PR #333, opened 2021-04-20
Use `non_witness_utxo` when making SegWit signatures to mitigate the "SegWit bug"

afilini:
### Description

Unless explicitly asked not to (by enabling a signer option), we will now by default require the presence of the `non_witness_utxo` when signing SegWit transactions as well, to mitigate the risk of an attacker tricking the wallet into paying more fees than it expects.

The `force_non_witness_utxo` option which was off by default has been replaced with `only_witness_utxo` which is also disabled by default, but since it's basically the opposite of the one we had previously, it effectively inverts the default behavior from not including the `non_witness_utxo` to providing it.

### Notes to the reviewers

This needed a bit of refactoring in the signing part to introduce some options that can be provided to tweak the behavior of the signer. I don't particularly like having that extra method on our `Signer` trait, but I couldn't come up with a better alternative.

#### All Submissions:

#### New Features:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -862,12 +862,10 @@ where
 
         // If we aren't allowed to use `witness_utxo`, ensure that every input has the
         // `non_witness_utxo`
-        if !sign_options.trust_witness_utxo {
-            for input in &psbt.inputs {
-                if input.non_witness_utxo.is_none() {
-                    return Err(Error::Signer(signer::SignerError::MissingNonWitnessUtxo));
-                }
-            }
+        if !sign_options.trust_witness_utxo
+            && psbt.inputs.iter().any(|i| i.non_witness_utxo.is_none())
+        {
+            return Err(Error::Signer(signer::SignerError::MissingNonWitnessUtxo));
         }
 
         for signer in self
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -46,7 +46,7 @@ pub use utils::IsDust;
 
 use address_validator::AddressValidator;
 use coin_selection::DefaultCoinSelectionAlgorithm;
-use signer::{Signer, SignerOrdering, SignersContainer};
+use signer::{SignOptions, Signer, SignerOrdering, SignersContainer};
 use tx_builder::{BumpFee, CreateTx, FeePolicy, TxBuilder, TxParams};
 use utils::{check_nlocktime, check_nsequence_rbf, After, Older, SecpCtx, DUST_LIMIT_SATOSHI};
 
@@ -706,7 +706,7 @@ where
     ///         .enable_rbf();
     ///     builder.finish()?
     /// };
-    /// let _ = wallet.sign(&mut psbt, None)?;
+    /// let _ = wallet.sign(&mut psbt, SignOptions::default())?;
     /// let tx = psbt.extract_tx();
     /// // broadcast tx but it's taking too long to confirm so we want to bump the fee
     /// let (mut psbt, _) =  {
@@ -716,7 +716,7 @@ where
     ///     builder.finish()?
     /// };
     ///
-    /// let _ = wallet.sign(&mut psbt, None)?;
+    /// let _ = wallet.sign(&mut psbt, SignOptions::default())?;
     /// let fee_bumped_tx = psbt.extract_tx();
     /// // broadcast fee_bumped_tx to replace original
     /// # Ok::<(), bdk::Error>(())
@@ -833,6 +833,11 @@ where
     /// Sign a transaction with all the wallet's signers, in the order specified by every signer's
     /// [`SignerOrdering`]
     ///
+    /// The [`SignOptions`] can be used to tweak the behavior of the software signers, and the way
+    /// the transaction is finalized at the end. Note that it can't be guaranteed that *every*
+    /// signers will follow the options, but the "software signers" (WIF keys and `xprv`) defined
+    /// in this library will.
+    ///
     /// ## Example
     ///
     /// ```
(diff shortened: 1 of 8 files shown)


src/wallet/mod.rs
```diff
@@ -982,6 +982,16 @@ where
         Ok((psbt, finished))
     }
 
+    /// Override the signing options for all the software signers
+    ///
+    /// Since the signers are wrapped in `Arc` they will only be updated if it's possible to take a
```
RCasatta: maybe better to fail if it's not possible to update then?
I see `Arc::get_mut` allows to check if there are shared references

src/wallet/signer.rs
```diff
@@ -427,17 +532,47 @@ impl SignersContainer {
     }
 }
 
+/// Options for a software signer
+///
+/// Adjust the behavior of our software signers, [`ExtendedKeySigner`] and [`PrivateKeySigner`]
+#[derive(Debug, Clone)]
```
RCasatta, line 433: Maybe we can add `Copy` since the struct content is small, will allow removing some explicit call to `clone` around

LLFourn:
Concept ACK. I think there might be a simpler approach.

1. Add another argument to `Wallet::sign` which indicates whether it should check for missing `non_witness_utxo`. Turning on this check should be decided dynamically by the caller because it depends on the context of the signing; it shouldn't be on the `Signer` I don't think. Consider adding a struct `SignOptions` passed to `Wallet::sign` with `assume_height` and `trust_witness_utxo` which is default false. Just do the check in `Wallet::sign`.
2. The "SegWit bug" only applies to `v0` outputs -- it's fixed in `v1` I think. So really you probably want three possible values here: (i) don't force, (ii) force v0 (iii) force all. I guess you need force all since some wallets may be stubborn about this or not upgraded.

afilini:
> Consider adding a struct SignOptions passed to Wallet::sign with assume_height and trust_witness_utxo which is default false

This was more or less how I implemented it initially, but there was the issue that signers like hardware wallets would basically ignore the options. I mean, I hope there's no way to ask an HW to sign a transaction while asking it to trust some parts without complaining :sweat_smile:  So I came to the conclusion that this is more of a Signer-specific thing, we can control it for our signers but not for all of them. Maybe a good compromise is to have a `set_options()` method on `Signer` that returns a `Result`, so you could try to set the options and it would work fine on our signers but fail on HWs.

> The "SegWit bug" only applies to v0 outputs -- it's fixed in v1 I think.

That's a good point

LLFourn:
> 
> 
> > Consider adding a struct SignOptions passed to Wallet::sign with assume_height and trust_witness_utxo which is default false
> 
> This was more or less how I implemented it initially, but there was the issue that signers like hardware wallets would basically ignore the options. I mean, I hope there's no way to ask an HW to sign a transaction while asking it to trust some parts without complaining 😅 So I came to the conclusion that this is more of a Signer-specific thing, we can control it for our signers but not for all of them. 

I think you misunderstood my suggestion: put the check in `Wallet::sign` before any signer is called to sign anything. To be clear:

```rust
    pub fn sign(&self, mut psbt: PSBT, sign_options: SignOptions) -> Result<(PSBT, bool), Error> {
        // this helps us doing our job later
        self.add_input_hd_keypaths(&mut psbt)?;
        
        if !sign_options.trust_witness_utxo {
             for input in psbt.input {
                 if input.non_witness_utxo.is_none() {
                     return Err(Error::MissingNonWitnessUtxo)
                  }
             }
        }

        for signer in self
            .signers
            .signers()
            .iter()
            .chain(self.change_signers.signers().iter())
        {
            if signer.sign_whole_tx() {
                signer.sign(&mut psbt, None, &self.secp)?;
            } else {
                for index in 0..psbt.inputs.len() {
                    signer.sign(&mut psbt, Some(index), &self.secp)?;
                }
            }
        }
```

This means it doesn't matter whether they are hardware wallets at all.

src/wallet/signer.rs
RCasatta: now the struct has more than one field

src/wallet/signer.rs
```diff
@@ -427,6 +427,43 @@ impl SignersContainer {
...
+    /// Whether the wallet should assume a specific height has been reached when trying to finalize
+    /// a transaction
+    ///
+    /// The wallet will only "use" a timelock to satisfy the spending policy of an input if the
+    /// timelock height has already been reached. This option allows overriding the "current height" to let the
+    /// wallet use timelocks in the future to spend a coin.
+    pub assume_height: Option<u32>,
+}
+
+impl Default for SignOptions {
```
RCasatta, line 458: This is equivalent to auto derive Default, do you prefer to be more explicit?

src/wallet/mod.rs
```diff
@@ -848,13 +853,23 @@ where
...
-    pub fn sign(&self, psbt: &mut PSBT, assume_height: Option<u32>) -> Result<bool, Error> {
+    pub fn sign(&self, psbt: &mut PSBT, sign_options: SignOptions) -> Result<bool, Error> {
         // this helps us doing our job later
         self.add_input_hd_keypaths(psbt)?;
 
+        // If we aren't allowed to use `witness_utxo`, ensure that every input has the
+        // `non_witness_utxo`
+        if !sign_options.trust_witness_utxo {
+            for input in &psbt.inputs {
+                if input.non_witness_utxo.is_none() {
```
RCasatta: we may want to avoid one nesting by using something like `if psbt.inputs.iter().map(|i| input.non_witness_uxo).any(Option::is_none)`

src/wallet/signer.rs
afilini: Yeah, because if somebody adds an option in the future they'll have to be explicit about the default value they want

src/wallet/mod.rs
afilini: Done in b1b1694

RCasatta (approved):

PR #338, opened 2021-05-06
Fix nightly Clippy warnings

tcharding:
### Description

Checking `bdk` with the nightly toolchain throws a bunch of warnings, ~all are trivial to fix~. Turns out one of them is not, the final patch of this PR may or may not be deemed an improvement, please review and I can drop if required.

Fix warnings emitted by Clippy when using nightly toolchain.

### Notes to the reviewers

While bdk does not appear to explicitly rely on nightly features I like to use nightly so I can pass `-Zunstable-options` to clippy. I do this so that Clippy doesn't hide warnings after the first time showing them, instead throwing warnings every check. 

#### All Submissions:

### DIFF
--- a/src/psbt/mod.rs
+++ b/src/psbt/mod.rs
@@ -17,6 +17,7 @@ pub trait PsbtUtils {
 }
 
 impl PsbtUtils for Psbt {
+    #[allow(clippy::all)] // We want to allow `manual_map` but it is too new.
     fn get_utxo_for(&self, input_index: usize) -> Option<TxOut> {
         let tx = &self.global.unsigned_tx;
 
--- a/src/descriptor/dsl.rs
+++ b/src/descriptor/dsl.rs
@@ -535,9 +535,7 @@ macro_rules! fragment_internal {
     ( @t , $( $tail:tt )* ) => ({
         $crate::fragment_internal!( @t $( $tail )* )
     });
-    ( @t ) => ({
-        ()
-    });
+    ( @t ) => ({});
 
     // Fallback to calling `fragment!()`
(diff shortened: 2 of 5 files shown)


src/psbt/mod.rs
```diff
@@ -24,16 +24,17 @@ impl PsbtUtils for PSBT {
...
         }
 
-        if let Some(input) = self.inputs.get(input_index) {
-            if let Some(wit_utxo) = &input.witness_utxo {
-                Some(wit_utxo.clone())
-            } else if let Some(in_tx) = &input.non_witness_utxo {
-                Some(in_tx.output[tx.input[input_index].previous_output.vout as usize].clone())
-            } else {
-                None
+        match self.inputs.get(input_index) {
```
RCasatta: I do agree this code looks less readable than before, if the specific lint is not usable maybe `#[allow(clippy::all)]` with a comment for this block?

tcharding: Cool, nice idea, I did not think of that. Will update PR as suggested. Thanks.

tcharding:
Please note, this PR now has an additional commit on the end of it (a new nightly Clippy warning snuck in during rebase on master).

notmandatory (approved):
Look like good cleanups and as usual good commit docs. ACK e1066e9

tcharding:
> Look like good cleanups and as usual good commit docs. ACK [e1066e9](https://github.com/bitcoindevkit/bdk/commit/e1066e955cf6c4fc0a20b4a895f064e70c696dae)

Thanks, appreciate the complement.  My commit log habits were formed in the fires of the linux-staging mailing list :)

PR #340, opened 2021-05-06
Update create transaction example code

tcharding:
### Description

The transaction builder changed a while ago, looks like some of the example code did not get updated.

Update the transaction creation code to use a mutable builder.

#### All Submissions:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -125,12 +125,15 @@
 //!     wallet.sync(noop_progress(), None)?;
 //!
 //!     let send_to = wallet.get_address(New)?;
-//!     let (psbt, details) = wallet.build_tx()
-//!         .add_recipient(send_to.script_pubkey(), 50_000)
-//!         .enable_rbf()
-//!         .do_not_spend_change()
-//!         .fee_rate(FeeRate::from_sat_per_vb(5.0))
-//!         .finish()?;
+//!     let (psbt, details) = {
+//!         let mut builder =  wallet.build_tx();
+//!         builder
+//!             .add_recipient(send_to.script_pubkey(), 50_000)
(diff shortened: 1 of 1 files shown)


afilini:
Looks like they are being ignored for some reason, can you also change them to `no_run` instead so that we can keep them updated going forward?

I did a quick test and there's also another test failing on line 145 which can be trivially fixed (replace `None` with `SignOptions::default()`)

tcharding:
Thanks for the review, will fix as suggested.

tcharding:
@afilini I fixed as suggested including changing all `ignore`s in `lib.rs` to `no_run`s (with code fixes). And, for the awesomely anal among us, I fixed the import lines to use `dbk` project style.

afilini:
I think the CI is failing because the tests use the electrum stuff but they are not feature-gated. It's kinda ugly but there's a way to fix it:

https://github.com/bitcoindevkit/bdk/blob/12641b9e8fa13cbc5eb34f55d18b423afc505aad/src/keys/mod.rs#L453-L472

afilini:
Since we have to do the release today and this isn't ready yet I've cherry-picked abbea4b on the release branch, so at least we will have the correct example on the published crate

tcharding:
Wow, this opens a whole can of worms. The `lib.rs` inner doc comment documents behaviour from the default set of features (as one would expect) but we cannot expect folks to build with the default set of features so we do not currently build these docs. Attempting to use the idea suggested above leads to this error:
```
note: inner doc comments like this (starting with `//!` or `/*!`) can only appear before items
```
I don't have the inclination to dig further into this at the moment. Closing this issue for another day.

PR #341, opened 2021-05-06
Check index out of bound also for tx inputs not only for psbt inputs

RCasatta:
### Description

A malformed PSBT, with more inputs in `psbt.inputs` than in `psbt.global.unsigned_tx.inputs` or viceversa can cause an out of bound panic

### Notes to the reviewers

Note this malformed PSBT is not deserialized from rust-bitcoin, failing with `ParseFailed("data not consumed entirely when explicitly deserializing")'` so it's probably impossible to raise the panic in normal usage condition, it's sane to have the check though.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/psbt/mod.rs
+++ b/src/psbt/mod.rs
@@ -41,6 +41,7 @@ impl PsbtUtils for PSBT {
 #[cfg(test)]
 mod test {
     use crate::bitcoin::consensus::deserialize;
+    use crate::bitcoin::TxIn;
     use crate::psbt::PSBT;
     use crate::wallet::test::{get_funded_wallet, get_test_wpkh};
     use crate::wallet::AddressIndex;
@@ -51,7 +52,7 @@ mod test {
 
     #[test]
     #[should_panic(expected = "InputIndexOutOfRange")]
-    fn test_psbt_malformed_legacy() {
+    fn test_psbt_malformed_psbt_input_legacy() {
         let psbt_bip: PSBT = deserialize(&base64::decode(PSBT_STR).unwrap()).unwrap();
         let (wallet, _, _) = get_funded_wallet(get_test_wpkh());
         let send_to = wallet.get_address(AddressIndex::New).unwrap();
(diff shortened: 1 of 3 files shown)


src/psbt/mod.rs
```diff
@@ -37,3 +37,66 @@ impl PsbtUtils for PSBT {
...
+        let send_to = wallet.get_address(AddressIndex::New).unwrap();
+        let mut builder = wallet.build_tx();
+        builder.add_recipient(send_to.script_pubkey(), 10_000);
+        let (mut psbt, _) = builder.finish().unwrap();
+        psbt.inputs.push(psbt_bip.inputs[0].clone());
+        let options = SignOptions {
+            trust_witness_utxo: true,
+            assume_height: None,
+        };
+        let _ = wallet.sign(&mut psbt, options).unwrap();
```
tcharding, line 67: Relying on this `unwrap` to indicate that the test passes while having `unwrap` calls further up the test can give a false positive i.e., the test can pass due to one of the earlier `unwrap`s panicking.

Using an assertion on `is_err` and removing the `should_panic` makes the test more robust.

```
assert!(wallet.sign(&mut psbt, options).is_err());
```
But looses the error type check, I came up with the rather ugly:
```
        match wallet.sign(&mut psbt, options) {
            Err(Error::Signer(e)) => assert_eq!(e, SignerError::InputIndexOutOfRange),
            Err(e) => panic!("Wrong error type: {:?}", e),
            Ok(_) => panic!("sign() should have failed"),
        }
```

tcharding (commented):
Small comment on the testing, feel free to ignore me :)

src/psbt/mod.rs
RCasatta: Yes, you are right, but it's a lot of code...

If we want to be more precise, I think we should consider using https://docs.rs/crate/assert_matches/1.5.0 as a dev-dep which will also be standardized one day  https://doc.rust-lang.org/nightly/std/macro.assert_matches.html

it would become something like:
`assert_matches!(wallet.sign(&mut psbt, options).unwrap_err(), SignerError::InputIndexOutOfRange)`

@afilini ?

afilini: I think we have a lot of tests that use the `should_panic`, so I wouldn't bother doing anything different here.

It makes sense to add `assert_matches`, but that should be a different PR that also updates all the other tests.

Also ideally we should avoid unwrapping completely if we use it, so the check could be against the `Result` as well instead of just the error variant:

`assert_matches!(wallet.sign(&mut psbt, options), Err(Error::Signer(SignerError::InputIndexOutOfRange)))`

afilini (approved):
I'm gonna merge this to `master` and the current release branch as well

PR #347, opened 2021-05-17
Put testutils inside main crate

LLFourn:
This means one less crate in the repo. I started off trying to put everything inside the main crate but got a greater appreciation of why it was separated in the first place. I was at least was able to remove the extra `testutils-macros` crate.

#### All Submissions:


Cargo.toml
```diff
@@ -54,13 +54,11 @@ all-keys = ["keys-bip39"]
...
 
 # Debug/Test features
-debug-proc-macros = ["bdk-macros/debug", "bdk-testutils-macros/debug"]
 test-electrum = ["electrum"]
 test-md-docs = ["electrum"]
 
 [dev-dependencies]
-bdk-testutils = "0.4"
-bdk-testutils-macros = "0.6"
+bdk-testutils = { path = "./testutils" }
```
LLFourn: note I made this relative again :)

src/blockchain/electrum.rs
```diff
@@ -47,9 +47,11 @@ pub struct ElectrumBlockchain(Client);
...
 #[cfg(test)]
 #[cfg(feature = "test-electrum")]
-#[bdk_blockchain_tests(crate)]
-fn local_electrs() -> ElectrumBlockchain {
-    ElectrumBlockchain::from(Client::new(&testutils::get_electrum_url()).unwrap())
+testutils::bdk_blockchain_tests! {
+    bdk => crate,
+    fn test_instance() -> ElectrumBlockchain {
+        ElectrumBlockchain::from(Client::new(&testutils::get_electrum_url()).unwrap())
+    }
```
LLFourn: Maybe this should just be at the end of the file I see no reason to clutter the start of it up here.

testutils/src/lib.rs
```diff
@@ -562,3 +564,12 @@ impl Deref for TestClient {
...
 }
+
+impl Default for TestClient {
+    fn default() -> Self {
+        let rpc_host_and_port =
+            env::var("BDK_RPC_URL").unwrap_or_else(|_| "127.0.0.1:18443".to_string());
+        let wallet = env::var("BDK_RPC_WALLET").unwrap_or_else(|_| "bdk-test".to_string());
+        Self::new(rpc_host_and_port, wallet)
+    }
+}
```
LLFourn: note this had to be done because of clippy.

LLFourn:
I just took this PR a lot further:

1. We now run tests against esplora as well in ci (except for the reorg test which doesn't work -- will open issue when merged)
2. All of testutils is just inside bdk now so this PR now removes two crates and therefore makes it easy to keep apis in sync between the tests and the lib. You can access the test macro from another crate with the `test-blockchains` feature. This also enables the tests in the local esplora and electrum implementations.
3. I wrote a shell script to actually run the tests and put it in root -- I doubt it will live for very long but it is at least useful for me locally so I thought I should include it.

run_blockchain_tests.sh
```diff
@@ -0,0 +1,54 @@
...
+export BDK_RPC_PASS=passw
+export BDK_RPC_URL=127.0.0.1:18443
+export BDK_RPC_WALLET=bdk-test
+export BDK_ELECTRUM_URL=tcp://127.0.0.1:60401
+
+cli(){
+    docker exec -it "$id" /root/bitcoin-cli -regtest -rpcuser=admin -rpcpassword=passw $@
+}
+
+eprintln "running getwalletinfo unitl bitcoind seems to be alive"
```
RCasatta: nit: "unitl" typo

RCasatta (approved):
I rebased the RPC PR on this so that I can more easily add tests for RPC and while none of the tests for RPC is working yet this LGTM

I find useful during development the ability to turn on logging that could be achieved by introducing `pub static INIT_LOGGING: Once = Once::new();` in blockchain_tests.rs and something like 
```
if std::env::var("RUST_LOG").is_ok() {
            crate::testutils::blockchain_tests::INIT_LOGGING.call_once(|| {
                env_logger::init();
            });
        }
```
in test_instance

also it would be nice if `run_blockchain_tests.sh` could take a test name so that is filtered for that

If you want to add in this PR this changes otherwise I can add in mine

LLFourn:
@RCasatta Good ideas. with env logger I just did `try_init` in the test init function which is idempotent. This seems to work.

`run_blockchain_tests.sh` has a usage now and can take an optional test name as an argument.

Issue #86, opened 2020-07-24
Provide easy way for wallet to connect to your own full node


moneyball:
I suggest adding this to the roadmap on the website.

Inspired by this tweet https://twitter.com/matt_odell/status/1286659125322162176?s=19

"All wallets should have the option to use your own node easily by simply scanning a QR code or pasting a connection code."

The BDK can build this and make it easy-to-use / default for any wallet based on BDK.

tcharding:
Can we close this issue since `RpcBlockchain` exists now?

PR #22, opened 2020-06-28
Bitcoin RPC Blockchain

justinmoon:

PR #348, opened 2021-05-17
Bitcoin core RPC backend

### LINKED ISSUE #79: Bitcoin core RPC client
https://github.com/rust-bitcoin/rust-bitcoincore-rpc

RCasatta:
updated version of #22 

closes #79 and #86

~~needs~~ rebased on top of #347 for common blockchain tests

### Description

Implements Blockchain trait for a backend using an RPC node

### Notes to the reviewers

Note we are obliged to use importmulti because bitcoin core doesn't support all descriptor https://bitcoindevkit.org/descriptors/#compatibility-matrix

use bitcoind crate as dev-dependency, requires env var pointing to a bitcoind executable, see for example https://github.com/RCasatta/bitcoind/blob/master/.github/workflows/test.yml

#### All Submissions:

#### New Features:

### DIFF
--- a/src/blockchain/mod.rs
+++ b/src/blockchain/mod.rs
@@ -43,6 +43,13 @@ pub use self::electrum::ElectrumBlockchain;
 #[cfg(feature = "electrum")]
 pub use self::electrum::ElectrumBlockchainConfig;
 
+#[cfg(feature = "rpc")]
+pub mod rpc;
+#[cfg(feature = "rpc")]
+pub use self::rpc::RpcBlockchain;
+#[cfg(feature = "rpc")]
+pub use self::rpc::RpcConfig;
+
 #[cfg(feature = "esplora")]
 #[cfg_attr(docsrs, doc(cfg(feature = "esplora")))]
 pub mod esplora;
@@ -52,6 +59,7 @@ pub use self::esplora::EsploraBlockchain;
 #[cfg(feature = "compact_filters")]
 #[cfg_attr(docsrs, doc(cfg(feature = "compact_filters")))]
 pub mod compact_filters;
+
 #[cfg(feature = "compact_filters")]
 pub use self::compact_filters::CompactFiltersBlockchain;
 
--- a/src/blockchain/rpc.rs
+++ b/src/blockchain/rpc.rs
@@ -280,11 +280,7 @@ impl Blockchain for RpcBlockchain {
     }
 
     fn get_tx(&self, txid: &Txid) -> Result<Option<Transaction>, Error> {
-        if self.capabilities.contains(&Capability::FullHistory) {
-            Ok(Some(self.client.get_raw_transaction(txid, None)?))
-        } else {
-            Ok(None)
-        }
+        Ok(Some(self.client.get_raw_transaction(txid, None)?))
     }
 
     fn broadcast(&self, tx: &Transaction) -> Result<(), Error> {
--- a/src/blockchain/rpc.rs
+++ b/src/blockchain/rpc.rs
@@ -58,7 +58,7 @@ pub struct RpcBlockchain {
     skip_blocks: Option<u32>,
 
     /// This is a fixed Address used as a hack key to store information on the node
-    _satoshi_address: Address,
+    _storage_address: Address,
 }
 
 /// RpcBlockchain configuration options
@@ -78,7 +78,7 @@ pub struct RpcConfig {
 
 impl RpcBlockchain {
     fn get_node_synced_height(&self) -> Result<u32, Error> {
-        let info = self.client.get_address_info(&self._satoshi_address)?;
+        let info = self.client.get_address_info(&self._storage_address)?;
         if let Some(GetAddressInfoResultLabel::Simple(label)) = info.labels.first() {
             Ok(label
                 .parse::<u32>()
@@ -93,7 +93,7 @@ impl RpcBlockchain {
     fn set_node_synced_height(&self, height: u32) -> Result<(), Error> {
         Ok(self
             .client
-            .set_label(&self._satoshi_address, &height.to_string())?)
+            .set_label(&self._storage_address, &height.to_string())?)
     }
 }
 
@@ -354,14 +354,15 @@ impl ConfigurableBlockchain for RpcBlockchain {
         }
 
         // this is just a fixed address used only to store a label containing the synced height in the node
-        let mut satoshi_address = Address::from_str("1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa").unwrap();
-        satoshi_address.network = network;
+        let mut storage_address =
+            Address::from_str("bc1qst0rewf0wm4kw6qn6kv0e5tc56nkf9yhcxlhqv").unwrap();
+        storage_address.network = network;
 
         Ok(RpcBlockchain {
             client,
             network,
             capabilities,
-            _satoshi_address: satoshi_address,
+            _storage_address: storage_address,
             skip_blocks: config.skip_blocks,
         })
     }
--- a/src/blockchain/rpc.rs
+++ b/src/blockchain/rpc.rs
@@ -242,17 +242,22 @@ impl Blockchain for RpcBlockchain {
             }
         }
 
-        let current_utxos: HashSet<LocalUtxo> = current_utxo
+        let current_utxos: HashSet<_> = current_utxo
             .into_iter()
-            .map(|u| LocalUtxo {
-                outpoint: OutPoint::new(u.txid, u.vout),
-                txout: TxOut {
-                    value: u.amount.as_sat(),
-                    script_pubkey: u.script_pub_key,
-                },
-                keychain: KeychainKind::External,
+            .map(|u| {
+                Ok(LocalUtxo {
+                    outpoint: OutPoint::new(u.txid, u.vout),
+                    keychain: db
+                        .get_path_from_script_pubkey(&u.script_pub_key)?
+                        .ok_or(Error::TransactionNotFound)?
+                        .0,
+                    txout: TxOut {
+                        value: u.amount.as_sat(),
+                        script_pubkey: u.script_pub_key,
+                    },
+                })
             })
-            .collect();
+            .collect::<Result<_, Error>>()?;
 
         let spent: HashSet<_> = known_utxos.difference(&current_utxos).collect();
         for s in spent {
--- a/src/blockchain/rpc.rs
+++ b/src/blockchain/rpc.rs
@@ -471,7 +471,7 @@ mod test {
 
         let config = RpcConfig {
             url: bitcoind.rpc_url(),
-            auth: Auth::CookieFile(bitcoind.cookie_file.clone()),
+            auth: Auth::CookieFile(bitcoind.config.cookie_file.clone()),
             network,
             wallet_name,
             skip_blocks: None,
@@ -523,7 +523,7 @@ mod test {
         generate(&bitcoind, 5);
         let config = RpcConfig {
             url: bitcoind.rpc_url(),
-            auth: Auth::CookieFile(bitcoind.cookie_file.clone()),
+            auth: Auth::CookieFile(bitcoind.config.cookie_file.clone()),
             network: Network::Regtest,
             wallet_name: "another-name".to_string(),
             skip_blocks: Some(103),
--- a/src/blockchain/rpc.rs
+++ b/src/blockchain/rpc.rs
@@ -1,0 +1,664 @@
+// Bitcoin Dev Kit
+// Written in 2021 by Riccardo Casatta <riccardo@casatta.it>
+//
+// Copyright (c) 2020-2021 Bitcoin Dev Kit Developers
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! Rpc Blockchain
+//!
+//! Backend that gets blockchain data from Bitcoin Core RPC
+//!
+//! ## Example
+//!
+//! ```no_run
+//! # use bdk::blockchain::{RpcConfig, RpcBlockchain, ConfigurableBlockchain};
+//! let config = RpcConfig {
+//!             url: "127.0.0.1:18332".to_string(),
+//!             auth: bitcoincore_rpc::Auth::CookieFile("/home/user/.bitcoin/.cookie".into()),
+//!             network: bdk::bitcoin::Network::Testnet,
+//!             wallet_name: "wallet_name".to_string(),
+//!             skip_blocks: None,
+//!         };
+//! let blockchain = RpcBlockchain::from_config(&config);
+//! ```
+
+use crate::bitcoin::consensus::deserialize;
+use crate::bitcoin::{Address, Network, OutPoint, Transaction, TxOut, Txid};
+use crate::blockchain::{Blockchain, Capability, ConfigurableBlockchain, Progress};
+use crate::database::{BatchDatabase, DatabaseUtils};
+use crate::descriptor::{get_checksum, IntoWalletDescriptor};
+use crate::wallet::utils::SecpCtx;
+use crate::{Error, FeeRate, KeychainKind, LocalUtxo, TransactionDetails};
+use bitcoincore_rpc::json::{
+    GetAddressInfoResultLabel, ImportMultiOptions, ImportMultiRequest,
+    ImportMultiRequestScriptPubkey, ImportMultiRescanSince,
+};
+use bitcoincore_rpc::jsonrpc::serde_json::Value;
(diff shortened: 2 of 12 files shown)


jb55:
oh hi, deja vu! I need this for a project, trying it now...

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,522 @@
...
+use crate::blockchain::{Blockchain, Capability, ConfigurableBlockchain, Progress};
+use crate::database::BatchDatabase;
+use crate::descriptor::{get_checksum, ExtendedDescriptor};
+use crate::{Error, FeeRate, KeychainKind, LocalUtxo, TransactionDetails};
+use bitcoincore_rpc::json::{
+    GetAddressInfoResultLabel, ImportMultiOptions, ImportMultiRequest,
+    ImportMultiRequestScriptPubkey, ImportMultiRescanSince,
+};
+use bitcoincore_rpc::jsonrpc::serde_json::Value;
+use bitcoincore_rpc::{Auth, Client, RpcApi};
```
jb55, line 42: I think we should re-export `bitcoincore_rpc` in `src/lib.rs` ? `RpcConfig` depends on `Auth` so you can't construct it otherwise.

```patch
diff --git a/src/lib.rs b/src/lib.rs
index f5323b1..477f023 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -216,6 +216,9 @@ extern crate bdk_macros;
 #[cfg(feature = "compact_filters")]
 extern crate lazy_static;
 
+#[cfg(feature = "rpc")]
+pub extern crate bitcoincore_rpc;
+
 #[cfg(feature = "electrum")]
 pub extern crate electrum_client;
```

jb55 (changes requested):

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,522 @@
...
+        db: &mut D,
+        progress_update: P,
+    ) -> Result<(), Error> {
+        debug!("sync");
+        let current_height = self.get_height()?;
+        let node_synced = self.get_node_synced_height()?;
+
+        //TODO if current_height == node_synced should check only the mempool
+
+        //TODO split the interval in chunk so that we can give progress update
```
jb55: instead of start/stop params which involves state, we could just periodically call `getwalletinfo` for the progress.

src/blockchain/rpc.rs
RCasatta: sure, adding...

Before final merge maybe we can assess if only Auth is needed we may want to import only that struct

src/blockchain/rpc.rs
RCasatta: yes, that is better

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,565 @@
...
+    ) -> Result<(), Error> {
+        debug!("sync");
+        let current_height = self.get_height()?;
+        let node_synced = self.get_node_synced_height()?;
+
+        //TODO if current_height == node_synced should check only the mempool
+
+        //TODO if it's a big rescan, spawn a thread that checks getwalletinfo and update progress
+        self.client
+            .rescan_blockchain(Some(node_synced as usize), Some(current_height as usize))?;
```
jb55, line 146: one thing I noticed is that if this call is ever interrupted during a long rescan from 0, `set_node_synced_height` is never called, and then the next time sync happens it will rescan from 0 again. not sure what the best way to handle this is.

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,565 @@
...
+    /// The bitcoin node authentication mechanism
+    pub auth: Auth,
+    /// The network we are using (it will be checked the bitcoin node network matches this)
+    pub network: Network,
+    /// The wallet name in the bitcoin node, consider using [wallet_name_from_descriptor] for this
+    pub wallet_name: String,
+}
+
+impl RpcBlockchain {
+    fn get_node_synced_height(&self) -> Result<u32, Error> {
```
jb55, line 80: what do you think about making these methods public? I was able to implement a "quick sync" for new wallets by setting node_synced_height to the current tip.

src/blockchain/rpc.rs
RCasatta: Yeah, right, maybe it's better to return to the initially proposed logic of doing the rescan in chunks, for example, every 10_000 blocks, this would solve both the progress (without a separated thread for the `getwalletinfo` calss) and make the restart from crash less bad

src/blockchain/rpc.rs
RCasatta: Maybe it should be a config parameter?

in compact_filters is passed at creation time https://docs.rs/bdk/0.7.0/bdk/blockchain/compact_filters/struct.CompactFiltersBlockchain.html#method.new

jb55:
>Yeah, right, maybe it's better to return to the initially proposed
>logic of doing the rescan in chunks, for example, every 10_000 blocks,
>this would solve both the progress (without a separated thread for the
>`getwalletinfo` calss) and make the restart from crash less bad

yup I came to the same conclusion after thinking about it some more 👍

jb55:
>Maybe it should be a config parameter?
>
>in compact_filters is passed at creation time https://docs.rs/bdk/0.7.0/bdk/blockchain/compact_filters/struct.CompactFiltersBlockchain.html#method.new

either way works for me! having a similar interface to the other backends sounds like a good thing.

RCasatta:
rebased on top of #347

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,617 @@
...
+            network: Network::Regtest,
+            wallet_name: "another-name".to_string(),
+            skip_blocks: Some(103),
+        };
+        let blockchain_skip = RpcBlockchain::from_config(&config).unwrap();
+        let db = MemoryDatabase::new();
+        let wallet_skip =
+            Wallet::new(DESCRIPTOR_PRIV, None, Network::Regtest, db, blockchain_skip).unwrap();
+        send_to_address(&bitcoind, &address, 100_000);
+        generate(&bitcoind, 1); // TODO why this is needed even if list_unspent should include zero conf and unsafe?
```
RCasatta: Does anybody know why this `generate` is needed? The `list_unspent` on line 154 should include unconfirmed...

jb55:
we should export this under the blockchain module like the others:

```diff
diff --git a/src/blockchain/mod.rs b/src/blockchain/mod.rs
index 95069ec..b9f5e3c 100644
--- a/src/blockchain/mod.rs
+++ b/src/blockchain/mod.rs
@@ -43,6 +43,11 @@ pub use self::electrum::ElectrumBlockchain;
 #[cfg(feature = "electrum")]
 pub use self::electrum::ElectrumBlockchainConfig;
 
+#[cfg(feature = "rpc")]
+pub use self::rpc::RpcBlockchain;
+#[cfg(feature = "rpc")]
+pub use self::rpc::RpcConfig;
+
 #[cfg(feature = "esplora")]
 #[cfg_attr(docsrs, doc(cfg(feature = "esplora")))]
 pub mod esplora;
                                                               
```

RCasatta:
Rebased on master and squashed

src/blockchain/rpc.rs
RCasatta: problem was that a `wallet.sync` is necessary before the transaction arrive in the mempool

src/blockchain/rpc.rs
RCasatta: reconsidered again, see https://github.com/bitcoindevkit/bdk/pull/348#discussion_r636016580

src/blockchain/mod.rs
```diff
@@ -43,6 +43,13 @@ pub use self::electrum::ElectrumBlockchain;
 #[cfg(feature = "electrum")]
 pub use self::electrum::ElectrumBlockchainConfig;
 
+#[cfg(feature = "rpc")]
+pub mod rpc;
+#[cfg(feature = "rpc")]
+pub use self::rpc::RpcBlockchain;
+#[cfg(feature = "rpc")]
+pub use self::rpc::RpcConfig;
```
afilini, line 51: The new RPC variant should be added to `src/blockchain/any.rs` as well

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,664 @@
...
+                client.create_wallet(&wallet_name, Some(true), None, None, None)?;
+                debug!("wallet created {:?}", wallet_name);
+            }
+        }
+
+        let blockchain_info = client.get_blockchain_info()?;
+        let network = match blockchain_info.chain.as_str() {
+            "main" => Network::Bitcoin,
+            "test" => Network::Testnet,
+            "regtest" => Network::Regtest,
```
afilini, line 339: Is signet supported as well?

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,664 @@
...
+        if rpc_version >= 210_000 {
+            let info: HashMap<String, Value> = client.call("getindexinfo", &[]).unwrap();
+            if info.contains_key("txindex") {
+                capabilities.insert(Capability::GetAnyTx);
+                capabilities.insert(Capability::AccurateFees);
+            }
+        }
+
+        // this is just a fixed address used only to store a label containing the synced height in the node
+        let mut satoshi_address = Address::from_str("1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa").unwrap();
```
afilini: Not sure if setting a label means that the node will start keeping track of the address, but if that's the case it would be better to use a different empty address since this one has a lot of utxos.

afilini: Also I think it's kinda confusing to call this "staoshi_address", maybe "storage_address" is a bit easier to understand

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,664 @@
...
+    #[derive(Deserialize)]
+    struct Name {
+        name: String,
+    }
+    #[derive(Deserialize)]
+    struct Result {
+        wallets: Vec<Name>,
+    }
+
+    let result: Result = client.call("listwalletdir", &[])?;
```
afilini: Maybe calling this `Result` isn't ideal because it shadows the `Result` from std. `CallResult` should be fine.

src/testutils/blockchain_tests.rs
```diff
@@ -370,6 +369,10 @@ macro_rules! bdk_blockchain_tests {
                 let test_client = TestClient::default();
                 let wallet = get_wallet_from_descriptors(&descriptors);
 
+                // rpc need to call import_multi before receiving any tx, otherwise will not see tx in the mempool
+                #[cfg(feature = "rpc")]
```
afilini: I think here you can use `test-rpc` instead, which at least won't slow down all the other tests only if `rpc` is enabled

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,664 @@
...
+        }
+
+        //TODO call rescan in chunks (updating node_synced_height) so that in case of
+        // interruption work can be partially recovered
+        debug!(
+            "rescan_blockchain from:{} to:{}",
+            node_synced, current_height
+        );
+        self.client
+            .rescan_blockchain(Some(node_synced as usize), Some(current_height as usize))?;
```
afilini, line 146: Maybe it's better to go back a few blocks in case of reorgs. Like, rescan from node_synced - 10 or even 100 if we really want to be paranoid

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,664 @@
...
+        //TODO supported up to 1_000 txs, should use since_blocks or do paging
+        let list_txs = self
+            .client
+            .list_transactions(None, Some(1_000), None, Some(true))?;
+        let mut list_txs_ids = HashSet::new();
+
+        for tx_result in list_txs.iter().filter(|t| {
+            // list_txs returns all conflicting tx we want to
+            // filter out replaced tx => unconfirmed and not in the mempool
+            t.info.confirmations > 0 || self.client.get_mempool_entry(&t.info.txid).is_ok()
```
afilini, line 187: Maybe here we should differentiate between the tx not being in the mempool and other random errors like networking issues

afilini: If that's not too hard

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,664 @@
...
+                    }
+                }
+
+                let td = TransactionDetails {
+                    transaction: Some(tx),
+                    txid: tx_result.info.txid,
+                    timestamp: tx_result.info.time,
+                    received,
+                    sent,
+                    fees: tx_result.fee.map(|f| f.as_sat().abs() as u64).unwrap_or(0), //TODO
```
afilini: Is the `TODO` here still relevant?

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,664 @@
...
+
+        let current_utxos: HashSet<LocalUtxo> = current_utxo
+            .into_iter()
+            .map(|u| LocalUtxo {
+                outpoint: OutPoint::new(u.txid, u.vout),
+                txout: TxOut {
+                    value: u.amount.as_sat(),
+                    script_pubkey: u.script_pub_key,
+                },
+                keychain: KeychainKind::External,
```
afilini: Is it correct here to hardcode the external keychain?

src/blockchain/rpc.rs
```diff
@@ -0,0 +1,664 @@
...
+            debug!("{:?} max {}", keykind, index);
+            db.set_last_index(keykind, index)?;
+        }
+
+        self.set_node_synced_height(current_height)?;
+        Ok(())
+    }
+
+    fn get_tx(&self, txid: &Txid) -> Result<Option<Transaction>, Error> {
+        if self.capabilities.contains(&Capability::FullHistory) {
```
afilini: The docs for `get_tx` are kinda generic and just say "fetch a tx given its txid". Since it's not guaranteed that the tx is part of the wallet, I would check for the `GetAnyTx` capability instead.

Also, maybe it's worth just trying with the node instead of avoiding the req entirely: for unconfirmed txs in its mempool the node should be able to reply even if it doesn't have `txindex` enabled.

src/blockchain/mod.rs
RCasatta: Yes, right, but I think we need to defer this (maybe following PR) until https://github.com/rust-bitcoin/rust-bitcoincore-rpc/pull/181 is merged and released (initial work for reference https://github.com/bitcoindevkit/bdk/commit/d13b9259d880c79aafb97c8cfa0ea6483389fda5)

src/blockchain/rpc.rs
RCasatta: This would probably slow every sync call without real reason? The node keep track of the addresses in case of reorgs without an explicit rescan.

So it makes me think this should probably go only in the setup call...

src/testutils/blockchain_tests.rs
RCasatta: changed in e1b037a, it required some changes in launch scripts

src/blockchain/rpc.rs
RCasatta: changed in 9b7ed08

src/blockchain/rpc.rs
RCasatta: changed in dffb753

src/blockchain/rpc.rs
RCasatta: no, fixed in 8185119

src/blockchain/rpc.rs
RCasatta: It's there because I have some doubt on the `unwrap_or(0)` I think there are cases where the node cannot compute the fees and we show 0 but it's unknown, maybe we should have TransactionDetails::fee field Option?

src/blockchain/rpc.rs
RCasatta: Not sure we can discriminate between random errors and not being in the mempool...

src/blockchain/rpc.rs
RCasatta: changed in ae5aa06

renamed and used a vanity address :) -> bc1qst0rewf0wm4kw6qn6kv0e5tc56nkf9yhcxlhqv
I'll send you the private in case someone decides to send btc to it

src/blockchain/rpc.rs
RCasatta: changed in ab98283

src/blockchain/rpc.rs
afilini: New bdk donation address! ahaha

src/blockchain/rpc.rs
afilini: Right, the rescan should only be required when importing new addresses, which is kinda the definition of what `setup` should do. I think you can just move this after the importmulti and it should work

RCasatta: 1639984b5658f22002f1478cc93791db963bca4e

src/blockchain/rpc.rs
RCasatta: ba2e3042cca6ba3dd4cf2b58880bdc759c3607a3

afilini (approved):

PR #349, opened 2021-05-17
Return AddressInfo from get_address to allow access to index

futurepaul:
This adds an `AddressInfo` struct that derefs to `Address` so that wallet's `get_address` method can return both an index and an address without breaking existing code (hopefully!).

The motivation is to be able to produce descriptors to pass to HWI for purpose of verifying an address. Without BDK telling me the index I would have to keep track of that state on my own.

This was discussed on Discord as preferable to creating an additional method (something like `get_address_and_index`) just for this purpose.

An even nicer outcome would be the ability to query the database for a list of previously seen addresses and indexes, but that will involve a lot of database stuff that I'm not familiar with.

### Notes to the reviewers

I'm a little uncertain on how verbose to be with documenting this struct.

Also I got two unrelated clippy errors. Not sure if I should fix those or rebase once they're fixed elsewhere.

#### All Submissions:

#### New Features:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -227,7 +227,7 @@ where
     D: BatchDatabase,
 {
     // Return a newly derived address using the external descriptor
-    fn get_new_address(&self) -> Result<(Address, u32), Error> {
+    fn get_new_address(&self) -> Result<AddressInfo, Error> {
         let incremented_index = self.fetch_and_increment_index(KeychainKind::External)?;
 
         let address_result = self
@@ -236,13 +236,16 @@ where
             .address(self.network);
 
         address_result
-            .map(|address| (address, incremented_index))
+            .map(|address| AddressInfo {
+                address,
+                index: incremented_index,
+            })
             .map_err(|_| Error::ScriptDoesntHaveAddressForm)
     }
 
     // Return the the last previously derived address if it has not been used in a received
     // transaction. Otherwise return a new address using [`Wallet::get_new_address`].
-    fn get_unused_address(&self) -> Result<(Address, u32), Error> {
+    fn get_unused_address(&self) -> Result<AddressInfo, Error> {
         let current_index = self.fetch_index(KeychainKind::External)?;
 
         let derived_key = self.descriptor.as_derived(current_index, &self.secp);
@@ -261,29 +264,32 @@ where
         } else {
             derived_key
                 .address(self.network)
-                .map(|address| (address, current_index))
+                .map(|address| AddressInfo {
+                    address,
+                    index: current_index,
+                })
                 .map_err(|_| Error::ScriptDoesntHaveAddressForm)
         }
     }
 
     // Return derived address for the external descriptor at a specific index
-    fn peek_address(&self, index: u32) -> Result<(Address, u32), Error> {
+    fn peek_address(&self, index: u32) -> Result<AddressInfo, Error> {
         self.descriptor
             .as_derived(index, &self.secp)
             .address(self.network)
-            .map(|address| (address, index))
+            .map(|address| AddressInfo { address, index })
             .map_err(|_| Error::ScriptDoesntHaveAddressForm)
     }
 
     // Return derived address for the external descriptor at a specific index and reset current
     // address index
-    fn reset_address(&self, index: u32) -> Result<(Address, u32), Error> {
+    fn reset_address(&self, index: u32) -> Result<AddressInfo, Error> {
         self.set_index(KeychainKind::External, index)?;
 
         self.descriptor
             .as_derived(index, &self.secp)
             .address(self.network)
-            .map(|address| (address, index))
+            .map(|address| AddressInfo { address, index })
             .map_err(|_| Error::ScriptDoesntHaveAddressForm)
     }
 
@@ -291,14 +297,12 @@ where
     /// available address index selection strategies. If none of the keys in the descriptor are derivable
     /// (ie. does not end with /*) then the same address will always be returned for any [`AddressIndex`].
     pub fn get_address(&self, address_index: AddressIndex) -> Result<AddressInfo, Error> {
-        let result = match address_index {
+        match address_index {
             AddressIndex::New => self.get_new_address(),
             AddressIndex::LastUnused => self.get_unused_address(),
             AddressIndex::Peek(index) => self.peek_address(index),
             AddressIndex::Reset(index) => self.reset_address(index),
-        };
-
-        result.map(|(address, index)| AddressInfo { index, address })
+        }
     }
 
     /// Return whether or not a `script` is part of this wallet (either internal or external)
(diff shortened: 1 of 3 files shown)


notmandatory:
Overall looks good! but have a couple things to do to fix the electrum integration tests:

1. Change Cargo.toml [dev-dependencies]:
   ```
   bdk-testutils-macros = { path = "testutils-macros"}
   ```

2. Change testutils-macros/src/lib.rs line 536. There may be a better way to fix this but seem to be because the result get wrapped in a Some() so the auto deref() doesn't work:
   ```
   fn test_sync_receive_coinbase() {
      let (wallet, descriptors, mut test_client) = init_single_sig();
      let wallet_addr = wallet.get_address(New).unwrap().address;
     ...
   ```

Also I verified clippy works with rust `1.51.0` (our current CI version) and no warnings, maybe you are using a newer rust version?

Finally we'd like to have all `bdk` commits signed, see GitHub ["Signing Commits"](https://docs.github.com/en/github/authenticating-to-github/signing-commits) doc if you haven't done it before. Feel free to rebase/force-push this PR's branch once you get the commits signed. If you run into any issues I'm happy to help.

src/wallet/mod.rs
```diff
@@ -196,24 +197,52 @@ pub enum AddressIndex {
...
 
-        self.descriptor
+        let address_result = self
+            .descriptor
             .as_derived(incremented_index, &self.secp)
-            .address(self.network)
+            .address(self.network);
+
+        address_result
+            .map(|address| (address, incremented_index))
```
futurepaul: I just realized I don't have to return a tuple here, I could return `AddressInfo` as well (in all four of these private methods). Thoughts on what would be preferable?

futurepaul:
Ah, my bad on misunderstanding what "sign" commits means.

The fact that this test breaks implies to me that there will be other current uses of `get_address` that will break as well?

Are we okay with this being a breaking change? Seems to hinge on if `AddressInfo` is the ideal thing going forward or if there will be another more preferable way of getting the index / descriptor for a given address.

afilini:
> Are we okay with this being a breaking change?

Personally I would say yes, I for sure like this much more than having a separate method to get address + index.

Maybe we can leave this open for a while to let everybody share their ideas, but in principle I'm not entirely against breaking the API (we've actually been doing it with every release so far, even if only in minor ways generally). It's still way too early to commit to a stable API forever IMHO.

testutils-macros/src/lib.rs
```diff
@@ -540,7 +540,7 @@ pub fn bdk_blockchain_tests(attr: TokenStream, item: TokenStream) -> TokenStream
 
                     test_client.generate(1, Some(wallet_addr));
 
-                    wallet.sync(noop_progress(), None).unwrap();
+                    wallet.sync(noop_progress(), None).unwrap().address;
```
notmandatory: oops I think you added `.address` to the wrong line, it should be on 536 above:
```rust
let wallet_addr = wallet.get_address(New).unwrap().address;
```

futurepaul: Ah you're right.

When I fix that I end up with a lot of broken tests, and it seems like the problem is coming from within the `testutils` macro:

```
test blockchain::electrum::generated_tests_local_electrs::test_sync_before_and_after_receive ... FAILED
test blockchain::electrum::generated_tests_local_electrs::test_sync_bump_fee_add_input ... FAILED
test blockchain::electrum::generated_tests_local_electrs::test_sync_bump_fee_add_input_no_change ... FAILED
test blockchain::electrum::generated_tests_local_electrs::test_sync_bump_fee_remove_change ... FAILED
test blockchain::electrum::generated_tests_local_electrs::test_sync_long_change_chain ... FAILED
```

```
---- blockchain::electrum::generated_tests_local_electrs::test_sync_before_and_after_receive stdout ----
thread 'blockchain::electrum::generated_tests_local_electrs::test_sync_before_and_after_receive' panicked at 'called `Result::unwrap()` on an `Err` value: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })', /home/futurepaul/.cargo/registry/src/github.com-1ecc6299db9ec823/bdk-testutils-0.4.0/src/lib.rs:304:85
```

It's unclear to me where this file stuff is happening. (My hunch is that addresses are being written somewhere and this API breakage broke that, but I can't quite pin it down).

Would be happy to pair on this if you hit me up on Discord. Feel a little in over my head messing with this macro.

notmandatory: If you are testing locally these tests are probably breaking because they require the `bitcoindevkit/electrs` docker container to be running and some env variables to be set (see https://github.com/bitcoindevkit/bitcoin-regtest-box#local-bdk-testing). But I'm also happy to walk you through it. Will PM you on Discord and we'll get it working.

notmandatory:
one other wrinkle is with #347 merged we now need to rebase your branch on master and will need to fix this macro issue in a different place. But we'll figure it out.

notmandatory (approved):
ACK 0ec064e. Looks good !

src/wallet/mod.rs
```diff
@@ -196,24 +197,55 @@ pub enum AddressIndex {
...
+    pub address: Address,
+}
+
+impl Deref for AddressInfo {
+    type Target = Address;
+
+    fn deref(&self) -> &Self::Target {
+        &self.address
+    }
+}
```
thomaseizinger, line 216: Just stumbled over this while upgrading to bdk 0.8.0. Not sure if it has been mentioned in the discussion but `Deref` is really only meant for smart pointers according to this documentation: https://doc.rust-lang.org/std/ops/trait.Deref.html

Would you accept a PR that removes this `impl` block again?

---

A separate discussion is on how to make access to the address in the struct more convenient. An alternative would be to make `get_new_address` return a tuple instead which can be instantly deconstructed at the call-site without having to import and name a type (`AddressInfo` in this case):

```rust
let (address, index) = wallet.get_new_address().unwrap();
```

or this if you don't care about the index:

```rust
let (address, _) = wallet.get_new_address().unwrap();
```

It is a question of taste really. The tuple approach looses you some documentation because the returned integer is effectively unnamed and `u32` itself doesn't say much about what it is.

afilini: I'm not so sure about this, I think there are a few other cases where we use `Deref` in the code that are not smart pointers. If there aren't any real downsides to using them I would just keep this as is

RCasatta: for reference, here https://github.com/pretzelhammer/rust-blog/blob/master/posts/tour-of-rusts-standard-library-traits.md#deref--derefmut there is some explanation of why `Deref` should be avoided for things that are not smart pointers

afilini: That's a good point, but I don't think we are abusing it in this case. We are not really trying to reproduce inheritance like in other OOP langs, it's just a convenience that allows the use of methods like `script_pubkey()` directly on `AddressInfo`

PR #353, opened 2021-05-26
[signer] Add an option to explicitly allow using non-`ALL` sighashes

### LINKED ISSUE #350: Add an explicit `TxBuilder` method to enable non-ALL sighashes
BDK signs using the sighash specified in the PSBT (or `ALL` if it's not provided). It's currently up to the user of the library to check that the PSBT they are signing doesn't have a dangerous sighash, but to avoid accidents it's probably better to have this check internally and force the user to explicitly opt-in to use non-ALL sighashes.

afilini:
### Description

Instead of blindly using the `sighash_type` set in a psbt input, we now only sign `SIGHASH_ALL` inputs by default, and require the user to explicitly opt-in to using other sighashes if they desire to do so.

Fixes #350

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Wallet
+- Added an option that must be explicitly enabled to allow signing using non-`SIGHASH_ALL` sighashes (#350)
+
 ## [v0.7.0] - [v0.6.0]
 
 ### Policy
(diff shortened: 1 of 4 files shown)


LLFourn:
ACK 881ca8d. I guess what you really want is to make sure that you are not signing any of your inputs with an unusual sighash (regardless of what others do). I don't think we need to account for this though since you are likely going to have to do a lot of manual verification work anyway in more exotic scenarios.

notmandatory (approved):
utACK 881ca8d. Test is easy to follow and covers expected cases.

afilini:
> I guess what you really want is to make sure that you are not signing any of your inputs with an unusual sighash (regardless of what others do). I don't think we need to account for this though since you are likely going to have to do a lot of manual verification work anyway in more exotic scenarios.

Yeah I tried to do that initially but it was a mess. If somebody needs that kind of "precision" they can just always opt-in on the BDK side and implement the check themselves before calling `sign()`

PR #355, opened 2021-05-27
[wallet] Verify unconfirmed transactions after syncing

### LINKED ISSUE #352: Verify signatures before bumping a transaction
When bumping a transaction, one may assume that its outputs were previously confirmed by the user.
However a malicious server can tweak unconfirmed transactions so that bumping them will result in sending the funds elsewhere.

Alice broadcasts tx A, `in1 -> out1`, which won't be confirmed.
Alice restores her wallet.
Server returns tx B, `in1 -> out2`, instead of tx A.
Alice bumps tx B, but instead of sending funds to out1, funds are sent to out2.

A possible solution consists in verifying the signatures of the transaction to be replaced.
For instance, unless unusual sighashes were used, tx B will fail verification, while tx A will pass.

This attack may be relevant for wallets relying on Electrum servers.
Although in most cases (I assume) the transaction is persisted when it's sent, so this attack is not viable, IMO verifying signatures before bumping is still worth the performance penalty.

afilini:
### Description

Verify the unconfirmed transactions we download against the consensus rules. This is currently exposed as an extra `verify` feature, since it depends on a pre-release version of `bitcoinconsensus`.

Closes #352

### Notes to the reviewers

Still missing some docs items but the code should be pretty much done

#### All Submissions:

#### New Features:


src/wallet/verify.rs
```diff
@@ -0,0 +1,171 @@
...
+    blockchain: &B,
+) -> Result<(), VerifyError> {
+    log::debug!("Verifying {}", tx.txid());
+
+    let serialized_tx = serialize(tx);
+
+    for (index, input) in tx.input.iter().enumerate() {
+        let prev_tx = if let Some(prev_tx) = database.get_raw_tx(&input.previous_output.txid)? {
+            prev_tx
+        } else if let Some(prev_tx) = blockchain.get_tx(&input.previous_output.txid)? {
```
RCasatta, line 47: we may want to locally cache the tx returned here to improve performance in a situation where txs is spending more input from the same tx. Maybe too edge case...

src/wallet/verify.rs
```diff
@@ -0,0 +1,171 @@
...
+
+impl fmt::Display for VerifyError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl std::error::Error for VerifyError {}
+
+impl From<Error> for VerifyError {
```
RCasatta, line 98: We have `impl_error!` for this and the following

src/wallet/verify.rs
```diff
@@ -0,0 +1,171 @@
...
+            Ok(42)
+        }
+        fn estimate_fee(&self, _target: usize) -> Result<FeeRate, Error> {
+            Ok(FeeRate::default_min_relay_fee())
+        }
+    }
+
+    #[test]
+    fn test_verify_fail_unsigned_tx() {
+        let prev_tx: Transaction = deserialize(&Vec::<u8>::from_hex("020000000101192dea5e66d444380e106f8e53acb171703f00d43fb6b3ae88ca5644bdb7e1000000006b48304502210098328d026ce138411f957966c1cf7f7597ccbb170f5d5655ee3e9f47b18f6999022017c3526fc9147830e1340e04934476a3d1521af5b4de4e98baf49ec4c072079e01210276f847f77ec8dd66d78affd3c318a0ed26d89dab33fa143333c207402fcec352feffffff023d0ac203000000001976a9144bfbaf6afb76cc5771bc6404810d1cc041a6933988aca4b956050000000017a91494d5543c74a3ee98e0cf8e8caef5dc813a0f34b48768cb0700").unwrap()).unwrap();
```
RCasatta, line 150: if this txs are on mainet or testnet maybe add a comment with the txid so that one could look on block explorer

src/wallet/verify.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+#[derive(Debug)]
+pub enum VerifyError {
+    MissingInputTx(Txid),
+    InvalidInput(OutPoint),
+
+    Consensus(bitcoinconsensus::Error),
+
+    Global(Box<Error>),
```
RCasatta, line 87: Maybe a comment why the Box is needed since Error is an enum (because Error contains a reference to VerifyError that would create a cycle)

src/wallet/verify.rs
afilini: I don't think I can use it for `From<crate::Error>` because we have to wrap that in a `Box`, but I can use it for the other one

RCasatta (approved):

Issue #334, opened 2021-04-21
signature problems with add_foreign_utxo


ulrichard:
I am trying to implement "proof of reserves". For the challenge txin I use add_foreign_utxo.
With the current master branch, it produces a seemingly valid proof psbt. But validating the signatures fails later on.
When I merge https://github.com/afilini/bdk/tree/fix/signer-witness-utxo, then I get the following error when trying to sign: "Missing non_witness_utxo on foreign utxo"

Now I don't know if this is because of a bug in my code or in bdk.

The code is in:
https://github.com/ulrichard/bdk/tree/feature/proof_of_reserves
more specifically in: "src/wallet/reserves.rs"

LLFourn:
So I think that PR (which I am going to review soon) makes non_witness_utxo mandatory. Your code isn't providing a non_witness_utxo: https://github.com/ulrichard/bdk/blob/2d2c4a1e32d2ef411cf784738c945e6145252dab/src/wallet/reserves.rs#L84 so it errors.

I will think about this problem while reviewing that PR later today :)

LLFourn:
To make things work with this PR you just need to do:

```rust
        builder
            .drain_wallet()
            .add_foreign_utxo(challenge_txin.previous_output, challenge_psbt_inp, 42)?
            .fee_absolute(0)
            .only_witness_utxo() // <-- you're missing this.
            .set_single_recipient(out_script_unspendable);
        let (psbt, _details) = builder.finish().unwrap();
```

ulrichard:
Cool, thanks for looking into this. Makes sense.

But after adding only_witness_utxo(), I still get MissingNonWitnessUtxo when running the tests.
I will check again, after the PR is merged.

LLFourn:
To debug you might try and insert a panic inside the places that `MissingNonWitnessUtxo` is called and see why this is occurring. Waiting until that PR is sorted out is also not a bad idea.

LLFourn:
@ulrichard if you use the new API that was merged in https://github.com/bitcoindevkit/bdk/pull/333/files#diff-18f5545e0b6d6f7d1d8ef9f744c9d6d1458baaffc419e69f826626c00dddc260R859 you should be able to do this by passing to `wallet.sign(&mut psbt, SignOptions { trust_witness_utxo: true, ..Default::default() }`. I think this issue can be closed.

ulrichard:
Thanks for the heads up @LLFourn 
Unfortunately, the errors persist.
[...]

PR #356, opened 2021-05-28
Implementing proof of reserves

ulrichard:
### Description

Adding a module that allows the construction of "proof of reserves" PSBTs and the verification thereof.

### Notes to the reviewers

There will be an accompanying PR in bdk-cli. I will paste the URL here, once it's created.
https://github.com/ulrichard/bdk-cli/tree/reserves
But it builds upon this one:
https://github.com/bitcoindevkit/bdk-cli/pull/28

#### All Submissions:

#### New Features:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ lazy_static = { version = "1.4", optional = true }
 tiny-bip39 = { version = "^0.8", optional = true }
 zeroize = { version = "<1.4.0", optional = true }
 bitcoinconsensus = { version = "0.19.0-3", optional = true }
-base64 = { version = "^0.11", optional = true }
+base64 = { version = "^0.13", optional = true }
 
 # Needed by bdk_blockchain_tests macro
 bitcoincore-rpc = { version = "0.13", optional = true }
@@ -74,7 +74,7 @@ lazy_static = "1.4"
 env_logger = "0.7"
 clap = "2.33"
 electrsd = { version= "0.6", features = ["trigger", "bitcoind_0_21_1"] }
-rstest = "^0.7"
+rstest = "^0.10"
 
 [[example]]
 name = "address_validator"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,6 @@ lazy_static = { version = "1.4", optional = true }
 tiny-bip39 = { version = "^0.8", optional = true }
 zeroize = { version = "<1.4.0", optional = true }
 bitcoinconsensus = { version = "0.19.0-3", optional = true }
-bitcoin_hashes = { version = "^0.9", optional = true }
 base64 = { version = "^0.11", optional = true }
 
 # Needed by bdk_blockchain_tests macro
@@ -61,7 +60,7 @@ async-interface = ["async-trait"]
 all-keys = ["keys-bip39"]
 keys-bip39 = ["tiny-bip39", "zeroize"]
 rpc = ["bitcoincore-rpc"]
-reserves = ["bitcoinconsensus", "bitcoin_hashes", "base64"]
+reserves = ["bitcoinconsensus", "base64", "electrum"]
 
 # Debug/Test features
 test-blockchains = ["bitcoincore-rpc", "electrum-client"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,8 @@ lazy_static = { version = "1.4", optional = true }
 tiny-bip39 = { version = "^0.8", optional = true }
 zeroize = { version = "<1.4.0", optional = true }
 bitcoinconsensus = { version = "0.19.0-3", optional = true }
+bitcoin_hashes = { version = "^0.9", optional = true }
+base64 = { version = "^0.11", optional = true }
 
 # Needed by bdk_blockchain_tests macro
 bitcoincore-rpc = { version = "0.13", optional = true }
@@ -59,7 +61,7 @@ async-interface = ["async-trait"]
 all-keys = ["keys-bip39"]
 keys-bip39 = ["tiny-bip39", "zeroize"]
 rpc = ["bitcoincore-rpc"]
-
+reserves = ["bitcoinconsensus", "bitcoin_hashes", "base64"]
 
 # Debug/Test features
 test-blockchains = ["bitcoincore-rpc", "electrum-client"]
@@ -73,6 +75,7 @@ lazy_static = "1.4"
 env_logger = "0.7"
 clap = "2.33"
 electrsd = { version= "0.6", features = ["trigger", "bitcoind_0_21_1"] }
+rstest = "^0.7"
 
 [[example]]
 name = "address_validator"
--- a/src/error.rs
+++ b/src/error.rs
@@ -63,10 +63,9 @@ pub enum Error {
     },
     /// Node doesn't have data to estimate a fee rate
     FeeRateUnavailable,
-    /// The proof of reserves is invalid. The reason is given in the string parameter.
-    ProofOfReservesInvalid(String),
-    /// Cannot verify the proof.
-    CannotVerifyProof,
+    /// Proof error
+    #[cfg(feature = "reserves")]
+    Proof(crate::wallet::reserves::ProofError),
     /// In order to use the [`TxBuilder::add_global_xpubs`] option every extended
     /// key in the descriptor must either be a master key itself (having depth = 0) or have an
     /// explicit origin provided
--- a/src/error.rs
+++ b/src/error.rs
@@ -63,6 +63,10 @@ pub enum Error {
     },
     /// Node doesn't have data to estimate a fee rate
     FeeRateUnavailable,
+    /// The proof of reserves is invalid. The reason is given in the string parameter.
+    ProofOfReservesInvalid(String),
+    /// Cannot verify the proof.
+    CannotVerifyProof,
     /// In order to use the [`TxBuilder::add_global_xpubs`] option every extended
     /// key in the descriptor must either be a master key itself (having depth = 0) or have an
     /// explicit origin provided
--- a/src/wallet/reserves.rs
+++ b/src/wallet/reserves.rs
@@ -1,26 +1,13 @@
-// Magical Bitcoin Library
-// Written in 2020 by
-//     Alekos Filini <alekos.filini@gmail.com>
+// Bitcoin Dev Kit
+// Written in 2020 by Alekos Filini <alekos.filini@gmail.com>
 //
-// Copyright (c) 2020 Magical Bitcoin
+// Copyright (c) 2020-2021 Bitcoin Dev Kit Developers
 //
-// Permission is hereby granted, free of charge, to any person obtaining a copy
-// of this software and associated documentation files (the "Software"), to deal
-// in the Software without restriction, including without limitation the rights
-// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-// copies of the Software, and to permit persons to whom the Software is
-// furnished to do so, subject to the following conditions:
-//
-// The above copyright notice and this permission notice shall be included in all
-// copies or substantial portions of the Software.
-//
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-// SOFTWARE.
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
 
 //! Proof of reserves
 //!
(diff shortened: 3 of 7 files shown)


Cargo.toml
```diff
@@ -67,6 +71,8 @@ env_logger = "0.7"
 base64 = "^0.11"
 clap = "2.33"
 serial_test = "0.4"
+rstest = "^0.7"
+tiny-bip39 = "^0.7"
```
notmandatory: The clippy error seems to be caused by this line, error goes away when I remove it. Could be because version `^0.8` is defined above (line 33) ?

notmandatory:
@nicolasburtey heard you on SLP and that you are looking into proof of reserve for your project. I would be interested to have your feedback on if you think this PR is in the right direction.

Cargo.toml
ulrichard: yes, you're right. Thanks for finding that.

src/wallet/reserves.rs
```diff
@@ -0,0 +1,603 @@
...
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
```
afilini: I think this is our older header, see one of our other files for the new one

src/wallet/reserves.rs
```diff
@@ -0,0 +1,603 @@
...
+    consensus::encode::serialize,
+    hash_types::{PubkeyHash, Txid},
+    util::{
+        address::Payload,
+        psbt::{Input, PartiallySignedTransaction as PSBT},
+    },
+    Network,
+};
+use bitcoin_hashes::{hash160, sha256d, Hash};
+use bitcoinconsensus;
```
afilini: We normally use the "module" import granularity, see https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=impo#imports_granularity

src/wallet/reserves.rs
```diff
@@ -0,0 +1,603 @@
...
+    /// We can currently not validate whether it was valid at a certain block height.
+    /// Returns the spendable amount of the proof.
+    fn verify_proof(&self, psbt: &PSBT, message: &str) -> Result<u64, Error> {
+        self.do_verify_proof(psbt, message)
+    }
+
+    /// implementation
+    fn do_create_proof(&self, message: &str) -> Result<PSBT, Error>;
+    /// implementation
+    fn do_verify_proof(&self, psbt: &PSBT, message: &str) -> Result<u64, Error>;
```
afilini: I don't understand why you would need those two other methods. Can't people just implement `create_proof` and `verify_proof` directly?

src/wallet/reserves.rs
```diff
@@ -0,0 +1,603 @@
...
+        let out_script_unspendable = bitcoin::Address {
+            payload: Payload::PubkeyHash(pkh),
+            network: self.network,
+        }
+        .script_pubkey();
+
+        let mut builder = self.build_tx();
+        builder
+            .drain_wallet()
+            .add_foreign_utxo(challenge_txin.previous_output, challenge_psbt_inp, 42)?
```
afilini, line 105: If you need this to stay in a fixed position you probably need to set the ordering to `Untouched` explicitly, otherwise the inputs might get shuffled around: https://docs.rs/bdk/0.7.0/bdk/wallet/tx_builder/struct.TxBuilder.html#method.ordering

src/wallet/reserves.rs
```diff
@@ -0,0 +1,603 @@
...
+    if let Some((i, _psbt_in)) =
+        psbt.inputs.iter().enumerate().find(|(_i, psbt_in)| {
+            psbt_in.witness_utxo.is_some() && psbt_in.non_witness_utxo.is_some()
+        })
+    {
+        return Err(Error::ProofOfReservesInvalid(format!(
+            "Witness and legacy input found at input #{}",
+            i
+        )));
+    }
```
afilini: This is not really an error, the spec allows having both. Actually most signer nowadays require both for segwit inputs

src/wallet/reserves.rs
```diff
@@ -0,0 +1,603 @@
...
+
+    // Verify the SIGHASH
+    if let Some((i, _psbt_in)) = psbt.inputs.iter().enumerate().find(|(_i, psbt_in)| {
+        psbt_in.sighash_type.is_some() && psbt_in.sighash_type != Some(SigHashType::All)
+    }) {
+        return Err(Error::ProofOfReservesInvalid(format!(
+            "Unsupported sighash type at input #{}",
+            i
+        )));
+    }
```
afilini: Here you should check the sighash that actually ended up in the signature, not just the psbt metadata. Otherwise an attacker can make a signature with sighash X and then just say "no I did it with Y" in the metadata, which is not right.

src/wallet/reserves.rs
```diff
@@ -0,0 +1,603 @@
...
+        .zip(tx.input.iter())
+        .enumerate()
+        .map(|(i, (psbt_in, tx_in))| {
+            if let Some(utxo) = &psbt_in.witness_utxo {
+                (i, &utxo.script_pubkey, utxo.value)
+            } else if let Some(nwtx) = &psbt_in.non_witness_utxo {
+                let outp = &nwtx.output[tx_in.previous_output.vout as usize];
+                (i, &outp.script_pubkey, outp.value)
+            } else {
+                panic!("must be either witness or legacy");
```
afilini: Ideally you shouldn't panic here, you can just return an error

src/wallet/reserves.rs
```diff
@@ -0,0 +1,603 @@
...
+        .inputs
+        .iter()
+        .zip(tx.input.iter())
+        .enumerate()
+        .map(|(i, (psbt_in, tx_in))| {
+            if let Some(utxo) = &psbt_in.witness_utxo {
+                (i, &utxo.script_pubkey, utxo.value)
+            } else if let Some(nwtx) = &psbt_in.non_witness_utxo {
+                let outp = &nwtx.output[tx_in.previous_output.vout as usize];
+                (i, &outp.script_pubkey, outp.value)
```
afilini: I think again here you are just trusting the value from the psbt metadata. This is dangerous: https://github.com/bitcoindevkit/bdk/issues/334#issuecomment-845059471

src/wallet/reserves.rs
```diff
@@ -0,0 +1,603 @@
...
+        .iter()
+        .zip(psbt.inputs.iter())
+        .map(|(tx_in, psbt_in)| {
+            if let Some(utxo) = &psbt_in.witness_utxo {
+                utxo.value
+            } else if let Some(nwtx) = &psbt_in.non_witness_utxo {
+                nwtx.output[tx_in.previous_output.vout as usize].value
+            } else {
+                0
+            }
```
afilini: Same here, you are just trusting the `value` from the psbt input

src/wallet/reserves.rs
```diff
@@ -0,0 +1,603 @@
...
+
+/// Construct a challenge input with the message
+fn challenge_txin(message: &str) -> TxIn {
+    let message = "Proof-of-Reserves: ".to_string() + message;
+    let message = sha256d::Hash::hash(message.as_bytes());
+    TxIn {
+        previous_output: OutPoint::new(Txid::from_hash(message), 0),
+        sequence: 0xFFFFFFFF,
+        script_sig: Builder::new().into_script(),
+        witness: Vec::new(),
```
afilini: You can probably replace those two lines with:

```suggestion
        ..Default::default(),
```

src/error.rs
```diff
@@ -64,6 +64,10 @@ pub enum Error {
         /// Required fee absolute value (satoshi)
         required: u64,
     },
+    /// The proof of reserves is invalid. The reason is given in the string parameter.
+    ProofOfReservesInvalid(String),
+    /// Cannot verify the proof.
+    CannotVerifyProof,
```
afilini: I would define a specific error enum for this, like we did with `SignerError` for instance. It's just to stop this `Error` from growing too much and providing some kind of a logical hierarchy for the errors

Cargo.toml
```diff
@@ -67,6 +71,7 @@ env_logger = "0.7"
 base64 = "^0.11"
 clap = "2.33"
 serial_test = "0.4"
+rstest = "^0.7"
```
afilini: I'm not a big fan of adding extra dependencies, even if they are only used during tests since they increase compilation times. Is there a different way to achieve the same result without adding this extra dependency?

Cargo.toml
```diff
@@ -31,6 +31,9 @@ cc = { version = ">=1.0.64", optional = true }
 socks = { version = "0.3", optional = true }
 lazy_static = { version = "1.4", optional = true }
 tiny-bip39 = { version = "^0.8", optional = true }
+bitcoinconsensus = { version = "0.19.0-3", optional = true }
+bitcoin_hashes = { version = "^0.9", optional = true }
```
afilini: You don't have to add this as an extra deps, `bitcoin` re-exports it as `bitcoin::hashes` iirc

src/error.rs
afilini: Also with an isolated enum I would try to define a variant for every error instead of just having one that contains a `String`

afilini (changes requested):

src/wallet/reserves.rs
ulrichard: That comes from 20 years of C++, where this was second nature. I'm still getting used to rust.
https://stackoverflow.com/questions/6481260/non-virtual-interface-design-pattern-in-c-c

afilini: Conceptually it makes a lot of sense, but unfortunately in Rust you can prevent anybody from implementing a different version of your methods, even if you provide an implementation yourself.

This is generally used to provide a generic default implementation that people can then override with a different one if they want to

src/wallet/reserves.rs
ulrichard: I didn't find this flag in the transaction, even after reading a couple of pages Ion the topic:
* https://raghavsood.com/blog/2018/06/10/bitcoin-signature-types-sighash
* https://docs.rs/bitcoin/0.26.0/bitcoin/blockdata/script/struct.Script.html
Can you guide me to the right direction a bit here?

src/wallet/reserves.rs
ulrichard: Do I understand your comment correctly, that I should fetch the previous transaction as referenced in OutPoint to determine the value?

Cargo.toml
ulrichard: I'm sure it would be possible with some macro magic. But I don't know rust macros good enough yet.

src/wallet/reserves.rs
ulrichard: Would it make sense to add a rustfmt.toml or .rustfmt.toml file to the project?

src/wallet/reserves.rs
```diff
@@ -0,0 +1,561 @@
...
+/// We can currently not validate whether it was valid at a certain block height.
+/// Returns the spendable amount of the proof.
+pub fn verify_proof(
+    psbt: &PSBT,
+    message: &str,
+    outpoints: Vec<(OutPoint, TxOut)>,
+    network: Network,
+) -> Result<u64, Error> {
+    let tx = psbt.clone().extract_tx();
+
```
ulrichard, line 140: I'm not sure if it would make sense to copy the outpoints into a sorted container for faster lookups here.

src/wallet/reserves.rs
ulrichard: Ok, now i'm passing in TxOut alongside OutPoint and looking up from there. That way I don't have to pass in an electrum client instance, and the blockchain lookup happens outside of the verification code.

nicolasburtey:
@notmandatory super cool I'll have a look. thanks.

notmandatory:
@ulrichard I have a meta question also about this PR. It seems that you don't need to access any wallet internals to create or validate a proof of reserve. So how would you feel about making this it's own repo here in the `bitcoindevkit` project? I think this would make it easier to iterate on the PoR concepts and release on your own schedule. And you can still have `pub extern crate bdk` so anyone using your crate will have full access to bdk wallet features.

ulrichard:
> @ulrichard I have a meta question also about this PR. It seems that you don't need to access any wallet internals to create or validate a proof of reserve. So how would you feel about making this it's own repo here in the `bitcoindevkit` project? I think this would make it easier to iterate on the PoR concepts and release on your own schedule. And you can still have `pub extern crate bdk` so anyone using your crate will have full access to bdk wallet features.

I'm still fairly new to rust, and thus still learning how to structure projects. If you think that's better, it's fine with me. 

1. Do you have naming conventions for stuff like that? How about "bdk-reserves"? 
2. Can I just create the project, and you add it to the `bitcoindevkit` project, or is it better if you create it first?

Cargo.toml
afilini: Ok, we can leave this open for the meantime and once everything else is done I'll help you out with the macro

src/wallet/reserves.rs
afilini: Yeah, that makes sense. I was assuming `cargo fmt` would enforce our version by default, but that's not the case

src/wallet/reserves.rs
ulrichard: So, I wanted to verify SIGHASH type with something like:

```rust
if let Some((i, _inp)) = tx.input.iter().enumerate().skip(1).find(|(_i, inp)| {
        if let Some(sht) = inp.script_sig.as_bytes().last() {
            *sht != 0 && SigHashType::from_u32(*sht as u32) != SigHashType::All
        } else {
            true
        }
    }) {
        return Err(Error::Proof(ProofError::UnsupportedSighashType(i)));
    }
```

But the script_sig is only populated in the multisig test cases. In the single-sig cases, 
`psbt.clone().extract_tx().input[1].script_sig.as_bytes().len()` 
is always 0

What could be the reason for that, and how should I cope with that?

src/wallet/reserves.rs
ulrichard: Upon closer inspection, it might be too early. It's still an unstable feature in rustfmt.

src/wallet/reserves.rs
ulrichard: Can I also interpret the last byte of the `witness: Vec<Vec<u8>>` as SigHashType?

ulrichard: With the following function, most unit tests pass. But the code doesn't look correct to me.

```rust
...
let tx = psbt.clone().extract_tx();
// Verify the SIGHASH
if let Some((i, _inp)) = tx
        .input
        .iter()
        .enumerate()
        .skip(1)
        .find(|(_i, inp)| !verify_sighash_type_all(inp))
{
     return Err(Error::Proof(ProofError::UnsupportedSighashType(i)));
}
...
```

```rust
fn verify_sighash_type_all(inp: &TxIn) -> bool {
    if inp.witness.is_empty() {
        if let Some(sht) = inp.script_sig.as_bytes().last() {
            if SigHashType::from_u32(*sht as u32) == SigHashType::All {
                return true;
            } else {
                return false;
            }
        } else {
            return false;
        }
    } else {
        for wit in &inp.witness {
            if wit.last().is_none() {
                continue;
            }
            let sht = SigHashType::from_u32(*wit.last().unwrap() as u32);
            if ![
                SigHashType::All,
                SigHashType::Single,
                SigHashType::AllPlusAnyoneCanPay,
                SigHashType::SinglePlusAnyoneCanPay,
            ]
            .contains(&sht)
            {
                return false;
            }
        }
        return true;
    }
}
```
But still the legacy multisig case fails. It has an `sht` value of 0xAE which doesn't really translate to a valid SigHashType.

notmandatory:
> I'm still fairly new to rust, and thus still learning how to structure projects. If you think that's better, it's fine with me.

Thanks, we're still figuring it out too, but I think in this case since proof of reserve is still an experimental topic it makes sense for it to live in it's own repo. 

>     1. Do you have naming conventions for stuff like that? How about "bdk-reserves"?

We don't have an official convention, the most important thing is that the name isn't taken in https://crates.io yet. The name `bdk-reserves` is a available. 

>     2. Can I just create the project, and you add it to the `bitcoindevkit` project, or is it better if you create it first?

It's probably easier if you create the repo in your account and then when you have it ready to re-review you can transfer it to the `bitcoindevkit` project. I don't think your code will need to change much, you should still be able to implementing the `ProofOfReserves` trait on `Wallet`.  And I think this could be a good dev model for how other experimental features (mixing, DLCs, stuff like that), could be built on `bdk`.

rajarshimaitra:
Hello @ulrichard , I haven't looked into the PR in detail yet, but judging from the overall objective is it something similar to [BIP322](https://github.com/bitcoin/bips/blob/master/bip-0322.mediawiki)? That BIP is a generic message signing protocol using specially formed Bitcoin transactions. And one of the major applications of such generic signing is "Proof of Reserve", as outlined [here](https://github.com/bitcoin/bips/blob/master/bip-0322.mediawiki#full-proof-of-funds). 

My understanding is BIP322 can be used for many things including proof of reserve, so maybe you might wanna try to include that standard in your implementation?

Unfortunately, there is no working implementation of BIP322 and `rust-miniscript` had a feature request open for it for some time now.  As far as my understanding, nobody in Blockstream has time to write it, but they would appreciate PRs. 

I have worked on a very rough sketch of it a few months ago trying to layout how basic BIP322 rust structures might look like. https://github.com/rust-bitcoin/rust-miniscript/pull/240

It's not complete but it does basic BIP322 functions of signing and verifying. That might be something of your interest? 

I am also happy to put in hands with you finishing BIP322 (it has been stale for some time now) and BDK can simply use that (it already has `rust-miniscript` dependency), if that aligns with your need.

Overall I feel this magic should reside in some upstream rust library like `rust-miniscript` or `rust-bitcoin` instead of wallet tools. Wallet tools like BDK can then simply expose simple APIs to use it.

ulrichard:
> Hello @ulrichard , I haven't looked into the PR in detail yet, but judging from the overall objective is it something similar to [BIP322](https://github.com/bitcoin/bips/blob/master/bip-0322.mediawiki)? That BIP is a generic message signing protocol using specially formed Bitcoin transactions. And one of the major applications of such generic signing is "Proof of Reserve", as outlined [here](https://github.com/bitcoin/bips/blob/master/bip-0322.mediawiki#full-proof-of-funds).
> 
> My understanding is BIP322 can be used for many things including proof of reserve, so maybe you might wanna try to include that standard in your implementation?
> 
> Unfortunately, there is no working implementation of BIP322 and `rust-miniscript` had a feature request open for it for some time now. As far as my understanding, nobody in Blockstream has time to write it, but they would appreciate PRs.
> 
> I have worked on a very rough sketch of it a few months ago trying to layout how basic BIP322 rust structures might look like. [rust-bitcoin/rust-miniscript#240](https://github.com/rust-bitcoin/rust-miniscript/pull/240)
> 
> It's not complete but it does basic BIP322 functions of signing and verifying. That might be something of your interest?
> 
> I am also happy to put in hands with you finishing BIP322 (it has been stale for some time now) and BDK can simply use that (it already has `rust-miniscript` dependency), if that aligns with your need.
> 
> Overall I feel this magic should reside in some upstream rust library like `rust-miniscript` or `rust-bitcoin` instead of wallet tools. Wallet tools like BDK can then simply expose simple APIs to use it.

Hi @rajarshimaitra, Yes BIP322 along with BIP127 were the main inspirations. Of which I followed BIP127 more closely. 
I didn't realize BIP322 could be implemented in miniscript. I'll have a look at your PR.

src/wallet/reserves.rs
ulrichard: I added some asserts temporarily to the unit tests to visualize the test data. 
They show a couple of things that I didn't expect from the information I found on the internet.
For the moment I adapted the verify_sighash_type_all() to let them pass, and added some comments.
I am curious to get some explanations on these values.

notmandatory:
Closing the PR and will work with @ulrichard on his https://github.com/ulrichard/bdk-reserves repo instead.

PR #358, opened 2021-06-01
Add a ureq version of esplora module

tcharding:
### Description

The `Blockchain` implementation for connecting to an Esplora instance is currently based on `reqwest`. Some users may not wish to use reqwest.

`ureq` is a simple HTTP client (no async) that is useful when `reqwest` is not suitable.

- Move `esplora.rs` -> `esplora/reqwest.rs`
- Add an implementation based on the `reqwest` esplora code but using `ureq`
- Add feature flags and conditional includes to re-export everything to the `esplora` module so we don't effect the rest of the code base.

This allows us to also remove the forced dependency on `tokio`.

Add comments to the manifest in an attempt to make it clear which features are needed under which circumstances.

### CI changes

This PR adds feature builds for esplora/ureq, async-interface, and in the WASM section adds reqwest as required. Please review carefully.

#### All Submissions:

#### New Features:

### DIFF
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Update toolchain
         run: rustup update
       - name: Test
-        run: cargo test --features all-keys,compiler,esplora,compact_filters --no-default-features
+        run: cargo test --features all-keys,compiler,esplora,ureq,compact_filters --no-default-features
                     
       - id: coverage
         name: Generate coverage
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ readme = "README.md"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-bdk-macros = "^0.4"
+bdk-macros = { path = "macros"} # TODO: Change this to version number after next release.
 log = "^0.4"
 miniscript = "5.1"
 bitcoin = { version = "~0.26.2", features = ["use-serde", "base64"] }
@@ -24,6 +24,7 @@ rand = "^0.7"
 sled = { version = "0.34", optional = true }
 electrum-client = { version = "0.7", optional = true }
 reqwest = { version = "0.11", optional = true, features = ["json"] }
+ureq = { version = "2.1", default-features = false, features = ["json"], optional = true }
 futures = { version = "0.3", optional = true }
 async-trait = { version = "0.1", optional = true }
 rocksdb = { version = "0.14", optional = true }
@@ -37,10 +38,6 @@ bitcoinconsensus = { version = "0.19.0-3", optional = true }
 # Needed by bdk_blockchain_tests macro
 bitcoincore-rpc = { version = "0.13", optional = true }
 
-# Platform-specific dependencies
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-tokio = { version = "1", features = ["rt"] }
-
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 async-trait = "0.1"
 js-sys = "0.3"
@@ -51,21 +48,32 @@ minimal = []
 compiler = ["miniscript/compiler"]
 verify = ["bitcoinconsensus"]
 default = ["key-value-db", "electrum"]
-electrum = ["electrum-client"]
-esplora = ["reqwest", "futures"]
 compact_filters = ["rocksdb", "socks", "lazy_static", "cc"]
 key-value-db = ["sled"]
-async-interface = ["async-trait"]
 all-keys = ["keys-bip39"]
 keys-bip39 = ["tiny-bip39", "zeroize"]
 rpc = ["bitcoincore-rpc"]
+
+# We currently provide mulitple implementations of `Blockchain`, all are
+# blocking except for the `EsploraBlockchain` which can be either async or
+# blocking, depending on the HTTP client in use.
+#
+# - Users wanting asynchronous HTTP calls should enable `async-interface` to get
+#   access to the asynchronous method implementations. Then, if Esplora is wanted,
+#   enable `esplora` AND `reqwest`.
+# - Users wanting blocking HTTP calls can use any of the other blockchain
+#   implementations (`compact_filters`, `electrum`, or `esplora`). Users wanting to
+#   use Esplora should enable `esplora` AND `ureq`.
+async-interface = ["async-trait"]
+electrum = ["electrum-client"]
+esplora = ["futures"]           # Requires one of: `ureq` or `reqwest`.
 
 
 # Debug/Test features
 test-blockchains = ["bitcoincore-rpc", "electrum-client"]
 test-electrum = ["electrum", "electrsd/electrs_0_8_10", "test-blockchains"]
 test-rpc = ["rpc", "electrsd/electrs_0_8_10", "test-blockchains"]
-test-esplora = ["esplora", "electrsd/legacy", "electrsd/esplora_a33e97e1", "test-blockchains"]
+test-esplora = ["esplora", "ureq", "electrsd/legacy", "electrsd/esplora_a33e97e1", "test-blockchains"]
 test-md-docs = ["electrum"]
 
 [dev-dependencies]
@@ -88,6 +96,6 @@ required-features = ["compiler"]
 [workspace]
 members = ["macros"]
 [package.metadata.docs.rs]
-features = ["compiler", "electrum", "esplora", "compact_filters", "rpc", "key-value-db", "all-keys", "verify"]
+features = ["compiler", "electrum", "esplora", "ureq", "compact_filters", "rpc", "key-value-db", "all-keys", "verify"]
 # defines the configuration attribute `docsrs`
 rustdoc-args = ["--cfg", "docsrs"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,13 +60,20 @@ rpc = ["bitcoincore-rpc"]
 #
 # - Users wanting asynchronous HTTP calls should enable `async-interface` to get
 #   access to the asynchronous method implementations. Then, if Esplora is wanted,
-#   enable `esplora` AND `reqwest`.
+#   enable `esplora` AND `reqwest` (`--features=use-esplora-reqwest`).
 # - Users wanting blocking HTTP calls can use any of the other blockchain
 #   implementations (`compact_filters`, `electrum`, or `esplora`). Users wanting to
-#   use Esplora should enable `esplora` AND `ureq`.
+#   use Esplora should enable `esplora` AND `ureq`  (`--features=use-esplora-ureq`).
+#
+# WARNING: Please take care with the features below, various combinations will
+# fail to build. We cannot currently build `bdk` with `--all-features`.
 async-interface = ["async-trait"]
 electrum = ["electrum-client"]
-esplora = ["futures"]           # Requires one of: `ureq` or `reqwest`.
+# MUST ALSO USE `--no-default-features`.
+use-esplora-reqwest = ["async-interface", "esplora", "reqwest", "futures"]
+use-esplora-ureq = ["esplora", "ureq"]
+# Typical configurations will not need to use `esplora` feature directly.
+esplora = []
 
 
 # Debug/Test features
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -121,26 +121,3 @@ pub fn maybe_await(expr: TokenStream) -> TokenStream {
 
     quoted.into()
 }
-
-/// Awaits if target_arch is "wasm32", uses `tokio::Runtime::block_on()` otherwise
-///
-/// Requires the `tokio` crate as a dependecy with `rt-core` or `rt-threaded` to build on non-wasm32 platforms.
-#[proc_macro]
-pub fn await_or_block(expr: TokenStream) -> TokenStream {
-    let expr: proc_macro2::TokenStream = expr.into();
-    let quoted = quote! {
-        {
-            #[cfg(all(not(target_arch = "wasm32"), not(feature = "async-interface")))]
-            {
-                tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap().block_on(#expr)
-            }
-
-            #[cfg(any(target_arch = "wasm32", feature = "async-interface"))]
-            {
-                #expr.await
-            }
-        }
-    };
-
-    quoted.into()
-}
--- a/src/blockchain/esplora/mod.rs
+++ b/src/blockchain/esplora/mod.rs
@@ -17,6 +17,7 @@
 //! Please note, to configure the Esplora HTTP client correctly use one of:
 //! Blocking:  --features='esplora,ureq'
 //! Async:     --features='async-interface,esplora,reqwest' --no-default-features
+use std::collections::HashMap;
 use std::fmt;
 use std::io;
 
@@ -24,6 +25,9 @@ use serde::Deserialize;
 
 use bitcoin::consensus;
 use bitcoin::{BlockHash, Txid};
+
+use crate::error::Error;
+use crate::FeeRate;
 
 #[cfg(all(
     feature = "esplora",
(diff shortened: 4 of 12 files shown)


notmandatory:
Supporting [SGX](https://fortanix.com/blog/2020/03/intel-sgx-technology-and-the-impact-of-processor-side-channel-attacks/)  sounds like a great feature for bitcoin projects like bdk. Any idea if `ureq` works with WASM? if it does I wonder if there's any reason to not just replace `reqwest` with `ureq`.

tcharding:
> Supporting [SGX](https://fortanix.com/blog/2020/03/intel-sgx-technology-and-the-impact-of-processor-side-channel-attacks/) sounds like a great feature for bitcoin projects like bdk. Any idea if `ureq` works with WASM? if it does I wonder if there's any reason to not just replace `reqwest` with `ureq`.

I can look into it and report back. Thanks

tcharding:
Some more information related to this PR, just for interests sake (cc @LLFourn)

The reason we cannot use `reqwest` is because it depends on `hyper` which depends on `socket2` which uses the `unix` target ([soure code](https://github.com/rust-lang/socket2/blob/f7023b4c810eb7b0fedf23a1752461c41765c797/src/lib.rs#L122)). Socket2 seems to handle a whole bunch of targets so its reasonable to think SGX support could be patched in, I don't personally plan on doing so though :)

LLFourn:
@tcharding Thanks. So what is the target you use for SGX? I guess what is needed for it to work is for the SGX environment to support non-blocking IO. It would be interesting to read more about this.

I think a way to make this PR more useful would be to enable the `ureq` implementation whenever the `async-interface` feature is disabled. Although this sounds messy (having two implementations) it might actually be much less messy than what we do now which is force the user to depend on tokio just to do `block_on`. @afilini  thoughts? There is of course quite some more work to do if we decide to go this route with this PR but @tcharding sounds keen.

LLFourn:
@notmandatory quick note: WASM in the browser needs a backend based on [web_sys](https://docs.rs/web-sys/0.3.51/web_sys/) like reqwest. `ureq` does not support this.

tcharding:
> @tcharding Thanks. So what is the target you use for SGX? I guess what is needed for it to work is for the SGX environment to support non-blocking IO. It would be interesting to read more about this.

We use this in `.cargo/config`:
```
[target.x86_64-fortanix-unknown-sgx]
runner = "ftxsgx-runner-cargo"

[build]
target = "x86_64-fortanix-unknown-sgx"
```
This builds using the target `sgx` (i.e., `#[cfg(target = sgx)]`), there are multiple sgx implementations so we can get more fine grained control using `vendor`. I've not used it before but its something like `vendor = fortanix`.

> I think a way to make this PR more useful would be to enable the `ureq` implementation whenever the `async-interface` feature is disabled. Although this sounds messy (having two implementations) it might actually be much less messy than what we do now which is force the user to depend on tokio just to do `block_on`. @afilini thoughts? There is of course quite some more work to do if we decide to go this route with this PR but @tcharding sounds keen.

Sweet, I'm keen to hack on this if its useful for the project.

afilini:
> I think a way to make this PR more useful would be to enable the ureq implementation whenever the async-interface feature is disabled. Although this sounds messy (having two implementations) it might actually be much less messy than what we do now which is force the user to depend on tokio just to do block_on.

Yup, this makes a lot of sense. I've been thinking about doing something similar for a while, at least since Matt suggested on Discord that we write our own http client. I'm not sure if I wanna go down that route (writing our client) right now, but at least for the non-wasm/non-async part we could use a much smaller client which is also easier to audit.

So yeah, I think it makes sense to have this kind of separation, and in the future if we really want to minimize our deps we can reimplement the http sync client ourselves using only the normal I/O syscalls (which I guess is the reason why reqwest doesn't work on SGX).

tcharding:
One thing I'm not all that comfortable with is that features are not additive, although that is the case before this PR. FTR I was building with `cargo check --no-default-features --features compiler,esplora,ureq,compact_filters,all-keys,keys-bip39,key-value-db && cargo check --no-default-features --features async-interface,reqwest,esplora`

Is there a better way to get full coverage when building?

LLFourn:
@tcharding see comment here: https://github.com/bitcoindevkit/bdk/pull/364#issuecomment-857300148

tcharding:
After much confusion I've realised that the request above from @LLFourn to provide `ureq` if `async-interface` is enabled is already in this PR (may or may not have been there when the comment was written, I can't remember when I did it :)

notmandatory:
Looks like this one needs a rebase to pickup the CI changes to test against rust `1.53.0`. And there may be a few more places to `cfg!` to get the WASM check to pass.

tcharding:
> Looks like this one needs a rebase to pickup the CI changes to test against rust `1.53.0`.

Cool, thanks.

> And there may be a few more places to `cfg!` to get the WASM check to pass.

This is failing because I did not modify the github actions for CI, this is done now. I've updated the initial PR comment to describe the CI changes.

LLFourn:
note @tcharding I am looking forward to reviewing this but I think it's best to wait until tests pass? Just checking to see if that makes sense to you.

tcharding:
> note @tcharding I am looking forward to reviewing this but I think it's best to wait until tests pass? Just checking to see if that makes sense to you.

~Yeah mate, no worries. I'm getting easily put of by CI in `bdk` lately project because it seems that I  have not yet worked out how to whisper the correct incantation to get my PRs past it :) I'll try to be more proactive.~

EDIT: Stop being a whinging bit** Tobin.

FTR and for anyone else reading this, I never expect any reviews until I have all green on CI.

notmandatory:
Hey there, have some good news / bad news for you on the CI front, I merged @RCasatta's new parallel integration test PR #375.  So you'll need to rebase again to pick it up, but it should all run much faster now!

tcharding:
> Hey there, have some good news / bad news for you on the CI front, I merged @RCasatta's new parallel integration test PR #375. So you'll need to rebase again to pick it up, but it should all run much faster now!

Great, thanks man.

macros/src/lib.rs
```diff
@@ -121,26 +121,3 @@ pub fn maybe_await(expr: TokenStream) -> TokenStream {
 
     quoted.into()
 }
-
```
afilini, line 124: Since you've changed the `bdk-macros` you should also update the `Cargo.toml` accordingly and use the local version. Otherwise when it's time to release this changes we will probably forget about them

src/blockchain/esplora/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+//! let blockchain = EsploraBlockchain::new("https://blockstream.info/testnet/api");
+//! # Ok::<(), bdk::Error>(())
+//! ```
+//!
+//! Esplora blockchain can use either `ureq` or `reqwest` for the HTTP client
+//! depending on your needs (blocking or async respectively).
+//!
+//! Please note, to configure the Esplora HTTP client correctly use one of:
+//! Blocking:  --features esplora,ureq
+//! Async:     --features async-features
```
afilini: This seems to imply that `async-features` is enough to enable the esplora async module, but I think you actually need the explicit `esplora` feature as well.

src/blockchain/esplora/ureq.rs
```diff
@@ -0,0 +1,437 @@
...
+    /// Invalid Hex data returned
+    Hex(bitcoin::hashes::hex::Error),
+
+    /// Transaction not found
+    TransactionNotFound(Txid),
+    /// Header height not found
+    HeaderHeightNotFound(u32),
+    /// Header hash not found
+    HeaderHashNotFound(BlockHash),
+}
```
afilini: Unless it's too complicated for some reason, I would move those structs/enums in the `mod.rs` file, so that we avoid duplicating them

src/blockchain/esplora/mod.rs
tcharding: Well two things. Most importantly this comment uses the wrong feature flag, its `async-interface` not `async-features`, so thanks for flagging this line. With `--features async-interface`, if I understand what I've done correctly, there is no need for explicit `esplora`. This was intentional, since there is only one async backend supported I figured we could enable it with
```
async-interface = ["async-trait", "esplora", "reqwest"]
```
I've pushed another commit attempting to clarify this further in `Cargo.toml`.

Are you happy with this design?

tcharding: 'Comfortable' is probably a better word that 'happy' - it still sucks that we cannot use `--all-features` :)

src/blockchain/esplora/ureq.rs
tcharding: Good thinking, can do.

tcharding:
Thanks for the review @afilini, will improve as suggested.

afilini:
Sorry, I forgot to point this out during the last review but you should also add the `ureq` feature to the list of features that docs.rs builds when producing the docs, otherwise it'll fail to build at the next release.

It's in the `[package.metadata.docs.rs]` section at the very end of `Cargo.toml`

src/blockchain/esplora/mod.rs
afilini: I'm not so sure about this, even though I don't particularly like the async stuff today I'm not completely excluding that we can implement more async `Blockchain` in the future.

I'd rather have two separate features, one to enable the async interface (which means that the trait will have async methods) and a different one to enable esplora.

This would also allow external users to implement their own custom async `Blockchain` without having to pull in reqwest and all the esplora code.

tcharding: Oh, good point. I did not think about users extending `bdk` - interesting things to think about when hacking on this repo that I'm not used to thinking about. Will see what I can come up with, thanks.

tcharding: Made `esplora`/`reqwest` independent of `async-interface`. Update docs in manifest to further assist in understanding whats required when using these features.

afilini (approved):

afilini:
Code looks really good, could you rebase on `master` to fix the conflicts?

I think we can merge it once that's done.

tcharding:
> Code looks really good, could you rebase on `master` to fix the conflicts?

Sure thing, I'll try get to it tomorrow.

> I think we can merge it once that's done.

Awesome!

tcharding:
With the addition of `stop_gap` to esplora (commit: `7e986fd Add stop_gap param to electrum and esplora blockchain configs`) this PR had some non-trivial merge conflicts. Seeing that the review process was complete but now the code will need to be re-reviewed I have squashed all commits down into a single patch.

tcharding:
Have I done something wrong here, CI has been running all weekend?

LLFourn:
Looks like they all passed here: https://github.com/bitcoindevkit/bdk/pull/358/checks?check_run_id=3139220301

tcharding:
> Looks like they all passed here: https://github.com/bitcoindevkit/bdk/pull/358/checks?check_run_id=3139220301

Thanks for looking. Why the yellow dots then?

afilini:
Oh, I think that's because you've changed a couple of jobs for the CI that were marked as "required" to allow merging the PR. I can remove them temporarily before merging and re-adding the new ones afterwards.

afilini:
I've noticed that there's one comment that hasn't been addressed. I couldn't place it on the relevant line because it's not part of the diff:

> Since you've changed the bdk-macros you should also update the Cargo.toml accordingly and use the local version. Otherwise when it's time to release this changes we will probably forget about them

Everything else looks really good, thanks!

tcharding:
> I've noticed that there's one comment that hasn't been addressed. I couldn't place it on the relevant line because it's not part of the diff:

My mistake, thanks for your patience reviewing this PR!

Added an additional patch with the toml change. The code builds cleanly while depending on released version of `macros` so I left this as a separate patch instead of squashing.

.github/workflows/code_coverage.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Update toolchain
         run: rustup update
       - name: Test
-        run: cargo test --features all-keys,compiler,esplora,compact_filters --no-default-features
+        run: cargo test --features all-keys,compiler,esplora,ureq,compact_filters --no-default-features
```
LLFourn, line 27: ~why does this flag need to be here shouldn't it just be ureq if esplora is flagged but `async-interface` is not?~

Cargo.toml
```diff
@@ -51,21 +48,32 @@ minimal = []
...
+#
+# - Users wanting asynchronous HTTP calls should enable `async-interface` to get
+#   access to the asynchronous method implementations. Then, if Esplora is wanted,
+#   enable `esplora` AND `reqwest`.
+# - Users wanting blocking HTTP calls can use any of the other blockchain
+#   implementations (`compact_filters`, `electrum`, or `esplora`). Users wanting to
+#   use Esplora should enable `esplora` AND `ureq`.
+async-interface = ["async-trait"]
+electrum = ["electrum-client"]
+esplora = ["futures"]           # Requires one of: `ureq` or `reqwest`.
```
LLFourn: makes more sense that `futures` is turned on by `async-interface` rather than `esplora` now.

src/blockchain/esplora/ureq.rs
```diff
@@ -0,0 +1,391 @@
...
+
+impl std::convert::From<UrlClient> for EsploraBlockchain {
+    fn from(url_client: UrlClient) -> Self {
+        EsploraBlockchain {
+            url_client,
+            stop_gap: 20,
+        }
+    }
+}
```
LLFourn, line 59: side note: I know we have this already in the existing one so we can ignore it but I find it questionable to insert this arbitrary undocumented choice of 20 here. I don't think we need this `From` implementation at all.

src/blockchain/esplora/ureq.rs
```diff
@@ -0,0 +1,391 @@
...
+        let fee_val = estimates
+            .into_iter()
+            .map(|(k, v)| Ok::<_, std::num::ParseIntError>((k.parse::<usize>()?, v)))
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(|e| Error::Generic(e.to_string()))?
+            .into_iter()
+            .take_while(|(k, _)| k <= &target)
+            .map(|(_, v)| v)
+            .last()
+            .unwrap_or(1.0);
```
LLFourn: I'd say this logic is complicated enough that we don't want it in two places.
The logic is also wrong. I ran into the bug yesterday -- the vec must be sorted first. I am not concerned whether this happens in this PR or another one :)

LLFourn (approved):
Nice work. LGTM. Would like to remove futures dependency to when we are actually using it.

.github/workflows/code_coverage.yml
LLFourn: (oh wait now I get it).

Cargo.toml
tcharding: oh yes, you are right in that this doesn't make sense. How about if `futures` is turned on by `reqwest` however because that's the only thing that uses them. (Commented without coding.)

tcharding:
Thanks for the review @LLFourn, I've read you comments and will action them shortly :)

afilini:
Sorry, I accidentally broke the CI in master with an older PR, if you could rebase once again after you are done addressing the comments that would be great!

tcharding:
> er with an older PR, if you could rebase once again after you are done addressing the comments that would be great!

No sweat, I'm glad actually, I was pondering whether to squash or not this morning, now I have an excuse :)

Cargo.toml
tcharding: I've had another think about how the features are combined and made an effort to make usage more ergonomic. Its in a separate patch to ease review commit: `c01f515 Improve feature combinations for ureq/reqwest`. From the commit log of that patch:
```
    ...

    With this patch applied we can do

    - `cargo check --no-default-features --features=use-esplora-reqwest`
    - `cargo check --no-default-features --features=use-esplora-ureq`
    - `cargo check --features=use-esplora-ureq`
    - `cargo check --no-default-features --features=async-trait`
```

src/blockchain/esplora/ureq.rs
tcharding: I tend to agree, but since its not part of this PR and since this one has been dragging on for a while now I'll leave this for another day.

Adding it to my TODO list.

src/blockchain/esplora/ureq.rs
tcharding: Moved logic to a separate function `esplora::into_fee_rate()`. Open to a better name.

Cargo.toml
tcharding: Includes update to CI

LLFourn: Good idea. This looks a better to me. I'd prefer to remove the `use-` from the front of these though.

PR #364, opened 2021-06-08
Add maybe_async attribute to CompactFiltersBlockchain

tcharding:
### Description

Currently we cannot build `bdk` using the following build command:

```
cargo check --no-default-features \
--features  async-interface,esplora,compact_filters
```

This is because the `maybe_async` attribute is missing from the `Blockchain` impl block for `CompactFiltersBlockchain`.

Add maybe_async attribute to CompactFiltersBlockchain `Blockchain` impl block.

#### All Submissions:

### DIFF
--- a/src/blockchain/compact_filters/mod.rs
+++ b/src/blockchain/compact_filters/mod.rs
@@ -221,6 +221,7 @@ impl CompactFiltersBlockchain {
     }
 }
 
+#[maybe_async]
 impl Blockchain for CompactFiltersBlockchain {
     fn get_capabilities(&self) -> HashSet<Capability> {
         vec![Capability::FullHistory].into_iter().collect()


tcharding:
Back to draft until I get CI passing :)

LLFourn:
NACK: The code you are making async is doing a lot of blocking io inside defeating the point. To make this async we'd have to do a lot of work. I think only the esplora backend is engineered to be async which means you can't compile `async-interface` with anything but esplora.

This sucks but I'm not sure what the solution is. One possible solution is to make everything async and get rid of the all the macros and switching. You can use `tokio::task::spawn_blocking` to spawn a thread to do it without re-egineering the internals. For now you should only use `async-interface` with esplora and probably only when compiling for wasm.

tcharding:
> NACK: The code you are making async is doing a lot of blocking io inside defeating the point. To make this async we'd have to do a lot of work. 

Woops, guess I should look at what I'm doing and not just build and YOLO

I think only the esplora backend is engineered to be async which means you can't compile `async-interface` with anything but esplora.

If only esplora is async and we are considering making esplora not async unless in WASM (using ureq) then what is the use of the featuer `async-interface`?

> This sucks but I'm not sure what the solution is. One possible solution is to make everything async and get rid of the all the macros and switching. You can use `tokio::task::spawn_blocking` to spawn a thread to do it without re-egineering the internals. 

Did all this come about when support for WASM was added? We can't be the only project facing this if that is the case?

> For now you should only use `async-interface` with esplora and probably only when compiling for wasm.

This should somehow be explicit in how we define the features :) 

Thanks for the review, will close and push this forward to another day.

LLFourn:
> 
> 
> > NACK: The code you are making async is doing a lot of blocking io inside defeating the point. To make this async we'd have to do a lot of work.
> 
> Woops, guess I should look at what I'm doing and not just build and YOLO
> 
> I think only the esplora backend is engineered to be async which means you can't compile `async-interface` with anything but esplora.
> 
> If only esplora is async and we are considering making esplora not async unless in WASM (using ureq) then what is the use of the featuer `async-interface`?

I guess just to let you do it outside of wasm if you want to. It needs to be better but I'm not sure which direction to go in.

afilini:
Yeah, the idea is just to allow the use of the async interface on non-wasm platforms when it's available (currently only esplora). I wasn't really expecting anybody to use it, but I thought it could be useful for testing.

Considering that pretty much all of our i/o code is non-async (database read/writes for instance) I don't think it really makes sense to invest time in making our backends async as well, other than when it's absolutely required, like with esplora for wasm. We would probably not gain anything in terms of performance and we would have to deal with runtimes and everything which complicates things quite a bit.

PR #365, opened 2021-06-09
Fix the `external_doc` CI error, make the examples `no_run`

afilini:
### Description

This PR fixes the issue we've been having with our CI on nightly runs, since the `feature(external_doc)` has been removed.

This also finishes the work started in #340 to make sure the CI compiles the examples we have in `src/lib.rs`.

### Notes to the reviewers

This needs to go in before we can make the `v0.8.0` release, because otherwise we can't use the CI to test our changes before publishing to crates.io like we do normally.

#### All Submissions:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,9 +14,6 @@
 // only enables the `doc_cfg` feature when
 // the `docsrs` configuration attribute is defined
 #![cfg_attr(docsrs, feature(doc_cfg))]
-// only enables the nightly `external_doc` feature when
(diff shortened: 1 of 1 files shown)


RCasatta (approved):
LGTM

just a note, I've seen crates using a generic `unstable` feature for nightly-only features that may be more generic than `test-md-docs` if we have other situations where it's useful to avoid features proliferations

afilini:
Yeah we can consider that in the future if we start adding more and more nightly-only features. I think right now this is the only one we have

PR #368, opened 2021-06-11
Change blockchain tests to use bitcoind cookie authentication

notmandatory:
### Description

Update CI to use new container images and use bitcoind cookie based authentication.

### Notes to the reviewers

I switched to using a versioned tag of the docker images so that we can push new `latest` versions of these images in the future without worrying about breaking the CI pipeline for this project.

#### All Submissions:

### DIFF
--- a/run_blockchain_tests.sh
+++ b/run_blockchain_tests.sh
@@ -14,10 +14,7 @@ eprintln(){
 }
 
 cleanup() {
-    if test "$id"; then
-        eprintln "cleaning up $blockchain docker container $id";
-        docker rm -fv "$id" > /dev/null;
-    fi
+    docker kill test_container
     trap - EXIT INT
 }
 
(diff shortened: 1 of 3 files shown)


run_blockchain_tests.sh
```diff
@@ -48,21 +45,23 @@ case "$blockchain" in
...
-export BDK_RPC_PASS=passw
+export BDK_RPC_AUTH=COOKIEFILE
+export BDK_RPC_COOKIEFILE=/tmp/regtest.cookie
 export BDK_RPC_URL=127.0.0.1:18443
 export BDK_RPC_WALLET=bdk-test
 export BDK_ELECTRUM_URL=tcp://127.0.0.1:60401
 
 cli(){
-    docker exec -it "$id" /root/bitcoin-cli -regtest -rpcuser=admin -rpcpassword=passw $@
+    docker exec -it test_container /root/bitcoin-cli -regtest -datadir=/root/.bitcoin $@
```
LLFourn: I'd prefer to keep this as `$id` and not use container names so different runs don't conflict.

LLFourn (approved):
ACK

run_blockchain_tests.sh
notmandatory: As long as the container is killed there won't be any data kept from a prior run, that's why I added the `docker --rm` option. I also tested with consecutive starting, generating blocks, killing the `test_container` and confirmed that the data is always reset back to a fresh state.

afilini: I think it still makes sense to have unique ids so that tests can be run in parallel maybe? Like multiple blockchains at the same time

LLFourn: sorry yes this is what I meant.

notmandatory: The problem we will have with parallel containers is we'll also need unique ports assigned to each container. Which may be hard to do now but will be easier with test frameworks like [electrsd](https://github.com/RCasatta/electrsd) and [bitcoind](https://github.com/RCasatta/bitcoind) that @RCasatta is working on. 

So I see your point, I'll switch this PR back to use `$id`. :slightly_smiling_face:

notmandatory: Now that I think about it, if we switch to test frameworks that configure and start the daemons that each test needs then we may not need docker containers at all.. anyway I'll still change it back.

afilini: I think it's still gonna be useful to have docker containers to use them as an environment with everything pre-installed. Otherwise every time you have to download core, compile electrs, etc.

I think eventually we will probably manage to remove all of our bash scripts and just keep using the containers to ship binaries

afilini (approved):

PR #370, opened 2021-06-12
Make fee and timestamp in TransactionDetails Options

### LINKED ISSUE #369: Transaction timestamp is 0 if transaction is seen by wallet before being mined
I'm not sure if this issue should be opened here but I noticed the problem downstream on both on `bdk-cli` and `bdk-jni`.

If you sync the wallet while an _incoming_ transaction you did not initiate (not a self-transfer) is in the mempool, the timestamp given to the transaction in the database will be 0. Note that this behaviour (putting a timestamp of 0 if transaction is seen but still in mempool) is not consistent across transactions, because if the transaction is a send (coming from the wallet) it will add the timestamp as time broadcast, whereas if it's a receive, it will put the timestamp at 0. 

An example of the problems this causes is that one would have to keep track of time of transactions separately instead of using the bdk timestamp, otherwise you run into situations like these (note the Jan 1, 1970 timestamp): 

I think it makes sense that the true timestamp of a transaction be the time at which the block containing that transaction is mined (not when the transaction was broadcast). In that sense the 0 is not bad, but in any case the most important part is that it should be consistent across receive/send transaction types.

Another small problem we have right now is that because the transaction data is not updated once the transaction is added to the cache/db, a transaction seen first when it was in the mempool will forever keep a timestamp of 0, even after more syncs of the wallet once the transaction has been mined. The only solution at the moment is to delete the wallet and recover (mobile) or delete the cache and recover (cli).

RCasatta:
### Description

Make fields `fee` (renamed from `fees` to be coherent with rpc) and `timestamp` `Option` to better handle the missing information.

Specify in docs `timestamp` refer to the timestamp of the block the tx is included in (removed the initialization to `now()` for sending tx)

UPDATE: as per @LLFourn suggestion, timestamp and height have been grouped in the `confirmation_time` field

Fix #369

### Notes to the reviewers

all tests are modified to use `unwrap_or(0)` in place where `fee` was used

In case a tx is confirmed but later reorged and included in the reorged block at the same height, the timestamp will be wrong. The fix would be detrimental for performance in the most common cases so I didn't implement it.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/testutils/blockchain_tests.rs
+++ b/src/testutils/blockchain_tests.rs
@@ -396,7 +396,7 @@ macro_rules! bdk_blockchain_tests {
                 assert_eq!(list_tx_item.txid, txid, "incorrect txid");
                 assert_eq!(list_tx_item.received, 50_000, "incorrect received");
                 assert_eq!(list_tx_item.sent, 0, "incorrect sent");
-                assert_eq!(list_tx_item.height, None, "incorrect height");
+                assert_eq!(list_tx_item.confirmation_time, None, "incorrect confirmation time");
             }
 
             #[test]
@@ -454,7 +454,7 @@ macro_rules! bdk_blockchain_tests {
                 assert_eq!(list_tx_item.txid, txid, "incorrect txid");
                 assert_eq!(list_tx_item.received, 105_000, "incorrect received");
                 assert_eq!(list_tx_item.sent, 0, "incorrect sent");
-                assert_eq!(list_tx_item.height, None, "incorrect height");
+                assert_eq!(list_tx_item.confirmation_time, None, "incorrect confirmation_time");
             }
 
             #[test]
@@ -515,7 +515,7 @@ macro_rules! bdk_blockchain_tests {
                 assert_eq!(list_tx_item.txid, txid, "incorrect txid");
                 assert_eq!(list_tx_item.received, 50_000, "incorrect received");
                 assert_eq!(list_tx_item.sent, 0, "incorrect sent");
-                assert_eq!(list_tx_item.height, None, "incorrect height");
+                assert_eq!(list_tx_item.confirmation_time, None, "incorrect confirmation_time");
 
                 let new_txid = test_client.bump_fee(&txid);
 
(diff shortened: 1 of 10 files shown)


thunderbiscuit:
Great stuff! 

Do I understand correctly that this `Option` will basically return one of two things, either the proper timestamp/fee, or return `None`? (just reading about the `Option` type).

LLFourn:
Concept ACK. Given that both height and timestamp will be `None` in the same case can we roll these into the same struct like `ConfirmationTime` or something?

RCasatta:
> Do I understand correctly that this `Option` will basically return one of two things, either the proper timestamp/fee, or return `None`? (just reading about the `Option` type).

yes, exactly

> Given that both height and timestamp will be None in the same case can we roll these into the same struct like ConfirmationTime or something?

will do

src/testutils/blockchain_tests.rs
```diff
@@ -749,12 +777,12 @@ macro_rules! bdk_blockchain_tests {
...
                 assert_eq!(new_details.received, 0, "incorrect received after change removal");
 
-                assert!(new_details.fees > details.fees, "incorrect fees");
+                assert!(new_details.fee.unwrap_or(0) > details.fee.unwrap_or(0), "incorrect fees");
             }
 
             #[test]
             #[serial]
-            fn test_sync_bump_fee_add_input() {
+            fn test_sync_bump_fee_add_input_simple() {
```
RCasatta, line 785: I added `simple`, because you can't launch tests singularly if they share a common prefix with others

LLFourn:
ACK 0bbfa5f989517540b0ca8bedd62a23066823fa25 it's a little interesting that you have to have this `::new()` function to check if they are both not options. It would be nice if this could be enforced during de-serialization instead.

afilini:
I haven't reviewed the code thoroughly yet, but I'm assuming this affects the database representation of our structs somehow. What happens if I try to load a db created with the older version?

RCasatta:
> I haven't reviewed the code thoroughly yet, but I'm assuming this affects the database representation of our structs somehow. What happens if I try to load a db created with the older version?

If my test is correct it works and it simply initializes `fee` and `confirmation_time` to `None`.

I used bdk-cli master, synced a wallet, then switched to this branch https://github.com/RCasatta/bdk-cli/tree/updated asked the balance of the same wallet, synced again and it works (my local bdk is on the options branch)

afilini (approved):

PR #371, opened 2021-06-16
Replace set_single_recipient with drain_to

LLFourn:
### Description

What set_single_recipient does turns out to be useful with multiple
recipients.
Effectively, set_single_recipient was simply creating a change
output that was arbitrarily required to be the only output.
But what if you want to send excess funds to one address but still have
additional recipients who receive a fixed value?
Generalizing this to `drain_to` simplifies the logic and removes several
error cases while also allowing new use cases.

### Notes to the reviewers

Note that this also changes `maintain_single_recipient` to `maintain_drain-to` which takes an argument specifying which output to reduce.

#### All Submissions:


src/wallet/tx_builder.rs
```diff
@@ -560,49 +560,79 @@ impl<'a, B, D: BatchDatabase, Cs: CoinSelectionAlgorithm<D>> TxBuilder<'a, B, D,
...
-    /// entirely given the higher new fee rate.
+    /// Bump the fee of a transaction made with [`drain_to`]
     ///
-    /// If extra inputs are added and they are not entirely consumed in fees, a change output will not
-    /// be added; the existing output will simply grow in value.
+    /// Unless extra inputs are specified with [`add_utxo`] or [`drain_wallet`], this function will
+    /// reduce the value of the `drain_to` output (removing it completely if necessary).
     ///
-    /// Fails if the transaction has more than one outputs.
+    /// Fails if the `drain_to` script can't be found amoung the `script_pubkey`s of the transaction
```
afilini: TIL https://www.askdifference.com/amoung-vs-among 😅

src/wallet/tx_builder.rs
```diff
@@ -560,49 +560,79 @@ impl<'a, B, D: BatchDatabase, Cs: CoinSelectionAlgorithm<D>> TxBuilder<'a, B, D,
...
     ///
+    /// [`drain_to`]: Self::drain_to
     /// [`add_utxo`]: Self::add_utxo
-    pub fn maintain_single_recipient(&mut self) -> Result<&mut Self, Error> {
-        let mut recipients = self.params.recipients.drain(..).collect::<Vec<_>>();
-        if recipients.len() != 1 {
-            return Err(Error::SingleRecipientMultipleOutputs);
+    /// [`drain_wallet`]: Self::drain_wallet
+    pub fn maintain_drain_to(&mut self, drain_to: Script) -> Result<&mut Self, Error> {
+        match self.params.recipients.iter().position(|x| x.0 == drain_to) {
```
afilini: nit: maybe deconstructing the tuple makes it a bit more readable? like `|(script, _)| script == drain_to`

src/wallet/mod.rs
```diff
@@ -722,8 +700,8 @@ where
     /// *repalce by fee* (RBF). If the transaction can be fee bumped then it returns a [`TxBuilder`]
     /// pre-populated with the inputs and outputs of the original transaction.
     ///
-    /// **NOTE**: if the original transaction was made with [`TxBuilder::set_single_recipient`],
-    /// the [`TxBuilder::maintain_single_recipient`] flag should be enabled to correctly reduce the
+    /// **NOTE**: if the original transaction was made with [`TxBuilder::drain_to`],
+    /// the [`TxBuilder::maintain_drain_to`] flag should be enabled to correctly reduce the
     /// only output's value in order to increase the fees.
```
afilini: nit: "only" doesn't necessarily apply here anymore, right?

afilini (commented):
Just a few minor comments. I really like this proposal, more functionality with less complexity!

src/wallet/mod.rs
LLFourn: Yes! Thanks. What about we just rename this to `bump_by_shrinking` to remove the verbal association  with `drain_to` because it may be useful without `drain_to` (it can sitll be mentioned in the docs).
Or thinking ahead to when `bump_fee` is just a tx builder method that adds all the inputs and outputs of the previous transaction (so you can bump multiple at once), we could have a `allow_shrinking(script_pubkey)` which would push the associated output into a list that should be reduced to pay for a fee bump before adding further inputs. I'll leave that for the next PR but that's what I am thinking.

afilini: I really like `allow_shrinking()`, I think it's very self explanatory. In the future we can extend that to take a `Vec` of script pubkeys instead of a single one

CHANGELOG.md
```diff
@@ -7,9 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Wallet
-#### Added
-- Bitcoin core RPC added as blockchain backend
-- Add a `verify` feature that can be enable to verify the unconfirmed txs we download against the consensus rules
+
+- Added Bitcoin core RPC added as blockchain backend
+- Added a `verify` feature that can be enable to verify the unconfirmed txs we download against the consensus rules
+- Removed and replaced `set_single_recipient` with more general `drain_to` and replaced `maintain_single_recipient` with `allow_shrinking`.
```
LLFourn, line 13: I think the correct style is to have "added" in the bullet rather than subheading added?

LLFourn:
I've just changed `maintain_single_recipient` to `allow_shrinking` as discussed.

(sorry to reviewers I accidentally put the whole thing into one commit while rebasing!)

src/wallet/tx_builder.rs
```diff
@@ -560,49 +560,81 @@ impl<'a, B, D: BatchDatabase, Cs: CoinSelectionAlgorithm<D>> TxBuilder<'a, B, D,
         self
     }
 
-    /// Set a single recipient that will get all the selected funds minus the fee. No change will
-    /// be created
+    /// Sets the address to *drain* excess coins to.
     ///
-    /// This method overrides any recipient set with [`set_recipients`](Self::set_recipients) or
-    /// [`add_recipient`](Self::add_recipient).
+    /// Usually, when there is excess coins they are sent to a change address generated by the
```
tcharding:
```suggestion
    /// Usually, when there are excess coins they are sent to a change address generated by the
```

src/wallet/tx_builder.rs
```diff
@@ -560,49 +560,81 @@ impl<'a, B, D: BatchDatabase, Cs: CoinSelectionAlgorithm<D>> TxBuilder<'a, B, D,
...
+    /// wallet. This option replaces the usual change address with an arbitrary `script_pubkey` of
+    /// your choosing. Just as with a change output, if the drain output is not needed (the excess
+    /// coins are too small) it will not be included in the resulting transaction. The only
+    /// difference is that it is valid to use `drain_to` without setting any ordinary recipients
+    /// with [`add_recipient`] (but it is perfectly find to add recipients as well).
```
tcharding:
```suggestion
    /// with [`add_recipient`] (but it is perfectly fine to add recipients as well).
```

src/wallet/tx_builder.rs
```diff
@@ -560,49 +560,81 @@ impl<'a, B, D: BatchDatabase, Cs: CoinSelectionAlgorithm<D>> TxBuilder<'a, B, D,
...
+    /// # use std::str::FromStr;
+    /// # use bitcoin::*;
+    /// # use bdk::*;
+    /// # use bdk::wallet::tx_builder::CreateTx;
+    /// # let to_address = Address::from_str("2N4eQYCbKUHCCTUjBJeHcJp9ok6J2GZsTDt").unwrap();
+    /// # let wallet = doctest_wallet!();
+    /// let mut tx_builder = wallet.build_tx();
+    ///
+    /// tx_builder
+    ///     // spend all outputs in this wallet
```
tcharding:
```suggestion
    ///     // Spend all outputs in this wallet.
```
Public docs, lets be anal about punctuation.

src/wallet/tx_builder.rs
```diff
@@ -560,49 +560,81 @@ impl<'a, B, D: BatchDatabase, Cs: CoinSelectionAlgorithm<D>> TxBuilder<'a, B, D,
...
+    ///     .drain_wallet()
+    ///     // send the excess (which is all the coins minus the fee) to this address
```
tcharding:
```suggestion
    ///     // Send the excess (which is all the coins minus the fee) to this address.
```

src/wallet/tx_builder.rs
```diff
@@ -560,49 +560,81 @@ impl<'a, B, D: BatchDatabase, Cs: CoinSelectionAlgorithm<D>> TxBuilder<'a, B, D,
...
-    /// [`add_utxo`]: Self::add_utxo
-    pub fn maintain_single_recipient(&mut self) -> Result<&mut Self, Error> {
-        let mut recipients = self.params.recipients.drain(..).collect::<Vec<_>>();
-        if recipients.len() != 1 {
-            return Err(Error::SingleRecipientMultipleOutputs);
+    /// Explicitly tells the wallet that it is allowed to reduce the fee of the output matching this
+    /// `script_pubkey` in order to bump the transaction fee. Without specifying this the wallet
+    /// will attempt to find a change output to shrink instead.
+    ///
+    /// **Note** that the output may shrunk to below the dust limit and therefore removed. If it is
```
tcharding:
```suggestion
    /// **Note** that the output may shrink to below the dust limit and therefore be removed. If it is
```

tcharding (commented):
Few docs fixes, the code all looks good to me.

afilini (approved):

PR #375, opened 2021-06-18
Parallel blockchain tests

RCasatta:
### Description

Our Blockchain tests must be executed serially because we start only 1 backend node and interleaving test could lead to wrong results.
This PR adds dev-dependencies `bitcoind` and `electrsd` making it easier to start/stop bitcoind and electrs independent processes for each test.
This allows to launch tests in parallel and make it easier to launch single test (providing an env var to bitcoind/electrs executables).

Even if launching multiple nodes it's more costly, parallel execution leads to improve test execution time:
test-electrum before:
`test result: ok. 17 passed; 0 failed; 0 ignored; 0 measured; 194 filtered out; finished in 83.41s`
test-electrum after:
`test result: ok. 17 passed; 0 failed; 0 ignored; 0 measured; 194 filtered out; finished in 25.41s`

Other than the parallelism performance are improved because electrsd support signalling to electrs to start indexing.

### Notes to the reviewers

I removed docker entirely, downloading needed executables. The other options was to keep docker as an executable provider, but I preferred the former.
UPDATE: bitcoind crate now automatically download the needed binary if the feature is specified, I plan to do the same for electrs. ~~After that we could decide to remove feature "test-*" since there will be no pre-requirements (`cargo test --features rpc` will run rpc tests)~~ Removing "test-*" would have required having some dep moved to non-dev dependencies so I avoided that. But now test-blockchains is activated whenever "test-*" is enabled

For bitcoind binary releases make it easier to download the binaries, while unfortunately electrs doesn't ship binaries (and we need also the outdated esplora version) so I did the build through [github action](https://github.com/RCasatta/electrsd/actions/workflows/build_electrs.yml) and hosted that on github electrsd repository. 

#### All Submissions:


src/testutils/blockchain_tests.rs
```diff
@@ -6,38 +6,52 @@ use bitcoin::{Address, Amount, Script, Transaction, Txid};
...
-        .unwrap();
-        let electrum = ElectrumClient::new(&get_electrum_url()).unwrap();
+    pub fn new(bitcoind_exe: String, electrs_exe: String) -> Self {
+        debug!("launching {} and {}", &bitcoind_exe, &electrs_exe);
+        let bitcoind = BitcoinD::new(bitcoind_exe).unwrap();
+
+        #[cfg(feature = "test-esplora")]
+        let http_enabled = true;
+        #[cfg(not(feature = "test-esplora"))]
+        let http_enabled = false;
```
afilini: I think this can be replaced with:

```suggestion
        let http_enabled = cfg!(feature = "test-esplora");
```

Cargo.toml
```diff
@@ -63,15 +62,14 @@ rpc = ["bitcoincore-rpc"]
...
+test-esplora = ["esplora", "electrsd/legacy"]
 test-md-docs = ["electrum"]
 
 [dev-dependencies]
 lazy_static = "1.4"
 env_logger = "0.7"
 clap = "2.33"
-serial_test = "0.4"
-bitcoind = "0.10.0"
+electrsd = { version="0.4.0", features = ["trigger"] }
```
afilini: nit: use `0.4` instead so that we can push patch-level updates if we need to

afilini:
> I removed docker entirely, downloading needed executables. The other options was to keep docker as an executable provider, but I preferred the former.

I remember we've talked about this IRL, but I don't remember the conclusion we reached. What was the rationale for removing the containers?

RCasatta:
> > I removed docker entirely, downloading needed executables. The other options was to keep docker as an executable provider, but I preferred the former.
> 
> I remember we've talked about this IRL, but I don't remember the conclusion we reached. What was the rationale for removing the containers?

So the problem I had with executables was basically that electrs doesn't release those (while bitcoin has "official" binaries), so the first version was building electrs and improving execution times with caching. However, caching ends up not being always leverage-able during development (eg. any change in toml invalidates it). In the end, I realized that releasing electrs executables somewhere (I used electrsd release but we can change that obviously) was simpler than releasing a docker image. So I think we can get rid of docker if we use it just to provide executables (port conflicts and process killing is handled at crate level).

Moreover, I think bitcoind and electrsd  crates could be improved in the future to automatically download binaries in `build.rs`

Cargo.toml
RCasatta: 48dcc45b9b7f7551411108e70ed9121e90bacd44

src/testutils/blockchain_tests.rs
RCasatta: nice, done in 48dcc45b9b7f7551411108e70ed9121e90bacd44

notmandatory:
I'm in agreement on the concept of downloading the needed bins instead of using the docker images to provide them to the CI. Even though the binaries are small, I'd also like to see the same [caching scheme](https://github.com/RCasatta/electrsd/pull/2) from your `electrsd` repo, or I can apply it in a follow-up PR. Also as part of this PR it'd be good to update the `run_blockchain_tests.sh` script to make local testing easier.

RCasatta:
>  Also as part of this PR it'd be good to update the run_blockchain_tests.sh script to make local testing easier.

I would like to do automatic bins download also for electrs, then a simple `cargo test --features rpc` or `cargo test --features electrum` will execute relative test without the need of a shell script

> I'd also like to see the same caching scheme from your electrsd repo,

Yeah right, the `~/.cargo/bitcoin`directory where binaries will end up is missing, I will update

notmandatory:
> I would like to do automatic bins download also for electrs, then a simple `cargo test --features rpc` or `cargo test --features electrum` will execute relative test without the need of a shell script

So you mean `testutils/blockchain_tests.rs` or the `electrsd` and `bitcoind` projects will download (if needed) and manage the right versions for the `electrs` and `bitcoind` binaries? if so then just adding the download location to the github actions cache should work fine.

RCasatta:
> > I would like to do automatic bins download also for electrs, then a simple `cargo test --features rpc` or `cargo test --features electrum` will execute relative test without the need of a shell script
> 
> `electrsd` and `bitcoind` projects will download (if needed) and manage the right versions for the `electrs` and `bitcoind` binaries? 

yes, defaulting to a specific version of bitcoind and electrs via a default feature (changeable with specific version with something like `cargo test --no-default-features --features bitcoind/0_17_1,rpc`

README.md
```diff
@@ -151,6 +151,25 @@ fn main() -> Result<(), bdk::Error> {
...
+### Integration testing
+
+Integration testing require testing features, for example:
+
+```
+cargo test --features test-electrum
+```
+
+The other options are `test-esplora` or `test-rpc`.
+Note that `electrs` and `bitcoind` binaries are automatically downloaded (on mac and linux), to specify your own one must use `--no-default-features` and provide `BITCOIND_EXE` and `ELECTRS_EXE` as environment variables.
```
notmandatory: nit: 
```suggestion
Note that `electrs` and `bitcoind` binaries are automatically downloaded (on mac and linux), to specify you already have installed binaries you must use `--no-default-features` and provide `BITCOIND_EXE` and `ELECTRS_EXE` as environment variables.
```

README.md
```diff
@@ -151,6 +151,25 @@ fn main() -> Result<(), bdk::Error> {
...
+```
+cargo test
+```
+
+### Integration testing
+
+Integration testing require testing features, for example:
+
+```
+cargo test --features test-electrum
```
LLFourn, line 167: 💡  Can the feature flag situation be improved? How about `test-blockchains` + `electrum` does this and we can get rid of `test-electrum`.

LLFourn (approved):
LGTM as long as @RCasatta is willing to accept PRs on his `electrsd` and `bitcoind` crates using docker instead of downloading the binaries.

README.md
RCasatta: The problem to do this is the different testing features required by esplora (`electrsd/esplora_a33e97e1`) and electrum (`electrsd/electrs_0_8_10`)

Previously, I even though to remove `test-*` completely (and simply run blockchain electrum test with electrum feature), in this case other than previously described issue there was another issue to include dep like `bitcoincore-rpc` only when testing.

RCasatta:
> LGTM as long as @RCasatta is willing to accept PRs on his `electrsd` and `bitcoind` crates using docker instead of downloading the binaries.

I will happily accept feature-gated docker PRs (and I am genuinely interested in the reasons to prefer it over downloading binaries at least in the bitcoind case where there are official binaries)

notmandatory (approved):
ACK , verified test-[rpc,electrum,esplora] all pass on my local linux and osx systems

PR #376, opened 2021-06-18
implement sqlite database

johncantrell97:
### Description

This PR implements all the Database related traits for a database backed by sqlite.  It currently uses cached prepared statements which simplify the implementation from trying to use actual prepared Statement objects owned by the SqliteDatabase struct.  It should have the same performance as long as the cache is large enough to hold all the statements.

In my own project I use a crate called rusqlite_migration to handle migrations but in an attempt to reduce dependencies I borrowed a simple migration implementation by jb55 that manages a version based on the length of the static migrations array.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/database/any.rs
+++ b/src/database/any.rs
@@ -65,6 +65,8 @@ macro_rules! impl_inner_method {
             $enum_name::Memory(inner) => inner.$name( $($args, )* ),
             #[cfg(feature = "key-value-db")]
             $enum_name::Sled(inner) => inner.$name( $($args, )* ),
+            #[cfg(feature = "sqlite")]
+            $enum_name::Sqlite(inner) => inner.$name( $($args, )* ),
         }
     }
 }
@@ -82,10 +84,15 @@ pub enum AnyDatabase {
     #[cfg_attr(docsrs, doc(cfg(feature = "key-value-db")))]
     /// Simple key-value embedded database based on [`sled`]
     Sled(sled::Tree),
+    #[cfg(feature = "sqlite")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "sqlite")))]
+    /// Sqlite embedded database using [`rusqlite`]
+    Sqlite(sqlite::SqliteDatabase),
 }
 
 impl_from!(memory::MemoryDatabase, AnyDatabase, Memory,);
 impl_from!(sled::Tree, AnyDatabase, Sled, #[cfg(feature = "key-value-db")]);
+impl_from!(sqlite::SqliteDatabase, AnyDatabase, Sqlite, #[cfg(feature = "sqlite")]);
 
 /// Type that contains any of the [`BatchDatabase::Batch`] types defined by the library
 pub enum AnyBatch {
@@ -95,6 +102,10 @@ pub enum AnyBatch {
     #[cfg_attr(docsrs, doc(cfg(feature = "key-value-db")))]
     /// Simple key-value embedded database based on [`sled`]
     Sled(<sled::Tree as BatchDatabase>::Batch),
+    #[cfg(feature = "sqlite")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "sqlite")))]
+    /// Sqlite embedded database using [`rusqlite`]
+    Sqlite(<sqlite::SqliteDatabase as BatchDatabase>::Batch),
 }
 
 impl_from!(
@@ -103,6 +114,7 @@ impl_from!(
     Memory,
 );
 impl_from!(<sled::Tree as BatchDatabase>::Batch, AnyBatch, Sled, #[cfg(feature = "key-value-db")]);
+impl_from!(<sqlite::SqliteDatabase as BatchDatabase>::Batch, AnyBatch, Sqlite, #[cfg(feature = "sqlite")]);
 
 impl BatchOperations for AnyDatabase {
     fn set_script_pubkey(
@@ -300,19 +312,25 @@ impl BatchDatabase for AnyDatabase {
             AnyDatabase::Memory(inner) => inner.begin_batch().into(),
             #[cfg(feature = "key-value-db")]
             AnyDatabase::Sled(inner) => inner.begin_batch().into(),
+            #[cfg(feature = "sqlite")]
+            AnyDatabase::Sqlite(inner) => inner.begin_batch().into(),
         }
     }
     fn commit_batch(&mut self, batch: Self::Batch) -> Result<(), Error> {
         match self {
             AnyDatabase::Memory(db) => match batch {
                 AnyBatch::Memory(batch) => db.commit_batch(batch),
-                #[cfg(feature = "key-value-db")]
-                _ => unimplemented!("Sled batch shouldn't be used with Memory db."),
+                _ => unimplemented!("Other batch shouldn't be used with Memory db."),
             },
             #[cfg(feature = "key-value-db")]
             AnyDatabase::Sled(db) => match batch {
                 AnyBatch::Sled(batch) => db.commit_batch(batch),
-                _ => unimplemented!("Memory batch shouldn't be used with Sled db."),
+                _ => unimplemented!("Other batch shouldn't be used with Sled db."),
+            },
+            #[cfg(feature = "sqlite")]
+            AnyDatabase::Sqlite(db) => match batch {
+                AnyBatch::Sqlite(batch) => db.commit_batch(batch),
+                _ => unimplemented!("Other batch shouldn't be used with Sqlite db."),
             },
         }
     }
@@ -337,6 +355,23 @@ impl ConfigurableDatabase for sled::Tree {
     }
 }
 
+/// Configuration type for a [`sqlite::SqliteDatabase`] database
+#[cfg(feature = "sqlite")]
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct SqliteDbConfiguration {
+    /// Main directory of the db
+    pub path: String,
+}
+
+#[cfg(feature = "sqlite")]
+impl ConfigurableDatabase for sqlite::SqliteDatabase {
+    type Config = SqliteDbConfiguration;
+
+    fn from_config(config: &Self::Config) -> Result<Self, Error> {
+        Ok(sqlite::SqliteDatabase::new(config.path.clone()))
+    }
+}
+
 /// Type that can contain any of the database configurations defined by the library
 ///
 /// This allows storing a single configuration that can be loaded into an [`AnyDatabase`]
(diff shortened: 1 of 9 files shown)


src/database/sqlite.rs
```diff
@@ -0,0 +1,894 @@
...
+        )?;
+        Ok(())
+    }
+
+    fn set_raw_tx(&mut self, transaction: &Transaction) -> Result<(), Error> {
+        match self.select_transaction_by_txid(&transaction.txid())? {
+            Some(_) => {
+                // should this ever really happen? wouldn't an adjusted tx
+                // result in a new txid
+                self.update_transaction(&transaction.txid(), &serialize(transaction))?;
```
tcharding: The comment is correct, AFAIU, if the transaction is already in the db we could just noop.

src/database/sqlite.rs
```diff
@@ -0,0 +1,894 @@
...
+                }))
+            },
+            None => {
+                Ok(None)
+            }
+        }
+    }
+
+    fn del_raw_tx(&mut self, txid: &Txid) -> Result<Option<Transaction>, Error> {
+        match self.select_transaction_by_txid(&txid)? {
```
tcharding: Double reference, same below. If you lint with clippy +nightly it will catch these for you.

src/database/sqlite.rs
```diff
@@ -0,0 +1,894 @@
...
+                Ok(None)
+            }
+        }
+    }
+
+    fn get_tx(&self, txid: &Txid, _include_raw: bool) -> Result<Option<TransactionDetails>, Error> {
+        // always returns with Option<Transaction> (if exists in db)
+        // are there cases where we NEED it to be None even if it exists in db?
+        // or is this just a performance improvement?
+        match self.select_transaction_details_by_txid(&txid)? {
```
tcharding: I would assume that `Err` is for db connection errors, `None` means not in db, and `Some` means found in db.

src/database/sqlite.rs
```diff
@@ -0,0 +1,894 @@
...
+        let keychain = serde_json::to_string(&keychain)?;
+        match self.select_last_deriviation_index_by_keychain(keychain)? {
+            Some(value) => {
+                Ok(Some(value))
+            },
+            None => {
+                Ok(None)
+            }
+        }
+    }
```
tcharding: Just for your interest, this can also be written as:
```
    fn get_last_index(&self, keychain: KeychainKind) -> Result<Option<u32>, Error> {
        let keychain = serde_json::to_string(&keychain)?;
        let value = self.select_last_deriviation_index_by_keychain(keychain)?;
        Ok(value)
    }
``

**Review by tcharding — COMMENTED** [review]
A few minor suggestions/ideas. Don't read too much into my review, I'm not that deep in `bdk`.

Cheers

File: src/database/sqlite.rs (hunk 3 above)
Comment (johncantrell97) [reply src/database/sqlite.rs]: this comment is actually referring to the fact that my select_transaction_details method ALWAYS includes the raw transaction in the result if it exists in the database.  The `get_tx` method accepts a `include_raw` flag for whether or not we are supposed to fetch the raw transaction.

So I was just worried about a potential case where the user sets include_raw to false but it's still included.  Feels like this should never be an actual problem other than perhaps a performance issue.

File: src/database/sqlite.rs (hunk 2 above)
Comment (johncantrell97) [reply src/database/sqlite.rs]: thanks, yeah I see that clippy finds it. cool.

File: src/database/sqlite.rs (hunk 1 above)
Comment (johncantrell97) [reply src/database/sqlite.rs]: yeah makes sense, I guess in theory there could be a change to the transaction struct that is outside of the txid calculation.  Can't hurt to keep the update here I suppose.

File: src/database/sqlite.rs (hunk 4 above)
Comment (johncantrell97) [reply src/database/sqlite.rs]: ah good point, looks cleaner.

File: src/database/any.rs [hunk 5]
```diff
@@ -300,19 +312,26 @@ impl BatchDatabase for AnyDatabase {
...
             AnyDatabase::Sled(inner) => inner.begin_batch().into(),
+            #[cfg(feature = "sqlite")]
+            AnyDatabase::Sqlite(inner) => inner.begin_batch().into(),
         }
     }
     fn commit_batch(&mut self, batch: Self::Batch) -> Result<(), Error> {
         match self {
             AnyDatabase::Memory(db) => match batch {
                 AnyBatch::Memory(batch) => db.commit_batch(batch),
                 #[cfg(feature = "key-value-db")]
```
notmandatory: I think we can remove this line, otherwise when `sqlite` feature is used with `--no-default-features` we get a `pattern Sqlite(_) not covered` error here.
```suggestion
```

notmandatory:
A couple places in the CI workflows also need the new `sqlite` feature added:
1. `cont_integration.yml`: build-test: strategy: matrix: features:
2. `nightly_docs.yml`: build_docs: steps: Build Docs:
    `run: cargo rustdoc --verbose --features=compiler,electrum,esplora,ureq,compact_filters,key-value-db,all-keys,sqlite -- --cfg docsrs -Dwarnings`

src/database/sqlite.rs
```diff
@@ -0,0 +1,987 @@
...
+    "CREATE TABLE last_derivation_indices (keychain TEXT, value INTEGER);",
+    "CREATE UNIQUE INDEX idx_indices_keychain ON last_derivation_indices(keychain);",
+    "CREATE TABLE checksums (keychain TEXT, checksum BLOB);",
+    "CREATE INDEX idx_checksums_keychain ON checksums(keychain);",
+];
+
+/// Sqlite database stored on filesystem
+///
+/// This is a permanent storage solution for devices and platforms that provide a filesystem.
+/// [`database`]: crate::database
```
notmandatory: You can verify docs created in CI with this command: 
```shell
cargo +nightly rustdoc --verbose --features=compiler,electrum,esplora,ureq,compact_filters,all-keys,sqlite -- --cfg docsrs -Dwarnings
```

```suggestion
/// [`crate::database`]
```

src/database/sqlite.rs
```diff
@@ -0,0 +1,987 @@
...
+    /// Instantiate a new SqliteDatabase instance by creating a connection
+    /// to the database stored at path
+    pub fn new(path: String) -> Self {
+        let connection = get_connection(&path).unwrap();
+        SqliteDatabase { path, connection }
+    }
+    /// Insert a script at the specific index and keychain
+    pub fn insert_script_pubkey(
+        &self,
+        keychain: String,
```
notmandatory, line 61: I think it'd be safer in this and other insert functions that take a `keychain` to use the type `KeychainKind`, then in the function convert it to a string before executing the insert or update.  The risk is if someone inserts data directly on the `SqliteDatabase` they could use any random value for `keychain`, and then the query functions in this block that assume the result for `keychain` will be decodable as `json::from_str(&keychain)` will fail.

notmandatory: May be useful to add these, seems to be how `rust-bitcoin` handles converting enums like `AddressType` to/from strings:
```rust
impl fmt::Display for KeychainKind {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        f.write_str(match *self {
            KeychainKind::Internal => "internal",
            KeychainKind::External => "external",
        })
    }
}

impl FromStr for KeychainKind {
    type Err = ();

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
            "internal" => Ok(KeychainKind::Internal),
            "external" => Ok(KeychainKind::External),
            _ => Err(()),
        }
    }
}
```

notmandatory:
Also this one needs a rebase to pickup latest `Blockchain` tests required for CI to pass.

src/database/sqlite.rs
johncantrell97: yeah we should probably do this.  I guess technically those functions are exposed as public but maybe they shouldn't be? They are functions used to implement the Database / BatchOperations traits which I assume are the main fn's used externally.

src/database/any.rs
```diff
@@ -300,19 +312,25 @@ impl BatchDatabase for AnyDatabase {
...
+            AnyDatabase::Sqlite(inner) => inner.begin_batch().into(),
         }
     }
     fn commit_batch(&mut self, batch: Self::Batch) -> Result<(), Error> {
         match self {
             AnyDatabase::Memory(db) => match batch {
                 AnyBatch::Memory(batch) => db.commit_batch(batch),
-                #[cfg(feature = "key-value-db")]
-                _ => unimplemented!("Sled batch shouldn't be used with Memory db."),
+                _ => unimplemented!("Other batch shouldn't be used with Memory db."),
```
notmandatory, line 324: Adding this should fix the clippy errors when `--no-default-features` is used.
```suggestion
               #[cfg(any(feature = "key-value-db", feature = "sqlite"))]
                _ => unimplemented!("Other batch shouldn't be used with Memory db."),
```

src/database/sqlite.rs
notmandatory: Yes I think only the `SqliteDatabase::new()` function needs to be public and then there's no way to insert a bad `keychain` value.

src/database/sqlite.rs
```diff
@@ -0,0 +1,987 @@
...
+            let raw_tx: Option<Vec<u8>> = row.get(7)?;
+            let tx: Option<Transaction> = match raw_tx {
+                Some(raw_tx) => {
+                    let tx: Transaction = deserialize(&raw_tx)?;
+                    Some(tx)
+                }
+                None => None,
+            };
+
+            let confirmation_time = match (timestamp, height) {
```
notmandatory: To make `clippy` for rust `1.53.0` happy (and a bit easier to read) should be:
```rust
let confirmation_time = match (height, timestamp) {
    (Some(height), Some(timestamp)) => Some(ConfirmationTime { height, timestamp }),
    _ => None,
};
```
This will also work for latest stable rust `1.55.0` which we'll update to soon I hope.

src/database/sqlite.rs
```diff
@@ -0,0 +1,987 @@
...
+                let raw_tx: Option<Vec<u8>> = row.get(6)?;
+                let tx: Option<Transaction> = match raw_tx {
+                    Some(raw_tx) => {
+                        let tx: Transaction = deserialize(&raw_tx)?;
+                        Some(tx)
+                    }
+                    None => None,
+                };
+
+                let confirmation_time = match (timestamp, height) {
```
notmandatory: Same `clippy` issue:
```rust
  let confirmation_time = match (height, timestamp) {
      (Some(height), Some(timestamp)) => Some(ConfirmationTime { height, timestamp }),
      _ => None,
  };
```

src/database/sqlite.rs
```diff
@@ -0,0 +1,987 @@
...
+            let txid: Vec<u8> = row.get(0)?;
+            let txid: Txid = deserialize(&txid)?;
+            let timestamp: Option<u64> = row.get(1)?;
+            let received: u64 = row.get(2)?;
+            let sent: u64 = row.get(3)?;
+            let fee: Option<u64> = row.get(4)?;
+            let height: Option<u32> = row.get(5)?;
+            let verified: bool = row.get(6)?;
+
+            let confirmation_time = match (timestamp, height) {
```
notmandatory: Same `clippy` issue:
```rust
let confirmation_time = match (height, timestamp) {
      (Some(height), Some(timestamp)) => Some(ConfirmationTime { height, timestamp }),
      _ => None,
  };
```

notmandatory (approved):
ACK 919522a456f40ffed8bb084d4f11a34e74197b10

PR #377, opened 2021-06-18
add Copy trait to Progress types

jb55:
I have this patch applied locally as it seems to be required in some multi-threaded contexts? Ideally I wouldn't have to keep a fork of bdk for my project so here we are.

### DIFF
--- a/src/blockchain/mod.rs
+++ b/src/blockchain/mod.rs
@@ -174,7 +174,7 @@ impl Progress for Sender<ProgressData> {
 }
 
 /// Type that implements [`Progress`] and drops every update received
-#[derive(Clone)]
+#[derive(Clone, Copy)]
 pub struct NoopProgress;
 
 /// Create a new instance of [`NoopProgress`]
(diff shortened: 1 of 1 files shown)


notmandatory:
According to [`Copy` trait docs](https://doc.rust-lang.org/std/marker/trait.Copy.html ) you should also add derive `Clone`.  Otherwise this looks fine to me.

jb55:
On Fri, Jun 18, 2021 at 04:31:26PM -0700, Steve Myers wrote:
>According to [`Copy` trait docs](https://doc.rust-lang.org/std/marker/trait.Copy.html ) you should
>also add derive `Clone`.  Otherwise this looks fine to me.

hmm? Clone is already being derived, unless I misunderstand what you're saying.

notmandatory (approved):
ACK 

Sorry, disregard my above comment.

notmandatory:
Looks like it just needs a rebase to pickup the recent CI change to latest Rust version then I can merge it.

jb55:
done

RCasatta:
Was trying to merge but our repo enforces signatures on commits, would you like to sign it? If not I can sign it

PR #378, opened 2021-06-19
Peer discovery for compact block filters

johncantrell97:
### Description

This adds a new p2p discovery module to the compact block filters blockchain implementation in BDK.  The high level idea is for a wallet that wants to be a light client using compact block filters to not need to specify an explicit peer to sync the chain.

The current implementation starts with the DNS seeds and performs a network search using `GetAddr` messages with connected peers.  It builds a directory with successfully connected addresses and their services flag.  It caches this directory to the filesystem to speed up discovery on subsequent startups of the wallet.

Currently there is no way to extract a peer or set of peers from this directory but that is a good next step.

It currently uses a configurable amount of worker threads to perform the crawl and I recommend spawning a thread to run the `manage` control loop that spawns the workers and handles worker messages.

The high level structs and architecture looks like this:

`AddressManager` -- the top level struct that gets instantiated to start everything.  The manager holds the directory, a shared list of addresses being searched, and channels for messaging between the worker threads and the manager.

`AddressDirectory` -- Right now it holds two indexes, one for cbf enabled addresses and the other for the rest.  Each index is a `VecDeque<NodeAddress>` where a `NodeAddress` is just an SocketAddr and ServicesFlag.  The indexes probably should not be VecDeque's, this was just easier for serializing from this to the filesystem and back into a the same structure the workers use for managing a list of addresses to search.  

`AddressDiscovery` -- a struct wrapped in Arc<Mutex> for sharing between worker threads.  It contains a list of addresses to be searched and which addresses have already been visited.  

`AddressWorker` -- a struct representing a network crawler worker.  it maintains a reference to the AddressDiscovery struct and a channel for sending messages to the AddressManager.

`DiscoveryProgress` -- a trait for receiving progress updates from the address manager.  this is modeled after the Progress trait used to track progress when syncing the blockchain. I provided implementations for noop, logger, and Sender<DiscoveryData>.

The changes to `peer` were to prevent the workers from getting stuck on a particular peer.  The current implementation blocks indefinitely when connecting and when receiving responses.  I added a connect_with_timeout and also changed the verack and version messages to timeout after 1 second.  We probably don't want to keep it exactly like this.  It might not even make sense to use the Peer struct when doing the discovery.  It also might make sense to rewrite some of the peer network messaging logic in a way that doesn't block and expect a single response.

Some example code if you want to try running an AddressManager search with BDK from this fork:

```
use std::thread;
use bdk::blockchain::compact_filters::{AddressManager, log_discovery_progress};
use bdk::bitcoin::network::constants::Network;
use simple_logger::SimpleLogger;

fn main() -> Result<(), bdk::Error> {
    SimpleLogger::new().init().unwrap();
    
    let network = Network::Bitcoin;
    let mut address_manager = AddressManager::new(network, "cache.json".to_string(), 8);
    
    let manager_handle = thread::spawn(move || {
        address_manager.manage(log_discovery_progress());
    });

    manager_handle.join().unwrap();
    Ok(())
}
```

### Notes to the reviewers

This is not really ready for a real review.  This is to help start documenting the idea and get Raj (+others?) in the loop as early as possible.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/blockchain/compact_filters/address_manager.rs
+++ b/src/blockchain/compact_filters/address_manager.rs
@@ -571,18 +571,48 @@ impl<P: DiscoveryProgress> AddressManager<P> {
     }
 
     /// Get all the known CBF addresses
-    pub fn get_known_cbfs(&self) -> Vec<SocketAddr> {
-        self.directory.cbf_nodes.iter().copied().collect()
+    pub fn get_known_cbfs(&self) -> Option<Vec<SocketAddr>> {
+        let addresses = self
+            .directory
+            .cbf_nodes
+            .iter()
+            .copied()
+            .collect::<Vec<SocketAddr>>();
+
+        match addresses.len() {
+            0 => None,
+            _ => Some(addresses),
+        }
     }
 
     /// Get all the known regular addresses
-    pub fn get_known_non_cbfs(&self) -> Vec<SocketAddr> {
-        self.directory.non_cbf_nodes.iter().copied().collect()
+    pub fn get_known_non_cbfs(&self) -> Option<Vec<SocketAddr>> {
+        let addresses = self
+            .directory
+            .non_cbf_nodes
+            .iter()
+            .copied()
+            .collect::<Vec<SocketAddr>>();
+
+        match addresses.len() {
+            0 => None,
+            _ => Some(addresses),
+        }
     }
 
     /// Get previously tried addresses
-    pub fn get_previously_tried(&self) -> Vec<SocketAddr> {
-        self.directory.previously_sent.iter().copied().collect()
+    pub fn get_previously_tried(&self) -> Option<Vec<SocketAddr>> {
+        let addresses = self
+            .directory
+            .previously_sent
+            .iter()
+            .copied()
+            .collect::<Vec<SocketAddr>>();
+
+        match addresses.len() {
+            0 => None,
+            _ => Some(addresses),
+        }
     }
 }
 
--- a/src/blockchain/compact_filters/address_manager.rs
+++ b/src/blockchain/compact_filters/address_manager.rs
@@ -690,7 +690,7 @@ mod test {
     use super::*;
 
     #[test]
-    fn test_crawl_times() {
+    fn test_address_manager() {
         // Initiate a manager with an non existent cache file name.
         // It will create a new cache file
         let mut manager = AddressManager::new(
@@ -704,9 +704,11 @@ mod test {
         .unwrap();
 
         // start the crawlers and time them
+        println!("Starting manager and initial fetch");
         let start = std::time::Instant::now();
         manager.fetch().unwrap();
         let duration1 = start.elapsed();
+        println!("Completed Initial fetch");
 
         // Create a new manager from existing cache and fetch again
         let mut manager = AddressManager::new(
@@ -720,31 +722,18 @@ mod test {
         .unwrap();
 
         // start the crawlers and time them
+        println!("Starting new fetch with previous cache");
         let start = std::time::Instant::now();
         manager.fetch().unwrap();
         let duration2 = start.elapsed();
+        println!("Completed new fetch()");
 
         println!("Time taken for initial crawl: {:#?}", duration1);
         println!("Time taken for next crawl {:#?}", duration2);
-    }
 
-    #[test]
-    fn test_buffer_management() {
-        // Initiate a manager with an non existent cache file name.
-        // It will create a new cache file
-        let mut manager = AddressManager::new(
-            Network::Bitcoin,
-            "addr_cache".to_string(),
-            20,
-            None,
-            None,
-            LogDiscoveryProgress,
-        )
-        .unwrap();
+        // Check Buffer Management
 
-        // Start the first fetch()
-        manager.fetch().unwrap();
-
+        println!("Checking buffer management");
         // Fetch few new address and ensure buffer goes to zero
         let mut addrs_list = Vec::new();
         for _ in 0..5 {
@@ -763,9 +752,12 @@ mod test {
 
         // Calling fetch again should start crawlers until buffer
         // requirements are matched.
+        println!("Address buffer exhausted, starting new fetch");
         manager.fetch().unwrap();
-
+        println!("Fetch Complete");
         // It should again have a cbf buffer of 5
         assert_eq!(manager.directory.get_cbf_buffer(), 5);
+
+        println!("Buffer management passed");
     }
 }
--- a/src/blockchain/compact_filters/address_manager.rs
+++ b/src/blockchain/compact_filters/address_manager.rs
@@ -1,0 +1,741 @@
+// Bitcoin Dev Kit
+// Written in 2021 by Rajarshi Maitra <rajarshi149@gmail.com>
+//                    John Cantrell <johncantrell97@protonmail.com> 
+//
+// Copyright (c) 2020-2021 Bitcoin Dev Kit Developers
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+use std::fs::File;
+use std::io::prelude::*;
+
+use std::collections::HashSet;
+use std::collections::VecDeque;
+use std::sync::{
+    mpsc::{channel, Receiver, SendError, Sender},
+    Arc, RwLock,
+};
+use std::thread::{self, JoinHandle};
+use std::time::Duration;
+
+use std::net::{SocketAddr, ToSocketAddrs};
+
+use std::sync::PoisonError;
+use std::sync::{MutexGuard, RwLockReadGuard, RwLockWriteGuard, WaitTimeoutResult};
+
+use serde::{Deserialize, Serialize};
+use serde_json::Error as SerdeError;
+
+use super::{Mempool, Peer, PeerError};
+
+use bitcoin::network::{
+    constants::{Network, ServiceFlags},
+    message::NetworkMessage,
+    Address,
+};
+
+/// Default address pool minimums
+const MIN_CBF_BUFFER: usize = 5;
+const MIN_NONCBF_BUFFER: usize = 5;
+
+/// A Discovery structure used by workers
+///
+/// Discovery can be initiated via a cache,
+/// Or it will start with default hardcoded seeds
+pub struct AddressDiscovery {
+    pending: VecDeque<SocketAddr>,
+    visited: HashSet<SocketAddr>,
+}
+
+impl AddressDiscovery {
+    fn new(network: Network, seeds: VecDeque<SocketAddr>) -> AddressDiscovery {
+        let mut network_seeds = AddressDiscovery::seeds(network);
+        let mut total_seeds = seeds;
+        total_seeds.append(&mut network_seeds);
+        AddressDiscovery {
+            pending: total_seeds,
+            visited: HashSet::new(),
+        }
+    }
+
+    fn add_pendings(&mut self, addresses: Vec<SocketAddr>) {
+        for addr in addresses {
+            if !self.pending.contains(&addr) && !self.visited.contains(&addr) {
+                self.pending.push_back(addr);
+            }
+        }
+    }
+
+    fn get_next(&mut self) -> Option<SocketAddr> {
+        match self.pending.pop_front() {
+            None => None,
+            Some(next) => {
+                self.visited.insert(next);
+                Some(next)
+            }
+        }
+    }
+
+    fn seeds(network: Network) -> VecDeque<SocketAddr> {
+        let mut seeds = VecDeque::new();
+
+        let port: u16 = match network {
+            Network::Bitcoin => 8333,
+            Network::Testnet => 18333,
+            Network::Regtest => 18444,
+            Network::Signet => 38333,
+        };
+
+        let seedhosts: &[&str] = match network {
+            Network::Bitcoin => &[
+                "seed.bitcoin.sipa.be",
+                "dnsseed.bluematt.me",
+                "dnsseed.bitcoin.dashjr.org",
+                "seed.bitcoinstats.com",
+                "seed.bitcoin.jonasschnelli.ch",
+                "seed.btc.petertodd.org",
+                "seed.bitcoin.sprovoost.nl",
+                "dnsseed.emzy.de",
+                "seed.bitcoin.wiz.biz",
+            ],
+            Network::Testnet => &[
+                "testnet-seed.bitcoin.jonasschnelli.ch",
+                "seed.tbtc.petertodd.org",
+                "seed.testnet.bitcoin.sprovoost.nl",
+                "testnet-seed.bluematt.me",
+            ],
+            Network::Regtest => &[],
+            Network::Signet => &[],
+        };
+
+        for seedhost in seedhosts.iter() {
+            if let Ok(lookup) = (*seedhost, port).to_socket_addrs() {
+                for host in lookup {
+                    if host.is_ipv4() {
+                        seeds.push_back(host);
+                    }
+                }
+            }
+        }
+        seeds
+    }
+}
+
+/// Crawler structure that will interface with Discovery and public bitcoin network
+///
+/// Address manager will spawn multiple crawlers in separate threads to discover new addresses.
+struct AddressWorker {
+    discovery: Arc<RwLock<AddressDiscovery>>,
+    sender: Sender<(SocketAddr, ServiceFlags)>,
+    network: Network,
+}
+
+impl AddressWorker {
+    fn new(
+        discovery: Arc<RwLock<AddressDiscovery>>,
+        sender: Sender<(SocketAddr, ServiceFlags)>,
+        network: Network,
+    ) -> AddressWorker {
+        AddressWorker {
+            discovery,
+            sender,
+            network,
+        }
+    }
+
+    fn try_receive_addr(&mut self, peer: &Peer) -> Result<(), AddressManagerError> {
+        if let Some(NetworkMessage::Addr(new_addresses)) =
+            peer.recv("addr", Some(Duration::from_secs(1)))?
+        {
+            self.consume_addr(new_addresses)?;
+        }
+
+        Ok(())
+    }
+
+    fn consume_addr(&mut self, addrs: Vec<(u32, Address)>) -> Result<(), AddressManagerError> {
+        let mut discovery_lock = self.discovery.write().map_err(PeerError::from)?;
+        let mut addresses = Vec::new();
+        for network_addrs in addrs {
+            if let Ok(socket_addrs) = network_addrs.1.socket_addr() {
+                addresses.push(socket_addrs);
+            }
+        }
+        discovery_lock.add_pendings(addresses);
+
+        Ok(())
+    }
+
+    fn work(&mut self) -> Result<(), AddressManagerError> {
+        loop {
+            let next_address = {
+                let mut address_discovery = self.discovery.write()?;
+                address_discovery.get_next()
+            };
+
+            match next_address {
+                Some(address) => {
+                    let potential_peer = Peer::connect_with_timeout(
+                        address,
+                        Duration::from_secs(1),
+                        Arc::new(Mempool::default()),
+                        self.network,
+                    );
+
+                    if let Ok(peer) = potential_peer {
+                        peer.send(NetworkMessage::GetAddr)?;
+                        self.try_receive_addr(&peer)?;
+                        self.try_receive_addr(&peer)?;
+                        self.sender.send((address, peer.get_version().services))?;
+                        // TODO: Investigate why close is being called on non existent connections
+                        // currently the errors are ignored
+                        peer.close().unwrap_or(());
+                    }
+                }
+                None => continue,
+            }
+        }
+    }
+}
+
+/// A dedicated cache structure, with cbf/non_cbf separation
+///
+/// [AddressCache] will interface with file i/o
+/// And can te turned into seeds. Generation of seed will put previously cached
+/// cbf addresses at front of the vec, to boost up cbf node findings
+#[derive(Serialize, Deserialize)]
+struct AddressCache {
+    banned_peers: HashSet<SocketAddr>,
+    cbf: HashSet<SocketAddr>,
+    non_cbf: HashSet<SocketAddr>,
+}
+
+impl AddressCache {
+    fn empty() -> Self {
+        Self {
+            banned_peers: HashSet::new(),
+            cbf: HashSet::new(),
+            non_cbf: HashSet::new(),
+        }
+    }
+
+    fn from_file(path: &str) -> Result<Option<Self>, AddressManagerError> {
+        let serialized: Result<String, _> = std::fs::read_to_string(path);
+        let serialized = match serialized {
+            Ok(contents) => contents,
+            Err(_) => return Ok(None),
+        };
+
+        let address_cache = serde_json::from_str(&serialized)?;
+        Ok(Some(address_cache))
+    }
+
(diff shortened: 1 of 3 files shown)


rajarshimaitra:
@johncantrell97 thank for working on this. This created a nice base to introduce the rest of the cbf functionalities.

I have been trying over the p2p manager, and have finalized how the address manager would look like. Added some changes, and updated via logically occurring commit. Below is a description of the change and rationales.

### Objective ###
In my mind a `PeerManager` should have a peer directory and an `AddressManager` (this PR). The sole purpose of `AddresssManager` is to handle crawling through the network and fetch valid and working node addresses. 

The Major interface between `AddressManager` and the user is via a `fetch()` method, which takes in number of cbf/non-cbf peers and provide freshly crawled addresses to user. 

Address manager should maintain a cache (file storage, json for now) and should effectively use it to quickly bootstrapped from previously found nodes. Caching should be done at the end of every fetch, and cache should only update if previously unknown address is found.

Address manager should not blindly trust previously found nodes, and should connect to verify they are still alive. For this reason `AddressDirectory` is the live in-memory store of freshly found addresses, and each instance of a manager will start with fresh `AddressDirectory`.

Once it has a populated cache, subsequent fetches should be fast. This is done by converting the existing cache into seeds and putting the previously known cbf addresses at front, so they get tried out first before non-cbf addresses. As the major time is taken for finding cbf peers in the network, this speeds up the process.  

Given the above objective and after few trials below is the `AddressManager` architecture I ended up settling for. Which is mostly similar to @johncantrell97 's initial sketch with some modifications.

image

### Change synopsis ###
Commits are atomic and should be easy to review them independently.

Added documentation.

Removed public visibility of most structures, as they will only be exposed to the manager.

Added a test case to demonstrate how the manager should be used to fetch fresh addresses from the network.

Each time `fetch()` is called, the manager will spawn new workers who will populate the directory, starting with a seed create from already existing cache.

The test includes a timer to compare time taken fetch 5 cbf and 5 non-cbf peers, once from and empty cache, and then from the populated cache of the first run.

In my local it took
- first crawl : 125 secs
- second crawl - 2.5 secs

Once the manager finds some cbf peers and cache them, subsequent fetches are mostly fast.

rajarshimaitra:
I am getting this error, randomly for some peers
```
<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Io(Os { code: 107, kind: NotConnected, message: "Transport endpoint is not connected" })', src/blockchain/compact_filters/p2p.rs:185:38
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/panicking.rs:92:14
   2: core::option::expect_none_failed
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/option.rs:1300:5
   3: core::result::Result<T,E>::unwrap
             at /home/raj/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:1037:23
   4: bdk::blockchain::compact_filters::p2p::AddressWorker::work
             at ./src/blockchain/compact_filters/p2p.rs:185:25
   5: bdk::blockchain::compact_filters::p2p::AddressManager<P>::spawn_workers::{{closure}}
             at ./src/blockchain/compact_filters/p2p.rs:428:17
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

It means `peer.close()` is being called on a peer which is not connected. Not sure why that would be the case. Will dig deeper. Suggestions are most welcome.

src/blockchain/compact_filters/p2p.rs
```diff
@@ -0,0 +1,539 @@
...
+                        address,
+                        Duration::from_secs(3),
+                        Arc::new(Mempool::default()),
+                        self.network,
+                    );
+
+                    if let Ok(peer) = potential_peer {
+                        peer.send(NetworkMessage::GetAddr).unwrap();
+                        self.try_receive_addr(&peer);
+                        self.try_receive_addr(&peer);
```
afilini: Are you supposed to receive two `addr` messages or is this just to give it one more chance to pickup the message before giving up?

src/blockchain/compact_filters/p2p.rs
```diff
@@ -0,0 +1,539 @@
...
+    }
+
+    fn from_file(path: &str) -> Option<Self> {
+        let serialized: Result<String, _> = std::fs::read_to_string(path);
+        let serialized = match serialized {
+            Ok(contents) => contents,
+            Err(_) => return None,
+        };
+
+        let address_cache = serde_json::from_str(&serialized).unwrap();
```
afilini: I would avoid the `unwrap()` here, either change the fn to return a `Result` or just return `None` like you do for the read errors

src/blockchain/compact_filters/p2p.rs
```diff
@@ -0,0 +1,539 @@
...
+        Some(address_cache)
+    }
+
+    fn write_to_file(&self, path: &str) -> Result<(), std::io::Error> {
+        let serialized = serde_json::to_string_pretty(&self).unwrap();
+
+        let mut cache_file = File::create(path).unwrap();
```
afilini: Same here, in a fn that returns a result there shouldn't really be any `unwraps`

src/blockchain/compact_filters/p2p.rs
```diff
@@ -0,0 +1,539 @@
...
+
+/// A dedicated cache structure, with cbf/non_cbf separation
+///
+/// [AddressCache] will interface with file i/o
+/// And can te turned into seeds. Generation of seed will put previously cached
+/// cbf addresses at front of the vec, to boost up cbf node findings
+#[derive(Serialize, Deserialize)]
+struct AddressCache {
+    cbf: Vec<SocketAddr>,
+    non_cbf: Vec<SocketAddr>,
```
afilini: I would change these to `HashSets`, I don't think the ordering of the items it's particularly important. This would have the advantage that loading the same cache multiple times effectively shuffles the items...

src/blockchain/compact_filters/p2p.rs
```diff
@@ -0,0 +1,539 @@
...
+
+        Ok(())
+    }
+
+    fn make_seeds(&self) -> VecDeque<SocketAddr> {
+        self.cbf
+            .iter()
+            .chain(self.non_cbf.iter())
+            .copied()
+            .collect()
```
afilini: ... to avoid using the same few seeds over and over again

rajarshimaitra:
Made a lot of updates to both API and internal structures. The objective is to make address crawling faster.

- Rebased the PR so previous commits also updated. Which doesn't need re-review. Previous iteration's last commit was https://github.com/bitcoindevkit/bdk/pull/378/commits/bfc13587bf1cfc1bc1854d962ea7027a3cce089e
So everything after that is new additions.

- Manager is modified to maintain a specific reserve of fresh addresses. (For now, the default is 5 for both cbf and non_cbf). The directory will keep track of all the previously supplied addresses to the user and will start workers to replenish the fresh buffer.

- The major API of the address manager is `fetch()`, which when called will fetch address from the network until buffer requirement is reached.

- Decided not to go with its own thread. So fetch will block the user thread until it returns. 

- Added banning logic to the manager and other internal structures.

- Major structural changes to make various comparisons and insertion efficient. So far I think I am quite satisfied with the speed of the crawl. the timer tests now return 35 secs and 1.5 secs for fresh and cached fetch respectively. Which I think is satisfactory. There are probably more possible optimization exists. Open for suggestions.

So far I think it finally has the right APIs to work with a PeerManager. Needs much housekeeping work (like Error structures, etc), which will be done once the API is more or less finalized.

src/blockchain/compact_filters/p2p.rs
```diff
@@ -0,0 +1,646 @@
...
+    /// - load up the existing cache.
+    /// - Update the cache with new found peers.
+    /// - check if address is in banlist
+    /// - run crawlers until buffer requirement is matched
+    /// - flush the current cache into disk
+    pub fn fetch(&mut self) {
+        self.spawn_workers();
+
+        // Get already existing cache
+        let mut cache = match AddressCache::from_file(&self.cache_filename) {
```
afilini: It looks like the cache is loaded in the constructor as well, maybe you can just store it in the struct to avoid re-loading it here

src/blockchain/compact_filters/p2p.rs
```diff
@@ -0,0 +1,646 @@
...
+
+    /// Crawl the Bitcoin network until given number of cbf/non_cbf nodes are found
+    ///
+    /// - This will start a bunch of crawlers.
+    /// - load up the existing cache.
+    /// - Update the cache with new found peers.
+    /// - check if address is in banlist
+    /// - run crawlers until buffer requirement is matched
+    /// - flush the current cache into disk
+    pub fn fetch(&mut self) {
```
afilini: Ideally I would change this to return a `Result` and remove the unwraps inside

src/blockchain/compact_filters/p2p.rs
```diff
@@ -0,0 +1,646 @@
...
+                            cache.add_address(addr, cbf);
+                        }
+                    }
+                }
+            }
+            self.progress.update(self.get_progress()).unwrap();
+        }
+
+        // When completed, flush the cache
+        cache.write_to_file(&self.cache_filename).unwrap();
```
afilini: Maybe it makes sense to write the cache periodically and not just at the end?

src/blockchain/compact_filters/p2p.rs
```diff
@@ -0,0 +1,646 @@
...
+    /// Get a new non_cbf address
+    pub fn get_new_non_cbf_address(&mut self) -> SocketAddr {
+        self.directory.get_new_address(false)
+    }
+
+    /// Ban an address
+    pub fn ban_peer(&mut self, addrs: &SocketAddr, cbf: bool) {
+        let mut cache = AddressCache::from_file(&self.cache_filename).unwrap();
+
+        cache.add_to_banlist(*addrs, cbf);
```
afilini: I think this is not being written back to disk. Adding the cache as a struct member would allow to reuse it here as well

src/blockchain/compact_filters/p2p.rs
```diff
@@ -0,0 +1,646 @@
...
+                Some(address) => {
+                    let potential_peer = Peer::connect_with_timeout(
+                        address,
+                        Duration::from_secs(1),
+                        Arc::new(Mempool::default()),
+                        self.network,
+                    );
+
+                    if let Ok(peer) = potential_peer {
+                        peer.send(NetworkMessage::GetAddr).unwrap();
```
afilini: I would remove the unwrap here and right down below, the `send` can fail if the underlying socket disconnects

src/blockchain/compact_filters/p2p.rs
```diff
@@ -0,0 +1,646 @@
...
+                                address,
+                                peer.get_version().services,
+                            ))
+                            .unwrap();
+                        // TODO: Investigate why close is being called on non existent connections
+                        // currently the errors are ignored
+                        peer.close().unwrap_or(());
+                    }
+                }
+                None => continue,
```
afilini: Maybe if we can't connect we should "ban" the peer?

I wonder if it makes sense to add a "ban until" timestamp to the cache, so that we can differentiate between a permanent ban (peer misbehaving) or a temporary "i can't connect right now" ban

src/blockchain/compact_filters/p2p.rs
rajarshimaitra: I guess it was to give it one more chance. I removed it in the last changes. But I am not fully sure that it doesn't have any impact.

I am having trouble syncing from the existing seed. New addresses are not updating in the pending list. But it works with a previously populated cache.

It could be related to this. Will investigate more.

src/blockchain/compact_filters/p2p.rs
rajarshimaitra: I am replacing unwraps from all the places, and creating error structure for `Peer`, `AddressManager`, `PeerManager` etc and then combined them into `CompactFiltersErrror`. So will handle all the unwraps in that following work.

We should have proper unwrap handling everywhere.

src/blockchain/compact_filters/p2p.rs
rajarshimaitra: https://github.com/bitcoindevkit/bdk/pull/378#discussion_r660346200

src/blockchain/compact_filters/p2p.rs
rajarshimaitra: Yup cache is now updated to `HashSet`.

src/blockchain/compact_filters/p2p.rs
rajarshimaitra: I am not sure I understand. 

What we are doing here is ensuring the CBF addresses are in front of the seed vector. In that way, those are the addresses the worker will try first, even before existing seeds.

Because finding cbf peers are much slower than finding regular peers, the discovery speeds up if we try already known cbf addresses first.

src/blockchain/compact_filters/p2p.rs
rajarshimaitra: Hmm.. I was trying to reduce the memory load of the structure. `cache` is initiated at the constructor and `fetch()`.

In the constructor, it's created to feed the seeds to the workers.

In `fetch`, it's created to update the disk storage.

In my head, I think these will be called very few times.

constructor: Once
fetch: Only when available fresh address reserve depletes under a threshold.

For this reason, I figured it might not be clever to carry a large cache around with the address manager. And only create when needed.

rajarshimaitra: Also note that the cache is essentially duplicated inside `Directory` which is part of the address manager. So thought it probably doesn't make sense to carry around duplicate data, only to use them occasionally.

src/blockchain/compact_filters/p2p.rs
rajarshimaitra: Ideally, all `pub` functions should give back `Result`. Working on it.

src/blockchain/compact_filters/p2p.rs
rajarshimaitra: Yes it does. But that involves threading. I thought of that first in general as it made more sense, but having multiple threads made stuff difficult. I already went into deadlock scenarios with just worker threads. So decided to keep stuff simple and ended up writing most of the things single-threaded and optimized around executions.  

As mentioned in https://github.com/bitcoindevkit/bdk/pull/378#discussion_r660353392 we can expect `fetch` to be called occasionally. As a result, cache will also update occasionally, and we will know exactly when it updates, only in a `fetch` call. 

Thus it made sense to not give its own dedicated loop and just use file i/o only when needed.

src/blockchain/compact_filters/p2p.rs
rajarshimaitra: Ah right, it wasn't being written to disk. Fixed that. 

Because of the reasons explained here https://github.com/bitcoindevkit/bdk/pull/378#discussion_r660353392 I am inclining against adding the cache to the structure.

src/blockchain/compact_filters/p2p.rs
rajarshimaitra: Agreed, will be fixed in error handling commits.

src/blockchain/compact_filters/p2p.rs
rajarshimaitra: Hmm, probably. Currently what it does is if it can't connect it will simply skip. I am not sure if connection failure should result into ban, as nodes can be offline temporarily, they may connect in next try. 

I think we definitely need to define banning more granularly, and I think best place to do that will be the `peermanager` (which I am sketching now, and it will use address manager inside). I have put a ban score flag into PeerManager, so we can have more granular banning logic like core does. 

And the objective here is the cache will maintain only the final ban list such that once it's banned it will never get into discovery again.

And then in peer manager, we can design different levels of banning, and decide whether to tell the address manager to "cache ban" the address or not.

rajarshimaitra:
Thanks @afilini for the review. Tried to address all of them. Overall the intention was to not involve threads and try to optimize around function calls and executions. That guided lot of the decisions. I think I have got it into a satisfactory level of efficiency that with an existing cache an address DB can be populated in few seconds with freshly connected cbf and non_cbf nodes. 

There's probably more optimizations possible.   

Following are the next TODOs after which I think I can get it out of draft.

- Handle results and options everywhere. This probably will also involve changes in `Peer` module as I think it should have its dedicated errors, and that decision will affect address manager errors too.
- Few minor possible changes in the `AddressManager` API as I sketch the basics of `PeerManager`.
- Few more tests (looking for ideas).

src/blockchain/compact_filters/p2p.rs
afilini: I was thinking that if the peers are stored in `Vec`s every call to this method would return the same few peers in the same order, unless they are explicitly removed from the cache.

If the caller uses a logic like "as soon as I reach 5 valid peers I'll stop looking" returning the seeds in the same order could lead to the caller just re-using the same peers over and over again, even across shutdowns.

Using `HashSet`s somewhat mitigates this I think.

rajarshimaitra: Ya, once we have `HashSet` the ordering is jumbled up. 

But it's still true that it's the same peers from the previous fetch that were cached in the last run.

I think we cannot do away with that repetition (that's the point of a cache anyway), So the manager checks the connection again, see if the peer is alive, see if it's in the ban list, check weather cbf/non_cbf, and then pass it on to the caller.

So at the end of all of it, we can be sure that the manager didn't just hand over a dead peer just because it had it in the cache last time.

rajarshimaitra:
Updated with Error handling. 

A new `PeerError` structure added.

All the other internal module-level errors will be finally combined into `CompactFiltersError`.

rajarshimaitra:
Updated the commit history consolidating all the changes.

Address Manager seems ready from my side. All the APIs I think are there required to make a basic Peer Manager.

This is now ready for public review.

src/blockchain/compact_filters/p2p.rs
johncantrell97: A peer can respond to get addresses in multiple replies and the current peer code ever only expects one reply.  We should probably at some point refactor some of the peer communication layer to be a bit more dynamic but until then I thought it might be useful to poll twice for addr replies.  

I think I logged it at some point and did see it receiving addresses on the second try so it might be helpful.

src/blockchain/compact_filters/p2p.rs
johncantrell97: At one point I had it writing the cache to file inside of the fetch() message loop from workers anytime I updated the memory cache.  I ran into a 'too many open files' issue though and so moved it out.  Could probably get away with some kind or probabilistic flush if we want to update it more often but not every iteration of the loop.

rajarshimaitra: Yes, That's the reason I took it out of the loop. So currently it's not flushing at each iteration. It flushes after the loop completes. So that problem doesn't exist now.

https://github.com/bitcoindevkit/bdk/blob/23fe55127e0a4c6871df0ff66ce8897aeff78143/src/blockchain/compact_filters/address_manager.rs#L541-L542

Also, note that the loop will not start in case the buffer requirement is already fulfilled. Flush will happen only once per `fetch()` or `ban()` call. Because these are the only calls where we expect the cache to be updated.

src/blockchain/compact_filters/p2p.rs
rajarshimaitra: Yes makes sense. At one point I have seen (without the second try_receive) that new pendings are not being updated in the discovery. Or it's getting updated with just 1 new address. As a result, the discovery pending list is being quickly exhausted and the workers get stuck.

I changed it back to double try and it didn't happen after that. But in any case, we should handle that situation properly. What to do when the pending list depletes in the discovery and workers don't have any more addresses to try. This can happen for multiple other reasons too, ex: A worker got new addresses and waiting for the write lock. In the meantime, other workers deplete the pending list and then nobody has any new addresses to try and they all wait.

To mitigate part of this I changed the pending addition task from the manager to worker. So instead of one manager thread trying to add pending to discovery it's all the worker threads and hopefully, someone will get to add new pending before the list depletes.

It's still nasty if that happens. There is no run time error, everything just pauses, and the user doesn't have any clue what went wrong.

The above refactoring removed the need for `AddressMessages` as it's just only one type of message the worker will send to the manager.

rajarshimaitra:
Rebased

notmandatory:
I had to cancel the CI workflows for this branch because the `compact_filters` tests seems to be hanging. Could this be some thread locking issue? @rajarshimaitra have you seen this happen on your local testing?

rajarshimaitra:
Hmm.. It seems the test got stuck somewhere. Which is not happening in my local.

I have combined the two tests into a single one, running two tests parallelly can provide weird errors. 
Added extra log points to see what step got stuck if it happens again.

There was a hang issue as I mentioned in https://github.com/bitcoindevkit/bdk/pull/378#discussion_r660344444, which after some refactoring didn't occur anymore in my local. It could be this, but unlikely. Maybe others can confirm from their locals. If it happens then one will see the pending number goes to zero in the test log, and the test hangs.

It seems to be working fine in my local and here's what I see.

```
> Executing task: cargo test --package bdk --lib -- blockchain::compact_filters::address_manager::test --nocapture <

   Compiling bdk v0.8.1-dev (/home/raj/github-repo/bdk)
    Finished test [unoptimized + debuginfo] target(s) in 9.25s
     Running target/debug/deps/bdk-fdc911b1c7b1070c

running 1 test
Starting manager and initial fetch
P2P Discovery: 4382 queued, 688 visited, 221 connected, 5 cbf_enabled
Completed Initial fetch
test blockchain::compact_filters::address_manager::test::test_address_manager ... test blockchain::compact_filters::address_manager::test::test_address_manager has been running for over 60 seconds
Starting new fetch with previous cache
P2P Discovery: 1359 queued, 40 visited, 20 connected, 5 cbf_enabled
Completed new fetch()
Time taken for initial crawl: 44.709318174s
Time taken for next crawl 2.871860807s
Checking buffer management
P2P Discovery: 9808 queued, 2916 visited, 465 connected, 10 cbf_enabled
Buffer management passed
test blockchain::compact_filters::address_manager::test::test_address_manager ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 203 filtered out; finished in 157.53s  
```

The test takes 157.53 secs. Which seems much lower than github action would mind. But it might take a bit more in the CI server. If that's the issue, then we should ignore this test for now until we figure out how to make it faster.

rajarshimaitra:
I am not sure why this test https://github.com/bitcoindevkit/bdk/pull/378/checks?check_run_id=3034931064 is stuck while all others are completed.

I am out of suggestions here. If the high test time was a problem, github would have complained in other tests too. Not really sure whats the problem. It seems it got cancelled before the address manager could initialise. The test will try to create a cache file in run directory also, but I suppose that should not be an issue.

Probably we should just ignore this test for now?

notmandatory:
I had to cancel the "Code Coverage" workflow, it was taking +2hrs for some reason. It seemed to be stalled on this: 
```
test blockchain::compact_filters::address_manager::test::test_address_manager has been running for over 60 seconds
```

I think it's OK to disable this test for now and then we can see if we can construct some focused regtest tests with https://github.com/RCasatta/bitcoind.

rajarshimaitra:
@notmandatory noted. 

We can disable the test and merge, or we can fix this here itself. 

I am working on creating a controlled regtest to do the test. Maybe in few more days. I think its better to get this thing done now, than come back later. 

We can merge this after that.

notmandatory:
> @notmandatory noted.
> 
> We can disable the test and merge, or we can fix this here itself.

After the long discussion on Discord I agree best to disable the `test_address_manager` test for now so we can merge this and have all the code together in the `compact_filters` branch.

notmandatory (approved):
ACK, merging to `compact_filters` branch.

notmandatory:
Looks like we lost the `#[ignore]` on the `test_address_manager()` test. I pushed it back up and if all the rest of the tests pass I should be able to do the merge.

PR #382, opened 2021-06-24
Add `InitWallet` structure

### LINKED ISSUE #354: New Wallet should force a sync call or offline mode
At the moment doing:

```
let wallet = Wallet::new(...)
let address = wallet.get_address();
wallet.sync()
```

may cause trouble because the address batch is created at get_address time so at sync time `Blockchain::setup` is not called

Moreover, during RPC backend development I realized if you do 

```
let wallet = Wallet::new(...)
let address = wallet.get_address();
send_to_address(&address, 50_000);
wallet.sync()
assert_eq!(wallet.get_balance(), 50_000);   //FAILS
```

the assert fails because `importmulti` is done at sync time when the received tx is already in the mempool

A possible solution could be that `Wallet::new` returns `WalletNotInitialized` (or a better name) which expose only `sync()` or `offline()` which returns a proper wallet or explicitly opt-in in an offline one

tcharding:
### Description

Currently when creating a `Wallet` there is a hard requirement that users call `wallet.sync()` before using the wallet.

In an effort to make the bdk library more intuitive add a new type `InitWallet` that wraps the `Wallet` and exposes a only three simple methods. A constructor, a `sync` method, and a `init` method that is a helper method calling `sync` for simple usage. `sync` consumes `self` and returns the inner `Wallet`.

Idea based on suggestion by @RCasatta in issue #354.

Fixes #354

### Notes to the reviewers

Draft because this is not necessarily a good idea, its a breaking API change, and it makes a lot of current documentation stale. This PR does not include all the docs fixes, would be done before raising non-draft PR. Are there many docs online in other places that this is going to make stale? Is that a consideration?

Also, adds an example usage in `bdk/examples/init_wallet.rs`. Done as two commits, the first adding the example with the current `Wallet::new` workflow and then, when we add the `InitWallet` in patch 2, includes modifications to the example.

I didn't think that hard about whether or not this is a good change, it just seemed like a fun thing to hack on late one evening, feel free to nack it :)

#### All Submissions:

#### New Features:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,6 +67,7 @@ test-esplora = ["esplora"]
 test-md-docs = ["electrum"]
 
 [dev-dependencies]
+anyhow = "1.0"
 lazy_static = "1.4"
 env_logger = "0.7"
 clap = "2.33"
--- a/examples/init_wallet.rs
+++ b/examples/init_wallet.rs
@@ -11,10 +11,10 @@
 use anyhow::Result;
 use bitcoin::Network;
 
-use bdk::blockchain::{ElectrumBlockchain, NoopProgress};
+use bdk::blockchain::ElectrumBlockchain;
 use bdk::database::MemoryDatabase;
 use bdk::electrum_client::Client;
-use bdk::Wallet;
+use bdk::InitWallet;
 
 const ELECTRUM_URL: &str = "ssl://electrum.blockstream.info:60002";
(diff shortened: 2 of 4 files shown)


RCasatta:
The problem of this API is that I may want to operate offline, for example just generating an address or with previously cached data.

Another implementation approach could be to have a phantom marker in the Wallet structure, which is `Offline` after `new`, and become `Online` after the `sync`
Some methods are always available (like `get_new_address` (note https://github.com/bitcoindevkit/bdk/issues/354#issuecomment-868320424) but others (like `get_balance` are available only for the Online version (Does not solve the "I want to operate on cached data" use case)

tcharding:
> The problem of this API is that I may want to operate offline, for example just generating an address or with previously cached data.

Any reason not to use `Wallet::new_offline` in that case?

tcharding:
I had a bit more of a fiddle with trying to make the new API better as suggested but ran into problems because of the type annotations on online/offline wallet `()` vs `B`.

Dropping this for now.

PR #386, opened 2021-07-08
Add Vbytes trait

tcharding:
### Description

We convert weight units into vbytes in various places. Lets add a trait to do it, this makes the code slightly cleaner.

#### All Submissions:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -780,7 +780,7 @@ where
             return Err(Error::IrreplaceableTransaction);
         }
 
-        let vbytes = tx.get_weight() as f32 / 4.0;
+        let vbytes = tx.get_weight().vbytes();
         let feerate = details.fee.ok_or(Error::FeeRateUnavailable)? as f32 / vbytes;
 
         // remove the inputs from the tx and process them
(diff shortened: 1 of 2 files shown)


rajarshimaitra:
ACK. 

I think it's useful.

afilini (approved):

src/wallet/mod.rs
```diff
@@ -1578,6 +1578,18 @@ where
...
 
+/// Trait implemented by types that can be used to measure weight units.
+pub trait Vbytes {
+    /// Convert weight units to virtual bytes.
+    fn vbytes(self) -> f32;
+}
+
+impl Vbytes for usize {
+    fn vbytes(self) -> f32 {
+        self as f32 / 4.0
```
harding, line 1589: Shouldn't this be rounded up?[1]  (Or is that being done somewhere else?  I did a quick git grep, but maybe I missed it.)

[1] "Virtual transaction size is defined as Transaction weight / 4 (rounded up to the next integer)" -- https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#transaction-size-calculations .  Software that has failed to round up has been known to occasionally generate unrelayable transactions when sending at the default minimum relay feerate, e.g. https://github.com/ElementsProject/lightning/commit/2e687b9b352c9092b5e8bd4a688916ac50b44af0

tcharding: Thanks for pointing this out. Looks like typically we divide a fee by the value returned by `vbytes` to get a fee rate, so this would mean currently without `ceil()` the fee rate is actually too high. 

But we have at least one call site that multiplies by vbytes and based on the docs you link our code is wrong. Thanks!

```
let fee =
            (TXIN_BASE_WEIGHT + weighted_utxo.satisfaction_weight).vbytes() * fee_rate.as_sat_vb();
        let effective_value = weighted_utxo.utxo.txout().value as i64 - fee.ceil() as i64;
```

tcharding: Attempted fix in: https://github.com/bitcoindevkit/bdk/pull/398

PR #389, opened 2021-07-09
[rpc] Show in the docs that the RPC APIs are feature-gated

afilini:
### Description

Show in the docs that in order to use the API under the `blockchain::rpc` module the `rpc` feature has to be enabled.

Not sure if it's worth doing a patch-level release just to fix this, I don't think it's that big of an issue.

#### All Submissions:


notmandatory (approved):
utACK. It's not a bug fix so I think this can wait for the next normal release.

PR #392, opened 2021-07-12
Return early if required UTXOs already big enough

tcharding:
### Description

If the required UTXO set is already bigger (including fees) than the
amount required for the transaction we can return early, no need to go
through the BNB algorithm or random selection.

### Notes to the reviewers

This is an optimisation, probably wants careful review since I only just started reading this code :)

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -345,6 +345,14 @@ impl<D: Database> CoinSelectionAlgorithm<D> for BranchAndBoundCoinSelection {
             .try_into()
             .expect("Bitcoin amount to fit into i64");
 
+        if curr_value > actual_target {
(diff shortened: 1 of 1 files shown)


afilini:
I can't think of anything that would break because of this, but just to be extra careful: could you also add a test? If anything just to make sure our test coverage also goes through that branch.

tcharding:
> I can't think of anything that would break because of this, but just to be extra careful: could you also add a test? If anything just to make sure our test coverage also goes through that branch.

Sure thing, can do. My bad, should have done this already.

tcharding:
Turns out this code path is already tested by the existing test `test_bnb_coin_selection_required_are_enough`, I've added an additional test `test_bnb_coin_selection_required_not_enough` to test for falsely entering the new code path.

afilini (approved):

PR #398, opened 2021-07-16
Fix vbytes and fee rate code

tcharding:
### Description

It was just pointed out that we are calculating the virtual bytes
incorrectly by forgetting to take the ceiling after division by 4 [1]

Add helper functions to encapsulate all weight unit -> virtual byte
calculations including fee to and from fee rate. This makes the code
easier to read, easier to write, and gives us a better chance that bugs
like this will be easier to see.

As an added bonus we can also stop using f32 values for fee amount,
which is by definition an amount in sats so should be a u64. This
removes a bunch of casts and the need for epsilon comparisons and just
deep down feels nice :)

[1] https://github.com/bitcoindevkit/bdk/pull/386#discussion_r670882678

Thanks to David Harding for finding this bug for us. (Solid surname too :)

#### All Submissions:

#### New Features:

#### Bugfixes:

Do I need these?

### DIFF
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -26,7 +26,7 @@
 //! ```
 //! # use std::str::FromStr;
 //! # use bitcoin::*;
-//! # use bdk::wallet::coin_selection::*;
+//! # use bdk::wallet::{self, coin_selection::*};
 //! # use bdk::database::Database;
 //! # use bdk::*;
 //! # const TXIN_BASE_WEIGHT: usize = (32 + 4 + 4 + 1) * 4;
@@ -41,7 +41,7 @@
 //!         optional_utxos: Vec<WeightedUtxo>,
 //!         fee_rate: FeeRate,
 //!         amount_needed: u64,
-//!         fee_amount: f32,
+//!         fee_amount: u64,
 //!     ) -> Result<CoinSelectionResult, bdk::Error> {
 //!         let mut selected_amount = 0;
 //!         let mut additional_weight = 0;
@@ -57,9 +57,8 @@
 //!                 },
 //!             )
 //!             .collect::<Vec<_>>();
-//!         let additional_fees = additional_weight as f32 * fee_rate.as_sat_vb() / 4.0;
-//!         let amount_needed_with_fees =
-//!             (fee_amount + additional_fees).ceil() as u64 + amount_needed;
+//!         let additional_fees = fee_rate.fee_wu(additional_weight);
+//!         let amount_needed_with_fees = (fee_amount + additional_fees) + amount_needed;
 //!         if amount_needed_with_fees > selected_amount {
 //!             return Err(bdk::Error::InsufficientFunds {
 //!                 needed: amount_needed_with_fees,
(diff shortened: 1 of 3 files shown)


src/wallet/mod.rs
```diff
@@ -1581,15 +1581,37 @@ where
...
+/// Calculate absolute fee in Satoshis from fee rate and weight unit size.
+pub fn fee_wu(fee_rate: FeeRate, wu: usize) -> u64 {
+    fee_vbytes(fee_rate, wu.vbytes())
+}
+
+/// Calculate absolute fee in Satoshis from fee rate and virtual byte size.
+pub fn fee_vbytes(fee_rate: FeeRate, vbytes: usize) -> u64 {
+    (fee_rate.as_sat_vb() * vbytes as f32).ceil() as u64
+}
+
```
afilini: Is there a reason why these functions aren't constructors implemented on `FeeRate` directly?

tcharding: None at all now you mention it. Will play around with this idea.

tcharding: Much nicer, thanks! I shortened the names also in line with the other `FeeRate` methods.

afilini (approved):

src/wallet/coin_selection.rs
```diff
@@ -57,9 +57,8 @@
 //!                 },
 //!             )
 //!             .collect::<Vec<_>>();
-//!         let additional_fees = additional_weight as f32 * fee_rate.as_sat_vb() / 4.0;
-//!         let amount_needed_with_fees =
-//!             (fee_amount + additional_fees).ceil() as u64 + amount_needed;
+//!         let additional_fees = wallet::fee_wu(fee_rate, additional_weight);
```
afilini: Sorry, I only noticed this after approving the changes. Shouldn't this be changed as well to use the new methods?

tcharding: Oh boy, seems I still don't know how to run cargo in a way that catches everything. What's your workflow, what cargo command do you run to check your work?

tcharding: I've squashed these two commits into a single one since the first patch was really a step in the development process (after implementing your suggestions @afilini). Since it may not be fresh in mind when merging, and/or others are reviewing, I think it is less review work to have a single patch.

tcharding: Apart from fixing the document code call site, no other changes were made while squashing. The patch is reasonably short now so its not too hard to verify the patch is ok again. Please re-review. Also, please say if this squashing is not what you like.

tcharding: Please review carefully the two changes that look like:
```
-        fee_amount += calc_fee_bytes(serialize(&drain_output).len() * 4);
+        fee_amount += fee_rate.fee_vb(serialize(&drain_output).len());
```
Thanks.

afilini: > What's your workflow, what cargo command do you run to check your work?

I'm not sure what went wrong here, usually just running `cargo test` or `cargo test --doc` should have caught it. Weird.

afilini (approved):

PR #400, opened 2021-07-19
[keys] limit version of zeroize to support rust 1.47+

### LINKED ISSUE #399: Switch to a different BIP39 implementation
The `tiny-bip39` dependency used in the `all-keys` feature depends on `zeroize` which appears to only support MSRV 1.51+

https://github.com/iqlusioninc/crates/

This is causing the following error in our CI workflow:
```
Downloaded zeroize v1.4.0
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/zeroize-1.4.0/Cargo.toml`

Caused by:
  feature `resolver` is required

  this Cargo does not support nightly features, but if you
  switch to nightly channel you can add
  `cargo-features = ["resolver"]` to enable this feature
```

https://github.com/bitcoindevkit/bdk/runs/3101954688?check_suite_focus=true

notmandatory:
### Description

This will fix #399 by limiting the transitive dependency zeroize to < 1.4.0 since zeroize >= 1.4.0 has a MSRV of `1.51+`. 

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ cc = { version = ">=1.0.64", optional = true }
 socks = { version = "0.3", optional = true }
 lazy_static = { version = "1.4", optional = true }
 tiny-bip39 = { version = "^0.8", optional = true }
+zeroize = { version = "<1.4.0", optional = true }
 bitcoinconsensus = { version = "0.19.0-3", optional = true }
 
 # Needed by bdk_blockchain_tests macro
@@ -56,7 +57,7 @@ compact_filters = ["rocksdb", "socks", "lazy_static", "cc"]
 key-value-db = ["sled"]
 async-interface = ["async-trait"]
 all-keys = ["keys-bip39"]
-keys-bip39 = ["tiny-bip39"]
+keys-bip39 = ["tiny-bip39", "zeroize"]
 rpc = ["bitcoincore-rpc"]
 
 


Cargo.toml
```diff
@@ -56,7 +57,7 @@ compact_filters = ["rocksdb", "socks", "lazy_static", "cc"]
 key-value-db = ["sled"]
 async-interface = ["async-trait"]
 all-keys = ["keys-bip39"]
-keys-bip39 = ["tiny-bip39"]
+keys-bip39 = ["tiny-bip39", "zeroize"]
```
rajarshimaitra, line 60: it seems `all-keys` and `keys-bip39` are duplicated feature. Is this intentional?

rajarshimaitra (commented):
Checked build locally. Seems good. 

One feature flag comment.

Cargo.toml
notmandatory: I think the idea is we eventually want to get away from using tiny-bip39, so want to use a feature name that won't change.

notmandatory: But for all-keys I think it's just  to be prepared if we want to have a separate optional key subfeatures or the all-keys feature.

rajarshimaitra: Makes Sense. Thanks for clarifying.

rajarshimaitra (approved):
ReACK.

sandipndev (approved):
ACK, limiting `zeroize` to `<1.4.0` should fix this.

afilini (approved):

Issue #396, opened 2021-07-15
Add support for sending to bech32m addresses


notmandatory:
Since `rust-bitcoin` [0.26.2](https://github.com/rust-bitcoin/rust-bitcoin/compare/0.26.2...master) supports sending to bech32m addresses `bdk` should also be able to support this feature. This issue may only require adding tests to verify support.

https://github.com/rust-bitcoin/rust-bitcoin/pull/601

notmandatory:
Closed with #402

Issue #63, opened 2020-07-27
Taproot / Schnorr


notmandatory:
This is a tracking issue for other issues and PRs related to supporting Taproot in BDK.

## Dependencies

[Overall rust-bitcoin org tracking project](https://github.com/orgs/rust-bitcoin/projects/3#card-53894767)

### rust-bitcoin/rust-bitcoin 

- [x] https://github.com/rust-bitcoin/rust-bitcoin/issues/503

### rust-bitcoin/rust-miniscript 

- [x] https://github.com/rust-bitcoin/rust-miniscript/pull/267
- [x] https://github.com/rust-bitcoin/rust-miniscript/pull/275
- [x] https://github.com/rust-bitcoin/rust-miniscript/pull/278
- [ ] https://github.com/rust-bitcoin/rust-miniscript/issues/182

## Send to Single Signature TR Address

- [x] bitcoindevkit/bdk#402  
- [ ] bitcoindevkit/bdk_wallet#189

## Create Single Signature TR Descriptor Wallet

- [ ]  bitcoindevkit/bdk#463

jrawsthorne:
Upstream issue to track changes needed for taproot https://github.com/rust-bitcoin/rust-bitcoin/issues/503

PR #402, opened 2021-07-20
[test] add tests to verify support for sending to bech32m address

sandipndev:
### Description

This PR is in reference to #396 
It adds some tests to confirm sending transactions to Bech32m addresses, introduced in [BIP350](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki)

**Note:** This change requires `bitcoin-core` `v22.0` to be incorporated into bitcoind.

### Checklists

#### New Features:

* [x] I've added tests for Bech32m adoption by `rust-bitcoin`

### DIFF
--- a/src/testutils/blockchain_tests.rs
+++ b/src/testutils/blockchain_tests.rs
@@ -145,9 +145,7 @@ impl TestClient {
 
         let bumped: serde_json::Value = self.call("bumpfee", &[txid.to_string().into()]).unwrap();
         let new_txid = Txid::from_str(&bumped["txid"].as_str().unwrap().to_string()).unwrap();
-
-        let monitor_script =
-            tx.vout[0].script_pub_key.addresses.as_ref().unwrap()[0].script_pubkey();
+        let monitor_script = Script::from_hex(&mut tx.vout[0].script_pub_key.hex.to_hex()).unwrap();
         self.wait_for_tx(new_txid, &monitor_script);
 
         debug!("Bumped {}, new txid {}", txid, new_txid);
@@ -910,6 +908,53 @@ macro_rules! bdk_blockchain_tests {
 
                 wallet.sync(noop_progress(), None).unwrap();
                 assert!(wallet.get_balance().unwrap() > 0, "incorrect balance after receiving coinbase");
+            }
+
+            #[test]
+            fn test_send_to_bech32m_addr() {
+                use std::str::FromStr;
+                use core_rpc::RpcApi;
+                use core_rpc::core_rpc_json::ImportDescriptorRequest;
+
+                let (wallet, descriptors, mut test_client) = init_single_sig();
+
+                let wallet_descriptor = "tr(tprv8ZgxMBicQKsPdBtxmEMPnNq58KGusNAimQirKFHqX2yk2D8q1v6pNLiKYVAdzDHy2w3vF4chuGfMvNtzsbTTLVXBcdkCA1rje1JG6oksWv8/86h/1h/0h/0/*)#y283ssmn";
+                let change_descriptor = "tr(tprv8ZgxMBicQKsPdBtxmEMPnNq58KGusNAimQirKFHqX2yk2D8q1v6pNLiKYVAdzDHy2w3vF4chuGfMvNtzsbTTLVXBcdkCA1rje1JG6oksWv8/86h/1h/0h/1/*)#47zsd9tt";
+
+                test_client.bitcoind.client
+                    .import_descriptors(
+                        vec![
+                            ImportDescriptorRequest::new(wallet_descriptor, false),
+                            ImportDescriptorRequest::new(change_descriptor, false),
+                        ]
+                    ).unwrap();
+
+                let node_addr = test_client.get_node_address(Some(core_rpc::core_rpc_json::AddressType::Bech32m));
+                assert_eq!(node_addr, bitcoin::Address::from_str("bcrt1pj5y3f0fu4y7g98k4v63j9n0xvj3lmln0cpwhsjzknm6nt0hr0q7qnzwsy9").unwrap());
+
+                let new_bech32m_addr = test_client.get_new_address(Some("test bech32m"), Some(core_rpc::core_rpc_json::AddressType::Bech32m)).unwrap();
+                assert_eq!(new_bech32m_addr, bitcoin::Address::from_str("bcrt1pxa4h86c5gc8x65un8nz546wy7hqxv7wljrv5sxukayh3xwnw23fs80jdf9").unwrap());
+
+                test_client.receive(testutils! {
+                    @tx ( (@external descriptors, 0) => 50_000 )
+                });
+
+                wallet.sync(noop_progress(), None).unwrap();
+                assert_eq!(wallet.get_balance().unwrap(), 50_000, "incorrect balance");
+
+                let mut builder = wallet.build_tx();
+                builder.add_recipient(node_addr.script_pubkey(), 25_000);
+                let (mut psbt, details) = builder.finish().unwrap();
+                let finalized = wallet.sign(&mut psbt, Default::default()).unwrap();
+                assert!(finalized, "Cannot finalize transaction");
+                let tx = psbt.extract_tx();
+                println!("{}", bitcoin::consensus::encode::serialize_hex(&tx));
+                wallet.broadcast(tx).unwrap();
+                wallet.sync(noop_progress(), None).unwrap();
+                assert_eq!(wallet.get_balance().unwrap(), details.received, "incorrect balance after send");
+
+                assert_eq!(wallet.list_transactions(false).unwrap().len(), 2, "incorrect number of txs");
+                assert_eq!(wallet.list_unspent().unwrap().len(), 1, "incorrect number of unspents");
             }
         }
     };
--- a/src/testutils/blockchain_tests.rs
+++ b/src/testutils/blockchain_tests.rs
@@ -913,48 +913,97 @@ macro_rules! bdk_blockchain_tests {
             #[test]
             fn test_send_to_bech32m_addr() {
                 use std::str::FromStr;
-                use core_rpc::RpcApi;
-                use core_rpc::core_rpc_json::ImportDescriptorRequest;
+                use serde;
+                use serde_json;
+                use serde::Serialize;
+                use bitcoincore_rpc::jsonrpc::serde_json::Value;
+                use bitcoincore_rpc::{Auth, Client, RpcApi};
 
                 let (wallet, descriptors, mut test_client) = init_single_sig();
+
+                // TODO remove once rust-bitcoincore-rpc with PR 199 released
+                // https://github.com/rust-bitcoin/rust-bitcoincore-rpc/pull/199
+                /// Import Descriptor Request
+                #[derive(Serialize, Clone, PartialEq, Eq, Debug)]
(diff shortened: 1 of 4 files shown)


RCasatta:
Ideally, I think it's a test to be added (also) to the `bdk_blockchain_tests!` macro

notmandatory:
Looks like bech32m addresses are not supported yet in `rust-bitcoin` 0.26.2. The PR that added support is https://github.com/rust-bitcoin/rust-bitcoin/pull/601 and it has other breaking changes that will also need to be fixed in `bdk`.  You can see the errors by updating `bdk`'s Cargo.toml like this:
```toml
bitcoin = { git = "https://github.com/rust-bitcoin/rust-bitcoin.git", branch = "master", features = ["use-serde", "base64"] }
```

Also @RCasatta has a good idea about also updating the `bdk_blockchain_tests` macro, so this could could be a bit larger task but also a good opportunity to learn about macros.  You'll also be able to get to know other parts of bdk by fixing the other errors to make `bdk` work with the `master` branch of `rust-bitcoin` (until a new released version is ready). Want to give it a go?

sandipndev:
Thank you for going through the problem.

We can develop using the master branch but should we release the Bech32m adoption before `rust-bitcoin` officially releases a version upgrade with the same?

In most places, there is a similar error after changing `Cargo.toml` to use the master branch of `rust-bitcoin`.
```bash
the trait `InnerXKey` is not implemented for `bitcoin::util::bip32`
```

I'm on fixing it!

notmandatory:
> We can develop using the master branch but should we release the Bech32m adoption before `rust-bitcoin` officially releases a version upgrade with the same?

You will need to keep this PR in "draft" mode until an official `rust-bitcoin` release is ready, and then you can put that version in the `Cargo.toml`. In the mean time you can still work on getting ready for that release by making your fixes against the `master` branch.

notmandatory:
The [0.27.0 version of `rust-bitcoin`](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/CHANGELOG.md#027---2020-07-21) is now released which includes `bech32m` support. I was able to get your new test to pass after these steps:

  1. In `Cargo.toml` update `bitcoin` version to `"^0.27"`
  2. Fork and make local clones of the `rust-miniscript` and `rust-electrum-client` projects and update both also to  the `rust-bitcoin` version `0.27`
  3. In `Cargo.toml` update the dependencies there for `rust-miniscript` and `rust-electrum-client` to the local path of your updated clones of these project, something like this:
     ```
     miniscript = { path = "../rust-miniscript" }
     ...
     electrum-client = { path = "../rust-electrum-client", optional = true }
     ```
     
Give it a try and then we can discuss additional tests (as @RCasatta suggested) and how to make the updates to those other projects (if no one else is already working on it).

sandipndev:
For the time being, I changed the dependencies to point to my git fork with updated `rust-bitcoin v0.27` in order to check if all the test cases pass. 

The added test case for `bech32m` is passing with `cargo test`.

But it looks like builds and some other test cases are still failing.

RCasatta:
From a quick look, it seems you need to use patch in Cargo.toml for the rust-bitcoin 0.27 dep so that dependencies of dependencies are replaced as well, see https://doc.rust-lang.org/edition-guide/rust-2018/cargo-and-crates-io/replacing-dependencies-with-patch.html. 

Otherwise, you may end up for something that seems the same Object but they are from different crate versions

sandipndev:
Turns out the error is coming from dev-dependency `electrs`, which uses `bitcoind v0.13`

```txt
   Compiling bitcoind v0.13.0
error[E0308]: mismatched types
   --> /home/sandi/.cargo/registry/src/github.com-1ecc6299db9ec823/bitcoind-0.13.0/src/lib.rs:184:45
    |
184 |             let client_result = Client::new(rpc_url.clone(), Auth::CookieFile(cookie_file.clone()));
    |                                             ^^^^^^^^^^^^^^^
    |                                             |
    |                                             expected `&str`, found struct `std::string::String`
    |                                             help: consider borrowing here: `&rpc_url`

error[E0308]: mismatched types
   --> /home/sandi/.cargo/registry/src/github.com-1ecc6299db9ec823/bitcoind-0.13.0/src/lib.rs:193:39
    |
193 |                     break Client::new(node_url_default, Auth::CookieFile(cookie_file.clone()))
    |                                       ^^^^^^^^^^^^^^^^
    |                                       |
    |                                       expected `&str`, found struct `std::string::String`
    |                                       help: consider borrowing here: `&node_url_default`
```

I can fix it locally by editing the specific file. But since it's a dependency of a dependency, is it possible to patch it to pass the CI tests?

sandipndev:
Waiting for the following PRs to get merged:
- https://github.com/RCasatta/rust-bitcoincore-rpc/pull/1
- https://github.com/RCasatta/bitcoind/pull/29

notmandatory:
I created a couple more PRs that also need to be merged to support this PR:

1. https://github.com/RCasatta/bitcoind/pull/30
2. https://github.com/RCasatta/electrsd/pull/21

~~1. https://github.com/RCasatta/rust-bitcoincore-rpc/pull/2~~ (optional, not really needed for this PR)

src/testutils/blockchain_tests.rs
```diff
@@ -875,6 +873,53 @@ macro_rules! bdk_blockchain_tests {
...
+                let wallet_descriptor = "tr(tprv8ZgxMBicQKsPdBtxmEMPnNq58KGusNAimQirKFHqX2yk2D8q1v6pNLiKYVAdzDHy2w3vF4chuGfMvNtzsbTTLVXBcdkCA1rje1JG6oksWv8/86h/1h/0h/0/*)#y283ssmn";
+                let change_descriptor = "tr(tprv8ZgxMBicQKsPdBtxmEMPnNq58KGusNAimQirKFHqX2yk2D8q1v6pNLiKYVAdzDHy2w3vF4chuGfMvNtzsbTTLVXBcdkCA1rje1JG6oksWv8/86h/1h/0h/1/*)#47zsd9tt";
+
+                test_client.bitcoind.client
+                    .import_descriptors(
+                        vec![
+                            ImportDescriptorRequest::new(wallet_descriptor, false),
+                            ImportDescriptorRequest::new(change_descriptor, false),
+                        ]
+                    ).unwrap();
```
sandipndev: Updated this section to match the newer prototype of `import_descriptors` in `rust-bitcoincore-rpc`

notmandatory:
Track via #63

notmandatory:
This needs to be reworked a little to use raw rpc calls to create our test `tr()` descriptor wallet as @RCasatta suggested. Once https://github.com/rust-bitcoin/rust-bitcoincore-rpc/pull/174 is merged we can rework the tests to use the lib instead.

Moving to '0.15.0' feature freeze so we have more time to work on it.

notmandatory:
I got the blockchain_tests to work without the `core-rpc` or `bitcoind` patches :tada:. Once I have my code cleaned up I'll add it as a commit to this PR. Moving this back into code freeze for `0.14.0`.

notmandatory (approved):
ACK afa1ab4ff8dba441d30bc745eca2d634fbd7f017

notmandatory:
FYI, here is an explanation from stackexchange that we followed for creating a P2TR wallet in `bitcoind` and then creating a deposit address for it: https://bitcoin.stackexchange.com/questions/108006/how-to-make-a-taproot-transaction-with-bitcoin-cli

sandipndev:
Awesome, thanks for fixing this PR. Calling the RPC commands, for now, seems to work perfectly and tests run - which is fantastic and ACK https://github.com/bitcoindevkit/bdk/pull/402/commits/afa1ab4ff8dba441d30bc745eca2d634fbd7f017 🎉

src/testutils/blockchain_tests.rs
```diff
@@ -145,9 +145,7 @@ impl TestClient {
 
         let bumped: serde_json::Value = self.call("bumpfee", &[txid.to_string().into()]).unwrap();
         let new_txid = Txid::from_str(&bumped["txid"].as_str().unwrap().to_string()).unwrap();
-
-        let monitor_script =
-            tx.vout[0].script_pub_key.addresses.as_ref().unwrap()[0].script_pubkey();
+        let monitor_script = Script::from_hex(&mut tx.vout[0].script_pub_key.hex.to_hex()).unwrap();
```
rajarshimaitra, line 148: Is there any special benefit of getting the script this way?

src/testutils/blockchain_tests.rs
```diff
@@ -911,6 +909,102 @@ macro_rules! bdk_blockchain_tests {
...
+                            timestamp: "now".to_string(),
+                        }
+                    }
+                }
+
+                // 1. Create and add descriptors to a test bitcoind node taproot wallet
+
+                // TODO replace once rust-bitcoincore-rpc with PR 174 released
+                // https://github.com/rust-bitcoin/rust-bitcoincore-rpc/pull/174
+                let _createwallet_result: Value = test_client.bitcoind.client.call("createwallet", &["taproot_wallet".into(),false.into(),true.into(),serde_json::to_value("").unwrap(), false.into(), true.into()]).unwrap();
```
rajarshimaitra, line 957: It seems we don't need to hold the create wallet result. So we can ignore the let binding here.

src/testutils/blockchain_tests.rs
```diff
@@ -911,6 +909,102 @@ macro_rules! bdk_blockchain_tests {
...
+                let wallet_descriptor = "tr(tprv8ZgxMBicQKsPdBtxmEMPnNq58KGusNAimQirKFHqX2yk2D8q1v6pNLiKYVAdzDHy2w3vF4chuGfMvNtzsbTTLVXBcdkCA1rje1JG6oksWv8/86h/1h/0h/0/*)#y283ssmn";
+                let change_descriptor = "tr(tprv8ZgxMBicQKsPdBtxmEMPnNq58KGusNAimQirKFHqX2yk2D8q1v6pNLiKYVAdzDHy2w3vF4chuGfMvNtzsbTTLVXBcdkCA1rje1JG6oksWv8/86h/1h/0h/1/*)#47zsd9tt";
+
+                let tr_descriptors = vec![
+                            ImportDescriptorRequest::new(wallet_descriptor, false),
+                            ImportDescriptorRequest::new(change_descriptor, false),
+                        ];
+
+                // TODO replace once rust-bitcoincore-rpc with PR 199 released
+                let _import_result: Value = taproot_wallet_client.call("importdescriptors", &[serde_json::to_value(tr_descriptors).unwrap()]).unwrap();
```
rajarshimaitra, line 971: Same as before, no need to hold the result.

rajarshimaitra (commented):
tACK https://github.com/bitcoindevkit/bdk/pull/402/commits/afa1ab4ff8dba441d30bc745eca2d634fbd7f017.

below are few minor question and suggestions.

src/testutils/blockchain_tests.rs
sandipndev: No, I think this could be removed out

notmandatory: I'm going to leave the unused results in for now since they don't hurt anything and we can remove them when we redo this test when updated dependencies are ready.

src/testutils/blockchain_tests.rs
notmandatory: I believe this was required because `.script_pubkey()` function didn't support P2TR scripts.

Issue #81, opened 2020-07-27
Compact block filter Blockchain backend


notmandatory:
This is an overview issue (based on [same issue for murmel](https://github.com/rust-bitcoin/murmel/issues/37)) for discussion and tracking the addition of a fully functional compact filters Blockchain backend to `bdk`. Ideally we'll have additional issues and/or PRs for each missing feature to link to this issue and we check off each feature as it's completed.

The below [BIP-157](https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki#client-operation)/[158](https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki#Block_Filters)  protocol features are based on the [BIP-157 Client Operation](https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki#client-operation) section. Please suggest additions and corrections in the comments.

Missing protocol features for `bdk` are indicated below with an unchecked box [ ], completed features with a checked box [x].

#### Terminology

- peer: bitcoin full node peer
- block header: bitcoin standard block header
- filter header: BIP-157 compact block filter header
- block filter: BIP-157/158 compact block filter

### Peer Discovery

1. - [x] Allow user to specify peer addresses
2. - [ ] Receive peer addresses via [addr] messages from connected peers
3. - [ ] Store new peer addresses
4. - [ ] If no provided or previously stored peers, find seed peer addresses via DNS

### Connect to Peers

1. - [ ] Allow user to specify number of peers to connect to
2. - [ ] Allow user to specify minimum number of BIP-157 peers to connect to
3. - [x] Connect to peers and request signalling bits via [version] message 
4. - [ ] Keep track of peers signalling service bit NODE_COMPACT_FILTERS
5. - [ ] Connect to specified number of peers
6. - [ ] Connect to minimum number of BIP-157 peers

### Request Block Headers

1. - [x] Request block headers via [getheaders] message to connected peers
2. - [x] Request block headers from genesis to most work block, 1 to 2000 blocks per request
3. - [ ] Optional: Request block headers from stored checkpoint block instead of genesis block
4. - [ ] Optional: Request block headers from DNS (see https://bitcoinheaders.net/)

### Receive Block Headers

1. - [x] Receive block headers [headers] from peers
2. - [x] Build chain of block headers from stored and received new block headers
3. - [x] Store new block headers and new block header tip
4. - [ ] Disconnect peers whose longest chain has significantly less work than known tip
[...]

PR #403, opened 2021-07-21
Add Peer Manager structure

rajarshimaitra:
### Description

This PR adds a Peer Manager structure. The Peer manager structure is responsible for marinating a live directory connected 
CBF/Non-CBF peers. The number of CBF/Non-CBF connections are configurable, and the default values are 2 and 3 respectively.

The Peer Manager manages an underlying address manager and uses it to fetch addresses from the network. The manager maintains a ban score for each individual peer. A threshold banscore of 100 is used to signal to ban of a peer. Once a peer address is banned, it will not be connected to subsequent runs.

The Peer Manager is intended to be used to handle parallel `cfilter` downloading from multiple peers.

### Notes to the reviewers

With this PR the first two segments of capabilities in https://github.com/bitcoindevkit/bdk/issues/81 (Peer Discovery, Connect to Peers) are potentially completed.
 
Please verify the same.

Tests are included but ignored to stop CI from hanging. It's useful to run them locally to check functionality.

Peer Manager Initialisation is quite fast if there is an existing cache of addresses (which will exist after the first run).

Some Dead Codes are allowed to suppress unnecessary warnings, this will be removed once Peer Manager is being used by the existing CBF blockchain backend. 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/blockchain/compact_filters/mod.rs
+++ b/src/blockchain/compact_filters/mod.rs
@@ -65,6 +65,7 @@ use rocksdb::{Options, SliceTransform, DB};
 
 mod address_manager;
 mod peer;
+mod peermngr;
 mod store;
 mod sync;
 
--- a/src/blockchain/compact_filters/peermngr.rs
+++ b/src/blockchain/compact_filters/peermngr.rs
@@ -1,0 +1,578 @@
+use super::address_manager::{AddressManager, AddressManagerError, DiscoveryProgress};
+use super::peer::{Mempool, Peer, PeerError, TIMEOUT_SECS};
+
+use std::net::SocketAddr;
+use std::sync::Arc;
+use std::time;
+
+use bitcoin::network::constants::{Network, ServiceFlags};
+use bitcoin::network::message::NetworkMessage;
+
+use std::path::PathBuf;
+
+use std::collections::BTreeMap;
+
+// Peer Manager Configuration constants
+const MIN_CBF_PEERS: usize = 2;
+const MIN_TOTAL_PEERS: usize = 5;
+const MIN_CRAWLER_THREADS: usize = 20;
+const BAN_SCORE_THRESHOLD: usize = 100;
+const RECEIVE_TIMEOUT: time::Duration = time::Duration::from_secs(TIMEOUT_SECS);
+
+#[allow(dead_code)]
+/// An Error structure describing Peer Management errors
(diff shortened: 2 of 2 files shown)


notmandatory:
Have you thought about how to possibly use mock Peers for testing? would be much easier if Peer and maybe PeerManager were traits that could be mocked. And PeerManager were used to create all new Peers. Or something like that.

rajarshimaitra:
> Have you thought about how to possibly use mock Peers for testing?

The only thing that comes to my mind to mock the p2p is to hand-create a p2p interface that can respond to network messages, if that's something we would need, I can start working on it. 

Also once we have the stitch between peer manager and current CBF `sync` method, we would need to have some framework to test it against various parallel syncing scenarios. I am not sure what's the best way to do that and also it would be quite a complex thing to design. I would need some suggestion and help if we decide to built our custom test framework for CBF backend testing. Will also post this in the discord for better visibility.  

For Peer Manager testing only though I have another idea. We can start with an already preloaded address manager with cbf and non-cbf addresses. In that way, we won't have to worry about testing address fetching, and can just test the peer manager logic. I think that's a possible way to test Peer Manager and I will work on it next.

rajarshimaitra:
> For Peer Manager testing only though I have another idea. We can start with an already preloaded address manager with cbf and non-cbf addresses.

After some trials, this also doesn't seem to be working. We can create a dummy address manager but the peer manager needs to connect to those addresses to establish a live peer structure. And bitcoin network addresses changes all the time. So if we just create a list of addresses without checking if they are live we will eventually have unresponsive addresses and the test will hang again.

To mitigate against this in the real network the real address manager always checks addresses for live connection before passing them on to the peer manager. But if we skip that crawling part in our test we get back the "unresponsive address" problem instead.

So as a result peer manager tests have to be ignored for the same reason as address manager, for now. Until we come up with some comprehensive solution to create a mock p2p network that can handle all of it.

rajarshimaitra:
@notmandatory for time being we can merge this into `compact_filters` branch, and remove from open PR list. 

We can let this branch go unsynced with `master`. Once work starts again on CBF, we can rebase and start from here..

notmandatory (approved):
ACK d2da3755f4cb35b73740d35f9835487bb4ce5655 for merging to `compact_filters` branch to consolidate for future CBF work.

Issue #391, opened 2021-07-10
Confirmed transactions show up as pending on application restart


thunderbiscuit:
This is happening on `bdk-jni`, and I'm not sure if it has to do with the db/cache on mobile or if it's coming from the bdk side.

@RCasatta could this be related to the #370 fix?

The issue is the following. Once you've synced while a transaction was pending, the new type `confirmation_time` is `None` (`null` on the Kotlin side). This is great. What's more, if you `sync` the wallet and the transaction has been confirmed, the vallue is updated to reflect the timestamp and a block height. All good. The problem seem to happen on application restart: a call to `transaction_list()` will return that transaction as pending again. You can then `sync` the wallet and it will update the `confirmation_time` value, but this sync has to be done (from what I can tell) every time. It's like the database doesn't remember that the `transactionDetails` were updated.

<center>
<img src="https://user-images.githubusercontent.com/39974688/125169707-68c58680-e179-11eb-99b2-3e3e3833f60c.gif" width="300px" />
</center>

RCasatta:
can you try with this patch?

```
diff --git a/src/database/keyvalue.rs b/src/database/keyvalue.rs
index 9978497..85de6ea 100644
--- a/src/database/keyvalue.rs
+++ b/src/database/keyvalue.rs
@@ -377,7 +377,9 @@ impl BatchDatabase for Tree {
     }
 
     fn commit_batch(&mut self, batch: Self::Batch) -> Result<(), Error> {
-        Ok(self.apply_batch(batch)?)
+        self.apply_batch(batch)?;
+        self.flush()?;
+        Ok(())
     }
 }

```

Just an untested quick try, I'll check it properly on Monday

thunderbiscuit:
I'm sorry I don't really know how to pull this in the JNI in a quick/test way... but I do want to help test it. I'm just not sure how.

afilini:
I had a brief discussion about this with @RCasatta, we think it might be a good idea to add a `flush()` method to `Database` and to explicitly call it after every write.

Then we'll implement it for `sled` to ensure writes are flushed to disk.

notmandatory:
I think the new `sqlite` database backend might solve this for mobile apps. Will test it once #376 is in a merged. If so we may not need the `sync()` function after all, but will need to document it if the `key-value-db` isn't recommended for mobile.

thunderbiscuit:
This issue is resolved on my end. Thanks!

PR #409, opened 2021-07-27
Add flush method to Database trait

RCasatta:
### Description

Expose `flush` method in the `Database` trait to explicitly flush db changes to disk

A step toward fixing https://github.com/bitcoindevkit/bdk/issues/391.

Following steps are
- [ ] Add call to flush from JNI
- [ ] Call the flush method from the app (ideally in a App Destructor callback or if not possible at the end of sync)

#### All Submissions:

#### New Features:

### DIFF
--- a/src/database/keyvalue.rs
+++ b/src/database/keyvalue.rs
@@ -367,6 +367,10 @@ impl Database for Tree {
             Ok(val)
         })
     }
+
+    fn flush(&mut self) -> Result<(), Error> {
+        Ok(Tree::flush(self).map(|_| ())?)
+    }
 }
 
 impl BatchDatabase for Tree {
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -134,6 +134,9 @@ pub trait Database: BatchOperations {
     ///
     /// It should insert and return `0` if not present in the database
     fn increment_last_index(&mut self, keychain: KeychainKind) -> Result<u32, Error>;
+
+    /// Force changes to be written to disk, returning the number of bytes written
+    fn flush(&mut self) -> Result<(), Error>;
 }
 
 /// Trait for a database that supports batch operations
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -135,7 +135,7 @@ pub trait Database: BatchOperations {
     /// It should insert and return `0` if not present in the database
     fn increment_last_index(&mut self, keychain: KeychainKind) -> Result<u32, Error>;
 
-    /// Force changes to be written to disk, returning the number of bytes written
+    /// Force changes to be written to disk
     fn flush(&mut self) -> Result<(), Error>;
 }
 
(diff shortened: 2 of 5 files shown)


src/database/keyvalue.rs
```diff
@@ -367,6 +367,10 @@ impl Database for Tree {
             Ok(val)
         })
     }
+
+    fn flush(&mut self) -> Result<usize, Error> {
+        Ok(Tree::flush(self)?)
```
rajarshimaitra: Would we use the "bytes written" value in some way? This will also force library users implementing their own DB's flush to return some number.

If it's not super useful maybe we shouldn't expose this as general API just because sled does it?

rajarshimaitra (commented):
ACK.

One API comment.

LLFourn:
We had a short discussion in meeting today. Have we considered just flushing after every write always on sled? I'm not really a fan of forcing the user to care about this.

thunderbiscuit:
I think one of the things discussed was that it was a performance hit on all applications when it might have been an issue mostly only for Android/mobile.

Would it be possible to have a flag turned on by default where it flushes to disk for all operations, but if you do care about efficiency and don't want the performance hit you can turn it off? Or maybe the opposite? Doesn't happen by default but you can have it flush to disk for all if you are working on mobile stuff (but then we're back to your point about having to make the user care about it).

LLFourn:
I really don't think that there would be a performance issue here. It's a wallet not a full node! In any case it's better to not optimize early. Despite it only being an issue on android I would prefer my wallet to flush to disk wherever possible.

RCasatta:
Flushing after every sync (maybe improving later to do it only if the state changed) without adding a new method was my first proposal, the safest path with only a small performance hit I think (sync is a network method, adding a write to disk shouldn't change much its execution time in percentual terms).

In the 14th July meeting, the general feedback was instead to have an explicit flush method (see also https://github.com/bitcoindevkit/bdk/issues/391#issuecomment-878302024).

I am in favor to close this and flush at sync but maybe @afilini has something to add

tcharding (approved):
Seems reasonable to me.

afilini:
I think the main reason why I like the explicit `flush()` method is just that it gives a way to the library to force a flush whenever it thinks one is needed (mainly after sync I guess).

We could fix this just for the sled db, but then when the external users implement the trait on other external types there's a chance they could make the same mistake. With an explicit method they are forced to provide an implementation for `flush()` and then the library decides when to use it.

I agree with @rajarshimaitra that maybe returning the number of changed bytes isn't ideal, because if you implement this for things like Postgres it's kinda meaningless, and it's not that the library needs it in any way.

RCasatta:
Rebased and removed the returned `usize`

rajarshimaitra:
ReACK https://github.com/bitcoindevkit/bdk/pull/409/commits/fe30716fa2ba59eb8e94cfa6a506597ab7febb1f

src/database/mod.rs
```diff
@@ -134,6 +134,9 @@ pub trait Database: BatchOperations {
     ///
     /// It should insert and return `0` if not present in the database
     fn increment_last_index(&mut self, keychain: KeychainKind) -> Result<u32, Error>;
+
+    /// Force changes to be written to disk, returning the number of bytes written
```
notmandatory: Needs to be updated to match last change:
```suggestion
    /// Force changes to be written to disk
```

notmandatory (approved):
utACK 19d74955e2c8c0701dc8acda8f3189489b393104

PR #411, opened 2021-07-28
exposing get_funded_wallet

ulrichard:
* moving get_funded_wallet out of the test section to make it available for https://github.com/ulrichard/bdk-reserves
* moving the wallet::network() getter function to make it more accessable for https://github.com/ulrichard/bdk-reserves

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -18,6 +18,7 @@ use std::collections::HashMap;
 use std::collections::{BTreeMap, HashSet};
 use std::fmt;
 use std::ops::{Deref, DerefMut};
+use std::str::FromStr;
 use std::sync::Arc;
 
 use bitcoin::secp256k1::Secp256k1;
@@ -55,6 +56,7 @@ use tx_builder::{BumpFee, CreateTx, FeePolicy, TxBuilder, TxParams};
 use utils::{check_nlocktime, check_nsequence_rbf, After, Older, SecpCtx, DUST_LIMIT_SATOSHI};
 
 use crate::blockchain::{Blockchain, Progress};
+use crate::database::memory::MemoryDatabase;
 use crate::database::{BatchDatabase, BatchOperations, DatabaseUtils};
 use crate::descriptor::derived::AsDerived;
 use crate::descriptor::policy::BuildSatisfaction;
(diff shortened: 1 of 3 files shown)


notmandatory:
This looks good, my only concern is exposing test functions in the release builds for `bdk`. I think it should be possible to put any test utility functions behind a feature flag. I'm working on testing that in [another fork ](https://github.com/notmandatory/bdk/tree/test_features)and if I get it to works I'll make some suggestions on how to add something like a `test-wallet` feature as part of this PR or as a follow-up PR.

notmandatory:
~~BTW, my PR to add the feature gates as a follow up to this PR is #412~~ Looks like enabling features only for tests isn't currently possible.

notmandatory (approved):
ACK 29f32d89d336d409c54d35c5fd08c739d82b861a

ulrichard:
Thanks for testing that stuff. 
Yes https://github.com/rust-lang/cargo/issues/2911 would be really cool.

tcharding:
Perhaps we should put this behind `#[cfg(debug_assertions)]` so its only available in debug builds?

ulrichard:
> Perhaps we should put this behind `#[cfg(debug_assertions)]` so its only available in debug builds?

That looks like a good idea. I quickly tried it and executed a few commands that came to my mind. The only one that failed was:
`cargo test --release`
But I don't know if that is actually used by anybody.

afilini:
I think fuzzing is generally done by compiling to release mode with tests enabled. I don't think it's really a problem if it's available in release mode to be honest, I would just leave it in there

src/wallet/mod.rs
```diff
@@ -1548,19 +1551,60 @@ where
     }
 }
 
+/// Return a fake wallet that appears to be funded for testing.
+pub fn get_funded_wallet(
```
afilini, line 1555: Did you try with `#[cfg(test)]` to see if you could hide this to non-testing environments?

I didn't realize this yesterday, I thought this was still part of the `test` module, which is only present when testing is enabled.

ulrichard: If I add this, then I get the following error in bdk-reserves: "unresolved import `bdk::wallet::get_funded_wallet`"

And if I try the following in bdk-reserves/Cargo.toml: "bdk = { ...  features = ["electrum", "test"] }" then I get the following error: "the package `bdk-reserves` depends on `bdk`, with features: `test` but `bdk` does not have these features."

It would be cool to make it work like this, but I don't know how.

afilini (approved):

PR #412, opened 2021-07-29
Add testutils and test-wallet features

notmandatory:
### Description

This PR creates two new feature flags, `testutils` to enable the testutils module, and `test-wallet` to enable the now exposed `wallet::get_funded_wallet()`  function. Exposing the `wallet::get_funded_wallet()` function and moving the  `Wallet::network()` function are needed to make testing easier for supporting projects like `bdk-reserves` without including this functionality in non-test builds.

### Notes to the reviewers

All doctests should work with the new `testutils` feature without having this module exposed for non-test builds.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,17 +62,20 @@ rpc = ["bitcoincore-rpc"]
 
 
 # Debug/Test features
+testutils = []
 test-blockchains = ["bitcoincore-rpc", "electrum-client"]
 test-electrum = ["electrum", "electrsd/electrs_0_8_10", "test-blockchains"]
 test-rpc = ["rpc", "electrsd/electrs_0_8_10", "test-blockchains"]
 test-esplora = ["esplora", "electrsd/legacy", "electrsd/esplora_a33e97e1", "test-blockchains"]
 test-md-docs = ["electrum"]
+test-wallet = []
 
 [dev-dependencies]
 lazy_static = "1.4"
 env_logger = "0.7"
 clap = "2.33"
 electrsd = { version= "0.8", features = ["trigger", "bitcoind_0_21_1"] }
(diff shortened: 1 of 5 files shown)


notmandatory:
@LLFourn I found another way here to add the `testutils` feature back in using `bdk` as it's own dev dependency. Too hacky? seem to work for doc tests and I don't see any down sides.

LLFourn:
@notmandatory 🤯 . Did you get that idea from somewhere? can we not just add testutils as a default feature I guess then it will be available during doctests I guess?

notmandatory:
It probably would work as a default feature but then anyone using BDK as a normal dependency would need to set no-default features or end with the unwanted 'testutils'. I ran across this in a old rustc bug where adding a crate as a test dep of itself used to panicked the compiler. But issue got fixed and seems to be supported.

notmandatory:
This is the issue comment that gave me the idea:

https://github.com/rust-lang/cargo/issues/2911#issuecomment-749580481

LLFourn:
They seem to say it doesn't work in that the hack enables it for non-dev as well: https://github.com/rust-lang/cargo/issues/2911#issuecomment-771895047

In rust you only pay for what you use so there is no real harm in not having it under a feature flag or as a default feature flag. My personal preference is to leave without and wait until https://github.com/rust-lang/rust/issues/67295 and https://github.com/rust-lang/cargo/issues/2911 are fixed.

notmandatory:
Ok ya I read through a few more related issues and looks like this is a hack that will [soon not be supported](https://github.com/rust-lang/cargo/issues/9518) and as you said probably isn't doing what I think it is. 

Closing this PR.

Issue #244, opened 2020-12-21
Reduce build time for `docs.rs`


afilini:
`docs.rs` has a timeout of 15 minutes for building the docs, which is very close to how long it takes to build `bdk`, mainly due to the `rocksdb` dependency for `compact_filters`. We should figure something out to reduce the build time, otherwise we risk getting build failures as the project grows.

notmandatory:
Fixed with #416 in [release 0.11.0](https://github.com/bitcoindevkit/bdk/releases?page=1)

PR #416, opened 2021-08-02
Disable default features for rocksdb

tcharding:
### Description

In an effort to reduce the build times of `rocksdb` we can set
`default-features` to false.

Please note, the build speed up is minimil

With default features:
```
cargo check --features compact_filters  890.91s user 47.62s system 352% cpu 4:26.55 total
```

Without default features:
```
cargo check --features compact_filters  827.07s user 47.63s system 352% cpu 4:08.39 total
```

Enable `snappy` since it seems like this is the current default compression
algorithm, therefore this patch (hopefully) makes no changes to the usage of the
`rocksdb` library in `bdk`. From the `rocksdb` code:

```
    /// Sets the compression algorithm that will be used for compressing blocks.
    ///
    /// Default: `DBCompressionType::Snappy` (`DBCompressionType::None` if
    /// snappy feature is not enabled).
    ///
    /// # Examples
    ///
    /// ```
    /// use rocksdb::{Options, DBCompressionType};
    ///
    /// let mut opts = Options::default();
    /// opts.set_compression_type(DBCompressionType::Snappy);
    /// ```
    pub fn set_compression_type(&mut self, t: DBCompressionType) {
        ....
```

### Notes to the reviewers

Done in an attempt to help with #244 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ reqwest = { version = "0.11", optional = true, features = ["json"] }
 ureq = { version = "2.1", default-features = false, features = ["json"], optional = true }
 futures = { version = "0.3", optional = true }
 async-trait = { version = "0.1", optional = true }
-rocksdb = { version = "0.14", optional = true }
+rocksdb = { version = "0.14", default-features = false, fetures = ["snappy"], optional = true }
 cc = { version = ">=1.0.64", optional = true }
 socks = { version = "0.3", optional = true }
 lazy_static = { version = "1.4", optional = true }


RCasatta (approved):
LGTM

for rocksdb: 
`default = ["snappy", "lz4", "zstd", "zlib", "bzip2"]`
it appears we don't need the other compression algorithms

tcharding:
> LGTM
> 
> for rocksdb:
> `default = ["snappy", "lz4", "zstd", "zlib", "bzip2"]`
> it appears we don't need the other compression algorithms

That was my guess also, I have zero experience with `rocksdb`. At first I tried to build with no features enabled and that builds just fine but from the docs I posted above I'm guessing that this means no compression which I guess we don't want.

Thanks for the review!

afilini (approved):

afilini:
I think this needs a rebase before we can merge it, because in master there are now two more CI jobs

PR #428, opened 2021-08-30
Fix imports in test macros

luckysori:
### Description

The test macros `testutils` and `populate_test_db` require users to put imports in scope of the generated code, which is not very ergonomic.

By using the re-exported versions of certain crates we ensure that these macros can be used without any extra imports.

#### All Submissions:


tcharding (approved):
For what its worth, ACK!

notmandatory (approved):
ACK acf157a99a305226203d2b55a567291a93c64720, looks like a nice improvement.

PR #430, opened 2021-08-30
[CI] Fixed esplora blockchain tests in CI

### LINKED ISSUE #431: [Bug] Esplora reqwest blockchain tests are failing
### Bug

It seems that `esplora-reqwest` blockchain tests have been failing for some time now.

The CI missed it because all the esplora tests were being skipped. 

The major sources of failure are `async` related. Especially calling `async` functions in `bdk_blockchain_tests` macro. 

Error sample 
```
error[E0599]: no method named `unwrap` found for opaque type `impl futures::Future` in the current scope
   --> src/testutils/blockchain_tests.rs:875:52
    |
875 |                   wallet.sync(noop_progress(), None).unwrap();
    |                                                      ^^^^^^ method not found in `impl futures::Future`
    | 
   ::: src/blockchain/esplora/mod.rs:125:1
    |
125 | / crate::bdk_blockchain_tests! {
126 | |     fn test_instance(test_client: &TestClient) -> EsploraBlockchain {
127 | |         EsploraBlockchain::new(&format!("http://{}",test_client.electrsd.esplora_url.as_ref().unwrap()), 20)
128 | |     }
129 | | }
    | |_- in this macro invocation
    |
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider `await`ing on the `Future` and calling the method on its `Output`
    |
875 |                 wallet.sync(noop_progress(), None).await.unwrap();
    |                                                    ^^^^^^

error: aborting due to 58 previous errors

Some errors have detailed explanations: E0195, E0308, E0599.
For more information about an error, try `rustc --explain E0195`.
```

### Reproduce: 

Edit test-esplora feature in cargo like `test-esplora = ["use-esplora-reqwest", "electrsd/legacy", "electrsd/esplora_a33e97e1", "test-blockchains"]`. 

then run `cargo test --features test-esplora`

### Analysis
As far as I understand, its happening because in the test macro we are calling wallet APIs in non async mode, whereas, for `esplora-reqwest` these are async functions.

https://github.com/bitcoindevkit/bdk/blob/721748e98f95f8f97d508322ae7871ec4c49f0c4/src/testutils/blockchain_tests.rs#L378-L380

### Thoughts
I tried putting in the `maybe_async` and `maybe_await` macros where applicable. But `#[test]` are not allowed to be async (or at least the vanilla flavour of them).

we can have async tests but with using `tokio` directly. But then, they cannot be non-async anymore.

Or maybe we can create some proc macros like `#[maybe_async_test]`? 

I noticed this issue few weeks back in bdk-cli https://github.com/bitcoindevkit/bdk-cli/pull/36#issue-702356592. Was confused why I couldn't reproduce it in BDK. The reason is the current `test-esplora` feature is broken, and is not compiling the blockchian tests. This is now fixed with https://github.com/bitcoindevkit/bdk/pull/430

rajarshimaitra:
### Description

Fixes #431 and esplora blockchain test skips in CI

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -78,15 +78,20 @@ jobs:
         run: cargo test --features test-md-docs --no-default-features -- doctest::ReadmeDoctests
 
   test-blockchains:
-    name: Test ${{ matrix.blockchain.name }}
+    name: Blockchain ${{ matrix.blockchain.features }}
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
         blockchain:
           - name: electrum
+            features: test-electrum
           - name: rpc
+            features: test-rpc
           - name: esplora
+            features: test-esplora,use-esplora-reqwest
+          - name: esplora
+            features: test-esplora,use-esplora-ureq    
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -104,8 +109,8 @@ jobs:
           toolchain: stable
           override: true
       - name: Test
-        run: cargo test --features test-${{ matrix.blockchain.name }} ${{ matrix.blockchain.name }}::bdk_blockchain_tests
-
+        run: cargo test --no-default-features --features ${{ matrix.blockchain.features }} ${{ matrix.blockchain.name }}::bdk_blockchain_tests
+  
   check-wasm:
     name: Check WASM
     runs-on: ubuntu-16.04
@@ -138,7 +143,6 @@ jobs:
         run: rustup update
       - name: Check
         run: cargo check --target wasm32-unknown-unknown --features use-esplora-reqwest --no-default-features
-
 
   fmt:
     name: Rust fmt
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ readme = "README.md"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-bdk-macros = "0.5"
+bdk-macros = { path = "macros"} # TODO: Change this to version number after next release.
 log = "^0.4"
 miniscript = "^6.0"
 bitcoin = { version = "^0.27", features = ["use-serde", "base64"] }
@@ -37,6 +37,10 @@ bitcoinconsensus = { version = "0.19.0-3", optional = true }
 
 # Needed by bdk_blockchain_tests macro
 core-rpc = { version = "0.14", optional = true }
+
+# Platform-specific dependencies
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+tokio = { version = "1", features = ["rt"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 async-trait = "0.1"
@@ -70,7 +74,7 @@ rpc = ["core-rpc"]
 async-interface = ["async-trait"]
 electrum = ["electrum-client"]
 # MUST ALSO USE `--no-default-features`.
-use-esplora-reqwest = ["async-interface", "esplora", "reqwest", "futures"]
+use-esplora-reqwest = ["esplora", "reqwest", "futures"]
 use-esplora-ureq = ["esplora", "ureq"]
 # Typical configurations will not need to use `esplora` feature directly.
 esplora = []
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -24,7 +24,7 @@ rand = "^0.7"
 sled = { version = "0.34", optional = true }
 electrum-client = { version = "0.8", optional = true }
 reqwest = { version = "0.11", optional = true, features = ["json"] }
-ureq = { version = "2.1", default-features = false, features = ["json"], optional = true }
+ureq = { version = "2.1", default-features = false, features = ["json", "tls"], optional = true }
```
notmandatory: Since `tls` the only default feature for ureq I'd change this line to:
```suggestion
ureq = { version = "2.1", features = ["json"], optional = true }
```
Also per below comment should be fine to have `tls` always enabled:
https://github.com/bitcoindevkit/bdk-cli/pull/41#discussion_r698333498

notmandatory (changes requested):
Looks good, also checkout related https://github.com/tcharding/bdk/pull/2. We should be able to test esplora with ureq and reqwest. If you can incorporate into this PR I'll close the one I did.

Cargo.toml
rajarshimaitra: Noted. Will be done one a different PR.

rajarshimaitra:
Updated with suggested refactors and async changes. 

Rearranged the commits.

Moved out `tls` changes.

notmandatory (approved):

notmandatory:
ACK 2caa590438205b3d9a60c46cabf90785d5b08069

.github/workflows/cont_integration.yml
```diff
@@ -78,15 +78,20 @@ jobs:
...
       matrix:
         blockchain:
           - name: electrum
+            features: test-electrum
           - name: rpc
+            features: test-rpc
           - name: esplora
+            features: test-esplora,use-esplora-reqwest
+          - name: esplora
+            features: test-esplora,use-esplora-ureq
```
tcharding, line 94: This line has trailing whitespace.

.github/workflows/cont_integration.yml
```diff
@@ -104,8 +109,8 @@ jobs:
           toolchain: stable
           override: true
       - name: Test
-        run: cargo test --features test-${{ matrix.blockchain.name }} ${{ matrix.blockchain.name }}::bdk_blockchain_tests
-
+        run: cargo test --no-default-features --features ${{ matrix.blockchain.features }} ${{ matrix.blockchain.name }}::bdk_blockchain_tests
+
```
tcharding, line 113: This line has trailing whitespace.

Cargo.toml
```diff
@@ -38,6 +38,10 @@ bitcoinconsensus = { version = "0.19.0-3", optional = true }
 # Needed by bdk_blockchain_tests macro
 core-rpc = { version = "0.14", optional = true }
 
+# Platform-specific dependencies
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+tokio = { version = "1", features = ["rt"] }
```
tcharding, line 43: Please correct me if I'm wrong, but this PR re-introduces a hard dependency on `tokio` just so we can test `esplora-reqwest` feature, that seems like a bad choice to me. A user of `bdk` that wishes to use `reqwest` will always be using `async`, why do we want to introduce a hard dependency on `tokio` just to be able to test in a manner that the library will not be used?

The `esplora-reqwest` feature can be tested coupled with either `async-interface` feature or using WASM target.

src/blockchain/esplora/reqwest.rs
```diff
@@ -106,19 +106,19 @@ impl Blockchain for EsploraBlockchain {
     }
 
     fn get_tx(&self, txid: &Txid) -> Result<Option<Transaction>, Error> {
-        Ok(self.url_client._get_tx(txid).await?)
+        Ok(await_or_block!(self.url_client._get_tx(txid))?)
```
tcharding, line 109: I do not like this usage of `await_or_block` because there is no reason to use `reqwest` in a blocking environment, we have `ureq` for that.

tcharding (commented):
Thanks for looking into this. I'm glad to see the things I missed in CI being remedied, however I'm not super fond of the re-introduction of hard dependency on `tokio` - seems like a step backwards to me.

tcharding:
All my comments are resolved by the implementation in #433, please see if you have any concerns with that approach @rajarshimaitra. Thanks.

notmandatory:
@tcharding @LLFourn as @rajarshimaitra noted on #433 even if this PR isn't the final solution it will unblock some other work (https://github.com/bitcoindevkit/bdk-cli/pull/41 and #429).  Also it's more of a roll back to how the `esplora::bdk_blockchain_tests` used to work before `ureq` was introduced.  Do you guys have any objection to merging this now while you work on a cleaner solution?

tcharding:
> @tcharding @LLFourn as @rajarshimaitra noted on #433 even if this PR isn't the final solution it will unblock some other work ([bitcoindevkit/bdk-cli#41](https://github.com/bitcoindevkit/bdk-cli/pull/41) and #429). Also it's more of a roll back to how the `esplora::bdk_blockchain_tests` used to work before `ureq` was introduced. Do you guys have any objection to merging this now while you work on a cleaner solution?

Don't stop the progress on my account :)

PR #415, opened 2021-08-02
Fix esplora integration test runner

tcharding:
### Description

Recently we added sub-modules to the `esplora` module for different HTTP
backends. We (Tobin) failed to update the integration test runner.

Modify the test running macro so that the backend is explicitly
selected:

  `cargo test --no-default-features --features=use-esplora-ureq,test-esplora`

### Notes to the reviewers

I can't actually get the integration tests to run to completion. They all (test-rpc, test-electrum, test-esplora) hang
```
$ echo $BITCOIND_EXE                                                                                                        ✘ 130 
/opt/bitcoin/bin/bitcoind
$ echo $ELECTRS_EXE
/home/tobin/.cargo/bin/electrs
$ cargo test --no-default-features --features=use-esplora-ureq,test-esplora
test blockchain::esplora::bdk_blockchain_tests::test_sync_address_reuse has been running for over 60 seconds
test blockchain::esplora::bdk_blockchain_tests::test_sync_after_send has been running for over 60 seconds
test blockchain::esplora::bdk_blockchain_tests::test_sync_before_and_after_receive has been running for over 60 seconds
test blockchain::esplora::bdk_blockchain_tests::test_sync_bump_fee_add_input_no_change has been running for over 60 seconds
```

However without this patch the integration tests do not run at all for esplora so I believe this is still correct. And it seems like a good way to ask if I am doing something wrong with how I'm trying to run the integration tests? Any suggestions please?

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [ ] I've added tests for the new feature
* [ ] I've added docs for the new feature
* [ ] I've updated `CHANGELOG.md`

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [ ] I'm linking the issue being fixed by this PR

PR #433, opened 2021-08-31
Fix esplora testing

tcharding:
### Description

Currently testing of esplora/reqwest is broken because of incorrect usage of, or lack of, `async`.

Enable testing of esplora/reqwest by using `maybe_await!` macro in tests and declaring all unit tests in `blockchain_tests.rs` to be async. This adds a dev dependency on `tokio`.

With this patch applied one can successfully run:
```
cargo test --features=test-esplora,use-esplora-reqwest --no-default-features
```

### Notes to the reviewers

This is, in my opinion, a better fix to https://github.com/bitcoindevkit/bdk/issues/431 than https://github.com/bitcoindevkit/bdk/pull/430 (i.e., supercedes #430) 

The primary advantage of this PR over #430 is it does not introduce a hard dependency on `tokio`.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -84,9 +84,9 @@ jobs:
       fail-fast: false
       matrix:
         blockchain:
+          # We test Esplora separately below.
           - name: electrum
           - name: rpc
-          - name: esplora
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -105,6 +105,34 @@ jobs:
           override: true
       - name: Test
         run: cargo test --features test-${{ matrix.blockchain.name }} ${{ matrix.blockchain.name }}::bdk_blockchain_tests
+
+  test-blockchains-esplora:
+    name: Test esplora-${{ matrix.backend.name }}
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        backend:
+          - name: ureq
+          - name: reqwest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ github.job }}-${{ hashFiles('**/Cargo.toml','**/Cargo.lock') }}
+      - name: Setup rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - name: Test
+        run: cargo test --no-default-features --features test-esplora,use-esplora-${{ matrix.backend.name }} esplora::bdk_blockchain_tests
 
   check-wasm:
     name: Check WASM
--- a/src/blockchain/esplora/ureq.rs
+++ b/src/blockchain/esplora/ureq.rs
@@ -377,3 +377,11 @@ impl ConfigurableBlockchain for EsploraBlockchain {
         Ok(EsploraBlockchain::new(config.base_url.as_str(), config.stop_gap).with_agent(agent))
     }
 }
+
+#[cfg(test)]
+#[cfg(feature = "test-esplora")]
+crate::bdk_blockchain_tests! {
+    fn test_instance(test_client: &TestClient) -> EsploraBlockchain {
+        EsploraBlockchain::new(&format!("http://{}",test_client.electrsd.esplora_url.as_ref().unwrap()), 20)
+    }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,7 +80,7 @@ esplora = []
 test-blockchains = ["core-rpc", "electrum-client"]
 test-electrum = ["electrum", "electrsd/electrs_0_8_10", "test-blockchains"]
 test-rpc = ["rpc", "electrsd/electrs_0_8_10", "test-blockchains"]
-test-esplora = ["esplora", "ureq", "electrsd/legacy", "electrsd/esplora_a33e97e1", "test-blockchains"]
+test-esplora = ["electrsd/legacy", "electrsd/esplora_a33e97e1", "test-blockchains"]
 test-md-docs = ["electrum"]
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,6 +88,7 @@ lazy_static = "1.4"
 env_logger = "0.7"
 clap = "2.33"
 electrsd = { version= "0.10", features = ["trigger", "bitcoind_0_21_1"] }
+tokio = { version = "1", features = ["macros", "rt"] }
 
 [[example]]
 name = "address_validator"
--- a/src/blockchain/esplora/reqwest.rs
+++ b/src/blockchain/esplora/reqwest.rs
@@ -355,6 +355,6 @@ impl ConfigurableBlockchain for EsploraBlockchain {
 #[cfg(feature = "test-esplora")]
 crate::bdk_blockchain_tests! {
     fn test_instance(test_client: &TestClient) -> EsploraBlockchain {
-        EsploraBlockchain::new(&format!("http://{}",test_client.electrsd.esplora_url.as_ref().unwrap()), None, 20)
+        EsploraBlockchain::new(&format!("http://{}",test_client.electrsd.esplora_url.as_ref().unwrap()), 20)
     }
 }
--- a/src/testutils/blockchain_tests.rs
+++ b/src/testutils/blockchain_tests.rs
@@ -375,10 +375,35 @@ macro_rules! bdk_blockchain_tests {
                 $block
             }
 
+            #[cfg(any(target_arch = "wasm32", feature = "async-interface"))]
+            async fn get_wallet_from_descriptors(descriptors: &(String, Option<String>), test_client: &TestClient) -> Wallet<$blockchain, MemoryDatabase> {
+                Wallet::new(&descriptors.0.to_string(), descriptors.1.as_ref(), Network::Regtest, MemoryDatabase::new(), get_blockchain(test_client)).await.unwrap()
+            }
+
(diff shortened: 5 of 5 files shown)


notmandatory:
This also looks like a good approach. Either way I agree if we only need `tokio` for testing it should be a dev-dependency. This PR also needs to have the CI pipeline updated to run the blockchain tests for esplora with `ureq` and `reqwest`, I think can copy from @rajarshimaitra PR. I'll check it out more closely tomorrow for a proper review.

src/blockchain/esplora/ureq.rs
```diff
@@ -377,3 +377,11 @@ impl ConfigurableBlockchain for EsploraBlockchain {
         Ok(EsploraBlockchain::new(config.base_url.as_str(), config.stop_gap).with_agent(agent))
     }
 }
+
+#[cfg(test)]
+#[cfg(feature = "test-esplora")]
+crate::bdk_blockchain_tests! {
```
notmandatory, line 383: I think you meant to move this to esplora/mod.rs as in your original #415 (and remove from esplora/reqwest.rs).

tcharding: If we want to be able to test both eslpora/reqwest _and_ esplora/ureq we need this to be in the ureq/reqwest submodules. I originally only put it in `mod.rs` thinking testing both was unnecessary, in hindsight this was wrong. With this change testing can be done by enabling feature `test-esplora` as well as the feature to enable the backend we want to test (e.g. `use-esplora-reqwest`).

tcharding:
> This PR also needs to have the CI pipeline updated to run the blockchain tests for esplora with ureq and reqwest

Oh, yes, my bad. Will add, thanks.

tcharding:
Patch `655560f Enable testing Esplora backends in CI` needs careful review please. The patch is pure cargo cult programing.

When I attempt to run the esplora integration test as it appears it will be run on CI, nothing runs? 

```
cargo test --no-default-features --features test-esplora,use-esplora-ureq esplora::bdk_blockchain_tests
    Finished test [unoptimized + debuginfo] target(s) in 0.05s
     Running unittests (/home/tobin/.cache/cargo/debug/deps/bdk-0eaddf4dc688116f)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 213 filtered out; finished in 0.00s
```

And without the test filter I get just the unit tests running
```
cargo test --no-default-features --features test-esplora,use-esplora-ureq esplora::bdk_blockchain_tests
...
test result: ok. 41 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 20.21s
```

RCasatta:
> When I attempt to run the esplora integration test as it appears it will be run on CI, nothing runs?

You have to give the right initial namespace, for example:

```
cargo test --no-default-features --features test-esplora,use-esplora-ureq blockchain::esplora
```

.github/workflows/cont_integration.yml
```diff
@@ -106,6 +106,34 @@ jobs:
...
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ github.job }}-${{ hashFiles('**/Cargo.toml','**/Cargo.lock') }}
+      - name: Setup rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - name: Test
+        run: cargo test --no-default-features --features test-esplora,use-esplora-${{ matrix.backend.name }} esplora::bdk_blockchain_tests
```
notmandatory, line 135: As @RCasatta suggested, this will work:
```suggestion
        run: cargo test --no-default-features --features test-esplora,use-esplora-${{ matrix.backend.name }} esplora::${{ matrix.backend.name }}::bdk_blockchain_tests
```
Or I think the better solution is to move the below chunk to `esplora/mod.rs` and remove it from `reqwest.rs` and `ureq.rs` (I confirmed it works):
```
#[cfg(test)]
#[cfg(feature = "test-esplora")]
crate::bdk_blockchain_tests! {
    fn test_instance(test_client: &TestClient) -> EsploraBlockchain {
        EsploraBlockchain::new(&format!("http://{}",test_client.electrsd.esplora_url.as_ref().unwrap()), 20)
    }
}
```
And then the above CI line does not need to change.

src/blockchain/esplora/ureq.rs
notmandatory: I locally tried moving this chunk to esplora/mod.rs and removing from `reqwest.rs` and `ureq.rs` and tests run fine with these commands:
```
cargo test --no-default-features --features test-esplora,use-esplora-ureq esplora::bdk_blockchain_tests
cargo test --no-default-features --features test-esplora,use-esplora-reqwest esplora::bdk_blockchain_tests
```

rajarshimaitra (commented):
tACK. 

Agreed with the concept that `reqwest` user should not have to have `tokio` just for testing.

All the tests passes as expected.

I am observing though, even in a release build `cargo build --release --features=use-esplora-reqwest --no-default-features`, `tokio` dependencies are still in the dependency tree because other dependencies already use it.
Cargo.lock after build:
```
...
[[package]]
name = "tokio"
version = "1.8.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "570c2eb13b3ab38208130eccd41be92520388791207fde783bda7c1e8ace28d4"
dependencies = [
 "autocfg",
 "bytes",
 "libc",
 "memchr",
 "mio",
 "pin-project-lite",
 "tokio-macros",
 "winapi 0.3.9",
]

[[package]]
name = "tokio-macros"
version = "1.3.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "54473be61f4ebe4efd09cec9bd5d16fa51d70ea0192213d754d2d500457db110"
dependencies = [
 "proc-macro2",
 "quote",
 "syn",
]

[[package]]
name = "tokio-native-tls"
version = "0.3.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "f7d995660bd2b7f8c1568414c1126076c13fbb725c40112dc0120b78eb9b717b"
dependencies = [
 "native-tls",
 "tokio",
]

[[package]]
name = "tokio-util"
version = "0.6.7"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "1caa0b0c8d94a049db56b5acf8cba99dc0623aab1b26d5b5f5e2d945846b3592"
dependencies = [
 "bytes",
 "futures-core",
 "futures-sink",
 "log",
 "pin-project-lite",
 "tokio",
]
```

I am not entirely sure whether there is any difference between "hard dependency" and "dependency of a dependency". Just by looking at `Cargo.lock`s from builds of both #433 and #430, it seems like it doesn't matter to our dependency tree. `tokio` is gonna be there in release mode whether we ask for it or not. `reqwest` depends on it (among few other crates).

Another minor concern I have is over code repetition. And there is really no other workaround if we choose not use `await_or_block!()` macro. This might create maintenance burden. But good thing is at least they are in testsutils only.

So far I haven't able to form any hard opinion on either approach. I think we should choose whichever makes our lives easier down the line. Maintaining a consistent async interfaces is a tricky thing for sure.

LLFourn:
I just had a chat with @tcharding  where we tried to come up with a way around the sprawling complexity of having multiple http backends. The result of the discussion was something like why not just have our own `HttpClient` and implement both async and non-async versions. This would just include `get`, `post`  methods and return the body a `Result<Vec<u8>, HttpError>` 

 In blockchain tests we would only use the blocking one and therefore not test request there. However the only logic in the http backend would be making a http request and returning the result so there is no real reason to test this as there is no business logic. 

At this point I think `HttpClient` should be a trait and esplora would use it like `Esplora<C: HttpClient>`. This also enables BYO http client (I know of one person who said they want this). This would make the http client a separate choice for the user to consider which might make the feature flags more coherent anyway.

I think I will take on this work myself starting next week if nobody has any objections.

LLFourn:
> I am observing though, even in a release build `cargo build --release --features=use-esplora-reqwest --no-default-features`, `tokio` dependencies are still in the dependency tree because other dependencies already use it.

note that reqwest has a hard dependency on tokio so this is what you should see.

notmandatory:
I think an `HttpClient` trait in the `esplora` module sounds like a good way decouple things, similar to having a `Database` trait with different backends. In this case will `bdk` still offer optional `esplora-ureq` or `esplora-reqwest` features? or will we just include the `ureq` as dev-dependencies for the esplora blockchain tests?  and in this case an end user (like `bdk-cli`) would implement it's own `ureq` based (or which ever) `HttpClient`? 

One other thing to keep in mind for the `HttpClient` trait is to support socks5 proxies, see #429.

LLFourn:
> 
> 
> I think an `HttpClient` trait in the `esplora` module sounds like a good way decouple things, similar to having a `Database` trait with different backends. In this case will `bdk` still offer optional `esplora-ureq` or `esplora-reqwest` features? or will we just include the `ureq` as dev-dependencies for the esplora blockchain tests? and in this case an end user (like `bdk-cli`) would implement it's own `ureq` based (or which ever) `HttpClient`?

My idea would be to have a `http-client-blocking` and `http-client-async` features and `ureq` and `reqwest` would not be re-exported anywhere at all in the api.  bdk-cli would not have to worry at all about this (much) but downstream crates would have to have their own http dependency if they need to make http requests which I think is fine.

> 
> One other thing to keep in mind for the `HttpClient` trait is to support socks5 proxies, see #429.

yes

tcharding:
Converted this to draft while development of the proposed `HttpClient` takes place. Cheers.

rajarshimaitra:
Since this is converted into draft can we have #430 merged for the mean time? I have some downstream bdk-cli PRs pending on the esplora fix.

tcharding:
> Since this is converted into draft can we have #430 merged for the mean time? I have some downstream bdk-cli PRs pending on the esplora fix.

Go for it

tcharding:
I think we can close this now, no further progress or discussion on this PR seems necessary.

PR #435, opened 2021-09-05
Add waste metric for coin selection

benthecarman:
Implementation of the coin selection waste metric from https://github.com/bitcoin/bitcoin/pull/22009

Later this can be used in the wallet to select the most optimal coin selection algo.

I plan on adding more of the tests but wanted to get this out there as I am new to rust and I am guessing I have some things I need to fix.

### DIFF
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -534,6 +534,49 @@ impl BranchAndBoundCoinSelection {
     }
 }
 
+struct WasteMetric;
+
+impl WasteMetric {
+    fn calculate_waste(
+        selected: Vec<WeightedUtxo>,
+        cost_of_change: Option<u64>,
+        amount_needed: u64,
+        fee_rate: FeeRate,
+        long_term_fee_rate: FeeRate,
+    ) -> i64 {
+        // Always consider the cost of spending an input now vs in the future.
+        let utxo_groups: Vec<OutputGroup> = selected
+            .into_iter()
+            .map(|u| OutputGroup::new(u, fee_rate))
+            .collect();
+
+        let waste: i64 = utxo_groups.iter().fold(0, |acc, utxo| {
+            let fee: i64 = utxo.fee as i64;
+            let long_term_fee: i64 = long_term_fee_rate
+                .fee_wu(TXIN_BASE_WEIGHT + utxo.weighted_utxo.satisfaction_weight)
+                as i64;
+
+            acc + fee - long_term_fee
+        });
+
+        let selected_effective_value: i64 = utxo_groups
+            .iter()
+            .fold(0, |acc, utxo| acc + utxo.effective_value);
+
+        match cost_of_change {
+            Some(change_cost) => {
+                // Consider the cost of making change and spending it in the future
+                // If we aren't making change, the caller should've set change_cost to 0
+                waste + change_cost as i64
+            }
+            None => {
+                // When we are not making change (change_cost == 0), consider the excess we are throwing away to fees
+                waste + selected_effective_value - amount_needed as i64
+            }
+        }
+    }
+}
+
 #[cfg(test)]
 mod test {
     use std::str::FromStr;
@@ -618,6 +661,26 @@ mod test {
             }),
         };
         vec![utxo; utxos_number]
+    }
+
+    fn generate_utxos_of_values(utxos_values: Vec<u64>) -> Vec<WeightedUtxo> {
+        utxos_values
+            .into_iter()
+            .map(|value| WeightedUtxo {
+                satisfaction_weight: P2WPKH_WITNESS_SIZE,
+                utxo: Utxo::Local(LocalUtxo {
+                    outpoint: OutPoint::from_str(
+                        "ebd9813ecebc57ff8f30797de7c205e3c7498ca950ea4341ee51a685ff2fa30a:0",
+                    )
+                    .unwrap(),
+                    txout: TxOut {
+                        value: value,
+                        script_pubkey: Script::new(),
+                    },
+                    keychain: KeychainKind::External,
+                }),
+            })
+            .collect()
     }
 
     fn sum_random_utxos(mut rng: &mut StdRng, utxos: &mut Vec<WeightedUtxo>) -> u64 {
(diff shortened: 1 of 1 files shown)


src/wallet/coin_selection.rs
```diff
@@ -534,6 +534,49 @@ impl BranchAndBoundCoinSelection {
     }
 }
 
+struct WasteMetric;
```
notmandatory, line 537: This looks like a good direction for improving coin selection, I haven't looked deeply into the code/algorithm but I do have a couple rust style suggestions.

1. Unless you plan on adding additional fields to the `WasteMetric` struct I think you can do without it.  
2. And if the idea for `calculate_waste` is as a helper function to implementations of the `CoinSelectionAlgorithm` trait then it may be better to make `calculate_waste` a "default implementation" on the `CoinSelectionAlgorithm` trait.

If you take this approach then possibly `CoinSelectionResult` could include a waste metric field.

benthecarman: The waste metric requires a long term fee rate (expected average fee rate of the wallet) so idk if it can be directly be added to `CoinSelectionResult`.

src/wallet/coin_selection.rs
```diff
@@ -534,6 +534,49 @@ impl BranchAndBoundCoinSelection {
...
+
+impl WasteMetric {
+    fn calculate_waste(
+        selected: Vec<WeightedUtxo>,
+        cost_of_change: Option<u64>,
+        amount_needed: u64,
+        fee_rate: FeeRate,
+        long_term_fee_rate: FeeRate,
+    ) -> i64 {
```
rajarshimaitra, line 546: Instead of returning `i64` why not return just return `Self`? And have `i64` or whatever as member of `WasteMetric`. If we are defining a struct anyway, lets use it. :) 

In this way we can also use this inside `CoinSelectionResult` later. This seems to me like a very nice tool to compare different selection algorithm and we can expose a simple API to users with some default configs (like long term fee rate etc).

rajarshimaitra (commented):
Thanks for the pull.

tested and concept ACK.

Can you please provide some more doc details within the code on waste metric? Is it used by core too? I would like to learn more about the rationale behind the waste calculations and see how that can be used to make selction-algo choices efficient.

notmandatory:
Some good suggestions came up in our team chat today from @LLFourn and @afilini about his PR. 
* your `calculate_waste` function could be implemented on the `CoinSelectionResult` struct, which already contains the `.selected` utxos. 
* then it would be straight forward to create a function (maybe on Wallet?) that does `coin_select()` for different  `CoinSelectionAlgorithm` implementations and returns the optimal result.

Also @rajarshimaitra reminded us that [tomorrows PR review club](https://bitcoincore.reviews/22009) will be on this topic, so good for anyone interested to attend.

notmandatory:
moving this one to DRAFT until we figure out how to integrate it.

csralvall:
Hey @benthecarman, are you still working on this? May I continue your work?

benthecarman:
> Hey @benthecarman, are you still working on this? May I continue your work?

Feel free to!

csralvall:
> Some good suggestions came up in our team chat today from @LLFourn and @afilini about his PR.
> 
> * your `calculate_waste` function could be implemented on the `CoinSelectionResult` struct, which already contains the `.selected` utxos.
> * then it would be straight forward to create a function (maybe on Wallet?) that does `coin_select()` for different  `CoinSelectionAlgorithm` implementations and returns the optimal result.
> 
> Also @rajarshimaitra reminded us that [tomorrows PR review club](https://bitcoincore.reviews/22009) will be on this topic, so good for anyone interested to attend.

Although `CoinSelectionResult` may be related to the `calculate_waste` function, and the shared selected parameter makes it more natural to think so, I found some drawbacks in this approach.
Schmidty noted in the PR review club that there were two waste metrics in use currently in _Bitcoin Core_. The one computed by the function `GetSelectionWaste` and the one used by BnB to improve the selection.
Currently, `GetSelectionWaste` in _Bitcoin Core_ master branch, is implemented as an independent function, not an associated one nor a method, and `SelectionResult` implements a public wrapper method around it to make comparisons between different algorithms. This approach allows the future refactoring of the BnB to deduplicate the calculation of the waste metric and, at the same time, the current comparison of the coin selection algorithms.
I suggest to follow the implementation of Bitcoin Core. To make `calculate_waste` an independent function, with a public wrapper (e.g.: `get_waste`) method around it in the CoinSelectionResult. This will give us the same flexibility mentioned above.

rajarshimaitra:
@csralvall That makes sense.. Eventually we would want to use the waste function to compare potential wastes between different coin selections.. It makes sense that it can be different than what BnB uses internally.. 

I think I understand the approach you are suggesting, although would be more clear when I see the code.. Feel free to update the PR (or make new one) and I would review..

src/wallet/coin_selection.rs
murchandamus: Just to clarify, the "Long Term Feerate Estimate" (LTFRE) is not the average feerate on the wallet, it's more of a minimum feerate that you expect to be able to spend at in the future. The thinking is that if you expect to be able to spend an input at a lower feerate in the future, you should delay spending it to then, hence, anything above that feerate is considered "wasteful". Since you always need to use at least one input, two solutions that each use one input of the same type will see their score increase by the same amount, so it doesn't affect their prioritization. However, if you e.g. have different output types or compare solutions with multiple inputs, you'll prefer the smaller input weight per this heuristic.

murchandamus:
> Can you please provide some more doc details within the code on waste metric? Is it used by core too? I would like to learn more about the rationale behind the waste calculations and see how that can be used to make selction-algo choices efficient.

I recently wrote a blog post about this, let me know if you have open questions, I'd be happy to incorporate feedback: https://murch.one/posts/waste-metric/

> It makes sense that it can be different than what BnB uses internally.

Yeah, that's actually a good point. So far, I don't see a reason why the two would diverge, but another contributor recently suggested that we could introduce additional heuristics based on privacy and confirmation-reliability to aid in prioritization of SelectionResults. These additional heuristics might only apply to the choice between the solutions from different algorithms rather than internally to BnB.

danielabrozzoni:

(Specifically, the coin selection is being rewritten from scratch)

PR #438, opened 2021-09-15
Fix `extract_policy()` on `pk_h()` operands

afilini:
### Description

It looks like we didn't have any test covering the `pk_h()` operand in our policy module, and sure enough there was a bug in there that basically made it work only with WIF keys, not extended keys.

While fixing this issue I noticed that:

1. We forgot to update the `descriptor!()` macro when we switched to the `IntoDescriptorKey` trait, so the `pk_h()` operand was only working with `DescriptorPublicKey` but not other key types that implement the conversion trait, so I fixed it as well.
2. The macro also didn't support `andor()` because we call it `and_or()`. To avoid breaking existing code I just added `andor()` as an alias, so both can be used. 

Finally I added a test that calls `extract_policy()` on a descriptor that contains `pk_h()`.

This issue was reported on discord: https://discord.com/channels/753336465005608961/753336465005608964/886978803690246206

### Notes to the reviewers

None I guess

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/descriptor/dsl.rs
+++ b/src/descriptor/dsl.rs
@@ -571,8 +571,9 @@ macro_rules! fragment {
     ( pk ( $key:expr ) ) => ({
         $crate::fragment!(c:pk_k ( $key ))
     });
-    ( pk_h ( $key_hash:expr ) ) => ({
-        $crate::impl_leaf_opcode_value!(PkH, $key_hash)
+    ( pk_h ( $key:expr ) ) => ({
+        let secp = $crate::bitcoin::secp256k1::Secp256k1::new();
+        $crate::keys::make_pkh($key, &secp)
     });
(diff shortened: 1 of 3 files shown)


notmandatory (approved):
ACK 326b64d ... but my comprehensions of these macros is limited. If anyone else want's to review please let me know, otherwise I'll merge as is.

notmandatory:
Is this a candidate for a fix release `0.11.1` ? or can it wait for the `0.12.0` release?

afilini:
I think it can wait, if we didn't notice there was a bug for such a long time it's probably not the biggest priority

decentclock:
I reported this issue on discord - can confirm this PR fixes it! Thank you!

PR #440, opened 2021-09-21
[fix-build] Fix version of zeroize_derive < 1.2.0

LLFourn:
zeroize_derive minor version bump 1.2 introduced a breaking change somehow. Fix it below until we figure out what to do! We already fix `zeroize` so this should not be a big deal.

update: For some reason I have to fix it exactly with `~1.1.0` for it to work otherwise cargo decides to use `1.0.0` which we can't have!

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -113,7 +113,7 @@ jobs:
   
   check-wasm:
     name: Check WASM
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-20.04
     env:
       CC: clang-10
       CFLAGS: -I/usr/include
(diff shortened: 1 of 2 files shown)


notmandatory (approved):
ACK c601ae32713d48b75c9401c4ce1bff8f00b19a3e

notmandatory:
I found the problem, the WASM job runner host `ubuntu-16-04` is no longer supported.
https://github.blog/changelog/2021-04-29-github-actions-ubuntu-16-04-lts-virtual-environment-will-be-removed-on-september-20-2021/

I'm going to investigate if we can get that job working with `ubuntu-20.04` or `ubuntu-latest`.

PR #442, opened 2021-09-23
Add use-serde feature

luckysori:
### Description

As consumers of `bdk` we want to access the `miniscript/use-serde` without adding a direct dependency to `miniscript` (adding a direct dependency means that we must keep its version in sync with that of `bdk`, which can be tricky). 

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Activate `miniscript/use-serde` feature to allow consumers of the library to access it via the re-exported `miniscript` crate.
+
 ## [v0.11.0] - [v0.10.0]
 
 - Added `flush` method to the `Database` trait to explicitly flush to disk latest changes on the db.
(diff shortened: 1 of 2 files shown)


notmandatory:
Since `bdk` already depends on `serde` and for `bitcoin` enables it feature  "use-serde", maybe we should also depend on `miniscript` with it's feature "use-serde" rather than make it a new optional feature? Is there an advantage to only making `serde` optional for `miniscript` ?

luckysori:
> Since `bdk` already depends on `serde` and for `bitcoin` enables it feature "use-serde", maybe we should also depend on `miniscript` with it's feature "use-serde" rather than make it a new optional feature? Is there an advantage to only making `serde` optional for `miniscript` ?

It's very possible that the introduction of the feature is not needed. The only advantage that I expect from having the feature is that someone that doesn't need the `miniscript/use-serde` feature doesn't need to compile extra code.

LLFourn:
+1 just enable miniscript serde by default without feature. We can always add an default `serde` feature later on if things change.

luckysori:
@notmandatory @LLFourn I've heeded your recommendation and updated the PR accordingly.

notmandatory (approved):
ACK eea8b135a477e48ee5d4665ef554355346503151

LLFourn (approved):

Issue #427, opened 2021-08-26
Allow transaction extraction from wallet


rajarshimaitra:
We recently changed the `maintain_single_recipient()` function to `allow_shrinking()`. But `allow_shrinking` takes in a scriptpubkey to identify which output to shrink. One use case of allow shrinking is in bdk-cli https://github.com/bitcoindevkit/bdk-cli/blob/4fd39b7734baa515867d00f6d710c4b2084de02d/src/lib.rs#L799-L801.

IIUC, BDK doesn't have currently any public API to fetch the scriptpubkey (or the transaction) to use in `allow_shrinking`. It seems like an useful function to have. People might wanna get the full transaction detail specific to an ID directly from their wallet.

Creating this issue to maintain discussion on this.

LLFourn:
I think what you're saying is that you can't inspect the tx to find the script pubkey of the output you want to shrink?
There is of course `list_transactions` but you awkwardly have to filter them to find the one you're looking for. The right solution is to be able to directly query the database yourself I think.
I had to add this functionality to my branch of `bdk` for `gun`. See: https://github.com/LLFourn/gun/blob/40eddd9010e5e3e535f3e6e8cef4fafb8cd5316e/src/party/mod.rs#L208

I plan for this change to make its way to main bdk eventually.

rajarshimaitra:
Yes that's exactly what I am looking for. Its little awkward to use list transactions, but it still can be done, and I can use that to handle breaking the situation in `bdk-cli`. Thanks for pointing that out.

Do you think its useful to have a public API to fetch a transaction with ID directly from the wallet? I suppose it would be a very simple function, as most of the codes to do that is available, but are private to the wallet. Only question is weather we would want it in a public API.

LLFourn:
@rajarshimaitra Yes I think `get_tx` on wallet makes sense given how the wallet api already allows inspecting the database.

LLFourn:
This can be closed @notmandatory

PR #443, opened 2021-09-23
Expose get_tx() method from `DB` to `Wallet`

rajarshimaitra:
### Description

Expose `get_tx()` method as discussed in #427.

I am wondering what kind of test can be included for this. 

[Edit]
It probably doesn't need separate test as the underlying wallet method is already tested here https://github.com/bitcoindevkit/bdk/blob/a685b22aa693e6d95a9719287faaa115dc5c0ec0/src/database/mod.rs#L334

#### All Submissions:

#### New Features:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -323,7 +323,7 @@ where
 
     /// Return the list of unspent outputs of this wallet
     ///
-    /// Note that this methods only operate on the internal database, which first needs to be
+    /// Note that this method only operates on the internal database, which first needs to be
     /// [`Wallet::sync`] manually.
     pub fn list_unspent(&self) -> Result<Vec<LocalUtxo>, Error> {
         self.database.borrow().iter_utxos()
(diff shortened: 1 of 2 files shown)


src/wallet/mod.rs
```diff
@@ -335,6 +335,15 @@ where
         self.database.borrow().get_utxo(&outpoint)
     }
 
+    /// Returns the `Tx` details corresponding to `txid` and optionally `raw_tx`
```
notmandatory: For clarity I think it's OK to duplicate most of the docs from the `list_transactions` function:
```suggestion
        /// Return a single transaction made or received by the wallet
        ///
        /// Optionally fill the [`TransactionDetails::transaction`] field with the raw transaction if
        /// `include_raw` is `true`.
        ///
        /// Note that this methods only operate on the internal database, which first needs to be
        /// [`Wallet::sync`] manually.
```

rajarshimaitra: Thanks yes that makes sense.. Updated..

notmandatory:
LGTM but needs a rebase to fix CHANGELOG conflict.

rajarshimaitra:
Updated and rebased..

rajarshimaitra:
Not sure why the CI is failing. It passed in my local.

notmandatory:
Looks like the [`zeroize_derive` version 1.1.0](
https://crates.io/crates/zeroize_derive/versions) we recently pinned to fix an error caused by that project changing their MSRV has now been yanked. I'll need to get that fixed in `bdk` before we release our `0.12.0` version.

notmandatory:
The build errors should be fixed once #444 is merged.

LLFourn (approved):

src/wallet/mod.rs
```diff
@@ -335,6 +335,21 @@ where
         self.database.borrow().get_utxo(&outpoint)
     }
 
+    /// Return a single transactions made and received by the wallet
+    ///
+    /// Optionally fill the [`TransactionDetails::transaction`] field with the raw transaction if
+    /// `include_raw` is `true`.
+    ///
+    /// Note that this methods only operate on the internal database, which first needs to be
```
shesek: `s/methods/method/`

notmandatory (approved):
ACK 2f38d3e526244cdb982b08abe9ecdf86af9c2257

PR #444, opened 2021-09-27
Downgrade tiny-bip39 to version prior to 0.8.0

notmandatory:
### Description

This is required until BDK MSRV is changed to 1.51 or we replace the `tiny-bip39` dependency (see #399).

### Notes to the reviewers

We were able to fix this issue with `zeroize_derive` by downgrading it to version `1.1.0` but that version was [yanked from crates.io](https://crates.io/crates/zeroize_derive/versions) two days ago. So now I'm downgrading `tiny-bip39` to a version prior to 0.8.

#### All Submissions:


notmandatory:
I will also cherry-pick this into the `release/0.12.0` branch.

afilini (approved):
Super annoying but I don't see any other obvious way to fix this

PR #448, opened 2021-10-13
Make MemoryDatabase Send + Sync

artfuldev:
### Description

We make the `MemoryDatabase` `Send + Sync`. This is [required](https://github.com/mozilla/uniffi-rs#thread-safety) for language bindings generation using [uniffi-rs](https://github.com/mozilla/uniffi-rs) and is a very small change.

#### All Submissions:

### DIFF
--- a/src/database/memory.rs
+++ b/src/database/memory.rs
@@ -14,6 +14,7 @@
 //! This module defines an in-memory database type called [`MemoryDatabase`] that is based on a
 //! [`BTreeMap`].
 
+use std::any::Any;
 use std::collections::BTreeMap;
 use std::ops::Bound::{Excluded, Included};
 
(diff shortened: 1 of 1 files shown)


LLFourn:
This LGTM. I think the idea here is to make sure that if you `Arc<Mutex<>>` the in memory database it will be send + sync as expected. Am I right?

No idea why these resolver test failures are reappearing in our CI.

artfuldev:
I'm not sure I understand the `Arc<Mutex<>>` part. The only way to make it automatically `Send + Sync` is to ensure the inner-most contained type `Box<dyn Any>` is also `Send + Sync` - that was the only violation, so I added the trait constraint there.

artfuldev:
> No idea why these resolver test failures are reappearing in our CI.

Is this something to be fixed by me?

LLFourn:
> Is this something to be fixed by me?

Nah we should fix this.

artfuldev:
> I think the idea here is to make sure that if you `Arc<Mutex<>>` the in memory database it will be send + sync as expected. Am I right?

Apologies, I think I finally understand what you meant.

yes, wrapping `MemoryDatabase` in a `Mutex` would make it `Send + Sync`, but when we try to use a `Wallet<B, MemoryDatabase>` the Wallet is no longer `Send + Sync` because `MemoryDatabase` isn't.

afilini:
The issue should be fixed now, you can try rebasing on the latest `master`

notmandatory (approved):
ACK 59f795f176061864c25d60b06c48682e3c3695e6.

PR #454, opened 2021-10-22
Fix/ci update 1.56

### LINKED ISSUE #439: Update CI pipelines to latest Rust stable `1.55.0`
The current stable Rust version is now [`1.55.0`](https://blog.rust-lang.org/2021/09/09/Rust-1.55.0.html). We should update our CI pipelines to this version and fix any resulting issues in the code, most likely due to new clippy checks.

afilini:
### Description

Update our CI to use `1.56`, pin some dependencies to maintain our MSRV.

Closes #439.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/blockchain/compact_filters/store.rs
+++ b/src/blockchain/compact_filters/store.rs
@@ -760,7 +760,7 @@ impl CfStore {
         let cf_headers: Vec<FilterHeader> = filter_hashes
             .into_iter()
             .scan(checkpoint, |prev_header, filter_hash| {
-                let filter_header = filter_hash.filter_header(&prev_header);
+                let filter_header = filter_hash.filter_header(prev_header);
                 *prev_header = filter_header;
 
                 Some(filter_header)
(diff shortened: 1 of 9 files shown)


rajarshimaitra (commented):
tACK 

Are those refs changes required for the previous version update changes? Its working ok too without the refs changes.

src/blockchain/compact_filters/store.rs
```diff
@@ -801,7 +801,7 @@ impl CfStore {
             .zip(headers.into_iter())
             .scan(checkpoint, |prev_header, ((_, filter_content), header)| {
                 let filter = BlockFilter::new(&filter_content);
-                if header != filter.filter_header(&prev_header) {
+                if header != filter.filter_header(prev_header) {
```
rajarshimaitra, line 804: Are moving objects instead of passing refs more efficient for these cases for some reason?

afilini: This was actually a clippy error, I'm not sure why the explicit `&` is not needed, but clearly the compiler can infer that it's a reference.

If you tried running clippy locally maybe you forgot to enable the `compact_filters` feature? This would explain why you didn't get anything

notmandatory (approved):
ACK d75d221540a0d47b7538b456211f92cdb4d4da8c

Issue #331, opened 2021-04-17
Support for MSRV shipped with Debian stable for default features 


notmandatory:
For improved build tool security we should support as MSRV the [rustc version shipped with Debian stable], currently `1.41.1`. This may not be ppossible for all features, but should be possible for our wallet with default features, key-value-db and electrum.

Discord discussion here: 
https://discord.com/channels/753336465005608961/753336465005608964/832682901950431323

If other features can also be made to work with a lower MSRV we should create separate issues for those. Another approach for libraries with higher MSRV may be to create a guide for verifying dependencies and/or creating reproducible builds.

[rustc version shipped with Debian stable]: https://tracker.debian.org/pkg/rustc

notmandatory:
The current `testing` debian release is 11 (bullseye) and it will ship with  `rustc 1.48.0`. I think it's worth considering that bullseye may be the `stable` debian by the time any mainnet ready projects adopt BDK. 

https://tracker.debian.org/pkg/rustc

notmandatory:
Something else to note is there appear to be CVEs that affect `rustc` versions prior to `1.52.0`. It looks like the debian `rustc` maintainer is trying to get the fixes back-ported to `1.48`. :crossed_fingers: 

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=986803****

[CVE-2021-28875](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-28875)
[CVE-2021-28876](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-28876)
[CVE-2021-28877](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-28877)
[CVE-2021-28878](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-28878)
[CVE-2021-28879](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-28879)
[CVE-2020-36317](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-36317)
[CVE-2020-36318](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-36318)

LLFourn:
[...]

PR #457, opened 2021-10-25
Update MSRV to 1.51, enable cargo resolver = "2"

notmandatory:
### Description

This is required to keep using dependencies that have switched to MSRV 1.51 which requires the resolver version 2: tiny-bip39, ahash, and object.

### Notes to the reviewers

Current Debian stable (bullseye) includes Rust 1.48.0, and testing (bookworm) includes Rust 1.50.0. If this is a concern I'm open to other suggestions on how to avoid running into the resolver errors we've been getting  hit with lately. 

Prior to merging this PR our branch checks will need to be updated to require the 1.51.0 tests instead of the ones for 1.48.0.

#### All Submissions:

#### New Features:

#### Bugfixes:

* [x] This pull request breaks the existing API,  **Requires update to Rust 1.51+**

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         rust:
           - 1.56.0 # STABLE
-          - 1.46.0 # MSRV
+          - 1.51.0 # MSRV
         features:
           - default
           - minimal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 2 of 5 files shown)


rajarshimaitra (approved):
tACK.

To me it seems MSRV to 1.51 is the right way to go.

Although interested to know if there's any hidden catch.

afilini:
Concept ACK on bumping the MSRV, but I would discuss about this in the call to see if we can get away with a lower version. The two errors I had to fix for #454 for instance were not related to the resolver, so maybe we can find a compromise to keep using a version that is more widely available without having too many issues.

notmandatory:
Rebased on `master` following #454

notmandatory:
I have a proposal for an updated MSRV policy in #331 and based on decision there will update this PR.

notmandatory:
Closing since #454 and #462 fixes our immediate issues. Will continue discussion on #331.

PR #459, opened 2021-10-26
Store the last sync time and block height

### LINKED ISSUE #413: BDK will select coinbase inputs that are not matured yet
I just did `generatetoaddress` to a wallet address and tried to create a transaction. BDK wallet did it but RPC returned: `"sendrawtransaction RPC error: {\"code\":-26,\"message\":\"bad-txns-premature-spend-of-coinbase, tried to spend coinbase at depth 38\"}"`

### LINKED ISSUE #455: Save the latest block height in the database
I'm working on a fix for #413 and #238: it would be useful for both of them to know the "current height" (to know if the coinbase outputs are mature or in the case of #238 to check if we have reached a given minimum number of confirmations).

Since we might not have a `Blockchain` when running calls like `get_balance`, we should save the block height in the db every time we sync (and maybe the timestamp as well).

This could also be used to, for instance, only sync automatically if the data are `x` minutes old or something like that.

afilini:
### Description

Update the `Database` trait to allow storing the block height and timestamp of the last `sync` of a `Wallet`. This will allow having a rough estimation of the "current" height, even in parts of the code that don't have network access (tx creation, computing the balance, etc), at least in non-airgapped wallets that are periodically synced.

After this is ready I'm planning to work on a PR to close #413, essentially making BDK aware of what can be spent and what is still not "mature" (initially only for coinbase UTXOs which are trivial, later on ideally also for any UTXO that has a timelock, which is a lot trickier).

Closes #455.

### Notes to the reviewers

Opening this in draft mode, since it includes some commits from #454.

Random open points:

1. I was thinking about renaming `ConfirmationTime` because that name doesn't really make sense in this context, but I couldn't come up with a decent alternative. I'm open to suggestions if you have any. 
2. I'm not sure what I've done with the sqlite database is ideal, let me know if you think I could use a better structure. Initially I used to add an entry in the table for every `sync`, and then I would only return the latest one. That seemed pretty inefficient, considering that I was storing stuff that were by definition never read, so I changed it to the current structure.. but this feels more NoSQL-like so I still don't know which one I like best.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/database/keyvalue.rs
+++ b/src/database/keyvalue.rs
@@ -82,6 +82,13 @@ macro_rules! impl_batch_operations {
             Ok(())
         }
 
+        fn set_last_sync_time(&mut self, ct: ConfirmationTime) -> Result<(), Error> {
+            let key = MapKey::LastSyncTime.as_map_key();
+            self.insert(key, serde_json::to_vec(&ct)?)$($after_insert)*;
+
+            Ok(())
+        }
+
         fn del_script_pubkey_from_path(&mut self, keychain: KeychainKind, path: u32) -> Result<Option<Script>, Error> {
             let key = MapKey::Path((Some(keychain), Some(path))).as_map_key();
             let res = self.remove(key);
@@ -167,6 +174,14 @@ macro_rules! impl_batch_operations {
                     Ok(Some(val))
                 }
             }
+        }
+
+        fn del_last_sync_time(&mut self) -> Result<Option<ConfirmationTime>, Error> {
+            let key = MapKey::LastSyncTime.as_map_key();
+            let res = self.remove(key);
+            let res = $process_delete!(res);
+
+            Ok(res.map(|b| serde_json::from_slice(&b)).transpose()?)
         }
     }
 }
@@ -342,6 +357,14 @@ impl Database for Tree {
             .transpose()
     }
 
+    fn get_last_sync_time(&self) -> Result<Option<ConfirmationTime>, Error> {
+        let key = MapKey::LastSyncTime.as_map_key();
+        Ok(self
+            .get(key)?
+            .map(|b| serde_json::from_slice(&b))
+            .transpose()?)
+    }
+
     // inserts 0 if not present
     fn increment_last_index(&mut self, keychain: KeychainKind) -> Result<u32, Error> {
         let key = MapKey::LastIndex(keychain).as_map_key();
@@ -469,5 +492,10 @@ mod test {
     #[test]
     fn test_last_index() {
         crate::database::test::test_last_index(get_tree());
+    }
+
+    #[test]
+    fn test_last_sync_time() {
+        crate::database::test::test_last_sync_time(get_tree());
     }
 }
--- a/src/database/keyvalue.rs
+++ b/src/database/keyvalue.rs
@@ -82,7 +82,7 @@ macro_rules! impl_batch_operations {
             Ok(())
         }
 
-        fn set_last_sync_time(&mut self, ct: ConfirmationTime) -> Result<(), Error> {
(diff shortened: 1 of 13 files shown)


notmandatory:
My rename suggestion for `ConfirmationTime` is simply `Confirmation`and then the field would be `TransactionDetails.confirmation`.

notmandatory:
It looks like this PR has come commits that are already in `master`, can it be rebased to simplify the history?
Don't need:
a348dbdcfe2fa9abc92eab7286dde1177fef65bd
548e43d928d1492784dc465a28fa99b9ca263b9f
548e43d928d1492784dc465a28fa99b9ca263b9f
548e43d928d1492784dc465a28fa99b9ca263b9f

LLFourn:
Can we think about getting rid of the block height request in `Wallet::new` after this and just using the one from the database.

afilini:
> Can we think about getting rid of the block height request in Wallet::new after this and just using the one from the database.

Yup, that would be great. With that you would only need the network to sync, everything else would work offline as well, which is reasonable

src/database/keyvalue.rs
```diff
@@ -82,6 +82,13 @@ macro_rules! impl_batch_operations {
             Ok(())
         }
 
+        fn set_last_sync_time(&mut self, ct: ConfirmationTime) -> Result<(), Error> {
```
LLFourn: So I'm working towards adding different types of syncs. e.g. only sync addresses given out (no stop-gap). Maybe only sync addresses given out recently or only sync external keychain etc.

Could we leave the door open to that by making a `SyncTime` struct which has the blocktime (whose type is unfortunately called `ConfirmationTime`) as a field? It wouldn't hurt to add the local time to it. It's possible we'd want more metadata about a sync in the future and it'd make it easier from a DB updating point of view to serialize it in this way.

LLFourn (changes requested):
LGTM -- one thing I'd like you to consider changing.

artfuldev:
Regarding `ConfirmationTime`, I think we can just use something like `BlockInfo` which has a few details from:
image

Currently, we use the height and timestamp as an indication of confirmation and last sync "time". Maybe we add hash next. And we can refer to it in a confirmed transaction as `block: BlockInfo` / while syncing as `last_synced: BlockInfo`.

src/database/keyvalue.rs
afilini: I tried implementing this but I got stuck because we don't have a way to get the tip hash with the current `Blockchain` trait.

Since I would have to make a breaking change to the trait I'm thinking I might as well split it up like you described during our last call.

I'm not sure if we should hold this PR until that's ready or just go ahead and merge it without block_hash and add that later on.

LLFourn: I think merge without that now and follow up with a PR that adds `Option<BlockHash>` to the struct. I still think that we don't want to add this to `BlockTime` but rather a struct like `SyncTime` that contains a `BlockTime`. Also `SyncTime` should probably contain the UTC time that the sync was finished (in this PR).

afilini: Yeah in my local branch I had a separate structure that contained the `BlockTime` and the block hash.

I will add just the UTC time for now, we'll add the block hash in a following PR

LLFourn: You might consider adding a field for it in the sqlite table otherwise we have to do a table update and I'm not sure we have a db upgrade plan yet :S?

notmandatory:
Looks like it needs a rebase to fix the CHANGELOG then is ready to go.

afilini:
Done!

notmandatory (approved):
ACK 5830226216218f2b3ce89206c58fb24c88ef8c9a

PR #166, opened 2020-11-16
Sync

RCasatta:
### This PR contains:

* **change sync logic**: biggest change in the sync algorithm is to drop the "clever but problematic" handling of the descendant transaction, see old fn `check_tx_and_descendant`. The rationale of the current algorithm is that internal address contains funds only in our transactions so you don't initially ask about them to the electrum server but check those in transactions retrieved from external addresses and only if present asks for spending tx. Doing so is problematic in case we have a deep change chain because these transactions must be asked serially. The algorithm in this PR simply asks for internal address as it's done for external so we can batch requests more, improving speed.
* **add timestamps to transactions**: at the moment timestamp field of transactions is not populated, this PR download transaction relative block headers to populate the timestamp
* **remove `listunspent` calls**: closes https://github.com/bitcoindevkit/bdk/issues/140, spent information is in our history, so we can avoid to ask this endpoint.
* `repl` example can use esplora so that testing this blockchain source is easier

### Results:

|wallet_name | max_addresses | branch | first sync | following sync
|-|-|-|-|-|
|change_deep | 200 | [master](ac18fb119f267c9e9349cb9cb984d14ea615e281) | 19.0s | 9.0s 
|change_deep | 200 | [sync](c4c03d6d7f4296ff5bbb3b7d8f12413f1e8dc5ec) | 2.3s | 1.2s 
|single_many_utxos | 200 | master | 16.0s | 2.0s 
|single_many_utxos | 200 | sync | 2.0s | 1.0s

### Future improvements:

* **Error handling on connection reset**. At the moment rate limiting on some electrum server (like esplora) could prevent big wallet (2000+ txs) to sync, we should handle connection failure and implement some exponential backoff time to keep servers happy
* At the moment Esplora API **miss raw blockheader** endpoint in production, now it has been updated https://github.com/Blockstream/electrs/commit/91076cff8bef143922abe455031ff80424de4d9c but not yet deployed (current code build the blockheader from a block header json)
* **Get rid of `max_addresses`** in electrum-like syncing mechanism: change cache addresses mechanism to calculate the addresses every team previous stop_gap batch contains txs (`max_addresses` could still be useful for parallelism in blockchain source like compact filters)
* **improve esplora concurrency** so that wasm wallets could benefit
* use rust-miniscript with secp context to speed up address derivations
* handle fee capabilities, avoid fetching previous txs if requested fee capability is not accurate

change_deep wallet:
```
[...]

PR #461, opened 2021-10-28
Restructure electrum/esplora sync logic

LLFourn:
## Description

This PR does dependency inversion on the previous sync logic for electrum and esplora captured in the trait `ElectrumLikeSync`. This means that the sync logic does not reference the blockchain at all. Instead the blockchain asks the sync logic (in `script_sync.rs`) what it needs to continue the sync and tries to retrieve it.

The initial purpose of doing this is to remove invocations of `maybe_await` in the abstract sync logic in preparation for completely removing `maybe_await` in the future. The other major benefit is it gives a lot more freedom for the esplora logic to use the rich data from the responses to complete the sync with less HTTP requests than it did previously.

## List of changes

- sync logic moved to `script_sync.rs` and `ElectrumLikeSync` is gone.
- esplora makes one http request per sync address. This means it makes half the number of http requests for a fully synced wallet and N*M less requests for a wallet which has N new transactions with M unique input transactions.
- electrum and esplora save less raw transactions in the database. Electrum still requests input transactions for each of its transactions to calculate the fee but it does not save them to the database anymore.
- The ureq and reqwest blockchain configuration is now unified into the same struct. This is the only API change. `read_timeout` and `write_timeout` have been removed in favor of a single `timeout` option which is set in both ureq and reqwest.
- ureq now does concurrent (parallel) requests using threads.
- An previously unnoticed bug has been fixed where by sending a lot of double spending transactions to the same address you could trick a bdk Esplora wallet into thinking it had a lot of unconfirmed coins. This is because esplora doesn't delete double spent transactions from its indexes immediately (not sure if this is a bug or a feature). A blockchain test is added for this.
- BONUS: The second commit in this PR fixes the feerate calculation for esplora and adds a test (the previous algorithm didn't work at all). I could have made a separate PR but since I was touching this file a lot I decided to fix it here.

## Notes to the reviewers

- The most important thing to review is the the logic in `script_sync.rs` is sound.
- Look at the two commits separately.
- I think CI is failing because of MSRV problems again!
- It would be cool to measure how much sync time is improved for your existing wallets/projects. For `gun` the speed improvements for modest but it is at least hammering the esplora server much less.
- I noticed the performance of reqwest in blocking is much worse in this patch than previously. This is because somehow reqwest is not re-using the connection for each request in this new code. I have no idea why. The plan is to get rid of the blocking reqwest implementation in a follow up PR.

#### All Submissions:


src/blockchain/utils.rs
```diff
@@ -23,366 +23,3 @@ use crate::database::{BatchDatabase, BatchOperations, DatabaseUtils};
...
-
-        // remove any spent utxo
-        for new_tx in new_txs.iter() {
-            for input in new_tx.input.iter() {
-                batch.del_utxo(&input.previous_output)?;
-            }
-        }
-
-        db.commit_batch(batch)?;
-        info!("finish setup, elapsed {:?}ms", start.elapsed().as_millis());
```
RCasatta, line 186: I would like to have the logging of the elapsed time for sync also in the new version (at least in electrum I don't see this)

src/blockchain/electrum.rs
```diff
@@ -71,10 +71,151 @@ impl Blockchain for ElectrumBlockchain {
...
+                                    (tx.tx_hash, tx_height)
+                                })
+                                .collect()
+                        })
+                        .collect();
+
+                    script_req.satisfy(txids_per_script)?
+                }
+
+                Request::Conftime(conftimereq) => {
```
RCasatta: nit: maybe `conftime_req` like `script_req`

src/blockchain/script_sync.rs
```diff
@@ -0,0 +1,367 @@
...
+
+struct State<'a, D> {
+    db: &'a D,
+    last_active_index: HashMap<KeychainKind, usize>,
+    tx_needed: VecDeque<Txid>,
+    conftime_needed: VecDeque<Txid>,
+    observed_txs: Vec<TransactionDetails>,
+}
+
+/// A reqeust for on-chain information
```
RCasatta: typo: request

src/blockchain/electrum.rs
```diff
@@ -71,10 +71,151 @@ impl Blockchain for ElectrumBlockchain {
...
+                    let needs_block_height = conftimereq
+                        .request()
+                        .filter_map(|txid| txid_to_height.get(txid).cloned())
+                        .filter(|height| block_times.get(height).is_none())
+                        .take(chunk_size)
+                        .collect::<HashSet<_>>();
```
RCasatta: since there are high probabilities of duplicates, this HashSet is likely smaller than `chunk_size`, I think the code should ensure to take `chunk_size` distinct elements

src/blockchain/electrum.rs
```diff
@@ -71,10 +71,151 @@ impl Blockchain for ElectrumBlockchain {
...
+                }
+                Request::Tx(txreq) => {
+                    let needs_block_height = txreq
+                        .request()
+                        .filter_map(|txid| txid_to_height.get(txid).cloned())
+                        .filter(|height| block_times.get(height).is_none())
+                        .take(chunk_size)
+                        .collect::<HashSet<_>>();
+
+                    let new_block_headers =
```
RCasatta: feels weird to request block headers in the `Request::Tx`, I guess it's problematic to do the `Tx` step before the `Conftime`?

src/blockchain/script_sync.rs
```diff
@@ -0,0 +1,367 @@
...
+    }))
+}
+
+pub struct ScriptReq<'a, D: BatchDatabase> {
+    state: State<'a, D>,
+    script_index: usize,
+    scripts_needed: VecDeque<Script>,
+    stop_gap: usize,
+    keychain: KeychainKind,
+    next_keychains: Vec<KeychainKind>,
```
RCasatta, line 56: I wonder why this can't be `Option<KeychainKind>`?

src/blockchain/esplora/reqwest.rs
```diff
@@ -196,71 +268,27 @@ impl UrlClient {
...
-        Ok(result)
+        last_seen: Option<Txid>,
+    ) -> Result<Vec<Tx>, EsploraError> {
+        let script_hash = sha256::Hash::hash(script.as_bytes()).into_inner().to_hex();
+        let url = match last_seen {
+            Some(last_seen) => format!(
+                "{}/scripthash/{}/txs/chain/{}",
+                self.url, script_hash, last_seen
+            ),
+            None => format!("{}/scripthash/{}/txs", self.url, script_hash),
```
RCasatta, line 282: This esplora endpoint is very informative, the downside is that we always ask for full information even we may already have it. I wonder if we can use the endpoint without `/txs` to have a faster response from the server and request the data only if we recognize it has changed. The first sync will be slower, but the following should be faster and use less bandwith

RCasatta (commented):
Looks great to me!

I did some tests with the test wallet I used in https://github.com/bitcoindevkit/bdk/pull/166 and performance are much better in esplora;
* before first:16s following:10s
* now: 6s

electrum performance is similar to master,  just slightly less now I think because of the HashSet containing less than chunk_size element

src/blockchain/utils.rs
LLFourn: Yes good idea. I always just use `dbg` for debugging so I don't naturally log stuff. Will add some logging in. Might also try and make "progress" work.

src/blockchain/electrum.rs
LLFourn: Good point. This was a little lazy from me. I'll attempt to do it properly.

src/blockchain/electrum.rs
LLFourn: It's not problematic to do the `Tx` step before `Conftime`. It's an arbitrary choice.

Your suggestion is good. Why not remove conftime request from the `Tx` step and put it into the `Conftime` step and do `Tx` first. Might make `script_sync.rs` a little more complicated but from the consumer's perspective it'd certainly be simpler.
I'll give it a shot.

src/blockchain/script_sync.rs
LLFourn: Just in case we invent new keychains. A new keychain could be added without changing sync logic.

src/blockchain/esplora/reqwest.rs
LLFourn: Interesting point. I don't think we have a better endpoint for finding txs assoiciated with an address? I think there's certainly a lot of room for avoiding this when not doing a full sync. So for the approach I'd say we introduce a partial sync. For the case of esplora there a bunch of other things you can do like asking about `outspend` to see if something has been spent or using `/scripthash/:hash/utxo` to see if an address has new utxos.

Another idea is you could even have a mode where a wallet doesn't keep track of all of its tx history but only keeps track of current utxos.

src/blockchain/electrum.rs
LLFourn: I did this. Much nicer.

src/blockchain/electrum.rs
LLFourn: Done.

src/blockchain/utils.rs
LLFourn: I've added helpful logging in `script_sync.rs` so you get it regardless of which blockchain you're using. Still TODO on the progress thing.

src/blockchain/script_sync.rs
RCasatta: ok

src/blockchain/esplora/reqwest.rs
RCasatta: >  I don't think we have a better endpoint for finding txs associated with an address?

I was imagining `GET /scripthash/:hash` could give hints if a call to `GET /scripthash/:hash/txs` is needed and be faster cause it doesn't contain txs information.

However, I think this and the different sync types you are suggesting can be investigated in other PRs since this is already a big improvement in comparison to what we have

src/blockchain/utils.rs
LLFourn: So I think I'm going to leave the "progress" to a follow up PR as it's slightly more involved than I expected.

src/blockchain/script_sync.rs
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    pub fn satisfy(
+        mut self,
+        tx_details: Vec<(Vec<Option<TxOut>>, Transaction)>,
+    ) -> Result<Request<'a, D>, Error> {
+        let tx_details: Vec<TransactionDetails> = tx_details
+            .into_iter()
+            .zip(self.state.tx_needed.iter())
+            .map(|((vin, tx), txid)| {
```
rajarshimaitra: Can we call it `((vout, tx), txid)`? Given `vin` here is `Vec<Option<TxOut>>` and its referred as  a `txout` variable in the next loop.

src/blockchain/script_sync.rs
```diff
@@ -0,0 +1,408 @@
...
+                        match (old_height, height) {
+                            (None, Some(_)) => {
+                                // It looks like the tx has confirmed since we last saw it -- we
+                                // need to know the confirmation time.
+                                self.tx_conftime_interested.insert(*txid);
+                            }
+                            (Some(old_height), Some(new_height)) if old_height != *new_height => {
+                                // The height of the tx has changed !? -- get the confirmation time.
+                                self.tx_conftime_interested.insert(*txid);
+                            }
```
rajarshimaitra: Shouldn't we push the tx into `observed_txs` list in these two cases?

src/blockchain/script_sync.rs
```diff
@@ -0,0 +1,408 @@
...
+                            (Some(_), None) => {
+                                details.confirmation_time = None;
+                                self.state.observed_txs.push(details);
+                            }
```
rajarshimaitra: Does this mean the tx as reorged? Or something unexpected? A comment here like other cases would help.

src/blockchain/utils.rs
```diff
@@ -23,366 +23,3 @@ use crate::database::{BatchDatabase, BatchOperations, DatabaseUtils};
 use crate::error::Error;
 use crate::types::{ConfirmationTime, KeychainKind, LocalUtxo, TransactionDetails};
 use crate::wallet::time::Instant;
-use crate::wallet::utils::ChunksIterator;
-
-#[derive(Debug)]
```
rajarshimaitra, line 28: It seems all the utilities are gone. Should the module be removed all together?

src/blockchain/electrum.rs
```diff
@@ -71,10 +71,139 @@ impl Blockchain for ElectrumBlockchain {
...
+                                    (tx.tx_hash, tx_height)
+                                })
+                                .collect()
+                        })
+                        .collect();
+
+                    script_req.satisfy(txids_per_script)?
+                }
+
+                Request::Conftime(conftime_req) => {
```
rajarshimaitra, line 114: Because in the logic of `script_sync` we will generally emit `Request::Tx` before `Request::ConfTime` it might make sense to follow that pattern in the match logic here too?

src/blockchain/esplora/mod.rs
```diff
@@ -107,10 +96,50 @@ impl fmt::Display for EsploraError {
...
+    /// create a config with default values given the base url and stop gap
+    pub fn new(base_url: String) -> Self {
+        Self {
+            base_url,
+            proxy: None,
+            timeout: None,
+            stop_gap: 20,
+            concurrency: None,
+        }
+    }
```
rajarshimaitra: We are setting a default value of 20 for stop gap here. Maybe we should make that configurable, as per the comment above the function def line? 

Or if we choose to use a specific default, may be its better to configure it like a `DEFAULT_STOP_GAP` const variable, with `stop_gap: Option<usize>` in `EsploraBlockchainConfig`?

src/blockchain/esplora/reqwest.rs
```diff
@@ -98,11 +94,91 @@ impl Blockchain for EsploraBlockchain {
...
+                            let mut related_txs: Vec<Tx> =
+                                self.url_client._scripthash_txs(script, None).await?;
+
+                            let n_confirmed =
+                                related_txs.iter().filter(|tx| tx.status.confirmed).count();
+                            // esplora pages on 25 confirmed transactions. If there's more than
+                            // 25 we need to keep requesting.
+                            if n_confirmed >= 25 {
+                                loop {
+                                    let new_related_txs: Vec<Tx> = self
```
rajarshimaitra: I am a little confused here. If esplora only returns max 25 conf txs, how would we get `n_confirmed => 25` here? The doc [here](https://github.com/Blockstream/esplora/blob/master/API.md#get-scripthashhashtxs) says more can be accessed with `last_seen`, but even if there are more than 25 conf txs, it seems the conf_count in a single response will be never more than 25, so, the loop won't trigger in that case?

There is also a max limit of 50 in-mempool txs/per page, as per the doc. Is that constrain also handled somehow within this logic?

Maybe instead of checking against `conf_count`, if we check for `total_count` (like its done in the inner loop), we can avoid this? Among the two constraints (25 for conf, 50 for unconf), The minimum constraint value is 25, so in case of a multi-pager response, the response total can never be less than 25. That can be the gate for asking for more?

src/blockchain/script_sync.rs
```diff
@@ -0,0 +1,408 @@
...
+    pub fn satisfy(
+        mut self,
+        tx_details: Vec<(Vec<Option<TxOut>>, Transaction)>,
+    ) -> Result<Request<'a, D>, Error> {
+        let tx_details: Vec<TransactionDetails> = tx_details
+            .into_iter()
+            .zip(self.state.tx_needed.iter())
+            .map(|((vin, tx), txid)| {
+                debug!("found tx_details for {}", txid);
+                assert_eq!(tx.txid(), *txid);
```
rajarshimaitra, line 175: Maybe we should assert before the log? Also is there any generic guaranties that `requirement` and `satisfaction` lists will always line up? Maybe we should specifically handle that somewhere?

src/blockchain/esplora/reqwest.rs
```diff
@@ -98,11 +94,91 @@ impl Blockchain for EsploraBlockchain {
...
+                    let full_txs = tx_req
+                        .request()
+                        .map(|txid| {
+                            let tx = tx_index.get(txid).expect("must be in index");
+                            (tx.previous_outputs(), tx.to_tx())
+                        })
+                        .collect();
+                    tx_req.satisfy(full_txs)?
+                }
+                Request::Finish(batch_update) => break batch_update,
```
rajarshimaitra, line 175: Similar to the electrum case, can the order of `Request::ConfTime` and `Request::Tx` be reversed to keep parity with sync logic definition?

src/blockchain/esplora/reqwest.rs
```diff
@@ -196,71 +268,27 @@ impl UrlClient {
...
-        }
-
-        Ok(result)
+        last_seen: Option<Txid>,
+    ) -> Result<Vec<Tx>, EsploraError> {
+        let script_hash = sha256::Hash::hash(script.as_bytes()).into_inner().to_hex();
+        let url = match last_seen {
+            Some(last_seen) => format!(
+                "{}/scripthash/{}/txs/chain/{}",
+                self.url, script_hash, last_seen
```
rajarshimaitra, line 280: What if the `last_seen` was an in-mempool transaction, or we need to extend on the unconfirmed txs list which is more than max limit of 50? This end point only gives confirmed txs.

src/blockchain/esplora/reqwest.rs
```diff
@@ -98,11 +94,91 @@ impl Blockchain for EsploraBlockchain {
...
+                            // esplora pages on 25 confirmed transactions. If there's more than
+                            // 25 we need to keep requesting.
+                            if n_confirmed >= 25 {
+                                loop {
+                                    let new_related_txs: Vec<Tx> = self
+                                        .url_client
+                                        ._scripthash_txs(
+                                            script,
+                                            Some(related_txs.last().unwrap().txid),
+                                        )
```
rajarshimaitra, line 124: Let me know if I am missing something. As per the `_scripthash_txs` logic, it seems we are only asking for confirmed txs list in this inner step of the loop? How are we ensuring at this point that all the unconfirmed txs have been fetched?

rajarshimaitra (commented):
Concept + tACK https://github.com/bitcoindevkit/bdk/pull/461/commits/bfd96a5e6ec0524dae3ad3f49996c382c18c00c2

Overall agreed with the idea of simplifying syncing interface. The current structure of sync logic seems very methodical and easy to extend into other blockchain type. Ran the blockchain tests manually, to see everything is working ok.

Couldn't do sync bench-marking with current master. Will do and report back.

Below are some conceptual questions and nits.

Thanks for working on this, it is a lot of work, and I learned a ton reviewing this.

src/blockchain/esplora/reqwest.rs
LLFourn: when last_seen is set we are always looking for more confirmed transactions.

src/blockchain/esplora/reqwest.rs
LLFourn: you can only have 50 unconfirmed transactions.

src/blockchain/script_sync.rs
LLFourn: > Maybe we should assert before the log? 

I don't think we need to micromanage the placement of this.

> Also is there any generic guaranties that `requirement` and `satisfaction` lists will always line up? Maybe we should specifically handle that somewhere?

I'm not sure there is anything you can do about this. It's a programmer error if they don't line up (hence the assert).

src/blockchain/esplora/reqwest.rs
LLFourn: > I am a little confused here. If esplora only returns max 25 conf txs, how would we get `n_confirmed => 25` here? The doc [here](https://github.com/Blockstream/esplora/blob/master/API.md#get-scripthashhashtxs) says more can be accessed with `last_seen`, but even if there are more than 25 conf txs, it seems the conf_count in a single response will be never more than 25, so, the loop won't trigger in that case?

The comment is wrong but the code is right. It's `>=` so 25 will trigger another loop. I fixed the comment.

> 
> There is also a max limit of 50 in-mempool txs/per page, as per the doc. Is that constrain also handled somehow within this logic?

There is no paging for mempool transactions.

src/blockchain/electrum.rs
LLFourn: The blockchain implementation should not rely on the implementation details of `script_sync`. It has changed during the course of this PR and may change again.  I think I don't keep to that in esplora because I expect the `ScriptReq` to come first. I'll try and fix that.

src/blockchain/script_sync.rs
LLFourn: observed_txs is only for txs that are finished. So I changed the name to `finished_txs`.

src/blockchain/script_sync.rs
LLFourn: done.

src/blockchain/utils.rs
LLFourn: done.

src/blockchain/esplora/mod.rs
LLFourn: You're right it should be a parameter. Fixed.

src/blockchain/esplora/reqwest.rs
LLFourn: See https://github.com/bitcoindevkit/bdk/pull/461#discussion_r743341458

src/blockchain/electrum.rs
LLFourn: On second thought, I think it's not worth doing in this PR since those branches won't be tested yet. When we do partial syncs where tx conf times are requested without having been found by ` script_req` first we can fix this.

rajarshimaitra: Yes makes sense. Can be done on a future PR.

corollary question, given the structure of `tx_needed` and `tx_missing_conftime` in sync state https://github.com/bitcoindevkit/bdk/blob/77266c4a9180df82bf98523df1a0bc7681bb1cc0/src/blockchain/script_sync.rs#L277-L281 

It seems to me that the logical order of requests is always `Script -> Tx -> Conftimes`. Because `Conftime` will need tx details which we can't fetch until we do a `Tx` request on the ID, and `Script` is what we always start with. Is this correct thinking? Or the order can change in different blockchain backend?

src/blockchain/esplora/reqwest.rs
rajarshimaitra: Yes, I have missread the doc for for esplora. Resolved..

src/blockchain/script_sync.rs
rajarshimaitra: Hmm that makes sense. It depends on the client's of sync logic. Also was wondering is there any specific reason for putting the assert only in `TxReq`? It isn't there for other reqs.

src/blockchain/esplora/reqwest.rs
rajarshimaitra: Resolved..

src/blockchain/esplora/reqwest.rs
rajarshimaitra: Sorry for dumb question, but is this 50 a protocol logic? Or something enforced by esplora? Can't I broadcast like 100 donation txs to a single script pubkey?

rajarshimaitra (commented):
ReACK https://github.com/bitcoindevkit/bdk/pull/461/commits/77266c4a9180df82bf98523df1a0bc7681bb1cc0

src/blockchain/electrum.rs
LLFourn: With the current internals yes that will be the order but you could easily change the internals to do it in a different way. The point is the blockchain should no care (ideally).

src/blockchain/script_sync.rs
LLFourn: If you can see another spot where we can do an assert to check consistency let's do it. It's possible I've missed one.

RCasatta (dismissed):
I think it needs a rebase because of the conflict in CHANGELOG, other than that
ACK 77266c4

LLFourn:
Just resolved conflicts with master.

rajarshimaitra:
Hey @LLFourn , any update on the fee optionality? Can we make fees optional in transactional details and have the script-sync not complain when all of the inputs are not found? This will allow to remove `--txindex` if we try to port this into RPC blockchain.

LLFourn:
@rajarshimaitra @RCasatta it seems odd to do that in this PR since the absent `TxOut`s path is not being tested or used here?

rajarshimaitra (approved):
Retested ACK https://github.com/bitcoindevkit/bdk/pull/461/commits/a630685a0ab95604cadbb423f65d8358ba4bc4e4

> @rajarshimaitra @RCasatta it seems odd to do that in this PR since the absent TxOuts path is not being tested or used here?

Yes I think that makes sense. Probably what we want is some way to allow unavailable TxOuts. The only difference would be that fees for those Tx would be absent. We will not miss any utxos because those inputs not found `get_transaction()` are non wallet utxos, so we don't need to keep track of them.

Currently having a missing input causes failure. I think that needs some further investigating and can be done on a future PR.

For now this is good to get in.

RCasatta:
> @rajarshimaitra @RCasatta it seems odd to do that in this PR since the absent TxOuts path is not being tested or used here?

I see from #466 that this logic doesn't fit perfectly with RPC sync, if this cannot be used there, then I do agree it's unnecessary to handle absent TxOuts here

utack a630685a0ab95604cadbb423f65d8358ba4bc4e4

notmandatory:
@LLFourn I'm planning to merge this PR today, thanks for all the work that went into it! and for addressing @RCasatta and @rajarshimaitra 's  review comments. But since this is such a big change I'm going to push it to the 0.15.0 release (0.14.0 is going out today or tomorrow) so it has some time in the `master` branch before we release it.

PR #462, opened 2021-11-01
Use rust-bip39 instead of tiny-bip39

### LINKED ISSUE #399: Switch to a different BIP39 implementation

rajarshimaitra:
### Description

Fixes #399. This can solve the current version conflict issues of `tiny-bip39` dependencies and replace that with `rust-bip39`. 

Few small refactoring changes were required for that.

`rust-bip39` doesn't have dedicated `MnemonicType` enum to denote mnemonic entropy (aka word count). To make the impl consistent with existing traits I have wrote a simple `WordCount` enum to capture that.   

### Notes to the reviewers

All previous tests are passing.

#### All Submissions:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,10 +35,6 @@ socks = { version = "0.3", optional = true }
 lazy_static = { version = "1.4", optional = true }
 
 bip39 = { version = "1.0.1", optional = true }
-# backtrace > 0.3.61 includes object v0.27 which doesn't compile on 1.46. this is used by
-# tiny-bip39
-backtrace = { version = "=0.3.61", optional = true }
-
 bitcoinconsensus = { version = "0.19.0-3", optional = true }
 
 # Needed by bdk_blockchain_tests macro
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,10 +34,7 @@ cc = { version = ">=1.0.64", optional = true }
 socks = { version = "0.3", optional = true }
 lazy_static = { version = "1.4", optional = true }
 
-# the latest 0.8 version of tiny-bip39 depends on zeroize_derive 1.2 which has MSRV 1.51 and our 
-# MSRV is 1.46, to fix this until we update our MSRV or replace the tiny-bip39 
-# dependency https://github.com/bitcoindevkit/bdk/issues/399 we can only use an older version
-tiny-bip39 = { version = "< 0.8", optional = true }
+bip39 = { version = "1.0.1", optional = true }
 # backtrace > 0.3.61 includes object v0.27 which doesn't compile on 1.46. this is used by
 # tiny-bip39
 backtrace = { version = "=0.3.61", optional = true }
@@ -65,7 +62,7 @@ sqlite = ["rusqlite", "ahash"]
 compact_filters = ["rocksdb", "socks", "lazy_static", "cc"]
 key-value-db = ["sled"]
 all-keys = ["keys-bip39"]
-keys-bip39 = ["tiny-bip39", "backtrace"]
+keys-bip39 = ["bip39"]
 rpc = ["core-rpc"]
 
 # We currently provide mulitple implementations of `Blockchain`, all are
--- a/src/keys/bip39.rs
+++ b/src/keys/bip39.rs
@@ -19,7 +19,23 @@ use bitcoin::Network;
 
 use miniscript::ScriptContext;
 
-pub use bip39::{Language, Mnemonic, MnemonicType, Seed};
+pub use bip39::{Language, Mnemonic};
+
+type Seed = [u8; 64];
+
+/// Type describing entropy length (aka word count) in the mnemonic
+pub enum WordsCount {
+    /// 12 words mnemonic (128 bits entropy)
+    Words12 = 128,
+    /// 15 words mnemonic (160 bits entropy)
+    Words15 = 160,
+    /// 18 words mnemonic (192 bits entropy)
+    Words18 = 192,
+    /// 21 words mnemonic (224 bits entropy)
+    Words21 = 224,
+    /// 24 words mnemonic (256 bits entropy)
+    Words24 = 256,
+}
 
 use super::{
     any_network, DerivableKey, DescriptorKey, ExtendedKey, GeneratableKey, GeneratedKey, KeyError,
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -65,7 +62,7 @@ sqlite = ["rusqlite", "ahash"]
 compact_filters = ["rocksdb", "socks", "lazy_static", "cc"]
 key-value-db = ["sled"]
 all-keys = ["keys-bip39"]
-keys-bip39 = ["tiny-bip39", "backtrace"]
+keys-bip39 = ["bip39"]
```
rajarshimaitra, line 61: Question: Instead of this chained features can we make it `all-keys=["bip39"]`?

notmandatory:
Need to make a small fix in `src/keys/mod.rs` at line 463:
```rust
   Mnemonic::parse_in(
        Language::English, "jelly crash boy whisper mouse ecology tuna soccer memory million news short"       
   )?
```
Then all tests should pass.

Cargo.toml
notmandatory: I believe the reason `keys-bip39` is there is in case we switched from `tiny-bip39` to some other bip39 crate (as you're doing now) then end users wouldn't need to change the feature name in their projects. Hopeful we won't need to change bip39 crates again, but just in case, I think it's OK to leave the `keys-bip39` feature here as it is.

src/keys/bip39.rs
```diff
@@ -60,7 +76,7 @@ impl<Ctx: ScriptContext> DerivableKey<Ctx> for Seed {
 impl<Ctx: ScriptContext> DerivableKey<Ctx> for MnemonicWithPassphrase {
     fn into_extended_key(self) -> Result<ExtendedKey<Ctx>, KeyError> {
         let (mnemonic, passphrase) = self;
-        let seed = Seed::new(&mnemonic, passphrase.as_deref().unwrap_or(""));
+        let seed: Seed = mnemonic.to_seed(passphrase.unwrap_or_else(|| "".to_string()));
```
notmandatory: Wouldn't it be simpler this way?
```suggestion
        let seed: Seed = mnemonic.to_seed(passphrase.unwrap_or("".to_string()));
```

notmandatory:
LGTM! just need to fix the little CI issue.

src/keys/bip39.rs
rajarshimaitra: Here `Seed` is just a type definition, not a structure. `rust-bip39` only requires `[u8; 64]` as the seed data. The key methods already had a `Seed` structure (from `tiny-bip39`). So I instead used a type alias. Thus we don't have any `Seed::new()` ctor. 

But agreed on the passphrase deref-dafault. That looks cleaner for `&str`.

Updated.

rajarshimaitra: Done..

notmandatory (approved):
ACK 8f26d3453a621d296119a51f02b4b4c8ba514042

Thanks!

notmandatory:
Ugh, sorry approved but then I noticed you should also be able to remove the `backtrace` dependency since it only had to be pinned because it was used in the `tiny-bip39`.

src/keys/bip39.rs
```diff
@@ -19,7 +19,23 @@ use bitcoin::Network;
 
 use miniscript::ScriptContext;
 
-pub use bip39::{Language, Mnemonic, MnemonicType, Seed};
+pub use bip39::{Language, Mnemonic};
+
+type Seed = [u8; 64];
+
+/// Type describing entropy length (aka word count) in the mnemonic
+pub enum WordCount {
```
artfuldev, line 27: Is there a reason we do not want to name this `MnemonicType`?

artfuldev: Is there a reason we don't want to name this `MnemonicType`?

artfuldev: I ask because it kind of used to exist earlier: https://docs.rs/bip39/0.5.1/bip39/enum.MnemonicType.html

rajarshimaitra: Right point. Reason is, `WordCount` seems more intuitive to me than `MnemonicType`. I had to look it up to see what it was doing. 

`MnemonicType` is an enum used in `tiny-bip39` which does more than simple word or entropy counting. It makes sense for the context of `tiny-bip39` impls, but it seemed too much for `rust-bip39` to use directly. It just takes an `usize` value as "word count".   

So, I decided to make our own simple enum for it and it made sense to give it an easier name.

rajarshimaitra:
@notmandatory thanks, yes that should be removed.

Updated..

rajarshimaitra:
Updated `WordCount` -> `WordsCount`

src/keys/bip39.rs
rajarshimaitra: PS: I think it should be `WordsCount` instead of `WordCount`.

notmandatory: I agree with @artfuldev that `MnemonicType` is less of a code change for downstream users but if you want to change it to something more meaningful I like your original name `WordCount` better.  If we're talking about how many words are in an article or book or something we generally say "the word count is 20" .. not "the words count is 20". See also: https://trends.google.com/trends/explore?q=%22words%20count%22,%22word%20count%22

notmandatory: oops, it's already merged. `WordsCount` will be fine. 😉

rajarshimaitra: Ah, thats a nice google trick to sort this kinda disputes.. 😂 

My bad, I will sneak the change back in some future PR.

rajarshimaitra: BTW that reminds me that this is probably a downstream API change. Even if we kept the name `MnemonicType` the bip39 API still changes. Maybe we should better document the API change somewhere?

notmandatory: Since it does change the API we probably should mention it in the changelog. We can add that in another PR or when we make the next release branch we can add it then.

rajarshimaitra: Updated in https://github.com/bitcoindevkit/bdk/pull/464

Issue #469, opened 2021-11-15
Intermittent errors when combining CompactFiltersBlockchain with MemoryDatabase


ulrichard:
My application runs totally fine with ElectrumBlockchain or RpcBlockchain, but it runs into frequent problems with CompactFiltersBlockchain. I constantly get CompactFilters(DataCorruption) or CompactFilters(Timeout) from wallet.sync(noop_progress(), None);
After I remove the contents of storage_dir, it works the first time, but on the second run, I start getting these errors again.
I also played with compact_filters in bdk-cli, and there everything was totally fine.
My local bitcoind node on localhost is the only one I'm connecting to.
It appears that the main difference is that bdk-cli uses sled database while my application uses MemoryDatabase.
Thus to reproduce the problem, I modified bdk-cli to also use MemoryDatabase in https://github.com/weareseba/bdk-cli/tree/memorydb
When running https://github.com/weareseba/bdk-cli/blob/memorydb/memorydb_compact_filters.sh in this configuration, I can reproduce the problem sometimes, but a lot less frequently than in my application.

rajarshimaitra:
IIRC Unlike other blockchains, comapct filters stores filter data specific to a given wallet descriptor. And adding that with a non-persistent wallet db like memory db can cause problems. At each run memory db will create new wallet descriptors that will not match with previously stored compact filters. 
- This will not happen in sled, because sled will persistently store the wallet data. 
- This won't be an issue for other blockchains because with new descriptor they will happily sync from scratch, and memory db is not persistent, so each run will create a new db, so you won't see any error.

I am not sure what your application is, but I suspect something like that is going on. In general for most of the cases you would want to use sled or any other persistent db, not the memory db.

If you also post the 1) error log, 2) bdk-cli usage comamnd and 3) the changes you have made in bdk-cli, I can try to reproduce it..

ulrichard:
I am working on a cold storage solution where only a subset of the machines has internet access. The descriptors are simple multisig, hence reproduceable. Since those wallets have only few transactions, I thought MemoryDatabase would be enough, so we don't have to care about persistence of additional data. But the system contains many different wallets. So, if we were going to use a persistent database, that would obviously have to be stored separately. And now I am not sure now if I also need to store the data for the compact filters separately per wallet? If so, that could be improved in the documentation.
https://docs.rs/bdk/0.13.0/bdk/blockchain/compact_filters/struct.CompactFiltersBlockchainConfig.html

[...]

PR #471, opened 2021-11-17
moving the function wallet_name_from_descriptor from blockchain/rpc.rs to wallet/mod.rs as it can be useful not only for rpc

ulrichard:
### Description

Moving the function wallet_name_from_descriptor from rpc.rs to mod.rs
Since the local cache for compact filters should be separate per wallet, this function can be useful not only for rpc.

### Notes to the reviewers

I thought about renaming it, but waited for opinions on that.

#### All Submissions:

### DIFF
--- a/src/blockchain/compact_filters/mod.rs
+++ b/src/blockchain/compact_filters/mod.rs
@@ -472,7 +472,7 @@ pub struct CompactFiltersBlockchainConfig {
     pub peers: Vec<BitcoinPeerConfig>,
     /// Network used
     pub network: Network,
-    /// Storage dir to save partially downloaded headers and full blocks
+    /// Storage dir to save partially downloaded headers and full blocks. Should be a separate directory per descriptor. Consider using [crate::wallet::wallet_name_from_descriptor] for this.
     pub storage_dir: String,
     /// Optionally skip initial `skip_blocks` blocks (default: 0)
     pub skip_blocks: Option<usize>,
--- a/src/blockchain/rpc.rs
+++ b/src/blockchain/rpc.rs
@@ -35,8 +35,6 @@ use crate::bitcoin::consensus::deserialize;
 use crate::bitcoin::{Address, Network, OutPoint, Transaction, TxOut, Txid};
 use crate::blockchain::{Blockchain, Capability, ConfigurableBlockchain, Progress};
 use crate::database::{BatchDatabase, DatabaseUtils};
-use crate::descriptor::{get_checksum, IntoWalletDescriptor};
-use crate::wallet::utils::SecpCtx;
 use crate::{BlockTime, Error, FeeRate, KeychainKind, LocalUtxo, TransactionDetails};
 use bitcoincore_rpc::json::{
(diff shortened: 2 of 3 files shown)


notmandatory:
I think the function name is OK, but it may belong in the `wallet` or maybe even the `database` module rather than `blockchain/mod.rs`. I could see using this in `bdk-cli` to generate a default wallet name if one isn't given.

rajarshimaitra (commented):
Concept + code review ACK. Yes it makes sense to put the compact filter data in wallet specific directory.

Inside bdk we are not using this function for any default scenarios. But maybe we should? Especially for default wallets in core. And in light of https://github.com/bitcoindevkit/bdk/issues/469 i think we should also handle unique directory for compact filters data.

Any thoughts on this @notmandatory @RCasatta ? 

PS: That can be handled in a separate PR. And this one will be useful for that impls. 

Also now that we are moving this function to a more generic use case than just wallet name, I think it would be better to rename it more intuitively, like `unique_name_from_descriptor()` or something.

notmandatory:
I think this is ok as is with code and name moving to the wallet mod so it can be used by `rpc` or `cbf` or any other reason to get a unique name for a wallet based on it's descriptor(s).  Maybe a more generic way to do it would be to add some functions in the `descriptor`  mod to take a Vec of `ExtendedDescriptor` and return a unique name.. but I don't see an immediate need.

notmandatory (approved):
ACK  9f8fc74

notmandatory:
@ulrichard sorry one thing I didn't notice, you need to set up commit signing and then `--amend` and re-force push your commit.  If you haven't setup signing with github see [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).

rajarshimaitra:
> I think this is ok as is with code and name moving to the wallet mod so it can be used by rpc or cbf or any other reason to get a unique name for a wallet based on it's descriptor(s).

Just to note here, the use case for the function in RPC and CBF are different. In RPC its the wallet name, in CBF its the directory name of the compact filter data. In CBF wallet name can be something entirely different. I just hope it doesn't add to any confusion.

notmandatory:
> Just to note here, the use case for the function in RPC and CBF are different. In RPC its the wallet name, in CBF its the directory name of the compact filter data. In CBF wallet name can be something entirely different. I just hope it doesn't add to any confusion.

In either case isn't this function still creating a short id for what bdk means by a "wallet" ie. a descriptor and optional change descriptor? In the case of RPC it's used to reference wallet related data in a core node and in CBF it references different data but still wallet related in the local filesystem.

rajarshimaitra:
> In either case isn't this function still creating a short id for what bdk means by a "wallet" ie. a descriptor and optional change descriptor? In the case of RPC it's used to reference wallet related data in a core node and in CBF it references different data but still wallet related in the local filesystem.

Yes I guess that's correct. The purpose is to generate a unique string from a given descriptor. The name is not a big deal as long as the purpose of the function is clear and why its was used.

notmandatory (approved):
re-ACK  2fc8114

ulrichard:
Thanks @notmandatory and @rajarshimaitra.

PR #490, opened 2021-11-30
Add IndexedChain trait and implement for Esplora & Electrum

johncantrell97:
### Description

This PR introduces the trait IndexedChain that Blockchains can implement if they have access to chain data.  There are lots of methods that could be added to this trait (the entire esplora/electrum apis) but I started with the data I needed to provide LDK with the chain data it needs to function.

I have implemented the trait for Esplora and Electrum blockchains.

#### All Submissions:

#### New Features:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bdk"
-version = "0.23.0"
+version = "0.24.0"
 edition = "2018"
 authors = ["Alekos Filini <alekos.filini@gmail.com>", "Riccardo Casatta <riccardo@casatta.it>"]
 homepage = "https://bitcoindevkit.org"


src/blockchain/electrum.rs
```diff
@@ -48,6 +48,63 @@ pub struct ElectrumBlockchain {
     stop_gap: usize,
 }
 
+impl IndexedChain for ElectrumBlockchain {
```
LLFourn: Since each of these queries requires a different underlying index I think it makes sense to have a different trait for each index e.g. `GetBlockHeader`, `GetPositionInBlock` etc.

LLFourn (changes requested):
Concept ACK

One thing I'd like to change.

In addition this should come with tests in a similar manner to how existing blockchain tests work (except use a different macro to generate tests specifically for this trait).

src/blockchain/mod.rs
```diff
@@ -84,6 +84,32 @@ pub enum Capability {
...
 
+/// The height and confirmation status of a transaction
+pub struct TxStatus {
+    /// Whether the transaction has at least one confirmation
+    pub confirmed: bool,
+    /// The height of the block the transaction was confirmed in
+    pub block_height: Option<u32>,
+}
+
+/// A trait Blockchains can implement if they support querying chain data
```
rajarshimaitra: By "support querying chain data" do you mean availability of certain indexing in those blockchain? In that case it might make sense to elaborate the doc a little to explain what are those special requirements. Also how `IndexedChain` is different from regular backends with the `Blockchain` trait.

In general a little explainer doc for this new trait would be very nice.

rajarshimaitra (commented):
Concept ack on breaking down the `Blockchain` into multiple traits. I think this will be eventually needed to make the lib more versatile.

Code changes looks good to me. If the difference between vanilla `Blockchain` and the other types are based on the capabilities (indexes) of different backends, then it might make sense to have such per capability traits as @LLFourn suggested.  

But for now, I think its a good start.

Just one doc request.

notmandatory:
Now that #501 is in you'll need to rebase on `master` to fix the CI issue.

jlopp:
Concept ACK - would love to see this functionality get merged.

LLFourn:
FWIW my plan is **not** to merge this PR or something similar but rather to reduce the surface area that the main BDK lib has with blockchain interactions. I have got feedback that people really want a fully featured rust interface to esplora's http api though. The question to me is do we really need traits in BDK that define this behaviour? It seems like a small thing to leave up to the application developer if they indeed want to have blockchain data polymorphism.

Issue #373, opened 2021-06-17
Support proxies in RPC


afilini:
We support a socks5 proxy for most of the other blockchain backends, we should do that for RPC as well

luisschwab:
Closing in favor of https://github.com/bitcoindevkit/bdk-bitcoind-client/issues/21

PR #493, opened 2021-12-03
Add Socks5 proxy transport for RPC blockchain

rajarshimaitra:
### Description

This attempts to solve https://github.com/bitcoindevkit/bdk/issues/373.

Because `bitcoincore_rpc::Client` doesn't support socks5 proxy internally, the easiest option for us is to implement our own 
socks5 transport and make it compatible with `bitcoincore_rpc::Client`. 

The `Transport` trait is implemented for `ProxyTransport` to make it compatible with core rpc Client. `ProxyTransport` is a simple Socks5 wrapper over `TcpStream` and handles HTTP data like regular connection, but passes it through a given proxy.

I have implemented the basic structure but getting a generic error while trying to connect through local tor proxy running at `127.0.0.1:9050`.

Error
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: RpcClient(JsonRpc(Transport(Io(Custom { kind: Other, error: "general SOCKS server failure" }))))', src/main.rs:77:62
stack backtrace:
```
The sample test code is here https://github.com/rajarshimaitra/bdk-example/blob/main/src/main.rs

I am investigating on the error. 

### Notes to the reviewers

Making the proxy transport probably warrant a dedicated module for rpc. Once the current approach is finalized, that refactoring can be done on a follow up PR. Suggestions welcome.  

#### All Submissions:

#### New Features:

### DIFF
--- a/src/blockchain/rpc_proxy.rs
+++ b/src/blockchain/rpc_proxy.rs
@@ -1,0 +1,241 @@
+// Bitcoin Dev Kit
+// Written in 2021 by Rajarshi Maitra <rajarshi149@gmail.com>
+//
+// Copyright (c) 2020-2021 Bitcoin Dev Kit Developers
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! A SOCKS5 proxy transport implementation for RPC blockchain.
+//!
+//! This is currently internal to the lib and only compatible with
+//! Bitcoin Core RPC.
+
+use super::rpc::Auth;
+use bitcoin::base64;
+use bitcoincore_rpc::jsonrpc::Error as JSONRPC_Error;
+use bitcoincore_rpc::jsonrpc::{Request, Response, Transport};
+use bitcoincore_rpc::Error as RPC_Error;
+use socks::Socks5Stream;
+use std::fs::File;
+use std::io::{BufRead, BufReader, Write};
+use std::time::{Duration, Instant};
+
+/// Errors that can be thrown by [`ProxyTransport`](crate::blockchain::rpc_proxy::ProxyTransport)
+#[derive(Debug)]
+pub enum RpcProxyError {
+    /// Bitcoin core rpc error
+    CoreRpc(bitcoincore_rpc::Error),
+    /// IO error
+    Io(std::io::Error),
+    /// Invalid RPC url
+    InvalidUrl,
+    /// Error serializing or deserializing JSON data
+    Json(serde_json::Error),
+    /// RPC timeout error
+    RpcTimeout,
+    /// Http Parsing Error
+    HttpParsing,
+    /// Http Timeout Error
+    HttpTimeout,
+    /// Http Response Code
+    HttpResponseCode(u16),
+}
+
+impl std::fmt::Display for RpcProxyError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl std::error::Error for RpcProxyError {}
+
+impl_error!(bitcoincore_rpc::Error, CoreRpc, RpcProxyError);
+impl_error!(std::io::Error, Io, RpcProxyError);
+impl_error!(serde_json::Error, Json, RpcProxyError);
+
+// We need to backport RpcProxyError to satisfy Transport trait bound
+impl From<RpcProxyError> for JSONRPC_Error {
+    fn from(e: RpcProxyError) -> Self {
+        Self::Transport(Box::new(e))
+    }
+}
+
+/// SOCKS5 proxy transport
+/// This is currently designed to work only with Bitcoin Core RPC
+pub(crate) struct ProxyTransport {
+    proxy_addr: String,
+    target_addr: String,
+    proxy_credential: Option<(String, String)>,
+    wallet_path: String,
+    rpc_auth: Option<String>,
+    timeout: Duration,
+}
+
+impl ProxyTransport {
+    /// Create a new ProxyTransport
+    pub(crate) fn new(
+        proxy_addr: &str,
+        rpc_url: &str,
+        proxy_credential: Option<(String, String)>,
+        rpc_auth: &Auth,
+    ) -> Result<Self, RpcProxyError> {
+        // Fetch the RPC address:port and wallet path from url
+        let (target_addr, wallet_path) = {
+            // the url will be of form "http://<rpc-host>:<rpc-port>/wallet/<wallet-file-name>"
+            (rpc_url[7..22].to_owned(), rpc_url[22..].to_owned())
+        };
+
+        // fetch username password from rpc authentication
+        let rpc_auth = {
+            if let (Some(user), Some(pass)) = Self::get_user_pass(rpc_auth)? {
+                let mut auth = user;
+                auth.push(':');
+                auth.push_str(&pass[..]);
+                Some(format!("Basic {}", &base64::encode(auth.as_bytes())))
+            } else {
+                None
+            }
+        };
+
+        Ok(ProxyTransport {
+            proxy_addr: proxy_addr.to_owned(),
+            target_addr,
+            proxy_credential,
+            wallet_path,
+            rpc_auth,
+            timeout: Duration::from_secs(15), // Same as regular RPC default
+        })
+    }
+
+    // Helper function to parse username:password pair for rpc Auth
+    fn get_user_pass(auth: &Auth) -> Result<(Option<String>, Option<String>), RpcProxyError> {
+        use std::io::Read;
+        match auth {
+            Auth::None => Ok((None, None)),
+            Auth::UserPass { username, password } => {
+                Ok((Some(username.clone()), Some(password.clone())))
+            }
+            Auth::Cookie { file } => {
+                let mut file = File::open(file)?;
+                let mut contents = String::new();
+                file.read_to_string(&mut contents)?;
+                let mut split = contents.splitn(2, ':');
+                Ok((
+                    Some(split.next().ok_or(RPC_Error::InvalidCookieFile)?.into()),
+                    Some(split.next().ok_or(RPC_Error::InvalidCookieFile)?.into()),
+                ))
+            }
+        }
+    }
+
+    // Try to read a line from a buffered reader. If no line can be read till the deadline is reached
+    // return a timeout error.
+    fn get_line<R: BufRead>(reader: &mut R, deadline: Instant) -> Result<String, RpcProxyError> {
+        let mut line = String::new();
+        while deadline > Instant::now() {
+            match reader.read_line(&mut line) {
+                // EOF reached for now, try again later
+                Ok(0) => std::thread::sleep(Duration::from_millis(5)),
+                // received useful data, return it
+                Ok(_) => return Ok(line),
+                // io error occurred, abort
+                Err(e) => return Err(e.into()),
+            }
+        }
+        Err(RpcProxyError::RpcTimeout)
+    }
+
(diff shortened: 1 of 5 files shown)


src/blockchain/rpc_proxy.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub(crate) fn new(
+        proxy_addr: &str,
+        rpc_url: &str,
+        proxy_credential: Option<(String, String)>,
+        rpc_auth: &Auth,
+    ) -> Result<Self, RpcProxyError> {
+        // Fetch the RPC address:port and wallet path from url
+        let (target_addr, wallet_path) = {
+            // the url will be of form "http://<rpc-host>:<rpc-port>/wallet/<wallet-file-name>"
+            (rpc_url[7..22].to_owned(), rpc_url[22..].to_owned())
```
afilini, line 89: Ideally here I would parse the url entirely, it's not reliable to just index into the string

src/blockchain/rpc_proxy.rs
```diff
@@ -0,0 +1,241 @@
...
+        };
+
+        // fetch username password from rpc authentication
+        let rpc_auth = {
+            if let (Some(user), Some(pass)) = Self::get_user_pass(rpc_auth)? {
+                let mut auth = user;
```
afilini, line 95: Why not `format!("{}:{}", user, pass)`?

src/blockchain/rpc_proxy.rs
```diff
@@ -0,0 +1,241 @@
...
+            target_addr,
+            proxy_credential,
+            wallet_path,
+            rpc_auth,
+            timeout: Duration::from_secs(15), // Same as regular RPC default
+        })
+    }
+
+    // Helper function to parse username:password pair for rpc Auth
+    fn get_user_pass(auth: &Auth) -> Result<(Option<String>, Option<String>), RpcProxyError> {
```
afilini, line 115: Since you seem to always require both username and password, I would change this to an option of a tuple instead of a tuple of options

src/blockchain/rpc_proxy.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    // Try to read a line from a buffered reader. If no line can be read till the deadline is reached
+    // return a timeout error.
+    fn get_line<R: BufRead>(reader: &mut R, deadline: Instant) -> Result<String, RpcProxyError> {
+        let mut line = String::new();
+        while deadline > Instant::now() {
+            match reader.read_line(&mut line) {
+                // EOF reached for now, try again later
+                Ok(0) => std::thread::sleep(Duration::from_millis(5)),
```
afilini, line 142: When EOF is reached there's nothing more to wait for, you should return an error here

afilini: The function is blocking, so it will internally wait for data and only return a string when there's something: https://doc.rust-lang.org/std/io/trait.BufRead.html#method.read_line

afilini (commented):
It seems a bit redundant to reimplement an http client in here: can you first try opening a pr to the jsonrpc library to see if they somehow let you build a [`Client`](https://docs.rs/jsonrpc/0.12.0/jsonrpc/client/struct.Client.html) from an existing TcpStream?

src/blockchain/rpc_proxy.rs
rajarshimaitra: Yes this does feel clunky. But this seemed easier for now without brain storming too much on a parser. Can you suggest a better way? Would update that in next revise.

src/blockchain/rpc_proxy.rs
rajarshimaitra: This is because the RPC server expects the auth string to be formatted in a certain way which is `base64(b"Basic {user}:{pass}")`. I just broke down the steps. This can be written in shorter lines.

src/blockchain/rpc_proxy.rs
rajarshimaitra: Yes, make sense. Will update in next revision.

src/blockchain/rpc_proxy.rs
rajarshimaitra: I wrote it this way because I wanted to introduce a timeout into it. Socks doesn't work with timeouts. Is there a way we can do timeouts without this?

rajarshimaitra:
Thanks @afilini for the look.

> It seems a bit redundant to reimplement an http client in here: can you first try opening a pr to the jsonrpc library to see if they somehow let you build a Client from an existing TcpStream?

I was looking for ways of how to do this easier in one of our upstream `bitcoincorerpc` or `jsonrpc`. But unfortunately the way `jsonrpc::Client` is written, it doesn't allow putting a `TcpStream` inside.

https://github.com/apoelstra/rust-jsonrpc/blob/ee78ede9a1eb543e0d64c4658019a6e4f7868d09/src/client.rs#L46-L53

The `Client` is wrapper over a `Transport` trait object. I think they did it make the lib flexible. So if we want to put a `TcpStream` in it, we have to make it `Transport`, which is essentially what I did in this PR with a `Socks5Stream`.

So weather we wanna do it in `jsonrpc` or anywhere, the approach kinda needs to be same, unless we propose to heavily refactor `jsonrpc::Client` which seems infeasible to me.

So instead of putting the whole thing in the upstream, I think its better to maintain it in our own repo for our need. The upstreams (both `bitcoincorerpc` and `jsonrpc`) provides specific APIs (`from_jsonrpc` and `with_transport` respectively) to enable this customization. So I think that's also their way of saying, "use your own `Transport` if you don't wanna use the default"?

src/blockchain/rpc_proxy.rs
afilini: I don't think so, `read_line()` is blocking and won't return until either the fd is closed or there's a new line available.

The timeout is usually set on the `TcpStream`, if that's not exposed by the socks library we can't do anything.

afilini:
Sorry I got confused, I meant the http client: https://github.com/apoelstra/rust-jsonrpc/blob/ee78ede9a1eb543e0d64c4658019a6e4f7868d09/src/simple_http.rs

Looking at it they don't store the TcpStream, so maybe that's not feasible.. but it sucks to have to just repeat the same code twice (also, not sure if there are licensing issues if we just copy their stuff in here).

src/blockchain/rpc_proxy.rs
afilini: Yeah it just seems a bit redundant here since the steps are so simple

src/blockchain/rpc_proxy.rs
afilini: A quick way without any extra dependencies could be:

1. check the prefix. It should be either "http://" or nothing
2. replace "http://" with ""
3. splitn with n = 2, separator = "/"
4. first part is the host, second part the wallet path

rajarshimaitra:
> Looking at it they don't store the TcpStream, so maybe that's not feasible.. but it sucks to have to just repeat the same code twice (also, not sure if there are licensing issues if we just copy their stuff in here).

The only thing similar with their `simplehttp` is the http read write part. And that's generic http stuffs. But ya it sucks to have the same http logic twice. 

There is another way I can think of is to propose changing their simplehttp client to include Socks5Stream inside depending on weather proxy options are provided. Then jsonrpc itself will be able provide proxy support. Do you think that would be a better approach?

Also any clue on the `generic socks failure` error? I can't seem to find any concrete reference in open net, and debugging inside isn't giving any more clue than this vague error message.

I have checked that my local tor proxy at port 9050 is working fine with curl.

notmandatory:
Now that #501 is in you can rebase on `master` to fix the CI issue.

rajarshimaitra:
> Now that https://github.com/bitcoindevkit/bdk/pull/501 is in you can rebase on master to fix the CI issue.

Yeah but I am still not sure if this is the best way to go for this..

notmandatory:
No problem, we can leave as draft until you decide how you want to proceed.

notmandatory:
My 2 cents on this is I don't think there is a compelling use-case for it. For mobile devices I think the preferred method will be to connect to public or home hosted bitcoind+electrs servers, or use compact block filters. But I'm not completely opposed to it if you have other uses in mind and/or find others who would use it.

rajarshimaitra:
One use case I can have in mind, is a personal mobile wallet connected to a home Umbrel node.. Umbrel can't be exposed outside home network without Tor and electrum server overhead just for personal usage doesn't make sense.. Bitcoin RPC is enough to operate a wallet that only handles someone's personal data.. Electrums are useful when they wanna expose blockchain service to public.. And then also users need trusted reputed  3rd party Electrum servers and that server have user transaction data available, so that creates potential privacy leak.. (Tor doesn't help there, because the endpoint itself is the point of failure) 

Many of the home use cases with vanilla core (Umbrel or something else) with mobile wallet integration will require Tor support for our RPC blcockhain.. They ideally will need Tor in the mobile too.. But not having Tor support for RPC restricts them to 
 - Either connect only in local home network.
 - Or delpoy electrum server too and then route to it via Tor. (In this case the wallet devs falls back on running their own electrum service, because its too much too ask to wallet users).
 
I personally don't find the electrum way friendly especially for home use cases.. Its an extra overhead (another 400 gb of storage) for home node runners with no added benefits..
 
I forgot which one but Specter/Sparrow wallet allows to connect to a home RPC over Tor.. That makes it very useful to have it operated from anywhere in the world.. And it doesn't need electrum too..

notmandatory:
According to the [`electrs` features](https://github.com/romanz/electrs#features) doc the index overhead should only be about 10%, you may be thinking of `esplora` which does use a lot more disk space and is harder to run on a home single board computer. And if someone does want to run only a simple `bitcoind` node and access it remotely this can also be done with BDK via compact block filters since we do support a single peer.  I don't recall if we support the TOR proxy for those connections, if not I think that'd be more useful since it would also allow you to connect to a (single) public node with a little better privacy than the electrum protocol. 

All that said, if you really think this is something folks will use then I'm not against it on any technical grounds, my primary concern is that it won't be used and doesn't look like a simple thing to add.

EDIT: The compact_filters Peer looks like it does already support connecting via a TOR proxy.

rajarshimaitra:
Yes, the Compact Filter way seems like another possibility, even though isn't meant for this purpose. Users also have to configure their node to allow compact filters since by default its not active for core. But then, it would open compact filter service of the node to public too, and that will cost them good bandwidth (recently pointed out by Murch in a core review, as there are not much CBF nodes out there, the burden of existing node is more, hopefully that will improve in future). But most home node setup packages comes with CBF disabled.  

Yes I agree, its not a easy thing to get into and also not dependent on BDK in any way. Tor support for RPC has to be done in lower level crates, and I am not sure if it will ever get done there too. Its not difficult to implement, but its difficult to organize PR and reviews in all those crates. 

But pushing the ball on that front  in https://github.com/apoelstra/rust-jsonrpc/issues/57

Lets see if it can get somewhere.

I really think it will be very useful. Otherwise the `RpcBlockhain` has very few limited application (only when the wallet and node are in same local network). With some upcoming changes in core, our RPC sync logic can be made more powerful and faster. But it would be sad if we can't use it in its full form in all home applications, for just this one missing piece.

notmandatory:
@rajarshimaitra here's an example of a socks proxy server, and is even in rust, you may be able to use to help with integration testing this feature without involving TOR. You'd have to spin up a socks5-server instance that proxies back to your localhost bitcoind. At least you should be able to verify the socks5 proxy protocol and auth is working (as long as socks5-server is correctly implementing the protocol from the server side).

https://crates.io/crates/socks5-server
~~https://github.com/shadowsocks/shadowsocks-rust~~ nevermind this one.. has non-standard client/server encryption

rajarshimaitra:
Thanks @notmandatory , I think this PR can be closed if you wanna clean up the PR list.. We are not going with this approach.. And I will try to test the socks5 proxy manually with your suggestion.. 

Update:

It worked :tada: .. https://github.com/rust-bitcoin/rust-bitcoincore-rpc/pull/249#issuecomment-1320834261

Thanks for the suggestion.. Opened the upstream PR for review.. :)

notmandatory:
Glad to hear the socks5 proxy suggestion helped. Closing this PR in favor of your rust-bitcoincore-rpc PR.

PR #495, opened 2021-12-07
Disable reqwest's default features

thomaseizinger:
### Description

﻿﻿By default, reqwest uses openssl for TLS. Any consumer wanting to use
rustls will thus pull in unnecessary dependencies.

#### All Submissions:

#### New Features:

* ~[ ] I've added tests for the new feature~
* ~[ ] I've added docs for the new feature~

#### Bugfixes:

* ~[ ] This pull request breaks the existing API~
* ~[ ] I've added tests to reproduce the issue which are now passing~
* ~[ ] I'm linking the issue being fixed by this PR~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unify ureq and reqwest esplora backends to have the same configuration parameters. This means reqwest now has a timeout parameter and ureq has a concurrency parameter.
 - Fixed esplora fee estimation.
 - Fixed generating WIF in the correct network format.
+- Disable `reqwest` default features.
+- Added `reqwest-default-tls` feature: Use this to restore the TLS defaults of reqwest if you don't want to add a dependency to it in your own manifest.
 
 ## [v0.14.0] - [v0.13.0]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ sled = { version = "0.34", optional = true }
 electrum-client = { version = "0.8", optional = true }
 rusqlite = { version = "0.25.3", optional = true }
 ahash = { version = "=0.7.4", optional = true }
-reqwest = { version = "0.11", optional = true, features = ["json"] }
+reqwest = { version = "0.11", optional = true, default-features = false, features = ["json"] }
 ureq = { version = "~2.2.0", features = ["json"], optional = true }
 futures = { version = "0.3", optional = true }
 async-trait = { version = "0.1", optional = true }
@@ -82,6 +82,8 @@ use-esplora-ureq = ["esplora", "ureq", "ureq/socks"]
 # Typical configurations will not need to use `esplora` feature directly.
 esplora = []
 
+# Use below feature with `use-esplora-reqwest` to enable reqwest default TLS support
+reqwest-default-tls = ["reqwest/default-tls"]
 
 # Debug/Test features
 test-blockchains = ["bitcoincore-rpc", "electrum-client"]


LLFourn (approved):
utACK 0b90ac0. I noticed this too but forgot to fix it. Thanks @thomaseizinger.

afilini:
Concept ACK, but I would also add another feature to enable explicit https support (like `reqwest-with-https`) so that if you want the default https implementation you don't have to include reqwest as a dependency of your project just to enable the feature

LLFourn:
@afilini good point. I don't think @thomaseizinger was trying to disable https completely. But if you disable default features I guess that's what happens. I think we should just add `rustls` in the reqwest feature flags. I think you always want https when using reqwest for now.

thomaseizinger:
> I would also add another feature to enable explicit https support (like `reqwest-with-https`) so that if you want the default https implementation you don't have to include reqwest as a dependency of your project just to enable the feature

I generally agree with this but reqwest has a lot of TLS related features: https://github.com/seanmonstar/reqwest/blob/baffb9c004fbd4fb235046683edb3f86ab74596b/Cargo.toml#L31-L43

I don't think it is worth it passing all those through and we would have to do that if we don't want to be opinionated on either openssl or rustls which is the entire point of why I opened this PR :)

notmandatory:
To make it easier for existing users who want to keep the default reqwest behavior I think @afilini 's suggestion makes sense. I'd call it something like `reqwest-default-tls`, and the whole change would be something like:

```toml
reqwest = { version = "0.11", optional = true, default-features = false, features = ["json"] }

# MUST ALSO USE `--no-default-features`
use-esplora-reqwest = ["esplora", "reqwest", "reqwest/socks", "futures"]

# Use below feature with `use-esplora-reqwest` to enable reqwest default TLS support
reqwest-default-tls = ["reqwest/default-tls"]
```

thomaseizinger:
Done!

rajarshimaitra (commented):
utACK https://github.com/bitcoindevkit/bdk/pull/495/commits/56527044bf5e1b1b0ae68d7666be7036d0091b9a

@thomaseizinger @notmandatory would it be better to include this feature operations somewhere in readme doc?

There is also this CI failure happening which I noticed in https://github.com/bitcoindevkit/bdk/pull/497 too.

rajarshimaitra:
Restarted the CI jobs. The failure is still there. Weirdly its not happening in my local.

```
$ cargo +1.46.0 clippy --all-targets --features async-interface --no-default-features -- -D warnings
    Checking bdk v0.14.1-dev (/home/raj/github-repo/bdk)
    Finished dev [unoptimized + debuginfo] target(s) in 2.36s
```

notmandatory:
Now that #501 is in you can rebase on `master` to fix the CI issue.

thomaseizinger:
> Now that #501 is in you can rebase on `master` to fix the CI issue.

Done.

thomaseizinger:
Rebased to fix conflicts. Is there something else blocking this from merging?

rajarshimaitra (approved):
LGTM https://github.com/bitcoindevkit/bdk/pull/495/commits/380a4f2588cf9800bde739eb3ba58ca67b785291

notmandatory (approved):
ACK 380a4f2588cf9800bde739eb3ba58ca67b785291

RCasatta:
> LGTM [380a4f2](https://github.com/bitcoindevkit/bdk/commit/380a4f2588cf9800bde739eb3ba58ca67b785291)

@rajarshimaitra, the `github-merge.py` tool use `ACK` keyword to show devs who acked the PR in the merge commit (see there isn't your LGTM in merge commit description https://github.com/bitcoindevkit/bdk/commit/5107ff80c1ae39dae217b7bc82ce22beb00790ed)description, so I suggest not using LGTM in this case (utACK or other variation are ok too)

rajarshimaitra:
Ah I see.. Thanks @RCasatta will keep that in mind..

PR #500, opened 2021-12-14
Add `verify` feature into blockchain tests

rajarshimaitra:
### Description
As pointed out by @LLFourn https://discord.com/channels/753336465005608961/753367451319926827/919895212434784316 our verification checks are not fully being CI tested. We should check for transaction verification in our sync process. This is already written in `wallet.sync()` but is feature gated by `verify` flag. 
https://github.com/bitcoindevkit/bdk/blob/64e88f0e006c68315142d53dc35b633327dde4b5/src/wallet/mod.rs#L1540-L1555
Thus running the `blockchain_tests` with just `verify` feature would be enough to check the verification process for wallet transactions.

For non wallet transaction verification we should probably include new blockchain tests.

#### All Submissions:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -89,13 +89,13 @@ jobs:
       matrix:
         blockchain:
           - name: electrum
-            features: test-electrum
+            features: test-electrum,verify
           - name: rpc
             features: test-rpc
           - name: esplora
-            features: test-esplora,use-esplora-reqwest
+            features: test-esplora,use-esplora-reqwest,verify
           - name: esplora
-            features: test-esplora,use-esplora-ureq    
+            features: test-esplora,use-esplora-ureq,verify
     steps:
       - name: Checkout
         uses: actions/checkout@v2


rajarshimaitra:
Rebased.

.github/workflows/cont_integration.yml
```diff
@@ -89,13 +89,13 @@ jobs:
       matrix:
         blockchain:
           - name: electrum
-            features: test-electrum
+            features: test-electrum,verify
           - name: rpc
-            features: test-rpc
+            features: test-rpc,verify
```
afilini: This is not necessary as rpc doesn't verify transactions

rajarshimaitra: Right, missed that..

rajarshimaitra:
There is a test failure in the sync verification at wallet module. Which wasn't checked previously. 

This part of the code is already refactored away in #502  and the error doesn't occur there. So I am just porting this PR's changes into 502 and closing this one.

Issue #498, opened 2021-12-12
Non-Wallet Tx verification doesn't work for RPC backend


rajarshimaitra:
**Describe the bug**  
Recently @LLFourn described a possible problem with updated sync logic in `verify_tx` logic in dev discord https://discord.com/channels/753336465005608961/753367451319926827/917624709426855946. I verified the issue exists currently even without the updated sync logic in `rpc` blockchain too.

`verify_tx` doesn't work for non wallet transactions with a rpc blockhain without `txindex`.

This happens because `get_raw_transaction` here https://github.com/bitcoindevkit/bdk/blob/64e88f0e006c68315142d53dc35b633327dde4b5/src/blockchain/rpc.rs#L327-L329 will return error if the transaction is not wallet related.

**To Reproduce**  
This can be reproduced with the test code added here
https://github.com/rajarshimaitra/bdk-example/blob/13097bab3f7742e8cc46c9b4b78be0fd7d44a315/src/main.rs#L276-L335

with the following error message
```
running 1 test
thread 'test::test_transaction_validation' panicked at 'called `Result::unwrap()` on an `Err` value: Global(Rpc(JsonRpc(Rpc(RpcError { code: -5, message: "No such mempool transaction. Use -txindex or provide a block hash to enable blockchain transaction queries. Use gettransaction for wallet transactions.", data: None }))))', src/main.rs:335:63
stack backtrace:
```

**Expected behavior**  
`verify_tx` should verify any transaction, not only just wallet txs.

**Possible recourse**
We should enforce `txindex` to be used in `rpc` backend when using `verify` feature.

LLFourn:
So on the call I think we agreed the solution here is not to verify the transactions on the rpc backend.

This bit of code has to be removed

https://github.com/bitcoindevkit/bdk/blob/64e88f0e006c68315142d53dc35b633327dde4b5/src/wallet/mod.rs#L1540-L1555

and instead each backend that benefits from verification (esplora, electrum) should verify the transactions as they come in during sync. Furthermore the `verified` field on `TransactionDetails` can be removed (this will be a non-breaking DB change hopefully).

rajarshimaitra:
> and instead each backend that benefits from verification (esplora, electrum) should verify the transactions as they come in during sync.

[...]

PR #502, opened 2021-12-16
Refactor verification logic

rajarshimaitra:
### Description

As discussed in https://github.com/bitcoindevkit/bdk/issues/498 and also in team call,
 - default verification from wallet sync is removed
 - `verify_tx` refactored as an wallet API
 - in `sync` verification added for electrum and esplora backends, gated by `verify` flag.
 - `verify` flag is removed from `TransactionDetails`.

### Notes to the reviewers

I haven't looked into `comapct_filters` to see how verification can fit there, but that will probably be required in future.

#### All Submissions:

#### Wallet API change:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Removed default verification from `wallet::sync`. sync-time verification is added in `script_sync` and is activated by `verify` feature flag.
+- `verify` flag removed from `TransactionDetails`.
+
 ## [v0.16.0] - [v0.15.0]
 
 - Disable `reqwest` default features.
@@ -18,12 +21,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Overhauled sync logic for electrum and esplora.
 - Unify ureq and reqwest esplora backends to have the same configuration parameters. This means reqwest now has a timeout parameter and ureq has a concurrency parameter.
 - Fixed esplora fee estimation.
-- Fixed generating WIF in the correct network format.
-- Disable `reqwest` default features.
-- Added `reqwest-default-tls` feature: Use this to restore the TLS defaults of reqwest if you don't want to add a dependency to it in your own manifest.
-- Removed default verification from `wallet::sync`. `sync` verification is added only for `electrum` and
-`esplora` blockchains. No `sync` verification for `rpc`. `verify_tx()` is refactored as an wallet api.
-`verify` flag removed from `TransactionDetails`.
 
 ## [v0.14.0] - [v0.13.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Overhauled sync logic for electrum and esplora.
 - Unify ureq and reqwest esplora backends to have the same configuration parameters. This means reqwest now has a timeout parameter and ureq has a concurrency parameter.
 - Fixed esplora fee estimation.
+- Fixed generating WIF in the correct network format.
+- Disable `reqwest` default features.
+- Added `reqwest-default-tls` feature: Use this to restore the TLS defaults of reqwest if you don't want to add a dependency to it in your own manifest.
+- Removed default verification from `wallet::sync`. `sync` verification is added only for `electrum` and
+`esplora` blockchains. No `sync` verification for `rpc`. `verify_tx()` is refactored as an wallet api.
+`verify` flag removed from `TransactionDetails`.
 
 ## [v0.14.0] - [v0.13.0]
 
--- a/src/blockchain/esplora/reqwest.rs
+++ b/src/blockchain/esplora/reqwest.rs
@@ -167,9 +167,27 @@ impl Blockchain for EsploraBlockchain {
                         .request()
                         .map(|txid| {
                             let tx = tx_index.get(txid).expect("must be in index");
-                            (tx.previous_outputs(), tx.to_tx())
+                            // Verify this transaction if requested via feature flag
+                            #[cfg(feature = "verify")]
+                            {
+                                use crate::wallet::verify::VerifyError;
+                                let prev_outs = tx.previous_outputs();
+                                let tx_bytes = serialize(&tx.to_tx());
+                                for (index, output) in prev_outs.iter().enumerate() {
+                                    if let Some(output) = output {
+                                        bitcoinconsensus::verify(
+                                            output.script_pubkey.to_bytes().as_ref(),
+                                            output.value,
+                                            &tx_bytes,
+                                            index,
+                                        )
+                                        .map_err(|e| VerifyError::from(e))?;
+                                    }
+                                }
+                            }
+                            Ok((tx.previous_outputs(), tx.to_tx()))
                         })
-                        .collect();
+                        .collect::<Result<_, Error>>()?;
                     tx_req.satisfy(full_txs)?
                 }
                 Request::Finish(batch_update) => break batch_update,
--- a/src/blockchain/esplora/reqwest.rs
+++ b/src/blockchain/esplora/reqwest.rs
@@ -167,24 +167,6 @@ impl Blockchain for EsploraBlockchain {
                         .request()
                         .map(|txid| {
                             let tx = tx_index.get(txid).expect("must be in index");
-                            // Verify this transaction if requested via feature flag
-                            #[cfg(feature = "verify")]
-                            {
-                                use crate::wallet::verify::VerifyError;
-                                let prev_outs = tx.previous_outputs();
-                                let tx_bytes = serialize(&tx.to_tx());
-                                for (index, output) in prev_outs.iter().enumerate() {
-                                    if let Some(output) = output {
-                                        bitcoinconsensus::verify(
-                                            output.script_pubkey.to_bytes().as_ref(),
-                                            output.value,
-                                            &tx_bytes,
-                                            index,
-                                        )
-                                        .map_err(|e| VerifyError::from(e))?;
-                                    }
-                                }
-                            }
                             Ok((tx.previous_outputs(), tx.to_tx()))
                         })
                         .collect::<Result<_, Error>>()?;
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -89,13 +89,13 @@ jobs:
       matrix:
         blockchain:
           - name: electrum
-            features: test-electrum
+            features: test-electrum,verify
           - name: rpc
             features: test-rpc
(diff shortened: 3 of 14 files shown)


rajarshimaitra:
Rebased.

LLFourn (commented):
utACK 426430d

Thanks for this @rajarshimaitra.
Looks like sqlite will need some extra care to get rid of the "verified" flag.

rajarshimaitra:
Thanks @LLFourn for the look. Sqlite was tricky. But finally all checks seems to be passing.. 🥳

Changes since last review
 - Added verification tests in CI (closing #500 )
 - Fixed verification flags in sqlite
 - Fixed verification flags in compact filters.

src/blockchain/esplora/reqwest.rs
```diff
@@ -167,9 +167,27 @@ impl Blockchain for EsploraBlockchain {
...
+                                            output.script_pubkey.to_bytes().as_ref(),
+                                            output.value,
+                                            &tx_bytes,
+                                            index,
+                                        )
+                                        .map_err(|e| VerifyError::from(e))?;
+                                    }
+                                }
+                            }
+                            Ok((tx.previous_outputs(), tx.to_tx()))
```
LLFourn, line 170: I think we could just do this once in `script_sync` rather than in these three different places? I get that we want to use `script_sync` logic for rpc soon which doesn't need to verify but we can just turn this off at runtime with a flag when we start the sync.

src/wallet/verify.rs
```diff
@@ -29,45 +29,57 @@ use crate::error::Error;
...
+    /// Performs script verification for a given transaction.
+    /// This uses bitcoinconsensus.
+    /// It can be used as an wallet tool to verify custom transactions by bitcoin's consensus rules.
+    ///
+    /// Note: All blockchains won't be able to verify all transactions.
+    ///
+    /// `electrum` and `esplora` support (confed/unconfed) *any* transaction verification,
+    /// `rpc` supports (confed/unconfed) *in-wallet* transaction verification
+    /// `compact_filters` supports (unconfed) *any* transaction verification.
+    pub fn verify_tx(&self, tx: &Transaction) -> Result<(), VerifyError> {
```
LLFourn: After thinking about this further I really think we should remove this API for now. 

If a user wants to verify a transaction it's unlikely they want to verify a transaction that's in the database. If it's in the database they can just check if it's in the database. If it is then it's verified (assuming you compiled with verification on).

If it's not in the database then it's really a blockchain specific thing in which case it should be a trait that is implemented on blockchains.

LLFourn (changes requested):

src/blockchain/esplora/reqwest.rs
rajarshimaitra: Yes this makes more sense. I will try to see where I can fit this in `script_sync`.

src/wallet/verify.rs
rajarshimaitra: I think I get what you are saying. 

Even if its an inherent `Blockchain` thing, we don't want to force user with its impl in their custom backends. So one way to do that will be to create a default implementation for any `Blockchain`. This can then be feature gated and blanket implemented. 

And the end result would look like `wallet.blockchain.verify()`, instead of current `wallet.verify()`. I am not sure if thats a big enough API win.

My envision of using the verify API is also mostly for non wallet transactions, because its a reasonable assumption that you trust your own DB. And they can enforce it at sync level with `--verify` feature flag for that purpose.

But there should be an easy way to verify non wallet txs. (Even random message verification at future). So the library needs the API somewhere.

So I think for this PR I will just let it be here. And we can make the move in future PR if we want.

rajarshimaitra:
@LLFourn as suggested moved verification logic to script_sync.

@notmandatory updated with rebase.

src/wallet/verify.rs
rajarshimaitra: @LLFourn  Do you feel satisfied with this reasoning?

LLFourn: @rajarshimaitra I'm not sure.

I've thought about this a lot and I think that transaction verification is a very context specific thing and I'm not sure this function fits any of those contexts. Do you have one in mind?

In what situations do you need to verify non-wallet transactions where you don't care about whether the transaction is actually valid but just if the inputs were to exist it passes the consensus rules. Don't you also want to check if the inputs has have been spent or otherwise exist or if the locktime on the tx has passed? Also note this function fails with backends without a transaction index. How are you meant to deal with this failure in the context where it is useful (which again I can't conceive of).

I can understand wanting to check a tx is valid without broadcasting it. Bitcoin core rpc [has a call](https://developer.bitcoin.org/reference/rpc/testmempoolaccept.html) for this. With esplora it would be a matter of checking whether the inputs have been spent and using libconsensus. So I see this as a blockchain specific check which this function doesn't implement properly. It may fail depending on the backend which is odd because the backend it fails on (rpc) is the only one that has a call to do this thing properly (testmempoolaccept).

I don't think it's worth holding up this PR on this but if we are unsure about this function then it makes sense to me to leave it where it is. Why break the API only to break it again later on?

rajarshimaitra: Hmm, I guess I see what you mean. My initial use case was, if I build a transaction with a custom script P2SH and I create my custom witness to spend it, a easy way to check if the the witness works would be useful. Something like `testmempoolaccept` of core RPC.

I don't have any strong opinion on where this should be placed, and no solution for the case of non-txindex blockchains. 

Currently its not breaking the wallet API. Using the lib with `--verify` flag will just open this extra function call on wallet, no other existing calls changes.

What do you suggest is the way forward with this PR?

LLFourn: > Hmm, I guess I see what you mean. My initial use case was, if I build a transaction with a custom script P2SH and I create my custom witness to spend it, a easy way to check if the the witness works would be useful. Something like `testmempoolaccept` of core RPC.

Wouldn't using libconsensus for the P2SH output be preferable in this case? testmempool would only be useful if you want to check the outputs actually exist at the same time and the network time allows for it etc. Another option is to use [`Interpreter`](https://docs.rs/miniscript/6.0.1/miniscript/interpreter/struct.Interpreter.html#method.from_txdata) from rust-miniscript.

> Currently its not breaking the wallet API. Using the lib with `--verify` flag will just open this extra function call on wallet, no other existing calls changes.

It's breaking the API because it removes the existing function.

> What do you suggest is the way forward with this PR?

I'd suggest the easiest way forward is to remove the changes related to the verify function from this PR and then make an issue or follow up PR where we can discuss whether this API is even a good idea or not. iirc @afilini introduced this function so his input is probably more useful than mine here.

rajarshimaitra:
@LLFourn  updated with suggested changes..

Some weird test failure is happening in ureq compilation.. Not sure why..

notmandatory:
@rajarshimaitra I think the CI was failing due to some problem with the recent rust `nightly` used by those jobs.. I pinned the failing jobs to `nightly-2022-01-25` and now they are passing. 

The reason the `Blockchain test-electrum` and two `Blockchain test-esplora` tests are waiting is because the name changed with the new `verify` feature added, I can fix this in the github checks config prior to merging.

rajarshimaitra:
Thanks @notmandatory for looking into it.. Yup they seem like something to do with recent rust.

LLFourn (approved):
utACK 98a3b32.

LGTM thanks @rajarshimaitra

CHANGELOG.md
```diff
@@ -18,6 +18,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Overhauled sync logic for electrum and esplora.
 - Unify ureq and reqwest esplora backends to have the same configuration parameters. This means reqwest now has a timeout parameter and ureq has a concurrency parameter.
 - Fixed esplora fee estimation.
+- Fixed generating WIF in the correct network format.
```
notmandatory: This line, and 22 and 23 should be in release 0.16.0, I think you need to do one final rebase before this is ready.

CHANGELOG.md
```diff
@@ -18,6 +18,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Disable `reqwest` default features.
+- Added `reqwest-default-tls` feature: Use this to restore the TLS defaults of reqwest if you don't want to add a dependency to it in your own manifest.
+- Removed default verification from `wallet::sync`. `sync` verification is added only for `electrum` and
```
notmandatory: Your new changelog lines should also be moved to Unreleased.

CHANGELOG.md
```diff
@@ -18,6 +18,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+`esplora` blockchains. No `sync` verification for `rpc`. `verify_tx()` is refactored as an wallet api.
```
notmandatory: Based on the comment "`verify_tx()` is refactored as a wallet api." I was expecting a new `wallet.verify_tx()` function, am I misunderstanding or did a commit not get pushed yet?

notmandatory (changes requested):
Overall looks good but I have a few changelog related comments.  Some may be fixed with a rebase.

CHANGELOG.md
rajarshimaitra: Moved them to release 0.16.

CHANGELOG.md
rajarshimaitra: Moved to unreleased.

CHANGELOG.md
rajarshimaitra: Updated the comment to better reflect current changes.

rajarshimaitra:
Thanks @notmandatory , I missed the changelog modification in last push.. Updated as suggested.

afilini (commented):
I like the concept and I think the code looks good. My only concern are the database changes that were made, I don't think those are backwards-compatible, especially in sqlite.

For sqlite I would just add another migration that removes the `verified` column.

For sled I don't think we have to do anything, because serde should ignore the extra fields when deserializing, but if you could check it (even manually, no need to add a test for this) it would be great.

LLFourn:
utACK 02f2302 thanks for fixing that up @notmandatory

notmandatory:
In 02f2302 I changed the `MIGRATIONS` to do a proper migration and remove the `transaction_details.verified` column  rather than altering the original `CREATE TABLE transaction_details ...` statement.  Since sqlite doesn't support `ALTER TABLE DROP column` I had to recreate the `transaction_details` table which seems to be the recommended way to do it.  

I also used my [PR](https://github.com/bitcoindevkit/bdk-cli/pull/71) version of `bdk-cli` to manually test `sync` and `list_transaction` before and after this PR change. With our without it works for `key-value-db` or `sqlite`.  I still fixed the `MIGRATIONS` so it can be an example for the future changes.

notmandatory:
~~I think all we need now is a rebase to pickup the pinned tokio version and CHANGELOG (:face_exhaling:) changes in `master`.~~

I removed my cherry-picked commits and added a merge commit from `master` at the end instead.

notmandatory (approved):

PR #505, opened 2021-12-19
Using dust value from rust-bitcoin in ``is_dust``

### LINKED ISSUE #472: Use dust_value from rust-bitcoin
Apparently rule for dust is more complicated than what we have here
https://github.com/bitcoindevkit/bdk/blob/9f04a9d82dcd7280efe520bf29d31779e2bde0d0/src/wallet/utils.rs#L41

see https://github.com/rust-bitcoin/rust-bitcoin/blob/ab97d2db1a4595058367edc80cd038f8467d538e/src/blockdata/script.rs#L457

jumbojets:
### Description

This PR aims to fix #472 . We can retrieve the dust value for a given ``bitcoin::blockdata::script::Script``, so I adjusted the ``is_dust`` function within the ``IsDust`` trait to receive such a ``&Script``. Thus, the ``is_dust`` function can make the proper comparison.
Let me know if you think that there could be a better interface than this.

Furthermore, because this new ``is_dust`` function provides a tighter upper bound on Bitcoin Core's ``GetDustThreshold()``, it actually invalidated a test. In the test, the drain output for a transaction was no longer considered dust and no longer included in the fee. Instead, the drain output was kicked back to the sender, invalidating the asserts in line 3436, 3437 and 3441 in ``src/wallet/mod.rs``. I increased the ``FeeRate`` in the test just enough that the drain output would be small enough to considered dust again and included in the total fee.

#### All Submissions:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -53,7 +53,7 @@ use address_validator::AddressValidator;
 use coin_selection::DefaultCoinSelectionAlgorithm;
 use signer::{SignOptions, Signer, SignerOrdering, SignersContainer};
 use tx_builder::{BumpFee, CreateTx, FeePolicy, TxBuilder, TxParams};
-use utils::{check_nlocktime, check_nsequence_rbf, After, Older, SecpCtx, DUST_LIMIT_SATOSHI};
+use utils::{check_nlocktime, check_nsequence_rbf, After, Older, SecpCtx};
 
 use crate::blockchain::{Blockchain, Progress};
 use crate::database::memory::MemoryDatabase;
@@ -601,7 +601,7 @@ where
         let recipients = params.recipients.iter().map(|(r, v)| (r, *v));
 
         for (index, (script_pubkey, value)) in recipients.enumerate() {
-            if value.is_dust() && !script_pubkey.is_provably_unspendable() {
+            if value.is_dust(script_pubkey) && !script_pubkey.is_provably_unspendable() {
                 return Err(Error::OutputBelowDustLimit(index));
             }
 
@@ -677,9 +677,9 @@ where
 
         if tx.output.is_empty() {
             if params.drain_to.is_some() {
-                if drain_val.is_dust() {
+                if drain_val.is_dust(&drain_output.script_pubkey) {
                     return Err(Error::InsufficientFunds {
-                        needed: DUST_LIMIT_SATOSHI,
+                        needed: drain_output.script_pubkey.dust_value().as_sat(),
                         available: drain_val,
                     });
                 }
@@ -688,7 +688,7 @@ where
             }
         }
 
-        if drain_val.is_dust() {
+        if drain_val.is_dust(&drain_output.script_pubkey) {
             fee_amount += drain_val;
         } else {
             drain_output.value = drain_val;
@@ -3424,7 +3424,7 @@ pub(crate) mod test {
             .unwrap();
 
         let mut builder = wallet.build_fee_bump(txid).unwrap();
-        builder.fee_rate(FeeRate::from_sat_per_vb(140.0));
+        builder.fee_rate(FeeRate::from_sat_per_vb(141.0));
         let (psbt, details) = builder.finish().unwrap();
 
         assert_eq!(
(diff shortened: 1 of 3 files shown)


jumbojets:
My apologies--this is my first time working on the repository . It appears that there were tests on features that I did not account for. Will promptly see how this change affect the tests and fix them!

notmandatory (approved):
ACK bf5994b14ade1fc0c37d3e3cda6946c26ceea631

Looks good, thanks for picking this issue up.

src/wallet/mod.rs
```diff
@@ -3424,7 +3424,7 @@ pub(crate) mod test {
             .unwrap();
 
         let mut builder = wallet.build_fee_bump(txid).unwrap();
-        builder.fee_rate(FeeRate::from_sat_per_vb(140.0));
+        builder.fee_rate(FeeRate::from_sat_per_vb(141.0));
```
rajarshimaitra, line 3427: Is this change because of the test failure?

src/wallet/utils.rs
```diff
@@ -30,16 +28,15 @@ pub(crate) const BLOCKS_TIMELOCK_THRESHOLD: u32 = 500000000;
 
 /// Trait to check if a value is below the dust limit
 // we implement this trait to make sure we don't mess up the comparison with off-by-one like a <
-// instead of a <= etc. The constant value for the dust limit is not public on purpose, to
-// encourage the usage of this trait.
+// instead of a <= etc.
```
rajarshimaitra, line 33: Maybe we can add a doc line here saying "We are perform dust value calculation for a given scriptpubkey using rust-bitcoin, to keep it compatible with network dust fee rate". Not needed but can be helpful for new readers.

rajarshimaitra (approved):
tACK https://github.com/bitcoindevkit/bdk/pull/505/commits/bf5994b14ade1fc0c37d3e3cda6946c26ceea631

Chnages looks good to me. Just one nit.

src/wallet/utils.rs
jumbojets: You're right that would be nice for new readers--I'll include a similar message. Thanks :)

rajarshimaitra (approved):
ReACK https://github.com/bitcoindevkit/bdk/pull/505/commits/a0c140bb29ab37103e4a24d296ac6fcea91da69e

src/wallet/utils.rs
```diff
@@ -28,18 +26,19 @@ pub(crate) const SEQUENCE_LOCKTIME_MASK: u32 = 0x0000FFFF;
...
     /// Check whether or not a value is below dust limit
-    fn is_dust(&self) -> bool;
+    fn is_dust(&self, script: &Script) -> bool;
 }
 
 impl IsDust for u64 {
-    fn is_dust(&self) -> bool {
-        *self <= DUST_LIMIT_SATOSHI
+    fn is_dust(&self, script: &Script) -> bool {
+        *self <= script.dust_value().as_sat()
```
afilini: I think this should be a lt, not lte: from the [docs in rust-bitcoin](https://docs.rs/bitcoin/latest/bitcoin/blockdata/script/struct.Script.html#method.dust_value):

> Gets the minimum value an output with this script should have in order to be broadcastable on today’s bitcoin network

I interpret this as "if the script has this value, then it is broadcastable", so it should be considered dust if it *does not* have this value, hence only if it's lower than it.

Ideally we should also have a test for this somewhere with known good/bad values

jumbojets: Sorry for the late response. Just now seeing this! Good find--I think your interpretation is correct, and I will correspondingly fix it. I will also try to find some good/bad values to construct a test.

jumbojets:
Resolved the issue that @afilini brought up.

Inspiration for the test cases comes from [this stack exchange answer](https://bitcoin.stackexchange.com/a/41082).

notmandatory (approved):
re ACK 5ac51dfe74c51c83fe270570d7ffc50888cf554f

afilini:
ACK 5ac51dfe74c51c83fe270570d7ffc50888cf554f

PR #515, opened 2021-12-31
Never delete spent utxos from the database

### LINKED ISSUE #414: Should be able to spend already unconfirmed spent inputs without RBF
When you do `add_utxo` you should be able to add a utxo that has already been spent in that there is a transaction that has spent it seen by the wallet. One way to do this would to not delete the utxo from the database after you see a transaction spending it but just mark the utxo with that txid.

danielabrozzoni:
### Description

A `is_spent` field is added to LocalUtxo; when a txo is spent we set
this field to true instead of deleting the entire utxo from the
database.
This allows us to create txs double-spending txs already in blockchain.
Listunspent won't return spent in mempool utxos, effectively excluding them from the coin selection and balance calculation
Fixes #414

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `verify` flag removed from `TransactionDetails`.
 - Add `get_internal_address` to allow you to get internal addresses just as you get external addresses.
 - added `ensure_addresses_cached` to `Wallet` to let offline wallets load and cache addresses in their database
+- Add `is_spent` field to `LocalUtxo`; when we notice that a utxo has been spent we set `is_spent` field to true instead of deleting it from the db.
 
 ### Sync API change
 
--- a/src/blockchain/rpc.rs
+++ b/src/blockchain/rpc.rs
@@ -249,7 +249,7 @@ impl WalletSync for RpcBlockchain {
         let mut list_txs_ids = HashSet::new();
 
         for tx_result in list_txs.iter().filter(|t| {
-            // list_txs returns all conflicting tx we want to
+            // list_txs returns all conflicting txs, we want to
             // filter out replaced tx => unconfirmed and not in the mempool
             t.info.confirmations > 0 || self.client.get_mempool_entry(&t.info.txid).is_ok()
         }) {
@@ -332,20 +332,23 @@ impl WalletSync for RpcBlockchain {
                             value: u.amount.as_sat(),
                             script_pubkey: u.script_pub_key,
                         },
+                        is_spent: false,
                     })),
                 },
             )
             .collect::<Result<HashSet<_>, Error>>()?;
 
         let spent: HashSet<_> = known_utxos.difference(&current_utxos).collect();
-        for s in spent {
-            debug!("removing utxo: {:?}", s);
-            db.del_utxo(&s.outpoint)?;
+        for utxo in spent {
+            debug!("setting as spent utxo: {:?}", utxo);
+            let mut spent_utxo = utxo.clone();
+            spent_utxo.is_spent = true;
+            db.set_utxo(&spent_utxo)?;
         }
         let received: HashSet<_> = current_utxos.difference(&known_utxos).collect();
-        for s in received {
-            debug!("adding utxo: {:?}", s);
-            db.set_utxo(s)?;
+        for utxo in received {
+            debug!("adding utxo: {:?}", utxo);
+            db.set_utxo(utxo)?;
         }
 
         for (keykind, index) in indexes {
--- a/src/types.rs
+++ b/src/types.rs
@@ -131,6 +131,8 @@ pub struct LocalUtxo {
     pub txout: TxOut,
     /// Type of keychain
     pub keychain: KeychainKind,
+    /// Whether this UTXO is spent or not
+    pub is_spent: bool,
 }
 
 /// A [`Utxo`] with its `satisfaction_weight`.
--- a/src/blockchain/compact_filters/mod.rs
+++ b/src/blockchain/compact_filters/mod.rs
@@ -163,11 +163,19 @@ impl CompactFiltersBlockchain {
             if let Some(previous_output) = database.get_previous_output(&input.previous_output)? {
                 inputs_sum += previous_output.value;
 
-                if database.is_mine(&previous_output.script_pubkey)? {
+                // this output is ours, we have a path to derive it
+                if let Some((keychain, _)) =
+                    database.get_path_from_script_pubkey(&previous_output.script_pubkey)?
+                {
                     outgoing += previous_output.value;
 
-                    debug!("{} input #{} is mine, removing from utxo", tx.txid(), i);
-                    updates.del_utxo(&input.previous_output)?;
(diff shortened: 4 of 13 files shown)


src/types.rs
```diff
@@ -131,6 +131,8 @@ pub struct LocalUtxo {
     pub txout: TxOut,
     /// Type of keychain
     pub keychain: KeychainKind,
+    /// This UTXO is used as an input of a mempool tx
```
rajarshimaitra: I think a simpler line could be "Whether this UTXO is spent by an unconfirmed tx".

rajarshimaitra (commented):
Initial Concept ACK. The approach looks good to me.

Would like to see a small tests for `list_unspent` also. Ensuring its working as expected will ensure all other application like coinselection, get_balance are working fine too.

This might also be a breaking change for DB APIs as `LocalUtxo` struct is modified. So would be better to include in the change log too.

danielabrozzoni:
Thanks for the review Raj, and sorry it took so long for me to update! I added the rpc code as well, so this PR is now ready :D

I agree with you regarding the test on `listunspent`, but I'm not sure where I should put it (I still haven't familiarized myself with the test structure...). For now, I've added a check in `test_double_spend`:
```rust
for utxo in wallet.list_unspent().unwrap() {
    // Making sure the TXO we just spent is not returned by list_unspent
    assert!(utxo.outpoint != initial_tx.input[0].previous_output, "wallet displays spent txo in unspents");
}
```
but let me know if you prefer it to be tested somewhere else as well :) 

Edit: Uhm, I did forget to update compact filters code 😭 Nevermind, this is still not ready, sorry...
Edit2: Ready 🎉 I'm not really sure if/where I should document this though...

src/blockchain/rpc.rs
```diff
@@ -226,6 +227,28 @@ impl Blockchain for RpcBlockchain {
...
+
+            let tx_result = self.client.get_transaction(&txid, Some(true))?;
+            let tx: Transaction = deserialize(&tx_result.hex)?;
+
+            // If it's unconfirmed, we may want to add its inputs to the currently spent in mempool
+            if tx_result.info.blockhash.is_none() {
+                for input in tx.input.iter() {
+                    // This input is mine if I have both the previous tx and the script_pubkey in
+                    // the db
+                    if let Some(o) = db.get_previous_output(&input.previous_output)? {
```
rajarshimaitra: Can we have a named variable here instead of `o`? It could be called a `txout`..

src/blockchain/rpc.rs
```diff
@@ -226,6 +227,28 @@ impl Blockchain for RpcBlockchain {
         }) {
```
rajarshimaitra, line 255: If you touch this file anyway, can you change the comment two lines above this to 
```
// list_txs returns all conflicting txs, we want to
// filter out replaced tx => unconfirmed and not in the mempool
```

src/blockchain/rpc.rs
```diff
@@ -315,6 +339,10 @@ impl Blockchain for RpcBlockchain {
             debug!("adding utxo: {:?}", s);
             db.set_utxo(s)?;
         }
+        for s in spent_in_mempool {
+            debug!("adding utxo: {:?}", s);
+            db.set_utxo(&s)?;
+        }
```
rajarshimaitra: Feel like named variable would help here too..

```suggestion
        for utxo in spent_in_mempool {
            debug!("adding utxo: {:?}", utxo);
            db.set_utxo(&utxo)?;
        }
```

rajarshimaitra (commented):
Hey @danielabrozzoni , sorry it also took me a long time to get back to it..

made a thorough pass.. tACK https://github.com/bitcoindevkit/bdk/pull/515/commits/f3317eb57df16e2bb03c8fd250060c12fd50db79

The code changes looks good to me.. I have also witnessed problem with esplora before.. Not sure this one is us or something upstream.. For now I think its good to go without it..

Below are just few non-blocking nits I have found.. If you happen to touch this again, feel free toi add them..

rajarshimaitra (approved):
ReACK https://github.com/bitcoindevkit/bdk/pull/515/commits/d39f4a4a3d341c468632b319bc96179fe0d937c1

afilini:
I know I previously said that we should delete UTXOs once confirmed (to avoid bloating the database), but I'm changing my mind: having them would help me with my work on #549, specifically in the code that tries to figure out if we've already performed the "handshake" with a payment code: I could list the transactions received by the notification address corresponding to that payment code, and if I find a transaction that spends our UTXO (here's where I need to have them even if spent) I know we did.

danielabrozzoni:
Avoid deleting utxos would made the code a bit cleaner 😏  I can modify this PR to never delete utxos, and then in future we can think of how to implement some kind of garbage collector.
What should the LocalUtxo struct look like? I could:
- Add a boolean `is_spent`, which is true if the UTXO is spent (both in mempool and already confirmed)
- Add a `spender_txid` field, which would be `None` if the UTXO is unspent, a `Txid` otherwise.

Do you think we would ever need a `spender_txid` field? If we don't, I'd go with `is_spent` (it's easier and saves space 😝). If we might, let's just go with the `spender_txid`.

afilini:
I think we've been having a ton of troubles lately because we try to over-optimize, so in this case I would go for the full `spender_txid`. We don't need it right now, but I'm sure somebody will in the future.

danielabrozzoni:
Actually when trying to implement it I noticed that the `spender_txid` solution is not trivial, and this PR has been opened for a while now. Better to go with `is_spent` now and improve later :)

afilini (approved):
ACK 227eeded0a5e69acd68ec1a3adb278a8639a3525

afilini:
There are conflicts with `master`, rebase and then I'll re-ack

afilini:
Re-ACK f2f0efc

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `verify` flag removed from `TransactionDetails`.
 - Add `get_internal_address` to allow you to get internal addresses just as you get external addresses.
 - added `ensure_addresses_cached` to `Wallet` to let offline wallets load and cache addresses in their database
+- Add `is_spent` field to `LocalUtxo`; when we notice that a utxo has been spent we set `is_spent` field to true instead of deleting it from the db.
```
LLFourn, line 13: post-merge comment: This should probably be called `LocalTxo` now since they're not necessarily ~spent~ unspent.

afilini: In my opinion it's fine to just call it "utxo", I think it's pretty common to refer to outputs in general as utxos. I think what we gain in terms of "naming consistency", we lose with confusion because "txo" doesn't sound as familiar as "utxo" to people (or at least it doesn't to myself :sweat_smile:)

LLFourn: Calling something "unspent" when it isn't unspent seems like it crossing a line...

I think they're usually called "txout"s rather than txos thought.

afilini: Yeah I like `LocalTxout` or `LocalTxOut` better. "txo", while correct, is a weird term imho, it's not used that much.

PR #519, opened 2022-01-06
Add wallet creation example using RPC

rajarshimaitra:
### Description

As mentioned in https://github.com/bitcoindevkit/bitcoindevkit.org/issues/83#issuecomment-1005190174 it would be helpful
to have wallet creation example code in the `example` directory.

This is an attempt to create the most simplistic bdk wallet with a RPC backend and demonstrate receiving, creating, signing
and broadcasting transaction.

The code is a refinement of the RPC tutorial https://bitcoindevkit.org/blog/bitcoin-core-rpc-demo/ with more elaborate doc comments.

### Notes to the reviewers

To automate the background RPC process `electrsd` is used which is already a `dev-dependency` of bdk.

#### All Submissions:

### DIFF
--- a/examples/rpcwallet.rs
+++ b/examples/rpcwallet.rs
@@ -1,0 +1,233 @@
+// Copyright (c) 2020-2021 Bitcoin Dev Kit Developers
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+use bdk::bitcoin::secp256k1::Secp256k1;
+use bdk::bitcoin::util::bip32::ExtendedPrivKey;
+use bdk::bitcoin::Amount;
+use bdk::bitcoin::Network;
+use bdk::bitcoincore_rpc::RpcApi;
+
+use bdk::blockchain::rpc::{Auth, RpcBlockchain, RpcConfig};
+use bdk::blockchain::ConfigurableBlockchain;
+
+use bdk::keys::bip39::{Language, Mnemonic, WordCount};
+use bdk::keys::{DerivableKey, ExtendedKey, GeneratableKey, GeneratedKey};
+
+use bdk::miniscript::miniscript::Segwitv0;
+
+use bdk::sled;
+use bdk::template::Bip84;
+use bdk::wallet::{signer::SignOptions, wallet_name_from_descriptor, AddressIndex, SyncOptions};
+use bdk::KeychainKind;
+use bdk::Wallet;
+
+use bdk::blockchain::Blockchain;
+
+use electrsd;
+
+use std::error::Error;
+use std::path::PathBuf;
+use std::str::FromStr;
+
+/// This example demonstrates a typical way to create a wallet and work with bdk.
+///
+/// This example bdk wallet is connected to a bitcoin core rpc regtest node,
+/// and will attempt to receive, create and broadcast transactions.
+///
+/// To start a bitcoind regtest node programmatically, this example uses
+/// `electrsd` library, which is also a bdk dev-dependency.
+///
+/// But you can start your own bitcoind backend, and the rest of the example should work fine.
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // -- Setting up background bitcoind process
+
+    println!(">> Setting up bitcoind");
+
+    // Start the bitcoind process
+    let bitcoind_conf = electrsd::bitcoind::Conf::default();
+
+    // electrsd will automatically download the bitcoin core binaries
+    let bitcoind_exe =
+        electrsd::bitcoind::downloaded_exe_path().expect("We should always have downloaded path");
+
+    // Launch bitcoind and gather authentication access
+    let bitcoind = electrsd::bitcoind::BitcoinD::with_conf(bitcoind_exe, &bitcoind_conf).unwrap();
+    let bitcoind_auth = Auth::Cookie {
+        file: bitcoind.params.cookie_file.clone(),
+    };
+
+    // Get a new core address
+    let core_address = bitcoind.client.get_new_address(None, None)?;
+
+    // Generate 101 blocks and use the above address as coinbase
+    bitcoind.client.generate_to_address(101, &core_address)?;
+
+    println!(">> bitcoind setup complete");
+    println!(
+        "Available coins in Core wallet : {}",
+        bitcoind.client.get_balance(None, None)?
+    );
+
+    // -- Setting up the Wallet
+
+    println!("\n>> Setting up BDK wallet");
+
+    // Get a random private key
+    let xprv = generate_random_ext_privkey()?;
+
+    // Use the derived descriptors from the privatekey to
+    // create unique wallet name.
+    // This is a special utility function exposed via `bdk::wallet_name_from_descriptor()`
+    let wallet_name = wallet_name_from_descriptor(
+        Bip84(xprv, KeychainKind::External),
+        Some(Bip84(xprv, KeychainKind::Internal)),
+        Network::Regtest,
+        &Secp256k1::new(),
+    )?;
+
+    // Create a database (using default sled type) to store wallet data
+    let mut datadir = PathBuf::from_str("/tmp/")?;
+    datadir.push(".bdk-example");
+    let database = sled::open(datadir)?;
+    let database = database.open_tree(wallet_name.clone())?;
+
+    // Create a RPC configuration of the running bitcoind backend we created in last step
+    // Note: If you are using custom regtest node, use the appropriate url and auth
+    let rpc_config = RpcConfig {
+        url: bitcoind.params.rpc_socket.to_string(),
+        auth: bitcoind_auth,
+        network: Network::Regtest,
+        wallet_name,
+        skip_blocks: None,
+    };
+
+    // Use the above configuration to create a RPC blockchain backend
+    let blockchain = RpcBlockchain::from_config(&rpc_config)?;
+
+    // Combine Database + Descriptor to create the final wallet
+    let wallet = Wallet::new(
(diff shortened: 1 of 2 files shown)


rajarshimaitra:
Apparently `cargo test` is trying to run the example which has some extra feature dependency and causing the test failure in CI.

I am not sure why thats not the case for other examples though (they also have special required deps). 

@notmandatory @afilini am I missing something in the cargo configuration?

rajarshimaitra:
Fixed build failure, all tests passing..

ConorOkus:
Trying to run this locally at the moment, but failing. What command should I run? `cargo run --example rpcwallet`?

rajarshimaitra:
@ConorOkus this should work `cargo run --features="all-keys key-value-db rpc" --example rpcwallet`.

ConorOkus:
tACK

LLFourn:
utACK does the example get run in the ci?

rajarshimaitra:
> utACK does the example get run in the ci?

I think currently it doesn't. But we can explicitly ask the CI to run examples I think. Haven't tried that out.

afilini:
I think it's ok if we just compile this example in the CI, I'm guessing what it does is already covered by specific unit tests

ConorOkus:
I'm trying to do a similar thing in LDK [here](https://github.com/lightningdevkit/rust-lightning/pull/1232/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R204)

afilini:
It's not as easy for BDK because different examples might require different features. So if you just compile them like you did, they will most likely fail.

You would have to get the list of features from the `Cargo.toml` and then compile using those

rajarshimaitra:
Retested

All looks good on this one to me.. Its good to go I think..

notmandatory:
This example LGTM but I'm going hold off merging it until #535 goes in so it can be refactored as needed first. Also removing from next milestone since we're already a bit behind and have some higher priority PRs to get in.

xraid:
has this been massaged to accommodate new stuff in merged https://github.com/bitcoindevkit/bdk/pull/535

rajarshimaitra:
> has this been massaged to accommodate new stuff in merged https://github.com/bitcoindevkit/bdk/pull/535

Good point, This will definitely have effect on the example. Will update it accordingly..

notmandatory:
I think this PR also needs a rebase to pick up the latest verify changes, in particular related to the CI Blockchain jobs.

rajarshimaitra:
Fixed the example in light of #535 and rebased..

examples/rpcwallet.rs
```diff
@@ -0,0 +1,236 @@
...
+
+use std::error::Error;
+use std::path::PathBuf;
+use std::str::FromStr;
+
+/// This example describes a typical code to create
+/// and work with bdk.
+///
+/// This example bdk wallet is connected to a bitcoin core rpc
+/// regtest node and will attempt to to receive, create and
```
notmandatory: typo
```suggestion
/// regtest node and will attempt to receive, create and
```

examples/rpcwallet.rs
```diff
@@ -0,0 +1,236 @@
...
+
+use bdk::blockchain::Blockchain;
+
+use electrsd;
+
+use std::error::Error;
+use std::path::PathBuf;
+use std::str::FromStr;
+
+/// This example describes a typical code to create
```
notmandatory: Small wording nit.. 
```suggestion
/// This example demonstrates a typical way to create a wallet 
```

examples/rpcwallet.rs
```diff
@@ -0,0 +1,236 @@
...
+/// and work with bdk.
+///
+/// This example bdk wallet is connected to a bitcoin core rpc
+/// regtest node and will attempt to to receive, create and
+/// broadcast transaction.
```
notmandatory:
```suggestion
/// broadcast transactions.
```

examples/rpcwallet.rs
```diff
@@ -0,0 +1,236 @@
...
+        auth: bitcoind_auth,
+        network: Network::Regtest,
+        wallet_name,
+        skip_blocks: None,
+    };
+
+    // Use the above configuration to create a RPC blockchain backend
+    let blockchain = RpcBlockchain::from_config(&rpc_config)?;
+
+    // Combine Database + Descriptor toi create the final wallet
```
notmandatory:
```suggestion
    // Combine Database + Descriptor to create the final wallet
```

notmandatory (commented):
I tried testing but got an error, do I need to do something else to run it?:
```shell
% cargo run --features keys-bip39,key-value-db,rpc --example rpcwallet
warning: field is never read: `network`
  --> src/blockchain/rpc.rs:58:5
   |
58 |     network: Network,
   |     ^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `bdk` (lib) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/examples/rpcwallet`
>> Setting up bitcoind
>> bitcoind setup complete
Available coins in Core wallet : 50.00000000 BTC

>> Setting up BDK wallet
Error: Rpc(JsonRpc(Transport(Timeout)))

```

rajarshimaitra:
Thanks @notmandatory for the review.. I fixed the nits.

> I tried testing but got an error, do I need to do something else to run it?:

This seems very weird. Its not happening in my local. And it seems the error is originating from `blockchain::rpc` not the example. Thats even more odd. But I am not sure how to reproduce it.. Probably because of a different rustc version you are using??

xraid:
i get : 

>> Setting up bitcoind
>> bitcoind setup complete
Available coins in Core wallet : 50.00000000 BTC

>> Setting up BDK wallet
Error: Rpc(JsonRpc(Transport(SocketError(Os { code: 35, kind: WouldBlock, message: "Resource temporarily unavailable" }))))

when i swap out in file `rpcwallet.rs` to use : 

`use bdk::bitcoincore_rpc::{Auth as rpc_auth, Client, RpcApi};`

with Client invocation dance => instead of using : 

`use electrsd;`

the BDK wallet finds bitcoind and all is good ...

While running a local regtest bitcoind with correct auth ...

rajarshimaitra:
Thanks @xraid for reporting.. I am still not being able to reproduce any of the issues. Its all working fine in my local as normal.

Did you mean you have to do the bitcoin core rpc setup manually instead of using `electrsd` crate and then it worked?

xraid:
the err says it blocks ? @afilini mentioned something about sharing connections to backends - that might not be the issue here, maybe it is the bitcoin.cookie in auth for the bdk-wallet ?

i changed and used Client instead of the electrsd lines [50...75] 

line 103 reuse bitcoind object for bdk-wallet that was used for core-wallet prior  
`    let rpc_config = RpcConfig {
        url: bitcoind.params.rpc_socket.to_string(),
        auth: bitcoind_auth,`

i am still to fresh in Rust and the BDK code base to make any sense in try debug, 

so i leave it to You and meanwhile watch ...

rajarshimaitra:
> the err says it blocks ? @afilini mentioned something about sharing connections to backends - that might not be the issue here, maybe it is the bitcoin.cookie in auth for the bdk-wallet ?

This should not be an issue because `elecrsd` deploys a local bitcoind only for the code in concern, and deletes it once the example ends.. So there should be only one instance of bdk trying to connect with the temporary bitcoind.

> i changed and used Client instead of the electrsd lines [50...75]

In that case did you manually deploy the regtest node in the background?

Ya I will keep digging to see if I can repro this..

rajarshimaitra:
> the err says it blocks ? @afilini mentioned something about sharing connections to backends - that might not be the issue here, maybe it is the bitcoin.cookie in auth for the bdk-wallet ?

This should not be an issue because `elecrsd` deploys a local bitcoind only for the code in concern, and deletes it once the example ends.. So there should be only one instance of bdk trying to connect with the temporary bitcoind.

> i changed and used Client instead of the electrsd lines [50...75]

In that case did you manually deploy the regtest node in the background?

Ya I will keep digging to see if I can repro this..

rajarshimaitra:
> the err says it blocks ? @afilini mentioned something about sharing connections to backends - that might not be the issue here, maybe it is the bitcoin.cookie in auth for the bdk-wallet ?

This should not be an issue because `elecrsd` deploys a local bitcoind only for the code in concern, and deletes it once the example ends.. So there should be only one instance of bdk trying to connect with the temporary bitcoind.

> i changed and used Client instead of the electrsd lines [50...75]

In that case did you manually deploy the regtest node in the background?

Ya I will keep digging to see if I can repro this..

xraid:
i run a local Nigiri regtest bitcoind electrum and used the `bdk::bitcoincore_rpc::Client` for core-wallet and for bdk-wallet i use : 

`bdk::blockchain::rpc::RpcBlockchain`

where i added auth and url in config : 

    // Set RPC username and password
    let auth = Auth::UserPass {
        username: "admin1".to_string(),
        password: "123".to_string()
    };
    // Set RPC url
    let mut rpc_url = "http://".to_string();
    rpc_url.push_str("127.0.0.1:18443");

    // Create a RPC configuration of the running bitcoind backend we created in last step
    // Note: If you are using custom regtest node, use the appropriate url and auth
    let rpc_config = RpcConfig {
        url: rpc_url, // bitcoind.params.rpc_socket.to_string(),
        auth: auth, //bitcoind_auth,
        network: Network::Regtest,
        wallet_name,
        skip_blocks: None,
    };

    // Use the above configuration to create a RPC blockchain backend
    let blockchain = RpcBlockchain::from_config(&rpc_config)?;

so there is 2 diff connections to bitcoind rpc - one for core and one for bdk 

Why the example with using the `use electrsd;` succeeds with first connect to core wallet and fails when trying to connect from initiated bdk wallet, i dont know ...

notmandatory (approved):
tACK 8a98e69e788dd79d030798465fbc0dec044d86f1

I verified this runs on my Linux (ubuntu 20.04) box.  I'm still getting an error on my macOS system, but I suspect it's something specific to the mac. I'd like to know what's blocking it on the mac but not a reason to hold up this PR.

PR #520, opened 2022-01-09
Change the visibility of the `derived` module and of the `AsDerived` trait

w0xlt:
There is currently no way to retrieve change addresses, so this PR changes the visibility of the `derived` module and the `AsDerived` trait.

With this change, the addresses can be retrieved using the function below, for example:

```rust
use bdk::descriptor::derived::AsDerived;

// ...

fn peek_change_address(&self, wallet: &Wallet<ElectrumBlockchain, MemoryDatabase>, index: u32) -> Result<AddressInfo, Error> {

        let result_descriptor = wallet.get_descriptor_for_keychain(bdk::KeychainKind::Internal);

        result_descriptor
            .as_derived(index, wallet.secp_ctx())
            .address(wallet.network())
            .map(|address| AddressInfo { index, address })
            .map_err(|_| Error::ScriptDoesntHaveAddressForm)
    }
```

### DIFF
--- a/src/descriptor/derived.rs
+++ b/src/descriptor/derived.rs
@@ -119,7 +119,7 @@ impl<'s> ToPublicKey for DerivedDescriptorKey<'s> {
     }
 }
 
-pub(crate) trait AsDerived {
+pub trait AsDerived {
     // Derive a descriptor and transform all of its keys to `DerivedDescriptorKey`
     fn as_derived<'s>(&self, index: u32, secp: &'s SecpCtx)
         -> Descriptor<DerivedDescriptorKey<'s>>;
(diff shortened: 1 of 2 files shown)


LLFourn:
I happened to need this feature for `gun`. Here's the (slightly messy) commit that adds the feature: https://github.com/LLFourn/bdk/commit/e0950b9a0dff93128c43ea05c4fe231786249fe2

I think this is a better way to do this. I'll see if I can clean up this commit and upstream it.

LLFourn:
Here I've made a PR that allows you to do this: https://github.com/bitcoindevkit/bdk/pull/522

rajarshimaitra:
ACK, #522 is much cleaner way to do it.

w0xlt:
Agree.
#522 is a better approach to achieve this change.
I will close this PR.

PR #522, opened 2022-01-11
Add API for internal addresses

LLFourn:
There are good reasons for applications to need to get internal
addresses too. For example creating a transactions that splits an output
into several smaller ones.

#### All Submissions:

#### New Features:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -237,7 +237,7 @@ impl<B, D> Wallet<B, D>
 where
     D: BatchDatabase,
 {
-    // Return a newly derived address using the external descriptor
+    // Return a newly derived address for the specified `keychain`.
     fn get_new_address(&self, keychain: KeychainKind) -> Result<AddressInfo, Error> {
         let incremented_index = self.fetch_and_increment_index(keychain)?;
 
@@ -254,8 +254,8 @@ where
             .map_err(|_| Error::ScriptDoesntHaveAddressForm)
     }
 
-    // Return the the last previously derived address if it has not been used in a received
-    // transaction. Otherwise return a new address using [`Wallet::get_new_address`].
+    // Return the the last previously derived address for `keychain` if it has not been used in a
+    // received transaction. Otherwise return a new address using [`Wallet::get_new_address`].
     fn get_unused_address(&self, keychain: KeychainKind) -> Result<AddressInfo, Error> {
         let current_index = self.fetch_index(keychain)?;
 
@@ -285,7 +285,7 @@ where
         }
     }
 
-    // Return derived address for the external descriptor at a specific index
+    // Return derived address for the descriptor of given [`KeychainKind`] at a specific index
     fn peek_address(&self, index: u32, keychain: KeychainKind) -> Result<AddressInfo, Error> {
         self.get_descriptor_for_keychain(keychain)
             .as_derived(index, &self.secp)
@@ -294,7 +294,7 @@ where
             .map_err(|_| Error::ScriptDoesntHaveAddressForm)
     }
 
-    // Return derived address for the external descriptor at a specific index and reset current
+    // Return derived address for `keychain` at a specific index and reset current
     // address index
     fn reset_address(&self, index: u32, keychain: KeychainKind) -> Result<AddressInfo, Error> {
         self.set_index(keychain, index)?;
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -238,11 +238,11 @@ where
     D: BatchDatabase,
 {
     // Return a newly derived address using the external descriptor
-    fn get_new_address(&self) -> Result<AddressInfo, Error> {
-        let incremented_index = self.fetch_and_increment_index(KeychainKind::External)?;
+    fn get_new_address(&self, keychain: KeychainKind) -> Result<AddressInfo, Error> {
+        let incremented_index = self.fetch_and_increment_index(keychain)?;
 
         let address_result = self
-            .descriptor
+            .get_descriptor_for_keychain(keychain)
             .as_derived(incremented_index, &self.secp)
             .address(self.network);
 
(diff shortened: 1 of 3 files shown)


rajarshimaitra (commented):
Concept + tACK https://github.com/bitcoindevkit/bdk/pull/522/commits/63f1ec0149dcf49518412f9c23ca17b1c4fe6c9e

I think it makes sense and the code looks good to me.

w0xlt (approved):
tACK 63f1ec0

I tested this commit with two wallets: 

1 - with internal and external descriptors: worked as expected.

2- with external descriptor only: `get_internal_address` returns the same addresses as  `get_address`.

Maybe if there is not an internal descriptor, a possible better approach would be to throw an error when calling `get_internal_address` .

But I think this behavior is not related to this PR. It happens because `get_descriptor_for_keychain(keychain)` apparently returns the same descriptor for internal and external keychain when the wallet is created with external descriptor only,

src/wallet/mod.rs
```diff
@@ -284,8 +286,8 @@ where
     }
 
     // Return derived address for the external descriptor at a specific index
```
rajarshimaitra: Just one nit, peek address can peek into both `external` and `internal` descriptors.
```suggestion
    // Return derived address for the descriptor of given [`KeychainKind`] at a specific index
```

rajarshimaitra:
> Maybe if there is not an internal descriptor, a possible better approach would be to throw an error when calling get_internal_address .
But I think this behavior is not related to this PR. It happens because get_descriptor_for_keychain(keychain) apparently returns the same descriptor for internal and external keychain when the wallet is created with external descriptor only,

I think the `get_descriptor_for_keychain()` is written that way to make wallet api agnostic of change descriptor. Weather its there or not, all the logic will simply follow "use external if there's no internal". This simplifies internal logic handling at many places and doesn't force us to handle "no change desc case" separately. and the lib user just keep this in mind while using the wallet. Its also documented in many places.

So it makes sense to me to keep the behavior intact in `peek_address` too.

src/wallet/mod.rs
```diff
@@ -238,11 +238,11 @@ where
     D: BatchDatabase,
 {
     // Return a newly derived address using the external descriptor
-    fn get_new_address(&self) -> Result<AddressInfo, Error> {
-        let incremented_index = self.fetch_and_increment_index(KeychainKind::External)?;
+    fn get_new_address(&self, keychain: KeychainKind) -> Result<AddressInfo, Error> {
+        let incremented_index = self.fetch_and_increment_index(keychain)?;
```
rajarshimaitra: Comment here needs to be also modified. Its not only `external` anymore.

src/wallet/mod.rs
```diff
@@ -256,10 +256,12 @@ where
 
     // Return the the last previously derived address if it has not been used in a received
     // transaction. Otherwise return a new address using [`Wallet::get_new_address`].
-    fn get_unused_address(&self) -> Result<AddressInfo, Error> {
-        let current_index = self.fetch_index(KeychainKind::External)?;
+    fn get_unused_address(&self, keychain: KeychainKind) -> Result<AddressInfo, Error> {
```
rajarshimaitra: Though this one doesn't mention `external` explicitly, maybe better to modify too with "given KeychainKind", for sake of completeness?

src/wallet/mod.rs
```diff
@@ -294,10 +296,10 @@ where
 
     // Return derived address for the external descriptor at a specific index and reset current
     // address index
-    fn reset_address(&self, index: u32) -> Result<AddressInfo, Error> {
-        self.set_index(KeychainKind::External, index)?;
+    fn reset_address(&self, index: u32, keychain: KeychainKind) -> Result<AddressInfo, Error> {
```
rajarshimaitra: Same, can reset both `external` and `internal`.

rajarshimaitra (dismissed):
ACK https://github.com/bitcoindevkit/bdk/pull/522/commits/89f88ce623953e64f8575d358844bc4d8f8124b9

Sorry I missed few others comment nits in last review, of the previous methods.

LLFourn:
comments addressed ready for further review.

notmandatory:
I think docs are breaking because of a regression in rust +nightly, there seems to be a bug fix PR in the works so I think we should wait a couple days and see if that fixes it.  See #538

notmandatory:
#538 is merged and fixes the docs issue, once you rebase this PR should be good to go.

src/wallet/mod.rs
```diff
@@ -4005,6 +4022,45 @@ pub(crate) mod test {
...
+        let wallet = Wallet::new_offline(
+            Bip84(key, KeychainKind::External),
+            None,
+            Network::Regtest,
+            MemoryDatabase::default(),
+        )
+        .unwrap();
+
+        assert_eq!(
+            wallet.get_address(AddressIndex::New).unwrap().address,
```
notmandatory: Shouldn't you be verifying that the internal address is the same as the external here?
```suggestion
            wallet.get_internal_address(AddressIndex::New).unwrap().address,
```

notmandatory (changes requested):
Looks good, one small thing to fix in the new test.

src/wallet/mod.rs
LLFourn: thanks fixed @notmandatory

notmandatory:
> #538 is merged and fixes the docs issue, once you rebase this PR should be good to go.

I think you still need to rebase since this went in, that will fix the rust docs CI job.

LLFourn:
@notmandatory new CI problem. Looks like tokio increased MSRV without major version bump. I think we should set MSRV for bdk to v1.56 (2021 edition) and leave it there.

notmandatory:
I guess we need to revisit and not close #331 .. at this point I agree the best option is set MSRV to rust 1.56, it's not the current stable for Debian.. but is being tested and should be shipped with next stable debian "bookworm" release. https://tracker.debian.org/pkg/rustc

LLFourn:
Rebased. Should be ready for merge after CI (hopefully) passes.

notmandatory (approved):
ACK 022256c91ac7a4d9324a1a7925394a948feec808

PR #523, opened 2022-01-11
Use u64 in blockheight parameter

### LINKED ISSUE #513: inconsistent types for numbers of block in get_block_count and skip_blocks
`bitcoincore_rpc::client::RpcApi pub fn get_block_count()` returns a `u64` value.

```rust
/// Returns the numbers of block in the longest chain.
fn get_block_count(&self) -> Result<u64> {
    self.call("getblockcount", &[])
}
```
But `bdk::blockchain::rpc pub struct RpcConfig` requires `u32` type for `skip_blocks`.

```rust
pub struct RpcConfig {
    /// The bitcoin node url
    pub url: String,
    /// The bitcoin node authentication mechanism
    pub auth: Auth,
    /// The network we are using (it will be checked the bitcoin node network matches this)
    pub network: Network,
    /// The wallet name in the bitcoin node, consider using [crate::wallet::wallet_name_from_descriptor] for this
    pub wallet_name: String,
    /// Skip this many blocks of the blockchain at the first rescan, if None the rescan is done from the genesis block
    pub skip_blocks: Option<u32>,
}
```

Although, in this case, `u64` type can apparently be converted without data loss to `u32`  via `let block_count = u32::try_from(get_block_count()).ok().unwrap()`, it might be better to have the same type for the same thing (numbers of block).

rajarshimaitra:
### Description

Fixes #513. 

Note that the wallet internal blockheight values are still `u32`. So we still need to convert the `u64` to `u32` in `wallet.mod`. This PR turns the height values to `u64` in `blockhain` module only to make it compatible with height values returned by `bitcoincore-rpc`.

Not sure how effective of a change is this. Opening it up for discussion.

#### All Submissions:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1556,7 +1556,7 @@ where
 
         let sync_time = SyncTime {
             block_time: BlockTime {
-                height: maybe_await!(self.client.get_height())?,
+                height: maybe_await!(self.client.get_height())? as u32,
                 timestamp: time::get_timestamp(),
             },
         };
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,7 +94,7 @@ test-md-docs = ["electrum"]
 lazy_static = "1.4"
 env_logger = "0.7"
 clap = "2.33"
-electrsd = { version= "0.13", features = ["trigger", "bitcoind_22_0"] }
+electrsd = { version= "0.14", features = ["trigger", "bitcoind_22_0"] }
 
 [[example]]
 name = "address_validator"
--- a/src/blockchain/any.rs
+++ b/src/blockchain/any.rs
@@ -155,7 +155,7 @@ impl Blockchain for AnyBlockchain {
         maybe_await!(impl_inner_method!(self, broadcast, tx))
     }
 
-    fn get_height(&self) -> Result<u32, Error> {
+    fn get_height(&self) -> Result<u64, Error> {
         maybe_await!(impl_inner_method!(self, get_height))
     }
     fn estimate_fee(&self, target: usize) -> Result<FeeRate, Error> {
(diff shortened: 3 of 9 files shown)


src/wallet/mod.rs
```diff
@@ -1556,7 +1556,7 @@ where
 
         let sync_time = SyncTime {
             block_time: BlockTime {
-                height: maybe_await!(self.client.get_height())?,
+                height: maybe_await!(self.client.get_height())? as u32,
```
w0xlt, line 1559: Maybe this field `BlockTime pub height: u32` can be changed to `u64` in other PR ?

w0xlt (approved):
tACK https://github.com/bitcoindevkit/bdk/pull/523/commits/28bdec2051c2eaea760e7ca972f4f88c0da39564

src/wallet/mod.rs
rajarshimaitra: It can be, then we have to change height to u64 in all of wallet code. That means everywhere. That felt more intrusive to me at this stage, so basically just contained the conversion only in blockchain level. This is where it it gets converted back to u32 for rest of the lib. Would do that if others think that's warranted too. 

I have scoured few other projects (rust-bitcoin, rust-lightning) and there is no consistent type for height. Both of these use u32, so as bdk.

LLFourn:
Concept NACK I don't really understand the motivation for this. Why not just do a checked cast from `u64` to `u32` in the bitcoincore rpc impl?

notmandatory:
If I did my math right a u32 represents ~81,940 years worth of block time. So seems little need for u64 anywhere for representing block height. 

(2^32 * 10) / 60 / 24 / 364

afilini:
Yeah I think we've been using `u32` because that's how long the field is in a Bitcoin block header, so I wouldn't change this.

I think `bitcoincore-rpc` should be changed instead, because it's clearly inconsistent with the other rust-bitcoin crates

rajarshimaitra:
Agreed with others. I too felt its unnecessary, just made the PR to address #513 and get this discussion. Closing it now.

PR #526, opened 2022-01-12
Add code example for each supported backend

w0xlt:
This PR adds code example for connecting to Esplora, Electrum Server, Neutrino and Bitcoin Core.
Also shows how to retrieve balance, sign and broadcast transactions.

To test:
```
cd examples/backend/
cargo run electrum
cargo run esplora
cargo run neutrino
cargo run rpc_core
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ bitcoincore-rpc = { version = "0.16", optional = true }
 
 # Platform-specific dependencies
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-tokio = { version = "1", features = ["rt"] }
+tokio = { version = "1", features = ["rt", "macros"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 getrandom = "0.2"
@@ -142,6 +142,16 @@ required-features = ["electrum", "hardware-signer"]
 name = "electrum_backend"
 path = "examples/electrum_backend.rs"
 required-features = ["electrum"]
+
+[[example]]
+name = "esplora_backend_synchronous"
+path = "examples/esplora_backend_synchronous.rs"
+required-features = ["use-esplora-ureq"]
+
+[[example]]
+name = "esplora_backend_asynchronous"
+path = "examples/esplora_backend_asynchronous.rs"
+required-features = ["use-esplora-reqwest", "reqwest-default-tls", "async-interface"]
 
 [workspace]
 members = ["macros"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -138,6 +138,11 @@ name = "hardware_signer"
 path = "examples/hardware_signer.rs"
 required-features = ["electrum", "hardware-signer"]
 
+[[example]]
+name = "electrum_backend"
+path = "examples/electrum_backend.rs"
+required-features = ["electrum"]
+
 [workspace]
 members = ["macros"]
 [package.metadata.docs.rs]
--- a/examples/electrum_backend.rs
+++ b/examples/electrum_backend.rs
@@ -1,0 +1,87 @@
+use std::str::FromStr;
+
+use bdk::bitcoin::util::bip32::ExtendedPrivKey;
+use bdk::bitcoin::Network;
+use bdk::blockchain::{Blockchain, ElectrumBlockchain};
+use bdk::database::MemoryDatabase;
+use bdk::template::Bip84;
+use bdk::wallet::export::FullyNodedExport;
+use bdk::{KeychainKind, SyncOptions, Wallet};
+
+use bdk::electrum_client::Client;
+use bdk::wallet::AddressIndex;
+use bitcoin::util::bip32;
+
+pub mod utils;
+
+use crate::utils::tx::build_signed_tx;
+
+/// This will create a wallet from an xpriv and get the balance by connecting to an Electrum server.
+/// If enough amount is available, this will send a transaction to an address.
+/// Otherwise, this will display a wallet address to receive funds.
+///
+/// This can be run with `cargo run --example electrum_backend` in the root folder.
+fn main() {
+    let network = Network::Testnet;
+
+    let xpriv = "tprv8ZgxMBicQKsPcx5nBGsR63Pe8KnRUqmbJNENAfGftF3yuXoMMoVJJcYeUw5eVkm9WBPjWYt6HMWYJNesB5HaNVBaFc1M6dRjWSYnmewUMYy";
+
+    let electrum_url = "ssl://electrum.blockstream.info:60002";
+
+    run(&network, electrum_url, xpriv);
+}
+
+fn create_wallet(network: &Network, xpriv: &ExtendedPrivKey) -> Wallet<MemoryDatabase> {
+    Wallet::new(
+        Bip84(*xpriv, KeychainKind::External),
+        Some(Bip84(*xpriv, KeychainKind::Internal)),
+        *network,
+        MemoryDatabase::default(),
+    )
+    .unwrap()
+}
+
+fn run(network: &Network, electrum_url: &str, xpriv: &str) {
+    let xpriv = bip32::ExtendedPrivKey::from_str(xpriv).unwrap();
+
+    // Apparently it works only with Electrs (not EletrumX)
+    let blockchain = ElectrumBlockchain::from(Client::new(electrum_url).unwrap());
+
+    let wallet = create_wallet(network, &xpriv);
+
+    wallet.sync(&blockchain, SyncOptions::default()).unwrap();
+
+    let address = wallet.get_address(AddressIndex::New).unwrap().address;
+
+    println!("address: {}", address);
+
+    let balance = wallet.get_balance().unwrap();
+
+    println!("Available coins in BDK wallet : {} sats", balance);
+
+    if balance.confirmed > 6500 {
+        // the wallet sends the amount to itself.
+        let recipient_address = wallet
+            .get_address(AddressIndex::New)
+            .unwrap()
+            .address
+            .to_string();
+
+        let amount = 5359;
+
+        let tx = build_signed_tx(&wallet, &recipient_address, amount);
+
+        blockchain.broadcast(&tx).unwrap();
+
+        println!("tx id: {}", tx.txid());
+    } else {
+        println!("Insufficient Funds. Fund the wallet with the address above");
+    }
+
+    let export = FullyNodedExport::export_wallet(&wallet, "exported wallet", true)
+        .map_err(ToString::to_string)
+        .map_err(bdk::Error::Generic)
+        .unwrap();
+
+    println!("------\nWallet Backup: {}", export.to_string());
+}
--- a/examples/esplora_backend_asynchronous.rs
+++ b/examples/esplora_backend_asynchronous.rs
@@ -1,0 +1,93 @@
+use std::str::FromStr;
+
+use bdk::blockchain::Blockchain;
+use bdk::{
+    blockchain::esplora::EsploraBlockchain,
+    database::MemoryDatabase,
+    template::Bip84,
+    wallet::{export::FullyNodedExport, AddressIndex},
+    KeychainKind, SyncOptions, Wallet,
+};
+use bitcoin::{
+    util::bip32::{self, ExtendedPrivKey},
+    Network,
+};
+
+pub mod utils;
+
+use crate::utils::tx::build_signed_tx;
+
+/// This will create a wallet from an xpriv and get the balance by connecting to an Esplora server,
+/// using non blocking asynchronous calls with `reqwest`.
+/// If enough amount is available, this will send a transaction to an address.
+/// Otherwise, this will display a wallet address to receive funds.
+///
+/// This can be run with `cargo run --no-default-features --features="use-esplora-reqwest, reqwest-default-tls, async-interface" --example esplora_backend_asynchronous`
+/// in the root folder.
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
+    let network = Network::Signet;
+
+    let xpriv = "tprv8ZgxMBicQKsPcx5nBGsR63Pe8KnRUqmbJNENAfGftF3yuXoMMoVJJcYeUw5eVkm9WBPjWYt6HMWYJNesB5HaNVBaFc1M6dRjWSYnmewUMYy";
+
+    let esplora_url = "https://explorer.bc-2.jp/api";
+
+    run(&network, esplora_url, xpriv).await;
+}
+
+fn create_wallet(network: &Network, xpriv: &ExtendedPrivKey) -> Wallet<MemoryDatabase> {
+    Wallet::new(
+        Bip84(*xpriv, KeychainKind::External),
+        Some(Bip84(*xpriv, KeychainKind::Internal)),
+        *network,
+        MemoryDatabase::default(),
+    )
+    .unwrap()
+}
+
+async fn run(network: &Network, esplora_url: &str, xpriv: &str) {
+    let xpriv = bip32::ExtendedPrivKey::from_str(xpriv).unwrap();
+
+    let blockchain = EsploraBlockchain::new(esplora_url, 20);
+
+    let wallet = create_wallet(network, &xpriv);
+
+    wallet
+        .sync(&blockchain, SyncOptions::default())
+        .await
+        .unwrap();
+
+    let address = wallet.get_address(AddressIndex::New).unwrap().address;
+
+    println!("address: {}", address);
+
+    let balance = wallet.get_balance().unwrap();
+
+    println!("Available coins in BDK wallet : {} sats", balance);
+
+    if balance.confirmed > 10500 {
+        // the wallet sends the amount to itself.
+        let recipient_address = wallet
+            .get_address(AddressIndex::New)
+            .unwrap()
+            .address
+            .to_string();
+
+        let amount = 9359;
+
+        let tx = build_signed_tx(&wallet, &recipient_address, amount);
+
+        let _ = blockchain.broadcast(&tx);
+
+        println!("tx id: {}", tx.txid());
+    } else {
+        println!("Insufficient Funds. Fund the wallet with the address above");
+    }
+
+    let export = FullyNodedExport::export_wallet(&wallet, "exported wallet", true)
+        .map_err(ToString::to_string)
+        .map_err(bdk::Error::Generic)
+        .unwrap();
+
+    println!("------\nWallet Backup: {}", export.to_string());
+}
--- a/examples/esplora_backend_synchronous.rs
+++ b/examples/esplora_backend_synchronous.rs
@@ -1,0 +1,89 @@
+use std::str::FromStr;
+
+use bdk::blockchain::Blockchain;
+use bdk::{
+    blockchain::esplora::EsploraBlockchain,
+    database::MemoryDatabase,
+    template::Bip84,
+    wallet::{export::FullyNodedExport, AddressIndex},
+    KeychainKind, SyncOptions, Wallet,
+};
+use bitcoin::{
+    util::bip32::{self, ExtendedPrivKey},
+    Network,
+};
+
+pub mod utils;
+
+use crate::utils::tx::build_signed_tx;
+
+/// This will create a wallet from an xpriv and get the balance by connecting to an Esplora server,
+/// using blocking calls with `ureq`.
+/// If enough amount is available, this will send a transaction to an address.
+/// Otherwise, this will display a wallet address to receive funds.
+///
+/// This can be run with `cargo run --features=use-esplora-ureq --example esplora_backend_synchronous`
+/// in the root folder.
+fn main() {
+    let network = Network::Signet;
+
+    let xpriv = "tprv8ZgxMBicQKsPcx5nBGsR63Pe8KnRUqmbJNENAfGftF3yuXoMMoVJJcYeUw5eVkm9WBPjWYt6HMWYJNesB5HaNVBaFc1M6dRjWSYnmewUMYy";
+
+    let esplora_url = "https://explorer.bc-2.jp/api";
+
+    run(&network, esplora_url, xpriv);
+}
+
+fn create_wallet(network: &Network, xpriv: &ExtendedPrivKey) -> Wallet<MemoryDatabase> {
+    Wallet::new(
+        Bip84(*xpriv, KeychainKind::External),
+        Some(Bip84(*xpriv, KeychainKind::Internal)),
+        *network,
+        MemoryDatabase::default(),
+    )
+    .unwrap()
+}
+
+fn run(network: &Network, esplora_url: &str, xpriv: &str) {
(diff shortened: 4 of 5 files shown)


rajarshimaitra (commented):
Concept ACK on adding more example to the project. But I think its possible without adding a new cargo project in the workspace. 

Can't we just have running examples like this https://github.com/bitcoindevkit/bdk/pull/519?

notmandatory:

w0xlt:
@notmandatory sorry for the delay.

I force pushed a new version that adds examples just for Electrum and Esplora, as there are already examples for Compact Filter and RPC backends.

I also updated the code to 0.20.1 and addressed @rajarshimaitra's suggestion not to add a new cargo project.

w0xlt:
The new push added the documentation suggested in https://github.com/bitcoindevkit/bdk/issues/685, explaining how to use `use-esplora-reqwest`.

Cargo.toml
```diff
@@ -116,6 +116,15 @@ name = "rpcwallet"
 path = "examples/rpcwallet.rs"
 required-features = ["keys-bip39", "key-value-db", "rpc"]
 
+[[example]]
+name = "electrum_backend"
+path = "examples/electrum_backend.rs"
+
```
rajarshimaitra: You also need to pass the required feature as electrum here, or else all the tests with `--no-default-features` will try to compile this example and fail..

examples/esplora_backend.rs
```diff
@@ -0,0 +1,81 @@
...
+/// This will create a wallet from an xpriv and get the balance by connecting to an Esplora server.
+/// If enough amount is available, this will send a transaction to an address.
+/// Otherwise, this will display a wallet address to receive funds.
+///
+/// This can be run with `cargo run --features="use-esplora-ureq" --example esplora_backend` in the root folder.
+///
+/// Users wanting asynchronous HTTP calls should change `required-features = ["use-esplora-ureq"]` to
+/// `required-features = ["use-esplora-reqwest"]` in `esplora_backend` target in Cargo.toml and run
+/// `cargo run --features="use-esplora-reqwest, reqwest-default-tls" --example esplora_backend`
+/// Without `reqwest-default-tls`, the asynchronous version will not work.
```
rajarshimaitra: This is not fully correct.. In order to use async calls they also need to use the `async-interface` feature along with `use-esplora-reqwest`.. And when they do that this same example will not work anymore.. For example you need to then call `wallet.sync().await.unwrap()` ..

So just changing the required feature doesn't work. if we need to show case the async reqwest thing we need to create another example..

What I would suggest here is use the async and reqwest in the example instead of ureq.. This example is very similar to the electrum one, so just doing another blocking esplora doesn't add much value.. Instead we should do a non blocking reqwest to show how to use the async-interface of bdk..

examples/electrum_backend.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    let xpriv = "tprv8ZgxMBicQKsPcx5nBGsR63Pe8KnRUqmbJNENAfGftF3yuXoMMoVJJcYeUw5eVkm9WBPjWYt6HMWYJNesB5HaNVBaFc1M6dRjWSYnmewUMYy";
+        "health lyrics appear aunt either wrist maple hover family episode seven maze";
+
+    let electrum_url = "ssl://electrum.blockstream.info:60002";
+
+    run(&network, electrum_url, xpriv);
+}
+
+pub fn load_or_create_wallet(
```
rajarshimaitra: This doesn't seem to be loading any wallet, and creating a new one everytime you run the example with memory database. So maybe just call it `create_wallet`??

rajarshimaitra (changes requested):
Concept ACK.. Few more fixes needs to be done..

examples/esplora_backend.rs
w0xlt: `wallet.sync().await.unwrap()` causes the error below:

```
error[E0277]: `Result<(), bdk::Error>` is not a future
  --> examples/esplora_backend.rs:52:53
   |
52 |     wallet.sync(&blockchain, SyncOptions::default()).await.unwrap();
   |                                                     ^^^^^^ `Result<(), bdk::Error>` is not a future
   |
   = help: the trait `std::future::Future` is not implemented for `Result<(), bdk::Error>`
   = note: Result<(), bdk::Error> must be a future or must implement `IntoFuture` to be awaited
   = note: required because of the requirements on the impl of `IntoFuture` for `Result<(), bdk::Error>`
help: remove the `.await`
   |
52 -     wallet.sync(&blockchain, SyncOptions::default()).await.unwrap();
52 +     wallet.sync(&blockchain, SyncOptions::default()).unwrap();
   | 
```

w0xlt: If I run `wallet.sync(&blockchain, SyncOptions::default()).unwrap();` using `use-esplora-reqwest, reqwest-default-tls`, it runs without error, but the wallet shows empty balance because apparently it does not sync.

rajarshimaitra: I think you also need to pass the `async-interface` feature.. Thats what makes the bdk reqwest codes async.. The error seems to be indicating that.. Its getting a regular `Result<>` whereas it expected a `Future`..

w0xlt: I had removed `async-interface` due to error below.
It is weird because `esplora_backend` doesn't have any reference to `electrum` .

```bash
$ cargo run --features="use-esplora-reqwest, reqwest-default-tls, async-interface" --example esplora_backend

   Compiling bdk v0.20.1-dev (/home/node01/Dev/wbdk)
error: Features async-interface and electrum are mutually exclusive and cannot be enabled together
   --> src/lib.rs:214:1
    |
214 | / compile_error!(
215 | |     "Features async-interface and electrum are mutually exclusive and cannot be enabled together"
216 | | );
    | |_^
```

rajarshimaitra: I think getting `reqwest` working is not always straight forward and might need some trial and errors.. I think these examples are still very useful without `reqwest` and with `ureq` instead.. That might help us move this one forward, and do the async example in a independent PR. 

How does that sound to you @w0xlt ??

Cargo.toml
w0xlt: Done in https://github.com/bitcoindevkit/bdk/pull/526/commits/8140d130edc065362ba636828c7a8ee5afb8f884

examples/esplora_backend.rs
w0xlt: I think it is working now.
Apparently https://github.com/bitcoindevkit/bdk/pull/526/commits/400cbad4388322f4f0e0007365bf1a68ea95a6b0 is working with `reqwest`.

examples/electrum_backend.rs
w0xlt: Done in https://github.com/bitcoindevkit/bdk/pull/526/commits/8140d130edc065362ba636828c7a8ee5afb8f884 and https://github.com/bitcoindevkit/bdk/pull/526/commits/400cbad4388322f4f0e0007365bf1a68ea95a6b0.

examples/electrum_backend.rs
```diff
@@ -0,0 +1,88 @@
...
+    wallet
+}
+
+pub fn run(network: &Network, electrum_url: &str, xpriv: &str) {
+    let xpriv = bip32::ExtendedPrivKey::from_str(xpriv).unwrap();
+
+    // Apparently it works only with Electrs (not EletrumX)
+    let blockchain = ElectrumBlockchain::from(Client::new(electrum_url).unwrap());
+
+    let wallet = load_or_create_wallet(&network, &xpriv);
```
rajarshimaitra: You need to update the function name here..

examples/esplora_backend.rs
```diff
@@ -0,0 +1,83 @@
...
+    .unwrap();
+
+    wallet
+}
+pub fn run(network: &Network, esplora_url: &str, xpriv: &str) {
+    let xpriv = bip32::ExtendedPrivKey::from_str(xpriv).unwrap();
+
+    let blockchain = EsploraBlockchain::new(esplora_url, 20);
+
+    let wallet = load_or_create_wallet(&network, &xpriv);
```
rajarshimaitra: same here..

examples/electrum_backend.rs
```diff
@@ -0,0 +1,88 @@
...
+/// This will create a wallet from an xpriv and get the balance by connecting to an Electrum server.
+/// If enough amount is available, this will send a transaction to an address.
+/// Otherwise, this will display a wallet address to receive funds.
+///
+/// This can be run with `cargo run --features="electrum" --example electrum_backend` in the root folder.
+fn main() {
+    let network = Network::Testnet;
+
+    let xpriv = "tprv8ZgxMBicQKsPcx5nBGsR63Pe8KnRUqmbJNENAfGftF3yuXoMMoVJJcYeUw5eVkm9WBPjWYt6HMWYJNesB5HaNVBaFc1M6dRjWSYnmewUMYy";
+        "health lyrics appear aunt either wrist maple hover family episode seven maze";
```
rajarshimaitra: I did not understand why this mnemonic is here..

Cargo.toml
```diff
@@ -116,6 +116,16 @@ name = "rpcwallet"
...
 
+[[example]]
+name = "electrum_backend"
+path = "examples/electrum_backend.rs"
+required-features = ["electrum"]
+
+[[example]]
+name = "esplora_backend"
+path = "examples/esplora_backend.rs"
+required-features = ["use-esplora-reqwest", "reqwest-default-tls"]
```
rajarshimaitra: You also need to update here to make it ureq only..

examples/electrum_backend.rs
```diff
@@ -0,0 +1,88 @@
...
+
+pub mod utils;
+
+use crate::utils::tx::build_signed_tx;
+
+/// This will create a wallet from an xpriv and get the balance by connecting to an Electrum server.
+/// If enough amount is available, this will send a transaction to an address.
+/// Otherwise, this will display a wallet address to receive funds.
+///
+/// This can be run with `cargo run --features="electrum" --example electrum_backend` in the root folder.
```
rajarshimaitra: nit: can omit the quote 
```suggestion
/// This can be run with `cargo run --features=electrum --example electrum_backend` in the root folder.
```

examples/esplora_backend.rs
```diff
@@ -0,0 +1,83 @@
...
+/// If enough amount is available, this will send a transaction to an address.
+/// Otherwise, this will display a wallet address to receive funds.
+///
+/// This can be run with `cargo run --features="use-esplora-reqwest, reqwest-default-tls" --example esplora_backend`
+/// in the root folder.
+///
+/// The configuration above uses asynchronous HTTP calls.
+/// To run it in a synchronous way, the user should change `required-features` to
+/// `required-features = ["use-esplora-ureq"] in `esplora_backend` target in Cargo.toml and run
+/// `cargo run --features="use-esplora-ureq" --example esplora_backend`
```
rajarshimaitra: Can we put this in a `Note` box or something.. Feels like the actual command to run it via reqwest is lost in the middle..

rajarshimaitra (changes requested):
Code looks good.. It need few more fixes to work.. Sorry if I reviewed prematurely.

examples/electrum_backend.rs
w0xlt: I originally used mnemonic instead of xpriv and forgot to removed it. 
Removed in 7a47022ed697e87cb3f978bdeb4aef0c90c39d33.

examples/electrum_backend.rs
w0xlt: My mistake. Done in 7a47022ed697e87cb3f978bdeb4aef0c90c39d33.

examples/esplora_backend.rs
w0xlt: My mistake. Done in a9c426e25d913d3fe4712a3a494ecf25081cfd18.

examples/electrum_backend.rs
w0xlt: Done in 7a47022ed697e87cb3f978bdeb4aef0c90c39d33.

examples/esplora_backend.rs
w0xlt: Done in a9c426e25d913d3fe4712a3a494ecf25081cfd18.

Cargo.toml
w0xlt: I didn't quite understand the suggestion. Could you elaborate more?

rajarshimaitra: Sorry my bad.. I misunderstood the doc comment.. This is fine..

examples/electrum_backend.rs
```diff
@@ -0,0 +1,89 @@
...
+    let network = Network::Testnet;
+
+    let xpriv = "tprv8ZgxMBicQKsPcx5nBGsR63Pe8KnRUqmbJNENAfGftF3yuXoMMoVJJcYeUw5eVkm9WBPjWYt6HMWYJNesB5HaNVBaFc1M6dRjWSYnmewUMYy";
+
+    let electrum_url = "ssl://electrum.blockstream.info:60002";
+
+    run(&network, electrum_url, xpriv);
+}
+
+pub fn create_wallet(network: &Network, xpriv: &ExtendedPrivKey) -> Wallet<MemoryDatabase> {
```
rajarshimaitra: Does these functions need to be public? This applies to all other such functions..

rajarshimaitra (commented):
tACK 95e9b7f5fc0ee5aa6586217c8b8ff6400eab6853

Code loos good.. Thanks for the quick update.. One last comment..

examples/electrum_backend.rs
w0xlt: Good catch. Done in f13d5e6bbb4d375037cd54cac694ca04003420c8 and aa5db56f8b2d42e4cf36b002a6e8afebdbd77d93.

examples/electrum_backend.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    let address = wallet.get_address(AddressIndex::New).unwrap().address;
+
+    println!("address: {}", address);
+
+    let balance = wallet.get_balance().unwrap();
+
+    println!("balance: {}", balance);
+
+    if balance > 6500 {
```
rajarshimaitra: balance isn't an integer anymore.. Thats why tests are failing..

rajarshimaitra (changes requested):
ACK f13d5e6bbb4d375037cd54cac694ca04003420c8

One last fix to pass the tests..

examples/electrum_backend.rs
w0xlt: Rebased and changed the comparison.

examples/esplora_backend.rs
```diff
@@ -0,0 +1,93 @@
...
+use crate::utils::tx::build_signed_tx;
+
+/// This will create a wallet from an xpriv and get the balance by connecting to an Esplora server.
+/// If enough amount is available, this will send a transaction to an address.
+/// Otherwise, this will display a wallet address to receive funds.
+///
+/// This can be run with `cargo run --features="use-esplora-reqwest, reqwest-default-tls" --example esplora_backend`
+/// in the root folder.
+///
+/// Note: The configuration above uses asynchronous HTTP calls.
```
rajarshimaitra, line 27: Sorry I have missed this before. Nothing major, feel free to ignore for this PR. 

Even though the calls are async in `reqwest` they are still blocking in our impl, using the `await_or_block` macro like this, https://github.com/bitcoindevkit/bdk/blob/06310f1dd06641c36cb1980ee34489e03eadf96b/src/blockchain/esplora/reqwest.rs#L126 

So to actually make the wallet calls async in the example you need to use the `async-interface` feature too.. In that case you will have `await` in your wallets calls in the example.. 

This PR is good to go on its own.. Maybe as a future PR an idea.

 - Transform this to a blocking example using `ureq`.
 - Create another fully `async` example using `reqwest`.

rajarshimaitra (approved):
tACK c7e1247777e3862e0e863e070bf3a7187bda8736

LGTM.. One suggestion maybe for a future PR.

examples/esplora_backend.rs
rajarshimaitra: Sorry I have missed this before. Nothing major, feel free to ignore for this PR. 

Even though the calls are async in `reqwest` they are still blocking in our impl, using the `await_or_block` macro like this, https://github.com/bitcoindevkit/bdk/blob/06310f1dd06641c36cb1980ee34489e03eadf96b/src/blockchain/esplora/reqwest.rs#L126 

So to actually make the wallet calls async in the example you need to use the `async-interface` feature too.. In that case you will have `await` in your wallets calls in the example.. 

This PR is good to go on its own.. Maybe as a future PR an idea.

 - Transform this to a blocking example using `ureq`.
 - Create another fully `async` example using `reqwest`.

w0xlt: That would be great. But, as mentioned above, the command `cargo run --features="use-esplora-reqwest, reqwest-default-tls, async-interface" --example esplora_backend` results in error.

I think the problem is here (`src/lib:210`)

```rust
#[cfg(all(feature = "async-interface", feature = "electrum"))]
compile_error!(
    "Features async-interface and electrum are mutually exclusive and cannot be enabled together"
);
```

Even there is no `electrum` in the features, it seems this is enabled by default. Is there a way to disable it ?

notmandatory: You can disable the default features (which are "key-value-db", "electrum") by adding `--no-default-features`, eg:
```
cargo run --no-default-features --features=use-esplora-reqwest,reqwest-default-tls,async-interface --example esplora_backend
```

w0xlt: @notmandatory this worked. Thanks.

@rajarshimaitra I created two files `examples/esplora_backend_synchronous.rs` and `examples/esplora_backend_asynchronous.rs` in a2b6a408f7589f2b9682181b7ab57447a28d6866.

w0xlt: I don't know why `CI / Rust fmt (pull_request)` failed. I  ran `cargo fmt --all` before committing.

examples/esplora_backend_asynchronous.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+async fn run(network: &Network, esplora_url: &str, xpriv: &str) {
+    let xpriv = bip32::ExtendedPrivKey::from_str(xpriv).unwrap();
+
+    let blockchain = EsploraBlockchain::new(esplora_url, 20);
+
+    let wallet = create_wallet(network, &xpriv);
+
+    await_or_block!(wallet.sync(&blockchain, SyncOptions::default())).unwrap();
```
rajarshimaitra: We don't need `await_or_block` as we know its gonna be async in this case..
```suggestion
    wallet.sync(&blockchain, SyncOptions::default()).await.unwrap();
```

examples/esplora_backend_asynchronous.rs
```diff
@@ -0,0 +1,91 @@
...
+    Network,
+};
+
+pub mod utils;
+
+use crate::utils::tx::build_signed_tx;
+
+use bdk_macros::await_or_block;
+
+/// This will create a wallet from an xpriv and get the balance by connecting to an Esplora server.
```
rajarshimaitra: Something like this.. 
```suggestion
/// This will create a wallet from an xpriv and get the balance by connecting to an Esplora server, using non blocking asynchronous calls with `reqwest`. 
```

examples/esplora_backend_synchronous.rs
```diff
@@ -0,0 +1,88 @@
...
+use bitcoin::{
+    util::bip32::{self, ExtendedPrivKey},
+    Network,
+};
+
+pub mod utils;
+
+use crate::utils::tx::build_signed_tx;
+
+/// This will create a wallet from an xpriv and get the balance by connecting to an Esplora server.
```
rajarshimaitra: Similar suggestion. The idea is to explain what this example is for in the first line. 
```suggestion
/// This will create a wallet from an xpriv and get the balance by connecting to an Esplora server, using blocking calls with `ureq`.  
```

Cargo.toml
```diff
@@ -119,6 +119,21 @@ name = "rpcwallet"
...
+
+[[example]]
+name = "esplora_backend_synchronous"
+path = "examples/esplora_backend_synchronous.rs"
+required-features = ["use-esplora-ureq"]
+
+[[example]]
+name = "esplora_backend_asynchronous"
+path = "examples/esplora_backend_asynchronous.rs"
+required-features = ["use-esplora-reqwest", "reqwest-default-tls"]
```
rajarshimaitra: shouldn't this include `async-interface` too?

examples/electrum_backend.rs
```diff
@@ -0,0 +1,87 @@
...
+
+pub mod utils;
+
+use crate::utils::tx::build_signed_tx;
+
+/// This will create a wallet from an xpriv and get the balance by connecting to an Electrum server.
+/// If enough amount is available, this will send a transaction to an address.
+/// Otherwise, this will display a wallet address to receive funds.
+///
+/// This can be run with `cargo run --features=electrum --example electrum_backend` in the root folder.
```
rajarshimaitra: This works with `cargo run --example electrum_backend` because `electrum` is there by default.

rajarshimaitra (changes requested):
ACK a2b6a408f7589f2b9682181b7ab57447a28d6866

Thanks for the split.. This now looks complete.. Few more nits..

rajarshimaitra:
> I don't know why CI / Rust fmt (pull_request) failed. I ran cargo fmt --all before committing.

It was some temporary problem with rust update server.. I restarted the tests and it passed..

examples/esplora_backend_asynchronous.rs
w0xlt: Done in e50fcb6c53964435d8a70f6fe396bf60bb75f1e2.

examples/esplora_backend_asynchronous.rs
w0xlt: Done in e50fcb6c53964435d8a70f6fe396bf60bb75f1e2.

Cargo.toml
w0xlt: Done in e50fcb6c53964435d8a70f6fe396bf60bb75f1e2.

examples/esplora_backend_synchronous.rs
w0xlt: Done in e50fcb6c53964435d8a70f6fe396bf60bb75f1e2.

examples/electrum_backend.rs
w0xlt: Done in ea2b9f5a433f9b04a257e61a2a9ddd7f1526b324. Thanks for the review.

rajarshimaitra (approved):
tACK e50fcb6c53964435d8a70f6fe396bf60bb75f1e2

w0xlt:
Rebased.
Is any additional action required ?

rajarshimaitra (approved):
Sorry took me some time to get back to this.. Thanks for pushing this through.. 

tACK f99a6b9f434e86a78695da72e9d28e6937c7e5db

All looks good to me, and great work separating the blocking and async example separate. This is good to go, no more changes should be required..

22f-cy (approved):
if [ "$PWD" = "$HOME" ]; then

Issue #525, opened 2022-01-12
`get_descriptor_for_keychain(KeychainKind::Internal)` should return `None` if there is no internal keychain


w0xlt:
The code below returns the external descriptor.

I understand that external descriptor will be used as (internal) change descriptor if it does not exist, but I think it would be better and less ambiguous if the function returned `None` or `Error` when explicitly calling  the internal descriptor.

This would also make a better distinction between external and internal descriptors and be more in line with the wallet structure.

```rust
let wallet = Wallet::new_offline(
        "<desc>",
        None,
        network,
        MemoryDatabase::default(),
)
.unwrap();

let desc = wallet.get_descriptor_for_keychain(KeychainKind::Internal);

println!("desc: {}", desc.to_string());
```

LLFourn:
Or maybe just `has_descriptor_for_keychain` returning a `bool`.

w0xlt:
Good suggestion.
I opened https://github.com/bitcoindevkit/bdk/pull/528 to add this function.

LLFourn:
As mentioned in #528 you can do this with `Wallet::public_descriptor` and checking if the return value  is `is_none()`.

PR #528, opened 2022-01-13
Add function to check internal descriptor

w0xlt:
As discussed in https://github.com/bitcoindevkit/bdk/pull/522#issuecomment-1011200611 and suggested in https://github.com/bitcoindevkit/bdk/issues/525#issuecomment-1011783112, there is no way of knowing whether the descriptor returned in `get_descriptor_for_keychain` is external or internal.

According to the first comment, this is a deliberate decision. So this PR adds a new function to check if internal descriptor exists.

This information can be important when showing a list of addresses and indicating which descriptor the address comes from, for example.

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1059,6 +1059,11 @@ where
         &self.secp
     }
 
+    /// Checks if the wallet has internal descriptor
+    pub fn has_internal_descriptor(&self) -> bool {
+        self.change_descriptor.is_some()
+    }
+
     /// Returns the descriptor used to create addresses for a particular `keychain`.
     pub fn get_descriptor_for_keychain(&self, keychain: KeychainKind) -> &ExtendedDescriptor {
         let (descriptor, _) = self._get_descriptor_for_keychain(keychain);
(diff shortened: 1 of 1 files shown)


rajarshimaitra (commented):
Though I am still unsure why would a lib user won't know if his wallet has a change_desc or not, but I think its no harm to explicitly have an API to determine that.

ACK https://github.com/bitcoindevkit/bdk/pull/528/commits/64acb93980fa58b3aef6be1a266e89bb727bba08

src/wallet/mod.rs
```diff
@@ -1059,6 +1059,11 @@ where
         &self.secp
     }
 
+    /// Checks if the wallet has internal descriptor
+    pub fn has_internal_descriptor(&self) -> bool {
```
LLFourn, line 1063: I think `has_descriptor_for_keychain(&self, keychain: KeychainKind)` would be better. I have in my mind to add other `KeychainKind`s later on and this would be helpful to that end.

LLFourn:
@w0xlt we discussed about this in the meeting today. Can you give us an idea what you would use this for?

w0xlt:
>Can you give us an idea what you would use this for?

Sure. This PR is related to how descriptors are implemented in the BDK. This behavior is described in https://github.com/bitcoindevkit/bdk/pull/522#pullrequestreview-849725900 and https://github.com/bitcoindevkit/bdk/pull/522#issuecomment-1011200611.

There is currently no way to retrieve internal (change) addresses.

With #522, this will be possible, but if there's no internal descriptor, `get_internal_address` will return the same addresses as `get_address`. That is, if there is no internal descriptor, there is still no way to know this.

This PR adds the ability to know if there is an internal descriptor and how to better process the results of `get_internal_address`.

The use case is: building an Electrum wallet using BDK. As the image below shows, it is necessary to distinguish between internal and external addresses.

image

notmandatory:
What if we add the `KeychainKind` to the `AddressInfo` struct returned by `get_internal_address` (and other functions that return `AddressInfo`) ? then there's be no need to do a `has_internal_descriptor` check and will help support other `KeychainKind` values going forward.

w0xlt:
I agree.

I think there is no need for a new function (`get_internal_address`).
The current function `get_address` can receive the `AddressIndex` and the `KeychainKind` as parameters. And `bdk::KeychainKind::External` can be the default value for this new parameter.

This way, it maintains backwards compatibility and still adds this new functionality.

I also don't see a problem if the BDK throws an exception if the user requires a `KeychainKind` that doesn't exist. This would make it easier to distinguish the addresses.

LLFourn:
It turns out we already have a method that gives you this anyway: `Wallet::public_descriptor` will return the descriptor for either keychain and `None` when there is no internal descriptor. What @notmandatory suggested seems like a reasonable improvement regardless.

PR #535, opened 2022-01-27
Remove blockchain from wallet

LLFourn:
While trying to finish #490 I thought that it'd be better to try the idea of getting rid of a lot of the async macros and just having tow different traits for sync and async stuff. While trying to do that I felt that this needed to be done first.
 
The goal of this change is to decouple the wallet from the blockchain trait. A wallet is something that keeps track of UTXOs and transactions (and can sign things). The only reason it should need to talk to the blockchain is if doing a `sync`. So we remove all superfluous calls to the blockchain and precisely define the requirements for the blockchain to be used to sync with two new traits: `WalletSync` and `GetHeight`.

1. Stop requesting height when wallet is created
2. `new_offline` is just `new` now.
3. a `WalletSync + GetHeight` is now the first argument to `sync`.
4. `SyncOptions` replaces the existing options to `sync` and allows for less friction when making breaking changes in the fuutre (no more noop_progress!).
5. We `Box` the `Progress` now to avoid type parameters
6. broadcast has been removed from `Wallet`. You just use the blockchain directly to broadcast now.

### Notes to the reviewers

This needs #502 before it can be merged but can reviewed now.
Be on the look up for stale documentation from this change.
Our doc build looks broken because of ureq or something.

#### All Submissions:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Removed default verification from `wallet::sync`. sync-time verification is added in `script_sync` and is activated by `verify` feature flag.
 - `verify` flag removed from `TransactionDetails`.
-- Removed Blockchain from Wallet.
-- Removed `Wallet::broadcast` (just use blockchain.broadcast)
+
+### Sync API change
+
+To decouple the `Wallet` from the `Blockchain` we've made major changes:
+
+- Removed `Blockchain` from Wallet.
+- Removed `Wallet::broadcast` (just use `Blockchain::broadcast`)
 - Depreciated `Wallet::new_offline` (all wallets are offline now)
-- Changed `Wallet::sync` to take a blockchain argument.
+- Changed `Wallet::sync` to take a `Blockchain`.
+- Stop making a request for the block height when calling `Wallet:new`.
+- Added `SyncOptions` to capture extra (future) arguments to `Wallet::sync`.
 
 ## [v0.16.1] - [v0.16.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ To decouple the `Wallet` from the `Blockchain` we've made major changes:
 - Changed `Wallet::sync` to take a `Blockchain`.
 - Stop making a request for the block height when calling `Wallet:new`.
 - Added `SyncOptions` to capture extra (future) arguments to `Wallet::sync`.
+- Removed `max_addresses` sync parameter which determined how many addresses to cache before syncing since this can just be done with `ensure_addreses_cached`.
 
 ## [v0.16.1] - [v0.16.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Removed default verification from `wallet::sync`. sync-time verification is added in `script_sync` and is activated by `verify` feature flag.
 - `verify` flag removed from `TransactionDetails`.
+- Removed Blockchain from Wallet.
+- Removed `Wallet::broadcast` (just use blockchain.broadcast)
+- Depreciated `Wallet::new_offline` (all wallets are offline now)
+- Changed `Wallet::sync` to take a blockchain argument.
 
 ## [v0.16.1] - [v0.16.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ To decouple the `Wallet` from the `Blockchain` we've made major changes:
 - Changed `Wallet::sync` to take a `Blockchain`.
 - Stop making a request for the block height when calling `Wallet:new`.
 - Added `SyncOptions` to capture extra (future) arguments to `Wallet::sync`.
-- Removed `max_addresses` sync parameter which determined how many addresses to cache before syncing since this can just be done with `ensure_addreses_cached`.
+- Removed `max_addresses` sync parameter which determined how many addresses to cache before syncing since this can just be done with `ensure_addresses_cached`.
 
 ## [v0.16.1] - [v0.16.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ To decouple the `Wallet` from the `Blockchain` we've made major changes:
 
 - Removed `Blockchain` from Wallet.
 - Removed `Wallet::broadcast` (just use `Blockchain::broadcast`)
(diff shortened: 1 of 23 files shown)


afilini (commented):
I think the code looks really good, but I haven't checked the documentation yet.

I will try to see if I can make it skip our dependencies in the doc build, it should also make it a lot faster.

notmandatory:
I created #538 for the broken docs, there's a quick way to fix it but will lose a docs feature from nightly we use.

afilini:
I was looking at #502, i figured we could make the `SyncOptions` struct a type defined by the `WalletSync` trait itself. The idea is that different backends might naturally have different options, for instance `electrum` and `esplora` could have an extra option to verify the transactions.

I don't think it would be too bad in terms of API, `Wallet::sync` would become something like:

```rust
pub fn sync<B: WalletSync + GetHeight>(
    &self,
    blockchain: &B,
    sync_opts: <B as WalletSync>::SyncOptions,
) -> Result<(), Error> {
```

LLFourn:
@afilini It'd be nice to change it to a runtime flag. I think if we want to go that direction it'd be fine to have a flag that is present for all backends that you just ignore in the case of RPC.

Part of my more longer term goal here is to make it possible to `Arc<dyn SyncWallet>` around a blockchain eventually. Adding an associated type would prevent this.

src/blockchain/any.rs
```diff
@@ -16,61 +16,17 @@
...
-//!     Network::Testnet,
-//!     MemoryDatabase::default(),
-//!     esplora_blockchain.into(),
-//! )?;
-//! # }
-//!
-//! # Ok::<(), bdk::Error>(())
-//! ```
-//!
 //! When paired with the use of [`ConfigurableBlockchain`], it allows creating wallets with any
```
notmandatory: No wallet needed, but creating any type of blockchain.
```suggestion
//! When paired with the use of [`ConfigurableBlockchain`], it allows creating any
```

src/wallet/mod.rs
```diff
@@ -232,11 +156,74 @@ impl fmt::Display for AddressInfo {
...
-// offline actions, always available
-impl<B, D> Wallet<B, D>
+#[derive(Debug, Default)]
+/// Options to a [`sync`].
+///
+/// [`sync`]: Wallet::sync
+pub struct SyncOptions {
+    /// The progress tracker which may be informed when progress is made.
+    pub progress: Option<Box<dyn Progress>>,
+    /// The maximum number of addresses sync on.
```
notmandatory:
```suggestion
    /// The maximum number of new addresses to derive and cache on sync.
```

notmandatory (commented):
Concept ACK plus a couple doc suggestions.

examples/compiler.rs
```diff
@@ -89,7 +89,7 @@ fn main() -> Result<(), Box<dyn Error>> {
         .transpose()
         .unwrap()
         .unwrap_or(Network::Testnet);
-    let wallet = Wallet::new_offline(&format!("{}", descriptor), None, network, database)?;
```
LLFourn, line 92: note to self: `new_offline` should just be depreciated and aliased to `new`.

LLFourn:
This is ready for final review and merge. I just pushed a few fixes:

1. `get_tx` has also been rolled into its own trait  because it was convenient for fixing the tests for `verify_tx`. [c0e75fc](https://github.com/bitcoindevkit/bdk/pull/535/commits/c0e75fc1a8560cdacf705ff60ef8833befd8280c)
2. I had to fix a bug in bitcoin rpc's sync which meant that if two wallets used the same blockchain they would get each other's transactions in their sync (because they both called loadwallet to the same backend). I'm not 100% clear why this wasn't happening before but it's fixed now. [035307e](https://github.com/bitcoindevkit/bdk/pull/535/commits/035307ef54ebc330b1760bbdd17bdab3b167cfeb)

icota:
This PR makes perfect sense. I was somewhat unhappy with the fact online wallets can't be initalised offline or why there even was a distinction between the two.

FWIW I've rebased a project I'm working on on top of this and all the tests pass. Code looks good too. Thanks @LLFourn 

tACK fbb50ad1c886a9cb716146fa823558d7d655c7b7

LLFourn:
Conflicts resolved. Ready for merge.

notmandatory (approved):
tACK d03aa851086c68592770942ecd18d1add7d204ee

Tested with `bdk-cli` (https://github.com/bitcoindevkit/bdk-cli/pull/73) and all looks good.

notmandatory:
@afilini do you want to give this a once over before it goes in?

LLFourn:
Since this hadn't been merged yet I took the opportunity to remove `max_addresses` as a sync parameter because:

1. Nobody uses it and there are no tests for it.
2. My next work will hopefully get rid of the address caching in the sync pipeline except when doing a full sync for the first time.
3. We have a method to do this explicitly now with #537

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ To decouple the `Wallet` from the `Blockchain` we've made major changes:
 - Changed `Wallet::sync` to take a `Blockchain`.
 - Stop making a request for the block height when calling `Wallet:new`.
 - Added `SyncOptions` to capture extra (future) arguments to `Wallet::sync`.
+- Removed `max_addresses` sync parameter which determined how many addresses to cache before syncing since this can just be done with `ensure_addreses_cached`.
```
thunderbiscuit: Small typo here: `ensure_addresses_cached`.

CHANGELOG.md
```diff
@@ -11,6 +11,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `get_internal_address` to allow you to get internal addresses just as you get external addresses.
 - added `ensure_addresses_cached` to `Wallet` to let offline wallets load and cache addresses in their database
 
+### Sync API change
+
+To decouple the `Wallet` from the `Blockchain` we've made major changes:
+
+- Removed `Blockchain` from Wallet.
+- Removed `Wallet::broadcast` (just use `Blockchain::broadcast`)
+- Depreciated `Wallet::new_offline` (all wallets are offline now)
```
afilini: Typo

```suggestion
- Deprecated `Wallet::new_offline` (all wallets are offline now)
```

src/wallet/mod.rs
```diff
@@ -175,69 +232,8 @@ where
...
-    }
-}
 
-// offline actions, always available
-impl<B, D> Wallet<B, D>
-where
-    D: BatchDatabase,
-{
-    // Return a newly derived address for the specified `keychain`.
+    // Return a newly derived address using the external descriptor
```
afilini: Is it always the external descriptor?

afilini (commented):
Just a couple of minor things, the code looks good to me.

There was one CI job stuck and I've just restarted it, hopefully it was just a random timeout and not an issue with the test.

PR #537, opened 2022-02-02
refactor wallet address caching into its own public method 

a5an0:
### Description

Currently, the only way to ensure that a wallet's internal database has addresses loaded and cached is through `Wallet::sync`: that function generates and caches up to a number of addresses if they aren't already in the database, and then uses the wallet's blockchain client to sync those addresses. If you are using an offline wallet, there is no mechanism to ensure that the database has addresses loaded. This is a problem for usecases like an offline wallet being used as a multisig signer and wanting to validate change addresses as `Wallet::is_mine` will only work properly if the owned-address is loaded in the database.

This PR takes the address caching functionality out of `Wallet::sync` and puts it in a new public method that is available to offline wallets. `Wallet::sync` uses this method internally.

#### All Submissions:

* [X] I've added docs for the new feature
* [X] I've updated `CHANGELOG.md`

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -316,6 +316,50 @@ where
         }
     }
 
+    /// Ensures that there are at least `max_addresses` addresses cached in the database if the
+    /// descriptor is derivable, or 1 address if it is not.
+    /// Will return `Ok(true)` if there are new addresses generated (either external or internal),
+    /// and `Ok(false)` if all the required addresses are already cached. This function is useful to
+    /// explicitly cache addresses in a wallet to do things like check [`Wallet::is_mine`] on
+    /// transaction output scripts.
+    pub fn ensure_addresses_cached(&self, max_addresses: u32) -> Result<bool, Error> {
+        let mut new_addresses_cached = false;
+        let max_address = match self.descriptor.is_deriveable() {
+            false => 0,
+            true => max_addresses,
+        };
+        debug!("max_address {}", max_address);
+        if self
+            .database
+            .borrow()
+            .get_script_pubkey_from_path(KeychainKind::External, max_address.saturating_sub(1))?
+            .is_none()
+        {
+            debug!("caching external addresses");
+            new_addresses_cached = true;
+            self.cache_addresses(KeychainKind::External, 0, max_address)?;
+        }
+
+        if let Some(change_descriptor) = &self.change_descriptor {
+            let max_address = match change_descriptor.is_deriveable() {
+                false => 0,
+                true => max_addresses,
+            };
+
+            if self
+                .database
+                .borrow()
+                .get_script_pubkey_from_path(KeychainKind::Internal, max_address.saturating_sub(1))?
+                .is_none()
+            {
+                debug!("caching internal addresses");
+                new_addresses_cached = true;
+                self.cache_addresses(KeychainKind::Internal, 0, max_address)?;
+            }
+        }
+        Ok(new_addresses_cached)
+    }
+
     /// Return whether or not a `script` is part of this wallet (either internal or external)
     pub fn is_mine(&self, script: &Script) -> Result<bool, Error> {
         self.database.borrow().is_mine(script)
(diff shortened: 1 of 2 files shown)


LLFourn:
Concept ACK. Like @a5an0 mentions, we need this when you give your descriptors over to another machine which generates addresses without your knowledge and then hands you a PSBT to sign. HW wallets like coldcard do this kind of thing too.

An important side note is that caching addresses is not even technically necessary for `sync` I think. You can just generate the new ones on the fly (up until `stop_gap`) as you look them up. Right now our sync logic depends on pre-caching but I want to fix this.

src/wallet/mod.rs
```diff
@@ -316,6 +316,51 @@ where
         }
     }
 
+    /// Ensures that there are at least `max_addresses` or [CACHE_ADDR_BATCH_SIZE] addresses
+    /// cached in the database if the descriptor is derivable, or 1 address if it is not.
+    /// Will return `Ok(true)` if there are new addresses generated (either external or internal),
+    /// and `Ok(false)` if at the required addresses are already cached. This function is used internally
```
notmandatory: I think this is what you meant.
```suggestion
    /// and `Ok(false)` if all the required addresses are already cached. This function is used internally
```

a5an0: Good catch! no typechecker for bad english 😅. will post an update

afilini:
Try rebasing on the lastest `master`, it should fix the CI

afilini:
Concept ACK btw, this seems useful.

> Right now our sync logic depends on pre-caching but I want to fix this.

Yes, this would be great. The current `sync`/`setup` behavior is also very inconsistent, we should think about what to do there.

notmandatory:
Once #502 is merged you'll need to rebase to pickup some CI changes that are in that PR. Then this one looks good to go.

a5an0:
rebased onto master to pick up CI changes

src/wallet/mod.rs
```diff
@@ -316,6 +316,51 @@ where
         }
     }
 
+    /// Ensures that there are at least `max_addresses` or `CACHE_ADDR_BATCH_SIZE` addresses
+    /// cached in the database if the descriptor is derivable, or 1 address if it is not.
+    /// Will return `Ok(true)` if there are new addresses generated (either external or internal),
+    /// and `Ok(false)` if all the required addresses are already cached. This function is used internally
+    /// by [`Wallet::sync`] in wallets that have a [`Blockchain`] client, and is useful for direct
```
LLFourn: Let's not talk about what Wallet::sync does internally since this doc could easily become outdated. After #535 is merged all wallets will be "offline" and can `sync` if/when a blockchain becomes available. So I think it's a good idea to remove the bulk of this sentence. It is worth mentioning that the reason you do this is so that `is_mine` works.

src/wallet/mod.rs
```diff
@@ -316,6 +316,51 @@ where
...
+    /// use in offline wallets where [`Wallet::sync`] is unavailable but addresses need to be loaded
+    /// to do things like check [`Wallet::is_mine`] on transaction output scripts.
+    pub fn ensure_addresses_cached(&self, max_addresses: Option<u32>) -> Result<bool, Error> {
```
LLFourn: I think `CACHE_ADDRESS_SIZE` should not talked about here. `max_addresses` should just be a `u32`.

src/wallet/mod.rs
```diff
@@ -316,6 +316,51 @@ where
...
+        let mut run_setup = false;
```
LLFourn:
```suggestion
        let mut new_addresses_cached = false;
```

`run_setup` doesn't make sense.

LLFourn (changes requested):
This LGTM just a few cosmetic things that I think we should fix up.

a5an0:
Thanks for the feedback! updated.

LLFourn (approved):
ACK 66c6158b29c2d1ee50ad1c2c41442fa9e3f06cb8 squash and merge

notmandatory (approved):
ACK edf2f0ce06d27a1434202b79306cd1e998b5614c

PR #546, opened 2022-02-16
AddressIndex improvements: LastUnused, FirstUnused, and get_batch_unused_addresses()

nickfarrow:
### Description
* Change `AddressIndex::LastUnused` to look back further than current_index
* Add `AddressIndex::FirstUnused`
* Add `get_batch_unused_addresses`

### Notes to the reviewers
Builds upon https://github.com/bitcoindevkit/bdk/pull/522

Currently BDK supports address indexing via `LastUnused`, which will return the address with `current_index` if it is unused, otherwise it will return a New address.

With this current logic, if you get two new addresses A1 and A2 and use A2, then `LastUnused` will give you a `New` address rather than the unused A1.

In order to more consistently utilize unused addresses i've added a new function `get_unused_key_indexes(keychain)` which returns a vector of indexes for the unused addresses in that keychain. Making use of this function, `LastUnused` now returns the most recent address that has not yet been used, and New if all addresses have been used.

In some cases it may be desirable to utilize unused addresses that reside earlier in the keychain. i.e. `AddressIndex::FirstUnused` in this PR. 

`FirstUnused` has the same caveat as LastUnused: that if the wallet has not yet detected an address has been used, it could return a used address.

Additionally a new public function `get_batch_unused_addresses` allows for retrieval of `N` unused addresses at once. Prioritizing unused addresses first, then populating the remaining with New addresses (like `FirstUnused`).

For example: if a wallet builds a transaction involving many `New` internal addresses but that transaction is never broadcast, then all of these addresses can now easily be used in a later transaction via `get_batch_unused_addresses`.

#### All Submissions:

#### New Features:

* [X] I've added tests for the new feature
* [X] I've added docs for the new feature
* [X] I've updated `CHANGELOG.md`

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -457,6 +457,39 @@ where
         Ok(new_addresses_cached)
     }
 
+    /// Return vector of n unused addresses from the [`KeychainKind`].
+    /// If less than n unused addresses are returned, the rest will be populated by new addresses.
+    /// The unused addresses returned are in order of oldest in keychain first, with increasing index.
+    pub fn get_batch_unused_addresses(
+        &self,
+        n: usize,
+        keychain: KeychainKind,
+    ) -> Result<Vec<AddressInfo>, Error> {
+        let unused_key_indexes = self.get_unused_key_indexes(keychain)?;
+        let mut addresses = unused_key_indexes
+            .iter()
+            .map(|i| {
+                let derived_key = self
+                    .get_descriptor_for_keychain(keychain)
+                    .as_derived(*i, &self.secp);
+
+                derived_key
+                    .address(self.network)
+                    .map(|address| AddressInfo {
+                        address,
+                        index: *i,
+                        keychain,
+                    })
+                    .map_err(|_| Error::ScriptDoesntHaveAddressForm)
+            })
+            .take(n)
+            .collect::<Result<Vec<_>, _>>()?;
+        for _ in 0..(n - addresses.len()) {
+            addresses.push(self.get_new_address(keychain)?)
+        }
+        Ok(addresses)
+    }
+
     /// Return whether or not a `script` is part of this wallet (either internal or external)
     pub fn is_mine(&self, script: &Script) -> Result<bool, Error> {
         self.database.borrow().is_mine(script)
@@ -3967,6 +4000,57 @@ pub(crate) mod test {
         assert_eq!(
             wallet.get_address(FirstUnused).unwrap().to_string(),
             "tb1q4er7kxx6sssz3q7qp7zsqsdx4erceahhax77d7"
+        );
+    }
+
+    #[test]
+    fn test_batch_unused_addresses() {
+        let descriptor = "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)";
+        let descriptors = testutils!(@descriptors (descriptor));
+        let wallet = Wallet::new(
+            &descriptors.0,
+            None,
+            Network::Testnet,
+            MemoryDatabase::new(),
+        )
+        .unwrap();
+
+        // get first two addresses, moving index
+        for _ in 0..2 {
+            let _ = wallet.get_address(New);
+        }
+
+        // use the second address
+        crate::populate_test_db!(
+            wallet.database.borrow_mut(),
+            testutils! (@tx ( (@external descriptors, 1) => 25_000 ) (@confirmations 1)),
+            Some(100),
+        );
+
+        assert_eq!(
+            wallet
+                .get_batch_unused_addresses(3, KeychainKind::External)
+                .unwrap(),
+            vec![
+                AddressInfo {
+                    index: 0,
+                    address: Address::from_str("tb1q6yn66vajcctph75pvylgkksgpp6nq04ppwct9a")
+                        .unwrap(),
+                    keychain: KeychainKind::External,
+                },
+                AddressInfo {
+                    index: 2,
+                    address: Address::from_str("tb1qzntf2mqex4ehwkjlfdyy3ewdlk08qkvkvrz7x2")
+                        .unwrap(),
+                    keychain: KeychainKind::External,
+                },
+                AddressInfo {
+                    index: 3,
+                    address: Address::from_str("tb1q32a23q6u3yy89l8svrt80a54h06qvn7gnuvsen")
+                        .unwrap(),
+                    keychain: KeychainKind::External,
+                }
+            ]
         );
     }
 
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -260,26 +260,25 @@ where
             .map_err(|_| Error::ScriptDoesntHaveAddressForm)
     }
 
-    // Return whether this address has been used in a transaction
-    fn has_address_been_used(&self, script_pk: &Script) -> bool {
-        let txns = self.list_transactions(true).unwrap_or_else(|_| vec![]);
-        txns.iter()
-            .flat_map(|tx_details| tx_details.transaction.as_ref())
-            .flat_map(|tx| tx.output.iter())
-            .any(|o| o.script_pubkey == *script_pk)
-    }
-
     // Return the the last previously derived address for `keychain` if it has not been used in a
     // received transaction. Otherwise return a new address using [`Wallet::get_new_address`].
     fn get_last_unused_address(&self, keychain: KeychainKind) -> Result<AddressInfo, Error> {
-        let mut unused_addresses = self.get_batch_unused_addresses(1, false, keychain)?;
-        Ok(unused_addresses.remove(0))
+        let unused_script_indexes = self.get_unused_script_indexes(keychain)?;
+        let current_index = &self.fetch_index(keychain)?;
+        if unused_script_indexes.contains(current_index) {
+            self.get_address(AddressIndex::Peek(*current_index))
+        } else {
+            self.get_new_address(keychain)
+        }
     }
 
-    // Return the the first address in the keychain which has not been a recipient of a transaction
+    // Return the the first address in the keychain which has not been used in a recieved transaction
+    // If they have all been used, return a new address using [`Wallet::get_new_address`].
     fn get_first_unused_address(&self, keychain: KeychainKind) -> Result<AddressInfo, Error> {
-        let mut unused_addresses = self.get_batch_unused_addresses(1, true, keychain)?;
-        Ok(unused_addresses.remove(0))
+        self.get_unused_script_indexes(keychain)?
+            .get(0)
+            .map(|index| self.get_address(AddressIndex::Peek(*index)))
+            .unwrap_or_else(|| self.get_new_address(keychain))
     }
 
     // Return derived address for the descriptor of given [`KeychainKind`] at a specific index
@@ -387,60 +386,39 @@ where
         Ok(new_addresses_cached)
     }
 
-    /// Return vector of n unused addresses from the [`KeychainKind`].
-    /// If less than n unused addresses are returned, the rest will be populated by new addresses.
-    /// The unused addresses returned are in order of oldest in keychain first, with increasing index.
-    pub fn get_batch_unused_addresses(
-        &self,
-        n: usize,
-        from_front: bool,
-        keychain: KeychainKind,
-    ) -> Result<Vec<AddressInfo>, Error> {
+    /// Return set of unused script indexes for the [`KeychainKind`].
+    pub fn get_unused_script_indexes(&self, keychain: KeychainKind) -> Result<Vec<u32>, Error> {
         let script_pubkeys = self
             .database
             .borrow()
             .iter_script_pubkeys(Some(keychain))
             .unwrap_or_else(|_| vec![]);
-
+        let txs = self.list_transactions(true).unwrap_or_else(|_| vec![]);
+        let tx_scripts: HashSet<&Script> = txs
+            .iter()
+            .flat_map(|tx_details| tx_details.transaction.as_ref())
+            .flat_map(|tx| tx.output.iter())
+            .map(|o| &o.script_pubkey)
+            .collect();
         let current_address_index = self.fetch_index(keychain)? as usize;
-        let check_indexes = if from_front {
-            (0..=current_address_index).collect::<Vec<_>>()
-        } else {
-            (0..=current_address_index).rev().collect::<Vec<_>>()
-        };
 
-        let mut unused_addresses = vec![];
-        for i in check_indexes {
-            // if we have made a pubkey at this index, check whether the address has been used.
-            if i < script_pubkeys.len() {
-                let script_pk = &script_pubkeys[i];
-                if self.has_address_been_used(script_pk) {
-                    continue;
+        let mut scripts_not_used: Vec<u32> = script_pubkeys
+            .iter()
+            .take(current_address_index + 1)
+            .enumerate()
+            .filter_map(|(i, script_pubkey)| {
+                if !tx_scripts.contains(script_pubkey) {
+                    Some(i as u32)
+                } else {
+                    None
                 }
-            }
-            if let Ok(unused_address) = self
-                .get_descriptor_for_keychain(keychain)
-                .as_derived(i as u32, &self.secp)
-                .address(self.network)
-                .map(|address| AddressInfo {
-                    address,
-                    index: i as u32,
-                    keychain,
-                })
-                .map_err(|_| Error::ScriptDoesntHaveAddressForm)
-            {
-                unused_addresses.push(unused_address);
-            }
-
-            if unused_addresses.len() >= n {
-                break;
-            }
+            })
+            .collect();
+        if script_pubkeys.is_empty() {
+            scripts_not_used.push(0);
         }
 
-        for _ in 0..(n - unused_addresses.len()) {
-            unused_addresses.push(self.get_new_address(keychain)?)
-        }
-        Ok(unused_addresses)
+        Ok(scripts_not_used)
     }
 
     /// Return whether or not a `script` is part of this wallet (either internal or external)
@@ -3927,7 +3905,7 @@ pub(crate) mod test {
     }
 
     #[test]
-    fn test_firstunused_address() {
+    fn test_first_unused_address() {
         let descriptor = "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)";
         let descriptors = testutils!(@descriptors (descriptor));
         let wallet = Wallet::new(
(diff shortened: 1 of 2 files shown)


LLFourn:
approach ACK. Needs rebase.

notmandatory:
Code changes look great but you'll need to do another rebase and can you add a signing key to Github and [sign your commits] when you do the rebase also? 

[sign your commits]: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

nickfarrow:
force push updated CI

src/wallet/mod.rs
```diff
@@ -253,32 +261,91 @@ where
...
+                    .as_derived(address_index, &self.secp);
+
+                let script_pubkey = derived_key.script_pubkey();
+
+                let found_used = self
+                    .list_transactions(true)?
+                    .iter()
+                    .flat_map(|tx_details| tx_details.transaction.as_ref())
+                    .flat_map(|tx| tx.output.iter())
+                    .any(|o| o.script_pubkey == script_pubkey);
```
rajarshimaitra: This check is not needed in this function anymore. We already checked that this key wasn't used in `get_unused_key_indexes`..

src/wallet/mod.rs
```diff
@@ -115,6 +116,13 @@ pub enum AddressIndex {
     /// caller is untrusted; for example when deriving donation addresses on-demand for a public
     /// web page.
     LastUnused,
+    /// Return the address for the first address in the keychain that has not been used in a received
+    /// transaction. Otherwise return a new address as with [`AddressIndex::New`].
+    ///
+    /// Use with caution, if the wallet has not yet detected an address has been used it could
+    /// return an already used address. This function is primarily meant for making use of addresses earlier
+    /// in the keychain that were infact never used.
+    FirstUnused,
```
rajarshimaitra, line 124: I feel the caution message is not necessary anymore?? As BDK will not give up used address anymore after this change??

src/wallet/mod.rs
```diff
@@ -115,6 +116,13 @@ pub enum AddressIndex {
     /// caller is untrusted; for example when deriving donation addresses on-demand for a public
     /// web page.
     LastUnused,
```
rajarshimaitra, line 117: The comment above this can also be updated with current changed behavior..

src/wallet/mod.rs
```diff
@@ -253,32 +261,91 @@ where
...
+            })
+            .collect();
+
+        Ok(unused_key_indexes)
+    }
 
-        if found_used {
+    // Return the the last previously derived address for `keychain` if it has not been used in a
+    // received transaction. Otherwise return a new address using [`Wallet::get_new_address`].
+    fn get_last_unused_address(&self, keychain: KeychainKind) -> Result<AddressInfo, Error> {
```
rajarshimaitra, line 265: I am thinking if we already got `get_first_unused_address` what we can use `get_last_unused_address` for?? Isn't it always better to use the first unused in all situations?

src/wallet/mod.rs
```diff
@@ -3872,6 +3973,87 @@ pub(crate) mod test {
...
+        crate::populate_test_db!(
+            wallet.database.borrow_mut(),
+            testutils! (@tx ( (@external descriptors, 0) => 25_000 ) (@confirmations 1)),
+            Some(100),
+        );
+
+        assert_eq!(
+            wallet.get_address(FirstUnused).unwrap().to_string(),
+            "tb1q4er7kxx6sssz3q7qp7zsqsdx4erceahhax77d7"
+        );
```
rajarshimaitra, line 3934: I would like to see the test situation here where we extract multiple addresses, use some of them and get back a previous unused one when called again.. That would correctly test the intended behavior.. Right now its just testing the vanilla situation..

rajarshimaitra (changes requested):
tACK 05aeb2306ddafb95b49b025b55ed3d260b816cb8

Below are some suggested modification..

src/wallet/mod.rs
LLFourn: Not necessarily I think. `last_unused_address` basically means the last address you gave out (unless it has been used). It's useful when you need a new address but you can't remember if you used the last one the app gave out to you.

[EDIT] it looks like this PR changes the behaviour of `last_unused` which I don't think was intended.

src/wallet/mod.rs
```diff
@@ -253,32 +261,91 @@ where
...
             .iter()
-            .flat_map(|tx_details| tx_details.transaction.as_ref())
-            .flat_map(|tx| tx.output.iter())
-            .any(|o| o.script_pubkey == script_pubkey);
+            .cloned()
+            .enumerate()
+            .filter_map(|(i, key)| {
+                if !self
+                    .list_transactions(true)
+                    .expect("getting transaction list")
```
LLFourn: don't panic.

src/wallet/mod.rs
```diff
@@ -253,32 +261,91 @@ where
...
-            .get_descriptor_for_keychain(keychain)
-            .as_derived(current_index, &self.secp);
-
-        let script_pubkey = derived_key.script_pubkey();
+        let derived_keys: Vec<Descriptor<DerivedDescriptorKey>> = (0..current_index + 1)
+            .map(|i| {
+                self.get_descriptor_for_keychain(keychain)
+                    .as_derived(i, &self.secp)
+            })
+            .collect();
```
LLFourn: This should be using the scripts (`iter_script_pubkeys`) in the database not re-deriving them which is costly.

src/wallet/mod.rs
```diff
@@ -253,32 +261,91 @@ where
             .map_err(|_| Error::ScriptDoesntHaveAddressForm)
     }
 
-    // Return the the last previously derived address for `keychain` if it has not been used in a
-    // received transaction. Otherwise return a new address using [`Wallet::get_new_address`].
-    fn get_unused_address(&self, keychain: KeychainKind) -> Result<AddressInfo, Error> {
+    // Return vector of unused address indexes
+    fn get_unused_key_indexes(&self, keychain: KeychainKind) -> Result<Vec<u32>, Error> {
```
LLFourn: Wouldn't it be simpler and more efficient just to do `has_address_been_used` and return a `bool`? You are iterating over all the the scripts and checking them all even when you may only be interested in a particular one. The alternative would be to return a `HashSet` of used addresses or something similar.

src/wallet/mod.rs
```diff
@@ -389,6 +457,39 @@ where
...
+    /// Return vector of n unused addresses from the [`KeychainKind`].
+    /// If less than n unused addresses are returned, the rest will be populated by new addresses.
+    /// The unused addresses returned are in order of oldest in keychain first, with increasing index.
+    pub fn get_batch_unused_addresses(
+        &self,
+        n: usize,
+        keychain: KeychainKind,
+    ) -> Result<Vec<AddressInfo>, Error> {
+        let unused_key_indexes = self.get_unused_key_indexes(keychain)?;
+        let mut addresses = unused_key_indexes
```
LLFourn: So to line up with my other comments this should just iter over script pubkeys from database and check if each one is used successively until you've found `n` of them (and then of course get more if you need).

src/wallet/mod.rs
LLFourn: Hmm I actually don't know a better test to write than this one? With the batch unused you can write a more complicated test but with `FristUnused` there's not much you can do.

LLFourn (changes requested):
Thanks @nickfarrow. Tests LGTM. See comments.

src/wallet/mod.rs
rajarshimaitra: Something like derive a bunch of address.. Only use some of them selectively so the address gaps are simulated.. Then check if the first unused is returned correctly.. Am I missing some details why that can't work??

src/wallet/mod.rs
rajarshimaitra: But even in the case if I don't remember if I used the last one I fetched, won't I still be wanting the first unused?? Won't that always ensure that address use gaps are minimized over the natural usage of the wallet?? 

If I ask last unused there might be other addresses too I could have used, just didn't knew.. So those gap remains until someone calls first unused again..

LLFourn: I don't have a strong rationale for `last_unused`'s existence tbh. We could remove it without complaint from me :)

src/wallet/mod.rs
LLFourn: It can work but I don't get why the gaps would effect the algorithm that finds the **first** unused. I mean I don't think that this will likely find a problem with the algorithm that this test wouldn't find.

src/wallet/mod.rs
nickfarrow: Yep I have changed LastUnused here to 'go back' and find the last unused address, rather than just checking only the address at current index and returning new otherwise. If someone gets addresses A1 and A2, and then uses A2. Do we want `LastUnused` to return new or A1?

This behavior made more sense to me for `LastUnused` address, not something like `CurrentAddressIfUnused`. Do you think this behavior is unexpected or too complex?

I agree that `FirstUnused` is better for minimizing address gaps, and yeah not exactly sure of when `LastUnused` would be preferable. Other than lazily getting the address you're about to use in something like
>when you need a new address but you can't remember if you used the last one the app gave out to you.

src/wallet/mod.rs
nickfarrow: I believe this caution is referring to the possibility that of:
1. Get a `LastUnused` address (or FirstUnused)
2. Use this address in some way that BDK doesn't immediately know of. Like in an unbroadcasted txn or an external address you intend to receive money to at some point in the future.
3. Then get an `LastUnused` address again (the same address) and use it for something else.

At later time, the address is reused once the unbroadcasted txn or delayed receive from step-2 occurs.

src/wallet/mod.rs
rajarshimaitra: Its not that the gaps would affect the algorithm, but to confirm that the behavior we are intending here is actually happening.. And this can be checked in single test for both first and last unused.. Once the behavior is pinned, we can decide later which one to use when or to keep both..

src/wallet/mod.rs
rajarshimaitra: I think for now its ok to keep both of them.. I think the documentation can be improved a bit to explain to users which one does what. Maybe with an added example like you given above.. 

We can decide later if we want to discard any..

src/wallet/mod.rs
rajarshimaitra: Ah yes that makes sense.. If the wallet doesnt detect transaction it would not consider it used..

notmandatory:

src/wallet/mod.rs
nickfarrow: Are you thinking we leave the current LastUnused as it currently exists? or rename to `CurrentAddressIfUnused` and implement `LastUnused` to mean "give me the last address that hasn't been used, otherwise new"

src/wallet/mod.rs
```diff
@@ -253,36 +260,26 @@ where
...
-            .get_descriptor_for_keychain(keychain)
-            .as_derived(current_index, &self.secp);
-
-        let script_pubkey = derived_key.script_pubkey();
-
-        let found_used = self
-            .list_transactions(true)?
-            .iter()
+    // Return whether this address has been used in a transaction
+    fn has_address_been_used(&self, script_pk: &Script) -> bool {
```
nickfarrow: actually checking a Script..

src/wallet/mod.rs
```diff
@@ -389,6 +387,62 @@ where
...
+            .database
+            .borrow()
+            .iter_script_pubkeys(Some(keychain))
+            .unwrap_or_else(|_| vec![]);
+
+        let current_address_index = self.fetch_index(keychain)? as usize;
+        let check_indexes = if from_front {
+            (0..=current_address_index).collect::<Vec<_>>()
+        } else {
+            (0..=current_address_index).rev().collect::<Vec<_>>()
```
nickfarrow: better way to do this?

src/wallet/mod.rs
```diff
@@ -389,6 +387,62 @@ where
         Ok(new_addresses_cached)
     }
 
+    /// Return vector of n unused addresses from the [`KeychainKind`].
+    /// If less than n unused addresses are returned, the rest will be populated by new addresses.
+    /// The unused addresses returned are in order of oldest in keychain first, with increasing index.
```
rajarshimaitra: This is not as per impl right now.. if `from_front` is set false, the addresses are returned in reverse order..

src/wallet/mod.rs
rajarshimaitra: This guy is better named as `is_scriptpubkey_used`..

src/wallet/mod.rs
```diff
@@ -389,6 +387,62 @@ where
...
+    pub fn get_batch_unused_addresses(
+        &self,
+        n: usize,
+        from_front: bool,
+        keychain: KeychainKind,
+    ) -> Result<Vec<AddressInfo>, Error> {
```
rajarshimaitra: I am not feeling comfortable with the API. `get_batched_unused` should not be concerned with `fornt` or `back`. Thats an impl detail for `LastUnused` or `FirstUnused`. And should not be exposed in public API..

This is also breaking the doc above. The order is not maintained anymore.. 

Better to handle the handle the first or last logic in in their respective functions itself than to handle in the batch function which is more generic.

src/wallet/mod.rs
```diff
@@ -389,6 +387,62 @@ where
...
+                })
+                .map_err(|_| Error::ScriptDoesntHaveAddressForm)
+            {
+                unused_addresses.push(unused_address);
+            }
+
+            if unused_addresses.len() >= n {
+                break;
+            }
+        }
```
rajarshimaitra: Try using some rust list comprehensions with iters and maps. Much of this code can be simplified..

src/wallet/mod.rs
```diff
@@ -3872,6 +3926,87 @@ pub(crate) mod test {
...
+                },
+                AddressInfo {
+                    index: 3,
+                    address: Address::from_str("tb1q32a23q6u3yy89l8svrt80a54h06qvn7gnuvsen")
+                        .unwrap(),
+                    keychain: KeychainKind::External,
+                }
+            ]
+        );
+    }
```
rajarshimaitra: Also need to assert that FirstUnused and LastUnused are working as intended..
 - Get 5 new addresses
 - Use only index 0 and 3
 - get_batch(3) should return index 0, 2, 4. current index should still be at 4.
 - get_first_unused() should return 0
 - get_last_unused() should return 4
 - get_batch(4) should return 0,2,4,5, and current index should be at 5.

src/wallet/mod.rs
```diff
@@ -389,6 +387,62 @@ where
...
+        } else {
+            (0..=current_address_index).rev().collect::<Vec<_>>()
+        };
+
+        let mut unused_addresses = vec![];
+        for i in check_indexes {
+            // if we have made a pubkey at this index, check whether the address has been used.
+            if i < script_pubkeys.len() {
+                let script_pk = &script_pubkeys[i];
+                if self.has_address_been_used(script_pk) {
```
rajarshimaitra: Here for each spk we are iterating over the entire transaction list. For wallets with large transaction this will can cause massive overhead.

Instead a better way would be to handle `Vec<Script>` in the `has_address_been_used` function. Call `list_transactions` only once, filter out all the spks that haven't been used and return then as a Vec.

rajarshimaitra (changes requested):
I am still not comfortable with the current state and approach of the PR. Some behaviour are not maintained, like get_batch(n, false, keychain) will give the list of addresses in reverse order. Not in ascending order of indexes.

Also I am thinking isn't it better to mark used addresses directly in the database? Knowing which one is used and not and saving the data seems to me more useful than figuring it out by transaction matching with the entire tx list, everytime we ask for an unused.. 

This will also simplify the LastUnused and FirstUnused fetching logic..

@LLFourn @afilini do you have any thought on this??

rajarshimaitra:
@nickfarrow also try to rebase on top of master instead of fetching and merging specific commit next time.. :)
That makes the commit history much cleaner and also applies your changes on top of current master..
just do `git rebase master` from the PR branch..

notmandatory:
I had to push this PR to the next release so the team can focus on #593, and after that one you'll probably need to rebase again. But then I promise we'll work on getting this one in. :-)

src/wallet/mod.rs
nickfarrow: Yep good idea, note this would then also check more addresses than necessary (not breaking early when finding n).
If this was stored in the database it could also break early if it's just a fast read

nickfarrow:
Yep I think this one needs some bigger discussion first around whether it is worthwhile to mark used addresses directly in the database as @rajarshimaitra suggested.

If this were the case, this could be simplified to a function `get_unused_addresses` that (quickly) gets an iterator over unused addresses in the database up to the current addressindex. With `FirstUnused` and `LastUnused` addresses at either ends.

This PR's `get_batch_unused_addresses`  (currently fetches `n`) would be superseded by `get_unused_addresses` where the user gets all the unused addresses and handles them how they desire.

Not sure what changed with old commits, possible I added signoff lines to the previous commits by mistake which may have updated them sorry. Will take care with next.

rajarshimaitra:
Upon further thoughts on this I have this rough idea of how it can be done:
 - Add a new column in the `script_pubkey` table named `used`. Which will have binary value 0/1, defaulting to 0.
 - At each `sync` when we add `utxo`s to the `Utxo` table, mark the pubkey as `used` too in the script_pubkey table.
 - Make a generic `get_batch_unused()` that will return a list of unused addresses, checking the `used` flag in the database.
 - Make `FirstUnused` as the front pop of the list, and `LastUnused` as the back pop of the list.. 
 
Pro:
 Much more scalable than transaction list scanning for wallets with large list of transactions. 
 
 Cons: 
 This is going to change the DB structure and the `BatchDatabase` API.
 
 I am willing to work on fleshing an impl out if this has Approach Acks..

afilini:
Keep in mind we also have the key/value db, we don't have tables and columns there. We can add a flag to mark a script as used (similarly to how I suggested adding a flag for scripts that we've already setup rather than relying on just the derivation index), but getting the list of unused addresses will still require scanning

src/wallet/mod.rs
LLFourn: `LastUnused`'s behaviour shouldn't change here.

src/wallet/mod.rs
LLFourn: The correct thing is to return a `impl DoubleEndedIterator` over unused addresses I think.

src/wallet/mod.rs
LLFourn: Return a [`impl DoubleEndedIterator`](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html) from the method instead.

src/wallet/mod.rs
LLFourn: Adding a check of the derivation index here would be good.

LLFourn (changes requested):
This PR is almost ready and provides something that is pretty useful. The main bits of work here are:

1. To restore the previous (someone nonsensical) behaviour of `LastUnused`
2. To add tests to check the wallets derivation index is correct after calling batch unused (can just try and get a new address after and check its index)

As @rajarshimaitra mentions the best way to implement is to index things properly which in bdk is currently done in the database backend. In `bdk_core` I've done indexing of unused addresses. Since I was the one who requested this feature and I'm focused on `bdk_core` we could simply close this PR and wait until it lands. Does anyone else want this feature presently? @nickfarrow what do you think?

rajarshimaitra:
ACK on @LLFourn that this can go in as it is without much further changes.. I will check the behavior once again.. If this lands through `bdk_core` eventually then we might not wanna do the `DataBase` way for now, and just keep things simple..

notmandatory:
Is this one OK to add this to the `0.22.0` milestone? looks like it's about ready and I don't want it to be overlooked.

LLFourn:
IMO this PR is suboptimal because of #701. I think it should be fixed first to make the code in this PR make sense. @nickfarrow?

nickfarrow:
Yep agree it may as well wait for improvements to `fetch_index`, that function is relied upon a few times here

danielabrozzoni:

PR #547, opened 2022-02-17
Document MSRV policy and change MSRV to 1.49.0

### LINKED ISSUE #331: Support for MSRV shipped with Debian stable for default features

### LINKED ISSUE #496: Consider not changing the version in `Cargo.toml` before the release is cut
In order to depend on unreleased versions of libraries, `cargo` offers a way to patch a certain version of a dependency via the `[patch.crates-io]`.

This however only works if the version mentioned in the project `Cargo.toml` file is the same one as the one set in the dependency's manifest.

I am currently depending on bdk 0.14 and trying to patch `bdk` across my dependency graph to latest HEAD. Unfortunately, that doesn't work:

```
❯ cargo t
    Updating git repository `https://github.com/bitcoindevkit/bdk`
warning: Patch `bdk v0.14.1-dev (https://github.com/bitcoindevkit/bdk#64e88f0e)` was not used
in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
```

Do people use the `[patch.crates-io]` feature? If yes, how are you dealing with pre-release version bumps in the manifest file?

notmandatory:
### Description

Set MSRV policy to be at most Debian "testing" release version of rustc. Update `build-test` CI job matrix to use the current latest stable version and MSRV ~~1.56.0~~ 1.49.0.

Fixes #331
Fixes #496 

### Notes to the reviewers

Fixes CI error caused by tokio MSRV changing to 1.49.0, see #522.

I also changed from a fixed stable version (was `1.56.0`) to the latest `stable` because we haven't been updating it manually and this will ensure we're always testing against current `stable` rust.

#### All Submissions:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -10,9 +10,9 @@ jobs:
     strategy:
       matrix:
         rust:
-          - version: 1.56.0 # STABLE
+          - version: stable # STABLE
             clippy: true
-          - version: 1.46.0 # MSRV
+          - version: 1.49.0 # MSRV
         features:
           - default
           - minimal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Increase project MSRV from 1.46.0 to 1.49.0, see new policy in README
+
 ## [v0.16.0] - [v0.15.0]
 
 - Disable `reqwest` default features.
--- a/README.md
+++ b/README.md
@@ -170,6 +170,16 @@ cargo test --features test-electrum
 The other options are `test-esplora` or `test-rpc`.
 Note that `electrs` and `bitcoind` binaries are automatically downloaded (on mac and linux), to specify you already have installed binaries you must use `--no-default-features` and provide `BITCOIND_EXE` and `ELECTRS_EXE` as environment variables.
 
+## Supported Rust Versions
+
+The BDK MSRV (minimum supported rust version) will be as low as possible to support required 
+dependencies and no higher than the [rustc](https://tracker.debian.org/pkg/rustc) version packaged 
+with the Debian ["testing"] release, and if possible the ["stable"] release. We will only increase 
+the BDK MSRV with a new MINOR release, never a PATCH release.
+
+["testing"]: https://www.debian.org/releases/testing/
+["stable"]: https://www.debian.org/releases/stable/
+
 ## License
 
 Licensed under either of
--- a/examples/compiler.rs
+++ b/examples/compiler.rs
@@ -85,7 +85,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     let network = matches
         .value_of("network")
-        .map(|n| Network::from_str(n))
+        .map(Network::from_str)
         .transpose()
         .unwrap()
         .unwrap_or(Network::Testnet);
--- a/src/blockchain/any.rs
+++ b/src/blockchain/any.rs
@@ -108,6 +108,7 @@ macro_rules! impl_inner_method {
 /// It allows switching backend at runtime
 ///
 /// See [this module](crate::blockchain::any)'s documentation for a usage example.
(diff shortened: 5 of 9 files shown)


notmandatory:
Prior to merging I also need to change the required github checks to ~~`1.56.0`~~ `1.49.0` and `stable`.

afilini:
I would avoid bumping the MSRV unless we get stuck or have issues somewhere. Maybe our policy could be "at most" debian testing, meaning that we will never go above, but we'll still try to support older versions if we can

notmandatory:
I should have linked to the PR that triggered this PR: https://github.com/bitcoindevkit/bdk/pull/522#issuecomment-1042393238

Maybe we don't need to bump the MSRV all the way to 1.56.0, I'll see if we can get away with rust `1.48.0`

notmandatory:
Looks like the [`tokio` MSRV](https://github.com/tokio-rs/tokio#supported-rust-versions) is now `1.49.0`.

notmandatory:
As mentioned by @moneyball on Discord, the LDK team pinned their `tokio` version to `~1.14` instead of increasing their MSRV, see https://github.com/lightningdevkit/rust-lightning/pull/1315. I believe one downside of this approach is we won't be getting new `tokio` fixes or features unless that team plans to back port them to `1.14.x` releases.

TheBlueMatt:
Note that tokio *does* backport fixes, so I'm not sure how much of a concern it is in practice. As for MSRV of 1.49, that seems exceedingly recent - notably most linux distros still ship 1.41 (and rustup is rather unacceptable for anything security-conscious, like Bitcoin).

notmandatory:
@TheBlueMatt, good points. I created https://github.com/bitcoindevkit/bdk/pull/550 as a fix to get our CI working and then we can take our time to decide about any MSRV changes.

rajarshimaitra (approved):
Review ACK https://github.com/bitcoindevkit/bdk/pull/547/commits/7469bd97e66f6888eb1143f40d1d21683cb116f4

edit: Just realized this might not be ready yet.. Will see again once final..

ulrichard:
I like and support this change. 
We recently upgraded all our other stuff to tokio 1.17 and it worked well with bdk up to 0.16. Now I tried upgrading bdk to 0.17 and got conflicts. 
With this change in place, we could upgrade to a newer bdk, without having to downgrade tokio.

notmandatory:
Another transitive dependency, `http`, has now updated their MSRV to 1.49, see #595.

TheBlueMatt:
Note that even if a dep updates their MSRV you can keep the looser bound on the MSRV and users can pin-back with something like `cargo update -p tokio --precise "1.14.1"`, which works to make the ldk-sample build with 1.41 even though by default it will try to build with a newer tokio.

notmandatory:
> Note that even if a dep updates their MSRV you can keep the looser bound on the MSRV and users can pin-back with something like `cargo update -p tokio --precise "1.14.1"`, which works to make the ldk-sample build with 1.41 even though by default it will try to build with a newer tokio.

That does seem like a clean way to do it, but if we used this approach would I also need to add that command to my CI pipeline matrix for testing with 1.46? If so I think it's easier to pin it in the `Cargo.toml` for now.

TheBlueMatt:
Yes, you'd have to do that in CI as well. I suppose it'd be one thing if it were a truly ancient rustc, but 1.41 is still the version shipped on many distros, so its kinda hard to demand that for every user :/

danielabrozzoni:

PR #550, opened 2022-02-17
Pin tokio version to ~1.14

notmandatory:
### Description

The `tokio` project recently changed their MSRV to `1.49.0`. This PR will pin the `tokio` dependency version to `~1.14` which is prior to their MSRV increase.

### Notes to the reviewers

The LDK team took this approach for their `tokio` dev-dependency, see https://github.com/lightningdevkit/rust-lightning/pull/1315. 

As long as `tokio` backports bug fixes  to`1.14.x` releases this should be safe. If we need any new features we can revisit this decision.

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Pin tokio dependency version to ~1.14 to prevent errors due to their new MSRV 1.49.0
+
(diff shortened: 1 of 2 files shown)


notmandatory:
I think we can use this as a short term fix while we discuss #331 to determine what our MSRV should be.

LLFourn:
ACK b04bb590f385db349f8dc70cc169674b530c4205

I agree this is the right thing to do for now.

LLFourn (approved):

Issue #120, opened 2020-10-06
Add oldest first coin selection next to largest first coin selection


murchandamus:
According to @afilini in the developer meeting today, BDK currently uses "dumb" largest first selection. Largest first selection tends to create small input sets in the short term but tends to grind the biggest UTXO to smaller chunks until it eventually causes a huge input set.

As discussed, it would be good to replace it with something that has a better average performance such as oldest first selection or random selection. Optimally, all of the above would be run and the best input set would be chosen per a simple heuristic among the candidate input sets. 

This task should coordinate with implementation efforts for #48.

kafaichoi:
I am interested in contributing to bdk. Can I work on this issue?

notmandatory:
Hi, yes it's all yours, I'll put your name on it. If you have any questions feel free to comment here or on the Bitcoin Dev Kit discord.

csralvall:
Hi! this issue also mentions random selection.
I see that SRD is implemented as a method for the BnB algorithm, but It might be useful to trait each one as a different algorithm, as in Bitcoin Core.
Do you think that it will be useful?

kafaichoi:
just made a draft PR https://github.com/bitcoindevkit/bdk/pull/557. I will really appreciate for a quick look to see if it's heading to right direction😬

danielabrozzoni:
Closed by #557

PR #148, opened 2020-10-27
[wallet] Add branch and bound coin selection

danielabrozzoni:
Fixes #48

PR #557, opened 2022-03-06
add OldestFirstCoinSelection

kafaichoi:
### Description
This PR is to add `OldestFirstCoinSelection`. See this issue for detail https://github.com/bitcoindevkit/bdk/issues/120

### Notes to the reviewers
Apologize in advance if the quality of this PR is too low.(I am newbie in both bitcoin wallet and rust).  

While this PR seemed very straight-forward to me in the first glance, it's actually a bit more complicated than I thought as it involves calling DB get the blockheight before sorting it.  

The current implementation should be pretty naive but I would like to get some opinion to see if I am heading to a right direction first before working on optimizations like

~~1. Avoiding calling DB for optional_utxos if  if the amount from required_utxos are already enough.~~ Probably not worth to do such optimization to keep code simpler?

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -97,6 +97,7 @@ use rand::seq::SliceRandom;
 use rand::thread_rng;
 #[cfg(test)]
 use rand::{rngs::StdRng, SeedableRng};
+use std::collections::HashMap;
 use std::convert::TryInto;
 
 /// Default coin selection algorithm used by [`TxBuilder`](super::tx_builder::TxBuilder) if not
@@ -199,6 +200,102 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
                 .into_iter()
                 .map(|utxo| (true, utxo))
                 .chain(optional_utxos.into_iter().rev().map(|utxo| (false, utxo)))
+        };
+
+        // Keep including inputs until we've got enough.
+        // Store the total input value in selected_amount and the total fee being paid in fee_amount
+        let mut selected_amount = 0;
+        let selected = utxos
+            .scan(
+                (&mut selected_amount, &mut fee_amount),
+                |(selected_amount, fee_amount), (must_use, weighted_utxo)| {
+                    if must_use || **selected_amount < amount_needed + **fee_amount {
+                        **fee_amount +=
+                            fee_rate.fee_wu(TXIN_BASE_WEIGHT + weighted_utxo.satisfaction_weight);
+                        **selected_amount += weighted_utxo.utxo.txout().value;
+
+                        log::debug!(
+                            "Selected {}, updated fee_amount = `{}`",
+                            weighted_utxo.utxo.outpoint(),
+                            fee_amount
+                        );
+
+                        Some(weighted_utxo.utxo)
+                    } else {
+                        None
+                    }
+                },
+            )
+            .collect::<Vec<_>>();
+
+        let amount_needed_with_fees = amount_needed + fee_amount;
+        if selected_amount < amount_needed_with_fees {
+            return Err(Error::InsufficientFunds {
+                needed: amount_needed_with_fees,
+                available: selected_amount,
+            });
+        }
+
+        Ok(CoinSelectionResult {
+            selected,
+            fee_amount,
+        })
+    }
+}
+
+/// OldestFirstCoinSelection always picks the utxo with the smallest blockheight to add to the selected coins next
+///
+/// This coin selection algorithm sorts the available UTXOs by blockheight and then picks them starting
+/// from the oldest ones until the required amount is reached.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct OldestFirstCoinSelection;
+
+impl<D: Database> CoinSelectionAlgorithm<D> for OldestFirstCoinSelection {
+    fn coin_select(
+        &self,
+        database: &D,
+        required_utxos: Vec<WeightedUtxo>,
+        mut optional_utxos: Vec<WeightedUtxo>,
+        fee_rate: FeeRate,
+        amount_needed: u64,
+        mut fee_amount: u64,
+    ) -> Result<CoinSelectionResult, Error> {
+        // query db and create a blockheight lookup table
+        let blockheights = optional_utxos
+            .iter()
+            .map(|wu| wu.utxo.outpoint().txid)
+            // fold is used so we can skip db query for txid that already exist in hashmap acc
+            .fold(Ok(HashMap::new()), |bh_result_acc, txid| {
+                bh_result_acc.and_then(|mut bh_acc| {
+                    if bh_acc.contains_key(&txid) {
+                        Ok(bh_acc)
+                    } else {
+                        database.get_tx(&txid, false).map(|details| {
+                            bh_acc.insert(
+                                txid,
+                                details.and_then(|d| d.confirmation_time.map(|ct| ct.height)),
+                            );
+                            bh_acc
+                        })
+                    }
+                })
+            })?;
+
+        // We put the "required UTXOs" first and make sure the optional UTXOs are sorted from
+        // oldest to newest according to blocktime
+        // For utxo that doesn't exist in DB, they will have lowest priority to be selected
+        let utxos = {
+            optional_utxos.sort_unstable_by_key(|wu| {
+                match blockheights.get(&wu.utxo.outpoint().txid) {
+                    Some(Some(blockheight)) => blockheight,
+                    _ => &u32::MAX,
+                }
+            });
+
+            required_utxos
+                .into_iter()
+                .map(|utxo| (true, utxo))
+                .chain(optional_utxos.into_iter().map(|utxo| (false, utxo)))
         };
 
         // Keep including inputs until we've got enough.
@@ -541,7 +638,7 @@ mod test {
     use bitcoin::{OutPoint, Script, TxOut};
 
     use super::*;
-    use crate::database::MemoryDatabase;
+    use crate::database::{BatchOperations, MemoryDatabase};
     use crate::types::*;
     use crate::wallet::Vbytes;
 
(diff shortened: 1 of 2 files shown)


src/wallet/coin_selection.rs
```diff
@@ -242,6 +243,102 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
...
+                .into_iter()
+                .map(|utxo| (true, utxo))
+                .chain(optional_utxos.into_iter().map(|utxo| (false, utxo)))
+        };
+
+        // Keep including inputs until we've got enough.
+        // Store the total input value in selected_amount and the total fee being paid in fee_amount
+        let mut selected_amount = 0;
+        let selected = utxos
+            .scan(
```
kafaichoi: this part of code is completely copy-and-paste from LargestFirstCoinSelection. I was trying to extract this logic to a new function `select_from_sorted_utxos` but found out that it's pretty hard to take a Iterator of struct as function argument. So I didn't continue on this one yet.

csralvall:
Concept ACK. I pulled the branch and run the tests, ` test_oldest_first_coin_selection_use_only_necessary` failed. I have to read further, but the approach looks good for me.

src/wallet/coin_selection.rs
kafaichoi: Thank you for catching this. I have fixed it

src/wallet/coin_selection.rs
```diff
@@ -201,45 +202,111 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
...
+    }
+}
 
-        let amount_needed_with_fees = amount_needed + fee_amount;
-        if selected_amount < amount_needed_with_fees {
-            return Err(Error::InsufficientFunds {
-                needed: amount_needed_with_fees,
-                available: selected_amount,
-            });
+/// Another simple and dumb coin selection
```
afilini: The first line is the only one that's shown on docs.rs unless you open the struct doc, so I would try to describe better what it actually does.

(i know you took inspiration from `LargestFirstCoinSelection`, but that one sucks as well, so I would do better here)

src/wallet/coin_selection.rs
```diff
@@ -201,45 +202,111 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
...
+                .into_iter()
+                .map(|utxo| (true, utxo))
+                .chain(optional_utxos.into_iter().map(|utxo| (false, utxo)))
+        };
+
+        select_sorted_utxso(utxos, fee_rate, amount_needed, fee_amount)
+    }
+}
+
+fn select_sorted_utxso<T>(
```
afilini: Typo: utx**so**

src/wallet/coin_selection.rs
```diff
@@ -201,45 +202,111 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
...
+
+        let mut txids: Vec<Txid> = optional_utxos
+            .iter()
+            .map(|wu| wu.utxo.outpoint().txid)
+            .collect::<Vec<Txid>>();
+
+        txids.dedup();
+
+        // query db and create a blockheight lookup table
+        let mut blockheights: HashMap<Txid, u32> = HashMap::new();
```
afilini: I think you could create this map without having to allocate an extra `Vec` for the txids: you could try mapping directly to a `(Txid, u32)` tuple by making the db lookup inside the closure. Then you can `collect()` into an `HashMap<Txid, u32>`.

Since errors could happen during the lookup, wrap everything in a `Result`, something like:

```rust
let blockheights = optional_utxos
    .iter()
    .map(|wu| wu.utxo.outpoint().txid)
    .filter_map(|txid| Ok(database.get_tx(&txid, false)?.and_then(|details| details.confirmation_time).map(|conf| (txid, conf.height))))
    .collect::<Result<HashMap<_, _>, Error>>()?;
```

afilini (commented):
I haven't looked at the tests yet, but overall I think you are on the right track!

I just have a few comments, mostly minor stuff.

src/wallet/coin_selection.rs
kafaichoi: good point. updated description. Please let me know if u find this descriptive enough?

src/wallet/coin_selection.rs
kafaichoi: My original ideal of having a separate txids(and perform dedup) is to avoid duplicated db lookup if there are multiple utxos from same tx. But you are right that there is no need to having allocate an extra Vec for it . We can surely do it in 1 iterator

Your example result in some compilation error. but I am able to get it working in the below code

```
        let blockheights = optional_utxos
            .iter()
            .map(|wu| wu.utxo.outpoint().txid)
            .map(|txid| {
                database
                    .get_tx(&txid, false)
                    .map(|detail| {
                        detail
                        .and_then(|detail| detail.confirmation_time)
                        .map(|conf| (txid, conf.height))
                    })
            })
            .filter_map(|x| x.transpose())
            .collect::<Result<HashMap<_, _>, Error>>()?;
```

However, my latest implementation is using fold as I believed that's the only way that we can skip db lookup in pure iterator approach? Please let me know if u find any issue in the latest approach? Thank you very much for your thorough review

src/wallet/coin_selection.rs
```diff
@@ -201,45 +202,114 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
...
-        Ok(CoinSelectionResult {
-            selected,
-            fee_amount,
-        })
+            required_utxos
+                .into_iter()
+                .map(|utxo| (true, utxo))
+                .chain(optional_utxos.into_iter().map(|utxo| (false, utxo)))
+        };
+
```
LLFourn, line 263: Consider just mapping the utxos to `(utxo, height)` and then sort the tuples by height and map back to utxos. The performance benefit is negligible.

src/wallet/coin_selection.rs
```diff
@@ -201,45 +202,114 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
...
+        select_sorted_utxos(utxos, fee_rate, amount_needed, fee_amount)
+    }
+}
+
+fn select_sorted_utxos<T>(
+    utxos: T,
```
LLFourn: Consider using `impl Iterator<..>` instead of type parameter here.

LLFourn (approved):
ACK f4fe25ae9125aaf5cbb585755b1968b2c3d90c2a

Nice thanks for writing tests. I suggested a small improvements for your consideration.

src/wallet/coin_selection.rs
kafaichoi: Thanks. it looks more concise.(type parameter is definitely redundant here)

src/wallet/coin_selection.rs
kafaichoi: sorry. u meant in line 251(before `sort_unstable_by_key`?) May I ask why this is a bit more performant?(some rust compiler optimization?)

LLFourn: Yeah I'm not sure why this comment showed up here :S

What I suggested is not more performant it's just simpler. My preference would be for the simpler code over performance. Worst case performance is the same. Up to you though.

kafaichoi: ahh I see but personally I think the current one look quite easy to follow. I will not change it for now unless other has stronger opinion?

src/wallet/coin_selection.rs
```diff
@@ -201,45 +202,111 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
...
-        let amount_needed_with_fees = amount_needed + fee_amount;
-        if selected_amount < amount_needed_with_fees {
-            return Err(Error::InsufficientFunds {
-                needed: amount_needed_with_fees,
-                available: selected_amount,
+        // We put the "required UTXOs" first and make sure the optional UTXOs are sorted from
+        // oldest to newest according to blocktime
+        // For utxo that doesn't exist in DB, they will have lowest priority to be selected
+        let utxos = {
+            optional_utxos.sort_unstable_by_key(|wu| {
```
Eunoia1729, line 251: Just a suggestion. How about using timestamp as the secondary key for sorting ?
There are no performance benefits but one advantage is that the sort will be relatively stable and selected utxos will be more predictable.

kafaichoi, line 251: emm but if blockheights is the same then their block timestamps should also be the same?

Eunoia1729: My bad, I didn't notice it was block timestamp. LGTM, kindly ignore.

src/wallet/coin_selection.rs
afilini: I also find this quite readable, I'd keep like this

src/wallet/coin_selection.rs
afilini: Yes, I really like this approach!

afilini (approved):
ACK 6931d0bd1f044bf73c0cb760c0eb0be19aea8de1

afilini:
(sorry it took me so long to re-review, thanks for the great contribution!)

src/wallet/coin_selection.rs
```diff
@@ -298,45 +261,52 @@ impl<D: Database> CoinSelectionAlgorithm<D> for OldestFirstCoinSelection {
...
-        }
+        select_sorted_utxos(utxos, fee_rate, amount_needed, fee_amount)
+    }
+}
 
-        Ok(CoinSelectionResult {
-            selected,
-            fee_amount,
-        })
+fn select_sorted_utxos(
```
murchandamus, line 268: As far as I can tell there are now three coin selection algorithms that order UTXOs in some fashion and then accumulate from the front: Largest First, Oldest First, and Single Random Draw.

I think the new function may replicate this part of SRD: https://github.com/bitcoindevkit/bdk/pull/148/files#diff-fd2a47357bd088bed85419710e86d949d1a172691db87c978be1336a725a42afR491-R502

Perhaps all three should use one the same accumulate function after the UTXO ordering?

murchandamus (commented):
I'm writing about this PR for the Optech newsletter. Do I understand right, that BDK will always use BnB/SRD by default and FIFO would only used on demand? It doesn't run all of the algorithms and then picks from the results, or did I miss that?

src/wallet/coin_selection.rs
afilini: Yes, that's a good point, SRD could probably be refactored to use this function

afilini:
> Do I understand right, that BDK will always use BnB/SRD by default and FIFO would only used on demand?

Yes, the [default algorithm](https://docs.rs/bdk/latest/bdk/wallet/coin_selection/type.DefaultCoinSelectionAlgorithm.html) in non-testing environments is `BranchAndBoundCoinSelection`, which is actually BnB + SRD in one (it tries with BnB and if it fails falls back to SRD).

When running tests it's changed to largest first i believe, to make tests more reproducible.

Users can change the algorithm when building a transaction using [`TxBuilder::coin_selection`](https://docs.rs/bdk/latest/bdk/wallet/tx_builder/struct.TxBuilder.html#method.coin_selection)

Issue #483, opened 2021-11-30
Improving Coinselection in BDK


rajarshimaitra:
**Description**  
  

Coinselection is the scheme to select a set of utxo, from all the available utxos, to create the most optimum spending transaction. Effective coinselection is beneficial for reducing transaction fees as well as improving operation privacy for a wallet. BDK currently uses two coinselection algorithms, [LargestFirst and BranchAndBound](https://docs.rs/bdk/0.13.0/bdk/wallet/coin_selection/index.html).

The aim of the project is to 

- include more coinselection algorithms in the library.
- Add a comparative mechanism to evaluate coinselection results from different algos.
- Use the most “effective” algorithm chosen by a “specific metric”.

**Expected Outcomes**  
 

- General understanding coinselection in Bitcoin protocol.
- Comparative study between different coinselection algorithms.
- Coming up with a metric to compare results of different coinselection algos.

  

**Resources**  
  
  
  
  
  

- What is [Coin Selection](https://bitcoin.stackexchange.com/questions/1077/what-is-the-coin-selection-algorithm)
- Generic Coin Selection [research paper](https://murch.one/wp-content/uploads/2016/11/erhardt2016coinselection.pdf)
- Bitcoin Core [PR Review Club](https://bitcoincore.reviews/22009) on "waste metric".
-  Proposed [waste metric in rust for BDK](https://github.com/bitcoindevkit/bdk/pull/435) to compare different coin selection algorithms.
- BDK Coin selection improvement [idea](https://github.com/bitcoindevkit/bdk/issues/120).
- Coin Selection [Literature](https://murch.one/wp-content/uploads/2016/11/erhardt2016coinselection.pdf).
- Scaling Bitcoin 2016, Milan, Coin Selection [PPT](https://scalingbitcoin.org/milan2016/presentations/D2%20-%204%20-%20Mark%20Erhardt.pdf) 
- Scaling Bitcoin 2016, Milan, Coin Selection [Transcript](https://scalingbitcoin.org/transcript/milan2016/coin-selection)

**Skills Required**  
  
  
  
  

- Experience with git. [Guide](https://git-scm.com/book/en/v2)
- Basic familiarity with rust. [First seven chapters of the book](https://doc.rust-lang.org/book/)
- Familiarity with Bitcoin’s Coin Selection algorithms.
- Familiarity with BDK’s [coinselection module](https://docs.rs/bdk/0.13.0/bdk/wallet/coin_selection/index.html).

[...]

Issue #147, opened 2020-10-27
Move change creation/addition inside `CoinSelectionAlgorithm`


afilini:
See https://github.com/bitcoindevkit/bdk/issues/121#issuecomment-716723002 for why that would be preferred.

We could simply tell the coin selection algorithm how much the change output costs in terms of weight units, and then we would get back a boolean from the coin selection that tells the wallet whether it should actually add the change or not.

danielabrozzoni:
Also, this is needed for correctly implementing the waste metric, as you need to know if you're going to create the change or not in order to calculate the `creation_cost` (see https://bitcoincore.reviews/22009 for a better explanation of the waste metric)

PR #558, opened 2022-03-06
Add waste metric for coin selection

csralvall:
### Description

Waste is a metric introduced by the BnB algorithm as part of its bounding
procedure. Later, it was included as a high level function to use in comparison
of different coin selection algorithms in bitcoin/bitcoin#22009.

This implementations considers waste as the sum of two values:
* Timing cost
* Creation cost
> waste = timing_cost + creation_cost

---

**Timing cost** is the cost associated to the current fee rate and some long
term fee rate used as a treshold to consolidate UTXOs.
> timing_cost = txin_size * current_fee_rate - txin_size * long_term_fee_rate

Timing cost can be negative if the `current_fee_rate` is cheaper than the
`long_term_fee_rate`, or zero if they are equal.

The name of this cost is derived from the action of making bets against some
kind of market (in this case the fee market) to gain profits using the
interpretation of the future behaviour of that market based on a partial
information game.

---

**Creation cost** is the cost associated to the surplus of coins besides the
transaction amount and transaction fees. It can happen in the form of a change
output or in the form of excess fees paid to the miner.

Change cost is derived from the cost of adding the extra output to the
transaction and spending that output in the future.
> change_cost = current_fee_rate * change_output_size + long_term_feerate * change_spend_size

Excess happens when there is not change, and the surplus of coins is spend as
part of the fees to the miner:
> excess = tx_total_value - tx_fees - target

Where _target_ is the desired amount to send.

Creation cost can be zero if there is a perfect match as result of the coin
selection algorithm.

---

So, waste can be zero or negative if the creation cost is zero and the timing
cost is less than or equal to zero

Later this can be used in the wallet to select the most optimal coin selection
algorithm.

This PR continues the work initiated by @benthecarman in #435 and address some
of the features described in #483.

The current PR is based mainly in the following sources:
- [GetSelectionWaste](https://github.com/bitcoin/bitcoin/blob/master/src/wallet/coinselection.h#L201) from Bitcoin Core.
- Draft PR #435.
- Bitcoin Core PR Review Club [#22009](https://bitcoincore.reviews/22009).

### Notes to the reviewers

While running `cargo clippy` some warnings may arise due to the lack of
documentation of the new functions, those will be resolved once the code
changes are finished. The same applies for the `CHANGELOG.md` file.

#### All Submissions:

#### New Features:

### DIFF
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -42,7 +42,7 @@
 //!         fee_rate: FeeRate,
 //!         amount_needed: u64,
 //!         fee_amount: u64,
-//!         cost_of_change: Option<u64>,
+//!         weighted_drain_output: WeightedTxOut,
 //!     ) -> Result<CoinSelectionResult, bdk::Error> {
 //!         let mut selected_amount = 0;
 //!         let mut additional_weight = 0;
@@ -67,13 +67,18 @@
 //!             });
 //!         }
 //!
-//!         let calculated_waste =
-//!             Waste::calculate(&all_utxos_selected, None, amount_needed, fee_rate)?;
+//!         let calculated_waste = Waste::calculate(
+//!             &all_utxos_selected,
+//!             weighted_drain_output,
+//!             amount_needed,
+//!             fee_rate,
+//!         )?;
 //!
 //!         Ok(CoinSelectionResult {
 //!             selected: all_utxos_selected.into_iter().map(|u| u.utxo).collect(),
 //!             fee_amount: fee_amount + additional_fees,
-//!             waste: calculated_waste,
+//!             waste: calculated_waste.0,
+//!             drain_output: calculated_waste.1,
 //!         })
 //!     }
 //! }
@@ -93,9 +98,13 @@
 //! # Ok::<(), bdk::Error>(())
 //! ```
 
-use crate::types::FeeRate;
+use crate::types::{FeeRate, WeightedTxOut};
+use crate::wallet::utils::IsDust;
 use crate::{database::Database, WeightedUtxo};
 use crate::{error::Error, Utxo};
+
+use bitcoin::consensus::encode::serialize;
+use bitcoin::TxOut;
 
 use rand::seq::SliceRandom;
 #[cfg(not(test))]
@@ -125,6 +134,8 @@ pub struct CoinSelectionResult {
     pub fee_amount: u64,
     /// Waste value of current coin selection
     pub waste: Waste,
+    /// Output to drain change amount
+    pub drain_output: Option<TxOut>,
 }
 
 impl CoinSelectionResult {
@@ -147,40 +158,43 @@ impl CoinSelectionResult {
 
 /// Metric introduced to measure the performance of different coin selection algorithms.
 ///
-/// This implementations considers waste as the sum of two values:
+/// This implementation considers "waste" the sum of two values:
 /// * Timing cost
 /// * Creation cost
 /// > waste = timing_cost + creation_cost
 ///
-/// **Timing cost** is the cost associated to the current fee rate and some long term fee rate used
-/// as a treshold to consolidate UTXOs.
+/// **Timing cost** is the cost associated with the current fee rate and some long term fee rate used
+/// as a threshold to consolidate UTXOs.
 /// > timing_cost = txin_size * current_fee_rate - txin_size * long_term_fee_rate
 ///
 /// Timing cost can be negative if the `current_fee_rate` is cheaper than the `long_term_fee_rate`,
 /// or zero if they are equal.
 ///
-/// **Creation cost** is the cost associated to the surplus of coins besides the transaction amount
-/// and transaction fees. It can happen in the form of a change output or in the form of excess
+/// **Creation cost** is the cost associated with the surplus of coins beyond the transaction amount
+/// and transaction fees. It can appear in the form of a change output or in the form of excess
 /// fees paid to the miner.
 ///
 /// Change cost is derived from the cost of adding the extra output to the transaction and spending
 /// that output in the future.
-/// > change_cost = current_fee_rate * change_output_size + long_term_feerate * change_spend_size
+/// > cost_of_change = current_fee_rate * change_output_size + long_term_feerate * change_spend_size
 ///
 /// Excess happens when there is no change, and the surplus of coins is spend as part of the fees
 /// to the miner:
 /// > excess = tx_total_value - tx_fees - target
 ///
-/// Where _target_ is the desired amount to send.
+/// Where _target_ is the amount needed to pay for the fees (minus input fees) and to fulfill the
+/// output values of the transaction.
+/// > target = sum(tx_outputs) + fee(tx_outputs) + fee(fixed_tx_parts)
 ///
 /// Creation cost can be zero if there is a perfect match as result of the coin selection
 /// algorithm.
 ///
 /// So, waste can be zero if creation and timing cost are zero. Or can be negative, if timing cost
-/// is negative and the creation cost is lower enough (less than the absolute value of timing
+/// is negative and the creation cost is low enough (less than the absolute value of timing
 /// cost).
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Waste(pub i64);
+// REVIEW: Add change_output field inside Waste struct?
 
 const LONG_TERM_FEE_RATE: FeeRate = FeeRate::from_sat_per_vb(5.0);
 
@@ -188,15 +202,19 @@ impl Waste {
     /// Calculate the amount of waste for the given coin selection
     ///
     /// - `selected`: the selected transaction inputs
-    /// - `cost_of_change`: the cost of creating change and spending it in the future
-    /// - `target`: the amount in satoshis of the aimed transaction output
+    /// - `weighted_drain_output`: transaction output intended to drain change plus it's spending
+    ///                            satisfaction weight. If script_pubkey belongs to a foreign descriptor, it's
+    ///                            satisfaction weight is zero.
+    /// - `target`: threshold in satoshis used to select UTXOs. It includes the sum of recipient
+    ///             outputs, the fees for creating the recipient outputs and the fees for fixed transaction
+    ///             parts
     /// - `fee_rate`: fee rate to use
     pub fn calculate(
         selected: &[WeightedUtxo],
-        cost_of_change: Option<u64>,
+        weighted_drain_output: WeightedTxOut,
         target: u64,
         fee_rate: FeeRate,
-    ) -> Result<Waste, Error> {
+    ) -> Result<(Waste, Option<TxOut>), Error> {
         // Always consider the cost of spending an input now vs in the future.
         let utxo_groups: Vec<_> = selected
             .iter()
@@ -213,28 +231,44 @@ impl Waste {
             acc + fee - long_term_fee
         });
 
+        // selected_effective_value: selected amount with fee discount for the selected utxos
         let selected_effective_value: i64 = utxo_groups
             .iter()
             .fold(0, |acc, utxo| acc + utxo.effective_value);
 
-        let creation_cost = match cost_of_change {
-            Some(change_cost) => {
-                // Consider the cost of making change and spending it in the future
-                // If we aren't making change, the caller should've set cost_of_change to None
-                if change_cost == 0 {
-                    return Err(Error::Generic(
-                        "if there is not cost_of_change, set to None, not zero".to_string(),
-                    ));
-                }
-                change_cost as i64
-            }
-            None => {
-                // When we are not making change, consider the excess we are throwing away to fees
-                selected_effective_value - target as i64
-            }
+        // target: sum(tx_outputs) + fees(tx_outputs) + fee(fixed_tx_parts)
+        // excess should always be greater or equal to zero
+        let excess = (selected_effective_value - target as i64) as u64;
+
+        let mut drain_output = weighted_drain_output.txout;
+
+        // change output fee
+        let change_output_cost = fee_rate.fee_vb(serialize(&drain_output).len());
+
+        let drain_val = excess.saturating_sub(change_output_cost);
+
+        let mut change_output = None;
+
+        // excess < change_output_size x fee_rate + dust_value
+        let creation_cost = if drain_val.is_dust(&drain_output.script_pubkey) {
+            // throw excess to fees
+            excess
+        } else {
+            // recover excess as change
+            drain_output.value = drain_val;
+
+            let change_input_cost = LONG_TERM_FEE_RATE
+                .fee_wu(TXIN_BASE_WEIGHT + weighted_drain_output.satisfaction_weight);
+
+            let cost_of_change = change_output_cost + change_input_cost;
+
+            // return change output
+            change_output = Some(drain_output);
+
+            cost_of_change
         };
 
-        Ok(Waste(timing_cost + creation_cost))
+        Ok((Waste(timing_cost + creation_cost as i64), change_output))
     }
 }
 
@@ -257,10 +291,9 @@ pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
     /// - `amount_needed`: the amount in satoshi to select
     /// - `fee_amount`: the amount of fees in satoshi already accumulated from adding outputs and
     ///                 the transaction's header
-    /// - `_cost_of_change`: the cost of creating change and spending it in the future
-    ///                      if there is change, it must be a positive number
-    ///                      must be None if there is no change
-    // REVIEW: refactor with less arguments?
+    /// - `weighted_drain_output`: transaction output intended to drain change plus it's spending
+    ///                            satisfaction weight. If script_pubkey belongs to a foreign descriptor, it's
+    ///                            satisfaction weight is zero.
     #[allow(clippy::too_many_arguments)]
     fn coin_select(
         &self,
@@ -270,7 +303,7 @@ pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
         fee_rate: FeeRate,
         amount_needed: u64,
         fee_amount: u64,
-        _cost_of_change: Option<u64>,
+        weighted_drain_output: WeightedTxOut,
     ) -> Result<CoinSelectionResult, Error>;
 }
 
@@ -290,7 +323,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
         fee_rate: FeeRate,
         amount_needed: u64,
         fee_amount: u64,
-        _cost_of_change: Option<u64>,
+        _weighted_drain_output: WeightedTxOut,
     ) -> Result<CoinSelectionResult, Error> {
         log::debug!(
             "amount_needed = `{}`, fee_amount = `{}`, fee_rate = `{:?}`",
@@ -329,7 +362,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for OldestFirstCoinSelection {
         fee_rate: FeeRate,
         amount_needed: u64,
         fee_amount: u64,
-        _cost_of_change: Option<u64>,
+        _weighted_drain_output: WeightedTxOut,
     ) -> Result<CoinSelectionResult, Error> {
         // query db and create a blockheight lookup table
         let blockheights = optional_utxos
@@ -416,6 +449,7 @@ fn select_sorted_utxos(
         selected,
         fee_amount,
         waste: Waste(0),
+        drain_output: None,
     })
 }
 
@@ -476,7 +510,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for BranchAndBoundCoinSelection {
         fee_rate: FeeRate,
         amount_needed: u64,
         fee_amount: u64,
-        _cost_of_change: Option<u64>,
+        _weighted_drain_output: WeightedTxOut,
     ) -> Result<CoinSelectionResult, Error> {
         // Mapping every (UTXO, usize) to an output group
         let required_utxos: Vec<OutputGroup<'_>> = required_utxos
@@ -710,6 +744,7 @@ impl BranchAndBoundCoinSelection {
             selected,
             fee_amount,
             waste: Waste(0),
+            drain_output: None,
         }
     }
 }
@@ -760,6 +795,16 @@ mod test {
             utxo(FEE_AMOUNT as u64 - 40, 1),
             utxo(200_000, 2),
         ]
+    }
+
+    fn get_test_weighted_txout() -> WeightedTxOut {
+        WeightedTxOut {
+            satisfaction_weight: P2WPKH_WITNESS_SIZE,
+            txout: TxOut {
+                value: 0,
+                script_pubkey: Script::new(),
+            },
+        }
     }
 
     fn setup_database_and_get_oldest_first_test_utxos<D: Database>(
@@ -893,6 +938,8 @@ mod test {
         let utxos = get_test_utxos();
         let database = MemoryDatabase::default();
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         let result = LargestFirstCoinSelection::default()
             .coin_select(
                 &database,
@@ -901,7 +948,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1.0),
                 250_000,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
 
@@ -915,6 +962,14 @@ mod test {
         let utxos = get_test_utxos();
         let database = MemoryDatabase::default();
 
+        let weighted_drain_txout = WeightedTxOut {
+            txout: TxOut {
+                value: 0,
+                script_pubkey: Script::new(),
+            },
+            satisfaction_weight: 0,
+        };
+
         let result = LargestFirstCoinSelection::default()
             .coin_select(
                 &database,
@@ -923,7 +978,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1.0),
                 20_000,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
 
@@ -937,6 +992,8 @@ mod test {
         let utxos = get_test_utxos();
         let database = MemoryDatabase::default();
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         let result = LargestFirstCoinSelection::default()
             .coin_select(
                 &database,
@@ -945,7 +1002,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1.0),
                 20_000,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
 
@@ -960,6 +1017,8 @@ mod test {
         let utxos = get_test_utxos();
         let database = MemoryDatabase::default();
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         LargestFirstCoinSelection::default()
             .coin_select(
                 &database,
@@ -968,7 +1027,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1.0),
                 500_000,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
     }
@@ -979,6 +1038,8 @@ mod test {
         let utxos = get_test_utxos();
         let database = MemoryDatabase::default();
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         LargestFirstCoinSelection::default()
             .coin_select(
                 &database,
@@ -987,7 +1048,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1000.0),
                 250_000,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
     }
@@ -997,6 +1058,8 @@ mod test {
         let mut database = MemoryDatabase::default();
         let utxos = setup_database_and_get_oldest_first_test_utxos(&mut database);
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         let result = OldestFirstCoinSelection::default()
             .coin_select(
                 &database,
@@ -1005,7 +1068,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1.0),
                 180_000,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
 
@@ -1054,6 +1117,8 @@ mod test {
         database.set_tx(&utxo1_tx_details).unwrap();
         database.set_tx(&utxo2_tx_details).unwrap();
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         let result = OldestFirstCoinSelection::default()
             .coin_select(
                 &database,
@@ -1062,7 +1127,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1.0),
                 180_000,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
 
@@ -1076,6 +1141,8 @@ mod test {
         let mut database = MemoryDatabase::default();
         let utxos = setup_database_and_get_oldest_first_test_utxos(&mut database);
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         let result = OldestFirstCoinSelection::default()
             .coin_select(
                 &database,
@@ -1084,7 +1151,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1.0),
                 20_000,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
 
@@ -1098,6 +1165,8 @@ mod test {
         let mut database = MemoryDatabase::default();
         let utxos = setup_database_and_get_oldest_first_test_utxos(&mut database);
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         let result = OldestFirstCoinSelection::default()
             .coin_select(
                 &database,
@@ -1106,7 +1175,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1.0),
                 20_000,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
 
@@ -1121,6 +1190,8 @@ mod test {
         let mut database = MemoryDatabase::default();
         let utxos = setup_database_and_get_oldest_first_test_utxos(&mut database);
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         OldestFirstCoinSelection::default()
             .coin_select(
                 &database,
@@ -1129,7 +1200,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1.0),
                 600_000,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
     }
@@ -1143,6 +1214,8 @@ mod test {
         let amount_needed: u64 =
             utxos.iter().map(|wu| wu.utxo.txout().value).sum::<u64>() - (FEE_AMOUNT + 50);
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         OldestFirstCoinSelection::default()
             .coin_select(
                 &database,
@@ -1151,7 +1224,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1000.0),
                 amount_needed,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
     }
@@ -1164,6 +1237,8 @@ mod test {
 
         let database = MemoryDatabase::default();
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         let result = BranchAndBoundCoinSelection::default()
             .coin_select(
                 &database,
@@ -1172,7 +1247,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1.0),
                 250_000,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
 
@@ -1186,6 +1261,8 @@ mod test {
         let utxos = get_test_utxos();
         let database = MemoryDatabase::default();
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         let result = BranchAndBoundCoinSelection::default()
             .coin_select(
                 &database,
@@ -1194,7 +1271,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1.0),
                 20_000,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
 
@@ -1208,6 +1285,8 @@ mod test {
         let utxos = get_test_utxos();
         let database = MemoryDatabase::default();
 
+        let weighted_drain_txout = get_test_weighted_txout();
+
         let result = BranchAndBoundCoinSelection::default()
             .coin_select(
                 &database,
@@ -1216,7 +1295,7 @@ mod test {
                 FeeRate::from_sat_per_vb(1.0),
                 299756,
                 FEE_AMOUNT,
-                None,
+                weighted_drain_txout,
             )
             .unwrap();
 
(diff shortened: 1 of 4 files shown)


Eunoia1729 (commented):
utACK

danielabrozzoni:
Tests are failing, can you please update the PR? :)
https://github.com/bitcoindevkit/bdk/pull/515 introduced the `is_spent` field in `LocalUtxo`, since you're testing coin selection (which uses only unspent utxos) you probably want to set it to false all the time :)

csralvall:
Yes, you are right! I will try to rebase my branch in top of master to get all last changes (including #515) and then apply the correction. Let me know if you have any doubt about this.

csralvall:
It seems that the only failing test remaining are related to missing documentation. I don't know if you want to discuss the implementation further, but I'm willing to add documentation to pass the test and then iterate again, in case something is missing.

danielabrozzoni:
I'll look into the code ASAP, sorry for the waiting :)

src/wallet/coin_selection.rs
```diff
@@ -110,16 +127,28 @@ pub type DefaultCoinSelectionAlgorithm = LargestFirstCoinSelection; // make the
...
+pub type Waste = i64;
 /// Result of a successful coin selection
 #[derive(Debug)]
 pub struct CoinSelectionResult {
     /// List of outputs selected for use as inputs
     pub selected: Vec<Utxo>,
     /// Total fee amount in satoshi
     pub fee_amount: u64,
+    /// Waste value of current coin selection
+    waste: Cell<Option<Waste>>,
```
danielabrozzoni: I'm not sure why you're using a `Cell` here, is it because you want `get_waste` to take `&self` instead of `&mut self`? If so, why is that preferred?

src/wallet/coin_selection.rs
```diff
@@ -135,6 +164,78 @@ impl CoinSelectionResult {
...
+            Some(waste) => waste,
+            None => {
+                let calculated_waste = calculate_waste(
+                    selected,
+                    cost_of_change,
+                    target,
+                    fee_rate,
+                    long_term_fee_rate,
+                )
+                .unwrap();
```
danielabrozzoni: When you're working on a library, I think you should either:
- If you're sure this will never panic (i.e. you're absolutely sure the function you're calling will never error), use `.expect("Reason why this will never panic")` (I'm not sure if we do that frequently in the code tbh, but it makes reviewing the code super easy)
- If you don't know if the function you're calling will panic or not (like, in this case), just propagate the error with `?`

src/wallet/coin_selection.rs
```diff
@@ -66,10 +67,25 @@
...
-//!         Ok(CoinSelectionResult {
-//!             selected: all_utxos_selected,
-//!             fee_amount: fee_amount + additional_fees,
-//!         })
+//!         let calculated_waste = Some(
+//!             calculate_waste(
+//!                 required_utxos
+//!                     .into_iter()
+//!                     .chain(optional_utxos)
+//!                     .collect::<Vec<WeightedUtxo>>(),
```
danielabrozzoni: This should be the selected utxos, not required + optional

src/wallet/coin_selection.rs
```diff
@@ -235,10 +336,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
             });
         }
 
-        Ok(CoinSelectionResult {
-            selected,
-            fee_amount,
-        })
+        Ok(CoinSelectionResult::new(selected, fee_amount, None))
```
danielabrozzoni: This waste should be calculated eventually, at least add a TODO

src/wallet/coin_selection.rs
```diff
@@ -135,6 +164,78 @@ impl CoinSelectionResult {
...
+        }
+    }
+}
+
+pub fn calculate_waste(
+    selected: Vec<WeightedUtxo>,
+    cost_of_change: Option<u64>,
+    target: u64,
+    fee_rate: FeeRate,
+    long_term_fee_rate: FeeRate,
```
danielabrozzoni: Since Bitcoin Core is hardcoding the long_term_fee_rate to 10, I'd say we do the same with a comment

src/wallet/coin_selection.rs
```diff
@@ -135,6 +164,78 @@ impl CoinSelectionResult {
...
+                .unwrap();
+                self.waste.set(Some(calculated_waste));
+                calculated_waste
+            }
+        }
+    }
+}
+
+pub fn calculate_waste(
+    selected: Vec<WeightedUtxo>,
```
danielabrozzoni: We don't need to take ownership of the vector, a simple reference is enough: `&[WeightedUtxo]`

src/wallet/coin_selection.rs
```diff
@@ -135,6 +164,78 @@ impl CoinSelectionResult {
...
+    cost_of_change: Option<u64>,
```
danielabrozzoni: After speaking with @afilini we concluded that inside the `coin_select` we don't really know the cost of change, so it should be added as a parameter to the `coin_select` method

danielabrozzoni (changes requested):
Concept ACK, the approach could be improved:
- `get_waste` should be dropped altogether, it's not really useful to cache the Waste (you just want to calculate it once at the end of the coin selection)
- instead of having a `pub fn calculate_waste`, better to go with: 
```
struct Waste {
    value: i64,
}

impl Waste {
    pub fn calculate (...) -> Result<Waste, Error> {...the whole calcualte_waste function is pasted here}
}
```

This is just a first look at it, I still haven't tested

csralvall:
Thanks for the review! Let me explain myself. I made a few assumptioms before
started working in this PR:

1. Originally I followed [this suggestion](https://github.com/bitcoindevkit/bdk/pull/435#discussion_r703057206)
   from Steve to drop the struct WasteMetric in favor of a helper function
   associated with the `CoinSelectionAlgorithm`. I tried this approach but the
   generic database bound imposed some constraints to the signature of the
   `calculate_waste` function that didn't make sense to me at that moment. So I
   decided to implement it as a general function related to the
   `CoinSelectionResult` through the `get_waste` method.

   I will try to use your proposed approach and create the `struct Waste`.

src/wallet/coin_selection.rs
csralvall: That's not currently possible. `calculate_waste` creates `OutputGroup`
from `Vec<WeightedUtxo>` to get the `fee` and `effective_value` of each
`WeightedUtxo`.
The `OutputGroup` constructor needs to take ownership of the
vector, like the `calculate_waste` function.
I can change `selected` by `utxo_groups: &[OutputGroup]`, but that
requires `OutputGroup` to be public.

src/wallet/coin_selection.rs
csralvall: Yes, that's conceptually correct but infesable with the actual method's
signature. Currently is not possible to consume `Vec<Utxo>` for
`calculate_waste`, just `Vec<WeightedUtxo>`. That might change also, if we
change the `selected` parameter by something else.
I suggest to finish the `calculate_waste` function signature first and then
apply this kind of changes. It may require some modifications to the
`AlwaysSpendEverything` algorithm outside of the scope modified by
`calculate_waste`, are you OK with that?

src/wallet/coin_selection.rs
csralvall: The use of `get_waste` and the `waste: Cell<<Option<Waste>>` field it's
because I was assuming a worst case scenario where the algorithms doesn't
necessarily set the waste metric and when an external source requires the
metric they should have a way to fulfill the request.
To modify the field, as it was, Rust required the whole instance to be
mutable, something undesirable, as the selected coins shouldn't be modified
after the algorithm execution. So I decided to use Cell as a way to mimic
the private/public behavior that I know from other programming languages.

Taking the above and what you said in the comments: _"you just want to
calculate it once at the end of the coin selection"_. I think that it should
be correct to set the waste field of `CoinSelectionResult` as: `pub waste:
Waste`
and, following your suggestion, remove the `get_waste` method altogether.

src/wallet/coin_selection.rs
csralvall: Yeah, thanks for pointing this out, I had a conceptual error here.
As `get_waste` will be removed, I will ignore it for now.

src/wallet/coin_selection.rs
danielabrozzoni: Oh, you're right!

> I can change selected by utxo_groups: &[OutputGroup], but that
requires OutputGroup to be public.

Mh, that's a bit suboptimal, it's ok to calculate the OutputGroup inside
the function IMHO. Let's just leave it like that :)

danielabrozzoni: ...Actually, playing a bit more with the `AlwaysSpendEverything` example, having a `Vec` is really suboptimal, and makes the whole experience quite painful 😂  I'll try to figure out something!

csralvall:
Some things that I've experienced while trying to make `Waste` a struct:
1. If it will be exposed through `CoinSelectionResult` it should be a public struct.
2. If we don't want `pub struct Waste` we can use `struct Waste` internally and pass the inner value to the `CoinSelectionResult` struct.
3. If we proceed with the public struct, an extra de-reference from the `waste` field of `CoinSelectionResult` will be required (`waste.value`).

All options seem cumbersome IMHO.

I reviewed the `coinselection` module from `bitcoin` and they have structured the `Waste` calculus in three different functions:
- `GetSelectionWaste` who does all the work and is not associated with any struct (for testing purposes).
- `ComputAndSetWaste` a setter (something similar to `get_waste`).
- `GetWaste` a getter.

Although all of that it's still a work in progress, because the values of the waste metric are hardcoded for `SelectCoinsBnB`.

afilini:
> If it will be exposed through CoinSelectionResult it should be a public struct.

Yes, because custom coin selection algorithms should also be able to calculate waste

> If we proceed with the public struct, an extra de-reference from the waste field of CoinSelectionResult will be required 

I don't really understand this, if you want you could also implement `Deref` so that `Waste` can behave like an `i64`. But I don't think it's that bad to just read the inner value. Also, if you only have one field you can also definite it as `struct Waste(pub i64)` and add a getter for the inner value so that you don't have to use `waste.0`.

I haven't reviewed the whole code yet, but I did spend some time trying to remove some useless cloning in the code: I pushed my code [here](https://github.com/afilini/bdk/commits/pr/558/fix-cloning), feel free to take the two top commits and apply them to your PR.

src/wallet/coin_selection.rs
csralvall: I've added `cost_of_change` as a parameter of coin_select.
`create_tx` in `src/wallet/mod.rs` may require further changes.

src/wallet/coin_selection.rs
csralvall: @afilini adjusted OutputGroup in [ef1c7be](https://github.com/bitcoindevkit/bdk/pull/558/commits/ef1c7be9ac5f8a35fc358a906ff2faa5c0fcd9e4), to make `selected: &[WeightedUtxo]` possible.

src/wallet/coin_selection.rs
csralvall: Fixed with [d2bf84a](https://github.com/bitcoindevkit/bdk/pull/558/commits/d2bf84af45283e9b896788b74383c805f110cf4b).

csralvall:
> I don't really understand this, if you want you could also implement `Deref` so that `Waste` can behave like an `i64`. But I don't think it's that bad to just read the inner value. Also, if you only have one field you can also definite it as `struct Waste(pub i64)` and add a getter for the inner value so that you don't have to use `waste.0`.

I didn't know about `Deref`, but now that I followed your suggestion I find `struct Waste(pub i64)` cleaner.
That's was all about, I prefer to have straight semantics in the code.
I don't know if there is a case of use for the getter if we can use destructuring assignment, but I can implement one if you think it's necessary.

> I haven't reviewed the whole code yet, but I did spend some time trying to remove some useless cloning in the code: I pushed my code [here](https://github.com/afilini/bdk/commits/pr/558/fix-cloning), feel free to take the two top commits and apply them to your PR.

Thanks for this! There are things that I'm still learning about Rust and made some clunky decisions.

src/wallet/coin_selection.rs
```diff
@@ -42,18 +42,19 @@
...
 //!         amount_needed: u64,
 //!         fee_amount: u64,
+//!         cost_of_change: Option<u64>,
 //!     ) -> Result<CoinSelectionResult, bdk::Error> {
 //!         let mut selected_amount = 0;
 //!         let mut additional_weight = 0;
 //!         let all_utxos_selected = required_utxos
 //!             .into_iter()
-//!             .chain(optional_utxos)
+//!             .chain(optional_utxos.into_iter())
```
danielabrozzoni: No need to have `into_iter` :)

src/wallet/coin_selection.rs
```diff
@@ -138,6 +159,59 @@ impl CoinSelectionResult {
...
+pub struct Waste(pub i64);
+
+impl Waste {
+    pub fn calculate(
+        selected: &[WeightedUtxo],
+        cost_of_change: Option<u64>,
+        target: u64,
+        fee_rate: FeeRate,
+        // NOTE: The value of long_term_fee_rate should be set to 10sat/vb.
+        long_term_fee_rate: FeeRate,
```
danielabrozzoni: No need for having it as a parameter, it can just be a constant :) 

```diff
+const LONG_TERM_FEE_RATE: FeeRate = FeeRate::from_sat_per_vb(10.0);
 
 /// Result of a successful coin selection
 #[derive(Debug)]
@@ -168,8 +169,6 @@ impl Waste {
         cost_of_change: Option<u64>,
         target: u64,
         fee_rate: FeeRate,
-        // NOTE: The value of long_term_fee_rate should be set to 10sat/vb.
-        long_term_fee_rate: FeeRate,
     ) -> Result<Waste, Error> {
         // Always consider the cost of spending an input now vs in the future.
         let utxo_groups: Vec<_> = selected
@@ -180,7 +179,7 @@ impl Waste {
         // If fee_rate < long_term_fee_rate timing cost can be negative
         let timing_cost: i64 = utxo_groups.iter().fold(0, |acc, utxo| {
             let fee: i64 = utxo.fee as i64;
-            let long_term_fee: i64 = long_term_fee_rate
+            let long_term_fee: i64 = LONG_TERM_FEE_RATE
                 .fee_wu(TXIN_BASE_WEIGHT + utxo.weighted_utxo.satisfaction_weight)
                 as i64;
```

src/wallet/coin_selection.rs
```diff
@@ -366,16 +441,17 @@ impl<D: Database> CoinSelectionAlgorithm<D> for BranchAndBoundCoinSelection {
         fee_rate: FeeRate,
         amount_needed: u64,
         fee_amount: u64,
+        cost_of_change: Option<u64>,
```
danielabrozzoni: Let's call it `_cost_of_change` for now, so clippy doesn't complain

src/wallet/coin_selection.rs
```diff
@@ -118,9 +128,20 @@ pub struct CoinSelectionResult {
...
 impl CoinSelectionResult {
+    /// Create new CoinSelectionResult
+    pub fn new(selected_utxos: Vec<Utxo>, fee_amount: u64, selection_waste: Waste) -> Self {
+        CoinSelectionResult {
+            selected: selected_utxos,
+            fee_amount,
+            waste: selection_waste,
+        }
+    }
+
```
danielabrozzoni: Not really needed, we can just create directly `CoinSelectionResult` (as we were doing before)

danielabrozzoni (commented):
This is looking really good! 😄
A couple of comments, small stuff. I still need to review tests 😭  Sorry!

Also, can you please squash commits? I'd say you just squash in one commit, putting Alekos and Ben as co-authors.

I don't know if it's better to merge with all the `TODOs` around (for calculating the waste and the cost of change) or not. @afilini wdyt?

csralvall:
Do you think that this is ready to start documenting it?

notmandatory:

danielabrozzoni:
Once it compiles, I think it's ready 😄

csralvall:
I added some `REVIEW` comments in `mod.rs` and `coin_selection.rs` because I need some clarification of what to do with those sections of code.

src/wallet/coin_selection.rs
```diff
@@ -138,6 +145,98 @@ impl CoinSelectionResult {
...
+///
+/// **Creation cost** is the cost associated to the surplus of coins besides the transaction amount
+/// and transaction fees. It can happen in the form of a change output or in the form of excess
+/// fees paid to the miner.
+///
+/// Change cost is derived from the cost of adding the extra output to the transaction and spending
+/// that output in the future.
+/// > change_cost = current_fee_rate * change_output_size + long_term_feerate * change_spend_size
+///
+/// Excess happens when there is not change, and the surplus of coins is spend as part of the fees
```
murchandamus:
```suggestion
/// Excess happens when there is no change, and the surplus of coins is spend as part of the fees
```

src/wallet/coin_selection.rs
```diff
@@ -138,6 +145,98 @@ impl CoinSelectionResult {
...
+/// to the miner:
+/// > excess = tx_total_value - tx_fees - target
+///
+/// Where _target_ is the desired amount to send.
+///
+/// Creation cost can be zero if there is a perfect match as result of the coin selection
+/// algorithm.
+///
+/// So, waste can be zero or negative if the creation cost is zero and the timing
+/// cost is less than or equal to zero.
```
murchandamus: Nit: The waste score can also end up being negative if the creation cost is simply low rather than zero.

```suggestion
/// So, waste can be zero or negative if the creation cost is low or zero and the timing
/// cost is less than or equal to zero.
```

src/wallet/coin_selection.rs
```diff
@@ -138,6 +145,98 @@ impl CoinSelectionResult {
...
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Waste(pub i64);
+
+const LONG_TERM_FEE_RATE: FeeRate = FeeRate::from_sat_per_vb(10.0);
```
murchandamus: While I picked the original 10 s/vB for long-term-feerate-estimate (LTFRE) per gut feeling and it has been working out decently, the blockspace market in the last year has allowed us to send at lower feerates than that value for large parts of the last 11 months.

One concern that I've been looking into recently is whether encouraging a consolidatory behavior for up to 10× the `minRelayTxFeerate` isn't causing wallets to spend more on fees than necessary. I've been running some coin selection [experiments](https://github.com/Xekyo/coin-selection-simulation/commit/80f615c450c5f774b5abe31015636431b5ac378f) based on feerates of the last year, and my tentative result is that a lower LTFRE would currently result in a lower total cost for the wallet. In the specific scenario that I was experimenting with, the best scoring LTFRE was 3 s/vB, but I would suggest setting 5 s/vB at this time, under the assumption that the progressing adoption of native segwit and Lightning will likely cause the blockspace market to remain relaxed for some time still, but there might be a few blockspace market demand events down the road if we get a bullrun.

Optimally, the LTFRE would be a dynamic value (H/T @josibake), but how to do that is on-going research. If the blockspace market generally becomes more competitive, the LTFRE should probably eventually be increased again, but I think the lower value might be good for at least another year?

```suggestion
const LONG_TERM_FEE_RATE: FeeRate = FeeRate::from_sat_per_vb(5.0);
```

src/wallet/coin_selection.rs
```diff
@@ -138,6 +145,98 @@ impl CoinSelectionResult {
...
+
+impl Waste {
+    /// Calculate the amount of waste for the given coin selection
+    ///
+    /// - `selected`: the selected transaction inputs
+    /// - `cost_of_change`: the cost of creating change and spending it in the future
+    /// - `target`: the amount in satoshis of the aimed transaction output
```
murchandamus: Possible documentation nit: I would expect the selection target to include the sum of the recipient amounts, the fees for creating the recipient outputs, and the fees for the (in practice) fixed transaction parts (version, locktime, input counter, output counter,  marker, flag). That way, only the inputs, the change, and additional fees from the inputs and `creation_cost` are up to be decided at the point of coin selection. It might be implemented differently here, but saying that `target` is just the recipient amounts might be slightly inaccurate.

src/wallet/coin_selection.rs
```diff
@@ -138,6 +145,98 @@ impl CoinSelectionResult {
...
+
+            acc + fee - long_term_fee
+        });
+
+        let selected_effective_value: i64 = utxo_groups
+            .iter()
+            .fold(0, |acc, utxo| acc + utxo.effective_value);
+
+        let creation_cost = match cost_of_change {
+            Some(change_cost) => {
```
murchandamus: You could perhaps detect whether the remainder after paying for the fixed transaction parts is sufficient to create a change, rather than making the caller responsible for knowing whether there will be change. A possible lower bound of satoshis in the remainder for toggling change creation could be `change_output_size×feerate + dust_threshold`.

src/wallet/coin_selection.rs
```diff
@@ -1339,4 +1495,140 @@ mod test {
...
+        let selected = generate_utxos_of_values(vec![100_000_000, 200_000_000]);
+        let amount_needed = 200_000_000;
+
+        let utxos: Vec<OutputGroup<'_>> = selected
+            .iter()
+            .map(|u| OutputGroup::new(u, fee_rate))
+            .collect();
+
+        let size_of_change = 31;
+        let cost_of_change: u64 = (size_of_change as f32 * fee_rate.as_sat_vb()) as u64;
```
murchandamus: This may just be my limited proficiency in reading Rust, but above you described that `cost_of_change` should be composed of the cost of creating a change output and the cost of spending it later. This seems to only be composed of the cost of creating the change output, but I don't see the cost of spending it.

src/wallet/coin_selection.rs
```diff
@@ -1339,4 +1495,140 @@ mod test {
...
+
+        let utxo_fee_diff: i64 = utxos.iter().fold(0, |acc, utxo| {
+            let fee_rate: i64 = utxo.fee as i64;
```
murchandamus: I'm confused at this variable name. Shouldn't this be a `fee` rather than a `fee_rate`? (Also in the following test cases.)

src/wallet/coin_selection.rs
```diff
@@ -1339,4 +1495,140 @@ mod test {
...
+
+        // Waste without change is the excess and difference between fee and long term fee
+        let waste = Waste::calculate(&selected, None, amount_needed, fee_rate);
+
+        assert_eq!(waste.unwrap(), Waste(utxo_fee_diff + excess as i64));
+    }
+
+    #[test]
+    fn test_calculate_waste_with_no_timing_cost_and_no_creation_cost() {
+        let fee_rate = FeeRate::from_sat_per_vb(10.0);
```
murchandamus: This might be a bit more obvious:

```suggestion
        let fee_rate = FeeRate::from_sat_per_vb(LONG_TERM_FEE_RATE);
```

murchandamus (commented):
Cool that you're working on this. I really like your description of the waste metric as the composite of a "timing cost" plus a "creation cost". I have a few questions and comments, I hope not too many are just misunderstandings owed to my low Rust reading proficiency.

src/wallet/coin_selection.rs
csralvall: Sure! I will try to reproduce those experiments locally. You get them running
`bitcoin/test/functional/sim_coinselect.py`, right?

Also, Is there any source to follow josibake's research?

src/wallet/coin_selection.rs
csralvall: What do you think of this?

> So, waste can be zero if creation and timing cost are zero. Or can be
> negative, if timing cost is negative and the creation cost is lower enough
> (less than the absolute value of timing cost).

src/wallet/coin_selection.rs
csralvall: Not a nit at all, it's an important remark! I learn a few thing while reviewing
this, thanks!

It can be changed to:

> target: threshold in satoshis used to select UTXOs. It includes the sum of
> recipient outputs, the fees for creating the recipient outputs and the fees
> for fixed transaction parts

WDYT?

src/wallet/coin_selection.rs
csralvall: You are right.

src/wallet/coin_selection.rs
csralvall: I think that it's desirable to do it, but it breaks the meaning of metric.
IMHO, this metric should only measure and provide a single outcome through
its result.
The intention is to use it as a tie breaker between the different
coin selection algorithms.
So, it will be computed after the coin selection algorithm. At that moment,
they should already have selected the UTXOs and decide whether there will be
change or not.
I think that we can reconsider this after it has been integrated with the
CoinSelectionAlgorithms.

src/wallet/coin_selection.rs
csralvall: So, using the [size-calulator](https://bitcoinops.org/en/tools/calc-size/) from optech, that same UTXO would weight 68 vbytes as a transaction input.
Am I missing something?

```rust
let change_out_size = 31;
let change_in_size = 68;
let cost_of_change: u64 = (
  change_out_size as f32 * fee_rate.as_sat_vb() - change_in_size as f32 * LONG_TERM_FEE_RATE.as_sat_vb()
) as u64;
```

src/wallet/coin_selection.rs
murchandamus: Yeah, but `bitcoin/test/functional/sim_coinselect.py` is not in the master branch yet. You can find it in @achow101's branch of Bitcoin Core here: https://github.com/achow101/bitcoin/tree/coin-sel-sim-trace
You will also need to enable tracing ([User-space, Statically Defined Tracing](https://github.com/bitcoin/bitcoin/blob/master/contrib/tracing/README.md)) to be able to collect the logs.

I don't think there is a public document on Josie's research yet.

src/wallet/coin_selection.rs
murchandamus: SGTM

src/wallet/coin_selection.rs
murchandamus: Right, if you use the heuristic to compare transaction candidates after coin selection, you'll know already whether a change is created. The automatic detection could be useful if you use it during coin selection to evaluate candidates of an algorithm that can optionally create change.

src/wallet/coin_selection.rs
```diff
@@ -1507,16 +1508,19 @@ mod test {
             .map(|u| OutputGroup::new(u, fee_rate))
             .collect();
 
-        let size_of_change = 31;
-        let cost_of_change: u64 = (size_of_change as f32 * fee_rate.as_sat_vb()) as u64;
+        let change_out_size = 31_f32;
+        let change_in_size = 68_f32;
+        let cost_of_change: u64 = (change_out_size * fee_rate.as_sat_vb()
+            - change_in_size * LONG_TERM_FEE_RATE.as_sat_vb())
```
murchandamus: I think the sign here is wrong. The two costs should add (unless one of them is tracked as negative for some reason).

 ```suggestion
             + change_in_size * LONG_TERM_FEE_RATE.as_sat_vb())
```

File: src/wallet/coin_selection.rs [hunk 21]
```diff
@@ -176,12 +176,13 @@ impl CoinSelectionResult {
 /// Creation cost can be zero if there is a perfect match as result of the coin selection
 /// algorithm.
 ///
-/// So, waste can be zero or negative if the creation cost is zero and the timing
-/// cost is less than or equal to zero.
+/// So, waste can be zero if creation and timing cost are zero. Or can be negative, if timing cost
+/// is negative and the creation cost is lower enough (less than the absolute value of timing
```
murchandamus:
```suggestion
/// is negative and the creation cost is low enough (less than the absolute value of timing
```

src/wallet/coin_selection.rs
```diff
@@ -138,6 +145,98 @@ impl CoinSelectionResult {
     }
 }
 
+/// Metric introduced to measure the performance of different coin selection algorithms.
+///
+/// This implementations considers waste as the sum of two values:
```
murchandamus:
```suggestion
/// This implementation considers "waste" the sum of two values:
```

src/wallet/coin_selection.rs
```diff
@@ -138,6 +145,98 @@ impl CoinSelectionResult {
...
+/// * Timing cost
+/// * Creation cost
+/// > waste = timing_cost + creation_cost
+///
+/// **Timing cost** is the cost associated to the current fee rate and some long term fee rate used
+/// as a treshold to consolidate UTXOs.
+/// > timing_cost = txin_size * current_fee_rate - txin_size * long_term_fee_rate
```
murchandamus:
```suggestion
/// **Timing cost** is the cost associated with the current fee rate and some long term fee rate used
/// as a threshold to consolidate UTXOs.
/// > timing_cost = txin_size * current_fee_rate - txin_size * long_term_fee_rate
```

src/wallet/coin_selection.rs
```diff
@@ -138,6 +145,98 @@ impl CoinSelectionResult {
...
+///
+/// Timing cost can be negative if the `current_fee_rate` is cheaper than the `long_term_fee_rate`,
+/// or zero if they are equal.
+///
+/// **Creation cost** is the cost associated to the surplus of coins besides the transaction amount
+/// and transaction fees. It can happen in the form of a change output or in the form of excess
+/// fees paid to the miner.
```
murchandamus:
```suggestion
/// **Creation cost** is the cost associated with the surplus of coins beyond the transaction amount
/// and transaction fees. It can appear in the form of a change output or in the form of excess
/// fees paid to the miner.
```

murchandamus (commented):
Read it over quickly again. Good changes!
I have couple more optional nits which you can feel free to ignore, but I think there may have slipped a sign error into the tests.

murchandamus:
I don't feel like I'm familiar enough to "ACK" here, but my prior comments appear to have been all addressed. So, I guess "Concept ACK"? :)

src/wallet/coin_selection.rs
```diff
@@ -157,6 +260,11 @@ pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
     /// - `amount_needed`: the amount in satoshi to select
     /// - `fee_amount`: the amount of fees in satoshi already accumulated from adding outputs and
     ///                 the transaction's header
+    /// - `_cost_of_change`: the cost of creating change and spending it in the future
+    ///                      if there is change, it must be a positive number
+    ///                      must be None if there is no change
+    // REVIEW: refactor with less arguments?
```
danielabrozzoni: I think it's fine to leave it as it is for now :)

src/wallet/coin_selection.rs
```diff
@@ -157,6 +260,11 @@ pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
     /// - `amount_needed`: the amount in satoshi to select
     /// - `fee_amount`: the amount of fees in satoshi already accumulated from adding outputs and
     ///                 the transaction's header
+    /// - `_cost_of_change`: the cost of creating change and spending it in the future
+    ///                      if there is change, it must be a positive number
+    ///                      must be None if there is no change
```
danielabrozzoni: The problem here is that you don't know if there's going to be a change or not before calling the `coin_selection` method...

In fact, at the moment we don't even calculate if we're going to have the change in `coin_select`! We do it later: https://github.com/bitcoindevkit/bdk/blob/fbd98b4c5a88a10f8dfc22b4c54187eba4d68afe/src/wallet/mod.rs#L768-L781

We obviously can't calculate the waste properly before knowing if there's a change or not, so we should move the change creation inside the coin_select method. See https://github.com/bitcoindevkit/bdk/issues/147

danielabrozzoni: To clarify: here, we should just take a `u64` (we can't know if we're going to have a change or not before the `coin_select`).
When we calculate the waste we should decide between the cost_of_change value or None based on whether we have the change or not

src/wallet/coin_selection.rs
```diff
@@ -138,6 +145,102 @@ impl CoinSelectionResult {
...
+    /// - `selected`: the selected transaction inputs
+    /// - `cost_of_change`: the cost of creating change and spending it in the future
+    /// - `target`: threshold in satoshis used to select UTXOs. It includes the sum of recipient
+    ///             outputs, the fees for creating the recipient outputs and the fees for fixed transaction
+    ///             parts
+
+    /// - `fee_rate`: fee rate to use
+    pub fn calculate(
+        selected: &[WeightedUtxo],
+        cost_of_change: Option<u64>,
```
danielabrozzoni: Really picky nit: I'd stay consistent between `change_cost` and `cost_of_change`, as far as I understand they're the same thing, so we should either use one name or the other in the code (whichever you find more clear)

src/wallet/coin_selection.rs
```diff
@@ -138,6 +145,102 @@ impl CoinSelectionResult {
...
+/// > excess = tx_total_value - tx_fees - target
+///
+/// Where _target_ is the desired amount to send.
+///
+/// Creation cost can be zero if there is a perfect match as result of the coin selection
+/// algorithm.
+///
+/// So, waste can be zero if creation and timing cost are zero. Or can be negative, if timing cost
+/// is negative and the creation cost is low enough (less than the absolute value of timing
+/// cost).
```
danielabrozzoni: This doc is great btw!! Good job :)

src/wallet/coin_selection.rs
danielabrozzoni: Btw, probably the cost of change should be calculated inside the `Waste::calculate` and not outside. 

> change_cost = current_fee_rate * change_output_size + long_term_feerate * change_spend_size

This means we need the following:
`current_fee_rate` -> we have it already
`change_output_size` -> we need it to have it as a param
`long_term_feerate` -> we have it, it's `LONG_TERM_FEE_RATE`
`change_spend_size` -> we need it to have it as a param

Which means we should either add two new parameters to the function (maybe it's better to merge those in a struct)

danielabrozzoni: As of exactly how to calculate the `change_output_size` and the `change_spend_size`, I'm figuring it out, and I'll get back to you

src/wallet/mod.rs
```diff
@@ -749,13 +749,22 @@ where
             params.bumping_fee.is_some(), // we mandate confirmed transactions if we're bumping the fee
         )?;
 
+        let weighted_drain_txout = WeightedTxOut {
```
danielabrozzoni, line 752: This needs to be updated? Do you need help figuring out how to do it? :)

csralvall: No, or yes :) Changes with this may collide with the solution for #147 . I don't know if I should address that in this same PR or not.

csralvall:
I realized that I should return the fee amount related to the change output
together with the change `TxOut`.
I could implement an `enum` like:

```rust
pub enum Excess {
    NoChange(Error::InsufficientFunds),
    // Change(change_output, change_output_fee)
    Change(TxOut, u64),
}
```

Then the signature of `Waste::calculate` would be:

```rust
    pub fn calculate(
        selected: &[WeightedUtxo],
        weighted_drain_output: WeightedTxOut,
        target: u64,
        fee_rate: FeeRate,
    ) -> Result<(Waste, Excess), Error>
```

And we can decide later if we join the change output with the other outputs
inside `coin_select` or `create_tx`. Either way, if we keep the change above,
`CoinSelectionResult` would look like this:

```rust
pub struct CoinSelectionResult {
    /// List of outputs selected for use as inputs
    pub selected: Vec<Utxo>,
    /// Total fee amount in satoshi
    pub fee_amount: u64,
    /// Waste value of current coin selection
    pub waste: Waste,
    /// Remaining amount after coin_select
    pub excess: Excess,
}
```

Through the use of `NoChange(err)` [here](https://github.com/bitcoindevkit/bdk/blob/098c448dfbdf1922d9911760da27d76df67af420/src/wallet/mod.rs#L801)
we can avoid passing `params.drain_wallet` as parameter to
`coin_select`.
And with `Change(output, output_fee)` we can keep the
same functionality of [this line](https://github.com/bitcoindevkit/bdk/blob/688ff96c8e925adcd8bf689388e95e22bc6c99ff/src/wallet/mod.rs#L810)
(that wouldn't be removed with the fix for #147 )
through `output.script_pubkey` and `output.value`

src/wallet/mod.rs
danielabrozzoni: I think you should fork from bdk master and try to implement #147 first (you can of course go back and look at this code if you feel like there are problems that you solved here already). In this way you'll come up with a solution that doesn't depend on the waste metric.

murchandamus:
Also, returning the total resulting fee with the coin selection result will make it easier to handle spending unconfirmed inputs when the parent transaction has a lower feerate than what you intend for the new transaction you're building to have. I mean, if you e.g. want to automatically bump the parent transaction to the same feerate so you achieve the intended feerate, and if you want to account for that additional in the waste metric later.

csralvall:
I'm not sure that I've understood what you said correctly. You were thinking in
something like the following situation?
- `Tx_A` ----> UTXO_A (fee rate = 5, confirmed)
- `Tx_B` ----> UTXO_B(fee rate = 3, unconfirmed)
- (UTXO_A, UTXO_B) ----> `Tx_C`(fee rate = 5).

The original transaction that generated UTXO_B is `Tx_B`, and coin select
re-bumps the fee for it (not implemented currently), lets call it `Tx_B_bis`.
You said that the extra amount that comes from [`Tx_B.fee_amount` - `Tx_B_bis.fee_amount`]
can now be considered as part of the fee amount of `Tx_C` and waste can account it?

I'm unsure because `fee_amount` isn't new for `CoinSelectionResult`.

As a side note, I'm planning to propose a modification for the fee amount
returned by `CoinSelectionResult`. To avoid carriage of state through multiple
functions, I was considering to just return the fee amount for the selected
inputs and then add it to the fee amount of the rest of the transaction (inside
`create_tx`).

danielabrozzoni:
```
pub enum Excess {
    NoChange(Error::InsufficientFunds),
    // Change(change_output, change_output_fee)
    Change(TxOut, u64),
}
```

Written like this, `Excess::NoChange` always contains an `Error::InsufficientFunds`. Is this what you want to achieve? If so, why? Couldn't you have no change while having enough funds?

murchandamus:
> I'm not sure that I've understood what you said correctly. You were thinking in something like the following situation?
> 
>     * `Tx_A` ----> UTXO_A (fee rate = 5, confirmed)
> 
>     * `Tx_B` ----> UTXO_B(fee rate = 3, unconfirmed)
> 
>     * (UTXO_A, UTXO_B) ----> `Tx_C`(fee rate = 5).
> 
> 
> The original transaction that generated UTXO_B is `Tx_B`, and coin select re-bumps the fee for it (not implemented currently), lets call it `Tx_B_bis`. You said that the extra amount that comes from [`Tx_B.fee_amount` - `Tx_B_bis.fee_amount`] can now be considered as part of the fee amount of `Tx_C` and waste can account it?

Yeah, if you spend an unconfirmed input from a transaction that paid a lower feerate than the new transaction you're creating, you need to add more fees to the new transaction so that the transaction package in total will have the intended feerate. I.e. in your example:

Tx_A— vsize: 200 vB, fee: 1000 sats, feerate: 5 sat/vB ↦ UTXO_A
Tx_B— vsize:150 vB, fee: 450 sats, feerate: 3 sat/vB ↦ UTXO_B

If Tx_C is spending UTXO_A and UTXO_B and has a vsize of 200 vB, how much fees should it pay to reach an effective feerate of 5 sat/vB?
⇒ Tx_A has the same feerate already, so doesn't need to be bumped. Tx_B pays less than Tx_C, so Tx_C has to add `(5 sat/vB - 3 sat/vB)×Tx_B.vsize = 300 sats` to bump Tx_B to 5 sat/vB, and must pay 200 vB × 5 sat/vB for itself, i.e. if Tx_C pays a total of 1,300 sat/vB the transaction package {Tx_A, Tx_B, Tx_C} will have an effective feerate of 5 sat/vB.

csralvall:
> ```
> pub enum Excess {
>     NoChange(Error::InsufficientFunds),
>     // Change(change_output, change_output_fee)
>     Change(TxOut, u64),
> }
> ```
> 
> Written like this, `Excess::NoChange` always contains an `Error::InsufficientFunds`. Is this what you want to achieve? If so, why? Couldn't you have no change while having enough funds?

I realized that I can't do that. The idea came up because I was to tied up to the code in `create_tx`.
I will probably turn to the structure used in #630.

danielabrozzoni:

(Specifically, the coin selection is being rewritten from scratch)

Issue #521, opened 2022-01-10
Coins not discovered within initial sync might never be found


thomaseizinger:
**Describe the bug**  

BDK does not seem to always find all coins on-chain if the wallet has been in use for a while and we don't have a local database.

**To Reproduce**  

1. Use in-memory database.
2. Have a wallet whose external descriptor has been used for more than CACHE_ADDR_BATCH_SIZE addresses.

I've made a private repository here (contains the testnet coin descriptor, please don't steal them :upside_down_face: ) and invited a few select maintainers of this repository: https://github.com/thomaseizinger/bdk-missing-coins-bug

**Expected behavior**  

BDK to always find all coins as long as they are within a specified `stop_gap`.

**Build environment**  
 - BDK tag/commit: **0.15.0**
 - OS+version: **Linux thomas-thinkpad 5.10.89-1-MANJARO**
 - Rust/Cargo version: **1.57.0**
 - Rust/Cargo target: **x86_64-unknown-linux-gnu**

**Additional context**  

I've tested this also with the esplora-ureq backend but without luck.

thomaseizinger:
I think the issue might be because we always go from `0` to `max_address_param`: https://github.com/bitcoindevkit/bdk/blob/fdb272e039b650c6f877b20b969fcc07fe086e84/src/wallet/mod.rs#L1506

Instead, what I think we should be doing is going from `last_index_with_activity` to `last_index_with_activity` + `max_address_param`! That way, consecutive syncs of the wallet will scan forward in batches until there are at least `max_address_param` addresses without activity.

LLFourn:
Yes. I've noticed this as well. IMO what we should be doing is caching on-demand rather than batching some arbitrary number stuff up front. I think #461 made this easier.

Issue #165, opened 2020-11-16
Futures returned in library are not send


LLFourn:
Send is explicitly opted out of here: https://github.com/bitcoindevkit/bdk/blob/43cb0331bffebfc103bf2adec956e998290e097d/macros/src/lib.rs#L49.

This is a bit of a problem. For example, after potentially receiving some funds you might like to do:

```
tokio::spawn(wallet.sync(...).then(|res| info!(the result of the sync was "{:?}", res))
```

To sync the wallet in the background. This is not possible since the future must be Send to send to `tokio::spawn`.
It will take some re-engineering of the internals to make this work. 

@afilini do you have opinions on how to address this?

afilini:
Isn't this somewhat connected to the discussion we had about making the wallet thread safe? If the wallet was thread safe then you could spawn a background thread and call `sync()` in that thread, without necessarily having to spawn a task using tokio and going through the async interface.

Actually, I think having a thread-safe wallet is probably the first requirement to then make the futures `Send`, otherwise it doesn't make sense to have a background task run `sync()` if you can't do anything else on the wallet because there's a Mutex that locks everything.

I guess the biggest problem with thread-safe wallets in general is that the database is in an inconsistent state during `sync`. We would probably need to have transactions of some sorts, so that you only commit the full changes at the end of `sync` and always keep the database in a consistent state.

afilini:
So unless I have misunderstood something, I guess there are a few steps before we can get to working on Send futures:

1. Make `Database` thread-safe
2. Add transactional operations to `Database`
3. Update all the `Blockchain` types to use the transactions with `Database`
4. Make `Blockchain` thread-safe, or wrap it in a Mutex. Here we need some kind of locking mechanism to avoid having multiple `sync` running at the same time
5. Make `Wallet` thread-safe, remove `RefCell`s. At this point you can already achieve the "sync in background" by spawning a thread and calling `sync` inside
6. ??? to make futures `Send`

notmandatory:
I agree the first priorities should be to make the `Database`, `Blockchain` and `Wallet` thread safe. For our `bdk-jni` android example the app code is spawning it's own threads from the JVM to sync the wallet or get balance updates, etc. So if `bdk` blocks on those calls it's not a problem.

LLFourn:
[...]

PR #560, opened 2022-03-08
Make futures send

### LINKED ISSUE #559: `maybe_async` complicates isomorphic use-cases
The `maybe_async` macro makes the function signature of each method different depending on whether the dependency is compiled for wasm32 or not. This isn't ideal for libraries that might be used in isomoprhic app frameworks like [iced](https://github.com/hecrj/iced), which can compile to either web or desktop-native targets.

I'm not sure what the best approach would be to fix this since [Rust's async story is so complex right now](https://www.ncameron.org/blog/portable-and-interoperable-async-rust/). If at some point we enter the shiny future, we can just remove usage of the bdk-macros crate altogether.

For now, I think it's possible to bypass this issue by also using the `bdk-macros` crate, but this is imperfect because it essentially perpetuates the problem, and "infects" all upstream dependents.

LLFourn:
Draft PR that will eventually fix #559 #165 #521

Don't review yet just here so coauthors @thomaseizinger @nickfarrow @notmandatory can keep track.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Removed default verification from `wallet::sync`. sync-time verification is added in `script_sync` and is activated by `verify` feature flag.
 - `verify` flag removed from `TransactionDetails`.
-- Removed Blockchain from Wallet.
-- Removed `Wallet::broadcast` (just use blockchain.broadcast)
+
+### Sync API change
+
+To decouple the `Wallet` from the `Blockchain` we've made major changes:
+
+- Removed `Blockchain` from Wallet.
+- Removed `Wallet::broadcast` (just use `Blockchain::broadcast`)
 - Depreciated `Wallet::new_offline` (all wallets are offline now)
-- Changed `Wallet::sync` to take a blockchain argument.
+- Changed `Wallet::sync` to take a `Blockchain`.
+- Stop making a request for the block height when calling `Wallet:new`.
+- Added `SyncOptions` to capture extra (future) arguments to `Wallet::sync`.
 
 ## [v0.16.1] - [v0.16.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ To decouple the `Wallet` from the `Blockchain` we've made major changes:
 - Changed `Wallet::sync` to take a `Blockchain`.
 - Stop making a request for the block height when calling `Wallet:new`.
 - Added `SyncOptions` to capture extra (future) arguments to `Wallet::sync`.
+- Removed `max_addresses` sync parameter which determined how many addresses to cache before syncing since this can just be done with `ensure_addreses_cached`.
 
 ## [v0.16.1] - [v0.16.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Removed default verification from `wallet::sync`. sync-time verification is added in `script_sync` and is activated by `verify` feature flag.
 - `verify` flag removed from `TransactionDetails`.
+- Removed Blockchain from Wallet.
+- Removed `Wallet::broadcast` (just use blockchain.broadcast)
+- Depreciated `Wallet::new_offline` (all wallets are offline now)
+- Changed `Wallet::sync` to take a blockchain argument.
 
 ## [v0.16.1] - [v0.16.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ To decouple the `Wallet` from the `Blockchain` we've made major changes:
 - Changed `Wallet::sync` to take a `Blockchain`.
 - Stop making a request for the block height when calling `Wallet:new`.
 - Added `SyncOptions` to capture extra (future) arguments to `Wallet::sync`.
-- Removed `max_addresses` sync parameter which determined how many addresses to cache before syncing since this can just be done with `ensure_addreses_cached`.
+- Removed `max_addresses` sync parameter which determined how many addresses to cache before syncing since this can just be done with `ensure_addresses_cached`.
 
 ## [v0.16.1] - [v0.16.0]
 
--- a/src/blockchain/esplora/ureq.rs
+++ b/src/blockchain/esplora/ureq.rs
@@ -24,7 +24,7 @@ use ureq::{Agent, Proxy, Response};
 use bitcoin::consensus::{deserialize, serialize};
 use bitcoin::hashes::hex::{FromHex, ToHex};
 use bitcoin::hashes::{sha256, Hash};
-use bitcoin::{BlockHeader, Script, Transaction, Txid};
+use bitcoin::{BlockHeader, Script, Transaction, TxOut, Txid};
 
 use super::api::Tx;
 use crate::blockchain::esplora::EsploraError;
@@ -46,8 +46,17 @@ struct UrlClient {
 #[derive(Debug)]
 pub struct EsploraBlockchain {
     url_client: UrlClient,
-    stop_gap: usize,
     concurrency: u8,
+}
+
+pub struct SyncSession {}
+pub struct WalletUpdate {
+    transactions: Vec<(Vec<Option<TxOut>>, Transaction)>,
+    last_active_index: Vec<usize>,
+}
+
+pub struct SetupSession {
+    scripts: Vec<(Box<dyn Iterator<Item = Script> + Send + 'static>, usize)>,
 }
 
 impl EsploraBlockchain {
@@ -59,7 +68,6 @@ impl EsploraBlockchain {
                 agent: Agent::new(),
             },
             concurrency: super::DEFAULT_CONCURRENT_REQUESTS,
-            stop_gap,
         }
     }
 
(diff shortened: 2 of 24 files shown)


johncantrell97:
probably not directly relevant for this PR but @notmandatory thought it might be helpful to share some recent thoughts I had about syncing a wallet when working with ldk and bdk.

LDK provides the `lightning-block-sync` utility (https://docs.rs/lightning-block-sync/0.0.105/lightning_block_sync/) which helps bring and keep your channel state in sync with the chain by polling a `BlockSource`.  I was thinking since I'm already using this for ldk it would be awesome to use the same interface for keeping the bdk wallet/database in sync.

I guess one way to look at it would be to implement BlockSource for a Blockchain and then use the Blockchain to keep channel state in sync.

The other way was to implement ldk's chain::Listen interface for either bdk's Wallet or Database objects so they could be fed blocks directly using lightning-block-sync utility and then use those blocks to update the database without needing a `Blockchain` at all.  

I guess you could actually do both of these ideas.  The latter is an alternative 'sync' method where the user is in charge of pushing blocks to the wallet instead of just calling "sync".

I'm not sure if these ideas are relevant to this work but thought I'd share just in case.

thomaseizinger:
They are somewhat relevant, thanks for sharing!

Our sync interface will center around _scripts_. You should be able to track a lightning channel by taking the multi-sig or punish transaction script and "syncing" that :)

LLFourn:
> The other way was to implement ldk's chain::Listen interface for either bdk's Wallet or Database objects so they could be fed blocks directly using lightning-block-sync utility and then use those blocks to update the database without needing a `Blockchain` at all.

I don't think it makes sense to feed blocks to the database directly. What if the block you are feeding is not the child of the last block you fed? In LDK they have a `BlockSource` which needs to provide any block via block header which makes it easy to solve this problem.

I think what you are after is being able to sync quickly every time there's a new block. Indeed our work here should make this much easier. So the API we are aiming for using esplora or electrum etc would be:

```rust
let sync_session = wallet.start_sync(options)?;
let update = esplora.sync(&sync_session)?;
wallet.apply_update(update)?;
```

So with a more real-time block approach we could make this something like:

```rust
loop {
   let update = block_source.look_for_updates(addresses).await;
   wallet.apply_update(update);
}
```
So every time `block_source` finds a new transactions related to the addresses it creates a wallet update which is applied immediately and then it goes back to looking.

The tricky thing will be updating `addresses` as the wallet produces more of them. But there are many ways you could go about this and it will depend on your application.

johncantrell97:
Yeah that api would work for esplora / electrum style implementation. 

As for making sure you feed blocks in order all you need to do is track the last block synced. LDK's "lightning-block-sync" module takes an object that implements Listen and it's last known good BlockHash and can bring it and keep it to tip. 

So you just need to implement logic for updating the wallet database for the block connected and block disconnected events. 

I ended up going this route for now but it's a bit of a hack given the limitations of the current bdk api. 

Will watch this PR and hopefully once finalized will allow for a cleaner integration.

LLFourn:
We've sketched some further changes to the syncing on the ureq. Here's an interesting questions that came up:

1. How do we replace the `Progress` trait?
  1. @afilini mentioned that we could just have a `Box<dyn FnMut(usize)>>` passed through and then no need for trait.
  2. @thomaseizinger mentioned that we could track progress by items being consumed from the iterator.
  3. Alternatively we could return an iterator of transactions associated with each address from the ureq implementation which would make it easy to track progress. It would also mean we could get rid of `last_active_index` tracking from the ureq implementation.

LLFourn:
Point 4 above doesn't really work since we need to know additional information like the block height/header the information is valid for. We can't return any results until we know this was consistent from start to finish.

I think most of the work done so far here should be replaced with an incremental update approach and the idea of a full sync is just when you use a stop gap to find updates. Here's some more detail:

```rust
// Generated by the descriptor watcher
struct SyncSession<I> {
   checkpoints: Vec<(u32, BlockHash)>,
   scripts: I // an iterator of scripts
} 

// Generated by the blockchain
struct SyncUpdate {
    transactions: Vec<Transaction>,
    checkpoint_invalidate: Option<(u32, BlockHash)>.
    new_checkpoint: (u32, BlockHash),
}
```

This is missing some detail that's in #165 but the point is that when you start a sync you give the blockchain a list of checkpoints (height and header) that you know about. The blockchain will then try and invalidate one or more of the checkpoints by finding the oldest checkpoint that is no longer valid (the block doesn't exist). It also returns a new checkpoint.

The descriptor watcher then deletes transactions if a checkpoint was invalidated. It then adds the new checkpoint to its list of checkpoints.

thoughts?

notmandatory:
This incremental scheme reminds me a bit of how CBF blockchain sync works (https://github.com/bitcoindevkit/bdk/issues/81), comparing block height/hash checkpoints with peers before downloading filters. 

Is a `Database` (with a `Wallet`) an example of a descriptor watcher? if using an async Database and Blockchain would it look something like below?
```rust
let sync_session = my_database.sync_session(&my_wallet).await?; // get checkpoints from db, scripts from wallet
let sync_update = my_blockchain.sync(sync_session).await?; // plus some sort of progress callback
my_database.update(sync_update).await?; // if database updates are large may need a progress callback here too
my_database.list_transactions().await?; // a bit awkward but I'm not sure how else to do this async
```

LLFourn:
> This incremental scheme reminds me a bit of how CBF blockchain sync works (#81), comparing block height/hash checkpoints with peers before downloading filters.
> 
Good. I hope this makes it easier.

> Is a `Database` (with a `Wallet`) an example of a descriptor watcher? if using an async Database and Blockchain would it look something like below?
> 

Yeah a `DescrioptorWatcher` or maybe `DescriptorManager` is essentially what a `Wallet` is now except it only has one keychain and maybe doesn't have signers. It just keeps a database of related stuff and processes updates. So yes a bit of a wrapped around a database (maybe!).

I dunno how we're going to integrate async databases into this atm. What @thomaseizinger has pointed out to me is that you can pretty much have all txs and indexes associated with your wallet in memory at the same time without it costing much. Sure an exchange with hundreds of thousands of tx's will need a gigabyte of memory but that's fine.

If we want async storage then what we might wanna do is provide a way of syncing the in memory database back (and maybe forth?) with a more permanent database via an event sourcing kind of approach which could be `async`. I don't have a strong vision here thought but I do want to try and avoid more async traits.

thomaseizinger:
> Point 4 above doesn't really work since we need to know additional information like the block height/header the information is valid for. We can't return any results until we know this was consistent from start to finish.
> 
> I think most of the work done so far here should be replaced with an incremental update approach and the idea of a full sync is just when you use a stop gap to find updates. Here's some more detail:
> 
> ```rust
> // Generated by the descriptor watcher
> struct SyncSession<I> {
>    checkpoints: Vec<(u32, BlockHash)>,
>    scripts: I // an iterator of scripts
> } 
> 
> // Generated by the blockchain
> struct SyncUpdate {
>     transactions: Vec<Transaction>,
>     checkpoint_invalidate: Option<(u32, BlockHash)>.
>     new_checkpoint: (u32, BlockHash),
> }
> ```
> 
> This is missing some detail that's in #165 but the point is that when you start a sync you give the blockchain a list of checkpoints (height and header) that you know about. The blockchain will then try and invalidate one or more of the checkpoints by finding the oldest checkpoint that is no longer valid (the block doesn't exist). It also returns a new checkpoint.
> 
> The descriptor watcher then deletes transactions if a checkpoint was invalidated. It then adds the new checkpoint to its list of checkpoints.
> 
> thoughts?

I think this makes sense. What I'd like to mention is that from a dependency PoV, having a (shared) struct is maybe not a good idea. If you want to move each blockchain backend into its own crate, they would have to depend on some shared crate that contains the struct that is _not_ bdk so that the bdk facade crate can then bundle all backends together and re-export them. But maybe we need such a shared "core" crate anyway so those structs could go in there.

thomaseizinger:
One more question: Why would we need to pass in _multiple_ checkpoints per script iterator? Wouldn't one be enough assuming that script iterator == descriptor and thus we call this backend API multiple times?

thomaseizinger:
> If we want async storage then what we might wanna do is provide a way of syncing the in memory database back (and maybe forth?) with a more permanent database via an event sourcing kind of approach which could be `async`.

Just a minor note: The syncing can be achieved in various ways, not necessarily just with event sourcing although it might lead itself naturally to it to record all changes to the memory database as events and upon sync, apply them to the persistent one (but not store the events there).

LLFourn:
> Just a minor note: The syncing can be achieved in various ways, not necessarily just with event sourcing although it might lead itself naturally to it to record all changes to the memory database as events and upon sync, apply them to the persistent one (but not store the events there).

I think that this can be modelled by the sync checkpoints themselves. Each sync checkpoint represents a change in the database and if we can make it easy to find all changes associated with each checkpoint I think we've won (that also makes it easy to roll back a checkpoint if that checkpoint becomes invalid).

> One more question: Why would we need to pass in _multiple_ checkpoints per script iterator? Wouldn't one be enough assuming that script iterator == descriptor and thus we call this backend API multiple times?

I don't understand what you meant in the second sentence. The reason why you pass in multiple is so that we know which checkpoints are invalid and which ones are still valid so we can apply the change even if one of them was invalid.

thomaseizinger:
> > One more question: Why would we need to pass in _multiple_ checkpoints per script iterator? Wouldn't one be enough assuming that script iterator == descriptor and thus we call this backend API multiple times?
> 
> I don't understand what you meant in the second sentence. The reason why you pass in multiple is so that we know which checkpoints are invalid and which ones are still valid so we can apply the change even if one of them was invalid.

I don't understand that. The checkpoint is the current height right as known to the wallet / database? Why is it not good enough for the sync backend to be told what the latest height is that we have in the database? The sync backend function (the one we prototyped for esplora) should IMO outright fail if the hash at the given height is no longer the same.

As a result, the overall sync function has to grab an earlier checkpoint from the DB and try syncing again. You wouldn't want to reimplement this logic for every backend so I think it shouldn't live in there but in an overall sync function that looks like this:

```rust
mod bdk {
	// Sub-module for blocking API, modeled after `reqwest::blocking`
	mod blocking {
			pub fn sync(backend: &impl FetchRelatedTransactionsBlocking, wallet: &Wallet) {
				// for each descriptor in wallet
				// fetch related transactions
				// apply result to wallet

				// if any error during syncing, apply fallback logic
				// for example, if sync backend fails because hash at height differs, restart this loop and grab an earlier checkpoint from the wallet
			}
	}

	pub async fn sync(backend: &impl FetchRelatedTransactions, wallet: &Wallet) {
		// same as above but async version of `FetchRelatedTransactions`
	}
}
```

Edit: IMO we should strive for the logic to be implement by every backend to be as minimal as possible and try to reuse as much as we can across all of them.

LLFourn:
@thomaseizinger Ok I think I can get on board with that approach. I don't mind leaving it up to another layer of logic to collect all the invalid checkpoints!

thomaseizinger:
> @thomaseizinger Ok I think I can get on board with that approach. I don't mind leaving it up to another layer of logic to collect all the invalid checkpoints!

What I'd like to point out is that such a design will push us towards `FetchRelatedTransactions` or whatever we call it to be a good building block. If for whatever reason, our sync logic is not suitable for a `bdk` user, then they can always copy-paste the above and modify to their liking and things will still work!

CHANGELOG.md
```diff
@@ -11,6 +11,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `get_internal_address` to allow you to get internal addresses just as you get external addresses.
 - added `ensure_addresses_cached` to `Wallet` to let offline wallets load and cache addresses in their database
 
+### Sync API change
+
+To decouple the `Wallet` from the `Blockchain` we've made major changes:
+
+- Removed `Blockchain` from Wallet.
+- Removed `Wallet::broadcast` (just use `Blockchain::broadcast`)
+- Depreciated `Wallet::new_offline` (all wallets are offline now)
```
cryptoquick, line 20: *Deprecated

src/blockchain/esplora/ureq.rs
```diff
@@ -87,13 +151,37 @@ impl Blockchain for EsploraBlockchain {
...
+impl WalletSync for EsploraBlockchain {
+    fn wallet_setup<D: BatchDatabase>(
         &self,
         database: &mut D,
-        _progress_update: P,
+        _progress_update: Box<dyn Progress>,
     ) -> Result<(), Error> {
         use crate::blockchain::script_sync::Request;
-        let mut request = script_sync::start(database, self.stop_gap)?;
+        let mut request = script_sync::start(database, 20)?;
```
cryptoquick, line 184: I'd suggest bringing this magic number out into a constant

cryptoquick (commented):
This is a clever approach. I just thought of another case where I could really make use of a unified interface between wasm and desktop builds; I'm looking to make a benchmark without it needing to be browser-side, so I took the time to look things over. You have my concept ACK.

LLFourn:
hey @cryptoquick, this idea balooned into https://bitcoindevkit.org/blog/bdk-core-pt1/
This is the approach I'm taking to solve this now. Feedback really welcome.

I'll close this.

PR #562, opened 2022-03-09
Expose more getters in Wallet and other useful descriptor traits

afilini:
### Description

Expose more getters and internal utilities for people who need to work with descriptors.

A good example of how this can be leveraged is in the second commit, which refactors the wallet export functionality to use the new public APIs rather than using members on `Wallet` directly.

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+
+- Added `Wallet::get_signers()`, `Wallet::descriptor_checksum()` and `Wallet::get_address_validators()`, exposed the `AsDerived` trait.
+
 ## [v0.17.0] - [v0.16.1]
 
 - Removed default verification from `wallet::sync`. sync-time verification is added in `script_sync` and is activated by `verify` feature flag.
--- a/src/wallet/export.rs
+++ b/src/wallet/export.rs
@@ -64,9 +64,10 @@ use std::str::FromStr;
 use serde::{Deserialize, Serialize};
 
 use miniscript::descriptor::{ShInner, WshInner};
-use miniscript::{Descriptor, DescriptorPublicKey, ScriptContext, Terminal};
+use miniscript::{Descriptor, ScriptContext, Terminal};
 
 use crate::database::BatchDatabase;
+use crate::types::KeychainKind;
 use crate::wallet::Wallet;
 
 /// Structure that contains the export of a wallet
@@ -117,8 +118,12 @@ impl WalletExport {
         include_blockheight: bool,
     ) -> Result<Self, &'static str> {
         let descriptor = wallet
-            .descriptor
-            .to_string_with_secret(&wallet.signers.as_key_map(wallet.secp_ctx()));
+            .get_descriptor_for_keychain(KeychainKind::External)
+            .to_string_with_secret(
+                &wallet
+                    .get_signers(KeychainKind::External)
+                    .as_key_map(wallet.secp_ctx()),
+            );
         let descriptor = remove_checksum(descriptor);
         Self::is_compatible_with_core(&descriptor)?;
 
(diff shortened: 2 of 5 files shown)


src/wallet/mod.rs
```diff
@@ -1132,6 +1160,11 @@ where
         descriptor
     }
 
+    /// Returns whether the wallet has a dedicated internal descriptor
+    pub fn has_internal_descriptor(&self) -> bool {
+        self.change_descriptor.is_some()
+    }
```
LLFourn: Note you can already just call `wallet.public_descriptor(KeychainKind::Internal).is_some()`

src/wallet/mod.rs
```diff
@@ -428,6 +428,21 @@ where
...
+    ///
+    /// Note that this methods only operate on the internal database, which first needs to be
+    /// [`Wallet::sync`] manually.
+    pub fn get_transaction(
+        &self,
+        txid: &Txid,
+        include_raw: bool,
+    ) -> Result<Option<TransactionDetails>, Error> {
+        self.database.borrow().get_tx(txid, include_raw)
+    }
```
rajarshimaitra: This function seems exactly same as `wallet::get_tx()`. I couldn't figure out why this one is duplicated..

src/wallet/mod.rs
```diff
@@ -456,13 +471,41 @@ where
...
 
+    /// Get the signers
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// # use bdk::{Wallet, KeychainKind};
+    /// # use bdk::bitcoin::Network;
+    /// # use bdk::database::MemoryDatabase;
+    /// let wallet = Wallet::new_offline("wpkh(tprv8ZgxMBicQKsPe73PBRSmNbTfbcsZnwWhz5eVmhHpi31HW29Z7mc9B4cWGRQzopNUzZUT391DeDJxL2PefNunWyLgqCKRMDkU1s2s8bAfoSk/84'/0'/0'/0/*)", None, Network::Testnet, MemoryDatabase::new())?;
```
rajarshimaitra: This line in the example needs to be modified in light of https://github.com/bitcoindevkit/bdk/pull/535

src/wallet/export.rs
```diff
@@ -142,12 +147,24 @@ impl WalletExport {
...
+                    .to_string_with_secret(
+                        &wallet
+                            .get_signers(KeychainKind::Internal)
+                            .as_key_map(wallet.secp_ctx()),
+                    );
+                Some(remove_checksum(descriptor))
+            }
         };
-        if export.change_descriptor() != wallet.change_descriptor.as_ref().map(desc_to_string) {
+        if export.change_descriptor() != change_descriptor {
```
rajarshimaitra, line 167: Does this line means we cannot have two completely different descriptors as external and internal? They change always have to be `/1/*`?

rajarshimaitra (commented):
ACK https://github.com/bitcoindevkit/bdk/pull/562/commits/b5cb0f73dac5aad60e2ee84479b07ca0cf0894e8

Code changes looks good to me. Tried the doc example and it works. Agree with the idea to expose more descriptor level functionality to users..

Below are few comments and questions

src/wallet/mod.rs
afilini: Because I'm an idiot and I didn't notice we already had that...

src/wallet/export.rs
afilini: Yes, unfortunately this export format only supports descriptors that match this specific structure. See [the spec here](https://github.com/Fonta1n3/FullyNoded/blob/10b7808c8b929b171cca537fb50522d015168ac9/Docs/Wallets/Wallet-Export-Spec.md#change-keys)

rajarshimaitra: Ya the spec does makes sense.. And its also nasty to generalize it.. 
I am just thinking on what happens when someone doesn't respect the the spec at wallet creation, and then finds it out at the export.. We do not seem to be enforcing this at `wallet::new`..

afilini: I don't think we should enforce this in the wallet constructor, it would be very limiting. I think it's fine that some descriptors can't be exported using this protocol, this is meant to be used to share descriptors with FullyNoded specifically and if you don't care about it you should not be limited in the set of descriptors you can use.

Maybe we should make it more clear that this is a FullyNoded-specific export, not a "generic" one. And we could design a different spec for arbitrary descriptors ourselves, or see if there's something else out there that we can reuse.

rajarshimaitra: > Maybe we should make it more clear that this is a FullyNoded-specific export, not a "generic" one.

Ya that makes sense.. I am just worried about someone not noticing this and get into trouble later. Ya explicitly clarifying it would be helpful. And we can figure the generic way later on..

notmandatory:
It looks like this one is ready for the next release, @rajarshimaitra or @LLFourn since you've both taken a look already would one of your like to approve it?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `get_internal_address` to allow you to get internal addresses just as you get external addresses.
 - added `ensure_addresses_cached` to `Wallet` to let offline wallets load and cache addresses in their database
 - Add `is_spent` field to `LocalUtxo`; when we notice that a utxo has been spent we set `is_spent` field to true instead of deleting it from the db.
+- Added `Wallet::get_signers()`, `Wallet::descriptor_checksum()` and `Wallet::get_address_validators()`, exposed the `AsDerived` trait.
```
rajarshimaitra: This should be in unreleased section?

rajarshimaitra (approved):
ReACK https://github.com/bitcoindevkit/bdk/pull/562/commits/11c091ef09cf849313b3f1d9d7b3ccde8c8b5974

Just one nit.

CHANGELOG.md
notmandatory: I fixed it with an amended last commit.

notmandatory (approved):
ACK 8cd055090d75669c67a44467cc0281465ef31270

Issue #549, opened 2022-02-17
Feature Request: BIP47 payment codes for private donations (PayNym)


satrinity402:
In the current environment, it would be extremely beneficial if bitcoin wallet developers using BDK could have an easy way to integrate [BIP47] payment codes into their wallets for receiving payments privately (especially donations) without needing to setup a server to generate fresh addresses for each payment. Would need functionality for sending to other BIP47 payment codes as well. 

Both Samourai Wallet and Sparrow Wallet are FOSS, and have integrated this functionality into their wallets.

[BIP47]: https://github.com/bitcoin/bips/blob/master/bip-0047.mediawiki

artdesignbySF:
I am for this enhancement and would love to see it adopted by all major wallets.

BitcoinQnA:
Another +1 here

satrinity402:
Looks like there is already a BIP47 implementation for Rust: https://github.com/rust-bitcoin/rust-bip47

modl21:
> I am for this enhancement and would love to see it adopted by all major wallets.

I echo this sentiment.

luke-jr:
The goal of a reusable donation address is great, but BIP47 is a bad way to do it. Recommend against implementing.

rottenwheel:
@luke-jr Yeah, that's why we should always post our legacy BTC addresses to receive donations; everyone can snoop in for eternity, right? No SegWit, that must be a bad way to do it too. We need to advise against BIP47 and SegWit, all the way.

- https://oxt.me/address/1A6eLsi6cmm2xKDSGThd8C2kfzDBXD8YTw
- https://twitter.com/LukeDashjr/status/1468009813913309193

A toast for Bitcoin's efficiency and privacy!

ElmoOutlaw:
Let's not get ahead of ourselves here!
We have spent years trying to ignore BIP47 and have spent a huge deal of energy and time trying to push the narrative against the Red Wall that we just can't possibly start to implement it into our wallets.
Don't get me wrong it's nothing to do with privacy, it's more of a save face kinda thang.
# Just HODL thru it y'all.

BitcoinQnA:
> The goal of a reusable donation address is great, but BIP47 is a bad way to do it. Recommend against implementing.

@luke-jr It would be useful if you could explain the technicals behind why you think this. Saying 'it's bad' doesn't help.

apemithrandir:
Are we doing a bounty for this one?
[...]

PR #569, opened 2022-03-15
[blockchain] Add traits to reuse `Blockchain`s across multiple wallets

afilini:
### Description

Add three new traits:
- `StatelessBlockchain` is used to tag `Blockchain`s that don't have any wallet-specic state, i.e. they can be used as-is to sync multiple wallets.
- `StatefulBlockchain` is the opposite of `StatelessBlockchain`: it provides a method to "clone" a `Blockchain` with an updated internal state (a new wallet checksum and, optionally, a different number of blocks to skip from genesis). Potentially this also allows reusing the underlying connection on `Blockchain` types that support it.
- `MultiBlockchain` is a generalization of this concept: it's implemented automatically for every type that implements `StatefulBlockchain` and for every `Arc<T>` where `T` is a `StatelessBlockchain`. This allows a piece of code that deals with multiple sub-wallets to just get a `&B: MultiBlockchain` without having to deal with stateful and statless blockchains individually.

These new traits have been implemented for Electrum, Esplora and RPC (the first two being stateless and the latter stateful). It hasn't been implemented on the CBF blockchain, because I don't think it would work in its current form (it throws away old block filters, so it's hard to go back and rescan).

This is the first step for bitcoindevkit/bdk#549, as BIP47 needs to sync many different descriptors internally.

It's also very useful for bitcoindevkit/bdk_wallet#188.

### Notes to the reviewers

This is still a draft because:
- I'm still wondering if these traits should "inherit" from `Blockchain` instead of the less-restrictive `WalletSync` + `GetHeight` which is the bare minimum to sync a wallet
- I need to write tests, at least for rpc which is stateful
- I need to add examples

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - added `OldestFirstCoinSelection` impl to `CoinSelectionAlgorithm`
 - New MSRV set to `1.56`
+- Add traits to reuse `Blockchain`s across multiple wallets (`BlockchainFactory` and `StatelessBlockchain`).
 
 
 ## [v0.18.0] - [v0.17.0]
--- a/src/blockchain/electrum.rs
+++ b/src/blockchain/electrum.rs
@@ -79,6 +79,8 @@ impl Blockchain for ElectrumBlockchain {
     }
 }
 
+impl StatelessBlockchain for ElectrumBlockchain {}
+
 impl GetHeight for ElectrumBlockchain {
     fn get_height(&self) -> Result<u32, Error> {
         // TODO: unsubscribe when added to the client, or is there a better call to use here?
@@ -320,8 +322,67 @@ impl ConfigurableBlockchain for ElectrumBlockchain {
 
 #[cfg(test)]
 #[cfg(feature = "test-electrum")]
-crate::bdk_blockchain_tests! {
-    fn test_instance(test_client: &TestClient) -> ElectrumBlockchain {
-        ElectrumBlockchain::from(Client::new(&test_client.electrsd.electrum_url).unwrap())
+mod test {
+    use std::sync::Arc;
+
+    use super::*;
+    use crate::database::MemoryDatabase;
+    use crate::testutils::blockchain_tests::TestClient;
+    use crate::wallet::{AddressIndex, Wallet};
+
+    crate::bdk_blockchain_tests! {
+        fn test_instance(test_client: &TestClient) -> ElectrumBlockchain {
+            ElectrumBlockchain::from(Client::new(&test_client.electrsd.electrum_url).unwrap())
+        }
+    }
+
+    fn get_factory() -> (TestClient, Arc<ElectrumBlockchain>) {
+        let test_client = TestClient::default();
+
+        let factory = Arc::new(ElectrumBlockchain::from(
+            Client::new(&test_client.electrsd.electrum_url).unwrap(),
+        ));
+
+        (test_client, factory)
+    }
+
+    #[test]
+    fn test_electrum_blockchain_factory() {
+        let (_test_client, factory) = get_factory();
+
+        let a = factory.build("aaaaaa", None).unwrap();
+        let b = factory.build("bbbbbb", None).unwrap();
+
+        assert_eq!(
+            a.client.block_headers_subscribe().unwrap().height,
+            b.client.block_headers_subscribe().unwrap().height
+        );
+    }
+
+    #[test]
+    fn test_electrum_blockchain_factory_sync_wallet() {
+        let (mut test_client, factory) = get_factory();
+
+        let db = MemoryDatabase::new();
+        let wallet = Wallet::new(
+            "wpkh(L5EZftvrYaSudiozVRzTqLcHLNDoVn7H5HSfM9BAN6tMJX8oTWz6)",
+            None,
+            bitcoin::Network::Regtest,
+            db,
+        )
+        .unwrap();
+
+        let address = wallet.get_address(AddressIndex::New).unwrap();
+
+        let tx = testutils! {
+            @tx ( (@addr address.address) => 50_000 )
+        };
+        test_client.receive(tx);
+
+        factory
+            .sync_wallet(&wallet, None, Default::default())
+            .unwrap();
+
+        assert_eq!(wallet.get_balance().unwrap(), 50_000);
     }
 }
--- a/src/blockchain/mod.rs
+++ b/src/blockchain/mod.rs
@@ -25,7 +25,8 @@ use bitcoin::{Transaction, Txid};
 
 use crate::database::BatchDatabase;
 use crate::error::Error;
-use crate::FeeRate;
+use crate::wallet::{wallet_name_from_descriptor, Wallet};
+use crate::{FeeRate, KeychainKind};
 
 #[cfg(any(
     feature = "electrum",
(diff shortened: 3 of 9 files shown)


afilini:
I didn't like how `StatefulBlockchain` worked, so I just removed that trait since it was very similar to `MultiBlockchain` anyway. I've also renamed `MultiBlockchain` to `BlockchainFactory` (@notmandatory used this term in yesterday's call and I really liked it).

I've also updated all these traits to inherit from `Blockchain` rather than `GetHeight + WalletSync`: the rationale is that it's easier to relax these requirements later rather than tightening them as it would be an API break.

I've added tests and documentation, I'm feeling confident to remove the draft state from this PR.

src/blockchain/rpc.rs
```diff
@@ -441,6 +441,59 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+///     url: "http://127.0.0.1:18332",
+///     auth: Auth::Cookie {
+///         file: "/home/user/.bitcoin/.cookie".to_string(),
+///     },
+///     network: Network::Testnet,
+///     wallet_name_prefix: "prefix-".to_string(),
+///     default_skip_blocks: 100_000,
+/// };
+/// let main_wallet_blockchain = factory.build("main_wallet", Some(200_000))?;
+/// # Ok(())
```
notmandatory, line 463: Missing closing brace in docs for `fn main()`.
```suggestion
/// # Ok(())
/// # }
```

src/blockchain/rpc.rs
```diff
@@ -441,6 +441,59 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
 
+/// Factory of [`RpcBlockchain`] instances, implements [`BlockchainFactory`]
+///
+/// Internally caches the node url and authentication params and allows getting many different [`RpcBlockchain`]
+/// objects for different wallet names and with different rescan heights.
+///
+/// ## Example
+///
+/// ```no_run
+/// # fn main() -> Result<(), Box<dyn std::error::Error>> {
```
notmandatory, line 452: Missing some uses in docs.
```suggestion
/// # use bdk::bitcoin::Network;
/// # use bdk::blockchain::BlockchainFactory;
/// # use bdk::blockchain::rpc::{Auth, RpcBlockchainFactory};
/// # fn main() -> Result<(), Box<dyn std::error::Error>> {
```

src/blockchain/rpc.rs
```diff
@@ -441,6 +441,59 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+/// let factory = RpcBlockchainFactory {
+///     url: "http://127.0.0.1:18332",
```
notmandatory: Missing `to_string()`.
```suggestion
///     url: "http://127.0.0.1:18332".to_string(),
```

src/blockchain/rpc.rs
```diff
@@ -441,6 +441,59 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+///     auth: Auth::Cookie {
+///         file: "/home/user/.bitcoin/.cookie".to_string(),
```
notmandatory: Needs a PathBuf so `.into()` works.
```suggestion
///         file: "/home/user/.bitcoin/.cookie".into(),
```

src/blockchain/mod.rs
```diff
@@ -164,6 +164,65 @@ pub trait ConfigurableBlockchain: Blockchain + Sized {
...
+```no_run
+# use bdk::Error;
+# use bdk::blockchain::*;
+# use bdk::database::*;
+# use bdk::wallet::*;
+fn sum_of_balances<B: BlockchainFactory>(blockchain_factory: B, wallets: &[Wallet<MemoryDatabase>]) -> Result<u64, Error> {
+    Ok(wallets
+        .iter()
+        .map(|w| -> Result<_, Error> {
+            w.sync(&blockchain_factory.build("wallet_1", None)?, SyncOptions::default())?;
```
notmandatory: Would it be more clear to use the `wallet_name_from_descriptor()` function in the example?
```suggestion
            let wallet_name = wallet_name_from_descriptor(...);
            w.sync(&blockchain_factory.build(wallet_name, None)?, SyncOptions::default())?;
```

src/blockchain/mod.rs
```diff
@@ -164,6 +164,65 @@ pub trait ConfigurableBlockchain: Blockchain + Sized {
...
+    type Inner: Blockchain;
+
+    /// Build a new blockchain for the given descriptor checksum
+    ///
+    /// If `override_skip_blocks` is `None`, the returned blockchain will inherit the number of blocks
+    /// from the factory. Since it's not possible to override the value to `None`, set it to
+    /// `Some(0)` to rescan from the genesis.
+    fn build(
+        &self,
+        checksum: &str,
```
notmandatory: Since we're not enforcing this to be some sort of checksum would `wallet_name` be a better field name?
```suggestion
        wallet_name: &str,
```

src/blockchain/rpc.rs
```diff
@@ -441,6 +441,59 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+///     },
+///     network: Network::Testnet,
+///     wallet_name_prefix: "prefix-".to_string(),
+///     default_skip_blocks: 100_000,
+/// };
+/// let main_wallet_blockchain = factory.build("main_wallet", Some(200_000))?;
```
notmandatory, line 462: Should we use a `wallet_name_from_descriptor` value here, or is that too pedantic?

src/blockchain/rpc.rs
```diff
@@ -441,6 +441,59 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+/// ```
+#[derive(Debug, Clone)]
+pub struct RpcBlockchainFactory {
+    /// The bitcoin node url
+    pub url: String,
+    /// The bitcoin node authentication mechanism
+    pub auth: Auth,
+    /// The network we are using (it will be checked the bitcoin node network matches this)
+    pub network: Network,
+    /// The prefix used to build the full wallet name for blockchains
```
notmandatory: I don't understand the purpose of `wallet_name_prefix` if we use unique wallet names. Or i guess this is just in case your core node happens to have a wallet by the same name?

src/blockchain/rpc.rs
```diff
@@ -456,3 +509,49 @@ crate::bdk_blockchain_tests! {
...
+mod test {
+    use super::*;
+    use crate::blockchain::*;
+    use crate::testutils::blockchain_tests::TestClient;
+
+    use bitcoin::Network;
+    use bitcoincore_rpc::RpcApi;
+
+    #[test]
+    fn test_rpc_blockchain_factory() {
```
notmandatory: It'd be nice to have a simple test like this for one of the stateless blockchain factories like `ElectrumBlockchain`.

notmandatory (commented):
Overall LGTM. I commented on some fixes needed in the rpc docs code, and a couple other suggestions.

src/blockchain/rpc.rs
afilini: It's true that it's not entirely required, but I think it could be useful to group all the sub-wallets that belong to the same "main wallet" without having to keep track of the prefix explicitly.

I'm gonna make it an `Option`, so if you are not interested in it you can just use `None` there

src/blockchain/mod.rs
afilini: I could use the descriptor checksum, just because it's shorter and easier to fit in a doc example

src/blockchain/rpc.rs
afilini: I think it's fine here, we don't even have a wallet to take the descriptor from in this example...

src/testutils/mod.rs
```diff
@@ -267,3 +267,5 @@ macro_rules! testutils {
         (external, internal)
     })
 }
+
+pub use testutils;
```
rajarshimaitra: I did not understand why we need to `use testutils` in `testutils::mod` itself..

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `get_internal_address` to allow you to get internal addresses just as you get external addresses.
 - added `ensure_addresses_cached` to `Wallet` to let offline wallets load and cache addresses in their database
 - Add `is_spent` field to `LocalUtxo`; when we notice that a utxo has been spent we set `is_spent` field to true instead of deleting it from the db.
+- Add traits to reuse `Blockchain`s across multiple wallets (`BlockchainFactory` and `StatelessBlockchain`).
```
rajarshimaitra: This should be in unrealesed section..

src/blockchain/electrum.rs
```diff
@@ -325,3 +327,29 @@ crate::bdk_blockchain_tests! {
         ElectrumBlockchain::from(Client::new(&test_client.electrsd.electrum_url).unwrap())
     }
 }
+
+#[cfg(test)]
+#[cfg(feature = "test-electrum")]
+mod test {
```
rajarshimaitra: This test can be converged with the previous `bdk_blockchain_tests!` macro in a single `mod test`..

src/blockchain/electrum.rs
```diff
@@ -325,3 +327,29 @@ crate::bdk_blockchain_tests! {
...
+        let factory = Arc::new(ElectrumBlockchain::from(
+            Client::new(&test_client.electrsd.electrum_url).unwrap(),
+        ));
+
+        let a = factory.build("aaaaaa", None).unwrap();
+        let b = factory.build("bbbbbb", None).unwrap();
+
+        assert_eq!(
+            a.client.block_headers_subscribe_raw().unwrap().height,
+            b.client.block_headers_subscribe().unwrap().height
```
rajarshimaitra: Instead of creating an `Arc::new()` does it make sense to have an api like `BlockchainFactory::from(T : Blockchain)` which returns and `Arc<Blockchain>`.

And then the trait can have a function like `BlockchainFactory::build_for_wallet` which returns the arc clone of the inner blockchain.

So the instantiation would look like
` let a = factory.build_for_wallet("aaaaa", None).unwrap()`.

To me it seems more clear that way for lib users to intuitively understand whats happening..

src/blockchain/mod.rs
```diff
@@ -164,6 +164,72 @@ pub trait ConfigurableBlockchain: Blockchain + Sized {
...
+    /// Build a new blockchain for the given descriptor wallet_name
+    ///
+    /// If `override_skip_blocks` is `None`, the returned blockchain will inherit the number of blocks
+    /// from the factory. Since it's not possible to override the value to `None`, set it to
+    /// `Some(0)` to rescan from the genesis.
+    fn build(
+        &self,
+        wallet_name: &str,
+        override_skip_blocks: Option<u32>,
+    ) -> Result<Self::Inner, Error>;
```
rajarshimaitra, line 217: Does it make sense to add auto syncing in the `build` function? So if I call `factory.build_for_wallet()` then it will take the wallet database and perform and auto sync/setup on first instantiation??

src/blockchain/rpc.rs
```diff
@@ -456,3 +517,49 @@ crate::bdk_blockchain_tests! {
...
+            wallet_name_prefix: Some("prefix-".into()),
+            default_skip_blocks: 0,
+        };
+
+        let a = factory.build("aaaaaa", None).unwrap();
+        assert_eq!(a.skip_blocks, Some(0));
+        assert_eq!(
+            a.client
+                .get_wallet_info()
+                .expect("Node connection is working")
```
rajarshimaitra: Nit
```suggestion
                .expect("Node connection isn't working")
```

src/blockchain/rpc.rs
```diff
@@ -456,3 +517,49 @@ crate::bdk_blockchain_tests! {
...
+                .wallet_name,
+            "prefix-aaaaaa"
+        );
+
+        let b = factory.build("bbbbbb", Some(100)).unwrap();
+        assert_eq!(b.skip_blocks, Some(100));
+        assert_eq!(
+            a.client
+                .get_wallet_info()
+                .expect("Node connection is working")
```
rajarshimaitra: Same Nit on error message.

src/blockchain/rpc.rs
```diff
@@ -456,3 +517,49 @@ crate::bdk_blockchain_tests! {
...
+                .wallet_name,
+            "prefix-aaaaaa"
+        );
+
+        let b = factory.build("bbbbbb", Some(100)).unwrap();
+        assert_eq!(b.skip_blocks, Some(100));
+        assert_eq!(
+            a.client
```
rajarshimaitra: Test failing.. 
```suggestion
            b.client
```

rajarshimaitra (changes requested):
Concept ACK a817545cf2ed0487ec00a933281092774b12852d

I think this would be really useful for multiple wallets to use the same backend. 

Below are some nits and api questions I had..

src/testutils/mod.rs
afilini: I think there were some issues with the macros not being exported correctly, I don't remember.

I could try removing it and seeing what happens

src/blockchain/mod.rs
afilini: I wouldn't do it here in the `build()`, this is just to get a blockchain which you could use for anything, not necessarily just to sync a wallet.

We could add another method to the trait to automatically infer the wallet name from a wallet reference, and maybe even one more to sync a wallet without even returning the new blockchain. Just create it internally, sync, and throw it away.

src/blockchain/rpc.rs
afilini: I interpreted this as "this is what i'm expecting", but yes I think I got it wrong

src/blockchain/electrum.rs
afilini: I think it would be a bit confusing because it would only work for `StatelessBlockchain`. For the others you'd still need to create it manually, like for rpc

afilini:
Rebased and addresses most comments

src/blockchain/mod.rs
notmandatory: Unless we have a specific use case for why someone would want to sync a wallet and not keep the blockchain here I'd say best not to add any new functionality.  But if a specific case comes up where it's needed it can always be added in a future PR.

src/blockchain/electrum.rs
rajarshimaitra: To clear my understanding further, what does it mean exactly by `StatelessBlockchain`?? Like a blockchain structure that doesn't contain or store and wallet level data?? And is interoperable with multiple wallets connecting to it?? 

So in that sense `electrum` and `esplora` are stateful, because they store wallet specific data in them??

src/blockchain/rpc.rs
```diff
@@ -456,3 +517,54 @@ crate::bdk_blockchain_tests! {
         RpcBlockchain::from_config(&config).unwrap()
     }
 }
```
rajarshimaitra: We can move this old test chunk into `mod test` too..

rajarshimaitra (approved):
tACK 6f91598ebc51ca36e52a396bc0429724ed57a35a

Code changes looks good to me. Now I understand the intention better. This will be really useful..

Just one non blocking nit. Can be done in a following PR.

notmandatory:
This one needs one more rebase to pickup the new MSRV CI changes and then it should be ready to go.

rajarshimaitra (approved):
tACK 8795da48397800effeffae8417c6e749ec488212

Issue #567, opened 2022-03-13
Remove Database::flush() function


notmandatory:
Remove `Database::flush()` function added in #409 ~~and call it internally in the `keyvalue::Database` impl in functions that create or update Tree data~~. 

Currently only the `keyvalue::Database` impl does anything in the `Database::flush` function and @thunderbiscuit and I confirmed it needs to be called to fix https://github.com/bitcoindevkit/bdk-ffi/issues/114 on Android. But the `Database::flush` is not used or needed for the `Memory` or `Sqlite` databases, see also https://github.com/bitcoindevkit/bdk-ffi/pull/116#issuecomment-1065942085.

EDIT: as discussed in [#575](https://github.com/bitcoindevkit/bdk/pull/575#issuecomment-1078585739) for now it's enough to only remove the `Database::flush()` function. The auto-flushing is probably not the right approach at this time.

danielabrozzoni:
Fixed by #575

PR #575, opened 2022-03-23
Remove database flush

kafaichoi:
### Description

This PR is to remove Database::flush. See this issue for detail https://github.com/bitcoindevkit/bdk/issues/567

### Notes to the reviewers
The 2nd commit is a small refactoring of adding a new private ivec_to_u32 to avoid too much code duplication. Please let me know if it's ok to include this in this PR or I should make it into a separate PR

Currently existing test cases are shared across for all Databaes implementation so I am not sure if we should add  specific test cases for keyvalue(Tree) for this auto-flush behaviour?(and I feel like it's more a implementation detail). Please let me know how should I proceed for test case in this PR

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ To decouple the `Wallet` from the `Blockchain` we've made major changes:
 - Stop making a request for the block height when calling `Wallet:new`.
 - Added `SyncOptions` to capture extra (future) arguments to `Wallet::sync`.
 - Removed `max_addresses` sync parameter which determined how many addresses to cache before syncing since this can just be done with `ensure_addresses_cached`.
+- remove `flush` method from the `Database` trait.
 
 ## [v0.16.1] - [v0.16.0]
 
--- a/src/database/keyvalue.rs
+++ b/src/database/keyvalue.rs
@@ -402,10 +402,6 @@ impl Database for Tree {
             Ok(val)
         })
     }
-
-    fn flush(&mut self) -> Result<(), Error> {
-        Ok(Tree::flush(self).map(|_| ())?)
-    }
 }
 
 impl BatchDatabase for Tree {
--- a/src/database/keyvalue.rs
+++ b/src/database/keyvalue.rs
@@ -166,16 +166,9 @@ macro_rules! impl_batch_operations {
         fn del_last_index(&mut self, keychain: KeychainKind) -> Result<Option<u32>, Error> {
             let key = MapKey::LastIndex(keychain).as_map_key();
             let res = self.remove(key);
-            let res = $process_delete!(res);
-
-            match res {
-                None => Ok(None),
-                Some(b) => {
-                    let array: [u8; 4] = b.as_ref().try_into().map_err(|_| Error::InvalidU32Bytes(b.to_vec()))?;
-                    let val = u32::from_be_bytes(array);
-                    Ok(Some(val))
-                }
-            }
+            $process_delete!(res)
+            .map(ivec_to_u32)
+            .transpose()
         }
 
         fn del_sync_time(&mut self) -> Result<Option<SyncTime>, Error> {
@@ -357,16 +350,7 @@ impl Database for Tree {
 
     fn get_last_index(&self, keychain: KeychainKind) -> Result<Option<u32>, Error> {
         let key = MapKey::LastIndex(keychain).as_map_key();
-        self.get(key)?
-            .map(|b| -> Result<_, Error> {
-                let array: [u8; 4] = b
-                    .as_ref()
-                    .try_into()
-                    .map_err(|_| Error::InvalidU32Bytes(b.to_vec()))?;
-                let val = u32::from_be_bytes(array);
-                Ok(val)
-            })
-            .transpose()
+        self.get(key)?.map(ivec_to_u32).transpose()
     }
 
     fn get_sync_time(&self) -> Result<Option<SyncTime>, Error> {
@@ -393,15 +377,17 @@ impl Database for Tree {
 
             Some(new.to_be_bytes().to_vec())
         })?
-        .map_or(Ok(0), |b| -> Result<_, Error> {
-            let array: [u8; 4] = b
-                .as_ref()
-                .try_into()
-                .map_err(|_| Error::InvalidU32Bytes(b.to_vec()))?;
-            let val = u32::from_be_bytes(array);
-            Ok(val)
-        })
+        .map_or(Ok(0), ivec_to_u32)
     }
+}
+
+fn ivec_to_u32(b: sled::IVec) -> Result<u32, Error> {
+    let array: [u8; 4] = b
+        .as_ref()
+        .try_into()
+        .map_err(|_| Error::InvalidU32Bytes(b.to_vec()))?;
+    let val = u32::from_be_bytes(array);
+    Ok(val)
 }
 
 impl BatchDatabase for Tree {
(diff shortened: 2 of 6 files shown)


notmandatory:
I think your code changes to remove flush from the Database trait and to refactor duplicate code into the `ivec_to_u32` function look good. 

But I'm having second thoughts on doing the auto-flush. I believe it's not meant to be used on every update since it will hurt performance and on most os/filesystems it is not needed since the os will automatically flush the data to disk if the process dies. For Android users it's reasonable to recommend using sqlite and not sled for now. In future bdk api changes that are currently being discussed the database will most likely be extracted from the `Wallet` and then if a user wants to use Sled on android they will be able directly call the `.flush()` function on Tree when they want such as when their app goes off screen or something like that. I'm going to update #567  to remove the part about auto flushing so you can remove the related code and tests from this PR. 

You also need to do a `cargo fmt` which CI checks for. Thanks!

kafaichoi:
> I think your code changes to remove flush from the Database trait and to refactor duplicate code into the `ivec_to_u32` function look good.
> 
> But I'm having second thoughts on doing the auto-flush. I believe it's not meant to be used on every update since it will hurt performance and on most os/filesystems it is not needed since the os will automatically flush the data to disk if the process dies. For Android users it's reasonable to recommend using sqlite and not sled for now. In future bdk api changes that are currently being discussed the database will most likely be extracted from the `Wallet` and then if a user wants to use Sled on android they will be able directly call the `.flush()` function on Tree when they want such as when their app goes off screen or something like that. I'm going to update #567 to remove the part about auto flushing so you can remove the related code and tests from this PR.
> 
> You also need to do a `cargo fmt` which CI checks for. Thanks!

That's very reasonable. Thank you so much for your comment. I have removed the auto-flush behaviour(and ran cargo fmt, clippy for each commit)

src/database/keyvalue.rs
```diff
@@ -166,16 +166,9 @@ macro_rules! impl_batch_operations {
...
-                None => Ok(None),
-                Some(b) => {
-                    let array: [u8; 4] = b.as_ref().try_into().map_err(|_| Error::InvalidU32Bytes(b.to_vec()))?;
-                    let val = u32::from_be_bytes(array);
-                    Ok(Some(val))
-                }
-            }
+            $process_delete!(res)
+            .map(ivec_to_u32)
+            .transpose()
```
notmandatory, line 171: I just noticed that there wasn't a `.transpose()` in the original code, did you mean to add it here?

kafaichoi: Yes. I believed that's the correct behaviour. it just replace the match expression.
`None ->Ok(None)`, `Some(Ok(_))` -> `Ok(Some(_))` and `Ok(Err(_))` -> `Err(_)`
https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.transpose

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ To decouple the `Wallet` from the `Blockchain` we've made major changes:
 - Stop making a request for the block height when calling `Wallet:new`.
 - Added `SyncOptions` to capture extra (future) arguments to `Wallet::sync`.
 - Removed `max_addresses` sync parameter which determined how many addresses to cache before syncing since this can just be done with `ensure_addresses_cached`.
+- remove `flush` method from the `Database` trait.
```
rajarshimaitra, line 55: This should be in `Unreleased` section?

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ To decouple the `Wallet` from the `Blockchain` we've made major changes:
...
+
```
rajarshimaitra, line 56: No need for the extra space after a changelog line..

CHANGELOG.md
```diff
@@ -437,4 +439,4 @@ final transaction is created by calling `finish` on the builder.
 [v0.16.0]: https://github.com/bitcoindevkit/bdk/compare/v0.15.0...v0.16.0
 [v0.16.1]: https://github.com/bitcoindevkit/bdk/compare/v0.16.0...v0.16.1
 [v0.17.0]: https://github.com/bitcoindevkit/bdk/compare/v0.16.1...v0.17.0
-[unreleased]: https://github.com/bitcoindevkit/bdk/compare/v0.17.0...HEAD
\ No newline at end of file
+[unreleased]: https://github.com/bitcoindevkit/bdk/compare/v0.17.0...HEAD
```
rajarshimaitra: I am seeing an extra blank line beneath this.. But its not showing in the change set in github.. So have a double check..

rajarshimaitra (commented):
tACK https://github.com/bitcoindevkit/bdk/pull/575/commits/6dc928216b542bb5721800ac24a78c079ba80955

With few nits..

rajarshimaitra:
@notmandatory so we don't need the explicit `flush` any more for any of the DBs?

notmandatory:
> @notmandatory so we don't need the explicit `flush` any more for any of the DBs?

The only database that actually implemented flush was Sled, and even for Sled it seems to only be needed for Android. That's why I suggested removing it and that mobile apps use sqlite instead. This https://github.com/bitcoindevkit/bdk/pull/409 is the PR that added it, and it was to trying to fix the issues that @thunderbiscuit ran into on Android.

afilini:
Maybe we should deprecate for one release before removing it? I don't know if anybody is using it, but it seems it wouldn't cost us anything to just delay the removal and it can make the transition a bit less painful for our users.

Concept ACK for the refactor but I'll take another look later to make sure the code is also good.

src/database/keyvalue.rs
notmandatory: Thanks for the explanation I love learning cool Rust like this!

notmandatory:
> Maybe we should deprecate for one release before removing it? I don't know if anybody is using it, but it seems it wouldn't cost us anything to just delay the removal and it can make the transition a bit less painful for our users.

Good point, I was assuming no one was using this but I think we did promise to deprecate APIs before removing them. If we do deprecate that should go in a new PR, so this one can be ready to go right after we cut the next release branch.

CHANGELOG.md
kafaichoi: My bad. mis-placement after rebase

CHANGELOG.md
kafaichoi: good catch. it looks like it's caused by I edited this line using vim which add newline at the end of each line. I have to remove the last line and use `echo -n` to avoid making change 😅

rajarshimaitra (approved):
ReACK 0e713f89bfc11bb91b22b788068461e141ae90a1

notmandatory:
I've assigned this to the release `0.19.0` and added #577 to deprecate `Database::flush()` in `0.18.0`. This PR can then be merged after we create the `release/0.19.0` feature freeze branch next week.

notmandatory:

notmandatory:
I had to push this PR to the next release so the team can focus on #593, and after that one you'll probably need to rebase again. But then I promise we get this one in. :-)

danielabrozzoni (approved):
tACK b1ace3c9fd439573ddaee01d075d95a673115f93

rajarshimaitra (approved):
tACK b1ace3c9fd439573ddaee01d075d95a673115f93

This looks ready for merge to me..

danielabrozzoni:
This PR needs rebase to fix the CHANGELOG.md conflicts and fix the CI, but then it's ready to merge 🙏🏻

danielabrozzoni:
re-ACK 5ff8320e3b1cb4e3971549a3e3c168f20762d04b

notmandatory:
Finally merged, thanks for seeing this one through!

PR #579, opened 2022-04-05
Faster sync by collecting esplora ureq thread handles

nickfarrow:
### Description
Speeds up esplora ureq syncing. Taken from https://github.com/bitcoindevkit/bdk/pull/560

The current sync just creates a map of scripts to joinhandles which doesn't yet spawn the sync threads due to lazy evaluation. In the following `handles.map()`, the thread handles are *sequentially* evaluated and joined. With the fix, the handles are collected so that the threads spawn in parallel, and then joined

### Notes to the reviewers

I had to add a `#[allow(clippy::needless_collect)]` so that it wouldn't complain about collecting and then iterating. (Perhaps clippy is partially responsible for this issue!)

Tested sync performance by doing a fresh sync on an existing [gun](https://gun.fun) wallet.
```
---- Before fix ---
real	0m13.121s
real	0m13.367s
real	0m13.211s

---- After fix ----
real	0m5.516s
real	0m5.251s
real	0m5.594s
```

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/blockchain/esplora/ureq.rs
+++ b/src/blockchain/esplora/ureq.rs
@@ -127,10 +127,11 @@ impl WalletSync for EsploraBlockchain {
                         .take(self.concurrency as usize)
                         .cloned();
 
-                    let handles = scripts.map(move |script| {
+                    let mut handles = vec![];
+                    for script in scripts {
                         let client = self.url_client.clone();
                         // make each request in its own thread.
-                        std::thread::spawn(move || {
+                        handles.push(std::thread::spawn(move || {
                             let mut related_txs: Vec<Tx> = client._scripthash_txs(&script, None)?;
 
                             let n_confirmed =
@@ -152,10 +153,11 @@ impl WalletSync for EsploraBlockchain {
                                 }
                             }
                             Result::<_, Error>::Ok(related_txs)
-                        })
-                    });
+                        }));
+                    }
 
                     let txs_per_script: Vec<Vec<Tx>> = handles
+                        .into_iter()
                         .map(|handle| handle.join().unwrap())
                         .collect::<Result<_, _>>()?;
                     let mut satisfaction = vec![];


notmandatory:
Now that you're fixing the parallel execution should we worry about running out of some resource if a wallet has lots of scripts and transactions to sync? seems like this could swamp a low power esplora server or at least exceed some point of diminishing returns for spawning new threads. You don't need to address it in this PR but but do you have any thoughts for how to address this issue? maybe some sort of thread pool?

LLFourn:
> Now that you're fixing the parallel execution should we worry about running out of some resource if a wallet has lots of scripts and transactions to sync? seems like this could swamp a low power esplora server or at least exceed some point of diminishing returns for spawning new threads. You don't need to address it in this PR but but do you have any thoughts for how to address this issue? maybe some sort of thread pool?

Note that the number of requests is limited. It does them in batches according to `concurrency`.

src/blockchain/esplora/ureq.rs
```diff
@@ -127,35 +127,40 @@ impl WalletSync for EsploraBlockchain {
...
-                                    let new_related_txs: Vec<Tx> = client._scripthash_txs(
-                                        &script,
-                                        Some(related_txs.last().unwrap().txid),
-                                    )?;
-                                    let n = new_related_txs.len();
-                                    related_txs.extend(new_related_txs);
-                                    // we've reached the end
-                                    if n < 25 {
-                                        break;
+                    #[allow(clippy::needless_collect)]
```
LLFourn: :bulb: To avoid this you could write a for loop and manually push the threads into a vec. It might not be idiomatic to spawn threads in an iterator. Not a blocker for me. It's probably worth a short comment as to why the ignore here is a good idea otherwise.

nickfarrow: good idea, much simpler

notmandatory (approved):
ACK adef166b22d5c95006df39fd8c57574624071189

Issue #561, opened 2022-03-09
Write own BIP39 implementation


jblachly:
As suggested in the code base:

https://github.com/bitcoindevkit/bdk/blob/adf7d0c126aab471d6c52cb3129c065c5c2a2063/src/keys/bip39.rs#L14-L17

I also recommend writing own BIP39 implementation.

In issue #399, BDK transitioned from `tiny-bip39` to `rust-bip39`, however, the latter package (which BDK now depends on) appears to be unmaintained or poorly maintained, with PRs and issues ignored for about a year.

https://github.com/rust-bitcoin/rust-bip39

At a minimum, BIP39 wordlists, and words -> entropy -> seed generation would be valuable.

notmandatory:
@jblachly I think it's too late to add this as a 2022 summer of bitcoin project, but I'll tag it as a good first issue for someone new.

jblachly:
Thanks; accidentally started with that template (there was no other feature request template) and didn't realize it would add an issue tag.

notmandatory:
No problem, it's on my todo list to add an enhancement issue template. 😅

violet360:
Hey, guys !!
I am a bitcoin enthusiast and an SoB participant who recently started with RUST and new to this project as well, I was looking for cool rust issues to work on so that I can improve my understanding and this one really caught my attention, I would love to work on this project.

Eunoia1729:
Hi @violet360! 
I've been working on this since past few days. Hope we can review/discuss if I'm going in the right direction. 
I can make a draft PR.

violet360:
Ok cool, for sure :)

evanlinjin:
@violet360 @Eunoia1729 How is progress on this issue so far?

danielabrozzoni:
@Eunoia1729 are you stll working on this? :)

Eunoia1729:
Sorry for the late response. 
I didnt progress much after that. Anyone interested can start with it, I'll be happy to join or help in any way.

vladimirfomene:
Hello @danielabrozzoni, has this been worked on? Can I pick it up?

danielabrozzoni:
Nobody that I know of, you can pick it up :)

atalw:
Hey, sorry that I didn't update here but I started working on this earlier [today](https://github.com/atalw/bdk/commits/bip-0039). Should I continue? Perhaps we could work together?

vladimirfomene:
[...]

PR #582, opened 2022-04-07
Expose bip39::Error

### LINKED ISSUE #581: bip39::Error is not exposed
**Describe the bug**  
Currently, there is no way to access the struct `bip39::Error`. It is useful to expose this structure so that one can create structs with `bip39::Error` type fields.

**Additional context**  
- [`bip39::Error`](https://github.com/rust-bitcoin/rust-bip39/blob/520e15a6e5b86940dae2d386d90b4b3616491d13/src/lib.rs#L99-L114)
- [Proposed location of where it should be exposed in BDK.](https://github.com/bitcoindevkit/bdk/blob/0621ca89d5f7f984b5c3281cd5b9de880342c54b/src/keys/bip39.rs#L22)

evanlinjin:
### Description

Expose `bip39::Error` (fixes #581 )

### Notes to the reviewers

I am aware that the `bip39` module plans to be rewritten (as per #561 ), however this seems like a rather straightforward and quick change that may be useful in the short/mid term.

#### Bugfixes:

* [x] Expose `bip39::Error`

### DIFF
--- a/src/keys/bip39.rs
+++ b/src/keys/bip39.rs
@@ -19,7 +19,7 @@ use bitcoin::Network;
 
 use miniscript::ScriptContext;
 
-pub use bip39::{Language, Mnemonic};
+pub use bip39::{Error, Language, Mnemonic};
 
 type Seed = [u8; 64];
 


afilini (approved):
ACK c752ccbddeaa5ae441cc74a1386c1e412a447104

PR #585, opened 2022-04-15
set coin type based on network 

### LINKED ISSUE #578: BIP44, BIP49, BIP84 templates don't set coin type
**Describe the bug**  

When using one of the BIP(44,49,84) templates the "coin type" part of the key derivation path is always set to 0, which represents bitcoin mainnet, no matter what the type of the key(s) used are.  These templates should use 1 if the type of the key is testnet, regtest, or signet.

(Originally reported on discord by marcosalejandro)

**To Reproduce**  
```rust
    // BIP44 `pkh(key/44'/{0,1}'/0'/{0,1}/*)`
    #[test]
    fn test_bip44_template_cointype() {
        use bitcoin::util::bip32::ChildNumber::Hardened;
        
        let xprvkey = bitcoin::util::bip32::ExtendedPrivKey::from_str("xprv9s21ZrQH143K2fpbqApQL69a4oKdGVnVN52R82Ft7d1pSqgKmajF62acJo3aMszZb6qQ22QsVECSFxvf9uyxFUvFYQMq3QbtwtRSMjLAhMf").unwrap();
        println!("bitcoin prvkey: {}", xprvkey.network);
        assert_eq!(Network::Bitcoin, xprvkey.network);
        let xdesc = Bip44(xprvkey, KeychainKind::Internal).build().unwrap();
        println!("xdesc = {:?}", xdesc);

        if let ExtendedDescriptor::Pkh(pkh) = xdesc.0 {
            let path : Vec<ChildNumber> = pkh.into_inner().full_derivation_path().into();
            let purpose = path.get(0).unwrap();
            println!("xdesc purpose = {:?}", purpose);
            assert!(matches!(purpose, Hardened { index: 44 }));
            let coin_type = path.get(1).unwrap();
            println!("xdesc coin_type = {:?}", coin_type);
            assert!(matches!(coin_type, Hardened { index: 0 }));
        }
        
        let tprvkey = bitcoin::util::bip32::ExtendedPrivKey::from_str("tprv8ZgxMBicQKsPcx5nBGsR63Pe8KnRUqmbJNENAfGftF3yuXoMMoVJJcYeUw5eVkm9WBPjWYt6HMWYJNesB5HaNVBaFc1M6dRjWSYnmewUMYy").unwrap();
        println!("testnet prvkey: {}", tprvkey.network);
        assert_eq!(Network::Testnet, tprvkey.network);
        let tdesc = Bip44(tprvkey, KeychainKind::Internal).build().unwrap();
        println!("tdesc = {:?}", tdesc);
        
        if let ExtendedDescriptor::Pkh(pkh) = tdesc.0 {
            let path : Vec<ChildNumber> = pkh.into_inner().full_derivation_path().into();
            let purpose = path.get(0).unwrap();
            println!("tdesc purpose = {:?}", purpose);
            assert!(matches!(purpose, Hardened { index: 44 }));
            let coin_type = path.get(1).unwrap();
            println!("tdesc coin_type = {:?}", coin_type);
            assert!(matches!(coin_type, Hardened { index: 1 }));
        }
    }
```
test output:
```shell
bitcoin prvkey: bitcoin
xdesc = (pkh(XPub(DescriptorXKey { origin: Some((65af6f2e, m/44'/0'/0')), xkey: ExtendedPubKey { network: Bitcoin, depth: 3, parent_fingerprint: 714b621c, child_number: Hardened { index: 0 }, public_key: PublicKey { compressed: true, key: PublicKey(c3ce4936cde277e08e4604bc3381520747dcdec0d8b3ee4b5e5428436a385a5a47456d0b8d75187bb488ce073dd4759a1738db9e356a5bdaaaaf5ae459f9ebfe) }, chain_code: f9fd1286c60681930bbc5e74c2769ac2898a3564502ebe5039905e010e2b5787 }, derivation_path: m/1, wildcard: Unhardened })), {XPub(DescriptorXKey { origin: Some((65af6f2e, m/44'/0'/0')), xkey: ExtendedPubKey { network: Bitcoin, depth: 3, parent_fingerprint: 714b621c, child_number: Hardened { index: 0 }, public_key: PublicKey { compressed: true, key: PublicKey(c3ce4936cde277e08e4604bc3381520747dcdec0d8b3ee4b5e5428436a385a5a47456d0b8d75187bb488ce073dd4759a1738db9e356a5bdaaaaf5ae459f9ebfe) }, chain_code: f9fd1286c60681930bbc5e74c2769ac2898a3564502ebe5039905e010e2b5787 }, derivation_path: m/1, wildcard: Unhardened }): XPrv(DescriptorXKey { origin: None, xkey: ExtendedPrivKey { network: Bitcoin, depth: 0, parent_fingerprint: 00000000, child_number: Normal { index: 0 }, private_key: [private key data], chain_code: 3d6a59ac46bf8fd55349efa02581a8f1439dea8a62fbfe4de3bdf7c4ba629578 }, derivation_path: m/44'/0'/0'/1, wildcard: Unhardened })}, {Bitcoin})
xdesc purpose = Hardened { index: 44 }
xdesc coin_type = Hardened { index: 0 }
testnet prvkey: testnet
tdesc = (pkh(XPub(DescriptorXKey { origin: Some((34b00776, m/44'/0'/0')), xkey: ExtendedPubKey { network: Testnet, depth: 3, parent_fingerprint: 1c00d808, child_number: Hardened { index: 0 }, public_key: PublicKey { compressed: true, key: PublicKey(f5f3776c7305f9adac93dc3b841d3018c54dd34c2e12cf45f6192daa679a99971c73da5604f0cc2a108a542095eef7e60587ca62d05d852f771bd5c2f8ad22db) }, chain_code: d7543f5ecbe9620f3249206b2fb085f4a905a8eb20f0d12c85407b5cd09ef515 }, derivation_path: m/1, wildcard: Unhardened })), {XPub(DescriptorXKey { origin: Some((34b00776, m/44'/0'/0')), xkey: ExtendedPubKey { network: Testnet, depth: 3, parent_fingerprint: 1c00d808, child_number: Hardened { index: 0 }, public_key: PublicKey { compressed: true, key: PublicKey(f5f3776c7305f9adac93dc3b841d3018c54dd34c2e12cf45f6192daa679a99971c73da5604f0cc2a108a542095eef7e60587ca62d05d852f771bd5c2f8ad22db) }, chain_code: d7543f5ecbe9620f3249206b2fb085f4a905a8eb20f0d12c85407b5cd09ef515 }, derivation_path: m/1, wildcard: Unhardened }): XPrv(DescriptorXKey { origin: None, xkey: ExtendedPrivKey { network: Testnet, depth: 0, parent_fingerprint: 00000000, child_number: Normal { index: 0 }, private_key: [private key data], chain_code: 07c7c6226c3cf2cd58bb416e0e124fa396a1559c95a3362c603c728c2c9de4f6 }, derivation_path: m/44'/0'/0'/1, wildcard: Unhardened })}, {Regtest, Signet, Testnet})
tdesc purpose = Hardened { index: 44 }
tdesc coin_type = Hardened { index: 0 }

assertion failed: matches!(coin_type, Hardened { index : 1 })
```

**Expected behavior**  

According to [BIP44 Coin Type] for `Testnet` the `ChildNumber` at index 1 should be `Hardened { index: 1 }`. 

**Build environment**  
 - BDK tag/commit: v0.17.0
 - OS+version: any
 - Rust/Cargo version: any
 - Rust/Cargo target: any

**Additional context**  

The Bip44, Bip49, and Bip84 macros all use the `descriptor::template::expand_make_bipxx` macro and this is where the coin type part of the derivation path is set.

[BIP44 Coin Type]:https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#Coin_type

jbesraa:
resolves #578 

### DIFF
--- a/src/descriptor/template.rs
+++ b/src/descriptor/template.rs
@@ -40,18 +40,19 @@ pub type DescriptorTemplateOut = (ExtendedDescriptor, KeyMap, ValidNetworks);
 /// use bdk::keys::{IntoDescriptorKey, KeyError};
 /// use bdk::miniscript::Legacy;
 /// use bdk::template::{DescriptorTemplate, DescriptorTemplateOut};
+/// use bitcoin::Network;
 ///
 /// struct MyP2PKH<K: IntoDescriptorKey<Legacy>>(K);
 ///
 /// impl<K: IntoDescriptorKey<Legacy>> DescriptorTemplate for MyP2PKH<K> {
-///     fn build(self) -> Result<DescriptorTemplateOut, DescriptorError> {
+///     fn build(self, network: Network) -> Result<DescriptorTemplateOut, DescriptorError> {
 ///         Ok(bdk::descriptor!(pkh(self.0))?)
 ///     }
 /// }
 /// ```
 pub trait DescriptorTemplate {
     /// Build the complete descriptor
-    fn build(self) -> Result<DescriptorTemplateOut, DescriptorError>;
+    fn build(self, network: Network) -> Result<DescriptorTemplateOut, DescriptorError>;
 }
 
 /// Turns a [`DescriptorTemplate`] into a valid wallet descriptor by calling its
@@ -62,7 +63,7 @@ impl<T: DescriptorTemplate> IntoWalletDescriptor for T {
         secp: &SecpCtx,
         network: Network,
     ) -> Result<(ExtendedDescriptor, KeyMap), DescriptorError> {
-        self.build()?.into_wallet_descriptor(secp, network)
+        self.build(network)?.into_wallet_descriptor(secp, network)
     }
 }
 
@@ -95,7 +96,7 @@ impl<T: DescriptorTemplate> IntoWalletDescriptor for T {
 pub struct P2Pkh<K: IntoDescriptorKey<Legacy>>(pub K);
 
 impl<K: IntoDescriptorKey<Legacy>> DescriptorTemplate for P2Pkh<K> {
-    fn build(self) -> Result<DescriptorTemplateOut, DescriptorError> {
+    fn build(self, _network: Network) -> Result<DescriptorTemplateOut, DescriptorError> {
         descriptor!(pkh(self.0))
     }
 }
@@ -130,7 +131,7 @@ impl<K: IntoDescriptorKey<Legacy>> DescriptorTemplate for P2Pkh<K> {
 pub struct P2Wpkh_P2Sh<K: IntoDescriptorKey<Segwitv0>>(pub K);
 
 impl<K: IntoDescriptorKey<Segwitv0>> DescriptorTemplate for P2Wpkh_P2Sh<K> {
-    fn build(self) -> Result<DescriptorTemplateOut, DescriptorError> {
+    fn build(self, _network: Network) -> Result<DescriptorTemplateOut, DescriptorError> {
         descriptor!(sh(wpkh(self.0)))
     }
 }
@@ -164,12 +165,12 @@ impl<K: IntoDescriptorKey<Segwitv0>> DescriptorTemplate for P2Wpkh_P2Sh<K> {
 pub struct P2Wpkh<K: IntoDescriptorKey<Segwitv0>>(pub K);
 
 impl<K: IntoDescriptorKey<Segwitv0>> DescriptorTemplate for P2Wpkh<K> {
-    fn build(self) -> Result<DescriptorTemplateOut, DescriptorError> {
+    fn build(self, _network: Network) -> Result<DescriptorTemplateOut, DescriptorError> {
         descriptor!(wpkh(self.0))
     }
 }
 
-/// BIP44 template. Expands to `pkh(key/44'/0'/0'/{0,1}/*)`
+/// BIP44 template. Expands to `pkh(key/44'/{0,1}'/0'/{0,1}/*)`
 ///
 /// Since there are hardened derivation steps, this template requires a private derivable key (generally a `xprv`/`tprv`).
 ///
@@ -193,21 +194,21 @@ impl<K: IntoDescriptorKey<Segwitv0>> DescriptorTemplate for P2Wpkh<K> {
 ///     MemoryDatabase::default()
 /// )?;
 ///
-/// assert_eq!(wallet.get_address(New)?.to_string(), "miNG7dJTzJqNbFS19svRdTCisC65dsubtR");
-/// assert_eq!(wallet.public_descriptor(KeychainKind::External)?.unwrap().to_string(), "pkh([c55b303f/44'/0'/0']tpubDDDzQ31JkZB7VxUr9bjvBivDdqoFLrDPyLWtLapArAi51ftfmCb2DPxwLQzX65iNcXz1DGaVvyvo6JQ6rTU73r2gqdEo8uov9QKRb7nKCSU/0/*)#xgaaevjx");
+/// assert_eq!(wallet.get_address(New)?.to_string(), "mmogjc7HJEZkrLqyQYqJmxUqFaC7i4uf89");
+/// assert_eq!(wallet.public_descriptor(KeychainKind::External)?.unwrap().to_string(), "pkh([c55b303f/44'/1'/0']tpubDCuorCpzvYS2LCD75BR46KHE8GdDeg1wsAgNZeNr6DaB5gQK1o14uErKwKLuFmeemkQ6N2m3rNgvctdJLyr7nwu2yia7413Hhg8WWE44cgT/0/*)#5wrnv0xt");
 /// # Ok::<_, Box<dyn std::error::Error>>(())
 /// ```
 pub struct Bip44<K: DerivableKey<Legacy>>(pub K, pub KeychainKind);
 
 impl<K: DerivableKey<Legacy>> DescriptorTemplate for Bip44<K> {
-    fn build(self) -> Result<DescriptorTemplateOut, DescriptorError> {
-        P2Pkh(legacy::make_bipxx_private(44, self.0, self.1)?).build()
+    fn build(self, network: Network) -> Result<DescriptorTemplateOut, DescriptorError> {
+        P2Pkh(legacy::make_bipxx_private(44, self.0, self.1, network)?).build(network)
     }
 }
 
 /// BIP44 public template. Expands to `pkh(key/{0,1}/*)`
 ///
-/// This assumes that the key used has already been derived with `m/44'/0'/0'`.
+/// This assumes that the key used has already been derived with `m/44'/0'/0'` for Mainnet or `m/44'/1'/0'` for Testnet.
 ///
 /// This template requires the parent fingerprint to populate correctly the metadata of PSBTs.
 ///
@@ -240,12 +241,12 @@ impl<K: DerivableKey<Legacy>> DescriptorTemplate for Bip44<K> {
 pub struct Bip44Public<K: DerivableKey<Legacy>>(pub K, pub bip32::Fingerprint, pub KeychainKind);
 
 impl<K: DerivableKey<Legacy>> DescriptorTemplate for Bip44Public<K> {
-    fn build(self) -> Result<DescriptorTemplateOut, DescriptorError> {
-        P2Pkh(legacy::make_bipxx_public(44, self.0, self.1, self.2)?).build()
+    fn build(self, network: Network) -> Result<DescriptorTemplateOut, DescriptorError> {
+        P2Pkh(legacy::make_bipxx_public(44, self.0, self.1, self.2)?).build(network)
     }
 }
 
-/// BIP49 template. Expands to `sh(wpkh(key/49'/0'/0'/{0,1}/*))`
+/// BIP49 template. Expands to `sh(wpkh(key/49'/{0,1}'/0'/{0,1}/*))`
 ///
 /// Since there are hardened derivation steps, this template requires a private derivable key (generally a `xprv`/`tprv`).
 ///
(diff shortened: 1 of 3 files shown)


afilini (changes requested):
On the surface this seems to work, but it's actually incomplete: there's nothing preventing a testnet bip44 wallet from being used on mainnet, or vice-versa.

My suggestion is to change the `DescriptorTemplate` trait to add a network argument to the `build()` method. That method is called when the descriptor template is passed to a wallet constructor, so we basically get the network from that. Which means the user only has to set the network once (in the wallet's constructor) and then everything magically works!

jbesraa:
hmm not sure I get that..
something like this?
```rust
pub trait DescriptorTemplate {
    /// Build the complete descriptor
    fn build(self, network: Network) -> Result<DescriptorTemplateOut, DescriptorError>;
}

Bip44(prvkey, KeychainKind::External).build(Network::Bitcoin)

pub struct Bip44<K: DerivableKey<Legacy>>(pub K, pub KeychainKind);

impl<K: DerivableKey<Legacy>> DescriptorTemplate for Bip44<K> {
    fn build(self, network: Network) -> Result<DescriptorTemplateOut, DescriptorError> {
        P2Pkh(legacy::make_bipxx_private(44, self.0, self.1, network)?).build(network)
    }
}
```

afilini:
Yes, exactly, something like that.

This will prevent you from having conflicts, i.e. specifying network `x` in the template constructor, but then trying to use it on a wallet for network `y`

jbesraa:
i see what u mean. 
we end up with a couple of unused `network` inputs
https://gist.github.com/jbesraa/1e807a730e45746371b2d1dbf65d47fd#file-bdk-desc-template-diff-rs-L26-L49
look for `_network`

afilini:
Yeah I think that's fine, some templates will use the network and some won't

src/descriptor/template.rs
```diff
@@ -197,11 +197,11 @@ impl<K: IntoDescriptorKey<Segwitv0>> DescriptorTemplate for P2Wpkh<K> {
 /// assert_eq!(wallet.public_descriptor(KeychainKind::External)?.unwrap().to_string(), "pkh([c55b303f/44'/0'/0']tpubDDDzQ31JkZB7VxUr9bjvBivDdqoFLrDPyLWtLapArAi51ftfmCb2DPxwLQzX65iNcXz1DGaVvyvo6JQ6rTU73r2gqdEo8uov9QKRb7nKCSU/0/*)#xgaaevjx");
```
rajarshimaitra: Doc needs to be updated with current template args (network field missing).

src/descriptor/template.rs
```diff
@@ -273,11 +273,11 @@ impl<K: DerivableKey<Legacy>> DescriptorTemplate for Bip44Public<K> {
 /// assert_eq!(wallet.public_descriptor(KeychainKind::External)?.unwrap().to_string(), "sh(wpkh([c55b303f/49\'/0\'/0\']tpubDC49r947KGK52X5rBWS4BLs5m9SRY3pYHnvRrm7HcybZ3BfdEsGFyzCMzayi1u58eT82ZeyFZwH7DD6Q83E3fM9CpfMtmnTygnLfP59jL9L/0/*))#gsmdv4xr");
```
rajarshimaitra: Re: doc update

src/descriptor/template.rs
```diff
@@ -466,6 +472,40 @@ mod test {
     use miniscript::descriptor::{DescriptorPublicKey, DescriptorTrait, KeyMap};
     use miniscript::Descriptor;
 
+    // BIP44 `pkh(key/44'/{0,1}'/0'/{0,1}/*)`
+    #[test]
+    fn test_bip44_template_cointype() {
```
rajarshimaitra, line 481: I feel like this test can be generalized to include cointype checking for all the templates 44, 49 and 84. Can be done in a single test. and can be named `test_template_cointype()`.

rajarshimaitra (changes requested):
Concept ACK.. Thanks for working on the bug..

But I feel its still incomplete. The public templates needs to be updated too.. And the docs also needs update.. Not sure if there's some macro problem doing the fix in public part, IIUC the same bug should be active on public templates too..

jbesraa:
@rajarshimaitra what do u think about the solution mentioned in prev message(gist)?

edit:
https://gist.github.com/jbesraa/1e807a730e45746371b2d1dbf65d47fd#file-bdk-desc-template-diff-rs

rajarshimaitra:
I looked at the diff.. But couldn't apply it.. Probably its on top of a commit not in the tree anymore.. Anyway if the issue is only having to pass unused args I think thats a better next step.. So probably worth to put another commit with the changes.. 

But something feels off to me. IIUC the network information should be needed for for all kind of keys.. So ideally we should only have one source of it and disperse it evenly through all the macros..  

Like how would this macro know which network to derive the key for??
```
pub struct P2Pkh<K: IntoDescriptorKey<Legacy>>(pub K);

impl<K: IntoDescriptorKey<Legacy>> DescriptorTemplate for P2Pkh<K> {
    fn build(self) -> Result<DescriptorTemplateOut, DescriptorError> {
        descriptor!(pkh(self.0))
    }
}
```
P2PKH mainnet and P2PKH testnet should have different coin types too..

So I think somewhere deeper a fix needs to be made so that we can disperse the network information evenly?? @afilini let me know if I am missing something here.

afilini:
> Like how would this macro know which network to derive the key for??

This is all handled by the `descriptor!()` macro, so you don't have to worry about it at this level.

If you look at the `DescriptorTemplateOut` type you'll see that it contains a `ValidNetworks` set. This is the intersection of all the valid network sets for all the keys. For example, if I have a key that's valid everywhere (let's say, a bip39 mnemonic) and a `xprv` the final valid network is the intersection of <everything> and `{ mainnet }`, which is just mainnet.

If you then try to load that descriptor in a testnet wallet, you'll get an error.

> P2PKH mainnet and P2PKH testnet should have different coin types too..

I think you are confusing the address encoding, which is different, with the script, which is the same. The base58 encoding uses different prefixes for mainnet/testnet, so you get a `1Something` address in mainnet and a `m/nsomething` on testnet. But this encoding only happens when you want to get a address string for a given script (which is done inside the wallet that knows the network). Here you are just defining the script, and that's the same on both chains

rajarshimaitra:
> If you look at the DescriptorTemplateOut type you'll see that it contains a ValidNetworks set.

Ok I see.. So the underlying macros already knows how to transmit back valid network set..

> I think you are confusing the address encoding, which is different, with the script, which is the same

Yes I was confused here. The `P2PKH` template will already have all the valid network set it needs.. 

I guess then the only work remaining here is to make `BipXXPublic` also set the coin type flag with network?? And will only need the same modification in the `make_bipxx_public` macro function.. And then a test to seal the behavior for all the bips..

afilini:
I don't think you can do it for the `BipXXPublic` structs because the coin type is part of the hardened derivation steps, so it has to be applied first. The `-Public` structs are meant to be used when you only know the pubkey.

We will have to update the documentation though, for example for bip44 it says:

> This assumes that the key used has already been derived with m/44'/0'/0'

This should be updated and contain a note that on testnet the path should be X and on mainnet Y.

notmandatory:

rajarshimaitra:
@jbesraa can you please simplify the commit history by doing `git rebase master` from your local branch?? The history has become confusing.. And don't know why `git log` is showing me different history than github.. 

In general anytime you wanna rebase your PR just do `git rebase master` and fix any conflict that occurs in rebase.. And not commit master into your branch.. That makes review much simpler..

afilini:
+1 for the rebase, plus i would just squash everything into a single commit for the same reason Raj mentioned

I'm also starting the CI so we can see if something is wrong there

danielabrozzoni (commented):
Concept ACK, can you please squash the commits in a single one? You can find a guide here: https://www.git-tower.com/learn/git/faq/git-squash

src/descriptor/template.rs
danielabrozzoni: I agree :)

jbesraa: sure, I can take care of that

src/descriptor/template.rs
```diff
@@ -466,6 +476,42 @@ mod test {
...
+        use bitcoin::util::bip32::ChildNumber::{self, Hardened};
+
+        let xprvkey = bitcoin::util::bip32::ExtendedPrivKey::from_str("xprv9s21ZrQH143K2fpbqApQL69a4oKdGVnVN52R82Ft7d1pSqgKmajF62acJo3aMszZb6qQ22QsVECSFxvf9uyxFUvFYQMq3QbtwtRSMjLAhMf").unwrap();
+        println!("bitcoin prvkey: {}", xprvkey.network);
```
rajarshimaitra: Better to avoid `println!`s in test, they clutter the test outputs unnecessarily..

src/descriptor/template.rs
```diff
@@ -466,6 +476,42 @@ mod test {
...
+        if let ExtendedDescriptor::Pkh(pkh) = xdesc.0 {
+            let path: Vec<ChildNumber> = pkh.into_inner().full_derivation_path().into();
+            let purpose = path.get(0).unwrap();
+            assert!(matches!(purpose, Hardened { index: 44 }));
+            let coin_type = path.get(1).unwrap();
+            assert!(matches!(coin_type, Hardened { index: 0 }));
+        }
+
+        let tprvkey = bitcoin::util::bip32::ExtendedPrivKey::from_str("tprv8ZgxMBicQKsPcx5nBGsR63Pe8KnRUqmbJNENAfGftF3yuXoMMoVJJcYeUw5eVkm9WBPjWYt6HMWYJNesB5HaNVBaFc1M6dRjWSYnmewUMYy").unwrap();
+        println!("testnet prvkey: {}", tprvkey.network);
```
rajarshimaitra: Same as above..

rajarshimaitra (changes requested):
tACK a8b0058cb4dd818264f5a9aa06ee9168b7c4ffb7

modulo https://github.com/bitcoindevkit/bdk/pull/585/files#r857772454

Code changes looks good to me.. Few small nits..

danielabrozzoni:
This needs rebase to pick up the CI changes :)

rajarshimaitra (approved):
tACK dd1438896de8f4d63d785764c31ec7345b0c0993

danielabrozzoni:
@jbesraa I'd love to see https://github.com/bitcoindevkit/bdk/pull/585/files#r857772454 fixed before merging :)

afilini (approved):
ACK dd1438896de8f4d63d785764c31ec7345b0c0993

There's still a comment pending about an improvement to get better test coverage, but the code freeze for the release is tomorrow and we'd like to have this merged, so we will go ahead and fix the test later on.

afilini:
Ah damn, I can't merge this because the commits aren't signed. Could you please fix this @jbesraa ?

notmandatory:
@jbesraa if you need any help setting up signing with github let me know or check out this link: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

afilini (approved):
re-ACK db9d43ed2f4146dbfdc28a72d0c9ae420f831fc3

danielabrozzoni:
re-ACK db9d43ed2f4146dbfdc28a72d0c9ae420f831fc3

Issue #485, opened 2021-11-30
BDK Descriptor Improvements: Add a generic descriptor trait


rajarshimaitra:
**Description**  
  
In order for BDK to extend beyond just a Bitcoin wallet, it needs a more powerful descriptor capability. This project aims at improving the descriptors used in the library. Instead of having a concrete descriptor type, design a descriptor trait that can be implemented on any generic structure. This will allow it to have more powerful descriptors, which will help in more advanced applications of BDK, as well as the Bitcoin wallet descriptor ecosystem further, like having a “Lightning Descriptor Template”, etc. 

**Expected Outcomes**  
 

- Through understanding of [descriptors](https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md).
- Basic understanding of the [rust-miniscript](https://github.com/rust-bitcoin/rust-miniscript) library.
- A descriptor trait in BDK library, that can be implemented for any generic structures. 

**Resources**  
  
  
  
  
  

**Skills Required**  
  
  
  
  

- Experience with git. [Guide](https://git-scm.com/book/en/v2)
- Basic familiarity with rust. [First seven chapters of the book](https://doc.rust-lang.org/book/)
- Familiarity with BDK’s descriptor usage.
- Familiarity with rust traits.

**Mentor(s)**  

**Difficulty** : Hard 

**Competency Test (optional)**  

- Install rust, compile and run all bdk examples and tests. 
- Read through the BDK docs.
- Make a dummy wallet with BDK with different blockchain backends. 
- Familiarity with basic rust, should be able to write basic custom trait implementations on foreign structures.

LLFourn:
What is a "Lightning Descriptor Template"?

rajarshimaitra:
> I don't think we actually store the descriptor anywhere?

@afilini would you be able to elaborate on this?

rajarshimaitra:
> I don't think we actually store the descriptor anywhere?

@afilini would you be able to elaborate on this?

afilini:
The problem is that afaik the BOLT 3 scripts can't be expressed using descriptors, or rather they can, but when encoded they produce different bitcoin scripts which makes them incompatible with the current lightning protocol.

[...]

PR #593, opened 2022-04-26
Add support for Taproot and `tr()` descriptors

### LINKED ISSUE #63: Taproot / Schnorr

afilini:
### Description

This is a work-in-progress PR to update BDK to rust-bitcoin `0.28` which introduces taproot support and a few other improvements. While updating we also introduce taproot support in BDK.

High level list of subtasks for this PR:
- [x] Update rust-bitcoin and rust-miniscript
- [x] Stop using deprecated structs
- [x] Add taproot metadata to psbts
- [x] Produce schnorr signatures
- [x] Finalize taproot txs
- [x] Support `tr()` descriptors in the `descriptor!()` macro
- [x] Write a lot of tests
  - [x] Interoperability with other wallets (Core + ?)
     - [x] Signing/finalizing a psbt made by core
     - [x] Producing a psbt that core can sign and finalize
  - [x] Creating psbts
    - [x] Verify the metadata are correct
    - [x] Verify sighashes are applied correctly
    - [x] Create a tx with a foreign taproot and non-taproot utxo
  - [x] Signing psbts
    - [x] Signing for a key spend
    - [x] Signing for a script spend
    - [x] Signing with a single (wif) key
    - [x] Signing with an xprv (with and without knowing the utxo being spent in the db)
    - [x] Signing with weird sighashes
  - [x] Policy module
    - [x] Simple key spend
    - [x] More complex tap tree with a few keys
    - [x] Verify both `contribution` and `satisfaction` of a PSBT input
  - [x] Wallet module
    - [x] Generate addresses

Fixes #63 

#### Milestone

I'm adding this to the `0.19` milestone because now that rust-bitcoin and rust-miniscript have been released we should not waiting too long to release a version of BDK that supports the new libraries.

#### API Breaks

Since this is an API-break because of the new version of rust-bitcoin and rust-miniscript, I'm also taking the chance to update a few things in our lib that I had been thinking about for a while.

One example is the signer interface, which had that weird `sign_whole_tx()` method. This has now been removed, and the `Signer` trait replaced with `TransactionSigner` and `InputSigner`. I'm also starting to think that the signer should not only look at the psbt to figure out what to do, but ideally it should also receive some information about the descriptor (for example, the type) to simplify the code.

One option is to add an extra parameter, but that would probably only be used by our internal signers and not much else (for example, if you ask an hardware wallet to sign, it will probably already know what kind of wallet you have).

Another option is to wrap `PrivateKey` and `DescriptorXKey<ExtendedPrivKey>` which are the two internal signers we support with a struct that contains metadata about the descriptor, and then implement the signer traits on that struct. We could construct this in `Wallet::new()`, after miniscript parses the descriptor.

#### MSRV Bump

Due to the update of `rust-electrum-client`, which in turn depends on an updated `webpki`, we will have to bump our MSRV beacuse 1.46 is not supported by the new `webpki` version.

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add traits to reuse `Blockchain`s across multiple wallets (`BlockchainFactory` and `StatelessBlockchain`).
 - Upgrade to rust-bitcoin `0.28`
 - If using the `sqlite-db` feature all cached wallet data is deleted due to a possible UTXO inconsistency, a wallet.sync will recreate it  
+- Update `PkOrF` in the policy module to become an enum
 
 ## [v0.18.0] - [v0.17.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Upgrade to rust-bitcoin `0.28`
 - If using the `sqlite-db` feature all cached wallet data is deleted due to a possible UTXO inconsistency, a wallet.sync will recreate it  
 - Update `PkOrF` in the policy module to become an enum
+- Add experimental support for Taproot, including:
+  - Support for `tr()` descriptors with complex tapscript trees
+  - Creation of Taproot PSBTs (BIP-371)
+  - Signing Taproot PSBTs (key spend and script spend)
+  - Support for `tr()` descriptors in the `descriptor!()` macro
 
 ## [v0.18.0] - [v0.17.0]
 
--- a/src/descriptor/dsl.rs
+++ b/src/descriptor/dsl.rs
@@ -77,11 +77,14 @@ macro_rules! impl_top_level_pk {
 #[macro_export]
 macro_rules! impl_top_level_tr {
     ( $internal_key:expr, $tap_tree:expr ) => {{
-        use $crate::miniscript::descriptor::{Descriptor, DescriptorPublicKey, Tr};
+        use $crate::miniscript::descriptor::{
+            Descriptor, DescriptorPublicKey, KeyMap, TapTree, Tr,
+        };
         use $crate::miniscript::Tap;
 
         #[allow(unused_imports)]
-        use $crate::keys::{DescriptorKey, IntoDescriptorKey};
+        use $crate::keys::{DescriptorKey, IntoDescriptorKey, ValidNetworks};
+
         let secp = $crate::bitcoin::secp256k1::Secp256k1::new();
 
         $internal_key
@@ -89,12 +92,19 @@ macro_rules! impl_top_level_tr {
             .and_then(|key: DescriptorKey<Tap>| key.extract(&secp))
             .map_err($crate::descriptor::DescriptorError::Key)
             .and_then(|(pk, mut key_map, mut valid_networks)| {
-                let tap_tree = $tap_tree.map(|(tap_tree, tree_keymap, tree_networks)| {
-                    key_map.extend(tree_keymap.into_iter());
-                    valid_networks = $crate::keys::merge_networks(&valid_networks, &tree_networks);
+                let tap_tree = $tap_tree.map(
+                    |(tap_tree, tree_keymap, tree_networks): (
+                        TapTree<DescriptorPublicKey>,
+                        KeyMap,
+                        ValidNetworks,
+                    )| {
+                        key_map.extend(tree_keymap.into_iter());
+                        valid_networks =
+                            $crate::keys::merge_networks(&valid_networks, &tree_networks);
 
-                    tap_tree
-                });
+                        tap_tree
+                    },
+                );
 
                 Ok((
                     Descriptor::<DescriptorPublicKey>::Tr(Tr::new(pk, tap_tree)?),
@@ -1178,5 +1188,36 @@ mod test {
         uncompressed_pk.compressed = false;
 
         descriptor!(wsh(v: pk(uncompressed_pk))).unwrap();
+    }
+
+    #[test]
+    fn test_dsl_tr_only_key() {
+        let private_key =
+            PrivateKey::from_wif("cSQPHDBwXGjVzWRqAHm6zfvQhaTuj1f2bFH58h55ghbjtFwvmeXR").unwrap();
+        let (descriptor, _, _) = descriptor!(tr(private_key)).unwrap();
+
+        assert_eq!(
+            descriptor.to_string(),
+            "tr(02e96fe52ef0e22d2f131dd425ce1893073a3c6ad20e8cac36726393dfb4856a4c)#heq9m95v"
+        )
+    }
+
+    #[test]
+    fn test_dsl_tr_simple_tree() {
+        let private_key =
+            PrivateKey::from_wif("cSQPHDBwXGjVzWRqAHm6zfvQhaTuj1f2bFH58h55ghbjtFwvmeXR").unwrap();
+        let (descriptor, _, _) =
+            descriptor!(tr(private_key, { pk(private_key), pk(private_key) })).unwrap();
+
+        assert_eq!(descriptor.to_string(), "tr(02e96fe52ef0e22d2f131dd425ce1893073a3c6ad20e8cac36726393dfb4856a4c,{pk(02e96fe52ef0e22d2f131dd425ce1893073a3c6ad20e8cac36726393dfb4856a4c),pk(02e96fe52ef0e22d2f131dd425ce1893073a3c6ad20e8cac36726393dfb4856a4c)})#xy5fjw6d")
+    }
+
+    #[test]
+    fn test_dsl_tr_single_leaf() {
+        let private_key =
+            PrivateKey::from_wif("cSQPHDBwXGjVzWRqAHm6zfvQhaTuj1f2bFH58h55ghbjtFwvmeXR").unwrap();
+        let (descriptor, _, _) = descriptor!(tr(private_key, pk(private_key))).unwrap();
+
+        assert_eq!(descriptor.to_string(), "tr(02e96fe52ef0e22d2f131dd425ce1893073a3c6ad20e8cac36726393dfb4856a4c,pk(02e96fe52ef0e22d2f131dd425ce1893073a3c6ad20e8cac36726393dfb4856a4c))#lzl2vmc7")
     }
 }
--- a/src/descriptor/dsl.rs
+++ b/src/descriptor/dsl.rs
@@ -75,6 +75,38 @@ macro_rules! impl_top_level_pk {
 
 #[doc(hidden)]
 #[macro_export]
+macro_rules! impl_top_level_tr {
+    ( $internal_key:expr, $tap_tree:expr ) => {{
+        use $crate::miniscript::descriptor::{Descriptor, DescriptorPublicKey, Tr};
+        use $crate::miniscript::Tap;
+
+        #[allow(unused_imports)]
+        use $crate::keys::{DescriptorKey, IntoDescriptorKey};
+        let secp = $crate::bitcoin::secp256k1::Secp256k1::new();
+
+        $internal_key
+            .into_descriptor_key()
+            .and_then(|key: DescriptorKey<Tap>| key.extract(&secp))
+            .map_err($crate::descriptor::DescriptorError::Key)
+            .and_then(|(pk, mut key_map, mut valid_networks)| {
+                let tap_tree = $tap_tree.map(|(tap_tree, tree_keymap, tree_networks)| {
+                    key_map.extend(tree_keymap.into_iter());
+                    valid_networks = $crate::keys::merge_networks(&valid_networks, &tree_networks);
+
+                    tap_tree
+                });
+
+                Ok((
+                    Descriptor::<DescriptorPublicKey>::Tr(Tr::new(pk, tap_tree)?),
+                    key_map,
+                    valid_networks,
+                ))
+            })
+    }};
+}
+
+#[doc(hidden)]
+#[macro_export]
 macro_rules! impl_leaf_opcode {
     ( $terminal_variant:ident ) => {{
         use $crate::descriptor::CheckMiniscript;
@@ -226,6 +258,62 @@ macro_rules! impl_sortedmulti {
             .and_then(|keys| $crate::keys::make_sortedmulti($thresh, keys, $build_desc, &secp))
     });
 
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! parse_tap_tree {
+    ( @merge $tree_a:expr, $tree_b:expr) => {{
+        use std::sync::Arc;
+        use $crate::miniscript::descriptor::TapTree;
+
+        $tree_a
+            .and_then(|tree_a| Ok((tree_a, $tree_b?)))
+            .and_then(|((a_tree, mut a_keymap, a_networks), (b_tree, b_keymap, b_networks))| {
+                a_keymap.extend(b_keymap.into_iter());
+                Ok((TapTree::Tree(Arc::new(a_tree), Arc::new(b_tree)), a_keymap, $crate::keys::merge_networks(&a_networks, &b_networks)))
+            })
+
+    }};
+
+    // Two sub-trees
+    ( { { $( $tree_a:tt )* }, { $( $tree_b:tt )* } } ) => {{
+        let tree_a = $crate::parse_tap_tree!( { $( $tree_a )* } );
+        let tree_b = $crate::parse_tap_tree!( { $( $tree_b )* } );
+
+        $crate::parse_tap_tree!(@merge tree_a, tree_b)
+    }};
+
+    // One leaf and a sub-tree
+    ( { $op_a:ident ( $( $minisc_a:tt )* ), { $( $tree_b:tt )* } } ) => {{
+        let tree_a = $crate::parse_tap_tree!( $op_a ( $( $minisc_a )* ) );
+        let tree_b = $crate::parse_tap_tree!( { $( $tree_b )* } );
+
+        $crate::parse_tap_tree!(@merge tree_a, tree_b)
+    }};
+    ( { { $( $tree_a:tt )* }, $op_b:ident ( $( $minisc_b:tt )* ) } ) => {{
+        let tree_a = $crate::parse_tap_tree!( { $( $tree_a )* } );
+        let tree_b = $crate::parse_tap_tree!( $op_b ( $( $minisc_b )* ) );
+
+        $crate::parse_tap_tree!(@merge tree_a, tree_b)
+    }};
+
+    // Two leaves
+    ( { $op_a:ident ( $( $minisc_a:tt )* ), $op_b:ident ( $( $minisc_b:tt )* ) } ) => {{
+        let tree_a = $crate::parse_tap_tree!( $op_a ( $( $minisc_a )* ) );
+        let tree_b = $crate::parse_tap_tree!( $op_b ( $( $minisc_b )* ) );
+
+        $crate::parse_tap_tree!(@merge tree_a, tree_b)
+    }};
+
+    // Single leaf
+    ( $op:ident ( $( $minisc:tt )* ) ) => {{
+        use std::sync::Arc;
+        use $crate::miniscript::descriptor::TapTree;
+
+        $crate::fragment!( $op ( $( $minisc )* ) )
+            .map(|(a_minisc, a_keymap, a_networks)| (TapTree::Leaf(Arc::new(a_minisc)), a_keymap, a_networks))
+    }};
 }
 
 #[doc(hidden)]
@@ -440,6 +528,15 @@ macro_rules! descriptor {
     });
     ( wsh ( $( $minisc:tt )* ) ) => ({
         $crate::impl_top_level_sh!(Wsh, new, new_sortedmulti, Segwitv0, $( $minisc )*)
+    });
+
+    ( tr ( $internal_key:expr ) ) => ({
+        $crate::impl_top_level_tr!($internal_key, None)
+    });
+    ( tr ( $internal_key:expr, $( $taptree:tt )* ) ) => ({
+        let tap_tree = $crate::parse_tap_tree!( $( $taptree )* );
+        tap_tree
+            .and_then(|tap_tree| $crate::impl_top_level_tr!($internal_key, Some(tap_tree)))
     });
 }
 
--- a/src/descriptor/dsl.rs
+++ b/src/descriptor/dsl.rs
@@ -579,6 +579,23 @@ impl<A, B, C> From<(A, (B, (C, ())))> for TupleThree<A, B, C> {
 
 #[doc(hidden)]
 #[macro_export]
+macro_rules! group_multi_keys {
+    ( $( $key:expr ),+ ) => {{
+        use $crate::keys::IntoDescriptorKey;
+
+        let keys = vec![
+            $(
+                $key.into_descriptor_key(),
+            )*
+        ];
+
+        keys.into_iter().collect::<Result<Vec<_>, _>>()
+            .map_err($crate::descriptor::DescriptorError::Key)
+    }};
+}
+
+#[doc(hidden)]
+#[macro_export]
 macro_rules! fragment_internal {
     // The @v prefix is used to parse a sequence of operands and return them in a vector. This is
     // used by operands that take a variable number of arguments, like `thresh()` and `multi()`.
@@ -737,21 +754,22 @@ macro_rules! fragment {
             .and_then(|items| $crate::fragment!(thresh_vec($thresh, items)))
     });
     ( multi_vec ( $thresh:expr, $keys:expr ) ) => ({
-        $crate::keys::make_multi($thresh, $keys)
-    });
-    ( multi ( $thresh:expr $(, $key:expr )+ ) ) => ({
-        use $crate::keys::IntoDescriptorKey;
         let secp = $crate::bitcoin::secp256k1::Secp256k1::new();
 
-        let keys = vec![
-            $(
-                $key.into_descriptor_key(),
-            )*
-        ];
+        $crate::keys::make_multi($thresh, $crate::miniscript::Terminal::Multi, $keys, &secp)
+    });
+    ( multi ( $thresh:expr $(, $key:expr )+ ) ) => ({
+        $crate::group_multi_keys!( $( $key ),* )
+            .and_then(|keys| $crate::fragment!( multi_vec ( $thresh, keys ) ))
+    });
+    ( multi_a_vec ( $thresh:expr, $keys:expr ) ) => ({
+        let secp = $crate::bitcoin::secp256k1::Secp256k1::new();
 
-        keys.into_iter().collect::<Result<Vec<_>, _>>()
-            .map_err($crate::descriptor::DescriptorError::Key)
-            .and_then(|keys| $crate::keys::make_multi($thresh, keys, &secp))
+        $crate::keys::make_multi($thresh, $crate::miniscript::Terminal::MultiA, $keys, &secp)
+    });
+    ( multi_a ( $thresh:expr $(, $key:expr )+ ) ) => ({
+        $crate::group_multi_keys!( $( $key ),* )
+            .and_then(|keys| $crate::fragment!( multi_a_vec ( $thresh, keys ) ))
     });
 
     // `sortedmulti()` is handled separately
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -14,15 +14,15 @@
 //! This module contains generic utilities to work with descriptors, plus some re-exported types
 //! from [`miniscript`].
 
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{BTreeMap, HashSet};
 use std::ops::Deref;
 
-use bitcoin::secp256k1;
 use bitcoin::util::bip32::{ChildNumber, DerivationPath, ExtendedPubKey, Fingerprint, KeySource};
 use bitcoin::util::{psbt, taproot};
+use bitcoin::{secp256k1, PublicKey, XOnlyPublicKey};
 use bitcoin::{Network, Script, TxOut};
 
-use miniscript::descriptor::{DescriptorType, InnerXKey};
+use miniscript::descriptor::{DescriptorType, InnerXKey, SinglePubKey};
 pub use miniscript::{
     descriptor::DescriptorXKey, descriptor::KeyMap, descriptor::Wildcard, Descriptor,
     DescriptorPublicKey, Legacy, Miniscript, ScriptContext, Segwitv0,
@@ -322,6 +322,16 @@ pub(crate) trait DescriptorMeta {
         hd_keypaths: &HdKeyPaths,
         secp: &'s SecpCtx,
     ) -> Option<DerivedDescriptor<'s>>;
+    fn derive_from_tap_key_origins<'s>(
+        &self,
+        tap_key_origins: &TapKeyOrigins,
+        secp: &'s SecpCtx,
+    ) -> Option<DerivedDescriptor<'s>>;
+    fn derive_from_psbt_key_origins<'s>(
+        &self,
+        key_origins: BTreeMap<Fingerprint, (&DerivationPath, SinglePubKey)>,
+        secp: &'s SecpCtx,
+    ) -> Option<DerivedDescriptor<'s>>;
     fn derive_from_psbt_input<'s>(
         &self,
         psbt_input: &psbt::Input,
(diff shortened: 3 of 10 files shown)


afilini:
I've updated the PR description with a list of ideas for tests that we need for this. The list is non-exhaustive, just a point from which we can start.

notmandatory:
I've started looking into the "Creating psbts" tests. 

Some notes on setting up a local `bitcoind -regtest` taproot wallet here: https://hackmd.io/@notmandatory/Sy2yOYeI5

rajarshimaitra:
I am finding a lot of version conflict while working on this https://github.com/bitcoindevkit/bdk-cli/issues/62#issuecomment-1124655758 while trying to update bdk-cli with the latest `bitcoincore-rpc` and `electrsd`.. 

I think the first two commit of this PR if can be merged independently of taproot into master would be really helpful.. 

It seems recently a lot has changed in the upstreams and we are using `bitcoincore-rpc` and `electrum-client` from two different sources (directly, and then from `electrsd`). And when we try to move `electrsd` from dev dependency all sort of version conflict is appearing in bdk-cli..

But to sort them in bdk-cli side I need to have the final updated version of Cargo.lock in bdk, which is done by the first two commits..

rajarshimaitra:
It also might make sense if we just make a patch release with the dependency upgrade, then the downstream libraries can fix them up. Or just having them in master would work too for time being.

afilini:
I don't think it can be released as patch upgrade because it would be a breaking change. I.e. if you depend on bdk `0.18` you would get the update and it would break your code.

I think I agree, I'm gonna open a new PR now with just the upgrade

afilini:
As discussed in our last meeting on Discord, I've added commit ea165b02bf7698242b8c7ff5485b23331c2ac832 to introduce the concept of a `SignerContext`, which helps the signer produce the right signature without having to rely on specific fields of the PSBT being set.

Unfortunately since the context is static (reused for all the signatures of the same signer) I couldn't store the list of taproot leaf hashes in there, because those change with every derivation index. So I only store whether the key matches the "internal_key", but then we still rely on the PSBT to figure out the list of leaf hashes to sign for.

notmandatory:
I'm working on some commits to add to this PR with tests for "Interoperability with other wallets ", see: https://github.com/afilini/bdk/pull/4. So far only testing with core (22.0).

afilini:
I'm working on the "signing psbts" section

src/wallet/signer.rs
```diff
@@ -301,48 +384,122 @@ impl Signer for PrivateKey {
...
+) {
+    use schnorr::TapTweak;
+
+    let keypair = secp256k1::KeyPair::from_seckey_slice(secp, secret_key.as_ref()).unwrap();
+    let keypair = match leaf_hash {
+        None => keypair
+            .tap_tweak(secp, psbt_input.tap_merkle_root)
+            .into_inner(),
+        Some(_) => keypair, // no tweak for script spend
+    };
```
rajarshimaitra, line 487: What if the secret key given here doesn't match with the pubkey?? Better to make a proactive check for that as we have the pubkey as an input already??

src/wallet/signer.rs
```diff
@@ -301,48 +384,122 @@ impl Signer for PrivateKey {
...
-
-    fn descriptor_secret_key(&self) -> Option<DescriptorSecretKey> {
-        Some(DescriptorSecretKey::SinglePriv(DescriptorSinglePriv {
-            key: *self,
-            origin: None,
-        }))
+fn produce_schnorr_sig(
+    secret_key: &secp256k1::SecretKey,
+    pubkey: XOnlyPublicKey,
+    leaf_hash: Option<taproot::TapLeafHash>,
```
rajarshimaitra, line 473: A comment somewhere here saying making the `leaf_hash` `None` will use the signature for key path spend, would be helpful. Even if just for internal use, because I will surely forget this.. 😅

src/wallet/signer.rs
```diff
@@ -229,19 +299,25 @@ impl Signer for DescriptorXKey<ExtendedPrivKey> {
...
                 } else {
                     None
                 }
             })
             .next()
         {
-            Some((pk, full_path)) => (pk, full_path.clone()),
+            Some((pk, full_path)) => (pk, full_path),
             None => return Ok(()),
         };
```
rajarshimaitra: Can't we just `collect()` them out here instead of `next() { ... }` block?? Found this hard to parse through..

src/wallet/signer.rs
```diff
@@ -301,48 +384,122 @@ impl Signer for PrivateKey {
...
-            .insert(pubkey, final_signature);
-
         Ok(())
     }
+}
 
-    fn sign_whole_tx(&self) -> bool {
-        false
-    }
+fn produce_ecsda_sig(
```
rajarshimaitra: Nit: because these functions modifies a psbt and not produce any signature, maybe they are better named as `sign_psbt_schnorr` or something?? 

Nothing blocking, may consider next time you touch this..

src/wallet/signer.rs
```diff
@@ -631,7 +822,62 @@ impl ComputeSighash for Segwitv0 {
...
+            )
+        } else if witness_utxos.iter().all(Option::is_some) {
+            all_witness_utxos.extend(witness_utxos.iter().filter_map(|x| x.as_ref()));
+            sighash::Prevouts::All(&all_witness_utxos)
+        } else {
+            return Err(SignerError::MissingWitnessUtxo);
+        };
+
+        // Assume no OP_CODESEPARATOR
+        let extra = extra.map(|leaf_hash| (leaf_hash, 0xFFFFFFFF));
```
rajarshimaitra, line 875: What happens when there are code separators?? I am still cloudy on where does the code separator comes from and what it does.. As far as I can figure out its somewhere in the witness script itself?? So we have to count them out and then input it here??

rajarshimaitra (commented):
Review ACK 8b709a595dc0da324b8cd357fbf4345f280b8bb9

Below are some non blocking comments and questions.. 

It seems the `wallet::signer` module can use some unit tests?? Is that necessary or they are covered elsewhere??

src/wallet/signer.rs
```diff
@@ -174,27 +174,46 @@ impl fmt::Display for SignerError {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SignerContext {
+    /// Legacy context
+    Legacy,
+    /// Segwit v0 context (BIP 143)
+    Segwitv0,
+    /// Taproot context (BIP 340)
+    Tap {
+        /// Whether the signer can sign for the internal key or not
+        internal_key: bool,
```
rajarshimaitra: Nit: Can it be named `is_internal_key` (intends a flag) instead??

rajarshimaitra (commented):
ACK 9034096089cb1067ad1a3bde94f43f74ed8dd2b2

just one nit..

src/wallet/signer.rs
afilini: I would skip the check here considering that we already have some of them in the `InputSigner` implementation, and an extra ec multiplication can be a bit slower.

Worst case we would just produce an invalid signature, or place a valid signature in the wrong spot.

src/wallet/signer.rs
afilini: If there are codeseparators and we don't account for them we may produce an invalid signature.

However, I haven't seen a single example script for taproot using op_codeseparator. The BIP mentions that there may be one very specific use case, but it doesn't even show how to build it.

Miniscript for sure doesn't use them, so as long as we only support miniscript descriptors we should be fine. We may have to reconsider this after #485, because then a user could make a custom script with the codeseparator, but I wouldn't worry too much about it.

src/wallet/signer.rs
afilini: Collecting allocates a new vec but we really only need the first item, so that would be a bit more wasteful.

I'll see if I can write this down more clearly.

afilini: Ok I switched to `find_map()` which directly returns the first non-`None` item, instead of mapping the iterator and taking the first item.

src/wallet/mod.rs
```diff
@@ -1494,17 +1503,21 @@ where
```
rajarshimaitra, line 1533: Unrelated to taproot, but in the lines above, we are deriving the `KeyChainKind` and `Child` index for the spk.. The `KeyChainKind` is also present in `LocalUtxo`. So is it better to make a match that these two are same? Or its always guaranted to be the same?

src/wallet/mod.rs
```diff
@@ -1496,17 +1528,35 @@ where
...
+                        .control_block(script_ver)
+                        .map(|cb| (cb, script_ver.clone()))
+                })
+                .collect();
+        } else {
+            psbt_input.bip32_derivation = derived_descriptor.get_hd_keypaths(&self.secp);
+        }
 
         psbt_input.redeem_script = derived_descriptor.psbt_redeem_script();
         psbt_input.witness_script = derived_descriptor.psbt_witness_script();
```
rajarshimaitra, line 1557: It seems to me that in these two cases we are not considering `Wpkh` types?? What happens for them??

https://github.com/bitcoindevkit/bdk/blob/4c1749a13a29da6be9812948f6aefb1ed74f166f/src/descriptor/mod.rs#L330-L349

src/wallet/mod.rs
```diff
@@ -4097,4 +4170,167 @@ pub(crate) mod test {
...
+        assert_eq!(
+            psbt.inputs[0].tap_key_origins.len(),
+            1,
+            "Empty input tap_key_origins"
+        );
+        assert_eq!(
+            psbt.outputs[0].tap_key_origins.len(),
+            1,
+            "Empty output tap_key_origins"
+        );
```
rajarshimaitra: Instead of only checking length feels like its better to assert the data structure itself?? The descriptor is hardcoded so they should not change..

src/wallet/mod.rs
```diff
@@ -4097,4 +4170,167 @@ pub(crate) mod test {
...
+        let mut builder = wallet.build_tx();
+        builder.add_recipient(addr.script_pubkey(), 25_000);
+        let (psbt, _) = builder.finish().unwrap();
+
+        assert!(
+            psbt.inputs[0].tap_merkle_root.is_some(),
+            "Empty input tap_merkle_root"
+        );
+        assert_eq!(psbt.inputs[0].tap_scripts.len(), 2, "Empty tap_scripts");
+    }
```
rajarshimaitra: Same test comment on asserting the data than checking length..

rajarshimaitra (commented):
tACK 58d5abc5225572fedd641644456f09fca31879d3

Few nits and comments..

src/wallet/signer.rs
rajarshimaitra: Yes.. Thats even more better..

src/wallet/signer.rs
rajarshimaitra: Okay ya makes sense, as long as we have some checks elsewhere..

src/wallet/signer.rs
rajarshimaitra: Ya that makes sense.. If its not that much useful no point worrying much on it.. I also haven't seen any codeseperator in the wild.. Maybe we can make a note somewhere saying BDK doesn't support it yet..

src/wallet/mod.rs
afilini: In general that kind of mismatch should only happen if the database is inconsistent, or if the user is actively manipulating the data.

Besides, all of the code right now assumes the database is the ultimate source of truth, so I think it's fine here to ignore the value of `LocalUtxo` and just use what comes out of it.

src/wallet/mod.rs
afilini: The `redeem_script` field is only populated for P2SH and P2WSH address types, so here we ignore all the others (P2TR, P2PKH, P2WPKH)

src/wallet/mod.rs
afilini: Yes, that's a good point

afilini:
I'm now working on the policy module tests, I will probably include some for the descriptor macro as well which were not part of the original list

notmandatory:
When you get a chance this needs a rebase to pick up 35feb107ed5969720ce54a6aa76b7b2176f6c7c1 to get CI to run the blockchain tests correctly.

src/wallet/mod.rs
```diff
@@ -1436,8 +1437,10 @@ where
                     psbt_input: foreign_psbt_input,
                     outpoint,
                 } => {
-                    // TODO: do not require non_witness_utxo for taproot utxos
-                    if !params.only_witness_utxo && foreign_psbt_input.non_witness_utxo.is_none() {
+                    if foreign_psbt_input.tap_internal_key.is_none()
```
danielabrozzoni: Can you add a small comment explaining why it's ok not having the `non_witness_utxo` if you don't have the tap internal key? Even a small comment with a URL to the doc is ok

src/wallet/signer.rs
```diff
@@ -631,7 +760,62 @@ impl ComputeSighash for Segwitv0 {
...
+            sighash::Prevouts::One(
+                input_index,
+                witness_utxos[input_index]
+                    .as_ref()
+                    .ok_or(SignerError::MissingWitnessUtxo)?,
+            )
+        } else if witness_utxos.iter().all(Option::is_some) {
+            all_witness_utxos.extend(witness_utxos.iter().filter_map(|x| x.as_ref()));
+            sighash::Prevouts::All(&all_witness_utxos)
+        } else {
```
danielabrozzoni, line 870: TODO: test with ANYONECANPAY, ALL and with missing witness utxos

src/descriptor/mod.rs
```diff
@@ -515,7 +523,64 @@ impl<'s> DerivedDescriptorMeta for DerivedDescriptor<'s> {
...
+            let key_origin = match pk.deref() {
+                DescriptorPublicKey::XPub(xpub) => {
+                    Some((xpub.root_fingerprint(secp), xpub.full_path(&[])))
+                }
+                DescriptorPublicKey::SinglePub(_) => None,
+            };
+
+            // If this is the internal key, we only insert the key origin if it's not None.
+            // For keys found in the tap tree we always insert a key origin (because the signer
+            // looks for it to know which leaves to sign for), but the key origin may be empty
```
danielabrozzoni: nit: I would rewrite this as
```
 // For keys found in the tap tree we always insert a key origin (because the signer
 // looks for it to know which leaves to sign for), even though it may be None
```

It's probably only me, but when I read it the first time I thought there was some kinda typo as you were repeating "key origin" twice 😅  Feel free to ignore.

src/wallet/mod.rs
```diff
@@ -1461,7 +1462,15 @@ where
...
 
-                psbt_output.bip32_derivation = derived_descriptor.get_hd_keypaths(&self.secp)?;
+                if desc.is_taproot() {
+                    psbt_output
+                        .tap_key_origins
+                        .append(&mut derived_descriptor.get_tap_key_origins(&self.secp));
+                } else {
+                    psbt_output
+                        .bip32_derivation
+                        .append(&mut derived_descriptor.get_hd_keypaths(&self.secp));
```
danielabrozzoni, line 1502: You were assigning `derived_descriptor.get_hd_keypaths(&self.secp)` to `psbt_output.bip32_derivation` before, now you're appending. Is there a specific reason for doing so? I'm trying to understand if I'm missing something 😅

src/wallet/mod.rs
```diff
@@ -1013,23 +1014,27 @@ where
         // this helps us doing our job later
         self.add_input_hd_keypaths(psbt)?;
 
-        // If we aren't allowed to use `witness_utxo`, ensure that every input but finalized one
-        // has the `non_witness_utxo`
+        // If we aren't allowed to use `witness_utxo`, ensure that every input (except p2tr and finalized ones)
+        // have the `non_witness_utxo`
```
danielabrozzoni: nit: has?

src/wallet/mod.rs
```diff
@@ -1013,23 +1014,27 @@ where
...
         // If the user hasn't explicitly opted-in, refuse to sign the transaction unless every input
-        // is using `SIGHASH_ALL`
+        // is using `SIGHASH_ALL` or `SIGHASH_DEFAULT` for taproot
         if !sign_options.allow_all_sighashes
             && !psbt.inputs.iter().all(|i| {
-                i.sighash_type.is_none() || i.sighash_type == Some(EcdsaSighashType::All.into())
+                i.sighash_type.is_none()
+                    || i.sighash_type == Some(EcdsaSighashType::All.into())
+                    || i.sighash_type == Some(SchnorrSighashType::All.into())
+                    || i.sighash_type == Some(SchnorrSighashType::Default.into())
```
danielabrozzoni, line 1037: Does this mean we don't return an error here if we have an input spending non-taproot outputs with `SchnorrSighashType::Default`?

src/wallet/mod.rs
```diff
@@ -4095,4 +4130,39 @@ pub(crate) mod test {
             "when there's no internal descriptor it should just use external"
         );
     }
+
+    #[test]
+    fn test_taproot_psbt_populate_tap_key_origins() {
```
rajarshimaitra, line 4195: In this test we are testing the script path key origin insertion.. Probably that's because you didn't had the script path tr descriptor at that time.. But now we can test script path key origin insertion using `get_test_tr_with_taptree_xprv` it seems?

rajarshimaitra (commented):
tACK f2f2d56ab77578ce1a6746ed30267cd3a3628676

Just one test suggestion..

src/wallet/mod.rs
```diff
@@ -4165,4 +4203,35 @@ pub(crate) mod test {
...
+                )
+                .unwrap()
+            ),
+        );
+        assert_eq!(
+            psbt.inputs[0].tap_scripts.clone().into_iter().collect::<Vec<_>>(),
+            vec![
+                (taproot::ControlBlock::from_slice(&Vec::<u8>::from_hex("c151494dc22e24a32fe9dcfbd7e85faf345fa1df296fb49d156e859ef3452012958b0afded0ee3149dbf6710d349dc30e55ae30c319c30efbc79efe19cf70f46a8").unwrap()).unwrap(), (from_str!("208aee2b8120a5f157f1223f72b5e62b825831a27a9fdf427db7cc697494d4a642ac"), taproot::LeafVersion::TapScript)),
+                (taproot::ControlBlock::from_slice(&Vec::<u8>::from_hex("c151494dc22e24a32fe9dcfbd7e85faf345fa1df296fb49d156e859ef345201295b9a515f7be31a70186e3c5937ee4a70cc4b4e1efe876c1d38e408222ffc64834").unwrap()).unwrap(), (from_str!("20b511bd5771e47ee27558b1765e87b541668304ec567721c7b880edc0a010da55ac"), taproot::LeafVersion::TapScript)),
+            ],
```
rajarshimaitra: Might make sense to also assert the `psbt_input.tap_internal_key` and `psbt_input.tap_merkle_root`?? 
Also same for the `psbt.output` taproot data..

rajarshimaitra (commented):
tACK 14d86623ca75c4be3f633d2495d96d2ac3f52535

src/wallet/mod.rs
afilini: It's kinda testing two different things, but I agree it makes sense to test that part as well. Here we are testing the key origins for key spend (because the leaf hash vec is empty), I'll add one more test for when the key appears in multiple leaves to make sure we are accounting for all of them

src/wallet/mod.rs
afilini: The merkle root is asserted a few lines above it. I'll change the test to send back to the same address as the input so that I can then assert that the input metadata are equal to the output

afilini: I just realized that these metadata are not present in the output, so there's not much to check there

danielabrozzoni (commented):
Concept ACK :)
I have a couple of comments, mainly regarding making the code easier to understand.
I still haven't looked into tests

src/wallet/mod.rs
afilini: Appending doesn't override what's already there. So let's say a wallet creates the PSBT with some key origins that we don't know of, then while we sign we would delete them, which would potentially make it impossible for a third wallet to sign if it needed those specific origins.

src/wallet/mod.rs
afilini: That's true, although it will fail later on in the signer code when it tires to "cast" the `PsbtSighashType` to an `EcdsaSighashType`.

Checking that you are using a "compatible" sighash type in my opinion falls in the "integrity check" kind of validation, which we don't really do right now in BDK. Maybe we should start, but that's not in the scope of this PR I think

src/descriptor/mod.rs
```diff
@@ -393,61 +411,124 @@ impl DescriptorMeta for ExtendedDescriptor {
...
 
+        // using `for_any_key` should make this stop as soon as we return `true`
+        self.for_any_key(|key| {
             if let DescriptorPublicKey::XPub(xpub) = key.as_key().deref() {
-                // Check if the key matches one entry in our `index`. If it does, `matches()` will
+                // Check if the key matches one entry in our `key_origins`. If it does, `matches()` will
                 // return the "prefix" that matched, so we remove that prefix from the full path
-                // found in `index` and save it in `derive_path`. We expect this to be a derivation
+                // found in `key_origins` and save it in `derive_path`. We expect this to be a derivation
                 // path of length 1 if the key is `wildcard` and an empty path otherwise.
```
rajarshimaitra, line 444: What happens when the `derive_path` is more than one childnumber long?? Should we assert it somewhere and throw an error as unexpected behavior?? There is no restriction on the input derivation path in key origin to be only extending it by one childnumber.. 

Right now I think if the input diverges from the expectation the `if let` statement at the end will be skipped and it won't return anything.. But that can be confusing to users..

rajarshimaitra (commented):
tACK 6fa5832a321b3596d726277107df59685e2904d9

src/wallet/mod.rs
```diff
@@ -4097,4 +4185,269 @@ pub(crate) mod test {
...
+        let (wallet, _, _) = get_funded_wallet(get_test_tr_with_taptree());
+        test_spend_from_wallet(wallet);
+
+        let (wallet, _, _) = get_funded_wallet(get_test_tr_with_taptree_xprv());
+        test_spend_from_wallet(wallet);
+    }
+
+    #[test]
+    fn test_taproot_sign_derive_index_from_psbt() {
+        env_logger::init();
```
danielabrozzoni: Debugging leftover?

src/wallet/mod.rs
```diff
@@ -4097,4 +4185,269 @@ pub(crate) mod test {
...
+                    ],
+                    (Default::default(), Default::default())
+                )
+            )],
+            "Wrong output tap_key_origins"
+        );
+    }
+
+    #[test]
+    fn test_taproot_psbt_input_tap_tree() {
```
danielabrozzoni, line 4291: Added a check for the tap tree:
```
diff --git a/src/wallet/mod.rs b/src/wallet/mod.rs
index 344bf11..8f7510f 100644
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -4284,6 +4284,8 @@ pub(crate) mod test {
 
     #[test]
     fn test_taproot_psbt_input_tap_tree() {
+        use crate::bitcoin::psbt::serialize::Deserialize;
+        use crate::bitcoin::psbt::TapTree;
         use bitcoin::hashes::hex::FromHex;
         use bitcoin::util::taproot;
 
@@ -4323,6 +4325,11 @@ pub(crate) mod test {
             psbt.inputs[0].tap_internal_key,
             psbt.outputs[0].tap_internal_key
         );
+
+        assert_eq!(
+            psbt.outputs[0].tap_tree,
+            Some(TapTree::deserialize(&Vec::<u8>::from_hex("01c022208aee2b8120a5f157f1223f72b5e62b825831a27a9fdf427db7cc697494d4a642ac01c0222051494dc22e24a32fe9dcfbd7e85faf345fa1df296fb49d156e859ef345201295ac",).unwrap()).unwrap())
+        );
     }
 
     fn test_spend_from_wallet(wallet: Wallet<AnyDatabase>) {
```

src/descriptor/dsl.rs
```diff
@@ -640,21 +764,22 @@ macro_rules! fragment {
...
     });
     ( multi_vec ( $thresh:expr, $keys:expr ) ) => ({
-        $crate::keys::make_multi($thresh, $keys)
+        let secp = $crate::bitcoin::secp256k1::Secp256k1::new();
+
+        $crate::keys::make_multi($thresh, $crate::miniscript::Terminal::Multi, $keys, &secp)
     });
     ( multi ( $thresh:expr $(, $key:expr )+ ) ) => ({
-        use $crate::keys::IntoDescriptorKey;
+        $crate::group_multi_keys!( $( $key ),* )
```
rajarshimaitra, line 772: Probably I am missing something, but why the input `key` pattern is `1 or many` and the pattern in code is `0 or many`? (the `+` in input and `*` in the code after `key`)

src/descriptor/dsl.rs
```diff
@@ -640,21 +764,22 @@ macro_rules! fragment {
...
-            )*
-        ];
-
-        keys.into_iter().collect::<Result<Vec<_>, _>>()
-            .map_err($crate::descriptor::DescriptorError::Key)
-            .and_then(|keys| $crate::keys::make_multi($thresh, keys, &secp))
+        $crate::keys::make_multi($thresh, $crate::miniscript::Terminal::MultiA, $keys, &secp)
+    });
+    ( multi_a ( $thresh:expr $(, $key:expr )+ ) ) => ({
+        $crate::group_multi_keys!( $( $key ),* )
```
rajarshimaitra, line 781: Similar question on key pattern..

src/keys/mod.rs
```diff
@@ -792,13 +792,18 @@ pub fn make_pkh<Pk: IntoDescriptorKey<Ctx>, Ctx: ScriptContext>(
 
 // Used internally by `bdk::fragment!` to build `multi()` fragments
 #[doc(hidden)]
-pub fn make_multi<Pk: IntoDescriptorKey<Ctx>, Ctx: ScriptContext>(
+pub fn make_multi<
+    Pk: IntoDescriptorKey<Ctx>,
+    Ctx: ScriptContext,
+    V: Fn(usize, Vec<DescriptorPublicKey>) -> Terminal<DescriptorPublicKey, Ctx>,
```
rajarshimaitra, line 798: Conceptual question.

The V here is a function but, in the usages of `make_multi` in dsl macros its used as, 
`$crate::keys::make_multi($thresh, $crate::miniscript::Terminal::Multi, $keys, &secp)`.

So `V` here is a `miniscript::Terminal::Multi` which is an enum..

How does this work??

rajarshimaitra (commented):
tACK 4c50a9e37f12685b8f786cb56a25a98323a0d7a5

src/keys/mod.rs
afilini: `Terminal::Multi` is not actually an enum, it's an enum *variant*. Just `Terminal` is the enum. Rust allows you to treat an enum variant as a function that creates an instance of that variant.

You may have used this pattern before when mapping errors since it's pretty common:

```rust
returns_a_result().map_err(MyErr::OtherVariant)
```

This is equivalent to

```rust
returns_a_result().map_err(|e| MyErr::OtherVariant(e))
```

src/descriptor/dsl.rs
afilini: I think it doesn't matter what you use when expanding repeated items, the compiler only looks at what you have in the "declaration" to match the tokens passed to the macro.

So in this case it should error if there's no key at all (because of the `+`), but afterwards you just want to iterate over whatever you have and for that you can just use `*`.

src/descriptor/mod.rs
afilini: This is actually guaranteed by miniscript, the `matches()` function returns a path of either length one or zero (see https://docs.rs/miniscript/7.0.0/src/miniscript/descriptor/key.rs.html#672).

afilini:
I should have addressed all the comments, here's the range-diff: https://gist.github.com/afilini/6e933f087971c83eb3b422a2a8b5f7aa

afilini:
I still have a few things I'd like to add to this, but I feel like this is growing a bit too much.

We should be done with the tests, how do you feel about merging this the way it is now, and adding more improvements in a follow-up PR? The main one I have in mind is a way to ask the signer to only sign for some specific leaf hashes, or for the internal key, since we currently sign for everything we can.

CHANGELOG.md
```diff
@@ -10,8 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New MSRV set to `1.56`
 - Unpinned tokio to `1`
 - Add traits to reuse `Blockchain`s across multiple wallets (`BlockchainFactory` and `StatelessBlockchain`).
-- Upgrade to rust-bitcoin `0.28`
 - If using the `sqlite-db` feature all cached wallet data is deleted due to a possible UTXO inconsistency, a wallet.sync will recreate it  
+- Update `PkOrF` in the policy module to become an enum
```
notmandatory, line 15: Should also add a note here about adding *experimental* `tr()` descriptor support, or something along those lines. :-)

notmandatory (approved):
ACK 3d1f72e7459524eca07c9546a6a959401f71305f

I was only able to grok parts of this, but I agree with merging it now as *experimental* `tr()` support so we can get it into users/testers hands. This is an amazing amount of work and I hope over time I and other will be able to help @afilini more on contributing to deep changes like this. From what I've seen even bitcoin core descriptor wallets don't yet have `tr()` support that is as extensive as this PR provides.

danielabrozzoni (approved):
Core review ACK 3d1f72e7459524eca07c9546a6a959401f71305f - I reviewed the code to the best of my knowledge, and it looks correct. I haven't done any manual testing, other than running `cargo test`, but I can do that after this is merged. 

I'm hyped about taproot being merged in BDK. Nice job!! 🚀

afilini:
I've opened #616 so that we don't forget the extra stuff that we can add after this pr.

I'll wait for a comment from @rajarshimaitra before merging, I'm assuming he must be close to the end since he started looking at the macro stuff.

src/descriptor/dsl.rs
rajarshimaitra: Ok, clear now.. Didn't knew that.. Thanks..

src/keys/mod.rs
rajarshimaitra: Oh this was interesting.. Cool stuff.. Yeah I have seen that pattern before, didn't realize that's what happening here..  Thanks..

rajarshimaitra (commented):
tACK 8ce8da187ec79f04a348011bc9c76f29c0352a48

CHANGELOG.md
```diff
@@ -10,8 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New MSRV set to `1.56`
 - Unpinned tokio to `1`
 - Add traits to reuse `Blockchain`s across multiple wallets (`BlockchainFactory` and `StatelessBlockchain`).
-- Upgrade to rust-bitcoin `0.28`
```
rajarshimaitra, line 13: Why we are removing this line?

CHANGELOG.md
rajarshimaitra: +1 

Yes experimental note for having taproot capabilities would be useful.. Also its not just the `tr()` descriptor, but the full suit of taproot capability we have now in bdk right??

rajarshimaitra (approved):
tACK 3d1f72e7459524eca07c9546a6a959401f71305f

All tests and changes look good so far.. I think its good to get merged as experimental taproot capabilities..

CHANGELOG.md
afilini: Good catch, this was probably a mistake in the rebase

afilini:
I noticed a new release of rust-bitcoin came out, I'm not sure what's new there but just in case I've updated the `Cargo.toml` to use that new version.

I've also updated the changelog with a brief list of what we support related to taproot.

CHANGELOG.md
```diff
@@ -12,6 +12,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add traits to reuse `Blockchain`s across multiple wallets (`BlockchainFactory` and `StatelessBlockchain`).
 - Upgrade to rust-bitcoin `0.28`
```
notmandatory, line 13:
```suggestion
- Upgrade to rust-bitcoin `0.28.1`
```
Looks like this rust-bitcoin version is only to enable higher MSRV for no_std CI tests. But still a good idea to do and update here in readme.

CHANGELOG.md
notmandatory: :+1: These notes look good.

notmandatory:
> I noticed a new release of rust-bitcoin came out, I'm not sure what's new there but just in case I've updated the `Cargo.toml` to use that new version.

Looks like the new `rust-bitcoin` 0.28.1 is only to make an exception and raise the MSRV to `1.47` for CI testing `no_std`, so probably doesn't affect `bdk` but won't hurt to specify as our min version.

PR #596, opened 2022-05-03
Bump MSRV to 1.56

afilini:
### Description

Following the discussion in #331, bump the MSRV to `1.56`. We already have other PRs bumping it to at least `1.51` (#593), but I'm felling like we are always lagging behind and our CI breaks regularly. As @LLFourn suggested, this PR makes a relatively large bump, hoping this buys us enough time to finish splitting up BDK, which will allow us to have a lower MSRV for the "core" crate.

#### All Submissions:

#### New Features:

### DIFF
--- a/src/blockchain/compact_filters/peer.rs
+++ b/src/blockchain/compact_filters/peer.rs
@@ -94,8 +94,7 @@ impl Mempool {
             TxIdentifier::Wtxid(wtxid) => self.0.read().unwrap().wtxids.get(&wtxid).cloned(),
         };
 
-        txid.map(|txid| self.0.read().unwrap().txs.get(&txid).cloned())
-            .flatten()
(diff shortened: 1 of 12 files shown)


src/blockchain/compact_filters/peer.rs
```diff
@@ -111,6 +110,7 @@ impl Mempool {
 
 /// A Bitcoin peer
 #[derive(Debug)]
+#[allow(clippy::dead_code)]
```
notmandatory: Have to remove `clippy::`
```suggestion
#[allow(dead_code)]
```

notmandatory (approved):
ACK cca69481eb7edb28532a4ed78185248d410050ac

I agree we need to update MSRV to support frequent changes in Blockchain (and Database) dependencies and once core bdk is in it's own crate we can work on supporting lower MSRVs there.

PR #599, opened 2022-05-05
Fix typo in docs for TxBuilder allow_shrinking method

thunderbiscuit:
### Description
This PR fixes a small typo in the documentation for the `allow_shrinking()` method on the `TxBuilder`.

### Notes to the reviewers
The sentence
```txt
Explicitly tells the wallet that it is allowed to reduce the fee of the output matching this `script_pubkey` in order to bump the transaction fee. Without specifying this the wallet will attempt to find a change output to shrink instead.
```
was changed for
```txt
Explicitly tells the wallet that it is allowed to reduce the amount of the output matching this `script_pubkey` in order to bump the transaction fee. Without specifying this the wallet will attempt to find a change output to shrink instead.
```

To reflect the fact that it's the _amount_ of the output that is being reduced in order to leave more for the fees.

### All Submissions:


notmandatory (approved):
ACK 4628a10191f451a7b0ade674e9f8568bf695e49c

Good fix.

PR #601, opened 2022-05-06
Add message signer & message signature verifier

tunikum:
### Description

Ths PR adds the possibility to 
* sign an arbitrary message 
* verify a given message signature 

using ECDSA.

#### All Submissions:

#### New Features:

### DIFF
--- a/src/keys/msgsig.rs
+++ b/src/keys/msgsig.rs
@@ -5,18 +5,6 @@ use bitcoin::secp256k1::{All, Message, Secp256k1, SecretKey};
 use bitcoin::util::misc::{signed_msg_hash, MessageSignature};
 use bitcoin::{Address, Network, PrivateKey, PublicKey};
 
-/// Trait for message signers
-pub trait MessageSigner<S> {
-    /// Sign an arbitrary message
-    fn sign(&self, msg: &str) -> S;
-}
-
-/// Trait for message signature verifiers
-pub trait MessageSignatureVerifier<S> {
-    /// Verify a signature against an arbitrary message
-    fn verify(&self, sig: S, msg: &str) -> bool;
-}
-
 /// A message signer using ECDSA
 pub struct EcdsaMessageSigner<'a> {
     secp: &'a Secp256k1<All>,
@@ -33,9 +21,7 @@ impl<'a> EcdsaMessageSigner<'a> {
     pub fn from_secret_key(secret_key: SecretKey, secp: &'a Secp256k1<All>) -> Self {
         EcdsaMessageSigner { secret_key, secp }
     }
-}
 
-impl<'a> MessageSigner<RecoverableSignature> for EcdsaMessageSigner<'a> {
     fn sign(&self, message: &str) -> RecoverableSignature {
         let msg_hash = signed_msg_hash(message);
         self.secp.sign_ecdsa_recoverable(
@@ -63,9 +49,7 @@ impl<'a> EcdsaMessageSignatureVerifier<'a> {
     pub fn from_address(address: Address, secp: &'a Secp256k1<All>) -> Self {
         EcdsaMessageSignatureVerifier { address, secp }
     }
-}
 
-impl<'a> MessageSignatureVerifier<RecoverableSignature> for EcdsaMessageSignatureVerifier<'a> {
     fn verify(&self, sig: RecoverableSignature, msg: &str) -> bool {
         let message_sig = MessageSignature::new(sig, false);
         let msg_hash = signed_msg_hash(msg);
--- a/src/keys/msgsig.rs
+++ b/src/keys/msgsig.rs
@@ -1,0 +1,128 @@
+
+/// Signing arbitrary messages and verify message signatures
+
+use bitcoin::hashes::Hash;
+use bitcoin::secp256k1::recovery::RecoverableSignature;
+use bitcoin::secp256k1::{All, Message, Secp256k1, SecretKey};
+use bitcoin::util::misc::{signed_msg_hash, MessageSignature};
+use bitcoin::{Address, Network, PrivateKey, PublicKey};
+
+/// Trait for message signers
+pub trait MessageSigner<S> {
+    /// Sign an arbitrary message
+    fn sign(&self, msg: &str) -> S;
+}
+
+/// Trait for message signature verifiers
+pub trait MessageSignatureVerifier<S> {
+    /// Verify a signature against an arbitrary message
+    fn verify(&self, sig: S, msg: &str) -> bool;
+}
+
+/// A message signer using ECDSA
+pub struct EcdsaMessageSigner {
+    secp: Secp256k1<All>,
+    secret_key: SecretKey,
+}
+
+impl EcdsaMessageSigner {
+
+    /// Creates message signer from a bitcoin ECDSA private key
+    pub fn from_prv(prv: PrivateKey) -> Self {
+        Self::from_secret_key(prv.key)
+    }
+
+    /// Creates message signer from an ECDSA secret key
+    pub fn from_secret_key(secret_key: SecretKey) -> Self {
+        EcdsaMessageSigner {
+            secret_key,
+            secp: Secp256k1::new(),
+        }
+    }
+}
+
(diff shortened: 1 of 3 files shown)


src/keys/msgsig.rs
```diff
@@ -0,0 +1,128 @@
...
+/// Signing arbitrary messages and verify message signatures
+
+use bitcoin::hashes::Hash;
+use bitcoin::secp256k1::recovery::RecoverableSignature;
+use bitcoin::secp256k1::{All, Message, Secp256k1, SecretKey};
+use bitcoin::util::misc::{signed_msg_hash, MessageSignature};
+use bitcoin::{Address, Network, PrivateKey, PublicKey};
+
+/// Trait for message signers
+pub trait MessageSigner<S> {
```
afilini: Are there any other signature types you'd plan to support on top of `RecoverableSignature`? I don't really understand why you need the generic here and on the verifier

src/keys/msgsig.rs
```diff
@@ -0,0 +1,128 @@
...
+        self.secp.sign_recoverable(
+            &Message::from_slice(&msg_hash.into_inner()[..]).unwrap(),
+            &self.secret_key,
+        )
+    }
+}
+
+/// A message signature verifier using ECDSA
+pub struct EcdsaMessageSignatureVerifier {
+    secp: Secp256k1<All>,
```
afilini: I would just take a reference here to avoid cloning the context. If you don't want to add a lifetime to the struct you could just change the api of `verify()` to take a secp argument

afilini: I mean, change the member declaration to be a reference and then take one in `from_pub` and `from_address`

src/keys/msgsig.rs
```diff
@@ -0,0 +1,128 @@
...
+            secp: Secp256k1::new(),
+        }
+    }
+}
+
+impl MessageSigner<RecoverableSignature> for EcdsaMessageSigner {
+    fn sign(&self, message: &str) -> RecoverableSignature {
+        let msg_hash = signed_msg_hash(message);
+        self.secp.sign_recoverable(
+            &Message::from_slice(&msg_hash.into_inner()[..]).unwrap(),
```
afilini: Either replace the unwrap with an expect (for reasonably unlikey errors) or change this function to return a `Result`

src/keys/msgsig.rs
```diff
@@ -0,0 +1,128 @@
...
+
+/// Trait for message signature verifiers
+pub trait MessageSignatureVerifier<S> {
+    /// Verify a signature against an arbitrary message
+    fn verify(&self, sig: S, msg: &str) -> bool;
+}
+
+/// A message signer using ECDSA
+pub struct EcdsaMessageSigner {
+    secp: Secp256k1<All>,
```
afilini: Same thing about secp as for the verifier

afilini (changes requested):
Overall this is a good start but it feels a bit disconnected from everything else in BDK.

There are a few ways to "connect" this to the Wallet, the easiest one is probably to iterate over the signers of a wallet and taking their key (if any) with [`descriptor_secret_key()`](https://docs.rs/bdk/0.18.0/bdk/wallet/signer/trait.Signer.html#method.descriptor_secret_key).

You could have an API exposed on the wallet that takes a message and a derivation index, and internally iterates on the signers, derives the keys to the right index and then signs the message.

You should also figure out a good way to handle this when you have multiple keys in your descriptor. You could return a list of signatures, or a map with key -> sig entries, I don't know.. There are a few ways to do it and I haven't thought about them in detail yet, I'd like to see your proposal since you've given this more thoughts than myself!

tunikum:
Thanks for the review. The additions in this PR are intentionally disconnected from the rest of BDK for single responsibility reasons but also because I was not sure if / how you want it to be connected - especially regarding multi-sig wallets and other signing algorithms.

So I chose to add the traits `MessageSigner` and `MessageSignatureVerifier`. With this I wanted to prevent that the wallet struct has to be extended with another method for each signing scheme / algorithm which is added (e.g. `sign_msg_ecdsa`, `sign_msg_schnorr`,…). Maybe this is just me having too much of a non-rust way of thinking?
Nevertheless, this is a design decision which I think should be made by the maintainers of BDK. 

I intended to join your dev call yesterday to discuss this. However, I didn’t manage to attend due to a change in my schedule…

src/keys/msgsig.rs
tunikum: Yes, exactly. The idea behind this is to add other signature types.

src/keys/msgsig.rs
tunikum: I agree, `secp` should at least be changed to a reference. However, regarding a `secp` argument at `verify()`, this would force BDK to have new methods on the wallet struct for every signing / verification algorithm.

src/keys/msgsig.rs
afilini: As far as I am aware there are no plans to switch to schnorr signatures for Bitcoin signed messages, so I don't think there's really too much room for upgrades here.

I was asking if you had at least one specific use case in memory (like a BIP which uses other sig algorithms) but if you don't have any examples I would just say it's better to drop the generic and stick to the classic ecdsa algorithm.

src/keys/msgsig.rs
afilini: I don't understand what do you mean here, can you elaborate a bit more?

My interpretation is that`verify()` doesn't have to be a method on the wallet, because you don't need anything from the wallet to verify a message (while for signing you need the private keys)

src/keys/msgsig.rs
tunikum: No, I had no other use cases in mind than multi-sig and possibly schonrr. But since the latter is not planned and multi-sig can be done with a map as suggested by you, I agree it's better to just add a sign method to the wallet.

src/keys/msgsig.rs
tunikum: I intially wanted to have similar interfaces for signing and verification. Therfore, I thought if signing is moved to the wallet, so should also verification. Combined with my assumption of having schnorr for message signatures at some point, BDK would end up with multiple `verify` methods.

However, this is not the case since schnorr is aparently not planned. Moreover, you  are right that `verify` does not have to be on the wallet - except you want verification to happen with the pubkey of a different `pk(...)` wallet. But this feels a little odd.

Therefore, I suggest that we change the PR to have 
* a method on the wallet for message signing which uses all private keys of the wallet
* message signature verification which is decoupled from the wallet

LLFourn:
Hey @q-src. I noticed that you didn't implement [BIP-137](https://github.com/bitcoin/bips/blob/master/bip-0137.mediawiki) or [BIP-322](https://github.com/bitcoin/bips/blob/master/bip-0322.mediawiki) are they not applicable to what you want to do?

It's not clear this functionality needs a trait around it in BDK. How did you imagine using these traits?

src/keys/msgsig.rs
gcomte: https://github.com/bitcoindevkit/bdk/pull/601/commits/fb772ed44768a899c82083bc38ed19d1393b606a ✔️

src/keys/msgsig.rs
gcomte: `secp` as a reference: https://github.com/bitcoindevkit/bdk/pull/601/commits/707f6c3de09a8e1d3cf2b8cd6401724f7490104f

src/keys/msgsig.rs
gcomte: `secp` as a reference: https://github.com/bitcoindevkit/bdk/pull/601/commits/707f6c3de09a8e1d3cf2b8cd6401724f7490104f

tunikum:
@LLFourn I was not aware of BIP-322 and therefore did not implement it. However, it looks very much like what I want to achieve. 

Apparently, BIP-322 is being added to rust-miniscript (https://github.com/rust-bitcoin/rust-miniscript/pull/444). Shall we wait until this is done and then use the miniscript implementation?

LLFourn:
@q-src TBH miniscript doesn't sound like the right place either. It might make sense to have a BIP322 implementation in `bdk` if it is supported somehow by the existing components of the library. I'm not sure if it is though.

PR #604, opened 2022-05-09
unpinning dependency tokio to just 1

ulrichard:
### Description

unpinning dependency tokio to just 1

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ bitcoincore-rpc = { version = "0.15", optional = true }
 
 # Platform-specific dependencies
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-tokio = { version = "~1.14", features = ["rt"] }
+tokio = { version = "1", features = ["rt"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 async-trait = "0.1"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - added `OldestFirstCoinSelection` impl to `CoinSelectionAlgorithm`
 - New MSRV set to `1.56`
+- Unpinned tokio to `1`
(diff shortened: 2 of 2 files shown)


ulrichard:
Or should it be version = "1" like it was before 5698c683c6f8543379ddf15dc24338fe4c86e7cc?

klochowicz:
> Or should it be version = "1" like it was before [5698c68](https://github.com/bitcoindevkit/bdk/commit/5698c683c6f8543379ddf15dc24338fe4c86e7cc)?

(unsolicited drive-by comment):
IMHO unpinning seems like a better solution as the main reason for pinning is not relevant anymore, since bdk bumped MSRV recently:
https://github.com/bitcoindevkit/bdk/commit/cca69481eb7edb28532a4ed78185248d410050ac

afilini:
Yeah agreed, just `1` is probably gonna work for a while now that we've caught up with the MSRV

Cargo.toml
```diff
@@ -42,7 +42,7 @@ bitcoincore-rpc = { version = "0.15", optional = true }
 
 # Platform-specific dependencies
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-tokio = { version = "~1.14", features = ["rt"] }
```
notmandatory, line 45: we need the `~` for a `1` major version since the range of allowed are the [same](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio).

Cargo.toml
```diff
@@ -42,7 +42,7 @@ bitcoincore-rpc = { version = "0.15", optional = true }
...
+tokio = { version = "~1", features = ["rt"] }
```
notmandatory: we (edit: don't) need the `~` for a `1` major version since the range of allowed are the [same](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio).

notmandatory (approved):
ACK 939a1156c65d0f347cb4397e0883573737d39761

PR #605, opened 2022-05-12
Fix sqlite database set_utxo to insert or update utxos

### LINKED ISSUE #591: SqliteDatabase set_utxo always inserts new utxo
**Describe the bug**  
When `set_utxo` is called the intention is to `insert_or_update` based on the outpoint.  The current implementation always inserts a new utxo and there is no unique constraint on the table.

**To Reproduce**  
Call `set_utxo` with the same &LocalUtxo twice.

**Expected behavior**  
When set_utxo is called twice with the same LocalUtxo there should only be a single utxo in the database.

**Additional context**  
This probably should use a unique index but it was created as just a regular index.  I assume we'll need to add new migration to handle it.  The problem is that adding the unique constraint can fail because people might already have databases with multiple utxos with the same outpoint. 

We would need to be able to run a data migration to remove the duplicates which is complicated for two reasons.  First, there's no concept of a data migration at the moment but I guess they are currently just sql statements.  Second, there's no way to know which one's to keep or delete, right? I guess is there ever a case where it would be "updating" a utxo? Is that even possible?

notmandatory:
### Description

This PR fixes #591 by:
1. Add sqlite `MIGRATIONS` statements to remove duplicate utxos and add unique utxos index on txid and vout.
2. Do an upsert (if insert fails update) instead of an insert in `set_utxo()`.
3. Update database::test::test_utxo to also verify `set_utxo()` doesn't insert duplicate utxos.

### Notes to the reviewers

I verified the updated `test_utxo` fails as expected before my fix and passes after the fix. I tested the new migrations using the below `bdk-cli` command and a manually updated sqlite db with duplicate utxos.
```shell
cargo run --no-default-features --features cli,sqlite-db,esplora-ureq -- wallet -w test1 --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
```

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/database/sqlite.rs
+++ b/src/database/sqlite.rs
@@ -41,6 +41,16 @@ static MIGRATIONS: &[&str] = &[
     "INSERT INTO transaction_details SELECT txid, timestamp, received, sent, fee, height FROM transaction_details_old;",
     "DROP TABLE transaction_details_old;",
     "ALTER TABLE utxos ADD COLUMN is_spent;",
+    // drop all data due to possible inconsistencies with duplicate utxos, re-sync required
+    "DELETE FROM checksums;",
+    "DELETE FROM last_derivation_indices;",
+    "DELETE FROM script_pubkeys;",
+    "DELETE FROM sync_time;",
+    "DELETE FROM transaction_details;",
+    "DELETE FROM transactions;",
+    "DELETE FROM utxos;",
+    "DROP INDEX idx_txid_vout;",
+    "CREATE UNIQUE INDEX idx_utxos_txid_vout ON utxos(txid, vout);"
 ];
 
 /// Sqlite database stored on filesystem
@@ -86,7 +96,7 @@ impl SqliteDatabase {
         script: &[u8],
         is_spent: bool,
     ) -> Result<i64, Error> {
-        let mut statement = self.connection.prepare_cached("INSERT INTO utxos (value, keychain, vout, txid, script, is_spent) VALUES (:value, :keychain, :vout, :txid, :script, :is_spent)")?;
+        let mut statement = self.connection.prepare_cached("INSERT INTO utxos (value, keychain, vout, txid, script, is_spent) VALUES (:value, :keychain, :vout, :txid, :script, :is_spent) ON CONFLICT(txid, vout) DO UPDATE SET value=:value, keychain=:keychain, script=:script, is_spent=:is_spent")?;
         statement.execute(named_params! {
             ":value": value,
             ":keychain": keychain,
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -90,12 +90,16 @@ jobs:
       matrix:
         blockchain:
           - name: electrum
+            testprefix: blockchain::electrum::test
             features: test-electrum,verify
           - name: rpc
+            testprefix: blockchain::rpc::test
             features: test-rpc
           - name: esplora
+            testprefix: esplora
             features: test-esplora,use-esplora-reqwest,verify
           - name: esplora
+            testprefix: esplora
             features: test-esplora,use-esplora-ureq,verify
     steps:
       - name: Checkout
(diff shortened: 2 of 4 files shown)


src/database/sqlite.rs
```diff
@@ -41,6 +41,9 @@ static MIGRATIONS: &[&str] = &[
     "INSERT INTO transaction_details SELECT txid, timestamp, received, sent, fee, height FROM transaction_details_old;",
     "DROP TABLE transaction_details_old;",
     "ALTER TABLE utxos ADD COLUMN is_spent;",
+    "DELETE FROM utxos WHERE rowid > (SELECT MIN(rowid) FROM utxos u2 WHERE utxos.txid = u2.txid AND utxos.vout = u2.vout);",
```
danielabrozzoni: I discussed a bit with @afilini about this, while theoretically the duplicated lines should contain the same info, we can't be so sure about that...
If the duplicated lines contain different info (for example: same `outpoint`, different `is_spent`), we might end up in an inconsistent state.
Alekos suggested wiping the entire db.
I suggested wiping it only if we find some duplicated outpoints with the different info - if this can be done in a single SQL query (should be doable, but I'm not entirely sure). What do you think?

danielabrozzoni: I discussed a bit with @afilini about this, while theoretically the duplicated lines should contain the same info, we can't be so sure about that...
If the duplicated lines contain different info (for example: same `outpoint`, different `is_spent`), we might end up in an inconsistent state.
Alekos suggested wiping the entire db.
I suggested wiping it only if we find some duplicated outpoints with the different info - if this can be done in a single SQL query (should be doable, but I'm not entirely sure). What do you think?

notmandatory: But I'm so proud of my little SQL de-duplicating! haha, anyway, rather than risk an inconsistent state I'll update this to always wipe the database, it's only a cache after all. And I have to agree with @afilini to wipe the data even if there are no dups since either way we will need to warn users that it MIGHT be wiped on this release and they'll need to be prepared to handle it with a sync() (and any reasonable app will do a sync at startup).

notmandatory:
I force-pushed 154b10b to DELETE all data from tables (except for script version table). Tested locally with bdk-cli and verified MIGRATIONS scripts all run and able to re-sync after. I also added a warning in the CHANGELOG that anyone using the `sqlite-db` feature will need to do a wallet.sync().

notmandatory:
This is now ready to go, I also added a fix for our blockchain integration tests in CI that haven't been running. 

Once this PR is merge I'm going to cherry pick it into a `0.18.1` fix release since @thunderbiscuit's SoB mentee found that our latest bdk-kotlin (and also I'll guess bdk-swift) release is broken, duplicating balances, because they use the SQLite DB.

danielabrozzoni (approved):
utACK 35feb107ed5969720ce54a6aa76b7b2176f6c7c1 - Code looks good, but I didn't do any local test to see if the db gets wiped

notmandatory:
Thanks for the utACK @danielabrozzoni, I'm going to merge now, should be good based on my manually testing, as discussed on the call this is what I did with the unreleased master branch version of `bdk-cli`:
1. Sync'd with dependency `bdk` 0.18.0 and new `sqlite-db` feature enabled:
```shell
cargo run --no-default-features --features cli,electrum,sqlite-db -- wallet --descriptor "wpkh(tpubEBr4i6yk5nf5DAaJpsi9N2pPYBeJ7fZ5Z9rmN4977iYLCGco1VyjB9tvvuvYtfZzjD5A8igzgw3HeWeeKFmanHYqksqZXYXGsw5zjnj7KM9/*)" sync
```
2. Manually checked sqlite DB file tables, verified old schema is there with utxo data, I used database tool in AndroidStudio but can also use `sqlite` cli tool
3. With same build of `bdk-cli` got wallet balance, sync'd wallet again and checked balance again, show's double utxos
4. Changed `bdk-cli` `Cargo.toml` dependency for `bdk` to this PR branch version
5. Re-sync'd with `bdk-cli` and verified correct balance
6. Manually checked sqlite DB tables again, verified new schema is there with new utxo unique constraints and no duplicates

danielabrozzoni:
Post-merge ACK 35feb107ed5969720ce54a6aa76b7b2176f6c7c1

PR #610, opened 2022-05-23
Populate the redeemScript for `sh(wsh(sortedmulti()))`

### LINKED ISSUE #609: Some PSBTs from BDK don't load in Electrum
**Describe the bug**  
When I try to load some PSBTs from BDK in electrum, I get the following error:
PSBTInputConsistencyFailure('PSBT input validation: If a witnessScript is provided, the scriptPubKey must be for that witnessScript')
First I suspect that it's just a flag in TxBuilder that I was missing. But include_output_redeem_witness_script didn't help, and for add_global_xpubs I would need to figure out how to avoid MissingKeyOrigin.
What's important is, that it all works nicely with "sh(" or "wsh(" but the problem occurs with "sh(wsh("

**To Reproduce**  
Download the attached example script, remove the .txt extension and run it:
[psbt_electrum.sh.txt](https://github.com/bitcoindevkit/bdk/files/8755912/psbt_electrum.sh.txt)
Try to open the resulting PSBT in Electrum

**Expected behavior**  
Electrum can load the PSBT and displays the information.

afilini:
### Description

Also explicitly match all the individual variants to ensure a similar problem
doesn't happen again.

Fixes #609

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -335,7 +335,12 @@ impl<'s> DescriptorScripts for DerivedDescriptor<'s> {
             DescriptorType::Sh => Some(self.explicit_script().unwrap()),
             DescriptorType::Bare => Some(self.explicit_script().unwrap()),
             DescriptorType::ShSortedMulti => Some(self.explicit_script().unwrap()),
-            _ => None,
+            DescriptorType::ShWshSortedMulti => Some(self.explicit_script().unwrap().to_v0_p2wsh()),
+            DescriptorType::Pkh
+            | DescriptorType::Wpkh
+            | DescriptorType::Tr
+            | DescriptorType::Wsh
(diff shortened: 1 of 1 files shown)


notmandatory:
It looks like this fix includes the TR type so it must be dependent on `rust-bitcoin` 0.28 and `rust-miniscript` 7.0, so that means we if we wanted to release it in a fix release we'd need some small change to work with the older versions of those dependencies right? Since @ulrichard isn't held up waiting for this I'm now leaning towards releasing it with the next regular release instead of a `0.18.1` fix release.

notmandatory (approved):
ACK 82de8b50da6e8b9752332b80da395c792c371def

afilini:
Yeah that's true, he said it's not blocking on discord so I would just go ahead and merge this.

Thanks for the review!

PR #614, opened 2022-05-25
Avoid using immature coinbase inputs

### LINKED ISSUE #413: BDK will select coinbase inputs that are not matured yet

danielabrozzoni:
### Description

With this PR we start considering how many confirmations a coinbase has. If it's not mature yet, we don't use it for building transactions. 
Fixes #413

### Notes to the reviewers

This PR is based on #611, review that one before reviewing this 😄 

007c5a78335a3e9f6c9c28a077793c2ba34bbb4e adds a coinbase parameter to `populate_test_db`, to specify if you want the db to be populated with immature coins. This is useful for `test_spend_coinbase`, but that's probably going to be the only use case.
I don't think it's a big deal to have a test function take an almost_always_useless parameter - it's not an exposed API, anyways. But, if you can come up with a different way of implementing `test_spend_coinbase` that doesn't require 007c5a78335a3e9f6c9c28a077793c2ba34bbb4e, even better! I looked for it for a while, but other than duplicating the whole `populate_test_db` code, which made the test way harder to comprehend, I didn't find any other way.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/database/memory.rs
+++ b/src/database/memory.rs
@@ -486,15 +486,23 @@ impl ConfigurableDatabase for MemoryDatabase {
 /// don't have `test` set.
 macro_rules! populate_test_db {
     ($db:expr, $tx_meta:expr, $current_height:expr$(,)?) => {{
+        $crate::populate_test_db!($db, $tx_meta, $current_height, (@coinbase false))
+    }};
+    ($db:expr, $tx_meta:expr, $current_height:expr, (@coinbase $is_coinbase:expr)$(,)?) => {{
         use std::str::FromStr;
         use $crate::database::BatchOperations;
         let mut db = $db;
         let tx_meta = $tx_meta;
         let current_height: Option<u32> = $current_height;
+        let input = if $is_coinbase {
+            vec![$crate::bitcoin::TxIn::default()]
+        } else {
+            vec![]
+        };
         let tx = $crate::bitcoin::Transaction {
             version: 1,
             lock_time: 0,
-            input: vec![],
+            input,
             output: tx_meta
                 .output
                 .iter()
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -73,6 +73,7 @@ use crate::testutils;
 use crate::types::*;
 
 const CACHE_ADDR_BATCH_SIZE: u32 = 100;
+const COINBASE_MATURITY: u32 = 100;
 
 /// A Bitcoin wallet
 ///
@@ -765,6 +766,7 @@ where
             params.drain_wallet,
             params.manually_selected_only,
             params.bumping_fee.is_some(), // we mandate confirmed transactions if we're bumping the fee
+            current_height,
         )?;
 
         let coin_selection = coin_selection.coin_select(
@@ -1335,6 +1337,7 @@ where
     /// Given the options returns the list of utxos that must be used to form the
     /// transaction and any further that may be used if needed.
     #[allow(clippy::type_complexity)]
+    #[allow(clippy::too_many_arguments)]
     fn preselect_utxos(
         &self,
         change_policy: tx_builder::ChangeSpendPolicy,
@@ -1343,6 +1346,7 @@ where
         must_use_all_available: bool,
         manual_only: bool,
         must_only_use_confirmed_tx: bool,
+        current_height: Option<u32>,
     ) -> Result<(Vec<WeightedUtxo>, Vec<WeightedUtxo>), Error> {
         //    must_spend <- manually selected utxos
         //    may_spend  <- all other available utxos
@@ -1361,23 +1365,44 @@ where
             return Ok((must_spend, vec![]));
         }
 
-        let satisfies_confirmed = match must_only_use_confirmed_tx {
-            true => {
-                let database = self.database.borrow();
-                may_spend
-                    .iter()
-                    .map(|u| {
-                        database
-                            .get_tx(&u.0.outpoint.txid, true)
-                            .map(|tx| match tx {
-                                None => false,
-                                Some(tx) => tx.confirmation_time.is_some(),
-                            })
+        let database = self.database.borrow();
+        let satisfies_confirmed = may_spend
+            .iter()
+            .map(|u| {
+                database
+                    .get_tx(&u.0.outpoint.txid, true)
+                    .map(|tx| match tx {
+                        // We don't have the tx in the db for some reason,
+                        // so we can't know for sure if it's mature or not.
+                        // We prefer not to spend it.
+                        None => false,
+                        Some(tx) => {
+                            // Whether the UTXO is mature and, if needed, confirmed
+                            let mut spendable = true;
+                            if must_only_use_confirmed_tx && tx.confirmation_time.is_none() {
+                                return false;
+                            }
+                            if tx
+                                .transaction
+                                .expect("We specifically ask for the transaction above")
+                                .is_coin_base()
+                            {
+                                if let Some(current_height) = current_height {
+                                    match &tx.confirmation_time {
+                                        Some(t) => {
+                                            // https://github.com/bitcoin/bitcoin/blob/c5e67be03bb06a5d7885c55db1f016fbf2333fe3/src/validation.cpp#L373-L375
+                                            spendable &= (current_height.saturating_sub(t.height))
+                                                >= COINBASE_MATURITY;
+                                        }
+                                        None => spendable = false,
+                                    }
+                                }
+                            }
+                            spendable
+                        }
                     })
-                    .collect::<Result<Vec<_>, _>>()?
-            }
-            false => vec![true; may_spend.len()],
-        };
+            })
+            .collect::<Result<Vec<_>, _>>()?;
 
         let mut i = 0;
         may_spend.retain(|u| {
@@ -4642,5 +4667,71 @@ pub(crate) mod test {
             sighash as u8,
             "The signature should have been made with the right sighash"
         );
+    }
+
+    #[test]
+    fn test_spend_coinbase() {
+        let descriptors = testutils!(@descriptors (get_test_wpkh()));
+        let wallet = Wallet::new(
+            &descriptors.0,
+            None,
+            Network::Regtest,
+            AnyDatabase::Memory(MemoryDatabase::new()),
+        )
+        .unwrap();
+
+        let confirmation_time = 5;
+
+        crate::populate_test_db!(
+            wallet.database.borrow_mut(),
+            testutils! (@tx ( (@external descriptors, 0) => 25_000 ) (@confirmations 0)),
+            Some(confirmation_time),
+            (@coinbase true)
+        );
+
+        let not_yet_mature_time = confirmation_time + COINBASE_MATURITY - 1;
+        let maturity_time = confirmation_time + COINBASE_MATURITY;
+
+        // The balance is nonzero, even if we can't spend anything
+        // FIXME: we should differentiate the balance between immature,
+        // trusted, untrusted_pending
+        // See https://github.com/bitcoindevkit/bdk/issues/238
+        let balance = wallet.get_balance().unwrap();
+        assert!(balance != 0);
+
+        // We try to create a transaction, only to notice that all
+        // our funds are unspendable
+        let addr = Address::from_str("2N1Ffz3WaNzbeLFBb51xyFMHYSEUXcbiSoX").unwrap();
+        let mut builder = wallet.build_tx();
+        builder
+            .add_recipient(addr.script_pubkey(), balance / 2)
+            .set_current_height(confirmation_time);
+        assert!(matches!(
+            builder.finish().unwrap_err(),
+            Error::InsufficientFunds {
+                needed: _,
+                available: 0
+            }
+        ));
+
+        // Still unspendable...
+        let mut builder = wallet.build_tx();
+        builder
+            .add_recipient(addr.script_pubkey(), balance / 2)
+            .set_current_height(not_yet_mature_time);
+        assert!(matches!(
+            builder.finish().unwrap_err(),
+            Error::InsufficientFunds {
+                needed: _,
+                available: 0
+            }
+        ));
+
+        // ...Now the coinbase is mature :)
+        let mut builder = wallet.build_tx();
+        builder
+            .add_recipient(addr.script_pubkey(), balance / 2)
+            .set_current_height(maturity_time);
+        builder.finish().unwrap();
     }
 }
--- a/src/wallet/tx_builder.rs
+++ b/src/wallet/tx_builder.rs
@@ -547,10 +547,15 @@ impl<'a, D: BatchDatabase, Cs: CoinSelectionAlgorithm<D>, Ctx: TxBuilderContext>
 
     /// Set the current blockchain height.
     ///
-    /// This will be used to set the nLockTime for preventing fee sniping. If the current height is
-    /// not provided, the last sync height will be used instead.
-    ///
+    /// This will be used to:
+    /// 1. Set the nLockTime for preventing fee sniping.
     /// **Note**: This will be ignored if you manually specify a nlocktime using [`TxBuilder::nlocktime`].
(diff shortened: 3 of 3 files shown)


src/wallet/mod.rs
```diff
@@ -1339,23 +1359,42 @@ where
...
+                        // We don't have the tx in the db for some reason,
+                        // so we can't know for sure if it's mature or not.
+                        // We prefer not to spend it.
+                        None => false,
+                        Some(tx) => {
+                            let mut confirmed = true;
+                            if tx
+                                .transaction
+                                .expect("We specifically ask for the transaction above")
+                                .is_coin_base()
```
rajarshimaitra: I feel like this part of the code can be simplified a little?? 

Instead of storing `satisfied_confirmed` in a `Vec<bool>` then removing the `may_spend` utxos as per the index of those satisfactions, why not just remove all the immature coinbase utxos from the `may_spend`  in just one single list operation?? Could make the intention of the logic much clearer..

Also what happens when the user manually provides and immature coinbase??

I think the nicest way would be to 
 - have lambda that checks for maturity given an utxo..
 - List out `may_spend` and `must_spend` where that maturity satisfaction fails.
 -  Error early if such failure occures in user given `must_spend`.

src/database/memory.rs
```diff
@@ -486,15 +486,23 @@ impl ConfigurableDatabase for MemoryDatabase {
...
         use std::str::FromStr;
         use $crate::database::BatchOperations;
         let mut db = $db;
         let tx_meta = $tx_meta;
         let current_height: Option<u32> = $current_height;
+        let input = if $is_coinbase {
+            vec![$crate::bitcoin::TxIn::default()]
+        } else {
+            vec![]
+        };
```
rajarshimaitra, line 501: Can't we just have a pure function here than a macro?? 😭😭😭

Life is already difficult enough.. 😂😂

rajarshimaitra (commented):
Concept ACK..

I have a comment on the selection logic..

src/database/memory.rs
danielabrozzoni: I just added a parameter to the `populate_test_db` macro to create coinbase inputs, should I duplicate the whole populate test db code just for using a function? ...

src/wallet/mod.rs
danielabrozzoni: > Instead of storing satisfied_confirmed in a Vec<bool> then removing the may_spend utxos as per the index of those satisfactions, why not just remove all the immature coinbase utxos from the may_spend in just one single list operation?? Could make the intention of the logic much clearer..

It was mostly because I didn't want to modify the old `satisfies_confirmed` logic but sure, I can change it 

> Also what happens when the user manually provides and immature coinbase??

I strongly disagree with removing the immature coinbase inputs from the `must_spend`, as the list is intended for the utxos that we **must** spend. The API is intended for advanced users that know what they need - and it's up to them to check if that certain utxo can even be spent (we don't even check if it's spent already or not, for example).

rajarshimaitra: > as the list is intended for the utxos that we must spend. 

I agree with that.. Just a flow thought that came while figuring the list operations.. Yes we should not remove it from there if that's the API intention.. 

But then again advanced users "not knowing" what they are doing, and spending an immature coinbase would never get them into blockchain as it is consensus enforced.. Will not cause fund loss though.. But no hard opinions one way or other.. If someone wanna do that, might have some good reason for it..  

Another way would be to modify the `get_available_utxos()` function, to not return immature coinbase.. This function is only used in `may_spend` extraction, and nowhere else.. So might make sense to enforce the behavior at that level itself..   

> It was mostly because I didn't want to modify the old satisfies_confirmed logic but sure, I can change it

Yes I think it gives us a good opportunity to clean up this logic here.. Unless there is something more I am missing.. But I think thats all this logic is doing..

src/database/memory.rs
rajarshimaitra: I wonder if there's a way to do this recursively.. Let me see if I can do it..
Feel free to ignore for now, and I will try to do it in a followup PR..

src/wallet/mod.rs
```diff
@@ -1339,23 +1359,42 @@ where
...
+                            {
+                                if let Some(current_height) = current_height {
+                                    if let Some(tx_conf_time) = &tx.confirmation_time {
+                                        // https://github.com/bitcoin/bitcoin/blob/c5e67be03bb06a5d7885c55db1f016fbf2333fe3/src/validation.cpp#L373-L375
+                                        confirmed &= current_height - tx_conf_time.height
+                                            >= COINBASE_MATURITY
```
vladimirfomene: Nits: shouldn't there be a semicolon after this statement.

vladimirfomene (commented):
Pulled changes, tested and reviewed code. I found it hard to review the macro part because I'm not very comfortable with macros. Tested ACK

src/wallet/mod.rs
danielabrozzoni: Right, thanks!

danielabrozzoni:
Rebased, and also updated `set_current_height` documentation

src/wallet/mod.rs
danielabrozzoni: > But then again advanced users "not knowing" what they are doing, and spending an immature coinbase would never get them into blockchain as it is consensus enforced.. Will not cause fund loss though.. But no hard opinions one way or other.. If someone wanna do that, might have some good reason for it..

I think it's reasonable to assume that whoever's using an advance API knows what they're doing. Also, they might want to presign a transaction and then hold it for a while, and so they wouldn't care about coinbase maturity.

> Another way would be to modify the get_available_utxos() function, to not return immature coinbase.. This function is only used in may_spend extraction, and nowhere else.. So might make sense to enforce the behavior at that level itself..

That would probably be slightly clearer, but it would require an additional db call (now we retrieve the tx once, and we use it to check if the tx is confirmed and if the inputs are mature)

What do you think of this? Is it slightly more readable? I'm sadly discarding every error `get_tx` throws at me, as I can't have `?` inside retain
```diff
         let database = self.database.borrow();
-        let satisfies_confirmed = may_spend
-            .iter()
-            .map(|u| {
-                database
-                    .get_tx(&u.0.outpoint.txid, true)
-                    .map(|tx| match tx {
-                        // We don't have the tx in the db for some reason,
-                        // so we can't know for sure if it's mature or not.
-                        // We prefer not to spend it.
-                        None => false,
-                        Some(tx) => {
-                            let mut confirmed = true;
-                            if tx
-                                .transaction
-                                .expect("We specifically ask for the transaction above")
-                                .is_coin_base()
-                            {
-                                if let Some(current_height) = current_height {
-                                    if let Some(tx_conf_time) = &tx.confirmation_time {
-                                        // https://github.com/bitcoin/bitcoin/blob/c5e67be03bb06a5d7885c55db1f016fbf2333fe3/src/validation.cpp#L373-L375
-                                        confirmed &= (current_height - tx_conf_time.height)
-                                            >= COINBASE_MATURITY;
-                                    } else {
-                                        confirmed = false;
-                                    }
-                                }
-                            }
-                            if must_only_use_confirmed_tx {
-                                confirmed &= tx.confirmation_time.is_some();
+
+        let satisfies_confirmed = |u: &LocalUtxo| {
+            database
+                .get_tx(&u.outpoint.txid, true)
+                .ok()
+                .flatten()
+                .map(|tx| {
+                    let mut confirmed = true;
+                    if tx
+                        .transaction
+                        .expect("We specifically ask for the transaction above")
+                        .is_coin_base()
+                    {
+                        if let Some(current_height) = current_height {
+                            if let Some(tx_conf_time) = &tx.confirmation_time {
+                                // https://github.com/bitcoin/bitcoin/blob/c5e67be03bb06a5d7885c55db1f016fbf2333fe3/src/validation.cpp#L373-L375
+                                confirmed &=
+                                    (current_height - tx_conf_time.height) >= COINBASE_MATURITY;
+                            } else {
+                                confirmed = false;
                             }
-                            confirmed
                         }
-                    })
-            })
-            .collect::<Result<Vec<_>, _>>()?;
+                    }
+                    if must_only_use_confirmed_tx {
+                        confirmed &= tx.confirmation_time.is_some();
+                    }
+                    confirmed
+                })
+                .unwrap_or(false)
+        };
 
         let mut i = 0;
         may_spend.retain(|u| {
             let retain = change_policy.is_satisfied_by(&u.0)
                 && !unspendable.contains(&u.0.outpoint)
-                && satisfies_confirmed[i];
+                && satisfies_confirmed(&u.0);
             i += 1;
             retain
         });

**Review by wszdexdrf — APPROVED** [review]
Tested ACK 93ca3b9bc9818011a115c8663fd6948abf1bcf67

File: src/wallet/mod.rs (hunk 1 above)
Comment (rajarshimaitra) [reply src/wallet/mod.rs]: I feel for now its ok to leave it as it.. Maybe in a later PR we can move this all together in `get_available_utxos()` and simplify this logic a bit..

**Review by rajarshimaitra — APPROVED** [review]
tACK 93ca3b9bc9818011a115c8663fd6948abf1bcf67

File: src/wallet/mod.rs [hunk 4]
```diff
@@ -1343,23 +1365,42 @@ where
...
+            .map(|u| {
+                database
+                    .get_tx(&u.0.outpoint.txid, true)
+                    .map(|tx| match tx {
+                        // We don't have the tx in the db for some reason,
+                        // so we can't know for sure if it's mature or not.
+                        // We prefer not to spend it.
+                        None => false,
+                        Some(tx) => {
+                            let mut confirmed = true;
```
afilini: Maybe you should rename this to something else, like "mature"?

src/wallet/mod.rs
```diff
@@ -1343,23 +1365,42 @@ where
...
+                            if tx
+                                .transaction
+                                .expect("We specifically ask for the transaction above")
+                                .is_coin_base()
+                            {
+                                if let Some(current_height) = current_height {
```
afilini, line 1390: I think you can replace the two nested if-lets with an `and_then()`

src/wallet/mod.rs
```diff
@@ -1343,23 +1365,42 @@ where
...
+                                    if let Some(tx_conf_time) = &tx.confirmation_time {
+                                        // https://github.com/bitcoin/bitcoin/blob/c5e67be03bb06a5d7885c55db1f016fbf2333fe3/src/validation.cpp#L373-L375
+                                        confirmed &= (current_height - tx_conf_time.height)
```
afilini: I would use `saturating_sub()` to avoid potential underflows with the subtraction

src/wallet/mod.rs
danielabrozzoni: We agreed in pvt that it's not really doable, I went for a match instead of the second `if let`

src/wallet/mod.rs
danielabrozzoni: Renamed to "spendable" as mature is not really appropriate either (the variable olds whether the utxo is mature and, if needed, confirmed)

src/wallet/mod.rs
```diff
@@ -1361,23 +1365,44 @@ where
...
+                                            // https://github.com/bitcoin/bitcoin/blob/c5e67be03bb06a5d7885c55db1f016fbf2333fe3/src/validation.cpp#L373-L375
+                                            spendable &= (current_height.saturating_sub(t.height))
+                                                >= COINBASE_MATURITY;
+                                        }
+                                        None => spendable = false,
+                                    }
+                                }
+                            }
+                            if must_only_use_confirmed_tx {
+                                spendable &= tx.confirmation_time.is_some();
```
evanlinjin: Would it make sense to have this check before the coinbase check?

This way we can return early if `must_only_use_confirmed_tx && tx.confirmation_time.is_none()`

src/wallet/mod.rs
```diff
@@ -1361,23 +1365,44 @@ where
...
+                        None => false,
+                        Some(tx) => {
+                            // Whether the UTXO is mature and, if needed, confirmed
+                            let mut spendable = true;
+                            if tx
+                                .transaction
+                                .expect("We specifically ask for the transaction above")
+                                .is_coin_base()
+                            {
+                                if let Some(current_height) = current_height {
```
evanlinjin, line 1390: If `current_height` is not explicitly provided, would it make sense to grab it from `Database::get_sync_time`?

src/database/memory.rs
```diff
@@ -486,15 +486,23 @@ impl ConfigurableDatabase for MemoryDatabase {
...
+    ($db:expr, $tx_meta:expr, $current_height:expr, (@coinbase $is_coinbase:expr)$(,)?) => {{
         use std::str::FromStr;
         use $crate::database::BatchOperations;
         let mut db = $db;
         let tx_meta = $tx_meta;
         let current_height: Option<u32> = $current_height;
+        let input = if $is_coinbase {
+            vec![$crate::bitcoin::TxIn::default()]
+        } else {
+            vec![]
```
rajarshimaitra, line 500: Not that its highly important but here the input remains always empty even if its there in `tx_meta`.. This doesn't fail any db tests because we never deal with inputs, but might make sense to get the inputs too from `tx_meta` for sake of completeness.. I have done it in this PR here https://github.com/bitcoindevkit/bdk/pull/624/files

Something like this 
```
        let input = if is_coinbase {
            // `TxIn::default()` creates a coinbase input, by definition.
            vec![TxIn::default()]
        } else {
            tx_meta
                .input
                .iter()
                .map(|test_input| {
                    let mut txin = TxIn::default();
                    txin.previous_output = OutPoint {
                        txid: test_input.txid,
                        vout: test_input.vout,
                    };

                    if let Some(seq) = test_input.sequence {
                        txin.sequence = seq;
                    }
                    txin
                })
                .collect()
        };
  ```
Not blocking on this PR. Can be done in a separate one later..

rajarshimaitra (approved):
ReACK 34069083a9573d11d2318718ede675b7e4fdab29

One non blocking note..

src/wallet/mod.rs
rajarshimaitra: We actually always get the current height from last sync height here if not provided. 
https://github.com/bitcoindevkit/bdk/blob/063d51fd7587815c733f24365a4de02fed072903/src/wallet/mod.rs#L622-L630

So theoretically this `current_height` will never be `None` here I think.. So we can as well use an `u32` instead of `Option<u32>` in `preselect_utxos`.

danielabrozzoni: > So theoretically this current_height will never be None here I think..

It can still be None if we never synced before (`get_sync_time` would return `None`)

rajarshimaitra: Ah yes that makes sense.. When we haven't synced ever it will be None at that time only.. But can't we just translate that None to 0 in that case and get rid of `Option`s all together in this level?? Not sure if it helps too much, but can simplify the code a bit.

src/wallet/mod.rs
afilini: Yeah that's a good point

src/wallet/mod.rs
afilini: I would keep the option here, it feels easier to understand to me

rajarshimaitra: Ya not a biggie.. Resolving this now..

src/database/memory.rs
rajarshimaitra: Resolving this too as its not blocking on this PR..

danielabrozzoni: Good idea, I think we can merge this in #624 :)

src/wallet/mod.rs
danielabrozzoni: Done in e85aa247cb85601e96f4d82b7a996f709559223f

afilini (approved):
ACK e85aa24

Issue #481, opened 2021-11-30
Refactor BDK Integration tests


rajarshimaitra:
**Description**  
  
BDK uses automated [integration testing](https://github.com/bitcoindevkit/bdk/blob/master/src/testutils/blockchain_tests.rs) to use in both local and github CI tests. This ensures the library is working as expected and there are no unexpected changed behaviour. Currently the testing framework is written using rust macros, which makes it a little difficult to parse through. The goal of this project is to use more intuitive rust structures and methods for the test framework. Similar to Boitcoin Core's [test framework](https://github.com/bitcoin/bitcoin/tree/master/test/functional/test_framework). This will help in easier understanding of the tests for new devs and also allow casual contributors to extend the tests. 

**Expected Outcomes**  
   
- Through understanding of integration tests in BDK.
- Good understanding of [rust macros](https://doc.rust-lang.org/book/ch19-06-macros.html).
- A flexible yet comprehensive struct based test framework for BDK, replacing the existing macros.
- Bonus: Add more integration tests into BDK.

**Resources**  
  
  
  
  
  
- [rust macro](https://doc.rust-lang.org/book/ch19-06-macros.html)
- [rust integration test](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html)

**Skills Required**  
  
  
  
  

- Experience with git. [Guide](https://git-scm.com/book/en/v2)
- Basic familiarity with rust. [First seven chapters of the book](https://doc.rust-lang.org/book/)
- Familiarity with BDK’s integration tests.
- Familiarity with rust macro, structure and traits.

**Mentor(s)**  
@rajarshimaitra 

**Difficulty**  : Medium

**Competency Test (optional)**  

- Install rust, compile and run all bdk examples and tests.
- Read through the BDK docs.
- Make a dummy wallet with BDK with different blockchain backends. 
- Familiarity with basic rust, should be able to write basic custom trait implementations on foreign structures.

notmandatory:
Closing this issue as blockchain tests have been rewritten for the `bdk_chain` module.

PR #624, opened 2022-06-06
Refactor testing logic from macros to pure functions

### LINKED ISSUE #719: [bug] get_funded_wallet() does not return balance
**Describe the bug**  
After recent changes in balance calculation, the get_funded_wallet test function will not show any balance because we are not setting the sync_time in the populate_test_db macro..

Our existing tests are not failing, because even if the balance is shown zero, it will still be able to produce transactions. But its causing test failure in downstream bdk-reserves crate where the balances are required to assert scenarios.

**To Reproduce**  
```
let (wallet, _, _) = get_funded_wallet(get_test_wpkh());
assert_eq!(wallet.get_balance().unwrap().confirmed, 0);
```
**Expected behavior**  
```
let (wallet, _, _) = get_funded_wallet(get_test_wpkh());
assert_eq!(wallet.get_balance().unwrap().confirmed, 50000);
```

rajarshimaitra:
### Description
This finishes the Summer of Bitcoin Task  #481 which aimed at refactoring the bvlockchain tests from macros to functions. This allows easier readability of the test logic using any kind of rust analysis tool and allows easy contribution by new contributors. Thanks to SoB particiapant @saikishore222 and @SanthoshAnguluri for the work. 

This PR also refactors the same for database tests and moves the test artifacts into their own `helper` module. 

### Notes to the reviewers

 - I am not sure if this requires changelog update. But Added an entry over there anyway as its quite a big change. 
 - Now that individual blockchain tests are to be specified in each module, please check that I am not missing any test cases. 
 

UPDATE: This PR now includes the fix for #719 . The fix is included here for easier review and to reduce duplicate review works. The last two commits includes the bug fix.

#### All Submissions:


rajarshimaitra:
Converting this into draft as I plan to assign my SoB mentees for further development on top of this PR.. 

@alekos can you provide just a quick concept review or any further thought?? That would be very helpful..

src/testutils/helpers.rs
```diff
@@ -0,0 +1,200 @@
...
+                    },
+                    ..Default::default()
+                };
+
+                if let Some(seq) = test_input.sequence {
+                    txin.sequence = seq;
+                }
+                txin
+            })
+            .collect()
```
danielabrozzoni: This is changing the behavior of the `populate_test_db` function, which is fine, and also needed for #682, but it means that the first commit is not really "move only"

cc @wszdexdrf: can you check that hardware wallet tests pass if you copy-paste these 10 lines in your PR?

src/testutils/helpers.rs
```diff
@@ -0,0 +1,200 @@
...
+use crate::{
+    database::{AnyDatabase, BatchDatabase, BatchOperations, MemoryDatabase},
+    testutils, BlockTime, KeychainKind, LocalUtxo, TransactionDetails, Wallet,
+};
+
+use super::TestIncomingTx;
+
+/// Populate a test database with a `TestIncomingTx`, as if we had found the tx with a `sync`.
+/// This is a hidden function, only useful for `DataBase` unit testing.
+pub fn populate_test_db(
```
danielabrozzoni, line 29: Should this be public?

danielabrozzoni (commented):
I just reviewed the first two commits for now, the changes looks good, but I'd avoid saying that the first commit is "move only": you're also changing the logic (see `txin` comment), and refactoring populate_test_db into a function.

It might be easier to review if the first commit really was move-only, as we could use `--color-moved=dimmed-zebra`, but I understand that it might be a bit of a hassle...

src/testutils/helpers.rs
rajarshimaitra: Yeah.. That's my bad.. Its not move only.. Updated the commit message..

src/testutils/helpers.rs
rajarshimaitra: Hmm.. Let me see what breaks if it isn't..

Update: It does have to be public to be used in the `doctest_wallet` macro.. Compiler screams otherwise.. `pub(crate)` doesn't work either..

rajarshimaitra:
Thanks @danielabrozzoni for the quick look.. I agree the first isn't move only anymore.. It was when I first made the commit.. Updated the commit message.. Also changed both the helper functions to private..

rajarshimaitra:
Thanks for the look @danielabrozzoni ..

> It might be easier to review if the first commit really was move-only, as we could use --color-moved=dimmed-zebra, but I understand that it might be a bit of a hassle...

It didn't made much sense to just do a move and then change logic.. And as I am making it from macro to function, it won't be a pure move anyway even without the logic change.. Might get more confusing than what it is to review.. Reverted the commit message though..

src/testutils/helpers.rs
wszdexdrf: This doesn't work with #682. Trezor gives a broken psbt (as before), and Ledger returns the error `IndexError: index out of range`

danielabrozzoni: I'll investigate then :)

src/testutils/helpers.rs
afilini: We can at least hide it from the docs

src/testutils/helpers.rs
```diff
@@ -0,0 +1,200 @@
...
+    let confirmation_time = if let Some(conf_time) = tx_meta.min_confirmations {
+        current_height
+            .checked_sub(conf_time as u32)
+            .map(|conf| BlockTime {
+                height: conf,
+                timestamp: 0,
+            })
+    } else {
+        None
+    };
```
afilini: I think this is very different from what we had before:

- In the macro if `min_confirmations` is zero we map that to `None` immediately
- The macro had an `unwrap()` on the `checked_sub()`. I understand it's better to handle errors, but in this case we want to fail if the height is too low for a given number of confirmations. This code instead silently makes the tx unconfirmed because it maps that case to `None`
- There's a `+1` missing which was present in the macro

src/wallet/mod.rs
```diff
@@ -1743,6 +1740,12 @@ where
         self.database.borrow()
     }
 
+    #[cfg(test)]
+    /// Return an mutable reference to the internal database
+    pub(crate) fn database_mut(&self) -> impl std::ops::DerefMut<Target = D> + '_ {
+        self.database.borrow_mut()
+    }
+
```
afilini, line 1760: I don't understand exactly why this is necessary

src/testutils/blockchain_tests.rs
```diff
@@ -358,1112 +359,1611 @@ where
...
-            use $crate::testutils;
-
-            use super::*;
-
-            #[allow(unused_variables)]
-            fn get_blockchain(test_client: &TestClient) -> $blockchain {
-                $( let $test_client = test_client; )?
-                $block
+// Marker enum to direct testClient to AnyBlockChain conversion
+pub enum BlockchainType {
```
afilini: This is a step backward compared to what we had before: the goal of the macro was not only to test our internal blockchains, but also to allow our users to test their blockchains against a set of edge cases that we maintain and that should hopefully be pretty complete.

This can also be done without a macro, but it needs to work with generics instead of `AnyBlockchain`, because `AnyBlockchain` cannot be extended outside of BDK (essentially it only contains our blockchain types).

Alternatively, instead of using the generics you might be able to use the [`impl Trait`](https://doc.rust-lang.org/reference/types/impl-trait.html), but we definitely need something more generic than this for our external users.

src/testutils/helpers.rs
rajarshimaitra: > In the macro if min_confirmations is zero we map that to None immediately


Is that the correct behavior though?? If minimum confirmation is 0, and current height is 100, I would expect tx_meta confirmation height to be at 100. We were setting it to `None` previously even when we were setting a transaction with min conf 0, which should make the transaction immediately confirmed. Maybe I am confused on the intention of min_confirmation here.. 

> The macro had an unwrap() on the checked_sub(). I understand it's better to handle errors, but in this case we want to fail if the height is too low for a given number of confirmations. This code instead silently makes the tx unconfirmed because it maps that case to None

I was very confused with the way the min_conf thing was used.. And it doesn't work if we are setting current_height to None.. Which is a requirement to fix #719.. So all that mixed together ended in the logic change.. But I get your point, will try to revert to previous behavior in the next push..  

> There's a +1 missing which was present in the macro

Yeah +1 should be there.. Will update in next push..

src/wallet/mod.rs
rajarshimaitra: Me neither.. But when I tried to directly use `self.database.borrow_mut()` in the function calls the compiler got angry and threw lots of problems that I couldn't subvert, and the easiest thing I found was to have this redundant function.. This does bother me, as we should not need something like this.. But for time being this works.. And I will try to remove it again and see if it can work.. But would prefer to do it in a separate PR..

src/testutils/blockchain_tests.rs
rajarshimaitra: Yes this makes sense.. I will work on it and update accordingly..

src/wallet/mod.rs
rajarshimaitra: Another problem faced is previously the `populate_test_db` macro was in the database module.. Which made accessing database easy.. But now as it has been taken out into `testutils::helper` we cannot just call `wallet.db` as that's private variable and only exposed via the `database()` api.. Which only gives immutable access.. So thats the major reason I had to make a mutable version of the same API.. But only for testing situations..

src/testutils/helpers.rs
danielabrozzoni: > May be I am missing something.. But why is this +1 required? If current height is 100, min_conf is 5. The tx confirmation height would be at 95. Which is simply the subtraction of two numbers.

Nope: if the current height is 100, and my transaction was included in block 100, then it has 1 confirmation. If the current height is 100, and my transaction is still in mempool, then it has 0 confirmations.

> Is that the correct behavior though?? If minimum confirmation is 0, and current height is 100, I would expect tx_meta confirmation height to be at 100. 

Nope, if the confirmation is 0 then the transaction was never included in a block, so there's no confirmation_time

>  If a transaction with min_conf = 1 makes it confirmed at height current_height - 1 I would expect min_conf=0 would make it confirmed at current_height. Not an unconfirmed transaction.

Well, but min_conf = 1 makes it confirmed at height `current_height`

rajarshimaitra: Hey @danielabrozzoni I updated my comments before I saw yours.. Yes it makes sense and I am trying to revert back to the previous behavior.. My bad.. I missed the +1 calculation there..

rajarshimaitra:
Hey @afilini I have updated the conftime calculation logic. I think now it is reverted as previous.. And it now expects the min_conf time to be lower than current height with a message. 

Also changed the blockchain tests to now run with a generic `init_blockchain()` function.. So I think for external users the test flow now would look like
 -  Create your own `init_blockchain()` function.
 -  Pass that into the macro.
 - Invoke the test cases names you wanna test on.. 
 
But the user still needs to manually specify the test cases name in the macro.. I couldn't find a good way to automate that part too.. 

kept the update commit separate for easier review.. But it might also makes sense to squash them, because many stuffs of the previous commits are now redundant..

rajarshimaitra:
I am also wondering if we should document it somewhere that our blockchain tests are public and how to use them. Libraries don't usually expose their tests. So might be helpful for users..

src/testutils/helpers.rs
rajarshimaitra: Done..

src/testutils/helpers.rs
rajarshimaitra: I am resolving this one as its not relevant to this PR.. Let me know if it needs to be open..

rajarshimaitra:
Thanks @afilini @danielabrozzoni for the initial look.

Pushed some updates

 - Rebased and resolved conflicts with `master`. Few changes added here previously are already done in `master`. Those are removed from this PR. 
  - Restructured all the commits. Including review suggestion and many follow up code updates.
  - Refer commit messages for a change summary.

This PR is ready for a second look.. :)

src/wallet/mod.rs
evanlinjin: We can make the `Wallet::database` field `pub(crate)`.

I tried it here: https://github.com/rajarshimaitra/bdk/pull/8

src/blockchain/esplora/mod.rs
```diff
@@ -149,19 +149,61 @@ impl_error!(std::num::ParseIntError, Parsing, EsploraError);
...
-        EsploraBlockchain::new(&format!("http://{}",test_client.electrsd.esplora_url.as_ref().unwrap()), 20)
-    }
-}
-
 const DEFAULT_CONCURRENT_REQUESTS: u8 = 4;
 
+#[cfg(feature = "test-esplora")]
 #[cfg(test)]
-mod test {
+pub mod test {
```
evanlinjin: This seems like an accidental change to `pub`. Let me know if I'm mistaken.

If the idea here is to make `test::init_blockchain` externally-accessible, we need to make `init_blockchain` `pub` as well.

src/blockchain/rpc.rs
```diff
@@ -870,32 +870,79 @@ impl BlockchainFactory for RpcBlockchainFactory {
...
+            test_sync_bump_fee_add_input_simple,
+            test_sync_bump_fee_add_input_no_change,
+            test_sync_receive_coinbase,
+            test_double_spend,
+            test_tx_chain,
+            test_get_block_hash,
+        )
+    ];
+
+    #[cfg(not(feature = "test-rpc-legacy"))]
```
evanlinjin, line 933: Nit: IMO writing this as `#[cfg(feature = "test-rpc")]` is better.

src/testutils/helpers.rs
```diff
@@ -0,0 +1,201 @@
...
+use crate::{
+    database::{AnyDatabase, BatchDatabase, BatchOperations, Database, MemoryDatabase, SyncTime},
+    testutils, BlockTime, KeychainKind, LocalUtxo, TransactionDetails, Wallet,
+};
+
+use super::TestIncomingTx;
+
+#[doc(hidden)]
+/// Populate a test database with a `TestIncomingTx`, as if we had found the tx with a `sync`.
+/// This is a hidden function, only useful for `DataBase` unit testing.
```
evanlinjin: Nit: Typo `Database`.

evanlinjin (commented):
LGTM, however it's really hard to read through all the changes of `blockchain_tests.rs`, but I am trusting that nothing dodgy is added. Also added some comments here and there.

I would love to see this merged soon so we can get started on https://discord.com/channels/753336465005608961/753367451319926827/1012298169830354966 :smile:

rajarshimaitra:
> LGTM, however it's really hard to read through all the changes of blockchain_tests.rs, but I am trusting that nothing dodgy is added. Also added some comments here and there.

Please don't trust. Verify! 😅

Ya the diff for the blockchain tests is pretty nasty.. Where as its just a single big move of all the tests out of the macro.. But for some reason that's how git is showing.. Let me know if breaking this diff into smaller pieces can help in review..  

One thing most important to verify that we are playing all the previous tests. 
 - No existing test was removed.
 - The call list at the blockchian module sites are covering everything they were covering previously.
 
 Thanks for the review @evanlinjin I will update them in the next push..

src/blockchain/esplora/mod.rs
rajarshimaitra: Huh.. Ya this should not be public.. We don't wanna publish `init_blockhain` because user would write that on their own for translating their custom blockchain..

src/blockchain/rpc.rs
rajarshimaitra: Agreed..

src/testutils/helpers.rs
rajarshimaitra: Thanks for the catch.. 👍

danielabrozzoni:
Moving this one to the next milestone, as it seems that there is still some work to do :)

rajarshimaitra:
Sorry I forgot about the last pending updates.. Yes this can wait for the next release.. There are few conflicts too.. I will resolve them after 0.23 is cut..

src/blockchain/rpc.rs
rajarshimaitra: On further look, it seems the intention of `test-rpc-legacy` is a to be a feature to disable certain tests.. So making it `test-rpc` won't work, as all the rpc tests are by default in `test-rpc`. This `not` thing, makes these tests disabled when `test-rpc-legacy` is turned on.. Maybe there was an easier way to do, but I won't mess with it in this PR..

rajarshimaitra:
Thanks for dealing with the mammoth this long everybody.. I think we are close to get this one in.. 

- Addressed @evanlinjin 's comments.
- Rebased on master.. 

This is now ready for a second look.. :)

danielabrozzoni:

PR #625, opened 2022-06-06
Restrict `drain_to` usage

### LINKED ISSUE #620: Subtle wallet drain related to TxParams
**Description of the bug**  
I discovered that if you create a transaction with no `recipient`,
`drain_wallet` set to false, **drain_to** set with some scriptpubkey, left all
the other `TxParams` as they are and finish the _psbt_, you still make a
transaction paying the fees for the fixed parts of the transaction and the fees
generated through coin selection.

It doesn't seem to be intentional and it drains funds from your wallet for free
as result.

**To Reproduce**  
I crafted a [proof of concept](https://github.com/csralvall/bdk/commit/acac621ca11073174787f2b72baff239f007638d).

**Expected behavior**  
I expect that a _psbt_ without recipients and not used to sweep a descriptor shouldn't be allowed to finalize.

**Build environment**  
 - BDK tag/commit: fbd98b4
 - OS+version: `5.10.105-1-MANJARO`
 - Rust version:  `rustc 1.58.1 (db9d1b20b 2022-01-20)`
 - Cargo version: `cargo 1.58.0 (f01b232bc 2022-01-19)`
 - Rust/Cargo target: `x86_64-unknown-linux-gnu`

**Additional context**  
I also attempted a fix in this [commit](https://github.com/csralvall/bdk/commit/79d5b6f2b26e739ddf6f206abaee180f0ecc72c3)
but there're conflicts with RBF policies.

It seems that `bump_fee` sets `drain_wallet` to false through the use of
`Default` values to create the new _psbt_ with the bumped fee.

I'm not familiar enough with **RBF** to know what I can and can't do, so I
would like to hear other ideas.

danielabrozzoni:
### Description
Before this commit, you could create a transaction with `drain_to` set
without specifying recipients, nor `drain_wallet`, nor `utxos`. What would
happen is that BDK would pick one input from the wallet and send
that one to `drain_to`, which is quite weird.
This PR restricts the usage of `drain_to`: if you want to use it as a
change output, you need to set recipients as well. If you want to send
a specific utxo to the `drain_to` address, you specify it through
`add_utxos`. If you want to drain the whole wallet, you set
`drain_wallet`. In any other case, if `drain_to` is set, we return a
`NoRecipients` error.

Fixes #620

#### All Submissions:

#### Bugfixes:

* [x] This pull request breaks the existing API - kinda? 

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -791,7 +791,14 @@ where
         let drain_val = (coin_selection.selected_amount() - outgoing).saturating_sub(fee_amount);
 
         if tx.output.is_empty() {
-            if params.drain_to.is_some() {
+            // Uh oh, our transaction has no outputs.
+            // We allow this when:
+            // - We have a drain_to address and the utxos we must spend (this happens,
+            // for example, when we RBF)
+            // - We have a drain_to address and drain_wallet set
+            // Otherwise, we don't know who we should send the funds to, and how much
+            // we should send!
+            if params.drain_to.is_some() && (params.drain_wallet || !params.utxos.is_empty()) {
                 if drain_val.is_dust(&drain_output.script_pubkey) {
                     return Err(Error::InsufficientFunds {
                         needed: drain_output.script_pubkey.dust_value().as_sat(),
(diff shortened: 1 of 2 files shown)


csralvall (commented):
Tested ACK.

I prepared the following table to check if all the possible execution paths
including the new conditions in `create_tx` are tested:

> x = _doesn't care conditions_

| `drain_to.is_some()` | `drain_wallet` | `!utxos.is_empty()` | `drain_val.is_dust(...)` | Test |
|:----:|:----:|:----:|:----:|:----:|
| False | x | x | x | test_create_tx_empty_recipients |
| True | False | False | False | test_create_tx_drain_to_no_drain_wallet_no_utxos (_NEW TEST_) |
| True | False | True | False | test_create_tx_drain_to_and_utxos (_NEW TEST_) - test_bump_fee_reduce_single_recipient |
| True | True | x | False | test_create_tx_drain_wallet_and_drain_to |
| True | True | x | True | test_create_tx_absolute_high_fee - test_create_tx_drain_to_dust_amount |
| True | False | True | True | ? |

I don't see much gain in including a test for the last case in terms of tested
functionality. However, it may worth the effort to document the possible ways
`create_tx` can be called.
Any test for *RBF* with a too high fee rate should do
the trick.

LLFourn (approved):
ACK 098c448dfbdf1922d9911760da27d76df67af420

src/wallet/mod.rs
```diff
@@ -2176,6 +2183,40 @@ pub(crate) mod test {
...
+        let utxos: Vec<_> = wallet
+            .get_available_utxos()
+            .unwrap()
+            .into_iter()
+            .map(|(u, _)| u.outpoint)
+            .collect();
+        let mut builder = wallet.build_tx();
+        builder
+            .drain_to(addr.script_pubkey())
+            .add_utxos(&utxos)
```
rajarshimaitra, line 2199: Not very pressing but maybe we can have another test, but instead of preselected utxos, set `drain_wallet` and `drain_to` and confirm behavior, to verify the other side of the conditional?

rajarshimaitra (commented):
tACK 098c448dfbdf1922d9911760da27d76df67af420

Code changes looks good to me. Just one comment.

src/wallet/mod.rs
danielabrozzoni: Do you mean like `test_create_tx_drain_wallet_and_drain_to` (already in master: https://github.com/bitcoindevkit/bdk/pull/625/files/098c448dfbdf1922d9911760da27d76df67af420#diff-18f5545e0b6d6f7d1d8ef9f744c9d6d1458baaffc419e69f826626c00dddc260L2138), or something else?

rajarshimaitra: My bad.. I missed that one.. Yup was talking about that only..

rajarshimaitra (approved):
ReAck 098c448dfbdf1922d9911760da27d76df67af420

notmandatory:
I think this needs a rebase to pickup the "Blockchain test-rpc-legacy"  CI test, then we can merge it.

danielabrozzoni:
Looks like Github Action got updated and now CI is broken 🔥
I'm working on it (but it might take a while) 🧯

danielabrozzoni:
Rebased :)

afilini (approved):
ACK 6a150368674046f796f5c37755896f16d8345fbc

Issue #597, opened 2022-05-03
Add additional information to `Policy` structure


notmandatory:
As discussed in bitcoindevkit/bdk#594 if we can somehow add the below information to the `Policy` structure it would be more useful in displaying the general signing conditions for a `Descriptor` to users. 

* signing xkey checksums
* signing xkey origin (master key fingerprint and derivation path), if available
* script type (ie. sh(), wsh(), sh(wsh), tr(), etc.)

bucko13:
I'm not sure this necessarily has to play a role, but thought it was at least worth raising and having for the record this recent discussion from the ledger team on the developer mailing list about a "wallet policy" extension for descriptors: 

https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2022-May/020423.html

notmandatory:
The plan going forward for `bdk` is to replace all the policy stuff with the `rust_miniscript` planning module approach. See: bitcoindevkit/bdk_wallet#4, bitcoindevkit/bdk_wallet#64, and bitcoindevkit/bdk_wallet#164.

PR #626, opened 2022-06-07
add extended descriptor multisig metadata to policy extraction

bucko13:
### Description

This is a first attempt at resolving #597. When using the policy module to parse descriptors, it can be useful to return additional information about the descriptor. This PR adds some of that in, namely key origin details about each key involved and the script type (p2sh, p2sh-p2wsh, or p2wsh). 

### Notes to the reviewers
An earlier version of this took advantage of `PkOrF` being a `struct`, but unfortunately just as I was ready to push up for a PR a bunch of changes came in that impacted this, specifically using an `enum` for `PkOrF` instead which changed a lot of the assumptions and tests. The enum makes most sense when these objects are relatively one dimensional (also makes for more straightforward tests), but when the key item can have more properties such as including more of the key origin, it got a little trickier. 

I only updated the tests for the test I added for this feature plus one other I saw breaking. I thought I'd put it up as a draft first to get any feedback before tweaking any further. It also wasn't clear if some of the tests were breaking for other reasons. 

The keys also come out a little weirder when serialized now then when they were done as a struct. Since they are basically tagged with the struct name, each key is an object with a single key `xpub_origin` whose value is an object with the origin details: `fingerprint` (which might be better called `xfp` or `root_fingerprint`?), `derivation_path`, and `xkey`).

I also considered whether or not to add the full_derivation_path or trailing path which you can sometimes have in a descriptor including wildcard (e.g. `/0/*`), but figured I could wait for that too. 

Note, I'm very new at Rust (this is my first production Rust code written), so I'm sure there's a lot that could be more idiomatic or that I've just done wrong completely. All suggestions on how to improve welcome!

### Checklists

* [ ] fix remaining tests
* [ ] add full_derivation_path (or similar) (?) 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/descriptor/policy.rs
+++ b/src/descriptor/policy.rs
@@ -41,10 +41,10 @@ use std::collections::{BTreeMap, HashSet, VecDeque};
 use std::fmt;
 
 use serde::ser::SerializeMap;
-use serde::{Serialize, Serializer};
+use serde::{Deserialize, Serialize, Serializer};
 
 use bitcoin::hashes::*;
-use bitcoin::util::bip32::Fingerprint;
+use bitcoin::util::bip32::{DerivationPath, ExtendedPubKey, Fingerprint};
 use bitcoin::{PublicKey, XOnlyPublicKey};
 
 use miniscript::descriptor::{
@@ -76,6 +76,19 @@ pub enum PkOrF {
     XOnlyPubkey(XOnlyPublicKey),
     /// An extended key fingerprint
     Fingerprint(Fingerprint),
+    /// serialization for xpub origin metadata
+    XpubOrigin(XpubOrigin),
+}
+
+/// Raw public key or extended key fingerprint
+#[derive(Debug, Clone, Default, Serialize, PartialEq, Eq, Hash)]
+pub struct XpubOrigin {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    fingerprint: Option<Fingerprint>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    derivation_path: Option<DerivationPath>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    xkey: Option<ExtendedPubKey>,
 }
 
 impl PkOrF {
@@ -89,9 +102,27 @@ impl PkOrF {
                 key: SinglePubKey::XOnly(pk),
                 ..
             }) => PkOrF::XOnlyPubkey(*pk),
-            DescriptorPublicKey::XPub(xpub) => PkOrF::Fingerprint(xpub.root_fingerprint(secp)),
+            DescriptorPublicKey::XPub(xpub) => PkOrF::XpubOrigin(XpubOrigin {
+                fingerprint: Some(xpub.root_fingerprint(secp)),
+                derivation_path: xpub.origin.as_ref().map(|origin| origin.clone().1),
+                xkey: Some(xpub.xkey),
+            }),
         }
     }
+}
+
+/// script type used to encode multisig script
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, Eq, PartialEq)]
+pub enum MultisigScriptType {
+    /// P2SH
+    #[serde(rename = "P2SH")]
+    P2sh,
+    /// P2SH-P2WSH
+    #[serde(rename = "P2SH_P2WSH")]
+    P2shP2wsh,
+    /// P2WSH
+    #[serde(rename = "P2WSH")]
+    P2wsh,
 }
 
 /// An item that needs to be satisfied
@@ -139,6 +170,9 @@ pub enum SatisfiableItem {
         keys: Vec<PkOrF>,
         /// The required threshold count
         threshold: usize,
+        /// The script type used to encode
+        #[serde(skip_serializing_if = "Option::is_none")]
+        script_type: Option<MultisigScriptType>,
     },
 
     // Complex item
@@ -578,6 +612,7 @@ impl Policy {
         threshold: usize,
         sorted: bool,
         secp: &SecpCtx,
+        script_type: Option<MultisigScriptType>,
     ) -> Result<Option<Policy>, PolicyError> {
         if threshold == 0 {
             return Ok(None);
@@ -621,6 +656,7 @@ impl Policy {
         let mut policy: Policy = SatisfiableItem::Multisig {
             keys: parsed_keys,
             threshold,
+            script_type,
         }
         .into();
         policy.contribution = contribution;
@@ -690,7 +726,9 @@ impl Policy {
 
                 Ok(requirements)
             }
-            SatisfiableItem::Multisig { keys, threshold } => {
+            SatisfiableItem::Multisig {
+                keys, threshold, ..
+            } => {
                 if selected.len() < *threshold {
                     return Err(PolicyError::NotEnoughItemsSelected(self.id.clone()));
                 }
@@ -951,7 +989,7 @@ impl<Ctx: ScriptContext + 'static> ExtractPolicy for Miniscript<DescriptorPublic
                 Some(SatisfiableItem::Hash160Preimage { hash: *hash }.into())
             }
             Terminal::Multi(k, pks) | Terminal::MultiA(k, pks) => {
-                Policy::make_multisig::<Ctx>(pks, signers, build_sat, *k, false, secp)?
+                Policy::make_multisig::<Ctx>(pks, signers, build_sat, *k, false, secp, None)?
             }
             // Identities
             Terminal::Alt(inner)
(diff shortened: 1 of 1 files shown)


src/descriptor/policy.rs
```diff
@@ -76,6 +76,19 @@ pub enum PkOrF {
     XOnlyPubkey(XOnlyPublicKey),
     /// An extended key fingerprint
     Fingerprint(Fingerprint),
+    /// serialization for xpub origin metadata
```
rajarshimaitra, line 79: Can we call this "An XPub with origin metadata"?? As per the other item docs and sounds a bit more clear to me.

src/descriptor/policy.rs
```diff
@@ -76,6 +76,19 @@ pub enum PkOrF {
     XOnlyPubkey(XOnlyPublicKey),
     /// An extended key fingerprint
     Fingerprint(Fingerprint),
```
rajarshimaitra, line 78: Seems like we also don't need this `FingerPrint` anymore as its included in the XpubOrigin??

src/descriptor/policy.rs
```diff
@@ -76,6 +76,19 @@ pub enum PkOrF {
...
+
+/// Raw public key or extended key fingerprint
+#[derive(Debug, Clone, Default, Serialize, PartialEq, Eq, Hash)]
+pub struct XpubOrigin {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    fingerprint: Option<Fingerprint>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    derivation_path: Option<DerivationPath>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    xkey: Option<ExtendedPubKey>,
```
rajarshimaitra, line 91: This seems to me representing a full XPub, not just the origin.. So we can name it better as just an `XPub`?

src/descriptor/policy.rs
```diff
@@ -89,11 +102,29 @@ impl PkOrF {
...
+    /// P2SH
+    #[serde(rename = "P2SH")]
+    P2sh,
+    /// P2SH-P2WSH
+    #[serde(rename = "P2SH_P2WSH")]
+    P2shP2wsh,
+    /// P2WSH
+    #[serde(rename = "P2WSH")]
+    P2wsh,
+}
```
rajarshimaitra, line 126: We can use `ScriptType` for any type of scripts.. Not just multisigs..

src/descriptor/policy.rs
```diff
@@ -139,6 +170,9 @@ pub enum SatisfiableItem {
         keys: Vec<PkOrF>,
         /// The required threshold count
         threshold: usize,
+        /// The script type used to encode
+        #[serde(skip_serializing_if = "Option::is_none")]
+        script_type: Option<MultisigScriptType>,
```
rajarshimaitra, line 175: I am feeling like we should be able to set the `ScriptType` in the `Policy` structure itself. Not just only multisigs.. And this should not be inside `SatisfiableItem`..

rajarshimaitra (changes requested):
I went through the code briefly, and I have some comments on the `MultisigScriptType` stuffs.. 

The script type should not multisig specific. We should be able to get the script type for any generic descriptors. So its should be a field inside the `Policy` structure itself.

But the problem of doing that is our `Policy` structure is recursive.. So we cannot just add an extra field inside the `Policy` itself, as it will then get recursively added in the internal policy nodes too.. 
 
 So Instead one way to do that would be to have a `GlobalPolicy` structure, which includes the global `ScriptType` and a `Policy`. Then the remaining logic of `Policy` structure remains the same, which can recursively include more policies in it. 
 
 Concept ACK on the XPub data..

bucko13:
Thanks for the review @rajarshimaitra! I’m afk for a little bit but at a high level this makes sense to me. Hopefully I can get some time to refactor when I’m back in September

rajarshimaitra:
No issues.. This is a valuable PR IMO.. It will be helpful to get more info out from our policy language..

src/descriptor/policy.rs
bucko13: Yep! Good catch.

src/descriptor/policy.rs
bucko13: Are you referring to the struct `XpubOrigin` or this field, `xkey`? 

In the contexts I'm most familiar working in, an xpub origin has more information than the xpub (extended public key) itself. Namely the origin is the xpub and the information describing how it was derived:  fingerprint for the root (`xfp` which might be a better name for that field) and the derivation path from root. The xpub only has its parent fingerprint and depth encoded (with some other data too of course.)  

As for renaming `xkey`, I forget why I put it as this, but I feel like I remember there being a reason, maybe some sort of naming clash. If that isn't a problem though then I'm not against renaming that to `xpub`.

src/descriptor/policy.rs
bucko13: Yep. Good point. I assume you're recommending the Enum be renamed to `ScriptType`? And maybe `P2tr should be included here now too if we're moving this to policy?

src/descriptor/policy.rs
bucko13: I see the confusion. Basically because there's already something called `origin` in the library (the field on the [DescriptorXKey struct](https://docs.rs/bdk/latest/bdk/descriptor/struct.DescriptorXKey.html#structfield.origin)) so it's a little confusing. I guess I'm trying to find some name that represents the full struct necessary for defining an xpub in a meaningful way (e.g. when verifying membership in a multisig wallet, which is one of the primary use cases we're building this out for). 

It's been a while since I've spent time in this code, so it's possible I'm missing something, but basically what I was originally thinking was where `XpubOrigin` is an object that contains an Extended Public Key plus it's derivation/origin metadata. 

OTOH, maybe I'm misunderstanding your comment and you're just asking to update this code comment 😅. In which case, maybe "serialization for xpub with derivation metadata" would be more clear to avoid the naming clash with origin?

bucko13:
Thanks again for the review @rajarshimaitra. Just to be clear, the request for updates is basically two parts:

1. We want a new `GlobalPolicy` struct that has two parts: `ScriptType` and `Policy`, where `Policy` is the normal `Policy` as it is now. 
2. Xpub origin data concept is good as is, but sounds like we want to clear up the language/naming a bit (see my comments in the PR on that)

src/descriptor/policy.rs
rajarshimaitra: I was actually referring to the struct name `XpubOrigin`.. The name suggests that its an `origin` for `Xpub`, but its more like an `XpubWithOrigin`. But I agree its mostly nomenclature and there's no *right* way to do it.. I think just the doc comment as something like  `/// Extended Publickey with origin information` is good enough to clarify what this struct is for.. The struct name can remain as it is..

src/descriptor/policy.rs
rajarshimaitra: Yup sounds good.. I think bdk can already recognize `p2tr`. A test with an example taproot descriptor into policy and checking if its parsing it correctly would be really nice..

src/descriptor/policy.rs
rajarshimaitra: OTOH if I am not wrong, after parsing a descriptor we will always have the origin information? If no fingerprint and path is provided in the descriptor, bdk assumes it as the master key? If that's the case we can simply name this struct `Xpub` and it will imply that origin information is already included in it.. As we never put an Xpub into a policy without its origin..

bucko13: Even if bdk never puts an xpub w/o origin details, if we're given one to interpret can you safely assume it's the root key? You don't need the origin information for watch-only wallets for example since you only need those details to tell the master key how to sign. We do something like this for unchained wallet configurations (which I want to use this tooling to convert to descriptors) where we only show actual origin information for keys that are yours (the clients) and blank out the origin info for those you don't have permissions to sign with in our system (for example the unchained key). We do this to preserve privacy and avoid data leaks. The xpub's relation to the master seed (origin information) should be of no concern to the other party that's signing if that key is not going to be signed with. So, in this context the origin information is optional and this would be true for all keys if you only want a watch-only descriptor.

rajarshimaitra: Yes I agree.. For watch-only we don't need the origin info and its not prudent to assume its a master key.. ACK on that..

danielabrozzoni:

(More specifically, we're going to use the miniscript "planning module" instead of Policies)

bucko13:
Got it. Thanks for the update! We do want to revisit this soon but sounds like it’ll make sense to do it on the new API. Largely if we can export the policy info (or whatever it’s called 😅) and have it accessible via wasm bindings than that is really all we were trying to get from this.

PR #630, opened 2022-06-13
Move change logic to coin_select

### LINKED ISSUE #147: Move change creation/addition inside `CoinSelectionAlgorithm`

csralvall:
### Description

The former way to compute and create change was inside `create_tx`, just after
performing coin selection.
It blocked the opportunity to have an "ensemble" algorithm to decide between
multiple coin selection algorithms based on a metric, like Waste.
Now, change is not created inside `coin_select` but the change amount and the
possibility to create change is decided inside the `coin_select` method. In
this way, change is associated with the coin selection algorithm that generated
it, and a method to decide between them can be implemented.

Fixes #147.

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-
 - Add `descriptor::checksum::get_checksum_bytes` method.
+- Add `Excess` enum to handle remaining amount after coin selection.
+- Move change creation from `Wallet::create_tx` to `CoinSelectionAlgorithm::coin_select`.
 
 ## [v0.20.0] - [v0.19.0]
 
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -29,6 +29,7 @@
 //! # use bdk::wallet::{self, coin_selection::*};
 //! # use bdk::database::Database;
 //! # use bdk::*;
+//! # use bdk::wallet::coin_selection::decide_change;
 //! # const TXIN_BASE_WEIGHT: usize = (32 + 4 + 4 + 1) * 4;
 //! #[derive(Debug)]
 //! struct AlwaysSpendEverything;
@@ -42,6 +43,7 @@
 //!         fee_rate: FeeRate,
 //!         amount_needed: u64,
 //!         fee_amount: u64,
+//!         drain_script: &Script,
 //!     ) -> Result<CoinSelectionResult, bdk::Error> {
 //!         let mut selected_amount = 0;
 //!         let mut additional_weight = 0;
@@ -59,16 +61,21 @@
 //!             .collect::<Vec<_>>();
 //!         let additional_fees = fee_rate.fee_wu(additional_weight);
 //!         let amount_needed_with_fees = (fee_amount + additional_fees) + amount_needed;
-//!         if amount_needed_with_fees > selected_amount {
+//!         if selected_amount < amount_needed_with_fees {
 //!             return Err(bdk::Error::InsufficientFunds {
 //!                 needed: amount_needed_with_fees,
 //!                 available: selected_amount,
 //!             });
 //!         }
 //!
+//!         let remaining_amount = selected_amount - amount_needed_with_fees;
+//!
+//!         let excess = decide_change(remaining_amount, fee_rate, drain_script);
+//!
 //!         Ok(CoinSelectionResult {
 //!             selected: all_utxos_selected,
 //!             fee_amount: fee_amount + additional_fees,
+//!             excess,
 //!         })
 //!     }
 //! }
@@ -89,8 +96,12 @@
 //! ```
 
 use crate::types::FeeRate;
+use crate::wallet::utils::IsDust;
 use crate::{database::Database, WeightedUtxo};
 use crate::{error::Error, Utxo};
+
+use bitcoin::consensus::encode::serialize;
+use bitcoin::Script;
 
 use rand::seq::SliceRandom;
 #[cfg(not(test))]
@@ -111,6 +122,27 @@ pub type DefaultCoinSelectionAlgorithm = LargestFirstCoinSelection; // make the 
 // prev_txid (32 bytes) + prev_vout (4 bytes) + sequence (4 bytes) + script_len (1 bytes)
 pub(crate) const TXIN_BASE_WEIGHT: usize = (32 + 4 + 4 + 1) * 4;
 
+#[derive(Debug)]
+/// Remaining amount after performing coin selection
+pub enum Excess {
+    /// It's not possible to create spendable output from excess using the current drain output
+    NoChange {
+        /// Threshold to consider amount as dust for this particular change script_pubkey
+        dust_threshold: u64,
+        /// Exceeding amount of current selection over outgoing value and fee costs
+        remaining_amount: u64,
+        /// The calculated fee for the drain TxOut with the selected script_pubkey
+        change_fee: u64,
+    },
+    /// It's possible to create spendable output from excess using the current drain output
+    Change {
+        /// Effective amount available to create change after deducting the change output fee
+        amount: u64,
+        /// The deducted change output fee
+        fee: u64,
+    },
+}
+
 /// Result of a successful coin selection
 #[derive(Debug)]
 pub struct CoinSelectionResult {
@@ -118,6 +150,8 @@ pub struct CoinSelectionResult {
     pub selected: Vec<Utxo>,
     /// Total fee amount in satoshi
     pub fee_amount: u64,
+    /// Remaining amount after deducing fees and outgoing outputs
+    pub excess: Excess,
 }
 
 impl CoinSelectionResult {
@@ -157,6 +191,8 @@ pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
     /// - `amount_needed`: the amount in satoshi to select
     /// - `fee_amount`: the amount of fees in satoshi already accumulated from adding outputs and
     ///                 the transaction's header
+    /// - `drain_script`: the script to use in case of change
+    #[allow(clippy::too_many_arguments)]
     fn coin_select(
         &self,
         database: &D,
@@ -165,6 +201,7 @@ pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
         fee_rate: FeeRate,
         amount_needed: u64,
         fee_amount: u64,
+        drain_script: &Script,
     ) -> Result<CoinSelectionResult, Error>;
 }
 
@@ -184,6 +221,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
         fee_rate: FeeRate,
         amount_needed: u64,
         fee_amount: u64,
+        drain_script: &Script,
     ) -> Result<CoinSelectionResult, Error> {
         log::debug!(
             "amount_needed = `{}`, fee_amount = `{}`, fee_rate = `{:?}`",
@@ -202,7 +240,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
                 .chain(optional_utxos.into_iter().rev().map(|utxo| (false, utxo)))
         };
 
-        select_sorted_utxos(utxos, fee_rate, amount_needed, fee_amount)
+        select_sorted_utxos(utxos, fee_rate, amount_needed, fee_amount, drain_script)
     }
 }
 
@@ -222,6 +260,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for OldestFirstCoinSelection {
         fee_rate: FeeRate,
         amount_needed: u64,
         fee_amount: u64,
+        drain_script: &Script,
     ) -> Result<CoinSelectionResult, Error> {
         // query db and create a blockheight lookup table
         let blockheights = optional_utxos
@@ -261,7 +300,33 @@ impl<D: Database> CoinSelectionAlgorithm<D> for OldestFirstCoinSelection {
                 .chain(optional_utxos.into_iter().map(|utxo| (false, utxo)))
         };
 
-        select_sorted_utxos(utxos, fee_rate, amount_needed, fee_amount)
+        select_sorted_utxos(utxos, fee_rate, amount_needed, fee_amount, drain_script)
+    }
+}
+
+/// Decide if change can be created
+///
+/// - `remaining_amount`: the amount in which the selected coins exceed the target amount
+/// - `fee_rate`: required fee rate for the current selection
+/// - `drain_script`: script to consider change creation
+pub fn decide_change(remaining_amount: u64, fee_rate: FeeRate, drain_script: &Script) -> Excess {
+    // drain_output_len = size(len(script_pubkey)) + len(script_pubkey) + size(output_value)
+    let drain_output_len = serialize(drain_script).len() + 8usize;
+    let change_fee = fee_rate.fee_vb(drain_output_len);
+    let drain_val = remaining_amount.saturating_sub(change_fee);
+
+    if drain_val.is_dust(drain_script) {
+        let dust_threshold = drain_script.dust_value().as_sat();
+        Excess::NoChange {
+            dust_threshold,
+            change_fee,
+            remaining_amount,
+        }
+    } else {
+        Excess::Change {
+            amount: drain_val,
+            fee: change_fee,
+        }
     }
 }
 
@@ -270,6 +335,7 @@ fn select_sorted_utxos(
     fee_rate: FeeRate,
     amount_needed: u64,
     mut fee_amount: u64,
+    drain_script: &Script,
 ) -> Result<CoinSelectionResult, Error> {
     let mut selected_amount = 0;
     let selected = utxos
@@ -296,6 +362,7 @@ fn select_sorted_utxos(
         .collect::<Vec<_>>();
 
     let amount_needed_with_fees = amount_needed + fee_amount;
+
     if selected_amount < amount_needed_with_fees {
         return Err(Error::InsufficientFunds {
             needed: amount_needed_with_fees,
@@ -303,9 +370,14 @@ fn select_sorted_utxos(
         });
     }
 
+    let remaining_amount = selected_amount - amount_needed_with_fees;
+
+    let excess = decide_change(remaining_amount, fee_rate, drain_script);
+
     Ok(CoinSelectionResult {
         selected,
         fee_amount,
+        excess,
     })
 }
 
@@ -366,6 +438,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for BranchAndBoundCoinSelection {
         fee_rate: FeeRate,
         amount_needed: u64,
         fee_amount: u64,
+        drain_script: &Script,
     ) -> Result<CoinSelectionResult, Error> {
         // Mapping every (UTXO, usize) to an output group
         let required_utxos: Vec<OutputGroup> = required_utxos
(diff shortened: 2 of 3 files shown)


src/wallet/coin_selection.rs
```diff
@@ -165,6 +186,7 @@ pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
         fee_rate: FeeRate,
         amount_needed: u64,
         fee_amount: u64,
+        _drain_output: &TxOut,
```
danielabrozzoni: Since this is the trait definition I wouldn't use the `_`, calling the parameter `drain_output`. Also, drain_output needs to be documented

src/wallet/coin_selection.rs
```diff
@@ -296,16 +321,31 @@ fn select_sorted_utxos(
...
     }
 
+    let remaining_amount = selected_amount - amount_needed_with_fees;
+
+    let change_fee = fee_rate.fee_vb(serialize(drain_output).len());
+
+    let drain_val = remaining_amount.saturating_sub(change_fee);
+
+    let excess = if drain_val.is_dust(&drain_output.script_pubkey) {
+        let target = drain_output.script_pubkey.dust_value().as_sat();
```
danielabrozzoni: What is target, exactly? The target you should reach for creating a change?

src/wallet/coin_selection.rs
```diff
@@ -111,13 +118,26 @@ pub type DefaultCoinSelectionAlgorithm = LargestFirstCoinSelection; // make the
 // prev_txid (32 bytes) + prev_vout (4 bytes) + sequence (4 bytes) + script_len (1 bytes)
 pub(crate) const TXIN_BASE_WEIGHT: usize = (32 + 4 + 4 + 1) * 4;
 
+#[derive(Debug)]
+/// Remaining amount after performing coin selection
+pub enum Excess {
+    /// It's not possible to create spendable output from excess using the current drain output
+    /// NoChange(target, remaining_amount, change_output_fee)
+    NoChange(u64, u64, u64),
```
danielabrozzoni: Maybe it would be better to have here and below a struct, instead of a tuple? With 3 elements of the same type, it starts to be difficult to remember which is what
OTOH, structs are more verbose and also you need to find a name for them.

danielabrozzoni (commented):
Concept ACK, it looks good to me! Just a couple of comments.

src/wallet/coin_selection.rs
csralvall: What do you think of variants with named fields, like [this](https://users.rust-lang.org/t/enum-field-expressions/2411/6).
It can be a middle ground between both options.
E.g.:
```rust
enum Excess {
      NoChange { dust_threshold: u64, remaining_amount: u64, change_fee: u64 },
      Change { amount: u64, fee: u64 }
}
```

src/wallet/coin_selection.rs
csralvall: The idea was that you must _hit_, at least, the dust value to create change (_the target_) . Now it doesn't seem so accurate. Maybe `dust_threshold` is.

src/wallet/coin_selection.rs
danielabrozzoni: That would work great!

src/wallet/coin_selection.rs
danielabrozzoni: Yeah `dust_threshold` is definitely easier to understand!

src/wallet/coin_selection.rs
afilini: ACK

(This is in reply to Daniela's comment about using named enum variants)

src/wallet/coin_selection.rs
```diff
@@ -597,6 +638,7 @@ impl BranchAndBoundCoinSelection {
         CoinSelectionResult {
             selected,
             fee_amount,
+            excess: Excess::Change(0, 0),
```
afilini: I suppose these are just temporary values, right?

src/wallet/coin_selection.rs
afilini: Technically here you don't need the full `TxOut`, the `Script` should be enough and I think it would be less confusing because it makes more sense that your coin selection depends on the change script, not some kind of output.

I would change this to a `drain_script` or something, then you'll have to replace the size estimation for the txout and it should be `serialize(script).len() + 8` (please double check this).

It's very important that you `serialize` the script and avoid using `Script::len()`: the former should also prefix the script with its length in VarInt format, which I believe is how it gets encoded in a `TxOut`. The latter only tells you the length of the script.

afilini (commented):
Concept ACK, this looks very similar to what I had in mind in terms of architecture

src/wallet/coin_selection.rs
csralvall: I'll change it by `drain_script: &Script` (with docs and no `_`) as @afilini suggested.

src/wallet/coin_selection.rs
csralvall: Yes, you're right. I'm trying to find a solution with a good balance between the parameter increment and the generality to reach each of the possible outputs of `BranchAndBound`.

src/wallet/coin_selection.rs
csralvall: Totally, this is great!
I double checked that `serialize(drain_output).len() == (serialize(drain_script).len() + 8)`.
Both are equal to: `size(len(script_pubkey)) + len(script_pubkey) + size(output_value)` where
`size(len(script_pubkey))` is the size of the `VarInt` field used to describe the length of the script.
So, it's safe for use.

danielabrozzoni:
Can you rebase on master to pick up the CI fixes?

src/wallet/coin_selection.rs
```diff
@@ -541,22 +628,40 @@ impl BranchAndBoundCoinSelection {
...
-            .collect();
+            .collect::<Vec<OutputGroup>>();
+
+        let selected_amount = selected_utxos
+            .iter()
+            .map(|u| u.effective_value)
+            .sum::<i64>();
+
+        // remaining_amount can't be negative as that would mean the
+        // selection wasn't successfull
```
danielabrozzoni: nit: `successful` (here and in other comments as well)

src/wallet/coin_selection.rs
```diff
@@ -541,22 +628,40 @@ impl BranchAndBoundCoinSelection {
...
         Ok(BranchAndBoundCoinSelection::calculate_cs_result(
             selected_utxos,
             required_utxos,
             fee_amount,
+            excess,
         ))
     }
 
+    #[allow(clippy::too_many_arguments)]
+    // REVIEW: Other approaches required more parameters or a large number of
```
danielabrozzoni: Related to this comment: maybe we should move the change_fee calculation inside of `decide_change`? What do you think? 
```diff
diff --git a/src/wallet/coin_selection.rs b/src/wallet/coin_selection.rs
index a9ca356..720a9e4 100644
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -300,7 +300,10 @@ impl<D: Database> CoinSelectionAlgorithm<D> for OldestFirstCoinSelection {
     }
 }
 
-fn decide_change(remaining_amount: u64, change_fee: u64, drain_script: &Script) -> Excess {
+fn decide_change(remaining_amount: u64, drain_script: &Script, fee_rate: &FeeRate) -> Excess {
+    // drain_output_len = size(len(script_pubkey)) + len(script_pubkey) + size(output_value)
+    let drain_output_len = serialize(drain_script).len() + 8usize;
+    let change_fee = fee_rate.fee_vb(drain_output_len);
     let drain_val = remaining_amount.saturating_sub(change_fee);
 
     if drain_val.is_dust(drain_script) {
@@ -360,11 +363,7 @@ fn select_sorted_utxos(
 
     let remaining_amount = selected_amount - amount_needed_with_fees;
 
-    // drain_output_len = size(len(script_pubkey)) + len(script_pubkey) + size(output_value)
-    let drain_output_len = serialize(drain_script).len() + 8usize;
-    let change_fee = fee_rate.fee_vb(drain_output_len);
-
-    let excess = decide_change(remaining_amount, change_fee, drain_script);
+    let excess = decide_change(remaining_amount, drain_script, &fee_rate);
 
     Ok(CoinSelectionResult {
         selected,
@@ -472,9 +471,6 @@ impl<D: Database> CoinSelectionAlgorithm<D> for BranchAndBoundCoinSelection {
             .try_into()
             .expect("Bitcoin amount to fit into i64");
 
-        // drain_output_len = size(len(script_pubkey)) + len(script_pubkey) + size(output_value)
-        let drain_output_len = serialize(drain_script).len() + 8usize;
-        let change_fee = fee_rate.fee_vb(drain_output_len);
 
         if curr_value > actual_target {
             let selected_amount = required_utxos
@@ -486,7 +482,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for BranchAndBoundCoinSelection {
             // actual_target = amount_needed + (fee_amount - vin_fees)
             let remaining_amount = (selected_amount - actual_target) as u64;
 
-            let excess = decide_change(remaining_amount, change_fee, drain_script);
+            let excess = decide_change(remaining_amount, drain_script, &fee_rate);
 
             return Ok(BranchAndBoundCoinSelection::calculate_cs_result(
                 vec![],
@@ -506,7 +502,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for BranchAndBoundCoinSelection {
                 fee_amount,
                 cost_of_change,
                 drain_script,
-                change_fee,
+                &fee_rate,
             )
             .unwrap_or_else(|_| {
                 self.single_random_draw(
@@ -516,7 +512,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for BranchAndBoundCoinSelection {
                     actual_target,
                     fee_amount,
                     drain_script,
-                    change_fee,
+                    &fee_rate,
                 )
             }))
     }
@@ -536,7 +532,7 @@ impl BranchAndBoundCoinSelection {
         fee_amount: u64,
         cost_of_change: f32,
         drain_script: &Script,
-        change_fee: u64,
+        fee_rate: &FeeRate,
     ) -> Result<CoinSelectionResult, Error> {
         // current_selection[i] will contain true if we are using optional_utxos[i],
         // false otherwise. Note that current_selection.len() could be less than
@@ -640,7 +636,7 @@ impl BranchAndBoundCoinSelection {
         // actual_target = amount_needed + (fee_amount - vin_fees)
         let remaining_amount = (selected_amount - actual_target) as u64;
 
-        let excess = decide_change(remaining_amount, change_fee, drain_script);
+        let excess = decide_change(remaining_amount, drain_script, fee_rate);
 
         Ok(BranchAndBoundCoinSelection::calculate_cs_result(
             selected_utxos,
@@ -661,7 +657,7 @@ impl BranchAndBoundCoinSelection {
         actual_target: i64,
         fee_amount: u64,
         drain_script: &Script,
-        change_fee: u64,
+        fee_rate: &FeeRate,
     ) -> CoinSelectionResult {
         #[cfg(not(test))]
         optional_utxos.shuffle(&mut thread_rng());
@@ -694,7 +690,7 @@ impl BranchAndBoundCoinSelection {
         // actual_target = amount_needed + (fee_amount - vin_fees)
         let remaining_amount = (selected_amount - actual_target) as u64;
 
-        let excess = decide_change(remaining_amount, change_fee, drain_script);
+        let excess = decide_change(remaining_amount, drain_script, fee_rate);
 
         BranchAndBoundCoinSelection::calculate_cs_result(
             selected_utxos,
```

This means that decide_change, bnb and single_random_draw will need a fee_rate parameter - but also that we don't have to repeat the `drain_output_len` and `change_fee` calculation in various parts of the code (this could come really handy when we separate SRD and BNB!).

danielabrozzoni (commented):
This is looking really good! I have a small comment about `decide_change`, otherwise it's good to go

Edit: also it needs a rebase for conflicts

src/wallet/coin_selection.rs
csralvall: I agree. Just a little gotcha, considering an x86_64 architecture, `FeeRate` is a `f32` so it takes 4 bytes and implements the copy trait. A reference size is `usize`, that is 8 bytes. Should we prefer references over moves or the other way around?

src/wallet/coin_selection.rs
csralvall: Fixed.

src/wallet/coin_selection.rs
danielabrozzoni: You're right, FeeRate also implements Copy - let's avoid the reference and just copy it :)

csralvall: Done.

danielabrozzoni (commented):
Can you squash the two commits together plz? Then I'll ACK :)

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Support for `tr()` descriptors in the `descriptor!()` macro
 - Add support for Bitcoin Core 23.0 when using the `rpc` blockchain
 - Add `remove_partial_sigs` and `try_finalize` to `SignOptions`
+- Add `Excess` enum to handle remaining amount after coin selection.
```
danielabrozzoni: These two should probably go under "Unreleased"

csralvall: Corrected.

danielabrozzoni (approved):
tACK b0f6a2089b44c3b18e97f8c2bf0183f65c67e958

src/wallet/coin_selection.rs
```diff
@@ -42,6 +43,7 @@
 //!         fee_rate: FeeRate,
 //!         amount_needed: u64,
 //!         fee_amount: u64,
+//!         _drain_script: &Script,
```
vladimirfomene: Forgot to remove the `_`, I think.

src/wallet/coin_selection.rs
```diff
@@ -261,7 +296,28 @@ impl<D: Database> CoinSelectionAlgorithm<D> for OldestFirstCoinSelection {
                 .chain(optional_utxos.into_iter().map(|utxo| (false, utxo)))
         };
 
-        select_sorted_utxos(utxos, fee_rate, amount_needed, fee_amount)
+        select_sorted_utxos(utxos, fee_rate, amount_needed, fee_amount, drain_script)
+    }
+}
+
+fn decide_change(remaining_amount: u64, fee_rate: FeeRate, drain_script: &Script) -> Excess {
+    // drain_output_len = size(len(script_pubkey)) + len(script_pubkey) + size(output_value)
```
vladimirfomene, line 313: From the comment for the `drain_output_len` computation, I was expecting, a third component in the form of the size of the `script_pubkey` in the computation below. I only see the addition of two components. Is there something, I'm missing?

vladimirfomene (commented):
tACK. Just Nits

src/wallet/coin_selection.rs
afilini: When you `serialize` a `Script` rust-bitcoin prefixes it with the length (which is a VarInt), so the first term amounts for both `serialize(len(script_pubkey)) + len(script_pubkey)`.

As I explained in the other comment, using `Script::len()` would have only returned the length of the script itself, without accounting for the VarInt preceeding it.

But the way this is done should be correct.

Also check out: https://github.com/bitcoindevkit/bdk/pull/630#discussion_r906696742

src/wallet/coin_selection.rs
danielabrozzoni: It's fine to use it here, as we're not using the drain_script anywhere in this function :)

src/wallet/coin_selection.rs
```diff
@@ -69,6 +71,7 @@
 //!         Ok(CoinSelectionResult {
 //!             selected: all_utxos_selected,
 //!             fee_amount: fee_amount + additional_fees,
+//!             excess: Excess::Change { amount: 0, fee: 0 },
```
afilini: This isn't the best in an example, because it probably creates an invalid tx (zero-value output for the change)

src/wallet/coin_selection.rs
```diff
@@ -408,10 +472,22 @@ impl<D: Database> CoinSelectionAlgorithm<D> for BranchAndBoundCoinSelection {
             .expect("Bitcoin amount to fit into i64");
 
         if curr_value > actual_target {
+            let selected_amount = required_utxos
+                .iter()
+                .map(|u| u.effective_value)
+                .sum::<i64>();
```
afilini: Isn't this literally `curr_value`? Why do you have to recompute it here?

src/wallet/coin_selection.rs
```diff
@@ -579,13 +677,31 @@ impl BranchAndBoundCoinSelection {
             })
             .collect::<Vec<_>>();
 
-        BranchAndBoundCoinSelection::calculate_cs_result(selected_utxos, required_utxos, fee_amount)
+        let selected_amount = selected_utxos
+            .iter()
+            .map(|u| u.effective_value)
+            .sum::<i64>();
```
afilini: You can probably merge this into the previous lines, replacing the `scan` with a `fold`. Keep a `(u64, Vec<_>)` as your internal state, every time you pick one utxo add the value to the accumulator and push it to the vec.

At the end you have both the selected utxos and the sum of their values

afilini (commented):
A few small comments.

Also I was thinking: do you think we should add some tests for the coin selection? I noticed that most of the tests were just updated to assume an empty drain_script and I'm actually a bit suprised none of the tests started failing (I was assuming at least some tests were asserting conditions on BDK not adding a change output, and reducing the size of the drain_script should cause them to potentially start adding one).

Can you double check if we are missing something there maybe?

src/wallet/coin_selection.rs
csralvall: Yes, I missed that. I can make `decide_change` public to compute the excess, but that would imply that it needs to be documented (I don't know if the function it's useful enough to be exposed in the docs) or I can mimic its behavior inside `AlwaysSpendEverything`.

src/wallet/coin_selection.rs
csralvall: So true!

src/wallet/coin_selection.rs
csralvall: Working on it, thanks!

csralvall:
@afilini 
> Also I was thinking: do you think we should add some tests for the coin selection? I noticed that most of the tests were just updated to assume an empty drain_script and I'm actually a bit suprised none of the tests started failing (I was assuming at least some tests were asserting conditions on BDK not adding a change output, and reducing the size of the drain_script should cause them to potentially start adding one).

They are not failing because the former code didn't make any change computations inside `coin_selection`, so the `excess` field who stores the change isn't touched by any of the tests inside `coin_selection.rs`. However, the correctness of change results is tested in the tests inside "mod", as the `decide_change` output is directly used there. Not the best option in terms of architecture, but at the beginning I thought they were enough.
That said, while reviewing your comments, I discovered the `selected_amount` I was using to compute the `remaining_amount` and then the change, was not considering the `required_utxos` amount in the BranchAndBoundAlgorithm and the tests didn't throw any error.
So, I'm in favor of adding or improving the tests. Like @rajarshimaitra said [here](https://github.com/bitcoindevkit/bdk/pull/666#pullrequestreview-1047902278), we may not be adequately testing the fee situations and edge cases.

> Can you double check if we are missing something there maybe?

So yes, I'm going to edit those empty scripts and see how I can improve the tests.

csralvall:
I've pushed some changes in this [branch](https://github.com/csralvall/bdk/commit/94951d4caa3e8b7068151a5d08a46a4b1af7c01a) underpinning the view that my proposal for the `coin_selection` module would have in the future.
The code is to display my ideas, so it won't compile yet.
I think that the return value of `coin_select` in that branch might be the final version. So, I'm not sure if it's worthy, at the end, to include the tests that I mentioned, since the functionality will be moved outside `coin_select` to a more general function.
I made some attempts to include tests, but all my ideas fell in circular logic, resembling highly to the `decide_change` function, already inside tested code.
The current checks and the fact that it's a refactoring and not an enhancement, are telling us that the PR didn't broke any current functionality. If something is wrong (and isn't reported by checks), it must be due to something previous, so a new PR addressing exclusively those errors and edge cases, or the tests for wallet in general, may be more appropriate.
With all this I mean that I've reviewed the current state of tests and any new change to the the modified code won't improve them any further.

afilini (approved):
ACK 32ae95f463f62c42c6d6aec62c1832a30298fce4

PR #634, opened 2022-06-16
Get block hash by its height

### LINKED ISSUE #603: Add `blockchain::GetBlockHash` trait and implement for blockchains
The new trait should have a function to get a block hash at a given height. The [Esplora API], [Electrum API], RPC, and CBF blockchains should all support getting a block hash at a given height.  This feature is needed to help support wallets that also use LDK, See https://github.com/bitcoindevkit/bdk-swift/issues/10.

[Esplora API]:https://github.com/Blockstream/esplora/blob/master/API.md#get-block-heightheight
[Electrum API]:https://electrumx.readthedocs.io/en/latest/protocol-methods.html#blockchain-block-header

vladimirfomene:
### Description
This PR create a new trait `blockchain::GetBlockHash` with a `get_block_hash` method which returns a block hash given the block height. This has been implemented for all blockchain backends. 
Fixes #603 

### Notes to the reviewers

I haven't updated the `CHANGELOG.md` and docs. Am I suppose to update it for this change?

#### All Submissions:

#### New Features:

### DIFF
--- a/src/blockchain/compact_filters/mod.rs
+++ b/src/blockchain/compact_filters/mod.rs
@@ -260,6 +260,16 @@ impl GetTx for CompactFiltersBlockchain {
     }
 }
 
+impl GetBlockHash for CompactFiltersBlockchain {
+    fn get_block_hash(&self, height: u64) -> Result<BlockHash, Error> {
+        self.headers
+            .get_block_hash(height as usize)?
+            .ok_or(Error::CompactFilters(
+                CompactFiltersError::BlockHashNotFound,
+            ))
+    }
+}
+
 impl WalletSync for CompactFiltersBlockchain {
     #[allow(clippy::mutex_atomic)] // Mutex is easier to understand than a CAS loop.
     fn wallet_setup<D: BatchDatabase>(
@@ -536,6 +546,8 @@ pub enum CompactFiltersError {
     InvalidFilter,
     /// The peer is missing a block in the valid chain
     MissingBlock,
+    /// Block hash at specified height not found
+    BlockHashNotFound,
     /// The data stored in the block filters storage are corrupted
     DataCorruption,
 
--- a/src/blockchain/electrum.rs
+++ b/src/blockchain/electrum.rs
@@ -98,6 +98,13 @@ impl GetTx for ElectrumBlockchain {
     }
 }
 
+impl GetBlockHash for ElectrumBlockchain {
+    fn get_block_hash(&self, height: u64) -> Result<BlockHash, Error> {
+        let block_header = self.client.block_header(height as usize)?;
+        Ok(block_header.block_hash())
+    }
+}
+
 impl WalletSync for ElectrumBlockchain {
     fn wallet_setup<D: BatchDatabase>(
         &self,
--- a/src/blockchain/mod.rs
+++ b/src/blockchain/mod.rs
@@ -21,7 +21,7 @@ use std::ops::Deref;
 use std::sync::mpsc::{channel, Receiver, Sender};
 use std::sync::Arc;
 
-use bitcoin::{Transaction, Txid};
+use bitcoin::{BlockHash, Transaction, Txid};
 
 use crate::database::BatchDatabase;
 use crate::error::Error;
@@ -87,7 +87,7 @@ pub enum Capability {
 
 /// Trait that defines the actions that must be supported by a blockchain backend
 #[maybe_async]
-pub trait Blockchain: WalletSync + GetHeight + GetTx {
+pub trait Blockchain: WalletSync + GetHeight + GetTx + GetBlockHash {
     /// Return the set of [`Capability`] supported by this backend
     fn get_capabilities(&self) -> HashSet<Capability>;
     /// Broadcast a transaction
@@ -108,6 +108,13 @@ pub trait GetHeight {
 pub trait GetTx {
     /// Fetch a transaction given its txid
     fn get_tx(&self, txid: &Txid) -> Result<Option<Transaction>, Error>;
+}
+
+#[maybe_async]
+/// Trait for getting block hash by block height
+pub trait GetBlockHash {
+    /// fetch block hash given its height
+    fn get_block_hash(&self, height: u64) -> Result<BlockHash, Error>;
 }
 
 /// Trait for blockchains that can sync by updating the database directly.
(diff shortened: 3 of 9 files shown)


src/blockchain/electrum.rs
```diff
@@ -30,6 +30,7 @@ use std::collections::{HashMap, HashSet};
 use log::{debug, error, info, trace};
 
 use bitcoin::{Transaction, Txid};
+use bitcoincore_rpc::RpcApi;
```
evanlinjin: Is this an accidental import?

src/blockchain/esplora/mod.rs
```diff
@@ -23,7 +23,9 @@ use std::io;
 
 use bitcoin::consensus;
 use bitcoin::{BlockHash, Txid};
+use bitcoincore_rpc::RpcApi;
```
evanlinjin: Is this also an accidental import?

evanlinjin (changes requested):

src/blockchain/esplora/mod.rs
notmandatory: I think this branch just needs a `cargo fmt` to fix imports and spacing. But it's still draft so will be fixed before ready for full review.

src/testutils/blockchain_tests.rs
```diff
@@ -1361,6 +1361,34 @@ macro_rules! bdk_blockchain_tests {
                 let finalized = wallet.sign(&mut psbt, Default::default()).unwrap();
                 assert_eq!(finalized, true);
             }
+
+            #[test]
+            fn test_get_block_hash() {
+                //create wallet with init_wallet
```
danielabrozzoni: tiny nit: add a space after `//`. Here and in all the other comments :)

src/testutils/blockchain_tests.rs
```diff
@@ -1361,6 +1361,34 @@ macro_rules! bdk_blockchain_tests {
...
+                let (_wallet, blockchain, _descriptors, mut test_client) = init_single_sig();
```
danielabrozzoni: nit: just use `_` for params you don't use, to improve readability

src/testutils/blockchain_tests.rs
```diff
@@ -1361,6 +1361,34 @@ macro_rules! bdk_blockchain_tests {
...
+
+                let height = test_client.bitcoind.client.get_blockchain_info().unwrap().blocks as u64;
+                let best_hash = test_client.bitcoind.client.get_best_block_hash().unwrap();
+
+                //use get_block_hash to get best block hash and compare with best_hash above
+                if let Some(block_hash) = blockchain.get_block_hash(height).unwrap() {
```
danielabrozzoni: In this way if `blockchain.get_block_hash(height).unwrap()` returns `None`, you don't complain, but I guess you should panic instead? What about doing:
```
let block_hash = blockchain.get_block_hash(height).unwrap().unwrap();
```

src/blockchain/mod.rs
```diff
@@ -110,6 +110,13 @@ pub trait GetTx {
     fn get_tx(&self, txid: &Txid) -> Result<Option<Transaction>, Error>;
 }
 
+#[maybe_async]
+/// Trait for getting block hash by block height
+pub trait GetBlockHash {
+    /// fetch block hash given its height
+    fn get_block_hash(&self, height: u64) -> Result<Option<BlockHash>, Error>;
```
danielabrozzoni: Here it seems like this method will return `Ok(None)` if the height is non-existent, `Ok(Some)` is everything went good, an Error in all the other cases. But then, when you implement it, you always either return an Error (using ?) or `Ok(Some(...))` - you never return `Ok(None)`.

Do you think it makes sense to change this method to:
```
fn get_block_hash(&self, height: u64) -> Result<BlockHash, Error> {
```

?

danielabrozzoni (commented):
Concept ACK, a couple of comments

src/blockchain/mod.rs
vladimirfomene: Good point! Yeah I think it will be best to change the signature of the method as you suggested given that I never return `Ok(None)`

src/testutils/blockchain_tests.rs
vladimirfomene: With the change to the method signature, this will just become
`let block_hash = blockchain.get_block_hash(height).unwrap();`

danielabrozzoni: Yep, it resolves itself :)

vladimirfomene:
@danielabrozzoni I have amended the code based on your recommendation. @evanlinjin I have also fixed the import issue. The test is passing for all blockchains but while testing against `rpc` blockchain, the following tests are failing:
failing-test

danielabrozzoni:
That looks like a spurious error...

In CI, compact filters compilation is failing. You can reproduce with `cargo build --features compact_filters --no-default-features`

src/blockchain/compact_filters/mod.rs
```diff
@@ -260,6 +260,18 @@ impl GetTx for CompactFiltersBlockchain {
...
 
+impl GetBlockHash for CompactFiltersBlockchain {
+    fn get_block_hash(&self, height: u64) -> Result<BlockHash, Error> {
+        let block_hash = self.headers.get_block_hash(height as usize)?;
+        match block_hash {
+            Some(block_hash) => Ok(block_hash),
+            None => Err(Error::CompactFilters(
+                CompactFiltersError::BlockHashNotFound,
+            )),
+        }
```
danielabrozzoni: nit: you can re-write this as
```
        self.headers.get_block_hash(height as usize)?.ok_or(Error::CompactFilters(
                CompactFiltersError::BlockHashNotFound,
            ))
```

`ok_or` makes it way more compact :)

danielabrozzoni (approved):
utACK d0efd4df2cdec79bca7d6bcc35d481a7546cae36

just a small nit

danielabrozzoni:
Also, you have conflicts with the changelog :(

vladimirfomene:
@danielabrozzoni, thanks for pointing that out. Fixed!

notmandatory (approved):
ACK 2af678aa846a82657d13bcda2ac0b4a9db4e2b80

Looks good !

PR #638, opened 2022-06-21
Fix CI, bump MSRV to 1.56.1

danielabrozzoni:
### Description
This PR fixes the CI, which is currently failing after a Github Actions update.
The MSRV is bumped to 1.56.1 (from 1.56.0), since that's `reqwest`'s current MSRV.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #640, opened 2022-06-23
Get balance in categories

### LINKED ISSUE #238: feature request: unconfirmed balances
At the moment `get_balance` returns the full wallet balance regardless of confirmations.

It would be nice to return unconfirmed balance as well or allow the caller to pass confirmation number.

wszdexdrf:
### Description
This changes `get_balance()` function so that it returns balance separated in 4 categories:
- available
- trusted-pending
- untrusted-pending
- immature

Fixes bitcoindevkit/bdk#238 

### Notes to the reviewers
Based on bitcoindevkit/bdk#614 

#### All Submissions:

#### New Features:

* [x] I've updated tests for the new feature


wszdexdrf:
I have not added tests since I feel that existing test should be fine. I can add some if need be.
Regarding the code, I went with the rule *"add instead of change"*

src/types.rs
```diff
@@ -240,6 +240,53 @@ impl BlockTime {
     }
 }
 
+/// Balance differentiated in various categories
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Default)]
+pub struct Balance {
+    /// All coinbase not matured yet
```
danielabrozzoni: nit: `All coinbase outputs not yet matured`

src/types.rs
```diff
@@ -240,6 +240,53 @@ impl BlockTime {
...
+    pub immature: u64,
+    /// In wallet change utxo, but not confirmed yet.
```
danielabrozzoni: nit: `Unconfirmed UTXO generated by a wallet tx` maybe? Does anyone have a better wording?

src/types.rs
```diff
@@ -240,6 +240,53 @@ impl BlockTime {
...
+    pub trusted_pending: u64,
+    /// Other wallet received utxo not confirmed yet.
+    pub untrusted_pending: u64,
```
danielabrozzoni, line 273: nit: `Unconfirmed UTXO received from an external wallet`

src/wallet/mod.rs
```diff
@@ -446,15 +446,51 @@ where
...
+    pub fn get_balance(&self) -> Result<Balance, Error> {
+        let mut immature = 0;
+        let mut trusted_pending = 0;
+        let mut untrusted_pending = 0;
+        let mut available = 0;
+        let utxos = self.list_unspent()?;
+        let database = self.database.borrow();
+        for u in utxos {
+            // Unwrap used since utxo set is created from database
+            let tx = database.get_tx(&u.outpoint.txid, true)?.unwrap();
```
danielabrozzoni: use `expect` instead of `unwrap`

src/wallet/mod.rs
```diff
@@ -4728,7 +4771,7 @@ pub(crate) mod test {
         // ...Now the coinbase is mature :)
         let mut builder = wallet.build_tx();
         builder
-            .add_recipient(addr.script_pubkey(), balance / 2)
+            .add_recipient(addr.script_pubkey(), balance.immature / 2)
```
danielabrozzoni: Why? At this point the coinbase is mature, the comment above says

src/wallet/mod.rs
```diff
@@ -4686,23 +4722,30 @@ pub(crate) mod test {
...
         let not_yet_mature_time = confirmation_time + COINBASE_MATURITY - 1;
         let maturity_time = confirmation_time + COINBASE_MATURITY;
 
-        // The balance is nonzero, even if we can't spend anything
-        // FIXME: we should differentiate the balance between immature,
-        // trusted, untrusted_pending
-        // See https://github.com/bitcoindevkit/bdk/issues/238
         let balance = wallet.get_balance().unwrap();
-        assert!(balance != 0);
+        assert!(balance.immature != 0);
```
danielabrozzoni: Check that the available amount is 0 here

src/wallet/mod.rs
```diff
@@ -4728,7 +4771,7 @@ pub(crate) mod test {
         // ...Now the coinbase is mature :)
         let mut builder = wallet.build_tx();
         builder
-            .add_recipient(addr.script_pubkey(), balance / 2)
+            .add_recipient(addr.script_pubkey(), balance.immature / 2)
             .set_current_height(maturity_time);
         builder.finish().unwrap();
```
danielabrozzoni, line 5353: Check that the available amount is non-zero, and that the immature amount is zero (after syncing)

danielabrozzoni (changes requested):
Concept ACK

- Some tests are failing, run them with `cargo test --features test-electrum` and `cargo test --features test-rpc`
- You should update the `CHANGELOG.md`
- You should add tests! Some ideas:
    + Receive a transaction, check that the output values get counted in the `untrusted_pending`, get it confirmed, check that now the amount is in `available`
    + Make a transaction, check that the change value is counted in `trusted_pending`, get it confirmed, check that now the amount is in `available`

src/wallet/mod.rs
wszdexdrf: The coins are mature for the supposed new transaction, but since the database is not updated hence they still are the same. For that to change, the sync time would need to be updated again. I am using `balance.immature` everywhere because it (the variable `balance`) was not updated hence all coins should be in immature category.

danielabrozzoni: Ok, then it's fine to leave this as is

danielabrozzoni:
Can you rebase to pick up the CI fixes (#638)? Thanks :)

src/wallet/mod.rs
```diff
@@ -445,15 +446,53 @@ where
...
+            // Unwrap used since utxo set is created from database
+            let tx = database
+                .get_tx(&u.outpoint.txid, true)?
+                .expect("Transaction not found in database");
+            if let Some(tx_conf_time) = &tx.confirmation_time {
+                if tx.transaction.expect("No transaction").is_coin_base() {
+                    let last_sync_height = self
+                        .database()
+                        .get_sync_time()?
+                        .map(|sync_time| sync_time.block_time.height);
```
afilini: This can be moved before the loop to avoid making extra unnecessary calls.

Moreover, I think you can just use `database` instead of `self.database()` since you've already borrowed it.

src/types.rs
```diff
@@ -240,6 +240,60 @@ impl BlockTime {
     }
 }
 
+/// Balance differentiated in various categories
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone, Default)]
```
afilini: I think here you should derive `Eq` as well, since this structure has the required properties:

> In addition to a == b and a != b being strict inverses, the equality must be (for all a, b and c):
> - reflexive: a == a;
> - symmetric: a == b implies b == a; and
> - transitive: a == b and b == c implies a == c.

src/types.rs
```diff
@@ -240,6 +240,60 @@ impl BlockTime {
...
+pub struct Balance {
+    /// All coinbase outputs not yet matured
+    pub immature: u64,
+    /// Unconfirmed UTXO generated by a wallet tx
+    pub trusted_pending: u64,
+    /// Unconfirmed UTXO received from an external wallet
+    pub untrusted_pending: u64,
+    /// All rest available balance
```
afilini: Maybe this is more clear if we document it as "Confirmed and immediately spendable balance"

afilini: ...which actually kind of opens a can of worms re: timelocked UTXOs.

In some cases it's pretty obvious what's timelocked and what's not, for example if your script is a simple `and(pk(key),after(value))`.

However, sometimes your timelock is only part of a branch, which means that your UTXOs may or may not be locked depending on how you'd like to spend them. I'm not sure what's the best way to fix this, ideally this function would take some kind of parameter to specify this, so that we can give an accurate answer to the user.

That said, I don't think it's a good idea to hold this PR while we discuss it, maybe we can open a different issue to keep track of that.

src/wallet/mod.rs
afilini: Actually, even better: if you move this out of the loop then you can use a match and if the value is `None` you can just return `Balance::default()` because it means you've never synced in the first place.

This also lets you remove the `if let Some(last_sync_height) = last_sync_height` and remove a bit of the nesting

src/types.rs
afilini: nit: `UTXOs` (plural)

src/types.rs
```diff
@@ -240,6 +240,60 @@ impl BlockTime {
...
+    pub available: u64,
+}
+
+impl Balance {
```
afilini, line 278: I would also add a `get_total()` here because just getting the whole balance is probably something most people will want to do

src/wallet/mod.rs
afilini: This ^

(in response to:  Check that the available amount is non-zero, and that the immature amount is zero (after syncing))

src/types.rs
```diff
@@ -240,6 +240,65 @@ impl BlockTime {
...
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Default)]
+pub struct Balance {
+    /// All coinbase outputs not yet matured
+    pub immature: u64,
+    /// Unconfirmed UTXOs generated by a wallet tx
+    pub trusted_pending: u64,
+    /// Unconfirmed UTXOs received from an external wallet
+    pub untrusted_pending: u64,
+    /// Confirmed and immediately spendable balance
+    pub available: u64,
```
evanlinjin: Would it make more sense to rename this as `confirmed`? Because, shouldn't `trusted_pending` be "available" in most cases?

src/types.rs
```diff
@@ -240,6 +240,65 @@ impl BlockTime {
     }
 }
 
+/// Balance differentiated in various categories
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Default)]
+pub struct Balance {
+    /// All coinbase outputs not yet matured
+    pub immature: u64,
```
evanlinjin, line 269: Would `immature_coinbase` be clearer? Or too wordy? :sweat_smile:

src/types.rs
```diff
@@ -240,6 +240,65 @@ impl BlockTime {
...
+    pub trusted_pending: u64,
+    /// Unconfirmed UTXOs received from an external wallet
+    pub untrusted_pending: u64,
+    /// Confirmed and immediately spendable balance
+    pub available: u64,
+}
+
+impl Balance {
+    /// Get sum of trusted_pending and available coins
+    pub fn get_trusted(&self) -> u64 {
```
evanlinjin: If we rename `available -> confirmed`, we can call this method `get_spendable`.

Would this be clearer? May need to ask everyone else's opinion.

src/wallet/mod.rs
```diff
@@ -445,15 +446,54 @@ where
...
+        {
+            Some(height) => height,
+            // None means database was never synced
+            None => return Ok(Balance::default()),
+        };
+        for u in utxos {
+            // Unwrap used since utxo set is created from database
+            let tx = database
+                .get_tx(&u.outpoint.txid, true)?
+                .expect("Transaction not found in database");
```
evanlinjin, line 487: Even though it would be the fault of the database, should we still return an error?

src/wallet/mod.rs
```diff
@@ -445,15 +446,54 @@ where
...
+            if let Some(tx_conf_time) = &tx.confirmation_time {
+                if tx.transaction.expect("No transaction").is_coin_base() {
+                    if (last_sync_height - tx_conf_time.height) >= COINBASE_MATURITY {
+                        available += u.txout.value;
+                    } else {
+                        immature += u.txout.value;
+                    }
+                } else {
+                    available += u.txout.value;
+                }
```
evanlinjin: I feel like we do not need nested `if-else` here.

```rust
if tx.is_coin_base() && coin_base_not_mature {
    immature += value;
} else {
    available += value;
}
```

evanlinjin (changes requested):
Concept ACK

I'm very excited for these changes, and great work!

I just have some thoughts on naming conventions and some changes in logic. Haven't really gone through everything yet though.

In regards to naming conventions, may need various second opinions.

afilini:
FYI: bitcoindevkit/bdk#614 got merged, so you can rebase on master to pickup the latest changes.

It also looks like there are some conflicts, so during the rebase you'll get a chance to fix them as well.

EDIT: I think the conflicts mostly come from the two commits you have in your branch from the old version of bitcoindevkit/bdk#614, so just dropping those and rebasing on master should fix everything.

src/wallet/mod.rs
danielabrozzoni: @wszdexdrf why has this been resolved? Is it fixed? It doesn't seem to me...

src/types.rs
danielabrozzoni: I like `confirmed`, but I'm wondering if then `untrusted_pending` is clear or not...

src/types.rs
danielabrozzoni: ACK for get_spendable :)

src/wallet/mod.rs
danielabrozzoni: IIRC we usually panic if we encounter db inconsistencies (such as this one), so I'd say it's ok to panic

src/wallet/mod.rs
danielabrozzoni: Oops, it was fixed a couple of lines above. Sorry!

src/types.rs
wszdexdrf: I feel `immature` coins are only ever meant in regards to coinbase coins, so I think it would be a bit too wordy. I thought of these names as if they had a imaginary "balance" or "coins" added at the end of them.

src/wallet/mod.rs
evanlinjin: I'll keep that in mind for my PRs!

src/wallet/mod.rs
```diff
@@ -446,15 +446,52 @@ where
...
-    pub fn get_balance(&self) -> Result<u64, Error> {
-        Ok(self
-            .list_unspent()?
-            .iter()
-            .fold(0, |sum, i| sum + i.txout.value))
+    pub fn get_balance(&self) -> Result<Balance, Error> {
+        let mut immature = 0;
+        let mut trusted_pending = 0;
+        let mut untrusted_pending = 0;
+        let mut available = 0;
```
danielabrozzoni: nit: should be called confirmed?

danielabrozzoni (approved):
tACK 74def0afa794b2124c288158e4453b6ec55028ed

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fee sniping discouraging through nLockTime - if the user specifies a `current_height`, we use that as a nlocktime, otherwise we use the last sync height (or 0 if we never synced)
 - Fix hang when `ElectrumBlockchainConfig::stop_gap` is zero.
 - Set coin type in BIP44, BIP49, and BIP84 templates
+- Return balance in separate categories, namely `available`, `trusted_pending`, `untrusted_pending` & `immature`.
```
afilini: `available` has been renamed to `confirmed`

Also I can see there are some whitespaces left at the end of the line, if you could remove them as well it would be great!

afilini (commented):
I hate to make you go through another round of fixing and rebasing but I swear, this is gonna be the last one!

The code looked great to me now, no complains there.

afilini (approved):
ACK 62e6b04fbeef1488ccdb09c04933a36fb26d6a92

src/types.rs
evanlinjin: That's an interesting point. If a coin is un-spendable due to a script-level timelock (such as OP_CLTV or OP_CSV) would we also use the term immature?

src/types.rs
evanlinjin: Untrusted unconfirmed? Maybe a little too much 😅

I personally think that untrusted_pending is clear enough (given our options):

- Untrusted: we don't know if someone will replace it.

- Pending: Not confirmed.

src/types.rs
danielabrozzoni: >  That's an interesting point. If a coin is un-spendable due to a script-level timelock (such as OP_CLTV or OP_CSV) would we also use the term immature?


Not at the moment! But I should open an issue to remember this. Also see https://github.com/bitcoindevkit/bdk/pull/640#discussion_r912108494

danielabrozzoni: ......which means that _maybe_ immature coinbase is clearer...

evanlinjin (approved):
ACK 62e6b04fbeef1488ccdb09c04933a36fb26d6a92

However, just a thought...

For tests that require checking balance, would it make sense to compare with the whole `Balance` struct?

For example:
```rust
assert_eq!(wallet.get_balance().unwrap(), Balance{ .. })
```

But I hate to be the one that delays merging, maybe we can have this as another task for another PR?

Otherwise, I've gone through everything and it all looks dandy to me :)

danielabrozzoni:
This needs rebasing to fix the conflicts.

One small note: please, when you've updated the code, can you click "resolve conversation" under the comments? This way the whole PR is way more readable. I've done it for you this time, I hope I didn't accidentally solve something that wasn't solved...

src/types.rs
danielabrozzoni: I've opened #676 to keep track of this, resolving this convo for now

src/wallet/mod.rs
```diff
@@ -5009,9 +5053,23 @@ pub(crate) mod test {
...
+            },
+        };
+        wallet
+            .database
+            .borrow_mut()
+            .set_sync_time(sync_time)
+            .unwrap();
+
+        let balance = wallet.get_balance().unwrap();
+        assert!(balance.immature == 0 && balance.confirmed != 0);
```
danielabrozzoni: Following evan's suggestion, I think that here we should assert_eq on the whole balance structure, just to make sure that if we have some weird inconsistencies they don't go unnoticed

danielabrozzoni (commented):
Just a tiny nit. Me and Alekos have been recently talking about lacking test coverage (mainly after we discovered #666 😈 ), and we realize that we should definitely be more nit-picker on tests. Sorry to delay merging even more, but I promise we're close!

src/testutils/blockchain_tests.rs
```diff
@@ -1059,20 +1072,25 @@ macro_rules! bdk_blockchain_tests {
...
                 wallet.sync(&blockchain, SyncOptions::default()).unwrap();
-                assert_eq!(wallet.get_balance().unwrap(), 0, "incorrect balance");
+                assert_eq!(wallet.get_balance().unwrap().immature, 0, "incorrect balance");
 
                 test_client.generate(1, Some(wallet_addr));
 
-                #[cfg(feature = "rpc")]
+                wallet.sync(&blockchain, SyncOptions::default()).unwrap();
+
+                #[cfg(not(feature="rpc"))]
```
danielabrozzoni: I swear I was about to merge when I noticed this...

Why is this line here? Shouldn't we just have an assert of the immature balance for every backend (not only "not rpc"), and then the code for making it mature?

```
// we consider the coinbase only when mature (100 blocks)
assert!(wallet.get_balance().unwrap().immature > 0, "incorrect balance after receiving coinbase");

// make it mature
let node_addr = test_client.get_node_address(None);
test_client.generate(100, Some(node_addr));
```

src/testutils/blockchain_tests.rs
```diff
@@ -486,16 +486,24 @@ macro_rules! bdk_blockchain_tests {
                 let (wallet, blockchain, descriptors, mut test_client) = init_single_sig();
 
                 wallet.sync(&blockchain, SyncOptions::default()).unwrap();
-                assert_eq!(wallet.get_balance().unwrap(), 0);
+                assert_eq!(wallet.get_balance().unwrap().untrusted_pending, 0);
```
danielabrozzoni: nit: this should be `get_total()`? It doesn't seem to me that there's a specific reason for using untrusted_pending, and we expect to have all the balance to zero at this point

wszdexdrf: That was like so because (1) when that line was added, `get_total()` didn't exist, and (2) Since I was checking untrusted pending later (equal to 50,000) I thought I should show only it to be equal to 0 before. I should probably change it to `get_total()`, because of... stricter testing 😅

src/testutils/blockchain_tests.rs
wszdexdrf: That is so because, what I could get from the previous code (and the accompanying comment) is that rpc doesn't report immature coins in balance. A quick test without it shows exactly that. Before the code only "matured" coinbase if backend in use was rpc, now I just test immature coins in all cases other that rpc, and mature coins anyway.

danielabrozzoni: Ok, but at this point if you drop the `#[cfg(not(feature="rpc"))]` and check the immature for the rpc as well, the tests pass, don't they?

wszdexdrf: Nope. Balance is reported as 0, 0, 0, 0. Atleast for me, with `test-rpc` feature.

danielabrozzoni: Ouch, yeah, now I see it. You won't be happy to hear this, but I really think this should be fixed before merging...

danielabrozzoni:
Rebased

afilini (approved):
ACK 0f03831274d3aa69da6e89729c65d66530bbd752

PR #607, opened 2022-05-16
BIP39 implementation

atalw:
### Description

This PR is a custom BIP39 implementation which closes #561. 

There are a couple of features that have been implemented
- Parse mnemonic string to `Mnenomic` type
- Generate `Mnemonic` from entropy
- Derive seed from Mnemonic (with and without passphrase)
- All language wordlists (with verification test to ensure they were untampered)
- Mnemonic test vectors from [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#test-vectors)
- Error handling

### Notes to the reviewers

I've tried to keep the function names the same as they were to maintain backward compatibility. There is however a file structure change in the favour of clarity.

I was initially using tabs so earlier commits may fail CI. I'll amend if needed.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [x] I've added tests for the new feature
* [x] I've added docs for the new feature
* [ ] I've updated `CHANGELOG.md`

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [x] I'm linking the issue being fixed by this PR

PR #644, opened 2022-06-29
BIP39 Implementation

### LINKED ISSUE #561: Write own BIP39 implementation

evanlinjin:
### Description

This is a continuation of PR #607 which closes #561

This PR includes commits for own implementation of `PBKFD2`.
I've also modified the `.gitignore`, I hope that is okay.

### Notes to the reviewers

Although complete, I still have some security concerns for the current implementation ([please check my comment below](https://github.com/bitcoindevkit/bdk/pull/644#issuecomment-1172943386)).

#### All Submissions:

#### New Features:

* [x] Add `pbkfd2` implementation


evanlinjin:
I'm thinking we can potentially do our own implementation for `unicode_normalization` as well?

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 - New MSRV set to `1.56.1`
 - Fee sniping discouraging through nLockTime - if the user specifies a `current_height`, we use that as a nlocktime, otherwise we use the last sync height (or 0 if we never synced)
+- Changed: Introduce own implementation of `bip39`.
```
danielabrozzoni: nit: remove the initial `Changed:`

src/keys/bip39/mod.rs
```diff
@@ -37,9 +88,149 @@ pub enum WordCount {
...
+        let mut indices: Vec<u16> = Vec::with_capacity(words.len());
+        let wordlist = language.wordlist();
+        let mut mnemonic_bits = vec![false; words.len() * 11];
+        for (i, word) in words.iter().enumerate() {
+            let idx = match wordlist.iter().position(|&x| x == *word) {
+                Some(i) => i as u16,
+                None => return Err(Error::InvalidWord(word.to_string())),
+            };
+
+            indices.push(idx);
```
danielabrozzoni: @vladimirfomene was saying, in #607:

> After calculating the indices for every word, you may want to recover the entropy, then calculate the checksum to compare it against the binary representation of mnemonic. In case, the calculated checksum does not match the checksum in the binary representation of the mnemonic, you will want to throw an error.


Looks like it's still a todo, so I'm posting it here

src/keys/bip39/mod.rs
```diff
@@ -37,9 +88,149 @@ pub enum WordCount {
...
+                }
+            }
+            words[i] = index;
+        }
+
+        Ok(Mnemonic { language, words })
+    }
+
+    /// Convert a mnemonic to a seed with an optional passphrase
+    fn to_seed(&self, passphrase: Option<String>) -> Seed {
```
danielabrozzoni: From #607:

@vladimirfomene: It might not be a good idea to change the type of passphrase from &str to option<string> as that has the potential of breaking code which consumes this method.

@atalw: It makes sense to have the passphrase as Option as it really is optional, so if a breaking change is okay we can go ahead with this.

-----------

Personally, I agree that we should try not to break the API, and leave `P: Into<Cow<'a, str>` (https://docs.rs/bip39/latest/src/bip39/lib.rs.html#479-486) here

danielabrozzoni (commented):
Copy-pasted some review comments from #607 that it seems still need to be addressed

src/keys/bip39/mod.rs
evanlinjin: In my opinion, having a `P: Into<Cow<'a, str>>` makes little sense based on our implementation. I propose just using a `&str`, this way for most people, the API shouldn't break.

CHANGELOG.md
evanlinjin: Resolved in b0033915e379cd365d2c1953b7ed6c6aee76679e

src/keys/bip39/mod.rs
evanlinjin: Resolved in b0033915e379cd365d2c1953b7ed6c6aee76679e

src/keys/bip39/mod.rs
evanlinjin: Changed in dd42594c78455d2e71c8cae86746ce9c0a2031c5. Let me know if it is sufficient!

https://github.com/bitcoindevkit/bdk/blob/dd42594c78455d2e71c8cae86746ce9c0a2031c5/src/keys/bip39/mod.rs#L204-L208

evanlinjin:
I still have some security concerns regarding the current implementation (although, I am no security expert, just based on what I've read on the internet). I will list them here, and hopefully someone knowledgeable enough could provide some clarity.

1. Should we avoid using heap memory, and keep everything on the stack? Apparently, heap memory is more prone to exploints. Reference: [github.com/shellphish/how2heap](https://github.com/shellphish/how2heap)

2. Is implementing `std::fmt::Display` and `Debug` a good idea? As it may leak the secrets to logs. We can potentially remove these implementations completely, or provide implementations with redacted secrets, or use an a crate such as [`secrey`](https://docs.rs/secrecy/latest/secrecy/index.html).

Thank you all in advance!

P.S. Test [`blockchain::esplora::bdk_blockchain_tests::test_sync_stop_gap_20`](https://github.com/bitcoindevkit/bdk/runs/7163681328?check_suite_focus=true) seems to fail occasionally.

danielabrozzoni:
> Should we avoid using heap memory, and keep everything on the stack? Apparently, heap memory is more prone to exploints. Reference: [github.com/shellphish/how2heap](https://github.com/shellphish/how2heap)

We could look into that, but it's better if we do so in a new PR. This one is already quite big, and the bigger it gets, the more difficult it is to collect reviews :)

> Is implementing std::fmt::Display and Debug a good idea? As it may leak the secrets to logs. We can potentially remove these implementations completely, or provide implementations with redacted secrets, or use an a crate such as [secrey](https://docs.rs/secrecy/latest/secrecy/index.html).

This, instead, I think should be tackled here: for now, avoiding Debug and Display (or manually implementing a really generic one) should be enough (with an appropriate comment on why we do so). I'd avoid adding YA dependency :)

evanlinjin:
Another aspect I've been thinking about, is the great majority of the time people will be using English (which shouldn't require Unicode normalization). For the passphrase, we can do a check only (and fail if not normalized).

Since normalization sometimes requires resizing the vector (so it's a heap operation), and for most people, it also means one less dependably.

evanlinjin:
> This, instead, I think should be tackled here: for now, avoiding Debug and Display (or manually implementing a really generic one) should be enough (with an appropriate comment on why we do so). I'd avoid adding YA dependency :)

Addressed in 08e1cc9ebfdb5f9e0f4bfd929eb71b94eedea89c.

afilini:
> Should we avoid using heap memory, and keep everything on the stack?

One advantage of this (on top of the extra safety) is that it would be much easier to then port to embedded hardware. We have many features in bdk which I guess are not really fit for hardware wallets, but mnemonics are for sure one that we'll need to have.

Are you able to do a rough estimation of how much longer/how much harder it would be to implement in this way?

src/keys/bip39/pbkdf2.rs
```diff
@@ -0,0 +1,87 @@
...
+        let block_index = (i + 1) as u32;
+
+        xor_sum(&hmac_prf, salt, iter_count, block_index, block);
+    }
+}
+
+/// Password is the UTF8-NFKD-normalized result of mnemonic words separated by space.
+fn make_password<'a, W>(words: W) -> String
+where
+    W: Iterator<Item = &'a str>,
```
afilini, line 64: You could define `Item` as another generic that implements `AsRef<str>`. This should allow you to pass a vec of strings as well if you want

src/keys/bip39/pbkdf2.rs
```diff
@@ -0,0 +1,87 @@
...
+{
+    words.collect::<Vec<&str>>().join(" ").nfkd().to_string()
+}
+
+/// Salt is the UTF8-NFKD-normalized result of (SALT_PREFIX + passphrase).
+fn make_salt(passphrase: &str) -> Cow<'static, str> {
+    let mut salt = Cow::from(SALT_PREFIX);
```
afilini, line 71: I understand the small performance benefit of reusing the ref as-is, but I don't think it's worth using `Cow` here especially considering that moving forward we'd like to avoid using the heap (even if we don't manage to finalize that transition in this PR)

src/keys/bip39/pbkdf2.rs
```diff
@@ -0,0 +1,87 @@
...
+const ITERATION_COUNT: u32 = 2048; // PBKDF2 iteration count
+const BLOCK_LEN: usize = sha512::Hash::LEN; // PBKDF2 block size
+
+/// Hmac engine used as the pseudo-random function.
+type HmacPRF = HmacEngine<sha512::Hash>;
+
+/// Make hmac-sha512 engine from password.
+/// The hmac engine is used as the pseudo-random function.
+fn make_prf(password: &str) -> HmacPRF {
+    HmacEngine::new(password.as_bytes())
```
afilini, line 27: I was gonna comment here that for extra safety we should re-normalize the string, but then I realized: wouldn't it be better to immediately convert strings to `&[u8]` **immediately** after normalization?

This would be like a marker for us, anything that's `&str` or similar is potentially not normalized, but as soon as we are done we just convert to bytes and forget about it.

With this change I guess you would make this function take a `&[u8]` directly, and do the conversion in the caller which as far as I can see is already normalizing correctly.

src/keys/bip39/pbkdf2.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+/// Generate block (of given block_index) by calculating xor sum of iterations of PRF.
+fn xor_sum(hmac_prf: &HmacPRF, salt: &str, iter_count: u32, block_index: u32, block: &mut [u8]) {
+    // for the first iteration, we concat: salt + block_index (as big-endian bytes)
+    let mut prev_u = Vec::with_capacity(salt.len() + 4);
```
afilini, line 33: Since the length is fixed i guess this could also be an array with static length (there should be a constant in bitcoin_hashes for this).

I think the code would probably still look decently good with `copy_from_slice`: https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice

afilini (commented):
This is just a partial review, I still haven't looked at all the files.

I just wanted to post this comments so that you could start thinking about them and see if they make any sense.

evanlinjin:
> > Should we avoid using heap memory, and keep everything on the stack?
> 
> One advantage of this (on top of the extra safety) is that it would be much easier to then port to embedded hardware. We have many features in bdk which I guess are not really fit for hardware wallets, but mnemonics are for sure one that we'll need to have.
> 
> Are you able to do a rough estimation of how much longer/how much harder it would be to implement in this way?

Less than a week. But I'm stuck into multi descriptor wallet business 😅😂

afilini:
Yes, multi-descriptor is definitely the priority right now. We'll get back to this once you are done there :)

.gitignore
```diff
@@ -3,3 +3,6 @@ Cargo.lock
 
 *.swp
 .idea
+
+# IDE
```
vladimirfomene, line 7: Nits: `.idea` is for IntelliJ. I don't know if it is necessary to have that `IDE` comment.

src/keys/bip39/mod.rs
```diff
@@ -0,0 +1,975 @@
...
+        }
+
+        let word_to_index_map = language.word_map();
+        let mut word_indexes = Vec::with_capacity(MS_MAX); // word indexes
+        let mut ent_cs_bits = Vec::with_capacity(MS_MAX * utils::U11_BITS); // ENT+CS bits
+
+        // parse word indexes and ENT+CS bits from mnemonic words
+        let parse_result = sentence_words
+            .iter()
+            .map(|&word| word_to_index_map.get(word).unwrap_or(&utils::U11_EOF))
```
vladimirfomene, line 170: Why not throw an invalid word error here if the word is not in the HashMap? What is the utility of having `&utils::U11_EOF` as the default value here?

src/keys/bip39/mod.rs
```diff
@@ -0,0 +1,975 @@
...
+        let ms = sentence_words.len();
+
+        // check word count is valid
+        if ms % MS_MULTIPLE != 0 || ms < MS_MIN || ms > MS_MAX {
+            return Err(Error::InvalidWordCount(ms));
+        }
+
+        let word_to_index_map = language.word_map();
+        let mut word_indexes = Vec::with_capacity(MS_MAX); // word indexes
+        let mut ent_cs_bits = Vec::with_capacity(MS_MAX * utils::U11_BITS); // ENT+CS bits
```
vladimirfomene, line 165: why not just use `ms` as your vector size instead of `MS_MAX`?

src/keys/bip39/mod.rs
```diff
@@ -0,0 +1,975 @@
...
+
+        // parse word indexes and ENT+CS bits from mnemonic words
+        let parse_result = sentence_words
+            .iter()
+            .map(|&word| word_to_index_map.get(word).unwrap_or(&utils::U11_EOF))
+            .try_for_each(|word_index| {
+                if *word_index > utils::U11_MAX {
```
vladimirfomene, line 172: Given that you are getting the `word_index` from the `word_map` is there a scenario where the `word_index` will be greater than `utils::U11_MAX`. I'm thinking if you throw an error for invalid words there will be no need for this if/else logic.

examples/rpcwallet.rs
```diff
@@ -217,13 +218,10 @@ fn main() -> Result<(), Box<dyn Error>> {
...
     // Generate a random mnemonic, and use that to create a "DerivableKey"
-    let mnemonic: GeneratedKey<_, _> = Mnemonic::generate((WordCount::Words12, Language::English))
-        .map_err(|e| e.expect("Unknown Error"))?;
-
-    // `Ok(mnemonic)` would also work if there's no passphrase and it would
-    // yield the same result as this construct with `password` = `None`.
-    Ok((mnemonic, password))
+    let mnemonic_with_passphrase: GeneratedKey<_, _> =
+        MnemonicWithPassphrase::generate((WordCount::Words12, Language::English, password))?;
+    Ok(mnemonic_with_passphrase)
```
vladimirfomene, line 226: Given that there is a change in the examples, I believe this will affect users. Is it possible to implement the BIP in such a way that it doesn't change anything for users.

src/keys/bip39/pbkdf2.rs
```diff
@@ -0,0 +1,87 @@
+// Bitcoin Dev Kit
+// Written in 2020 by Alekos Filini <alekos.filini@gmail.com>
+//
+// Copyright (c) 2020-2022 Bitcoin Dev Kit Developers
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
```
vladimirfomene, line 10: It will be great to have a reference to the PBKDF2 RFC link as part of this module's documentation. https://datatracker.ietf.org/doc/html/rfc2898

src/keys/bip39/wordlists/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+            Language::Italian => &italian::WORDS,
+            #[cfg(feature = "czech")]
+            Language::Czech => &czech::WORDS,
+            #[cfg(feature = "portuguese")]
+            Language::Portuguese => &portuguese::WORDS,
+        }
+    }
+
+    /// Generate word map for given language.
+    pub fn word_map(&self) -> HashMap<&str, u16> {
```
vladimirfomene, line 85: can we write a test for this method?

vladimirfomene (commented):
Thanks for helping us move this forward! Just a couple of questions and comments.

src/keys/bip39/wordlists/mod.rs
evanlinjin: What kind of test are you suggesting?

vladimirfomene: I was thinking of writing a test to make sure we have correct word to indices mapping in the hashmap.

rajarshimaitra:
Is this a good idea to have it in bdk_core eventually? Or we wanna do key generation outside of core separately?

danielabrozzoni:
We closed #561, let's close this one as well :)

PR #645, opened 2022-06-30
Allow signing only specific leaf hashes

### LINKED ISSUE #616: Allow signing only specific leaf hashes
Currently our signer will sign everything it can, including all the taproot leaves that contain the key. We should add an option to choose which leaf hashes we should sign (plus the internal key sig if we can) so that the user has control over the process.

danielabrozzoni:
Fixes #616

#### All Submissions:

#### New Features:

### DIFF
--- a/src/wallet/signer.rs
+++ b/src/wallet/signer.rs
@@ -58,6 +58,7 @@
 //!         &self,
 //!         psbt: &mut psbt::PartiallySignedTransaction,
 //!         input_index: usize,
+//!         _sign_options: &SignOptions,
 //!         _secp: &Secp256k1<All>,
 //!     ) -> Result<(), SignerError> {
 //!         self.device.hsm_sign_input(psbt, input_index)?;
@@ -241,6 +242,7 @@ pub trait InputSigner: SignerCommon {
         &self,
         psbt: &mut psbt::PartiallySignedTransaction,
         input_index: usize,
+        sign_options: &SignOptions,
         secp: &SecpCtx,
     ) -> Result<(), SignerError>;
 }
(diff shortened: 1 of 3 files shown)


afilini (commented):
Concept ACK, but I would also add a way to disable signing for the internal key (which is required if you want to force the use of a a script-spend path)

danielabrozzoni:
>  I would also add a way to disable signing for the internal key (which is required if you want to force the use of a a script-spend path)

Done :)

danielabrozzoni:
Rebased

src/wallet/signer.rs
```diff
@@ -677,6 +696,39 @@ pub struct SignOptions {
...
+pub enum TapLeavesOptions {
+    /// The signer will sign all the leaves it has a key for.
+    All,
+    /// The signer won't sign leaves other than the ones specified. Note that it could still ignore
+    /// some of the specified leaves, if it doesn't have the right key to sign them.
+    Include(Vec<taproot::TapLeafHash>),
+    /// The signer won't sign the specified leaves.
+    Exclude(Vec<taproot::TapLeafHash>),
+    /// The signer won't sign any leaf. This effectively means that it will try signing only for
+    /// the key-path.
```
afilini: I would remove the comment about "effectively signing for the key-path" because that's controlled by another option

src/wallet/signer.rs
```diff
@@ -401,26 +410,36 @@ impl InputSigner for SignerWrapper<PrivateKey> {
...
-                    sign_psbt_schnorr(
-                        &self.inner,
-                        x_only_pubkey,
-                        Some(lh),
-                        &mut psbt.inputs[input_index],
-                        hash,
-                        hash_ty,
-                        secp,
-                    );
+                if sign_options.tap_leaves_options != TapLeavesOptions::None {
```
afilini: I would remove this if and move it into the match. `None` always returns false, `All` always true

danielabrozzoni:
Fixed

afilini (approved):
ACK a713a5a0629c9a643708a4b0d8d6ac3a87a13195

PR #647, opened 2022-06-30
Multi descriptor: Decouple transaction building logic

### LINKED ISSUE #188: BDK Descriptor Improvements: Add multi descriptor capability in wallet
**Description**  
  
In order for BDK to extend beyond just a Bitcoin wallet, it needs a more powerful descriptor capability. This project aims at improving the descriptors used in the library. Currently BDK uses only a single descriptor for a single wallet. Giving a new descriptor to BDK will create a brand new wallet.

The goal of this project is to create a wallet that can handle multiple descriptors within itself (a "super wallet"), and can send funds from any or all of them together.   

**Expected Outcomes**  
 

- Through understanding of [descriptors](https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md).
- Basic understanding of the [rust-miniscript](https://github.com/rust-bitcoin/rust-miniscript) library.
- Multi descriptor capability in BDK. 

**Resources**  
  
  
  
  
  

**Skills Required**  
  
  
  
  

- Experience with git. [Guide](https://git-scm.com/book/en/v2)
- Basic familiarity with rust. [First seven chapters of the book](https://doc.rust-lang.org/book/)
- Familiarity with BDK’s descriptor usage.

**Mentor(s)**  

**Difficulty**: Hard  

**Competency Test (optional)**  

- Install rust, compile and run all bdk examples and tests. 
- Read through the BDK docs.
- Make a dummy wallet with BDK with different blockchain backends. 
- Familiarity with basic rust, should be able to write basic custom trait implementations on foreign structures.

### LINKED ISSUE #486: BDK Descriptor Improvements: 02 - Add multi descriptor capability in wallet
**Description**  
  
In order for BDK to extend beyond just a Bitcoin wallet, it needs a more powerful descriptor capability. This project aims at improving the descriptors used in the library. Currently BDK uses only a single descriptor for a single wallet. Giving a new descriptor to BDK will create a brand new wallet.

Add multi descriptor capability in the wallet, so a single wallet can handle multiple descriptors. This can be useful for many lightning applications.

**Expected Outcomes**  
 

- Through understanding of [descriptors](https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md).
- Basic understanding of the [rust-miniscript](https://github.com/rust-bitcoin/rust-miniscript) library.
- Multi descriptor capability in BDK. 

**Resources**  
  
  
  
  
  

**Skills Required**  
  
  
  
  

- Experience with git. [Guide](https://git-scm.com/book/en/v2)
- Basic familiarity with rust. [First seven chapters of the book](https://doc.rust-lang.org/book/)
- Familiarity with BDK’s descriptor usage.

**Mentor(s)**  

**Difficulty**: Hard  

**Competency Test (optional)**  

- Install rust, compile and run all bdk examples and tests. 
- Read through the BDK docs.
- Make a dummy wallet with BDK with different blockchain backends. 
- Familiarity with basic rust, should be able to write basic custom trait implementations on foreign structures.

evanlinjin:
## Description

Closes bitcoindevkit/bdk_wallet#188

The end goal of this PR is to have the ability to a single wallet with multiple descriptors.

## Notes to the reviewers

In the Discord channel @ `#multi-descriptor-wallets`, @afilini [suggested a roadmap](https://discord.com/channels/753336465005608961/991846404068749405/993812164685082655):

> I think a reasonable roadmap would be:
> 1. pr to add "database factories" or something similar
> 2. pr to add "multi-wallet" support (collection of the current Wallet struct). The subwallets will only use the "external" keychain, and at the multi-wallet level you can define what to use as internal/external
> 3. pr to remove the two keychains from the wallet struct, deal with renaming and deprecating stuff (users should switch from Wallet to MultiWallet with two subwallets etc)

The "database factories" structure is already implemented in a separate PR: bitcoindevkit/bdk#654 (And can be used later).

This PR mainly addresses the initial portion of step 2.

### My Approach

The most difficult aspect of adding multi-descriptor capability is in the transaction creation. Everything else from retrieving address, UTXOs, signing, transactions, balances and syncing are trivial tasks. Hence, my approach is to first decouple the transaction building logic from `Wallet` so that it can be reused for multi-descriptor (or even for users of BDK that don't want a wallet). In that being said, I have kept what @afilini said in mind (paraphrasing)...

* Avoid resulting in ***a huge pr that breaks everything***.
* ... ***I'm sure it's gonna take a while to get pr 2 to work well and it's probably better if we avoid replacing the current `Wallet` with it immediately.***

Therefore, I have ensured the following things for my approach:

1. Instead of moving logic out of `Wallet`, I have turned various member functions into static functions (renamed to start with `_`). The original functions will call these static functions. This way, we do not change the "external interface" of `Wallet` and more-or-less keep these bits and pieces in the same place (so hopefully git will not complain so much when we merge).

2. I have introduced a new trait `SpendableCollection` to replace the functionality of `Wallet`/`Database` within `TxBuilder`. To keep `Wallet` working with the new modified `TxBuilder`, a struct `SpendablesFromDb` is formed which implements `SpendableCollection`, and wraps `Database` (and some fields of `Wallet`).

### Noteworthy changes

Not in any particular order...

* Introduction of the `SpendableCollection` trait (and `SpendableCollectionInner` which defines it's "explicitly required" methods). These traits represent a collection of owned and spendable UTXO's, Descriptors and their associated transactions. Note that it's members `SpendableCollectionInner::get_{external|internal}_descriptors` returns `Vec`s (so we can support multiple descriptor implementations in the very near future).
* The `CoinSelectionAlgorithm` trait's type constraint has been moved to it's function definition (the constrained type was only used as input anyway) - I hope this change is okay?
* Instead of referencing descriptors by `KeychainKind`, I have changed to use the `ExtendedDescriptor` directly (again, for muti-descriptors in the future). This change has some small caveats...
  * https://github.com/bitcoindevkit/bdk/blob/574dcc4ee3f045cf5a3a0cc61667b5aeff370be5/src/wallet/spendable_collection.rs#L73-L101
  * https://github.com/bitcoindevkit/bdk/blob/574dcc4ee3f045cf5a3a0cc61667b5aeff370be5/src/wallet/mod.rs#L5279-L5283
* `TxBuilder`, other than needing the `BatchDatabase` constraint with `SpendableCollection`, uses static members from `Wallet` (as mentioned above). Hence, we needed to add another phantom constraint imposed by the constraints of those `Wallet` methods (this can be removed once we ACTUALLY move stuff out of `Wallet`).
* TODO @evanlinjin: I have probably forgotten to mention something here... I will keep this space updated.

### Tests

It seems that I have not broken the functionality of `Wallet` and all the non-doc tests continue to pass.

### Next steps

Once @afilini and everyone else is happy with my approach thus far, I will create a multi-descriptor implementation of `SpendableCollection` and do tests to see if multi-descriptor transaction creation works as expected.

Since `CoinSelectionAlgorithm` now takes a `SpendableCollection` as an input (instead of a `BatchDatabase`, I need to ensure that `SpendableCollection` is satisfactory for all coin selection algorithms.

I also still need to decouple Fee-bumping logic.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -141,15 +141,15 @@ impl CoinSelectionResult {
 
 /// Trait for generalized coin selection algorithms
 ///
-/// This trait can be implemented to make the [`Wallet`](super::Wallet) use a customized coin
-/// selection algorithm when it creates transactions.
+/// This trait can be implemented to make the [`TxBuilder`](super::tx_builder::TxBuilder) use a
+/// customized coin selection algorithm when it creates transactions.
 ///
 /// For an example see [this module](crate::wallet::coin_selection)'s documentation.
 pub trait CoinSelectionAlgorithm: std::fmt::Debug {
     /// Perform the coin selection
     ///
-    /// - `database`: a reference to the wallet's database that can be used to lookup additional
-    ///               details for a specific UTXO
+    /// - `multi_tracker`: a reference to the multi-descriptor tracker that can be used to lookup
+    ///                     additional details for a specific UTXO
     /// - `required_utxos`: the utxos that must be spent regardless of `amount_needed` with their
     ///                     weight cost
     /// - `optional_utxos`: the remaining available utxos to satisfy `amount_needed` with their
@@ -158,9 +158,9 @@ pub trait CoinSelectionAlgorithm: std::fmt::Debug {
     /// - `amount_needed`: the amount in satoshi to select
     /// - `fee_amount`: the amount of fees in satoshi already accumulated from adding outputs and
     ///                 the transaction's header
-    fn coin_select<Sc: MultiTracker>(
+    fn coin_select<Mt: MultiTracker>(
         &self,
-        database: &Sc,
+        multi_tracker: &Mt,
         required_utxos: Vec<WeightedUtxo>,
         optional_utxos: Vec<WeightedUtxo>,
         fee_rate: FeeRate,
@@ -177,9 +177,9 @@ pub trait CoinSelectionAlgorithm: std::fmt::Debug {
 pub struct LargestFirstCoinSelection;
 
 impl CoinSelectionAlgorithm for LargestFirstCoinSelection {
-    fn coin_select<D: MultiTracker>(
+    fn coin_select<Mt: MultiTracker>(
         &self,
-        _database: &D,
+        _multi_tracker: &Mt,
         required_utxos: Vec<WeightedUtxo>,
         mut optional_utxos: Vec<WeightedUtxo>,
         fee_rate: FeeRate,
@@ -215,9 +215,9 @@ impl CoinSelectionAlgorithm for LargestFirstCoinSelection {
 pub struct OldestFirstCoinSelection;
 
 impl CoinSelectionAlgorithm for OldestFirstCoinSelection {
-    fn coin_select<D: MultiTracker>(
+    fn coin_select<Mt: MultiTracker>(
         &self,
-        database: &D,
+        multi_tracker: &Mt,
         required_utxos: Vec<WeightedUtxo>,
         mut optional_utxos: Vec<WeightedUtxo>,
         fee_rate: FeeRate,
@@ -234,7 +234,7 @@ impl CoinSelectionAlgorithm for OldestFirstCoinSelection {
                     if bh_acc.contains_key(&txid) {
                         Ok(bh_acc)
                     } else {
-                        database.get_tx(&txid).map(|details| {
+                        multi_tracker.get_tx(&txid).map(|details| {
                             let conf_time =
                                 details.and_then(|tx| tx.confirmed.map(|conf| conf.height));
                             bh_acc.insert(txid, conf_time);
@@ -362,9 +362,9 @@ impl BranchAndBoundCoinSelection {
 const BNB_TOTAL_TRIES: usize = 100_000;
 
 impl CoinSelectionAlgorithm for BranchAndBoundCoinSelection {
-    fn coin_select<D: MultiTracker>(
+    fn coin_select<Mt: MultiTracker>(
         &self,
-        _database: &D,
+        _multi_tracker: &Mt,
         required_utxos: Vec<WeightedUtxo>,
         optional_utxos: Vec<WeightedUtxo>,
         fee_rate: FeeRate,
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -236,7 +236,7 @@ impl CoinSelectionAlgorithm for OldestFirstCoinSelection {
                     } else {
                         database.get_tx(&txid).map(|details| {
                             let conf_time =
-                                details.and_then(|(_, bt_opt)| bt_opt.map(|bt| bt.height));
+                                details.and_then(|tx| tx.confirmed.map(|conf| conf.height));
                             bh_acc.insert(txid, conf_time);
                             println!(
                                 "OldestFirstCoinSelection: insert({}, {:?})",
@@ -628,11 +628,6 @@ mod test {
 
     const FEE_AMOUNT: u64 = 50;
 
-    // static DESCRIPTOR: ExtendedDescriptor =
-    //     "wsh(pk(cRYhyJQzA92kLQUtJZbVVQrBoQwSm4WT23sanTXYrKoUMjX8ph1Z))"
-    //         .into_wallet_descriptor(&SecpCtx::new(), Network::Regtest)
-    //         .unwrap()
-    //         .0;
     fn new_desc() -> ExtendedDescriptor {
         "wsh(pk(cRYhyJQzA92kLQUtJZbVVQrBoQwSm4WT23sanTXYrKoUMjX8ph1Z))"
             .into_wallet_descriptor(&SecpCtx::new(), Network::Regtest)
@@ -1025,10 +1020,7 @@ mod test {
         let desc = new_desc();
         let db = new_db();
         let cache = new_collection(&desc, &db);
-        let utxos = {
-            // let database: &mut MemoryDatabase = cache.borrow_mut();
-            setup_database_and_get_oldest_first_test_utxos(&mut *db.borrow_mut())
-        };
+        let utxos = { setup_database_and_get_oldest_first_test_utxos(&mut *db.borrow_mut()) };
 
         let result = OldestFirstCoinSelection::default()
             .coin_select(
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -26,14 +26,14 @@
 //! ```
 //! # use std::str::FromStr;
 //! # use bitcoin::*;
-//! # use bdk::wallet::{self, coin_selection::*, spendable_collection::*};
+//! # use bdk::wallet::{self, coin_selection::*, multi_tracker::*};
 //! # use bdk::*;
 //! # const TXIN_BASE_WEIGHT: usize = (32 + 4 + 4 + 1) * 4;
 //! #[derive(Debug)]
 //! struct AlwaysSpendEverything;
 //!
 //! impl CoinSelectionAlgorithm for AlwaysSpendEverything {
-//!     fn coin_select<D: SpendableCollection>(
+//!     fn coin_select<D: MultiTracker>(
 //!         &self,
 //!         database: &D,
 //!         required_utxos: Vec<WeightedUtxo>,
@@ -99,7 +99,7 @@ use rand::{rngs::StdRng, SeedableRng};
 use std::collections::HashMap;
 use std::convert::TryInto;
 
-use super::spendable_collection::SpendableCollection;
+use super::multi_tracker::MultiTracker;
 
 /// Default coin selection algorithm used by [`TxBuilder`](super::tx_builder::TxBuilder) if not
 /// overridden
@@ -158,7 +158,7 @@ pub trait CoinSelectionAlgorithm: std::fmt::Debug {
     /// - `amount_needed`: the amount in satoshi to select
     /// - `fee_amount`: the amount of fees in satoshi already accumulated from adding outputs and
     ///                 the transaction's header
-    fn coin_select<Sc: SpendableCollection>(
+    fn coin_select<Sc: MultiTracker>(
         &self,
         database: &Sc,
         required_utxos: Vec<WeightedUtxo>,
@@ -177,7 +177,7 @@ pub trait CoinSelectionAlgorithm: std::fmt::Debug {
 pub struct LargestFirstCoinSelection;
 
 impl CoinSelectionAlgorithm for LargestFirstCoinSelection {
-    fn coin_select<D: SpendableCollection>(
+    fn coin_select<D: MultiTracker>(
         &self,
         _database: &D,
         required_utxos: Vec<WeightedUtxo>,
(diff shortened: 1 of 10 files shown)


src/wallet/tx_builder.rs
```diff
@@ -117,23 +123,24 @@ impl TxBuilderContext for BumpFee {}
...
-pub(crate) struct TxParams {
+pub(crate) struct TxParams<'a> {
     pub(crate) recipients: Vec<(Script, u64)>,
     pub(crate) drain_wallet: bool,
     pub(crate) drain_to: Option<Script>,
     pub(crate) fee_policy: Option<FeePolicy>,
-    pub(crate) internal_policy_path: Option<BTreeMap<String, Vec<usize>>>,
-    pub(crate) external_policy_path: Option<BTreeMap<String, Vec<usize>>>,
+    pub(crate) descriptor_policy_paths:
+        BTreeMap<&'a ExtendedDescriptor, BTreeMap<String, Vec<usize>>>,
```
afilini, line 143: Maybe we can avoid the lifetime by using something else as a key? like the descriptor's checksum?

src/wallet/multi_tracker.rs
```diff
@@ -0,0 +1,405 @@
+///! MultiTracker
```
afilini, line 12: This will eventually need the usual header that we have in all the other files

src/wallet/multi_tracker.rs
```diff
@@ -0,0 +1,405 @@
...
+    ) -> Result<Box<dyn Iterator<Item = Arc<dyn TransactionSigner>> + '_>, Error>;
+
+    /// Obtains owned UTXOs alongside their satisfaction weights.
+    ///
+    /// TODO @evanlinjin: We should have our own structure for Unspents.
+    /// Fields that make sense =>
+    /// * Confirmed height.
+    /// * Descriptor + Child index.
+    /// * Satisfaction weight of scriptPubKey (weight of corresponding scriptSig/witness data).
+    fn iter_unspent(&self) -> Result<Box<dyn Iterator<Item = (LocalUtxo, usize)> + '_>, Error>;
```
afilini, line 91: We actually have a `WeightedUtxo` struct that is basically a utxo + the weight, maybe that's more convenient here

src/wallet/multi_tracker.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// Note that if the script is not owned, None shoud be returned.
+    fn get_path_of_script_pubkey(
+        &self,
+        script: &Script,
+    ) -> Result<Option<(DescriptorItem, u32)>, Error> {
+        let descriptors = self.iter_descriptors()?.collect::<Vec<_>>();
+
+        let secp = SecpCtx::new();
+
+        for index in 0..2100_u32 {
```
afilini: Wait, that's a lot! Deriving all the descriptors up to index 2100 may take a very long time. Do you think it makes sense to add an extra argument for a "max index"? Most of the time I guess the caller would only need to derive up to 100/200

src/wallet/multi_tracker.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// Output may not be owned by us, just be part of a transaction that we are keeping track of.
+    fn get_output(&self, outpoint: &OutPoint) -> Result<Option<TxOut>, Error> {
+        self.get_tx(&outpoint.txid)?.map_or_else(
+            || Err(Error::InvalidOutpoint(*outpoint)),
+            |tx_item| {
+                Ok(tx_item
+                    .raw
+                    .output
+                    .get(outpoint.vout as usize)
+                    .map(Clone::clone))
```
afilini: Should we return an error if a tx doesn't have the requested vout? I kinda feel like these are swapped around, we should return `None` if we don't know about the outpoint, but if we know the tx and the vout is missing, then it's clearly invalid

src/wallet/multi_tracker.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// Obtains a new change address without a descriptor provided explicitly
+    #[deprecated]
+    fn new_auto_address(&self) -> Result<AddressInfo, Error> {
+        let (mut internal, mut external): (Vec<_>, Vec<_>) = self
+            .iter_descriptors()?
+            .partition(|item| item.keychain == KeychainKind::Internal);
+
+        internal.append(&mut external);
+
+        let item = internal.iter().chain(external.iter()).next().unwrap();
```
afilini: I think here you are "appending" twice. First directly to the vector, and then when iterating you chain with external.

`append` probably empties the `external` vector, so in the end the code works fine but it looks weird.

src/wallet/multi_tracker.rs
```diff
@@ -0,0 +1,405 @@
...
+            .db
+            .borrow()
+            .iter_utxos()?
+            .into_iter()
+            .filter(|utxo| !utxo.is_spent)
+            .filter_map(move |utxo| {
+                let (item, _) = self
+                    // @evanlinjin: Will panic with default implementation on timeout.
+                    .get_path_of_script_pubkey(&utxo.txout.script_pubkey)
+                    .unwrap()?;
```
afilini, line 310: Oh is that a question mark on an option? I thought they were still unstable, it's awesome we can finally use them 😍

src/wallet/multi_tracker.rs
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    fn get_path_of_script_pubkey(
+        &self,
+        script: &Script,
+    ) -> Result<Option<(DescriptorItem, u32)>, Error> {
+        // TODO: Add as struct field
+        let desc_map = self
+            .iter_descriptors()?
+            .map(|desc| (desc.descriptor.to_string(), desc))
```
afilini, line 369: I think here again you can just use the checksum which is shorter

src/wallet/multi_tracker.rs
```diff
@@ -0,0 +1,405 @@
...
+        let item = internal.iter().chain(external.iter()).next().unwrap();
+
+        #[allow(deprecated)]
+        self.new_address(&item.descriptor)
+    }
+}
+
+/// Implements [`MultiTracker`] with one external descriptor and an optional internal descriptor.
+// #[derive(Clone)]
+pub struct LegacyTracker<'a, D> {
```
afilini, line 185: Is there a specific reason why you chose to not let this struct own the descriptors and the database and instead use references?

afilini: Is it to make this easier to clone for example? I see that in the `TxBuilder` you have to clone this, in that case it makes sense, but then the question is: why can't you store a reference in the `TxBuilder` like we used to do before with the `Wallet`?

src/wallet/coin_selection.rs
```diff
@@ -144,7 +145,7 @@ impl CoinSelectionResult {
 /// selection algorithm when it creates transactions.
 ///
 /// For an example see [this module](crate::wallet::coin_selection)'s documentation.
-pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
+pub trait CoinSelectionAlgorithm: std::fmt::Debug {
     /// Perform the coin selection
     ///
     /// - `database`: a reference to the wallet's database that can be used to lookup additional
```
afilini: The documentation will have to be updated eventually, just writing this as a reminder

src/wallet/coin_selection.rs
```diff
@@ -144,7 +145,7 @@ impl CoinSelectionResult {
 /// selection algorithm when it creates transactions.
 ///
 /// For an example see [this module](crate::wallet::coin_selection)'s documentation.
-pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
+pub trait CoinSelectionAlgorithm: std::fmt::Debug {
```
afilini, line 148: I'm trying to figure out how these changes affect my original idea for the coin selection: the reason why the `D` generic was defined here was that we wanted to offer a way to use custom databases in a custom coin selection without being restricted on the API in any way.

For example, let's say I have my custom database that exposes some kind of metric that I'd like to use for the coin selection, then I can implement `CoinSelectionAlgorithm<MyCustomDb>` and this would make the cs get a `&MyCustomDb` (not a generic `&dyn Database`), which would allow me to use any custom API I have.

It's not the end of the world to lose this ability because again it was something that probably nobody used, but I think it was kind of cool and I'm wondering if we can somehow preserve this even with the `MultiTracker`.

src/wallet/coin_selection.rs
```diff
@@ -649,21 +675,52 @@ mod test {
         ]
     }
 
+    // creates a tx with given utxo as single output.
+    // updates utxo to contain correct txid in outpoint.
+    fn tx(utxo: &mut WeightedUtxo, lock_time: u32) -> (Transaction, Txid) {
+        let local = match &mut utxo.utxo {
+            Utxo::Local(local) => local,
+            _ => panic!(),
```
afilini: We may want to at least provide a good error message here

afilini (commented):
I still feel like I'm trying to wrap my head around this, but conceptually I think it makes sense and I really like how this allows us to slowly move away from `Wallet` by making it "backward compatible" with `MultiTracker`.

src/wallet/coin_selection.rs
evanlinjin: But since CoinSelectionAlgorithm is a trait, we can have a struct implement CoinSelectionAlgorithm (with special databases and parameters as fields) so we can still do those custom algorithms without the need for the generic type. Let me know if I've missed something though

src/wallet/tx_builder.rs
evanlinjin: I will change to use checksum once https://github.com/bitcoindevkit/bdk/pull/671 is approved and merged :)

src/wallet/multi_tracker.rs
evanlinjin: However, `WeightedUtxo` can be of a foreign `Utxo` so it is not as "direct" as using an union here (since this method is about iterating unspent, and owned UTXOs). Should I keep it as is?

src/wallet/multi_tracker.rs
evanlinjin: This is the default implementation of the method and I intended it just to be there for convenience. But you are right, so I've changed it to check for 200 iterations, and added a comment saying that this method should be overloaded. Would that be enough? It seems inelegant to add another input imo.

src/wallet/multi_tracker.rs
evanlinjin: Resolved in 7649de62edc4610107c852185e87780a05a92ee6

src/wallet/multi_tracker.rs
evanlinjin: Addressed in 400dbd19d9ae4c094da74213cb22ec1eef8a2e69

src/wallet/multi_tracker.rs
evanlinjin: Will wait on https://github.com/bitcoindevkit/bdk/pull/671 :)

src/wallet/multi_tracker.rs
evanlinjin: @afilini It is just to avoid copying data in memory. The assumption is that one would only use `LegacyTracker` with the existence of `Wallet` (which is a good assumption imo). Since, the whole point of `LegacyTracker` is to get `Wallet` to "implement" `MultiTracker`, since transaction-building and signing logic is to be wrapped around `MultiTracker`. Right now this tx logic uses static functions of `Wallet` (which is confusing), but will avoid annoying git as much as possible. Those static functions, responsible for tx logic, should eventually be moved out of `Wallet` when the time is right.

src/wallet/coin_selection.rs
evanlinjin: Addressed in 2a302d2c7ef344331a23ee094b4bc52b106297c7

src/wallet/coin_selection.rs
evanlinjin: Addressed in 7432b07ad36942de3884d35906748d79c7e76183

evanlinjin:
This work will be replaced with BDK Core/Chain.

PR #648, opened 2022-07-01
test: BDK won't add unconf inputs when fee bumping 

### LINKED ISSUE #144: `add_utxo` with unconfirmed output will be used in `bump_fee` in violation of RBF rules
I am fairly sure this is the case. It needs a test in any case.

BIP125 requires this: https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki#implementation-details

danielabrozzoni:
### Description

Closes #144 

### Notes to reviewers

#144 is describing a bug that doesn't seem to happen in BDK master anymore (BDK not respecting BIP125 rule 2). This PR just adds a test to check that the bug is fixed.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -899,8 +899,6 @@ where
     /// # Ok::<(), bdk::Error>(())
     /// ```
     // TODO: support for merging multiple transactions while bumping the fees
-    // TODO: option to force addition of an extra output? seems bad for privacy to update the
-    // change
     pub fn build_fee_bump(
         &self,
         txid: Txid,
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -2262,7 +2262,6 @@ pub(crate) mod test {
             .drain_to(drain_addr.script_pubkey())
             .drain_wallet();
         let (psbt, details) = builder.finish().unwrap();
-        dbg!(&psbt);
         let outputs = psbt.unsigned_tx.output;
 
         assert_eq!(outputs.len(), 2);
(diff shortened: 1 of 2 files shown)


src/wallet/mod.rs
```diff
@@ -3836,6 +3833,52 @@ pub(crate) mod test {
...
+        let addr = Address::from_str("2N1Ffz3WaNzbeLFBb51xyFMHYSEUXcbiSoX").unwrap();
+        let mut builder = wallet.build_tx();
+        builder
+            .drain_wallet()
+            .drain_to(addr.script_pubkey())
+            .enable_rbf();
+        let (psbt, mut original_details) = builder.finish().unwrap();
+        // Now we receive one transaction with 0 confirmations. We won't be able to use that for
+        // fee bumping, as it's still unconfirmed!
+        crate::populate_test_db!(
```
evanlinjin, line 3880: Would it make sense to test replacing txs with different amounts of inputs (where inputs are of various states of unconfirmed/confirmed)?

Not sure if it should be included for the scope of this PR though...

evanlinjin:
@notmandatory It seems this PR is just test additions, should the **`bug`** label be removed?

LLFourn:
Isn't this intentional? [BIP125](https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki)

 > The replacement transaction may only include an unconfirmed input if that input was included in one of the original transactions. (An unconfirmed input spends an output from a currently-unconfirmed transaction.)

evanlinjin:
> Isn't this intentional? [BIP125](https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki)
> 
> > The replacement transaction may only include an unconfirmed input if that input was included in one of the original transactions. (An unconfirmed input spends an output from a currently-unconfirmed transaction.)

This ticket is wrongly labelled as `bug`, this is just a some additional tests from my understanding

@danielabrozzoni means that we shouldn't allow introducing new unconfirmed inputs when RBF.

danielabrozzoni:
Yeah, I removed the "bug" label and added the "test" one. #144 was describing a bug that has been fixed since, but here I'm adding a test to check that the bug doesn't appear anymore. I'll update the review notes to make it clear!

src/wallet/mod.rs
danielabrozzoni: I'm not sure what the "end goal" of such test would be. What are you testing exactly? That we would choose only confirmed inputs for fee bumping?

evanlinjin: For example, we can test whether BDK allows for RBF of tx with unconfirmed parent.
```
Given:
* Wallet with 2 UTXOs (1 confirmed, 1 unconfirmed).
* We attempt to "drain wallet", resulting in a tx with 2 inputs 
  (1 unconfirmed, 1 confirmed) and 1 output to drain address.

When: 
* Attempt RBF of "drain tx".

Expect:
* Successful RBF of "drain tx".
```

evanlinjin: @danielabrozzoni checking whether we only pick confirmed new inputs would be great too!

danielabrozzoni: Right, I'll add it, thanks!

danielabrozzoni: > For example, we can test whether BDK allows for RBF of tx with unconfirmed parent. [...]

I've added `test_bump_fee_unconfirmed_input` testing this scenario

> checking whether we only pick confirmed new inputs would be great too!

Mh, I think this is already tested (at least partially) in `test_bump_fee_unconfirmed_inputs_only`, and I can't seem to find a better way to test it... Maybe I could send a lot of unconfirmed inputs and one unconfirmed output to the wallet, and check that BDK will only pick the confirmed input for fee bumping. What do you think? Is it a test worth adding in your opinion?

afilini (approved):
ACK 5d00f8238886a993ef21056e5b3e216a4aae6951

src/wallet/mod.rs
evanlinjin: @danielabrozzoni Fair, let's skip it :P I like the new test!

PR #652, opened 2022-07-03
Fix: Hang when `ElectrumBlockchainConfig::stop_gap == 0`

### LINKED ISSUE #651: Setting electrum client's stop_gap as 0 results in hang
**Describe the bug**  

Setting `ElectrumBlockchainConfig::stop_gap` as `0` results in infinite loop during wallet setup/sync.
https://github.com/bitcoindevkit/bdk/blob/063d51fd7587815c733f24365a4de02fed072903/src/blockchain/electrum.rs#L302

**To Reproduce**  
* Create wallet with `ElectrumBlockchainConfig::stop_gap` set as `0`
* Call `Wallet::sync`

**Expected behavior**  
* Electrum blockchain should still be functional with `stop_gap == 0`

**Build environment**  
 - BDK tag/commit: `v0.19.0` 063d51fd7587815c733f24365a4de02fed072903
 - OS+version: Fedora Linux 5.18.5-200.fc36.x86_64
 - Rust/Cargo version: cargo 1.62.0 (a748cf5a3 2022-06-08)
 - Rust/Cargo target: x86_64-unknown-linux-gnu

**Additional context**

Example code:
```rust
let mut test_client = TestClient::default();

let factory = Arc::new(
    ElectrumBlockchain::from_config(&ElectrumBlockchainConfig {
        url: test_client.electrsd.electrum_url.clone(),
        socks5: None,
        retry: 0,
        timeout: None,
        stop_gap: 0, // STOP_GAP IS ZERO
    })
    .unwrap(),
);

let wallet = Wallet::new(
    "wpkh(L5EZftvrYaSudiozVRzTqLcHLNDoVn7H5HSfM9BAN6tMJX8oTWz6)",
    None,
    bitcoin::Network::Regtest,
    MemoryDatabase::new(),
)
.unwrap();

// HANGS HERE INDEFINITELY!!!
factory
    .sync_wallet(&wallet, None, Default::default())
    .unwrap();
```

evanlinjin:
* Ensure `chunk_size` is > 0 during wallet sync.

* Slight refactoring for better readability.

* Add test: `test_electrum_blockchain_factory_sync_with_stop_gaps`

### Description

`Wallet::sync` hangs indefinitely when syncing with Electrum with `stop_gap` set as 0.

The culprit is having `chunk_size` set as `stop_gap`. A zero value results in syncing not being able to progress.

Fixes #651

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/blockchain/electrum.rs
+++ b/src/blockchain/electrum.rs
@@ -147,19 +147,12 @@ impl WalletSync for ElectrumBlockchain {
 
                 Request::Conftime(conftime_req) => {
                     // collect up to chunk_size heights to fetch from electrum
-                    let needs_block_height = {
-                        let mut needs_block_height = HashSet::with_capacity(chunk_size);
-                        conftime_req
-                            .request()
-                            .filter_map(|txid| txid_to_height.get(txid).cloned())
-                            .filter(|height| block_times.get(height).is_none())
-                            .take(chunk_size)
-                            .for_each(|height| {
-                                needs_block_height.insert(height);
-                            });
-
-                        needs_block_height
-                    };
+                    let needs_block_height = conftime_req
+                        .request()
+                        .filter_map(|txid| txid_to_height.get(txid).cloned())
+                        .filter(|height| block_times.get(height).is_none())
+                        .take(chunk_size)
+                        .collect::<HashSet<u32>>();
 
                     let new_block_headers = self
                         .client
@@ -329,6 +322,7 @@ mod test {
     use super::*;
     use crate::database::MemoryDatabase;
     use crate::testutils::blockchain_tests::TestClient;
+    use crate::testutils::configurable_blockchain_tests::ConfigurableBlockchainTester;
     use crate::wallet::{AddressIndex, Wallet};
 
     crate::bdk_blockchain_tests! {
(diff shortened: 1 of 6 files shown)


evanlinjin:
I realized for the test, I'm using a single key descriptor, when I really should be using an extended key, since we are testing the stop gap.

So this is still WIP.

src/blockchain/electrum.rs
```diff
@@ -385,4 +386,116 @@ mod test {
 
         assert_eq!(wallet.get_balance().unwrap(), 50_000);
     }
+
+    #[test]
+    fn test_electrum_blockchain_factory_sync_with_stop_gaps() {
```
LLFourn: can we put this in `blockchain_tests` so it covers all blockchains?

evanlinjin: That's a good idea, I'll do so in a bit.

LLFourn:
Interesting bug. I'm not sure about the semantics of `stop_gap = 0`? Maybe the correct thing is to do nothing. If I am going to stop after I find a gap of 0 then I should just stop right away no? (you start off with a gap of 0).

evanlinjin:
> Interesting bug. I'm not sure about the semantics of `stop_gap = 0`? Maybe the correct thing is to do nothing. If I am going to stop after I find a gap of 0 then I should just stop right away no? (you start off with a gap of 0).

From my understanding, the gap between 0 and 1 is zero, as there is no gap. These are integers, so there are no numbers between 0 and 1.

src/blockchain/electrum.rs
```diff
@@ -108,7 +108,10 @@ impl WalletSync for ElectrumBlockchain {
         let mut block_times = HashMap::<u32, u32>::new();
         let mut txid_to_height = HashMap::<Txid, u32>::new();
         let mut tx_cache = TxCache::new(database, &self.client);
-        let chunk_size = self.stop_gap;
+
+        // Ensure chunk_size > 0, otherwise we will result in an infinite loop
```
LLFourn: I understand better what is going on now. We must find a gap greater than `stop_gap` in order to stop.
So I think this comment should be: 

> set the chunk_size to the smallest value that we can find a gap greater than stop gap.

It is also a performance issue to set it to `stop_gap` since you will do two requests when you should only do one on any empty wallet (for example).

LLFourn:
ACK. See https://github.com/bitcoindevkit/bdk/pull/652#discussion_r912656091 

`chunk_size` was set wrongly before. Thanks!

src/blockchain/electrum.rs
evanlinjin: Changed in 475b005eb9151284885600896e8f0f50a45cd1b2.

src/blockchain/electrum.rs
evanlinjin: > can we put this in `blockchain_tests` so it covers all blockchains?

I feel like this is too much work for this PR. Should we get this merged and I can create a second ticket?

Each test case requires creating a new `Blockchain` with a specified `stop_gap`.

I am also not very familiar with macros in Rust, so this may take a while.

LLFourn: Hmmmm. Ok I see and agree with this point. Not all blockchains have "stop gaps" anyways.

evanlinjin: We can definitely do something universal I think. Can create a custom test trait with a method such as `generate_blockchain_with_stop_gap` and have that trait as an input.

src/blockchain/electrum.rs
```diff
@@ -145,18 +148,16 @@ impl WalletSync for ElectrumBlockchain {
                 Request::Conftime(conftime_req) => {
                     // collect up to chunk_size heights to fetch from electrum
                     let needs_block_height = {
-                        let mut needs_block_height_iter = conftime_req
+                        let mut needs_block_height = HashSet::with_capacity(chunk_size);
```
afilini: I think here you can just construct the `HashSet` from the iterator, using the `FromIterator` trait:

```rust
let needs_block_height = conftime_req./*<long chain>*/.collect();
```

src/blockchain/electrum.rs
```diff
@@ -385,4 +386,116 @@ mod test {
...
+        // Amount (in satoshis) provided to a single address (which expects to have a balance)
+        const AMOUNT_PER_TX: u64 = 1000;
+
+        // [stop_gap, actual_gap, addrs_before, addrs_after]
+        //
+        // [0]     stop_gap: Passed to [`ElectrumBlockchainConfig`]
+        // [1]   actual_gap: Range size of address indexes without a balance
+        // [2] addrs_before: Range size of address indexes (before gap) which contains a balance
+        // [3]  addrs_after: Range size of address indexes (after gap) which contains a balance
+        let test_vectors: Vec<[u64; 4]> = vec![
```
afilini: Wouldn't it be better to use `usize` here? I see that you have to cast these values in a few places

afilini (commented):
Concept ACK, a couple of things:

- did you check if this happens with esplora as well? The two blockchains are very similar in terms of architecture and code
- once you are done fixing the various comments, could you squash the commits into the first one?

evanlinjin:
> Concept ACK, a couple of things:
> 
> - did you check if this happens with esplora as well? The two blockchains are very similar in terms of architecture and code
> - once you are done fixing the various comments, could you squash the commits into the first one?

Thank you for the review! I'll test all Blockchains with a stop_gap and do some rebasing after ☺️

evanlinjin:
> did you check if this happens with esplora as well? The two blockchains are very similar in terms of architecture and code

@afilini I've made a trait that can be reused for testing both esplora and electrum. I've also made it expandable, so in the future we can implement more tests in it. Let me know whether these changes make sense!

src/blockchain/electrum.rs
evanlinjin: I've done something relatively universal, we can continue this conversation here: https://github.com/bitcoindevkit/bdk/pull/652#issuecomment-1173789385

src/blockchain/electrum.rs
evanlinjin: It seems that if we do this, other variables will need casting... Should we leave it as is?

evanlinjin:
> once you are done fixing the various comments, could you squash the commits into the first one?

@afilini I feel like it's better to keep it to two commits. One if for the actual fix, the other is introducing a new testing trait that can be shared between different blockchain implementations.

What do you think?

evanlinjin:
@afilini Should this fix also be part of [Release 0.20.0 Feature Freeze](https://github.com/bitcoindevkit/bdk/milestone/16)?

Ensuring that the `Blockchain` implementations respect the `stop_gap` parameter can potentially be important for avoiding address reuse.

The test I have included checks for this.

https://github.com/bitcoindevkit/bdk/blob/612da165f8cfbc2b0aee0a37b2cdc44d6da017c8/src/testutils/configurable_blockchain_tests.rs#L131-L135

src/blockchain/electrum.rs
afilini: Yeah at a quick glance I thought using usize would be fine, but if you need to cast anyway there's no reason to change it

afilini:
> @afilini I feel like it's better to keep it to two commits. One if for the actual fix, the other is introducing a new testing trait that can be shared between different blockchain implementations.

Yeah two commits works for me, as long as they are self-contained (they should both compile and run `cargo test` with no errors) and they have a good commit message

> @afilini Should this fix also be part of [Release 0.20.0 Feature Freeze](https://github.com/bitcoindevkit/bdk/milestone/16)?

I don't think this is as critical as that other one, but if this is ready in time yes, this is also important

afilini (approved):
ACK 612da165f8cfbc2b0aee0a37b2cdc44d6da017c8

PR #653, opened 2022-07-03
Fix: Wallet sync may decrement address index

### LINKED ISSUE #649: Syncing causes get_address index to decrement
This only happens with certain xpubs. Code:
```
    #[test]
    fn test_address_looping() {
        let wallet =
            Wallet::new("wpkh([5d14cd2a/84h/1h/0h]tpubDCWhawC5a8Rgx6y7rk5qHtueax2MVWfdfobzEcmcSvQUDYq94dnqyx6KAFbxCocxQnnLuFcRYFWmvXS9DtWRYqJeU33pcvsam9AaozJXS1P/0/*)",
                        None,
                        Network::Testnet,
                        MemoryDatabase::new()).unwrap();

        for _ in 0..15 {
            let address = wallet.get_address(AddressIndex::New).unwrap();
            println!("{}:  {}", address.index, address.address.to_string());
        }

        println!("wallet.sync()");
        let client = get_electrum_client(-1, "ssl://electrum.blockstream.info:60002").unwrap();
        let blockchain = ElectrumBlockchain::from(client);

        wallet.sync(&blockchain, SyncOptions { progress: None }).unwrap();

        for _ in 0..15 {
            let address = wallet.get_address(AddressIndex::New).unwrap();
            println!("{}:  {}", address.index, address.address.to_string());
        }
    }
```
Output:
```
0:  tb1qghhpvphu6rd6ygwuurw3p9cg7se84taj2vssdu
1:  tb1q79crajpg7y838vmqngerx6eqv5tuytzlk7cx3n
2:  tb1qcae3r5r6pv7p59s9vy3depphz3vffuqhrxm0e3
3:  tb1qemh2gefeeydc8k649tqlhz4egr3pktnt72h9xz
4:  tb1qckls5f7xrz29k3jtuw67ul6dp6mlz39m6r0xmj
5:  tb1qvycmqxyffvllmt7u8732nr7l576s9al6qlfleq
6:  tb1qrxm4qpdjevpzjmh7x7048kng6ys0audzz50jwn
7:  tb1q285u2994gsyuplqajm0ttadh238pp2clsutxc3
8:  tb1q5cfe3nrwtl4qfxyavnagxe7lqchskdjvkeqw50
9:  tb1qpa8375ve0um755svsm7u8r73ltk5xc0hwmp240
10:  tb1qc3nd440fqzx3zzwvm2n47g5dn2ahpcs74e58rw
11:  tb1qur22cvhcwtez5d7f0jeajg6ul22chqvcqwv3pg
12:  tb1qyud6sa4dseyx5k8vzp9yhq2nef9p5zrcv7pjny
13:  tb1qq50glc8hh20k8fh39u9xedshy6plra69cx4xxt
14:  tb1q92llgymvnl90y2a6ku2fztfw4heg400psph740
wallet.sync()
3:  tb1qemh2gefeeydc8k649tqlhz4egr3pktnt72h9xz
4:  tb1qckls5f7xrz29k3jtuw67ul6dp6mlz39m6r0xmj
5:  tb1qvycmqxyffvllmt7u8732nr7l576s9al6qlfleq
6:  tb1qrxm4qpdjevpzjmh7x7048kng6ys0audzz50jwn
7:  tb1q285u2994gsyuplqajm0ttadh238pp2clsutxc3
8:  tb1q5cfe3nrwtl4qfxyavnagxe7lqchskdjvkeqw50
9:  tb1qpa8375ve0um755svsm7u8r73ltk5xc0hwmp240
10:  tb1qc3nd440fqzx3zzwvm2n47g5dn2ahpcs74e58rw
11:  tb1qur22cvhcwtez5d7f0jeajg6ul22chqvcqwv3pg
12:  tb1qyud6sa4dseyx5k8vzp9yhq2nef9p5zrcv7pjny
13:  tb1qq50glc8hh20k8fh39u9xedshy6plra69cx4xxt
14:  tb1q92llgymvnl90y2a6ku2fztfw4heg400psph740
15:  tb1qzyzavz6qc9qn52hxrfhpddztas84r69ytqzgyh
16:  tb1qgfwvg3tpvkedzl5exmmwjlpzx74rtmkw0dtemg
17:  tb1qh6vsnhff77tr9g0yg5jum0n3h5ccesnhldfsjw
```
One would expect `get_address` subsequent to `sync` to return an address at index 15.

**Build environment**  
 - BDK tag/commit: a328607d27feb5d556df99ec6beabc41cc655160
 - OS+version: Ubuntu 22.04
 - Rust/Cargo version: 1.54.0
 - Rust/Cargo target: x86_64

evanlinjin:
### Description

Fixes #649 

It is critical to ensure `Wallet::get_address` with `AddressIndex::new` always returns a new and unused address.

This bug seems to be Electrum-specific. The fix is to check address index updates to  ensure that newly suggested indexes are not smaller than indexes already in database.

### Notes to the reviewers

I have written new tests in `/testutils/blockchain_tests.rs` that tests all `Blockchain` implementations.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/blockchain/script_sync.rs
+++ b/src/blockchain/script_sync.rs
@@ -314,6 +314,22 @@ impl<'a, D: BatchDatabase> State<'a, D> {
         let finished_txs = make_txs_consistent(&self.finished_txs);
         let observed_txids: HashSet<Txid> = finished_txs.iter().map(|tx| tx.txid).collect();
         let txids_to_delete = existing_txids.difference(&observed_txids);
+
+        // Ensure `last_active_index` does not decrement database's current state.
+        let index_updates = self
+            .last_active_index
+            .iter()
+            .map(|(keychain, sync_index)| {
+                let sync_index = *sync_index as u32;
+                let index_res = match self.db.get_last_index(*keychain) {
+                    Ok(Some(db_index)) => Ok(std::cmp::max(db_index, sync_index)),
+                    Ok(None) => Ok(sync_index),
+                    Err(err) => Err(err),
+                };
+                index_res.map(|index| (*keychain, index))
+            })
+            .collect::<Result<Vec<(KeychainKind, u32)>, _>>()?;
+
         let mut batch = self.db.begin_batch();
 
         // Delete old txs that no longer exist
@@ -377,8 +393,10 @@ impl<'a, D: BatchDatabase> State<'a, D> {
             batch.set_tx(finished_tx)?;
         }
 
-        for (keychain, last_active_index) in self.last_active_index {
-            batch.set_last_index(keychain, last_active_index as u32)?;
+        // apply index updates
+        for (keychain, new_index) in index_updates {
+            debug!("updating index ({}, {})", keychain.as_byte(), new_index);
+            batch.set_last_index(keychain, new_index)?;
         }
 
         info!(
--- a/src/testutils/blockchain_tests.rs
+++ b/src/testutils/blockchain_tests.rs
@@ -372,6 +372,7 @@ macro_rules! bdk_blockchain_tests {
             use $crate::blockchain::Blockchain;
             use $crate::database::MemoryDatabase;
             use $crate::types::KeychainKind;
+            use $crate::wallet::AddressIndex;
             use $crate::{Wallet, FeeRate, SyncOptions};
             use $crate::testutils;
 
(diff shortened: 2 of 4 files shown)


evanlinjin:
### TODO
- [ ] I should add a test for possible regressions. Ensure that wallet index increments when expected after sync.
- [ ] Fix CI errors

evanlinjin:
These tests should be made universal so we can test all `Blockchain` implementations. I'm on it now.

evanlinjin:
Update: this bug affects `RpcBlockchain` as well. Probably in a different manner though.

src/blockchain/script_sync.rs
```diff
@@ -314,6 +314,26 @@ impl<'a, D: BatchDatabase> State<'a, D> {
...
+
+        // Ensure `last_active_index` does not decrement database's current state.
+        let index_updates = self
+            .last_active_index
+            .iter()
+            .filter_map(|(keychain, index)| {
+                let db_index = self
+                    .db
+                    .get_last_index(*keychain)
+                    .unwrap_or(None)
```
afilini: I think here ideally we should propagate the error instead of "downgrading" it to a `None` (otherwise again we may overwrite the last derivation index accidentally).

On the other end, I'm not sure what would be the best way to do it: the only thing I could think of would be to replace this `filter_map` with just a map, and return a `Result` in the closure so that you can collect a `Result<Vec<_>>` and propagate the error with `?`. Then you'll have to filter out the `None`s before iterating the list.

evanlinjin: This is my current solution, would this be satisfactory?

https://github.com/bitcoindevkit/bdk/blob/edf5ab7ea72189fc978271fee362159c02415c48/src/blockchain/script_sync.rs#L318-L344

src/testutils/blockchain_tests.rs
```diff
@@ -651,6 +652,61 @@ macro_rules! bdk_blockchain_tests {
...
+                let (wallet, blockchain, _descriptors, mut test_client) = init_single_sig();
+
+                const ADDRS_TO_FUND: u32 = 7;
+                const ADDRS_TO_IGNORE: u32 = 11;
+
+                let mut first_addr_index: u32 = 0;
+
+                (0..ADDRS_TO_FUND + ADDRS_TO_IGNORE).for_each(|i| {
+                    let new_addr = wallet.get_address(AddressIndex::New).unwrap();
+                    println!("new_addr.index: {}", new_addr.index);
```
notmandatory: I think you forgot to remove this `println!`.

evanlinjin: Removed!

src/blockchain/script_sync.rs
```diff
@@ -314,6 +314,35 @@ impl<'a, D: BatchDatabase> State<'a, D> {
...
+                match db_result {
+                    // Only update on OK db_result with db index < proposed index
+                    Ok(db_index) if db_index < index => Some((*keychain, index)),
+                    _ => None,
+                }
+            })
+            .collect::<Vec<_>>();
+
+        // report error on db fail
+        db_result?;
```
rajarshimaitra: We can just match the whole thing in one iterator and we don't have to initialize and pass around the errors..
This comparison logic can be written something like this.. 
```
        let index_updates = self
            .last_active_index
            .iter()
            .map(|(keychain, sync_index)| {
                let db_index = match self.db.get_last_index(*keychain).transpose() {
                    Some(Ok(db_index)) => Ok(std::cmp::max(db_index as usize, *sync_index)),
                    Some(e) => e.map(|i| i as usize),
                    None => Ok(*sync_index),
                };
                db_index.and_then(|index| Ok((*keychain, index as usize)))
            })
            .collect::<Result<Vec<(KeychainKind, usize)>, _>>()?;
```
To me this looks much simple to read..

rajarshimaitra (changes requested):
Concept ACK..

I verified that the test is only failing for Electrum, but not for RPC or Esplora.. Maybe worth exploring further.. 

I have a code suggestion for the same checking logic.. It seems we can get the check done in much sorter codes.

src/blockchain/script_sync.rs
evanlinjin: @rajarshimaitra Thank you for providing an elegant solution. I'll add a shoutout in the commit (almost feel bad for just copying it).

evanlinjin: I did not know you could collect into a `Result<>`, that is pretty useful.

evanlinjin: I don't think we need to `transpose` with your solution:
```rust
        let index_updates = self
            .last_active_index
            .iter()
            .map(|(keychain, sync_index)| {
                let sync_index = *sync_index as u32;
                let index_res = match self.db.get_last_index(*keychain) {
                    Ok(Some(db_index)) => Ok(std::cmp::max(db_index, sync_index)),
                    Ok(None) => Ok(sync_index),
                    Err(err) => Err(err),
                };
                index_res.map(|index| (*keychain, index))
            })
            .collect::<Result<Vec<(KeychainKind, u32)>, _>>()?;
```

src/blockchain/script_sync.rs
evanlinjin: Continued here: https://github.com/bitcoindevkit/bdk/pull/653#discussion_r916074702

src/blockchain/script_sync.rs
rajarshimaitra: Yeah converting `Vec<Result<>>` into `Result<Vec<>>` is very useful and allows us to do a lot just with iterators.. 

Yes agree, this looks much better than the original..

rajarshimaitra (approved):
tACK ac5f0af2f7465910c2e1fdd41f83a31126a91de7

Thanks for the quick update..

afilini:
ACK af6bde3997ac2fe4cfdbb691f2af2e802f0ab7c6

I'll wait for the CI to finish and then I'll merge this.

PR #654, opened 2022-07-05
Multi descriptor: Introduce `DatabaseFactory` trait

evanlinjin:
### Description

This is part of bitcoindevkit/bdk_wallet#188 to add multi-descriptor wallet support to BDK.

As each `DescriptorTracker` should have it's own database, and a `MultiTracker` structure will handle an aggregate of `DescriptorTracker`s, we need a factory structure that the `MultiTracker` can use to create `Database`s for each `DescriptorTracker`.

#### All Submissions:

#### New Features:

### DIFF
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -27,10 +27,12 @@
 use serde::{Deserialize, Serialize};
 
 use bitcoin::hash_types::Txid;
-use bitcoin::{OutPoint, Script, Transaction, TxOut};
+use bitcoin::{Network, OutPoint, Script, Transaction, TxOut};
 
+use crate::descriptor::ExtendedDescriptor;
 use crate::error::Error;
 use crate::types::*;
+use crate::wallet::utils::SecpCtx;
 
 pub mod any;
 pub use any::{AnyDatabase, AnyDatabaseConfig};
@@ -212,12 +214,28 @@ pub(crate) trait DatabaseUtils: Database {
 
 impl<T: Database> DatabaseUtils for T {}
 
+/// A factory trait that builds databases which share underlying configurations and/or storage
+/// paths.
+pub trait DatabaseFactory: Sized {
+    /// Inner type to build
+    type Inner: BatchDatabase;
+
+    /// Builds the defined [`DatabaseFactory::Inner`] type.
+    fn build(
+        &self,
+        descriptor: &ExtendedDescriptor,
+        network: Network,
+        secp: &SecpCtx,
+    ) -> Result<Self::Inner, Error>;
+}
+
 #[cfg(test)]
 pub mod test {
     use std::str::FromStr;
 
     use bitcoin::consensus::encode::deserialize;
     use bitcoin::hashes::hex::*;
+    use bitcoin::util::bip32::{self, DerivationPath, ExtendedPubKey};
     use bitcoin::*;
 
     use super::*;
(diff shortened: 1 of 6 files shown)


src/database/sqlite.rs
```diff
@@ -970,11 +975,49 @@ pub fn migrate(conn: &Connection) -> rusqlite::Result<()> {
     Ok(())
 }
 
+/// A [`DatabaseFactory`] implementation that builds [`SqliteDatabase`].
+///
+/// Each built database is stored in path of format: `<path_root>_<hash>.<path_ext>`
+/// Where `hash` contains identifying data derived with inputs provided by
+/// [`DatabaseFactory::build`] or [`DatabaseFactory::build_with_change`] calls.
+pub struct SqliteDatabaseFactory<P> {
```
afilini, line 983: I had an idea for this, let me know what you think: in theory it should be possible to just have a single database by adding an extra column to all the relevant tables that identifies to which descriptor the rows belong.

So my idea for the factory would be something that wraps the db (probably in an `Arc`) and all the actual `SqliteDatabase` struct would have a member inside to keep track of what id it belongs to. If you create a `SqliteDatabase` directly this could either be `None` or just an empty string.

Then we just have to add an additional `WHERE` clause to the queries and everything should work magically.

Why we may want to do this? Mainly because I guess having a single file makes it easier to work with, backup, etc. And also this opens the possibility to add more methods in the future that iterate over the whole db, which would probably bring a performance benefit to `MultiTracker`.

Disadvantages? I guess the extra `WHERE` clause might cause a small performance hit? But we can also add an index on that column if we feel like it.

src/database/mod.rs
```diff
@@ -212,12 +214,39 @@ pub(crate) trait DatabaseUtils: Database {
...
+
+    /// Builds the defined [`DatabaseFactory::Inner`] type with the addition of a change descriptor.
+    fn build_with_change(
+        &self,
+        descriptor: ExtendedDescriptor,
+        change_descriptor: Option<ExtendedDescriptor>,
+        network: Network,
+        secp: &SecpCtx,
+    ) -> Result<Self::Inner, Error>;
+}
```
afilini: I wonder if it makes sense to just drop this since eventually I guess we will reach the point where there's a 1-to-1 relationship between descriptor and database. Considering this will be used mainly with `MultiTracker` I don't think we are gonna need the `build_with_change` too much, maybe just for `LegacyTracker` but we can work around it I think.

src/database/mod.rs
```diff
@@ -212,12 +214,39 @@ pub(crate) trait DatabaseUtils: Database {
...
+/// A factory trait that builds databases which share underlying configurations and/or storage
+/// paths.
+pub trait DatabaseFactory: Sized {
+    /// Inner type to build
+    type Inner: BatchDatabase;
+
+    /// Builds the defined [`DatabaseFactory::Inner`] type.
+    fn build(
+        &self,
+        descriptor: ExtendedDescriptor,
```
afilini: I would change this to just take a reference, to avoid cloning the descriptor needlessly

afilini (commented):
Code looks good, there are just a couple of things I would have done differently. But I'm not sure my ideas are better, I'm just proposing here so we can discuss them

src/database/mod.rs
evanlinjin: We won't need it for LegacyTracker as it only requires one database ☺️

afilini: Yeah but that database should take into account both the external and internal descriptor for its name/key or whatever right?

But again, I don't think that's a strong enough reason to keep this, we will figure out something else

afilini: Ah yes you are right, we don't need to construct that database from a factory anyway

src/database/sqlite.rs
evanlinjin: I think this is a great idea! I would assume that performance wouldn't be too much of an issue since we are already using indexes for almost all tables/queries.

However, I feel like this should be in it's separate issue/PR since we need to change the "nature" of `SqliteDatabase`? What do you think?

notmandatory: I think @afilini's suggestion can be done in this PR if you want to give it a shot. Once that's in this should be ready for general review right?

evanlinjin: @notmandatory Thank you for including this PR in the last call (which I failed to attend :cry: ). I have not touched this PR since the last review because I am skeptical on whether this is actually needed.

My current thinking is that `Database` itself is NOT an ideal model for implementing multi-descriptor, and the correct approach seems to be that of which `bdk_core` is doing. I think more discussion is needed before I would be willing to continue/close this PR.

notmandatory: No problem, I think it's fine to leave this one on hold and see how a better approach shapes up with `bdk_core`.

PR #662, opened 2022-07-10
Consolidate `fee_amount` and `amount_needed`

### LINKED ISSUE #641: Consolidate `amount_needed` and `fee_amount` in one parameter
Currently, the `coin_select` method from `CoinSelectionAlgorithm` has
[seven parameters](https://github.com/bitcoindevkit/bdk/blob/2283444f727b85663b19d100af86db4392ebc891/src/wallet/coin_selection.rs#L160-L168).
One of them, `amount_needed` (the sum of the outgoing vout values) is only used
to compound the amount that the selected utxos must satisfy to create the
transaction.
Another parameter involved in that compounded amount is
`fee_amount`. It has two use cases. The first one, to [pass the accumulated fees](https://github.com/bitcoindevkit/bdk/blob/2283444f727b85663b19d100af86db4392ebc891/src/wallet/coin_selection.rs#L158-L159)
for each txout and transaction’s header. The other is to [accumulate](https://github.com/bitcoindevkit/bdk/blob/2283444f727b85663b19d100af86db4392ebc891/src/wallet/coin_selection.rs#L275-L296)
the resulting fees from the current selection (in `LargestFirstCoinSelection`
and `OldestFirstCoinSelection`). 

Both parameters can be consolidated in one parameter like:
```rust
/// `target_amount`: the outgoing amount in satoshis and the fees already
///                  accumulated from added outputs and transaction’s header.
```

I think that this’ll still be valid in the future, as coin selection algorithms
consider those values fixed at the moment of `coin_select`. Also, the algorithm
would return only the `fee_amount` of the tx inputs in vin, which would reduce
the carry of the state through functions, remove the need for [variable shadowing](https://github.com/bitcoindevkit/bdk/blob/2283444f727b85663b19d100af86db4392ebc891/src/wallet/mod.rs#L760)
and differentiate the vin fee amount from the rest of fees.

The following code illustrates some of the ideas above:

- New `coin_select` signature ([old](https://github.com/bitcoindevkit/bdk/blob/2283444f727b85663b19d100af86db4392ebc891/src/wallet/coin_selection.rs#L160-L168)):
```rust
    fn coin_select(
        &self,
        database: &D,
        required_utxos: Vec<WeightedUtxo>,
        fee_rate: FeeRate,
        target_amount: u64,
    ) -> Result<CoinSelectionResult, Error>;
```

- New `CoinSelectionResult` meaning ([old](https://github.com/bitcoindevkit/bdk/blob/2283444f727b85663b19d100af86db4392ebc891/src/wallet/coin_selection.rs#L115-L121)):
```rust
/// Result of a successful coin selection
#[derive(Debug)]
pub struct CoinSelectionResult {
    /// List of outputs selected for use as inputs
    pub selected: Vec<Utxo>,
    /// Total fee amount for the selected utxos
    pub fee_amount: u64,
}
```

- New `select_sorted_utxos` ([old](https://github.com/bitcoindevkit/bdk/blob/2283444f727b85663b19d100af86db4392ebc891/src/wallet/coin_selection.rs#L268-L310)):
```rust
fn select_sorted_utxos(
    utxos: impl Iterator<Item = (bool, WeightedUtxo)>,
    fee_rate: FeeRate,
    target_amount: u64,
) -> Result<CoinSelectionResult, Error> {
    let mut selected_amount = 0;
    let mut fee_amount = 0; // fee_amount is local
    let selected = utxos
        .scan(
            (&mut selected_amount, &mut fee_amount),
            |(selected_amount, fee_amount), (must_use, weighted_utxo)| {
            // target_amount doesn't change (fee_amount accumulates the selected utxos fees)
                if must_use || **selected_amount < target_amount + **fee_amount {
                    **fee_amount +=
                        fee_rate.fee_wu(TXIN_BASE_WEIGHT + weighted_utxo.satisfaction_weight);
                    **selected_amount += weighted_utxo.utxo.txout().value;
                    ...
                    Some(weighted_utxo.utxo)
                } else {
                    None
                }
            },
        )
        .collect::<Vec<_>>();
        ...
}
```

- Replace of `actual_target` in `coin_select` ([old](https://github.com/bitcoindevkit/bdk/blob/2283444f727b85663b19d100af86db4392ebc891/src/wallet/coin_selection.rs#L390)):
> ~~let actual\_target = fee\_amount + amount\_needed;~~
> actual\_target = **target_amount** 

- Change `single_random_draw` ([old](https://github.com/bitcoindevkit/bdk/blob/2283444f727b85663b19d100af86db4392ebc891/src/wallet/coin_selection.rs#L553-L560)) and `bnb` ([old](https://github.com/bitcoindevkit/bdk/blob/2283444f727b85663b19d100af86db4392ebc891/src/wallet/coin_selection.rs#L444-L453)) methods:
```rust
fn single_random_draw(
    &self,
    required_utxos: Vec<OutputGroup>,
    mut optional_utxos: Vec<OutputGroup>,
    curr_value: i64,
    target_amount: u64,
) -> CoinSelectionResult
```
```rust
fn bnb(
    &self,
    required_utxos: Vec<OutputGroup>,
    mut optional_utxos: Vec<OutputGroup>,
    mut curr_value: i64,
    mut curr_available_value: i64,
    target_amount: u64,
    cost_of_change: f32,
) -> Result<CoinSelectionResult, Error>
```

- Change `calculate_cs_result` ([old](https://github.com/bitcoindevkit/bdk/blob/2283444f727b85663b19d100af86db4392ebc891/src/wallet/coin_selection.rs#L585-L601)):
```rust
fn calculate_cs_result(
    mut selected_utxos: Vec<OutputGroup>,
    mut required_utxos: Vec<OutputGroup>,
) -> CoinSelectionResult {
    selected_utxos.append(&mut required_utxos);
    let fee_amount = selected_utxos.iter().map(|u| u.fee).sum::<u64>();
    let selected = selected_utxos
        .into_iter()
        .map(|u| u.weighted_utxo.utxo)
        .collect::<Vec<_>>();

    CoinSelectionResult {
        selected,
        fee_amount,
    }
}
```

- Sum vin fee amount to total fee amount instead of using "variable shadowing"
  ([old](https://github.com/bitcoindevkit/bdk/blob/2283444f727b85663b19d100af86db4392ebc891/src/wallet/mod.rs#L760)):
```rust
  fee_amount += coin_selection.fee_amount;
```

This implies some backward incompatible changes so I want to know if there is
support for this, and hear ideas/opinions before opening any PR.

csralvall:
### Description

Before this commit `fee_amount` and `amount_needed` were passed as independent
parameters. From the perspective of coin selection algorithms, they are always
used jointly for the same purpose, to create a coin selection with a total
effective value greater than it's summed values.

This commit removes the abstraction that the use of the two parameter
introduced by consolidating both into a single parameter, `target_amount`, who
carries their values added up.

Resolves: bitcoindevkit/bdk#641

### Notes to the reviewers

I just updated old tests and didn't create new ones because almost all changes
are renames and "logic changes" (like the addition of the selection fee) are
tested in the modified tests.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -41,8 +41,7 @@
 //!         required_utxos: Vec<WeightedUtxo>,
 //!         optional_utxos: Vec<WeightedUtxo>,
 //!         fee_rate: FeeRate,
-//!         amount_needed: u64,
-//!         fee_amount: u64,
+//!         target_amount: u64,
 //!         drain_script: &Script,
 //!     ) -> Result<CoinSelectionResult, bdk::Error> {
 //!         let mut selected_amount = 0;
@@ -60,7 +59,7 @@
 //!             )
 //!             .collect::<Vec<_>>();
 //!         let additional_fees = fee_rate.fee_wu(additional_weight);
-//!         let amount_needed_with_fees = (fee_amount + additional_fees) + amount_needed;
+//!         let amount_needed_with_fees = additional_fees + target_amount;
 //!         if selected_amount < amount_needed_with_fees {
 //!             return Err(bdk::Error::InsufficientFunds {
 //!                 needed: amount_needed_with_fees,
@@ -74,7 +73,7 @@
 //!
 //!         Ok(CoinSelectionResult {
 //!             selected: all_utxos_selected,
-//!             fee_amount: fee_amount + additional_fees,
+//!             fee_amount: additional_fees,
 //!             excess,
 //!         })
 //!     }
@@ -148,7 +147,7 @@ pub enum Excess {
 pub struct CoinSelectionResult {
     /// List of outputs selected for use as inputs
     pub selected: Vec<Utxo>,
-    /// Total fee amount in satoshi
+    /// Total fee amount for the selected utxos in satoshis
     pub fee_amount: u64,
     /// Remaining amount after deducing fees and outgoing outputs
     pub excess: Excess,
@@ -183,14 +182,13 @@ pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
     ///
     /// - `database`: a reference to the wallet's database that can be used to lookup additional
     ///               details for a specific UTXO
-    /// - `required_utxos`: the utxos that must be spent regardless of `amount_needed` with their
+    /// - `required_utxos`: the utxos that must be spent regardless of `target_amount` with their
     ///                     weight cost
-    /// - `optional_utxos`: the remaining available utxos to satisfy `amount_needed` with their
+    /// - `optional_utxos`: the remaining available utxos to satisfy `target_amount` with their
     ///                     weight cost
     /// - `fee_rate`: fee rate to use
-    /// - `amount_needed`: the amount in satoshi to select
-    /// - `fee_amount`: the amount of fees in satoshi already accumulated from adding outputs and
-    ///                 the transaction's header
+    /// - `target_amount`: the outgoing amount in satoshis and the fees already
+    ///                    accumulated from added outputs and transaction’s header.
     /// - `drain_script`: the script to use in case of change
     #[allow(clippy::too_many_arguments)]
     fn coin_select(
@@ -199,8 +197,7 @@ pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
         required_utxos: Vec<WeightedUtxo>,
         optional_utxos: Vec<WeightedUtxo>,
         fee_rate: FeeRate,
-        amount_needed: u64,
-        fee_amount: u64,
+        target_amount: u64,
         drain_script: &Script,
     ) -> Result<CoinSelectionResult, Error>;
 }
@@ -219,14 +216,12 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
         required_utxos: Vec<WeightedUtxo>,
         mut optional_utxos: Vec<WeightedUtxo>,
         fee_rate: FeeRate,
-        amount_needed: u64,
-        fee_amount: u64,
+        target_amount: u64,
         drain_script: &Script,
     ) -> Result<CoinSelectionResult, Error> {
         log::debug!(
-            "amount_needed = `{}`, fee_amount = `{}`, fee_rate = `{:?}`",
-            amount_needed,
-            fee_amount,
+            "target_amount = `{}`, fee_rate = `{:?}`",
+            target_amount,
             fee_rate
         );
 
@@ -240,7 +235,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for LargestFirstCoinSelection {
                 .chain(optional_utxos.into_iter().rev().map(|utxo| (false, utxo)))
         };
 
-        select_sorted_utxos(utxos, fee_rate, amount_needed, fee_amount, drain_script)
+        select_sorted_utxos(utxos, fee_rate, target_amount, drain_script)
     }
 }
 
@@ -258,8 +253,7 @@ impl<D: Database> CoinSelectionAlgorithm<D> for OldestFirstCoinSelection {
         required_utxos: Vec<WeightedUtxo>,
         mut optional_utxos: Vec<WeightedUtxo>,
         fee_rate: FeeRate,
-        amount_needed: u64,
-        fee_amount: u64,
+        target_amount: u64,
         drain_script: &Script,
     ) -> Result<CoinSelectionResult, Error> {
         // query db and create a blockheight lookup table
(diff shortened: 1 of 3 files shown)


src/wallet/coin_selection.rs
```diff
@@ -1148,23 +1129,23 @@ mod test {
...
+                target_amount,
             )
             .unwrap();
 
         assert_eq!(result.selected.len(), 1);
         assert_eq!(result.selected_amount(), 100_000);
         let input_size = (TXIN_BASE_WEIGHT + P2WPKH_WITNESS_SIZE).vbytes();
-        let epsilon = 0.5;
-        assert!((1.0 - (result.fee_amount as f32 / input_size as f32)).abs() < epsilon);
+        // the final fee rate should be exactly the same that the fee rate given
```
danielabrozzoni: nit: the same "as"

Also, is there a specific reason for changing this line? I think we should be more precise with the feerate and always check that it's >= than what the user asked, but maybe we shouldn't change it in this PR?

danielabrozzoni (commented):
Concept ACK, a small comment :)

src/wallet/coin_selection.rs
csralvall: I thought that as we were using the first utxo's effective value as target amount an from this PR on we don't have to consider the accumulated fee amount of the overall transaction, it was a good way to illustrate the idea that BnB would produce an exact match. I don't have strong opinions on this, so I can roll it back.

afilini: Maybe it's just me but I always try to avoid comparing floats with strict equality because there's inherent rounding happening and I'm always afraid it will cause something to break sooner or later.

Maybe in this case it won't happen, but just to be safe I would keep an epsilon, even if very small.

csralvall: Do you have any suggestion for its value? I think it should be less than 0.5 because we are not considering the overall `fee_amount` of the transaction as part of the `fee_rate` for the selection. But I don't know how to avoid magical numbers and use something more "analytical".

evanlinjin: It does seem that this test should be for "an exact match" whilst taking FP rounding errors into consideration. I also agree that `0.5` is probably too big ([machine epsilon for `binary32` is like 24 decimal places](https://en.wikipedia.org/wiki/Machine_epsilon) which roughly means, based on my limited understanding, numbers being stored with 23-24 digits of precision).

I would say, having `epsilon` as `0.01` should be plenty lol (no point going any smaller imo).

evanlinjin:
Concept ACK for me.

This PR is rather straight forward, and for the context on my work with multi-descriptor support (#486), it would be good to merge this PR soon (since I will also be modifying the `wallet/{mod|coin_selection}.rs` files).

Other than the concerns mentioned in https://github.com/bitcoindevkit/bdk/pull/662#discussion_r921181198, I think this PR is good to merge.

src/wallet/coin_selection.rs
csralvall: LGTM, done!

danielabrozzoni:
Please rebase to fix the conflicts, then I'll ACK :)

src/wallet/coin_selection.rs
csralvall: I found there is a constant defined for `f32` machine epsilon, probably for this kind of cases. I'll use it instead of `0.01`.

danielabrozzoni:
Code review ACK 6493bbf6db30c944c95321e359759126e8fa1350

csralvall:
Rebased with master.
- I've added an extra line to the `CHANGELOG.md` about the new meaning of
`fee_amount` in `CoinSelectionResult` which wasn't explicitly described before.
- I've removed the `Resolves: ...` part of the commit as I've realized it creates
a lot of automatic logs in related PRs and Issues, each time we add new changes,
duplicating info and making issue navigation cumbersome.
At the end, it's the maintainer who will put that line in the merge commit.
Maybe we can use some format to avoid the trigger and at the same time suggest
the fixed issues to the maintainer.
If you think it's not a serious issue, I can continue writing commits like
before.

src/wallet/coin_selection.rs
```diff
@@ -1505,12 +1480,11 @@ mod test {
             utxos,
             0,
             target_amount as i64,
-            FEE_AMOUNT,
```
danielabrozzoni, line 1508: In all the other tests you do `target_amount = *whatever it was before* + FEE_AMOUNT", maybe you should do the same here?

csralvall: Sure! Although, now that I review the test again, it's impossible for it to fail :eyes:.

danielabrozzoni: Yeah, you can't do much with unit testing on the SDR, other than hoping it doesn't panic

danielabrozzoni (approved):
tACK 6eb299b72f3126ae3d1e234b2a8039041273528f

danielabrozzoni:
I rebased by myself, now it's ready for merge

re-ACK e8df3d2d91927edb9a339c664f0603c47622e4b0 - I tested with the fuzzer, run it for 13,000,000 iterations, couldn't find any crash :)

Issue #660, opened 2022-07-08
create_tx returns wrong fee in TransactionDetails


icota:
Code:
```
    #[test]
    fn test_create_psbt() {
        let wallet =
            Wallet::new("wpkh([96f9192b/84h/0h/0h]tpubDCxzhZZE31g2EqSv1UajMAw5Hd62htydz9r2XBkrccHgBh8uw3n62zr6Zjmj64tfTk8Tjxo6VctjUMAh5DXWTErfQPC6RmQhTdtNnXuTXTQ/0/*)",
                        Some("wpkh([96f9192b/84h/0h/0h]tpubDCxzhZZE31g2EqSv1UajMAw5Hd62htydz9r2XBkrccHgBh8uw3n62zr6Zjmj64tfTk8Tjxo6VctjUMAh5DXWTErfQPC6RmQhTdtNnXuTXTQ/1/*)"),
                        Network::Testnet,
                        MemoryDatabase::new()).unwrap();

        let client = get_electrum_client(-1, "ssl://electrum.blockstream.info:60002").unwrap();
        let blockchain = ElectrumBlockchain::from(client);
        wallet.sync(&blockchain, SyncOptions { progress: None }).unwrap();

        let send_to = Address::from_str("tb1ql7w62elx9ucw4pj5lgw4l028hmuw80sndtntxt").unwrap();

        let mut builder = wallet.build_tx();
        builder
            .add_recipient(send_to.script_pubkey(), 8630)
            .enable_rbf()
            .fee_rate(FeeRate::from_sat_per_vb(2.01));

        let (psbt, details) = builder.finish().unwrap();

        let encoded = base64::encode(&serialize(&psbt));

        println!("{:#?}", details);
        println!("Serialised PSBT: {}", encoded);
    }
```
Output:
```
TransactionDetails {
    transaction: None,
    txid: 5b26b2f1d02a3e4071b59eafb9f850be7d0473932c89fab1eb87431786e97454,
    received: 0,
    sent: 8859,
    fee: Some(
        286,
    ),
    confirmation_time: None,
}

[...]

PR #666, opened 2022-07-12
Various fixes to the `fee_amount` calculation in `create_tx`

### LINKED ISSUE #160: How to calculate TXIN_BASE_WEIGHT?
PR #159 adds a `TXIN_BASE_WEIGHT` constant, which represents the `base weight of a Txin, not counting the weight needed for satisfaying it`. At the moment this constant is calculated as:
```
// prev_txid (32 bytes) + prev_vout (4 bytes) + sequence (4 bytes) + script_len (1 bytes)
pub const TXIN_BASE_WEIGHT: usize = (32 + 4 + 4 + 1) * 4;
```

My doubt is: should the `script_len` be included in the calculation? Miniscript docs say that the satisfaction weight `Includes the weight of the VarInts encoding the scriptSig and witness stack length.` (see https://docs.rs/miniscript/3.0.0/miniscript/descriptor/enum.Descriptor.html#method.max_satisfaction_weight)

Removing the `script_len` (i.e. defining `TXIN_BASE_WEIGHT` as `(32 + 4 + 4) * 4`) causes electrs tests to fail (`min relay fee not met`, see .https://github.com/bitcoindevkit/bdk/runs/1336260689?check_suite_focus=true).

I'm trying to understand if there's something I'm not seeing, or if there's some bug in bdk, or if I should ask directly in miniscript repo :)

danielabrozzoni:
### Description

This PR mainly fixes two bugs:
1. TXIN_BASE_WEIGHT wrongly included the `script_len` (Fixes #160)
2. We wouldn't take into account the segwit header in the fee calculation, which could have resulted in a transaction with a lower feerate than the requested one
3. In tests we used to push 108 bytes on the witness as a fake signature, but we should have pushed 106 instead

I also add a test to reproduce the conditions of #660, to check if it's solved. Turns out it's been solved already in #630, but if you're curious about what the bug was, here it is: https://github.com/bitcoindevkit/bdk/issues/660#issuecomment-1196436776
#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1871,6 +1871,14 @@ pub(crate) mod test {
     use crate::signer::{SignOptions, SignerError};
     use crate::wallet::AddressIndex::{LastUnused, New, Peek, Reset};
 
+    // The satisfaction size of a P2WPKH is 112 WU =
+    // 1 (elements in witness) + 1 (OP_PUSH) + 33 (pk) + 1 (OP_PUSH) + 72 (signature + sighash) + 1*4 (script len)
+    // On the witness itself, we have to push once for the pk (33WU) and once for signature + sighash (72WU), for
+    // a total of 105 WU.
+    // Here, we push just once for simplicity, so we have to add an extra byte for the missing
+    // OP_PUSH.
+    const P2WPKH_FAKE_WITNESS_SIZE: usize = 106;
+
     #[test]
     fn test_cache_addresses_fixed() {
         let db = MemoryDatabase::new();
@@ -2007,7 +2015,7 @@ pub(crate) mod test {
             $(
                 $( $add_signature )*
                 for txin in &mut tx.input {
-                    txin.witness.push([0x00; 108]); // fake signature
+                    txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
                 }
             )*
 
@@ -3236,7 +3244,7 @@ pub(crate) mod test {
         let txid = tx.txid();
         // skip saving the new utxos, we know they can't be used anyways
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -3296,7 +3304,7 @@ pub(crate) mod test {
         let txid = tx.txid();
         // skip saving the new utxos, we know they can't be used anyways
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -3362,7 +3370,7 @@ pub(crate) mod test {
         let mut tx = psbt.extract_tx();
         let txid = tx.txid();
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -3406,7 +3414,7 @@ pub(crate) mod test {
         let mut tx = psbt.extract_tx();
         let txid = tx.txid();
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -3462,7 +3470,7 @@ pub(crate) mod test {
         let mut tx = psbt.extract_tx();
         let txid = tx.txid();
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -3519,7 +3527,7 @@ pub(crate) mod test {
         let mut tx = psbt.extract_tx();
         let txid = tx.txid();
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -3560,7 +3568,7 @@ pub(crate) mod test {
         let txid = tx.txid();
         // skip saving the new utxos, we know they can't be used anyways
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -3623,7 +3631,7 @@ pub(crate) mod test {
         let txid = tx.txid();
         // skip saving the new utxos, we know they can't be used anyways
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -3694,7 +3702,7 @@ pub(crate) mod test {
         let txid = tx.txid();
         // skip saving the new utxos, we know they can't be used anyways
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -3765,7 +3773,7 @@ pub(crate) mod test {
         let txid = tx.txid();
         // skip saving the new utxos, we know they can't be used anyways
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -3826,7 +3834,7 @@ pub(crate) mod test {
         let txid = tx.txid();
         // skip saving the new utxos, we know they can't be used anyways
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -3897,7 +3905,7 @@ pub(crate) mod test {
         let txid = tx.txid();
         // skip saving the new utxos, we know they can't be used anyways
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -3976,7 +3984,7 @@ pub(crate) mod test {
         let mut tx = psbt.extract_tx();
         let txid = tx.txid();
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
@@ -4020,7 +4028,7 @@ pub(crate) mod test {
         let mut tx = psbt.extract_tx();
         let txid = tx.txid();
         for txin in &mut tx.input {
-            txin.witness.push([0x00; 108]); // fake signature
+            txin.witness.push([0x00; P2WPKH_FAKE_WITNESS_SIZE]); // fake signature
             wallet
                 .database
                 .borrow_mut()
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -736,8 +736,6 @@ where
         let mut outgoing: u64 = 0;
         let mut received: u64 = 0;
 
-        fee_amount += fee_rate.fee_wu(tx.weight());
-
         let recipients = params.recipients.iter().map(|(r, v)| (r, *v));
 
         for (index, (script_pubkey, value)) in recipients.enumerate() {
@@ -753,12 +751,13 @@ where
                 script_pubkey: script_pubkey.clone(),
                 value,
             };
-            fee_amount += fee_rate.fee_vb(serialize(&new_out).len());
 
             tx.output.push(new_out);
 
             outgoing += value;
         }
+
+        fee_amount += fee_rate.fee_wu(tx.weight());
 
         if params.change_policy != tx_builder::ChangeSpendPolicy::ChangeAllowed
             && self.change_descriptor.is_none()
@@ -851,6 +850,9 @@ where
                     script_pubkey: drain_script,
                 };
 
+                // TODO: We should pay attention when adding a new output: this might increase
+                // the lenght of the "number of vouts" parameter by 2 bytes, potentially making
+                // our feerate too low
                 tx.output.push(drain_output);
             }
         };
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -759,6 +759,17 @@ where
 
         fee_amount += fee_rate.fee_wu(tx.weight());
 
+        // Segwit transactions' header is 2WU larger than legacy txs' header,
+        // as they contain a witness marker (1WU) and a witness flag (1WU) (see BIP144).
+        // At this point we really don't know if the resulting transaction will be segwit
+        // or legacy, so we just add this 2WU to the fee_amount - overshooting the fee amount
+        // is better than undershooting it.
+        // If we pass a fee_amount that is slightly higher than the final fee_amount, we
+        // end up with a transaction with a slightly higher fee rate than the requested one.
+        // If, instead, we undershoot, we may end up with a feerate lower than the requested one
+        // - we might come up with non broadcastable txs!
+        fee_amount += fee_rate.fee_wu(2);
+
         if params.change_policy != tx_builder::ChangeSpendPolicy::ChangeAllowed
             && self.change_descriptor.is_none()
         {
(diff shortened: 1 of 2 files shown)


src/wallet/mod.rs
```diff
@@ -831,12 +847,18 @@ where
         }
 
         if drain_val.is_dust(&drain_output.script_pubkey) {
-            fee_amount += drain_val;
+            // We are not going to add a drain output. This means that both the value
+            // of the output and the fees needed for spending it will go in fees
+            fee_amount += drain_val + drain_fee;
```
LLFourn: Is this change the thing that fixes the bug that was found in #660?

[edit] I'm a little bit confused -- we are not adding the drain output but the comment on `drain_fee` says the fee to be paid *if* we add the drain output. This comment seems wrong then (iirc written by me -- sorry!).

Ok I think I see what we got wrong here. The above question can be disregarded.

src/wallet/mod.rs
```diff
@@ -3692,7 +3714,7 @@ pub(crate) mod test {
             .unwrap();
 
         let mut builder = wallet.build_fee_bump(txid).unwrap();
-        builder.fee_rate(FeeRate::from_sat_per_vb(141.0));
+        builder.fee_rate(FeeRate::from_sat_per_vb(142.0));
```
LLFourn: Ok so my understanding is the feerate is being changed here to keep round test values. Is there some analytic way we could calculate `142.0` here so it doesn't look like "just set it to whatever to make the test pass".

src/wallet/mod.rs
danielabrozzoni: > Is this change the thing that fixes the bug that was found in https://github.com/bitcoindevkit/bdk/issues/660?

Not really, I thought so, but the bug is still there!

You're right, this specific line of code doesn't seem to make much sense. I'm still trying to figure out what the bug could be - but I'm near! I'll put a comment when I'm ready, for now this PR is not really ready for review :)

src/wallet/mod.rs
```diff
@@ -750,6 +750,20 @@ where
...
+        // is better than undershooting it.
+        // If we pass a fee_amount that is slightly higher than the final fee_amount, we
+        // end up with a transaction with a slightly higher fee rate than the requested one.
+        // If, instead, we undershoot, we may end up with a feerate lower than the requested one
+        // - we might come up with non broadcastable txs!
+        fee_amount += fee_rate.fee_wu(2);
+        // FIXME: we should apply this same undershooting/overshooting heuristic to the drain output
+        // as well: when we add a drain_output that we didn't consider previously in the
+        // fee_amount, we're lowering the transaction fee rate.
+
```
csralvall: I don't understand the following:
As this is done before coin selection, the selected amount will be considering the extra fees derived of these 2 WUs as part of the amount to pay, so any excess (dropped as fees or used as change) will be produced based on the same extra fee.
How would the `drain_output` reduce the transaction fee rate?

src/wallet/mod.rs
```diff
@@ -805,9 +805,14 @@ where
             }
         };
 
-        fee_amount += fee_rate.fee_vb(serialize(&drain_output).len());
-
-        let drain_val = (coin_selection.selected_amount() - outgoing).saturating_sub(fee_amount);
+        // sum(inputs) - sum(outputs)
+        let delta = coin_selection.selected_amount() - outgoing;
+        // Fee to be paid if we add the drain_output
+        let drain_fee = fee_rate.fee_vb(serialize(&drain_output).len());
```
csralvall: I'm pretty sure that the old code was cumbersome, but correct, and this is just an equivalent way to write it.

Using `drain_fee` as `fee_rate.fee_vb(serialize(&drain_output).len())`, 
in the old code `drain_fee` was added to `fee_amount` (l808). So, `drain_val`
became the effective remaining value of the selection (l810).

If `drain_val` was dust, its value was added to `fee_amount` (l833), effectively computing
the following:
```
fee_amount = change_fee + drain_val + fee_amount
```
If it wasn't, the effective value `drain_val` was sent, as `change_fee`
was already considered inside `fee_amount`.

Note that:
```
delta = selected_amount - outgoing = drain_fee + drain_val + fee_amount
```

src/wallet/mod.rs
```diff
@@ -837,6 +836,9 @@ where
             if self.is_mine(&drain_output.script_pubkey)? {
                 received += drain_val;
             }
+            // TODO: We should pay attention when adding a new output: this might increase
+            // the lenght of the "number of vouts" parameter by 1 byte, potentially making
+            // our feerate too low
```
csralvall: Just to warn here that the addition of the draining output can increase the size of length field of the vout vector (and any VarInt in general), not by 1 but 2 or 4 bytes, depending on its previous length. More about this [here](https://learnmeabitcoin.com/technical/varint) and [here](https://docs.rs/bitcoin/latest/src/bitcoin/consensus/encode.rs.html#361-401).

evanlinjin: If we assume every output is a `P2WPKH`, an output will be 31 bytes [8(value) + 1(varint) + 22(scriptPubKey)]. So the weight of an output will be 124 (31 x 4). The max number of outputs we can fit in a block (assuming that we have nothing else, but just outputs) will be 32,258. This fits well within a u16 (or a 3 byte varint).

However, for transactions that are broadcasted, the max tx weight is 400,000, so we can only fit 3225 `P2WPKH` outputs.

Either way, I feel like a good assumption would be to say an increase of only 2 bytes is possible.

evanlinjin (commented):
Changes seem logical (based on my limited understanding), will try give a proper review later.

src/wallet/mod.rs
danielabrozzoni: I updated the PR (and this line changed), now it should be good :)

LLFourn: Thanks. The code makes sense now. At least more sense than the code before (that I think I wrote). If you could annotate somehow the change that fixes #660 or explain how it fixed it I'd appreciate it (I know this is still in draft -- take your time).

src/wallet/mod.rs
afilini: Didn't we say that it should be the coin selection to decide whether to add the drain output or not? I think we've been talking about this forever, the current logic of checking the remainder after the coin selection is done is suboptimal, so we should just give the coin selection the weight cost of the drain output and it will decide whether to add it or not.

Which should remove this problem completely if I understand it correctly

src/wallet/mod.rs
```diff
@@ -805,9 +805,14 @@ where
...
 
-        fee_amount += fee_rate.fee_vb(serialize(&drain_output).len());
-
-        let drain_val = (coin_selection.selected_amount() - outgoing).saturating_sub(fee_amount);
+        // sum(inputs) - sum(outputs)
+        let delta = coin_selection.selected_amount() - outgoing;
+        // Fee to be paid if we add the drain_output
+        let drain_fee = fee_rate.fee_vb(serialize(&drain_output).len());
+        // Value of the potential drain output
+        let drain_val = (coin_selection.selected_amount() - outgoing)
```
afilini: You've already computed `(coin_selection.selected_amount() - outgoing)`, so you can replace it with `delta`

src/wallet/mod.rs
afilini: Just for context: the value of the feerate here must be high enough that during rbf we are forced to add a new input and also that we have to remove the change that we had previously.

So in theory:

```
new_tx_weight = tx.weight() + extra_input_weight + input_satisfaction_weight - change_output_weight
fee_abs = 750000 - 45000 - epsilon (10, 20 sats?)
fee_rate = fee_abs / new_tx_weight
```

Note that the `tx.weight()` should be measured *after* we push the fake signatures, or alternatively an additional `input_satisfaction_weight` should be added to the `new_tx_weight` to account for that.

src/wallet/mod.rs
danielabrozzoni: > As this is done before coin selection, the selected amount will be considering the extra fees derived of these 2 WUs as part of the amount to pay, so any excess (dropped as fees or used as change) will be produced based on the same extra fee.
How would the drain_output reduce the transaction fee rate?

Yeah, you're right, at this point I'm not even sure anymore. I'm positive that sometimes we return a tx with a lower fee rate, but I can't figure out why.
Anywoo, I'll remove the comment as it's obviously wrong, and I'll keep working on this for another PR :)

>  I think we've been talking about this forever

Eh, review #630 then...

src/wallet/mod.rs
danielabrozzoni: Ouch, thanks!

src/wallet/mod.rs
csralvall: LGTM. I didn't think in all that, great explanation, thanks!

evanlinjin:
The changes in this PR seem to make sense.

However, to ensure that this PR actually fixes #660, imo it will be helpful to write a test that exactly replicates the conditions which caused the problems in #660.

I.e.

* Start with wallet with balance of 8859 sats.
* Create tx with one recipient, sending 8630 sats & fee rate 2.01 vb/sat.
* Ensure created tx's sent + fee <= starting_balance.

I'm assuming the problem pointed out in #660 is actually, what should be the correct behavior when the fee_rate suggested surpasses the available wallet balance.

evanlinjin:
Another idea would be to create psbts in a loop with different recipient amounts and ensure each psbt created does not surplus starting balance.

In that being said, this seems like a really annoying bug to iron out, lots of moving parts and things to think about.

src/wallet/mod.rs
danielabrozzoni: Done in 0639f62aeae3b4fb0a3d37ce60d7c8013d638881

src/wallet/mod.rs
danielabrozzoni: Oh, TIL! I wasn't sure how varint worked. I just updated the comment, thanks 🙏🏻

src/wallet/mod.rs
danielabrozzoni: Yeah, you're probably right. I think the main bug was just adding the drain_value to the fee_amount even when there was no drain output.

But I had to modify those lines anyways, so I just cleaned them up a bit ✨ 🧹

danielabrozzoni:
> However, to ensure that this PR actually fixes https://github.com/bitcoindevkit/bdk/issues/660, imo it will be helpful to write a test that exactly replicates the conditions which caused the problems in https://github.com/bitcoindevkit/bdk/issues/660.

Thanks, I've just added the test in da712c45806c663045d6c08f1bb0c99b1b84cd6c :)

> Another idea would be to create psbts in a loop with different recipient amounts and ensure each psbt created does not surplus starting balance.

I wrote it:
```diff
diff --git a/src/wallet/mod.rs b/src/wallet/mod.rs
index 10c8cab..2f65cb2 100644
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -3967,6 +3967,33 @@ pub(crate) mod test {
         builder.finish().unwrap();
     }
 
+    #[test]
+    fn test_fee_amount_many_cases() {
+        use rand::{SeedableRng, Rng, rngs::StdRng};
+        let seed = [0; 32];
+        let mut rng: StdRng = SeedableRng::from_seed(seed);
+        let (wallet, descriptors, _) = get_funded_wallet(get_test_wpkh());
+        let addr = Address::from_str("2N1Ffz3WaNzbeLFBb51xyFMHYSEUXcbiSoX").unwrap();
+        for _ in 0..100 {
+            let input_value = rng.gen_range(46_000, 100_000);
+            let incoming_txid = crate::populate_test_db!(
+                wallet.database.borrow_mut(),
+                testutils! (@tx ( (@external descriptors, 0) => input_value ) (@confirmations 1)),
+                Some(100),
+            );
+            let mut builder = wallet.build_tx();
+            builder
+                .add_recipient(addr.script_pubkey(), 45_000)
+                .add_utxo(OutPoint::new(incoming_txid, 0))
+                .unwrap()
+                .enable_rbf();
+            let (psbt, details) = builder.finish().unwrap();
+            let fee_amount = psbt.inputs.iter().fold(0, |acc, i| acc + i.witness_utxo.as_ref().unwrap().value) -
+                psbt.unsigned_tx.output.iter().fold(0, |acc, o| acc + o.value);
+            assert_eq!(fee_amount, details.fee.unwrap());
+        }
+    }
+
     #[test]
     fn test_sign_single_xprv() {
         let (wallet, _, _) = get_funded_wallet("wpkh(tprv8ZgxMBicQKsPd3EupYiPRhaMooHKUHJxNsTfYuScep13go8QFfHdtkG9nRkFGb7busX4isf6X9dURGCoKgitaApQ6MupRhZMcELAxTBRJgS/*)");
```

but then I couldn't find a way to make it fail on master (I tried changing the number of iterations and the `input_value` values), and we already have quite some tests checking the fee amount that wouldn't catch these bugs, so I figured it was useless.

This should be ready for review. I've also added a TL;DR of the various bug fixed in the PR description.

Sorry everyone if it took a while!

src/wallet/mod.rs
danielabrozzoni: I've added an explanation of the bug in the commit message (da712c45806c663045d6c08f1bb0c99b1b84cd6c), but the TL;DR is that we were using `saturating_sub` for calculating `drain_val`, which means that the `fee_amount` calculation would end up wrong when `drain_val` was supposed to be negative.

I fear that the commit message is not super explicating, so if you have ideas on how to improve it, please let me know!

rajarshimaitra (commented):
Initial concept ACK.. I will make another deeper pass. For now I had few comment.. 

It took me sometime to wrap my head around whats going on around the fee thingy.. I *guess* this is solving the problem of #660 but I haven't tried it out yet. So far all the tests seems to be passing. But I don't think we are adequately testing the fee situations and edge cases. One that you have already mentioned in the TODO comment about adding new outputs.  

The only fee related tests happens in the wallet module are by this macro https://github.com/bitcoindevkit/bdk/blob/3644a452c16717c259c3d6c1b3aff7c9b636a800/src/wallet/mod.rs#L1983-L2010

Which is used to check weather the fee is above certain threshold or not.. But it never asserts weather the fee reported in `TransactionDetails` is correct for the transaction. In fact it calculates the feerate from the result detail and then performs the check. So if the rate is wrong in the result, its wrong in the tests too.. 

I think we need to assert that the fees reported in the create transaction process is the correct fee.

But this is just one part of the problem..

The fee calculation is quite dynamic and its happening all over the places in the `create_tx` function. So its hard to reason about it.. Maybe we can streamline it somehow?? 

Another (unrelated to this PR) doubt I had is why the returning fee rate is 0 here? https://github.com/bitcoindevkit/bdk/blob/2de264ea0f9e63251fe1f21670808461b86425bb/src/wallet/mod.rs#L708

So then if we use a `FeePolicy::FeeAmount` in policy then all the calculations you added after it is multiplied with a 0 fee rate?? 

Overall I also feel we should have some more dedicated fee related tests like the one you already added, to adequately handle these cases.

danielabrozzoni:
Some good news: #660 has been solved by #630! 😁 
This PR is still useful tho, I promise. I've added a test to make sure that #660 doesn't happen again

> But I don't think we are adequately testing the fee situations and edge cases. One that you have already mentioned in the TODO comment about adding new outputs.
The only fee related tests happens in the wallet module are by this macro 
Which is used to check weather the fee is above certain threshold or not.. But it never asserts weather the fee reported in TransactionDetails is correct for the transaction. In fact it calculates the feerate from the result detail and then performs the check. So if the rate is wrong in the result, its wrong in the tests too..

Right, I didn't notice! I've added the fee check in `assert_fee_rate`. Regarding the test I posted above, speaking with alekos we agreed that it's better to introduce fuzzing in bdk instead. PR coming soon (two weeks :tm:)

> The fee calculation is quite dynamic and its happening all over the places in the create_tx function. So its hard to reason about it.. Maybe we can streamline it somehow??

Yeah, you're right in saying that it's quite complicated and scattered around, but I don't know how we could improve it honestly :(

> So then if we use a FeePolicy::FeeAmount in policy then all the calculations you added after it is multiplied with a 0 fee rate??

Yep! That's a cool trick to make sure that the fee returned in the transaction is the same as the user asked - maybe not so clear tho.
All around the coin selection and the create_tx code we have a fee_rate variable, to calculate the fee for adding each new utxo etc. We're basically saying "this thing has weight x, and we are using fee rate y", and then from x and y we can calculate the fee in sats from adding it to the transaction.
So, when the user tells you to have a 200 sats fees,  you set the fee rate to zero. In this way, every time we're trying to add to the fee amount the weight of  something, the fee added will be 0, and at the end of the calculations we'll have a tx with the fee specified by the user.

rajarshimaitra (approved):
tACK 09e044516718b3639cb1059dea7296925d5944e9

Changes looks good to me.. :)

rajarshimaitra:
> Yep! That's a cool trick to make sure that the fee returned in the transaction is the same as the user asked - maybe not so clear tho.

Ahh that's really neat.. Yup I missed that.. Thanks for updating the test macro and the tests.. It seems we are covering #660 all ok now..  Very much liked the docs in the test, makes them reading much easier.. 

To me this looks good to go.. 

> Yeah, you're right in saying that it's quite complicated and scattered around, but I don't know how we could improve it honestly :(

Yeah me neither.. I would try to think over it and see if I can see some ways.. :)

afilini (approved):
ACK 09e044516718b3639cb1059dea7296925d5944e9

afilini:
Code looks good, I'll try running this with my fuzzer before merging just to ensure we haven't missed something obvious

afilini:
Ok we need some time to investigate this, because with this PR it looks like we sometimes end up with a feerate lower than we ask for. It's probably caused by the lowering of `TXIN_BASE_WEIGHT`, but we can't understand exactly how at the moment.

danielabrozzoni:
Everything seems to be fixed now, I just pushed three additional commits (please don't hate me ❤️ )

Turns out that a constant in tests was wrong. I also added another check to assert_fee_rate, and added a test for high-R signatures (see #695)

src/wallet/mod.rs
```diff
@@ -753,13 +751,25 @@ where
...
                 value,
             };
-            fee_amount += fee_rate.fee_vb(serialize(&new_out).len());
 
             tx.output.push(new_out);
 
             outgoing += value;
         }
 
+        fee_amount += fee_rate.fee_wu(tx.weight());
```
afilini, line 760: An interesting thing to note here: while rust-bitcoin serializes txs with no inputs as segwit txs (hence, including the marker and flag bytes), the `Transaction::weight()` method doesn't!

I opened an issue to point out the inconsistency: https://github.com/rust-bitcoin/rust-bitcoin/issues/1159

This is just to say that the current code that explicitly pushes the two extra weight units is actually correct, but this may change in the future if the `weight` method ends up being modified.

afilini (approved):
ACK 419dc248b667db05295cd4c68347c4ef51f51023

PR #668, opened 2022-07-13
Add testing on Compact Block filters in CI

vladimirfomene:
Setup testing on Compact block filters blockchain
backend.

### Description

While working on #634, I realized we don't have a setup for carrying out test on Compact Block Filters blockchain in CI. This PR is a work in progress solution to get that setup. 

### Notes to the reviewers
For some reason the tests are taking too long to run. 

compact-block-filter-ci

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/blockchain/compact_filters/mod.rs
+++ b/src/blockchain/compact_filters/mod.rs
@@ -580,3 +580,22 @@ impl From<crate::error::Error> for CompactFiltersError {
         CompactFiltersError::Global(Box::new(err))
     }
 }
+
+#[cfg(test)]
+#[cfg(feature = "test-compact-filters")]
+mod test {
+
+    use super::*;
+
+    crate::bdk_blockchain_tests! {
+        fn test_instance(test_client: &TestClient) -> CompactFiltersBlockchain {
+            let bitcoind_port = test_client.bitcoind.params.rpc_socket.port();
+            let num_threads = 2;
+            let mempool = Arc::new(Mempool::default());
+            let peers = (0..num_threads)
+            .map(|_| Peer::connect(format!("localhost:{}", bitcoind_port), Arc::clone(&mempool), Network::Regtest))
+            .collect::<Result<_, _>>().unwrap();
+            CompactFiltersBlockchain::new(peers, "./wallet-filters", Some(0)).unwrap()
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/blockchain/compact_filters/mod.rs
```diff
@@ -580,3 +580,22 @@ impl From<crate::error::Error> for CompactFiltersError {
...
+
+    crate::bdk_blockchain_tests! {
+        fn test_instance(test_client: &TestClient) -> CompactFiltersBlockchain {
+            let bitcoind_port = test_client.bitcoind.params.rpc_socket.port();
+            let num_threads = 2;
+            let mempool = Arc::new(Mempool::default());
+            let peers = (0..num_threads)
+            .map(|_| Peer::connect(format!("localhost:{}", bitcoind_port), Arc::clone(&mempool), Network::Regtest))
+            .collect::<Result<_, _>>().unwrap();
+            CompactFiltersBlockchain::new(peers, "./wallet-filters", Some(0)).unwrap()
```
rajarshimaitra, line 598: I haven't tested yet, but I don't think this logic will work.. 

 - You can't get compact filters from rpc port. You have to connect to the p2p port.
 - The test client doesn't serve compact filters by default. You need to set a extra bitcoind args parameter accordingly to turn that on the test client.. Which can be done under a feature gate..

rajarshimaitra (changes requested):
Overall concept ACK on the idea of testing compact filter syncing in regtest.. Thanks for opening this up..  We might need this after integrating nakamoto too.. See here https://github.com/bitcoindevkit/bdk/issues/680#issuecomment-1221540035

But it needs the following additions to work properly..

vladimirfomene:
@rajarshimaitra , I will definitely get back to this once we integrate with Nakamoto.

rajarshimaitra:
Upon further thinking.. I think this can go into bdk even without nakamoto.. From the PoV of a testing framework we will always have only one single backend.. And CBF operation with single peer will be same between current bdk's cbf module and nakamoto.. 

So it might make sense to get this done even before we get nakamoto and bring the bdk's cbf impl into testing.. In that case, was when we shift to nakamoto we will already have a something to test it against, at least single peer scenario.. 

On high level what needs to be done to get compact_filters to testing is
 - Configure `bitcoind` in `TestClient` as compact blockfilter serving.. This should happen under a feature flag..
 - Get the p2p port of the bitcoind as pass that into compact filter blockchain.. 
  
@vladimirfomene if you are open to work on this I would suggest to carry on, as you are already quite close to the final thing..

danielabrozzoni:

PR #669, opened 2022-07-15
Add `get_block_headers` and check timestamps 

### LINKED ISSUE #183: timestamp timelocks are not satisfied
It looks like the logic of `finalize_psbt` only checks if height based timelocks are satisfied:

```rust
                  match desc.satisfy(
                        &mut tmp_input,
                        (
                            PsbtInputSatisfier::new(psbt, n),
                            After::new(current_height, false),
                            Older::new(current_height, create_height, false),
                        ),
                    ) {
```

In addition, we should check satisfaction of timestamp based timelocks. I will fix this in `bdk_core` but I thought this was worth noting and fixing ahead of then.

### LINKED ISSUE #642: timestamp timelocks are not satisfied
It looks like the logic of `finalize_psbt` only checks if height based timelocks are satisfied:

```rust
                  match desc.satisfy(
                        &mut tmp_input,
                        (
                            PsbtInputSatisfier::new(psbt, n),
                            After::new(current_height, false),
                            Older::new(current_height, create_height, false),
                        ),
                    ) {
```

In addition timestamp we should check satisfaction of timestamp based timelocks. I will fix this in `bdk_core` but I thought this was worth noting and fixing ahead of then.

wszdexdrf:
### Description

This changes `GetBlockHash` to `GetBlockInfo`. A new function is added to the new trait (in addition to the old one `get_block_hash`) which is `get_block_headers`, which returns block headers for a given block height. It is implemented on all blockchain backends.

This also changes `After` and `Older` defined in `wallet/utils.rs` to accept and check against timestamps.

### Notes to the reviewers

This PR was meant to fix bitcoindevkit/bdk_wallet#183, but further testing revealed that (a) the checking functions in `wallet/utils.rs` aren't actually called, the implementations in `miniscript/satisfy.rs` are called, which already check timestamps; and (b) the checking functions in `wallet/utils.rs` have a slightly different purpose (checking if the transaction can be broadcasted) compared to miniscript implementations (checking if the transaction timelocks are satisfied). Hence I changed the `wallet/utils.rs` functions, without changing the current flow of code. 
Hence the issue is already fixed and this PR barely changes anything.

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `Excess` enum to handle remaining amount after coin selection.
 - Move change creation from `Wallet::create_tx` to `CoinSelectionAlgorithm::coin_select`.
 - Change the interface of `SqliteDatabase::new` to accept any type that implement AsRef<Path>
+- Change trait `GetBlockHash` to `GetBlockInfo`. A new function is added to the new trait `get_block_header` which expects a block height and returns corresponding block header. This is implemented on every blockchain backend.
 
 ## [v0.20.0] - [v0.19.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Move change creation from `Wallet::create_tx` to `CoinSelectionAlgorithm::coin_select`.
 - Change the interface of `SqliteDatabase::new` to accept any type that implement AsRef<Path>
 - Change trait `GetBlockHash` to `GetBlockInfo`. A new function is added to the new trait `get_block_header` which expects a block height and returns corresponding block header. This is implemented on every blockchain backend.
+- Amend the struct `After` and `Older` to accept and check against timestamps as well.
 
 ## [v0.20.0] - [v0.19.0]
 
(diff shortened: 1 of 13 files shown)


rajarshimaitra (commented):
Concept ACK on f2634fc2ccdac7d140c4630cd6287bae356c9ea2..

I will go through the 2nd commit more thoroughly, but I really think that should have been a separate PR.. 

So if you are still working on it, my initial thought is, why not just remove the `GetBlockHash` trait all together?? A hash from a header is just a `.block_hash()` call away on the Header.. So the Header is really what we want most of the time, for time, hash whatever.. 

So if you are changing the trait anyway, why not just simply have a `GetHeader` trait?

danielabrozzoni:
Sorry, but I'm a bit lost... You're saying that the functions in `wallet/utils.rs` aren't actually called, then why is this PR modifying them? Can't we just wipe them and use the miniscript ones?

wszdexdrf:
Maybe I should explain. The reason I changed the functions (which were not actually used) was because I was unsure of which way to solve the issue. On one hand, the miniscript's function solve the problem, but they don't actually solve it strictly correctly (using median time past etc.). Therefore I changed the functions (in hopes that maybe we could change the calls in order to call our implementation (somehow) which would check more thoroughly (using block times and what-not). Though I didn't change the calls, I changed the functions in order to show how exactly the code will look, before changing how it works. Hence the reason this was a draft PR.

PR #671, opened 2022-07-16
Introduce `get_checksum_bytes` method and improvements

evanlinjin:
### Description

`get_checksum_bytes()` returns a descriptor checksum as `[u8; 8]` instead of `String`, potentially improving performance and memory usage.

In addition to this, since descriptors only use characters that fit within a UTF-8 8-bit code unit ([US-ASCII](https://www.charset.org/charsets/us-ascii)), there is no need to use the `char` type (which is 4 bytes). This can also potentially bring in some performance and memory-usage benefits.

### Notes to the reviewers

This is useful because we will be using descriptor checksums for indexing operations in the near future (multi-descriptor wallets bitcoindevkit/bdk_wallet#188 ).

Refer to comments by @afilini :
* https://github.com/bitcoindevkit/bdk/pull/647#discussion_r921184366
* https://github.com/bitcoindevkit/bdk/pull/647#discussion_r921914696
* https://github.com/bitcoindevkit/bdk/pull/654#discussion_r921980876

#### All Submissions:

#### New Features:

### DIFF
--- a/src/descriptor/checksum.rs
+++ b/src/descriptor/checksum.rs
@@ -14,12 +14,10 @@
 //! This module contains a re-implementation of the function used by Bitcoin Core to calculate the
 //! checksum of a descriptor
 
-use std::iter::FromIterator;
-
 use crate::descriptor::DescriptorError;
 
-const INPUT_CHARSET: &str =  "0123456789()[],'/*abcdefgh@:$%{}IJKLMNOPQRSTUVWXYZ&+-.;<=>?!^_|~ijklmnopqrstuvwxyzABCDEFGH`#\"\\ ";
-const CHECKSUM_CHARSET: &str = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";
+const INPUT_CHARSET: &[u8] = b"0123456789()[],'/*abcdefgh@:$%{}IJKLMNOPQRSTUVWXYZ&+-.;<=>?!^_|~ijklmnopqrstuvwxyzABCDEFGH`#\"\\ ";
+const CHECKSUM_CHARSET: &[u8] = b"qpzry9x8gf2tvdw0s3jn54khce6mua7l";
 
 fn poly_mod(mut c: u64, val: u64) -> u64 {
     let c0 = c >> 35;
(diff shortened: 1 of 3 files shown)


src/descriptor/checksum.rs
```diff
@@ -43,15 +41,17 @@ fn poly_mod(mut c: u64, val: u64) -> u64 {
     c
 }
 
-/// Compute the checksum of a descriptor
-pub fn get_checksum(desc: &str) -> Result<String, DescriptorError> {
```
notmandatory, line 47: I don't know if anyone is using `get_checksum()` but since it's public our policy is to deprecate it for at least one release before completely removing it. We haven't been super strict about it, but we do need to be more consistent. I think you could make a simple function to call your new `get_checksum_bytes` and convert to a String.

evanlinjin: @notmandatory I haven't removed it at all! And yes, `get_checksum` calls `get_checksum_bytes` 😅 (line 79)

evanlinjin: Check line 79

vladimirfomene (commented):
Tested ACK

src/descriptor/checksum.rs
```diff
@@ -14,12 +14,10 @@
...
 //! checksum of a descriptor
 
-use std::iter::FromIterator;
-
 use crate::descriptor::DescriptorError;
 
-const INPUT_CHARSET: &str =  "0123456789()[],'/*abcdefgh@:$%{}IJKLMNOPQRSTUVWXYZ&+-.;<=>?!^_|~ijklmnopqrstuvwxyzABCDEFGH`#\"\\ ";
-const CHECKSUM_CHARSET: &str = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";
+const INPUT_CHARSET: &[u8] = "0123456789()[],'/*abcdefgh@:$%{}IJKLMNOPQRSTUVWXYZ&+-.;<=>?!^_|~ijklmnopqrstuvwxyzABCDEFGH`#\"\\ ".as_bytes();
+const CHECKSUM_CHARSET: &[u8] = "qpzry9x8gf2tvdw0s3jn54khce6mua7l".as_bytes();
```
afilini: On both those lines I think you can use byte string literals: https://doc.rust-lang.org/reference/tokens.html#byte-string-literals

It should make the code more portable as it works on older rust versions as well (`as_bytes()` is a const fn only since 1.39) and it will also cause a compile error if we accidentally put a non-ascii character in there: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=81fadd81957dbc1c56844a279a9d239f

afilini (commented):
Concept ACK, just a small nit

src/descriptor/checksum.rs
evanlinjin: @afilini I just updated this! Hopefully the tests pass.

evanlinjin:
@afilini @notmandatory This is ready to go imo. :smiley:

afilini (approved):
ACK 6db5b4a094263a07eeb82ba92b47754d25a8d0d3

PR #672, opened 2022-07-17
Fix wallet sync not finding coins of addresses which are not cached

### LINKED ISSUE #451: BDK should always scan addresses it has utxos for
At the moment BDK stops syncing after it finds `stop_gap` addresses that don't have any associated tx. But what if has a utxo for an address after stop gap in the database. Despite BDK not scanning for them it can still find these utxos because they might be on a transaction that bdk is already interested for other reasons.

I think another way of thinking about it is that bdk should always sync addresses it has given out regardless if they fall after `stop_gap` empty addresses. In fact, stop gap shouldn't be used except for when doing an initial sync. BDK should just look up to the highest index address in the database (i.e. the highest index of an address it has given out).

### LINKED ISSUE #521: Coins not discovered within initial sync might never be found

evanlinjin:
Fixes #521
Fixes #451

^ However, only for electrum-based `Blockchain` implementations. For RPC and Compact Block Filters, syncing works differently, and so are the bugs - I have created a separate ticket for this (#677).

### Description

Previously, electrum-based blockchain implementations only synced for `scriptPubKey`s that are already cached in `Database`.

This PR introduces a feedback mechanism, that uses `stop_gap` and the difference between "current index" and "last active index" to determine whether we need to cache more `scriptPubKeys`.

The `WalletSync::wallet_setup` trait now may return an `Error::MissingCachedScripts` error which contains the number of extra `scriptPubKey`s to cache, in order to satisfy `stop_gap` for the next call.

`Wallet::sync` now calls `WalletSync` in a loop, caching in-between subsequent calls (if needed).

#### Notes to reviewers

1. The caveat to this solution is that it is not the most efficient. Every call to `WalletSync::wallet_setup` starts polling the Electrum-based server for `scriptPubKey`s starting from index 0.

    However, I feel like this solution is the least "destructive" to the API of `Blockchain`. Also, once the `bdk_core` sync logic is integration, we can select specific ranges of `scriptPubKey`s to sync.

2. Also note that this PR only fixes the issue for electrum-based `Blockchain` implementations (currently `blockchain::electrum` and `blockchain::esplora` only).

3. Another thing to note is that, although `Database` assumes 1-2 keychains, the current `WalletSync` "feedback" only returns one number (which is interpreted as the larger "missing count" of the two keychains). This is done for simplicity, and because we are planning to only have one keychain per database in the future.

    https://github.com/bitcoindevkit/bdk/blob/f0c876e7bf38566c0d224cbe421ee312ffc06660/src/blockchain/mod.rs#L157-L161

4. Please have a read of https://github.com/bitcoindevkit/bdk/pull/672#issuecomment-1186929465 for additional context.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/blockchain/script_sync.rs
+++ b/src/blockchain/script_sync.rs
@@ -5,6 +5,7 @@ returns associated transactions i.e. electrum.
 #![allow(dead_code)]
 use crate::{
     database::{BatchDatabase, BatchOperations, DatabaseUtils},
+    error::MissingCachedScripts,
     wallet::time::Instant,
     BlockTime, Error, KeychainKind, LocalUtxo, TransactionDetails,
 };
@@ -34,11 +35,12 @@ pub fn start<D: BatchDatabase>(db: &D, stop_gap: usize) -> Result<Request<'_, D>
     let scripts_needed = db
         .iter_script_pubkeys(Some(keychain))?
         .into_iter()
-        .collect();
+        .collect::<VecDeque<_>>();
     let state = State::new(db);
 
     Ok(Request::Script(ScriptReq {
         state,
+        initial_scripts_needed: scripts_needed.len(),
         scripts_needed,
         script_index: 0,
         stop_gap,
@@ -50,6 +52,7 @@ pub fn start<D: BatchDatabase>(db: &D, stop_gap: usize) -> Result<Request<'_, D>
 pub struct ScriptReq<'a, D: BatchDatabase> {
     state: State<'a, D>,
     script_index: usize,
+    initial_scripts_needed: usize, // if this is 1, we assume the descriptor is not derivable
     scripts_needed: VecDeque<Script>,
     stop_gap: usize,
     keychain: KeychainKind,
@@ -113,43 +116,71 @@ impl<'a, D: BatchDatabase> ScriptReq<'a, D> {
             self.script_index += 1;
         }
 
-        for _ in txids {
-            self.scripts_needed.pop_front();
-        }
+        self.scripts_needed.drain(..txids.len());
 
-        let last_active_index = self
+        // last active index: 0 => No last active
+        let last = self
             .state
             .last_active_index
             .get(&self.keychain)
-            .map(|x| x + 1)
-            .unwrap_or(0); // so no addresses active maps to 0
+            .map(|&l| l + 1)
+            .unwrap_or(0);
+        // remaining scripts left to check
+        let remaining = self.scripts_needed.len();
+        // difference between current index and last active index
+        let current_gap = self.script_index - last;
 
-        Ok(
-            if self.script_index > last_active_index + self.stop_gap
-                || self.scripts_needed.is_empty()
-            {
-                debug!(
-                    "finished scanning for transactions for keychain {:?} at index {}",
-                    self.keychain, last_active_index
-                );
-                // we're done here -- check if we need to do the next keychain
-                if let Some(keychain) = self.next_keychains.pop() {
-                    self.keychain = keychain;
-                    self.script_index = 0;
-                    self.scripts_needed = self
-                        .state
-                        .db
-                        .iter_script_pubkeys(Some(keychain))?
-                        .into_iter()
-                        .collect();
-                    Request::Script(self)
-                } else {
-                    Request::Tx(TxReq { state: self.state })
-                }
-            } else {
-                Request::Script(self)
-            },
-        )
+        // this is a hack to check whether the scripts are coming from a derivable descriptor
+        // we assume for non-derivable descriptors, the initial script count is always 1
+        let is_derivable = self.initial_scripts_needed > 1;
+
+        debug!(
+            "sync: last={}, remaining={}, diff={}, stop_gap={}",
+            last, remaining, current_gap, self.stop_gap
+        );
+
+        if is_derivable {
+            if remaining > 0 {
+                // we still have scriptPubKeys to do requests for
+                return Ok(Request::Script(self));
+            }
+
+            if last > 0 && current_gap < self.stop_gap {
+                // current gap is not large enough to stop, but we are unable to keep checking since
+                // we have exhausted cached scriptPubKeys, so return error
+                let err = MissingCachedScripts {
+                    last_count: self.script_index,
+                    missing_count: self.stop_gap - current_gap,
+                };
+                return Err(Error::MissingCachedScripts(err));
+            }
+
+            // we have exhausted cached scriptPubKeys and found no txs, continue
+        }
+
+        debug!(
+            "finished scanning for txs of keychain {:?} at index {:?}",
+            self.keychain, last
+        );
+
+        if let Some(keychain) = self.next_keychains.pop() {
+            // we still have another keychain to request txs with
+            let scripts_needed = self
+                .state
+                .db
+                .iter_script_pubkeys(Some(keychain))?
+                .into_iter()
+                .collect::<VecDeque<_>>();
+
+            self.keychain = keychain;
+            self.script_index = 0;
+            self.initial_scripts_needed = scripts_needed.len();
+            self.scripts_needed = scripts_needed;
+            return Ok(Request::Script(self));
+        }
+
+        // We have finished requesting txids, let's get the actual txs.
+        Ok(Request::Tx(TxReq { state: self.state }))
     }
 }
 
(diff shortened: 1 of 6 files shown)


evanlinjin:
After talking with @LLFourn in a private conversation, it seems that the solution provided in this PR does not fully solve #521.

### The problem that still exists

The problem that still exists is as follows. Given:

* A wallet established with a keychain, which contains a transaction in which at least one input spends from an owned `scriptPubKey`, and at least one output sends to a different owned `scriptPubKey`.
* The `Database` only contains the cache of one of these `scriptPubKey`s.

If the wallet performs a sync, and attempts to compile data (such as owned outputs, UTXOs and balances) from the transaction, it will miss including data from an owned `scriptPubKey`.

### The fix?

Currently, we define `stop_gap` as the largest expected (child index) gap between two "used" `scriptPubKey`s (where "used" means, contained within atleast one transaction output).

I propose adding an additional constraint to the definition of `stop_gap`. Given a transaction, any two `scriptPubKey`s referenced by that transaction (either being spent, or received to) should not have a difference in index greater than `stop_gap`.

If we can enforce the new definition of `stop_gap` in code, users of BDK can have a more predictable sync/update experience, hence fixing #521.

### How does everything work currently?

Analyzing the syncing logic (of electrum-based `Blockchain` implementations), we see that syncing is defined in 4 steps (as per defined in `blockchain::Request`.

https://github.com/bitcoindevkit/bdk/blob/9165faef95faa5cae731d1ad028d21e658a15263/src/blockchain/script_sync.rs#L15-L25

The first two steps are as follows:

1. `Request::Script`: First we send the electrum-based server a list of `scriptPubKey`s, obtaining a bunch of `txid`s (and their confirmation times).
2. `Request::TxReq`: We obtain the full raw transactions of `txid`s (if not already exist in `Database`) and compile metadata of the transaction (such as sent/received amounts).

Then the final step is where a batch update is prepared for the `Database`, in which one of these steps is to "set every UTXO we observed" (in the above steps).

As you can see, compiling sent/received balances from transactions, as well as finding owned UTXOs of new transactions require checking against the cache of owned `scriptPubKey`s. If we "under-cache" before these steps, we will "miss out" on finding crucial data.

We cannot fully fix this problem (unless we check with every possible `scriptPubKey` - but that is ridiculously expensive in computation), so we use our new definition of `stop_gap` to ensure sync behaves as expected for the user.

### How do we enforce the new `stop_gap` definition?

So we need to ensure the following are fulfilled before compiling metadata (such as new outputs, utxos and balances):

1. We have obtained all transactions of owned `scriptPubKey`s (where `scriptPubKey`s are searched for in chronological order starting from child index 0), in which the last `stop_gap+1` `scriptPubKey`s are not associated with any transactions. 
2. We have cached a `stop_gap` number of `scriptPubKey`s above the last `scriptPubKey` index we queried for during `Request::Script`.

### What changes do we need to make?

This should now all be implemented in this commit: 12afbe31ded597ff2a57d3d4ed4f0db83ee4f008

### In conclusion

Wallet sync is a difficult problem since we are either sacrificing performance for reliability, or sacrificing reliability for performance. The best thing we can do, is to make the syncing logic as predictable as we can for the developer/user.

Additionally, let's design a good API for sync logic.

evanlinjin:
Update: I found a fix without breaking API at all and solving both bugs (for electrum-based blockchains only). It is still WIP as I still need to write tests.

evanlinjin:
I just had a thought. I feel like to improve the efficiently, the `WalletSync` implementation can internally self-cache `txid`s before returning `Error::MissingCachedScripts`. This way, subsequent calls can use the cache to avoid re-polling the electrum-based server (for the `scriptPubKey`s that are already requested for).

src/blockchain/script_sync.rs
```diff
@@ -117,39 +120,69 @@ impl<'a, D: BatchDatabase> ScriptReq<'a, D> {
             self.scripts_needed.pop_front();
```
afilini: Unrelated to your PR, but I would replace this for loop with:

```
self.scripts_needed.drain(..txids.len());
```

Instead of explicitly iterating. I think it would be a bit more readable and potentially faster (an optimized drain implementation could drop all the items at once instead of iterating internally.

src/wallet/mod.rs
```diff
@@ -1689,16 +1689,75 @@ where
...
+                }
+
+                // cache and try again
+                for keychain in [KeychainKind::External, KeychainKind::Internal] {
+                    let (_, keychain) = self._get_descriptor_for_keychain(keychain);
+                    let from = self
+                        .database
+                        .borrow()
+                        .iter_script_pubkeys(Some(keychain))?
+                        .len() as _;
```
afilini: I would try to use the returned `last_count` value from the `MissingCachedScripts` error, to avoid this lookup which could be relatively slow in case the database has a lot of scripts.

src/wallet/mod.rs
```diff
@@ -1689,16 +1689,75 @@ where
...
-                blockchain.wallet_setup(self.database.borrow_mut().deref_mut(), progress,)
-            )?;
+
+        // first run, we record progress...
+        let mut sync_res = if run_setup {
+            maybe_await!(blockchain.wallet_setup(self.database.borrow_mut().deref_mut(), progress))
         } else {
-            maybe_await!(blockchain.wallet_sync(self.database.borrow_mut().deref_mut(), progress,))?;
+            maybe_await!(blockchain.wallet_sync(self.database.borrow_mut().deref_mut(), progress))
+        };
```
afilini: I think you can avoid duplicating this piece of code by moving this inside the loop, and then breaking if the descriptor is not derivable.

Something like:

```rust
for _ in 0..MAX_ROUNDS {
    let sync_res = // .... ;

    if !is_derivable {
        break;
    }

    // do everything else to get the database ready for the next iteration
}
```

afilini: I'm not sure what to do with the progress, I feel like we should have our own "internal" progress (that can be cloned and reused) that wraps the real progress and maps the first iteration to a maximum of like 70%, then if there are more iterations each has a lower "weight", like the second iteration going to 0 to 100 only causes the overall progress to go from 70 to 80 maybe and then at the end we just set it to 100.

It's not great but I can't think of any other way to predict how many iterations there's going to be. We should adjust those weights based on probability, most of the times it's just one (so that gets a larger weight).

If you have better ideas let me know.

afilini (commented):
Concept ACK, a few comments.

src/wallet/mod.rs
evanlinjin: I did something else to avoid duplicating code, let me if you are happy with my solution :)

Regarding the `SyncProgress` object... Because right now electrum-based blockchains don't actually report progress at all, and only electrum-based blockchains will return `Error::MissingCachedScripts` (for now), I feel like we shouldn't bother changing behavior for it in this PR?

Maybe we should do another ticket to fix progress-reporting behavior. I realized that not all sync attempts will report progress (for example, rpc only reports sync progress on it's initial sync).

evanlinjin:
@afilini @LLFourn would it make sense to try to fix the same bugs in the other `Blockchain` implementations in this PR?

afilini:
I think it would be nice to have everything in a single PR, if you can fix those as well we can merge everything together.

evanlinjin:
> I think it would be nice to have everything in a single PR, if you can fix those as well we can merge everything together.

Fair, I've created a separate ticket (#677) since it seems the "nature" of the bugs are a little different, but I'll try tackle them in this PR as well.

evanlinjin:
@afilini I started working on fixing sync for RPC... But because sync works very different, so the nature of the bug is also really different. I reckon it should be tackled in a separate PR because it feels like a completely different set of problems.

I've created a separate ticket here (#677), would it be okay to work on this in a different PR? What do you think?

afilini:
Yes, that makes sense, I'll go ahead and merge this one then

afilini (approved):
ACK 5c940c33cb1f1a166c6e28e8bbfd9cdaef6c1ab6

PR #682, opened 2022-07-25
Add a custom signer for hardware wallets

wszdexdrf:
Also adds a new test in CI for building and testing on a virtual
hardware wallet.

### Description

This PR would enable BDK users to sign transactions using a hardware wallet. It is just the beginning hence there are no complex features, but I hope not for long.
I have added a test in CI for building a ledger emulator and running the new test on it. The test is similar to the one on bitcoindevkit/rust-hwi.

### Notes to the reviewers
The PR is incomplete (and wouldn't work, as the rust-hwi in `cargo.toml` is pointing to a local crate, temporarily) as a small change is required in rust-hwi (https://github.com/bitcoindevkit/rust-hwi/pull/42).

#### All Submissions:

#### New Features:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 - Add capacity to create FeeRate from sats/kvbytes and sats/kwu.
 - Rename `as_sat_vb` to `as_sat_per_vb`. Move all `FeeRate` test to `types.rs`.
+- Add custom Harware Wallet Signer `HwiSigner` in `src/wallet/harwaresigner/` module.
 
 ## [v0.21.0] - [v0.20.0]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ rocksdb = { version = "0.14", default-features = false, features = ["snappy"], o
 cc = { version = ">=1.0.64", optional = true }
 socks = { version = "0.3", optional = true }
 lazy_static = { version = "1.4", optional = true }
+hwi = { version = "0.2.2", optional = true }
 
 bip39 = { version = "1.0.1", optional = true }
 bitcoinconsensus = { version = "0.19.0-3", optional = true }
@@ -61,6 +62,7 @@ key-value-db = ["sled"]
 all-keys = ["keys-bip39"]
 keys-bip39 = ["bip39"]
 rpc = ["bitcoincore-rpc"]
+hardware-signer = ["hwi"]
 
 # We currently provide mulitple implementations of `Blockchain`, all are
 # blocking except for the `EsploraBlockchain` which can be either async or
@@ -93,6 +95,7 @@ test-rpc = ["rpc", "electrsd/electrs_0_8_10", "electrsd/bitcoind_22_0", "test-bl
 test-rpc-legacy = ["rpc", "electrsd/electrs_0_8_10", "electrsd/bitcoind_0_20_0", "test-blockchains"]
 test-esplora = ["electrsd/legacy", "electrsd/esplora_a33e97e1", "electrsd/bitcoind_22_0", "test-blockchains"]
 test-md-docs = ["electrum"]
+test-hardware-signer = ["hardware-signer"]
 
 [dev-dependencies]
 lazy_static = "1.4"
--- a/ci/Dockerfile.ledger
+++ b/ci/Dockerfile.ledger
@@ -1,0 +1,9 @@
+# Taken from bitcoindevkit/rust-hwi
+FROM ghcr.io/ledgerhq/speculos
+
+RUN apt-get update
+RUN apt-get install wget -y
+RUN wget "https://github.com/LedgerHQ/speculos/blob/master/apps/nanos%23btc%232.1%231c8db8da.elf?raw=true" -O /speculos/btc.elf
+ADD automation.json /speculos/automation.json
+
+ENTRYPOINT ["python", "./speculos.py", "--automation", "file:automation.json", "--display", "headless", "--vnc-port", "41000", "btc.elf"]
--- a/src/database/memory.rs
+++ b/src/database/memory.rs
@@ -490,11 +490,10 @@ macro_rules! populate_test_db {
         let mut db = $db;
         let tx_meta = $tx_meta;
         let current_height: Option<u32> = $current_height;
-        let input = if $is_coinbase {
-            vec![$crate::bitcoin::TxIn::default()]
-        } else {
-            vec![]
-        };
+        let mut input = vec![$crate::bitcoin::TxIn::default()];
+        if !$is_coinbase {
+            input[0].previous_output.vout = 0;
+        }
         let tx = $crate::bitcoin::Transaction {
             version: 1,
             lock_time: 0,
--- a/src/wallet/hardwaresigner.rs
+++ b/src/wallet/hardwaresigner.rs
@@ -1,0 +1,64 @@
+// Bitcoin Dev Kit
+// Written in 2020 by Alekos Filini <alekos.filini@gmail.com>
+//
+// Copyright (c) 2020-2021 Bitcoin Dev Kit Developers
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! HWI Signer
+//!
+//! This module contains a simple implementation of a Custom signer for rust-hwi
+
+use bitcoin::psbt::PartiallySignedTransaction;
+use bitcoin::secp256k1::{All, Secp256k1};
+use bitcoin::util::bip32::Fingerprint;
+
+use hwi::error::Error;
+use hwi::types::{HWIChain, HWIDevice};
+use hwi::HWIClient;
+
+use crate::signer::{SignerCommon, SignerError, SignerId, TransactionSigner};
+
+#[derive(Debug)]
+/// Custom signer for Hardware Wallets
+///
+/// This ignores `sign_options` and leaves the decisions up to the hardware wallet.
+pub struct HWISigner {
+    fingerprint: Fingerprint,
+    client: HWIClient,
+}
+
+impl HWISigner {
+    /// Create a instance from the specified device and chain
+    pub fn from_device(device: &HWIDevice, chain: HWIChain) -> Result<HWISigner, Error> {
+        let client = HWIClient::get_client(device, false, chain)?;
+        Ok(HWISigner {
+            fingerprint: device.fingerprint,
+            client,
+        })
+    }
+}
+
+impl SignerCommon for HWISigner {
+    fn id(&self, _secp: &Secp256k1<All>) -> SignerId {
+        SignerId::Fingerprint(self.fingerprint)
+    }
+}
+
+/// This implementation ignores `sign_options`
+impl TransactionSigner for HWISigner {
+    fn sign_transaction(
+        &self,
+        psbt: &mut PartiallySignedTransaction,
+        _sign_options: &crate::SignOptions,
+        _secp: &crate::wallet::utils::SecpCtx,
+    ) -> Result<(), SignerError> {
+        psbt.combine(self.client.sign_tx(psbt)?.psbt)
+            .expect("Failed to combine HW signed psbt with passed PSBT");
+        Ok(())
+    }
+}
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -48,6 +48,9 @@ pub(crate) mod utils;
 #[cfg_attr(docsrs, doc(cfg(feature = "verify")))]
 pub mod verify;
 
+#[cfg(feature = "hardware-signer")]
+pub mod hardwaresigner;
+
 pub use utils::IsDust;
 
 #[allow(deprecated)]
@@ -5413,5 +5416,34 @@ pub(crate) mod test {
             .unwrap();
         // ...and checking that everything is fine
         assert_fee_rate!(psbt, details.fee.unwrap_or(0), fee_rate);
+    }
+
+    #[cfg(feature = "test-hardware-signer")]
+    #[test]
+    fn test_create_signer() {
+        use crate::wallet::hardwaresigner::HWISigner;
+        use hwi::types::HWIChain;
+        use hwi::HWIClient;
+
+        let devices = HWIClient::enumerate().unwrap();
+        let device = devices.first().expect("No devices found");
+        let client = HWIClient::get_client(device, true, HWIChain::Regtest).unwrap();
+        let descriptors = client.get_descriptors(None).unwrap();
+        let custom_signer = HWISigner::from_device(device, HWIChain::Regtest).unwrap();
+
+        let (mut wallet, _, _) = get_funded_wallet(&descriptors.internal[0]);
+        wallet.add_signer(
+            KeychainKind::External,
+            SignerOrdering(200),
+            Arc::new(custom_signer),
+        );
+
+        let addr = wallet.get_address(LastUnused).unwrap();
+        let mut builder = wallet.build_tx();
+        builder.drain_to(addr.script_pubkey()).drain_wallet();
+        let (mut psbt, _) = builder.finish().unwrap();
+
+        let finalized = wallet.sign(&mut psbt, Default::default()).unwrap();
+        assert!(finalized);
     }
 }
--- a/src/wallet/signer.rs
+++ b/src/wallet/signer.rs
@@ -159,6 +159,16 @@ pub enum SignerError {
     InvalidSighash,
     /// Error while computing the hash to sign
     SighashError(sighash::Error),
+    /// Error while signing using hardware wallets
+    #[cfg(feature = "hardware-signer")]
+    HWIError(hwi::error::Error),
+}
+
+#[cfg(feature = "hardware-signer")]
+impl From<hwi::error::Error> for SignerError {
+    fn from(e: hwi::error::Error) -> Self {
+        SignerError::HWIError(e)
+    }
 }
 
 impl From<sighash::Error> for SignerError {
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -172,3 +172,32 @@ jobs:
         run: rustup update
       - name: Check fmt
         run: cargo fmt --all -- --config format_code_in_doc_comments=true --check
+
+  test_harware_wallet:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - version: 1.60.0 # STABLE
+          - version: 1.56.1 # MSRV
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Build simulator image
+      run: docker build -t hwi/ledger_emulator ./ci -f ci/Dockerfile.ledger
+    - name: Run simulator image
+      run: docker run --name simulator --network=host hwi/ledger_emulator &
+    - name: Install Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.9'
+    - name: Install python dependencies
+      run: pip install hwi==2.1.1 protobuf==3.20.1
+    - name: Set default toolchain
+      run: rustup default ${{ matrix.rust.version }}
+    - name: Set profile
+      run: rustup set profile minimal
+    - name: Update toolchain
+      run: rustup update
+    - name: Test
+      run: cargo test --features test-hardware-signer
--- a/ci/automation.json
+++ b/ci/automation.json
@@ -1,0 +1,30 @@
+{
+    "version": 1,
+    "rules": [
+        {
+            "regexp": "Address \\(\\d/\\d\\)|Message hash \\(\\d/\\d\\)|Confirm|Fees|Review|Amount",
+            "actions": [
+                [ "button", 2, true ],
+                [ "button", 2, false ]
+            ]
+        },
+        {
+            "text": "Sign",
+            "conditions": [
+                [ "seen", false ]
+            ],
+            "actions": [
+                [ "button", 2, true ],
+                [ "button", 2, false ],
+                [ "setbool", "seen", true ]
+            ]
+        },
+        {
+            "regexp": "Approve|Sign|Accept",
+            "actions": [
+                [ "button", 3, true ],
+                [ "button", 3, false ]
+            ]
+        }
+    ]
+}


evanlinjin:
This is exciting that `rust-hwi` is finally getting integrated into BDK :)

I'm wondering whether "Hardware Wallet" is the best name since we have a struct named `Wallet`, and there is also a portion of the Bitcoin community wary of the term "Hardware Wallet" anyway. The proposed alternative term is "Hardware Signer" and since our signing traits are already called `{}Signer`s, I think this may be a better term.

Cargo.toml
```diff
@@ -33,6 +33,7 @@ rocksdb = { version = "0.14", default-features = false, features = ["snappy"], o
 cc = { version = ">=1.0.64", optional = true }
 socks = { version = "0.3", optional = true }
 lazy_static = { version = "1.4", optional = true }
+hwi = { path = "../rust-hwi", optional = true }
```
danielabrozzoni: This should be a `version =` kinda path

src/hardwarewallet/signer.rs
```diff
@@ -0,0 +1,50 @@
...
+impl TransactionSigner for HWISigner {
+    fn sign_transaction(
+        &self,
+        psbt: &mut PartiallySignedTransaction,
+        _secp: &Secp256k1<All>,
+    ) -> Result<(), SignerError> {
+        psbt.combine(
+            self.client
+                .sign_tx(psbt)
+                .expect("Hardware Wallet couldn't sign transaction")
```
danielabrozzoni: Maybe in this case we should return an error, not panic?

wszdexdrf:
> I'm wondering whether "Hardware Wallet" is the best name since we have a struct named `Wallet`

Now that you say it, it makes sense. This isn't anywhere close to the `Wallet` struct in terms of functionality... so maybe Hardware Signer is better...

danielabrozzoni (commented):
Sorry for the early review, I was just too excited to leave this PR alone 😁 

This is looking great! I'd split this in multiple commits, at least once for the actual code and one for the CI.

Cargo.toml
wszdexdrf: Yep. I specified a local path since I know this won't work on the published version of hwi (and I did the testing on a local crate so just left it like that 😅). I opened a PR on rust-hwi for this.

danielabrozzoni: Ah ok, I can merge and tag a new minor version then

src/hardwarewallet/signer.rs
wszdexdrf: Hm.. The thing is, the errors (exceptions) returned by `sign_tx` are all mostly of the same type (`BadArgumentError`), only with different messages. So making good use of them (converting them into correct error types) will be rather... difficult. The other option is to create a separate type for python exceptions, which I feel isn't really solving the problem, just moving it.

danielabrozzoni: > only with different messages

But maybe those messages are useful to the user, to understand why the hw is not signing?

Cargo.toml
danielabrozzoni: Btw, while I wait before tagging the new version, you can use

`hwi = { git = "https://github.com/bitcoindevkit/rust-hwi.git" }`

so that you can run the CI :) Also, you can specify a branch, in case you have more fixes to rust-hwi but can't wait for my (slow :snail:) merges

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

src/hardwaresigner/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn test_create_signer() {
+        let descriptors = get_first_device().get_descriptors(None).unwrap();
+        let devices = HWIClient::enumerate().unwrap();
+        let custom_signer =
+            HWISigner::from_device(devices.first().unwrap(), types::HWIChain::Test).unwrap();
+
+        let mut wallet = Wallet::new(
+            &descriptors.internal[0],
+            Some(&descriptors.receive[0]),
+            Network::Testnet,
```
danielabrozzoni: As I was telling you in private, this should be regtest

src/hardwaresigner/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+                "No devices found. Either plug in a hardware wallet, or start a simulator.",
+            ),
+            true,
+            types::HWIChain::Test,
+        )
+        .unwrap()
+    }
+
+    #[test]
+    fn test_create_signer() {
```
danielabrozzoni: In BDK we have two different types of tests:
- unit tests
- integration tests, which use blockchain backends

This should probably be a unit test, and should have the same structure as the other ones. See, for example, `test_sign_single_xprv` in `src/wallet/mod.rs`: there's no blockchain backend being created, only the transaction creation/signing part is tested. This test should also probably be moved in `src/wallet/mod.rs`.

Then, integration tests could be useful as well. To write one, you should modify the `bdk_blockchain_tests` in `src/testutils/blockchain_tests.rs`. Those tests will be run against every backend we support.

danielabrozzoni:
I've just published hwi 0.2.1 btw :)

danielabrozzoni:
Btw, I don't like that much having the hw module under src/hardwaresigner, it feels out of place... I'd say it should be either in src/wallet/signer.rs, or maybe you could make signer.rs a module and put the hardwaresigner in src/wallet/signer/hardwaresigner.rs, as you wish

src/database/memory.rs
```diff
@@ -493,7 +493,10 @@ macro_rules! populate_test_db {
         let input = if $is_coinbase {
             vec![$crate::bitcoin::TxIn::default()]
         } else {
-            vec![]
+            vec![$crate::bitcoin::TxIn {
```
danielabrozzoni: This is not really well-formatted.

Also, regarding the commit name: "make changes" is not really informative...

src/wallet/mod.rs
```diff
@@ -5414,4 +5417,44 @@ pub(crate) mod test {
...
+
+    #[cfg(feature = "hardware-signer")]
+    #[test]
+    fn test_create_signer() {
+        use hwi::{types, HWIClient};
+
+        use crate::wallet::hardwaresigner::signer::HWISigner;
+
+        let devices = HWIClient::enumerate().unwrap();
+        // let derivation_path = DerivationPath::from_str("m/84'/1'/0'/0/0").unwrap();
```
danielabrozzoni: Leftover?

src/wallet/mod.rs
```diff
@@ -5414,4 +5417,44 @@ pub(crate) mod test {
...
+        let addr = wallet.get_address(LastUnused).unwrap();
+        let mut builder = wallet.build_tx();
+        builder.drain_to(addr.script_pubkey()).drain_wallet();
+        let (mut psbt, _) = builder.finish().unwrap();
+
+        let finalized = wallet.sign(&mut psbt, Default::default()).unwrap();
+        assert!(finalized);
+
+        // let extracted = psbt.extract_tx();
+        // assert_eq!(extracted.input[0].witness.len(), 2);
```
danielabrozzoni: Should this be uncommented?

src/wallet/signer.rs
```diff
@@ -129,7 +129,7 @@ impl From<Fingerprint> for SignerId {
 }
 
 /// Signing error
-#[derive(Debug, PartialEq, Eq, Clone)]
+#[derive(Debug)]
```
danielabrozzoni: Why are you removing PartialEq, Eq, Clone? Can't you make `hwi::error::Error` implement those?

src/wallet/hardwaresigner/signer.rs
```diff
@@ -0,0 +1,53 @@
...
+    fn id(&self, _secp: &Secp256k1<All>) -> SignerId {
+        SignerId::Fingerprint(self.fingerprint)
+    }
+}
+
+impl TransactionSigner for HWISigner {
+    fn sign_transaction(
+        &self,
+        psbt: &mut PartiallySignedTransaction,
+        _sign_options: &crate::SignOptions,
```
danielabrozzoni: It should be documented somewhere that we are ignoring the `sign_options`

src/wallet/signer.rs
wszdexdrf: Unfortunately, no. `hwi::error::Error` contains `serde_json::error::Error` and `std::io::Error` which do not implement these traits. I could create a new implementation, but I don't think it would be useful (for `hwi::error::Error`). If these traits are absolutely necessary, then I think I could add an implementation.

src/wallet/hardwaresigner/signer.rs
```diff
@@ -0,0 +1,55 @@
+//! HWI Signer
```
afilini: This needs the usual header

src/wallet/hardwaresigner/signer.rs
```diff
@@ -0,0 +1,55 @@
...
+use bitcoin::{
+    psbt::PartiallySignedTransaction,
+    secp256k1::{All, Secp256k1},
+    util::bip32::Fingerprint,
+};
+use hwi::{
+    error::Error,
+    types::{HWIChain, HWIDevice},
+    HWIClient,
+};
```
afilini: We normally use a different import style

src/wallet/hardwaresigner/signer.rs
```diff
@@ -0,0 +1,55 @@
...
+/// This implementation ignores `sign_options`
+impl TransactionSigner for HWISigner {
+    fn sign_transaction(
+        &self,
+        psbt: &mut PartiallySignedTransaction,
+        _sign_options: &crate::SignOptions,
+        _secp: &crate::wallet::utils::SecpCtx,
+    ) -> Result<(), SignerError> {
+        psbt.combine(self.client.sign_tx(psbt)?.psbt)
+            .expect("Failed to combine HW signed psbt with passed PSBT");
```
afilini: Is it safe to have an "expect" here? In which cases can the psbt combine fail?

src/wallet/signer.rs
afilini: Maybe it's just easier to change hwi so that its error type implements those

src/wallet/mod.rs
```diff
@@ -5414,4 +5417,40 @@ pub(crate) mod test {
         // ...and checking that everything is fine
         assert_fee_rate!(psbt, details.fee.unwrap_or(0), fee_rate);
     }
+
+    #[cfg(feature = "hardware-signer")]
+    #[test]
+    fn test_create_signer() {
```
afilini, line 5423: Maybe this test should be ignored by default? Since it depends on external setup

afilini: Or alternatively you can add a separate feature to test hardware signers, and if explicitly enabled we run this test

src/wallet/signer.rs
danielabrozzoni: We could use `String` instead of `hwi::error::Error` here, or `String` inside hwi instead of `serde_json::error::Error` and `std::io::Error`

src/wallet/hardwaresigner/signer.rs
wszdexdrf: Looking at the code for `psbt.combine()`, I could find 2 possible errors that can be returned, (1) when the 2 psbts describe different txs (shouldn't happen as the hardware wallet should only sign the tx) and (2) `CombineInconsistentKeySources ` which, AFAIK, shouldn't happen unless the user wipes and resets the device _while_ he is signing a transaction XD.

Cargo.toml
```diff
@@ -33,6 +33,8 @@ rocksdb = { version = "0.14", default-features = false, features = ["snappy"], o
 cc = { version = ">=1.0.64", optional = true }
 socks = { version = "0.3", optional = true }
 lazy_static = { version = "1.4", optional = true }
+# To be changed as soon as https://github.com/bitcoindevkit/rust-hwi/pull/44 is merged and released
```
danielabrozzoni: Done :)

src/wallet/hardwaresigner/signer.rs
afilini: Thanks for looking into this!

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Change the meaning of the `fee_amount` field inside `CoinSelectionResult`: from now on the `fee_amount` will represent only the fees asociated with the utxos in the `selected` field of `CoinSelectionResult`.
 - New `RpcBlockchain` implementation with various fixes.
 - Return balance in separate categories, namely `confirmed`, `trusted_pending`, `untrusted_pending` & `immature`.
+- Add a custom Signer for Hardware Wallets.
```
afilini: This should be moved up in the unreleased section. Sorry for not noticing it before

vladimirfomene (commented):
utACK

src/database/memory.rs
```diff
@@ -493,7 +493,13 @@ macro_rules! populate_test_db {
...
+                previous_output: $crate::bitcoin::OutPoint {
+                    txid: Default::default(),
+                    vout: 5,
+                },
+                ..Default::default()
```
rajarshimaitra: Do we really need to hardcode transaction inputs in this way?? Seems little aggressive to me.. We are not using the transaction inputs from the `tx_meta` in any of the tests, so not an issue for now, but in future this can get awkward.. And whats the issue when you tried the #624 way?

src/wallet/hardwaresigner/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! Hardware wallets
+//!
+//! This module contains modules required for working with Hardware Wallets.
+//! This module can contain various vendor specific code, if required.
+
+pub mod signer;
```
rajarshimaitra: Why we need a separate module folder for this? Are we expecting more signer modules in future?

src/wallet/mod.rs
```diff
@@ -5414,4 +5417,40 @@ pub(crate) mod test {
...
+
+        let devices = HWIClient::enumerate().unwrap();
+        let client = HWIClient::get_client(
+            devices.first().expect(
+                "No devices found. Either plug in a hardware wallet, or start a simulator.",
+            ),
+            true,
+            types::HWIChain::Regtest,
+        )
+        .unwrap();
```
rajarshimaitra: When running the test I am getting an error at this line 
```
thread 'wallet::test::test_create_signer' panicked at 'called `Result::unwrap()` on an `Err` value: PyErr("TypeError: get_client() takes from 2 to 4 positional arguments but 5 were given")', src/wallet/mod.rs:5436:10
```

Am I doing something wrong??

ci/Dockerfile.ledger
```diff
@@ -0,0 +1,9 @@
+# Taken from bitcoindevkit/rust-hwi
+FROM ghcr.io/ledgerhq/speculos
+
+RUN apt-get update
+RUN apt-get install wget -y
+RUN wget "https://github.com/LedgerHQ/speculos/blob/master/apps/nanos%23btc%232.1%231c8db8da.elf?raw=true" -O /speculos/btc.elf
+ADD automation.json /speculos/automation.json
+
+ENTRYPOINT ["python", "./speculos.py", "--automation", "file:automation.json", "--display", "headless", "--vnc-port", "41000", "btc.elf"]
```
rajarshimaitra, line 9: This is really nice that we can have ci testing with docker for devices. Is it possible to have such docker for other major vendor like coldcard and trezor?

rajarshimaitra (commented):
Review ACK.. This is really exciting.. Very much pumped for BDK HWI integration. Thanks for working on this.. Below I have few comments..

src/wallet/hardwaresigner/mod.rs
wszdexdrf: Yes, I think there should be more signers/functions, specifically for hardware wallets, in the future. Hardware wallets differ greatly in function (even vis-a-vis HWI) and to take full advantage of them, we would need some vendor/device specific code in both rust-hwi and BDK in the future.

src/wallet/mod.rs
wszdexdrf: Make sure you have hwi library >=2.1.1. The latest version of hwi through pip should work fine.

ci/Dockerfile.ledger
wszdexdrf: Yes indeed! We do test with trezor emulator on rust-hwi (and coldcard is work-in-progress). I didn't want to bog down BDK tests, as the emulators take a lot of time to build and run, hence I thought we would do more testing on rust-hwi and then just test one device with BDK. 

Just to add, only ledger supplies a pre-built docker image. For the rest, we build one ourselves. Hence the long time.

src/database/memory.rs
wszdexdrf: I am not really sure why this is so, but the trezor device doesn't behave properly if we leave that empty. When I test signing a psbt with trezor using code from #624, (and what is there on master), trezor (emulator) returns a broken psbt which cannot be deserialized.

rajarshimaitra: Ok I did some debugging and its failing because we actually don't set the transaction inputs in `tx_meta` in the `testutils` macro, unless its given in the macro call. 
https://github.com/bitcoindevkit/bdk/blob/c9b1b6d076cedcc389128d7d48d287aa2d1aa70d/src/testutils/mod.rs#L130-L133

And in `get_funded_wallet` we are calling `testutils` without any input. Because its a fake transaction, there's no input to specify.. So it made sense to have empty input for generic cases, but would fail where an input is expected. Like this test here..  

So if you add a hardcoded fake input in `get_funded_wallet` and then apply the #624 logic in `populate_test_db` it will work as expected.. I would suggest hard coding the input in `get_funded_wallet` instead of `populate_test_db`, as the later can be used in cases where we do know about transaction inputs..

Another option is to call `testutils` with `@coinbase true` in `get_funded_wallet`, so it will use the `TxIn::deafult()` value, which will also work..

src/wallet/hardwaresigner/mod.rs
rajarshimaitra: That makes sense.. Thanks..

src/wallet/mod.rs
rajarshimaitra: Yup.. That was problem.. Working now..

rajarshimaitra (commented):
tACK 2860c418838821a3aa9a16c468880e9d2621206f

src/database/memory.rs
wszdexdrf: A new arguement in `get_funded_wallet()` for `@coinbase`. I think this would be the best (and easiest). With that I could skip the changes in `populate_test_db`.

src/wallet/hardwaresigner/mod.rs
danielabrozzoni: I'm not sure if we'll ever add more signers, `HWI` is specifically built for handling the specific code for vendors/devices

src/wallet/mod.rs
```diff
@@ -1867,8 +1870,9 @@ where
 }
 
 /// Return a fake wallet that appears to be funded for testing.
-pub fn get_funded_wallet(
+pub fn get_funded_wallet_with_coinbase(
     descriptor: &str,
+    coinbase: bool,
```
rajarshimaitra: Why not just always call coinbase?? There is no way for us to ever specify any inputs in `get_funded_wallet()` anyway.. So doesn't hurt to always have the coinbase input than empty inputs, which is not a valid transaction.. 

And having the funding transaction, coinbase, even if its fake, makes semantic sense..

src/wallet/mod.rs
```diff
@@ -1902,11 +1906,17 @@ pub fn get_funded_wallet(
...
 
     (wallet, descriptors, txid)
 }
 
+/// Return a fake wallet that appears to be funded for testing.
+pub fn get_funded_wallet(
+    descriptor: &str,
+) -> (Wallet<AnyDatabase>, (String, Option<String>), bitcoin::Txid) {
+    get_funded_wallet_with_coinbase(descriptor, false)
+}
```
rajarshimaitra: I couldn't think of any situation where we might need empty inputs in this transaction.. So this function feels redundant..

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Add a custom Signer for Hardware Wallets.
```
rajarshimaitra: "Add custom Harware Wallet Signer`HwiSigner` in `src/wallet/harwaresigner/` module". 

Or something along this line to elaborate a bit on it would be helpful.

rajarshimaitra (changes requested):
Thanks for the quick update.. Few more comments..

src/wallet/hardwaresigner/mod.rs
rajarshimaitra: Yeah thats what I was thinking too.. IIUC HWI can handle any vendor's device if they support the API.. So in that case we should just need one module file to cover it..

src/wallet/mod.rs
wszdexdrf: > And having the funding transaction, coinbase, even if its fake, makes semantic sense..

I agree with that, 100%. However, it creates some problems when height is specified in a test, as coinbase inputs need to mature before being used. A few tests were failing if I used coinbase in all tests.

src/wallet/hardwaresigner/mod.rs
wszdexdrf: Okay. I will refactor the code into one file.

src/wallet/mod.rs
rajarshimaitra: Right.. And this gets fixed by just adding a 100 to the height values of those tests..

wszdexdrf: Okay. I was hoping I could get by without changing the test... I will fix them.

rajarshimaitra: Yeah those are out of context a bit.. But small changes only in two tests is not that critical IMO.. Thanks for the quick updates.. I will go over the fully once more..

src/wallet/mod.rs
```diff
@@ -5414,4 +5417,40 @@ pub(crate) mod test {
         // ...and checking that everything is fine
         assert_fee_rate!(psbt, details.fee.unwrap_or(0), fee_rate);
     }
+
+    #[cfg(feature = "test-hardware-signer")]
+    #[test]
+    fn test_create_signer() {
+        use hwi::{types, HWIClient};
```
danielabrozzoni: nit: instead of including `types` and always using `types::HWIChain`, just include `HWIChain`

src/wallet/mod.rs
```diff
@@ -5414,4 +5417,40 @@ pub(crate) mod test {
...
+
```
danielabrozzoni: nit: remove this empty line

src/wallet/mod.rs
```diff
@@ -5414,4 +5417,40 @@ pub(crate) mod test {
...
+            devices.first().expect(
+                "No devices found. Either plug in a hardware wallet, or start a simulator.",
+            ),
+            true,
+            types::HWIChain::Regtest,
+        )
+        .unwrap();
+        let descriptors = client.get_descriptors(None).unwrap();
+        let custom_signer =
+            HWISigner::from_device(devices.first().unwrap(), types::HWIChain::Regtest).unwrap();
```
danielabrozzoni: nit: you could save `devices.first().expect("...")` (above) in a variable, and use it here as well

src/wallet/hardwaresigner.rs
```diff
@@ -0,0 +1,63 @@
...
+use bitcoin::util::bip32::Fingerprint;
+
+use hwi::error::Error;
+use hwi::types::{HWIChain, HWIDevice};
+use hwi::HWIClient;
+
+use crate::signer::{SignerCommon, SignerError, SignerId, TransactionSigner};
+
+#[derive(Debug)]
+/// Custom Signer for Hardware Wallets
```
danielabrozzoni: nit: signer shouldn't be capitalized

src/wallet/hardwaresigner.rs
```diff
@@ -0,0 +1,63 @@
...
+/// This ignores `sign_options` and leaves the decisions upto the hardware wallet.
```
danielabrozzoni: upto -> up to

Also, you should leave an empty line above this one

src/wallet/mod.rs
danielabrozzoni: Sorry if I'm late to the conversation, but can't we just make sure that in `populate_test_db` we create a sane transaction with inputs in all cases?

```diff
diff --git a/src/database/memory.rs b/src/database/memory.rs
index 7d806eb..ad1eda9 100644
--- a/src/database/memory.rs
+++ b/src/database/memory.rs
@@ -490,11 +490,10 @@ macro_rules! populate_test_db {
         let mut db = $db;
         let tx_meta = $tx_meta;
         let current_height: Option<u32> = $current_height;
-        let input = if $is_coinbase {
-            vec![$crate::bitcoin::TxIn::default()]
-        } else {
-            vec![]
-        };
+        let mut input = vec![$crate::bitcoin::TxIn::default()];
+        if !$is_coinbase {
+            input[0].previous_output.vout = 0;
+        }
         let tx = $crate::bitcoin::Transaction {
             version: 1,
             lock_time: 0,
```

So that we don't have to modify tests...

What do you think?

rajarshimaitra: One thing that comes to mind while setting the input in `populate_test_db` is, what if we need to insert a fake transaction in the db, where the input matters? Even if we set them in `tx_meta` of `testutils`, it will get ignored in this function.. And we will always have a specific input in any transaction added via populate_test_db.. 

What I think can be more useful (I partially tried this in #624) make sure populate test db creates the transaction as per the `tx_meta` passed in.. If it doesn't have inputs (which is the case for `get_funded_wallet`) make it a coinbase..

danielabrozzoni: > make sure populate test db creates the transaction as per the tx_meta passed in..

Right, that's a good idea!

>  If it doesn't have inputs (which is the case for get_funded_wallet) make it a coinbase..

What about the opposite: if it doesn't have the inputs, we use a dummy default (like all zeros for the txid, and 0 for the vout, so it's not a coinbase) - I think this would be cleaner, as:
- We rarely need coinbase inputs for tests
- Coinbase inputs are easy to specify (just use `Outpoint::null()`)
- Coinbase inputs might bring some troubles in tests (you need to have 100 more confirmations etc)

rajarshimaitra: That makes total sense to me.. I don't think we will need this input in any situation, and for coinbase situations we are already specifying them.. ACK on that will be more cleaner..

danielabrozzoni (approved):
ACK 86c1a75ac1af5f38080bdf59651aa62ca40782c8 - thanks for taking care of this!

rajarshimaitra (approved):
tACK 86c1a75ac1af5f38080bdf59651aa62ca40782c8

afilini (approved):
ACK 138acc3b7d137788d0518182e2167504e58ebc48

PR #683, opened 2022-07-25
Fix wallet sync for `RpcBlockchain`

### LINKED ISSUE #677: `WalletSync` issues with `RpcBlockchain` and `CompactFiltersBlockchain`
### The bug

After looking through the codebase for `RpcBlockchain`, it looks as if `importdescriptors` would only be called on the first sync, or not at all if there were already more than 100 `scriptPubKey`s cached before the first sync attempt. In other words, the Bitcoin Core wallet would only be aware of 100 `scriptPubKey`s or none at all.

The `CompactFiltersBlockchain` implementation may have similar issues to the electrum-based `Blockchain` implementations (fixed in #672), where owned outputs can be missed when compiling data from raw transactions (when `scriptPubKey`s are under-cached), and subsequent syncs will not rediscover these missed outputs.

Note that these deductions are not tested, but mere assumptions from scanning over the codebase.

### Tests to write

I recommend writing the following tests (to confirm the existence of these bugs):

1. After caching greater-than-100 `scriptPubKey`s before doing initial `Wallet::sync`, we should be able to detect correct balance and owned UTXO set (assuming that all txs interact with owned `scriptPubKey`s with derivation index less-than-100).

2. Given a positive integer that is greater-than-0 (`x`), after doing the initial `Wallet::sync`, create a self-send transaction, sending from `scriptPubKey` at derivation index 99 to derivation index 99+`x`. Cache at-least `x` more `scriptPubKey`s and re-sync. `Wallet` should have the correct balance and output-set.

### Additional context

Feel free to look at PR #672 (which fixes #521 and #451 for electrum-based blockchains). However, note that the sync logic differs from RPC and Compact Block Filters, so the fix will be different.

If I have understood the codebase correctly, and these bugs exist, it will be good to have a fix ASAP since these are critical problems.

evanlinjin:
Fixes #677

### Description

Unfortunately to fix all the problems, I had to do a complete re-implementation of `RpcBlockchain`.

**The new implementation fixes the following:**
* We can track more than 100 scriptPubKeys
* We can obtain more than 1000 transactions per sync
* Transaction "metadata" for already-syned transactions are updated when we introduce new scriptPubKeys

**`RpcConfig` changes:**
* Introduce `RpcSyncParams`.
* Remove `RpcConfig::skip_blocks` (this is replaced by `RpcSyncParams::start_time`).

### Notes to the reviewers

* The `RpcConfig` structure is changed. It will be good to confirm whether this is an okay change.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/blockchain/rpc.rs
+++ b/src/blockchain/rpc.rs
@@ -424,24 +424,29 @@ impl DbState {
             // check if tx has an immature coinbase output (add to updated UTXOs)
             // this is required because `listunspent` does not include immature coinbase outputs
             if tx_res.detail.category == GetTransactionResultDetailCategory::Immature {
-                // let vout = tx_res.detail.vout;
-                // let txout = raw_tx.output.get(vout as usize).cloned().ok_or_else(|| {
-                //     Error::Generic(format!(
-                //         "Core RPC returned detail with invalid vout '{}' for tx '{}'",
-                //         vout, tx_res.info.txid,
-                //     ))
-                // })?;
-                // println!("got immature detail!");
+                let txout = raw_tx
+                    .output
+                    .get(tx_res.detail.vout as usize)
+                    .cloned()
+                    .ok_or_else(|| {
+                        Error::Generic(format!(
+                            "Core RPC returned detail with invalid vout '{}' for tx '{}'",
+                            tx_res.detail.vout, tx_res.info.txid,
+                        ))
+                    })?;
 
-                // if let Some((keychain, _)) = db.get_path_from_script_pubkey(&txout.script_pubkey)? {
-                //     let utxo = LocalUtxo {
-                //         outpoint: OutPoint::new(tx_res.info.txid, d.vout),
-                //         txout,
-                //         keychain,
-                //         is_spent: false,
-                //     };
-                //     self.updated_utxos.insert(utxo);
-                // }
+                if let Some((keychain, index)) =
+                    db.get_path_from_script_pubkey(&txout.script_pubkey)?
+                {
+                    let utxo = LocalUtxo {
+                        outpoint: OutPoint::new(tx_res.info.txid, tx_res.detail.vout),
+                        txout,
+                        keychain,
+                        is_spent: false,
+                    };
+                    self.updated_utxos.insert(utxo);
+                    self._update_last_index(keychain, index);
+                }
             }
 
             // update tx deltas
@@ -511,7 +516,7 @@ impl DbState {
         let new_utxos = core_utxos.difference(&self.utxos).cloned();
 
         // add to updated utxos
-        self.updated_utxos = spent_utxos.chain(new_utxos).collect();
+        self.updated_utxos.extend(spent_utxos.chain(new_utxos));
 
         Ok(self)
     }
@@ -604,6 +609,7 @@ impl DbState {
         // update utxos
         self.updated_utxos
             .iter()
+            .inspect(|&utxo| println!("updating: {:?}", utxo.txout))
             .try_for_each(|utxo| batch.set_utxo(utxo))?;
 
         // update last indexes
--- a/src/blockchain/rpc.rs
+++ b/src/blockchain/rpc.rs
@@ -26,21 +26,23 @@
 //!     },
 //!     network: bdk::bitcoin::Network::Testnet,
 //!     wallet_name: "wallet_name".to_string(),
-//!     skip_blocks: None,
+//!     sync_params: None,
 //! };
 //! let blockchain = RpcBlockchain::from_config(&config);
 //! ```
 
-use crate::bitcoin::consensus::deserialize;
 use crate::bitcoin::hashes::hex::ToHex;
-use crate::bitcoin::{Address, Network, OutPoint, Transaction, TxOut, Txid};
+use crate::bitcoin::{Network, OutPoint, Transaction, TxOut, Txid};
 use crate::blockchain::*;
-use crate::database::{BatchDatabase, DatabaseUtils};
+use crate::database::{BatchDatabase, BatchOperations, DatabaseUtils};
 use crate::descriptor::get_checksum;
+use crate::error::MissingCachedScripts;
 use crate::{BlockTime, Error, FeeRate, KeychainKind, LocalUtxo, TransactionDetails};
+use bitcoin::Script;
 use bitcoincore_rpc::json::{
-    GetAddressInfoResultLabel, ImportMultiOptions, ImportMultiRequest,
-    ImportMultiRequestScriptPubkey, ImportMultiRescanSince,
+    GetTransactionResult, GetTransactionResultDetailCategory, ImportMultiOptions,
+    ImportMultiRequest, ImportMultiRequestScriptPubkey, ImportMultiRescanSince,
+    ListTransactionResult, ScanningDetails,
 };
 use bitcoincore_rpc::jsonrpc::serde_json::{json, Value};
 use bitcoincore_rpc::Auth as RpcAuth;
@@ -49,7 +51,8 @@ use log::debug;
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
 use std::path::PathBuf;
-use std::str::FromStr;
+use std::thread;
+use std::time::Duration;
 
 /// The main struct for RPC backend implementing the [crate::blockchain::Blockchain] trait
 #[derive(Debug)]
@@ -60,11 +63,8 @@ pub struct RpcBlockchain {
     is_descriptors: bool,
     /// Blockchain capabilities, cached here at startup
     capabilities: HashSet<Capability>,
-    /// Skip this many blocks of the blockchain at the first rescan, if None the rescan is done from the genesis block
-    skip_blocks: Option<u32>,
-
-    /// This is a fixed Address used as a hack key to store information on the node
-    _storage_address: Address,
+    /// Sync parameters.
+    sync_params: RpcSyncParams,
 }
 
 /// RpcBlockchain configuration options
@@ -78,8 +78,33 @@ pub struct RpcConfig {
     pub network: Network,
     /// The wallet name in the bitcoin node, consider using [crate::wallet::wallet_name_from_descriptor] for this
     pub wallet_name: String,
-    /// Skip this many blocks of the blockchain at the first rescan, if None the rescan is done from the genesis block
-    pub skip_blocks: Option<u32>,
+    /// Sync parameters
+    pub sync_params: Option<RpcSyncParams>,
+}
+
+/// Sync parameters for Bitcoin Core RPC.
+///
+/// In general, BDK tries to sync `scriptPubKey`s cached in [`crate::database::Database`] with
+/// `scriptPubKey`s imported in the Bitcoin Core Wallet. These parameters are used for determining
+/// how the `importdescriptors` RPC calls are to be made.
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct RpcSyncParams {
+    /// The minimum number of scripts to scan for on initial sync.
+    pub start_script_count: usize,
+    /// Time in unix seconds in which initial sync will start scanning from (0 to start from genesis).
+    pub start_time: u64,
+    /// RPC poll rate (in seconds) to get state updates.
+    pub poll_rate_sec: u64,
+}
+
+impl Default for RpcSyncParams {
+    fn default() -> Self {
+        Self {
+            start_script_count: 100,
+            start_time: 0,
+            poll_rate_sec: 3,
+        }
+    }
 }
 
 /// This struct is equivalent to [bitcoincore_rpc::Auth] but it implements [serde::Serialize]
@@ -112,27 +137,6 @@ impl From<Auth> for RpcAuth {
             Auth::UserPass { username, password } => RpcAuth::UserPass(username, password),
             Auth::Cookie { file } => RpcAuth::CookieFile(file),
         }
-    }
-}
-
-impl RpcBlockchain {
-    fn get_node_synced_height(&self) -> Result<u32, Error> {
-        let info = self.client.get_address_info(&self._storage_address)?;
-        if let Some(GetAddressInfoResultLabel::Simple(label)) = info.labels.first() {
-            Ok(label
-                .parse::<u32>()
-                .unwrap_or_else(|_| self.skip_blocks.unwrap_or(0)))
-        } else {
-            Ok(self.skip_blocks.unwrap_or(0))
-        }
-    }
-
-    /// Set the synced height in the core node by using a label of a fixed address so that
-    /// another client with the same descriptor doesn't rescan the blockchain
-    fn set_node_synced_height(&self, height: u32) -> Result<(), Error> {
-        Ok(self
-            .client
-            .set_label(&self._storage_address, &height.to_string())?)
     }
 }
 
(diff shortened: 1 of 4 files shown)


evanlinjin:
@danielabrozzoni I tried applying the fixes from PR #687 into this one, however there are some problems. You can see my attempt here: https://github.com/evanlinjin/bdk/pull/1

Let's talk about this over a call tomorrow.

evanlinjin:
We should probably check whether this fixes #598

src/blockchain/rpc.rs
```diff
@@ -178,224 +182,61 @@ impl GetBlockHash for RpcBlockchain {
...
-            debug!("rescan_blockchain from:{} to:{}", node_synced, sync_up_to);
-            self.client
-                .rescan_blockchain(Some(node_synced as usize), Some(sync_up_to as usize))?;
-            progress_update.update((sync_up_to as f32) / (current_height as f32), None)?;
-
-            self.set_node_synced_height(sync_up_to)?;
+        // this tells Core wallet where to sync from for imported scripts
+        let start_epoch = db
+            .get_sync_time()?
+            .map_or(self.sync_params.start_time, |st| st.block_time.timestamp);
```
evanlinjin: Maybe we should have a new parameter `RpcSyncParams::force_start_time` that forces every sync to use `sync_params.start_time`. This is for users that prefer reliability above all else.

src/blockchain/rpc.rs
```diff
@@ -178,224 +182,61 @@ impl GetBlockHash for RpcBlockchain {
...
+        let db_spks = db.iter_script_pubkeys(None)?;
+        let db_spk_count = db_spks.len();
+
+        // this is a hack to check whether the scripts are coming from a derivable descriptor
+        // we assume for non-derivable descriptors, the initial script count is always 1 and
+        // since we are working with two descriptors, we use 2 to determine derivability
+        let is_derivable = db_spk_count > 2;
+
+        // ensure db scripts meet start script count requirements
+        if is_derivable && db_spks.len() < self.sync_params.start_script_count {
```
danielabrozzoni: you can use `db_spk_count`

src/blockchain/rpc.rs
```diff
@@ -495,6 +330,486 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+            .into_iter()
+            .map(|tx| (tx.txid, tx))
+            .collect::<HashMap<_, _>>();
+        let utxos = db.iter_utxos()?.into_iter().collect::<HashSet<_>>();
+        let last_indexes = [KeychainKind::External, KeychainKind::Internal]
+            .iter()
+            .filter_map(|keychain| {
+                db.get_last_index(*keychain)
+                    .map(|v| v.map(|i| (*keychain, i)))
+                    .transpose()
```
danielabrozzoni: maybe here it's better for readability if you use match?

src/blockchain/rpc.rs
```diff
@@ -495,6 +330,486 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+                match &db_tx.transaction {
+                    Some(raw_tx) => raw_tx,
+                    None => {
+                        updated = true;
+                        db_tx.transaction.insert(client.get_raw_transaction(
+                            &tx_res.info.txid,
+                            tx_res.info.blockhash.as_ref(),
+                        )?)
+                    }
+                };
```
danielabrozzoni: It's a bit unclear that in the None case, you're inserting in the same thing you're matching on (took me a second). What about:

```rust
                match &mut db_tx.transaction {
                    Some(raw_tx) => raw_tx,
                    tx_opt => {
                        updated = true;
                        tx_opt.insert(client.get_raw_transaction(
                            &tx_res.info.txid,
                            tx_res.info.blockhash.as_ref(),
                        )?)
                    }
```

src/blockchain/rpc.rs
```diff
@@ -495,6 +330,486 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+                let txid = self.retained_txs.get(&db_tx.txid)?;
+                self._sent_from_raw_tx(db, db_tx.transaction.as_ref()?)
+                    .map(|sent| {
+                        if db_tx.sent != sent {
+                            Some((*txid, sent))
+                        } else {
+                            None
+                        }
+                    })
+                    .transpose()
```
danielabrozzoni: Same as above, a match here might be more readable

src/blockchain/rpc.rs
```diff
@@ -495,6 +330,486 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
danielabrozzoni, line 490: You might not need to collect into a Vec, as you're iterating them just below

src/blockchain/rpc.rs
```diff
@@ -495,6 +330,486 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+
+        // record send updates
+        sent_updates.into_iter().for_each(|(txid, sent)| {
```
danielabrozzoni: If you go with the "avoid collect" suggestion, here you can use a `for` and then you can remove the `Result` with a `?`

src/blockchain/rpc.rs
danielabrozzoni: OTOH... Why do you need two loops? Can't you record the updates directly when you found them?

src/blockchain/rpc.rs
```diff
@@ -495,6 +330,486 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+            self.updated_txs.insert(txid);
+        });
+
+        // obtain UTXOs from Core wallet
+        let core_utxos = client
+            .list_unspent(Some(0), None, None, Some(true), None)?
+            .into_iter()
+            .filter_map(|utxo_res| {
+                db.get_path_from_script_pubkey(&utxo_res.script_pub_key)
+                    .transpose()
```
danielabrozzoni: Again, maybe better with a match

src/blockchain/rpc.rs
```diff
@@ -495,6 +330,486 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+                                },
+                                is_spent: false,
+                            }
+                        })
+                    })
+            })
+            .collect::<Result<HashSet<_>, Error>>()?;
+
+        // mark "spent utxos" to be updated in database
+        let spent_utxos = self.utxos.difference(&core_utxos).cloned().map(|mut utxo| {
```
danielabrozzoni, line 523: I'm wondering what happens if:

Someone sends us a transaction `txid` with one utxo `A` for us, then it RBFs the transaction and it becomes `txid1` with one utxo `A1` for us. What happens to utxo `A`? Do we mark it as spent?

If so, can you think of a way of fixing this? Otherwise I'll just open an issue :)

src/blockchain/rpc.rs
```diff
@@ -495,6 +330,486 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+
+        // add to updated utxos
+        self.updated_utxos.extend(spent_utxos.chain(new_utxos));
+
+        Ok(self)
+    }
+
+    /// We want to filter out conflicting transactions.
+    /// Only accept transactions that are already confirmed, or existing in mempool.
+    fn _filter_tx(client: &Client, res: GetTransactionResult) -> Option<GetTransactionResult> {
```
danielabrozzoni: Even if it's internal stuff, better to remove the initial `_`.

Doing so, the compiler tells us that `filter_tx` is never used... Is that right?

src/blockchain/rpc.rs
```diff
@@ -495,6 +330,486 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+                index
+            });
+
+        if updated {
+            debug!(
+                "db state: last_index updated for {}: {}",
+                keychain.as_byte(),
+                index
+            );
+            self.updated_last_indexes.insert(keychain);
```
danielabrozzoni: I'm not sure if it's worth it to have both a `last_indexes` and a `updated_last_indexes`, as it's quite complex and not a great performance improvement. You could simply have `last_indexes` and at the end always rewrite the last index of each keychain in the database - without really caring if you didn't modify it. It's just two rows to update, not thousands, and I feel like it would make the code significantly easier to read.

src/blockchain/rpc.rs
```diff
@@ -495,6 +330,486 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+    /// Prepare db batch operations.
+    fn update_batch<D: BatchDatabase>(&self, batch: &mut D::Batch) -> Result<(), Error> {
+        let mut del_txs = 0_u32;
+
+        // delete stale txs from db
+        // stale = not retained
+        self.txs
+            .keys()
+            .filter(|&txid| !self.retained_txs.contains(txid))
+            .try_for_each(|txid| batch.del_tx(txid, false).map(|_| del_txs += 1))?;
```
danielabrozzoni: I don't think this will work, as usually `batch.del_tx` will return `Some` only if you're deleting something that has just been inserted in the batch.

src/blockchain/rpc.rs
```diff
@@ -495,6 +330,486 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+        // update last indexes
+        self.updated_last_indexes
+            .iter()
+            .map(|keychain| self.last_indexes.get_key_value(keychain).unwrap())
+            .try_for_each(|(&keychain, &index)| batch.set_last_index(keychain, index))?;
+
+        info!(
+            "db batch updates: del_txs={}, update_txs={}, update_utxos={}",
+            del_txs,
+            self.updated_txs.len(),
```
danielabrozzoni, line 634: Is this count correct? Above, you `filter_map` the `updated_txs` before inserting them

src/blockchain/rpc.rs
```diff
@@ -495,6 +330,486 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+        // stale = not retained
+        self.txs
+            .keys()
+            .filter(|&txid| !self.retained_txs.contains(txid))
+            .try_for_each(|txid| batch.del_tx(txid, false).map(|_| del_txs += 1))?;
+
+        // update txs
+        self.updated_txs
+            .iter()
+            .filter_map(|txid| self.txs.get(txid))
```
danielabrozzoni: Maybe if `self.txs.get(txid)` never returns `None` it's better to expect with an explanation?


....Unless there are some cases where `None` might be returned! But I can't think of any

src/blockchain/rpc.rs
danielabrozzoni: Also, maybe you should check those per KeychainKind: `start_script_count = 100` imho should mean 100 cached for the external and 100 cached for the internal descriptor. Otherwise you risk having  `start_script_count = 100` and, instead of 50/50, having 100 cached for the external and 0 for the internal.

If that's the case, maybe we should rename `count` -> `index`? So you can say "index = 50" instead of `count = 100` and it's clearer

src/blockchain/rpc.rs
danielabrozzoni: 👍🏻

danielabrozzoni (commented):
This review was a collaborative effort between me and @afilini, but I want to make it clear that if some comments are stupid, it's only his fault. (jk) (well, maybe not)

Concept ACK, some comments

src/blockchain/rpc.rs
evanlinjin: What would be the correct behavior? Delete it, or add another field such as "replaced"?

src/blockchain/rpc.rs
evanlinjin: I feel like it's a matter of taste, but I've change it :/
```rust
            .filter_map(|keychain| match db.get_last_index(*keychain) {
                Ok(li_opt) => li_opt.map(|li| Ok((*keychain, li))),
                Err(err) => Some(Err(err)),
            })
```

src/blockchain/rpc.rs
evanlinjin: Added.

src/blockchain/rpc.rs
evanlinjin: Good catch. Changed.

```rust
        let ext_spks = db.iter_script_pubkeys(Some(KeychainKind::External))?;
        let int_spks = db.iter_script_pubkeys(Some(KeychainKind::Internal))?;

        // This is a hack to see whether atleast one of the keychains comes from a derivable
        // descriptor. We assume that non-derivable descriptors always has a script count of 1.
        let last_count = std::cmp::max(ext_spks.len(), int_spks.len());
        let has_derivable = last_count > 1;

        // If atleast one descriptor is derivable, we need to ensure scriptPubKeys are sufficiently
        // cached.
        if has_derivable && last_count < params.start_script_count {
            let inner_err = MissingCachedScripts {
                last_count,
                missing_count: params.start_script_count - last_count,
            };
            debug!("requesting more spks with: {:?}", inner_err);
            return Err(Error::MissingCachedScripts(inner_err));
        }
```

src/blockchain/rpc.rs
evanlinjin: Great idea. Changed.

src/blockchain/rpc.rs
evanlinjin: It's because we cannot have two mutable references (`self` is already a mutable reference).

The solution I changed to is to clone `self.txs` first. However, I am thinking now `DbState::update_state` should really returns `DbStateUpdates` instead of having `self` mutable? But I am already spending too much time on this PR!!! haha

src/blockchain/rpc.rs
evanlinjin: @danielabrozzoni Yup, let's make this a separate issue haha

src/blockchain/rpc.rs
evanlinjin: Done.

src/blockchain/rpc.rs
evanlinjin: Fixed.

src/blockchain/rpc.rs
evanlinjin: Sorted.

```rust
        // update txs
        self.updated_txs
            .iter()
            .inspect(|&txid| debug!("updating tx: {}", txid))
            .try_for_each(|txid| batch.set_tx(self.txs.get(txid).unwrap()))?;
```

src/blockchain/rpc.rs
evanlinjin: I only filter the deleted txs, in which case I have the `del_txs` variable :)

src/blockchain/rpc.rs
evanlinjin: Changed.

src/blockchain/rpc.rs
evanlinjin: This is my current solution:

```rust
        // update tx sent fields from tx inputs
        self.txs
            .clone() // clone is required as we cannot have more than two mutable references
            .values()
            .try_for_each(|db_tx| -> Result<(), Error> {
                // only bother to update txs that are retained
                if self.retained_txs.get(&db_tx.txid).is_some() {
                    // if we do not have the raw_tx, we cannot calculate the sent value
                    if let Some(raw_tx) = db_tx.transaction.as_ref() {
                        let sent = self.sent_from_raw_tx(raw_tx)?;
                        self.txs
                            .entry(db_tx.txid)
                            .and_modify(|db_tx| db_tx.sent = sent);
                    }
                }
                Ok(())
            })?;
```

src/blockchain/rpc.rs
evanlinjin: Removed!

src/blockchain/rpc.rs
evanlinjin: My solution:

```rust
        // obtain UTXOs from Core wallet
        let core_utxos = client
            .list_unspent(Some(0), None, None, Some(true), None)?
            .into_iter()
            .filter_map(|utxo_entry| {
                let path_result = self
                    .db
                    .get_path_from_script_pubkey(&utxo_entry.script_pub_key)
                    .transpose()?;

                let utxo_result = match path_result {
                    Ok((keychain, index)) => {
                        self.update_last_index(keychain, index);
                        Ok(Self::make_local_utxo(utxo_entry, keychain, false))
                    }
                    Err(err) => Err(err),
                };

                Some(utxo_result)
            })
            .collect::<Result<HashSet<_>, Error>>()?;
```

with:

```rust
    fn make_local_utxo(entry: ListUnspentResultEntry, keychain: KeychainKind, is_spent: bool) -> LocalUtxo {
        LocalUtxo {
            outpoint: OutPoint::new(entry.txid, entry.vout),
            txout: TxOut {
                value: entry.amount.as_sat(),
                script_pubkey: entry.script_pub_key,
            },
            keychain,
            is_spent,
        }
    }
```

src/blockchain/rpc.rs
afilini: In that case I would just delete it, but since it doesn't really affect the balance or anything we don't have to do it in this pr right now

evanlinjin: I feel like it will be a nice feature for the wallet to keep track of replaced transactions as well.

src/blockchain/rpc.rs
danielabrozzoni: Is it? You're still doing:

```
if batch.del_tx(txid, false)?.is_some() {
    debug!("deleting tx: {}", txid);
    del_txs += 1;
}
```

But my point is that `del_tx` will delete a transaction even when returning None...

Ping @afilini for an explanation, I don't remember why is the BatchDatabase doing so

afilini: It's because the batch object is basically disconnected from the actual database, it's just a collection of operations that are then applied at once at the end

evanlinjin: @afilini @danielabrozzoni Thanks for pointing this out! I'll actually fix it now

src/blockchain/rpc.rs
```diff
@@ -495,6 +285,507 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+        // add to updated utxos
+        self.updated_utxos.extend(spent_utxos.chain(new_utxos));
+
+        Ok(self)
+    }
+
+    /// Calculates received amount from raw tx.
+    fn received_from_raw_tx(db: &D, raw_tx: &Transaction) -> Result<u64, Error> {
+        raw_tx.output.iter().try_fold(0_u64, |recv, txo| {
+            let v = if db.is_mine(&txo.script_pubkey)? {
```
afilini, line 540: nit: you can use `self.db` and remove the argument

src/blockchain/rpc.rs
```diff
@@ -495,6 +285,507 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+    fn new(db: &'a D, params: &'a RpcSyncParams, prog: &'a dyn Progress) -> Result<Self, Error> {
+        let ext_spks = db.iter_script_pubkeys(Some(KeychainKind::External))?;
+        let int_spks = db.iter_script_pubkeys(Some(KeychainKind::Internal))?;
+
+        // This is a hack to see whether atleast one of the keychains comes from a derivable
+        // descriptor. We assume that non-derivable descriptors always has a script count of 1.
+        let last_count = std::cmp::max(ext_spks.len(), int_spks.len());
+        let has_derivable = last_count > 1;
+
+        // If atleast one descriptor is derivable, we need to ensure scriptPubKeys are sufficiently
```
afilini: nit: at least

src/blockchain/rpc.rs
```diff
@@ -495,6 +285,507 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+            // update tx deltas
+            self.retained_txs.insert(tx_res.info.txid);
+            if updated {
+                self.updated_txs.insert(tx_res.info.txid);
+            }
+        }
+
+        // update tx sent fields from tx inputs
+        self.txs
+            .clone() // clone is required as we cannot have more than two mutable references
```
afilini: nit: more than one (or alternatively just "two")

src/blockchain/rpc.rs
```diff
@@ -495,6 +285,507 @@ fn list_wallet_dir(client: &Client) -> Result<Vec<String>, Error> {
...
+
+            // update tx deltas
+            self.retained_txs.insert(tx_res.info.txid);
+            if updated {
+                self.updated_txs.insert(tx_res.info.txid);
+            }
+        }
+
+        // update tx sent fields from tx inputs
+        self.txs
```
afilini: I have a couple of suggestions here to improve this a bit:

1. instead of cloning the whole thing you can call `values()` first, so at least you avoid cloning the keys
2. the check that looks for the tx in `retained_txs` could be moved into a `filter()` method, which maybe makes this a bit more readable. Also this check could be moved before the `clone()` to again avoid cloning things you don't need

afilini (commented):
I just have a few very small comment, code looks very good now!

src/blockchain/rpc.rs
danielabrozzoni: Resolving as I opened #700

afilini:
Also if you could rebase to fix the conflicts in changelog.md

src/blockchain/rpc.rs
evanlinjin: English is such an annoying language. Rules do not make sense. Anyway, thank you for pointing this out! haha

src/blockchain/rpc.rs
evanlinjin: Looking at how `received_from_raw_tx` is used, it is used while a mutable reference to a field of `self` still exists.

The suggested change will result in:
```
error[E0502]: cannot borrow `*self` as immutable because it is also borrowed as mutable
```

src/blockchain/rpc.rs
evanlinjin: @afilini Thank you for the suggestions!

However, using `.values().cloned()` would not work, as calling `.cloned()` only clones one element at a time (and we have a mutable reference of the whole `self`).

In my original commit, I had all filters in an `.filter_map` and had two loop (one for marking changes and one for applying changes) so we avoid the mutable-reference problems.

I think it is best to kinda go back to my original solution, but improve commenting and split "filter logic" into multiple `.filter..()` calls.

This is what I have now:

```rust
        // obtain vector of `TransactionDetails::sent` changes
        let sent_updates = self
            .txs
            .values()
            // only bother to update txs that are retained
            .filter(|db_tx| self.retained_txs.contains(&db_tx.txid))
            // only bother to update txs where the raw tx is accessable
            .filter_map(|db_tx| (db_tx.transaction.as_ref().map(|tx| (tx, db_tx.sent))))
            // recalcuate sent value, only update txs in which sent value is changed
            .filter_map(|(raw_tx, old_sent)| {
                self.sent_from_raw_tx(raw_tx)
                    .map(|sent| {
                        if sent != old_sent {
                            Some((raw_tx.txid(), sent))
                        } else {
                            None
                        }
                    })
                    .transpose()
            })
            .collect::<Result<Vec<_>, _>>()?;

        // record send updates
        sent_updates.iter().for_each(|&(txid, sent)| {
            // apply sent field changes
            self.txs.entry(txid).and_modify(|db_tx| db_tx.sent = sent);
            // mark tx as modified
            self.updated_txs.insert(txid);
        });
```

This was my original:

https://github.com/bitcoindevkit/bdk/blob/9d787bf122edf9e5aa6fb43f2315860c1bfb732a/src/blockchain/rpc.rs#L454-L476

evanlinjin: I just had a thought. If we can guarantee that txs returned from `listtransactions` RPC call are in chronological order (which I assume it would be), we can have the `TransactionDetails::sent` recalculations in the same loop as the recalculations of all the other fields.

The only benefit of this is for a slight performance gain, so maybe no point to bother? (I would assume that reducing IO would contribute the most to performance)

afilini: Yeah I remember this now, I think I was the one who suggested merging the two loops :sweat_smile: 

In the end I like what you came up with!

afilini (approved):
ACK 5eeba6cced9a6fa0ad8ee4f64d04e1948620eac8

PR #689, opened 2022-07-29
Add allow_dust method to TxBuilder

terror:
We needed this for testing our wallet with dust, does this look like a reasonable feature? If so, I'll go ahead and add a test, update the changelog, etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add capacity to create FeeRate from sats/kvbytes and sats/kwu.
 - Rename `as_sat_vb` to `as_sat_per_vb`. Move all `FeeRate` test to `types.rs`.
 - Add custom Harware Wallet Signer `HwiSigner` in `src/wallet/harwaresigner/` module.
+- Add `allow_dust` method on `TxBuilder`.
 
 ## [v0.21.0] - [v0.20.0]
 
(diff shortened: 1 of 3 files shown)


danielabrozzoni:
Concept ACK! I think this could be useful for other purposes as well, like creating a transaction with a 1 satoshi notification output to someone

danielabrozzoni:
Concept ACK! Transactions with below-dust ouputs are non-standard, but still valid, so I think BDK should be able to create them. It could also be useful in some very rare cases, like for creating a notification output to an address.

src/wallet/tx_builder.rs
```diff
@@ -560,6 +561,12 @@ impl<'a, D: BatchDatabase, Cs: CoinSelectionAlgorithm<D>, Ctx: TxBuilderContext>
         self.params.current_height = Some(height);
         self
     }
+
+    /// Set whether or not the dust limit is checked.
```
danielabrozzoni, line 565: Maybe we could also explain here that if the dust is allowed you might end up with a transaction that is non standard

danielabrozzoni:
The code looks good to me, can you please squash all the commits into the first one?

danielabrozzoni:
Mh... There are still two commits, and the latest one's message is a bit weird. I'd use interactive git rebases to fix (https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) - I would reword the first commit to change the message, and drop entirely the second commit

terror:
> Mh... There are still two commits, and the latest one's message is a bit weird. I'd use interactive git rebases to fix (https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) - I would reword the first commit to change the message, and drop entirely the second commit

@danielabrozzoni Should be good to go now. PTAL

danielabrozzoni (approved):
utACK 61ad455872ab78df744763c0d69d63c6683f4814

vladimirfomene (commented):
Tested ACK

rajarshimaitra (commented):
Review ACK.. Code looks good to me.. A small test at this stage would be really nice..

danielabrozzoni:
@rajarshimaitra do you have in mind any other additional test? There's already a `test_allow_dust_limit`, I can't think of any other that could be useful

rajarshimaitra (approved):
tACK 61ad455872ab78df744763c0d69d63c6683f4814

Huh.. I have no idea how I missed the test in last pass.. Thanks @danielabrozzoni yup thats all, we don't need anything else..

This looks good to go..

CHANGELOG.md
```diff
@@ -15,14 +15,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Change the meaning of the `fee_amount` field inside `CoinSelectionResult`: from now on the `fee_amount` will represent only the fees asociated with the utxos in the `selected` field of `CoinSelectionResult`.
 - New `RpcBlockchain` implementation with various fixes.
 - Return balance in separate categories, namely `confirmed`, `trusted_pending`, `untrusted_pending` & `immature`.
+- Add `allow_dust` method on `TxBuilder`.
```
afilini: You should rebase and move this under the new unreleased section

danielabrozzoni: Hey @terror, sorry for the ping, but the feature freeze is tomorrow, which means that if you fix this comment by tomorrow we can merge the PR and release `allow_dust` as a part of bdk 0.22! :)

danielabrozzoni (approved):
ACK 257a687344815e6c2610d30d41f8885a0cc2990a

danielabrozzoni:
Oh no, I just noticed... @terror can you sign your commit? Otherwise I can't merge. If you prefer, I can sign it with my key.

danielabrozzoni:
tACK bfd7b2f65debe57edc5d563c071f0bfce592a8d2

PR #692, opened 2022-07-30
Fix derivable-descriptor detection during wallet sync

evanlinjin:
### Description

This fix is for electrum-based blockchains.

Previously, it was assumed that containing > 1 spks in database means the descriptor is derivable, however each database tracks two descriptors so it should really check > 2.

### Notes to the reviewers

This is a touch-up on work done in https://github.com/bitcoindevkit/bdk/pull/672

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/blockchain/script_sync.rs
+++ b/src/blockchain/script_sync.rs
@@ -131,8 +131,9 @@ impl<'a, D: BatchDatabase> ScriptReq<'a, D> {
         let current_gap = self.script_index - last;
 
         // this is a hack to check whether the scripts are coming from a derivable descriptor
-        // we assume for non-derivable descriptors, the initial script count is always 1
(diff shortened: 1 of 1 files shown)


danielabrozzoni (approved):
utACK 9d2bea240d43ac89b13897b9a79fa8965fd1c24b

danielabrozzoni:
Can you add a test to reproduce the bug? I suppose that having one wallet with two non-derivable descriptors should do it

evanlinjin:
@danielabrozzoni Thanks for the tip! I realized I confused myself. :sweat_smile: 

I forgot that `ScriptReq::satisfy` works with "one keychain at a time" so `let is_derivable = self.initial_scripts_needed > 1;` (the original) is correct.

I'll close this PR.

PR #694, opened 2022-07-30
Add assertions in the FeeRate constructor

afilini:
### Description

Disallow negative, NaN, infinite or subnormal fee rate values.

### Notes to the reviewers

This commit is technically an API break because it makes the `FeeRate::from_sat_per_vb` function non-const. I think it's worth it compared to the risk of having completely nonsensical fee rates (that can break the coin selection in interesting ways). 

EDIT: it's also a breaking change because our code can now panic in scenarios where it didn't before. Again, I think it's worth it.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -51,14 +51,34 @@ impl AsRef<[u8]> for KeychainKind {
 pub struct FeeRate(f32);
 
 impl FeeRate {
+    /// Create a new instance checking the value provided
+    ///
+    /// ## Panics
+    ///
+    /// Panics if the value is not [normal](https://doc.rust-lang.org/std/primitive.f32.html#method.is_normal) (except if it's a positive zero) or negative.
+    fn new_checked(value: f32) -> Self {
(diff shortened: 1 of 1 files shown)


danielabrozzoni (approved):
utACK 00e5c1d796598bca1e7722ae6bf85b63c88a4677

src/types.rs
```diff
@@ -51,14 +51,28 @@ impl AsRef<[u8]> for KeychainKind {
 pub struct FeeRate(f32);
 
 impl FeeRate {
+    /// Create a new instance checking the value provided
+    ///
+    /// Panics if the value is not [normal](https://doc.rust-lang.org/std/primitive.f32.html#method.is_normal) (except if it's a positive zero) or negative.
```
kaiwolfram, line 58: Panic descriptions are usually under a `/// # Panics` section

danielabrozzoni:
re-ACK 235011feef8a6faadc08b814e199e5d5ced2f3a0

PR #704, opened 2022-08-06
Fix `rpc::CoreTxIter` logic.

evanlinjin:
### Description

This fixes a bug where `CoreTxIter` attempts to call `listtransactions` immediately after a tx result is filtered (instead of being returned), when in fact, the correct logic will be to pop another tx result.

The new logic also ensures that tx results are returned in chonological order. The test `test_list_transactions` verifies this. We also now ensure that `page_size` is between the range `[0 to 1000]` otherwise an error is returned.

Some needless cloning is removed from `from_config` as well as logging improvements.

### Notes to the reviewers

This is an oversight by me (sorry) for PR #683

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/blockchain/rpc.rs
+++ b/src/blockchain/rpc.rs
@@ -201,27 +201,26 @@ impl ConfigurableBlockchain for RpcBlockchain {
     /// Returns RpcBlockchain backend creating an RPC client to a specific wallet named as the descriptor's checksum
     /// if it's the first time it creates the wallet in the node and upon return is granted the wallet is loaded
     fn from_config(config: &Self::Config) -> Result<Self, Error> {
-        let wallet_name = config.wallet_name.clone();
-        let wallet_url = format!("{}/wallet/{}", config.url, &wallet_name);
-        debug!("connecting to {} auth:{:?}", wallet_url, config.auth);
+        let wallet_url = format!("{}/wallet/{}", config.url, &config.wallet_name);
 
         let client = Client::new(wallet_url.as_str(), config.auth.clone().into())?;
         let rpc_version = client.version()?;
 
-        let loaded_wallets = client.list_wallets()?;
-        if loaded_wallets.contains(&wallet_name) {
-            debug!("wallet already loaded {:?}", wallet_name);
-        } else if list_wallet_dir(&client)?.contains(&wallet_name) {
-            client.load_wallet(&wallet_name)?;
-            debug!("wallet loaded {:?}", wallet_name);
+        info!("connected to '{}' with auth: {:?}", wallet_url, config.auth);
+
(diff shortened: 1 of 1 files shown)


src/blockchain/rpc.rs
```diff
@@ -695,81 +701,54 @@ where
...
+                Some(page_size),
+                Some(page_size * page_index),
+                Some(true),
+            )
+        })
+        // take until returned rpc call is empty or until error
+        // TODO: replace with the following when MSRV is 1.57.0:
+        // `.map_while(|res| res.map(|l| if l.is_empty() { None } else { Some(l) }).transpose())`
+        .take_while(|res| {
+            if got_err || res.is_ok() && res.as_ref().unwrap().is_empty() {
```
afilini: I'm not sure if that works, but maybe you can replace the second condition with something like `matches!(res, Ok(&[]))`.

But if you prefer it this way, that also works for me. Up to you!

afilini:
Can you come up with an example that would break without this PR (i.e. without iterating in chronological order)? You don't have to write it in code right now, just describe it. We can add it later to the blockchain tests so that we can cover more "edge-cases".

evanlinjin:
@afilini the "chronological order" is just a good to have.

The real fix is this:

> There was also a bug where CoreTxIter will attempt to call listtransactions again when a tx result is filtered (the correct logic would be to pop another tx result).

Sorry I should have made it more clear.

It just so happened that the test I wrote for the iterator didn't work initially (and I didn't know why) but then I realized that with different page sizes results were returned in different orders so I decided to change the logic to make things more predictable.

evanlinjin:
@afilini I just updated the PR description and commit message to better describe the intention.

src/blockchain/rpc.rs
evanlinjin: My solution:
```rust
matches!(res, Ok(list) if list.is_empty())
```

evanlinjin:
@afilini Right now the `RpcBlockchain` sync logic assumes that transactions returned by `CoreTxIter` may not in chronological order (this is why we need to calculate `TransactionDetails::received` in a separate loop. In the future we can change this to slightly improve performance (because we are using the new `list_transactions` function).

The chronological nature of transactions returned is more of an "implementation detail" for determining field values of `TransactionDetails` rather than a potential bug. As from my understanding, `Database::iter_txs` does not require transactions to be in chronological order.

afilini:
Ah ok, that's fine.

The other bug was clear to me, instead of an `if let` it should have been a `while let` to iterate until you find a tx that is not filtered. I couldn't figure out why the chronological order was needed, thanks for the explanation!

afilini (approved):
ACK 74e2c477f124489a2357921ca879cc82e24da5fd

PR #713, opened 2022-08-12
Add datatype for is_spent sqlite column

vladimirfomene:
### Description

During table creation, Sqlite does not throw an error when a column datatype is not defined. In addition, the datatype provided during table creation does not put a constraint on the type of data that can be put in that column. So you can easily put a string value in an integer column. Despite this, I think it is important for us to add the datatype for clarity.

### Notes to the reviewers

You can read more about how Sqlite dynamic typing [here](https://www.sqlite.org/faq.html). I have amended our `migrate` code with a new commit. The idea is to run migrations in a transaction so that they either succeed or fail. This prevents us from having the database in an inconsistent state at any point in time.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/database/sqlite.rs
+++ b/src/database/sqlite.rs
@@ -52,7 +52,14 @@ static MIGRATIONS: &[&str] = &[
     "DELETE FROM transactions;",
     "DELETE FROM utxos;",
     "DROP INDEX idx_txid_vout;",
-    "CREATE UNIQUE INDEX idx_utxos_txid_vout ON utxos(txid, vout);"
+    "CREATE UNIQUE INDEX idx_utxos_txid_vout ON utxos(txid, vout);",
+    "BEGIN TRANSACTION;\
+    ALTER TABLE utxos RENAME TO utxos_old;\
+    CREATE TABLE utxos (value INTEGER, keychain TEXT, vout INTEGER, txid BLOB, script BLOB, is_spent BOOLEAN DEFAULT 0);\
+    INSERT INTO utxos SELECT value, keychain, vout, txid, script, is_spent FROM utxos_old;\
+    DROP TABLE utxos_old;\
+    CREATE UNIQUE INDEX idx_utxos_txid_vout ON utxos(txid, vout);\
+    COMMIT;"
 ];
 
 /// Sqlite database stored on filesystem
--- a/src/database/sqlite.rs
+++ b/src/database/sqlite.rs
@@ -53,13 +53,11 @@ static MIGRATIONS: &[&str] = &[
     "DELETE FROM utxos;",
     "DROP INDEX idx_txid_vout;",
     "CREATE UNIQUE INDEX idx_utxos_txid_vout ON utxos(txid, vout);",
-    "BEGIN TRANSACTION;\
-    ALTER TABLE utxos RENAME TO utxos_old;\
-    CREATE TABLE utxos (value INTEGER, keychain TEXT, vout INTEGER, txid BLOB, script BLOB, is_spent BOOLEAN DEFAULT 0);\
-    INSERT INTO utxos SELECT value, keychain, vout, txid, script, is_spent FROM utxos_old;\
-    DROP TABLE utxos_old;\
-    CREATE UNIQUE INDEX idx_utxos_txid_vout ON utxos(txid, vout);\
-    COMMIT;"
+    "ALTER TABLE utxos RENAME TO utxos_old;",
+    "CREATE TABLE utxos (value INTEGER, keychain TEXT, vout INTEGER, txid BLOB, script BLOB, is_spent BOOLEAN DEFAULT 0);",
+    "INSERT INTO utxos SELECT value, keychain, vout, txid, script, is_spent FROM utxos_old;",
+    "DROP TABLE utxos_old;",
+    "CREATE UNIQUE INDEX idx_utxos_txid_vout ON utxos(txid, vout);"
 ];
 
 /// Sqlite database stored on filesystem
@@ -921,8 +919,8 @@ impl BatchDatabase for SqliteDatabase {
 }
 
 pub fn get_connection<T: AsRef<Path>>(path: &T) -> Result<Connection, Error> {
-    let connection = Connection::open(path)?;
-    migrate(&connection)?;
+    let mut connection = Connection::open(path)?;
+    migrate(&mut connection)?;
     Ok(connection)
 }
 
@@ -957,28 +955,41 @@ pub fn set_schema_version(conn: &Connection, version: i32) -> rusqlite::Result<u
     )
 }
 
-pub fn migrate(conn: &Connection) -> rusqlite::Result<()> {
+pub fn migrate(conn: &mut Connection) -> Result<(), Error> {
     let version = get_schema_version(conn)?;
     let stmts = &MIGRATIONS[(version as usize)..];
-    let mut i: i32 = version;
 
-    if version == MIGRATIONS.len() as i32 {
+    // begin transaction, all migration statements and new schema version commit or rollback
+    let tx = conn.transaction()?;
+
+    // execute every statement and return `Some` new schema version
+    // if execution fails, return `Error::Rusqlite`
+    // if no statements executed returns `None`
+    let new_version = stmts
+        .iter()
+        .enumerate()
+        .map(|version_stmt| {
+            log::info!(
+                "executing db migration {}: `{}`",
+                version + version_stmt.0 as i32 + 1,
+                version_stmt.1
+            );
+            tx.execute(version_stmt.1, [])
+                // map result value to next migration version
+                .map(|_| version_stmt.0 as i32 + version + 1)
+        })
+        .last()
+        .transpose()?;
+
+    // if `Some` new statement version, set new schema version
+    if let Some(version) = new_version {
+        set_schema_version(&tx, version)?;
+    } else {
         log::info!("db up to date, no migration needed");
-        return Ok(());
     }
 
-    for stmt in stmts {
-        let res = conn.execute(stmt, []);
-        if res.is_err() {
-            println!("migration failed on:\n{}\n{:?}", stmt, res);
-            break;
-        }
-
-        i += 1;
-    }
-
-    set_schema_version(conn, i)?;
-
+    // commit transaction
+    tx.commit()?;
     Ok(())
 }
 


src/database/sqlite.rs
```diff
@@ -42,7 +42,7 @@ static MIGRATIONS: &[&str] = &[
     "CREATE TABLE transaction_details (txid BLOB, timestamp INTEGER, received INTEGER, sent INTEGER, fee INTEGER, height INTEGER);",
     "INSERT INTO transaction_details SELECT txid, timestamp, received, sent, fee, height FROM transaction_details_old;",
     "DROP TABLE transaction_details_old;",
-    "ALTER TABLE utxos ADD COLUMN is_spent;",
+    "ALTER TABLE utxos ADD COLUMN is_spent BOOLEAN CHECK (is_spent IN (0, 1));",
```
danielabrozzoni: The problem with this approach is that databases that already did this migration won't be changed (see the migration code: https://github.com/bitcoindevkit/bdk/blob/master/src/database/sqlite.rs#L953-L976)

Maybe it's better to add YA migration altering the utxo table? Idk if it's possible to change the column type

afilini: Yes it's probably better to add this as an extra migration

vladimirfomene: @danielabrozzoni @afilini, it is not possible to modify an existing column datatype in sqlite. So using something like `ALTER COLUMN` will not work here. Another option will be to rename the table to `table_name_old` and copy all its content to the new table. Here is an example from the migration code:
```sql
CREATE TABLE transaction_details (txid BLOB, timestamp INTEGER, received INTEGER, sent INTEGER, fee INTEGER, height INTEGER, verified INTEGER DEFAULT 0);
ALTER TABLE transaction_details RENAME TO transaction_details_old;
CREATE TABLE transaction_details (txid BLOB, timestamp INTEGER, received INTEGER, sent INTEGER, fee INTEGER, height INTEGER);
INSERT INTO transaction_details SELECT txid, timestamp, received, sent, fee, height FROM transaction_details_old;
DROP TABLE transaction_details_old;
```
    
   This is a lot of gymnastics for a simple datatype change. Another possibility will be to allow the current implementation. What do you guys think?

danielabrozzoni: I'd honestly go with renaming the old table and moving everything to a new one... It is a bit of gymnastic, but it's important that our type remain strict for everyone, not only for new users of bdk

vladimirfomene: @danielabrozzoni , updated following your feedback.

src/database/sqlite.rs
```diff
@@ -43,6 +43,10 @@ static MIGRATIONS: &[&str] = &[
     "INSERT INTO transaction_details SELECT txid, timestamp, received, sent, fee, height FROM transaction_details_old;",
     "DROP TABLE transaction_details_old;",
     "ALTER TABLE utxos ADD COLUMN is_spent;",
+    "ALTER TABLE utxos RENAME TO utxos_old;",
+    "CREATE TABLE utxos (value INTEGER, keychain TEXT, vout INTEGER, txid BLOB, script BLOB, is_spent BOOLEAN DEFAULT 0);",
+    "INSERT INTO utxos SELECT value, keychain, vout, txid, script, is_spent FROM utxos_old;",
+    "DROP TABLE utxos_old;",
```
danielabrozzoni: This should be at the end of the migrations. 

See: https://github.com/bitcoindevkit/bdk/blob/master/src/database/sqlite.rs#L953-L975

We apply all the migrations we have, and then we save the index of the migration array of the last migration we applied. For this reason, the migration array must be append-only: otherwise clients when migrating might re-do steps they've already done!

If we kept the code like that, a client which is in master at the latest migration (`"CREATE UNIQUE INDEX idx_utxos_txid_vout ON utxos(txid, vout);"`) would re-do the last 4 migrations of the array (from `"DELETE FROM transaction_details;"` to `"CREATE UNIQUE INDEX idx_utxos_txid_vout ON utxos(txid, vout);"`), instead of applying the migrations you just added. Does that make sense?

For the same reason, you shouldn't delete the `"DROP INDEX idx_txid_vout;"` line.

danielabrozzoni (changes requested):

src/database/sqlite.rs
vladimirfomene: Thanks @danielabrozzoni! That makes sense, sorry I missed it the first time.

danielabrozzoni (approved):
tACK 0c9dabed69b4edd8e389a7ccaf65b66635158937 - I created an example on master creating a wallet with sqlite, I manually checked the`utxos` table with the `sqlite3` command, and `is_spent` didn't have a type. I then switched to this branch, ran the example again, inspected the table and found the same exact data, but the `is_spent` column's type was correct.

In case someone else wants to manually try this out, this is the code I used:
```
use bdk::blockchain::ElectrumBlockchain;
use bdk::database::SqliteDatabase;
use bdk::electrum_client::Client;
use bdk::{SyncOptions, Wallet};

fn main() -> Result<(), bdk::Error> {
    let client = Client::new("ssl://electrum.blockstream.info:60002")?;
    let blockchain = ElectrumBlockchain::from(client);
    let wallet = Wallet::new(
        "wpkh([c258d2e4/84h/1h/0h]tpubDDYkZojQFQjht8Tm4jsS3iuEmKjTiEGjG6KnuFNKKJb5A6ZUCUZKdvLdSDWofKi4ToRCwb9poe1XdqfUnP4jaJjCB2Zwv11ZLgSbnZSNecE/0/*)",
        Some("wpkh([c258d2e4/84h/1h/0h]tpubDDYkZojQFQjht8Tm4jsS3iuEmKjTiEGjG6KnuFNKKJb5A6ZUCUZKdvLdSDWofKi4ToRCwb9poe1XdqfUnP4jaJjCB2Zwv11ZLgSbnZSNecE/1/*)"),
        bitcoin::Network::Testnet,
        SqliteDatabase::new("/tmp/bdk.sqlite"),
    )?;

    wallet.sync(&blockchain, SyncOptions::default())?;

    println!("Descriptor balance: {} SAT", wallet.get_balance()?);

    Ok(())
}
```

src/database/sqlite.rs
```diff
@@ -52,6 +52,11 @@ static MIGRATIONS: &[&str] = &[
     "DELETE FROM transactions;",
     "DELETE FROM utxos;",
     "DROP INDEX idx_txid_vout;",
+    "CREATE UNIQUE INDEX idx_utxos_txid_vout ON utxos(txid, vout);",
+    "ALTER TABLE utxos RENAME TO utxos_old;",
+    "CREATE TABLE utxos (value INTEGER, keychain TEXT, vout INTEGER, txid BLOB, script BLOB, is_spent BOOLEAN DEFAULT 0);",
+    "INSERT INTO utxos SELECT value, keychain, vout, txid, script, is_spent FROM utxos_old;",
+    "DROP TABLE utxos_old;",
```
afilini, line 59: To avoid potentially leaving the database in a semi-inconsistent state I would pack all these commands in a single string, so that it counts as a single migration. Plus i would make it into a transaction with `BEGIN/COMMIT` (see: https://www.sqlite.org/lang_transaction.html) so that either it completes fully or fails.

notmandatory:
Hey I forgot to discuss this one in our call today. Since it needs a little more work I'm moving it to the 0.23 milestone.

danielabrozzoni:
> Since it needs a little more work I'm moving it to the 0.23 milestone.

Well if Vlad can update this one in the next ~10 hours, we can squeeze it in, otherwise no problem with waiting for the next release :)

vladimirfomene:
@afilini @danielabrozzoni I have added a new commit to our migration code which runs the migration statement in a transaction in such a way that they either succeed or fail. Let me know what you think about it.

src/database/sqlite.rs
```diff
@@ -950,7 +955,7 @@ pub fn set_schema_version(conn: &Connection, version: i32) -> rusqlite::Result<u
     )
 }
 
-pub fn migrate(conn: &Connection) -> rusqlite::Result<()> {
+pub fn migrate(conn: &mut Connection) -> rusqlite::Result<()> {
```
notmandatory: The new transaction is a good improvement, but I'd also simplify it by including the version update in the same transaction. Here's my suggestions for how to do it, I also threw in some "functional style", and removed unneeded functions (only called by migrate), let me know what you think:
```rust
pub fn migrate(conn: &mut Connection) -> Result<(), Error> {
    let version = get_schema_version(conn)?;
    let stmts = &MIGRATIONS[(version as usize)..];

    // begin transaction, all migration statements and new schema version commit or rollback
    let tx = conn.transaction()?;

    // execute every statement and return `Some` new version
    // if execution fails, return `Error::Rusqlite`
    // if no statements executed returns `None`
    let new_version = stmts
        .iter()
        .enumerate()
        .map(|version_stmt| {
            log::info!(
                "executing db migration {}: `{}`",
                version_stmt.0,
                version_stmt.1
            );
            tx.execute(version_stmt.1, [])
                // map result value to next migration version
                .map(|_| version_stmt.0 as i32 + 1)
        })
        .last()
        .transpose()?;

    // if `Some` new statement version set new schema version
    if let Some(version) = new_version {
        tx.execute(
            "UPDATE version SET version=:version",
            named_params! {":version": version},
        )?;
    } else {
        log::info!("db up to date, no migration needed");
    }
    
    // commit transaction
    tx.commit()?;
    Ok(())
}
```

vladimirfomene: This definitely looks cleaner and more idiomatic. Thanks, I will test it out.

vladimirfomene:
@notmandatory, I have updated the change as recommended.

src/database/sqlite.rs
```diff
@@ -950,28 +955,41 @@ pub fn set_schema_version(conn: &Connection, version: i32) -> rusqlite::Result<u
...
+    // if no statements executed returns `None`
+    let new_version = stmts
+        .iter()
+        .enumerate()
+        .map(|version_stmt| {
+            log::info!(
+                "executing db migration {}: `{}`",
+                version_stmt.0,
+                version_stmt.1
+            );
```
rajarshimaitra: If the intent of this log is to show migration steps with index number, currently its showing below for a db upgrade..
```
executing db migration 0: `ALTER TABLE utxos RENAME TO utxos_old;`
executing db migration 1: `CREATE TABLE utxos (value INTEGER, keychain TEXT, vout INTEGER, txid BLOB, script BLOB, is_spent BOOLEAN DEFAULT 0);`
executing db migration 2: `INSERT INTO utxos SELECT value, keychain, vout, txid, script, is_spent FROM utxos_old;`
executing db migration 3: `DROP TABLE utxos_old;`
executing db migration 4: `CREATE UNIQUE INDEX idx_utxos_txid_vout ON utxos(txid, vout);`
```
The indexing starts at 0 for a migration operation.  
We might wanna index the migration with the actual line number of the full migration array? In that way it will be easier to debug later at which migration step problem occurred..

src/database/sqlite.rs
```diff
@@ -950,28 +955,41 @@ pub fn set_schema_version(conn: &Connection, version: i32) -> rusqlite::Result<u
...
+            tx.execute(version_stmt.1, [])
+                // map result value to next migration version
+                .map(|_| version_stmt.0 as i32 + 1)
```
rajarshimaitra: I did not understand the purpose of this line..

rajarshimaitra (commented):
tACK cd3e33e9a03a1f29eb15032db425e72be72156c5

Verified that `is_spent` datatype indeed changes and extra migration is added over an existing database successfully.. 

This is good to go.. Just one non blocking observation.

src/database/sqlite.rs
afilini: +1

src/database/sqlite.rs
vladimirfomene: Great idea! I will implement it.

src/database/sqlite.rs
vladimirfomene: The goal was to increment the database version after every statement runs but having had a second look at it, I think it should have been `map(|_| version_stmt.0 as i32 + version + 1)`

src/database/sqlite.rs
vladimirfomene: Done

src/database/sqlite.rs
vladimirfomene: I have updated this line to `map(|_| version_stmt.0 as i32 + version + 1)`.

rajarshimaitra (approved):
tACK 54d768412abf218d2646aba211bb79b9dd287118

Verified that the log is matching the migration statement index. Db schema version is being set correctly after migration..

afilini (approved):
ACK 54d768412abf218d2646aba211bb79b9dd287118

PR #717, opened 2022-08-14
[bug-fix] Set the sync time in populate test db.

rajarshimaitra:
### Description

After recent changes in balance calculation, the `get_funded_wallet` test function will not show any balance because we are not setting the `sync_time` in the `populate_test_db` macro.. 

Our existing tests are not failing, because even if the balance is shown zero, it will still be able to produce transactions. But its causing test failure in downstream `bdk-reserves` crate where the balances are required to assert scenarios. 

### Notes to the reviewers

This will be replayed in #624 too, as this function is being refactored there.. For now opening it as separate PR for quick merge.  

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/database/memory.rs
+++ b/src/database/memory.rs
@@ -520,6 +520,13 @@ macro_rules! populate_test_db {
                 timestamp: 0,
             });
 
+        let sync_time = BlockTime {
+            height: current_height.unwrap(),
+            timestamp: 0
+        };
+
+        db.set_sync_time(SyncTime{block_time: sync_time}).unwrap();
+
         let tx_details = $crate::TransactionDetails {
             transaction: Some(tx.clone()),
             txid,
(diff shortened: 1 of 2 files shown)


rajarshimaitra:
This is becoming a more intrusive change than I previously thought.. There are test cases thats working on the assumption that sync height is never set, even when `populate_test_db` is taking in current height as `Some(100)`.. Also the current `populate_test_db` cannot handle `None` values without further changes.. 

I am trying next to replay the fix on top of #624.. I think that will reduce the review effort, as all these functions are refactored there anyway.. And its not a major bug, so can wait for sometime, if the fix isn't just few lines anymore.. 

More suggestions welcome in the mean time.. 

Idea: The other way of fixing this bug is, do not check for sync time at `get_balance()`, just give a balance back if there's unspent outputs, which I think what was the case pre v0.21.

rajarshimaitra:
Closing this in favor of #624  where the full fix is included..

PR #718, opened 2022-08-15
Verify signatures after signing

quad:
### Description

Verify signatures after signing

As per [BIP-340, footnote 14][fn]:
> Verifying the signature before leaving the signer prevents random or
> attacker provoked computation errors. This prevents publishing invalid
> signatures which may leak information about the secret key. It is
> recommended, but can be omitted if the computation cost is prohibitive.

[fn]: https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#cite_note-14

### Notes to the reviewers

How do we test this?

#### All Submissions:

### DIFF
--- a/src/wallet/signer.rs
+++ b/src/wallet/signer.rs
@@ -475,10 +475,10 @@ fn sign_psbt_ecdsa(
     hash_ty: EcdsaSighashType,
     secp: &SecpCtx,
 ) {
-    let sig = secp.sign_ecdsa(
-        &Message::from_slice(&hash.into_inner()[..]).unwrap(),
-        secret_key,
-    );
+    let msg = &Message::from_slice(&hash.into_inner()[..]).unwrap();
+    let sig = secp.sign_ecdsa(msg, secret_key);
+    secp.verify_ecdsa(msg, &sig, &pubkey.inner)
+        .expect("invalid or corrupted ecdsa signature");
 
     let final_signature = ecdsa::EcdsaSig { sig, hash_ty };
     psbt_input.partial_sigs.insert(pubkey, final_signature);
@@ -504,10 +504,10 @@ fn sign_psbt_schnorr(
         Some(_) => keypair, // no tweak for script spend
     };
 
-    let sig = secp.sign_schnorr(
-        &Message::from_slice(&hash.into_inner()[..]).unwrap(),
-        &keypair,
-    );
+    let msg = &Message::from_slice(&hash.into_inner()[..]).unwrap();
+    let sig = secp.sign_schnorr(msg, &keypair);
+    secp.verify_schnorr(&sig, msg, &XOnlyPublicKey::from_keypair(&keypair))
+        .expect("invalid or corrupted schnorr signature");
 
     let final_signature = schnorr::SchnorrSig { sig, hash_ty };
 


src/wallet/signer.rs
```diff
@@ -504,10 +504,10 @@ fn sign_psbt_schnorr(
         Some(_) => keypair, // no tweak for script spend
     };
 
-    let sig = secp.sign_schnorr(
-        &Message::from_slice(&hash.into_inner()[..]).unwrap(),
-        &keypair,
-    );
+    let msg = &Message::from_slice(&hash.into_inner()[..]).unwrap();
+    let sig = secp.sign_schnorr(msg, &keypair);
+    secp.verify_schnorr(&sig, msg, &XOnlyPublicKey::from_keypair(&keypair))
```
quad, line 509: Confusingly, `secp.verify_schnorr(&sig, msg, &pubkey)` failed the unit tests.

But, [below on line 517](https://github.com/bitcoindevkit/bdk/pull/718/files#diff-78369b4827c629390df1a56a9ec4cae57346aa2f03cb66bcf9455cf75411923cL517), we attach the resulting signature to `pubkey`.

That seems wrong. Is that a bug?

danielabrozzoni:
Since the footnote says "prevents **random** or attacker provoked computation errors", maybe we should re-try if the verification fails, and after one retry, return an error instead of panicking?

quad:
I wouldn’t retry.

The effect of a random computation error likely isn’t limited to just the internals of the signature call. Either the CPU or more likely the RAM is in an invalid state. Failing fast is the correct behaviour as, regardless, the caller needs to be able to handle failure. See: https://docs.kernel.org/admin-guide/ras.html

In the case of an attacker provoked error, re-running leaks more information in a tighter loop. Halting on detection of glitch attacks is the norm. And many classes of side channel attacks are specifically mitigated by slowing down the loop. See: https://www.blackhat.com/docs/eu-15/materials/eu-15-Giller-Implementing-Electrical-Glitching-Attacks.pdf

src/wallet/signer.rs
afilini: When doing a key-spend the key is tweaked with the merkle root (line 501-503), so the public key of the `keypair` variable is gonna be different from the "original" pubkey.

That said, key-spend signatures are stored in the `tap_key_sig` psbt field, which is not a map and thus doesn't contain the pubkey at all (unlike `tap_script_sigs`)

jmecom:
IMO, if a fault occurs, the caller should be made aware and the library should not silently retry until success occurs. The correct response to a fault detection may vary from system-to-system, and so the library should be flexible. I'll quote the [Hardware Hacking Handbook](https://nostarch.com/hardwarehacking) on some possible fault responses:

```A fault response is what to do when a fault is detected. The goal here is to reduce the chances of a successful attack to the point where an attacker will give up. On the one end of the spectrum, you can implement a program exit, OS reboot, or chip reset. These actions will delay an attacker but in principle still allow them infinite tries. Somewhere in the middle of the spectrum is signaling a backend system to flag this device as suspicious and perhaps disable the account. On the other end of the spectrum, you can implement permanent measures like wiping keys, accounts, or even burning fuses that disallow the chip from booting up.```

danielabrozzoni:
Yeah, you're right, re-trying doesn't seem to be a good idea, sorry :)

afilini (approved):
ACK 31eed5f78cccb4e31f0bb1ce86ebee4a631b5807

afilini:
Wait, I can't merge this unless the commits are signed: @quad you can either sign them yourself, or if you prefer I can sign them with my own key, let me know.

quad:
Feel free to sign it with your key.

quad:
@afilini I've signed the commit. Would you please re-approve running workflows?

afilini (approved):
re-ACK 7b1ad1b62914a26d6f445364ace4e784bb2901c2

Issue #573, opened 2022-03-18
Remove old `LocalUtxo`s


danielabrozzoni:
After #515 we never delete TxOuts from the database; this can obviously be a problem for wallets that do many transactions and applications that run on devices with limited memory. We should add a way of manually deleting the spent txouts.

xraid:
maybe choose archive or delete
auto archive (set time interval - size count) - then manually delete

more of a concern for wallet implementation choice of db management ?

maybe a API of encrypted remote store in own "cloud" or other type bucket

vladimirfomene:
@danielabrozzoni , can I pick this up?

danielabrozzoni:
I just assigned this to you 😁 

So, I don't know exactly how this will look like, but to give a bit of context: in #414 and #515 there should be enough discussion to cover why we went for never deleting the utxos in the first place. Now, we should find some way to do so, and IMHO it should be pretty configurable (as everything in BDK). We could have an automatic mode ("delete all the utxos which have been spent at least 10 blocks ago", "keep at most 500 utxos"), but also a manual API would be cool ("delete all the utxos" or "delete this specific utxo").

We haven't discussed much about it, so feel free to jot down your thoughts on discord or, even better, create a PR with a POC and let people discuss there. Thanks 🙏🏻

notmandatory:
Closing since persistence is done differently in `bdk_wallet` 1.0.

PR #725, opened 2022-08-17
Add API to delete spent UTXOs from database

vladimirfomene:
### Description

We currently store spent UTXOs in the database with an `is_spent` field and we don't provide users with a way to delete these UTXOs. The issue here is that these could easily bloat the database or memory. This PR provides methods that can allow users to delete spent UTXOs from the database following certain criteria like size of spent utxos and number of confirmations.
This PR fixes issue #573

### Notes to the reviewers

I haven't added this change to the change log and written test for it. I'm thinking we can use this PR as a starting point to discuss how best we can solve this issue. I think having an API that users can call manually is a great start towards solving this issue but it  might also be necessary to implement an automatic version of this. 

#### All Submissions:

#### New Features:

### DIFF
--- a/src/database/keyvalue.rs
+++ b/src/database/keyvalue.rs
@@ -532,4 +532,9 @@ mod test {
     fn test_check_descriptor_checksum() {
         crate::database::test::test_check_descriptor_checksum(get_tree());
     }
+
+    #[test]
+    fn test_del_spent_utxos() {
+        crate::database::test::test_del_spent_utxos(get_tree());
+    }
 }
--- a/src/database/memory.rs
+++ b/src/database/memory.rs
@@ -687,4 +687,9 @@ mod test {
     fn test_check_descriptor_checksum() {
         crate::database::test::test_check_descriptor_checksum(get_tree());
     }
+
+    #[test]
+    fn test_del_spent_utxos() {
+        crate::database::test::test_del_spent_utxos(get_tree());
+    }
 }
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -158,6 +158,33 @@ pub trait Database: BatchOperations {
     ///
     /// It should insert and return `0` if not present in the database
     fn increment_last_index(&mut self, keychain: KeychainKind) -> Result<u32, Error>;
+
+    /// Delete a list of spent utxos from database. Delete all spent utxos if  list is `None`.
+    fn del_spent_utxos(
+        &mut self,
+        to_delete: Option<Vec<OutPoint>>,
+    ) -> Result<Vec<LocalUtxo>, Error> {
+        if let Some(to_delete) = to_delete {
+            let to_delete = to_delete
+                .iter()
+                .filter_map(|out| self.get_utxo(out).transpose())
+                .collect::<Result<Vec<_>, _>>()?;
+            let deleted_utxos = to_delete
+                .iter()
+                .filter(|utxo| utxo.is_spent)
+                .filter_map(|out| self.del_utxo(&out.outpoint).transpose())
+                .collect::<Result<Vec<_>, _>>()?;
+            Ok(deleted_utxos)
+        } else {
+            let deleted_utxos = self
+                .iter_utxos()?
+                .iter()
+                .filter(|utxo| utxo.is_spent)
+                .filter_map(|out| self.del_utxo(&out.outpoint).transpose())
+                .collect::<Result<Vec<_>, _>>()?;
+            Ok(deleted_utxos)
+        }
+    }
 }
 
 /// Trait for a database that supports batch operations
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -680,5 +680,135 @@ pub mod test {
         assert!(res.is_err());
     }
 
+    pub fn test_del_spent_utxos<D: Database>(mut db: D) {
+        // Query database for utxos, to prove that it is empty
+        assert_eq!(db.iter_utxos().unwrap().len(), 0);
+
+        let (first_utxo, _second_utxo, third_utxo, fourth_utxo) =
+            setup_del_spent_utxo_test(&mut db);
+        // test that insertion was successful
+        assert_eq!(db.iter_utxos().unwrap().len(), 4);
+
+        // call del_spent_utxos with None
+        let mut res = db.del_spent_utxos(None).unwrap();
+        res.sort_by(|a, b| a.txout.value.cmp(&b.txout.value));
+        // verify that the spent ones has been deleted
+        assert_eq!(
+            res,
+            vec![first_utxo.clone(), third_utxo.clone(), fourth_utxo.clone()]
+        );
+        assert_eq!(db.iter_utxos().unwrap().len(), 1);
+        // re-insert the deleted utxo into database
+        db.set_utxo(&first_utxo).unwrap();
+        db.set_utxo(&third_utxo).unwrap();
+        db.set_utxo(&fourth_utxo).unwrap();
+
+        // call del_spent_utxos with vector of utxos
+        let mut res = db
+            .del_spent_utxos(Some(vec![first_utxo.outpoint, third_utxo.outpoint]))
+            .unwrap();
+        res.sort_by(|a, b| a.txout.value.cmp(&b.txout.value));
+        assert_eq!(res, vec![first_utxo.clone(), third_utxo.clone()]);
+        let utxos = db.iter_utxos().unwrap();
+        assert_eq!(utxos.len(), 2);
+        assert!(utxos.contains(&fourth_utxo));
+    }
+
+    fn setup_del_spent_utxo_test<D: Database>(
+        db: &mut D,
+    ) -> (LocalUtxo, LocalUtxo, LocalUtxo, LocalUtxo) {
+        // insert four utxos into database
+        let first_outpoint = OutPoint::from_str(
+            "c1b4e695098210a31fe02abffe9005cffc051bbe86ff33e173155bcbdc5821e3:0",
+        )
+        .unwrap();
+        let first_script = Script::from(
+            Vec::<u8>::from_hex("76a914db4d1141d0048b1ed15839d0b7a4c488cd368b0e88ac").unwrap(),
+        );
+        let first_txout = TxOut {
+            value: 133742,
+            script_pubkey: first_script,
+        };
+        let first_utxo = LocalUtxo {
+            txout: first_txout,
+            outpoint: first_outpoint,
+            keychain: KeychainKind::External,
+            is_spent: true,
+        };
+
+        db.set_utxo(&first_utxo).unwrap();
+
+        let second_outpoint = OutPoint::from_str(
+            "fc9e4f9c334d55c1dc535bd691a1c159b0f7314c54745522257a905e18a56779:1",
+        )
+        .unwrap();
+
+        let second_script = Script::from(
+            Vec::<u8>::from_hex("76a914824d8a679134215d6d21d25bde3cc63f89ec92eb88ac").unwrap(),
+        );
+
+        let second_txout = TxOut {
+            value: 2257563,
+            script_pubkey: second_script,
+        };
+
+        let second_utxo = LocalUtxo {
+            txout: second_txout,
+            outpoint: second_outpoint,
+            keychain: KeychainKind::External,
+            is_spent: false,
+        };
+
+        db.set_utxo(&second_utxo).unwrap();
+
+        let third_outpoint = OutPoint::from_str(
+            "26e14e606105b250e64849cc14a484ce58f0f7f7064e662862001661b726427b:1",
+        )
+        .unwrap();
+
+        let third_script = Script::from(
+            Vec::<u8>::from_hex("76a9140d5a9a6f7aae31ebb3b72bbfd05935de5765e0e688ac").unwrap(),
+        );
+
+        let third_txout = TxOut {
+            value: 1119096,
+            script_pubkey: third_script,
+        };
+
+        let third_utxo = LocalUtxo {
+            txout: third_txout,
+            outpoint: third_outpoint,
+            keychain: KeychainKind::External,
+            is_spent: true,
+        };
+
+        db.set_utxo(&third_utxo).unwrap();
+
+        let fourth_outpoint = OutPoint::from_str(
+            "d27cff02d45817a11ac01d0d93a2e439f2d643b5d8bc57f4f7f56aa571104e8c:0",
+        )
+        .unwrap();
+
+        let fourth_script = Script::from(
+            Vec::<u8>::from_hex("76a914899490496bfd1228e7ad5b0e156f1fc50a8f6f7688ac").unwrap(),
+        );
+
+        let fourth_txout = TxOut {
+            value: 36662433,
+            script_pubkey: fourth_script,
+        };
+
+        let fourth_utxo = LocalUtxo {
+            txout: fourth_txout,
+            outpoint: fourth_outpoint,
+            keychain: KeychainKind::External,
+            is_spent: true,
+        };
+
+        db.set_utxo(&fourth_utxo).unwrap();
+
+        (first_utxo, second_utxo, third_utxo, fourth_utxo)
+    }
+
     // TODO: more tests...
 }
--- a/src/database/sqlite.rs
+++ b/src/database/sqlite.rs
@@ -1096,4 +1096,9 @@ pub mod test {
     fn test_check_descriptor_checksum() {
         crate::database::test::test_check_descriptor_checksum(get_database());
     }
+
+    #[test]
+    fn test_del_spent_utxos() {
+        crate::database::test::test_del_spent_utxos(get_database());
+    }
 }


src/database/keyvalue.rs
```diff
@@ -200,6 +200,29 @@ impl BatchOperations for Batch {
     impl_batch_operations!({}, process_delete_batch);
 }
 
+trait Helpers {
+    fn delete_spent_utxo_set(
+        &mut self,
+        deleted_utxos: &mut Vec<LocalUtxo>,
+        to_del: &[LocalUtxo],
+    ) -> Result<(), Error>;
+}
```
rajarshimaitra: This seems redundant to me. Why have another Trait when the actual delete call is just `self.del_utxo()`?

src/database/memory.rs
```diff
@@ -126,6 +128,45 @@ impl MemoryDatabase {
...
+                }
+            })
+            .collect()
+    }
+
+    fn delete_spent_utxo_set(
+        &mut self,
+        deleted_utxos: &mut Vec<LocalUtxo>,
+        to_del: &Vec<LocalUtxo>,
+    ) -> Result<(), Error> {
```
rajarshimaitra: Instead of passing in a mutable list to collect deleted utxos why not just return them in the function call?? Thats the final API anyway. And that will allow you to have your downstream calls more ergonomic.

src/database/memory.rs
```diff
@@ -449,6 +490,54 @@ impl Database for MemoryDatabase {
...
+    fn del_spent_utxos(&mut self, spent_utxos: Vec<LocalUtxo>) -> Result<Vec<LocalUtxo>, Error> {
+        let mut deleted_utxos: Vec<LocalUtxo> = vec![];
+        if spent_utxos.is_empty() {
+            let all_spent_utxos: Vec<LocalUtxo> = self.select_spent_utxos();
+            self.delete_spent_utxo_set(&mut deleted_utxos, &all_spent_utxos)?;
+        } else {
+            self.delete_spent_utxo_set(&mut deleted_utxos, &spent_utxos)?;
+        }
+
+        Ok(deleted_utxos)
```
rajarshimaitra: if you don't pass in the deleted utxo collector this could would simply look like 
```
if spent_utxos.is_empty() {
           Ok(self.delete_spent_utxo_set(&self.select_spent_utxos())?);
        } else {
            Ok(self.delete_spent_utxo_set(&spent_utxos)?);
        }
```

src/database/mod.rs
```diff
@@ -158,6 +170,17 @@ pub trait Database: BatchOperations {
     ///
     /// It should insert and return `0` if not present in the database
     fn increment_last_index(&mut self, keychain: KeychainKind) -> Result<u32, Error>;
+
+    /// Delete UTXOs provided as params, if `spent_utxos` is empty it deletes all spent
+    /// UTXOs in the database.
+    fn del_spent_utxos(&mut self, spent_utxos: Vec<LocalUtxo>) -> Result<Vec<LocalUtxo>, Error>;
```
rajarshimaitra: I feel the API would be more clear if we asked for `Option<Vec<OutPoint>>` .. `LocalUtxo` is an internal structure we use to store more metadata about an utxo.. But an utxo could be identified by an `OutPoint`. And we should not burden users to provide a `LocalUtxo`. We should be able to fetch a `LocalUtxo` given an `OutPoint`. 

And delete all utxos if user provide `None` instead of a list of outpoints..

src/database/mod.rs
```diff
@@ -55,6 +55,18 @@ pub struct SyncTime {
...
+/// spent UTXOs
+#[derive(Clone, Debug)]
+pub struct DelCriteria {
+    /// Number of Confirmation on UTXOs to
+    /// make it eligible for deletion
+    pub confirmations: Option<u32>,
+    /// Max number of allowable spent
+    /// UTXOs in the database
+    pub threshold_size: Option<u64>,
+}
```
rajarshimaitra: The comments over the field could be made in a single line.. They are not that big..

rajarshimaitra (commented):
Concept ACK on we should have a utxo deleting API..

Below are few comments over the API and some internal codes..

src/database/keyvalue.rs
vladimirfomene: The idea was to make the code shorter and more readable by abstraction the functionality of deleting a list of spent utxos.

src/database/mod.rs
vladimirfomene: Okay, I will change this.

src/database/mod.rs
vladimirfomene: Adding the option makes total sense, I will make this change. Is `LocalUtxo` really an internal structure, though? It is exposed in the docs and returned in methods like `iter_utxos` and `get_utxo`.

rajarshimaitra: Yes it is.. Along with all other bdk specific types.. They are useful as a holder of extra meta data around common structures, this one around an outpoint.. So they makes sense as function outputs.. But not as inputs.. Because then you are forcing users to find out extra data that they don't really need for the API call..

src/database/keyvalue.rs
rajarshimaitra: While I agree in general abstracting away makes the code shorter in one place.. But you also end up distributing chunks of logic around the codebase. Now theres one more extra place to look at while reading the deletion logic.. Which can make reading through actually more difficult.. 

Having all the business logic in a single place can also add to readability..  Here to me, it seems the trait is only taking away 2 lines of code which could have been in the main function itself.. And its only used for `keyvalue`..

src/database/keyvalue.rs
```diff
@@ -379,6 +379,77 @@ impl Database for Tree {
...
+            .iter_utxos()?
+            .into_iter()
+            .filter(|utxo| utxo.is_spent)
+            .collect();
+        let spent_utxo_size = spent_utxos.len();
+        if let Some(threshold_size) = criteria.threshold_size {
+            // if the threshold size is smaller than current size, delete everything
+            if threshold_size < spent_utxo_size as u64 {
+                //delete all spent utxos
+                return self.del_spent_utxos(None);
```
rajarshimaitra: How about deleting only the excess utxos sorted by confirmations??

src/database/memory.rs
```diff
@@ -126,6 +128,42 @@ impl MemoryDatabase {
...
+                deleted_utxos.push(utxo);
+            }
+        }
+
+        Ok(deleted_utxos)
+    }
+
+    fn count_spent_utxos(&self) -> usize {
+        self.select_spent_utxos().len()
+    }
```
rajarshimaitra: Do we really need a function for this?

rajarshimaitra (commented):
Went over the code, and here are some more thoughts and comments.. Sorry for being too verbose on approach.. But I think we can do some cool stuffs here.. 

 - It seems to me the the basic operations of getting and deleting an utxos are already part of the `Database` trait.. So heres a wild thought.. Why not have a default deleting utxo implementation in `Database` trait itself?? 

I am not sure if its possible, but have you tried that route?? Could be really interesting and users won't have to define it for their custom type anymore if it can be default implemented.. Will also help us concentrate the deletion logic in just one place in the whole codebase, and not have codes for each impls..  

It will work **iff**, all the basic method calls you need to make for the selection and deletion are part of the `Database` trait..   

I also feel the `delete_spent_utxo_by_criteria()` should use the `delete_spent_utxo()` call internally and use that function to delete selected utxo upon given criteria.. Should not need its own separate code to make delete calls.. Only needs to pass on a correct vec of utxos to delete.. 

The two criteria condition `confirmation` and `threshold_size` should be asserted together.. For example
if confirmation = 5, threshold_size = 10, after deleting the database should have <10 utxos, all with confirmation<5.. I am not very if thats what happening currently..  
 

I feel this will separate the logic a lot, and reuse it where possible.. And code will be much shorter in that way.. And if we can actually do the default way, that would be the best thing to do in terms of maintainability..  

 - To verify that our `DelCriteria` is working as expected we should also include a test for it too..

src/database/keyvalue.rs
```diff
@@ -379,6 +379,77 @@ impl Database for Tree {
...
+                .filter(|utxo| utxo.is_spent)
+                .collect();
+
+            for spent_utxo in all_spent_utxos {
+                if let Some(utxo) = self.del_utxo(&spent_utxo.outpoint)? {
+                    deleted_utxos.push(utxo);
+                }
+            }
+            Ok(deleted_utxos)
+        }
```
rajarshimaitra: Just some generic code suggestions. Feel free to ignore for this PR.. This code also looks good to me..

Rust actually allows us to compress our code a lot, if we are only interacting with vectors of stuffs.. For functions which takes in `Vec<stuff1>` and spits out `Vec<stuffs2>`, the whole function logic can be boiled down to mapping one kind of vector into another kind of vector.

Doing this here would look something like this.. 

```
if let Some(utxos_to_delete) = spent_outpoints {
            utxos_to_delete
                .iter()
                .filter_map(|utxo| self.del_utxo(utxo).transpose())
                .collect::<Result<Vec<_>, _>>()
        } else {
            self.iter_utxos()?
                .iter()
                .filter(|utxo| utxo.is_spent)
                .filter_map(|spent_utxo| self.del_utxo(&spent_utxo.outpoint).transpose())
                .collect::<Result<Vec<LocalUtxo>, _>>()
        }
``` 

Efficiently using the rust `iter()` and `map()` abstractions allows us to have compact as well as highly efficient code. This is also called "Ergonomic Rust".. Kinda same idea of "Pythonic"..   

PS: There's always an one-liner 😛 (if all you are dealing with is vector of stuffs)..

rajarshimaitra: Here you can also observe, the only two calls to `self` is `iter_utxos` and `del_utxo`.. Both are part of the `Database` trait and already defined.. So this code could in theory sit inside the trait definition itself..

vladimirfomene: @rajarshimaitra, thanks for this! I will go with the one-liners, they are great!

src/database/memory.rs
vladimirfomene: This has been removed

src/database/keyvalue.rs
vladimirfomene: Will deleting just the excess not mean they will have to do frequent cleanups?

vladimirfomene: Will deleting just the excess not mean they will have to do frequent cleanups?

vladimirfomene:
Thanks @rajarshimaitra for the great feedback! I have pushed some changes with the following:
- [x] Made the code more ergonomic.
- [x] Implemented the functionality as defaults methods on the Database trait.
- [x] Use `del_spent_utxos` in `delete_spent_utxos_by_criteria`.

What is left:
I have not added test because I believe a lot will still change on the current implementation. I have also not amended the code so that the two criteria can work in conjunction.

vladimirfomene:
Thanks @rajarshimaitra for the great feedback! I have pushed some changes with the following:
- [x] Made the code more ergonomic.
- [x] Implemented the functionality as defaults methods on the Database trait.
- [x] Use `del_spent_utxos` in `delete_spent_utxos_by_criteria`.

src/database/mod.rs
```diff
@@ -158,6 +168,73 @@ pub trait Database: BatchOperations {
     ///
     /// It should insert and return `0` if not present in the database
     fn increment_last_index(&mut self, keychain: KeychainKind) -> Result<u32, Error>;
+
+    /// Delete UTXOs provided as params, if `spent_utxos` is empty it deletes all spent
+    /// UTXOs in the database.
+    fn del_spent_utxos(
+        &mut self,
+        spent_outpoints: Option<Vec<OutPoint>>,
```
rajarshimaitra: probably better to name it something like `to_delete`? All utxos in this context are "spent".

src/database/mod.rs
```diff
@@ -158,6 +168,73 @@ pub trait Database: BatchOperations {
     ///
     /// It should insert and return `0` if not present in the database
     fn increment_last_index(&mut self, keychain: KeychainKind) -> Result<u32, Error>;
+
+    /// Delete UTXOs provided as params, if `spent_utxos` is empty it deletes all spent
+    /// UTXOs in the database.
```
rajarshimaitra: Something like this can sound clearer?
```suggestion
    /// Delete a list of spent utxos from database. Delete all spent utxos if  list is `None`. 
```

rajarshimaitra (commented):
Review ACK  c1cae755b25d53e117cf86a8838ce01d822a295f

Some nits and comments..

src/database/mod.rs
```diff
@@ -158,6 +168,73 @@ pub trait Database: BatchOperations {
...
+                            > confirmations.unwrap())
+                    {
+                        filtered_outpoints.push(spent_utxo.outpoint);
+                    }
+                }
+                Ok(self.del_spent_utxos(Some(filtered_outpoints))?)
+            }
+            None => Ok(vec![]),
+        }
+    }
```
rajarshimaitra: I am not much sure on the criteria part.. If the wallet dev wants to delete utxos upon some criteria they should be able to do that on their own in their code. And such criteria in practice can be very diverse. It doesn't help for us to anticipate them and provide defaults. More options at such low level database API is more confusing..

But if it has to stay, I would expect a behavior from the structure something like below..

Both the criteria are defined as options. So they are both free to be set independent of each other. So when both is set I would expect an `and` of them, and when one is `None` it should not have any effect.

So there code here should do something like this, instead of deleting    per each criteria.
 - Choose all utxos to delete by confirmation criteria.
 - Apply threshold criteria on remaining utxos and get how many more to delete, if any. If some, select them by decreasing confirmation time from utxo list.
 - Compile both of the two list together. 
  - Delete those utxos..

src/database/mod.rs
```diff
@@ -158,6 +168,73 @@ pub trait Database: BatchOperations {
...
+        let spent_utxos: Vec<LocalUtxo> = self
+            .iter_utxos()?
+            .into_iter()
+            .filter(|utxo| utxo.is_spent)
+            .collect();
+        if let Some(threshold_size) = criteria.threshold_size {
+            // if the threshold size is smaller than current size, delete everything
+            if threshold_size < spent_utxos.len() as u64 {
+                //delete all spent utxos
+                return self.del_spent_utxos(None);
```
danielabrozzoni: I agree with what Raj is saying below: you can't really delete all the utxos if there are more than `threshold_size`, you probably just want to delete the oldest ones (=first spent ones)

src/database/mod.rs
```diff
@@ -55,6 +55,16 @@ pub struct SyncTime {
     pub block_time: BlockTime,
 }
 
+/// Structure encapsulates criteria used for deleting
+/// spent UTXOs
+#[derive(Clone, Debug)]
+pub struct DelCriteria {
+    /// Number of confirmations on UTXOs to make it eligible for deletion
+    pub confirmations: Option<u32>,
```
danielabrozzoni: This is a weird criteria IMHO... It would make sense to say "delete all the utxos spent more than N blocks ago", but that's not what you're doing: you're deleting the UTXOs that have been _received_ more than N blocks ago.
For example

```
Current time: block 15
Utxo A: reiceived at block 9, spent at block 10
Utxo B: received at block 1, spent at block 14
```

You probably want to delete utxo A, but keep  utxo B: it's been spent only since one block, so there could still be reorg, you might have to feebump it if it goes back in mempool, etc.

danielabrozzoni (commented):
Concept ACK, but this definitely needs some tests, showing how to delete by confirmations and/or by threshold size, and checking that the deletions are correct. This is also useful for reviewers to check if the code is correct and/or if the API itself is easy to understand

> I have not added test because I believe a lot will still change on the current implementation.

You're probably right, on the other hand we don't want to merge something broken, and tests are the easiest way to find out if it's all ok

src/database/mod.rs
vladimirfomene: @rajarshimaitra don't you think it is more descriptive to keep `del_spent_utxos`?

src/database/mod.rs
vladimirfomene: Sure, will make the change

vladimirfomene: Sure, will make the change

src/database/mod.rs
danielabrozzoni: I think he means using `to_delete` instead of `spent_outpoints`

src/database/mod.rs
vladimirfomene: Okay, I will go with Raj's suggestion

src/database/mod.rs
vladimirfomene: Noted. I will change the implementation to "delete all the utxos spent more than N blocks ago".

src/database/mod.rs
rajarshimaitra: Yes that's what I meant..

vladimirfomene:
@rajarshimaitra and @danielabrozzoni , I have amended the PR based on your recommendations. I have also added tests for the methods.

danielabrozzoni:
@vladimirfomene did you also fix https://github.com/bitcoindevkit/bdk/pull/725#discussion_r967649903 ? Looking at the code, it seems to me it hasn't been

vladimirfomene:
Yes I did

src/database/mod.rs
```diff
@@ -765,5 +765,285 @@ pub mod test {
...
+
+        let (first_utxo, _second_utxo, third_utxo, fourth_utxo) =
+            setup_del_spent_utxo_test(&mut db, false);
+        // test that insertion was successful
+        assert_eq!(db.iter_utxos().unwrap().len(), 4);
+
+        // call del_spent_utxos with None
+        let mut res = db.del_spent_utxos(None).unwrap();
+        res.sort_by(|a, b| a.txout.value.cmp(&b.txout.value));
+        // verify that the one spent utxo has been deleted.
```
danielabrozzoni: nit: "verify that the spent ones has been deleted" is slightly easier to understand

src/database/mod.rs
```diff
@@ -765,5 +765,285 @@ pub mod test {
...
+        db.set_utxo(&first_utxo).unwrap();
+        db.set_utxo(&third_utxo).unwrap();
+        db.set_utxo(&fourth_utxo).unwrap();
+
+        // call del_spent_utxos with vector of utxos
+        let mut res = db
+            .del_spent_utxos(Some(vec![
+                first_utxo.outpoint,
+                third_utxo.outpoint,
+                fourth_utxo.outpoint,
```
danielabrozzoni: Maybe here you should call del_spent_utxos with just 2 of the 3 utxos that have been spent, so that we can check that `del_spent_utxos(Some(v))` deletes all the utxos in `v` and not all the spent ones

src/database/mod.rs
```diff
@@ -158,6 +168,108 @@ pub trait Database: BatchOperations {
...
+
+    /// Delete a list of spent utxos from database. Delete all spent utxos if  list is `None`.
+    fn del_spent_utxos(
+        &mut self,
+        to_delete: Option<Vec<OutPoint>>,
+    ) -> Result<Vec<LocalUtxo>, Error> {
+        if let Some(to_delete) = to_delete {
+            let deleted_utxos = to_delete
+                .iter()
+                .filter_map(|out| self.del_utxo(out).transpose())
```
danielabrozzoni: I think we should also check that the `to_delete` txouts are effectively spent, otherwise we might have users accidentally deleting unspent txouts and then having problems with sync/balance

src/database/mod.rs
```diff
@@ -158,6 +168,108 @@ pub trait Database: BatchOperations {
...
+            .iter()
+            .filter_map(|utxo| self.get_tx(&utxo.outpoint.txid, false).transpose())
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let txs_conf_heights = tx_details
+            .iter()
+            .filter_map(|details| details.confirmation_time.as_ref().map(|conf| conf.height))
+            .collect::<Vec<_>>();
+        let utxos_heights: Vec<(&LocalUtxo, &u32)> =
+            spent_utxos.iter().zip(txs_conf_heights.iter()).collect();
```
danielabrozzoni: Can you really `zip` here? I think there are cases where `txs_conf_heights` is shorter than `spent_utxos`:
- if `self.get_tx(&utxo.outpoint.txid, false).transpose()` is `None`
- if `details.confirmation_time.as_ref().map(|conf| conf.height)` is `None`

src/database/mod.rs
```diff
@@ -158,6 +168,108 @@ pub trait Database: BatchOperations {
...
+
+        let mut to_delete: Vec<&LocalUtxo> = vec![];
+        let mut failed_confirmation_criteria: Vec<(&LocalUtxo, &u32)> = vec![];
+
+        // Choose all utxos to delete by confirmation criteria.
+        if criteria.confirmations.is_some() {
```
danielabrozzoni: nit: this could be an `if let`

src/database/mod.rs
```diff
@@ -158,6 +168,108 @@ pub trait Database: BatchOperations {
...
+        }
+
+        // apply threshold criteria on spent utxos
+        if criteria.threshold_size.is_some() {
+            if failed_confirmation_criteria.is_empty() {
+                failed_confirmation_criteria.extend(utxos_heights.iter());
+            }
+
+            // only select on threshold if there are spent utxos to be deleted
+            if spent_utxos.len() - to_delete.len() > 0 {
```
danielabrozzoni: Use `spent_utxos.len() > to_delete.len()` instead: if `spent_utxos.len() - to_delete.len()` is less than zero, I think Rust would panic (subtraction underflow)

src/database/mod.rs
```diff
@@ -158,6 +168,108 @@ pub trait Database: BatchOperations {
...
+                    .map(|(utxo, _height)| *utxo)
+                    .collect::<Vec<&LocalUtxo>>(),
+            );
+
+            // add utxos that failed the test to `failed_confirmation_criteria`
+            failed_confirmation_criteria.extend(fail.iter());
+        }
+
+        // apply threshold criteria on spent utxos
+        if criteria.threshold_size.is_some() {
```
danielabrozzoni: same: I would use `if let`, so you don't have to unwrap below

src/database/mod.rs
```diff
@@ -158,6 +168,108 @@ pub trait Database: BatchOperations {
...
+            if failed_confirmation_criteria.is_empty() {
+                failed_confirmation_criteria.extend(utxos_heights.iter());
+            }
+
+            // only select on threshold if there are spent utxos to be deleted
+            if spent_utxos.len() - to_delete.len() > 0 {
+                let qty_to_delete = spent_utxos.len() as u64
+                    - criteria.threshold_size.unwrap()
+                    - to_delete.len() as u64;
```
danielabrozzoni: I think this could underflow as well... you might want to use https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_sub to avoid problems

```
let qty_to_delete = spent_utxos.len() as u64
                               - to_delete.len() as u64
                               .saturanting_sub(criteria.threshold_size.unwrap());
```

(I didn't compile this code, it probably needs some more parenthesis to work)

src/database/mod.rs
```diff
@@ -158,6 +168,108 @@ pub trait Database: BatchOperations {
...
+            if current_block_height.is_none() {
+                return Err(Error::Generic(String::from(
+                        "You must have a non-None `current_block_height` when using the `confirmations` criteria",
+                    )));
+            }
+
+            // split utxos_heights pair based on confirmation criteria
+            let (pass, fail): (Vec<_>, Vec<_>) =
+                utxos_heights.iter().partition(|(_utxo, height)| {
+                    (current_block_height.unwrap() - **height) >= criteria.confirmations.unwrap()
```
danielabrozzoni: I would use `saturating_sub` for `current_block_height.unwrap() - **height`, to avoid panics

src/database/mod.rs
vladimirfomene: Done

src/database/mod.rs
vladimirfomene: done

src/database/mod.rs
vladimirfomene: Done

vladimirfomene:
I have removed the `del_spent_utxos_by_criteria` method and its test. We decided to remove the `del_spent_utxos_by_criteria` because we don't currently store data that tells us how many blocks ago a utxo was spent. We only store data that tells us how many blocks ago, a utxo was received. This makes it hard to calculate the number of blocks passed since a utxo was spent and therefore difficult to use that as a criteria for deletion.

notmandatory:
Hey sorry but I need to take this out of the 0.23 release, and there's a chance that with the up-coming `bdk_core` work  we may not need it.  But this PR is a still a good bit of work so let's leave it open until we have more info.

rajarshimaitra:
My 2 sats would be to get it into bdk anyway even if we might not need it after bdk_core integration.. A full featured bdk_core integration would take a bit of time, and this can be useful feature to add for wallet devs anyway who might wanna stay with the older versions of bdk..

danielabrozzoni:
In my opinion getting this in might cause more disrupt than help: 
- It takes resources away from bdk-core (I can't work on bdk_core and review on bdk at the same time :))
- We don't have proof that people need this right now 
I agree that it *might* be useful to get this one in for the ones that want to stay on bdk `0.something` instead of switching to `1.0`, but IMHO we should wait for the bdk_core integration to be almost ready before deciding this

danielabrozzoni:

PR #726, opened 2022-08-17
[bug-fix] Set the db sync height

### LINKED ISSUE #719: [bug] get_funded_wallet() does not return balance

rajarshimaitra:
### Description

Fixes #719 

Previously we weren't setting the db sync height in populate_test_db
macro even when current height is provided.. This creates a bug that
get_funded_wallet will return 0 balance.

This PR fixes the populate_test_db macro and updates tests which were
previously dependent on the unsynced wallet behavior.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/database/memory.rs
+++ b/src/database/memory.rs
@@ -486,7 +486,8 @@ macro_rules! populate_test_db {
     }};
     ($db:expr, $tx_meta:expr, $current_height:expr, (@coinbase $is_coinbase:expr)$(,)?) => {{
         use std::str::FromStr;
-        use $crate::database::BatchOperations;
+        use $crate::database::SyncTime;
+        use $crate::database::{BatchOperations, Database};
         let mut db = $db;
         let tx_meta = $tx_meta;
         let current_height: Option<u32> = $current_height;
@@ -512,13 +513,31 @@ macro_rules! populate_test_db {
         };
 
         let txid = tx.txid();
+        // Set Confirmation time only if current height is provided.
+        // panics if `tx_meta.min_confirmation` is Some, and current_height is None.
         let confirmation_time = tx_meta
             .min_confirmations
             .and_then(|v| if v == 0 { None } else { Some(v) })
             .map(|conf| $crate::BlockTime {
-                height: current_height.unwrap().checked_sub(conf as u32).unwrap() + 1,
+                height: current_height.expect("Current height is needed for testing transaction with min-confirmation values").checked_sub(conf as u32).unwrap() + 1,
                 timestamp: 0,
             });
+
+        // Set the database sync_time.
+        // Check if the current_height is less than already known sync height, apply the max
+        // If any of them is None, the other will be applied instead.
+        // If both are None, this will not be set.
+        if let Some(height) = db.get_sync_time().unwrap()
+                                .map(|sync_time| sync_time.block_time.height)
+                                .max(current_height) {
+            let sync_time = SyncTime {
+                block_time: BlockTime {
+                    height,
+                    timestamp: 0
+                }
+            };
+            db.set_sync_time(sync_time).unwrap();
+        }
 
         let tx_details = $crate::TransactionDetails {
             transaction: Some(tx.clone()),
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1927,6 +1927,12 @@ pub(crate) mod test {
     const P2WPKH_FAKE_WITNESS_SIZE: usize = 106;
 
     #[test]
+    fn test_get_funded_wallet_balance() {
+        let (wallet, _, _) = get_funded_wallet(get_test_wpkh());
+        assert_eq!(wallet.get_balance().unwrap().confirmed, 50000);
+    }
+
+    #[test]
     fn test_cache_addresses_fixed() {
         let db = MemoryDatabase::new();
         let wallet = Wallet::new(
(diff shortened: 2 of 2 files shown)


src/wallet/mod.rs
```diff
@@ -2862,7 +2898,21 @@ pub(crate) mod test {
...
+            Network::Regtest,
+            AnyDatabase::Memory(MemoryDatabase::new()),
+        )
+        .unwrap();
+
+        let tx_meta = testutils! {
+                @tx ( (@external descriptors, 0) => 50_000 ) (@confirmations 1)
+        };
+
+        // Add the transaction to our db, but do not sync the db.
```
danielabrozzoni: Same here, a comment would be really helpful!

src/wallet/mod.rs
```diff
@@ -2364,7 +2385,22 @@ pub(crate) mod test {
...
```
danielabrozzoni: Can we add a comment here explaining why we don't want to sync the db?

rajarshimaitra: Done.. Yes would be helpful..

rajarshimaitra:
Updated with review comments..

src/wallet/mod.rs
```diff
@@ -2364,7 +2385,23 @@ pub(crate) mod test {
...
+            AnyDatabase::Memory(MemoryDatabase::new()),
+        )
+        .unwrap();
+
+        let tx_meta = testutils! {
+                @tx ( (@external descriptors, 0) => 50_000 ) (@confirmations 1)
+        };
+
+        // Add the transaction to our db, but do not sync the db. Unsynced db
+        // should trigger the default sequqence value for a new transaction as 0xFFFFFFFF
```
vladimirfomene: Nits: spelling error on `sequence`

src/wallet/mod.rs
```diff
@@ -2862,7 +2899,22 @@ pub(crate) mod test {
...
```
vladimirfomene: Nits: spelling error on `sequence`

vladimirfomene (commented):
utACK

src/wallet/mod.rs
rajarshimaitra: Thanks.. Done..

rajarshimaitra:
Thanks @vladimirfomene for the catches.. Updated..

danielabrozzoni (approved):
utACK b8ce842ce01752ce693759bbfa5007f5619c2424

src/database/memory.rs
```diff
@@ -512,13 +513,29 @@ macro_rules! populate_test_db {
...
             .and_then(|v| if v == 0 { None } else { Some(v) })
             .map(|conf| $crate::BlockTime {
-                height: current_height.unwrap().checked_sub(conf as u32).unwrap() + 1,
+                height: height.checked_sub(conf as u32).unwrap() + 1,
                 timestamp: 0,
-            });
+            })
+        } else {
+            None
+        };
```
afilini: I don't think it's ok to silently drop the `min_confirmations` if the current height is not provided. I would rather keep what we had before with a better error message, by changing the `unwrap()` into an `expect()`.

src/database/memory.rs
```diff
@@ -512,13 +513,29 @@ macro_rules! populate_test_db {
...
+
+        // Set the database sync height if current height is provided
+        if let Some(height) = current_height {
+            let sync_time = SyncTime {
+                block_time: BlockTime {
+                    height,
+                    timestamp: 0
+                }
+            };
+            db.set_sync_time(sync_time).unwrap();
```
afilini, line 539: I think here you should only keep the most recent sync time, so instead of just setting it you should first check if something is present and compare the two

src/database/memory.rs
rajarshimaitra: Makes sense..

src/database/memory.rs
rajarshimaitra: It should.. Done.,.

rajarshimaitra:
Thanks @afilini  for the look.. Updated as suggested.. In the last dev call we talked about releasing this as a patch version for some downstream libs to update.. See https://github.com/bitcoindevkit/bdk-cli/pull/102#issuecomment-1226928903

danielabrozzoni:
Sorry I didn't notice the two issues that Alekos has found, I should have inspected the code better.

re-utACK a72a97b663fa71cb66e91f1187dc257762dbb2a7

notmandatory:
@rajarshimaitra  even though we talked about doing a fix release for this, since this is a test framework fix and code freeze for 0.22 is next Wed. (Aug 31), can we skip doing a fix release only for this PR and wait for 0.22 to put it out (Sept 7)?

rajarshimaitra:
@notmandatory sounds good to me.. Only needed for the bdk-cli PR.. That can wait a bit..

src/database/memory.rs
```diff
@@ -512,14 +513,32 @@ macro_rules! populate_test_db {
...
+                height: current_height.expect("Current height is needed for testing transaction with min-confirmation values").checked_sub(conf as u32).unwrap() + 1,
                 timestamp: 0,
             });
 
+        // Set the database sync_time.
+        // Check if the current_height is less than already known sync height, apply the max
+        // If any of them is None, the other will be applied instead.
+        // If both are None, this will not be set.
+        if let Some(height) = db.get_sync_time().unwrap()
+                                .and_then(|sync_time| Some(sync_time.block_time.height))
```
afilini: nit: You can replace `and_then` with `map` because you always return `Some()`

rajarshimaitra:
Updated and rebased..

afilini (commented):
ACK 08668ac46247d527cc53af5b6f359b1fa4e3b6aa

PR #544, opened 2022-02-16
Update DEVELOPMENT_CYCLE.md to work with [patch.crates-io]

notmandatory:
### Description

Update DEVELOPMENT_CYCLE and release instructions to make [overriding dependencies] possible for downstream projects with unreleased `bdk` versions for development and testing. Also simplifies the release process by capturing changelog information in the `pull_request_template` and recording release changelog information in the release tag message instead of in a `CHANGELOG.md` file which causes too many merge conflicts and complicates the release process.

Fixes bitcoindevkit/bdk#536 
Fixes bitcoindevkit/bdk#496 

### Notes to the reviewers

The primary changes to our current release process are:

1. Don't add `-dev` or `-rc.x` to unreleased `bdk` cargo versions because those extensions do not work with [overriding dependencies].
2. Increment the `master` branch version as soon as a `release/MAJOR.MINOR` branch is created, the next release `release/MAJOR.MINOR` branch version with be **MAJOR.MINOR.PATCH**, and the `master` branch development version will be **MAJOR.MINOR+1.0**; either version can be used with [overriding dependencies].
4. Remove the `bdk` version from the `src/lib.rs` file so that it doesn't need to be changed on every release, because it isn't needed in the rust docs for most developers and removing it will help simplify the release process.
5. The new release process is now documented as a checklist in a new `release.md` github issue template.
6. Putting changelog information in the release tag message is how the tokio project does it. ~~After this PR is merged I will replace old tags with new ones containing changelog information and then do a new PR to remove the CHANGELOG.md file.~~ After this PR is merged I don't think we need to update old tags, only rename the CHANGELOG.md file to CHANGELOG-OLD.md with a note to check tags for future change log info.

[overriding dependencies]: https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #727, opened 2022-08-17
Add `process_and_select_coins` function

csralvall:
### Description

There were some common tasks identified in the construction of a coin selection
result that the different algorithms performed separately. Those tasks were
blended inside the algorithms or separated into functions, but the underlying
actions were the same:
- Do some preprocessing to the inputs.
- Select utxos based in the algorithm criterium.
- Decide change.
- Build the coin selection result.

This commit intends to create a single function that consumes different
algorithms and perform all the mentioned tasks as a pipeline taking utxos as
input and producing coin selection results at the end.

The new function reduces the work needed to implement other coin selection
algorithms, modularizes the code not related directly to the selection process
and reduces the number of lines to review and understand its logic.

~As a side effect of the implementation, the `SingleRandomDraw` algorithm was
extracted from its host algorithm, `BranchAndBound`, and introduced as a
`FallBack` algorithm. But I consider this a feature.~

### Notes to the reviewers

I'm not totally satisfied with some parts of the code.

For example, I don't like to have "ad-hoc" parameters in the current signature
of `coin_select` just for the sake of `BranchAndBound` compatibility. I thought
of some solutions to this, like the addition of a `cost_of_change` field into the
struct or the recomputation of the available value inside
`BranchAndBound::coin_select` (duplicating the amount of work), but I would
like to hear other opinions.

Another thing that bothers me is the conversion of `target_amount` from u64 to
i64 and the other way back. I must be obfuscated with it because I can't see
another way to do it.

This PR will set up the structure needed to integrate the Waste metric (#558)
with the coin selection module.

**BREAKING CHANGES**: This PR modifies the API of `coin_select` and introduce a new
function, `get_selection`, that supersedes its functionality.

### Changelog notice

- Added `process_and_select_coins()` function.
- Added `OutputGroup`s as a public structure that holds the `WeightedUtxo`, its
  effective value, and its fees.
- Changed `CoinSelectionAlgorithm::coin_select()` signature. The scope of the
  selection was reduced to select utxos only from `optional_utxos`. It only
  returns the selected `OutputGroups` from now on.
- Added `WeightedScriptPubkey` struct.

#### All Submissions:

#### New Features:

* [x] I've ~added~ updated the tests for the new feature

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -13,7 +13,7 @@ use std::convert::AsRef;
 use std::ops::Sub;
 
 use bitcoin::blockdata::transaction::{OutPoint, Transaction, TxOut};
-use bitcoin::{hash_types::Txid, util::psbt};
+use bitcoin::{hash_types::Txid, util::psbt, Script};
 
 use serde::{Deserialize, Serialize};
 
@@ -175,6 +175,18 @@ pub struct WeightedUtxo {
     pub satisfaction_weight: usize,
     /// The UTXO
     pub utxo: Utxo,
+}
+
+/// A wallet owned script_pubkey with the weight of the `witness` + `scriptSig` in [weight units]
+/// to spend from it.
+/// Useful to optimize reducing the waste metric in coin selection algorithms.
+/// [weight units]: <https://en.bitcoin.it/wiki/Weight_units>
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct WeightedScriptPubkey {
+    /// The weight of the witness data and `scriptSig` expressed in [weight units].
+    pub satisfaction_weight: usize,
+    /// The script_pubkey
+    pub script_pubkey: Script,
 }
 
 #[derive(Debug, Clone, PartialEq)]
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -29,7 +29,6 @@
 //! # use bdk::wallet::{self, coin_selection::*};
 //! # use bdk::database::Database;
 //! # use bdk::*;
-//! # use bdk::wallet::coin_selection::decide_change;
 //! # const TXIN_BASE_WEIGHT: usize = (32 + 4 + 4) * 4;
 //! #[derive(Debug)]
 //! struct AlwaysSpendEverything;
@@ -38,44 +37,12 @@
 //!     fn coin_select(
 //!         &self,
 //!         database: &D,
-//!         required_utxos: Vec<WeightedUtxo>,
-//!         optional_utxos: Vec<WeightedUtxo>,
+//!         optional_utxos: Vec<OutputGroup>,
 //!         fee_rate: FeeRate,
 //!         target_amount: u64,
-//!         drain_script: &Script,
-//!     ) -> Result<CoinSelectionResult, bdk::Error> {
-//!         let mut selected_amount = 0;
-//!         let mut additional_weight = 0;
-//!         let all_utxos_selected = required_utxos
-//!             .into_iter()
-//!             .chain(optional_utxos)
-//!             .scan(
-//!                 (&mut selected_amount, &mut additional_weight),
-//!                 |(selected_amount, additional_weight), weighted_utxo| {
-//!                     **selected_amount += weighted_utxo.utxo.txout().value;
-//!                     **additional_weight += TXIN_BASE_WEIGHT + weighted_utxo.satisfaction_weight;
-//!                     Some(weighted_utxo.utxo)
-//!                 },
-//!             )
-//!             .collect::<Vec<_>>();
-//!         let additional_fees = fee_rate.fee_wu(additional_weight);
-//!         let amount_needed_with_fees = additional_fees + target_amount;
-//!         if selected_amount < amount_needed_with_fees {
-//!             return Err(bdk::Error::InsufficientFunds {
-//!                 needed: amount_needed_with_fees,
-//!                 available: selected_amount,
-//!             });
-//!         }
-//!
-//!         let remaining_amount = selected_amount - amount_needed_with_fees;
-//!
-//!         let excess = decide_change(remaining_amount, fee_rate, drain_script);
-//!
-//!         Ok(CoinSelectionResult {
-//!             selected: all_utxos_selected,
-//!             fee_amount: additional_fees,
-//!             excess,
-//!         })
+//!         available_value: i64,
+//!     ) -> Result<Vec<OutputGroup>, bdk::Error> {
+//!         Ok(optional_utxos)
 //!     }
 //! }
 //!
@@ -94,9 +61,9 @@
 //! # Ok::<(), bdk::Error>(())
 //! ```
 
-use crate::types::FeeRate;
+use crate::database::Database;
+use crate::types::{FeeRate, WeightedScriptPubkey, WeightedUtxo};
 use crate::wallet::utils::IsDust;
-use crate::{database::Database, WeightedUtxo};
 use crate::{error::Error, Utxo};
 
 use bitcoin::consensus::encode::serialize;
@@ -171,6 +138,171 @@ impl CoinSelectionResult {
     }
 }
 
+/// Perform the coin selection
+///
+/// - `algorithm`: the algorithm to use to select the UTXOs for the transaction
+/// - `database`: database reference to use by those algorithms that require extra information
+///               about UTXOs
+/// - `required_utxos`: the utxos that must be spent regardless of `target_amount` with their
+///                     weight cost
+/// - `optional_utxos`: the remaining available utxos to satisfy `target_amount` with their
+///                     weight cost
+/// - `fee_rate`: fee rate to use
+/// - `target_amount`: the outgoing amount in satoshis and the fees already
+///                    accumulated from added outputs and transaction’s header.
+/// - `weighted_drain_script`: the drain script used to create the change output associated with
+///                            the max satisfaction weight needed to spend it
+pub fn process_and_select_coins<D: Database, Cs: CoinSelectionAlgorithm<D>>(
+    algorithm: Cs,
+    database: &D,
+    required_utxos: Vec<WeightedUtxo>,
+    optional_utxos: Vec<WeightedUtxo>,
+    fee_rate: FeeRate,
+    target_amount: u64,
+    weighted_drain_script: &WeightedScriptPubkey,
+) -> Result<CoinSelectionResult, Error> {
+    // ####################################################################
+    // ######################### PREPROCESSING ############################
+    // ####################################################################
+
+    // Mapping every (UTXO, usize) to an output group
+    let mut required_output_groups: Vec<OutputGroup> = required_utxos
+        .into_iter()
+        .map(|u| OutputGroup::new(u, fee_rate))
+        .collect();
+
+    // Mapping every (UTXO, usize) to an output group.
+    let optional_output_groups: Vec<OutputGroup> = optional_utxos
+        .into_iter()
+        .map(|u| OutputGroup::new(u, fee_rate))
+        .filter(|u| u.effective_value.is_positive())
+        .collect();
+
+    let (required_effective_value, required_fees) = required_output_groups
+        .iter()
+        .fold((0, 0), |(eff_value, fees), x| {
+            (eff_value + x.effective_value, fees + x.fee)
+        });
+
+    let (optional_effective_value, optional_fees) = optional_output_groups
+        .iter()
+        .fold((0, 0), |(eff_value, fees), x| {
+            (eff_value + x.effective_value, fees + x.fee)
+        });
+
+    // `required_effective_value` and `optional_effective_value` are both the sum of *effective_values* of
+    // the UTXOs. For the optional UTXOs (optional_effective_value) we filter out UTXOs with
+    // negative effective value, so it will always be positive.
+    //
+    // Since we are required to spend the required UTXOs (required_effective_value) we have to consider
+    // all their effective values, even when negative, which means that required_effective_value could
+    // be negative as well.
+    //
+    // If the sum of required_effective_value and optional_effective_value is negative or lower than our target,
+    // we can immediately exit with an error, as it's guaranteed we will never find a solution
+    // if we actually run the coin selection algorithm
+    let total_effective_value = optional_effective_value + required_effective_value;
+    let total_value: Result<u64, _> = total_effective_value.try_into();
+    match total_value {
+        Ok(v) if v >= target_amount => {}
+        _ => {
+            // Assume we spend all the UTXOs we can (all the required + all the optional with
+            // positive effective value), sum their value and their fee cost.
+            let utxo_fees = optional_fees + required_fees;
+            let utxo_value = total_effective_value + utxo_fees as i64;
+
+            // Add to the target the fee cost of the UTXOs
+            return Err(Error::InsufficientFunds {
+                needed: target_amount + utxo_fees,
+                available: utxo_value as u64,
+            });
+        }
+    }
+
+    let target_amount_i64 = target_amount
+        .try_into()
+        .expect("Bitcoin amount to fit into i64");
+
+    // ####################################################################
+    // ######################### COIN SELECTION ###########################
+    // ####################################################################
+
+    let (output_groups, effective_value, fee_amount) =
+        if required_effective_value > target_amount_i64 {
+            (
+                required_output_groups,
+                required_effective_value,
+                required_fees,
+            )
+        } else if total_effective_value == target_amount_i64 {
+            let mut selected_output_groups = optional_output_groups;
+            selected_output_groups.append(&mut required_output_groups);
+            (
+                selected_output_groups,
+                total_effective_value,
+                optional_fees + required_fees,
+            )
+        } else {
+            // from now on, target_amount can only be positive
+            let target_amount = (target_amount_i64 - required_effective_value) as u64;
+            let mut selected_output_groups = algorithm.coin_select(
+                database,
+                optional_output_groups,
+                fee_rate,
+                target_amount,
+                optional_effective_value,
+            )?;
+            let selected_effective_value: i64 = selected_output_groups
+                .iter()
+                .map(|og| og.effective_value)
+                .sum();
+            let selected_fees: u64 = selected_output_groups.iter().map(|og| og.fee).sum();
+            selected_output_groups.append(&mut required_output_groups);
+            (
+                selected_output_groups,
+                selected_effective_value + required_effective_value,
+                selected_fees + required_fees,
+            )
+        };
+
+    // ####################################################################
+    // ############################ GET EXCESS ############################
+    // ####################################################################
+
+    // if coin_select finish, it means it found a valid coin selection. The effective value of that
+    // coin selection should be greater than target_amount (an u64) so it's safe to apply the
+    // conversion from i64 to u64
+    let effective_value = effective_value as u64;
+
+    // remaining_amount = selected_amount - (target_amount + selection_fees)
+    let remaining_amount = effective_value.saturating_sub(target_amount);
+
+    let excess = decide_change(
+        remaining_amount,
+        fee_rate,
+        &weighted_drain_script.script_pubkey,
+    );
+
+    // ####################################################################
+    // ############################ GET UTXOS #############################
+    // ####################################################################
+
+    let selected = output_groups
+        .into_iter()
+        .map(|x| x.weighted_utxo.utxo)
+        .collect::<Vec<_>>();
+
+    // ####################################################################
+    // ##################### BUILD COIN SELECTION RESULT ##################
+    // ####################################################################
+
+    Ok(CoinSelectionResult {
+        selected,
+        fee_amount,
+        excess,
+    })
+}
+
 /// Trait for generalized coin selection algorithms
 ///
 /// This trait can be implemented to make the [`Wallet`](super::Wallet) use a customized coin
@@ -182,24 +314,20 @@ pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
     ///
     /// - `database`: a reference to the wallet's database that can be used to lookup additional
     ///               details for a specific UTXO
-    /// - `required_utxos`: the utxos that must be spent regardless of `target_amount` with their
-    ///                     weight cost
-    /// - `optional_utxos`: the remaining available utxos to satisfy `target_amount` with their
+    /// - `optional_utxos`: the utxos available for selection to satisfy `target_amount` with their
     ///                     weight cost
     /// - `fee_rate`: fee rate to use
     /// - `target_amount`: the outgoing amount in satoshis and the fees already
     ///                    accumulated from added outputs and transaction’s header.
-    /// - `drain_script`: the script to use in case of change
-    #[allow(clippy::too_many_arguments)]
+    /// - `available_value`: the total effective value of all the optional utxos
     fn coin_select(
         &self,
         database: &D,
-        required_utxos: Vec<WeightedUtxo>,
-        optional_utxos: Vec<WeightedUtxo>,
+        optional_utxos: Vec<OutputGroup>,
         fee_rate: FeeRate,
         target_amount: u64,
-        drain_script: &Script,
-    ) -> Result<CoinSelectionResult, Error>;
+        available_value: i64,
+    ) -> Result<Vec<OutputGroup>, Error>;
 }
 
 /// Simple and dumb coin selection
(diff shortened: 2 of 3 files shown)


src/wallet/coin_selection.rs
```diff
@@ -567,27 +571,28 @@ impl BranchAndBoundCoinSelection {
...
 
                 // If we found a solution better than the previous one, or if there wasn't previous
                 // solution, update the best solution
-                if best_selection_value.is_none() || curr_value < best_selection_value.unwrap() {
+                if best_selection_value.is_none() || selected_value < best_selection_value.unwrap()
+                {
                     best_selection = current_selection.clone();
-                    best_selection_value = Some(curr_value);
+                    best_selection_value = Some(selected_value);
                 }
```
murchandamus, line 587: I have the impression that this optimizes for the smallest excess rather than the smallest waste. I was wondering whether this is an intentional design choice, and if so, whether you could elaborate the thought process.

src/wallet/coin_selection.rs
```diff
@@ -567,27 +571,28 @@ impl BranchAndBoundCoinSelection {
...
 
                 // If we found a perfect match, break here
-                if curr_value == target_amount {
+                if selected_value == target_amount {
                     break;
                 }
```
murchandamus, line 592: This may not necessarily be the solution with the best waste score (but of course an optimal solution if minimizing the excess).

murchandamus (commented):
I'm not sure how BDK prefers to structure its commits, but on Bitcoin Core, we like to separate renaming of things and relocating without functional changes from functional changes to separate commits, i.e. it might be easier to review if the relocation of SRD were a separate commit.

src/wallet/coin_selection.rs
csralvall: Yes, of course! I realized we passed the `required_utxos` as parameter to all
the algorithms, but in the end, they worked as a fixed value in all coin
selections, and the `required_utxos` should be included no-matter-what in the
CoinSelectionResult, as its original logic.
In BranchAndBound, something similar happened with `curr_value` because
it was the `required_utxos`'s summed amount.
`curr_value`, as implemented at that moment, wasn't a fixed value, but it had a
lower bound: the amount compounded by `required_utxos`. So, my idea was to
remove the `required_utxos` parameter, and related parameters like `curr_value`,
from the `coin_select` signature, and create a new `target_amount` subtracting the
effective value of `required_utxos`. The coin selection would depend from
that moment on the `optional_utxos` and the new target.
`curr_value` followed the same logic with the new target: the lower bound became
zero because we were already considering the `required_utxos`'s amount in the new
`target_amount`, and then there would be no need to take it as a parameter.
The change of the name was because it eased my thinking while making the
changes, I mean, in the end, `curr_value` held the value of all the
`selected_utxos`, so it was the "selected_value".

csralvall:
> I'm not sure how BDK prefers to structure its commits, but on Bitcoin Core, we like to separate renaming of things and relocating without functional changes from functional changes to separate commits, i.e. it might be easier to review if the relocation of SRD were a separate commit.

I thought that squashing things was the way to proceed but I can split this in two commits or more.

src/wallet/coin_selection.rs
murchandamus: I think we're talking past each other. What I was asking about seems unrelated to the required_utxos to me. The Branch and Bound Coin Selection Strategy aims to find a changeless input set that minimizes a transaction's _waste_ (which is a proxy for the transaction's cost in the context of the wallet's lifetime). To that end, we keep the input set that produces the lowest `curr_waste`, rather than the lowest `selected_value`.

image
(from [`/src/wallet/coinselection.cpp`](https://github.com/bitcoin/bitcoin/blob/master/src/wallet/coinselection.cpp#L86-L110))

The motivation here is that e.g. at high feerates you should prefer using two inputs that drop 100 sats more to fees, instead of paying the fees for five inputs that overshoot the target slightly less.

So, when I saw

```
                if best_selection_value.is_none() || selected_value < best_selection_value.unwrap()
                {
                    best_selection = current_selection.clone();
                    best_selection_value = Some(selected_value);
                }
 ```

 I was wondering whether you were optimizing for the smallest overshoot on purpose, or whether you also wanted to optimize for smallest waste instead, as we do in Bitcoin Core.

Comment (danielabrozzoni) [reply src/wallet/coin_selection.rs:587]: @Xekyo I think you're right in saying that we're not trying to minimize the waste: my understanding of the bnb coin selection wasn't optimal when I wrote the code, and it hasn't been updated ever since 😰 Thanks for pointing this out!
I opened an issue for this (https://github.com/bitcoindevkit/bdk/issues/733), and we'll try to fix asap :)

On the other end, I don't think it should be fixed in this PR, as, If I understand correctly, here @csralvall  is just refactoring the code structure (changing variable names, such as going from `curr_value` to `selected_value`, trying to abstract some operations with the new `get_selection` function, etc) - being a big enough PR already, I think we can work on fixing the algorithm in a subsequent one :)

File: src/wallet/mod.rs [hunk 3]
```diff
@@ -836,23 +838,40 @@ where
...
+                    .address
+                    .script_pubkey(),
+            };
+
+            let satisfaction_weight = if params.drain_to.is_none() {
+                self.get_descriptor_for_keychain(KeychainKind::Internal)
+                    .max_satisfaction_weight()
+                    .unwrap()
+            } else {
+                0
```
danielabrozzoni, line 860: This is wrong, you should use miniscript to get the max satisfaction weight here as well

src/wallet/coin_selection.rs
```diff
@@ -38,44 +37,16 @@
...
-//!         optional_utxos: Vec<WeightedUtxo>,
+//!         optional_utxos: Vec<OutputGroup>,
 //!         fee_rate: FeeRate,
 //!         target_amount: u64,
-//!         drain_script: &Script,
-//!     ) -> Result<CoinSelectionResult, bdk::Error> {
+//!         available_value: i64,
+//!     ) -> Result<(Vec<OutputGroup>, u64), bdk::Error> {
 //!         let mut selected_amount = 0;
 //!         let mut additional_weight = 0;
```
danielabrozzoni: I don't understand why you're leaving these two here

src/wallet/coin_selection.rs
```diff
@@ -38,44 +37,16 @@
...
-//!             .collect::<Vec<_>>();
-//!         let additional_fees = fee_rate.fee_wu(additional_weight);
-//!         let amount_needed_with_fees = additional_fees + target_amount;
-//!         if selected_amount < amount_needed_with_fees {
-//!             return Err(bdk::Error::InsufficientFunds {
-//!                 needed: amount_needed_with_fees,
-//!                 available: selected_amount,
-//!             });
-//!         }
+//!         let additional_fees = optional_utxos.iter().map(|og| og.fee).sum();
```
danielabrozzoni: This makes me wonder, should `coin_select` return the additional fees at all? We can just compute them in `get_selection`, as it's the same piece of code for every algorithm

Or maybe some algorithms (like bnb) do need to calculate them anyways, so we might as well just use what the algorithm calculated?

src/wallet/mod.rs
```diff
@@ -52,7 +52,10 @@ pub use utils::IsDust;
 
 #[allow(deprecated)]
 use address_validator::AddressValidator;
-use coin_selection::DefaultCoinSelectionAlgorithm;
+use coin_selection::{
+    get_selection, DefaultCoinSelectionAlgorithm,
+    Excess::{Change, NoChange},
```
rajarshimaitra: Naming request: A name like `HandleExcess` could be more clearer here.. Its a simple enum, that specifies how to handle the change if any.  

Not a major issue, a general rust street rule is not to import enum variants directly. Call them with the  enum name in code like `Excess::Change` etc. The name of the enum gives useful context at the usage location. And its always just two item long.

src/types.rs
```diff
@@ -167,6 +167,19 @@ pub struct WeightedUtxo {
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct WeightedScript {
+    /// The weight of the witness data and `scriptSig` expressed in [weight units]. This is used to
+    /// properly compute the cost of change for waste metric.
+    ///
+    /// [weight units]: https://en.bitcoin.it/wiki/Weight_units
+    pub satisfaction_weight: usize,
+    /// The script_pubkey
+    pub script_pubkey: Script,
+}
```
rajarshimaitra: It was becoming too many comments and mostly just nits, so did a suggestion instead.. How about we have something like below to express this struct? 

Rationales:
 - Cleaner to specify explainer details in the struct doc than in field doc.
 - Its not really a `Script`, its a `script_pubkey`, which is also `Script` type. So all `script_punbkeys` are `Script`, but not all `Script`s are `script_pubkey`. Thus seems better to me as `WeightedScriptPubkey`.  

```suggestion
/// An wallet owned script_pubkey with the weight of the `witness` + `scriptSig` in [weight units] to spend from it.
/// This is useful for coninselection algorithm to to optimize for reducing waste metric.
/// [weight units]: https://en.bitcoin.it/wiki/Weight_units 
#[derive(Debug, Clone, PartialEq)]
pub struct WeightedScriptPubkey {
    /// The weight of the `witness` and `scriptSig` expressed in [weight units].
    pub satisfaction_weight: usize,
    /// The script_pubkey
    pub script_pubkey: Script,
}
```

src/wallet/coin_selection.rs
csralvall: +1
I was confused because the logic is the same as in the master branch. This PR is mainly refactoring.

src/wallet/coin_selection.rs
```diff
@@ -325,63 +454,46 @@ pub fn decide_change(remaining_amount: u64, fee_rate: FeeRate, drain_script: &Sc
 }
 
 fn select_sorted_utxos(
-    utxos: impl Iterator<Item = (bool, WeightedUtxo)>,
-    fee_rate: FeeRate,
+    utxos: impl Iterator<Item = OutputGroup>,
     target_amount: u64,
-    drain_script: &Script,
-) -> Result<CoinSelectionResult, Error> {
+) -> Result<(Vec<OutputGroup>, u64), Error> {
```
danielabrozzoni: This doesn't need to be a Result

src/wallet/coin_selection.rs
```diff
@@ -116,6 +87,9 @@ use std::convert::TryInto;
 pub type DefaultCoinSelectionAlgorithm = BranchAndBoundCoinSelection;
 #[cfg(test)]
 pub type DefaultCoinSelectionAlgorithm = LargestFirstCoinSelection; // make the tests more predictable
+/// Algorithm to use in case of an error in the chosen coin selection algorithm
+// If everything fails, simple random selection should work
+pub type FallbackCoinSelectionAlgorithm = SingleRandomDrawCoinSelection;
```
danielabrozzoni: First, I agree with Murch that it would have been better to have this split in a different commit - but I understand it's not easy to see when to squash and when not to squash :) If you can, review would be much easier with two commits!

Second, this is a bit suboptimal: not every CS algorithm has a fallback algorithm! Instead, me and @afilini came up with this idea of having a `FallbackWrapper` algorithm struct (you should be able to copy-paste this code and it should already work):
```diff
diff --git a/src/wallet/coin_selection.rs b/src/wallet/coin_selection.rs
index 49bafa9..ef6139b 100644
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -84,12 +84,10 @@ use std::convert::TryInto;
 /// Default coin selection algorithm used by [`TxBuilder`](super::tx_builder::TxBuilder) if not
 /// overridden
 #[cfg(not(test))]
-pub type DefaultCoinSelectionAlgorithm = BranchAndBoundCoinSelection;
+pub type DefaultCoinSelectionAlgorithm =
+    FallbackWrapper<BranchAndBoundCoinSelection, SingleRandomDrawCoinSelection>;
 #[cfg(test)]
 pub type DefaultCoinSelectionAlgorithm = LargestFirstCoinSelection; // make the tests more predictable
-/// Algorithm to use in case of an error in the chosen coin selection algorithm
-// If everything fails, simple random selection should work
-pub type FallbackCoinSelectionAlgorithm = SingleRandomDrawCoinSelection;
 
 // Base weight of a Txin, not counting the weight needed for satisfying it.
 // prev_txid (32 bytes) + prev_vout (4 bytes) + sequence (4 bytes)
@@ -250,25 +248,13 @@ pub fn get_selection<D: Database, Cs: CoinSelectionAlgorithm<D>>(
         } else {
             // from now on, target_amount can only be positive
             let target_amount = (target_amount_i64 - required_effective_value) as u64;
-            let (mut selected_utxos, selected_fees) = match algorithm.coin_select(
+            let (mut selected_utxos, selected_fees) = algorithm.coin_select(
                 database,
                 optional_utxos.clone(),
                 fee_rate,
                 target_amount,
                 optional_effective_value,
-            ) {
-                Ok(selection) => selection,
-                Err(_err) => {
-                    let fallback_algorithm = FallbackCoinSelectionAlgorithm {};
-                    fallback_algorithm.coin_select(
-                        database,
-                        optional_utxos,
-                        fee_rate,
-                        target_amount,
-                        optional_effective_value,
-                    )?
-                }
-            };
+            )?;
             let selected_effective_value: i64 =
                 selected_utxos.iter().map(|og| og.effective_value).sum();
             selected_utxos.append(&mut required_utxos);
             @@ -344,6 +330,59 @@ pub trait CoinSelectionAlgorithm<D: Database>: std::fmt::Debug {
     ) -> Result<(Vec<OutputGroup>, u64), Error>;
 }
 
+#[derive(Debug)]
+pub struct FallbackWrapper<M, F> {
+    main: M,
+    fallback: F,
+}
+
+impl<M, F> Default for FallbackWrapper<M, F>
+where
+    M: Default,
+    F: Default,
+{
+    fn default() -> Self {
+        FallbackWrapper {
+            main: M::default(),
+            fallback: F::default(),
+        }
+    }
+}
+
+impl<M, F, D> CoinSelectionAlgorithm<D> for FallbackWrapper<M, F>
+where
+    M: CoinSelectionAlgorithm<D>,
+    F: CoinSelectionAlgorithm<D>,
+    D: Database,
+{
+    fn coin_select(
+        &self,
+        database: &D,
+        optional_utxos: Vec<OutputGroup>,
+        fee_rate: FeeRate,
+        target_amount: u64,
+        available_value: i64,
+    ) -> Result<(Vec<OutputGroup>, u64), Error> {
+        self.main
+            .coin_select(
+                database,
+                optional_utxos.clone(),
+                fee_rate,
+                target_amount,
+                available_value,
+                )
+            })
+    }
+}
+
 /// Simple and dumb coin selection
 ///
 /// This coin selection algorithm sorts the available UTXOs by value and then picks them starting
```

This is basically introducing a `FallbackWrapper`, which you can build around your favorite CS algorithms. We implemented CoinSelectionAlgorithm on the wrapper, so that you just need to call `coin_select`, and we automatically take care of calling the fallback in case of errors.

src/wallet/coin_selection.rs
```diff
@@ -171,6 +145,178 @@ impl CoinSelectionResult {
...
+/// - `required_utxos`: the utxos that must be spent regardless of `target_amount` with their
+///                     weight cost
+/// - `optional_utxos`: the remaining available utxos to satisfy `target_amount` with their
+///                     weight cost
+/// - `fee_rate`: fee rate to use
+/// - `target_amount`: the outgoing amount in satoshis and the fees already
+///                    accumulated from added outputs and transaction’s header.
+/// - `weighted_drain_script`: the drain script used to create the change output associated with
+///                            the max satisfaction weight needed to spend it
+pub fn get_selection<D: Database, Cs: CoinSelectionAlgorithm<D>>(
```
danielabrozzoni: Mh, I'd like a more explicative name. What about `process_and_select_coins`?

src/wallet/coin_selection.rs
```diff
@@ -171,6 +145,178 @@ impl CoinSelectionResult {
...
+        .collect();
+
+    // Mapping every (UTXO, usize) to an output group.
+    let optional_utxos: Vec<OutputGroup> = optional_utxos
+        .into_iter()
+        .map(|u| OutputGroup::new(u, fee_rate))
+        .filter(|u| u.effective_value.is_positive())
+        .collect();
+
+    let req_values = required_utxos.iter().fold((0, 0), |(eff_value, fees), x| {
```
danielabrozzoni: You can unpack this directly instead of doing it below:
```
let (required_effective_value, required_fees) = required_utxos.iter().etc...
```

danielabrozzoni (commented):
I gave a first quick pass, I'm not done reviewing :) (but I chose to publish the review so you can start working on these fixes!)

rajarshimaitra (commented):
Concept ACK..

I made a first shallow pass and for now just have few nit comments and name changing requests.. I will go through the full thing again once more..

src/wallet/coin_selection.rs
danielabrozzoni: I just noticed that calling `BranchAndBounCoinSelection::default()` is suboptimal (it's using a default `size_of_change`)... I'm working on finding a solution...

src/wallet/coin_selection.rs
```diff
@@ -641,32 +646,25 @@ impl BranchAndBoundCoinSelection {
...
-            selected_utxos,
-            required_utxos,
-            excess,
-        ))
+        Ok((selected_utxos, fee_amount))
     }
+}
+
+/// Single Random Draw coin selection
+#[derive(Debug)]
```
danielabrozzoni: Same as the other empty CS algs, this should derive `#[derive(Debug, Default, Clone, Copy)]`

src/wallet/mod.rs
csralvall: Didn't we consider the `drain_to` script pubkey as something external? the satisfaction weight for the wallet is zero because it won't spend the created output.

src/wallet/coin_selection.rs
csralvall: Me neither.

src/wallet/coin_selection.rs
csralvall: Great idea! `bnb` isn't the problem, `select_sorted_utxos` is, but I can rethink it to remove the tuple from the return value.

src/wallet/coin_selection.rs
csralvall: I may be missing something, but shouldn't be a call to `self.callback.coin_select` inside the impl of `coin_select` for `FallbackWrapper`?

src/wallet/coin_selection.rs
csralvall: Sure, that's better. I can proceed with your suggested name and change it again in the future if we find a better one.

src/wallet/mod.rs
csralvall: > Naming request: A name like `HandleExcess` could be more clearer here.. Its a simple enum, that specifies how to handle the change if any.

Well, I thought about that before and didn't like it because `Excess` is just a "thing" and not an action per se. The logic inside `create_tx` is the one that handles the `Excess`. That was my reasoning. If I didn't convince you, let me know and I will change it.

> Not a major issue, a general rust street rule is not to import enum variants directly. Call them with the enum name in code like `Excess::Change` etc. The name of the enum gives useful context at the usage location. And its always just two item long.

Good rule! I'll apply it.

src/types.rs
csralvall: Agree. I'll apply the changes locally to avoid conflicts (I've several changes in the backlog).

csralvall:
Thanks, @Xekyo, @danielabrozzoni, @rajarshimaitra, and @afilini for taking the time to review this. I've several changes ahead. Here is a list to keep you updated:
- [x] Split the addition of `get_selection` from changes to `CoinSelectionAlgorithm`s (`SingleRandomDraw`).
- [ ] Create `FallbackWrapper` (instead of `FallbackCoinSelectionAlgorithm`). ~It will probably go in a separate commit together with the split of `BranchAndBound`.~ It will included in a separate PR.
- [x] Address ~all~ the majority of nits.

src/wallet/mod.rs
danielabrozzoni: > Didn't we consider the drain_to script pubkey as something external?

Yeah

> the satisfaction weight for the wallet is zero because it won't spend the created output.

I admit I'm a bit undecided on this... But I think we should still consider the satisfaction weight here. Even tho it's not us spending the output, someone still will have to, so I think it's fair to try to reduce their waste.

src/wallet/coin_selection.rs
danielabrozzoni: Sorry, I probably copied it wrong 🤦🏻‍♀️ Yeah, it should be:
```diff
+impl<M, F, D> CoinSelectionAlgorithm<D> for FallbackWrapper<M, F>
+where
+    M: CoinSelectionAlgorithm<D>,
+    F: CoinSelectionAlgorithm<D>,
+    D: Database,
+{
+    fn coin_select(
+        &self,
+        database: &D,
+        optional_utxos: Vec<OutputGroup>,
+        fee_rate: FeeRate,
+        target_amount: u64,
+        available_value: i64,
+    ) -> Result<(Vec<OutputGroup>, u64), Error> {
+        self.main
+            .coin_select(
+                database,
+                optional_utxos.clone(),
+                fee_rate,
+                target_amount,
+                available_value,
+            )
+            .or_else(|_| {
+                self.fallback.coin_select(
+                    database,
+                    optional_utxos.clone(),
+                    fee_rate,
+                    target_amount,
+                    available_value,
+                )
+            })
+    }
+}
+
```

src/wallet/coin_selection.rs
murchandamus: Ah, I see. Thanks for explaining

src/wallet/mod.rs
csralvall: Ok, I've to find a way to compute the `satisfaction_weight` for an arbitrary `Script`, I've tried before without success.

src/wallet/coin_selection.rs
```diff
@@ -171,6 +145,178 @@ impl CoinSelectionResult {
...
+    // all their effective values, even when negative, which means that required_effective_value could
+    // be negative as well.
+    //
+    // If the sum of required_effective_value and optional_effective_value is negative or lower than our target,
+    // we can immediately exit with an error, as it's guaranteed we will never find a solution
+    // if we actually run the coin selection algorithm
+    let total_value: Result<u64, _> =
+        (optional_effective_value + required_effective_value).try_into();
+    match total_value {
+        Ok(v) if v >= target_amount => {}
```
evanlinjin, line 207: If `total_value == target_amount`, we could just skip coin selection right?

src/wallet/mod.rs
```diff
@@ -836,23 +838,40 @@ where
...
             required_utxos,
             optional_utxos,
             fee_rate,
             outgoing + fee_amount,
-            &drain_script,
+            &weighted_drain_script,
         )?;
+
         fee_amount += coin_selection.fee_amount;
         let excess = &coin_selection.excess;
```
evanlinjin: Nit: We can get rid of the `excess` variable.

src/wallet/coin_selection.rs
csralvall: Yes!

src/wallet/mod.rs
csralvall: Do you mean to replace all occurrences of `excess` with `coin_selection.excess`?

src/wallet/coin_selection.rs
csralvall: > I just noticed that calling `BranchAndBounCoinSelection::default()` is suboptimal (it's using a default `size_of_change`)... I'm working on finding a solution...

But the user can still pass `algorithm` as:
```rust
DefaultCoinSelectionAlgorithm {
   main: BranchAndBoundCoinCoinSelection::new(<user_set_size_of_change>),
   fallback: SingleRandomDrawCoinSelection::default()
}
```
This default is not giving up the flexibility above.

csralvall: I'll continue working with this in another PR.

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ page. See [DEVELOPMENT_CYCLE.md](DEVELOPMENT_CYCLE.md) for more details.
...
+- Add capacity to create FeeRate from sats/kvbytes and sats/kwu.
+- Rename `as_sat_vb` to `as_sat_per_vb`. Move all `FeeRate` test to `types.rs`.
+- Add custom Harware Wallet Signer `HwiSigner` in `src/wallet/harwaresigner/` module.
+- Added `process_and_select_coins()` function.
+- Added `OutputGroup`s as a public structure that holds the `WeightedUtxo`, its
+  effective value, and its fees.
+- Changed `CoinSelectionAlgorithm::coin_select()` signature. The scope of the
+  selection was reduced to select utxos only from `optional_utxos`. It only
+  returns the selected `OutputGroups` from now on.
+- Added `WeightedScriptPubkey` struct.
```
danielabrozzoni: We don't use the CHANGELOG.md anymore :) (see https://github.com/bitcoindevkit/bdk/pull/544). You should add a "### Changelog" subtitle in the PR description with the whole changelog.

src/wallet/coin_selection.rs
csralvall: This doesn't apply here anymore.

src/wallet/mod.rs
danielabrozzoni: How is this going? Do you need help?

csralvall: I'm not advancing in figuring out how to solve this. I've tried to implement a solution using miniscript functions without success. Another idea, which I don't like, is to pass the foreign weight as a parameter of the transaction.
Also, I don't want to implement something to parse the script and compute the weight. It would be pointless because miniscript implements those functions. But there isn't an implementation for an arbitrary scriptpubkey.

danielabrozzoni:
I don't think this will get into the milestone 0.23, I'm sorry :( I need some more time to review it and think about the API changes. This will introduce another API break in the `coin_selection` function, and I'd prefer to avoid further breakage in the future...

csralvall:
Sure, I understand.

evanlinjin (commented):
This is an excellent undertaking of trying to streamline coin selection in BDK.

However, there seems to be an attempt of "not changing too much" (which is understandable), but the transaction building model in BDK is inconvenient (and potentially wrong) and should be reworked from scratch (in my opinion).

Let's discuss tomorrow over the call and come up with a "grand plan". Your insights will be very appreciated (based on my judgement of your work).

NACK (for now) :cry:

danielabrozzoni:
After discussing this with the team, we came to the conclusion that we have to close this PR, as the `bdk_core` update makes it obsolete.

Just a reminder: even though this exact code won't be used in bdk, the ideas discovered while writing/reviewing it surely will. Thanks for your work, Cesar :)

Issue #524, opened 2022-01-11
Feerate should be returned from Transaction construction


LLFourn:
When you get TransactionDetails from `tx_builder.finish()` it has the fee for the overall transaction but does not include the feerate of the resulting transaction. There is no simple way for the caller to replicate this information since the transaction does not have witness data at this point. This is annoying because after constructing the psbt the user should know the feerate before signing it.

This can either be done by returning another value from `builder.finish()` or by adding a (perhaps optional) field to `TransactionDetails`.

csralvall:
Would make sense to save the overall `weight` or `size` of the transaction and the `fee rate` instead of the `fee amount`? I think that the `weight` would be useful in the `TransactionDetails`, and we can implement a method to compute the `fee amount`. WDYT?

thunderbiscuit:
Just want to comment to say that this is also a feature I'd use if implemented.

i5hi:
I am giving this issue a try and have a few thoughts:

Transaction size can be useful since you can calculate the fee_rate with it. 

When using create_tx does it make sense to have Some(Transaction) always returned so its easy to calculate size?

Or we can just have size as a field in TransactionDetail?

fee_rate helps too but feels repetitive since there is already a fee field.

LLFourn:
> When using create_tx does it make sense to have Some(Transaction) always returned so its easy to calculate size?

Even if you did that tx doesn't have the correct weight since it is missing signatures.

i5hi:
> > When using create_tx does it make sense to have Some(Transaction) always returned so its easy to calculate size?
> 
> Even if you did that tx doesn't have the correct weight since it is missing signatures.

The max satisfaction weight from the descriptor would give you that, right?

LLFourn:
> The max satisfaction weight from the descriptor would give you that, right?

If all the inputs are yours and you know which input had which descriptor. But in general no.

thunderbiscuit:
I think we're bumping into a related issue when we attempted to RBF bump a transaction on bdk-ffi. 

BIP 125 says that nodes will relay transactions for which the fees are bumped by _at least the amount of the minimum transaction relay rate_, i.e. if the minimum is 1 sat/vbyte your "currently paying" 3.0 sat/vbyte transaction will need to have a replacement transaction with at least 4.0 sat/vbyte to be related by nodes.

[...]

PR #728, opened 2022-08-21
Add fee_amount() and fee_rate() functions to PsbtUtils trait

notmandatory:
### Description

The purpose of the PR is to provide a more convenient way to calculate the transaction fee amount and fee rate for a PSBT. This PR adds `fee_amount` and `fee_rate` functions to the existing `PsbtUtils` trait and implements them for `PartiallySignedBitcoinTransaction`. The `fee_rate` value is only valid if the PSBT it is called on is fully signed and finalized. 

See related discussion: https://github.com/bitcoindevkit/bdk-ffi/issues/179

### Changelog

Added
- PsbtUtils.fee_amount(), calculates the PSBT total transaction fee amount in Sats.
- PsbtUtils.fee_rate(), calculates the PSBT FeeRate, the value is only accurate AFTER the PSBT is finalized.

### Notes to the reviewers

Ideally I'd like `fee_rate` to return an `Option` and return `None` if the PSBT isn't finalized. But I'm not quite sure how to determine if a PSBT is finalized without having a `Wallet` and running it through the finalize code first. Or there might be a way to fill in missing signatures with properly sized fake data prior to calculating the fee rate. For now I think it's enough to do this simple approach with usage warning in the rust docs.

#### All Submissions:

#### New Features:

### DIFF
--- a/src/psbt/mod.rs
+++ b/src/psbt/mod.rs
@@ -9,11 +9,22 @@
 // You may not use this file except in accordance with one or both of these
 // licenses.
 
+use crate::FeeRate;
 use bitcoin::util::psbt::PartiallySignedTransaction as Psbt;
 use bitcoin::TxOut;
 
 pub trait PsbtUtils {
     fn get_utxo_for(&self, input_index: usize) -> Option<TxOut>;
+
+    /// The total transaction fee amount, sum of input amounts minus sum of output amounts, in Sats.
+    /// If the PSBT is missing a TxOut for an input returns None.
+    fn fee_amount(&self) -> Option<u64>;
+
+    /// The transaction's fee rate. This value will only be accurate if calculated AFTER the
+    /// `PartiallySignedTransaction` is finalized and all witness/signature data is added to the
+    /// transaction.
+    /// If the PSBT is missing a TxOut for an input returns None.
+    fn fee_rate(&self) -> Option<FeeRate>;
 }
 
 impl PsbtUtils for Psbt {
@@ -37,6 +48,27 @@ impl PsbtUtils for Psbt {
             None
         }
     }
+
+    fn fee_amount(&self) -> Option<u64> {
+        let tx = &self.unsigned_tx;
+        let utxos: Option<Vec<TxOut>> = (0..tx.input.len()).map(|i| self.get_utxo_for(i)).collect();
+
+        utxos.map(|inputs| {
+            let input_amount: u64 = inputs.iter().map(|i| i.value).sum();
+            let output_amount: u64 = self.unsigned_tx.output.iter().map(|o| o.value).sum();
+            input_amount
+                .checked_sub(output_amount)
+                .expect("input amount must be greater than output amount")
+        })
+    }
+
+    fn fee_rate(&self) -> Option<FeeRate> {
+        let fee_amount = self.fee_amount();
+        fee_amount.map(|fee| {
+            let weight = self.clone().extract_tx().weight();
+            FeeRate::from_wu(fee, weight)
+        })
+    }
 }
 
 #[cfg(test)]
@@ -44,8 +76,9 @@ mod test {
     use crate::bitcoin::TxIn;
     use crate::psbt::Psbt;
     use crate::wallet::AddressIndex;
+    use crate::wallet::AddressIndex::New;
     use crate::wallet::{get_funded_wallet, test::get_test_wpkh};
-    use crate::SignOptions;
+    use crate::{psbt, FeeRate, SignOptions};
     use std::str::FromStr;
 
     // from bip 174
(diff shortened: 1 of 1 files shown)


src/psbt/mod.rs
```diff
@@ -37,15 +46,35 @@ impl PsbtUtils for Psbt {
             None
         }
     }
+
+    fn fee_amount(&self) -> u64 {
```
vladimirfomene: I'm wondering why the `output_amount` is calculated using the `unsigned_tx` outputs and not the PSBT `outputs`.

vladimirfomene (commented):
Tested ACK. Just have one question.

thunderbiscuit:
Linking bitcoindevkit/bdk_wallet#187 to this.

src/psbt/mod.rs
notmandatory: The PSBT and transactions inputs and outputs should all be the same. But your questions made me think about moving these functions to the TransactionDetails struct, or maybe as new util trait on Transaction. Will noodle around with that and see how it works out.

notmandatory: Also I just remembered the reason I needed to use PSBT to calculate the `fee_amount` is because it contains the full details for the transaction inputs, which includes the input amounts, if you only look at the Transaction structure it contains essentially "pointers" to the unspent outputs that become the inputs (the `OutPoint`) and this structure does not contain the input amounts.  So I'm leaving these helper functions on the `PsbtUtils` trait.

rajarshimaitra (commented):
Concept ACK. It will be helpful to get the feerate in the unsigned stage. 

And if I understand correctly, the ask of #524 is to have the fee rate in unsigned transaction specifically, which is a special annoying problem. As we don't know the size before fully signing it. Its straight forward to calculate the fee rate from a signed tx just from the transaction data itself. We know both the fee and the size after signing. 

I think one way could be to return the `FeeRate` in the `TransactionDetails` itself from `create_tx()` function. The fee rate value exists in the `create_tx()` function, we just happen to throw it away after coinselection. here
https://github.com/bitcoindevkit/bdk/blob/0a3734ed2b9aeebdc7f13e64c6544d35788a0699/src/wallet/mod.rs#L848-L855

So instead why not just keep the value alive and store it in the `TransactionDetails`.

Edge Case: If I understand correctly the interplay between `fee_rate` and `fee_amount` happening here 
https://github.com/bitcoindevkit/bdk/blob/0a3734ed2b9aeebdc7f13e64c6544d35788a0699/src/wallet/mod.rs#L743-L770

It seems if the user sets a `FeePolicy::FeeAmount(fee)` then `fee_rate` will always be 0, and in that case we should have that value as `None` in the `TrasnactionDetails`.

rajarshimaitra:
Another idea.. We have a `max_satisfaction_weight()` on the descriptor of the wallet.. We can use this in create transaction to cater  for the missing signature data and make a rough fee rate computation, in the case of the user providing `FeePolicy::FeeAmount(fee)`.  It might be off from the exact fee rate, but feels worth than not having it at all?? 

Bad idea: The wallet might have inputs for which it can't produce satisfaction.

notmandatory:
@rajarshimaitra you're right that the initial request was for an estimate of the fee rate before signing. Since I didn't see an easy way to do that so I went for giving the exact fee rate of the PSBT after it's finalized. I didn't think of using the `max_satisfaction_weight()`. I'm going to try that approach in a separate PR to add something like an "estimated_fee_rate" to the `TransactionDetails` result when `Wallet.create_tx()` is called.

rajarshimaitra:
@notmandatory  Noted.. I am thinking is it even useful to provide a fee rate method for signed PSBTs?? For a signed transactions its generally available readily from the PSBT data itself..

rajarshimaitra (approved):
@notmandatory Noted.. This can be done in separate PR..

tACK 7a9ed145ec7861ac02b27b6c1877168c004918c3

Although I am thinking if its much useful to provide this as a separate function, as its mostly straight forward calculation from any signed transaction data. But I guess having extra utility never hurts..

danielabrozzoni (approved):
utACK 7a9ed145ec7861ac02b27b6c1877168c004918c3

I've also added this PR to the milestone, as it seems super close to getting in :)

notmandatory:
BTW, I didn't add a CHANGELOG message since I'm assuming we'll have bitcoindevkit/bdk#544 in for this next release. If not I'll need to update CHANGELOG before this goes in.

src/psbt/mod.rs
```diff
@@ -37,15 +46,35 @@ impl PsbtUtils for Psbt {
...
+        let input_amount = &self
+            .inputs
+            .iter()
+            .fold(0, |acc, i| acc + i.witness_utxo.as_ref().unwrap().value);
```
afilini: There are a couple of things I would change here:

- First of all, sometimes you have the `non_witness_utxo` instead of the `witness_utxo`
- On top of that, it's not guaranteed that you'll have either one of them, a PSBT would still be valid. So `unwrap`ping here is probably not a good idea. I would suggest you make both methods return an `Option`, so that you can return `None` when it's not possible to calculate the fee because the PSBT is missing some metadata

afilini: For the first point, you can use `get_utxo_for()` which is designed specifically for this, when you need the `TxOut` that's being spent but you don't know if you have the `witness_utxo` or the `non_witness_utxo` available

afilini: Also, can you maybe add a test for all these cases?

- a psbt with witness_utxo
- one with non_witness_utxo
- one without any of them

notmandatory:
rebased to pickup ci and other changes

src/psbt/mod.rs
notmandatory: I'll work on these, but moved this PR to next release so more time to update and re-review.

src/psbt/mod.rs
```diff
@@ -37,15 +48,39 @@ impl PsbtUtils for Psbt {
             None
         }
     }
+
+    fn fee_amount(&self) -> Option<u64> {
+        let tx = &self.unsigned_tx;
+        let utxos: Option<Vec<TxOut>> = (0..tx.input.len()).map(|i| self.get_utxo_for(i)).collect();
+
+        utxos.map(|inputs| {
+            let input_amount = inputs.iter().fold(0, |acc, i| acc + i.value);
```
afilini: I think you can use [`.sum()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.sum) here instead of manually implementing it with `.fold()`.

src/psbt/mod.rs
```diff
@@ -37,15 +48,39 @@ impl PsbtUtils for Psbt {
...
+
+            let output_amount = &self
+                .unsigned_tx
+                .output
+                .iter()
+                .fold(0, |acc, o| acc + o.value);
```
afilini: Same thing about `sum()` here

src/psbt/mod.rs
```diff
@@ -37,15 +48,39 @@ impl PsbtUtils for Psbt {
...
+
+            input_amount - output_amount
+        })
+    }
+
+    fn fee_rate(&self) -> Option<FeeRate> {
+        let fee_amount = self.fee_amount();
+        let weight = self.clone().extract_tx().weight();
```
afilini: It's a small nit, but I would move this line inside the closure so that you end up cloning the whole PSBT only if `fee_amount` is `Some()`. Otherwise it's just wasted time, because you'll throw away the weight immediately

src/psbt/mod.rs
```diff
@@ -37,15 +48,39 @@ impl PsbtUtils for Psbt {
...
+        utxos.map(|inputs| {
+            let input_amount = inputs.iter().fold(0, |acc, i| acc + i.value);
+
+            let output_amount = &self
+                .unsigned_tx
+                .output
+                .iter()
+                .fold(0, |acc, o| acc + o.value);
+
+            input_amount - output_amount
```
afilini: Just to be extra careful I would either add an assertion that `input_amount` is greater than or equal `output_amount`, or you could use `checked_sub().expect()` to achieve the same result.

It should never happen, but PSBTs are untrusted input so I think it's better to be careful.

src/psbt/mod.rs
notmandatory: ugh 🤦 thanks  fixed!

src/psbt/mod.rs
notmandatory: fixed

src/psbt/mod.rs
notmandatory: Oh ya good idea, fixed.

src/psbt/mod.rs
notmandatory: Safety first, fixed.

afilini (approved):
ACK ab41679368be79e8645915b40be554cc07557931

I've also checked the coveralls report and it looks like all the added lines are actually covered by tests! I mean, I was expecting it but it's nice to see the report match the expectation :)

Issue #372, opened 2021-06-17
Better address setup


afilini:
**Description**

Instead of using the derivation index we should store for each address whether or not it has been "setup" with the blockchain backend. We could merge `Blockchain::setup()` with `Blockchain::sync()` and just let the function decide what to do with each address based on its status.

Probably depends on #359 since we would have to change the db interface.

**Expected Outcomes** 
* Update blockchain module traits and structs as needed and fix corresponding tests
* Add add or combine tests if needed to cover updated functionality 
* Update any modified public traits and structs documentation 

**Resources**  
* [BDK Website](https://bitcoindevkit.org)
* [BDK github repo](https://github.com/bitcoindevkit/bdk)
* [BDK Discord](https://discord.gg/dstn4dQ)

**Skills Required**  
* Experience with git. [Guide](https://git-scm.com/book/en/v2)
* Basic familiarity with rust. [First seven chapters of the book](https://doc.rust-lang.org/book/)
* Familiarity with bitcoin descriptors and address generation
* Familiarity with changes made for schema migration in #359

**Mentor(s)**  
@notmandatory 

**Difficulty**  
Medium

**Competency Test**  
* Install rust, compile and run all bdk examples and tests. 
* Read through the BDK docs.
* Make a dummy wallet with BDK with key-value and sqlite backends. 
* Familiarity with basic rust, should be able to write basic custom trait implementations on foreign structures.

notmandatory:
Per our team chat yesterday I changed this into a summer-of-bitcoin project.

LLFourn:
I think this one needs to be expanded up on since I don't know what I would do with it if I had to tackle it. In particular I'm not totally clear on the motivation and what the result should look like.

notmandatory:
The genesis of this issue is #354. 

My understanding is when addresses are requested (ie. `get_address`) prior to a `sync` and then `setup` isn't called. But by keeping track on individual addresses if they're setup or not at least the first part of #354 is solved since they can be setup if needed every time sync is called. 

In my mind this change will create a table for all `get_address` scripts generated (or at least their indexes) with a boolean indicating if they are "setup" or not. Then when sync is called a query will be done to find any not-setup scripts, and if any are found call setup on them. I think the reason to not setup each script as it's requested is that it would be a performance hit. Also we allow random indexes to be requested so keeping only a range of index not setup wouldn't work.

But maybe there's a better way to address the two issues from #354 and we can explore that with whoever takes this on.

[...]

PR #740, opened 2022-09-01
Make `RpcBlockchain` import more efficient

### LINKED ISSUE #715: Large `importdescriptors` call slow, possible to reduce number of watched addresses?
BDK is issuing a large importdescriptors call, which is turning out to be very slow. See [this issue for some of the gruesome details](https://github.com/bitcoin/bitcoin/issues/25724).

I think the call is large because BDK is asking Bitcoin Core to watch 100 addresses. Is there a way to reduce this number? I don't see it in the RpcBlockchain options.

evanlinjin:
Fixes #715

### Description

Introduces `RpcImportParams` that keeps track of next derivation index to start import from for next sync. This avoids re-importing scripts/descriptors into Bitcoin Core for every sync cycle. However, the user will be responsible for persisting the `RpcImportParams` structure; the `RpcBlockchain::get_import_parameters` and `RpcBlockchain::replace_import_parameters` functions can be used for this purpose.

Also add pagination for importing scripts/descriptors.

Also had to make `clippy` happy by deriving `Eq` for various structures.

### Notes to the reviewers

Refer to: https://github.com/bitcoindevkit/bdk/issues/715#issuecomment-1233656187

### Changelog notice

* Introduce `RpcImportParams` to make sync more efficient
* Introduce `RpcSyncParams::page_size` to restrict request/response array size of RPC calls

#### All Submissions:

#### New Features:

#### Bugfixes:

~* [ ] I've added tests to reproduce the issue which are now passing~ (This is a performance bug fix)

### DIFF
--- a/src/blockchain/rpc.rs
+++ b/src/blockchain/rpc.rs
@@ -77,7 +77,7 @@ impl Deref for RpcBlockchain {
 }
 
 /// RpcBlockchain configuration options
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
 pub struct RpcConfig {
     /// The bitcoin node url
     pub url: String,
@@ -96,7 +96,7 @@ pub struct RpcConfig {
 /// In general, BDK tries to sync `scriptPubKey`s cached in [`crate::database::Database`] with
 /// `scriptPubKey`s imported in the Bitcoin Core Wallet. These parameters are used for determining
 /// how the `importdescriptors` RPC calls are to be made.
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
 pub struct RpcSyncParams {
     /// The minimum number of scripts to scan for on initial sync.
     pub start_script_count: usize,
@@ -373,7 +373,7 @@ impl<'a, D: BatchDatabase> DbState<'a, D> {
 
     /// Sync states of [BatchDatabase] and Core wallet.
     /// First we import all `scriptPubKey`s from database into core wallet
-    fn sync_with_core(&mut self, client: &Client, is_descriptor: bool) -> Result<&mut Self, Error> {
+    fn sync_with_core(&mut self, client: &Client, use_desc: bool) -> Result<&mut Self, Error> {
         // this tells Core wallet where to sync from for imported scripts
         let start_epoch = if self.params.force_start_time {
             self.params.start_time
@@ -385,7 +385,7 @@ impl<'a, D: BatchDatabase> DbState<'a, D> {
 
         // sync scriptPubKeys from Database to Core wallet
         let scripts_iter = self.ext_spks.iter().chain(&self.int_spks);
-        if is_descriptor {
+        if use_desc {
             import_descriptors(client, start_epoch, scripts_iter)?;
         } else {
             import_multi(client, start_epoch, scripts_iter)?;
--- a/src/blockchain/rpc.rs
+++ b/src/blockchain/rpc.rs
@@ -27,6 +27,7 @@
 //!     network: bdk::bitcoin::Network::Testnet,
 //!     wallet_name: "wallet_name".to_string(),
 //!     sync_params: None,
+//!     import_params: None,
 //! };
 //! let blockchain = RpcBlockchain::from_config(&config);
 //! ```
@@ -47,8 +48,9 @@ use bitcoincore_rpc::json::{
 use bitcoincore_rpc::jsonrpc::serde_json::{json, Value};
 use bitcoincore_rpc::Auth as RpcAuth;
 use bitcoincore_rpc::{Client, RpcApi};
-use log::{debug, info};
+use log::{debug, info, warn};
 use serde::{Deserialize, Serialize};
+use std::cell::RefCell;
 use std::collections::{HashMap, HashSet};
 use std::ops::Deref;
 use std::path::PathBuf;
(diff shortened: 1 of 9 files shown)


notmandatory:
The idea of paging makes good sense, but is there a reason to not save the next derivation index in the existing wallet DB as we do with other sync related data?

evanlinjin:
> The idea of paging makes good sense, but is there a reason to not save the next derivation index in the existing wallet DB as we do with other sync related data? 

Thanks for the comment @notmandatory .

The "next derivation index" in this context means the index after the last imported into Bitcoin Core. Ideally, we should be able to obtain this information easily from Bitcoin Core itself, but the RPC interface does not allow it.

This is different to how "next derivation index" should be interpreted in `Database`, which is "the next index in which we get a fresh address".

The ideal would be to have the two "next derivation index" concepts in sync. This would be possible if "user-get-new-address" and "import-new-address-into-bitcoin-core" could be performed atomically. Another solution would be to mark "imported-into-bitcoin-core" addresses in the `Database`. Both these solutions are non-viable (due to the restrictions of the "model" of BDK).

So in essence, I did this hack because: restrictive RPC interface in Core + restrictive "bounds" of the current BDK "model".

src/blockchain/rpc.rs
```diff
@@ -710,6 +786,38 @@ where
     Ok(())
 }
 
+fn pagenated_import<'a, S>(
```
afilini: I think the correct term is "paginated"

src/blockchain/rpc.rs
```diff
@@ -710,6 +786,38 @@ where
...
+    start_epoch: u64,
+    poll_rate_sec: u64,
+    page_size: usize,
+    scripts_iter: S,
+    progress: &dyn Progress,
+) -> Result<(), Error>
+where
+    S: Iterator<Item = &'a Script> + Clone,
+{
+    (0_usize..)
```
afilini, line 801: Alternative option, if you like it better: `(0_usize..).step_by(page_size)`. So you can iterate over the start index for each page

src/blockchain/rpc.rs
```diff
@@ -710,6 +786,38 @@ where
...
+        .map(|page_index| {
+            scripts_iter
+                .clone()
```
afilini, line 805: Does this clone all the elements in the iterator? If that's the case, maybe we should replace these few lines with a custom iterator that just consumes `scripts_iter` once and returns the item in batches.

src/blockchain/rpc.rs
```diff
@@ -66,6 +68,8 @@ pub struct RpcBlockchain {
     capabilities: HashSet<Capability>,
     /// Sync parameters.
     sync_params: RpcSyncParams,
+    /// Import parameters.
+    import_params: RefCell<RpcImportParams>,
```
afilini, line 72: Just to be safe I would replace this with an `Arc<RpcImportParams>`. Then you can use `Arc::make_mut()` whenever you want to replace it

afilini:
I think #372 would fix the same issue this PR tries to tackle, because it would allow us to save in the database which addresses have already been imported. It's been blocked all this time because we haven't figured out a good way to do database migrations, but maybe it's time we get back to it.

Also, a suggestion: maybe you don't remember, but the old rpc implementation used the label of a fixed address as a "storage section" to save essentially the `RpcImportParams`. What do you think about moving back to that model? Keeping your current architecture, but instead of asking the user to persist that structure, we just do it automatically with that hack. It would also have the added benefit of automatically resetting if the user switches to a clean node, otherwise that requires manual intervention (manually marking the node as "clean" so that bdk reimports everything)

notmandatory:
I'm putting this on the list to discuss tomorrow, but since it looks like it requires more discussion will probably have to remove it from the 0.23.0 feature freeze.

src/blockchain/rpc.rs
evanlinjin: The iterator for `Vec` uses pointers so I think it is okay. Also the elements here are references too.

```rust
#[stable(feature = "rust1", since = "1.0.0")]
#[must_use = "iterators are lazy and do nothing unless consumed"]
pub struct Iter<'a, T: 'a> {
    ptr: NonNull<T>,
    end: *const T, // If T is a ZST, this is actually ptr+len.  This encoding is picked so that
    // ptr == end is a quick test for the Iterator being empty, that works
    // for both ZST and non-ZST.
    _marker: PhantomData<&'a T>,
}
```

src/blockchain/rpc.rs
evanlinjin: This won't work as the `WalletSync` is NOT `&mut self`. We can do `Arc<Mutex<RpcImportParams>>`? But I think this is too much. I feel like the user can just handle thread-safety elsewhere?

src/blockchain/rpc.rs
evanlinjin: Done.

evanlinjin:
> Also, a suggestion: maybe you don't remember, but the old rpc implementation used the label of a fixed address as a "storage section" to save essentially the RpcImportParams. What do you think about moving back to that model?

@afilini I am currently unconvinced that the previous model is a good idea. It will cause problems if a different `Wallet` with the same descriptors attempts to sync with the same Bitcoin Core (scriptPubKeys will be skipped). Also it is not elegant.

Even with what I have in the PR, and if the user does not persist `RpcSyncParams`, only the first sync of the `RpcBlockchain` lifetime will be inefficient. I think this in itself is decent enough.

> I think https://github.com/bitcoindevkit/bdk/issues/372 would fix the same issue this PR tries to tackle, because it would allow us to save in the database which addresses have already been imported. It's been blocked all this time because we haven't figured out a good way to do database migrations, but maybe it's time we get back to it.

It does sound like this would fix the issue, but I think #372 would be more elegant with "birthdays" of scriptPubKeys. So database will store birthdays of scriptPubKeys and "last-sync" for each different "blockchain". This way we can compare which scriptPubKeys we need to import, without having to do a map of which scriptPubKey has been synced to which "blockchains".

evanlinjin:
BDK Chain will replace the need for this

PR #707, opened 2022-08-09
Retry failed RPC calls on error

afilini:
### Description

Whenever the transport errors with "WouldBlock" we wait a bit and retry the call.

This partially fixes #650, at least the errors happening with the RPC blockchain reported here: https://github.com/bitcoindevkit/bdk/issues/650#issuecomment-1182990519

### Notes to the reviewers

I'm a bit puzzled because in my tests I still have failures, but this time with the error:

```
Rpc(JsonRpc(Rpc(RpcError { code: -4, message: "Wallet is currently rescanning. Abort existing rescan or wait.", data: None })))
```

even though the wallet have just been created. I even tried adding another "await_wallet_scan" before our `importdescriptors` and it didn't make any difference.

To consistently reproduce this issue I ran the tests in a large (>=16 vCPUs) instance on AWS.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [ ] I've added tests for the new feature
* [ ] I've added docs for the new feature
* [ ] I've updated `CHANGELOG.md`

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [x] I'm linking the issue being fixed by this PR

PR #741, opened 2022-09-01
Retry failed RPC calls on error

### LINKED ISSUE #650: CI: `blockchain::esplora::bdk_blockchain_tests::test_sync_stop_gap_20` occationally fails
**Describe the bug**  

CI: `blockchain::esplora::bdk_blockchain_tests::test_sync_stop_gap_20` occationally fails.

**The command that fails**

```bash
$ cargo test --no-default-features --features test-esplora,use-esplora-ureq,verify esplora::bdk_blockchain_tests
```

**Occurences**
- https://github.com/bitcoindevkit/bdk/runs/7163681328?check_suite_focus=true

evanlinjin:
This PR replaces #707 (which attempted to fix #650), but is based on top of work done in #740.

### Description

Refer to #707 and #740

I don't think we can ever fix #650 fully, however the work done in this PR seems to reduce the failure rate.

In addition to the work done by @afilini in #707, this PR introduces logic that catches error code -4 during descriptor/scriptPubKey import, and treats it like a successful import (so we continue syncing as usual). The caveat of this is that some sync cycles will miss out on importing some scriptPubKeys/descriptors, but is this better than a complete failure?

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/blockchain/any.rs
+++ b/src/blockchain/any.rs
@@ -194,7 +194,7 @@ impl_from!(boxed rpc::RpcBlockchain, AnyBlockchain, Rpc, #[cfg(feature = "rpc")]
 /// );
 /// # }
 /// ```
-#[derive(Debug, serde::Serialize, serde::Deserialize, Clone, PartialEq)]
+#[derive(Debug, serde::Serialize, serde::Deserialize, Clone, PartialEq, Eq)]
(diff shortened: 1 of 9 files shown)


evanlinjin:
@afilini do you think this is still worth working on?

PR #744, opened 2022-09-02
Add psbt_signer.rs example

notmandatory:
### Description

Adding a simple example of how to create a PSBT with a watch only wallet and then sign it with a signing wallet. 

### Notes to the reviewers

This example was inspired by a question from a user.

### Changelog notice

none.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/examples/psbt_signer.rs
+++ b/examples/psbt_signer.rs
@@ -1,0 +1,102 @@
+// Copyright (c) 2020-2022 Bitcoin Dev Kit Developers
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+use bdk::blockchain::{Blockchain, ElectrumBlockchain};
+use bdk::database::MemoryDatabase;
+use bdk::wallet::AddressIndex;
+use bdk::SyncOptions;
+use bdk::{FeeRate, SignOptions, Wallet};
+use bitcoin::{Address, Network};
+use electrum_client::Client;
+use std::error::Error;
+use std::str::FromStr;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // test keys created with `bdk-cli key generate` and `bdk-cli key derive` commands
+    let signing_external_descriptor = "wpkh([e9824965/84'/1'/0']tprv8fvem7qWxY3SGCQczQpRpqTKg455wf1zgixn6MZ4ze8gRfHjov5gXBQTadNfDgqs9ERbZZ3Bi1PNYrCCusFLucT39K525MWLpeURjHwUsfX/0/*)";
+    let signing_internal_descriptor = "wpkh([e9824965/84'/1'/0']tprv8fvem7qWxY3SGCQczQpRpqTKg455wf1zgixn6MZ4ze8gRfHjov5gXBQTadNfDgqs9ERbZZ3Bi1PNYrCCusFLucT39K525MWLpeURjHwUsfX/1/*)";
+
+    let watch_only_external_descriptor = "wpkh([e9824965/84'/1'/0']tpubDCcguXsm6uj79fSQt4V2EF7SF5b26zCuG2ZZNsbNQuw5G9YWSJuGhg2KknQBywRq4VGTu41zYTCh3QeVFyBdbsymgRX9Mrts94SW7obEdqs/0/*)";
+    let watch_only_internal_descriptor = "wpkh([e9824965/84'/1'/0']tpubDCcguXsm6uj79fSQt4V2EF7SF5b26zCuG2ZZNsbNQuw5G9YWSJuGhg2KknQBywRq4VGTu41zYTCh3QeVFyBdbsymgRX9Mrts94SW7obEdqs/1/*)";
+
+    // create client for Blockstream's testnet electrum server
+    let blockchain =
+        ElectrumBlockchain::from(Client::new("ssl://electrum.blockstream.info:60002")?);
+
+    // create watch only wallet
+    let watch_only_wallet: Wallet<MemoryDatabase> = Wallet::new(
+        watch_only_external_descriptor,
+        Some(watch_only_internal_descriptor),
+        Network::Testnet,
+        MemoryDatabase::default(),
+    )?;
+
+    // create signing wallet
+    let signing_wallet: Wallet<MemoryDatabase> = Wallet::new(
+        signing_external_descriptor,
+        Some(signing_internal_descriptor),
+        Network::Testnet,
+        MemoryDatabase::default(),
+    )?;
+
+    println!("Syncing watch only wallet.");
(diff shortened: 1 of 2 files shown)


notmandatory:
@vladimirfomene please take a look and see if it works for you.

vladimirfomene (commented):
Tested ACK

vladimirfomene (commented):
Tested ACK

examples/psbt_signer.rs
```diff
@@ -0,0 +1,102 @@
...
+    )?;
+
+    #[cfg(feature = "electrum")]
+    {
+        println!("Sync watch only wallet");
+        watch_only_wallet.sync(&blockchain, SyncOptions::default())?;
+    }
+
+    // get deposit address
+    let deposit_address = watch_only_wallet.get_address(AddressIndex::New).unwrap();
```
danielabrozzoni: nit: you should probably `?` here, instead of unwrapping

examples/psbt_signer.rs
```diff
@@ -0,0 +1,102 @@
...
+    let balance = watch_only_wallet.get_balance()?;
+    println!("Watch only wallet balance in SAT: {}", balance);
+
+    //Create the PSBT sending 10,000 SATs back to the same wallet
+    let send_to_address = watch_only_wallet.get_address(AddressIndex::New)?;
+    let mut builder = watch_only_wallet.build_tx();
+    builder
+        .add_recipient(send_to_address.script_pubkey(), 10_000)
+        .enable_rbf()
+        .do_not_spend_change()
```
danielabrozzoni: Do we need this? This adds all the change outputs to the "unspendable" list (see https://docs.rs/bdk/latest/bdk/wallet/tx_builder/struct.TxBuilder.html#method.do_not_spend_change), I fear that adding it in a example without a real reason for doing so might confuse users, and they might just copy it and then wonder why they can't spend most of their funds...

examples/psbt_signer.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    // get deposit address
+    let deposit_address = watch_only_wallet.get_address(AddressIndex::New).unwrap();
+    println!(
+        "Fund watch only wallet by sending testnet bitcoin to this address: {addr}\nExplorer URL: https://mempool.space/testnet/address/{addr}",
+        addr = deposit_address.address
+    );
+
+    let balance = watch_only_wallet.get_balance()?;
```
danielabrozzoni: I would modify this code a bit, to make the example easier to use:

First, check the balance, if it's less than ~10_200 sats (you're sending 10_000 sats, I'm adding a 200 sats margin for fees), print the message asking the user to fund the wallet, and exit. Otherwise, continue with creating the transaction.

examples/psbt_signer.rs
```diff
@@ -0,0 +1,102 @@
...
+    println!("Watch only wallet balance in SAT: {}", balance);
+
+    //Create the PSBT sending 10,000 SATs back to the same wallet
+    let send_to_address = watch_only_wallet.get_address(AddressIndex::New)?;
```
danielabrozzoni: nit: maybe print the send_to_address in as well, so that we know how to recover the funds :) Otherwise, we might just want to hardcode here a testnet faucet address.

danielabrozzoni (commented):
Concept ACK

It looks good to me, I think it should also be added to the `Cargo.toml` in the examples section :)

examples/psbt_signer.rs
notmandatory: Good point, I removed it, was only there from the test code I was originally asked to look at.

examples/psbt_signer.rs
notmandatory: Another good idea ! When wallet balance < required I show the user the URL to uo1.net testnet faucet pre-prepopulated with address to send to (couldn't prepopulate the amount 🤷 ), when balance unconfirmed I show an message to wait, and when adequate balance confirmed voilà PSBT signed and broadcast.

examples/psbt_signer.rs
notmandatory: Done! do the new messages I added make sense?

notmandatory:
I incorporated the suggested changes and also added a list of any unconfirmed tx if the wallet balance is less than the desired 10k sats. Ready to re-review! 🙏

danielabrozzoni (approved):
Looking great!

ACK fa998de4b1849adcf43351dbec05b3de43e6694e

examples/psbt_signer.rs
```diff
@@ -0,0 +1,102 @@
...
+            );
+        }
+    } else {
+        println!("Creating a PSBT sending 9800 SATs plus fee to the u01.net testnet faucet return address 'tb1ql7w62elx9ucw4pj5lgw4l028hmuw80sndtntxt'.");
+        let return_address = Address::from_str("tb1ql7w62elx9ucw4pj5lgw4l028hmuw80sndtntxt")?;
+        let mut builder = watch_only_wallet.build_tx();
+        builder
+            .add_recipient(return_address.script_pubkey(), 9_800)
+            .enable_rbf()
+            .fee_rate(FeeRate::from_sat_per_vb(1.0));
```
afilini, line 82: I would use `.drain_wallet().drain_to(return_address.script_pubkey())` to make sure a change is not created. It's also a chance to show that BDK can do more than "just send x to this address".

afilini:
I was also about to suggest using the `descriptor!()` macro but decided not to because I think it could be a bit overwhelming at first. Most users will probably have a string for their descriptor and this would look more familiar to them.

examples/psbt_signer.rs
notmandatory: That's a good idea but I do like having some coin selection and change going on, which I think will be how most new projects will be doing rather than sweeping all funds.

danielabrozzoni: Now that I think about it, the problem of receiving 10_000 and sending 9_800 is that you create a small change (~150 sats?), which will be uneconomical to spend (the fees for spending it are greater than the utxo value) - which BDK's coin selection won't pick, unless it's explicitly instructed to do so.

What this means is: after running this example many times, we might have the wallet full of these rather small utxos which we can't spend, and users will start to see some weird errors: for example, they might see that the balance is 10_000 (~70 utxos, ~150 each), but the coin selection will fail saying that there are not enough funds.

We can prevent this either by sweeping all funds every time, or by asking users to fill the wallet with 10_000 sats, but then spending only a smaller number, like 5_000

danielabrozzoni: Oh, nevermind, the CS is smart enough that it picks 200 sats as fee, leaving no residue in the wallet :)

notmandatory: Yes I guess also in my testing it used the change for fee and left an empty wallet balance. Yea for BDK coin selection! So we think it's ok to do it this way and not use `drain_to` to ensure there's no change?

danielabrozzoni: I think it's fine :)

afilini: Yeah the end result should be the same because the leftover is lower than the dust threshold. I can go ahead and merge it like this, my suggestion was based on what I would have done personally, but it also looks good like this.

afilini (approved):
ACK fa998de4b1849adcf43351dbec05b3de43e6694e

PR #745, opened 2022-09-07
Add tests to improve coverage

vladimirfomene:
### Description

This PR add more test to the database module and also fixes certain bugs discovered by the written test. I also amended the name used for the database parameter in the test functions.

### Notes to the reviewers

This contributes to fixing bitcoindevkit/bdk_wallet#179 

#### All Submissions:

### DIFF
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -217,6 +217,7 @@ pub mod test {
     use std::str::FromStr;
 
     use bitcoin::consensus::encode::deserialize;
+    use bitcoin::consensus::serialize;
     use bitcoin::hashes::hex::*;
     use bitcoin::*;
 
@@ -322,8 +323,24 @@ pub mod test {
     }
 
     pub fn test_raw_tx<D: Database>(mut tree: D) {
-        let hex_tx = Vec::<u8>::from_hex("0100000001a15d57094aa7a21a28cb20b59aab8fc7d1149a3bdbcddba9c622e4f5f6a99ece010000006c493046022100f93bb0e7d8db7bd46e40132d1f8242026e045f03a0efe71bbb8e3f475e970d790221009337cd7f1f929f00cc6ff01f03729b069a7c21b59b1736ddfee5db5946c5da8c0121033b9b137ee87d5a812d6f506efdd37f0affa7ffc310711c06c7f3e097c9447c52ffffffff0100e1f505000000001976a9140389035a9225b3839e2bbf32d826a1e222031fd888ac00000000").unwrap();
-        let tx: Transaction = deserialize(&hex_tx).unwrap();
+        let hex_tx = Vec::<u8>::from_hex("02000000000101f58c18a90d7a76b30c7e47d4e817adfdd79a6a589a615ef36e360f913adce2cd0000000000feffffff0210270000000000001600145c9a1816d38db5cbdd4b067b689dc19eb7d930e2cf70aa2b080000001600140f48b63160043047f4f60f7f8f551f80458f693f024730440220413f42b7bc979945489a38f5221e5527d4b8e3aa63eae2099e01945896ad6c10022024ceec492d685c31d8adb64e935a06933877c5ae0e21f32efe029850914c5bad012102361caae96f0e9f3a453d354bb37a5c3244422fb22819bf0166c0647a38de39f21fca2300").unwrap();
+        let mut tx: Transaction = deserialize(&hex_tx).unwrap();
+
+        tree.set_raw_tx(&tx).unwrap();
+
+        let txid = tx.txid();
+
+        assert_eq!(tree.get_raw_tx(&txid).unwrap(), Some(tx.clone()));
+
+        // mutate transaction's witnesses
+        for tx_in in tx.input.iter_mut() {
+            tx_in.witness = Witness::new();
+        }
+
+        let updated_hex_tx = serialize(&tx);
+
+        // verify that mutation was successful
+        assert_ne!(hex_tx, updated_hex_tx);
 
         tree.set_raw_tx(&tx).unwrap();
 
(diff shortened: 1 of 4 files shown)


danielabrozzoni:
I'd like to see b91afdb49bcec89007b254883fa3c21420e2f382 split into multiple commits: at least one different commit for each bug discovered and fixed, with an explanation of the bug found. You modified quite a bit of code in `src/database/keyvalue.rs` and `src/database/memory.rs`, but without a clear commit message it's nearly to impossible to understand your intentions.

vladimirfomene:
@danielabrozzoni, I have implemented the changes requested and explained reason for the change in the commit message.

src/database/mod.rs
```diff
@@ -441,5 +458,180 @@ pub mod test {
...
+
+        let res = db.del_path_from_script_pubkey(&script).unwrap();
+
+        assert!(res.is_none());
+
+        let _res = db.set_script_pubkey(&script, keychain, path);
+        let (chain, child) = db.del_path_from_script_pubkey(&script).unwrap().unwrap();
+
+        assert_eq!(chain, keychain);
+        assert_eq!(path, child);
```
danielabrozzoni: I think you should also iter script pub keys to check if they were actually deleted from the db
(and do the same for the others `del_` tests)

src/database/mod.rs
```diff
@@ -441,5 +458,180 @@ pub mod test {
...
+        assert_eq!(res, 0);
+
+        let _res = db.increment_last_index(keychain);
+
+        let res = db.del_last_index(keychain).unwrap().unwrap();
+
+        assert_eq!(res, 1);
+    }
+
+    pub fn test_check_descriptor_checksum<D: Database>(mut db: D) {
```
danielabrozzoni, line 639: I think you should add a couple of comments here to make the test more understandable... If I understand correctly, the first check_descriptor_checksum sets the checksum to "1cead456", and the second one compares the stored checksum with "1cead454", and returns an error. Does that make sense?

src/database/sqlite.rs
```diff
@@ -467,7 +467,7 @@ impl SqliteDatabase {
         &self,
         txid: &[u8],
     ) -> Result<Option<TransactionDetails>, Error> {
-        let mut statement = self.connection.prepare_cached("SELECT transaction_details.timestamp, transaction_details.received, transaction_details.sent, transaction_details.fee, transaction_details.height, transactions.raw_tx FROM transaction_details, transactions WHERE transaction_details.txid=transactions.txid AND transaction_details.txid=:txid")?;
+        let mut statement = self.connection.prepare_cached("SELECT timestamp, received, sent, fee, height FROM transaction_details WHERE txid=:txid")?;
```
danielabrozzoni, line 470: I'd modify the commit message of 27c6735d122ea7ef61f19e1eeac23643ccc096e8: 
The main commit line shouldn't be "Fix failing test: test_del_tx" : the purpose of this commit is to fix a bug you've found, not to fix a test (the commit doesn't even have lines changing tests, so it's definitely not fixing a test) :)

danielabrozzoni (commented):
First of all, some styling nits:
1. You shouldn't put "Fix failing test: test_del_tx" as the commit message for 27c6735d122ea7ef61f19e1eeac23643ccc096e8 - your commit is not fixing a test. Your commit is fixing a bug you've found. This is very different, as "Fix failing test" makes the reader think that in the commit you are modifying the actual test behavior.
2. This is a quite big nit, but I usually first put commits that fix bugs and then commits that add tests, not the other way around (in this PR d8f502660f30327f5ffe1ef6a961641bdfad539e is before 27c6735d122ea7ef61f19e1eeac23643ccc096e8). The reason for that is: it helps me write a clearer commit message, as that I'm not tempted to cite any test in the bug-fixing commit, but also it makes the git history bisectable, which is always a nice to have.

The added tests look good and they actually improve the code coverage :) See https://coveralls.io/builds/52344495

I am a bit confused about 27c6735d122ea7ef61f19e1eeac23643ccc096e8: to me, it doesn't seem that you need to modify `select_transaction_details_by_txid`, you can just remove the transaction from  the returned object if `include_raw` is false:
```
diff --git a/src/database/sqlite.rs b/src/database/sqlite.rs
index 562c934..4fb6337 100644
--- a/src/database/sqlite.rs
+++ b/src/database/sqlite.rs
@@ -744,11 +744,16 @@ impl BatchOperations for SqliteDatabase {
         include_raw: bool,
     ) -> Result<Option<TransactionDetails>, Error> {
         match self.select_transaction_details_by_txid(txid)? {
-            Some(transaction_details) => {
+            Some(mut transaction_details) => {
                 self.delete_transaction_details_by_txid(txid)?;
 
                 if include_raw {
                     self.delete_transaction_by_txid(txid)?;
+                } else {
+                    // Since we want to return what we actually deleted from
+                    // the database, we make sure to put a "None" in the
+                    // transaction field here.
+                    transaction_details.transaction = None;
                 }
                 Ok(Some(transaction_details))
             }
```

Is there something I'm not seeing?

vladimirfomene:
@danielabrozzoni, initially I made the transaction field `None` on the returned object and yes it fixed the bug. The other issue here is that if I was calling `get_tx` with `include_raw = false`, behind the scene `select_transaction_details_by_txid` will get the transaction even though it is not needed as it does a join query on `transactions` and `transactiondetails`.

src/database/mod.rs
vladimirfomene: Yes, your thinking about the logic is right. I will add comments here.

src/database/sqlite.rs
```diff
@@ -467,7 +467,7 @@ impl SqliteDatabase {
         &self,
         txid: &[u8],
     ) -> Result<Option<TransactionDetails>, Error> {
-        let mut statement = self.connection.prepare_cached("SELECT transaction_details.timestamp, transaction_details.received, transaction_details.sent, transaction_details.fee, transaction_details.height, transactions.raw_tx FROM transaction_details, transactions WHERE transaction_details.txid=transactions.txid AND transaction_details.txid=:txid")?;
```
afilini: I would keep this query the same as it was before, because while it doesn't make too much difference for sqlite I think it would be a lot slower to make two queries over network rather than one. So if in the future we add support for postgres or mysql it will probably be based on the sqlite code, and for this reason I would try to be forward-thinking here.

danielabrozzoni: If it makes the code cleaner, we might decide to pass a `include_raw` parameter to `select_transaction_details_by_txid` and, even though we will still fetch the transaction from the db using this query, decide not to return it if it's not needed

vladimirfomene: I will go with just setting the transaction field of the return object to None.

src/database/mod.rs
```diff
@@ -619,13 +631,21 @@ pub mod test {
...
+        assert!(res.is_none());
     }
 
     pub fn test_check_descriptor_checksum<D: Database>(mut db: D) {
+        // insert checksum associated to keychain
         let checksum = "1cead456".as_bytes();
         let keychain = KeychainKind::External;
         let _res = db.check_descriptor_checksum(keychain, checksum);
 
+        // check if `check_descriptor_checksum` throws
```
danielabrozzoni, line 645: Those comments are great, but included in the wrong commit - should be in fd48469b617b2a0335d8a982d913434e8b6df77e, not 2ea1f574928b85ee00829efa9cfc4b23e8a712df

src/database/mod.rs
```diff
@@ -494,7 +491,10 @@ pub mod test {
         let (chain, child) = db.del_path_from_script_pubkey(&script).unwrap().unwrap();
 
         assert_eq!(chain, keychain);
-        assert_eq!(path, child);
+        assert_eq!(child, path);
+
+        let res = db.get_path_from_script_pubkey(&script).unwrap();
+        assert!(res.is_none());
```
danielabrozzoni, line 497: Same for this test additions (and more below): should be in https://github.com/bitcoindevkit/bdk/commit/fd48469b617b2a0335d8a982d913434e8b6df77e, not https://github.com/bitcoindevkit/bdk/commit/2ea1f574928b85ee00829efa9cfc4b23e8a712df

rajarshimaitra (approved):
tACK e65edbf53cb14e273ff99ed0576843dd3367385d

Thanks for the new tests..

danielabrozzoni (approved):
Code review ACK e65edbf53cb14e273ff99ed0576843dd3367385d

PR #746, opened 2022-09-08
Add mnemonic_to_descriptors example

vladimirfomene:
### Description

Using *bdk-cli* it is simple for a new user to generate a mnemonic phrase and descriptors. This might 
not be clear for new users when using *bdk* itself.

### Notes to the reviewers

This was initially requested by one user but might be relevant for other users as well.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/examples/mnemonic_to_descriptors.rs
+++ b/examples/mnemonic_to_descriptors.rs
@@ -1,0 +1,60 @@
+// Copyright (c) 2020-2021 Bitcoin Dev Kit Developers
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+use bdk::bitcoin::secp256k1::Secp256k1;
+use bdk::bitcoin::util::bip32::DerivationPath;
+use bdk::bitcoin::Network;
+use bdk::descriptor;
+use bdk::descriptor::IntoWalletDescriptor;
+use bdk::keys::bip39::{Language, Mnemonic, WordCount};
+use bdk::keys::{GeneratableKey, GeneratedKey};
+use bdk::miniscript::Tap;
+use bdk::Error as BDK_Error;
+use std::error::Error;
+use std::str::FromStr;
+
+/// This example demonstrates how to generate a mnemonic phrase
+/// using BDK and use that to generate a descriptor string.
+fn main() -> Result<(), Box<dyn Error>> {
+    let secp = Secp256k1::new();
+
+    // In this example we are generating a 12 words mnemonic phrase
+    // but it is also possible generate 15, 18, 21 and 24 words
+    // using their respective `WordCount` variant.
+    let mnemonic: GeneratedKey<_, Tap> =
+        Mnemonic::generate((WordCount::Words12, Language::English))
+            .map_err(|_| BDK_Error::Generic("Mnemonic generation error".to_string()))?;
+
+    println!("Mnemonic phrase: {}", *mnemonic);
+    let mnemonic_with_passphrase = (mnemonic, None);
+
+    // define external and internal derivation key path
+    let external_path = DerivationPath::from_str("m/86h/0h/0h/0").unwrap();
+    let internal_path = DerivationPath::from_str("m/86h/0h/0h/1").unwrap();
+
+    // generate external and internal descriptor from mnemonic
+    let (external_descriptor, ext_keymap) =
+        descriptor!(tr((mnemonic_with_passphrase.clone(), external_path)))?
+            .into_wallet_descriptor(&secp, Network::Testnet)?;
+    let (internal_descriptor, int_keymap) =
+        descriptor!(tr((mnemonic_with_passphrase, internal_path)))?
+            .into_wallet_descriptor(&secp, Network::Testnet)?;
+
+    println!("tpub external descriptor: {}", external_descriptor);
+    println!("tpub internal descriptor: {}", internal_descriptor);
+    println!(
+        "tprv external descriptor: {}",
+        external_descriptor.to_string_with_secret(&ext_keymap)
+    );
+    println!(
+        "tprv internal descriptor: {}",
+        internal_descriptor.to_string_with_secret(&int_keymap)
+    );
+
+    Ok(())
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -153,6 +153,11 @@ name = "esplora_backend_asynchronous"
 path = "examples/esplora_backend_asynchronous.rs"
 required-features = ["use-esplora-reqwest", "reqwest-default-tls", "async-interface"]
 
+[[example]]
+name = "mnemonic_to_descriptors"
+path = "examples/mnemonic_to_descriptors.rs"
+required-features = ["all-keys"]
+
 [workspace]
 members = ["macros"]
 [package.metadata.docs.rs]


examples/mnemonic_to_descriptors.rs
```diff
@@ -0,0 +1,75 @@
...
+    let master_xprv = xkey.into_xprv(Network::Testnet).ok_or_else(|| {
+        BDK_Error::Generic("Privatekey info not found (should not happen)".to_string())
+    })?;
+
+    let fingerprint = master_xprv.fingerprint(&secp);
+    let phrase = mnemonic
+        .word_iter()
+        .fold("".to_string(), |phrase, w| phrase + w + " ")
+        .trim()
+        .to_string();
```
afilini: Can't you just use `mnemonic.to_string()` here?

examples/mnemonic_to_descriptors.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    // In this example we are generating a 12 words mnemonic phrase
+    // but it is also possible generate 15, 18, 21 and 24 words
+    // using their respective `WordCount` variant.
+    let mnemonic: GeneratedKey<_, miniscript::BareCtx> =
+        Mnemonic::generate((WordCount::Words12, Language::English))
+            .map_err(|_| BDK_Error::Generic("Mnemonic generation error".to_string()))?;
+
+    let mnemonic = mnemonic.into_key();
+    let xkey: ExtendedKey = (mnemonic.clone(), None).into_extended_key()?;
```
afilini: I would add a comment here that `None` is the passphrase because it took me a second to figure it out

afilini: Alternatively, I think you can simplify this to just `mnemonic.clone().into_extended_key()?`

examples/mnemonic_to_descriptors.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    // You can replace this derivation path with one of your choosing
+    let path = DerivationPath::from_str("m/84h/0h/0h").unwrap();
+
+    let derived_xprv = &master_xprv.derive_priv(&secp, &path)?;
+
+    let origin: KeySource = (fingerprint, path);
+
+    let derived_xprv_desc_key: DescriptorKey<Segwitv0> =
+        derived_xprv.into_descriptor_key(Some(origin), DerivationPath::default())?;
```
afilini: Instead of deriving manually, I think you can let bdk do it:

```rust
let derived_xprv_desc_key: DescriptorKey<Segwitv0> = (mnemonic, path).into_descriptor_key()?;
```

examples/mnemonic_to_descriptors.rs
```diff
@@ -0,0 +1,75 @@
...
+/// It then uses the master extended private key with a derivation
+/// path to generate child extended private and public keys. Finally,
+/// it uses the derivation path and these keys to get the descriptors.
+fn main() -> Result<(), Box<dyn Error>> {
+    let secp = Secp256k1::new();
+
+    // In this example we are generating a 12 words mnemonic phrase
+    // but it is also possible generate 15, 18, 21 and 24 words
+    // using their respective `WordCount` variant.
+    let mnemonic: GeneratedKey<_, miniscript::BareCtx> =
```
afilini: You should use the same ctx here for consistency (you used `Segwitv0` down below)

examples/mnemonic_to_descriptors.rs
afilini: `mnemonic` can also be replaced with `(mnemonic, None)` if you want to make it clear that it supports passphrases as well

examples/mnemonic_to_descriptors.rs
vladimirfomene: Done

examples/mnemonic_to_descriptors.rs
vladimirfomene: Done

examples/mnemonic_to_descriptors.rs
vladimirfomene: added comment

examples/mnemonic_to_descriptors.rs
vladimirfomene: Done

examples/mnemonic_to_descriptors.rs
```diff
@@ -0,0 +1,72 @@
...
+/// it uses the derivation path and these keys to get the descriptors.
+fn main() -> Result<(), Box<dyn Error>> {
+    let secp = Secp256k1::new();
+
+    // In this example we are generating a 12 words mnemonic phrase
+    // but it is also possible generate 15, 18, 21 and 24 words
+    // using their respective `WordCount` variant.
+    let mnemonic: GeneratedKey<_, Segwitv0> =
+        Mnemonic::generate((WordCount::Words12, Language::English))
+            .map_err(|_| BDK_Error::Generic("Mnemonic generation error".to_string()))?;
```
rajarshimaitra, line 31: This got me thinking, why don't we have a feature gated error variant for `bip39::Error` for `bdk::Error`.  Doesn't have to be included here, but probably an idea for future PR..

examples/mnemonic_to_descriptors.rs
```diff
@@ -0,0 +1,72 @@
...
+        let desc_pubkey = desc_seckey
+            .as_public(&secp)
+            .map_err(|e| BDK_Error::Generic(e.to_string()))?;
+        println!("xpub {}", desc_pubkey);
+        println!("xprv {}", desc_seckey);
+    } else {
+        return Err(Box::new(BDK_Error::Key(Message(
+            "Invalid key variant".to_string(),
+        ))));
+    }
```
rajarshimaitra: As per the description of the example, it would be good to extend this a little further and create a full descriptor string with the `Xpriv`.. And then another step of showing how to create a watch only descriptor with the `Xpub` part alone..

rajarshimaitra (commented):
ACK.. Thanks for writing this.. Would be a really helpful example.. Few comments below..

examples/mnemonic_to_descriptors.rs
vladimirfomene: I will do that!

examples/mnemonic_to_descriptors.rs
vladimirfomene: I don't know if this will be necessary in the long term given that we are planning to implement our own version of BIP 39 and that will have an error enum.

danielabrozzoni:
I agree with Raj that this example should also cover creating the actual descriptor, instead of stopping at the xprv

However, you don't really need to manipulate the mnemonic that much to create a descriptor with bdk, as the `descriptor!` macro also accepts mnemonics (or pairs `(mnemonics, passphrase)`)

```
    let mnemonic: GeneratedKey<_, Tap> =
        Mnemonic::generate((WordCount::Words12, Language::English))
            .map_err(|_| BDK_Error::Generic("Mnemonic generation error".to_string()))?;
   
    println!("Mnemonic phrase: {}", mnemonic);
    let mnemonic_with_passphrase = (mnemonic, None);

    let external_path = DerivationPath::from_str("m/86h/0h/0h/0").unwrap();
    let internal_path = DerivationPath::from_str("m/86h/0h/0h/1").unwrap();

    let external_descriptor = descriptor!(tr((mnemonic_with_passphrase.clone(), external_path)))?;
    let internal_descriptor = descriptor!(tr((mnemonic_with_passphrase, internal_path)))?;
    println!("{}", external_descriptor);
    println!("{}", internal_descriptor);
```

This greatly simplifies your whole example, and I think it's the right way to go, as it shows how powerful the `descriptor!` macro really is :)

vladimirfomene:
Thanks @danielabrozzoni for helping me with the debugging! @notmandatory this was built on top of #800.

rajarshimaitra (approved):
tACK 247030559d65c1c7d962f972f60872042f6ecf9a

I will go through the network fix more closely later.. If this is scheduled for 0.25, I think this can go in now..

notmandatory:
I merged PR #800, go ahead and re-base this PR, just to make sure the history will look good.

notmandatory (approved):
ACK df905a8d5ef8e275188389486ac685b29f0d8c9f

PR #747, opened 2022-09-09
Run code coverage on every PR

danielabrozzoni:
#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #748, opened 2022-09-09
Run the code coverage with different features

danielabrozzoni:
### Description

Based on #747

This runs multiple codecov jobs and then merges the results all together. This increases the report precision, as we make sure to run tests with all the features.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -1,4 +1,4 @@
-on: [push]
+on: [push, pull_request]
 
 name: Code Coverage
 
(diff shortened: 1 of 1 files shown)


danielabrozzoni:
This doesn't really seem to work...
What I'm trying to do is: run multiple tests with different features and then merge the coveralls reports.
This gives some weird results with esplora and electrum - for example, the coverage of the electrum module actually decreased, even tho we're still testing with all the electrum features!
The problem is probably that coveralls has to merge reports with different source codes (different features -> different source code), which leads to a wrong result.

PR #751, opened 2022-09-13
nakamoto as Compact Block Filter Blockchain

rajarshimaitra:
### Description

An implementation of [nakamoto](https://github.com/cloudhead/nakamoto) as a BDK Blockchain to serve as an alter BIP157 syncing mechanism.  

Commit 1 :  adds a new module `src/blockchain/compact_filters/nakamoto.rs` which defines a `CbfNode` struct which holds the `namakoto-client` handlers. Which the implements `Blockchain` traits.  222a2b466626c07cae4db048c10014ca44514c3e

Commit 2 : Integrates the `CbfNode` backend with the bdk's blockchain test framework. 

### Notes to Reviewers
The current CI tests won't work because nakamoto doesn't support our MSRV. Though you can run the modified tests in your local with toolchain 1.63 or higher. 

Few of the tests are intermittently failing and needs further investigation. But overall this PR is ready for review of to be used for testing..  

Thanks for all the help and support from @cloudhead and others who pitched in for various fixes in upstream.  

#### All Submissions:


Cargo.toml
```diff
@@ -39,6 +40,13 @@ bitcoinconsensus = { version = "0.19.0-3", optional = true }
 # Needed by bdk_blockchain_tests macro and the `rpc` feature
 bitcoincore-rpc = { version = "0.15", optional = true }
 
+# Nakamoto crates for the CBF blockchain
+nakamoto-client = { git = "https://github.com/cloudhead/nakamoto.git", branch = "master", optional = true }
+nakamoto-net =  { git = "https://github.com/cloudhead/nakamoto.git", branch = "master", optional = true }
+nakamoto-net-poll =  { git = "https://github.com/cloudhead/nakamoto.git", branch = "master", optional = true }
+nakamoto-p2p = { git = "https://github.com/cloudhead/nakamoto.git", branch = "master", optional = true }
+nakamoto-common = { git = "https://github.com/cloudhead/nakamoto.git", branch = "master", optional = true }
```
cloudhead: You should be able to just import the `nakamoto` crate. It exposes these externally.

src/blockchain/compact_filters/nakamoto.rs
```diff
@@ -0,0 +1,677 @@
...
+                    }
+                }
+                Event::BlockMatched {
+                    hash: _,
+                    header,
+                    height,
+                    transactions,
+                } => {
+                    let timestamp = header.time;
+                    for tx in transactions {
```
cloudhead, line 559: Note that *all* transactions of the matching block are included here, so you have to make sure to discard the ones you aren't interested in.

src/blockchain/compact_filters/nakamoto.rs
```diff
@@ -0,0 +1,677 @@
...
+                                    None
+                                }
+                            } else {
+                                None
+                            }
+                        })
+                        .collect::<HashSet<_>>();
+
+                    for txid in db_txs {
+                        unconfirm_tx(database, &txid)?;
```
cloudhead, line 549: Note that nakamoto should already be emitting a `TxStatus::Reverted` for each tx in a reverted block, so you shouldn't have to handle it here.

src/blockchain/compact_filters/nakamoto.rs
```diff
@@ -0,0 +1,677 @@
...
+                    progress_update.update(
+                        (height as f32 / tip as f32) * 100.0,
+                        Some("Sync Progress".into()),
+                    )?;
+                    // We break the sync loop once we reach current height.
+                    if height == tip {
+                        debug!("Sync complete at : {}", height);
+                        self.last_sync_height.set(height as u32);
+                        break;
+                    }
```
cloudhead, line 628: I assume you have another mechanism to restart the event loop when a new block header is received.

src/blockchain/compact_filters/nakamoto.rs
```diff
@@ -0,0 +1,677 @@
...
+                        }
+                        TxStatus::Acknowledged { peer } => {
+                            debug!("txid: {}, status : ACK by {}", txid, peer);
+                        }
+                        TxStatus::Confirmed { height, .. } => {
+                            debug!("txid: {}, status : Confirmed at {}", txid, height);
+                        }
+                        TxStatus::Reverted => {
+                            debug!("Transaction reverted due to reorg: txid : {}", txid);
+                            let _ = unconfirm_tx(database, &txid)?;
```
cloudhead: FYI, nakamoto will re-announce reverted transactions. I don't think we fire another event for them other than this one (but we could).

cloudhead (commented):
Looks good to me.

Cargo.toml
rajarshimaitra: Thanks for the tip.. I missed it.. Yes that's much more cleaner..

src/blockchain/compact_filters/nakamoto.rs
rajarshimaitra: We are actually handling that in the `add_tx` itself.. So any unrelated transaction will not get added in the db.. But ya the will be iterated over here but with not much extra over head as some basic `if let` statements in `add_tx` won't trigger for them. Maybe there's a better way to organize this, but for start it made sense to consolidate all transaction conflicts and addition logic in single place in `add_tx`..

Here we are just checking if the transaction hasn't been already processed by us..

src/blockchain/compact_filters/nakamoto.rs
rajarshimaitra: Hmm okay makes sense.. We still handle reverted cases in `TxStatus::Reverted`.. I am commenting this section out for now..

src/blockchain/compact_filters/nakamoto.rs
rajarshimaitra: No, this is basically a hack to stop the sync loop for the reorg tests in regetst.. Ideally an `invalidateblock` call on the regtest should also update the tip height, and the sync should automatically stop at the new tip height, but that wasn't happening.. So instead this hack is to force stop the sync..  This is feature gated as `test`, and cannot be used in production..

src/blockchain/compact_filters/nakamoto.rs
rajarshimaitra: I think this transaction reversion event should be sufficient to catch them..

rajarshimaitra:
Thanks @cloudhead for the review and suggestion.. It seems the structure of the crate has changed in recent master.. So I pinned the rev to `14902d3d5cc9442cdda82419f0f24277d86b243b`. I couldn't find any release tag in github, and the published `v0.3.0` in crates.io doesn't seem to include some latest fixes.. Anyway, this works for now, until we get a more stable release of nakamoto out in crates.io..

rajarshimaitra:
@thunderbiscuit this now includes `AnyBlockhain` and `ConfigurableBlockchain` updates..

src/blockchain/compact_filters/nakamoto.rs
```diff
@@ -0,0 +1,687 @@
...
+    /// Last Sync height
+    last_sync_height: Cell<u32>,
+    // A Hack to handle reorg sync cases.
+    #[cfg(test)]
+    break_sync_at: Cell<Option<u32>>,
+}
+
+/// Nakamoto CBF Client Configuration
+#[derive(Debug, Clone, PartialEq, Eq, serde::Deserialize, serde::Serialize)]
+pub struct CBFConfig {
```
thunderbiscuit: small nit: I think to keep in line with the other two `ElectrumBlockchainConfig` and `EsploraBlockchainConfig` you'd want to name this one `CbfBlockchainConfig`, or maybe `NakamotoBlockchainConfig` (just a suggestion).

src/blockchain/compact_filters/nakamoto.rs
```diff
@@ -0,0 +1,687 @@
...
+    network: bitcoin::Network,
+    datadir: Option<PathBuf>,
```
thunderbiscuit: We'd need these two fields to be public if possible (which is in line with the `EsploraBlockchainConfig` and `ElectrumBlockchainConfig` fields).

thunderbiscuit (commented):
Hey Raj! So I'm trying to expose this in the bindings and it surfaces two small things: (1) I think the names for the other blockchain config structs are slightly different, (2) the fields in the config struct are currently private but I'd need access to them.

To get an idea for how we use the config structs (and how we wrap them up slightly, note the `ElectrumBlockchainConfig` to `ElectrumConfig` same for the Esplora) you can take a look at lines 98ish to 225ish of [this source file](https://github.com/bitcoindevkit/bdk-ffi/blob/master/src/lib.rs).

src/blockchain/compact_filters/nakamoto.rs
cloudhead: Ah gotcha, ok.

src/blockchain/compact_filters/nakamoto.rs
rajarshimaitra: Done..

src/blockchain/compact_filters/nakamoto.rs
rajarshimaitra: Done..

rajarshimaitra:
@thunderbiscuit updated as suggested..

rajarshimaitra:
One more extra commit is added to disable to existing compact_filters logic, so we don't need to unnecessarily build rocksdb and other stuffs not used for nakamoto client..

thunderbiscuit:
Working on the FFI integration here: bitcoindevkit/bdk-ffi#207
With this wallet: [Devkit Wallet Nakamoto branch](https://github.com/thunderbiscuit/devkit-wallet/tree/nakamoto)

danielabrozzoni:

PR #754, opened 2022-09-16
Fix the release process

afilini:
### Description

A few things I noticed while doing the 0.22 release.

Follow-up of #544 

#### All Submissions:

### DIFF
--- a/.github/ISSUE_TEMPLATE/minor_release.md
+++ b/.github/ISSUE_TEMPLATE/minor_release.md
@@ -21,9 +21,9 @@ assignees: ''
 
 <--add notices from PRs merged since the prior release, see ["keep a changelog"]-->
 
-### Checklist 
+### Checklist
 
-Release numbering must follow [Semantic Versioning]. These steps assume the current `master` 
+Release numbering must follow [Semantic Versioning]. These steps assume the current `master`
 branch **development** version is *MAJOR.MINOR.0*.
 
 #### On the day of the feature freeze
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/minor_release.md
```diff
@@ -37,34 +37,43 @@ Change the `master` branch to the next MINOR+1 version:
   - The commit message should be "Bump version to MAJOR.MINOR+1.0".
 - [ ] Create PR and merge the `bump_dev_MAJOR_MINOR+1` branch to `master`.
   - Title PR "Bump version to MAJOR.MINOR+1.0".
-  
+
 Create a new release branch and release candidate tag:
 
 - [ ] Double check that your local `master` is up-to-date with the upstream repo.
 - [ ] Create a new branch called `release/MAJOR.MINOR+1` from `master`.
+- [ ] Bump the `release/MAJOR.MINOR+1` branch to `MAJOR.MINOR+1.0-rc.1` version.
```
notmandatory, line 45: Was wondering if there's a particular reason to also update the cargo version to include the `-rc.x` part of the version. Is it just to be more explicit so pre-release users know it's not official yet? 

It's not a big deal for me either way, but the reason I left it off was that I figured anyone doing rc testing would be pointing to the release branch or rc tag  in a [patch.crates-io] statement. And then could just remove that patch line once the release was out, also was trying to avoid extra Cargo.toml file editing steps.

afilini: It's just to make sure the version is reflected into cargo, otherwise from cargo's perspective all of those are the same version, it's just different tags pointing to them.

So for example when I run a cargo build I see "bdk 0.22-rc.1" instead of just "bdk 0.22" which may be confusing.

notmandatory (approved):
ACK e1fa0b66954cc56dae055464ab5bb40b7a5043be

PR #756, opened 2022-09-18
Remove genesis_block lazy initialization

### LINKED ISSUE #752: Remove genesis_block lazy initialization
Remove this genesis block initialization:

https://github.com/bitcoindevkit/bdk/blob/7b12f35698ff468f676665704fb274c36cb305df/src/blockchain/compact_filters/store.rs#L40-L45

by using https://docs.rs/bitcoin/0.29.1/bitcoin/blockdata/constants/fn.genesis_block.html#

After that lazy_static could probably be removed or set as dev-dep only

shobitb:
### Description

This commit fixes #752

cargo test runs successfully.

### Notes to the reviewers

Hi, newbie here learning Rust and BDK!  I've removed the lazy_static block in this commit, and when learning about lazy_static also came across something called [once_cell](https://doc.rust-lang.org/std/cell/struct.OnceCell.html), [soon to be available in stdlib](https://github.com/rust-lang/rust/issues/74465).  It's like lazy_static but faster and is not a macro.  Shall I keep the lazy_static and create an issue to switch to once_cell in the future, or remove the lazy_static (as I've done) and create an issue anyway? 

#### All Submissions:


danielabrozzoni:
Hey, thanks for your contribution :)
There are some compilation errors when compiling with the `compact_filter` feature, you can see them if you compile like this:
```
cargo build --features compact_filters
```

Regarding `lazy_static` vs `once_cell`: in this specific case I'd say it's better to use `genesis_block` (as you're doing), but in the other cases, sure, once_cell would be better! Feel free to open an issue :)

shobitb:
Thanks @danielabrozzoni! I've updated my commit and ensured compact_filters compiles successfully.

```
bdk % cargo build --features compact_filters
   Compiling bdk v0.22.0 (/Users/shobit/bdk)
    Finished dev [unoptimized + debuginfo] target(s) in 27.82s
```

afilini (commented):
You should also update the `Cargo.toml` to remove lazy_static from the compact_filters feature since it's not needed anymore.

Actually I think you may even be able to remove the lazy_static dependency (not the dev-dependency which is used in the tests) entirely, I don't think it's used anywhere else

shobitb:
> You should also update the `Cargo.toml` to remove lazy_static from the compact_filters feature since it's not needed anymore.
> 
> Actually I think you may even be able to remove the lazy_static dependency (not the dev-dependency which is used in the tests) entirely, I don't think it's used anywhere else

Got it, and thank you for taking a look!  I've updated the commit after removing the dependency (but not dev-dependency).

vladimirfomene (commented):
Tested ACK. Thanks for the contribution!

danielabrozzoni:
Code looks good to me, can you please squash all commits? You have three separate commits with the same title/message, it should be just one.

https://stackoverflow.com/a/5189600

shobitb:
> Code looks good to me, can you please squash all commits? You have three separate commits with the same title/message, it should be just one.
> 
> https://stackoverflow.com/a/5189600

Done!  Thanks for the link.

notmandatory (approved):
ACK e6f2d029fa9708f98599c1bd4ef74d232b111c5a

PR #763, opened 2022-09-27
Fix `Wallet::descriptor_checksum` to actually return the checksum

evanlinjin:
### Description

`Wallet::descriptor_checksum` should return the checksum, not the descriptor without the checksum.

### Notes to the reviewers

Please merge.

### Changelog notice

Fix `Wallet::descriptor_checksum` to actually return the descriptor checksum.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1842,7 +1842,7 @@ where
             .to_string()
             .split_once('#')
             .unwrap()
-            .0
+            .1
             .to_string()
     }
 }
@@ -1937,6 +1937,22 @@ pub(crate) mod test {
     // Here, we push just once for simplicity, so we have to add an extra byte for the missing
     // OP_PUSH.
     const P2WPKH_FAKE_WITNESS_SIZE: usize = 106;
+
+    #[test]
+    fn test_descriptor_checksum() {
+        let (wallet, _, _) = get_funded_wallet(get_test_wpkh());
+        let checksum = wallet.descriptor_checksum(KeychainKind::External);
+        assert_eq!(checksum.len(), 8);
+
+        let raw_descriptor = wallet
+            .descriptor
+            .to_string()
+            .split_once('#')
+            .unwrap()
+            .0
+            .to_string();
+        assert_eq!(get_checksum(&raw_descriptor).unwrap(), checksum);
+    }
 
     #[test]
     fn test_get_funded_wallet_balance() {


evanlinjin:
So these lines here...

https://github.com/bitcoindevkit/bdk/blob/b14e4ee3a0d7fb722a8d2c08705ad913ff04a366/src/wallet/mod.rs#L208
https://github.com/bitcoindevkit/bdk/blob/b14e4ee3a0d7fb722a8d2c08705ad913ff04a366/src/wallet/mod.rs#L217

Essentially, we are getting the checksum of (descriptors+checksum)... This does not seem like wanted behavior?

This needs to be fixed, but in a way that it doesn't result in incompatible databases. I think I have a way of fixing this while being backwards compatible.

For adding new db checksums: Use checksum of (descriptor).
For verifying db checksums: Either checksum(descriptor) or checksum(descriptor+checksum) is okay.

@afilini @notmandatory Is this a good proposal?

notmandatory:
To add another wrinkle to this discussion, there's also a non-pub function in `miniscript::descriptor::checksum::desc_checksum()` which is what's used in the `Display` impl for all the script types in the `miniscript` crate. I think the `bdk` version of `get_checksum()` was probably added because we couldn't use the `miniscript` one. 

I think this is relevant in that we probably shouldn't implement our own `get_checksum()` and certainly  not have two ways in our project to get it. 

EDIT: see ACK below. 
~~@evanlinjin I'd like to get this cleaned up, but it's not urgent right? can we move it to the 0.24 milestone so we have more time to discuss?~~

notmandatory (approved):
ACK af0b3698c691cd08ece9ae201cc9699f0d57acc8

Even though there may be more we can do to cleanup how checksums are done, this certainly fixes a clear bug and would be nice to get into the 0.23 release.

evanlinjin:
@notmandatory yes I agree with you. I think this PR fixes an API issue, everything else I've mentioned is really more internal.

danielabrozzoni:
utACK af0b3698c691cd08ece9ae201cc9699f0d57acc8

danielabrozzoni:
Oh nevermind, it looked too fun, I had to test it! 😄 

ACK af0b3698c691cd08ece9ae201cc9699f0d57acc8 - I run the test you added with the old code, and verified that the bug was there. I then run the test (with a few more dbg!() expressions) and manually verified that the problem is fixed.

PR #764, opened 2022-09-28
Use the esplora client crate

afilini:
### Description

Use the external esplora client crate now that it's published

### Changelog notice

- Start using the external esplora client crate
- Deprecate the `use-esplora-reqwest` and `use-esplora-ureq` features in favor of `use-esplora-async` and `use-esplora-blocking`

#### All Submissions:

#### New Features:

#### Bugfixes:


Cargo.toml
```diff
@@ -79,13 +78,13 @@ hardware-signer = ["hwi"]
 async-interface = ["async-trait"]
 electrum = ["electrum-client"]
 # MUST ALSO USE `--no-default-features`.
-use-esplora-reqwest = ["esplora", "reqwest", "reqwest/socks", "futures"]
-use-esplora-ureq = ["esplora", "ureq", "ureq/socks"]
+use-esplora-reqwest = ["esplora", "esplora-client/async", "futures"]
+use-esplora-ureq = ["esplora", "esplora-client/blocking"]
```
tnull: I think keeping these names is a bit strange given the `ureq`/`reqwest` dependencies have been removed. Maybe these features could also be renamed `use-esplora-async`/`use-esplora-blocking`, while initially keeping the deprecated features around as aliases to not break compatibility?

notmandatory: I also like this idea. Add the new feature names, but keep the old ones around for a release with comment in `Cargo.toml` and we'll put in tag changelog that old feature names are deprecated.

afilini: Apparently there's no way to deprecate a feature, should we just add a note to the changelog?

notmandatory: What you did with making the old ureq/reqwest features aliases for the new names is all I was expecting, plus adding your change log note.

notmandatory (approved):
ACK 5f50be91a2186fd7ddabcbe03e9774d9f1b1d180

Cargo.toml
vladimirfomene: In line with @tnull point, I was wondering if it doesn't also make sense to rename the `ureq.rs` and `reqwest.rs` file under the `esplora` module to `blocking.rs` and `async.rs` respectively.

tnull (commented):
LGTM, mod https://github.com/bitcoindevkit/bdk/pull/764#discussion_r983096231

Cargo.toml
afilini: I did consider it but since the modules are not public it doesn't make any difference for the public API, so it's just an internal thing

tnull: Hm, but doesn't that mean that there is even less of a reason not to clean it up?

afilini: Yeah, maybe, I personally don't mind but I'm not against cleaning it up either.

I replied like that because for some reason I though this PR had already been merged by Steve and so I thought it was too late to fix it and I didn't want to open another PR just for that. But since this is still open yeah I can do it!

notmandatory: I'll do a quick re-review and merge it now. :-)

notmandatory (approved):
ACK d7bfe68e2df270ab799d36ebf3563e178ae50c6e

PR #765, opened 2022-09-29
Fix how descriptor checksums are calculated

evanlinjin:
### Description

Previously, the methods `get_checksum_bytes` and `get_checksum` do not check input data to see whether the input data already has a checksum.

This PR does the following:

* Introduce a `exclude_hash: bool` flag for `get_checksum_bytes`, that excludes the checksum portion of the original data when calculating the checksum. In addition to this, if the calculated checksum does not match the original checksum, an error is returned for extra safety.
* Ensure `Wallet` is still backwards compatible with databases created with the "checksum inception" bug.

### Notes to the reviewers

Thank you.

### Changelog notice

Fix the "checksum inception" bug, where we may accidentally calculate the checksum of a descriptor that already has a checksum.

#### All Submissions:

#### Bugfixes:


src/descriptor/checksum.rs
```diff
@@ -41,12 +41,21 @@ fn poly_mod(mut c: u64, val: u64) -> u64 {
     c
 }
 
-/// Computes the checksum bytes of a descriptor
-pub fn get_checksum_bytes(desc: &str) -> Result<[u8; 8], DescriptorError> {
+/// Computes the checksum bytes of a descriptor.
+/// `exclude_hash = true` ignores all data after the first '#' (inclusive).
+pub fn get_checksum_bytes(mut desc: &str, exclude_hash: bool) -> Result<[u8; 8], DescriptorError> {
```
rajarshimaitra: Instead of innocently naming it `exlcude_hash` can we name it something dangerous like `exclude_inception_bug`? That will clear the intent that this value should never be set to `false` unless wrong checksum calculation is intended.

In the current state always using this value as `true` works for both the cases of w/wo checksum in descriptor. So Ideally everywhere else we should just use `get_checksum` and avoid using this function. Better to remove from public interface also?

src/descriptor/mod.rs
```diff
@@ -84,19 +85,15 @@ impl IntoWalletDescriptor for &str {
...
-            let parts: Vec<&str> = self.splitn(2, '#').collect();
-            if !get_checksum(parts[0])
-                .ok()
-                .map(|computed| computed == parts[1])
-                .unwrap_or(false)
-            {
-                return Err(DescriptorError::InvalidDescriptorChecksum);
+        let descriptor = match self.split_once('#') {
+            Some((desc, original_checksum)) => {
+                let checksum = get_checksum_bytes(desc, false)?;
```
rajarshimaitra: Why not just use `get_checksum()` here? Its guaranteed that `desc` won't have checksum at this stage.

src/wallet/mod.rs
```diff
@@ -243,6 +244,19 @@ where
...
+    /// check is necessary for backwards compatibility of a checksum-inception bug.
+    fn db_checksum(db: &mut D, desc: &str, kind: KeychainKind) -> Result<(), Error> {
+        let checksum = get_checksum_bytes(desc, true)?;
+        if db.check_descriptor_checksum(kind, checksum).is_ok() {
+            return Ok(());
+        }
+
+        let checksum_inception = get_checksum_bytes(desc, false)?;
+        db.check_descriptor_checksum(kind, checksum_inception)
+    }
```
rajarshimaitra: Instead of creating another function here, is it possible to move this logic inside `check_descriptor_checksum` itself? This hacky bug fix doesn't need to be so up high in the library. And can be confusing later on.

src/wallet/mod.rs
```diff
@@ -1943,15 +1957,38 @@ pub(crate) mod test {
...
-        let raw_descriptor = wallet
-            .descriptor
-            .to_string()
-            .split_once('#')
-            .unwrap()
-            .0
-            .to_string();
-        assert_eq!(get_checksum(&raw_descriptor).unwrap(), checksum);
+    #[test]
+    fn test_db_checksum() {
```
rajarshimaitra, line 1883: Also wondering if its possible to reproduce the bug in test itself.. Master should fail in the test while this PR should pass. This backward compatibility test can stay too..

rajarshimaitra (commented):
Concept ACK.. Great job finding this out.. I reproduced the bug by the following method

 - Created a descriptor, once with and then without checksum. Calculate checksum manually using `get_checksum()`. 
 -  Calculate checksum using `wallet_name_from_descriptor` which calls the same `get_checksum()` function but ensures to exclude the checksum if present in descriptor. 
 - Read the checksum value stored in db after running bdk-cli using w/wo checksum descriptor.

Here's the trial outputs for reference
```
Master:
desc w checksum: wpkh(032b0558078bec38694a84933d659303e2575dae7e91685911454115bfd64487e3)#wc9dk083

get_checksum() : arwnyfm2
wallet name : wc9dk083
get_checksum  bytes  [97, 114, 119, 110, 121, 102, 109, 50]
wallet name bytes : [119, 99, 57, 100, 107, 48, 56, 51]
checksum in db : [97, 114, 119, 110, 121, 102, 109, 50]

desc w/o checksum: wpkh(032b0558078bec38694a84933d659303e2575dae7e91685911454115bfd64487e3)

get_checksum : wc9dk083
wallet name : wc9dk083
get_checksum  bytes  [119, 99, 57, 100, 107, 48, 56, 51]
wallet name bytes : [119, 99, 57, 100, 107, 48, 56, 51]
checksum in db: [97, 114, 119, 110, 121, 102, 109, 50]

PR 765:
desc w checksum: wpkh(032b0558078bec38694a84933d659303e2575dae7e91685911454115bfd64487e3)#wc9dk083
get_checksum : wc9dk083
wallet name : wc9dk083
get_checksum  bytes  [119, 99, 57, 100, 107, 48, 56, 51]
wallet name bytes : [119, 99, 57, 100, 107, 48, 56, 51]
checksum in db : [119, 99, 57, 100, 107, 48, 56, 51]

desc w/o checksum: wpkh(032b0558078bec38694a84933d659303e2575dae7e91685911454115bfd64487e3)

get_checksum : wc9dk083
wallet name : wc9dk083
get_checksum  bytes  [119, 99, 57, 100, 107, 48, 56, 51]
wallet name bytes : [119, 99, 57, 100, 107, 48, 56, 51]
checksum in db : [119, 99, 57, 100, 107, 48, 56, 51]
```
Observation: 
 - Master is clearly storing wrong checksum in db for desc with checksum data. But it seems master is also storing the wrong checksum in case of descriptor w/o checksum. Check the Master Case 2, `checksum in db` value. I am not sure why this is happening. This means master will forcefully include the checksum in checksum calculation, even if its not there?
 
  - In any case this PR seems to be working as expected, and both `get_checksum` and checksum stored in db is matching with `wallet_name_from_descriptor`, so everything seems consistent.    

Few comments below on code.

src/descriptor/checksum.rs
danielabrozzoni: This is tough. I still think we need to keep `get_checksum_bytes` in the public interface, but I feel like the `exclude_hash` logic should be hidden (maybe in an internal `get_checksum_internal`?).

Also, I want everyone that updates to bdk 0.23 and used these methods to know that something has changed (optimally by displaying a warning or something when compiling), or otherwise I suppose some nasty bugs will arise. Same for `get_checksum`.

I honestly don't know how to handle this...

danielabrozzoni (commented):
I reviewed only the first commit for now, but I feel like we need to tell our users about the bug in some way:

- I don't want to silently change the behavior of `get_checksum` - if some users used it the same way we do, they need to implement some way of switching to the new checksums - or just keep using the old ones
- On the other hand... exposing a boolean `exclude_hash` is a bit confusing. All the new users, and some of the old ones, don't need to know what's `exclude_hash` is about.

I'm thinking that we might achieve something useful with deprecating old methods, but I'm really not sure... Any ideas?

src/descriptor/checksum.rs
evanlinjin: @danielabrozzoni @rajarshimaitra Maybe we can just rename the functions all together, this way no one gets confused?

```rust
// we use this internally
fn calc_checksum_bytes_internal(desc: &str, exclude_hash: bool) {}

// avoids checksum like a boss
pub fn calc_checksum_bytes(desc: &str) {} 

// also avoids checksum like a boss
pub fn calc_checksum(desc: &str) {}
```

src/descriptor/mod.rs
evanlinjin: It's to avoid extra calculations.

src/wallet/mod.rs
evanlinjin: We don't have the same logic for `get_checksum_bytes` in master though?

src/wallet/mod.rs
evanlinjin: `check_descriptor_checksum` is part of a trait.

src/descriptor/checksum.rs
danielabrozzoni: This is cool. Also, I'd deprecate the old ones with a comprehensive message, explaining that the new ones might return a different result

rajarshimaitra: ACK on the above.. Keeping the API same would be helpful.. The name changes looks good to me.. That with a good explainer would do..

src/descriptor/mod.rs
rajarshimaitra: Okay agreed..

src/wallet/mod.rs
rajarshimaitra: Ah right.. Trait.. 😄

src/wallet/mod.rs
rajarshimaitra: Hmm.. On 2nd look it seems to me we are covering all the behaviors necessary.. I will think more if there's more needed to be covered..

src/descriptor/checksum.rs
notmandatory: @evanlinjin looks like everyone is onboard with the above proposed naming, but if I understand correctly these changes are needed:
 - [ ] rename `get_checksum_bytes` to `calc_checksum_bytes_internal` and make not public
 - [ ] add deprecated public `calc_checksum_bytes` that calls `calc_checksum_bytes_internal` with exclude_hash == false, this replicates current behavior but warns current users they need to change to `calc_checksum`.
 - [ ] add `calc_checksum` that calls `calc_checksum_bytes_internal` with exclude_hash == true, this is new function with new name that users should use instead of depricated `calc_checksum_bytes`
 - [ ] in new PR for 0.25 release (asuming this PR goes into 0.24) consolidate correct behavior into `calc_checksum` and remove deprecated and internal functions.

We can discuss on the call Tues too.

evanlinjin: I'm assuming if we implement `calc_checksum_bytes_internal`, we wouldn't need `calc_checksum_internal`, but I might be wrong.

notmandatory: I added a commit to:
1. make  `calc_checksum_bytes_internal` pub(crate)
2. add backward compatible but deprecated `get_checksum_bytes` and `get_checksum` functions
3. rename replacement functions `calc_checksum_bytes` and `calc_checksum`
4. added a simple test to prove the new functions work as expected with a descriptor that already includes a checksum

I think this covers all the bases. Let me know if it looks OK or if you see anything else I should do.

src/descriptor/checksum.rs
```diff
@@ -72,37 +84,84 @@ pub fn get_checksum_bytes(desc: &str) -> Result<[u8; 8], DescriptorError> {
...
+        .map(|b| unsafe { String::from_utf8_unchecked(b.to_vec()) })
+}
+
+// TODO in release 0.25.0, remove get_checksum_bytes and get_checksum
+// TODO in release 0.25.0, consolidate calc_checksum_bytes_internal into calc_checksum_bytes
+
+/// Compute the checksum bytes of a descriptor
+#[deprecated(
+    since = "0.24.0",
+    note = "Use new `calc_checksum_bytes` function which excludes any existing hash before calculating the checksum hash bytes. See https://github.com/bitcoindevkit/bdk/pull/765."
```
rajarshimaitra: Nit: Maybe a bit more clear like this?

```suggestion
    note = "Use new `calc_checksum_bytes` function which excludes any existing checksum in the  descriptor string before calculating the checksum hash bytes. See https://github.com/bitcoindevkit/bdk/pull/765."
```

src/descriptor/checksum.rs
```diff
@@ -72,37 +84,84 @@ pub fn get_checksum_bytes(desc: &str) -> Result<[u8; 8], DescriptorError> {
...
+    // test calc_checksum() function; it should return the same value as Bitcoin Core even if the
+    // descriptor string includes a checksum hash
+    #[test]
+    fn test_calc_checksum_with_checksum_hash() {
+        let desc = "wpkh(tprv8ZgxMBicQKsPdpkqS7Eair4YxjcuuvDPNYmKX3sCniCf16tHEVrjjiSXEkFRnUH77yXc6ZcwHHcLNfjdi5qUvw3VDfgYiH5mNsj5izuiu2N/1/2/*)#tqz0nc62";
+        assert_eq!(calc_checksum(desc).unwrap(), "tqz0nc62");
+
+        let desc = "pkh(tpubD6NzVbkrYhZ4XHndKkuB8FifXm8r5FQHwrN6oZuWCz13qb93rtgKvD4PQsqC4HP4yhV3tA2fqr2RbY5mNXfM7RxXUoeABoDtsFUq2zJq6YK/44'/1'/0'/0/*)#lasegmfs";
+        assert_eq!(calc_checksum(desc).unwrap(), "lasegmfs");
     }
```
rajarshimaitra: How about also asserting here that with wrong checksum string at the end it should fail.. Can be included in the test case here itself.. This will allow us to hit the last `if let` clause of `calc_checksum_bytes_internal` that returns `DescriptorError::InvalidDescriptorChecksum`..

The failure case we are testing in the test next to this one is covering a `InvalidDescriptorCharacter` error, which happens in the descriptor parsing logic of `calc_checksum_bytes_internal`..

src/wallet/mod.rs
```diff
@@ -203,28 +204,28 @@ where
...
         let (descriptor, keymap) = into_wallet_descriptor_checked(descriptor, &secp, network)?;
-        database.check_descriptor_checksum(
+        Self::db_checksum(
+            &mut database,
+            &descriptor.to_string(),
             KeychainKind::External,
-            get_checksum(&descriptor.to_string())?.as_bytes(),
         )?;
         let signers = Arc::new(SignersContainer::build(keymap, &descriptor, &secp));
+
```
rajarshimaitra: Is this newline meant to be here?

rajarshimaitra (commented):
Thanks @notmandatory for carrying the work.. This looks almost complete and good to go..

tACK  60057a7bf78f01bfb1149251f384de6432f94d59

Below are few nits and testing suggestions..

src/wallet/mod.rs
```diff
@@ -1943,15 +1957,38 @@ pub(crate) mod test {
...
+        db.check_descriptor_checksum(KeychainKind::External, checksum_inception)
+            .expect("failed to save checksum inception");
+        Wallet::db_checksum(&mut db, &desc, KeychainKind::External)
+            .expect("db that uses checksum inception should be supported");
+
+        let mut db = MemoryDatabase::new();
+        db.check_descriptor_checksum(KeychainKind::External, checksum_invalid)
+            .expect("failed to save invalid checksum");
+        Wallet::db_checksum(&mut db, &desc, KeychainKind::External)
+            .expect_err("db that uses invalid checksum should fail");
```
rajarshimaitra, line 1907: Question, should we also check for invalid checksum error here too??

src/descriptor/checksum.rs
rajarshimaitra: Resolving this one as fixed in latest push..

evanlinjin (commented):
ACK 60057a7bf78f01bfb1149251f384de6432f94d59

@notmandatory Thank you for moving the PR forwards. Looks good to me.

src/wallet/mod.rs
notmandatory: I think it's OK as it is, especially keeping in mind that with `bdk_core` this stuff probably won't be needed.

src/descriptor/checksum.rs
notmandatory: good idea, fixed

src/descriptor/checksum.rs
notmandatory: Agreed, I took this wording and also used it in the deprecation node and the docs.

notmandatory (approved):
ACK 648282e60283f579ba1644fe506ae1370786da96

src/wallet/mod.rs
notmandatory: oops, not sure how it got there, I removed it.

notmandatory:
I think this one is ready to merge, but I'm going to hold off until #770 is merged since I think it will be easier to rebase this PR on that one.

PR #766, opened 2022-09-29
Deprecate `list_unspent` which has an unnecessary `.collect()`

evanlinjin:
### Description

Introduce `Wallet::iter_unspent`, which replaces `Wallet::list_unspent`. Thus avoiding an unnecessary `.collect()` for most use cases.

### Changelog notice

Introduce `Wallet::iter_unspent`, which replaces `Wallet::list_unspent`. Thus avoiding an unnecessary `.collect()` for most use cases.

#### All Submissions:

#### New Features:

### DIFF
--- a/src/testutils/blockchain_tests.rs
+++ b/src/testutils/blockchain_tests.rs
@@ -455,7 +455,7 @@ macro_rules! bdk_blockchain_tests {
                 assert!(wallet.database().deref().get_sync_time().unwrap().is_some(), "sync_time hasn't been updated");
 
                 assert_eq!(wallet.get_balance().unwrap().untrusted_pending, 50_000, "incorrect balance");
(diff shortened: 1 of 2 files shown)


danielabrozzoni:
Concept ACK! I just wonder, will this method change again after the bdk-core integration, or do you think it can remain the same? I'd avoid breaking this API in 0.24, and then immediately deprecate `iter_unspent` in 1.0
From my understanding of bdk_core, I think it can remain the same, but I might be wrong

evanlinjin:
> Concept ACK! I just wonder, will this method change again after the bdk-core integration, or do you think it can remain the same? I'd avoid breaking this API in 0.24, and then immediately deprecate `iter_unspent` in 1.0 From my understanding of bdk_core, I think it can remain the same, but I might be wrong

Good point. Let's discuss later today.

rajarshimaitra:
Concept ACK..

PR #767, opened 2022-09-29
Remove generic `D: Database` from `CoinSelectionAlgorithm`

### LINKED ISSUE #281: Remove Database type paramter from CoinSelectionAlgorithm
The rational for adding the database was iirc that the user may have some custom database and their coin selection algorithm may be coupled with particular data in their database.

This could easily be the case but it could also be that they need look up information from an internal HTTP API for some kind of business logic.
Wherever they get their additional data from it can be done without implementing a coin selection algorithm for a particular database. Instead, the type implementing `CoinSelectionAlgorithm` can just store a reference to the database/api where it gets its additional data from.

To fix this issue I suggest:
- Remove the type parameter from `CoinSelectionAlgorithm` and just pass in a `&dyn Database`.
- Make TxBuilder take a `& impl CoinSelectionAlgorithm` to `coin_selection` (i.e. take a reference).

evanlinjin:
Fixes #281 

### Description

Since `CoinSelectionAlgorithm` is a trait, if the database is really needed, one can create a struct that contains a database field which implements the `CoinSelectionAlgorithm` trait.

Most `CoinSelectionAlgorithm` implementations do not require a database.

### Changelog notice

The `CoinSelectionAlgorithm` trait no longer depends on a database.

#### All Submissions:

#### New Features:

* [x] This pull request breaks the existing API (it removes a single input from a trait's method)

### DIFF
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -34,10 +34,9 @@
 //! #[derive(Debug)]
 //! struct AlwaysSpendEverything;
 //!
-//! impl<D: Database> CoinSelectionAlgorithm<D> for AlwaysSpendEverything {
+//! impl CoinSelectionAlgorithm for AlwaysSpendEverything {
 //!     fn coin_select(
 //!         &self,
-//!         database: &D,
 //!         required_utxos: Vec<WeightedUtxo>,
 //!         optional_utxos: Vec<WeightedUtxo>,
 //!         fee_rate: FeeRate,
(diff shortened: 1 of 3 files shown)


evanlinjin:
**To do**

- [ ] `LargestFirstCoinSelection` should use `Ref<D>` as a field (otherwise it will be inconvenient/confusing to use). I've converted this to draft again.
- [ ] Write an example of using `LargestFirstCoinSelection` with `Wallet`.

PR #774, opened 2022-10-04
`FeeRate` calculation fixes

evanlinjin:
### Description

Weight units should always be whole units. However, virtual bytes will need decimal places and the conversion from weight unit to vbytes should not be rounded.

This commit has the following changes:
* vbytes should always be represented in `f32`
* conversion between vbytes and weight units should never be rounded
* simple calculations (such as those for feerates) should be explicitly defined (instead of using multiple small functions)

### Notes to the reviewers

Some "feerate" values for tests needed to be increased. This is because we were calling `.ceil()` on weight unit to vbytes conversion before (which is wrong, and results in more "vbytes" than what is real).

### Changelog notice

Improve accuracy of fee/feerate calculations and conversions. Also renamed the `Vbytes` trait to `WeightUnits`.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -98,13 +98,15 @@ impl FeeRate {
 
     /// Calculate fee rate from `fee` and weight units (`wu`).
     pub fn from_wu(fee: u64, wu: usize) -> FeeRate {
-        Self::from_vb(fee, wu.vbytes())
+        let vbytes = wu as f32 / 4.0;
+        let rate = fee as f32 / vbytes;
+        Self::new_checked(rate)
     }
 
     /// Calculate fee rate from `fee` and `vbytes`.
-    pub fn from_vb(fee: u64, vbytes: usize) -> FeeRate {
-        let rate = fee as f32 / vbytes as f32;
-        Self::from_sat_per_vb(rate)
+    pub fn from_vb(fee: u64, vbytes: f32) -> FeeRate {
+        let rate = fee as f32 / vbytes;
+        Self::new_checked(rate)
     }
 
     /// Return the value as satoshi/vbyte
@@ -114,12 +116,13 @@ impl FeeRate {
 
     /// Calculate absolute fee in Satoshis using size in weight units.
     pub fn fee_wu(&self, wu: usize) -> u64 {
-        self.fee_vb(wu.vbytes())
+        let vbytes = wu as f32 / 4.0;
+        (self.0 * vbytes).ceil() as u64
     }
 
     /// Calculate absolute fee in Satoshis using size in virtual bytes.
-    pub fn fee_vb(&self, vbytes: usize) -> u64 {
-        (self.as_sat_per_vb() * vbytes as f32).ceil() as u64
+    pub fn fee_vb(&self, vbytes: f32) -> u64 {
+        (self.0 * vbytes).ceil() as u64
     }
 }
 
@@ -140,13 +143,13 @@ impl Sub for FeeRate {
 /// Trait implemented by types that can be used to measure weight units.
 pub trait Vbytes {
     /// Convert weight units to virtual bytes.
-    fn vbytes(self) -> usize;
+    fn vbytes(self) -> f32;
 }
 
 impl Vbytes for usize {
-    fn vbytes(self) -> usize {
+    fn vbytes(self) -> f32 {
         // ref: https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#transaction-size-calculations
-        (self as f32 / 4.0).ceil() as usize
+        self as f32 / 4.0
     }
 }
 
--- a/src/types.rs
+++ b/src/types.rs
@@ -141,13 +141,13 @@ impl Sub for FeeRate {
 }
 
 /// Trait implemented by types that can be used to measure weight units.
-pub trait Vbytes {
+pub trait WeightUnits {
     /// Convert weight units to virtual bytes.
-    fn vbytes(self) -> f32;
+    fn to_vbytes(self) -> f32;
 }
 
-impl Vbytes for usize {
-    fn vbytes(self) -> f32 {
+impl WeightUnits for usize {
+    fn to_vbytes(self) -> f32 {
         // ref: https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#transaction-size-calculations
         self as f32 / 4.0
     }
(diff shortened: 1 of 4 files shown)


src/types.rs
```diff
@@ -114,12 +116,13 @@ impl FeeRate {
...
-        self.fee_vb(wu.vbytes())
+        let vbytes = wu as f32 / 4.0;
+        (self.0 * vbytes).ceil() as u64
     }
 
     /// Calculate absolute fee in Satoshis using size in virtual bytes.
-    pub fn fee_vb(&self, vbytes: usize) -> u64 {
-        (self.as_sat_per_vb() * vbytes as f32).ceil() as u64
+    pub fn fee_vb(&self, vbytes: f32) -> u64 {
+        (self.0 * vbytes).ceil() as u64
```
vladimirfomene, line 125: Why don't you use `self.as_sat_per_vb` instead of `self.0`? I feel it is more descriptive

src/types.rs
```diff
@@ -114,12 +116,13 @@ impl FeeRate {
 
     /// Calculate absolute fee in Satoshis using size in weight units.
     pub fn fee_wu(&self, wu: usize) -> u64 {
-        self.fee_vb(wu.vbytes())
+        let vbytes = wu as f32 / 4.0;
+        (self.0 * vbytes).ceil() as u64
```
vladimirfomene, line 120: With your change to the wu -> vb conversion, `self.fee_vb(wu.vbytes())` does exactly what these lines do?

src/types.rs
```diff
@@ -98,13 +98,15 @@ impl FeeRate {
...
+        Self::new_checked(rate)
     }
 
     /// Calculate fee rate from `fee` and `vbytes`.
-    pub fn from_vb(fee: u64, vbytes: usize) -> FeeRate {
-        let rate = fee as f32 / vbytes as f32;
-        Self::from_sat_per_vb(rate)
+    pub fn from_vb(fee: u64, vbytes: f32) -> FeeRate {
+        let rate = fee as f32 / vbytes;
+        Self::new_checked(rate)
```
vladimirfomene, line 109: I think using `Self::from_sat_per_vb(rate)` instead of  `Self::new_checked(rate)` is more descriptive

src/types.rs
```diff
@@ -98,13 +98,15 @@ impl FeeRate {
 
     /// Calculate fee rate from `fee` and weight units (`wu`).
     pub fn from_wu(fee: u64, wu: usize) -> FeeRate {
-        Self::from_vb(fee, wu.vbytes())
+        let vbytes = wu as f32 / 4.0;
+        let rate = fee as f32 / vbytes;
+        Self::new_checked(rate)
```
vladimirfomene, line 103: With your change to the wu -> vb conversion, `Self::from_vb(fee, wu.vbytes())` does exactly the same thing as the lines above.

src/types.rs
evanlinjin: As `as_sat_per_vb` is an external interface, I personally find that `self.0` is more "direct". For someone reading the code, it is less steps to figure out what exactly is going on. Instead of going, *"What does `as_sat_per_vb` do? Oh it just returns `self.0`"*.

src/types.rs
evanlinjin: It's better to be explicit with simple calculations.

danielabrozzoni:
> Weight units should always be whole units. 

Why? 1 WU = 0.25 sat/vbyte. If you can have 0.1 sat/vbyte, then you also need decimal weight units. Or am I misinterpreting what you're saying? (I haven't read the code yet)

Anywoo, as we were saying on discord, it would be cool to handle feerates the core way :tm: (https://discord.com/channels/753336465005608961/753367451319926827/1026784080714534932)

Should we just move to this method instead? If so, can you update this PR, or you have too much on your plate already? If you need help just ping :)

evanlinjin:
@danielabrozzoni 

> Why? 1 WU = 0.25 sat/vbyte. If you can have 0.1 sat/vbyte, then you also need decimal weight units. Or am I misinterpreting what you're saying? (I haven't read the code yet)

Think about it this way, if we look at the smallest possible "unit" of a transaction, for example, an `VarInt(1)` that records witness stack length, the weight of this `VarInt` is 1. This is the smallest possible "unit", and so, weight units should always be whole integers.

I'm talking about weight units (`wu`), not feerate (`sats/wu`). Feerates (whether `sats/wu` or `sats/vb`), should always be in floats (unless we work in kilo-{wu|vb} units).

> Should we just move to this method instead? If so, can you update this PR, or you have too much on your plate already? If you need help just ping :)

I agree that avoiding floats and working in kilo-everything seems to be the way to go. However, this PR is more of a bug fix, not an attempt at changing behavior.

If you look at how we did conversions from wu to vb, you see that vb is rounded up `.ceil`.

https://github.com/bitcoindevkit/bdk/blob/7de8be46c04128328d8dbd1de7b29033a290fde8/src/types.rs#L146-L151

This is only correct if we are considering a transaction as a whole (as per [BIP-0141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#transaction-size-calculations))

However, many times we are converting from wu to vb for only a portion of a tx (such as satisfaction weight for coin selection) and then adding the portions up. This will result in overshooting weight calculations of the entire tx by quite a bit.

evanlinjin:
Although this PR will increase the "correctness" of our weight calculations, if there are bugs in miniscript `max_satisfaction_weight` we may end up undershooting, which is dangerous. I think this should NOT be merged as of now until we confirm that miniscript is correct.

For reference:
* https://github.com/rust-bitcoin/rust-miniscript/pull/472
* https://github.com/rust-bitcoin/rust-miniscript/pull/473
* https://github.com/rust-bitcoin/rust-miniscript/pull/474

evanlinjin:
Also, if https://github.com/rust-bitcoin/rust-miniscript/pull/476 gets merged, we need to do +5 for [`TXIN_BASE_WEIGHT`](https://github.com/bitcoindevkit/bdk/blob/master/src/wallet/coin_selection.rs#L122).
* +1 WU for witness field stack size varint
* +4 WU for scriptSig size varint

LLFourn:
> This is only correct if we are considering a transaction as a whole (as per [BIP-0141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#transaction-size-calculations))
> 
> However, many times we are converting from wu to vb for only a portion of a tx (such as satisfaction weight for coin selection) and then adding the portions up. This will result in overshooting weight calculations of the entire tx by quite a bit.

I agree with this. The API should only take `fee_for(tx: &Transaction)` otherwise you are almost certainly using it incorrectly.

ValuedMammal:
@nondiremanuel This should be mostly resolved by #1216

PR #775, opened 2022-10-04
Workaround for `pkh` max satisfaction weight

evanlinjin:
### Description

There is a bug in miniscript where they fail to take into consideration an `OP_PUSH..` (4 weight units) when calculating
`max_satisfaction_weight` for `pkh` script types.

We add this weight explicitly for `pkh` script types during coin selection.

### Notes to the reviewers

The alternative is to patch miniscript (as @afilini mentioned).

### Changelog notice

Implement workaround for miniscript `max_satisfaction_weight` bug with `pkh` scripts.

#### All Submissions:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -30,7 +30,7 @@ use bitcoin::{
     Txid, Witness,
 };
 
-use miniscript::descriptor::DescriptorTrait;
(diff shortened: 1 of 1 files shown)


evanlinjin:
@afilini We call `.max_satisfaction_weight` in multiple places. We should do a patch like you mentioned. Closing for now.

Issue #695, opened 2022-07-31
Implement signature grinding


evanlinjin:
**Describe the bug**  
BDK does not do signature grinding for ecdsa signatures.

**Expected behavior**  
BDK should have signature grinding.

**Additional context**  

This is what BDK currently does.

https://github.com/bitcoindevkit/bdk/blob/a1477405d1adebd9a95689af17d3912edf1e57ef/src/wallet/signer.rs#L460-L463

But we have this to use.

https://github.com/rust-bitcoin/rust-secp256k1/blob/7fde3325075f16ec4b8c2c624fee523cc5225047/src/ecdsa/mod.rs#L341-L344
```rust
    /// Constructs a signature for `msg` using the secret key `sk`, RFC6979 nonce
    /// and "grinds" the nonce by passing extra entropy if necessary to produce
    /// a signature that is less than 71 - `bytes_to_grind` bytes. The number
    /// of signing operation performed by this function is exponential in the
    /// number of bytes grinded.
    /// Requires a signing capable context.
    pub fn sign_ecdsa_grind_r(&self, msg: &Message, sk: &SecretKey, bytes_to_grind: usize) -> Signature {
        let len_check = |s : &ffi::Signature| der_length_check(s, 71 - bytes_to_grind);
        self.sign_grind_with_check(msg, sk, len_check)
    }
```

**More context**

After we are certain that the BDK ECDSA signer(s) implementation uses signature grinding, we should optimize fee-rate calculation to take into consideration of grinding.

Do we need another signer method to inform whether the signer does ecdsa signature grinding?

LLFourn:
eak concept NACK.

If my understanding is up to date we currently  use `max_satisfaction_weight()` to figure out weight for coin selection which will not use the grinded value. So fixing this won't improve much other than increase the feerate.

It is also not possible to control this on most hardware wallets. If people want less weight I think they should just use taproot addresses rather than demand libraries to have this wacky parameter to save a few  bytes.

evanlinjin:
> eak concept NACK.
> 
> If my understanding is up to date we currently  use `max_satisfaction_weight()` to figure out weight for coin selection which will not use the grinded value. So fixing this won't improve much other than increase the feerate.
> 
> It is also not possible to control this on most hardware wallets. If people want less weight I think they should just use taproot addresses rather than demand libraries to have this wacky parameter to save a few  bytes.

In all else being said, increasing the fee rate slightly is a side-effect most people will be happy with imo.

[...]

PR #779, opened 2022-10-07
Add signature grinding for ECDSA signatures

vladimirfomene:
### Description

This PR adds a new field called `allow_grinding`
in the Signer's `SignOptions` struct that is used
to determine whether or not to grind an ECDSA signature during the signing process.

### Changelog notice

Breaking change: the BDK Signer now produces low-R signatures by default, saving one byte. If you want to preserve the original behavior, set `allow_grinding` in the `SignOptions` to `false`.

### Notes to the reviewers

This PR resolves issue #695 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -5524,6 +5524,7 @@ pub(crate) mod test {
                     SignOptions {
                         remove_partial_sigs: false,
                         try_finalize: false,
+                        allow_grinding: false,
                         ..Default::default()
                     },
                 )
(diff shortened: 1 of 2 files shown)


src/wallet/signer.rs
```diff
@@ -718,6 +724,11 @@ pub struct SignOptions {
     ///
     /// Defaults to `true`, i.e., we always try to sign with the taproot internal key.
     pub sign_with_tap_internal_key: bool,
+
+    /// Whether we should grind ECDSA signature to ensure signing with low r as bitcoin core
```
danielabrozzoni: nit: I'd remove the "as bitcoin core" part

src/wallet/mod.rs
```diff
@@ -5546,6 +5548,35 @@ pub(crate) mod test {
...
+                    remove_partial_sigs: false,
+                    allow_grinding: true,
+                    ..Default::default()
+                },
+            )
+            .unwrap();
+
+        let key = psbt.inputs[0].partial_sigs.keys().next().unwrap();
+        let sig_len = psbt.inputs[0].partial_sigs[key].sig.serialize_der().len();
+        assert!(sig_len < 71);
```
danielabrozzoni: nit: maybe assert that it's exactly 70?

danielabrozzoni (commented):
Concept ACK, the code looks good to me!

src/wallet/mod.rs
```diff
@@ -5546,6 +5548,35 @@ pub(crate) mod test {
         assert_fee_rate!(psbt, details.fee.unwrap_or(0), fee_rate);
     }
 
+    #[test]
+    fn test_fee_rate_sign_grinding_low_r() {
+        let (wallet, _, _) = get_funded_wallet("wpkh(tprv8ZgxMBicQKsPd3EupYiPRhaMooHKUHJxNsTfYuScep13go8QFfHdtkG9nRkFGb7busX4isf6X9dURGCoKgitaApQ6MupRhZMcELAxTBRJgS/*)");
```
rajarshimaitra, line 5557: A description of the goal of the test like the above one would be helpful..

rajarshimaitra (commented):
ACK.. Code LGTM..

src/wallet/mod.rs
vladimirfomene: Makes sense

rajarshimaitra (approved):
ACK 68dd6d20311b28f5c8e8c9657dce8cddae4f7aa3

danielabrozzoni:
ACK 68dd6d20311b28f5c8e8c9657dce8cddae4f7aa3

PR #785, opened 2022-10-26
Fix wallet export rescan height

LLFourn:
It would return the latest transaction height rather than the earliest as the height to rescan from.

Found by @evanlinjin  and I while implementing `bdk_core` stuff into bdk's wallet.

### Changelog notice

- Fix wallet export transaction height

#### All Submissions:

### DIFF
--- a/src/wallet/export.rs
+++ b/src/wallet/export.rs
@@ -134,15 +134,11 @@ impl FullyNodedExport {
         let blockheight = match wallet.database.borrow().iter_txs(false) {
             _ if !include_blockheight => 0,
             Err(_) => 0,
-            Ok(txs) => {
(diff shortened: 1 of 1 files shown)


notmandatory:
The **Changelog notice** section is where you'd put what ever change log message (if any) you want put in the release notes changelog (which we now manually put together and add to the release tag and on Github releases page).

rajarshimaitra (approved):
tACK e9bbb8724f92ada33228bbfc22acdaefbd79c44e

Yup makes sense to export the "sync from" height rather than last seen height..

notmandatory (approved):
ACK e9bbb8724f92ada33228bbfc22acdaefbd79c44e

PR #789, opened 2022-10-27
Add wasm compatible database using gloo_storage::LocalStorage

### LINKED ISSUE #450: Wasm compatible Database for bdk
Hi, i find that i need to have a database on wasm and i think i'm not the first nor the last. I'm currently working in the browser and it's all great with bdk but it's very cumbersome for the (future) users to wait for the synchronization every time.
So i think in two options:
- To make a version of MemoryDatabase in which we save the serialized structures on LocalStorage
- Implement the database tratis for IndexedDB. web_sys has support for IndexedDB so i think it can be a good possibility.
I really need this feature and I will have time for work on it, so what do you think is the best path and the requirements and niceties for this?

benthecarman:
### Description

Closes #450

Adds a new database type for wasm builds. This is basically copied from https://github.com/BitcoinDevShop/mutiny-web-poc/pull/5

### Notes to the reviewers

I am not sure if the tests are working correctly, couldnt figure out how to properly run with them without `wasm-pack`

#### All Submissions:

#### New Features:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ rocksdb = { version = "0.14", default-features = false, features = ["snappy"], o
 cc = { version = ">=1.0.64", optional = true }
 socks = { version = "0.3", optional = true }
 hwi = { version = "0.3.0", optional = true }
+gloo-storage = { version = "0.2.2", optional = true }
 
 bip39 = { version = "1.0.1", optional = true }
 bitcoinconsensus = { version = "0.19.0-3", optional = true }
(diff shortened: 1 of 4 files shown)


notmandatory:
@futurepaul  sorry for the delay in commenting on this PR, it looks like a great addition but right now we're trying to avoid adding new features like this to `bdk` until we know how it will fit in with the `bdk_core` release 1.0 refactoring that the team is focused on.  Rather than wait for that to shake out, since you're only implementing a `bdk` public trait how would you feel about making a little repo for this in your account? That way anyone who wants to try it out can easily add it as a dependency and you'll be able to iterate quickly on improvements. Then in a few months once the new bdk 1.0 APIs are solidified we can revisit pulling this into the `bitcoindevkit` org.

PR #793, opened 2022-11-03
bdk v1.0.0-alpha.0

LLFourn:
We prepare the BDK repo for a major restructuring 🔥. This PR maintains the existing wallet API as much as possible and adds very little. 

## Things Done

- database modules removed
- blockchain gutted but new esplora syncing code added (this will be gone soon hopefully).
- minimal API changes.
- Many macros removed.
- no longer applicable examples removed.
- Much conditional compilation removed. Can compile with `--all-features` now.
- All wallet tests passing
- TestClient moved into its own repo
- Example using `esplora` 

## APIs changed

- wallet no longer has a `sync` method. This is replaced with `apply_wallet_scan`.
- address "caching" is gone. You can just change the derivation index with `ensure_derived_up_to` which sets your derivation to at least the argument. Unlike `ensure_addresses_cached` used to do this will alter what getting a new address gives you.
- `AddressIndex::Reset` is gone. This thing didn't make much sense and is hard to do with the more sane internals we've established. Changing the derivation index changes what script pubkeys the wallet will search so this is dangerous. We plan to add method like `trim_unused` which lowers the derivation index to the highest unused index. Applications must handle giving out old addresses manually now (which I think is good).

## Unfinished work

- [x] esplora example doesn't work for mempool transactions yet (seems like our esplora in testclient doesn't index mempool??).
- [x] we need to figure out a way to retrieve and store transaction timestamps (we're currently just setting them to `u64::MAX`). In `bdk_core` we never got around to doing this but it needs to be done.
- [x] A few insights we got from doing this PR should be applied to bdk_core first.
- [x] doctests not working.

### Notes to the reviewers

Try not to review the actual changes. This PR will be forced pushed a bit so it will be likely wasted.
I think I did a faithful job of translating the tests. A bit of review here would be helpful.

I *do* think it would be good to merge this PR soon into the v1 branch so we have something to work off once unfinished work is done.
Checking out the branch and poke around and give feedback would be the most helpful thing.

Run the (sort of) working example:

```
cargo run --example esplora --features="bdk_test_client/bitcoind_22_0 bdk_test_client/esplora esplora"
```
#### All Submissions:

#### Bugfixes:

* [ ] I'm linking the issue being fixed by this PR (there's too many!)


.github/workflows/cont_integration.yml
```diff
@@ -14,21 +14,9 @@ jobs:
...
-          - compiler
-          - rpc
-          - verify
-          - async-interface
-          - use-esplora-async
-          - sqlite
-          - sqlite-bundled
+          - --no-default-features
+          - --all-features
+          - --features=default
```
danielabrozzoni: You should modify the build, clippy, test steps below:

Instead of `cargo build --features ${{ matrix.features }} --no-default-features`, it should be `cargo build ${{ matrix.features }}`

LLFourn: oops yep forgot to do that thanks. Looks like @evanlinjin has done this.

src/wallet/mod.rs
```diff
@@ -366,67 +332,236 @@ where
...
+            if let Some(prev_height) = height.checked_sub(1) {
+                let exp_hash = *update
+                    .checkpoints
+                    .entry(prev_height)
+                    .or_insert(block.header.prev_blockhash);
+
+                if block.header.prev_blockhash != exp_hash {
+                    panic!("block hash does not match!");
+                }
+            }
```
afilini: I guess otherwise you should check it's genesis block

src/blockchain/esplora.rs
```diff
@@ -0,0 +1,215 @@
...
+
+/// Extension trait for the blocking client
+pub trait BlockingClientExt {
+    /// Scan a single keychain given an iterator over its script pubkeys.
+    ///
+    /// Returns the index of the last script pubkey in the iterator that was present in a
+    /// transaction output along with the update. After `stop_gap` script pubkeys have no related
+    /// transactions it terminates (or if the iterator is exhausested).
+    ///
+    /// `parallel_requests` indicates that the client may make that may **additional** requests in
```
afilini: "that the client may make **additional** requests"

src/blockchain/esplora.rs
```diff
@@ -0,0 +1,215 @@
...
+                            }
+                        }
+                    }
+
+                    Result::<_, Error>::Ok((index, related_txs))
+                }))
+            })
+            .collect::<Vec<_>>();
+
+        let n_handles = handles.len();
```
afilini: This can be moved down i guess? `n_handles` is not used in the loop

src/blockchain/esplora.rs
```diff
@@ -0,0 +1,215 @@
...
+        let blocks_at_end = self
+            .get_blocks(None)?
+            .into_iter()
+            .map(|block| BlockId {
+                hash: block.id,
+                height: block.time.height,
+            })
+            .collect::<Vec<_>>();
+
+        if !blocks_at_end.contains(&tip_at_start) {
```
afilini: Esplora returns the last 10 blocks, so in theory here we should also make sure that the height of `tip_at_start` is in this range, otherwise we need to fetch more blocks.

src/blockchain/esplora.rs
```diff
@@ -0,0 +1,215 @@
...
```
afilini: Same as above

afilini (commented):
I still need to finish a couple of files but since it's taking me so long I'll start by posting those comments

src/wallet/mod.rs
```diff
@@ -121,15 +112,6 @@ pub enum AddressIndex {
...
     Peek(u32),
-    /// Return the address for a specific descriptor index and reset the current descriptor index
-    /// used by `AddressIndex::New` and `AddressIndex::LastUsed` to this value.
-    ///
-    /// Use with caution, if an index is given that is less than the current descriptor index
-    /// then the returned address and subsequent addresses returned by calls to `AddressIndex::New`
-    /// and `AddressIndex::LastUsed` may have already been used. Also if the index is reset to a
-    /// value earlier than the [`crate::blockchain::Blockchain`] stop_gap (default is 20) then a
-    /// larger stop_gap should be used to monitor for all possibly used addresses.
-    Reset(u32),
```
danielabrozzoni, line 132: Could you expand a bit on why is Reset being removed?

src/wallet/mod.rs
```diff
@@ -158,63 +140,78 @@ impl fmt::Display for AddressInfo {
...
-/// [`sync`]: Wallet::sync
-pub struct SyncOptions {
-    /// The progress tracker which may be informed when progress is made.
-    pub progress: Option<Box<dyn Progress>>,
+#[derive(Debug, Clone, PartialEq)]
+/// Error applying an update to the wallet.
+pub enum ApplyUpdateErr {
+    /// The data in the update was somehow incompatible with the existing chain's data.
+    Chain(UpdateFailure),
+    /// The update was compatible but it was missing **full** transactions for some txids it found.
```
danielabrozzoni: Just to make sure that I understand: this error is thrown if the Update is containing a certain `txidA` in the `new_txids`, but there's not a transaction with txid equal to `txidA` in the blocks between the last valid checkpoint and `new_tip`. Is that right?

Edit: I tried to read `apply_update` and concluded that I'm probably wrong

src/wallet/mod.rs
```diff
@@ -227,112 +224,82 @@ where
...
 
-        let found_used = self
-            .list_transactions(true)?
-            .iter()
-            .flat_map(|tx_details| tx_details.transaction.as_ref())
-            .flat_map(|tx| tx.output.iter())
-            .any(|o| o.script_pubkey == script_pubkey);
+        let address = definite_descriptor
+            .address(self.network)
+            .expect("descriptor must have address form");
```
danielabrozzoni: Does this mean that we're dropping support for descriptors without an address form, or that we trust our users to not call `get_new_address` if their descriptor can't have an address form?

src/wallet/mod.rs
```diff
@@ -366,67 +332,236 @@ where
...
-            .database
-            .borrow()
-            .iter_utxos()?
-            .into_iter()
-            .filter(|l| !l.is_spent)
-            .collect())
+    pub fn list_unspent(&self) -> Vec<LocalUtxo> {
+        self.spk_tracker
+            .iter_unspent(&self.sparse_chain, &self.tx_graph)
+            .map(|((keychain, _), txo)| LocalUtxo {
```
danielabrozzoni: This reminds me: the 1.0 release is a good opportunity to rename `LocalUtxo` to `LocalTxOut` :)

src/wallet/mod.rs
```diff
@@ -366,67 +332,236 @@ where
...
+                    }
+                }
+                None => continue,
+            }
+
+            // skip if already in chain
+            // we check chain instead of graph, as graph may be updated without commiting to chain
+            if self.sparse_chain.transaction_height(tx.txid()).is_some() {
+                continue;
+            }
```
danielabrozzoni: This checks seems relatively cheap, maybe it should be done first, before the match?

src/wallet/mod.rs
```diff
@@ -366,67 +332,236 @@ where
...
+    ) -> impl DoubleEndedIterator<Item = BlockId> + 'a {
+        self.sparse_chain.range_checkpoints(range)
+    }
+
+    /// Returns the latest checkpoint.
+    pub fn latest_checkpoint(&self) -> Option<BlockId> {
+        self.sparse_chain.latest_checkpoint()
+    }
+
+    /// Inserts a transaction into the graph if the transaction is deemed relevant.
```
danielabrozzoni: What does it mean for a tx to be relevant? To me, it seems that it means:
- it's a wallet transaction
- we have never seen it before

Is this correct?

danielabrozzoni (commented):
Just an initial review, I still need to look at half of `src/wallet/mod.rs` (I reviewed until before `apply_update`) and a bunch of other files. I'm posting this half-baked review so that you can start replying to all my questions (sorry, lol)

src/wallet/mod.rs
LLFourn: The reason is to create the invariant that every UTXO you have must have a derivation index less than or equal to your current derivation index. Clearly this cannot hold if you allow "resetting" your derivation index. Why do we want this invariant? It simplifies API and internals. There is no longer two indexes you are tracking: (i) where you are cached up to, (ii) where you are giving out addresses at. You just have one number which does handles both things.

However we should allow for "trimming" your derivation index down to your last active index. This will be needed in order to do `stop_gap` style syncing with things like CBF and `rpc` block streaming. The algorithm would be to set your derivation index to `last_active_index + stop_gap` before applying each block. Then when you've reached the tip you can trim to the `last_active_index` and the wallet becomes operational.

src/wallet/mod.rs
LLFourn: It is thrown when we are told there is a relevant `txid` in some block but we haven't provided the actual transaction. This comes up with electrum which first returns you a list of `txids` which you would go and turn into an update. The wallet would then tell you which transactions it actually needs you to download with that error. You go and fetch them and re-apply it.

src/wallet/mod.rs
LLFourn: The latter I think. Needs documentation.

src/wallet/mod.rs
LLFourn: This method was included by accident. Sorry! It was something sketched out earlier that won't ever exist I think.

LLFourn:
Just pushed changes to:

1. Remove some stuff that wasn't relevant to this PR and snuck through unintentionally. This includes `apply_blocks` which @evanlinjin pushed recently. We agreed that it would be better for that to come in with the changes that will use it.
2. Fix the esplora example by waiting a bit. H/t @afilini @danielabrozzoni.

tnull:
I'm currently trying wrap my head around how/if this new syncing mechanism could be used to better integrate with LDK. IIUC, the main idea here is that instead of having a monolithic `sync()` method, this follows three-step protocol:

1. the wallet generates a list of items (descriptors) it is interested in
2. this list is passed to the Esplora client which retrieves the relevant updates
3. the updates are applied to the `Wallet`

I think generally a similar scheme could be used to update the LDK-internal wallet (e.g., via the [`Confirm`](https://docs.rs/lightning/latest/lightning/chain/trait.Confirm.html) interface). However, I think in order to make this work we'd need (just some assorted notes for now):

- The methods allowing to express interest in a set of items and to apply them would need to be based on suitable abstractions (e.g, an `UpdatableWallet` trait or similar, etc.), which a part of LDK can implement/work with.
- To be able to express interest in the status of particular outputs and/or txids, not just descriptors.
-  Have access to additional data items for each applied update (block headers, index of transactions within a block, confirmation height) for any registered items (transactions/output spends).
- The ordering in which the updates are applied/returned needs to be deterministic and chain order. In particular any unconfirmations of items would need to be handed over before any reconfirmations.

Would be interested in any feedback regarding this, in particular whether we could find any abstractions that would allow us to further integrate the sync mechanisms or the wallets themselves, for that matter.

LLFourn:
>     * The methods allowing to express interest in a set of items and to apply them would need to be based on suitable abstractions (e.g, an `UpdatableWallet` trait or similar, etc.), which a part of LDK can implement/work with.
> 

Can these abstraction be defined in LDK or do you think they need to be in BDK?

>     * To be able to express interest in the status of particular outputs and/or txids, not just descriptors.

I added an issue for this: https://github.com/LLFourn/bdk_core_staging/issues/84. Currently we don't require you pass a descriptor (at least over at the bdk_core repo). You just pass in an iterator of script pubkeys you are interested in. I think we can add to that iterators of outpoints and txids to that.

> 
>     * Have access to additional data items for each applied update (block headers, index of transactions within a block, confirmation height) for any registered items (transactions/output spends).

We have designed the internals to allow this. I created an issue (https://github.com/LLFourn/bdk_core_staging/issues/85) for someone to actually do the work to include the all this info. Currently (though not in this PR yet) we only include the confirmation height as "additional data" (and sometimes the confirmation time as well). 

> 
>     * The ordering in which the updates are applied/returned needs to be deterministic and chain order. In particular any unconfirmations of items would need to be handed over before any reconfirmations.

I don't understand exactly this requirement but our update logic makes it hard to get this wrong I hope. There's an issue to make the ordering even more picky https://github.com/LLFourn/bdk_core_staging/issues/70 but I'm not sure if this addresses your concern. 

> 
> Would be interested in any feedback regarding this, in particular whether we could find any abstractions that would allow us to further integrate the sync mechanisms or the wallets themselves, for that matter.

Thanks for making this comment and apologies for the late reply. I'm going to be updating this PR soon so you can have a look at the state of the art.

tnull:
> Can these abstraction be defined in LDK or do you think they need to be in BDK?

Hm, I think BDK would need to provide some kind of API that we can work with, e.g., some trait-based interface that we could implement.

> I added an issue for this: [LLFourn/bdk_core_staging#84](https://github.com/LLFourn/bdk_core_staging/issues/84). Currently we don't require you pass a descriptor (at least over at the bdk_core repo). You just pass in an iterator of script pubkeys you are interested in. I think we can add to that iterators of outpoints and txids to that.

> We have designed the internals to allow this. I created an issue ([LLFourn/bdk_core_staging#85](https://github.com/LLFourn/bdk_core_staging/issues/85)) for someone to actually do the work to include the all this info. Currently (though not in this PR yet) we only include the confirmation height as "additional data" (and sometimes the confirmation time as well).

Nice, looking forward to these!

> I don't understand exactly this requirement but our update logic makes it hard to get this wrong I hope. There's an issue to make the ordering even more picky [LLFourn/bdk_core_staging#70](https://github.com/LLFourn/bdk_core_staging/issues/70) but I'm not sure if this addresses your concern.

Yeah, I think the details of the implementation should lie on our side, but the main point w.r.t. BDK would be that the interface would give us the updates in topological chain order.

> Thanks for making this comment and apologies for the late reply. I'm going to be updating this PR soon so you can have a look at the state of the art.

Sounds good, thank you! :)

LLFourn:
Status update.

Here's the scope of the PR and where we are on it:

1. We've decided to expand the scope of this PR to include persistence and as we agreed it is important to guarantee that every time you give out an address you have persisted that fact before you return it.
2. We've decided to turn the repo in to a cargo workspace and move all the concrete blockchain data sources  out of the `bdk` crate so it has minimal dependencies. I'll do that as part of this PR with a minimal working.

So remaining:
- [ ] Make wallet auto-persist to something with a simple trait that allows you to insert changesets perhaps with some distinction between updating chain data and derivation indices. Derivation indices must be guaranteed while chain data can just be refetched.
- [ ] Fix docs and doc tests
- [ ] Make repo into a workspace.

There is one issue that should be done in bdk_core staging: https://github.com/LLFourn/bdk_core_staging/issues/114

LLFourn:
Status update:

1. I implemented persistence as I imagined. See `persist.rs`. Any change you make to the wallet will be `staged` and then you can call `commit` to save it to the persistence backend. Some changes (like getting a new address) will be committed right away for safety reasons. 
2. All wallet tests have been moved into the `/tests` folder.
3. Added `Wallet::cancel_tx` which allows you to tell the wallet you will no longer be using a tx that the builder generated. This means the wallet will use that change address in the future (since it was never broadcast). In the future this will also unreserve utxos (see bitcoindevkit/bdk_wallet#166).

Remaining for `v1.0.0-alpha.0` (help needed):

1. Doc review and making doc tests work
2. Make bdk into a workspace
3. put `bdk_esplora` and `bdk_electrum` ported from `bdk_core_staging` examples.

LLFourn:
Just reading back @tnull's comments:

> Yeah, I think the details of the implementation should lie on our side, but the main point w.r.t. BDK would be that the interface would give us the updates in topological chain order.

Ok so our design does not enforce this: "changesets" can insert transactions that are ancestors of transactions that have already appeared in previous changesets. This will never happen in most workflows (and can be guaranteed never to happen if you update your state purely by applying new blocks sequentially). 

Other than sending bdk's changesets over to LDK maybe we can just create a custom "update" for it from our own chain data. We store and index a lot of stuff in memory so if the requirement is answering questions like "have these outputs been spent and by what tx" it should be easy. More sophisticated things like "notify me *when* this output is spent (or is no longer spent)" should be doable as well (but not yet implemented in an easy to consume way).

LLFourn:
One thing I think I forgot to implement is an API to expose `SpkTxOutIndex::send_and_received` method which is needed for it to be feature compatible with legacy `bdk` since `TransactionDetails` had `sent` and `received` fields and that was stored in the DB. From now on we want to calculate it on the fly.

Cargo.toml
```diff
@@ -122,25 +69,16 @@ required-features = ["compiler"]
...
+name = "mnemonic_to_descriptors"
+path = "examples/mnemonic_to_descriptors.rs"
+required-features = ["all-keys"]
 
 [workspace]
-members = ["macros"]
+members = ["bdk_test_client"]
 [package.metadata.docs.rs]
-features = ["compiler", "electrum", "esplora", "use-esplora-blocking", "compact_filters", "rpc", "key-value-db", "sqlite", "all-keys", "verify", "hardware-signer"]
+all-feautres = true
```
notmandatory: typo
```suggestion
all-features = true
```

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "bdk"
-version = "0.24.0"
+version = "0.27.0"
```
notmandatory: May as well update this as part of this PR:
```suggestion
version = "1.0.0-alpha"
```

LLFourn:
I've just make the repo into a workspace. Existing `bdk` is in `crates` (it will be joined by `bdk_chain` and `bdk_esplora` etc eventually). There is also an `example-crates` directory where you can put examples that are crates themselves and add them to the workspace. I put to placeholder crates there to demonstrate for electrum and esplora.

The final task for `bdk-v1.0.0-aplha.0` is to make those two example crates using `Wallet`, to clean up documentation/README and make CI pass.

.github/workflows/code_coverage.yml
```diff
@@ -38,13 +38,7 @@ jobs:
...
         run: if [[ ! -e ~/.cargo/bin/grcov ]]; then cargo install grcov; fi
       - name: Test
-        # WARNING: this is not testing the following features: test-esplora, test-hardware-signer, async-interface
-        # This is because some of our features are mutually exclusive, and generating various reports and
-        # merging them doesn't seem to be working very well.
-        # For more info, see:
-        # - https://github.com/bitcoindevkit/bdk/issues/696
-        # - https://github.com/bitcoindevkit/bdk/pull/748#issuecomment-1242721040
-        run: cargo test --features all-keys,compact_filters,compiler,key-value-db,sqlite,sqlite-bundled,test-electrum,test-rpc,verify
+        run: cargo test --all-features
```
danielabrozzoni, line 41: If you want to test with the feature hardware-signer, you need to install hwi and start the hardware simulator. See the last commit of https://github.com/bitcoindevkit/bdk/pull/873

LLFourn: Ah I see so we didn't test with it before. I'm not sure I want to fix that but I also don't want to go back to feature listing. What do you think about commenting this test out for now and doing #872 before `v1.0.0`?

LLFourn: (I merged your fix so let's see how far that gets us).

.github/workflows/cont_integration.yml
```diff
@@ -52,76 +40,12 @@ jobs:
...
-            target
-          key: ${{ runner.os }}-cargo-test-md-docs-${{ hashFiles('**/Cargo.toml','**/Cargo.lock') }}
-      - name: Set default toolchain
-        run: rustup default nightly
-      - name: Set profile
-        run: rustup set profile minimal
-      - name: Update toolchain
-        run: rustup update
-      - name: Test
-        run: cargo test --features test-md-docs --no-default-features -- doctest::ReadmeDoctests
```
danielabrozzoni, line 83: I suppose we're still going to have some examples in the README, so we shouldn't remove the checks from the CI

crates/bdk/src/descriptor/mod.rs
```diff
@@ -682,23 +708,40 @@ mod test {
 
         let secp = Secp256k1::new();
 
-        let xpub = bip32::ExtendedPubKey::from_str("xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL").unwrap();
+        let xprv = bip32::ExtendedPrivKey::from_str("xprv9s21ZrQH143K3c3gF1DUWpWNr2SG2XrG8oYPpqYh7hoWsJy9NjabErnzriJPpnGHyKz5NgdXmq1KVbqS1r4NXdCoKitWg5e86zqXHa8kxyB").unwrap();
         let path = bip32::DerivationPath::from_str("m/0").unwrap();
 
         // here `to_descriptor_key` will set the valid networks for the key to only mainnet, since
         // we are using an "xpub"
```
danielabrozzoni: Old comment

crates/bdk/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+//!
+//! If you are new to BDK we recommended that you use the default features which will enable
+//! basic descriptor wallet functionality. More advanced users can disable the `default` features
+//! (`--no-default-features`) and build the BDK library with only the features you need.
+
+//! Below is a list of the available feature flags and the additional functionality they provide.
+//!
+//! * `all-keys`: all features for working with bitcoin keys
+//! * `async-interface`: async functions in bdk traits
+//! * `keys-bip39`: [BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) mnemonic codes for generating deterministic keys
```
danielabrozzoni: I'd modify it so that it's updated to the current state of the features:
```
diff --git a/crates/bdk/src/lib.rs b/crates/bdk/src/lib.rs
index e48d356..0b190c9 100644
--- a/crates/bdk/src/lib.rs
+++ b/crates/bdk/src/lib.rs
@@ -81,7 +81,7 @@
 //!
 //! BDK uses a set of [feature flags](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section)
 //! to reduce the amount of compiled code by allowing projects to only enable the features they need.
-//! By default, BDK enables two internal features, `key-value-db` and `electrum`.
+//! By default, BDK enables only the `std` internal feature.
 //!
 //! If you are new to BDK we recommended that you use the default features which will enable
 //! basic descriptor wallet functionality. More advanced users can disable the `default` features
@@ -90,8 +90,9 @@
 //! Below is a list of the available feature flags and the additional functionality they provide.
 //!
 //! * `all-keys`: all features for working with bitcoin keys
-//! * `async-interface`: async functions in bdk traits
 //! * `keys-bip39`: [BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) mnemonic codes for generating deterministic keys
+//! * `file-store`: enables the storage of the blockchain data
+//! * `compiler`: miniscript compiler
```
I omitted `hardware-signer` because of #872

Feel free to adapt it if you don't like it though!

crates/bdk/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+//!   single-sig wallets, multisigs, timelocked contracts and more.
+//! * It supports multiple blockchain backends and databases, allowing developers to choose exactly what's right for their projects.
+//! * It is built to be cross-platform: the core logic works on desktop, mobile, and even WebAssembly.
+//! * It is very easy to extend: developers can implement customized logic for blockchain backends, databases, signers, coin selection, and more, without having to fork and modify this library.
+//!
+//! ## Generate a few addresses
+//!
+//! ### Example
+//! ```
+//! use bdk::{Wallet};
```
danielabrozzoni: nit: `use bdk::Wallet`

danielabrozzoni (commented):
This is looking really nice! I have a few docs-related nits

src/wallet/mod.rs
evanlinjin: Personally I think we should re-enable this feature to reset to the largest of:

1. First unused index
2. Parameter in `Reset(param)`

It shouldn't be complicated to implement with the `KeychainTxOutIndex` API.

LLFourn: I think we would need to "unreveal" method in `KeychainTxOutIndex` first I think. I still find this very questionable. Why would you reveal an address and then not actually reveal it. In any case this feature shouldn't be done as an `AddressIndex`.

.github/workflows/cont_integration.yml
LLFourn: I was going to change how this is done actually. I was just gonna include the README as a doc in the `lib.rs` using: https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html

src/wallet/mod.rs
LLFourn: Note i didn't do this here. I think it's possible we could get rid of this type in favor of using a `bdk_chain` one.

notmandatory (approved):
ACK 8d56d5503d97a3c60ccb1ae966f89b38b23f67af

CI updated, and tests passing. Organization looks good. There are small nits we could still fix but it's good enough for a first alpha release and will be easier to collaborate on additional changes once this PR is merged.

danielabrozzoni (approved):
I pushed a commit with some small doc fixes.

ACK bc3e05c6c671e927d0f954dac602f89a452b3d86

PR #796, opened 2022-11-11
Remove deprecated code and feature aliases

notmandatory:
### Description

A little housekeeping to remove recent and older deprecated code and feature aliases.

### Notes to the reviewers

I also removed the descriptor checksum inception check and corresponding test since the ability to make bad checksums was deprecated and now removed.

### Changelog notice

Removed deprecated
- get_checksum_bytes and get_checksum functions
- ConfirmationTime type
- WalletExport type
- Wallet.new_offline() function
- Cargo feature aliases use-esplora-request and use-esplora-ureq

#### All Submissions:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -59,7 +59,7 @@ use utils::{check_nsequence_rbf, After, Older, SecpCtx};
 use crate::blockchain::{GetHeight, NoopProgress, Progress, WalletSync};
 use crate::database::memory::MemoryDatabase;
 use crate::database::{AnyDatabase, BatchDatabase, BatchOperations, DatabaseUtils, SyncTime};
-use crate::descriptor::checksum::calc_checksum_bytes_internal;
+use crate::descriptor::checksum::calc_checksum_bytes;
 use crate::descriptor::policy::BuildSatisfaction;
 use crate::descriptor::{
     calc_checksum, into_wallet_descriptor_checked, DerivedDescriptor, DescriptorMeta,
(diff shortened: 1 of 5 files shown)


src/wallet/mod.rs
```diff
@@ -1894,12 +1875,6 @@ pub(crate) mod test {
         Wallet::db_checksum(&mut db, &desc, KeychainKind::External)
             .expect("db that uses actual checksum should be supported");
 
-        let mut db = MemoryDatabase::new();
-        db.check_descriptor_checksum(KeychainKind::External, checksum_inception)
-            .expect("failed to save checksum inception");
-        Wallet::db_checksum(&mut db, &desc, KeychainKind::External)
-            .expect("db that uses checksum inception should be supported");
-
```
rajarshimaitra, line 1902: I am not entirely sure, but will this create a hard error for database that had "inception bug" checksum?? In v0.24, the logic was if the db had such bad checksum, it will be accepted in the test.. But the bad checksum would still remain in db.. So when someone updates from v0.24 to v0.25 with bad checksum, this should now then cause hard error?

rajarshimaitra (commented):
Code review ACK.. Changes looks good to me.. One comment regarding the checksum inception bug..

src/wallet/mod.rs
notmandatory: I don't think just using the new checksum will cause a hard error. But if a wallet is created with a name and checksum A but then that same wallet name is used and the descriptor now  has checksum B that will be an error.  I expect most wallets use the `Default` name which is the checksum inself, so changing this checksum code will just create a new wallet name and duplicate the cache of wallet data on sync. Either way I think we've been clear enough in the deprecation last release that this is changing.

notmandatory:
This isn't urgent so moving to the 0.26 release

notmandatory:
Closing this, it's a lot to review and can be done as part of the bdk_core 1.0 release.

Issue #804, opened 2022-11-25
Ability to set *both* stop_gap and validate_domain for ElectrumBlockchain


icota:
Naive "fix" attempt in https://github.com/bitcoindevkit/bdk/pull/805

rajarshimaitra:
This can be closed after merge of #805 ..

PR #805, opened 2022-11-25
electrum: add validate_domain to ElectrumBlockchainConfig

icota:
### Description

The purpose of the PR is to be able to configure both `stop_gap` **and** `validate_domain`. Perhaps there are nicer ways.

#### All Submissions:

#### Bugfixes:

Issue in https://github.com/bitcoindevkit/bdk/issues/804

### DIFF
--- a/src/blockchain/any.rs
+++ b/src/blockchain/any.rs
@@ -178,7 +178,8 @@ impl_from!(boxed rpc::RpcBlockchain, AnyBlockchain, Rpc, #[cfg(feature = "rpc")]
 ///    "type" : "electrum",
 ///    "url" : "ssl://electrum.blockstream.info:50002",
 ///    "retry": 2,
-///    "stop_gap": 20
+///    "stop_gap": 20,
+///    "validate_domain": true
 /// }"#,
 /// )
 /// .unwrap();
@@ -190,6 +191,7 @@ impl_from!(boxed rpc::RpcBlockchain, AnyBlockchain, Rpc, #[cfg(feature = "rpc")]
 ///         socks5: None,
 ///         timeout: None,
 ///         stop_gap: 20,
+///         validate_domain: true,
 ///     })
 /// );
 /// # }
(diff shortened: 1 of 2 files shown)


tnull (commented):
This currently fails multiple tests in which the new field is missing from `ElectrumBlockchainConfig`.

icota:
@tnull this PR is more of a conversation starter since I'm not sure breaking the API would be acceptable anyway. I can close it and we can discuss in https://github.com/bitcoindevkit/bdk/issues/804

notmandatory:
I think it's OK to implement this as a breaking change as you've done, but you also need to update the related tests that create a new `ElectrumBlockchainConfig`, setting your new parameter to false.

afilini:
Concept ACK, but the default should be `true`, right? I guess it makes more sense to opt-out of validating a domain, the default should be to do it

tnull:
> Concept ACK, but the default should be `true`, right? I guess it makes more sense to opt-out of validating a domain, the default should be to do it

I had the same thought, but I believe the default returned by `ConfigBuilder::new()` is actually `true`:
https://github.com/bitcoindevkit/rust-electrum-client/blob/129081999ce96f1ed33877a2aecad4e2550764f3/src/config.rs#L119-L128

However, I agree that it should probably also be set to `true` in the doc example given.

Also, needs a rebase now to make CI happy.

BitcoinQnA:
> Concept ACK, but the default should be `true`, right? I guess it makes more sense to opt-out of validating a domain, the default should be to do it

@afilini forgive my naivety here, but how would this default affect BDK users connecting to their own node at home? The default for this is over Tor, which carries no domain/cert.

notmandatory:
You all are right the default for `validate_domain` in the examples should be `true` since this is what the electrum client `Config` already defaults to.

@icota this PR also needs to be rebased to fix a recent CI issue. Thanks!

notmandatory (approved):
ACK 2451c00268a8b518212c59132c263bc186d1b0f9

PR #812, opened 2022-12-06
Implement ordering for TransactionDetails

benthecarman:
### Description

Pulled from https://github.com/BitcoinDevShop/mutiny-web-poc/pull/189

Wallets should be able to sort the transactions easily, this makes it so you can just all `sort` on a list of tx details instead of needing to implement the sort_by yourself

#### All Submissions:

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -247,6 +247,20 @@ pub struct TransactionDetails {
     pub confirmation_time: Option<BlockTime>,
 }
 
+impl PartialOrd for TransactionDetails {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for TransactionDetails {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.confirmation_time
+            .cmp(&other.confirmation_time)
+            .then_with(|| self.txid.cmp(&other.txid))
+    }
+}
+
 /// Block height and timestamp of a block
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Default)]
 pub struct BlockTime {
@@ -254,6 +268,20 @@ pub struct BlockTime {
     pub height: u32,
     /// confirmation block timestamp
     pub timestamp: u64,
+}
+
+impl PartialOrd for BlockTime {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for BlockTime {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.height
+            .cmp(&other.height)
+            .then_with(|| self.timestamp.cmp(&other.timestamp))
+    }
 }
 
 /// **DEPRECATED**: Confirmation time of a transaction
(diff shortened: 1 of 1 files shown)


src/types.rs
```diff
@@ -247,6 +247,32 @@ pub struct TransactionDetails {
...
+    }
+}
+
+impl Ord for TransactionDetails {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // Sort by timestamp, but if there's no timestamp put it first,
+        // if timestamps are equal, compare by txid
+        match (&self.confirmation_time, &other.confirmation_time) {
+            (Some(a), Some(b)) => {
+                let cmp = a.timestamp.cmp(&b.timestamp);
```
tnull: Maybe it would make sense to implement `Ord` also for `BlockTime`, which would first compare the height then the timestamp. Here we then can just to compare the entire `confirmation_time`?

benthecarman: Good idea, done

src/types.rs
```diff
@@ -247,6 +247,32 @@ pub struct TransactionDetails {
...
+    }
+}
+
+impl Ord for TransactionDetails {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // Sort by timestamp, but if there's no timestamp put it first,
+        // if timestamps are equal, compare by txid
+        match (&self.confirmation_time, &other.confirmation_time) {
+            (Some(a), Some(b)) => {
+                let cmp = a.cmp(&b);
```
tnull: I think you can make this (and below) quite a bit more concise by chaining the `cmp`s via [`then_with`](https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.then_with). 

Btw., `Ord` is implemented for `Option<T>`, so you should be able to just do `self.confirmation_time.cmp(&other.confirmation_time).then_with(...)`, no need for the `match`.

src/types.rs
```diff
@@ -256,6 +282,23 @@ pub struct BlockTime {
...
 
+impl PartialOrd for BlockTime {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for BlockTime {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        let height_cmp = self.height.cmp(&other.height);
```
tnull: Same as above, you can just chain the `cmp`s via `then_with`

src/types.rs
benthecarman: Done, thanks!

src/types.rs
```diff
@@ -247,6 +247,21 @@ pub struct TransactionDetails {
...
 
+impl PartialOrd for TransactionDetails {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for TransactionDetails {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // Sort by timestamp, but if there's no timestamp put it first,
```
tnull: nit: I think this comment is superfluous now.

tnull (commented):
Looks good to me.

tnull:
Ah, this probably would also benefit from a really basic unit test checking that the ordering works as expected.

benthecarman:
Added a test

tnull (commented):
Looks good from my side!

notmandatory:
Looks like a good reasonable default ordering scheme to me. But you need to fix a clippy error, and rebase to fix a recent CI issue.

benthecarman:
> Looks like a good reasonable default ordering scheme to me. But you need to fix a clippy error, and rebase to fix a recent CI issue.

fixed

danielabrozzoni (approved):
utACK d3d07564f28b68723630fcb03becdc9d5b8d72ae

Thanks for taking care of this! :)

PR #813, opened 2022-12-07
Add `thiserror` for better display error formatting

### LINKED ISSUE #555: Error formatting is poor
Right now BDK just uses the output of `Debug` to create the display implementation. There should be a nice informative description for each error. See [`thiserror`](https://docs.rs/thiserror/latest/thiserror/).

yukibtc:
### Description

Resolve #555 

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/blockchain/compact_filters/mod.rs
+++ b/src/blockchain/compact_filters/mod.rs
@@ -50,7 +50,6 @@
 //! ```
 
 use std::collections::HashSet;
-use std::fmt;
 use std::path::Path;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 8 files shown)


src/blockchain/compact_filters/mod.rs
```diff
@@ -534,54 +533,62 @@ impl ConfigurableBlockchain for CompactFiltersBlockchain {
...
+    #[error("Peer doesn't advertise the BLOOM service flag")]
     PeerBloomDisabled,
 
     /// No peers have been specified
+    #[error("No peers have been specified")]
     NoPeers,
 
     /// Internal database error
+    #[error("Internal database error: {0}")]
     Db(rocksdb::Error),
```
rajarshimaitra: As we are using `thiserror` the from impls can also be automated using `Variant(#[from] Error_type)` syntax.. 

Same fro other conversions..

rajarshimaitra (changes requested):
Concept ACK..

One extra change for from impls as we are using `thiserror` so lets use its full power..

rajarshimaitra (commented):
Review ACK.. Changes looks good to me.. Would let others comment with their take on using `thiserror` as a dep..

notmandatory:
I'm in favor of this approach, but since it's adding a new dependency I think we should discuss also at our next bdk dev call, I've added it to the agenda.

notmandatory:
@afilini in our call today we discussed making `thiserror` an optional dependency, but on second thought I don't think there's a reason to make it optional since it [only adds macros](https://docs.rs/thiserror/latest/thiserror/#details) to generate nice looking `std::error::Error` implementations, so doesn't change the bdk lib API and has no requirement to also use `anyhow`.

danielabrozzoni:
> I don't think there's a reason to make it optional since it [only adds macros](https://docs.rs/thiserror/latest/thiserror/#details) to generate nice looking std::error::Error implementations

The reason for making it optional is that some projects want to cut dependencies to a minimum, and wouldn't use bdk if we include too many.

I agree with @afilini that it would be nice to have `thiserror` but behind a feature enabled by default

yukibtc:
Add `thiserror` as optional dependency, in my opinion, make the code more messed up, less maintainable.
If you prefer to have fewer dependencies, it's better to avoid to add `thiserror` also as optional dependency and continue in the way of PR #814.
The result of PR #813 and #814 for the dev that use BDK library is the same, but for you the work it's doubled when you need to make some changes to error enums (using the approach of `thiserror` as optional feature).
Keep both approach it's a bad idea, in my opinion.

notmandatory:
> Add `thiserror` as optional dependency, in my opinion, make the code more messed up, less maintainable. If you prefer to have fewer dependencies, it's better to avoid to add `thiserror` also as optional dependency and continue in the way of PR #814. The result of PR #813 and #814 for the dev that use BDK library is the same, but for you the work it's doubled when you need to make some changes to error enums (using the approach of `thiserror` as optional feature). Keep both approach it's a bad idea, in my opinion.

I agree if we want to avoid making `thiserror` a required dependency it's better to just go with your approach in #814. 

Reviewers, please check out #814, I've targeted it for the 0.27 release.

PR #815, opened 2022-12-08
Bump CI Rust stable version to 1.65.0 

yukibtc:
### Description

* Fix clippy for Rust 1.65.0  
* Bump CI Rust stable version to 1.65.0

#### All Submissions:

* [X] I've signed all my commits
* [X] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [X] I ran `cargo fmt` and `cargo clippy` before committing

PR #816, opened 2022-12-08
TOR+Esplora example

### LINKED ISSUE #196: Integrated TOR proxy
Create example code and blog post for how to use integrated rust tor client to sync bdk wallet with tor enabled blockchain service such as blockstream.info (http://explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion/). 

see:  

https://github.com/MagicalBitcoin/libtor
https://gitlab.torproject.org/tpo/core/arti

rorp:
### Description

A self contained example that demonstrates using a TOR proxy with the Esplora blockchain client.

Blog post PR https://github.com/bitcoindevkit/bitcoindevkit.org/pull/134

Fixes https://github.com/bitcoindevkit/bdk_wallet/issues/196

#### All Submissions:

#### New Features:

#### Bugfixes:


notmandatory:
Concept looks good, please rebase to fix the CI error.

examples/esplora_backend_with_tor.rs
```diff
@@ -0,0 +1,173 @@
...
+
+fn create_wallet(network: &Network, xpriv: &ExtendedPrivKey) -> Wallet<MemoryDatabase> {
+    Wallet::new(
+        Bip84(*xpriv, KeychainKind::External),
+        Some(Bip84(*xpriv, KeychainKind::Internal)),
+        *network,
+        MemoryDatabase::default(),
+    )
+    .unwrap()
+}
```
rajarshimaitra: This function seems redundant.. We can put this into the main itself..

examples/esplora_backend_with_tor.rs
```diff
@@ -0,0 +1,173 @@
...
+/// in the root folder.
+fn main() {
+    let network = Network::Signet;
+
+    let xpriv = "tprv8ZgxMBicQKsPcx5nBGsR63Pe8KnRUqmbJNENAfGftF3yuXoMMoVJJcYeUw5eVkm9WBPjWYt6HMWYJNesB5HaNVBaFc1M6dRjWSYnmewUMYy";
+
+    let esplora_url =
+        "http://explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion/testnet/api";
+
+    run(&network, esplora_url, xpriv);
```
rajarshimaitra: I don't think its useful to have a separate `run` function.. Just put everything inside main.. Its easier to read that way..

examples/esplora_backend_with_tor.rs
```diff
@@ -0,0 +1,173 @@
...
+    } else {
+        println!("Insufficient Funds. Fund the wallet with the address above");
+    }
+
+    let export = FullyNodedExport::export_wallet(&wallet, "exported wallet", true)
+        .map_err(ToString::to_string)
+        .map_err(bdk::Error::Generic)
+        .unwrap();
+
+    println!("------\nWallet Backup: {}", export.to_string());
```
rajarshimaitra: I know this is how the other esplora examples were written, but I don't think its useful to spit out an export.. Its kinda out of scope for the intent of the example..

examples/esplora_backend_with_tor.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    println!("Tor started");
+
+    format!("127.0.0.1:{}", socks_port)
+}
+
+fn truncate_log(filename: &String) {
+    let path = std::path::Path::new(filename);
+    if path.exists() {
+        let file = File::options()
```
rajarshimaitra: This will not work in `rust 1.56.1`.. You need to use `std::fs::OpenOption`.. Thats the error in CI..

Cargo.toml
```diff
@@ -146,13 +148,18 @@ required-features = ["electrum"]
 [[example]]
 name = "esplora_backend_synchronous"
 path = "examples/esplora_backend_synchronous.rs"
-required-features = ["use-esplora-ureq"]
+required-features = ["use-esplora-blocking"]
 
 [[example]]
 name = "esplora_backend_asynchronous"
 path = "examples/esplora_backend_asynchronous.rs"
 required-features = ["use-esplora-reqwest", "reqwest-default-tls", "async-interface"]
```
rajarshimaitra: If we are changing feature alias anyway, why not change this to `use-esplora-async` too.. Would be consistent with other examples..

examples/esplora_backend_with_tor.rs
```diff
@@ -0,0 +1,173 @@
...
+use esplora_client::Builder;
+use libtor::{LogDestination, LogLevel, Tor, TorFlag};
+
+pub mod utils;
+
+use crate::utils::tx::build_signed_tx;
+
+/// This will create a wallet from an xpriv and get the balance by connecting to an Esplora server
+/// over Tor network, using blocking calls with `ureq`.
+/// If enough amount is available, this will send a transaction to an address.
```
rajarshimaitra: nit 

```suggestion
/// If enough amount is available, this will send a transaction to it's own address.
```

rajarshimaitra (changes requested):
Concept ACK, thanks for the example.. I think its more useful to demo how to run tor from the code.. But good to see that its actually working.. I have funded the wallet address with some coins so hopefully the full example will work again.. 

Below are few general comments..

Cargo.toml
```diff
@@ -109,6 +109,8 @@ env_logger = "0.7"
 electrsd = "0.21"
 # Move back to importing from rust-bitcoin once https://github.com/rust-bitcoin/rust-bitcoin/pull/1342 is released
 base64 = "^0.13"
+libtor = "47.8.0+0.4.7.x"
+ureq = {version = "*", features = ["json", "socks-proxy"]}
```
rajarshimaitra: Why do we need `ureq`?

examples/esplora_backend_with_tor.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    let xpriv = bip32::ExtendedPrivKey::from_str(xpriv_str).unwrap();
+
+    let socks_addr = start_tor();
+
+    let client = Builder::new(esplora_url)
+        .proxy(&format!("socks5://{}", socks_addr))
+        .timeout(120) // seconds
+        .build_blocking()
+        .expect("Should never fail with no proxy and timeout");
```
rajarshimaitra: Just unwrap it instead of `expect`. And the expect msg seems wrong. We ARE using a proxy.

examples/esplora_backend_with_tor.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    let wallet = Wallet::new(
+        Bip84(xpriv, KeychainKind::External),
+        Some(Bip84(xpriv, KeychainKind::Internal)),
+        network,
+        MemoryDatabase::default(),
+    )
+    .unwrap();
+
+    println!("Syncing the wallet");
```
rajarshimaitra: Where did the transaction creation part go? That should be included. I was only talking about printing the wallet export as unnecessary..

rajarshimaitra (changes requested):
Needs few more changes.. And can you please squash the commits and rebase on master?

Cargo.toml
rorp: I believe BDK uses `ureq` under the hood to make HTTP requests. This line enables  `socks-proxy` feature, which is not enabled by default. This feature is required to connect to Tor. The example fails without it.

examples/esplora_backend_with_tor.rs
rorp: As per the issue `how to use integrated rust tor client to sync bdk wallet` so there's no need in  the transaction creation part at all.

rorp:
I'm sorry, I'm not a Git power user. I don't feel comfortable with interactive rebase. I always use GitHub's "Squash And Merge". It really does wonders, IMHO.

Cargo.toml
rajarshimaitra: No it doesn't.. You are using `esplora-client` internals to create a client which isn't the intended API in BDK.. You can write something like below and it would work. The `esplora-client` crate is already compiled with those features.. 

```
let esplora_config = EsploraBlockchainConfig{
        base_url: esplora_url.into(),
        proxy: Some(format!("socks5://{}", socks_addr)),
        concurrency: Some(4),
        stop_gap: 20,
        timeout: Some(60)
    };

let blockchain = EsploraBlockchain::from_config(&esplora_config).unwrap();
```

examples/esplora_backend_with_tor.rs
rajarshimaitra: I don't think so.. The sync doesn't make any sense if there's no wallet data to sync.. You are just creating an empty wallet and calling sync.. It doesn't complete in theory until you get to see an updated balance..

rajarshimaitra:
> I'm sorry, I'm not a Git power user. I don't feel comfortable with interactive rebase.

Probably a good time google on those and get those basics done as they are rudimentary to contributing to open source projects.. You don't need interactive rebase.. `git rebase master` from your branch will work.. But you need to sanitize the commit history before this can be merged..

Cargo.toml
rorp: Now I'm lost. Could you show exact code how to enable socks5 in `esplora-client`, please. This is my first BDK PR, I'm not that familiar with BDK's internals. Probably I'm missing something. 

I deleted the `ureq = ...` line from `Cargo.toml`, copy/pasted the code above, ran the example and got 
```bash
$ cargo clean
$ cargo run --features="use-esplora-blocking" --example esplora_backend_with_tor
    Finished dev [unoptimized + debuginfo] target(s) in 1.70s
     Running `target/debug/examples/esplora_backend_with_tor`
Staring Tor in /var/folders/38/pp4xfv2j7dbc6sr3gnln7jtr0000gn/T//bdk-tor-example
Dec 28 08:23:39.977 [notice] Tor 0.4.7.10 running on Darwin with Libevent 2.1.12-stable, OpenSSL 1.1.1s, Zlib 1.2.11, Liblzma N/A, Libzstd N/A and Unknown N/A as libc.
Dec 28 08:23:39.977 [notice] Tor can't help you if you use it wrong! Learn how to be safe at https://support.torproject.org/faq/staying-anonymous/
Dec 28 08:23:39.977 [notice] Configuration file ".torrc" not present, using reasonable defaults.
Dec 28 08:23:39.978 [notice] Opening Socks listener on 127.0.0.1:19050
Dec 28 08:23:39.978 [notice] Opened Socks listener connection (ready) on 127.0.0.1:19050
Tor started
Connecting to http://explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion/testnet/api via SOCKS5 proxy 127.0.0.1:19050
Syncing the wallet
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Esplora(Ureq(Transport(Transport { kind: ConnectionFailed, message: Some("Connect error"), url: Some(Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion")), port: None, path: "/testnet/api/scripthash/bb833461984bc2f06978e5e45c86478939c0d122625ec3b0a6a231e3c9b5378e/txs", query: None, fragment: None }), source: Some(Custom { kind: Other, error: "SOCKS feature disabled." }), response: None })))', examples/esplora_backend_with_tor.rs:69:54
```

rorp:
> Probably a good time google on those and get those basics done as they are rudimentary to contributing to open source projects.. 

Probably. Nobody complained about me not being a git power user for many years of contributing to open source projects, though. I'd rather keep using automation...

Cargo.toml
rorp: I noticed that BDK actually uses two different `ureq` versions

```bash
   Compiling ureq v2.2.0
   Compiling ureq v1.5.5
```

Maybe that's the root cause. One of them is compiled with socks and the other is not, and for some reason the linker uses the one without socks support...

Just a guess...

rajarshimaitra: Yeah this is a bug in `esplora-clinet` which is fixed here https://github.com/bitcoindevkit/rust-esplora-client/pull/34.. So this needs to wait for a new release of esplora-client..

rajarshimaitra:
@rorp this needs a rebase on master. With esplora_client updated now this configuration https://github.com/bitcoindevkit/bdk/pull/816#discussion_r1058119715 should work.

examples/esplora_backend_with_tor.rs
```diff
@@ -0,0 +1,178 @@
...
+        .timeout(120) // seconds
+        .build_blocking()
+        .unwrap();
+
+    println!(
+        "Connecting to {} via SOCKS5 proxy {}",
+        &esplora_url, &socks_addr
+    );
+
+    let blockchain = EsploraBlockchain::from_client(client, 20);
```
rajarshimaitra: This is a valid way of building the client, but not the intended API for bdk for such cases. 
Users using bdk, should not have to use `elsplora_client` in their dependency. Handling `esplora-client` is already done internally in bdk. 

`EsploraBlockhain` in bdk implements `ConfigurableBlockchain` which has an API `from_config`. https://github.com/bitcoindevkit/bdk/blob/eac739d3954937e49705a7336b4a017652802996/src/blockchain/esplora/blocking.rs#L215-L218

So you can create an esplora config using this structure https://github.com/bitcoindevkit/bdk/blob/eac739d3954937e49705a7336b4a017652802996/src/blockchain/esplora/mod.rs#L36-L62

Fields are public, so you can just set individual values.. 

And then just call `EsploraBlockhain::from_config(&config)`.

rajarshimaitra (changes requested):

examples/esplora_backend_with_tor.rs
rorp: 👍

w0xlt (commented):
Concept ACK

notmandatory:
@rorp we're getting an error with `rustls`. Can you check if there's a way to get around this without changing our MSRV to 1.57? Also please rebase to pickup our new stable rust version change from bitcoindevkit/bdk#815. Thanks! 

In the meantime I'm going to have to remove this from 0.27 milestone. 
```
error: package `rustls v0.20.8` cannot be built because it requires rustc 1.57 or newer, while the currently active rustc version is 1.56.1
```
See also: https://github.com/bitcoindevkit/bdk/actions/runs/3895859305/jobs/6739813321

notmandatory:
This should be ready to go after @rajarshimaitra's final review. I'll rebase/squash the commits for the original author.

rajarshimaitra:
It seems the signet esplora server the example was using went offline..
```
Staring Tor in /tmp/bdk-tor-example
Feb 21 18:49:57.051 [notice] Tor 0.4.7.13 running on Linux with Libevent 2.1.12-stable, OpenSSL 3.0.2, Zlib 1.2.11, Liblzma N/A, Libzstd N/A and Glibc 2.35 as libc.
Feb 21 18:49:57.051 [notice] Tor can't help you if you use it wrong! Learn how to be safe at https://support.torproject.org/faq/staying-anonymous/
Feb 21 18:49:57.051 [notice] Configuration file "/home/raj/.torrc" not present, using reasonable defaults.
Feb 21 18:49:57.052 [notice] Opening Socks listener on 127.0.0.1:19050
Feb 21 18:49:57.052 [notice] Opened Socks listener connection (ready) on 127.0.0.1:19050
Tor started
Connecting to http://explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion/testnet/api via SOCKS5 proxy 127.0.0.1:19050
Syncing the wallet
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Esplora(Ureq(Transport(Transport { kind: ConnectionFailed, message: Some("Connect error"), url: Some(Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion")), port: None, path: "/testnet/api/scripthash/2a69c29645663a231f39008866d65f855038570c1664987845f60dcad9f99253/txs", query: None, fragment: None }), source: Some(Custom { kind: TimedOut, error: "SOCKS proxy: explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion:80 timed out connecting after 29999ms." }) })))', examples/esplora_backend_with_tor.rs:83:54
```

Any other esplora onion endpoint we can use?

rajarshimaitra (approved):
ReACK d4b3f562d5088612959429add24b0fc93d80d539

Needs a rebase and its good to go..

notmandatory:
Thanks for all your work on this but we've gotten to the point where I'd like to hold off on merging any new features and only merge critical bug fixes to the `release/0.27` branch. 

Can you rebase this after the new `bdk_core_staging` stuff in bitcoindevkit/bdk#793 is merged to `master` branch? then we can merge it in for one of the upcoming 1.0.0-alpha releases.

danielabrozzoni:
Hey, can you please rebase this one on master?

rorp:
@danielabrozzoni Done!

Cargo.toml
```diff
@@ -1,5 +1,21 @@
...
+    "example-crates/keychain_tracker_example_cli",
+    "example-crates/wallet_electrum",
+    "example-crates/wallet_esplora",
+    "example-crates/wallet_esplora_async",
+    "example-crates/wallet_esplora_tor",
+    "nursery/tmp_plan",
+    "nursery/coin_select"
+]
+
+default-members = [
```
danielabrozzoni, line 18: Why do we need both members and default-members?

rorp: To prevent `example-crates/wallet_esplora_tor` to be built by default. It depends on `libtor`, which requires a C compiler, but BDK as a library doesn't and most important it shouldn't.

In this configuration one can build BDK without installing the additional dependencies simply by running
```bash
cargo build
```

And after installing the additional dependencies one can build this example with
```bash
cargo build -p bdk-esplora-wallet-tor-example
```

I agree, this is not the most elegant solution, so I'm open for suggestions how to improve it.

oleonardolima:
I'll open a new PR that covers Tor support relying on `arti-client` through https://github.com/bitcoindevkit/rust-esplora-client/pull/67.

Is there any active development on this one? cc @rorp @notmandatory

Otherwise, I can reuse the example of `libtor` here as an alternative there too.

notmandatory:
@oleonardolima I'll close this one so you can do a new PR based on the arti-client.

Issue #344, opened 2021-05-07
assert_matches!


RCasatta:
As briefly discussed in https://github.com/bitcoindevkit/bdk/pull/341#discussion_r627858886 and following

we should add https://docs.rs/crate/assert_matches/1.5.0 as a dev-dep which will also be standardized one day https://doc.rust-lang.org/nightly/std/macro.assert_matches.html

And changes our pattern matching test to be like:
```
assert_matches!(wallet.sign(&mut psbt, options), Err(Error::Signer(SignerError::InputIndexOutOfRange)))
```

eupn:
The [`matches!`](https://doc.rust-lang.org/std/macro.matches.html) macro has been stabilized since `1.42.0`.

We can ([and already do](https://github.com/bitcoindevkit/bdk/search?q=assert%21%28matches%21%28)) use it like this: `assert!(matches!(...))` to have behavior similar to `assert_matches!` (though the errors may be less informative).

This way we'll avoid an additional dev-dependency.

RCasatta:
> though the errors may be less informative

Yeah that's what I like, when `assert!(matches!())` test fails I end up printing the values before and launch tests with `--nocapture` which is unpleasant

LLFourn:
what's wrong with

```
assert_eq!(wallet.sign(&mut psbt, options), Err(Error::Signer(SignerError::InputIndexOutOfRange)))
```
?

RCasatta:
> what's wrong with
> 
> ```
> assert_eq!(wallet.sign(&mut psbt, options), Err(Error::Signer(SignerError::InputIndexOutOfRange)))
> ```
> 
> ?

You may want to test something that doesn't `impl Eq` or do stuff like:

```
assert_matches!(&policy.item, Signature(pk_or_f) if pk_or_f.fingerprint.unwrap() == fingerprint));
```


By the way, I think it would be good to have `assert_matches!` but if people have doubts it's absolutely fine without it

notmandatory:
closed by #821

PR #819, opened 2022-12-11
[344] Add assert_matches

Synesso:
### Description

Adds testing dependency `assert_matches` and applies to the to assertion mentioned in #344 

### Notes to the reviewers

I have only applied this macro to the location suggested in the issue and not elsewhere, as I am not familiar enough with the project.

#### All Submissions:


Synesso:
I saw build failures that seemed unrelated to my change, so I force pushed master/head to see if the build would still fail. Unfortunately that caused the PR to instantly close. Starting again...

PR #821, opened 2022-12-12
[#344] Add assert_matches

Synesso:
[#344] Add assert_matches

### DIFF
--- a/src/descriptor/checksum.rs
+++ b/src/descriptor/checksum.rs
@@ -133,6 +133,7 @@ pub fn get_checksum(desc: &str) -> Result<String, DescriptorError> {
 mod test {
     use super::*;
     use crate::descriptor::calc_checksum;
+    use assert_matches::assert_matches;
 
     // test calc_checksum() function; it should return the same value as Bitcoin Core
     #[test]
@@ -155,16 +156,16 @@ mod test {
         assert_eq!(calc_checksum(desc).unwrap(), "lasegmfs");
 
         let desc = "wpkh(tprv8ZgxMBicQKsPdpkqS7Eair4YxjcuuvDPNYmKX3sCniCf16tHEVrjjiSXEkFRnUH77yXc6ZcwHHcLNfjdi5qUvw3VDfgYiH5mNsj5izuiu2N/1/2/*)#tqz0nc26";
-        assert!(matches!(
-            calc_checksum(desc).err(),
-            Some(DescriptorError::InvalidDescriptorChecksum)
-        ));
+        assert_matches!(
+            calc_checksum(desc),
+            Err(DescriptorError::InvalidDescriptorChecksum)
+        );
 
         let desc = "pkh(tpubD6NzVbkrYhZ4XHndKkuB8FifXm8r5FQHwrN6oZuWCz13qb93rtgKvD4PQsqC4HP4yhV3tA2fqr2RbY5mNXfM7RxXUoeABoDtsFUq2zJq6YK/44'/1'/0'/0/*)#lasegmsf";
-        assert!(matches!(
-            calc_checksum(desc).err(),
-            Some(DescriptorError::InvalidDescriptorChecksum)
-        ));
+        assert_matches!(
+            calc_checksum(desc),
+            Err(DescriptorError::InvalidDescriptorChecksum)
+        );
     }
 
     #[test]
@@ -172,9 +173,9 @@ mod test {
         let sparkle_heart = unsafe { std::str::from_utf8_unchecked(&[240, 159, 146, 150]) };
         let invalid_desc = format!("wpkh(tprv8ZgxMBicQKsPdpkqS7Eair4YxjcuuvDPNYmKX3sCniCf16tHEVrjjiSXEkFRnUH77yXc6ZcwHHcL{}fjdi5qUvw3VDfgYiH5mNsj5izuiu2N/1/2/*)", sparkle_heart);
 
-        assert!(matches!(
-            calc_checksum(&invalid_desc).err(),
-            Some(DescriptorError::InvalidDescriptorCharacter(invalid_char)) if invalid_char == sparkle_heart.as_bytes()[0]
-        ));
+        assert_matches!(
+            calc_checksum(&invalid_desc),
+            Err(DescriptorError::InvalidDescriptorCharacter(invalid_char)) if invalid_char == sparkle_heart.as_bytes()[0]
+        );
     }
 }
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -581,6 +581,7 @@ impl DescriptorMeta for ExtendedDescriptor {
 mod test {
     use std::str::FromStr;
 
+    use assert_matches::assert_matches;
     use bitcoin::consensus::encode::deserialize;
     use bitcoin::hashes::hex::FromHex;
     use bitcoin::secp256k1::Secp256k1;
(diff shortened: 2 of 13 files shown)


evanlinjin:
Thank you for the PR. Couldn't we just use `assert_eq` instead of adding an external dependency?

danielabrozzoni:
Approved the CI run :)

> Thank you for the PR. Couldn't we just use assert_eq instead of adding an external dependency?

Note that it's just a dev dependency (it's useful only in tests), so it's not too bad. Also, see: https://github.com/bitcoindevkit/bdk/issues/344#issuecomment-873536670

Synesso:
Do these failing code cov and doc builds need to pass?

danielabrozzoni:
I merged #822 to fix the CI, try rebasing :)

notmandatory:
I'm all for this change since it's only a dev dependency and [assert_matches!](https://github.com/rust-lang/rust/issues/82775) is on track to end up in the `std` lib eventually. 

How would you fee about updating the ~29 spots in the code where we use `assert!(matches!(...))` to use `assert_matches!()` instead?

Synesso:
Updated @danielabrozzoni @notmandatory PTAL

Synesso:
Pushed a change to fix a linting error (seemed unrelated to my changes).

src/descriptor/checksum.rs
```diff
@@ -155,26 +156,26 @@ mod test {
         assert_eq!(calc_checksum(desc).unwrap(), "lasegmfs");
 
         let desc = "wpkh(tprv8ZgxMBicQKsPdpkqS7Eair4YxjcuuvDPNYmKX3sCniCf16tHEVrjjiSXEkFRnUH77yXc6ZcwHHcLNfjdi5qUvw3VDfgYiH5mNsj5izuiu2N/1/2/*)#tqz0nc26";
-        assert!(matches!(
+        assert_matches!(
             calc_checksum(desc).err(),
             Some(DescriptorError::InvalidDescriptorChecksum)
-        ));
+        );
```
danielabrozzoni: You can drop the `.err()` here:

```diff
         assert_matches!(
-            calc_checksum(desc).err(),
-            Some(DescriptorError::InvalidDescriptorChecksum)
+            calc_checksum(desc),
+            Err(DescriptorError::InvalidDescriptorChecksum)
         );
```

and same below

src/descriptor/mod.rs
```diff
@@ -852,10 +847,7 @@ mod test {
         let result = into_wallet_descriptor_checked(descriptor, &secp, Network::Testnet);
 
         assert!(result.is_err());
-        assert!(matches!(
-            result.unwrap_err(),
-            DescriptorError::HardenedDerivationXpub
-        ));
+        assert_matches!(result.unwrap_err(), DescriptorError::HardenedDerivationXpub);
```
danielabrozzoni: Can be written as
```
        assert_matches!(result, Err(DescriptorError::HardenedDerivationXpub));
```

and also you can drop the `assert!` in the line above. I think it was there because `assert!(matches!(` is not really informative when there's a failure - but the error reporting of `assert_matches` is fine, so we don't need those tricks anymore :)

src/wallet/mod.rs
```diff
@@ -4425,11 +4426,9 @@ pub(crate) mod test {
             result.is_err(),
             "Signing should have failed because the TX uses non-standard sighashes"
         );
-        assert!(
-            matches!(
-                result.unwrap_err(),
-                Error::Signer(SignerError::NonStandardSighash)
-            ),
+        assert_matches!(
+            result.unwrap_err(),
```
danielabrozzoni: You can avoid the `unwrap_err`, and match on `Err(Error::Signer(SignerError::NonStandardSighash))`

danielabrozzoni (commented):
A few nits, but otherwise it's looking really good!

Can you please squash your commits together when you update?

Synesso:
Squashed and requested changes applied. TY

src/descriptor/checksum.rs
```diff
@@ -155,26 +156,26 @@ mod test {
         assert_eq!(calc_checksum(desc).unwrap(), "lasegmfs");
 
         let desc = "wpkh(tprv8ZgxMBicQKsPdpkqS7Eair4YxjcuuvDPNYmKX3sCniCf16tHEVrjjiSXEkFRnUH77yXc6ZcwHHcLNfjdi5qUvw3VDfgYiH5mNsj5izuiu2N/1/2/*)#tqz0nc26";
-        assert!(matches!(
-            calc_checksum(desc).err(),
+        assert_matches!(
+            calc_checksum(desc),
             Some(DescriptorError::InvalidDescriptorChecksum)
```
danielabrozzoni: As the CI pointed out, this should be `Err(...)`, not `Some`

danielabrozzoni (commented):
There are still some instances of `.err()` and `.unwrap_err()` around, can you fix them?

rajarshimaitra:
I approved a run.. Lets see what happens.. :)

rajarshimaitra:
Congrats.. All checks passing.. Will do a deeper review tomorrow..

src/testutils/configurable_blockchain_tests.rs
```diff
@@ -1,8 +1,9 @@
+use assert_matches::assert_matches;
 use bitcoin::Network;
 
 use crate::{
-    blockchain::ConfigurableBlockchain, database::MemoryDatabase, testutils, wallet::AddressIndex,
-    Wallet,
+    blockchain::ConfigurableBlockchain, database::MemoryDatabase, error::Error,
+    signer::SignerError, testutils, wallet::AddressIndex, Wallet,
```
rajarshimaitra: I am seeing warning in the additions here as they aren't seemed to be used in this module..

src/testutils/configurable_blockchain_tests.rs
```diff
@@ -1,8 +1,9 @@
+use assert_matches::assert_matches;
```
rajarshimaitra: Same unused warning here.. `assert_matches` not used..

src/testutils/configurable_blockchain_tests.rs
```diff
@@ -234,6 +235,7 @@ where
     builder.add_recipient(address2.script_pubkey(), TRANSFER_AMOUNT / 2);
     let (mut psbt, details) = builder.finish().unwrap();
     assert!(wallet.sign(&mut psbt, Default::default()).unwrap());
+
```
rajarshimaitra: Why this new line?

rajarshimaitra (dismissed):
ACK.. No strong opinion on `assert_matches` over `matches`.. Below are few extra changes that I think isn't required..

src/testutils/configurable_blockchain_tests.rs
Synesso: Removed. I didn't see warnings in IntelliJ, nor from cargo. 
What should I be doing/using to see these?

src/testutils/configurable_blockchain_tests.rs
Synesso: fixed

src/testutils/configurable_blockchain_tests.rs
Synesso: fixed.

src/testutils/configurable_blockchain_tests.rs
rajarshimaitra: If you are using a editor linting, you need to activate certain features to activate those segment of code.. This warning was visible if `test-blockchain` feature was activated.. Different part of the code base are activated with different features, so without the feature on that code will not be analyzed and warning wont be shown in the editor.. I don't know how to do it in IntelliJ, must be some settings that you have to add..

The other way is just build with various features like `cargo build --feature test-blockchain` etc..

rajarshimaitra: I use rust-analyzer with VSCode.. But other editors should work too..

Synesso: Found some useful config in IJ. Thanks.
Is this good to merge now?

notmandatory (approved):
ACK 14bc9c0e356ac1c864e7c37aaadfe4696e999693

Thanks for updating all the `assert!(matches!())`.

notmandatory:
@rajarshimaitra if you're OK with the latest push I think you need to resolve your review changes or re-approve or just let me know and I'll go ahead with merging it.

danielabrozzoni (approved):
utACK 14bc9c0e356ac1c864e7c37aaadfe4696e999693

rajarshimaitra:
Sorry I was AFK for last 3 days.. Made another pass over this and all looks good to me..

PR #838, opened 2023-01-19
Add small clarification to docs

thunderbiscuit:
### Description
Very small fixes to documentation:
1. I got a DM last week from a user who thought we had a bug with our timestamps. It turns out he was using the milliseconds version of the Unix timestamp in his project and didn't realize we were giving out a standard Unix timestamp. The docs now mention this explicitly.
2. I noticed some small inconsistencies in the documentation on the public templates while porting them over to Kotlin. This PR also fixes that so that all templates use a common documentation wording.

#### All Submissions:

### DIFF
--- a/examples/compiler.rs
+++ b/examples/compiler.rs
@@ -29,7 +29,7 @@ use bdk::wallet::AddressIndex::New;
 use bdk::{KeychainKind, Wallet};
 
 /// Miniscript policy is a high level abstraction of spending conditions. Defined in the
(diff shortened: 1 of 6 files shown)


notmandatory (approved):
ACK bd18338bf2e08e6d463593a81b4d5901c30ce04a

notmandatory (approved):
Re-ACK f0cec015b59ab484e81ca675d1170af5eb553911

Putting this into 0.27.0 since it is only doc changes.

PR #858, opened 2023-02-10
Remove unused function calls and return values in sqlite

silvestrst-crypto:
### Description

Some method return values were never used, thus resulting in unnecessary calls to `self.connection.last_insert_rowid()`.

This change removes these calls, and changes the corresponding method return type to `Result<(), Error>`.

#### All Submissions:

* [x] I ran all `sqlite` tests successfully

### DIFF
--- a/src/database/sqlite.rs
+++ b/src/database/sqlite.rs
@@ -97,7 +97,7 @@ impl SqliteDatabase {
         keychain: String,
         child: u32,
         script: &[u8],
-    ) -> Result<i64, Error> {
+    ) -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


rajarshimaitra:
I think the intention of these returns was for internal debugging purposes. To help the caller check back which row got last modified after the call. Though we don't seem to use them much anywhere, not sure if removing them fully is okay..

silvestrst-crypto:
It just feels a little counter-intuitive to have unused code, that spans across two methods for debugging. I guess the alternative would be to put it inside `log::debug`, but that could get fairly verbose.

danielabrozzoni:

PR #876, opened 2023-02-28
Fix TxOrdering::Shuffle to also shuffle the inputs

### LINKED ISSUE #865: TxOdering::Shuffle is not shuffling the inputs
TxOdering::Shuffle is not shuffling the inputs (only the outputs).

Bug found while doing bdk core integration. I think it's moderately serious so I think this fix should be backported.

notmandatory:
### Description

Fixes #865 

This is a small change to make it so that `TxOrdering:Shuffle` also shuffles the inputs. 

### Notes to the reviewers

It looks like input shuffling was meant to be excluded in #38 when the `TxOrdering::Shuffle` option was added since the corresponding test check that inputs are not shuffled. 

### Changelog notice

Fixed

- The `TxOrdering::Shuffle` option in the `TxBuilder` now also shuffles the inputs. 

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/wallet/tx_builder.rs
+++ b/src/wallet/tx_builder.rs
@@ -703,6 +703,7 @@ impl TxOrdering {
                 #[cfg(test)]
                 let mut rng = rand::rngs::StdRng::seed_from_u64(12345);
 
+                tx.input.shuffle(&mut rng);
                 tx.output.shuffle(&mut rng);
             }
             TxOrdering::Bip69Lexicographic => {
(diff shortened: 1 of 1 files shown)


notmandatory:
@afilini I'm a little worried that you intended to not shuffle inputs since your test verifies they're not shuffled. Do you remember is there's any reason not to shuffle the inputs?

notmandatory:
This doesn't seem like a critical issue so I'll re-do it for a future 1.0.0 alpha release after #793 is merged.

notmandatory:
Already fixed!

PR #878, opened 2023-03-01
Copy stuff from bdk_core_staging to bdk 

danielabrozzoni:
This is based on #793 and copies all the repos from bdk_core_staging to bdk. I've used commit 250b4f1dcce10805adfb2f201901675eb6076554, I suppose we're going to frequently change it, at least while there are still PRs being merged on the other repo :) 

I still need to add bdk_esplora!


LLFourn:
I pushed commits to move everything else over and get `cargo test` in the workspace to work.

This was merged into #793. Thanks!

PR #879, opened 2023-03-02
#793: rebase, fix CI

danielabrozzoni:
Additions to #793:

- Disable the test for the hardware wallets. The latest version of HWI broke *something* and now I don't have the time to figure out what.
- Fix the clippy job, fix most of the clippy warnings
- Make bdk_file_store use bincode v1 so that our MSRV can be rust 1.57
- Make bdk_esplora compile in WASM. We still need to make it work though, see #880 
- CI uses actions-rs/toolchain for rust instead of manually downloading rustup
- Disable coveralls, we'll enable it eventually
- Add  the wallet_esplora example and the wallet_electrum example
- Misc


.github/workflows/cont_integration.yml
```diff
@@ -10,118 +10,34 @@ jobs:
...
-          - use-esplora-async
-          - sqlite
-          - sqlite-bundled
+          - --no-default-features
+          - --all-features
+          - --features=default
     steps:
       - name: checkout
         uses: actions/checkout@v2
       - name: Generate cache key
```
notmandatory, line 23: This can be removed since we're switching to actions/cache.

.github/workflows/code_coverage.yml
```diff
@@ -16,44 +16,37 @@ jobs:
...
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
         with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+            toolchain: "1.65.0"
```
notmandatory, line 22: I think this also be 1.67.0 to match stable version used in CI pipeline.
```suggestion
            toolchain: "1.67.0"
```

.github/workflows/cont_integration.yml
```diff
@@ -132,72 +48,39 @@ jobs:
...
-      - name: Add target wasm32
-        run: rustup target add wasm32-unknown-unknown
-      - name: Update toolchain
-        run: rustup update
-      - name: Check
-        run: cargo check --target wasm32-unknown-unknown --features async-interface,use-esplora-async,dev-getrandom-wasm --no-default-features
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+            toolchain: "1.65.0"
```
notmandatory, line 59: Same as above, good to use same stable version everywhere.
```suggestion
            toolchain: "1.67.0"
```

.github/workflows/cont_integration.yml
```diff
@@ -132,72 +48,39 @@ jobs:
...
-      - name: Set profile
-        run: rustup set profile minimal
-      - name: Add rustfmt
-        run: rustup component add rustfmt
-      - name: Update toolchain
-        run: rustup update
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+            toolchain: "1.65.0"
```
notmandatory, line 81:
```suggestion
            toolchain: "1.67.0"
```

nursery/coin_select/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "bdk_coin_select"
+version = "0.0.1"
+authors = [ "LLFourn <lloyd.fourn@gmail.com>" ]
```
notmandatory, line 4: Even though @LLFourn is the primary contributor on this, for consistency and since other's will be contributing in the future I think best not to have individual names here.
```suggestion
authors = ["Bitcoin Dev Kit Developers"]
```

notmandatory (commented):
I have a few small comments, no show stoppers. I primarily focused on the CI and overall organization not the new/refactored code. 

If @LLFourn is OK having this go in before #793 I think we should do it so new PRs towards our first alpha release will be easier to review.

LLFourn:
I reset #793 to ab9242d10ddaaafae3f6433966658af882d1f735 (the last commit here)

Issue #880, opened 2023-03-03
Implement EsploraExt for esplora_client::AsyncClient


danielabrozzoni:
**Describe the enhancement**  
https://github.com/bitcoindevkit/bdk/blob/97631328ff0dcfc9b72637cb0d6106dd3cb69e2e/crates/esplora/src/lib.rs#L309-L318
**Use case**  
Using bdk_esplora in wasm

evanlinjin:
This is done in #885 which is merged into #793. Essentially, we now have `EsploraAsyncExt`.

PR #885, opened 2023-03-07
Implement EsploraExt for async client

vladimirfomene:
### Description
This PR solves issue #880.  It is implementing the EsploraExt trait on the Esplora Async client by defining an async version of that trait and implementing it on the async esplora client. 

### Notes to the reviewers

I went with the copy pasting approach. I believe it is okay since we are just doing here.

#### All Submissions:

#### New Features:

#### Bugfixes:


crates/esplora/Cargo.toml
```diff
@@ -14,6 +14,8 @@ readme = "README.md"
 [dependencies]
 bdk_chain = { path = "../chain", version = "0.3.1", features = ["serde", "miniscript"] }
 esplora-client = { version = "0.3", default-features = false }
+async-trait = "0.1.66"
+futures = "0.3.26"
```
evanlinjin: IMO these should be optional dependencies. It might make it easier to have two files, one for `EsploraExt` and one for `EsploraAsyncExt`.

evanlinjin: I can get it done

evanlinjin: Completed in https://github.com/danielabrozzoni/bdk/pull/3

LLFourn:
This was merged into https://github.com/danielabrozzoni/bdk/pull/3 which got merged into #793

PR #889, opened 2023-03-15
Add docs for `EsploraAsyncExt` and make doctests runnable

evanlinjin:
### Description

Add docs for `EsploraAsyncExt` and make various doctests runnable.

### DIFF
--- a/crates/esplora/src/async_ext.rs
+++ b/crates/esplora/src/async_ext.rs
@@ -12,6 +12,13 @@ use futures::stream::{FuturesOrdered, TryStreamExt};
 
 use crate::map_confirmation_time;
 
+/// Trait to extend [`esplora_client::AsyncClient`] functionality.
+///
+/// This is the async version of [`EsploraExt`]. Refer to
+/// [crate-level documentation] for more.
+///
+/// [`EsploraExt`]: crate::EsploraExt
+/// [crate-level documentation]: crate
 #[cfg(feature = "async")]
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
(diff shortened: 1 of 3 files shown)


crates/esplora/src/async_ext.rs
```diff
@@ -12,6 +12,12 @@ use futures::stream::{FuturesOrdered, TryStreamExt};
 
 use crate::map_confirmation_time;
 
+/// Trait to extend [`esplora_client::AsyncClient`] functionality.
+///
+/// This is the async version of [`EsploraExt`]. Refer to [crate-level documentation] for more.
```
rajarshimaitra: I couldn't find any more info on the crate-level documentation. Maybe something more will be added later?

rajarshimaitra (commented):
tACK 136a0b130cd8fda167e099823ec775635b5f7cce

crates/esplora/src/async_ext.rs
evanlinjin: Woops you are right. Maybe "refer to crate level documentation for usage" will be better?

evanlinjin: Fixed.

rajarshimaitra: Umm.. It also doesn't have any usage docs too. Can't we just point to the `keychain_tracker_esplora` crate for usage? That's where they will find how to use this trait.

evanlinjin: How about now?

rajarshimaitra (approved):
ACK 877b6587875e75407567d9080f486eba2cb08460

ChidiChuks:
@evanlinjin Great work on this PR! Considering the increasing adoption of Miniscript for expressive and secure policy construction in Bitcoin, would it be worth referencing or aligning this implementation with ongoing Miniscript support? This could help ensure compatibility and future-proofing, especially for more complex script use cases.

PR #890, opened 2023-03-15
Simplify `reveal_to_target` logic

evanlinjin:
### Description

- Remove unnecessary check
- Better comments
- Better variable names
- Added another test

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -63,7 +63,7 @@ pub struct KeychainTxOutIndex<K> {
     inner: SpkTxOutIndex<(K, u32)>,
     // descriptors of each keychain
     keychains: BTreeMap<K, Descriptor<DescriptorPublicKey>>,
-    // last stored indexes
+    // last revealed indexes
     last_revealed: BTreeMap<K, u32>,
     // lookahead settings for each keychain
     lookahead: BTreeMap<K, u32>,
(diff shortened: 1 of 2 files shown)


1440000bytes:
test

crates/chain/src/keychain/txout_index.rs
```diff
@@ -381,40 +381,39 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
 
-        // if we can reveal new indexes, the latest revealed index goes here
-        let mut revealed_index = None;
+        debug_assert_eq!(
+            next_reveal_index + lookahead,
+            self.next_store_index(keychain)
+        );
+
+        // if we need to reveal new indexes, the latest revealed index goes here
+        let mut reveal_to_index = None;
```
vladimirfomene, line 393: Maybe you should call this `latest_reveal_index`. It might remove the need to have a comment.

vladimirfomene (approved):
Nice one. This version of the code is easy to follow. I just have a nit.

crates/chain/tests/test_keychain_txout_index.rs
```diff
@@ -210,13 +210,44 @@ fn test_lookahead() {
...
+// - scanning txouts with spks above last stored index
+// expect:
+// - last revealed index should increase as expected
+// - last used index should change as expected
+#[test]
+fn test_scan_with_lookahead() {
+    let (mut txout_index, external_desc, _) = init_txout_index();
+    txout_index.set_lookahead_for_all(10);
+
+    let spks: BTreeMap<u32, Script> = [0, 10, 20, 30]
```
LLFourn, line 225: :bulb: after the loop try with 41 and check that it doesn't find it

evanlinjin: Good idea

LLFourn (approved):

crates/chain/src/keychain/txout_index.rs
evanlinjin: I think we probably need a comment anyway to inform that this variable is to be mutated "if we need to reveal new indices".

PR #893, opened 2023-03-15
Set bitcoin, miniscript default-features = false, bump esplora-client version to 0.4

notmandatory:
### Description

Carrying over relevant maintenance changes from release 0.27.2 to bdk `master`. Setting bitcoin and miniscript `default-features = false`. Bumped esplora-client to 0.4.

### Notes to the reviewers

The `default-features = false` is to let bdk be unbiased for things like no-std.

### Changelog notice

- Set default-features = false for rust-bitcoin and rust-miniscript #882
- Update esplora client dependency to version 0.4 #884

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #894, opened 2023-03-16
Better no-std support

evanlinjin:
This replaces #893

### Description

Carrying over relevant maintenance changes from release 0.27.2 to bdk `master`.

- Use `default-features = false` for `miniscript` and `bitcoin`
- Introduce `std`  features for `bdk`, `bdk_chain` and `bdk_esplora`

### Notes to the reviewers

The `default-features = false`for `bitcoin` and `miniscript` is to let `bdk` be unbiased for things like no-std.

### Changelog notice

- Set default-features = false for rust-bitcoin and rust-miniscript #894
- Introduce `std`  features for `bdk`, `bdk_chain` and `bdk_esplora`

#### All Submissions:

### DIFF
--- a/crates/chain/Cargo.toml
+++ b/crates/chain/Cargo.toml
@@ -13,18 +13,18 @@ readme = "README.md"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-bitcoin = { version = "0.29" }
+bitcoin = { version = "0.29", default-features = false }
 serde_crate = { package = "serde", version = "1", optional = true, features = ["derive"] }
 
 # Use hashbrown as a feature flag to have HashSet and HashMap from it.
 # note version 0.13 breaks outs MSRV.
-hashbrown = { version = "0.12", optional = true,  features = ["serde"] }
-miniscript = { version = "9.0.0", optional = true  }
+hashbrown = { version = "0.12", optional = true, features = ["serde"] }
+miniscript = { version = "9.0.0", optional = true, default-features = false }
 
 [dev-dependencies]
 rand = "0.8"
 
 [features]
-default = ["std", "miniscript"]
-std = []
+default = ["std"]
+std = ["bitcoin/std"]
 serde = ["serde_crate", "bitcoin/serde" ]
--- a/crates/esplora/Cargo.toml
+++ b/crates/esplora/Cargo.toml
@@ -12,13 +12,18 @@ readme = "README.md"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-bdk_chain = { path = "../chain", version = "0.4.0", features = ["serde", "miniscript"] }
+bdk_chain = { path = "../chain", version = "0.4.0", default-features = false, features = ["serde", "miniscript"] }
 esplora-client = { version = "0.5", default-features = false }
 async-trait = { version = "0.1.66", optional = true }
 futures = { version = "0.3.26", optional = true }
 
+# use these dependencies if you need to enable their /no-std features
+bitcoin = { version = "0.29", optional = true, default-features = false }
+miniscript = { version = "9.0.0", optional = true, default-features = false }
+
 [features]
-default = ["blocking"]
+default = ["std", "async-https", "blocking"]
+std = ["bdk_chain/std"]
 async = ["async-trait", "futures", "esplora-client/async"]
 async-https = ["async", "esplora-client/async-https"]
 blocking = ["esplora-client/blocking"]
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -32,6 +32,34 @@ jobs:
       - name: Test
         run: cargo test ${{ matrix.features }}
 
+  check-no-std:
+    name: Check no_std
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+          profile: minimal
+          # target: "thumbv6m-none-eabi"
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v2.2.1
+      - name: Check bdk_chain
+        working-directory: ./crates/chain
+        # TODO "--target thumbv6m-none-eabi" should work but currently does not
+        run: cargo check --no-default-features --features bitcoin/no-std,miniscript/no-std,hashbrown
+      - name: Check bdk
+        working-directory: ./crates/bdk
+        # TODO "--target thumbv6m-none-eabi" should work but currently does not
+        run: cargo check --no-default-features --features bitcoin/no-std,miniscript/no-std,bdk_chain/hashbrown
+      - name: Check esplora
+        working-directory: ./crates/esplora
+        # TODO "--target thumbv6m-none-eabi" should work but currently does not
+        run: cargo check --no-default-features --features bitcoin/no-std,miniscript/no-std,bdk_chain/hashbrown
+
   check-wasm:
     name: Check WASM
     runs-on: ubuntu-20.04
@@ -57,10 +85,10 @@ jobs:
         uses: Swatinem/rust-cache@v2.2.1
       - name: Check bdk
         working-directory: ./crates/bdk
-        run: cargo check --target wasm32-unknown-unknown --features dev-getrandom-wasm
+        run: cargo check --target wasm32-unknown-unknown --no-default-features --features bitcoin/no-std,miniscript/no-std,bdk_chain/hashbrown,dev-getrandom-wasm
       - name: Check esplora
         working-directory: ./crates/esplora
-        run: cargo check --target wasm32-unknown-unknown --features async --no-default-features
+        run: cargo check --target wasm32-unknown-unknown --no-default-features --features bitcoin/no-std,miniscript/no-std,bdk_chain/hashbrown,async
 
   fmt:
     name: Rust fmt
--- a/crates/bdk/Cargo.toml
+++ b/crates/bdk/Cargo.toml
@@ -15,11 +15,11 @@ rust-version = "1.57"
 [dependencies]
 log = "=0.4.18"
 rand = "^0.8"
-miniscript = { version = "9", features = ["serde"] }
-bitcoin = { version = "0.29", features = ["serde", "base64", "rand"] }
+miniscript = { version = "9", features = ["serde"], default-features = false }
+bitcoin = { version = "0.29", features = ["serde", "base64", "rand"], default-features = false }
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1.0" }
-bdk_chain = { path = "../chain", version = "0.4.0", features = ["miniscript", "serde"] }
+bdk_chain = { path = "../chain", version = "0.4.0", features = ["miniscript", "serde"], default-features = false }
 
 # Optional dependencies
 hwi = { version = "0.5", optional = true, features = [ "use-miniscript"] }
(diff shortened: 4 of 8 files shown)


crates/chain/Cargo.toml
```diff
@@ -13,18 +13,19 @@ readme = "../README.md"
...
+hashbrown = { version = "0.12", optional = true, features = ["serde"] }
+miniscript = { version = "9.0.0", optional = true, default-features = false }
 
 [dev-dependencies]
 rand = "0.8"
 
 [features]
 default = ["std", "miniscript"]
-std = []
+std = ["bitcoin/std", "miniscript/std"]
```
LLFourn: :x: I don't think we need `miniscript/std` for anything. This will unnecessarily add miniscript as a dependency if someone enables the `std` feature.

crates/chain/Cargo.toml
```diff
@@ -13,18 +13,19 @@ readme = "../README.md"
...
+no-std = ["hashbrown", "bitcoin/no-std", "miniscript/no-std"]
```
LLFourn: Why does rust bitcoin and rust miniscript make no-std a feature :S

I am looking at the source of rust miniscript and rust bitcoin and it looks like the `no-std` feature doesn't do anyhing (absence of `std` makes it no_std)`

Let's not add this feature if this is the case.

notmandatory: Since there's no way to enable the `no-std` dependencies due to the absence of `std` I don't see another way. But maybe we could add these only to the `crates/bdk` project, and just document what dependencies need to be manually added for anyone using the `bdk_chain` in a no std project other than with bdk?

LLFourn: > Since there's no way to enable the `no-std` dependencies due to the absence of `std` I don't see another way. 

Ok so I see that `hashbrown` needs to be enabled to have `HashMap`. I think what we do in `bdk_chain` is the right way to solve this specific problem: Use `BTreeMap` instead unless `hashbrown` feature is enabled which makes it faster by using an actual hashbrown hashmap. If that's too offensive then just make `hashbrown` a mandatory dependency. That seems way less harmful than this feature flag.

I think adding this feature just to enable features in another crate is not a good solution:
- Crates depending on `bdk` will have to do the same thing and so the `no-std` thing spreads everywhere.
- It means `--all-features` enabling `std` and `no-std`  which just seems awful.

I think just telling people that they have to manually depend on `bitcoin` and or `miniscript` and enable their `no-std` feature manually is probably the right way to go for now. I hope we can remove `no-std` upstream.

notmandatory: Ok, I'm on-board with not adding a `no-std` feature and just documenting what needs to be manually enabled by downstream projects who are building for that environment. I think `no-std` is going to be uncommon/advance user thing anyway and certainly a lower support priority relative to the rest of the 1.0.0 changes going on. I'll add an issue to document how to use `bdk` in a no std environment and then this PR can be just the changes for `std` and bumping `esplora-client`, how does that sound?

LLFourn: @tcharding gave us some good advice:

image

we can just always enable the `no-std` feature and optionally enable std.

[edit] this solves the problem for rust-bitcoin only. I wonder if we can do the same thing for rust-miniscript too.

tcharding: Unfortunately you'll get a dependency on `hashbrown` that is unneeded in "std" if you do the same thing for `miniscript`.

tcharding: I had a quick look and I rekon it should work for miniscript also.

evanlinjin:
@notmandatory Do you want to take over this ticket? I think you are allowed to directly push to my branch.

notmandatory:
A good explanation of how to verify your crate works with `no_std` here: https://github.com/KodrAus/rust-no-std

I'm expanding the scope of this a bit tackle #920 also.

notmandatory:
I ran into a couple issues with getting `no_std` to work with `bdk_chain`. 

1. `alloc::collections::BTreeMap<K,V>` requires `K` to be `Sized`. So in `SpkTxOutIndex` I changed `K` to a `Box<Script>` 
    I don't know what downstream issues this will cause, any concerns using a `Box<Script>`? 
2.  in `bitcoin` 0.29.x `sync` needs to be changed to `alloc::sync`, this looks like it was fixed in `bitcoin` 0.30.0.
    I opened #929.

notmandatory:
BTW, I'm testing `no_std` building like this:

```shell
% cargo build -p bdk_chain --no-default-features --features bitcoin/no-std,hashbrown --target thumbv6m-none-eabi
```

tcharding:
> I ran into a couple issues with getting `no_std` to work with `bdk_chain`.
> 
> 1. `alloc::collections::BTreeMap<K,V>` requires `K` to be `Sized`. So in `SpkTxOutIndex` I changed `K` to a `Box<Script>`

In case its useful, as of `rust-bitcoin` 0.30.0 we have `ScriptBuf` which is `Sized`.

LLFourn:
@tcharding so my understanding is that `Script` is renamed to `ScriptBuf` and `Script` is now way to operate on a slice representing a script. I wasn't paying enough attention at the time but if I was I would have suggested leaving `Script` and making `ScriptSlice` for the new type to avoid breaking everything?

In any case we need to `s/Script/ScriptBuf/g` across the project.

tcharding:
Correct. We used Script/ScriptBuf to mimic types in the stdlib (Path/PathBuf), out of interest do types exist named as you suggest Foo/FooSlice?

crates/esplora/Cargo.toml
```diff
@@ -13,7 +13,7 @@ readme = "README.md"
 
 [dependencies]
 bdk_chain = { path = "../chain", version = "0.3.1", features = ["serde", "miniscript"] }
```
benthecarman: You need default features off on bdk_chain here too

benthecarman:
Not sure if this is from an issue on master or not but seems to break if you do 
```
$ cargo check --target wasm32-unknown-unknown --no-default-features --features no-std --features dev-getrandom-wasm
```

I get this error
```
    Checking bdk_chain v0.3.1 (/home/ben/projects/bdk/crates/chain)
error[E0277]: `RefCell<core::option::Option<Arc<TaprootSpendInfo>>>` cannot be shared between threads safely
   --> crates/chain/src/keychain/txout_index.rs:566:6
    |
566 | ) -> impl Iterator<Item = (u32, Script)> + Clone + Send + 'a
    |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `RefCell<core::option::Option<Arc<TaprootSpendInfo>>>` cannot be shared between threads safely
    |
    = help: within `Descriptor<DescriptorPublicKey>`, the trait `Sync` is not implemented for `RefCell<core::option::Option<Arc<TaprootSpendInfo>>>`
    = note: required because it appears within the type `Mutex<Option<Arc<TaprootSpendInfo>>>`
    = note: required because it appears within the type `Tr<DescriptorPublicKey>`
    = note: required because it appears within the type `Descriptor<DescriptorPublicKey>`
    = note: required for `&Descriptor<DescriptorPublicKey>` to implement `Send`
    = note: required because it appears within the type `Cow<'_, Descriptor<DescriptorPublicKey>>`
note: required because it's used within this closure
   --> crates/chain/src/keychain/txout_index.rs:579:13
    |
579 |             move |index| -> Result<_, miniscript::descriptor::ConversionError> {
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: required because it appears within the type `Map<TakeWhile<TakeWhile<R, [closure@txout_index.rs:575:21]>, [closure@txout_index.rs:577:21]>, [closure@txout_index.rs:579:13]>`
    = note: required because it appears within the type `TakeWhile<Map<TakeWhile<TakeWhile<R, [closure@txout_index.rs:575:21]>, [closure@txout_index.rs:577:21]>, [closure@txout_index.rs:579:13]>, fn(&Result<(u32, Script), ConversionError>) -> bool {Result::<(u32, Script), ConversionError>::is_ok}>`
    = note: required because it appears within the type `Map<TakeWhile<Map<TakeWhile<TakeWhile<R, [closure@txout_index.rs:575:21]>, [closure@txout_index.rs:577:21]>, [closure@txout_index.rs:579:13]>, fn(&Result<(u32, Script), ConversionError>) -> bool {Result::<(u32, Script), ConversionError>::is_ok}>, fn(Result<(u32, Script), ConversionError>) -> (u32, Script) {Result::<(u32, Script), ConversionError>::unwrap}>`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `bdk_chain` (lib) due to previous error
```

notmandatory:
I'm stumped too. I get the same error as above (`error[E0277]: RefCell<core::option::Option<Arc<TaprootSpendInfo>>> cannot be shared between threads safely  --> crates/chain/src/keychain/txout_index.rs:566:6`) with the build command:
```
cargo build -p bdk_chain --no-default-features --features miniscript/no-std
```
But the same build command with `std` enabled works fine:
```
cargo build -p bdk_chain --no-default-features --features miniscript/std
```
I don't understand why the `TaprootSpendInfo` type becomes `Send + Sync` when `std` is enabled but not with `no-std`.

LLFourn:
> I'm stumped too. I get the same error as above (`error[E0277]: RefCell<core::option::Option<Arc<TaprootSpendInfo>>> cannot be shared between threads safely --> crates/chain/src/keychain/txout_index.rs:566:6`) with the build command:
> 
> ```
> cargo build -p bdk_chain --no-default-features --features miniscript/no-std
> ```
> 
> But the same build command with `std` enabled works fine:
> 
> ```
> cargo build -p bdk_chain --no-default-features --features miniscript/std
> ```
> 
> I don't understand why the `TaprootSpendInfo` type becomes `Send + Sync` when `std` is enabled but not with `no-std`.

Probably they use a refcell when they don't have access to `std::sync::Mutex`. RefCell will not be `Sync`. This shouldn't be a problem and #927 might fix this (or at least it would be a first step).

notmandatory:
> 
> Probably they use a refcell when they don't have access to `std::sync::Mutex`. RefCell will not be `Sync`. This shouldn't be a problem and #927 might fix this (or at least it would be a first step).

Ok, you are 100% right, in `rust-miniscript` I found where they swap in `RefCell` for `Mutex` when `not(feature="std")`. https://github.com/rust-bitcoin/rust-miniscript/blob/4a3ba11c2fd5063be960741d557f3f7a28041e1f/src/lib.rs#L795

tcharding:
cc @sanket1729 in case he wants to chime in.

notmandatory:
I have a work-around for the `Mutex` -> `RefCell` issue, I removed the `Send` trait requirements of the `range_descriptor_spks()` function. I'll need @danielabrozzoni to let me know how badly this butchers her code 😬. 

https://github.com/bitcoindevkit/bdk/pull/894/files#diff-4f6395be13ef1a3eabfcbd5303acdd487fa20097ab838581778076dd8df29f53

LLFourn:
I think `range_descriptor_spks` simply needs to be replaced with #927 for this to work. Or at least I think from there you can make this RefCell <-> Mutex thing to only make the iterator  `!Sync` but it would let it be `Send` which is all we need here.

benthecarman:
Looks like this will eventually need to add `impl Ord for Condition` or you will get this error:

```
error[E0277]: the trait bound `Condition: Ord` is not satisfied
    --> crates/bdk/src/descriptor/policy.rs:251:40
     |
251  |         map.serialize_entry(&k_string, v)?;
     |             ---------------            ^ the trait `Ord` is not implemented for `Condition`
     |             |
     |             required by a bound introduced by this call
     |
     = help: the trait `serde::Serialize` is implemented for `BTreeSet<T>`
     = note: required for `BTreeSet<Condition>` to implement `serde::Serialize`
```

I have this in a branch here if you want: https://github.com/benthecarman/bdk/commit/d7459e94e22d7b1b7334800e89f5015a154ab298

notmandatory:
Should esplora-client be bumped to [v0.5.0](https://github.com/bitcoindevkit/rust-esplora-client/releases/tag/v0.5.0)?

notmandatory:
Per discord conversation added this to alpha.1 milestone. @evanlinjin would you rather we get this in before or after #976? I can take it over if it needs any rebasing.

notmandatory (approved):
ACK 7ab84be9c747fd73fc5d1bd0706e7027a66aadb5

notmandatory:
I've rebased this PR and removed from the commit message and PR description the part about updating the `esplora-client` version since that ended up being done in a different PR.

PR #882, opened 2023-03-04
Set `default-features = false` for rust-bitcoin and rust-miniscript

benthecarman:
Same as https://github.com/bitcoindevkit/rust-esplora-client/pull/39

Should let bdk be unbiased for things like no-std

PR #897, opened 2023-03-16
Revert: Set default-features = false for rust-bitcoin and rust-miniscript

notmandatory:
### Description

Since the `bitcoin` and `miniscript` dependencies require enabling either `std` or `no-std` the only way we can make this change is to add a corresponding `std` and `no-std` features and requiring our downstream users to enable one. But BDK doesn't really support `no-std` yet and we don't want to break downstream projects on a patch release. So unfortunately I need to revert #882. 

See also: rust-bitcoin/rust-miniscript#533

### Notes to the reviewers

I've also yanked release `0.27.2` on crates.io and will be sure to test future releases with downstream bdk-cli and bdk-ffi projects to catch any issues like this.

### Changelog notice

None

#### All Submissions:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,8 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 bdk-macros = "^0.6"
 log = "^0.4"
-miniscript = { version = "9.0", default-features = false, features = ["serde"] }
-bitcoin = { version = "0.29.2", default-features = false, features = ["serde", "base64", "rand"] }
+miniscript = { version = "9.0", features = ["serde"] }
+bitcoin = { version = "0.29.1", features = ["serde", "base64", "rand"] }
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1.0" }
 rand = "^0.8"
@@ -31,7 +31,7 @@ async-trait = { version = "0.1", optional = true }
 rocksdb = { version = "0.14", default-features = false, features = ["snappy"], optional = true }
 cc = { version = ">=1.0.64", optional = true }
 socks = { version = "0.3", optional = true }
-hwi = { version = "0.5", optional = true, features = ["use-miniscript"] }
+hwi = { version = "0.5", optional = true, features = [ "use-miniscript"] }
 
 bip39 = { version = "1.0.1", optional = true }
 bitcoinconsensus = { version = "0.19.0-3", optional = true }
@@ -104,8 +104,6 @@ test-hardware-signer = ["hardware-signer"]
 dev-getrandom-wasm = ["getrandom/js"]
 
 [dev-dependencies]
-miniscript = { version = "9.0", features = ["std"] }
-bitcoin = { version = "0.29.2", features = ["std"] }
 lazy_static = "1.4"
 env_logger = "0.7"
 electrsd = "0.22"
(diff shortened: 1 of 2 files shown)


rajarshimaitra (commented):
ACK.. One comment..

Cargo.toml
```diff
@@ -14,8 +14,8 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 bdk-macros = "^0.6"
 log = "^0.4"
-miniscript = { version = "9.0", default-features = false, features = ["serde"] }
-bitcoin = { version = "0.29.2", default-features = false, features = ["serde", "base64", "rand"] }
+miniscript = { version = "9.0", features = ["serde"] }
+bitcoin = { version = "0.29.1", features = ["serde", "base64", "rand"] }
```
rajarshimaitra, line 18: Is it intentional to downgrade `bitcoin` to `0.29.1`? I ran some tests with latest version and all seemed to work well with it. Am I missing something..

Cargo.toml
```diff
@@ -31,7 +31,7 @@ async-trait = { version = "0.1", optional = true }
 rocksdb = { version = "0.14", default-features = false, features = ["snappy"], optional = true }
 cc = { version = ">=1.0.64", optional = true }
 socks = { version = "0.3", optional = true }
-hwi = { version = "0.5", optional = true, features = ["use-miniscript"] }
+hwi = { version = "0.5", optional = true, features = [ "use-miniscript"] }
```
tnull:
```suggestion
hwi = { version = "0.5", optional = true, features = ["use-miniscript"] }
```

Cargo.toml
```diff
@@ -14,8 +14,8 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 bdk-macros = "^0.6"
 log = "^0.4"
-miniscript = { version = "9.0", default-features = false, features = ["serde"] }
```
tnull, line 17: Mh, rather than outright reverting it, could we maybe leave  `default-features = false` and set `miniscript/std` in BDK's `default` feature? This way downstream projects could opt-out of `std` by setting `default-features = false` for BDK. 

@benthecarman Would this work for you?

benthecarman: Yeah that would work

Cargo.toml
notmandatory: This was part of the original commit and since I used the `git revert` command undid the bitcoin 0.29.2 change too. To keep it simple I'd rather not make any other changes in this PR but to revert the original commit (and corresponding changelog entry). Also even if we leave this at 0.29.1 anyone who does a `cargo update` will still get the latest 0.29.2 version.

Cargo.toml
notmandatory: Since I used the git revert command it undid this spacing fix too. To keep it simple I'd rather not make any other changes in this PR but to revert the original commit. This can be fixed next time we touch the Cargo.toml file.

Cargo.toml
notmandatory: This should work but will still force any downstream projects that use `bdk` and have `default-features = false` to now have to also enable a new `std` feature. That's the kind of breaking change I want to avoid on a patch release. Ideally I want to defer all breaking changes to the 1.0 release. 

Besides Mutiny are there any other downstream projects that need to disable `std` for `bitcoin` and `miniscript`?

LLFourn:
In another PR I mentioned that we should always just enable the `no-std` feature for now.

notmandatory:
I'm closing this PR in favor of bumping the maintenance release version to 0.28.0 and add a default `std` feature. See #930 and ACK there if you agree.

PR #909, opened 2023-03-19
Make outputs of CLI examples nice

### LINKED ISSUE #905: Make outputs of CLI examples nice.
The CLI output is a bit messy and this makes it difficult to read.

I suggest putting everything into tables and to reduce the amount of data in the `list` commands and to add additional commands to see more details about a particular item.

We can use a dependency such as https://docs.rs/tabled/latest/tabled/ to achieve this.

LagginTimes:
Closes #905

Initial work done towards #905 using https://docs.rs/tabled/latest/tabled/.

The following commands are now in table form:
```
address index
address list
balance
txout list
```

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -32,6 +32,34 @@ jobs:
       - name: Test
         run: cargo test ${{ matrix.features }}
 
+  check-no-std:
+    name: Check no_std
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install Rust toolchain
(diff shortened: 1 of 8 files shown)


evanlinjin (approved):
I think this is objectively better than what we had before. Any further improvements can be done later imo.

remix7531 (commented):
Looks a lot better. However, this make it a lot harder to grep output.

bdk-table

notmandatory:
I agree with @remix75 this output format looks good but makes it hard to parse. In many of our existing tutorials with the old `bdk-cli` we use `jq` to parse the json output. If we made these tables the default output style can we also add (in a separate PR is OK) a new `--json` flag (or something like that) to format command outputs as json ?

PR #913, opened 2023-03-20
Implement utxo reservation

vladimirfomene:
### Description

This PR solves issue bitcoindevkit/bdk_wallet#166. The current design keeps a set of UTXOs that have been reserved by transactions. UTXOs are released when a transaction is canceled. I'm currently thinking it might also make sense to release UTXOs once a transaction gets broadcasted (this idea is not implemented in this PR). 

### Notes to the reviewers

With the new `ChainOracle` design, it no longer makes sense to have this as part of the keychain tracker. It might make sense to have this inside the bdk wallet. 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -334,9 +334,9 @@ impl<D> Wallet<D> {
     }
 
     /// Return the list of unspent outputs of this wallet
-    pub fn list_unspent(&self) -> Vec<LocalUtxo> {
+    pub fn list_unspent(&self, include_reserved_utxos: bool) -> Vec<LocalUtxo> {
         self.keychain_tracker
-            .full_utxos()
+            .full_utxos(include_reserved_utxos)
             .map(|(&(keychain, derivation_index), utxo)| LocalUtxo {
                 outpoint: utxo.outpoint,
                 txout: utxo.txout,
@@ -388,9 +388,9 @@ impl<D> Wallet<D> {
 
     /// Returns the utxo owned by this wallet corresponding to `outpoint` if it exists in the
     /// wallet's database.
-    pub fn get_utxo(&self, op: OutPoint) -> Option<LocalUtxo> {
+    pub fn get_utxo(&self, op: OutPoint, include_reserved_utxos: bool) -> Option<LocalUtxo> {
         self.keychain_tracker
-            .full_utxos()
+            .full_utxos(include_reserved_utxos)
             .find_map(|(&(keychain, derivation_index), txo)| {
                 if op == txo.outpoint {
                     Some(LocalUtxo {
@@ -527,11 +527,12 @@ impl<D> Wallet<D> {
 
     /// Return the balance, separated into available, trusted-pending, untrusted-pending and immature
     /// values.
-    pub fn get_balance(&self) -> Balance {
-        self.keychain_tracker.balance(|keychain| match keychain {
-            KeychainKind::External => false,
-            KeychainKind::Internal => true,
-        })
+    pub fn get_balance(&self, include_reserved_utxos: bool) -> Balance {
+        self.keychain_tracker
+            .balance(include_reserved_utxos, |keychain| match keychain {
+                KeychainKind::External => false,
+                KeychainKind::Internal => true,
+            })
     }
 
     /// Add an external signer
@@ -866,6 +867,7 @@ impl<D> Wallet<D> {
             params.drain_wallet,
             params.manually_selected_only,
             params.bumping_fee.is_some(), // we mandate confirmed transactions if we're bumping the fee
+            params.include_reserved_utxos,
             current_height.map(LockTime::to_consensus_u32),
         );
 
@@ -908,6 +910,10 @@ impl<D> Wallet<D> {
                 witness: Witness::new(),
             })
             .collect();
+
+        coin_selection.selected.iter().for_each(|u| {
+            let _ = self.keychain_tracker.mark_reserved(u.outpoint());
+        });
 
         if tx.output.is_empty() {
             // Uh oh, our transaction has no outputs.
@@ -1370,6 +1376,13 @@ impl<D> Wallet<D> {
                 txout_index.unmark_used(&keychain, index);
             }
         }
+
+        // release all the UTXOs that were reserved by this transaction
+        tx.input.iter().for_each(|input| {
+            let _ = self
+                .keychain_tracker
+                .unmark_reserved(&input.previous_output);
+        });
     }
 
     fn map_keychain(&self, keychain: KeychainKind) -> KeychainKind {
@@ -1391,8 +1404,8 @@ impl<D> Wallet<D> {
         Some(descriptor.at_derivation_index(child))
     }
 
-    fn get_available_utxos(&self) -> Vec<(LocalUtxo, usize)> {
-        self.list_unspent()
+    fn get_available_utxos(&self, include_reserved_utxos: bool) -> Vec<(LocalUtxo, usize)> {
+        self.list_unspent(include_reserved_utxos)
             .into_iter()
             .map(|utxo| {
                 let keychain = utxo.keychain;
@@ -1417,11 +1430,12 @@ impl<D> Wallet<D> {
         must_use_all_available: bool,
         manual_only: bool,
         must_only_use_confirmed_tx: bool,
+        include_reserved_utxos: bool,
         current_height: Option<u32>,
     ) -> (Vec<WeightedUtxo>, Vec<WeightedUtxo>) {
         //    must_spend <- manually selected utxos
         //    may_spend  <- all other available utxos
-        let mut may_spend = self.get_available_utxos();
+        let mut may_spend = self.get_available_utxos(include_reserved_utxos);
 
         may_spend.retain(|may_spend| {
             !manually_selected
--- a/crates/bdk/src/wallet/tx_builder.rs
+++ b/crates/bdk/src/wallet/tx_builder.rs
@@ -97,6 +97,7 @@ impl TxBuilderContext for BumpFee {}
 /// // non-chaining
 /// let (psbt2, details) = {
 ///     let mut builder = wallet.build_tx();
+///     builder.include_reserved_utxos(true);
 ///     builder.ordering(TxOrdering::Untouched);
 ///     for addr in &[addr1, addr2] {
 ///         builder.add_recipient(addr.script_pubkey(), 50_000);
@@ -150,6 +151,7 @@ pub(crate) struct TxParams {
     pub(crate) bumping_fee: Option<PreviousFee>,
     pub(crate) current_height: Option<LockTime>,
     pub(crate) allow_dust: bool,
+    pub(crate) include_reserved_utxos: bool,
 }
 
 #[derive(Clone, Copy, Debug)]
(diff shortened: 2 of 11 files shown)


LLFourn:
> With the new `ChainOracle` design, it no longer makes sense to have this as part of the keychain tracker. It might make sense to have this inside the bdk wallet.

Concept ACK. I also suggested that in response to bitcoindevkit/bdk_wallet#166. Can you change the PR to only modify bdk::Wallet?

vladimirfomene:
Thanks @LLFourn for the feedback! Let me work on that update.

crates/bdk/src/wallet/mod.rs
```diff
@@ -388,9 +388,9 @@ impl<D> Wallet<D> {
 
     /// Returns the utxo owned by this wallet corresponding to `outpoint` if it exists in the
     /// wallet's database.
-    pub fn get_utxo(&self, op: OutPoint) -> Option<LocalUtxo> {
+    pub fn get_utxo(&self, op: OutPoint, include_reserved_utxos: bool) -> Option<LocalUtxo> {
```
danielabrozzoni, line 391: I don't think we should have a `include_reserved_utxos` parameter here... If I want bdk to retrieve a specific outpoint, I usually don't care if it's reserved or not.

I would just call `full_utxos(true)` a couple of lines below and that's it

crates/bdk/src/wallet/mod.rs
```diff
@@ -527,11 +527,12 @@ impl<D> Wallet<D> {
 
     /// Return the balance, separated into available, trusted-pending, untrusted-pending and immature
     /// values.
-    pub fn get_balance(&self) -> Balance {
-        self.keychain_tracker.balance(|keychain| match keychain {
-            KeychainKind::External => false,
-            KeychainKind::Internal => true,
-        })
+    pub fn get_balance(&self, include_reserved_utxos: bool) -> Balance {
```
danielabrozzoni, line 530: Similarly to above, when users ask for a balance I suppose they always want the reserved utxos anyways

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -274,12 +276,20 @@ impl<'a, D, Cs: CoinSelectionAlgorithm, Ctx: TxBuilderContext> TxBuilder<'a, D,
     ///
     /// These have priority over the "unspendable" utxos, meaning that if a utxo is present both in
     /// the "utxos" and the "unspendable" list, it will be spent.
-    pub fn add_utxos(&mut self, outpoints: &[OutPoint]) -> Result<&mut Self, Error> {
+    pub fn add_utxos(
```
danielabrozzoni, line 279: Same here, if you want to manually add utxos you probably don't care whether they're unlocked or not. 

Unless... we might want to return a specific error (not `UnknownUtxo`, a more detailed one, like `LockedUtxo`) if you try to add a locked utxo to a txbuilder that doesn't have `include_reserved_utxos`. I'm not sure exactly what's the way forward here.

danielabrozzoni (commented):
Concept ACK, thanks for working on this! I think I'll do something similar to what you did with `include_reserved_utxos` in the new TxBuilder.
I left some comments regarding the Wallet API - I'm worried that by exposing `include_reserved_utxos` frequently we might make the API harder to use, without a real benefit. 

Also, should we have some method on `Wallet` to query and modify the reserved utxos?

crates/bdk/src/wallet/mod.rs
vladimirfomene: Makes sense.

crates/bdk/src/wallet/tx_builder.rs
vladimirfomene: I will prefer creating an error like `LockedUtxo`. Make more sense to me.

vladimirfomene:
@danielabrozzoni, what do you think the potential use cases will be for adding an API for querying and modifying the locked utxos?

danielabrozzoni:
Think about a scenario where there are multiple bdk wallets able to communicate between each other, each one of them crafting transactions and not broadcasting them immediately. In this case you don't want to have conflicting txs and you want to manually tell each wallet "also lock this specific utxo". But it's a rather exotic scenario so we don't have to implement it right now.

benthecarman:
> what do you think the potential use cases will be for adding an API for querying and modifying the locked utxos?

I could see this useful for things you queue up. I could see using this when a user queues some utxos for a coinjoin, until that coinjoin happens those utxos should be locked, so if they go to send those coins should be locked in the coin selection screen. Being able to see which utxos are locked would be important for that.

DanGould:
This would be nice in the case a payjoin fails because someone goes offline and the receiver bans their utxo (to prevent probing), the signing algorithm could automatically pick a utxo outside of the reserved set.

DanGould:
I'm curious what the status of this PR is. Tracking and expiring intermediate but unconfirmed payjoin states (pending original psbt vs augmented payjoin psbt) requires locking. Otherwise subsequent sign attempts spend outputs of pending payjoin state that may never end up confirmed on chain.

@vladimirfomene are you still invested in seeing this PR completed?
@LLFourn how much bandwidth does the BDK team have to review this PR considering the 1.0 stabilization?

Would either of you like a hand or consider my contributions to get this reservation issue fixed?

LLFourn:
@DanGould reconsidering the problem here. Is there something wrong with just using: [`add_unspendable`](https://docs.rs/bdk/1.0.0-alpha.8/bdk/wallet/tx_builder/struct.TxBuilder.html#method.add_unspendable) when you don't want to spend a txout because you know at the application level which transactions you intend on broadcasting and which you don't.

Also you could consider just adding the transaction to the wallet with `insert_tx` which will also remove the txs from being spendable while also adding any utxos on that new tx to the pool but I'm guessing you don't want to do that here.

DanGould:
Edit: I believe I've been able to fix this problem by having payjoin only contribute confirmed inputs. This is a slight degradation vs being able spend unconfirmed-but-seen-in-mempool transactions but it works without reservation for now.

Please bear with me as I outline in detail what I'm trying to do and where I've run into the snag. I'm trying to watch unconfirmed, pending transactions that one of the two payjoin parties have signed but which may never be broadcast, and are out of the one party's control to broadcast. I am running BDK 1.0.0-alpha.5.

## Payjoin Sender

The payjoin sender first signs an Original PSBT that's a typical transaction sending funds to the receiver. It wants to add this transaction to its wallet because, even though it never plans on broadcasting this TX, if for some reason the payjoin session times out, the receiver may extract and broadcast this original fallback transaction, making the sender's original tx change spendable. If the sender receives a payjoin PSBT with a signed receiver input, it can sign and broadcast the payjoin and remove the Original tx from its wallet because those funds have already been spent in the payjoin.

## Payjoin Receiver

The payjoin receiver augments an incoming original PSBT with its input to create a payjoin PSBT missing the sender's signature, since the sender's original PSBT sighash_all signature becomes invalid when the transaction is augmented. The receiver adds this payjoin transaction to its wallet, sends it to the sender to receive a signature, and waits to see it broadcast. If that transaction never gets signed, the receiver may choose to broadcast the original PSBT tx if it's an automated payment processor or otherwise let the payjoin session expire. During the time this payjoin PSBT in the hands of the sender in the wild and not considered expired nor seen in mempool, the receiver shouldn't try to create more transactions spending the same input as the payjoin PSBT nor spend change outputs from the payjoin PSBT since they might never be confirmed. I would understand these to be 'reserved' utxos.

--- 

I wasn't considering `add_unspendable`, ~but I suppose that could work if I can add pending wallet UTXOs as unspendable and later remove them once a conflicting broadcast is detected. Will this still produce a pending transaction when I call `wallet.transactions()`?.~ because that's a TxBuilder method and I'm looking for a way to track pending transactions that may later become invalid. Building a transaction is not a problem. The external UTXO is added to a psbt manually, not via TxBuilder.

<details>
<summary>`list_transactions` implementation</summary>

  ```rs
  pub fn list_transactions(
      &self,
      include_raw: bool,
  ) -> Result<Vec<TransactionDetails>, MutinyError> {
      if let Ok(wallet) = self.wallet.try_read() {
          let txs = wallet
              .transactions()
              .filter_map(|tx| {
                  // skip txs that were not relevant to our bdk wallet
                  if wallet.spk_index().is_tx_relevant(tx.tx_node.tx) {
                      let (sent, received) = wallet.spk_index().sent_and_received(tx.tx_node.tx);
  
                      let transaction = if include_raw {
                          Some(tx.tx_node.tx.clone())
                      } else {
                          None
                      };
  
                      // todo bdk is making an easy function for this
                      // calculate fee if possible
                      let inputs = tx
                          .tx_node
                          .tx
                          .input
                          .iter()
                          .map(|txin| {
                              wallet
                                  .spk_index()
                                  .txout(txin.previous_output)
                                  .map(|(_, _, txout)| txout.value)
                          })
                          .sum::<Option<u64>>();
                      let outputs = tx.tx_node.tx.output.iter().map(|txout| txout.value).sum();
                      let fee = inputs.map(|inputs| inputs.saturating_sub(outputs));
  
                      Some(TransactionDetails {
                          transaction,
                          txid: tx.tx_node.txid,
                          received,
                          sent,
                          fee,
                          confirmation_time: tx.chain_position.cloned().into(),
                          labels: vec![],
                      })
                  } else {
                      None
                  }
              })
              .collect();
          return Ok(txs);
      }
      log_error!(
          self.logger,
          "Could not get wallet lock to list transactions"
      );
      Err(MutinyError::WalletOperationFailed)
  }
  ```
</details>

[My current attempt (one commit in a PR)](https://github.com/MutinyWallet/mutiny-node/pull/820/commits/7880f9de97d8423a2943de3beee88807022bbf42) has the sender `insert_tx` an Original transaction it sends to the receiver. The sender then calls `cancel_tx` on the original tx if it receives, signs, and broadcasts an updated payjoin.

Similarly, the receiver also calls `insert_tx` on the payjoin before it sends its payjoin psbt it to the sender to await the payjoin broadcast from the sender.

However, I'm running into a problem where subsequent attempts at receiving payjoin try to spend the output of the prior payjoin tx in a new payjoin. That prior payjoin change output has never been broadcast, so the sender rejects it because it does not exist from the sender perspective. My understanding is that if the payjoin tx input and output was marked as unspendable or "reserved," then it would NOT appear as a `LocalOutput` from `list_unspent` and chosen for spending in subsequent transactions.

 ([here's the original referenced commit](https://github.com/DanGould/mutiny-node/commit/7880f9de97d8423a2943de3beee88807022bbf42) in a branch separate from the PR in case that PR is updated)

LLFourn:
@DanGould thanks for the context.

1. Inserting transactions before they are broadcast: I actually thought this would be fine but it looks like we've got bug that means that if you insert a transaction before broadcasting it, it will still show up as "unconfirmed" and therefore utxos from it will show up in your utxo list (and utxos it spends will be gone). Note `cancel_tx` won't help here. I made an issue here: https://github.com/bitcoindevkit/bdk/issues/1396. So for now you should never insert transactions before they are broadcast.
2. How to "reserve" utxos: Even if you are not using `TxBuilder` you can still manually `.filter` the list of utxos right? I'm not sure what value we'd provide by having an API for doing this more behind the scenes. It seems to me to be better if the application is explicitly in control of enforcing this.

The feature this PR is aimed at providing was more about stopping multiple threads from creating transactions that spend the same utxos. The plan wasn't to persist this list of reserved utxos (just have them in memory). It was just there to "lock" the utxo in the short time between a transaction being created and it being confirmed to broadcast. This might be useful to you too though. I think the thing to keep in mind is that we don't want to persist these lists of reserved utxos if you want that you'll have to do it at the application level.

I'd hesitate to encourage you to take on this PR as I there could be quite some bikesheading around what to do with the balance, when to unreserve utxos etc. Only take it on if you have significant time to dedicate to figuring out the right answers to these (and other) questions.

evanlinjin:
Stale.

PR #924, opened 2023-03-23
Introduce new `bdk_coin_select` implementation

LLFourn:
This is porting over https://github.com/LLFourn/bdk_core_staging/pull/46

I made it compile on 1.48.0 and added a script that we can use in CI to check that all the crates that should be able to compile on 1.48.0 can.

TODO:

- finish documentation
- update examples to use this

I've had [a request](https://github.com/wizardsardine/liana/issues/51#issuecomment-1457911498) to do a release of the separately before it is used in `bdk` itself so that other projects can try it out.

cc @darosior


notmandatory:
The script approach looks good to me. I tried my sub-directory idea (`cargo +1.48.0 build --manifest-path msrv-1.48.0/Cargo.toml`) and it doesn't work.

nursery/coin_select/src/coin_selector.rs
```diff
@@ -1,173 +1,171 @@
...
+/// [`CoinSelector`] is responsible for selecting and deselecting from a set of canididates.
+///
+/// You can do this manually by calling methods like [`select`] or automatically with methods like [`branch_and_bound`].
+///
+/// [`select`]: CoinSelector::select
+/// [`branch_and_bound`]: CoinSelector::branch_and_bound
+#[derive(Debug, Clone)]
+pub struct CoinSelector<'a> {
+    base_weight: u32,
+    candidates: &'a [Candidate],
```
evanlinjin, line 16: I think it will be much more helpful if we had:
```rust
pub struct CoinSelector<'c, C> {
    candidates: &'c [C],
    /* ... */
}

impl<'c, C: AsRef<Candidate>> CoinSelector<'c, C> {}
```

This way `C` can contain everything required to complete the rest of tx building.

You can even do:

```rust
    impl AsRef<Candidate> for Candidate {
        fn as_ref(&self) -> &Candidate {
            self
        }
    }
```

LLFourn: I think to do this properly we'd have to make `Candidate` a trait. I'd leave it like this for now until we see how tx building turns out and make it more flexible where we find pain points.

on the other hand there's no real harm to doing what you suggested now so maybe I can give it a shot.

nursery/coin_select/src/lib.rs
```diff
@@ -1,33 +1,61 @@
...
+pub use feerate::*;
+pub mod change_policy;
+
+/// Txin "base" fields include `outpoint` (32+4) and `nSequence` (4) and 1 byte for the scriptSig
+/// length.
+pub const TXIN_BASE_WEIGHT: u32 = (32 + 4 + 4 + 1) * 4;
+
+/// The weight of a TXOUT without the `scriptPubkey` (and script pubkey length field).
+/// Just the weight of the value field.
+pub const TXOUT_BASE_WEIGHT: u32 = 4 * core::mem::size_of::<u64>() as u32; // just the value
```
darosior: nit: Technically it's an `i64`, but the size is the same anyways.

```suggestion
pub const TXOUT_BASE_WEIGHT: u32 = 4 * core::mem::size_of::<i64>() as u32; // just the value
```

nursery/coin_select/src/coin_selector.rs
```diff
@@ -176,440 +174,424 @@ impl<'a> CoinSelector<'a> {
...
-            ExcessStrategyKind::ToRecipient => core::write!(f, "to_recipient"),
-            ExcessStrategyKind::ToDrain => core::write!(f, "to_drain"),
+    /// Is not the "none" drain
+    pub fn is_some(&self) -> bool {
+        !self.is_none()
+    }
+
+    pub fn new_tr_keyspend() -> Self {
+        Self {
+            weight: TXOUT_BASE_WEIGHT + TR_SPK_WEIGHT,
```
darosior, line 551: You are missing the scriptPubKey length (and therefore 4 weight units)?

darosior (commented):
Impressive work, thanks for making it a separate crate.

We started poking at using this in Liana (see https://github.com/wizardsardine/liana/pull/560), so i'm curious what's the state of this PR?

nursery/coin_select/src/lib.rs
LLFourn: I actually didn't know that bitcoin core treated these at `i64` internally. I think from a consensus perspective either is correct? i.e. if we changed the definition to be `u64` the protocol doesn't change at all.

nursery/coin_select/src/coin_selector.rs
LLFourn: Yep well spotted. I added a 1 byte spk length to `TXOUT_BASE_WEIGHT`. All this weight stuff was added on a whim while I was trying to use this and is not very well tested. I'm not sure if it's worth having since you should probably get the numbers from rust bitcoin and rust miniscript which do a better job of telling you this stuff now.

LLFourn:
Hey @darosior. Thanks for taking a stab at using this that was really helpful. I think the status is waiting for @evanlinjin to help me finish it. Here's a list of TODOs:

1. Add minimize fee metric (it's a better default that waste see https://github.com/wizardsardine/liana/pull/560#discussion_r1236237579) 
2. Allow easy weighted combination of different metrics (see same link above)
3. Document everything including nice README.md exmaples
4. Make DWIM branch and bound helper function (https://github.com/wizardsardine/liana/pull/560#discussion_r1236232805)
5. Make `DrainWeight` (https://github.com/wizardsardine/liana/pull/560#discussion_r1236246084)
6. Add `CoinSelector::fund_outputs` (see https://github.com/wizardsardine/liana/pull/560#discussion_r1236251001) -- another point is that in general we need to know the number of outputs to get the varint for the outputs correct which I think we are cheating on atm.
7. Test cases against real world transactions to see if coin selections have precisely the real world feerate that was expected. (EDIT: I've just done a few of these against the feerate shown mempool.space).
8. Rewrite examples to use this crate.
9. Remove all `TODO` in the new crate and make issues for further work.

I wonder if @evanlinjin can put this as priority after #1002.

nursery/coin_select/src/coin_selector.rs
```diff
@@ -1,173 +1,171 @@
...
-        WeightedValue {
-            value,
-            weight,
-            input_count: 1,
-            is_segwit,
+impl Default for Target {
+    fn default() -> Self {
+        Self {
+            feerate: FeeRate::default_min_relay_fee(),
+            min_fee: 0, // TODO figure out what the actual network rule is for this
```
darosior, line 37: There is no explicit standardness limit for absolute fee (aside from replacement rules). It's implicitly bounded by the minimum feerate rule since transactions can't be empty.

nursery/coin_select/src/coin_selector.rs
```diff
@@ -1,173 +1,171 @@
...
-/// single UTXO, or a group of UTXOs that should be spent together.
+/// A target value to select for along with feerate constraints.
 #[derive(Debug, Clone, Copy)]
-pub struct WeightedValue {
-    /// Total value of the UTXO(s) that this [`WeightedValue`] represents.
+pub struct Target {
+    /// The minimum feerate that the selection must have
+    pub feerate: FeeRate,
+    /// The minimum fee the selection must have
+    pub min_fee: u64,
```
darosior, line 28: Since this only needs be specified for RBF, how about making it `Optional`?

nursery/coin_select/src/coin_selector.rs
```diff
@@ -176,440 +174,426 @@ impl<'a> CoinSelector<'a> {
...
+            .selected()
+            .find(|(_, wv)| wv.is_segwit)
+            .map(|_| 2)
+            .unwrap_or(0);
+        let vin_count_varint_extra_weight = {
+            let input_count = self.selected().map(|(_, wv)| wv.input_count).sum::<usize>();
+            (varint_size(input_count) - 1) * 4
+        };
+
+        self.selected_weight() + witness_header_extra_weight + vin_count_varint_extra_weight
```
darosior: For transaction that have at least one Segwit input, you also need to consider the added witness stack size for all inputs that are *not* spending from a Segwit Script. This is always 1 per legacy input since for them this stack is always empty.

More information about this here: https://github.com/bitcoin/bitcoin/pull/26567/files#diff-6e06b309cd494ef5da4e78aa0929a980767edd12342137f268b9219167064d13R62-R69.

LLFourn: Thanks. I fixed this and added some tests 4e0c907b9d297b7df9e95495dc6ca26d4ecd490e

Interestingly doing this properly broke some prop tests. The prop tester would find edge cases where the branch and bound algorithm found what looked like a good solution with lots of legacy inputs but then it had to add one segwit input and made the weight blow up for the existing inputs. This could probably be fixed but I don't think it's worth the effort and code complexity just to have slightly better selections when you mix legacy and segwit for waste metric etc.

nursery/coin_select/src/coin_selector.rs
LLFourn: Yep. I'll leave removing all these `TODO`s around the code for @evanlinjin who I think created most of them. Things that need addressing should that when this is merged it's TODO free.

nursery/coin_select/src/coin_selector.rs
LLFourn: This would mean `Some(0)` and `None` have the same semantics which I think outweighs the better intuition provided by having it as an `Option`. Code docs should for sure explain that the only time you need to use this is to comply with RBF rules.

darosior:
So https://github.com/bitcoindevkit/bdk/pull/1002 was replaced by https://github.com/bitcoindevkit/bdk/pull/1034 which was merged. @evanlinjin do you still have this as a priority?

evanlinjin:
@darosior thanks for asking. I'm literally planning to start work on this tomorrow. This will be a priority. I should have a rough timeline of when I think I can get this completed by the end of the week.

Also note that I also need to look into a couple of bugs as well. Will keep you posted.

darosior:
That's great news! Thanks, please do.

notmandatory:
This is replaced by #1072.

Issue #904, opened 2023-03-17
Spks iterators should have their own iterator type


evanlinjin:
Something like this because with the current method using `range_descriptor_spks` the returned iterator will not be able quickly call `skip` or `nth` on the iterator. It will do all the intermediate derivations.

```rust
use bitcoin::{
    secp256k1::{Secp256k1, VerifyOnly},
    Script,
};
use miniscript::{Descriptor, DescriptorPublicKey};

/// An iterator over a descriptor's script pubkeys.
///
// TODO: put this into miniscript
#[derive(Clone, Debug)]
pub struct SpkIter {
    descriptor: Descriptor<DescriptorPublicKey>,
    index: usize,
    secp: Secp256k1<VerifyOnly>,
    end: usize,
}

impl SpkIter {
    /// Creates a new script pubkey iterator starting at 0 from a descriptor
    pub fn new(descriptor: Descriptor<DescriptorPublicKey>) -> Self {
        let secp = Secp256k1::verification_only();
        let end = if descriptor.has_wildcard() {
            // Because we only iterate over non-hardened indexes there are 2^31 values
            (1 << 31) - 1
        } else {
            0
        };

        Self {
            descriptor,
            index: 0,
            secp,
            end,
        }
    }
}

impl Iterator for SpkIter {
    type Item = (u32, Script);

    fn nth(&mut self, n: usize) -> Option<Self::Item> {
        self.index = self.index.saturating_add(n);
        self.next()
    }

    fn next(&mut self) -> Option<Self::Item> {
        let index = self.index;
        if index > self.end {
            return None;
        }

        let script = self
            .descriptor
            .at_derivation_index(self.index as u32)
            .derived_descriptor(&self.secp)
            .expect("the descritpor cannot need hardened derivation")
            .script_pubkey();

        self.index += 1;

        Some((index as u32, script))
    }
}
```

Original ticket: https://github.com/LLFourn/bdk_core_staging/issues/190

evanlinjin:
@LagginTimes is currently working on this. I can't seem to assign him to the GitHub ticket for some reason.

notmandatory:
> @LagginTimes is currently working on this. I can't seem to assign him to the GitHub ticket for some reason.

It's something with GitHub where we can only assign issues to people who are part of the organization. I think your comment here is sufficient.

evanlinjin:
#927 completed this

PR #927, opened 2023-03-24
Custom spk iterator

LagginTimes:
PR for #904 

**What is done:**
- Added `SpkIterator` with `new` and `new_with_range` constructors, along with its own `nth()` and `next()` implementations
- Replaced all instances of `range_descriptor_spks` with the new `SpkIterator::new_with_range` constructor
- Created wildcard and non-wildcard unit tests for `nth()` and `next()` functions for `SpkIterator`

**What is in-progress:**
- ~@evanlinjin: Discussing the necessity of `Cow<>` in this implementation~
- ~@LagginTimes: Documentation for the new functions~

### DIFF
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -2,18 +2,16 @@ use crate::{
     collections::*,
     indexed_tx_graph::{Indexer, OwnedIndexer},
     miniscript::{Descriptor, DescriptorPublicKey},
-    ForEachTxOut, SpkTxOutIndex,
+    spk_iter::BIP32_MAX_INDEX,
+    ForEachTxOut, SpkIterator, SpkTxOutIndex,
 };
-use alloc::{borrow::Cow, vec::Vec};
-use bitcoin::{secp256k1::Secp256k1, OutPoint, Script, TxOut};
+use alloc::vec::Vec;
+use bitcoin::{OutPoint, Script, TxOut};
 use core::{fmt::Debug, ops::Deref};
 
 use crate::Append;
 
 use super::DerivationAdditions;
-
-/// Maximum [BIP32](https://bips.xyz/32) derivation index.
-pub const BIP32_MAX_INDEX: u32 = (1 << 31) - 1;
 
 /// A convenient wrapper around [`SpkTxOutIndex`] that relates script pubkeys to miniscript public
 /// [`Descriptor`]s.
@@ -243,10 +241,9 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         let next_reveal_index = self.last_revealed.get(keychain).map_or(0, |v| *v + 1);
         let lookahead = self.lookahead.get(keychain).map_or(0, |v| *v);
 
-        for (new_index, new_spk) in range_descriptor_spks(
-            Cow::Borrowed(descriptor),
-            next_store_index..next_reveal_index + lookahead,
-        ) {
+        for (new_index, new_spk) in
+            SpkIterator::new_with_range(descriptor, next_store_index..next_reveal_index + lookahead)
+        {
             let _inserted = self
                 .inner
                 .insert_spk((keychain.clone(), new_index), new_spk);
@@ -266,13 +263,13 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// derivable script pubkeys.
     pub fn spks_of_all_keychains(
         &self,
-    ) -> BTreeMap<K, impl Iterator<Item = (u32, Script)> + Clone> {
+    ) -> BTreeMap<K, SpkIterator<Descriptor<DescriptorPublicKey>>> {
         self.keychains
             .iter()
             .map(|(keychain, descriptor)| {
                 (
                     keychain.clone(),
-                    range_descriptor_spks(Cow::Owned(descriptor.clone()), 0..),
+                    SpkIterator::new_with_range(descriptor.clone(), 0..),
                 )
             })
             .collect()
@@ -284,13 +281,13 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// # Panics
     ///
     /// This will panic if the `keychain` does not exist.
-    pub fn spks_of_keychain(&self, keychain: &K) -> impl Iterator<Item = (u32, Script)> + Clone {
+    pub fn spks_of_keychain(&self, keychain: &K) -> SpkIterator<Descriptor<DescriptorPublicKey>> {
         let descriptor = self
             .keychains
             .get(keychain)
             .expect("keychain must exist")
             .clone();
-        range_descriptor_spks(Cow::Owned(descriptor), 0..)
+        SpkIterator::new_with_range(descriptor, 0..)
     }
 
     /// Convenience method to get [`revealed_spks_of_keychain`] of all keychains.
@@ -370,7 +367,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         &mut self,
         keychains: &BTreeMap<K, u32>,
     ) -> (
-        BTreeMap<K, impl Iterator<Item = (u32, Script)>>,
+        BTreeMap<K, SpkIterator<Descriptor<DescriptorPublicKey>>>,
         DerivationAdditions<K>,
     ) {
         let mut additions = DerivationAdditions::default();
@@ -380,7 +377,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
             let (new_spks, new_additions) = self.reveal_to_target(keychain, index);
             if !new_additions.is_empty() {
                 spks.insert(keychain.clone(), new_spks);
-                additions.append(new_additions);
+                additions.append(new_additions.clone());
             }
         }
 
@@ -405,7 +402,10 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         &mut self,
         keychain: &K,
         target_index: u32,
-    ) -> (impl Iterator<Item = (u32, Script)>, DerivationAdditions<K>) {
+    ) -> (
+        SpkIterator<Descriptor<DescriptorPublicKey>>,
+        DerivationAdditions<K>,
+    ) {
         let descriptor = self.keychains.get(keychain).expect("keychain must exist");
         let has_wildcard = descriptor.has_wildcard();
 
@@ -430,7 +430,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
 
         // we range over indexes that are not stored
         let range = next_reveal_index + lookahead..=target_index + lookahead;
-        for (new_index, new_spk) in range_descriptor_spks(Cow::Borrowed(descriptor), range) {
+        for (new_index, new_spk) in SpkIterator::new_with_range(descriptor, range) {
             let _inserted = self
                 .inner
                 .insert_spk((keychain.clone(), new_index), new_spk);
@@ -447,16 +447,13 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
                 let _old_index = self.last_revealed.insert(keychain.clone(), index);
                 debug_assert!(_old_index < Some(index));
                 (
-                    range_descriptor_spks(
-                        Cow::Owned(descriptor.clone()),
-                        next_reveal_index..index + 1,
-                    ),
+                    SpkIterator::new_with_range(descriptor.clone(), next_reveal_index..index + 1),
                     DerivationAdditions(core::iter::once((keychain.clone(), index)).collect()),
                 )
             }
             None => (
-                range_descriptor_spks(
-                    Cow::Owned(descriptor.clone()),
+                SpkIterator::new_with_range(
+                    descriptor.clone(),
                     next_reveal_index..next_reveal_index,
                 ),
                 DerivationAdditions::default(),
@@ -586,34 +583,4 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     pub fn apply_additions(&mut self, additions: DerivationAdditions<K>) {
         let _ = self.reveal_to_target_multi(&additions.0);
     }
-}
-
-fn range_descriptor_spks<'a, R>(
-    descriptor: Cow<'a, Descriptor<DescriptorPublicKey>>,
-    range: R,
-) -> impl Iterator<Item = (u32, Script)> + Clone + Send + 'a
-where
-    R: Iterator<Item = u32> + Clone + Send + 'a,
-{
-    let secp = Secp256k1::verification_only();
-    let has_wildcard = descriptor.has_wildcard();
-    range
-        .into_iter()
-        // non-wildcard descriptors can only have one derivation index (0)
-        .take_while(move |&index| has_wildcard || index == 0)
-        // we can only iterate over non-hardened indices
-        .take_while(|&index| index <= BIP32_MAX_INDEX)
-        .map(
-            move |index| -> Result<_, miniscript::descriptor::ConversionError> {
-                Ok((
-                    index,
-                    descriptor
-                        .at_derivation_index(index)
-                        .derived_descriptor(&secp)?
-                        .script_pubkey(),
-                ))
-            },
-        )
-        .take_while(Result::is_ok)
-        .map(Result::unwrap)
 }
--- a/crates/chain/src/spk_iter.rs
+++ b/crates/chain/src/spk_iter.rs
@@ -1,0 +1,215 @@
+use crate::{
+    bitcoin::{secp256k1::Secp256k1, Script},
+    miniscript::{Descriptor, DescriptorPublicKey},
+};
+use core::{borrow::Borrow, ops::Bound, ops::RangeBounds};
+
+/// Maximum [BIP32](https://bips.xyz/32) derivation index.
+pub const BIP32_MAX_INDEX: u32 = (1 << 31) - 1;
+
+/// An iterator for derived script pubkeys.
+///
+/// [`SpkIterator`] is an implementation of the [`Iterator`] trait which possesses its own `next()`
+/// and `nth()` functions, both of which circumvent the unnecessary intermediate derivations required
+/// when using their default implementations.
+///
+/// ## Examples
+///
+/// ```
+/// use bdk_chain::SpkIterator;
+/// # use miniscript::{Descriptor, DescriptorPublicKey};
+/// # use bitcoin::{secp256k1::Secp256k1};
+/// # use std::str::FromStr;
+/// # let secp = bitcoin::secp256k1::Secp256k1::signing_only();
+/// # let (descriptor, _) = Descriptor::<DescriptorPublicKey>::parse_descriptor(&secp, "wpkh([73c5da0a/86'/0'/0']xprv9xgqHN7yz9MwCkxsBPN5qetuNdQSUttZNKw1dcYTV4mkaAFiBVGQziHs3NRSWMkCzvgjEe3n9xV8oYywvM8at9yRqyaZVz6TYYhX98VjsUk/1/0)").unwrap();
+/// # let external_spk_0 = descriptor.at_derivation_index(0).script_pubkey();
+/// # let external_spk_3 = descriptor.at_derivation_index(3).script_pubkey();
+/// # let external_spk_4 = descriptor.at_derivation_index(4).script_pubkey();
+///
+/// // Creates a new script pubkey iterator starting at 0 from a descriptor.
+/// let mut spk_iter = SpkIterator::new(&descriptor);
+/// assert_eq!(spk_iter.next(), Some((0, external_spk_0)));
+/// assert_eq!(spk_iter.next(), None);
+/// ```
+#[derive(Clone)]
+pub struct SpkIterator<D> {
+    next_index: u32,
+    end: u32,
+    descriptor: D,
+    secp: Secp256k1<bitcoin::secp256k1::VerifyOnly>,
+}
+
+impl<D> SpkIterator<D>
+where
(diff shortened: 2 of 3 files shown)


crates/chain/src/keychain/txout_index.rs
```diff
@@ -588,3 +588,63 @@ where
...
+
+        self.next_index += 1;
+
+        Some((self.next_index, script))
+    }
+
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        //let n = self.next_index + n as u32;
+        //self.descriptor.at_derivation_index(n);
+        self.next_index = self.next_index.saturating_add(n as u32);
```
evanlinjin: I think `checking_add` is more accurate.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -588,3 +588,63 @@ where
         .take_while(Result::is_ok)
         .map(Result::unwrap)
 }
+
+pub struct SpkIterator<'a> {
+    next_index: u32,
+    end: u32,
+    descriptor: Cow<'a, Descriptor<DescriptorPublicKey>>,
+    secp: Secp256k1<bitcoin::secp256k1::VerifyOnly>,
+}
```
evanlinjin: Let's start using this in the actual codebase! Create one (or more) constructors for this and replace the calls to `range_descriptor_spks()`. The other way is to make `range_descriptor_spks()` a constructor of `SpkIterator`.

crates/chain/tests/test_keychain_txout_index.rs
```diff
@@ -367,3 +368,19 @@ fn test_non_wildcard_derivations() {
     assert_eq!(revealed_spks.count(), 0);
     assert!(revealed_additions.is_empty());
 }
+
+#[test]
+fn test_spkiterator_wildcard() {
```
evanlinjin: Another test for non-wildcard will be nice!

crates/chain/src/keychain/txout_index.rs
evanlinjin: Okay, I see you have a nice constructor below `SpkIterator::new`. We need another one that allows us to set a range.

benthecarman:
@LagginTimes any updates here, this is blocking #894

crates/chain/src/keychain/txout_index.rs
```diff
@@ -588,3 +588,63 @@ where
...
+
+impl<'a> SpkIterator<'a> {
+    /// Creates a new script pubkey iterator starting at 0 from a descriptor
+    pub fn new(descriptor: Cow<'a, Descriptor<DescriptorPublicKey>>) -> Self {
+        let secp = Secp256k1::verification_only();
+        let end = if descriptor.has_wildcard() {
+            // Because we only iterate over non-hardened indexes there are 2^31 values
+            (2 ^ 31) - 1
```
notmandatory: You should use the `BIP32_MAX_INDEX` const here. But if you did want to do it with a x^y it'd have to be `u32::pow(2, 31) -1`.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -588,3 +588,63 @@ where
...
+pub struct SpkIterator<'a> {
+    next_index: u32,
+    end: u32,
+    descriptor: Cow<'a, Descriptor<DescriptorPublicKey>>,
+    secp: Secp256k1<bitcoin::secp256k1::VerifyOnly>,
+}
+
+impl<'a> SpkIterator<'a> {
+    /// Creates a new script pubkey iterator starting at 0 from a descriptor
+    pub fn new(descriptor: Cow<'a, Descriptor<DescriptorPublicKey>>) -> Self {
```
notmandatory: Why does the `descriptor` param need to be a `Cow<>` ? I've never used a `Cow<>` but from the docs it looks like it's needed if `Descriptor<DescriptorPublicKey>>` is mutable and I don't see here where it needs to be. This way should work:
```suggest
pub fn new(descriptor: &'a Descriptor<DescriptorPublicKey>) -> Self {
```

crates/chain/src/keychain/txout_index.rs
```diff
@@ -588,3 +588,63 @@ where
         .take_while(Result::is_ok)
         .map(Result::unwrap)
 }
+
+pub struct SpkIterator<'a> {
+    next_index: u32,
+    end: u32,
```
notmandatory: Since `end` is always 0 or `BIP32_MAX_INDEX` I don't think you need the `end` field. You only need to check in the logic (as you're doing) if `descriptor.has_wildcard()`.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -588,3 +588,63 @@ where
...
+impl<'a> Iterator for SpkIterator<'a> {
+    type Item = (u32, Script);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // for non-wildcard descriptors:
+        // * we expect the first element to be Some((0, spk)), then will return None after.
+        // for wildcard descriptors:
+        // * we expect it to keep iterating until exhausted.
+        let has_wildcard = self.descriptor.has_wildcard();
+        if self.next_index > self.end || (has_wildcard && self.next_index > BIP32_MAX_INDEX) {
```
notmandatory:
```suggestion
        if (!has_wildcard && self.next_index > 0) || (has_wildcard && self.next_index > BIP32_MAX_INDEX) {
```

crates/chain/src/keychain/txout_index.rs
notmandatory: In this case `saturating_add` should be OK since the `.next()` call checks that the `next_index` value isn't greater than `BIP32_MAX_INDEX`.

crates/chain/src/keychain/txout_index.rs
notmandatory: You can disregard my comment above. I think we need the `Cow<>` so that  `SpkIterator` will be Send+Sync, but I still need to verify it solves my issue in #894.

crates/chain/src/keychain/txout_index.rs
evanlinjin: @notmandatory I don't think this is correct, I see where you are coming from, but sometimes we want to bound the iterator to a specific range. Having the `end` index here would be helpful for that, unless if you have a better idea in mind?

notmandatory: Ah ok, I was just looking at the `new` constructor and didn't see any way to change the `end` field there. Since the `SpkIterator` fields are private they'll need to be made `pub` or `pub(crate)` or add some way to set `end` right?

evanlinjin: Agreed!

crates/chain/src/keychain/txout_index.rs
```diff
@@ -559,32 +559,75 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-            },
-        )
-        .take_while(Result::is_ok)
-        .map(Result::unwrap)
+    secp: Secp256k1<bitcoin::secp256k1::VerifyOnly>,
+}
+
+impl<'a> SpkIterator<'a> {
+    // Creates a new script pubkey iterator starting at 0 from a descriptor
+    pub fn new(descriptor: Cow<'a, Descriptor<DescriptorPublicKey>>) -> Self {
```
evanlinjin: It'll be good to have another constructor where you can specify start index and an optional end index.

crates/chain/tests/test_keychain_txout_index.rs
evanlinjin: Tests to make sure `end` is enforced correctly.

evanlinjin (changes requested):

crates/chain/src/keychain/txout_index.rs
```diff
@@ -559,32 +559,75 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+
+        self.next_index += 1;
+
+        Some(output)
+    }
+
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        //let n = self.next_index + n as u32;
+        //self.descriptor.at_derivation_index(n);
+        self.next_index = self.next_index.saturating_add(n as u32);
```
evanlinjin: `n` (which is `usize`) may be >32bit (depending on the architecture). We should use a fallible method `TryInto::try_into` and use `u32::MAX` for the error case.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -559,32 +559,75 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     }
 }
 
-fn range_descriptor_spks<'a, R>(
+fn range_descriptor_spks(
+    descriptor: Cow<'_, Descriptor<DescriptorPublicKey>>,
+    range: Range<u32>,
```
evanlinjin: Instead of taking in a concrete type `Range` (I know I recommended this, I'm sorry), take in a generic `RangeBounds`.

https://doc.rust-lang.org/core/ops/trait.RangeBounds.html

This way you can still use `a..=b`

evanlinjin (changes requested):

crates/chain/src/keychain/txout_index.rs
evanlinjin: Good point, however I would still prefer `checking_add` as I find it "cleaner" and easier to reason about. This has become a nit though.

crates/chain/src/keychain/txout_index.rs
evanlinjin: I used `Cow` because the iterator is used for both internal logic (where we don't want to do an unnecessary clone of the descriptor), and also returned (where we do want to clone the descriptor as we don't want to hold a reference to our structures).

crates/chain/src/keychain/txout_index.rs
evanlinjin: We have a second constructor `new_with_range` for this now!

crates/chain/src/keychain/txout_index.rs
evanlinjin: Completed.

crates/chain/tests/test_keychain_txout_index.rs
```diff
@@ -367,3 +368,49 @@ fn test_non_wildcard_derivations() {
...
+
+    assert_eq!(external_spk.next().unwrap(), (0, external_spk_0));
+    assert_eq!(external_spk.nth(15).unwrap(), (16, external_spk_16));
+    assert_eq!(external_spk.nth(3).unwrap(), (20, external_spk_20));
+    assert_eq!(external_spk.next().unwrap(), (21, external_spk_21));
+    assert_eq!(
+        external_spk.nth(max_index as usize).unwrap(),
+        (BIP32_MAX_INDEX, external_spk_max)
+    );
+    assert_eq!(external_spk.nth(0), None);
```
evanlinjin: It'll be good to have tests for checking the `end` bound is enforced with a wildcard descriptor.

evanlinjin:
I think the next step would be to squash all these commits in this PR into one commit and cherry-pick it onto #894 to see if we can fix https://github.com/bitcoindevkit/bdk/pull/894#issuecomment-1494857766.

crates/chain/src/keychain/txout_index.rs
LLFourn: > You can disregard my comment above. I think we need the `Cow<>` so that `SpkIterator` will be Send+Sync, but I still need to verify it solves my issue in #894.

It only needs to be `Send + 'static` I think. What do we need `Sync` for? It's only being accessed from one thread. We need to remove the `Cow` so it can be `'static`. Can't this just take `T: Borrow<Descriptor<DescriptorPublicKey>>` or something like that.

LagginTimes: With @evanlinjin's assistance, `Cow<>` has removed in lieu of `T: Borrow<Descriptor<DescriptorPublicKey>>`.

notmandatory: > > You can disregard my comment above. I think we need the `Cow<>` so that `SpkIterator` will be Send+Sync, but I still need to verify it solves my issue in #894.
> 
> It only needs to be `Send + 'static` I think. What do we need `Sync` for? 

Agreed, doesn't need to be `Sync`.  And make sense to use `Borrow<>` here.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -559,32 +555,107 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+/// ```
+/// use bdk_chain::keychain::SpkIterator;
+/// # use miniscript::{Descriptor, DescriptorPublicKey};
+/// # use bitcoin::{secp256k1::Secp256k1};
+/// # use std::str::FromStr;
+/// # let secp = bitcoin::secp256k1::Secp256k1::signing_only();
+/// # let (descriptor, _) = Descriptor::<DescriptorPublicKey>::parse_descriptor(&secp, "wpkh([73c5da0a/86'/0'/0']xprv9xgqHN7yz9MwCkxsBPN5qetuNdQSUttZNKw1dcYTV4mkaAFiBVGQziHs3NRSWMkCzvgjEe3n9xV8oYywvM8at9yRqyaZVz6TYYhX98VjsUk/1/0)").unwrap();
+///
+/// // Creates a new script pubkey iterator starting at 0 from a descriptor.
+/// let _spk_iter = SpkIterator::new(&descriptor);
```
LLFourn: no need for `_` prefix in example. Use the iterator in the example.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -559,32 +555,107 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+            BIP32_MAX_INDEX
+        } else {
+            0
+        };
+
+        SpkIterator::new_with_range(descriptor, 0..=end)
+    }
+
+    /// Creates a new script pubkey iterator from a descriptor with a given range.
+    pub fn new_with_range<R>(descriptor: D, range: R) -> Self
```
LLFourn: I don't think we need `new_with_range`. You can just `.skip` to the start of the range and `.take` the length of the range you want.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -559,32 +555,107 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+where
+    D: Borrow<Descriptor<DescriptorPublicKey>>,
+{
+    type Item = (u32, Script);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // For non-wildcard descriptors, we expect the first element to be Some((0, spk)), then None after.
+        // For wildcard descriptors, we expect it to keep iterating until exhausted.
+        let has_wildcard = self.descriptor.borrow().has_wildcard();
+        if self.next_index >= self.end || (has_wildcard && self.next_index > BIP32_MAX_INDEX) {
```
LLFourn: You only need to check `self.next_index >= self.end`. `end` must be at most `BIP32_MAX_INDEX`.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -418,16 +417,13 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-                        next_reveal_index..index + 1,
-                    ),
+                    SpkIterator::new_with_range(descriptor.clone(), next_reveal_index..index + 1),
                     DerivationAdditions(core::iter::once((keychain.clone(), index)).collect()),
                 )
             }
             None => (
-                range_descriptor_spks(
-                    Cow::Owned(descriptor.clone()),
+                SpkIterator::new_with_range(
```
LLFourn, line 455: The return value of this function should have `SpkIterator` not `impl Iterator`

LLFourn (changes requested):
Looking good. I don't think we want `new_with_range`.

I think once it's ready we should just merge this. No need to cherry pick. You can make a test checking whether this has done its job by making a dummy trait in a test like `trait TestSendStatic: Send + 'static {}` and `impl TestSendStatic for SpkIterator<Descriptor<DescriptorPublicKey>> { }`. This will fail to compile if it didn't work.
Put the `SpkIterator` into its own `spk_iter.mod` file. Put tests for it at the end of the module. 
Make sure methods that return an `SpkIterator` return it and not an `impl Iterator`.

crates/chain/src/keychain/txout_index.rs
evanlinjin: That is a good point. So we don't even need the `end` field of `SpkIterator`.

evanlinjin: However, I would prefer that we keep this method, just redefine it to use `.skip` and `.take`. This is a helpful method to use imo.

crates/chain/src/keychain/txout_index.rs
evanlinjin: I guess we can have a `debug_assert` somewhere to ensure `self.next_index` is never larger than `BIP32_MAX_INDEX`.

crates/chain/src/keychain/txout_index.rs
evanlinjin: @LLFourn Are you concerned because we are using the `RangeBounds` trait in a "weird" way?

LLFourn: I just don't know why we want to create a method that just calls `.skip` and `.take` when the caller can just do this. We don't have to document it or test it then.

evanlinjin:
@LLFourn Here is my argument for leaving `SpkIterator::new_with_range` in.

`KeychainTxOutIndex::reveal_to_target`, in this PR, now returns an explicit concrete type of `SpkIterator`. This cannot be achieved if we depend on `.skip` and `.take`. Instead we would need to return an implicit `impl Iterator<...>`.

For me, this justifies leaving it as is, since it provides a better API.

crates/chain/src/keychain/txout_index.rs
evanlinjin: Please refer to https://github.com/bitcoindevkit/bdk/pull/927#issuecomment-1527938350

crates/chain/src/spk_iter.rs
```diff
@@ -0,0 +1,208 @@
...
+/// # use std::str::FromStr;
+/// # let secp = bitcoin::secp256k1::Secp256k1::signing_only();
+/// # let (descriptor, _) = Descriptor::<DescriptorPublicKey>::parse_descriptor(&secp, "wpkh([73c5da0a/86'/0'/0']xprv9xgqHN7yz9MwCkxsBPN5qetuNdQSUttZNKw1dcYTV4mkaAFiBVGQziHs3NRSWMkCzvgjEe3n9xV8oYywvM8at9yRqyaZVz6TYYhX98VjsUk/1/0)").unwrap();
+/// # let external_spk_0 = descriptor.at_derivation_index(0).script_pubkey();
+/// # let external_spk_3 = descriptor.at_derivation_index(3).script_pubkey();
+/// # let external_spk_4 = descriptor.at_derivation_index(4).script_pubkey();
+///
+/// // Creates a new script pubkey iterator starting at 0 from a descriptor.
+/// let mut spk_iter = SpkIterator::new(&descriptor);
+/// assert_eq!(spk_iter.next().unwrap(), (0, external_spk_0));
```
evanlinjin: Just a small nit:

```suggestion
/// assert_eq!(spk_iter.next(), Some((0, external_spk_0)));
```

Same goes to the `assert_eq!`s below.

evanlinjin (approved):
ACK 8b7555927d0b326fa1b6ed8fa44ab73b0bb88c3b

This looks good to me. I guess the last thing to do is to do the "dummy trait" test as mentioned by @LLFourn: https://github.com/bitcoindevkit/bdk/pull/927#pullrequestreview-1393215840.

There is also a small nit.

Also note that I rebased for you (sorry!), so you need to force pull.

crates/chain/src/spk_iter.rs
```diff
@@ -0,0 +1,208 @@
...
+    where
+        R: RangeBounds<u32>,
+    {
+        let mut end = match range.end_bound() {
+            Bound::Included(end) => *end + 1,
+            Bound::Excluded(end) => *end,
+            Bound::Unbounded => u32::MAX,
+        };
+        // Because `end` is exclusive, we want the maximum value to be BIP32_MAX_INDEX + 1.
+        if end > BIP32_MAX_INDEX + 1 {
```
LLFourn: nit `end = end.min(BIP32_MAX_INDEX + 1)` is easier to read and more compact.

LLFourn:
> ACK [8b75559](https://github.com/bitcoindevkit/bdk/commit/8b7555927d0b326fa1b6ed8fa44ab73b0bb88c3b)
> 
> This looks good to me. I guess the last thing to do is to do the "dummy trait" test as mentioned by @LLFourn: [#927 (review)](https://github.com/bitcoindevkit/bdk/pull/927#pullrequestreview-1393215840).

I agree. 


> @LLFourn Here is my argument for leaving `SpkIterator::new_with_range` in.
> 
> `KeychainTxOutIndex::reveal_to_target`, in this PR, now returns an explicit concrete type of `SpkIterator`. This cannot be achieved if we depend on `.skip` and `.take`. Instead we would need to return an implicit `impl Iterator<...>`.
> 
> For me, this justifies leaving it as is, since it provides a better API.

Do you mean it provides a API internally? Then I suggest we make it a private function and remove the documentation for it.

LLFourn (approved):
Nits aside LGTM

evanlinjin (changes requested):
We need to make this build with:

```bash
$ cargo build --no-default-features -p bdk_chain
```

To fix this, only include `spk_iter` module if cargo feature `"miniscript"` is enabled.

Also a nit: move `keychain::BIP32_MAX_INDEX` into `spk_iter` module.

evanlinjin (approved):
ACK 10fe32e6f10091c0c43aba3f66e0f700a06f8c89

Issue #895, opened 2023-03-16
Redesigning bdk_chain around a "ChainOracle"


LLFourn:
BDK currently depends heavily on a `SparseChain` to tell it whether transactions are in the chain or not. This proposal is to redesign it around a more generic `ChainOracle` trait which can tell you whether a `(height, block_hash)` is in the chain or not and use this information to decide whether a transaction is in the chain. We can do this because for each transaction we can always know a block header and hash that implies that it is in the chain.

A `SparseChain` can also fill the role of a `ChainOracle` but this change would allow us use other sources of chain data to fulfil the need. For example with our pending `nakamoto` cbf implementation we could use the on-chain block headers as the `ChainOracle`.

This proposal is to remove `SparseChain` as a core mechanism and instead pass in a `impl ChainOracle` into methods that need to know about the current state of the chain.

```rust
trait ChainOracle {
    type Error;
    fn is_in_best_chain(&self, block_height: u32, block_hash: BlockHash) -> Result<bool, Self::Error>;
}
```

## Removing transactions from `SparseChain`

I think we should remove transactions from the sparse chain. Without needing to "re-org" the chain there is no real need to have transaction location so closely coupled to the blocks.

Instead of storing the "position" of a transaction within `SparseChain` we could easily store it in `TxGraph` in a monotone way. Imagine a field like this in `TxGraph`:

```
positions: HashSet<(Txid, (u32, BlockHash))>
```

Then when you want to figure out if `txid` is in the chain you can query the heights and hashes that it has been seen in and see if any of those blocks are in the canonical chain.

Note that we don't need this position to be the blockhash the transaction is in. We could call this the "anchor block" -- if it's in the chain then this txid is in the chain. This means we don't need to make additional queries to electrum/esplora to figure out the block hash that it's actually in.

So maybe it could be more like this:

```rust
positions: HashSet<(Txid, A)>
```

where `A` does the `Anchor` trait which has one method to return the block hash and height whose presence implies the existence -- it may be the height the transaction is actually at or it may be simple the tip that implies its existence. This allows different applications to store different data about their transaction's positions.

## Indexing the chain data

Now that the chain is out of the way and outside of the transaction system we can put the indexing of transactions next to the graph.

```rust
pub struct IndexedTxGraph<I> {
    graph: TxGraph<T>,
    indexer: I
}

// Where I does a trait like this
pub trait TxIndexer {
    // index a whole tx
    fn index_tx(&mut self, tx: &Transaction);
[...]

Issue #859, opened 2023-02-10
RPC sync failure 


rajarshimaitra:
## `Resource temporarily unavailable` failure in RPC

`[2023-02-10T06:50:19Z ERROR bdk_cli] Rpc(JsonRpc(Transport(SocketError(Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }))))`

It's been known for some time that our rpc sync produces intermittent failures with the above error. This has been reported across many instances. But it doesn't caught in our CI. I have also struggled before to replicate this. Sometimes happens locally while running many test instances parallel. But it doesn't happen while running the sync manually. This made it very hard to debug.

First observed while upgrading `electrsd`'s bitcoind version to `0.23`: https://github.com/bitcoindevkit/bdk/pull/613#issuecomment-1138535751

Reported for mac by @notmandatory here : https://github.com/bitcoindevkit/bdk/issues/749

Reported by @evanlinjin  here : https://github.com/bitcoindevkit/bdk/issues/650#issuecomment-1183790197

And as @afilini  [suggested](https://github.com/bitcoindevkit/bdk/issues/650#issuecomment-1184262191), it's an internal TCP error, either by hitting bitcoin core's parallel request limits, or something even lower at os level. 

Recently this has been reported to be happening consistently in `signet`, and I have been able to reproduce it enough to be able to debug successfully.

## To Reproduce  

Sync any bdk wallet, from a fresh database, in  `signet/testnet` network using the RPC blcokchain.

Below is a `sync` output with `bdk-cli` in `debug` mode. And the sync ends with the above error. 
```
[2023-02-10T06:50:04Z DEBUG bdk_cli] network: Signet
[2023-02-10T06:50:04Z INFO  bdk::database::sqlite] db up to date, no migration needed
[2023-02-10T06:50:04Z DEBUG bdk_cli::utils] database opened successfully
[2023-02-10T06:50:04Z DEBUG bitcoincore_rpc::client] JSON-RPC request: getnetworkinfo []
[2023-02-10T06:50:04Z INFO  bdk::blockchain::rpc] connected to 'http://127.0.0.1:38332/wallet/coldcard-signet' with auth: UserPass { username: "mempool", password: "mempool" }
[2023-02-10T06:50:04Z DEBUG bitcoincore_rpc::client] JSON-RPC request: listwallets []
[2023-02-10T06:50:04Z INFO  bdk::blockchain::rpc] wallet already loaded: coldcard-signet
[...]

PR #928, opened 2023-03-24
Add rpc wallet example crates

rajarshimaitra:
### Description

This PR adds the RPC wallet syncing codes. This includes the `rpcwallet` library, `keychain_tracker_rpc` command line app, and `wallet_rpc` BDK wallet example.

This works on the existing structure of `bdk_chain` and might require update after https://github.com/bitcoindevkit/bdk/issues/895.

### Notes to the reviewers

This still has many rough ages. 
 - The `RpcClient` currently initiates a legacy wallet, into which we import all the tracker addresses. This still causes the problem reported in https://github.com/bitcoindevkit/bdk/issues/859. The work-around for now added in the code is to ignore the `resource unavailable error` and carry on. Looking for more suggestions. One future hope is, using the core's descriptor wallet and work with descriptor imports. But still might not fix this RPC issue.
 
  - I added a new `Wallet` API `dump_spks` which is useful for getting all the existing spks from the tracker without rederiving them. 
  
  - Basic docs are added. If you see anything missing or inappropriate let me know.  
   
  - What should be the approach of testing? Should we think of recreating the previous `blockchian tests`?

### Changelog notice

 - Add the new rpc-wallet syncing examples.

#### All Submissions:


notmandatory:
I seem to recall a discussion about syncing a new 1.0 wallet with RPC using full blocks, or I could be conflating with a different discussion. Either way.. in theory would it be possible to sync a bdk 1.0 wallet by scanning all the blocks from the birth date of the wallet, and then updating with new blocks every time the chain tip changes? This question came up in a discussion I was having with @bodymindarts about improving sync performance for very large (200K+) address wallets.

evanlinjin:
@notmandatory That would be a separate example!

evanlinjin:
Closing as it is stale and is based on codebase from a while back. In the future we may want this, however we had no requests so far, and I would assume block-by-block would be the way to go.

PR #930, opened 2023-03-28
Add default `std` feature, prep for 0.28.0 release

notmandatory:
### Description

Add new `std` feature which enables `bitcoin/std` and `miniscript/std`, and enable `std` feature in `default` feature set.

### Notes to the reviewers

When using `bdk` in a wasm environment then `default` features should be disabled and `bitcoin/no-std` and `miniscript/no-std` must be enabled.  This is a breaking change for anyone using `bdk` with `--no-default-features`.

I also updated the workflows to also run for `release/*` branches in addition to `master`.

### Changelog notice

Changed

- Added new `std` feature as part of `default` features, `std` must be enabled unless building for wasm.

#### All Submissions:

#### New Features:

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -2,6 +2,9 @@ name: Audit
 
 on:
   push:
+    branches:
+      - 'master'
+      - 'release/*'
     paths:
       - '**/Cargo.toml'
       - '**/Cargo.lock'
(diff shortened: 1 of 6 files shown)


rajarshimaitra (approved):
ACK cbcbdd120d48742ed70fcb47c159266ae6c5b1ae

I tested the builds with a few regular combinations. Looks good to me.

PR #926, opened 2023-03-24
Introduce redesigned `bdk_chain` structures

evanlinjin:
### Description

This is part of bitcoindevkit/bdk#895

The initial `bdk_chain` structures allowed updating to be done without blocking IO (alongside many other benefits). However, the requirement to have transactions "perfectly positioned" in our `SparseChain` increased complexity of the syncing API. Updates needed to be meticulously crafted to properly connect with the original `SparseChain`. Additionally, it forced us to keep a local copy of the "best chain" data (which may not always be needed depending on the chain source).

The redesigned structures, as introduced by this PR, fixes these shortcomings.

1. Instead of `SparseChain`, we introduce the ability to `Anchor` a transaction to multiple blocks that may or may not be in the same chain history. We expand `TxGraph` to records these anchors (while still maintaining the *monotone* nature of `TxGraph`). When updating our new `TxGraph`, we don't need to complicated *are-we-connected* checks that we need for `SparseChain`. 
2. The chain source, in combination with our new`TxGraph` is all that we need to determine the "chain position" of a transaction. The chain source only needs to implement a single trait `ChainOracle`. This typically does not need to be updated (as it is remote), although we can have a special `ChainOracle` implementation that is stored locally. This only needs block height and hash information, reducing the scope of *non-monotine* structures that need to be updated.

**What is done:**

* `TxGraph` includes anchors (ids of blocks that the tx is seen in) and last-seem unix timestamp (for determining which non-confirmed tx we should consider as part of "best chain" if there are conflicts). This structure continues to be fully "monotone"; we can introduce data in any manner and not risk resulting in an inconsistent state.
* `LocalChain` includes the "checkpoint" logic of `SparseChain` but removes the `txid` data. `LocalChain` implements the `ChainOracle` trait. Any blockchain-source can also implement the `ChainOracle` trait.
* `IndexedTxGraph` is introduced and contains two fields; an internal `TxGraph` struct and a `TxIndex` implementation. These two fields will be updated atomically and can replace the functionality of `keychain::Tracker`.

**What is in-progress:**

* ~@evanlinjin: The `TxIndex` trait should not have `is_{}_relevant` methods as we cannot guarantee this across all transaction indexes. We should introduce extension traits for these (https://github.com/bitcoindevkit/bdk/pull/926#discussion_r1159286393).~
[...]

PR #931, opened 2023-03-29
Test esplora ext

vladimirfomene:
### Description

This adds test for scanning using both the blocking and async Esplora client. This code is subject to change given the redesign of BDK chain. I will leave this as a draft and will pick it up once #926 .

### DIFF
--- a/crates/esplora/Cargo.toml
+++ b/crates/esplora/Cargo.toml
@@ -17,6 +17,15 @@ esplora-client = { version = "0.3", default-features = false }
 async-trait = { version = "0.1.66", optional = true }
 futures = { version = "0.3.26", optional = true }
 
+[dev-dependencies]
+electrsd = { version = "0.22", features = ["legacy", "esplora_a33e97e1", "bitcoind_22_0"] }
+tokio = { version = "1.20.1", features = ["full"] }
+anyhow = "1"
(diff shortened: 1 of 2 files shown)


notmandatory:
Can we keep this in the alpha.1 milestone? does it need anything more than a rebase now that #976 is merged?

notmandatory:
Moved this to beta release so we can focus final alpha releases on functional changes, leaving additional test/example/doc changes to beta milestone.

LLFourn:
I think this PR is being superseded by @LagginTimes if I understand correctly. @vladimirfomene how to avoid wasting your time by having PRs like this slip is something we need to discuss. Not that it was totally wasted - this PR was a good starting point for future work but I feel like:

1. The task was not clearly described -- do we have an issue for this? Some parts of the approach needed to be guided. This PR uses `KeychainTxOutIndex` when it should use as little from `bdk_chain` as possible since it's just trying to test esplora.
2. We were doing it ahead of time -- I think what we needed was a bug in the implementation that we could test and fix with the most minimal test possible. Several have been discovered recently which now gives us the real world problem to verify we've solved with testing.

vladimirfomene:
Yes this PR was definitely ahead of its time

PR #937, opened 2023-04-10
Update persistance With new `IndexedTxGraph`

rajarshimaitra:
### Description

This builds on top of #926 and updates the persistence module to include the new IndexedTxGraph.

### Notes to the reviewers

The examples and wallet module isn't updated as I will wait for some reviews and concept ACKs on the overall change.

Had to add one extra trait `Empty` to make the generics fit together.  

### Changelog notice

Introduce `IndexedTxGraphStore` which persists an `IndexedTxGraph` and replace `KeychainStore`. 

#### All Submissions:


crates/chain/src/keychain/persist.rs
evanlinjin, line 1: Instead of modifying this structure (which is specifically for persisting the `KeychainTracker`), I think it's best to create a new file:

`chain/src/persist.rs`

Also, because we have the `Append` trait, the `Persist::stage` field only needs to implement `Append` imo.

```rust
pub struct Persist<A, B, T> {
    backend: B,
    stage: A,
    phantom: PhantomData<T>,
}

impl<A: Append + Default, B: PersistBackend> Persist<A, B> { /* TODO */ }
```

`PersistBackend` will look as follows:

```rust
pub trait PersistBackend<A, T> {
    type WriteError: core::fmt::Debug;
    type LoadError: core::fmt::Debug;
    fn append(&mut self, additions: A) -> Result<(), Self::WriteError>;
    fn load(&mut self, tracker: &mut T) -> Result<(), Self::LoadError>;
}
```

The intention here is that we can make tracker `T` anything we want (tracker is the in-memory representation of all our data). I would expect that `IndexedTxGraph` would be nested somewhere inside the `T` implementation. Because we will support different tracker implementation, the corresponding additions `A` implementation also needs to be a generic. `A` only needs to implement `Append` and `Default`.

evanlinjin (changes requested):
Thank you for starting work on this!

rajarshimaitra:
closing in favor of #943

PR #944, opened 2023-04-21
Use pass by reference for TxId when getting a transaction

benthecarman:
### Description

Small API cleanup I found

#### All Submissions:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -412,7 +412,8 @@ impl<D> Wallet<D> {
     /// Optionally fill the [`TransactionDetails::transaction`] field with the raw transaction if
     /// `include_raw` is `true`.
     pub fn get_tx(&self, txid: Txid, include_raw: bool) -> Option<TransactionDetails> {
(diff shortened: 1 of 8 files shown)


evanlinjin:
@benthecarman Thank you for the PR, what is the intention for this change?

benthecarman:
> what is the intention for this change?

AFIAK this should be a small performance improvement as before a whole `TxId` was being passed around so it had to use its implementation of `Copy` instead of just using the reference.

evanlinjin:
> > what is the intention for this change?
> 
> AFIAK this should be a small performance improvement as before a whole `TxId` was being passed around so it had to use its implementation of `Copy` instead of just using the reference.

What measure do we have to prove we have a "small performance improvement" from passing by value vs by reference?

benthecarman:
Hmm that was my gut intuition but after doing some reading online seems to be a more nuanced debate. Benchmarking this would probably be a decent amount of effort for not much payoff.

Feel free to close if you'd like, was at least a learning experience for me haha

LLFourn:
Yeah I think the best practice is to always pass `Copy` things by value and let LLVM figure out the rest.

PR #953, opened 2023-04-27
Add sat_per_kwu to FeeRate

benthecarman:
Adds a simple conversion tool for going to kilo weight units

### DIFF
--- a/crates/bdk/src/types.rs
+++ b/crates/bdk/src/types.rs
@@ -114,6 +114,11 @@ impl FeeRate {
         self.0
     }
 
+    /// Return the value as satoshi/kwu
+    pub fn sat_per_kwu(&self) -> f32 {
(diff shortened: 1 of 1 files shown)


crates/bdk/src/types.rs
```diff
@@ -114,6 +114,11 @@ impl FeeRate {
         self.0
     }
 
+    /// Return the value as satoshi/kwu
+    pub fn as_sat_per_kwu(&self) -> f32 {
```
LLFourn: using `as` usually implies the conversion was zero cost. I think here `sats_per_kwu` would be fine.

benthecarman: fixed!

notmandatory (approved):
ACK 217ea3321a937c36da161e7e7dce3bdac46238c0

Issue #958, opened 2023-04-28
More tests for `IndexedTxGraph` and `TxGraph`


evanlinjin:
**Describe the enhancement**  

Now that #926 is merged, let's look at having more fine-grained tests.

Please comment below if you wish to work on a test (to avoid duplicate work!).

### For `TxGraph`

- [ ] Test behavior of inserting `Anchor`s without the corresponding transaction existing in the graph. We should allow for this behavior.
    1. Insert various anchors for a non-existing `txid`. Ensure `Additions` are as expected.
    2. Merge the resultant `Additions` of the previous step, and attempt to recover a new `TxGraph` with them (this should succeed).
    3. Now, insert the actual transaction of `txid`. Ensure anchors still exist.
    4. Now, attempt recovery of the `TxGraph` again with the appended additions.
    - @remix75 

- [ ] Test `TxGraph::get_chain_position` and `TxGraph::list_chain_txs` behavior with unconfirmed conflicts. "Unconfirmed" can either mean that the txs do not have anchors, or that the anchors are not in the best chain. Both cases need to be tested. `TxGraph` should always pick the transaction with the latest `last_seen` value. What happens if the conflicts have the same `last_seen` value? - @rajarshimaitra

- [ ] Test `TxGraph::get_chain_position` and `TxGraph::list_chain_txs` behavior with unconfirmed transactions (txs with no anchors) that conflict with transactions that are confirmed. The unconfirmed transactions should not be givne a chain position or be listed. - @LagginTimes

rajarshimaitra:
> Test TxGraph::get_chain_position and TxGraph::list_chain_txs behavior with unconfirmed conflicts. 

I will give this one a shot.

remix7531:
> Test behavior of inserting Anchors without the corresponding transaction existing in the graph. We should allow for this behavior.

I will work on this

remix7531:
> Merge the resultant Additions of the previous step, and attempt to recover a new TxGraph with them

> Now, attempt recovery of the TxGraph again with the appended additions.

What is meant by recovery in this context? We can insert things into a `TxGraph` and get information from it, but how is a `TxGraph` recovered?

evanlinjin:
> What is meant by recovery in this context? We can insert things into a `TxGraph` and get information from it, but how is a `TxGraph` recovered?

[...]

PR #961, opened 2023-05-02
Add test for inserting anchor without tx

remix7531:
### Description

This adds a test case described in [issue958](https://github.com/bitcoindevkit/bdk/issues/958)

#### All Submissions:


crates/chain/tests/test_tx_graph.rs
```diff
@@ -880,3 +881,57 @@ fn test_additions_last_seen_append() {
         );
     }
 }
+
+#[test]
+fn test_inserting_anchor_without_tx() {
+    let mut graph = TxGraph::<BlockId>::default();
+    let txs = [common::new_tx(0), common::new_tx(1)];
+    let chain: LocalChain = (0..=5)
+        .map(|ht| (ht, BlockHash::hash(format!("Block Hash {}", ht).as_bytes())))
```
evanlinjin: We have a `h!("data")` that can be used.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -880,3 +881,57 @@ fn test_additions_last_seen_append() {
...
+        additions,
+        Additions {
+            tx: Default::default(),
+            txout: Default::default(),
+            anchors: BTreeSet::from([(chain.get_block(0).unwrap(), txs[0].txid()),]),
+            last_seen: Default::default(),
+        }
+    );
+
+    graph.apply_additions(additions);
```
evanlinjin: This is interesting. You are re-adding things that are already added. I don't think this should be here, but it would be something interesting to test in a separate test.

The result of adding things that are already added must be no changes being made.

evanlinjin (changes requested):
Thanks for the work so far. Remember that we also want to test recovering the appended additions into a new graph. Refer to https://github.com/bitcoindevkit/bdk/issues/958#issuecomment-1532394373

crates/chain/tests/test_tx_graph.rs
remix7531: No. I am using [`insert_anchor_preview()`](https://bitcoindevkit.org/docs-rs/bdk/nightly/latest/bdk_chain/tx_graph/struct.TxGraph.html#method.insert_anchor_preview) to generate additions.

crates/chain/tests/test_tx_graph.rs
remix7531: `h!` only allows `literals`. Going to change to `expr` to be more general

crates/chain/tests/test_tx_graph.rs
```diff
@@ -822,3 +822,71 @@ fn test_additions_last_seen_append() {
...
+    assert_eq!(graph, graph_recovered);
+
+    // We can not directly check whether an Anchor is part of a TxGraph
+    // so we check if a txid is connected to the correct Anchor / Block.
+    assert_eq!(
+        graph
+            .get_chain_position(&chain, chain.tip().unwrap(), txs[0].txid())
+            .unwrap(),
+        ObservedAs::Confirmed(&con_hei_anchor)
+    );
```
evanlinjin: Thanks for this, there is a `TxGraph::all_anchors` method now.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -822,3 +822,71 @@ fn test_additions_last_seen_append() {
...
+
+#[test]
+fn test_inserting_anchor_without_tx() {
+    let mut graph = TxGraph::<ConfirmationHeightAnchor>::default();
+    let txs = [common::new_tx(0), common::new_tx(1)];
+    let chain: LocalChain = (0..2)
+        .map(|i| (i, h!(i.to_string())))
+        .collect::<BTreeMap<u32, BlockHash>>()
+        .into();
+    let con_hei_anchor = ConfirmationHeightAnchor {
```
evanlinjin: Nit: just name this variable `anchor`.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -822,3 +822,71 @@ fn test_additions_last_seen_append() {
...
+            .unwrap(),
+        ObservedAs::Confirmed(&con_hei_anchor)
+    );
+
+    let additions2 = graph.insert_tx(txs[0].clone());
+    let additions3 = graph.insert_tx(txs[1].clone());
+
+    // graph should be recoverable from additions
+    graph_recovered.apply_additions(additions2);
+    graph_recovered.apply_additions(additions3);
```
evanlinjin: This is cool, but I intended we test "merging" the additions first.

```rust
let mut acc_additions = Additions::default();

// maybe have this as the body of a loop over txs
{
    let additions = graph.insert_tx(tx);
    assert_eq!(additions, todo!());
    acc_additions.append(additions);
}

// recover graph with accumulated additions
let mut recovered_graph = TxGraph::default();
recovered_graph.apply_additions(acc_additions);
assert_eq!(recovered_graph, graph, "accumulated additions from 'graph' should recover the same 'graph'");
```

crates/chain/tests/test_tx_graph.rs
```diff
@@ -822,3 +822,71 @@ fn test_additions_last_seen_append() {
...
+        .collect::<BTreeMap<u32, BlockHash>>()
+        .into();
+    let con_hei_anchor = ConfirmationHeightAnchor {
+        anchor_block: BlockId {
+            height: 0,
+            hash: chain.blocks().get(&0).unwrap().clone(),
+        },
+        confirmation_height: 0,
+    };
+    let additions = graph.insert_anchor(txs[0].txid(), con_hei_anchor);
```
evanlinjin: Could we test inserting multiple anchors as well? (as per spec)

crates/chain/tests/test_tx_graph.rs
```diff
@@ -822,3 +822,71 @@ fn test_additions_last_seen_append() {
         );
     }
 }
+
+#[test]
```
evanlinjin: I think it will be nice to have a comment summarizing what is tested here.

evanlinjin (changes requested):
Good work so far. Few notes.

1. Ensure CI passes.
2. Ensure GitHub can verify your commits (Maybe you need to give your PGP or SSH public key to GitHub?).
3. Please rebase your commits into one.
4. Use loops where you can (for example if you are inserting multiple anchors or multiple txs).

danielabrozzoni (commented):
A few improvements can be made:
- Fix the warnings `cargo clippy` is throwing
- Use a more descriptive commit message. Instead of "tests for issue958", describe which test you're adding in the title, and add "Partially fixes #958" in the description. In general, it should be possible to understand what a commit is doing from the title/description without having to open github

The rest looks good to me.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -822,3 +822,95 @@ fn test_additions_last_seen_append() {
...
+        Additions {
+            tx: Default::default(),
+            txout: Default::default(),
+            anchors: BTreeSet::from([
+                (anchors[0], txs[0].txid()),
+                (anchors[1], txs[1].txid()),
+                (anchors[2], txs[2].txid()),
+            ]),
+            last_seen: Default::default(),
+        }
```
evanlinjin: Nit, for future reference.

```suggestion
        Additions {
            anchors: BTreeSet::from([
                (anchors[0], txs[0].txid()),
                (anchors[1], txs[1].txid()),
                (anchors[2], txs[2].txid()),
            ]),
            ..Default::default()
        }
```

crates/chain/tests/test_tx_graph.rs
```diff
@@ -822,3 +822,95 @@ fn test_additions_last_seen_append() {
...
+    for tx in &txs[1..] {
+        let additions = graph.insert_tx(tx.clone());
+        assert_eq!(
+            additions,
+            Additions {
+                tx: BTreeSet::from([tx.clone()]),
+                txout: Default::default(),
+                anchors: BTreeSet::default(),
+                last_seen: Default::default(),
+            }
```
evanlinjin: Nit
```suggestion
            Additions {
                tx: BTreeSet::from([tx.clone()]),
                ..Default::default()
            }
```

crates/chain/tests/test_tx_graph.rs
```diff
@@ -822,3 +822,95 @@ fn test_additions_last_seen_append() {
...
+    for tx in txs.iter().take(5).skip(4) {
+        assert_eq!(
+            graph
+                .try_get_chain_position(&chain, chain.tip().unwrap(), tx.txid())
+                .unwrap()
+                .unwrap(),
+            ObservedAs::Unconfirmed(0)
+        );
+    }
+}
```
evanlinjin: I don't think it makes sense to use `get_chain_position` for this test. It feels "out of scope". We want a direct way of seeing if we have the correct anchors inserted for a given tx. I would use `get_tx_node` here. This also simplifies the test as we don't need a `LocalChain`.

Note that `get_tx_node` only returns `Some` for full transactions.

evanlinjin: I think checking both `get_tx_node` and `all_anchors` will be good here.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -822,3 +822,95 @@ fn test_additions_last_seen_append() {
...
+    );
+
+    // graph should be recoverable from additions
+    graph_recovered.apply_additions(anchor_additions);
+    assert_eq!(graph, graph_recovered);
+
+    // Check whether Anchors are part of TxGraph
+    for i in 0..3 {
+        assert!(graph.all_anchors().contains(&(anchors[i], txs[i].txid())));
+    }
```
evanlinjin: This is another way of writing this, but I guess it's a matter of taste. Not mandatory (no pun intended).

```suggestion
    assert_eq!(
        graph.all_anchors(),
        &anchors
            .iter()
            .cloned()
            .zip(txs.iter().map(Transaction::txid))
            .take(3)
            .collect()
    );
```

evanlinjin (changes requested):
Thank you again, sorry I think it will be better to reduce the scope of this test.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -822,3 +822,90 @@ fn test_additions_last_seen_append() {
...
+            }
+        );
+    }
+
+    // Transactions 3 and 4 should not have an Anchor
+    for tx in txs.iter().take(5).skip(3) {
+        assert!(graph.get_tx_node(tx.txid()).unwrap().anchors.is_empty());
+    }
+
+    assert!(false)
```
evanlinjin: Woops!

notmandatory:
Since it looks like still a bit of work today I'm moving this out of the alpha.1 milestone.

remix7531:
I do not understand what is missing here. The workflows failed because of [this](https://github.com/bitcoindevkit/bdk/actions/runs/5177377132/jobs/9624816846#step:5:13) 
`error: package log v0.4.19 cannot be built because it requires rustc 1.60.0 or newer, while the currently active rustc version is 1.57.0`

Just tested the workflows locally and it seams to work now.

danielabrozzoni:
Hey @remix7531, sorry for the late response! I think you encountered a problem with our CI, that's why it would work locally but fail here. Can you try rebasing?

oleonardolima:
Is this something still needed for 1.0.0-beta milestone ? (cc @notmandatory @ValuedMammal)

I got a bit confused because I noticed that the #958 issue already has been closed but without the tests covered by this PR.

ValuedMammal:
I think it wouldn't hurt to have a test similar to this

notmandatory:
This testing could help with verifying sqlite schema question in #1712

evanlinjin:
Tests for inserting anchor without tx already exist: https://github.com/bitcoindevkit/bdk/blob/f6c1c61e7611966e20a88aace932a08aabd543f2/crates/chain/tests/test_tx_graph.rs#L1074-L1102

PR #963, opened 2023-05-03
Various tweaks to redesigned structures

evanlinjin:
### Description

The following tweaks are made:

* `TxGraph::try_get_chain_position` used to always exclude unconfirmed transactions with last_seen value of 0. However, what is the point of including a transaction in the graph if it cannot be part of the chain history? Additionally, maybe sometimes we don't wish to use the last_seen field at all. The new behavior will consider unconfirmed transactions with last_seen of 0.
* Introduce `LocalChain::insert_block`. This is necessary as a `LocalChain` is updated with another `LocalChain`. The update `LocalChain` needs a simple way to insert blocks.
* `BlockId` should not implement `Anchor`. Discussion: https://github.com/bitcoindevkit/bdk/pull/926#discussion_r1181373238.
* I have removed `TxGraph::relevant_heights` as it is ambiguous. A simpler and more flexible alternative method `TxGraph::all_anchors` is introduced.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/indexed_tx_graph.rs
+++ b/crates/chain/src/indexed_tx_graph.rs
@@ -12,6 +12,7 @@ use crate::{
 /// A struct that combines [`TxGraph`] and an [`Indexer`] implementation.
 ///
 /// This structure ensures that [`TxGraph`] and [`Indexer`] are updated atomically.
+#[derive(Debug)]
 pub struct IndexedTxGraph<A, I> {
     /// Transaction index.
     pub index: I,
@@ -27,12 +28,14 @@ impl<A, I: Default> Default for IndexedTxGraph<A, I> {
     }
 }
 
-impl<A: Anchor, I: Indexer> IndexedTxGraph<A, I> {
+impl<A, I> IndexedTxGraph<A, I> {
     /// Get a reference of the internal transaction graph.
     pub fn graph(&self) -> &TxGraph<A> {
         &self.graph
     }
+}
 
+impl<A: Anchor, I: Indexer> IndexedTxGraph<A, I> {
     /// Applies the [`IndexedAdditions`] to the [`IndexedTxGraph`].
     pub fn apply_additions(&mut self, additions: IndexedAdditions<A, I::Additions>) {
         let IndexedAdditions {
@@ -217,7 +220,7 @@ impl<A: Anchor, I: OwnedIndexer> IndexedTxGraph<A, I> {
         C: ChainOracle,
         F: FnMut(&Script) -> bool,
     {
-        let tip_height = chain_tip.anchor_block().height;
+        let tip_height = chain_tip.height;
 
         let mut immature = 0;
         let mut trusted_pending = 0;
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -6,13 +6,6 @@ use bitcoin::BlockHash;
 use crate::{BlockId, ChainOracle};
 
 /// This is a local implementation of [`ChainOracle`].
-///
-/// TODO: We need a cache/snapshot thing for chain oracle.
-/// * Minimize calls to remotes.
-/// * Can we cache it forever? Should we drop stuff?
-/// * Assume anything deeper than (i.e. 10) blocks won't be reorged.
-/// * Is this a cache on txs or block? or both?
-/// TODO: Parents of children are confirmed if children are confirmed.
 #[derive(Debug, Default, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct LocalChain {
     blocks: BTreeMap<u32, BlockHash>,
@@ -71,6 +64,11 @@ impl LocalChain {
         }
     }
 
+    /// Get a reference to the inner map of block height to hash.
+    pub fn inner(&self) -> &BTreeMap<u32, BlockHash> {
+        &self.blocks
+    }
+
     pub fn tip(&self) -> Option<BlockId> {
         self.blocks
             .iter()
@@ -78,11 +76,9 @@ impl LocalChain {
             .map(|(&height, &hash)| BlockId { height, hash })
     }
 
-    /// Get a block at the given height.
-    pub fn get_block(&self, height: u32) -> Option<BlockId> {
-        self.blocks
-            .get(&height)
-            .map(|&hash| BlockId { height, hash })
+    /// Get a [`BlockHash`] at the given height.
+    pub fn get_blockhash(&self, height: u32) -> Option<BlockHash> {
+        self.blocks.get(&height).cloned()
     }
 
     /// This is like the sparsechain's logic, expect we must guarantee that all invalidated heights
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -173,6 +173,31 @@ impl LocalChain {
     pub fn heights(&self) -> BTreeSet<u32> {
         self.blocks.keys().cloned().collect()
     }
+
+    /// Insert a block of [`BlockId`] into the [`LocalChain`].
+    ///
+    /// # Error
+    ///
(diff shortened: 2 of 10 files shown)


crates/chain/src/local_chain.rs
```diff
@@ -85,8 +85,23 @@ impl LocalChain {
             .map(|&hash| BlockId { height, hash })
     }
 
-    /// This is like the sparsechain's logic, expect we must guarantee that all invalidated heights
-    /// are to be re-filled.
+    /// Get the most recent blocks of the best chain.
+    ///
+    /// `max_count` is the upper bound of the number of elements to return.
+    pub fn recent_blocks_of_best_chain(&self, max_count: usize) -> BTreeMap<u32, BlockHash> {
```
LLFourn: Seems like another premature convenience method? It's not used in this PR so maybe wait until the PR where you actually want it to introduce it. The caller can do `.iter().rev().take(..)` for now.

I'm also not sure this is the *right* API. Perhaps this should take a height and give you all blocks >= that height and only the tip if that's empty. I don't know if you necessarily want `n` blocks because getting blocks that are 10,000 blocks deep already is a waste of time. I don't know the answer but I can't really decide here.

crates/chain/src/tx_graph.rs
```diff
@@ -349,6 +349,11 @@ impl<A> TxGraph<A> {
             .filter(move |(_, conflicting_txid)| *conflicting_txid != txid)
     }
 
+    /// Iterates over all transaction anchors known by [`TxGraph`].
+    pub fn all_anchors(&self) -> impl ExactSizeIterator<Item = &(A, Txid)> + DoubleEndedIterator {
```
LLFourn: why not just return the `&BTreeSet<..>`. Much more useful in a `.anchors` method.

LLFourn (changes requested):
I left some comments. Other misc things I noticed that are not changed in this PR but could be:

1. `LocalChain::get_block(height) -> BlockId` allows for invalid representation. It should be `get_block_hash(height) -> BlockHash`.
2. It would be nice if `LocalChain` had a way to doing `.blocks` and just getting a reference to the inner BTreeMap without using the `AsRef` implementation.
3. There are all these `TODOs` as doc comments for Local chain. Can you make issues out of the TODOs and remove them from comments.

crates/chain/src/local_chain.rs
evanlinjin: The problem with taking a height is you can't be sure if the method will return any blocks at all, or return a block that you can connect to.

The benefit of `max_count` is that it is the lower bound of the "height range size" returned (so you can use a number that is large enough to guarantee no reorgs). It also acts as the upper bound of elements to clone.

In any case, I agree that this is premature. Will remove.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -27,12 +28,14 @@ impl<A, I: Default> Default for IndexedTxGraph<A, I> {
     }
 }
 
-impl<A: Anchor, I: Indexer> IndexedTxGraph<A, I> {
+impl<A, I> IndexedTxGraph<A, I> {
```
danielabrozzoni, line 31: I'm not sure how idiomatic this is, since we always end up using IndexedTxGraph with `A: Anchor, I: Indexer`, but I'm not opposed to this change if it's useful for us somehow

crates/chain/tests/test_tx_graph.rs
```diff
@@ -694,33 +694,43 @@ fn test_chain_spends() {
         .iter()
         .zip([&tx_0, &tx_1].into_iter())
         .for_each(|(ht, tx)| {
-            let block_id = local_chain.get_block(*ht).expect("block expected");
-            let _ = graph.insert_anchor(tx.txid(), block_id);
+            // let block_id = local_chain.get_block(*ht).expect("block expected");
```
danielabrozzoni: This can be removed?

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -206,86 +207,112 @@ fn test_list_owned_txouts() {
...
+    let fetch =
+        |height: u32,
+         graph: &IndexedTxGraph<ConfirmationHeightAnchor, KeychainTxOutIndex<String>>| {
+            let txouts = graph
+                .list_owned_txouts(
+                    &local_chain,
+                    local_chain
+                        .get_blockhash(height)
+                        .map(|hash| BlockId { height, hash })
+                        .unwrap(),
```
danielabrozzoni: nit: just save this in a `tip` variable instead of repeating it over and over again (unless it's changing between calls, but it doesn't seem so?)

danielabrozzoni (approved):
ACK a56d289eef0ea241bfe29b4c1069966353e4ca57

A few comments but they don't have to be addressed

evanlinjin:
After a call with @LLFourn, the follow things will be added (either to this PR, or a separate PR):

1. The `PersistBackend<T, C>` trait currently has a generic `T` for the "in memory representation". This also constrains `C` to "accompany" `T`, thus requiring us to have something like `Tracker` in #965. The "onioning" of these structures is unfortunate. However, there is a way to do `PersistBackend` without `T` and just rely on `C`! Let's experiment with that to see if it cleans up the API and reduces "onioning".

2. `ChainOracle::get_chain_tip(&self) -> Result<Option<BlockId>, Self::Error>` will be added. This is needed as we need something to input as the "static block" for the `is_block_in_best_chain` call. Additionally, a structure that is capable of implementing `ChainOracle` functionality should also have the functionality to return the tip of the best chain.

3. `OwnedIndexer` is bad. Here is something better:

    ```rust
    pub OwnedTxOuts {
        SpkIndex;
        Iter: Iterator<Item = (Self::SpkIndex, OutPoint)>;
        fn owned_txouts(&self) -> Self::Iter;
    }
    ```
    This thing iterates over all outpoints that we own (alongside the spk index we give it). To get the owned UTXO set, we filter over this. Same thing for balance.

crates/chain/src/local_chain.rs
```diff
@@ -71,18 +64,21 @@ impl LocalChain {
         }
     }
 
+    /// Get a reference to the inner map of block height to hash.
+    pub fn inner(&self) -> &BTreeMap<u32, BlockHash> {
```
LLFourn: just call this `blocks` it's not really a wrapper so `inner` doesn't make much sense.

crates/chain/src/local_chain.rs
```diff
@@ -71,18 +64,21 @@ impl LocalChain {
...
     }
 
-    /// Get a block at the given height.
-    pub fn get_block(&self, height: u32) -> Option<BlockId> {
-        self.blocks
-            .get(&height)
-            .map(|&hash| BlockId { height, hash })
+    /// Get a [`BlockHash`] at the given height.
+    pub fn get_blockhash(&self, height: u32) -> Option<BlockHash> {
+        self.blocks.get(&height).cloned()
```
LLFourn: oh right since we already expose the `BTreeMap` we don't even need this at all?

crates/chain/src/local_chain.rs
```diff
@@ -173,6 +169,31 @@ impl LocalChain {
     pub fn heights(&self) -> BTreeSet<u32> {
         self.blocks.keys().cloned().collect()
     }
+
```
LLFourn, line 167: same with the above we don't need it just expose the `BTreeMap`. Essentially LocalChain is just keeping a consistent BTreeMap so it only needs insert methods?

LLFourn (approved):
Left another idea. Approved. Happy to merge when @evanlinjin is happy.

crates/chain/src/indexed_tx_graph.rs
evanlinjin: Even if it's not useful directly, it may allow building logic around/above it not require as many generic constraints. I would leave it as is unless if there are downsides to this?

PR #965, opened 2023-05-04
Implement persistence with the new structures

evanlinjin:
### Description

This is part of #895 and #971

* Introduce a more generic version of the `keychain::persist::*` structures that only needs a single generic for the changeset type.

Additional changes:

* The `Append` trait has a new method `is_empty`.
* Introduce `Store` structure for `bdk_file_store` (which implements `PersistBackend`).

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/persist.rs
+++ b/crates/chain/src/persist.rs
@@ -1,0 +1,89 @@
+use core::convert::Infallible;
+
+use crate::Append;
+
+/// `Persist` wraps a [`PersistBackend`] (`B`) to create a convenient staging area for changes (`C`)
+/// before they are persisted.
+///
+/// Not all changes to the in-memory representation needs to be written to disk right away, so
+/// [`Persist::stage`] can be used to *stage* changes first and then [`Persist::commit`] can be used
+/// to write changes to disk.
+#[derive(Debug)]
+pub struct Persist<B, C> {
+    backend: B,
+    stage: C,
+}
+
+impl<B, C> Persist<B, C>
+where
+    B: PersistBackend<C>,
+    C: Default + Append,
+{
+    /// Create a new [`Persist`] from [`PersistBackend`].
+    pub fn new(backend: B) -> Self {
+        Self {
+            backend,
+            stage: Default::default(),
+        }
+    }
+
+    /// Stage a `changeset` to be commited later with [`commit`].
+    ///
+    /// [`commit`]: Self::commit
+    pub fn stage(&mut self, changeset: C) {
+        self.stage.append(changeset)
+    }
+
+    /// Get the changes that have not been commited yet.
+    pub fn staged(&self) -> &C {
+        &self.stage
+    }
+
+    /// Commit the staged changes to the underlying persistance backend.
+    ///
+    /// Returns a backend-defined error if this fails.
+    pub fn commit(&mut self) -> Result<(), B::WriteError> {
+        let mut temp = C::default();
+        core::mem::swap(&mut temp, &mut self.stage);
+        self.backend.write_changes(&temp)
+    }
+}
+
+/// A persistence backend for [`Persist`].
+///
+/// `C` represents the changeset; a datatype that records changes made to in-memory data structures
+/// that are to be persisted, or retrieved from persistence.
+pub trait PersistBackend<C> {
+    /// The error the backend returns when it fails to write.
+    type WriteError: core::fmt::Debug;
+
+    /// The error the backend returns when it fails to load changesets `C`.
+    type LoadError: core::fmt::Debug;
+
+    /// Writes a changeset to the persistence backend.
+    ///
+    /// It is up to the backend what it does with this. It could store every changeset in a list or
+    /// it inserts the actual changes into a more structured database. All it needs to guarantee is
+    /// that [`load_from_persistence`] restores a keychain tracker to what it should be if all
+    /// changesets had been applied sequentially.
+    ///
+    /// [`load_from_persistence`]: Self::load_from_persistence
+    fn write_changes(&mut self, changeset: &C) -> Result<(), Self::WriteError>;
+
+    /// Return the aggregate changeset `C` from persistence.
+    fn load_from_persistence(&mut self) -> Result<C, Self::LoadError>;
+}
+
+impl<C: Default> PersistBackend<C> for () {
+    type WriteError = Infallible;
+
+    type LoadError = Infallible;
+
+    fn write_changes(&mut self, _changeset: &C) -> Result<(), Self::WriteError> {
+        Ok(())
+    }
+
+    fn load_from_persistence(&mut self) -> Result<C, Self::LoadError> {
+        Ok(C::default())
+    }
+}
--- a/crates/chain/src/tx_data_traits.rs
+++ b/crates/chain/src/tx_data_traits.rs
@@ -64,20 +64,35 @@ impl<A: Anchor> Anchor for &'static A {
 pub trait Append {
     /// Append another object of the same type onto `self`.
     fn append(&mut self, other: Self);
+
+    /// Returns whether the structure is considered empty.
+    fn is_empty(&self) -> bool;
 }
 
 impl Append for () {
     fn append(&mut self, _other: Self) {}
+
+    fn is_empty(&self) -> bool {
+        true
+    }
 }
 
 impl<K: Ord, V> Append for BTreeMap<K, V> {
     fn append(&mut self, mut other: Self) {
         BTreeMap::append(self, &mut other)
     }
+
+    fn is_empty(&self) -> bool {
+        BTreeMap::is_empty(self)
+    }
 }
 
 impl<T: Ord> Append for BTreeSet<T> {
     fn append(&mut self, mut other: Self) {
         BTreeSet::append(self, &mut other)
+    }
+
+    fn is_empty(&self) -> bool {
+        BTreeSet::is_empty(self)
     }
 }
--- a/crates/chain/src/tx_data_traits.rs
+++ b/crates/chain/src/tx_data_traits.rs
@@ -1,6 +1,7 @@
 use crate::collections::BTreeMap;
 use crate::collections::BTreeSet;
 use crate::BlockId;
+use alloc::vec::Vec;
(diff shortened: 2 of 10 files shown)


crates/file_store/src/store.rs
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    /// Append a new changeset to the file and truncate the file to the end of the appended
+    /// changeset.
+    ///
+    /// The truncation is to avoid the possibility of having a valid but inconsistent changeset
+    /// directly after the appended changeset.
+    ///
+    /// **WARNING**: This method does not detect whether the changeset is empty or not, and will
+    /// append an empty changeset to the file (not catastrophic, just a waste of space).
```
LLFourn: I think it should do this check. I'd add `is_empty` onto `Append`.

evanlinjin: Added in d2e7b48e2c76a34f065705053f6e6955d6878d22

crates/chain/src/persist.rs
```diff
@@ -0,0 +1,89 @@
...
+        let mut temp = C::default();
+        core::mem::swap(&mut temp, &mut self.stage);
+        self.backend.write_changes(&temp)
+    }
+}
+
+/// A persistence backend for [`Persist`].
+///
+/// `T` represents the tracker, the in-memory data structure which we wish to persist.
+pub trait PersistBackend<T, C> {
```
bodymindarts: Has any consideration been given to making this an [`#[async_trait]`](https://docs.rs/async-trait/latest/async_trait/)? IO operations like write / load should generally not be executed synchronously - especially if the persistence backend is a networked database.

LLFourn: This is a good point. Some background on this trait. We'd rather this trait not exist but in order to preserve `Wallet`'s (over) simple API that automatically persists every change you make to it we had to add this trait. I think that if you are using a networked backend you are doing something sufficiently sophisticated to justify not using `bdk::Wallet` (and not using this trait) which is more meant as an example starting point kind of thing.

@evanlinjin thoughts?

evanlinjin: Yes.

In general, the structures in `bdk_chain` have a simple API and easy to design persistence around without requiring a trait. Methods that make changes return a "changeset" which the caller can persist.

evanlinjin: Maybe if we want to do an async version of `Wallet`, then it may be a good idea to have an async trait.

bodymindarts: I see so for more sophisticated use cases you suggest using `bdk_chain` directly. Thanks for this insight!

crates/file_store/src/store.rs
```diff
@@ -0,0 +1,289 @@
...
+impl<'a, C> Store<'a, C>
+where
+    C: Default + Append + serde::Serialize + serde::de::DeserializeOwned,
+{
+    /// Creates a new store from a [`File`].
+    ///
+    /// The file must have been opened with read and write permissions.
+    ///
+    /// [`File`]: std::fs::File
+    pub fn new(magic: &'a [u8], mut db_file: File) -> Result<Self, FileError> {
```
LLFourn, line 54: there is no docs for this very *magical* function argument.

crates/file_store/src/store.rs
```diff
@@ -0,0 +1,289 @@
...
+        db_file.rewind()?;
+
+        let mut magic_buf = Vec::from_iter((0..).take(magic.len()));
```
LLFourn:
```suggestion
        let mut magic_buf = vec![0u8;magic.len];
```

crates/file_store/src/entry_iter.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+impl<'t, T> Iterator for EntryIter<'t, T>
+where
+    T: serde::de::DeserializeOwned,
+{
+    type Item = Result<T, IterError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
LLFourn, line 40: The logic of `EntryIter` seems to have gotten more complex without much benefit. I think it was better to let the caller seek to the right position before creating the `EntryIter` but perhaps I am missing something.

crates/file_store/src/store.rs
```diff
@@ -0,0 +1,289 @@
...
+        Copy,
+        PartialOrd,
+        Ord,
+        PartialEq,
+        Eq,
+        Hash,
+        serde::Serialize,
+        serde::Deserialize,
+    )]
+    enum TestKeychain {
```
LLFourn: this is unused?

crates/file_store/src/store.rs
```diff
@@ -0,0 +1,289 @@
...
+            match self {
+                Self::External => write!(f, "external"),
+                Self::Internal => write!(f, "internal"),
+            }
+        }
+    }
+
+    #[derive(Debug, Default, serde::Serialize, serde::Deserialize)]
+    struct TestChangeSet {
+        pub changes: Vec<String>,
```
LLFourn: :bulb: : just impl `Append` for `Vec<T>` and use it directly in the test.

It could be cool to also implement `Append` for any `(A, B) where A: Append, B: Append` etc so you can easily add your "updated to" kind of variable for block syncing. Can be done at another time.

crates/chain/src/tx_data_traits.rs
```diff
@@ -64,20 +64,35 @@ impl<A: Anchor> Anchor for &'static A {
 pub trait Append {
     /// Append another object of the same type onto `self`.
     fn append(&mut self, other: Self);
+
+    /// Returns whether the structure is considered empty.
+    fn is_empty(&self) -> bool;
```
LLFourn, line 70: Let's say I wanted to implement `Append` for a `u32` to represent a monotonically increasing value -- I guess `u32::MIN` would return true for `is_empty`.

LLFourn (approved):
I left a few nits. Otherwise LGTM.

crates/file_store/src/entry_iter.rs
evanlinjin: I think the changes result in a simpler API. Before, when we construct an `EntryIter` we need to return a `Result` (because seek may error), and the `EntryIter` itself also iterates over `Result`s.

crates/file_store/src/store.rs
evanlinjin: Unfortunately.

crates/chain/src/tx_data_traits.rs
evanlinjin: I think implementing `Append` for `u32` is a rather confusing combination. `Option<u32>` or using a wrapped type makes more sense for me.

crates/chain/src/persist.rs
```diff
@@ -0,0 +1,89 @@
...
+/// `Persist` wraps a [`PersistBackend`] (`B`) to create a convenient staging area for changes (`C`)
+/// before they are persisted.
+///
+/// Not all changes to the in-memory representation needs to be written to disk right away, so
+/// [`Persist::stage`] can be used to *stage* changes first and then [`Persist::commit`] can be used
+/// to write changes to disk.
+#[derive(Debug)]
+pub struct Persist<B, C> {
+    backend: B,
+    stage: C,
```
danielabrozzoni, line 14: nit: maybe it should be S for stage, instead of C?

crates/chain/src/persist.rs
```diff
@@ -0,0 +1,89 @@
...
+        &self.stage
+    }
+
+    /// Commit the staged changes to the underlying persistance backend.
+    ///
+    /// Returns a backend-defined error if this fails.
+    pub fn commit(&mut self) -> Result<(), B::WriteError> {
+        let mut temp = C::default();
+        core::mem::swap(&mut temp, &mut self.stage);
+        self.backend.write_changes(&temp)
```
danielabrozzoni, line 48: Mhhhh maybe put a small comment here, it took me a second to figure out what you were trying to do (commit changes and clean self.stage at the same time)

Also, a shorter way of writing this, but not necessarily a cleaner one, might be:
```
self.backend.write_changes(core::mem::take(&mut self.stage))
```

crates/file_store/src/entry_iter.rs
```diff
@@ -0,0 +1,100 @@
...
+///
+/// Reads and returns an entry each time [`next`] is called. If an error occurs while reading the
+/// iterator will yield a `Result::Err(_)` instead and then `None` for the next call to `next`.
+///
+/// [`next`]: Self::next
+pub struct EntryIter<'t, T> {
+    db_file: Option<&'t mut File>,
+
+    /// The file position for the first read of `db_file`.
+    start_pos: Option<u64>,
```
danielabrozzoni, line 20: nit: maybe expand in the docs why it's an Option and what happens when this is None

crates/chain/src/persist.rs
LLFourn: actually shouldn't we only clear it in the case that it succeeds.

evanlinjin: That's a good point

evanlinjin:
```rust
    /// Commit the staged changes to the underlying persistance backend.
    ///
    /// Changes that are committed (if any) are returned.
    ///
    /// # Error
    ///
    /// Returns a backend-defined error if this fails.
    pub fn commit(&mut self) -> Result<Option<C>, B::WriteError> {
        if self.stage.is_empty() {
            return Ok(None);
        }
        self.backend
            .write_changes(&self.stage)
            // if written successfully, take and return `self.stage`
            .map(|_| Some(core::mem::take(&mut self.stage)))
    }
```

evanlinjin: 96b10751325dc1bced09f9e864438216400e4ba3

crates/chain/src/persist.rs
evanlinjin: `C` for changeset though :open_mouth:

danielabrozzoni: Then do `changeset: C` :P

notmandatory: I support `changeset: C` 🙂

PR #966, opened 2023-05-04
Multi Descriptor Wallet API [MVP Refactoring]

### LINKED ISSUE #188: BDK Descriptor Improvements: Add multi descriptor capability in wallet

### LINKED ISSUE #486: BDK Descriptor Improvements: 02 - Add multi descriptor capability in wallet

rajarshimaitra:
### Description
Fixes bitcoindevkit/bdk_wallet#188 .

This describes the minimal viable changes to make the wallet APIs generic on K. This gives us a good starting point on to base the remaining questions of multi-desc wallet designs.

The commits are divided into logical chunks. I couldn't separate the major wallet module refactoring commit easily, so that's still a bit pain to review. Following are the commit summaries; more details are added in the commit message.

 -  https://github.com/bitcoindevkit/bdk/pull/966/commits/7aefa75bc9403c4947d3b78c8dc90049b550ca27: Removes keychain info from `LocalUTXO` as it's redundant and forces us to mark `LocalUTXO` and all other derived structure generic on `K`.
 
 - https://github.com/bitcoindevkit/bdk/pull/966/commits/aba7b96baaa0decd01a4e69ea8a64cc500c03d23: Modifies the `TxBuilder` to only work on the default wallet. Because we don't have a suitable `TxParam` that can satisfy multi-desc requirements, we keep the `TxBuilder` default wallet only. This maintains all the previous API and behaviors of `TxBuilder`. Better multi-desc support for `TxBuilder` and `TxParams` can be done on top of this in a future PR.

 - https://github.com/bitcoindevkit/bdk/pull/966/commits/ce6ee5c3d94d01d908aada8dcb0825244518805d: The Major refactoring change. This changes the wallet API to work with generic `K`. Creates a default wallet to work with `KeychainKind`. And adds the following new multi-desc APIs,
     - new_multi_descriptor() : create a multi descriptor wallet.
     - add_descriptor() : Adds a descriptor and marks it with a `K`.
     - get_address_by_keychain (): Get an address for a given keychain.
     - list_utxos_by_keychain (): does what it says.
     - list_transactions_by_keychain() : ''
     - get_balance_by_keychain() : Get a balance for one keychain. The unconfirmed balance is reported as trusted.
     - get_balance_for_keychains() : Get accumulated balance for a list of keychains. uses `should_trust` predicate to determine "untrusted" keychains.
     - create_tx_multi() : The multi-descriptor create transaction API. 
   
The rest of the commits are self-explanatory.

### Notes to Reviewers

The `Balance` type is moved out from `bdk_chain::keychain` module into `bdk::types`. This is to solve the problem described here https://github.com/bitcoindevkit/bdk/pull/966#issuecomment-1536043323. It feels easier to not burden `bdk_chain` to hand us out the balances. It already gives us all the data we need to calculate the balance, so `bdk_chain` can be less opinionated about balance calculations and we can handle the required `get_balance_*` functions at the wallet layer. 

The `create_tx_multi` API takes in a list of keychains to spend from. And a list of change keychains to generate change addresses from. For `change_keychains` is an iterator instead of a single keychain because later, we wanna add the recipient address matching capability of the change address. There could be other ways of doing it, but as it's the simplest of all the floating solutions, kept the API accordingly. Currently, multiple keychains do nothing, and a single one is selected at random to generate change. This will be handled in a separate PR.

### TODO

Tests for basic multi-desc APIs. The newly added APIs and their behaviors should be asserted. 

#### All Submissions:

#### New Features:


crates/bdk/src/types.rs
```diff
@@ -167,7 +167,7 @@ pub struct LocalUtxo {
     /// Transaction output
     pub txout: TxOut,
     /// Type of keychain
-    pub keychain: KeychainKind,
+    //pub keychain: K,
```
evanlinjin: Isn't this important information for an UTXO?

evanlinjin: I see that, instead of returning `LocalUtxo` for the methods, you are returning a tuple `(LocalUtxo, K)` because you have removed the `keychain` field in `LocalUtxo`.

I think it would be better to keep the `keychain: K` field in `LocalUTXO<K>`, and constrain methods that can only use `K` as `KeychainKind` to return/use `LocalUtxo<KeychainKind>`.

crates/bdk/src/wallet/mod.rs
```diff
@@ -615,792 +457,977 @@ impl<D> Wallet<D> {
...
+    /// [Note]: Currently the change address is being selected randomly from the
+    /// given list. This can be improved to select a keychain matching the recipeient's
+    /// address type.
+    pub(crate) fn create_tx_multi<Cs: coin_selection::CoinSelectionAlgorithm>(
+        &mut self,
+        utxo_keychains: Vec<K>,
+        change_keychains: Vec<K>,
+        coin_selection: Cs,
+        params: TxParams,
+    ) -> Result<(psbt::PartiallySignedTransaction, TransactionDetails), Error>
```
evanlinjin: This seems like the wrong API to me. There should only be one keychain which we use to generate the change output (`change_keychain`).

To achieve this, I would replace `TxParams::change_policy` (really bad name, it really means `change_spend_policy`) with something like `spend_keychains: BTreeSet<K>`. What do you think?

Also, I would change `utxo_keychains` to use a generic constrained to `IntoIterator<Item = K>`. Maybe `spend_keychains` will be a better name?

evanlinjin (changes requested):
Thanks for the work! These are my initial thoughts after a preliminary run-through.

I still haven't fully conceptualized how we plan to progress with this to get to our desired API.

I would have done things a little differently. I.e. trying to split out `TxBuilder` completely first. It would be a lot more tedious, but my intuition tells me it leaves us with something cleaner to work with.

Would love to hear everyone's thoughts on this.

crates/bdk/src/wallet/mod.rs
vladimirfomene: I agree with @evanlinjin here. Having only one change_descriptor will also simplify the code. We can have some logic which chooses the right change_keychain outside of this method.

crates/bdk/src/types.rs
rajarshimaitra: > Isn't this important information for an UTXO?

IMO no. These types (also LocalUTXO and WeightedUTXO) are conceptually equivalent for every keychain. They should not be generic on K. Making such primitives types generic would be awkward because we will end up having all other derived types generic too. Like `CoinSelection<K>`. It doesn't make semantic sense.

I think it's preferable to stop the generic keychain progression at `TxBuilder` level only. There are always ways to pass in the keychain with a given utxo. 

Previously it was done because it was convenient. but wasn't "necessary", and it was okay to do because keychains weren't generic.

crates/bdk/src/wallet/mod.rs
rajarshimaitra: >  There should only be one keychain which we use to generate the change output (change_keychain).

This was discussed in the last call. And the context was https://github.com/bitcoindevkit/bdk/issues/486#issuecomment-1509310886. Allowing for choosing different descriptors for different address types.

There were multiple ways to go about this: 
 - @notmandatory : Make it automated by the wallet and don't give change keychainsl in this API at all.
 
 - me: Specify one change keychain at wallet creation time and automatically derive and store all other three types of keychains internally.
 
 - consensus: The easiest way for prototyping is to allow the user to specify multiple change keychais, Have a logic inside to match the keychain corresponding to the recipient address type. If the keychain isn't there in the list, fall back on a default one. Document this behavior. Give an example demo on how to use it properly. If they don't need the feature, they can just pass in a single keychain.

Thus the current API is only to facilitate that. 
 
I agree that this is not ideal and surely can be improved to do more smarter things, but that should be done in a separate PR. 

>  I would replace TxParams::change_policy (really bad name, it really means change_spend_policy) with something like spend_keychains: BTreeSet<K>. What do you think?

The whole `TxParams` needs to change to handle multi-desc. Not just because of this particular issue. It's questionable whether we should even have the TxParams or not. Would be better to keep that work separate. 

Currently, I don't wanna make TxPramas generic on K at this stage. As that would creep up the generic to TxBuilder which can't handle generic keychains anyway right now. 

On the name, yes I totally agree, it's a confusing name. But not touching it here.

crates/bdk/src/wallet/mod.rs
```diff
@@ -615,792 +457,977 @@ impl<D> Wallet<D> {
...
-                if self.is_mine(&drain_script) {
-                    received += amount;
-                }
-                fee_amount += fee;
+        let internal_requirements = internal_policies
+            .iter()
+            .map(|(policy, _)| {
+                Ok(policy.get_condition(
+                    params
+                        .external_policy_path
```
vladimirfomene: Shouldn't this logic be using `internal_policy_path`? I'm not very familiar with policy stuff but given that we now have multiple descriptors should we not keep track of multiple policy paths in the `TxParams` and not just the internal and external one?

rajarshimaitra:
Thanks for the look. 

To give a bit of context. This PR isn't meant to be the final API or the internal design for multi descriptor wallet. There are many potential improvements to make. Some of them I mentioned in the description. But without this, it was hard to figure out what exactly the new designs for other structures should be. 

The aim of this PR is to do the basic changes needed to make the wallet generic on keychains. Which itself was a big refactoring change because of the ways bdk internals were written. Due to this, it's hard to understand what needs to be changed/refactored/redesigned in the other parts of the wallet module.  

> I would have done things a little differently. I.e. trying to split out TxBuilder completely first. It would be a lot more tedious, but my intuition tells me it leaves us with something cleaner to work with.

Completely with you that TxBuilder and wallet decoupling should happen, but that's an orthogonal discussion. And should be handled via another parallel PR. I think @danielabrozzoni and @vladimirfomene has also started thinking about it. 

The aim of this PR is to show the refactoring can be done without a lot of scope-creeping with the existing design and can guide us on how the new design for TxBuilder and TxParams should fit in the multi-descriptor world. That's a separate discussion on its own.

crates/bdk/src/wallet/mod.rs
rajarshimaitra: > Also, I would change utxo_keychains to use a generic constrained to IntoIterator<Item = K>. Maybe spend_keychains will be a better name?

Done..

vladimirfomene (commented):
Thanks Raj! We finally have something on which we can have conversations around. I think we missing an API to get balance per keychain, but we can certainly add it later.

rajarshimaitra:
Thanks for the look Vlad. 

> I think we missing an API to get balance per keychain, but we can certainly add it later.

Good point. I commented this API out because KeychainTracker doesn't impl balance by keychain. This does provide a good starting point to discuss how we should do it. Keychaintracker won't be used eventually, so maybe we do it over the new IndexedTxGraph API and unlock this one over the wallet once the wallet is ported on the new design? 

On the current keychaintracker, it seems pretty easy to do, but it probably will be wasted work. On the IndexedTxGraph it's a bit harder because it works on the `FullTxOut` which doesn't have keychain info. 

Maybe @evanlinjin would have more suggestions on this.

crates/bdk/src/wallet/mod.rs
rajarshimaitra: Good catch. This was a bug. And it doesn't get caught in tests because it doesn't test with complex multi-path change descriptors. We should add such tests.  Fixed in next push.

benthecarman:
Super excited for this one!

crates/bdk/src/wallet/mod.rs
rajarshimaitra: > I agree with @evanlinjin here. Having only one change_descriptor will also simplify the code. We can have some logic which chooses the right change_keychain outside of this method.

I don't see how that will simplify the situation. Few questions:
 - Choose from what? Does the user store different change descriptors while creating the wallet?  
 - If we only give one descriptor as change, does the wallet generate the remaining ones itself? How does it mark them with a K? 
 - If the user gives different keychain at different calls will it recreate all the types of them and track them too?   
 
I was thinking along that line too, but it feels trying to automate this doesn't make it simpler. As @LLFourn also suggested in the comment here https://github.com/bitcoindevkit/bdk/issues/486#issuecomment-1515738913 It's much simpler to just make the user explicitly specify the change. Because in the application layer user will always know about their keychains. If they don't want that special privacy feature, they just specify one single change keychain. This also removes the need to internally mark one or more change keychains at the wallet creation time.

rajarshimaitra:
Pushed some review comment fixes and moved all the remaining wallet APIs (that fits) into the generic definition.

crates/bdk/src/wallet/mod.rs
rajarshimaitra: I have no strong opinions on this though. This is just the simplest way of doing it with the least amount of complexity on the wallet side, so went with it. If we can come up with a more elegant design we should definitely do it.

crates/bdk/src/types.rs
rajarshimaitra: Resolving this as its clear from the changesets that this field is redundant.

crates/bdk/src/wallet/mod.rs
rajarshimaitra: Resolving this, as this will be handled later in a separate PR.

rajarshimaitra:
- Rearranged the commit history. 
 - Added the rest of the remaining APIs for multi-desc wallet. 
 - Cleaned up some rough edges. 
 - Updated the PR description to reflect current state. 
 
This is ready for review from my side. The changeset should look much simpler now. I will move on to add some basic tests to ensure the new multi-desc APIs work.

crates/bdk/src/wallet/mod.rs
```diff
@@ -269,416 +284,625 @@ impl<D> Wallet<D> {
...
-        }
+    /// When this categorization is not needed, passing in `should_trust : false` will default to all keychain balances being
+    /// "trusted".
+    pub fn get_balance_for_keychains(
+        &self,
+        keychains: impl IntoIterator<Item = K>,
+        should_trust: impl Fn(&K) -> bool,
+    ) -> Balance {
+        keychains.into_iter().fold(Balance::default(), |acc, k| {
+            let mut balance = self.get_balance_by_keychain(&k);
```
vladimirfomene: I'm thinking we can have a boolean parameter on `get_balance_for_keychain`. The API can look like `get_balance_by_keychain(&self, keychain: &K, should_trust: bool)`. This will remove the need for `make_untrusted` and also remove this `if` condition.

crates/bdk/src/wallet/mod.rs
```diff
@@ -964,176 +1192,200 @@ impl<D> Wallet<D> {
...
-                .collect(),
-            utxos: original_utxos,
-            bumping_fee: Some(tx_builder::PreviousFee {
-                absolute: fee,
-                rate: feerate.as_sat_per_vb(),
-            }),
-            ..Default::default()
-        };
+    /// Iterator over all keychains in this wallet
+    pub fn keychanins(&self) -> &BTreeMap<K, ExtendedDescriptor> {
```
vladimirfomene: Nits: I think we are suppose to have `keychains` here instead of `keychanins`.

vladimirfomene (commented):
@rajarshimaitra, I also initially thought we will need to decouple the wallet from the transaction builder but I no longer have a strong opinion on that. I think we just need to have a `TxParams` and `TxBuilder` work for our multi-descriptor wallet.

crates/bdk/src/wallet/mod.rs
rajarshimaitra: That's a nice idea.. I also didn't like the awkward `make_untrusted` thingy.. Doing it in the next push..

crates/bdk/src/wallet/mod.rs
rajarshimaitra: Totally..

rajarshimaitra:
>  I think we just need to have a TxParams and TxBuilder work for our multi-descriptor wallet.

I would further decouple this problem and fix the `TxParams` first for multi-desc. My current understanding is the `TxParams` is already good enough to serve the basic multi-desc operations, especially because it doesn't deal with keychains. We only need a new multi-desc `policy_path` field for complex descriptors. Things will be more clear once we have the tests.

evanlinjin:
> To give a bit of context. This PR isn't meant to be the final API or the internal design for multi descriptor wallet. There are many potential improvements to make. Some of them I mentioned in the description. But without this, it was hard to figure out what exactly the new designs for other structures should be.

I'm aware. I'm saying that I don't think this is the correct intermediary step since we have now figured that `TxBuilder` is deeply coupled with `Wallet`. Knowing this, I think the most intuitive approach is to first decouple `TxBuilder`.

> Completely with you that TxBuilder and wallet decoupling should happen, but that's an orthogonal discussion. And should be handled via another parallel PR. I think @danielabrozzoni and @vladimirfomene has also started thinking about it.

I'm not opposed to experimenting with the `Wallet` API while `TxBuilder` is rebuilt. However, I would want the new `TxBuilder` merged before merging this PR. Merging this PR in it's current state leaves `Wallet` with a confused API.

rajarshimaitra:
> I'm not opposed to experimenting with the Wallet API while TxBuilder is rebuilt. However, I would want the new TxBuilder merged before merging this PR. Merging this PR in it's current state leaves Wallet with a confused API.

ACK on this and turning this into draft until we have the new TxBuilder. Happy to open it up later and replay the refactoring on the new APIs.

philipp1992:
Hi,
can you give this PR more attention please? We need that urgently.
kind regards
Philipp

notmandatory:
> Hi, can you give this PR more attention please? We need that urgently. kind regards Philipp

This can currently be done using the `bdk_chain` crate but won't be a feature for `bdk_wallet` until the future 2.0 milestone.

notmandatory:
A lot has changed with bdk since this PR, but I'm leaving it open for research when we get to doing this for 2.0.

evanlinjin:
I don't think this should be left open since the approach here is wrong and this PR is very outdated. bitcoindevkit/bdk#1539 is a more sane approach. Additionally, we have ticket bitcoindevkit/bdk_wallet#84 to reference all related PRs open or not.

PR #967, opened 2023-05-04
Reproduce desciptor checksum

bodymindarts:
This PR documents an issue that the descriptor checksum does not coincide with what bitcoind expects.


rajarshimaitra:
Thanks for the report.

Few more info required:
 - which version of bdk you are using.
 - what steps did you took to produce the descriptors (assuming you got it from bdk)
 
My initial hunch is, this looks very much like a previous bug we had and fixed here https://github.com/bitcoindevkit/bdk/pull/765. This should not happen in latest (v0.28).

Issue #971, opened 2023-05-05
Furthur changes to redesign structures.


evanlinjin:
**Describe the enhancement**

After a call with @LLFourn, the follow things will be added (either to this PR, or a separate PR):

1. ~The `PersistBackend<T, C>` trait currently has a generic `T` for the "in memory representation". This also constrains `C` to "accompany" `T`, thus requiring us to have something like `Tracker` in #965. The "onioning" of these structures is unfortunate. However, there is a way to do `PersistBackend` without `T` and just rely on `C`! Let's experiment with that to see if it cleans up the API and reduces "onioning".~
    -- Completed by #965

2. `ChainOracle::get_chain_tip(&self) -> Result<Option<BlockId>, Self::Error>` will be added. This is needed as we need something to input as the "static block" for the `is_block_in_best_chain` call. Additionally, a structure that is capable of implementing `ChainOracle` functionality should also have the functionality to return the tip of the best chain.

3. ~`OwnedIndexer` is bad. Here is something better:~

    ```rust
    pub trait OwnedTxOuts {
        type SpkIndex;
        type Iter: Iterator<Item = (Self::SpkIndex, OutPoint)>;
        fn owned_txouts(&self) -> Self::Iter;
    }
    ```
    ~This thing iterates over all outpoints that we own (alongside the spk index we give it). To get the owned UTXO set, we filter over this. Same thing for balance.~
    -- Turns out `OwnedTxOuts` is not needed either, all removed in #975

evanlinjin:
For point number 3, I think neither the `OwnedIndexer` or `OwnedTxOuts` traits are needed. Will have a PR soon.

evanlinjin:
I don't know whether 2. is actually needed atm.

notmandatory:
Should be fixed in #976

notmandatory:
@evanlinjin can we close this now that #976 is merged?

evanlinjin:
Fixed in #976

PR #975, opened 2023-05-10
Improve `txout` listing and balance APIs for redesigned structures

evanlinjin:
### Description

As noted in https://github.com/bitcoindevkit/bdk/issues/971#issuecomment-1542408941.

Instead of relying on a `OwnedIndexer` trait to filter for relevant txouts, we move the listing and balance methods from `IndexedTxGraph` to `TxGraph` and add an additional input (list of relevant outpoints) to these methods.

This provides a simpler implementation and a more flexible API.

#### Other Fixes

The `Persist::commit` method is fixed in 96b10751325dc1bced09f9e864438216400e4ba3.

Previously, regardless of whether writing to persistence backend is successful or not, the logic always cleared `self.staged`. This is changed to only clear `self.staged` after successful write.

Additionally, the written changeset (if any) is returned, and `PersistBackend::write_changes` will not be called if `self.staged` is empty.

### Notes to the reviewers

Yes, slightly more boilerplate to do the same things, but less code to maintain and a much more flexible API. Very worth it IMO.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/indexed_tx_graph.rs
+++ b/crates/chain/src/indexed_tx_graph.rs
@@ -1,12 +1,9 @@
-use core::convert::Infallible;
-
 use alloc::vec::Vec;
-use bitcoin::{OutPoint, Script, Transaction, TxOut};
+use bitcoin::{OutPoint, Transaction, TxOut};
 
 use crate::{
-    keychain::Balance,
     tx_graph::{Additions, TxGraph},
-    Anchor, Append, BlockId, ChainOracle, FullTxOut, ObservedAs,
+    Anchor, Append,
 };
 
 /// A struct that combines [`TxGraph`] and an [`Indexer`] implementation.
@@ -29,6 +26,14 @@ impl<A, I: Default> Default for IndexedTxGraph<A, I> {
 }
 
 impl<A, I> IndexedTxGraph<A, I> {
+    /// Construct a new [`IndexedTxGraph`] with a given `index`.
+    pub fn new(index: I) -> Self {
+        Self {
+            index,
+            graph: TxGraph::default(),
+        }
+    }
+
     /// Get a reference of the internal transaction graph.
     pub fn graph(&self) -> &TxGraph<A> {
         &self.graph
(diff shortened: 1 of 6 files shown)


crates/chain/src/indexed_tx_graph.rs
```diff
@@ -29,6 +26,14 @@ impl<A, I: Default> Default for IndexedTxGraph<A, I> {
 }
 
 impl<A, I> IndexedTxGraph<A, I> {
+    /// Construct a new [`IndexedTxGraph`] with a given `index`.
+    pub fn new(index: I) -> Self {
```
LLFourn, line 30: Why can't you pass in both graph and index here. Just index everything in the graph.

crates/chain/src/tx_graph.rs
```diff
@@ -762,107 +763,190 @@ impl<A: Anchor> TxGraph<A> {
...
         &'a self,
         chain: &'a C,
         chain_tip: BlockId,
-    ) -> impl Iterator<Item = Result<FullTxOut<ObservedAs<A>>, C::Error>> + 'a {
-        self.try_list_chain_txouts(chain, chain_tip)
-            .filter(|r| matches!(r, Ok(txo) if txo.spent_by.is_none()))
+        outpoints: impl IntoIterator<Item = (S, OutPoint)> + 'a,
+    ) -> impl Iterator<Item = Result<(S, FullTxOut<ObservedAs<A>>), C::Error>> + 'a {
+        self.try_filter_chain_txouts(chain, chain_tip, outpoints)
+            .filter(|r| !matches!(r, Ok((_, full_txo)) if full_txo.spent_by.is_some()))
```
LLFourn: no.

crates/chain/src/tx_graph.rs
```diff
@@ -762,107 +763,190 @@ impl<A: Anchor> TxGraph<A> {
             .map(|r| r.expect("oracle is infallible"))
     }
 
-    /// List outputs that are in `chain` with `chain_tip`.
+    /// Get a filtered list of outputs from the given `outpoints` that are in `chain` with
+    /// `chain_tip`.
     ///
-    /// Floating ouputs are not iterated over.
+    /// `outpoints` is a list of outpoints we are interested in, coupled with the associated txout's
+    /// script pubkey index (`S`).
```
LLFourn: why does this have to be a script pubkey index. It's just any data you want to associate with the outpoint. You should suggest if you don't have a natural choice suggest to the reader they can just do `vec_of_outpoints.iter().enumerate()` for example.

The type parameter should probably be named `I` and emphasize that it's there for convenience  and can be `()` if you just don't want one. It would be nice if you could even do without it but it looks like necessary plumbing to make getting the balance easy.

LLFourn (changes requested):

crates/chain/src/tx_graph.rs
evanlinjin: Clippy will complain otherwise 🫤

crates/chain/src/indexed_tx_graph.rs
evanlinjin: Indexing data may result in `I::Additions`. Then the method needs to return both `Self` and `IndexedAdditions`. I don't think this is the right API.

The intention for constructing `IndexedTxGraph` with only an index is so you can set up things like keychains and lookahead settings. I don't think you would ever want to add transactions into `TxGraph` before this stage.

crates/chain/src/tx_graph.rs
evanlinjin: Jokes, there is a way of doing it that improves readability by alot.

evanlinjin: https://github.com/bitcoindevkit/bdk/blob/fb75aa94a9f47d53ce7246c14d739fc53dc38197/crates/chain/src/tx_graph.rs#L867-L873

crates/chain/src/tx_graph.rs
evanlinjin: ~How about instead of `I`, we have:~

```rust
impl IntoIterator<Item = impl AsRef<OutPoint>>
```

Okay, bad idea. There is a lot of friction in getting things to implement `AsRef`.

evanlinjin: Changed generic bound to `OI` (outpoint identifier) and improved docs: ed89de752cf20cb80deefda3ddf65a137b669288

crates/chain/src/indexed_tx_graph.rs
LLFourn: Ok understood. I think there should be some way of taking an existing graph and and index and creating an indexed tx graph but we can leave that for another time.

LLFourn (approved):
I think this is better. Looking forward to seeing how this works in the examples.

LLFourn:
ACK ed89de752cf20cb80deefda3ddf65a137b669288

Issue #797, opened 2022-11-14
misleading spendable balance


zoedberg:
**Describe the bug**  

I've noticed a strange behaviour of the `get_balance` API. 

**To Reproduce**  

Creating a new wallet and sending some funds to it produces the following balance:
```
spendable: 0
total: 10000
immature: 0
confirmed: 0
trusted pending: 0
untrusted pending: 10000
```
Since the spendable balance is `0` I assumed it was not possible to send funds, but still I've been able to spend the funds and bdk now reports the following balance:
```
spendable: 8859
total: 8859
immature: 0
confirmed: 0
trusted pending: 8859
untrusted pending: 0
```

Maybe this is not a bug and is just a matter of naming, but to me `spendable` should reflect the amount of funds that is possible to spend.

**Expected behavior**  

I see that you documented `spendable` as the "sum of trusted_pending and confirmed coins", so IMO a more fitting name could be `safe_to_spend` balance.

Moreover we probably need also a `spendable` balance defined as the sum of funds that can be spent, that I think will be the sum of all balances except the `immature` one.

**Build environment**  
 - BDK tag/commit: `v0.24.0` 
 - OS+version: `debian 11`
 - Rust/Cargo version: `1.65.0`
 - Rust/Cargo target: `x86_64-unknown-linux-gnu`

danielabrozzoni:
Concept ACK for using `safe_to_spend` instead of `spendable`, you're right that it is way more clear!

danielabrozzoni:
I had a chat with @hrik2001 and he's going to work on this one :)

danielabrozzoni:
We just discussed this issue in https://github.com/bitcoindevkit/.github/issues/35:

- `safe_to_spend` is superior in every way to `spendable`
- changing the definition of `spendable` "under the hood" is actually quite risky, as users using `spendable` might not notice
- we need to have timelocks in the balance as well (see bitcoindevkit/bdk_wallet#180)
- we have to avoid keep breaking the API at each release, as it's quite annoying

For these reasons, it's better if we delay fixing this issue until we have a clear path for bitcoindevkit/bdk_wallet#180, and at that point, fix both at the same time in BDK 1.0

PoolloverNathan:
My vote goes to depreciating `spendable` making it an alias to a new field `safe_to_spend`, and adding `possibly_safe_to_spend` or something like that that has the above behavior.

LLFourn:
[...]

PR #976, opened 2023-05-11
Reimplement `Wallet`, `ElectrumExt` and `Esplora{Async}Ext` with redesigned structures.

### LINKED ISSUE #938: Update Wallet to work with `IndexedTxGraph`
This is a placeholder for updating Wallet prior to merging #926, #937, and #TBD (update examples).

evanlinjin:
### Description

Closes #938

* Update `Wallet` to use redesigned structures.
* Update `bdk_electrum::ElectrumExt` to produce updates for redesigned structures.
* Update `bdk_esplora::EsploraExt` and `bdk_esplora::EsploraAsyncExt` to produce updates for redesigned structures.
* Added `example-crates/example_cli` library for implementing examples with redesigned structures.
* Added `example-crate/example_electrum` which is an electrum CLI wallet using the redesigned structures.
* Updated `example-crate/{wallet_electrum|wallet_esplora|wallet_esplora_async}` examples to use redesigned structures.
* Remove all old structures.

### Notes to the reviewers

~These changes bump our `all-features` MSRV to `1.60.0` because of the introduction of `bdk_esplora`. As long as the `bdk_chain` and `bdk_wallet` crates hit a MSRV of `1.48.0`, it will be fine (this work is done in #987).~ No longer needed due to #993 

~I had to comment out the examples that use `Wallet` with our chain sources. Once we update the helper-packages for those chain sources, we can also update the examples.~

Possible future improvements for `ElectrumExt`:

* Remove requirement to retry obtaining ALL data after reorg is detected. Transactions can be anchored to a lower block (not block tip), and an `assume_final_depth` value can be used.

* The logic to finalize an update with confirmation time can be improved during reorgs to not require returning an error.

* Use the subscription model of electrum, as intended by the API.

#### All Submissions:

#### New Features:


example-crates/wallet_electrum/src/main.rs
```diff
@@ -1,104 +1,105 @@
...
-use bdk_file_store::KeychainStore;
+// use bdk::{
+//     bitcoin::{Address, Network},
+//     SignOptions, Wallet,
+// };
+// use bdk_electrum::{
+//     electrum_client::{self, ElectrumApi},
+//     ElectrumExt,
+// };
+// use bdk_file_store::KeychainStore;
```
LLFourn, line 11: since this is not a draft can we get rid of all these comments?

crates/bdk/src/wallet/mod.rs
```diff
@@ -1685,52 +1758,62 @@ impl<D> Wallet<D> {
...
+        let (_, derivation_additions) = self
+            .indexed_graph
+            .index
+            .reveal_to_target_multi(&update.keychain);
+        changeset.append(derivation_additions.into());
+        changeset.append(self.indexed_graph.apply_update(update.graph).into());
+
+        let changed = !changeset.is_empty();
+        if changed {
+            self.persist.stage(changeset);
```
LLFourn: No need for this branch. If it is empty it'll just be a noop.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1685,52 +1758,62 @@ impl<D> Wallet<D> {
...
+    where
+        D: PersistBackend<ChangeSet>,
+    {
         self.persist.staged()
     }
 
     /// Get a reference to the inner [`TxGraph`](bdk_chain::tx_graph::TxGraph).
-    pub fn as_graph(&self) -> &bdk_chain::tx_graph::TxGraph {
-        self.keychain_tracker.graph()
+    pub fn as_graph(&self) -> &TxGraph<ConfirmationTimeAnchor> {
```
```suggestion
    pub fn tx_graph(&self) -> &TxGraph<ConfirmationTimeAnchor> {
```

I think `as_` doesn't feel right. `Wallet` is not wrapping a `TxGraph` it just has one you can access. Also it's nice to have `tx_` in front of `graph` since it's not clear what it's a graph of.

crates/bdk/src/wallet/mod.rs
```diff
@@ -497,41 +528,80 @@ impl<D> Wallet<D> {
         &mut self,
         tx: Transaction,
         position: ConfirmationTime,
-    ) -> Result<bool, chain_graph::InsertTxError<ConfirmationTime>> {
-        let changeset = self.keychain_tracker.insert_tx(tx, position)?;
-        let changed = changeset.is_empty();
-        self.persist.stage(changeset);
+        seen_at: Option<u64>,
```
LLFourn: surely you can either insert a tx in a conirmed position OR with a seent_at but not both. You should specify the `seen_at` at the Unconfirmed branch of `ConfirmationTime` or something.

crates/bdk/src/wallet/mod.rs
```diff
@@ -472,10 +498,15 @@ impl<D> Wallet<D> {
...
-        let changed = changeset.is_empty();
-        self.persist.stage(changeset);
+    ) -> Result<bool, local_chain::InsertBlockNotMatchingError>
+    where
+        D: PersistBackend<ChangeSet>,
+    {
+        let changeset = self.chain.insert_block(block_id)?;
+        let changed = !changeset.is_empty();
+        if changed {
+            self.persist.stage(changeset.into());
```
LLFourn: another redundant branch.

LLFourn (changes requested):
Design looks good. I think we should try and avoid adding another argument to `insert_tx` on Wallet.

crates/bdk/src/wallet/mod.rs
evanlinjin: Removed all unnecessary branches.

crates/bdk/src/wallet/mod.rs
evanlinjin: Changed.

crates/bdk/src/wallet/mod.rs
evanlinjin: Updated to have the `ConfirmationTime::Unconfirmed` variant have a `last_seen` field.

example-crates/example_cli/src/lib.rs
```diff
@@ -0,0 +1,780 @@
...
+pub fn run_address_cmd<A, X>(
+    graph: &mut KeychainTxGraph<A>,
+    db: &Mutex<Database<'_, A, X>>,
+    network: Network,
+    cmd: AddressCmd,
+) -> anyhow::Result<()>
+where
+    ChangeSet<A, X>: Default + Append + DeserializeOwned + Serialize,
+{
+    let process_spk = |spk_i: u32, spk: &Script, index_additions: DerivationAdditions<Keychain>| {
```
LLFourn: Make what is happening in this closure so easy to do it doesn't need a closure.

For example, make a few `From` implementations or constructor functions easily create the `ChangeSet` in one line.
There is no need for the branch on `is_empty`.

crates/electrum/src/v2.rs
```diff
@@ -0,0 +1,507 @@
...
+        let mut graph_update = TxGraph::<A>::new(new_txs);
+        for (txid, anchors) in self.graph_update {
+            if let Some(seen_at) = seen_at {
+                let _ = graph_update.insert_seen_at(txid, seen_at);
+            }
+            for anchor in anchors {
+                let _ = graph_update.insert_anchor(txid, anchor);
+            }
+        }
+        dbg!(graph_update.full_txs().count());
```
LLFourn:
```suggestion
```
I'm guess this wasn't meant to be here.

crates/electrum/src/v2.rs
```diff
@@ -0,0 +1,507 @@
...
+
+        if update_tip.hash != client.block_header(update_tip.height as _)?.block_hash() {
+            // [TODO] We should alter the logic so we won't have to return an error. This is to
+            // [TODO] ensure obtained block times are "anchored" to our tip. If we exclude this, it
+            // [TODO] should be "safe" as well. Tx confirmation times would just slightly vary.
+            return Err(Error::Message(format!(
+                "tip changed during update: update_tip={:?}",
+                update_tip
+            )));
+        }
```
LLFourn: I think we should just note that the confirmation time might not be precisely correct if there was a re-org. Throwing an error here is worse than returning slightly off times.

I think electrum really intends for you to use the merkle proof API. Best solution is to do that and attach confirmation time of each block to `LocalChain`.

crates/electrum/src/v2.rs
```diff
@@ -0,0 +1,507 @@
...
+    Ok(update)
+}
+
+fn determine_tx_anchor(
+    anchor_block: BlockId,
+    raw_height: i32,
+    txid: Txid,
+) -> Option<ConfirmationHeightAnchor> {
+    if txid
+        == Txid::from_hex("4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b")
```
LLFourn: this deserves a comment why it's necessary.

example-crates/example_cli/src/lib.rs
```diff
@@ -0,0 +1,728 @@
...
+        // take mutable ref to construct tx -- it is only open for a short time while building it.
+        let (tx, change_info) = create_tx(graph, chain, keymap, cs_algorithm, address, value)?;
+
+        if let Some((index_additions, (change_keychain, index))) = change_info {
+            // We must first persist to disk the fact that we've got a new address from the
+            // change keychain so future scans will find the tx we're about to broadcast.
+            // If we're unable to persist this, then we don't want to broadcast.
+            db.lock()
+                .unwrap()
+                .stage(C::from(KeychainAdditions::from(index_additions)));
```
LLFourn: The comment says we're going to try and persist this to disk but then it just stages it.

example-crates/example_cli/src/lib.rs
```diff
@@ -0,0 +1,728 @@
...
+                coin_select,
+                address,
+                value,
+                broadcast,
+            )
+        }
+    }
+}
+
+pub fn prepare_index<S: clap::Subcommand, CTX: secp256k1::Signing>(
```
LLFourn: Does this need to be `pub`. Does it need to exist at all? I think you can just inline it.

LLFourn (changes requested):
The examples are nice.

Instead of commenting out all these files can you just remove them from workspace for a while?

crates/electrum/src/v2.rs
evanlinjin: I think so too, but probably for a separate PR?

crates/bdk/src/wallet/mod.rs
```diff
@@ -497,41 +468,70 @@ impl<D> Wallet<D> {
...
+            ConfirmationTime::Unconfirmed { last_seen } => (None, Some(last_seen)),
+        };
+
+        let changeset: ChangeSet = self.indexed_graph.insert_tx(&tx, anchor, last_seen).into();
+        let changed = !changeset.is_empty();
         self.persist.stage(changeset);
         Ok(changed)
     }
 
     #[deprecated(note = "use Wallet::transactions instead")]
```
danielabrozzoni: Can't we just remove this whole method?

crates/bdk/src/wallet/mod.rs
```diff
@@ -1685,52 +1689,62 @@ impl<D> Wallet<D> {
     /// transactions related to your wallet into it.
     ///
     /// [`commit`]: Self::commit
-    pub fn apply_update(&mut self, update: Update) -> Result<(), UpdateError>
+    pub fn apply_update(&mut self, update: Update) -> Result<bool, UpdateNotConnectedError>
```
danielabrozzoni, line 1707: nit: the doc should explain the return type

crates/bdk/src/wallet/mod.rs
```diff
@@ -1685,52 +1689,62 @@ impl<D> Wallet<D> {
...
         self.persist.stage(changeset);
-        Ok(())
+        Ok(changed)
     }
 
     /// Commits all curently [`staged`] changed to the persistence backend returning and error when this fails.
     ///
     /// [`staged`]: Self::staged
-    pub fn commit(&mut self) -> Result<(), D::WriteError>
+    pub fn commit(&mut self) -> Result<bool, D::WriteError>
```
danielabrozzoni, line 1730: ditto above

example-crates/example_cli/src/lib.rs
evanlinjin: Already changed in 6f5a2cc56e95b1fb16a88590cf2698a7a75c1209

crates/electrum/src/v2.rs
```diff
@@ -0,0 +1,507 @@
...
+
+impl<K, A> Default for ElectrumUpdate<K, A> {
+    fn default() -> Self {
+        Self {
+            graph_update: Default::default(),
+            chain_update: Default::default(),
+            keychain_update: Default::default(),
+        }
+    }
+}
```
danielabrozzoni: I think this can be derived

example-crates/example_cli/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "example_cli"
```
danielabrozzoni, line 2: Am I right in saying that `example_cli` will eventually replace `keychain_tracker_example_cli`?

example-crates/example_cli/src/lib.rs
```diff
@@ -0,0 +1,728 @@
...
+    #[clap(name = "txout")]
+    TxOut {
+        #[clap(subcommand)]
+        txout_cmd: TxOutCmd,
+    },
+    /// Send coins to an address.
+    Send {
+        value: u64,
+        address: Address,
+        #[clap(short, default_value = "largest-first")]
```
danielabrozzoni: I would use bnb by default

example-crates/example_cli/src/lib.rs
```diff
@@ -0,0 +1,728 @@
...
+
+#[derive(Subcommand, Debug, Clone)]
+pub enum TxOutCmd {
+    List {
+        /// Return only spent outputs.
+        #[clap(short, long)]
+        spent: bool,
+        /// Return only unspent outputs.
+        #[clap(short, long)]
+        unspent: bool,
```
danielabrozzoni, line 155: nit: This could probably be an enum spent_only/unspent_only/all. Same for confirmed.

example-crates/example_cli/src/lib.rs
```diff
@@ -0,0 +1,728 @@
...
+        outputs.push(change_output)
+    }
+
+    let mut transaction = Transaction {
+        version: 0x02,
+        lock_time: chain
+            .get_chain_tip()?
+            .and_then(|block_id| LockTime::from_height(block_id.height).ok())
+            .unwrap_or(LockTime::ZERO)
+            .into(),
```
danielabrozzoni, line 525: This works because we never add timelocks to the assets, so they won't be used, and the lock_time can be the current block for fee sniping purposes.

...But it took me a sec to figure it out, so maybe can you add a small comment?

example-crates/example_electrum/src/main.rs
```diff
@@ -0,0 +1,323 @@
...
+    Keychain,
+};
+
+const DB_MAGIC: &[u8] = b"bdk_example_electrum";
+const DB_PATH: &str = ".bdk_electrum_example.db";
+const ASSUME_FINAL_DEPTH: usize = 10;
+
+#[derive(Subcommand, Debug, Clone)]
+enum ElectrumCommands {
+    /// Scans the addresses in the wallet using the esplora API.
```
danielabrozzoni: nit: electrum, not esplora :)

example-crates/example_electrum/src/main.rs
```diff
@@ -0,0 +1,323 @@
...
+    Scan {
+        /// When a gap this large has been found for a keychain, it will stop.
+        #[clap(long, default_value = "5")]
+        stop_gap: usize,
+        #[clap(flatten)]
+        scan_options: ScanOptions,
+    },
+    /// Scans particular addresses using the esplora API.
```
danielabrozzoni: ditto above

example-crates/example_electrum/src/main.rs
```diff
@@ -0,0 +1,323 @@
...
+            return res;
+        }
+    };
+
+    let response = match electrum_cmd.clone() {
+        ElectrumCommands::Scan {
+            stop_gap,
+            scan_options,
+        } => {
+            let (keychain_spks, c) = {
```
danielabrozzoni: Can you give a more explanatory name to `c`?

danielabrozzoni (commented):
A few comments, I'm still not done reviewing electrum v2. I agree with lloyd that the old ~~electrum~~ example files should be deleted :)

example-crates/example_cli/src/lib.rs
evanlinjin: Changed to this:

```rust
            // We must first persist to disk the fact that we've got a new address from the
            // change keychain so future scans will find the tx we're about to broadcast.
            // If we're unable to persist this, then we don't want to broadcast.
            {
                let db = &mut *db.lock().unwrap();
                db.stage(C::from(KeychainAdditions::from(index_additions)));
                db.commit()?;
            }
```

crates/electrum/src/v2.rs
evanlinjin: The problem with using `derive(Default)` is that `K` and `A` will also be forced to implement `Default` (which are unnecessary bounds).

example-crates/example_cli/Cargo.toml
evanlinjin: Yes

example-crates/example_cli/src/lib.rs
evanlinjin: This idea deserves to be addressed in a separate ticket/PR!

crates/electrum/src/v2.rs
evanlinjin: Issue created: https://github.com/bitcoindevkit/bdk/issues/980

example-crates/example_cli/src/lib.rs
evanlinjin: Inlined.

crates/bdk/src/wallet/mod.rs
evanlinjin: Removed.

crates/bdk/src/wallet/mod.rs
evanlinjin: Added.

crates/bdk/src/wallet/mod.rs
evanlinjin: Done.

example-crates/example_cli/src/lib.rs
evanlinjin: Changed.

example-crates/example_electrum/src/main.rs
evanlinjin: `local_chain`.

evanlinjin:
Oh yeah I still need to remove the commented-out examples.

crates/bdk/src/wallet/mod.rs
```diff
@@ -497,41 +469,57 @@ impl<D> Wallet<D> {
...
+                let tip_height = tip.map(|b| b.height);
+                if Some(height) > tip_height {
+                    return Err(InsertTxError::ConfirmationHeightCannotBeGreaterThanTip {
+                        tip_height,
+                        tx_height: height,
+                    });
+                }
+                (
+                    Some(ConfirmationTimeAnchor {
+                        anchor_block: tip.expect("already checked if tip_height > height"),
```
LLFourn: shouldn't this be the tightest block. i.e. the lowest hight in local chain that is `>=` the height.

crates/bdk/src/wallet/mod.rs
```diff
@@ -497,41 +469,57 @@ impl<D> Wallet<D> {
         &mut self,
         tx: Transaction,
         position: ConfirmationTime,
-    ) -> Result<bool, chain_graph::InsertTxError<ConfirmationTime>> {
-        let changeset = self.keychain_tracker.insert_tx(tx, position)?;
-        let changed = changeset.is_empty();
+    ) -> Result<bool, InsertTxError>
```
LLFourn, line 494: I actually hadn't considered the implications of this change to wallet. I think what you've done makes sense.

crates/bdk/src/wallet/mod.rs
```diff
@@ -249,15 +263,15 @@ impl<D> Wallet<D> {
 
     /// Iterator over all keychains in this wallet
     pub fn keychanins(&self) -> &BTreeMap<KeychainKind, ExtendedDescriptor> {
```
vladimirfomene: Nits: This should be `keychains`

crates/bdk/src/wallet/mod.rs
```diff
@@ -472,10 +441,13 @@ impl<D> Wallet<D> {
...
-        let changeset = self.keychain_tracker.insert_checkpoint(block_id)?;
-        let changed = changeset.is_empty();
-        self.persist.stage(changeset);
+    ) -> Result<bool, local_chain::InsertBlockNotMatchingError>
+    where
+        D: PersistBackend<ChangeSet>,
+    {
+        let changeset = self.chain.insert_block(block_id)?;
+        let changed = !changeset.is_empty();
+        self.persist.stage(changeset.into());
```
vladimirfomene, line 469: Correct me if I'm wrong but shouldn't we only persist if `changed` is true?

crates/bdk/src/wallet/mod.rs
```diff
@@ -497,41 +469,57 @@ impl<D> Wallet<D> {
...
+                    }),
+                    None,
+                )
+            }
+            ConfirmationTime::Unconfirmed { last_seen } => (None, Some(last_seen)),
+        };
+
+        let changeset: ChangeSet = self.indexed_graph.insert_tx(&tx, anchor, last_seen).into();
+        let changed = !changeset.is_empty();
         self.persist.stage(changeset);
```
vladimirfomene, line 526: Again is there a need to stage these changes if `changed` is false.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1019,16 +1003,21 @@ impl<D> Wallet<D> {
...
+        let chain_tip = self.chain.tip().unwrap_or_default();
+
+        let mut tx = graph
+            .get_tx(txid)
+            .ok_or(Error::TransactionNotFound)?
+            .clone();
+
+        let pos = graph
+            .get_chain_position(&self.chain, chain_tip, txid)
+            .ok_or(Error::TransactionNotFound)?;
```
vladimirfomene, line 1042: Maybe we need a better error variant here. For saying the associated chain position for the txid was not found.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1681,56 +1673,71 @@ impl<D> Wallet<D> {
...
+        let mut changeset: ChangeSet = self.chain.apply_update(update.chain)?.into();
+        let (_, index_additions) = self
+            .indexed_graph
+            .index
+            .reveal_to_target_multi(&update.keychain);
+        changeset.append(ChangeSet::from(IndexedAdditions::from(index_additions)));
+        changeset.append(self.indexed_graph.apply_update(update.graph).into());
+
+        let changed = !changeset.is_empty();
         self.persist.stage(changeset);
```
vladimirfomene, line 1720: As above, I believe we should only stage changes if `changed` is false.

vladimirfomene (commented):
Great work Evan! Just have a couple of comments

crates/bdk/src/wallet/mod.rs
evanlinjin: That was my intuition originally as well. However, @LLFourn suggested that the branch is not needed as `stage()` wouldn't do anything with an empty changeset anyway.

crates/bdk/src/wallet/mod.rs
evanlinjin: I agree that the errors aren't great.

However, I disagree about having a single error enum with a large number of variants. I think we should fix errors in one go with a separate PR, otherwise it's duplicate work.

evanlinjin: Created a separate issue.

crates/chain/src/chain_data.rs
```diff
@@ -99,44 +47,31 @@ impl TxHeight {
 )]
 pub enum ConfirmationTime {
     Confirmed { height: u32, time: u64 },
-    Unconfirmed,
+    Unconfirmed { last_seen: u64 },
```
LLFourn: Can we document each variant. Specify the units of last_seen and time.

It's time to add `#[warn(missing_docs)]` on this project and make sure CI is failing on it. Fine to do it in a followup PR but i think it should be done before making the next alpha release.

crates/bdk/src/wallet/mod.rs
```diff
@@ -83,19 +86,18 @@ const COINBASE_MATURITY: u32 = 100;
 pub struct Wallet<D = ()> {
     signers: Arc<SignersContainer>,
     change_signers: Arc<SignersContainer>,
-    keychain_tracker: KeychainTracker<KeychainKind, ConfirmationTime>,
-    persist: persist::Persist<KeychainKind, ConfirmationTime, D>,
+    chain: LocalChain,
+    indexed_graph: IndexedTxGraph<ConfirmationTimeAnchor, KeychainTxOutIndex<KeychainKind>>,
+    persist: Persist<D, ChangeSet>, // [TODO] Use a different `ChangeSet`
```
LLFourn: What does this TODO mean? Can you remove any TODOs you've made during this PR and make an issue for it and reference the issue instead of just `//TODO` in code.

crates/bdk/src/wallet/mod.rs
```diff
@@ -271,23 +285,24 @@ impl<D> Wallet<D> {
...
         let (index, spk) = match address_index {
             AddressIndex::New => {
-                let ((index, spk), changeset) = txout_index.reveal_next_spk(&keychain);
+                let ((index, spk), index_additions) = txout_index.reveal_next_spk(&keychain);
                 let spk = spk.clone();
 
-                self.persist.stage(changeset.into());
+                self.persist
+                    .stage(ChangeSet::from(IndexedAdditions::from(index_additions)));
                 self.persist.commit().expect("TODO");
```
LLFourn: Created https://github.com/bitcoindevkit/bdk/issues/996

LLFourn (approved):
Fantastic work. ACK 965cbe6dde99a1a58168cc0f4bdd320f0aba10e1.

Left nits and notes.

example-crates/example_electrum/src/main.rs
```diff
@@ -106,29 +141,40 @@ fn main() -> anyhow::Result<()> {
...
-                let local_chain = tracker.chain().checkpoints().clone();
-                (spk_iterators, local_chain)
+
+                let c = chain
+                    .blocks()
+                    .iter()
+                    .rev()
+                    .take(ASSUME_FINAL_DEPTH)
+                    .map(|(k, v)| (*k, *v))
+                    .collect::<BTreeMap<u32, BlockHash>>();
```
LLFourn, line 151: Made issue about the ergonomics here: https://github.com/bitcoindevkit/bdk/issues/997

crates/chain/src/chain_data.rs
evanlinjin: Documentation and `#[warn(missing_docs)]` added in 250458ae01e7157c97f3c88f572e4d0ff21c019f

crates/bdk/src/wallet/mod.rs
evanlinjin: This TODO is not needed, there is nothing to TODO here. Removed.

crates/bdk/src/wallet/mod.rs
evanlinjin: Partially resolved in 250458ae01e7157c97f3c88f572e4d0ff21c019f

I think we should get rid of the address helper methods and just have one method that uses `KeychainKind` as a second input.

crates/bdk/src/wallet/mod.rs
evanlinjin: Yes, changed in 9653b06ff06a3ae605393256b45b0c6685230291

crates/bdk/src/wallet/mod.rs
evanlinjin: Added an additional comment in 9653b06ff06a3ae605393256b45b0c6685230291

crates/file_store/README.md
```diff
@@ -1,9 +1,9 @@
 # BDK File Store
 
 This is a simple append-only flat file implementation of
-[`Persist`](`bdk_chain::keychain::persist::Persist`).
+[`Persist`](`bdk_chain::Persist`).
 
-The main structure is [`KeychainStore`](`crate::KeychainStore`), which can be used with [`bdk`]'s
+The main structure is [`Store`](`crate::Store`), which can be used with [`bdk`]'s
```
danielabrozzoni, line 6: Do you think we will eventually build some convenience structs on top of `Store`, similar to what `KeychainStore` was, or will users just use `Store`?

crates/chain/src/chain_data.rs
```diff
@@ -98,45 +48,44 @@ impl TxHeight {
...
 
-impl sparse_chain::ChainPosition for ConfirmationTime {
-    fn height(&self) -> TxHeight {
-        match self {
-            ConfirmationTime::Confirmed { height, .. } => TxHeight::Confirmed(*height),
-            ConfirmationTime::Unconfirmed => TxHeight::Unconfirmed,
-        }
+impl ConfirmationTime {
+    /// Construct an unconfirmed variant using the given `last_seen` time in unix seconds.
+    pub fn unconfirmed(last_seen: u64) -> Self {
```
danielabrozzoni, line 67: Is there a reason for not having also the `confirmed` constructor? 
I'd say either have both or none of them, the asymmetry here bothers me 😅

crates/chain/src/chain_data.rs
```diff
@@ -235,75 +185,32 @@ impl Anchor for ConfirmationTimeAnchor {
...
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
-pub struct FullTxOut<P> {
+pub struct FullTxOut<A> {
     /// The location of the `TxOut`.
     pub outpoint: OutPoint,
     /// The `TxOut`.
     pub txout: TxOut,
     /// The position of the transaction in `outpoint` in the overall chain.
-    pub chain_position: P,
+    pub chain_position: ChainPosition<A>,
```
danielabrozzoni, line 194: I think this makes things way clearer btw :)

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -203,6 +208,24 @@ impl<A: Anchor, IA: Append> Append for IndexedAdditions<A, IA> {
...
+            graph_additions,
+            ..Default::default()
+        }
+    }
+}
+
+impl<A, K> From<DerivationAdditions<K>> for IndexedAdditions<A, DerivationAdditions<K>> {
+    fn from(index_additions: DerivationAdditions<K>) -> Self {
+        Self {
+            graph_additions: Default::default(),
```
danielabrozzoni, line 223: tiny tiny nit but if you're using `..Default::default()` a couple of lines above I think you should use it here too

crates/bdk/src/wallet/mod.rs
```diff
@@ -271,91 +291,99 @@ impl<D> Wallet<D> {
...
     {
-        self._get_address(address_index, KeychainKind::Internal)
+        self._get_address(KeychainKind::Internal, address_index)
+            .expect("persistence backend must not fail")
     }
 
-    fn _get_address(&mut self, address_index: AddressIndex, keychain: KeychainKind) -> AddressInfo
+    /// Return a derived address using the specified `keychain` (external/internal).
+    ///
+    /// If `keychain` is [`KeychainKind::External`], external addresses will be derived (used for
```
danielabrozzoni, line 302: nit: it might be easier to read written as "If `keychain` is [`KeychainKind::External`], external addresses (used for receiving funds) will be derived."

It might be just me though

(Ditto below w/ internal addresses)

crates/bdk/src/wallet/mod.rs
```diff
@@ -271,91 +291,99 @@ impl<D> Wallet<D> {
...
+    /// receiving funds).
+    ///
+    /// If `keychain` is [`KeychainKind::Internal`], internal addresses will be derived (used for
+    /// creating change outputs). If the wallet does not have an internal keychain, it will use the
```
danielabrozzoni, line 306: Mhh I would expect an error if the user asks for an internal address without having the internal keychain

crates/bdk/src/wallet/mod.rs
```diff
@@ -472,66 +460,91 @@ impl<D> Wallet<D> {
...
     pub fn get_balance(&self) -> Balance {
-        self.keychain_tracker.balance(|keychain| match keychain {
-            KeychainKind::External => false,
-            KeychainKind::Internal => true,
-        })
+        self.indexed_graph.graph().balance(
+            &self.chain,
+            self.chain.tip().unwrap_or_default(),
+            self.indexed_graph.index.outpoints().iter().cloned(),
+            |&(k, _), _| k == KeychainKind::Internal,
```
danielabrozzoni, line 546: This reminds me that I had a nice chat w/ some friends regarding the meaning of "trusted" and "untrusted" keychain and how it doesn't really make sense as a distinction. I will post a message on discord explaining (eventually 😅 ) or directly update the balance issue we already have 

But the TL;DW is: addresses are public anyways, everyone can send money to our change addresses, so you should distinguish between trusted and untrusted based on whether you did a certain tx or not, not based on the keychain used

danielabrozzoni (commented):
Partial review for now because I'm too tired to finish it, I looked at `bdk`, `chain` and `file_store`, they look good to me, I just have a few questions/nits

crates/esplora/README.md
```diff
@@ -27,7 +27,7 @@ To use the extension traits:
 // for blocking
 use bdk_esplora::EsploraExt;
 // for async
-use bdk_esplora::EsploraAsyncExt;
+// use bdk_esplora::EsploraAsyncExt;
```
danielabrozzoni, line 30: Why can't it be uncommented? Doesn't compile?

crates/esplora/Cargo.toml
```diff
@@ -13,12 +13,12 @@ readme = "README.md"
...
 [dependencies]
 bdk_chain = { path = "../chain", version = "0.4.0", features = ["serde", "miniscript"] }
-esplora-client = { version = "0.3", default-features = false }
+esplora-client = { version = "0.5", default-features = false }
 async-trait = { version = "0.1.66", optional = true }
 futures = { version = "0.3.26", optional = true }
 
 [features]
-default = ["async-https", "blocking"]
+default = ["blocking"]
```
danielabrozzoni, line 21: Why isn't async-https included by default anymore? Is it incompatible with blocking?

crates/esplora/src/async_ext.rs
```diff
@@ -100,47 +92,35 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
...
-                break;
+
+            if update.chain.insert_block(tip_at_start).is_ok() {
+                break (update, tip_at_start);
             }
-        }
-        let tip_at_start = BlockId {
-            height: self.get_height().await?,
-            hash: self.get_tip_hash().await?,
         };
```
danielabrozzoni, line 123: I really need some more comments in order to understand what's happening here 😅  To me it seems that you're trying to build an update to the local chain that doesn't have any block that might have been reorged:
1 You traverse the local chain backwards
2 For each block height, you retrieve the current block hash
3 You insert the block height and new hash in the update
4 If the hash is the same as the one you already had in the local chain, then you break as you removed any possible reorg
5. You insert the tip, if it's all good then you're done, otherwise you start from scratch as it means you were updating it mid-reorg, and it's better to start again

Does this sound fair? If so, can you copy-paste these steps in the code as comments? 
(Also, do you think it would work if you swapped step 3 and 4? To me it seems more logical to insert in the update only blocks that actually changed, but i don't know how the localchain works, and it might need at least one block in common to be able to update...)

crates/esplora/src/async_ext.rs
```diff
@@ -149,7 +129,7 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
             type IndexWithTxs = (u32, Vec<esplora_client::Tx>);
 
             loop {
-                let futures: FuturesOrdered<_> = (0..parallel_requests)
+                let futures = (0..parallel_requests)
```
danielabrozzoni, line 132: I know you haven't modified this, but I think we can make the code below a bit clearer. Copy-pasting it here as I can't comment on this PR (it's too far away from modified stuff). Lines 143 to 158:
```
                            if n_confirmed >= 25 {
                                loop {
                                    let new_related_txs = client
                                        .scripthash_txs(
                                            &script,
                                            Some(related_txs.last().unwrap().txid),
                                        )
                                        .await?;
                                    let n = new_related_txs.len();
                                    related_txs.extend(new_related_txs);
                                    // we've reached the end
                                    if n < 25 {
                                        break;
                                    }
                                }
                            }

```

I think we could substitute `if n_confirmed >= 25 { loop { ` with a while loop (so that we don't even need the break on like 155). Also I think we should say "n_confirmed == 25" since if I understand what's happening it can  never be more than 25

crates/esplora/src/async_ext.rs
```diff
@@ -180,36 +160,23 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
...
                 let n_futures = futures.len();
 
-                let idx_with_tx: Vec<IndexWithTxs> = futures.try_collect().await?;
-
-                for (index, related_txs) in idx_with_tx {
+                for (index, related_txs) in futures.try_collect::<Vec<IndexWithTxs>>().await? {
                     if related_txs.is_empty() {
                         empty_scripts += 1;
                     } else {
                         last_active_index = Some(index);
```
danielabrozzoni, line 171: Waaaaait are the responses in `idx_with_tx` ordered? Shouldn't we do `last_active_index = max(last_active_index, index);`?

crates/esplora/src/async_ext.rs
```diff
@@ -267,50 +235,35 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
...
-            } else {
-                false
-            }
-        };
-
-        if reorg_occurred {
-            // A reorg occurred, so let's find out where all the txids we found are in the chain now.
-            // XXX: collect required because of weird type naming issues
+        if tip_at_start.hash != self.get_block_hash(tip_at_start.height).await? {
+            // A reorg occurred, so let's find out where all the txids we found are now in the chain
```
danielabrozzoni, line 249: It seems to me that this piece of code is important to find out what happened after a reorg, but we do it only if the reorg happens during the scan, and not every time there's a reorg. Am I wrong?

example-crates/wallet_esplora/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "bdk-esplora-wallet-example"
+name = "wallet_esplora"
```
danielabrozzoni, line 2: A C K ✨

danielabrozzoni (commented):
I ended up reviewing all the esplora code, and now I'm full of questions, even regarding stuff that wasn't touched in this PR. Sorry haha

example-crates/example_cli/src/lib.rs
```diff
@@ -0,0 +1,736 @@
...
+    pub db_path: PathBuf,
+
+    #[clap(env = "BDK_CP_LIMIT", long, default_value = "20")]
+    pub cp_limit: usize,
+
+    #[clap(subcommand)]
+    pub command: Commands<S>,
+}
+
+#[allow(clippy::almost_swapped)]
```
danielabrozzoni, line 52: Weird, I don't have warnings for clippy on cargo 1.69 nor 1.57 🤷🏻‍♀️

notmandatory:
I finally tested this branch using a signet descriptor setup with a lot of addresses to scan and a slightly modified version of `example-crates/wallet_esplora_async`. My test descriptor has 1001 UTXO (500 sats each) spread over 200_000 addresses. Good news is it sync'd without a problem 🎉 . Bad news is it did take a bit of time. I'll retest with more parallel threads and see if that helps. 

cc @bodymindarts 

Test Params:
```rust
const STOP_GAP: usize = 250;
const PARALLEL_REQUESTS: usize = 500;
const ESPLORA_API_URL: &str = "https://mempool.space/signet/api";
const EXTERNAL_DESC: &str = "wpkh([f6c175d5/84'/1'/0']tpubDD3t7WfvBLbiVB1BbxZjeusz76Xb1Hdv1AgjdyEuVWHzbBNN1Fio9X5rfUpk6JWyuTM7YHLQ7CsetUYMtWm3USAFaHzYY4ymd1FXfum76p5/0/*)#2xxpq0d7";
const INTERNAL_DESC: &str = "wpkh([f6c175d5/84'/1'/0']tpubDD3t7WfvBLbiVB1BbxZjeusz76Xb1Hdv1AgjdyEuVWHzbBNN1Fio9X5rfUpk6JWyuTM7YHLQ7CsetUYMtWm3USAFaHzYY4ymd1FXfum76p5/1/*)#mjrqa6ax";
```

Test Results:

```
Wallet balance after syncing: 500500 sats
Wallet sync took 16 minutes and 31 seconds
```
DB file size: 26K

PS. I've been unable to get this test descriptor to sync with bdk 0.28, still trying to nailing down the issue. 

EDIT1:

With `PARALLEL_REQUESTS: usize = 1000;` the sync time was:
```
Wallet sync took 9 minutes and 4 seconds
```
The sync time should scale well with more threads as long as the esplora server is properly sized.

EDIT2: 

With `PARALLEL_REQUESTS` at 2000:
```
Wallet sync took 6 minutes and 21 seconds
```

crates/bdk/src/wallet/mod.rs
notmandatory: This is how the current BDK API works, instead of changing in this PR can we make an issue and discuss there?

crates/bdk/src/wallet/mod.rs
notmandatory: I added a note on #797 so we don't forget to continue the discussion there.

crates/esplora/Cargo.toml
notmandatory: For what it's worth, I tried putting `"async-https"` and the below `use bdk_esplora:EsploraAsyncExt;` back in and all tests build and run fine.

notmandatory (commented):
This is a monumental piece of work. I like all the new type naming. Also it's much easier to read with old code removed.  At our team call we'll discuss where it needs more detailed review and how to do it. If everyone is OK with the `chain`, `electrum` and `esplora` crates I think anything else can be fixed in future alpha releases.

crates/bdk/src/wallet/mod.rs
```diff
@@ -83,19 +86,83 @@ const COINBASE_MATURITY: u32 = 100;
...
+    fn from(indexed_additions: IndexedAdditions<A, DerivationAdditions<K>>) -> Self {
+        Self {
+            indexed_additions,
+            ..Default::default()
+        }
+    }
+}
+
+impl<K, A> From<DerivationAdditions<K>> for ChangeSet<K, A> {
+    fn from(index_additions: DerivationAdditions<K>) -> Self {
```
vladimirfomene: Nits: might be better to call the parameter to the `from` method `derived_additions`.

crates/file_store/README.md
evanlinjin: What convenience structs existed on top of `KeychainStore`? What do you think is missing with `Store`?

crates/chain/src/chain_data.rs
evanlinjin: No particular reason, I only created the `unconfirmed` constructor because it became tedious.

crates/chain/src/chain_data.rs
evanlinjin: This was only possible after getting rid of the old structures.

example-crates/example_cli/src/lib.rs
```diff
@@ -0,0 +1,736 @@
+pub use anyhow;
```
danielabrozzoni, line 1: This whole file would be way easier to review if there was one move-only commit, followed by commits changing stuff, if needed. In this way, I could just review the commits changing the logic. Just something to keep in mind for next time!

example-crates/example_electrum/src/main.rs
```diff
@@ -192,54 +242,77 @@ fn main() -> anyhow::Result<()> {
...
             }
 
-            let local_chain = tracker.chain().checkpoints().clone();
-            // drop lock on tracker
-            drop(tracker);
+            let c = chain
+                .blocks()
+                .iter()
+                .rev()
+                .take(ASSUME_FINAL_DEPTH)
```
danielabrozzoni, line 261: I'm a bit lost. It seems that you're taking the latest ASSUME_FINAL_DEPTH blocks. Why? Can you add a comment in the code explaining what's happening?

crates/electrum/src/electrum_ext.rs
```diff
@@ -0,0 +1,486 @@
...
+impl<K, A: Anchor> ElectrumUpdate<K, A> {
+    pub fn missing_full_txs<A2>(&self, graph: &TxGraph<A2>) -> Vec<Txid> {
+        self.graph_update
+            .keys()
+            .filter(move |&&txid| graph.as_ref().get_tx(txid).is_none())
+            .cloned()
+            .collect()
+    }
+
+    pub fn finalize(
```
danielabrozzoni, line 40: A comment explaining what it means to finalize an update would be helpful

crates/electrum/src/electrum_ext.rs
```diff
@@ -0,0 +1,486 @@
...
+            outpoints,
+            usize::MAX,
+            batch_size,
+        )
+    }
+}
+
+impl ElectrumExt<ConfirmationHeightAnchor> for Client {
+    fn get_tip(&self) -> Result<(u32, BlockHash), Error> {
+        // TODO: unsubscribe when added to the client, or is there a better call to use here?
```
danielabrozzoni, line 175: What does it mean to "unsubscribe when added to the client"?

crates/electrum/src/electrum_ext.rs
```diff
@@ -0,0 +1,486 @@
...
+                Some(ConfirmationHeightAnchor {
+                    anchor_block,
+                    confirmation_height: h,
+                })
+            }
+        }
+    }
+}
+
+fn populate_with_outpoints<K>(
```
danielabrozzoni, line 346: Pls put a comment explaining what this method does. What is being populate with outpoints? The update?

crates/electrum/src/electrum_ext.rs
```diff
@@ -0,0 +1,486 @@
...
+    for outpoint in outpoints {
+        let txid = outpoint.txid;
+        let tx = client.transaction_get(&txid)?;
+        debug_assert_eq!(tx.txid(), txid);
+        let txout = match tx.output.get(outpoint.vout as usize) {
+            Some(txout) => txout,
+            None => continue,
+        };
+        // attempt to find the following transactions (alongside their chain positions), and
+        // add to our sparsechain `update`:
```
danielabrozzoni, line 362: not a sparsechain anymore :)

crates/electrum/src/electrum_ext.rs
```diff
@@ -0,0 +1,486 @@
...
+                hash: current_hash,
+            })
+            .expect("This never errors because we are working with a fresh chain");
+
+        if current_hash == existing_hash {
+            break;
+        }
+    }
+
+    // Insert the new tip so new transactions will be accepted into the sparsechain.
```
danielabrozzoni, line 295: Not a sparsechain anymore

crates/electrum/src/electrum_ext.rs
```diff
@@ -0,0 +1,486 @@
...
+}
+
+impl ElectrumExt<ConfirmationHeightAnchor> for Client {
+    fn get_tip(&self) -> Result<(u32, BlockHash), Error> {
+        // TODO: unsubscribe when added to the client, or is there a better call to use here?
+        self.block_headers_subscribe()
+            .map(|data| (data.height as u32, data.header.block_hash()))
+    }
+
+    fn scan<K: Ord + Clone>(
```
danielabrozzoni, line 180: I think this whole method needs comments explaining what's going on

danielabrozzoni (commented):
I finished reviewing electrum as well, I haven't really looked into electrum_ext as I had troubles figuring out what was happening, I left a few comments asking for more comments (lol)

It's probably obvious at this point but I really think it would have helped w/ reviewing and updating if this whole PR was split in at least 3 or 4 smaller ones (wallet, esplora, electrum, example cli). Something to keep in mind for next time!

crates/chain/src/chain_data.rs
danielabrozzoni: Then I would add the `confirmed` as well

crates/electrum/src/electrum_ext.rs
evanlinjin: The electrum API has a subscription model. We are subscribing to new tips coming in (we get the first tip), but we don't process later tips, so we should unsubscribe from receiving more tips.

crates/chain/src/indexed_tx_graph.rs
evanlinjin: Sorry @danielabrozzoni, I'm going to ignore this one :sweat_smile:

crates/electrum/src/electrum_ext.rs
danielabrozzoni: > The electrum API has a subscription model. We are subscribing to new tips coming in (we get the first tip), but we don't process later tips, so we should unsubscribe from receiving more tips.

Yep, but why "when added to the client" - unsubscribe when what is added to the client, the first tip?

crates/chain/src/indexed_tx_graph.rs
danielabrozzoni: 🤷🏻‍♀️

crates/esplora/src/async_ext.rs
evanlinjin: > it can never be more than 25

I don't think this statement is correct.
https://github.com/Blockstream/esplora/blob/master/API.md#get-scripthashhashtxs

I think we should leave it as is. These are style nits.

crates/esplora/src/async_ext.rs
evanlinjin: Yes, the responses are in order. Did you miss the fact that we are collecting into a `FuturesOrdered`?

crates/esplora/src/async_ext.rs
evanlinjin: I'll update the comments to try make thinks clearer. However, by principle, I don't think comments should say what is happening (because you can just read the code). The "why" is more important.

crates/esplora/Cargo.toml
evanlinjin: I think this was an attempt to keep MSRV to 1.57 after the `base64` mishap that is now fixed. Will change it back.

crates/esplora/src/async_ext.rs
evanlinjin: If a reorg happens during our scan, our view of blocks may not be consistent so we need to retry. We just need to ensure our update is consistent. If our update connects with our original view, we can assume that the resultant state is also consistent.

example-crates/example_cli/src/lib.rs
evanlinjin: I think it was caused by this regression (https://github.com/rust-lang/rust-clippy/issues/10421) which is fixed in the latest version for `cargo clippy`. Will remove.

example-crates/example_cli/src/lib.rs
evanlinjin: Apologies. Will keep in mind for next time.

example-crates/example_electrum/src/main.rs
evanlinjin: Will add a comment. Basically anything as deep or deeper than `ASSUME_FINAL_DEPTH` will not be reorged out (based on our assumption). This is essentially the minimum checkpoints from the tip to include (from the original chain) to guarantee that we will find a point of agreement.

crates/electrum/src/electrum_ext.rs
evanlinjin: Yes, the update. We already have a comment on `scan` explaining the `outpoints` input though?

example-crates/example_electrum/src/main.rs
LLFourn: Note that #1002 is removing this ASSUME_FINAL_DEPTH thing.

LLFourn (approved):
ACK 75f8b81d58a985669ce7302fe235ad68eddc0d47

I think the way to resolve the remaining discussion here is to make issues for them if the concerns still exist after #1002

danielabrozzoni (approved):
Partial ACK 75f8b81d58a985669ce7302fe235ad68eddc0d47 - the Wallet code looks good to me, I don't have a good enough understanding of the esplora/electrum code to confidently ACK it.

Issue #906, opened 2023-03-17
Tests for blockchain-extension crates.


evanlinjin:
Blockchain crates such as `bdk_esplora` and `bdk_electrum` should have comprehensive tests on their extension trait methods (outputs of these methods should be as expected given different scenarios).

Although we want to test the extension methods under the same set of scenarios, the syncing/scanning logic of each example is unique, so we want to check the outputs individually. Additionally, we may change the API of these extension methods so we do not want to deal with the hassle of maintaining additional generic/wrapper code.

We will provide the extension methods real blockchain data via the bitcoind and electrsd crates.

We will maintain a list of scenarios in this ticket description (I will update it via suggestions in the comments below).

### Scenarios

* Ensure `stop_gap` functions correctly. Does the methods correctly process a transaction history with self-sends right at the boundaries of the input `stop_gap`?
* If we provide the methods with "already-reorged" data, is the output as expected? I.e. Does the returned `SparseChain`/`ChainGraph` contain checkpoints at `agreement_point` and `invalid_from` indexes.

---

Original ticket: https://github.com/LLFourn/bdk_core_staging/issues/156

vladimirfomene:
@evanlinjin, I will be taking this up

notmandatory:
Current code test coverage is good, see current CodeCov stats below.

| crate | Line Coverage | Functions |
|-------|---------------|-----------|
| bitcoind_rpc | 81.6 % , 129 / 158 | 50.0 % , 15 / 30 |
| electrum	| 77.8 % , 273 / 351 | 45.5 % , 20 / 44 |
| esplora/src | 91.3 %  , 1023 / 1120 | 54.5 % ,  61 / 112 |

Closing this issue but feel free to open new issues on specific areas that need better testing.

PR #979, opened 2023-05-16
Test electrum ext

vladimirfomene:
### Description

This PR tests our Electrum scanning logic

### Notes to the reviewers

I will leave this in draft for now because the test still occasionally fail. I'm still trying to figure out why but the root cause of the bug is hard to find because the bug occurs occasionally.

#### All Submissions:

### DIFF
--- a/crates/electrum/Cargo.toml
+++ b/crates/electrum/Cargo.toml
@@ -14,3 +14,11 @@ readme = "README.md"
 [dependencies]
 bdk_chain = { path = "../chain", version = "0.4.0", features = ["serde", "miniscript"] }
 electrum-client = { version = "0.12" }
+
+[dev-dependencies]
+anyhow = "1"
+electrsd = { version= "0.22", features = ["bitcoind_22_0", "electrs_0_9_1"] }
+# zip versions after 0.6.3 don't work with our MSRV 1.57.0
+ zip = "=0.6.3"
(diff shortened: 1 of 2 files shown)


evanlinjin:
@vladimirfomene This doesn't seem to build for `1.57.0`. Could you figure out a way, to have the CI skip these tests for `1.57.0`?

Also, could you provide more details for the failing tests? Thank you.

vladimirfomene:
@evanlinjin, was able to get the test to compile with 1.57. Here are the two errors I occasionally get. 
two-test-failed

LLFourn:
We have a bug in #1145 that having tests like this PR introduced may have caught. I think we want to introduce a test in that PR to reproduce the issue. If this PR already reproduces the issue then we could use this one. In any case the groundwork here should be useful. I think this PR can be improved by not using concepts like `miniscript` or `KeychainTxOutIndex` since we are only trying to test the electrum logic (not `bdk_chain`). it's easy enough to call the APIs here by just putting some spks in a `Vec` or whatever. In the case of #1145's bug we just need to check the chain returned is correct.  

@vladimirfomene wdyt?

notmandatory:
Test coverage is good for electrum client see #906. We can add more testing for new scenarios as needed. Closing this PR as it's out of date.

PR #985, opened 2023-05-22
Fix caching

notmandatory:
### Description

The `Wallet.ensure_addresses_cached()` function was always regenerating and saving new scripts pubkeys starting from index 0.  This is inefficient and for very large wallets and can cause sync to fail. 

### Notes to the reviewers

This issue was discovered by @bodymindarts while testing on a 200K+ transaction wallet.

Since this is more of a performance issue than functional/scaling problem I only added a test to verify that with multiple cached batches created during sync, the transaction to addresses in each cached batch are found. An integration test syncing 200K+ tx would take too long to complete. 

### Changelog notice

* Fix script pubkey caching during sync to only cache starting from the current highest script index instead of 0.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/blockchain/compact_filters/peer.rs
+++ b/src/blockchain/compact_filters/peer.rs
@@ -74,7 +74,7 @@ impl Mempool {
 
     /// Look-up a transaction in the mempool given an [`Inventory`] request
     pub fn get_tx(&self, inventory: &Inventory) -> Option<Transaction> {
-        let identifer = match inventory {
+        let identifier = match inventory {
             Inventory::Error
             | Inventory::Block(_)
             | Inventory::WitnessBlock(_)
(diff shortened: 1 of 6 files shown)


notmandatory:
@bodymindarts would you be able to test this branch against your large test descriptor?

notmandatory:
I fixed a few spelling typos to trigger a new build. Not sure why github wouldn't run the actions pipeline initially.

notmandatory:
Looks like some tests in CI are failing, I'll work on fixing those.

bodymindarts:
Just for clarification. It has > 200k _addresses_. About half and half from the public and the change descriptor. I'm not sure how many transactions but same order-of-magnitude.

I will try it out as soon as possible!

notmandatory:
Removed from 0.29 milestone and closing this PR since issues are already fixed in 1.0 syncing mechanism.

PR #995, opened 2023-05-27
Add big wallet example

notmandatory:
### Description

Adding an example that creates a wallet that uses up to address index 200K and times how long it takes to sync. I also changed the esplora concurrency value to be a usize since there shouldn't be any reason it can't use more then 256 connections. 

### Changelog notice

- Change EsploraConfig concurrency value from a u8 to a usize
- Add new big_wallet_esplora example to test a wallet that monitor up to 200K addresses

#### All Submissions:


nondiremanuel:
Is this part of 1.0 version or should I assign it to maintenance?

notmandatory:
This particular PR was build for the release/0.28 branch so needs to be in the maintenance branch. I'll make an issue to do something similar for the 1.0 release.

notmandatory:
Big wallets (thousands of addresses to track) should use the new `bitcoind_rpc` client, or the future CBF (BIP157/158) client instead of esplora.

PR #1002, opened 2023-06-03
Implement linked-list `LocalChain` and add rpc-chain module/example

### LINKED ISSUE #997: LocalChain should be implemented as a cheaply cloneable singly linked list
See bad ergonomics here:

https://github.com/bitcoindevkit/bdk/blob/965cbe6dde99a1a58168cc0f4bdd320f0aba10e1/example-crates/example_electrum/src/main.rs#L145-L152

We should just be able to cheaply clone the whole thing (without copying all the data using referencing counting and linked list). From this you should be able to get an owned iterator of `BlockId`.

evanlinjin:
Fixes bitcoindevkit/bdk#997 

### Description

This PR changes the `LocalChain` implementation to have blocks stored as a linked-list. To verify the API of the linked-list `LocalChain`, the `bitcoind_rpc` crate and example is added (which is a block-by-block chain source). Additionally, the electrum/esplora crates and examples are also updated for the new API, but the logic is also updated to better-ensure consistency when obtaining a view of blocks.

A linked-list `LocalChain` allows the data-source thread to hold a shared reference to a single checkpoint and have access to the whole history of checkpoints without cloning or having a lock on `LocalChain`. This idea was originally proposed by @LLFourn's.

### Notes to the reviewers

Sorry this is a massive PR again. The changes can be "grouped" as follows:
* Change `LocalChain` API to use linked-list.
  * The `apply_update` method is renamed to `update` and the input is now the `CheckPoint` tip. An additional input is added (`introduce_older_blocks: bool`) which allows the caller to specify whether they wish to introduce new checkpoints below the highest point of agreement. This is required optimization as the update tip's chain and the original chain may share a large portion of history. 
  * A `_check_consistency` method is added, which `assert!`s whether the internal tip's history is consistent with the internal index.
* Introduce `prune_and_apply_update` methods for both `IndexedTxGraph` and `Wallet`. This prunes out irrelevant transactions from the `TxGraph` update before applying it. This will be used by block-by-block data sources as each block update is a `TxGraph`.
* Update `electrum` crate to use the new `LocalChain` API. We cannot 100% guarantee consistency with `electrum` updates (yet), but this is an improvement.
* Update `esplora` to ensure 100% consistency. `LocalChain` and `TxGraph` updates are now handled separately in `ExploraExt`. We make use of the `block_hash` field of tx statuses from the esplora API.
* Introduce `bitcoind_rpc` crate and example.

### Changelog notice

* Change `LocalChain` API to use a linked-list of block data.
* Add the `bitcoind_rpc` crate and example.
* Fix the `esplora` crate to guarantee a consistent view of blocks when formulating an update.
* Add `example_rpc` CLI example for `bitcoind_rpc` crate.
* Fix: Remove duplicate temp-plan crate.

#### All Submissions:

#### New Features:

#### Bugfixes:


crates/chain/src/local_chain.rs
```diff
@@ -1,14 +1,116 @@
...
+/// A structure that represents changes to [`LocalChain`].
+pub type ChangeSet = BTreeMap<u32, Option<BlockHash>>;
+
+/// Represents a block of [`LocalChain`].
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct CheckPoint {
+    /// Block height.
+    pub height: u32,
+    /// Block hash.
+    pub hash: BlockHash,
```
LLFourn: Why not just hacve `block_id` here instead of these two fields.

crates/chain/src/local_chain.rs
```diff
@@ -1,14 +1,116 @@
...
+    }
+}
+
+/// A structure that iterates over checkpoints backwards.
+pub struct CheckPointIter {
+    current: Option<Arc<CheckPoint>>,
+}
+
+impl Iterator for CheckPointIter {
+    type Item = Arc<CheckPoint>;
```
LLFourn: Why doesn't this just iterate `BlockId`.

crates/chain/src/local_chain.rs
```diff
@@ -17,224 +119,450 @@ impl ChainOracle for LocalChain {
...
+    ///
+    /// An error will occur if the update does not correctly connect with `self`.
+    ///
+    /// Refer to [module-level documentation] for more.
+    ///
+    /// [module-level documentation]: crate::local_chain
+    pub fn determine_changeset(
+        &self,
+        new_tip: &Arc<CheckPoint>,
+        must_include_height: Option<u32>,
```
LLFourn: why isn't the update another `LocalChain`?

crates/chain/src/local_chain.rs
```diff
@@ -1,14 +1,116 @@
...
+use alloc::{sync::Arc, vec::Vec};
 use bitcoin::BlockHash;
 
-use crate::{BlockId, ChainOracle};
+/// A structure that represents changes to [`LocalChain`].
+pub type ChangeSet = BTreeMap<u32, Option<BlockHash>>;
+
+/// Represents a block of [`LocalChain`].
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct CheckPoint {
```
LLFourn: This type should be private in my mind. It's an implementation detail for how you can produce an iterator for `BlockId` with a single `Arc::clone`.

crates/chain/src/local_chain.rs
```diff
@@ -1,14 +1,116 @@
...
+}
 
 /// This is a local implementation of [`ChainOracle`].
 #[derive(Debug, Default, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct LocalChain {
-    blocks: BTreeMap<u32, BlockHash>,
+    checkpoints: BTreeMap<u32, Arc<CheckPoint>>,
+}
+
+impl FromIterator<Arc<CheckPoint>> for LocalChain {
```
LLFourn: Not sure what this is for. You can create a `LocalChain` from a single `&CheckPoint` by going backwards. Why do you want to create a localchain from an iterator of `CheckPoint`. Also since I think `CheckPoint` should be private I also think this probably shouldn't exist.

crates/chain/src/local_chain.rs
```diff
@@ -1,14 +1,116 @@
...
+    fn from_iter<T: IntoIterator<Item = Arc<CheckPoint>>>(iter: T) -> Self {
+        let mut chain = Self {
+            checkpoints: iter.into_iter().map(|cp| (cp.height, cp)).collect(),
+        };
+        chain.fix_links(0);
+        chain
+    }
+}
+
+impl IntoIterator for LocalChain {
```
LLFourn: I guess we don't want this because it consumes `LocalChain` so is not very useful given how we expect to use it.

crates/chain/src/local_chain.rs
```diff
@@ -17,224 +119,450 @@ impl ChainOracle for LocalChain {
...
     }
 
-    /// Get a reference to a map of block height to hash.
-    pub fn blocks(&self) -> &BTreeMap<u32, BlockHash> {
-        &self.blocks
+    /// Constructs a [`LocalChain`] from an iterator of [`BlockId`]s.
+    ///
+    /// The blocks do not need to be in order. If multiple blocks of the same height are introduced,
+    /// the latest block will be used.
+    pub fn from_blocks(blocks: impl IntoIterator<Item = BlockId>) -> Self {
```
LLFourn: Not sure about this API. Can't we just force them to pass in a `BTreeSet<BlockId>` so it's ordered properly and so we can set `prev` in a natural way.

The docs must say that it is up to the caller to make sure the blocks are in the same chain.

LLFourn (commented):
Initial comments. The main difference I have is that the user should not see `Arc` anywhere. It's an internal implementation detail.

crates/chain/src/local_chain.rs
evanlinjin: No specific reason, just that it seems more convenient this way if we are exposing `CheckPoint`. I.e. you can get blockhash by `cp.hash` instead of `cp.block.hash`.

crates/chain/src/local_chain.rs
evanlinjin: My thinking is that `CheckPoint` just contains more data, so why not just expose that?

crates/chain/src/local_chain.rs
evanlinjin: Formulating a `LocalChain` means we need to think about which checkpoints to include in the update. I think it's a much easier API if the chain-data-source-thread just spits out new checkpoints it is aware of, and the `determine_changeset`. Of course, you can continue to add data to an update `LocalChain` and throw that down a channel, but the internal `BTreeMap` will keep growing.

crates/chain/src/local_chain.rs
evanlinjin: What are the downsides of exposing `Arc`?

crates/chain/src/local_chain.rs
LLFourn: Ok I think that makes sense. so what is the point of `must_include_height`? The doc comments don't mention it.

crates/chain/src/local_chain.rs
evanlinjin: Changed.

crates/chain/src/local_chain.rs
evanlinjin: Removed `must_include_height`. We break early if the internal raw pointers match. Otherwise we iterate over the entire ancestry of `new_tip`.

crates/chain/src/local_chain.rs
evanlinjin: @LLFourn and myself discussed about this face-to-face and we agree that iterating over checkpoints is a useful feature to have since we can compare raw pointers, making updates more efficient. However, `Arc` should be an internal implementation detail.

These changes have all been implemented.

crates/chain/src/local_chain.rs
evanlinjin: `Arc` is now hidden.

crates/chain/src/local_chain.rs
evanlinjin: Removed.

crates/chain/src/local_chain.rs
evanlinjin: Removed.

crates/chain/src/local_chain.rs
evanlinjin: Changed as suggested.

crates/chain/src/local_chain.rs
```diff
@@ -1,14 +1,149 @@
...
+impl std::error::Error for NewCheckPointError {}
+
+impl CheckPoint {
+    /// Construct a [`CheckPoint`] from a [`BlockId`].
+    pub fn new(block: BlockId) -> Self {
+        Self(Arc::new(CPInner { block, prev: None }))
+    }
+
+    /// Construct a [`CheckPoint`] of `block` with a previous checkpoint.
+    pub fn new_with_prev(
```
LLFourn: Do we need this as a public API?

What about if you can just create a `CheckPoint` from a `BTreeSet` of `BlockId` which are assumed to be in the same chain. This would remove the error case that comes with this method.

crates/chain/src/local_chain.rs
```diff
@@ -17,224 +152,339 @@ impl ChainOracle for LocalChain {
...
+        apply();
         Ok(changeset)
     }
 
+    /// Get or insert a `block_id`.
+    ///
+    /// # Errors
+    ///
+    /// Replacing the block hash of an existing checkpoint will result in an error.
+    pub fn get_or_insert(
```
LLFourn: What's wrong with `get_block_hash` and `insert_block`. Why does this API need to exist?

crates/chain/src/local_chain.rs
```diff
@@ -17,224 +152,339 @@ impl ChainOracle for LocalChain {
...
-            .last()
-            .map(|(&height, &hash)| BlockId { height, hash })
+    /// Returns whether the [`LocalChain`] is empty (has no checkpoints).
+    pub fn is_empty(&self) -> bool {
+        self.checkpoints.is_empty()
     }
 
-    /// This is like the sparsechain's logic, expect we must guarantee that all invalidated heights
-    /// are to be re-filled.
-    pub fn determine_changeset(&self, update: &Self) -> Result<ChangeSet, UpdateNotConnectedError> {
```
LLFourn, line 89: Don't you still need this? You want to `determine_changeset`, apply to persistent storage and then update in memory if it succeeds. In order to only have `update` we'd have to make `LocalChain` monotone. We could do this after this PR pretty easily I think and so leave it as `update` if we promise to fix it later.

crates/chain/src/local_chain.rs
evanlinjin: Ideally we want to create with a previous pointer. This way updates from the data source can connect in memory.

crates/chain/src/local_chain.rs
evanlinjin: `update` returns a closure first. The local chain only updates when the returned closure is called. It is done in this way so we can make use of the pointers during the update.

crates/chain/src/local_chain.rs
LLFourn: What about a method on checkpoint like extend which creates a new one on top of an existing one?

crates/chain/src/local_chain.rs
LLFourn: I don't really like this. I can just call `apply_changeset` instead of calling this closure to do the same thing?. It'd be fine with just applying the changeset right away and trying to fix #1005 eventually.

crates/chain/src/local_chain.rs
evanlinjin: Heights may not increment in sequence, so don't we still need an error? We can also have a Result, where the error variant is the old checkpoint (meaning failed to extend).

crates/chain/src/local_chain.rs
evanlinjin: Yes, but you won't get some of the benefits of the linked list. What do you not like about this API?

crates/esplora/src/blocking_ext.rs
```diff
@@ -66,60 +62,75 @@ pub trait EsploraExt {
...
+        for (&height, &hash) in new_blocks.iter() {
+            last_cp = Some(
+                CheckPoint::new_with_prev(BlockId { height, hash }, last_cp)
+                    .expect("heights should not conflict"),
+            );
         }
 
+        let tip = last_cp.expect("must have atleast one checkpoint");
+        let mut update = LocalUpdate::<K, ConfirmationTimeAnchor>::new(tip.clone());
+
```
LLFourn, line 202: I have a feeling that this could be a bit simpler. The algorithm is.

1. Record tip
2. Get relevant transactions anchored at the tip
3. Check tip hasn't changed -- if not extend whichever previous tip still exists in chain with blocks that conflict with those that aren't and finally the new tip.

So I'd put most of this logic at the end. I feel like there might be a more concise way of expressing it.

evanlinjin: Do you mind being a bit more specific about the part you are talking about? Thanks

LLFourn: The logic mentioned in point (3). Right now that happens at the record tip step rather than at the end.

crates/chain/src/local_chain.rs
LLFourn: It's redundant which makes it confusing. Is the point I might decide not to apply it? In which situation would I not want to apply it?

evanlinjin: Ah I see what you mean, I'll remove the closure and have the method apply changes right away.

crates/chain/src/local_chain.rs
LLFourn: Yes sorry this wouldn't remove the error case but just make it a more natural API (pushing onto the end of a list).

notmandatory:
A little off-topic for this PR, but how'd you'all feel about adding type aliases for the block height and unix time instead of just using a `u32` and `u64`? It would make the code easier to read and I don't think there are any drawbacks. Could put them in `chain/src/lib.rs`.

```rust
// The bitcoin block height, 0 is the genesis block.
pub type BlockHeight = u32;

// Number of seconds that have elapsed since 00:00:00 UTC on 1 January 1970, the Unix epoch.
pub type UnixSeconds = u64;
```

evanlinjin:
@notmandatory Thank you for the suggestion! Personally, the aliases don't make the code easier to read for me. However, I am open for the change if enough people see the benefit.

notmandatory:
I did a little local regtest testing with the `rpc_example` CLI and created a README with my steps: https://github.com/bitcoindevkit/bdk/pull/1019/commits/4d365919d5c797e43040fc20f83f50133a4691c2

@evanlinjin do you think this is useful? If so feel free to pull the README file into your PR or I can add it later in a new PR. For the record I didn't find any problems in this basic manual testing.

crates/esplora/src/blocking_ext.rs
evanlinjin: I've added a separate method for constructing the "update tip" (named `construct_update_tip`). It's probably not what you had in mind @LLFourn, but I think it's the only way of guaranteeing that we create a consistent view of checkpoints during the update.

The logic of finding the point of agreement can screw up consistency, if there is a double-reorg during the update. Please refer to the diagram.

photo_2023-06-29_13-07-37

Essentially, we need to atomically fetch the latest x blocks from electrum/esplora to guarantee consistency. x is the assume-final depth.

crates/chain/src/local_chain.rs
evanlinjin: Reverted back to `insert_block`!

crates/chain/src/local_chain.rs
evanlinjin: Changed the API to what you suggested @LLFourn

crates/chain/src/local_chain.rs
evanlinjin: Done.

crates/chain/src/local_chain.rs
```diff
@@ -19,215 +130,295 @@ impl ChainOracle for LocalChain {
...
+    /// [module-level documentation]: crate::local_chain
+    pub fn update(&mut self, new_tip: CheckPoint) -> Result<ChangeSet, CannotConnectError> {
+        let mut updated_cps = BTreeMap::<u32, CheckPoint>::new();
+        let mut agreement_height = Option::<u32>::None;
+        let mut agreement_ptr_matches = false;
+
+        for cp in new_tip.iter() {
+            let block = cp.block_id();
+
+            match self.checkpoints.get(&block.height) {
```
LLFourn: I think this algorithm does not require us to use `BTreeMap`s at all. We are just trying to figure out whether and where two linked lists connect to each other with our given chain rules. This can be done by traversing the linked lists backwards in tandem. The `BTreeMap` should just be there to index the linked list items by height once we've figured out our new chain.

I don't mind implementing it like this if you don't see my vision here.

LLFourn (commented):
Comments at this stage:

1. The test look great. They need to also test what happens if you build the update chain from the existing chain local chain. This could be done as new tests and/or having a generic transformation on the existing tests to make one of the update chain nodes be cloned form the same `Arc` as the local chain.
2. I think the `BtreeMap` in `LocalChain` should just index a `CheckPoint` chain. It should not have `CheckPoint`s in it. Will explain more in code comment.
3. There are too many named loops in the bitcoin rpc code. Try to rewrite it so that loop labels are not used as GOTOs.
4. The bitcoin RPC code is using iterators in a kind of unconventional way where you can return `Some(_)` after `None`. I think the idea is cute and works but needs some docs to explain in which situation you would use it as an iterator and which you would manually call `.next` on it.

crates/chain/src/local_chain.rs
evanlinjin: @LLFourn today I tried implementing it the way as you suggested. However, the algorithm to traverse the linked lists in tandem is difficult (I will not claim that it is impossible).

One of the reasons that makes this a difficult problem is that our update logic requires a specific rule which states:

> The update chain must have a checkpoint that overlaps the original chain's checkpoint which is one higher than the highest point of agreement.

However, depending on where we have our points of agreements (or lack of), we will need to have to include/exclude specific ranges of blocks from the original and update chains, while ensuring we don't unnecessarily remove links for the final resultant chain.

Essentially, it is much easier to not care about links until the final step (hence the `fix_links` method).

Would you be okay to keep this as is and we fix `LocalChain` for good in #1005?

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// [`BitcoindRpcIter`] attempts to avoid re-emitting transactions.
+    Mempool {
+        /// The checkpoint of the last-seen tip.
+        cp: CheckPoint,
+        /// Subset of mempool transactions.
+        txs: Vec<(Transaction, u64)>,
+    },
+}
+
+/// A closure that transforms a [`BitcoindRpcItem`] into a [`ConfirmationHeightAnchor`].
```
vladimirfomene: Update `BitcoindRpcItem` to `BitcoindRpcUpdate`.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+    ConfirmationHeightAnchor {
+        anchor_block: BlockId {
+            height: info.height as _,
+            hash: info.hash,
+        },
+        confirmation_height: info.height as _,
+    }
+}
+
+/// A closure that transforms a [`BitcoindRpcItem`] into a [`ConfirmationTimeAnchor`].
```
vladimirfomene: Also update here

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+    pub fn is_mempool(&self) -> bool {
+        matches!(self, Self::Mempool { .. })
+    }
+
+    /// Returns whether the update is of a block.
+    pub fn is_block(&self) -> bool {
+        matches!(self, Self::Block { .. })
+    }
+
+    /// Transforms the [`BitcoindRpcItem`] into a [`LocalUpdate`].
```
vladimirfomene: Also update `BitcoindRpcItem`

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+            self.last_emission_was_mempool = false;
+            None
+        } else {
+            Some(self.next_update())
+        }
+    }
+}
+
+impl<'a> BitcoindRpcEmitter<'a> {
+    /// Constructs a new [`BitcoindRpcIter`] with the provided [`bitcoincore_rpc::Client`].
```
vladimirfomene: Update `BitcoindRpcIter` to `BitcoindRpcEmitter`

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -90,6 +90,38 @@ where
         }
     }
 
+    /// Apply `update`, but filters out irrelevant transactions.
```
vladimirfomene, line 93: Should pruning irrelevant transactions not be something we do by default (maybe in `apply_update`)? I'm thinking we will save space in doing so.

crates/chain/src/local_chain.rs
```diff
@@ -19,215 +144,229 @@ impl ChainOracle for LocalChain {
     fn is_block_in_chain(
         &self,
         block: BlockId,
-        static_block: BlockId,
+        chain_tip: BlockId,
```
vladimirfomene, line 167: Sorry if this is was tackled in other PRs. Why don't we just return `Result<bool, Self::Error>`? Is the error here infallible?

crates/chain/src/local_chain.rs
```diff
@@ -247,4 +386,185 @@ impl core::fmt::Display for InsertBlockNotMatchingError {
...
+impl std::error::Error for CannotConnectError {}
+
+fn merge(
+    original_tip: CheckPoint,
+    update_tip: CheckPoint,
+) -> Result<(CheckPoint, ChangeSet), CannotConnectError> {
+    let mut o_visited = BTreeMap::<u32, CheckPoint>::new();
+    let mut o_iter = original_tip
+        .iter()
+        .inspect(|cp| {
```
vladimirfomene: Does this not make more sense as a `for_each`?

crates/chain/src/local_chain.rs
```diff
@@ -247,4 +386,185 @@ impl core::fmt::Display for InsertBlockNotMatchingError {
...
+            o_visited.insert(cp.height(), cp.clone());
+        })
+        .peekable();
+
+    let mut u_visited = BTreeMap::<u32, CheckPoint>::new();
+    let u_iter = update_tip.iter().inspect(|cp| {
```
vladimirfomene: Same here

crates/chain/src/local_chain.rs
```diff
@@ -247,4 +386,185 @@ impl core::fmt::Display for InsertBlockNotMatchingError {
...
+            "introduced chain cannot connect with the original chain, try include height {}",
+            self.try_include_height,
+        )
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for CannotConnectError {}
+
+fn merge(
```
vladimirfomene: Please can you add more comments to this function, I'm a bit confused on its inner workings.

vladimirfomene (commented):
My first pass of review. Couldn't finish to the end but will definitely go through it again maybe over the weekend.

crates/chain/src/local_chain.rs
evanlinjin: No, as `for_each` consumes the entire iterator.

crates/chain/src/local_chain.rs
evanlinjin: https://github.com/bitcoindevkit/bdk/pull/1002/commits/7223adb9f801544fa24180fdabc8f6b159da8c31#r1257138658

crates/chain/src/local_chain.rs
evanlinjin: Sorry about that. I've converted this PR back to a draft so people don't get confused. Note that @LLFourn has already rewritten this method with good comments.

crates/chain/src/local_chain.rs
evanlinjin: @LLFourn has been persistent to have this change (and for good reason) since it has simplified our implementation (see the `local_chain::merge_chains` method). These changes are included in this PR.

crates/chain/src/local_chain.rs
vladimirfomene: thanks @evanlinjin !

crates/chain/src/local_chain.rs
```diff
@@ -2,15 +2,137 @@
...
+    ///
+    /// Returns an `Err(self)` if there is block which does not have a greater height than the
+    /// previous one.
+    pub fn extend(self, blocks: impl IntoIterator<Item = BlockId>) -> Result<Self, Self> {
+        let mut curr = self.clone();
+        for block in blocks {
+            curr = curr.push(block).map_err(|_| self.clone())?;
+        }
+        Ok(curr)
+    }
```
evanlinjin: There is already a trait called `Extend` with a method with the exact name and similar name: https://doc.rust-lang.org/std/iter/trait.Extend.html

I think we should rename this to `extend_chain` or something else.

crates/chain/src/local_chain.rs
evanlinjin: The error here is `Infallible` (as defined 5 lines above). The signature of the trait method `ChainOracle::is_block_in_chain` returns an option wrapped in a result. `Ok(None)` basically means "I don't know". The plan is to have additional logic to handle this specific scenario (not implemented).

crates/esplora/src/blocking_ext.rs
evanlinjin: Fixed and redesigned in a47837c0a80d9d860b4c5774a978205ca82d7e64

crates/chain/src/indexed_tx_graph.rs
evanlinjin: For spk-based syncing (such as esplora and electrum) there is nothing to prune. If we enforce pruning, it will interfere with the ability of adding transactions for feerate calculations (for example, a payjoin send).

Via @LLFourn's advice, I've currently split this into two methods: 68b3c5841477d5318e1b144db9d6927a0bf400d4

crates/chain/src/local_chain.rs
evanlinjin: Renamed to `extend_with_blocks` in 4f327134815cd52789cf13a595927e1eaf1cbfc0

LLFourn: nit I prefer `extend`. iirc `CheckPoint` doesn't implement `Iterator` so this is not really a concern. There are plenty of things in `std` library that have a method `extend` that are not the trait. If you are renaming this then you should rename `push` similarly as well.

evanlinjin:
After a call with @LLFourn, I will make the following changes to this PR before splitting it up:

- [x] `LocalChain::update` should take in a new struct (`local_chain::Update`) that combines the `CheckPoint` input with the `introduce_older_blocks: bool`. This is because the boolean argument is determined by the chain data source, and not by the caller/user.
- [x] Don't prematurely optimize. Remove the same-pointer optimization for `local_chain::merge_chains`.
- [x] Remove `derive(Eq, PartialEq, PartialOrd, Ord)` for `CheckPoint`. We don't know what this means.

Because our PR(s) still need to compile, we will split them as follows:

1. Introduce linked-list `LocalChain` and update esplora/electrum/wallet.
2. Introduce bitcoind rpc crate & example. We will need to add `prune_and_apply_update` for `IndexedTxGraph` and `Wallet` (potentially).
3. Add esplora cli example.

This is the best we can do to ensure that our PRs still compiles at every commit. After the PR-split, I will explore the following:

- [ ] `BlockId` can impl `Anchor` but the semantics should be documented. I.e. it means the tx is confirmed in the anchor block (same block).
- [ ] `BitcoindRpcEmitter` should have separate methods for emitting blocks and mempool.
- [ ] `BitcoindUpdate::into_update`'s input closure should be `anchor_filter_map` which returns an `Option<A>`.

evanlinjin:
Replaced by bitcoindevkit/bdk#1034

PR #1004, opened 2023-06-05
Add TxBuilder step 1: CoinControl

danielabrozzoni:
#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -43,6 +43,7 @@ pub mod coin_selection;
 pub mod export;
 pub mod signer;
 pub mod tx_builder;
+pub mod tx_builder_2;
 pub(crate) mod utils;
 
 #[cfg(feature = "hardware-signer")]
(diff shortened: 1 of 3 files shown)


danielabrozzoni:
Replying to @evanlinjin's comment https://gist.github.com/evanlinjin/41e6a23cc08e888884bded327e290866?permalink_comment_id=4587231#gistcomment-4587231 here:

> - Let's make a PR soon! It's easier to discuss that way (like on the PR).
> - Let's disallow duplicate UTXOs. Need some sort of filter HashSet or BTreeSet.
> - It will be better to use utxos: impl IntoIterator so the caller won't need to .collect first if their UTXOs is not a Vec.

ACK 

> - add_planned_utxos() essentially forces the caller to group UTXOs using the same descriptor + assets together first. Would this be a problem?

I don't think it is, especially since I have a new `list_unspent_by_keychain` method in the wallet. But if you see any better method to do this, lmk! 

> Thoughts on spec-writing in general

Agreed, I'll try to apply those heuristics to the txbuilder docs, thanks!

PR #1007, opened 2023-06-16
Ignore rocksdb audit error RUSTSEC-2022-0046

### LINKED ISSUE #1006: release/0.28 audit failing due to RUSTSEC-2022-0046.html
**Describe the bug**  

For some reason the CI audit just caught that there is an issue with `rocksdb` due to "Out-of-bounds read when opening multiple column families with TTL".  See:

https://rustsec.org/advisories/RUSTSEC-2022-0046.html

**To Reproduce**  

```
cargo update
cargo audit
```

**Expected behavior**  

Should not fail audit tests.

**Build environment**  
 - BDK tag/commit: release/0.28 
 - OS+version: any
 - Rust/Cargo version: ci tests against 1.57.0 and 1.65.0
 - Rust/Cargo target: any 

**Additional context**  

The `rocksdb` dependency is only used when the `compact_filters` feature is enabled. The reason we haven't update our `rocksdb` version earlier is that newer versions don't support our MSRV of 1.57. See also #480. 

I propose we disable the audit warning for this issue since `compact_filters` is an experimental feature and we don't use the rocksdb "multiple column families with TTL" feature mentioned in this advisory.

notmandatory:
### Description

Fixes #1006.

### Notes to the reviewers

The `compact_filters` feature is marked as **experimental** and we don't use the rocksdb "multiple column families with TTL" feature mentioned in this advisory. Also this feature will be completely reworked for the bdk 1.0 release.

### Changelog notice

None

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,2 @@
+[advisories]
+ignore = ["RUSTSEC-2022-0046"]


junderw:
Concept ACK

I verified the vulnerability and the fact that the current usage wouldn't trigger it anyways.

Being an experimental feature compounds upon the reasons for the ACK.

danielabrozzoni (approved):
ACK f7d0852e92640791711cd1cd580a0b972df3da64 - bdk is not affected by this vulnerability, as we never call `rocksdb::DBWithThreadMode::open_cf_descriptors_with_ttl()`

PR #1011, opened 2023-06-20
[ci] fix docsrs error for bdk crate

### LINKED ISSUE #955: docs don't build on docs.rs for 1.0.0-alpha.0
See: https://docs.rs/crate/bdk/1.0.0-alpha.0

can we somehow make sure in CI that the docs will build on docs.rs before merging?

notmandatory:
### Description

Fixes #955

### Notes to the reviewers

- Fixed by adding `#![cfg_attr(docsrs, feature(doc_cfg))]` to `bdk` crate lib.rs.
- ~~Update nightly_docs workflow to use similar build command that docs.rs uses (and fails if above not done).~~
- ~~For example docs artifact produced see: https://github.com/bitcoindevkit/bdk/actions/runs/5326442890~~

### Changelog notice

none

#### All Submissions:

### DIFF
--- a/crates/bdk/src/lib.rs
+++ b/crates/bdk/src/lib.rs
@@ -1,4 +1,10 @@
 #![doc = include_str!("../README.md")]
+// only enables the `doc_cfg` feature when the `docsrs` configuration attribute is defined
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(
+    docsrs,
+    doc(html_logo_url = "https://github.com/bitcoindevkit/bdk/raw/master/static/bdk.png")
+)]
 #![no_std]
 #![warn(missing_docs)]
 


thunderbiscuit (approved):
ACK 789cb997bc531243c9b34cbb29415c807b698d52.

crates/bdk/Cargo.toml
```diff
@@ -62,8 +62,10 @@ rustdoc-args = ["--cfg", "docsrs"]
 name = "mnemonic_to_descriptors"
 path = "examples/mnemonic_to_descriptors.rs"
 required-features = ["all-keys"]
+doc-scrape-examples = true
```
LLFourn: I've never heard of this feature before. Interested to see how this works. Unfortunately our new examples are in their own crates so this won't work there.

.github/workflows/nightly_docs.yml
```diff
@@ -17,10 +25,9 @@ jobs:
         run: rustup update
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2.2.1
-      - name: Build docs
-        run: cargo doc --no-deps
-        env:
-          RUSTDOCFLAGS: '--cfg docsrs -Dwarnings'
```
LLFourn: shouldn't you keep `-Dwarnings`.

crates/bdk/Cargo.toml
notmandatory: At least the `bdk` crate still has some code in it's own `examples`, and in the future it would be nice to put some simple examples in all the crates just to get the code added to the docs.

.github/workflows/nightly_docs.yml
notmandatory: The `-Dwarnings` isn't in the command that docs.rs uses so I left it off. I did confirm that with the new command string I added catches the `feature(doc_cfg)` issue when run locally or remotely. Also I can't see where in the docs that `-Dwarnings` is a flag that rustdoc cares about.
https://doc.rust-lang.org/cargo/commands/cargo-rustdoc.html

notmandatory: Correction: it looks like `--cfg docsrs` and `-Dwarnings` are both flags passed through to `rustc`.  I'm pretty sure other jobs like `clippy` will fail first, but doesn't hurt to add it here also. I added it back as part of the `build.rustdocflags` and force pushed.

notmandatory:
Ugh. Now with the `-Dwarnings` I seem to be getting a new error that I can't repro running the same command as on CI locally. Will keep looking into it tomorrow.

crates/bdk/Cargo.toml
notmandatory: Nevermind, I removed the `doc-scrape-examples` feature, can add back later if we want it.

.github/workflows/nightly_docs.yml
notmandatory: I rolled back all changes except the one I think really mattered which was adding `#![cfg_attr(docsrs, feature(doc_cfg))]` to bdk crate lib.rs.  

If that's not enough to fix docs.rs I'll take another stab at it for the next alpha.

notmandatory:
@LLFourn I simplified this down to just what should be needed to fix the original issue that was breaking the docs.rs rustdocs. Please approve if OK with you, this is the last thing before we can cut the alpha.1 release.

LLFourn (approved):
LGTM thanks @notmandatory

PR #1012, opened 2023-06-21
Bump env_logger to 0.10.

monaka:
### Notes to the reviewers

This will reduce using `atty` that is unmainted.
See also: https://rustsec.org/advisories/RUSTSEC-2021-0145

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bdk/Cargo.toml
+++ b/crates/bdk/Cargo.toml
@@ -45,7 +45,7 @@ dev-getrandom-wasm = ["getrandom/js"]
 
 [dev-dependencies]
 lazy_static = "1.4"
-env_logger = "0.7"
+env_logger = "0.10"
 # Move back to importing from rust-bitcoin once https://github.com/rust-bitcoin/rust-bitcoin/pull/1342 is released
 base64 = "^0.13"
 assert_matches = "1.5.0"


tnull:
Unfortunately, `env_logger` 0.10 seems to require rustc 1.60 which doesn't meet MSRV policy. Up to 0.9.3, `env_logger` only required rustc 1.41.0, which may therefore be an option. However, this version is still dependent on `atty`, so at least for the given reason it doesn't make much sense to bump it to 0.9.3 either.

evanlinjin:
Thank you for this PR, unfortunately it doesn't seem possible due to MSRV requirements (thanks to @tnull for pointing this out). Closing for now.

apoelstra:
If the dep can really be bumped with no changes to the codebase, it may be possible to provide a ranged version that covers multiple major versions.

notmandatory:
@evanlinjin we should be able to bump `env_logger` to `0.10` and just `cargo update --precise` it back to `0.7` in CI for verifying build/tests with our MSRV (1.57), as in #1046.

evanlinjin:
> @evanlinjin we should be able to bump `env_logger` to `0.10` and just `cargo update --precise` it back to `0.7` in CI for verifying build/tests with our MSRV (1.57), as in #1046.

If we set `env_logger = "0.10"`, version `0.7` would not be compatible: https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility

We need to see whether setting a range (from `0.7` to `0.10`) does not break builds at any version of `env_logger` (as @apoelstra suggested).

@monaka Sorry I closed the PR too early. Seems like there is a lot more options to explore. I'll create a ticket.

Edit: Ticket created (#1067).

PR #1014, opened 2023-06-22
Impl Ord for Condition

benthecarman:
This was missed in #894 

Tried using `master` for Mutiny and we get compile errors without having this

### DIFF
--- a/crates/bdk/src/descriptor/policy.rs
+++ b/crates/bdk/src/descriptor/policy.rs
@@ -39,7 +39,7 @@
 use crate::collections::{BTreeMap, HashSet, VecDeque};
 use alloc::string::String;
 use alloc::vec::Vec;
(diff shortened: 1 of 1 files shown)


notmandatory:
Clippy is complaining because you implemented `Ord` but using the default derived version of `PartialOrd`.  Where in your code do you need `Ord`? If it's for sorting something can you pass in your own function there instead of making this the global `Ord` impl for bdk?

benthecarman:
This is just so you can compile bdk with no-std and wasm32

can test with `cargo check --no-default-features --features bitcoin/no-std,miniscript/no-std,bdk_chain/hashbrown --target wasm32-unknown-unknown`

benthecarman:
Figured out this issue, need to include bdk_chain with the `hashbrown` feature

PR #1017, opened 2023-06-26
Fix cargo features

evanlinjin:
### Description

The following commands failed on `master`:
```
$ cargo build -p bdk_esplora
$ cargo build -p bdk_electrum
```

This PR introduces the following fixes:
* `bdk_chain/std` should also enable `miniscript/std`
* use the version of `hashbrown` that `bitcoin` and `miniscript` uses

### Changelog notice

* Change the version of `hashbrown` to `0.11` (the version that `bitcoin` and `miniscript` uses).
* Fix the `std` feature flags for `bdk_chain`.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/Cargo.toml
+++ b/crates/chain/Cargo.toml
@@ -13,12 +13,13 @@ readme = "README.md"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+# For no-std, remember to enable the bitcoin/no-std feature
 bitcoin = { version = "0.29", default-features = false }
 serde_crate = { package = "serde", version = "1", optional = true, features = ["derive"] }
 
 # Use hashbrown as a feature flag to have HashSet and HashMap from it.
 # note version 0.13 breaks outs MSRV.
-hashbrown = { version = "0.12", optional = true, features = ["serde"] }
+hashbrown = { version = "0.11", optional = true, features = ["serde"] }
 miniscript = { version = "9.0.0", optional = true, default-features = false }
 
 [dev-dependencies]
@@ -26,5 +27,5 @@ rand = "0.8"
 
 [features]
 default = ["std"]
-std = ["bitcoin/std"]
-serde = ["serde_crate", "bitcoin/serde" ]
+std = ["bitcoin/std", "miniscript/std"]
+serde = ["serde_crate", "bitcoin/serde"]


crates/chain/Cargo.toml
```diff
@@ -13,18 +13,19 @@ readme = "README.md"
...
 
 [dev-dependencies]
 rand = "0.8"
 
 [features]
 default = ["std"]
-std = ["bitcoin/std"]
-serde = ["serde_crate", "bitcoin/serde" ]
+std = ["bitcoin/std", "miniscript/std"]
+serde = ["serde_crate", "bitcoin/serde", "hashbrown/serde"]
```
notmandatory: If we add `"hashbrown/serde"` to the `serde` feature then whenever `serde` is enabled `hashbrown` will be too. Since `hashbrown` is meant primarily for `no-std` builds how about this instead:

```suggestion
serde = ["serde_crate", "bitcoin/serde"]
hashbrown_serde = ["hashbrown/serde"]
```
Or what do you think is more likely?
1. someone will be making a `no-std` build with `hashbrown` but they don't want `serde`  
2. someone will make a default (`std`) build with `serde` but they don't want `hashbrown`

I think 2 is more likely, but even though awkward I think it's better to add a `hashbrown_serde` feature to give the users complete control, especially since it only effect the niche users making `no-std` builds.

evanlinjin: That is a good point. I think `2` is the more likely scenario. However, I think we do not need an additional feature flag, since the user can just enable `"hashbrown/serde"`.

notmandatory: Yes, agree, no need for a `hashbrown_serde` feature, using `"hashbrown/serde"` makes more sense.

notmandatory: oh wait.. but this build fails:
```
cargo check --no-default-features --features bitcoin/no-std,miniscript/no-std,bdk_chain/hashbrown
```
only way to do it would be:
```
cargo check --no-default-features --features bitcoin/no-std,miniscript/no-std,bdk_chain/hashbrown_serde
```
EDIT: above is from the `crates/bdk` directory as in the `check-no-std` CI test.

evanlinjin: I think not enabling `hashbrown/serde` by default results in some confusing cargo feature dependencies. I think I will revert it back to always enabling `hashbrown/serde` whenever `hashbrown` is enabled, and we can discuss this later on?

notmandatory (approved):
ACK 26ade1172622f66cb95c831c4a8e7535409e8e81

Issue #916, opened 2023-03-22
`rust-bitcoin` 0.30.0 upgrade may be annoying


tcharding:
Hi,

We just released v0.30.0 of `rust-bitcoin`, assuming you guys are going to upgrade, can I ask a favour please. If the upgrade is painful can you please give me feedback. There is a fair bit of churn going on with crate smashing and I'd like to make it as painless as possible for downstream. I don't personally work on any code that uses `rust-bitcoin` so I have a tendency to get tunnel vision.

If its super painful, just prod me and I can lend a hand.

Thanks

notmandatory:
Thanks, we'll also need rust-bitcoin/rust-miniscript#527 to be released, but we can still start getting a bdk PR ready.

tcharding:
Oh of course, I probably should help out there first huh :)

notmandatory:
Reopening this issue since @danielabrozzoni is working on it in #1023.

nondiremanuel:
@notmandatory shall we close this issue? I saw that #1023 was merged

notmandatory:
Yes! fixed by #1023.

PR #1023, opened 2023-06-28
Update rust bitcoin

danielabrozzoni:
### Description

Updates to rust-bitcoin 0.30.0 and miniscript 0.10.0

Not covered in this PR:
- https://github.com/bitcoindevkit/bdk/issues/1036. Although the latter is deprecated, I think it's better if I update it in a separate PR, as this one is pretty big already. 
- https://github.com/bitcoindevkit/bdk/issues/1037
- https://github.com/bitcoindevkit/bdk/issues/1038

Heads up, I'm explicitly denying multipath descriptors until we have better tests for them. See the commit message of 23fba7aee9b55bf59db73c296b9fb025a256502e

### Changelog notice

- Update to `rust-bitcoin` 0.30.0 and `miniscript` 10.0.0

#### All Submissions:

### DIFF
--- a/crates/bdk/Cargo.toml
+++ b/crates/bdk/Cargo.toml
@@ -15,14 +15,14 @@ rust-version = "1.57"
 [dependencies]
 log = "0.4"
 rand = "^0.8"
-miniscript = { version = "9", features = ["serde"], default-features = false }
-bitcoin = { version = "0.29", features = ["serde", "base64", "rand"], default-features = false }
+miniscript = { version = "10.0.0", features = ["serde"], default-features = false }
+bitcoin = { version = "0.30.0", features = ["serde", "base64", "rand-std"], default-features = false }
 serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1.0" }
 bdk_chain = { path = "../chain", version = "0.5.0", features = ["miniscript", "serde"], default-features = false }
 
 # Optional dependencies
-hwi = { version = "0.5", optional = true, features = [ "use-miniscript"] }
+hwi = { version = "0.7.0", optional = true, features = [ "miniscript"] }
 bip39 = { version = "1.0.1", optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
@@ -46,8 +46,6 @@ dev-getrandom-wasm = ["getrandom/js"]
 [dev-dependencies]
 lazy_static = "1.4"
 env_logger = "0.7"
-# Move back to importing from rust-bitcoin once https://github.com/rust-bitcoin/rust-bitcoin/pull/1342 is released
-base64 = "^0.13"
 assert_matches = "1.5.0"
 
 [package.metadata.docs.rs]
--- a/crates/bdk/src/descriptor/dsl.rs
+++ b/crates/bdk/src/descriptor/dsl.rs
@@ -516,13 +516,14 @@ macro_rules! descriptor {
         use $crate::miniscript::descriptor::{Descriptor, DescriptorPublicKey};
 
         $crate::impl_top_level_pk!(Pkh, $crate::miniscript::Legacy, $key)
+            .and_then(|(a, b, c)| Ok((a.map_err(|e| miniscript::Error::from(e))?, b, c)))
             .map(|(a, b, c)| (Descriptor::<DescriptorPublicKey>::Pkh(a), b, c))
     });
     ( wpkh ( $key:expr ) ) => ({
         use $crate::miniscript::descriptor::{Descriptor, DescriptorPublicKey};
 
         $crate::impl_top_level_pk!(Wpkh, $crate::miniscript::Segwitv0, $key)
-            .and_then(|(a, b, c)| Ok((a?, b, c)))
+            .and_then(|(a, b, c)| Ok((a.map_err(|e| miniscript::Error::from(e))?, b, c)))
             .map(|(a, b, c)| (Descriptor::<DescriptorPublicKey>::Wpkh(a), b, c))
     });
     ( sh ( wpkh ( $key:expr ) ) ) => ({
@@ -532,7 +533,7 @@ macro_rules! descriptor {
         use $crate::miniscript::descriptor::{Descriptor, DescriptorPublicKey, Sh};
 
         $crate::impl_top_level_pk!(Wpkh, $crate::miniscript::Segwitv0, $key)
-            .and_then(|(a, b, c)| Ok((a?, b, c)))
+            .and_then(|(a, b, c)| Ok((a.map_err(|e| miniscript::Error::from(e))?, b, c)))
             .and_then(|(a, b, c)| Ok((Descriptor::<DescriptorPublicKey>::Sh(Sh::new_wpkh(a.into_inner())?), b, c)))
     });
     ( sh ( $( $minisc:tt )* ) ) => ({
@@ -702,7 +703,7 @@ macro_rules! fragment {
         $crate::keys::make_pkh($key, &secp)
     });
     ( after ( $value:expr ) ) => ({
-        $crate::impl_leaf_opcode_value!(After, $crate::bitcoin::PackedLockTime($value)) // TODO!! https://github.com/rust-bitcoin/rust-bitcoin/issues/1302
+        $crate::impl_leaf_opcode_value!(After, $crate::miniscript::AbsLockTime::from_consensus($value))
     });
     ( older ( $value:expr ) ) => ({
         $crate::impl_leaf_opcode_value!(Older, $crate::bitcoin::Sequence($value)) // TODO!!
@@ -796,7 +797,6 @@ macro_rules! fragment {
 #[cfg(test)]
 mod test {
     use alloc::string::ToString;
-    use bitcoin::hashes::hex::ToHex;
     use bitcoin::secp256k1::Secp256k1;
     use miniscript::descriptor::{DescriptorPublicKey, KeyMap};
     use miniscript::{Descriptor, Legacy, Segwitv0};
@@ -805,8 +805,8 @@ mod test {
 
     use crate::descriptor::{DescriptorError, DescriptorMeta};
     use crate::keys::{DescriptorKey, IntoDescriptorKey, ValidNetworks};
+    use bitcoin::bip32;
     use bitcoin::network::constants::Network::{Bitcoin, Regtest, Signet, Testnet};
-    use bitcoin::util::bip32;
     use bitcoin::PrivateKey;
 
     // test the descriptor!() macro
@@ -822,18 +822,15 @@ mod test {
         assert_eq!(desc.is_witness(), is_witness);
         assert_eq!(!desc.has_wildcard(), is_fixed);
         for i in 0..expected.len() {
-            let index = i as u32;
-            let child_desc = if !desc.has_wildcard() {
-                desc.at_derivation_index(0)
-            } else {
-                desc.at_derivation_index(index)
-            };
+            let child_desc = desc
+                .at_derivation_index(i as u32)
+                .expect("i is not hardened");
             let address = child_desc.address(Regtest);
             if let Ok(address) = address {
                 assert_eq!(address.to_string(), *expected.get(i).unwrap());
             } else {
                 let script = child_desc.script_pubkey();
-                assert_eq!(script.to_hex().as_str(), *expected.get(i).unwrap());
+                assert_eq!(script.to_hex_string(), *expected.get(i).unwrap());
             }
         }
     }
@@ -1178,9 +1175,7 @@ mod test {
     }
 
     #[test]
-    #[should_panic(
-        expected = "Miniscript(ContextError(CompressedOnly(\"04b4632d08485ff1df2db55b9dafd23347d1c47a457072a1e87be26896549a87378ec38ff91d43e8c2092ebda601780485263da089465619e0358a5c1be7ac91f4\")))"
-    )]
+    #[should_panic(expected = "Miniscript(ContextError(UncompressedKeysNotAllowed))")]
     fn test_dsl_miniscript_checks() {
         let mut uncompressed_pk =
             PrivateKey::from_wif("L5EZftvrYaSudiozVRzTqLcHLNDoVn7H5HSfM9BAN6tMJX8oTWz6").unwrap();
--- a/crates/bdk/src/descriptor/error.rs
+++ b/crates/bdk/src/descriptor/error.rs
@@ -32,11 +32,11 @@ pub enum Error {
     InvalidDescriptorCharacter(u8),
 
     /// BIP32 error
-    Bip32(bitcoin::util::bip32::Error),
+    Bip32(bitcoin::bip32::Error),
     /// Error during base58 decoding
-    Base58(bitcoin::util::base58::Error),
+    Base58(bitcoin::base58::Error),
     /// Key-related error
-    Pk(bitcoin::util::key::Error),
+    Pk(bitcoin::key::Error),
     /// Miniscript error
     Miniscript(miniscript::Error),
     /// Hex decoding error
@@ -81,9 +81,9 @@ impl fmt::Display for Error {
 #[cfg(feature = "std")]
 impl std::error::Error for Error {}
 
-impl_error!(bitcoin::util::bip32::Error, Bip32);
-impl_error!(bitcoin::util::base58::Error, Base58);
-impl_error!(bitcoin::util::key::Error, Pk);
+impl_error!(bitcoin::bip32::Error, Bip32);
+impl_error!(bitcoin::base58::Error, Base58);
+impl_error!(bitcoin::key::Error, Pk);
 impl_error!(miniscript::Error, Miniscript);
 impl_error!(bitcoin::hashes::hex::Error, Hex);
 impl_error!(crate::descriptor::policy::PolicyError, Policy);
--- a/crates/bdk/src/descriptor/error.rs
+++ b/crates/bdk/src/descriptor/error.rs
@@ -22,6 +22,8 @@ pub enum Error {
     InvalidDescriptorChecksum,
     /// The descriptor contains hardened derivation steps on public extended keys
     HardenedDerivationXpub,
+    /// The descriptor contains multipath keys
+    MultiPath,
 
     /// Error thrown while working with [`keys`](crate::keys)
     Key(crate::keys::KeyError),
@@ -63,6 +65,10 @@ impl fmt::Display for Error {
             Self::HardenedDerivationXpub => write!(
                 f,
                 "The descriptor contains hardened derivation steps on public extended keys"
+            ),
+            Self::MultiPath => write!(
+                f,
+                "The descriptor contains multipath keys, which are not supported yet"
             ),
             Self::Key(err) => write!(f, "Key error: {}", err),
             Self::Policy(err) => write!(f, "Policy error: {}", err),
--- a/crates/bdk/src/descriptor/mod.rs
+++ b/crates/bdk/src/descriptor/mod.rs
@@ -18,17 +18,17 @@ use crate::collections::BTreeMap;
 use alloc::string::String;
 use alloc::vec::Vec;
 
-use bitcoin::util::bip32::{ChildNumber, DerivationPath, ExtendedPubKey, Fingerprint, KeySource};
-use bitcoin::util::{psbt, taproot};
-use bitcoin::{secp256k1, PublicKey, XOnlyPublicKey};
+use bitcoin::bip32::{ChildNumber, DerivationPath, ExtendedPubKey, Fingerprint, KeySource};
+use bitcoin::{key::XOnlyPublicKey, secp256k1, PublicKey};
+use bitcoin::{psbt, taproot};
 use bitcoin::{Network, TxOut};
 
 use miniscript::descriptor::{
-    DefiniteDescriptorKey, DescriptorSecretKey, DescriptorType, InnerXKey, SinglePubKey,
+    DefiniteDescriptorKey, DescriptorMultiXKey, DescriptorSecretKey, DescriptorType,
+    DescriptorXKey, InnerXKey, KeyMap, SinglePubKey, Wildcard,
 };
 pub use miniscript::{
-    descriptor::DescriptorXKey, descriptor::KeyMap, descriptor::Wildcard, Descriptor,
-    DescriptorPublicKey, Legacy, Miniscript, ScriptContext, Segwitv0,
+    Descriptor, DescriptorPublicKey, Legacy, Miniscript, ScriptContext, Segwitv0,
 };
 use miniscript::{ForEachKey, MiniscriptKey, TranslatePk};
 
@@ -59,16 +59,16 @@ pub type DerivedDescriptor = Descriptor<DefiniteDescriptorKey>;
 /// Alias for the type of maps that represent derivation paths in a [`psbt::Input`] or
 /// [`psbt::Output`]
 ///
-/// [`psbt::Input`]: bitcoin::util::psbt::Input
-/// [`psbt::Output`]: bitcoin::util::psbt::Output
+/// [`psbt::Input`]: bitcoin::psbt::Input
+/// [`psbt::Output`]: bitcoin::psbt::Output
 pub type HdKeyPaths = BTreeMap<secp256k1::PublicKey, KeySource>;
 
 /// Alias for the type of maps that represent taproot key origins in a [`psbt::Input`] or
 /// [`psbt::Output`]
 ///
-/// [`psbt::Input`]: bitcoin::util::psbt::Input
-/// [`psbt::Output`]: bitcoin::util::psbt::Output
-pub type TapKeyOrigins = BTreeMap<bitcoin::XOnlyPublicKey, (Vec<taproot::TapLeafHash>, KeySource)>;
+/// [`psbt::Input`]: bitcoin::psbt::Input
+/// [`psbt::Output`]: bitcoin::psbt::Output
+pub type TapKeyOrigins = BTreeMap<XOnlyPublicKey, (Vec<taproot::TapLeafHash>, KeySource)>;
 
 /// Trait for types which can be converted into an [`ExtendedDescriptor`] and a [`KeyMap`] usable by a wallet in a specific [`Network`]
 pub trait IntoWalletDescriptor {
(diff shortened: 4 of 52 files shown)


crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -38,20 +38,22 @@
 //!         &self,
 //!         required_utxos: Vec<WeightedUtxo>,
 //!         optional_utxos: Vec<WeightedUtxo>,
-//!         fee_rate: FeeRate,
+//!         fee_rate: bdk::FeeRate,
 //!         target_amount: u64,
-//!         drain_script: &Script,
+//!         drain_script: &ScriptBuf,
```
evanlinjin: Instead of `&ScriptBuf`, consider having `&Script`.

crates/chain/src/chain_data.rs
```diff
@@ -104,7 +104,7 @@ impl Default for BlockId {
     fn default() -> Self {
         Self {
             height: Default::default(),
-            hash: BlockHash::from_inner([0u8; 32]),
+            hash: BlockHash::from_slice(&[0u8; 32]).expect("len is 32"),
```
evanlinjin: Consider using `BlockHash::all_zeros()`.

apoelstra: Also, `from_inner` was renamed to `from_byte_array`, so for other fixed arrays you still don't need to `.expect`.

crates/chain/tests/test_spk_txout_index.rs
```diff
@@ -1,18 +1,21 @@
...
+    let spk2 = ScriptBuf::from_hex("00142b57404ae14f08c3a0c903feb2af7830605eb00f").unwrap();
 
     let mut index = SpkTxOutIndex::default();
     index.insert_spk(0, spk1.clone());
     index.insert_spk(1, spk2.clone());
 
     let tx1 = Transaction {
         version: 0x02,
-        lock_time: PackedLockTime(0),
+        lock_time: LockTime::Blocks(Height::ZERO),
```
apoelstra: In e83ea8e1204833049e8d1dd5c6b2be4c981e463d:

There is a `Locktime::ZERO` constant you can use directly.

Also, as a style thing, I find it clearer to import `absolute` and `relative` modules and then to use `absolute::LockTime` everywhere so it's clear what kind of locktime is under consideration.

crates/bdk/src/descriptor/dsl.rs
```diff
@@ -702,7 +703,7 @@ macro_rules! fragment {
         $crate::keys::make_pkh($key, &secp)
     });
     ( after ( $value:expr ) ) => ({
-        $crate::impl_leaf_opcode_value!(After, $crate::bitcoin::PackedLockTime($value)) // TODO!! https://github.com/rust-bitcoin/rust-bitcoin/issues/1302
+        $crate::impl_leaf_opcode_value!(After, $crate::bitcoin::absolute::LockTime::from_consensus($value).into()) // TODO!! https://github.com/rust-bitcoin/rust-bitcoin/issues/1302
```
apoelstra: In d12242ec746ffc51da64672d7c7f7e014258d190:

You can do `$crate::miniscript::AbsLockTime::from_consensus($value)` which is much shorter than this and avoids the `.into()`. Can probably also drop the TODO at that point unless you want to file an issue on rust-miniscript suggesting that `AbsLockTIme` implement `From<u32>`.

crates/bdk/src/descriptor/mod.rs
```diff
@@ -494,7 +512,7 @@ impl DescriptorMeta for ExtendedDescriptor {
             false
         });
 
-        path_found.map(|path| self.at_derivation_index(path))
+        path_found.map(|path| self.at_derivation_index(path).expect("TODO"))
```
apoelstra: In d12242ec746ffc51da64672d7c7f7e014258d190:

This TODO can just be `unwrap` with a comment explaining that `path_found` came out of a child index and therefore is in range.

crates/bdk/src/descriptor/mod.rs
```diff
@@ -545,7 +563,7 @@ impl DescriptorMeta for ExtendedDescriptor {
             return None;
         }
 
-        let descriptor = self.at_derivation_index(0);
+        let descriptor = self.at_derivation_index(0).expect("TODO");
```
apoelstra: In d12242ec746ffc51da64672d7c7f7e014258d190:

You can just unwrap. This function only fails if the index is out of range, and 0 is not.

apoelstra: Ditto at several other places.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1403,7 +1404,7 @@ impl<D> Wallet<D> {
             .index
             .index_of_spk(&txout.script_pubkey)?;
         let descriptor = self.get_descriptor_for_keychain(keychain);
-        Some(descriptor.at_derivation_index(child))
+        Some(descriptor.at_derivation_index(child).expect("TODO"))
```
apoelstra: In d12242ec746ffc51da64672d7c7f7e014258d190:

These TODOs could be `.ok()?` which wil ljust return None if the child index is somehow out of range.

example-crates/example_electrum/src/main.rs
```diff
@@ -84,6 +84,7 @@ fn main() -> anyhow::Result<()> {
         Network::Testnet => "ssl://electrum.blockstream.info:60002",
         Network::Regtest => "tcp://localhost:60401",
         Network::Signet => "tcp://signet-electrumx.wakiyamap.dev:50001",
+        _ => todo!(),
```
apoelstra: In a219219691bb1c6d47e8f8a9df3100f92eee0baf:

`todo` is not a bad choice for this but I'd panic with "Unknown network" or something so it's clear this isn't a code todo.

apoelstra:
Done reviewing 81ef698ed9c1fdc6dee1bc2757fa7c9209273343.

Overall looks great. I apologize for what an effort this was.

notmandatory:
This should fix #916

crates/bdk/Cargo.toml
```diff
@@ -15,8 +15,9 @@ rust-version = "1.57"
 [dependencies]
 log = "=0.4.18"
 rand = "^0.8"
-miniscript = { version = "9", features = ["serde"], default-features = false }
-bitcoin = { version = "0.29", features = ["serde", "base64", "rand"], default-features = false }
+miniscript = { version = "10", features = ["serde"], default-features = false }
+bitcoin = { version = "0.30", features = ["serde", "base64", "rand"], default-features = false }
+secp256k1 = { version = "0.27", features = [ "rand-std" ] }
```
tcharding:
```suggestion
bitcoin = { version = "0.30", features = ["serde", "base64", "rand-std"], default-features = false }
```
I guess the secp dependency was just to enable "rand-std"? `rust-bitcoin` has "rand-std" that can do that for you.

crates/bdk/src/descriptor/mod.rs
```diff
@@ -600,7 +617,7 @@ mod test {
             "wpkh(02b4632d08485ff1df2db55b9dafd23347d1c47a457072a1e87be26896549a8737)",
         )
         .unwrap();
-        let psbt: psbt::PartiallySignedTransaction = deserialize(
+        let psbt = psbt::PartiallySignedTransaction::deserialize(
```
tcharding:
```suggestion
        let psbt = psbt::Psbt::deserialize(
```

`bitcoin::psbt::Psbt` is available now. Next release hopefully `bitcoin::Psbt` will be (https://github.com/rust-bitcoin/rust-bitcoin/pull/1938)

crates/bdk/src/descriptor/policy.rs
```diff
@@ -68,7 +68,7 @@ use crate::wallet::utils::{After, Older, SecpCtx};
 use super::checksum::calc_checksum;
 use super::error::Error;
 use super::XKeyUtils;
-use bitcoin::util::psbt::{Input as PsbtInput, PartiallySignedTransaction as Psbt};
+use bitcoin::psbt::{Input as PsbtInput, PartiallySignedTransaction as Psbt};
```
tcharding: Stylistic choice but you could use `psbt::Input` and `Psbt`. Along with:

```
use bitcoin::psbt::{self, Psbt};
```

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -38,20 +38,22 @@
...
 //!     ) -> Result<CoinSelectionResult, bdk::Error> {
 //!         let mut selected_amount = 0;
-//!         let mut additional_weight = 0;
+//!         let mut additional_weight = Weight::from_wu(0);
```
tcharding:
```suggestion
//!         let mut additional_weight = Weight::ZERO;
```

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -188,7 +193,7 @@ pub trait CoinSelectionAlgorithm: core::fmt::Debug {
         optional_utxos: Vec<WeightedUtxo>,
         fee_rate: FeeRate,
         target_amount: u64,
-        drain_script: &Script,
+        drain_script: &ScriptBuf,
```
tcharding: @evanlinjin already mentioned it but just FYI the `ScriptBuf` is to `Script` as `PathBuf` is to `Path` and `String` is to `&str` - so typically one would always using either `&Script` or `ScriptBuf` depending on whether you want to own the script data or not.

crates/bdk/src/wallet/signer.rs
```diff
@@ -499,12 +550,12 @@ fn sign_psbt_ecdsa(
...
     pubkey: PublicKey,
     psbt_input: &mut psbt::Input,
-    hash: bitcoin::Sighash,
+    hash: impl bitcoin::hashes::Hash,
     hash_ty: EcdsaSighashType,
     secp: &SecpCtx,
     allow_grinding: bool,
 ) {
-    let msg = &Message::from_slice(&hash.into_inner()[..]).unwrap();
+    let msg = &Message::from_slice(&hash[..]).unwrap();
```
tcharding: I think you can just do `Message::from(hash)`.

crates/bdk/src/wallet/signer.rs
```diff
@@ -523,12 +574,10 @@ fn sign_psbt_schnorr(
     pubkey: XOnlyPublicKey,
     leaf_hash: Option<taproot::TapLeafHash>,
     psbt_input: &mut psbt::Input,
-    hash: taproot::TapSighashHash,
-    hash_ty: SchnorrSighashType,
+    hash: sighash::TapSighash,
```
tcharding: Stylistic choice but I'd do 
```suggestion
    hash: TapSighash,
```

crates/bdk/src/wallet/signer.rs
```diff
@@ -849,19 +903,26 @@ impl ComputeSighash for Legacy {
     }
 }
 
-fn p2wpkh_script_code(script: &Script) -> Script {
+fn p2wpkh_script_code(script: &ScriptBuf) -> ScriptBuf {
```
tcharding: We have `ScriptBuf::p2wpkh_script_code` now (I actually took the code from bdk when adding it :)

crates/esplora/Cargo.toml
```diff
@@ -13,7 +13,7 @@ readme = "README.md"
 
 [dependencies]
 bdk_chain = { path = "../chain", version = "0.4.0", default-features = false, features = ["serde", "miniscript"] }
-esplora-client = { version = "0.5", default-features = false }
+esplora-client = { git = "https://github.com/tcharding/rust-esplora-client.git", branch = "06-10-upgrade-bitcoin", default-features = false }
```
tcharding: I have my branches cofigured to auto-delete on merge so I changed this to
```
esplora-client = { git = "https://github.com/bitcoindevkit/rust-esplora-client.git", branch = "master", default-features = false }
```

crates/bdk/src/wallet/signer.rs
```diff
@@ -849,19 +903,26 @@ impl ComputeSighash for Legacy {
...
         .push_opcode(opcodes::all::OP_DUP)
         .push_opcode(opcodes::all::OP_HASH160)
-        .push_slice(&script[2..])
+        .push_slice(script_bytes)
         .push_opcode(opcodes::all::OP_EQUALVERIFY)
         .push_opcode(opcodes::all::OP_CHECKSIG)
         .into_script()
 }
 
 impl ComputeSighash for Segwitv0 {
```
tcharding, line 904: I'm not exactly sure bit it looks like the implementations of this `ComputeSighash` trait contain loads of stuff that should be done by `rust-bitcoin::psbt` (complex enough to require a separate PR though). The taproot stuff is not done yet but for legacy and segwitv0 you should be able to use the new signing stuff in `psbt` (see `Psbt::sign` and `Psbt::sighash_ecdsa`). If you do play around with it and find it lacking please raise an issue.

crates/bdk/src/wallet/signer.rs
```diff
@@ -912,7 +973,7 @@ impl ComputeSighash for Segwitv0 {
                 } else if psbt_input
                     .redeem_script
                     .as_ref()
-                    .map(Script::is_v0_p2wpkh)
+                    .map(|s| s.as_script().is_v0_p2wpkh())
```
tcharding: No need to call `as_script` explicitly, `ScriptBuf` derefs to `Script` so all the methods on `Script` are available on `ScriptBuf`.

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -27,7 +31,7 @@
 //!     // Create a transaction with one output to `to_address` of 50_000 satoshi
 //!     .add_recipient(to_address.script_pubkey(), 50_000)
 //!     // With a custom fee rate of 5.0 satoshi/vbyte
-//!     .fee_rate(FeeRate::from_sat_per_vb(5.0))
+//!     .fee_rate(bdk::FeeRate::from_sat_per_vb(5.0))
```
tcharding, line 30: I didn't look super closely but you should be able to get rid of your `FeeRate` now and use `bitcoin::FeeRate`.

crates/bdk/tests/wallet.rs
```diff
@@ -725,7 +729,7 @@ fn test_create_tx_set_redeem_witness_script_p2wsh_p2sh() {
     builder.drain_to(addr.script_pubkey()).drain_wallet();
     let (psbt, _) = builder.finish().unwrap();
 
-    let script = Script::from(
+    let script = ScriptBuf::from(
```
tcharding: Can just call `SciptBuf::from_hex` directly here.

crates/bdk/tests/wallet.rs
```diff
@@ -809,7 +813,10 @@ fn test_create_tx_add_utxo() {
         )
         .unwrap();
 
-    let addr = Address::from_str("2N1Ffz3WaNzbeLFBb51xyFMHYSEUXcbiSoX").unwrap();
+    let addr = Address::from_str("2N1Ffz3WaNzbeLFBb51xyFMHYSEUXcbiSoX")
+        .unwrap()
+        .require_network(Network::Testnet)
+        .unwrap();
```
tcharding: In test code, if you want, you can just call `assumed_checked` instead of `require_network`.

crates/chain/src/spk_iter.rs
```diff
@@ -151,12 +150,25 @@ mod test {
     #[allow(clippy::iter_nth_zero)]
     fn test_spkiterator_wildcard() {
         let (_, external_desc, _) = init_txout_index();
-        let external_spk_0 = external_desc.at_derivation_index(0).script_pubkey();
-        let external_spk_16 = external_desc.at_derivation_index(16).script_pubkey();
-        let external_spk_20 = external_desc.at_derivation_index(20).script_pubkey();
-        let external_spk_21 = external_desc.at_derivation_index(21).script_pubkey();
+        let external_spk_0 = external_desc
+            .at_derivation_index(0)
```
tcharding: Just fyi if you want to keep these all on one line you can use `#[rustfmt::skip]` on the function.

nursery/tmp_plan/src/lib.rs
```diff
@@ -16,15 +16,17 @@
 //! witness/script_sig for the input.
 use bdk_chain::{bitcoin, collections::*, miniscript};
 use bitcoin::{
-    blockdata::{locktime::LockTime, transaction::Sequence},
+    address::WitnessVersion,
+    bip32::{DerivationPath, Fingerprint, KeySource},
+    // TODO: is locktime absolute?
```
tcharding: Did you want this in there still?

nursery/tmp_plan/src/lib.rs
```diff
@@ -201,7 +203,7 @@ where
                     ..
                 } => PlanState::Complete {
                     final_script_sig: None,
-                    final_script_witness: Some(Witness::from_vec(witness)),
+                    final_script_witness: Some(Witness::from_slice(&witness)),
```
tcharding:
```suggestion
                    final_script_witness: Some(Witness::from(witness)),
```

tcharding (commented):
Wow you guys really use `rust-bitcoin` a lot. I hope we can stabalize the API soon so you don't have to go through all this work too many times. Thanks for being patient with our wild API changes.

crates/bdk/src/wallet/tx_builder.rs
danielabrozzoni: I opened an issue for this one: https://github.com/bitcoindevkit/bdk/issues/1037 as I feel it needs more discussion: at the moment `bdk::FeeRate` has a couple methods that `bitcoin::FeeRate` doesn't have, and although switching is possible, I think we should discuss whether it's better to switch immediately or to first submit a patch to you guys :)

crates/bdk/src/wallet/signer.rs
danielabrozzoni: Fair, thanks! This PR is big enough already, but I opened https://github.com/bitcoindevkit/bdk/issues/1038 so that I can't forget :)

danielabrozzoni:
Thanks for the reviews, they've been super helpful! 

I just updated the code and rebased on top of master. CI should be fine now.

apoelstra (approved):
utACK 6259639a680161ac78caabee5a9a6a84e866c1a7

crates/bdk/src/wallet/signer.rs
tcharding: Awesome, thanks.

crates/chain/src/spk_txout_index.rs
```diff
@@ -152,11 +152,11 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
         use bitcoin::hashes::Hash;
         use core::ops::Bound::*;
         let min_op = OutPoint {
-            txid: Txid::from_inner([0x00; 32]),
+            txid: Txid::from_slice(&[0x00; 32]).expect("len is 32"),
```
tcharding: nit: `Txid::all_zeros()`

crates/bdk/src/wallet/signer.rs
```diff
@@ -499,12 +548,12 @@ fn sign_psbt_ecdsa(
...
     pubkey: PublicKey,
     psbt_input: &mut psbt::Input,
-    hash: bitcoin::Sighash,
+    hash: impl bitcoin::hashes::Hash,
     hash_ty: EcdsaSighashType,
     secp: &SecpCtx,
     allow_grinding: bool,
 ) {
-    let msg = &Message::from_slice(&hash.into_inner()[..]).unwrap();
+    let msg = &Message::from_slice(&hash[..]).unwrap();
```
tcharding: Missed this one, you did the one below :)
```rust
let msg = &Message::from(hash);
```

tcharding (approved):
utACK 23fba7aee9b55bf59db73c296b9fb025a256502e

tcharding:
Clean! Nice one.

crates/bdk/src/descriptor/dsl.rs
```diff
@@ -824,16 +824,16 @@ mod test {
         for i in 0..expected.len() {
             let index = i as u32;
             let child_desc = if !desc.has_wildcard() {
-                desc.at_derivation_index(0)
+                desc.at_derivation_index(0).unwrap()
             } else {
-                desc.at_derivation_index(index)
+                desc.at_derivation_index(index).unwrap()
             };
```
evanlinjin: **Nit:** I think this can be simplified to this.

```rust
let child_desc = desc
    .at_derivation_index(i as u32)
    .expect("must not be hardened");
```

For non-wildcard or single keys, `at_derivation_index` basically just wraps the original key in `DefiniteDescriptorKey` (and `index` is not used anyway).

crates/bdk/src/descriptor/error.rs
```diff
@@ -64,6 +66,7 @@ impl fmt::Display for Error {
                 f,
                 "The descriptor contains hardened derivation steps on public extended keys"
             ),
+            Self::MultiPath => write!(f, "The descriptor contains multipath keys"),
```
evanlinjin:
```suggestion
            Self::MultiPath => write!(f, "The descriptor contains multipath keys, which are not supported yet"),
```

I think it makes sense to make it explicit that mutipath keys aren't inherently problematic, it's just we don't support it yet.

crates/bdk/src/descriptor/template.rs
```diff
@@ -565,30 +565,30 @@ mod test {
...
-        let xprvkey = bitcoin::util::bip32::ExtendedPrivKey::from_str("xprv9s21ZrQH143K2fpbqApQL69a4oKdGVnVN52R82Ft7d1pSqgKmajF62acJo3aMszZb6qQ22QsVECSFxvf9uyxFUvFYQMq3QbtwtRSMjLAhMf").unwrap();
+        let xprvkey = bitcoin::bip32::ExtendedPrivKey::from_str("xprv9s21ZrQH143K2fpbqApQL69a4oKdGVnVN52R82Ft7d1pSqgKmajF62acJo3aMszZb6qQ22QsVECSFxvf9uyxFUvFYQMq3QbtwtRSMjLAhMf").unwrap();
         assert_eq!(Network::Bitcoin, xprvkey.network);
         let xdesc = Bip44(xprvkey, KeychainKind::Internal)
             .build(Network::Bitcoin)
             .unwrap();
 
         if let ExtendedDescriptor::Pkh(pkh) = xdesc.0 {
-            let path: Vec<ChildNumber> = pkh.into_inner().full_derivation_path().into();
+            let path: Vec<ChildNumber> = pkh.into_inner().full_derivation_path().unwrap().into();
```
evanlinjin, line 577:
```suggestion
            let path: Vec<ChildNumber> = pkh.into_inner().full_derivation_path().expect("must not be multipath key").into();
```

nursery/coin_select/src/coin_selector.rs
```diff
@@ -94,18 +94,14 @@ impl CoinSelectorOpt {
...
             output: txouts.to_vec(),
         };
-        let base_weight = tx.weight();
-        // this awkward calculation is necessary since TxOut doesn't have \.weight()
-        let drain_weight = {
-            tx.output.push(drain_output.clone());
-            tx.weight() - base_weight
-        };
+        let base_weight = tx.weight().to_wu();
+        let drain_weight = drain_output.weight();
```
evanlinjin: My only concern with this, is this weight calculation does not take into account varint changes for output count.

crates/electrum/Cargo.toml
```diff
@@ -13,4 +13,6 @@ readme = "README.md"
 
 [dependencies]
 bdk_chain = { path = "../chain", version = "0.5.0", features = ["serde", "miniscript"] }
-electrum-client = { version = "0.12" }
+electrum-client = { version = "0.17" }
+# rustls 0.21.2 increased their MSRV
+rustls = { version = "=0.21.1", optional = true, features = ["dangerous_configuration"] }
```
evanlinjin: I think it will make sense to use #1046 here.

crates/chain/src/spk_txout_index.rs
```diff
@@ -152,11 +152,11 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
         use bitcoin::hashes::Hash;
         use core::ops::Bound::*;
         let min_op = OutPoint {
-            txid: Txid::from_inner([0x00; 32]),
+            txid: Txid::all_zeros(),
             vout: u32::MIN,
         };
         let max_op = OutPoint {
-            txid: Txid::from_inner([0xff; 32]),
+            txid: Txid::from_slice(&[0xff; 32]).expect("len is 32"),
```
evanlinjin: Nit:

```suggestion
            txid: Txid::from_byte_array([0xff; Txid::LEN]),
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -1403,13 +1407,14 @@ impl<D> Wallet<D> {
...
         let descriptor = self.get_descriptor_for_keychain(keychain);
-        Some(descriptor.at_derivation_index(child))
+        descriptor.at_derivation_index(child).ok()
     }
 
     fn get_available_utxos(&self) -> Vec<(LocalUtxo, usize)> {
         self.list_unspent()
             .map(|utxo| {
                 let keychain = utxo.keychain;
+                #[allow(deprecated)]
```
evanlinjin, line 1417: This is a good call. Saves time.

evanlinjin (commented):
Thank you for this work!

Can I suggest to use `.expect()` instead of `.unwrap()` in some places?

I understand that it becomes tedious to do it everywhere, especially if there are multiple calls to `.at_derivation_index()` (in which case I would suggest to leave a comment explaining why we don't expect a hardened wildcard).

An explanation as to why we do not expect an error goes a long way.

evanlinjin (changes requested):
I have looked through everything and tested the examples and made sure they all still work :smile:

In my opinion, these are mandatory changes:

* Do not pin dependencies: https://github.com/bitcoindevkit/bdk/pull/1023#discussion_r1281591844
* Better description for multipath error: https://github.com/bitcoindevkit/bdk/pull/1023#discussion_r1281446160
* Incorrect weight calculation: https://github.com/bitcoindevkit/bdk/pull/1023#discussion_r1281584907

After which, I will give an ACK.

nursery/coin_select/src/coin_selector.rs
danielabrozzoni: Ahh right, I missed that, thanks!

crates/electrum/Cargo.toml
danielabrozzoni: Ugh, of course 🤦🏻‍♀️

crates/bdk/src/descriptor/template.rs
danielabrozzoni: I'm all in for using expect more often, but I don't think I'll do it in this specific case, as it's a test and it's obvious from the descriptor (just a couple of lines above) that it's not a multipath. But yeah, I'm replacing unwrap with expect in the rest of the code :)

danielabrozzoni:
Rebased and fixed review comments.

evanlinjin (approved):
ACK 1da3b304bb0aa29e3bb013e0a583c34ec390b32d

Looked through everything and made sure the examples still worked as expected (which they did).

Issue #1025, opened 2023-07-04
Release 1.0.0-alpha.1


notmandatory:
### Summary

The BDK 1.0.0-alpha release should be used for **experimentation only**, APIs are still unstable and the code is not fully tested. This alpha.1 release introduces the new `ChainOracle` struct for more efficient chain syncing. A new `std` default feature was added for `bdk`, `bdk_chain` and `bdk_esplora` crates; when disabled these crates can be used in `no-std` projects. BDK 1.0.0-alpha.x docs are now published to docs.rs.

### Commit

f4d2a76661d465dd0640d54470b95cc7b5ece979

### Changelog

Fixed

- Fixed a bug in the policy condition calculation. #932
- Pin base64 to 0.21.0 #990
- Fix docsrs publishing for bdk crate. #1011

Changed

- Refactor `bdk_chain` to use new `ChainOracle` structure. #926 #963 #965 #975 #976 
- Better no std support. #894
  - Set `default-features = false` for `rust-bitcoin` and `rust-miniscript`.
  - Introduce `std` default feature for `bdk`, `bdk_chain` and `bdk_esplora`.

Added

- Add a simple conversion tool for going to kilo weight units. #953
- Add Custom spk iterator. #927 
- Add taproot descriptor template (BIP-86). #840

### Checklist

Release numbering must follow [Semantic Versioning]. These steps assume the current `master`
branch **development** version is *MAJOR.MINOR.0*.

#### Before publishing the release

Change the `master` branch to the next MINOR+1 version:

- [x] Switch to the `master` branch.
- [x] Create a new PR branch called `bump_dev_1.0.0-alpha.1`.
- [x] Bump the `crates` versions and README files
  - [x] `bdk` to `1.0.0-alpha.1`.
  - [x] `bdk_chain` to `0.5.0`
  - [x] `bdk_electrum` to `0.3.0`
  - [x] `bdk_esplora` to `0.3.0`
  - [x] `bdk_file_store` to `0.2.0`
- [x] Bump the `example-crates` versions and README files
    - [x] `example_cli` to `0.2.0`
    - [x] `example_electrum` to `0.2.0`
    - [x] `wallet_electrum_example` to `0.2.0`
    - [x] `wallet_esplora` to `0.2.0`
    - [x] `wallet_esplora_async` to `0.2.0`
- [x] Commit the changes with message "Bump version to 1.0.0-alpha.1".
[...]

PR #1028, opened 2023-07-06
Add CreateTxError and use as error type for TxBuilder::finish()

### LINKED ISSUE #994: `bdk::Wallet` should have context-specific error types
I agree that the errors aren't great.

However, I disagree about having a single error enum with a large number of variants. I think we should fix errors in one go with a separate PR, otherwise it's duplicate work.

_Originally posted by @evanlinjin in https://github.com/bitcoindevkit/bdk/pull/976#discussion_r1204823980_

### LINKED ISSUE #996: Wallet::get_address should return an Result
Since the persistent backend is fallible. There are `expect("TODO")` in there now which seem to have slipped through the cracks. It would be good to differentiate between those calls that are fallible and those that are infallible.
I think `AddressIndex` should go in favor of different methods that return different addresses since the error cases seem to be different.

I think with an infallible persistence backend we can have `new_address` but if it's fallible then it should be `try_new_address` and return a `Result`.

cc @notmandatory. IIRC you were thinking about how to redesign this.

notmandatory:
### Description

To remove some places where there were `.expect("TODO")` I added a new `CreateTxError` type which is returned from `TxBuilder::finish()`. I also updated related tests and doc tests.

Fixes https://github.com/bitcoindevkit/bdk/issues/996#issuecomment-1621036206

Also added fallible `Wallet::try_get_address()` and `try_get_internal_address()`  to return `Result` with a possible `D:WriteError` when a PersistBackend is used. This should fix bitcoindevkit/bdk#996.

I removed catch-all bdk::Error and replaced usages with new types and updated related functions, fixes bitcoindevkit/bdk#994.

### Notes to the reviewers

~~I didn't add all possible bdk::Error types that `Wallet::create_tx()` and `TxBuilder::finish()` functions might throw. It's probably not too much more work but will take a bit more research so I want to make sure this is the right general approach first.~~

I added `anyhow` to the dev-dependencies so I could remove some `.expect()` lines from the docs tests and make the examples closer to what an end user should do.  I also used the `anyhow!()` macro to replace a few places that were using the `bdk::Error::Generic` in example code.

I also moved the module level error.rs file to wallet/error.rs so no one would be tempted to make any new catch all errors and to make it clear that all the errors in it are wallet module related. 

### Changelog notice

Changed

- Updated bdk module to use new context specific error types
  - wallet: MiniscriptPsbtError, CreateTxError, BuildFeeBumpError error enums
  - coin_selection: module Error enum
- Renamed fallible Wallet address functions to try_get_address() and try_get_internal_address()

Removed

- Removed catch-all top level bdk::Error enum
- Removed impl_error macro

Added

-  Added infallible Wallet get_address(), get_internal_address functions

#### All Submissions:

#### Bugfixes:


crates/bdk/src/wallet/mod.rs
```diff
@@ -200,6 +201,61 @@ pub enum InsertTxError {
...
+#[derive(Debug)]
+/// Error returned from [`TxBuilder::finish`]
+pub enum CreateTxError<P> {
+    /// There was a problem with the descriptors passed in
+    Descriptor(DescriptorError),
+    /// We were unable to write wallet data to the persistence backend
+    Persist(P),
+    /// There was a problem while extracting and manipulating policies
+    Policy(PolicyError),
+    /// TODO: replace this with specific error types
```
evanlinjin: I think it will be a good idea to remove the `CreateTxError::Bdk` branch in this PR as well. I think removing this will improve the API significantly (since the caller won't have to handle a large list of redundant error variants).

This can be our opportunity to avoid using `Error::Generic` and have "concrete" variants for timelock/tx-version errors.

If needed, we can have `CreateTxError` be a variant of `bdk::Error`, although I would avoid this if possible.

evanlinjin (changes requested):
Thank you for taking this on!

I appreciate your attempt to be conservative in the amount of changes made. However, the current state of this PR does not result in a better API since the caller still has to handle a large list of redundant error variants.

If it was a PR to purely remove the `.expect("TODO")`s, I would just return a `bdk::Error::Generic` or introduce a new variant in `bdk::Error`.

However, I would prefer this PR to fix the error-handling API for tx creation completely (as it seems you already have done the majority of the work).

evanlinjin:
But we probably just need a quick fix for the next release.

crates/bdk/src/wallet/mod.rs
notmandatory: If you like the general approach of adding a new `CreateTxError` enum but would rather that I replace the `Bbk` variant with the specific errors for this PR then I'm happy to do that. I just wanted to make sure I was on the right track before chasing down all the required variants.

notmandatory: I should be able to get this knocked out today or tomorrow so we don't delay the alpha.1 too much more.

notmandatory:
I added fallible `Wallet::try_get_address()` and `try_get_internal_address()`  to return `Result` with a possible `D:WriteError` when a PersistBackend is used. This should fix bitcoindevkit/bdk#996.

notmandatory:
I removed the `CreateTxError::Bdk` variant and moved all the needed variants from the `bdk::Error` to the `CreateTxError` enum, plus created concrete variants for any errors that were using `Generic`. A few other things I'd like to do, but don't need to be in this PR:

* add `BuildFeeBumpError` enum for `Wallet::build_fee_bump` 
* add `SignError` enum for `Wallet::sign`
* add `FinalizeError` enum for `Wallet::finalize` 
* eliminate (or greatly diminish) `bdk::Error`
* move the wallet error enums and related code to a wallet::error file
* add fallible and infallible  versions of any others Wallet functions it makes sense for

notmandatory:
Per comment: https://github.com/bitcoindevkit/bdk/issues/1025#issuecomment-1635184287

I'm moving this to future alpha release.

crates/bdk/src/wallet/mod.rs
```diff
@@ -159,14 +160,32 @@ impl Wallet {
             NewError::Persist(_) => unreachable!("no persistence so it can't fail"),
         })
     }
+
+    /// Infallibly return a derived address using the external descriptor, see [`AddressIndex`] for
+    /// available address index selection strategies. If none of the keys in the descriptor are derivable
+    /// (i.e. does not end with /*) then the same address will always be returned for any [`AddressIndex`].
+    pub fn get_address(&mut self, address_index: AddressIndex) -> AddressInfo {
+        self.try_get_address(address_index).unwrap()
+    }
```
LLFourn, line 271: This is implemented on `Wallet<()>` but it should be implemented on `Wallet<D> where D: PersistBackend<WriteError=core::convert::Infallible>`

example-crates/wallet_electrum/src/main.rs
```diff
@@ -26,7 +26,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         Network::Testnet,
     )?;
 
-    let address = wallet.get_address(bdk::wallet::AddressIndex::New);
+    let address = wallet
+        .try_get_address(bdk::wallet::AddressIndex::New)
+        .expect("new_address");
```
LLFourn: You should not panic when this fails. It indicates that persistence failed. A good way to do this might be to use `anyhow::Context::context` and then `?` the error. It doesn't look like the examples depend on `anyhow` but they should and the main function should return `anyohw::Result`.

LLFourn (dismissed):
+1 on adding `CreateTxError`. left some comments.

nondiremanuel:
@notmandatory we didn't talk about this PR in the last calls. Is it correct to keep it into alpha.3 milestone? Shall we discuss it in the next call?

notmandatory:
> @notmandatory we didn't talk about this PR in the last calls. Is it correct to keep it into alpha.3 milestone? Shall we discuss it in the next call?

I don't think we need to discuss it until after alpha.2 is released. Then I'll have to rebase it and fix a few small things and will bug folks for a re-review.

notmandatory:
I've rebased and now need to do fixes suggested in comments.

crates/bdk/src/wallet/mod.rs
notmandatory: If I do this it looks like I also have to add a `db: D` parameter to `new_no_persist()`, and then all callers (tests for now) need to specify a `db: ()`. Is there scenario you're thinking of where D is some other `Infallible` database?

notmandatory:
This should also fix bitcoindevkit/bdk#994.  I'm in the process now of removing wallet::Error enum and replacing with specific errors. 

~~I'm also adding `thiserror` crate for less boilerplate error code.~~ We can't use `thiserror` because it isn't supported with no_std. :-(

notmandatory:
Moved this back to draft since still have more work to do before ready for a re-review.

.github/workflows/cont_integration.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         rust:
           - version: stable
             clippy: true
-          - version: 1.57.0 # MSRV
+          - version: 1.63.0 # MSRV
```
notmandatory, line 15: This MSRV bump should be in a separate PR once we get the go-ahead from the LDK team.

crates/bdk/src/wallet/mod.rs
notmandatory: So this took a bit longer than a day or two to finally wrap up, but should be all done now. 😅

example-crates/wallet_electrum/src/main.rs
notmandatory: I added `anyhow` and removed the `.expect()` but didn't add the `.context()` because I don't think it's needed for this simple example.

notmandatory:
This is ready for a re-review with the caveat that after CI fixes are made on `master` branch I'll need to rebase and remove the CI related commits I added here.

.github/workflows/cont_integration.yml
realeinherjar: Just a heads-up and to link the PRs.
This depends on #1183

crates/bdk/Cargo.toml
```diff
@@ -45,8 +45,9 @@ dev-getrandom-wasm = ["getrandom/js"]
 
 [dev-dependencies]
 lazy_static = "1.4"
-env_logger = "0.7"
+env_logger = { version = "0.7", default-features = false }
```
realeinherjar: I am a little bit picky with changes to `Cargo.toml`.
I understand the `anyhow` addition, but why changing `env_logger` to `default-features = false`?
And I could not find this documented in the PR.

crates/bdk/src/descriptor/error.rs
```diff
@@ -10,7 +10,7 @@
 // licenses.
 
 //! Descriptor errors
-
+#[cfg(feature = "std")]
```
realeinherjar: What is the impact for `no-std`?
What would be the returned errors in `no-std`?

notmandatory: Good question, I config'd out `fmt::Display` for errors here for consistence since it was done in other parts of the bdk package. Technically I think we could leave it and it wouldn't break `no_std` code but I believe it would make the final binary larger which could be a problem for some `no_std` embedded devices.  

@afilini any thoughts on if we should or should not include `fmt::Display` impls for errors with `no_std`?

evanlinjin: I don't think it will hurt to impl Display for `no_std`. Also curious on the reasoning.

notmandatory: To answer @realeinherjar there would still be returned error enums but just not formatted .to_string() or in a display string. 

I also found this list of suggestions for making small `no_std` binaries, including the recommendation to not use `core::fmt`. Looks like it only applies if someone is using `bdk` and needs an executable smaller than 20 kilobytes.

[Remove core::fmt with #![no_main] and Careful Usage of libstd](https://github.com/johnthagen/min-sized-rust#remove-corefmt-with-no_main-and-careful-usage-of-libstd)

.github/workflows/cont_integration.yml
notmandatory: Yup I'll remove these commits and rebase on #1183 after it's merged.

crates/bdk/Cargo.toml
notmandatory: I originally changed this to remove one additional MSRV pinning, but was probably being a bit to aggressive so I put it back as it was. I also put it back in #1183.

crates/bdk/src/descriptor/error.rs
notmandatory: I removed the `no_std` cfgs I added on `fmt:Display`. This change would have to be done for all bdk crates for it to have any effect on `no_std` binary size and I don't think are blocked by this. I created a new issue #1198 targeted for the 2.0 milestone.

crates/bdk/src/wallet/mod.rs
notmandatory: @LLFourn I'd like to wrap up this PR, but want to make sure you're OK leaving this as it is. I don't see any reason to pass in a `db:D` parameter for `new_no_persist()` when it'd always be `()`.

notmandatory:
I've removed the cfg for fmt::Display and created a new issue for that change (if we want to do it). Tests passing and ready for re-review.

crates/bdk/examples/mnemonic_to_descriptors.rs
ValuedMammal, line 1: unrelated to this PR, but the derivation path should correspond to the network, e.g. `m/86h/1h/0h/0`

crates/bdk/src/wallet/mod.rs
```diff
@@ -1396,15 +1442,16 @@ impl<D> Wallet<D> {
     /// };
     /// let  finalized = wallet.sign(&mut psbt, SignOptions::default())?;
```
ValuedMammal: nit: whitespace
```suggestion
    /// let finalized = wallet.sign(&mut psbt, SignOptions::default())?;
```

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -540,7 +554,7 @@ impl<'a, D, Cs: CoinSelectionAlgorithm, Ctx: TxBuilderContext> TxBuilder<'a, D,
     /// Returns the [`BIP174`] "PSBT" and summary details about the transaction.
```
ValuedMammal:
```suggestion
    /// Returns a new [`Psbt`] per [BIP174].
    ///
    /// [BIP174]: https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki
```

crates/bdk/src/wallet/error.rs
```diff
@@ -0,0 +1,407 @@
...
+            ),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for AddUtxoError {}
+
+#[derive(Debug)]
+/// Error returned from [`TxBuilder::add_utxo`] and [`TxBuilder::add_utxos`]
```
ValuedMammal:
```suggestion
/// Error returned from [`TxBuilder::add_foreign_utxo`]
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -1416,7 +1463,7 @@ impl<D> Wallet<D> {
                 .filter(|i| i.tap_internal_key.is_none() && i.tap_merkle_root.is_none())
                 .any(|i| i.non_witness_utxo.is_none())
         {
-            return Err(Error::Signer(signer::SignerError::MissingNonWitnessUtxo));
+            return Err(SignError::Signer(SignerError::MissingNonWitnessUtxo));
```
ValuedMammal: Maybe `Wallet::sign` and `Wallet::finalize_psbt` can return `Result<bool, signer::SignerError>` to avoid having to wrap it in a new type.. but that might require `signer::SignerError` to absorb an extra variant `MiniscriptPsbt`

crates/bdk/src/wallet/mod.rs
```diff
@@ -38,6 +38,7 @@ use bitcoin::{
 };
 use core::fmt;
 use core::ops::Deref;
+use descriptor::error::Error as DescriptorError;
 use miniscript::psbt::{PsbtExt, PsbtInputExt, PsbtInputSatisfier};
 
 use bdk_chain::tx_graph::CalculateFeeError;
```
ValuedMammal, line 44: should `CalculateFeeError` also implement std error so we don't have to call `expect` on `Wallet::calculate_fee`?

ValuedMammal (commented):
I made minor comments, but nothing that should block this from moving forward

crates/bdk/src/wallet/mod.rs
notmandatory: Good idea! fixed.

crates/bdk/src/wallet/mod.rs
notmandatory: Makes sense, I added the implementation.

crates/bdk/examples/mnemonic_to_descriptors.rs
notmandatory: I created #1209 to check this in all our docs and examples.

notmandatory:
I pushed a couple more commits to add changes suggested by @ValuedMammal. I'll squash commits to clean up the history prior to merging.

matthiasdebernardini:
Hey! This needs fixing too, "not enough" 

https://github.com/bitcoindevkit/bdk/blob/d6a0cf0795d9f8a1550e614e96b339a18fc537dc/crates/bdk/src/descriptor/policy.rs#L524

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -100,14 +100,16 @@
...
 use crate::WeightedUtxo;
-use crate::{error::Error, Utxo};
 
 use alloc::vec::Vec;
 use bitcoin::consensus::encode::serialize;
 use bitcoin::{Script, Weight};
 
 use core::convert::TryInto;
+use core::fmt;
+use core::fmt::Formatter;
```
vladimirfomene: Nits `use core::fmt::{self, Formatter};`

crates/bdk/src/wallet/mod.rs
```diff
@@ -67,11 +70,11 @@ use crate::descriptor::{
     calc_checksum, into_wallet_descriptor_checked, DerivedDescriptor, DescriptorMeta,
```
vladimirfomene: nit: add `self` on the import here to avoid importing on line 61

crates/bdk/src/wallet/mod.rs
```diff
@@ -224,6 +227,24 @@ impl Wallet {
...
     }
+
+    /// Infallibly return a derived address using the external descriptor, see [`AddressIndex`] for
+    /// available address index selection strategies. If none of the keys in the descriptor are derivable
+    /// (i.e. does not end with /*) then the same address will always be returned for any [`AddressIndex`].
+    pub fn get_address(&mut self, address_index: AddressIndex) -> AddressInfo {
+        self.try_get_address(address_index).unwrap()
+    }
+
+    /// Infallibly return a derived address using the internal (change) descriptor.
```
vladimirfomene, line 273: I'm curious to know why we needed to create infallible version of the address functions.

vladimirfomene (approved):
Just nits. LGTM. ACK 35a108b4300473277e6f575a8f781add5d555a48

crates/bdk/src/wallet/mod.rs
notmandatory: I think the only reason for this is so that non-persistent / memory only applications don't need to handle a persistence error which will never happen. This is probably a rare case but I suspect there are some embedded system applications where this would be useful.

evanlinjin: This is an excellent addition, thank you @notmandatory.

evanlinjin:
I don't think this should be addressed in this PR, but I think we should use idiomatic naming for methods: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

I.e. `Wallet::get_address` should be `Wallet::address`.

evanlinjin (approved):
ACK 35a108b4300473277e6f575a8f781add5d555a48

Thank you for this work. Now the errors are more domain-specific and detailed.

crates/bdk/src/wallet/mod.rs
danielabrozzoni: I think Lloyd meant that you should keep `new_no_persist` in `impl Wallet`, but move the other two methods in `impl Wallet<D: ...>`:
```diff
diff --git a/crates/bdk/src/wallet/mod.rs b/crates/bdk/src/wallet/mod.rs
index 051f9730..061a3db5 100644
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -226,7 +226,12 @@ impl Wallet {
             NewError::Persist(_) => unreachable!("no persistence so it can't fail"),
         })
     }
+}
 
+impl<D> Wallet<D>
+where
+    D: PersistBackend<ChangeSet, WriteError = core::convert::Infallible>,
+{
     /// Infallibly return a derived address using the external descriptor, see [`AddressIndex`] for
     /// available address index selection strategies. If none of the keys in the descriptor are derivable
     /// (i.e. does not end with /*) then the same address will always be returned for any [`AddressIndex`].

```

To me, it makes sense to have `get_address` and `get_internal_address` for all the wallets that have infallible backends (which could be either `()` or some other database)

danielabrozzoni (commented):
Fantastic work, than you.

I think Lloyd's comment should be fixed, as it should be fairly trivial. I left a comment on that.

notmandatory:
> I think Lloyd's comment should be fixed, as it should be fairly trivial. I left a comment on that.

Ohhhh! 🤦‍♂️ thanks for the help here, now I get it. I'm going to push your fix, rebase, clean up the history a little, and then it should be ready to merge.

crates/bdk/src/wallet/mod.rs
notmandatory: Fixed!

PR #1031, opened 2023-07-14
remove duplicate `is_empty`

vladimirfomene:
Duplicate `is_empty` method

#### All Submissions:

### DIFF
--- a/crates/chain/src/keychain.rs
+++ b/crates/chain/src/keychain.rs
@@ -46,11 +46,6 @@ pub use txout_index::*;
 pub struct DerivationAdditions<K>(pub BTreeMap<K, u32>);
 
 impl<K> DerivationAdditions<K> {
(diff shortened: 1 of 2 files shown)


crates/chain/src/keychain.rs
```diff
@@ -72,6 +67,7 @@ impl<K: Ord> Append for DerivationAdditions<K> {
         self.0.append(&mut other.0);
     }
 
+    /// Returns whether the additions are empty.
     fn is_empty(&self) -> bool {
```
benthecarman, line 71: The other one was pub, this one is not

vladimirfomene: I saw that but also saw that the Append trait is public.

danielabrozzoni (approved):
ACK f5074ee3aee319137db1ddfa1e99f7e111abcfd4

Issue #1022, opened 2023-06-28
Unify changeset APIs between all structures


LLFourn:
- Each structure has a different name for its changeset. For example `TxGraph` it's called `Additions` while IndexedTxGraph is called `IndexedAdditions`. Rather they should have the same name and users should just use different module prefixes e.g. `tx_graph::Additions` or `indexed_tx_graph::Additions`.
- Since these are monotone they don't need a "preview" methods. You can always just apply the update right away. I'd also include `LocalChain` in this policy despite it not being monotone ([yet](#1005)).

evanlinjin:
Since we are already on this, can I request for some more renamings?

- [ ] `CanonicalTx::observed_as` should be renamed to `CanonicalTx::chain_position`.
- [ ] `CanonicalTx::node` should be renamed to `CanonicalTx::tx_node`.

I think this will provide more clarity.

Thanks @danielabrozzoni

danielabrozzoni:
bitcoindevkit/bdk#1065 fixed this

PR #1034, opened 2023-07-19
Implement linked-list `LocalChain` and update chain-src crates/examples

### LINKED ISSUE #997: LocalChain should be implemented as a cheaply cloneable singly linked list

evanlinjin:
Fixes #997 
Replaces #1002 

### Description

This PR changes the `LocalChain` implementation to have blocks stored as a linked-list. This allows the data-src thread to hold a shared ref to a single checkpoint and have access to the whole history of checkpoints without cloning or keeping a lock on `LocalChain`.

The APIs of `bdk::Wallet`, `esplora` and `electrum` are also updated to reflect these changes. Note that the `esplora` crate is rewritten to anchor txs in the confirmation block (using the esplora API's tx status block_hash). This guarantees 100% consistency between anchor blocks and their transactions (instead of anchoring txs to the latest tip). `ExploraExt` now has separate methods for updating the `TxGraph` and `LocalChain`.

A new method `TxGraph::missing_blocks` is introduced for finding "floating anchors" of a `TxGraph` update (given a chain).

Additional changes:

* `test_local_chain.rs` is refactored to make test cases easier to write. Additional tests are also added.
* Examples are updated.
* Exclude example-cli `*.db` files in `.gitignore`.
* Rm duplicate `bdk_tmp_plan` module.

### Notes to the reviewers

This is the smallest possible division of #1002 without resulting in PRs that do not compile. Since we have changed the API of `LocalChain`, we also need to change `esplora`, `electrum` crates and examples alongside `bdk::Wallet`.

### Changelog notice

* Implement linked-list `LocalChain`. This allows the data-src thread to hold a shared ref to a single checkpoint and have access to the whole history of checkpoints without cloning or keeping a lock on `LocalChain`.
* Rewrote `esplora` chain-src crate to anchor txs to their confirmation blocks (using esplora API's tx-status `block_hash`).

#### All Submissions:

#### New Features:


crates/bdk/Cargo.toml
```diff
@@ -47,7 +47,7 @@ dev-getrandom-wasm = ["getrandom/js"]
 lazy_static = "1.4"
 env_logger = "0.7"
 # Move back to importing from rust-bitcoin once https://github.com/rust-bitcoin/rust-bitcoin/pull/1342 is released
-base64 = "^0.13"
+base64 = "^0.21"
```
LLFourn: what happened here.

crates/chain/src/local_chain.rs
```diff
@@ -19,215 +164,271 @@ impl ChainOracle for LocalChain {
...
+    pub fn from_changeset(changeset: ChangeSet) -> Self {
+        let mut chain = Self::default();
+        chain.apply_changeset(&changeset);
 
-    /// Get a reference to a map of block height to hash.
-    pub fn blocks(&self) -> &BTreeMap<u32, BlockHash> {
-        &self.blocks
-    }
+        #[cfg(debug_assertions)]
+        chain._check_consistency(Some(&changeset));
```
LLFourn: Why not use `debug_assert!` here so avoid this `cfg` thingy.

crates/chain/src/local_chain.rs
```diff
@@ -19,215 +164,271 @@ impl ChainOracle for LocalChain {
...
+        self.tip.clone()
     }
 
-    /// Updates [`LocalChain`] with an update [`LocalChain`].
+    /// Returns whether the [`LocalChain`] is empty (has no checkpoints).
+    pub fn is_empty(&self) -> bool {
+        self.tip.is_none()
+    }
+
+    /// Updates [`Self`] with the given `update_tip`.
```
LLFourn: These docs are stale. Also this should be `self` and not `[Self]` It doesn't update the type it updates the instance of the type which is `self`. Also it's preferable to say "Updates the chain..." rather than updates `self`".

crates/chain/src/local_chain.rs
```diff
@@ -19,215 +164,271 @@ impl ChainOracle for LocalChain {
...
+    /// be true.
     ///
-    /// [`determine_changeset`]: Self::determine_changeset
-    /// [`apply_changeset`]: Self::apply_changeset
-    pub fn apply_update(&mut self, update: Self) -> Result<ChangeSet, UpdateNotConnectedError> {
-        let changeset = self.determine_changeset(&update)?;
-        self.apply_changeset(changeset.clone());
-        Ok(changeset)
+    /// The method returns [`ChangeSet`] on success. This represents the applied changes to
+    /// [`Self`].
```
LLFourn: again using `[Self]`.

crates/chain/src/tx_graph.rs
```diff
@@ -598,6 +598,31 @@ impl<A: Clone + Ord> TxGraph<A> {
...
+    /// Find missing block heights of `chain`.
+    ///
+    /// This works by scanning through anchors, and seeing whether the anchor block of the anchor
+    /// exists in the [`LocalChain`].
+    pub fn missing_blocks<'a>(&'a self, chain: &'a LocalChain) -> impl Iterator<Item = u32> + 'a {
+        self.anchors
+            .iter()
+            .map(|(a, _)| a.anchor_block())
+            .filter({
+                let mut last_block = Option::<BlockId>::None;
```
LLFourn: nit: declare this variable before the statement to make it clear the filter has state.

crates/chain/src/keychain.rs
```diff
@@ -89,24 +86,32 @@ impl<K> AsRef<BTreeMap<K, u32>> for DerivationAdditions<K> {
...
-/// A structure to update [`KeychainTxOutIndex`], [`TxGraph`] and [`LocalChain`]
-/// atomically.
-#[derive(Debug, Clone, PartialEq)]
+/// A structure to update [`KeychainTxOutIndex`], [`TxGraph`] and [`LocalChain`] atomically.
+///
+/// [`LocalChain`]: local_chain::LocalChain
+#[derive(Debug, Clone)]
 pub struct LocalUpdate<K, A> {
     /// Last active derivation index per keychain (`K`).
     pub keychain: BTreeMap<K, u32>,
```
vladimirfomene: nit: maybe change this field to `keychains` or `last_active_indices`. Maybe this can be pushed to another PR.

crates/chain/src/tx_graph.rs
```diff
@@ -598,6 +598,31 @@ impl<A: Clone + Ord> TxGraph<A> {
...
+                        last_block = Some(*block);
+                        true
+                    }
+                }
+            })
+            .filter_map(|block| match chain.heights().get(&block.height) {
+                Some(chain_hash) if *chain_hash == block.hash => None,
+                _ => Some(block.height),
+            })
+    }
```
LLFourn, line 662: I recall in the previous PR that this function was implemented incorrectly and I requested tests be written. I re-requesting this. I still believe this function is wrong. it needs tests for:

1. Two transactions with anchors the same anchor which is missing from the chain: the missing block appears once in the iterator.
2. Two transactions with *different* anchors at the same height: the missing block appears once in the chain.
3. One transaction with two anchors at the same height. The chain doesn't include this height. The iterator should include one item.
4. One transaction with two anchors at the same height. The chain includes this height but doesn't match either anchor. The iterator should be empty.
5. One transaction with two anchors at two different heights. One of the anchor height is included in the chain with the same hash as the anchor. This should return nothing.
6. One transaction with two anchors at two different heights A and B. height  A is included in the chain with a different hash to as the anchor. The iterator should yield only B.
7. One transaction with two anchors at two different heights. Neither height is in the chain. This should yield both heights in the iterator.

The essential principle is that it should return any height that could help it determine whether one of the transactions in the graph is in the chain or not. If it can determine that already then it shouldn't return anything for that tx.

crates/chain/src/local_chain.rs
```diff
@@ -2,15 +2,160 @@
...
+    }
+}
+
+impl From<LocalChain> for BTreeMap<u32, BlockHash> {
+    fn from(value: LocalChain) -> Self {
+        value.index
+    }
+}
+
+impl From<ChangeSet> for LocalChain {
```
LLFourn: why does this exist? To me it doesn't make sense to construct a `LocalChain` from a changeset especially since the changeset is non-monotone.

If you want to do this just do

 ```rust
local_chain = changeset.into_iter().filter_map(core::convert::identity).collect::<BTreeMap<u32, BlockHash>>().into();
```

File: crates/chain/src/local_chain.rs [hunk 9]
```diff
@@ -2,15 +2,160 @@
...
+use alloc::sync::Arc;
 use bitcoin::BlockHash;
 
-use crate::{BlockId, ChainOracle};
+/// A structure that represents changes to [`LocalChain`].
+pub type ChangeSet = BTreeMap<u32, Option<BlockHash>>;
+
+/// A blockchain of [`LocalChain`].
+///
+/// The in a linked-list with newer blocks pointing to older ones.
```
vladimirfomene: Nits: phrase typo

example-crates/wallet_esplora_async/src/main.rs
```diff
@@ -53,19 +54,22 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            [],
-            STOP_GAP,
-            PARALLEL_REQUESTS,
-        )
+    let (update_graph, last_active_indices) = client
+        .update_tx_graph(keychain_spks, None, None, STOP_GAP, PARALLEL_REQUESTS)
         .await?;
-    println!();
+    let get_heights = wallet.tx_graph().missing_blocks(wallet.local_chain());
+    let new_tip = client.update_local_chain(prev_tip, get_heights).await?;
```
LLFourn: why can't this just return `local_chain::Update`?

LLFourn (changes requested):
Looks great. A few API things to touch up in my view.

crates/chain/src/local_chain.rs
```diff
@@ -19,215 +164,271 @@ impl ChainOracle for LocalChain {
...
+        CheckPointIter {
+            current: self.tip.as_ref().map(|tip| tip.0.clone()),
+        }
     }
-}
 
-#[cfg(feature = "std")]
-impl std::error::Error for UpdateNotConnectedError {}
+    /// Get a reference to the internal index mapping the height to block hash
+    pub fn heights(&self) -> &BTreeMap<u32, BlockHash> {
```
vladimirfomene: I'm thinking we should update this method name, maybe `index` or `get_index`.

crates/chain/src/tx_graph.rs
```diff
@@ -598,6 +598,31 @@ impl<A: Clone + Ord> TxGraph<A> {
...
 impl<A: Anchor> TxGraph<A> {
+    /// Find missing block heights of `chain`.
+    ///
+    /// This works by scanning through anchors, and seeing whether the anchor block of the anchor
+    /// exists in the [`LocalChain`].
+    pub fn missing_blocks<'a>(&'a self, chain: &'a LocalChain) -> impl Iterator<Item = u32> + 'a {
+        self.anchors
+            .iter()
+            .map(|(a, _)| a.anchor_block())
+            .filter({
```
vladimirfomene: is there a way to simplify this logic using the ChainOracle `is_in_best_chain` method?

crates/chain/tests/test_local_chain.rs
```diff
@@ -1,180 +1,300 @@
...
+            update: chain_update![(1, h!("B")), (2, h!("C"))],
+            exp: ExpectedResult::Ok {
+                changeset: &[(1, Some(h!("B")))],
+                init_changeset: &[(1, Some(h!("B"))), (2, Some(h!("C"))), (3, Some(h!("D")))],
+            },
+        },
+        // Introduce an older checkpoint (A) that is not directly behind PoA
+        //        | 1 | 2 | 3
+        // chain  |     B   C
+        // update | A       C
```
vladimirfomene, line 128: Diagram needs to be updated

crates/electrum/src/electrum_ext.rs
```diff
@@ -1,34 +1,46 @@
...
 };
-use electrum_client::{Client, ElectrumApi, Error};
+use electrum_client::{Client, ElectrumApi, Error, HeaderNotification};
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap, HashSet},
     fmt::Debug,
 };
 
+/// We assume that a block of this depth and deeper cannot be reorged.
+const ASSUME_FINAL_DEPTH: u32 = 8;
```
vladimirfomene, line 15: Any specific reason why we going with 8 and not 7 or 6?

crates/electrum/src/electrum_ext.rs
```diff
@@ -1,34 +1,46 @@
...
+/// the full transactions from Electrum and finalize the update.
 #[derive(Debug, Clone)]
 pub struct ElectrumUpdate<K, A> {
+    /// Map of [`Txid`]s to associated [`Anchor`]s.
     pub graph_update: HashMap<Txid, BTreeSet<A>>,
-    pub chain_update: LocalChain,
+    /// The latest chain tip, as seen by the Electrum server.
+    pub new_tip: local_chain::CheckPoint,
+    /// Last-used index update for [`KeychainTxOutIndex`](bdk_chain::keychain::KeychainTxOutIndex).
     pub keychain_update: BTreeMap<K, u32>,
```
vladimirfomene, line 29: Nits: Maybe `keychains_update`

crates/electrum/src/electrum_ext.rs
```diff
@@ -75,6 +93,7 @@ impl<K> ElectrumUpdate<K, ConfirmationHeightAnchor> {
         missing: Vec<Txid>,
     ) -> Result<LocalUpdate<K, ConfirmationTimeAnchor>, Error> {
         let update = self.finalize(client, seen_at, missing)?;
+        // client.batch_transaction_get(txid)
```
vladimirfomene: Why is this not remove completely?

crates/esplora/src/async_ext.rs
```diff
@@ -74,16 +86,119 @@ pub trait EsploraAsyncExt {
...
+                    // here unconditionally. We only break if we guarantee there are no new heights
+                    // below our current.
+                    if old_block.hash == new_hash {
+                        earliest_agreement_cp = Some(old_cp);
+
+                        let first_new_height = *new_blocks
+                            .keys()
+                            .next()
+                            .expect("must have atleast one new block");
+                        if first_new_height <= old_block.height {
```
vladimirfomene: @evanlinjin do you mind expanding on why this condition is necessary? I'm not able to understand from the comment

crates/esplora/src/async_ext.rs
```diff
@@ -92,178 +207,116 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
...
+            let handles = txids
+                .by_ref()
+                .take(parallel_requests)
+                .filter(|&txid| graph.get_tx(txid).is_none())
+                .map(|txid| {
+                    let client = self.clone();
+                    async move { client.get_tx_status(&txid).await.map(|s| (txid, s)) }
+                })
+                .collect::<FuturesOrdered<_>>();
+            // .collect::<Vec<JoinHandle<Result<(Txid, TxStatus), Error>>>>();
```
vladimirfomene: I believe, this is no longer necessary.

crates/esplora/src/lib.rs
```diff
@@ -14,16 +14,22 @@ mod async_ext;
...
-    tip_at_start: BlockId,
-) -> Option<ConfirmationTimeAnchor> {
-    match (tx_status.block_time, tx_status.block_height) {
-        (Some(confirmation_time), Some(confirmation_height)) => Some(ConfirmationTimeAnchor {
-            anchor_block: tip_at_start,
-            confirmation_height,
-            confirmation_time,
-        }),
-        _ => None,
+const ASSUME_FINAL_DEPTH: u32 = 15;
```
vladimirfomene, line 17: is this because the max block summary return by mempoo.space is 15?

vladimirfomene (commented):
Just a few nits and few questions.

crates/bdk/Cargo.toml
evanlinjin: Rm-ed.

crates/chain/src/local_chain.rs
evanlinjin: Changed in 66c6c33

crates/chain/src/local_chain.rs
evanlinjin: Rm-ed

example-crates/wallet_esplora_async/src/main.rs
evanlinjin: Good catch. Changed to return the beautiful `Update` structure as you have requested.

crates/chain/src/local_chain.rs
evanlinjin: Good catch, thank you!

crates/chain/src/local_chain.rs
evanlinjin: I don't think `index` is a better name as that can be both a verb and a noun...

crates/chain/src/tx_graph.rs
evanlinjin: I don't think that will make sense to use that here as we need a chain tip.

crates/electrum/src/electrum_ext.rs
evanlinjin: No reason

crates/esplora/src/async_ext.rs
evanlinjin: The comparison operator was actually the wrong way around so thank you for pointing out this line!

We are trying to find the checkpoint to build our update chain from. Finding a point of agreement is not enough if we wish to introduce blocks lower than the point of agreement (if we allow this, new blocks below point of agreement will NOT be included in the update).

Therefore, the rule to find our "checkpoint base" is: checkpoint must be a point of agreement && `height <= lowest_introduced_block_height`.

crates/esplora/src/lib.rs
evanlinjin: It's just a large enough number in which we can guarantee with close to full certainty that we don't have a reorg at that depth.

crates/esplora/src/async_ext.rs
evanlinjin: Rm-ed

crates/electrum/src/electrum_ext.rs
evanlinjin: It's finnneeee haha

crates/electrum/src/electrum_ext.rs
evanlinjin: Any large enough number is good enough I think.

crates/chain/src/keychain.rs
evanlinjin: Changed.

crates/esplora/src/async_ext.rs
evanlinjin: And note that anything equal or lower to `ASSUME_FINAL_DEPTH` is considered a point of agreement.

crates/chain/src/tx_graph.rs
evanlinjin: > One transaction with two anchors at two different heights A and B. height A is included in the chain with a different hash to as the anchor. The iterator should yield only B.

Isn't it possible that the transaction is confirmed at height A anyway? My intuition tells me we should still return both heights.

evanlinjin: @LLFourn I think I did it, except for the test case I don't agree with. 4a704fe700cc608a4760173be1d4a9a506a2b73a

crates/chain/src/keychain.rs
```diff
@@ -89,24 +86,33 @@ impl<K> AsRef<BTreeMap<K, u32>> for DerivationAdditions<K> {
...
-    pub chain: LocalChain,
+    ///
+    /// [`LocalChain`]: local_chain::LocalChain
+    pub chain: local_chain::Update,
 }
 
-impl<K, A> Default for LocalUpdate<K, A> {
-    fn default() -> Self {
+impl<K, A> LocalUpdate<K, A> {
+    /// Construct a [`LocalUpdate`] with a given [`CheckPoint`] tip.
```
vladimirfomene: Maybe this needs to change to `Construct a [`LocalUpdate`] with a given [`local_chain::Update`]`

vladimirfomene (approved):
Just nits:

evanlinjin:
Something that I need help with is how we should test the `bdk_esplora` crate. Should we do a full blown integration-test environment? What's the best way to spin up an esplora backend? Docker? Or should we mock up the esplora API?

vladimirfomene:
> Something that I need help with is how we should test the `bdk_esplora` crate. Should we do a full blown integration-test environment? What's the best way to spin up an esplora backend? Docker? Or should we mock up the esplora API?

Using `electrsd` in conjunction with `bitcoind` allows us to have integration test environment both for Esplora and Electrum.

crates/chain/src/tx_graph.rs
LLFourn: > > One transaction with two anchors at two different heights A and B. height A is included in the chain with a different hash to as the anchor. The iterator should yield only B.
> 
> Isn't it possible that the transaction is confirmed at height A anyway? My intuition tells me we should still return both heights.

In order for it to be in the chain blocks `A` or `B` must be in the chain. We know `A` isn't because at the height of `A` we have a block hash that conflicts with it so it must be at `B` or not at all so we need to fetch the block at the same height as `B`.

crates/chain/src/tx_graph.rs
```diff
@@ -598,6 +598,51 @@ impl<A: Clone + Ord> TxGraph<A> {
 }
 
 impl<A: Anchor> TxGraph<A> {
+    /// Find missing block heights of `chain`.
+    ///
+    /// This works by scanning through anchors, and seeing whether the anchor block of the anchor
+    /// exists in the [`LocalChain`]. The returned iterator does not output duplicate heights.
+    pub fn missing_blocks<'a>(&'a self, chain: &'a LocalChain) -> impl Iterator<Item = u32> + 'a {
```
LLFourn: Call it missing heights?

crates/chain/tests/test_tx_graph.rs
```diff
@@ -822,3 +822,127 @@ fn test_additions_last_seen_append() {
...
+            &[2],
+        ),
+        new_scenario(
+            "2 txs with different anchors at the same height, one of the anchors is missing",
+            &[
+                (h!("tx_1"), new_anchor(2, h!("B1"))),
+                (h!("tx_2"), new_anchor(2, h!("B2"))),
+            ],
+            &[(1, h!("A")), (2, h!("B1"))],
+            &[2],
```
LLFourn: I don't get it. Height `2` is already in the chain. It's not missing. Why is it returned here?

crates/chain/tests/test_tx_graph.rs
```diff
@@ -822,3 +822,127 @@ fn test_additions_last_seen_append() {
...
+            &[3],
+        ),
+        new_scenario(
+            "tx with 2 anchors at the same height, chain has this height but does not match either anchor",
+            &[
+                (h!("tx"), new_anchor(4, h!("D1"))),
+                (h!("tx"), new_anchor(4, h!("D2"))),
+            ],
+            &[(4, h!("D3")), (5, h!("E"))],
+            &[4],
```
LLFourn: Again `4` is already in the chain.

crates/chain/src/local_chain.rs
```diff
@@ -2,15 +2,166 @@
...
+pub struct CheckPoint(Arc<CPInner>);
+
+/// The internal contents of [`CheckPoint`].
+#[derive(Debug, Clone)]
+struct CPInner {
+    /// Block id (hash and height).
+    block: BlockId,
+    /// Previous checkpoint (if any).
+    prev: Option<Arc<CPInner>>,
+}
```
danielabrozzoni, line 34: I'm wondering why you chose to create CPInner and CheckPoint instead of using LinkedList<BlockId>. Is it because you want to be able to hold a single CheckPoint?

crates/chain/src/local_chain.rs
```diff
@@ -2,15 +2,166 @@
...
 
-use alloc::collections::BTreeMap;
+use crate::collections::BTreeMap;
+use crate::{BlockId, ChainOracle};
+use alloc::sync::Arc;
 use bitcoin::BlockHash;
 
-use crate::{BlockId, ChainOracle};
+/// A structure that represents changes to [`LocalChain`].
+pub type ChangeSet = BTreeMap<u32, Option<BlockHash>>;
```
danielabrozzoni, line 14: Am I right in saying that if the changeset contains, for example, `(32, None)`, this would mean "the block with height 32 doesn't exist anymore"?

crates/chain/src/local_chain.rs
```diff
@@ -19,215 +170,250 @@ impl ChainOracle for LocalChain {
...
+        chain
     }
-}
 
-impl From<BTreeMap<u32, BlockHash>> for LocalChain {
-    fn from(value: BTreeMap<u32, BlockHash>) -> Self {
-        Self { blocks: value }
+    /// Construct a [`LocalChain`] from a given `checkpoint` tip.
+    pub fn from_tip(tip: CheckPoint) -> Self {
+        let mut _self = Self {
```
danielabrozzoni: nit: this should be `self` as it is used below

crates/chain/src/local_chain.rs
```diff
@@ -19,215 +170,250 @@ impl ChainOracle for LocalChain {
...
-        // the first block's height to invalidate in the local chain
-        let invalidate_from_height = self.blocks.range(invalidate_lb..).next().map(|(&h, _)| h);
+    /// Applies the given `update` to the chain.
+    ///
+    /// The method returns [`ChangeSet`] on success. This represents the applied changes to `self`.
+    ///
+    /// To update, the `update_tip` must *connect* with `self`. If `self` and `update_tip` has a
+    /// mutual checkpoint (same height and hash), it can connect if:
+    /// * The mutual checkpoint is the tip of `self`.
+    /// * An ancestor of `update_tip` has a height which is of the checkpoint one higher than the
```
danielabrozzoni: I'm really lost here. Does this mean that if I want to remove a block from `self`, then I have to provide another block with the same height?

Let's say self is `1 -> 2 -> 3 -> 4` and the update is `3 -> 4' -> 5'`. As far as I understand, this is valid (I'm removing block `4` and replacing it with `4'`). Instead, if self was `1 -> 2 -> 3 -> 4` and the update was `3 -> 5'`, this wouldn't be valid, as I'm removing block 4 without giving a replacement. Does this make sense?

crates/chain/src/local_chain.rs
```diff
@@ -2,15 +2,166 @@
...
+    pub introduce_older_blocks: bool,
+}
 
 /// This is a local implementation of [`ChainOracle`].
-#[derive(Debug, Default, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Default, Clone)]
 pub struct LocalChain {
-    blocks: BTreeMap<u32, BlockHash>,
+    tip: Option<CheckPoint>,
+    index: BTreeMap<u32, BlockHash>,
```
danielabrozzoni, line 150: I don't understand what's the logic behind the tip checkpoint and the index. Does the index contain all the blocks we know, and the checkpoint contains only some of them? If so, how do you chose which blocks are going to be checkpoints?

crates/chain/src/local_chain.rs
```diff
@@ -19,215 +170,250 @@ impl ChainOracle for LocalChain {
...
+    }
 
-impl core::fmt::Display for UpdateNotConnectedError {
-    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(
-            f,
-            "the update cannot connect with the chain, try include block at height {}",
-            self.0
-        )
+    /// Iterate over checkpoints in decending height order.
```
danielabrozzoni: nit: descending

crates/chain/src/local_chain.rs
```diff
@@ -19,215 +170,250 @@ impl ChainOracle for LocalChain {
...
+    pub fn iter_checkpoints(&self) -> CheckPointIter {
```
danielabrozzoni, line 384: Do you think it makes sense to have the method name reflect the fact that the checkpoints are iterated backwards, from highest to lowest height?

crates/chain/src/local_chain.rs
danielabrozzoni: What about `blocks`? I find `heights` to be weird, it's not returning just heights, block hashes as well

danielabrozzoni (commented):
I reviewed the LocalChain code and left a few questions. 

No need to wait for my complete review before merging, I don't think I can provide a meaningful review or a sensible ACK

crates/chain/src/local_chain.rs
LLFourn: Correct.

crates/chain/src/local_chain.rs
LLFourn: Do you mean why not use `LinkedList` from `std::collections`? The `clone` operation for that one is `O(n)` while this one is `O(1)`. The point of this `CheckPoint` is that you can clone it cheaply and pass it off to a blockchain source.

`CPInner` is needed to avoid exposing `Arc` to the user I think. It should just be called `CheckPointInner`.

crates/chain/src/local_chain.rs
```diff
@@ -2,15 +2,166 @@
...
+/// A structure that represents changes to [`LocalChain`].
+pub type ChangeSet = BTreeMap<u32, Option<BlockHash>>;
+
+/// A [`LocalChain`] checkpoint is used to find the agreement point between two chains and as a
+/// transaction anchor.
+///
+/// Each checkpoint contains the height and hash of a block ([`BlockId`]).
+///
+/// Internaly, checkpoints are nodes of a linked-list. This allows the caller to view the entire
+/// chain without holding a lock to [`LocalChain`].
```
LLFourn: Also mention that you can cheaply clone a `CheckPoint` without copying the whole list.

s/linked-list/reference counted linked-list/

crates/chain/src/local_chain.rs
```diff
@@ -2,15 +2,166 @@
...
+        } else {
+            Err(self)
+        }
+    }
+
+    /// Extends the checkpoint linked list by a iterator of block ids.
+    ///
+    /// Returns an `Err(self)` if there is block which does not have a greater height than the
+    /// previous one.
+    pub fn extend_with_blocks(
```
LLFourn: There is `extend_with_blocks` and `push`. Either choose `push_block` and `extend_with_blocks` or just `push` and `extend`.

crates/chain/src/local_chain.rs
```diff
@@ -2,15 +2,166 @@
...
+        CheckPointIter {
+            current: Some(self.0),
+        }
+    }
+}
+
+/// A struct to update [`LocalChain`].
+///
+/// This is used as input for [`LocalChain::apply_update`]. It contains the update's chain `tip` and
+/// a `bool` which signals whether this update can introduce blocks below the original chain's tip
```
LLFourn: referencing the `bool` type literally is not super helpful. Consider instead:

> It contains the update's chain `tip` and a flag `introduce_older_blocks` which signals whether this update intends to introduce missing blocks in the original chain.

crates/chain/src/local_chain.rs
LLFourn: The index indexes what's in `tip` so you can quickly query it for a certain height. The blocks in the index are the same as in `tip`.

crates/chain/src/local_chain.rs
LLFourn: I don't think you can assign to `self`.

crates/chain/src/local_chain.rs
```diff
@@ -19,215 +170,250 @@ impl ChainOracle for LocalChain {
...
-
-        // the lower bound of the range to invalidate
-        let invalidate_lb = match agreement_height {
-            Some(height) if height == update_tip => u32::MAX,
-            Some(height) => height + 1,
-            None => 0,
-        };
+    /// Returns whether the [`LocalChain`] is empty (has no checkpoints).
+    pub fn is_empty(&self) -> bool {
+        self.tip.is_none()
```
LLFourn:
```suggestion
        let res = self.tip.is_none();
        debug_assert_eq!(res, self.index.is_empty());
        res
```

crates/chain/src/local_chain.rs
LLFourn: Correct. I wonder if this documentation can be improved by writing it less like code and explain what it means to connect. I think the concept is easier to understand if we say 

> There must be no ambiguity about which of the existing chain's blocks are still valid and which are now invalid. That is, the new chain must implicitly  connect to a definite block in the existing chain and invalidate the block after it (if it exists) by including a block at the same height but with a different hash to explicitly exclude it as a connection point.

crates/chain/src/local_chain.rs
evanlinjin: `CPInner` means "checkpoint inner" and it's there purely to hide the `Arc`. It's still a linked list as `prev` holds a pointer to the previous node.

crates/chain/src/local_chain.rs
evanlinjin: Correct!

crates/chain/src/local_chain.rs
evanlinjin: I don't think that's legal. We might get arrested by the rustacean police.

crates/chain/src/local_chain.rs
danielabrozzoni: Of course, I'm just wondering why you're not using the std LinkedList: https://doc.rust-lang.org/std/collections/struct.LinkedList.html

crates/chain/src/local_chain.rs
evanlinjin: You say that you are lost, but prove that you are not lost. Please provide guidance on how we can word this better.

crates/chain/src/local_chain.rs
evanlinjin: Grazie mille

crates/chain/src/local_chain.rs
evanlinjin: I think that will provide better clarity. Do you have any suggestions? @LLFourn what are your thoughts?

crates/chain/src/local_chain.rs
evanlinjin: Yeah we probably need to include this in the docs.

crates/chain/src/local_chain.rs
evanlinjin: One requirement we have is that each node should be both `Sync + Send`. This way, the update thread can have to total view of the chain without keeping a lock on `LocalChain`.

Also, we only need a singly linked-list so `LinkedList` might be overkill.

crates/chain/src/local_chain.rs
evanlinjin: The index indexes the total history that is available from `tip`.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1697,24 +1697,25 @@ impl<D> Wallet<D> {
...
-    ///
-    /// This returns whether the `update` resulted in any changes.
+    /// Returns whether the `update` resulted in any changes.
     ///
     /// Usually you create an `update` by interacting with some blockchain data source and inserting
     /// transactions related to your wallet into it.
     ///
     /// [`commit`]: Self::commit
-    pub fn apply_update(&mut self, update: Update) -> Result<bool, UpdateNotConnectedError>
+    pub fn apply_update(&mut self, update: Update) -> Result<bool, CannotConnectError>
```
LLFourn: This bool to know whether something changed doesn't really need to exist I think.

crates/chain/src/local_chain.rs
danielabrozzoni: I can do that as a part of #1022

example-crates/wallet_esplora_async/src/main.rs
```diff
@@ -53,19 +54,19 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            [],
-            [],
-            STOP_GAP,
-            PARALLEL_REQUESTS,
-        )
+    let (update_graph, last_active_indices) = client
+        .update_tx_graph(keychain_spks, None, None, STOP_GAP, PARALLEL_REQUESTS)
         .await?;
-    println!();
+    let get_heights = wallet.tx_graph().missing_heights(wallet.local_chain());
```
LLFourn:
```suggestion
    let missing_heights = wallet.tx_graph().missing_heights(wallet.local_chain());
```

crates/chain/src/local_chain.rs
evanlinjin: No need, I'll do it here. Probably need to rebase anyway for the other suggested changes.

example-crates/example_electrum/src/main.rs
```diff
@@ -296,7 +278,9 @@ fn main() -> anyhow::Result<()> {
 
         let indexed_additions = {
             let mut additions = IndexedAdditions::<ConfirmationHeightAnchor, _>::default();
-            let (_, index_additions) = graph.index.reveal_to_target_multi(&final_update.keychain);
+            let (_, index_additions) = graph
+                .index
+                .reveal_to_target_multi(&final_update.last_active_indices);
```
LLFourn, line 283: This is not specifically to do with this PR. I think we are making suboptimal use of our API. Only in `scan` do we need to do this work so it should only occur in that branch to make that clear to the user. The changes can be staged there and only committed out here.

LLFourn (changes requested):
Thanks. Made some mostly nits and documentation suggestions.

example-crates/example_electrum/src/main.rs
evanlinjin: With sync, it's still possible to find a transaction that will increase last active indices right?

crates/chain/src/local_chain.rs
LLFourn: Delete the method. You can just get the tip and iter backwards right?

[edit] oh but I see that since the tip in an `Option` that's annoying. I think documenting the order is fine for me.

example-crates/example_electrum/src/main.rs
LLFourn: No. 

[edit] ok yes. But the point is we never need to do `reveal_to_target_multi` in `sync`. I want to get across that `reveal_to_target_multi` is a specialized operation that you won't be doing during normal wallet operation.

LLFourn:
ACK b206a985cffaa9b614841219371faa53ba23dfc3

example-crates/example_electrum/src/main.rs
evanlinjin: @LLFourn are you saying we should depend on `lookahead` + scanning in the transactions?

LLFourn: In sync you don't need to "look ahead" you are only scanning the addresses you've already revealed.

LLFourn (approved):
Sorry forgot to approve

PR #1040, opened 2023-07-20
Add cli esplora example

evanlinjin:
### Description

This PR builds on top of #1034 and adds a cli-example for our `esplora` chain-src crate.

### Notes to the reviewers

Don't merge this until #1034 is merged. The only relevant commit is 5ff0412d8edba933760c79a40be29c8365719096.

### Changelog notice

* Add cli-example for `esplora`.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -1066,6 +1066,45 @@ impl<A> ChangeSet<A> {
             })
             .chain(self.txouts.iter().map(|(op, txout)| (*op, txout)))
     }
+
+    /// Iterates over the heights of that the new transaction anchors in this changeset.
+    ///
+    /// This is useful if you want to find which heights you need to fetch data about in order to
+    /// confirm or exclude these anchors.
+    ///
+    /// See also: [`TxGraph::missing_heights`]
+    pub fn anchor_heights(&self) -> impl Iterator<Item = u32> + '_
+    where
+        A: Anchor,
+    {
+        let mut dedup = None;
+        self.anchors
+            .iter()
+            .map(|(a, _)| a.anchor_block().height)
+            .filter(move |height| {
+                let duplicate = dedup == Some(*height);
+                dedup = Some(*height);
+                !duplicate
+            })
+    }
+
+    /// Returns an iterator for the [`anchor_heights`] in this changeset that are not included in
+    /// `local_chain`. This tells you which heights you need to include in `local_chain` in order
+    /// for it to conclusively act as a [`ChainOracle`] for the transaction anchors this changeset
+    /// will add.
+    ///
+    /// [`ChainOracle`]: crate::ChainOracle
+    /// [`anchor_heights`]: Self::anchor_heights
+    pub fn missing_heights_from<'a>(
+        &'a self,
+        local_chain: &'a LocalChain,
+    ) -> impl Iterator<Item = u32> + 'a
+    where
+        A: Anchor,
+    {
+        self.anchor_heights()
+            .filter(move |height| !local_chain.blocks().contains_key(height))
+    }
 }
 
 impl<A: Ord> Append for ChangeSet<A> {
--- a/example-crates/example_esplora/src/main.rs
+++ b/example-crates/example_esplora/src/main.rs
@@ -1,0 +1,324 @@
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    io::{self, Write},
+    sync::Mutex,
+};
+
+use bdk_chain::{
+    bitcoin::{Address, Network, OutPoint, ScriptBuf, Txid},
+    indexed_tx_graph::IndexedTxGraph,
+    keychain::WalletChangeSet,
+    local_chain::{CheckPoint, LocalChain},
+    Append, ConfirmationTimeAnchor,
+};
+
+use bdk_esplora::{esplora_client, EsploraExt};
+
+use example_cli::{
+    anyhow::{self, Context},
+    clap::{self, Parser, Subcommand},
+    Keychain,
+};
+
+const DB_MAGIC: &[u8] = b"bdk_example_esplora";
+const DB_PATH: &str = ".bdk_esplora_example.db";
+
+#[derive(Subcommand, Debug, Clone)]
+enum EsploraCommands {
+    /// Scans the addresses in the wallet using the esplora API.
+    Scan {
+        /// When a gap this large has been found for a keychain, it will stop.
+        #[clap(long, default_value = "5")]
+        stop_gap: usize,
+        #[clap(flatten)]
+        scan_options: ScanOptions,
+    },
+    /// Scan for particular addresses and unconfirmed transactions using the esplora API.
+    Sync {
+        /// Scan all the unused addresses.
+        #[clap(long)]
+        unused_spks: bool,
+        /// Scan every address that you have derived.
+        #[clap(long)]
+        all_spks: bool,
+        /// Scan unspent outpoints for spends or changes to confirmation status of residing tx.
+        #[clap(long)]
+        utxos: bool,
+        /// Scan unconfirmed transactions for updates.
+        #[clap(long)]
+        unconfirmed: bool,
+        #[clap(flatten)]
+        scan_options: ScanOptions,
+    },
+}
+
+#[derive(Parser, Debug, Clone, PartialEq)]
+pub struct ScanOptions {
+    /// Max number of concurrent esplora server requests.
+    #[clap(long, default_value = "1")]
+    pub parallel_requests: usize,
+}
+
+fn main() -> anyhow::Result<()> {
+    let (args, keymap, index, db, init_changeset) = example_cli::init::<
+        EsploraCommands,
+        WalletChangeSet<Keychain, ConfirmationTimeAnchor>,
+    >(DB_MAGIC, DB_PATH)?;
+
+    // Contruct `IndexedTxGraph` and `LocalChain` with our initial changeset. They are wrapped in
+    // `Mutex` to display how they can be used in a multithreaded context. Technically the mutexes
+    // aren't strictly needed here.
+    let graph = Mutex::new({
+        let mut graph = IndexedTxGraph::new(index);
+        graph.apply_changeset(init_changeset.index_tx_graph);
+        graph
+    });
+    let chain = Mutex::new({
+        let mut chain = LocalChain::default();
+        chain.apply_changeset(&init_changeset.chain);
+        chain
+    });
+
+    let esplora_url = match args.network {
+        Network::Bitcoin => "https://blockstream.info/api",
+        Network::Testnet => "https://blockstream.info/testnet/api",
+        Network::Regtest => "http://localhost:3002",
+        Network::Signet => "https://mempool.space/signet/api",
+        _ => panic!("unsupported network"),
+    };
+
+    let client = esplora_client::Builder::new(esplora_url).build_blocking()?;
+
+    let esplora_cmd = match &args.command {
+        // These are commands that are handled by this example (sync, scan).
+        example_cli::Commands::ChainSpecific(esplora_cmd) => esplora_cmd,
+        // These are general commands handled by example_cli. Execute the cmd and return.
+        general_cmd => {
+            let res = example_cli::handle_commands(
+                &graph,
+                &db,
+                &chain,
+                &keymap,
+                args.network,
+                |tx| {
+                    client
+                        .broadcast(tx)
+                        .map(|_| ())
+                        .map_err(anyhow::Error::from)
+                },
+                general_cmd.clone(),
+            );
+
+            db.lock().unwrap().commit()?;
+            return res;
+        }
+    };
+
+    // Prepare the `IndexedTxGraph` update based on whether we are scanning or syncing.
+    // Scanning: We are iterating through spks of all keychains and scanning for transactions for
+    //   each spk. We start with the lowest derivation index spk and stop scanning after `stop_gap`
+    //   number of consecutive spks have no transaction history. A Scan is done in situations of
+    //   wallet restoration. It is a special case. Applications should use "sync" style updates
+    //   after an initial scan.
+    // Syncing: We only check for specified spks, utxos and txids to update their confirmation
+    //   status or fetch missing transactions.
+    let indexed_tx_graph_changeset = match &esplora_cmd {
+        EsploraCommands::Scan {
+            stop_gap,
+            scan_options,
+        } => {
+            let keychain_spks = graph
+                .lock()
+                .expect("mutex must not be poisoned")
+                .index
+                .spks_of_all_keychains()
+                .into_iter()
+                // This `map` is purely for logging.
+                .map(|(keychain, iter)| {
+                    let mut first = true;
+                    let spk_iter = iter.inspect(move |(i, _)| {
+                        if first {
+                            eprint!("\nscanning {}: ", keychain);
+                            first = false;
+                        }
+                        eprint!("{} ", i);
+                        // Flush early to ensure we print at every iteration.
+                        let _ = io::stderr().flush();
+                    });
+                    (keychain, spk_iter)
+                })
+                .collect::<BTreeMap<_, _>>();
+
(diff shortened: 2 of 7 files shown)


example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+        Network::Regtest => "http://localhost:3002",
+        Network::Signet => "https://mempool.space/signet/api",
+    };
+
+    let client = esplora_client::Builder::new(esplora_url).build_blocking()?;
+
+    // Match the given command. Exectute and return if command is provided by example_cli
+    let esplora_cmd = match &args.command {
+        // Command that are handled by the specify example
+        example_cli::Commands::ChainSpecific(electrum_cmd) => electrum_cmd,
```
LLFourn:
```suggestion
        example_cli::Commands::ChainSpecific(esplora_command) => esplora_command,
```

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+                        let _ = io::stdout().flush();
+                    });
+                    (keychain, spk_iter)
+                })
+                .collect::<BTreeMap<_, _>>();
+
+            drop(graph);
+
+            client
+                .update_tx_graph(
```
LLFourn, line 157: This method should not be called `update_tx_graph` since in the context of that library it is not updating anything. It's returning transaction relevant to a bunch of filters. I'd just call them `scan_txs` and `scan_txs_with_keychains`.

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+                },
+                general_cmd.clone(),
+            );
+
+            db.lock().unwrap().commit()?;
+            return res;
+        }
+    };
+
+    let (update_graph, update_keychain_indices) = match &esplora_cmd {
```
LLFourn: `update_keychain_indicies` should not be returned here since it only relevant in one branch. Just process it in the `scan` branch.

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+                            first = false;
+                        }
+                        eprint!("{} ", i);
+                        let _ = io::stdout().flush();
+                    });
+                    (keychain, spk_iter)
+                })
+                .collect::<BTreeMap<_, _>>();
+
+            drop(graph);
```
LLFourn: nit: I think we should avoid explicit dropping where we can avoid it. It sends a weird message in the code. You can just write it like:

```rust
let keychain_spks = {
    let graph = graph.lock().unwrap();
   // 
};
```
In this case since it's only used once I don't even think you need to assign it because you collect at the end:

```rust
let keychain_spks = graph
                .lock()
                .unwrap()
                .index
                .spks_of_all_keychains()
                .into_iter()
                .map(|(keychain, iter)| {
                    let mut first = true;
                    let spk_iter = iter.inspect(move |(i, _)| {
                        if first {
                            eprint!("\nscanning {}: ", keychain);
                            first = false;
                        }
                        eprint!("{} ", i);
                        let _ = io::stdout().flush();
                    });
                    (keychain, spk_iter)
                })
                .collect::<BTreeMap<_, _>>();
```

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+    println!();
+
+    let (heights_to_fetch, tip) = {
+        let chain = &*chain.lock().unwrap();
+        let heights_to_fetch = update_graph.missing_heights(chain).collect::<Vec<_>>();
+        let tip = chain.tip();
+        (heights_to_fetch, tip)
+    };
+
+    #[cfg(debug_assertions)]
```
LLFourn: `#[cfg(debug_assertions)]` is a bit distracting here.

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+    println!(
+        "new chain: {:?}",
+        update.tip.iter().map(|cp| cp.height()).collect::<Vec<_>>()
+    );
+    println!("new tip: {}", update.tip.height());
+
+    // check that all anchors are part of the new tip's history
+    #[cfg(debug_assertions)]
```
LLFourn: Am I meant to do this in my own code under a `cfg(debug_assertion)`? Can you explain why I have to do this. What is specific to esplora about this information?.

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    let db_changeset: LocalChangeSet<Keychain, ConfirmationTimeAnchor> = {
+        let mut chain = chain.lock().unwrap();
+        let mut graph = graph.lock().unwrap();
+
+        let chain_changeset = chain.apply_update(update)?;
+
+        let indexed_additions = {
+            let mut additions = IndexedAdditions::default();
+            let (_, index_additions) = graph.index.reveal_to_target_multi(&update_keychain_indices);
```
LLFourn: Can we redesign this so `reveal_to_target_multi` is only done in the `scan` branch and not for `sync`. We really want to get across that `scan` is the unusual choice which is slow and requires this patch up logic. Most of the time the application should be picky about what data it fetches by looking at what is likely to have updates.. We also need to a comment explaining why reveal_to_target_multi **must** be done first when you're using a stop gap type scan.

LLFourn (changes requested):

LLFourn:
Thanks. In general this example could do with a lot more comments explaining what's happening and why. With examples it's better to be on the side of too many comments rather than too little.

example-crates/example_esplora/src/main.rs
evanlinjin: How about `create_tx_graph_update`?

example-crates/example_esplora/src/main.rs
evanlinjin: You are right, this does not belong here. It was just a convenient place to put it when I was debugging and I left it there.

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,341 @@
...
+    clap::{self, Parser, Subcommand},
+    Keychain,
+};
+
+const DB_MAGIC: &[u8] = b"bdk_example_esplora";
+const DB_PATH: &str = ".bdk_esplora_example.db";
+
+#[derive(Subcommand, Debug, Clone)]
+enum EsploraCommands {
+    /// Scans the addresses in the wallet sing the esplora API.
```
benthecarman: > sing

using

example-crates/example_esplora/src/main.rs
evanlinjin: I think this should be handled in a separate PR? I made a ticket: #1058

notmandatory:
While testing I found a bug in `esplora`  `blocking_ext.rs` and also exists in `async_ext.rs`. When syncing the `stop_gap` is set to `usize::MAX` and the `update_tx_graph` functions fail on the addition here:

```rust
                if last_index > last_active_index.map(|i| i + stop_gap as u32) {
                    break;
                }
```

Easy enough to fix: 
```rust
                if last_index > last_active_index.map(|i| i.saturating_add(stop_gap as u32)) {
                    break;
                }
```

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,326 @@
...
+        chain.apply_changeset(&init_changeset.chain_changeset);
+        chain
+    });
+
+    let esplora_url = match args.network {
+        Network::Bitcoin => "https://blockstream.info/api",
+        Network::Testnet => "https://blockstream.info/testnet/api",
+        Network::Regtest => "http://localhost:3002",
+        Network::Signet => "https://mempool.space/signet/api",
+        _ => panic!("unsuported network"),
```
vladimirfomene: Nits: wrong spelling for unsupported.

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,326 @@
...
+                .map(|(keychain, iter)| {
+                    let mut first = true;
+                    let spk_iter = iter.inspect(move |(i, _)| {
+                        if first {
+                            eprint!("\nscanning {}: ", keychain);
+                            first = false;
+                        }
+                        eprint!("{} ", i);
+                        // Flush early to ensure we print at every iteration.
+                        let _ = io::stdout().flush();
```
vladimirfomene: Quick question. I thought eprint! prints to stderr. Will flushing stdout also flush stderr?

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,326 @@
...
+
+                if *all_spks {
+                    let all_spks = graph
+                        .index
+                        .all_spks()
+                        .iter()
+                        .map(|(k, v)| (*k, v.clone()))
+                        .collect::<Vec<_>>();
+                    spks = Box::new(spks.chain(all_spks.into_iter().map(|(index, script)| {
+                        eprintln!("scanning {:?}", index);
```
vladimirfomene, line 214: shouldn't we also flush here?

example-crates/example_esplora/src/main.rs
LLFourn: Hmm good point. No it shouldn't. I wonder if this line is actually needed!

LLFourn:
LGTM but i think we are waiting on https://github.com/bitcoindevkit/bdk/pull/1065 before this one.

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    // Up to this point, we have only created a `TxGraph` update, but not an update for our
+    // `ChainOracle` implementation (`LocalChain`). The `TxGraph` update may contain chain anchors,
+    // so we need the corresponding blocks to exist in our `LocalChain`. Here, we find the heights
+    // of missing blocks in `LocalChain`.
+    //
+    // Getting the local chain tip is only for printing to stdout.
+    let (missing_block_heights, tip) = {
+        let chain = &*chain.lock().unwrap();
+        let heights_to_fetch = graph_update.missing_heights(chain).collect::<Vec<_>>();
```
LLFourn: I think this can be simplified. Notice we can figure out heights_to_fetch by just iterating over the TxGraph changeset's added anchors. Each new anchor height is almost certainly relevant to us and can't see any harm in just fetching the block at that height even if somehow it's not. I'm glad that `missing_heights` exists and is correct but I don't think it's needed here.

This means this example can be changed the update is applied and an `indexed_tx_graph::ChangeSet` is returned from each match arm. Then we can go and update the chain. We *could* even persist the change in each match arm and lazily update the chain and persist that afterwards as this demonstrates the decoupling of chain and graph nicely.

LLFourn (changes requested):

danielabrozzoni:
I just pushed:
- Evan's commit, rebased
- A commit on top, to be squashed, fixing most of the review comments
- I rebased this whole PR on #1093 since I suppose it's going to be merged before this one 

> While testing I found a bug in esplora blocking_ext.rs and also exists in async_ext.rs. When syncing the stop_gap is set to usize::MAX and the update_tx_graph functions fail on the addition here:

@notmandatory I don't understand how to reproduce this. Does it happen with every call to sync?

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,344 @@
...
+    // reason, we want retrieve the heights of the newly added anchors, and fetch the corresponding
+    // blocks.
+
+    // We get the heights of all the anchors introduced by the changeset, and the local chain tip.
+    // Note that the latter is only used for logging.
+    let (missing_block_heights, tip) = {
+        let chain = &*chain.lock().unwrap();
+        let heights_to_fetch = changeset
+            .index_tx_graph
+            .graph
```
LLFourn, line 299: I think we can add a convenience  method to all the types that have a `tx_graph::ChangeSet` in them called `relevant_heights` (return an `Iterator<Item = u32>`) to clear this up. This code would work but it would often fetch data redundantly from the chain because it doesn't filter by what we already know about the chain. So I think this code should be more like:

```rust
let heights_to_fetch = changeset.relevant_heights().filter(|height| !chain.blocks.contains_key(height)).collect::<Vec<_>>();
```
In addition, since this mistake is clearly easy to make we could also have `missing_heights(&self, local_chain: &LocalChain) -> impl Iterator<Item = u32> + '_`:

```rust
let heights_to_fetch = changeset.missing_hegihts(&chain).collect::<Vec<_>>();
```

example-crates/example_esplora/src/main.rs
```diff
@@ -0,0 +1,344 @@
...
+    // We want the `LocalChain` to have data about all the anchors in the `TxGraph` - for this
+    // reason, we want retrieve the heights of the newly added anchors, and fetch the corresponding
+    // blocks.
+
+    // We get the heights of all the anchors introduced by the changeset, and the local chain tip.
+    // Note that the latter is only used for logging.
+    let (missing_block_heights, tip) = {
+        let chain = &*chain.lock().unwrap();
+        let heights_to_fetch = changeset
+            .index_tx_graph
```
LLFourn: side note: this should be named `indexed_tx_graph` since that's the name of the module it's from. Probably my fault! Would be happy if you renamed it in this PR.

example-crates/example_esplora/src/main.rs
LLFourn: Unresolving this since it was only partially addressed. Here's what I think is remaining to do:

Do not declare a mutable changeset outside the match arms and the progressively update it. Instead just make each match arm return a `indexed_tx_graph::ChangeSet`. This means the graph part of the update is completed in the match arms and then you only have to do the local chain update outside. This is feels simpler and makes the separation between the two updates more stark.

LLFourn:
I've made some requested changes. Rather than go back and forth too much I made a PR to show what I had in mind: https://github.com/evanlinjin/bdk/pull/6

(note I accidentally push to this PR branch at some point. Ignore that. I've reset it back to how it was).

LLFourn (dismissed):

danielabrozzoni:
Changes from my last push (019fa4c0489d2738addc0b4cec22d1d89312e182):
- I reviewed and cherry-picked @LLFourn's commits
- I removed my fix from #1093 since it will take some time before it's ready
- CI is broken, I'm waiting for #1105, and then I'll rebase again

@evanlinjin I'd say this is ready to merge. Can you review the commits pushed on top, squash them, and ACK? Then I'll ACK as well :)

danielabrozzoni:
Rebased after #1048 in order to fix CI

crates/chain/src/tx_graph.rs
```diff
@@ -1066,6 +1066,33 @@ impl<A> ChangeSet<A> {
...
             .chain(self.txouts.iter().map(|(op, txout)| (*op, txout)))
     }
+
+    /// Iterates over the heights of that the new transaction anchors in this changeset.
+    ///
+    /// This is useful if you want to find which heights you need to fetch data about in order to
+    /// confirm or exclude these anchors.
+    ///
+    /// See also: [`Self::missing_heights`]
+    pub fn relevant_heights(&self) -> impl Iterator<Item=u32> + '_ where A: Anchor {
```
evanlinjin: **Nit:** I would name this `anchor_heights`. I think `relevant` here seems ambiguous.

evanlinjin (commented):
Thanks for taking this forward @danielabrozzoni. Just a really small nit. Looks good otherwise.

evanlinjin:
I also tested everything. Works fine.

danielabrozzoni (approved):
ACK f41cc1cb374d6bae553dd997ebb325ced21025f0

notmandatory (approved):
ACK f41cc1cb374d6bae553dd997ebb325ced21025f0

I was also not able to reproduce the sync error I saw before.

notmandatory:
I'm dismissing @LLFourn blocking review so this can be merged since it looks like comments are all addressed.

notmandatory:
> ACK [f41cc1c](https://github.com/bitcoindevkit/bdk/commit/f41cc1cb374d6bae553dd997ebb325ced21025f0)
> 
> I was also not able to reproduce the sync error I saw before.

I found the sync error mentioned above, created tests to repro and added fix in #1110 .

PR #1041, opened 2023-07-20
Add `bitcoind_rpc` chain source module.

evanlinjin:
### Description

This PR builds on top of #1034 and adds the `bitcoind_rpc` chain-src module and example.

### Notes to the reviewers

Don't merge this until #1034 is in!

### Changelog notice

* Add `bitcoind_rpc` chain-source module.
* Add `example_bitcoind_rpc` example module.
* Add `AnchorFromBlockPosition` trait which are for anchors that can be constructed from a given block, height and position in block.
* Add helper methods to `IndexedTxGraph` and `TxGraph` for batch operations and applying blocks directly.
* Add helper methods to `CheckPoint` for easier construction from a block `Header`.

### Checklists

* [x] Add test: we should detect when an initially-confirmed transaction is "unconfirmed" during a reorg.
* [x] Improve `example_bitcoind_rpc`: add `live` command.
* [x] Improve docs.
* [x] Reintroduce `CheckPoint`.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/bitcoind_rpc/Cargo.toml
+++ b/crates/bitcoind_rpc/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "bdk_bitcoind_rpc"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+# For no-std, remember to enable the bitcoin/no-std feature
+bitcoin = { version = "0.30", default-features = false }
+bitcoincore-rpc = { version = "0.17" }
+
+[dev-dependencies]
+bdk_chain = { path = "../chain", version = "0.5", features = ["serde", "miniscript"] }
+bitcoind = { version = "0.33", features = ["25_0"] }
+anyhow = { version = "1" }
+
+[features]
+default = ["std"]
+std = ["bitcoin/std"]
+serde = ["bitcoin/serde"]
--- a/crates/bitcoind_rpc/Cargo.toml
+++ b/crates/bitcoind_rpc/Cargo.toml
@@ -9,13 +9,13 @@ edition = "2021"
 # For no-std, remember to enable the bitcoin/no-std feature
 bitcoin = { version = "0.30", default-features = false }
 bitcoincore-rpc = { version = "0.17" }
+bdk_chain = { path = "../chain", version = "0.5", default-features = false }
 
 [dev-dependencies]
-bdk_chain = { path = "../chain", version = "0.5", features = ["serde", "miniscript"] }
 bitcoind = { version = "0.33", features = ["25_0"] }
 anyhow = { version = "1" }
 
 [features]
 default = ["std"]
-std = ["bitcoin/std"]
-serde = ["bitcoin/serde"]
+std = ["bitcoin/std", "bdk_chain/std"]
+serde = ["bitcoin/serde", "bdk_chain/serde"]
--- a/crates/bitcoind_rpc/src/lib.rs
+++ b/crates/bitcoind_rpc/src/lib.rs
@@ -9,8 +9,7 @@
 //! mempool.
 #![warn(missing_docs)]
 
-use std::collections::BTreeMap;
-
+use bdk_chain::{local_chain::CheckPoint, BlockId};
 use bitcoin::{block::Header, Block, BlockHash, Transaction};
 pub use bitcoincore_rpc;
 use bitcoincore_rpc::bitcoincore_rpc_json;
@@ -24,7 +23,7 @@ pub struct Emitter<'c, C> {
     client: &'c C,
     start_height: u32,
 
-    emitted_blocks: BTreeMap<u32, BlockHash>,
+    last_cp: Option<CheckPoint>,
     last_block: Option<bitcoincore_rpc_json::GetBlockResult>,
 
     /// The latest first-seen epoch of emitted mempool transactions. This is used to determine
@@ -37,14 +36,29 @@ pub struct Emitter<'c, C> {
 }
 
 impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
-    /// Constructs a new [`Emitter`] with the provided [`bitcoincore_rpc::Client`].
+    /// Construct a new [`Emitter`] with the given RPC `client` and `start_height`.
     ///
     /// `start_height` is the block height to start emitting blocks from.
-    pub fn new(client: &'c C, start_height: u32) -> Self {
+    pub fn from_height(client: &'c C, start_height: u32) -> Self {
         Self {
             client,
             start_height,
-            emitted_blocks: BTreeMap::new(),
+            last_cp: None,
+            last_block: None,
+            last_mempool_time: 0,
+            last_mempool_tip: None,
+        }
+    }
+
+    /// Construct a new [`Emitter`] with the given RPC `client` and `checkpoint`.
+    ///
+    /// `checkpoint` is used to find the latest block which is still part of the best chain. The
+    /// [`Emitter`] will emit blocks starting right above this block.
+    pub fn from_checkpoint(client: &'c C, checkpoint: CheckPoint) -> Self {
+        Self {
+            client,
+            start_height: 0,
+            last_cp: Some(checkpoint),
             last_block: None,
             last_mempool_time: 0,
             last_mempool_tip: None,
@@ -114,7 +128,7 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
             .collect::<Result<Vec<_>, _>>()?;
 
         self.last_mempool_time = latest_time;
-        self.last_mempool_tip = self.emitted_blocks.iter().last().map(|(&height, _)| height);
+        self.last_mempool_tip = self.last_cp.as_ref().map(|cp| cp.height());
 
         Ok(txs_to_emit)
     }
@@ -135,7 +149,7 @@ enum PollResponse {
     NoMoreBlocks,
     /// Fetched block is not in the best chain.
     BlockNotInBestChain,
-    AgreementFound(bitcoincore_rpc_json::GetBlockResult),
+    AgreementFound(bitcoincore_rpc_json::GetBlockResult, CheckPoint),
     AgreementPointNotFound,
 }
 
@@ -146,7 +160,10 @@ where
     let client = emitter.client;
 
     if let Some(last_res) = &emitter.last_block {
-        assert!(!emitter.emitted_blocks.is_empty());
+        assert!(
+            emitter.last_cp.is_some(),
+            "must not have block result without last cp"
+        );
 
         let next_hash = match last_res.nextblockhash {
             None => return Ok(PollResponse::NoMoreBlocks),
@@ -160,7 +177,7 @@ where
         return Ok(PollResponse::Block(res));
     }
 
-    if emitter.emitted_blocks.is_empty() {
+    if emitter.last_cp.is_none() {
         let hash = client.get_block_hash(emitter.start_height as _)?;
 
         let res = client.get_block_info(&hash)?;
@@ -170,15 +187,15 @@ where
         return Ok(PollResponse::Block(res));
     }
 
-    for (&_, hash) in emitter.emitted_blocks.iter().rev() {
-        let res = client.get_block_info(hash)?;
+    for cp in emitter.last_cp.iter().flat_map(CheckPoint::iter) {
+        let res = client.get_block_info(&cp.hash())?;
         if res.confirmations < 0 {
             // block is not in best chain
             continue;
         }
 
         // agreement point found
-        return Ok(PollResponse::AgreementFound(res));
+        return Ok(PollResponse::AgreementFound(res, cp));
     }
 
     Ok(PollResponse::AgreementPointNotFound)
@@ -196,9 +213,28 @@ where
         match poll_once(emitter)? {
             PollResponse::Block(res) => {
                 let height = res.height as u32;
-                let item = get_item(&res.hash)?;
-                assert_eq!(emitter.emitted_blocks.insert(height, res.hash), None);
+                let hash = res.hash;
+                let item = get_item(&hash)?;
+
+                let this_id = BlockId { height, hash };
+                let prev_id = res.previousblockhash.map(|prev_hash| BlockId {
+                    height: height - 1,
+                    hash: prev_hash,
+                });
+
+                match (&mut emitter.last_cp, prev_id) {
+                    (Some(cp), _) => *cp = cp.clone().push(this_id).expect("must push"),
+                    (last_cp, None) => *last_cp = Some(CheckPoint::new(this_id)),
+                    // When the receiver constructs a local_chain update from a block, the previous
+                    // checkpoint is also included in the update. We need to reflect this state in
+                    // `Emitter::last_cp` as well.
+                    (last_cp, Some(prev_id)) => {
+                        *last_cp = Some(CheckPoint::new(prev_id).push(this_id).expect("must push"))
+                    }
+                }
+
                 emitter.last_block = Some(res);
+
                 return Ok(Some((height, item)));
             }
             PollResponse::NoMoreBlocks => {
@@ -209,11 +245,11 @@ where
                 emitter.last_block = None;
                 continue;
             }
-            PollResponse::AgreementFound(res) => {
+            PollResponse::AgreementFound(res, cp) => {
                 let agreement_h = res.height as u32;
 
                 // get rid of evicted blocks
-                emitter.emitted_blocks.split_off(&(agreement_h + 1));
+                emitter.last_cp = Some(cp);
 
                 // The tip during the last mempool emission needs to in the best chain, we reduce
                 // it if it is not.
@@ -226,7 +262,11 @@ where
                 continue;
             }
             PollResponse::AgreementPointNotFound => {
-                emitter.emitted_blocks.clear();
+                // We want to clear `last_cp` and set `start_height` to the first checkpoint's
+                // height. This way, the first checkpoint in `LocalChain` can be replaced.
+                if let Some(last_cp) = emitter.last_cp.take() {
+                    emitter.start_height = last_cp.height();
+                }
                 emitter.last_block = None;
                 continue;
             }
--- a/crates/bitcoind_rpc/src/lib.rs
+++ b/crates/bitcoind_rpc/src/lib.rs
@@ -33,7 +33,7 @@ pub struct Emitter<'c, C> {
 
     /// The last emitted block during our last mempool emission. This is used to determine whether
     /// there has been a reorg since our last mempool emission.
-    last_mempool_tip: Option<(u32, BlockHash)>,
+    last_mempool_tip: Option<u32>,
 }
 
 impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
@@ -65,12 +65,17 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
     pub fn mempool(&mut self) -> Result<Vec<(Transaction, u64)>, bitcoincore_rpc::Error> {
         let client = self.client;
 
-        let prev_mempool_tip = match self.last_mempool_tip {
-            // use 'avoid-re-emission' logic if there is no reorg
-            Some((height, hash)) if self.emitted_blocks.get(&height) == Some(&hash) => height,
-            _ => 0,
-        };
+        // This is the emitted tip height during the last mempool emission.
+        let prev_mempool_tip = self
+            .last_mempool_tip
+            // We use `start_height - 1` as we cannot guarantee that the block at
+            // `start_height` has been emitted.
+            .unwrap_or(self.start_height.saturating_sub(1));
 
+        // Mempool txs come with a timestamp of when the tx is introduced to the mempool. We keep
+        // track of the latest mempool tx's timestamp to determine whether we have seen a tx
+        // before. `prev_mempool_time` is the previous timestamp and `last_time` records what will
+        // be the new latest timestamp.
         let prev_mempool_time = self.last_mempool_time;
         let mut latest_time = prev_mempool_time;
 
@@ -109,11 +114,7 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
             .collect::<Result<Vec<_>, _>>()?;
 
         self.last_mempool_time = latest_time;
-        self.last_mempool_tip = self
-            .emitted_blocks
-            .iter()
-            .last()
-            .map(|(&height, &hash)| (height, hash));
+        self.last_mempool_tip = self.emitted_blocks.iter().last().map(|(&height, _)| height);
 
         Ok(txs_to_emit)
     }
@@ -209,7 +210,18 @@ where
                 continue;
             }
             PollResponse::AgreementFound(res) => {
-                emitter.emitted_blocks.split_off(&(res.height as u32 + 1));
+                let agreement_h = res.height as u32;
+
+                // get rid of evicted blocks
+                emitter.emitted_blocks.split_off(&(agreement_h + 1));
+
+                // The tip during the last mempool emission needs to in the best chain, we reduce
+                // it if it is not.
+                if let Some(h) = emitter.last_mempool_tip.as_mut() {
+                    if *h > agreement_h {
+                        *h = agreement_h;
+                    }
+                }
                 emitter.last_block = Some(res);
                 continue;
             }
--- a/crates/bitcoind_rpc/src/lib.rs
+++ b/crates/bitcoind_rpc/src/lib.rs
@@ -1,0 +1,243 @@
+//! This crate is used for emitting blockchain data from the `bitcoind` RPC interface (excluding the
+//! RPC wallet API).
+//!
+//! [`Emitter`] is the main structure which sources blockchain data from [`bitcoincore_rpc::Client`].
+//!
+//! To only get block updates (exclude mempool transactions), the caller can use
+//! [`Emitter::next_block`] or/and [`Emitter::next_header`] until it returns `Ok(None)` (which means
+//! the chain tip is reached). A separate method, [`Emitter::mempool`] can be used to emit the whole
+//! mempool.
+#![warn(missing_docs)]
+
+use std::collections::BTreeMap;
+
+use bitcoin::{block::Header, Block, BlockHash, Transaction};
+pub use bitcoincore_rpc;
+use bitcoincore_rpc::bitcoincore_rpc_json;
+
+/// A structure that emits data sourced from [`bitcoincore_rpc::Client`].
+///
+/// Refer to [module-level documentation] for more.
+///
+/// [module-level documentation]: crate
+pub struct Emitter<'c, C> {
+    client: &'c C,
+    start_height: u32,
+
+    emitted_blocks: BTreeMap<u32, BlockHash>,
+    last_block: Option<bitcoincore_rpc_json::GetBlockResult>,
+
+    /// The latest first-seen epoch of emitted mempool transactions. This is used to determine
+    /// whether a mempool transaction is already emitted.
+    last_mempool_time: usize,
+
+    /// The last emitted block during our last mempool emission. This is used to determine whether
+    /// there has been a reorg since our last mempool emission.
+    last_mempool_tip: Option<(u32, BlockHash)>,
+}
+
+impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
+    /// Constructs a new [`Emitter`] with the provided [`bitcoincore_rpc::Client`].
+    ///
+    /// `start_height` is the block height to start emitting blocks from.
+    pub fn new(client: &'c C, start_height: u32) -> Self {
+        Self {
+            client,
+            start_height,
+            emitted_blocks: BTreeMap::new(),
+            last_block: None,
+            last_mempool_time: 0,
+            last_mempool_tip: None,
+        }
+    }
+
+    /// Emit mempool transactions, alongside their first-seen unix timestamps.
+    ///
+    /// Ideally, this method would only emit the same transaction once. However, if the receiver
+    /// filters transactions based on whether it alters the output set of tracked script pubkeys,
+    /// there are situations where we would want to re-emit. For example, if an emitted mempool
+    /// transaction spends a tracked UTXO which is confirmed at height `h`, but the receiver has
+    /// only seen up to block of height `h-1`, we want to re-emit this transaction until the
+    /// receiver has seen the block at height `h`.
+    ///
+    /// In other words, we want to re-emit a transaction if we cannot guarantee it's ancestors are
+    /// already emitted.
+    pub fn mempool(&mut self) -> Result<Vec<(Transaction, u64)>, bitcoincore_rpc::Error> {
+        let client = self.client;
+
+        let prev_mempool_tip = match self.last_mempool_tip {
+            // use 'avoid-re-emission' logic if there is no reorg
+            Some((height, hash)) if self.emitted_blocks.get(&height) == Some(&hash) => height,
+            _ => 0,
+        };
+
+        let prev_mempool_time = self.last_mempool_time;
+        let mut latest_time = prev_mempool_time;
+
+        let txs_to_emit = client
+            .get_raw_mempool_verbose()?
+            .into_iter()
+            .filter_map({
+                let latest_time = &mut latest_time;
+                move |(txid, tx_entry)| -> Option<Result<_, bitcoincore_rpc::Error>> {
+                    let tx_time = tx_entry.time as usize;
+                    if tx_time > *latest_time {
+                        *latest_time = tx_time;
+                    }
+
+                    // Avoid emitting transactions that are already emitted if we can guarantee
+                    // blocks containing ancestors are already emitted. The bitcoind rpc interface
+                    // provides us with the block height that the tx is introduced to the mempool.
+                    // If we have already emitted the block of height, we can assume that all
+                    // ancestor txs have been processed by the receiver.
+                    let is_already_emitted = tx_time <= prev_mempool_time;
+                    let is_within_height = tx_entry.height <= prev_mempool_tip as _;
+                    if is_already_emitted && is_within_height {
+                        return None;
+                    }
+
+                    let tx = match client.get_raw_transaction(&txid, None) {
+                        Ok(tx) => tx,
+                        // the tx is confirmed or evicted since `get_raw_mempool_verbose`
+                        Err(err) if err.is_not_found_error() => return None,
+                        Err(err) => return Some(Err(err)),
+                    };
+
+                    Some(Ok((tx, tx_time as u64)))
+                }
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        self.last_mempool_time = latest_time;
+        self.last_mempool_tip = self
+            .emitted_blocks
+            .iter()
+            .last()
+            .map(|(&height, &hash)| (height, hash));
+
+        Ok(txs_to_emit)
+    }
+
+    /// Emit the next block height and header (if any).
+    pub fn next_header(&mut self) -> Result<Option<(u32, Header)>, bitcoincore_rpc::Error> {
+        poll(self, |hash| self.client.get_block_header(hash))
+    }
+
+    /// Emit the next block height and block (if any).
+    pub fn next_block(&mut self) -> Result<Option<(u32, Block)>, bitcoincore_rpc::Error> {
+        poll(self, |hash| self.client.get_block(hash))
+    }
+}
+
+enum PollResponse {
+    Block(bitcoincore_rpc_json::GetBlockResult),
+    NoMoreBlocks,
+    /// Fetched block is not in the best chain.
+    BlockNotInBestChain,
+    AgreementFound(bitcoincore_rpc_json::GetBlockResult),
+    AgreementPointNotFound,
+}
+
+fn poll_once<C>(emitter: &Emitter<C>) -> Result<PollResponse, bitcoincore_rpc::Error>
+where
+    C: bitcoincore_rpc::RpcApi,
+{
+    let client = emitter.client;
+
+    if let Some(last_res) = &emitter.last_block {
+        assert!(!emitter.emitted_blocks.is_empty());
+
+        let next_hash = match last_res.nextblockhash {
+            None => return Ok(PollResponse::NoMoreBlocks),
+            Some(next_hash) => next_hash,
+        };
+
+        let res = client.get_block_info(&next_hash)?;
+        if res.confirmations < 0 {
+            return Ok(PollResponse::BlockNotInBestChain);
+        }
+        return Ok(PollResponse::Block(res));
+    }
+
+    if emitter.emitted_blocks.is_empty() {
+        let hash = client.get_block_hash(emitter.start_height as _)?;
+
+        let res = client.get_block_info(&hash)?;
+        if res.confirmations < 0 {
+            return Ok(PollResponse::BlockNotInBestChain);
+        }
+        return Ok(PollResponse::Block(res));
+    }
+
+    for (&_, hash) in emitter.emitted_blocks.iter().rev() {
+        let res = client.get_block_info(hash)?;
+        if res.confirmations < 0 {
+            // block is not in best chain
+            continue;
+        }
+
+        // agreement point found
+        return Ok(PollResponse::AgreementFound(res));
+    }
+
+    Ok(PollResponse::AgreementPointNotFound)
+}
+
+fn poll<C, V, F>(
+    emitter: &mut Emitter<C>,
+    get_item: F,
+) -> Result<Option<(u32, V)>, bitcoincore_rpc::Error>
+where
+    C: bitcoincore_rpc::RpcApi,
+    F: Fn(&BlockHash) -> Result<V, bitcoincore_rpc::Error>,
+{
+    loop {
+        match poll_once(emitter)? {
+            PollResponse::Block(res) => {
+                let height = res.height as u32;
+                let item = get_item(&res.hash)?;
+                assert_eq!(emitter.emitted_blocks.insert(height, res.hash), None);
+                emitter.last_block = Some(res);
+                return Ok(Some((height, item)));
+            }
+            PollResponse::NoMoreBlocks => {
+                emitter.last_block = None;
+                return Ok(None);
+            }
+            PollResponse::BlockNotInBestChain => {
+                emitter.last_block = None;
+                continue;
+            }
+            PollResponse::AgreementFound(res) => {
+                emitter.emitted_blocks.split_off(&(res.height as u32 + 1));
+                emitter.last_block = Some(res);
+                continue;
+            }
+            PollResponse::AgreementPointNotFound => {
+                emitter.emitted_blocks.clear();
+                emitter.last_block = None;
+                continue;
+            }
+        }
+    }
+}
+
+/// Extends [`bitcoincore_rpc::Error`].
+pub trait BitcoindRpcErrorExt {
+    /// Returns whether the error is a "not found" error.
+    ///
+    /// This is useful since [`Emitter`] emits [`Result<_, bitcoincore_rpc::Error>`]s as
+    /// [`Iterator::Item`].
+    fn is_not_found_error(&self) -> bool;
+}
+
+impl BitcoindRpcErrorExt for bitcoincore_rpc::Error {
+    fn is_not_found_error(&self) -> bool {
+        if let bitcoincore_rpc::Error::JsonRpc(bitcoincore_rpc::jsonrpc::Error::Rpc(rpc_err)) = self
+        {
+            rpc_err.code == -5
+        } else {
+            false
+        }
+    }
+}
--- a/crates/bitcoind_rpc/src/lib.rs
+++ b/crates/bitcoind_rpc/src/lib.rs
@@ -1,5 +1,5 @@
-//! This crate is used for emitting blockchain data from the `bitcoind` RPC interface (excluding the
-//! RPC wallet API).
+//! This crate is used for emitting blockchain data from the `bitcoind` RPC interface. It does not
+//! use the wallet RPC API, so this crate can be used with wallet-disabled Bitcoin Core nodes.
 //!
 //! [`Emitter`] is the main structure which sources blockchain data from [`bitcoincore_rpc::Client`].
 //!
@@ -23,7 +23,14 @@ pub struct Emitter<'c, C> {
     client: &'c C,
     start_height: u32,
 
+    /// The checkpoint of the last-emitted block that is in the best chain. If it is later found
+    /// that the block is no longer in the best chain, it will be popped off from here.
     last_cp: Option<CheckPoint>,
+
+    /// The block result returned from rpc of the last-emitted block. As this result contains the
+    /// next block's block hash (which we use to fetch the next block), we set this to `None`
+    /// whenever there are no more blocks, or the next block is no longer in the best chain. This
+    /// gives us an opportunity to re-fetch this result.
     last_block: Option<bitcoincore_rpc_json::GetBlockResult>,
 
     /// The latest first-seen epoch of emitted mempool transactions. This is used to determine
@@ -67,15 +74,14 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
 
     /// Emit mempool transactions, alongside their first-seen unix timestamps.
     ///
-    /// Ideally, this method would only emit the same transaction once. However, if the receiver
-    /// filters transactions based on whether it alters the output set of tracked script pubkeys,
-    /// there are situations where we would want to re-emit. For example, if an emitted mempool
-    /// transaction spends a tracked UTXO which is confirmed at height `h`, but the receiver has
-    /// only seen up to block of height `h-1`, we want to re-emit this transaction until the
-    /// receiver has seen the block at height `h`.
+    /// This method emits each transaction only once, unless we cannot guarantee the transaction's
+    /// ancestors are already emitted.
     ///
-    /// In other words, we want to re-emit a transaction if we cannot guarantee it's ancestors are
-    /// already emitted.
+    /// To understand why, consider a receiver which filters transactions based on whether it
+    /// alters the UTXO set of tracked script pubkeys. If an emitted mempool transaction spends a
+    /// tracked UTXO which is confirmed at height `h`, but the receiver has only seen up to block
+    /// of height `h-1`, we want to re-emit this transaction until the receiver has seen the block
+    /// at height `h`.
     pub fn mempool(&mut self) -> Result<Vec<(Transaction, u64)>, bitcoincore_rpc::Error> {
         let client = self.client;
 
--- a/crates/bitcoind_rpc/tests/test_emitter.rs
+++ b/crates/bitcoind_rpc/tests/test_emitter.rs
@@ -757,3 +757,56 @@ fn mempool_during_reorg() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+/// If blockchain re-org includes the start height, emit new start height block
+///
+/// 1. mine 101 blocks
+/// 2. emmit blocks 99a, 100a
+/// 3. invalidate blocks 99a, 100a, 101a
+/// 4. mine new blocks 99b, 100b, 101b
+/// 5. emmit block 99b
+///
+/// The block hash of 99b should be different than 99a, but their previous block hashes should
+/// be the same.
+#[test]
+fn no_agreement_point() -> anyhow::Result<()> {
+    const PREMINE_COUNT: usize = 101;
+
+    let env = TestEnv::new()?;
+
+    // start height is 99
+    let mut emitter = Emitter::new(&env.client, (PREMINE_COUNT - 2) as u32);
+
+    // mine 101 blocks
+    env.mine_blocks(PREMINE_COUNT, None)?;
+
+    // emit block 99a
+    let (_, block_header_99a) = emitter.next_header()?.expect("block 99a header");
+    let block_hash_99a = block_header_99a.block_hash();
+    let block_hash_98a = block_header_99a.prev_blockhash;
+
+    // emit block 100a
+    let (_, block_header_100a) = emitter.next_header()?.expect("block 100a header");
+    let block_hash_100a = block_header_100a.block_hash();
+
+    // get hash for block 101a
+    let block_hash_101a = env.client.get_block_hash(101)?;
+
+    // invalidate blocks 99a, 100a, 101a
+    env.client.invalidate_block(&block_hash_99a)?;
+    env.client.invalidate_block(&block_hash_100a)?;
+    env.client.invalidate_block(&block_hash_101a)?;
+
+    // mine new blocks 99b, 100b, 101b
+    env.mine_blocks(3, None)?;
+
+    // emit block header 99b
+    let (_, block_header_99b) = emitter.next_header()?.expect("block 99b header");
+    let block_hash_99b = block_header_99b.block_hash();
+    let block_hash_98b = block_header_99b.prev_blockhash;
+
+    assert_ne!(block_hash_99a, block_hash_99b);
+    assert_eq!(block_hash_98a, block_hash_98b);
+
+    Ok(())
+}
--- a/crates/bitcoind_rpc/tests/test_emitter.rs
+++ b/crates/bitcoind_rpc/tests/test_emitter.rs
@@ -368,7 +368,8 @@ fn test_into_tx_graph() -> anyhow::Result<()> {
     // must receive mined block which will confirm the transactions.
     {
         let (height, block) = emitter.next_block()?.expect("must get mined block");
-        let _ = chain.apply_update(block_to_chain_update(&block, height))?;
+        let _ = chain
+            .apply_update(CheckPoint::from_header(&block.header, height).into_update(false))?;
         let indexed_additions = indexed_tx_graph.apply_block_relevant(block, height);
         assert!(indexed_additions.graph.txs.is_empty());
         assert!(indexed_additions.graph.txouts.is_empty());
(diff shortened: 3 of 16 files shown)


crates/bitcoind_rpc/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+name = "bdk_bitcoind_rpc"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+bdk_chain = { path = "../chain", version = "0.5.0", features = ["serde", "miniscript"] }
+bitcoincore-rpc = { version = "0.17.0" }
+anyhow = { version = "1" }
```
notmandatory: Do we need to add `anyhow` ? None of the other crates use it, so would like to not have to add it for this one.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,476 @@
+//! This crate is used for updating [`bdk_chain`] structures with data from the `bitcoind` RPC
```
LLFourn: But not the rpc wallet API :warning: :warning:

crates/bitcoind_rpc/Cargo.toml
evanlinjin: Woops, I think that was only meant as a dev dependency! Thanks for pointing this out.

example-crates/example_rpc/src/main.rs
```diff
@@ -0,0 +1,308 @@
...
+
+                    let graph_update = {
+                        let tx_filter = bdk_bitcoind_rpc::indexer_filter(
+                            &mut graph.index,
+                            &mut indexed_additions.index_additions,
+                        );
+                        let anchor_map = bdk_bitcoind_rpc::confirmation_time_anchor;
+                        item.into_tx_graph_update(tx_filter, anchor_map)
+                    };
+                    indexed_additions.append(graph.apply_update(graph_update));
```
LLFourn: This seems to be more complicated than it needs to be. Why can't I just pass off a bunch of transactions whether they be in a block or just a chunk of mempool transactions to the indexed tx graph and have it tell me the changeset after applying them. Passing off the `indexer` to the bitcoin_rpc logic seems like a hack. bitcoind_rpc shouldn't know about this type. It's a totally separate concern.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+            EmittedUpdate::Mempool(e) => e.into_tx_graph_update(tx_filter),
+        }
+    }
+}
+
+/// An emitted block.
+#[derive(Debug, Clone)]
+pub struct EmittedBlock {
+    /// The checkpoint constructed from the block's height/hash and connected to the previous block.
+    pub cp: CheckPoint,
```
LLFourn: Is this needed. Won't the logic receiving this block be able to figure out where to connect it. `Block` has the previous block hash in it which should have already been emitted if the caller didn't know about it already.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+        }
+        tx_graph
+    }
+}
+
+/// An emitted subset of mempool transactions.
+#[derive(Debug, Clone)]
+pub struct EmittedMempool {
+    /// The checkpoint of the last-seen tip.
+    pub cp: CheckPoint,
```
LLFourn: please explain in the docs how I am meant to use this information.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+    /// Subset of mempool transactions.
+    pub txs: Vec<(Transaction, u64)>,
```
LLFourn: document what the `u64` is for.

example-crates/example_rpc/src/main.rs
```diff
@@ -0,0 +1,308 @@
...
+                    if is_mempool {
+                        if !live {
+                            return Ok(());
+                        }
+                        if await_flag(&sigterm_flag, LIVE_POLL_DUR_SECS) {
+                            return Ok(());
+                        }
+                    }
+                }
+                unreachable!()
```
LLFourn: This `unreachable` thing biases me against providing this iterator API. It ended up conceptually weird to me. IWouldn't it be clearer just to emit blocks and process them in a `loop`.

example-crates/example_rpc/src/main.rs
```diff
@@ -0,0 +1,308 @@
...
+                (Some(path), _, _) => Auth::CookieFile(path.clone()),
+                (_, Some(user), Some(pass)) => Auth::UserPass(user.clone(), pass.clone()),
+                (_, Some(_), None) => panic!("rpc auth: missing rpc_pass"),
+                (_, None, Some(_)) => panic!("rpc auth: missing rpc_user"),
+            },
+        )?
+    };
+
+    match rpc_cmd {
+        RpcCommands::Scan {
```
LLFourn: Maybe just `sync` instead of `scan` since we use that for scanning addresses. This is rather syncing local state with remote state like the sync commands.

LLFourn (commented):
I've made some comments. It looks awesome. I think this will be a "flagship" example on how to use bdk so I think we should spend the time to get it right. The block data processing pipeline is too complicated for what the problem is. I think `IndexedTxGraph` needs to be able to process block data and filter out what's relevant itself. This is needed for CBF block processing as well so calling out to weird helper methods in the bitcoind_rpc crate seems counter productive.

notmandatory:
@evanlinjin this looks like a really powerful way to sync with a core node. I read through the code and I think I understand the state-model approach within the `Emitter::emit_block`. Have you and @LLFourn chatted yet about his ideas to simplify it somehow? From an outside user the APIs look easy enough to use but if there are opportunities to simplify the internals it's worth discussing.

You look to be covering the main test scenarios, 1. sync blocks before and after reorg, 2. sync txs in mempool and after being included in a block.  Is it worth adding a test to verify confirmed tx are unconfirmed after a reorg?

example-crates/example_rpc/src/main.rs
evanlinjin: The way we are using it here is a bit weird. We have a flag for both "live" mode and "sync-once" mode. In my opinion, the iterator will be useful in a "sync-once" mode. The "unreachable" thing is not a necessity or a direct result of using the iterator.

crates/bitcoind_rpc/src/lib.rs
evanlinjin: I think it makes our API a bit more flexible.

For example:
* The caller pulled data from esplora to height 100.
* The caller waits a while.
* The caller creates wallet with birthday at height 200.

If we create a `local_chain::Update` at height 200 with checkpoints derived from `Block` only, we can't connect to the chain at height 100.

crates/bitcoind_rpc/src/lib.rs
evanlinjin: I removed `EmittedMempool::cp`. It's not useful.

example-crates/example_rpc/src/main.rs
evanlinjin: This is addressed in 26209de743c4cd134fcd53a1db2b7c121b27230c

crates/bitcoind_rpc/src/lib.rs
LLFourn: > I think it makes our API a bit more flexible.
> 
> For example:
> 
>     * The caller pulled data from esplora to height 100.
> 
>     * The caller waits a while.
> 
>     * The caller creates wallet with birthday at height 200.
> 
> 
> If we create a `local_chain::Update` at height 200 with checkpoints derived from `Block` only, we can't connect to the chain at height 100.

In this case you would pass in the block at `100` as `last_cp` to the emitter and then it would emit every block from 100 to 200 in order which means it would connect to `100`. Am I missing something?

evanlinjin: What if we want to sync from 200 without removing old blocks?

What if the caller only has the Genesis block and wallet birthday is at height 805016

notmandatory:
Looks like we have another MSRV issue. Need to add these:

in `cont_integration.yml`
```
          cargo update -p log --precise "0.4.18"
          cargo update -p tempfile --precise "3.6.0"
          cargo update -p minreq --precise "2.8.1"              <======= add this line to fix
          cargo update -p rustls:0.21.6 --precise "0.21.1"
          cargo update -p tokio:1.32.0 --precise "1.29.1"
          cargo update -p flate2:1.0.27 --precise "1.0.26"
          cargo update -p reqwest --precise "0.11.18"
          cargo update -p h2 --precise "0.3.20"
          cargo update -p rustls-webpki --precise "0.100.1"
```

in `README.md` 
```
# log 0.4.19 has MSRV 1.60.0+
cargo update -p log --precise "0.4.18"
# tempfile 3.7.0 has MSRV 1.63.0+
cargo update -p tempfile --precise "3.6.0"
# minreq 2.9.0 prevents pinning rustls to 0.21.1        <====== add corresponding docs
cargo update -p minreq --precise "2.8.1"
# rustls 0.21.2 has MSRV 1.60.0+
cargo update -p rustls:0.21.6 --precise "0.21.1"
# tokio 1.30 has MSRV 1.63.0+
cargo update -p tokio:1.32.0 --precise "1.29.1"
# flate2 1.0.27 has MSRV 1.63.0+
cargo update -p flate2:1.0.27 --precise "1.0.26"
# reqwest 0.11.19 has MSRV 1.63.0+
cargo update -p reqwest --precise "0.11.18"
# h2 0.3.21 has MSRV 1.63.0+
cargo update -p h2 --precise "0.3.20"
# rustls-webpki 0.100.2 has MSRV 1.60.0+.                         <======== small update
cargo update -p rustls-webpki --precise "0.100.1"
```

notmandatory:
@vladimirfomene is working on an rpc wallet example and reorg confirmed transaction tests.

crates/bitcoind_rpc/src/lib.rs
LLFourn: These are certainly questions. What are the answers? :sweat_smile:

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -0,0 +1,300 @@
...
+impl TestEnv {
+    fn new() -> anyhow::Result<Self> {
+        let daemon = match std::env::var_os("TEST_BITCOIND") {
+            Some(bitcoind_path) => bitcoind::BitcoinD::new(bitcoind_path),
+            None => bitcoind::BitcoinD::from_downloaded(),
+        }?;
+        let client = bitcoincore_rpc::Client::new(
+            &daemon.rpc_url(),
+            bitcoincore_rpc::Auth::CookieFile(daemon.params.cookie_file.clone()),
+        )?;
```
RCasatta, line 35: This works ok but I was wondering if there are reasons not to use
```
 // use BITCOIND_EXE instead of TEST_BITCOIND, see https://docs.rs/bitcoind/0.33.0/bitcoind/fn.exe_path.html
let daemon = bitcoind::BitcoinD::new(bitcoind::exe_path()?)?; 
```
and then using the inner `daemon.client` instead of creating a new one

evanlinjin: I think that will work with the latest version of `bitcoind`. However, we were using the old version before, where there was a custom package name for the `bitcoincore-rpc` dependency. Will change soon!

evanlinjin:
@LLFourn suggested that we try "decouple" the `bdk_bitcoind_rpc` API from `bdk_chain` structures. The reason for this is as follows:

* `bdk_chain` structures should be directly updateable by inputting raw bitcoin data such as `Block`s, `Transaction`s, etc. We shouldn't need to *transform* the data into `bdk_chain` types before applying it to `TxGraph`, `LocalChain`, etc.

* The will allow us to update the `bdk_chain` and `bdk_bitcoind_rpc` crates orthogonal to one another.

* This will make `bdk_bitcoind_rpc` better when used with non-`bdk_chain` projects. This will reduce a dependency.

crates/bitcoind_rpc/src/lib.rs
evanlinjin: @LLFourn I'm very confused what you mean by this comment haha.

crates/bitcoind_rpc/tests/test_emitter.rs
evanlinjin: Updated in c87b1b08bbfe63c22961093572b16828d9c355fe

---

Edit: I had to revert c87b1b08bbfe63c22961093572b16828d9c355fe because latest version of `bitcoind` is breaking MSRV.

crates/bitcoind_rpc/src/lib.rs
LLFourn: I am trying to find out the answer to my question and the questions you posed in response to my question.

example-crates/example_rpc/src/main.rs
```diff
@@ -0,0 +1,346 @@
...
+                        })
+                    }
+                };
+                let update_tip_height = chain_update.as_ref().map(|u| u.tip.height());
+
+                let tx_graph_update: Vec<TxItem<Option<ConfirmationTimeAnchor>>> = match &emission {
+                    Emission::Mempool(txs) => {
+                        txs.iter().map(|tx| (&tx.tx, None, Some(tx.time))).collect()
+                    }
+                    Emission::Block(b) => {
```
LLFourn: Can we have `apply_block(&block, height)` on `TxGraph` and `IndexedTxGraph`? I don't think you want users doing this manually when it should be that simple. The anchor type of `A` for `TxGraph` should be constrained so that it can be extracted from a block header and the position of the transaction in the block. Probably need a new trait for this like `FromBlockPosition`. `FromBlockPosition` can take `(&block, position)`

We could take this opportunity to have a different anchor for this one which includes block position.

`IndexedTxGraph` should also have `apply_block_relevant` or something like that which only adds the relevant transactions which you'd use here.

For mempool dumping I'd add `batch_insert_unconfirmed` and `batch_insert_relevant_unconfirmed`. These can take a iter of `(tx, last_seen)`.

example-crates/example_rpc/src/main.rs
```diff
@@ -0,0 +1,346 @@
...
+    Sync {
+        /// Starting block height to fallback to if no point of agreement if found
+        #[clap(env = "FALLBACK_HEIGHT", long, default_value = "0")]
+        fallback_height: u32,
+        /// The unused-scripts lookahead will be kept at this size
+        #[clap(long, default_value = "10")]
+        lookahead: u32,
+        /// Whether to be live!
+        #[clap(long, default_value = "false")]
+        live: bool,
```
LLFourn: Sorry to request this late. Can we have two command: `sync` and `live`. The `sync` command handler should just have the minimal logic to apply all the latest blocks and the mempool and then exit (nice to print how many blocks and mempool txs were processed). The live command should do what the current code does. This is so a developers can more easily see how to interact with the crate in the two cases.

example-crates/example_rpc/src/main.rs
```diff
@@ -0,0 +1,346 @@
...
+    let rpc_cmd = match args.command {
+        example_cli::Commands::ChainSpecific(rpc_cmd) => rpc_cmd,
+        general_cmd => {
+            let res = example_cli::handle_commands(
+                &graph,
+                &db,
+                &chain,
+                &keymap,
+                args.network,
+                |_| Err(anyhow::anyhow!("use `tx` instead")),
```
LLFourn: Why?

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,321 @@
...
+    }
+}
+
+/// This is the [`Iterator::Item`] of [`Iter`]. This can either represent a block/header or the set
+/// of mempool transactions.
+///
+/// Refer to [module-level documentation] for more.
+///
+/// [module-level documentation]: crate
+pub enum Emission<B> {
```
LLFourn: Why do these different data requests (mempool, next block) needed to be handled in the same "stream" of events. Looking at the caller code it doesn't seem to simplify things, they are always matched on and handled differently (this could be because of the difference in vision I have kept pushing you into a corner where this was the case).

Furthermore, A user is likely to want to check for the next block vs dump the mempool at different times. E.g. dump the mempool every 30s but check for next block every 5s (actually it would be nice if we did this in our example). I strongly believe the caller should make a conscious decision about when to do this. It is not a natural decision to me to dump the mempool after you've finished looking for new blocks (and then continuing to dump it at every second call to `next` which feels even weirder).

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,321 @@
...
+            client,
+            start_height,
+            emitted: BTreeMap::new(),
+            last: None,
+        }
+    }
+
+    /// Emit all mempool transactions.
+    pub fn mempool(&self) -> impl Iterator<Item = Result<MempoolTx, bitcoincore_rpc::Error>> + '_ {
+        let res = self.client.get_raw_mempool();
```
LLFourn: side question: there a way of custom dumping the mempool so you only get transactions that entered after a certain time?

example-crates/example_rpc/src/main.rs
evanlinjin: We cannot construct the `Client` before knowing the client's url and params, and they are defined in the `Commands::ChainSpecific` enum's variants. We cannot modify `Commands::Send`.

example-crates/example_rpc/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "example_rpc"
```
LLFourn: please rename directory and package `example_bitcoind_rpc`. Actually it should probably be `example_bitcoind_rpc_block_polling`.

example-crates/example_rpc/src/main.rs
LLFourn: Just for ref I made a PR to this branch fixing this: https://github.com/evanlinjin/bdk/pull/7

This PR can be merged now or I can re-do it to master after this is merged.

crates/bitcoind_rpc/src/lib.rs
notmandatory: I don't see anything in the RPC options for `get_raw_mempool` to filter what it returns. Probably the only way to make this more efficient is to plug into a `zmq` topic like `rawtx` and get block and mempool tx updates that way. But figuring out if/how to use `zmq` should be a post 1.0 thing to do.

https://bitcoincore.org/en/doc/25.0.0/rpc/blockchain/getrawmempool/

crates/bitcoind_rpc/tests/test_emitter.rs
notmandatory: I was able to use to the latest version of `electrsd` (which depends on latest 0.33 `bitcoind`) without breaking MSRV with this commit: https://github.com/bitcoindevkit/bdk/pull/1110/commits/7be6f86343e30c9e07bcfd52342a203ea7556026

So I think we should be able to keep latest `bitcoind`.

example-crates/example_rpc/src/main.rs
evanlinjin: @LLFourn I don't think having `apply_block` on `TxGraph` makes sense, since we have no `Indexer`. I'll add it to `IndexedTxGraph` though.

LLFourn: I'm not sure what you meant here. Do you mean you think we should only add `apply_block_relevant` and not add `apply_block` to either of them?

I would add `apply_block` too. You can use the `TxGraph` for general analysis. Maybe I want to know what % of new outputs are spent within the next 10 blocks etc. You don't need an indexer to make `TxGraph` useful.

evanlinjin: Woops, I assumed that you meant `apply_block` would filter for relevant txs (as that was what the original codebase did) and didn't finish reading your comment.

example-crates/example_rpc/src/main.rs
evanlinjin: Rebased and added to this PR as 2db28e0376e8f7b195d78e23cd5283e7a5c21a45

example-crates/example_rpc/src/main.rs
evanlinjin: introduced here: 09bdffab2fd08290157c0f0e31058dcaefbbe308

crates/bitcoind_rpc/src/lib.rs
evanlinjin: If we scan the mempool for relevant txs before scanning the latest tip, we may end up excluding mempool transactions that send from our UTXO pool (as we need the prev txs to determine relevancy).

This may be problematic if we intend to have non-duplicate tx dumps from the mempool (as suggested in https://github.com/bitcoindevkit/bdk/pull/1041#discussion_r1315406687). We will end up excluding an actually-relevant mempool transaction which sends from our wallet.

If this tx takes a while to confirm, we can be fully synced up to the tip and not see it in the wallet until it is confirmed. The wallet may accidentally rbf the tx if the tx stays unconfirmed.

I think it is good policy to scan the mempool only after scanning the latest tip.

crates/bitcoind_rpc/src/lib.rs
evanlinjin: We can use `get_raw_mempool` to only gets txids. Then, we can use the `get_mempool_entry` on an individual txid to get the time that the tx entered the mempool.

notmandatory:
If you do a rebase it should fix the MSRV errors. The only job not passing right now is the audit job ☹️, but I have an indirect fix for that in https://github.com/RCasatta/bitcoind/pull/141.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+            last_block: None,
+            last_mempool_time: 0,
+        }
+    }
+
+    /// Emit mempool transactions.
+    ///
+    /// This avoids re-emitting transactions (where viable). We can do this if all blocks
+    /// containing ancestor transactions are already emitted.
+    pub fn mempool(&mut self) -> Result<Vec<MempoolTx>, bitcoincore_rpc::Error> {
```
evanlinjin: We should have the following tests checking the behavior when `Emitter::mempool` is called multiple times.

- [x] The receiver (`bdk_chain` structures) is synced to the chain tip, and there is txs in the mempool. When we call `Emitter::mempool` multiple times, mempool txs should not be re-emitted.

- [x] In the above scenario, if the network's chain extends with empty blocks (and our receiver stays up to sync), we expect that the txs do not get re-emitted.

- [x] The network is at height `h`. The receiver is synced to height `h-1`. When a mempool tx `A` is introduced at height `h`, we should always re-emit it.

- [x] In the above scenario, if the receiver syncs up to `h`, we should still re-emit (right after). However, subsequent calls to `Emitter::mempool` should not re-emit.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+            .into_iter()
+            .filter_map(
+                move |(txid, tx_entry)| -> Option<Result<_, bitcoincore_rpc::Error>> {
+                    let tx_time = tx_entry.time as usize;
+                    if tx_time > latest_time {
+                        latest_time = tx_time;
+                    }
+
+                    // Avoid emitting transactions that are already emitted if we can guarantee
+                    // blocks containing ancestors are already emitted. We only check block height
```
LLFourn: This comment doesn't make sense to me. I don't see anything where you guarantee blocks containing ancestors are already emitted (but also don't understand why that's relevant).

evanlinjin: The `Indexer` may think a transaction is irrelevant for the time being (since it does not have the prev outpoints that it owns). We should keep re-emitting such a transaction.

A bitcoind's tx entry gives us the block height in which the tx is introduced to the mempool. If our `Emitter`'s last-emitted-block-state meets or surpasses this block height, we can guarantee that we have scanned or ancestors (except in case of reorg).

For reorg, we force-re-emit everything.

crates/bitcoind_rpc/src/lib.rs
notmandatory: I added a test for the 3rd checkbox and a CI fix in #1140, if it looks OK feel free to copy/paste or cherry-pick, which ever is easier.

evanlinjin: @notmandatory Thank you for the PR! Unfortunately we duplicated work, and I think my solution covers a broader set of situations, so will stick with that.

crates/bitcoind_rpc/tests/test_emitter.rs
evanlinjin: I have upgraded to the latest `bitcoind` crate. However, I think it is best to keep the implementation as is, just in case `bitcoincore_rpc` versions diverges between `bitcoind` and our dependencies?

crates/bitcoind_rpc/src/lib.rs
notmandatory: No problem! closed my PR.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,322 @@
+//! This crate is used for emitting blockchain data from the `bitcoind` RPC interface (excluding the
+//! RPC wallet API).
+//!
+//! [`Emitter`] is the main structure which sources blockchain data from [`bitcoincore_rpc::Client`].
+//!
+//! To only get block updates (exlude mempool transactions), the caller can use
```
notmandatory: nit: typo
```suggestion
//! To only get block updates (exclude mempool transactions), the caller can use
```

notmandatory (approved):
ACK 409033a8cf75e8216dc8611d7708e8703e36fcc2

A few little nits need fixing, but overall looks like a powerful way to sync via bitcoind rpc. But since it's so new I recommend we put a little "experimental" disclaimer in the release notes until it's gotten more user real-world testing.

notmandatory:
I created a little test for the "no agreement point" situation. Feel free to cherry pick or copy/pasta if you think it's useful.

https://github.com/bitcoindevkit/bdk/pull/1143/commits/e728c53c0d3406ea66aa9fba59953f7188db7c1a

evanlinjin:
The `Emitter::mempool` method is taking ~20seconds (today, with a busy mempool). This is because we need call `getrawtransaction` for every mempool entry that requires re-emission.

Possible ways to optimize in the future:

1. Cache transactions in the emitter.
2. Use multiple threads to call `getrawtransaction`.
3. A combination of the above.

This issue should not exist if we sync to tip before emitting from mempool (need to check).

example-crates/example_bitcoind_rpc_polling/src/main.rs
```diff
@@ -0,0 +1,366 @@
...
+        RpcCommands::Sync { rpc_args } => {
+            let RpcArgs {
+                fallback_height,
+                lookahead,
+                ..
+            } = rpc_args;
+
+            let mut chain = chain.lock().unwrap();
+            let mut graph = graph.lock().unwrap();
+            let mut db = db.lock().unwrap();
```
LLFourn: :x: don't keep locks open over IO.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -145,27 +141,149 @@ where
...
+    /// Batch insert unconfirmed transactions.
+    ///
+    /// Items of `txs` are tuples containing the transaction and an optional *last seen* timestamp.
+    /// The *last seen* communicates when the transaction is last seen in the mempool which is used
+    /// for conflict-resolution in [`TxGraph`] (refer to [`TxGraph::insert_seen_at`] for details).
+    ///
+    /// To filter out irrelevant transactions, use [`batch_insert_relevant_unconfirmed`] instead.
+    ///
+    /// [`batch_insert_relevant_unconfirmed`]: IndexedTxGraph::batch_insert_relevant_unconfirmed
+    pub fn batch_insert_unconfirmed<'t>(
```
LLFourn: Please add the same method to `TxGraph`.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -145,27 +141,149 @@ where
...
+            unconfirmed_txs
+                .into_iter()
+                .map(|(tx, last_seen)| (tx, core::iter::empty(), last_seen)),
+        )
+    }
+
+    /// Batch insert unconfirmed transactions.
+    ///
+    /// Items of `txs` are tuples containing the transaction and an optional *last seen* timestamp.
+    /// The *last seen* communicates when the transaction is last seen in the mempool which is used
```
LLFourn: What's the motivation for making `last_seen` optional. I think it should be mandatory.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -145,27 +141,149 @@ where
...
+/// This tuple contains fields in the following order:
+/// * A reference to the transaction.
+/// * A collection of [`Anchor`]s.
+/// * An optional last-seen timestamp.
+///
+/// This is used as a input item of [`batch_insert_relevant`] and [`batch_insert`].
+///
+/// [`batch_insert_relevant`]: IndexedTxGraph::batch_insert_relevant
+/// [`batch_insert`]: IndexedTxGraph::batch_insert
+pub type InsertTxItem<'t, A> = (&'t Transaction, A, Option<u64>);
```
LLFourn: Not sure about introducing this concept. Is it so bad to ask the user to insert anchors/last_seen using their separate calls?

[EDIT] I just realized they are unfortunately not separate calls in indexed tx graph and made a comment elsewhere about fixing that.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -145,27 +141,149 @@ where
...
+        changeset
+    }
+
+    /// Batch insert transactions.
+    ///
+    /// All transactions in `txs` will be inserted. To filter out irrelevant transactions, use
+    /// [`batch_insert_relevant`] instead.
+    ///
+    /// [`batch_insert_relevant`]: IndexedTxGraph::batch_insert_relevant
+    pub fn batch_insert<'t>(
```
LLFourn: Don't think this method is worth it. Doesn't do anything beyond calling `insert_tx` in a loop and appending changesets (it's surprising that it's not implemented like this).

example-crates/example_cli/src/lib.rs
```diff
@@ -664,45 +666,213 @@ where
...
+                    (tx, None)
+                }
+            };
+
+            match (broadcast)(chain_specfic, &transaction) {
+                Ok(_) => {
+                    println!("Broadcasted Tx : {}", transaction.txid());
+
+                    let keychain_changeset =
+                        graph.lock().unwrap().insert_tx(&transaction, None, None);
```
LLFourn: Why does indexed tx graph `insert_tx` take three arguments but tx graph only take one. Eeek. We should make this the same. I like `TxGraph`'s API better FWIW.

crates/chain/src/local_chain.rs
```diff
@@ -39,6 +40,39 @@ impl CheckPoint {
...
+    ///
+    /// If `header` is of the genesis block, the checkpoint won't have a [`prev`] node. Otherwise,
+    /// we return a checkpoint linked with the previous block.
+    ///
+    /// [`prev`]: CheckPoint::prev
+    pub fn from_header(header: &bitcoin::block::Header, height: u32) -> Self {
+        let hash = header.block_hash();
+        let this_block_id = BlockId { height, hash };
+
+        if header.prev_blockhash == BlockHash::all_zeros() {
```
LLFourn: isn't `if height == 0` strictly more correct here? This avoids the underflow on line 57 also. `match height.checked_sub(1)` might be the way to go.

crates/chain/src/local_chain.rs
```diff
@@ -39,6 +40,39 @@ impl CheckPoint {
...
+            return Self::new(this_block_id);
+        }
+
+        let height = height - 1;
+        let hash = header.prev_blockhash;
+        let prev_block_id = BlockId { height, hash };
+
+        CheckPoint::new(prev_block_id)
+            .extend(core::iter::once(this_block_id))
```
LLFourn: don't we have a method like `push` for single things?

example-crates/example_bitcoind_rpc_polling/src/main.rs
```diff
@@ -0,0 +1,366 @@
...
+
+            graph.index.set_lookahead_for_all(lookahead);
+            // we start at a height lower than last-seen tip in case of reorgs
+            let start_height = chain.tip().as_ref().map_or(fallback_height, |cp| {
+                cp.height().saturating_sub(ASSUME_FINAL_DEPTH)
+            });
+
+            let rpc_client = rpc_args.new_client()?;
+            let mut emitter = Emitter::new(&rpc_client, start_height);
```
LLFourn: Hmm iirc this used to be a checkpoint instead of `start_height` so we didn't have this `ASSUME_FINAL_DEPTH` thing. That was much better but I see that it needed to be sacrificed to remove the depend on `bdk_chain`. Could we make it an `Iterator` of `(height, hash)` and make an easy way to create one of those from a `CheckPoint` and use the same clever logic we had before?

Either that or we could have a `bdk_core` crate that just defines some types like `BlockId`, `Checkpoint` that we can hopefully not make breaking changes to (at least no where near as many as we'll make to `bdk_chain`).

example-crates/example_bitcoind_rpc_polling/src/main.rs
```diff
@@ -0,0 +1,366 @@
...
+
+            let mut last_db_commit = Instant::now();
+            let mut last_print = Instant::now();
+
+            while let Some((height, block)) = emitter.next_block()? {
+                let chain_update =
+                    CheckPoint::from_header(&block.header, height).into_update(false);
+                let chain_changeset = chain.apply_update(chain_update)?;
```
LLFourn: Can we guarantee that emitter won't give us out of order blocks here? If so I would `expect` this. If not maybe we can figure out how we can give this guarantee?

LLFourn (changes requested):
This is coming together really nicely. The `sync` code is really easy to follow. This is presents probably the optimal way of using the library if you are creating a service that is meant to be on all the time tracking deposits/withdrawals/LN channels etc.

I've left several questions and ways I think this can be improved.

LLFourn:
> The `Emitter::mempool` method is taking ~20seconds (today, with a busy mempool). This is because we need call `getrawtransaction` for every mempool entry that requires re-emission.
> 
> Possible ways to optimize in the future:
> 
>     1. Cache transactions in the emitter.
> 
>     2. Use multiple threads to call `getrawtransaction`.
> 
>     3. A combination of the above.
> 
> 
> This issue should not exist if we sync to tip before emitting from mempool (need to check).

Couldn't we also just have a kind of "stream" of mempool txs that never terminates and just continuously emits mempool txs you haven't seen before? So rename current `Emitter` to `BlockEmitter` and make it only emit blocks. Then create a second thing like `MempoolStream` which polls for new things and emits them when it finds them as best you can with the rpc API (probably via caching). `MempoolStream` would be continuously polled in a separate thread during the lifetime of the application I guess.

Side remark: we may want to name the crate like `bitcoin_rpc_polling` rather than `bitcoin_rpc` given we have other ways of using `bitcoin_rpc` i.e. the wallet which might have different dependencies?

example-crates/example_bitcoind_rpc_polling/src/main.rs
evanlinjin: But they aren't being used anywhere else so we'll we wasting resources locking/unlocking?

example-crates/example_bitcoind_rpc_polling/src/main.rs
evanlinjin: I like the `bdk_core` crate idea. If we decide to go down this approach, let's make these changes a separate issue?

**EDIT:** I would prefer it if this version of `bitcoind_rpc` still depended on `bdk_chain` for initializing `Emitter` with `CheckPoint`. Then we introduce the `bdk_core` crate in a later version?

evanlinjin:
> Couldn't we also just have a kind of "stream" of mempool txs that never terminates and just continuously emits mempool txs you haven't seen before? So rename current `Emitter` to `BlockEmitter` and make it only emit blocks. Then create a second thing like `MempoolStream` which polls for new things and emits them when it finds them as best you can with the rpc API (probably via caching). `MempoolStream` would be continuously polled in a separate thread during the lifetime of the application I guess.

Currently, `Emitter::mempool` relies on the last block emitted (during the last mempool emission) to determine whether we need to re-emit a mempool tx.

* If the last block emitted is no longer in the best chain, we re-emit everything in the mempool (this can be improved in the future, however, given the rarity of reorgs I think this is good enough).
* If the last block emitted during the last mempool emission is still in the best chain, we only need to emit txs introduced after that given height.

Unless we have some way for the `BlockEmitter` to inform the `MempoolEmitter` about the emitted blocks?

P.S. The re-emission logic works, and does wonders when we are synced close to the tip:

image

crates/chain/src/indexed_tx_graph.rs
evanlinjin: I really want to crack a @notmandatory joke here.

crates/chain/src/indexed_tx_graph.rs
evanlinjin: ~~What if we want to update with a mixed list of unconfirmed/confirmed transactions? I would prefer it if we just had `batch_insert` and `batch_insert_relevant` which takes in `InsertTxItem` (we can then get rid of the other methods). I think this is better because having too many methods that does similar but slightly different things can be very confusing.~~

**EDIT:** After thinking about it, we only need batch methods for unconfirmed anyway since it requires additional logic.

example-crates/example_cli/src/lib.rs
evanlinjin: I guess the only benefit of having `insert_tx` also insert anchors and last_seen is we can reduce the number of times we calculate the txid :sweat_smile:.

I agree that the API between `TxGraph` and `IndexedTxGraph` should be relatively uniform. So `IndexedTxGraph` should have separate methods for inserting tx, anchor and last_seen.

However, I think this should be tackled in a separate issue.

example-crates/example_bitcoind_rpc_polling/src/main.rs
evanlinjin: ~We can guarantee that emitter won't give us out of order blocks here.~

**EDIT:** Actually there is one scenario where the emitted block won't connect. It's when there is a reorg that happens lower than our start_height. We'll need to make use of `CheckPoint`.

crates/chain/src/indexed_tx_graph.rs
evanlinjin: Do you mind expanding on what you mean by what's not implemented like this?

evanlinjin: I think I understand what you mean now. This method is now removed.

example-crates/example_bitcoind_rpc_polling/src/main.rs
LLFourn: Sounds like a plan. The most important thing is the code has been written not to depend on details of `bdk_chain` at least and we can switch it to `bdk_core` later.

example-crates/example_bitcoind_rpc_polling/src/main.rs
LLFourn: Yes we need to pass in the iterator of hash and height (or `CheckPoint`) to solve this.

LLFourn:
> Currently, `Emitter::mempool` relies on the last block emitted (during the last mempool emission) to determine whether we need to re-emit a mempool tx.
> 
>     * If the last block emitted is no longer in the best chain, we re-emit everything in the mempool (this can be improved in the future, however, given the rarity of reorgs I think this is good enough).
> 
>     * If the last block emitted during the last mempool emission is still in the best chain, we only need to emit txs introduced after that given height.

I think I understand the whole solution now thanks. It does seem to be the simplest solution that gives the ordering guarantees. Nice!

example-crates/example_cli/src/lib.rs
LLFourn: > I guess the only benefit of having `insert_tx` also insert anchors and last_seen is we can reduce the number of times we calculate the txid 😅.

How does it?

> However, I think this should be tackled in a separate issue.

Why? Does it need further discussion? I'd just do this now and forget about it.

example-crates/example_bitcoind_rpc_polling/src/main.rs
LLFourn: But if they were being used somewhere else the whole application would be stuck. We're in an example. The point of it is to show people (and ourselves) how to use the API in a thread safe context so that other parts of the application can access the these components while a sync is going on.

example-crates/example_cli/src/lib.rs
evanlinjin: Done.

evanlinjin:
Currently it takes upwards of 20s to load wallet from db (for the cli example). Probably something to look into for the next release.

example-crates/example_cli/src/lib.rs
```diff
@@ -664,45 +666,213 @@ where
         Commands::ChainSpecific(_) => unreachable!("example code should handle this!"),
         Commands::Address { addr_cmd } => {
             let graph = &mut *graph.lock().unwrap();
-            run_address_cmd(graph, db, network, addr_cmd)
```
danielabrozzoni, line 667: In d92709a:

Am I correct in saying that this is just a copy-paste of the internals of `run_address_cmd`? What's the reason behind this? If the method is not needed anymore, can we delete it?

example-crates/example_electrum/src/main.rs
```diff
@@ -66,8 +103,10 @@ type ChangeSet = (
 );
 
 fn main() -> anyhow::Result<()> {
+    // let (args, keymap, index, db, (disk_local_chain, disk_tx_graph)) =
```
danielabrozzoni: In d92709a: these two lines should probably be removed and not commented

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,243 @@
+//! This crate is used for emitting blockchain data from the `bitcoind` RPC interface (excluding the
```
danielabrozzoni: " from the `bitcoind` RPC interface (excluding the RPC wallet API)."

What do you mean? Is this saying that we don't use any wallet-related RPCs?

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+            client,
+            start_height,
+            emitted_blocks: BTreeMap::new(),
+            last_block: None,
+            last_mempool_time: 0,
+            last_mempool_tip: None,
+        }
+    }
+
+    /// Emit mempool transactions, alongside their first-seen unix timestamps.
```
danielabrozzoni, line 75: Mhh, why are we talking about first-seen timestamp here, while we usually (in the txgraph, from what I've seen) talk about last seen timestamp?

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+    ///
+    /// Ideally, this method would only emit the same transaction once. However, if the receiver
+    /// filters transactions based on whether it alters the output set of tracked script pubkeys,
+    /// there are situations where we would want to re-emit. For example, if an emitted mempool
+    /// transaction spends a tracked UTXO which is confirmed at height `h`, but the receiver has
+    /// only seen up to block of height `h-1`, we want to re-emit this transaction until the
+    /// receiver has seen the block at height `h`.
+    ///
+    /// In other words, we want to re-emit a transaction if we cannot guarantee it's ancestors are
```
danielabrozzoni: typo: its ancestors

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+        };
+
+        let prev_mempool_time = self.last_mempool_time;
+        let mut latest_time = prev_mempool_time;
+
+        let txs_to_emit = client
+            .get_raw_mempool_verbose()?
+            .into_iter()
+            .filter_map({
+                let latest_time = &mut latest_time;
```
danielabrozzoni, line 106: It seems to me that this line was put here just to please the compiler, but actually it works fine even if you remove it. Unless there's some logic difference that I don't see.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+            emitted_blocks: BTreeMap::new(),
+            last_block: None,
+            last_mempool_time: 0,
+            last_mempool_tip: None,
+        }
+    }
+
+    /// Emit mempool transactions, alongside their first-seen unix timestamps.
+    ///
+    /// Ideally, this method would only emit the same transaction once. However, if the receiver
```
danielabrozzoni: nit: It might be clearer if you put it in a different way, first explaining when a tx might be emitted more than once, and then giving the reason:
```
This method emits each transaction only once, unless when we cannot guarantee the transaction's ancestors
are already emitted.

To understand why, consider a receiver filtering transactions based on...
```

(The first sentence might need some rewording :))

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+///
+/// Refer to [module-level documentation] for more.
+///
+/// [module-level documentation]: crate
+pub struct Emitter<'c, C> {
+    client: &'c C,
+    start_height: u32,
+
+    emitted_blocks: BTreeMap<u32, BlockHash>,
+    last_block: Option<bitcoincore_rpc_json::GetBlockResult>,
```
danielabrozzoni, line 34: This needs a bit of documentation to explain when the option is None, as one would expect that it's None only if the Emitter never emitted blocks, but instead you set it to None in various scenarios (no more block, block not in the best chain, etc...).

danielabrozzoni: Same for emitted blocks, we should clarify that it's blocks emitted that are in the best chain, and so we can clear the map if we can't find agreement, or remove some blocks when they get reorged out

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -0,0 +1,759 @@
...
+        count: usize,
+        address: Option<Address>,
+    ) -> anyhow::Result<Vec<BlockHash>> {
+        let coinbase_address = match address {
+            Some(address) => address,
+            None => self.client.get_new_address(None, None)?.assume_checked(),
+        };
+        let block_hashes = self
+            .client
+            .generate_to_address(count as _, &coinbase_address)?;
```
danielabrozzoni, line 50: nit: If generate_to_address takes something different than a usize, why not having mine_blocks take that as well?

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -0,0 +1,759 @@
...
+        let res = self.mine_blocks(count, None);
+        assert_eq!(
+            self.client.get_block_count()?,
+            start_height,
+            "reorg should not result in height change"
+        );
+        res
+    }
+
+    fn reorg_empty_blocks(&self, count: usize) -> anyhow::Result<Vec<(usize, BlockHash)>> {
```
danielabrozzoni, line 137: nit: reorg_with_empty_blocks might be better, otherwise it seems that you're reorging out empty blocks

example-crates/example_bitcoind_rpc_polling/src/main.rs
```diff
@@ -0,0 +1,366 @@
...
+use example_cli::{
+    anyhow,
+    clap::{self, Args, Subcommand},
+    Keychain,
+};
+
+const DB_MAGIC: &[u8] = b"bdk_example_rpc";
+const DB_PATH: &str = ".bdk_example_rpc.db";
+
+const CHANNEL_BOUND: usize = 10;
```
danielabrozzoni, line 30: nit: needs docs :)

crates/bitcoind_rpc/src/lib.rs
evanlinjin: Yes, how do you recommend it worded?

crates/bitcoind_rpc/src/lib.rs
evanlinjin: last_seen will be copied into the closure and the original last_seen will stay the same.

crates/bitcoind_rpc/src/lib.rs
danielabrozzoni: This crate is used for emitting blockchain data from the `bitcoind` RPC interface. It doesn't use the wallet RPC API, so that you can use this crate even with wallet-disabled Bitcoin Core nodes.

How does this sound?

crates/bitcoind_rpc/src/lib.rs
danielabrozzoni: > The original last_seen will stay the same.

Is it? The new latest_time is just a mutable reference to the original latest_time, so you're modifying the original as well inside the closure.

But if I'm wrong, and in the closure you have a different latest_time that you're modifying only inside the closure, why is that needed? In the closure you only modify latest_time, you never read it.

danielabrozzoni: Sorry I didn't try, removing the line some tests fail, so it is needed. I just don't get why :(

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -209,7 +210,18 @@ where
...
-                emitter.emitted_blocks.split_off(&(res.height as u32 + 1));
+                let agreement_h = res.height as u32;
+
+                // get rid of evicted blocks
+                emitter.emitted_blocks.split_off(&(agreement_h + 1));
+
+                // The tip during the last mempool emission needs to in the best chain, we reduce
+                // it if it is not.
+                if let Some(h) = emitter.last_mempool_tip.as_mut() {
+                    if dbg!(*h) > dbg!(agreement_h) {
```
notmandatory: Did you mean to leave these `dbg!` macros here?

example-crates/example_cli/src/lib.rs
evanlinjin: Woops, duplicates are now removed.

crates/bitcoind_rpc/src/lib.rs
evanlinjin: latest_time implements copy, so it won't be moved into the closure, it'll get copied.

example-crates/example_bitcoind_rpc_polling/src/main.rs
evanlinjin: I'll change it, but it bugs me. I'm wondering whether we need to show developers how to use mutexes in our examples. For me, it would seem like common sense to know how to wrap any structure with mutexes/locks.

crates/bitcoind_rpc/src/lib.rs
evanlinjin: removed

example-crates/example_bitcoind_rpc_polling/src/main.rs
evanlinjin: Made an issue: https://github.com/bitcoindevkit/bdk/issues/1155

crates/bitcoind_rpc/src/lib.rs
evanlinjin: The RPC API provides the timestamp of when a tx is first seen in the mempool. It would still make sense to prioritize later first-seen timestamps during conflicts.

I think I tried using `seen_at` instead of `last_seen`.

Or we can argue we are still getting the last `first_seen` for a tx coming in from the mempool.

evanlinjin:
@danielabrozzoni Sorry, I'm going to ignore your nits due to how long this PR is taking. We can revise `TestEnv` in the future. 

I also think we should put `TestEnv` in a separate crate so it can be shared for testing chain source crates.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -72,97 +72,197 @@ impl<A: Anchor, I: Indexer> IndexedTxGraph<A, I>
...
     ///
-    /// Relevancy is determined by the [`Indexer::is_tx_relevant`] implementation of `I`. Irrelevant
-    /// transactions in `txs` will be ignored. `txs` do not need to be in topological order.
+    /// Relevancy is determined by the internal [`Indexer::is_tx_relevant`] implementation of `I`.
+    /// Irrelevant tansactions in `txs` will be ignored.
     ///
-    /// `anchors` can be provided to anchor the transactions to blocks. `seen_at` is a unix
-    /// timestamp of when the transactions are last seen.
-    pub fn insert_relevant_txs<'t>(
+    /// Items of `txs` are tuples containing the transaction and an optional *last seen* timestamp.
```
LLFourn:
```suggestion
    /// Items of `txs` are tuples containing the transaction and an *last seen* timestamp.
```

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -72,97 +72,197 @@ impl<A: Anchor, I: Indexer> IndexedTxGraph<A, I>
...
+                .map(|(tx, seen_at)| (tx.clone(), seen_at)),
         );
-        changeset
+
+        ChangeSet { graph, indexer }
+    }
+
+    /// Batch insert unconfirmed transactions.
+    ///
+    /// Items of `txs` are tuples containing the transaction and an optional *last seen* timestamp.
```
LLFourn:
```suggestion
    /// Items of `txs` are tuples containing the transaction and an *last seen* timestamp.
```

LLFourn (approved):
Thanks this looks excellent. It would be nice to know why it's taking 20s before merging though? Is it a function of the size of the db file -- how big is it?

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    /// The block result returned from rpc of the last-emitted block. As this result contains the
+    /// next block's block hash (which we use to fetch the next block), we set this to `None`
+    /// whenever there are no more blocks, or the next block is no longer in the best chain. This
+    /// gives us an opportunity to re-fetch this result.
+    last_block: Option<bitcoincore_rpc_json::GetBlockResult>,
+
+    /// The latest first-seen epoch of emitted mempool transactions. This is used to determine
+    /// whether a mempool transaction is already emitted.
+    last_mempool_time: usize,
```
vladimirfomene, line 38: Why define this as `usize`? Can we have it as `u64` to avoid conversion later?

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+/// [module-level documentation]: crate
+pub struct Emitter<'c, C> {
+    client: &'c C,
+    start_height: u32,
+
+    /// The checkpoint of the last-emitted block that is in the best chain. If it is later found
+    /// that the block is no longer in the best chain, it will be popped off from here.
+    last_cp: Option<CheckPoint>,
+
+    /// The block result returned from rpc of the last-emitted block. As this result contains the
```
vladimirfomene, line 30: I think using `GetBlockResult` will make this sentence clearer.

evanlinjin:
@LLFourn the file is 104MB, with a start height of 778,000

**UPDATE:** Okay I accidentally called `batch_insert_unconfirmed` instead of `batch_insert_relevant_confirmed`. So most of the content is mempool txs.

**UPDATE 2:** Yeah with these settings, when a wallet is fully synced, loading from db takes less than 100ms.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -72,97 +72,197 @@ impl<A: Anchor, I: Indexer> IndexedTxGraph<A, I>
...
+    /// Batch insert unconfirmed transactions.
+    ///
+    /// Items of `txs` are tuples containing the transaction and a *last seen* timestamp. The
+    /// *last seen* communicates when the transaction is last seen in the mempool which is used for
+    /// conflict-resolution in [`TxGraph`] (refer to [`TxGraph::insert_seen_at`] for details).
+    ///
+    /// To filter out irrelevant transactions, use [`batch_insert_relevant_unconfirmed`] instead.
+    ///
+    /// [`batch_insert_relevant_unconfirmed`]: IndexedTxGraph::batch_insert_relevant_unconfirmed
+    pub fn batch_insert_unconfirmed(
```
vladimirfomene, line 203: When will this be useful? Can we add docs around when a user will want to use this API?

example-crates/example_bitcoind_rpc_polling/src/main.rs
```diff
@@ -0,0 +1,368 @@
...
+const DB_MAGIC: &[u8] = b"bdk_example_rpc";
+const DB_PATH: &str = ".bdk_example_rpc.db";
+
+/// The mpsc channel bound for emissions from [`Emitter`].
+const CHANNEL_BOUND: usize = 10;
+/// Delay for printing status to stdout.
+const STDOUT_PRINT_DELAY: Duration = Duration::from_secs(6);
+/// Delay between mempool emissions.
+const MEMPOOL_EMIT_DELAY: Duration = Duration::from_secs(30);
+/// Delay for commiting to persistance.
```
vladimirfomene: Nit: committing

example-crates/example_bitcoind_rpc_polling/src/main.rs
```diff
@@ -0,0 +1,368 @@
...
+                            synced_to.hash(),
+                            synced_to.height(),
+                            balance.total()
+                        );
+                    }
+                }
+            }
+
+            let mempool_txs = emitter.mempool()?;
+            let graph_changeset = graph.lock().unwrap().batch_insert_unconfirmed(mempool_txs);
```
vladimirfomene: Why not use the relevant variant of the `batch_insert_unconfirmed` here?

vladimirfomene (commented):
Great work @evanlinjin ! Just a couple of questions and nits.

example-crates/example_bitcoind_rpc_polling/src/main.rs
evanlinjin: Great spot, yes this is a bug.

crates/chain/src/indexed_tx_graph.rs
evanlinjin: cc @LLFourn

evanlinjin: When we want to track transactions, but don't want to filter them out.

notmandatory (approved):
Re ACK 85c62532a55cfc94eade4d20ca3075dc4cd4882e

I don't see any critical issues remaining and any final nits can be fixed in follow-up PRs.

PR #1046, opened 2023-07-24
Avoid pinning dependencies, use --precise in ci

### LINKED ISSUE #1035: Log dependency fix makes use as library very hard
bdk depends on `log = "=0.4.18"`, which is really annoying for anyone having another version of `log` in their dependency tree.

danielabrozzoni:
Fixes #1035

#### All Submissions:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -71,7 +71,6 @@ jobs:
         uses: actions/checkout@v2
         # Install a recent version of clang that supports wasm32
       - run: wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add - || exit 1
-      - run: sudo apt-add-repository "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-10 main" || exit 1
       - run: sudo apt-get update || exit 1
       - run: sudo apt-get install -y libclang-common-10-dev clang-10 libc6-dev-i386 || exit 1
       - name: Install Rust toolchain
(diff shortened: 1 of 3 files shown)


crates/bdk/Cargo.toml
```diff
@@ -13,7 +13,7 @@ edition = "2021"
 rust-version = "1.57"
 
 [dependencies]
-log = "=0.4.18"
+log = "^0.4"
```
adoerr: Since a caret requirement is just another syntax for the default strategy, I would simply use `"0.4"` for consistency reasons

notmandatory:
LGTM. One thing to add is a short MSRV section in `bdk/README.md` similar to what the `rust-bitcoin` project did (https://github.com/rust-bitcoin/rust-bitcoin/blob/master/README.md#minimum-supported-rust-version-msrv). For example:

## Minimum Supported Rust Version (MSRV)

This library should compile with any combination of features with **Rust 1.57.0**.

To build with the MSRV you will need to pin dependencies as follows:

```
# log 0.4.19 has MSRV 1.60.0+
cargo update -p log --precise "0.4.18"
# tempfile 3.7.0 has MSRV 1.63.0
cargo update -p tempfile --precise "3.6.0"
```

.github/workflows/cont_integration.yml
```diff
@@ -27,6 +27,9 @@ jobs:
             profile: minimal
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2.2.1
+      - name: Pin dependencies
```
notmandatory:
```suggestion
      - name: Pin dependencies for MSRV
```

danielabrozzoni:
Thanks! Fixed all your comments, and added a fix for the wasm build.

adoerr (approved):
LGTM

evanlinjin (approved):
ACK ffb7c795e15d2d879f2d857e7607729ca4a53155

PR #1048, opened 2023-07-27
 Remove TransactionDetails from Wallet API

### LINKED ISSUE #922: Kill TransactionDetails
Kill TransactionDetails, get details of txs via methods from Wallet instead. Ensure rich data we have from `bdk_chain` can be observed from `Wallet`.

### LINKED ISSUE #1015: Wallet::sent_and_received doesn't exist
This kind of utility function should be exposed to users of `Wallet`. others like `balance` have been and this is technically a regression since `TransactionDetails` used to offer this information.

There may be opportunities to further delegate this to the txout index elsewhere in wallet since I see that "sent" and "received" being calculated internally in places.

The same is true for `calculate_fee` of `TxGraph` -- it should be there on `Wallet`.

notmandatory:
### Description

Removed `TransactionDetails` and changed `Wallet::get_tx` to return a `CanonicalTx`, and `TxBuilder::finish` to return only a `PartiallySignedTransaction`. This should fix #922 and fix #1015.  

I also added `Wallet` functions to get a `Transaction` send and receive amounts, fee, and `FeeRate`.

see: https://github.com/bitcoindevkit/bdk/issues/922#issuecomment-1652814975

### Notes to the reviewers

Alot of wallet tests had to change since `TxBuilder::finish` only returns a PSBT now. 

I added a new `CalculateFeeError` which follows changes coming in #1028.

### Changelog notice

Added
- Wallet::sent_and_received function
- Wallet::calculate_fee and Wallet::calculate_fee_rate functions
- Wallet::error::CalculateFeeError
- Wallet::insert_txout function to allow inserting foreign TxOuts

BREAKING CHANGES:

Removed
- TransactionDetails struct

Changed
- Wallet::get_tx now returns CanonicalTx instead of TransactionDetails
- TxBuilder::finish now returns only a PartiallySignedTransaction

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -40,6 +40,7 @@ use core::fmt;
 use core::ops::Deref;
 use miniscript::psbt::{PsbtExt, PsbtInputExt, PsbtInputSatisfier};
 
+use bdk_chain::tx_graph::CalculateFeeError;
 #[allow(unused_imports)]
 use log::{debug, error, info, trace};
 
@@ -66,7 +67,7 @@ use crate::descriptor::{
     calc_checksum, into_wallet_descriptor_checked, DerivedDescriptor, DescriptorMeta,
     ExtendedDescriptor, ExtractPolicy, IntoWalletDescriptor, Policy, XKeyUtils,
 };
-use crate::error::{CalculateFeeError, Error, MiniscriptPsbtError};
+use crate::error::{Error, MiniscriptPsbtError};
 use crate::psbt::PsbtUtils;
 use crate::signer::SignerError;
 use crate::types::*;
@@ -434,11 +435,7 @@ impl<D> Wallet<D> {
     ///
     /// Note `tx` does not have to be in the graph for this to work.
     pub fn calculate_fee(&self, tx: &Transaction) -> Result<u64, CalculateFeeError> {
-        match self.indexed_graph.graph().calculate_fee(tx) {
-            None => Err(CalculateFeeError::MissingTxOut),
-            Some(fee) if fee < 0 => Err(CalculateFeeError::NegativeFee(fee)),
-            Some(fee) => Ok(u64::try_from(fee).unwrap()),
-        }
+        self.indexed_graph.graph().calculate_fee(tx)
     }
 
     /// Calculate the `FeeRate` for a given transaction.
@@ -1072,13 +1069,12 @@ impl<D> Wallet<D> {
             return Err(Error::IrreplaceableTransaction);
         }
 
-        let fee = graph.calculate_fee(&tx).ok_or(Error::FeeRateUnavailable)?;
-        if fee < 0 {
-            // It's available but it's wrong so let's say it's unavailable
-            return Err(Error::FeeRateUnavailable)?;
-        }
-        let fee = fee as u64;
-        let feerate = FeeRate::from_wu(fee, tx.weight());
+        let fee = self
+            .calculate_fee(&tx)
+            .map_err(|_| Error::FeeRateUnavailable)?;
+        let fee_rate = self
+            .calculate_fee_rate(&tx)
+            .map_err(|_| Error::FeeRateUnavailable)?;
 
         // remove the inputs from the tx and process them
         let original_txin = tx.input.drain(..).collect::<Vec<_>>();
@@ -1162,7 +1158,7 @@ impl<D> Wallet<D> {
             utxos: original_utxos,
             bumping_fee: Some(tx_builder::PreviousFee {
                 absolute: fee,
-                rate: feerate.as_sat_per_vb(),
+                rate: fee_rate.as_sat_per_vb(),
             }),
             ..Default::default()
         };
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -431,10 +431,17 @@ impl<D> Wallet<D> {
             .next()
     }
 
-    /// Inserts the given foreign `TxOut` at `OutPoint` into the wallet's transaction graph. Any
-    /// inserted foreign TxOuts are not persisted until [`Self::commit`] is called.
+    /// Inserts a [`TxOut`] at [`OutPoint`] into the wallet's transaction graph.
+    /// Any inserted TxOuts are not persisted until [`commit`] is called.
+    ///
+    /// This can be used to add a `TxOut` that the wallet doesn't own but is used as an input to
+    /// a [`Transaction`] passed to the [`calculate_fee`] or [`calculate_fee_rate`] functions.
     ///
     /// Only insert TxOuts you trust the values for!
+    ///
+    /// [`calculate_fee`]: Self::calculate_fee
+    /// [`calculate_fee_rate`]: Self::calculate_fee_rate
+    /// [`commit`]: Self::commit
     pub fn insert_txout(&mut self, outpoint: OutPoint, txout: TxOut)
     where
         D: PersistBackend<ChangeSet>,
@@ -445,12 +452,30 @@ impl<D> Wallet<D> {
 
     /// Calculates the fee of a given transaction. Returns 0 if `tx` is a coinbase transaction.
     ///
-    /// To calculate the fee for a [`Transaction`] that depends on foreign [`TxOut`] values you must
-    /// first manually insert the foreign TxOuts into the tx graph using the [`insert_txout`] function.
-    /// Only insert TxOuts you trust the values for!
+    /// To calculate the fee for a [`Transaction`] with inputs not owned by this wallet you must
+    /// manually insert the TxOut(s) into the tx graph using the [`insert_txout`] function.
     ///
     /// Note `tx` does not have to be in the graph for this to work.
     ///
+    /// # Examples
+    ///
+    /// ```rust, no_run
+    /// # use bitcoin::Txid;
+    /// # use bdk::Wallet;
+    /// # let mut wallet: Wallet<()> = todo!();
+    /// # let txid:Txid = todo!();
+    /// let tx = wallet.get_tx(txid).expect("transaction").tx_node.tx;
+    /// let fee = wallet.calculate_fee(tx).expect("fee");
+    /// ```
+    ///
+    /// ```rust, no_run
+    /// # use bitcoin::psbt::PartiallySignedTransaction;
+    /// # use bdk::Wallet;
+    /// # let mut wallet: Wallet<()> = todo!();
+    /// # let mut psbt: PartiallySignedTransaction = todo!();
+    /// let tx = &psbt.clone().extract_tx();
+    /// let fee = wallet.calculate_fee(tx).expect("fee");
+    /// ```
     /// [`insert_txout`]: Self::insert_txout
     pub fn calculate_fee(&self, tx: &Transaction) -> Result<u64, CalculateFeeError> {
         self.indexed_graph.graph().calculate_fee(tx)
@@ -458,12 +483,30 @@ impl<D> Wallet<D> {
 
     /// Calculate the [`FeeRate`] for a given transaction.
     ///
-    /// To calculate the fee rate for a [`Transaction`] that depends on foreign [`TxOut`] values you
-    /// must first manually insert the foreign TxOuts into the tx graph using the [`insert_txout`] function.
-    /// Only insert TxOuts you trust the values for!
+    /// To calculate the fee rate for a [`Transaction`] with inputs not owned by this wallet you must
+    /// manually insert the TxOut(s) into the tx graph using the [`insert_txout`] function.
     ///
     /// Note `tx` does not have to be in the graph for this to work.
     ///
+    /// # Examples
+    ///
+    /// ```rust, no_run
+    /// # use bitcoin::Txid;
+    /// # use bdk::Wallet;
+    /// # let mut wallet: Wallet<()> = todo!();
+    /// # let txid:Txid = todo!();
+    /// let tx = wallet.get_tx(txid).expect("transaction").tx_node.tx;
+    /// let fee_rate = wallet.calculate_fee_rate(tx).expect("fee rate");
+    /// ```
+    ///
+    /// ```rust, no_run
+    /// # use bitcoin::psbt::PartiallySignedTransaction;
+    /// # use bdk::Wallet;
+    /// # let mut wallet: Wallet<()> = todo!();
+    /// # let mut psbt: PartiallySignedTransaction = todo!();
+    /// let tx = &psbt.clone().extract_tx();
+    /// let fee_rate = wallet.calculate_fee_rate(tx).expect("fee rate");
+    /// ```
     /// [`insert_txout`]: Self::insert_txout
     pub fn calculate_fee_rate(&self, tx: &Transaction) -> Result<FeeRate, CalculateFeeError> {
         self.calculate_fee(tx).map(|fee| {
@@ -473,10 +516,31 @@ impl<D> Wallet<D> {
     }
 
     /// Computes total input value going from script pubkeys in the index (sent) and the total output
-    /// value going to script pubkeys in the index (received) in `tx`. For the `sent` to be computed
-    /// correctly, the output being spent must have already been scanned by the index. Calculating
-    /// received just uses the [`Transaction`] outputs directly, so it will be correct even if it has
-    /// not been scanned.
+    /// value going to script pubkeys in the index (received) in `tx`.
+    ///
+    /// For the `sent` to be computed correctly, the outputs being spent must have already been
+    /// scanned by the index. Calculating received just uses the [`Transaction`] outputs directly,
+    /// so it will be correct even if it has not been scanned.
+    ///
+    /// # Examples
+    ///
+    /// ```rust, no_run
+    /// # use bitcoin::Txid;
+    /// # use bdk::Wallet;
+    /// # let mut wallet: Wallet<()> = todo!();
+    /// # let txid:Txid = todo!();
+    /// let tx = wallet.get_tx(txid).expect("transaction").tx_node.tx;
+    /// let (sent, received) = wallet.sent_and_received(tx);
+    /// ```
+    ///
+    /// ```rust, no_run
+    /// # use bitcoin::psbt::PartiallySignedTransaction;
+    /// # use bdk::Wallet;
+    /// # let mut wallet: Wallet<()> = todo!();
+    /// # let mut psbt: PartiallySignedTransaction = todo!();
+    /// let tx = &psbt.clone().extract_tx();
+    /// let (sent, received) = wallet.sent_and_received(tx);
+    /// ```
     pub fn sent_and_received(&self, tx: &Transaction) -> (u64, u64) {
         self.indexed_graph.index.sent_and_received(tx)
     }
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -66,7 +66,7 @@ use crate::descriptor::{
     calc_checksum, into_wallet_descriptor_checked, DerivedDescriptor, DescriptorMeta,
     ExtendedDescriptor, ExtractPolicy, IntoWalletDescriptor, Policy, XKeyUtils,
 };
-use crate::error::{Error, MiniscriptPsbtError};
+use crate::error::{CalculateFeeError, Error, MiniscriptPsbtError};
 use crate::psbt::PsbtUtils;
 use crate::signer::SignerError;
 use crate::types::*;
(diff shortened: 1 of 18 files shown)


LLFourn:
I think the plan is to deprecate `TransactionDetails` altogether. What about just doing #1015 or is the plan to change this test to test the things in #1015 when they're added?

EDIT: this comment was made before seeing the discussion in #1015 -- I've commented over there.

crates/bdk/src/wallet/mod.rs
```diff
@@ -427,27 +428,97 @@ impl<D> Wallet<D> {
...
+        let fee_amount = self.fee_amount(tx);
+        fee_amount.map(|fee| {
+            let weight = tx.weight();
+            FeeRate::from_wu(fee, weight)
+        })
+    }
+
+    /// Return `TransactionAmounts` for a `Transaction` in relation to the `Wallet` and it's
+    /// descriptors.
+    pub fn sent_and_received(&self, tx: &Transaction) -> TransactionAmounts {
```
notmandatory: Can I rename this to `tx_amounts`? the name `sent_and_received` is a bit long an awkward.

danielabrozzoni: ACK

crates/bdk/src/wallet/mod.rs
```diff
@@ -427,27 +427,97 @@ impl<D> Wallet<D> {
...
+            .graph()
+            .all_txouts()
+            .filter_map(|(op, txo)| {
+                if op.eq(outpoint) {
+                    Some(txo.clone())
+                } else {
+                    None
+                }
+            })
+            .next()
```
danielabrozzoni: nit: you can use `find_map`
```
    pub fn get_txout(&self, outpoint: &OutPoint) -> Option<TxOut> {
        self.indexed_graph
            .graph()
            .all_txouts()
            .find_map(|(op, txo)| {
                if op.eq(outpoint) {
                    Some(txo.clone())
                } else {
                    None
                }
            })
    }
```

danielabrozzoni (approved):
I just have a nit 

ACK 40d5af39e256c2df3ea468e5f6cc32e5898cc77d

crates/bdk/src/wallet/mod.rs
notmandatory: Thanks for the tip! but while fixing it I found that there's already a `TxGraph::get_txout()` function so I'm using that instead and was able to remove this whole function.

notmandatory:
@LLFourn continuing https://github.com/bitcoindevkit/bdk/issues/1015#issuecomment-1657935215 here. 

Are you ok with the name `TotalSentReceived` for the struct and `total_sent_received` for the function? I just don't think `and` in the names is as useful as `total`. But if you really hate it I can live with `SentAndReceived`/`sent_and_received`.  

I'm now calling the `TxGraph::calculate_fee` function but added a friendlier `Result` return type and new `CalculateFeeError`. If you like this approach should I move it down to TxGraph?

crates/bdk/src/wallet/mod.rs
```diff
@@ -427,27 +427,49 @@ impl<D> Wallet<D> {
...
     ///
-    /// Optionally fill the [`TransactionDetails::transaction`] field with the raw transaction if
-    /// `include_raw` is `true`.
-    pub fn get_tx(&self, txid: Txid, include_raw: bool) -> Option<TransactionDetails> {
+    /// Note `tx` does not have to be in the graph for this to work.
+    pub fn calculate_fee(&self, tx: &Transaction) -> Result<u64, CalculateFeeError> {
+        match self.indexed_graph.graph().calculate_fee(tx) {
+            None => Err(CalculateFeeError::MissingTxOut),
+            Some(fee) if fee < 0 => Err(CalculateFeeError::NegativeFee(fee)),
+            Some(fee) => Ok(u64::try_from(fee).unwrap()),
```
LLFourn: I think this API is better and it should just be pushed down to `TxGrpah`. @evanlinjin?

crates/chain/src/spk_txout_index.rs
```diff
@@ -334,4 +343,33 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
             .any(|output| self.spk_indices.contains_key(&output.script_pubkey));
         input_matches || output_matches
     }
+
+    /// Return `TotalSentReceived` for a `Transaction` in relation to the `SpkTxOutIndex`.
```
LLFourn: Isn't this meant to replace `sent_and_received`? (it's still here).
After seeing this I think the tuple `(u64, u64)` as it was before since  has the order of the returned value is in the name `total_sent_received` which implies `sent` is being returned before received. Not a major concern so feel free to go with whatever feels right.

In any case please just rename the existing function `sent_and_received` and change the output type (if you still want a custom type).

LLFourn (commented):
Concept ACK. See minor comments.

crates/chain/src/spk_txout_index.rs
notmandatory: Hey sorry I just realized you already had the `sent_and_received` function implemented in the `chain` crate. Makes sense so I just exposed it as it is there in the `bdk` API with the same name and tuple result type.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -474,7 +482,7 @@ fn test_calculate_fee_on_coinbase() {
 
     let graph = TxGraph::<()>::default();
 
-    assert_eq!(graph.calculate_fee(&tx), Some(0));
+    assert_matches!(graph.calculate_fee(&tx), Ok(0));
```
LLFourn: nit `assert_eq` here? I think you could make all of these `assert_eq` and avoid the dev dependency.

crates/bdk/src/wallet/mod.rs
```diff
@@ -969,20 +991,8 @@ impl<D> Wallet<D> {
...
-        let sent = coin_selection.local_selected_amount();
         let psbt = self.complete_transaction(tx, coin_selection.selected, params)?;
-
-        let transaction_details = TransactionDetails {
-            transaction: None,
-            txid,
-            confirmation_time: ConfirmationTime::Unconfirmed { last_seen: 0 },
-            received,
-            sent,
-            fee: Some(fee_amount),
```
LLFourn, line 989: I am concerned that we can't reproduce this `fee` value unless we return it here. When you use `add_foreign_utxo` to the builder you are telling it the value of the input. This won't exist in the `TxGraph` (although it could we have a `mut self` here!).

So we can:
1) add back the second return value to return the fee
2) Keep only one return value but insert the `TxOut`s into the `TxGraph` so we can later query for fee and get the right answer (they would be un-anchored  i.e. assumed to be in the mempool). Thse txouts would be committed to disk so we could query the fee of this tx later on.
3) Require the user inserts the foreign utxo into the wallet manually before they can use it in `add_foreign_utxo`.
4) Simply recommend that the user always insert the foreign UTXO into the wallet manually if they want to be able to calculate the fees of the resulting transaction. This makes the user think about whether they can actually trust the values they have in this UTXO.

I think I lean towards (4) because we have planning module integration coming up and that may change transaction building a lot. I notice that in order to do (3) or (4) we'd need to add `insert_txout` method on wallet that calls the graph's method.

crates/chain/tests/test_tx_graph.rs
notmandatory: Fixed, I was getting an error before but it was because I forgot to derive `Eq`, `PartialEq` for `CalculateFeeError`, it works now without the extra dev dep.

crates/bdk/src/wallet/mod.rs
notmandatory: Interesting problem. I agree the last option (4) makes sense, I can add some documentation to explain the situation. I'll also add a test to reproduce and confirm it works as expected.

crates/bdk/src/wallet/mod.rs
evanlinjin: Personally I prefer `sent_and_received` as I find it more descriptive.

crates/bdk/src/wallet/mod.rs
evanlinjin: The other option will change the `TxGraph::calculate_fee` method to include missing outputs.

```rust
pub fn calculate_fee(
    &self, 
    tx: &Transactions, 
    missing_outputs: &HashMap<OutPoint, TxOut>,
) -> Result<i64, CalculateFeeError> {
    unimplemented!()
}
```

Edit: rethinking about it, I think 4 is still the most practical solution. My suggestion will by clunky after a tx is created and the caller wants to calculate the fee.

crates/bdk/src/wallet/mod.rs
evanlinjin: I think that will be nice. To improve the API further, I would have the `CalculateFeeError::MissingTxOut` specify the missing txout outpoints as well.

notmandatory:
Rebased post #1023 and #1034 being merged.

crates/bdk/src/wallet/mod.rs
notmandatory: In a future `1.x` release we could add a `calculate_fee_with_outputs` function is there's any demand for it.

crates/bdk/src/wallet/mod.rs
```diff
@@ -428,27 +429,48 @@ impl<D> Wallet<D> {
             .next()
     }
 
-    /// Return a single transactions made and received by the wallet
+    /// Calculates the fee of a given transaction. Returns 0 if `tx` is a coinbase transaction.
     ///
-    /// Optionally fill the [`TransactionDetails::transaction`] field with the raw transaction if
-    /// `include_raw` is `true`.
-    pub fn get_tx(&self, txid: Txid, include_raw: bool) -> Option<TransactionDetails> {
+    /// Note `tx` does not have to be in the graph for this to work.
```
evanlinjin, line 458: Should we note that we need the outputs that this tx spends to exist in the internal tx graph for this to work?

notmandatory: Yup, I'll add the same note I'm adding to the `TxGraph::calculate_fee` to any functions that use it.

crates/bdk/src/wallet/mod.rs
```diff
@@ -428,27 +429,48 @@ impl<D> Wallet<D> {
...
+    /// value going to script pubkeys in the index (received) in `tx`. For the `sent` to be computed
+    /// correctly, the output being spent must have already been scanned by the index. Calculating
+    /// received just uses the transaction outputs directly, so it will be correct even if it has not
+    /// been scanned.
+    pub fn sent_and_received(&self, tx: &Transaction) -> (u64, u64) {
+        self.indexed_graph.index.sent_and_received(tx)
+    }
+
+    /// Return a single `CanonicalTx` made and received by the wallet or `None` if it doesn't
+    /// exist in the wallet
```
evanlinjin: Because of the complexity of the `CanonicalTx` structure, I think this method deserves documentation with a bit more detail. Let me know if this will suffice.

```suggestion
    /// Get a single transaction from the wallet as a [`CanonicalTx`] (if the transaction exists).
    ///
    /// `CanonicalTx` contains the full transaction alongside metadata such as:
    /// * Blocks that the transaction is [`Anchor`]ed in. These may or may not be blocks that exist
    ///   in the best chain.
    /// * The [`ChainPosition`] of the transaction in the best chain - whether the transaction is
    ///   confirmed or unconfirmed. If the transaction is confirmed, the anchor which proves the
    ///   confirmation is provided. If the transaction is unconfirmed, the unix timestamp of when
    ///   the transaction was last seen in the mempool is provided.
    ///
    /// ```rust, no_run
    /// use bdk::{chain::ChainPosition, Wallet};
    /// # let wallet: Wallet<()> = todo!();
    /// # let my_txid: bdk::bitcoin::Txid = todo!();
    ///
    /// let canonical_tx = wallet.get_tx(my_txid).expect("panic if tx does not exist");
    ///
    /// // get reference to full transaction
    /// println!("my tx: {:#?}", canonical_tx.node.tx);
    ///
    /// // list all transaction anchors
    /// for anchor in canonical_tx.mode.anchors {
    ///     println!("tx is anchored by block of hash {}", anchor.anchor_block().hash);
    /// }
    ///
    /// // get confirmation status of transaction
    /// match canonical_tx.observed_as {
    ///     ChainPosition::Confirmed(anchor) => println!(
    ///         "tx is confirmed at height {}, we know this since {}:{} is in the best chain",
    ///         anchor.confirmation_height,
    ///         anchor.anchor_block.height,
    ///         anchor.anchor_block.hash,
    ///     ),
    ///     ChainPosition::Unconfirmed(last_seen) => println!(
    ///         "tx is last seen at {}, it is unconfirmed as it is not anchored in the best chain",
    ///         last_seen,
    ///     ),
    /// }
    /// ```
    ///
    /// [`Anchor`]: bdk_chain::Anchor
```

crates/bdk/src/wallet/mod.rs
notmandatory: To test the wallet calculate_fee  and fee_rate with and without foreign utxos I added `Wallet::insert_txout`, which exposes the underlying TxGraph::insert_txout. For this new function I'm also persisting the change, as long as the txout data is trusted this is OK isn't it? 

I also updated docs with warnings about missing utxos, and to trust any added. Is it clear enough?

crates/chain/src/tx_graph.rs
```diff
@@ -241,33 +250,50 @@ impl<A> TxGraph<A> {
...
+                    (sum, missing_outpoints)
+                }
+            },
+        );
+
+        let inputs_sum = if inputs_sum.1.is_empty() {
+            Ok(inputs_sum.0 as i64)
+        } else {
+            Err(CalculateFeeError::MissingTxOut(inputs_sum.1))
+        }?;
```
evanlinjin: Nit:
```suggestion
        let (inputs_sum, missing_outputs) = tx.input.iter().fold(
            (0_i64, Vec::new()),
            |(mut sum, mut missing_outpoints), txin| match self.get_txout(txin.previous_output) {
                None => {
                    missing_outpoints.push(txin.previous_output);
                    (sum, missing_outpoints)
                }
                Some(txout) => {
                    sum += txout.value as i64;
                    (sum, missing_outpoints)
                }
            },
        );
        if !missing_outputs.is_empty() {
            return Err(CalculateFeeError::MissingTxOut(missing_outputs));
        }
```

crates/bdk/tests/wallet.rs
```diff
@@ -88,6 +87,62 @@ fn test_get_funded_wallet_balance() {
...
+    let (wallet, _) = get_funded_wallet(get_test_wpkh());
+    assert_eq!(wallet.get_balance().confirmed, 50000);
+    let mut tx_amounts: Vec<(Txid, (u64, u64))> = wallet
+        .transactions()
+        .map(|ct| (ct.node.txid, wallet.sent_and_received(ct.node.tx)))
+        .collect();
+    tx_amounts.sort_by(|a1, a2| a1.0.cmp(&a2.0));
+
+    assert_eq!(tx_amounts.len(), 2);
+    assert_matches!(tx_amounts.get(0), Some((_, (76_000, 50_000))))
```
evanlinjin: How did we know what values we are expecting here?

crates/bdk/tests/wallet.rs
```diff
@@ -88,6 +87,62 @@ fn test_get_funded_wallet_balance() {
...
+        })
+        .collect();
+    tx_fee_amounts.sort_by(|a1, a2| a1.0.cmp(&a2.0));
+
+    assert_eq!(tx_fee_amounts.len(), 2);
+    assert_matches!(
+        tx_fee_amounts.get(1),
+        Some((_, Err(CalculateFeeError::MissingTxOut(_))))
+    );
+    assert_matches!(tx_fee_amounts.get(0), Some((_, Ok(1000))))
```
evanlinjin: How did we know this fees here? Would it be possible to send a tx from `bitcoind` and have `bitcoind` give us the fee/feerate and we can compare the result with what we get from BDK?

evanlinjin (changes requested):
I'm liking how this looks. I've got some suggestions to documentation and tests. I'll like to give an ACK after all is addressed (or dismissed).

crates/bdk/tests/wallet.rs
notmandatory: Thanks for catching this, I got sloppy here. Will fix it and add better docs too.

crates/bdk/tests/wallet.rs
notmandatory: Since we're just calling rust-bitcoin's calculations here I don't think we need to re-validate their code. But I can document the calculation for anyone who wants to see where the final number comes from.

crates/bdk/src/wallet/mod.rs
```diff
@@ -428,27 +429,115 @@ impl<D> Wallet<D> {
...
-    /// `include_raw` is `true`.
-    pub fn get_tx(&self, txid: Txid, include_raw: bool) -> Option<TransactionDetails> {
+    /// Only insert TxOuts you trust the values for!
+    pub fn insert_txout(&mut self, outpoint: OutPoint, txout: TxOut) -> Result<(), D::WriteError>
+    where
+        D: PersistBackend<ChangeSet>,
+    {
+        let additions = self.indexed_graph.insert_txout(outpoint, &txout);
+        self.persist.stage(ChangeSet::from(additions));
+        self.persist.commit()?;
```
LLFourn: We are not meant to commit here. What if you want to insert a bunch of txouts? This would do IO for every single call.

notmandatory: Makes sense. I removed the commit() and Result<>, and updated the docs.

crates/bdk/tests/wallet.rs
```diff
@@ -1002,11 +1049,13 @@ fn test_add_foreign_utxo() {
         .only_witness_utxo()
         .add_foreign_utxo(utxo.outpoint, psbt_input, foreign_utxo_satisfaction)
         .unwrap();
-    let (mut psbt, details) = builder.finish().unwrap();
+    let mut psbt = builder.finish().unwrap();
+    let fee = psbt.fee_amount();
```
LLFourn: :bulb: : We are using `psbt.fee_amount()` a lot in all of these tests but we also introduced our own fee calc method in this PR. Maybe we can test for free our one against the PSBT one with some macro like `let fee = check_fee!(wallet, psbt)` which would internally assert that psbt.fee_amount was the same as what the wallet calculates and then returns it.

crates/bdk/src/wallet/mod.rs
```diff
@@ -430,27 +431,113 @@ impl<D> Wallet<D> {
...
+    pub fn calculate_fee_rate(&self, tx: &Transaction) -> Result<FeeRate, CalculateFeeError> {
+        self.calculate_fee(tx).map(|fee| {
+            let weight = tx.weight();
+            FeeRate::from_wu(fee, weight)
+        })
+    }
+
+    /// Computes total input value going from script pubkeys in the index (sent) and the total output
+    /// value going to script pubkeys in the index (received) in `tx`. For the `sent` to be computed
+    /// correctly, the output being spent must have already been scanned by the index. Calculating
```
```suggestion
    /// correctly, the outputs being spent must have already been scanned by the index. Calculating
```

crates/bdk/src/wallet/mod.rs
LLFourn: Nit: I'm concerned that the notion of a "foreign" txout is ill-defined and will be short lived. It should just say here that the txout must have been inserted into the wallet which won't happen naturally if the transaction has inputs that were not owned by this wallet (as you mention you can remedy this by `insert_txout`). The specific documentation that this problem can occur for you when you use `add_foreign_utxo` in the builder should reside on `add_foreign_utxo` to avoid using the term "foreign" here. Perhaps reference `add_foreign_utxo` here and the likely source of your problems in this doc.

LLFourn (approved):
ACK 1992c62efd4c68a062f91f9d02da685c25ba1b40

left some nits

crates/bdk/src/wallet/mod.rs
danielabrozzoni:
```
    /// match canonical_tx.observed_as {
    ///     ChainPosition::Confirmed(anchor) => println!(
    ///         "tx is confirmed at height {}, we know this since {}:{} is in the best chain",
    ///         anchor.confirmation_height,
    ///         anchor.anchor_block.height,
    ///         anchor.anchor_block.hash,
    ///     ),
```

Is this correct? I'm still confused about anchors etc, but from my understanding not all the anchors of a transaction have to be in the best chain. A transaction could have an anchor A1 in the best chain, and an anchor A2 which is not in the best chain - in this case, we know that the tx is confirmed because of A1, but we can't say "tx is confirmed at height h2, we know this since A2 is in the best chain"

notmandatory: What I understand this is saying is that if the `canonical_tx.observed_as` is `ChainPosition::Confirmed` then it will contain the anchor that is in the best chain even if there may be other anchors that aren't in the best chain.  If the `canonical_tx` isn't anchored to the best chain it will have `ChainPosition::Unconfirmed` with the last seen time.  The magic looks like it happens in `TxGraph::try_get_chain_position`.

crates/bdk/src/wallet/mod.rs
notmandatory: How do these changes look? https://github.com/bitcoindevkit/bdk/pull/1048/commits/ec707899049274fe091ae568a068cd6a1ca02dc3

crates/bdk/src/wallet/mod.rs
evanlinjin: Yes, the `ChainPosition::Confirmed` variant only contains one anchor, which is the anchor that anchors the tx in the best chain.

crates/bdk/tests/wallet.rs
notmandatory: Good idea, I added it in https://github.com/bitcoindevkit/bdk/pull/1048/commits/f1d868d94ae02e6b30771ace01c84b3bed98c839, and was able to remove the test I added for checking fees with a foreign utxo that's now covered by the existing txbuilder foreign utxo tests.

crates/bdk/src/wallet/mod.rs
```diff
@@ -430,27 +431,118 @@ impl<D> Wallet<D> {
...
+            let weight = tx.weight();
+            FeeRate::from_wu(fee, weight)
+        })
+    }
+
+    /// Computes total input value going from script pubkeys in the index (sent) and the total output
+    /// value going to script pubkeys in the index (received) in `tx`. For the `sent` to be computed
+    /// correctly, the outputs being spent must have already been scanned by the index. Calculating
+    /// received just uses the [`Transaction`] outputs directly, so it will be correct even if it has
+    /// not been scanned.
```
evanlinjin: Nit: I think it will be better to have two sections for these docs.

https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

```suggestion
    /// Computes total input value going from script pubkeys in the index (sent) and the total output
    /// value going to script pubkeys in the index (received) in `tx`.
    ///
    /// For the `sent` to be computed correctly, the outputs being spent must have already been scanned
    /// by the index. Calculating received just uses the [`Transaction`] outputs directly, so it will
    /// be correct even if it has not been scanned.
```

notmandatory: I hadn't read the documentation guide, good suggestion. I also added some simple examples.

notmandatory:
I've rebased, fixed docs and fixed a new MSRV issue.  Just need @evanlinjin to ACK and then I'll merge, which will also fix `master` branch CI.

crates/bdk/src/wallet/mod.rs
notmandatory: Fixed docs in 465d53cc88b32bf8098dc61e9afdbb28d22b213d (rebased with other suggestions from @evanlinjin ).

evanlinjin (approved):
ACK 5fb5061645ae92d37091a215e48e84423fea48a3

PR #1033, opened 2023-07-19
Backport new taproot descriptor template (BIP86)

notmandatory:
### Description

This PR solves bitcoindevkit/bdk_wallet#169 for the release/0.28 branch. This PR adds a P2TR descriptor template and a BIP86 taproot descriptor template. With this, users can now create a taproot descriptor with templates.

### Notes to the reviewers

The commit from bitcoindevkit/bdk#840 is cherry-picked from the `master` branch to the `release/0.28` branch without any changes.

### Changelog notice

Add taproot descriptor template (BIP-86).

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [x] I've added tests for the new feature
* [x] I've added docs for the new feature

PR #1008, opened 2023-06-16
Backporting "Fix policy condition calculation" onto release/0.28

junderw:
Fixes #942

### Description

This backports "Fix policy condition calculation" (#932) onto release/0.28

### Notes to the reviewers

Currently kept the Author the same and the committer is myself. Let me know if this is not the desired outcome of the history. (I have made no modifications myself)

### Changelog notice

Backported "Fix policy condition calculation"

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #840, opened 2023-01-23
create taproot descriptor template

vladimirfomene:
### Description

This PR solves bitcoindevkit/bdk_wallet#169. This PR adds a P2TR
descriptor template and a BIP86 taproot
descriptor template. With this, users
can now create a taproot descriptor with templates.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [x] I've added tests for the new feature
* [x] I've added docs for the new feature

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [x] I'm linking the issue being fixed by this PR

PR #1057, opened 2023-08-03
release: bump version to 0.28.1

notmandatory:
Summary

This patch release backports (from the BDK 1.0 dev branch) a fix for a bug in the policy condition calculation and adds a new taproot single key descriptor template (BIP-86). The policy condition calculation bug can cause issues when a policy subtree fails due to missing info even if it's not selected when creating a new transaction, errors on unused policy paths are now ignored.

Fixed

- Backported #932 fix for policy condition calculation #1008

Added

-  Backported #840 taproot descriptor template (BIP-86) #1033

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [v0.28.1]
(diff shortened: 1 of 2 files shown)


thunderbiscuit (approved):
ACK 7231039e8144f1eb27f4114c942b2b0cf793f071.

PR #964, opened 2023-05-03
Test conflicting tx situations.

rajarshimaitra:
### Description

partial fixes #958 . These tests attempt to cover all the conflicting scenarios. 

The following tests are included:

 - Mempool conflicts are chosen by the latest last seen.
 - Conflicts in mempool and orphaned block filters out the orphaned block tx.
 - Conflicts with mempool and confirmed tx, filters out mempool.
 - Conflicts with two mempool tx at same `seen_at`.

 Some helper functions in the common module.
### Notes to the reviewers

As noted by @evanlinjin on the test description.
> What happens if the conflicts have the same last_seen value?

A possible solution for this situation is mentioned in the comment of the last test, which covers this case. I will attempt a PR to enforce the behavior if that is a reasonable approach. 

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #1064, opened 2023-08-07
Better tests for transaction conflict handling

### LINKED ISSUE #1063: Better tests for transaction conflict handling.
**Describe the enhancement**  

This continues on the work suggested in ticket #985 and implemented in #964.

However, I realized that it is tedious to construct `TxGraph` histories, making it difficult to test multiple scenarios without an overwhelming amount of code. I suggested a `TxTemplate` structure in https://github.com/bitcoindevkit/bdk/pull/964#issuecomment-1551054290:

```rust
struct TxTemplate<'a, K> {
    pub tx_name: &'a str,
    pub spends: &'a [(&'a str, usize)], // (tx_name, vout)
    pub outputs: &'a [TxOutTemplate<K>],
    pub anchors: &'a [(u32, BlockId)],
    pub last_seen: Option<u64>,
}

struct TxOutTemplate<K> {
    pub value: u64,
    pub owned_spk: Option<(K, u32)>, // some = derive a spk from K, none = random spk
}

fn init_graph<'a, K: 'a>(
    descriptors: HashMap<K, Descriptor<DescriptorPublicKey>>,
    tx_templates: impl IntoIterator<Item = TxTemplate<'a, K>>,
) -> (
    IndexedTxGraph<ConfirmationHeightAnchor, KeychainTxOutIndex<K>>,
    HashMap<&'a str, Txid>,
) {
    todo!()
}
```

Alongside this, we need a `Scenario` structure to describe the specific scenario we are testing, and the expected results of various calls to `bdk_chain` structure methods (this is how we should that the final state is expected).

```rust
struct Scenario<'a> {
    /// Name of the test scenario
    name: &'a str,
    /// Transaction templates
    tx_templates: &'a [TxTemplate],
    /// Names of txs that must exist in the output of `list_chain_txs`
    exp_chain_txs: HashSet<&'a str>,
    /// Outpoints of UTXOs that must exist in the output of `filter_chain_unspents`
    exp_unspents: HashSet<(&'a str, u32)>,
    /// Expected balances
    exp_balance: Balance,
}
```

**Scenarios**

- [ ] 2 unconfirmed txs with different `last_seen`s conflict. The most recent tx should be the only tx that appears in the list methods.
- [ ] 3 unconfirmed txs with different `last_seen`s conflict. The most recent tx should be the only TX that appears in the list methods.
- [ ] An unconfirmed tx `U` conflicts with a tx anchored in orphaned block `O`. `O` has higher `last_seen`. `O` should be the only tx that appears in the list methods.
- [ ] An unconfirmed tx `U` conflicts with a tx anchored in orphaned block `O`. `U` has higher `last_seen`. `U` should be the only tx that appears in the list methods.
- [ ] Multiple unconfirmed txs conflict with a confirmed tx. None of the unconfirmed txs should appear in the list methods.
- [ ] `B` and `B'` conflict. `C` spends `B`. `B'` is anchored in best chain. `B` and `C` should not appear in the list methods.
- [ ] `B` and `B'` conflict. `C` spends `B`. `B` is anchored in best chain. `B'` should not appear in the list methods.
- [ ] `B` and `B'` conflict. `C` spends both `B` and `B'`. `C` is impossible. Try this with only `B` confirmed vs only `B'` confirmed vs none confirmed.

**Further work**

Our current structures do not handle unconfirmed conflicts properly when the conflicting transactions have the same `last_seen` timestamp (thanks to @rajarshimaitra for pointing this out).

https://github.com/bitcoindevkit/bdk/blob/fe7bba1d9fcbfd9188fb036f7f2b488e62992a96/crates/chain/tests/test_indexed_tx_graph.rs#L692-L695

I propose we deal with this as follows:

When two unconfirmed txs conflict, but have the same last_seen, we can check the input's sequences (using RBF rules). Otherwise, we sort by feerate (if possible). We fallback to lexicographical sorting of txids.

LagginTimes:
### Description

Fixes bitcoindevkit/bdk#1063.

This PR introduces a new `TxTemplate` struct to test different transaction conflict scenarios in `TxGraph`.
The following transaction conflict scenarios are tested:
- 2 unconfirmed txs with different last_seens conflict. The most recent tx should be the only tx that appears in the list methods.
- 3 unconfirmed txs with different last_seens conflict. The most recent tx should be the only tx that appears in the list methods.
- An unconfirmed tx U conflicts with a tx anchored in orphaned block O. O has higher last_seen. O should be the only tx that appears in the list methods.
- An unconfirmed tx U conflicts with a tx anchored in orphaned block O. U has higher last_seen. U should be the only tx that appears in the list methods.
- Multiple unconfirmed txs conflict with a confirmed tx. None of the unconfirmed txs should appear in the list methods.
- B and B' conflict. C spends B. B' is anchored in best chain. B and C should not appear in the list methods.
- B and B' conflict. C spends B. B is anchored in best chain. B' should not appear in the list methods.
- B and B' conflict. C spends both B and B'. C is impossible.
- B and B' conflict. C spends both B and B'. C is impossible. B' is confirmed.
- B and B' conflict. C spends both B and B'. C is impossible. D spends C.

These tests revealed that `TxGraph::walk_conflicts` was not checking ancestors of the root tx for conflicts. `TxGraph::walk_conflicts` has been refactored to check for conflicting ancestor transactions by using a new `TxAncestors` iterator in `TxGraph`.

### Changelog notice

- Introduced `tx_template` module
- Introduced `TxGraph::TxAncestors` iterator
- Refactored `TxGraph::walk_conflicts` to use `TxGraph::TxAncestors`
- Added `walk_ancestors` to `TxGraph`

### Checklists
All Submissions:

- [x] I've signed all my commits
- [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
- [x] I ran cargo fmt and cargo clippy before committing

#### New Features:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -381,8 +381,9 @@ impl<A> TxGraph<A> {
     /// transaction's inputs (spends). The conflicting txids are returned with the given
     /// transaction's vin (in which it conflicts).
     ///
-    /// Note that this only returns directly conflicting txids and does not include descendants of
-    /// those txids (which are technically also conflicting).
+    /// Note that this only returns directly conflicting txids and won't include:
+    /// - descendants of conflicting transactions (which are technically also conflicting)
+    /// - transactions conflicting with the given transaction's ancestors
     pub fn direct_conflicts_of_tx<'g>(
         &'g self,
         tx: &'g Transaction,
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -324,10 +324,10 @@ impl<A> TxGraph<A> {
     /// The supplied closure takes in two inputs `(depth, descendant_txid)`:
     ///
     /// * `depth` is the distance between the starting `txid` and the `descendant_txid`. I.e., if the
-    ///     descendant is spending an output of the starting `txid`; the `depth` will be 1.
+    ///     descendant is spending an output of the starting `txid` then `depth` will be 1.
     /// * `descendant_txid` is the descendant's txid which we are considering to walk.
     ///
-    /// The supplied closure returns an `Option<T>`, allowing the caller to map each node it vists
+    /// The supplied closure returns an `Option<T>`, allowing the caller to map each node it visits
     /// and decide whether to visit descendants.
     pub fn walk_descendants<'g, F, O>(&'g self, txid: Txid, walk_map: F) -> TxDescendants<A, F>
     where
@@ -1173,7 +1173,7 @@ impl<'g, A, F> TxDescendants<'g, A, F> {
         descendants
     }
 
-    /// Creates a `TxDescendants` from multiple starting transactions that include the starting
+    /// Creates a `TxDescendants` from multiple starting transactions that includes the starting
     /// `txid`s when iterating.
     pub(crate) fn from_multiple_include_root<I>(
         graph: &'g TxGraph<A>,
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -55,6 +55,7 @@ use crate::{
     ChainOracle, ChainPosition, FullTxOut,
 };
 use alloc::vec::Vec;
+use alloc::collections::vec_deque::VecDeque;
 use bitcoin::{OutPoint, Script, Transaction, TxOut, Txid};
 use core::{
     convert::Infallible,
@@ -317,6 +318,30 @@ impl<A> TxGraph<A> {
         self.spends
             .range(start..=end)
             .map(|(outpoint, spends)| (outpoint.vout, spends))
+    }
+
+    /// Creates an iterator that filters and maps ancestor transactions.
+    ///
+    /// The iterator starts with the ancestors of the supplied `tx` (ancestor transactions of `tx`
+    /// are transactions spent by `tx`). The supplied transaction is excluded from the iterator.
+    ///
+    /// The supplied closure takes in two inputs `(depth, ancestor_tx)`:
+    ///
+    /// * `depth` is the distance between the starting `Transaction` and the `ancestor_tx`. I.e., if
+    ///    the `Transaction` is spending an output of the `ancestor_tx` then `depth` will be 1.
+    /// * `ancestor_tx` is the `Transaction`'s ancestor which we are considering to walk.
+    ///
+    /// The supplied closure returns an `Option<T>`, allowing the caller to map each `Transaction`
+    /// it visits and decide whether to visit ancestors.
+    pub fn walk_ancestors<'g, F, O>(
+        &'g self,
+        tx: &'g Transaction,
+        walk_map: F,
+    ) -> TxAncestors<'g, A, F>
+    where
+        F: FnMut(usize, &'g Transaction) -> Option<O> + 'g,
+    {
+        TxAncestors::new_exclude_root(self, tx, walk_map)
     }
 
     /// Creates an iterator that filters and maps descendants from the starting `txid`.
@@ -1134,6 +1159,126 @@ impl<A: Ord> Append for ChangeSet<A> {
 impl<A> AsRef<TxGraph<A>> for TxGraph<A> {
     fn as_ref(&self) -> &TxGraph<A> {
         self
+    }
+}
+
+/// An iterator that traverses ancestors of a given root transaction.
+///
+/// The iterator excludes partial transactions.
+///
+/// This `struct` is created by the [`walk_ancestors`] method of [`TxGraph`].
+///
+/// [`walk_ancestors`]: TxGraph::walk_ancestors
+pub struct TxAncestors<'g, A, F> {
+    graph: &'g TxGraph<A>,
+    visited: HashSet<Txid>,
+    queue: VecDeque<(usize, &'g Transaction)>,
+    filter_map: F,
+}
+
+impl<'g, A, F> TxAncestors<'g, A, F> {
+    /// Creates a `TxAncestors` that includes the starting `Transaction` when iterating.
+    pub(crate) fn new_include_root(
+        graph: &'g TxGraph<A>,
+        tx: &'g Transaction,
+        filter_map: F,
+    ) -> Self {
+        Self {
+            graph,
+            visited: Default::default(),
+            queue: [(0, tx)].into(),
+            filter_map,
+        }
+    }
+
+    /// Creates a `TxAncestors` that excludes the starting `Transaction` when iterating.
+    pub(crate) fn new_exclude_root(
+        graph: &'g TxGraph<A>,
+        tx: &'g Transaction,
+        filter_map: F,
+    ) -> Self {
+        let mut ancestors = Self {
+            graph,
+            visited: Default::default(),
+            queue: Default::default(),
+            filter_map,
+        };
+        ancestors.populate_queue(1, tx);
+        ancestors
+    }
+
+    /// Creates a `TxAncestors` from multiple starting `Transaction`s that includes the starting
+    /// `Transaction`s when iterating.
+    #[allow(unused)]
+    pub(crate) fn from_multiple_include_root<I>(
+        graph: &'g TxGraph<A>,
+        txs: I,
+        filter_map: F,
+    ) -> Self
+    where
+        I: IntoIterator<Item = &'g Transaction>,
+    {
+        Self {
+            graph,
+            visited: Default::default(),
+            queue: txs.into_iter().map(|tx| (0, tx)).collect(),
+            filter_map,
+        }
+    }
+
+    /// Creates a `TxAncestors` from multiple starting `Transaction`s that excludes the starting
+    /// `Transaction`s when iterating.
+    #[allow(unused)]
+    pub(crate) fn from_multiple_exclude_root<I>(
+        graph: &'g TxGraph<A>,
+        txs: I,
+        filter_map: F,
+    ) -> Self
+    where
+        I: IntoIterator<Item = &'g Transaction>,
+    {
+        let mut ancestors = Self {
+            graph,
+            visited: Default::default(),
+            queue: Default::default(),
+            filter_map,
+        };
+        for tx in txs {
+            ancestors.populate_queue(1, tx);
+        }
+        ancestors
+    }
+
+    fn populate_queue(&mut self, depth: usize, tx: &'g Transaction) {
+        let ancestors = tx
+            .input
+            .iter()
+            .map(|txin| txin.previous_output.txid)
+            .filter(|&prev_txid| self.visited.insert(prev_txid))
+            .filter_map(|prev_txid| self.graph.get_tx(prev_txid))
+            .map(|tx| (depth, tx));
+        self.queue.extend(ancestors);
+    }
+}
+
+impl<'g, A, F, O> Iterator for TxAncestors<'g, A, F>
+where
+    F: FnMut(usize, &'g Transaction) -> Option<O>,
+{
+    type Item = O;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            // we have exhausted all paths when queue is empty
+            let (ancestor_depth, tx) = self.queue.pop_front()?;
+            // ignore paths when user filters them out
+            let item = match (self.filter_map)(ancestor_depth, tx) {
+                Some(item) => item,
+                None => continue,
+            };
+            self.populate_queue(ancestor_depth + 1, tx);
+            return Some(item);
+        }
     }
 }
 
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -1145,7 +1145,7 @@ impl<A> AsRef<TxGraph<A>> for TxGraph<A> {
 pub struct TxDescendants<'g, A, F> {
     graph: &'g TxGraph<A>,
     visited: HashSet<Txid>,
-    stack: Vec<(usize, Txid)>,
+    queue: VecDeque<(usize, Txid)>,
     filter_map: F,
 }
 
@@ -1156,7 +1156,7 @@ impl<'g, A, F> TxDescendants<'g, A, F> {
         Self {
             graph,
             visited: Default::default(),
-            stack: [(0, txid)].into(),
+            queue: [(0, txid)].into(),
             filter_map,
         }
     }
@@ -1166,10 +1166,10 @@ impl<'g, A, F> TxDescendants<'g, A, F> {
         let mut descendants = Self {
             graph,
             visited: Default::default(),
-            stack: Default::default(),
+            queue: Default::default(),
             filter_map,
         };
-        descendants.populate_stack(1, txid);
+        descendants.populate_queue(1, txid);
         descendants
     }
 
@@ -1186,7 +1186,7 @@ impl<'g, A, F> TxDescendants<'g, A, F> {
         Self {
             graph,
             visited: Default::default(),
-            stack: txids.into_iter().map(|txid| (0, txid)).collect(),
+            queue: txids.into_iter().map(|txid| (0, txid)).collect(),
             filter_map,
         }
     }
@@ -1205,25 +1205,25 @@ impl<'g, A, F> TxDescendants<'g, A, F> {
         let mut descendants = Self {
             graph,
             visited: Default::default(),
-            stack: Default::default(),
+            queue: Default::default(),
             filter_map,
         };
         for txid in txids {
-            descendants.populate_stack(1, txid);
+            descendants.populate_queue(1, txid);
         }
         descendants
     }
 }
 
 impl<'g, A, F> TxDescendants<'g, A, F> {
-    fn populate_stack(&mut self, depth: usize, txid: Txid) {
+    fn populate_queue(&mut self, depth: usize, txid: Txid) {
         let spend_paths = self
             .graph
             .spends
             .range(tx_outpoint_range(txid))
             .flat_map(|(_, spends)| spends)
             .map(|&txid| (depth, txid));
-        self.stack.extend(spend_paths);
+        self.queue.extend(spend_paths);
     }
 }
 
@@ -1235,8 +1235,8 @@ where
 
     fn next(&mut self) -> Option<Self::Item> {
         let (op_spends, txid, item) = loop {
-            // we have exhausted all paths when stack is empty
-            let (op_spends, txid) = self.stack.pop()?;
+            // we have exhausted all paths when queue is empty
+            let (op_spends, txid) = self.queue.pop_front()?;
             // we do not want to visit the same transaction twice
             if self.visited.insert(txid) {
                 // ignore paths when user filters them out
@@ -1246,7 +1246,7 @@ where
             }
         };
 
-        self.populate_stack(op_spends + 1, txid);
+        self.populate_queue(op_spends + 1, txid);
         Some(item)
     }
 }
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -54,8 +54,8 @@ use crate::{
     collections::*, keychain::Balance, local_chain::LocalChain, Anchor, Append, BlockId,
     ChainOracle, ChainPosition, FullTxOut,
 };
-use alloc::vec::Vec;
 use alloc::collections::vec_deque::VecDeque;
+use alloc::vec::Vec;
 use bitcoin::{OutPoint, Script, Transaction, TxOut, Txid};
 use core::{
     convert::Infallible,
@@ -697,8 +697,9 @@ impl<A: Anchor> TxGraph<A> {
             }
         }
 
-        // The tx is not anchored to a block which is in the best chain, let's check whether we can
-        // ignore it by checking conflicts!
+        // The tx is not anchored to a block which is in the best chain, which means that it
+        // might be in mempool, or it might have been dropped already.
+        // Let's check conflicts to find out!
         let tx = match tx_node {
             TxNodeInternal::Whole(tx) => tx,
             TxNodeInternal::Partial(_) => {
@@ -707,17 +708,70 @@ impl<A: Anchor> TxGraph<A> {
             }
         };
 
-        // If a conflicting tx is in the best chain, or has `last_seen` higher than this tx, then
-        // this tx cannot exist in the best chain
-        for conflicting_tx in self.walk_conflicts(tx, |_, txid| self.get_tx_node(txid)) {
-            for block in conflicting_tx.anchors.iter().map(A::anchor_block) {
-                if chain.is_block_in_chain(block, chain_tip)? == Some(true) {
-                    // conflicting tx is in best chain, so the current tx cannot be in best chain!
+        // We want to retrieve all the transactions that conflict with us, plus all the
+        // transactions that conflict with our unconfirmed ancestors, since they conflict with us
+        // as well.
+        // We only traverse unconfirmed ancestors since conflicts of confirmed transactions
+        // cannot be in the best chain.
+
+        // First of all, we retrieve all our ancestors. Since we're using `new_include_root`, the
+        // resulting array will also include `tx`
+        let unconfirmed_ancestor_txs =
+            TxAncestors::new_include_root(self, tx, |_, ancestor_tx: &Transaction| {
+                let tx_node = self.get_tx_node(ancestor_tx.txid())?;
+                // We're filtering the ancestors to keep only the unconfirmed ones (= no anchors in
+                // the best chain)
+                for block in tx_node.anchors {
+                    match chain.is_block_in_chain(block.anchor_block(), chain_tip) {
+                        Ok(Some(true)) => return None,
+                        Err(e) => return Some(Err(e)),
+                        _ => continue,
+                    }
+                }
+                Some(Ok(tx_node))
+            })
+            .collect::<Result<Vec<_>, C::Error>>()?;
+
+        // We determine our tx's last seen, which is the max between our last seen,
+        // and our unconf descendants' last seen.
+        let unconfirmed_descendants_txs =
+            TxDescendants::new_include_root(self, tx.txid(), |_, descendant_txid: Txid| {
+                let tx_node = self.get_tx_node(descendant_txid)?;
+                // We're filtering the ancestors to keep only the unconfirmed ones (= no anchors in
+                // the best chain)
+                for block in tx_node.anchors {
+                    match chain.is_block_in_chain(block.anchor_block(), chain_tip) {
+                        Ok(Some(true)) => return None,
+                        Err(e) => return Some(Err(e)),
+                        _ => continue,
+                    }
+                }
+                Some(Ok(tx_node))
+            })
+            .collect::<Result<Vec<_>, C::Error>>()?;
+
+        let tx_last_seen = unconfirmed_descendants_txs
+            .iter()
+            .max_by_key(|tx| tx.last_seen_unconfirmed)
+            .map(|tx| tx.last_seen_unconfirmed)
+            .expect("descendants always includes at least one transaction (the root tx");
+
+        // Now we traverse our ancestors and consider all their conflicts
+        for tx_node in unconfirmed_ancestor_txs {
+            // We retrieve all the transactions conflicting with this specific ancestor
+            let conflicting_txs = self.walk_conflicts(tx_node.tx, |_, txid| self.get_tx_node(txid));
+
+            // If a conflicting tx is in the best chain, or has `last_seen` higher than this ancestor, then
+            // this tx cannot exist in the best chain
+            for conflicting_tx in conflicting_txs {
+                for block in conflicting_tx.anchors {
+                    if chain.is_block_in_chain(block.anchor_block(), chain_tip)? == Some(true) {
+                        return Ok(None);
+                    }
+                }
+                if conflicting_tx.last_seen_unconfirmed > tx_last_seen {
                     return Ok(None);
                 }
-            }
-            if conflicting_tx.last_seen_unconfirmed > *last_seen {
-                return Ok(None);
             }
         }
 
--- a/crates/chain/tests/common/tx_template.rs
+++ b/crates/chain/tests/common/tx_template.rs
@@ -1,0 +1,136 @@
+use rand::distributions::{Alphanumeric, DistString};
+use std::collections::HashMap;
+
+use bdk_chain::{tx_graph::TxGraph, BlockId, SpkTxOutIndex};
+use bitcoin::{
+    locktime::absolute::LockTime, secp256k1::Secp256k1, OutPoint, ScriptBuf, Sequence, Transaction,
+    TxIn, TxOut, Txid, Witness,
+};
+use miniscript::Descriptor;
+
+/// Template for creating a transaction in `TxGraph`.
+///
+/// The incentive for transaction templates is to create a transaction history in a simple manner to
+/// avoid having to explicitly hash previous transactions to form previous outpoints of later
+/// transactions.
+#[derive(Clone, Copy, Default)]
+pub struct TxTemplate<'a, A> {
+    /// Uniquely identifies the transaction, before it can have a txid.
+    pub tx_name: &'a str,
+    pub inputs: &'a [TxInTemplate<'a>],
+    pub outputs: &'a [TxOutTemplate],
+    pub anchors: &'a [A],
+    pub last_seen: Option<u64>,
+}
+
+#[allow(dead_code)]
+pub enum TxInTemplate<'a> {
+    /// This will give a random txid and vout.
+    Bogus,
+
+    /// This is used for coinbase transactions because they do not have previous outputs.
+    Coinbase,
+
+    /// Contains the `tx_name` and `vout` that we are spending. The rule is that we must only spend
+    /// from tx of a previous `TxTemplate`.
+    PrevTx(&'a str, usize),
+}
+
+pub struct TxOutTemplate {
+    pub value: u64,
+    pub spk_index: Option<u32>, // some = get spk from SpkTxOutIndex, none = random spk
+}
+
+#[allow(unused)]
+impl TxOutTemplate {
+    pub fn new(value: u64, spk_index: Option<u32>) -> Self {
+        TxOutTemplate { value, spk_index }
+    }
+}
+
+#[allow(dead_code)]
+pub fn init_graph<'a>(
+    tx_templates: impl IntoIterator<Item = &'a TxTemplate<'a, BlockId>>,
+) -> (TxGraph<BlockId>, SpkTxOutIndex<u32>, HashMap<&'a str, Txid>) {
+    let (descriptor, _) = Descriptor::parse_descriptor(&Secp256k1::signing_only(), "tr(tprv8ZgxMBicQKsPd3krDUsBAmtnRsK3rb8u5yi1zhQgMhF1tR8MW7xfE4rnrbbsrbPR52e7rKapu6ztw1jXveJSCGHEriUGZV7mCe88duLp5pj/86'/1'/0'/0/*)").unwrap();
+    let mut graph = TxGraph::<BlockId>::default();
+    let mut spk_index = SpkTxOutIndex::default();
+    (0..10).for_each(|index| {
+        spk_index.insert_spk(
+            index,
+            descriptor
+                .at_derivation_index(index)
+                .unwrap()
+                .script_pubkey(),
+        );
+    });
+    let mut tx_ids = HashMap::<&'a str, Txid>::new();
+
+    for (bogus_txin_vout, tx_tmp) in tx_templates.into_iter().enumerate() {
+        let tx = Transaction {
+            version: 0,
+            lock_time: LockTime::ZERO,
+            input: tx_tmp
+                .inputs
+                .iter()
+                .map(|input| match input {
+                    TxInTemplate::Bogus => TxIn {
+                        previous_output: OutPoint::new(
+                            bitcoin::hashes::Hash::hash(
+                                Alphanumeric
+                                    .sample_string(&mut rand::thread_rng(), 20)
+                                    .as_bytes(),
+                            ),
+                            bogus_txin_vout as u32,
+                        ),
+                        script_sig: ScriptBuf::new(),
+                        sequence: Sequence::default(),
+                        witness: Witness::new(),
+                    },
+                    TxInTemplate::Coinbase => TxIn {
+                        previous_output: OutPoint::null(),
+                        script_sig: ScriptBuf::new(),
+                        sequence: Sequence::MAX,
+                        witness: Witness::new(),
+                    },
+                    TxInTemplate::PrevTx(prev_name, prev_vout) => {
+                        let prev_txid = tx_ids.get(prev_name).expect(
+                            "txin template must spend from tx of template that comes before",
+                        );
+                        TxIn {
+                            previous_output: OutPoint::new(*prev_txid, *prev_vout as _),
+                            script_sig: ScriptBuf::new(),
+                            sequence: Sequence::default(),
+                            witness: Witness::new(),
+                        }
+                    }
+                })
+                .collect(),
+            output: tx_tmp
+                .outputs
+                .iter()
+                .map(|output| match &output.spk_index {
+                    None => TxOut {
+                        value: output.value,
+                        script_pubkey: ScriptBuf::new(),
+                    },
+                    Some(index) => TxOut {
+                        value: output.value,
+                        script_pubkey: spk_index.spk_at_index(index).unwrap().to_owned(),
+                    },
+                })
+                .collect(),
+        };
+
+        tx_ids.insert(tx_tmp.tx_name, tx.txid());
+        spk_index.scan(&tx);
+        let _ = graph.insert_tx(tx.clone());
+        for anchor in tx_tmp.anchors.iter() {
+            let _ = graph.insert_anchor(tx.txid(), *anchor);
+        }
+        if let Some(seen_at) = tx_tmp.last_seen {
+            let _ = graph.insert_seen_at(tx.txid(), seen_at);
+        }
+    }
+    (graph, spk_index, tx_ids)
+}
--- a/crates/chain/tests/test_tx_graph.rs
+++ b/crates/chain/tests/test_tx_graph.rs
@@ -610,7 +610,7 @@ fn test_descendants_no_repeat() {
         .collect::<Vec<_>>();
 
     let mut graph = TxGraph::<()>::default();
-    let mut expected_txids = BTreeSet::new();
+    let mut expected_txids = Vec::new();
 
     // these are NOT descendants of `tx_a`
     for tx in txs_not_connected {
@@ -625,19 +625,14 @@ fn test_descendants_no_repeat() {
         .chain(core::iter::once(&tx_e))
     {
         let _ = graph.insert_tx(tx.clone());
-        assert!(expected_txids.insert(tx.txid()));
+        expected_txids.push(tx.txid());
     }
 
     let descendants = graph
         .walk_descendants(tx_a.txid(), |_, txid| Some(txid))
         .collect::<Vec<_>>();
 
-    assert_eq!(descendants.len(), expected_txids.len());
-
-    for txid in descendants {
-        assert!(expected_txids.remove(&txid));
-    }
-    assert!(expected_txids.is_empty());
+    assert_eq!(descendants, expected_txids);
 }
 
 #[test]
--- a/crates/chain/tests/test_tx_graph.rs
+++ b/crates/chain/tests/test_tx_graph.rs
@@ -5,7 +5,7 @@ use bdk_chain::{
     collections::*,
     local_chain::LocalChain,
     tx_graph::{ChangeSet, TxGraph},
-    Anchor, Append, BlockId, ChainPosition, ConfirmationHeightAnchor,
+    Anchor, Append, BlockId, ChainOracle, ChainPosition, ConfirmationHeightAnchor,
 };
 use bitcoin::{
     absolute, hashes::Hash, BlockHash, OutPoint, ScriptBuf, Transaction, TxIn, TxOut, Txid,
@@ -494,6 +494,208 @@ fn test_calculate_fee_on_coinbase() {
     let graph = TxGraph::<()>::default();
 
     assert_eq!(graph.calculate_fee(&tx), Ok(0));
+}
+
+// `test_walk_ancestors` uses the following transaction structure:
+//
+//     a0
+//    /  \
+//   b0   b1   b2
+//  /  \   \  /
+// c0  c1   c2  c3
+//    /      \  /
+//   d0       d1
+//             \
+//              e0
+//
+// where b0 and b1 spend a0, c0 and c1 spend b0, d0 spends c1, etc.
+#[test]
+fn test_walk_ancestors() {
+    let local_chain: LocalChain = (0..=20)
+        .map(|ht| (ht, BlockHash::hash(format!("Block Hash {}", ht).as_bytes())))
+        .collect::<BTreeMap<u32, BlockHash>>()
+        .into();
+    let tip = local_chain.tip().expect("must have tip");
+
+    let tx_a0 = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint::new(h!("op0"), 0),
+            ..TxIn::default()
+        }],
+        output: vec![TxOut::default(), TxOut::default()],
+        ..common::new_tx(0)
+    };
+
+    // tx_b0 spends tx_a0
+    let tx_b0 = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint::new(tx_a0.txid(), 0),
+            ..TxIn::default()
+        }],
+        output: vec![TxOut::default(), TxOut::default()],
+        ..common::new_tx(0)
+    };
+
+    // tx_b1 spends tx_a0
+    let tx_b1 = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint::new(tx_a0.txid(), 1),
+            ..TxIn::default()
+        }],
+        output: vec![TxOut::default()],
+        ..common::new_tx(0)
+    };
+
+    let tx_b2 = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint::new(h!("op1"), 0),
+            ..TxIn::default()
+        }],
+        output: vec![TxOut::default()],
+        ..common::new_tx(0)
+    };
+
+    // tx_c0 spends tx_b0
+    let tx_c0 = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint::new(tx_b0.txid(), 0),
+            ..TxIn::default()
+        }],
+        output: vec![TxOut::default()],
+        ..common::new_tx(0)
+    };
+
+    // tx_c1 spends tx_b0
+    let tx_c1 = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint::new(tx_b0.txid(), 1),
+            ..TxIn::default()
+        }],
+        output: vec![TxOut::default()],
+        ..common::new_tx(0)
+    };
+
+    // tx_c2 spends tx_b1 and tx_b2
+    let tx_c2 = Transaction {
+        input: vec![
+            TxIn {
+                previous_output: OutPoint::new(tx_b1.txid(), 0),
+                ..TxIn::default()
+            },
+            TxIn {
+                previous_output: OutPoint::new(tx_b2.txid(), 0),
+                ..TxIn::default()
+            },
+        ],
+        output: vec![TxOut::default()],
+        ..common::new_tx(0)
+    };
+
+    let tx_c3 = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint::new(h!("op2"), 0),
+            ..TxIn::default()
+        }],
+        output: vec![TxOut::default()],
+        ..common::new_tx(0)
+    };
+
+    // tx_d0 spends tx_c1
+    let tx_d0 = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint::new(tx_c1.txid(), 0),
+            ..TxIn::default()
+        }],
+        output: vec![TxOut::default()],
+        ..common::new_tx(0)
+    };
+
+    // tx_d1 spends tx_c2 and tx_c3
+    let tx_d1 = Transaction {
+        input: vec![
+            TxIn {
+                previous_output: OutPoint::new(tx_c2.txid(), 0),
+                ..TxIn::default()
+            },
+            TxIn {
+                previous_output: OutPoint::new(tx_c3.txid(), 0),
+                ..TxIn::default()
+            },
+        ],
+        output: vec![TxOut::default()],
+        ..common::new_tx(0)
+    };
+
+    // tx_e0 spends tx_d1
+    let tx_e0 = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint::new(tx_d1.txid(), 0),
+            ..TxIn::default()
+        }],
+        output: vec![TxOut::default()],
+        ..common::new_tx(0)
+    };
+
+    let mut graph = TxGraph::<BlockId>::new(vec![
+        tx_a0.clone(),
+        tx_b0.clone(),
+        tx_b1.clone(),
+        tx_b2.clone(),
+        tx_c0.clone(),
+        tx_c1.clone(),
+        tx_c2.clone(),
+        tx_c3.clone(),
+        tx_d0.clone(),
+        tx_d1.clone(),
+        tx_e0.clone(),
+    ]);
+
+    [&tx_a0, &tx_b1].iter().for_each(|&tx| {
+        let _ = graph.insert_anchor(tx.txid(), tip.block_id());
+    });
+
+    let ancestors = [
+        graph
+            .walk_ancestors(&tx_c0, |depth, tx| Some((depth, tx)))
+            .collect::<Vec<_>>(),
+        graph
+            .walk_ancestors(&tx_d0, |depth, tx| Some((depth, tx)))
+            .collect::<Vec<_>>(),
+        graph
+            .walk_ancestors(&tx_e0, |depth, tx| Some((depth, tx)))
+            .collect::<Vec<_>>(),
+        // Only traverse unconfirmed ancestors of tx_e0 this time
+        graph
+            .walk_ancestors(&tx_e0, |depth, tx| {
+                let tx_node = graph.get_tx_node(tx.txid())?;
+                for block in tx_node.anchors {
+                    match local_chain.is_block_in_chain(block.anchor_block(), tip.block_id()) {
+                        Ok(Some(true)) => return None,
+                        _ => continue,
+                    }
+                }
+                Some((depth, tx_node.tx))
+            })
+            .collect::<Vec<_>>(),
+    ];
+
+    let expected_ancestors = [
+        vec![(1, &tx_b0), (2, &tx_a0)],
+        vec![(1, &tx_c1), (2, &tx_b0), (3, &tx_a0)],
+        vec![
+            (1, &tx_d1),
+            (2, &tx_c2),
+            (2, &tx_c3),
+            (3, &tx_b1),
+            (3, &tx_b2),
+            (4, &tx_a0),
+        ],
+        vec![(1, &tx_d1), (2, &tx_c2), (2, &tx_c3), (3, &tx_b2)],
+    ];
+
+    for (txids, expected_txids) in ancestors.iter().zip(expected_ancestors.iter()) {
+        assert_eq!(txids, expected_txids);
+    }
 }
 
 #[test]
--- a/crates/chain/tests/test_tx_graph_conflicts.rs
+++ b/crates/chain/tests/test_tx_graph_conflicts.rs
@@ -1,0 +1,629 @@
+#[macro_use]
+mod common;
+
+use std::collections::{BTreeSet, HashSet};
+
+use bdk_chain::{keychain::Balance, BlockId};
+use bitcoin::{OutPoint, Script};
+use common::*;
+
+#[allow(dead_code)]
+struct Scenario<'a> {
+    /// Name of the test scenario
+    name: &'a str,
+    /// Transaction templates
+    tx_templates: &'a [TxTemplate<'a, BlockId>],
+    /// Names of txs that must exist in the output of `list_chain_txs`
+    exp_chain_txs: HashSet<&'a str>,
+    /// Outpoints that must exist in the output of `filter_chain_txouts`
+    exp_chain_txouts: HashSet<(&'a str, u32)>,
+    /// Outpoints of UTXOs that must exist in the output of `filter_chain_unspents`
+    exp_unspents: HashSet<(&'a str, u32)>,
+    /// Expected balances
+    exp_balance: Balance,
+}
+
+/// This test ensures that [`TxGraph`] will reliably filter out irrelevant transactions when
+/// presented with multiple conflicting transaction scenarios using the [`TxTemplate`] structure.
+/// This test also checks that [`TxGraph::list_chain_txs`], [`TxGraph::filter_chain_txouts`],
+/// [`TxGraph::filter_chain_unspents`], and [`TxGraph::balance`] return correct data.
+#[test]
+fn test_tx_conflict_handling() {
+    // Create Local chains
+    let local_chain = local_chain!(
+        (0, h!("A")),
+        (1, h!("B")),
+        (2, h!("C")),
+        (3, h!("D")),
+        (4, h!("E")),
+        (5, h!("F")),
+        (6, h!("G"))
+    );
+    let chain_tip = local_chain
+        .tip()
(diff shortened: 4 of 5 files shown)


crates/chain/tests/common/tx_template.rs
```diff
@@ -0,0 +1,131 @@
...
+};
+use miniscript::{Descriptor, DescriptorPublicKey};
+
+use super::TestKeychain;
+
+/// Transaction template.
+pub struct TxTemplate<'a, K, A> {
+    /// Uniquely identifies the transaction, before it can have a txid.
+    pub tx_name: &'a str,
+    pub spends: &'a [TxInTemplate<'a>],
```
evanlinjin: Lets rename this to `inputs`.

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -1,19 +1,597 @@
...
+    /// Names of txs that must exist in the output of `list_chain_txs`
+    exp_chain_txs: HashSet<&'a str>,
+    /// Outpoints of UTXOs that must exist in the output of `filter_chain_unspents`
+    exp_unspents: HashSet<(&'a str, u32)>,
+    /// Expected balances
+    exp_balance: Balance,
+}
+
+#[test]
+fn test_scenarios() {
```
evanlinjin: I think Raj had comments explaining what was being tested. We still need them. Also the test name should be changed back to what it was (or have something better).

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -1,19 +1,597 @@
...
+
+    for scenario in scenarios {
+        let (temp_graph, temp_tx_ids) = init_graph(
+            graph.clone(),
+            graph.index.keychains().clone().into_iter().collect(),
+            scenario.tx_templates.iter(),
+        );
+        graph = temp_graph;
+        tx_ids.extend(temp_tx_ids);
+    }
```
evanlinjin: Looks like we aren't using `Scenario` for the purpose that they are designed for.

If there is no way to use it for these tests here, we can just not use it.

evanlinjin (commented):
Thanks for working on this, glad to see the progress. Just some comments!

crates/chain/tests/test_indexed_tx_graph.rs
LagginTimes: This was my next step, and is what I am currently working on.

crates/chain/tests/common/tx_template.rs
LagginTimes: Done.

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -1,19 +1,331 @@
...
+        exp_balance: Balance {
+            immature: 0,              // coinbase matured
+            trusted_pending: 15000,   // tx5
+            untrusted_pending: 20000, // tx4
+            confirmed: 80000,         // tx1 + tx3
+        },
+    }];
+
+    // A helper lambda to extract and filter data from the graph.
+    let fetch =
```
evanlinjin: We don't need a separate closure here since we are only calling it in one place, but I guess this is a placeholder test so it'll be removed anyway?

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -1,19 +1,379 @@
...
+            // TODO Assert indexes
+            trusted_spks.push(script.to_owned());
+        }
+    }
+    {
+        for _ in 0..10 {
+            let ((_, script), _) = graph.index.reveal_next_spk(&TestKeychain::Internal);
+            untrusted_spks.push(script.to_owned());
+        }
+    }
```
evanlinjin: Try make use of `Iterator` and `Iterator::collect` here.

crates/chain/tests/common/tx_template.rs
```diff
@@ -0,0 +1,139 @@
...
+
+pub struct TxOutTemplate<K> {
+    pub value: u64,
+    pub owned_spk: Option<(K, u32)>, // some = derive a spk from K, none = random spk
+}
+
+#[allow(dead_code)]
+pub fn init_graph<'a>(
+    graph: IndexedTxGraph<ConfirmationHeightAnchor, KeychainTxOutIndex<TestKeychain>>,
+    descriptors: HashMap<TestKeychain, Descriptor<DescriptorPublicKey>>,
```
evanlinjin: I just realized we probably don't need to have multiple descriptors here. Since we are testing conflicts. At most, we just need to know whether the spk is tracked or not tracked (so a single descriptor is fine).

`KeychainTxOutIndex<()>` will do I think.

crates/chain/tests/common/tx_template.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+pub struct TxOutTemplate<K> {
+    pub value: u64,
+    pub owned_spk: Option<(K, u32)>, // some = derive a spk from K, none = random spk
+}
+
+#[allow(dead_code)]
+pub fn init_graph<'a>(
+    graph: IndexedTxGraph<ConfirmationHeightAnchor, KeychainTxOutIndex<()>>,
```
evanlinjin: Nit: I think it will be clearer if we just had a `descriptor` input, and not construct an `IndexedTxGraph` from an `IndexedTxGraph`.

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -1,19 +1,442 @@
...
+                tx1,
+                tx_bestchain_conflict_1, // B
+                tx_bestchain_conflict_2, // B'
+                tx_bestchain_spend_1,    // C
+            ],
+            // B and C should not appear in the list methods, but C is appearing
+            exp_chain_txs: HashSet::from([
+                ("tx1", 0),
+                ("tx_bestchain_conflict_2", 0),
+                ("tx_bestchain_spend_1", 0),
```
evanlinjin: Let's keep the tests accurate (even though there may be a bug in `TxGraph`).

crates/chain/tests/common/tx_template.rs
LLFourn: Why do we need descriptors at all? You can just insert script pubkeys into `SpkTxOutIndex`. I am a little unsure as to why we need an indexed graph either. Doesn't transaction conflict handling exclusively in the domain of methods on `TxGraph` which are passed a `ChainOracle`.

evanlinjin: ~@LLFourn Good point.~

Edit:

Wait, we need to descriptor, because of how `TxOutTemplate` works.

https://github.com/bitcoindevkit/bdk/blob/a86f76ae0cfd3dd333699c8c0bf70a72723cc3d0/crates/chain/tests/common/tx_template.rs#L55-L58

But we should shove the spks into `SpkTxoutIndex`. @LagginTimes also, we don't need to `K` generic anymore.

crates/chain/tests/common/tx_template.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    /// This is used for coinbase transactions because they do not have previous outputs.
+    Coinbase,
+
+    /// Contains the `tx_name` and `vout` that we are spending. The rule is that we must only spend
+    /// a previous transaction.
+    PrevTx(&'a str, usize),
+}
+
+pub struct TxOutTemplate<K> {
```
evanlinjin: Please get rid of this generic!

evanlinjin:
Please rebase on `master` to fix the CI.

crates/chain/tests/common/tx_template.rs
```diff
@@ -0,0 +1,145 @@
...
+                let _ = graph.insert_tx(tx.clone());
+            }
+            for anchor in tx_tmp.anchors.iter() {
+                let _ = graph.insert_anchor(tx.txid(), *anchor);
+            }
+            if let Some(seen_at) = tx_tmp.last_seen {
+                let _ = graph.insert_seen_at(tx.txid(), seen_at);
+            }
+        };
+    }
```
evanlinjin: Let's not filter out transactions. If the transaction is in the scenario, shove it in the graph.

crates/chain/tests/common/tx_template.rs
LagginTimes: `<K>`.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -14,7 +14,7 @@ use crate::{
 /// A struct that combines [`TxGraph`] and an [`Indexer`] implementation.
 ///
 /// This structure ensures that [`TxGraph`] and [`Indexer`] are updated atomically.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
evanlinjin: Why is this change necessary? Ideally we want to avoid cloning `IndexedTxGraph` at it may become problematic.

LagginTimes: This was probably left over from old code/debugging use and has been removed.

evanlinjin:
It's looking good so far. We've discussed getting rid of the recursion. After that is done, let's clean up the commit messages:

```
test(chain): add `TxTemplate` tests...  # add a proper body & comment out failing tests
fix(chain): fix ...                     # enable the now-passing tests
```

crates/chain/src/tx_graph.rs
```diff
@@ -655,25 +657,47 @@ impl<A: Anchor> TxGraph<A> {
...
         };
 
-        // If a conflicting tx is in the best chain, or has `last_seen` higher than this tx, then
-        // this tx cannot exist in the best chain
-        for conflicting_tx in self.walk_conflicts(tx, |_, txid| self.get_tx_node(txid)) {
-            for block in conflicting_tx.anchors.iter().map(A::anchor_block) {
-                if chain.is_block_in_chain(block, chain_tip)? == Some(true) {
-                    // conflicting tx is in best chain, so the current tx cannot be in best chain!
+        while !check_tx_conflicts.is_empty() {
+            let (tx, last_seen, ancestor_height) = check_tx_conflicts.pop().unwrap();
```
evanlinjin: **Nit:** I think this is better as we can avoid the `.unwrap()`.

```suggestion
        loop {
            let (tx, last_seen, ancestor_height) = match check_tx_conflicts.pop() {
                Some(item) => item,
                None => break,
            };
```

crates/chain/src/tx_graph.rs
```diff
@@ -655,25 +657,47 @@ impl<A: Anchor> TxGraph<A> {
...
+                    .iter()
+                    .map(|txin| txin.previous_output.txid)
+                    .for_each(|prev_txid| {
+                        if let Some((tx_node, _, last_seen)) = self.txs.get(&prev_txid) {
+                            match tx_node {
+                                TxNodeInternal::Whole(tx) => {
+                                    check_tx_conflicts.push((tx, last_seen, ancestor_height + 1));
+                                }
+                                TxNodeInternal::Partial(_) => {}
+                            };
```
evanlinjin: **Nit:**

```suggestion
                            if let TxNodeInternal::Whole(tx) = tx_node {
                                check_tx_conflicts.push((tx, last_seen, ancestor_height + 1));
                            }
```

crates/chain/src/tx_graph.rs
```diff
@@ -641,6 +641,8 @@ impl<A: Anchor> TxGraph<A> {
         chain_tip: BlockId,
         txid: Txid,
     ) -> Result<Option<ChainPosition<&A>>, C::Error> {
+        let mut check_tx_conflicts: Vec<(&Transaction, &u64, u32)> = Vec::new();
```
evanlinjin: Add a comment as to why this is needed. Something like this:

```
// This is a stack of previous spends of `txid`. Anything that conflicts with `txid`'s ancestors also conflict with `txid`.
```

crates/chain/src/tx_graph.rs
```diff
@@ -655,25 +657,47 @@ impl<A: Anchor> TxGraph<A> {
...
+                    }
+                }
+                if conflicting_tx.last_seen_unconfirmed > *last_seen {
                     return Ok(None);
                 }
             }
-            if conflicting_tx.last_seen_unconfirmed > *last_seen {
-                return Ok(None);
+
+            // Add this tx's direct ancestor(s) to the queue to check for conflicts, with a hard
```
```suggestion
            // Add this tx's direct ancestor(s) to the stack to check for conflicts, with a hard
```

evanlinjin (changes requested):
Thank for the excellent work. A few notes.

1. Can you please add a proper PR description? We need a short description of what we are testing, and the bug that was found and how it was fixed.

2. We've identified a shortcoming of `TxGraph::walk_conflicts`. It does not find all conflicts since it does not check conflicts of ancestors of the root tx. Please include this in the docs of `walk_conflicts`. Also make a separate ticket to fix `walk_conflicts` for good.

crates/chain/tests/common/tx_template.rs
```diff
@@ -0,0 +1,141 @@
...
+    for (bogus_txin_vout, tx_tmp) in tx_templates.into_iter().enumerate() {
+        let tx = Transaction {
+            version: 0,
+            lock_time: LockTime::ZERO,
+            input: tx_tmp
+                .inputs
+                .iter()
+                .map(|input| match input {
+                    TxInTemplate::Bogus => TxIn {
+                        // #TODO have actual random data
```
evanlinjin: Let's complete this!

A simple solution is to use the current unix timestamp as the seed (which you can hash/modulo).

Also add some bogus inputs to test transactions.

evanlinjin (changes requested):
Last thing.

evanlinjin (approved):
ACK 6bd514518d90c221d33e0cbad4e5db5b9476510d

However, we should consider changing the behavior of `TxGraph::walk_conflicts` to walk ancestors to a given depth and check conflicts on those ancestors. This can be a separate issue.

crates/chain/tests/common/tx_template.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+impl<'a, A> Copy for TxTemplate<'a, A> {}
+
+impl<'a, A> Clone for TxTemplate<'a, A> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
```
danielabrozzoni: Clone and Copy can be derived I think :)

crates/chain/tests/common/tx_template.rs
```diff
@@ -0,0 +1,141 @@
...
+                    },
+                    TxInTemplate::Coinbase => TxIn {
+                        previous_output: OutPoint::null(),
+                        script_sig: ScriptBuf::new(),
+                        sequence: Sequence::MAX,
+                        witness: Witness::new(),
+                    },
+                    TxInTemplate::PrevTx(prev_name, prev_vout) => {
+                        let prev_txid = tx_ids.get(prev_name).expect(
+                            "txin template must spend from tx of template that comes before",
```
danielabrozzoni, line 98: nit: It's good that there's an expect here, but maybe it should be written in the method docs as well

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -153,89 +153,89 @@ fn test_tx_conflict_handling() {
         ..Default::default()
     };
 
-    // let tx_bestchain_conflict_1: TxTemplate<'_, ConfirmationHeightAnchor> = TxTemplate {
```
danielabrozzoni: (in 1b2d552ba5333fb011125c26fd1d95190a7a425f and 6e75d051f5fd9c9e61d189893d18234ba01c86fc): sorry for being picky about commit hygiene, but I don't think we should have a commit adding a commented test and another commit uncommenting it. You can either add the test in the first commit without comments, or add it in the second commit along with the fix

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -1,473 +0,0 @@
-#[macro_use]
```
danielabrozzoni: In 9aa22dd62554c0138928b14f5f4332ed040de369: why is this being removed? Can we have the reason in the commit message? Are you moving these tests somewhere else or just completely rewriting it?

crates/chain/tests/common/tx_template.rs
```diff
@@ -1,9 +1,10 @@
+use rand::distributions::{Alphanumeric, DistString};
```
danielabrozzoni, line 1: nit: I would squash 760eaa246dbb1fc4329f7d32b00f86a0d0b9231e into 1b2d552ba5333fb011125c26fd1d95190a7a425f, but if you prefer to keep them separate it's fine as well

crates/chain/tests/common/tx_template.rs
```diff
@@ -0,0 +1,141 @@
...
+    Coinbase,
+
+    /// Contains the `tx_name` and `vout` that we are spending. The rule is that we must only spend
+    /// a previous transaction.
+    PrevTx(&'a str, usize),
+}
+
+pub struct TxOutTemplate {
+    pub value: u64,
+    pub spk_index: Option<u32>, // some = get spk from SpkTxOutIndex, none = random spk
```
danielabrozzoni, line 41: I'm curious, why are you inserting the index instead of the whole spk? I'd say not to worry too much about cloning since it's just tests, clarity should be preferred.

crates/chain/src/tx_graph.rs
```diff
@@ -339,6 +339,9 @@ impl<A> TxGraph<A> {
     /// Creates an iterator that both filters and maps conflicting transactions (this includes
     /// descendants of directly-conflicting transactions, which are also considered conflicts).
     ///
+    /// Note that this does not find all conflicts, because we are not checking ancestors of the
```
danielabrozzoni: nit in the commit message of 6e75d051f5fd9c9e61d189893d18234ba01c86fc: you have a duplicated line "fix(chain): add check for ancestor tx conflicts in `TxGraph`" at the end

crates/chain/src/tx_graph.rs
```diff
@@ -319,6 +319,12 @@ impl<A> TxGraph<A> {
             .map(|(outpoint, spends)| (outpoint.vout, spends))
     }
 
+    /// Creates an iterator of ancestor transactions from the starting `Transaction`, up to a
+    /// specified depth limit.
```
danielabrozzoni: nit: we should document that this doesn't include the tx itself. I know it seems obvious but it might not be!

crates/chain/src/tx_graph.rs
```diff
@@ -357,7 +368,8 @@ impl<A> TxGraph<A> {
     /// transaction's vin (in which it conflicts).
     ///
     /// Note that this only returns directly conflicting txids and does not include descendants of
-    /// those txids (which are technically also conflicting).
+    /// those txids or conflicting txids of the given transaction's ancestors (which are technically
+    /// also conflicting).
```
danielabrozzoni: nit: I would rewrite it as:
```
/// Note that this only returns directly conflicting txids and does not include:
/// - descendants of conflicting transactions (which are technically also conflicting)
/// - transactions conflicting with the given transaction's ancestors
```

crates/chain/src/tx_graph.rs
```diff
@@ -674,7 +686,7 @@ impl<A: Anchor> TxGraph<A> {
         // The tx is not anchored to a block which is in the best chain, let's check whether we can
         // ignore it by checking conflicts!
         let tx = match tx_node {
-            TxNodeInternal::Whole(tx) => tx,
+            TxNodeInternal::Whole(tx) => tx, // check_tx_conflicts.push((tx, last_seen, 0)),
```
danielabrozzoni: Is this supposed to be here?

crates/chain/src/tx_graph.rs
```diff
@@ -1110,6 +1122,118 @@ impl<A> ForEachTxOut for TxGraph<A> {
...
 
+/// An iterator that traverses transaction ancestors to a specified depth.
+///
+/// This `struct` is created by the [`walk_ancestors`] method of [`TxGraph`].
+///
+/// [`walk_ancestors`]: TxGraph::walk_ancestors
+pub struct TxAncestors<'g, A> {
+    graph: &'g TxGraph<A>,
+    visited: HashSet<Txid>,
+    stack: Vec<(usize, Transaction)>,
```
danielabrozzoni: nit: document what this usize is (depth?)

crates/chain/src/tx_graph.rs
```diff
@@ -1110,6 +1122,118 @@ impl<A> ForEachTxOut for TxGraph<A> {
...
+}
+
+impl<'g, A> TxAncestors<'g, A> {
+    fn populate_stack(&mut self, depth: usize, tx: Transaction) {
+        if depth <= self.depth_limit && self.visited.insert(tx.txid()) {
+            tx.input
+                .iter()
+                .map(|txin| txin.previous_output.txid)
+                .for_each(|prev_txid| {
+                    if let Some((TxNodeInternal::Whole(tx), _, _)) = self.graph.txs.get(&prev_txid)
```
danielabrozzoni: Do I understand correctly that we don't traverse partial txs? If that's the case, we should document it

crates/chain/tests/test_tx_graph.rs
```diff
@@ -496,6 +496,209 @@ fn test_calculate_fee_on_coinbase() {
...
+//
+//     a0
+//    /  \
+//   b0   b1   b2
+//  /  \   \  /
+// c0  c1   c2  c3
+//    /      \  /
+//   d0       d1
+//             \
+//              e0
```
danielabrozzoni, line 509: I know Evan praised you already, but I feel like repeating it - thank you for this sketch of the txgraph, it is truly helpful!

crates/chain/tests/test_tx_graph.rs
```diff
@@ -496,6 +496,209 @@ fn test_calculate_fee_on_coinbase() {
...
+        tx_c2.clone(),
+        tx_c3.clone(),
+        tx_d0.clone(),
+        tx_d1.clone(),
+        tx_e0.clone(),
+    ]);
+
+    let mut ancestors = vec![
+        graph
+            .walk_ancestors(tx_e0.clone(), 25)
```
danielabrozzoni: TxAncestors is already an iterator, you don't need to collect and iter again:
```
        graph
            .walk_ancestors(tx_e0.clone(), 25)
            .map(|(depth, tx)| (depth, tx.txid()))
            .collect::<Vec<_>>(),
```

crates/chain/tests/test_tx_graph.rs
```diff
@@ -496,6 +496,209 @@ fn test_calculate_fee_on_coinbase() {
...
+            (2, tx_c2.txid()),
+            (2, tx_c3.txid()),
+            (1, tx_d1.txid()),
+        ],
+        vec![(3, tx_a0.txid()), (2, tx_b0.txid()), (1, tx_c1.txid())],
+        vec![(1, tx_b0.txid())],
+    ];
+
+    while let Some(mut txids) = ancestors.pop() {
+        let mut expected_txids = expected_ancestors.pop().unwrap();
```
danielabrozzoni: Commenting just in case you ever need it, you can just zip the two vectors in this case. But popping works just fine!

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,1282 @@
...
+            exp_unspents: HashSet::from([("tx_conflict_3", 0)]),
+            exp_balance: Balance {
+                immature: 0,
+                trusted_pending: 40000,
+                untrusted_pending: 0,
+                confirmed: 0,
+            },
+        },
+        Scenario {
+            name: "unconfirmed tx conflicts with tx in orphaned block, orphaned higher unseen",
```
danielabrozzoni: Here and below, I think you meant `last_seen` instead of unseen?

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,1282 @@
...
+        tx_name: "tx_orphaned_conflict_1",
+        inputs: &[TxInTemplate::PrevTx("tx1", 0)],
+        outputs: &[TxOutTemplate {
+            value: 30000,
+            spk_index: Some(2),
+        }],
+        anchors: &[ConfirmationHeightAnchor {
+            anchor_block: BlockId {
+                height: 5,
+                hash: h!("Orphaned Block"),
```
danielabrozzoni: Btw, I think the orphaned block should have height = 4, as we don't have a block with height 5 in the best chain. But I don't think it really matters for the test scenario

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,1282 @@
...
+            exp_unspents: HashSet::from([("tx_confirmed_conflict", 0)]),
+            exp_balance: Balance {
+                immature: 0,
+                trusted_pending: 0,
+                untrusted_pending: 0,
+                confirmed: 50000,
+            },
+        },
+        Scenario {
+            name: "B and B' conflict, C spends B, B' is anchored in best chain",
```
danielabrozzoni: You should specify that B and B' both spend from A, which is in the best chain

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,1282 @@
...
+    /// Expected balances
+    exp_balance: Balance,
+}
+
+/// This test ensures that [`TxGraph`] will reliably filter out irrelevant transactions when
+/// presented with multiple conflicting transaction scenarios using the [`TxTemplate`] structure.
+/// This test also checks that [`TxGraph::filter_chain_txouts`], [`TxGraph::filter_chain_unspents`],
+/// and [`TxGraph::balance`] return correct data.
+#[test]
+fn test_tx_conflict_handling() {
```
danielabrozzoni, line 31: nit: this test is quite difficult to follow as there are a lot of test scenarios, and I found myself going back and forth between the tx declaration and the test cases; it would be helpful to split it into smaller tests.

danielabrozzoni (commented):
Thanks a lot for your work! This is just a partial review, I reviewed everything apart from the tests in `crates/chain/tests/test_tx_graph_conflicts.rs`, I only reviewed `test_tx_conflict_handling` in that file, I'm going to finish the review in the next few days. 

#1063 includes a "Further work" section that I don't think is addressed here, am I correct? If that's the case, we should open a new issue before merging this PR

crates/chain/tests/common/tx_template.rs
evanlinjin: Thanks for the comment! The thinking is that it is meant to be a "template" so it should be easy to read and write out. An index is perfect for this, instead of having the full spk.

crates/chain/src/tx_graph.rs
evanlinjin: "may not include"

crates/chain/src/tx_graph.rs
```suggestion
    stack: Vec<(usize, &'g Transaction)>,
```

The iterator should use references of transactions instead of cloning them. You'll need to play around with lifetimes to make this happen.

crates/chain/src/tx_graph.rs
```diff
@@ -319,6 +319,12 @@ impl<A> TxGraph<A> {
             .map(|(outpoint, spends)| (outpoint.vout, spends))
     }
 
+    /// Creates an iterator of ancestor transactions from the starting `Transaction`, up to a
+    /// specified depth limit.
+    pub fn walk_ancestors(&self, tx: Transaction, depth_limit: usize) -> TxAncestors<A> {
```
```suggestion
    pub fn walk_ancestors<'g>(&'g self, tx: &'g Transaction, depth_limit: usize) -> TxAncestors<'g, A> {
```

also make `TxAncestors` iterate references to `Transaction`s.

crates/chain/src/tx_graph.rs
```diff
@@ -337,18 +343,23 @@ impl<A> TxGraph<A> {
...
         walk_map: F,
     ) -> TxDescendants<A, F>
     where
         F: FnMut(usize, Txid) -> Option<O> + 'g,
     {
-        let txids = self.direct_conflicts_of_tx(tx).map(|(_, txid)| txid);
+        let mut txids = HashSet::new();
+        for (_, tx) in TxAncestors::new_include_root(self, tx.clone(), depth_limit) {
+            txids.extend(self.direct_conflicts_of_tx(&tx).map(|(_, txid)| txid));
+        }
```
evanlinjin: Nit: this can be written in a more "functional" manner.

crates/chain/tests/test_tx_graph.rs
evanlinjin: @LagginTimes I would go ahead and use [`zip`](https://doc.rust-lang.org/core/iter/trait.Iterator.html#method.zip) here.

Another trick would be to collect into a `BTreeSet` so you don't need to `.sort()`.

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,1282 @@
...
+};
+use common::*;
+use miniscript::{Descriptor, DescriptorPublicKey};
+
+#[allow(dead_code)]
+struct Scenario<'a> {
+    /// Name of the test scenario
+    name: &'a str,
+    /// Transaction templates
+    tx_templates: &'a [TxTemplate<'a, ConfirmationHeightAnchor>],
```
evanlinjin: Since #1069 is merged now, we should use `BlockId` as the anchor to simplify these tests.

crates/chain/tests/test_tx_graph_conflicts.rs
evanlinjin: I agree with this. I think we should declare the tx templates in the actual scenarios.

To make it more readable, we can also do the following:
* Use `BlockId` as the anchor (maybe add a `block_id!(1, "B")` macro so each anchor declaration is on one line).
* Add `TxOutTemplate::new(value, spk_index)` method, or a macro. This way each txout declaration is on one line.

evanlinjin (changes requested):

crates/chain/tests/common/tx_template.rs
evanlinjin: But `A` will need to implement Copy and Clone, which most anchors do anyway? I don't mind either way.

crates/chain/tests/test_tx_graph_conflicts.rs
evanlinjin: @danielabrozzoni is there a document/article you refer to for best practice for commit hygiene? Will be helpful for future reference.

evanlinjin:
@danielabrozzoni bitcoindevkit/bdk#1102 is the ticket for the "further work" section.

crates/chain/tests/common/tx_template.rs
danielabrozzoni: That's the case already I think?

crates/chain/tests/test_tx_graph_conflicts.rs
danielabrozzoni: I really like https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#committing-patches

But in general, for big PRs it makes sense to review each commit separately, so you should make sure that commits are not modifying what's done in previous commits in the same PR

crates/chain/tests/test_indexed_tx_graph.rs
LagginTimes: It was decided that the file name should be renamed to `test_tx_graph_conflicts`. The old one, `test_indexed_tx_graph` was therefore removed.

crates/chain/src/tx_graph.rs
```diff
@@ -683,7 +695,7 @@ impl<A: Anchor> TxGraph<A> {
 
         // If a conflicting tx is in the best chain, or has `last_seen` higher than this tx, then
         // this tx cannot exist in the best chain
-        for conflicting_tx in self.walk_conflicts(tx, |_, txid| self.get_tx_node(txid)) {
+        for conflicting_tx in self.walk_conflicts(tx, 25, |_, txid| self.get_tx_node(txid)) {
```
evanlinjin: I think it would make sense to make the `depth_limit` input configurable.

However, because `try_get_chain_position` is called in so many places, would it make sense to make all these methods take in an `ancestor_depth_limit` argument?

The other option is to have the `ancestor_depth_limit` global for the entire `TxGraph`, but would there be a situation where we want different depth limits for the same graph?

Is there even a way to avoid the depth limit entirely?

LagginTimes:
I've addressed all requested changes except for the ones related to commit hygiene, which will be addressed at a later date.

crates/chain/src/tx_graph.rs
LLFourn: I think as long as we accept this `25` is a temporary solution it's fine to leave it fixed. Let's put it in a constant.

crates/chain/src/tx_graph.rs
```diff
@@ -337,27 +347,35 @@ impl<A> TxGraph<A> {
...
     /// Creates an iterator that both filters and maps conflicting transactions (this includes
-    /// descendants of directly-conflicting transactions, which are also considered conflicts).
+    /// ancestors and descendants of directly-conflicting transactions, which are also considered
+    /// conflicts).
     ///
     /// Refer to [`Self::walk_descendants`] for `walk_map` usage.
     pub fn walk_conflicts<'g, F, O>(
         &'g self,
         tx: &'g Transaction,
+        depth_limit: usize,
```
LLFourn: I think that we don't want to add this parameter here. I see why this approach needs this limit but I don't think it belongs in the public API. `walk_conflicts` should let you walk all the transactions that would have to not exist for this transaction to be canonical assuming the parents of this tx are canonical. This makes sense in the context of broadcasting a transaction: you want to be able to find out which txs this will remove when you broadcast it and it is accepted. In this case you already know that all the parents are canonical (otherwise your broadcast will fail anyway).

I think that you can still do this full conflict list (including transitive ancestor conflicts) with internal logic which just calls walk conflicts for each ancestor up to a certain depth.

crates/chain/src/tx_graph.rs
LLFourn: Rather than just going to `25` it would be better to continue getting ancestor conflicts until you reach the first confirmed ancestor because then it no longer matters what its conflicts are. I'm not sure if this be done easily.

I wrote what I think the long term plan to approach the general problem should be in: https://github.com/bitcoindevkit/bdk/issues/1099#issuecomment-1704559684 and remove this limit anyway.

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,1348 @@
...
+    assert_eq!(txouts.len(), 3);
+    assert_eq!(
+        txouts
+            .iter()
+            .filter(|(_, txout)| matches!(txout.chain_position, ChainPosition::Unconfirmed(100)))
+            .count(),
+        2
+    );
+
+    // We have two mempool utxos both at same chain position.
```
LLFourn: why do we assert the same thing twice.

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,1348 @@
...
+    let local_chain = LocalChain::from(
+        (0..150)
+            .map(|i| (i as u32, h!("random")))
+            .collect::<BTreeMap<u32, BlockHash>>(),
+    );
+
+    // Initiate IndexedTxGraph
+
+    let (desc_1, _) = Descriptor::parse_descriptor(&Secp256k1::signing_only(), "tr(tprv8ZgxMBicQKsPd3krDUsBAmtnRsK3rb8u5yi1zhQgMhF1tR8MW7xfE4rnrbbsrbPR52e7rKapu6ztw1jXveJSCGHEriUGZV7mCe88duLp5pj/86'/1'/0'/0/*)").unwrap();
+    let (desc_2, _) = Descriptor::parse_descriptor(&Secp256k1::signing_only(), "tr(tprv8ZgxMBicQKsPd3krDUsBAmtnRsK3rb8u5yi1zhQgMhF1tR8MW7xfE4rnrbbsrbPR52e7rKapu6ztw1jXveJSCGHEriUGZV7mCe88duLp5pj/86'/1'/0'/1/*)").unwrap();
```
LLFourn: This test looks like it's testing aspects of `IndexedTxGraph` so it should be in `test_indexed_tx_graph.rs`.

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,1348 @@
...
+
+/// Ensure [`IndexedTxGraph::insert_relevant_txs`] can successfully index transactions NOT presented
+/// in topological order.
+///
+/// Given 3 transactions (A, B, C), where A has 2 owned outputs. B and C spends an output each of A.
+/// Typically, we would only know whether B and C are relevant if we have indexed A (A's outpoints
+/// are associated with owned spks in the index). Ensure insertion and indexing is topological-
+/// agnostic.
+#[test]
+fn insert_relevant_txs() {
```
LLFourn: This doesn't seem to be about testing conflicts but it's in `test_tx_graph_conflicts.rs`.

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,1348 @@
...
+        assert_eq!(
+            balance, scenario.exp_balance,
+            "\n[{}] 'balance' failed",
+            scenario.name
+        );
+    }
+}
+
+#[allow(unused)]
+pub fn single_descriptor_setup() -> (
```
LLFourn: we shouldn't need an indexed tx graph to test conflicts.

crates/chain/tests/test_indexed_tx_graph.rs
LLFourn: `test_index_tx_graph.rs` should have tests that specifically test things the indexed tx graph does. i.e. after you insert a tx into it, the indexer has been called and asked to index that tx. e.g. `insert_relevant_txs` should be in here since it has nothing to do with conflicts as far as I can tell.

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,1348 @@
...
+use common::*;
+use miniscript::{Descriptor, DescriptorPublicKey};
+
+#[allow(dead_code)]
+struct Scenario<'a> {
+    /// Name of the test scenario
+    name: &'a str,
+    /// Transaction templates
+    tx_templates: &'a [TxTemplate<'a, BlockId>],
+    /// Names of txs that must exist in the output of `list_chain_txs`
```
LLFourn, line 16: the output of `list_chain_txs` is never tested anywhere. Can we test it like this comment implies we should?

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,1348 @@
...
+                untrusted_pending: 20000, // tx4
+                confirmed: 80000          // tx1 + tx3
+            }
+        );
+    }
+}
+
+/// Check conflicts between two in mempool transactions. Tx with older `seen_at` is filtered out.
+#[test]
+fn test_unconfirmed_conflicts() {
```
LLFourn: Is there a reason this test and the others that use `setup_conflicts` cannot be incorporated into the "scenarios" structure above?

LLFourn (changes requested):
Hey @LagginTimes. Thanks for looking at how to make the conflict handling better. I your approach is the simplest way to get the tests passing to fix the intended bug. Great work. In practice the implementation might be inefficient with a deep mempool and lots of conflicting txs because of the repeated ancestor traversal. In https://github.com/bitcoindevkit/bdk/issues/1099#issuecomment-1704559684 I put forward a plan to solve this. Having said that, getting tests passing is more important for now and we can iterate on that later.

I've left comments explaining how this PR can be improved but in summary:

1. Leave the public API the same (except for adding the lovely new ancestors traversal feature)
2. In the tx graph conflict tests, don't use an indexed tx graph.
3. restore `test_indexed_tx_graph.rs` and put things in there that test its API (e.g. `insert_relevant_txs`.

crates/chain/tests/test_tx_graph_conflicts.rs
LagginTimes: This was code that was incorporated from #964 and has not been refactored with new `TxTemplate` yet. My PR description should have mentioned the fact that almost all of the code taken from #964 has thus far been left as-is.

crates/chain/tests/test_tx_graph_conflicts.rs
LagginTimes: This is also mostly unmodified code from #964. In discussion with @evanlinjin, one of the next steps was to refactor the #964 tests with the `TxTemplate` scenario structure.

crates/chain/tests/test_tx_graph_conflicts.rs
evanlinjin: @LagginTimes let's extract what is good from #964 into "scenarios".

crates/chain/src/tx_graph.rs
```diff
@@ -319,14 +319,19 @@ impl<A> TxGraph<A> {
             .map(|(outpoint, spends)| (outpoint.vout, spends))
     }
 
-    /// Creates an iterator of ancestor transactions from the starting `Transaction` up to a
-    /// specified depth limit, not including the original `Transaction`.
-    pub fn walk_ancestors<'g>(
+    /// Creates an iterator of unconfirmed ancestor transactions from the starting `Transaction`,
```
LLFourn: `walk_ancestors` should traverse all ancestors. It sounds like a general method traversing ancestors in a graph. You could have another method for `walk_unconfirmed_ancestors`.

LagginTimes: The next step, after discussion with @evanlinjin, was to refactor `TxAncestors` again to take in a closure for versatility (similar to `TxDescendants`), which will allow `walk_ancestors` to traverse all ancestors. The work for this is nearly complete.

crates/chain/src/tx_graph.rs
```diff
@@ -205,6 +205,26 @@ impl<A> TxGraph<A> {
         self.get_tx_node(txid).map(|n| n.tx)
     }
 
+    /// Get a transaction by txid. This only returns `Some` for full, unconfirmed transactions.
+    pub fn get_unconfirmed_tx<'g, C>(
```
evanlinjin: Try not to introduce new API if it's not needed. This method is only used internally, and as a public API, `get_chain_position` can be used instead.

crates/chain/src/tx_graph.rs
```diff
@@ -205,6 +205,26 @@ impl<A> TxGraph<A> {
...
+        &'g self,
+        txid: Txid,
+        chain: &'g C,
+        chain_tip: BlockId,
+    ) -> Result<Option<&'g Transaction>, C::Error>
+    where
+        A: Anchor,
+        C: ChainOracle,
+    {
+        let tx_node = self.get_tx_node(txid).expect("tx must exist");
```
evanlinjin: Given that the intention of this method is to be a public API, we should not panic if tx does not exist, since the caller provides the `txid`.

crates/chain/src/tx_graph.rs
```diff
@@ -699,18 +725,24 @@ impl<A: Anchor> TxGraph<A> {
...
-        // If a conflicting tx is in the best chain, or has `last_seen` higher than this tx, then
-        // this tx cannot exist in the best chain
-        for conflicting_tx in self.walk_conflicts(tx, 25, |_, txid| self.get_tx_node(txid)) {
-            for block in conflicting_tx.anchors.iter().map(A::anchor_block) {
-                if chain.is_block_in_chain(block, chain_tip)? == Some(true) {
-                    // conflicting tx is in best chain, so the current tx cannot be in best chain!
+        // Populate a list of unconfirmed ancestor txs to check for conflicts
+        for tx in TxAncestors::new_include_root(self, tx, |_, tx: &Transaction| {
+            self.get_unconfirmed_tx(tx.txid(), chain, chain_tip)
+                .expect("error is infallible")
```
evanlinjin: We cannot guarantee that the `ChainOracle` implementation has an [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html) error type.

A solution is to have the closure return `Option<Result<&tx, C::Error>>`, and we collect the ancestor txs into `Result<Vec<_>, C::Error>`. You'll need to play around with the lifetimes.

evanlinjin (changes requested):
Good work so far.

Please remove `TxGraph::get_unconfirmed_tx` and make sure we aren't panicing in unexpected ways.

crates/chain/src/tx_graph.rs
```diff
@@ -681,11 +705,29 @@ impl<A: Anchor> TxGraph<A> {
...
+
         // If a conflicting tx is in the best chain, or has `last_seen` higher than this tx, then
         // this tx cannot exist in the best chain
-        for conflicting_tx in self.walk_conflicts(tx, |_, txid| self.get_tx_node(txid)) {
-            for block in conflicting_tx.anchors.iter().map(A::anchor_block) {
-                if chain.is_block_in_chain(block, chain_tip)? == Some(true) {
+        for conflicting_tx in ancestor_txs
+            .iter()
+            .flat_map(|&tx| self.walk_conflicts(tx, |_, txid| self.get_tx_node(txid)))
+        {
```
evanlinjin: Really small nit: for clarity, I would declare the iterator as a variable. Also, we can use `into_iter` to consume `ancestor_txs` since we aren't using it anymore.
```suggestion
        let conflicting_txs = ancestor_txs
            .into_iter()
            .flat_map(|ancestor| self.walk_conflicts(ancestor, |_, txid| self.get_tx_node(txid)));

        for conflicting_tx in conflicting_txs {
```

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,583 @@
...
+        assert_eq!(
+            utxos, exp_utxos,
+            "\n[{}] 'filter_chain_unspents' failed",
+            scenario.name
+        );
+        assert_eq!(
+            balance, scenario.exp_balance,
+            "\n[{}] 'balance' failed",
+            scenario.name
+        );
```
evanlinjin: Nit: I think this will make it easier to read.

```suggestion
        let txs = tx_graph
            .list_chain_txs(&local_chain, chain_tip)
            .map(|tx| tx.tx_node.txid)
            .collect::<BTreeSet<_>>();
        let exp_txs = scenario
            .exp_chain_txs
            .iter()
            .map(|txid| *exp_tx_ids.get(txid).expect("txid must exist"))
            .collect::<BTreeSet<_>>();
        assert_eq!(
            txs, exp_txs,
            "\n[{}] 'list_chain_txs' failed",
            scenario.name
        );

        let txouts = tx_graph
            .filter_chain_txouts(
                &local_chain,
                chain_tip,
                spk_index.outpoints().iter().cloned(),
            )
            .map(|(_, full_txout)| full_txout.outpoint)
            .collect::<BTreeSet<_>>();
        let exp_txouts = scenario
            .exp_chain_txouts
            .iter()
            .map(|(txid, vout)| OutPoint {
                txid: *exp_tx_ids.get(txid).expect("txid must exist"),
                vout: *vout,
            })
            .collect::<BTreeSet<_>>();
        assert_eq!(
            txouts, exp_txouts,
            "\n[{}] 'filter_chain_txouts' failed",
            scenario.name
        );

        let utxos = tx_graph
            .filter_chain_unspents(
                &local_chain,
                chain_tip,
                spk_index.outpoints().iter().cloned(),
            )
            .map(|(_, full_txout)| full_txout.outpoint)
            .collect::<BTreeSet<_>>();
        let exp_utxos = scenario
            .exp_unspents
            .iter()
            .map(|(txid, vout)| OutPoint {
                txid: *exp_tx_ids.get(txid).expect("txid must exist"),
                vout: *vout,
            })
            .collect::<BTreeSet<_>>();
        assert_eq!(
            utxos, exp_utxos,
            "\n[{}] 'filter_chain_unspents' failed",
            scenario.name
        );

        let balance = tx_graph.balance(
            &local_chain,
            chain_tip,
            spk_index.outpoints().iter().cloned(),
            |_, spk: &Script| spk_index.index_of_spk(spk).is_some(),
        );
        assert_eq!(
            balance, scenario.exp_balance,
            "\n[{}] 'balance' failed",
            scenario.name
        );
```

crates/chain/src/tx_graph.rs
```diff
@@ -319,6 +319,28 @@ impl<A> TxGraph<A> {
...
+    /// the original `Transaction`.
+    ///
+    /// The supplied closure takes in two inputs `(depth, ancestor_tx)`:
+    ///
+    /// * `depth` is the distance between the starting `Transaction` and the `ancestor_tx`. I.e., if
+    ///    the `Transaction` is spending an output of the `ancestor_tx`; the `depth` will be 1.
+    /// * `ancestor_tx` is the `Transaction`'s ancestor which we are considering to walk.
+    ///
+    /// The supplied closure returns an `Option<T>`, allowing the caller to map each `Transaction`
+    /// it visitsand decide whether to visit ancestors.
```
evanlinjin: typo

```suggestion
    /// it visits and decide whether to visit ancestors.
```

crates/chain/src/tx_graph.rs
```diff
@@ -319,6 +319,28 @@ impl<A> TxGraph<A> {
             .map(|(outpoint, spends)| (outpoint.vout, spends))
     }
 
+    /// Creates an iterator of ancestor transactions from the starting `Transaction`, not including
+    /// the original `Transaction`.
```
evanlinjin: **Nit:** I would first mention that this iterator that filters and maps. Then go on to explain where we start iterating from.

```suggestion
    /// Creates an iterator that filters and maps ancestor transactions.
    ///
    /// The iterator starts with the ancestors of the supplied `tx` (ancestor transactions of `tx`
    /// are transactions spent by `tx`). The supplied transaction is excluded from the iterator.
```

crates/chain/src/tx_graph.rs
```diff
@@ -337,7 +359,8 @@ impl<A> TxGraph<A> {
     }
 
     /// Creates an iterator that both filters and maps conflicting transactions (this includes
-    /// descendants of directly-conflicting transactions, which are also considered conflicts).
+    /// ancestors and descendants of directly-conflicting transactions, which are also considered
+    /// conflicts).
```
```
    /// Creates an iterator that both filters and maps directly-conflicting transactions and their
    /// descendants.
    ///
    /// Directly conflicting transactions of `tx` will have atleast one identical output spend as
    /// `tx`.
    ///
    /// **WARNING:** This will not include all conflicts as conflicts with `tx`'s ancestors and
    /// those ancestor's descendants are excluded.
    ///
    /// Refer to [`Self::walk_descendants`] for `walk_map` usage.
```

I think this is better.

crates/chain/src/tx_graph.rs
```diff
@@ -681,11 +705,29 @@ impl<A: Anchor> TxGraph<A> {
             }
         };
 
+        // Traverse tx and its unconfirmed ancestor txs to check for conflicts
```
evanlinjin: Code comments should explain ***why***!

```suggestion
        // Since transactions that conflict with the root tx are also conflicts, we need to traverse
        // all ancestor's conflicts. We only traverse unconfirmed ancestors since conflicts of 
        // confirmed transactions cannot be in the best chain.
```

crates/chain/src/tx_graph.rs
```diff
@@ -1149,6 +1191,123 @@ impl<A> ForEachTxOut for TxGraph<A> {
     }
 }
 
+/// An iterator that traverses unconfirmed ancestors that are whole `Transaction`s.
```
```suggestion
/// An iterator that traverses ancestors of a given root transaction.
///
/// The iterator excludes partial transactions.
```

crates/chain/src/tx_graph.rs
```diff
@@ -1149,6 +1191,123 @@ impl<A> ForEachTxOut for TxGraph<A> {
...
+        let (ancestor_depth, tx, item) = loop {
+            // we have exhausted all paths when stack is empty
+            let (ancestor_depth, tx) = self.stack.pop()?;
+            // ignore paths when user filters them out
+            if let Some(item) = (self.filter_map)(ancestor_depth, tx) {
+                break (ancestor_depth, tx, item);
+            }
+        };
+        self.populate_stack(ancestor_depth + 1, tx);
+        Some(item)
```
evanlinjin: Another nit.

```suggestion
        loop {
            // we have exhausted all paths when stack is empty
            let (ancestor_depth, tx) = self.stack.pop()?;

            // ignore paths when user filters them out
            let item = match (self.filter_map)(ancestor_depth, tx) {
                Some(item) => item,
                None => continue,
            };

            self.populate_stack(ancestor_depth + 1, tx);
            return Some(item);
        }
```

crates/chain/tests/test_tx_graph_conflicts.rs
evanlinjin: I just had an idea. I think it will be better if this is a method of `Scenario`. Then we can split these scenarios into separate test methods. This will make it easier to spot which test fails. If we do this, we can remove the `Scenario::name` field and just use the test function's name.

Let's do this is a separate PR, let's create a ticket for it.

evanlinjin (changes requested):
I'm really liking how this is turning out. I have some documentation suggestions. Otherwise, it's just a bunch of nits.

evanlinjin:
In terms of code, I am happy with everything - good job! I'll ACK this once we fix the commit messages.

* 809a98595188c9e70dc2d4a8a3829818bea6dcb6 should be split into two. One for the `chore` of moving tests. The other one for documentation changes and other tweaks.
* 686789b9f0d004161e1d57e14582333fec76739b and 2ade5474185d6950605366546b5a3de41a47c22a should be merged (as mentioned by @danielabrozzoni).

crates/chain/tests/common/tx_template.rs
```diff
@@ -0,0 +1,132 @@
...
+use std::collections::HashMap;
+
+use bdk_chain::{tx_graph::TxGraph, BlockId, SpkTxOutIndex};
+use bitcoin::{
+    locktime::absolute::LockTime, secp256k1::Secp256k1, OutPoint, ScriptBuf, Sequence, Transaction,
+    TxIn, TxOut, Txid, Witness,
+};
+use miniscript::Descriptor;
+
+/// Transaction template.
```
vladimirfomene: can we add more doc here to help anyone trying to understand the code why we decided to use this struct?

evanlinjin:
@LagginTimes Remember to rebase, we have conflicts!

danielabrozzoni:
>  I'll ACK this once we fix the commit messages.

In case you have never done that, https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History can be of help :)

evanlinjin (approved):
ACK 452627dacf4aae9e3636317223f37dae3f33397b

crates/chain/src/tx_graph.rs
```diff
@@ -319,6 +319,30 @@ impl<A> TxGraph<A> {
             .map(|(outpoint, spends)| (outpoint.vout, spends))
     }
 
+    /// Creates an iterator that filters and maps ancestor transactions.
+    ///
+    /// The iterator starts with the ancestors of the supplied `tx` (ancestor transactions of `tx`
```
LLFourn, line 325: I'm not sure this is true looking at the algorithm. If tx has ancestors `A` and `B`, it will pop off `A` from the stack and then add `A`'s ancestors to the stack. Unless I'm mistaken this means you'll visit `A`'s ancestors before visiting `B`.

The tests you've added don't test for the order they are emitted. Shouldn't they?
You could fix this in this PR or in a follow up because the problem seems to predate (it exists in `TxDescendants`).

crates/chain/src/tx_graph.rs
```diff
@@ -319,6 +319,30 @@ impl<A> TxGraph<A> {
...
+    /// are transactions spent by `tx`). The supplied transaction is excluded from the iterator.
+    ///
+    /// The supplied closure takes in two inputs `(depth, ancestor_tx)`:
+    ///
+    /// * `depth` is the distance between the starting `Transaction` and the `ancestor_tx`. I.e., if
+    ///    the `Transaction` is spending an output of the `ancestor_tx`; the `depth` will be 1.
```
```suggestion
    ///    the `Transaction` is spending an output of the `ancestor_tx` then `depth` will be 1.
```
^Grammatical error also exists in the `walk_descendants` function (`;` shouldn't be used to join conditional clauses; it's for independent clauses).

crates/chain/src/tx_graph.rs
```diff
@@ -337,7 +361,8 @@ impl<A> TxGraph<A> {
     }
 
     /// Creates an iterator that both filters and maps conflicting transactions (this includes
-    /// descendants of directly-conflicting transactions, which are also considered conflicts).
+    /// ancestors and descendants of directly-conflicting transactions, which are also considered
```
LLFourn: How does this walk ancestors? It only seems to walk direct conflicts and their descendants.

crates/chain/src/tx_graph.rs
```diff
@@ -681,11 +707,32 @@ impl<A: Anchor> TxGraph<A> {
             }
         };
 
+        // Since transactions that conflict with the root tx are also conflicts, we need to traverse
+        // all ancestor's conflicts. We only traverse unconfirmed ancestors since conflicts of
+        // confirmed transactions cannot be in the best chain.
+        let ancestor_txs =
```
```suggestion
        let unconfirmed_ancestor_txs =
```

crates/chain/src/tx_graph.rs
```diff
@@ -356,8 +381,9 @@ impl<A> TxGraph<A> {
     /// transaction's inputs (spends). The conflicting txids are returned with the given
     /// transaction's vin (in which it conflicts).
     ///
-    /// Note that this only returns directly conflicting txids and does not include descendants of
-    /// those txids (which are technically also conflicting).
+    /// Note that this only returns directly conflicting txids and may not include:
+    /// - descendants of conflicting transactions (which are technically also conflicting)
+    /// - transactions conflicting with the given transaction's ancestors
     pub fn direct_conflicts_of_tx<'g>(
```
LLFourn, line 387: side note. This function shouldn't be suffixed `_of_tx` unless we also suffix the other method so it's `walk_conflicts_of_tx`. I think we should just remove the suffix from here.

crates/chain/src/tx_graph.rs
```diff
@@ -681,11 +707,32 @@ impl<A: Anchor> TxGraph<A> {
...
+        let conflicting_txs = ancestor_txs
+            .into_iter()
+            .flat_map(|ancestor| self.walk_conflicts(ancestor, |_, txid| self.get_tx_node(txid)));
+
         // If a conflicting tx is in the best chain, or has `last_seen` higher than this tx, then
         // this tx cannot exist in the best chain
-        for conflicting_tx in self.walk_conflicts(tx, |_, txid| self.get_tx_node(txid)) {
-            for block in conflicting_tx.anchors.iter().map(A::anchor_block) {
-                if chain.is_block_in_chain(block, chain_tip)? == Some(true) {
+        for conflicting_tx in conflicting_txs {
```
LLFourn: I don't think the logic in this loop is right. Consider a transaction `B` with parent `A` which conflicts with `A'`. The last seen order is `B > A' > A`. Now this function will say that `B` is in the chain but `A`  is not because it conflicts with `A'` which has a higher last seen. I looked for a test for this to see if I'm wrong but it doesn't seem to be one.

LLFourn (changes requested):
Thanks @LagginTimes. I've pointed out what I think is a bug that I think should be fixed here. The other things I've identified I leave to you to resolve them here or we can do it in follow up work.

The key things we should (eventually) resolve are:

1. The worst case time complexity of the algorithm to determine whether a single tx is in the chain is roughly linear in the number of unconfirmed transactions. If we are determining whether a bunch of txouts are in the chain and every unconfirmed tx has one this means it's quadratic. I fear by peppering the mempool with a bunch of unconfirmed txs and RBFing them a lot you could denial of service attack systems using this algorithm.
2. I think it'd be better if `TxDescendents` and `TxAncestors` should be proximity to the root txid in the graph. The current order is depth first and at what "depth" the node is visited at depends on the order of the outputs or inputs that are being traversed. The iterator can give a depth of `7` to a tx that has a direct relationship to the root tx.

danielabrozzoni:
As discussed with @evanlinjin, I'm picking this up as it's high priority :)

crates/chain/src/tx_graph.rs
danielabrozzoni: Agreed, opened #1138 as this PR is quite big already

crates/chain/src/tx_graph.rs
danielabrozzoni: Agreed, I was confused as well. Fixed.

crates/chain/src/tx_graph.rs
danielabrozzoni: Agreed, I fixed by using a `VecDeque` in both `TxAncestors` and `TxDescendants`. Also, the tests now check the order.

crates/chain/src/tx_graph.rs
danielabrozzoni: Agreed, I fixed by using `direct_conflicts_of_tx` instead of `walk_conflicts`, and considering the last_seen of each ancestor instead of the last_seen of the root tx:
```diff
-        let conflicting_txs = ancestor_txs
-            .into_iter()
-            .flat_map(|ancestor| self.walk_conflicts(ancestor, |_, txid| self.get_tx_node(txid)));
-
-        // If a conflicting tx is in the best chain, or has `last_seen` higher than this tx, then
-        // this tx cannot exist in the best chain
-        for conflicting_tx in conflicting_txs {
-            for block in conflicting_tx.anchors {
-                if chain.is_block_in_chain(block.anchor_block(), chain_tip)? == Some(true) {
-                    // conflicting tx is in best chain, so the current tx cannot be in best chain!
+        for tx_node in unconfirmed_ancestor_txs {
+            // We retrieve all the transactions conflicting with this specific ancestor
+            let conflicting_txs = self
+                .direct_conflicts_of_tx(tx_node.tx)
+                .filter_map(|(_, txid)| self.get_tx_node(txid));
+
+            // If a conflicting tx is in the best chain, or has `last_seen` higher than this ancestor, then
+            // this tx cannot exist in the best chain
+            // Note that `direct_conflicts_of_tx`, which we used above to retrieve conflicts, doesn't include conflicting
+            // descendants, so comparing `last_seen` is safe
+            for conflicting_tx in conflicting_txs {
+                for block in conflicting_tx.anchors {
+                    if chain.is_block_in_chain(block.anchor_block(), chain_tip)? == Some(true) {
+                        // conflicting tx is in best chain, so the current tx cannot be in best chain!
+                        return Ok(None);
+                    }
+                }
+                if conflicting_tx.last_seen_unconfirmed > tx_node.last_seen_unconfirmed {
                     return Ok(None);
                 }
             }
-            if conflicting_tx.last_seen_unconfirmed > *last_seen {
-                return Ok(None);
-            }

```

I added the scenario "B and B' spend A and conflict, C spends B, all the transactions are unconfirmed, B' has higher last_seen than B" in the conflicting tests, which would previously fail and now passes. Lmk if it seems correct to you!

danielabrozzoni (commented):
I just pushed df1ee4f495478f6131df77a4cf5dbda115923fac:
- Rewrite commit history
- Fix review comments

As a part of fixing the review comments I added a scenario in `test_tx_graph_conflicts`, I just wanted to say @LagginTimes, the `Scenario/TxTemplate` mechanism is super convenient! Good job :)

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -0,0 +1,627 @@
...
+                TxTemplate {
+                    tx_name: "C",
+                    inputs: &[TxInTemplate::PrevTx("B", 0)],
+                    outputs: &[TxOutTemplate::new(30000, Some(3))],
+                    last_seen: Some(25),
+                    ..Default::default()
+                },
+            ],
+            // B and C should not appear in the list methods
+            exp_chain_txs: HashSet::from(["A", "B'"]),
```
LLFourn: Isn't what we want to happen here is that `B` and `C` are canonical?

`C` has the more recent last seen which should transitively mean that `B` is chosen over `B'` . In other words the effective `last_seen` of a node is the max of all its descendants.

This PR at least makes the answer consistent but it seems we are actually asserting suboptimal behavior here. Can be done in a follow up.

LLFourn (approved):
Approved. Consider fixing https://github.com/bitcoindevkit/bdk/pull/1064/files#r1344900282
ACK df1ee4f495478f6131df77a4cf5dbda115923fac

crates/chain/tests/test_tx_graph_conflicts.rs
danielabrozzoni: Just to give a bit of context, my reasoning behind this was: how would a node react if it saw B, then B', then C? And I think it would go like:
- B: fine, I'll keep this
- B': fine, I'll keep this, and drop B
- C: I'll drop this, as it's spending B, which I dropped already

Which I don't know if it's coherent with the rest of the txgraph and the last_seen logic (it probably isn't).

Anyways, I just pushed a fix, it should be fine now :)

evanlinjin (approved):
ACK 6d601a7e885bfc627594b794b1a460be47799eea

Thanks @danielabrozzoni for taking this forward and @LagginTimes for the initial work.

crates/chain/tests/test_tx_graph_conflicts.rs
LLFourn: > Which I don't know if it's coherent with the rest of the txgraph and the last_seen logic (it probably isn't).

Indeed. TxGraph is montone so doesn't record the order of things entering into it. Our conflict resolution mechanism is `last_seen` so I think we should take whichever subgraph has the tx with the latest one. Hopefully we can make this efficient!

PR #1069, opened 2023-08-09
Implement Anchor for BlockId

### LINKED ISSUE #1062: Implement `Anchor` for `BlockId`
**Describe the enhancement**  

The semantics of this would be that the confirmation block and the anchor block will be the same block. This should be clearly documented.

vladimirfomene:
### Description

This PR fixes #1062. 
It implements the `Anchor` trait on `BlockId`. 

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/chain_data.rs
+++ b/crates/chain/src/chain_data.rs
@@ -87,6 +87,9 @@ impl From<ChainPosition<ConfirmationTimeAnchor>> for ConfirmationTime {
 }
 
 /// A reference to a block in the canonical chain.
+///
+/// `BlockId` implements [`Anchor`]. When a transaction is anchored to `BlockId`, the confirmation
+/// block and anchor block are the same block.
 #[derive(Debug, Clone, PartialEq, Eq, Copy, PartialOrd, Ord, core::hash::Hash)]
 #[cfg_attr(
     feature = "serde",
@@ -98,6 +101,12 @@ pub struct BlockId {
     pub height: u32,
     /// The hash of the block.
     pub hash: BlockHash,
+}
+
+impl Anchor for BlockId {
+    fn anchor_block(&self) -> Self {
+        *self
+    }
 }
 
 impl Default for BlockId {
--- a/crates/chain/src/chain_data.rs
+++ b/crates/chain/src/chain_data.rs
@@ -140,6 +140,8 @@ impl From<(&u32, &BlockHash)> for BlockId {
 }
 
 /// An [`Anchor`] implementation that also records the exact confirmation height of the transaction.
+///
+/// Refer to [`Anchor`] for more details.
 #[derive(Debug, Default, Clone, PartialEq, Eq, Copy, PartialOrd, Ord, core::hash::Hash)]
 #[cfg_attr(
     feature = "serde",
@@ -168,6 +170,8 @@ impl Anchor for ConfirmationHeightAnchor {
 
 /// An [`Anchor`] implementation that also records the exact confirmation time and height of the
 /// transaction.
+///
+/// Refer to [`Anchor`] for more details.
 #[derive(Debug, Default, Clone, PartialEq, Eq, Copy, PartialOrd, Ord, core::hash::Hash)]
 #[cfg_attr(
     feature = "serde",
--- a/crates/chain/src/tx_data_traits.rs
+++ b/crates/chain/src/tx_data_traits.rs
@@ -44,6 +44,58 @@ impl ForEachTxOut for Transaction {
 /// assume that transaction A is also confirmed in the best chain. This does not necessarily mean
 /// that transaction A is confirmed in block B. It could also mean transaction A is confirmed in a
 /// parent block of B.
+///
+/// ```
+/// # use bdk_chain::local_chain::LocalChain;
+/// # use bdk_chain::tx_graph::TxGraph;
+/// # use bdk_chain::BlockId;
+/// # use bdk_chain::ConfirmationHeightAnchor;
+/// # use bdk_chain::example_utils::*;
+/// # use bitcoin::hashes::Hash;
+///
+/// // Initialize the local chain with two blocks.
+/// let chain = LocalChain::from_blocks(
+///     [
+///         (1, Hash::hash("first".as_bytes())),
+///         (2, Hash::hash("second".as_bytes())),
+///     ]
+///     .into_iter()
+///     .collect(),
+/// );
+///
+/// // Transaction to be inserted into `TxGraph`s with different anchor types.
+/// let tx = tx_from_hex(RAW_TX_1);
+///
+/// // Insert `tx` into a `TxGraph` that uses `BlockId` as the anchor type.
+/// // When a transaction is anchored with `BlockId`, the anchor block and the confirmation block of
+/// // the transaction is the same block.
+/// let mut graph_a = TxGraph::<BlockId>::default();
+/// let _ = graph_a.insert_tx(tx.clone());
+/// graph_a.insert_anchor(
+///     tx.txid(),
+///     BlockId {
+///         height: 1,
+///         hash: Hash::hash("first".as_bytes()),
+///     },
+/// );
+///
+/// // Insert `tx` into a `TxGraph` that uses `ConfirmationHeightAnchor` as the anchor type.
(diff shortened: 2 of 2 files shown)


notmandatory:
Is it possible to demonstrate using this in a test or example?

vladimirfomene:
@notmandatory let me look into it.

crates/chain/src/chain_data.rs
```diff
@@ -87,6 +87,7 @@ impl From<ChainPosition<ConfirmationTimeAnchor>> for ConfirmationTime {
 }
 
 /// A reference to a block in the canonical chain.
+/// When used as an anchor block, it is the same as the confirmation block.
```
evanlinjin: I prefer it written like this, with the `Anchor` stuff in a separate section.

```suggestion
///
/// `BlockId` implements [`Anchor`]. When a transaction is anchored to `BlockId`, the confirmation
/// block and anchor block are the same block.
```

crates/chain/src/chain_data.rs
```diff
@@ -100,6 +101,13 @@ pub struct BlockId {
     pub hash: BlockHash,
 }
 
+impl Anchor for BlockId {
+    // Here the returned anchored [`BlockId`] is the same as confirmation block.
```
evanlinjin: I don't think we need this, as we have explained this in the docs.

evanlinjin (changes requested):
Thanks for this! Just some thoughts.

vladimirfomene:
I think I will be using this for the Nakamoto syncing logic, @evanlinjin do you have an idea on how to write an example or test for this?

evanlinjin:
> I think I will be using this for the Nakamoto syncing logic, @evanlinjin do you have an idea on how to write an example or test for this?

I think a [documentation test](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html) for `Anchor` will be nice.

Like using a `LocalChain` with a single block, creating multiple `TxGraph`s with different anchor types and anchoring a single tx into the `TxGraph`s with the different anchor types.

evanlinjin:
And did `tokio` just break our MSRV?

crates/chain/src/tx_data_traits.rs
```diff
@@ -44,6 +44,52 @@ impl ForEachTxOut for Transaction {
...
+/// // the ConfirmationHeightAnchor, the anchor block is a bound on
+/// // the confirmation block of the transaction.
+/// graph_a.insert_anchor(
+///     tx.txid(),
+///     BlockId {
+///         height: 1,
+///         hash: Hash::hash("first".as_bytes()),
+///     },
+/// );
+/// graph_a.insert_anchor(
```
evanlinjin:
```suggestion
/// graph_b.insert_anchor(
```

evanlinjin: Also, I would prefer if the `graph_a` operations were together and the `graph_b` operations were together.

vladimirfomene:
Will rebase once #1077 gets merged

crates/chain/src/tx_data_traits.rs
```diff
@@ -44,6 +44,55 @@ impl ForEachTxOut for Transaction {
...
+///     .into_iter()
+///     .collect(),
+/// );
+///
+/// // Insert transaction into TxGraphs with different anchor types.
+/// // Insert different anchor types for the same transaction.
+/// // When using the BlockId as anchor, the anchor block is the
+/// // confirmation block of the transaction. Where as when using
+/// // the ConfirmationHeightAnchor, the anchor block is a bound on
+/// // the confirmation block of the transaction.
```
danielabrozzoni: nit: maybe we should explicitly say in the ConfirmationHeightAnchor (and ConfirmationTimeAnchor) docs that being anchored to a block doesn't mean that the transaction is confirmed in that block?

danielabrozzoni: Also, what do you mean by "the anchor block is *a bound on* the confirmation block"? I think it could be rephrased in a clearer way.

danielabrozzoni (commented):
Just a couple of nits :)

crates/chain/src/tx_data_traits.rs
vladimirfomene: I have updated the docs for ConfirmationHeightAnchor and ConfirmationTimeAnchor. I have also removed the confusing clause.

crates/chain/src/chain_data.rs
```diff
@@ -131,6 +140,9 @@ impl From<(&u32, &BlockHash)> for BlockId {
 }
 
 /// An [`Anchor`] implementation that also records the exact confirmation height of the transaction.
+/// When a transaction is anchored in [`Self::anchor_block`], it does not mean
+/// it is confirmed in that block. It just means [`Self::anchor_block`] is in the best chain and
+/// therefore our transaction is also in the best chain.
```
evanlinjin: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

Can we stick to this structure when writing documentation?

We want the first section to be a short sentence, and have a more detailed description in a separate section.

Please also fix this for `ConfirmationTimeAnchor`.

evanlinjin:
```suggestion
///
/// Refer to [`Anchor`] for more.
```

Instead of including this explanation for every struct that implements `Anchor`, I think it's better to move this description to `Anchor`'s documentation. We can just have "Refer to Anchor for more" at the end here.

crates/chain/src/tx_data_traits.rs
```diff
@@ -44,6 +44,53 @@ impl ForEachTxOut for Transaction {
...
+///     ]
+///     .into_iter()
+///     .collect(),
+/// );
+///
+/// // Insert transaction into TxGraphs with different anchor types.
+/// // Insert different anchor types for the same transaction.
+/// // When using the BlockId as anchor, the anchor block is the
+/// // confirmation block of the transaction.
+/// let tx = tx_from_hex(RAW_TX_1);
```
evanlinjin:
```suggestion
/// // Transaction to be inserted into `TxGraph`s with different anchor types.
/// let tx = tx_from_hex(RAW_TX_1);
///
/// // Insert `tx` into a `TxGraph` that uses `BlockId` as the anchor type.
/// // When a transaction is anchored with `BlockId`, the anchor block and the confirmation block of
/// // the transaction is the same block.
```

crates/chain/src/tx_data_traits.rs
```diff
@@ -44,6 +44,53 @@ impl ForEachTxOut for Transaction {
...
+/// let mut graph_a = TxGraph::<BlockId>::default();
+/// let _ = graph_a.insert_tx(tx.clone());
+/// graph_a.insert_anchor(
+///     tx.txid(),
+///     BlockId {
+///         height: 1,
+///         hash: Hash::hash("first".as_bytes()),
+///     },
+/// );
+///
```
evanlinjin, line 81:
```suggestion
///
/// // Insert `tx` into a `TxGraph` that uses `ConfirmationHeightAnchor` as the anchor type.
/// // When a transaction is anchored with `ConfirmationHeightAnchor`, the anchor block and
/// // confirmation block can be different. However, the confirmation block cannot be higher than
/// // the anchor block and block blocks must be in the same chain for the anchor to be valid.
```

evanlinjin (changes requested):
Good work, we are nearly there. Documentation can be better.

crates/chain/src/tx_data_traits.rs
```diff
@@ -44,6 +44,59 @@ impl ForEachTxOut for Transaction {
...
+/// graph_b.insert_anchor(
+///     tx.txid(),
+///     ConfirmationHeightAnchor {
+///         anchor_block: BlockId {
+///             height: 2,
+///             hash: Hash::hash("second".as_bytes()),
+///         },
+///         confirmation_height: 1,
+///     },
+/// );
```
evanlinjin: Nit: Wouldn't it be clearer to keep `graph_a` operations together and `graph_b` operations together? The other option would be to declare/define `graph_a` and `graph_b` in the same place.

evanlinjin (approved):
ACK 18e8da393779617e8a501d7e21b0a9db1a8b11f1

PR #1071, opened 2023-08-09
Update rust bitcoin (BDK 0.28)

danielabrozzoni:
### Changelog notice

- Update rust-bitcoin to 0.30

#### All Submissions:

### DIFF
--- a/examples/utils/mod.rs
+++ b/examples/utils/mod.rs
@@ -13,7 +13,10 @@ pub(crate) mod tx {
         // Create a transaction builder
         let mut tx_builder = wallet.build_tx();
 
-        let to_address = Address::from_str(recipient_address).unwrap();
+        let to_address = Address::from_str(recipient_address)
+            .unwrap()
+            .require_network(wallet.network())
+            .unwrap();
 
         // Set recipient of the transaction
         tx_builder.set_recipients(vec![(to_address.script_pubkey(), amount)]);
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -24,10 +24,11 @@ use std::sync::Arc;
 use bitcoin::secp256k1::Secp256k1;
 
 use bitcoin::consensus::encode::serialize;
-use bitcoin::util::psbt;
+use bitcoin::psbt;
+use bitcoin::sighash::{EcdsaSighashType, TapSighashType};
 use bitcoin::{
-    Address, EcdsaSighashType, LockTime, Network, OutPoint, SchnorrSighashType, Script, Sequence,
-    Transaction, TxOut, Txid, Witness,
+    absolute, Address, Network, OutPoint, Script, ScriptBuf, Sequence, Transaction, TxOut, Txid,
+    Weight, Witness,
 };
 
 use miniscript::psbt::{PsbtExt, PsbtInputExt, PsbtInputSatisfier};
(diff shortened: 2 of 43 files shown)


danielabrozzoni:
CBF is broken but I'm too tired to look into it right now 😅

notmandatory:
FYI in the past couple days some dependencies broke for 1.57.0 MSRV, I'm fixing it in #1074 .

examples/utils/mod.rs
```diff
@@ -13,7 +13,9 @@ pub(crate) mod tx {
         // Create a transaction builder
         let mut tx_builder = wallet.build_tx();
 
-        let to_address = Address::from_str(recipient_address).unwrap();
+        let to_address = Address::from_str(recipient_address)
+            .unwrap()
+            .assume_checked();
```
notmandatory: Small thing, but since we want folks to use new best practice I think we should pass in `Network` and use `require_network()` here, and then will also need to return a `Result<Transaction, Error>`.

src/wallet/mod.rs
```diff
@@ -4803,7 +4912,7 @@ pub(crate) mod test {
         let (wallet, _, _) = get_funded_wallet(get_test_tr_repeated_key());
         let addr = wallet.get_address(AddressIndex::New).unwrap();
 
-        let path = vec![("e5mmg3xh".to_string(), vec![0])]
+        let path = vec![("rn4nre9c".to_string(), vec![0])]
```
notmandatory, line 4915: Is it expected that this policy path changed?

src/wallet/signer.rs
```diff
@@ -382,6 +381,49 @@ impl InputSigner for SignerWrapper<DescriptorXKey<ExtendedPrivKey>> {
...
+    fn id(&self, secp: &SecpCtx) -> SignerId {
+        SignerId::from(self.root_fingerprint(secp))
+    }
+
+    fn descriptor_secret_key(&self) -> Option<DescriptorSecretKey> {
+        Some(DescriptorSecretKey::MultiXPrv(self.signer.clone()))
+    }
+}
+
+impl InputSigner for SignerWrapper<DescriptorMultiXKey<ExtendedPrivKey>> {
```
notmandatory, line 411: With this do we now support BIP 389 multixkey's in descriptors?

notmandatory (commented):
LGTM! I just have one suggestion and a question.

src/wallet/signer.rs
danielabrozzoni: Not really, this is some of the code needed to support multikey, but it's not all we need and using multikey right now is a bit broken (I tried some time ago, I don't rememeber how).

For this exact reason, f7e72223f8e21e238da8e60cb540086701e3fdf0 denies multipath keys

examples/utils/mod.rs
danielabrozzoni: Fixed, I didn't return a result as we `unwrap` everywhere in the method

src/wallet/mod.rs
danielabrozzoni: I'm not sure why it's happening, I had to change it on the 1.0 PR as well... Anyways, I checked manually and this one should be correct

examples/utils/mod.rs
notmandatory: Good call, agreed that it's best to keep it simple like the rest and unwrap.

notmandatory:
FYI the CI issue is probably the same one @vladimirfomene ran into yesterday on his PR. I should be able to get a PR in today to fix it.

danielabrozzoni:
Now it should be fine :)

notmandatory (approved):
ACK d1af252ac43d79e17a2751cfefea0e4596e0a02a

Issue #1066, opened 2023-08-07
Absolute fee constraint in TxBuilder not being respected


sjeohp:
When creating a tx the absolute fee amount when set does not always match the actual fee amount. I see this is because the amount is added to several times, so it must be deliberate, but for someone using the library it is unintuitive. What is the rationale?

notmandatory:
Can you also give a little code example and which version of bdk are you using? This would help us know which documentation needs to be improved.

sjeohp:
crates/bdk/wallet/mod.rs
    fn create_tx

sjeohp:
For example 

    let mut tx_builder = wallet.build_tx();
    tx_builder.fee_absolute(fee);
    let (psbt, details) = tx_builder.finish()?;

In my experience the fee will not necessarily match the fee that was set.

LLFourn:
It's worth pointing out that the "absolute fee" is a really a minimum absolute fee -- it's possible to overshoot it slightly since adding a change output to drian the remaining excess might not be viable.

How far off is the fee vs expected?

sjeohp:
Making tiny txs on testnet the fee varies by 100-150%.

Is there a way to upper bound it? Or reduce variability at the cost of something else?

LLFourn:
Appreciated the work looking into this. No but I am not convinced that the implementation isn't simply wrong with respect to absolute fees. Is the fee wrong even when a change output is added? That would be the next question since the builder adds a change output and the resulting tx does not have the correct fee that would mean there's a bug I think.

sjeohp:
Yes same thing with the change descriptor.

With `<desc>` having 4500 satoshis and `<change_desc>` having 1000 satoshis or 0 (no difference):
`bdk-cli wallet -d <desc> -c <change_desc>  create_tx --to <receiver_addr>:500 --fee_abs 250` returns a fee of 500 most times, sometimes 250.

Also I don't exactly understand this:

`bdk-cli wallet -d <DESC_A> -c <DESC_B> get_balance` returns the balance of both wallets combined but needs to be `sync`ed with both combined. If both are synced separately or inverted like `bdk-cli wallet -d <DESC_B> -c <DESC_A> sync` the balance will be wrong and `create_tx` might fail.

LLFourn:
> Yes same thing with the change descriptor.
> 
[...]

PR #1072, opened 2023-08-09
Introduce new `bdk_coin_select` implementation

evanlinjin:
### Description

This replaces bitcoindevkit/bdk#924 (which in turn, ports over https://github.com/LLFourn/bdk_core_staging/pull/46).

As stated by @LLFourn in https://github.com/bitcoindevkit/bdk/pull/924#issuecomment-1600066366, we will complete the following in this PR:

- [x] Add minimize fee metric (it's a better default that waste see https://github.com/wizardsardine/liana/pull/560#discussion_r1236237579) 
- [x] Allow easy weighted combination of different metrics (see same link above)
- [x] Document everything including nice README.md exmaples
- [x] Make DWIM branch and bound helper function (https://github.com/wizardsardine/liana/pull/560#discussion_r1236232805)
- [x] Make `DrainWeight` (https://github.com/wizardsardine/liana/pull/560#discussion_r1236246084)
- [x] Add `CoinSelector::fund_outputs` (see https://github.com/wizardsardine/liana/pull/560#discussion_r1236251001) -- another point is that in general we need to know the number of outputs to get the varint for the outputs correct which I think we are cheating on atm.
- [x] Test cases against real world transactions to see if coin selections have precisely the real world feerate that was expected. (EDIT: I've just done a few of these against the feerate shown mempool.space).
- [x] Rewrite examples to use this crate.
- [x] Remove all `TODO` in the new crate and make issues for further work.

### Notes to the reviewers

~

### Changelog notice

Introduce new `bdk_coin_select` module

#### All Submissions:

#### New Features:


notmandatory:
This looks like something worthy of a 2.0 release milestone. Will current 1.0 coin selection be at least on par with pre-1.0 BDK coin selection?

LLFourn:
> This looks like something worthy of a 2.0 release milestone. Will current 1.0 coin selection be at least on par with pre-1.0 BDK coin selection?

The plan for this PR is to release the crate on crates.io. Gutting BDK wallet coin selection and replacing it with this is another problem altogether. Using it in bdk wallet would require the planning module bitcoindevkit/bdk_wallet#139 is first ready to be released. I think the plan is to have a crate `bdk_tx_builder` or something that will put all these things together and can be used without `Wallet`.

notmandatory:
> > This looks like something worthy of a 2.0 release milestone. Will current 1.0 coin selection be at least on par with pre-1.0 BDK coin selection?
> 
> The plan for this PR is to release the crate on crates.io. Gutting BDK wallet coin selection and replacing it with this is another problem altogether. Using it in bdk wallet would require the planning module bitcoindevkit/bdk_wallet#139 is first ready to be released. I think the plan is to have a crate `bdk_tx_builder` or something that will put all these things together and can be used without `Wallet`.

Got it, I won't mark this for any BDK major release milestones. But eventually integrating this into BDK along with bitcoindevkit/bdk_wallet#139 sounds like a good goals for a future 2.0 milestone.

notmandatory:
Per discussion in  bitcoindevkit/.github#72 this will be stand-alone framework, not to be integrate with BDK wallet until a future 2.0 milestone.

nondiremanuel:
Even if it will be ready together with alpha.3, being that it won't be integrated until a future 2.0 version, I would assign it to the 2.0 milestone. In general I propose to use milestones to signal which will be the release that will include a specific PR, regardless of the timing of development (so even if it is ready earlier), but maybe I didn't understand correctly the topic here

LLFourn:
Depending on how things go we could replace the existing coin selection with this one without doing the full redesign of tx building in `bdk` and put in `alpha.3`. For example, instead of:

```
tx_builder = tx_builder.coin_selection::<SomeCoinSelectionAlgo>();
```
This could be improved to:

```rust
tx_builder = tx_builder.coin_selection(|coin_selector| {
    /*This will be called once we've got all the candidates. Do your coin selection with coin_selector and return it*/
});
```
Internally `tx_builder` would store a `where for<'a> F: FnOnce(CoinSelector<'a>) -> Result<CoinSelector<'a>, E>` which would have a sensible default (`F` and `E` would become type parameters on `TxBuilder`). The benefit would be getting the flexibility and power and correctness of the new coin selection code. This would likely fix bitcoindevkit/bdk#1066 which arises mainly from the current architecture of coin selection where each coin selection algorithm can choose to return coin selections that violate the constraints given to it (in this case absolute fee).

@evanlinjin would be a better judge of whether this is a good idea.

evanlinjin:
@darosior I'll give a guesstimate of two more weeks before this is in a good state.

darosior:
Cool, thanks for keeping me up to date.

-------- Original Message --------

darosior:
We should be able to get back to testing this on our end within the next week or two, for what it's worth.

darosior:
Let me know when you think it's a state close enough to final for us to revive https://github.com/wizardsardine/liana/pull/560. (No rush, @jp1ac4 has plenty on his plate already!)

evanlinjin:
@darosior I'm confident in the stability, accuracy and performance of the `LowestFee` metric (the metric suggested by @LLFourn in https://github.com/wizardsardine/liana/pull/560#discussion_r1236237579) because of the extensive amount of prop-testing and real-world testing done. This is what I would recommend using for the bnb metric.

For the change policy, I think using `min_value_and_waste` will make the most sense.

Currently, the `Waste` metric's bound function is too tight but I haven't had too much time to look deeply into it (yet).

My next steps is to add more documentation and improve the accuracy of the `Waste` metric's bound function.

Sorry for the delays, I have overestimated myself.

LLFourn:
IMO we could just leave improving the waste metric for future work. Nothing terrible happens if the bounds are a bit too tight.

What's left to be done in the actual API?

evanlinjin:
> IMO we could just leave improving the waste metric for future work. Nothing terrible happens if the bounds are a bit too tight.
> 
> What's left to be done in the actual API?

I think just documentation...

LLFourn:
Awesome! Let's try and get this one out there with a basic README.md example on how to use it and just do a release and see what happens.

darosior:
For what it's worth we've resumed working on this on our side. @jp1ac4 will resume working on https://github.com/wizardsardine/liana/issues/51, starting with updating his PR at https://github.com/wizardsardine/liana/pull/560 to the new API.

nursery/coin_select/src/coin_selector.rs
```diff
@@ -35,7 +35,8 @@ impl Default for Target {
     fn default() -> Self {
         Self {
             feerate: FeeRate::default_min_relay_fee(),
-            min_fee: 0, // TODO figure out what the actual network rule is for this
+            // https://bitcoin.stackexchange.com/questions/69282/what-is-the-min-relay-min-fee-code-26
+            min_fee: 1000,
```
jp1ac4, line 39: I think the `relayfee` mentioned in that link is rather a feerate given in BTC/kvB:
https://github.com/bitcoin/bitcoin/blob/82ea4e787c791acbc85fd3043dd6bae038cba4f2/src/rpc/net.cpp#L643

evanlinjin: Thanks for pointing this out. Definitely a bad mistake.

evanlinjin:
@jp1ac4 @darosior I'm working on some breaking changes to the API.

After a discussion with @LLFourn, we've decided to include `DrainWeights` in `Target` and keep `Target` in the `CoinSelector`. This simplifies the API and gives us an easy pathway to take into account the output count varint weight changes when including the drain output.

How this simplifies the API:

* The target & drain weights are included in the `CoinSelector`, so they don't need to be passed in as method inputs.
* Change policy can be simplified to just: `Fn(&CoinSelector) -> bool`.

How we can take into account the output varint weight changes when including drain:

```rust
pub struct Target {
    /*other fields*/

    pub base_weight: u32,
    pub drain_weights: DrainWeights,
}

impl Target {
    pub fn new(
        feerate: FeeRate,
        recipient_outputs: impl IntoIterator<Item = (u32 /*weight*/, u64 /*value*/)>
        drain_outputs: impl IntoIterator<Item = DrainWeights>,
    ) -> Self {
        todo!()
    }
}
```

This calculates the `base_weight` and the `drain_weight` (which includes output varint changes) that are both included in `Target`.

nursery/coin_select/src/coin_selector.rs
darosior: Outside of replacements there is no minimum absolute fee rule. It's implicit from the minimum feerate (1 sat/vb) and how transactions cannot have a size of 0.

darosior:
Sounds good. On our side the implementation is getting there and should be ready to merge soon-ish.

evanlinjin:
@darosior would you prefer a release before the breaking changes?

LLFourn:
> @jp1ac4 @darosior I'm working on some breaking changes to the API.
> 
> After a discussion with @LLFourn, we've decided to include `DrainWeights` in `Target` and keep `Target` in the `CoinSelector`. This simplifies the API and gives us an easy pathway to take into account the output count varint weight changes when including the drain output.

So in my mind `Target` and `DrainWeights` would still be separate things just both exposed and settable publicly. `Target` would include the number of outputs and `DrainWeights` would include the number of outputs in the draining output. I also think you need to include the "base weight" in `Target` as it is coupled to the number of outputs.

Now I thought further -- if we create this invariant I think this also allows us to reduce the `change_policy` to a single integer -- i.e. the threshold of excess coin which should trigger the inclusion of the drain output. Wdyt @evanlinjin?

darosior:
> @darosior would you prefer a release before the breaking changes?

Yes, that would be great.

darosior:
Ran into a crash while testing this on our end:
```
  2023-11-09T14:57:43.020092Z ERROR liana:55: panic occurred at line 183 of file /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/metrics/lowest_fee.rs: Some("value:weight ratio must stay the same: vpw=2375810 perfect_vpw=2375809.8 perfect_value=300001060 perfect_weight=126.27318")
   0: liana::setup_panic_hook::{{closure}}
             at /home/darosior/.cargo/git/checkouts/liana-d9e0035d1d73e045/ae5c1c3/src/lib.rs:49:18
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/alloc/src/boxed.rs:2021:9
      std::panicking::rust_panic_with_hook
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:711:13
   2: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:599:13
   3: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/sys_common/backtrace.rs:170:18
   4: rust_begin_unwind
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:595:5
   5: core::panicking::panic_fmt
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:67:14
   6: bdk_coin_select::metrics::lowest_fee::slurp
             at /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/metrics/lowest_fee.rs:183:13
   7: <bdk_coin_select::metrics::lowest_fee::LowestFee<C> as bdk_coin_select::bnb::BnbMetric>::bound
             at /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/metrics/lowest_fee.rs:160:36
   8: bdk_coin_select::bnb::BnbIter<M>::consider_adding_to_queue
             at /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/bnb.rs:93:21
   9: bdk_coin_select::bnb::BnbIter<M>::new
             at /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/bnb.rs:87:9
  10: bdk_coin_select::coin_selector::CoinSelector::bnb_solutions
             at /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/coin_selector.rs:478:9
  11: bdk_coin_select::coin_selector::CoinSelector::run_bnb
             at /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/coin_selector.rs:493:33
  12: liana::commands::utils::select_coins_for_spend
             at /home/darosior/.cargo/git/checkouts/liana-d9e0035d1d73e045/ae5c1c3/src/commands/utils.rs:170:21
  13: liana::commands::<impl liana::DaemonControl>::create_spend
             at /home/darosior/.cargo/git/checkouts/liana-d9e0035d1d73e045/ae5c1c3/src/commands/mod.rs:479:13
  14: <liana_gui::daemon::embedded::EmbeddedDaemon as liana_gui::daemon::Daemon>::create_spend_tx
             at src/daemon/embedded.rs:91:9
  15: liana_gui::app::state::spend::step::DefineSpend::auto_select_coins
             at src/app/state/spend/step.rs:189:15
```

@jp1ac4 points out:
> All values are `f32` but not all have decimal points. From some quick tests, I found that `2375809.9 as f32` prints as `2375810` while `2375809.8 as f32` prints as `2375809.8`. So from that it seems an issue with type conversions and rounding.

(I found surprising you'd opt for using floats here. If you'd consider changing for `u64` that would avoid float footguns as well as some conversions downstream.)

LLFourn:
> Ran into a crash while testing this on our end:
> 
> ```
>   2023-11-09T14:57:43.020092Z ERROR liana:55: panic occurred at line 183 of file /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/metrics/lowest_fee.rs: Some("value:weight ratio must stay the same: vpw=2375810 perfect_vpw=2375809.8 perfect_value=300001060 perfect_weight=126.27318")
>    0: liana::setup_panic_hook::{{closure}}
>              at /home/darosior/.cargo/git/checkouts/liana-d9e0035d1d73e045/ae5c1c3/src/lib.rs:49:18
>    1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
>              at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/alloc/src/boxed.rs:2021:9
>       std::panicking::rust_panic_with_hook
>              at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:711:13
>    2: std::panicking::begin_panic_handler::{{closure}}
>              at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:599:13
>    3: std::sys_common::backtrace::__rust_end_short_backtrace
>              at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/sys_common/backtrace.rs:170:18
>    4: rust_begin_unwind
>              at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:595:5
>    5: core::panicking::panic_fmt
>              at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:67:14
>    6: bdk_coin_select::metrics::lowest_fee::slurp
>              at /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/metrics/lowest_fee.rs:183:13
>    7: <bdk_coin_select::metrics::lowest_fee::LowestFee<C> as bdk_coin_select::bnb::BnbMetric>::bound
>              at /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/metrics/lowest_fee.rs:160:36
>    8: bdk_coin_select::bnb::BnbIter<M>::consider_adding_to_queue
>              at /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/bnb.rs:93:21
>    9: bdk_coin_select::bnb::BnbIter<M>::new
>              at /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/bnb.rs:87:9
>   10: bdk_coin_select::coin_selector::CoinSelector::bnb_solutions
>              at /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/coin_selector.rs:478:9
>   11: bdk_coin_select::coin_selector::CoinSelector::run_bnb
>              at /home/darosior/.cargo/git/checkouts/bdk-0dee5b29605e7dd4/7607975/nursery/coin_select/src/coin_selector.rs:493:33
>   12: liana::commands::utils::select_coins_for_spend
>              at /home/darosior/.cargo/git/checkouts/liana-d9e0035d1d73e045/ae5c1c3/src/commands/utils.rs:170:21
>   13: liana::commands::<impl liana::DaemonControl>::create_spend
>              at /home/darosior/.cargo/git/checkouts/liana-d9e0035d1d73e045/ae5c1c3/src/commands/mod.rs:479:13
>   14: <liana_gui::daemon::embedded::EmbeddedDaemon as liana_gui::daemon::Daemon>::create_spend_tx
>              at src/daemon/embedded.rs:91:9
>   15: liana_gui::app::state::spend::step::DefineSpend::auto_select_coins
>              at src/app/state/spend/step.rs:189:15
> ```
> 
> @jp1ac4 points out:
> 
> > All values are `f32` but not all have decimal points. From some quick tests, I found that `2375809.9 as f32` prints as `2375810` while `2375809.8 as f32` prints as `2375809.8`. So from that it seems an issue with type conversions and rounding.
> 
> (I found surprising you'd opt for using floats here. If you'd consider changing for `u64` that would avoid float footguns as well as some conversions downstream.)

We are using `f32` in here because this function returns a value to help score a coin selection using a heuristic. I think using other representations comes with more risk than benefits for this e.g. integers can overflow. The imprecision will not cause serious problems since it at worst only causes the heuristic to be off. Having a more accurate heuristic is not necessarily a benefit if the accuracy comes with a performance cost.

But maybe I'm missing something. Why did you introduce this debug assertion @evanlinjin? We're you getting sub-optimal selections during proptesting? I assume it was just some sanity check you introduced during development and forgot to remove. The 0.1 seems very arbitrary.

Happy to explore other approaches if they can be shown to improve the results of the selection. There's a lot of space for bench-marking and optimization. I'd say for now just remove this debug assertion.

evanlinjin:
@LLFourn Yes I was getting non-optimal results. This assert is to check whether the "perfect weight" calculation is correct. The `0.01` is arbitrary and is used to compensate for the round-off error(s). I'll remove this debug statement for now.

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -836,7 +836,7 @@ mod test {
         let drain_script = ScriptBuf::default();
         let target_amount = 250_000 + FEE_AMOUNT;
 
-        let result = LargestFirstCoinSelection::default()
+        let result = LargestFirstCoinSelection
```
LLFourn, line 839: huh why are we changing this file?

build-msrv-crates.sh
```diff
@@ -0,0 +1,16 @@
...
+    cleanup
+    exit $signal;
+' INT
+
+cleanup() {
+    mv Cargo.tmp.toml Cargo.toml 2>/dev/null
+}
+
+cp Cargo.toml Cargo.tmp.toml
+cp Cargo.1.48.0.toml Cargo.toml
```
LLFourn, line 13: I think we should instead just put this code in a separate repo in the bdk org and have our own MSRV for it.

nursery/coin_select/README.md
```diff
@@ -0,0 +1,189 @@
...
+candidate UTXOs and a transaction `base_weight`. The `base_weight` includes the recipient outputs
+and mandatory inputs (if any).
+
+```rust
+use std::str::FromStr;
+use bdk_coin_select::{ CoinSelector, Candidate, TXIN_BASE_WEIGHT };
+use bitcoin::{ Address, Network, Transaction, TxIn, TxOut };
+
+// You should use miniscript to figure out the satisfaction weight for your coins!
+const TR_SATISFACTION_WEIGHT: u32 = 66;
```
LLFourn, line 18: We actually have this as a const in `bdk_coin_select`. Keep the comment saying something "In general, you can use miniscript to figure out the satisfaction weight of your inputs".

nursery/coin_select/README.md
```diff
@@ -0,0 +1,189 @@
...
+    .expect("network must match");
+
+let candidates = vec![
+    Candidate {
+        // How many inputs does this candidate represent. Needed so we can 
+        // figure out the weight of the varint that encodes the number of inputs.
+        input_count: 1,
+        // the value of the input
+        value: 1_000_000,
+        // the total weight of the input(s). This doesn't include
```
LLFourn, line 35: What doesn't it include (I know this was probably my fault)?. Maybe just remove that.

nursery/coin_select/README.md
```diff
@@ -0,0 +1,189 @@
...
+    input: vec![],
+    // include your recipient outputs here
+    output: vec![TxOut {
+        value: 900_000,
+        script_pubkey: recipient_addr.script_pubkey(),
+    }],
+    lock_time: bitcoin::absolute::LockTime::from_height(0).unwrap(),
+    version: 0x02,
+};
+let base_weight = base_tx.weight().to_wu() as u32;
```
LLFourn, line 61: since rust bitcoin uses u64 here should we also? It can be u32 internally but we can cast it.

nursery/coin_select/README.md
```diff
@@ -0,0 +1,189 @@
...
+println!("base weight: {}", base_weight);
+
+// You can now select coins!
+let mut coin_selector = CoinSelector::new(&candidates, base_weight);
+coin_selector.select(0);
```
LLFourn, line 66: Let's show:

- how to check whether the selection is finished?
- figure out which coins were selected

nursery/coin_select/README.md
```diff
@@ -0,0 +1,189 @@
...
+// The drain output(s) may or may not be included in the final tx. We calculate
+// the drain weight to include the output length varint weight changes from
+// including the drain output(s).
+let drain_output_weight = {
+    let mut tx_with_drain = base_tx.clone();
+    tx_with_drain.output.push(TxOut {
+        script_pubkey: drain_addr.script_pubkey(),
+        ..Default::default()
+    });
+    tx_with_drain.weight().to_wu() as u32 - base_weight
```
LLFourn, line 112: Can we just use [`TxOut::weight`](https://docs.rs/bitcoin/latest/bitcoin/blockdata/transaction/struct.TxOut.html#method.weight)?

nursery/coin_select/README.md
```diff
@@ -0,0 +1,189 @@
...
+// We use a change policy that introduces a change output if doing so reduces
+// the "waste" and that the change output's value is at least that of the 
+// `dust_limit`.
+let change_policy = min_value_and_waste(
+    drain_weights,
+    dust_limit,
+    long_term_feerate,
+);
+
+// This metric minimizes transaction fee. The `long_term_feerate` is used to
```
LLFourn, line 166:
```suggestion
// This metric minimizes transaction fees paid over time. The `long_term_feerate` is used to
```

nursery/coin_select/src/coin_selector.rs
```diff
@@ -1,617 +1,715 @@
...
+    ///
+    /// The method keeps trying until no better solution can be found, or we reach `max_rounds`. If
+    /// a solution is found, the score is returned. Otherwise, we error with [`NoBnbSolution`].
+    ///
+    /// Use [`CoinSelector::bnb_solutions`] to access the branch and bound iterator directly.
+    pub fn run_bnb<M: BnbMetric>(
+        &mut self,
+        metric: M,
+        max_rounds: usize,
+    ) -> Result<Ordf32, NoBnbSolution> {
```
LLFourn, line 498: When i was suggesting this method I was thinking that it would fall back to `select_until_target_met` if it can't find anything. This would mean you don't need this error type -- rather you can just have `InsufficientFunds`.

Do you think it's worth having `run_bnb_with_fallback()` that does that?

Also another sneaky request -- if `#[cfg(feature = "std")]` can you add a method that lets you pass in a `std::time::Duration` and checks the time passed every 1_000 rounds and stops after the time has been exceeded.

nursery/coin_select/src/feerate.rs
```diff
@@ -0,0 +1,91 @@
+use crate::float::Ordf32;
+use core::ops::{Add, Sub};
+
+/// Fee rate
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+// Internally stored as satoshi/weight unit
+pub struct FeeRate(Ordf32);
```
LLFourn, line 7: might be worth manually implementing `Debug` so the format prints sats per vb in the first place like: `8 sat/vb (2 sat/wu)`

nursery/coin_select/src/feerate.rs
```diff
@@ -0,0 +1,91 @@
...
+        Self::from_sat_per_vb(rate)
+    }
+
+    /// Return the value as satoshi/vbyte.
+    pub fn as_sat_vb(&self) -> f32 {
+        self.0 .0 * 4.0
+    }
+
+    /// Return the value as satoshi/wu.
+    pub fn spwu(&self) -> f32 {
```
LLFourn, line 72:
```suggestion
    pub fn sat_per_wu(&self) -> f32 {
```

we can keep `fn spwu` as `pub(crate)` but in the API it should be more descriptive.

nursery/coin_select/src/metrics/lowest_fee.rs
```diff
@@ -0,0 +1,175 @@
...
+                }
+                None => 0.0,
+            }
+    }
+
+    fn calc_metric_lb(&self, cs: &CoinSelector<'_>, drain_weights: Option<DrainWeights>) -> f32 {
+        match drain_weights {
+            // with change
+            Some(drain_weights) => {
+                (cs.input_weight() + drain_weights.output_weight) as f32
```
LLFourn, line 51: Probably worth throwing in the `base_weight` in here too to make the metric match the result.

I had a helper method on `CoinSelector` for this: `implied_fee` -- might be more elegant to make it public and use it. You're not taking into account the `min_fee` which seems like it could cause the wrong answer.

nursery/coin_select/src/metrics/lowest_fee.rs
```diff
@@ -0,0 +1,175 @@
...
+where
+    for<'a, 'b> C: Fn(&'b CoinSelector<'a>, Target) -> Drain,
+{
+    fn calc_metric(&self, cs: &CoinSelector<'_>, drain_weights: Option<DrainWeights>) -> f32 {
+        self.calc_metric_lb(cs, drain_weights)
+            + match drain_weights {
+                Some(_) => {
+                    let selected_value = cs.selected_value();
+                    assert!(selected_value >= self.target.value);
+                    (cs.selected_value() - self.target.value) as f32
```
LLFourn, line 41: what's going on here? When you do have a change output why do you add this extra input minus target to the value :confused:.

darosior:
Another API consideration: for RBF the minimum absolute fee of a replacement transaction depends on the size of itself (since by rule 4 the fee must be at least the sum of the replaced transactions' fees plus 1sat/vb worth of the replacement transaction's size).

Therefore when selecting coins for the replacement transaction the caller can't set `min_fee` to the right value, since it depends on the result of the coin selection it's trying to get. For now i think our best bet is to try again with `min_fee = prev_min_fee + replacement_tx_size` until it either errors or succeeds, but it's meh.

h/t @jp1ac4.

LLFourn:
> Another API consideration: for RBF the minimum absolute fee of a replacement transaction depends on the size of itself (since by rule 4 the fee must be at least the sum of the replaced transactions' fees plus 1sat/vb worth of the replacement transaction's size).
> 
> Therefore when selecting coins for the replacement transaction the caller can't set `min_fee` to the right value, since it depends on the result of the coin selection it's trying to get. For now i think our best bet is to try again with `min_fee = prev_min_fee + replacement_tx_size` until it either errors or succeeds, but it's meh.
> 
> h/t @jp1ac4.

The `min_fee` is for rule 3. You use the previous feerate + 1 s/vb to satisfy rule 4. At least this is what i had in mind let me know if I missed something!

darosior:
> The `min_fee` is for rule 3. You use the previous feerate + 1 s/vb to satisfy rule 4. At least this is what i had in mind let me know if I missed something!

Unfortunately it's not that simple. [Rule 4](https://github.com/bitcoin/bitcoin/blob/master/doc/policy/mempool-replacements.md#current-replace-by-fee-policy) is about absolute fees. Consider this situation (`Tx (fee, vsize)`). We try replacing the following transaction chain: `A (100, 100)` <- `B (100, 100)`. We set `min_fee` to `200` and `feerate` to 2 s/vb (ie previous feerate + 1 s/vb). We get a replacement  `A' (200, 100)`.

This fails [the Rule 4 check](https://github.com/bitcoin/bitcoin/blob/d752349029ec7a76f1fd440db2ec2e458d0f3c99/src/policy/rbf.cpp#L172) since `A'` should not only have a higher feerate than conflicts (6) and a higher fee than the sum of the conflicts' fees (3), but it should also have 1 s/vb * its size additional fees (4). `A'` should have a fee of `300`.

Of course in this trivial example we could have set `min_fee` to `300`. But what if the coin selection makes it so `A'` is `200` vbytes large instead? Failure again. It's really something which can't be done outside the coin selection.

LLFourn:
@darosior thanks for the correction. I always simplified this rule to "higher feerate" in my mind. TIL. I don't blame myself too much since this is a very bizarre rule. It obviates the need for rule 3 (satisfying rule 4 will always satisfy rule 3). I do believe that higher abs fee + 1svb higher feerate than tx being replaced is far more intuitive and simpler.

I feel as if there was some discussion about simplifying these rules led by @glozow. I wonder if there was any progress. 

So if I understand correctly we should add another constraint to `Target` that handles RBF constraints which handles the abs fee of what is being replaced and has a "inremental relay" feerate. I think we should do this in a follow up PR @evanlinjin.

Once @evanlinjin responds to my comments above we're going to:

1. Move this crate to a new repo because it is so nicely decoupled from everything else
2. Release it
3. Make this PR just update the examples and use the released version (luckily we didn't impl RBF in the examples yet!)

darosior:
> I do believe that higher abs fee + 1svb higher feerate than tx being replaced is far more intuitive and simpler.

And doesn't achieve the purpose of this rule, which is to prevent free relay/churn. For instance with your rule i can replace `txA(100_000,100_000)` hundreds (almost a thousand) of times without paying any more fee by every time crafting a smaller `txA'(100_000,50_000)`, `txA'(100_000,33_333)`, etc.. replacement.

> I feel as if there was some discussion about simplifying these rules led by @glozow. I wonder if there was any progress.

There's been plenty of discussions around improving RBF, but it was mostly in the context of solving pinning (transaction v3 and ephemeral anchors). There's also been work into trying to patch some holes in the current RBF and make it more incentive compatible along with other mempool operations (for instance, correctly considering the ancestor feerate) but it turns out it's very complicated to do without a whole rewrite of the mempool. This culminated in the [Cluster Mempool](https://github.com/bitcoin/bitcoin/issues/27677) proposal, which should make for a more approachable interface for wallets.

TL;DR: there is many issues with RBF. People are aware. It's hard. :)

LLFourn:
> > I do believe that higher abs fee + 1svb higher feerate than tx being replaced is far more intuitive and simpler.
> 
> And doesn't achieve the purpose of this rule, which is to prevent free relay/churn. For instance with your rule i can replace `txA(100_000,100_000)` hundreds (almost a thousand) of times without paying any more fee by every time crafting a smaller `txA'(100_000,50_000)`, `txA'(100_000,33_333)`, etc.. replacement.

My claim was that transaction fees (abs fee, and feerate) should just be enforced to be monotonically increasing which is contradiction to "I can replace hundreds of times without paying any more fee". Your response sounds like you thought I was suggesting that enforcing only a greater feerate was sufficient. If my understanding is wrong please make another attempt to clean it up!

> There's been plenty of discussions around improving RBF, but it was mostly in the context of solving pinning (transaction v3 and ephemeral anchors). There's also been work into trying to patch some holes in the current RBF and make it more incentive compatible along with other mempool operations (for instance, correctly considering the ancestor feerate) but it turns out it's very complicated to do without a whole rewrite of the mempool. This culminated in the [Cluster Mempool](https://github.com/bitcoin/bitcoin/issues/27677) proposal, which should make for a more approachable interface for wallets.
> 
> TL;DR: there is many issues with RBF. People are aware. It's hard. :)

Thanks the cluster mempool work looks like it could be a great starting point to go deeper into mempool theory.

nursery/coin_select/src/metrics/lowest_fee.rs
jp1ac4: Should this rather be in the `None` case for the excess? I've just encountered one scenario where I get the same score for two different changeless selections, but one of these selections has greater excess.

darosior:
*(I'm happy to move this discussion elsewhere btw, it looks like we are slowly shifting OT of the code review here)*

---

> > > I do believe that higher abs fee + 1svb higher feerate than tx being replaced is far more intuitive and simpler.
> > 
> > 
> > And doesn't achieve the purpose of this rule, which is to prevent free relay/churn. For instance with your rule i can replace `txA(100_000,100_000)` hundreds (almost a thousand) of times without paying any more fee by every time crafting a smaller `txA'(100_000,50_000)`, `txA'(100_000,33_333)`, etc.. replacement.
> 
> My claim was that transaction fees (abs fee, and feerate) should just be enforced to be monotonically increasing which is contradiction to "I can replace hundreds of times without paying any more fee".

My bad then i indeed misunderstood "higher abs fee + 1svb" for "higher or equal abs fee + 1svb". But i'm still confused because it was in the context of your criticism of rule 4. Are you saying as long as it's paying as small as 1 satoshi more in abs fees we should accept it? That's not free, but still pretty cheap churn+relay.  And that's [precisely the rationale given for rule 4](https://github.com/bitcoin/bitcoin/blob/master/doc/policy/mempool-replacements.md#current-replace-by-fee-policy):
> Try to prevent DoS attacks where an attacker causes the network to repeatedly relay transactions each paying a tiny additional amount in fees, e.g. just 1 satoshi.

LLFourn:
> _(I'm happy to move this discussion elsewhere btw, it looks like we are slowly shifting OT of the code review here)_

All good this discussion will be useful when we come to fix this.

> 
> My bad then i indeed misunderstood "higher abs fee + 1svb" for "higher or equal abs fee + 1svb". But i'm still confused because it was in the context of your criticism of rule 4. Are you saying as long as it's paying as small as 1 satoshi more in abs fees we should accept it? That's not free, but still pretty cheap churn+relay. And that's [precisely the rationale given for rule 4](https://github.com/bitcoin/bitcoin/blob/master/doc/policy/mempool-replacements.md#current-replace-by-fee-policy):

No my bad looking back at it I used "+" to mean logical conjunction. Whoops.

Right I am suggesting that 1 satoshi more in abs fee is good enough for a small tx replacing a bigger tx (since that's the scenario where the abs fee would be stricter than the feerate rule). Or if 1 sat was not a big enough step up have a minimum like you do for feerate e.g. 100 sats.

After thinking about this my misunderstanding is that this would be simpler -- but actually it's not. Providing the previous tx fee and the prev feerate is actually equivalent to providing the previous transaction fee and weight. I don't know how I got stuck on this. I am still confused as to why there is rule 3 when rule 4 suffices for everything.

~So I propose we do something like this for the target:~

```rust
enum TargetFee {
    Rate(FeeRate),
    Replacement {
        prev_fee: u64,
        prev_weight: u32,
        incremental_relay_feerate: FeeRate,
    }
}
```

No this doesn't make sense. It should be:

```rust
struct TargetFee {
    feerate: FeeRate,
    replacement: Option<Replacement>,
}

struct Replacement {
    prev_fee: u64,
    prev_weight: u32,
    incremental_relay_feerate: FeeRate,
}
```

Feerate is orthogonal to whether we are replacing. It's two different constraints. So I'd say get rid of `min_fee` and do this.

evanlinjin:
`bdk_coin_select` now exists in a [separate repository](https://github.com/bitcoindevkit/coin-select) so we can close this PR.

notmandatory:
@evanlinjin should we have an issue for `bdk` to switch to using this new crate?

PR #1075, opened 2023-08-09
release: bump version to 0.29.0

notmandatory:
Because of the esplora-client update to 0.5.0 the 0.28.1 release needs to be yanked and replaced with this 0.29.0 release.


notmandatory:
Decided to do this is a 0.28.2 release instead of 0.29.0

PR #1084, opened 2023-08-21
Enhance bdk chain structures

### LINKED ISSUE #1061: Further enhancements to `Update` and `ChangeSet` structures
**Describe the enhancement**  

As suggested by @LLFourn:

1. LocalUpdate should be moved to bdk Wallet. 
2. We shouldn't use LocalUpdate in electrum API (it doesn't need to be there IMO). 
3. Delete LocalChangeSet or move it to bdk Wallet. We should implement `Append` for tuples (A,B) where A: Append, B:Append and use this as our changeset in the examples instead of `LocalChangeSet`. i.e. we can just go type ChangeSet = (local_chain::ChangeSet, indexed_tx_graph::ChangeSet<...>)

vladimirfomene:
### Description

Fixes bitcoindevkit/bdk#1061 

### Changelog notice

- Move WalletUpdate to the wallet module
- Remove ForEachTxout trait completely
- Refactor ElectrumExt to not use WalletUpdate.

#### All Submissions:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -110,7 +110,7 @@ pub struct WalletUpdate<K, A> {
     /// Update for the [`LocalChain`].
     ///
     /// [`LocalChain`]: local_chain::LocalChain
-    pub chain: local_chain::Update,
+    pub chain: Option<local_chain::Update>,
 }
 
 impl<K, A> WalletUpdate<K, A> {
@@ -119,7 +119,7 @@ impl<K, A> WalletUpdate<K, A> {
         Self {
             last_active_indices: BTreeMap::new(),
             graph: TxGraph::default(),
-            chain: chain_update,
+            chain: Some(chain_update),
         }
     }
 }
@@ -1942,7 +1942,11 @@ impl<D> Wallet<D> {
     where
         D: PersistBackend<ChangeSet>,
     {
-        let mut changeset = ChangeSet::from(self.chain.apply_update(update.chain)?);
+        let mut changeset = match update.chain {
+            Some(chain_update) => ChangeSet::from(self.chain.apply_update(chain_update)?),
+            None => ChangeSet::default(),
+        };
+
         let (_, index_changeset) = self
             .indexed_graph
             .index
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -22,10 +22,10 @@ use alloc::{
 pub use bdk_chain::keychain::Balance;
 use bdk_chain::{
     indexed_tx_graph,
-    keychain::{KeychainTxOutIndex, WalletChangeSet},
+    keychain::{self, KeychainTxOutIndex},
     local_chain::{self, CannotConnectError, CheckPoint, CheckPointIter, LocalChain},
     tx_graph::{CanonicalTx, TxGraph},
-    Append, BlockId, ChainPosition, ConfirmationTime, ConfirmationTimeAnchor, FullTxOut,
+    Anchor, Append, BlockId, ChainPosition, ConfirmationTime, ConfirmationTimeAnchor, FullTxOut,
     IndexedTxGraph, Persist, PersistBackend,
 };
 use bitcoin::consensus::encode::serialize;
@@ -120,6 +120,62 @@ impl<K, A> WalletUpdate<K, A> {
             last_active_indices: BTreeMap::new(),
             graph: TxGraph::default(),
             chain: chain_update,
+        }
+    }
+}
+
+/// A structure that records the corresponding changes as result of applying an [`WalletUpdate`].
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct WalletChangeSet<K, A> {
+    /// Changes to the [`LocalChain`].
+    ///
+    /// [`LocalChain`]: local_chain::LocalChain
+    pub chain: local_chain::ChangeSet,
+
+    /// ChangeSet to [`IndexedTxGraph`].
+    ///
+    /// [`IndexedTxGraph`]: bdk_chain::indexed_tx_graph::IndexedTxGraph
+    #[serde(bound(
+        deserialize = "K: Ord + serde::Deserialize<'de>, A: Ord + serde::Deserialize<'de>",
+        serialize = "K: Ord + serde::Serialize, A: Ord + serde::Serialize",
+    ))]
+    pub index_tx_graph: indexed_tx_graph::ChangeSet<A, keychain::ChangeSet<K>>,
+}
+
+impl<K, A> Default for WalletChangeSet<K, A> {
+    fn default() -> Self {
+        Self {
+            chain: Default::default(),
+            index_tx_graph: Default::default(),
+        }
+    }
+}
+
+impl<K: Ord, A: Anchor> Append for WalletChangeSet<K, A> {
+    fn append(&mut self, other: Self) {
+        Append::append(&mut self.chain, other.chain);
+        Append::append(&mut self.index_tx_graph, other.index_tx_graph);
+    }
+
+    fn is_empty(&self) -> bool {
+        self.chain.is_empty() && self.index_tx_graph.is_empty()
+    }
+}
+
+impl<K, A> From<local_chain::ChangeSet> for WalletChangeSet<K, A> {
+    fn from(chain: local_chain::ChangeSet) -> Self {
+        Self {
+            chain,
+            ..Default::default()
+        }
+    }
+}
+
+impl<K, A> From<indexed_tx_graph::ChangeSet<A, keychain::ChangeSet<K>>> for WalletChangeSet<K, A> {
+    fn from(index_tx_graph: indexed_tx_graph::ChangeSet<A, keychain::ChangeSet<K>>) -> Self {
+        Self {
+            index_tx_graph,
+            ..Default::default()
         }
     }
 }
@@ -277,7 +333,7 @@ impl<D> Wallet<D> {
 
         let changeset = db.load_from_persistence().map_err(NewError::Persist)?;
         chain.apply_changeset(&changeset.chain);
-        indexed_graph.apply_changeset(changeset.indexed_tx_graph);
+        indexed_graph.apply_changeset(changeset.index_tx_graph);
 
         let persist = Persist::new(db);
 
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -22,7 +22,7 @@ use alloc::{
 pub use bdk_chain::keychain::Balance;
 use bdk_chain::{
     indexed_tx_graph,
-    keychain::{KeychainTxOutIndex, WalletChangeSet, WalletUpdate},
+    keychain::{KeychainTxOutIndex, WalletChangeSet},
     local_chain::{self, CannotConnectError, CheckPoint, CheckPointIter, LocalChain},
     tx_graph::{CanonicalTx, TxGraph},
     Append, BlockId, ChainPosition, ConfirmationTime, ConfirmationTimeAnchor, FullTxOut,
@@ -93,6 +93,35 @@ pub struct Wallet<D = ()> {
     persist: Persist<D, ChangeSet>,
     network: Network,
     secp: SecpCtx,
+}
+
+/// A structure to update [`KeychainTxOutIndex`], [`TxGraph`] and [`LocalChain`] atomically.
+///
+/// [`LocalChain`]: local_chain::LocalChain
+#[derive(Debug, Clone)]
+pub struct WalletUpdate<K, A> {
+    /// Contains the last active derivation indices per keychain (`K`), which is used to update the
+    /// [`KeychainTxOutIndex`].
+    pub last_active_indices: BTreeMap<K, u32>,
+
+    /// Update for the [`TxGraph`].
+    pub graph: TxGraph<A>,
+
+    /// Update for the [`LocalChain`].
+    ///
+    /// [`LocalChain`]: local_chain::LocalChain
+    pub chain: local_chain::Update,
+}
+
+impl<K, A> WalletUpdate<K, A> {
+    /// Construct a [`WalletUpdate`] with a given [`local_chain::Update`].
+    pub fn new(chain_update: local_chain::Update) -> Self {
+        Self {
+            last_active_indices: BTreeMap::new(),
+            graph: TxGraph::default(),
+            chain: chain_update,
+        }
+    }
 }
 
 /// The update to a [`Wallet`] used in [`Wallet::apply_update`]. This is usually returned from blockchain data sources.
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -25,7 +25,7 @@ use bdk_chain::{
     keychain::{self, KeychainTxOutIndex},
     local_chain::{self, CannotConnectError, CheckPoint, CheckPointIter, LocalChain},
     tx_graph::{CanonicalTx, TxGraph},
-    Anchor, Append, BlockId, ChainPosition, ConfirmationTime, ConfirmationTimeAnchor, FullTxOut,
+    Append, BlockId, ChainPosition, ConfirmationTime, ConfirmationTimeAnchor, FullTxOut,
     IndexedTxGraph, Persist, PersistBackend,
 };
 use bitcoin::consensus::encode::serialize;
@@ -95,73 +95,51 @@ pub struct Wallet<D = ()> {
     secp: SecpCtx,
 }
 
-/// A structure to update [`Wallet`].
+/// An update to [`Wallet`].
 ///
 /// It updates [`bdk_chain::keychain::KeychainTxOutIndex`], [`bdk_chain::TxGraph`] and [`local_chain::LocalChain`] atomically.
-#[derive(Debug, Clone)]
-pub struct WalletUpdate<K, A> {
+#[derive(Debug, Clone, Default)]
+pub struct Update {
     /// Contains the last active derivation indices per keychain (`K`), which is used to update the
     /// [`KeychainTxOutIndex`].
-    pub last_active_indices: BTreeMap<K, u32>,
+    pub last_active_indices: BTreeMap<KeychainKind, u32>,
 
-    /// Update for the [`TxGraph`].
-    pub graph: TxGraph<A>,
+    /// Update for the wallet's internal [`TxGraph`].
+    pub graph: TxGraph<ConfirmationTimeAnchor>,
 
-    /// Update for the [`LocalChain`].
+    /// Update for the wallet's internal [`LocalChain`].
     ///
     /// [`LocalChain`]: local_chain::LocalChain
     pub chain: Option<local_chain::Update>,
 }
 
-impl<K, A> Default for WalletUpdate<K, A> {
-    fn default() -> Self {
-        Self {
-            last_active_indices: BTreeMap::new(),
-            graph: TxGraph::default(),
-            chain: None,
-        }
-    }
-}
-
-/// A structure that records the corresponding changes as result of applying an [`WalletUpdate`].
-#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
-pub struct WalletChangeSet<K, A> {
+/// The changes made to a wallet by applying an [`Update`].
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize, Default)]
+pub struct ChangeSet {
     /// Changes to the [`LocalChain`].
     ///
     /// [`LocalChain`]: local_chain::LocalChain
     pub chain: local_chain::ChangeSet,
 
-    /// ChangeSet to [`IndexedTxGraph`].
+    /// Changes to [`IndexedTxGraph`].
     ///
     /// [`IndexedTxGraph`]: bdk_chain::indexed_tx_graph::IndexedTxGraph
-    #[serde(bound(
-        deserialize = "K: Ord + serde::Deserialize<'de>, A: Ord + serde::Deserialize<'de>",
-        serialize = "K: Ord + serde::Serialize, A: Ord + serde::Serialize",
-    ))]
-    pub index_tx_graph: indexed_tx_graph::ChangeSet<A, keychain::ChangeSet<K>>,
+    pub indexed_tx_graph:
+        indexed_tx_graph::ChangeSet<ConfirmationTimeAnchor, keychain::ChangeSet<KeychainKind>>,
 }
 
-impl<K, A> Default for WalletChangeSet<K, A> {
-    fn default() -> Self {
-        Self {
-            chain: Default::default(),
-            index_tx_graph: Default::default(),
-        }
-    }
-}
-
-impl<K: Ord, A: Anchor> Append for WalletChangeSet<K, A> {
+impl Append for ChangeSet {
     fn append(&mut self, other: Self) {
         Append::append(&mut self.chain, other.chain);
-        Append::append(&mut self.index_tx_graph, other.index_tx_graph);
+        Append::append(&mut self.indexed_tx_graph, other.indexed_tx_graph);
     }
 
     fn is_empty(&self) -> bool {
-        self.chain.is_empty() && self.index_tx_graph.is_empty()
+        self.chain.is_empty() && self.indexed_tx_graph.is_empty()
     }
 }
 
-impl<K, A> From<local_chain::ChangeSet> for WalletChangeSet<K, A> {
+impl From<local_chain::ChangeSet> for ChangeSet {
     fn from(chain: local_chain::ChangeSet) -> Self {
         Self {
             chain,
@@ -170,20 +148,21 @@ impl<K, A> From<local_chain::ChangeSet> for WalletChangeSet<K, A> {
     }
 }
 
-impl<K, A> From<indexed_tx_graph::ChangeSet<A, keychain::ChangeSet<K>>> for WalletChangeSet<K, A> {
-    fn from(index_tx_graph: indexed_tx_graph::ChangeSet<A, keychain::ChangeSet<K>>) -> Self {
+impl From<indexed_tx_graph::ChangeSet<ConfirmationTimeAnchor, keychain::ChangeSet<KeychainKind>>>
+    for ChangeSet
+{
+    fn from(
+        indexed_tx_graph: indexed_tx_graph::ChangeSet<
+            ConfirmationTimeAnchor,
+            keychain::ChangeSet<KeychainKind>,
+        >,
+    ) -> Self {
         Self {
-            index_tx_graph,
+            indexed_tx_graph,
             ..Default::default()
         }
     }
 }
-
-/// The update to a [`Wallet`] used in [`Wallet::apply_update`]. This is usually returned from blockchain data sources.
-pub type Update = WalletUpdate<KeychainKind, ConfirmationTimeAnchor>;
-
-/// The changeset produced internally by [`Wallet`] when mutated.
-pub type ChangeSet = WalletChangeSet<KeychainKind, ConfirmationTimeAnchor>;
 
 /// The address index selection strategy to use to derived an address from the wallet's external
 /// descriptor. See [`Wallet::get_address`]. If you're unsure which one to use use `WalletIndex::New`.
@@ -332,7 +311,7 @@ impl<D> Wallet<D> {
 
         let changeset = db.load_from_persistence().map_err(NewError::Persist)?;
         chain.apply_changeset(&changeset.chain);
-        indexed_graph.apply_changeset(changeset.index_tx_graph);
+        indexed_graph.apply_changeset(changeset.indexed_tx_graph);
 
         let persist = Persist::new(db);
 
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -95,9 +95,9 @@ pub struct Wallet<D = ()> {
     secp: SecpCtx,
 }
 
-/// A structure to update [`KeychainTxOutIndex`], [`TxGraph`] and [`LocalChain`] atomically.
+/// A structure to update [`Wallet`].
 ///
-/// [`LocalChain`]: local_chain::LocalChain
+/// It updates [`bdk_chain::keychain::KeychainTxOutIndex`], [`bdk_chain::TxGraph`] and [`local_chain::LocalChain`] atomically.
 #[derive(Debug, Clone)]
 pub struct WalletUpdate<K, A> {
     /// Contains the last active derivation indices per keychain (`K`), which is used to update the
@@ -113,13 +113,12 @@ pub struct WalletUpdate<K, A> {
     pub chain: Option<local_chain::Update>,
 }
 
-impl<K, A> WalletUpdate<K, A> {
-    /// Construct a [`WalletUpdate`] with a given [`local_chain::Update`].
-    pub fn new(chain_update: local_chain::Update) -> Self {
+impl<K, A> Default for WalletUpdate<K, A> {
+    fn default() -> Self {
         Self {
             last_active_indices: BTreeMap::new(),
             graph: TxGraph::default(),
-            chain: Some(chain_update),
+            chain: None,
         }
     }
 }
--- a/crates/chain/src/indexed_tx_graph.rs
+++ b/crates/chain/src/indexed_tx_graph.rs
@@ -233,7 +233,18 @@ pub trait Indexer {
     /// Scan and index the given `outpoint` and `txout`.
     fn index_txout(&mut self, outpoint: OutPoint, txout: &TxOut) -> Self::ChangeSet;
 
-    /// Scan and index the given transaction.
+    /// Scans a transaction for relevant outpoints, which are stored and indexed internally.
+    ///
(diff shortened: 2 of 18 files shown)


LLFourn:
What's the motivation for removing `ForEachTxOut`? I think it serves a purpose in that it allows you to scan several things that are bunches of txouts (including things that are not transactions).

evanlinjin:
> What's the motivation for removing `ForEachTxOut`? I think it serves a purpose in that it allows you to scan several things that are bunches of txouts (including things that are not transactions).

This was my suggestion and I wrote it on the ticket. I don't think it is needed? The `Indexer` trait has the method `Indexer::index_txout(&mut self, outpoint: OutPoint, txout: &TxOut) -> Self::ChangeSet`, and we can have another method:

```rust
pub trait Indexer {
    /* ... */

    fn index_txouts<'a>(
        &mut self,
        txouts: impl IntoIterator<Item = (OutPoint, &'a TxOut)>,
    ) -> Self::ChangeSet
    where
        Self::ChangeSet: Append + Default,
    {
        let mut changeset = Self::ChangeSet::default();
        for (outpoint, txout) in txouts {
            Append::append(&mut changeset, self.index_txout(outpoint, txout));
        }
        changeset
    }
}
```

Now anything that implements `IntoIterator<Item = (OutPoint, &TxOut)>` will work.

vladimirfomene:
Will rebase this once MSRV fixes make it to master.

crates/bdk/src/wallet/mod.rs
```diff
@@ -94,6 +94,91 @@ pub struct Wallet<D = ()> {
...
+    /// [`KeychainTxOutIndex`].
+    pub last_active_indices: BTreeMap<K, u32>,
+
+    /// Update for the [`TxGraph`].
+    pub graph: TxGraph<A>,
+
+    /// Update for the [`LocalChain`].
+    ///
+    /// [`LocalChain`]: local_chain::LocalChain
+    pub chain: local_chain::Update,
```
evanlinjin: @vladimirfomene Another request, can you try wrap this with an `Option` please?

I think it'll be important to not have to update with a `CheckPoint`.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -124,13 +124,11 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-    /// [`ForEachTxout`]: crate::ForEachTxOut
-    pub fn scan(&mut self, txouts: &impl ForEachTxOut) -> super::ChangeSet<K> {
+    pub fn scan(&mut self, tx: &bitcoin::Transaction) -> super::ChangeSet<K> {
         let mut changeset = super::ChangeSet::<K>::default();
-        txouts.for_each_txout(|(op, txout)| changeset.append(self.scan_txout(op, txout)));
+        for (op, txout) in tx.output.iter().enumerate() {
+            changeset.append(self.scan_txout(OutPoint::new(tx.txid(), op as u32), txout));
+        }
         changeset
     }
```
evanlinjin: I think we can move `scan()` to the `Indexer::index_tx()` implementation of `KeychainTxOutIndex`. `scan_txout` can also be replaced by `Indexer::index_txout`.

I would also do the same for `SpkTxoutIndex`.

evanlinjin (changes requested):

crates/bdk/src/wallet/mod.rs
```diff
@@ -94,6 +94,91 @@ pub struct Wallet<D = ()> {
     secp: SecpCtx,
 }
 
+/// A structure to update [`KeychainTxOutIndex`], [`TxGraph`] and [`LocalChain`] atomically.
```
evanlinjin: We will need to change this to something more understandable. This is no longer the `bdk_chain` crate.

```rust
/// A structure to update [`Wallet`].
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -94,6 +94,91 @@ pub struct Wallet<D = ()> {
...
+
+impl<K, A> WalletUpdate<K, A> {
+    /// Construct a [`WalletUpdate`] with a given [`local_chain::Update`].
+    pub fn new(chain_update: local_chain::Update) -> Self {
+        Self {
+            last_active_indices: BTreeMap::new(),
+            graph: TxGraph::default(),
+            chain: Some(chain_update),
+        }
+    }
```
evanlinjin: Since the `chain` field is now an `Option`, we can implement `Default` for `WalletUpdate` instead of this method.

Don't use `derive()`, otherwise we will need to have `K` and `A` implement `Default` for `WalletUpdate` to be `Default`.

crates/electrum/src/electrum_ext.rs
```diff
@@ -539,9 +526,8 @@ fn populate_with_spks<K, I: Ord + Clone>(
             }
 
             for tx in spk_history {
-                let tx_entry = update.graph_update.entry(tx.tx_hash).or_default();
```
evanlinjin, line 542: I think you have accidentally removed the ability for us to fetch unconfirmed/mempool transactions.

Notice the `or_default()` call here.

crates/electrum/src/electrum_ext.rs
```diff
@@ -1,29 +1,28 @@
...
+/// Represents an update fetched from an Electrum server.
 ///
-/// To provide a complete update to [`TxGraph`], you'll need to call [`Self::missing_full_txs`] to
+/// To get a complete [`Self::graph_update`], you'll need to call [`Self::missing_full_txs`] to
 /// determine the full transactions missing from [`TxGraph`]. Then call [`Self::finalize`] to fetch
 /// the full transactions from Electrum and finalize the update.
 #[derive(Debug, Clone)]
 pub struct ElectrumUpdate<K, A> {
-    /// Map of [`Txid`]s to associated [`Anchor`]s.
-    pub graph_update: HashMap<Txid, BTreeSet<A>>,
```
evanlinjin, line 26: What is the motivation of this change?

evanlinjin (changes requested):
Thanks for the work so far. Sorry about my incremental reviews.

crates/bdk/src/wallet/mod.rs
vladimirfomene: I'm thinking we can have something like:
```
/// A structure to update [`Wallet`].

/// It updates [`bdk_chain::KeychainTxoutIndex`], [`bdk_chain::TxGraph`] and [`bdk_chain::LocalChain`] atomically.
```

crates/bdk/src/wallet/mod.rs
vladimirfomene: Let me do this

LLFourn:
> > What's the motivation for removing `ForEachTxOut`? I think it serves a purpose in that it allows you to scan several things that are bunches of txouts (including things that are not transactions).
> 
> This was my suggestion and I wrote it on the ticket. I don't think it is needed? The `Indexer` trait has the method `Indexer::index_txout(&mut self, outpoint: OutPoint, txout: &TxOut) -> Self::ChangeSet`, and we can have another method:

So if I want to index a `Block` what should I do?. I can iterate manually over every `TxOut` and call this but that seems much more cumbersome than it needs to be?

crates/electrum/src/electrum_ext.rs
vladimirfomene: The motivation was to get rid of `WalletUpdate` but still have a way to get all the graph updates that we were exposing through `WalletUpdate`.

vladimirfomene:
@LLFourn I'm thinking you will use `index_tx` here instead of `index_txout`.

crates/electrum/src/electrum_ext.rs
```diff
@@ -118,7 +108,7 @@ impl<K> ElectrumUpdate<K, ConfirmationHeightAnchor> {
             .collect::<HashMap<u32, u64>>();
 
         let graph_changeset = {
-            let old_changeset = TxGraph::default().apply_update(update.graph.clone());
+            let old_changeset = TxGraph::default().apply_update(self.graph_update);
```
LLFourn: Not part of this PR necessarily but the code below seems super complicated and uses changesets for no reason? How about we just have a function on `tx_graph.map_anchor(|anchor| get_new_anchor(anchor))` which consumes self and returns a new `TxGraph` with the new anchor.

crates/electrum/src/electrum_ext.rs
LLFourn: The main motivation for removing `WalletUpdate` from here is so `keychain_update` doesn't have to be part of this struct since it is not always necessary to have it. The one that does the keychain scan should just return a tuple with the keychain update being an element. This removes a type parameter. The `new_tip` also seems orthogonal -- it can just be returned as another element in the tuple. So `ElectrumUpdate` should only have the graph info in it. 

The particular change made here seems a bit dangerous since we can't enforce the user doesn't use the `TxGraph` as an update before it's finalized. If we do what i suggest above I think it's fine to leave this as a `HashMap` and then consume self in the `finalize` methods.

crates/electrum/src/electrum_ext.rs
```diff
@@ -54,29 +55,19 @@ impl<K, A: Anchor> ElectrumUpdate<K, A> {
...
         let new_txs = client.batch_transaction_get(&missing)?;
-        let mut graph_update = TxGraph::<A>::new(new_txs);
-        for (txid, anchors) in self.graph_update {
+        for tx in new_txs {
             if let Some(seen_at) = seen_at {
-                let _ = graph_update.insert_seen_at(txid, seen_at);
-            }
-            for anchor in anchors {
-                let _ = graph_update.insert_anchor(txid, anchor);
+                let _ = self.graph_update.insert_seen_at(tx.txid(), seen_at);
```
LLFourn: I think this is wrong but not due to this PR. We should only set `seen_at` if the transaction is not anchored. Right @evanlinjin?

crates/chain/src/keychain/txout_index.rs
LLFourn: I think this is a good idea. Also `index_tx` can have a default impl in the trait.

LLFourn:
> @LLFourn I'm thinking you will use `index_tx` here instead of `index_txout`.

Understood. I think might be as easy as we need to make it. Thanks.

crates/electrum/src/electrum_ext.rs
evanlinjin: I think it's okay to have both. The transaction may get reorged out and the last_seen will become useful

crates/electrum/src/electrum_ext.rs
```diff
@@ -58,7 +58,7 @@ impl<K, A: Anchor> ElectrumUpdate<K, A> {
         client: &Client,
         seen_at: Option<u64>,
         missing: Vec<Txid>,
-    ) -> Result<WalletUpdate<K, A>, Error> {
+    ) -> Result<(TxGraph<A>, BTreeMap<K, u32>, local_chain::CheckPoint), Error> {
```
evanlinjin: Please return a `local_chain::Update` instead of `local_chain::CheckPoint`.

evanlinjin (changes requested):

crates/electrum/src/electrum_ext.rs
evanlinjin: Created issue: #1096

crates/chain/src/keychain/txout_index.rs
evanlinjin: Created ticket: #1097

evanlinjin (approved):
ACK fb8777a9c3ecfec0aeca35dfbd773e51b0cf8e56

vladimirfomene:
Thanks @evanlinjin for these improvements!

notmandatory:
Per Evan moved this to Alpha.2 since it's required for RPC PR

notmandatory (approved):
ACK c56728ff1315e0deaf256af07fd1ff5e18fced8a

I confirmed `example_electrum` and `example_esplora` scan and sync commands still work. Used my test signet descriptors:

```shell
cargo run -- --network signet "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/0/*)" "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/1/*)" scan
```

notmandatory:
One small nit, it looks like most of the commits include breaking changes so commit messages should include a `!` to denote commit breaks old APIs.  Doesn't need to be fixed for this PR but to keep in mind for future.

evanlinjin:
ACK c56728ff1315e0deaf256af07fd1ff5e18fced8a

crates/bdk/src/wallet/mod.rs
```diff
@@ -95,6 +95,90 @@ pub struct Wallet<D = ()> {
     secp: SecpCtx,
 }
 
+/// A structure to update [`Wallet`].
```
```suggestion
/// An update to [`Wallet`].
``` 

There is no need to refer to it as a structure. Anyone can see it is a `struct`.

crates/bdk/src/wallet/mod.rs
```diff
@@ -95,6 +95,90 @@ pub struct Wallet<D = ()> {
...
+///
+/// It updates [`bdk_chain::keychain::KeychainTxOutIndex`], [`bdk_chain::TxGraph`] and [`local_chain::LocalChain`] atomically.
+#[derive(Debug, Clone)]
+pub struct WalletUpdate<K, A> {
+    /// Contains the last active derivation indices per keychain (`K`), which is used to update the
+    /// [`KeychainTxOutIndex`].
+    pub last_active_indices: BTreeMap<K, u32>,
+
+    /// Update for the [`TxGraph`].
```
```suggestion
    /// Update for the wallet's internal [`TxGraph`].
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -95,6 +95,90 @@ pub struct Wallet<D = ()> {
...
+    pub graph: TxGraph<A>,
+
+    /// Update for the [`LocalChain`].
```
```suggestion
    /// Update for the wallet's internal [`LocalChain`].
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -95,6 +95,90 @@ pub struct Wallet<D = ()> {
...
+    fn default() -> Self {
+        Self {
+            last_active_indices: BTreeMap::new(),
+            graph: TxGraph::default(),
+            chain: None,
+        }
+    }
+}
+
+/// A structure that records the corresponding changes as result of applying an [`WalletUpdate`].
```
```suggestion
/// The changes made to a wallet by applying an [`WalletUpdate`].
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -95,6 +95,90 @@ pub struct Wallet<D = ()> {
...
+    pub chain: local_chain::ChangeSet,
+
+    /// ChangeSet to [`IndexedTxGraph`].
+    ///
+    /// [`IndexedTxGraph`]: bdk_chain::indexed_tx_graph::IndexedTxGraph
+    #[serde(bound(
+        deserialize = "K: Ord + serde::Deserialize<'de>, A: Ord + serde::Deserialize<'de>",
+        serialize = "K: Ord + serde::Serialize, A: Ord + serde::Serialize",
+    ))]
+    pub index_tx_graph: indexed_tx_graph::ChangeSet<A, keychain::ChangeSet<K>>,
```
```suggestion
    pub indexed_tx_graph: indexed_tx_graph::ChangeSet<A, keychain::ChangeSet<K>>,
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -95,6 +95,90 @@ pub struct Wallet<D = ()> {
...
+    fn from(index_tx_graph: indexed_tx_graph::ChangeSet<A, keychain::ChangeSet<K>>) -> Self {
+        Self {
+            index_tx_graph,
+            ..Default::default()
+        }
+    }
+}
+
 /// The update to a [`Wallet`] used in [`Wallet::apply_update`]. This is usually returned from blockchain data sources.
 pub type Update = WalletUpdate<KeychainKind, ConfirmationTimeAnchor>;
```
LLFourn: Remove these type Alias since they are no longer needed.

crates/bdk/src/wallet/mod.rs
```diff
@@ -95,6 +95,90 @@ pub struct Wallet<D = ()> {
     secp: SecpCtx,
 }
 
+/// A structure to update [`Wallet`].
+///
+/// It updates [`bdk_chain::keychain::KeychainTxOutIndex`], [`bdk_chain::TxGraph`] and [`local_chain::LocalChain`] atomically.
+#[derive(Debug, Clone)]
+pub struct WalletUpdate<K, A> {
```
```suggestion
pub struct Update {
```

`K` and `A` are `KeychainKind` and `ConfirmationTimeAnchor` and are invariant at the moment.

crates/bdk/src/wallet/mod.rs
```diff
@@ -95,6 +95,90 @@ pub struct Wallet<D = ()> {
...
+            last_active_indices: BTreeMap::new(),
+            graph: TxGraph::default(),
+            chain: None,
+        }
+    }
+}
+
+/// A structure that records the corresponding changes as result of applying an [`WalletUpdate`].
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct WalletChangeSet<K, A> {
```
```suggestion
pub struct ChangeSet {
```

Same reasoning as above.

crates/bdk/src/wallet/mod.rs
LLFourn: I don't recall how this name change happened but it shouldn't have.

crates/chain/src/spk_txout_index.rs
```diff
@@ -54,19 +53,35 @@ impl<I> Default for SpkTxOutIndex<I> {
 }
 
 impl<I: Clone + Ord> Indexer for SpkTxOutIndex<I> {
-    type ChangeSet = ();
+    type ChangeSet = BTreeSet<I>;
```
LLFourn: This doesn't make sense to me. The changeset is meant to keep note of an index's persisted state. But `SpkTxOutIndex` has no persisted state and what is returned from `index_txout` doesn't represent a change to state at all. This shoulds stay as `()`.

crates/electrum/src/electrum_ext.rs
```diff
@@ -15,90 +14,67 @@ use std::{
...
-        Self {
-            new_tip,
-            graph_update: HashMap::new(),
-            keychain_update: BTreeMap::new(),
-        }
-    }
+/// determine the full transactions missing from [`TxGraph`]. Then call [`Self::finalize`] to
+/// fetch the full transactions from Electrum and finalize the update.
+#[derive(Debug, Default, Clone)]
+pub struct IncompleteTxGraph<A>(HashMap<Txid, BTreeSet<A>>);
```
LLFourn: Can we call this just `RelevantTxids` there's no need to conceptually couple this to a `TxGraph`.

Let me try and point out what seems overly complicated here. I know these problems were not introduced in this PR but i think now's the time to fix them.

1. The data you get from electrum is always anchored with `ConfirmationHeight` so why is there a type parameter here.
2. Why is there a type parameter on `ElectrumExt`. Extension traits don't have type parameters in general.
3. These "finalize" methods should just be `into_tx_graph` and `into_confirmation_time_tx_graph`

example-crates/example_electrum/src/main.rs
```diff
@@ -60,19 +60,22 @@ pub struct ScanOptions {
...
 }
 
-type ChangeSet = WalletChangeSet<Keychain, ConfirmationHeightAnchor>;
+type ChangeSet = (
+    local_chain::ChangeSet,
+    indexed_tx_graph::ChangeSet<ConfirmationHeightAnchor, keychain::ChangeSet<Keychain>>,
+);
 
 fn main() -> anyhow::Result<()> {
     let (args, keymap, index, db, init_changeset) =
```
LLFourn: :bulb: deconstruct `init_changeset` into `(disk_local_chain, disk_tx_graph)` to avoid using `.1` and `.0` below.

example-crates/wallet_electrum/src/main.rs
```diff
@@ -52,14 +53,21 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         })
         .collect();
 
-    let electrum_update = client.scan(prev_tip, keychain_spks, None, None, STOP_GAP, BATCH_SIZE)?;
+    let (chain_update, incomplete_graph_update, keychain_update) =
```
LLFourn:
```suggestion
    let (chain_update, relevant_txids, keychain_update) =
```

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -233,7 +233,18 @@ pub trait Indexer {
     /// Scan and index the given `outpoint` and `txout`.
     fn index_txout(&mut self, outpoint: OutPoint, txout: &TxOut) -> Self::ChangeSet;
 
-    /// Scan and index the given transaction.
+    /// Scans a transaction for relevant outpoints, which are stored and indexed internally.
+    ///
+    /// If the matched script pubkey is part of the lookahead, the last stored index is updated for
+    /// the script pubkey's keychain and the [`ChangeSet`] returned will reflect the
+    /// change.
```
LLFourn: lookahead has nothing to do with this trait. That's an API of `KeychainTxOutIndex`. I think this documentation is missing the point (probably because it was copy pasted) for this methid. The documentation of the trait should mention that it's primarily used by `IndexedTxGraph`. Then we can explain here that `index_tx` is called when there's a new transaction added to the graph.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -200,14 +176,11 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// Set the lookahead count for `keychain`.
     ///
     /// The lookahead is the number of scripts to cache ahead of the last stored script index. This
```
LLFourn:
```suggestion
    /// The lookahead is the number of scripts to cache ahead of the last revealed script index. This
```

crates/chain/src/keychain/txout_index.rs
```diff
@@ -200,14 +176,11 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-    /// is useful during a scan via [`scan`] or [`scan_txout`].
+    /// is useful during a scan via [`Indexer::index_tx`] or [`Indexer::index_txout`].
```
LLFourn: Since we are not necessarily encouraging people to "scan" things with these methods (it's mostly done automatically) I don't think this sentence is doing much for us. Instead explain when it is useful at a higher level i.e. when you are processing block data and want to find outputs you own beyond your last revealed index because the last revealed index may not be accurate (some other system may have revealed script pubkeys beyond it).

crates/chain/src/lib.rs
```diff
@@ -100,3 +100,11 @@ pub mod collections {
 
 /// How many confirmations are needed f or a coinbase output to be spent.
 pub const COINBASE_MATURITY: u32 = 100;
+
+impl<A, IA> From<indexed_tx_graph::ChangeSet<A, IA>>
```
LLFourn: Not sure if we need this. If we do it shouldn't be here (it seems to be about `indexed_tx_graph`)

crates/electrum/src/electrum_ext.rs
```diff
@@ -223,9 +204,8 @@ impl ElectrumExt<ConfirmationHeightAnchor> for Client {
 
         let update = loop {
             let (tip, _) = construct_update_tip(self, prev_tip.clone())?;
-            let mut update = ElectrumUpdate::<K, ConfirmationHeightAnchor>::new(tip.clone());
-            let cps = update
-                .new_tip
+            let mut graph_update = IncompleteTxGraph::<ConfirmationHeightAnchor>::default();
```
LLFourn: In general can we avoid the term "graph" in this code (unless we are actually returning a `TxGraph`). We are just finding relevant transactions to set of script pubkeys.

crates/electrum/src/electrum_ext.rs
```diff
@@ -165,6 +136,7 @@ pub trait ElectrumExt<A> {
     /// The scan for each keychain stops after a gap of `stop_gap` script pubkeys with no associated
     /// transactions. `batch_size` specifies the max number of script pubkeys to request for in a
     /// single batch request.
+    #[allow(clippy::type_complexity)]
```
LLFourn: Hmm I don't think we should do this. Can we just package these tuple types into their own types. Consider bundling (`local_chain::Update, RelevantTxids)` into its own type.

LLFourn (changes requested):
Hey @vladimirfomene thanks for doing this refactoring.

I've pointed out a bunch of ways we can simplify things in both wallet types and electrum_ext. Many of the changes requested are not to do with this PR but I think this is the place to do them.

[edit] I also pushed 41042069809e3eeb4a8cc8a5a8db1af2c57c4a11

crates/chain/src/keychain/txout_index.rs
```diff
@@ -120,6 +112,36 @@ impl<K: Clone + Ord + Debug> Indexer for KeychainTxOutIndex<K> {
...
+    /// the script pubkey's keychain and the [`super::ChangeSet`] returned will reflect the
+    /// change.
+    ///
+    /// Typically, this method is used in two situations:
+    ///
+    /// 1. After loading transaction data from the disk, you may scan over all the txouts to restore all
+    /// your txouts.
+    /// 2. When getting new data from the chain, you usually scan it before incorporating it into
+    /// your chain state.
+    pub fn scan(&mut self, tx: &bitcoin::Transaction) -> super::ChangeSet<K> {
```
evanlinjin: Why are we reintroducing `scan` and `scan_txout`? The `Indexer` methods already do the same thing.

Same question for `SpkTxOutIndex`.

vladimirfomene: I can see how to get rid of `scan` and `scan_txout` in `TxoutIndex` but I don't see how to get rid of it in `SpkTxoutIndex` as `scan_txout` in `TxoutIndex` depends on `scan_txout` in `SpkTxoutIndex` returning an optional index. Do you mind sharing some code sample of how this could work?

evanlinjin: Thanks for explaining, I see the problem now.

In `KeychainTxOutIndex`, the signature of the methods are exactly the same so we don't need to duplicate them (rm `scam` and `scan_txout`). For `SpkTxOutIndex`, they are different, so we need different methods.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -199,15 +174,10 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-    ///
-    /// # Panics
-    ///
-    /// This will panic if the `keychain` does not exist.
-    ///
-    /// [`scan`]: Self::scan
-    /// [`scan_txout`]: Self::scan_txout
+    /// The lookahead is the number of scripts to cache ahead of the last revealed script index. This
+    /// is useful to find outputs you own when processing block data that lie beyond the last revealed
+    /// index. This is because in certain scenarios we might not know the last revealed indexit in like
```
LLFourn: Thanks this explanation is better. There's a typo in this sentence. I think it just needs a small rewrite.

PS does this method still panic if the keychain doesn't exist? If so we should probably still mention that.

crates/electrum/src/electrum_ext.rs
```diff
@@ -140,21 +114,17 @@ impl<K> ElectrumUpdate<K, ConfirmationHeightAnchor> {
...
-            },
-            chain: update.chain,
-        })
+        let mut new_graph = TxGraph::default();
+        new_graph.apply_changeset(graph_changeset);
+        Ok(new_graph)
     }
 }
 
+type ElectrumUpdate = (local_chain::Update, RelevantTxids);
```
LLFourn: I think it would be better if this was a struct and explain why the update is in this form (we have to update the chain and the txids at the same time since we anchor the txids to the same chain tip that we check before and after we gather the txids).

LLFourn (approved):
This LGTM! ACK ee01b0fd632c1ff4ab95ec6a50e6af61a1eab7df

left a couple of small comments.

crates/chain/src/keychain/txout_index.rs
vladimirfomene: Let me check. If it still panics, I will mention it.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -225,15 +225,21 @@ impl<A, K> From<keychain::ChangeSet<K>> for ChangeSet<A, keychain::ChangeSet<K>>
...
 }
 
-/// Represents a structure that can index transaction data.
+impl<A, IA> From<ChangeSet<A, IA>> for (local_chain::ChangeSet, ChangeSet<A, IA>) {
+    fn from(indexed_changeset: ChangeSet<A, IA>) -> Self {
+        (local_chain::ChangeSet::default(), indexed_changeset)
+    }
+}
+
+/// Defines methods to index transaction data.
```
evanlinjin:
```suggestion
/// A trait for indexing transaction data.
///
/// Types which implement this trait can be used to construct an [`IndexedTxGraph`]. This trait
/// should rarely be called directly.
```

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -225,15 +225,21 @@ impl<A, K> From<keychain::ChangeSet<K>> for ChangeSet<A, keychain::ChangeSet<K>>
     }
 }
 
-/// Represents a structure that can index transaction data.
+impl<A, IA> From<ChangeSet<A, IA>> for (local_chain::ChangeSet, ChangeSet<A, IA>) {
+    fn from(indexed_changeset: ChangeSet<A, IA>) -> Self {
+        (local_chain::ChangeSet::default(), indexed_changeset)
+    }
+}
```
evanlinjin: This is rather odd. Where is this useful?

crates/chain/src/spk_txout_index.rs
```diff
@@ -77,49 +76,37 @@ impl<I: Clone + Ord> Indexer for SpkTxOutIndex<I> {
...
-            $self.spk_txouts.insert((spk_i.clone(), $op));
-            $self.unused.remove(&spk_i);
-        }
-        spk_i
-    }};
-}
-
 impl<I: Clone + Ord> SpkTxOutIndex<I> {
-    /// Scans an object containing many txouts.
+    /// Scans a transaction for txouts.
```
evanlinjin:
```suggestion
    /// Scans a transaction's outputs for matching script pubkeys.
```

crates/chain/src/spk_txout_index.rs
```diff
@@ -77,49 +76,37 @@ impl<I: Clone + Ord> Indexer for SpkTxOutIndex<I> {
...
-    ///
-    /// [`ForEachTxout`]: crate::ForEachTxOut
-    pub fn scan(&mut self, txouts: &impl ForEachTxOut) -> BTreeSet<I> {
+    pub fn scan(&mut self, tx: &Transaction) -> BTreeSet<I> {
         let mut scanned_indices = BTreeSet::new();
 
-        txouts.for_each_txout(|(op, txout)| {
-            if let Some(spk_i) = scan_txout!(self, op, txout) {
+        for (i, txout) in tx.output.iter().enumerate() {
+            let op = OutPoint::new(tx.txid(), i as u32);
```
evanlinjin: **Nit:** I think it would be better to create the `txid` outside the loop to avoid rehashing the transaction in every iteration.

evanlinjin (changes requested):
Mostly nits and documentation suggestions.

crates/chain/src/indexed_tx_graph.rs
vladimirfomene: It is useful in `example_cli::handle_commands`. If not implemented we get the following error, `the trait From<bdk_chain::indexed_tx_graph::ChangeSet<ConfirmationTimeAnchor, bdk_chain::keychain::ChangeSet<Keychain>>> is not implemented for (BTreeMap<u32, std::option::Option<BlockHash>>, bdk_chain::indexed_tx_graph::ChangeSet<ConfirmationTimeAnchor, bdk_chain::keychain::ChangeSet<Keychain>>)`

crates/chain/src/indexed_tx_graph.rs
vladimirfomene: What do you mean by `This trait should rarely be called directly`?

crates/chain/src/spk_txout_index.rs
vladimirfomene: I agree

crates/chain/src/indexed_tx_graph.rs
evanlinjin: The caller shouldn't be calling the trait methods directly. It should all be handled in `IndexedTxGraph`.

---

An example of stdlib documentation using similar wording: https://doc.rust-lang.org/stable/std/iter/trait.Product.html

crates/chain/src/indexed_tx_graph.rs
evanlinjin: @vladimirfomene surely there must be a way around this?

vladimirfomene: Let me look for another solution

evanlinjin: Thanks, let me know what you find.

evanlinjin (approved):
ACK 1ff806c67f4da9ba58b7c7689fde0fe41a34a6f5

PR #1093, opened 2023-08-24
fix: spks_of_all_keychains() shouldn't return an infinite iterator for non-wildcard descriptors

danielabrozzoni:
### Description

When you pass a non-wildcard descriptor in `new_with_range`, we make
sure that the range length is at most 1; if that's not the case, we
shorten it.
We would previously use `new_with_range` without this check and with a
non-wildcard descriptor in `spks_of_all_keychains`, this meant creating
a spkiterator that would go on producing the same spks over and over
again, causing some issues with syncing on electrum/esplora.

To reproduce the bug, run in `example-crates/example_electrum`:
```
cargo run "sh(wsh(or_d(c:pk_k(cPGudvRLDSgeV4hH9NUofLvYxYBSRjju3cpiXmBg9K8G9k1ikCMp),c:pk_k(cSBSBHRrzqSXFmrBhLkZMzQB9q4P9MnAq92v8d9a5UveBc9sLX32))))#zp9pcfs9" scan
```

### Changelog notice

- Fixed a bug where `KeychainTxOutIndex::spks_of_all_keychains`/`KeychainTxOutIndex::spks_of_keychain` would return an iterator yielding infinite spks even for non-wildcard descriptors.

#### All Submissions:

### DIFF
--- a/crates/chain/src/spk_iter.rs
+++ b/crates/chain/src/spk_iter.rs
@@ -45,34 +45,41 @@ where
 {
     /// Creates a new script pubkey iterator starting at 0 from a descriptor.
     pub fn new(descriptor: D) -> Self {
-        let end = if descriptor.borrow().has_wildcard() {
-            BIP32_MAX_INDEX
-        } else {
-            0
-        };
-
-        SpkIterator::new_with_range(descriptor, 0..=end)
+        SpkIterator::new_with_range(descriptor, 0..=BIP32_MAX_INDEX)
     }
 
     // Creates a new script pubkey iterator from a descriptor with a given range.
+    // If the descriptor doesn't have a wildcard, we shorten whichever range you pass in
+    // to have length <= 1. This means that if you pass in 0..0 or 0..1 the range will
+    // remain the same, but if you pass in 0..10, we'll shorten it to 0..1
     pub(crate) fn new_with_range<R>(descriptor: D, range: R) -> Self
     where
         R: RangeBounds<u32>,
     {
+        let start = match range.start_bound() {
+            Bound::Included(start) => *start,
+            Bound::Excluded(start) => *start + 1,
+            Bound::Unbounded => u32::MIN,
+        };
+
         let mut end = match range.end_bound() {
             Bound::Included(end) => *end + 1,
             Bound::Excluded(end) => *end,
             Bound::Unbounded => u32::MAX,
         };
+
         // Because `end` is exclusive, we want the maximum value to be BIP32_MAX_INDEX + 1.
         end = end.min(BIP32_MAX_INDEX + 1);
 
+        if !descriptor.borrow().has_wildcard() {
+            // The length of the range should be at most 1
+            if end != start {
+                end = start + 1;
+            }
+        }
+
         Self {
-            next_index: match range.start_bound() {
-                Bound::Included(start) => *start,
-                Bound::Excluded(start) => *start + 1,
-                Bound::Unbounded => u32::MIN,
-            },
+            next_index: start,
             end,
             descriptor,
             secp: Secp256k1::verification_only(),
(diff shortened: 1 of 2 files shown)


evanlinjin (changes requested):
Thank you for spotting and fixing this @danielabrozzoni!

I think `SpkIterator::new_with_range` should also detect non-wildcard descriptors, and the implementation of `SpkIterator::new` can be changed to this:

```rust
    /// Creates a new script pubkey iterator starting at 0 from a descriptor.
    pub fn new(descriptor: D) -> Self {
        SpkIterator::new_with_range(descriptor, 0..=BIP32_MAX_INDEX)
    }
```

This means, `SpkIterator:new_with_range` will sometimes return empty iterators.

LLFourn:
+1 for what @evanlinjin suggested above.

danielabrozzoni:
> I think SpkIterator::new_with_range should also detect non-wildcard descriptors, and the implementation of SpkIterator::new can be changed to this:

I tried to implement this by setting `end = next_index` in `new_with_range` for non-wildcard descriptors, but I'm seeing some tests failing, investigating now...

danielabrozzoni:
> I tried to implement this by setting end = next_index in new_with_range for non-wildcard descriptors, but I'm seeing some tests failing, investigating now...

Turns out that this was wrong. I'm now shortening whichever range passed to `new_with_range` to have length at most 1 if the descriptor doesn't have a wildcard, and it works perfectly. Lmk what you think :)

crates/chain/src/spk_iter.rs
```diff
@@ -195,6 +202,22 @@ mod test {
 
         let mut external_spk = SpkIterator::new(&no_wildcard_descriptor);
 
+        assert_eq!(external_spk.nth(0).unwrap(), (0, external_spk_0.clone()));
```
evanlinjin: Nit: this will give better errors. Will be really cool if we can tweak this for all asserts in the same PR.

```suggestion
        assert_eq!(external_spk.nth(0), Some((0, external_spk_0.clone())));
```

evanlinjin (commented):
Thanks for fixing this. This was my oversight.

As discussed (via DM), it'll be better to have non-wildcard descriptors only returns something on derivation index 0.

The follow tests should pass:

```rust
// non index-0 should NOT return an spk
assert_eq!(SpkIterator::new_with_range(&no_wildcard_descriptor, 1..1).next(), None);
assert_eq!(SpkIterator::new_with_range(&no_wildcard_descriptor, 1..=1).next(), None);
assert_eq!(SpkIterator::new_with_range(&no_wildcard_descriptor, 1..2).next(), None);
assert_eq!(SpkIterator::new_with_range(&no_wildcard_descriptor, 1..=2).next(), None);
```

crates/chain/src/spk_iter.rs
```diff
@@ -45,34 +45,43 @@ where
...
+
         // Because `end` is exclusive, we want the maximum value to be BIP32_MAX_INDEX + 1.
         end = end.min(BIP32_MAX_INDEX + 1);
 
+        if !descriptor.borrow().has_wildcard() {
+            // The length of the range should be at most 1
+            if end != start {
+                end = start + 1;
+            }
+        }
```
evanlinjin, line 81: Nit: I guess they end up doing the same thing anyway.

```suggestion

        if descriptor.borrow().has_wildcard() {
            // Because `end` is exclusive, we want the maximum value to be BIP32_MAX_INDEX + 1.
            end = end.min(BIP32_MAX_INDEX + 1);
        } else {
            // Non-wildcard should only return spk at index 0.
            end = end.min(1);
        }
```

evanlinjin (approved):
ACK e48b911c8d23cceab2071cc92085fe6675d55ca9

Tested with `example_electrum` with non-wildcard descriptor.
Left a non-critical nit.

crates/chain/src/spk_iter.rs
evanlinjin: @danielabrozzoni said she likes her way better.

danielabrozzoni: Sorry :(

PR #1095, opened 2023-08-25
Update `try_filter_chain_unspents` to check for what is confirmed and spendable. 

kngako:
### Description

Updated `tx_grap::try_filter_chain_unspents` to consider any output unspent using the `is_confirmed_and_spendable` method instead of checking whether `full_txo.spent_by.is_none()`.

### Notes to the reviewers

Since transaction can be replaced a utxo shouldn't be considered spent until there is a confirmation for that occurring.

#### All Submissions:

#### New Features:

#### Bugfixes:


evanlinjin:
Thank you for this PR. I see the intention is to change the behavior of `try_filter_chain_unspents` to only return UTXOs that are already confirmed.

This may be a better default for getting a list of UTXO candidates for coin selection. However this is not the purpose of this method. We want to return all UTXOs and have the caller responsible for filtering further (if needed). I.e. caller wants to only spend UTXOs with more than x confirmations.

kngako:
Hello @evanlinjin, think you for clarifying. Part of the reason I made this change is because once a UTXO is spent it is no longer returned to the caller (even though the caller may still control the UTXO, especially if it's spent via a RBF transaction which may need to be replaced). 

Would adding logic to require that the spend be confirmed be sufficient?

Ultimately we go from

```rust
.filter(|r| match r {
    // keep unspents, drop confirmed spends
    Ok((_, full_txo)) => full_txo.spent_by.is_none(),
    // keep errors
    Err(_) => true,
})
```

To 
```rust
.filter(|r| match r {
    // keep unspents, drop confirmed spends
    Ok((_, full_txo)) => {
        match &full_txo.spent_by {
            Some((chain_position, _)) => !chain_position.is_confirmed(),
            None => true,
        }
    },
    // keep errors
    Err(_) => true,
})
```

evanlinjin:
@kngako the problem with doing this, is after coin selection, we may accidentally create a transaction that does RBF.

For what you want to do, I think it is better to use `filter_chain_txouts` and filter the results from there.

evanlinjin:
Closing for now. Please use approach as suggested in https://github.com/bitcoindevkit/bdk/pull/1095#issuecomment-1699251759.

kngako:
Thank you.

kngako:
Oh, wait. Another remark. The `TxBuilder`.`add_utxos` function makes a call to `get_utxo` which then makes a call to `filter_chain_unspents` and ultimately means TxBuilder cannot build RBF transactions as txo outputs with unconfirmed spends will not be returned by the wallets `get_utxo` function.

Also the `add_foreign_utxo` which is there to add a UTXO not owned by the wallet seems like torture when the wallet does actually own the utxo.

---

Not sure if then `filter_chain_txouts` should be called in the `get_utxo` function instead of `filter_chain_unspents`?

evanlinjin:
@kngako usually for RBF, you want to replace a specific transaction. You can get the specific transaction, and therefore get it's spends.

LLFourn:
IMO what @kngako is getting at makes sense to me. You should be able to add whatever txout you want even if it's already spent in tx builder. It could be added as another method `add_txout` which doesn't check whether it's spent. You can already do this with `add_foreign_utxo`. Should probably rename `add_foreign_utxo` to `add_foreign_txout` as well since we don't know if it actually is unspent. I think we're aiming to rewrite tx builder on this principle so I don't think we need an issue for this.

PR #1109, opened 2023-09-01
Further improve unconfirmed tx conflict resolution

### LINKED ISSUE #1102: Further improve unconfirmed tx conflict resolution
**Describe the enhancement**  

At the end of ticket #1063's description, there is a **Further work** section which states:

> Our current structures do not handle unconfirmed conflicts properly when the conflicting transactions have the same `last_seen` timestamp (thanks to @rajarshimaitra for pointing this out).
>
> https://github.com/bitcoindevkit/bdk/blob/fe7bba1d9fcbfd9188fb036f7f2b488e62992a96/crates/chain/tests/test_indexed_tx_graph.rs#L692-L695
>
> I propose we deal with this as follows:
>
> When two unconfirmed txs conflict, but have the same last_seen, we can check the input's sequences (using RBF rules). Otherwise, we sort by feerate (if possible). We fallback to lexicographical sorting of txids.

LagginTimes:
### Description

Fixes #1102. If a conflicting tx has the same `last_seen`, then we check lexicographical sorting of txids.

### Notes to the reviewers

The tests for this fix exist in the `TxTemplate` structure in #1064 which may need to be merged first.

#### All Submissions:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -789,6 +789,12 @@ impl<A: Anchor> TxGraph<A> {
                 if conflicting_tx.last_seen_unconfirmed > tx_last_seen {
                     return Ok(None);
                 }
+                if conflicting_tx.last_seen_unconfirmed == *last_seen
+                    && conflicting_tx.txid() > tx.txid()
+                {
+                    // Conflicting tx has priority if txid of conflicting tx > txid of original tx
+                    return Ok(None);
+                }
             }
         }
 
--- a/crates/chain/tests/test_tx_graph_conflicts.rs
+++ b/crates/chain/tests/test_tx_graph_conflicts.rs
@@ -50,10 +50,10 @@ fn test_tx_conflict_handling() {
             tx_templates: &[
                 TxTemplate {
                     tx_name: "tx1",
-                    inputs: &[TxInTemplate::Bogus],
                     outputs: &[TxOutTemplate::new(40000, Some(0))],
                     anchors: &[block_id!(1, "B")],
                     last_seen: None,
+                    ..Default::default()
                 },
                 TxTemplate {
                     tx_name: "tx_conflict_1",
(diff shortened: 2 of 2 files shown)


LLFourn (commented):
Having considered it again I'm not sure if it's a good idea to make this change. It's a half-fix. This tie-break will not necessarily reflect what full mempool nodes decide since it doesn't consider the feerates of parents etc. I do think it's a good idea to tie-break consistently and deterministically (so just arbitrarily choose txid by lexical order). We could do the simplest thing here and then make an issue to implement logic which reflects mempool rules of bitcoin core.

If we were to do this one I'd expect to have a few more tests. I don't think each branch here can be reach with each test.

cc @danielabrozzoni @evanlinjin wdyt?

danielabrozzoni:
I agree with Lloyd's points, tie-breaking by tx feerate doesn't make much sense, as it's the package feerate that counts. We can go with tie-breaking by txid for now.

I also wonder how often it happens that two transactions have the same last_seen.

LagginTimes:
Updated to tiebreak by txid only.

LLFourn (approved):
ACK c4aea756792f26a169da1e113dca5e3a51716f12

These two commits should be squashed.
Thanks @LagginTimes.

notmandatory:
I'm hoping we can fix the CI issues with #1183 but if that doesn't happen #1182 and then this PR will need a rebase.

nondiremanuel:
As per today's discussion in the Lib Team Call, this PR probably needs to be rebased after the ci fix of #1182

LagginTimes:
Rebased with #1182 CI fixes.

LagginTimes:
Rebased with #1182 CI fixes.

crates/chain/src/tx_graph.rs
```diff
@@ -789,6 +789,12 @@ impl<A: Anchor> TxGraph<A> {
                 if conflicting_tx.last_seen_unconfirmed > tx_last_seen {
                     return Ok(None);
                 }
+                if conflicting_tx.last_seen_unconfirmed == *last_seen
+                    && conflicting_tx.txid() > tx.txid()
+                {
+                    // Conflicting tx has priority if txid of conflicting tx > txid of original tx
```
danielabrozzoni, line 795: nit: this should be documented, but I don't want to hold up merging, so I'm going to do it as a part of #1188

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -70,14 +70,12 @@ fn test_tx_conflict_handling() {
                     ..Default::default()
                 },
             ],
-            // correct output if filtered by fee rate: tx1, tx_conflict_1
-            exp_chain_txs: HashSet::from(["tx1", "tx_conflict_1", "tx_conflict_2"]),
-            exp_chain_txouts: HashSet::from([("tx1", 0), ("tx_conflict_1", 0), ("tx_conflict_2", 0)]),
-            // correct output if filtered by fee rate: tx_conflict_1
-            exp_unspents: HashSet::from([("tx_conflict_1", 0), ("tx_conflict_2", 0)]),
+            exp_chain_txs: HashSet::from(["tx1", "tx_conflict_2"]),
```
danielabrozzoni, line 73: Same as above, this needs docs, will do it as a part of #1188

danielabrozzoni (approved):
ACK afbf83c8b0d5c97495e6ceaa39963415edef6911

PR #1110, opened 2023-09-04
fix(esplora): use saturating_add in update_tx_graph()

notmandatory:
### Description

This fixes overflow error when calling update_tx_graph() from update_tx_graph_without_keychain().

### Notes to the reviewers

You can reproduce the error by reverting 66a2bf5ef2a70f8de41941d2f07bec776d152b96.

The tests could use some cleanup but get the job done for this PR.

### Changelog notice

None

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -35,11 +35,12 @@ jobs:
           cargo update -p rustls:0.21.7 --precise "0.21.1"
           cargo update -p rustls:0.20.9 --precise "0.20.8"
           cargo update -p tokio:1.32.0 --precise "1.29.1"
+          cargo update -p tokio-util --precise "0.7.8"
           cargo update -p flate2:1.0.27 --precise "1.0.26"
           cargo update -p reqwest --precise "0.11.18"
           cargo update -p h2 --precise "0.3.20"
           cargo update -p rustls-webpki:0.100.3 --precise "0.100.1"
-          cargo update -p rustls-webpki:0.101.5 --precise "0.101.1"
+          cargo update -p rustls-webpki:0.101.6 --precise "0.101.1"
           cargo update -p zip:0.6.6 --precise "0.6.2"
       - name: Build
         run: cargo build ${{ matrix.features }}
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -38,7 +38,9 @@ jobs:
           cargo update -p flate2:1.0.27 --precise "1.0.26"
           cargo update -p reqwest --precise "0.11.18"
           cargo update -p h2 --precise "0.3.20"
-          cargo update -p rustls-webpki --precise "0.100.1"
+          cargo update -p rustls-webpki:0.100.3 --precise "0.100.1"
+          cargo update -p rustls-webpki:0.101.5 --precise "0.101.1"
+          cargo update -p zip:0.6.6 --precise "0.6.2"
       - name: Build
         run: cargo build ${{ matrix.features }}
       - name: Test
--- a/crates/esplora/Cargo.toml
+++ b/crates/esplora/Cargo.toml
@@ -21,7 +21,7 @@ futures = { version = "0.3.26", optional = true }
 bitcoin = { version = "0.30.0", optional = true, default-features = false }
 miniscript = { version = "10.0.0", optional = true, default-features = false }
 
-[dev-dependencies]
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 electrsd = { version= "0.25.0", features = ["bitcoind_25_0", "esplora_a33e97e1", "legacy"] }
 tokio = { version = "1", features = ["rt", "rt-multi-thread", "macros"] }
 
--- a/crates/esplora/Cargo.toml
+++ b/crates/esplora/Cargo.toml
@@ -21,6 +21,10 @@ futures = { version = "0.3.26", optional = true }
 bitcoin = { version = "0.30.0", optional = true, default-features = false }
 miniscript = { version = "10.0.0", optional = true, default-features = false }
 
+[dev-dependencies]
+electrsd = { version= "0.25.0", features = ["bitcoind_25_0", "esplora_a33e97e1", "legacy"] }
+tokio = { version = "1", features = ["rt", "rt-multi-thread", "macros"] }
+
 [features]
 default = ["std", "async-https", "blocking"]
 std = ["bdk_chain/std"]
--- a/crates/esplora/tests/blocking_ext.rs
+++ b/crates/esplora/tests/blocking_ext.rs
@@ -1,0 +1,114 @@
+use bdk_esplora::EsploraExt;
+use electrsd::bitcoind::bitcoincore_rpc::RpcApi;
+use electrsd::bitcoind::{self, anyhow, BitcoinD};
+use electrsd::{Conf, ElectrsD};
+use esplora_client::{self, BlockingClient, Builder};
+use std::str::FromStr;
+use std::thread::sleep;
+use std::time::Duration;
+
+use bdk_chain::bitcoin::{Address, Amount, BlockHash, Txid};
+
+struct TestEnv {
+    bitcoind: BitcoinD,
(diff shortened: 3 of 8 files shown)


LLFourn:
I'm confused. How can this possibly overflow in practice?

notmandatory:
I ran into this when I was testing sync with esplora but only checking unused spks, based on how the cli example works. I could be misunderstanding how that's supposed to be done, but I thought I could just give the esplora client my spks via `update_tx_graph_without_keychain()` and I'd get back any new tx. But since this passes a stop_gap of u32::Max it causes an overflow with any last_active_index > 0.

notmandatory:
I was also able to repro this issue (using the original code) using the `example-crates/example_esplora` with these steps:

1. create a new address
    ```shell
     cargo run -- --network signet "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/0/*)" "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/1/*)" address new
   ```
3. send from faucet to new address
4. sync unused spks (don't think the parallel requests matter but that's how I did it)
    ```shell
    cargo run -- --network signet "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/0/*)" "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/1/*)" sync --unused-spks --parallel-requests 2
    ```

crates/esplora/tests/blocking_ext.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    let graph_update = env.client.update_tx_graph_without_keychain(
+        misc_spks.into_iter(),
+        vec![].into_iter(),
+        vec![].into_iter(),
+        1,
+    )?;
+
+    assert_eq!(graph_update.full_txs().count(), 2);
```
LLFourn: :bulb: test that the txids match those returned from `send_to_address`.

crates/esplora/Cargo.toml
```diff
@@ -21,9 +21,21 @@ futures = { version = "0.3.26", optional = true }
...
 default = ["std", "async-https", "blocking"]
 std = ["bdk_chain/std"]
 async = ["async-trait", "futures", "esplora-client/async"]
 async-https = ["async", "esplora-client/async-https"]
 blocking = ["esplora-client/blocking"]
+
+# This feature is used to run `cargo check` in our CI targeting wasm. It's not recommended
+# for libraries to explicitly include the "getrandom/js" feature, so we only do it when
+# necessary for running our CI. See: https://docs.rs/getrandom/0.2.8/getrandom/#webassembly-support
+dev-getrandom-wasm = ["getrandom/js"]
```
LLFourn: why do we need this. Where do we need to use `getrandom` at all here?

notmandatory: It looks like `getrandom` is brought in via the `electrsd` dev-dependency I added. One improvement I can make is to only include `getrandom` as a dev-dependency since that's the only place it's needed.

```
# from esplora directory
$ cargo tree -i getrandom                                                                                                                
getrandom v0.2.10
└── rand_core v0.6.4
    ├── rand v0.8.5
    │   └── secp256k1 v0.27.0
    │       └── bitcoin v0.30.1
    │           ├── bdk_chain v0.5.0 (/Users/steve/git/notmandatory/bdk/crates/chain)
    │           │   └── bdk_esplora v0.3.0 (/Users/steve/git/notmandatory/bdk/crates/esplora)
    │           ├── bitcoincore-rpc-json v0.17.0
    │           │   └── bitcoincore-rpc v0.17.0
    │           │       └── bitcoind v0.33.1
    │           │           └── electrsd v0.25.0
    │           │               [dev-dependencies]
    │           │               └── bdk_esplora v0.3.0 (/Users/steve/git/notmandatory/bdk/crates/esplora)
```

LLFourn: I think you don't need it at all since `electrsd` doesn't run in WASM anyway.

crates/esplora/tests/blocking_ext.rs
notmandatory: Fixed. 

There should be a way to consolidate blocking and async tests and TestEnv (probably with macros as Alekos originally did), I'll create an issue.

crates/esplora/Cargo.toml
notmandatory: For some reason `cargo check` with the WASM target is still including `getrandom` and throwing the error. I'll see if I can find some way to force it not to.

LLFourn: yeah because it's compiling dev depends in cargo check right? so it's gonna try and compile electrsd for WASM. So you need to `#[cfg(not(target_arch = "wasm32"))]` and not depend on the thing in the Cargo.toml if the arch is WASM.

Tbh I'm not sure why secp256k1 depends on `rand` maybe we can disable a feature that prevents including rand? That would be much simpler.

notmandatory: I figured out the issue. Using `[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]` in the `crates/esplora/Cargo.toml` didn't work until I enabled `resolver = "2"` in the top level workspace `Cargo.toml`. I would have expected it to already be using the new resolver since all our crates are set to `edition = "2021"`, but apparently the resolver also needs to be set for the whole workspace. 

https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#cargos-new-feature-resolver

notmandatory:
Rebased on master.

.github/workflows/cont_integration.yml
```diff
@@ -38,7 +38,9 @@ jobs:
           cargo update -p flate2:1.0.27 --precise "1.0.26"
           cargo update -p reqwest --precise "0.11.18"
           cargo update -p h2 --precise "0.3.20"
-          cargo update -p rustls-webpki --precise "0.100.1"
+          cargo update -p rustls-webpki:0.100.3 --precise "0.100.1"
+          cargo update -p rustls-webpki:0.101.5 --precise "0.101.1"
```
LLFourn: Why do we depend on two slightly different versions of webpki?

LLFourn (approved):
Thanks @notmandatory. Just to be clear I fully understand why it fails. I missed that `u32::MAX` was how we were disabling it.

ACK 7be6f86343e30c9e07bcfd52342a203ea7556026

.github/workflows/cont_integration.yml
notmandatory: Looks like `rustls-webpki` 0.100.x is required by the `rustls` version used by `electrum-client`. 

And `rustls-webpki` 0.101.x is required directly by the version of `minreq` used by our dev-dependencies `bitcoind` and `electrsd`. 

I think the ultimate solution for this will be to get all our dependencies to use the same major+minor version of `rustls`.

notmandatory:
Rebased and fixed two more ci MSRV issues. Had to pin `tokio-util` and update the pinned version of `rustls-webpki:0.101.6`.

notmandatory:
FYI, after I merged this PR the audit job failed due to the new dev dependency on  the `bitcoind`, I'm not sure why it didn't fail when I created the PR. The issue should be fixed with a new `bitcoind` patch release after https://github.com/RCasatta/bitcoind/pull/141 is merged.

PR #1111, opened 2023-09-05
Add fee-bump to example-cli example crate

sjeohp:
### Description

Opening for feedback. 

This adds a `run_bump_cmd` to `example_cli` and tweaks `nursery/coin_select/src/coin_selector.rs` to allow required utxos.

#### All Submissions:

* [ *] I've signed all my commits

#### New Features:

#### Bugfixes:

### DIFF
--- a/example-crates/example_cli/src/lib.rs
+++ b/example-crates/example_cli/src/lib.rs
@@ -3,26 +3,28 @@ use anyhow::Context;
 use bdk_coin_select::{coin_select_bnb, CoinSelector, CoinSelectorOpt, WeightedValue};
 use bdk_file_store::Store;
 use serde::{de::DeserializeOwned, Serialize};
-use std::{cmp::Reverse, collections::HashMap, path::PathBuf, sync::Mutex, time::Duration};
+use std::{cmp::Reverse, collections::HashMap, fmt, path::PathBuf, sync::Mutex, time::Duration};
 
 use bdk_chain::{
     bitcoin::{
         psbt::Prevouts, secp256k1::Secp256k1, util::sighash::SighashCache, Address, LockTime,
-        Network, Sequence, Transaction, TxIn, TxOut,
+        Network, OutPoint, Sequence, Transaction, TxIn, TxOut, Txid,
     },
     indexed_tx_graph::{IndexedAdditions, IndexedTxGraph},
     keychain::{DerivationAdditions, KeychainTxOutIndex},
+    local_chain::LocalChain,
     miniscript::{
         descriptor::{DescriptorSecretKey, KeyMap},
         Descriptor, DescriptorPublicKey,
     },
-    Anchor, Append, ChainOracle, DescriptorExt, FullTxOut, Persist, PersistBackend,
+    Anchor, Append, ChainOracle, ChainPosition, DescriptorExt, FullTxOut, Persist, PersistBackend,
 };
 pub use bdk_file_store;
 pub use clap;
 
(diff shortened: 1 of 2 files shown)


example-crates/example_cli/src/lib.rs
```diff
@@ -633,6 +667,112 @@ pub fn planned_utxos<A: Anchor, O: ChainOracle, K: Clone + bdk_tmp_plan::CanDeri
...
+    }
+
+    create_tx(
+        graph,
+        chain,
+        keymap,
+        cs_algorithm,
+        Some(new_feerate),
+        tx.output,
+        original_outpoints,
```
LLFourn, line 772: We only need to spend one of the original outpoints to RBF the transaction here.

example-crates/example_cli/src/lib.rs
```diff
@@ -633,6 +667,112 @@ pub fn planned_utxos<A: Anchor, O: ChainOracle, K: Clone + bdk_tmp_plan::CanDeri
...
+    let mut tx = graph
+        .graph()
+        .get_tx(txid)
+        .ok_or(Error::TransactionNotFound)?
+        .clone();
+
+    if !tx
+        .input
+        .iter()
+        .any(|txin| txin.sequence.to_consensus_u32() <= 0xFFFFFFFD)
```
LLFourn, line 733: I think that the `sequence` type might have a helper method for this check. I would just emit a warning here rather than error out.

example-crates/example_cli/src/lib.rs
```diff
@@ -633,6 +667,112 @@ pub fn planned_utxos<A: Anchor, O: ChainOracle, K: Clone + bdk_tmp_plan::CanDeri
...
+    }
+
+    // remove the inputs from the tx and process them
+    let original_outpoints = tx
+        .input
+        .drain(..)
+        .map(|txin| txin.previous_output)
+        .collect::<Vec<_>>();
+
+    if tx.output.len() > 1 {
```
LLFourn, line 745: I think we should avoid this "figure out which one is change" algorithm. Instead the API should be that we want to double spend a transaction (or even set of transactions) with a new spend. We don't need to enforce that the new spend has the same txouts as the old one. If you want to do ordinary RBF you just put the same destination address in that you used to create the original transaction. This means the application is responsible for recalling the intent of any particular transaction when they want to RBF it so that the new transaction's effect matches the old one.

So to be clear I think the CLI interface should just be to add a `--replace <txid>` flag to the `send` command which would just make sure that it spends at least one input of any `--replace` transaction and has a higher feerate and higher absolute fee than that transaction.

example-crates/example_cli/src/lib.rs
```diff
@@ -506,7 +537,10 @@ where
     let (_, selection_meta) = selection.best_strategy();
 
     // get the selected utxos
-    let selected_txos = selection.apply_selection(&candidates).collect::<Vec<_>>();
+    let selected_txos = selection
+        .apply_selection(&candidates)
+        .chain(required.iter())
```
LLFourn, line 542: So required utxos should just be selected in the `coin_selector` (then `apply_selection` is bound to include it).

LLFourn (commented):
Thanks @sjeohp. Left some comments. It will take some time to get this feature in since there's a lot in the pipline atm.

notmandatory:
I put this in the beta.0 milestone since it's only changing example code.

evanlinjin:
Stale. Also note that we are working on correct RBF logic in `bdk-tx`.

PR #1115, opened 2023-09-10
fix: remove deprecated `max_satisfaction_weight`

### LINKED ISSUE #1036: Use `max_weight_to_satisfy` instead of `max_satisfaction_weight`
Needs #1023

`max_satisfaction_weight` is deprecated, we should switch to using `max_weight_to_satisfy` instead.

Note that the switch is not an easy search and replace, the two methods have slightly different definitions

realeinherjar:
### Description

Closes #1036.

- Change deprecated `max_satisfaction_weight` to `max_weight_to_satisfy`
- Remove `#[allow(deprecated)]` flags

### Notes to the reviewers

I've changed all `max_satisfaction_weight()` to `max_weight_to_satisfy()`
in `Wallet.get_available_utxo()` and `Wallet.build_fee_bump()`.
Checking the docs on the `miniscript` crate for `max_weight_to_satisfy` has the following note:

> Since this method uses `segwit_weight` instead of `legacy_weight`,
> if you want to include only legacy inputs in your transaction,
> you should remove 1WU from each input's `max_weight_to_satisfy`
> for a more accurate estimate.

Hence, I also added 1WU (`4`) to the `max_weight_to_satisfy()` calls.
But I think that there is a better way to do this instead of hardcoding `4`.
Please advise.

### Fixed

Replace the deprecated `max_satisfaction_weight` from `rust-miniscript` to `max_weight_to_satisfy`.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -28,7 +28,6 @@ use bdk_chain::{
     Append, BlockId, ChainPosition, ConfirmationTime, ConfirmationTimeAnchor, FullTxOut,
     IndexedTxGraph, Persist, PersistBackend,
 };
-use bitcoin::consensus::encode::serialize;
 use bitcoin::psbt;
 use bitcoin::secp256k1::Secp256k1;
 use bitcoin::sighash::{EcdsaSighashType, TapSighashType};
@@ -36,6 +35,7 @@ use bitcoin::{
     absolute, Address, Network, OutPoint, Script, ScriptBuf, Sequence, Transaction, TxOut, Txid,
     Weight, Witness,
 };
+use bitcoin::consensus::encode::serialize;
 use core::fmt;
 use core::ops::Deref;
 use miniscript::psbt::{PsbtExt, PsbtInputExt, PsbtInputSatisfier};
@@ -1288,7 +1288,7 @@ impl<D> Wallet<D> {
                         let satisfaction_weight = self
                             .get_descriptor_for_keychain(keychain)
                             .max_weight_to_satisfy()
-                            .unwrap();
+                            .unwrap() + 4;
                         WeightedUtxo {
                             utxo: Utxo::Local(LocalUtxo {
                                 outpoint: txin.previous_output,
@@ -1298,7 +1298,7 @@ impl<D> Wallet<D> {
                                 derivation_index,
                                 confirmation_time,
                             }),
-                            satisfaction_weight,
+                            satisfaction_weight
                         }
                     }
                     None => {
@@ -1621,7 +1621,7 @@ impl<D> Wallet<D> {
                     utxo,
                     self.get_descriptor_for_keychain(keychain)
                         .max_weight_to_satisfy()
-                        .unwrap(),
+                        .unwrap() + 4,
                 )
             })
             .collect()
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -56,7 +56,6 @@ pub mod hardwaresigner;
 
 pub use utils::IsDust;
 
-#[allow(deprecated)]
 use coin_selection::DefaultCoinSelectionAlgorithm;
 use signer::{SignOptions, SignerOrdering, SignersContainer, TransactionSigner};
 use tx_builder::{BumpFee, CreateTx, FeePolicy, TxBuilder, TxParams};
@@ -1286,10 +1285,9 @@ impl<D> Wallet<D> {
 
                 let weighted_utxo = match txout_index.index_of_spk(&txout.script_pubkey) {
                     Some(&(keychain, derivation_index)) => {
-                        #[allow(deprecated)]
                         let satisfaction_weight = self
                             .get_descriptor_for_keychain(keychain)
-                            .max_satisfaction_weight()
+                            .max_weight_to_satisfy()
                             .unwrap();
                         WeightedUtxo {
                             utxo: Utxo::Local(LocalUtxo {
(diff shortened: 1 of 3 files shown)


danielabrozzoni:
I'm pretty sure that adding a +4 is not enough to make the code correct, but proving it isn't as easy as I thought.

Adding +4 is fine for segwit transactions, but it's not for legacy ones. Looking at the bdk txbuilding tests, it seems to me that we never test with legacy descriptors:

https://github.com/bitcoindevkit/bdk/blob/43fb0b20df92492bee763e26fd0b365af4f076bb/crates/bdk/tests/common.rs#L111-L156

This should be improved. I tried to quickly add a `get_test_pkh` to (hopefully) make `test_bump_fee_add_input` fail, but `assert_fee_rate!` works for segwit inputs only (it uses the witness_utxo).

I would tackle this issue in steps:
1. Improve the current testing environment to also test for legacy inputs (as stated above, this would include improving the current asset_fee_rate)
2. If I'm correct, step 1 would introduce some tests that currently fail 
3. Improve the current code. I would start with renaming the `satisfaction_weight` in `WeightedUtxo` to `weight_to_satisfy` for clarity, and adding a small description copy-pasting the meaning of `weight_to_satisfy` (can be copied from rust-ms). Then, change the code that uses `WeightedUtxo::weight_to_satisfy`. At the moment we have a `TXIN_BASE_WEIGHT` constant that we would add to `satisfaction_weight`, for example, see:

https://github.com/bitcoindevkit/bdk/blob/43fb0b20df92492bee763e26fd0b365af4f076bb/crates/bdk/src/wallet/coin_selection.rs#L311

We should instead have different constants (I think you can just use `Txin::default().legacy_weight()` and `Txin::default().segwit_weight()` and add the weight_to_satisfy to those constants based on whether the tx is segwit or not.

Just a small note: I understand that this issue is much bigger and more complicated than what it seemed. Feel free to ask for help when/if you need it, and to solve only parts of it (ex: only improve the testing framework) if you think it's best :)

realeinherjar:
> Adding +4 is fine for segwit transactions, but it's not for legacy ones

I totally agree with you.
The issue for me is not what needs to be done, but how are we gonna do it...
I know that this is urgent in the 1.0 roadmap


> 1. Improve the current testing environment to also test for legacy inputs (as stated above, this would include improving the current asset_fee_rate)

Yes, also agree that this is step 1.
I will work on that straight away.

> Just a small note: I understand that this issue is much bigger and more complicated than what it seemed. Feel free to ask for help when/if you need it, and to solve only parts of it (ex: only improve the testing framework) if you think it's best :)

Yes, sure thanks. This PR could also be more than 2 hands.

> `TxIn::default().legacy_weight()` and `TxIn::default().segwit_weight()`

I was exactly thinking that.
My early drafts used it.
**I have one question about implementation**:
What would be a good approach to figure it out which weight constant to add?
I was trying to do a `match` on [`Descriptor`](https://docs.rs/miniscript/latest/miniscript/descriptor/enum.Descriptor.html) from `miniscript`.
However, I was trying to do it inside `Wallet.build_fee_bump()` and `TxBuilder.add_utxos()`.
But, seeing your comments it seems that a better place would be `CoinSelectionAlgorithm.select_sorted_utxos()`?

danielabrozzoni:
> This PR could also be more than 2 hands.

I think the more people work on it, the harder it gets to communicate what needs to be done. Ofc many can help with reviews, but I'll let you do the code if you feel up for it :)

> But, seeing your comments it seems that a better place would be CoinSelectionAlgorithm.select_sorted_utxos()?

I was initially thinking of some kind of `is_segwit` parameter in `coin_select`, but I now realize this isn't the clean way to go. The coin selection should only select utxos, the logic for calculating the weight should be done outside.

Matching on the DescriptorType is a good start (see https://docs.rs/miniscript/latest/miniscript/descriptor/enum.DescriptorType.html#method.segwit_version), but unfortunately not enough, as you also need to consider foreign utxos (consider a user with a legacy descriptor, adding a segwit foreign utxo). 

One thing we might want to do is to consider all txs to be segwit by default, but then the user can customize this using a `set_segwit` method in the tx builder. We would then modify `WeightedUtxo::satisfaction_weight` to include the txin base weight (instead of it being added while selecting utxos).

However, I don't think this is a clean solution, it seems quite hacky and not really clear for the user.

We could also:
- Reconsider our priorities and push this issue to bdk 2.0 (or 3.0? I can't remember), when we'll have a decent interface for coin selection and tx building. Our fee estimation is quite broken anyways... We can keep using `max_satisfaction_weight` and calling it a day
- (Similarly to what it's done in this PR), assume segwit transactions, and overestimate for legacy transactions.

Thoughts? cc @evanlinjin @LLFourn

realeinherjar:
> * Reconsider our priorities and push this issue to bdk 2.0 (or 3.0? I can't remember), when we'll have a decent interface for coin selection and tx building. Our fee estimation is quite broken anyways... We can keep using `max_satisfaction_weight` and calling it a day

That is also valid, this was deprecated in Jan 2023 (see https://github.com/rust-bitcoin/rust-miniscript/pull/476).
However, even if we close this by pushing this to 2.0+, I would still open a new PR to test more corner cases in fee bump and tx build (I didn't like that my hardcoded `4` passed all tests).

realeinherjar:
I already have tests that fail the `+ 4` hack in b902be2e38cfc1350fda8712be8db77d05973845 using a legacy PKH wallet.

notmandatory:

storopoli:
I'm taking over this in a new PR with some tweaks and a rebase

Issue #1107, opened 2023-09-01
BlockId should not have a default and passing a chain tip should be an `Option`


LLFourn:
It does not make sense for a BlockId to have a default. We use the `Default` to semantically mean a null block. e.g.
```
self.indexed_graph
            .graph()
            .filter_chain_unspents(
                &self.chain,
                self.chain.tip().map(|cp| cp.block_id()).unwrap_or_default(),
                self.indexed_graph.index.outpoints().iter().cloned(),
            )
```

I think these should just take an `Option` for the tip since that's the rusty way of communicating null. When it's `None` it should just be assumed that nothing is in the chain.
Maybe the tip should take an argument that is `Option<B> where B: Into<BlockId>` to avoid the `map`.

On a related note `ChainOracle` has `get_chain_tip` despite this not being used anywhere. I'd say we should remove it. Perhaps we will want this later but we can just make a new trait for that like `GetChainTip`.

SpaghettiSats:
I agree with you, `BlockId` should not have a `Default` and `chain_tip` should definitely be an `Option`. But I don't think `get_chain_tip` should be removed. I don't have such a broad view of the project, but I think it can be useful from the perspective of wallet developers, for example as used in example crates.
I would love to work on this if others also find it necessary.

LLFourn:
Hey @SpaghettiSats. Getting the chain tip is clearly very useful, it just doesn't do anything by being in the trait at the moment. In the examples or your app you will almost always be dealing with a concrete impl of `ChainOracle` so you can just call its concrete method.

There is a way the library could make use of `get_chain_tip` and actually justify its existence and that's to have a more convenient version of `filter_chain_unspents` that doesn't take the tip argument but just fetches it automatically from chain (the first argument). We could consider making `filter_chain_unspents` the name of that convenient version and have `filter_subchain_unspents` or something have the API of the current method. Of course we'd need to do this for the fallible version as well. We should also do it for all the other methods that allow passing a chain and a chain tip e.g. balance. So it's quite a bit of typing but it would make the API nicer!

That's the direction I'd take if we want to keep it. Please have a go at that if you feel so inspired.

SpaghettiSats:
Thanks for the explanation. Now I understand. I will try to do so!

LLFourn:
See discussion: https://github.com/bitcoindevkit/bdk/pull/1116#issuecomment-1752205748
This should be fixed with: https://github.com/bitcoindevkit/bdk/issues/1079

Issue #1079, opened 2023-08-16
Create LocalChain with genesis block hardwired for Network


notmandatory:
**Describe the enhancement**

`LocalChain` will no longer implement `Default`. Instead, it will require a genesis hash to construct. This ensures that we will always have at least one checkpoint in `LocalChain` (the genesis block).

`ChainOracle::get_chain_tip` will no longer need to return an `Option` as we guarantee that we at least have the genesis block.

```rust
pub trait ChainOracle {
    /* other trait methods and types */

    /// Get the best chain's chain tip.
    fn get_chain_tip(&self) -> Result<BlockId, Self::Error>;
}
```

**Implementation details**

`LocalChain` will need to ensure that the genesis checkpoint cannot be replaced while updating. I think it would also make sense to persist the genesis checkpoint, this way recovery from persistence will fail with the wrong genesis hash.

**Changes to chain-source crates**

Any method that takes in an option of a checkpoint should just take in the checkpoint (no `Option`).

For the `bdk_bitcoind_rpc` crate, instead of emitting `(u32, Block)`, we want to emit `(CheckPoint, Block)`. This is because for the first emission (given a start height that is >1), we can no longer rely on the block header to construct the chain update (since we now need to connect to the genesis block!). Additionally, if there is no previous checkpoint (`Emitter::last_cp`), the first emission's checkpoint should also include the genesis block.

**Use case**

This fixes #1107 without requiring methods of `TxGraph` and `IndexedTxGraph` to take in an `Option` of `BlockId` as the *static block*.

The additional benefit of this is that if the caller accidentally updates `LocalChain` with dat from the wrong network, the update can never connect (since the genesis block cannot be replaced).

**Additional context**

Suggested by @evanlinjin in bitcoindevkit/.github#72.

Comment on `PR#1116`: https://github.com/bitcoindevkit/bdk/pull/1116#issuecomment-1752042855

danielabrozzoni:
I'm taking a shot at this one :)

LLFourn:
What is the benefit of this and what is the planned API?

notmandatory:
I think the idea is that you can create a new LocalChain prepopulated with the appropriate genesis BlockId and this would prevent blocks from the wrong chain from being added to it. Is that what you had in mind @evanlinjin ?

LLFourn:
So this would be an optional kind of constructor like `LocalChain::new_mainnet`?

evanlinjin:
[...]

PR #1116, opened 2023-09-11
BlockId should not have a default and passing a chain tip should be an Option #1107 

SpaghettiSats:
### Description

This pull request addresses the issue discussed in [GitHub Issue #1107](https://github.com/bitcoindevkit/bdk/issues/1107). The primary changes made in this PR are as follows:

1. **BlockId Default Removal**: Removed the default implementation for `BlockId` as it doesn't make sense for it to have a default value. Instead, we rely on `Option<BlockId>` to semantically represent a null block.

2. **Chain Tip as Option**: Changed the `chain_tip` parameter in various methods to accept an `Option<BlockId>` instead of using `unwrap_or_default()`. This aligns with Rust conventions for handling null values.

3. **ChainOracle's `get_chain_tip`**: While keeping `get_chain_tip`, it is now made more meaningful by adding functionality to fetch the chain tip from the `chain` argument when calling certain methods. This makes the `get_chain_tip` method more useful and justified.

4. **API Improvement**: Introduced a more convenient version of `filter_chain_unspents` that automatically fetches the chain tip from the `chain` argument, simplifying usage. The original method is retained as `filter_subchain_unspents` to maintain backward compatibility. Similar improvements are made for the following methods: `get_chain_position`, `get_chain_spend`, `list_chain_txs`, `filter_chain_txouts`,  `balance` and of course also for their fallible versions.

### Notes to the reviewers

- This PR aims to improve the API by aligning it with Rust conventions for handling null values and making it more user-friendly.
- Carefully reviewed the changes to ensure consistency with existing code and adherence to project guidelines.
- Tests have been added for the new features and adjustments.

#### All Submissions:

- [] I've signed all my commits.
- [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md).
- [x] I ran `cargo fmt` and `cargo clippy` before committing.

#### New Features:

- [x] I've added tests for the new feature.
- [x] I've added docs for the new feature

#### Bugfixes:

- [x] This pull request breaks the existing API.
- [ ] I've added tests to reproduce the issue which are now passing (Not applicable in this case).
- [x] I'm linking the issue being fixed by this PR ([GitHub Issue #1107](https://github.com/bitcoindevkit/bdk/issues/1107)).

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -435,9 +435,8 @@ impl<D> Wallet<D> {
     pub fn list_unspent(&self) -> impl Iterator<Item = LocalUtxo> + '_ {
         self.indexed_graph
             .graph()
-            .filter_subchain_unspents(
+            .filter_chain_unspents(
                 &self.chain,
-                self.chain.tip().map(|cp| cp.block_id()),
                 self.indexed_graph.index.outpoints().iter().cloned(),
             )
             .map(|((k, i), full_txo)| new_local_utxo(k, i, full_txo))
@@ -485,11 +484,7 @@ impl<D> Wallet<D> {
         let (&spk_i, _) = self.indexed_graph.index.txout(op)?;
         self.indexed_graph
             .graph()
-            .filter_subchain_unspents(
-                &self.chain,
-                self.chain.tip().map(|cp| cp.block_id()),
-                core::iter::once((spk_i, op)),
-            )
+            .filter_chain_unspents(&self.chain, core::iter::once((spk_i, op)))
             .map(|((k, i), full_txo)| new_local_utxo(k, i, full_txo))
             .next()
     }
@@ -658,11 +653,7 @@ impl<D> Wallet<D> {
         let graph = self.indexed_graph.graph();
 
         Some(CanonicalTx {
-            chain_position: graph.get_subchain_position(
-                &self.chain,
-                self.chain.tip().map(|cp| cp.block_id()),
-                txid,
-            )?,
+            chain_position: graph.get_chain_position(&self.chain, txid)?,
             tx_node: graph.get_tx_node(txid)?,
         })
     }
@@ -748,17 +739,14 @@ impl<D> Wallet<D> {
     pub fn transactions(
         &self,
     ) -> impl Iterator<Item = CanonicalTx<'_, Transaction, ConfirmationTimeAnchor>> + '_ {
-        self.indexed_graph
-            .graph()
-            .list_subchain_txs(&self.chain, self.chain.tip().map(|cp| cp.block_id()))
+        self.indexed_graph.graph().list_chain_txs(&self.chain)
     }
 
     /// Return the balance, separated into available, trusted-pending, untrusted-pending and immature
     /// values.
     pub fn get_balance(&self) -> Balance {
-        self.indexed_graph.graph().subchain_balance(
+        self.indexed_graph.graph().balance(
             &self.chain,
-            self.chain.tip().map(|cp| cp.block_id()),
             self.indexed_graph.index.outpoints().iter().cloned(),
             |&(k, _), _| k == KeychainKind::Internal,
         )
@@ -1238,7 +1226,6 @@ impl<D> Wallet<D> {
     ) -> Result<TxBuilder<'_, D, DefaultCoinSelectionAlgorithm, BumpFee>, Error> {
         let graph = self.indexed_graph.graph();
         let txout_index = &self.indexed_graph.index;
-        let chain_tip = self.chain.tip().map(|cp| cp.block_id());
 
         let mut tx = graph
             .get_tx(txid)
@@ -1246,7 +1233,7 @@ impl<D> Wallet<D> {
             .clone();
 
         let pos = graph
-            .get_subchain_position(&self.chain, chain_tip, txid)
+            .get_chain_position(&self.chain, txid)
             .ok_or(Error::TransactionNotFound)?;
         if let ChainPosition::Confirmed(_) = pos {
             return Err(Error::TransactionConfirmed);
@@ -1278,7 +1265,7 @@ impl<D> Wallet<D> {
                 let txout = &prev_tx.output[txin.previous_output.vout as usize];
 
                 let confirmation_time: ConfirmationTime = graph
-                    .get_subchain_position(&self.chain, chain_tip, txin.previous_output.txid)
+                    .get_chain_position(&self.chain, txin.previous_output.txid)
                     .ok_or(Error::UnknownUtxo)?
                     .cloned()
                     .into();
@@ -1473,8 +1460,6 @@ impl<D> Wallet<D> {
         psbt: &mut psbt::PartiallySignedTransaction,
         sign_options: SignOptions,
     ) -> Result<bool, Error> {
-        let chain_tip = self.chain.tip().map(|cp| cp.block_id());
-
         let tx = &psbt.unsigned_tx;
         let mut finished = true;
 
@@ -1489,7 +1474,7 @@ impl<D> Wallet<D> {
             let confirmation_height = self
                 .indexed_graph
                 .graph()
-                .get_subchain_position(&self.chain, chain_tip, input.previous_output.txid)
+                .get_chain_position(&self.chain, input.previous_output.txid)
                 .map(|chain_position| match chain_position {
                     ChainPosition::Confirmed(a) => a.confirmation_height,
                     ChainPosition::Unconfirmed(_) => u32::MAX,
@@ -1642,7 +1627,6 @@ impl<D> Wallet<D> {
         must_only_use_confirmed_tx: bool,
         current_height: Option<u32>,
     ) -> (Vec<WeightedUtxo>, Vec<WeightedUtxo>) {
-        let chain_tip = self.chain.tip().map(|cp| cp.block_id());
         //    must_spend <- manually selected utxos
         //    may_spend  <- all other available utxos
         let mut may_spend = self.get_available_utxos();
@@ -1671,7 +1655,7 @@ impl<D> Wallet<D> {
                 let confirmation_time: ConfirmationTime = match self
                     .indexed_graph
                     .graph()
-                    .get_subchain_position(&self.chain, chain_tip, txid)
+                    .get_chain_position(&self.chain, txid)
                 {
                     Some(chain_position) => chain_position.cloned().into(),
                     None => return false,
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -435,7 +435,7 @@ impl<D> Wallet<D> {
     pub fn list_unspent(&self) -> impl Iterator<Item = LocalUtxo> + '_ {
         self.indexed_graph
             .graph()
-            .filter_chain_unspents(
+            .filter_subchain_unspents(
                 &self.chain,
                 self.chain.tip().map(|cp| cp.block_id()),
                 self.indexed_graph.index.outpoints().iter().cloned(),
@@ -485,7 +485,7 @@ impl<D> Wallet<D> {
         let (&spk_i, _) = self.indexed_graph.index.txout(op)?;
         self.indexed_graph
             .graph()
-            .filter_chain_unspents(
+            .filter_subchain_unspents(
                 &self.chain,
                 self.chain.tip().map(|cp| cp.block_id()),
                 core::iter::once((spk_i, op)),
@@ -658,7 +658,7 @@ impl<D> Wallet<D> {
         let graph = self.indexed_graph.graph();
 
         Some(CanonicalTx {
-            chain_position: graph.get_chain_position(
+            chain_position: graph.get_subchain_position(
                 &self.chain,
                 self.chain.tip().map(|cp| cp.block_id()),
                 txid,
@@ -750,13 +750,13 @@ impl<D> Wallet<D> {
     ) -> impl Iterator<Item = CanonicalTx<'_, Transaction, ConfirmationTimeAnchor>> + '_ {
         self.indexed_graph
             .graph()
-            .list_chain_txs(&self.chain, self.chain.tip().map(|cp| cp.block_id()))
+            .list_subchain_txs(&self.chain, self.chain.tip().map(|cp| cp.block_id()))
     }
 
     /// Return the balance, separated into available, trusted-pending, untrusted-pending and immature
     /// values.
     pub fn get_balance(&self) -> Balance {
-        self.indexed_graph.graph().balance(
+        self.indexed_graph.graph().subchain_balance(
             &self.chain,
             self.chain.tip().map(|cp| cp.block_id()),
             self.indexed_graph.index.outpoints().iter().cloned(),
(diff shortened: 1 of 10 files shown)


crates/chain/src/local_chain.rs
```diff
@@ -171,11 +171,16 @@ impl From<BTreeMap<u32, BlockHash>> for LocalChain {
...
         &self,
         block: BlockId,
-        chain_tip: BlockId,
+        chain_tip: Option<B>,
     ) -> Result<Option<bool>, Self::Error> {
+        let chain_tip: BlockId = match chain_tip {
+            Some(x) => x.into(),
+            None => return Ok(None),
+        };
+
```
realeinherjar: Wouldn't it better to return `Result<bool, Self::Error>`?
`Result<Option<bool>, Self::Error>` seems unnecessary

realeinherjar (commented):
Do we really need to `.clone()` all the `Into<BlockId>`?
Can't we pass by reference in some of these?

crates/bdk/src/wallet/mod.rs
```diff
@@ -372,9 +372,9 @@ impl<D> Wallet<D> {
     pub fn list_unspent(&self) -> impl Iterator<Item = LocalUtxo> + '_ {
         self.indexed_graph
             .graph()
-            .filter_chain_unspents(
+            .filter_subchain_unspents(
```
LLFourn: this looks like these can just be `filter_chain_unspents` as it's just using `self.chain.tip`.

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -469,5 +475,80 @@ fn test_list_owned_txouts() {
...
+        assert_eq!(unconfirmed_utxos_txid, unconfirmed_utxos_txid_);
+        assert_eq!(
+            confirmed_txouts_txid.len() + unconfirmed_txouts_txid.len(),
+            txouts.len()
+        );
+        assert_eq!(
+            confirmed_utxos_txid.len() + unconfirmed_utxos_txid.len(),
+            utxos.len()
+        );
+        assert_eq!(balance, balance_);
```
LLFourn: Not sure if these tests are worth maintaining them. Let's try and minimize changes to this file and leave them out for now as they will conflict a lot with [#1102 ](https://github.com/bitcoindevkit/bdk/pull/1109)

crates/chain/tests/test_tx_graph.rs
```diff
@@ -724,7 +724,23 @@ fn test_chain_spends() {
 
     // Assert that confirmed spends are returned correctly.
     assert_eq!(
-        graph.get_chain_spend(&local_chain, tip.block_id(), OutPoint::new(tx_0.txid(), 0)),
+        graph.get_subchain_spend(
```
LLFourn: No need to test both methods. Just test `get_chain_spend`.

example-crates/example_cli/src/lib.rs
```diff
@@ -608,11 +608,11 @@ pub fn planned_utxos<A: Anchor, O: ChainOracle, K: Clone + bdk_tmp_plan::CanDeri
     chain: &O,
     assets: &bdk_tmp_plan::Assets<K>,
 ) -> Result<Vec<(bdk_tmp_plan::Plan<K>, FullTxOut<A>)>, O::Error> {
-    let chain_tip = chain.get_chain_tip()?.unwrap_or_default();
+    let chain_tip = chain.get_chain_tip()?;
     let outpoints = graph.index.outpoints().iter().cloned();
     graph
         .graph()
-        .try_filter_chain_unspents(chain, chain_tip, outpoints)
+        .try_filter_subchain_unspents(chain, chain_tip, outpoints)
```
LLFourn: This should just be `try_filter_chain_unspents`

LLFourn (commented):
Thanks for doing this tedious work @SpaghettiSats.  General comments:

1. No need to add specific tests for these. Try to avoid as much as possible changes to the test files in this PR. I think it's quite hard to implement these convenience methods incorrectly.
2. Docs need to explain what a subchain is. Liberally refer to the other method e.g. `subchain_X is like [X] except it restricts the chain to a custom tip`. The tip doesn't even need to be in the same chain as the tip. You can use this to find information about a point in the past or on a fork (if your chain oracle supports that).
3. I think after looking at it we don't need to do the `Into<BlockId>` thing because it's already convenient enough now that we have all these variants. I'd just take a `BlockId`. Should address @realeinherjar's complaint above.
5. Use the convenience methods wherever possible in `wallet/mod.rs` and in the tests etc

crates/chain/src/local_chain.rs
LLFourn: There are three possibilities:

- `Some(true)`: The block is in the chain
- `Some(false)`: The block is not in the chain
- `None` we don't know if the block is in the chain.

evanlinjin: Maybe in the future, we can consider turning this into an enum?

SpaghettiSats:
> Thanks for doing this tedious work @SpaghettiSats. General comments:
> 
>     1. No need to add specific tests for these. Try to avoid as much as possible changes to the test files in this PR. I think it's quite hard to implement these convenience methods incorrectly.
> 
>     2. Docs need to explain what a subchain is. Liberally refer to the other method e.g. `subchain_X is like [X] except it restricts the chain to a custom tip`. The tip doesn't even need to be in the same chain as the tip. You can use this to find information about a point in the past or on a fork (if your chain oracle supports that).
> 
>     3. I think after looking at it we don't need to do the `Into<BlockId>` thing because it's already convenient enough now that we have all these variants. I'd just take a `BlockId`. Should address @realeinherjar's complaint above.
> 
>     4. Use the convenience methods wherever possible in `wallet/mod.rs` and in the tests etc

Thanks for the feedback. I'll work on these suggestions.

crates/chain/src/local_chain.rs
LLFourn: Not sure. Would just add documentation as far as I can see. Documenting the trait method seems clear enough. I could be wrong.

crates/chain/src/tx_graph.rs
```diff
@@ -644,6 +644,11 @@ impl<A: Anchor> TxGraph<A> {
 
     /// Get the position of the transaction in `chain` with tip `chain_tip`.
     ///
+    /// This method is like [`try_get_chain_position`] except it restricts the
+    /// chain to a custom tip. The tip doesn't even need to be in the same chain as the tip.
```
```suggestion
    /// chain to a custom tip. The tip doesn't even need to be in the same chain as the current tip.
```

LLFourn (approved):
ACK be8665085aa114ff03c58b62f22133171c14621a

Thanks @SpaghettiSats. I think another review from someone to confirm that these extra methods are worth maintaining for the convenience they offer.

SpaghettiSats:
I have rebased my branch and signed all the commits I have made, please if there are any other issues tell me and I will try to fix them.

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -469,5 +475,80 @@ fn test_list_owned_txouts() {
                 confirmed: 80000          // tx1 + tx3
             }
         );
+
+        // Test that not specifing the chain_tip return the same results
```
danielabrozzoni: 025c4beef733ce76f45b5517e2ba895000d56de5 is adding some tests that 7adfedb992ab30b98edd03f08b54377e5505c73c is removing, can you squash the two commits together instead?

crates/chain/src/chain_oracle.rs
```diff
@@ -17,7 +17,7 @@ pub trait ChainOracle {
     fn is_block_in_chain(
         &self,
         block: BlockId,
-        chain_tip: BlockId,
+        chain_tip: Option<&BlockId>,
```
danielabrozzoni: Mh, wouldn't it be better if we just had `BlockId` here? I can't really think of any case where someone would call `is_block_in_chain(block, None)`, giving that it would always return `None`

crates/chain/src/tx_graph.rs
```diff
@@ -644,19 +644,25 @@ impl<A: Anchor> TxGraph<A> {
...
     ///
-    /// [`get_chain_position`]: Self::get_chain_position
-    pub fn try_get_chain_position<C: ChainOracle>(
+    /// [`try_get_chain_position`]: Self::try_get_chain_position
+    /// [`get_subchain_position`]: Self::get_subchain_position
+    pub fn try_get_subchain_position<C: ChainOracle>(
         &self,
         chain: &C,
-        chain_tip: BlockId,
+        chain_tip: Option<BlockId>,
```
danielabrozzoni, line 665: Same as my comment above, why having an Option here?

danielabrozzoni (commented):
Thank you so much for this! I agree that removing the `Default` implementation on `BlockId` and instead using `Option` is much clearer, and I like the new methods that automatically fetch the chain tip from the chain oracle instead of requiring the user to provide one, they make the API way clearer.

However, I wonder whether it makes sense to have `Option<BlockId>` in some places. I left two comments about this, there are many other places where I'm left wondering the same thing, but I didn't comment on those as I'm probably just missing something :)

SpaghettiSats:
First of all thank you for your review @danielabrozzoni. I agree with you. I will try to fix everything today.

crates/chain/src/tx_graph.rs
SpaghettiSats: After a deeper look, I think we should keep this `Option` for 2 reasons: for consistency between methods and because it is convenient to check if the `chain_tip` is `None` in just one place instead of having to check every time before calling the method (the `ChainOracle` give an `Option`). If you think that is still the case to just pass the `BlockId` let me know and I will do it.

danielabrozzoni: > for consistency between methods

My point was that most (all?) of these methods should have a `BlockId` instead of an `Option`, I just commented a couple of them for brevity. 


I see that it's convenient to have an Option here, but at the same time it makes the API a bit more difficult to understand. 

Eventually it all comes down to personal preference, if you want to keep an `Option` here it's fine, but I think the decision should at least be documented properly

LLFourn: I think I agree with @danielabrozzoni here. It causes friction to call it with the output of `get_chain_tip` but that friction is born by us with this PR mostly so +1.

SpaghettiSats: I get it, I'll fix the problem this weekend.

evanlinjin:
Approach NACK

Thank you for taking this on and communicating clearly.

Overall, I agree that `BlockId` should not implement `Default`.

However, I think it is important that all `TxGraph` methods that need to fetch chain positions should still take in `chain_tip: CheckPoint`. Especially when multiple calls to these methods are made. The `chain_tip` ensures consistency in case of state changes of our `ChainOracle`.

`Wallet::build_fee_bump` is an example of how the new `get_chain_position` can potentially be misused. It is okay here because we are using `LocalChain` (which is in `Wallet` and we are holding a mutable reference over it). However, imagine if this is a `ChainOracle` implementation based on a client interface to a full/cbf node.

Additionally, now we have extra methods doing different but similar things. This increases the complexity of our API.

**An Alternative Proposal:**

I think we should require all `ChainOracle` implementations to be aware of the genesis block. `LocalChain` will no longer implement `Default`, and the caller needs to provide the genesis hash to construct it. The signature of `ChainOracle::get_chain_tip` would be changed to:

```rust
fn get_chain_tip(&self) -> Result<BlockId, Self::Error>;
```

So all `TxGraph` methods (which require getting chain positions) can take in `chain_tip: CheckPoint` (no `Option`), since we can guarantee to always at least have the genesis checkpoint.

The additional benefit to this, is if the caller accidentally updates `LocalChain` with data from the wrong network, the update can never connect (since the genesis block can't be replaced).

**Related:**

- https://github.com/bitcoindevkit/bdk/issues/1079

LLFourn:
> Approach NACK
> 
> Thank you for taking this on and communicating clearly.
> 
> Overall, I agree that `BlockId` should not implement `Default`.
> 
> However, I think it is important that all `TxGraph` methods that need to fetch chain positions should still take in `chain_tip: CheckPoint`. Especially when multiple calls to these methods are made. The `chain_tip` ensures consistency in case of state changes of our `ChainOracle`.

I don't understand. They don't currently take in `chain_tip: CheckPoint`, they take a `BlockId` are you proposing to change this?

So to be clear, @evanlinjin's concern can be shown clearly in this change: https://github.com/bitcoindevkit/bdk/pull/1116/files#diff-ba2cf4a27dd1b461a84dc49953334c1efe322887febbcbe6f928057c36e7f7ec

The two calls should be using the same chain tip. They are because `LocalChain` has a lock on it during the two calls but we don't generally guarantee this for `ChainOracle` so if we were to switch it then it could come up with inconsistent answers. It's difficult to see how using an old chain tip for one call and then a newer chain tip for the next call is going to be worse than using an old chain tip for both calls -- but the concern is that users will be doing this without even knowing it.

> 
> `Wallet::build_fee_bump` is an example of how the new `get_chain_position` can potentially be misused. It is okay here because we are using `LocalChain` (which is in `Wallet` and we are holding a mutable reference over it). However, imagine if this is a `ChainOracle` implementation based on a client interface to a full/cbf node.
> 

I imagined it but couldn't see what exactly the problem is. What's the concern?

> Additionally, now we have extra methods doing different but similar things. This increases the complexity of our API.
> 

Convenience methods do add maintenance complexity but reduce complexity for the user. I think it depends on whether the convenience methods send the right message. If people are meant to use a method 90% of the time in one way then it makes sense to have a convenience method that simplifies that for them and reassures them that they are not "missing out" on the right way of doing something by always using it that way. I thought that's what my suggestion was doing but it actually was not. You're meant to pass in the the same block id over multiple calls to these methods to keep the answers consistent but the convenience methods stops you from doing that. Perhaps we could include more nuanced documentation on the methods to make this clearer.

> **An Alternative Proposal:**
> 
> I think we should require all `ChainOracle` implementations to be aware of the genesis block. `LocalChain` will no longer implement `Default`, and the caller needs to provide the genesis hash to construct it. The signature of `ChainOracle::get_chain_tip` would be changed to:
> 
> ```rust
> fn get_chain_tip(&self) -> Result<BlockId, Self::Error>;
> ```
> 
> So all `TxGraph` methods (which require getting chain positions) can take in `chain_tip: CheckPoint` (no `Option`), since we can guarantee to always at least have the genesis checkpoint.
> 
> The additional benefit to this, is if the caller accidentally updates `LocalChain` with data from the wrong network, the update can never connect (since the genesis block can't be replaced).
> 
> **Related:**
> 
>     * [Create LocalChain with genesis block hardwired for Network #1079](https://github.com/bitcoindevkit/bdk/issues/1079)

I don't think this is really an alternative proposal -- it's compatible with this PR also.

evanlinjin:
@LLFourn sorry I meant `BlockId`

LLFourn:
To be clear in conclusion, I think we should not take this PR's approach for now and instead just wait on implementing @evanlinjin's proposal. I guess we can leave `get_chain_tip` on there. After further thinking I do see why it's useful even if we don't use it in our our API.

Sorry for originally suggesting this direction @SpaghettiSats. At least we have a better direction going forward. I think the `Default` on `BlockId` can be removed at the same time as requirement that `LocalChain` always be constructed with a genesis `BlockId`.

LLFourn:
So since we are in fact going in the direction of #1079 let's close this. Once again sorry for the wasted work @SpaghettiSats. I wonder if BDK team can help prevent me from wasting other people's time like this. It seems I casually circumvented the issue validation and execution engine of the BDK team and it cost us.

An obvious suggestion is that we have ways of labeling issues that have actually been groomed by the core team that are ready for others to pick up.

Could you bring this up in the next dev meeting: @evanlinjin @notmandatory

notmandatory:
I think for grooming new issues we should use the project "Status" field and only mark issues as "Todo" we're sure are ready and should be worked on.

PR #1126, opened 2023-09-15
Remove `TxGraph::missing_heights` as `ChangeSet::missing_heights_from` should have replaced it

evanlinjin:
### Description

`ChangeSet::missing_heights_from` is introduced as a simpler alternative to `TxGraph::missing_heights`. We should remove `TxGraph::missing_heights` and use `ChangeSet::missing_heights_from` instead.

### Notes to the reviewers

WIP

### Changelog notice

Remove `TxGraph::missing_heights` and use `ChangeSet::missing_heights_from` instead.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bdk/src/lib.rs
+++ b/crates/bdk/src/lib.rs
@@ -43,6 +43,7 @@ pub use types::*;
 pub use wallet::signer;
 pub use wallet::signer::SignOptions;
 pub use wallet::tx_builder::TxBuilder;
+pub use wallet::Update;
 pub use wallet::Wallet;
 
 /// Get the version of BDK at runtime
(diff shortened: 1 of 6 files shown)


notmandatory:
I'm moving this back to new alpha.4 release since it looks like a large-ish functional change.

LLFourn:
I think I've expressed directly to evan that I think this is a bad change. `TxGraph::missing_heights` is a useful method on its own. The names should be made consistent though.

LLFourn:
Haven't heard anything back from @evanlinjin  so closing this for now.

Issue #1101, opened 2023-08-29
Reintroduce descriptor data to `bdk::Wallet` persistence


evanlinjin:
**Describe the enhancement**

The current stable version of BDK stores the descriptor checksum in the database. This ensures the caller does not use the wrong descriptor with the database. We should reintroduce this.

**Implementation proposal**

Add `verify_checksum` method to the `Persist` trait.

LLFourn:
This is a good point. I seem to recall that things turned out this way because I didn't even plan on having any keychain data in the database in general and would leave that up to the application. This changed in order to simplify the lives of `Wallet` users.

Why don't we just persist the public descriptor in the database? We could even have this as part of the changeset for `KeychainTxOutIndex`. This would make sense since inserting a descriptor is a change you are making to the thing. This would remove the error case where you pass in a descriptor that doesn't match something in the database. Instead you just load from the database. New keychain descriptors can naturally be added later.

I seem to remember @afilini mentioning that it was to maintain some kind of forward privacy against an attacker who can read your database at some point (and then loses access). This would prevent them from learning the full descriptor at that point. I think this is a very weak guarantee:

1. You get to see all the spks at one point which lets you follow those coins forward. Using common input ownership heuristic and change heuristic you get a good idea about all future coins ownership.
2. The material that produces the descriptor is usually on disk anyway. If you have access to the database you probably got access to that as well.

If you want strong privacy against a database compromiser then there are two better choices:
    1. Don't store the data at all and just sync each time i.e. not persistence
    2. Encrypt the data using some secret. If you have a private descriptor (with secrets) then this is an obvious choice to derive a key from it and encrypt the database. A key derived from an application password would be good to. We could make this quite easy without much effort for `bdk_file_persist`.

notmandatory:
Had a little discussion about this with @evanlinjin and rather than persist the full descriptors I like the idea of offering a different trait for storing the descriptors, in case the wallet user wants to store those in a different more secure way. I still think we need the `verify_checksum` or should it be `verify_descriptor_checksums`? since we're confirming the persisted wallet data is for the right set of descriptors, currently 1 or 2, could be more in the future.

LLFourn:
- What's the need for a trait for this? Aren't you suggesting not storing descriptors and just letting the application provide them when creating the keychain txout index.
[...]

PR #1128, opened 2023-09-19
feat: add bdk_sqlite crate implementing PersistBackend

notmandatory:
### Description

Add "bdk_sqlite_store" crate implementing `PersistBackend` backed by a SQLite database.

### Notes to the reviewers

In addition to adding a SQLite based `PersistenceBackend` this PR also:
 
* add  `CombinedChangeSet` in `bdk_persist` and update `wallet` crate to use it. 
* updates the `wallet/tests` to also use this new sqlite store crate.
* updates `example-crates/wallet_esplora_async` to use this new sqlite store crate.
* fixes out of date README instructions for MSRV.
* bumps the ci `build_docs` job to rust `nightly-2024-05-12`.

### Changelog notice

Changed

- Update Wallet to use CombinedChangeSet for persistence.

Added

- Add  CombinedChangeSet in bdk_persist crate.
- Add bdk_sqlite crate implementing SQLite based PersistenceBackend storage for CombinedChangeSet.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/sqlite/Cargo.toml
+++ b/crates/sqlite/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "bdk_sqlite"
+version = "0.1.0"
+edition = "2021"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/bitcoindevkit/bdk"
+documentation = "https://docs.rs/bdk_sqlite"
+description = "A simple SQLite based implementation of Persist for Bitcoin Dev Kit."
+keywords = ["bitcoin", "persist", "persistence", "bdk", "sqlite"]
+authors = ["Bitcoin Dev Kit Developers"]
+readme = "README.md"
+
+[dependencies]
+anyhow = { version = "1", default-features = false }
+bdk_chain = { path = "../chain", version = "0.14.0", features = ["serde", "miniscript"] }
+bdk_persist = { path = "../persist", version = "0.2.0", features = ["serde"] }
+rusqlite = { version = "0.31.0", features = ["bundled"] }
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
--- a/crates/wallet/tests/wallet.rs
+++ b/crates/wallet/tests/wallet.rs
@@ -1,9 +1,12 @@
+use std::path::Path;
 use std::str::FromStr;
 
 use assert_matches::assert_matches;
 use bdk_chain::collections::BTreeMap;
 use bdk_chain::COINBASE_MATURITY;
 use bdk_chain::{BlockId, ConfirmationTime};
+use bdk_persist::PersistBackend;
+use bdk_sqlite::rusqlite::Connection;
 use bdk_wallet::descriptor::{calc_checksum, IntoWalletDescriptor};
 use bdk_wallet::psbt::PsbtUtils;
 use bdk_wallet::signer::{SignOptions, SignerError};
@@ -69,166 +72,198 @@ const P2WPKH_FAKE_WITNESS_SIZE: usize = 106;
 const DB_MAGIC: &[u8] = &[0x21, 0x24, 0x48];
 
 #[test]
-fn load_recovers_wallet() {
-    let temp_dir = tempfile::tempdir().expect("must create tempdir");
-    let file_path = temp_dir.path().join("store.db");
-
-    // create new wallet
-    let wallet_spk_index = {
-        let db = bdk_file_store::Store::create_new(DB_MAGIC, &file_path).expect("must create db");
-        let mut wallet = Wallet::new(get_test_tr_single_sig_xprv(), None, db, Network::Testnet)
-            .expect("must init wallet");
-
-        wallet.reveal_next_address(KeychainKind::External).unwrap();
-        wallet.spk_index().clone()
-    };
-
-    // recover wallet
+fn load_recovers_wallet() -> anyhow::Result<()> {
+    fn run<B, FN, FR>(filename: &str, create_new: FN, recover: FR) -> anyhow::Result<()>
+    where
+        B: PersistBackend<bdk_wallet::wallet::ChangeSet> + Send + Sync + 'static,
+        FN: Fn(&Path) -> anyhow::Result<B>,
+        FR: Fn(&Path) -> anyhow::Result<B>,
     {
-        let db = bdk_file_store::Store::open(DB_MAGIC, &file_path).expect("must recover db");
-        let wallet = Wallet::load(db).expect("must recover wallet");
-        assert_eq!(wallet.network(), Network::Testnet);
-        assert_eq!(
-            wallet.spk_index().keychains().collect::<Vec<_>>(),
-            wallet_spk_index.keychains().collect::<Vec<_>>()
-        );
-        assert_eq!(
-            wallet.spk_index().last_revealed_indices(),
-            wallet_spk_index.last_revealed_indices()
-        );
-        let secp = Secp256k1::new();
-        assert_eq!(
-            *wallet.get_descriptor_for_keychain(KeychainKind::External),
-            get_test_tr_single_sig_xprv()
-                .into_wallet_descriptor(&secp, wallet.network())
-                .unwrap()
-                .0
-        );
+        let temp_dir = tempfile::tempdir().expect("must create tempdir");
+        let file_path = temp_dir.path().join(filename);
+
+        // create new wallet
+        let wallet_spk_index = {
+            let db = create_new(&file_path).expect("must create db");
+            let mut wallet = Wallet::new(get_test_tr_single_sig_xprv(), None, db, Network::Testnet)
+                .expect("must init wallet");
+
+            wallet.reveal_next_address(KeychainKind::External).unwrap();
+            wallet.spk_index().clone()
+        };
+
+        // recover wallet
+        {
+            let db = recover(&file_path).expect("must recover db");
+            let wallet = Wallet::load(db).expect("must recover wallet");
+            assert_eq!(wallet.network(), Network::Testnet);
+            assert_eq!(
+                wallet.spk_index().keychains().collect::<Vec<_>>(),
+                wallet_spk_index.keychains().collect::<Vec<_>>()
+            );
+            assert_eq!(
+                wallet.spk_index().last_revealed_indices(),
+                wallet_spk_index.last_revealed_indices()
+            );
+            let secp = Secp256k1::new();
+            assert_eq!(
+                *wallet.get_descriptor_for_keychain(KeychainKind::External),
+                get_test_tr_single_sig_xprv()
+                    .into_wallet_descriptor(&secp, wallet.network())
+                    .unwrap()
+                    .0
+            );
+        }
+
+        // `new` can only be called on empty db
+        {
+            let db = recover(&file_path).expect("must recover db");
+            let result = Wallet::new(get_test_tr_single_sig_xprv(), None, db, Network::Testnet);
+            assert!(matches!(result, Err(NewError::NonEmptyDatabase)));
+        }
+
+        Ok(())
     }
 
-    // `new` can only be called on empty db
-    {
-        let db = bdk_file_store::Store::open(DB_MAGIC, &file_path).expect("must recover db");
-        let result = Wallet::new(get_test_tr_single_sig_xprv(), None, db, Network::Testnet);
-        assert!(matches!(result, Err(NewError::NonEmptyDatabase)));
-    }
+    run(
+        "store.db",
+        |path| Ok(bdk_file_store::Store::create_new(DB_MAGIC, path)?),
+        |path| Ok(bdk_file_store::Store::open(DB_MAGIC, path)?),
+    )?;
+    run(
+        "store.sqlite",
+        |path| Ok(bdk_sqlite::Store::new(Connection::open(path)?)?),
+        |path| Ok(bdk_sqlite::Store::new(Connection::open(path)?)?),
+    )?;
+
+    Ok(())
 }
 
 #[test]
-fn new_or_load() {
-    let temp_dir = tempfile::tempdir().expect("must create tempdir");
-    let file_path = temp_dir.path().join("store.db");
-
-    // init wallet when non-existent
-    let wallet_keychains: BTreeMap<_, _> = {
-        let db = bdk_file_store::Store::open_or_create_new(DB_MAGIC, &file_path)
-            .expect("must create db");
-        let wallet = Wallet::new_or_load(get_test_wpkh(), None, db, Network::Testnet)
-            .expect("must init wallet");
-        wallet.keychains().map(|(k, v)| (*k, v.clone())).collect()
-    };
-
-    // wrong network
+fn new_or_load() -> anyhow::Result<()> {
+    fn run<B, F>(filename: &str, new_or_load: F) -> anyhow::Result<()>
+    where
+        B: PersistBackend<bdk_wallet::wallet::ChangeSet> + Send + Sync + 'static,
+        F: Fn(&Path) -> anyhow::Result<B>,
     {
-        let db =
-            bdk_file_store::Store::open_or_create_new(DB_MAGIC, &file_path).expect("must open db");
-        let err = Wallet::new_or_load(get_test_wpkh(), None, db, Network::Bitcoin)
-            .expect_err("wrong network");
-        assert!(
-            matches!(
+        let temp_dir = tempfile::tempdir().expect("must create tempdir");
+        let file_path = temp_dir.path().join(filename);
+
+        // init wallet when non-existent
+        let wallet_keychains: BTreeMap<_, _> = {
+            let db = new_or_load(&file_path).expect("must create db");
+            let wallet = Wallet::new_or_load(get_test_wpkh(), None, db, Network::Testnet)
+                .expect("must init wallet");
+            wallet.keychains().map(|(k, v)| (*k, v.clone())).collect()
+        };
+
+        // wrong network
+        {
+            let db = new_or_load(&file_path).expect("must create db");
+            let err = Wallet::new_or_load(get_test_wpkh(), None, db, Network::Bitcoin)
+                .expect_err("wrong network");
+            assert!(
+                matches!(
+                    err,
+                    bdk_wallet::wallet::NewOrLoadError::LoadedNetworkDoesNotMatch {
+                        got: Some(Network::Testnet),
+                        expected: Network::Bitcoin
+                    }
+                ),
+                "err: {}",
                 err,
-                bdk_wallet::wallet::NewOrLoadError::LoadedNetworkDoesNotMatch {
-                    got: Some(Network::Testnet),
-                    expected: Network::Bitcoin
-                }
-            ),
-            "err: {}",
-            err,
-        );
-    }
+            );
+        }
 
-    // wrong genesis hash
-    {
-        let exp_blockhash = BlockHash::all_zeros();
-        let got_blockhash =
-            bitcoin::blockdata::constants::genesis_block(Network::Testnet).block_hash();
+        // wrong genesis hash
+        {
+            let exp_blockhash = BlockHash::all_zeros();
+            let got_blockhash =
+                bitcoin::blockdata::constants::genesis_block(Network::Testnet).block_hash();
 
-        let db =
-            bdk_file_store::Store::open_or_create_new(DB_MAGIC, &file_path).expect("must open db");
-        let err = Wallet::new_or_load_with_genesis_hash(
-            get_test_wpkh(),
-            None,
-            db,
-            Network::Testnet,
-            exp_blockhash,
-        )
-        .expect_err("wrong genesis hash");
-        assert!(
-            matches!(
+            let db = new_or_load(&file_path).expect("must open db");
+            let err = Wallet::new_or_load_with_genesis_hash(
+                get_test_wpkh(),
+                None,
+                db,
+                Network::Testnet,
+                exp_blockhash,
+            )
+            .expect_err("wrong genesis hash");
+            assert!(
+                matches!(
+                    err,
+                    bdk_wallet::wallet::NewOrLoadError::LoadedGenesisDoesNotMatch { got, expected }
+                    if got == Some(got_blockhash) && expected == exp_blockhash
+                ),
+                "err: {}",
                 err,
-                bdk_wallet::wallet::NewOrLoadError::LoadedGenesisDoesNotMatch { got, expected }
-                if got == Some(got_blockhash) && expected == exp_blockhash
-            ),
-            "err: {}",
-            err,
-        );
-    }
+            );
+        }
 
-    // wrong external descriptor
-    {
-        let exp_descriptor = get_test_tr_single_sig();
-        let got_descriptor = get_test_wpkh()
-            .into_wallet_descriptor(&Secp256k1::new(), Network::Testnet)
-            .unwrap()
-            .0;
+        // wrong external descriptor
+        {
+            let exp_descriptor = get_test_tr_single_sig();
+            let got_descriptor = get_test_wpkh()
+                .into_wallet_descriptor(&Secp256k1::new(), Network::Testnet)
+                .unwrap()
+                .0;
 
-        let db =
-            bdk_file_store::Store::open_or_create_new(DB_MAGIC, &file_path).expect("must open db");
-        let err = Wallet::new_or_load(exp_descriptor, None, db, Network::Testnet)
-            .expect_err("wrong external descriptor");
-        assert!(
-            matches!(
+            let db = new_or_load(&file_path).expect("must open db");
+            let err = Wallet::new_or_load(exp_descriptor, None, db, Network::Testnet)
+                .expect_err("wrong external descriptor");
+            assert!(
+                matches!(
+                    err,
+                    bdk_wallet::wallet::NewOrLoadError::LoadedDescriptorDoesNotMatch { ref got, keychain }
+                    if got == &Some(got_descriptor) && keychain == KeychainKind::External
+                ),
+                "err: {}",
                 err,
-                bdk_wallet::wallet::NewOrLoadError::LoadedDescriptorDoesNotMatch { ref got, keychain }
-                if got == &Some(got_descriptor) && keychain == KeychainKind::External
-            ),
-            "err: {}",
-            err,
-        );
-    }
+            );
+        }
 
-    // wrong internal descriptor
-    {
-        let exp_descriptor = Some(get_test_tr_single_sig());
-        let got_descriptor = None;
+        // wrong internal descriptor
+        {
+            let exp_descriptor = Some(get_test_tr_single_sig());
+            let got_descriptor = None;
 
-        let db =
-            bdk_file_store::Store::open_or_create_new(DB_MAGIC, &file_path).expect("must open db");
-        let err = Wallet::new_or_load(get_test_wpkh(), exp_descriptor, db, Network::Testnet)
-            .expect_err("wrong internal descriptor");
-        assert!(
-            matches!(
+            let db = new_or_load(&file_path).expect("must open db");
+            let err = Wallet::new_or_load(get_test_wpkh(), exp_descriptor, db, Network::Testnet)
+                .expect_err("wrong internal descriptor");
+            assert!(
+                matches!(
+                    err,
+                    bdk_wallet::wallet::NewOrLoadError::LoadedDescriptorDoesNotMatch { ref got, keychain }
+                    if got == &got_descriptor && keychain == KeychainKind::Internal
+                ),
+                "err: {}",
                 err,
-                bdk_wallet::wallet::NewOrLoadError::LoadedDescriptorDoesNotMatch { ref got, keychain }
-                if got == &got_descriptor && keychain == KeychainKind::Internal
-            ),
-            "err: {}",
-            err,
-        );
+            );
+        }
+
+        // all parameters match
+        {
+            let db = new_or_load(&file_path).expect("must open db");
+            let wallet = Wallet::new_or_load(get_test_wpkh(), None, db, Network::Testnet)
+                .expect("must recover wallet");
+            assert_eq!(wallet.network(), Network::Testnet);
+            assert!(wallet
+                .keychains()
+                .map(|(k, v)| (*k, v.clone()))
+                .eq(wallet_keychains));
+        }
+        Ok(())
     }
 
-    // all parameters match
-    {
-        let db =
-            bdk_file_store::Store::open_or_create_new(DB_MAGIC, &file_path).expect("must open db");
-        let wallet = Wallet::new_or_load(get_test_wpkh(), None, db, Network::Testnet)
-            .expect("must recover wallet");
-        assert_eq!(wallet.network(), Network::Testnet);
-        assert!(wallet
-            .keychains()
-            .map(|(k, v)| (*k, v.clone()))
-            .eq(wallet_keychains));
-    }
+    run("store.db", |path| {
+        Ok(bdk_file_store::Store::open_or_create_new(DB_MAGIC, path)?)
+    })?;
+    run("store.sqlite", |path| {
+        Ok(bdk_sqlite::Store::new(Connection::open(path)?)?)
+    })?;
+
+    Ok(())
 }
 
 #[test]
--- a/nursery/coin_select/src/bnb.rs
+++ b/nursery/coin_select/src/bnb.rs
@@ -305,341 +305,341 @@ where
     }?
 }
 
-#[cfg(all(test, feature = "miniscript"))]
-mod test {
-    use bitcoin::secp256k1::Secp256k1;
-
-    use crate::coin_select::{evaluate_cs::evaluate, ExcessStrategyKind};
-
-    use super::{
-        coin_select_bnb,
-        evaluate_cs::{Evaluation, EvaluationError},
-        tester::Tester,
-        CoinSelector, CoinSelectorOpt, Vec, WeightedValue,
-    };
-
-    fn tester() -> Tester {
-        const DESC_STR: &str = "tr(xprv9uBuvtdjghkz8D1qzsSXS9Vs64mqrUnXqzNccj2xcvnCHPpXKYE1U2Gbh9CDHk8UPyF2VuXpVkDA7fk5ZP4Hd9KnhUmTscKmhee9Dp5sBMK)";
-        Tester::new(&Secp256k1::default(), DESC_STR)
-    }
-
-    fn evaluate_bnb(
-        initial_selector: CoinSelector,
-        max_tries: usize,
-    ) -> Result<Evaluation, EvaluationError> {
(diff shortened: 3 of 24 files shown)


LLFourn:
hey @notmandatory. What do you think about the approach of just putting the the data in actual tables rather than serializing it in some blob? For reference the tables we'd have for this would be:

1. Raw transactions (txid, tx_binary)
2. Block heights and hashes  (height, hash) keyed by hash
3. Transaction index data. This is currently an integer per keychain for `KeychainTxOutIndex` which `Wallet` uses. This might also [include the descriptor in the future](https://github.com/bitcoindevkit/bdk/issues/1101#issuecomment-1701929749).

The advantage would be that the database would be intelligible on the sqlite cli and from other applications. The disadvantage would be that we couldn't use sqlite for just any type of changeset. You'd probably have some extra trait over changesets that applications would have to implement on their changesets that return the inner `indexed_tx_graph::ChangeSet` and the `local_chain::ChangeSet`.

notmandatory:
> hey @notmandatory. What do you think about the approach of just putting the the data in actual tables rather than serializing it in some blob? 

It's funny you should mention it, I started out with the multi tables/columns approach but thought maybe it wasn't in the spirit of the `PersistBackend` trait.  Also I wasn't sure it was worth the trouble if it could only be used with one `ChangeSet` type. But I'll give it another shot this week, probably in a different PR in case we decide to go back to this blobs approach.

LLFourn:
> > hey @notmandatory. What do you think about the approach of just putting the the data in actual tables rather than serializing it in some blob?
> 
> It's funny you should mention it, I started out with the multi tables/columns approach but thought maybe it wasn't in the spirit of the `PersistBackend` trait. Also I wasn't sure it was worth the trouble if it could only be used with one `ChangeSet` type. But I'll give it another shot this week, probably in a different PR in case we decide to go back to this blobs approach.

Sure if we do go blobs I think we should use `bincode` rather than JSON.

notmandatory:
FYI I rebased off of #1203 so only look at my commits.

notmandatory:
I've implemented the `PersistBackend` trait for  SQLite with the below change set type. Next step is to test it out with the examples.

```rust
struct ChangeSet<K: Ord + for<'de> Deserialize<'de> + Serialize>(
    local_chain::ChangeSet,
    indexed_tx_graph::ChangeSet<ConfirmationTimeHeightAnchor, keychain::ChangeSet<K>>,
);
```

EDIT: I updated PR to use `ConfirmationTimeHeightAnchor` instead of `ConfirmationHeightAnchor`.

crates/sqlite_store/src/store.rs
```diff
@@ -0,0 +1,1095 @@
...
+    // confirmation height u32
+    // confirmation time unix epoch seconds
+    // txid is transaction hash hex string (reversed)
+    "CREATE TABLE anchor
+    (block_hash TEXT NOT NULL REFERENCES block (hash),
+     confirmation_height INTEGER,
+     confirmation_time INTEGER,
+     txid TEXT NOT NULL REFERENCES tx(txid))
+    STRICT;",
+];
```
storopoli: What if we move this to `.sql` file?
We get syntax highlighting and maybe some IDE introspection and autocomplete?

crates/sqlite_store/src/store.rs
```diff
@@ -0,0 +1,1095 @@
...
+    // Outpoint vout
+    // TxOut value as SATs
+    // TxOut script consensus encoded
+    "CREATE TABLE txout
+    (txid TEXT NOT NULL,
+     vout INTEGER NOT NULL,
+     value INTEGER NOT NULL,
+     script BLOB NOT NULL,
+     PRIMARY KEY (txid, vout))
+    STRICT;",
```
storopoli: Aren't both of these the same type: `&indexed_tx_graph::ChangeSet<A, keychain::ChangeSet<K>>`?

What are the differences between these two?

Why not add them to the same table and add a column to differentiate them.
Them everything would be in the `tx` table?

crates/sqlite_store/src/store.rs
notmandatory: I'd love to get this out of the code but to include it in a .sql file I'd have to bring in another dependency that includes static files into the code and I'd rather not do that just for better IDE support.

crates/sqlite_store/src/store.rs
notmandatory: These two tables correspond to the `txs` and `txouts` fields of the `tx_graph::ChangeSet<A>` struct and as I understand it are not directly related. The way tx_graph module docs explain it is the `txs` are the "whole" transactions our wallet knows about, and `txouts` are the "partial" transactions where we only know the txid and it's outputs and amounts. 

In tx_graph.rs:
```
pub struct ChangeSet<A = ()> {
    /// Added transactions.
    pub txs: BTreeSet<Transaction>,
    /// Added txouts.
    pub txouts: BTreeMap<OutPoint, TxOut>,
    /// Added anchors.
    /// Added last-seen unix timestamps of transactions.
    pub last_seen: BTreeMap<Txid, u64>,
}
```

crates/sqlite_store/src/store.rs
storopoli: What about?
```rust
const MIGRATIONS: &str = include_str!("path/to/schema.sql");
```

crates/sqlite_store/src/store.rs
storopoli: Okay, I see! Thanks for the explanation!

crates/sqlite_store/src/store.rs
notmandatory: thanks will try this approach!

crates/sqlite_store/src/store.rs
```diff
@@ -0,0 +1,979 @@
...
+                        .map_err(WriteError::Sqlite)?;
+                    Ok(())
+                }
+            }
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Load blocks.
```
rustaceanrob: Load network (right?)

notmandatory:
I've rebased this PR on the latest version of #1203 and everything seems to be working fine in the new `sqlite_store` crate. There are still some build errors in other crates that I'll make sure are fixed after #1203 is merged and I rebase on master.

notmandatory:
Rebased on latest version #1203.

oleonardolima:
@notmandatory I got curious, what's the decision behind using `rusqlite` instead of something like `sqlx` ?

_I'm not sure if this has been discussed before, please let me know if I should check some previous issue/discussion._

crates/sqlite_store/src/schema.rs
ValuedMammal, line 1: I ran the wallet example once and queried the schema version and came up with 8, I think because the migration executes every statement while incrementing the version before updating it. I assume we want the schema version to start at either 0 or 1?

crates/sqlite_store/schema/schema_0.sql
```diff
@@ -0,0 +1,67 @@
...
+
+-- keychain is the json serialized keychain structure as JSONB,
+-- descriptor_id is a sha256::Hash id of the descriptor string w/o the checksum,
+-- descriptor is the complete descriptor string,
+-- last active index is a u32
+CREATE TABLE keychain
+(
+    keychain      BLOB PRIMARY KEY NOT NULL,
+    descriptor    TEXT UNIQUE      NOT NULL,
+    descriptor_id BLOB UNIQUE      NOT NULL,
```
evanlinjin: Because two different keychains can share the same descriptor, the `descriptor` and `descriptor_id` columns should not be `UNIQUE`.

Also, because the `descriptor_id` is derived from the `descriptor`, I'm wondering whether we need to store it?

crates/sqlite_store/schema/schema_0.sql
```diff
@@ -0,0 +1,67 @@
...
+-- network is the valid network for all other table data
+CREATE TABLE network
+(
+    name TEXT UNIQUE NOT NULL
+) STRICT;
+
+-- keychain is the json serialized keychain structure as JSONB,
+-- descriptor_id is a sha256::Hash id of the descriptor string w/o the checksum,
+-- descriptor is the complete descriptor string,
+-- last active index is a u32
```
ValuedMammal:
```suggestion
-- last revealed index is a u32
```

crates/sqlite_store/schema/schema_0.sql
ValuedMammal: descriptor_id could also be TEXT and displayed as a hex str the same way as txid and blockhash

crates/sqlite_store/schema/schema_0.sql
```diff
@@ -0,0 +1,67 @@
...
+-- Outpoint txid hash hex string (reversed)
+-- Outpoint vout
+-- TxOut value as SATs
+-- TxOut script consensus encoded
+CREATE TABLE txout
+(
+    txid   TEXT    NOT NULL,
+    vout   INTEGER NOT NULL,
+    value  INTEGER NOT NULL,
+    script BLOB    NOT NULL,
```
ValuedMammal: just an idea, the script pubkey could be TEXT and stored as hex or asm string

crates/sqlite_store/src/schema.rs
notmandatory: Oh ya I can see how this could be misleading, that "version" is just the index of the last line in the schema file that has been applied. Should I rename it to something else like "index" ? I don't think we need need a "version" id since the schema file is committed and versioned in source control.

crates/sqlite_store/schema/schema_0.sql
notmandatory: My thinking for using BLOBs for descriptor_id is that it's an internal value we don't log anywhere and that no one would need to plug into a block explorer, unlike txid which should be easy to lookup.

crates/sqlite_store/schema/schema_0.sql
notmandatory: Same as for descriptor_id I don't know how useful it would be for debugging to have a human readable version of the script. Is there some particular reason you had in mind for making it human readable?

crates/sqlite_store/src/schema.rs
ValuedMammal: so what's the version table used for? sorry I know you just tried to explain it

crates/sqlite_store/schema/schema_0.sql
ValuedMammal: maybe at a glance you could see if a script was OP_RETURN for example, but it probably makes more sense to just load the changesets into an app that can decode everything for you.

crates/sqlite_store/src/store.rs
```diff
@@ -0,0 +1,893 @@
...
+            .expect("create new memory db store");
+        // let mut store =
+        //     Store::<Keychain, ConfirmationTimeHeightAnchor>::new(Path::new("test_agg.sqlite"))
+        //         .expect("create new file db store");
+
+        store.write_changes(&changeset).expect("write changeset");
+
+        let agg_changeset = store.load_from_persistence().expect("aggregated changeset");
+
+        assert_eq!(agg_changeset, Some(changeset)); // TODO assert
```
ValuedMammal: Is there something else this TODO comment is referring to?

crates/sqlite_store/src/schema.rs
notmandatory: The way it works is all the sql statements for migrating (building or changing) the DB schema are loaded from the *.sql files into an array. The statements in the MIGRATIONS array and executed starting with the current version number (or 0 if no version table). If the last statement executed was 8th in the array then we're at schema version 8 and update the version table to that.  

Next time the app runs it loads the version table and gets 8, there are still only 0..8 statements in the migrations list so no  new migration statements are executed. 

If in the future we need to make some new DB schema changes, say add a new field or table, we add a new schema file (or append to the current file) the new statements we want to run. Next time we start the app we execute statements 9...n and then update the versions table again to n.

notmandatory:
> @notmandatory I got curious, what's the decision behind using `rusqlite` instead of something like `sqlx` ?
> 
> _I'm not sure if this has been discussed before, please let me know if I should check some previous issue/discussion._

The reason for using `rusqlite` is just simplicity and fewer dependencies since all I want is a sqlite store for use in mobile apps. The sqlx crate has alot of dependencies, uses async, and I think is more suited for server apps. 

But I think once we have this blocking/sqlite only store working its worth also creating a `bdk_sqlx_store` crate with async and at least postgres support for server apps.

crates/sqlite_store/src/store.rs
notmandatory: Good question, I think I put that there before I had the assert and forgot to remove it.  Removed.

crates/sqlite_store/schema/schema_0.sql
```diff
@@ -0,0 +1,65 @@
...
+) STRICT;
+
+-- txid is transaction hash hex string (reversed)
+-- whole_tx is a consensus encoded transaction,
+-- last seen is a u64 unix epoch seconds
+CREATE TABLE tx
+(
+    txid      TEXT PRIMARY KEY NOT NULL,
+    whole_tx  BLOB             NOT NULL,
+    last_seen INTEGER
```
evanlinjin: `last_seen` can be added for a `txid`, even when there is no full tx or partial txouts. Would making the `whole_tx` column `NULL`-able make sense?

ValuedMammal (approved):
ACK acf3834d0d0824c15a396940c3597c027180a0b7

I appreciate you taking the time to implement this and making it easy for reviewers. The approach to use private traits in store.rs looks like a straightforward solution with minimal headaches. I noticed the sqlite store `ChangeSet` doesn't implement `Default`, and during `read` the changesets aren't appended, but with this implementation that doesn't appear necessary.

storopoli (approved):
ACK acf3834d0d0824c15a396940c3597c027180a0b7

evanlinjin (changes requested):
Unfortunately I haven't had time to look through everything yet. Trying to get it done soon.

crates/sqlite_store/schema/schema_0.sql
```diff
@@ -0,0 +1,65 @@
...
+-- confirmation height u32
+-- confirmation time unix epoch seconds
+-- txid is transaction hash hex string (reversed)
+CREATE TABLE anchor
+(
+    block_hash          TEXT NOT NULL REFERENCES block (hash),
+    confirmation_height INTEGER,
+    confirmation_time   INTEGER,
+    txid                TEXT NOT NULL REFERENCES tx (txid)
+) STRICT;
```
evanlinjin: Would this duplicate anchors stored in the db? The entire `(anchor, txid)` should uniquely identify the row (according to my shallow understanding of SQL).

evanlinjin: Also, how important is it to store `confirmation_height` and `confirmation_time` as separate fields? We could make this more "generic" if we had an anchor-blob field that is a unique identifier and block_hash + txid.

crates/sqlite_store/schema/schema_0.sql
notmandatory: Oh ya good call on the `descriptor` (and `descriptor_id`) not being unique anymore. But I need `descriptor_id` when I'm persisting `txout_index::ChangeSet.last_revealed` which is keyed by `descriptor_id`, so I need to use the `descriptor_id` column to update the keychain(s) `last_revealed` values. See: `update_last_revealed()`.

crates/sqlite_store/schema/schema_0.sql
notmandatory: Another good catch, I changed `tx.whole_tx` to be `NULL`-able and changed `insert_txs()` and `update_last_seen()` to be "UPSERT" meaning they `INSERT` a new `tx` record if the txid doesn't exist, otherwise they `UPDATE` either the `whole_tx` or `last_seen`.  I also updated `select_txs()` to only select tx where  `tx.whole_tx` is not `NULL`.

crates/sqlite_store/schema/schema_0.sql
notmandatory: If the software is correct we shouldn't get duplicate `anchor` records, but the purpose of the DB constraints are mainly to catch software issues. So I added a `UNIQUE` constraint for the `anchor` table that includes all the `anchor` fields.

As for `anchor.confirmation_height` and `anchor.confirmation_time` columns I'd prefer to keep them separate to make debugging a little easier.  They're both human readable and if one or both aren't used (depending on anchor type) it doesn't really cost anything to leave them as `NULL`.

notmandatory: I can also see how making `anchor` a blob would make this DB schema work with any anchor type. But do you expect there to be any more types? Making things too generic makes it harder to reason about.

notmandatory:
I force pushed fixes mentioned above. I also updated the tests to persist multiple change sets with the scenarios:
* add a tx with last_seen but no whole_tx
* adding a whole_tx to the tx that only had a last seen
* add second anchors for existing tx1 and tx2

notmandatory:
> ACK [acf3834](https://github.com/bitcoindevkit/bdk/commit/acf3834d0d0824c15a396940c3597c027180a0b7)
> 
> I appreciate you taking the time to implement this and making it easy for reviewers. The approach to use private traits in store.rs looks like a straightforward solution with minimal headaches. I noticed the sqlite store `ChangeSet` doesn't implement `Default`, and during `read` the changesets aren't appended, but with this implementation that doesn't appear necessary.

@ValuedMammal thanks for mentioning the missing `Default`. I ended up having to add `Default` to the `ChangeSet` type after all to test with multiple changesets and in the process did find an issue where append wouldn't work when appending from `network = None` to `network = Some(xx)`.

crates/sqlite_store/schema/schema_0.sql
evanlinjin: I do expect more `Anchor` types, and I think it's not too much work to make it flexible. One thing I had in mind is SPV proof anchors. However, I expect that it will be a very useful thing for users to save any anchor type.

In terms of debugging persistence backends, the only thing this is required is to make sure an aggregation of persisted changesets (using `Append::append`) is the same as the changeset recovered from the persistence. Or did you mean for something else?

notmandatory: Ok I'll try reworking this with `Anchor` as a JSONB blob. After sleeping on it, I think this will probably simplify the code a lot. 

The type of debugging I was thinking about was for devs to be able to run queries on the DB to double check values they see from their app. But with JSONB we can still query sub-fields of the `Anchor` blob which should be fine.

notmandatory: @evanlinjin I pushed my changes to make `Anchor` a blob, it did simplify the code. Here's an example query that shows how someone can still query the `Anchor` data (if they know what they're looking for).
```sql
SELECT block_hash, txid, datetime(json_extract(anchor, '$.confirmation_time'), 'unixepoch') AS confirmation_time
FROM anchor_tx
ORDER BY confirmation_time;
```
```csv
block_hash,txid,confirmation_time
0000006504d5a5efe4bdbd86a7052a4dddcd467b35c2ebde3689099746d9f037,f3e362bec29de16bb261341d8b59b4bd89f0a18b30efec094c855296f0bfe0ec,2023-08-01 07:23:03
000000f8721a9c4e1d99b833bdf01174b0cfa9cff0f72926ceee5e08074d2c7c,923662af0add7264d65a7c05d0e1ecf8bb86e90ffe4ed7f9d1b4c3b86f78ed4c,2023-09-01 19:39:34
000000f4289f0a969dee95f3d23951b9b375f97a5f1545dd93e09032070af48b,ae8221bc9a61b435c915343b58035a007d25867457ec42386c4cb9794af0060f,2023-09-01 20:40:08
000000f4289f0a969dee95f3d23951b9b375f97a5f1545dd93e09032070af48b,04f454e23050f8da0d3931359ca3fa275121ae0f4167c7fb2a5d80376bbb142c,2023-09-01 20:40:08
000000f4289f0a969dee95f3d23951b9b375f97a5f1545dd93e09032070af48b,d1faf0bbbf5f0cd8366c5e2be944e30ab4af19134c1c97988ec45c6ad7bee339,2023-09-01 20:40:08
etc..
```

crates/sqlite_store/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+mod schema;
+pub mod store;
+#[cfg(feature = "wallet")]
+#[cfg_attr(docsrs, doc(cfg(feature = "wallet")))]
+pub mod wallet;
+
+use bdk_chain::bitcoin::Network;
+use bdk_chain::{indexed_tx_graph, keychain, local_chain, Anchor, Append};
+use serde::{Deserialize, Serialize};
+
```
evanlinjin: I think we should publicly re-export the `rusqlite` crate. Therefore the caller can re-use our version instead of explicitly including the dependency in `Cargo.toml`.

crates/sqlite_store/src/store.rs
```diff
@@ -0,0 +1,827 @@
...
+
+impl<K, A> Store<K, A>
+where
+    K: Ord + for<'de> Deserialize<'de> + Serialize + Send,
+    A: Anchor + for<'de> Deserialize<'de> + Serialize + Send,
+{
+    /// Creates a new store from a [`Path`].
+    ///
+    /// The file must be able to be opened with read and write permissions.
+    pub fn new<P: AsRef<Path>>(path: P) -> Result<Self, rusqlite::Error> {
```
evanlinjin: I think we should let the caller construct the `Connection` and we have `Connection` as the input. Not only could we remove the `new_memory` method, but also allows the caller to set custom flags. Or maybe there is an argument saying that this is dangerous in some way?

crates/sqlite_store/src/wallet.rs
```diff
@@ -0,0 +1,32 @@
...
+            .read()
+            .map_err(|e| anyhow!(e).context("unable to load wallet changes"))?;
+
+        Ok(changeset.map(|changeset| bdk_wallet::wallet::ChangeSet {
+            network: changeset.network,
+            chain: changeset.chain,
+            indexed_tx_graph: changeset.tx_graph,
+        }))
+    }
+}
```
evanlinjin: It is kind of annoying that we need to have a feature flag just to enable this... Maybe we should put the `wallet::ChangeSet` into `Persist`? Also, what if callers want to have `Store` for any `PersistBackend<C>`? I'll have a play around...

evanlinjin (commented):
This is looking really good.

evanlinjin:
Also, what do you think about naming the crate `bdk_sqlite`? The term `store` doesn't add any more meaning to the name.

crates/wallet/tests/wallet.rs
evanlinjin, line 1: Instead of ONLY testing with the sqlite store, I think it would be better to extract the main test logic into helper functions and have an input closure that creates the `PersistBackend`. Then we can run tests with both backends.

evanlinjin: I think by principle, we should never reduce what we are testing.

notmandatory: Makes sense, I'll work on it.

crates/sqlite_store/src/wallet.rs
notmandatory: Ya I agree the feature flag is annoying. I'm fine with moving `wallet::ChangeSet` into the persist crate since wallet always depends on persist. I'll give it a try. 

As for `Store` supporting any changeset type, I think the simplest approach is a new `impl PersistBackend<>` for each one. I don't expect there'd be too many of them and it's simple coding.

crates/sqlite_store/src/store.rs
notmandatory: I prefer keeping things simple and `new(path)` is the only one the users would generally need. For `new_memory()` as described in the docs it's primarily for testing. Also sqlite doesn't support sharing a connection so safer to create it internally. 

If we find a case where a user needs to tweek the connection flags then we can add a `new_with_flags(path,flags)` but for now I can't think of any reason they'd need to.

crates/wallet/tests/wallet.rs
evanlinjin: I can work on changing/adding this if you haven't started?

crates/wallet/src/types.rs
```diff
@@ -20,6 +20,7 @@ use serde::{Deserialize, Serialize};
 
 /// Types of keychains
 #[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, Hash, Ord, PartialOrd)]
+#[serde(tag = "type")]
```
evanlinjin: For some reason, this makes `bdk_file_store` fail. Looking into it.

evanlinjin: Okay, so [`bincode` encodes all enums as `u32`](https://github.com/bincode-org/bincode?tab=readme-ov-file#why-does-bincode-not-respect-repru8)...

evanlinjin: However, removing this breaks `bdk_sqlite_store` with some keychains/anchor types due to the following:

1. How the `json_extract()` method works in sqlite: https://sqlite.org/json1.html#jexb
2. We are storing keychains/anchors as JSONB

The `json_extract()` method may not always return JSON values, and may fail parsing by `serde_json`.

evanlinjin: We should just use `json()` instead of `json_extract()`.

crates/sqlite_store/src/store.rs
evanlinjin: Honestly, I feel like the current implementation is sacrificing a lot of flexibility for very little extra convenience.

crates/wallet/src/types.rs
notmandatory: Oh ya I should have documented that I ran into this issue with the non-struct `KeychainKind` enum. I'll try using `json()`, I think I tried it already but maybe not. Worst case we could store `keychain` as a simple json string for maximum compatibility, it just feels less cool.

crates/wallet/tests/wallet.rs
notmandatory: I haven't started working on fixing the testing yet, if you can do it (your) tomorrow that'd be great and I'll work on the other changes today.

crates/wallet/src/types.rs
evanlinjin: Fixed here: https://github.com/notmandatory/bdk/pull/7/commits/06565995723b2c8bfe0d2e9c3d51b19e6b9619da

Sorry the commit mixed other things in as well.

evanlinjin: Using JSONB seems to be a good idea. Fixed in the commit above. I tend to try things out while I review, it feels more thorough for me.

crates/wallet/tests/wallet.rs
evanlinjin: https://github.com/notmandatory/bdk/commit/06565995723b2c8bfe0d2e9c3d51b19e6b9619da

Done here!

crates/sqlite_store/src/wallet.rs
evanlinjin: Wait, I don't think moving the `wallet::ChangeSet` into `bdk_persist` makes sense. It will add `bdk_chain` dependencies to `bdk_persist` which will destroy the point of it.

evanlinjin: I'm okay with the feature flag for now, although I do have an idea of using transform closures.

```rust
struct TransformedStore<K, A, C> {
    store: Store<K, A>,
    transform: Box<dyn Fn(ChangeSet) -> C>,
    transform_back: Box<dyn Fn(C) -> ChangeSet>,
}

impl TransformedStore {
    fn new(store: Store) -> Self
    where
        C: From<ChangeSet> + Into<ChangeSet>,
    { todo!() }

    fn with_transform(store: Store, transform, transform_back) -> Self { todo!() }
}
```

evanlinjin: The above code does not compile, but I hope the point comes across haha.

evanlinjin:
I'm very happy with this codebase, thanks for spearheading this effort @notmandatory. Just the things I've mentioned above and I'll be happy to ACK.

crates/sqlite_store/src/store.rs
notmandatory: I merged your patch to use `Connection`; it doesn't hurt to give the extra flexibility and with your change the API is still easy to use.

crates/sqlite_store/src/wallet.rs
notmandatory: Can we just stick with the `wallet` feature flag for now?

notmandatory:
> Also, what do you think about naming the crate `bdk_sqlite`? The term `store` doesn't add any more meaning to the name.

~~On second thought can we leave the name `bdk_sqlite_store`? I started to rename it but then to be consistent I'd have to rename the crate folder to `crates/sqlite` which seems weird since it's not the `sqlite` project but the way to use `sqlite` to store `bdk` data.~~ I renamed it in a0084dca59ff5a8de86a34380afb9c6940df1c04.

crates/wallet/src/types.rs
notmandatory: I merged your commits, just had to fix some outdated docs on the last commit.

notmandatory:
I ran into a docs error that I fixed by removing the rust doc link in the README. Can add it back later after the crate is published but it's not that important.

nursery/coin_select/src/bnb.rs
evanlinjin, line 1: This file should be removed? We should be using `bdk_coin_select` imported from our new repo.

evanlinjin: Okay it's not removed yet, that's pretty annoying. This is out of scope for this PR though.

crates/sqlite_store/src/wallet.rs
evanlinjin: https://github.com/notmandatory/bdk/pull/8

I did a proof-of-concept here. However, this still requires a `wallet` feature flag. I'm thinking whether it will make sense to put `ChangeSet` structures in `bdk_chain`. Maybe we can put a versioned aggregate-`ChangeSet` structures here (instead of `bdk_wallet`).

evanlinjin:
~~While thinking about how we can do away with the `wallet` feature flag, I had a rethink about how we were structuring everything for the persistence stuff. Here are my new thoughts.~~

~~Initially, we moved `PersistBackend` out of `bdk_chain` and into a new crate (`bdk_persist`) because we did not want `bdk_file_store` to be dependent on `bdk_chain`. This will only make sense if we expect `bdk_file_store` to be used outside of the context of BDK (which I assume to be the case). The introduction of `bdk_persist` then became a side-effect because we do not expect `bdk_persist` to be used outside of the scope of BDK (so ideally, it would also be contained in `bdk_chain`).~~

~~I think this was the wrong approach. We should have separated `bdk_file_store` into 2 crates:~~
- ~~`flatfile_store` - This would live outside of the `bitcoindevkit/bdk` repo and not have any bdk dependencies. This can be used by other projects that do not require BDK.~~
- ~~`bdk_flatfile_store` - This will be part of `bitcoindevkit/bdk`.~~

~~With this new structure, we can do away with `bdk_persist` because `bdk_flatfile_store` will always be used in the context of BDK and we have `flatfile_store` that has no BDK dependencies.~~

~~In addition to this, we can have combined-`ChangeSet` structures (that combine changesets from `local_chain`, `indexed_tx_graph`, etc.) which we maintain within `bdk_chain`. These changesets can be used in `bdk_wallet` and `bdk_sqlite_store`. This will get rid of the `wallet` feature flag requirement here.~~

Edit: Okay I got this wrong. We made `bdk_persist` because we did not want `bdk_chain` to be dependent on `anyhow`. In this case, it would be fine to have `bdk_persist` depend on `bdk_chain`. We can contain combined-changeset structures within `bdk_persist`.

notmandatory:
I renamed

> Also, what do you think about naming the crate `bdk_sqlite`? The term `store` doesn't add any more meaning to the name.

We discussed the naming at our 17:00 UTC BDK call today and consensus agrees with the shorter naming so I made the change in a0084dca59ff5a8de86a34380afb9c6940df1c04.

notmandatory:
@evanlinjin I merged your latest PR commits but removed `DbCommitment` and the `transform()` logic.  

`bdk_sqlite` already depends on `bdk_persist` so I don't see any reason why  `Store` shouldn't use `persist::changesets::CombinedChangeSet` directly instead of trying to transform it to a `DbCommitment`. If this looks OK to you I'll squash and cleanup the commits.

evanlinjin:
ConceptACK for removing `DbCommitment` and transform stuff for now.

Although there is an argument that the transform stuff might be useful if the caller wants to have custom changesets? However, I guess they can impl PersistBackend themselves. Let's leave it out for now. 

Let's squash!

crates/sqlite/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+[dependencies]
+anyhow = { version = "1", default-features = false }
+bdk_chain = { path = "../chain", version = "0.14.0", features = ["serde", "miniscript"] }
+bdk_persist = { path = "../persist", version = "0.2.0", features = ["serde"] }
+rusqlite = { version = "0.31.0", features = ["bundled"] }
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+
+# optional
+bdk_wallet = { path = "../wallet", version = "1.0.0-alpha.11", optional = true }
```
evanlinjin: We shouldn't need this anymore right?

evanlinjin (commented):
Overall I'm quite happy how this turned out.

In terms of styling, is the use of traits just to "section" the code? If that is the case, I would have preferred alternative ways to do so (i.e. you can have multiple `impl Store {}` blocks with doc comments at the top of each block). Using traits makes it harder to read and adds weird quirks (i.e. the funky thing you needed to do with `db_transaction`).

I'll ACK once we get rid of `bdk_wallet` dependency from `bdk_sqlite`.

notmandatory:
@evanlinjin great styling tip, I redid as you suggested and it look much better (I don't know what I was thinking with the traits 🙃 ) and removed the wallet features, sorry I forgot to remove that last push.

I also reorganized and squashed commits so it should be ready to merge. Thanks for all the help!

notmandatory:
I amended/force pushed another good suggestion from @evanlinjin to impl `PersistBackend` generic over types that can be converted `From` and `Into` `CombinedChageSet`:
```rust
impl<K, A, C> bdk_persist::PersistBackend<C> for Store<K, A>
where
    K: Ord + for<'de> Deserialize<'de> + Serialize + Send,
    A: Anchor + for<'de> Deserialize<'de> + Serialize + Send,
    C: Clone + From<CombinedChangeSet<K, A>> + Into<CombinedChangeSet<K, A>>,
{
    fn write_changes(&mut self, changeset: &C) -> anyhow::Result<()> {
        self.write(&changeset.clone().into())
            .map_err(|e| anyhow::anyhow!(e).context("unable to write changes to sqlite database"))
    }

    fn load_from_persistence(&mut self) -> anyhow::Result<Option<C>> {
        self.read()
            .map(|c| c.map(Into::into))
            .map_err(|e| anyhow::anyhow!(e).context("unable to read changes from sqlite database"))
    }
}
```

evanlinjin (approved):
ACK 475c5024ecc5d288f35fabb8fab2b62dbf67653b

PR #1135, opened 2023-09-28
ci: Pin time, cc in order to keep MSRV

danielabrozzoni:
#### All Submissions:


notmandatory (approved):
I understand having to pin `time` since a new version came out a few days ago, but not sure why `cc` broke. But I did confirm without pinning `cc` the tests fail. 

ACK ea6411c6850eef4f6c171b3035b3d70a8b2a9dc5

PR #1139, opened 2023-09-30
feat: add blockchain traits module with esplora_async impl

notmandatory:
### Description

WIP to make a simpler API for users to do common sync operations.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bdk/src/blockchain/mod.rs
+++ b/crates/bdk/src/blockchain/mod.rs
@@ -1,0 +1,67 @@
+//! Blockchain
+//!
+//! This module provides traits that can be used to fetch [`Update`] data from the bitcoin
+//! blockchain and perform other common functions needed by a [`Wallet`] user. There is an async
+//! and blocking version of each trait.
+//!
+//! Also provides are implementations of these traits for the commonly used blockchain client protocols
+//! [Electrum], [Esplora] and [BitcoinCoreRPC]. Creators of new or custom blockchain clients should
+//! implement which ever of these traits that are applicable.
+//!
+//! [`Update`]: crate::wallet::Update
+//! [Electrum]: TBD
+//! [Esplora]: TBD
+//! [BitcoinCoreRPC]: TBD
+
+#[cfg(feature = "async")]
+mod async_traits;
+
+#[cfg(feature = "blocking")]
+mod blocking_traits;
+
+#[cfg(feature = "esplora_async")]
+mod esplora_async;
+
+// sync modes
+
+/// Defines the options when syncing spks with an Electrum or Esplora blockchain client.
+#[derive(Debug)]
+pub struct SpkSyncMode {
+    /// Sync all spks the wallet has ever derived
+    pub all_spks: bool,
+    /// Sync only derived spks that have not been used, only applies if `all_spks` is false
+    pub unused_spks: bool,
+    /// Sync wallet utxos
+    pub utxos: bool,
+    /// Sync unconfirmed transactions
+    pub unconfirmed_tx: bool,
+}
+
+impl Default for SpkSyncMode {
+    fn default() -> Self {
+        Self {
+            all_spks: false,
+            unused_spks: true,
+            utxos: true,
+            unconfirmed_tx: true,
+        }
+    }
+}
+
+// trait errors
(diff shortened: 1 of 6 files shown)


LLFourn:
Concept NACK.

1. This reintroduces the problem of keeping a lock over the Wallet while doing IO. Removing this is the main goal of `v1`. You can overcome this by taking in a `Arc<Mutex<Wallet>>`.
2. This introduces lots of dependencies to `bdk::Wallet` making long term MSRV policy difficult and we'd be back with different MSRVs per feature.
3. These traits are only appropriate for esplora and electrum i.e. spks style syncing. By having them it makes it seem as if we support this kind of syncing as a more first class citizen to others. My personal feeling is that we should push development of CBF as the best way to use the library when that line of development is accomplished.
4. If we really want some helper methods to simplify use make a trait in `bdk_chain` that `Wallet` (or `Arc<Mutex<Wallet>>`) can implement, then you can have DWIM methods in `bdk_esplora` and `bdk_electrum` that sync against any type that implements this trait. The trait would let you query (i) current local chain tip checkpoint and (ii) spks, txids etc of interest.

crates/bdk/src/blockchain/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+impl Default for SpkSyncMode {
+    fn default() -> Self {
+        Self {
+            all_spks: false,
+            unused_spks: true,
+            utxos: true,
+            unconfirmed_tx: true,
+        }
+    }
+}
```
realeinherjar, line 49: From our discussion in FFI bindings meeting: can this be an `enum`?

tnull:
I want to @LLFourn's points that re-introducing separate async/blocking traits might make it hard to switch between syncing methods during runtime. If these are introduced, I'd prefer if we could make sure all relevant traits are implemented for all syncing methods (i.e., also have async bitcoind/electrum implementations, for instance).

evanlinjin:
Is the purpose of introducing these traits to make it esier to implement FFI bindings?

notmandatory:
The initial reason I started looking into this, as @evanlinjin mentioned above, is so the mobile bindings and Rust wallet users don't need to rewrite the same sync logic themselves. Instead I thought I could just throw the the scan/sync logic from the `wallet_*` examples behind some trait impls. 

> Concept NACK.
> 
>     1. This reintroduces the problem of keeping a lock over the Wallet while doing IO. Removing this is the main goal of `v1`. You can overcome this by taking in a `Arc<Mutex<Wallet>>`.

Based on below I can see even if I could do what I'm trying to do without a Mutex it isn't the right approach.

>     2. This introduces lots of dependencies to `bdk::Wallet` making long term MSRV policy difficult and we'd be back with different MSRVs per feature.

Yes this is a problem and I agree we don't want to get back into all the MSRV and feature problems.

>     3. These traits are only appropriate for esplora and electrum i.e. spks style syncing. By having them it makes it seem as if we support this kind of syncing as a more first class citizen to others. My personal feeling is that we should push development of CBF as the best way to use the library when that line of development is accomplished.

Long term I'd like to see users using CBF too, but most users today need esplora/electrum so I want to make their lives as easy as possible.

>     4. If we really want some helper methods to simplify use make a trait in `bdk_chain` that `Wallet` (or `Arc<Mutex<Wallet>>`) can implement, then you can have DWIM methods in `bdk_esplora` and `bdk_electrum` that sync against any type that implements this trait. The trait would let you query (i) current local chain tip checkpoint and (ii) spks, txids etc of interest.

I'm going to close this PR, but I like your suggestion for how to add some sort of helper traits in the `bdk_chain` crate. I'll try noodling around with that in a new PR. Thanks for all the feedback !

LLFourn:
> I'm going to close this PR, but I like your suggestion for how to add some sort of helper traits in the `bdk_chain` crate. I'll try noodling around with that in a new PR. Thanks for all the feedback !

Awesome. I'd say start with something like this: https://github.com/bitcoindevkit/bdk/issues/1153#issuecomment-1752263555 which simplifies it down to three lines. Shouldn't need any traits either. That approach is better than what I was hinting at above.

PR #1146, opened 2023-10-03
fix: Remove dependency on unnecessary bdk_chain features

LLFourn:


danielabrozzoni (approved):
utACK 5e79b81a6a3f7012153fea469720fab153266479

Issue #1149, opened 2023-10-05
Release 0.29.0


notmandatory:
### Summary

This maintenance release updates our `rust-bitcoin` dependency to 0.30.x and fixes a wallet balance bug when a wallet has more than one coinbase transaction.

### Commit

213c270ab475e2afbf6863271550a537cd203c15

### Changelog

- Update rust-bitcoin to 0.30 #1071
- Fix a bug when syncing coinbase utxos on electrum #1090

### Checklist

Release numbering must follow [Semantic Versioning]. These steps assume the current `release/0.28`
branch **development** version is *0.28.2*.

#### On the day of the release

Create the `release/0.29` branch and set to the `0.29.0` version:

- [x] From `release/0.28` branch create a new `release/0.29` branch. Push the branch to upstream.
- [x] Create a new PR branch called `bump_dev_0_29`.
- [x] Bump the `bump_dev_0_29` branch to the next development MINOR+1 version.
  - Change the `Cargo.toml` version value to `0.29.0`.
  - Update the `CHANGELOG.md` file.
  - The commit message should be "Bump version to 0.29.0".
- [x] Create PR and merge the `bump_dev_0_29` branch to `release/0.29`.
  - Title PR "Bump version to 0.29.0".

Tag and publish new release:

- [x] Add a tag to the `HEAD` commit in the `release/0.29` branch.
  - The tag name should be `v0.29.0`
  - The first line of the tag message should be "Release 0.29.0".
  - In the body of the tag message put a copy of the **Summary** and **Changelog** for the release.
  - Make sure the tag is signed, for extra safety use the explicit `--sign` flag.
- [x] Wait for the CI to finish one last time.
- [x] Push the new tag to the `bitcoindevkit/bdk` repo.
- [x] Publish **all** the updated crates to crates.io.
- [x] Create the release on GitHub.
  - Go to "tags", click on the dots on the right and select "Create Release".
  - Set the title to `Release 0.29.0`.
  - In the release notes body put the **Summary** and **Changelog**.
  - Use the "+ Auto-generate release notes" button to add details from included PRs.
  - Until we reach a `1.0.0` release check the "Pre-release" box.
- [x] Make sure the new release shows up on [crates.io] and that the docs are built correctly on [docs.rs].
- [x] Announce the release, using the **Summary**, on Discord, Twitter and Nostr.
- [x] Celebrate 🎉

[Semantic Versioning]: https://semver.org/
[...]

PR #1150, opened 2023-10-05
Bump version to 0.29.0

notmandatory:
See #1149 

I also had to fix MSRV 1.57 version pinning in CI pipelines.


thunderbiscuit (approved):
ACK 213c270ab475e2afbf6863271550a537cd203c15. Dealing with the pinning for the MSRV looks like quite a pain...

Issue #1094, opened 2023-08-24
Add integration tests for electrum, esplora, bitcoind_rpc crates


notmandatory:
**Describe the enhancement**  

Basic blockchain integration tests need to be added to confirm scan and sync in different scenarios. 

**Use case**  

Example scan and sync test scenarios:

- single xkey or single key descriptors
- multisig xkey or single key descriptors
- multisig with mix of xkey and single key descriptors
- chain reorg
- synchronous or async

**Additional context**

Originally suggested by @danielabrozzoni  on discord: https://discord.com/channels/753336465005608961/753367451319926827/1144280548802572319

notmandatory:
To reduce duplicate code this should be done after bitcoindevkit/bdk#2098, with similar consolidated `TestEnv` for `electrum` and `bitcoind_rpc` crates.

danielabrozzoni:
One of my biggest doubt for this is how to test examples. Running `cargo test` automatically checks that examples build, but we also need to test that they work correctly (otherwise we might have problems such as bitcoindevkit/bdk#1125 or bitcoindevkit/bdk#1151).

We might decide to duplicate the examples code into our tests, but then we need to make sure that the two stay up to date one with the other.

nondiremanuel:
According to our previous discussion, for the moment I assign the beta release to both this and the bitcoindevkit/bdk#2098 issue and then if we can make it in time for alpha.3 they can be included

notmandatory:
Blockchain crates have good test coverage, see bitcoindevkit/bdk#906. We should open new issues for specific scenarios that need more testing.

PR #1152, opened 2023-10-06
fix(wallet_esplora): missing_heights uses the graph update

### LINKED ISSUE #1151: bdk + esplora not seeing an old transaction as confirmed
**Describe the bug**  
When syncing using bdk and esplora, we might not see that a certain tx is confirmed, and consider it unconfirmed forever

**To Reproduce**  
See https://github.com/sebastianmontero/bdk-esplora-test

**Expected behavior**  
bdk correctly seeing a transaction as confirmed when it is

**Build environment**  
 - BDK tag/commit: 4ee11aae12f5f5b636008195998833268d5edc66

danielabrozzoni:
Fixes #1151.

When wallet_esplora_* was used to sync a wallet containing a transaction confirmed some time ago (more than 10-15 blocks ago), the transaction would be stuck in an "unconfirmed" state forever.

At the first scan time, `update_local_chain` would just fetch the last 10 to 15 blocks (depending on the server used), and `tx_graph.missing_heights` wouldn't return the tx's confirmation block as it was called on the original, non-updated tx_graph.
So, after the first scan, we would have a transaction in memory with an anchor that doesn't exist in our local_chain, and try_get_chain_position would return unconfirmed.

When scanning again, missing_heights would find the missing anchor, but `update_local_chain` wouldn't include it as it's older than ASSUME_FINAL_DEPTH.

The missing block would be downloaded every time, but never included in the local_chain, and the transaction would remain unconfirmed forever.

Here we call missing_heights on the updated graph, so that it can correctly return the anchor height, and `update_local_chain` can fetch it and include it in the chain.

### Notes to the reviewers

I'm not sure if this is the right approach, so I'm opening this PR to gather feedback. I still need to add tests, I'll do so once we decide if this is the right way to go or not.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/example-crates/wallet_esplora_async/src/main.rs
+++ b/example-crates/wallet_esplora_async/src/main.rs
@@ -56,7 +56,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let (update_graph, last_active_indices) = client
         .scan_txs_with_keychains(keychain_spks, None, None, STOP_GAP, PARALLEL_REQUESTS)
         .await?;
-    let missing_heights = wallet.tx_graph().missing_heights(wallet.local_chain());
+    let missing_heights = update_graph.missing_heights(wallet.local_chain());
     let chain_update = client.update_local_chain(prev_tip, missing_heights).await?;
     let update = Update {
         last_active_indices,
--- a/example-crates/wallet_esplora_blocking/src/main.rs
+++ b/example-crates/wallet_esplora_blocking/src/main.rs
@@ -55,7 +55,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let (update_graph, last_active_indices) =
         client.scan_txs_with_keychains(keychain_spks, None, None, STOP_GAP, PARALLEL_REQUESTS)?;
-    let missing_heights = wallet.tx_graph().missing_heights(wallet.local_chain());
+    let missing_heights = update_graph.missing_heights(wallet.local_chain());
     let chain_update = client.update_local_chain(prev_tip, missing_heights)?;
     let update = Update {
         last_active_indices,


danielabrozzoni:
Cherry-picked cdfb91f30fa788fe4215f3176e97a347cfefeaba from #1041 for ci :(

realeinherjar (approved):
Good one.
Another MSRV break in CI...
tACK cbf0ad93972507af90b51ee3090ad7fac10bb14f (with `mempool.space` not `blockstream.info`)

evanlinjin:
This is clearly my oversight. Thanks for identifying and fixing it.

Can you rearrange the commits to that the MSRV fix comes before the `wallet_esplora..` commit? Thanks.

Will give this a test, ACK and merge tomorrow.

example-crates/wallet_esplora_async/src/main.rs
```diff
@@ -56,12 +56,18 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+
     let missing_heights = wallet.tx_graph().missing_heights(wallet.local_chain());
     let chain_update = client.update_local_chain(prev_tip, missing_heights).await?;
     let update = Update {
         last_active_indices,
-        graph: update_graph,
         chain: Some(chain_update),
+        ..Default::default()
     };
     wallet.apply_update(update)?;
```
notmandatory: Couldn't this be fixed internally in the `wallet.apply_update()` function? Either way I can also use this fix in #1139.

evanlinjin: The problem is we need two calls to esplora, and need to apply some changes to our wallet between them.

I think the most unaesthetic thing here is:

```rust
let missing_heights = wallet.tx_graph().missing_heights(wallet.local_chain());
```

Maybe we can create a `Wallet::missing_heights` method.

evanlinjin: My proof of concept: https://github.com/danielabrozzoni/bdk/pull/5

example-crates/wallet_esplora_async/src/main.rs
```diff
@@ -56,12 +56,18 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let (update_graph, last_active_indices) = client
         .scan_txs_with_keychains(keychain_spks, None, None, STOP_GAP, PARALLEL_REQUESTS)
         .await?;
+    let update = Update {
+        graph: update_graph,
+        ..Default::default()
+    };
+    wallet.apply_update(update)?;
+
     let missing_heights = wallet.tx_graph().missing_heights(wallet.local_chain());
```
LLFourn:
```suggestion
    let missing_heights = update_graph.missing_heights(wallet.local_chain());
```

example-crates/wallet_esplora_async/src/main.rs
LLFourn: But this unaesthetic thing is the bug. We're not meant to ask the wallet what's missing we're meant to ask the update.

This is a one line fix. See above. Please let me know if I'm mistaken.

example-crates/wallet_esplora_async/src/main.rs
danielabrozzoni: Ouch of course, this is the right way to do it! I'm updating the code.

danielabrozzoni:
I updated the code to match @LLFourn suggestion; other than being way cleaner, I later found out that my original solution was broken as well.

FYI, doing
```rust
    let update = Update {
        last_active_indices,
        graph: update_graph,
        chain: Some(chain_update),
    };
    wallet.apply_update(update)?;
```
yields a different result than
```rust
    let update = Update {
        graph: update_graph,
        ..Default::default()
    };
    wallet.apply_update(update)?;
    
    let update = Update {
        last_active_indices,
        chain: Some(chain_update),
        ..Default::default()
    };
    wallet.apply_update(update)?;
```

I haven't looked much into it, but when the graph update is added separately from the rest of the update, `list_unspent` gives a wrong result. I'm not sure if this is expected (it might be, but I certainly didn't expect it!), or a bug.

danielabrozzoni:
Regarding tests, we have an issue open for integration tests: #1094. Should I start to tackle it in this PR, or can we merge and think about it in the next milestone?

evanlinjin:
> I haven't looked much into it, but when the graph update is added separately from the rest of the update, list_unspent gives a wrong result. I'm not sure if this is expected (it might be, but I certainly didn't expect it!), or a bug.

We should always apply the index update first. When the index update and graph update is in the same `wallet::Update`, the logic of `Wallet::apply_update` applies the index update first. If you don't apply the index update first, the transactions in the graph may not be indexed properly (some outputs will not end up being tracked, even though it's spk is part of the keychain).

Calling `missing_heights` on the unapplied `TxGraph` should get the same result as calling it on the wallet's `TxGraph`. The difference is that the prior is more efficient (so the new changes are welcome).

evanlinjin:
As a side note, I realized there is no way to set the keychain index's lookahead in `Wallet`. This will pose a problem for block-by-block chain sources.

evanlinjin (approved):
ACK b1461f05d05a0b9cf056a9604032468064ca1a05

PR #1157, opened 2023-10-09
chore: pin `byteorder` and `webpki` to fix MSRV

evanlinjin:
* pin `byteorder` to 1.4.3
* pin `webpki` to 0.22.2
* it's tokio:1.33.0 that needs to be pinned to 1.29.1
* fix README MSRV comments


danielabrozzoni (approved):
utACK 1db3f87a4850c11b010ceafeba8a64ee0e54e022

PR #1158, opened 2023-10-09
doc(bdk): Clarify the absolute_fee docs

### LINKED ISSUE #1066: Absolute fee constraint in TxBuilder not being respected

danielabrozzoni:
Fixes #1066

#### All Submissions:

### DIFF
--- a/crates/bdk/src/wallet/tx_builder.rs
+++ b/crates/bdk/src/wallet/tx_builder.rs
@@ -188,6 +188,10 @@ impl<'a, D, Cs: CoinSelectionAlgorithm, Ctx: TxBuilderContext> TxBuilder<'a, D, 
     /// * btc/kvB (0.00001000 btc/kvB == 1 sats/vB) using FeeRate::from_btc_per_kvb
     /// * sats/kwu (250 sats/kwu == 1 sats/vB) using FeeRate::from_sat_per_kwu
     /// Default is 1 sat/vB (see min_relay_fee)
+    ///
+    /// Note that this is really a minimum feerate -- it's possible to
+    /// overshoot it slightly since adding a change output to drain the remaining
+    /// excess might not be viable.
     pub fn fee_rate(&mut self, fee_rate: FeeRate) -> &mut Self {
         self.params.fee_policy = Some(FeePolicy::FeeRate(fee_rate));
         self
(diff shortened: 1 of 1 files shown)


crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -198,6 +198,10 @@ impl<'a, D, Cs: CoinSelectionAlgorithm, Ctx: TxBuilderContext> TxBuilder<'a, D,
     /// If anyone sets both the fee_absolute method and the fee_rate method,
     /// the FeePolicy enum will be set by whichever method was called last,
     /// as the FeeRate and FeeAmount are mutually exclusive.
+    ///
+    /// None that this is a really a minimum absolute fee -- it's possible to
```
thunderbiscuit: I think this is a typo: `None => Note`

danielabrozzoni: Ooops, thanks for spotting it!

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -198,6 +198,10 @@ impl<'a, D, Cs: CoinSelectionAlgorithm, Ctx: TxBuilderContext> TxBuilder<'a, D,
     /// If anyone sets both the fee_absolute method and the fee_rate method,
     /// the FeePolicy enum will be set by whichever method was called last,
     /// as the FeeRate and FeeAmount are mutually exclusive.
+    ///
+    /// Note that this is a really a minimum absolute fee -- it's possible to
```
kaiwolfram:
```suggestion
    /// Note that this is really a minimum absolute fee -- it's possible to
```

danielabrozzoni: Ah! Turns out I had two different typos in the same line 😅  Fixed, thanks!

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -198,6 +198,10 @@ impl<'a, D, Cs: CoinSelectionAlgorithm, Ctx: TxBuilderContext> TxBuilder<'a, D,
     /// If anyone sets both the fee_absolute method and the fee_rate method,
     /// the FeePolicy enum will be set by whichever method was called last,
     /// as the FeeRate and FeeAmount are mutually exclusive.
+    ///
+    /// Note that this is really a minimum absolute fee -- it's possible to
+    /// overshoot it slightly since adding a change output to drain the remaining
+    /// excess might not be viable.
```
evanlinjin, line 208: Couldn't we say this about the feerate as well?

danielabrozzoni: Fair, I updated the `fee_rate` description, lmk if it makes sense to you

notmandatory (approved):
ACK 0ecc0280c032301dc638e75e54a8691d66d50274

New docs makes sense to me.

PR #1161, opened 2023-10-09
ref(hwi): Move hwi out of bdk

### LINKED ISSUE #872: Move hardware signer into its own crate
It has its own very bespoke python runtime dependency and needs hardware to test (if I understand correctly). It doesn't need to be in the main `bdk` crate and I think it's desirable not to have non cargo dependencies for tests in the main `bdk` crate.

danielabrozzoni:
Fixes #872

### Description

This commit moves the `hardwaresigner` outside of bdk and inside `bdk_hwi`

### Notes to the reviewers

There are currently two issues with the code:
- `TransactionSigner` dictates that `sign_transaction` must return a `SignerError` - but being `SignerError` defined inside of bdk, we can't modify it to include an hwi specific error! I don't know how we could fix this (other than getting rid of the trait altogether :)); for now I just added a `SignerError::Generic` variant, lmk if you have better ideas!
- The hwi tests used the bdk utils to get a funded wallet for testing, which aren't available in `bdk_hwi`, which made me realize - maybe we should expose them so that we can use them across our crates, and also our users can use them to test their code?
For now, I just left the test commented.

### Changelog notice

- The old `hardwaresigner` module has been moved out of `bdk` and inside `bdk_hwi`.

#### All Submissions:

#### New Features:

~~* [ ] I've added tests for the new feature~~

### DIFF
--- a/crates/bdk/src/wallet/signer.rs
+++ b/crates/bdk/src/wallet/signer.rs
@@ -80,6 +80,7 @@
 //! ```
 
 use crate::collections::BTreeMap;
+use alloc::string::String;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::cmp::Ordering;
@@ -162,16 +163,10 @@ pub enum SignerError {
     SighashError(sighash::Error),
     /// Miniscript PSBT error
     MiniscriptPsbt(MiniscriptPsbtError),
-    /// Error while signing using hardware wallets
-    #[cfg(feature = "hardware-signer")]
-    HWIError(hwi::error::Error),
-}
-
-#[cfg(feature = "hardware-signer")]
-impl From<hwi::error::Error> for SignerError {
-    fn from(e: hwi::error::Error) -> Self {
-        SignerError::HWIError(e)
-    }
+    /// To be used only by external libraries implementing [`InputSigner`] or
+    /// [`TransactionSigner`], so that they can return their own custom errors, without having to
+    /// modify [`SignerError`] in BDK.
+    External(String),
 }
 
 impl From<sighash::Error> for SignerError {
@@ -196,8 +191,7 @@ impl fmt::Display for SignerError {
             Self::InvalidSighash => write!(f, "Invalid SIGHASH for the signing context in use"),
             Self::SighashError(err) => write!(f, "Error while computing the hash to sign: {}", err),
             Self::MiniscriptPsbt(err) => write!(f, "Miniscript PSBT error: {}", err),
-            #[cfg(feature = "hardware-signer")]
-            Self::HWIError(err) => write!(f, "Error while signing using hardware wallets: {}", err),
+            Self::External(err) => write!(f, "{}", err),
         }
     }
 }
--- a/crates/hwi/src/signer.rs
+++ b/crates/hwi/src/signer.rs
@@ -1,0 +1,94 @@
+use bdk::bitcoin::bip32::Fingerprint;
+use bdk::bitcoin::psbt::PartiallySignedTransaction;
+use bdk::bitcoin::secp256k1::{All, Secp256k1};
+
+use hwi::error::Error;
+use hwi::types::{HWIChain, HWIDevice};
+use hwi::HWIClient;
+
+use bdk::signer::{SignerCommon, SignerError, SignerId, TransactionSigner};
+
+#[derive(Debug)]
+/// Custom signer for Hardware Wallets
+///
+/// This ignores `sign_options` and leaves the decisions up to the hardware wallet.
+pub struct HWISigner {
+    fingerprint: Fingerprint,
+    client: HWIClient,
+}
+
+impl HWISigner {
+    /// Create a instance from the specified device and chain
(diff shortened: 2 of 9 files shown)


nondiremanuel:
As per today's discussion in the Lib Team Call, this PR probably needs to be rebased after the ci fix of #1182

crates/hwi/src/signer.rs
```diff
@@ -0,0 +1,94 @@
...
+        &self,
+        psbt: &mut PartiallySignedTransaction,
+        _sign_options: &bdk::SignOptions,
+        _secp: &Secp256k1<All>,
+    ) -> Result<(), SignerError> {
+        psbt.combine(
+            self.client
+                .sign_tx(psbt)
+                .map_err(|e| {
+                    SignerError::Generic(format!("While signing with hardware wallet: {}", e))
```
ValuedMammal: would this work?

```suggestion
                    SignerError::HWIError(e.to_string())
```

ValuedMammal (commented):
ACK

I also agree it would be nice to move some common bdk test utils to a shared crate - similar to #1171

crates/hwi/src/signer.rs
danielabrozzoni: In order for this to work, we would need to add the `HWIError` variant inside of bdk, which is not really desirable since it wouldn't be used inside bdk but in an outside crate.

evanlinjin:
This PR moves `harwaresigner` into a separate crate, but the new crate still depends on `bdk`. Ideally, `bdk_hwi` should not depend on `bdk`. The best way to do this is to have the `Signer` traits in a separate crate as well - however, this is also non-ideal since those traits will most likely be scrapped in the future.

I would really love to have hwi as a separate crate. However, I don't think it makes sense to do it with the current state of `Wallet`.

Introducing `SignerError::Generic` is also non-ideal. Now the error variant for HWI is ambiguous.

In conclusion, I think we should try this again later down the line.

danielabrozzoni:
> Ideally, bdk_hwi should not depend on bdk

Why? We already have `rust_hwi` available if someone wants to use hardware wallets without using bdk. `bdk_hwi` is supposed to be for people that want to use hwi and bdk together, so I don't see why it shouldn't depend on bdk.

> Introducing SignerError::Generic is also non-ideal. Now the error variant for HWI is ambiguous.

True, but at the moment everyone trying to implement the Signer trait in a crate outside of bdk encounters the same problem, so I think we should find a way to fix this anyways (without having to wait for a complete refactoring on the signer/wallet, since that will take a while). We could add either good docs to `SignerError::Generic`, create a `SignerError::External` (for external crates), or maybe we could refactor the trait to return a generic `Box<dyn std::error::Error>` (I haven't tested this so it might not be feasible)

I agree that this code is very suboptimal given the state of the Wallet, but we won't refactor the Wallet until bdk 2.0; do we want to leave hwi inside bdk, untested, until then?

nondiremanuel:
@evanlinjin do you have any feedback on Daniela's last comment?

notmandatory:

danielabrozzoni:
Done, thanks!

crates/bdk/src/wallet/signer.rs
```diff
@@ -162,16 +163,8 @@ pub enum SignerError {
...
-    HWIError(hwi::error::Error),
-}
-
-#[cfg(feature = "hardware-signer")]
-impl From<hwi::error::Error> for SignerError {
-    fn from(e: hwi::error::Error) -> Self {
-        SignerError::HWIError(e)
-    }
+    /// A generic error
+    Generic(String),
```
notmandatory: I'm not a big fan of the `SignerError::Generic` since I just removed `bdk::Error::Generic` in #1028. But in this case I don't see an simple way around it. If it becomes annoying we can fix it in 2.0 release.

notmandatory (approved):
ACK f1124d4d37b90d3c5d263e5dd5a8bb3df1d4d214

notmandatory:
Can merge this one after the new CI issues are fixed with #1247 (after it's ACKd and merged).

evanlinjin (approved):
ACK f1124d4d37b90d3c5d263e5dd5a8bb3df1d4d214

I don't like `SignerError::Generic`, however I don't see a better option without taking too much time. We will change the API greatly for 2.0 anyway (as @danielabrozzoni mentioned: https://github.com/bitcoindevkit/bdk/pull/1161#issuecomment-1822340841).

danielabrozzoni:
Do you think it would be better  to rename `SignerError::Generic` to `SignerError::ExternalTraitImplementationError` or something better? This would make it clear that it's needed only for external libraries implementing the Signer trait.

notmandatory:
How about `SignerError::External` with doc that explains this is for external trait implementation errors? Or I'd be fine with the longer version too.

danielabrozzoni:
Thanks, I rebased and changed `Error::Generic` to `Error::External`. LMK if the docs look ok or if they can be improved!

notmandatory (approved):
reACK 105d70e9743f651d141c6e0d901d60d583afd4d9

PR #1163, opened 2023-10-10
fix(electrum): fixed chain sync issue

### LINKED ISSUE #1125: Calling `example_electrum sync` on an unconfirmed tx may result in the tx being stuck as unconfirmed eternally (according to the cli))

evanlinjin:
### Description

This replaces #1145 and is based on this comment: https://github.com/bitcoindevkit/bdk/pull/1145#discussion_r1351806136. I pushed forward with this as I think it is a really good idea to include this in our `alpha.2` milestone.

This may or may not fix #1125 (however, neither @LagginTimes or myself can replicate the issue @danielabrozzoni has been running into - check #1145 for convo). We can confirm whether it does once we have proper tests. However, these changes here obviously fix a logical flaw.

* Fixed a logic error in `construct_update_tip` that caused the local chain tip to always be a block behind the actual tip.
* Docs are added to clarify the `construct_update_tip` logic.
* `ASSUME_FINAL_DEPTH` is renamed to `MAX_REORG_DEPTH`.

### Notes to the reviewers

This is tested manually (we need to add a proper test framework in the near future). I can perform the following successfully:

* `example_electrum scan` can detect incoming transactions when unconfirmed, and detect that is becomes confirm in later calls.
* `exampl_electrum sync` can do that same.

### Changelog notice

* Fixed a logic error in `construct_update_tip` that caused the local chain tip to always be a block behind the actual tip.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/electrum/src/electrum_ext.rs
+++ b/crates/electrum/src/electrum_ext.rs
@@ -11,8 +11,8 @@ use std::{
     str::FromStr,
 };
 
-/// We assume that a block of this depth and deeper cannot be reorged.
-const ASSUME_FINAL_DEPTH: u32 = 8;
+/// We assume that a blocks deeper than this depth cannot be reorged.
+const MAX_REORG_DEPTH: u32 = 6;
 
 /// Represents updates fetched from an Electrum server, but excludes full transactions.
 ///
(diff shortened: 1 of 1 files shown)


danielabrozzoni (approved):
ACK 55ad9e840a9379a2562b3823c6c3d3ea19949fd1 - I'm still encountering the bug every now and then, but I still want to merge this PR as it fixes an obvious issue.

LLFourn:
See comment here: https://github.com/bitcoindevkit/bdk/pull/1145#discussion_r1353664906

I still think that the renaming here is embedding a mistaken concept. There is no `MAX_REORG_DEPTH` for this algorithm (it can withstand a reorg of any depth).

Also can we please have a test that reproduces the issue and fixes it in this PR. This does not block `alpha.2` IMO. You can release a new version of the electrum crate (patch release) after `alpha.2`.

evanlinjin:
I agree with @LLFourn's points. This can be continued on #1145

PR #1171, opened 2023-10-12
chore: extract `TestEnv` into separate crate

LagginTimes:
### Description

`TestEnv` is extracted into its own crate with `electrsd` support added so that `TestEnv` can also serve `esplora` and `electrum`.
The tests in the `esplora` crate have also been updated to use `TestEnv`.

The `tx_can_become_unconfirmed_after_reorg()` test in `test_electrum` suggests that the electrum issue where a reorged tx would be stuck at confirmed does not exist anymore.

### Notes to the reviewers

The code for `tx_can_become_unconfirmed_after_reorg()` was adapted from the same test in `bitcoind_rpc/test_emitter`. This electrum version of the test requires extra review, as I am uncertain if I used the API efficiently.

#### All Submissions:

### DIFF
--- a/.github/workflows/nightly_docs.yml
+++ b/.github/workflows/nightly_docs.yml
@@ -17,6 +17,8 @@ jobs:
         run: rustup update
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2.2.1
+      - name: Pin dependencies for MSRV
+        run: cargo update -p home --precise "0.5.5"
       - name: Build docs
         run: cargo doc --no-deps
         env:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
     "crates/esplora",
     "crates/bitcoind_rpc",
     "crates/hwi",
+    "crates/testenv",
     "example-crates/example_cli",
     "example-crates/example_electrum",
     "example-crates/example_esplora",
--- a/crates/bitcoind_rpc/Cargo.toml
+++ b/crates/bitcoind_rpc/Cargo.toml
@@ -19,7 +19,7 @@ bitcoincore-rpc = { version = "0.17" }
 bdk_chain = { path = "../chain", version = "0.11", default-features = false }
 
 [dev-dependencies]
-bitcoind = { version = "0.33", features = ["25_0"] }
+bdk_testenv = { path = "../testenv", default_features = false }
 anyhow = { version = "1" }
 
 [features]
--- a/crates/bitcoind_rpc/tests/test_emitter.rs
+++ b/crates/bitcoind_rpc/tests/test_emitter.rs
@@ -2,160 +2,14 @@ use std::collections::{BTreeMap, BTreeSet};
 
 use bdk_bitcoind_rpc::Emitter;
 use bdk_chain::{
-    bitcoin::{Address, Amount, BlockHash, Txid},
+    bitcoin::{Address, Amount, Txid},
     keychain::Balance,
     local_chain::{self, CheckPoint, LocalChain},
     Append, BlockId, IndexedTxGraph, SpkTxOutIndex,
 };
-use bitcoin::{
-    address::NetworkChecked, block::Header, hash_types::TxMerkleNode, hashes::Hash,
-    secp256k1::rand::random, Block, CompactTarget, OutPoint, ScriptBuf, ScriptHash, Transaction,
-    TxIn, TxOut, WScriptHash,
-};
-use bitcoincore_rpc::{
-    bitcoincore_rpc_json::{GetBlockTemplateModes, GetBlockTemplateRules},
-    RpcApi,
-};
-
-struct TestEnv {
-    #[allow(dead_code)]
-    daemon: bitcoind::BitcoinD,
-    client: bitcoincore_rpc::Client,
-}
-
-impl TestEnv {
-    fn new() -> anyhow::Result<Self> {
-        let daemon = match std::env::var_os("TEST_BITCOIND") {
-            Some(bitcoind_path) => bitcoind::BitcoinD::new(bitcoind_path),
-            None => bitcoind::BitcoinD::from_downloaded(),
-        }?;
-        let client = bitcoincore_rpc::Client::new(
-            &daemon.rpc_url(),
-            bitcoincore_rpc::Auth::CookieFile(daemon.params.cookie_file.clone()),
-        )?;
-        Ok(Self { daemon, client })
-    }
-
-    fn mine_blocks(
-        &self,
-        count: usize,
-        address: Option<Address>,
-    ) -> anyhow::Result<Vec<BlockHash>> {
-        let coinbase_address = match address {
-            Some(address) => address,
-            None => self.client.get_new_address(None, None)?.assume_checked(),
-        };
-        let block_hashes = self
-            .client
-            .generate_to_address(count as _, &coinbase_address)?;
-        Ok(block_hashes)
-    }
-
-    fn mine_empty_block(&self) -> anyhow::Result<(usize, BlockHash)> {
-        let bt = self.client.get_block_template(
-            GetBlockTemplateModes::Template,
-            &[GetBlockTemplateRules::SegWit],
-            &[],
-        )?;
-
-        let txdata = vec![Transaction {
-            version: 1,
-            lock_time: bitcoin::absolute::LockTime::from_height(0)?,
-            input: vec![TxIn {
-                previous_output: bitcoin::OutPoint::default(),
-                script_sig: ScriptBuf::builder()
-                    .push_int(bt.height as _)
-                    // randomn number so that re-mining creates unique block
-                    .push_int(random())
-                    .into_script(),
-                sequence: bitcoin::Sequence::default(),
-                witness: bitcoin::Witness::new(),
-            }],
-            output: vec![TxOut {
-                value: 0,
-                script_pubkey: ScriptBuf::new_p2sh(&ScriptHash::all_zeros()),
-            }],
-        }];
-
-        let bits: [u8; 4] = bt
-            .bits
-            .clone()
-            .try_into()
-            .expect("rpc provided us with invalid bits");
-
-        let mut block = Block {
-            header: Header {
-                version: bitcoin::block::Version::default(),
-                prev_blockhash: bt.previous_block_hash,
-                merkle_root: TxMerkleNode::all_zeros(),
-                time: Ord::max(bt.min_time, std::time::UNIX_EPOCH.elapsed()?.as_secs()) as u32,
-                bits: CompactTarget::from_consensus(u32::from_be_bytes(bits)),
-                nonce: 0,
-            },
-            txdata,
-        };
-
-        block.header.merkle_root = block.compute_merkle_root().expect("must compute");
-
-        for nonce in 0..=u32::MAX {
-            block.header.nonce = nonce;
-            if block.header.target().is_met_by(block.block_hash()) {
-                break;
-            }
-        }
-
-        self.client.submit_block(&block)?;
-        Ok((bt.height as usize, block.block_hash()))
-    }
-
-    fn invalidate_blocks(&self, count: usize) -> anyhow::Result<()> {
-        let mut hash = self.client.get_best_block_hash()?;
-        for _ in 0..count {
-            let prev_hash = self.client.get_block_info(&hash)?.previousblockhash;
-            self.client.invalidate_block(&hash)?;
-            match prev_hash {
-                Some(prev_hash) => hash = prev_hash,
-                None => break,
-            }
-        }
-        Ok(())
-    }
-
-    fn reorg(&self, count: usize) -> anyhow::Result<Vec<BlockHash>> {
-        let start_height = self.client.get_block_count()?;
-        self.invalidate_blocks(count)?;
-
-        let res = self.mine_blocks(count, None);
-        assert_eq!(
-            self.client.get_block_count()?,
-            start_height,
-            "reorg should not result in height change"
-        );
-        res
-    }
-
-    fn reorg_empty_blocks(&self, count: usize) -> anyhow::Result<Vec<(usize, BlockHash)>> {
-        let start_height = self.client.get_block_count()?;
-        self.invalidate_blocks(count)?;
-
-        let res = (0..count)
-            .map(|_| self.mine_empty_block())
-            .collect::<Result<Vec<_>, _>>()?;
-        assert_eq!(
-            self.client.get_block_count()?,
-            start_height,
-            "reorg should not result in height change"
-        );
-        Ok(res)
-    }
-
-    fn send(&self, address: &Address<NetworkChecked>, amount: Amount) -> anyhow::Result<Txid> {
-        let txid = self
-            .client
-            .send_to_address(address, amount, None, None, None, None, None, None)?;
-        Ok(txid)
-    }
-}
+use bdk_testenv::TestEnv;
+use bitcoin::{hashes::Hash, Block, OutPoint, ScriptBuf, WScriptHash};
+use bitcoincore_rpc::RpcApi;
 
 /// Ensure that blocks are emitted in order even after reorg.
 ///
@@ -166,17 +20,22 @@ impl TestEnv {
 #[test]
 pub fn test_sync_local_chain() -> anyhow::Result<()> {
     let env = TestEnv::new()?;
-    let (mut local_chain, _) = LocalChain::from_genesis_hash(env.client.get_block_hash(0)?);
-    let mut emitter = Emitter::new(&env.client, local_chain.tip(), 0);
+    let network_tip = env.rpc_client().get_block_count()?;
+    let (mut local_chain, _) = LocalChain::from_genesis_hash(env.rpc_client().get_block_hash(0)?);
+    let mut emitter = Emitter::new(env.rpc_client(), local_chain.tip(), 0);
 
-    // mine some blocks and returned the actual block hashes
+    // Mine some blocks and return the actual block hashes.
+    // Because initializing `ElectrsD` already mines some blocks, we must include those too when
+    // returning block hashes.
     let exp_hashes = {
-        let mut hashes = vec![env.client.get_block_hash(0)?]; // include genesis block
-        hashes.extend(env.mine_blocks(101, None)?);
+        let mut hashes = (0..=network_tip)
+            .map(|height| env.rpc_client().get_block_hash(height))
+            .collect::<Result<Vec<_>, _>>()?;
+        hashes.extend(env.mine_blocks(101 - network_tip as usize, None)?);
         hashes
     };
 
-    // see if the emitter outputs the right blocks
+    // See if the emitter outputs the right blocks.
     println!("first sync:");
     while let Some(emission) = emitter.next_block()? {
         let height = emission.block_height();
@@ -207,7 +66,7 @@ pub fn test_sync_local_chain() -> anyhow::Result<()> {
         "final local_chain state is unexpected",
     );
 
-    // perform reorg
+    // Perform reorg.
     let reorged_blocks = env.reorg(6)?;
     let exp_hashes = exp_hashes
         .iter()
@@ -216,7 +75,7 @@ pub fn test_sync_local_chain() -> anyhow::Result<()> {
         .cloned()
         .collect::<Vec<_>>();
 
-    // see if the emitter outputs the right blocks
+    // See if the emitter outputs the right blocks.
     println!("after reorg:");
     let mut exp_height = exp_hashes.len() - reorged_blocks.len();
     while let Some(emission) = emitter.next_block()? {
@@ -272,16 +131,25 @@ fn test_into_tx_graph() -> anyhow::Result<()> {
     let env = TestEnv::new()?;
 
     println!("getting new addresses!");
-    let addr_0 = env.client.get_new_address(None, None)?.assume_checked();
-    let addr_1 = env.client.get_new_address(None, None)?.assume_checked();
-    let addr_2 = env.client.get_new_address(None, None)?.assume_checked();
+    let addr_0 = env
+        .rpc_client()
+        .get_new_address(None, None)?
+        .assume_checked();
+    let addr_1 = env
+        .rpc_client()
+        .get_new_address(None, None)?
+        .assume_checked();
+    let addr_2 = env
+        .rpc_client()
+        .get_new_address(None, None)?
+        .assume_checked();
     println!("got new addresses!");
 
     println!("mining block!");
     env.mine_blocks(101, None)?;
     println!("mined blocks!");
 
-    let (mut chain, _) = LocalChain::from_genesis_hash(env.client.get_block_hash(0)?);
+    let (mut chain, _) = LocalChain::from_genesis_hash(env.rpc_client().get_block_hash(0)?);
     let mut indexed_tx_graph = IndexedTxGraph::<BlockId, _>::new({
         let mut index = SpkTxOutIndex::<usize>::default();
         index.insert_spk(0, addr_0.script_pubkey());
@@ -290,7 +158,7 @@ fn test_into_tx_graph() -> anyhow::Result<()> {
         index
     });
 
-    let emitter = &mut Emitter::new(&env.client, chain.tip(), 0);
+    let emitter = &mut Emitter::new(env.rpc_client(), chain.tip(), 0);
 
     while let Some(emission) = emitter.next_block()? {
         let height = emission.block_height();
@@ -306,7 +174,7 @@ fn test_into_tx_graph() -> anyhow::Result<()> {
     let exp_txids = {
         let mut txids = BTreeSet::new();
         for _ in 0..3 {
-            txids.insert(env.client.send_to_address(
+            txids.insert(env.rpc_client().send_to_address(
                 &addr_0,
                 Amount::from_sat(10_000),
                 None,
@@ -342,7 +210,7 @@ fn test_into_tx_graph() -> anyhow::Result<()> {
 
     // mine a block that confirms the 3 txs
     let exp_block_hash = env.mine_blocks(1, None)?[0];
-    let exp_block_height = env.client.get_block_info(&exp_block_hash)?.height as u32;
+    let exp_block_height = env.rpc_client().get_block_info(&exp_block_hash)?.height as u32;
     let exp_anchors = exp_txids
         .iter()
         .map({
@@ -386,10 +254,10 @@ fn ensure_block_emitted_after_reorg_is_at_reorg_height() -> anyhow::Result<()> {
 
     let env = TestEnv::new()?;
     let mut emitter = Emitter::new(
-        &env.client,
+        env.rpc_client(),
         CheckPoint::new(BlockId {
             height: 0,
-            hash: env.client.get_block_hash(0)?,
+            hash: env.rpc_client().get_block_hash(0)?,
         }),
         EMITTER_START_HEIGHT as _,
     );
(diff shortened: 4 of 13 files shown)


evanlinjin:
The commit message does not need `()`.

evanlinjin (changes requested):
We need to change the internals to use `electrsd` so the test env can serve `esplora` and `electrum` as well.

Can use https://github.com/bitcoindevkit/bdk/pull/979 as a reference to see what electrum-server methods we need.

LagginTimes:
> We need to change the internals to use `electrsd` so the test env can serve `esplora` and `electrum` as well.

I will look into this().

vladimirfomene:
I think having reorg functions for all backends will help. I had to use a hack from Steve to implement reorg for electrum. You should see the code in #979. Read more here: https://bitcoin.stackexchange.com/questions/114044/how-can-i-simulate-a-reorg-for-testing.

evanlinjin:
> I think having reorg functions for all backends will help. I had to use a hack from Steve to implement reorg for electrum. You should see the code in #979. Read more here: https://bitcoin.stackexchange.com/questions/114044/how-can-i-simulate-a-reorg-for-testing. 

But reorg in Bitcoind should also result in reorg elsewhere?

evanlinjin:
@vladimirfomene that's a great point.

https://github.com/bitcoindevkit/bdk/blob/release/0.29/src/testutils/blockchain_tests.rs

Maybe the old `testutils` is something to base this work on.

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -189,15 +43,23 @@ fn block_to_chain_update(block: &bitcoin::Block, height: u32) -> local_chain::Up
...
+    let tip = env.rpc_client().get_block_count()?;
+    let mut emitter = Emitter::from_height(env.rpc_client(), tip as u32);
 
     // mine some blocks and returned the actual block hashes
     let exp_hashes = {
-        let mut hashes = vec![env.client.get_block_hash(0)?]; // include genesis block
-        hashes.extend(env.mine_blocks(101, None)?);
+        let mut hashes = (0..=tip)
+            .map(|height| env.rpc_client().get_block_hash(height))
+            .collect::<Result<Vec<_>, _>>()?; // includes genesis block
```
evanlinjin: Can we leave a comment explaining why we need to do this? Thanks!

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -189,15 +43,23 @@ fn block_to_chain_update(block: &bitcoin::Block, height: u32) -> local_chain::Up
...
+        hashes.extend(env.mine_blocks(101 - tip as usize, None, &env.bitcoind)?);
         hashes
     };
 
+    (0..tip).for_each(|height| {
+        let changeset = BTreeMap::from([(height as u32, Some(exp_hashes[height as usize]))]);
+        local_chain.apply_changeset(&changeset);
+    });
```
evanlinjin: This also needs a comment saying why.

Nit: I would iterate over `exp_hashes` (using `.enumerate` and `.take`). I would also use the `LocalChain::from_blocks` method to construct the `local_chain` here instead.

crates/bitcoind_rpc/Cargo.toml
```diff
@@ -12,6 +12,7 @@ bitcoincore-rpc = { version = "0.17" }
 bdk_chain = { path = "../chain", version = "0.6", default-features = false }
 
 [dev-dependencies]
+testenv = { path = "../testenv", version = "0.1.0", default_features = false }
 bitcoind = { version = "0.33", features = ["25_0"] }
```
evanlinjin: We can get rid of this test dependency now.

crates/testenv/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        self.bitcoind.client.submit_block(&block)?;
+        Ok((bt.height as usize, block.block_hash()))
+    }
+
+    pub fn invalidate_blocks(
+        &self,
+        count: usize,
+        bitcoind: &electrsd::bitcoind::BitcoinD,
```
evanlinjin: Why do we need this as a parameter when we already have a `BitcoinD` in `self`?

Edit: Okay I see your intention here. Concept ACK, Approach NACK. Let's have two `bitcoinD`s in `TestEnv` instead.

crates/testenv/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+        let res = self.mine_blocks(count, None, bitcoind);
+        assert_eq!(
+            bitcoind.client.get_block_count()?,
+            start_height,
+            "reorg should not result in height change"
+        );
+        res
+    }
+
+    pub fn reorg_empty_blocks(&self, count: usize) -> anyhow::Result<Vec<(usize, BlockHash)>> {
```
evanlinjin, line 214: Don't we also need to make this method `electrs`-safe?

crates/testenv/src/lib.rs
evanlinjin, line 1: Please add documentation, which should include important context as to why some things are implemented in a certain way.

I.e. The requirement of https://github.com/bitcoindevkit/bdk/pull/1171#issuecomment-1760465589 would not be immediately apparent, so a note of this would be mandatory.

evanlinjin (changes requested):
Thanks for taking this forward. The code looks pretty good.

The documentation was a little lacking. It's important to communicate why things are implemented the way they are.

crates/testenv/src/lib.rs
LagginTimes: It seems that the simple electrsd reorg test completed without having to pass in `bitcoinD` references. I have removed the extraneous `bitcoind` parameters.

LagginTimes:
A simple electrsd reorg test was written and runs without having to pass in `bitcoinD` references to `mine_blocks()` and `reorg()`. I have therefore removed the extraneous `bitcoind` input parameters from `TestEnv`. Will have this commit sorted out tomorrow.

LagginTimes:
A `test_chain_sync()` test has been added that reproduces the issue described in #1125 and also confirms that the fix in #1145 at least partially fixes the problem (in that there may be additional bugs as seen by @danielabrozzoni).

crates/bitcoind_rpc/Cargo.toml
```diff
@@ -19,7 +19,7 @@ bitcoincore-rpc = { version = "0.17" }
 bdk_chain = { path = "../chain", version = "0.6", default-features = false }
 
 [dev-dependencies]
-bitcoind = { version = "0.33", features = ["25_0"] }
+testenv = { path = "../testenv", version = "0.1.0", default_features = false }
```
evanlinjin: Let's not include the `version` here.

crates/electrum/Cargo.toml
```diff
@@ -12,6 +12,11 @@ readme = "README.md"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-bdk_chain = { path = "../chain", version = "0.6.0", default-features = false }
+bdk_chain = { path = "../chain", version = "0.6.0", features = [ "miniscript" ] }
 electrum-client = { version = "0.18" }
 #rustls = { version = "=0.21.1", optional = true, features = ["dangerous_configuration"] }
+
+[dev-dependencies]
+testenv = { path = "../testenv", version = "0.1.0", default-features = false }
```
evanlinjin: Please remove the `version` field.

crates/electrum/Cargo.toml
```diff
@@ -12,6 +12,11 @@ readme = "README.md"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-bdk_chain = { path = "../chain", version = "0.6.0", default-features = false }
+bdk_chain = { path = "../chain", version = "0.6.0", features = [ "miniscript" ] }
```
evanlinjin: This will force the crate to use `bdk_chain` with default features and `"miniscript"` when it's not strictly needed.

Can you revert this and add a `bdk_chain` dependency under `[dev-dependencies]`?

crates/esplora/Cargo.toml
```diff
@@ -21,6 +21,9 @@ futures = { version = "0.3.26", optional = true }
 bitcoin = { version = "0.30.0", optional = true, default-features = false }
 miniscript = { version = "10.0.0", optional = true, default-features = false }
 
+[dev-dependencies]
+testenv = { path = "../testenv", version = "0.1.0", default_features = false }
```
evanlinjin: rm `version` field thanks!

crates/testenv/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "testenv"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+# For no-std, remember to enable the bitcoin/no-std feature
+bitcoin = { version = "0.30", default-features = false }
```
evanlinjin: `bdk_chain` already re-exports `bitcoin` so we don't need this.

crates/testenv/src/lib.rs
evanlinjin: Please add documentation to all public methods and structures.

evanlinjin (changes requested):

crates/bitcoind_rpc/Cargo.toml
LLFourn: Why wouldn't you include the version? I think you need it to publish on crates.io

crates/electrum/Cargo.toml
LLFourn: Why do we need miniscript to test this crate?

crates/bitcoind_rpc/Cargo.toml
evanlinjin: This is a dev dependency. So when we release a new version of `test_env` we won't have to change all the version fields of these dev dependencies. Less things to change for a release

crates/electrum/Cargo.toml
LagginTimes: Including `miniscript` was to get a very early version of `test_electrum` to work properly. It is no longer needed, but I forgot to remove it.

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -188,13 +42,17 @@ fn block_to_chain_update(block: &bitcoin::Block, height: u32) -> local_chain::Up
 #[test]
 pub fn test_sync_local_chain() -> anyhow::Result<()> {
     let env = TestEnv::new()?;
-    let (mut local_chain, _) = LocalChain::from_genesis_hash(env.client.get_block_hash(0)?);
-    let mut emitter = Emitter::new(&env.client, local_chain.tip(), 0);
+    let tip = env.rpc_client().get_block_count()?;
+    let (mut local_chain, _) = LocalChain::from_genesis_hash(env.rpc_client().get_block_hash(0)?);
+    let mut emitter = Emitter::new(env.rpc_client(), local_chain.tip(), 0);
 
     // mine some blocks and returned the actual block hashes
```
evanlinjin: nit

```suggestion
    // mine some blocks and returned the actual block hashes
    // because creating an instance of `ElectrsD` automatically mines blocks... we need to... 
```

crates/electrum/tests/test_electrum.rs
```diff
@@ -0,0 +1,238 @@
...
+            None,
+            None,
+            5,
+        )?;
+
+        let missing = relevant_txids.missing_full_txs(recv_graph.graph());
+        let graph_update =
+            relevant_txids.into_confirmation_time_tx_graph(&client, None, missing)?;
+        let _ = recv_chain.apply_update(chain_update);
+        let _ = recv_graph.apply_update(graph_update);
```
evanlinjin, line 177: Can we check the changeset here to see if a new anchor is added?

crates/testenv/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+                    .or_else(|| electrsd::bitcoind::downloaded_exe_path().ok())
+                    .expect(
+                "you need to provide an env var BITCOIND_EXE or specify a bitcoind version feature",
+                );
+                electrsd::bitcoind::BitcoinD::with_conf(
+                    bitcoind_exe,
+                    &electrsd::bitcoind::Conf::default(),
+                )
+            }
+        }?;
```
evanlinjin: We only need one ENV here. For consistency, use `BITCOIND_EXE` and scrap `TEST_BITCOIND`.

crates/testenv/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+                .assume_checked(),
+        };
+        let block_hashes = self
+            .bitcoind
+            .client
+            .generate_to_address(count as _, &coinbase_address)?;
+        Ok(block_hashes)
+    }
+
+    /// Mine an empty block.
```
evanlinjin:
```suggestion
    /// Mine a block that is guaranteed to be empty even with transactions in the mempool.
```

evanlinjin (changes requested):
Very close!

evanlinjin:
In terms of commit-hygiene, I think this PR should be at least two commits. One for extracting the `TestEnv`, one for introducing the `electrum` tests.

ValuedMammal (commented):
Light review ACK, hoping to do a deeper dive on `test_electrum.rs` soon.

update: tested Ack. I don't see any major issues. Great work @LagginTimes

update: made a comment on 53186e2a4d5c4a2e71580ea27679de99f5cc4782

crates/electrum/tests/test_electrum.rs
ValuedMammal: > Can we check the changeset here to see if a new anchor is added?

We would expect to see new anchors inserted into graph when the initially sent txns confirm, but what would be the expectation for anchors *after* reorg?

crates/electrum/tests/test_electrum.rs
```diff
@@ -0,0 +1,240 @@
...
+    // Sync up to tip.
+    wait_for_block(&env, &client)?;
+    let ElectrumUpdate {
+        chain_update,
+        relevant_txids,
+    } = client.sync(recv_chain.tip(), [spk_to_track.clone()], None, None, 5)?;
+
+    let missing = relevant_txids.missing_full_txs(recv_graph.graph());
+    let graph_update = relevant_txids.into_confirmation_time_tx_graph(&client, None, missing)?;
+    let _ = recv_chain.apply_update(chain_update);
```
ValuedMammal:
```suggestion
    let chain_changeset = recv_chain.apply_update(chain_update).unwrap();
    assert_eq!(chain_changeset.len(), REORG_COUNT);
```

This should be true, but right now `chain_changeset` only contains 8 elements. I think it's because we only fetch `CHAIN_SUFFIX_LENGTH` number of blocks (8) from electrum, which adds an extra wrinkle I'm not sure was intended.

If we set `REORG_COUNT` to 8, then this is resolved, and we won't find new anchors added post-reorg.

LagginTimes: Good catch. Fixed.

crates/electrum/tests/test_electrum.rs
```diff
@@ -0,0 +1,240 @@
...
+    // Sync up to tip.
+    wait_for_block(&env, &client)?;
+    let ElectrumUpdate {
+        chain_update,
+        relevant_txids,
+    } = client.sync(recv_chain.tip(), [spk_to_track], None, None, 5)?;
+
+    let missing = relevant_txids.missing_full_txs(recv_graph.graph());
+    let graph_update = relevant_txids.into_confirmation_time_tx_graph(&client, None, missing)?;
+    let _ = recv_chain.apply_update(chain_update);
```
ValuedMammal: I'm tempted to use `?` here, to check the Result of `LocalChain::apply_update` is ok. Not sure if you have an opinion, or if this matters. The same suggestion would apply to lines 190, 219 below.

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -166,17 +20,22 @@ impl TestEnv {
 #[test]
 pub fn test_sync_local_chain() -> anyhow::Result<()> {
     let env = TestEnv::new()?;
-    let (mut local_chain, _) = LocalChain::from_genesis_hash(env.client.get_block_hash(0)?);
-    let mut emitter = Emitter::new(&env.client, local_chain.tip(), 0);
+    let tip = env.rpc_client().get_block_count()?;
```
evanlinjin: Nit: 

```suggestion
    let network_tip = env.rpc_client().get_block_count()?;
```

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -166,17 +20,22 @@ impl TestEnv {
...
+    let (mut local_chain, _) = LocalChain::from_genesis_hash(env.rpc_client().get_block_hash(0)?);
+    let mut emitter = Emitter::new(env.rpc_client(), local_chain.tip(), 0);
 
-    // mine some blocks and returned the actual block hashes
+    // Mine some blocks and return the actual block hashes.
+    // We must first add the hashes of the genesis block and the block automatically generated from
+    // initializing `ElectrsD`.
```
evanlinjin: Nit: I think this is clearer?

```suggestion
    // Mine some blocks and return the actual block hashes.
    // Because initializing `ElectrsD` already mines some blocks, we must include those too when returning block hashes.
```

crates/electrum/tests/test_electrum.rs
```diff
@@ -0,0 +1,240 @@
...
+/// 1. Mine 101 blocks.
+/// 2. Send a tx.
+/// 3. Mine extra block to confirm sent tx.
+/// 4. Check [`Balance`] to ensure tx is confirmed.
+#[test]
+fn scan_detects_confirmed_tx() -> Result<()> {
+    const SEND_AMOUNT: Amount = Amount::from_sat(10_000);
+
+    let env = TestEnv::new()?;
+    let client = electrum_client::Client::new(env.electrsd.electrum_url.as_str())?;
```
evanlinjin, line 35: Now #1306 is merged, I think we can just use `env.electrum_client()`.

crates/electrum/tests/test_electrum.rs
```diff
@@ -0,0 +1,240 @@
...
+        if client.block_headers_pop()?.is_some() {
+            return Ok(());
+        }
+
+        if delay.as_millis() < 512 {
+            delay = delay.mul_f32(2.0);
+        }
+        std::thread::sleep(delay);
+    }
+}
```
evanlinjin: I think this can be a method on `TestEnv`. Maybe rename it to `wait_until_electrum_sees_block` or something more descriptive. It's also not immediately apparent why this is useful so please write some docs!

crates/testenv/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "testenv"
```
evanlinjin: Please rename to `bdk_testenv`.

crates/testenv/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+        let res = self.mine_blocks(count, None);
+        assert_eq!(
+            self.bitcoind.client.get_block_count()?,
+            start_height,
+            "reorg should not result in height change"
+        );
+        res
+    }
+
+    /// Reorg with a number of empty blocks of a given size `count`.
```
evanlinjin, line 213: Docs can be a bit more descriptive here. I.e. will mine empty blocks even though there are txs in the mempool. Or: refer to `mine_empty_blocks` for more.

evanlinjin (changes requested):
This is looking great. This is my initial round of reviews. I still need to look at the introduced electrum tests.

evanlinjin:
I think you will need to do a really annoying rebase :/ sorry about that!

notmandatory:

crates/electrum/tests/test_electrum.rs
evanlinjin: Getting the versions of the crates to match up right now is impossible. I think what you have is optimal for now.

evanlinjin:
I think git is being weird, can you re-rebase on master? Make sure your remote is up to date!

evanlinjin:
Also could you please add a simple README for `bdk_testenv`? Thanks!

crates/testenv/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        }?;
+
+        let electrsd = match std::env::var_os("ELECTRS_EXE") {
+            Some(env_electrs_exe) => electrsd::ElectrsD::new(env_electrs_exe, &bitcoind),
+            None => {
+                let mut electrsd_conf = electrsd::Conf::default();
+                electrsd_conf.http_enabled = true;
+                let electrs_exe = electrsd::downloaded_exe_path()
+                    .expect("electrs version feature must be enabled");
+                electrsd::ElectrsD::with_conf(electrs_exe, &bitcoind, &electrsd_conf)
```
ValuedMammal, line 46: If `ELECTRS_EXE` is Some, we're calling `new` with no `electrsd::Conf`. We should configure `with_conf` in both cases.

crates/testenv/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+    // Reset `electrsd` so that new blocks can be seen.
+    pub fn reset_electrsd(mut self) -> anyhow::Result<Self> {
+        let electrsd = match std::env::var_os("ELECTRS_EXE") {
+            Some(env_electrs_exe) => electrsd::ElectrsD::new(env_electrs_exe, &self.bitcoind),
+            None => {
+                let mut electrsd_conf = electrsd::Conf::default();
+                electrsd_conf.http_enabled = true;
+                let electrs_exe = electrsd::downloaded_exe_path()
+                    .expect("electrs version feature must be enabled");
+                electrsd::ElectrsD::with_conf(electrs_exe, &self.bitcoind, &electrsd_conf)
```
ValuedMammal, line 74: If `ELECTRS_EXE` is Some, we're calling `new` with no `electrsd::Conf`. We should configure `with_conf` in both cases.

crates/testenv/src/lib.rs
oleonardolima: I think that in the background it's already doing that, right ? 🤔 https://docs.rs/electrsd/0.25.0/src/electrsd/lib.rs.html#143-146

But I agree that a standard form would be better :)

storopoli:
I know that we're testing it in other crates.
But do we plan to expose this in the future?
If yes, I think we should add tests for the `test_env` crate (I know that this sounds like an oxymoron, but bear with me ...)

ValuedMammal:
Now that #1351 is open, I made LagginTimes/bdk#1 to test stop_gap in `ElectrumExt::full_scan`

Cargo.toml
```diff
@@ -7,7 +7,7 @@ members = [
     "crates/electrum",
     "crates/esplora",
     "crates/bitcoind_rpc",
-    "crates/hwi",
+    "crates/testenv",
```
danielabrozzoni, line 11: Why are we removing hwi from the Cargo.toml?

evanlinjin (approved):
ACK 29036977e11696007825b49c894c8b06d92337b2

Cargo.toml
evanlinjin: Damn I didn't catch this

evanlinjin: @LagginTimes is it because something is failing to build?

LagginTimes: I believe this may have been an accidental deletion as a result of a merge. Thanks for catching this.

.github/workflows/nightly_docs.yml
```diff
@@ -17,6 +17,8 @@ jobs:
         run: rustup update
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2.2.1
+      - name: Pin dependencies for MSRV
```
notmandatory, line 20: This shouldn't be needed. I was able to build the locally docs (`cargo doc --no-deps`) with `nightly-2022-12-14` without any errors.

crates/testenv/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "bdk_testenv"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+# For no-std, remember to enable the bitcoin/no-std feature
+# bitcoin = { version = "0.30", default-features = false }
```
notmandatory: Does this commented line need to be here? I don't see how it is related to the above comment.

notmandatory (commented):
Overall looks like a great improvement, related tests are much easier to read too. I just have one comment/question about a CI change.

.github/workflows/nightly_docs.yml
LagginTimes: Publish Nightly Docs / Build docs will unfortunately fail without this change.

> error: package `home v0.5.9` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.68.0-nightly
Either upgrade to rustc 1.70.0 or newer, or use
cargo update -p home@0.5.9 --precise ver
where `ver` is the latest version of `home` supporting rustc 1.68.0-nightly

storopoli (approved):
ACK e8dc8d3d7dd30aa3b0826b6ad50399a57a105491

.github/workflows/nightly_docs.yml
notmandatory: That's strange, but anyway not a big deal, more important to get the rest of this PR in. Thanks for checking.

crates/electrum/tests/test_electrum.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    Ok(())
+}
+
+/// Ensure that confirmed txs that are reorged become unconfirmed.
+///
+/// 1. Mine 101 blocks.
+/// 2. Mine 11 blocks with a confirmed tx in each.
+/// 3. Perform 11 separate reorgs on each block with a confirmed tx.
```
oleonardolima: @LagginTimes Is this documentation up-to-date, shouldn't it be 8 as it's on `REORG_COUNT` ?

oleonardolima: I mean, [here](https://github.com/bitcoindevkit/bdk/pull/1171/commits/ac4345462ef01c0d80210b6e4038483ce83b4685#diff-6febce892f6a2b6f549cd1fec8fcff9f5512bf62417685b073c53a6a0b252c6cR156-R160) you are creating only 8txs that will take part in the reorganization later, right?

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -744,7 +624,7 @@ fn mempool_during_reorg() -> anyhow::Result<()> {
         // emission.
         // TODO: How can have have reorg logic in `TestEnv` NOT blacklast old blocks first?
```
oleonardolima, line 625: @LagginTimes general question: after this PR is merged, do you think we'll be able to solve this TODO ?

oleonardolima (commented):
Overall looks pretty good!

I just left one minor comment about the documentation that I was uncertain about, and a general question for further discussion.

crates/bitcoind_rpc/tests/test_emitter.rs
evanlinjin: Unfortunately no. This can be done in a follow-up PR!

crates/electrum/tests/test_electrum.rs
LagginTimes: > I mean, [here](https://github.com/bitcoindevkit/bdk/pull/1171/commits/ac4345462ef01c0d80210b6e4038483ce83b4685#diff-6febce892f6a2b6f549cd1fec8fcff9f5512bf62417685b073c53a6a0b252c6cR156-R160) you are creating only 8txs that will take part in the reorganization later, right?

This is correct, the docs weren't updated at the time of the change to 8.

crates/esplora/Cargo.toml
```diff
@@ -21,6 +21,9 @@ futures = { version = "0.3.26", optional = true }
 bitcoin = { version = "0.30.0", optional = true, default-features = false }
 miniscript = { version = "10.0.0", optional = true, default-features = false }
 
+[dev-dependencies]
+bdk_testenv = { path = "../testenv", default_features = false }
+
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 electrsd = { version= "0.25.0", features = ["bitcoind_25_0", "esplora_a33e97e1", "legacy"] }
```
evanlinjin: Do we need to make the tests run in WASM? Because `electrsd` is a mandatory dependency of `bdk_testenv`.

notmandatory: Right now for CI we only do a `cargo check` for WASM and don't rerun the tests for that target. Since I haven't heard of any WASM issues that slipped through by not running the tests I don't think we need to change `TestEnv` to work with WASM.

crates/bitcoind_rpc/tests/test_emitter.rs
oleonardolima: Cool, thanks!

oleonardolima (approved):
ACK 04d0ab5a978fbfd45a5c795f9142ff4a9ba29ee3

notmandatory (approved):
ACK 04d0ab5a978fbfd45a5c795f9142ff4a9ba29ee3

Great job, this will be a big help with testing.

evanlinjin (approved):
ACK 7c9ba3cfc855dc0845476fe923dac4e47cb5c06a

PR #1172, opened 2023-10-12
Introduce block-by-block API to `bdk::Wallet` and add RPC wallet example

vladimirfomene:
### Description

Introduce block-by-block API for `bdk::Wallet`. A `wallet_rpc` example is added to demonstrate syncing `bdk::Wallet` with the `bdk_bitcoind_rpc` chain-source crate.

The API of `bdk_bitcoind_rpc::Emitter` is changed so the receiver knows how to connect to the block emitted.

### Changelog notice

Added
* `Wallet` methods to apply full blocks (`apply_block` and `apply_block_connected_to`) and a method to apply a batch of unconfirmed transactions (`apply_unconfirmed_txs`).
* `CheckPoint::from_block_ids` convenience method.
* `LocalChain` methods to apply a block header (`apply_header` and `apply_header_connected_to`).
* Test to show that `LocalChain` can apply updates that are shorter than original. This will happen during reorgs if we sync wallet with `bdk_bitcoind_rpc::Emitter`.

Fixed
* `InsertTxError` now implements `std::error::Error`.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -23,7 +23,9 @@ pub use bdk_chain::keychain::Balance;
 use bdk_chain::{
     indexed_tx_graph,
     keychain::{self, KeychainTxOutIndex},
-    local_chain::{self, CannotConnectError, CheckPoint, CheckPointIter, LocalChain},
+    local_chain::{
+        self, ApplyHeaderError, CannotConnectError, CheckPoint, CheckPointIter, LocalChain,
+    },
     tx_graph::{CanonicalTx, TxGraph},
     Append, BlockId, ChainPosition, ConfirmationTime, ConfirmationTimeHeightAnchor, FullTxOut,
     IndexedTxGraph, Persist, PersistBackend,
@@ -31,8 +33,8 @@ use bdk_chain::{
 use bitcoin::secp256k1::{All, Secp256k1};
 use bitcoin::sighash::{EcdsaSighashType, TapSighashType};
 use bitcoin::{
-    absolute, Address, Network, OutPoint, Script, ScriptBuf, Sequence, Transaction, TxOut, Txid,
-    Weight, Witness,
+    absolute, Address, Block, Network, OutPoint, Script, ScriptBuf, Sequence, Transaction, TxOut,
+    Txid, Weight, Witness,
 };
 use bitcoin::{consensus::encode::serialize, BlockHash};
 use bitcoin::{constants::genesis_block, psbt};
@@ -427,6 +429,55 @@ pub enum InsertTxError {
         tx_height: u32,
     },
 }
+
+impl fmt::Display for InsertTxError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            InsertTxError::ConfirmationHeightCannotBeGreaterThanTip {
+                tip_height,
+                tx_height,
+            } => {
+                write!(f, "cannot insert tx with confirmation height ({}) higher than internal tip height ({})", tx_height, tip_height)
+            }
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for InsertTxError {}
+
+/// An error that may occur when applying a block to [`Wallet`].
+#[derive(Debug)]
+pub enum ApplyBlockError {
+    /// Occurs when the update chain cannot connect with original chain.
+    CannotConnect(CannotConnectError),
+    /// Occurs when the `connected_to` hash does not match the hash derived from `block`.
+    UnexpectedConnectedToHash {
+        /// Block hash of `connected_to`.
+        connected_to_hash: BlockHash,
+        /// Expected block hash of `connected_to`, as derived from `block`.
+        expected_hash: BlockHash,
+    },
+}
+
+impl fmt::Display for ApplyBlockError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ApplyBlockError::CannotConnect(err) => err.fmt(f),
+            ApplyBlockError::UnexpectedConnectedToHash {
+                expected_hash: block_hash,
+                connected_to_hash: checkpoint_hash,
+            } => write!(
+                f,
+                "`connected_to` hash {} differs from the expected hash {} (which is derived from `block`)",
+                checkpoint_hash, block_hash
+            ),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for ApplyBlockError {}
 
 impl<D> Wallet<D> {
     /// Initialize an empty [`Wallet`].
@@ -2302,7 +2353,7 @@ impl<D> Wallet<D> {
         self.persist.commit().map(|c| c.is_some())
     }
 
-    /// Returns the changes that will be staged with the next call to [`commit`].
+    /// Returns the changes that will be committed with the next call to [`commit`].
     ///
     /// [`commit`]: Self::commit
     pub fn staged(&self) -> &ChangeSet
@@ -2325,6 +2376,86 @@ impl<D> Wallet<D> {
     /// Get a reference to the inner [`LocalChain`].
     pub fn local_chain(&self) -> &LocalChain {
         &self.chain
+    }
+
+    /// Introduces a `block` of `height` to the wallet, and tries to connect it to the
+    /// `prev_blockhash` of the block's header.
+    ///
+    /// This is a convenience method that is equivalent to calling [`apply_block_connected_to`]
+    /// with `prev_blockhash` and `height-1` as the `connected_to` parameter.
+    ///
+    /// [`apply_block_connected_to`]: Self::apply_block_connected_to
+    pub fn apply_block(&mut self, block: Block, height: u32) -> Result<(), CannotConnectError>
+    where
+        D: PersistBackend<ChangeSet>,
+    {
+        let connected_to = match height.checked_sub(1) {
+            Some(prev_height) => BlockId {
+                height: prev_height,
+                hash: block.header.prev_blockhash,
+            },
+            None => BlockId {
+                height,
+                hash: block.block_hash(),
+            },
+        };
+        self.apply_block_connected_to(block, height, connected_to)
+            .map_err(|err| match err {
+                ApplyHeaderError::InconsistentBlocks => {
+                    unreachable!("connected_to is derived from the block so must be consistent")
+                }
+                ApplyHeaderError::CannotConnect(err) => err,
+            })
+    }
+
+    /// Applies relevant transactions from `block` of `height` to the wallet, and connects the
+    /// block to the internal chain.
+    ///
+    /// The `connected_to` parameter informs the wallet how this block connects to the internal
+    /// [`LocalChain`]. Relevant transactions are filtered from the `block` and inserted into the
+    /// internal [`TxGraph`].
+    pub fn apply_block_connected_to(
+        &mut self,
+        block: Block,
+        height: u32,
+        connected_to: BlockId,
+    ) -> Result<(), ApplyHeaderError>
+    where
+        D: PersistBackend<ChangeSet>,
+    {
+        let mut changeset = ChangeSet::default();
+        changeset.append(
+            self.chain
+                .apply_header_connected_to(&block.header, height, connected_to)?
+                .into(),
+        );
+        changeset.append(
+            self.indexed_graph
+                .apply_block_relevant(block, height)
+                .into(),
+        );
+        self.persist.stage(changeset);
+        Ok(())
+    }
+
+    /// Apply relevant unconfirmed transactions to the wallet.
+    ///
+    /// Transactions that are not relevant are filtered out.
+    ///
+    /// This method takes in an iterator of `(tx, last_seen)` where `last_seen` is the timestamp of
+    /// when the transaction was last seen in the mempool. This is used for conflict resolution
+    /// when there is conflicting unconfirmed transactions. The transaction with the later
+    /// `last_seen` is prioritied.
+    pub fn apply_unconfirmed_txs<'t>(
+        &mut self,
+        unconfirmed_txs: impl IntoIterator<Item = (&'t Transaction, u64)>,
+    ) where
+        D: PersistBackend<ChangeSet>,
+    {
+        let indexed_graph_changeset = self
+            .indexed_graph
+            .batch_insert_relevant_unconfirmed(unconfirmed_txs);
+        self.persist.stage(ChangeSet::from(indexed_graph_changeset));
     }
 }
 
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -2385,7 +2385,7 @@ impl<D> Wallet<D> {
     /// with `prev_blockhash` and `height-1` as the `connected_to` parameter.
     ///
     /// [`apply_block_connected_to`]: Self::apply_block_connected_to
-    pub fn apply_block(&mut self, block: Block, height: u32) -> Result<(), CannotConnectError>
+    pub fn apply_block(&mut self, block: &Block, height: u32) -> Result<(), CannotConnectError>
     where
         D: PersistBackend<ChangeSet>,
     {
@@ -2416,7 +2416,7 @@ impl<D> Wallet<D> {
     /// internal [`TxGraph`].
     pub fn apply_block_connected_to(
         &mut self,
-        block: Block,
+        block: &Block,
         height: u32,
         connected_to: BlockId,
     ) -> Result<(), ApplyHeaderError>
--- a/crates/bitcoind_rpc/src/lib.rs
+++ b/crates/bitcoind_rpc/src/lib.rs
@@ -43,11 +43,13 @@ pub struct Emitter<'c, C> {
 }
 
 impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
-    /// Construct a new [`Emitter`] with the given RPC `client`, `last_cp` and `start_height`.
+    /// Construct a new [`Emitter`].
     ///
-    /// * `last_cp` is the check point used to find the latest block which is still part of the best
-    ///   chain.
-    /// * `start_height` is the block height to start emitting blocks from.
+    /// `last_cp` informs the emitter of the chain we are starting off with. This way, the emitter
+    /// can start emission from a block that connects to the original chain.
+    ///
+    /// `start_height` starts emission from a given height (if there are no conflicts with the
+    /// original chain).
     pub fn new(client: &'c C, last_cp: CheckPoint, start_height: u32) -> Self {
         Self {
             client,
@@ -127,13 +129,58 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
     }
 
     /// Emit the next block height and header (if any).
-    pub fn next_header(&mut self) -> Result<Option<(u32, Header)>, bitcoincore_rpc::Error> {
-        poll(self, |hash| self.client.get_block_header(hash))
+    pub fn next_header(&mut self) -> Result<Option<BlockEvent<Header>>, bitcoincore_rpc::Error> {
+        Ok(poll(self, |hash| self.client.get_block_header(hash))?
+            .map(|(checkpoint, block)| BlockEvent { block, checkpoint }))
     }
 
     /// Emit the next block height and block (if any).
-    pub fn next_block(&mut self) -> Result<Option<(u32, Block)>, bitcoincore_rpc::Error> {
-        poll(self, |hash| self.client.get_block(hash))
+    pub fn next_block(&mut self) -> Result<Option<BlockEvent<Block>>, bitcoincore_rpc::Error> {
+        Ok(poll(self, |hash| self.client.get_block(hash))?
+            .map(|(checkpoint, block)| BlockEvent { block, checkpoint }))
+    }
+}
+
+/// A newly emitted block from [`Emitter`].
+#[derive(Debug)]
+pub struct BlockEvent<B> {
+    /// Either a full [`Block`] or [`Header`] of the new block.
+    pub block: B,
+
+    /// The checkpoint of the new block.
+    ///
+    /// A [`CheckPoint`] is a node of a linked list of [`BlockId`]s. This checkpoint is linked to
+    /// all [`BlockId`]s originally passed in [`Emitter::new`] as well as emitted blocks since then.
+    /// These blocks are guaranteed to be of the same chain.
+    ///
+    /// This is important as BDK structures require block-to-apply to be connected with another
+    /// block in the original chain.
+    pub checkpoint: CheckPoint,
+}
+
+impl<B> BlockEvent<B> {
+    /// The block height of this new block.
+    pub fn block_height(&self) -> u32 {
+        self.checkpoint.height()
+    }
+
+    /// The block hash of this new block.
+    pub fn block_hash(&self) -> BlockHash {
+        self.checkpoint.hash()
+    }
+
+    /// The [`BlockId`] of a previous block that this block connects to.
+    ///
+    /// This either returns a [`BlockId`] of a previously emitted block or from the chain we started
+    /// with (passed in as `last_cp` in [`Emitter::new`]).
+    ///
+    /// This value is derived from [`BlockEvent::checkpoint`].
+    pub fn connected_to(&self) -> BlockId {
+        match self.checkpoint.prev() {
+            Some(prev_cp) => prev_cp.block_id(),
+            // there is no previous checkpoint, so just connect with itself
+            None => self.checkpoint.block_id(),
+        }
     }
 }
 
@@ -203,7 +250,7 @@ where
 fn poll<C, V, F>(
     emitter: &mut Emitter<C>,
     get_item: F,
-) -> Result<Option<(u32, V)>, bitcoincore_rpc::Error>
+) -> Result<Option<(CheckPoint, V)>, bitcoincore_rpc::Error>
 where
     C: bitcoincore_rpc::RpcApi,
     F: Fn(&BlockHash) -> Result<V, bitcoincore_rpc::Error>,
@@ -215,13 +262,14 @@ where
                 let hash = res.hash;
                 let item = get_item(&hash)?;
 
-                emitter.last_cp = emitter
+                let new_cp = emitter
                     .last_cp
                     .clone()
                     .push(BlockId { height, hash })
                     .expect("must push");
+                emitter.last_cp = new_cp.clone();
                 emitter.last_block = Some(res);
-                return Ok(Some((height, item)));
+                return Ok(Some((new_cp, item)));
             }
             PollResponse::NoMoreBlocks => {
                 emitter.last_block = None;
--- a/crates/chain/src/indexed_tx_graph.rs
+++ b/crates/chain/src/indexed_tx_graph.rs
@@ -224,20 +224,26 @@ where
     /// Irrelevant transactions in `txs` will be ignored.
     pub fn apply_block_relevant(
         &mut self,
-        block: Block,
+        block: &Block,
         height: u32,
     ) -> ChangeSet<A, I::ChangeSet> {
         let block_id = BlockId {
             hash: block.block_hash(),
             height,
         };
-        let txs = block.txdata.iter().enumerate().map(|(tx_pos, tx)| {
-            (
-                tx,
-                core::iter::once(A::from_block_position(&block, block_id, tx_pos)),
-            )
-        });
-        self.batch_insert_relevant(txs)
+        let mut changeset = ChangeSet::<A, I::ChangeSet>::default();
+        for (tx_pos, tx) in block.txdata.iter().enumerate() {
+            changeset.indexer.append(self.index.index_tx(tx));
+            if self.index.is_tx_relevant(tx) {
+                let txid = tx.txid();
+                let anchor = A::from_block_position(block, block_id, tx_pos);
+                changeset.graph.append(self.graph.insert_tx(tx.clone()));
+                changeset
+                    .graph
+                    .append(self.graph.insert_anchor(txid, anchor));
+            }
+        }
+        changeset
     }
 
     /// Batch insert all transactions of the given `block` of `height`.
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -39,6 +39,28 @@ impl CheckPoint {
         Self(Arc::new(CPInner { block, prev: None }))
     }
 
+    /// Construct a checkpoint from a list of [`BlockId`]s in ascending height order.
+    ///
+    /// # Errors
+    ///
+    /// This method will error if any of the follow occurs:
+    ///
+    /// - The `blocks` iterator is empty, in which case, the error will be `None`.
+    /// - The `blocks` iterator is not in ascending height order.
+    /// - The `blocks` iterator contains multiple [`BlockId`]s of the same height.
+    ///
+    /// The error type is the last successful checkpoint constructed (if any).
+    pub fn from_block_ids(
+        block_ids: impl IntoIterator<Item = BlockId>,
+    ) -> Result<Self, Option<Self>> {
+        let mut blocks = block_ids.into_iter();
+        let mut acc = CheckPoint::new(blocks.next().ok_or(None)?);
+        for id in blocks {
+            acc = acc.push(id).map_err(Some)?;
+        }
+        Ok(acc)
+    }
+
     /// Construct a checkpoint from the given `header` and block `height`.
     ///
     /// If `header` is of the genesis block, the checkpoint won't have a [`prev`] node. Otherwise,
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -370,12 +370,16 @@ impl LocalChain {
         Ok(changeset)
     }
 
-    /// Update the chain with a given [`Header`] and a `connected_to` [`BlockId`].
+    /// Update the chain with a given [`Header`] at `height` which you claim is connected to a existing block in the chain.
     ///
-    /// The `header` will be transformed into checkpoints - one for the current block and one for
-    /// the previous block. Note that a genesis header will be transformed into only one checkpoint
-    /// (as there are no previous blocks). The checkpoints will be applied to the chain via
-    /// [`apply_update`].
+    /// This is useful when you have a block header that you want to record as part of the chain but
+    /// don't necessarily know that the `prev_blockhash` is in the chain.
+    ///
+    /// This will usually insert two new [`BlockId`]s into the chain: the header's block and the
+    /// header's `prev_blockhash` block. `connected_to` must already be in the chain but is allowed
+    /// to be `prev_blockhash` (in which case only one new block id will be inserted).
+    /// To be successful, `connected_to` must be chosen carefully so that `LocalChain`'s [update
+    /// rules][`apply_update`] are satisfied.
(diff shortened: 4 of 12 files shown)


crates/bdk/src/wallet/mod.rs
```diff
@@ -1949,6 +1949,18 @@ impl<D> Wallet<D> {
         Ok(())
     }
 
+    /// Stages wallet changes in memory for later commit.
+    ///
+    /// To commit changes to the persistence backend, call [`commit`].
+    ///
+    /// [`commit`]: Self::commit
+    pub fn stage(&mut self, changeset: ChangeSet)
```
evanlinjin: I think it's better not to have this. Just call `self.persist.stage` at the end of method definitions.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1986,6 +1998,42 @@ impl<D> Wallet<D> {
...
+        assert_ne!(lookahead, 0, "lookahead must be greater than 0");
+        self.indexed_graph.index.set_lookahead_for_all(lookahead);
+    }
+
+    /// Insert all the block's relevant transactions into the IndexedTxGraph.
+    pub fn apply_block_relevant(
+        &mut self,
+        block: Block,
+        height: u32,
+    ) -> Result<ChangeSet, CannotConnectError> {
```
evanlinjin: Instead of returning a changeset, just stage it internally.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1986,6 +1998,42 @@ impl<D> Wallet<D> {
...
+    }
+
+    /// Batch insert unconfirmed transactions into the IndexedTxGraph.
+    /// Filtering out those that are not relevant.
+    ///
+    /// Read more here: [`self.indexed_graph.batch_insert_relevant_unconfirmed()`]
+    pub fn batch_insert_relevant_unconfirmed<'t>(
+        &mut self,
+        unconfirmed_txs: impl IntoIterator<Item = (&'t Transaction, u64)>,
+    ) -> indexed_tx_graph::ChangeSet<ConfirmationTimeAnchor, keychain::ChangeSet<KeychainKind>>
```
evanlinjin: Don't return changeset.

example-crates/wallet_rpc/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+use bdk_file_store::Store;
+use std::str::FromStr;
+
+const DB_MAGIC: &str = "bdk-rpc-wallet-example";
+const FALLBACK_HEIGHT: u32 = 2532323;
+const SEND_AMOUNT: u64 = 5000;
+const LOOKAHEAD: u32 = 20;
+const RPC_USER: &str = "bitcoin";
+const RPC_PASS: &str = "password";
+const RPC_URL: &str = "127.0.0.1:18332";
```
evanlinjin: Please don't hard-code RPC authentication and url. Either have environmental variables, or use the `bitcoind` crate.

evanlinjin (changes requested):

example-crates/wallet_rpc/src/main.rs
vladimirfomene: I think environment variables here make more sense.

vladimirfomene:
@evanlinjin apology for the delay on this one. I'm doing all necessary changes today.

evanlinjin:
Is the intention to have the user supply their own descriptor? Or are we hardcoding the descriptor?

vladimirfomene:
I think we should hardcode the descriptor as we do in other wallet example

crates/bdk/src/wallet/mod.rs
```diff
@@ -1986,6 +1986,49 @@ impl<D> Wallet<D> {
...
     }
+
+    /// Set lookahead for all keychains.
+    pub fn set_lookahead_for_all(&mut self, lookahead: u32) {
+        // Having a lookahead of 0, means if we sync the chain
+        // we would not find any update since we don't have any
+        // store scripts in our indexer.
+        assert_ne!(lookahead, 0, "lookahead must be greater than 0");
+        self.indexed_graph.index.set_lookahead_for_all(lookahead);
+    }
```
ValuedMammal, line 2409: Is there a nicer way to handle the case when `lookahead` is 0 besides panicking? If not, it should be mentioned in the function's doc comment, but we might also consider returning a `Result`, or setting some default value for lookahead.

example-crates/wallet_rpc/src/main.rs
```diff
@@ -0,0 +1,91 @@
...
+        Network::Testnet,
+    )?;
+
+    let address = wallet.get_address(AddressIndex::New);
+    println!("Generated Address: {}", address);
+
+    let balance = wallet.get_balance();
+    println!("Wallet balance before syncing: {} sats", balance.total());
+
+    let rpc_client = Client::new(&args[0], Auth::UserPass(args[1].clone(), args[2].clone()))?;
```
ValuedMammal: If I recall, `args[0]` refers to the command executable, and subsequent arguments begin at index 1. I was only able to run this example after incrementing each of the subscripts.

bonus: You could check for the correct number of command line args early in the program and remind the user if something is missing.

```rust
if args.len() < 6 {
    println!("Usage: wallet_rpc <RPC_URL> <RPC_USER> <RPC_PASS> <LOOKAHEAD> <FALLBACK_HEIGHT>");
    std::process::exit(1);
}
```

example-crates/wallet_rpc/src/main.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    let balance = wallet.get_balance();
+    println!("Wallet balance after syncing: {} sats", balance.total());
+
+    if balance.total() < SEND_AMOUNT {
+        println!(
+            "Please send at least {} sats to the receiving address",
+            SEND_AMOUNT
+        );
+        std::process::exit(0);
```
ValuedMammal: nit:
```suggestion
        std::process::exit(1);
```
otherwise, this is equivalent to `return Ok(())` afaik

ValuedMammal (commented):
I did successfully run the regtest example, and after modifying `wallet_rpc/src/main.rs` to correctly parse the arg vector, I ran the example on both testnet and signet. I had to play with setting an appropriate fallback height to find the wallet txs, and on one occasion the program sat hanging while applying blocks, but I didn't investigate why. I wonder if this example might also be adapted to run on regtest, but that can be saved for future work

crates/bdk/src/wallet/mod.rs
vladimirfomene: I think returning an error here is better. I have also added more comments about this in the docs.

example-crates/wallet_rpc/src/main.rs
vladimirfomene: I like the bonus

example-crates/wallet_rpc/src/main.rs
vladimirfomene: I used zero here instead of 1 because it is what is used in other examples. I have changed it to 1 though.

vladimirfomene:
@ValuedMammal, I have added a note for other reviewers to check if the program hangs while applying blocks.

example-crates/example_bitcoind_rpc_polling/README.md
```diff
@@ -0,0 +1,68 @@
...
+   ```
+   bitcoin-cli -datadir=/tmp/regtest/bitcoind -regtest -rpcuser=<your-rpc-username> -rpcpassword=<your-rpc-password> generatetoaddress 101 $BITCOIND_ADDRESS
+   ```
+6. Verify test bitcoind wallet balance.
+   ```
+   bitcoin-cli -rpcwallet="test" -datadir=/tmp/regtest/bitcoind -regtest -rpcuser=<your-rpc-username> -rpcpassword=<your-rpc-password> getbalances
+   ```
+7. Set descriptor env and get address from RPC CLI wallet.
+   ```
+   export DESCRIPTOR="wpkh(tprv8ZgxMBicQKsPfK9BTf82oQkHhawtZv19CorqQKPFeaHDMA4dXYX6eWsJGNJ7VTQXWmoHdrfjCYuDijcRmNFwSKcVhswzqs4fugE8turndGc/1/*)"
```
evanlinjin, line 36: Based on how the example is written, this is not used. However, I would find it useful if the example allows for providing a custom descriptor.

evanlinjin: @vladimirfomene can you please address this? Thank you

vladimirfomene: This README was not meant for the wallet example but the CLI example. I think we should prioritize ease of use of this example rather than customizability.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1986,6 +1986,55 @@ impl<D> Wallet<D> {
...
+        let mut changeset = ChangeSet::from(self.chain.apply_update(chain_update)?);
+        changeset.append(ChangeSet::from(
+            self.indexed_graph.apply_block_relevant(block, height),
+        ));
+        self.persist.stage(changeset);
+        Ok(())
+    }
+
+    /// Batch insert unconfirmed transactions into the IndexedTxGraph.
+    /// Filtering out those that are not relevant.
```
danielabrozzoni: nit: I would connect the two sentences using a `,`: 
```
/// Batch insert unconfirmed transactions into the IndexedTxGraph,
/// filtering out those that are not relevant.
```

example-crates/example_bitcoind_rpc_polling/README.md
danielabrozzoni: It's ok if the descriptor can't be customized, but I think that evan is right, if this variable is not used we should simply remove this line

danielabrozzoni (commented):
Looks good to me!

example-crates/example_bitcoind_rpc_polling/README.md
evanlinjin: ~@vladimirfomene if we are prioritizing ease of use, what do you think about using the `bitcoind` crate and just ran a regtest environment locally? This way it's a single command and we can print stuff out to stdout.~

~I'll let you decide what you think is most appropriate. However, at the very least, please make the README and the example executable consistent.~

---

Edit: @vladimirfomene sorry about the misunderstanding, I see what you mean now! this README is for the `example_cli` example! :sweat_smile:

danielabrozzoni: Ahhh gotcha! Resolving.

danielabrozzoni (approved):
ACK d97875b9151e7155450b823c6d98321cf933d890

darosior (commented):
I would suggest the wallet use a default lookahead limit other than `0`. I expect it would be quite surprising to the user of the high-level `Wallet` interface to see that he has to explicitly call `set_lookahead` on the wallet for it to be able to notice a utxo received on next_deriv_index + 1.

Doesn't need to be in this PR, but wanted to raise it here since it's tightly related to the functionality introduced.

I think `1_000` would be a good default. This is what the Bitcoin Core wallet uses.

EDIT: this can actually be tackled in parallel of this PR. I've done so here: https://github.com/bitcoindevkit/bdk/pull/1229.

crates/bdk/src/wallet/mod.rs
```diff
@@ -2312,6 +2312,55 @@ impl<D> Wallet<D> {
...
+        block: Block,
+        height: u32,
+    ) -> Result<(), CannotConnectError>
+    where
+        D: PersistBackend<ChangeSet>,
+    {
+        let chain_update = CheckPoint::from_header(&block.header, height).into_update(false);
+        let mut changeset = ChangeSet::from(self.chain.apply_update(chain_update)?);
+        changeset.append(ChangeSet::from(
+            self.indexed_graph.apply_block_relevant(block, height),
```
darosior: Not saying this should be changed, but worth pointing out: internally `apply_block_relevant` will use `apply_batch_relevant`. The latter can be used on any batch and therefore can't make any assumption on the topology, for this reason it loops twice on the list of transactions (once for updating the spk index, once for updating the tx graph). In the case of valid blocks, where transactions must be topologically ordered, we can just use a single loop.

Not a huge optimization, but still, avoids a few thousands iterations.

crates/bdk/src/wallet/mod.rs
```diff
@@ -2312,6 +2312,55 @@ impl<D> Wallet<D> {
...
+            return Err(Error::Generic(
+                "lookahead must be greater than 0".to_string(),
+            ));
+        }
+        self.indexed_graph.index.set_lookahead_for_all(lookahead);
+        Ok(())
+    }
+
+    /// Insert all the block's relevant transactions into the IndexedTxGraph.
+    pub fn apply_block_relevant(
```
LLFourn: My main concern with this PR is we don't have a plan for the suite of methods we'll need as I talked about over in https://github.com/vladimirfomene/bdk/pull/2#issuecomment-1823605394

If we don't implement them in this PR we should at least have a plan so we don't have to rename things later and do proper documentation for this method.

- `apply_block(block, height)` applies the block when the previous block is in the wallet. Otherwise errors.
- `apply_block_connected_to(block, height, connected_to)` applies the block assuming that the previous block hash of the block is in a descendant of `connected_to`. Will error out of `connected_to` is not in the wallet (it's fine for `connected_to` to be the previous block hash -- then this is the same as `apply_block`).
- `apply_block_assume_connected(block, height)` -- first calls `apply_block` internally but if that was not connected then assumes that the block does connect to the existing chain and so finds the right spot.

Does this make sense? If so what we have to do here is improve the documentation for this method and add these other methods in another PR.

Another question is how these fit in with the "batch insert" methods. Can we expose those to wallet users in addition to (or instead of) the above?

crates/bdk/src/wallet/mod.rs
LLFourn: Sure. Worth a comment that we could make an assumption on the topology of the order of the transactions to make a bespoke implementation.

notmandatory:
Please rebase now that bitcoindevkit/bdk#1183 has been merged.

evanlinjin:
### Requirements

1. Ability to **ONLY** insert *relevant* checkpoints. *Relevant* means checkpoints to blocks which
contain relevant transactions. This is important for doing block-by-block syncing. I.e. full node
without CBF, a CBF node (we still want to filter out false-positives), and silent payments (for the 
future).

2. Ability for the block-source to handle reorgs (mid-sync) without requesting data from 
`bdk::Wallet`.

### Why `apply_block_connected_to` and `apply_block_assume_connected` Does Not Satisfy

Both these methods applies all block checkpoints, no matter if they contain relevant transactions or
not. To solve this, we can have another method, `apply_block_relevant_assume_connected`, that does
not apply checkpoints of blocks containing no relevant transctions. However, this cannot handle
reorgs mid-sync in an elegant way.

Let's assume we have `apply_block_relevant_assume_connected` which filters checkpoints, and the
folowing scenario plays out:

```
block_height        | 1 | 2 | 3 |
emitted_initially     A   B   C
relevant              A       C
emitted_post_reorg    A   B'
```

`emitted_initially` is the checkpoints that the chain-source has emitted. `relevant` is the
checkpoints that end up being stored in `LocalChain`. `emitted_post_reorg` is the checkpoints that
the chain-source re-emits due to reorg. As can be seen, `A, B'` (the update) cannot connect with 
`A, C`. We need `A, B', C'` as the update.

### My proposal

```rust
/// Introduces transactions of the given block to the wallet.
///
/// Only relevant transactions are inserted. Transactions are inserted alongside their anchors.
fn introduce_block_txs(&mut self, block: &Block, height: u32) { todo!() }

/// Introduces a chain `tip` to the wallet as a `CheckPoint`.
///
/// This updates the `last_synced_to_height: u32` parameter to the height of `tip` if `tip` can
/// connect with the internal `LocalChain`.
///
/// This method attempts to insert the `tip`, but only if it contains relevant transactions.
fn introduce_tip(&mut self, tip: CheckPoint) -> Result<(), CannotConnectError> { todo!() }
```

The chain-source emitter is responsible for emitting full blocks and checkpoints (that connects the
current block to previously emitted blocks).

The block is first processed by `introduce_block_txs`. This inserts relevant transactions and
associated anchors into the wallet.

Then we call `introduce_tip`. If the block contains relevent transactions, the `LocalChain` is
updated with this new tip (and only the tip, since we want to skip irrelevant checkpoints). If the
tip is irrelevant, we only update the `last_synced_to_height: u32` value.

#### How does `introduce_tip` work?

Imagine a situation where the emitter has emitted block height 1 (with hash `A`) and height 2 (with
hash `B`). `1:A` is considered relevant and `2:B` is not. The state of the wallet's `LocalChain`
would be a single checkpoint `[1:A]`.

If the next emission is `3:C` and it contains relevant transactions, the `tip` input of
`introduce_tip` may contain the update chain `[1:A, 2:B, 3:C]` (but of course, we only want to
insert `3:C` and not `2:B`). The logic of `introduce_tip` will iterate the update chain backwards
to determine whether `3:C` can connect with the wallet chain (in this case, it can via `1:A`). With
this knowledge, `try_apply_tip` will create a trimmed update chain `[1:A, 3:C]` that is then applied
to just apply the new tip.

`introduce_tip` also needs to keep checkpoints that are needed to invalidate original checkpoints.
I.e. with a original chain `[1:A, 3:C, 4:D]` and an update chain `[1:A, 3:C, 4:D', 5:E]`.
`apply_tip` still needs to keep height 4 in the update even though it is not the tip. The
stripped-update will be `[3:C, 4:D', 5:E]`.

#### Some tips cannot connect, but we don't return error

Given a scenario with an original chain `[1:A, 3:C, 5:E]` (where the original tip is at height 5).
If there is a 2-block reorg, the chain-source emitter will be tempted to emit block at height 4 (as
that is earliest block reorged). The update chain may be `[1:A, 2:B, 3:C, 4:D']`. This cannot be
connect because we cannot know if `4:D'` and `5:E` belongs in the same chain. However, it is not the
end of the world as the next block emitted will be of height 5. So we just ignore this and wait for
the next emission instead of returning an error.

#### Initiating syncs

Because we are only include checkpoints which contain relevant transactions, we need somewhere
else to track the last-synced-to-height. This value is used when creating a new instance of a
chain-source-emitter. We need to track the last-synced-to-height in the wallet's changeset.

```rust
impl Emitter {
    /// We use `last_cp` for reorg detection. Otherwise, we start emitting from
    /// `last_height - assume_final_depth` where the first emission will connect to `last_cp`.
    fn new(last_height: u32, last_cp: CheckPoint) -> Self { todo!() }
}
```

#### Optimizing `introduce_tip`

Because we are only inserting checkpoints with relevant transactions, inserted checkpoints will be
few and far apart.

I.e. If the original chain is `[1:A, 4:D]` and the next relevant checkpoint is at height 4000,
this means we need to do 3996 (4000-4) iterations just to find out if checkpoint at height 4000 can
connect to `4:D`.

A solution is to cache the most recent irrelevant checkpoints. For example, when we introduce
checkpoint at height 5 (which is irrelevant), we cache it and associate it with our highest relevant
checkpoint `4:D`. We keep doing this so when we get to height 4000, we can iterate from the
introduced `tip` and find out that the previous node (at height 3999) is the same as the height 3999
that is cached. We also know that the cached checkpoint of height 3999 is connected to `4:D`. We can
safely create a trimmed-update of `[4:D, 4000:X]`.

### Changes to Bitcoind RPC chain source

We need to emit `CheckPoint`s alongside blocks.

crates/bdk/src/wallet/mod.rs
evanlinjin: @LLFourn after reading your points in #1245, I agree with this. However, can we not have `apply_block_assume_connected`? I think the emitter should always be explicit with which block to connect to.

evanlinjin:
@vladimirfomene Is it alright if I take over this PR? Thank you in advance!

I agree with @LLFourn's proposal (https://github.com/bitcoindevkit/bdk/pull/1172#discussion_r1406779771) and will implement `apply_block` and `apply_block_connected_to` like so.

notmandatory:
I fixed some new CI issues with bitcoindevkit/bdk#1247, it just needs to be ACKd and merged and then this PR rebased.

vladimirfomene:
> @vladimirfomene Is it alright if I take over this PR? Thank you in advance!
> 
> I agree with @LLFourn's proposal ([#1172 (comment)](https://github.com/bitcoindevkit/bdk/pull/1172#discussion_r1406779771)) and will implement `apply_block` and `apply_block_connected_to` like so.

Go for it

crates/bdk/src/wallet/mod.rs
evanlinjin: Implemented in d55a40799428754d9e33302f9c6057287b8a759f

crates/bdk/src/wallet/mod.rs
```diff
@@ -2330,6 +2379,120 @@ impl<D> Wallet<D> {
...
+
+        let chain_update = local_chain::Update {
+            tip: {
+                let this_block = BlockId {
+                    height,
+                    hash: block_hash,
+                };
+                let block_ids = match height {
+                    h if h > connected_to.height => vec![connected_to, this_block],
+                    h if h < connected_to.height => vec![this_block, connected_to],
```
LLFourn: I don't think we should allow this connected_to height to be < the height being added. The caller can do this switcharoo if this is explicitly what they want. The order of the two should be obvious in sane approaches.

What happened to the `prev_blockhash` block. If it's not the same as `connected_to` then there should be three elements here.

crates/bdk/src/wallet/mod.rs
```diff
@@ -2330,6 +2379,120 @@ impl<D> Wallet<D> {
...
+    /// [`LocalChain`]. Relevant transactions are filtered from the `block` and inserted into the
+    /// internal [`TxGraph`].
+    pub fn apply_block_connected_to(
+        &mut self,
+        block: &Block,
+        height: u32,
+        connected_to: BlockId,
+    ) -> Result<(), ApplyBlockError>
+    where
+        D: PersistBackend<ChangeSet>,
```
LLFourn, line 2424: Most of the logic in here should be in `LocalChain` in a public method like `apply_block_header_connected_to` and this PR should bring in tests for it.

crates/bdk/src/wallet/mod.rs
```diff
@@ -2330,6 +2379,120 @@ impl<D> Wallet<D> {
...
+                };
+                let mut tip = Option::<CheckPoint>::None;
+                for id in block_ids {
+                    let new_tip = match tip {
+                        Some(prev_id) => prev_id.push(id).expect("must push"),
+                        None => CheckPoint::new(id),
+                    };
+                    tip = Some(new_tip);
+                }
+                tip.unwrap()
```
```suggestion
let tip = block_ids[1..].iter().fold(CheckPoint::new(block_ids[0]), |tip, id| tip.push(id).expect("in order"))
```
Or something :^)

Can make a method on CheckPoint like `::from_iter` to reduce cognitive load.

crates/bdk/src/wallet/mod.rs
```diff
@@ -2330,6 +2379,120 @@ impl<D> Wallet<D> {
...
+            self.indexed_graph.apply_block_relevant(block, height),
+        ));
+        self.persist.stage(changeset);
+        Ok(())
+    }
+
+    /// Apply relevant unconfirmed transactions to the wallet.
+    ///
+    /// Transactions that are not relevant are filtered out.
+    pub fn process_unconfirmed_txs<'t>(
```
LLFourn: So we're processing txs here of course but could we say `apply_unconfirmed_txs` instead?

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -224,20 +224,26 @@ where
...
-                core::iter::once(A::from_block_position(&block, block_id, tx_pos)),
-            )
-        });
-        self.batch_insert_relevant(txs)
+        let mut changeset = ChangeSet::<A, I::ChangeSet>::default();
+        for (tx_pos, tx) in block.txdata.iter().enumerate() {
+            changeset.indexer.append(self.index.index_tx(tx));
+            if self.index.is_tx_relevant(tx) {
+                let txid = tx.txid();
+                let anchor = A::from_block_position(block, block_id, tx_pos);
```
LLFourn, line 239: This was done previously in #1041 but I missed it. Why can't we just have `BlockPoistion: Into<A>` bound instead of adding our own conversion trait?

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -127,13 +127,45 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
-        poll(self, |hash| self.client.get_block(hash))
+    pub fn next_block(&mut self) -> Result<Option<Emission<Block>>, bitcoincore_rpc::Error> {
+        Ok(poll(self, |hash| self.client.get_block(hash))?
+            .map(|(checkpoint, block)| Emission { block, checkpoint }))
+    }
+}
+
+/// An emission from [`Emitter`].
+#[derive(Debug)]
+pub struct Emission<B> {
```
LLFourn: This name is a bit annoying. "Oh an emitter, I wonder if what it emits. Oh Emissions".  Can we name it based on the contents. Maybe `NewBlock`.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -127,13 +127,45 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
+    /// Either a full [`Block`] or [`Header`].
+    pub block: B,
+    /// The checkpoint of the emitted block/header.
+    pub checkpoint: CheckPoint,
```
LLFourn, line 158: The reader of the documentation will be interested in what can go wrong if you only have `block` here without `CheckPoint`. Similarly, can we explain why in `Emiiter::new` we have both `start_height` and `last_cp` -- why can't `start_height` just be `last_cp.height()` always. I have a feeling these two things are related.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -127,13 +127,45 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
+    pub fn block_hash(&self) -> BlockHash {
+        self.checkpoint.hash()
+    }
+
+    /// The [`BlockId`] of where this emitted block should connect to.
+    pub fn connected_to(&self) -> BlockId {
+        match self.checkpoint.prev() {
+            Some(prev_cp) => prev_cp.block_id(),
+            // there is no previous checkpoint, so just connect with itself
+            None => self.checkpoint.block_id(),
```
LLFourn, line 182: Why isn't it connected to the block's prev block hash block? This should never emit the genesis block right so this will always work. The key thing I am confused about (which there is a high probability you've already explained it to me months ago) is why we even need this checkpoint thing in the emission.

LLFourn (changes requested):
Main point of of interest for me is why we are returning a `CheckPoint` in the emissions and not just a block or block header. The documentation of the module doesn't explain this clearly and it's not clear why I am using the checkpoint in the example to update the local chain rather than the block header itself. What would go wrong if I just used the block and it's prev block hash to update the local chain?

During review I opened #1260 which I think could also be fixed during this PR.

crates/bitcoind_rpc/src/lib.rs
evanlinjin: `start_height` is the height we start sync at when `last_cp` is the genesis block. Ie. If `start_height` is 1000 and `last_cp` is 0, the first emission will be 1000 (which will be connected to 0).

crates/bdk/src/wallet/mod.rs
evanlinjin: This is designed this way to avoid adding another error variant. Would the current implementation be problematic?

crates/bdk/src/wallet/mod.rs
evanlinjin: Addressed in fc2914af10aaca899af52d65e01d39c3d91ef8d5. I made a convenience method on `CheckPoint`.

crates/bdk/src/wallet/mod.rs
evanlinjin: Added in d38af7a563660d497dd83a525cb2dbfd8815d3ac

crates/bdk/src/wallet/mod.rs
evanlinjin: Done.

crates/bdk/src/wallet/mod.rs
evanlinjin: Okay I decided to return an error if `connected_to` height is greater than the current block.

crates/chain/src/indexed_tx_graph.rs
evanlinjin: It's because there are 3 inputs and doing `A: From<(&Block, BlockId, usize)>` is confusing and we won't know what the `usize` is for.

An alternative would be to have:

```rust
/// Parameters for constructing an [`Anchor`].
pub struct AnchorParams<'b> {
    pub block: &'b Block,
    pub block_id: BlockId,
    pub tx_pos: usize,
}

fn some_method<A: Anchor + From<AnchorParams>>() { todo!() }
```

What do you think? (`AnchorParams` will be nicely documented of course!)

crates/bitcoind_rpc/src/lib.rs
evanlinjin: @LLFourn and I have agreed on `BlockEvent` in a private call.

crates/bitcoind_rpc/src/lib.rs
evanlinjin: I've improved the documentation for `BlockEvent` as well as `Emitter::new`.

crates/bitcoind_rpc/src/lib.rs
evanlinjin: As discussed in the call, if `start_height` is > 1 and the original chain's tip is 0, the first emission will not connect to the prev block.

If the caller constructs `Emitter` from the tip of a `LocalChain`, this should not happen if the bitcoind genesis hash is the same as `LocalChain`'s genesis hash. However, we can't guarantee that the caller will do this. The alternative would be to return an error, but that just makes the API cumbersome.

crates/chain/src/indexed_tx_graph.rs
evanlinjin: I've created an issue to further this discussion: https://github.com/bitcoindevkit/bdk/issues/1266

evanlinjin:
@LLFourn I don't think bitcoindevkit/bdk_wallet#117 should be addressed in this PR. Those seem like substantial changes and I feel like it requires further discussion.

LLFourn (commented):
This LGTM but I'd like to ask whether the changes to the file store actually result in a significant performance improvement? I think the issue is a design bug in the bincode API which is fixed in the v2 release which returns the amount of data read from the reader. Maybe we can just wait for that instead of making this more complicated.

evanlinjin:
@LLFourn wrapping the file in `BufReader` is a significant improvement.

For the counting reader, we still need to back track on any type of error. To backtrack, we still need the position before the read. It's either we seek to get the position, or we store the position.

Side note, maybe `BufReader` caches seeks?

evanlinjin:
@LLFourn I moved `bdk_file_store` changes into a separate PR: bitcoindevkit/bdk#1270

crates/bdk/src/wallet/mod.rs
```diff
@@ -428,6 +430,55 @@ pub enum InsertTxError {
...
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for InsertTxError {}
+
+/// An error that may occur when applying a block to [`Wallet`].
+#[derive(Debug)]
+pub enum ApplyBlockError {
```
ValuedMammal, line 451: Did you intend to use this for either `apply_block` or `apply_block_connected_to`?

crates/bdk/src/wallet/mod.rs
```diff
@@ -2326,6 +2377,86 @@ impl<D> Wallet<D> {
...
+    }
+
+    /// Apply relevant unconfirmed transactions to the wallet.
+    ///
+    /// Transactions that are not relevant are filtered out.
+    ///
+    /// This method takes in an iterator of `(tx, last_seen)` where `last_seen` is the timestamp of
+    /// when the transaction was last seen in the mempool. This is used for conflict resolution
+    /// when there is conflicting unconfirmed transactions. The transaction with the later
+    /// `last_seen` is prioritied.
```
ValuedMammal, line 2448: typo: prioritied

crates/chain/src/local_chain.rs
```diff
@@ -347,6 +370,91 @@ impl LocalChain {
...
+            },
+        };
+        self.apply_header_connected_to(header, height, connected_to)
+            .map_err(|err| match err {
+                ApplyHeaderError::InconsistentBlocks => {
+                    unreachable!("connected_to is derived from the block so is always consistent")
+                }
+                ApplyHeaderError::CannotConnect(err) => err,
+            })
+    }
```
ValuedMammal, line 460: Can this function return `ApplyHeaderError`, or is mapping the err serving some greater purpose?

example-crates/example_bitcoind_rpc_polling/README.md
ValuedMammal, line 1: feel free to ignore: I found this example a bit difficult to follow. It might be better suited for a website tutorial, but if we decide to keep it, I made a few edits in terms of style ValuedMammal/bdk@0caeb791139b2f99e0bcefe424d93567933dbcd0

ValuedMammal (commented):
I think overall this is good. It's nice to see everything come together in `wallet_rpc` in an elegant fashion, can confirm the examples work as intended.

In b486ecc7297dff9a8a97955f14afd58e8b018294 'feat(wallet): introduce block-by-block api', the commit description could be updated (still referring to methods `process_block`, etc).

We should perhaps include more links to example crates in the workspace README (`wallet_rpc`, `example_esplora`, `example_bitcoind_rpc`).

crates/chain/src/local_chain.rs
LLFourn: It should probably be called `ApplyHeaderConnectedTo` error.

crates/chain/src/local_chain.rs
```diff
@@ -347,6 +370,91 @@ impl LocalChain {
         Ok(changeset)
     }
 
+    /// Update the chain with a given [`Header`] and a `connected_to` [`BlockId`].
+    ///
+    /// The `header` will be transformed into checkpoints - one for the current block and one for
+    /// the previous block. Note that a genesis header will be transformed into only one checkpoint
+    /// (as there are no previous blocks). The checkpoints will be applied to the chain via
+    /// [`apply_update`].
```
LLFourn: The documentation doesn't tell us what the method is for and is written in the passive voice which makes it harder to read. It should start by saying that it's going to insert the block hash into the chain along with the previous block if it's not already there. The documentation shouldn't mention `apply_update` -- the reader doesn't care and if we change this then the documentation will be broken.

LLFourn (approved):
ACK db58b8a418d5db26b75041d5e3d0961d2ff61ecc

example-crates/example_bitcoind_rpc_polling/README.md
evanlinjin: This looks great, can you do a PR?

LLFourn:
self-ACK: a4f28c079e3bc4500e8329aeed230d8100d38617

I added one commit to improve the documentation of `apply_header_connected_to`

example-crates/example_bitcoind_rpc_polling/README.md
ValuedMammal: > This looks great, can you do a PR?

Onto this branch or on master?

stevenroose:
I would like to use this :)

evanlinjin:
> I would like to use this :)

Me too! cc. @kcalvinalvin

evanlinjin (approved):
ACK a4f28c079e3bc4500e8329aeed230d8100d38617

PR #1175, opened 2023-10-12
Move Balance out of keychain

### LINKED ISSUE #128: Merge `tx_data_traits.rs` into `chain_data.rs`.
Remove the `tx_data_traits.rs` file and move all of it's contents into `chain_data.rs`. Also move the `Balance` struct into `chain_data.rs`.

See https://discord.com/channels/753336465005608961/753367451319926827/1161983570538147932

### LINKED ISSUE #1173: Move `Balance` out of `keychain`
It should be a top level struct. This can be accomplished by adding a `pub use` in the `lib.rs`

acerbisgianluca:
### Description

Closes bitcoindevkit/bdk_wallet#128. This PR makes `Balance` available at the top-level of the `chain` crate.

#### All Submissions:

### DIFF
--- a/crates/chain/src/lib.rs
+++ b/crates/chain/src/lib.rs
@@ -37,6 +37,7 @@ pub use tx_graph::TxGraph;
 mod chain_oracle;
 pub use chain_oracle::*;
 mod persist;
+pub use keychain::Balance;
 pub use persist::*;
 
 #[doc(hidden)]


evanlinjin (changes requested):
Thank you for taking a crack at this! However, we want to move the actual definition of `Balance` to `lib.rs`. Currently, you are just doing a [re-export](https://doc.rust-lang.org/beta/rustdoc/write-documentation/re-exports.html).

acerbisgianluca:
Hey @evanlinjin, thanks for your feedback!

I'm new to Rust development, but I see that `lib.rs` basically contains only modules and re-exports (correct me if I'm wrong obv), so should I move the whole `Balance` definiton to `lib.rs` anyway? I think that the issue is about re-exporting it.

evanlinjin:
My bad, I see the issue seems to suggest reexporting. However, I don't think `Balance` belongs in the `keychain` module at all as it's not `keychain`-specific?

cc. @danielabrozzoni

danielabrozzoni:
Sorry @evanlinjin, I initially thought you meant to just re-export Balance, my issue was wrongly phrased.

I agree with Evan that having `Balance` inside `keychain.rs` is weird, as there's no keychain-related method that uses the Balance! (When I created the issue I totally thought that was the case, which shows how little I know about bdk_chain still 😅)

I also agree with @acerbisgianluca that having it inside `lib.rs` is a bit weird, since it would be the only struct defined there.

From a quick `git grep` I see that the only structure using the balance is the `TxGraph` (as it should be), @evanlinjin should we move `Balance` inside `src/tx_graph.rs`?  Otherwise, having it in `lib.rs` is fine as well for me.

evanlinjin:
After a chat with @danielabrozzoni, we concluded that it'll be better to defer these decisions.

I also mentioned that I'm also not a fan of the `tx_data_traits.rs` and `chain_data.rs` files (there doesn't seem to be a clear enough distinction between the types declared in each).

Let's continue discussion on the ticket and make sure we all agree before moving forwards. Please also provide your thoughts there!

notmandatory:
Moved to alpha.4 since this is a functional change we should figure out before the beta release.

wthrajat:
Hi @evanlinjin!
> Let's continue discussion on the ticket and make sure we all agree before moving forwards. Please also provide your thoughts there!

Have you reached to a conclusion?

evanlinjin:
@acerbisgianluca we have reached a final conclusion for bitcoindevkit/bdk_wallet#128 

> Remove the `tx_data_traits.rs` file and move all of it's contents into `chain_data.rs`. Also move the `Balance` struct into `chain_data.rs`.

Please update this PR when you have time!

notmandatory:
A small change but I think we should push to 2.0 milestone.

notmandatory:
Closing this one as it's replaced by bitcoindevkit/bdk#1493.

PR #1176, opened 2023-10-12
Implement default index_tx method

### LINKED ISSUE #137: `Indexer::index_tx` can have a default impl in the trait.
I think this is a good idea. Also `index_tx` can have a default impl in the trait.

_Originally posted by @LLFourn in https://github.com/bitcoindevkit/bdk/pull/1084#discussion_r1305441253_

### LINKED ISSUE #1097: `Indexer::index_tx` can have a default impl in the trait.
I think this is a good idea. Also `index_tx` can have a default impl in the trait.

_Originally posted by @LLFourn in https://github.com/bitcoindevkit/bdk/pull/1084#discussion_r1305441253_

vladimirfomene:
### Description

This PR fixes bitcoindevkit/bdk_wallet#137 

### Changelog

This PR implements a default implementation for the Indexer's `index_tx` method. It was recommended on one of the reviews of bitcoindevkit/bdk_wallet#137. 

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/indexed_tx_graph.rs
+++ b/crates/chain/src/indexed_tx_graph.rs
@@ -44,6 +44,18 @@ impl<A, I> IndexedTxGraph<A, I> {
 }
 
 impl<A: Anchor, I: Indexer> IndexedTxGraph<A, I> {
+    /// Determines the [`ChangeSet`] between `self` and an empty [`IndexedTxGraph`].
+    pub fn initial_changeset(&self) -> ChangeSet<A, I::ChangeSet> {
+        let graph = self.graph.initial_changeset();
+        let indexer = self.index.initial_changeset();
+        ChangeSet { graph, indexer }
+    }
+}
+
+impl<A: Anchor, I: Indexer> IndexedTxGraph<A, I>
+where
+    I::ChangeSet: Default + Append,
+{
     /// Applies the [`ChangeSet`] to the [`IndexedTxGraph`].
     pub fn apply_changeset(&mut self, changeset: ChangeSet<A, I::ChangeSet>) {
         self.index.apply_changeset(changeset.indexer);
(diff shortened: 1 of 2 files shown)


evanlinjin:
Can we include a `# Changelog` section in the PR description? Also please remove the checklist items that aren't needed for this PR (or mark them as completed).

vladimirfomene:
@LLFourn was your thinking here that we have a default implementation so that we can get rid of the implementation on `KeychainTxOutIndex` and `SpkTxOutIndex`? Sorry I should have asked this before even creating this PR.

evanlinjin:
> @LLFourn was your thinking here that we have a default implementation so that we can get rid of the implementation on `KeychainTxOutIndex` and `SpkTxOutIndex`? Sorry I should have asked this before even creating this PR.

Yes. Please reuse the default implementation where possible/viable.

evanlinjin (approved):
self-ACK 680fe941d7cb76e275696648458a14cdc5ca9c3b

LLFourn:
I want to register a last-minute Concept-NACK on this.

- `index_txout` is really `index_dangling_txout` because it's only called when you insert a dangling txout. Dangling txouts trade off soundess for convenience. They are unsound because the server that told you about them may be fabricating them. They are a convenient feature since they can be used to figure out fees so you can have a consistent display to the user (ever tx in the list has a feerate). Furthermore, you can of course have a policy of only turning actual transactions into dangling txouts by first downloading the real transaction and then only keeping the txouts you're interested in.
- This PR changes the defualt implementation of `index_tx` to index the txouts of a transaction the same way as you do for dangling txouts. This is a weird default. The implementer of trait should be asking themselves "how do I want to index txouts that I'm not sure even exist" but this PR hides this decision.
- Our current indexer implementations do index dangling txout as if they existed and you could spend them -- this could be a mistake . At the very least I think there could be a boolean passed to the txout indexers to turn this off on and on depending on whether you want to fully trust the dangling utxos in your wallet.
- You might object that this isn't a concern because servers like electrum can always lie about full transactions too but this isn't the case in theory -- you can request merkle proofs from electrum which have to be in a block and you can verify the proof of work of the block and even receive proof that the block is in the main chain. We need an issue to tackle this at some point.

PR #1177, opened 2023-10-13
Upgrade bitcoin/miniscript dependencies

### LINKED ISSUE #1196: Upgrade to `rust-bitcoin` 0.31.0
**Describe the enhancement**  
`rust-bitcoin` 0.31.0 was recently released and includes changes that so far kept LDK from upgrading to 0.3X.Y (cf. https://github.com/lightningdevkit/rust-lightning/issues/2124#issuecomment-1791464639).

**Additional context**
Probably currently blocked on https://github.com/rust-bitcoin/rust-miniscript/issues/622

tcharding:
Upgrade:
    
- bitcoin to v0.31.0
- miniscript to v11.0.0

Fix: #1196

### DIFF
--- a/crates/bdk/src/wallet/signer.rs
+++ b/crates/bdk/src/wallet/signer.rs
@@ -19,13 +19,12 @@
 //! # use core::str::FromStr;
 //! # use bitcoin::secp256k1::{Secp256k1, All};
 //! # use bitcoin::*;
-//! # use bitcoin::psbt;
 //! # use bdk::signer::*;
 //! # use bdk::*;
 //! # #[derive(Debug)]
 //! # struct CustomHSM;
 //! # impl CustomHSM {
-//! #     fn hsm_sign_input(&self, _psbt: &mut psbt::PartiallySignedTransaction, _input: usize) -> Result<(), SignerError> {
+//! #     fn hsm_sign_input(&self, _psbt: &mut Psbt, _input: usize) -> Result<(), SignerError> {
 //! #         Ok(())
 //! #     }
 //! #     fn connect() -> Self {
@@ -55,7 +54,7 @@
 //! impl InputSigner for CustomSigner {
 //!     fn sign_input(
 //!         &self,
-//!         psbt: &mut psbt::PartiallySignedTransaction,
+//!         psbt: &mut Psbt,
 //!         input_index: usize,
 //!         _sign_options: &SignOptions,
 //!         _secp: &Secp256k1<All>,
@@ -87,13 +86,13 @@ use core::cmp::Ordering;
 use core::fmt;
 use core::ops::{Bound::Included, Deref};
 
-use bitcoin::bip32::{ChildNumber, DerivationPath, ExtendedPrivKey, Fingerprint};
+use bitcoin::bip32::{ChildNumber, DerivationPath, Fingerprint, Xpriv};
 use bitcoin::hashes::hash160;
 use bitcoin::secp256k1::Message;
 use bitcoin::sighash::{EcdsaSighashType, TapSighash, TapSighashType};
 use bitcoin::{ecdsa, psbt, sighash, taproot};
 use bitcoin::{key::TapTweak, key::XOnlyPublicKey, secp256k1};
-use bitcoin::{PrivateKey, PublicKey};
+use bitcoin::{PrivateKey, Psbt, PublicKey};
 
 use miniscript::descriptor::{
     Descriptor, DescriptorMultiXKey, DescriptorPublicKey, DescriptorSecretKey, DescriptorXKey,
@@ -264,7 +263,7 @@ pub trait InputSigner: SignerCommon {
     /// Sign a single psbt input
     fn sign_input(
         &self,
-        psbt: &mut psbt::PartiallySignedTransaction,
+        psbt: &mut Psbt,
         input_index: usize,
         sign_options: &SignOptions,
         secp: &SecpCtx,
@@ -279,7 +278,7 @@ pub trait TransactionSigner: SignerCommon {
     /// Sign all the inputs of the psbt
     fn sign_transaction(
         &self,
-        psbt: &mut psbt::PartiallySignedTransaction,
+        psbt: &mut Psbt,
         sign_options: &SignOptions,
         secp: &SecpCtx,
     ) -> Result<(), SignerError>;
@@ -288,7 +287,7 @@ pub trait TransactionSigner: SignerCommon {
 impl<T: InputSigner> TransactionSigner for T {
     fn sign_transaction(
         &self,
-        psbt: &mut psbt::PartiallySignedTransaction,
+        psbt: &mut Psbt,
         sign_options: &SignOptions,
         secp: &SecpCtx,
     ) -> Result<(), SignerError> {
@@ -300,7 +299,7 @@ impl<T: InputSigner> TransactionSigner for T {
     }
 }
 
-impl SignerCommon for SignerWrapper<DescriptorXKey<ExtendedPrivKey>> {
+impl SignerCommon for SignerWrapper<DescriptorXKey<Xpriv>> {
     fn id(&self, secp: &SecpCtx) -> SignerId {
         SignerId::from(self.root_fingerprint(secp))
     }
@@ -310,10 +309,10 @@ impl SignerCommon for SignerWrapper<DescriptorXKey<ExtendedPrivKey>> {
     }
 }
 
-impl InputSigner for SignerWrapper<DescriptorXKey<ExtendedPrivKey>> {
+impl InputSigner for SignerWrapper<DescriptorXKey<Xpriv>> {
     fn sign_input(
         &self,
-        psbt: &mut psbt::PartiallySignedTransaction,
+        psbt: &mut Psbt,
         input_index: usize,
         sign_options: &SignOptions,
         secp: &SecpCtx,
@@ -396,7 +395,7 @@ fn multikey_to_xkeys<K: InnerXKey + Clone>(
         .collect()
 }
 
-impl SignerCommon for SignerWrapper<DescriptorMultiXKey<ExtendedPrivKey>> {
+impl SignerCommon for SignerWrapper<DescriptorMultiXKey<Xpriv>> {
     fn id(&self, secp: &SecpCtx) -> SignerId {
         SignerId::from(self.root_fingerprint(secp))
     }
@@ -406,10 +405,10 @@ impl SignerCommon for SignerWrapper<DescriptorMultiXKey<ExtendedPrivKey>> {
     }
 }
 
-impl InputSigner for SignerWrapper<DescriptorMultiXKey<ExtendedPrivKey>> {
+impl InputSigner for SignerWrapper<DescriptorMultiXKey<Xpriv>> {
     fn sign_input(
         &self,
-        psbt: &mut psbt::PartiallySignedTransaction,
+        psbt: &mut Psbt,
         input_index: usize,
         sign_options: &SignOptions,
         secp: &SecpCtx,
@@ -438,7 +437,7 @@ impl SignerCommon for SignerWrapper<PrivateKey> {
 impl InputSigner for SignerWrapper<PrivateKey> {
     fn sign_input(
         &self,
-        psbt: &mut psbt::PartiallySignedTransaction,
+        psbt: &mut Psbt,
         input_index: usize,
         sign_options: &SignOptions,
         secp: &SecpCtx,
@@ -577,7 +576,7 @@ fn sign_psbt_schnorr(
     hash_ty: TapSighashType,
     secp: &SecpCtx,
 ) {
-    let keypair = secp256k1::KeyPair::from_seckey_slice(secp, secret_key.as_ref()).unwrap();
+    let keypair = secp256k1::Keypair::from_seckey_slice(secp, secret_key.as_ref()).unwrap();
     let keypair = match leaf_hash {
         None => keypair
             .tap_tweak(secp, psbt_input.tap_merkle_root)
@@ -852,7 +851,7 @@ pub(crate) trait ComputeSighash {
     type SighashType;
 
     fn sighash(
-        psbt: &psbt::PartiallySignedTransaction,
+        psbt: &Psbt,
         input_index: usize,
         extra: Self::Extra,
     ) -> Result<(Self::Sighash, Self::SighashType), SignerError>;
@@ -864,7 +863,7 @@ impl ComputeSighash for Legacy {
     type SighashType = EcdsaSighashType;
 
     fn sighash(
-        psbt: &psbt::PartiallySignedTransaction,
+        psbt: &Psbt,
         input_index: usize,
         _extra: (),
     ) -> Result<(Self::Sighash, Self::SighashType), SignerError> {
@@ -913,7 +912,7 @@ impl ComputeSighash for Segwitv0 {
     type SighashType = EcdsaSighashType;
 
     fn sighash(
-        psbt: &psbt::PartiallySignedTransaction,
+        psbt: &Psbt,
         input_index: usize,
         _extra: (),
     ) -> Result<(Self::Sighash, Self::SighashType), SignerError> {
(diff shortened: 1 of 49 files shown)


notmandatory:
Thanks for the headsup and trial patch. Any rough ETA for 0.31 ? Since this would be a breaking change we need to figure out if it could make it in to our 1.0-beta (this year 🤞) or safer to  target for a 2.0 release next year.

tcharding:
There isn't anything massive in this one, I wouldn't stress about getting it into 1.0-beta. You could even skip this one and jump straight to the next one if your 2.0 comes out after March/April.

tcharding:
ETA for v0.31.0 would be in the next couple weeks I'd say.

realeinherjar:
Closes #1191.

tcharding:
`rust-bitcoin` release is out but we are working on `miniscript` still so not much help to you guys yet. Just mentioning in case you see it.

notmandatory:
@tcharding I see that rust-miniscript 11.0 was released with rust-bitcoin/rust-miniscript#618, if the changes aren't too big can we get this into our 1.0-alpha.4 milestone?

tcharding:
Deleted message, first notification of the morning and I totally misunderstood the comment. I'll fix this PR up.

tcharding:
The upgrade is currently stuck on the `rust-electrum-client`, PR looks read to go just needs review please: https://github.com/bitcoindevkit/rust-electrum-client/pull/121

I think I have all the other crate upgrades queued up. I can check them over and take of draft soon as `rust-electrum-client` is upgraded.

notmandatory:
I'm moving this to up to the 1.0.0-alpha.4 release since rust-bitcoin 0.31.0 and rust-miniscript 11.0.0 are released. @tcharding has also created PRs to upgrade the other crates we depend on, we just need to get them merged and released.

tcharding:
CI is expected to be red because I have paths sections in the manifest, to test locally you'll need to either use git wortkrees or update the paths.

tcharding:
> should be good to go

I rescind my claim :)

Has one test failure:

```
cargo test --all --all-features
...
failures:

---- test_bump_fee_add_input_change_dust stdout ----
thread 'test_bump_fee_add_input_change_dust' panicked at crates/bdk/tests/wallet.rs:2109:33:
called `Result::unwrap()` on an `Err` value: CoinSelection(InsufficientFunds { needed: 75159, available: 75000 })


failures:
    test_bump_fee_add_input_change_dust
```

ValuedMammal:
> ```
> cargo test --all --all-features
> ...
> failures:
> 
> ---- test_bump_fee_add_input_change_dust stdout ----
> thread 'test_bump_fee_add_input_change_dust' panicked at crates/bdk/tests/wallet.rs:2109:33:
> called `Result::unwrap()` on an `Err` value: CoinSelection(InsufficientFunds { needed: 75159, available: 75000 })
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> 
> 
> failures:
>     test_bump_fee_add_input_change_dust
> ```

https://github.com/bitcoindevkit/bdk/blob/c01983d02aac6ee49c1bb34ade833e5d72bc8738/crates/bdk/tests/wallet.rs#L2058-L2064

It appears the point of this test is to check that when bumping fee, a new input is added and one output is taken away when compared with the original tx. 

We know what the expected fee is from the value of the ins/outs (50k + 25k - 45K), so why not just set the bumpfee using `builder.fee_absolute()` and leave the rest as is without bothering to play with weights. Any objections to this @tcharding ?

```rust
    // We set a fee high enough that during rbf we are forced to add
    // a new input and also that we have to remove the change
    // that we had previously
    // inputs: 50k + 25k
    // outputs: 45K
    let fee = 50_000 + 25_000 - 45_000;
    builder.fee_absolute(fee);
    let psbt = builder.finish().unwrap();
```

crates/esplora/Cargo.toml
```diff
@@ -13,16 +13,16 @@ readme = "README.md"
 
 [dependencies]
 bdk_chain = { path = "../chain", version = "0.9.0", default-features = false }
-esplora-client = { version = "0.6.0", default-features = false }
+esplora-client = { version = "0.7.0", default-features = false, path = "../../../../rust-esplora-client/02-07-upgrade-bitcoin-0.31" }
```
notmandatory: The `rust-esplora-client` 0.7.0 is now published on crates.io so this can be fixed.

tcharding: Thanks @notmandatory, will do.

tcharding:
Looks like `hwi` is the final missing piece. The `bitcoin` upgrade is done there already.

notmandatory:
> Looks like `hwi` is the final missing piece. The `bitcoin` upgrade is done there already.

Thanks! didn't have `hwi` on my list but will work on getting it ready for a release.

tcharding:
Thanks man, we are trying really hard to get to a place where you don't have to upgrade 300 crates constantly to get new versions of `github.com/rust-bitcoin` code.

notmandatory:
The latest and greatest rust-hwi is published and ready to go: https://crates.io/crates/hwi

tcharding:
FTR this is upgrade only with minimal changes, I did not attempt an improvements that may be possible with the upgrade  eg., using `Amount` instead of `u64` (I just converted everywhere using `to_sat()` and `from_sat()`) - thanks.

tcharding:
ATM I don't know what the CI failure means:
- bug in this patch
- bug in the way weight and/or fee rate are used
- Or, most likely, a difference in the `bitcoin::FeeRate` from the original `bdk::FeeRate`.

Needs further investigation.

notmandatory:
I did some digging into the broken test and found I could fix it by removing the line in `Wallet::create_tx()` where we always add the fee for 2 WU when calculating the new transaction fee_amount. All unit tests still pass with this line removed: https://github.com/tcharding/bdk/blob/df0a46a2731d1aebd734529537525cb2959de889/crates/bdk/src/wallet/mod.rs#L1511

My theory is your changes in https://github.com/rust-bitcoin/rust-bitcoin/pull/2076/commits/c34e3cc7cc03ec0dd69d679bfa560e32474ab0c3 to fix `Transaction::weight()` now already take into account the 2 WU for segwit. Does that make sense?

tcharding:
Sounds plausible and likely. Thanks for investigating for me. I'll re-spin this with your fix.

tcharding:
@notmandatory  you mean the bug was hiding in the line of code with **9 lines of comments** describing why it was there, I probably could have looked a bit harder huh. Thanks for finding it man. I included it in the same patch, if you think its needs pulling out separately I can re-do it.

crates/esplora/Cargo.toml
```diff
@@ -13,17 +13,17 @@ readme = "README.md"
...
 # use these dependencies if you need to enable their /no-std features
-bitcoin = { version = "0.30.0", optional = true, default-features = false }
-miniscript = { version = "10.0.0", optional = true, default-features = false }
+bitcoin = { version = "0.31.0", optional = true, default-features = false }
+miniscript = { version = "11.0.0", optional = true, default-features = false }
 
 [dev-dependencies]
-bdk_testenv = { path = "../testenv", version = "0.1.0", default_features = false }
-electrsd = { version= "0.25.0", features = ["bitcoind_25_0", "esplora_a33e97e1", "legacy"] }
+bdk_testenv = { path = "../testenv", default_features = false }
```
tcharding, line 25: This does not change anything but reduces the maintenance burden because if the version of `bdk_testenv` gets bumped it does not need changing here. This change, and others like it, could be pulled out into a separate patch if you would like me to.

crates/bdk/src/wallet/signer.rs
```diff
@@ -952,40 +951,39 @@ impl ComputeSighash for Segwitv0 {
...
-        Ok((
-            sighash::SighashCache::new(&psbt.unsigned_tx).segwit_signature_hash(
-                input_index,
-                &script,
-                value,
-                sighash,
-            )?,
-            sighash,
-        ))
+        Ok((sighash, sighash_type))
```
ValuedMammal, line 986: This looks better, thank you.

ValuedMammal (approved):
ACK 984c758f964f8fee687d21885b43b0b4705f318c

ValuedMammal:
The only thing I'm confused about (on my end) is why I see this when trying to build locally

```sh
error[E0277]: the trait bound `bitcoin::bitcoin_hashes::sha256d::Hash: ThirtyTwoByteHash` is not satisfied
   --> crates/bdk/src/wallet/signer.rs:537:13
    |
533 |         sign_psbt_ecdsa(
    |         --------------- required by a bound introduced by this call
...
537 |             hash,
    |             ^^^^ the trait `ThirtyTwoByteHash` is not implemented for `bitcoin::bitcoin_hashes::sha256d::Hash`
    |
    = help: the following other types implement trait `ThirtyTwoByteHash`:
              bitcoin::secp256k1::bitcoin_hashes::sha256t::Hash<T>
              bitcoin::secp256k1::bitcoin_hashes::sha256::Hash
              bitcoin::secp256k1::bitcoin_hashes::sha256d::Hash
              bitcoin::LegacySighash
              bitcoin::SegwitV0Sighash
              bitcoin::TapSighash
note: required by a bound in `sign_psbt_ecdsa`
```

crates/esplora/Cargo.toml
notmandatory: I thought all dependencies needed a specific version to publish the crate but I just tested it out and looks like dev-dependencies don't need a version so feel free to remove for `bdk_testenv`.

notmandatory:
> @notmandatory you mean the bug was hiding in the line of code with **9 lines of comments** describing why it was there, I probably could have looked a bit harder huh. Thanks for finding it man. I included it in the same patch, if you think its needs pulling out separately I can re-do it.

Same patch is fine, thanks!

apoelstra:
@ValuedMammal two things:

1. We are moving away from the `ThirtyTwoByteHash` in favor of just `Into<Message>`, since having a dedicated trait that crossed both rust-secp256k1 and bitcoin-hashes was making our upgrades more fragile.
2. We specifically dropped the `ThirtyTwoByteHash` bound on `sha256d::Hash` because as far as we were aware nobody was signing bare hashes like this. How are you obtaining this hash? The rust-bitcoin sighashing methods all return one of the `bitcoin::*Sighash` types.

notmandatory (approved):
ACK 984c758f964f8fee687d21885b43b0b4705f318c

Thanks for doing the update, this was a lot of work!

ValuedMammal:
@apoelstra Thanks for adding the extra context. 

BDK has a trait `ComputeSighash` where the associated type `Sighash` looks to be correct for each script context `Legacy`, `Segwitv0`, and `Tap`. I think part of the problem in this case is we're trying to call `sign_psbt_ecdsa` for both legacy and segwitv0 sighashes by first calling `to_raw_hash`.

https://github.com/bitcoindevkit/bdk/blob/53791eb6c5352d9bc7c8ed1cc3062e48f5a5116c/crates/bdk/src/wallet/signer.rs#L521-L542

Instead of being fancy, I think if we just call `sign_psbt_ecdsa` individually for each case, then this won't be an issue.

```rust
    match self.ctx {
        SignerContext::Segwitv0 => {
            let (hash, hash_ty) = Segwitv0::sighash(psbt, input_index, ())?;
            sign_psbt_ecdsa(
                &self.inner,
                pubkey,
                &mut psbt.inputs[input_index],
                hash,
                hash_ty,
                secp,
                sign_options.allow_grinding,
            );
        }
        SignerContext::Legacy => {
            let (hash, hash_ty) = Legacy::sighash(psbt, input_index, ())?;
            sign_psbt_ecdsa(
                &self.inner,
                pubkey,
                &mut psbt.inputs[input_index],
                hash,
                hash_ty,
                secp,
                sign_options.allow_grinding,
            );
        }
        _ => return Ok(()), // handled above
    };
```

apoelstra:
> Instead of being fancy, I think if we just call `sign_psbt_ecdsa` individually for each case, then this won't be an issue.

Gotcha! Yep, this makes sense -- and I suspect that in future we may try to split out the two `sign_psbt_ecdsa`s into separate methods. (Though I'm not sure; it may be that we let the pre-Taproot API be somewhat flexible/poorly defined and just focus on Taproot.)

tcharding:
Note @apoelstra this is the 0.31 upgrade not the 0.32 upgrade (using secp 0.28 so `ThirtyTwoByteHash` should still be implemented on `hashes` types, we only just removed it in secp 0.29). I'm working on the 0.32-rc1 upgrade as well, its not done yet though.

apoelstra:
@tcharding ah, then I suspect the cause of the error is actually https://github.com/rust-bitcoin/rust-secp256k1/issues/673 ... which is why we are removing `ThirtyTwoByteHash` :)

tcharding:
> Instead of being fancy, I think if we just call sign_psbt_ecdsa individually for each case

The `bdk` signing is not using the PSBT signing API from `rust-bitcoin`, you guys are signing manually using `secp` (and at least as it is now you can't because `rust-bitcoin`s PSBT signing API does not support grinding).

> The only thing I'm confused about (on my end) is why I see this when trying to build locally

Can you check your lock file and see what version of `hashes`, secp, and bitcoin you are pulling in please - just to make sure its not broken at our end.

Thanks

tcharding:
Just for the record, this PR does not attempt to use any of the improvements in the new `rust-bitcoin` version (assuming you see them as improvements). It just makes the code build. E.g., keeps standard integer types instead of using new ADT -  `Weight::to_wu` and `Weight::from_wu` and uses sats `Amount::to_sat` and `Amount::from_sat` instead of `Amount`.

ValuedMammal:
> Can you check your lock file and see what version of `hashes`, secp, and bitcoin you are pulling in please - just to make sure its not broken at our end.

I'm glad you brought this up. While troubleshooting I did a `cargo clean` but never thought to remove the lock file, oops. That seems to have fixed the compile error.

Before resetting Cargo.lock
- bitcoin 0.31.2, hashes 0.13
- secp 0.28.2, hashes 0.12

After
- bitcoin 0.31.2, hashes 0.13
- secp 0.28.2, hashes 0.13

tcharding:
Sweet, everything works then. Thanks

oleonardolima (approved):
ACK 984c758f964f8fee687d21885b43b0b4705f318c

oleonardolima:
Closes #1381

evanlinjin:
Unfortunately this does not compile when rebased on top of `master`. Should be an easy fix though.

```
error[E0277]: the trait bound `bitcoin::bitcoin_hashes::sha256d::Hash: ThirtyTwoByteHash` is not satisfied
   --> crates/bdk/src/wallet/signer.rs:537:13
    |
533 |         sign_psbt_ecdsa(
    |         --------------- required by a bound introduced by this call
...
537 |             hash,
    |             ^^^^ the trait `ThirtyTwoByteHash` is not implemented for `bitcoin::bitcoin_hashes::sha256d::Hash`
    |
    = help: the following other types implement trait `ThirtyTwoByteHash`:
              bitcoin::secp256k1::bitcoin_hashes::sha256t::Hash<T>
              bitcoin::secp256k1::bitcoin_hashes::sha256::Hash
              bitcoin::secp256k1::bitcoin_hashes::sha256d::Hash
              bitcoin::LegacySighash
              bitcoin::SegwitV0Sighash
              bitcoin::TapSighash
note: required by a bound in `sign_psbt_ecdsa`
   --> crates/bdk/src/wallet/signer.rs:551:40
    |
547 | fn sign_psbt_ecdsa(
    |    --------------- required by a bound in this function
...
551 |     hash: impl bitcoin::hashes::Hash + bitcoin::secp256k1::ThirtyTwoByteHash,
    |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `sign_ps
bt_ecdsa`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `bdk` (lib) due to previous error
```

notmandatory:
The corresponding LDK issue is lightningdevkit/rust-lightning#2745.

notmandatory:
> Unfortunately this does not compile when rebased on top of `master`. Should be an easy fix though.

I just did a test-rebase (https://github.com/bitcoindevkit/bdk/pull/1405/checks) and CI builds fine for me. Maybe when you tried it you didn't do a `cargo update`?

storopoli (approved):
ACK 984c758f964f8fee687d21885b43b0b4705f318c

PR #1178, opened 2023-10-15
`LocalChain` with hardwired genesis block

### LINKED ISSUE #1079: Create LocalChain with genesis block hardwired for Network

### LINKED ISSUE #1107: BlockId should not have a default and passing a chain tip should be an `Option`

evanlinjin:
closes #1079
closes #1107

### Description

Many methods of `TxGraph` require a `chain_tip: BlockId` input to use against a `ChainOracle` implementation. This is used to ask the `ChainOracle` implementation whether a certain block exists in the chain identified by the `chain_tip`. This guarantees that the `TxGraph` methods will return a consistent history of transactions.

However, the `ChainOracle` trait's `get_chain_tip` method returns an option of `BlockId`. It becomes unclear what to do when `get_chain_tip` returns `None`.

This PR changes the `ChainOracle::get_chain_tip` method to always return a `BlockId` (no `Option`). `LocalChain` now hardwires the genesis block in order to implement `ChainOracle`.

`bdk::Wallet` and `bdk_file_store::Store` are changed to have separate constructor methods for initializing a fresh instance and recovering a previous instance from persistence.

### Changelog notice

- Changed `ChainOracle::get_chain_tip` method to return a `BlockId` instead of an `Option` of a `BlockId`.
- Refactored `LocalChain` so that the genesis `BlockId` is hardwired. This way, the `ChainOracle::get_chain_tip` implementation can always return a tip.
- Add `is_empty` method to `PersistBackend`. This returns true when there is no data in the persistence.
- Changed `Wallet::new` to initialize a fresh wallet only.
- Added `Wallet::load` to restore an instance of a wallet.
- Replaced `Store::new` with separate methods to create/open the database file.

#### All Submissions:

#### New Features:


crates/bdk/README.md
```diff
@@ -72,7 +72,7 @@ fn main() {
     let db = ();
 
     let descriptor = "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/0'/0'/0/*)";
-    let mut wallet = Wallet::new(descriptor, None, db, Network::Testnet).expect("should create");
+    let mut wallet = Wallet::new(descriptor, None, db, Network::Testnet, None).expect("should create");
```
LLFourn: I'd say either make new functions to support this or create a `Params` type or a `WalletBuilder` or something like that.  It's hard to figure out what these `None`s mean just by looking at the signature.

crates/bdk/src/wallet/mod.rs
```diff
@@ -310,7 +358,9 @@ impl<D> Wallet<D> {
         };
 
         let changeset = db.load_from_persistence().map_err(NewError::Persist)?;
-        chain.apply_changeset(&changeset.chain);
+        chain
+            .apply_changeset(&changeset.chain)
+            .map_err(|_| NewError::InvalidPersistenceGenesis)?;
```
LLFourn: I think we should consider a explicit API  which distinguishes between initializing a wallet for the first time and loading it -- so `::new` and `::load`. It should be simpler for developers and could avoid this error case. Furthermore consider the implications of also doing this: https://github.com/bitcoindevkit/bdk/issues/1101#issuecomment-1701929749 and we also persist the `Network` in the first changeset (waving hands a bit here). Then load could load everything from in one go without the possibility of contradiction between runtime data and persisted data.

So `new` would take what it takes now but would persist the genesis block, network and keychains immediately. `load` would get everything from persistence and applications would just inspect it at runtime to find out what kind of wallet the user had. If we really want this kind of function we could have `new_or_load` as well which just tries to load, checks everything is consistent with the paramters (throwing errors like above if not) or if it can't be loaded, making a new one.

crates/chain/src/genesis.rs
```diff
@@ -0,0 +1,54 @@
+//! Bitcoin genesis hashes.
```
notmandatory: Why can't we use `rust-bitcoin` for this?

https://docs.rs/bitcoin/latest/bitcoin/blockdata/constants/fn.genesis_block.html#

evanlinjin: I didn't know this existed. 😮

evanlinjin: Removed.

crates/bdk/README.md
evanlinjin: Let's do this in a separate PR, once #1101 is completed also.

LLFourn: Do what? Please don't add another param to this function in this PR.

evanlinjin: Sorry I misread your initial comment. I created a separate method for now.

crates/bdk/src/wallet/mod.rs
evanlinjin: I think `Wallet::new` probably needs a check as to whether the underlying `PersistenceBackend` is empty. Otherwise, there may be accidental overwrites.

LLFourn: I don't think the `PersistenceBackend` has an overwrite API though?

evanlinjin: When `bdk_file_store::Store` is initialized, the logical cursor is positioned after the "magic bytes".

`Wallet::new` does not load any data from the `PersistenceBackend`.

So the next write will start at the position after "magic bytes" (if I'm not mistaken).

`Wallet::new` always writes the initial changeset.

LLFourn: So this sounds like an implementation specific problem. I don't think that `Wallet::new` needs to check anything here. We can just have a safe method in `bdk_file_store` that errors out when you create a new db for a file that already exists. Remember the data we are storing is fine to delete. It can always be reconstructed.

evanlinjin:
I'm really close. Remaining tasks:

- [x] Make `bdk_file_store`'s constructor methods "safe" so we don't overwrite existing valid data. As mentioned here: https://github.com/bitcoindevkit/bdk/pull/1178#discussion_r1374036437
- [x] Documentation and tests for `Wallet`'s constructors.

crates/bdk/src/wallet/mod.rs
evanlinjin: I now have separate methods on `bdk_file_store::Store` for initializing and loading.

LLFourn: So I didn't get feedback on the idea of doing the changes I suggested in #1101 as part of this. Is this PR intended to be a stepping stone towards doing #1101 or do you not think #1101 is viable. I doesn't seem valuable to do this change and then change the API again when we do #1101.

crates/chain/src/persist.rs
```diff
@@ -80,6 +80,15 @@ pub trait PersistBackend<C> {
 
     /// Return the aggregate changeset `C` from persistence.
     fn load_from_persistence(&mut self) -> Result<C, Self::LoadError>;
+
+    /// Returns whether the persistence backend contains no data.
+    fn is_empty(&mut self) -> Result<bool, Self::LoadError>
```
LLFourn: why does this need to be on the trait. I don't see why we need it.

crates/bdk/src/wallet/mod.rs
evanlinjin: ~How about we first move the descriptors to a field in the wallet's ChangeSet. Then in a separate PR, we change the keychain txout index to have changesets with descriptors? This way the API of wallet stays consistent.~

^ The above is a bad idea.

I think it's best to wait for @danielabrozzoni's PR for #1101 and have that merged first and base this work on that.

crates/chain/src/persist.rs
evanlinjin: This is for the `Wallet::new_or_load` method. So that we know whether we need to initialize (where we commit the first changeset) or load from existing. Technically, we can check whether required fields exists (i.e. network type and descriptors)... Would you prefer this way instead?

**Edit:** After thinking about it, I think a better option would be for `PersistBackend::load_from_persistence` method to return `Result<Option<C>, Self::LoadError>`. It is helpful to know the difference between having no data in persistence v.s. only having persisted a `Default` changeset.

crates/bdk/src/wallet/mod.rs
```diff
@@ -126,12 +126,18 @@ pub struct ChangeSet {
...
+
+    /// Stores the network type of the wallet.
+    pub network: Option<Network>,
 }
 
 impl Append for ChangeSet {
     fn append(&mut self, other: Self) {
         Append::append(&mut self.chain, other.chain);
         Append::append(&mut self.indexed_tx_graph, other.indexed_tx_graph);
+        if other.network.is_some() {
```
LLFourn, line 140: Should probably `debug_assert` at least that these are not conflicting.

A cleaner less brittle approach would be to have some fixed `Header` data that you can set in persistence. Make an issue if you think it's worth discussing.

LLFourn (approved):
ACK 632b8067cd173700c04aa1b848da49e13c0dea13

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -43,29 +43,12 @@ pub struct Emitter<'c, C> {
 }
 
 impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
-    /// Construct a new [`Emitter`] with the given RPC `client` and `start_height`.
-    ///
-    /// `start_height` is the block height to start emitting blocks from.
-    pub fn from_height(client: &'c C, start_height: u32) -> Self {
+    /// TODO
```
vladimirfomene: I wonder what we can do so that we don't forget to write this doc.

crates/bdk/src/wallet/mod.rs
```diff
@@ -221,99 +231,215 @@ impl Wallet {
...
     },
 }
 
-#[cfg(feature = "std")]
-impl<P: core::fmt::Display + core::fmt::Debug> std::error::Error for NewError<P> {}
-
 impl<D> Wallet<D> {
-    /// Create a wallet from a `descriptor` (and an optional `change_descriptor`) and load related
-    /// transaction data from `db`.
+    /// Initialize an empty [`Wallet`].
```
vladimirfomene, line 405: What does `empty` here mean? From my understanding, it means the wallet has an empty state (empty IndexedTxGraph, empty localchain). We might want to improve the doc here.

crates/bdk/src/wallet/mod.rs
```diff
@@ -326,6 +452,143 @@ impl<D> Wallet<D> {
...
+        D: PersistBackend<ChangeSet>,
+    {
+        let changeset = db
+            .load_from_persistence()
+            .map_err(LoadError::Load)?
+            .ok_or(LoadError::NotInitialized)?;
+        Self::load_from_changeset(descriptor, change_descriptor, db, changeset)
+    }
+
+    fn load_from_changeset<E: IntoWalletDescriptor>(
```
vladimirfomene, line 478: nits: I will change the name of this method to `init_from_changeset`. Feel free to ignore if it doesn't resonate.

crates/file_store/src/store.rs
```diff
@@ -33,30 +33,64 @@ where
...
     }
 }
 
 impl<'a, C> Store<'a, C>
 where
-    C: Default + Append + serde::Serialize + serde::de::DeserializeOwned,
+    C: Append + serde::Serialize + serde::de::DeserializeOwned,
 {
-    /// Creates a new store from a [`File`].
+    /// Create a new [`Store`] file in write-only mode; error if the file exists.
```
vladimirfomene, line 45: Is it really in write only mode given that the file is also given read option on line 65

vladimirfomene (commented):
Great work Evan! I just have some nits.

crates/bitcoind_rpc/src/lib.rs
notmandatory: Looks like docs here should be something like:
```suggestion
    /// Construct a new [`Emitter`] with the given RPC `client`, `last_cp` and `start_height`.
    ///
    /// `last_cp` is the check point used to find the latest block which is still part of the best chain.
    /// `start_height` is the block height to start emitting blocks from.
```

notmandatory (approved):
ACK e342a8c3dd5ce7775756df8ae7bbbe2eb3f61863

Looks good, nice to have that chain tip Option removed!

notmandatory:
This should also close #1107 .

notmandatory:
@evanlinjin looks like this one has a conflict so needs a rebase before I can merge it.

evanlinjin:
@notmandatory don't merge yet, I want to address comments and make clippy happier.

PR #1165, opened 2023-10-10
ci: move to Nix

realeinherjar:
### Description

This would make all the tests possible to run locally.
It enhances developer experience and facilitates onboarding of new contributors.

Additions:

- Crate auditing capabilities using [`rustsec/advisory-db`](https://github.com/rustsec/advisory-db).
- Local and CI caching of all artifacts from running `cargo {check,build,test}` in the whole workspace.
   Superseds `.github/workflows/audit.yml`.
- Pre-commit checks for everything that we enforce (gpg-signed/conventional commits, and some goodies like typos, so bitcoindevkit/bdk#1086 kind of things are not necessary anymore)
- Instructions and rationale in `NIX.md`

Closes bitcoindevkit/bdk_wallet#129.
Superseds bitcoindevkit/bdk#1122 and bitcoindevkit/bdk#1156.

Pinneds Dependencies:

- `bitcoind`: pinned to `0.25.0`
- `electrs`: pinned to [`Blockstream's esplora`](https://github.com/Blockstream/electrs) using the Fedimint deployment, check https://github.com/jkitman/nixpkgs/commit/61ccef8bc0a010a21ccdeb10a92220a47d8149ac and https://github.com/fedimint/fedimint/blob/master/flake.nix#L5

TODO:

- [x] Fix `llvm` deps.
- [x] Fix `openssl` errors.
- [x] Fix dependencies on MacOS.
- [x] Fix `WASM`.
- [x] Fix `MSRV`.
   Add a custom `CargoMSRV.lock`? (Or maybe a `CargoMSRV.toml` with the pinned MSRV dependencies and then `cargo generate-lockfile` with Rust MSRV?)
   Depends on solving https://github.com/ipetkov/crane/issues/422
- [x] move all the logic from `cont_integration.yml` to `flake.nix` (all the `cargo update -p`)
  From the Fedimint suggestion we'll use [`crane`](https://crane.dev).
  This would allow caching of a lot of things
  Then the user would call `nix buid -L .#MSRV --keep-failed` and so on.
  This would also eliminate all the multiple `runs-on` in `cont_integration.yml` to a single one where each step would be a `name` and `run` the `nix build` command.
- [x] Add [`DeterminateSystems/magic-nix-cache-action`](https://github.com/DeterminateSystems/magic-nix-cache-action) to cache stuff in CI.
[...]

PR #1179, opened 2023-10-16
build(esplora): Add async-https-rustls flag to esplora client

thunderbiscuit:
### Description
The bdk_esplora crate currently doesn't expose the [`async-https-rustls` flag offered by the rust-esplora-client](https://github.com/bitcoindevkit/rust-esplora-client/blob/ef1925e1ee1b951f15a9339282858bea27c6168f/Cargo.toml#L44) crate and instead requires users to build using the `default-tls` flag on reqwest, which uses the platform-specific openssl library when compiling. This creates complications for cross-compilation, notably for our Android builds that currently support 3 architectures (`arm64-v8a`, `armeabi-v7a`, and `x86_64`). In order to solve this we can either compile the openssl libraries for each of the platforms we want to support, or use the rustls-tls version of reqwest. The second options is much easier and requires less fiddling with the internals of the Android native development kit and cross-compilation rabbit holes.

Before we merge this I want to make sure I understand the tradeoffs between the `native-tls` and the `rustls-tls` and confirm that there are not potential issues there, but from what I understand they should provide the same functionality/security, and because these are already available/exposed in reqwest and rust-esplora-client, I think this should be a fairly straightforward additional feature we offer.

### Changelog notice
```txt
Added:
  - New async-https-rustls feature flag for the bdk_esplora crate, allowing to compile rust-esplora-client using rustls-tls instead of the default native-tls.
```

#### All Submissions:


realeinherjar (approved):
ACK 7330c7f384324689f376140e62ea3e82bfe86c5e

realeinherjar:
Also in my up-to-date `CargoMSRV.lock` in #1165 I had to use `rustls` versions `0.21.1` and `0.20.8`.
I see that MSRV CI is failing.

tnull:
> Before we merge this I want to make sure I understand the tradeoffs between the native-tls and the rustls-tls and confirm that there are not potential issues there, but from what I understand they should provide the same functionality/security, and because these are already available/exposed in reqwest and rust-esplora-client, I think this should be a fairly straightforward additional feature we offer.

Really interested to hear about your findings. I think generally using the platform's TLS implementation (i.e., `native-tls`) would be preferable as it would allow users to receive critical security fixes via OS upgrades, instead of having to wait and eventually upgrade once rustls, esplora-client, and BDK ship new versions/builds. It would also be preferable from an MSRV perspective as shipping newer `rustls` versions might conflict with the MSRV requirements, potentially leaving the choice between not shipping fixes and breaking MSRV.

I however heard before that making `native-tls` work on Android isn't as trivial as `rustls`, however, I haven't looked too closely into it yet. So, again, would be really interested to hear about your learnings.

nondiremanuel:
Lib Team Call comment: useful for android binding, possibly to be included in the next bdk_esplora release. Since it should be quick adding a flag, we added to alpha.3

notmandatory (approved):
ACK 6817ca9bcbff957efb58f96886a30b6bd87c0cde

realeinherjar (approved):
ACK https://github.com/bitcoindevkit/bdk/commit/6817ca9bcbff957efb58f96886a30b6bd87c0cde

danielabrozzoni (approved):
I am not yet sure if it's better to use rustls-tls or native-tls, however, I think it's important to offer the choice to our users (since, as you said, we already offer it in rust-esplora-client).

I found this post on the rust forum: https://users.rust-lang.org/t/any-reasons-to-prefer-native-tls-over-rustls/37626/6. It seems that the only drawback is what @tnull said, about not getting security patches shipped with the OS, and possible MSRV issues. Someone mentioned that rustls wasn't audited, but that changed in Jun 2020. 

ACK 6817ca9bcbff957efb58f96886a30b6bd87c0cde

PR #1182, opened 2023-10-23
chore: fix MSRV for flate2

evanlinjin:


evanlinjin:
It's still broken, will revisit later (unless if someone would like to take over).

realeinherjar:
I think that `rustls:0.21.7` needs to be changed to `rustls:0.21.8`

notmandatory:
This is ready to re-review and merge. I don't want to have a broken CI holding up merging other PRs while we wait to figure out if it's OK to bump our MSRV with #1183.

notmandatory:
I had to downgrade `hashbrown` since I can't even do a `cargo update` otherwise. It looks like the 0.10 and 0.11 versions are pointing to yanked versions of `ahash`.

notmandatory (approved):
ACK 95250fc44ed1372e77995d04ac2b380e0cb126a0

realeinherjar (approved):
ACK 95250fc44ed1372e77995d04ac2b380e0cb126a0

evanlinjin (commented):
ACK 95250fc44ed1372e77995d04ac2b380e0cb126a0

PR #1186, opened 2023-10-29
Clean up clippy allows

### LINKED ISSUE #1127: Way too many `clippy::allow`
It seems we've got in the habit of doing `clippy::allow` at many points that clippy complains about something. I count ~20 places. I think we should fix most these and refrain from doing it in the future (don't merge new PRs with `clippy::allow` in them).

ValuedMammal:
closes #1127 

There are several instances in the code where we allow clippy lints that would otherwise be flagged during regular checks. It would be preferable to minimize the number of "clippy allow" attributes either by fixing the affected areas or setting a specific configuration in `clippy.toml`. In cases where we have to allow a particular lint, it should be documented why the lint doesn't apply.

For context see https://github.com/bitcoindevkit/bdk/issues/1127#issuecomment-1784256647 as well as the commit message details.

One area I'm unsure of is whether `Box`ing a large error in 4fc2216 is the right approach. Logically it makes sense to avoid allocating a needlessly heavy `Result`, but I haven't studied the implications or tradeoffs of such a change.

#### All Submissions:

### DIFF
--- a/crates/bdk/src/psbt/mod.rs
+++ b/crates/bdk/src/psbt/mod.rs
@@ -35,24 +35,16 @@ pub trait PsbtUtils {
 }
 
 impl PsbtUtils for Psbt {
-    #[allow(clippy::all)] // We want to allow `manual_map` but it is too new.
     fn get_utxo_for(&self, input_index: usize) -> Option<TxOut> {
         let tx = &self.unsigned_tx;
+        let input = self.inputs.get(input_index)?;
 
-        if input_index >= tx.input.len() {
-            return None;
-        }
-
-        if let Some(input) = self.inputs.get(input_index) {
-            if let Some(wit_utxo) = &input.witness_utxo {
-                Some(wit_utxo.clone())
-            } else if let Some(in_tx) = &input.non_witness_utxo {
-                Some(in_tx.output[tx.input[input_index].previous_output.vout as usize].clone())
-            } else {
-                None
-            }
-        } else {
-            None
+        match (&input.witness_utxo, &input.non_witness_utxo) {
+            (Some(_), _) => input.witness_utxo.clone(),
+            (_, Some(_)) => input.non_witness_utxo.as_ref().map(|in_tx| {
+                in_tx.output[tx.input[input_index].previous_output.vout as usize].clone()
+            }),
+            _ => None,
         }
     }
 
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -12,7 +12,7 @@
 //! Wallet
 //!
 //! This module defines the [`Wallet`].
-use crate::collections::{BTreeMap, HashMap, HashSet};
+use crate::collections::{BTreeMap, HashMap};
 use alloc::{
     boxed::Box,
     string::{String, ToString},
@@ -1518,15 +1518,8 @@ impl<D> Wallet<D> {
             return Err(CreateTxError::ChangePolicyDescriptor);
         }
 
-        let (required_utxos, optional_utxos) = self.preselect_utxos(
-            params.change_policy,
-            &params.unspendable,
-            params.utxos.clone(),
-            params.drain_wallet,
-            params.manually_selected_only,
-            params.bumping_fee.is_some(), // we mandate confirmed transactions if we're bumping the fee
-            Some(current_height.to_consensus_u32()),
-        );
+        let (required_utxos, optional_utxos) =
+            self.preselect_utxos(&params, Some(current_height.to_consensus_u32()));
 
         // get drain script
         let drain_script = match params.drain_to {
@@ -2063,17 +2056,26 @@ impl<D> Wallet<D> {
 
     /// Given the options returns the list of utxos that must be used to form the
     /// transaction and any further that may be used if needed.
-    #[allow(clippy::too_many_arguments)]
     fn preselect_utxos(
         &self,
-        change_policy: tx_builder::ChangeSpendPolicy,
-        unspendable: &HashSet<OutPoint>,
-        manually_selected: Vec<WeightedUtxo>,
-        must_use_all_available: bool,
-        manual_only: bool,
-        must_only_use_confirmed_tx: bool,
+        params: &TxParams,
         current_height: Option<u32>,
     ) -> (Vec<WeightedUtxo>, Vec<WeightedUtxo>) {
+        let TxParams {
+            change_policy,
+            unspendable,
+            utxos,
+            drain_wallet,
+            manually_selected_only,
+            bumping_fee,
+            ..
+        } = params;
+
+        let manually_selected = utxos.clone();
+        // we mandate confirmed transactions if we're bumping the fee
+        let must_only_use_confirmed_tx = bumping_fee.is_some();
+        let must_use_all_available = *drain_wallet;
+
         let chain_tip = self.chain.tip().block_id();
         //    must_spend <- manually selected utxos
         //    may_spend  <- all other available utxos
@@ -2088,7 +2090,7 @@ impl<D> Wallet<D> {
 
         // NOTE: we are intentionally ignoring `unspendable` here. i.e manual
         // selection overrides unspendable.
-        if manual_only {
+        if *manually_selected_only {
             return (must_spend, vec![]);
         }
 
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -2290,9 +2290,6 @@ impl<D> Wallet<D> {
     ) -> Result<(), MiniscriptPsbtError> {
         // We need to borrow `psbt` mutably within the loops, so we have to allocate a vec for all
         // the input utxos and outputs
-        //
-        // Clippy complains that the collect is not required, but that's wrong
-        #[allow(clippy::needless_collect)]
         let utxos = (0..psbt.inputs.len())
             .filter_map(|i| psbt.get_utxo_for(i).map(|utxo| (true, i, utxo)))
             .chain(
--- a/crates/esplora/src/async_ext.rs
+++ b/crates/esplora/src/async_ext.rs
@@ -6,10 +6,13 @@ use bdk_chain::{
     local_chain::{self, CheckPoint},
     BlockId, ConfirmationTimeHeightAnchor, TxGraph,
 };
-use esplora_client::{Error, TxStatus};
+use esplora_client::TxStatus;
 use futures::{stream::FuturesOrdered, TryStreamExt};
 
 use crate::anchor_from_status;
+
+/// [`esplora_client::Error`]
+type Error = Box<esplora_client::Error>;
 
 /// Trait to extend the functionality of [`esplora_client::AsyncClient`].
 ///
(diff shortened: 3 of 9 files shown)


ValuedMammal:
The first attempt failed CI due to `clippy::iter_nth_zero`. Of course, there's no reason to think the lint could be disregarded in the first place.. I believe I used the wrong invocation of `cargo clippy` on the command line, allowing it to slip by. I suppose the nix experience should eventually help prevent silly mistakes.

After taking another look at spk_iter.rs, I think it's fine to allow the exception (as it was originally), since we're specifically testing `SpkIterator`'s custom implementation of `nth`. An alternative would be to write the tests in a way that doesn't require calling `nth(0)`, but for now this seems like a trivial case not worth worrying about.

nursery/tmp_plan/src/requirements.rs
```diff
@@ -117,6 +117,7 @@ impl From<bip32::Error> for SigningError {
 impl std::error::Error for SigningError {}
 
 impl RequiredSignatures<DescriptorPublicKey> {
+    #[allow(clippy::too_many_arguments)]
```
LLFourn: feel free to just #[allow(clippy::all)]  in this module's lib.rs

example-crates/example_cli/src/lib.rs
```diff
@@ -183,18 +183,17 @@ impl core::fmt::Display for Keychain {
     }
 }
 
-#[allow(clippy::type_complexity)]
+// Alias changeset info in `create_tx`
+pub type ChangeSet = (keychain::ChangeSet<Keychain>, (Keychain, u32));
```
danielabrozzoni: Instead of defining this alias, can we change the return parameter of create_tx to return a structure? It can contain:
- transaction
- keychain_changeset
- change_keychain
- change_index

I think this would make the code a bit clearer, what do you think?

crates/esplora/src/async_ext.rs
```diff
@@ -6,11 +6,17 @@ use bdk_chain::{
...
+use esplora_client::TxStatus;
 use futures::{stream::FuturesOrdered, TryStreamExt};
 
 use crate::{anchor_from_status, ASSUME_FINAL_DEPTH};
 
+/// `Box` a large [`esplora_client::Error`]
+type Error = Box<esplora_client::Error>;
+
+/// Alias `Result` to use our `Error`
+type Result<T, E = Error> = core::result::Result<T, E>;
```
danielabrozzoni: Why is this line needed?

danielabrozzoni (commented):
Overall it looks really good, thanks a lot!
A few questions/suggestions:

crates/esplora/src/async_ext.rs
ValuedMammal: You're right it's not needed, let's get rid of it. The thought was that it allows omitting the `Err` variant in a function signature similar to what you get from `anyhow::Result`, but I neglected to actually remove them. In this case, it's probably better to be explicit and keep `Error` in the fn signatures.

example-crates/example_cli/src/lib.rs
ValuedMammal: Yes, I like that. However, since three of these fields are either all `Some` or all `None`, it might make sense to have a struct just for `change_info` and continue to return a tuple of `(transaction, change_info)` from `create_tx`

For example, here I introduce `CreateTxChange` and make its members match the names of the existing logic in fn `handle_commands`, does that make sense?

```
--- a/example-crates/example_cli/src/lib.rs
+++ b/example-crates/example_cli/src/lib.rs
@@ -181,21 +181,24 @@ impl core::fmt::Display for Keychain {
             Keychain::Internal => write!(f, "internal"),
         }
     }
 }
 
-// Alias changeset info in `create_tx`
-pub type ChangeSet = (keychain::ChangeSet<Keychain>, (Keychain, u32));
+pub struct CreateTxChange {
+    pub index_changeset: keychain::ChangeSet<Keychain>,
+    pub change_keychain: Keychain,
+    pub index: u32,
+}
 
 pub fn create_tx<A: Anchor, O: ChainOracle>(
     graph: &mut KeychainTxGraph<A>,
     chain: &O,
     keymap: &HashMap<DescriptorPublicKey, DescriptorSecretKey>,
     cs_algorithm: CoinSelectionAlgo,
     address: Address,
     value: u64,
-) -> anyhow::Result<(Transaction, Option<ChangeSet>)>
+) -> anyhow::Result<(Transaction, Option<CreateTxChange>)>
 where
     O::Error: std::error::Error + Send + Sync + 'static,
 {
     let mut changeset = keychain::ChangeSet::default();
 
@@ -385,11 +388,15 @@ where
             }
         }
     }
 
     let change_info = if selection_meta.drain_value.is_some() {
-        Some((changeset, (internal_keychain, change_index)))
+        Some(CreateTxChange {
+            index_changeset: changeset,
+            change_keychain: internal_keychain,
+            index: change_index,
+        })
     } else {
         None
     };
 
     Ok((transaction, change_info))
@@ -593,11 +600,16 @@ where
                 let graph = &mut *graph.lock().unwrap();
                 // take mutable ref to construct tx -- it is only open for a short time while building it.
                 let (tx, change_info) =
                     create_tx(graph, chain, keymap, coin_select, address, value)?;
 
-                if let Some((index_changeset, (change_keychain, index))) = change_info {
+                if let Some(CreateTxChange {
+                    index_changeset,
+                    change_keychain,
+                    index,
+                }) = change_info
+                {
                     // We must first persist to disk the fact that we've got a new address from the
                     // change keychain so future scans will find the tx we're about to broadcast.
                     // If we're unable to persist this, then we don't want to broadcast.
                     {
                         let db = &mut *db.lock().unwrap();
```

danielabrozzoni: Yes, sounds good!

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -531,10 +517,19 @@ impl BranchAndBoundCoinSelection {
         optional_utxos.sort_unstable_by_key(|a| a.effective_value);
         optional_utxos.reverse();
 
+        let mut curr_value = required_utxos
```
danielabrozzoni: I don't think it's worth the hassle to refactor the cs module, since it's going away soon. Also see https://github.com/bitcoindevkit/bdk/issues/1127#issuecomment-1784313256

ValuedMammal:
rebasing now. dropped all the changes to `coin_selection.rs`, and made `ref(tmp_plan)` the last commit

crates/bdk/src/wallet/mod.rs
ValuedMammal, line 1: I guess I'm on the fence regarding `preselect_utxos`. @LLFourn would you prefer we left this untouched so it's clear at the call site exactly what information we're passing?

ValuedMammal:
Reminder to bump msrv in clippy.toml
```
warning: the MSRV in `clippy.toml` and `Cargo.toml` differ; using `1.57.0` from `clippy.toml`
```

ValuedMammal:
Changes in most recent push:

- re-refactored `Psbt::get_utxo_for` to use a match block. I think it conveys the same thing in a more concise way. [compare](https://github.com/bitcoindevkit/bdk/compare/6cdc7f1a95f2a26f65b87ead00cc57b328f18f51..6107725756493ff5f0afa94c25c11d0a598933a1#diff-504dcd23159cb3b14ff38d9c2148749e8a213d292c346c55035f36c0aae7ccd6)

nursery/tmp_plan/src/lib.rs
```diff
@@ -12,7 +12,7 @@
 //! doing coin selection. Furthermore it provides which subset of those keys and hash pre-images you
 //! will actually need as well as what locktime or sequence number you need to set.
 //!
-//! Once you've obstained signatures, hash pre-images etc required by the plan, it can create a
+//! Once you've obtained signatures, hash pre-images etc required by the plan, it can create a
```
LLFourn: I'd prefer we didn't touch this crate and just allowed all.

LLFourn:
Please revert all changes to the tmp plan crate and just blanket allow everything. Other than that this looks good to go.

ValuedMammal:
@LLFourn appreciate your input, thanks for clarifying.

example-crates/example_cli/src/lib.rs
```diff
@@ -643,17 +651,19 @@ where
     }
 }
 
-#[allow(clippy::type_complexity)]
-pub fn init<CS: clap::Subcommand, S: clap::Args, C>(
-    db_magic: &[u8],
-    db_default_path: &str,
-) -> anyhow::Result<(
+// Alias the `Result` of `init`
+pub type InitialState<CS, S, C> = (
```
LLFourn: This could actually be made into a struct where each field had a nice comment if you felt up to it.

LLFourn (approved):
ACK c4f121109198df3a3f96b507fe76a7f14146a135

Thanks @ValuedMammal!

example-crates/example_cli/src/lib.rs
ValuedMammal: Would you rename this?

ValuedMammal:
Is anyone opposed to making the type `InitialState` in `example_cli` into a struct with comments?

```rust
/// The initial state returned from [`init`].
pub struct Init<CS: clap::Subcommand, S: clap::Args, C> {
    /// Arguments parsed by the cli.
    pub args: Args<CS, S>,
    /// Descriptor keymap.
    pub keymap: KeyMap,
    /// Keychain-txout index.
    pub index: KeychainTxOutIndex<Keychain>,
    /// Persistence backend.
    pub db: Mutex<Database<C>>,
    /// Initial changeset.
    pub init_changeset: C,
}
```

danielabrozzoni:
@ValuedMammal go for it!

crates/bdk/src/psbt/mod.rs
```diff
@@ -35,24 +35,20 @@ pub trait PsbtUtils {
...
     fn get_utxo_for(&self, input_index: usize) -> Option<TxOut> {
         let tx = &self.unsigned_tx;
 
-        if input_index >= tx.input.len() {
-            return None;
-        }
+        let input = match self.inputs.get(input_index) {
+            Some(input) => input,
+            None => return None,
+        };
```
evanlinjin: Nit:

```suggestion
        let input = self.inputs.get(input_index)?;
```

example-crates/example_cli/src/lib.rs
```diff
@@ -388,43 +388,46 @@ where
...
     } else {
         None
     };
 
     Ok((transaction, change_info))
 }
 
-#[allow(clippy::type_complexity)]
+// Alias the elements of `Result` of `planned_utxos`
+pub type Plan<K, A> = (bdk_tmp_plan::Plan<K>, FullTxOut<A>);
```
evanlinjin: Nit: can we rename this to `PlannedUTXO`? Thanks!

evanlinjin (commented):
Looking good! Just a couple of NITs

crates/bdk/src/psbt/mod.rs
ValuedMammal: indeed, this is better.

example-crates/example_cli/src/lib.rs
ValuedMammal: I hope you don't mind I used `PlannedUtxo`.

ValuedMammal:
- Rebased onto f099b42005b5310cee51fa2eaa29f94d76753bd7
- Addressed comments by @evanlinjin 

Thank you to all the reviewers! @danielabrozzoni @LLFourn

evanlinjin (approved):
ACK 1c15cb2f9169bb08a4128c183eeeca80fff97ab7

example-crates/example_cli/src/lib.rs
evanlinjin: Woops sorry I was in Golang world these two weeks.

PR #1183, opened 2023-10-25
Bump MSRV to 1.63.0

notmandatory:
### Description

fixes #331 

### Notes to the reviewers

We won't merge this PR until LDK merges lightningdevkit/rust-lightning#2681.

There are alot of clippy checks to fix at 1.63.0 so I left the clippy MSRV at 1.57.0 for now.

### Changelog notice

Changed

- MSRV is now 1.63.0 for bdk, chain, and bitcoind_rpc crates

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #1188, opened 2023-10-30
doc: Improve TxGraph & co docs

danielabrozzoni:
#### All Submissions:

### DIFF
--- a/crates/chain/src/chain_data.rs
+++ b/crates/chain/src/chain_data.rs
@@ -147,6 +147,8 @@ impl From<(&u32, &BlockHash)> for BlockId {
 
 /// An [`Anchor`] implementation that also records the exact confirmation height of the transaction.
 ///
+/// Note that the confirmation block and the anchor block can be different here.
+///
 /// Refer to [`Anchor`] for more details.
 #[derive(Debug, Default, Clone, PartialEq, Eq, Copy, PartialOrd, Ord, core::hash::Hash)]
 #[cfg_attr(
@@ -185,6 +187,8 @@ impl AnchorFromBlockPosition for ConfirmationHeightAnchor {
 
 /// An [`Anchor`] implementation that also records the exact confirmation time and height of the
 /// transaction.
+///
+/// Note that the confirmation block and the anchor block can be different here.
 ///
 /// Refer to [`Anchor`] for more details.
 #[derive(Debug, Default, Clone, PartialEq, Eq, Copy, PartialOrd, Ord, core::hash::Hash)]
--- a/crates/chain/src/indexed_tx_graph.rs
+++ b/crates/chain/src/indexed_tx_graph.rs
@@ -1,7 +1,5 @@
-//! Contains the [`IndexedTxGraph`] structure and associated types.
-//!
-//! This is essentially a [`TxGraph`] combined with an indexer.
-
+//! Contains the [`IndexedTxGraph`] and associated types. Refer to the
+//! [`IndexedTxGraph`] documentation for more.
 use alloc::vec::Vec;
 use bitcoin::{Block, OutPoint, Transaction, TxOut, Txid};
 
@@ -11,9 +9,9 @@ use crate::{
     Anchor, AnchorFromBlockPosition, Append, BlockId,
 };
 
-/// A struct that combines [`TxGraph`] and an [`Indexer`] implementation.
+/// The [`IndexedTxGraph`] combines a [`TxGraph`] and an [`Indexer`] implementation.
 ///
-/// This structure ensures that [`TxGraph`] and [`Indexer`] are updated atomically.
+/// It ensures that [`TxGraph`] and [`Indexer`] are updated atomically.
 #[derive(Debug)]
 pub struct IndexedTxGraph<A, I> {
     /// Transaction index.
@@ -266,7 +264,7 @@ where
     }
 }
 
-/// A structure that represents changes to an [`IndexedTxGraph`].
+/// Represents changes to an [`IndexedTxGraph`].
 #[derive(Clone, Debug, PartialEq)]
 #[cfg_attr(
     feature = "serde",
--- a/crates/chain/src/lib.rs
+++ b/crates/chain/src/lib.rs
@@ -12,9 +12,8 @@
 //!    you do it synchronously or asynchronously. If you know a fact about the blockchain, you can just
 //!    tell `bdk_chain`'s APIs about it, and that information will be integrated, if it can be done
 //!    consistently.
-//! 2. Error-free APIs.
-//! 3. Data persistence agnostic -- `bdk_chain` does not care where you cache on-chain data, what you
-//!    cache or how you fetch it.
+//! 2. Data persistence agnostic -- `bdk_chain` does not care where you cache on-chain data, what you
+//!    cache or how you retrieve it from persistent storage.
 //!
 //! [Bitcoin Dev Kit]: https://bitcoindevkit.org/
 
--- a/crates/chain/src/tx_data_traits.rs
+++ b/crates/chain/src/tx_data_traits.rs
@@ -5,21 +5,25 @@ use alloc::vec::Vec;
 
 /// Trait that "anchors" blockchain data to a specific block of height and hash.
 ///
-/// [`Anchor`] implementations must be [`Ord`] by the anchor block's [`BlockId`] first.
-///
-/// I.e. If transaction A is anchored in block B, then if block B is in the best chain, we can
+/// If transaction A is anchored in block B, and block B is in the best chain, we can
 /// assume that transaction A is also confirmed in the best chain. This does not necessarily mean
 /// that transaction A is confirmed in block B. It could also mean transaction A is confirmed in a
 /// parent block of B.
 ///
+/// Every [`Anchor`] implementation must contain a [`BlockId`] parameter, and must implement
+/// [`Ord`]. When implementing [`Ord`], the anchors' [`BlockId`]s should take precedence
+/// over other elements inside the [`Anchor`]s for comparison purposes, i.e., you should first
+/// compare the anchors' [`BlockId`]s and then care about the rest.
+///
+/// The example shows different types of anchors:
 /// ```
 /// # use bdk_chain::local_chain::LocalChain;
 /// # use bdk_chain::tx_graph::TxGraph;
 /// # use bdk_chain::BlockId;
 /// # use bdk_chain::ConfirmationHeightAnchor;
+/// # use bdk_chain::ConfirmationTimeHeightAnchor;
 /// # use bdk_chain::example_utils::*;
 /// # use bitcoin::hashes::Hash;
-///
 /// // Initialize the local chain with two blocks.
 /// let chain = LocalChain::from_blocks(
 ///     [
@@ -47,6 +51,7 @@ use alloc::vec::Vec;
 /// );
 ///
 /// // Insert `tx` into a `TxGraph` that uses `ConfirmationHeightAnchor` as the anchor type.
+/// // This anchor records the anchor block and the confirmation height of the transaction.
 /// // When a transaction is anchored with `ConfirmationHeightAnchor`, the anchor block and
 /// // confirmation block can be different. However, the confirmation block cannot be higher than
 /// // the anchor block and both blocks must be in the same chain for the anchor to be valid.
@@ -60,6 +65,25 @@ use alloc::vec::Vec;
 ///             hash: Hash::hash("second".as_bytes()),
 ///         },
 ///         confirmation_height: 1,
+///     },
+/// );
+///
+/// // Insert `tx` into a `TxGraph` that uses `ConfirmationTimeHeightAnchor` as the anchor type.
+/// // This anchor records the anchor block, the confirmation height and time of the transaction.
+/// // When a transaction is anchored with `ConfirmationTimeHeightAnchor`, the anchor block and
+/// // confirmation block can be different. However, the confirmation block cannot be higher than
+/// // the anchor block and both blocks must be in the same chain for the anchor to be valid.
+/// let mut graph_c = TxGraph::<ConfirmationTimeHeightAnchor>::default();
+/// let _ = graph_c.insert_tx(tx.clone());
+/// graph_c.insert_anchor(
+///     tx.txid(),
+///     ConfirmationTimeHeightAnchor {
+///         anchor_block: BlockId {
+///             height: 2,
+///             hash: Hash::hash("third".as_bytes()),
+///         },
+///         confirmation_height: 1,
+///         confirmation_time: 123,
 ///     },
 /// );
 /// ```
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -1,12 +1,32 @@
 //! Module for structures that store and traverse transactions.
 //!
-//! [`TxGraph`] is a monotone structure that inserts transactions and indexes the spends. The
-//! [`ChangeSet`] structure reports changes of [`TxGraph`] but can also be applied to a
-//! [`TxGraph`] as well. Lastly, [`TxDescendants`] is an [`Iterator`] that traverses descendants of
-//! a given transaction.
+//! [`TxGraph`] contains transactions and indexes them so you can easily traverse the graph of those transactions.
+//! `TxGraph` is *monotone* in that you can always insert a transaction -- it doesn't care whether that
+//! transaction is in the current best chain or whether it conflicts with any of the
+//! existing transactions or what order you insert the transactions. This means that you can always
+//! combine two [`TxGraph`]s together, without resulting in inconsistencies.
+//! Furthermore, there is currently no way to delete a transaction.
+//!
+//! Transactions can be either whole or partial (i.e., transactions for which we only
+//! know some outputs, which we usually call "floating outputs"; these are usually inserted
+//! using the [`insert_txout`] method.).
+//!
+//! The graph contains transactions in the form of [`TxNode`]s. Each node contains the
+//! txid, the transaction (whole or partial), the blocks it's anchored in (see the [`Anchor`]
+//! documentation for more details), and the timestamp of the last time we saw
+//! the transaction as unconfirmed.
 //!
 //! Conflicting transactions are allowed to coexist within a [`TxGraph`]. This is useful for
-//! identifying and traversing conflicts and descendants of a given transaction.
+//! identifying and traversing conflicts and descendants of a given transaction. Some [`TxGraph`]
+//! methods only consider "canonical" (i.e., in the best chain or in mempool) transactions,
+//! we decide which transactions are canonical based on anchors `last_seen_unconfirmed`;
+//! see the [`try_get_chain_position`] documentation for more details.
+//!
+//! The [`ChangeSet`] reports changes made to a [`TxGraph`]; it can be used to either save to
+//! persistent storage, or to be applied to another [`TxGraph`].
+//!
+//! Lastly, you can use [`TxAncestors`]/[`TxDescendants`] to traverse ancestors and descendants of
+//! a given transaction, respectively.
 //!
 //! # Applying changes
 //!
@@ -49,6 +69,8 @@
 //! let changeset = graph.apply_update(update);
 //! assert!(changeset.is_empty());
 //! ```
+//! [`try_get_chain_position`]: TxGraph::try_get_chain_position
+//! [`insert_txout`]: TxGraph::insert_txout
 
 use crate::{
     collections::*, keychain::Balance, local_chain::LocalChain, Anchor, Append, BlockId,
@@ -91,7 +113,7 @@ impl<A> Default for TxGraph<A> {
     }
 }
 
-/// An outward-facing view of a (transaction) node in the [`TxGraph`].
+/// A transaction node in the [`TxGraph`].
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub struct TxNode<'a, T, A> {
     /// Txid of the transaction.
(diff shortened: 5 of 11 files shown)


crates/chain/src/tx_graph.rs
```diff
@@ -687,8 +701,20 @@ impl<A: Anchor> TxGraph<A> {
...
+    /// tx's conflicting transactions, and tx's ancestors' conflicting transactions;
+    /// if in these transactions we find one which is anchored to a block in the best chain,
+    /// then `tx` can't be in mempool.
+    ///
+    /// Otherwise, we decide which transactions are in mempool based on their
+    /// `last_seen_unconfirmed`. We first of all calculate `tx`'s `max_last_seen_unconfirmed`,
+    /// which is the max `last_seen_unconfirmed` between `tx` and all its descendants.
+    /// We then look at all the conflicts of `tx`, and if all of them have
+    /// `last_seen_unconfirmed` < `max_last_seen_unconfirmed`, then we consider `tx` to be still
+    /// in mempool.
```
danielabrozzoni: I did my best, but I don't think these docs are clear enough. Any suggestions?

crates/chain/src/lib.rs
```diff
@@ -12,9 +12,8 @@
 //!    you do it synchronously or asynchronously. If you know a fact about the blockchain, you can just
 //!    tell `bdk_chain`'s APIs about it, and that information will be integrated, if it can be done
 //!    consistently.
-//! 2. Error-free APIs.
-//! 3. Data persistence agnostic -- `bdk_chain` does not care where you cache on-chain data, what you
-//!    cache or how you fetch it.
+//! 2. Data persistence agnostic -- `bdk_chain` does not care where you cache on-chain data, what you
+//!    cache or how you retrieve it from persistence.
```
realeinherjar:
```suggestion
//!    cache, or how you retrieve it from persistent storage.
```

crates/chain/src/tx_data_traits.rs
```diff
@@ -5,21 +5,22 @@ use alloc::vec::Vec;
 
 /// Trait that "anchors" blockchain data to a specific block of height and hash.
 ///
-/// [`Anchor`] implementations must be [`Ord`] by the anchor block's [`BlockId`] first.
-///
-/// I.e. If transaction A is anchored in block B, then if block B is in the best chain, we can
+/// If transaction A is anchored in block B, then if block B is in the best chain, we can
```
realeinherjar:
```suggestion
/// If transaction A is anchored in block B, and if block B is in the best chain, then we can
```

crates/chain/src/tx_data_traits.rs
```diff
@@ -5,21 +5,22 @@ use alloc::vec::Vec;
...
 /// assume that transaction A is also confirmed in the best chain. This does not necessarily mean
 /// that transaction A is confirmed in block B. It could also mean transaction A is confirmed in a
 /// parent block of B.
 ///
+/// Every [`Anchor`] implementation must contain a [`BlockId`] parameter, and must implement
+/// [`Ord`]. When implementing [`Ord`], the first elements to compare must be the anchors'
```
realeinherjar: This is confusing.
What first elements mean?
You mean the `self.other` in [`std::cmp::Ord`](https://doc.rust-lang.org/std/cmp/trait.Ord.html)?

Maybe?
```suggestion
/// [`Ord`]. When implementing [`Ord`], the elements to compare must be the anchors'
```

crates/chain/src/tx_graph.rs
```diff
@@ -1,12 +1,25 @@
...
+//! certain outputs, which we usually call "floating outputs").
+//!
+//! The graph contains transactions in the form of [`TxNode`] structures. Each node contains the
+//! txid, the transaction (whole or partial), the blocks it's anchored in (see the [`Anchor`]
+//! documentation for more details), and the timestamp of the last time we saw
+//! the transaction as unconfirmed.
 //!
 //! Conflicting transactions are allowed to coexist within a [`TxGraph`]. This is useful for
-//! identifying and traversing conflicts and descendants of a given transaction.
+//! identifying and traversing conflicts and descendants of a given transaction. Some [`TxGraph`]
```
realeinherjar, line 20:
```suggestion
//! identifying, and traversing conflicts and descendants of a given transaction. Some [`TxGraph`]
```

crates/chain/src/tx_graph.rs
```diff
@@ -1,12 +1,25 @@
...
+//! methods only consider "canonical" (i.e., in the best chain or in mempool) transactions,
+//! we decide which transactions are canonical based on anchors `last_seen_unconfirmed`;
+//! see the [`try_get_chain_position`] documentation for more details.
+//!
+//! The [`ChangeSet`] structure reports changes of [`TxGraph`]; it can be used to either save to persistance,
```
realeinherjar:
```suggestion
//! The [`ChangeSet`] structure reports changes of [`TxGraph`]; it can be used to either save to persistent storage,
```

or 

```suggestion
//! The [`ChangeSet`] structure reports changes of [`TxGraph`]; it can be used to either save to persistence,
```

crates/chain/src/tx_graph.rs
```diff
@@ -1,12 +1,25 @@
...
+//! or to be applied to another [`TxGraph`].
+//!
+//! Lastly, you can use [`TxAncestors`]/[`TxDescendants`] to traverse ancestors and descendants of
+//! a given transaction.
```
realeinherjar:
```suggestion
//! Lastly, you can use [`TxAncestors`]/[`TxDescendants`] to traverse ancestors and descendants of
//! a given transaction, respectively.
```

crates/chain/src/tx_graph.rs
```diff
@@ -687,8 +701,20 @@ impl<A: Anchor> TxGraph<A> {
 
     /// Get the position of the transaction in `chain` with tip `chain_tip`.
     ///
-    /// If the given transaction of `txid` does not exist in the chain of `chain_tip`, `None` is
-    /// returned.
+    /// If the given `txid` does not exist in the chain of `chain_tip`, and
+    /// we believe that the transaction is not in mempool anymore, `None` is returned.
+    ///
+    /// We determine whether the transaction `tx` is in the mempool or not by looking at
```
realeinherjar: Ok either use the `the` consistently or drop it consistently.
```suggestion
    /// If the given `txid` does not exist in the chain of `chain_tip`, and
    /// we believe that the transaction is not in the mempool anymore, `None` is returned.
    ///
    /// We determine whether the transaction `tx` is in the mempool or not by looking at
```

or

```suggestion
    /// If the given `txid` does not exist in the chain of `chain_tip`, and
    /// we believe that the transaction is not in mempool anymore, `None` is returned.
    ///
    /// We determine whether the transaction `tx` is in mempool or not by looking at
```

Also if adding `the`, check all the `in mempool` in this PR to propagate consistently.

crates/chain/src/tx_graph.rs
```diff
@@ -687,8 +701,20 @@ impl<A: Anchor> TxGraph<A> {
...
+    /// tx's conflicting transactions, and tx's ancestors' conflicting transactions;
```
realeinherjar:
```suggestion
    /// `tx`'s conflicting transactions, and `tx`'s ancestors' conflicting transactions;
```

Consistently mention `tx`

crates/chain/src/tx_graph.rs
```diff
@@ -687,8 +701,20 @@ impl<A: Anchor> TxGraph<A> {
...
+    /// if in these transactions we find one which is anchored to a block in the best chain,
+    /// then `tx` can't be in mempool.
+    ///
+    /// Otherwise, we decide which transactions are in mempool based on their
+    /// `last_seen_unconfirmed`. We first of all calculate `tx`'s `max_last_seen_unconfirmed`,
+    /// which is the max `last_seen_unconfirmed` between `tx` and all its descendants.
+    /// We then look at all the conflicts of `tx`, and if all of them have
+    /// `last_seen_unconfirmed` < `max_last_seen_unconfirmed`, then we consider `tx` to be still
```
realeinherjar:
```suggestion
    /// We then look at all the conflicts of `tx`, and if for all of them
    /// `last_seen_unconfirmed` < `max_last_seen_unconfirmed` holds, then we consider `tx` to be still
```

realeinherjar (commented):
Some suggestions

crates/chain/src/tx_data_traits.rs
danielabrozzoni: > What first elements mean?

That when you do the comparison in `Ord`, you must first compare the `BlockId`s, and then compare whatever you want if the blockids are the same. How would you rephrase it?

realeinherjar: Oh I see. Then maybe?
```suggestion
/// [`Ord`]. When implementing [`Ord`], the anchors' [`BlockId`]s should take precedence over other elements inside the [`Anchor`]s for comparison purposes.
```

Something like this. Feel free to rephrase it.

danielabrozzoni: Love it, thanks!

crates/chain/src/lib.rs
evanlinjin: I don't mind either `persistence` or `persistent storage`.

crates/chain/src/tx_data_traits.rs
evanlinjin: Nit: I would also remove the `if`.

```suggestion
/// If transaction A is anchored in block B, and block B is in the best chain, we can
```

crates/chain/src/tx_graph.rs
evanlinjin: I don't think a comma is needed here.

crates/chain/src/tx_graph.rs
evanlinjin: Would it be better if we explain it like so?

> There are two factors that we use to approximate whether an unconfirmed transaction still exists in the mempool...
> 1. Conflicts with confirmed txs!
> 2. `last_seen`! We prioritize later `last_seen`s because...

crates/chain/src/chain_data.rs
```diff
@@ -186,6 +189,9 @@ impl AnchorFromBlockPosition for ConfirmationHeightAnchor {
 /// An [`Anchor`] implementation that also records the exact confirmation time and height of the
 /// transaction.
 ///
+/// Note that the confirmation block and the anchor block can be different here. If you
+/// know they're always going to be the same, you can use the [`BlockId`] anchor instead.
```
LLFourn: This advice doesn't make sense I guess since you would lose the `confirmation_time` info (unless we had a way of storing `confirmation_time` in `LocalChain`/`CheckPoint` -- we could turn this into an issue if we feel enthusiastic about the idea but it might be too early).

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -1,6 +1,8 @@
-//! Contains the [`IndexedTxGraph`] structure and associated types.
+//! Contains the [`Indexer`] trait, the [`IndexedTxGraph`] structure,
+//! and the associated [`ChangeSet`].
```
LLFourn: Dunno how valuable this module level documentation is here. Docs.rs will display a list of these just below.

Could just have instead "See [`IndexedTxGraph`] for documentation about how the components of this module relate to each other".

crates/chain/src/lib.rs
LLFourn: persistence is not really meant to be used as a noun in this context IMO.

crates/chain/src/tx_graph.rs
```diff
@@ -1,12 +1,25 @@
 //! Module for structures that store and traverse transactions.
 //!
-//! [`TxGraph`] is a monotone structure that inserts transactions and indexes the spends. The
-//! [`ChangeSet`] structure reports changes of [`TxGraph`] but can also be applied to a
-//! [`TxGraph`] as well. Lastly, [`TxDescendants`] is an [`Iterator`] that traverses descendants of
-//! a given transaction.
+//! [`TxGraph`] is a monotone structure that contains transactions and indexes the spends. It can
```
LLFourn: I felt this bit needed a bit of filling out. May need integration into surrounding.

```suggestion
//! [`TxGraph`] contains transactions and indexes them so you can easily traverse the graph of those transactions.
`TxGraph` is *monotone* in that you can always insert a transaction -- it doesn't care whether that transaction is in the current best chain or whether it conflicts with any of the existing transactions or what order you insert the transactions.
Furthermore, there is currently no way to delete a transaction.
```

crates/chain/src/tx_graph.rs
```diff
@@ -1,12 +1,25 @@
...
+//! contain either full transactions, or partial transactions (transactions for which we have only
+//! certain outputs, which we usually call "floating outputs").
+//!
+//! The graph contains transactions in the form of [`TxNode`] structures. Each node contains the
```
```suggestion
//! The graph contains transactions in the form of [`TxNode`]s. Each node contains the
```

crates/chain/src/tx_graph.rs
LLFourn: persistent storage is better.

In general can we avoid calling things "structures" unless we are really trying to highlight some property of them. This can just be "[`ChangeSet`] reports changes made to to a [`TxGraph`]".

crates/chain/src/tx_graph.rs
LLFourn: Yes but without exclamation marks :P

Also probably not "still exists" but instead "currently exists".

danielabrozzoni: That's better for sure, but I still think that the point 2 (`last_seen`) is not really clear. I'm refactoring this piece of doc as:
```
    /// If the given `txid` does not exist in the chain of `chain_tip`, and
    /// we believe that the transaction is not in mempool anymore, `None` is returned.
    ///
    /// There are two factors that we use to approximate whether an unconfirmed
    /// transaction currently exists in mempool:
    /// 1. Conflicts with confirmed txs. If the transaction `tx` or any of its unconfirmed
    /// ancestors conflict with a confirmed transaction, then `tx` can't be in mempool.
    /// 2. The `last_seen_unconfirmed` parameter. Given two conflicting transactions, we say
    /// that the one with a higher `last_seen_unconfirmed` (i.e., we saw it later) has a higher
    /// chance of still being in mempool. When trying to figure out if `tx`
    /// is still in mempool, we first of all calculate `tx`'s `max_last_seen_unconfirmed`,
    /// which is the max `last_seen_unconfirmed` between `tx` and all its descendants.
    /// We then look at all the conflicts of `tx`, and if for all of them
    /// `last_seen_unconfirmed` < `max_last_seen_unconfirmed` holds,
    /// then we consider `tx` to be still in mempool.
```

How would you rephrase it?

crates/chain/src/tx_graph.rs
danielabrozzoni: I don't like it either, sorry

crates/chain/src/chain_data.rs
danielabrozzoni: Ah, I didn't notice this. Thanks! I wonder if it would make sense to have some comment explaining the differences between the various anchor implementations, and if so, where it would make sense to put it. I feel like it's something important to have, but that doesn't really belong to the `Anchor` trait.

danielabrozzoni:
Ready for another rounds of reviews, thanks everyone!

crates/chain/src/chain_data.rs
LLFourn: Yes I think so. It's a good idea to explain why each anchor exists on the implementation. It's mostly reflective of what we get from different blockchain data sources..

crates/chain/src/tx_graph.rs
```diff
@@ -1,12 +1,30 @@
...
-//! [`TxGraph`] as well. Lastly, [`TxDescendants`] is an [`Iterator`] that traverses descendants of
-//! a given transaction.
+//! [`TxGraph`] contains transactions and indexes them so you can easily traverse the graph of those transactions.
+//! `TxGraph` is *monotone* in that you can always insert a transaction -- it doesn't care whether that
+//! transaction is in the current best chain or whether it conflicts with any of the
+//! existing transactions or what order you insert the transactions.
+//! Furthermore, there is currently no way to delete a transaction.
+//!
+//! Transactions can be either whole or partial (i.e., transactions for which we only
+//! know some outputs, which we usually call "floating outputs").
```
vladimirfomene: I think it will be good to document scenarios in which we get partial transactions.

crates/chain/src/tx_graph.rs
```diff
@@ -127,7 +146,7 @@ impl Default for TxNodeInternal {
     }
 }
 
-/// An outwards-facing view of a transaction that is part of the *best chain*'s history.
+/// A transaction that is included in a certain chain.
```
vladimirfomene: Since a canonical transaction could still be in the mempool, it might be good to mention it here as well.

vladimirfomene (commented):
Just nits.

crates/chain/src/chain_data.rs
danielabrozzoni: I realized we already have a short explanation of the various types of anchors in the Anchor docs, I expanded it a little. It still doesn't go deep in the *why* different anchors exist, but it's a start.

crates/chain/src/tx_graph.rs
danielabrozzoni: I didn't realize, thanks!

crates/chain/src/tx_graph.rs
danielabrozzoni: Fair, thanks :)

crates/chain/src/tx_data_traits.rs
```diff
@@ -5,21 +5,26 @@ use alloc::vec::Vec;
...
+/// If transaction A is anchored in block B, and block B is in the best chain, we can
 /// assume that transaction A is also confirmed in the best chain. This does not necessarily mean
 /// that transaction A is confirmed in block B. It could also mean transaction A is confirmed in a
 /// parent block of B.
 ///
+/// Every [`Anchor`] implementation must contain a [`BlockId`] parameter, and must implement
+/// [`Ord`]. When implementing [`Ord`], the anchors' [`BlockId`]s should take precedence
+/// over other elements inside the [`Anchor`]s for comparison purposes, i.e., you should first
+/// compare the anchors' [`BlockId`]s and then care about the rest.
+/// [`BlockId`]s.
```
evanlinjin: This looks like a mistake. Please remove

crates/chain/src/tx_graph.rs
```diff
@@ -1,12 +1,31 @@
 //! Module for structures that store and traverse transactions.
 //!
-//! [`TxGraph`] is a monotone structure that inserts transactions and indexes the spends. The
-//! [`ChangeSet`] structure reports changes of [`TxGraph`] but can also be applied to a
-//! [`TxGraph`] as well. Lastly, [`TxDescendants`] is an [`Iterator`] that traverses descendants of
-//! a given transaction.
+//! [`TxGraph`] contains transactions and indexes them so you can easily traverse the graph of those transactions.
+//! `TxGraph` is *monotone* in that you can always insert a transaction -- it doesn't care whether that
```
evanlinjin, line 4: Maybe it's also worth mentioning that we can always combine two TxGraphs without resulting in inconsistencies.

danielabrozzoni: Done, thanks!

danielabrozzoni:
I pushed a few additions as specified in #1109

crates/chain/src/tx_graph.rs
evanlinjin:
```
/// Get the position of the transaction in a chain with tip `chain_tip`.
///
/// Chain data is fetched from a [`ChainOracle`] implementation - `chain`.
///
/// This method returns `Ok(None)` if the transaction is not found in the chain, and no longer
/// belongs in the mempool. The following factors are used to approximate whether an 
/// unconfirmed transaction exists in the mempool (not evicted):
///
/// 1. Unconfirmed transactions that conflict with confirmed transactions are evicted.
/// 2. Unconfirmed transactions that spend from transactions that are evicted, are also
///    evicted.
/// 3. Given two conflicting unconfirmed transactions, the transaction with the lower
///    `last_seen_unconfirmed` parameter is evicted. A transaction's `last_seen_unconfirmed` 
///    parameter is the max of all it's descendants' `last_seen_unconfirmed` parameters. If the
///    final `last_seen_unconfirmed`s are the same, the transaction with the lower `txid` (by
///    lexicographical order) is evicted.
///
/// # Error
///
/// An error will occur if the [`ChainOracle`] implementation (`chain`) fails. If the
/// [`ChainOracle`] is infallible, [`get_chain_position`] can be used instead.
///
/// [`get_chain_position`]: Self::get_chain_position
```

@danielabrozzoni What do you think about this?

danielabrozzoni: Looks good! Updated, thanks

notmandatory (approved):
ACK a5e087a09e283f40347488e5a35f7f55ecb6ed73

It looks like all comments have been addresses and this is a nice improvement to the docs. @LLFourn and @evanlinjin if you don't have any other suggestions can we merge this now?

LLFourn:
There are still points where we refer to types as "structs" or "structures". For example:

> /// This struct is created by the [`walk_ancestors`] method of [`TxGraph`].

Can just be:

> /// Returned by the [`walk_ancestors`] method of [`TxGraph`] 

This is my only nit. For some reason this redundancy of calling something which will obviously be displayed as a `struct` a struct makes me seethe.

danielabrozzoni:
I pushed some final changes to the docs, removing the "struct"/"structure" terms where I found them, but the MSRV CI is broken :(

notmandatory:
> I pushed some final changes to the docs, removing the "struct"/"structure" terms where I found them, but the MSRV CI is broken :(

~~I just fixed the CI issue in #1236, you can cherry pick it in to this PR or rebase after it gets merged.~~ It's better to merge #1183 instead.

notmandatory:

danielabrozzoni:
Done, thanks!

notmandatory (approved):
Re-ACK 0adff9c35f109a15740212b4dce74c03950c85ed

PR #1190, opened 2023-11-01
Add `Wallet::list_output` method

### LINKED ISSUE #1184: Add `Wallet::list_outputs` method.
As pointed out by @darosior on [Discord](https://discord.com/channels/753336465005608961/978744259693916230/1166660800312778812), there is no `list_unspent` equivalent in `Wallet` that lists all owned outputs (spent or unspent).

evanlinjin:
Fixes #1184

### Description

Introduce `Wallet::list_output` method that lists all outputs (both spent and unspent) in a consistent history.

### Changelog notice

- Rename `LocalUtxo` to `LocalOutput`.
- Add `Wallet::list_output` method.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -814,15 +814,20 @@ impl<D> Wallet<D> {
     }
 
     /// Inserts a [`TxOut`] at [`OutPoint`] into the wallet's transaction graph.
-    /// Any inserted TxOuts are not persisted until [`commit`] is called.
     ///
-    /// This can be used to add a `TxOut` that the wallet doesn't own but is used as an input to
-    /// a [`Transaction`] passed to the [`calculate_fee`] or [`calculate_fee_rate`] functions.
+    /// This is used for providing a previous output's value so that we can use [`calculate_fee`]
+    /// or [`calculate_fee_rate`] on a given transaction. Outputs inserted with this method will
+    /// not be returned in [`list_unspent`] or [`list_output`].
     ///
-    /// Only insert TxOuts you trust the values for!
+    /// Any inserted `TxOut`s are not persisted until [`commit`] is called.
+    ///
+    /// **WARNING:** This should only be used to add `TxOut`s that the wallet does not own. Only
+    /// insert `TxOut`s that you trust the values for!
     ///
     /// [`calculate_fee`]: Self::calculate_fee
     /// [`calculate_fee_rate`]: Self::calculate_fee_rate
+    /// [`list_unspent`]: Self::list_unspent
+    /// [`list_output`]: Self::list_output
     /// [`commit`]: Self::commit
     pub fn insert_txout(&mut self, outpoint: OutPoint, txout: TxOut)
     where
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -737,7 +737,7 @@ impl<D> Wallet<D> {
     }
 
     /// Return the list of unspent outputs of this wallet
-    pub fn list_unspent(&self) -> impl Iterator<Item = LocalUtxo> + '_ {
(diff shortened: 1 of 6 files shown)


notmandatory (approved):
ACK 63dea4611c4fd985eb6d843883dbb8259d4e4f25

One small NIT is I'd label 723658fce55320866d2f62b2837edd03b2cbeea3 as a "refactor!" even though it's only a rename:

https://softwareengineering.stackexchange.com/questions/381736/renaming-%e2%8a%88-refactoring

vladimirfomene:
ACK a84446f2ba3f1d2732148e2d833f9c3fbec16291

crates/bdk/src/wallet/mod.rs
```diff
@@ -443,6 +443,20 @@ impl<D> Wallet<D> {
             .map(|((k, i), full_txo)| new_local_utxo(k, i, full_txo))
     }
 
+    /// List all relevant outputs (includes both spent and unspent).
```
danielabrozzoni: Does this include both confirmed and unconfirmed outputs? In any case, can this be specified?

Also, we should specify that floating outputs are not included, as said in the `try_filter_chain_txouts` docs.

crates/bdk/tests/wallet.rs
```diff
@@ -139,6 +139,25 @@ fn test_get_funded_wallet_tx_fee_rate() {
     assert_eq!(tx_fee_rate.as_sat_per_vb(), 8.849558);
 }
 
+#[test]
+fn test_list_output() {
+    let (wallet, txid) = get_funded_wallet(get_test_wpkh());
```
danielabrozzoni, line 242: nit: It would be better if this test also included an unconfirmed transaction, but if you don't want to, that's fine as well

crates/bdk/src/wallet/mod.rs
evanlinjin: > Does this include both confirmed and unconfirmed outputs? In any case, can this be specified?

Yes it does. I'll mention it.

> Also, we should specify that floating outputs are not included, as said in the try_filter_chain_txouts docs.

Do we want to expose the concept of "floating outputs" to the `Wallet`? The documentation of `Wallet::insert_txout` already states; *"This can be used to add a `TxOut` that the wallet does not own..."*.

crates/bdk/tests/wallet.rs
evanlinjin: I think it should be fine as `filter_chain_txouts` is thoroughly tested?

notmandatory (approved):
Re-ACK 01037cd5d67033bc701c7d8d2f9e8cac12732f8d

crates/bdk/src/wallet/mod.rs
danielabrozzoni: > Do we want to expose the concept of "floating outputs" to the Wallet?

I agree that we might not want that, but then list_outputs should also list floating outputs... Otherwise someone could insert manually a txout and then not seeing it here.

evanlinjin: @danielabrozzoni what I mean is that the exposed methods to add a "floating `TxOut`" explicitly state that the output will not be added as an owned output (we can make these docs clearer).

Non-owned outputs won't be listed.

evanlinjin: As discussed, we should improve docs of `insert_txout`!

evanlinjin: Clarified docs in 278210bb8918923005e80968df3dfcc3372c0a5c

notmandatory (approved):
re-ACK 278210bb8918923005e80968df3dfcc3372c0a5c

New docs do a good job of clarifying for users which outputs they'll get from list_output, and which they won't from insert_txout.

danielabrozzoni (approved):
ACK 278210bb8918923005e80968df3dfcc3372c0a5c

Thanks for the docs update! Looking fantastic.

Issue #1112, opened 2023-09-05
`scan_without_keychain` and `scan`'s names should be inverted


LLFourn:
There are some naming problems in electrum and esplora code.

- `scan` should mean without a keychain
- `scan_with_keychain` should be the name of the one that scans with a keychain

Same applies to `update_tx_graph_without_keychain` and `update_tx_graph` in esplora (which I would likewise rename to `scan` and `scan_with_keychain`).

The problem with the current naming is that it makes it seem `scan` (with a keychain) is the simple default and that `scan_without_keychain` is the exception. But scanning with a keychain (stop_gap) is very inefficient and should only be used when restoring from seedwords. Mutiny wallet caused a lowkey denial of service attack on mempool.space because they were always scanning with keychain. 

- it needs to be documented on `scan_with_keychain` that it should be used sparingly and explain the circumstances where it should be used.
- I would remove all the other arguments from `scan_with_keychain` other than the `keychain_spks` in both esplora and electrum. This emphasizes the special case that it should be used for.
- Perhaps we should change `scan` to `sync` instead since that's the nomenclature we have elsewhere.

vladimirfomene:
@LLFourn do you think this can be tagged "good first issue"?

LLFourn:
sure but it should be done with high priority anyway.

PR #1194, opened 2023-11-02
Add wallet sync_request and full_scan_request functions

### LINKED ISSUE #1153: [wallet] Consider having a higher-level method for syncing
Looking at wallet_esplora or wallet_electrum, syncing takes quite a bit of code, and I suppose than 99% of users will just want to update the whole wallet in one big batch, without any fancy optimization. We should consider having a higher level method similar to the old `wallet.sync()` to reduce the amount of code copy-pasted from our examples.

Maybe we could have `sync_esplora`, `sync_electrum`, `sync_bitcoind` and all of them need a wallet passed in?

notmandatory:
This is WIP.  So far only demonstrating with async esplora, if approach looks good I'll do the rest.

### Description

Adding new functions to for esplora and electrum light client ext modules to simplify full_scan and sync operations.  

This is another attempt to fix #1153. 

#### All Submissions:

#### New Features:

#### Bugfixes:


evanlinjin:
@notmandatory do you mind adding a more detailed explanation as to how this pushes us in a direction to have a higher-level method/methods for syncing a wallet?

LLFourn:
@notmandatory what about doing it like I suggested here: https://github.com/bitcoindevkit/bdk/issues/1153#issuecomment-1752263555

notmandatory:
Suggestion from https://github.com/bitcoindevkit/bdk/issues/1153#issuecomment-1752263555

```rust
let things_I_am_interested_in = wallet.lock().unwrap().start_sync();
let update = electrum_or_esplora.sync(things_i_am_interested_in)?;
wallet.lock().unwrap().apply_update(update)?:
```

@evanlinjin where I'm at right now is trying to figure out what wallet data is needed for scan and sync and how to get what we need from existing wallet functions.  I should be able to get back to the suggested solution above but am stuck on a few questions maybe you and @LLFourn can help with.

1. does it make sense to have different `wallet.start_sync` and `wallet.start_scan` since those operations need different info? 

2. `things_I_am_interested_in` and `update` need to be tuples right? I don't see any place to put new structs for these without creating unwanted cross-crate-dependencies. It will be trivial to make a `impl From` to convert the tuple returned from any of the clients back into a wallet `Update`.

3. For the `things_I_am_interested_in` does it make sense to clone and collect what I can? I know the goal is to keep this thread safe, is cloning what I need for syncing the only/best option?

What I've found so far is for a `client.scan()` we need three things from the wallet:

```rust
keychain_spks: BTreeMap<K, impl IntoIterator<IntoIter = impl Iterator<Item = (u32, ScriptBuf)> + Send> + Send,>,
local_chain: &LocalChain,
local_tip: Option<CheckPoint>,
```

and for a `client.sync()` we need five things:
```rust
spks: Vec<ScriptBuf>,
local_chain: &LocalChain,
local_tip: Option<CheckPoint>,
outpoints: Vec<OutPoint>,
txids: Vec<Txid>,
```

If you guys have any better ideas for these types I'll try to make them work. Thanks!

notmandatory:
I added Wallet `start_scan` and `start_sync` functions as an example of what I have in mind. Once I have confirmation the types are correct I need to add some tests and probably better docs and then also implement `scan` and `sync` for EsploraExt and ElectrumExt and update related docs and examples.

LLFourn:
> Suggestion from [#1153 (comment)](https://github.com/bitcoindevkit/bdk/issues/1153#issuecomment-1752263555)
> 
> ```rust
> let things_I_am_interested_in = wallet.lock().unwrap().start_sync();
> let update = electrum_or_esplora.sync(things_i_am_interested_in)?;
> wallet.lock().unwrap().apply_update(update)?:
> ```
> 
> @evanlinjin where I'm at right now is trying to figure out what wallet data is needed for scan and sync and how to get what we need from existing wallet functions. I should be able to get back to the suggested solution above but am stuck on a few questions maybe you and @LLFourn can help with.
> 
>     1. does it make sense to have different `wallet.start_sync` and `wallet.start_scan` since those operations need different info?

I hadn't thought of making `start_scan` as well. But I think we're renaming this to `rescan`  and fixing "scan" arguments to just what we need https://github.com/bitcoindevkit/bdk/issues/1112.

> 
>     2. `things_I_am_interested_in` and `update` need to be tuples right? I don't see any place to put new structs for these without creating unwanted cross-crate-dependencies. It will be trivial to make a `impl From` to convert the tuple returned from any of the clients back into a wallet `Update`.

I don't think they should be tuples? Just a `struct` definition containing all spks, txids etc. The definition should be in `bdk_chain`.

> 
>     3. For the `things_I_am_interested_in` does it make sense to clone and collect what I can? I know the goal is to keep this thread safe, is cloning what I need for syncing the only/best option?

Yes. Things should be cloned and collected out of wallet. FWIW I think we overuse `IntoIterator` currently. These can just be `Vec`.

I made another issue to try and collect requirements here: #1195

crates/esplora/src/async_ext.rs
ValuedMammal, line 1: As a side note, it looks like the trait bounds on `keychain_spks`  in `scan_txs_with_keychain` can be simplified to `BTreeMap<K, I>` where `K: Ord + Clone + Send, I: Iterator<Item = (u32, ScriptBuf)> + Send` similar to what you have for the new `scan` function, since we pass in an iterator of spks and not a collection. I second @LLFourn 's thought that other params like request_heights, txids, and outpoints could just be Vec or BTreeSet.

notmandatory:
This will need a little rework after #1178 is merged.

notmandatory:
Also based on poll on discord I'm renaming function names to `full_scan` and `sync`, see: https://discord.com/channels/753336465005608961/753367451319926827/1171551358353158164

crates/esplora/src/async_ext.rs
notmandatory: Hey sorry I didn't respond sooner to this. I'm going to rework this PR once the alpha.3 release is done and incorporate your suggestions.

crates/chain/src/request.rs
```diff
@@ -0,0 +1,32 @@
+use crate::collections::BTreeMap;
+use crate::local_chain::CheckPoint;
+use alloc::vec::Vec;
+use bitcoin::{OutPoint, ScriptBuf, Txid};
+
+/// Data required to perform a blockchain client sync.
```
evanlinjin: Imo we should be clear that these structures are only useful for spk-based syncing.

example-crates/wallet_esplora_async/src/main.rs
```diff
@@ -24,62 +33,93 @@ async fn main() -> Result<(), anyhow::Error> {
...
+            // 4. get blockchain update from original request checkpoint and missing heights
+            let chain_update = client
+                .update_local_chain(request.checkpoint, missing_heights)
+                .await?;
+            // 5. create wallet update
+            Ok((
+                Update {
+                    graph: graph_update,
+                    chain: Some(chain_update),
+                    ..Update::default()
```
evanlinjin: I was thinking about creating some sort of builder structure for ChangeSets. Maybe we can do something similar for Updates?

```rust
pub struct UpdateBuilder<U> { u: U }

impl<U> UpdateBuilder<U> {
    pub fn chain(self, u: impl Into<U>) -> Self { todo!() }
    pub fn build(self) -> U { self.u }
}
```

However changesets can be appended, but updates may not? Maybe `chain` needs to return a result. Would this even be a good idea?

notmandatory: I think the default struct construction above is pretty clear and easy to use.

notmandatory:
I've rebased post alpha.4 tag.

benthecarman:
This would be great, would fix the most annoying part of using bdk for me!

example-crates/wallet_esplora_async/src/main.rs
ValuedMammal: Builder pattern would make sense for building a sync/scan request. I think the current approach leaves the door open to potentially evolve in that direction.

LLFourn: Why would you need a builder pattern for changesets when you should never create these yourself?
It could be helpful for updates but I don't see what's wrong with the code above?

Oh except I wonder if we could have the concept of the "empty chain" which just has the geneisis block or whatever rather than having to have an `Option` there (but I guess not if you want `Default` to work so ignore that).

crates/bdk/src/wallet/mod.rs
```diff
@@ -109,6 +110,38 @@ pub struct Update {
...
+
+impl From<(TxGraph<ConfirmationTimeHeightAnchor>, local_chain::Update)> for Update {
+    fn from((graph, chain): (TxGraph<ConfirmationTimeHeightAnchor>, local_chain::Update)) -> Self {
+        Self {
+            graph,
+            chain: Some(chain),
+            ..Default::default()
+        }
+    }
+}
```
LLFourn: Are these `From` impls really important. These kinds of things are hard for users to find even if they are useful. It's better not to have a codebase full of secret tricks.

crates/bdk/src/wallet/mod.rs
```diff
@@ -2353,6 +2386,70 @@ impl<D> Wallet<D> {
...
+        // We want to search for whether our UTXOs are spent, and spent by which transaction.
+        let outpoints: Vec<OutPoint> = self.list_unspent().map(|utxo| utxo.outpoint).collect();
+
+        // Sync unconfirmed TX
+        // We want to search for whether our unconfirmed transactions are now confirmed.
+        let txids: Vec<Txid> = self
+            .transactions()
+            .filter(|canonical_tx| !canonical_tx.chain_position.is_confirmed())
+            .map(|canonical_tx| canonical_tx.tx_node.txid)
+            .collect();
```
LLFourn: Performance note: In theory there are faster ways to get the set of unconfirmed transactions in the graph. We haven't implemented them though.

example-crates/wallet_esplora_async/src/main.rs
```diff
@@ -24,62 +32,93 @@ async fn main() -> Result<(), anyhow::Error> {
...
-        .all_unbounded_spk_iters()
-        .into_iter()
-        .map(|(k, k_spks)| {
-            let mut once = Some(());
-            let mut stdout = std::io::stdout();
-            let k_spks = k_spks
-                .inspect(move |(spk_i, _)| match once.take() {
-                    Some(_) => print!("\nScanning keychain [{:?}]", k),
-                    None => print!(" {:<3}", spk_i),
-                })
```
LLFourn, line 51: I think the main thing this PR is missing is a way of restoring these prints. I don't think we can lose the ability to monitor progress here. This is the main technical challenge of this improvement. Let me know if you want to discuss. I'm not even sure what I'd do here but I have some ideas.

example-crates/wallet_esplora_async/src/main.rs
```diff
@@ -24,62 +32,93 @@ async fn main() -> Result<(), anyhow::Error> {
...
+            ))
+        }
+        Some(cmd) if cmd == "sync" => {
+            info!("Start sync...");
+            // 1. get data required to do a wallet sync, if also syncing previously used addresses set unused_spks_only = false
+            let request = wallet.sync_request(true);
+            // 2. sync unused wallet spks (addresses), unconfirmed tx, and utxos
+            let graph_update = client
+                .sync(
+                    request.spks,
```
LLFourn: I'm guessing the point of this PR is to make the esplora/electrum apis take the request as the argument here?

crates/bdk/src/wallet/mod.rs
notmandatory: Will remove.

crates/chain/src/request.rs
notmandatory: I updated the docs.

example-crates/wallet_esplora_async/src/main.rs
notmandatory: This will be tricky to do so I would like to discuss and get some advice.

example-crates/wallet_esplora_async/src/main.rs
notmandatory: Updated to use the request structs.

example-crates/wallet_esplora_async/src/main.rs
notmandatory: Can we push the feature to report sync/full_scan progress to a follow-up PR? I think this one is going to have enough in it to review.

LLFourn: I'd prefer not to lose features in PRs -- if we don't know how to restore this functionality then that's a deal breaker for this approach so it's better to sort it out now.

crates/esplora/src/blocking_ext.rs
```diff
@@ -139,22 +137,24 @@ impl EsploraExt for esplora_client::BlockingClient {
         })
     }
 
-    fn full_scan<K: Ord + Clone>(
+    fn full_scan<K: Ord + Clone + Send>(
         &self,
-        keychain_spks: BTreeMap<K, impl IntoIterator<Item = (u32, ScriptBuf)>>,
+        request: FullScanRequest<
+            K,
+            impl IntoIterator<IntoIter = impl Iterator<Item = (u32, ScriptBuf)> + Send> + Send + Clone,
```
LLFourn: This should just Box an iterator internally. No need for a type parameter. You could even use a concrete SpkIter I guess.

notmandatory:
A quick update. I've spent the past few days trying to figure out how to change the FullScanRequest to be able to add an inspect function for the SPKs but I can't figure out a way to do it. The main issue I run into if I try to use the same basic pattern as with SyncRequest is that I can't "capture" the inspect_spk function when I'm trying to add it to the iterators in the BTreeMap<K, I> that FullScanRequest holds.  I think this isn't causing a problem for SyncRequest because it uses collected vectors. I'm open to any ideas on how to do this, the complexity around closures, generics, and thread safety is exceeding my rust capabilities.

LLFourn:
@notmandatory to do it on `FullScanRequest` I think your going to need to store a `Box<dyn Iterator<...>>` for each keychain. Presumably you want to inspect each iterator with the same function. So something like:

```rust
self.spks_by_keychain = self.spks_by_keychain.drain(..).map(|keychain, spk_iter| Box::new(spk_iter.inspect(INSPECT_FN)).collect();
```
This is kinda like what we do now in the examples.

notmandatory:
I finally got this to build with an inspect function for the `FullScanRequest`. My next steps are to add tests, update examples to use inspect functions, and verify this API will work with work being done in bitcoindevkit/bdk-ffi#445.

example-crates/wallet_esplora_async/src/main.rs
notmandatory: I wasn't able to print the keychain because there isn't a way to have a generic K in the inspect function and capture it across threads. But I am able to use the index and spk in the function which I did for this example.

notmandatory:
I rebased this PR on #1380. In the process I also figured out a way to use the generic K for the function passed into `FullScanRequest::inspect_spks()`. To do it I had to wrap the Boxed function in an Arc and use the `Fn` type instead of `FnMut`. Before I was making a copy of the inspect function for every inspected item, this new way using a reference to the same function which should be more efficient.

example-crates/wallet_esplora_async/src/main.rs
notmandatory: 📌 Resolving this for now without adding a builder since that can be done later and this PR already has a lot to review.

crates/chain/src/request.rs
notmandatory: done.

crates/esplora/src/async_ext.rs
notmandatory: `SyncRequest` is now using `Vec`s and `FullScanRequest` is using `Iterator<Item = (u32, ScriptBuf)> + Send`.

notmandatory:
closing in favor of #1413

PR #1202, opened 2023-11-09
fix(chain): filter coinbase tx not in best chain

### LINKED ISSUE #1144: Coinbase transaction that is not anchored in the best chain should always be filtered out
**Describe the bug**  

Coinbase transactions cannot exist in the mempool and be unconfirmed. `TxGraph::try_get_chain_position` should always return `None` for coinbase transactions which are not anchored in the best chain.

**To Reproduce**  

A proper test has not been written for this scenario. The test would mine blocks which has the coinbase send to a tracked/owned spk. A reorg should not result in an unconfirmed balance.

**Expected behavior**  

Coinbase transactions cannot be "unconfirmed".

LagginTimes:
### Description

Fixes #1144.
Coinbase transactions cannot exist in the mempool and be unconfirmed. `TxGraph::try_get_chain_position` should always return `None` for coinbase transactions not anchored in best chain.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/tests/test_tx_graph_conflicts.rs
+++ b/crates/chain/tests/test_tx_graph_conflicts.rs
@@ -46,6 +46,49 @@ fn test_tx_conflict_handling() {
 
     let scenarios = [
         Scenario {
+            name: "coinbase tx cannot be in mempool and be unconfirmed",
+            tx_templates: &[
+                TxTemplate {
+                    tx_name: "unconfirmed_coinbase",
+                    inputs: &[TxInTemplate::Coinbase],
+                    outputs: &[TxOutTemplate::new(5000, Some(0))],
+                    ..Default::default()
+                },
+                TxTemplate {
+                    tx_name: "confirmed_genesis",
+                    inputs: &[TxInTemplate::Bogus],
+                    outputs: &[TxOutTemplate::new(10000, Some(1))],
+                    anchors: &[block_id!(1, "B")],
+                    last_seen: None,
+                },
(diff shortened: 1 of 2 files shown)


crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -45,6 +45,46 @@ fn test_tx_conflict_handling() {
         .unwrap_or_default();
 
     let scenarios = [
+        Scenario {
+            name: "coinbase tx cannot be in mempool and be unconfirmed",
+            tx_templates: &[
+                TxTemplate {
+                    tx_name: "coinbase",
```
danielabrozzoni: nit: I would call this "unconfirmed_coinbase"

danielabrozzoni (commented):
The code looks good to me.

In #1144, @evanlinjin says:

> A proper test has not been written for this scenario. The test would mine blocks which has the coinbase send to a tracked/owned spk. A reorg should not result in an unconfirmed balance.

Are we ok with the test that this PR includes, or better to also add the one evan is suggesting?

notmandatory (approved):
ACK 991cb77b6fbeedbf52d1bd9aa6b3d680f8269969

I verified new test fails without this fix. If an additional test is needed and not easy to do with current framework can we make a new issue and put in the beta milestone?

danielabrozzoni (approved):
ACK 991cb77b6fbeedbf52d1bd9aa6b3d680f8269969

PR #1204, opened 2023-11-11
chore: remove bdk dependency on log and dev dependency on env_logger

notmandatory:
### Description

As suggested by @TheBlueMatt we shouldn't use the `log` crate because even though it is a rust-lang project is may depend on some other crates that aren't as well maintained. Since we were only use `log` in a few places and not it any other bdk workspace projects except the `bdk` package I removed it. 

### Notes to the reviewers

For consistency I also removed the `env_logger` from the dev dependencies since it was only being used in a few places in a couple examples and println! is adequate for examples. 

### Changelog notice

None

#### All Submissions:

### DIFF
--- a/crates/bdk/Cargo.toml
+++ b/crates/bdk/Cargo.toml
@@ -13,7 +13,6 @@ edition = "2021"
 rust-version = "1.57"
 
 [dependencies]
-log = "0.4"
 rand = "^0.8"
 miniscript = { version = "10.0.0", features = ["serde"], default-features = false }
 bitcoin = { version = "0.30.0", features = ["serde", "base64", "rand-std"], default-features = false }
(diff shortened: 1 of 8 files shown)


evanlinjin (approved):
ACK 18e9e0a645384eddc3f66519cd0d5e0ff6eb3933

danielabrozzoni:
Should we open an issue to introduce some sort of diy logger as ldk does?

ACK 18e9e0a645384eddc3f66519cd0d5e0ff6eb3933

notmandatory:
> Should we open an issue to introduce some sort of diy logger as ldk does?

I just created a placeholder issue.

notmandatory:
rebased

PR #1207, opened 2023-11-13
ci: Pin jobserver after cc

danielabrozzoni:
Since now only cc depends on jobserver, it should be pinned after we pin cc

#### All Submissions:


notmandatory (approved):
ACK 298f6cb1e881806db82e40071169dfba91f8f95c

PR #1141, opened 2023-10-01
Refactor: Migrate bdk::FeeRate to bitcoin::FeeRate

junderw:
Closes #1136

### Description

This PR will eventually migrate FeeRate to rust-bitcoin's FeeRate.

### Notes to the reviewers

1. This is currently a WIP that is meant for discussion. I have made a lot of arbitrary decisions just so that we can see the full diff and go over each decision again as a team. Don't assume any decision was made with explicit intent, I just tried to make everything compile and tests pass. Everything is on the table for discussion.
2. ***If anyone wants to, feel free to take over this PR at any time.*** The legwork of figuring out all the places that need changing and which tests are affected took a few hours of trial and error. If you look at this PR and think you want to take over and finish off the more interesting parts (investigating the changes and why they were necessary / whether there are better more appropriate ways to change it) feel free. Just let me know. This issue is low priority for me personally, so don't expect this to get done in a quick manner unless someone takes it from me. I will be happy to answer questions about the new FeeRate API though.

#### Changed

- bdk::FeeRate was replaced with bitcoin::FeeRate

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #1216, opened 2023-11-15
Migrate to bitcoin::FeeRate

### LINKED ISSUE #1136: Replace bdk FeeRate with rust-bitcoin FeeRate
**Describe the enhancement**  

We should be able to replace the bdk `FeeRate` struct with the `FeeRate` from rust-bitcoin since they serve the same purpose and it's less confusing to users to have only one version of the type. Any missing functions from the bdk::FeeRate should be upstreamed to rust-bitcoin or moved to a new bdk trait implemented on rust-bitcoin's `FeeRate`. 

**Use case**  

All existing uses for bdk's `FeeRate`.

**Additional context**

https://docs.rs/bitcoin/0.30.1/bitcoin/blockdata/fee_rate/struct.FeeRate.html
https://docs.rs/bdk/1.0.0-alpha.1/bdk/struct.FeeRate.html

ValuedMammal:
### Description

This follows a similar approach to bitcoindevkit/bdk#1141 namely to remove `FeeRate` from `bdk::types` and instead defer to the upstream implementation for all fee rates. The idea is that making the switch allows BDK to benefit from a higher level abstraction, leaving the implementation details largely hidden.

As noted in bitcoindevkit/bdk#774, we should avoid extraneous conversions that can result in deviations in estimated transaction size and calculated fee amounts, etc. This would happen for example whenever calling a method like `FeeRate::to_sat_per_vb_ceil`. The only exception I would make is if we must return a fee rate error to the user, we might prefer to display it in the more familiar sats/vb, but this would only be useful if the rate can be expressed as a float.

### Notes to the reviewers

`bitcoin::FeeRate` is an integer whose native unit is sats per kilo-weight unit. In order to facilitate the change, a helper method `feerate_unchecked` is added and used only in wallet tests and psbt tests as necessary to convert existing fee rates to the new type. It's "unchecked" in the sense that we're not checking for integer overflow, because it's assumed we're passing a valid fee rate in a unit test.

Potential follow-ups can include:
- [x] Constructing a proper `FeeRate` from a `u64` in all unit tests, and thus obviating the need for the helper `feerate_unchecked` going forward.
- [x] Remove trait `Vbytes`.
- Consider adding an extra check that the argument to `TxBuilder::drain_to` is within "standard" size limits.
- Consider refactoring `coin_selection::select_sorted_utxos` to be efficient and readable.

closes bitcoindevkit/bdk#1136 

### Changelog notice

- Removed `FeeRate` type. All fee rates are now rust-bitcoin [`FeeRate`](https://docs.rs/bitcoin/latest/bitcoin/blockdata/fee_rate/struct.FeeRate.html).
- Removed trait `Vbytes`.

#### All Submissions:

### DIFF
--- a/crates/bdk/src/types.rs
+++ b/crates/bdk/src/types.rs
@@ -46,19 +46,6 @@ impl AsRef<[u8]> for KeychainKind {
     }
 }
 
-/// Trait implemented by types that can be used to measure weight units.
-pub trait Vbytes {
-    /// Convert weight units to virtual bytes.
-    fn vbytes(self) -> usize;
-}
-
-impl Vbytes for usize {
-    fn vbytes(self) -> usize {
-        // ref: https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#transaction-size-calculations
-        (self as f32 / 4.0).ceil() as usize
-    }
-}
-
 /// An unspent output owned by a [`Wallet`].
 ///
 /// [`Wallet`]: crate::Wallet
--- a/crates/bdk/src/types.rs
+++ b/crates/bdk/src/types.rs
@@ -11,11 +11,10 @@
 
 use alloc::boxed::Box;
 use core::convert::AsRef;
-use core::ops::Sub;
 
 use bdk_chain::ConfirmationTime;
 use bitcoin::blockdata::transaction::{OutPoint, Sequence, TxOut};
-use bitcoin::{psbt, Weight};
+use bitcoin::psbt;
 
 use serde::{Deserialize, Serialize};
 
@@ -44,103 +43,6 @@ impl AsRef<[u8]> for KeychainKind {
             KeychainKind::External => b"e",
             KeychainKind::Internal => b"i",
         }
-    }
-}
-
-/// Fee rate
-#[derive(Debug, Copy, Clone, PartialEq, PartialOrd)]
-// Internally stored as satoshi/vbyte
-pub struct FeeRate(f32);
-
-impl FeeRate {
-    /// Create a new instance checking the value provided
-    ///
-    /// ## Panics
-    ///
-    /// Panics if the value is not [normal](https://doc.rust-lang.org/std/primitive.f32.html#method.is_normal) (except if it's a positive zero) or negative.
-    fn new_checked(value: f32) -> Self {
-        assert!(value.is_normal() || value == 0.0);
-        assert!(value.is_sign_positive());
-
-        FeeRate(value)
-    }
-
-    /// Create a new instance of [`FeeRate`] given a float fee rate in sats/kwu
-    pub fn from_sat_per_kwu(sat_per_kwu: f32) -> Self {
-        FeeRate::new_checked(sat_per_kwu / 250.0_f32)
-    }
-
-    /// Create a new instance of [`FeeRate`] given a float fee rate in sats/kvb
-    pub fn from_sat_per_kvb(sat_per_kvb: f32) -> Self {
-        FeeRate::new_checked(sat_per_kvb / 1000.0_f32)
-    }
-
-    /// Create a new instance of [`FeeRate`] given a float fee rate in btc/kvbytes
-    ///
-    /// ## Panics
-    ///
-    /// Panics if the value is not [normal](https://doc.rust-lang.org/std/primitive.f32.html#method.is_normal) (except if it's a positive zero) or negative.
-    pub fn from_btc_per_kvb(btc_per_kvb: f32) -> Self {
-        FeeRate::new_checked(btc_per_kvb * 1e5)
-    }
-
-    /// Create a new instance of [`FeeRate`] given a float fee rate in satoshi/vbyte
-    ///
-    /// ## Panics
-    ///
-    /// Panics if the value is not [normal](https://doc.rust-lang.org/std/primitive.f32.html#method.is_normal) (except if it's a positive zero) or negative.
-    pub fn from_sat_per_vb(sat_per_vb: f32) -> Self {
-        FeeRate::new_checked(sat_per_vb)
-    }
-
-    /// Create a new [`FeeRate`] with the default min relay fee value
-    pub const fn default_min_relay_fee() -> Self {
-        FeeRate(1.0)
-    }
-
-    /// Calculate fee rate from `fee` and weight units (`wu`).
-    pub fn from_wu(fee: u64, wu: Weight) -> FeeRate {
-        Self::from_vb(fee, wu.to_vbytes_ceil() as usize)
-    }
-
-    /// Calculate fee rate from `fee` and `vbytes`.
-    pub fn from_vb(fee: u64, vbytes: usize) -> FeeRate {
-        let rate = fee as f32 / vbytes as f32;
-        Self::from_sat_per_vb(rate)
-    }
-
-    /// Return the value as satoshi/vbyte
-    pub fn as_sat_per_vb(&self) -> f32 {
-        self.0
-    }
-
-    /// Return the value as satoshi/kwu
-    pub fn sat_per_kwu(&self) -> f32 {
-        self.0 * 250.0_f32
-    }
-
-    /// Calculate absolute fee in Satoshis using size in weight units.
-    pub fn fee_wu(&self, wu: Weight) -> u64 {
-        self.fee_vb(wu.to_vbytes_ceil() as usize)
-    }
-
-    /// Calculate absolute fee in Satoshis using size in virtual bytes.
-    pub fn fee_vb(&self, vbytes: usize) -> u64 {
-        (self.as_sat_per_vb() * vbytes as f32).ceil() as u64
-    }
-}
-
-impl Default for FeeRate {
-    fn default() -> Self {
-        FeeRate::default_min_relay_fee()
-    }
-}
-
-impl Sub for FeeRate {
-    type Output = Self;
-
-    fn sub(self, other: FeeRate) -> Self::Output {
-        FeeRate(self.0 - other.0)
     }
 }
 
@@ -242,75 +144,5 @@ impl Utxo {
             Utxo::Local(_) => None,
             Utxo::Foreign { sequence, .. } => *sequence,
         }
-    }
-}
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn can_store_feerate_in_const() {
-        const _MIN_RELAY: FeeRate = FeeRate::default_min_relay_fee();
-    }
-
-    #[test]
-    #[should_panic]
-    fn test_invalid_feerate_neg_zero() {
-        let _ = FeeRate::from_sat_per_vb(-0.0);
-    }
-
-    #[test]
-    #[should_panic]
-    fn test_invalid_feerate_neg_value() {
-        let _ = FeeRate::from_sat_per_vb(-5.0);
-    }
-
-    #[test]
-    #[should_panic]
-    fn test_invalid_feerate_nan() {
-        let _ = FeeRate::from_sat_per_vb(f32::NAN);
-    }
-
-    #[test]
-    #[should_panic]
-    fn test_invalid_feerate_inf() {
-        let _ = FeeRate::from_sat_per_vb(f32::INFINITY);
-    }
-
-    #[test]
-    fn test_valid_feerate_pos_zero() {
-        let _ = FeeRate::from_sat_per_vb(0.0);
-    }
-
-    #[test]
-    fn test_fee_from_btc_per_kvb() {
-        let fee = FeeRate::from_btc_per_kvb(1e-5);
-        assert!((fee.as_sat_per_vb() - 1.0).abs() < f32::EPSILON);
-    }
-
-    #[test]
-    fn test_fee_from_sat_per_vbyte() {
-        let fee = FeeRate::from_sat_per_vb(1.0);
-        assert!((fee.as_sat_per_vb() - 1.0).abs() < f32::EPSILON);
-    }
-
-    #[test]
-    fn test_fee_default_min_relay_fee() {
-        let fee = FeeRate::default_min_relay_fee();
-        assert!((fee.as_sat_per_vb() - 1.0).abs() < f32::EPSILON);
-    }
-
-    #[test]
-    fn test_fee_from_sat_per_kvb() {
-        let fee = FeeRate::from_sat_per_kvb(1000.0);
-        assert!((fee.as_sat_per_vb() - 1.0).abs() < f32::EPSILON);
-    }
-
-    #[test]
-    fn test_fee_from_sat_per_kwu() {
-        let fee = FeeRate::from_sat_per_kwu(250.0);
-        assert!((fee.as_sat_per_vb() - 1.0).abs() < f32::EPSILON);
-        assert_eq!(fee.sat_per_kwu(), 250.0);
     }
 }
--- a/crates/bdk/src/wallet/coin_selection.rs
+++ b/crates/bdk/src/wallet/coin_selection.rs
@@ -41,7 +41,7 @@
 //!         &self,
 //!         required_utxos: Vec<WeightedUtxo>,
 //!         optional_utxos: Vec<WeightedUtxo>,
-//!         fee_rate: bdk::FeeRate,
+//!         fee_rate: FeeRate,
 //!         target_amount: u64,
 //!         drain_script: &Script,
 //!     ) -> Result<CoinSelectionResult, coin_selection::Error> {
@@ -61,7 +61,7 @@
 //!                 },
 //!             )
 //!             .collect::<Vec<_>>();
-//!         let additional_fees = fee_rate.fee_wu(additional_weight);
+//!         let additional_fees = (fee_rate * additional_weight).to_sat();
 //!         let amount_needed_with_fees = additional_fees + target_amount;
 //!         if selected_amount < amount_needed_with_fees {
 //!             return Err(coin_selection::Error::InsufficientFunds {
@@ -101,10 +101,10 @@
 //! ```
 
 use crate::chain::collections::HashSet;
-use crate::types::FeeRate;
 use crate::wallet::utils::IsDust;
 use crate::Utxo;
 use crate::WeightedUtxo;
+use bitcoin::FeeRate;
 
 use alloc::vec::Vec;
 use bitcoin::consensus::encode::serialize;
@@ -313,7 +313,8 @@ impl CoinSelectionAlgorithm for OldestFirstCoinSelection {
 pub fn decide_change(remaining_amount: u64, fee_rate: FeeRate, drain_script: &Script) -> Excess {
     // drain_output_len = size(len(script_pubkey)) + len(script_pubkey) + size(output_value)
     let drain_output_len = serialize(drain_script).len() + 8usize;
-    let change_fee = fee_rate.fee_vb(drain_output_len);
+    let change_fee =
+        (fee_rate * Weight::from_vb(drain_output_len as u64).expect("overflow occurred")).to_sat();
     let drain_val = remaining_amount.saturating_sub(change_fee);
 
     if drain_val.is_dust(drain_script) {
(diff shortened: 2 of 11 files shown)


ValuedMammal:
After looking at some [related work](https://github.com/rust-bitcoin/rust-bitcoin/pull/2064#issuecomment-1736297479), I'm not totally opposed to displaying only whole numbers of fee rates. The difference between 99 and 100 sat/vb seems less significant than the difference between 1 and 2 sat/vb.

Some other conveniences I think would be handy:
- Saturating addition/subtraction of `FeeRate`s
- A sanity check that a given rate when applied to a minimum valid tx size doesn't produce an amount greater than `MAX_MONEY`

crates/bdk/src/wallet/mod.rs
```diff
@@ -2385,6 +2385,17 @@ fn create_signers<E: IntoWalletDescriptor>(
...
+#[macro_export]
+#[doc(hidden)]
+macro_rules! floating_rate {
+    ($rate:expr) => {{
+        use $crate::bitcoin::blockdata::constants::WITNESS_SCALE_FACTOR;
+        // sat_kwu / 250.0 -> sat_vb
+        $rate.to_sat_per_kwu() as f64 / ((1000 / WITNESS_SCALE_FACTOR) as f64)
+    }};
+}
+
```
ValuedMammal, line 2576: suggest change doc to

```rust
/// Transforms a [`FeeRate`] to `f64` with unit of sat/vb
```

notmandatory:
>     * A sanity check that a given rate when applied to a minimum valid tx size doesn't produce an amount greater than `MAX_MONEY`

The fee and fee rate sanity check should be done in a new PR, I created bitcoindevkit/bdk_wallet#120 for this issue.

ValuedMammal:
Rebased to 553bdf92

- Resolved conflict with bitcoindevkit/bdk#1158 
- Fixed MSRV compile error

evanlinjin:
I think I have a controversial opinion about these changes. Because we plan to redo the entire tx building logic (after 1.0), and the current tx builder works okay as it is... Would this be worth the effort?

notmandatory:

ValuedMammal:
> I think I have a controversial opinion about these changes. Because we plan to redo the entire tx building logic (after 1.0), and the current tx builder works okay as it is... Would this be worth the effort?

I think it's fair to have a deeper discussion about this. I don't have strong feelings either way. I think I lean toward making the change because 1) ease of maintainability by aligning with rust bitcoin, 2) the units of a feerate can always be converted to any unit you like.

Granted I'm not as familiar with some of the long-standing design efforts of BDK, so I'm happy to get other opinions or hold off until everyone's in agreement.

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -313,7 +313,7 @@ impl CoinSelectionAlgorithm for OldestFirstCoinSelection {
 pub fn decide_change(remaining_amount: u64, fee_rate: FeeRate, drain_script: &Script) -> Excess {
     // drain_output_len = size(len(script_pubkey)) + len(script_pubkey) + size(output_value)
     let drain_output_len = serialize(drain_script).len() + 8usize;
-    let change_fee = fee_rate.fee_vb(drain_output_len);
+    let change_fee = (fee_rate * Weight::from_vb_unchecked(drain_output_len as u64)).to_sat();
```
storopoli: I think we should use [`from_vb`](https://docs.rs/bitcoin/latest/bitcoin/blockdata/weight/struct.Weight.html#method.from_vb) here instead of `from_vb_unchecked`.

`from_vb` will do additional checks for overflow and returns an `Option` that we can put a nice error message in `.expect()`.

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -456,7 +460,7 @@ impl CoinSelectionAlgorithm for BranchAndBoundCoinSelection {
             .iter()
             .fold(0, |acc, x| acc + x.effective_value);
 
-        let cost_of_change = self.size_of_change as f32 * fee_rate.as_sat_per_vb();
+        let cost_of_change = (Weight::from_vb_unchecked(self.size_of_change) * fee_rate).to_sat();
```
storopoli: Same as the other comment, `from_vb`

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -893,7 +897,7 @@ mod test {
             .coin_select(
                 utxos,
                 vec![],
-                FeeRate::from_sat_per_vb(1.0),
+                FeeRate::from_sat_per_vb_unchecked(1),
```
storopoli, line 901: These are fine to leave `*vb_unchecked` since they are tests

crates/bdk/src/wallet/mod.rs
```diff
@@ -1651,7 +1648,7 @@ impl<D> Wallet<D> {
     /// let mut psbt =  {
     ///     let mut builder = wallet.build_fee_bump(tx.txid())?;
     ///     builder
-    ///         .fee_rate(bdk::FeeRate::from_sat_per_vb(5.0));
+    ///         .fee_rate(FeeRate::from_sat_per_vb_unchecked(5));
```
storopoli: This is an example. I think we should use the more strict `from_sat_per_vb` to check for overflow.

Users will adhere to examples as the good practices for implementation.

crates/bdk/src/wallet/mod.rs
```diff
@@ -2563,6 +2560,17 @@ fn create_signers<E: IntoWalletDescriptor>(
     Ok((signers, change_signers))
 }
 
+/// Transforms a [`FeeRate`] to `f64` with unit of sat/vb.
```
storopoli:
```suggestion
/// Transforms a [`FeeRate`] to `f64` with unit as sat/vb.
```

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -31,7 +31,7 @@
 //!     // Create a transaction with one output to `to_address` of 50_000 satoshi
 //!     .add_recipient(to_address.script_pubkey(), 50_000)
 //!     // With a custom fee rate of 5.0 satoshi/vbyte
-//!     .fee_rate(bdk::FeeRate::from_sat_per_vb(5.0))
+//!     .fee_rate(FeeRate::from_sat_per_vb_unchecked(5))
```
storopoli: This is an example. I think we should use the more strict `from_sat_per_vb` to check for overflow.

Users will adhere to examples as the good practices for implementation.

crates/bdk/tests/psbt.rs
```diff
@@ -10,6 +11,12 @@ use common::*;
 // from bip 174
 const PSBT_STR: &str = "cHNidP8BAKACAAAAAqsJSaCMWvfEm4IS9Bfi8Vqz9cM9zxU4IagTn4d6W3vkAAAAAAD+////qwlJoIxa98SbghL0F+LxWrP1wz3PFTghqBOfh3pbe+QBAAAAAP7///8CYDvqCwAAAAAZdqkUdopAu9dAy+gdmI5x3ipNXHE5ax2IrI4kAAAAAAAAGXapFG9GILVT+glechue4O/p+gOcykWXiKwAAAAAAAEHakcwRAIgR1lmF5fAGwNrJZKJSGhiGDR9iYZLcZ4ff89X0eURZYcCIFMJ6r9Wqk2Ikf/REf3xM286KdqGbX+EhtdVRs7tr5MZASEDXNxh/HupccC1AaZGoqg7ECy0OIEhfKaC3Ibi1z+ogpIAAQEgAOH1BQAAAAAXqRQ1RebjO4MsRwUPJNPuuTycA5SLx4cBBBYAFIXRNTfy4mVAWjTbr6nj3aAfuCMIAAAA";
 
+fn _feerate_unchecked(sat_vb: f64) -> FeeRate {
+    // 1 sat_vb / 4wu_vb * 1000kwu_wu = 250 sat_kwu
+    let sat_kwu = (sat_vb * 250.0).ceil() as u64;
+    FeeRate::from_sat_per_kwu(sat_kwu)
+}
+
```
storopoli: I might be missing something but I don't think you are using this function.

crates/bdk/tests/wallet.rs
```diff
@@ -55,6 +55,12 @@ fn receive_output_in_latest_block(wallet: &mut Wallet, value: u64) -> OutPoint {
     receive_output(wallet, value, anchor)
 }
 
+fn _feerate_unchecked(sat_vb: f64) -> FeeRate {
+    // 1 sat_vb / 4wu_vb * 1000kwu_wu = 250 sat_kwu
+    let sat_kwu = (sat_vb * 250.0).ceil() as u64;
+    FeeRate::from_sat_per_kwu(sat_kwu)
+}
+
```
storopoli: I might be missing something but I don't think you are using this function.

crates/bdk/tests/wallet.rs
```diff
@@ -753,7 +765,7 @@ fn test_create_tx_drain_to_dust_amount() {
     builder
         .drain_to(addr.script_pubkey())
         .drain_wallet()
-        .fee_rate(FeeRate::from_sat_per_vb(453.0));
+        .fee_rate(FeeRate::from_sat_per_vb_unchecked(499));
```
storopoli: Why changing from 453 to 499?

crates/bdk/tests/wallet.rs
```diff
@@ -1568,8 +1590,9 @@ fn test_bump_fee_reduce_change() {
         .insert_tx(tx, ConfirmationTime::Unconfirmed { last_seen: 0 })
         .unwrap();
 
+    let feerate = FeeRate::from_sat_per_kwu(625); // 2.5 sat/vb
```
storopoli, line 1587: The math is not matching for me.
Am I doing something wrong?

`dbg!(psbt.clone().extract_tx().weight().to_wu() as f64 * 2.5);`

returns `1140.0`

crates/bdk/tests/wallet.rs
```diff
@@ -1663,9 +1686,10 @@ fn test_bump_fee_reduce_single_recipient() {
...
```
storopoli, line 1689: The math is not matching for me.
Am I doing something wrong?

`dbg!(psbt.clone().extract_tx().weight().to_wu() as f64 * 2.5);`

returns `830.0`

crates/bdk/tests/wallet.rs
```diff
@@ -2298,7 +2322,7 @@ fn test_fee_amount_negative_drain_val() {
     let send_to = Address::from_str("tb1ql7w62elx9ucw4pj5lgw4l028hmuw80sndtntxt")
         .unwrap()
         .assume_checked();
-    let fee_rate = FeeRate::from_sat_per_vb(2.01);
+    let fee_rate = FeeRate::from_sat_per_kwu(500);
```
storopoli, line 2325: Again, I am trying to follow the math:

`dbg!(psbt.clone().extract_tx().weight().to_wu() as f64 * 2.01)` gives `659.28`

storopoli (changes requested):
Some changes proposed.
Mostly important is to use [`from_vb`](https://docs.rs/bitcoin/latest/bitcoin/blockdata/weight/struct.Weight.html#method.from_vb) instead of `from_vb_unchecked` in code and examples.

Tests are fine to use the unchecked version.

ValuedMammal:
Thank you @storopoli , many good points brought up. I'm working through some of the comments now

crates/bdk/src/wallet/mod.rs
ValuedMammal: will fix

crates/bdk/tests/psbt.rs
ValuedMammal: This was kept in case someone in the future feels the need to construct a `bitcoin::FeeRate` from `f64`, since `bitcoin::FeeRate::from_sat_per_vb` only accepts an integer as param. We can also remove it now if you think it won't be missed.

crates/bdk/tests/wallet.rs
ValuedMammal: Admittedly, this was a hack to make the test pass, and `499` was the first value chosen because it gives the impression of a "very high fee rate." The original `453` must have been right on the edge of passing. Evidently `454` would make this pass as well.

crates/bdk/tests/wallet.rs
ValuedMammal: It looks like you're multiplying the tx weight by 2.5, which doesn't result in a feerate. To arrive at units of sat/kwu, you could do something like

`fee_amount * 1000 / tx.weight().to_wu()`

This will be more accurate if you `extract_tx` only after it has been signed.

crates/bdk/src/wallet/coin_selection.rs
ValuedMammal: I agree this is a concern. My opinion is that an insanely sized `drain_script` would overflow the block size before overflowing a `u64`, so this check should ideally occur during tx building and throw an error if the drain script exceeds either `MAX_STANDARD_TX_WEIGHT` (400kwu) or more generously `MAX_BLOCK_WEIGHT` (4Mwu). Note that if the drain script is derived from a typical bitcoin address, then this should never happen. I think adding these sanity checks early in tx building would make the code more robust, instead of cramming the logic into this particular line. What would be your take on that?

I would apply the same reasoning to `size_of_change` in the constructor of `BranchAndBoundCoinSelection`, but from what I hear, the coin select module is soon to be replaced so I don't know how relevant this is now.

In any case, your suggestion is sound, so we can either fixup this line of code, or I can add a note in the PR description about adding these extra consensus/standardness checks as a followup.

crates/bdk/tests/psbt.rs
storopoli: No, this is good. It's in the test suite.
Let's add some docstrings then what you think?

crates/bdk/tests/wallet.rs
storopoli: Let's add some docstrings here as well, what you think?

crates/bdk/tests/wallet.rs
storopoli: I can confirm that 454 passes the tests, can we use that?

```suggestion
        .fee_rate(FeeRate::from_sat_per_vb_unchecked(454));
```

crates/bdk/tests/wallet.rs
storopoli: Thanks!

crates/bdk/src/wallet/coin_selection.rs
ValuedMammal: I may be overthinking. We can do both - check for overflow here, and make a note to check the drain script size elsewhere

storopoli: I don't think we need to check for overflow, in an error handling situation.
Just an `.expect()` with a nice error message would be sufficient in my opinion.
It will give enough context for the user of the source of the error.

crates/bdk/src/types.rs
```diff
@@ -247,70 +149,5 @@ impl Utxo {
...
-        assert!((fee.as_sat_per_vb() - 1.0).abs() < f32::EPSILON);
-    }
-
-    #[test]
-    fn test_fee_from_sat_per_kwu() {
-        let fee = FeeRate::from_sat_per_kwu(250.0);
-        assert!((fee.as_sat_per_vb() - 1.0).abs() < f32::EPSILON);
-        assert_eq!(fee.sat_per_kwu(), 250.0);
-    }
+    //use super::*;
```
storopoli: I don't think this adds anything. Let's remove it instead of commenting?

storopoli (changes requested):
Just a minor thing.

crates/bdk/src/types.rs
ValuedMammal: noted

ValuedMammal:
Recent changes:

- Using checked vsize-to-weight conversions in documentation and during coin selection.
- Added docstring to the test helper `feerate_unchecked` to explain its purpose, and moved it to bdk/tests/common.rs

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -344,9 +345,12 @@ fn select_sorted_utxos(
             (&mut selected_amount, &mut fee_amount),
```
evanlinjin, line 345: I have a few thoughts here (I know this was like this before this PR, but I think we should address them).

1. Use a `for` loop instead of `.scan`. `.scan` makes this hard to read.
2. Because multiplying weight with feerate rounds up, we should keep an aggregate selection weight (instead of an aggregate `fee_amount`) and calculate the `fee_amount` on every iteration with the aggregate selection weight. This will give a more accurate result. Ref: https://docs.rs/bitcoin/latest/src/bitcoin/blockdata/fee_rate.rs.html#137-143.
3. This method needs doc comments. This methods requires that the `utxos` input is sorted by must-be-used UTXOs first otherwise it won't work as expected.

For 2., it is to be noted that what we have in this PR is already a lot better than what we had before (since our old `FeeRate::fee_wu` was a lot more aggressive when rounding up than the `FeeRate as Mul<Weight>` impl).

---

Edit: Ignore this! This PR is strictly better than what we had before. Also for point no. 2, we can't do the same for `OutputGroup` anyway so we shouldn't bother ourselves here.

In the future, we should focus our efforts on implementing our new `TxBuilder` than maintaining this old code that is to be scrapped.

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -387,9 +391,10 @@ struct OutputGroup {
 
 impl OutputGroup {
     fn new(weighted_utxo: WeightedUtxo, fee_rate: FeeRate) -> Self {
-        let fee = fee_rate.fee_wu(Weight::from_wu(
-            (TXIN_BASE_WEIGHT + weighted_utxo.satisfaction_weight) as u64,
-        ));
+        let fee = (fee_rate
+            * Weight::from_wu((TXIN_BASE_WEIGHT + weighted_utxo.satisfaction_weight) as u64))
```
evanlinjin, line 395: ~~Nit: Since we are doing this in multiple places, I think we should create a method `WeightedUtxo::txin_weight` that returns `TXIN_BASE_WEIGHT + self.satisfaction_weight`.~~

Edit: Ignore this also.

crates/bdk/src/types.rs
evanlinjin, line 1: We can also get rid of `Vbytes` now since it is only used in one test. Yay!

evanlinjin (commented):
Thanks for working on this, fee/weight calculations should be more accurate than before since the old `FeeRate::fee_wu` was rounding up virtual bytes to whole units, whereas the `Mul<FeeRate> for Weight` impl (which we are using now) is rounding up kilo-wus.

However, I think we should try our best to avoid these changes in the future. This took a long time to review for me as I had to revisit old code to understand the ramifications of these changes. If there is no bug, we should just leave it as is so we can focus our efforts on the new `TxBuilder`.

As a final nit, let's get rid of `Vbytes` and I will happily ACK this!

crates/bdk/src/wallet/coin_selection.rs
ValuedMammal: I agree this is a bit difficult to make sense of.

ValuedMammal:
- Rebased onto d77a7f2ff13419d394e6a19f9942344a7c19a724
- Removed trait Vbytes in 011841905822a4ce58469bf1128cd018c85ce6d8

crates/bdk/tests/common.rs
```diff
@@ -154,3 +154,16 @@ pub fn get_test_tr_with_taptree_xprv() -> &'static str {
...
 }
+
+/// Construct a new [`FeeRate`] from the given raw `sat_vb` feerate. This is
+/// useful in cases where we want to create a feerate from a `f64`, as the
+/// traditional [`FeeRate::from_sat_per_vb`] method will only accept an integer.
+///
+/// **Note** this 'quick and dirty' conversion should only be used when the input
+/// parameter has units of `satoshis/vbyte` **AND** is not expected to overflow,
+/// or else the resulting value will be inaccurate.
+fn feerate_unchecked(sat_vb: f64) -> FeeRate {
```
ValuedMammal: I guess this should be made `pub` for it to be at all useful.

ValuedMammal:
> Posting clippy checks here instead.
error: file opened with `create`, but `truncate` behavior not defined
  --> crates/file_store/src/store.rs:64:14
   |
64 |             .create(true)
   |              ^^^^^^^^^^^^- help: add: `.truncate(true)`
   |
   = help: if you intend to overwrite an existing file entirely, call `.truncate(true)`

@evanlinjin ?

evanlinjin:
@ValuedMammal rebase on master should fix it now!

evanlinjin (approved):
ACK 475a77219a6e7649c91b7cfeb44871cee9d0f374

PR #1225, opened 2023-11-20
esplora: fix incorrect gap limit check in blocking client

darosior:
The gap limit was checked such as if the last_index was not None but the last_active_index was, we'd consider having reached it. But the last_index is never None for this check. This effectively made it so the gap limit was always 1: if the first address isn't used last_active_index will be None and we'd return immediately.

Fix this by avoiding error-prone Option comparisons and correctly checking we've reached the gap limit before breaking out of the loop.

### DIFF
--- a/crates/esplora/src/async_ext.rs
+++ b/crates/esplora/src/async_ext.rs
@@ -259,7 +259,13 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
                     }
                 }
 
-                if last_index > last_active_index.map(|i| i.saturating_add(stop_gap as u32)) {
+                let last_index = last_index.expect("Must be set since handles wasn't empty.");
+                let past_gap_limit = if let Some(i) = last_active_index {
+                    last_index > i.saturating_add(stop_gap as u32)
+                } else {
+                    last_index >= stop_gap as u32
+                };
+                if past_gap_limit {
                     break;
                 }
             }
--- a/crates/esplora/src/blocking_ext.rs
+++ b/crates/esplora/src/blocking_ext.rs
@@ -250,7 +250,13 @@ impl EsploraExt for esplora_client::BlockingClient {
                     }
                 }
 
-                if last_index > last_active_index.map(|i| i.saturating_add(stop_gap as u32)) {
+                let last_index = last_index.expect("Must be set since handles wasn't empty.");
+                let past_gap_limit = if let Some(i) = last_active_index {
+                    last_index > i.saturating_add(stop_gap as u32)
+                } else {
+                    last_index >= stop_gap as u32
+                };
+                if past_gap_limit {
                     break;
                 }
             }
--- a/crates/esplora/tests/async_ext.rs
+++ b/crates/esplora/tests/async_ext.rs
@@ -3,6 +3,7 @@ use electrsd::bitcoind::bitcoincore_rpc::RpcApi;
 use electrsd::bitcoind::{self, anyhow, BitcoinD};
 use electrsd::{Conf, ElectrsD};
 use esplora_client::{self, AsyncClient, Builder};
+use std::collections::{BTreeMap, HashSet};
 use std::str::FromStr;
 use std::thread::sleep;
 use std::time::Duration;
@@ -113,5 +114,123 @@ pub async fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
     let mut expected_txids = vec![txid1, txid2];
     expected_txids.sort();
     assert_eq!(graph_update_txids, expected_txids);
+    Ok(())
+}
+
+/// Test the bounds of the address scan depending on the gap limit.
+#[tokio::test]
+pub async fn test_async_update_tx_graph_gap_limit() -> anyhow::Result<()> {
+    let env = TestEnv::new()?;
+    let _block_hashes = env.mine_blocks(101, None)?;
+
+    // Now let's test the gap limit. First of all get a chain of 10 addresses.
+    let addresses = [
+        "bcrt1qj9f7r8r3p2y0sqf4r3r62qysmkuh0fzep473d2ar7rcz64wqvhssjgf0z4",
+        "bcrt1qmm5t0ch7vh2hryx9ctq3mswexcugqe4atkpkl2tetm8merqkthas3w7q30",
+        "bcrt1qut9p7ej7l7lhyvekj28xknn8gnugtym4d5qvnp5shrsr4nksmfqsmyn87g",
+        "bcrt1qqz0xtn3m235p2k96f5wa2dqukg6shxn9n3txe8arlrhjh5p744hsd957ww",
+        "bcrt1q9c0t62a8l6wfytmf2t9lfj35avadk3mm8g4p3l84tp6rl66m48sqrme7wu",
+        "bcrt1qkmh8yrk2v47cklt8dytk8f3ammcwa4q7dzattedzfhqzvfwwgyzsg59zrh",
+        "bcrt1qvgrsrzy07gjkkfr5luplt0azxtfwmwq5t62gum5jr7zwcvep2acs8hhnp2",
+        "bcrt1qw57edarcg50ansq8mk3guyrk78rk0fwvrds5xvqeupteu848zayq549av8",
+        "bcrt1qvtve5ekf6e5kzs68knvnt2phfw6a0yjqrlgat392m6zt9jsvyxhqfx67ef",
+        "bcrt1qw03ddumfs9z0kcu76ln7jrjfdwam20qtffmkcral3qtza90sp9kqm787uk",
+    ];
+    let addresses: Vec<_> = addresses
+        .into_iter()
+        .map(|s| Address::from_str(s).unwrap().assume_checked())
+        .collect();
+    let spks: Vec<_> = addresses
+        .iter()
+        .enumerate()
+        .map(|(i, addr)| (i as u32, addr.script_pubkey()))
+        .collect();
+    let mut keychains = BTreeMap::new();
+    keychains.insert(0, spks);
+
+    // Then receive coins on the 4th address.
+    let txid_4th_addr = env.bitcoind.client.send_to_address(
+        &addresses[3],
+        Amount::from_sat(10000),
+        None,
+        None,
+        None,
+        None,
+        Some(1),
+        None,
+    )?;
+    let _block_hashes = env.mine_blocks(1, None)?;
+    while env.client.get_height().await.unwrap() < 103 {
+        sleep(Duration::from_millis(10))
+    }
+
+    // A scan with a gap limit of 2 won't find the transaction, but a scan with a gap limit of 3
+    // will.
+    let (graph_update, active_indices) = env
+        .client
+        .scan_txs_with_keychains(
+            keychains.clone(),
+            vec![].into_iter(),
+            vec![].into_iter(),
+            2,
+            1,
+        )
+        .await?;
+    assert!(graph_update.full_txs().next().is_none());
+    assert!(active_indices.is_empty());
+    let (graph_update, active_indices) = env
+        .client
+        .scan_txs_with_keychains(
+            keychains.clone(),
+            vec![].into_iter(),
+            vec![].into_iter(),
+            3,
+            1,
+        )
+        .await?;
+    assert_eq!(graph_update.full_txs().next().unwrap().txid, txid_4th_addr);
+    assert_eq!(active_indices[&0], 3);
+
+    // Now receive a coin on the last address.
+    let txid_last_addr = env.bitcoind.client.send_to_address(
+        &addresses[addresses.len() - 1],
+        Amount::from_sat(10000),
+        None,
+        None,
+        None,
+        None,
+        Some(1),
+        None,
+    )?;
+    let _block_hashes = env.mine_blocks(1, None)?;
+    while env.client.get_height().await.unwrap() < 104 {
+        sleep(Duration::from_millis(10))
+    }
+
+    // A scan with gap limit 4 won't find the second transaction, but a scan with gap limit 5 will.
+    // The last active indice won't be updated in the first case but will in the second one.
+    let (graph_update, active_indices) = env
+        .client
+        .scan_txs_with_keychains(
+            keychains.clone(),
+            vec![].into_iter(),
+            vec![].into_iter(),
+            4,
+            1,
+        )
+        .await?;
+    let txs: HashSet<_> = graph_update.full_txs().map(|tx| tx.txid).collect();
+    assert_eq!(txs.len(), 1);
+    assert!(txs.contains(&txid_4th_addr));
+    assert_eq!(active_indices[&0], 3);
+    let (graph_update, active_indices) = env
+        .client
+        .scan_txs_with_keychains(keychains, vec![].into_iter(), vec![].into_iter(), 5, 1)
+        .await?;
+    let txs: HashSet<_> = graph_update.full_txs().map(|tx| tx.txid).collect();
+    assert_eq!(txs.len(), 2);
+    assert!(txs.contains(&txid_4th_addr) && txs.contains(&txid_last_addr));
+    assert_eq!(active_indices[&0], 9);
+
     Ok(())
 }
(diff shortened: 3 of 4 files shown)


darosior:
It's a bit surprising this bug wouldn't be caught by any test. I'll look into adding one.

tnull:
From a real quick comparison in looks like the changed code is part of `async_ext.rs`. Does the async variant therefore also suffer from the same issue and needs to be fixed?

crates/esplora/src/blocking_ext.rs
```diff
@@ -250,7 +250,8 @@ impl EsploraExt for esplora_client::BlockingClient {
                     }
                 }
 
-                if last_index > last_active_index.map(|i| i.saturating_add(stop_gap as u32)) {
+                let last_index = last_index.expect("Must be set since handles wasn't empty.");
+                if last_index > last_active_index.unwrap_or(0) + stop_gap as u32 {
```
darosior:
```suggestion
                if last_index > last_active_index.unwrap_or(0).saturating_add(stop_gap as u32) {
```
Since the version without keychain set the gap limit to uint32::MAX.

darosior:
> Does the async variant therefore also suffer from the same issue and needs to be fixed?

Probably, but to be honest i'm pretty limited in time at the moment so i'm only writing a unit test to demonstrate the bug for the blocking client. I'd prefer if another contributor took care of fixing the async client.

darosior:
Pushed a unit test exercising this logic as a separate commit, so it can be cherry-picked on master to demonstrate the bug.

If somebody wants to take care of fixing the async client they can reuse the same logic from this test, i think.

darosior:
Ok so actually porting the fix and the test to the async client was actually trivial, so i just did it here. The same goes for the async unit test: it can be cherry-picked on master to demonstrate the bug.

LLFourn:
> It's a bit surprising this bug wouldn't be caught by any test. I'll look into adding one.

I can see you've done this thanks. I think we are in the middle of setting up a proper test environment for this crate in #1171. I think you can leave @evanlinjin and @LagginTimes to polish of this PR after that.

crates/esplora/tests/async_ext.rs
```diff
@@ -115,3 +116,121 @@ pub async fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
...
+        None,
+        Some(1),
+        None,
+    )?;
+    let _block_hashes = env.mine_blocks(1, None)?;
+    while env.client.get_height().await.unwrap() < 103 {
+        sleep(Duration::from_millis(10))
+    }
+
+    // A scan with a gap limit of 1 won't find the transaction, but a scan with a gap limit of 2
```
danielabrozzoni: I might have a wrong definition of stop_gap, but I've always thought of it as "how many empty spks I go through before I quit searching". Given my definition, a stop_gap of 4 should find the transaction, and not of 2? Because a stop_gap of 3 would look at spk0, spk1, spk2, and then quit; while a stop_gap of 4 would look at spk0, spk1, spk2, spk3 (<- find a tx here! :)), spk4, spk5, spk6, spk7, and then quit.

crates/esplora/src/async_ext.rs
```diff
@@ -259,7 +259,12 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
                     }
                 }
 
-                if last_index > last_active_index.map(|i| i.saturating_add(stop_gap as u32)) {
+                let last_index = last_index.expect("Must be set since handles wasn't empty.");
+                if last_index
+                    > last_active_index
+                        .unwrap_or(0)
```
danielabrozzoni: Is this right? A last_active_index of `None` and a `last_active_index` of 0 are not the same, they differ by 1 spk seen. Maybe this should be `unwrap_or(-1)`?

EDIT: well you can't `unwrap_or(-1)` as this is a u32, but still, I think the two cases should be differentiated

danielabrozzoni (commented):
I'm not very knowledgeable on the esplora code, I have a few questions

crates/esplora/tests/async_ext.rs
darosior: Yes, it seemed a bit off to me as well: it seems to be finding more than it should. But i didn't want to fix this here so i just tested that there is indeed a threshold at which we don't find it, but then we do find it. (But not on master.)

It's also unclear whether we want to be fixing this: it seems good to be conservative with this.

crates/esplora/src/async_ext.rs
darosior: Heh, true. Can't use -1 though as it's unsigned. So it looks like i need to do something like this to be accurate (and still clear):
```diff
diff --git a/crates/esplora/src/blocking_ext.rs b/crates/esplora/src/blocking_ext.rs
index 7ee4625..bde24f8 100644
--- a/crates/esplora/src/blocking_ext.rs
+++ b/crates/esplora/src/blocking_ext.rs
@@ -251,11 +251,12 @@ impl EsploraExt for esplora_client::BlockingClient {
                 }
 
                 let last_index = last_index.expect("Must be set since handles wasn't empty.");
-                if last_index
-                    > last_active_index
-                        .unwrap_or(0)
-                        .saturating_add(stop_gap as u32)
-                {
+                let past_gap_limit = if let Some(i) = last_active_index {
+                    last_index > i.saturating_add(stop_gap as u32)
+                } else {
+                    last_index >= stop_gap as u32
+                };
+                if past_gap_limit {
                     break;
                 }
             }
```

However looks like this uncovered a crash now..
```
running 1 test
test test_update_tx_graph_gap_limit ... FAILED

failures:

---- test_update_tx_graph_gap_limit stdout ----
thread 'test_update_tx_graph_gap_limit' panicked at crates/esplora/tests/blocking_ext.rs:183:47:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:595:5
   1: core::panicking::panic_fmt
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:67:14
   2: core::panicking::panic
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:117:5
   3: core::option::Option<T>::unwrap
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/option.rs:935:21
   4: blocking_ext::test_update_tx_graph_gap_limit
             at ./tests/blocking_ext.rs:183:16
   5: blocking_ext::test_update_tx_graph_gap_limit::{{closure}}
             at ./tests/blocking_ext.rs:120:44
   6: core::ops::function::FnOnce::call_once
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/ops/function.rs:250:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    test_update_tx_graph_gap_limit
```

danielabrozzoni: Interesting! I'm looking into it. I want to merge your fix, but first I want to make sure that we're not introducing a bigger bug.

darosior: Of course! Thanks for looking into it.

danielabrozzoni: The failure you see is in the tests, because you removed one off-by-one (the `unwrap_or(0)`), but there's another off-by-one in the code posted.
I started by updating the tests following my comment https://github.com/bitcoindevkit/bdk/pull/1225#discussion_r1402460854, so that we don't find the tx to spk3 with stop_gap = 3, but we do find it with stop_gap = 4; and we don't find the tx to spk9 with a stop_gap of 5, but we find it with a stop_gap of 6:
```diff
diff --git a/crates/esplora/tests/async_ext.rs b/crates/esplora/tests/async_ext.rs
index ec79cd07..409daf04 100644
--- a/crates/esplora/tests/async_ext.rs
+++ b/crates/esplora/tests/async_ext.rs
@@ -164,7 +164,7 @@ pub async fn test_async_update_tx_graph_gap_limit() -> anyhow::Result<()> {
         sleep(Duration::from_millis(10))
     }
 
-    // A scan with a gap limit of 1 won't find the transaction, but a scan with a gap limit of 2
+    // A scan with a gap limit of 3 won't find the transaction, but a scan with a gap limit of 4
     // will.
     let (graph_update, active_indices) = env
         .client
@@ -172,7 +172,7 @@ pub async fn test_async_update_tx_graph_gap_limit() -> anyhow::Result<()> {
             keychains.clone(),
             vec![].into_iter(),
             vec![].into_iter(),
-            1,
+            3,
             1,
         )
         .await?;
@@ -184,7 +184,7 @@ pub async fn test_async_update_tx_graph_gap_limit() -> anyhow::Result<()> {
             keychains.clone(),
             vec![].into_iter(),
             vec![].into_iter(),
-            2,
+            4,
             1,
         )
         .await?;
@@ -207,7 +207,7 @@ pub async fn test_async_update_tx_graph_gap_limit() -> anyhow::Result<()> {
         sleep(Duration::from_millis(10))
     }
 
-    // A scan with gap limit 4 won't find the second transaction, but a scan with gap limit 5 will.
+    // A scan with gap limit 5 won't find the second transaction, but a scan with gap limit 6 will.
     // The last active indice won't be updated in the first case but will in the second one.
     let (graph_update, active_indices) = env
         .client
@@ -215,7 +215,7 @@ pub async fn test_async_update_tx_graph_gap_limit() -> anyhow::Result<()> {
             keychains.clone(),
             vec![].into_iter(),
             vec![].into_iter(),
-            4,
+            5,
             1,
         )
         .await?;
@@ -225,7 +225,7 @@ pub async fn test_async_update_tx_graph_gap_limit() -> anyhow::Result<()> {
     assert_eq!(active_indices[&0], 3);
     let (graph_update, active_indices) = env
         .client
-        .scan_txs_with_keychains(keychains, vec![].into_iter(), vec![].into_iter(), 5, 1)
+        .scan_txs_with_keychains(keychains, vec![].into_iter(), vec![].into_iter(), 6, 1)
         .await?;
     let txs: HashSet<_> = graph_update.full_txs().map(|tx| tx.txid).collect();
     assert_eq!(txs.len(), 2);
```

Then, the code should be modified to:
```
+                let past_gap_limit = if let Some(i) = last_active_index {
+                    last_index >= i.saturating_add(stop_gap as u32)
+                } else {
+                    // If the last active is None, last_index + 1 is the number of spks we searched
+                    // If we searched `stop_gap` or more spks without finding a tx, then we're past
+                    // the gap_limit
+                    last_index + 1 >= stop_gap as u32
+                };
```

Using `>=` instead of `>` in the first arm, and `last_index + 1` in the second arm.

Does this sound fair?

crates/esplora/src/blocking_ext.rs
```diff
@@ -250,7 +250,12 @@ impl EsploraExt for esplora_client::BlockingClient {
                     }
                 }
 
-                if last_index > last_active_index.map(|i| i.saturating_add(stop_gap as u32)) {
+                let last_index = last_index.expect("Must be set since handles wasn't empty.");
+                if last_index
+                    > last_active_index
+                        .unwrap_or(0)
+                        .saturating_add(stop_gap as u32)
+                {
```
evanlinjin:
```suggestion
                let stop_index = match last_active_index {
                    Some(active_index) => (stop_gap as u32)
                        .saturating_add(1)
                        .saturating_add(active_index),
                    None => stop_gap as u32,
                };
                if last_index >= stop_index {
```

A stop gap of 0 expects a tx history for every spk. Therefore, if we have `stop_gap = 0`, `last_index = 0` and `last_active_index = None`, we expect to break here.

crates/esplora/tests/blocking_ext.rs
```diff
@@ -110,5 +111,118 @@ pub fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
...
+        1,
+        1,
+    )?;
+    assert!(graph_update.full_txs().next().is_none());
+    assert!(active_indices.is_empty());
+    let (graph_update, active_indices) = env.client.scan_txs_with_keychains(
+        keychains.clone(),
+        vec![].into_iter(),
+        vec![].into_iter(),
+        2,
```
evanlinjin:
```suggestion
        3,
```

Because with a `stop_gap` of 3, we allow 3 consecutive spks with no history. This will be for spk indices 0, 1 and 2.

evanlinjin (changes requested):
Thanks for spotting this major bug and quickly implementing the fix with tests to go with. The changes here are strictly better than what we had before.

However, `stop_gap` is now implemented slightly incorrectly. If you do not have time to make these changes, I could do it for you!

evanlinjin:
~@danielabrozzoni sorry I didn't see your review before posting mine. Guess we are talking about the same thing haha.~

Okay I think we both agree that `stop_gap` is not implemented correctly, but I think my suggestion is correcter.

crates/esplora/src/blocking_ext.rs
darosior: ~~See https://github.com/bitcoindevkit/bdk/pull/1225#discussion_r1402485388, fixing this uncovers a crash. I'm checking Daniela's suggestion now.~~

My bad i misread the crash as being in `src/blocking_ext.rs` but it's in fact in `tests/blocking_ext.rs`. So your suggestion is the same as mine here https://github.com/bitcoindevkit/bdk/pull/1225#discussion_r1402485388 and i think is correct.

crates/esplora/tests/blocking_ext.rs
danielabrozzoni: https://electrum.readthedocs.io/en/latest/faq.html#what-is-the-gap-limit
https://docs.btcpayserver.org/FAQ/Wallet/#the-gap-limit-problem

I think this means that the stop gap should be 4? If you have a stop gap of 3, you generate 3 addresses without history (spk 0, 1, 2) and then you stop.

crates/esplora/src/async_ext.rs
darosior: Partially. You are right for the case where there is a last active index. I had the logic off by one, i was considering the next index. But i think you are wrong for the case where there is no last active index. For instance take a gap limit of 1. Your code would make us break if we just scanned index 0 unsuccessfully, although we should also scan index 1. (Only scanning index 0 is gap limit 0.)

darosior: No actually i think you are also wrong in the first case. Take a gap limit of 1. If our last active index is 4 and we just scanned index 5 unsuccessfully we should also scan index 6. (Only scanning index 5 is gap limit 0.)

So i think my patch was in fact correct https://github.com/bitcoindevkit/bdk/pull/1225#discussion_r1402485388. :)

danielabrozzoni: Mhh, let's start from the tests. In the first case, what's the min stop gap that must find the tx? Is it 3, or is it 4? 

I really want to have a clear definition in my mind of gap limit before reasoning about this :)

darosior:
Ok, just pushed a version which fixes the off-by-one when the last active index is `None`. Thanks both for chiming in!

darosior:
Ok the discussion about what the gap limit should be is now happening in two threads. Let's centralize it here. I'm going to close both threads.

First of all let me point out how this PR does not change existing behaviour for the case where there is a last active index. We might want to change that, but maybe let's keep this for another PR: one bug at a time heh?

That said, whether it happens here or in a follow-up, Daniela [raises a good point](https://github.com/bitcoindevkit/bdk/pull/1225#discussion_r1404241584). If we define "gap limit" as "the number of inactive derivation indexes before we stop" then the current implementation is wrong: as demonstrated by the newly introduced test with a gap limit of 3 we scan up to the 4th derivation index.

danielabrozzoni (approved):
> First of all let me point out how this PR does not change existing behaviour for the case where there is a last active index. We might want to change that, but maybe let's keep this for another PR: one bug at a time heh?

Fair enough! I've opened #1227 for discussion about this - even if the current definition of stop_gap is correct, it should be clearly documented :)

ACK 43aed386bc33f8dfca53ea2ec166dea1ce3d42d3

darosior:
> I've opened https://github.com/bitcoindevkit/bdk/issues/1127 for discussion

https://github.com/bitcoindevkit/bdk/issues/1227

evanlinjin (approved):
ACK 43aed386bc33f8dfca53ea2ec166dea1ce3d42d3

PR #1203, opened 2023-11-10
Include the descriptor in `keychain::Changeset`

danielabrozzoni:
Fixes bitcoindevkit/bdk#1101

- Moves keychain::ChangeSet inside `keychain/txout_index.rs` as now the `ChangeSet` depends on miniscript
- Slightly cleans up tests by introducing some constant descriptors
- The KeychainTxOutIndex's internal SpkIterator now uses DescriptorId
instead of K. The DescriptorId -> K translation is made at the
KeychainTxOutIndex level.
- The keychain::Changeset is now a struct, which includes a map for last
revealed indexes, and one for newly added keychains and their
descriptor.

### Changelog notice

API changes in bdk:
- Wallet::keychains returns a `impl Iterator` instead of `BTreeMap`
- Wallet::load doesn't take descriptors anymore, since they're stored in the db
- Wallet::new_or_load checks if the loaded descriptor from db is the same as the provided one
  
API changes in bdk_chain:
- `ChangeSet` is now a struct, which includes a map for last revealed
      indexes, and one for keychains and descriptors.
- `KeychainTxOutIndex::inner` returns a `SpkIterator<(DescriptorId, u32)>`
- `KeychainTxOutIndex::outpoints` returns a `BTreeSet` instead of `&BTreeSet`
- `KeychainTxOutIndex::keychains` returns a `impl Iterator` instead of
      `&BTreeMap`
- `KeychainTxOutIndex::txouts` doesn't return a ExactSizeIterator anymore
- `KeychainTxOutIndex::last_revealed_indices` returns a `BTreeMap`
      instead of `&BTreeMap`
- `KeychainTxOutIndex::add_keychain` has been renamed to `KeychainTxOutIndex::insert_descriptor`, and now it returns a ChangeSet
- `KeychainTxOutIndex::reveal_next_spk` returns Option
- `KeychainTxOutIndex::next_unused_spk` returns Option
- `KeychainTxOutIndex::unbounded_spk_iter` returns Option
- `KeychainTxOutIndex::next_index` returns Option
- `KeychainTxOutIndex::reveal_to_target` returns Option
- `KeychainTxOutIndex::revealed_keychain_spks` returns Option
- `KeychainTxOutIndex::unused_keychain_spks` returns Option
- `KeychainTxOutIndex::last_revealed_index` returns Option
- `KeychainTxOutIndex::keychain_outpoints` returns Option
- `KeychainTxOutIndex::keychain_outpoints_in_range` returns Option
[...]

PR #1229, opened 2023-11-27
Use a universal lookahead value for `KeychainTxOutIndex` and have a reasonable default

darosior:
### Description

The `bdk::Wallet` is currently created without setting any lookahead value for the keychain. This implicitly makes it a lookahead of 0. As this is a high-level interface we should avoid footguns and aim for a reasonable default.

To fix this, we have also decided to change `KeychainTxOutIndex` to have a default lookahead. Additionally, we have simplified the `KeychainTxOutIndex` API to have a single `lookahead` that is ONLY set at construction `KeychainTxOutIndex::new(lookahead: u32) -> Self`. This avoids the footguns of having methods which allows the caller to decrease the `lookahead` (which will panic).

### Notes to the reviewers

~A way to set this value externally is introduced in #1172. This PR only aims to use a saner default than 0. `1_000` is the value used by the Bitcoin Core wallet, and that seems reasonable to me.~

Edit: we should NOT allow setting the `lookahead` value after-the-fact. Instead, the `lookahead` should be provided to the wallet's constructor.

@evanlinjin: I don't think additional tests are necessary as no additional features are added, and the surface area of the API is decreased. The original tests already thoroughly test the `lookahead` concept.

#### All Submissions:

*(This section was updated as this PR changed from being a simple setting to introducing a new method.)*

~* [ ] I've added tests~
* [x] I've added docs


darosior:
Not sure how i would implement a test which:
1. Creates a fresh wallet
2. Sends a transaction on the second address from the keychain
3. Asserts the wallet notices it (this would fail on master)

LLFourn:
See the tests in `/tests` for where the wallet tests are. There's a bunch of them that test little things like this.

Historical note that I actually pushed against having a lookahead for a little while. I felt like it encouraged a design where people had multiple devices giving out addresses for the same keychain which I feel is bad practice. The wallet giving out addresses for the xpub knows the last one that has been produced so it's technically not possible for it exist on the blockchain and application developers should be using this invariant.

My complaints were put aside when @evanlinjin pointed out the impossibility of doing block-by-block updating without a lookahead. The default was probably set to 0 to placate me because I still wanted the developer to consciously engage this feature for that scenario. ~But I think this is fine for `Wallet` -- I'd leave 0 as the default for `KeychainTxOutIndex`~ changed my mind just have a nice default in both places.

crates/bdk/src/wallet/mod.rs
```diff
@@ -456,6 +459,7 @@ impl<D> Wallet<D> {
         let secp = Secp256k1::new();
         let (chain, chain_changeset) = LocalChain::from_genesis_hash(genesis_hash);
         let mut index = KeychainTxOutIndex::<KeychainKind>::default();
+        index.set_lookahead_for_all(DEFAULT_LOOKAHEAD);
```
notmandatory: I created a small test and found you need to move where you set the lookahead to below the `create_signers()` function. But then some other tests began to fail that we'll need to dig into a bit. The changes and test are here: https://github.com/bitcoindevkit/bdk/pull/1230/commits/9d3cb2b3e2907a9fefa5e5389d524daaa752aa99. Feel free to copy/paste them into your PR or cherry pick and squash the commit into yours. 

I confirmed that without your DEFAULT_LOOKAHEAD my new test fails.

evanlinjin: Thanks for pointing this out.

The `create_signers` method calls `KeychainTxOutIndex::add_keychain`. `KeychainTxOutIndex::set_lookahead_for_all` method sets the lookaheads for already-added keychains. So if you call `add_keychain` after `set_lookahead_for_all`, the keychain added will always start with a lookahead of 0.

This can be a confusing API. I propose we add an additional parameter to the `add_keychain` method:

```rust
pub fn add_keychain(&mut self, keychain: K, descriptor: Descriptor<DescriptorPublicKey>, lookahead: u32) {
    todo!()
}
```

notmandatory: I like this solution, it also makes it more obvious that if you're adding a keychain you should think about the look ahead.

darosior: Note: it seems the `txout_index` logic is assuming the lookahead is never reduced (for instance it's asserted here https://github.com/bitcoindevkit/bdk/blob/55b680c194b218061254ea898d892c9a5e4cdd3b/crates/chain/src/keychain/txout_index.rs#L393-L396 and here https://github.com/bitcoindevkit/bdk/blob/55b680c194b218061254ea898d892c9a5e4cdd3b/crates/chain/src/keychain/txout_index.rs#L414).

As i introduce a way to set the lookindex in `add_keychain` as suggested above, i'm going to remove `set_lookahead` to prevent this footgun. The alternative obviously would be to drop the asserts and make sure the code doesn't actually assumes this but:
1. This sounds like a lot more work
2. I'm not familiar enough with this code to make this assessment

Does that make sense?

darosior: Alright so i've implemented @evanlinjin's suggestion and simply modified the first assertion from above to be a `>=` instead of an equality check. I've left out the commit which drops the methods to set the lookahead externally (https://github.com/darosior/bdk/commit/4166ea5e37c859f2f07e33b69c08d183c65ded4c) but i'm still interested in your thoughts about that.

benthecarman:
Does this mean it is watching the next 1000 addresses? That seems pretty expensive for syncing

crates/chain/src/keychain/txout_index.rs
```diff
@@ -142,15 +142,22 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// # Panics
     ///
     /// This will panic if a different `descriptor` is introduced to the same `keychain`.
-    pub fn add_keychain(&mut self, keychain: K, descriptor: Descriptor<DescriptorPublicKey>) {
+    pub fn add_keychain(
+        &mut self,
+        keychain: K,
+        descriptor: Descriptor<DescriptorPublicKey>,
+        lookahead: u32,
```
LLFourn: My pref would be to have `add_keychain_with_lookahead` and have `add_keychain` which just sets a default of `1_000`. Apps looking to save memory can use `add_keychain_with_lookahead` and provide 0 as the argument but I'm now more inclined to have a large default here as we do in `Wallet`.

Having two methods provides a good place to explain the concept of lookahead and means we don't have some unexplained integer argument at the end of `add_keychain`.

If you don't feel comfortable writing the docs and explaining the design rationale I'm sure we could chip in here.

LLFourn:
> Does this mean it is watching the next 1000 addresses? That seems pretty expensive for syncing

No what you actually go out and fetch from the chain is under your control and this lookahead doesn't effect it. This is about what the txout index will find when you feed it transaction. It's only relevant when you are doing block style syncing -- it shouldn't effect at all electrum/esplora.

crates/chain/src/keychain/txout_index.rs
evanlinjin: I'll be happy with that. Need good docs though!

darosior: It's already documented, but i can point to where it is, sure.

darosior: Done.

> If you don't feel comfortable writing the docs and explaining the design rationale I'm sure we could chip in here.

Feel free to, i've simply pointed to the existing doc about `lookahead`.

crates/bdk/src/wallet/mod.rs
evanlinjin: @darosior That's a good point. I think we should still be able to increase the lookahead. However, decreasing the lookahead should not make any sense. Also, maybe in the future we should persist the lookahead as well (since we are going to be persisting the descriptor - #1203).

I think the best course of action for now, would be to have `set_lookahead` and `set_lookahead_for_all` fail if the caller tries to set a lower lookahead.

```rust
/// ...
/// Lookahead cannot decrease.
/// On failure, we return the current lookahead value.
pub fn set_lookahead(&mut self, keychain: &K, lookahead: u32) -> Result<(), u32> { todo!() }

/// ...
/// Lookaheads cannot decrease.
/// On failure, return the highest current lookahead value.
pub fn set_lookahead_for_all(&mut self, lookahead: u32) -> Result<(), (K, u32)> { todo!() }
```

What do you guys think? Is there merit in decreasing the lookahead in the future?

LLFourn: It don't think it should be persisted. It's a runtime policy thing and is more dependent on your hardware than it is on your keychain. I'd just document that these methods can only increase it.

notmandatory:

evanlinjin:
@darosior If it is alright with you, could I take over this PR? I think this is important for syncing from block-by-block chain sources.

crates/bdk/src/wallet/mod.rs
evanlinjin: I will be okay with just asserting that the lookahead cannot decrease.

evanlinjin:
After discussion with @LLFourn, we have come to agreement to use a universal `lookahead` value which is specified when creating the `KeychainTxOutIndex`.

```rust
pub fn new(lookahead: u32) -> Self { todo!() }
```

I will move this PR forward in this manner.

notmandatory:
I fixed the CI issues with #1247, it just needs to be ACKd and merged and then this PR rebased.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -142,57 +156,27 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// # Panics
     ///
     /// This will panic if a different `descriptor` is introduced to the same `keychain`.
+    ///
+    /// [`add_keychain_with_lookahead`]: Self::add_keychain_with_lookahead
```
notmandatory: Nit, this line and the one above aren't needed and should be removed.

notmandatory:
The main changes looks good, but when I tried to do some manual `regtest` testing with the `example_bitcoind_rpc_polling` example code I ran into an issue where the genesis block isn't set. I was able to get past that by adding the below to the example `main.rs`, but pretty sure it's the right way to fix it.
```rust
if init_changeset.0.is_empty() {
        let genesis_hash = genesis_block(Regtest).block_hash();
        dbg!(genesis_hash);
        let  (_chain, changeset) = LocalChain::from_genesis_hash(genesis_hash);
        init_changeset.0 = changeset;
        db.lock().unwrap().stage(init_changeset.clone());
        db.lock().unwrap().commit()?;
    }
 ```

**Review by notmandatory — APPROVED** [review]
tACK bc796f412acdc3d2cd96f7aec0f24fa47c1fe889

I manually tested using `example_bitcoind_rpc_polling` cli with the steps in gist: https://gist.github.com/notmandatory/6fa85dd942089833629bdb40d2b54858

File: crates/chain/src/keychain/txout_index.rs [hunk 4]
```diff
@@ -118,6 +114,24 @@ impl<K: Clone + Ord + Debug> Indexer for KeychainTxOutIndex<K> {
     }
 }
 
+impl<K> KeychainTxOutIndex<K> {
+    /// Construct a [`KeychainTxOutIndex`] with the given `lookahead`.
+    ///
+    /// The lookahead is the number of scripts to cache ahead of the last revealed script index.
```
LLFourn:
```suggestion
    /// The lookahead is the number of script pubkeys to cache ahead of the last revealed script index.
```

LLFourn:
I pushed a commit to improve the lookahead doc slightly.

self-ACK: c9467dcbb27bd497006846dcfcefb0c2cf8823c4

LLFourn (approved):
ACK 1def76f1f17fc876534c798883099c58de57c0fd

PR #1070, opened 2023-08-09
Rename methods of esplora ext

vladimirfomene:
### Description

This PR fixes #1058. Built on top of #1040 

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [ ] I've added tests for the new feature
* [ ] I've added docs for the new feature

PR #1235, opened 2023-12-07
Refactor/rename electrum_ext and esplora_ext to have sync and full_scan functions

### LINKED ISSUE #1112: `scan_without_keychain` and `scan`'s names should be inverted

notmandatory:
### Description

fixes #1112

Simple function renaming plus updated docs:

1. electrum_ext: rename functions `scan_without_keychain` to `sync` and `scan` to `full_scan` 
2. esplora_ext: rename functions `scan_txs` to `sync` and `scan_txs_with_keychains` to `full_scan`

### Notes to the reviewers

The esplora_ext changes were partially fixed in #1070 but I renamed again so the functions match names ~~suggested in #1112~~ agreed on in discord poll, `sync` and `full_scan`.

### Changelog notice

Changed

- electrum_ext: rename functions scan_without_keychain to sync and scan to full_scan 
- esplora_ext: rename functions scan_txs to sync and scan_txs_with_keychains to full_scan

#### All Submissions:

### DIFF
--- a/crates/chain/src/lib.rs
+++ b/crates/chain/src/lib.rs
@@ -1,4 +1,4 @@
-//! This crate is a collection of core structures for [Bitcoin Dev Kit] (alpha release).
+//! This crate is a collection of core structures for [Bitcoin Dev Kit].
 //!
 //! The goal of this crate is to give wallets the mechanisms needed to:
 //!
--- a/crates/electrum/src/electrum_ext.rs
+++ b/crates/electrum/src/electrum_ext.rs
@@ -134,64 +134,54 @@ pub struct ElectrumUpdate {
 
 /// Trait to extend [`Client`] functionality.
 pub trait ElectrumExt {
-    /// Scan the blockchain (via electrum) for the data specified and returns updates for
-    /// [`bdk_chain`] data structures.
+    /// Full scan the keychain scripts specified with the blockchain (via an Electrum client) and
+    /// returns updates for [`bdk_chain`] data structures.
     ///
     /// - `prev_tip`: the most recent blockchain tip present locally
     /// - `keychain_spks`: keychains that we want to scan transactions for
-    /// - `txids`: transactions for which we want updated [`Anchor`]s
-    /// - `outpoints`: transactions associated with these outpoints (residing, spending) that we
-    ///     want to included in the update
     ///
-    /// The scan for each keychain stops after a gap of `stop_gap` script pubkeys with no associated
+    /// The full scan for each keychain stops after a gap of `stop_gap` script pubkeys with no associated
     /// transactions. `batch_size` specifies the max number of script pubkeys to request for in a
     /// single batch request.
-    fn scan<K: Ord + Clone>(
+    fn full_scan<K: Ord + Clone>(
         &self,
         prev_tip: CheckPoint,
         keychain_spks: BTreeMap<K, impl IntoIterator<Item = (u32, ScriptBuf)>>,
-        txids: impl IntoIterator<Item = Txid>,
-        outpoints: impl IntoIterator<Item = OutPoint>,
         stop_gap: usize,
         batch_size: usize,
     ) -> Result<(ElectrumUpdate, BTreeMap<K, u32>), Error>;
 
-    /// Convenience method to call [`scan`] without requiring a keychain.
+    /// Sync a set of scripts with the blockchain (via an Electrum client) for the data specified
+    /// and returns updates for [`bdk_chain`] data structures.
     ///
-    /// [`scan`]: ElectrumExt::scan
-    fn scan_without_keychain(
+    /// - `prev_tip`: the most recent blockchain tip present locally
+    /// - `misc_spks`: an iterator of scripts we want to sync transactions for
+    /// - `txids`: transactions for which we want updated [`Anchor`]s
+    /// - `outpoints`: transactions associated with these outpoints (residing, spending) that we
+    ///     want to include in the update
+    ///
+    /// `batch_size` specifies the max number of script pubkeys to request for in a single batch
+    /// request.
+    ///
+    /// If the scripts to sync are unknown, such as when restoring or importing a keychain that
+    /// may include scripts that have been used, use [`full_scan`] with the keychain.
+    ///
+    /// [`full_scan`]: ElectrumExt::full_scan
+    fn sync(
         &self,
         prev_tip: CheckPoint,
         misc_spks: impl IntoIterator<Item = ScriptBuf>,
         txids: impl IntoIterator<Item = Txid>,
         outpoints: impl IntoIterator<Item = OutPoint>,
         batch_size: usize,
-    ) -> Result<ElectrumUpdate, Error> {
-        let spk_iter = misc_spks
-            .into_iter()
-            .enumerate()
-            .map(|(i, spk)| (i as u32, spk));
-
-        let (electrum_update, _) = self.scan(
-            prev_tip,
-            [((), spk_iter)].into(),
-            txids,
-            outpoints,
-            usize::MAX,
-            batch_size,
-        )?;
-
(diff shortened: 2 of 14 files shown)


LLFourn:
1. I think that people preferred some other names. This didn't make it into the issue :(. I remmber there was a discord poll cc @danielabrozzoni 
2. It would be really good if we could do as I mentioned in the issue and remove the arguments to `scan_with_keychain` (or whatever it ends up being called) except for the stop gap and the keychain and spks (and chain specific options like parallel requests). `scan_with_keychain` should be a minimal and sepcialized function.
3. The documentation needs to be updated to reflect this. One bit of it even calls `scan_with_keychain` "the star of the show" when we are trying to explain that it is NOT that. It is the opposite of that. It is the ugly inflexible understudy who is only called up when you need to recover from seed!

LLFourn:
https://discord.com/channels/753336465005608961/753367451319926827/1171551358353158164 here it is!

notmandatory:
Oh ya, I was thinking that poll was for `Wallet` but makes sense we also need to change the names and docs here, will do it. The winners of the poll that I'll use are `full_scan` and `sync`.

notmandatory:
Functions renamed plus I updated the docs.

crates/chain/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-//! This crate is a collection of core structures for [Bitcoin Dev Kit] (alpha release).
+//! This crate is a collection of core structures for [Bitcoin Dev Kit].
```
notmandatory, line 1: Alpha version status is already in `Cargo.toml`.

crates/electrum/src/electrum_ext.rs
```diff
@@ -134,19 +134,19 @@ pub struct ElectrumUpdate {
...
     /// - `txids`: transactions for which we want updated [`Anchor`]s
     /// - `outpoints`: transactions associated with these outpoints (residing, spending) that we
     ///     want to included in the update
     ///
-    /// The scan for each keychain stops after a gap of `stop_gap` script pubkeys with no associated
+    /// The full scan for each keychain stops after a gap of `stop_gap` script pubkeys with no associated
     /// transactions. `batch_size` specifies the max number of script pubkeys to request for in a
     /// single batch request.
-    fn scan<K: Ord + Clone>(
+    fn full_scan<K: Ord + Clone>(
```
LLFourn, line 146: please remove `txids` and `outpoints` from this.

notmandatory: Fixed, and for esplora. I also removed copy/paste mistake in the sync() docs that mentioned stop gap and made some params refs.

crates/electrum/src/electrum_ext.rs
```diff
@@ -134,64 +134,54 @@ pub struct ElectrumUpdate {
...
 
-    /// Convenience method to call [`scan`] without requiring a keychain.
+    /// Sync a set of scripts with the blockchain (via an Electrum client) for the data specified
+    /// and returns updates for [`bdk_chain`] data structures.
+    ///
+    /// - `prev_tip`: the most recent blockchain tip present locally
+    /// - `misc_spks`: an iterator of scripts we want to sync transactions for
+    /// - `txids`: transactions for which we want updated [`Anchor`]s
+    /// - `outpoints`: transactions associated with these outpoints (residing, spending) that we
+    ///     want to included in the update
```
ValuedMammal:
```suggestion
    ///     want to include in the update
```

crates/esplora/src/async_ext.rs
```diff
@@ -36,58 +36,45 @@ pub trait EsploraAsyncExt {
...
         >,
-        txids: impl IntoIterator<IntoIter = impl Iterator<Item = Txid> + Send> + Send,
-        outpoints: impl IntoIterator<IntoIter = impl Iterator<Item = OutPoint> + Send> + Send,
         stop_gap: usize,
         parallel_requests: usize,
     ) -> Result<(TxGraph<ConfirmationTimeHeightAnchor>, BTreeMap<K, u32>), Error>;
 
-    /// Convenience method to call [`scan_txs_with_keychains`] without requiring a keychain.
+    /// Sync a set of scripts with with the blockchain (via an Esplora client) for the data
+    /// specified and return a [`TxGraph`] and a map of last active indices.
```
ValuedMammal:
```suggestion
    /// Sync a set of scripts with the blockchain (via an Esplora client) for the data
    /// specified and return a [`TxGraph`].
```

The function doesn't appear to return a 'map of last active indices.' The same is true for the blocking client method.

ValuedMammal (approved):
The code looks good. I ran the esplora tests - all ok.

crates/esplora/src/async_ext.rs
notmandatory: thanks good catches! I fixed for blocking_ext and electrum too.

crates/electrum/src/electrum_ext.rs
```diff
@@ -134,64 +134,54 @@ pub struct ElectrumUpdate {
...
     ///
-    /// The scan for each keychain stops after a gap of `stop_gap` script pubkeys with no associated
+    /// The full scan for each keychain stops after a gap of `stop_gap` script pubkeys with no associated
     /// transactions. `batch_size` specifies the max number of script pubkeys to request for in a
     /// single batch request.
-    fn scan<K: Ord + Clone>(
+    fn full_scan<K: Ord + Clone>(
         &self,
-        prev_tip: CheckPoint,
+        prev_tip: &CheckPoint,
```
LLFourn: I think we shouldn't use references here. Although we can it seems like the wrong hint to give to the user. The design of all the other parameters is that this function owns all of them so there can be no lifetime issues when spawning off a new thread to run this. The value returned by `LocalChain::tip` should be passed trait into this -- borrowing it beforehand doesn't help anyone and just leads to a compiler error before you add a reference and pass that it instead.

LLFourn (approved):
ACK d36db4d8df8173d92b9cdeb4fa9877ca77f56d20

Much nicer thanks. 

side note: the docs here are stale: [d36db4d8df8173d92b9cdeb4fa9877ca77f56d20](https://github.com/bitcoindevkit/bdk/pull/1235/files#diff-edcdb067974af68901999ca2335e8aabe0b6e0768845bb637a87f5dbe5d34124L22). There is no argument `prev_tip` to `update_local_chain`.

notmandatory:
> side note: the docs here are stale: [d36db4d8df8173d92b9cdeb4fa9877ca77f56d20](https://github.com/bitcoindevkit/bdk/pull/1235/files#diff-edcdb067974af68901999ca2335e8aabe0b6e0768845bb637a87f5dbe5d34124L22). There is no argument `prev_tip` to `update_local_chain`.

fixed!

crates/electrum/src/electrum_ext.rs
notmandatory: Fixed.

PR #1247, opened 2023-12-20
ci: pin home dependency to 0.5.5 and check_clippy to rust stable version

notmandatory:
### Description

Fixed 1.63 MSRV error by pinning `home` dependency to `0.5.5`, and `clippy` error by changing `check_clippy` job to using rust `stable` version. 

### Notes to the reviewers

It's OK to use rust `stable` version for clippy because we already have a `clippy.toml` file to tell it which version of the clippy rules to check against.

### Changelog notice

None

#### All Submissions:


evanlinjin (approved):
ACK e7fbc8bcf36de110e88095dba8be0c62e2c8e0ef

PR #1250, opened 2023-12-28
Introduce `bdk::Wallet` builder API

evanlinjin:
### Description

This changes the wallet-construction API to use the builder pattern. This is a simpler API and allows us to avoid multiple constructor methods.

I.e. previously, we had `Wallet::new` and `Wallet::new_with_genesis_hash`. If in the future, we added more parameters, it will turn into an absolute mess.

### Notes to the reviewers

This PR is based on #1229. Consider reviewing/merging that first!

I think it makes sense to have the `lookahead` configurable in `bdk::Wallet` and persisted. However, @LLFourn is against this idea of persisting the `lookahead`. This can be of discussion for a future change.

### Changelog notice

* Changed wallet-construction API to use the builder pattern.
* Changed `NewError` to `InitError` and `NewOrLoadError` to `InitOrLoadError`.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/bdk/src/wallet/builder.rs
+++ b/crates/bdk/src/wallet/builder.rs
@@ -1,0 +1,98 @@
+use bdk_chain::PersistBackend;
+use bitcoin::constants::genesis_block;
+use bitcoin::{BlockHash, Network};
+
+use crate::descriptor::{DescriptorError, IntoWalletDescriptor};
+use crate::wallet::{ChangeSet, Wallet};
+
+use super::{InitError, InitOrLoadError};
+
+/// Helper structure to initialize a fresh [`Wallet`] instance.
+pub struct WalletBuilder<E> {
+    pub(super) descriptor: E,
+    pub(super) change_descriptor: Option<E>,
+    pub(super) network: Network,
+    pub(super) custom_genesis_hash: Option<BlockHash>,
+}
+
+impl<E> WalletBuilder<E> {
+    /// Start building a fresh [`Wallet`] instance.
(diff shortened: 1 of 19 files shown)


notmandatory:
Concept ACK, this pattern should also be easier to extend to support future features like custom coin selection. But I'd like to push it to the alpha.4 release, alpha.3 is already long overdue.

evanlinjin:
@notmandatory if we wish to make the `lookahead` configurable in `bdk::Wallet` for alpha.3, I think it will be a good idea to include this in the release. Otherwise, I am okay with pushing this back.

LLFourn:
Concept ACK

crates/bdk/src/wallet/builder.rs
```diff
@@ -0,0 +1,98 @@
...
+use bitcoin::constants::genesis_block;
+use bitcoin::{BlockHash, Network};
+
+use crate::descriptor::{DescriptorError, IntoWalletDescriptor};
+use crate::wallet::{ChangeSet, Wallet};
+
+use super::{InitError, InitOrLoadError};
+
+/// Helper structure to initialize a fresh [`Wallet`] instance.
+pub struct WalletBuilder<E> {
```
ValuedMammal, line 11: How do you feel about just calling this `Builder`, i.e. `bdk::wallet::Builder` ?

evanlinjin:
It makes sense to complete this AFTER #1203 is merged (as that PR also touches wallet constructors).

notmandatory:

evanlinjin:
Builder pattern already introduced.

PR #1253, opened 2024-01-02
Remove deprecated checksum routines

theStack:
### Description

This PR removes the routines `get_checksum` and `get_checksum_bytes` (in the bdk crate, descriptor/checksum.rs), which have been deprecated in 0.24.0. Consequently, the routine `calc_checksum_bytes_internal` is consolidated into `calc_checksum_bytes`, as the boolean parameter `exclude_hash` is not needed anymore. See also the two TODOs in the touched file.

### Notes to the reviewers

In the second commit, the signature of the function `calc_checksum_bytes` slightly changes, as the `desc` parameter now is declared as `mut`, in order to change the local variable within the function. My rust experience is rather limited, so I'm not sure if this is a problem for users. IIUC, this is comparable to changing a pass-by-value parameter in C++ from `const std::string desc` to `std::string desc`, which is relevant only for the function implementation, but doesn't change the interface.

### Changelog notice

- Remove deprecated `get_checksum` and `get_checksum_bytes` routines

#### All Submissions:

### DIFF
--- a/crates/bdk/src/descriptor/checksum.rs
+++ b/crates/bdk/src/descriptor/checksum.rs
@@ -107,28 +107,7 @@ pub fn calc_checksum(desc: &str) -> Result<String, DescriptorError> {
         .map(|b| unsafe { String::from_utf8_unchecked(b.to_vec()) })
 }
 
-// TODO in release 0.25.0, remove get_checksum_bytes and get_checksum
 // TODO in release 0.25.0, consolidate calc_checksum_bytes_internal into calc_checksum_bytes
-
-/// Compute the checksum bytes of a descriptor
-#[deprecated(
-    since = "0.24.0",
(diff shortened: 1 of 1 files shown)


danielabrozzoni (approved):
ACK ed91a4bdb49276057eb7458e1a94653553be24e7

Looks good to me, thanks for taking the time to clean up the code! :)

evanlinjin:
@theStack thank you for the contribution. However our policy is that all commits need to be signed otherwise we cannot merge! Please sign your commits!

theStack:
> @theStack thank you for the contribution. However our policy is that all commits need to be signed otherwise we cannot merge! Please sign your commits!

The commits are signed, but apparently I forgot to upload my GPG public key to GitHub. Did that now, so the commits are shown as "Verified". Is there anything more to do?

danielabrozzoni:
It should be fine now, thanks!

PR #1256, opened 2024-01-06
cherry-pick feat(wallet)!: add `NonEmptyDatabase` variant to `NewError`

notmandatory:
### Description

`NewError` is the error type when constructing a wallet with `Wallet::new`. We want this to return an error when the database already contains data (in which case, the caller should use `load` or `new_or_load`).

### Notes to the reviewers

This is cherry-picked from #1172 so that we can add it to the alpha.3 release. 

### Changelog notice

Change
- Return `NonEmptyDatabase` error when constructing a wallet with `Wallet::new` if the file already contains data (in which case, the caller should use `load` or `new_or_load`).

#### All Submissions:

#### New Features:


evanlinjin (approved):
ACK a1d34afa2455d93d23a36fd78d94aa79c7e493d1

PR #875, opened 2023-02-28
Bump bip39 crate to v2.0.0

tnull:
### Description

The updated version of `rust-bip39` was just released. It also [bumped the pinned version](https://github.com/rust-bitcoin/rust-bip39/pull/41) of `unicode-normalization`, which previously had lead to some incompatibilities.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #1259, opened 2024-01-07
Bump `bip39` dependency to v2.0

tnull:
We previously bumped the `bip39` version to 2.0 [in the 0.2X release branch](https://github.com/bitcoindevkit/bdk/pull/875). Back then, bumping it in `master` was [assumed unnecessary](https://github.com/bitcoindevkit/bdk/pull/875#issuecomment-1483990088). It seems this was erroneous, as the current `1.0.1` dependency is present since https://github.com/bitcoindevkit/bdk/pull/793, which was merged before the bump in `release/0.27`.

Here, we therefore bump the crate version in `master` after all.

### DIFF
--- a/crates/bdk/Cargo.toml
+++ b/crates/bdk/Cargo.toml
@@ -22,7 +22,7 @@ bdk_chain = { path = "../chain", version = "0.6.0", features = ["miniscript", "s
 
 # Optional dependencies
 hwi = { version = "0.7.0", optional = true, features = [ "miniscript"] }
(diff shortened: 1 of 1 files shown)


tnull:
@notmandatory Any chance this could still make it in to 1.0-alpha.3?

notmandatory:
Thanks for catching this. I'll make sure it's in the alpha.3 release.

notmandatory (approved):
ACK 8694624bd5f63b738448c250d7b1042f219636da

PR #1261, opened 2024-01-08
Refactor `reveal_to_target` and `next_store_index`

danielabrozzoni:
### Description

As a part of #1203, I found myself having to modify `reveal_to_target`, but had some troubles understanding exactly what the method was doing.

This PR refactors `reveal_to_target` to be, in my opinion, a bit clearer. We now exist prematurely if `next_reveal_index` < `target_index`; this way, we don't need to keep track of `next_reveal_index`, as it would always be equal to `Some(target_index)`.
As a part of trying to understand `reveal_to_target` I had to read through `next_store_index`, I slightly improved it for clarity reasons as well. 

#### All Submissions:

### DIFF
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -390,7 +390,11 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
             let _inserted = self
                 .inner
                 .insert_spk((keychain.clone(), new_index), new_spk);
-            debug_assert!(_inserted, "must not have existing spk",);
+            debug_assert!(_inserted, "must not have existing spk");
+            debug_assert!(
+                has_wildcard || new_index == 0,
+                "non-wildcard descriptors must not iterate past index 0"
+            );
         }
 
         let _old_index = self.last_revealed.insert(keychain.clone(), target_index);
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -364,51 +364,41 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         let has_wildcard = descriptor.has_wildcard();
 
         let target_index = if has_wildcard { target_index } else { 0 };
-        let next_reveal_index = self.last_revealed.get(keychain).map_or(0, |v| *v + 1);
+        let next_reveal_index = self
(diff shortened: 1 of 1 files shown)


crates/chain/src/keychain/txout_index.rs
```diff
@@ -206,9 +206,10 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     fn next_store_index(&self, keychain: &K) -> u32 {
         self.inner()
             .all_spks()
-            .range((keychain.clone(), u32::MIN)..(keychain.clone(), u32::MAX))
+            .keys()
+            .filter(|(k, _)| k == keychain)
```
evanlinjin: I advise against using `filter` here. `filter` will iterate over every single spk whereas `range` only iterates over spks of the given keychain.

Additionally, `btree::Range` has an optimized implementation of `Iterator::last` (https://doc.rust-lang.org/src/alloc/collections/btree/map.rs.html#1981) which we will be missing out on if we use `filter`.

evanlinjin (changes requested):
Thank you for this change. The code is clearer now. However, please revert the change to `next_store_index`.

evanlinjin:
@danielabrozzoni I added another `debug_assert` statement while I was at it: a28bbb262a6e039930ce933a12d11904af907c02

crates/chain/src/keychain/txout_index.rs
danielabrozzoni: Good catch! I added a comment explaining what is `range` doing and why we should use it

crates/chain/src/keychain/txout_index.rs
```diff
@@ -206,9 +206,11 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     fn next_store_index(&self, keychain: &K) -> u32 {
         self.inner()
             .all_spks()
+            // This range is filtering out the spks with a keychain different than
+            // `keychain`. We don't use filter here as range is more optimized.
```
evanlinjin, line 210: I would prefer *"Range through all spks of `keychain`."*. However, I feel like this comment is not needed as the codebase already shows this.

Definitely not a blocker though.

evanlinjin (approved):
ACK 761189ab2b679fbfee45b33435d66ed98a135e2a

PR #1267, opened 2024-01-11
Simplify `Esplora::update_local_chain` and add tests

### LINKED ISSUE #1199: bdk + esplora `update_local_chain` logic might be off
See the discussion in #1151: https://github.com/bitcoindevkit/bdk/issues/1151#issuecomment-1752555182 (and following comments)

evanlinjin:
Fixes #1199

### Description

After a second look at the `update_local_chain` implementations, it was clear that they were over complicated. This PR simplifies the `EsploraExt::update_local_chain` method(s) of the `bdk_esplora` crate and adds a whole bunch of tests.

### Notes to the reviewers

The description of #1199 is very brief, however @danielabrozzoni mentioned about potentially-problematic logic with `ASSUME_FINAL_DEPTH`. The logic was indeed convoluted and it did not need to be that way. This PR removes the need for `ASSUME_FINAL_DEPTH`. 

### Changelog notice

Fixed
- Simplified `EsploraExt::update_local_chain` logic.

#### All Submissions:

#### Bugfixes:

~* [ ] I've added tests to reproduce the issue which are now passing~ (there are now lots of tests though)

### DIFF
--- a/crates/esplora/src/async_ext.rs
+++ b/crates/esplora/src/async_ext.rs
@@ -85,20 +85,20 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
         local_tip: CheckPoint,
         request_heights: impl IntoIterator<IntoIter = impl Iterator<Item = u32> + Send> + Send,
     ) -> Result<local_chain::Update, Error> {
-        let new_tip_height = self.get_height().await?;
-
         // Atomically fetch latest blocks from Esplora. This way, we avoid creating an update with
         // an inconsistent set of blocks (assuming that a reorg depth cannot be greater than the
         // latest blocks fetched).
-        let mut fetched_blocks = {
-            let heights = (0..=new_tip_height).rev();
-            let hashes = self
-                .get_blocks(Some(new_tip_height))
-                .await?
-                .into_iter()
-                .map(|b| b.id);
-            heights.zip(hashes).collect::<BTreeMap<u32, BlockHash>>()
-        };
+        let mut fetched_blocks = self
+            .get_blocks(None)
+            .await?
+            .into_iter()
+            .map(|b| (b.time.height, b.id))
+            .collect::<BTreeMap<u32, BlockHash>>();
+        let new_tip_height = fetched_blocks
+            .keys()
+            .last()
+            .copied()
+            .expect("must have atleast one block");
 
         // fetch blocks of heights that the caller is interested in, reusing latest blocks that are
         // already fetched.
--- a/crates/esplora/src/async_ext.rs
+++ b/crates/esplora/src/async_ext.rs
@@ -2,14 +2,14 @@ use async_trait::async_trait;
 use bdk_chain::collections::btree_map;
 use bdk_chain::{
     bitcoin::{BlockHash, OutPoint, ScriptBuf, Txid},
-    collections::{BTreeMap, BTreeSet},
+    collections::BTreeMap,
     local_chain::{self, CheckPoint},
     BlockId, ConfirmationTimeHeightAnchor, TxGraph,
 };
 use esplora_client::{Error, TxStatus};
 use futures::{stream::FuturesOrdered, TryStreamExt};
 
-use crate::{anchor_from_status, ASSUME_FINAL_DEPTH};
+use crate::anchor_from_status;
 
 /// Trait to extend the functionality of [`esplora_client::AsyncClient`].
 ///
@@ -85,10 +85,11 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
         local_tip: CheckPoint,
         request_heights: impl IntoIterator<IntoIter = impl Iterator<Item = u32> + Send> + Send,
     ) -> Result<local_chain::Update, Error> {
-        let request_heights = request_heights.into_iter().collect::<BTreeSet<_>>();
         let new_tip_height = self.get_height().await?;
 
-        // atomically fetch blocks from esplora
+        // Atomically fetch latest blocks from Esplora. This way, we avoid creating an update with
+        // an inconsistent set of blocks (assuming that a reorg depth cannot be greater than the
+        // latest blocks fetched).
         let mut fetched_blocks = {
             let heights = (0..=new_tip_height).rev();
             let hashes = self
@@ -99,7 +100,8 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
             heights.zip(hashes).collect::<BTreeMap<u32, BlockHash>>()
         };
 
-        // fetch heights that the caller is interested in
+        // fetch blocks of heights that the caller is interested in, reusing latest blocks that are
+        // already fetched.
         for height in request_heights {
             // do not fetch blocks higher than remote tip
             if height > new_tip_height {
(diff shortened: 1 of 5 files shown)


evanlinjin:
@notmandatory this doesn't really change the API

crates/esplora/src/async_ext.rs
```diff
@@ -99,89 +100,48 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
...
-                })
-        };
+
+            // We have found point of agreement so the update will connect!
+            if fetched_hash == local_hash {
+                break;
+            }
+        }
+
+        // TODO: Use `CheckPoint::from_block_ids` once #1172 is merged
```
LLFourn: This is merged now.

LLFourn (commented):
Concept ACK. One thing looks like it needs to be updated.

crates/esplora/src/async_ext.rs
evanlinjin: https://github.com/bitcoindevkit/bdk/pull/1172

But it's not? 😅

LLFourn: Derp. Merge it then.

crates/esplora/src/async_ext.rs
```diff
@@ -85,10 +85,11 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
         local_tip: CheckPoint,
         request_heights: impl IntoIterator<IntoIter = impl Iterator<Item = u32> + Send> + Send,
     ) -> Result<local_chain::Update, Error> {
-        let request_heights = request_heights.into_iter().collect::<BTreeSet<_>>();
         let new_tip_height = self.get_height().await?;
 
-        // atomically fetch blocks from esplora
+        // Atomically fetch latest blocks from Esplora. This way, we avoid creating an update with
+        // an inconsistent set of blocks (assuming that a reorg depth cannot be greater than the
```
LLFourn: What is this assumption I don't get it. There is no assumption being made here that I can see.

crates/esplora/src/async_ext.rs
```diff
@@ -85,10 +85,11 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
...
+        // latest blocks fetched).
```
LLFourn: Not touched as part of this PR but below:

```rust

        let mut fetched_blocks = {
            let heights = (0..=new_tip_height).rev();
            let hashes = self
                .get_blocks(Some(new_tip_height))
                .await?
                .into_iter()
                .map(|b| b.id);
            heights.zip(hashes).collect::<BTreeMap<u32, BlockHash>>()
        };
```
Why are we zipping to get the heights of the blocks. This looks really error prone. Surely esplora will tell us the heights of the blocks we fetched.

LLFourn: But furthermore, why do we even do `get_height` at the start here. Just do `get_blocks(None)` which will get from the tip.

LLFourn (commented):
I found some more things that could be simplified as part of this.

crates/esplora/src/async_ext.rs
evanlinjin: If the reorg depth is greater than the latest block count that we fetched atomically, and we had to get another block lower than the atomic-fetch count... This additional block may not connect with the atomically-fetched blocks.

crates/esplora/src/async_ext.rs
evanlinjin: I didn't realize there was a height available because it's nested in the `time` field.

LLFourn:
self-ACK: 216648bcfd2c622890bf7f50fb2eaa4396c42ed1

Pushed a comment and docs better explaining what @evanlinjin  was trying to explain to me here: https://github.com/bitcoindevkit/bdk/pull/1267#discussion_r1469287813

evanlinjin (commented):
ACK 216648bcfd2c622890bf7f50fb2eaa4396c42ed1

Thank you @LLFourn for the improved docs.

crates/esplora/src/async_ext.rs
```diff
@@ -99,89 +100,41 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
             heights.zip(hashes).collect::<BTreeMap<u32, BlockHash>>()
         };
 
-        // fetch heights that the caller is interested in
+        // fetch blocks of heights that the caller is interested in, reusing latest blocks that are
```
danielabrozzoni: nit: I would rephrase it as

// fetch blocks of heights that the caller is interested in, skipping the fetching part for blocks that are already in `fetched_blocks`

crates/esplora/src/async_ext.rs
```diff
@@ -99,89 +100,41 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
...
-                    let first_new_height = *fetched_blocks
-                        .keys()
-                        .next()
-                        .expect("must have at least one new block");
-                    if first_new_height >= local_block.height {
-                        break;
-                    }
-                }
+        // Ensure `fetched_blocks` can create an update that connects with the original chain.
+        for (height, local_hash) in local_tip.iter().map(|cp| (cp.height(), cp.hash())) {
```
danielabrozzoni, line 127: nit: I would call it `local_height`

crates/esplora/src/async_ext.rs
```diff
@@ -99,89 +100,41 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
...
-
-                    let first_new_height = *fetched_blocks
-                        .keys()
-                        .next()
-                        .expect("must have at least one new block");
-                    if first_new_height >= local_block.height {
-                        break;
-                    }
-                }
+        // Ensure `fetched_blocks` can create an update that connects with the original chain.
```
danielabrozzoni: nit: I would explicitly state here that the way you're ensuring that the update connects is by looking for a agreement point

crates/esplora/src/async_ext.rs
```diff
@@ -99,89 +100,41 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
...
-                // building the chain will fail
-                .split_off(&(first_cp.height() + 1))
-                .into_iter()
-                .map(|(height, hash)| BlockId { height, hash })
-                .fold(first_cp, |prev_cp, block| {
-                    prev_cp.push(block).expect("must extend checkpoint")
-                })
-        };
+
+            // We have found point of agreement so the update will connect!
```
danielabrozzoni, line 139: What if no point of agreement is found? Is this possible?

crates/esplora/tests/blocking_ext.rs
```diff
@@ -202,3 +232,161 @@ pub fn test_update_tx_graph_gap_limit() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn update_local_chain() -> anyhow::Result<()> {
+    const TIP_HEIGHT: u32 = 50;
+
+    let env = TestEnv::new()?;
+    let b = {
```
danielabrozzoni: I don't really like one character variable names, can we change this in `blocks`, and `bdc` in `bitcoind_client`?

crates/esplora/tests/blocking_ext.rs
```diff
@@ -202,3 +232,161 @@ pub fn test_update_tx_graph_gap_limit() -> anyhow::Result<()> {
...
+            .chain((2..).zip(env.mine_blocks((TIP_HEIGHT - 1) as usize, None)?))
+            .collect::<BTreeMap<_, _>>()
+    };
+    // so new blocks can be seen by Electrs
+    let env = env.reset_electrsd()?;
+
+    struct TestCase {
+        name: &'static str,
+        chain: LocalChain,
+        heights: &'static [u32],
```
danielabrozzoni: I know it is verbose, but can we call this `request_heights`?

danielabrozzoni (commented):
A few nits, otherwise LGTM

crates/esplora/src/async_ext.rs
evanlinjin: This should never happen because the genesis block must connect! However, let's say hypothetically, the genesis doesn't connect (because the wallet and esplora are on different networks), the update formed will try replace the genesis, resulting in the `local_chain::MissingGenesisError` which is the error that we want!

crates/esplora/src/async_ext.rs
evanlinjin: We are fetching remote blocks of the same height and comparing it though. I think keeping the variable name as `height` would be better?

danielabrozzoni: Fair enough

crates/esplora/src/async_ext.rs
```diff
@@ -85,103 +91,60 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
...
+            .map(|b| (b.time.height, b.id))
+            .collect::<BTreeMap<u32, BlockHash>>();
+        let new_tip_height = fetched_blocks
+            .keys()
+            .last()
+            .copied()
+            .expect("must have atleast one block");
 
-        // fetch heights that the caller is interested in
+        // Fetch blacks of heights that the caller is interested in, skipping blocks that are
```
notmandatory: typo:
```suggestion
        // Fetch blocks of heights that the caller is interested in, skipping blocks that are
```

crates/esplora/src/blocking_ext.rs
```diff
@@ -78,102 +84,57 @@ impl EsploraExt for esplora_client::BlockingClient {
...
+            .map(|b| (b.time.height, b.id))
+            .collect::<BTreeMap<u32, BlockHash>>();
+        let new_tip_height = fetched_blocks
+            .keys()
+            .last()
+            .copied()
+            .expect("must atleast have one block");
 
-        // fetch heights that the caller is interested in
+        // Fetch blacks of heights that the caller is interested in, skipping blocks that are
```
notmandatory: same typo:
```suggestion
        // Fetch blocks of heights that the caller is interested in, skipping blocks that are
```

notmandatory (commented):
Found a couple typos that should be fixed.

crates/esplora/src/async_ext.rs
evanlinjin: No one saw anything.

notmandatory:
To fix CI in `cont_integration.yml` try replacing:
```
          cargo update -p zip --precise "0.6.2"
```
with: 
```
          cargo update -p zstd-sys --precise "2.0.8+zstd.1.5.5"`
```

evanlinjin (commented):
ACK c5afbaa95d7bc699981d6fe7f9964a41b7587d5c

Thank you @notmandatory for spotting the typos and fixing the CI!

notmandatory (approved):
utACK c5afbaa95d7bc699981d6fe7f9964a41b7587d5c

PR #1269, opened 2024-01-12
Revamp `KeychainTxOutIndex` API to be safer

### LINKED ISSUE #1268: Sync command grabs histories from lookahead spks.
**Describe the bug**  

There is a problem with the sync command in the esplora/electrum examples. Because of the 1000 lookahead default, we are trying to fetch txs for each spk (inc. lookahead).

We are fetching spks using `SpkTxOutIndex::all_spks()` and `SpkTxOutIndex::unused_spks()` to pass into `{Electrum|Esplora}::sync` (which includes lookahead spks).

**To Reproduce**  

Run `sync` command with esplora/electrum cli examples (not wallet).

**Expected behavior**  

We only fetch histories of revealed spks.

**Suggested implementation**

We probably need methods on `KeychainTxOutIndex` for non-lookahead versions for `all_spks` and `unused_spks`.

Maybe it is a bad idea to expose methods of `SpkTxOutIndex` from `KeychainTxOutIndex`?

evanlinjin:
Closes #1268

### Description

Previously `SpkTxOutIndex` methods can be called from `KeychainTxOutIndex` due to the `DeRef` implementation. However, the internal `SpkTxOut` will also contain lookahead spks resulting in an error-prone API.

`SpkTxOutIndex` methods are now not directly callable from `KeychainTxOutIndex`. Methods of `KeychainTxOutIndex` are renamed for clarity. I.e. methods that return an unbounded spk iter are prefixed with `unbounded`.

In addition to this, I also optimized the peek-address logic of `bdk::Wallet` using the optimized `<SpkIterator as Iterator>::nth` implementation.

### Notes to the reviewers

This is mostly refactoring, but can also be considered a bug-fix (as the API before was very problematic).

### Changelog notice

Changed
* Wallet's peek-address logic is optimized by making use of `<SpkIterator as Iterator>::nth`.
* `KeychainTxOutIndex` API is refactored to better differentiate between methods that return unbounded vs stored spks.
* `KeychainTxOutIndex` no longer directly exposes `SpkTxOutIndex` methods via `DeRef`. This was problematic because `SpkTxOutIndex` also contains lookahead spks which we want to hide.

Added
* `SpkIterator::descriptor` method which gets a reference to the internal descriptor. 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -811,8 +811,7 @@ impl<D> Wallet<D> {
         self.chain.tip()
     }
 
-    /// Returns a iterators of all the script pubkeys for the `Internal` and `External` variants in
-    /// `KeychainKind`.
+    /// Get unbounded script pubkey iterators for both `Internal` and `External` keychains.
     ///
     /// This is intended to be used when doing a full scan of your addresses (e.g. after restoring
     /// from seed words). You pass the `BTreeMap` of iterators to a blockchain data source (e.g.
@@ -826,7 +825,7 @@ impl<D> Wallet<D> {
         self.indexed_graph.index.all_unbounded_spk_iters()
     }
 
-    /// Gets an iterator over all the script pubkeys in a single keychain.
+    /// Get an unbounded script pubkey iterator for the given `keychain`.
     ///
     /// See [`all_unbounded_spk_iters`] for more documentation
     ///
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -262,6 +262,11 @@ where
     /// Infallibly return a derived address using the external descriptor, see [`AddressIndex`] for
     /// available address index selection strategies. If none of the keys in the descriptor are derivable
     /// (i.e. does not end with /*) then the same address will always be returned for any [`AddressIndex`].
+    ///
+    /// # Panics
+    ///
+    /// This panics when the caller requests for an address of derivation index greater than the
+    /// BIP32 max index.
     pub fn get_address(&mut self, address_index: AddressIndex) -> AddressInfo {
         self.try_get_address(address_index).unwrap()
     }
@@ -273,6 +278,11 @@ where
     /// see [`AddressIndex`] for available address index selection strategies. If none of the keys
     /// in the descriptor are derivable (i.e. does not end with /*) then the same address will always
     /// be returned for any [`AddressIndex`].
+    ///
+    /// # Panics
+    ///
+    /// This panics when the caller requests for an address of derivation index greater than the
+    /// BIP32 max index.
     pub fn get_internal_address(&mut self, address_index: AddressIndex) -> AddressInfo {
         self.try_get_internal_address(address_index).unwrap()
     }
@@ -649,6 +659,11 @@ impl<D> Wallet<D> {
     ///
     /// A `PersistBackend<ChangeSet>::WriteError` will result if unable to persist the new address
     /// to the `PersistBackend`.
+    ///
+    /// # Panics
+    ///
+    /// This panics when the caller requests for an address of derivation index greater than the
+    /// BIP32 max index.
     pub fn try_get_address(
         &mut self,
         address_index: AddressIndex,
@@ -669,6 +684,11 @@ impl<D> Wallet<D> {
     /// see [`AddressIndex`] for available address index selection strategies. If none of the keys
     /// in the descriptor are derivable (i.e. does not end with /*) then the same address will always
     /// be returned for any [`AddressIndex`].
+    ///
+    /// # Panics
+    ///
+    /// This panics when the caller requests for an address of derivation index greater than the
+    /// BIP32 max index.
     pub fn try_get_internal_address(
         &mut self,
         address_index: AddressIndex,
@@ -691,6 +711,11 @@ impl<D> Wallet<D> {
     /// See [`AddressIndex`] for available address index selection strategies. If none of the keys
     /// in the descriptor are derivable (i.e. does not end with /*) then the same address will
     /// always be returned for any [`AddressIndex`].
+    ///
+    /// # Panics
+    ///
+    /// This panics when the caller requests for an address of derivation index greater than the
+    /// BIP32 max index.
     fn _get_address(
         &mut self,
         keychain: KeychainKind,
@@ -710,12 +735,14 @@ impl<D> Wallet<D> {
                 let ((index, spk), index_changeset) = txout_index.next_unused_spk(&keychain);
                 (index, spk.into(), Some(index_changeset))
             }
-            AddressIndex::Peek(index) => {
-                let (index, spk) = txout_index
-                    .unbounded_spk_iter(&keychain)
-                    .take(index as usize + 1)
-                    .last()
-                    .unwrap();
+            AddressIndex::Peek(mut peek_index) => {
+                let mut spk_iter = txout_index.unbounded_spk_iter(&keychain);
+                if !spk_iter.descriptor().has_wildcard() {
+                    peek_index = 0;
+                }
+                let (index, spk) = spk_iter
+                    .nth(peek_index as usize)
+                    .expect("derivation index is out of bounds");
                 (index, spk, None)
             }
         };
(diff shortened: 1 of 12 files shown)


crates/bdk/src/wallet/mod.rs
```diff
@@ -792,36 +819,36 @@ impl<D> Wallet<D> {
     ///
     /// Note carefully that iterators go over **all** script pubkeys on the keychains (not what
     /// script pubkeys the wallet is storing internally).
-    pub fn spks_of_all_keychains(
+    pub fn all_unbounded_spk_iters(
```
danielabrozzoni, line 822: The doc for this method is missing a \`, can you fix it please? Line 814, there's no ` at the start of the word "External"

example-crates/example_esplora/src/main.rs
```diff
@@ -235,32 +235,32 @@ fn main() -> anyhow::Result<()> {
                 if *all_spks {
                     let all_spks = graph
                         .index
-                        .all_spks()
-                        .iter()
-                        .map(|(k, v)| (*k, v.clone()))
+                        .revealed_spks()
```
danielabrozzoni, line 238: Why is this iterating over revealed only? Was using `all_spks()` a bug?

crates/bdk/src/wallet/mod.rs
```diff
@@ -792,36 +819,36 @@ impl<D> Wallet<D> {
...
 
     /// Gets an iterator over all the script pubkeys in a single keychain.
     ///
-    /// See [`spks_of_all_keychains`] for more documentation
+    /// See [`all_unbounded_spk_iters`] for more documentation
     ///
-    /// [`spks_of_all_keychains`]: Self::spks_of_all_keychains
-    pub fn spks_of_keychain(
+    /// [`all_unbounded_spk_iters`]: Self::all_unbounded_spk_iters
+    pub fn unbounded_spk_iter(
```
danielabrozzoni, line 833: Is "unbounded" meant to be the contrary of "stored only"? 

To me, "unbounded_spk_iter" makes it sound like that there is a way to get a "bounded" spk iter, but then there's no method called like that. Maybe "all" would be a better word to communicate "it's all of the spks, not the stored only" - but then I wouldn't know how to call "all_unbounded_spk_iters" 😅 

This is definitely not a blocker, but I'm not convinced that "unbounded" is the right word here

crates/chain/src/keychain/txout_index.rs
```diff
@@ -133,8 +125,11 @@ impl<K> KeychainTxOutIndex<K> {
     }
 }
 
+/// Methods that are *re-exposed* from the internal [`SpkTxOutIndex`].
 impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// Return a reference to the internal [`SpkTxOutIndex`].
+    ///
+    /// **WARNING:** The internal index will contain lookahead spks.
```
danielabrozzoni: nit: maybe we should expand a little bit on stored vs revealed vs lookahead.

danielabrozzoni (commented):
A few comments. I still haven't finished my review, but I wanted to post this in the meantime :)

example-crates/example_esplora/src/main.rs
evanlinjin: Yes `all_spks` includes lookahead spks because it if `Deref`ed from `SpkTxOutIndex`. We want to hide all lookahead spks.

crates/bdk/src/wallet/mod.rs
evanlinjin: I don't think `all` is a better option. It can be interpreted as "all-revealed" or "all-stored". I'll be open for a better name, let me know if you have any further ideas.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -144,8 +139,103 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+}
+
+impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
+    /// Iterate through all keychain identifiers.
+    pub fn keychains(&self) -> impl Iterator<Item = &K> + '_ {
+        self.keychains.keys()
+    }
+
+    /// Return a reference to the internal map of keychain to descriptors.
+    pub fn keychain_descriptors(&self) -> &BTreeMap<K, Descriptor<DescriptorPublicKey>> {
```
LLFourn: I don't think we need `keychains` and `keychain_descriptors`.

crates/chain/src/keychain/txout_index.rs
LLFourn: I think we should do this on the main type. It already does this a bit but it could be clearer.

crates/bdk/src/wallet/mod.rs
LLFourn: I like unbounded better. Although there is no "bounded" version it implies the others are bounded (which they are). "unbounded" tells the caller that it's never going to stop so it would only make sense to use this in a "stop gap" context.

crates/chain/src/keychain/txout_index.rs
evanlinjin: 83e7b7ec402bef27c3f83876f346ec2abd23eff1

LLFourn (approved):
ACK 83e7b7ec402bef27c3f83876f346ec2abd23eff1

crates/chain/src/keychain/txout_index.rs
```diff
@@ -133,8 +165,12 @@ impl<K> KeychainTxOutIndex<K> {
     }
 }
 
+/// Methods that are *re-exposed* from the internal [`SpkTxOutIndex`].
 impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// Return a reference to the internal [`SpkTxOutIndex`].
+    ///
+    /// **WARNING:** The internal index will contain lookahead spks. Refer to
+    /// [struct-level docs](KeychainTxOutIndex) for more about `lookahead`.
```
danielabrozzoni, line 176: Do you think we will eventually avoid exposing `inner`, or is it useful for some use cases, and so exposing it with a warning is good enough?

example-crates/example_esplora/src/main.rs
danielabrozzoni: Should we clarify this change in the changelog?

danielabrozzoni (commented):
From my understanding, this PR is making some `SpkTxOutIndex` methods slightly more difficult to call (you need to use `KeychainTxOutIndex::inner` instead of just using `Deref`).

These are SpkTxOutIndex methods that are not directly exposed from `KeychainTxOutIndex`:
- `all_spks`
- `insert_spk`
- `outputs_in_range`
- `txouts`
- `txouts_in_tx`

It makes sense to me that `insert_spk` and `all_spks` are not exposed: you should let the `KeychainTxOutIndex` insert the spks, and `all_spks` exposes the lookahead spks.

But I just can't figure out why `outputs_in_range`, `txouts`, and `txouts_in_tx` shouldn't be exposed as well - is there a particular reason?

crates/chain/src/keychain/txout_index.rs
evanlinjin: I would assume in some cases getting lookahead spks will be handy.

I.e. we haven't finalized how a CBF chain-source will look like and how it will interact with `bdk_chain` structures. However, it might make sense to match lookahead spks against GCS (filters). Or maybe using an *unbounded* spk iterator makes more sense?

Maybe later on, we determine that it will make sense to use helper methods on `KeychainTxOutIndex` instead. However, for most usecases now, outputting lookahead spks is almost always non-intentional.

evanlinjin: In summary, I think we should keep `.inner` with a warning (for now).

evanlinjin:
> But I just can't figure out why `outputs_in_range`, `txouts`, and `txouts_in_tx` shouldn't be exposed as well - is there a particular reason?

@danielabrozzoni I didn't include them because they weren't used in the examples. I guess it doesn't hurt to include?

danielabrozzoni (approved):
ACK 71fff1613d1c167d180636857063e181f6acedf1

PR #1270, opened 2024-01-13
Improve performance of `bdk_file_store::EntryIter`

evanlinjin:
### Description

`EntryIter` performance is improved by reducing syscalls. The underlying file reader is wrapped with `BufReader` (to reduce calls to `read` and `seek`).

Two new tests are introduced. One ensures correct behavior when the last changeset write is too short. The other ensures the next write position is correct after a short read.

### Notes to the reviewers

This is extracted from #1172 as suggested by https://github.com/bitcoindevkit/bdk/pull/1172#pullrequestreview-1817465627.

### Changelog notice

Changed
* `EntryIter` performance is improved by reducing syscalls.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/file_store/src/entry_iter.rs
+++ b/crates/file_store/src/entry_iter.rs
@@ -1,11 +1,13 @@
 use bincode::Options;
 use std::{
     fs::File,
-    io::{self, Seek},
+    io::{self, BufReader, Seek},
     marker::PhantomData,
 };
 
 use crate::bincode_options;
+
+type EntryReader<'t> = CountingReader<BufReader<&'t mut File>>;
 
 /// Iterator over entries in a file store.
 ///
@@ -14,7 +16,7 @@ use crate::bincode_options;
 ///
 /// [`next`]: Self::next
 pub struct EntryIter<'t, T> {
-    db_file: Option<&'t mut File>,
+    db_file: Option<EntryReader<'t>>,
 
     /// The file position for the first read of `db_file`.
     start_pos: Option<u64>,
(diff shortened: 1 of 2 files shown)


crates/file_store/src/entry_iter.rs
```diff
@@ -38,44 +40,43 @@ where
...
 
-impl From<io::Error> for IterError {
-    fn from(value: io::Error) -> Self {
-        IterError::Io(value)
+impl<'t, T> Drop for EntryIter<'t, T> {
+    fn drop(&mut self) {
+        // This syncs the underlying file's offset with the buffer's position. This way, we
+        // maintain the correct position to start the next read/write.
+        #[allow(clippy::seek_from_current)]
+        let _ = self.db_file.seek(io::SeekFrom::Current(0));
```
LLFourn: Can you change this so we instead just seek the underlying file to the `stream_position` that the BufReader returns (by recording it and then into_inner or whatever to get the file). This is more explicit about what we're trying to do and doesn't have clippy ignores :see_no_evil:.

Something feels off about this. It we be cool if we could get rid of the implicit state being stored in the file position and make the user explicitly start writing at a certain position that we've advised instead.

evanlinjin: @LLFourn How is this?

```rust
        if let Ok(pos) = self.db_file.stream_position() {
            let _ = self.db_file.get_mut().seek(io::SeekFrom::Start(pos));
        }
```

Can we do explicit position writes in a separate PR? I think it is out of scope for this PR (+ I would like to use these changes here).

LLFourn (approved):
ACK 51bd01b3dd3d68c1628d74be9dc6d0f1cf1f15fe

PR #1272, opened 2024-01-13
chore: move Balance out of keychain module

### LINKED ISSUE #128: Merge `tx_data_traits.rs` into `chain_data.rs`.

startup-dreamer:
### Description

Tries to Fix: bitcoindevkit/bdk_wallet#128 

Moved `Balance` from keychain.rs to tx_graph.rs and did the rexport for `bdk::wallet` module.

#### All Submissions:


evanlinjin:
Thank you for this contribution.

However, the discussion about where to put everything in `bdk_chain` is not final. I rather we move everything in one PR (after we come to consensus, which has not happened yet).

For the meantime, I will provide my thoughts on the matter. I think we should have a file `chain/types.rs` that contains all smaller public traits and structs (replacing `chain_data.rs` and `tx_data_traits.rs` files). `Balance` should be in `types.rs`.

Also note that this is a duplicate of: bitcoindevkit/bdk#1175

LLFourn:
It would be great to avoid a `types.rs`. This is totally unhelpful name because everything we declare is a type. It's fine to put things in `lib.rs` if there's no where else it belongs.

It would be fine to put `tx_data_traits.rs` into `chain_data.rs`. I don't recall why `tx_data_traits.rs` was created.

startup-dreamer:
Closing this in support of bitcoindevkit/bdk#1283 from [This comment](https://github.com/bitcoindevkit/bdk_wallet/issues/128)

PR #1286, opened 2024-01-17
doc, example(bdk): fix derivation path in mnemonic_to_descriptors

ValuedMammal:
This will help avoid confusion with mainnet descriptors.

#### All Submissions:

### DIFF
--- a/crates/bdk/examples/mnemonic_to_descriptors.rs
+++ b/crates/bdk/examples/mnemonic_to_descriptors.rs
@@ -33,8 +33,8 @@ fn main() -> Result<(), anyhow::Error> {
     let mnemonic_with_passphrase = (mnemonic, None);
 
     // define external and internal derivation key path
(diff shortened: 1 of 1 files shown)


danielabrozzoni (approved):
ACK d494f63d08f60ce01c22ff08f7e88a80a7be2522

A note for the future: I didn't immediately understand why you were doing this change, I had to search BIP86 and then BIP44 only to find out that the coin_type was wrong. It would be better if the PR description or the commit message included a short explanation of what was wrong and how the commit is fixing it. Something like:
```
doc, example(bdk): fix derivation path in mnemonic_to_descriptors

We were using a mainnet derivation path (coin_type = 0) in a testnet example. This commit changes the coin_type to be 1 (= Bitcoin Testnet, as defined in BIP44).
```

(^ The wording of this commit message can be improved haha, it's just an example)

Thanks for your contribution!

PR #1292, opened 2024-01-18
fix(store): Remove lifetime

LLFourn:
Remove gratuitous use of lifetimes in the main persistence struct `Store`. Having lifetimes on this means that you have to keep the magic bytes alive longer than the database which is particularly offensive if you have to send the database to another thread. On top of that the bytes aren't even read.


LLFourn:
I pushed another commit to this to always make sure that we're committing things in example_cli (I noticed a bug while fixing clippy errors).

evanlinjin (approved):
ACK e6433fb2c1526cff37e3a17cc71986759ebac30f

PR #1293, opened 2024-01-22
Use a BufReader in bdk_file_store

LLFourn:
PR over #1270 to simplify it. Key observation is that you can use stream_position in `BufReader` without making a system call. I think @evanlinjin  introduced CountingReader to keep track of where we are in the stream but `BufReader` already does this (correct me if I'm wrong).
Take a close look please @evanlinjin. It passes the tests but I might be missing something.

### DIFF
--- a/crates/file_store/src/entry_iter.rs
+++ b/crates/file_store/src/entry_iter.rs
@@ -1,11 +1,13 @@
 use bincode::Options;
 use std::{
     fs::File,
-    io::{self, Seek},
+    io::{self, BufReader, Seek},
     marker::PhantomData,
 };
 
 use crate::bincode_options;
+
+type EntryReader<'t> = CountingReader<BufReader<&'t mut File>>;
 
 /// Iterator over entries in a file store.
 ///
@@ -14,7 +16,7 @@ use crate::bincode_options;
 ///
 /// [`next`]: Self::next
 pub struct EntryIter<'t, T> {
-    db_file: Option<&'t mut File>,
+    db_file: Option<EntryReader<'t>>,
 
     /// The file position for the first read of `db_file`.
     start_pos: Option<u64>,
@@ -24,7 +26,7 @@ pub struct EntryIter<'t, T> {
 impl<'t, T> EntryIter<'t, T> {
     pub fn new(start_pos: u64, db_file: &'t mut File) -> Self {
         Self {
-            db_file: Some(db_file),
+            db_file: Some(CountingReader::new(BufReader::new(db_file))),
             start_pos: Some(start_pos),
             types: PhantomData,
         }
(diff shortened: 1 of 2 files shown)


crates/file_store/src/entry_iter.rs
```diff
@@ -38,44 +40,34 @@ where
...
+            if let Some(start) = self.start_pos.take() {
+                self.db_file.seek(io::SeekFrom::Start(start))?;
+            }
 
-            match bincode_options().deserialize_from(&*f) {
-                Ok(changeset) => {
-                    f.stream_position()?;
-                    Ok(Some(changeset))
-                }
+            let pos_before_read = self.db_file.stream_position()?;
```
evanlinjin, line 51: * `Seek::stream_position` is equivalent to `self.seek(SeekFrom::Current(0))` [(ref)](https://doc.rust-lang.org/std/io/struct.BufReader.html#method.stream_position).
* Every call to `BufReader::seek` discards the buffer [(ref)](https://doc.rust-lang.org/std/io/struct.BufReader.html#method.seek).

We are now discarding the buffer on every call to `next`.

This is how `BufReader::seek` is implemented:

```rust
    fn seek(&mut self, pos: SeekFrom) -> io::Result<u64> {
        let result: u64;
        if let SeekFrom::Current(n) = pos {
            let remainder = (self.buf.filled() - self.buf.pos()) as i64;
            // it should be safe to assume that remainder fits within an i64 as the alternative
            // means we managed to allocate 8 exbibytes and that's absurd.
            // But it's not out of the realm of possibility for some weird underlying reader to
            // support seeking by i64::MIN so we need to handle underflow when subtracting
            // remainder.
            if let Some(offset) = n.checked_sub(remainder) {
                result = self.inner.seek(SeekFrom::Current(offset))?;
            } else {
                // seek backwards by our remainder, and then by the offset
                self.inner.seek(SeekFrom::Current(-remainder))?;
                self.discard_buffer();
                result = self.inner.seek(SeekFrom::Current(n))?;
            }
        } else {
            // Seeking with Start/End doesn't care about our buffer length.
            result = self.inner.seek(pos)?;
        }
        self.discard_buffer();
        Ok(result)
    }
```

LLFourn: > * `Seek::stream_position` is equivalent to `self.seek(SeekFrom::Current(0))` [(ref)](https://doc.rust-lang.org/std/io/struct.BufReader.html#method.stream_position). 

That's not what that reference says :)

evanlinjin: Oh damn I missed that!

crates/file_store/src/entry_iter.rs
```diff
@@ -40,45 +40,34 @@ where
...
-        if let Some(r) = self.db_file.as_mut() {
-            // This syncs the underlying file's offset with the buffer's position. This way, no data
-            // is lost with future reads.
-            let _ = r.stream_position();
-        }
+            }
+        })()
+        .transpose()
     }
 }
```
evanlinjin, line 72: I think we still need a `Drop` implementation that syncs the underlying file's offset. However, we should use `self.seek(SeekFrom::Current(0))`.

LLFourn: Can we write a test that demonstrates this need? I think I understand the rationale I'll see if I can come up one.

[EDIT] or maybe we can use a `BufReader` in the main type (not just the iterator).

evanlinjin: @LLFourn I don't think using a BufReader with the internal file getting written to is a good idea. We'll need to do some crazy stuff with `.get_mut` and flush the buffer with every write.

I thought that the test I added demonstrated the need for the `Drop` impl but clearly not. We need a test to write after a failed read without closing the file in-between.

evanlinjin:
@LLFourn I cherry-picked these commits onto #1270. Can we close this PR and continue discussion on #1270?

evanlinjin:
This is continued in #1270

PR #1294, opened 2024-01-22
Expose `SpkIterator::new_with_range`

evanlinjin:
### Description

This method is useful, I'm not sure why it is not public.

I've also updated the docs and tests.

### Changelog notice

Added
* Expose `SpkIterator::new_with_range`

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/spk_iter.rs
+++ b/crates/chain/src/spk_iter.rs
@@ -43,18 +43,24 @@ impl<D> SpkIterator<D>
 where
     D: Borrow<Descriptor<DescriptorPublicKey>>,
 {
-    /// Creates a new script pubkey iterator starting at 0 from a descriptor.
+    /// Create a new script pubkey iterator from `descriptor`.
+    ///
+    /// This iterates from derivation index 0 and stops at index 0x7FFFFFFF (as specified in
+    /// BIP-32). Non-wildcard descriptors will only return one script pubkey at derivation index 0.
+    ///
+    /// Use [`new_with_range`](SpkIterator::new_with_range) to create an iterator with a specified
+    /// derivation index range.
     pub fn new(descriptor: D) -> Self {
         SpkIterator::new_with_range(descriptor, 0..=BIP32_MAX_INDEX)
     }
 
-    // Creates a new script pubkey iterator from a descriptor with a given range.
-    // If the descriptor doesn't have a wildcard, we shorten whichever range you pass in
-    // to have length <= 1. This means that if you pass in 0..0 or 0..1 the range will
-    // remain the same, but if you pass in 0..10, we'll shorten it to 0..1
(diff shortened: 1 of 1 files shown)


crates/chain/src/spk_iter.rs
```diff
@@ -43,18 +43,24 @@ impl<D> SpkIterator<D>
...
-    // Creates a new script pubkey iterator from a descriptor with a given range.
-    // If the descriptor doesn't have a wildcard, we shorten whichever range you pass in
-    // to have length <= 1. This means that if you pass in 0..0 or 0..1 the range will
-    // remain the same, but if you pass in 0..10, we'll shorten it to 0..1
-    // Also note that if the descriptor doesn't have a wildcard, passing in a range starting
-    // from n > 0, will return an empty iterator.
-    pub(crate) fn new_with_range<R>(descriptor: D, range: R) -> Self
+    /// Create a new script pubkey iterator from `descriptor` and a given `range`.
+    ///
+    /// The end bound of the range will be shortened to 0 (inclusive) for non-wildcard descriptors
```
LLFourn: "the end bound of the range will be shortened to 0 (inclusive)" is a very complicated way of saying "non-wildcard descriptors will only emit a single item".

crates/chain/src/spk_iter.rs
```diff
@@ -73,11 +79,9 @@ where
...
         end = end.min(BIP32_MAX_INDEX + 1);
 
-        if !descriptor.borrow().has_wildcard() {
-            // The length of the range should be at most 1
-            if end != start {
-                end = start + 1;
-            }
+        // A non-wildcard descriptor will only have index 0.
+        if !descriptor.borrow().has_wildcard() && end > 1 {
+            end = 1;
```
LLFourn: what happens if we just remove this. It seems to be checked in `next`. Looks unnecessary.

LLFourn (changes requested):

crates/chain/src/spk_iter.rs
evanlinjin: Removed.

crates/chain/src/spk_iter.rs
```diff
@@ -250,6 +249,14 @@ mod test {
             SpkIterator::new_with_range(&no_wildcard_descriptor, 1..=2).next(),
             None
         );
+        assert_eq!(
```
danielabrozzoni, line 252: I'm not sure why these tests are being added, but I won't complain about adding new tests 😛

danielabrozzoni (approved):
utACK 8305e648494cbdd1f6438904ec367f8442fba606

LLFourn (approved):
ACK 8305e648494cbdd1f6438904ec367f8442fba606

LLFourn (approved):
DELETED COMMENT

PR #1299, opened 2024-01-25
fix(file_store): rm lifetime from `FileError`

evanlinjin:
### Description

The lifetime on the error needed to be the same as the input magic bytes which was annoying.

### Changelog notice

Fixed
* Remove `bdk_file_store::FileError` lifetime.

#### All Submissions:


LLFourn (approved):
ACK 5611c9e42adb4106f48039e19ab0797b266c7fff

Issue #1265, opened 2024-01-10
Option for Electrum and Esplora APIs to include floating `TxOut`s in updates for fee calculation


evanlinjin:
**Describe the enhancement**  

BDK does not store the transaction fee in a separate field. It needs to be calculated from the difference between prev outputs and outputs created.

For transactions that are received from an external wallet, we would not have prev txout data by default. It'll be handy to have an option to get prev txouts for these transactions and have them inserted as floating txouts in the `TxGraph`.

**Use case**  

Some wallets would want to be able to calculate fees for all transactions that it includes.

**Additional context**

LLFourn:
In esplora it should not be an option it just should be done. The information is already fetched it just isn't inserted. With electrum I guess you have to make it an option.

evanlinjin:
That's a good point @LLFourn. Should be straightforward for Esplora then.

LagginTimes:
I can take a look at this.

ValuedMammal:
possibly related #1419

evanlinjin:
@LagginTimes will also implement this for Electrum.

LagginTimes:
The `electrum` portion of this ticket has been addressed in #1403.

PR #1308, opened 2024-01-31
feat(esplora): include previous `TxOut`s for fee calculation

LagginTimes:
### Description

Partially implements #1265.

The previous `TxOut` for transactions received from an external wallet are added as floating `TxOut`s to `TxGraph` to allow for fee calculation.

### Notes to the reviewers

Currently only the `esplora` portion of #1265 has been implemented.
The `electrum` portion will potentially be done in a new PR, as discussed on the 1/30/24 Lib call.

#### To Do:
* [X] Implement `electrum` portion of #1265.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/esplora/src/async_ext.rs
+++ b/crates/esplora/src/async_ext.rs
@@ -1,7 +1,7 @@
 use async_trait::async_trait;
 use bdk_chain::collections::btree_map;
 use bdk_chain::{
-    bitcoin::{BlockHash, OutPoint, ScriptBuf, Txid},
+    bitcoin::{BlockHash, OutPoint, ScriptBuf, TxOut, Txid},
     collections::BTreeMap,
     local_chain::{self, CheckPoint},
     BlockId, ConfirmationTimeHeightAnchor, TxGraph,
@@ -203,6 +203,24 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
                         let _ = graph.insert_tx(tx.to_tx());
                         if let Some(anchor) = anchor_from_status(&tx.status) {
                             let _ = graph.insert_anchor(tx.txid, anchor);
+                        }
+
+                        let previous_outputs = tx.vin.iter().filter_map(|vin| {
+                            let prevout = vin.prevout.as_ref()?;
+                            Some((
+                                OutPoint {
+                                    txid: vin.txid,
+                                    vout: vin.vout,
+                                },
+                                TxOut {
+                                    script_pubkey: prevout.scriptpubkey.clone(),
+                                    value: prevout.value,
+                                },
+                            ))
+                        });
+
+                        for (outpoint, txout) in previous_outputs {
+                            let _ = graph.insert_txout(outpoint, txout);
                         }
                     }
                 }
--- a/crates/esplora/tests/async_ext.rs
+++ b/crates/esplora/tests/async_ext.rs
@@ -109,6 +109,28 @@ pub async fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
         )
         .await?;
 
+    // Check to see if we have the floating txouts available from our two created transactions'
+    // previous outputs in order to calculate transaction fees.
+    for tx in graph_update.full_txs() {
+        // Retrieve the calculated fee from `TxGraph`, which will panic if we do not have the
+        // floating txouts available from the transactions' previous outputs.
+        let fee = graph_update.calculate_fee(tx.tx).expect("Fee must exist");
+
+        // Retrieve the fee in the transaction data from `bitcoind`.
+        let tx_fee = env
+            .bitcoind
+            .client
+            .get_transaction(&tx.txid, None)
+            .expect("Tx must exist")
+            .fee
+            .expect("Fee must exist")
+            .abs()
+            .to_sat() as u64;
+
+        // Check that the calculated fee matches the fee from the transaction data.
+        assert_eq!(fee, tx_fee);
+    }
+
     let mut graph_update_txids: Vec<Txid> = graph_update.full_txs().map(|tx| tx.txid).collect();
     graph_update_txids.sort();
     let mut expected_txids = vec![txid1, txid2];
--- a/crates/esplora/tests/blocking_ext.rs
+++ b/crates/esplora/tests/blocking_ext.rs
@@ -136,6 +136,28 @@ pub fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
         1,
     )?;
 
+    // Check to see if we have the floating txouts available from our two created transactions'
+    // previous outputs in order to calculate transaction fees.
+    for tx in graph_update.full_txs() {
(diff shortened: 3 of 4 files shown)


crates/esplora/src/async_ext.rs
```diff
@@ -204,6 +204,30 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
...
+                                            vout: vin.vout,
+                                        },
+                                        TxOut {
+                                            script_pubkey: po.scriptpubkey,
+                                            value: po.value,
+                                        },
+                                    )
+                                })
+                            })
+                            .collect();
```
evanlinjin: Here are a couple of principles when writing rust code:

1. Avoid unnecessary cloning.
2. Make use of the lazy property of iterators (don't prematurely collect if you don't have to!).
3. Using `?` can make the codebase easier to read.

Please also update the `blocking_ext.rs` code.

```suggestion
                        let previous_outputs = tx.vin.iter().filter_map(|vin| {
                            let prevout = vin.prevout.as_ref()?;
                            Some((
                                OutPoint {
                                    txid: vin.txid,
                                    vout: vin.vout,
                                },
                                TxOut {
                                    script_pubkey: prevout.scriptpubkey.clone(),
                                    value: prevout.value,
                                },
                            ))
                        });
```

evanlinjin (changes requested):
Looking good! Just some small changes.

crates/esplora/tests/async_ext.rs
```diff
@@ -109,6 +109,19 @@ pub async fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
         )
         .await?;
 
+    // Check to see if we have the floating txouts available from our two created transactions'
+    // previous outputs in order to calculate transaction fees.
+    let fee_calculations = graph_update
+        .full_txs()
+        .map(|tx| graph_update.calculate_fee(tx.tx).is_ok())
```
evanlinjin: I think we should check whether the fee calculation is correct (instead of whether a fee calculation has occurred).

You can either use Esplora's [`GET /tx/:txid`](https://github.com/Blockstream/esplora/blob/master/API.md#get-txtxid) endpoint or bitcoind RPC's `gettransaction` call to get the fee.

This is the last thing and that's an ACK from me!

crates/esplora/tests/async_ext.rs
```diff
@@ -109,6 +109,27 @@ pub async fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
         )
         .await?;
 
+    // Check to see if we have the floating txouts available from our two created transactions'
+    // previous outputs in order to calculate transaction fees.
+    graph_update.full_txs().for_each(|tx| {
+        let calculate_fee_result = graph_update.calculate_fee(tx.tx);
+        let get_tx_result = env.bitcoind.client.get_transaction(&tx.txid, None);
+
+        // Check if fee calculation and transaction retrieval were successful
```
danielabrozzoni: It seems to me that you are checking that the fee calculation and transaction retrieval had the same result, not that they were successful. Is this what you intended to do? If so, in which cases both bitcoind and our tx_graph can't calculate the fee and we are ok with it?

crates/esplora/tests/async_ext.rs
```diff
@@ -109,6 +109,27 @@ pub async fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
...
+        let tx_fee_exists = get_tx_result.as_ref().map_or(false, |tx| tx.fee.is_some());
+        assert_eq!(calculate_fee_result.is_ok(), tx_fee_exists);
+
+        // Check if the calculated fee value matches the fee from transaction retrieval
+        if let (Ok(fee), Ok(transaction)) = (calculate_fee_result, get_tx_result) {
+            let tx_fee = transaction
+                .fee
+                .unwrap_or_else(|| panic!("Fee not found for transaction {:?}", tx.txid))
```
danielabrozzoni: You are sure that `transaction.fee` exists, as at this point `calculate_fee_result` is ok, and you check above that `tx_fee_exist` is equal to `calculate_fee_result.is_ok()`. This can just be an `unwrap` or `expect(*explanation of why this must exist*)`

crates/esplora/tests/blocking_ext.rs
```diff
@@ -136,6 +136,32 @@ pub fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
...
+                .abs()
+                .to_sat() as u64;
+            assert_eq!(fee, tx_fee);
+        }
+    });
+    // Ensure there are exactly two fee calculations.
+    // assert_eq!(fee_calculations.len(), 2);
+
+    // Ensure that both fee calculations are successful.
+    // assert!(fee_calculations.iter().all(|&is_ok| is_ok));
```
danielabrozzoni: Should be removed

LagginTimes: Oops.

crates/esplora/tests/async_ext.rs
LagginTimes: My intention was that both were successful. I see the logical hole I left and it should be fixed now.

crates/esplora/tests/async_ext.rs
```diff
@@ -109,6 +109,27 @@ pub async fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
         )
         .await?;
 
+    // Check to see if we have the floating txouts available from our two created transactions'
+    // previous outputs in order to calculate transaction fees.
+    graph_update.full_txs().for_each(|tx| {
+        let calculate_fee_result = graph_update.calculate_fee(tx.tx);
+        let get_tx_result = env.bitcoind.client.get_transaction(&tx.txid, None);
```
evanlinjin: Is there a situation where `bitcoind` cannot get the tx or the fee of the tx here? It'll be easier to read if we just return the error for `get_transaction` and unwrap the fee.

evanlinjin: I think we should avoid storing `Result`s when we don't have to. It makes it easier to read if error handling goes directly after getting the result.

> Try `TxGraph::calculate_fee` -> We `expect` no error as we have all prev outputs! -> Store fee value in variable.

evanlinjin: Because `.for_each` takes in a closure which returns nothing, whereas we want to handle our errors, I think we should avoid using it here. The test returns a Result, might as well use it!

evanlinjin (approved):
ACK 499059549f150f40f24e841349c1f907ff160635

Although I do not like the use of `.for_each` (as mentioned in https://github.com/bitcoindevkit/bdk/pull/1308#discussion_r1477312297), I also do not think it should be a blocker.

evanlinjin (approved):
re-ACK 552f11cb5f66a9366dbc05c801fcae043ebe5b82

danielabrozzoni (approved):
ACK 552f11cb5f66a9366dbc05c801fcae043ebe5b82

PR #1200, opened 2023-11-08
fix(bdk): Check if we're using the correct internal key before signing

danielabrozzoni:
### Description

Fixes #1142

We would previously sign with whatever x_only_pubkey we had in hand, without first checking if it was the right key or not. This effectively meant that adding multiple taproot PrivateKey signers would produce unbroadcastable transactions.

### Changelog notice

- Fix a bug related to taproot signing with internal keys. We would previously sign with the first private key we had, without checking if it was the correct internal key or not.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [x] I've added tests to reproduce the issue which are now passing
* [x] I'm linking the issue being fixed by this PR

PR #1310, opened 2024-01-31
Remove extra taproot fields when finalizing PSBT

### LINKED ISSUE #1243: BIP 371 noncompliance: Taproot fields remain after adding `PSBT_IN_FINAL_SCRIPTWITNESS` in finalize_psbt
**Describe the bug**  

Finalizers should remove `PSBT_IN_TAP_*` and `PSBT_OUT_TAP_BIP32_DERIVATION` after `PSBT_IN_FINAL_SCRIPTWITNESS` is constructed according to [BIP 371 spec](https://github.com/bitcoin/bips/blob/master/bip-0371.mediawiki#specification)

**To Reproduce**  

Fund a psbt with a taproot input. Call `wallet.sign` on it. See `PSBT_IN_TAP_KEY_SIG` present on output.

**Expected behavior**  

Fund a psbt with a taproot input. Call `wallet.sign` on it. See `PSBT_IN_TAP_KEY_SIG` not present.

**Build environment**  
 - BDK-CLI tag/commit: 1.0.0-alpha.1
 - Rust/Cargo version:  1.63.0, 1.76.0-nightly
 - Rust/Cargo target: wasm32-unknown-unknown

**Additional context**  

I ran into this problem trying to take payjoin BDK-output Original PSBTs and process them at the receiver with `walletprocesspsbt` on bitcoind, and bitcoind failed to skip the inputs that it encountered having PSBT_IN_TAP_KEY_SIG belonging to the sender. Typically it'd just skip them.

I believe the issue could be fixed by calling `clear()` on the aforementioned fields here

https://github.com/bitcoindevkit/bdk/blob/b13505c1c3e4d38c3cf47f34e7b18efb822b3b33/crates/bdk/src/wallet/mod.rs#L1883-L1887

ValuedMammal:
We currently allow removing `partial_sigs` from a finalized PSBT, which is relevant to non-taproot inputs, however taproot related PSBT fields were left in place despite the recommendation of BIP371 to remove them once the `final_script_witness` is constructed. This can cause confusion for parsers that encounter extra taproot metadata in an already satisfied input.

Fix this by introducing a new member to SignOptions `remove_taproot_extras`, which when true will remove extra taproot related data from a PSBT upon successful finalization. This change makes removal of all taproot extras the default but configurable.

fixes #1243

### Notes to the reviewers

If there's a better or more descriptive name for `remove_taproot_extras`, I'm open to changing it.

### Changelog notice

Fixed an [issue](https://github.com/bitcoindevkit/bdk/issues/1243) finalizing taproot inputs following BIP371

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -1972,11 +1972,26 @@ impl<D> Wallet<D> {
                             if sign_options.remove_partial_sigs {
                                 psbt_input.partial_sigs.clear();
                             }
+                            if sign_options.remove_taproot_extras {
+                                // We just constructed the final witness, clear these fields.
+                                psbt_input.tap_key_sig = None;
+                                psbt_input.tap_script_sigs.clear();
+                                psbt_input.tap_scripts.clear();
+                                psbt_input.tap_key_origins.clear();
+                                psbt_input.tap_internal_key = None;
+                                psbt_input.tap_merkle_root = None;
+                            }
                         }
                         Err(_) => finished = false,
                     }
                 }
                 None => finished = false,
+            }
+        }
+
+        if finished && sign_options.remove_taproot_extras {
+            for output in &mut psbt.outputs {
+                output.tap_key_origins.clear();
             }
         }
 
--- a/crates/bdk/tests/psbt.rs
+++ b/crates/bdk/tests/psbt.rs
@@ -161,16 +161,26 @@ fn test_psbt_fee_rate_with_missing_txout() {
 fn test_psbt_multiple_internalkey_signers() {
     use bdk::signer::{SignerContext, SignerOrdering, SignerWrapper};
     use bdk::KeychainKind;
-    use bitcoin::{secp256k1::Secp256k1, PrivateKey};
-    use miniscript::psbt::PsbtExt;
+    use bitcoin::key::TapTweak;
+    use bitcoin::secp256k1::{schnorr, KeyPair, Message, Secp256k1, XOnlyPublicKey};
+    use bitcoin::sighash::{Prevouts, SighashCache, TapSighashType};
+    use bitcoin::{PrivateKey, TxOut};
     use std::sync::Arc;
 
     let secp = Secp256k1::new();
-    let (mut wallet, _) = get_funded_wallet(get_test_tr_single_sig());
+    let wif = "cNJmN3fH9DDbDt131fQNkVakkpzawJBSeybCUNmP1BovpmGQ45xG";
+    let desc = format!("tr({})", wif);
+    let prv = PrivateKey::from_wif(wif).unwrap();
+    let keypair = KeyPair::from_secret_key(&secp, &prv.inner);
+
+    let (mut wallet, _) = get_funded_wallet(&desc);
+    let to_spend = wallet.get_balance().total();
     let send_to = wallet.get_address(AddressIndex::New);
     let mut builder = wallet.build_tx();
-    builder.add_recipient(send_to.script_pubkey(), 10_000);
+    builder.drain_to(send_to.script_pubkey()).drain_wallet();
     let mut psbt = builder.finish().unwrap();
+    let unsigned_tx = psbt.unsigned_tx.clone();
+
     // Adds a signer for the wrong internal key, bdk should not use this key to sign
     wallet.add_signer(
         KeychainKind::External,
@@ -183,10 +193,32 @@ fn test_psbt_multiple_internalkey_signers() {
             },
         )),
     );
-    let _ = wallet.sign(&mut psbt, SignOptions::default()).unwrap();
-    // Checks that we signed using the right key
-    assert!(
-        psbt.finalize_mut(&secp).is_ok(),
-        "The wrong internal key was used"
-    );
+    let finalized = wallet.sign(&mut psbt, SignOptions::default()).unwrap();
+    assert!(finalized);
+
+    // To verify, we need the signature, message, and pubkey
+    let witness = psbt.inputs[0].final_script_witness.as_ref().unwrap();
+    assert!(!witness.is_empty());
+    let signature = schnorr::Signature::from_slice(witness.iter().next().unwrap()).unwrap();
+
+    // the prevout we're spending
+    let prevouts = &[TxOut {
+        script_pubkey: send_to.script_pubkey(),
+        value: to_spend,
+    }];
+    let prevouts = Prevouts::All(prevouts);
+    let input_index = 0;
+    let mut sighash_cache = SighashCache::new(unsigned_tx);
+    let sighash = sighash_cache
+        .taproot_key_spend_signature_hash(input_index, &prevouts, TapSighashType::Default)
+        .unwrap();
+    let message = Message::from(sighash);
+
+    // add tweak. this was taken from `signer::sign_psbt_schnorr`
+    let keypair = keypair.tap_tweak(&secp, None).to_inner();
+    let (xonlykey, _parity) = XOnlyPublicKey::from_keypair(&keypair);
+
+    // Must verify if we used the correct key to sign
+    let verify_res = secp.verify_schnorr(&signature, &message, &xonlykey);
+    assert!(verify_res.is_ok(), "The wrong internal key was used");
 }
--- a/crates/bdk/src/wallet/signer.rs
+++ b/crates/bdk/src/wallet/signer.rs
@@ -782,6 +782,16 @@ pub struct SignOptions {
     /// Defaults to `true` which will remove partial signatures during finalization.
     pub remove_partial_sigs: bool,
 
+    /// Whether to remove taproot specific fields from the PSBT on finalization.
+    ///
+    /// For inputs this includes the taproot internal key, merkle root, and individual
+    /// scripts and signatures. For both inputs and outputs it includes key origin info.
+    ///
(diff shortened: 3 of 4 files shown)


crates/bdk/tests/psbt.rs
```diff
@@ -183,7 +183,15 @@ fn test_psbt_multiple_internalkey_signers() {
...
+            &mut psbt,
+            SignOptions {
+                remove_taproot_extras: false,
+                ..Default::default()
+            }
+        )
+        .unwrap());
     // Checks that we signed using the right key
     assert!(
         psbt.finalize_mut(&secp).is_ok(),
```
ValuedMammal: @danielabrozzoni Curiously, removing `tap_key_sig` from a psbt input made this fail somewhere during `finalize_mut`, hence this change. But this shouldn't affect the meaning of the test, other than to assert finalization succeeded in `Wallet::sign`. Just wanted to run this by you because I think it was one of your tests.

danielabrozzoni: The reason for this is that `finalize_mut` doesn't really make sense here: `wallet.sign` is already finalizing the transaction (and you are asserting that, thanks for adding it!), and the only reason why `finalize_mut` wouldn't complain before is because we left the taproot_extras in the psbt, and so miniscript was able to re-finalize it. Can you please add a commit removing this assert with finalize_mut? Thanks!

ValuedMammal: Actually, I don't believe asserting that `sign` returns true is sufficient to check *which key* did the signing, whereas `finalize_mut` will take steps to validate sigs, if I understand. If we just have wallet not try to finalize, then this will make more sense.

danielabrozzoni: Your approach works, but the previous one worked too:
- `tap_signatures` is an Option, it is not a BTreeMap like `partial_sigs`
- The bug was that we would sign with whatever key we had in hand, thus replacing the previous signature that was in the PSBT
- In the wallet at the time of signing the key with higher precedence is the right one, so we are signing with the right one first, and the wrong one second. Without the bug fix, we would replace the right signature with the wrong one and we wouldn't be able to finalize. After the fix, we check that we're signing with the right key, and so adding a wrong key that signs last doesn't really matter (it is not used for signing anyways)
- If the tx is finalized it means that we had the right signature in place

Perhaps instead of rewriting the test, we could add comments better explaining what's going on?

ValuedMammal: Thank you. How about we take your original suggestion and get rid of `finalize_mut`. 

The only problem I see is that this test (below) passes on the pre #1200 buggy code, when it shouldn't?

<details>

<summary>expand</summary>

```rust
fn test_psbt_multiple_internalkey_signers() {
    use bdk::signer::{SignerContext, SignerOrdering, SignerWrapper};
    use bdk::KeychainKind;
    use bitcoin::{secp256k1::Secp256k1, PrivateKey};
    use miniscript::psbt::PsbtExt;
    use std::sync::Arc;

    let secp = Secp256k1::new();
    let (mut wallet, _) = get_funded_wallet(get_test_tr_single_sig());
    let send_to = wallet.get_address(AddressIndex::New);
    builder.add_recipient(send_to.script_pubkey(), 10_000);
    // Adds a signer for the wrong internal key, bdk should not use this key to sign
    wallet.add_signer(
        KeychainKind::External,
        // A signerordering lower than 100, bdk will use this signer first
        SignerOrdering(0),
        Arc::new(SignerWrapper::new(
            PrivateKey::from_wif("5J5PZqvCe1uThJ3FZeUUFLCh2FuK9pZhtEK4MzhNmugqTmxCdwE").unwrap(),
            SignerContext::Tap {
                is_internal_key: true,
            },
        )),
    );
    let finalized = wallet.sign(&mut psbt, SignOptions::default()).unwrap();
    assert!(finalized);
    
    // Checks that we signed using the right key
    let err = psbt.finalize_mut(&secp);
    dbg!(err);
}
```

</details>

so `finalize_mut` was doing its job, even though the wallet considers it finalized. So what made the tx unbroadcastable, I wonder? 

This is probably a bit out of scope for the current PR, but in the meantime I'll think of some other ways we can test the 'correct internal key' logic if you think that would be valuable.

danielabrozzoni: Oh, you're right, thanks! I think the problem lays in the fact that `finalized` is calculated based on "do we have the descriptor to satisfy this input? If that's so, and we can satisfy every input, then the transaction must be finalized". 

I am a bit torn here. The only solution that comes to my mind is to get rid of the test - we realized it wasn't even testing what I thought it was, and atm we don't have any function to check "did we sign with the right key?". 

@evanlinjin inputs?

danielabrozzoni: Edit: we can also comment the test for now instead of removing it, with a comment pointing to this discussion

ValuedMammal: Referring to this

https://github.com/bitcoindevkit/bdk/blob/7aca88474ac241bc0938ff989026463bdaf8f3ab/crates/bdk/src/wallet/signer.rs#L459-L467
 

>  The bug was that we would sign with whatever key we had in hand, thus replacing the previous signature that was in the PSBT
> 
>  In the wallet at the time of signing the key with higher precedence is the right one, so we are signing with the right one first, and the wrong one second.


My interpretation is that we sign with the wrong key first, since it has the lower SignerOrdering, and then skip signing for subsequent signers, as at that point `tap_key_sig` is no longer None.


> After the fix, we check that we're signing with the right key, and so adding a wrong key that signs last doesn't really matter (it is not used for signing anyways)

Agreed

> Edit: we can also comment the test for now instead of removing it, with a comment pointing to this discussion

I'm ok with this.

danielabrozzoni: > My interpretation is that we sign with the wrong key first, since it has the lower SignerOrdering, and then skip signing for subsequent signers, as at that point tap_key_sig is no longer None.

Yes, that's exactly what we would do

ValuedMammal:
I have a bit more clarity on things. The wallet considers a psbt finalized if it can produce a signature and fill in the witness or scriptsig, and I think that's enough for our purposes. However the resulting transaction can still be rejected by the network for other reasons (the coins were already spent, a timelock wasn't met, etc).

We still need to ensure the correct key is used for signing, and the most straightforward way is to verify the signature against the intended pubkey with `verify_schnorr`. This requires access to more of the internals, like computing the sighash, so the test for this can ~~possibly just go in `signer.rs`, where we simulate the behavior of the wallet by looping over different signers, calling `sign_input`,~~ and verifying the signature we end up with against the expected xonly key.

edit: I found a way to do it by fixing up the existing test.

danielabrozzoni:
This is fantastic, thanks!
Just a little nit, can you reorder the commits so that they're atomical and hygienic? (More info here: https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#committing-patches)

I'm thinking, instead of having 4 commits, just having two, one updating the test `test_psbt_multiple_internalkey_signers`, one removing the extra taproot fields when finalizing

ValuedMammal:
> I'm thinking, instead of having 4 commits, just having two, one updating the test `test_psbt_multiple_internalkey_signers`, one removing the extra taproot fields when finalizing

For sure, will squash

crates/bdk/src/wallet/mod.rs
```diff
@@ -1972,6 +1972,18 @@ impl<D> Wallet<D> {
                             if sign_options.remove_partial_sigs {
                                 psbt_input.partial_sigs.clear();
                             }
+                            if sign_options.remove_taproot_extras
+                                && psbt_input.tap_internal_key.is_some()
```
evanlinjin: Nit: I guess this is not strictly necessary. Also, it might be the case that the signer decides to clear some fields and not all.
```suggestion
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -1980,6 +1992,15 @@ impl<D> Wallet<D> {
             }
         }
 
+        if finished && sign_options.remove_taproot_extras {
+            for output in &mut psbt.outputs {
+                // Only clear derivations for taproot outputs.
+                if output.tap_internal_key.is_some() {
+                    output.bip32_derivation.clear();
+                }
```
evanlinjin:
```suggestion
                output.bip32_derivation.clear();
```

I think we should clear this for all finalized outputs.

evanlinjin (changes requested):
Thank you for doing this! I prefer if we remove taproot extra fields in all cases when witness/scriptsig is finalized (instead of just when the `tap_internal_key` is present). Otherwise, it looks good to me!

crates/bdk/src/wallet/mod.rs
ValuedMammal: The rationale was that without this condition, we would clear bip32 derivations even for non taproot spends. Maybe `bip32_derivation` doesn't belong in the list of "taproot extras"?

evanlinjin: I see. What is the rationale of clearing the `bip_derivation` field?

ValuedMammal: Ah, I'm thinking the taproot derivation info is already represented by `tap_key_origins`, meaning `bip32_derivation` isn't relevant here.

evanlinjin (approved):
ACK 5840ce473e430de4c4e3698734e9667cc476fee4

PR #1316, opened 2024-02-02
tx_builder: Support setting explicit nSequence for foreign inputs

### LINKED ISSUE #1315: No way to set nSequence value for foreign utxos
The TxBuilder API for foreign UTXOs does not support setting the nSequence value for those inputs in any way.

stevenroose:
Fixes https://github.com/bitcoindevkit/bdk/issues/1315.

### DIFF
--- a/crates/bdk/src/wallet/tx_builder.rs
+++ b/crates/bdk/src/wallet/tx_builder.rs
@@ -390,6 +390,22 @@ impl<'a, D, Cs: CoinSelectionAlgorithm, Ctx: TxBuilderContext> TxBuilder<'a, D, 
         psbt_input: psbt::Input,
         satisfaction_weight: usize,
     ) -> Result<&mut Self, AddForeignUtxoError> {
+        self.add_foreign_utxo_with_sequence(
+            outpoint,
+            psbt_input,
+            satisfaction_weight,
+            Sequence::MAX,
+        )
+    }
+
+    /// Same as [add_foreign_utxo](TxBuilder::add_foreign_utxo) but allows to set the nSequence value.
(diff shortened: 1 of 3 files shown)


crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -389,6 +389,22 @@ impl<'a, D, Cs: CoinSelectionAlgorithm, Ctx: TxBuilderContext> TxBuilder<'a, D,
...
+    ) -> Result<&mut Self, AddForeignUtxoError> {
+        self.add_foreign_utxo_with_sequence(
+            outpoint,
+            psbt_input,
+            satisfaction_weight,
+            Sequence::MAX,
+        )
+    }
+
+    /// Same as [add_foreign_utxo] but allows to set the nSequence value.
```
evanlinjin:
```suggestion
    /// Same as [add_foreign_utxo](TxBuilder::add_foreign_utxo) but allows to set the nSequence value.
```

stevenroose: fixed

evanlinjin (approved):
ACK 9bb39a3a3fcdebeec53a8a771091954c845db7a5

When we change to use PSBT v2, we can remove this (as the sequence will be part of the psbt input).

stevenroose:
Oh yeah that's true. I was looking for a sequence field in the input, but psbt v1 works on already-finish tx structure I guess. Not sure how the psbtv2 flow will work for this. I guess it has one more "creation" phase to go through.

danielabrozzoni (approved):
utACK 9bb39a3a3fcdebeec53a8a771091954c845db7a5

PR #1257, opened 2024-01-06
ci: move to Nix

storopoli:
### Description
This would make all the tests possible to run locally. It enhances developer experience and facilitates onboarding of new contributors.

(Updated MSRV to 1.63)

Additions:

* Crate auditing capabilities using [`rustsec/advisory-db`](https://github.com/rustsec/advisory-db).
* Local and CI caching of all artifacts from running `cargo {check,build,test}` in the whole workspace.
  Superseds `.github/workflows/audit.yml`.
* Pre-commit checks for everything that we enforce (gpg-signed/conventional commits, and some goodies like typos, so [docs: fix spelling errors #1086](https://github.com/bitcoindevkit/bdk/pull/1086) kind of things are not necessary anymore)
* Instructions and rationale in `NIX.md`

Closes bitcoindevkit/bdk_wallet#129. Superseds bitcoindevkit/bdk#1122, bitcoindevkit/bdk#1156, bitcoindevkit/bdk#1165.

Pinneds Dependencies:

* `bitcoind`: pinned to 0.25.0 in nixpkgs.
* `electrs`: pinned to [`Blockstream's esplora`](https://github.com/Blockstream/electrs) using the Fedimint deployment, check [jkitman/nixpkgs@61ccef8](https://github.com/jkitman/nixpkgs/commit/61ccef8bc0a010a21ccdeb10a92220a47d8149ac) and https://github.com/fedimint/fedimint/blob/master/flake.nix#L5

TODO:

* [x]  Fix `llvm` deps.
* [x]  Fix `openssl` errors.
* [x]  Fix dependencies on MacOS.
* [x]  Fix `WASM`.
* [x]  Fix `MSRV`.
  Add a custom `CargoMSRV.lock`? (Or maybe a `CargoMSRV.toml` with the pinned MSRV dependencies and then `cargo generate-lockfile` with Rust MSRV?)
  Depends on solving [fix: build `crane-utils` using a different toolchain ipetkov/crane#422](https://github.com/ipetkov/crane/issues/422)
* [x]  move all the logic from `cont_integration.yml` to `flake.nix` (all the `cargo update -p`)
  From the Fedimint suggestion we'll use [`crane`](https://crane.dev).
  This would allow caching of a lot of things
  Then the user would call `nix buid -L .#MSRV --keep-failed` and so on.
  This would also eliminate all the multiple `runs-on` in `cont_integration.yml` to a single one where each step would be a `name` and `run` the `nix build` command.
[...]

PR #1156, opened 2023-10-08
ci: replace toolchain

realeinherjar:
### Description

This replaces `actions-rs/toolchain` for `dtolnay/rust-toolchain` in all CI runs.

### Notes to the reviewers

Following the discussion in https://github.com/rust-bitcoin/rust-bitcoin/issues/2113,
[`actions-rs/toolchain`](https://github.com/actions-rs/toolchain) is an "Unofficial GitHub Actions for Rust programming language" which has only one person in the organization, @svartalf, which appears to be MIA for a long time.
People are asking in vain for these fixes, check https://github.com/actions-rs/toolchain/issues/221 (almost a year long).
 
There are a bunch of warnings in CI, e.g. https://github.com/bitcoindevkit/bdk/actions/runs/6443147276/job/17499339102#step:6:71.

[`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain) is by a well-established Rust developer, and it is well-maintaned (and has more stars than `actions-rs/toolchain`.

Also [`rust-bitcoin` already replaced `actions-rs/toolchain` with `dtolnay/rust-toolchain` in most every place](https://github.com/rust-bitcoin/rust-bitcoin/issues/2113#issuecomment-1751846698)

### Changelog notice

Replaced `actions-rs/toolchain` with `dtolnay/rust-toolchain` in CI.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #1122, opened 2023-09-12
ci: Check commits in PRs

realeinherjar:
Closes bitcoindevkit/bdk_wallet#135.
- Creates `commits.yml` in `.github/workflows/` directory with a new GH Action
  that has 2 jobs:

  1. `signed-commits`: uses [`1Password/check-signed-commits-action`(https://github.com/1Password/check-signed-commits-action)
     to check if all commits in the PR are signed.
  1. `conventional-commits`: uses [`cocogitto/cocogitto-action`](https://github.com/cocogitto/cocogitto-action)
      based on [`cocogitto`](https://crates.io/crates/cocogitto)
      to check if all commits in the PR follows the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.2/),
      i.e. they are prefixed with `fix`, `refactor`, `feat` etc.

- Clarifies the "Commit policy" `CONTRIBUTING.md`:
  - Rust-based CLI tool suggestion to locally check for the commits: [`cocogitto/cocogitto`](https://github.com/cocogitto/cocogitto).
  - Suggestion of two git hooks, see below.

- Two [Git Hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)
  to the new directory `ci/git-hooks`
  - `signed-commits.sh` hook (a `pre-push` hook)
     will not allow the user to push to remote if the last commit is not signed.
     This is a good sanity test,if the last commit is signed then there is a high chance
     of any other previous are also signed.
  - `conventional-commits.sh` hook (a `commit-msg` hook) will not allow the user to commit
     if the commit message does not satisfy the Conventional Commits template.

### Notes to the reviewers

Since this is a big enforcement change,
we can for now not make this CI check mandatory for a PR to be merged.
When appropriate we can turn check mandatory.

Also none of these actions have write permissions.

### Changelog notice

Added CI checks for signed commits and conventional commits guidelines.

#### All Submissions:

* [X] I've signed all my commits
* [X] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [X] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [ ] I've added tests for the new feature
* [X] I've added docs for the new feature (in `CONTRIBUTING.md`)

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [X] I'm linking the issue being fixed by this PR

PR #1320, opened 2024-02-04
ci: move to Nix (simpler, easy to maintain version)

### LINKED ISSUE #129: ci: workflows failing due to "Connection timed out" errors
**Describe the bug**  

CI workflows are randomly failing due to "Connection timed out" errors when downloading the `bitcoind` binary from "https://bitcoincore.org/bin". 

```
error: failed to run custom build command for `bitcoind v0.30.0`

Caused by:
  process didn't exit successfully: `/home/runner/work/bdk/bdk/target/debug/build/bitcoind-99a74cf3b9077890/build-script-build` (exit status: 101)
  --- stdout
  filename:bitcoin-22.0-x86_64-linux-gnu.tar.gz version:22.0 hash:59ebd25dd82a51638b7a6bb914586201e67db67b919b2a1ff08925a7936d1b16
  url:https://bitcoincore.org/bin//bitcoin-core-22.0/bitcoin-22.0-x86_64-linux-gnu.tar.gz

  --- stderr
  thread 'main' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bitcoind-0.30.0/build.rs:98:49:
  called `Result::unwrap()` on an `Err` value: IoError(Os { code: 110, kind: TimedOut, message: "Connection timed out" })
```

For example: https://github.com/bitcoindevkit/bdk/actions/runs/6459716661/job/17538503354?pr=1160

**To Reproduce**  

Push a new PR for `master` or `release/0.29` branch. 

**Expected behavior**  

There should be no random failures. 

**Build environment**  
 - BDK tag/commit: `master` and `release/0.29` branches
 - OS+version: linux or macos, likely all other
 - Rust/Cargo version: stable
 - Rust/Cargo target: linux or macos, likely all others

**Additional context**  

https://github.com/RCasatta/bitcoind/pull/144

### LINKED ISSUE #1162: ci: workflows are randomly failing due to "Connection timed out" errors
**Describe the bug**  

CI workflows are randomly failing due to "Connection timed out" errors when downloading the `bitcoind` binary from "https://bitcoincore.org/bin". 

```
error: failed to run custom build command for `bitcoind v0.30.0`

Caused by:
  process didn't exit successfully: `/home/runner/work/bdk/bdk/target/debug/build/bitcoind-99a74cf3b9077890/build-script-build` (exit status: 101)
  --- stdout
  filename:bitcoin-22.0-x86_64-linux-gnu.tar.gz version:22.0 hash:59ebd25dd82a51638b7a6bb914586201e67db67b919b2a1ff08925a7936d1b16
  url:https://bitcoincore.org/bin//bitcoin-core-22.0/bitcoin-22.0-x86_64-linux-gnu.tar.gz

  --- stderr
  thread 'main' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bitcoind-0.30.0/build.rs:98:49:
  called `Result::unwrap()` on an `Err` value: IoError(Os { code: 110, kind: TimedOut, message: "Connection timed out" })
```

For example: https://github.com/bitcoindevkit/bdk/actions/runs/6459716661/job/17538503354?pr=1160

**To Reproduce**  

Push a new PR for `master` or `release/0.29` branch. 

**Expected behavior**  

There should be no random failures. 

**Build environment**  
 - BDK tag/commit: `master` and `release/0.29` branches
 - OS+version: linux or macos, likely all other
 - Rust/Cargo version: stable
 - Rust/Cargo target: linux or macos, likely all others

**Additional context**  

https://github.com/RCasatta/bitcoind/pull/144

storopoli:
## Description

This is a simpler version of bitcoindevkit/bdk#1257.
Although not fully reproducible (we are fetching Cargo dependencies instead of declaring all of them).
It is WAY easier to maintain and provides most of the benefits and Dev-Experience from Nix. 

You can switch flawlessly from Rust stable, MSRV, and nightly using either:

- `nix develop .`: latest Rust stable
- `nix develop .#stable`: latest Rust stable
- `nix develop .#msrv`: Rust MRSV
- `nix develop .#nightly`: Rust Nightly

It handles all the annoyances of `Cargo.lock`s and `cargo update`s.

Also you don't need to worry about installing binaries from `bitcoind` or `electrs` crates,
since we fetch these from `nixpkgs` (with pinned versions) and add them to your local environment.
Additionally, all the extra dependencies that you need for WASM are handled as well.
Finally, no need to set ENV vars: they are also set all by default to you.

If you are committing typos, or failing to adhere to our guidelines in [CONTRIBUTING.md](CONTRIBUTING.md);
we won't let you commit or push.
This is handled by Nix's automated `pre-commit` hooks which gives helpful error messages.

Please check the [Nix.md](Nix.md) file to learn more about Nix and what we are aiming to do with the Nix integration into BDK.

I am taking the liberty of tagging certain people that are Nix enthusiast to help the discussion and review.
Cc @notmandatory, @dpc, @casey-bowman, @plebhash, @matthiasdebernardini, @gudnuf, @oleonardolima. 

### Notes to the reviewers

- Pre-commit checks for everything that we enforce:
  - Checks for typos in the source code and documentation.
  - Checks if all the commits are GPG-signed and follow the conventional commits style.
    (again, please check [CONTRIBUTING.md](CONTRIBUTING.md))
  - Runs `cargo clippy` in all workspace.
  - Runs `cargo fmt` in all workspace.
  - Runs `nixpgs-fmt` in all workspace (for `.nix` files)
- Instructions and rationale in `Nix.md`

Closes bitcoindevkit/bdk_wallet#129. Superseds bitcoindevkit/bdk#1122, bitcoindevkit/bdk#1156, bitcoindevkit/bdk#1165, bitcoindevkit/bdk#1257.

Pinneds Dependencies:

* `bitcoind`: pinned to 0.27.0 in `nixpkgs` version `24.05`.
* `electrs`: pinned to [`Blockstream's esplora`](https://github.com/Blockstream/electrs).

### Changelog notice

- We moved to Nix-based CI and Local development environment to better onboard new developers and enhance the productivity of developers.

#### All Submissions:

#### New Features:

#### Bugfixes:


dpc:
> This can be fixed with -- --test-threads=2. Shall we add?

Why not increase [bitcoind's rpc threads and work queue](https://github.com/fedimint/fedimint/blob/32bac957b23c0facfaff32baaa427d4898bce06d/devimint/src/cfg/bitcoin.conf#L9C10-L10C10) from the potato-like defaults instead?

dpc:
So just the dev shells, no crane etc.? Seems reasonable. Dev shells are like 80% of benefits for very little Nix involvment.

dpc (approved):

crates/bdk/Cargo.toml
```diff
@@ -13,6 +13,7 @@ edition = "2021"
 rust-version = "1.63"
 
 [dependencies]
+log = "^0.4"
```
notmandatory: We shouldn't need this dependency, see #1222 for a bit of discussion about it.

storopoli:
> > This can be fixed with -- --test-threads=2. Shall we add?
> 
> Why not increase [bitcoind's rpc threads and work queue](https://github.com/fedimint/fedimint/blob/32bac957b23c0facfaff32baaa427d4898bce06d/devimint/src/cfg/bitcoin.conf#L9C10-L10C10) from the potato-like defaults instead?

Good call I did this in a1997cf324aa5d5c70ce0eaa46a8c27f16d7c4a2

```rust
bitcoind_conf.args.push("-rpcworkqueue=1024");
bitcoind_conf.args.push("-rpcthreads=64");
```

crates/bdk/Cargo.toml
storopoli: Good call. I did forked out a very old state of the bdk repo
Removed now.

evanlinjin:
> > > This can be fixed with -- --test-threads=2. Shall we add?
> > 
> > 
> > Why not increase [bitcoind's rpc threads and work queue](https://github.com/fedimint/fedimint/blob/32bac957b23c0facfaff32baaa427d4898bce06d/devimint/src/cfg/bitcoin.conf#L9C10-L10C10) from the potato-like defaults instead?
> 
> Good call I did this in [a1997cf](https://github.com/bitcoindevkit/bdk/commit/a1997cf324aa5d5c70ce0eaa46a8c27f16d7c4a2)
> 
> ```rust
> bitcoind_conf.args.push("-rpcworkqueue=1024");
> bitcoind_conf.args.push("-rpcthreads=64");
> ```

64 threads is way too much. I think the ideal is to have 1 thread per CPU core. I think a good compromise is 4 (between desktop and CI).

storopoli:
> > > > This can be fixed with -- --test-threads=2. Shall we add?
> > > 
> > > 
> > > Why not increase [bitcoind's rpc threads and work queue](https://github.com/fedimint/fedimint/blob/32bac957b23c0facfaff32baaa427d4898bce06d/devimint/src/cfg/bitcoin.conf#L9C10-L10C10) from the potato-like defaults instead?
> > 
> > 
> > Good call I did this in [a1997cf](https://github.com/bitcoindevkit/bdk/commit/a1997cf324aa5d5c70ce0eaa46a8c27f16d7c4a2)
> > ```rust
> > bitcoind_conf.args.push("-rpcworkqueue=1024");
> > bitcoind_conf.args.push("-rpcthreads=64");
> > ```
> 
> 64 threads is way too much. I think the ideal is to have 1 thread per CPU core. I think a good compromise is 4 (between desktop and CI).

4 is the default. So no change then? I can easily delete a1997cf324aa5d5c70ce0eaa46a8c27f16d7c4a2 (that's why I did it in a separate commit 😄)

dpc:
> 64 threads is way too much. I think the ideal is to have 1 thread per CPU core. I think a good compromise is 4 (between desktop and CI).

Its a io bound workload. Absent async capabilities default 4 is laughable even for one core and I don't even know why bitcoind is doing it.

evanlinjin:
> > 64 threads is way too much. I think the ideal is to have 1 thread per CPU core. I think a good compromise is 4 (between desktop and CI).
> 
> Its a io bound workload. Absent async capabilities default 4 is laughable even for one core and I don't even know why bitcoind is doing it.

I didn't realize it was io-bounded. However, in our test environment, I doubt there will be 64 concurrent calls. How about we meet in the middle and say 16?

dpc:
I was on my phone before, so let me elaborate.

> I didn't realize it was io-bounded. However, in our test environment, I doubt there will be 64 concurrent calls. How about we meet in the middle and say 16?

From a perspective of most workloads e.g. asking for blockchain data, `bitcoind` (and `electrs`) acts very much like database software, e.g. posgresql. Postgresql is an [interesting case because it utilizes process-based model](https://tech.slashdot.org/story/23/06/19/1723259/postgresql-reconsiders-its-process-based-model) - meaning it will spawn not just a thread, but a whole process for every client connection - waaay heavier than just a thread.

Even for that heavy model, the recommendations will be [to limit it around around 500](https://cloud.ibm.com/docs/databases-for-postgresql?topic=databases-for-postgresql-managing-connections).

It takes many threads to saturate modern IO device. [Just an example from a yt video I had on hand benchamarking filesystems](https://youtu.be/AhjL1SHku7M?si=ZnSzjDUFpRvorQa5&t=241). Each bar is number of workers. And these are dedicated IO-generating workers, generating heavy load in a tight loop. So in this benchmark, it will take like 4 workers. Possibly one worker is even more threads. But an rpc thread in bitcoind by necessity will be spending lots of time reading things and writing stuff to a tcp connection, doing some checks, re-encodings and what not. To actually get a maximum IO performance there will need to be a lot of them. I'd say 64 per core.


On my mostly idle laptop at this very moment there are 2k threads running:

image

On a modern Linux a thread is just a stack, which due to CoW optimizations will really amount to 1 or 2 pages (4K page size) of physically used memory plus a tiny (1K?) amount to track it all in the kernel.

There's really not much to save on thread counts in a grand scheme of things.

Personally, I like to run as lots and lots of tests in parallel, especially any sort of integration and e2e tests, which tend to have lots of dead time - waiting for stuff (like disk and network IO - latency-bound things). The usual limit to this is the available memory. Additional benefit is also that it tends to expose problems more often, as all operations tend to have more ... timing jitter.

So I'm really surprised that the default in `bitcoind` is 4. That's too little even for an old generation raspberry pi.

CONTRIBUTING.md
```diff
@@ -31,8 +31,8 @@ facilitates social contribution, easy testing and peer review.
 To contribute a patch, the workflow is as follows:
 
   1. Fork Repository
-  2. Create topic branch
-  3. Commit patches
+  1. Create topic branch
+  1. Commit patches
```
oleonardolima, line 35: @storopoli nit: is this index right ?

CONTRIBUTING.md
```diff
@@ -48,15 +48,36 @@ hesitate to split it into multiple small, focused PRs.
...
+Commits should cover both the issue fixed and the solution's rationale.
+These [guidelines](https://chris.beams.io/posts/git-commit/) should be kept in mind.
+
+We enforce two commit styles in our CI:
+
+1. Commits should be signed with GPG using a key with a valid email address.
+1. Commit messages should follow the ["Conventional Commits 1.0.0"](https://www.conventionalcommits.org/en/v1.0.0/)
+   to make commit histories easier to read by humans and automated tools.
+   Commits starting with `Merge ...` which GitHub automatically generates
+   are exempt from this rule.
```
oleonardolima, line 85: @storopoli nit: same here

CONTRIBUTING.md
oleonardolima: there are also some other ones in the other .md files, but commenting here to not spam the PR

NIX.md
```diff
@@ -0,0 +1,138 @@
...
+
+- [NixOS Wiki](https://nixos.wiki/).
+- [Nix Reference Manual](https://nixos.org/manual/nix/stable/).
+- [`nixpkgs` Manual](https://nixos.org/manual/nixpkgs/stable/).
+- [Official documentation for getting things done with Nix](https://nix.dev/).
+
+There's also the [`nix-bitcoin` project](https://nixbitcoin.org/).
+
+To install Nix please follow the [official Nix documentation](https://nixos.org/manual/nix/stable/installation/installation.html).
+You may need to enable Flakes support, check the instructions at [NixOS Wiki entry on Flakes](https://nixos.wiki/wiki/Flakes).
```
oleonardolima: @storopoli wdyt about also mentioning the https://github.com/DeterminateSystems/nix-installer ?
it can be a helpful alternative way to install/remove it :)

NIX.md
```diff
@@ -0,0 +1,138 @@
...
+
+All the above can easily be accomplished with Nix.
+Nix is available for macOS, Linux, and Windows;
+while also being easy to install.
+Nix has a rich [community](https://nixos.org/community/) and one can find help
+on the [Nix's Forums](https://discourse.nixos.org/),
+[Nix's Discord](https://discord.gg/RbvHtGa),
+and [Nix's Matrix channel](https://matrix.to/#/#community:nixos.org).
+Additionally, Nix issues and errors can be searched in any search engine,
+and in the [Nix's stackoverflow](https://stackoverflow.com/questions/tagged/nix+or+nixpkgs+or+nixos+or+nixops).
```
oleonardolima: @storopoli just throwing some suggestions, maybe mentioning https://zero-to-nix.com/ as well ?

CONTRIBUTING.md
storopoli: It's good practice to use 1. indexing in `.md` files (see why [here](https://www.seancdavis.com/posts/pro-tip-use-1-for-every-ordered-list-item-when-writing-markdown/)).

Under the hood everything becomes

```html
<ol>
  <li>Create topic branch</li>
  <li>Commit patches</li>
</ol>
```

Then diffing and maintaining is a breeze.
Because if you move or remove item 2, you are screwed.
You need to rename all the other ones.
And the diff will suck.
With this you are fine.
Move, move, remove, `while true { list.shuffle() }`, you are good to go.

CONTRIBUTING.md
storopoli: Same as above

NIX.md
storopoli: Yes, despite being my preferred way, it is not the official installer.
Maybe a footnote what you think?

NIX.md
storopoli: I've added rebased and squashed. Thanks for the suggestion. Great resource.

CONTRIBUTING.md
oleonardolima: Cool, I didn't know that ;)
Thanks for the explanation and resource #TIL

NIX.md
oleonardolima: Sure, a footnote is good enough

NIX.md
oleonardolima: Great, thanks!

oleonardolima (approved):

NIX.md
matthiasdebernardini: https://github.com/cachix/devenv/issues/509#issuecomment-1919469287 @storopoli they are considering it

matthiasdebernardini:
Can you have it print at the beginning, "First time takes a good long while..." or something like that. Some parts seem so slow that you'd think the shell crashed or something like that. Would be good for the user to know this. 

Also I am having 8 tests fail, is that just me?
```
running 8 tests
test mempool_re_emits_if_tx_introduction_height_not_reached ... FAILED
test mempool_avoids_re_emission ... FAILED
test no_agreement_point ... FAILED
test mempool_during_reorg ... FAILED
test tx_can_become_unconfirmed_after_reorg ... FAILED
test test_sync_local_chain ... FAILED
test ensure_block_emitted_after_reorg_is_at_reorg_height ... FAILED
test test_into_tx_graph ... FAILED

failures:

---- mempool_re_emits_if_tx_introduction_height_not_reached stdout ----
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)

---- mempool_avoids_re_emission stdout ----
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)

---- no_agreement_point stdout ----
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)

---- mempool_during_reorg stdout ----
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)

---- tx_can_become_unconfirmed_after_reorg stdout ----
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)

---- test_sync_local_chain stdout ----
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)

---- ensure_block_emitted_after_reorg_is_at_reorg_height stdout ----
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)

---- test_into_tx_graph stdout ----
getting new addresses!
got new addresses!
mining block!
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)


failures:
    ensure_block_emitted_after_reorg_is_at_reorg_height
    mempool_avoids_re_emission
    mempool_during_reorg
    mempool_re_emits_if_tx_introduction_height_not_reached
    no_agreement_point
    test_into_tx_graph
    test_sync_local_chain
    tx_can_become_unconfirmed_after_reorg

test result: FAILED. 0 passed; 8 failed; 0 ignored; 0 measured; 0 filtered out; finished in 24.35s

error: test failed, to rerun pass `-p bdk_bitcoind_rpc --test test_emitter`
(nix:nix-shell-env) matthiasdebernardini@Matthiass-MacBook-Pro-2:~/Projects/bdk_nix$
```

storopoli:
> Also I am having 8 tests fail, is that just me?

Check the open questions: "I still get some random failures (at least locally) because of the tests DDoS'ing the poor electrs/bitcoind local binaries. This can be fixed with `-- --test-threads=2`. Shall we add?"

> Can you have it print at the beginning, "First time takes a good long while..." or something like that. Some parts seem so slow that you'd think the shell crashed or something like that. Would be good for the user to know this.

Does cargo install `big-huge-thing` prints something like that? I don't think it's necessary

NIX.md
storopoli: Ok, convinced, let's change the instructions...

flake.nix
```diff
@@ -0,0 +1,221 @@
...
+          name = "signed-commits";
+          runtimeInputs = [ pkgs.git ];
+          text = builtins.readFile ./ci/commits_verify_signature.sh;
+        };
+
+        # Toolchains
+        # latest stable
+        stable = pkgs.rust-bin.stable.latest.default.override {
+          targets = [ "wasm32-unknown-unknown" ]; # wasm
+        };
```
yanganto, line 71: The Rust version 1.63.0 is the minimum supported in msrv. It is good to set the default shell with this version, else developer will easily write code that can not pass the clippy.

flake.nix
```diff
@@ -0,0 +1,221 @@
...
+            _shellHook = (self.checks.${system}.pre-commit-check.shellHook or "");
+          in
+          {
+            default = pkgs.mkShell ({
+              shellHook = "${_shellHook}";
+              buildInputs = buildInputs ++ WASMInputs ++ [ stable ];
+              inherit nativeBuildInputs;
+            } // envVars // {
+              shellHook = "${cargoUpdate} ${_shellHook}";
+            });
```
yanganto: The Rust version 1.63.0 is the minimum supported in msrv. It is good to set the default shell with this version, else developer will easily write code that can not pass the clippy because it is not always the same between different Rust versions.

storopoli: Good point. @notmandatory what do you think? Shall we make MSRV the default devshell?

flake.nix
```diff
@@ -0,0 +1,221 @@
...
+  inputs = {
+    # stable nixpkgs (let's not YOLO on unstable)
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
+
+    # pin dependencies to a specific version
+    # find instructions here:
+    # <https://lazamar.co.uk/nix-versions>
+
+    # bitcoind pinned to 0.25.1
+    nixpkgs-bitcoind.url = "github:nixos/nixpkgs?rev=53793ca0aecf67164c630ca34dbfde552a8ab085";
```
yanganto: There are always backports on bitcoin `25.X`.  The latest one was three weeks ago.  The current way to pin bitcoin `0.25.1` is not updated because it is based on the date `2023-10-19`.  And pin the all `nixos/nixpkgs` is pretty heavy.

May we pin bitcoind with a correct version with hash but not `nixos/nixpkgs`?

flake.nix
```diff
@@ -0,0 +1,221 @@
...
+    # TODO: pin to 0.26.0 once it stops to fail in darwin
+    # pinned to 0.26.0
+    # nixpkgs-bitcoind.url = "github:nixos/nixpkgs?rev=f5375ec98618347da6b036a5e06381ab7380db03";
+
+    # Blockstream's esplora
+    # inspired by fedimint CI
+    nixpkgs-kitman.url = "github:jkitman/nixpkgs?rev=61ccef8bc0a010a21ccdeb10a92220a47d8149ac";
```
yanganto: There maybe same issue as bitcoind.

flake.nix
storopoli: > May we pin bitcoind with a correct version with hash but not nixos/nixpkgs?

What do you mean by that?
Downloading the binary straight from `https://bitcoin.org/`?

yanganto: No, downloading bitcoind will not be a proper way in nix shell, and the binary will not always work for everyone.

I open a PR to this feature branch.
https://github.com/storopoli/bdk/pull/1

yanganto: @evanlinjin If it is good to put it into another issue and keep this PR as simple as possible.  I am happy to be assigned to the other one.

storopoli: Oh noice! Yes, a full derivation. Now I get the idea!

flake.nix
storopoli: working on https://github.com/storopoli/bdk/pull/2

nondiremanuel:
@storopoli does this PR substitute bitcoindevkit/bdk#1257? Shall we close that or do we need to leave it open?

storopoli:
> @storopoli does this PR substitute bitcoindevkit/bdk#1257? Shall we close that or do we need to leave it open?

We are close but I need some help with https://github.com/storopoli/bdk/pull/2

flake.nix
storopoli: Done in [96c1758](https://github.com/bitcoindevkit/bdk/pull/1320/commits/96c17585ec9f6aedcbf2efc8013619cd430fe7b4)

flake.nix
notmandatory: It should be fine to use the current stable Rust for the devshell. We use a project wide `clippy.toml` with `msrv="1.63.0"` so what ever version of rust you're using it will only check against the 1.63.0 rules.

storopoli:
> @storopoli does this PR substitute bitcoindevkit/bdk#1257? Shall we close that or do we need to leave it open?

This is ready to review

evanlinjin:
> @storopoli does this PR substitute bitcoindevkit/bdk#1257? Shall we close that or do we need to leave it open?

Yes it does. Closing bitcoindevkit/bdk#1257

Edit:

> This is a simpler version of bitcoindevkit/bdk#1257.
Although not fully reproducible (we are fetching Cargo dependencies instead of declaring all of them).

Okay maybe not. It's just an alternative?

yanganto:
Hey, This is good, just set the msrv as default and lock your project, then all ci can pass.
https://github.com/storopoli/bdk/pull/3

Also, there is more than one way to sign a commit, not only GPG can sign.
I use the ssh sign,but the CI seems only work with GPG sign.

yanganto:
I have misunderstood that this pr will be droped, so I try to fix it without comment first.🙏 Looking forward to use nix here.😁

storopoli:
@yanganto amazing work. Thanks for the contribution.
Ok this is ready to be merged.

yanganto (approved):

oleonardolima (approved):
utACK 8e9288a358070c965107bab394b5d5b91a04386c

yanganto:
@notmandatory How do you think about this PR?  I love this one, and it will be good for me because I use NixOS, I always need a Nix shell when developing something.  It is also good for any newcomer to the project, he does not need to worry about the environment after this PR.

storopoli:
I've squashed everything into a single commit so that we're ready to merge.

oleonardolima (approved):
tACK

> **NOTE:** I wasn't able to run it on the Nix installed natively on macOS (as expected I guess), but I was able to test it over OrbStack.

storopoli:
> **NOTE:** I wasn't able to run it on the Nix installed natively on macOS (as expected I guess)

Not expected, I am able to run natively in MacOS. Can you post a stacktrace?

oleonardolima:
> > **NOTE:** I wasn't able to run it on the Nix installed natively on macOS (as expected I guess)
> 
> Not expected, I am able to run natively in MacOS. Can you post a stacktrace?

Sure, the specific error is related to CC linker 🤔 

I'll send you the detailed stacktrace.

<details>
<summary> cc linker failure - stacktrace </summary>

```
error: linking with `cc` failed: exit status: 1
  |
[...]
  = note: ld: framework not found SystemConfiguration
          clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
```
</details>

yanganto:
> > > **NOTE:** I wasn't able to run it on the Nix installed natively on macOS (as expected I guess)
> > 
> > Not expected, I am able to run natively in MacOS. Can you post a stacktrace?
> 
> Sure, the specific error is related to CC linker 🤔 
> 
> I'll send you the detailed stacktrace.
> 
> <details>
> <summary> cc linker failure - stacktrace </summary>
> 
> ```
> error: linking with `cc` failed: exit status: 1
>   |
> [...]
>   = note: ld: framework not found SystemConfiguration
>           clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
> ```
> </details>

What package is this happened when nix try to build?  I do not have a mac machine but I am happy to try fix It if there is much more context from your end.

oleonardolima:
>> [...]
> 
> What package is this happened when nix try to build? I do not have a mac machine but I am happy to try fix It if there is much more context from your end.

It happens when I try to run some command like: `cargo test` or `cargo clippy`, it's not really deterministic on which package it fails during the compilation phase.

yanganto:
Hi @oleonardolima 
Do you go into the nix shell?

You need do 
1. `nix develop` 
(After this you will be in the shell. It will be a long run when the first time go in, and it will build up everything you need with a determined version.  The second time you go into the shell, it will be quick)
2. other commands you want to do will in the nix shell, ex `cargo fmt`

Or your can do it one liner
`nix develop -c cargo fmt`

The error from your end is about `clang-16`, and I see the dev shell setup clang with a determined version 14, and set up in the shell here.
https://github.com/storopoli/bdk/blob/7c81fbd617ad9200f02ca3cda179405156568f04/flake.nix#L76

So if you are not in the nix shell, nix will not setup the environment for you.

oleonardolima:
> Hi @oleonardolima Do you go into the nix shell?
> 
> You need do
> 
> 1. `nix develop`
>    (After this you will be in the shell. It will be a long run when the first time go in, and it will build up everything you need with a determined version.  The second time you go into the shell, it will be quick)
> 2. other commands you want to do will in the nix shell, ex `cargo fmt`
> 
> Or your can do it one liner `nix develop -c cargo fmt`
> 
> The error from your end is about `clang-16`, and I see the dev shell setup clang with a determined version 14, and set up in the shell here. https://github.com/storopoli/bdk/blob/7c81fbd617ad9200f02ca3cda179405156568f04/flake.nix#L76
> 
> So if you are not in the nix shell, nix will not setup the environment for you.

Yes, I'm running it inside the nix shell.

It's weird because for the msrv shell it works, I only get this error on stable shell.

yanganto:
> Yes, I'm running it inside the nix shell.
> 
> It's weird because for the msrv shell it works, I only get this error on the stable shell.

There are two things need to know.  The clippy or fmt will be different based on the different versions of Rust.

### The first thing is to use msrv, if we design to support it.

The older one will be good, and much more restrictive, so a code that passes the older one is much more passed to the new one, the trade-off is we can not use the new syntax.

When we say we support MSRV, we will not want to pop a lot of errors or warns when a user uses the MSRV version of the crate we provided and do clippy or fmt.  If so, we should use msrv.

### The other thing is the stable version is not a live thing in a determined build system

It only means the latest stable version when you lock it based on the time, namely based on the `flake.lock`.  So it is still a fixed version of Rust.  You can double check if the what is rust version you running locally if the version of rust is not the same as fixed with `nix develop .#stable`.  If not, you are not in the nix shell for stable rust and just fall back to your local system.
The thing we call it stable because we will update schedulely in CI in the future.

---

BTW, @notmandatory, in bitcoindevkit/bdk#1366, the `.github/workflows/cont_integration.yml` is changed the version to stable. it will run based on the time.  There will be a different meaning of the stable version here, and it will not be reproducible in the future.  And it will stand on the opposite side of this PR.  If you want to do that, you can reject this PR, and I will not take time on the PR after you clearly reject this.
```diff
-          - version: 1.65.0 # STABLE
+         - version: stable
```

yanganto:
Sorry, @notmandatory 
I misunderstood that.  bitcoindevkit/bdk#1366 is doing on the release branch not a main branch. 
You did not mean to reject this PR.
:pray:

yanganto:
@matthiasdebernardini 
> Can you have it print at the beginning, "First time takes a good long while..." or something like that. Some parts seem so slow that you'd think the shell crashed or something like that. Would be good for the user to know this.

We can not do this print.  It will run when the user uses `nix develop` or other automatic daemons when the user goes into the folder.  We can only print after the environment is ready. We can provide a cache on the environment in the future, so the time will based on the downloading not on the compiling.

yanganto:
Hi @ValuedMammal 

If you have time, please take a look at this. The current ci failure is a 429 error, which already opened in bitcoindevkit/bdk#1120, and nothing about this PR. I just wanted to let you know that everything is done.

- current PR provided a reproducible environment (every dependency and version of the software in the system are locked) for developing, CI
- this will be the first step to make things robust, I believe you heard something only works on someone's machine, build server, or CI server in a lot of project.  Nix is a mature solution to solve this.

By the way, nix solution also provides a nix shell, that you can work in the same environment as CI server, build machine, or other developer's laptop.  This will greatly reduce the cost of team communication.

People love Rust, because it is more robust, but Rust only takes care of the application layer, because it still relies on things in the system, i.g. `openssl-sys`.  These kinds of issues only be solved by Nix.

You possibly wonder or are aware we do not know the nix shell, but actually, we also do not know the bash shell, cat, echo, sed, ..., etc., also Linux kernel.  We just use the shell and building application on top of it, nix just exposes these and lets us know, and uses a hash lock on everything we are using and produces a reproducible environment. The thing we don't know in detail is still unknown, but it is locked and becomes reproducible.


We use nix this way.
`nix develop -c command ...`
We still can easily take down nix back to a normal shell in the local environment.
`command ...`

I believe introducing nix in the project before it goes too big will be helpful.  And I hope to get your agreement on this PR.

storopoli:
@yanganto any idea why the `pre-commit-nix` `clippy` hook is complaining now?

```bash
last 25 log lines:
       > clippy...................................................................Failed
       > - hook id: clippy
       > - exit code: 101
       >
       >     Updating crates.io index
       > warning: spurious network error (3 tries remaining): [35] SSL connect error (OpenSSL/3.0.13: error:16000069:STORE routines::unregistered scheme)
       > warning: spurious network error (2 tries remaining): [35] SSL connect error (OpenSSL/3.0.13: error:16000069:STORE routines::unregistered scheme)
       > warning: spurious network error (1 tries remaining): [35] SSL connect error (OpenSSL/3.0.13: error:16000069:STORE routines::unregistered scheme)
       > error: failed to get `bip39` as a dependency of package `bdk v1.0.0-alpha.7 (/private/tmp/nix-build-pre-commit-run.drv-0/src/crates/bdk)`
       >
       > Caused by:
       >   failed to query replaced source registry `crates-io`
       >
       > Caused by:
       >   download of config.json failed
       >
       > Caused by:
       >   failed to download from `https://index.crates.io/config.json`
       >
       > Caused by:
       >   [35] SSL connect error (OpenSSL/3.0.13: error:16000069:STORE routines::unregistered scheme)
       >
       > nixpkgs-fmt..............................................................Passed
       > rustfmt..................................................................Passed
       > typos....................................................................Passed
       For full logs, run 'nix log /nix/store/qk9jkq6w1g91wsrzg2dkvq32pn1j9x2b-pre-commit-run.drv'.
```

Offending config https://github.com/storopoli/bdk/blob/936dc781a3a8356be2b4361a385d08f56ff4b7bf/flake.nix#L139-L142:

```nix
clippy = {
  enable = true;
  entry = lib.mkForce "${rust}/bin/cargo-clippy clippy --all-targets --all-features -- -D warnings";
};
```

yanganto:
If I were you I would remove the clippy and not do it in the check.  There is no network access in the check by design because network access is a side effect.  The check is better for static things otherwise, you need to make sure you have have cache in the nix build system so it will not network access when checked.

Rust clippy is not a simple static thing if you know Rust deeper, it will compile, so just do a thing under `nix develop -c ...` not in the check.  If you want to hack on this, do it in another PR not now, please :pray:.  I supper want this PR.  It is not worth it to hack a no-network static environment and want to put something before it against the design of the check (I know we can do and I have done it before).

Also, I do not think `pre-commit-check` is a good thing.  There is nothing hurt in letting the CI server run for others, people may want to contribute a little thing (like typo) from a weak laptop, I believe this contribution is welcome but we put a lot of things before his commit.  This is not kind and it is exactly mean in my view.

If you accept my point then forget someone's best practice and remove `pre-commit-check`, use a simple and stupid way(KISS), in the ci machine setup with nix, then go into the nix shell and do `cargo clippy`.

storopoli:
> in the ci machine setup with nix, then go into the nix shell and do `cargo clippy`.

Yes that was my intended workaround as well, see 582638586e6b9bac4c259e84ab7dddd29d8b984f.
Thanks!

crates/testenv/src/lib.rs
```diff
@@ -20,12 +20,12 @@ pub struct TestEnv {
...
     /// Construct a new [`TestEnv`] instance with default configurations.
     pub fn new() -> anyhow::Result<Self> {
-        let bitcoind = match std::env::var_os("BITCOIND_EXE") {
+        let bitcoind = match std::env::var_os("BITCOIND_EXEC") {
             Some(bitcoind_path) => electrsd::bitcoind::BitcoinD::new(bitcoind_path),
             None => {
                 let bitcoind_exe = electrsd::bitcoind::downloaded_exe_path()
                     .expect(
-                "you need to provide an env var BITCOIND_EXE or specify a bitcoind version feature",
+                "you need to provide an env var BITCOIND_EXEC or specify a bitcoind version feature",
```
storopoli: These were driving me crazy. It was not trivial to find.
These were wrong according to the `bitcoind`/`electrsd` docs.
Glad that I was able to fix it.

Cc @LagginTimes.

storopoli:
This rebasing is turning out to be not trivial.
@yanganto can you take a look at the CI errors when you have the time, please?

yanganto:
You can see the error
```
   Downloaded js-sys v0.3.68
error: package `powerfmt v0.2.0` cannot be built because it requires rustc 1.67.0 or newer, while the currently active rustc version is 1.63.0
```

---

Hi @storopoli 
What is your base?  It seems not rebased on main branch.
You can see the dependency in the main is 
https://github.com/bitcoindevkit/bdk/blob/80e190b3e734a46ba73925d381fe4f1a9941bf2a/crates/esplora/Cargo.toml#L26
But now your feature branch is 
https://github.com/storopoli/bdk/blob/488cdc3b8004b0c9b77ce160675dfcddbfcf7c70/crates/esplora/Cargo.toml#L26
```rust
electrsd = { version = "0.26.0", features = ["legacy"] }
```
And this possibly needs 1.68 as MSRV, or make an optional feature in electrsd and avoid that.  (base on `electrsd 0.26` -> `zip 0.6` -> `time 0.3.7`) 

Currently, time msrv is `1.68`, but I think it is not worthy to dig here, because the branch is rebased to something wrong place. 


---

Hi @ValuedMammal @evanlinjin 
Rust crate is possibly dependent on the Rust version.  If we do not use msrv to develop, it will possibly upgrade the dependency, not suitable for msrv and we don't know. (Dev will always say it should work but actually, it does not)  That is also the reason why we need to fix the environment and nix help on this.
 
This is what I did a similar task before https://github.com/storopoli/bdk/pull/3 for the PR.  If this is still in a not merged PR, then a similar thing will keep happening.  So that is why the nix needs to be introduced in the early stage of a project.

storopoli:
> What is your base? It seems not rebased on main branch. You can see the dependency in the main is
> 
> https://github.com/bitcoindevkit/bdk/blob/80e190b3e734a46ba73925d381fe4f1a9941bf2a/crates/esplora/Cargo.toml#L26
> 
> 
> But now your feature branch is
> https://github.com/storopoli/bdk/blob/488cdc3b8004b0c9b77ce160675dfcddbfcf7c70/crates/esplora/Cargo.toml#L26
> ```rust
> electrsd = { version = "0.26.0", features = ["legacy"] }
> ```

@yanganto this is the exact change that we we're proposing with the Nix CI.
If you 

See the `esplora/Cargo.toml` from previous iterations before the rebasing attempt.
This is an example from your last PR to fix devshells (https://github.com/yanganto/bdk/blob/361c81bfaf988c65df24394aa349efb22bf843c7/crates/esplora/Cargo.toml#L25):

```toml
electrsd = { version = "0.26.0", features = ["legacy"] }
```

If we include `"bitcoind_25_0", "esplora_a33e97e1"` they will trigger downloads.
This was working before.

I 03680515d62c5650f731ae376d9ff89a560e251b and 864f78c4679daf0e71c8a4a80289ee3c04c8da95 fixes that in the `testenv/Cargo.toml` (and some other things) which is the major new moving part in this rebasing round.

I tried to fix the clippy warnings in 9373abbebb0798658bc657413dcc665653798b5d but now I am getting this [error](https://github.com/bitcoindevkit/bdk/actions/runs/8434622007/job/23098216883#step:6:981):

```shell
error[E0786]: found invalid metadata files for crate `bitcoin`
   --> crates/chain/src/spk_iter.rs:202:20
    |
202 |         let secp = bitcoin::secp256k1::Secp256k1::signing_only();
    |                    ^^^^^^^
    |
    = note: invalid metadata version found: /home/runner/work/bdk/bdk/target/debug/deps/libbitcoin-9e061ce2009267ea.rmeta
```

I cannot reproduce this locally with `nix develop -L .#msrv --command cargo clippy --all-targets --all-features -- -D warnings`

yanganto:
> If we include `"bitcoind_25_0", "esplora_a33e97e1"` they will trigger downloads. This was working before.
Okay, I got it. Sorry about this.

> 
> ```shell
> error[E0786]: found invalid metadata files for crate `bitcoin`
>    --> crates/chain/src/spk_iter.rs:202:20
>     |
> 202 |         let secp = bitcoin::secp256k1::Secp256k1::signing_only();
>     |                    ^^^^^^^
>     |
>     = note: invalid metadata version found: /home/runner/work/bdk/bdk/target/debug/deps/libbitcoin-9e061ce2009267ea.rmeta
> ```
> 
> I cannot reproduce this locally with `nix develop -L .#msrv --command cargo clippy --all-targets --all-features -- -D warnings`

The error is 
```
error: could not compile `bdk_chain` due to 34 previous errors
```
The bdk crate can not be built when using clippy.
Your local crate is built before, so no fresh build again.  
The error is under bdk crate nothing about nix.  
When doing a fresh machine with msrv Rust, we will get this just like we see in CI.

Forgot the crate bug, I will just do clippy after the build, leave the hidden bug still in the crate, and focus on the ci itself here.  Also, I found the pre-commit-check does not use msrv.
https://github.com/storopoli/bdk/pull/4

ValuedMammal:
>     * current PR provided a reproducible environment (every dependency and version of the software in the system are locked) for developing, CI
> 
>     * this will be the first step to make things robust, I believe you heard something only works on someone's machine, build server, or CI server in a lot of project.  Nix is a mature solution to solve this.

@yanganto Thank you for the insightful comments, and especially for the work you put into this. True I haven't given this the review time it deserves, though I'm willing to be open minded when it comes to improving the build system.

storopoli:
Alright, this is now once again ready to merge to `master`.
Since this is on the `beta` milestone, we'll wait for the `alpha` PRs to land.
I don't see any more painful `Cargo.toml`s changes like the ones that occurred in `master` in the future.
So rebasing will probably be easier now.

Again, thanks @yanganto for the amazing work.

yanganto (approved):

storopoli:
Another painful rebase 😓.
But this is again ready to merge once we get to `beta` milestones.

README.md
```diff
@@ -62,16 +64,22 @@ Fully working examples of how to use these components are in `/example-crates`:
...
 This library should compile with any combination of features with Rust 1.63.0.
 
 To build with the MSRV you will need to pin dependencies as follows:
 
 ```shell
-cargo update -p zstd-sys --precise "2.0.8+zstd.1.5.5"
-cargo update -p time --precise "0.3.20"
 cargo update -p home --precise "0.5.5"
 cargo update -p proptest --precise "1.2.0"
+cargo update -p bumpalo --precise "3.14.0"
```
storopoli: Another rebase all green now.
@notmandatory check the new `bumpalo` CI pin.
We might need to fix that in the near future (or maybe now) ir regular CI.

notmandatory: Looks like this is just a dependency for WASM and it should be OK to only run those tests with `stable` rust.

storopoli:
rebased and ready for another round of reviews.
CI is fixed by bitcoindevkit/bdk#1529 and bitcoindevkit/bdk#1530.

oleonardolima:
@storopoli I'm still unsure why, but I'm getting `cc` linking errors on macos:

- nix develop .: Rust MSRV
I'm able to start the shell, but the `cargo test` command fails due to `cc` linking issues
- nix develop .#msrv: Rust MRSV
- nix develop .#stable: latest Rust stable
- nix develop .#nightly: Rust Nightly
On both the three above, I'm not able to start the shell, it fails while trying to build the electrs-deps
``` bash
       > error[E0425]: cannot find function `download_filename` in this scope
       >   --> /nix/store/400y52h1b4c939vx0zf9l79wd6krdf0j-vendor-cargo-deps/d1c72e7ae5a19faff71fdfa4daedf2b91f137ca4d842bac976a2cfa82c727fad/electrumd-0.1.0/build.rs:27:29
       >    |
       > 27 |     let download_filename = download_filename();
       >    |                             ^^^^^^^^^^^^^^^^^ not found in this scope
       >
```

storopoli:
I'll try to debug it.

storopoli:
Yeah we're screwed!
Maybe we need to drop Blockstream's esplora `electrs` and use the `electrs` crate directly.

Because somehow `electrumd` is being compiled (despite being a `dev-dependency`)
in [`Blockstream/electrs` `Cargo.toml`](https://github.com/Blockstream/electrs/blob/e11cd561a197f8c0e4839721b4e1877b2a41053f/Cargo.toml#L63):

```rs
electrumd = { version = "0.1.0", features = [ "4_1_5" ] }
```

and [`electrumd` hardcodes an AppImage in the `build.rs`](https://github.com/shesek/electrumd/blob/c42fc5f3394e21da3893294fb5112de1d884e389/build.rs#L10-L14) (AMAZING!):

```rs
#[cfg(all(target_os = "linux", target_arch = "x86_64"))]
fn download_filename() -> String {
    format!("electrum-{}-x86_64.AppImage", &VERSION)
}
// other platforms are currently unsupported
```

I am trying to remove blockstream's esplora and add `pkgs.electrsd` from `nixpkgs`.

Any ideas @dpc?

storopoli:
Ok 2301e80a3ea6ecdc9961190d11d80b4643b4fc56 should fix MacOS as well.
Copied from fedimint how they package Blockstream's `esplora`.
@oleonardolima and @matthiasdebernardini this should work on darwin arch.
Can you try it please? Run:

```bash
nix develop -L .#msrv --command cargo test --all-features
```

storopoli:
added a `justfile` and `just` to `buildInputs`

matthiasdebernardini:
I was finally able to get all the tests to pass using this command

`nix develop -L .#msrv --command cargo test --all-features -- --test-threads=2`

is there anything else to look after?

storopoli:
> is there anything else to look after?

Try without the `--test-threads=2` please?

matthiasdebernardini:
Could not get it to work, which is why I added the test threads argument.

```
➜  bdk git:(js/nix-simple) nix develop -L .#msrv --command cargo test --all-features
warning: input 'rust-overlay' has an override for a non-existent input 'flake-utils'
    Finished test [unoptimized + debuginfo] target(s) in 0.40s
warning: the following packages contain code that will be rejected by a future version of Rust: libc v0.2.155
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 7`
     Running unittests src/lib.rs (target/debug/deps/bdk_bitcoind_rpc-7626c46bad197f9a)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/test_emitter.rs (target/debug/deps/test_emitter-dc28ba34d3b67736)

running 8 tests
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)
test mempool_during_reorg ... FAILED
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)
test mempool_avoids_re_emission ... FAILED
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)
test no_agreement_point ... FAILED
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)
test mempool_re_emits_if_tx_introduction_height_not_reached ... FAILED
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)
test test_sync_local_chain ... FAILED
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)
test tx_can_become_unconfirmed_after_reorg ... FAILED
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)
test ensure_block_emitted_after_reorg_is_at_reorg_height ... FAILED
Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)
test test_into_tx_graph ... FAILED

failures:

---- mempool_during_reorg stdout ----
thread 'mempool_during_reorg' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/test/src/lib.rs:184:5

---- mempool_avoids_re_emission stdout ----
thread 'mempool_avoids_re_emission' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/test/src/lib.rs:184:5

---- no_agreement_point stdout ----
thread 'no_agreement_point' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/test/src/lib.rs:184:5

---- mempool_re_emits_if_tx_introduction_height_not_reached stdout ----
thread 'mempool_re_emits_if_tx_introduction_height_not_reached' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/test/src/lib.rs:184:5

---- test_sync_local_chain stdout ----
thread 'test_sync_local_chain' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/test/src/lib.rs:184:5

---- tx_can_become_unconfirmed_after_reorg stdout ----
thread 'tx_can_become_unconfirmed_after_reorg' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/test/src/lib.rs:184:5

---- ensure_block_emitted_after_reorg_is_at_reorg_height stdout ----
thread 'ensure_block_emitted_after_reorg_is_at_reorg_height' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/test/src/lib.rs:184:5

---- test_into_tx_graph stdout ----
getting new addresses!
got new addresses!
mining block!
thread 'test_into_tx_graph' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/test/src/lib.rs:184:5


failures:
    ensure_block_emitted_after_reorg_is_at_reorg_height
    mempool_avoids_re_emission
    mempool_during_reorg
    mempool_re_emits_if_tx_introduction_height_not_reached
    no_agreement_point
    test_into_tx_graph
    test_sync_local_chain
    tx_can_become_unconfirmed_after_reorg

test result: FAILED. 0 passed; 8 failed; 0 ignored; 0 measured; 0 filtered out; finished in 37.59s

error: test failed, to rerun pass '-p bdk_bitcoind_rpc --test test_emitter'
```

storopoli:
Ok, that's strange. Maybe a MacOS thing. I have 12 cores in my laptop and 32 in my desktop andi don't get these errors.

dpc:
> Error: JSON-RPC error: transport error: Couldn't connect to host: Resource temporarily unavailable (os error 35)


Possibly too low `ulimit`. See https://github.com/fedimint/fedimint/blob/d7135bce5bda0e623c84d89736d363961df7663e/flake.nix#L314

matthiasdebernardini:
I'm on the latest macOS with a high core count macbook.

I ran this command
```
➜  bdk git:(js/nix-simple) ulimit -Sn 1024
```

and got the same error

oleonardolima:
I tried it on my side too and got the same issue now.
I would guess it's how the tests are structured.

dpc:
Try setting the ulimit to higher numbers. 1024 is still quite small for a lot of things running concurrently.

dpc:
Also, if you are using single bitcoind instance for all tests, you might want to increase the queue size and rpc thread count on it. Otherwise bitcoind randomly rejects connections, even when there isn't all that many of them yet. Though to that I'd add that the production code should probably do some retries as well.

Anyway, I'm guessing. Unfortunately non trivial tests that involve lots of different software pieces etc. often require some debugging and refactoring/tweaking to work reliably.

https://github.com/ordinals/ord/issues/3364#issuecomment-2111172271


https://github.com/fedimint/fedimint/blob/1ca5227e0ceb8c4f3b0ca122364653e0f017a38c/devimint/src/cfg/bitcoin.conf#L9

oleonardolima:
> Try setting the ulimit to higher numbers. 1024 is still quite small for a lot of things running concurrently.

I tested with `ulimit` size up to 16384, and it still fails. I would guess it's something other than the `ulimit`.

> Also, if you are using single bitcoind instance for all tests, you might want to increase the queue size and rpc thread count on it. Otherwise bitcoind randomly rejects connections, even when there isn't all that many of them yet. Though to that I'd add that the production code should probably do some retries as well.

I did a little bit of it and further testing, the main problem occurs in the `bdk_testenv::TestEnv::mine_blocks`, it's the same lineno failure for all the tests. Specifically when it's trying to first mine some blocks, and calling the RPC command `generatetoaddress`.

It's really weird, because I checked it and it's running the bitcoind on different ports, datadirs, and such 🤔. 

> https://github.com/fedimint/fedimint/blob/1ca5227e0ceb8c4f3b0ca122364653e0f017a38c/devimint/src/cfg/bitcoin.conf#L9

I think we should give it a try.

oleonardolima:
> > https://github.com/fedimint/fedimint/blob/1ca5227e0ceb8c4f3b0ca122364653e0f017a38c/devimint/src/cfg/bitcoin.conf#L9
> 
> I think we should give it a try.

FWIW, I'm not sure if I did something wrong, but it still didn't solve the issue 🤔

storopoli:
rebased to latest `master`.
Added the new checks in rust `stable` for the example-crates.
Once bitcoindevkit/bdk#1599 is merged I can delete the 35070555e1cea64f663a9801adedee2944f7e10b, since it is the same commit

storopoli:
rebased to `master` after bitcoindevkit/bdk#1599 was merged

thesimplekid:
I stole this for cdk just fyi been working well. https://github.com/cashubtc/cdk/blob/main/flake.nix

storopoli:
rebased `master` after bitcoindevkit/bdk#1561 was merged

Nix.md
```diff
@@ -0,0 +1,147 @@
...
+They also provide pre-commit git hooks covered [above](#bdks-nix-based-tests-and-ci).
+
+Regarding the typos check, there might be some false positives.
+In that case you can add a regex rule to filter out the typos in the `.typos.toml` file.
+We already included all the false positives we've found so far with some explainable comments.
+Hence, you should be able to follow the examples and add your own.
+Additionally, you can find more information in [`crate-ci/typos`](https://github.com/crate-ci/typos).
+
+As a final note on the `devShell`s, they can be conveniently automated with
+[`nix-community/nix-direnv`](https://github/nix-community/nix-direnv).
```
notmandatory, line 142:
```suggestion
[`nix-community/nix-direnv`](https://github.com/nix-community/nix-direnv).
```

.github/workflows/cont_integration.yml
```diff
@@ -3,149 +3,133 @@ on: [push, pull_request]
...
-            override: true
-            profile: minimal
+        uses: actions/checkout@v4
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@v14
+      - name: Nix Cache
+        uses: DeterminateSystems/magic-nix-cache-action@v8
       - name: Rust Cache
-        uses: Swatinem/rust-cache@v2.2.1
-      - name: Pin dependencies for MSRV
```
notmandatory, line 41: As a library project I'd rather not use a checked in `Cargo.lock` to handle the MSRV version pinning problem.  Can we do it the way @thesimplekid  [did it in his project](https://github.com/cashubtc/cdk/blob/88b79eb9e9620dc9ff9f803ec7702afae4ae0c1b/flake.nix#L123)?

flake.nix
```diff
@@ -0,0 +1,188 @@
...
+              buildInputs = buildInputs ++ [ coverage_toolchain pkgs.lcov ];
+              inherit nativeBuildInputs;
+              RUSTFLAGS = "-Cinstrument-coverage";
+              RUSTDOCFLAGS = "-Cinstrument-coverage";
+              LLVM_PROFILE_FILE = "./target/coverage/%p-%m.profraw";
+            } // envVars);
+          in
+          {
+            inherit msrv stable nightly coverage;
+            default = msrv;
```
notmandatory, line 184: I'd rather make the `stable` shell the default since that's what most people will be using. We just need the MSRV shell tests to be run in CI or as a pre-push check if that's possible.

.github/workflows/update_flake_lock.yml
```diff
@@ -0,0 +1,21 @@
+name: Update Flake Lock File
```
notmandatory, line 1: This may require more testing to get it to use the right permissions and signing, as we had to do in #1580.  Might be a good idea to put this into a separate PR since it can be done manually for now.

Nix.md
```diff
@@ -0,0 +1,147 @@
...
+
+```shell
+nix develop .#SHELL
+```
+
+Where `SHELL` is the name of the `devShell` you want to use.
+For all the shells we provide they come with the WASM Rust target toolchain.
+We provide the following shells:
+
+- `msrv`: MSRV Rust version.
```
notmandatory, line 125: I'd rather have `stable` as the default.

README.md
```diff
@@ -62,28 +65,34 @@ Fully working examples of how to use these components are in `/example-crates`:
 [`electrum-client`]: https://docs.rs/electrum-client/
 [`bdk_chain`]: https://docs.rs/bdk-chain/
 
+## Dependencies
+
+- `bitcoind_rpc` depends on `bitcoind` being installed and available in `PATH` or in the `BITCOIND_EXEC` ENV variable.
+- `esplora` depends on [Blockstream's version of `electrs`](https://github.com/Blockstream/electrs)
+   being installed and available in `PATH` or in the `ELECTRS_EXEC` ENV variable.
+
```
notmandatory, line 73: Worth putting a reference here to the Nix.md file and that using the Nix devshells will setup the bitcoind and electrs env.

justfile
```diff
@@ -0,0 +1,56 @@
...
+  nixpkgs-fmt $(echo **.nix)
+
+# run tests
+test: build
+  #!/usr/bin/env bash
+  set -euo pipefail
+  if [ ! -f Cargo.toml ]; then
+    cd {{invocation_directory()}}
+  fi
+  cargo test
```
notmandatory, line 43: You shouldn't need to call `build` first but you do need the `ARGS`.  I'd also set the test threads so it works on macos until we can get that figured out.

```suggestion
test *ARGS="--workspace --all-targets -- --test-threads=2":
  #!/usr/bin/env bash
  set -euo pipefail
  if [ ! -f Cargo.toml ]; then
    cd {{invocation_directory()}}
  fi
  cargo test {{ARGS}}
```

notmandatory (changes requested):
Overall this looks like a great use of Nix to setup a local dev env use the pre-checks to prevent common commit problems. 

I would like to figure out the macos test threads issue, but that can be done in a follow-up PR. I also have a few other comments below. 

Thanks all the hard work getting this to work.

justfile
storopoli: Done! 0267841eef6243516bcdc95da0d1cb777fb687c5

README.md
storopoli: Done! 0267841eef6243516bcdc95da0d1cb777fb687c5

Nix.md
storopoli: Done! 0267841eef6243516bcdc95da0d1cb777fb687c5

.github/workflows/update_flake_lock.yml
storopoli: Done! 0267841eef6243516bcdc95da0d1cb777fb687c5

flake.nix
storopoli: Done! 0267841eef6243516bcdc95da0d1cb777fb687c5

Nix.md
storopoli: Done! 0267841eef6243516bcdc95da0d1cb777fb687c5

.github/workflows/cont_integration.yml
storopoli: Done in bf151a9b1babb42798bf852e17e3e118938ef364 since I don't know if it will work.
After we verify I can squash all into a single commit.

storopoli:
> Thanks all the hard work getting this to work.

My pleasure.

Ok the trick that @thesimplekid has that does not need a `Cargo.lock` seems to work. Thanks! (I did some extra stuff that calls `cargo update` in the non-MSRV `shellHooks`)
I'll squash everything into a single commit and ask for a new round of reviews.

PS: Since we are changing the GitHub CI workflow names, we might need to update the branch protection rules for `master` after this gets merged.

notmandatory:
I just ran into an error with the pre-push hook that it didn't think 2cf46a2a9e513d3f2612b7c363446fb1acf958ae is signed because (I think) it was signed with an ssh key instead of pgp. Can you repro this @storopoli ? I found it when pushing latest master to my github fork repo.

storopoli:
You are hitting this: https://blog.dbrgn.ch/2021/11/16/git-ssh-signatures/

You need to whitelist @ValuedMammal's ssh key in `error: gpg.ssh.allowedSignersFile needs to be configured and exist for ssh signature verification`

notmandatory:
@storopoli sorry to have to bump this one from 1.0 but I think we need to have some more discussions about how nix helps the project and where/how it should be introduced. Also there are multiple new things being introduced here and probably better to put ci, hooks, justfile changes in new prs. But when you're free let's schedule a time for a live that. Thanks!

storopoli:
Sure, I'll move this to a draft for now.

storopoli:
I am closing this since I won't be able to provide the full maintainability support.
Anyone is free to pick it up, glad to help if needed for a future PR with reviews or suggestions.

PR #1323, opened 2024-02-05
Updated is_mine documentation

matthiasdebernardini:
### Description

To clarify the purpose of a method and to indicate how to get more information from the wallet

### Notes to the reviewers

We can also change the how is_mine works by calling `derivation_of_spk` directly, so its more clear how the internals work. This way rather than having the docs explicitly call out the method, they can jump to source.

#### All Submissions:

* [ x] I've signed all my commits
* [ x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [ x] I ran `cargo fmt` and `cargo clippy` before committing

### DIFF
--- a/crates/bitcoind_rpc/Cargo.toml
+++ b/crates/bitcoind_rpc/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bdk_bitcoind_rpc"
-version = "0.10.0"
+version = "0.11.0"
 edition = "2021"
 rust-version = "1.63"
 homepage = "https://bitcoindevkit.org"
@@ -16,7 +16,7 @@ readme = "README.md"
 # For no-std, remember to enable the bitcoin/no-std feature
 bitcoin = { version = "0.31", default-features = false }
 bitcoincore-rpc = { version = "0.18" }
-bdk_chain = { path = "../chain", version = "0.14", default-features = false }
+bdk_chain = { path = "../chain", version = "0.15", default-features = false }
 
 [dev-dependencies]
 bdk_testenv = { path = "../testenv", default-features = false }
--- a/crates/chain/Cargo.toml
+++ b/crates/chain/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bdk_chain"
-version = "0.14.0"
+version = "0.15.0"
 edition = "2021"
 rust-version = "1.63"
 homepage = "https://bitcoindevkit.org"
--- a/crates/electrum/Cargo.toml
+++ b/crates/electrum/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bdk_electrum"
-version = "0.13.0"
+version = "0.14.0"
 edition = "2021"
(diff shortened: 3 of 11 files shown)


LLFourn:
Sure I support reimplementing in terms of `derivation_of_spk`. Also there might be a better name for it.

matthiasdebernardini:
```
    /// Finds how the wallet derived the script pubkey `spk`.
    ///
    /// Will only return `Some(_)` if the wallet has given out the spk.
    pub fn derivation_of_spk(&self, spk: &Script) -> Option<(KeychainKind, u32)> {
        self.indexed_graph.index.index_of_spk(spk)
    }
```

> Also there might be a better name for it

@LLFourn rename this? Could be `find_spk_derivation` or `find_spk`? 

Also with regards to;

> Will only return `Some(_)` if the wallet has given out the spk

I think this means to say that it will return `Some(_)` if the spk is findable/derivable, to me, "has given out" means its tracking which spks its been giving out which sounds inaccurate.

LLFourn:
> ```
>     /// Finds how the wallet derived the script pubkey `spk`.
>     ///
>     /// Will only return `Some(_)` if the wallet has given out the spk.
>     pub fn derivation_of_spk(&self, spk: &Script) -> Option<(KeychainKind, u32)> {
>         self.indexed_graph.index.index_of_spk(spk)
>     }
> ```
> 
> > Also there might be a better name for it
> 
> @LLFourn rename this? Could be `find_spk_derivation` or `find_spk`?

I don't like `find` because it implies it's not just doing a lookup. `find` in rust is used in iterators to go through each element and return the first match.

> 
> Also with regards to;
> 
> > Will only return `Some(_)` if the wallet has given out the spk
> 
> I think this means to say that it will return `Some(_)` if the spk is findable/derivable, to me, "has given out" means its tracking which spks its been giving out which sounds inaccurate.

I think it was accurate. That's one of the main functions of the `KeychainTxOutIndex` is to track which has given out. However I think this will now return results for things that have not yet been given out because of the "lookahead" feature so I do think the docs needs to be updated to say that if it matches a script pubkey that is has derived internally.

notmandatory:
Since this is just a docs and internal implementation change that doesn't change the API I'd like to move it to a post 1.0 milestone.

crates/bdk/src/wallet/mod.rs
```diff
@@ -814,7 +814,10 @@ impl<D> Wallet<D> {
         })
     }
 
-    /// Return whether or not a `script` is part of this wallet (either internal or external)
+    /// Determines if `script` is associated with either the internal or external keychain of the wallet.
+    ///
+    /// To determine if the `script` belongs specifically to either
+    /// the internal or external keychain, use the `wallet.derivation_of_spk()` method.
```
evanlinjin: Can we create a link for `wallet.derivation_of_spk`?

evanlinjin (changes requested):

matthiasdebernardini:
I closed this but would like to revisit this post 1.0 release.

PR #1324, opened 2024-02-06
[chain] Make KeychainTxOutIndex more range based

LLFourn:
KeychainTxOut index should try and avoid "all" kind of queries. There may be subranges of interest. If the user wants "all" they can just query "..".

The ideas is that KeychainTxOutIndex should be designed to be able to incorporate many unrelated keychains that can be managed in the same index. We should be able to see the "net_value" of a transaction to a specific subrange. e.g. imagine a collaborative custody service that manages all their user descriptors inside the same `KeychainTxOutIndex`. One user in their service may pay another so when you are analyzing how much a transaction is spending for a particular user you need to do analyze a particular sub-range.

### Notes to the reviewers

- I didn't change `unused_spks` to follow this rule because I want to delete that method some time in the future. `unused_spks` is being used in the examples for syncing but it shouldn't be (the discussion as to why will probably surface in #1194).
- I haven't applied this reasoning to the methods that return `BTreeMap`s e.g. `all_unbounded_spk_iters`. It probably should be but I haven't made up my mind yet.

This probably belongs after #1194

### Changelog notice

- `KeychainTxOutIndex` methods modified to take ranges of keychains instead.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -1015,7 +1015,7 @@ impl<D> Wallet<D> {
     /// let (sent, received) = wallet.sent_and_received(tx);
     /// ```
     pub fn sent_and_received(&self, tx: &Transaction) -> (u64, u64) {
-        self.indexed_graph.index.sent_and_received(tx)
+        self.indexed_graph.index.sent_and_received(tx, ..)
     }
 
     /// Get a single transaction from the wallet as a [`CanonicalTx`] (if the transaction exists).
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -268,15 +268,14 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         self.inner.unmark_used(&(keychain, index))
     }
 
-    /// Computes total input value going from script pubkeys in the index (sent) and the total output
-    /// value going to script pubkeys in the index (received) in `tx`. For the `sent` to be computed
-    /// correctly, the output being spent must have already been scanned by the index. Calculating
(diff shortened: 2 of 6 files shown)


ValuedMammal:
If I understand, you're using range in a couple different ways: in keychain/txout_index to range over keychains, and in spk_txout_index to range over derivation indexes - with `map_to_inner_bounds` having some mediating role

Am I right in thinking these are similar but distinct concepts? or is it just two ways of looking at the same thing?

LLFourn:
> If I understand, you're using range in a couple different ways: in keychain/txout_index to range over keychains, and in spk_txout_index to range over derivation indexes - with `map_to_inner_bounds` having some mediating role
> 
> Am I right in thinking these are similar but distinct concepts? or is it just two ways of looking at the same thing?

`SpkTxOutIndex` they are similar. `KeychainTxOutIndex` is an `SpkTxOutIndex` with a specialized structure where each group of spks is under something called a "keychain". In `SpkTxOutIndex` each spk is under its own unique key. In order to query across multiple keychains in `KeychainTxOutIndex`, you need to be able to query across sub-ranges of indices in `SpkTxOutIndex`.

notmandatory:
Do we have a current need for this or can it be moved to a post 1.0 release as a non-breaking internal optimization?

LLFourn:
It's an API change for `bdk_chain` and doesn't effect `bdk_wallet` I don't think. it doesn't need a milestone since if I understand that system is only meant for prioritizing tasks with respect to `bdk_wallet` where as the order of other tasks is left to the discretion of the devs.

evanlinjin (approved):
ACK fac228337c79e1a7c406500c177e049e7f9f5da7

Issue #1147, opened 2023-10-05
[chain] Consider Merging IndexedTxGraph and TxGraph into one thing


LLFourn:
currently:

```
#[derive(Debug)]
pub struct IndexedTxGraph<A, I> {
    /// Transaction index.
    pub index: I,
    graph: TxGraph<A>,
}
```

But it could also just be:

```
#[derive(Clone, Debug, PartialEq)]
pub struct TxGraph<A = (), I = ()> {
    // all transactions that the graph is aware of in format: `(tx_node, tx_anchors, tx_last_seen)`
    txs: HashMap<Txid, (TxNodeInternal, BTreeSet<A>, u64)>,
    spends: BTreeMap<OutPoint, HashSet<Txid>>,
    anchors: BTreeSet<(A, Txid)>,

    // This atrocity exists so that `TxGraph::outspends()` can return a reference.
    // FIXME: This can be removed once `HashSet::new` is a const fn.
    empty_outspends: HashSet<Txid>,
    pub index: I
}
```

But we would have to give up the derive `PartialEq` on `TxGraph` because I don't think we want to force indexes to implement it.

The motivation is avoiding having to duplicate every mutating API on `TxGraph` to `IndexedTxGraph`. We seem to have some trouble doing this. For example, `IndexedTxGraph` and `TxGraph` both have `insert_tx` but they have different arguments atm but they should be the same. Also in #1041I a method is being added to `IndexedTxGraph` but it should also be on `TxGraph` but the PR author forgot.

The downside is that `TxGraph` would not longer be a pure data structure but since we don't use it like that I don't think this matters.

evanlinjin:
"the PR author"

notmandatory:
I moved this to alpha.4 since we want to get any functional changes in before a beta release.

LLFourn:
FWIW I think this is still a good idea.

nondiremanuel:
Can we postpone it to a 1.1 version @notmandatory?

evanlinjin:
I think it is okay to postpone this to after 1.0.0. I think it may simplify the API and make it cleaner to use with less foot-guns. However, it is also not critical to have this.

PR #1325, opened 2024-02-06
Add `map_anchors` for `TxGraph`

### LINKED ISSUE #1295: `TxGraph<A>` should be able to transform anchor types.
**Describe the enhancement**  

```rust
impl<A1, A2: From<A1>> From<TxGraph<A1>> for TxGraph<A2> {
    // TODO
}
```

If `A2: From<A1>` is impossible, we can have a method:

```rust
impl<A> TxGraph<A> {
    pub fn map_anchors<A2, F>(self, f: F) -> TxGraph<A2> where F: Fn(A) -> A2 { todo!() }
}
```

**Use case**  

Some chain source methods return `TxGraph`s. It will be good to update original `TxGraph`s with different anchor types.

yanganto:
### Description
Fix bitcoindevkit/bdk#1295 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -452,6 +452,21 @@ impl<A> TxGraph<A> {
 }
 
 impl<A: Clone + Ord> TxGraph<A> {
+    /// Transform the [`TxGraph`] to have [`Anchor`]s of another type.
+    ///
+    /// This takes in a closure of signature `FnMut(A) -> A2` which is called for each [`Anchor`] to
+    /// transform it.
+    pub fn map_anchors<A2: Clone + Ord, F>(self, f: F) -> TxGraph<A2>
+    where
+        F: FnMut(A) -> A2,
+    {
+        let mut new_graph = TxGraph::<A2>::default();
+        new_graph.apply_changeset(self.initial_changeset().map_anchors(f));
+        new_graph
+    }
+}
+
+impl<A: Clone + Ord> TxGraph<A> {
     /// Construct a new [`TxGraph`] from a list of transactions.
     pub fn new(txs: impl IntoIterator<Item = Transaction>) -> Self {
         let mut new = Self::default();
@@ -1293,6 +1308,26 @@ impl<A: Ord> Append for ChangeSet<A> {
             && self.txouts.is_empty()
             && self.anchors.is_empty()
             && self.last_seen.is_empty()
+    }
+}
+
+impl<A: Ord> ChangeSet<A> {
+    /// Transform the [`ChangeSet`] to have [`Anchor`]s of another type.
+    ///
+    /// This takes in a closure of signature `FnMut(A) -> A2` which is called for each [`Anchor`] to
+    /// transform it.
+    pub fn map_anchors<A2: Ord, F>(self, mut f: F) -> ChangeSet<A2>
+    where
+        F: FnMut(A) -> A2,
+    {
+        ChangeSet {
+            txs: self.txs,
+            txouts: self.txouts,
+            anchors: BTreeSet::<(A2, Txid)>::from_iter(
+                self.anchors.into_iter().map(|(a, txid)| (f(a), txid)),
+            ),
+            last_seen: self.last_seen,
+        }
     }
 }
 
--- a/crates/chain/tests/common/tx_template.rs
+++ b/crates/chain/tests/common/tx_template.rs
@@ -1,7 +1,7 @@
 use rand::distributions::{Alphanumeric, DistString};
 use std::collections::HashMap;
 
-use bdk_chain::{tx_graph::TxGraph, BlockId, SpkTxOutIndex};
+use bdk_chain::{tx_graph::TxGraph, Anchor, SpkTxOutIndex};
 use bitcoin::{
     locktime::absolute::LockTime, secp256k1::Secp256k1, OutPoint, ScriptBuf, Sequence, Transaction,
     TxIn, TxOut, Txid, Witness,
@@ -49,11 +49,11 @@ impl TxOutTemplate {
 }
 
 #[allow(dead_code)]
-pub fn init_graph<'a>(
-    tx_templates: impl IntoIterator<Item = &'a TxTemplate<'a, BlockId>>,
-) -> (TxGraph<BlockId>, SpkTxOutIndex<u32>, HashMap<&'a str, Txid>) {
+pub fn init_graph<'a, A: Anchor + Copy + 'a>(
+    tx_templates: impl IntoIterator<Item = &'a TxTemplate<'a, A>>,
+) -> (TxGraph<A>, SpkTxOutIndex<u32>, HashMap<&'a str, Txid>) {
     let (descriptor, _) = Descriptor::parse_descriptor(&Secp256k1::signing_only(), "tr(tprv8ZgxMBicQKsPd3krDUsBAmtnRsK3rb8u5yi1zhQgMhF1tR8MW7xfE4rnrbbsrbPR52e7rKapu6ztw1jXveJSCGHEriUGZV7mCe88duLp5pj/86'/1'/0'/0/*)").unwrap();
-    let mut graph = TxGraph::<BlockId>::default();
+    let mut graph = TxGraph::<A>::default();
     let mut spk_index = SpkTxOutIndex::default();
     (0..10).for_each(|index| {
         spk_index.insert_spk(
--- a/crates/chain/tests/common/tx_template.rs
+++ b/crates/chain/tests/common/tx_template.rs
@@ -49,7 +49,7 @@ impl TxOutTemplate {
 }
 
 #[allow(dead_code)]
-pub fn init_graph<'a, A: Anchor + Copy + 'a>(
+pub fn init_graph<'a, A: Anchor + Clone + 'a>(
     tx_templates: impl IntoIterator<Item = &'a TxTemplate<'a, A>>,
 ) -> (TxGraph<A>, SpkTxOutIndex<u32>, HashMap<&'a str, Txid>) {
     let (descriptor, _) = Descriptor::parse_descriptor(&Secp256k1::signing_only(), "tr(tprv8ZgxMBicQKsPd3krDUsBAmtnRsK3rb8u5yi1zhQgMhF1tR8MW7xfE4rnrbbsrbPR52e7rKapu6ztw1jXveJSCGHEriUGZV7mCe88duLp5pj/86'/1'/0'/0/*)").unwrap();
@@ -126,7 +126,7 @@ pub fn init_graph<'a, A: Anchor + Copy + 'a>(
         spk_index.scan(&tx);
         let _ = graph.insert_tx(tx.clone());
         for anchor in tx_tmp.anchors.iter() {
-            let _ = graph.insert_anchor(tx.txid(), *anchor);
+            let _ = graph.insert_anchor(tx.txid(), anchor.clone());
         }
         if let Some(seen_at) = tx_tmp.last_seen {
             let _ = graph.insert_seen_at(tx.txid(), seen_at);
--- a/crates/chain/tests/test_tx_graph.rs
+++ b/crates/chain/tests/test_tx_graph.rs
@@ -10,7 +10,9 @@ use bdk_chain::{
 use bitcoin::{
     absolute, hashes::Hash, BlockHash, OutPoint, ScriptBuf, Transaction, TxIn, TxOut, Txid,
 };
+use common::*;
 use core::iter;
+use rand::RngCore;
 use std::vec;
 
 #[test]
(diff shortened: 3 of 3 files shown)


evanlinjin:
Thank you for this PR! What is the purpose of commit a223c0c?

crates/chain/src/tx_graph.rs
```diff
@@ -449,6 +449,32 @@ impl<A> TxGraph<A> {
     pub fn is_empty(&self) -> bool {
         self.txs.is_empty()
     }
+
+    /// Map TxGraph for another anchor type
```
```suggestion
    /// Transform the [`TxGraph`] to have [`Anchor`]s of another type.
    ///
    /// This takes in a closure of signature `Fn(A) -> A2` which is called for each [`Anchor`] to
    /// transform it.
```


Some notes on writing documentation:

* Use the following structure (as suggested [here](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components)), but maybe skip the example for simple methods (such as this one):

    ```
    [short sentence explaining what it is]
    
    [more detailed explanation]
    
    [at least one code example that users can copy/paste to try it]
    
    [even more advanced explanations if necessary]
    ```

* Use links when referencing structures (such as `TxGraph`).

crates/chain/src/tx_graph.rs
```diff
@@ -449,6 +449,32 @@ impl<A> TxGraph<A> {
...
+    pub fn map_anchors<A2: Ord, F>(self, f: F) -> TxGraph<A2>
+    where
+        F: Fn(A) -> A2,
```
evanlinjin: I think using `FnMut` will be better. This gives the caller more flexibility.

Also note that [`Iterator::map` takes in `FnMut`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map).

crates/chain/src/tx_graph.rs
```diff
@@ -449,6 +449,32 @@ impl<A> TxGraph<A> {
...
+                            tx_node_interal.clone(),
+                            BTreeSet::from_iter(set.into_iter().map(&f)),
+                            amt,
+                        ),
+                    )
+                },
+            )),
+            anchors: BTreeSet::<(A2, Txid)>::from_iter(
+                self.anchors.into_iter().map(|(a, txid)| (f(a), txid)),
+            ),
```
evanlinjin: This is written in a way which assumes the caller's implementation of `F` is deterministic. A.K.A passing in the same `A` on separate calls will give us the same `A2` result.

The benefit of NOT having this assumption is:
1. It's safer.
2. It's probably more performant (as there will be no duplicate calls on `F`).

evanlinjin: We should write a test to make sure the resultant `TxGraph` is reconstructible.

1. Call `.map_anchor` on `graph_a` with a very strange `F` that is non-deterministic (let's call the resultant `TxGraph` - `graph_b`).
2. Call `graph_b.initial_changeset`. Construct a `graph_c` with the resultant changeset.
3. `assert_eq!(graph_b, graph_c)`.

Also another test that shows `graph_a` and `graph_b` have the same transactions, and last_seens. Also the same anchor count.

evanlinjin (changes requested):
Thanks for this work. This will be a useful feature to have (talking from personal experience).

yanganto:
> Thank you for this PR! What is the purpose of commit [a223c0c](https://github.com/bitcoindevkit/bdk/commit/a223c0c63726dcff203d32fe1fc3edae1525b5a2)?

I see the doc
```rust
/// The generic `A` should be a [`Anchor`] implementation.
```

And it was not required in the enum, if it is required, we may not need an additional description on the document.

This is a chore, and not about this topic, so it is separated in another commit.

evanlinjin:
@yanganto the key `K` for a `BTreeSet` in the stdlib is only useful if it impl's `Ord`, but it is not constrained in the struct definition ([check here](https://doc.rust-lang.org/src/alloc/collections/btree/map.rs.html#172-176)).

The reason for this is that generic constraints stack on top of each other and will start becoming tedious if types depend on types that have generics with constraints. So when we define anything, we avoid enforcing generic constraints wherever we can.

yanganto:
> @yanganto the key `K` for a `BTreeSet` in the stdlib is only useful if it impl's `Ord`, but it is not constrained in the struct definition ([check here](https://doc.rust-lang.org/src/alloc/collections/btree/map.rs.html#172-176)).
> 
> The reason for this is that generic constraints stack on top of each other and will start becoming tedious if types depend on types that have generics with constraints. So when we define anything, we avoid enforcing generic constraints wherever we can.

I think there is no `Ord` constrain on `BTreeSet` struct and there is no doc on the struct because the real constraint is from the [impl](https://doc.rust-lang.org/src/alloc/collections/btree/map.rs.html#291-295), and the `get`, `take`, `replace` are the real constraints with `Ord`.  I dropped the commit here and maybe dig into this later on.  It is not a chore and out of this thread.

crates/chain/src/tx_graph.rs
yanganto: We can not use `FnMut` here, and we use `FnOnce` which also gives caller the flexibility.

yanganto:
@evanlinjin Does the test case look good from your end? If it is good, I will squash all the commit and resign it. :pray:

evanlinjin:
> > @yanganto the key `K` for a `BTreeSet` in the stdlib is only useful if it impl's `Ord`, but it is not constrained in the struct definition ([check here](https://doc.rust-lang.org/src/alloc/collections/btree/map.rs.html#172-176)).
> > 
> > The reason for this is that generic constraints stack on top of each other and will start becoming tedious if types depend on types that have generics with constraints. So when we define anything, we avoid enforcing generic constraints wherever we can.
> 
> I think there is no `Ord` constrain on `BTreeSet` struct and there is no doc on the struct because the real constraint is from the [impl](https://doc.rust-lang.org/src/alloc/collections/btree/map.rs.html#291-295), and the `get`, `take`, `replace` are the real constraints with `Ord`.  I dropped the commit here and maybe dig into this later on.  It is not a chore and out of this thread.

That's exactly my point. We should only put generic constraints where it's strictly necessary.

crates/chain/tests/common/tx_template.rs
```diff
@@ -49,11 +49,11 @@ impl TxOutTemplate {
 }
 
 #[allow(dead_code)]
-pub fn init_graph<'a>(
-    tx_templates: impl IntoIterator<Item = &'a TxTemplate<'a, BlockId>>,
-) -> (TxGraph<BlockId>, SpkTxOutIndex<u32>, HashMap<&'a str, Txid>) {
+pub fn init_graph<'a, A: Anchor + Copy + 'a>(
```
evanlinjin: Currently, all our `Anchor` implementations impl `Copy`, but that might not always be the case. I think we should loosen the constraint to be `A: Anchor + Clone + 'a`.

crates/chain/src/tx_graph.rs
yanganto: Got it.  It now maps only once on ChangeSet.

yanganto:
There are more tx and blocks in the test case, and ready for review.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -1172,3 +1174,71 @@ fn test_missing_blocks() {
...
+            inputs: &[TxInTemplate::PrevTx("genesis", 0)],
+            outputs: &[TxOutTemplate::new(20000, Some(2))],
+            anchors: &[block_id!(2, "B")],
+            ..Default::default()
+        },
+        TxTemplate {
+            tx_name: "block 2",
+            inputs: &[TxInTemplate::PrevTx("block 1", 0)],
+            outputs: &[TxOutTemplate::new(30000, Some(3))],
+            anchors: &[block_id!(3, "C")],
```
LLFourn: Can we try one of the txs having multiple anchors here.

crates/chain/src/tx_graph.rs
```diff
@@ -1296,6 +1311,28 @@ impl<A: Ord> Append for ChangeSet<A> {
...
+    where
+        F: FnMut(A) -> A2 + Clone,
+    {
+        ChangeSet {
+            txs: self.txs,
+            txouts: self.txouts,
+            anchors: BTreeSet::<(A2, Txid)>::from_iter(
+                self.anchors
+                    .into_iter()
+                    .map(|(a, txid)| (f(a).clone(), txid)),
```
LLFourn: what's with the `.clone()` here? If it's not necessary please remove from the `F` bounds as well.

yanganto: Many thanks :pray:

crates/chain/tests/test_tx_graph.rs
yanganto: Good idea.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -1172,3 +1174,71 @@ fn test_missing_blocks() {
...
+        assert_eq!(new_txnode.anchors.len(), tx_node.anchors.len());
+
+        let mut new_anchors = new_txnode.anchors.iter();
+        for anchor in tx_node.anchors.iter() {
+            assert_eq!(new_anchors.next().unwrap().anchor_block, *anchor);
+        }
+        assert!(new_anchors.next().is_none());
+    }
+    assert!(new_txs.next().is_none());
+}
```
evanlinjin: There are still some checks missing.

* The `TxGraph::anchors` field is not checked. We can use the `TxGraph::all_anchors` method to get it.
* The current test only checks whether all txs in `graph` are in `new_graph`, but it doesn't check for duplicates or potential additions.

Suggestion: If we collect things we want to check into `BTreeSet`, it makes the checks more complete and easier to read.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -1172,3 +1174,71 @@ fn test_missing_blocks() {
...
+        },
+        TxTemplate {
+            tx_name: "block 1",
+            inputs: &[TxInTemplate::PrevTx("genesis", 0)],
+            outputs: &[TxOutTemplate::new(20000, Some(2))],
+            anchors: &[block_id!(2, "B")],
+            ..Default::default()
+        },
+        TxTemplate {
+            tx_name: "block 2",
```
evanlinjin: Nit: Can we rename these tx names? tx1, tx2, tx3 are fine.

crates/chain/src/tx_graph.rs
```diff
@@ -1296,6 +1311,26 @@ impl<A: Ord> Append for ChangeSet<A> {
     }
 }
 
+impl<A: Clone + Ord> ChangeSet<A> {
+    /// Transform the [`ChangeSet`] to have [`Anchor`]s of another type.
+    ///
+    /// This takes in a closure of signature `FnMut(A) -> A2` which is called for each [`Anchor`] to
+    /// transform it.
+    pub fn map_anchors<A2: Clone + Ord, F>(self, mut f: F) -> ChangeSet<A2>
```
evanlinjin: The `Clone` bound is not needed for `A` here.

crates/chain/tests/test_tx_graph.rs
yanganto: I got it. The all_anchors of a graph are checked, and the tx_node iterator may not follow the block_id order and will be in a different order, so it still checks with a loop.

evanlinjin: The nature of `BTree{Set|Map}`s is that they are always ordered (hence you can `.collect` and compare the results). So no, comparing inside a loop is not required.

Personally I think this is cleaner, but this is personal preference.

crates/chain/tests/common/tx_template.rs
evanlinjin: Is this done?

yanganto: Sorry about delay, we are using clone for this now.

evanlinjin (approved):
ACK 5489f905a434ecc06867603c7c421e3e50d993ca

yanganto:
Thanks for the review.

crates/chain/src/tx_graph.rs
```diff
@@ -451,6 +451,21 @@ impl<A> TxGraph<A> {
...
+    /// Transform the [`TxGraph`] to have [`Anchor`]s of another type.
+    ///
+    /// This takes in a closure of signature `FnMut(A) -> A2` which is called for each [`Anchor`] to
+    /// transform it.
+    pub fn map_anchors<A2: Clone + Ord, F>(self, f: F) -> TxGraph<A2>
+    where
+        F: FnMut(A) -> A2,
+    {
+        let mut new_graph = TxGraph::<A2>::default();
+        new_graph.apply_changeset(self.initial_changeset().map_anchors(f));
```
LLFourn, line 464: :pick:  This seems weird because we allocate a giant changeset and then just apply it. I think we can just repeat the process you do in `ChangeSet`.

LLFourn (approved):
ACK 5489f905a434ecc06867603c7c421e3e50d993ca

I guess we don't want to add this method to `IndexedTxGraph`?. I think this is another reason to do #1147.

crates/chain/src/tx_graph.rs
yanganto: There are multiple anchor-related fields in `TxGraph`, and only one in `ChangeSet`.  The function caller passed in possible non-determined, so it is good to map it once here to avoid issues from a non-determined function.

LLFourn: Huh isn't there only one acnhor related field on `TxGraph` also?

evanlinjin: > Huh isn't there only one acnhor related field on `TxGraph` also?

We also have anchors in `TxNode` which should have 1 to 1 mapping to the `anchors` field. I think the implementation right now is safer and easier to read. If in the future, there is clear performance problems we can do something more efficient.

evanlinjin:
> I guess we don't want to add this method to `IndexedTxGraph`?. I think this is another reason to do bitcoindevkit/bdk_wallet#130.

My gripe with bitcoindevkit/bdk_wallet#130 is that I'm assuming the `Indexer` interface is too restrictive and I would like to think about it further before making it mandatory by including it in `TxGraph`. I.e. with Utreexo and for optimizing updating the "transaction/utxo view" of a purely block-by-block sync. For the second point, it will be good if the indexer returned which outputs and inputs are relevant straight away. This way, we can update our view of the UTXO set without reconstructing it from the `TxGraph`.

Otherwise, yes, we need `map_anchors` on `IndexedTxGraph` as well (for now).

PR #1334, opened 2024-02-09
Reorder fields in ConfirmationHeightAnchor fields so Ord DWIM

LLFourn:
Something that is confirmed more recently should be greater than something that is confirmed earlier regardless of anchor.


evanlinjin (approved):
ACK adc95137ac2e52f4d45c4ec778703001c95a362f

PR #1335, opened 2024-02-09
fix(chain): tx_graph::ChangeSet::is_empty

LLFourn:
🙈

### Changelog notice

- Fix bug in `tx_graph::ChangeSet::is_empty` where is returns true even when it wasn't empty

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -67,7 +67,10 @@
 //!
 //! // if we apply it again, the resulting changeset will be empty
 //! let changeset = graph.apply_update(update);
-//! assert!(changeset.is_empty());
+//! assert!({
+//!     use bdk_chain::Append;
+//!     changeset.is_empty()
+//! });
 //! ```
 //! [`try_get_chain_position`]: TxGraph::try_get_chain_position
 //! [`insert_txout`]: TxGraph::insert_txout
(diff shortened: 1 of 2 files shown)


crates/chain/src/tx_graph.rs
```diff
@@ -1214,7 +1214,10 @@ impl<A> Default for ChangeSet<A> {
 impl<A> ChangeSet<A> {
     /// Returns true if the [`ChangeSet`] is empty (no transactions or txouts).
     pub fn is_empty(&self) -> bool {
-        self.txs.is_empty() && self.txouts.is_empty()
+        self.txs.is_empty()
+            && self.txouts.is_empty()
+            && self.anchors.is_empty()
+            && self.last_seen.is_empty()
     }
```
evanlinjin: `tx_graph::ChangeSet` implements `Append` and `Append` already has a method `is_empty`. So this is a duplicate and we can just remove it.

The caveat (which is not a big one imo), is the caller needs to `use bdk_chain::Append` to call it.

evanlinjin:
Woops! Thank you for catching this. We had duplicate methods so I think we should get rid of one of them.

evanlinjin:
@LLFourn I pushed a commit on top, let me know if you are happy with it

LLFourn:
Thanks @evanlinjin I like that solution. I took a moment to clear up the docs. It looked cluttered to `use bdk_chian::Append` there just above the call so I hid it in the example. Users will find out the trait needs to be imported when they try and use it.

Self-ACK: 13ab5a835d59341e387fbbefec12fe4f48e5f3c8

evanlinjin (approved):
ACK 13ab5a835d59341e387fbbefec12fe4f48e5f3c8

notmandatory:
Hey sorry I didn't get this in the release notes for alpha.6, it wasn't on the BDK project so I didn't see it. It still shows up on the github releases page for anyone really interested.

Issue #1336, opened 2024-02-09
last_seen is wrong on both electrum and esplora


LLFourn:
1. Esplora never sets `last_seen`
2. Electrum workflow sets it even for transactions it sees as confirmed.

`last_seen` should be strictly set when the transaction is seen in the mempool. If you set it superfluously you create db churn. Every change to it is added to the changeset and written to disk.

evanlinjin:
For anyone new reading this ticket, I'll provide some more context on:

> `last_seen` should be strictly set when the transaction is seen in the mempool. If you set it superfluously you create db churn. Every change to it is added to the changeset and written to disk.

Our one and only `PersistenceBackend` implementation (`bdk_file_store`) is an append only flat file (hence, the db churn). There is no disastrous consequence if we have a `last_seen` for every transaction. However, it will be good to minimize data written.

evanlinjin:
I didn't include `last_seen` with Esplora because of API limitations. Someone please prove me wrong!

LLFourn:
@evanlinjin API limitations? I think you're just meant to get the `SystemTime::now`. That's what we do in the electrum example why can't the same thing work in esplora?

evanlinjin:
> @evanlinjin API limitations? I think you're just meant to get the `SystemTime::now`. That's what we do in the electrum example why can't the same thing work in esplora?
> 

I forgot that's what we did. Does esplora return conflicting unconfirmed txs though? (For some reason I remember it did) If it does, the timestamp won't be helpful.

LLFourn:
@evanlinjin I think it does. That's a good point but there's nothing we can do about that here. The `TxGraph` will resolve that.

evanlinjin:
If it does, would it be useful to use System time to get the timestamp for last_seen?

LLFourn:
It's better than it not having `last_seen` at all?

evanlinjin:
> It's better than it not having `last_seen` at all?

How so? The last seen is only used for conflict resolution and with Esplora acting in this way, all unconfirmed transactions will have the same last seen.

LLFourn:
It's informational. You can display this to the user in the app. It gives you an ordering on unconfirmed transactions. You can use it to tell when a tx was replaced (roughly).

ValuedMammal:
[...]

PR #1337, opened 2024-02-10
Fix `last_seen` in Esplora and Electrum chains [WIP]

rustaceanrob:
### Description

The last seen unix time should be set for mempool transactions (#1336). I started working on this in the blocking Esplora client.

### Notes to the reviewers

I decided to check-in with a WIP to get some clarifications from maintainers. There is a method on `tx_graph` that allows an insertion into the map for storing `last_seen` . This method is `insert_seen_at`, which I think could be called `insert_last_seen` without any ambiguity. When a transaction has been confirmed, there is no method to remove the `Txid` from the map so we don't continually store `last_seen` when a TX has been confirmed. 

Should I add a `remove_seen_at` in `tx_graph`? Or if my naming alternative makes sense, add a `remove_last_seen`?

Thanks (:

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/esplora/src/blocking_ext.rs
+++ b/crates/esplora/src/blocking_ext.rs
@@ -314,7 +314,7 @@ impl EsploraExt for esplora_client::BlockingClient {
                             if !status.confirmed {
                                 let last_seen = anchor.confirmation_time;
                                 let _ = graph.insert_seen_at(op.txid, last_seen);
-                            } 
+                            }
                             // else, graph.remove_seen_at or similar as the tx is confirmed
                         }
                     }
--- a/crates/esplora/src/blocking_ext.rs
+++ b/crates/esplora/src/blocking_ext.rs
@@ -295,6 +295,10 @@ impl EsploraExt for esplora_client::BlockingClient {
                 let status = self.get_tx_status(&op.txid)?;
                 if let Some(anchor) = anchor_from_status(&status) {
                     let _ = graph.insert_anchor(op.txid, anchor);
+                    if !status.confirmed {
+                        let last_seen = anchor.confirmation_time;
+                        let _ = graph.insert_seen_at(op.txid, last_seen);
+                    }
                 }
             }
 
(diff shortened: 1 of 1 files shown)


evanlinjin:
Concept ACK, approach NACK.

This will be a great feature to have! However, I think we are limited by the Esplora/Electrum APIs (let me know if I have missed something).

Note that your approach here is incorrect. Please check the Esplora API docs: https://github.com/Blockstream/esplora/blob/master/API.md#transaction-format

`status.block_time` is only available for confirmed transactions (so will always be `None`).

rustaceanrob:
Oh got it! That is trickier to implement than I thought. Am I correct in that the `last_seen` field is only intended for mempool transactions, and would be appropriate to remove once a transaction has been confirmed? In that case a `remove_seen_at` on `tx_graph` could be beneficial later.

evanlinjin:
> Oh got it! That is trickier to implement than I thought. Am I correct in that the `last_seen` field is only intended for mempool transactions, and would be appropriate to remove once a transaction has been confirmed? In that case a `remove_seen_at` on `tx_graph` could be beneficial later.

`last_seen` just means "last seen in the mempool". It is used to resolve conflicts between unconfirmed transactions (not necessarily mempool transactions - BDK would not know, it just approximates). The current assumption is that for two conflicting unconfirmed transactions, the transaction with the more recent last_seen takes precedence (however, this approximation can be improved - see bitcoindevkit/bdk_wallet#109).

Currently, the nature of `TxGraph` is fully monotone. Nothing can be added or removed. `TxGraph` is combined with a `ChainOracle` implementation to know whether a transaction is confirmed or not. When a transaction gets reorged out (the block which the tx is anchored to is no longer in the best chain), BDK considered it unconfirmed. If a conflicting transaction with a more recent `last_seen` gets introduced, the prior transaction is considered evicted.

LLFourn:
> Currently, the nature of `TxGraph` is fully monotone. Nothing can be added or removed.

I think things can be added!

evanlinjin:
> > Currently, the nature of `TxGraph` is fully monotone. Nothing can be added or removed.
> 
> I think things can be added!

Sorry *nothing can be removed 😅

PR #1338, opened 2024-02-10
Redo tx test

### LINKED ISSUE #107: The `insert_txouts` test in `test_tx_graph.rs` needs redoing.
**Describe the bug**  

https://github.com/bitcoindevkit/bdk/blob/7aca88474ac241bc0938ff989026463bdaf8f3ab/crates/chain/tests/test_tx_graph.rs#L16-L17

This test misunderstands `Anchor`s. There is no such thing as an "unconfirmed anchor", and `ChainPosition` does not implement `Anchor`. We should rewrite this test to be comprehensive.

**Proposed test scenarios**

- [ ] Insert tx with txid `A`, then insert partial txout(s) of the same txid. Expect: No changes to `TxGraph` (full tx should still exist in graph). Returned changeset is empty.
- [ ] Insert partial txout with txid `B`. Then insert full tx with txid `B`. Expect: the full tx should replace the partial tx. Check graph and returned changeset to confirm this.
- [ ] Do the above two tests with different counts of partial txids.

### LINKED ISSUE #1331: The `insert_txouts` test in `test_tx_graph.rs` does not make any sense.
**Describe the bug**  

https://github.com/bitcoindevkit/bdk/blob/7aca88474ac241bc0938ff989026463bdaf8f3ab/crates/chain/tests/test_tx_graph.rs#L16-L17

This test misunderstands `Anchor`s. There is no such thing as an "unconfirmed anchor", and `ChainPosition` does not implement `Anchor`. We should rewrite this test to be comprehensive.

**Proposed test scenarios**

- [ ] Insert tx with txid `A`, then insert partial txout(s) of the same txid. Expect: No changes to `TxGraph` (full tx should still exist in graph). Returned changeset is empty.
- [ ] Insert partial txout with txid `B`. Then insert full tx with txid `B`. Expect: the full tx should replace the partial tx. Check graph and returned changeset to confirm this.
- [ ] Do the above two tests with different counts of partial txids.

yanganto:
### Description
Fix bitcoindevkit/bdk_wallet#107

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/tests/test_tx_graph.rs
+++ b/crates/chain/tests/test_tx_graph.rs
@@ -11,192 +11,211 @@ use bitcoin::{
     absolute, hashes::Hash, BlockHash, OutPoint, ScriptBuf, Transaction, TxIn, TxOut, Txid,
 };
 use core::iter;
+use std::str::FromStr;
 use std::vec;
 
 #[test]
+fn insert_partial_tx() {
+    let mut graph = TxGraph::<ChainPosition<BlockId>>::default();
+    let _ = graph.insert_txout(
+        OutPoint::new(h!("txA"), 1),
+        TxOut {
+            value: 10_000,
+            script_pubkey: ScriptBuf::new(),
+        },
+    );
+
+    assert_eq!(graph.full_txs().collect::<Vec<_>>().len(), 0);
+
+    let changeset = graph.initial_changeset();
+    assert_eq!(changeset.txouts.len(), 1);
+    assert_eq!(
+        changeset.txouts().next(),
+        Some((
+            OutPoint {
+                txid: Txid::from_str(
+                    "0619eff85fd33f55f61af87ee1b612d3e2a26d1aed5f130e9bc74e6d0928d643"
+                )
+                .unwrap(),
+                vout: 1
+            },
+            &TxOut {
+                value: 10_000,
+                script_pubkey: ScriptBuf::new()
+            }
+        ))
+    );
+}
+
+#[test]
+fn insert_full_tx_after_partial_tx_inserted() {
+    let mut graph = TxGraph::<ChainPosition<BlockId>>::default();
+    let op = OutPoint::new(h!("txB"), 1);
+    let _ = graph.insert_txout(
+        op,
+        TxOut {
+            value: 10_000,
+            script_pubkey: ScriptBuf::new(),
+        },
+    );
+    assert_eq!(graph.full_txs().collect::<Vec<_>>().len(), 0);
+
+    let _ = graph.insert_tx(Transaction {
+        version: 0x01,
+        lock_time: absolute::LockTime::ZERO,
+        input: vec![TxIn {
+            previous_output: op,
+            ..Default::default()
+        }],
+        output: vec![TxOut {
+            value: 10_000,
+            script_pubkey: ScriptBuf::new(),
+        }],
+    });
+
+    assert_eq!(graph.full_txs().collect::<Vec<_>>().len(), 1);
+
+    let changeset = graph.initial_changeset();
+    assert_eq!(changeset.txouts.len(), 1);
+    assert_eq!(
+        changeset.txouts().next(),
+        Some((
+            OutPoint {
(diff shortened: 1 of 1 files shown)


crates/chain/src/chain_data.rs
```diff
@@ -30,16 +30,6 @@ impl<A: Clone> ChainPosition<&A> {
...
-impl<A: Anchor> ChainPosition<A> {
-    /// Determines the upper bound of the confirmation height.
-    pub fn confirmation_height_upper_bound(&self) -> Option<u32> {
-        match self {
-            ChainPosition::Confirmed(a) => Some(a.confirmation_height_upper_bound()),
-            ChainPosition::Unconfirmed(_) => None,
-        }
-    }
-}
-
```
evanlinjin: Why is this removed?

evanlinjin: Looking at the commit message for a7cbeab734361126957e7f5466a494a35dd8e48e, it seems you have gotten confused. We are not implementing `Anchor` for `ChainPosition` here, we are requiring that `A: Anchor` so that method `confirmation_height_upper_bound` is available.

yanganto: My bad.

evanlinjin: Np, thanks for the correction!

crates/chain/tests/test_tx_graph.rs
```diff
@@ -11,192 +11,211 @@ use bitcoin::{
...
+    let changeset = graph.initial_changeset();
+    assert_eq!(changeset.txouts.len(), 1);
+    assert_eq!(
+        changeset.txouts().next(),
+        Some((
+            OutPoint {
+                txid: Txid::from_str(
+                    "0619eff85fd33f55f61af87ee1b612d3e2a26d1aed5f130e9bc74e6d0928d643"
+                )
+                .unwrap(),
```
evanlinjin, line 39: nit: can we use `h!("txA")` here? Or even better, have a variable for the outpoint and txout.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -11,192 +11,211 @@ use bitcoin::{
...
+            }
+        ))
+    );
+}
 
-    // Another (OutPoint, TxOut) tuple to be used as update as partial transaction.
-    let update_ops = [(
-        OutPoint::new(h!("tx2"), 0),
+#[test]
+fn insert_full_tx_after_partial_tx_inserted() {
```
evanlinjin, line 51: This test is implemented differently than what is stated in the ticket. We want to reintroduce the partial tx as a full tx. In other words, the partial tx should have the same txid as the full tx.

evanlinjin: Also we are missing the test where we insert a partial tx after a full tx.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -11,192 +11,211 @@ use bitcoin::{
...
-    assert_eq!(
-        changeset,
-        ChangeSet {
-            txs: [update_txs.clone()].into(),
-            txouts: update_ops.clone().into(),
-            anchors: [(conf_anchor, update_txs.txid()), (unconf_anchor, h!("tx2"))].into(),
-            last_seen: [(h!("tx2"), 1000000)].into()
-        }
+#[test]
+fn insert_txouts() {
```
evanlinjin, line 99: I really meant for this to be part of `insert_full_tx_after_partial_tx` and `insert_partial_tx_after_full_tx` (not implemented yet). I.e. as scenarios that we loop through.

evanlinjin (changes requested):
Thanks for the work on this! It's good to see `ChainPosition` removed as an anchor.

However, please go over the ticket before marking it for review. It seems you have missed a couple of things.

Additionally, can you add docstr comments for each test? I think it's handy to have.

evanlinjin:
Tests already fixed in bitcoindevkit/bdk#1569

PR #1341, opened 2024-02-12
fix!(chain): Make spk revelation return owned spks

LLFourn:
Unfortunately there's no way to return a lifetime to something from a `&mut self` that doesn't extend the lifetime of the mutable borrow. So it's better not borrow at all here.

I think @evanlinjin tried to make this change some time ago but I NACK'd it. Now I understand why it's unavoidable.  Sorry!

### DIFF
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -5,7 +5,8 @@ use crate::{
     spk_iter::BIP32_MAX_INDEX,
     SpkIterator, SpkTxOutIndex,
 };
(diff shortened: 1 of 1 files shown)


notmandatory:
This needs a rebase but otherwise looks good to me.

oleonardolima (commented):
Concept ACK

notmandatory:
@LLFourn  is it OK if someone else takes this over and rebases it?

ValuedMammal:
I think it can be done in #1438

notmandatory:
Closing this PR as equivalent fixes are included in #1451.  a5142decfcd0581bf871755b8130782666329017

PR #1345, opened 2024-02-15
fix: remove deprecated max_satisfaction_weight

### LINKED ISSUE #1036: Use `max_weight_to_satisfy` instead of `max_satisfaction_weight`

storopoli:
### Description
Continuation of #1115.
Closes #1036.

* Change deprecated `max_satisfaction_weight` to `max_weight_to_satisfy`
* Remove `#[allow(deprecated)]` flags

### Notes to the reviewers

I've changed all `max_satisfaction_weight()` to `max_weight_to_satisfy()` in `Wallet.get_available_utxo()` and `Wallet.build_fee_bump()`. Checking the docs on the `miniscript` crate for `max_weight_to_satisfy` has the following note:

We are testing if the underlying descriptor `is.segwit()` or `.is_taproot`,
then adding 4WU if true or leaving as it is otherwise.

Another thing, we are not testing in BDK tests for legacy (pre-segwit) descriptors.
Should I also add them to this PR?

### Fixed
Replace the deprecated `max_satisfaction_weight` from `rust-miniscript` to `max_weight_to_satisfy`.

#### All Submissions:
* [x]  I've signed all my commits
* [x]  I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x]  I ran `cargo fmt` and `cargo clippy` before committing

#### Bugfixes:
* [ ]  This pull request breaks the existing API
* [ ]  I've added tests to reproduce the issue which are now passing
* [x]  I'm linking the issue being fixed by this PR

### DIFF
--- a/crates/bdk/src/wallet/coin_selection.rs
+++ b/crates/bdk/src/wallet/coin_selection.rs
@@ -32,7 +32,6 @@
 //! # use bdk::*;
 //! # use bdk::wallet::coin_selection::decide_change;
 //! # use anyhow::Error;
-//! # const TXIN_BASE_WEIGHT: usize = (32 + 4 + 4) * 4;
 //! #[derive(Debug)]
 //! struct AlwaysSpendEverything;
 //!
@@ -55,7 +54,8 @@
 //!                 |(selected_amount, additional_weight), weighted_utxo| {
 //!                     **selected_amount += weighted_utxo.utxo.txout().value;
 //!                     **additional_weight += Weight::from_wu(
-//!                         (TXIN_BASE_WEIGHT + weighted_utxo.satisfaction_weight) as u64,
+//!                         (TxIn::default().segwit_weight() + weighted_utxo.satisfaction_weight)
+//!                             as u64,
 //!                     );
 //!                     Some(weighted_utxo.utxo)
 //!                 },
@@ -109,6 +109,7 @@ use bitcoin::FeeRate;
 use alloc::vec::Vec;
 use bitcoin::consensus::encode::serialize;
 use bitcoin::OutPoint;
+use bitcoin::TxIn;
 use bitcoin::{Script, Weight};
 
 use core::convert::TryInto;
@@ -118,10 +119,6 @@ use rand::seq::SliceRandom;
 /// Default coin selection algorithm used by [`TxBuilder`](super::tx_builder::TxBuilder) if not
 /// overridden
 pub type DefaultCoinSelectionAlgorithm = BranchAndBoundCoinSelection;
-
-// Base weight of a Txin, not counting the weight needed for satisfying it.
-// prev_txid (32 bytes) + prev_vout (4 bytes) + sequence (4 bytes)
-pub(crate) const TXIN_BASE_WEIGHT: usize = (32 + 4 + 4) * 4;
 
 /// Errors that can be thrown by the [`coin_selection`](crate::wallet::coin_selection) module
 #[derive(Debug)]
@@ -347,10 +344,10 @@ fn select_sorted_utxos(
                 if must_use || **selected_amount < target_amount + **fee_amount {
                     **fee_amount += (fee_rate
                         * Weight::from_wu(
-                            (TXIN_BASE_WEIGHT + weighted_utxo.satisfaction_weight) as u64,
+                            (TxIn::default().segwit_weight() + weighted_utxo.satisfaction_weight)
+                                as u64,
                         ))
                     .to_sat();
-
                     **selected_amount += weighted_utxo.utxo.txout().value;
                     Some(weighted_utxo.utxo)
                 } else {
@@ -392,9 +389,10 @@ struct OutputGroup {
 impl OutputGroup {
     fn new(weighted_utxo: WeightedUtxo, fee_rate: FeeRate) -> Self {
         let fee = (fee_rate
-            * Weight::from_wu((TXIN_BASE_WEIGHT + weighted_utxo.satisfaction_weight) as u64))
+            * Weight::from_wu(
+                (TxIn::default().segwit_weight() + weighted_utxo.satisfaction_weight) as u64,
+            ))
         .to_sat();
-
         let effective_value = weighted_utxo.utxo.txout().value as i64 - fee as i64;
         OutputGroup {
             weighted_utxo,
@@ -744,7 +742,7 @@ mod test {
     use core::str::FromStr;
 
     use bdk_chain::ConfirmationTime;
-    use bitcoin::{Amount, OutPoint, ScriptBuf, TxOut};
+    use bitcoin::{Amount, OutPoint, ScriptBuf, TxIn, TxOut};
 
     use super::*;
     use crate::types::*;
@@ -754,9 +752,9 @@ mod test {
     use rand::seq::SliceRandom;
     use rand::{Rng, RngCore, SeedableRng};
 
-    // n. of items on witness (1WU) + signature len (1WU) + signature and sighash (72WU)
-    // + pubkey len (1WU) + pubkey (33WU) + script sig len (1 byte, 4WU)
-    const P2WPKH_SATISFACTION_SIZE: usize = 1 + 1 + 72 + 1 + 33 + 4;
+    // signature len (1WU) + signature and sighash (72WU)
+    // + pubkey len (1WU) + pubkey (33WU)
+    const P2WPKH_SATISFACTION_SIZE: usize = 1 + 72 + 1 + 33;
 
     const FEE_AMOUNT: u64 = 50;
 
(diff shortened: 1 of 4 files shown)


kaiwolfram (commented):
‼️‼️ 3️⃣✖️🇸 🇦 🇲 🇪 🎌 🇨 🇴 🇩 🇪 ‼️‼️
🫴4️⃣🤔

crates/bdk/src/wallet/mod.rs
```diff
@@ -2045,13 +2050,18 @@ impl<D> Wallet<D> {
...
-                #[allow(deprecated)]
-                (
-                    utxo,
+                (utxo, {
+                    let is_segwit = self.get_descriptor_for_keychain(keychain).is_witness()
+                        || self.get_descriptor_for_keychain(keychain).is_taproot();
+                    let segwit_add = match is_segwit {
+                        true => 4,
+                        false => 0,
+                    };
```
evanlinjin: Is the intention here to include the `scriptSigLen` weight? If so, please name the variable appropriately.

However, non-segwit inputs also have this field. Wouldn't it be correct to always have this addition?

evanlinjin (changes requested):
Thank you for having a go at this, however the approach here is incorrect.

First off, it's incorrect to add 4WU to only segwit inputs.

Looking at the docs of [`max_weight_to_satisfy`](https://docs.rs/miniscript/latest/miniscript/descriptor/enum.Descriptor.html#method.max_weight_to_satisfy), we need to change the definition of our `TX_IN_BASE_WEIGHT` to be `TxIn::default().segwit_weight()`.

After that, to make the tests pass, we need to change `P2WPKH_SATISFACTION_SIZE` to take in to account the `scriptSigLen` and `witnessLen`.

I think it's okay to assume all transactions are segwit transactions for now? Unless if someone can figure out how to take away the witness len elegantly (for non-segwit txs).

crates/bdk/src/wallet/mod.rs
evanlinjin: Ignore the comment above. Just remove this and add the extra weights to `TX_IN_BASE_WEIGHT`.

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -120,8 +120,9 @@ use rand::seq::SliceRandom;
 pub type DefaultCoinSelectionAlgorithm = BranchAndBoundCoinSelection;
 
 // Base weight of a Txin, not counting the weight needed for satisfying it.
-// prev_txid (32 bytes) + prev_vout (4 bytes) + sequence (4 bytes)
-pub(crate) const TXIN_BASE_WEIGHT: usize = (32 + 4 + 4) * 4;
+// prev_txid (32 bytes) + prev_vout (4 bytes) + sequence (4 bytes) + script_len
+// (1 byte).
+pub(crate) const TXIN_BASE_WEIGHT: usize = (32 + 4 + 4 + 1) * 4;
```
storopoli: @evanlinjin You mean something like this?

evanlinjin: You're missing the witness length

evanlinjin:
Can I please have some input on my statement here? Thanks!

> I think it's okay to assume all transactions are segwit transactions for now? Unless if someone can figure out how to take away the witness len elegantly (for non-segwit txs).

crates/bdk/src/wallet/mod.rs
storopoli: Great, sorry for the missundestanding.
I think I've fixed this in 52d6283ecaa343980dc8ef2da2ee3b0b28aaa45e

storopoli:
> I think it's okay to assume all transactions are segwit transactions for now? Unless if someone can figure out how to take away the witness len elegantly (for non-segwit txs).

I agree. We currently don't test any non-segwit tx anyways...

evanlinjin:
> I think it's okay to assume all transactions are segwit transactions for now? Unless if someone can figure out how to take away the witness len elegantly (for non-segwit txs).

Because we already assume this:

https://github.com/bitcoindevkit/bdk/blob/master/crates%2Fbdk%2Fsrc%2Fwallet%2Fmod.rs#L1505-L1513

So I think it is okay.

evanlinjin:
@storopoli can we please rebase the two commits into one and I'll ACK it.

storopoli:
> @storopoli can we please rebase the two commits into one and I'll ACK it.

Done, thanks mate!

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -318,8 +319,19 @@ impl<'a, D, Cs, Ctx> TxBuilder<'a, D, Cs, Ctx> {
...
+                        .is_witness()
+                        || wallet
+                            .get_descriptor_for_keychain(utxo.keychain)
+                            .is_taproot();
+                    let segwit_add = match is_segwit {
+                        true => 4,
+                        false => 0,
+                    };
+                    descriptor.max_weight_to_satisfy().unwrap() + segwit_add
+                };
```
evanlinjin: This is still wrong.

evanlinjin (changes requested):

crates/bdk/src/wallet/tx_builder.rs
storopoli: Sorry about that. Should be good now.

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -749,9 +747,9 @@ mod test {
     use rand::seq::SliceRandom;
     use rand::{Rng, RngCore, SeedableRng};
 
-    // n. of items on witness (1WU) + signature len (1WU) + signature and sighash (72WU)
-    // + pubkey len (1WU) + pubkey (33WU) + script sig len (1 byte, 4WU)
-    const P2WPKH_SATISFACTION_SIZE: usize = 1 + 1 + 72 + 1 + 33 + 4;
+    // n. of items on witness (1WU) signature and sighash (72WU)
+    // + pubkey len (1WU) + pubkey (33WU)
+    const P2WPKH_SATISFACTION_SIZE: usize = 1 + 72 + 1 + 33;
```
ValuedMammal: I think our accounting is right but just for clarity I would keep signature len in the above comment and drop witness length varint - which should already be included in [segwit_weight](https://docs.rs/bitcoin/0.30.2/bitcoin/blockdata/transaction/struct.TxIn.html#method.segwit_weight).

```suggestion
    // signature len (1WU) + signature and sighash (72WU)
    // + pubkey len (1WU) + pubkey (33WU)
    const P2WPKH_SATISFACTION_SIZE: usize = 1 + 72 + 1 + 33;
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -54,7 +54,6 @@ pub(crate) mod utils;
 pub mod error;
 pub use utils::IsDust;
 
-#[allow(deprecated)]
```
ValuedMammal, line 57: ❓ I'm actually not sure why this was here before

ValuedMammal (commented):
Approach ACK.

For simplicity I think it's fair to assume segwit weight everywhere. Unless someone is doing legacy transactions with hundreds of inputs (which is certainly possible), I don't think you'd notice a difference. We could consider carving out a special case for legacy-only transactions in a future iteration of tx building.

To the note on testing legacy descriptors I don't have an immediate comment on it.

crates/bdk/src/wallet/mod.rs
storopoli: Me neither, but always glad to nuke it out of the codebase.
More strict lints checks are always welcomed...

crates/bdk/src/wallet/coin_selection.rs
storopoli: I like this. I will add the commit and then squash it later

crates/bdk/src/wallet/mod.rs
storopoli: OOOOOO sorry. Yes, this was here because we were using the deprecated `rust-miniscript`'s `max_satisfaction_weight`.

That's the whole point of the PR 😂

evanlinjin:
@storopoli sorry can you do a rebase on master? thanks

storopoli:
> @storopoli sorry can you do a rebase on master? thanks

Don't worry man. Sure, rebased to master :)

evanlinjin (approved):
ACK 798ed8ced25156049126645435127a22245e916f

PR #1349, opened 2024-02-16
Fix `KeychainTxOutIndex::lookahead_to_target`

evanlinjin:
### Description

This method was not used (so it was untested) and it was not working. This fixes it.

The old implementation used `.next_store_index` which returned the keychain's last index stored in `.inner` (which include lookahead spks). This is WRONG because `.replenish_lookahead` needs the difference from last revealed.

### Changelog notice

Fix `KeychainTxOutIndex::lookahead_to_target`

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/tests/test_keychain_txout_index.rs
+++ b/crates/chain/tests/test_keychain_txout_index.rs
@@ -386,3 +386,98 @@ fn test_non_wildcard_derivations() {
         1,
     );
 }
+
+/// Check that calling `lookahead_to_target` stores the expected spks.
+#[test]
+fn lookahead_to_target() {
+    #[derive(Default)]
+    struct TestCase {
+        lookahead: u32,                      // Global lookahead value
+        external_last_revealed: Option<u32>, // Last revealed index for external keychain
+        internal_last_revealed: Option<u32>, // Last revealed index for internal keychain
+        external_target: Option<u32>,        // Call `lookahead_to_target(External, u32)`
+        internal_target: Option<u32>,        // Call `lookahead_to_target(Internal, u32)`
+    }
+
+    let test_cases = &[
+        TestCase {
+            lookahead: 0,
+            external_target: Some(100),
+            ..Default::default()
+        },
+        TestCase {
+            lookahead: 10,
+            internal_target: Some(99),
+            ..Default::default()
+        },
+        TestCase {
+            lookahead: 100,
+            internal_target: Some(9),
+            external_target: Some(10),
+            ..Default::default()
+        },
+        TestCase {
+            lookahead: 12,
+            external_last_revealed: Some(2),
+            internal_last_revealed: Some(2),
+            internal_target: Some(15),
+            external_target: Some(13),
+        },
+        TestCase {
+            lookahead: 13,
+            external_last_revealed: Some(100),
+            internal_last_revealed: Some(21),
+            internal_target: Some(120),
+            external_target: Some(130),
+        },
+    ];
+
+    for t in test_cases {
+        let (mut index, _, _) = init_txout_index(t.lookahead);
+
+        if let Some(last_revealed) = t.external_last_revealed {
+            let _ = index.reveal_to_target(&TestKeychain::External, last_revealed);
+        }
+        if let Some(last_revealed) = t.internal_last_revealed {
+            let _ = index.reveal_to_target(&TestKeychain::Internal, last_revealed);
+        }
+
+        let keychain_test_cases = [
+            (
+                TestKeychain::External,
+                t.external_last_revealed,
+                t.external_target,
+            ),
+            (
+                TestKeychain::Internal,
+                t.internal_last_revealed,
+                t.internal_target,
+            ),
+        ];
+        for (keychain, last_revealed, target) in keychain_test_cases {
+            if let Some(target) = target {
+                let original_last_stored_index = match last_revealed {
+                    Some(last_revealed) => Some(last_revealed + t.lookahead),
+                    None => t.lookahead.checked_sub(1),
+                };
+                let exp_last_stored_index = match original_last_stored_index {
+                    Some(original_last_stored_index) => {
+                        Ord::max(target, original_last_stored_index)
+                    }
+                    None => target,
+                };
+                index.lookahead_to_target(&keychain, target);
+                let keys = index
+                    .inner()
+                    .all_spks()
+                    .range((keychain.clone(), 0)..=(keychain.clone(), u32::MAX))
+                    .map(|(k, _)| k.clone())
+                    .collect::<Vec<_>>();
+                let exp_keys = core::iter::repeat(keychain)
+                    .zip(0_u32..=exp_last_stored_index)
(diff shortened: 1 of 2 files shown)


crates/chain/tests/test_keychain_txout_index.rs
```diff
@@ -386,3 +386,98 @@ fn test_non_wildcard_derivations() {
...
+        if let Some(last_revealed) = t.external_last_revealed {
+            let _ = index.reveal_to_target(&TestKeychain::External, last_revealed);
+        }
+
+        if let Some(last_revealed) = t.internal_last_revealed {
+            let _ = index.reveal_to_target(&TestKeychain::Internal, last_revealed);
+        }
+
+        if let Some(external_target) = t.external_target {
+            let normal_last = match t.external_last_revealed {
```
danielabrozzoni: We usually use "last" as in "last_revealed", but here it seems to be used as "last stored". If that's so, can we change this variable name?

crates/chain/tests/test_keychain_txout_index.rs
```diff
@@ -386,3 +386,98 @@ fn test_non_wildcard_derivations() {
...
+        }
+
+        if let Some(internal_target) = t.internal_target {
+            let normal_last = match t.internal_last_revealed {
+                Some(last_revealed) => Some(last_revealed + t.lookahead),
+                None => t.lookahead.checked_sub(1),
+            };
+
+            index.lookahead_to_target(&TestKeychain::Internal, internal_target);
+            let external_keys = index
```
danielabrozzoni: nit: should be `internal_keys`, `exp_internal_keys`

crates/chain/tests/test_keychain_txout_index.rs
```diff
@@ -386,3 +386,98 @@ fn test_non_wildcard_derivations() {
...
+                .inner()
+                .all_spks()
+                .range((TestKeychain::Internal, 0)..=(TestKeychain::Internal, u32::MAX))
+                .map(|(k, _)| k.clone())
+                .collect::<Vec<_>>();
+            let exp_external_keys = core::iter::repeat(TestKeychain::Internal)
+                .zip(0_u32..=Ord::max(internal_target, normal_last.unwrap_or(0)))
```
danielabrozzoni: I don't think this `normal_last.unwrap_or(0)` is ok. If you didn't have any spk stored, then exp_external_keys should have the index 0 in it, but in this case it would (you're ranging through 0..=0). 

I think you can just expect that `normal_last` is never going to be zero

danielabrozzoni (commented):
Unrelated, but going through `replenish_lookahead` I realized it could just use `self.next_index` instead of `let next_reveal_index = self.last_revealed.get(&descriptor_id).map_or(0, |v| *v + 1);`. Can you fix? Or I can open an issue if you prefer!

crates/chain/tests/test_keychain_txout_index.rs
evanlinjin: I think this is okay in terms of correctness (but it is hard to read). The smallest possible value of `internal_target` is 0, so `normal_last.unwrap(0)` being 0 will not affect anything.

`normal_last` is "what is normally the last stored value if `lookahead_to_target` was not called" - a value of `None` here means no spks were stored. `lookahead_to_target`'s input target_value is inclusive and the smallest possible input is 0, so after calling this method, the minimum resultant last_stored_index is 0.

However, I'll change it for the sake of readability.

evanlinjin:
> Unrelated, but going through `replenish_lookahead` I realized it could just use `self.next_index` instead of `let next_reveal_index = self.last_revealed.get(&descriptor_id).map_or(0, |v| *v + 1);`. Can you fix?

I don't mind adding to this PR (as it is small). However, I'm not sure what we are trying to fix here?

crates/chain/tests/test_keychain_txout_index.rs
```diff
@@ -386,3 +386,98 @@ fn test_non_wildcard_derivations() {
...
+/// Check that calling `lookahead_to_target` stores the expected spks.
+#[test]
+fn lookahead_to_target() {
+    #[derive(Default)]
+    struct TestCase {
+        lookahead: u32,                      // Global lookahead value
+        external_last_revealed: Option<u32>, // Last revealed index for external keychain
+        internal_last_revealed: Option<u32>, // Last revealed index for internal keychain
+        external_target: Option<u32>,        // Call `lookahead_to_target(External, u32)`
+        internal_target: Option<u32>,        // Call `lookahead_to_target(Internal, u32)`
```
storopoli: Why not move the comments to the field's docstrings?

```suggestion
fn lookahead_to_target() {
    #[derive(Default)]
    struct TestCase {
        /// Global lookahead value
        lookahead: u32,
        /// Last revealed index for external keychain
        external_last_revealed: Option<u32>,
        /// Last revealed index for internal keychain
        internal_last_revealed: Option<u32>,
        /// Call `lookahead_to_target(External, u32)`
        external_target: Option<u32>,
        /// Call `lookahead_to_target(Internal, u32)`
        internal_target: Option<u32>,
```

evanlinjin: It's a styling preference. Is there a clear advantage of having doc string here?

storopoli: No :)
Only if we ever document it.
But makes consistent with the docstring for the struct (the one you defined)

LLFourn:
Did we consider removing this method? Do we believe there is still a good reason to do lookaheads outside of the lookahead setting?

evanlinjin:
> Did we consider removing this method? Do we believe there is still a good reason to do lookaheads outside of the lookahead setting?

I'm currently using it for something 😅 (that's how I found the bug). To be fair, what I'm doing is pretty niche but I don't think it hurts to keep it?

crates/chain/tests/test_keychain_txout_index.rs
LLFourn: > It's a styling preference. Is there a clear advantage of having doc string here?

Yes it allows your editor to tell you the purpose of the field without having to jump to the definition.

+1 for using docstrings for this.

evanlinjin:
This is ready for review!

crates/chain/tests/test_keychain_txout_index.rs
notmandatory: I marked this as resolved since it looks like change was made.

crates/chain/tests/test_keychain_txout_index.rs
notmandatory: Also looks done so marked resolved.

crates/chain/tests/test_keychain_txout_index.rs
notmandatory: This was made more readable so marking as resolved.

notmandatory (approved):
ACK b290b2950221578d73c3194a1accfd080f83631a

The test framework was more complicated to review than the change! but good coverage.

PR #1351, opened 2024-02-16
fix: define and document `stop_gap`

### LINKED ISSUE #1227: Clearly define and document `stop_gap`
When reviewing #1225 I noticed that the definition of `stop_gap` we currently use is different from *the one I had in mind*. In any case, we should clearly define it and document it.

The PR fixes a bug and then adds a gap limit test. In the test, we generate 10 spks, and then we send sats on the 4th one (`spks[3]`). What is the minimum gap limit we should use in order to see the transaction?

The test currently doesn't find any transaction with `stop_gap = 2`, but does find a transaction with `stop_gap = 3`. I think this is wrong.
If we define the stop gap as "the maximum number of consecutive unused addresses" (as electrum does https://electrum.readthedocs.io/en/latest/faq.html#what-is-the-gap-limit) (and also btcpayserver, it seems? https://docs.btcpayserver.org/FAQ/Wallet/#the-gap-limit-problem), I think `stop_gap = 3` shouldn't find the tx, but `stop_gap = 4` should:

- With `stop_gap = 3`, we search `spk0`, `spk1`, `spk2`, none of this contain a tx, and since 3 is the maximum number of consecutive unused addresses, we stop
- With `stop_gap = 4` we search `spk0`, `spk1`, `spk2`, `spk3`, we find the transaction, and then we continue searching until we stop at `spk7` (included)

Also, with this definition, `stop_gap` should be always nonzero I think? As, if you use `stop_gap = 0` and check `spk0` and it doesn't have any tx, you have checked one unused address, and violated the definition (0 should be "the maximum number of consecutive unused addresses" ).

Does this make sense?

storopoli:
### Description

- changes the code implementation to "the maximum number of consecutive unused addresses" in esplora async and blocking extensions.
- for all purposes treats `stop_gap = 0` as `stop_gap = 1`.
- renames `past_gap_limit` to `gap_limit_reached` to indicate we want to break once the gap limit is reached and not go further in `full_scan`, as suggested in https://github.com/bitcoindevkit/bdk/issues/1227#issuecomment-1859040463
- change the tests according to the new implementation.
- add notes on what `stop_gap` means and links to convergent definition in other Bitcoin-related software.

Closes #1227.

### Notes to the reviewers

We can iterate over the wording and presentation of the `stop_gap` definition
and details.

### Changelog notice

- BREAKING: change `stop_gap` definition and effects in `full_scan`
  to reflect the common definitions in most Bitcoin-related software.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/esplora/src/async_ext.rs
+++ b/crates/esplora/src/async_ext.rs
@@ -52,6 +52,19 @@ pub trait EsploraAsyncExt {
     /// The full scan for each keychain stops after a gap of `stop_gap` script pubkeys with no associated
     /// transactions. `parallel_requests` specifies the max number of HTTP requests to make in
     /// parallel.
+    ///
+    /// ## Note
+    ///
+    /// `stop_gap` is defined as "the maximum number of consecutive unused addresses".
+    /// For example, with a `stop_gap` of  3, `full_scan` will keep scanning
+    /// until it encounters 3 consecutive script pubkeys with no associated transactions.
+    ///
+    /// This follows the same approach as other Bitcoin-related software,
+    /// such as [Electrum](https://electrum.readthedocs.io/en/latest/faq.html#what-is-the-gap-limit),
+    /// [BTCPay Server](https://docs.btcpayserver.org/FAQ/Wallet/#the-gap-limit-problem),
+    /// and [Sparrow](https://www.sparrowwallet.com/docs/faq.html#ive-restored-my-wallet-but-some-of-my-funds-are-missing).
+    ///
+    /// A `stop_gap` of 0 will be treated as a `stop_gap` of 1.
     async fn full_scan<K: Ord + Clone + Send>(
         &self,
         keychain_spks: BTreeMap<
@@ -162,6 +175,7 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
         let parallel_requests = Ord::max(parallel_requests, 1);
         let mut graph = TxGraph::<ConfirmationTimeHeightAnchor>::default();
         let mut last_active_indexes = BTreeMap::<K, u32>::new();
+        let stop_gap = Ord::max(stop_gap, 1);
 
         for (keychain, spks) in keychain_spks {
             let mut spks = spks.into_iter();
@@ -226,12 +240,12 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
                 }
 
                 let last_index = last_index.expect("Must be set since handles wasn't empty.");
-                let past_gap_limit = if let Some(i) = last_active_index {
-                    last_index > i.saturating_add(stop_gap as u32)
+                let gap_limit_reached = if let Some(i) = last_active_index {
+                    last_index >= i.saturating_add(stop_gap as u32)
                 } else {
-                    last_index >= stop_gap as u32
+                    last_index + 1 >= stop_gap as u32
                 };
-                if past_gap_limit {
+                if gap_limit_reached {
                     break;
                 }
             }
--- a/crates/esplora/src/blocking_ext.rs
+++ b/crates/esplora/src/blocking_ext.rs
@@ -50,6 +50,19 @@ pub trait EsploraExt {
     /// The full scan for each keychain stops after a gap of `stop_gap` script pubkeys with no associated
     /// transactions. `parallel_requests` specifies the max number of HTTP requests to make in
     /// parallel.
+    ///
+    /// ## Note
+    ///
+    /// `stop_gap` is defined as "the maximum number of consecutive unused addresses".
+    /// For example, with a `stop_gap` of  3, `full_scan` will keep scanning
+    /// until it encounters 3 consecutive script pubkeys with no associated transactions.
+    ///
+    /// This follows the same approach as other Bitcoin-related software,
+    /// such as [Electrum](https://electrum.readthedocs.io/en/latest/faq.html#what-is-the-gap-limit),
+    /// [BTCPay Server](https://docs.btcpayserver.org/FAQ/Wallet/#the-gap-limit-problem),
+    /// and [Sparrow](https://www.sparrowwallet.com/docs/faq.html#ive-restored-my-wallet-but-some-of-my-funds-are-missing).
+    ///
+    /// A `stop_gap` of 0 will be treated as a `stop_gap` of 1.
     fn full_scan<K: Ord + Clone>(
         &self,
         keychain_spks: BTreeMap<K, impl IntoIterator<Item = (u32, ScriptBuf)>>,
@@ -149,6 +162,7 @@ impl EsploraExt for esplora_client::BlockingClient {
         let parallel_requests = Ord::max(parallel_requests, 1);
         let mut graph = TxGraph::<ConfirmationTimeHeightAnchor>::default();
         let mut last_active_indexes = BTreeMap::<K, u32>::new();
+        let stop_gap = Ord::max(stop_gap, 1);
 
         for (keychain, spks) in keychain_spks {
             let mut spks = spks.into_iter();
@@ -216,12 +230,12 @@ impl EsploraExt for esplora_client::BlockingClient {
                 }
 
                 let last_index = last_index.expect("Must be set since handles wasn't empty.");
-                let past_gap_limit = if let Some(i) = last_active_index {
-                    last_index > i.saturating_add(stop_gap as u32)
+                let gap_limit_reached = if let Some(i) = last_active_index {
+                    last_index >= i.saturating_add(stop_gap as u32)
                 } else {
-                    last_index >= stop_gap as u32
+                    last_index + 1 >= stop_gap as u32
                 };
-                if past_gap_limit {
+                if gap_limit_reached {
                     break;
                 }
             }
--- a/crates/esplora/tests/async_ext.rs
+++ b/crates/esplora/tests/async_ext.rs
@@ -91,9 +91,9 @@ pub async fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
     Ok(())
 }
 
-/// Test the bounds of the address scan depending on the gap limit.
+/// Test the bounds of the address scan depending on the `stop_gap`.
 #[tokio::test]
-pub async fn test_async_update_tx_graph_gap_limit() -> anyhow::Result<()> {
+pub async fn test_async_update_tx_graph_stop_gap() -> anyhow::Result<()> {
     let env = TestEnv::new()?;
     let base_url = format!("http://{}", &env.electrsd.esplora_url.clone().unwrap());
     let client = Builder::new(base_url.as_str()).build_async()?;
@@ -140,12 +140,12 @@ pub async fn test_async_update_tx_graph_gap_limit() -> anyhow::Result<()> {
         sleep(Duration::from_millis(10))
     }
 
-    // A scan with a gap limit of 2 won't find the transaction, but a scan with a gap limit of 3
+    // A scan with a gap limit of 3 won't find the transaction, but a scan with a gap limit of 4
     // will.
-    let (graph_update, active_indices) = client.full_scan(keychains.clone(), 2, 1).await?;
+    let (graph_update, active_indices) = client.full_scan(keychains.clone(), 3, 1).await?;
     assert!(graph_update.full_txs().next().is_none());
     assert!(active_indices.is_empty());
-    let (graph_update, active_indices) = client.full_scan(keychains.clone(), 3, 1).await?;
+    let (graph_update, active_indices) = client.full_scan(keychains.clone(), 4, 1).await?;
     assert_eq!(graph_update.full_txs().next().unwrap().txid, txid_4th_addr);
     assert_eq!(active_indices[&0], 3);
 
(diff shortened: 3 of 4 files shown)


evanlinjin:
Thanks for this work.

I'm actually in favor of renaming the variable instead of changing the definition.

I think if `stop_gap` is renamed to `allowed_empty_history_count` it can avoid the confusion and we can avoid an API that panics on it being 0.

storopoli:
> I think if `stop_gap` is renamed to `allowed_empty_history_count` it can avoid the confusion and we can avoid an API that panics on it being 0.

I think that `stop_gap` is more widely used and will be a better fit for users without having to "reinvent the wheel".
There's no need to add more complexity for something that is de-facto a `stop_gap` like all major wallets and Bitcoin-related software defines.

We can discuss on how to handle `0`.

crates/esplora/src/blocking_ext.rs
```diff
@@ -216,12 +228,12 @@ impl EsploraExt for esplora_client::BlockingClient {
                 }
 
                 let last_index = last_index.expect("Must be set since handles wasn't empty.");
-                let past_gap_limit = if let Some(i) = last_active_index {
+                let gap_limit_reached = if let Some(i) = last_active_index {
                     last_index > i.saturating_add(stop_gap as u32)
```
ValuedMammal:
```suggestion
                    last_index >= i.saturating_add(stop_gap as u32)
```

Imagine 10 addresses and we send a tx to the 4th address and the 10th address (like in the test). We need to scan 6 times in order to find the tx at the 10th address (5, 6, 7, 8, 9, 10). So here, when `i` is 3 and the `last_index` scanned is 9, we have

`9 >= 3 + 6`

and we break. 

This means the second half of the test `test_update_tx_graph_gap_limit` should read

`// A scan with gap limit 5 won't find the second transaction, but a scan with gap limit 6 will.`

crates/esplora/src/blocking_ext.rs
```diff
@@ -145,6 +156,7 @@ impl EsploraExt for esplora_client::BlockingClient {
         stop_gap: usize,
         parallel_requests: usize,
     ) -> Result<(TxGraph<ConfirmationTimeHeightAnchor>, BTreeMap<K, u32>), Error> {
+        assert!(stop_gap > 0, "stop_gap must be higher than 0");
```
ValuedMammal: We should be able to handle being given a stop gap of 0 without blowing up.

ValuedMammal (commented):
I think the code in bdk_electrum suffers from the same stop gap bug. If so, we should try to fix that in this PR also, and probably have a test for it.

In terms of Changelog I consider this a bugfix but I wouldn't call it a breaking change, per se ?

danielabrozzoni:
I agree with José here, although `allowed_empty_history_count` would let us avoid the problem with `stop_gap` = 0, I don't think it's worth it to re-invent the wheel and introduce a new concept. We should stick with what users already know. 

I'm fine with both panicking or returning an error if the stop_gap is 0.

crates/esplora/src/blocking_ext.rs
storopoli: @notmandatory any ideas? To panic or to not panic?

evanlinjin:
I'll argue that `allowed_empty_count` is not a difficult concept to understand, so if it helps us avoid the `stop_gap=0` situation then it's very much worth it.

Returning an error when `stop_gap=0` is overdoing it imo. If we go with `stop_gap`, I much prefer a panic when it is 0.

crates/esplora/src/blocking_ext.rs
notmandatory: The way I think about the panics is if an app user (not dev) is able to set it to an invalid value then we should throw an error. I know wallets like [Sparrow](https://www.sparrowwallet.com/docs/faq.html#ive-restored-my-wallet-but-some-of-my-funds-are-missing) DO allow a user to manually set their stopgap when importing a wallet.  if we panic when a user sets it to 0 we're giving them a way to crash their app and that doesn't seem right. I'd prefer to only see panics on situations that can/should never happen due to end user interaction. I'd only expect a panic to catch some serious logic error found during unit or fuzz testing.

danielabrozzoni:
> I'll argue that allowed_empty_count is not a difficult concept to understand

You're right that it's not difficult to understand, but it's yet another thing to find out and learn about bdk. My intuition is that for a developer that wants to use bdk, there's already a lot to be learned, should we really add yet another concept, even if small?

> Returning an error when stop_gap=0 is overdoing it imo.

I would agree if we didn't already return a result, but since we do, it's probably not so bad to add a new error variant. But again, both error and panic are ok for me.

storopoli:
We can discuss Error/Panic on Tue's call

crates/esplora/src/blocking_ext.rs
storopoli: returning an Error now if `stop_gap == 0`

danielabrozzoni:
> I would agree if we didn't already return a result, but since we do, it's probably not so bad to add a new error variant. 

Gosh, I just realized we return a esplora error, and so we can't create a new error variant easily 😅

storopoli:
> > I would agree if we didn't already return a result, but since we do, it's probably not so bad to add a new error variant.
> 
> Gosh, I just realized we return a esplora error, and so we can't create a new error variant easily 😅

I have a stash that wraps the errors:

```rust
/// Custom error type for handling errors related to `esplora_client` operations
/// and additional conditions specific to our application logic, such as invalid parameters.
#[derive(Debug)]
enum EsploraError {
    /// Represents errors originating from the `esplora_client` crate.
    EsploraClientError(esplora_client::Error),
    /// Error to indicate an invalid `stop_gap` value, specifically when `stop_gap` is 0.
    InvalidStopGap,
}

impl From<esplora_client::Error> for EsploraError {
    fn from(err: esplora_client::Error) -> EsploraError {
        EsploraError::EsploraClientError(err)
    }
}

impl std::fmt::Display for EsploraError {
    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
        match *self {
            EsploraError::EsploraClientError(ref err) => write!(f, "Esplora client error: {}", err),
            EsploraError::InvalidStopGap => write!(f, "Invalid stop_gap: stop_gap cannot be 0"),
        }
    }
}

impl std::error::Error for EsploraError {}
```

Not beautiful but works 😂

evanlinjin:
Can we not create an error variant for this? It's more baggage which callers will have to deal with.

storopoli:
> Can we not create an error variant for this? It's more baggage which callers will have to deal with.

We cannot extend enums variants from other crates in Rust.
That's why I propose the wrapper above.

storopoli:
Discussed in 2024-02-27 BDK Lib Meeting:

- Not introduce an Error Variant
- Not panic
- Threat `stop_gap = 0` as `1` and document this thoroughly.

crates/esplora/src/blocking_ext.rs
```diff
@@ -50,6 +50,19 @@ pub trait EsploraExt {
     /// The full scan for each keychain stops after a gap of `stop_gap` script pubkeys with no associated
     /// transactions. `parallel_requests` specifies the max number of HTTP requests to make in
     /// parallel.
+    ///
+    /// ## Note
+    ///
+    /// `stop_gap` is defined as "the maximum number of consecutive unused addresses".
+    /// For example, with `stop_gap = 3`, `full_scan` will keep searching for script pubkeys
```
ValuedMammal:
```suggestion
    /// For example, with a `stop_gap` of  3, `full_scan` will keep scanning
```

crates/esplora/src/blocking_ext.rs
```diff
@@ -50,6 +50,19 @@ pub trait EsploraExt {
...
+    /// until it encounters 3 consecutive script pubkeys with no associated transactions.
+    ///
+    /// This follows the same approach as other Bitcoin-related software,
+    /// such as [Electrum](https://electrum.readthedocs.io/en/latest/faq.html#what-is-the-gap-limit),
+    /// [BTCPay Server](https://docs.btcpayserver.org/FAQ/Wallet/#the-gap-limit-problem),
+    /// and [Sparrow](https://www.sparrowwallet.com/docs/faq.html#ive-restored-my-wallet-but-some-of-my-funds-are-missing).
+    ///
+    /// `stop_gap = 0` will be treated as `stop_gap = 1` for all purposes.
```
ValuedMammal:
```suggestion
    /// A `stop_gap` of 0 will be treated as a `stop_gap` of 1.
```

I actually don't see a problem with the way stop_gap was already defined, but I guess it doesn't hurt to add clarifying details. Links to external sources seem like a nice addition.

My only suggestion is to use more plain words in place of pseudo expressions like `stop_gap = 0`.

crates/esplora/tests/blocking_ext.rs
```diff
@@ -239,14 +239,14 @@ pub fn test_update_tx_graph_gap_limit() -> anyhow::Result<()> {
         sleep(Duration::from_millis(10))
     }
 
-    // A scan with gap limit 4 won't find the second transaction, but a scan with gap limit 5 will.
+    // A scan with gap limit 4 won't find the second transaction, but a scan with gap limit 6 will.
     // The last active indice won't be updated in the first case but will in the second one.
     let (graph_update, active_indices) = env.client.full_scan(keychains.clone(), 4, 1)?;
```
ValuedMammal:
```suggestion
    // A scan with gap limit 5 won't find the second transaction, but a scan with gap limit 6 will.
    // The last active indice won't be updated in the first case but will in the second one.
    let (graph_update, active_indices) = env.client.full_scan(keychains.clone(), 5, 1)?;
```

nit:

If the new code is right, we will want to now see that a scan with gap limit 5 won't find the second transaction.

storopoli: Done in b9ba9b5978c2dfb1ee4cd29ed565365bd62ee75f

crates/esplora/src/blocking_ext.rs
storopoli: Done in b9ba9b5978c2dfb1ee4cd29ed565365bd62ee75f

crates/esplora/src/blocking_ext.rs
storopoli: Done in b9ba9b5978c2dfb1ee4cd29ed565365bd62ee75f

crates/esplora/src/blocking_ext.rs
storopoli: Resolved since we are treating `stop_gap` 0 as 1.

crates/esplora/src/blocking_ext.rs
danielabrozzoni: Hey, this was marked as resolved, but the code didn't really change. Unresolving.

crates/esplora/tests/blocking_ext.rs
```diff
@@ -214,12 +214,12 @@ pub fn test_update_tx_graph_gap_limit() -> anyhow::Result<()> {
         sleep(Duration::from_millis(10))
     }
 
-    // A scan with a gap limit of 2 won't find the transaction, but a scan with a gap limit of 3
+    // A scan with a stop_gap of 3 won't find the transaction, but a scan with a gap limit of 5
```
danielabrozzoni: This is wrong, it should be:
- stop_gap of 3 won't find
- stop_gap of 4 will

The test is passing right now only because the code is broken (see my above comment)

crates/esplora/tests/blocking_ext.rs
```diff
@@ -239,14 +239,14 @@ pub fn test_update_tx_graph_gap_limit() -> anyhow::Result<()> {
         sleep(Duration::from_millis(10))
     }
 
-    // A scan with gap limit 4 won't find the second transaction, but a scan with gap limit 5 will.
+    // A scan with gap limit 4 won't find the second transaction, but a scan with gap limit 6 will.
```
danielabrozzoni: Same as above: it makes no sense  to check with `stop_gap=n` and then with `stop_gap=n+2`. You want to check that the stop_gap is exactly n (in this case, 6), and that n-1 doesn't work. 

This should be:
stop_gap of 5: doesn't find
stop_gap of 6: finds

crates/esplora/tests/blocking_ext.rs
danielabrozzoni: This wasn't resolved, see my comment

danielabrozzoni (changes requested):
There are still some issues with your code, that were pointed out by other contributors but weren't resolved.

I don't want to be harsh, and I know it's not easy, but please try to put attention to details. Making sure to incorporate all the feedback that is given is important for speeding up review time.

storopoli:
@danielabrozzoni sorry about the oversight and thank you for pointing that out!
I have corrected the logic and math on `stop_gap` in a new commit

crates/esplora/tests/async_ext.rs
```diff
@@ -186,12 +186,14 @@ pub async fn test_async_update_tx_graph_gap_limit() -> anyhow::Result<()> {
...
+    // A scan with a gap limit of 3 won't find the transaction, but a scan with a gap limit of 4
     // will.
-    let (graph_update, active_indices) = env.client.full_scan(keychains.clone(), 2, 1).await?;
+    // FIXME: this part of the test is erroring
+    let (graph_update, active_indices) = env.client.full_scan(keychains.clone(), 3, 1).await?;
     assert!(graph_update.full_txs().next().is_none());
     assert!(active_indices.is_empty());
-    let (graph_update, active_indices) = env.client.full_scan(keychains.clone(), 3, 1).await?;
+    // NOTE: here onwards is OK
+    let (graph_update, active_indices) = env.client.full_scan(keychains.clone(), 4, 1).await?;
```
storopoli: Anyone knows why this is happening?
It is failing in:

```rust
assert!(graph_update.full_txs().next().is_none());
assert!(active_indices.is_empty());
```

everything else works...

crates/esplora/src/async_ext.rs
```diff
@@ -241,9 +241,9 @@ impl EsploraAsyncExt for esplora_client::AsyncClient {
 
                 let last_index = last_index.expect("Must be set since handles wasn't empty.");
                 let gap_limit_reached = if let Some(i) = last_active_index {
-                    last_index > i.saturating_add(stop_gap as u32)
+                    last_index >= i.saturating_add(stop_gap as u32)
                 } else {
-                    last_index + 1 >= stop_gap as u32
+                    last_index > stop_gap as u32
```
danielabrozzoni: In  08f04585e8bd900b9f88708b1193136480cd2061: this shouldn't have changed, `last_index + 1 >= stop_gap` was correct

storopoli: Yes you are absolutely right. I thought that I was simplifying the code but the expressions are not equal.
I think that all tests should pass now.
I've changed it back and amended in 95230d7d2b3d662277ec6a476451fd5ff4c3bd08

danielabrozzoni (commented):
Sweet, can you remove the FIXME and NOTE in the tests and squash the two commits together? Then it's ready to merge for me

danielabrozzoni (approved):
ACK 0c5220d0c46a305a17543359c30ed3c537b1d3b4

ValuedMammal (approved):
Just to note: the fix for electrum may be as simple as changing a `>` to `>=` but right now you kind of have to test manually to convince yourself that's the case

storopoli:
Rebased to `master`.
Cc @danielabrozzoni and @notmandatory.

notmandatory (approved):
ACK 7c1861aab9a76f2ba4a2ae372a16c9669a33f5de

PR #1360, opened 2024-02-25
feat(chain): create BlockHeight & UnixSeconds struct

15IITian:
bitcoindevkit/bdk_wallet#147

### DIFF
--- a/crates/chain/src/lib.rs
+++ b/crates/chain/src/lib.rs
@@ -21,6 +21,7 @@
 #![warn(missing_docs)]
 
 pub use bitcoin;
+pub use bitcoin::absolute::Height;
 mod spk_txout_index;
 pub use spk_txout_index::*;
 mod chain_data;
@@ -37,6 +38,8 @@ mod chain_oracle;
 pub use chain_oracle::*;
 mod persist;
 pub use persist::*;
+use std::fmt;
+use std::ops::Deref;
 
 #[doc(hidden)]
 pub mod example_utils;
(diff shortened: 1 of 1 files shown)


15IITian:
Please tell any changes required till so far?

15IITian:
image

I created this command to directly subtract height value from a variable of type BlockHeight ->
so should I create two commands which will give output in `u32` and `BlockHeight` respectively.
as in various places sometimes we need result in either types .

notmandatory:
This all looks good to me, be sure to fix the docs and clippy errors. Then also check where these new types should be used in the chain module code. Thanks!

15IITian:
> Then also check where these new types should be used in the chain module code

But IMO, I have to make crates other than `bdk-chain` to be compatible with these structs  as they are using `bdk-chain` structs in them.

crates/chain/src/lib.rs
```diff
@@ -97,5 +100,58 @@ pub mod collections {
...
+}
+impl Default for UnixSeconds {
+    fn default() -> Self {
+        UnixSeconds(0)
+    }
+}
+
+#[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone, PartialEq, Eq, Copy, PartialOrd, Ord, core::hash::Hash)]
+pub struct BlockHeight(Height);
```
evanlinjin, line 121: What is the rationale of having another type that wraps `bitcoin::absolute::Height`?

15IITian: > What is the rationale of having another type that wraps `bitcoin::absolute::Height`?

As Height struct in rust-bitcoin neither have `Default` trait implemention and  nor we can  implement it in our codebase.
Thus wrapped it .

15IITian:
https://github.com/bitcoindevkit/bdk/blob/d77a7f2ff13419d394e6a19f9942344a7c19a724/crates/chain/tests/test_local_chain.rs#L73

In the `test` folder of `bdk-chain` crate ->  
There is redundancy of  creating  `hash` and `height` of same values like this function has .
Suggestion: 
   * Either  in each function , we can define some varibles containing these values  and use them like this -> 
   
image

   (But still redundant at the function level)
  * Or make  const variables containing them.

crates/chain/src/lib.rs
evanlinjin: Thanks for the explanation, this probably requires a doc comment.

15IITian: Raised an issue for lack of this trait implementation  in `rust-bitcoin`

https://github.com/rust-bitcoin/rust-bitcoin/issues/2532

15IITian:
Once this  https://github.com/rust-bitcoin/rust-bitcoin/pull/2533 for adding Default trait implementation  in rust-bitcoin will  get successfully merged.
Both `locktime::absolute::Height` and  `locktime::absolute::Time` will have their default implementation but these changes will get reflected in their future releases. 
So Should we skip the idea of using `Height` struct  for now?

15IITian:
image

It would be solved after `Height` & `Time` structs will have their default implementation ,which IMO would be reflected in  their future releases.
So should we stop this work for now ?

notmandatory:
@15IITian yes it's fine to put this PR on hold until rust-bitcoin/rust-bitcoin#2533 is merged.

crates/chain/src/lib.rs
```diff
@@ -97,5 +100,58 @@ pub mod collections {
     pub use hashbrown::hash_map;
 }
 
+/// Number of seconds that have elapsed since 00:00:00 UTC on 1 January 1970, the Unix epoch.
+#[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone, PartialEq, Eq, Copy, PartialOrd, Ord, core::hash::Hash)]
+pub struct UnixSeconds(u64);
```
Kixunil, line 106: `bitcoin::absolute::Time` may be more appropriate.

crates/chain/src/lib.rs
```diff
@@ -97,5 +100,58 @@ pub mod collections {
...
+
+impl Default for BlockHeight {
+    fn default() -> Self {
+        return BlockHeight(
+            Height::from_consensus(0).expect("Failed to create a default BlockHeight"),
+        );
+    }
+}
+
+impl Deref for BlockHeight {
```
Kixunil, line 131: [This is an anti-pattern](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html)

crates/chain/src/lib.rs
15IITian: > `bitcoin::absolute::Time` may be more appropriate.


image

Kixunil: Well clearly the default is ambiguous so if there's a sensible default in your struct you should implement it manually and if there isn't you shouldn't implement it.

notmandatory: > `bitcoin::absolute::Time` may be more appropriate.

Good point, I didn't see `Time` and `Height` are both exposed under `bitcoin::absolute::`. Should certainly use those paths for `bdk_chain`.

notmandatory:
> @15IITian yes it's fine to put this PR on hold until [rust-bitcoin/rust-bitcoin#2533](https://github.com/rust-bitcoin/rust-bitcoin/pull/2533) is merged.

You should also be able to continue this PR using `bitcoin::absolute::Time` and `bitcoin::absolute::Height` without those types having implemented default values yet. You can adapt any places that expect defaults to use `Time::MIN` and `Height::MIN`. Then if/when these types get defaults it's only a small internal change that should not affect any of our APIs.

notmandatory:
@15IITian I created bitcoindevkit/bdk#1372 to switch the u64s for Time and I'll also replace the u32's with Height so I'm closing this PR.

PR #1361, opened 2024-02-25
docs: update docs to point to `derivation_of_spk`

### LINKED ISSUE #1042: The Wallet.is_mine() function should return K
**Describe the enhancement**  

The `Wallet.is_mine(&self, script: &[Script])` function currently returns a `bool` representing "whether or not a script is part of this wallet (either internal or external)". It would be more useful to know which `KeychainKind` the script was found for, or `None` if not one of my scripts:

```
pub fn is_mine(&self, script: &[Script]) -> Option<KeychainKind>
```

**Use case**  

A request came up on discord about how to figure out which output is the change. Having `is_mine()` tell you which chain a script is for would make this easy to figure out.

15IITian:
Updated docs  of `Wallet.is_mine()` function to point `derivation_of_spk` function to find whether a `KeychainKind` exits  for the given `script` or not.

fix #1042

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -815,6 +815,8 @@ impl<D> Wallet<D> {
     }
 
     /// Return whether or not a `script` is part of this wallet (either internal or external)
+    /// Use `derivation_of_spk` function to find the `KeychainKind` for the `script`, or
+    /// None if there's no  associated `Keychain`.
     pub fn is_mine(&self, script: &Script) -> bool {
         self.indexed_graph.index.index_of_spk(script).is_some()
     }


crates/bdk/src/wallet/mod.rs
```diff
@@ -815,6 +815,8 @@ impl<D> Wallet<D> {
     }
 
     /// Return whether or not a `script` is part of this wallet (either internal or external)
+    /// Use `derivation_of_spk` function to find the `KeychainKind` for the `script`, or
+    /// None if there's no  associated `Keychain`.
```
evanlinjin, line 819: Please follow rust doc conventions. Ref: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

We want a separate section for the more detailed description. Also, for clarity, I would reword it like so: "If you wish to find the `KeychainKind` for the given `script`, use `derivation_of_spk` instead."

In addition to this, please have a link for `derivation_of_spk`.

evanlinjin (changes requested):

crates/bdk/src/wallet/mod.rs
15IITian: Would it make sense to have  seperate section telling about the  `derivation_of_spk` in detail  and include that point ?

evanlinjin: If you want to expand on `derivation_of_spk`, it should be part of the docs on `derivation_of_spk`, not here.

evanlinjin:
Closing because stale.

PR #1366, opened 2024-03-04
Bump MSRV 1.63.0 AND add deprecate TxBuilder::allow_shrinking()

### LINKED ISSUE #1342: allow_shrinking not working as expected
Using [`build_fee_bump`](https://docs.rs/bdk/0.29.0/bdk/wallet/struct.Wallet.html#method.build_fee_bump) to replace transaction along with [`allow_shrinking`](https://docs.rs/bdk/0.29.0/bdk/wallet/tx_builder/struct.TxBuilder.html#method.allow_shrinking) results in the specified address output amount increasing and consuming all of the input utxo amount.

**To Reproduce** 
Create a transaction with one output ADDRESS and specify an x amount to be sent. sign and broadcast.
Sync wallet. Create a bump fee txbuilder and call allow_shrinking(ADDRESS) and specify a higher feerate fee_rate

Example Code:

```rust

fn create_original_tx(){
     let (wallet, blockchain) = init();
     wallet.sync(&blockchain, SyncOptions::default()).unwrap();
     let addr = bdk::bitcoin::Address::from_str("bcrt1q6m7tyf45hd5swf7ug5f3r35y6htga0m2tlxhfh").unwrap();
     println!("Descriptor balance: {} SAT", wallet.get_balance().unwrap());
     let script = addr.script_pubkey().clone();
     tx_builder
         .add_recipient(script.clone(), 10000)
         .fee_rate(FeeRate::from_sat_per_vb(1.0))
         .enable_rbf();
     let (mut psbt, _tx_details) = tx_builder.finish().unwrap();
     let  finalized = wallet.sign(&mut psbt, bdk::SignOptions::default()).unwrap();
     assert!(finalized, "we should have signed all the inputs");
     let tx = psbt.clone().extract_tx();
     blockchain.broadcast(&tx).unwrap();
     let txid = psbt.extract_tx().txid();
     println!("Txid:{:?}",txid );
 }

fn build_bump_psbt(){
     let (wallet, blockchain) = init();
     let addr = bdk::bitcoin::Address::from_str("bcrt1q6m7tyf45hd5swf7ug5f3r35y6htga0m2tlxhfh").unwrap();
     wallet.sync(&blockchain, SyncOptions::default()).unwrap();
     let mut tx_builder = wallet.build_fee_bump(Txid::from_str("f8f8d1e566eb2ae963825628e4a3ef249a82aa71b8bfb5389b5d935d8864eebe").unwrap()).unwrap();
     tx_builder
         .allow_shrinking(addr.script_pubkey()).unwrap()
         .fee_rate(FeeRate::from_sat_per_vb(2.5));
     let (mut psbt, _tx_details) = tx_builder.finish().unwrap();
     let  finalized = wallet.sign(&mut psbt, bdk::SignOptions::default()).unwrap();
     assert!(finalized, "we should have signed all the inputs");
     let tx = psbt.clone().extract_tx();
     blockchain.broadcast(&tx).unwrap();
     let txid = psbt.extract_tx().txid();
     println!("Txid2 :{:?}",txid );
 }

```
**Expected behavior**  
Expect the output ADDRESS to have the same amount or less. Amount will be less if the input transaction cannot cover the fee as well as the originally specified amount for ADDRESS

However  output ADDRESS is having a higher amount than the origincal. It is actually consuming all the input amount.

**Build environment**  
 - BDK v0.28.2
 - macOS 13.2
 - Rust/Cargo version: 1.76.0

**Additional context**  

Original transaction outputs
Screenshot 2024-02-09 at 19 31 09

Transaction outputs after fee bump and allow shrinking:
Screenshot 2024-02-09 at 19 32 05

notmandatory:
### Description

fixes #1342 

* bumps MSRV to 1.63
* fixed clippy errors for new MSRV
* add warning to docs for TxBuilder::allow_shrinking().
* add wallet test to confirm behavior of TxBuilder::allow_shrinking() when not also using drain_wallet().

### Notes to the reviewers

This is a quick fix, long term we should try to make this safer to use.

#### All Submissions:


notmandatory:
Rather than try to figure out how to get this to build with MSRV 1.57 I'm bumping it to 1.63.0. 

~~For some reason CI isn't running at all so I still working on that.~~

thunderbiscuit:
I think the docs now correctly describe what the method is doing, but it feels like a bug to me, and the docs now simply describe the bug.

The intended behaviour (as I imagine it at least) is what is described in the first line of the doc:
> Explicitly tells the wallet that it is allowed to reduce the amount of the output matching this `script_pubkey` in order to bump the transaction fee.

In other words, if you're bumping the fee on a transaction, you can tell the wallet to not apply default behaviour (use the change output to pay for the extra fees), but instead use one of your payees' output. If the extra fee required is 1000 sat, the output should simply shrink by 1000 sat, those sats go to the transaction fee, and nothing else about the transaction change. If the method ends up triggering a switcharoo of where change goes and now throws all change into that payees' output, it's so far off what I'd expect would happen from the name of the method that I'd consider it a bug.

If the current behaviour is useful in some way (I can't think of a use case but maybe there are?) we could rename the method to better describe what it does. Otherwise I think it would be better to fix the method to match what users might expect from an `allow_shrinking()` method, or disable it entirely if there is no bandwidth to fix at the moment.

notmandatory:
~~@thunderbiscuit I think you are right now that I'm looking at the code again it looks like coin selection WILL add more inputs to pay the fee so will drain the whole wallet to the `allow_shrinking()` address if enabled..  so either we need to remove allow_shrinking() or always turn on manually_selected_only() for RBF or for RBF when using `allow_shrinking()`.  I'll update my test to confirm.~~ Update, I was wrong about this, see below.

notmandatory:
I updated the test to confirm that given a wallet with 2 UTXOs, if the original TX only uses one UTXO then the new fee bumped TX with `allow_shrinking()` will also only use that one UTXO and NOT bring in the other available in the wallet. 

I think this function is doing what it says which is allow shrinking one output to pay for a new larger TX fee. The only scenario I can think of where this behavior would be unexpected is if you also manually add new inputs with your fee bump tx builder. Should I explicitly warn against this scenario? I'd hope that the warning would cover that scenario, I'd rather not change the function name for the 0.xx branch, I'd rather have that discussion for the 1.0 bdk wallet api.

LLFourn:
> I think the docs now correctly describe what the method is doing, but it feels like a bug to me, and the docs now simply describe the bug.

Yes it's a design bug IMO: https://github.com/bitcoindevkit/bdk/issues/1374

notmandatory:
How about we just deprecate `allow_shrinking()` in this release (and in corresponding bdk-ffi API) in case anyone is using it, and then remove it completely in the next release (and in the 1.0.0 api)? Then we have time to redesign txbuilder for 1.1 or 2.0 milestone.

.github/workflows/cont_integration.yml
```diff
@@ -18,9 +18,9 @@ jobs:
     strategy:
       matrix:
         rust:
-          - version: 1.65.0 # STABLE
+          - version: stable
```
danielabrozzoni, line 21: This means that if a new version of rust is released, with a new version of clippy with stricter rules, our ci would immediately fail the clippy check. Are we ok with this?

src/wallet/signer.rs
```diff
@@ -805,9 +805,10 @@ pub struct SignOptions {
 }
 
 /// Customize which taproot script-path leaves the signer should sign.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
 pub enum TapLeavesOptions {
     /// The signer will sign all the leaves it has a key for.
+    #[default]
```
danielabrozzoni, line 811: Ohh this is so cool!

danielabrozzoni (approved):
utACK 5bc00e44e6b0a7f83a878a03bcd0df5a8719e8ba - looks good to me, I only have one comment regarding the clippy ci

.github/workflows/cont_integration.yml
storopoli: I don't think so because of https://github.com/bitcoindevkit/bdk/blob/52f395555792a2f24ae7248d98da318bdad5dadb/clippy.toml#L1

Ref: https://doc.rust-lang.org/stable/clippy/configuration.html

danielabrozzoni: I think it means that clippy won't suggest any improvement that breaks msrv, not that clippy is pinned to version 1.63.0.

storopoli: From the Ref above:

> Projects that intend to support old versions of Rust can disable lints pertaining to newer features by specifying the minimum supported Rust version (MSRV) in the clippy configuration file.

So I think that you worry about

> if a new version of rust is released, with a new version of clippy with stricter rules, our ci would immediately fail the clippy check

would not fail right?

danielabrozzoni: I think it would fail.
Let's say that rust 1.80 introduces the method `pizza()`, and clippy 1.80 introduces two new rules:
1. `pizza()` should always be used instead of `format()`
2. Every variable name should start with a `x` (this makes your code more beautiful)

If your msrv is 1.79, clippy wouldn't complain about rule 1: you can't use `pizza()`, as it doesn't exist in rust 1.79! But I'm kind of sure that it would still warn you about rule 2, since it's doable in earlier versions of rust too. 

Anyways, I don't know how often clippy adds rules that don't pertain to new rust features, so it might not be a big deal after all.

storopoli: Yes but first MSRV CI would fail because you cannot include `pizza()` since it does not exist in MSRV.

And even if you don't include it the `clippy.toml` MSRV config will make sure to not enable the `pizza_instead of format` lint because of

> Projects that intend to support old versions of Rust can disable lints pertaining to newer features by specifying the minimum supported Rust version (MSRV) in the clippy configuration file.

danielabrozzoni: > Yes but first MSRV CI would fail because you cannot include pizza() since it does not exist in MSRV.

I never said I included pizza in my code. I said that clippy had that rule, but wouldn't enforce it since my MSRV wouldn't permit it. 

The point is not the rule1, it is rule2 that would make the CI fail.

storopoli: Rule 2 is a "newer feature" that I've mentioned above.
So CI with `clippy.toml` would pass.

danielabrozzoni: No, rule 2 is a new clippy lint that can be applied to also older versions of rust, not a new feature.
Anyways, I don't have the time to try this out, I'd say it's ok to merge and if you ever notice that clippy breaks when a new version is released, and it's too annoying to fix, you can change the ci again.

storopoli: I don't think that this is how the clippy lints are versioned.
Because if you run clippy with Rust 1.X.0 it means that you are running a snapshot of the codebase of rust, cargo, clippy, and rustfmt from the time that 1.X.0 was released.

I see you scenario happening with some backport like 1.X.1 or 1.X.2 etc.

.github/workflows/code_coverage.yml
```diff
@@ -21,19 +21,16 @@ jobs:
...
-      - name: Set default toolchain
-        run: rustup default nightly
-      - name: Set profile
-        run: rustup set profile minimal
-      - name: Add llvm tools
-        run: rustup component add llvm-tools-preview
-      - name: Update toolchain
-        run: rustup update
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
```
storopoli: Since we are touching this, we should be using https://github.com/dtolnay/rust-toolchain instead of the abandoned unmaintained `actions-rs`.

notmandatory:
I had to disable the code coverage and blockchain integration tests.  They used to work but not are failing even though there haven't been any code changes since they last run successfully. I think time is better spent on 1.0 than trouble shooting these old flaky tests.

PR #1368, opened 2024-03-05
docs: replace `LocalUtxo` to `LocalOutput`

### LINKED ISSUE #1249: Better names for methods
It looks like `LocalUtxo` was renamed to `LocalOutput` in master, which seems like a great improvement in clarity. I came across #665 trying to determine whether or not `Wallet::get_utxo` actually returns *unspent* outputs or not. I'd propose renaming the rest of the wallet API (and database API) to reflect this type change in 1.x

15IITian:
#### All Submissions:

 fixes #1249

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -2245,7 +2245,7 @@ impl<D> Wallet<D> {
         Ok(psbt)
     }
 
-    /// get the corresponding PSBT Input for a LocalUtxo
+    /// get the corresponding PSBT Input for a LocalOutput
     pub fn get_psbt_input(
         &self,
         utxo: LocalOutput,


CHANGELOG.md
```diff
@@ -205,7 +205,7 @@ BDK and LDK together.
 - `verify` flag removed from `TransactionDetails`.
 - Add `get_internal_address` to allow you to get internal addresses just as you get external addresses.
 - added `ensure_addresses_cached` to `Wallet` to let offline wallets load and cache addresses in their database
-- Add `is_spent` field to `LocalUtxo`; when we notice that a utxo has been spent we set `is_spent` field to true instead of deleting it from the db.
```
danielabrozzoni: I don't think we should be modifying the old changelog - at the time, when we published those versions, LocalOutput was indeed called LocalUtxo :) 

Can you please revert your changes on the CHANGELOG.md file?

15IITian: I got  your point .

notmandatory:
@Ademan is this what you had in mind for #1249 ? If so It's a very small and non-functional change and I'd prefer it be done as basic house keeping in some other more substantial PR.

15IITian:
>  It's a very small and non-functional change and I'd prefer it be done as basic house keeping in some other more substantial PR.

If so we should close it .

notmandatory:
Yes I'll close this PR and someone can fix it as housekeeping in another PR.

PR #1369, opened 2024-03-06
feat(chain): add `get` and `range` methods to `CheckPoint`

### LINKED ISSUE #1354: The `TxGraph::missing_heights` implementation is problematic.
The logic of `missing_heights` is explained clearly in this comment:

```rust
// Usually, if a height of a tx anchor is missing from the chain, we would want to return
// this height in the iterator. The exception is when the tx is confirmed in chain. All the
// other missing-height anchors of this tx can be skipped.
```

However, if a transaction is confirmed in a block that is about to turn stale (get reorged out), we
still want to fetch the new block hash. I suspect this is where our "transaction is never confirmed"
bug comes from.

**My proposed solution:**

Change the exception to:

```rust
// ... The exception is when the tx is confirmed in chain with a depth greater than the provided
// `max_reorg_depth`.
```

And we change the signature of the method to be:

```rust
pub fn missing_heights<'a>(
    &'a self,
    chain: &'a LocalChain,
    max_reorg_depth: u32,
) -> impl Iterator<Item = u32> + 'a { todo!() }
```

With some nice documentation for `max_reorg_depth`.

**An alternative solution:**

Remove the stated "exception". We can continue to consider a height "missing" when there is a
floating anchor at that height (anchor that does not point to a block in the best chain).

The downside of this approach is that there are situations where certain heights will always be
considered "missing". I think this is somewhat acceptable as reorgs are rare and this situation only
happens due to reorgs.

Because of the behavior change, if we do this solution, we should rename the method to
`floating_anchor_heights`.

**How to reproduce with `bdk_esplora`:**

The following should be written as a test.

* Wallet receives money via transaction (A).
* Transaction gets confirmed at confirmation height (h).
* (Sync wallet with `bdk_esplora`),
* A reorg happens, and (A) gets re-confirmed at height (h') where `h' >= h`.
* (Sync wallet with `bdk_esplora`).
* At this point onwards, transaction (A) is forever stuck as unconfirmed!

**Going forward:**

The test should be written first while we wait for the discussion on which solution to do matures.

evanlinjin:
Partially fixes #1354 

### Description

These methods allow us to query for checkpoints contained within the linked list by height and height range. This is useful to determine checkpoints to fetch for chain sources without having to refer back to the `LocalChain`.

Currently this is not implemented efficiently, but in the future, we will change `CheckPoint` to use a skip list structure.

### Notes to the reviewers

Please refer to the conversation in #1354 for more details. In summary, both `TxGraph::missing_heights` and `tx_graph::ChangeSet::missing_heights_from` are problematic in their own ways. Additionally, it's a better API for chain sources if we only need to lock our receiving structures twice (once to obtain initial state and once for applying the update). Instead of relying on methods such as `EsploraExt::update_local_chain` to get relevant checkpoints, we can use these query methods instead. This allows up to get rid of `::missing_heights`-esc methods and remove the need to lock receiving structures in the middle of the sync.

### Changelog notice

* Added `get` and `range` methods to `CheckPoint` (and in turn, `LocalChain`). This simulates an API where we have implemented a skip list of checkpoints (to implement in the future). This is a better API because we can query for any height or height range with just a checkpoint tip instead of relying on a separate checkpoint index (which needs to live in `LocalChain`).
* Changed `LocalChain` to have a faster `Eq` implementation. We now maintain an xor value of all checkpoint block hashes. We compare this xor value to determine whether two chains are equal.
* Added `PartialEq` implementation for `CheckPoint` and `local_chain::Update`.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -1127,18 +1127,14 @@ impl<D> Wallet<D> {
                 // anchor tx to checkpoint with lowest height that is >= position's height
                 let anchor = self
                     .chain
-                    .blocks()
                     .range(height..)
-                    .next()
+                    .last()
                     .ok_or(InsertTxError::ConfirmationHeightCannotBeGreaterThanTip {
                         tip_height: self.chain.tip().height(),
                         tx_height: height,
                     })
-                    .map(|(&anchor_height, &hash)| ConfirmationTimeHeightAnchor {
-                        anchor_block: BlockId {
-                            height: anchor_height,
-                            hash,
-                        },
+                    .map(|anchor_cp| ConfirmationTimeHeightAnchor {
+                        anchor_block: anchor_cp.block_id(),
                         confirmation_height: height,
                         confirmation_time: time,
                     })?;
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -246,17 +246,6 @@ pub struct LocalChain {
     tip: CheckPoint,
 }
 
-// TODO: Figure out whether we can get rid of this
-impl From<LocalChain> for BTreeMap<u32, BlockHash> {
-    fn from(value: LocalChain) -> Self {
-        value
-            .tip
-            .iter()
-            .map(|cp| (cp.height(), cp.hash()))
-            .collect()
-    }
-}
-
 impl ChainOracle for LocalChain {
     type Error = Infallible;
 
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -1,10 +1,12 @@
 //! The [`LocalChain`] is a local implementation of [`ChainOracle`].
 
 use core::convert::Infallible;
+use core::ops::RangeBounds;
 
 use crate::collections::BTreeMap;
 use crate::{BlockId, ChainOracle};
 use alloc::sync::Arc;
+use alloc::vec::Vec;
 use bitcoin::block::Header;
 use bitcoin::BlockHash;
 
@@ -148,6 +150,36 @@ impl CheckPoint {
     pub fn iter(&self) -> CheckPointIter {
         self.clone().into_iter()
     }
+
+    /// Get checkpoint at `height`.
+    ///
+    /// Returns `None` if checkpoint at `height` does not exist`.
+    pub fn get(&self, height: u32) -> Option<Self> {
+        self.range(height..=height).next()
+    }
+
+    /// Iterate checkpoints over a height range.
+    ///
+    /// Note that we always iterate checkpoints in reverse height order (iteration starts at tip
+    /// height).
+    pub fn range<R>(&self, range: R) -> impl Iterator<Item = CheckPoint>
+    where
+        R: RangeBounds<u32>,
+    {
+        let start_bound = range.start_bound().cloned();
+        let end_bound = range.end_bound().cloned();
+        self.iter()
+            .skip_while(move |cp| match end_bound {
+                core::ops::Bound::Included(inc_bound) => cp.height() > inc_bound,
+                core::ops::Bound::Excluded(exc_bound) => cp.height() >= exc_bound,
+                core::ops::Bound::Unbounded => false,
+            })
+            .take_while(move |cp| match start_bound {
+                core::ops::Bound::Included(inc_bound) => cp.height() >= inc_bound,
+                core::ops::Bound::Excluded(exc_bound) => cp.height() > exc_bound,
+                core::ops::Bound::Unbounded => true,
+            })
+    }
 }
 
 /// Iterates over checkpoints backwards.
@@ -205,18 +237,28 @@ pub struct Update {
 #[derive(Debug, Clone)]
 pub struct LocalChain {
     tip: CheckPoint,
-    index: BTreeMap<u32, BlockHash>,
 }
 
 impl PartialEq for LocalChain {
     fn eq(&self, other: &Self) -> bool {
-        self.index == other.index
+        self.iter_checkpoints()
+            .map(|cp| cp.block_id())
+            .collect::<Vec<_>>()
+            == other
+                .iter_checkpoints()
+                .map(|cp| cp.block_id())
+                .collect::<Vec<_>>()
     }
 }
 
+// TODO: Figure out whether we can get rid of this
 impl From<LocalChain> for BTreeMap<u32, BlockHash> {
     fn from(value: LocalChain) -> Self {
-        value.index
+        value
+            .tip
+            .iter()
+            .map(|cp| (cp.height(), cp.hash()))
+            .collect()
     }
 }
 
@@ -228,18 +270,16 @@ impl ChainOracle for LocalChain {
         block: BlockId,
         chain_tip: BlockId,
     ) -> Result<Option<bool>, Self::Error> {
-        if block.height > chain_tip.height {
-            return Ok(None);
+        let chain_tip_cp = match self.tip.get(chain_tip.height) {
+            // we can only determine whether `block` is in chain of `chain_tip` if `chain_tip` can
+            // be identified in chain
+            Some(cp) if cp.hash() == chain_tip.hash => cp,
+            _ => return Ok(None),
+        };
+        match chain_tip_cp.get(block.height) {
+            Some(cp) => Ok(Some(cp.hash() == block.hash)),
+            None => Ok(None),
         }
-        Ok(
-            match (
-                self.index.get(&block.height),
-                self.index.get(&chain_tip.height),
-            ) {
-                (Some(cp), Some(tip_cp)) => Some(*cp == block.hash && *tip_cp == chain_tip.hash),
-                _ => None,
-            },
-        )
     }
 
     fn get_chain_tip(&self) -> Result<BlockId, Self::Error> {
@@ -250,7 +290,7 @@ impl ChainOracle for LocalChain {
 impl LocalChain {
     /// Get the genesis hash.
     pub fn genesis_hash(&self) -> BlockHash {
-        self.index.get(&0).copied().expect("must have genesis hash")
+        self.tip.get(0).expect("genesis must exist").hash()
     }
 
     /// Construct [`LocalChain`] from genesis `hash`.
@@ -259,7 +299,6 @@ impl LocalChain {
         let height = 0;
         let chain = Self {
             tip: CheckPoint::new(BlockId { height, hash }),
-            index: core::iter::once((height, hash)).collect(),
         };
         let changeset = chain.initial_changeset();
         (chain, changeset)
@@ -276,7 +315,6 @@ impl LocalChain {
         let (mut chain, _) = Self::from_genesis_hash(genesis_hash);
         chain.apply_changeset(&changeset)?;
 
-        debug_assert!(chain._check_index_is_consistent_with_tip());
         debug_assert!(chain._check_changeset_is_applied(&changeset));
 
         Ok(chain)
@@ -284,18 +322,11 @@ impl LocalChain {
 
     /// Construct a [`LocalChain`] from a given `checkpoint` tip.
     pub fn from_tip(tip: CheckPoint) -> Result<Self, MissingGenesisError> {
-        let mut chain = Self {
-            tip,
-            index: BTreeMap::new(),
-        };
-        chain.reindex(0);
-
-        if chain.index.get(&0).copied().is_none() {
+        let genesis_cp = tip.iter().last().expect("must have at least one element");
+        if genesis_cp.height() != 0 {
             return Err(MissingGenesisError);
         }
-
-        debug_assert!(chain._check_index_is_consistent_with_tip());
-        Ok(chain)
+        Ok(Self { tip })
     }
 
     /// Constructs a [`LocalChain`] from a [`BTreeMap`] of height to [`BlockHash`].
@@ -303,12 +334,11 @@ impl LocalChain {
     /// The [`BTreeMap`] enforces the height order. However, the caller must ensure the blocks are
     /// all of the same chain.
     pub fn from_blocks(blocks: BTreeMap<u32, BlockHash>) -> Result<Self, MissingGenesisError> {
-        if !blocks.contains_key(&0) {
+        if blocks.get(&0).is_none() {
             return Err(MissingGenesisError);
         }
 
         let mut tip: Option<CheckPoint> = None;
-
         for block in &blocks {
             match tip {
                 Some(curr) => {
@@ -321,13 +351,9 @@ impl LocalChain {
             }
         }
 
-        let chain = Self {
-            index: blocks,
+        Ok(Self {
             tip: tip.expect("already checked to have genesis"),
-        };
-
-        debug_assert!(chain._check_index_is_consistent_with_tip());
-        Ok(chain)
+        })
     }
 
     /// Get the highest checkpoint.
@@ -494,9 +520,7 @@ impl LocalChain {
                 None => LocalChain::from_blocks(extension)?.tip(),
             };
             self.tip = new_tip;
-            self.reindex(start_height);
 
-            debug_assert!(self._check_index_is_consistent_with_tip());
             debug_assert!(self._check_changeset_is_applied(changeset));
         }
 
@@ -509,16 +533,16 @@ impl LocalChain {
     ///
     /// Replacing the block hash of an existing checkpoint will result in an error.
     pub fn insert_block(&mut self, block_id: BlockId) -> Result<ChangeSet, AlterCheckPointError> {
-        if let Some(&original_hash) = self.index.get(&block_id.height) {
+        if let Some(original_cp) = self.tip.get(block_id.height) {
+            let original_hash = original_cp.hash();
             if original_hash != block_id.hash {
                 return Err(AlterCheckPointError {
                     height: block_id.height,
                     original_hash,
                     update_hash: Some(block_id.hash),
                 });
-            } else {
-                return Ok(ChangeSet::default());
             }
+            return Ok(ChangeSet::default());
         }
 
         let mut changeset = ChangeSet::default();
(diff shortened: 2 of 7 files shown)


crates/chain/src/local_chain.rs
```diff
@@ -148,6 +149,34 @@ impl CheckPoint {
...
+
+    /// Find checkpoint at `height`.
+    ///
+    /// Returns `None` if checkpoint at `height` does not exist`. If you are querying for multiple
+    /// heights, it will be more efficient to use [`query_multi`].
+    ///
+    /// [`query_multi`]: CheckPoint::query_multi
+    pub fn query(&self, height: u32) -> Option<Self> {
+        self.iter()
+            .take_while(|cp| cp.height() >= height)
```
LLFourn: No need for this `take_while`.

crates/chain/src/local_chain.rs
```diff
@@ -148,6 +149,34 @@ impl CheckPoint {
...
+            .find(|cp| cp.height() == height)
+    }
+
+    /// Find checkpoints of the given `heights`.
+    ///
+    /// This returns an iterator which iterates over the requested heights and whether a
+    /// corresponding checkpoint is found.
+    pub fn query_multi(&self, heights: BTreeSet<u32>) -> impl Iterator<Item = (u32, Option<Self>)> {
```
LLFourn: Seems like a premature optimization. I understand that this is faster than calling query several times but the way to speed `query` up is with a skip list.

LLFourn (commented):
I'm slightly concerned we're introducing these methods without using them anywhere. Especially `query_multi` which I don't think needs to exist yet. I suggest cutting `query_multi`.

crates/chain/src/local_chain.rs
evanlinjin: I think we need this. If we don't have this and we don't have a direct hit, we will iterate the entire chain.

LLFourn: Ah good point thanks! (comment that)

crates/chain/src/local_chain.rs
evanlinjin: Even with a skip list, this will be faster? I also see this as a convenience method.

LLFourn: > Even with a skip list, this will be faster? This is also a convenience method, like super convenient. I honestly think this is worth it.

Things that are faster are not necessarily better (if done before we have evidence that we need faster). Can we add the convenience method at the point that we need the convenience?

I think in general this PR doesn't really belong at this point because we don't get to see how we actually use the API. It seems like a simple thing to do as part of a larger PR that deletes the BTreeMap from LocalChain and uses checkpoint queries instead. Can we just make that this PR?

crates/chain/src/local_chain.rs
evanlinjin: Comment added.

crates/chain/src/local_chain.rs
evanlinjin: I realized that you are right. I think it is a great policy to only introduce convenience methods if they are used. `CheckPoint::query_multi` turns out to be not very useful. However, I introduced `CheckPoint::query_from` which will return `Some` if there is a checkpoint at `height` or greater. This is a sufficient replacement for `BTreeMap::<u32, BlockHash>::range` which is used in `bdk::Wallet::insert_tx`.

crates/chain/src/local_chain.rs
```diff
@@ -205,18 +224,39 @@ pub struct Update {
 #[derive(Debug, Clone)]
 pub struct LocalChain {
     tip: CheckPoint,
-    index: BTreeMap<u32, BlockHash>,
+    xor: BlockHashXor,
```
LLFourn: I think this is a bit premature. I don't know that we need really fast equality for anything.  We didn't have fast equality before. If we were to do it, I'd just do it in `CheckPoint`.

evanlinjin: > If we were to do it, I'd just do it in `CheckPoint`.

Do you mean implement fast equality in `CheckPoint`? So every checkpoint will have an `xor` field?

evanlinjin: Yes, I agree that adding the xor right now is premature.

crates/chain/src/local_chain.rs
```diff
@@ -34,6 +34,20 @@ struct CPInner {
     prev: Option<Arc<CPInner>>,
 }
 
+impl PartialEq for CheckPoint {
+    fn eq(&self, other: &Self) -> bool {
```
LLFourn, line 39: :wrench: https://doc.rust-lang.org/core/iter/trait.Iterator.html#method.eq

I think it can be implemented with this.

LLFourn (approved):
ACK aa84ecf9d0a6f37e56c6fbebe3a2370f22f8acfd

crates/chain/tests/test_local_chain.rs
```diff
@@ -528,6 +528,45 @@ fn checkpoint_from_block_ids() {
...
+    ];
+
+    for t in test_cases.into_iter() {
+        let tip = t.chain.tip();
+        for h in t.query_range.0..=t.query_range.1 {
+            let query_result = tip.query(h);
+            let exp_hash = t.chain.query(h).map(|cp| cp.hash());
+            match query_result {
+                Some(cp) => {
+                    assert_eq!(Some(cp.hash()), exp_hash);
```
ValuedMammal, line 568: It kind of looks like we're getting the checkpoint hash twice here from the same query result.

crates/chain/src/local_chain.rs
LLFourn: > > If we were to do it, I'd just do it in `CheckPoint`.
> 
> Do you mean implement fast equality in `CheckPoint`? So every checkpoint will have an `xor` field?

Yes. As discussed in #1380 we probably do want to do this for `CheckPoint` in the future.

crates/chain/src/local_chain.rs
ValuedMammal: Would that not cause a recursion though?

LLFourn: Oh interesting point. We still want to map to the `BlockId` so the eq used in the iterator is just `BlockId`'s. That should be fine!

crates/chain/tests/test_local_chain.rs
```diff
@@ -528,6 +528,45 @@ fn checkpoint_from_block_ids() {
...
+                    assert_eq!(cp.height(), h);
+                }
+                None => assert!(query_result.is_none()),
```
notmandatory: 🔧 Since this match arm already told us `query_result` is `None` did you mean to do this instead?
```suggestion
                None => assert!(exp_hash.is_none()),
```

crates/chain/tests/test_local_chain.rs
notmandatory: I think this is OK, I read this line as saying: if the `query_result` is some checkpoint, then the hash of that checkpoint must equal the expected hash. 

⛏️ On the other hand I don't see how this test could ever fail since both `tip.query(h)` and `t.chain.query(h)` call `Checkpoint::query()` with `h`.

notmandatory (approved):
ACK aa84ecf9d0a6f37e56c6fbebe3a2370f22f8acfd

Appreciate the focused scope for this series of chain PRs, much easier to review. I have one small test code suggestion but otherwise looks good to me.

crates/chain/tests/test_local_chain.rs
evanlinjin: I realized how I wrote this was wrong. We shouldn't be using `query` for both what is to be checked and what we are checking against.

crates/chain/tests/test_local_chain.rs
evanlinjin: Fixed.

crates/chain/tests/test_local_chain.rs
evanlinjin: > I realized how I wrote this was wrong. We shouldn't be using `query` for both what is to be checked and what we are checking against.

Fixed.

crates/chain/src/local_chain.rs
evanlinjin: Good point @ValuedMammal. I'll map the iterator to block id first.

evanlinjin: Updated to use `Iterator::eq`.

notmandatory (approved):
re-ACK 2b006da9042156b6ad9ce5a4237fc24973d0d72a

evanlinjin:
I would hold off merging this until this discussion reaches a consensus: https://github.com/bitcoindevkit/bdk/pull/1380#discussion_r1540554140

evanlinjin:
I have agreed with @LLFourn's suggestion of implementing `CheckPoint::range` and have updated the PR to implement it. Refer to the conversation here: https://github.com/bitcoindevkit/bdk/pull/1380#discussion_r1547182439

This PR is ready for review/merge.

evanlinjin:
Ahhh I probably still need to add a few tests.

crates/chain/src/local_chain.rs
```diff
@@ -148,6 +158,39 @@ impl CheckPoint {
...
     }
+
+    /// Query for checkpoint at `height`.
+    ///
+    /// Returns `None` if checkpoint at `height` does not exist`.
+    pub fn query(&self, height: u32) -> Option<Self> {
+        self.iter()
+            // optimization to avoid iterating the entire chain if we do not get a direct hit
+            .take_while(|cp| cp.height() >= height)
+            .find(|cp| cp.height() == height)
```
jirijakes: Perhaps opportunity to reuse `range`, since `query` is its special case? It seems that even performance-wise it would not be worse.

```suggestion
	self.range(height..=height).next()
```

crates/chain/src/local_chain.rs
```diff
@@ -148,6 +158,39 @@ impl CheckPoint {
     pub fn iter(&self) -> CheckPointIter {
         self.clone().into_iter()
     }
+
+    /// Query for checkpoint at `height`.
+    ///
+    /// Returns `None` if checkpoint at `height` does not exist`.
+    pub fn query(&self, height: u32) -> Option<Self> {
```
jirijakes: The name feels hard to discover. If I am interested in finding a checkpoint at a _height_, I would first look for a method containing _height_, something like `find_at_height()`, `find_by_height()`, `get_at_height()`, `at_height()` or so.

Similarly `range` and the equivalent methods on `LocalChain`.

evanlinjin: What about `query_height`? cc @LLFourn

LLFourn: I'd go with just `.get_hash` or better yet just `get`. `query` is not a good name indeed.

crates/chain/src/local_chain.rs
```diff
@@ -148,6 +158,39 @@ impl CheckPoint {
...
+        self.iter()
+            // optimization to avoid iterating the entire chain if we do not get a direct hit
+            .take_while(|cp| cp.height() >= height)
+            .find(|cp| cp.height() == height)
+    }
+
+    /// Iterate checkpoints over a height range.
+    ///
+    /// Note that we always iterate checkpoints in reverse height order (iteration starts at tip
```
LLFourn, line 170: :pick: there's no need to start this sentence with `Note that`. It doesn't start at the tip it starts at the highest height within the range.

The implementation looks correct fwiw. I still feel like we are adding methods like this before we actually need them. My bias is to keep it for now since you've done it but we really have to stop adding these kinds of methods.

evanlinjin: This is needed though, like we're actually using it

crates/chain/src/local_chain.rs
evanlinjin: Let's go with `.get`.

crates/chain/src/local_chain.rs
```diff
@@ -202,21 +245,19 @@ pub struct Update {
...
-    index: BTreeMap<u32, BlockHash>,
-}
-
-impl PartialEq for LocalChain {
-    fn eq(&self, other: &Self) -> bool {
-        self.index == other.index
-    }
 }
 
+// TODO: Figure out whether we can get rid of this
```
ValuedMammal: You could make this be the implementation for the `LocalChain::blocks` function and have it return `BTreeMap<u32, BlockHash>`, possibly even taking an extra `range: impl RangeBounds<u32>` parameter.

crates/chain/src/local_chain.rs
```diff
@@ -1,6 +1,8 @@
 //! The [`LocalChain`] is a local implementation of [`ChainOracle`].
 
 use core::convert::Infallible;
+use core::ops::RangeBounds;
+use core::u32;
```
ValuedMammal: Is this `core::u32` needed?

ValuedMammal (commented):
Overall looks good.

`get` sounds like a better name than `query`. My other suggestion would have been to call it `find`, since that is how it's implemented.

Interesting that `LocalChain` now delegates virtually everything to the checkpoint. Curious to see if the structure of `LocalChain` becomes more relevant when skip list is involved.

crates/chain/src/local_chain.rs
LLFourn: I see one place it's being used where [as we've discussed](https://github.com/bitcoindevkit/bdk/pull/1380#discussion_r1540554140) it's not necessary.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1128,18 +1128,14 @@ impl<D> Wallet<D> {
                 // anchor tx to checkpoint with lowest height that is >= position's height
                 let anchor = self
                     .chain
-                    .blocks()
                     .range(height..)
-                    .next()
+                    .last()
```
evanlinjin, line 1131: @LLFourn `CheckPoint::range` is also used here.

LLFourn: Thanks. I missed that one and had forgot about the API of this function. On a side note this is interesting because we are anchoring a tx not to the block that it's in. So even if we made a rule that esplora/electrum etc must tell us about the block that the tx is in we'd still need the concept of anchoring if we want to keep this API.

LLFourn:
ACK cdbc131ef68182a4317beb145232ec8e87d6299c

LLFourn (approved):

crates/chain/src/local_chain.rs
evanlinjin: I blame my autocomplete engine.

evanlinjin: Removed

crates/chain/src/local_chain.rs
evanlinjin: Changed.

crates/chain/src/local_chain.rs
evanlinjin: This is an interesting idea! However, I like to use @LLFourn's policy when it comes to introducing helper methods/structures. We should only introduce such things if we introduce how/where it is used within the PR. This ensures the usefulness of the things we are introducing.

I've decided to remove this `impl` completely as it was not being used.

crates/chain/src/local_chain.rs
evanlinjin: Removed.

crates/chain/src/local_chain.rs
evanlinjin: Conversation continues here: https://github.com/bitcoindevkit/bdk/pull/1369#discussion_r1550753230

evanlinjin:
I've addressed all review comments and removed the `From<LocalChain>` impl for `BTreeMap<u32, BlockHash>` as it wasn't being used.

crates/chain/src/local_chain.rs
```diff
@@ -276,39 +299,30 @@ impl LocalChain {
...
+        Ok(Self { tip })
     }
 
     /// Constructs a [`LocalChain`] from a [`BTreeMap`] of height to [`BlockHash`].
     ///
     /// The [`BTreeMap`] enforces the height order. However, the caller must ensure the blocks are
     /// all of the same chain.
     pub fn from_blocks(blocks: BTreeMap<u32, BlockHash>) -> Result<Self, MissingGenesisError> {
-        if !blocks.contains_key(&0) {
+        if blocks.get(&0).is_none() {
```
jirijakes, line 321: Just curious: why this change? Original looked a bit better readable.

crates/chain/src/local_chain.rs
```diff
@@ -542,33 +550,41 @@ impl LocalChain {
...
+            // The checkpoint below the earliest checkpoint to remove will be the new tip.
+            Some(Some(new_tip)) => new_tip,
+            // If there is no checkpoint below the earliest checkpoint to remove, it means the
+            // "earliest checkpoint to remove" is the genesis block. We disallow removing the
+            // genesis block.
+            Some(None) => return Err(MissingGenesisError),
+            // If there is nothing to remove, we return an empty changeset.
+            None => return Ok(ChangeSet::default()),
+        };
+        Ok(changeset)
```
jirijakes, line 575: Not too important, but it took me a while to understand what changeset is returned when and what is when assigned to `self.tip` and when it is left untouched.

Perhaps?

```suggestion
        match remove_from.map(|cp| cp.prev()) {
            // The checkpoint below the earliest checkpoint to remove will be the new tip.
            Some(Some(new_tip)) => {
                self.tip = new_tip;
                Ok(changeset)
            }
            // If there is no checkpoint below the earliest checkpoint to remove, it means the
            // "earliest checkpoint to remove" is the genesis block. We disallow removing the
            // genesis block.
            Some(None) => Err(MissingGenesisError),
            // If there is nothing to remove, we return an empty changeset.
            None => Ok(ChangeSet::default()),
        }
```

jirijakes (approved):
The PR helped me better understand some internals. Still learning.

Only have one question and one minor suggestion (not too important).

ACK 6d78f0e0fbaf7b25499ed671fdd2c1266a0d47c7

crates/chain/src/local_chain.rs
evanlinjin: `.contains_key` no longer exists because `blocks` is a different type now.

LLFourn (approved):
ACK 53942cced492138174638a8087b08f643a8d41ad

crates/chain/src/local_chain.rs
evanlinjin: Your suggestion is definitely better, but I'll ignore it now because I want this merged and rebasing is annoying 😅

PR #1373, opened 2024-03-14
Wrap transactions as `Arc<Transaction>` in `TxGraph`

evanlinjin:
### Description

This PR makes `TxGraph` store transactions as `Arc<Transaction>`.

`Arc<Transaction>` can be shared between the chain-source and receiving structures. This allows the chain-source to keep the already-fetched transactions (save bandwith) and have a shared pointer to the transaction (save memory).

Our current logic to avoid re-fetching transactions is to refer back to the receiving structures. However, this means an additional round of locking our receiving structures.

### Notes to the reviewers

This will make more sense once I update the esplora/electrum chain sources to make use of both bitcoindevkit/bdk#1369 and this PR. The result would be chain sources which would only require locking the receiving structures twice (once for initiating the update, and once for applying the update).

### Changelog notice

* Changed `TxGraph` to store transactions as `Arc<Transaction>`. This allows chain-sources to cheaply keep a copy of already-fetched transactions.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -942,7 +942,7 @@ impl<D> Wallet<D> {
     /// # let mut wallet: Wallet<()> = todo!();
     /// # let txid:Txid = todo!();
     /// let tx = wallet.get_tx(txid).expect("transaction").tx_node.tx;
-    /// let fee = wallet.calculate_fee(tx).expect("fee");
+    /// let fee = wallet.calculate_fee(&tx).expect("fee");
     /// ```
     ///
     /// ```rust, no_run
@@ -973,7 +973,7 @@ impl<D> Wallet<D> {
     /// # let mut wallet: Wallet<()> = todo!();
     /// # let txid:Txid = todo!();
     /// let tx = wallet.get_tx(txid).expect("transaction").tx_node.tx;
-    /// let fee_rate = wallet.calculate_fee_rate(tx).expect("fee rate");
+    /// let fee_rate = wallet.calculate_fee_rate(&tx).expect("fee rate");
     /// ```
     ///
     /// ```rust, no_run
@@ -981,8 +981,8 @@ impl<D> Wallet<D> {
     /// # use bdk::Wallet;
     /// # let mut wallet: Wallet<()> = todo!();
     /// # let mut psbt: PartiallySignedTransaction = todo!();
-    /// let tx = &psbt.clone().extract_tx();
-    /// let fee_rate = wallet.calculate_fee_rate(tx).expect("fee rate");
+    /// let tx = psbt.clone().extract_tx();
+    /// let fee_rate = wallet.calculate_fee_rate(&tx).expect("fee rate");
     /// ```
     /// [`insert_txout`]: Self::insert_txout
     pub fn calculate_fee_rate(&self, tx: &Transaction) -> Result<FeeRate, CalculateFeeError> {
@@ -1003,8 +1003,8 @@ impl<D> Wallet<D> {
     /// # use bdk::Wallet;
     /// # let mut wallet: Wallet<()> = todo!();
     /// # let txid:Txid = todo!();
-    /// let tx = wallet.get_tx(txid).expect("transaction").tx_node.tx;
-    /// let (sent, received) = wallet.sent_and_received(tx);
+    /// let tx = wallet.get_tx(txid).expect("tx exists").tx_node.tx;
+    /// let (sent, received) = wallet.sent_and_received(&tx);
     /// ```
     ///
     /// ```rust, no_run
@@ -1065,7 +1065,7 @@ impl<D> Wallet<D> {
     pub fn get_tx(
         &self,
         txid: Txid,
-    ) -> Option<CanonicalTx<'_, Transaction, ConfirmationTimeHeightAnchor>> {
+    ) -> Option<CanonicalTx<'_, Arc<Transaction>, ConfirmationTimeHeightAnchor>> {
         let graph = self.indexed_graph.graph();
 
         Some(CanonicalTx {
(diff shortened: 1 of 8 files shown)


evanlinjin:
TODO: Make sure all the examples still work.

LLFourn:
Main concern: I don't think we need `T` here. That seem premature. We can just concretely have an `Arc<Transaction>`.

crates/bdk/src/wallet/mod.rs
```diff
@@ -982,7 +982,7 @@ impl<D> Wallet<D> {
     /// # let mut wallet: Wallet<()> = todo!();
     /// # let mut psbt: PartiallySignedTransaction = todo!();
     /// let tx = &psbt.clone().extract_tx();
-    /// let fee_rate = wallet.calculate_fee_rate(tx).expect("fee rate");
+    /// let fee_rate = wallet.calculate_fee_rate(&tx).expect("fee rate");
```
LLFourn, line 985: I don't think this is needed.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1006,7 +1006,7 @@ impl<D> Wallet<D> {
     /// # let mut wallet: Wallet<()> = todo!();
     /// # let txid:Txid = todo!();
     /// let tx = wallet.get_tx(txid).expect("transaction").tx_node.tx;
```
LLFourn: why does this say `expect("transaction")`. Shouldn't it say `expect("exists")`.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1006,7 +1006,7 @@ impl<D> Wallet<D> {
...
-    /// let (sent, received) = wallet.sent_and_received(tx);
+    /// let (sent, received) = wallet.sent_and_received(&tx);
```
LLFourn, line 1007: Interested to understand why this `&` is needed now. Before tx was of type `Transaction` right, so shouldn't it needed to have been borrowed before?

LLFourn:
Can this PR not be done on top of bitcoindevkit/bdk#1369. I think it can go before and is pretty unrelated.

crates/bdk/src/wallet/mod.rs
evanlinjin: Before, `TxNode`'s `tx` field is of a reference to `T`. I changed it in this PR to just be `tx: T`. Note that `T` is now `Arc<Transaction>`.

Because `sent_and_received` takes in a `&Transaction`, we expect to call `.as_ref` on `Arc<Transaction>`. However, we can make use of the [auto-deref properties of `AsRef` implementations](https://doc.rust-lang.org/std/convert/trait.AsRef.html#generic-implementations) (such as `Arc`), so just doing `&tx` is enough.

crates/bdk/src/wallet/mod.rs
evanlinjin: I removed the extra reference on the previous line.

evanlinjin:
@LLFourn removed the dependency on bitcoindevkit/bdk#1369

crates/bdk/src/wallet/mod.rs
ValuedMammal: Maybe we can map to shared ref `&` internally and avoid returning `Arc<Transaction>` in `get_tx{_node}`

On second thought, the PR description implies that we specifically DO want to return an arc pointer.

crates/chain/src/tx_graph.rs
```diff
@@ -30,10 +31,22 @@
...
 //!
-//! Methods that apply changes to [`TxGraph`] will return [`ChangeSet`].
-//! [`ChangeSet`] can be applied back to a [`TxGraph`] or be used to inform persistent storage
+//! Methods that change the state of [`TxGraph`] will return [`ChangeSet`]s.
+//! [`ChangeSet`]s can be applied back to a [`TxGraph`] or be used to inform persistent storage
 //! of the changes to [`TxGraph`].
 //!
+//! # Generics
+//!
+//! Anchors are represented as generics within `TxGraph<A, T>`. To make use of all functionality of
```
ValuedMammal: At the moment tx graph is still `TxGraph<A>`

crates/chain/tests/test_tx_graph.rs
```diff
@@ -664,17 +668,17 @@ fn test_walk_ancestors() {
 
     let ancestors = [
         graph
-            .walk_ancestors(&tx_c0, |depth, tx| Some((depth, tx)))
+            .walk_ancestors(Arc::new(tx_c0.clone()), |depth, tx| Some((depth, tx)))
```
ValuedMammal: I'm guessing this also works.

```suggestion
            .walk_ancestors(tx_c0, |depth, tx| Some((depth, tx)))
```

crates/bdk/src/wallet/mod.rs
evanlinjin: Yes, I think we should return a concrete `Arc` for this API to be useful (cheaply share transactions between chain-source, receiving-structs and changesets).

LLFourn (approved):
ACK 8ab58af093ff295b86c3f8189aa96dea313092c4

ValuedMammal (approved):

crates/chain/src/tx_graph.rs
```diff
@@ -430,7 +444,7 @@ impl<A> TxGraph<A> {
     /// Note that this only returns directly conflicting txids and won't include:
     /// - descendants of conflicting transactions (which are technically also conflicting)
     /// - transactions conflicting with the given transaction's ancestors
-    pub fn direct_conflitcs<'g>(
+    pub fn direct_conflicts<'g>(
```
notmandatory, line 447: 🏕️ nice cleanup!

notmandatory (approved):
ACK 8ab58af093ff295b86c3f8189aa96dea313092c4

May need a rebase.

PR #1377, opened 2024-03-16
ci: fix dependabot for audit and cont_integration workflows

notmandatory:
### Description

We are getting errors on dependabot generated PRs. See:

https://github.com/bitcoindevkit/bdk/actions/runs/8310897352
https://github.com/bitcoindevkit/bdk/actions/runs/8310897348

To fix this one of the recommended solutions is to exclude `push` triggered workflow runs for the dependabot created PR branches, `push` trigger is sufficient. See: [Error: 403 "Resource not accessible by integration"](https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/resource-not-accessible).

### Notes to the reviewers

An alternative fix is to remove the `push` trigger but this seemed like a smaller change and the audit workflow only uses the `push` trigger and I'm not sure if it will break if we switch it to use `pull_request` instead. 

It's a little annoying having dependabot create PRs but since it only triggers on security related issues it's worth having so we don't forget to do those updates.

### Changelog notice

None.

#### All Submissions:

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -5,6 +5,8 @@ on:
     paths:
       - '**/Cargo.toml'
       - '**/Cargo.lock'
+    branches-ignore:
+      - 'dependabot/**'
   schedule:
     - cron: '0 0 * * 0' # Once per week
 
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -1,4 +1,8 @@
-on: [push, pull_request]
+on:
+  push:
+    branches-ignore:
+      - 'dependabot/**'
+  pull_request:
 
 name: CI
 


notmandatory:
@storopoli have you run into this issue before with dependabot and any thoughts on if this is the best way to fix it?

storopoli:
Digging a little on this one by looking at:

- [Error: 403 "Resource not accessible by integration"](https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/resource-not-accessible)
- [Workflow syntax for GitHub Actions - Permissions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions)

We can definitely add the ignore as you are doing,
or we can try to add the necessary permissions as well.
Something like this:

```yml
jobs:
  security_audit:
    runs-on: ubuntu-20.04
    permissions:
      security_events: write
```

I am fine with both approaches.
I think that security audits would be nice to have in dependabot PRs,
since it is primarily updating dependencies.
I am fond of having a `Cargo.lock` committed and updating it through dependabot,
see [Dependabot `cargo`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#cargo).
It will be much easier than the current pinning approach.

The Nix CI will add a `Cargo.lock` (#1320) and the next logical step,
which I was already planning,
was to add Cargo updates to dependabot.

oleonardolima:
Is this still an issue ? Also, I haven't seen any suggestion/updates by dependabot in a while 🤔.

FWIW, fedimint uses a separate `dependabot.yml` CI job AFAICT it works fine, this one: https://github.com/fedimint/fedimint/blob/master/.github/dependabot.yml

notmandatory:
@oleonardolima thanks for the tip, I'll take a look at how fedimint is doing it and possibly rework this PR.

notmandatory:
This doesn't look to be an issue anymore, will open an new issue if we get any more dependabot problems.  

I opened a new PR to use a new rust audit task since the old `actions-rs/audit-check` is no longer maintained. See: #1646

PR #1378, opened 2024-03-17
Update bdk README

### LINKED ISSUE #1044: README example is wrong
```rust
use bdk::{bitcoin::Network, wallet::{AddressIndex, Wallet}};

fn main() {
    // a type that implements `Persist`
    let db = ();

    let descriptor = "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/0'/0'/0/*)";
    let mut wallet = Wallet::new(descriptor, None, db, Network::Testnet).expect("should create");

    // get a new address (this increments revealed derivation index)
    println!("revealed address: {}", wallet.get_address(AddressIndex::New));
    println!("staged changes: {:?}", wallet.staged());
    // persist changes
    wallet.commit().expect("must save");
}
```

This indicates two things that are not helpful:
1. You should do `Wallet::new` when you don't want persistence. We have a method for creating a wallet without persistence. If we want to show them how to use actual persistence then let's use that method.
2. You do not need to call `commit` after doing `get_address` to persist the changes -- it does it automatically (in fact `staged` should be empty which indicates to me no one has run this example :sweat_smile:) . There's also no point in committing anyway since the db is explicitly `()` so it's confusing.

ValuedMammal:
fixes #1044 

### Notes

The code snippet is a compile fail because variables `txout` and `outpoint` are not initialized. Any other suggestions are welcome.

There is still some old commented code in the file that's probably not needed unless we want to provide another example snippet, say for getting new address info.

* [x] Fix broken links
* [x] Use a new unused tprv
* [ ] Try to make persistence example make sense, or maybe just do away with it

#### All Submissions:

### DIFF
--- a/crates/bdk/README.md
+++ b/crates/bdk/README.md
@@ -26,7 +26,7 @@
 
 ## `bdk`
 
-The `bdk` crate provides the [`Wallet`](`crate::Wallet`) type which is a simple, high-level
+The `bdk` crate provides the [`Wallet`] type which is a simple, high-level
 interface built from the low-level components of [`bdk_chain`]. `Wallet` is a good starting point
 for many simple applications as well as a good demonstration of how to use the other mechanisms to
 construct a wallet. It has two keychains (external and internal) which are defined by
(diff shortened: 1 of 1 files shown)


crates/bdk/README.md
```diff
@@ -218,9 +218,11 @@ submitted for inclusion in the work by you, as defined in the Apache-2.0
 license, shall be dual licensed as above, without any additional terms or
 conditions.
 
+[`Wallet`]: https://docs.rs/bdk/1.0.0-alpha.7/bdk/wallet/struct.Wallet.html
```
ValuedMammal, line 221: Perhaps this could also point to 'latest', only problem is that the latest on docs.rs is still on 0.29

notmandatory: Once we get a final 1.0.0 we can remove the "pre-release" flag and make it the latest on crates.io and docs.rs.

evanlinjin (approved):
ACK 06d7dc5c3a890f0a141de4cf6f421fe766d99ec3

PR #1380, opened 2024-03-21
Simplified `EsploraExt` API

### LINKED ISSUE #1354: The `TxGraph::missing_heights` implementation is problematic.

evanlinjin:
Fixes #1354

### Description

Built on top of both #1369 and #1373, we simplify the `EsploraExt` API by removing the `update_local_chain` method and having `full_scan` and `sync` update the local chain in the same call. The `full_scan` and `sync` methods now takes in an additional input (`local_tip`) which provides us with the view of the `LocalChain` before the update. These methods now return structs `FullScanUpdate` and `SyncUpdate`.

The examples are updated to use this new API. `TxGraph::missing_heights` and `tx_graph::ChangeSet::missing_heights_from` are no longer needed, therefore they are removed.

Additionally, we used this opportunity to simplify the logic which updates `LocalChain`. We got rid of the `local_chain::Update` struct (which contained the update `CheckPoint` tip and a `bool` which signaled whether we want to introduce blocks below point of agreement). It turns out we can use something like `CheckPoint::insert` so the chain source can craft an update based on the old tip. This way, we can make better use of `merge_chains`' optimization that compares the `Arc` pointers of the local and update chain (before we were crafting the update chain NOT based on top of the previous local chain). With this, we no longer need the `Update::introduce_older_block` field since the logic will naturally break when we reach a matching `Arc` pointer.

### Notes to the reviewers

* Obtaining the `LocalChain`'s update now happens within `EsploraExt::full_scan` and `EsploraExt::sync`. Creating the `LocalChain` update is now split into two methods (`fetch_latest_blocks` and `chain_update`) that are called before and after fetching transactions and anchors.
* We need to duplicate code for `bdk_esplora`. One for blocking and one for async.

### Changelog notice

* Changed `EsploraExt` API so that sync only requires one round of fetching data. The `local_chain_update` method is removed and the `local_tip` parameter is added to the `full_scan` and `sync` methods.
* Removed `TxGraph::missing_heights` and `tx_graph::ChangeSet::missing_heights_from` methods.
* Introduced `CheckPoint::insert` which allows convenient checkpoint-insertion. This is intended for use by chain-sources when crafting an update.
* Refactored `merge_chains` to also return the resultant `CheckPoint` tip.
* Optimized the update `LocalChain` logic - use the update `CheckPoint` as the new `CheckPoint` tip when possible.

#### All Submissions:

#### New Features:


evanlinjin:
@notmandatory ready for you to base #1194 on!

crates/esplora/src/blocking_ext.rs
```diff
@@ -68,250 +56,368 @@ pub trait EsploraExt {
...
+        .filter({
+            let mut prev_height = Option::<u32>::None;
+            move |h| match prev_height.replace(*h) {
+                None => true,
+                Some(prev_h) => prev_h != *h,
             }
+        })
+        // filter out heights that surpass the update tip
+        .filter(|h| *h <= update_tip_height)
+        .rev();
```
LLFourn: :pick: I feel it's better to put the `.rev` first unless there's some reason it has to be last. The comment above says we're iterating backwards but it seems to by lying.

:pick: Can the `.filter` be `.take_while` here?

crates/esplora/src/blocking_ext.rs
```diff
@@ -68,250 +56,368 @@ pub trait EsploraExt {
...
+
+    // We keep track of a checkpoint node of `local_tip` to make traversing the linked-list of
+    // checkpoints more efficient.
+    let mut curr_cp = local_tip.clone();
 
-            if let Some(last_active_index) = last_active_index {
-                last_active_indexes.insert(keychain, last_active_index);
+    for h in anchor_heights {
+        if let Some(cp) = curr_cp.query_from(h) {
```
LLFourn: Don't I just want to know if we have the hash at `h`. Why use `query_from` rather than `query`.

Also it looks like `query_from` should just be generalized into a `range(..)` type method to have more idiomatic and easy to understand name (if it needs to exist right now).

crates/esplora/src/blocking_ext.rs
```diff
@@ -68,250 +56,368 @@ pub trait EsploraExt {
...
+            curr_cp = cp.clone();
+            if cp.height() == h {
+                // blocks that already exist in checkpoint linked-list is also stored in
+                // `update_blocks` because we want to keep higher checkpoints of `local_chain`
+                update_blocks.insert(h, cp.hash());
```
LLFourn: :confused:  I don't understand why you want to do this. Probably can be answered by the question below.

crates/esplora/src/blocking_ext.rs
```diff
@@ -68,250 +56,368 @@ pub trait EsploraExt {
...
-            .map(|(g, _)| g)?;
-
-        let mut txids = txids.into_iter();
+    Ok(local_chain::Update {
+        tip: CheckPoint::from_block_ids(
+            update_blocks
+                .into_iter()
+                .map(|(height, hash)| BlockId { height, hash }),
+        )
+        .expect("must be in order"),
```
LLFourn: Isn't it much more straightforward to just extend the existing chain by pushing onto the `local_tip` one from the point of agreement. Why are we constructing an entirely new one.

crates/esplora/src/blocking_ext.rs
```diff
@@ -68,250 +56,368 @@ pub trait EsploraExt {
...
+        introduce_older_blocks: true,
+    })
+}
+
+/// This performs a full scan to get an update for the [`TxGraph`] and
+/// [`KeychainTxOutIndex`](bdk_chain::keychain::KeychainTxOutIndex).
+#[doc(hidden)]
+pub fn full_scan_for_index_and_graph_blocking<K: Ord + Clone>(
```
LLFourn: why is this a `pub fn`

LLFourn (commented):
Left some comments so I can understand a few things better before giving a full review.

crates/esplora/src/blocking_ext.rs
evanlinjin: Because we mark the update to introduce older checkpoints, and in which case we will iterate the whole chain.

crates/esplora/src/blocking_ext.rs
evanlinjin: So we can test it in the separate test folder. The other option is to move the tests in `src`.

crates/esplora/src/blocking_ext.rs
LLFourn: Hmm don't we know if the thing actually does introduce older checkpoints? We don't have to fix `introduce_older_blocks` to `true`. Also building from the existing checkpoint will allow us to stop early if the `Arc` pointers match.

IIRC `introduce_older_blocks` was needed so we could set it to false for block-by-blocking syncing where if it were assumed to be true and some edge case happened then the `Arc` pointers wouldn't match and the update would be computationally intensive.

We could probably overcome this with a block `xor` state accumulator thingo that you had in the other PR. So I'd work on doing this in the most straightforward way and then probably remove the whole `introduce_older_blocks` thing in favor of block xor to detect if something has changed instead of relying on `Arc` pointers.

evanlinjin: I see what you are saying. I think I did try to construct the update from the old tip, but it turned out more complicated than I expected? We have to traverse the old chain to find a height just lower than our update's earliest height. I think I can try look at this problem again to see if I can simplify it.

crates/esplora/src/blocking_ext.rs
evanlinjin: The `.filter` cannot be `.take_while` here since we are only filtering out top-most (highest) blocks (if there is anything to filter out at all).

I think it makes sense to put the `.rev` first.

crates/esplora/src/blocking_ext.rs
evanlinjin: We are using `.query_from` as an optimization; `.query_from(height)` returns either the checkpoint of the given `height`, or the one just greater than `height`. Either way, we store the returned `CheckPoint` in `curr_cp` which we can call `.query_from` on again.

Just for some context, I originally wanted to create an update by extending from `curr_cp` (but I deemed it too complicated).

crates/esplora/src/blocking_ext.rs
evanlinjin: If we ensure that our update has a checkpoint at previously-had heights, we can avoid an disjoint update? I think I need to prove that this may happen though via a test (as tests still pass when I remove this line).

evanlinjin: Okay I thought about it and it will not create a disjoint update so we don't need this line.

crates/esplora/src/blocking_ext.rs
LLFourn: Yeah I'd move that test in here if it doesn't test the public API.

crates/esplora/src/blocking_ext.rs
LLFourn: Perhaps it would feel less complicated if you made an `insert(height, hash)` method on `CheckPoint` which consumed `self` and returned a new chain with that height and hash inserted at the right point.

I think we should remove the whole `introduce_older_blocks` thing in this PR and just assume it's true and make it faster later. I think we know how to do that now.

notmandatory:
> @notmandatory ready for you to base #1194 on!

I finished rebasing #1194 on this PR and everything is working fine.

evanlinjin:
@notmandatory I forgot to mention I also need to rebase this PR on master 😬

crates/esplora/src/blocking_ext.rs
evanlinjin: @LLFourn I think the `range`-type methods usually return a list of items (while we want this method to return one item). I think a better name would be `query_until` (with the same function signature)...

crates/esplora/src/blocking_ext.rs
evanlinjin: I think removing the `introduce_older_blocks` thing in this PR is a bad idea. It will make block-by-block chain sources substantially slower (and they are already being used).

I forgot that the `merge_chains` method already has an optimization to stop when we reach a checkpoint with a matching pointer. This should be good enough to break early for spk-based chain sources.

However, for block-by-block chain sources, we aren't given the opportunity to re-introduce local checkpoints to the `Emitter`.

I think we can work to remove `introduce_older_blocks` in a separate PR.

evanlinjin: Is it important to construct the update from the older checkpoint? Is the intention to simplify the logic? I don't think it will make things simpler.

LLFourn: I made a PR to this branch to show how much simpler things could be: https://github.com/evanlinjin/bdk/pull/9

crates/esplora/src/blocking_ext.rs
LLFourn: If you want what `query_from(whatever)` does you can do `range(whatever..).next()` :)

crates/esplora/src/blocking_ext.rs
LLFourn: On removing introduce_older_blocks I'm inclined to do it now. Make it right, then make it fast. I also think it will be fast anyway except in the unlikely case of a re-org (Arc's pointers will match so won't slow down in practice unless I'm forgetting how block by block syncing works). People using `alpha` things should be ready for unexpected slowdown. They don't have to update to the latest `alpha` until it's fixed.

I'll leave it up to you.

crates/esplora/src/blocking_ext.rs
evanlinjin: Updated #1369 to have `CheckPoint::range`. However, I still need to do tests.

crates/esplora/src/blocking_ext.rs
evanlinjin: daf588f016ec3118c875db8ed6b55fa03683f0f6

Let me know what u think

crates/chain/src/local_chain.rs
```diff
@@ -354,29 +355,26 @@ impl LocalChain {
...
-            update.introduce_older_blocks,
-        )?;
-        // `._check_index_is_consistent_with_tip` and `._check_changeset_is_applied` is called in
-        // `.apply_changeset`
-        self.apply_changeset(&changeset)
-            .map_err(|_| CannotConnectError {
-                try_include_height: 0,
-            })?;
+    pub fn apply_update(&mut self, update: CheckPoint) -> Result<ChangeSet, CannotConnectError> {
+        let (changeset, can_replace) = merge_chains(self.tip.clone(), update.clone())?;
```
ValuedMammal: I think this would look better if `merge_chains` was implemented on `LocalChain` and took in `&mut self` and `other`. Also if we change `merge_chains` to `merge_and_apply_changeset`, then the business of `can_replace` can stay inside the same function.

idea: evanlinjin/bdk#10

crates/chain/src/local_chain.rs
```diff
@@ -354,29 +355,26 @@ impl LocalChain {
...
+        if can_replace {
+            self.tip = update;
+        } else {
+            // `._check_index_is_consistent_with_tip` and `._check_changeset_is_applied` is called in
+            // `.apply_changeset`
```
ValuedMammal:
```suggestion
            // `._check_changeset_is_applied` is called in `.apply_changeset`
```

crates/chain/src/local_chain.rs
evanlinjin: I like the idea of returning the result in `merge_chains`. However, removing the `unreachable!` statement doesn't make it nicer imo since we have to rely on `.unwrap`s (also this is just a styling change so I'll just ignore it).

ValuedMammal: agree, the extra refactor is out of scope and more for my own sanity.

evanlinjin: I've updated to have `merge_chains` return the checkpoint here: 877f7284985c9f1186f902cb4ae24753087eea4d

crates/esplora/src/blocking_ext.rs
```diff
@@ -1,55 +1,38 @@
...
 };
 use esplora_client::TxStatus;
 
 use crate::anchor_from_status;
+use crate::FullScanUpdate;
+use crate::SyncUpdate;
 
 /// [`esplora_client::Error`]
-type Error = Box<esplora_client::Error>;
+pub type Error = Box<esplora_client::Error>;
```
evanlinjin, line 19: Why did I make this public?

evanlinjin: I think it's fine though.

LLFourn (approved):
ACK 96a9aa6e63474dbd93a2ef969eef5b07c79e6491

PR #1385, opened 2024-03-22
Fix last seen unconfirmed

### LINKED ISSUE #1336: last_seen is wrong on both electrum and esplora

ValuedMammal:
A method `update_last_seen_unconfirmed` is added for `TxGraph` that allows inserting a `seen_at` time for all unanchored transactions.

fixes #1336 

### Changelog notice

Changed:
- Methods `into_tx_graph` and `into_confirmation_time_tx_graph`for `RelevantTxids` are changed to no longer accept a `seen_at` parameter.

Added:
- Added method `update_last_seen_unconfirmed` for `TxGraph`.

#### All Submissions:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -541,12 +541,75 @@ impl<A: Clone + Ord> TxGraph<A> {
 
     /// Inserts the given `seen_at` for `txid` into [`TxGraph`].
     ///
-    /// Note that [`TxGraph`] only keeps track of the latest `seen_at`.
+    /// Note that [`TxGraph`] only keeps track of the latest `seen_at`. To batch
+    /// update all unconfirmed transactions with the latest `seen_at`, see
+    /// [`update_last_seen_unconfirmed`].
+    ///
+    /// [`update_last_seen_unconfirmed`]: Self::update_last_seen_unconfirmed
     pub fn insert_seen_at(&mut self, txid: Txid, seen_at: u64) -> ChangeSet<A> {
         let mut update = Self::default();
         let (_, _, update_last_seen) = update.txs.entry(txid).or_default();
         *update_last_seen = seen_at;
         self.apply_update(update)
+    }
+
+    /// Update the last seen time for all unconfirmed transactions.
+    ///
+    /// This method updates the last seen unconfirmed time for this [`TxGraph`] by inserting
+    /// the given `seen_at` for every transaction not yet anchored to a confirmed block,
+    /// and returns the [`ChangeSet`] after applying all updates to `self`.
+    ///
+    /// This is useful for keeping track of the latest time a transaction was seen
+    /// unconfirmed, which is important for evaluating transaction conflicts in the same
+    /// [`TxGraph`]. For details of how [`TxGraph`] resolves conflicts, see the docs for
+    /// [`try_get_chain_position`].
+    ///
+    /// A normal use of this method is to call it with the current system time. Although
+    /// block headers contain a timestamp, using the header time would be less effective
+    /// at tracking mempool transactions, because it can drift from actual clock time, plus
+    /// we may want to update a transaction's last seen time repeatedly between blocks.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # use bdk_chain::example_utils::*;
+    /// # use std::time::UNIX_EPOCH;
+    /// # let tx = tx_from_hex(RAW_TX_1);
+    /// # let mut tx_graph = bdk_chain::TxGraph::<()>::new([tx]);
+    /// let now = std::time::SystemTime::now()
+    ///     .duration_since(UNIX_EPOCH)
+    ///     .expect("valid duration")
+    ///     .as_secs();
+    /// let changeset = tx_graph.update_last_seen_unconfirmed(now);
+    /// assert!(!changeset.last_seen.is_empty());
+    /// ```
+    ///
+    /// Note that [`TxGraph`] only keeps track of the latest `seen_at`, so the given time must
+    /// by strictly greater than what is currently stored for a transaction to have an effect.
+    /// To insert a last seen time for a single txid, see [`insert_seen_at`].
+    ///
+    /// [`insert_seen_at`]: Self::insert_seen_at
+    /// [`try_get_chain_position`]: Self::try_get_chain_position
+    pub fn update_last_seen_unconfirmed(&mut self, seen_at: u64) -> ChangeSet<A> {
+        let mut changeset = ChangeSet::default();
+        let unanchored_txs: Vec<Txid> = self
+            .txs
+            .iter()
+            .filter_map(
+                |(&txid, (_, anchors, _))| {
+                    if anchors.is_empty() {
+                        Some(txid)
+                    } else {
+                        None
+                    }
+                },
+            )
+            .collect();
+
+        for txid in unanchored_txs {
+            changeset.append(self.insert_seen_at(txid, seen_at));
+        }
+        changeset
     }
 
     /// Extends this graph with another so that `self` becomes the union of the two sets of
--- a/crates/electrum/src/electrum_ext.rs
+++ b/crates/electrum/src/electrum_ext.rs
@@ -40,15 +40,11 @@ impl RelevantTxids {
     pub fn into_tx_graph(
         self,
         client: &Client,
-        seen_at: Option<u64>,
         missing: Vec<Txid>,
     ) -> Result<TxGraph<ConfirmationHeightAnchor>, Error> {
         let new_txs = client.batch_transaction_get(&missing)?;
         let mut graph = TxGraph::<ConfirmationHeightAnchor>::new(new_txs);
         for (txid, anchors) in self.0 {
-            if let Some(seen_at) = seen_at {
-                let _ = graph.insert_seen_at(txid, seen_at);
-            }
             for anchor in anchors {
                 let _ = graph.insert_anchor(txid, anchor);
             }
@@ -67,10 +63,9 @@ impl RelevantTxids {
     pub fn into_confirmation_time_tx_graph(
         self,
         client: &Client,
-        seen_at: Option<u64>,
         missing: Vec<Txid>,
     ) -> Result<TxGraph<ConfirmationTimeHeightAnchor>, Error> {
-        let graph = self.into_tx_graph(client, seen_at, missing)?;
+        let graph = self.into_tx_graph(client, missing)?;
 
         let relevant_heights = {
             let mut visited_heights = HashSet::new();
--- a/crates/electrum/tests/test_electrum.rs
+++ b/crates/electrum/tests/test_electrum.rs
@@ -68,7 +68,7 @@ fn scan_detects_confirmed_tx() -> Result<()> {
     } = client.sync(recv_chain.tip(), [spk_to_track], None, None, 5)?;
 
     let missing = relevant_txids.missing_full_txs(recv_graph.graph());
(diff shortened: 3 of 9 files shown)


notmandatory:
This looks fine to me, but my understanding is that a time value always should be provided, even if it's made up as long the the values are increasing. If we make this optional we should at least explain what the downside is of using None, which I believe was unneeded DB updates.

ValuedMammal:
> This looks fine to me, but my understanding is that a time value always should be provided, even if it's made up as long the the values are increasing. If we make this optional we should at least explain what the downside is of using None, which I believe was unneeded DB updates.

Ok, I think I agree it should be required.

rustaceanrob:
I agree on required. It should also make for a better UX as the end user "refreshes" their wallet.

evanlinjin:
Sorry about the conflicts! We have merged #1171

ValuedMammal:
electrum: `into_tx_graph` for RelevantTxids takes an optional `seen_at`, that could be why I was undecided before whether it should always be provided.

@evanlinjin Can we assume in electrum_ext if a relevant txid has no associated anchors, it must be unconfirmed and therefore be an appropriate time to `insert_seen_at`?

evanlinjin:
@ValuedMammal that sounds right to me! Let's add a couple of tests and I'll be happy to ACK and merge.

**My proposed tests for both `ElectrumExt` and `EsploraExt`:** When syncing, ensure that transactions that stay unconfirmed between rounds have their `last_seen` increased and transactions that stay confirmed have unchanging `last_seen`s.

ValuedMammal:
Update on test coverage:

For electrum, repeated syncing doesn't update the last seen timestamp, because if a tx is already present in graph, TxGraph won't report it as missing.

For esplora I did successfully check that timestamps increase for repeat syncs as long as a tx is unconfirmed. When the tx confirms, the timestamp is unchanged.

Note: I don't love the fact that every test needs to mine 100 regtest blocks, as that slows down CI. We should perhaps think about consolidating various assertions into more monolithic tests.

crates/electrum/tests/test_electrum.rs
```diff
@@ -190,3 +193,57 @@ fn tx_can_become_unconfirmed_after_reorg() -> Result<()> {
...
+    assert_eq!(tx.last_seen_unconfirmed, 0);
+
+    // Send tx 2
+    let txid = env.send(&addr, amt)?;
+
+    // Mine empty block and sync. TxGraph should record last seen unconfirmed
+    let _ = env.mine_empty_block();
+    env.wait_until_electrum_sees_block()?;
+    let ElectrumUpdate { relevant_txids, .. } = client.sync(chain.tip(), [spk], None, None, 5)?;
+    let missing = relevant_txids.missing_full_txs(graph.graph());
```
LLFourn: we want to avoid writing these kinds of tests that use `apply_update` and `missing_full_txs` because we're going to change the mechanisms on how this works. Tests against the blockchain backends should just check that it returns the right data (not that it is right after applying it as an update).

LLFourn (commented):
The changes here look exactly how I imagined they should. Thanks! But now that I see it, isn't it simpler just to have method like `update_last_seen_unconfirmed` on `TxGraph` which sets the last seen of all unanchored transactions? We can just call this on the update `TxGraph` and set it to `now` before applying it. This means we wouldn't need to pass it in, we can just mutate the `TxGraph` afterwards.

At least presently it doesn't look like we lose anything by doing this since the backends are both just setting whatever you tell them when you pass the time in. I was thinking that maybe esplora or electrum would tell you the last time it was seen in the API but I guess that's not the case. Mempool.space does have this api: https://mempool.space/docs/api/rest#get-transaction-times but that's "first seen" rather than last seen.

crates/electrum/tests/test_electrum.rs
ValuedMammal: You're right, it's a bit overkill.

ValuedMammal:
> The changes here look exactly how I imagined they should. Thanks! But now that I see it, isn't it simpler just to have method like `update_last_seen_unconfirmed` on `TxGraph` which sets the last seen of all unanchored transactions? We can just call this on the update `TxGraph` and set it to `now` before applying it. This means we wouldn't need to pass it in, we can just mutate the `TxGraph` afterwards.

I'll have another look at this.

ValuedMammal:
- Updated the PR description to reflect the approach using `TxGraph`, so the above comments are mostly outdated. This way has the advantage of not breaking the current API for electrum/esplora.

crates/electrum/src/electrum_ext.rs
```diff
@@ -46,8 +46,10 @@ impl RelevantTxids {
         let new_txs = client.batch_transaction_get(&missing)?;
         let mut graph = TxGraph::<ConfirmationHeightAnchor>::new(new_txs);
         for (txid, anchors) in self.0 {
-            if let Some(seen_at) = seen_at {
-                let _ = graph.insert_seen_at(txid, seen_at);
+            if anchors.is_empty() {
+                if let Some(seen_at) = seen_at {
```
notmandatory: ❓ I know it breaks the current API but shouldn't `seen_at` be required and non-optional here too?

ValuedMammal: I don't know whether it makes more sense to make this `seen_at` parameter required, or to remove the param and use the new `update_last_seen_unconfirmed` method on the update `TxGraph`.

LLFourn: I think it's fine as it is. Will likely be radically changed soon.

[EDIT] but when you did:
```rust
 l    let mut graph_update =
        relevant_txids.into_confirmation_time_tx_graph(&client, None, missing)?;
    let now = std::time::UNIX_EPOCH.elapsed().unwrap().as_secs();
    let _ = graph_update.update_last_seen_unconfirmed(now);
```
You are choosing to pass in `None` but then manually update it. If this is what you want to do in the examples then just remove the option from these functions. It's confusing to use both in the examples.

LLFourn (changes requested):
LGTM just need to make the electrum examples consistent (decide whether to pass it into the `into_confirmation_time_tx_graph` or just remove the Option and do it manually outside).

evanlinjin (commented):
This looks great! However, there may be a situation where the chain-source returns anchors that do not point to a block in the best chain. Those unconfirmed transactions will therefore have no "timestamp". Would it be better if `TxGraph::update_last_seen_unconfirmed` takes in a checkpoint tip so unconfirmed transactions are transactions that do not have an anchor that points to a checkpoint?

ValuedMammal:
> This looks great! However, there may be a situation where the chain-source returns anchors that do not point to a block in the best chain. Those unconfirmed transactions will therefore have no "timestamp". Would it be better if `TxGraph::update_last_seen_unconfirmed` takes in a checkpoint tip so unconfirmed transactions are transactions that do not have an anchor that points to a checkpoint?

I had not thought of that

crates/electrum/src/electrum_ext.rs
ValuedMammal: Fixed. I made `seen_at` a required parameter for `into_tx_graph`.

LLFourn:
> This looks great! However, there may be a situation where the chain-source returns anchors that do not point to a block in the best chain. Those unconfirmed transactions will therefore have no "timestamp". Would it be better if `TxGraph::update_last_seen_unconfirmed` takes in a checkpoint tip so unconfirmed transactions are transactions that do not have an anchor that points to a checkpoint?

`last_seen_unconfirmed` is for transactions that we've actually seen as unconfirmed in a fully validating mempool. Why would we set it for transactions that we've only seen as confirmed? Or are you saying esplora will allow us to determine that a tx is both in the mempool but was also confirmed in some non-canonical block? If that's the case then we have to pass in the timestamp into the sync/scan function so the logic can set both. The suggestion that we add and use `TxGraph::update_last_seen_unconfirmed` was based on the assumption that esplora and electrum would not allow us to distinguish this case.

evanlinjin:
@LLFourn the problem is we can't get blocks and transactions atomically. We either get blocks before getting transactions, or get transactions before getting blocks.

There might be a situation when a reorg happens between fetching txs and blocks. The tx might be sent back to mempool because it's not included in replacement block.

LLFourn:
> @LLFourn the problem is we can't get blocks and transactions atomically. We either get blocks before getting transactions, or get transactions before getting blocks.
> 
> There might be a situation when a reorg happens between fetching txs and blocks. The tx might be sent back to mempool because it's not included in replacement block.

...or it might not be if the re-org confirms a double spend. It might be in the mempool it might not be so we don't set `last_seen` because we don't know. Where's the problem?

crates/chain/src/tx_graph.rs
```diff
@@ -549,6 +549,32 @@ impl<A: Clone + Ord> TxGraph<A> {
         self.apply_update(update)
     }
 
+    /// Update the last seen unconfirmed time for all unanchored transactions.
+    ///
+    /// Returns the [`ChangeSet`] after applying any updates to `self`. Note that
+    /// [`TxGraph`] only keeps track of the latest `seen_at`.
```
LLFourn: It'd be nice if the docs explain more about why you would want to use this method. Are there some docs about how "last seen" works around that we could link to?

One question that comes to mind is why isn't it good enough to just set the `last_seen` to the most recent block header time. An example, showing how to get the right time from `std::time` would be good.

LLFourn (approved):
ACK b21ea0fe53710cf5ae037d9da48d48355e9f654e

From looking at it, the changes in `electrum_ext.rs` seem to do exactly the same thing that would be achieved if you just called `update_last_seen_unconfirmed` so it would make sense to just do that instead of passing in `seen_at`. I'm pretty sure @evanlinjin will change this soon so I don't think it matters that much.

crates/chain/src/tx_graph.rs
ValuedMammal: Pushed a docs only commit, let me know what you think.

@evanlinjin Is there a preference for keeping `into_tx_graph` as shown in c551dfba304464cf0814a9211b66d34e513f53b7 versus dropping the `seen_at` parameter in favor of the new `update_last_seen_unconfirmed`?

Also, passing in a chain tip is an interesting idea that could be its own method if we could demonstrate a use for it, but isn't `TxGraph` supposed to be agnostic to whether a tx is "canonical"?

LLFourn: docs are :100:. Your docs explain nicely why we don't want to use the block header time.

LLFourn (approved):
ACK 449dd9adec6aa95babfb339ab7ce815e5f7e0e74

crates/chain/src/tx_graph.rs
evanlinjin: @ValuedMammal do you mean dropping the `seen_at` parameter from `into_tx_graph`? I think that will be nice thanks!

evanlinjin (commented):
LGTM! Thanks for the quality work. Let's drop the `last_seen` parameter from `.into_tx_graph` and I'll ACK/merge.

ValuedMammal:
> LGTM! Thanks for the quality work. Let's drop the `last_seen` parameter from `.into_tx_graph` and I'll ACK/merge.

Done. Thank you to the reviewers @evanlinjin @LLFourn @notmandatory @rustaceanrob

evanlinjin (approved):
ACK a2a64ffb6e92baf46a912f36294f3f4f521a528a

Issue #256, opened 2020-12-30
Inconsistent checking of RBF rules


LLFourn:
[BIP125] rules 3 and 4 are not being checked properly according to my understanding:

https://github.com/bitcoindevkit/bdk/blob/c2b2da7601fb570952cfd858f3821b3a4fd87ec9/src/wallet/mod.rs#L730-L746

It should be checking rules 3 and 4 regardless of the fee policy chosen in txbuilder. 

[BIP125]: https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki#Implementation_Details

afilini:
I think we can kinda ignore rule 3 at the moment, since we don't support merging multiple unconfirmed transactions, so this:

> The replacement transaction pays an absolute fee of at least the sum paid by the original transactions.

For us basically means that we should just pay at least the previous fee.

I agree on rule 4 though, the `FeeAmount` branch should check for `*amount < details.fees + _something_`.

And now that I think about it, it's kind of a chicken-and-egg problem: rule 4 says that we should pay at least the size **of the new** transaction, but at this point we don't know it yet. So even the part above where `required_feerate` is computed is wrong.. Not sure what's the best way to handle this, I've gotta think about it a bit more..

evanlinjin:
Coin selection algorithms should take `min_absolute_fee` and `target_feerate` into consideration on the get-go.`bdk_core` will solve this issue, WIP here: https://github.com/LLFourn/bdk_core_staging/pull/21

nondiremanuel:
We discussed it in the call and decided to close this issue. Please comment if you think it should be re-opened

notmandatory:
Per discussion in dev call today this doesn't seem to be fixed yet.

LLFourn:
I volunteer to try and fix RBF API with `bdk_coin_select`. Don't have an ETA yet.

notmandatory:
Thanks @LLFourn , if you can fix with integration of `bdk_coin_select` that'd be great.

notmandatory:
@LLFourn Since we're not going to have have a `bdk_coin_select` crate for 1.0 can I defer this issue to the 2.0 milestone?

notmandatory:
Yes I'd like to make the extracted `bdk_coin_select` one of the priorities for 2.0 so if it's better to fix this together with that I'll update the milestone.

LLFourn:
[...]

PR #1386, opened 2024-03-25
Remove TxBuilder allow_shrinking() and unneeded context param

### LINKED ISSUE #1374: RBF API is awkward (remove allow_shrinking)
A fee bump shouldn't try to interpret the existing transaction and replicate it -- it should just create a tx with the appropriate fee that spends one or more of the existing transaction's outputs.

The builder API should just be `tx_builder.replace(tx)`. You could call it several times to replace several transactions.

_Originally posted by @LLFourn in https://github.com/bitcoindevkit/bdk/issues/1342#issuecomment-1940509301_

NB I think our RBF code is probably wrong but that's another issue. I've done a lot of work in the coin selection crate to get this correct. Getting the API right here and removing footgun `allow_shrinking` is what's important.

notmandatory:
### Description

Remove wallet::TxBuilder::allow_shrinking() and unneeded TxBuilder context param.

Fixes bitcoindevkit/bdk#1374 

### Notes to the reviewers

The allow_shrinking function was the only one using the TxBuilder FeeBump context and it's useful to have CreateTx context  functions available for building FeeBump transactions, see updated tests.

### Changelog notice

Changed

- Removed TxBuilder::allow_shrinking() function.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -59,7 +59,7 @@ pub use utils::IsDust;
 
 use coin_selection::DefaultCoinSelectionAlgorithm;
 use signer::{SignOptions, SignerOrdering, SignersContainer, TransactionSigner};
-use tx_builder::{BumpFee, CreateTx, FeePolicy, TxBuilder, TxParams};
+use tx_builder::{FeePolicy, TxBuilder, TxParams};
 use utils::{check_nsequence_rbf, After, Older, SecpCtx};
 
 use crate::descriptor::policy::BuildSatisfaction;
@@ -1293,12 +1293,11 @@ impl Wallet {
     /// ```
     ///
     /// [`TxBuilder`]: crate::TxBuilder
-    pub fn build_tx(&mut self) -> TxBuilder<'_, DefaultCoinSelectionAlgorithm, CreateTx> {
+    pub fn build_tx(&mut self) -> TxBuilder<'_, DefaultCoinSelectionAlgorithm> {
         TxBuilder {
             wallet: alloc::rc::Rc::new(core::cell::RefCell::new(self)),
             params: TxParams::default(),
             coin_selection: DefaultCoinSelectionAlgorithm::default(),
-            phantom: core::marker::PhantomData,
         }
     }
 
@@ -1684,7 +1683,7 @@ impl Wallet {
     pub fn build_fee_bump(
         &mut self,
         txid: Txid,
-    ) -> Result<TxBuilder<'_, DefaultCoinSelectionAlgorithm, BumpFee>, BuildFeeBumpError> {
+    ) -> Result<TxBuilder<'_, DefaultCoinSelectionAlgorithm>, BuildFeeBumpError> {
         let graph = self.indexed_graph.graph();
         let txout_index = &self.indexed_graph.index;
         let chain_tip = self.chain.tip().block_id();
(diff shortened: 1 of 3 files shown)


notmandatory:
@LLFourn I did a little more than just remove TxBuilder::allow_shrinking in this PR. If you think it's too much to remove the TxBuilder context I can put it back in, but then I'll need to either remove the two tests I updated or add the functions I needed to work for either context.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1431,7 +1430,6 @@ impl<D> Wallet<D> {
         };
 
         let (fee_rate, mut fee_amount) = match params.fee_policy.unwrap_or_default() {
-            //FIXME: see https://github.com/bitcoindevkit/bdk/issues/256
```
notmandatory: 📌 Removed because issue said it is fixed by 1.0 work but is it? 

I think we're supposed to make sure both the absolute fee amount AND fee rate is at least as big or bigger than the original transaction but it looks like we're only checking one or the other based on the `FeePolicy`.

notmandatory: I'm re-opening #256 since it seems it isn't fixed.

ValuedMammal:
LGTM

Conceptually, I think `allow_shrinking` is still a valuable method to have in the api for `TxBuilder`, but for the moment it's worth reconsidering how this will work in relation to RBF, coin selection, etc.

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -124,11 +123,10 @@ impl TxBuilderContext for BumpFee {}
 /// [`finish`]: Self::finish
 /// [`coin_selection`]: Self::coin_selection
 #[derive(Debug)]
-pub struct TxBuilder<'a, D, Cs, Ctx> {
+pub struct TxBuilder<'a, D, Cs> {
```
ValuedMammal: I see we're dropping the `Ctx` type parameter as it won't be needed, although we still have `pub trait TxBuilderContext` in this file - I assume as a reminder of how to approach it next time?

notmandatory: 🔧 Oh good catch, I should remove the `TxBuilderContext` also, if we need/want to add it back later it will still be in the commit history.

notmandatory: Fixed.

ValuedMammal:
ACK 1ab9204aee85cf041d9a27c8d927abcbfdb7c11f

ValuedMammal (approved):

oleonardolima (approved):
ACK 1ab9204aee85cf041d9a27c8d927abcbfdb7c11f

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -613,9 +596,79 @@ impl<'a, D, Cs, Ctx> TxBuilder<'a, D, Cs, Ctx> {
...
+        self.params.recipients.push((script_pubkey, amount));
+        self
+    }
+
+    /// Add data as an output, using OP_RETURN
+    pub fn add_data<T: AsRef<PushBytes>>(&mut self, data: &T) -> &mut Self {
+        let script = ScriptBuf::new_op_return(data);
+        self.add_recipient(script, 0u64);
+        self
+    }
```
oleonardolima: @notmandatory I did another pass on this one, and it raised a question: do we already want to use `bitcoin::Amount` here, as we're already touching/moving this code? wdyt ?

notmandatory: Good point, after I re-base I'll also update these `u64` to `Amount`.

oleonardolima (approved):
ACK 97e25080b7b99cc9d54eda11066500a72693d75b

ValuedMammal:
ACK 97e25080b7b99cc9d54eda11066500a72693d75b

Issue #1363, opened 2024-02-29
Remove persistence backend generics from Wallet<T> 


LLFourn:
The only reason we have `T` and don't just have a `Box<dyn PersitenceBackend>` internally is the custom error types. We could easily make standard error enum for persistence failures. This would remove the generic and therefore make using the wallet easier. Especially it would be easier to pass across FFI boundries.

ValuedMammal:
fixed by #1387

Issue #1155, opened 2023-10-07
Introduce `bdk_core` crate


evanlinjin:
**Describe the enhancement**  

This will contain common types shared across bdk crates.

**Use case**  

Better crate structure. Allows chain-source crates to be non-dependent on `bdk_chain`.

**Additional context**

* Discussion: https://github.com/bitcoindevkit/bdk/pull/1041#discussion_r1346777053

danielabrozzoni:
Concept ACK, the name `bdk_types` might be more explicative

LLFourn:
I think this crate will include some logic and not just types. E.g. the `CheckPoint` I think is the main thing that's caused the desire to have this. 

I did suggest this but fwiw I am not entirely convinced it's a good idea. In retrospect I like the idea of just not depending on any `bdk` crate and just using `impl Iterator<Item=(u32, BlockHash)>` instead of `CheckPoint`. I'd see if we cannot exhaust all other ways of doing this before adding this crate.

evanlinjin:
Right now the RPC emitter must emit blocks contiguously (cannot skip blocks), because we are emitting `(u32, Block)` and we construct the chain update from the block header.

If the emitter emits `(Checkpoint, Block)` instead, we will have the flexibility to skip blocks.

I think the ability to skip blocks is useful for block-by-block chain sources (especially with CBF). Also, if we go along with #1079 and still want to ability to specify a `start_height` (instead of starting from genesis), we need this as well.

evanlinjin:
> I did suggest this but fwiw I am not entirely convinced it's a good idea.

@LLFourn can you please expand on why it's not a good idea? Thanks

LLFourn:
I don't know that it's not a good idea I just am not convinced it is compared to other possible directions. Happy to leave this call to you if you are totally convinced that we cannot get a good API without it.

notmandatory:
On the one hand I hate adding yet another crate to the project, but if these types will be more or less unchanging and will be needed for the CBF client then I support it. I also think it's easier to read and talk about code with specific types like `(Checkpoint, Block)` instead of dealing with all the places we have a u32 or u64 and having to think about is this a height or a timestamp or maybe some sort of index into transaction or block data, etc. Can we also put a type alias for u64 unix epoch timestamps here?

danielabrozzoni:
> On the one hand I hate adding yet another crate to the project

[...]

PR #1387, opened 2024-03-26
fix(wallet): remove the generic from wallet

rustaceanrob:
### Description

The `PersistenceBackend` uses generics to describe errors returned while applying the change set to the persistence layer. This change removes generics wherever possible and introduces a new public error enum. Removing the generics from `PersistenceBackend` errors is the first step towards #1363 

*Update*: I proceeded with removing the generics from `Wallet` by introducing a `Box<dyn PersistenceBackend>` . 

### Notes to the reviewers

This one sort of blew up in the number of changes due to the use of generics for most of the `Wallet` error variants. The generics were only used for the persistence errors, so I removed the generics from higher level errors whenever possible. The error variants of `PersistenceBackend` may also be more expressive, but I will level that up for discussion and make any changes required.

### Changelog notice

- Changed `PersistenceBackend` errors to depend on the `anyhow` crate.
- Remove the generic `T` from `Wallet`

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bdk/Cargo.toml
+++ b/crates/bdk/Cargo.toml
@@ -13,6 +13,7 @@ edition = "2021"
 rust-version = "1.63"
 
 [dependencies]
+anyhow = { version = "1", default-features = false }
 rand = "^0.8"
 miniscript = { version = "11.0.0", features = ["serde"], default-features = false }
 bitcoin = { version = "0.31.0", features = ["serde", "base64", "rand-std"], default-features = false }
--- a/crates/bdk/src/wallet/error.rs
+++ b/crates/bdk/src/wallet/error.rs
@@ -47,11 +47,11 @@ impl std::error::Error for MiniscriptPsbtError {}
 /// Error returned from [`TxBuilder::finish`]
 ///
 /// [`TxBuilder::finish`]: crate::wallet::tx_builder::TxBuilder::finish
-pub enum CreateTxError<P> {
+pub enum CreateTxError {
     /// There was a problem with the descriptors passed in
     Descriptor(DescriptorError),
-    /// We were unable to write wallet data to the persistence backend
-    Persist(P),
+    /// We were unable to load wallet data from or write wallet data to the persistence backend
+    Persist(anyhow::Error),
     /// There was a problem while extracting and manipulating policies
     Policy(PolicyError),
     /// Spending policy is not compatible with this [`KeychainKind`]
@@ -119,17 +119,14 @@ pub enum CreateTxError<P> {
     MiniscriptPsbt(MiniscriptPsbtError),
 }
 
-impl<P> fmt::Display for CreateTxError<P>
-where
-    P: fmt::Display,
-{
+impl fmt::Display for CreateTxError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Self::Descriptor(e) => e.fmt(f),
             Self::Persist(e) => {
                 write!(
                     f,
-                    "failed to write wallet data to persistence backend: {}",
+                    "failed to load wallet data from or write wallet data to persistence backend: {}",
                     e
                 )
             }
@@ -214,38 +211,38 @@ where
     }
 }
 
-impl<P> From<descriptor::error::Error> for CreateTxError<P> {
+impl From<descriptor::error::Error> for CreateTxError {
     fn from(err: descriptor::error::Error) -> Self {
         CreateTxError::Descriptor(err)
     }
 }
 
-impl<P> From<PolicyError> for CreateTxError<P> {
+impl From<PolicyError> for CreateTxError {
     fn from(err: PolicyError) -> Self {
         CreateTxError::Policy(err)
     }
 }
 
-impl<P> From<MiniscriptPsbtError> for CreateTxError<P> {
+impl From<MiniscriptPsbtError> for CreateTxError {
     fn from(err: MiniscriptPsbtError) -> Self {
         CreateTxError::MiniscriptPsbt(err)
     }
 }
 
-impl<P> From<psbt::Error> for CreateTxError<P> {
+impl From<psbt::Error> for CreateTxError {
     fn from(err: psbt::Error) -> Self {
         CreateTxError::Psbt(err)
     }
 }
 
-impl<P> From<coin_selection::Error> for CreateTxError<P> {
+impl From<coin_selection::Error> for CreateTxError {
     fn from(err: coin_selection::Error) -> Self {
         CreateTxError::CoinSelection(err)
     }
 }
 
 #[cfg(feature = "std")]
-impl<P: core::fmt::Display + core::fmt::Debug> std::error::Error for CreateTxError<P> {}
+impl std::error::Error for CreateTxError {}
 
 #[derive(Debug)]
 /// Error returned from [`Wallet::build_fee_bump`]
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -82,12 +82,12 @@ const COINBASE_MATURITY: u32 = 100;
 ///
 /// [`signer`]: crate::signer
 #[derive(Debug)]
-pub struct Wallet<D = ()> {
+pub struct Wallet {
     signers: Arc<SignersContainer>,
     change_signers: Arc<SignersContainer>,
     chain: LocalChain,
     indexed_graph: IndexedTxGraph<ConfirmationTimeHeightAnchor, KeychainTxOutIndex<KeychainKind>>,
-    persist: Persist<D, ChangeSet>,
+    persist: Persist<ChangeSet>,
     network: Network,
     secp: SecpCtx,
 }
@@ -236,7 +236,7 @@ impl Wallet {
         Self::new(descriptor, change_descriptor, (), network).map_err(|e| match e {
             NewError::NonEmptyDatabase => unreachable!("mock-database cannot have data"),
             NewError::Descriptor(e) => e,
-            NewError::Write(_) => unreachable!("mock-write must always succeed"),
+            NewError::Persist(_) => unreachable!("mock-write must always succeed"),
         })
     }
 
@@ -251,15 +251,12 @@ impl Wallet {
             .map_err(|e| match e {
                 NewError::NonEmptyDatabase => unreachable!("mock-database cannot have data"),
                 NewError::Descriptor(e) => e,
-                NewError::Write(_) => unreachable!("mock-write must always succeed"),
+                NewError::Persist(_) => unreachable!("mock-write must always succeed"),
             })
     }
 }
 
-impl<D> Wallet<D>
-where
-    D: PersistBackend<ChangeSet, WriteError = core::convert::Infallible>,
-{
+impl Wallet {
     /// Infallibly return a derived address using the external descriptor, see [`AddressIndex`] for
     /// available address index selection strategies. If none of the keys in the descriptor are derivable
     /// (i.e. does not end with /*) then the same address will always be returned for any [`AddressIndex`].
@@ -296,19 +293,16 @@ where
 /// [`new`]: Wallet::new
 /// [`new_with_genesis_hash`]: Wallet::new_with_genesis_hash
 #[derive(Debug)]
-pub enum NewError<W> {
+pub enum NewError {
     /// Database already has data.
     NonEmptyDatabase,
     /// There was problem with the passed-in descriptor(s).
     Descriptor(crate::descriptor::DescriptorError),
     /// We were unable to write the wallet's data to the persistence backend.
-    Write(W),
+    Persist(anyhow::Error),
 }
 
-impl<W> fmt::Display for NewError<W>
-where
-    W: fmt::Display,
-{
+impl fmt::Display for NewError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             NewError::NonEmptyDatabase => write!(
@@ -316,13 +310,13 @@ where
                 "database already has data - use `load` or `new_or_load` methods instead"
             ),
             NewError::Descriptor(e) => e.fmt(f),
-            NewError::Write(e) => e.fmt(f),
+            NewError::Persist(e) => e.fmt(f),
         }
     }
 }
 
 #[cfg(feature = "std")]
-impl<W> std::error::Error for NewError<W> where W: core::fmt::Display + core::fmt::Debug {}
+impl std::error::Error for NewError {}
 
 /// The error type when loading a [`Wallet`] from persistence.
 ///
@@ -330,11 +324,11 @@ impl<W> std::error::Error for NewError<W> where W: core::fmt::Display + core::fm
 ///
 /// [`load`]: Wallet::load
 #[derive(Debug)]
-pub enum LoadError<L> {
+pub enum LoadError {
     /// There was a problem with the passed-in descriptor(s).
     Descriptor(crate::descriptor::DescriptorError),
     /// Loading data from the persistence backend failed.
-    Load(L),
+    Persist(anyhow::Error),
     /// Wallet not initialized, persistence backend is empty.
     NotInitialized,
     /// Data loaded from persistence is missing network type.
@@ -343,14 +337,11 @@ pub enum LoadError<L> {
     MissingGenesis,
 }
 
-impl<L> fmt::Display for LoadError<L>
-where
-    L: fmt::Display,
-{
+impl fmt::Display for LoadError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             LoadError::Descriptor(e) => e.fmt(f),
-            LoadError::Load(e) => e.fmt(f),
+            LoadError::Persist(e) => e.fmt(f),
             LoadError::NotInitialized => {
                 write!(f, "wallet is not initialized, persistence backend is empty")
             }
@@ -361,7 +352,7 @@ where
 }
 
 #[cfg(feature = "std")]
-impl<L> std::error::Error for LoadError<L> where L: core::fmt::Display + core::fmt::Debug {}
+impl std::error::Error for LoadError {}
 
 /// Error type for when we try load a [`Wallet`] from persistence and creating it if non-existent.
 ///
@@ -370,13 +361,11 @@ impl<L> std::error::Error for LoadError<L> where L: core::fmt::Display + core::f
 /// [`new_or_load`]: Wallet::new_or_load
 /// [`new_or_load_with_genesis_hash`]: Wallet::new_or_load_with_genesis_hash
 #[derive(Debug)]
-pub enum NewOrLoadError<W, L> {
+pub enum NewOrLoadError {
     /// There is a problem with the passed-in descriptor.
     Descriptor(crate::descriptor::DescriptorError),
-    /// Writing to the persistence backend failed.
-    Write(W),
-    /// Loading from the persistence backend failed.
-    Load(L),
+    /// Either writing to or loading from the persistence backend failed.
+    Persist(anyhow::Error),
     /// Wallet is not initialized, persistence backend is empty.
     NotInitialized,
     /// The loaded genesis hash does not match what was provided.
@@ -395,16 +384,15 @@ pub enum NewOrLoadError<W, L> {
     },
 }
 
-impl<W, L> fmt::Display for NewOrLoadError<W, L>
-where
-    W: fmt::Display,
-    L: fmt::Display,
-{
+impl fmt::Display for NewOrLoadError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             NewOrLoadError::Descriptor(e) => e.fmt(f),
-            NewOrLoadError::Write(e) => write!(f, "failed to write to persistence: {}", e),
-            NewOrLoadError::Load(e) => write!(f, "failed to load from persistence: {}", e),
+            NewOrLoadError::Persist(e) => write!(
+                f,
+                "failed to either write to or load from persistence, {}",
+                e
+            ),
             NewOrLoadError::NotInitialized => {
                 write!(f, "wallet is not initialized, persistence backend is empty")
             }
@@ -419,12 +407,7 @@ where
 }
 
 #[cfg(feature = "std")]
-impl<W, L> std::error::Error for NewOrLoadError<W, L>
-where
-    W: core::fmt::Display + core::fmt::Debug,
-    L: core::fmt::Display + core::fmt::Debug,
-{
-}
+impl std::error::Error for NewOrLoadError {}
 
 /// An error that may occur when inserting a transaction into [`Wallet`].
 #[derive(Debug)]
@@ -488,17 +471,14 @@ impl fmt::Display for ApplyBlockError {
 #[cfg(feature = "std")]
 impl std::error::Error for ApplyBlockError {}
 
-impl<D> Wallet<D> {
+impl Wallet {
     /// Initialize an empty [`Wallet`].
     pub fn new<E: IntoWalletDescriptor>(
         descriptor: E,
         change_descriptor: Option<E>,
-        db: D,
+        db: impl PersistBackend<ChangeSet> + Send + Sync + 'static,
         network: Network,
-    ) -> Result<Self, NewError<D::WriteError>>
-    where
-        D: PersistBackend<ChangeSet>,
-    {
+    ) -> Result<Self, NewError> {
         let genesis_hash = genesis_block(network).block_hash();
         Self::new_with_genesis_hash(descriptor, change_descriptor, db, network, genesis_hash)
     }
@@ -510,13 +490,10 @@ impl<D> Wallet<D> {
     pub fn new_with_genesis_hash<E: IntoWalletDescriptor>(
         descriptor: E,
         change_descriptor: Option<E>,
-        mut db: D,
+        mut db: impl PersistBackend<ChangeSet> + Send + Sync + 'static,
         network: Network,
         genesis_hash: BlockHash,
-    ) -> Result<Self, NewError<D::WriteError>>
-    where
-        D: PersistBackend<ChangeSet>,
-    {
+    ) -> Result<Self, NewError> {
         if let Ok(changeset) = db.load_from_persistence() {
             if changeset.is_some() {
                 return Err(NewError::NonEmptyDatabase);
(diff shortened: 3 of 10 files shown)


evanlinjin:
Thanks for working on this. In the future, please state in the ticket that you'll be working on it, otherwise we'll result in duplicate work.

rustaceanrob:
Will do! Thanks for the tip

rustaceanrob:
It is not obvious why this change is useful for removing `D` from Wallet. I am going to continue work towards that goal on this branch.

crates/chain/src/persist.rs
```diff
@@ -1,6 +1,26 @@
-use core::convert::Infallible;
-
 use crate::Append;
+use core::fmt;
+
+#[derive(Debug, Clone)]
+/// Errors returned by [`PersistBackend`] when staging and committing changes.
```
notmandatory: 🔧 This description doesn't capture that this can also be thrown for load errors. How about something simple like:
```suggestion
/// Errors returned by [`PersistBackend`] when writing or loading data.
```

rustaceanrob: Agreed. I am also working on more expressive error variants by wrapping the errors.

rustaceanrob:
Broke no_std with an error variant. Will fix soon.

crates/chain/src/persist.rs
```diff
@@ -1,33 +1,62 @@
...
-
 use crate::Append;
+use alloc::boxed::Box;
+use core::fmt;
+
+#[derive(Debug)]
+/// Errors returned by [`PersistBackend`] when writing or loading changes.
+pub enum PersistBackendError {
+    /// Happens when there is an standard IO error writing to the backend file.
+    IoError,
```
LLFourn: Shouldn't we include the `std::io::Error` here? The variant will need to be `#[cfg(feature = "std")]` of course.

crates/chain/src/persist.rs
```diff
@@ -1,33 +1,62 @@
...
+    /// Happens when there is an error iterating over the persistent backend file.
+    IterError,
```
LLFourn: This error variant doesn't make sense to me since it only applies to backends that iter (also it's not clear what the caller is meant to do with this error).

crates/chain/src/persist.rs
```diff
@@ -1,33 +1,62 @@
...
+    /// Occurs when there is an error loading the persistent backend file.
+    LoadError,
```
LLFourn: Should probably have different error enum for loading/appending.

crates/chain/src/persist.rs
rustaceanrob: Yes we should. I was unaware you are able to feature gate variants like this. The marco just goes above the variant?

LLFourn:
So after looking and thinking about this more closely I'd expect two error enums, one for each interaction:

```rust
enum LoadError {
   #[cfg(feature = "std")]
   StdIo(std::io::Error),
   Other(String),
}

enum CommitError {
    #[cfg(feature = "std")]
   StdIo(std::io::Error),
   Other(String),
   Inconsistent(String) // maybe?
}
```

But looking at it we don't really get much value from this structure unless we add specific error variants that occur with our changesets that we can report and react to. But our changeset API is (mostly) makes invalid representations impossible. In so far as invalid representations are possible they depend on the concrete changeset which the backend doesn't have access to.

The correct thing to do here I think is to return `Box<dyn core::error::Error>` (except that doesn't exist yet!). How about we just use `anyhow::Error` to do this. It would be better to not have this dependency but I think replicating what `anyhow` does (by `Box`ing our own error trait) is the next best solution and seems like a waste of time when almost every rust app depends on anyhow anyway. I think we should just accept it and wait for `core::error::Error`.

cc @evanlinjin

crates/chain/src/persist.rs
rustaceanrob: This error was intended to capture the error in `file_store/store.rs`. As far as I understand, not all users of `bdk_wallet` will use `file_store`, so I wasn't sure how to wrap it in the `PersistBackendError`. In `store`, this error occurs when applying the change set, so I guess `ApplyChangeError` could specify there is something wrong when attempting that action?

rustaceanrob:
Added my interpretation of how we would use `anyhow` here. Easy enough to add on or revert back.

crates/file_store/src/store.rs
```diff
@@ -173,7 +172,7 @@ where
     ///
     /// The truncation is to avoid the possibility of having a valid but inconsistent changeset
     /// directly after the appended changeset.
-    pub fn append_changeset(&mut self, changeset: &C) -> Result<(), io::Error> {
+    pub fn append_changeset(&mut self, changeset: &C) -> Result<(), anyhow::Error> {
```
evanlinjin: Instead of changing the signature of this method, could we just map the error when we use it in `PersistBackend::write_changes`?

crates/file_store/src/store.rs
```diff
@@ -206,13 +211,13 @@ pub struct AggregateChangesetsError<C> {
     pub iter_error: IterError,
 }
 
-impl<C> std::fmt::Display for AggregateChangesetsError<C> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        std::fmt::Display::fmt(&self.iter_error, f)
+impl<C> fmt::Display for AggregateChangesetsError<C> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        core::fmt::Display::fmt(&self.iter_error, f)
```
evanlinjin: Nit: could we either revert this change, or use `fmt::Formatter`/`fmt::Result`/`fmt::Display`?

crates/chain/src/persist.rs
```diff
@@ -1,33 +1,39 @@
...
-    pub fn new(backend: B) -> Self {
+    pub fn new(backend: Box<dyn PersistBackend<C>>) -> Self {
         Self {
             backend,
             stage: Default::default(),
         }
     }
 
-    /// Stage a `changeset` to be committed later with [`commit`].
+    /// Stage a `changeset` to be committed later stdwith [`commit`].
```
LLFourn: `stdwith` D:

LLFourn (commented):
This looks great. Main question is whether we're comfortable with this in `bdk_chain` or we should just make a new `bdk_persist` crate which defines this trait. It is rather orthogonal to everything else in `bdk_chain` and now it adds another dependency. If `bdk_core` existed (#1155) likely it wouldn't even depend on `bdk_chain`. "persist" could also be a feature flag in `bdk_chain`.

As a side note, for some reason we use this trait in the `bdk_chain` examples but I'm not sure why. They could easily be rewritten to not depend on this trait.

cc @thunderbiscuit wdyt about `anyhow::Error`. How do you handle errors in the FFI?

crates/bdk/src/wallet/mod.rs
```diff
@@ -371,13 +369,11 @@ impl<L> std::error::Error for LoadError<L> where L: core::fmt::Display + core::f
...
 #[derive(Debug)]
-pub enum NewOrLoadError<W, L> {
+pub enum NewOrLoadError {
     /// There is a problem with the passed-in descriptor.
     Descriptor(crate::descriptor::DescriptorError),
     /// Writing to the persistence backend failed.
-    Write(W),
-    /// Loading from the persistence backend failed.
-    Load(L),
+    Persist(anyhow::Error),
```
evanlinjin, line 368: We might need to update the docs as this variant can represent either a writing or loading error now.

crates/bdk/src/wallet/mod.rs
```diff
@@ -396,16 +392,11 @@ pub enum NewOrLoadError<W, L> {
...
-    W: fmt::Display,
-    L: fmt::Display,
-{
+impl fmt::Display for NewOrLoadError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             NewOrLoadError::Descriptor(e) => e.fmt(f),
-            NewOrLoadError::Write(e) => write!(f, "failed to write to persistence: {}", e),
-            NewOrLoadError::Load(e) => write!(f, "failed to load from persistence: {}", e),
+            NewOrLoadError::Persist(e) => write!(f, "failed to write to persistence, {}", e),
```
evanlinjin:
```suggestion
            NewOrLoadError::Persist(e) => write!(f, "failed to either write to or load from persistence, {}", e),
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -489,17 +475,14 @@ impl fmt::Display for ApplyBlockError {
...
 impl std::error::Error for ApplyBlockError {}
 
-impl<D> Wallet<D> {
+impl Wallet {
     /// Initialize an empty [`Wallet`].
     pub fn new<E: IntoWalletDescriptor>(
         descriptor: E,
         change_descriptor: Option<E>,
-        db: D,
+        db: Box<dyn PersistBackend<ChangeSet>>,
```
evanlinjin: Should we hide the `Box` from the caller and call `Box::new` internally? 
```suggestion
        db: impl PersistBackend<ChangeSet> + 'static,
```

crates/chain/src/persist.rs
```diff
@@ -1,33 +1,39 @@
...
+}
+
+impl<C> Persist<C>
 where
-    B: PersistBackend<C>,
     C: Default + Append,
 {
     /// Create a new [`Persist`] from [`PersistBackend`].
-    pub fn new(backend: B) -> Self {
+    pub fn new(backend: Box<dyn PersistBackend<C>>) -> Self {
```
evanlinjin: I think it is nicer to hide the `Box` from the caller here. I can't think of a benefit for exposing it?

```suggestion
    pub fn new(backend: impl PersistBackend<C> + 'static) -> Self {
```

evanlinjin (changes requested):
This looks great! However, I'm wondering whether it's better to introduce our own lightweight trait instead on using `anyhow` (which is another dependency).

```rust
pub trait PersistError: core::any::Any + Debug {}

pub fn an_example_method() -> Result<(), Box<dyn PersistError>> { todo!() }
```

The downsides of this is that the `PersistenceBackend` implementations may need to wrap their returned errors in a new type to implement `PersistError`. However, I see this as a minor compromise as they need to implement `PersistenceBackend` anyway?

cc. @LLFourn @ValuedMammal

crates/bdk/src/wallet/mod.rs
evanlinjin: @LLFourn what are your thoughts about combining write/load variants?

rustaceanrob:
Committed those changes for review

crates/bdk/src/wallet/mod.rs
LLFourn: Fine to keep them as one variant but as you said doc strings need to be updated if so. I don't see a strong motivation to merge them.

crates/bdk/Cargo.toml
```diff
@@ -13,6 +13,7 @@ edition = "2021"
 rust-version = "1.63"
 
 [dependencies]
+anyhow = { version = "1.0.81", default-features = false }
```
LLFourn: version should just be ` = "1" `

LLFourn:
> The downsides of this is that the `PersistenceBackend` implementations may need to wrap their returned errors in a new type to implement `PersistError`. However, I see this as a minor compromise as they need to implement `PersistenceBackend` anyway?
> 
> cc. @LLFourn @ValuedMammal

I don't think this is a minor downside for the implementer. It means wrapping all errors from the underlying backend in a type and implementing a trait on the wrapped types. Also then what about downcasting? What about backtraces? `anyhow::Error` just does everything right and there's no point in trying to replicate it. 

I'm leaning towards putting the persist trait in its own crate `bdk_persist`. It never made much sense in `bdk_chain`, it was just convenient to put it there but it no longer is.

crates/bdk/src/wallet/mod.rs
rustaceanrob: For the comment above the variant, how is "Either writing to or loading from the persistence backend failed." Verbatim of the `NewOrLoadError::Persist` error message.

rustaceanrob:
Should I continue by breaking out into a bdk persist crate? Or is that best left for another PR?

ValuedMammal:
> The correct thing to do here I think is to return `Box<dyn core::error::Error>` (except that doesn't exist yet!). How about we just use `anyhow::Error` to do this. It would be better to not have this dependency but I think replicating what `anyhow` does (by `Box`ing our own error trait) is the next best solution and seems like a waste of time when almost every rust app depends on anyhow anyway. I think we should just accept it and wait for `core::error::Error`.
> 

I'd be in favor of finding a solution that doesn't require anyhow, if possible. Currently, the `PersistBackend` errors are only constrained by `Debug`, so a new `PersistError` struct (or enum) could be something like

```rust
pub struct PersistError(pub Box<dyn fmt::Debug + Send + Sync>);
```

and make `PersistBackend` implementations return this concrete error, which still feels a bit backwards, but in practice could be as simple as writing a `From` conversion or constructing the type directly.

The persist module could also just live in `bdk` crate along with wallet, no?

evanlinjin:
@rustaceanrob we haven't reached a consensus on this. Feel free to put in your input.

LLFourn:
> > The correct thing to do here I think is to return `Box<dyn core::error::Error>` (except that doesn't exist yet!). How about we just use `anyhow::Error` to do this. It would be better to not have this dependency but I think replicating what `anyhow` does (by `Box`ing our own error trait) is the next best solution and seems like a waste of time when almost every rust app depends on anyhow anyway. I think we should just accept it and wait for `core::error::Error`.
> 

Replicating what anyhow does 100% is definitely not in scope. You can attach context anyhow errors and backtraces are automatically attached if your rust version is new enough. Of course we don't need this but it is nice to have.

> I'd be in favor of finding a solution that doesn't require anyhow, if possible. Currently, the `PersistBackend` errors are only constrained by `Debug`, so a new `PersistError` struct (or enum) could be something like
> 
> ```rust
> pub struct PersistError(pub Box<dyn fmt::Debug + Send + Sync>);
> ```
> 

And also `Display`. But we need to go further and make it `+ core::any::Any + 'static` so you can downcast the error to the concrete one. You will likely want to get the specific backend error at the application level e.g. the concrete postgres error. Then impl `From<E> where: E: std::error::Error`. This is quite a lot of stuff to maintain to avoid a dependency which virtually every application uses (and should use!). The question boils down to whether we want to maintain our own dynamic error system when there is already a standard choice for this very common problem in the rust ecosystem. I'd say no but it's less my decision to make than those who will actually be doing the work :)

> and make `PersistBackend` implementations return this concrete error, which still feels a bit backwards, but in practice could be as simple as writing a `From` conversion or constructing the type directly.
> 
> The persist module could also just live in `bdk` crate along with wallet, no?

Yes that could also work but I have a feeling the persist trait will be updated much less frequently than `bdk_wallet`. It wouldn't make sense to have to bump the version on the sqlite crate because some orthogonal API change happened in `bdk_wallet`.

thunderbiscuit:
> cc @thunderbiscuit wdyt about anyhow::Error. How do you handle errors in the FFI?

I'm not familiar with anyhow but reading up on it now. In general I favour not reimplementing the wheel, and think that using tried-and-true libraries is more robust over time (i.e., I am in favour of using anyhow instead of building similar behaviour in-house). One question I'd ask is if we add the anyhow dependency here, does it mean we'll have 2 approaches to errors across the codebase? Or would it be used in this single instance only?

As for FFI, we can't pass the errors directly (very!) unfortunately, so we roughly have to create new ones and map the Rust errors to those. This change would not impact us too much either way; do whatever is best for the Rust side of things and we'll adapt.

rustaceanrob:
I understand the reluctance to bring in a new dependency, but as @LLFourn described, the behavior that we are looking for would just be a re-implementation of what `anyhow` already does. `anyhow` also adds an additional end-user benefit of context and backtraces that feel out of scope for this PR, but improve the end-user developer experience imo. I am personally not seeing enough drawbacks compared to both the end-user and maintainer benefits gained from a drop in replacement, but I am new here of course! 

On structure, I am not seeing `chain` import `bdk`, so I'm not sure about moving it to the wallet side of things. This is type of common behavior that should rarely change seems like a prime candidate for `bdk_core`. If I were to weigh in on structure, I would leave things how they are for this PR and wait for a definitive restructuring decision, or start off `bdk_core` with `bdk_core/persist`. 

> As for FFI, we can't pass the errors directly (very!) unfortunately, so we roughly have to create new ones and map the Rust > errors to those. This change would not impact us too much either way; do whatever is best for the Rust side of things and > we'll adapt.

I have contributed to code with public `anyhow` result types along with UniFFI bindings. I will reach out to him to see what his experience was. Happy to help with UniFFI chores.

LLFourn (approved):
Fine to leave it here for now.  I still have a weak preference for moving it into `bdk_persist` rather than `bdk_core` (don't really want `bdk_core` to depend on `anyhow`). 

ACK bfadae799bad187b42732614b5c8738c5eb26819

Might be nice to clean up the commit history before merging. I think this should just be one commit on top of master if we can manage it.

example-crates/example_cli/src/lib.rs
```diff
@@ -29,7 +29,7 @@ pub type KeychainChangeSet<A> = (
     local_chain::ChangeSet,
     indexed_tx_graph::ChangeSet<A, keychain::ChangeSet<Keychain>>,
 );
-pub type Database<C> = Persist<Store<C>, C>;
+pub type Database<C> = Persist<C>;
```
ValuedMammal: nit: No need for this type alias now, just use `Persist` everywhere.

crates/bdk/src/wallet/mod.rs
```diff
@@ -19,6 +19,7 @@ use alloc::{
     sync::Arc,
     vec::Vec,
 };
+use anyhow::Result;
```
ValuedMammal: Instead of importing this globally, can we just prefix the `Result`s that use it with `anyhow::`. That way it's easy to spot where it is being used.

crates/bdk/Cargo.toml
ValuedMammal, line 1: anyhow is in here twice, so I guess we can drop it from dev-dependencies?

crates/chain/src/persist.rs
```diff
@@ -1,26 +1,33 @@
...
-#[derive(Debug)]
-pub struct Persist<B, C> {
-    backend: B,
+pub struct Persist<C> {
+    backend: Box<dyn PersistBackend<C>>,
     stage: C,
 }
 
-impl<B, C> Persist<B, C>
+impl<C: fmt::Debug> fmt::Debug for Persist<C> {
```
ValuedMammal, line 16: Alternatively, we can define the trait as

```rust
pub trait PersistBackend<C>: fmt::Debug {...}
```

and still be able to `#[derive(Debug)]` as before. I think you'd have to add a bound `C: fmt::Debug` for example in the trait impl for `Store`. Feel free to ignore if you think it's inconsequential.

ValuedMammal (commented):
I acknowledge all the points in the previous discussion, and I can offer these review comments :)

crates/chain/src/persist.rs
rustaceanrob: Just going to leave as is to keep my change minimal for this commit

crates/bdk/Cargo.toml
rustaceanrob: Dropped. Will adjust if CI fails. The library uses `anyhow = { version = "1", default-features = false }`, so hopefully the CI doesn't use any additional features.

rustaceanrob:
Should I squash commits locally? I only ask because I think there's also a feature to do it on the merge GUI.

crates/bdk/Cargo.toml
ValuedMammal: yep I see now. maybe better to keep it as dev dependency as well in order to get default features.

crates/bdk/src/keys/mod.rs
```diff
@@ -632,7 +632,6 @@ pub trait GeneratableKey<Ctx: ScriptContext>: Sized {
     /// Generate a key given the options with a random entropy
     fn generate(options: Self::Options) -> Result<GeneratedKey<Self, Ctx>, Self::Error> {
         use rand::{thread_rng, Rng};
-
```
ValuedMammal: nit: Seems unnecessary

crates/bdk/src/wallet/error.rs
```diff
@@ -214,38 +211,44 @@ where
...
 }
 
-impl<P> From<coin_selection::Error> for CreateTxError<P> {
+impl From<coin_selection::Error> for CreateTxError {
     fn from(err: coin_selection::Error) -> Self {
         CreateTxError::CoinSelection(err)
     }
 }
 
+impl From<anyhow::Error> for CreateTxError {
```
ValuedMammal: Do we actually need this? I was able to build and test without it.

crates/bdk/src/wallet/mod.rs
```diff
@@ -232,11 +232,11 @@ impl Wallet {
         descriptor: E,
         change_descriptor: Option<E>,
         network: Network,
-    ) -> Result<Self, DescriptorError> {
+    ) -> Result<Self, descriptor::error::Error> {
```
ValuedMammal: nit: `DescriptorError` looks nicer.

crates/file_store/src/store.rs
```diff
@@ -25,16 +24,17 @@ impl<C> PersistBackend<C> for Store<C>
...
-
-    type LoadError = IterError;
-
-    fn write_changes(&mut self, changeset: &C) -> Result<(), Self::WriteError> {
-        self.append_changeset(changeset)
+    fn write_changes(&mut self, changeset: &C) -> anyhow::Result<()> {
+        self.append_changeset(changeset).map_err(|error| {
+            anyhow::Error::msg(error.to_string())
+                .context("Failed to write changes to the persistence backend.")
+        })
```
ValuedMammal: I would use `anyhow` macro.

```suggestion
        self.append_changeset(changeset)
            .map_err(|e| anyhow!(e).context("failed to write changes to persistence backend"))
```

crates/file_store/src/store.rs
```diff
@@ -25,16 +24,17 @@ impl<C> PersistBackend<C> for Store<C>
...
     }
 
-    fn load_from_persistence(&mut self) -> Result<Option<C>, Self::LoadError> {
-        self.aggregate_changesets().map_err(|e| e.iter_error)
+    fn load_from_persistence(&mut self) -> anyhow::Result<Option<C>> {
+        self.aggregate_changesets().map_err(|e| {
+            anyhow::Error::msg(e.to_string()).context("Error loading from persistence backend.")
+        })
```
ValuedMammal: We still want to map to `iter_error`.

```suggestion
        self.aggregate_changesets()
            .map_err(|e| anyhow!(e.iter_error).context("error loading from persistence backend"))
```

ValuedMammal (changes requested):
Sorry, I have a couple more notes.

Primary suggestions are for `store.rs`, and I'm also wondering whether the `From` impl in `wallet::error` module is needed. Any nits are definitely not blockers.

As far as squashing, my preference is to use [`git rebase --interactive`](https://git-scm.com/docs/git-rebase).

crates/bdk/src/wallet/error.rs
rustaceanrob: When previously adding new variants I thought it might be useful. Agree that this doesn't make much sense to add at the moment.

rustaceanrob:
Squashed to [c958fb4](https://github.com/bitcoindevkit/bdk/commit/c958fb468702203baf7e1c8f5f0ac8d89d3b9505) with changes suggested from @ValuedMammal. Should be ready for CI

crates/chain/src/persist.rs
```diff
@@ -1,26 +1,32 @@
-use core::convert::Infallible;
-
 use crate::Append;
+use alloc::boxed::Box;
+use core::fmt;
 
 /// `Persist` wraps a [`PersistBackend`] (`B`) to create a convenient staging area for changes (`C`)
```
jirijakes: The documentation is outdated now, `B` no more.

jirijakes (commented):
This change silently removes `Sync` and `Send` auto traits from `Persist` and therefore from anything that uses it (e. g. `Wallet`). 

One way to deal with it could be:

```rs
pub struct Persist<C> {
    backend: Box<dyn PersistBackend<C> + Send + Sync>,
    stage: C,
}
```

but that would require every backend to be `Send` + `Sync`, even when not needed.

I am not sure how big of a problem this is, but given that `Wallet` until now tried to be `Sync` + `Send`, it may be an issue.

rustaceanrob:
As far as I can tell `Wallet` is already [not `Sync`](https://docs.rs/bdk/latest/bdk/wallet/struct.Wallet.html#synthetic-implementations)? Let me know if I am misunderstanding something!

jirijakes:
> As far as I can tell `Wallet` is already not `Sync`?

It is in [alpha.8](https://docs.rs/bdk/1.0.0-alpha.8/bdk/wallet/struct.Wallet.html#impl-Sync-for-Wallet%3CD%3E).

But I really don't know whether this was some accident or deliberate action. I don't know whether this is something to care about or not.

rustaceanrob:
Interesting. I tried messing with it and I am not aware of any other work-arounds besides the one you proposed. Between `Sync` and `Send`, not having `Send` seems to be the most problematic as earlier releases of `Wallet` were `Send`. Is it reasonable for all backends to be `Send` and `Sync`? @LLFourn @evanlinjin

ValuedMammal:
That's a great point @jirijakes what made you realize that? I mean is it evident from looking at the code, or is there something more subtle going on? I do think it's generally desirable to maintain thread safety of the wallet.

jirijakes:
> what made you realize that?

Rather a coincidence. A few days ago I read articles (mainly [this](https://predr.ag/blog/semver-in-rust-tooling-breakage-and-edge-cases/)) about this kind of semver breakage. Having the articles fresh in mind and seeing `Box`, I was curious.

rustaceanrob:
I updated the PR to include the `Sync + Send` trait bounds on the `PersistBackend` when creating a new `Wallet`. Although this may be a breaking change for wallets that implement their own store, they should easily be able to enforce `Sync + Send` on their `ChangeSet` generic. The benefit gained of `Wallet` being thread safe feels like a reasonable tradeoff, especially in the server context. Thank you @jirijakes for that insight.

LLFourn:
Yeah this change  implies that we must force `PersistBackend` to be `Send + Sync`. Before `Send` and `Sync` depended on `PersistBackend` impl. Now since we are using a trait object we have to make a decision for the user and the right decision is to add `Send + Sync` to the trait object. It's a slight regression but easily worth it for what we get. For things that are not `Send` or `Sync` you can always just make a channel or something like that to communicate with it (where the channel sender would be `Send + Sync`).

evanlinjin (approved):
ACK e51af49ffa951e14203ac2b465ade351dd90f6cd

This is good work. However, I'm really not a fan of introducing the `anyhow` dependency to the `bdk_chain` crate.

I think we should create a separate RP which introduces a separate `bdk_persist` crate.

PR #1390, opened 2024-03-27
Make Wallet require a change descriptor

### LINKED ISSUE #1383: Wallet must have unique Internal and External descriptors
Per discussion in our last backlog review call, since the bdk wallet APIs are opinionated about how to correctly build and use an on-chain wallet we should make unique internal (change address) and external (receive address) descriptors mandatory when creating a new `Wallet`. This will also allow us to remove current behavior of getting new change addresses from the external descriptor if no internal descriptor is given. Using external addresses for internal (change) breaks other wallet features like asking the tx builder to create transactions that don't spend change.  

If a user wants to create a wallet without an internal descriptor they can do what ever they want using the lower level `bdk_chain` and other crates.

ValuedMammal:
All `Wallet` constructors are modified to require a change descriptor, where previously it was optional. Additionally we enforce uniqueness of the change descriptor to avoid ambiguity when deriving scripts and ensure the wallet will always have two distinct keystores.

Notable changes

* Add error `DescriptorError::ExternalAndInternalAreTheSame`
* Remove error `CreateTxError::ChangePolicyDescriptor`
* No longer rely on `map_keychain`

fixes #1383

### Changelog notice

Changed:

Constructing a Wallet now requires two distinct descriptors.

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -242,7 +242,7 @@ pub enum LoadError {
     /// Data loaded from persistence is missing genesis hash.
     MissingGenesis,
     /// Data loaded from persistence is missing descriptor.
-    MissingDescriptor,
+    MissingDescriptor(KeychainKind),
 }
 
 impl fmt::Display for LoadError {
@@ -255,7 +255,9 @@ impl fmt::Display for LoadError {
             }
             LoadError::MissingNetwork => write!(f, "loaded data is missing network type"),
             LoadError::MissingGenesis => write!(f, "loaded data is missing genesis hash"),
-            LoadError::MissingDescriptor => write!(f, "loaded data is missing descriptor"),
+            LoadError::MissingDescriptor(k) => {
+                write!(f, "loaded data is missing descriptor for keychain {k:?}")
+            }
         }
     }
 }
@@ -517,14 +519,14 @@ impl Wallet {
             .indexer
             .keychains_added
             .get(&KeychainKind::External)
-            .ok_or(LoadError::MissingDescriptor)?
+            .ok_or(LoadError::MissingDescriptor(KeychainKind::External))?
             .clone();
         let change_descriptor = changeset
             .indexed_tx_graph
             .indexer
             .keychains_added
             .get(&KeychainKind::Internal)
-            .ok_or(LoadError::MissingDescriptor)?
+            .ok_or(LoadError::MissingDescriptor(KeychainKind::Internal))?
             .clone();
 
         let (signers, change_signers) =
@@ -596,10 +598,12 @@ impl Wallet {
                         expected: genesis_hash,
                         got: None,
                     },
-                    LoadError::MissingDescriptor => NewOrLoadError::LoadedDescriptorDoesNotMatch {
-                        got: None,
-                        keychain: KeychainKind::External,
-                    },
+                    LoadError::MissingDescriptor(keychain) => {
+                        NewOrLoadError::LoadedDescriptorDoesNotMatch {
+                            got: None,
+                            keychain,
+                        }
+                    }
                 })?;
                 if wallet.network != network {
                     return Err(NewOrLoadError::LoadedNetworkDoesNotMatch {
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -617,10 +617,10 @@ impl Wallet {
                 let (expected_descriptor, expected_descriptor_keymap) = descriptor
                     .into_wallet_descriptor(&wallet.secp, network)
                     .map_err(NewOrLoadError::Descriptor)?;
-                let wallet_descriptor = wallet.public_descriptor(KeychainKind::External).cloned();
-                if wallet_descriptor != Some(expected_descriptor.clone()) {
+                let wallet_descriptor = wallet.public_descriptor(KeychainKind::External);
+                if wallet_descriptor != &expected_descriptor {
                     return Err(NewOrLoadError::LoadedDescriptorDoesNotMatch {
-                        got: wallet_descriptor,
+                        got: Some(wallet_descriptor.clone()),
                         keychain: KeychainKind::External,
                     });
                 }
@@ -644,11 +644,10 @@ impl Wallet {
                     change_descriptor
                         .into_wallet_descriptor(&wallet.secp, network)
                         .map_err(NewOrLoadError::Descriptor)?;
-                let wallet_change_descriptor =
-                    wallet.public_descriptor(KeychainKind::Internal).cloned();
-                if wallet_change_descriptor != Some(expected_change_descriptor.clone()) {
+                let wallet_change_descriptor = wallet.public_descriptor(KeychainKind::Internal);
+                if wallet_change_descriptor != &expected_change_descriptor {
                     return Err(NewOrLoadError::LoadedDescriptorDoesNotMatch {
-                        got: wallet_change_descriptor,
+                        got: Some(wallet_change_descriptor.clone()),
                         keychain: KeychainKind::Internal,
                     });
                 }
@@ -1858,22 +1857,24 @@ impl Wallet {
             KeychainKind::Internal => &self.change_signers,
         };
 
-        match self.public_descriptor(keychain) {
-            Some(desc) => Ok(desc.extract_policy(signers, BuildSatisfaction::None, &self.secp)?),
-            None => Ok(None),
-        }
+        self.public_descriptor(keychain).extract_policy(
+            signers,
+            BuildSatisfaction::None,
+            &self.secp,
+        )
     }
 
     /// Return the "public" version of the wallet's descriptor, meaning a new descriptor that has
     /// the same structure but with every secret key removed
     ///
     /// This can be used to build a watch-only version of a wallet
-    pub fn public_descriptor(&self, keychain: KeychainKind) -> Option<&ExtendedDescriptor> {
+    pub fn public_descriptor(&self, keychain: KeychainKind) -> &ExtendedDescriptor {
         self.indexed_graph
             .index
             .keychains()
             .find(|(k, _)| *k == &keychain)
             .map(|(_, d)| d)
+            .expect("keychain must exist")
     }
 
     /// Finalize a PSBT, i.e., for each input determine if sufficient data is available to pass
@@ -1980,7 +1981,7 @@ impl Wallet {
 
     /// Returns the descriptor used to create addresses for a particular `keychain`.
     pub fn get_descriptor_for_keychain(&self, keychain: KeychainKind) -> &ExtendedDescriptor {
-        self.public_descriptor(keychain).expect("keychain exists")
+        self.public_descriptor(keychain)
     }
 
     /// The derivation index of this wallet. It will return `None` if it has not derived any addresses.
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -166,7 +166,7 @@ impl Wallet {
     /// Creates a wallet that does not persist data.
     pub fn new_no_persist<E: IntoWalletDescriptor>(
         descriptor: E,
-        change_descriptor: Option<E>,
+        change_descriptor: E,
         network: Network,
     ) -> Result<Self, DescriptorError> {
         Self::new(descriptor, change_descriptor, (), network).map_err(|e| match e {
@@ -179,7 +179,7 @@ impl Wallet {
     /// Creates a wallet that does not persist data, with a custom genesis hash.
     pub fn new_no_persist_with_genesis_hash<E: IntoWalletDescriptor>(
         descriptor: E,
-        change_descriptor: Option<E>,
+        change_descriptor: E,
         network: Network,
         genesis_hash: BlockHash,
     ) -> Result<Self, crate::descriptor::DescriptorError> {
@@ -398,7 +398,7 @@ impl Wallet {
     /// Initialize an empty [`Wallet`].
     pub fn new<E: IntoWalletDescriptor>(
         descriptor: E,
-        change_descriptor: Option<E>,
+        change_descriptor: E,
         db: impl PersistBackend<ChangeSet> + Send + Sync + 'static,
         network: Network,
     ) -> Result<Self, NewError> {
@@ -412,7 +412,7 @@ impl Wallet {
     /// for syncing from alternative networks.
     pub fn new_with_genesis_hash<E: IntoWalletDescriptor>(
         descriptor: E,
-        change_descriptor: Option<E>,
+        change_descriptor: E,
         mut db: impl PersistBackend<ChangeSet> + Send + Sync + 'static,
         network: Network,
         genesis_hash: BlockHash,
@@ -524,7 +524,8 @@ impl Wallet {
             .indexer
             .keychains_added
             .get(&KeychainKind::Internal)
-            .cloned();
+            .ok_or(LoadError::MissingDescriptor)?
+            .clone();
 
         let (signers, change_signers) =
             create_signers(&mut index, &secp, descriptor, change_descriptor, network)
@@ -551,7 +552,7 @@ impl Wallet {
     /// This method will fail if the loaded [`Wallet`] has different parameters to those provided.
     pub fn new_or_load<E: IntoWalletDescriptor>(
         descriptor: E,
-        change_descriptor: Option<E>,
+        change_descriptor: E,
         db: impl PersistBackend<ChangeSet> + Send + Sync + 'static,
         network: Network,
     ) -> Result<Self, NewOrLoadError> {
@@ -573,7 +574,7 @@ impl Wallet {
     /// useful for syncing from alternative networks.
     pub fn new_or_load_with_genesis_hash<E: IntoWalletDescriptor>(
         descriptor: E,
-        change_descriptor: Option<E>,
+        change_descriptor: E,
         mut db: impl PersistBackend<ChangeSet> + Send + Sync + 'static,
         network: Network,
         genesis_hash: BlockHash,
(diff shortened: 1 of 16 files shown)


ValuedMammal:
Do we want to keep the change descriptor optional for `example_cli`? I'm guessing it won't hurt anything since we don't actually use the wallet.

crates/bdk/examples/compiler.rs
```diff
@@ -42,12 +42,23 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     // Create a `wsh` type descriptor from the policy.
     // `policy.compile()` returns the resulting miniscript from the policy.
-    let descriptor = Descriptor::new_wsh(policy.compile()?)?;
+    let descriptor = Descriptor::new_wsh(policy.compile()?)?.to_string();
 
-    println!("Compiled into following Descriptor: \n{}", descriptor);
+    println!("Compiled into Descriptor: \n{}", descriptor);
```
storopoli: We are using edition 2021, so you can do:
```suggestion
    println!("Compiled into Descriptor: \n{descriptor}");
```

crates/bdk/examples/compiler.rs
```diff
@@ -42,12 +42,23 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+    // Do the same for another (internal) keychain
+    let policy_str = "or(10@thresh(2,pk(029ffbe722b147f3035c87cb1c60b9a5947dd49c774cc31e94773478711a929ac0),pk(025f05815e3a1a8a83bfbb03ce016c9a2ee31066b98f567f6227df1d76ec4bd143),pk(025625f41e4a065efc06d5019cbbd56fe8c07595af1231e7cbc03fafb87ebb71ec)),1@and(pk(03deae92101c790b12653231439f27b8897264125ecb2f46f48278603102573165),older(12960)))";
+    println!("Compiling internal policy: \n{}", policy_str);
+
+    let policy = Concrete::<String>::from_str(policy_str)?;
+    let internal_descriptor = Descriptor::new_wsh(policy.compile()?)?.to_string();
+    println!(
+        "Compiled into internal Descriptor: \n{}",
+        internal_descriptor
+    );
```
storopoli: We are using edition 2021, so you can do:
```suggestion
    println!("Compiled into internal Descriptor: \n{internal_descriptor}");
```

crates/bdk/src/descriptor/error.rs
```diff
@@ -79,6 +84,7 @@ impl fmt::Display for Error {
             Self::Pk(err) => write!(f, "Key-related error: {}", err),
             Self::Miniscript(err) => write!(f, "Miniscript error: {}", err),
             Self::Hex(err) => write!(f, "Hex decoding error: {}", err),
+            Self::NonUniqueInternal(d) => write!(f, "Non unique change descriptor: {}", d),
```
storopoli: Same for all these `write!`s:

```suggestion
            Self::NonUniqueInternal(d) => write!(f, "Non unique change descriptor: {d}"),
```

crates/bdk/src/descriptor/template.rs
```diff
@@ -79,7 +79,9 @@ impl<T: DescriptorTemplate> IntoWalletDescriptor for T {
 ///
 /// let key =
 ///     bitcoin::PrivateKey::from_wif("cTc4vURSzdx6QE6KVynWGomDbLaA75dNALMNyfjh3p8DRRar84Um")?;
-/// let mut wallet = Wallet::new_no_persist(P2Pkh(key), None, Network::Testnet)?;
+/// let key2 =
```
storopoli: Isn't it better to give better names than `key` and `key2`?

crates/bdk/src/descriptor/template.rs
```diff
@@ -107,7 +109,9 @@ impl<K: IntoDescriptorKey<Legacy>> DescriptorTemplate for P2Pkh<K> {
 ///
 /// let key =
 ///     bitcoin::PrivateKey::from_wif("cTc4vURSzdx6QE6KVynWGomDbLaA75dNALMNyfjh3p8DRRar84Um")?;
-/// let mut wallet = Wallet::new_no_persist(P2Wpkh_P2Sh(key), None, Network::Testnet)?;
+/// let key2 =
```
storopoli: Isn't it better to give better names than `key` and `key2`?

crates/bdk/src/descriptor/template.rs
```diff
@@ -136,7 +140,9 @@ impl<K: IntoDescriptorKey<Segwitv0>> DescriptorTemplate for P2Wpkh_P2Sh<K> {
 ///
 /// let key =
 ///     bitcoin::PrivateKey::from_wif("cTc4vURSzdx6QE6KVynWGomDbLaA75dNALMNyfjh3p8DRRar84Um")?;
-/// let mut wallet = Wallet::new_no_persist(P2Wpkh(key), None, Network::Testnet)?;
+/// let key2 =
+///     bitcoin::PrivateKey::from_wif("cVpPVruEDdmutPzisEsYvtST1usBR3ntr8pXSyt6D2YYqXRyPcFW")?;
```
storopoli: Isn't it better to give better names than `key` and `key2`?

crates/bdk/src/descriptor/template.rs
```diff
@@ -164,7 +170,9 @@ impl<K: IntoDescriptorKey<Segwitv0>> DescriptorTemplate for P2Wpkh<K> {
 ///
 /// let key =
 ///     bitcoin::PrivateKey::from_wif("cTc4vURSzdx6QE6KVynWGomDbLaA75dNALMNyfjh3p8DRRar84Um")?;
-/// let mut wallet = Wallet::new_no_persist(P2TR(key), None, Network::Testnet)?;
+/// let key2 =
```
storopoli: Isn't it better to give better names than `key` and `key2`?

crates/bdk/src/wallet/export.rs
```diff
@@ -268,13 +262,14 @@ mod test {
     #[test]
     #[should_panic(expected = "Incompatible change descriptor")]
     fn test_export_no_change() {
-        // This wallet explicitly doesn't have a change descriptor. It should be impossible to
+        // This wallet has a bad change descriptor (no glob). It should be impossible to
```
storopoli: "bad" is arbitrary. Can't we be more technical here.

storopoli (commented):
Some comments

crates/bdk/src/descriptor/template.rs
ValuedMammal: How about `key_a`, `key_b`

storopoli: You've used `descriptor` and `descriptor_change` in the PR somewhere.
I like that. It is informative.
Because here `key` is the receiving/external "address type",
and `key2` is the change/internal "address type.

casey-bowman:
If two descriptors are the same except that each has a different extended public key that is replaced by the corresponding extended private key, do these qualify as distinct?

Also, to clarify, the receive descriptors for any two wallets must also be distinct with this change, too, correct?
UPDATE - Oh, I see the issue #1383

crates/bdk/examples/compiler.rs
ValuedMammal: I opened #1394 to discuss code style.

crates/bdk/examples/compiler.rs
ValuedMammal: https://github.com/bitcoindevkit/bdk/pull/1390#discussion_r1545444089

crates/bdk/src/descriptor/error.rs
ValuedMammal: https://github.com/bitcoindevkit/bdk/pull/1390#discussion_r1545444089

crates/bdk/src/descriptor/template.rs
ValuedMammal: Readers of the code will see `key` is a bitcoin private key.

To expand on that, here `key` is being used to build a descriptor "template" which can then be passed to `Wallet::new_no_persist`, as all `DescriptorTemplate` implementors are also `IntoWalletDescriptor`.

crates/bdk/src/descriptor/template.rs
ValuedMammal: https://github.com/bitcoindevkit/bdk/pull/1390#discussion_r1545445207

crates/bdk/src/descriptor/template.rs
ValuedMammal: https://github.com/bitcoindevkit/bdk/pull/1390#discussion_r1545445207

crates/bdk/src/descriptor/template.rs
ValuedMammal: https://github.com/bitcoindevkit/bdk/pull/1390#discussion_r1545445207

crates/bdk/src/wallet/export.rs
ValuedMammal: The technical reason in this case is "no glob". The absence of a glob `*` a.k.a. "wildcard" in the change descriptor is enough for `export_wallet` to hit the expected error.

crates/bdk/src/descriptor/template.rs
evanlinjin: `key_external` and `key_internal`?

crates/bdk/src/wallet/mod.rs
```diff
@@ -2558,22 +2531,31 @@ fn create_signers<E: IntoWalletDescriptor>(
...
+    index.add_keychain(KeychainKind::Internal, descriptor);
+
+    // Check we didn't try to add an overlapping keychain
+    if index.spk_at_index(KeychainKind::Internal, 0).is_none() {
+        let change_descriptor = index
+            .keychains()
+            .get(&KeychainKind::Internal)
+            .expect("we added the keychain");
+        return Err(DescriptorError::NonUnique(change_descriptor.to_string()));
+    }
```
evanlinjin: I don't quite understand how this checks that the internal/external keychains are overlapping. Right now the wallet creates a `KeychainTxOutIndex` with a lookahead value of 25 (this is the default value), so the if statement will always evaluate to `false`.

evanlinjin (commented):
I'm concerned with the way we are checking whether the 2 keychains are "distinct". I have a few questions.

1. How can we guarantee that `into_wallet_descriptor` returns the same public-descriptor string for all non-distinct descriptors?
2. How can we ensure that a non-wildcard descriptor and a wildcard descriptor has no overlaps (the non-wildcard descriptor's spk can be derived from the wildcard descriptor).

For 2. also refer to my comment here: https://github.com/bitcoindevkit/bdk/issues/1383#issuecomment-2029651474

crates/bdk/src/wallet/mod.rs
ValuedMammal: you're right, it won't catch overlap if the non-wildcard descriptor has a derivation index greater than `DEFAULT_LOOKAHEAD`, so that's a problem.

ValuedMammal: We could derive the spk for the non-wildcard descriptor, and then do an exhaustive search in the wildcard descriptor for an index matching that spk. It would of course be more efficient if we could infer the definite derivation index.

ValuedMammal: Reverted that idea. My current thinking is to do nothing and document that it is a user error to introduce a non unique keychain.

LLFourn: I agree with this. We don't need to provide guarantees here. We can check equality of descriptors but other than that it's up to the programmer not to introduce "overlapping" descriptors.

crates/bdk/src/descriptor/error.rs
```diff
@@ -42,6 +45,8 @@ pub enum Error {
     Miniscript(miniscript::Error),
     /// Hex decoding error
     Hex(bitcoin::hashes::hex::Error),
+    /// The provided wallet descriptors are not unique
+    NonUnique(ExtendedDescriptor),
```
LLFourn:
```suggestion
    ExternalAndInternalAreTheSame,
```
+more obvious name. I don't think we need the Extended descriptor in there. It doesn't help you debug anything really.

LLFourn (approved):
ACK 04c8750f54075014af10b92aa800fb7dc596cfa7

I'm going to be away so not going to get another chance to look at this. It LGTM. Feel free to keep shedding. Thanks for taking on this rather tedious PR.

ValuedMammal:
Small changes

- Fixed `key_*` names in descriptor template docs
- Changed the new descriptor error to just `ExternalAndInternalAreTheSame`

ValuedMammal:
>     1. How can we guarantee that `into_wallet_descriptor` returns the same public-descriptor string for all non-distinct descriptors?
> 
>     2. How can we ensure that a non-wildcard descriptor and a wildcard descriptor has no overlaps (the non-wildcard descriptor's spk can be derived from the wildcard descriptor).
> 

To point 1) I don't think you're objecting to checking equality for two values of `Descriptor<DescriptorPublicKey>`, which is only true for identical descriptors, i.e. ones that derive 100% of the same scripts. Maybe this was addressed in the discussion around point 2?

Note that the public descriptor will be common whether the wallet is given a public or private descriptor. So for example this should pass:

```rust
#[test]
fn same_descriptor_public_private() {
    // public + private of same descriptor should fail to create wallet
    let desc = "wpkh(tprv8ZgxMBicQKsPdcAqYBpzAFwU5yxBUo88ggoBqu1qPcHUfSbKK1sKMLmC7EAk438btHQrSdu3jGGQa6PA71nvH5nkDexhLteJqkM4dQmWF9g/84'/1'/0'/0/*)";
    let change = "wpkh([3c31d632/84'/1'/0']tpubDCYwFkks2cg78N7eoYbBatsFEGje8vW8arSKW4rLwD1AU1s9KJMDRHE32JkvYERuiFjArrsH7qpWSpJATed5ShZbG9KsskA5Rmi6NSYgYN2/0/*)";

    let err = Wallet::new_no_persist(desc, change, Network::Testnet);
    assert!(matches!(
        err,
        Err(DescriptorError::ExternalAndInternalAreTheSame),
    ));
}
```

crates/bdk/examples/compiler.rs
ValuedMammal: I do like this suggestion. I got in the habit of using the old style since I don't think the old rust 1.57 supported it. But also it's common to want to use an expression like a function call as the argument to the format string, in which case it would have to go after the double quotes.

ValuedMammal:
Rebased on 358e842dcda0eb867d229881823c76dba1d7cce1

storopoli (approved):
ACK cddefbbcff8728ff10a125aada34f39ae1c65d4a

ValuedMammal:
Rebased on 7607b492 and added two commits, but this may need more discussion and testing to decide if it's a reasonable solution

storopoli (approved):
Totally agree with the infallible.

ACK aec4ddad1daa241b64dd3787ed790135878d22ea

Note that you might want to update the original PR message details.

notmandatory:
> Note that the public descriptor will be common whether the wallet is given a public or private descriptor. So for example this should pass:
> 
> ```rust
> #[test]
> fn same_descriptor_public_private() {
>     // public + private of same descriptor should fail to create wallet
>     let desc = "wpkh(tprv8ZgxMBicQKsPdcAqYBpzAFwU5yxBUo88ggoBqu1qPcHUfSbKK1sKMLmC7EAk438btHQrSdu3jGGQa6PA71nvH5nkDexhLteJqkM4dQmWF9g/84'/1'/0'/0/*)";
>     let change = "wpkh([3c31d632/84'/1'/0']tpubDCYwFkks2cg78N7eoYbBatsFEGje8vW8arSKW4rLwD1AU1s9KJMDRHE32JkvYERuiFjArrsH7qpWSpJATed5ShZbG9KsskA5Rmi6NSYgYN2/0/*)";
> 
>     let err = Wallet::new_no_persist(desc, change, Network::Testnet);
>     assert!(matches!(
>         err,
>         Err(DescriptorError::ExternalAndInternalAreTheSame),
>     ));
> }
> ```

This test does pass and I think is an adequate level of error checking. If we run into cases where users are adding duplicates descriptors that aren't caught we can add more sophisticated checking later and it won't break the API.

It's good documentation to add this test to `tests/wallet.rs` and I think you need a rebase and then this one looks ready to merge.

notmandatory (approved):
ACK aec4ddad1daa241b64dd3787ed790135878d22ea

storopoli (approved):
ACK aec4ddad1daa241b64dd3787ed790135878d22ea

crates/wallet/src/wallet/mod.rs
```diff
@@ -517,14 +514,15 @@ impl Wallet {
             .indexer
             .keychains_added
             .get(&KeychainKind::External)
-            .ok_or(LoadError::MissingDescriptor)?
+            .expect("keychain must have been added")
```
ValuedMammal: Is it safe to `expect` the descriptor will always be recovered, or is it better to keep the `MissingDescriptor` error? since technically we were handling the error even for the external descriptor which was always present

notmandatory: This is a tricky question. My original thought is that our code will always persist the right data so missing the network or either of the descriptors should be a panic. 

But on the other hand we're always loading changeset data from outside storage. And that means we could load the wrong data or possibly something outside our code could have changed the data. 

Based on that idea that at runtime we don't completely control the data we're loading, for both the internal and external keychains I'd like to throw the `LoadError::MissingDescriptor` instead of panic.

ValuedMammal: Ok, will fix. Do you think the `MissingDescriptor` variant should hold the keychain kind so it can pass it along to `NewOrLoadError::LoadedDescriptorDoesNotMatch` ?

notmandatory: Oh ya, I agree that it would be a more useful error if you include the missing `KeychainKind`.

ValuedMammal:
@storopoli I took your advice about more precise wording from this comment https://github.com/bitcoindevkit/bdk/pull/1390#discussion_r1541215579 and I even put the variable inside the format string in 0ffcd9a1cf376ef01b1c32f006aafa944c7fda2d :)

storopoli (approved):
ACK 0ffcd9a1cf376ef01b1c32f006aafa944c7fda2d

notmandatory (approved):
re-ACK 8bc3d35f6c83d078551c4bc57f72f13324865486

PR #1392, opened 2024-03-27
bdk_hwi, set cargo bdk version to 1.0.0-alpha.8

notmandatory:
### Description

Need to set a specific version for the `bdk_wallet` dependency in order to be able to publish `bdk_hwi` to crates.io. I also added a README file since it is also a warning if a crate is published without one.

```shell
cargo publish -p bdk_hwi
    Updating crates.io index
error: all dependencies must have a version specified when publishing.
dependency `bdk` does not specify a version
Note: The published dependency will use the version from crates.io,
the `path` specification will be removed from the dependency declaration.
```

#### All Submissions:

### DIFF
--- a/crates/hwi/Cargo.toml
+++ b/crates/hwi/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bdk_hwi"
-version = "0.2.0"
+version = "0.1.0"
 edition = "2021"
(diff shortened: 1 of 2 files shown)


notmandatory:
Will make these fixes in #1449.

PR #1395, opened 2024-04-03
feat(wallet): Allow user provided RNG, make rand an optional dependency

### LINKED ISSUE #871: Remove dependency on rand
Because we invoke `thread_rng` directly we have to also depend on `getrandom` in order to set certain feature flags to get things to compile on wasm.

Can we just depend on `rand_core` and let users pass in anything implement `RngCore`? This would simplify dependencies a bit. The place where this will be most tricky is in tx building but I hope it can be done.

We should try and attempt this before `v1`.

rustaceanrob:
### Description

WIP towards removing `rand` fixes #871 

The `rand` dependency was imported explicitly, but `rand` is also implicitly used through the `rand-std` feature flag on `bitcoin`.

### Notes to he reviewers

**Updated:**

`rand` was used primarily in two parts of `bdk`. Particularly in signing and in building a transaction.

Signing:
- Used implicitly in [`sign_schnorr`](https://docs.rs/bitcoin/latest/bitcoin/key/struct.Secp256k1.html#method.sign_schnorr), but nowhere else within `signer`. 

Transaction ordering:
- Used to shuffle the inputs and outputs of a transaction, the default
- Used in the single random draw __as a fallback__ to branch and bound during coin selection. Branch and bound is the default coin selection option.

See conversation for proposed solutions.

### Changelog notice

- Remove the `rand` dependency from `bdk`

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/coin_selection.rs
+++ b/crates/wallet/src/wallet/coin_selection.rs
@@ -31,8 +31,6 @@
 //! # use bdk_wallet::*;
 //! # use bdk_wallet::wallet::coin_selection::decide_change;
 //! # use anyhow::Error;
-//! # use rand::{thread_rng, RngCore};
-//!
 //! #[derive(Debug)]
 //! struct AlwaysSpendEverything;
 //!
@@ -94,7 +92,7 @@
 //! let psbt = {
 //!     let mut builder = wallet.build_tx().coin_selection(AlwaysSpendEverything);
 //!     builder.add_recipient(to_address.script_pubkey(), Amount::from_sat(50_000));
-//!     builder.finish_with_aux_rand(&mut thread_rng())?
+//!     builder.finish()?
 //! };
 //!
 //! // inspect, sign, broadcast, ...
--- a/crates/wallet/src/wallet/coin_selection.rs
+++ b/crates/wallet/src/wallet/coin_selection.rs
@@ -31,6 +31,8 @@
 //! # use bdk_wallet::*;
 //! # use bdk_wallet::wallet::coin_selection::decide_change;
 //! # use anyhow::Error;
+//! # use rand::{thread_rng, RngCore};
+//!
 //! #[derive(Debug)]
 //! struct AlwaysSpendEverything;
 //!
@@ -92,7 +94,7 @@
 //! let psbt = {
 //!     let mut builder = wallet.build_tx().coin_selection(AlwaysSpendEverything);
 //!     builder.add_recipient(to_address.script_pubkey(), Amount::from_sat(50_000));
-//!     builder.finish()?
+//!     builder.finish_with_aux_rand(&mut thread_rng())?
 //! };
 //!
 //! // inspect, sign, broadcast, ...
@@ -114,8 +116,9 @@ use bitcoin::{Script, Weight};
 
 use core::convert::TryInto;
 use core::fmt::{self, Formatter};
-use rand::seq::SliceRandom;
+use rand_core::RngCore;
 
+use super::utils::shuffle_slice;
 /// Default coin selection algorithm used by [`TxBuilder`](super::tx_builder::TxBuilder) if not
 /// overridden
 pub type DefaultCoinSelectionAlgorithm = BranchAndBoundCoinSelection;
@@ -516,27 +519,16 @@ impl CoinSelectionAlgorithm for BranchAndBoundCoinSelection {
             ));
         }
 
-        Ok(self
-            .bnb(
-                required_utxos.clone(),
-                optional_utxos.clone(),
-                curr_value,
-                curr_available_value,
-                target_amount,
-                cost_of_change,
-                drain_script,
-                fee_rate,
-            )
-            .unwrap_or_else(|_| {
-                self.single_random_draw(
-                    required_utxos,
-                    optional_utxos,
-                    curr_value,
-                    target_amount,
-                    drain_script,
-                    fee_rate,
-                )
-            }))
+        self.bnb(
+            required_utxos.clone(),
+            optional_utxos.clone(),
+            curr_value,
+            curr_available_value,
+            target_amount,
+            cost_of_change,
+            drain_script,
+            fee_rate,
+        )
     }
 }
 
@@ -663,40 +655,6 @@ impl BranchAndBoundCoinSelection {
         ))
     }
 
-    #[allow(clippy::too_many_arguments)]
-    fn single_random_draw(
-        &self,
-        required_utxos: Vec<OutputGroup>,
-        mut optional_utxos: Vec<OutputGroup>,
-        curr_value: i64,
-        target_amount: i64,
-        drain_script: &Script,
-        fee_rate: FeeRate,
-    ) -> CoinSelectionResult {
-        optional_utxos.shuffle(&mut rand::thread_rng());
-        let selected_utxos = optional_utxos.into_iter().fold(
-            (curr_value, vec![]),
-            |(mut amount, mut utxos), utxo| {
-                if amount >= target_amount {
-                    (amount, utxos)
-                } else {
-                    amount += utxo.effective_value;
-                    utxos.push(utxo);
-                    (amount, utxos)
-                }
-            },
-        );
-
-        // remaining_amount can't be negative as that would mean the
-        // selection wasn't successful
-        // target_amount = amount_needed + (fee_amount - vin_fees)
-        let remaining_amount = (selected_utxos.0 - target_amount) as u64;
-
-        let excess = decide_change(remaining_amount, fee_rate, drain_script);
-
-        BranchAndBoundCoinSelection::calculate_cs_result(selected_utxos.1, required_utxos, excess)
-    }
-
     fn calculate_cs_result(
         mut selected_utxos: Vec<OutputGroup>,
         mut required_utxos: Vec<OutputGroup>,
@@ -715,6 +673,58 @@ impl BranchAndBoundCoinSelection {
             excess,
         }
     }
+}
+
+// Pull UTXOs at random until we have enough to meet the target
+pub(crate) fn single_random_draw(
+    required_utxos: Vec<WeightedUtxo>,
+    optional_utxos: Vec<WeightedUtxo>,
+    target_amount: u64,
+    drain_script: &Script,
+    fee_rate: FeeRate,
+    rng: &mut impl RngCore,
+) -> CoinSelectionResult {
+    let target_amount = target_amount
+        .try_into()
+        .expect("Bitcoin amount to fit into i64");
+
+    let required_utxos: Vec<OutputGroup> = required_utxos
+        .into_iter()
+        .map(|u| OutputGroup::new(u, fee_rate))
+        .collect();
+
+    let mut optional_utxos: Vec<OutputGroup> = optional_utxos
+        .into_iter()
+        .map(|u| OutputGroup::new(u, fee_rate))
+        .collect();
+
+    let curr_value = required_utxos
+        .iter()
+        .fold(0, |acc, x| acc + x.effective_value);
+
+    shuffle_slice(&mut optional_utxos, rng);
+
+    let selected_utxos =
+        optional_utxos
+            .into_iter()
+            .fold((curr_value, vec![]), |(mut amount, mut utxos), utxo| {
+                if amount >= target_amount {
+                    (amount, utxos)
+                } else {
+                    amount += utxo.effective_value;
+                    utxos.push(utxo);
+                    (amount, utxos)
+                }
+            });
+
+    // remaining_amount can't be negative as that would mean the
+    // selection wasn't successful
+    // target_amount = amount_needed + (fee_amount - vin_fees)
+    let remaining_amount = (selected_utxos.0 - target_amount) as u64;
+
+    let excess = decide_change(remaining_amount, fee_rate, drain_script);
+
+    BranchAndBoundCoinSelection::calculate_cs_result(selected_utxos.1, required_utxos, excess)
 }
 
 /// Remove duplicate UTXOs.
@@ -740,6 +750,7 @@ where
 mod test {
     use assert_matches::assert_matches;
     use core::str::FromStr;
+    use rand::rngs::StdRng;
 
     use bdk_chain::ConfirmationTime;
     use bitcoin::{Amount, ScriptBuf, TxIn, TxOut};
@@ -748,8 +759,7 @@ mod test {
     use crate::types::*;
     use crate::wallet::coin_selection::filter_duplicates;
 
-    use rand::rngs::StdRng;
-    use rand::seq::SliceRandom;
+    use rand::prelude::SliceRandom;
     use rand::{Rng, RngCore, SeedableRng};
 
     // signature len (1WU) + signature and sighash (72WU)
@@ -1090,13 +1100,12 @@ mod test {
     }
 
     #[test]
+    #[ignore = "SRD fn was moved out of BnB"]
     fn test_bnb_coin_selection_success() {
         // In this case bnb won't find a suitable match and single random draw will
         // select three outputs
         let utxos = generate_same_value_utxos(100_000, 20);
-
         let drain_script = ScriptBuf::default();
-
         let target_amount = 250_000 + FEE_AMOUNT;
 
         let result = BranchAndBoundCoinSelection::default()
@@ -1136,6 +1145,7 @@ mod test {
     }
 
     #[test]
+    #[ignore = "no exact match for bnb, previously fell back to SRD"]
     fn test_bnb_coin_selection_optional_are_enough() {
         let utxos = get_test_utxos();
         let drain_script = ScriptBuf::default();
@@ -1154,6 +1164,26 @@ mod test {
         assert_eq!(result.selected.len(), 2);
         assert_eq!(result.selected_amount(), 300000);
         assert_eq!(result.fee_amount, 136);
+    }
+
+    #[test]
+    fn test_single_random_draw_function_success() {
+        let seed = [0; 32];
+        let mut rng: StdRng = SeedableRng::from_seed(seed);
+        let mut utxos = generate_random_utxos(&mut rng, 300);
+        let target_amount = sum_random_utxos(&mut rng, &mut utxos) + FEE_AMOUNT;
+        let fee_rate = FeeRate::from_sat_per_vb_unchecked(1);
+        let drain_script = ScriptBuf::default();
+        let result = single_random_draw(
+            vec![],
+            utxos,
+            target_amount,
+            &drain_script,
+            fee_rate,
+            &mut rng,
+        );
+        assert!(result.selected_amount() > target_amount);
+        assert_eq!(result.fee_amount, (result.selected.len() * 68) as u64);
     }
 
     #[test]
@@ -1408,34 +1438,6 @@ mod test {
                 .unwrap();
             assert_eq!(result.selected_amount(), target_amount as u64);
         }
-    }
-
-    #[test]
-    fn test_single_random_draw_function_success() {
-        let seed = [0; 32];
-        let mut rng: StdRng = SeedableRng::from_seed(seed);
-        let mut utxos = generate_random_utxos(&mut rng, 300);
-        let target_amount = sum_random_utxos(&mut rng, &mut utxos) + FEE_AMOUNT;
-
-        let fee_rate = FeeRate::from_sat_per_vb_unchecked(1);
-        let utxos: Vec<OutputGroup> = utxos
-            .into_iter()
-            .map(|u| OutputGroup::new(u, fee_rate))
-            .collect();
-
-        let drain_script = ScriptBuf::default();
-
-        let result = BranchAndBoundCoinSelection::default().single_random_draw(
-            vec![],
-            utxos,
-            0,
-            target_amount as i64,
-            &drain_script,
-            fee_rate,
-        );
-
-        assert!(result.selected_amount() > target_amount);
-        assert_eq!(result.fee_amount, (result.selected.len() * 68) as u64);
     }
 
     #[test]
--- a/crates/wallet/src/wallet/tx_builder.rs
+++ b/crates/wallet/src/wallet/tx_builder.rs
@@ -20,7 +20,6 @@
 //! # use bdk_wallet::wallet::ChangeSet;
 //! # use bdk_wallet::wallet::error::CreateTxError;
 //! # use anyhow::Error;
-//! # use rand::thread_rng;
 //! # let to_address = Address::from_str("2N4eQYCbKUHCCTUjBJeHcJp9ok6J2GZsTDt").unwrap().assume_checked();
 //! # let mut wallet = doctest_wallet!();
 //! // create a TxBuilder from a wallet
@@ -35,7 +34,7 @@
 //!     .do_not_spend_change()
 //!     // Turn on RBF signaling
 //!     .enable_rbf();
-//! let psbt = tx_builder.finish_with_aux_rand(&mut thread_rng())?;
+//! let psbt = tx_builder.finish()?;
 //! # Ok::<(), anyhow::Error>(())
 //! ```
 
@@ -57,7 +56,7 @@ use crate::{KeychainKind, LocalOutput, Utxo, WeightedUtxo};
 /// A transaction builder
 ///
 /// A `TxBuilder` is created by calling [`build_tx`] or [`build_fee_bump`] on a wallet. After
-/// assigning it, you set options on it until finally calling [`finish_with_aux_rand`] to consume the builder and
+/// assigning it, you set options on it until finally calling [`finish`] to consume the builder and
 /// generate the transaction.
 ///
 /// Each option setting method on `TxBuilder` takes and returns `&mut self` so you can chain calls
@@ -71,7 +70,6 @@ use crate::{KeychainKind, LocalOutput, Utxo, WeightedUtxo};
 /// # use bdk_wallet::wallet::ChangeSet;
 /// # use bdk_wallet::wallet::error::CreateTxError;
 /// # use anyhow::Error;
-/// # use rand::thread_rng;
 /// # let mut wallet = doctest_wallet!();
 /// # let addr1 = Address::from_str("2N4eQYCbKUHCCTUjBJeHcJp9ok6J2GZsTDt").unwrap().assume_checked();
 /// # let addr2 = addr1.clone();
@@ -82,7 +80,7 @@ use crate::{KeychainKind, LocalOutput, Utxo, WeightedUtxo};
 ///         .ordering(TxOrdering::Untouched)
 ///         .add_recipient(addr1.script_pubkey(), Amount::from_sat(50_000))
 ///         .add_recipient(addr2.script_pubkey(), Amount::from_sat(50_000));
-///     builder.finish_with_aux_rand(&mut thread_rng())?
+///     builder.finish()?
 /// };
 ///
 /// // non-chaining
@@ -92,7 +90,7 @@ use crate::{KeychainKind, LocalOutput, Utxo, WeightedUtxo};
 ///     for addr in &[addr1, addr2] {
 ///         builder.add_recipient(addr.script_pubkey(), Amount::from_sat(50_000));
 ///     }
-///     builder.finish_with_aux_rand(&mut thread_rng())?
+///     builder.finish()?
 /// };
 ///
 /// assert_eq!(psbt1.unsigned_tx.output[..2], psbt2.unsigned_tx.output[..2]);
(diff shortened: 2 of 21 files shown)


LLFourn:
> ### Description
> 
> WIP towards removing `rand` #871
> 
> The `rand` dependency was imported explicitly, but `rand` is also implicitly used through the `rand-std` feature flag on `bitcoin`. In this PR, I used the `bitcoin::secp256k1::rand` re-export from `bitcoin` to remove the explicit import of `rand`.
> 
> Per my understanding, the `rand-std` feature flag is used for signing transactions, and was being used outside of `tx_builder` and `coin_selection`. I wanted to open up some discussion on if there should be additional effort to remove `rand-std` as a feature flag from `bitcoin`, since `rand` is brought in there anyway.

Good question. I think we want to remove the dependency on the bitcoin feature too because otherwise we have the same problem with wasm. If we want randomness during signing (it's a good default), then we should make the caller pass in an Rng and call: [`sign_schnorr_wth_aux_rand`](https://docs.rs/bitcoin/latest/bitcoin/key/struct.Secp256k1.html#method.sign_schnorr_with_aux_rand) rather than the one that needs the rand feature. We could also make randomness here optional (deterministic by default) by adding a field to `SignOptions` to pass in an `Option<&mut R> where R: RngCore`

> ### Notes to the reviewers
> 
> While experimenting with this, allowing the user to pass in an `impl RngCore` to finish a transaction resulted in redundant `impl RngCore` arguments for algorithms that implement the coin selection trait. I just want to confirm `tx.finish(&mut rng)` is the type of API change that is worth removing the `rand-std` feature flag from `bitcoin` before proceeding.

Yeah, I think it's worth it tbh if that's what it comes to. Annoying thing is that not every `TxBuilder` ordering needs randomness. Shuffling could be done outside of building in theory but that would require an extra step by the user to engage it.

rustaceanrob:
Thanks, will keep chipping at this

crates/bdk/Cargo.toml
```diff
@@ -14,7 +14,7 @@ rust-version = "1.63"
 
 [dependencies]
 miniscript = { version = "10.0.0", features = ["serde"], default-features = false }
-bitcoin = { version = "0.30.0", features = ["serde", "base64", "rand-std"], default-features = false }
+bitcoin = { version = "0.30.0", features = ["serde", "base64"], default-features = false }
```
LLFourn: we should also be removing the `target = "wasm32"` dependencies (or making them dev dependencies to get tests to pass).

rustaceanrob:
> If we want randomness during signing (it's a good default), then we should make the caller pass in an Rng and call: [sign_schnorr_wth_aux_rand](https://docs.rs/bitcoin/latest/bitcoin/key/struct.Secp256k1.html#method.sign_schnorr_with_aux_rand) rather than the one that needs the rand feature. We could also make randomness here optional (deterministic by default) by adding a field to SignOptions to pass in an Option<&mut R> where R: RngCore

After thinking about this I am a little confused at how the `sign_schnorr` family of functions works. [sign_schnorr_wth_aux_rand](https://docs.rs/bitcoin/latest/bitcoin/key/struct.Secp256k1.html#method.sign_schnorr_with_aux_rand) makes sense and is easy enough to implement with `SignOptions`. [sign_schnorr_no_aux_rand](https://docs.rs/bitcoin/latest/bitcoin/key/struct.Secp256k1.html#method.sign_schnorr_no_aux_rand) is tripping me up. How is it possible to sign with no auxiliary random entropy? Is there something going here with a deterministic nonce based on the secret and the message? In the case where the user signs with `SignOptions::default()`, which I imagine most of the time, what is the implication for repeated nonces? 

Another aside, to shuffle inputs and outputs, and I think in BnB, an extension trait on slices from `rand` was used to call `shuffle` on `Vec<T>`. A way for me to add this functionality back was to use a crate called [shuffle](https://docs.rs/shuffle/latest/shuffle/index.html), which allows slice shuffling with any `impl RngCore`. So far I am not a fan of it, as it adds new possibility of errors and another dependency. I will have to home brew a [shuffle algorithm](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle) to avoid using the crate `shuffle`. Might take some time and a more careful review.

LLFourn:
> > If we want randomness during signing (it's a good default), then we should make the caller pass in an Rng and call: [sign_schnorr_wth_aux_rand](https://docs.rs/bitcoin/latest/bitcoin/key/struct.Secp256k1.html#method.sign_schnorr_with_aux_rand) rather than the one that needs the rand feature. We could also make randomness here optional (deterministic by default) by adding a field to SignOptions to pass in an Option<&mut R> where R: RngCore
> 
> After thinking about this I am a little confused at how the `sign_schnorr` family of functions works. [sign_schnorr_wth_aux_rand](https://docs.rs/bitcoin/latest/bitcoin/key/struct.Secp256k1.html#method.sign_schnorr_with_aux_rand) makes sense and is easy enough to implement with `SignOptions`. [sign_schnorr_no_aux_rand](https://docs.rs/bitcoin/latest/bitcoin/key/struct.Secp256k1.html#method.sign_schnorr_no_aux_rand) is tripping me up. How is it possible to sign with no auxiliary random entropy? Is there something going here with a deterministic nonce based on the secret and the message? 

Yes. See [BIP340](https://bips.xyz/340#user-content-default-signing). The secret and the message are hashed together if the aux rand bytes are null so that' swhat `no_aux_rand` is doing.

> In the case where the user signs with `SignOptions::default()`, which I imagine most of the time, what is the implication for repeated nonces?
> 

There are no implications in most scenarios. People making a Bitcoin HSM using BDK that needs to be robust against an attacker who can inject faults into the power supply or something should look for the option and add randomness. It'd be nice if we could easily make it the default or nudge people towards. We'll have an opportunity when we redesign transaction building. But for now making it optional is fine. 

> Another aside, to shuffle inputs and outputs, and I think in BnB, an extension trait on slices from `rand` was used to call `shuffle` on `Vec<T>`. A way for me to add this functionality back was to use a crate called [shuffle](https://docs.rs/shuffle/latest/shuffle/index.html), which allows slice shuffling with any `impl RngCore`. So far I am not a fan of it, as it adds new possibility of errors and another dependency. I will have to home brew a [shuffle algorithm](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle) to avoid using the crate `shuffle`. Might take some time and a more careful review.

Definitely shouldn't add a dependency. I think rolling your own is the right thing to do in this case. I wonder why we would be shuffling things in BnB but hopefully that goes away soon.

rustaceanrob:
> Definitely shouldn't add a dependency. I think rolling your own is the right thing to do in this case. I wonder why we would be shuffling things in BnB but hopefully that goes away soon.

I vastly overestimated the difficulty. No issues on the shuffling algo. 

> There are no implications in most scenarios. People making a Bitcoin HSM using BDK that needs to be robust against an attacker who can inject faults into the power supply or something should look for the option and add randomness. It'd be nice if we could easily make it the default or nudge people towards. We'll have an opportunity when we redesign transaction building. But for now making it optional is fine.

I am thinking of a few ways to go about this. There can be an `Option<[u8; 32]>` called `aux_rand` on `SignOptions`, but I assume this will be frequently `None`. Another approach, easiest for the bindings, would be to make an explicit argument on `sign` named `aux_rand` that takes an `Option<[u8; 32]>` that forces the user to opt out of adding randomness by at least acknowledging they are opting out. Likewise, an `Option<Box<dyn RngCore>>` or even a forced `&mut impl RngCore` could be passed to `sign`. I like this approach, but bindings would have to alter this API I would imagine.

rustaceanrob:
_Updated_: The PR now uses `Option<[u8; 32]>` as a `SignOptions` for additional entropy on Schnorr signatures. While I think this is the lowest hanging fruit so-to-speak, I think it would be somewhat hard to find for users. Adding a generic over `SignOptions` propagates all the way up to `Wallet`, adding a generic. Removing generics from the wallet is mentioned as an issue (#1363), so I don't think this is much of an option. 

`SignOptions` could contain a `Option<Box<dyn RngCore>>`, if the entropy should remain as an option for users. 

I like the idea of having a user pass an `Option<R>` where `R: RngCore` when calling `sign` on wallet. They would have to explicitly opt out of using additional randomness, whereas they can passively ignore this feature if it is a member of `SignOptions`.

On building transactions, using randomness depends on the `TxOrdering` and if the user selected the single random draw coin selection algorithm. I added a mandatory `&mut R` to `finish`, but I would like to hear other suggestions here!

cc @ValuedMammal

LLFourn:
> _Updated_: The PR now uses `Option<[u8; 32]>` as a `SignOptions` for additional entropy on Schnorr signatures. While I think this is the lowest hanging fruit so-to-speak, I think it would be somewhat hard to find for users. Adding a generic over `SignOptions` propagates all the way up to `Wallet`, adding a generic. Removing generics from the wallet is mentioned as an issue (#1363), so I don't think this is much of an option.

`Option<[u8;32]>` being passed down to the schnorr signing function is not really using it in the right way. Each call is meant to have fresh randomness. I think you can add a generic to `SignOptions` without it going on `Wallet`. You're idea of doing a `Box<dyn RngCore>` is also fine.

TBH I'm thinking we should just drop randomness for the schnorr signing function for now (use the `no_aux_rand` function). I don't imagine BDK's PSBT signing being used in a context where synthetic nonce's theoretical benefits will translate into real world security. It is desirable to have the ability but I imagine all this stuff being re-imagined in the future.

rustaceanrob:
> TBH I'm thinking we should just drop randomness for the schnorr signing function for now (use the `no_aux_rand` function). 

I would prefer this. Ultimately I think this is a feature that can be handled outside of this PR.

I thought longer about input/output shuffling and greater concept of `TxOrdering`. The `TxOrdering` the user selects should determine whether or not there should be randomness involved. To reflect this I decided to add `Box<dyn RngCore>` on the `Shuffle` enumeration. An `Option<Box<dyn RngCore>>` on the `TxParams` does not properly couple the `TxOrdering` and the source for randomness, and would lead to confusing error cases when the user does not provide `RngCore`. Due to this change, `TxOrdering` is unable to keep `Shuffle` as the `Default` enumeration. Rather than use `Untouched` as a default, I do not think this enum should derive `Default`. 

From the [docs](https://doc.rust-lang.org/std/default/trait.Default.html): 
> Sometimes, you want to fall back to some kind of default value, and don’t particularly care what it is.

I would argue we do care what the ordering is for BDK user transactions as there are privacy implications, and the user should hand select the `TxOrdering`. I demonstrated in all example crates how to use `Shuffle`. 

Along the same vein, I took the opportunity to try to resolve #534 by deprecating `BIP69` lexicographical sorting. As discussed in that thread, `BIP69` does not preserve privacy as intended. Likewise, I added a `Custom` enumeration for the user to provide arbitrary sorting functions for inputs and outputs.

I referenced the `coin_selection` crate for how the coin selection process might look in the future. It looks like single random draw would not be implemented, as it would require a dependency. I removed single random draw for now, but there should be a deterministic fallback (i.e. largest first), as there are high failure rates when trying to create 0 change transactions. Either that, or more consideration into adding `rand_core` in future coin selection.

The `TxOrdering` enum for reference: 

```rust
pub enum TxOrdering {
    /// Randomized
    Shuffle(Box<dyn RngCore>),
    /// Unchanged
    Untouched,
    /// BIP69 / Lexicographic
    #[deprecated = "BIP69 does not improve privacy as was the intention of the BIP"]
    Bip69Lexicographic,
    /// A custom ordering of inputs and outputs
    Custom {
        /// The custom function to order the inputs of the transaction
        input_ordering: Box<dyn Fn(&TxIn, &TxIn) -> core::cmp::Ordering>,
        /// The custom function to order the outputs of the transaction
        output_ordering: Box<dyn Fn(&TxOut, &TxOut) -> core::cmp::Ordering>,
    },
}
```

ValuedMammal:
Here's how I think we can keep single random draw as a fallback to BNB: don't implement SRD on the type `BranchAndBoundCoinSelection` directly, but instead pull it out of the impl block and make it `pub(crate)`. Then call it directly in `Wallet::create_tx` in the case that BNB fails, passing roughly the same arguments. Add a parameter to `create_tx` called `mut rng: Option<Box<dyn RngCore>>` that can be used both during coin selection as well as tx ordering. Also add a parameter for the rng to the function `TxOrdering::sort_tx`. This lets us avoid dealing with the `Default` behavior of `TxParams`.

I was a bit skeptical of adding an argument to `TxBuilder::finish` because that seems like a poor design, but actually this is probably the right move considering the bulk of the logic occurs after `finish` is called. As an aside: given the chance, I would probably replace the name `build_tx` with something like `Wallet::new_tx` or `TxBuilder::new` and replace the name `finish` with `build_tx` to reflect this point. In any case, `finish` can then accept an argument `rng: Box<dyn RngCore>`, similar to what you propose, only generalized to work whenever randomness is needed. Then we add a second function `finish_no_aux_rand` that takes no additional arguments and passes `None` as the optional rng to `create_tx`. I'm still open to debate if you have another opinion.

crates/bdk/src/wallet/utils.rs
```diff
@@ -110,6 +112,16 @@ impl<Pk: MiniscriptKey + ToPublicKey> Satisfier<Pk> for Older {
     }
 }
 
+// The Fisher-Yates shuffling algorithm to randomly order a list of N items: [https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle]
+pub(crate) fn shuffle_slice<T>(list: &mut [T], rng: &mut Box<dyn RngCore>) {
+    let mut current_index = list.len() - 1;
```
ValuedMammal: Can we be sure this won't be called with an empty `list` ?

rustaceanrob: No! I will add that base case

rustaceanrob:
>  In any case, finish can then accept an argument rng: Box<dyn RngCore>, similar to what you propose, only generalized to work whenever randomness is needed. Then we add a second function finish_no_aux_rand that takes no additional arguments and passes None as the optional rng to create_tx. I'm still open to debate if you have another opinion.

I am hesitant about `finish_no_aux_rand` because let's say a user has some `TxOrdering::Custom` and defaults to BnB. They try `finish_no_aux_rand` and BnB fails and throws an error because there was no RNG source. That leaves the user to investigate why BnB should have randomness at all. Why I proposed having the `RngCore` associated with the enum variant on `TxOrdering` was so the user clearly understands why randomness is occurring. To your point, it would be strange if not nonsensical to have arguments `TxOrdering::Shuffle(Box<dyn RngCore>` followed by some other `Box<dyn RngCore>` for the SRD fallback. So we are left with a general `Box<dyn RngCore>`. If we go this generalized approach, I would rather the user only have `finish` and be forced to provide the RNG so there aren't any new error cases with this PR. 

> Here's how I think we can keep single random draw as a fallback to BNB

Agree with this. I looked at the `coin_selection` crate though, and there was a zero dependency policy. In an effort of being future oriented, I'd like to confirm `core_rand` can be made an exception and SRD will be implemented there. If not, we should find a zero dependency (deterministic-selection) solution.

cc @evanlinjin

ValuedMammal:
> They try `finish_no_aux_rand` and BnB fails and throws an error because there was no RNG source.

In that case, fallback to LargestFirst. maybe simpler to just require the rng in all cases, though.

rustaceanrob:
I exchanged a message with Murch on that. He said that would be a bad idea because they might end up grinding down their UTXO pool. `LargestFirst` isn't something we want to have happening in the background. I was going to go with that but got convinced otherwise

ValuedMammal:
Seeing that it's common for our upstream deps to make rand-std a cargo feature, the idea came up that BDK should do the same, but that still leaves the question of how to obtain randomness when the feature is not enabled. I think we agree that a discussion of the optimal coin-select strategy probably belongs in a separate PR. If we were to do nothing else but try and drop the dependency on `rand`, I think it would look similar to what you have in this PR - while keeping SRD in the picture, since currently this is the only thing BDK knows to fallback on (even if `bdk_coin_select` doesn't include SRD, should that necessarily preclude the wallet from implementing it?). But if the end result is deemed awkward or unwieldy, I don't rule out possibly taking a step back to reevaluate. This is sort of the vision I have in mind:

Add the parameter `rng: &mut impl RngCore` in these places
- `generate` for GeneratableKey
- `TxBuilder::finish`
- `Wallet::create_tx`
- `single_random_draw` (now a standalone function)
- `TxOrdering::sort_tx`
- the new `shuffle_slice`

With a minimal approach, we simply require a user-provided rng everywhere. I don't think this interferes with the idea of custom tx ordering, nor does it create strange new error cases.

> new snippet of calling `coin_select` from inside the function `create_tx`

```rust
        // Try BNB. If that fails, fallback to single random draw
        let coin_selection = match coin_selection.coin_select(
            required_utxos.clone(),
            optional_utxos.clone(),
            fee_rate,
            outgoing + fee_amount,
            &drain_script,
        ) {
            Ok(res) => res,
            Err(e) => match e {
                coin_selection::Error::InsufficientFunds { .. } => {
                    return Err(CreateTxError::CoinSelection(e));
                }
                coin_selection::Error::BnBNoExactMatch
                | coin_selection::Error::BnBTotalTriesExceeded => {
                    coin_selection::single_random_draw(
                        required_utxos,
                        optional_utxos,
                        fee_rate,
                        outgoing + fee_amount,
                        &drain_script,
                        rng,
                    )
                }
            },
        };
```

LLFourn:
The plan above sounds good. As an aside I'm going to work on replacing the coin selection code with `bdk_coin_select`. We will probably fall back to "largest first" if BnB fails -- but in this case BnB is very unlikely to fail because it is allowed to find solutions with change. Just noting this because the problem of randomness during coin selection will go away (hopefully!).

rustaceanrob:
> This is sort of the vision I have in mind:

Will prep this for tomorrows PR review club. There is still plenty to discuss but I think we are in agreement on how this should be implemented. Thanks for the input @ValuedMammal 

> but in this case BnB is very unlikely to fail because it is allowed to find solutions with change

I'm interested to see how this works. Admittedly coin selection is an area I need to study up on, but relaxing BnB to find solutions with change makes randomness a non-issue as you said.

storopoli:
> Will prep this for tomorrows PR review club.

It's on Thursday no?

@rustaceanrob

rustaceanrob:
Oops mixed it up with team meeting. Regardless of me rebasing and making these changes, we should have a discussion on feature flags for the review club

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -839,15 +844,14 @@ pub enum TxOrdering {
...
             TxOrdering::Untouched => {}
             TxOrdering::Shuffle => {
-                use rand::seq::SliceRandom;
-                let mut rng = rand::thread_rng();
-                tx.input.shuffle(&mut rng);
-                tx.output.shuffle(&mut rng);
+                shuffle_slice(&mut tx.input, rng);
+                shuffle_slice(&mut tx.output, rng);
             }
+            #[allow(deprecated)]
```
ValuedMammal: 📌 While the "swiss army knife" analogy in https://github.com/bitcoindevkit/bdk/issues/534#issuecomment-1022975295 is compelling, I'm also in favor of just removing the `Bip69Lexicographic` variant now.

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -1411,34 +1421,6 @@ mod test {
         }
     }
 
-    #[test]
-    fn test_single_random_draw_function_success() {
```
ValuedMammal: Maybe worth keeping a version of this test to be sure `single_random_draw` still works as intended.

crates/bdk/src/wallet/utils.rs
```diff
@@ -110,6 +111,19 @@ impl<Pk: MiniscriptKey + ToPublicKey> Satisfier<Pk> for Older {
     }
 }
 
+// The Fisher-Yates shuffling algorithm to randomly order a list of N items: [https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle]
```
ValuedMammal: totally unimportant suggestion:

```suggestion
// The Knuth shuffling algorithm based on the original Fisher-Yates method: [https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle]
```

Also, wouldn't hurt to have a unit test for `shuffle_slice`.

ValuedMammal (commented):
Approach ACK

crates/bdk/src/keys/mod.rs
```diff
@@ -20,6 +20,8 @@ use core::marker::PhantomData;
 use core::ops::Deref;
 use core::str::FromStr;
 
+use rand_core::RngCore;
```
storopoli: We might as well restrict to crytographic-secure PRNG with [`CryptoRng`](https://docs.rs/rand_core/latest/rand_core/trait.CryptoRng.html) trait instead?

rustaceanrob:
Updated with the notes from PR review club. Will have a few follow up PRs after this, but this ready for review.

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -718,6 +677,57 @@ impl BranchAndBoundCoinSelection {
...
 }
 
+pub(crate) fn single_random_draw(
+    required_utxos: Vec<WeightedUtxo>,
+    optional_utxos: Vec<WeightedUtxo>,
+    target_amount: u64,
+    drain_script: &Script,
+    fee_rate: FeeRate,
+    rng: &mut impl RngCore,
+) -> CoinSelectionResult {
```
storopoli: This is a public function without docstrings.
I know that the original one hasn't, but it was a non-public internal function.

Can we put some docstrings here?

crates/bdk/src/wallet/utils.rs
```diff
@@ -110,6 +111,19 @@ impl<Pk: MiniscriptKey + ToPublicKey> Satisfier<Pk> for Older {
     }
 }
 
+// The Knuth shuffling algorithm based on the original Fisher-Yates method: [https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle]
```
storopoli: This is not a docstring, my suggestion:

```suggestion
/// The Knuth shuffling algorithm based on the original
/// [Fisher-Yates method](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle).
```

storopoli (changes requested):
Also, remind me again what we decided on what to tackle in this PR yesterday?

What happens if someone doesn't have/bring his own PRNG?
He/She can't finish the transaction builder, since we only expose `finish_with_aux_rand` and no more `finish` (no PRNG)?

crates/bdk/src/wallet/coin_selection.rs
rustaceanrob: `pub(crate)` is not visible to users. I can add developer comments?

crates/bdk/src/wallet/utils.rs
rustaceanrob: Same comment as above

rustaceanrob:
> What happens if someone doesn't have/bring his own PRNG?
He/She can't finish the transaction builder, since we only expose finish_with_aux_rand and no more finish (no PRNG)?

In a `no_std` environment like embedded devices you can compile `rand`, in WASM you can use the `getrandom` build for JS, and the bindings team can just use `rand`. Otherwise the user is just using normal Rust builds. I am unsure how this function signature would effect downstream users at all, especially because it doesn't have to be cryptographic secure. I was going to make a separate PR to introduce a feature flag to bring back `finish` as a method to build transactions, but that was mostly for user convenience and shouldn't be covered here.

storopoli:
> I was going to make a separate PR to introduce a feature flag to bring back `finish` as a method to build transactions, but that was mostly for user convenience and shouldn't be covered here.

Yes, that was what I was recollecting. Great, agreed!

crates/bdk/src/wallet/coin_selection.rs
storopoli: Ok, then sorry about the nagging. It's fine as it is.

crates/bdk/src/wallet/utils.rs
storopoli: Ok, then sorry about the nagging. It's fine as it is.

storopoli (approved):
ACK 012aafeca63e152b5bc635e0b5ff6202bf896cfd

crates/bdk/src/wallet/coin_selection.rs
ValuedMammal: > Can we put some docstrings here?

I see what you're saying, I think it could go either way

crates/bdk/src/wallet/utils.rs
```diff
@@ -182,4 +198,45 @@ mod test {
...
+    }
+
+    #[test]
+    fn test_shuffle_slice_multi_vec() {
+        let seed = [0; 32];
+        let mut rng: StdRng = SeedableRng::from_seed(seed);
+        let mut test: Vec<u8> = vec![0, 1, 2, 4, 5];
+        shuffle_slice(&mut test, &mut rng);
+        assert_eq!(test, &[2, 1, 0, 4, 5]);
+        println!("{:?}", test);
```
ValuedMammal: You might have forgotten to remove this `println`, otherwise this looks great

ValuedMammal (approved):
ACK 012aafec

Not sure what caused the code coverage job to stall, you could try pushing again.

crates/bdk/src/wallet/utils.rs
rustaceanrob: Yeah this might have broke coveralls? Either way amending that.

notmandatory:
@rustaceanrob can you please rebase this one to pick up all the recent changes? Then if it's all still good please move it to ready to review and hopefully can be one of the next ones in.

rustaceanrob:
rebased and CI passing

storopoli (approved):
ACK 736e14e69ac9ee4b1cc416582eb865ed90786179

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -1090,13 +1100,12 @@ mod test {
     }
 
     #[test]
+    #[ignore]
```
notmandatory: ❓ Why are these tests ignored?

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -758,14 +763,12 @@ pub enum TxOrdering {
 
 impl TxOrdering {
     /// Sort transaction inputs and outputs by [`TxOrdering`] variant
-    pub fn sort_tx(&self, tx: &mut Transaction) {
+    pub fn sort_tx(self, tx: &mut Transaction, rng: &mut impl RngCore) {
```
notmandatory: ❓ For consistency shouldn't this be called `sort_tx_with_aux_rand`? Then in the follow-up PR that adds the `rand` feature flag we can cfg back in a `sort_tx` without the `rng` param?

crates/wallet/src/wallet/coin_selection.rs
rustaceanrob: To avoid adding `RngCore` to the `CoinSelection` trait, `single_random_draw` was moved into its own function. From the code comment:
>  // In this case bnb won't find a suitable match and single random draw will select three outputs

So this test does not actually test "bnb_coin_selection", instead it tests that the case in which BnB fails defaults back to single random draw silently. In this PR the BnB no longer fails silently, and single random draw is called explicitly in the error case. See `crates/wallet/src/wallet/coin_selection.rs` line 520 in the diff. I need to take my own advice and add the ignore reason. Either that or we remove it.

For the second one, I think there is no exact match, and the BnB previously failed back to SRD, but I will double check.

crates/wallet/src/wallet/tx_builder.rs
rustaceanrob: Makes sense, will amend soon.

rustaceanrob: Renamed

crates/wallet/src/wallet/coin_selection.rs
rustaceanrob: Added `ignore` reasons

notmandatory:
I added #1479 as a reminder to add back `finish()` and `sort_tx()` functions with `rand` when `std` is enabled.

rustaceanrob:
@notmandatory was nice enough to go through the effort to make each of the public functions and methods that require an `RngCore` have a feature-gated alternative that uses the same naming as before. His commit is strictly better than my first commit because 1. it reduces the size of the diff and makes the changes more legible 2. would be non-breaking for the majority of users, excluding WASM. I cherry-picked it and added it to this PR, would appreciate a quick review.

notmandatory:
Thanks for pulling in my commit @rustaceanrob, this now fixes #1479 too.

ValuedMammal (approved):
ACK 4bddb0de6262fb4014d51baf8c9453eb45a3d0ef

Thank you @rustaceanrob @notmandatory

notmandatory (approved):
ACK 4bddb0de6262fb4014d51baf8c9453eb45a3d0ef

PR #1397, opened 2024-04-04
Introduce proptesting, starting with `CheckPoint::range`

evanlinjin:
### Description

This is based on bitcoindevkit/bdk#1369. It made sense for me to introduce the `CheckPoint::range` test as a proptest. I think we should also start introducing it for other methods too.

### Changelog notice

* Added proptest for `CheckPoint::range`.

#### All Submissions:

#### New Features:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -33,6 +33,7 @@ jobs:
           cargo update -p zstd-sys --precise "2.0.8+zstd.1.5.5"
           cargo update -p time --precise "0.3.20"
           cargo update -p home --precise "0.5.5"
+          cargo update -p proptest --precise "1.2.0"
       - name: Build
         run: cargo build ${{ matrix.features }}
       - name: Test
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ cargo update -p time --precise "0.3.20"
 cargo update -p jobserver --precise "0.1.26"
 # home 0.5.9 has MSRV 1.70.0
 cargo update -p home --precise "0.5.5"
+# proptest 1.4.0 has MSRV 1.65.0
+cargo update -p proptest --precise "1.2.0"
 ```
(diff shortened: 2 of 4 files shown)


.github/workflows/cont_integration.yml
```diff
@@ -33,6 +33,7 @@ jobs:
           cargo update -p zstd-sys --precise "2.0.8+zstd.1.5.5"
           cargo update -p time --precise "0.3.20"
           cargo update -p home --precise "0.5.5"
+          cargo update -p proptest --precise "1.2.0"
```
LLFourn, line 36: Why do we need all these pins?

We don't have to test on MSRV we just have to check that it builds on MSRV. dev-dependencies shouldn't have to be pinned. Or at least that's the policy I'd like to adopt.

See sophisticated technology on how to do this: https://github.com/bitcoindevkit/coin-select/blob/990226a982bd6036bed5a7674fe57d6f9d48e49e/.github/workflows/test.yml#L40

cc @notmandatory

evanlinjin: Maybe in the future PR? I commented on discord as well about this.

LLFourn: Ok issue created: https://github.com/bitcoindevkit/bdk/issues/1398

LLFourn (approved):
ACK c5e5cd0c032d5ba646576fd308ac4e566483129e

LGTM. Also in case you hadn't noticed there is also this tool: https://proptest-rs.github.io/proptest/proptest/state-machine.html

It works very nicely when you have a "model" you want to replicate. So imagine you want to compare your `CheckPoint` to how a `BTreeMap` would behave. You generate state transitions and apply them to both and compare them to see if the `CheckPoint` properly imitates a `BTreeMap`. Of course you don't need this crate to do it, it's just a structured way of doing it.

PR #1402, opened 2024-04-11
[wallet] Improve address API

### LINKED ISSUE #898: Improve wallet address API
`get_address` with an enum of `AddressIndex` no longer makes much sense because the `Peek` variant doesn't require `Wallet` to be mutable but the others do.

Instead I think the methods should roughly follow what's in `KeychainTracker` right now.

- `LastUnused` should be removed and replaced with a method `unused_addresses` which returns and iterator. If you really want the last one then you can do `wallet.unused_addresses().last()`. I know this is not what `LastUnused` does but this is what most people thinks it does. I'm not sure what 
- `get_address` should instead just take an index and so should do what `Peek` currently does.
- getting a new address should be handled by a new method `reveal_next_address()`
- And revealing up to a certain index can be `reveal_addresses_to`
- The `mark_used` and `unmark_used` feature in `KeychainTxOutIndex` should probably be available to users of `Wallet`.
- The methods should all take a `KeychainKind` as their argument (rather than having separate methods for each).

ValuedMammal:
Improvements to the wallet address API, see commit messages for details.

### Notes to the reviewers

The logic of getting addresses is roughly the same as before when using `AddressIndex`, following this mapping:

- `New` -> `reveal_next_address`
- `LastUnused` -> `next_unused_address` (assuming this is what `LastUnused` really means)
- `Peek` -> `peek_address`

Wondering whether it makes sense to expose [`is_used`](https://github.com/bitcoindevkit/bdk/blob/358e842dcda0eb867d229881823c76dba1d7cce1/crates/chain/src/keychain/txout_index.rs#L236) for Wallet as well.

fixes #898

### Changelog notice

Added:

- Added Wallet methods:
  - `peek_address`
  - `reveal_next_address`
  - `next_unused_address`
  - `reveal_addresses_to`
  - `list_unused_addresses`
  - `mark_used`
  - `unmark_used`

Removed:

- Removed Wallet methods:
  - `get_address`
  - `get_internal_address`
  - `try_get_address`
  - `try_get_internal_address`

- Removed type AddressIndex

### Checklists

### Feature

### DIFF
--- a/crates/bdk/src/wallet/mod.rs
+++ b/crates/bdk/src/wallet/mod.rs
@@ -179,7 +179,7 @@ impl
 }
 
 /// The address index selection strategy to use to derived an address from the wallet's external
-/// descriptor. See [`Wallet::get_address`]. If you're unsure which one to use use `WalletIndex::New`.
+/// descriptor.
 #[derive(Debug)]
 pub enum AddressIndex {
     /// Return a new address after incrementing the current descriptor index.
@@ -253,36 +253,6 @@ impl Wallet {
                 NewError::Descriptor(e) => e,
                 NewError::Persist(_) => unreachable!("mock-write must always succeed"),
             })
-    }
-}
-
-impl Wallet {
-    /// Infallibly return a derived address using the external descriptor, see [`AddressIndex`] for
-    /// available address index selection strategies. If none of the keys in the descriptor are derivable
-    /// (i.e. does not end with /*) then the same address will always be returned for any [`AddressIndex`].
-    ///
-    /// # Panics
-    ///
-    /// This panics when the caller requests for an address of derivation index greater than the
-    /// BIP32 max index.
-    pub fn get_address(&mut self, address_index: AddressIndex) -> AddressInfo {
-        self.try_get_address(address_index).unwrap()
-    }
-
-    /// Infallibly return a derived address using the internal (change) descriptor.
-    ///
-    /// If the wallet doesn't have an internal descriptor it will use the external descriptor.
-    ///
-    /// see [`AddressIndex`] for available address index selection strategies. If none of the keys
-    /// in the descriptor are derivable (i.e. does not end with /*) then the same address will always
-    /// be returned for any [`AddressIndex`].
-    ///
-    /// # Panics
-    ///
-    /// This panics when the caller requests for an address of derivation index greater than the
-    /// BIP32 max index.
-    pub fn get_internal_address(&mut self, address_index: AddressIndex) -> AddressInfo {
-        self.try_get_internal_address(address_index).unwrap()
     }
 }
 
@@ -803,13 +773,7 @@ impl Wallet {
     /// # Errors
     ///
     /// If writing to persistent storage fails.
-    pub fn reveal_next_address(
-        &mut self,
-        keychain: KeychainKind,
-    ) -> Result<AddressInfo, D::WriteError>
-    where
-        D: PersistBackend<ChangeSet>,
-    {
+    pub fn reveal_next_address(&mut self, keychain: KeychainKind) -> anyhow::Result<AddressInfo> {
         let keychain = self.map_keychain(keychain);
         let ((index, spk), index_changeset) = self.indexed_graph.index.reveal_next_spk(&keychain);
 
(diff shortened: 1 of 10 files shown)


evanlinjin:
If `peek_address` and `get_address` does the same thing, can we get rid of `get_address`?

ValuedMammal:
> If `peek_address` and `get_address` does the same thing, can we get rid of `get_address`?

Yes. I also feel that replacing `get_address` with `peek_address(KeychainKind::External, 0)` is slightly less convenient than before. `get_address` appears quite often in wallet tests and (although tests shouldn't be a driver of API choices) I think should be replaced by something similarly convenient.

How about something like this:

```rust
    /// Get the first derived address of the external keychain without revealing it.
    ///
    /// Shorthand for calling [`peek_address`](Self::peek_address) at index 0.
    pub fn first_address(&self) -> AddressInfo {
        self.peek_address(KeychainKind::External, 0)
    }
```

evanlinjin:
@ValuedMammal Do you mean inconvenient as in we need to refactor many of the tests? As you mentioned, I don't think avoiding large test refactors should be a driving factor in API design.

For the `first_address` method you proposed, is the main motivation for non-wildcard descriptors? I think it makes sense if it was marketed as such.

ValuedMammal:
> Do you mean inconvenient as in we need to refactor many of the tests? As you mentioned, I don't think avoiding large test refactors should be a driving factor in API design.
> 
> For the `first_address` method you proposed, is the main motivation for non-wildcard descriptors? I think it makes sense if it was marketed as such.

"Inconvenient" in the sense that it requires an additional function arg compared to the current `get_address(New)`.

I think any descriptor is capable of producing a first address, wildcard or not. Alternatively, we drop `get_address` and use `peek_address` everywhere. Simpler the better. Actually, `next_unused_address` works just fine here as well!

ValuedMammal:
- Rebased on 52f395555792a2f24ae7248d98da318bdad5dadb
- Finally disposed of `get_address`

crates/bdk/src/wallet/mod.rs
```diff
@@ -703,116 +648,159 @@ impl<D> Wallet<D> {
...
-    /// A `PersistBackend<ChangeSet>::WriteError` will result if unable to persist the new address
-    /// to the `PersistBackend`.
+    /// Peek an address of the given `keychain` at `index` without revealing it.
     ///
     /// # Panics
     ///
     /// This panics when the caller requests for an address of derivation index greater than the
-    /// BIP32 max index.
-    pub fn try_get_address(
+    /// [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) max index.
```
evanlinjin, line 628: I think it will be good to note that for non-wildcard descriptors, we will return the address at every provided index.

evanlinjin (changes requested):
LGTM, just a small addition to docs and I will be happy to ACK/merge!

crates/bdk/src/wallet/mod.rs
```diff
@@ -703,116 +648,159 @@ impl<D> Wallet<D> {
...
-                if !spk_iter.descriptor().has_wildcard() {
-                    peek_index = 0;
-                }
-                let (index, spk) = spk_iter
-                    .nth(peek_index as usize)
-                    .expect("derivation index is out of bounds");
-                (index, spk, None)
-            }
-        };
+        let ((index, spk), index_changeset) = self.indexed_graph.index.next_unused_spk(&keychain);
```
ValuedMammal, line 710: I believe this should also use `map_keychain`. My mistake for missing it, fixed in 0648323ec6642425a2b977ba95d4d485eef399c2

ValuedMammal:
The last push includes documentation improvements and the fix for `map_keychain`.

📌 We're using `expect` when converting an spk to a bitcoin `Address` under the assumption that the script will be valid. Is this something we can always guarantee? Food for thought I guess.

LagginTimes:
Everything looks pretty good to me. :+1:

crates/bdk/src/wallet/mod.rs
evanlinjin: Good catch

evanlinjin (approved):
ACK d39b319ddf6bf3365d5d01841a8d366c6231db06

PR #1408, opened 2024-04-18
Fix: enable blocking-https-rustls feature on esplora client

thunderbiscuit:
The [`blocking` feature on the rust-esplora-client library](https://github.com/bitcoindevkit/rust-esplora-client/blame/master/Cargo.toml#L35) changed from ureq to minreq, which does not come with https enabled by default, breaking previously working code that simply enabled the `blocking` feature on the `bdk_esplora` crate.

This change will enable what is currently the "default https" [flag for the minreq library](https://docs.rs/minreq/latest/minreq/#https-or-https-rustls) when using the `blocking` feature on bdk_esplora, reverting that breaking change.

### Notes to the reviewers

Another way we could do this (let me know if this is preferable) is to add a new feature called `blocking-https-rustls`:
```rust
blocking = ["esplora-client/blocking"]
blocking-https-rustls = ["esplora-client/blocking-https-rustls"]
```

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/esplora/Cargo.toml
+++ b/crates/esplora/Cargo.toml
@@ -27,9 +27,10 @@ electrsd = { version= "0.27.1", features = ["bitcoind_25_0", "esplora_a33e97e1",
 tokio = { version = "1", features = ["rt", "rt-multi-thread", "macros"] }
 
 [features]
-default = ["std", "async-https", "blocking"]
+default = ["std", "async-https", "blocking-https-rustls"]
 std = ["bdk_chain/std"]
 async = ["async-trait", "futures", "esplora-client/async"]
 async-https = ["async", "esplora-client/async-https"]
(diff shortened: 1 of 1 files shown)


notmandatory:
🔧 I'd prefer you add the `blocking-https-rustls` feature which should be used in the `default` feature since I expect it to be the most common case. But I'd like to keep the plain `blocking` feature for those rare cases when someone doesn't want the extra https dependencies (ie. they're using a TOR or some other proxy).

thunderbiscuit:
Sounds good. I fixed the features as per recommendation.

storopoli (commented):
ACK d3a14d411d2a1a37f07e42c28eae3a1836aec14d

evanlinjin (approved):
ACK d3a14d411d2a1a37f07e42c28eae3a1836aec14d

PR #1411, opened 2024-04-24
feat: update `keychain::Balance` to use `bitcoin::Amount`

### LINKED ISSUE #823: `Balance` should use `bitcoin::Amount`
**Describe the bug**  
Currently, the `Balance` type returned by `get_balance()` only uses non-descript `u64`s, while not even stating a denomination anywhere in the documentation (I assume it's sats?). To avoid any confusion and make this much less error prone BDK should reuse the `Amount` type provided by `rust-bitcoin`.

Also, the documentation could generally be improved, e.g., clarifying what exactly it means that some funds are `immature`/`trusted_pending`/`untrusted_pending`/`confirmed`.

**To Reproduce**  
Call `Wallet::get_balance()`.

**Expected behavior**  
Receive a meaningful return value.

**Additional context**  
This is in particular confusing since similar `u64` values are used in the Lightning space very often to denominate millisatoshis. It's best practice to append `_msat` to the names of such variables, which however still doesn't make use of the type system to avoid any conversion errors (we can't, as `Amount` doesn't  allow to express millisatoshis). As BDK doesn't have to deal with millisatoshis, it really should make use of `Amount` here.

oleonardolima:
fixes #823 

### Description

It's being used on `Balance`, and throughout the code, an `u64` represents the amount, which relies on the user to infer its sats, not millisats, or any other representation.

It updates the usage of `u64` on `Balance`, and other APIs:
- `TxParams::add_recipient`
- `KeyChainTxOutIndex::sent_and_received`, `KeyChainTxOutIndex::net_value`
-  `SpkTxOutIndex::sent_and_received`, `SpkTxOutIndex::net_value`

### Notes to the reviewers

It updates some of the APIs to expect the `bitcoin::Amount`, but it does not update internal usage of u64, such as `TxParams` still expects and uses `u64`, please see the PR comments for related discussion.

### Changelog notice

- Changed the `keychain::Balance` struct fields to use `Amount` instead of `u64`.
- Changed the `add_recipient` method on `TxBuilder` implementation to expect `bitcoin::Amount`.
- Changed the `sent_and_received`, and `net_value` methods on `KeyChainTxOutIndex` to expect `bitcoin::Amount`.
- Changed the `sent_and_received`, and `net_value` methods on `SpkTxOutIndex` to expect `bitcoin::Amount`.

#### All Submissions:

#### New Features:

#### Bugfixes:


crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -718,6 +718,7 @@ impl<'a, Cs: CoinSelectionAlgorithm> TxBuilder<'a, Cs, CreateTx> {
         self
     }
 
+    // TODO: (@leonardo) Should this expect/use `bitcoin::Amount` instead ? Would it be a huge breaking change ?
```
oleonardolima: what do you folks think ? do you see any problem in tackling this change as well ?
cc @notmandatory @ValuedMammal @storopoli

crates/chain/src/spk_txout_index.rs
```diff
@@ -270,6 +270,7 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
         self.spk_indices.get(script)
     }
 
+    // TODO: (@leonardo) Should this also be updated to return `(bitcoin::Amount, bitcoin::Amount)` instead of (u64, u64)
```
oleonardolima: same as above, what do you folks think ? do you see any problem in tackling this change as well ?
cc @notmandatory @ValuedMammal @storopoli

crates/bdk/src/wallet/tx_builder.rs
oleonardolima: _NOTE:_ the changes aren't hard (will do on separate commits anyway), it just impact some APIs

ValuedMammal: yeah I'm open to having this change assuming we call `to_sat` before pushing onto recipients.

oleonardolima: hmm, that's ok too, but why we can't we use Amount on the recipients, as well ?

crates/chain/src/spk_txout_index.rs
storopoli: I don't see an issue in changing this to `Amount`.

crates/bdk/src/wallet/tx_builder.rs
storopoli: Same, go for it.

ValuedMammal: > why we can't we use Amount on the recipients, as well ?

I don't think it adds anything, because the library assumes all amounts are in satoshis and we should keep this invariant.

notmandatory: @ValuedMammal do you mean not to use `Amount` for the `amount` param? I agree internally in `TxParams` it should stay sats but for the public API wouldn't `Amount` be safer?

notmandatory: @oleonardolima yes thanks for taking this one on, it's one we'd still like to get into the 1.0 alpha.

ValuedMammal: yes for changing this method `add_recipient` to accept an `Amount`. I don't think the type in `TxParams` needs to change in this PR

oleonardolima: Good, I think we should update the `set_recipient` to accept `Amount` too, right ? @ValuedMammal @notmandatory wdyt ?

oleonardolima:
I think it's ready for a new round of reviews

_I'll move the previous conversations to resolved after it, and if it's everything looks good on this new review round)_

storopoli:
> I think it's ready for a new round of reviews

Can you move it from Draft PR to a full PR?

oleonardolima:
> > I think it's ready for a new round of reviews
> 
> Can you move it from Draft PR to a full PR?

Yes, forgot to do it

oleonardolima:
~Oh, I forgot to run the tests on docs, working on it 😅~
Done!

crates/bdk/src/wallet/tx_builder.rs
notmandatory: Yes anywhere in the public API where we can replace a `u32` (or `i64`) representing sats with `Amount` (or `SignedAmount`) I'd like to do it it.

crates/bdk/src/wallet/coin_selection.rs
```diff
@@ -92,7 +92,10 @@
 //!     .unwrap();
 //! let psbt = {
 //!     let mut builder = wallet.build_tx().coin_selection(AlwaysSpendEverything);
-//!     builder.add_recipient(to_address.script_pubkey(), 50_000);
+//!     builder.add_recipient(
+//!         to_address.script_pubkey(),
+//!         bitcoin::Amount::from_sat(50_000),
+//!     );
```
storopoli: Isn't this covered in the

```rust
//! # use bitcoin::*;
```

on line 28?

```suggestion
//!     builder.add_recipient(
//!         to_address.script_pubkey(),
//!         Amount::from_sat(50_000),
//!     );
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -1175,7 +1175,7 @@ impl Wallet {
     /// let psbt = {
     ///    let mut builder =  wallet.build_tx();
     ///    builder
-    ///        .add_recipient(to_address.script_pubkey(), 50_000);
+    ///        .add_recipient(to_address.script_pubkey(), bitcoin::Amount::from_sat(50_000));
```
storopoli: It's already imported in line 40

```suggestion
    ///        .add_recipient(to_address.script_pubkey(), Amount::from_sat(50_000));
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -1557,7 +1557,7 @@ impl Wallet {
     /// let mut psbt = {
     ///     let mut builder = wallet.build_tx();
     ///     builder
-    ///         .add_recipient(to_address.script_pubkey(), 50_000)
+    ///         .add_recipient(to_address.script_pubkey(), bitcoin::Amount::from_sat(50_000))
```
storopoli: Same

```suggestion
    ///         .add_recipient(to_address.script_pubkey(), Amount::from_sat(50_000))
```

crates/bdk/src/wallet/mod.rs
```diff
@@ -1730,7 +1730,7 @@ impl Wallet {
     /// # let to_address = Address::from_str("2N4eQYCbKUHCCTUjBJeHcJp9ok6J2GZsTDt").unwrap().assume_checked();
     /// let mut psbt = {
     ///     let mut builder = wallet.build_tx();
-    ///     builder.add_recipient(to_address.script_pubkey(), 50_000);
+    ///     builder.add_recipient(to_address.script_pubkey(), bitcoin::Amount::from_sat(50_000));
```
storopoli: Same

```suggestion
    ///     builder.add_recipient(to_address.script_pubkey(), Amount::from_sat(50_000));
```

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -29,7 +29,7 @@
 //!
 //! tx_builder
 //!     // Create a transaction with one output to `to_address` of 50_000 satoshi
-//!     .add_recipient(to_address.script_pubkey(), 50_000)
+//!     .add_recipient(to_address.script_pubkey(), bitcoin::Amount::from_sat(50_000))
```
storopoli: Already imported in line 18

```suggestion
//!     .add_recipient(to_address.script_pubkey(), Amount::from_sat(50_000))
```

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -94,8 +94,8 @@ impl TxBuilderContext for BumpFee {}
 ///     let mut builder = wallet.build_tx();
 ///     builder
 ///         .ordering(TxOrdering::Untouched)
-///         .add_recipient(addr1.script_pubkey(), 50_000)
-///         .add_recipient(addr2.script_pubkey(), 50_000);
+///         .add_recipient(addr1.script_pubkey(), bitcoin::Amount::from_sat(50_000))
+///         .add_recipient(addr2.script_pubkey(), bitcoin::Amount::from_sat(50_000));
```
storopoli: You are importing in L50

```suggestion
///         .add_recipient(addr1.script_pubkey(), Amount::from_sat(50_000))
///         .add_recipient(addr2.script_pubkey(), Amount::from_sat(50_000));
```

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -104,7 +104,7 @@ impl TxBuilderContext for BumpFee {}
 ///     let mut builder = wallet.build_tx();
 ///     builder.ordering(TxOrdering::Untouched);
 ///     for addr in &[addr1, addr2] {
-///         builder.add_recipient(addr.script_pubkey(), 50_000);
+///         builder.add_recipient(addr.script_pubkey(), bitcoin::Amount::from_sat(50_000));
```
storopoli: imported L50

```suggestion
///         builder.add_recipient(addr.script_pubkey(), Amount::from_sat(50_000));
```

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -274,7 +274,10 @@ impl<'a, Cs, Ctx> TxBuilder<'a, Cs, Ctx> {
     ///
     /// let builder = wallet
     ///     .build_tx()
-    ///     .add_recipient(to_address.script_pubkey(), 50_000)
+    ///     .add_recipient(
+    ///         to_address.script_pubkey(),
+    ///         bitcoin::Amount::from_sat(50_000),
```
storopoli: imported L50

```suggestion
    ///         Amount::from_sat(50_000),
```

crates/bdk/tests/psbt.rs
```diff
@@ -1,5 +1,6 @@
-use bdk::bitcoin::{Amount, FeeRate, Psbt, TxIn};
+use bdk::bitcoin::{FeeRate, Psbt, TxIn};
 use bdk::{psbt, KeychainKind, SignOptions};
+use bitcoin::Amount;
```
storopoli: Why this change?

crates/chain/src/keychain.rs
```diff
@@ -90,26 +90,26 @@ impl<K> AsRef<BTreeMap<K, u32>> for ChangeSet<K> {
...
     /// This is the balance you can spend right now that shouldn't get cancelled via another party
     /// double spending it.
-    pub fn trusted_spendable(&self) -> u64 {
+    pub fn trusted_spendable(&self) -> bitcoin::Amount {
         self.confirmed + self.trusted_pending
     }
 
     /// Get the whole balance visible to the wallet.
-    pub fn total(&self) -> u64 {
+    pub fn total(&self) -> bitcoin::Amount {
```
storopoli: Why not import `bitcoin::Amount` and reduce the clutter?

crates/chain/src/keychain/txout_index.rs
```diff
@@ -285,7 +289,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// This calls [`SpkTxOutIndex::net_value`] internally.
     ///
     /// [`sent_and_received`]: Self::sent_and_received
-    pub fn net_value(&self, tx: &Transaction, range: impl RangeBounds<K>) -> i64 {
+    pub fn net_value(&self, tx: &Transaction, range: impl RangeBounds<K>) -> SignedAmount {
```
storopoli, line 292: Why this needs to be a `SignedAmount`?

crates/chain/src/spk_txout_index.rs
```diff
@@ -301,9 +305,11 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
     /// for calling [`sent_and_received`] and subtracting sent from received.
     ///
     /// [`sent_and_received`]: Self::sent_and_received
-    pub fn net_value(&self, tx: &Transaction, range: impl RangeBounds<I>) -> i64 {
+    pub fn net_value(&self, tx: &Transaction, range: impl RangeBounds<I>) -> SignedAmount {
```
storopoli, line 308: Why this needs to be a `SignedAmount`?

crates/chain/src/tx_graph.rs
```diff
@@ -1155,27 +1155,27 @@ impl<A: Anchor> TxGraph<A> {
...
         mut trust_predicate: impl FnMut(&OI, &Script) -> bool,
     ) -> Result<Balance, C::Error> {
-        let mut immature = 0;
-        let mut trusted_pending = 0;
-        let mut untrusted_pending = 0;
-        let mut confirmed = 0;
+        let mut immature = bitcoin::Amount::ZERO;
+        let mut trusted_pending = bitcoin::Amount::ZERO;
+        let mut untrusted_pending = bitcoin::Amount::ZERO;
+        let mut confirmed = bitcoin::Amount::ZERO;
```
storopoli: Why not import `bitcoin::Amount` and reduce the clutter?

crates/chain/tests/test_spk_txout_index.rs
```diff
@@ -20,14 +22,23 @@ fn spk_txout_sent_and_received() {
...
+    );
+    assert_eq!(
+        index.sent_and_received(&tx1, ..1),
+        (Amount::from_sat(0), Amount::from_sat(42_000))
+    );
+    assert_eq!(
+        index.sent_and_received(&tx1, 1..),
+        (Amount::from_sat(0), Amount::from_sat(0))
+    );
+    assert_eq!(index.net_value(&tx1, ..), SignedAmount::from_sat(42_000));
```
storopoli, line 37: Why this needs to be a `SignedAmount`?

crates/chain/tests/test_spk_txout_index.rs
```diff
@@ -53,10 +64,19 @@ fn spk_txout_sent_and_received() {
...
+    );
+    assert_eq!(
+        index.sent_and_received(&tx2, ..1),
+        (Amount::from_sat(42_000), Amount::from_sat(30_000))
+    );
+    assert_eq!(
+        index.sent_and_received(&tx2, 1..),
+        (Amount::from_sat(0), Amount::from_sat(20_000))
+    );
+    assert_eq!(index.net_value(&tx2, ..), SignedAmount::from_sat(8_000));
```
storopoli, line 79: Why this needs to be a `SignedAmount`?

storopoli (changes requested):
Some suggestions.

Most important, why we are using `SignedAmount` in some places?

crates/chain/src/keychain/txout_index.rs
oleonardolima: It's because `net_value` used `i64` previously, and to keep the standard (and makes sense in my opinion given the  fn name) I used the `SignedAmount`, same for the other comments, they are all `net_value` references.

crates/bdk/tests/psbt.rs
oleonardolima: I have no idea, probably rust-analyzer stuff 🤔, fixed it.

crates/chain/src/keychain.rs
oleonardolima: Done, much better now :)

oleonardolima:
> Some suggestions.
> 
> Most important, why we are using `SignedAmount` in some places?

It's because net_value used i64 previously, and to keep the standard (and makes sense in my opinion given the fn name) I used the SignedAmount, same for the other comments, they are all net_value references.

_(replying on the overall comment too)_

crates/chain/src/keychain/txout_index.rs
storopoli: Oh I see...

storopoli (approved):
ACK 362a67aefd5baea6a321721df05e9aff79894c86

Great work!

crates/chain/src/spk_txout_index.rs
```diff
@@ -301,9 +305,11 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
...
         let (sent, received) = self.sent_and_received(tx, range);
-        received as i64 - sent as i64
+        (received - sent)
+            .to_signed()
+            .expect("Amount conversion to SignedAmount error")
```
ValuedMammal: I think you want to do `to_signed` on each Amount before doing the subtraction.

oleonardolima: Yes, you're right, fixed it.

example-crates/example_cli/src/lib.rs
```diff
@@ -506,7 +506,7 @@ where
             let chain = &*chain.lock().unwrap();
             fn print_balances<'a>(
                 title_str: &'a str,
-                items: impl IntoIterator<Item = (&'a str, u64)>,
+                items: impl IntoIterator<Item = (&'a str, Amount)>,
```
ValuedMammal, line 509: This function intends to print out a value in sats, so for this to work we need another `to_sat` inside the function.

```diff
             fn print_balances<'a>(
                 title_str: &'a str,
                 items: impl IntoIterator<Item = (&'a str, Amount)>,
             ) {
                 println!("{}:", title_str);
                 for (name, amount) in items.into_iter() {
-                    println!("    {:<10} {:>12} sats", name, amount)
+                    println!("    {:<10} {:>12} sats", name, amount.to_sat())
```

crates/chain/src/spk_txout_index.rs
```diff
@@ -301,9 +305,14 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
...
-    pub fn net_value(&self, tx: &Transaction, range: impl RangeBounds<I>) -> i64 {
+    pub fn net_value(&self, tx: &Transaction, range: impl RangeBounds<I>) -> SignedAmount {
         let (sent, received) = self.sent_and_received(tx, range);
-        received as i64 - sent as i64
+
+        let error_msg = "`Amount::to_signed()` conversion to `SignedAmount` error";
+        let signed_sent = sent.to_signed().expect(error_msg);
+        let signed_received = received.to_signed().expect(error_msg);
+
+        signed_received - signed_sent
```
ValuedMammal:
```suggestion
        received.to_signed().expect("valid SignedAmount")
            - sent.to_signed().expect("valid SignedAmount")
```

example-crates/example_cli/src/lib.rs
oleonardolima: Nice, good catch! Fixed it.

crates/bdk/src/wallet/mod.rs
```diff
@@ -978,7 +978,7 @@ impl Wallet {
     /// let tx = &psbt.clone().extract_tx().expect("tx");
     /// let (sent, received) = wallet.sent_and_received(tx);
     /// ```
-    pub fn sent_and_received(&self, tx: &Transaction) -> (u64, u64) {
+    pub fn sent_and_received(&self, tx: &Transaction) -> (Amount, Amount) {
```
ValuedMammal, line 981: In this case, the docs for this method shouldn't say that the returned amounts are "in satoshis".

crates/bdk/tests/wallet.rs
```diff
@@ -561,14 +561,15 @@ fn test_create_tx_change_policy_no_internal() {
...
-        .add_recipient(addr.script_pubkey(), 25_000)
+        .add_recipient(addr.script_pubkey(), Amount::from_sat(25_000))
         .do_not_spend_change();
     assert!(matches!(
         builder.finish(),
         Err(CreateTxError::ChangePolicyDescriptor)
     ));
 }
 
+// TODO: (@leonardo) Can we update this one to return an `bitcoin::Amount` instead ?
```
ValuedMammal: If you mean can we change `Wallet::calculate_fee` to return an `Amount`, I think possibly yes. But to avoid scope creep we should push it to a future PR.

```suggestion
```

ValuedMammal (commented):
Sorry my reviews have been intermittent. I think this is good.

The commit history could be cleaned up slightly:

- In df92624b011668216bba9908f6bd4fbb6db06457 we want to refer to `TxBuilder::add_recipient` (not CoinSelectionAlgorithm)

- Some changes in later commits (like fixing imports and wallet examples) could be squashed into earlier ones, but this not a blocker in my opinion.

Some other thoughts: I agree there is potentially more to do like updating `calculate_fee` and `Psbt::fee_amount` to use `Amount`. I'm still unsure how we should handle amounts internally. Honestly I don't mind leaving it as `u64`. If we decide that all amounts should be strongly typed, then `Amount` seems like a logical choice, my only concern is we should avoid the possibility of mixing up denominations.

crates/bdk/tests/wallet.rs
oleonardolima: Yes, that's what I mean. Cool, I'll address it in another PR.

crates/bdk/src/wallet/mod.rs
oleonardolima: Yes, updated it, thanks!

oleonardolima:
> Sorry my reviews have been intermittent. I think this is good.
> 
> The commit history could be cleaned up slightly:
> 
> * In [df92624](https://github.com/bitcoindevkit/bdk/commit/df92624b011668216bba9908f6bd4fbb6db06457) we want to refer to `TxBuilder::add_recipient` (not CoinSelectionAlgorithm)
> * Some changes in later commits (like fixing imports and wallet examples) could be squashed into earlier ones, but this not a blocker in my opinion.
> 
> Some other thoughts: I agree there is potentially more to do like updating `calculate_fee` and `Psbt::fee_amount` to use `Amount`. I'm still unsure how we should handle amounts internally. Honestly I don't mind leaving it as `u64`. If we decide that all amounts should be strongly typed, then `Amount` seems like a logical choice, my only concern is we should avoid the possibility of mixing up denominations.

@ValuedMammal No worries, thanks for the comments and suggestions! I tried to address them, please let me know if something remains or if new needed changes appear.

crates/bdk/src/wallet/mod.rs
```diff
@@ -950,10 +950,10 @@ impl Wallet {
     /// [`insert_txout`]: Self::insert_txout
     pub fn calculate_fee_rate(&self, tx: &Transaction) -> Result<FeeRate, CalculateFeeError> {
         self.calculate_fee(tx)
-            .map(|fee| bitcoin::Amount::from_sat(fee) / tx.weight())
+            .map(|fee| Amount::from_sat(fee) / tx.weight())
     }
 
-    /// Compute the `tx`'s sent and received amounts (in satoshis).
+    /// Compute the `tx`'s sent and received amounts (in [`bitcoin::Amount`]).
```
ValuedMammal: What do you think of this?

```suggestion
    /// Compute the `tx`'s sent and received [`Amount`]s.
```

ValuedMammal (approved):
ACK 2e6aeb7499d9c4fadce4024fe26666976447ac2e

ValuedMammal:
Actually, changing `calculate_fee` and `Psbt::fee_amount` should be fairly easy to add to this PR but I'll let you decide if you want to tackle it here or later. Besides that, I'm not sure if there's still more places in the public API that would benefit from using `Amount`.

oleonardolima:
> Actually, changing `calculate_fee` and `Psbt::fee_amount` should be fairly easy to add to this PR but I'll let you decide if you want to tackle it here or later. Besides that, I'm not sure if there's still more places in the public API that would benefit from using `Amount`.

I updated `calculate_fee` and `fee_amount` on ac762f4a862db227432aa7daeeee424d2b653af3, but ended up touching other parts as well.

I would rather move this commit ac762f4a862db227432aa7daeeee424d2b653af3 to another PR, and keep only the previous ones here, wdyt ?

oleonardolima:
> > Actually, changing `calculate_fee` and `Psbt::fee_amount` should be fairly easy to add to this PR but I'll let you decide if you want to tackle it here or later. Besides that, I'm not sure if there's still more places in the public API that would benefit from using `Amount`.
> 
> I updated `calculate_fee` and `fee_amount` on [ac762f4](https://github.com/bitcoindevkit/bdk/commit/ac762f4a862db227432aa7daeeee424d2b653af3), but ended up touching other parts as well.
> 
> I would rather move this commit [ac762f4](https://github.com/bitcoindevkit/bdk/commit/ac762f4a862db227432aa7daeeee424d2b653af3) to another PR, and keep only the previous ones here, wdyt ?

I moved it to the other PR #1426, we can start discussing the changes on that one after this one is merged.

evanlinjin (approved):
ACK 22aa534d7648e5808414ea3adfcfb702572bd6c9

PR #1413, opened 2024-04-25
Introduce universal sync/full-scan structures for spk-based syncing

### LINKED ISSUE #1153: [wallet] Consider having a higher-level method for syncing

evanlinjin:
Fixes #1153
Replaces #1194

### Description

Introduce universal structures that represent sync/full-scan requests/results.

### Notes to the reviewers

This is based on #1194 but is different in the following ways:
* The functionality to print scan/sync progress is not reduced.
* `SyncRequest` and `FullScanRequest` is simplified and fields are exposed for more flexibility.

### Changelog notice

* Add universal structures for initiating/receiving sync/full-scan requests/results for spk-based syncing.
* Updated `bdk_esplora` chain-source to make use of new universal sync/full-scan structures.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/spk_client.rs
+++ b/crates/chain/src/spk_client.rs
@@ -1,0 +1,315 @@
+//! Helper types for spk-based blockchain clients.
+
+use core::{fmt::Debug, ops::RangeBounds};
+
+use alloc::{boxed::Box, collections::BTreeMap, vec::Vec};
+use bitcoin::{OutPoint, Script, ScriptBuf, Txid};
+
+use crate::{local_chain::CheckPoint, ConfirmationTimeHeightAnchor, TxGraph};
+
+/// Data required to perform a spk-based blockchain client sync.
+///
+/// A client sync fetches relevant chain data for a known list of scripts, transaction ids and
+/// outpoints. The sync process also updates the chain from the given [`CheckPoint`].
+pub struct SyncRequest {
+    /// A checkpoint for the current chain [`LocalChain::tip`].
+    /// The sync process will return a new chain update that extends this tip.
+    ///
+    /// [`LocalChain::tip`]: crate::local_chain::LocalChain::tip
+    pub chain_tip: CheckPoint,
+    /// Transactions that spend from or to these indexed script pubkeys.
+    pub spks: Box<dyn Iterator<Item = ScriptBuf> + Send>,
+    /// Transactions with these txids.
+    pub txids: Box<dyn Iterator<Item = Txid> + Send>,
+    /// Transactions with these outpoints or spent from these outpoints.
+    pub outpoints: Box<dyn Iterator<Item = OutPoint> + Send>,
+}
+
+impl SyncRequest {
+    /// Construct a new [`SyncRequest`] from a given `cp` tip.
+    pub fn from_chain_tip(cp: CheckPoint) -> Self {
+        Self {
+            chain_tip: cp,
+            spks: Box::new(core::iter::empty()),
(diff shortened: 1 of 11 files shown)


example-crates/wallet_esplora_async/src/main.rs
```diff
@@ -37,34 +36,44 @@ async fn main() -> Result<(), anyhow::Error> {
...
+            generate_inspect(KeychainKind::Internal),
+        );
 
     let mut update = client
-        .full_scan(prev_tip, keychain_spks, STOP_GAP, PARALLEL_REQUESTS)
+        .full_scan(request, STOP_GAP, PARALLEL_REQUESTS)
         .await?;
     let now = std::time::UNIX_EPOCH.elapsed().unwrap().as_secs();
-    let _ = update.tx_graph.update_last_seen_unconfirmed(now);
+    let _ = update.graph_update.update_last_seen_unconfirmed(now);
```
storopoli, line 75: Great! I was going through this example yesterday and I was having tons of issues because even without this PR it was un-reproducible code.
Thanks!

storopoli (commented):
Concept ACK.
These are great changes.


I imagine that you are planning to implement this to electrum as well? Either in this PR or in a separate PR?

example-crates/wallet_esplora_async/src/main.rs
evanlinjin: I'm sorry I don't quite understand?

example-crates/example_esplora/src/main.rs
```diff
@@ -319,29 +307,48 @@ fn main() -> anyhow::Result<()> {
...
-            )?;
+            let total_spks = request.spks.len();
+            let total_txids = request.txids.len();
+            let total_ops = request.outpoints.len();
+            request = request
+                .inspect_spks({
+                    let mut visited = 0;
+                    move |_| {
+                        visited += 1;
+                        eprintln!(" [ {:>6.2}% ]", (visited * 100) as f32 / total_spks as f32)
```
notmandatory, line 326: 😃 nice touch showing progress as percents.

crates/chain/src/spk_client.rs
```diff
@@ -0,0 +1,384 @@
...
+            chain_tip: cp,
+            spks: Box::new(core::iter::empty()),
+            txids: Box::new(core::iter::empty()),
+            outpoints: Box::new(core::iter::empty()),
+        }
+    }
+
+    /// Set the [`Script`]s that will be synced against.
+    ///
+    /// This consumes the [`SyncRequest`] and returns the updated one.
```
notmandatory, line 41: 😃  I like how you changed all the Sync/FullScan Request functions to consume and return Self, makes them much easier to use.

notmandatory (approved):
tACK c0374a0eeb155f1cb4f9d99223f6bc65d24f7df8

Thanks for the improvements! 

The examples ran fine. But I did have to change the wallet examples to use signet and the signet esplora server `http://signet.bitcoindevkit.net` since others are rate limiting and so the scans were failing with 429 errors.

PR #1416, opened 2024-04-30
refactor(chain)!: Change tx_last_seen to `Option<u64>`

### LINKED ISSUE #1396: [chain] last_seen_unconfirmed should have a default of `None` not 0
Current implementation of `TxGraph` `insert_tx` is:

```rust
    /// Inserts the given transaction into [`TxGraph`].
    ///
    /// The [`ChangeSet`] returned will be empty if `tx` already exists.
    pub fn insert_tx(&mut self, tx: Transaction) -> ChangeSet<A> {
        let mut update = Self::default();
        update.txs.insert(
            tx.txid(),
            (TxNodeInternal::Whole(tx.into()), BTreeSet::new(), 0),
        );
        self.apply_update(update)
    }
```

The 0 in the tuple represents the last time the transaction has been seen unconfirmed. I think it is wrong to set `0` here because you should be able to insert transactions into the graph (or wallet) that have never been broadcast and have never been seen. When you list canonical transactions[^1] you don't want to list transactions that have been inserted but never actually seen. In other words the field should be an `Option` which is still monotonically increasing with `None` being the smallest value.

I don't think this change affects the changeset struct, it just changes how you interpret a missing `last_seen`. 

[^1]: Incidentally we should rename `try_list_chain_txs` to `canonincal_transactions`

### LINKED ISSUE #1446: `Wallet::insert_tx` should not have `position` input
**Describe the problem**  

The `Wallet::insert_tx` method allows us to add a transaction and instantly confirm it by anchoring the transaction to the wallet tip. This is dangerous as the user is not aware of how it works internally, and may anchor the transaction to the wrong tip (which will result in nonsensical history later on).

The correct way to insert confirmed or in-mempool transactions is via a chain-source which creates a `wallet::Update`.

I propose changing this method to only insert transactions without anchors or last seen. Therefore, I think this will be tackled in bitcoindevkit/bdk#1416 (assuming that bitcoindevkit/bdk#1416 is a good idea). In other words, this method will just insert transactions that have not left "application memory", or unbroadcasted transactions.

ValuedMammal:
The PR changes the type of last_seen to `Option<u64>` for `txs` member of `TxGraph`.

This fixes an issue where unbroadcast and otherwise non-canonical transactions were returned from methods `list_chain_txs` and `Wallet::transactions` because every new tx inserted had a last_seen of 0 making it appear unconfirmed.

fixes #1446 
fixes #1396 

### Changelog notice

Changed
- Member `last_seen_unconfirmed` of `TxNode` is changed to `Option<u64>`
- Renamed `TxGraph` method `list_chain_txs` to `list_canonical_txs`
- Changed `Wallet::insert_tx` to take a single `tx: Transaction` as parameter

Added
- Add method `txs_with_no_anchor_or_last_seen` for `TxGraph`
- Add method `unbroadcast_transactions` for `Wallet`

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -258,6 +258,19 @@ impl<A> TxGraph<A> {
             })
     }
 
+    /// Iterate over graph transactions with no anchors or last-seen.
+    pub fn txs_with_no_anchor_or_last_seen(
+        &self,
+    ) -> impl Iterator<Item = TxNode<'_, Arc<Transaction>, A>> {
+        self.full_txs().filter_map(|tx| {
+            if tx.anchors.is_empty() && tx.last_seen_unconfirmed.is_none() {
+                Some(tx)
+            } else {
+                None
+            }
+        })
+    }
+
     /// Get a transaction by txid. This only returns `Some` for full transactions.
     ///
     /// Refer to [`get_txout`] for getting a specific [`TxOut`].
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -109,10 +109,11 @@ use core::{
 /// [module-level documentation]: crate::tx_graph
 #[derive(Clone, Debug, PartialEq)]
 pub struct TxGraph<A = ()> {
-    // all transactions that the graph is aware of in format: `(tx_node, tx_anchors, tx_last_seen)`
-    txs: HashMap<Txid, (TxNodeInternal, BTreeSet<A>, Option<u64>)>,
+    // all transactions that the graph is aware of in format: `(tx_node, tx_anchors)`
+    txs: HashMap<Txid, (TxNodeInternal, BTreeSet<A>)>,
     spends: BTreeMap<OutPoint, HashSet<Txid>>,
     anchors: BTreeSet<(A, Txid)>,
+    last_seen: HashMap<Txid, u64>,
 
     // This atrocity exists so that `TxGraph::outspends()` can return a reference.
     // FIXME: This can be removed once `HashSet::new` is a const fn.
@@ -125,6 +126,7 @@ impl<A> Default for TxGraph<A> {
             txs: Default::default(),
             spends: Default::default(),
             anchors: Default::default(),
+            last_seen: Default::default(),
             empty_outspends: Default::default(),
         }
     }
@@ -210,7 +212,7 @@ impl<A> TxGraph<A> {
     ///
     /// This includes txouts of both full transactions as well as floating transactions.
     pub fn all_txouts(&self) -> impl Iterator<Item = (OutPoint, &TxOut)> {
-        self.txs.iter().flat_map(|(txid, (tx, _, _))| match tx {
+        self.txs.iter().flat_map(|(txid, (tx, _))| match tx {
             TxNodeInternal::Whole(tx) => tx
                 .as_ref()
                 .output
@@ -232,7 +234,7 @@ impl<A> TxGraph<A> {
     pub fn floating_txouts(&self) -> impl Iterator<Item = (OutPoint, &TxOut)> {
         self.txs
             .iter()
-            .filter_map(|(txid, (tx_node, _, _))| match tx_node {
+            .filter_map(|(txid, (tx_node, _))| match tx_node {
                 TxNodeInternal::Whole(_) => None,
                 TxNodeInternal::Partial(txouts) => Some(
                     txouts
@@ -247,12 +249,12 @@ impl<A> TxGraph<A> {
     pub fn full_txs(&self) -> impl Iterator<Item = TxNode<'_, Arc<Transaction>, A>> {
         self.txs
             .iter()
-            .filter_map(|(&txid, (tx, anchors, last_seen))| match tx {
+            .filter_map(|(&txid, (tx, anchors))| match tx {
                 TxNodeInternal::Whole(tx) => Some(TxNode {
                     txid,
                     tx: tx.clone(),
                     anchors,
-                    last_seen_unconfirmed: *last_seen,
+                    last_seen_unconfirmed: self.last_seen.get(&txid).copied(),
                 }),
                 TxNodeInternal::Partial(_) => None,
             })
@@ -283,11 +285,11 @@ impl<A> TxGraph<A> {
     /// Get a transaction node by txid. This only returns `Some` for full transactions.
     pub fn get_tx_node(&self, txid: Txid) -> Option<TxNode<'_, Arc<Transaction>, A>> {
         match &self.txs.get(&txid)? {
-            (TxNodeInternal::Whole(tx), anchors, last_seen) => Some(TxNode {
+            (TxNodeInternal::Whole(tx), anchors) => Some(TxNode {
                 txid,
                 tx: tx.clone(),
                 anchors,
-                last_seen_unconfirmed: *last_seen,
+                last_seen_unconfirmed: self.last_seen.get(&txid).copied(),
             }),
             _ => None,
         }
@@ -517,7 +519,6 @@ impl<A: Clone + Ord> TxGraph<A> {
             (
                 TxNodeInternal::Partial([(outpoint.vout, txout)].into()),
                 BTreeSet::new(),
-                None,
             ),
         );
         self.apply_update(update)
@@ -531,7 +532,7 @@ impl<A: Clone + Ord> TxGraph<A> {
         let mut update = Self::default();
         update.txs.insert(
             tx.compute_txid(),
-            (TxNodeInternal::Whole(tx), BTreeSet::new(), None),
+            (TxNodeInternal::Whole(tx), BTreeSet::new()),
         );
         self.apply_update(update)
     }
@@ -572,8 +573,7 @@ impl<A: Clone + Ord> TxGraph<A> {
     /// [`update_last_seen_unconfirmed`]: Self::update_last_seen_unconfirmed
     pub fn insert_seen_at(&mut self, txid: Txid, seen_at: u64) -> ChangeSet<A> {
         let mut update = Self::default();
-        let (_, _, update_last_seen) = update.txs.entry(txid).or_default();
-        *update_last_seen = Some(seen_at);
+        update.last_seen.insert(txid, seen_at);
         self.apply_update(update)
     }
 
@@ -620,7 +620,7 @@ impl<A: Clone + Ord> TxGraph<A> {
             .txs
             .iter()
             .filter_map(
-                |(&txid, (_, anchors, _))| {
+                |(&txid, (_, anchors))| {
                     if anchors.is_empty() {
                         Some(txid)
                     } else {
@@ -669,10 +669,10 @@ impl<A: Clone + Ord> TxGraph<A> {
                 });
 
             match self.txs.get_mut(&txid) {
-                Some((tx_node @ TxNodeInternal::Partial(_), _, _)) => {
+                Some((tx_node @ TxNodeInternal::Partial(_), _)) => {
                     *tx_node = TxNodeInternal::Whole(wrapped_tx.clone());
                 }
-                Some((TxNodeInternal::Whole(tx), _, _)) => {
+                Some((TxNodeInternal::Whole(tx), _)) => {
                     debug_assert_eq!(
                         tx.as_ref().compute_txid(),
                         txid,
@@ -680,10 +680,8 @@ impl<A: Clone + Ord> TxGraph<A> {
                     );
                 }
                 None => {
-                    self.txs.insert(
-                        txid,
-                        (TxNodeInternal::Whole(wrapped_tx), BTreeSet::new(), None),
-                    );
+                    self.txs
+                        .insert(txid, (TxNodeInternal::Whole(wrapped_tx), BTreeSet::new()));
                 }
             }
         }
@@ -692,9 +690,8 @@ impl<A: Clone + Ord> TxGraph<A> {
             let tx_entry = self.txs.entry(outpoint.txid).or_default();
 
             match tx_entry {
-                (TxNodeInternal::Whole(_), _, _) => { /* do nothing since we already have full tx */
-                }
-                (TxNodeInternal::Partial(txouts), _, _) => {
+                (TxNodeInternal::Whole(_), _) => { /* do nothing since we already have full tx */ }
+                (TxNodeInternal::Partial(txouts), _) => {
                     txouts.insert(outpoint.vout, txout);
                 }
             }
@@ -702,15 +699,15 @@ impl<A: Clone + Ord> TxGraph<A> {
 
         for (anchor, txid) in changeset.anchors {
             if self.anchors.insert((anchor.clone(), txid)) {
-                let (_, anchors, _) = self.txs.entry(txid).or_default();
+                let (_, anchors) = self.txs.entry(txid).or_default();
                 anchors.insert(anchor);
             }
         }
 
         for (txid, new_last_seen) in changeset.last_seen {
-            let (_, _, last_seen) = self.txs.entry(txid).or_default();
-            if Some(new_last_seen) > *last_seen {
-                *last_seen = Some(new_last_seen);
+            let last_seen = self.last_seen.entry(txid).or_default();
+            if new_last_seen > *last_seen {
+                *last_seen = new_last_seen;
             }
         }
     }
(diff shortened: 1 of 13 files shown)


crates/chain/tests/test_tx_graph.rs
```diff
@@ -959,10 +959,10 @@ fn test_chain_spends() {
         }))
     );
 
-    // Even if unconfirmed tx has a last_seen of 0, it can still be part of a chain spend.
+    // Even if unconfirmed tx has a last_seen of None, it can still be part of a chain spend.
```
ValuedMammal: Is this a valid assumption?

If instead we were to filter unseen txs in `try_get_chain_position`, it has the same desired effect on canonical txs, but then this assertion no longer holds. In other words, the above comment will say
```
// An unconfirmed and unseen tx cannot be part of a chain spend.
```

Experimenting with that approach in 03d0515ec7a1712c1614d0b0e0947e97463e816d. This might be preferable in that it applies the same reasoning (that an unseen tx does not have a chain position) to `list_chain_spends` and `filter_chain_txouts`.

storopoli:
Isn't `Default` for `Option<T>` `None`? If yes, maybe we can simplify with `..Default::default()`?

crates/chain/src/tx_graph.rs
```diff
@@ -794,7 +794,15 @@ impl<A: Anchor> TxGraph<A> {
...
-            Some(v) => v,
+            Some(val) => {
+                let (_, anchors, last_seen) = val;
+                if anchors.is_empty() && last_seen.is_none() {
+                    // We don't consider unanchored, unseen transactions
+                    // to have a chain position.
+                    return Ok(None);
+                }
+                val
+            }
```
evanlinjin: I like the idea of moving this logic into `try_get_chain_position`. However, for transactions with `last_seen` as `None`, but with an anchor that anchors the transaction in the best chain, wouldn't the transaction be in the best chain?

My understanding of `last_seen` is that a value of `None` signals that it has not been seen in the mempool. However, this fact shouldn't invalidate the fact that the transaction can be confirmed in the best chain?

crates/chain/src/chain_data.rs
```diff
@@ -10,7 +10,7 @@ pub enum ChainPosition<A> {
     /// The chain data is seen as confirmed, and in anchored by `A`.
     Confirmed(A),
     /// The chain data is not confirmed and last seen in the mempool at this timestamp.
-    Unconfirmed(u64),
+    Unconfirmed(Option<u64>),
```
evanlinjin: Would it be fair to say that unconfirmed transactions without a last-seen will not have a `ChainPosition`? Would it make sense to rename `ChainPosition` to `CanonicalPosition`?

crates/chain/src/chain_data.rs
```diff
@@ -58,14 +58,16 @@ pub enum ConfirmationTime {
     /// The transaction is unconfirmed
     Unconfirmed {
         /// The last-seen timestamp in unix seconds.
-        last_seen: u64,
+        last_seen: Option<u64>,
```
evanlinjin: Maybe it's not worth it to put an option here either? Assuming that we agree that unconfirmed txs without a last-seen are not worth returning? The `ConfirmationTime` type is only used in `bdk_wallet` as a field in `LocalOutput` (which is a UTXO). There is no point showing a UTXO that is not part of the canonical chain right?

crates/chain/src/tx_graph.rs
evanlinjin: Should we rename this method to `try_get_canonical_position`?

crates/chain/src/tx_graph.rs
```diff
@@ -965,18 +978,24 @@ impl<A: Anchor> TxGraph<A> {
...
     /// returned item.
     ///
-    /// If the [`ChainOracle`] is infallible, [`list_chain_txs`] can be used instead.
+    /// If the [`ChainOracle`] is infallible, [`canonical_transactions`] can be used instead.
     ///
-    /// [`list_chain_txs`]: Self::list_chain_txs
-    pub fn try_list_chain_txs<'a, C: ChainOracle + 'a>(
+    /// [`full_txs`]: Self::full_txs
+    /// [`canonical_transactions`]: Self::canonical_transactions
+    pub fn try_canonical_transactions<'a, C: ChainOracle + 'a>(
```
evanlinjin: Would `try_list_canonical_transactions` be a better name?

crates/chain/tests/common/tx_template.rs
```diff
@@ -131,9 +131,7 @@ pub fn init_graph<'a, A: Anchor + Clone + 'a>(
         for anchor in tx_tmp.anchors.iter() {
             let _ = graph.insert_anchor(tx.txid(), anchor.clone());
         }
-        if let Some(seen_at) = tx_tmp.last_seen {
-            let _ = graph.insert_seen_at(tx.txid(), seen_at);
-        }
+        let _ = graph.insert_seen_at(tx.txid(), tx_tmp.last_seen.unwrap_or(0));
```
evanlinjin: Why are we force-inserting a `last_seen` when the template `last_seen` is `None`?

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -784,13 +784,21 @@ mod test {
 
     fn get_test_utxos() -> Vec<WeightedUtxo> {
```
evanlinjin: These aren't really useful UTXOs when they are not seen at all right?

evanlinjin:
@LLFourn's [comment](https://github.com/bitcoindevkit/bdk/issues/1396#issuecomment-2036248375)...

> `None` would be in no ones mempool (not even yours). You can have a transaction that is valid but has never left application memory.

summarizes the main usecase of these changes.

With this context in mind, this PR description needs to highlight how we should handle non-broadcasted transactions in `TxGraph`. How should these non-broadcasted transactions influence methods `filter_chain_txouts`/`filter_chain_unspents`-esc methods.

How can the caller list these non-broadcasted transactions? What if the caller would like to include UTXOs of non-broadcasted transactions when listing their UTXO set? What if the caller would NOT like to do so? What if the caller would like to replace non-broadcasted transactions? Currently, we do transaction-precedence based on last_seen. Wouldn't we need something like this for non-broadcasted transactions too?

It seems that many things have not be thought through clearly. Let's have some more concrete discussion about how everything should look after these changes.

crates/chain/tests/common/tx_template.rs
ValuedMammal: This was a fix for `test_tx_graph_conflicts`. Because we're interested in testing tx conflicts, we have to control for the fact that txs without a last_seen would now be filtered out, so we just assume all txs are canonical.

edit: An alternative is to patch up some of the scenarios in `test_tx_graph_conflicts` to ensure every tx template is given some value of last_seen. That might make things clearer.

crates/chain/src/tx_graph.rs
ValuedMammal: I think having a chain position should imply the tx is canonical, so I think these are synonymous.

crates/chain/src/chain_data.rs
ValuedMammal: > There is no point showing a UTXO that is not part of the canonical chain right?

I'm leaning towards no, a utxo owned by the wallet shouldn't qualify as a `LocalOutput` if it doesn't come from a tx in the canonical chain.

ValuedMammal:
While it's possible to store a transaction in `TxGraph` before broadcasting it, I hesitate to advertise this as a feature, because there's not an easy way to make any owned txouts available for coin selection. The design of `TxGraph` requires all txs have a canonical chain position in order to count toward the total balance. You can manipulate the graph into considering a transaction canonical, but without a more robust way to abandon a tx and restore an input's unspent status, the only way to revert it is to do a full wallet rescan.

As an aside, if you're trying to build a chain of dependent transactions without broadcasting them, it should be possible to bypass coin selection by adding any owned txouts as a foreign utxo to a new `TxBuilder`.

evanlinjin:
Another priority review for next week.

notmandatory:
@ValuedMammal if this is already rebased and ready to go please change the status to "ready to review".

evanlinjin:
> This fixes an issue where unbroadcast and otherwise non-canonical transactions were returned from methods 

How would a non-canonical tx be returned by the methods?

crates/wallet/src/wallet/mod.rs
```diff
@@ -2299,6 +2228,16 @@ impl Wallet {
         self.indexed_graph.graph()
     }
 
+    /// Get a mutable reference to the inner [`TxGraph`].
+    ///
+    /// **WARNING**: It's not recommended to make changes manually to the transaction
```
notmandatory: ❓ Is this only being exposed for testing purposes? If so can we avoid move the tests that need it into a test module within this crate so it can be made `pub(crate)` instead? I hate to leave this foot-gun out there in out public API where is can (and likely will) be misused. 

Alternatively can existing "safe" APIs be used in tests instead of adding this function?

notmandatory (commented):
Overall looks good to me, just one question.

crates/wallet/src/wallet/mod.rs
ValuedMammal: I agree I don't like that it is `pub`. What do you think about having methods `insert_anchor` and `insert_seen_at` for Wallet? If not, we can try moving the tests that need it

ValuedMammal:
> > This fixes an issue where unbroadcast and otherwise non-canonical transactions were returned from methods
> 
> How would a non-canonical tx be returned by the methods?

For example in `test_list_owned_txouts` tx3 is registered as spending tx2 at block 1 when tx3 doesn't confirm until block 2 and is never explicitly given a last_seen in the test. Does that seem unexpected?

crates/wallet/src/wallet/mod.rs
notmandatory: ~~If this function is really only needed for tests can you leave it public but feature gate it with `#[cfg(test)]` ?Otherwise if `insert_anchor` and `insert_seen_at` can do their thing safely I like that approach, but again if they're only needed for testing I'd rather they not be in the public API.~~ 

Sorry for my slow reply, I see you already added these functions. They don't look any more dangerous than constructing a bogus wallet `Update` and applying it with `apply_update()`.  I still worry about complicating the wallet API, do you envision any other uses for these new functions except for testing? If not then how do you feel about putting them behind the `#[cfg(test)]` feature flag?

ValuedMammal: I think that means moving a lot of things back to unit tests including everything that uses `get_funded_wallet`. Should I do this in a separate PR?

notmandatory: ~~Can't you leave the tests were they are and just hide those new functions from the public API with the `#[cfg(test)]` ? I haven't tried it but I think it should work that way.~~ Drat, I just learned that `#[cfg(test)]` functions are not available for the integration tests, I guess you were telling me that. I'm ok with leaving the functions in the public API for now then as long as they're reasonably safe to use.

crates/wallet/src/wallet/mod.rs
```diff
@@ -906,6 +877,23 @@ impl Wallet {
         self.stage.append(additions.into());
     }
 
+    /// Inserts an `anchor` for a transaction with the given `txid`.
+    ///
+    /// This stages the changes, you must persist them later.
+    pub fn insert_anchor(&mut self, txid: Txid, anchor: ConfirmationTimeHeightAnchor) {
+        let indexed_graph_changeset = self.indexed_graph.insert_anchor(txid, anchor);
+        self.stage.append(indexed_graph_changeset.into());
+    }
```
evanlinjin: I feel like we should include a warning that the caller is responsible for ensuring the correctness of the information provided. If the caller anchors a tx in a block where the tx does not actually belong in, the canonical history of the wallet may be messed up permanently (since the `TxGraph` is monotone).

crates/chain/src/tx_graph.rs
```diff
@@ -971,10 +981,10 @@ impl<A: Anchor> TxGraph<A> {
     /// If the [`ChainOracle`] implementation (`chain`) fails, an error will be returned with the
     /// returned item.
     ///
-    /// If the [`ChainOracle`] is infallible, [`list_chain_txs`] can be used instead.
+    /// If the [`ChainOracle`] is infallible, [`list_canonical_transactions`] can be used instead.
     ///
-    /// [`list_chain_txs`]: Self::list_chain_txs
-    pub fn try_list_chain_txs<'a, C: ChainOracle + 'a>(
+    /// [`list_canonical_transactions`]: Self::list_canonical_transactions
+    pub fn try_list_canonical_transactions<'a, C: ChainOracle + 'a>(
```
evanlinjin: Many of our methods use `tx` as a name for `transaction`. What do you think about renaming this to `try_list_canonical_txs` for consistency?

crates/wallet/src/wallet/mod.rs
evanlinjin: I think it's okay to leave potentially-dangerous methods in wallet as long as it's documented. Potentially-dangerous also means powerful (sometimes).

evanlinjin:
I feel like we should include a method for returning "unseen and unanchored" transactions. Not sure what to name it, maybe `unseen_txs` is enough? I guess this would be on `TxGraph` and re-exposed on `Wallet`. The intention is to have a method that can be helpful for determining which txs still need to be broadcasted.

crates/wallet/src/wallet/mod.rs
ValuedMammal: Resolved by removing `insert_anchor` and `insert_seen_at`, using `apply_update` instead.

notmandatory (approved):
ACK 42fab2bcc6ba5f0df0385c017851d3589e7575cd

This looks good and all comments addressed.

crates/chain/src/tx_graph.rs
```diff
@@ -110,7 +110,7 @@ use core::{
 #[derive(Clone, Debug, PartialEq)]
 pub struct TxGraph<A = ()> {
     // all transactions that the graph is aware of in format: `(tx_node, tx_anchors, tx_last_seen)`
-    txs: HashMap<Txid, (TxNodeInternal, BTreeSet<A>, u64)>,
+    txs: HashMap<Txid, (TxNodeInternal, BTreeSet<A>, Option<u64>)>,
```
LLFourn: This should not be in this tuple I think. `last_seen` should just been its own `HashMap<Txid, u64>` if the value would be `None` here it would just not have an entry in the map. This is more idomatic and easier to reason about imo.

crates/chain/src/tx_graph.rs
```diff
@@ -258,6 +258,27 @@ impl<A> TxGraph<A> {
             })
     }
 
+    /// All full transactions filtered by a filter map `f`.
+    pub fn full_txs_by<'a, F, O>(&'a self, f: F) -> impl Iterator<Item = O> + 'a
+    where
+        F: FnMut(TxNode<'a, Arc<Transaction>, A>) -> Option<O> + 'a,
+    {
+        self.full_txs().filter_map(f)
+    }
```
LLFourn: This merthod has no reason to exist. Anyone cal just call `full_txs()` and then `.filter_map` on the output.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -1117,7 +1107,58 @@ fn update_last_seen_unconfirmed() {
...
+            .next()
+            .unwrap()
+            .last_seen_unconfirmed
+            .unwrap(),
+        2
+    );
+}
+
+#[test]
+fn list_canonical_txs() {
```
LLFourn: :pick: It would be good if this name could give us more information about what this is testing `transactions_inserted_into_tx_graph_are_not_canonical_until_they_have_an_anchor_in_best_chain`. I know we need to do better on this but we might as well start here.

crates/wallet/src/wallet/mod.rs
```diff
@@ -2321,6 +2250,14 @@ impl Wallet {
         self.indexed_graph.graph()
     }
 
+    /// Iterate over transactions in the wallet that are unseen and unanchored likely
+    /// because they haven't been broadcast.
+    pub fn unbroadcast_transactions(
+        &self,
+    ) -> impl Iterator<Item = TxNode<'_, Arc<Transaction>, ConfirmationTimeHeightAnchor>> {
+        self.as_ref().txs_with_no_anchor_or_last_seen()
```
LLFourn: `as_ref` is very ambiguous here

crates/chain/src/tx_graph.rs
```diff
@@ -560,7 +581,7 @@ impl<A: Clone + Ord> TxGraph<A> {
     pub fn insert_seen_at(&mut self, txid: Txid, seen_at: u64) -> ChangeSet<A> {
         let mut update = Self::default();
         let (_, _, update_last_seen) = update.txs.entry(txid).or_default();
-        *update_last_seen = seen_at;
+        let _ = update_last_seen.insert(seen_at);
```
LLFourn: More idomatic just to assign to `Some(seen_at)`.

LLFourn (dismissed):
Thanks. I've made a few suggestions about how to clean up the internals and tests. API looks good.

crates/wallet/src/wallet/mod.rs
ValuedMammal: using `tx_graph` method on Wallet instead

ValuedMammal:
I'm interested in implementing the suggestion to make last-seen a new field in `TxGraph` that is `HashMap<Txid, u64>` if it can be done in this PR cleanly

notmandatory (approved):
Re ACK af75817d4bedbb5e148812d1073fd0729a23358b

@LLFourn (or anyone else) any other suggestions before we merge this one?

PR #1420, opened 2024-05-02
Bump bdk version to 1.0.0-alpha.10

danielabrozzoni:
### Description

bdk_chain to 0.13.0
bdk_bitcoind_rpc to 0.9.0
bdk_electrum to 0.12.0
bdk_esplora to 0.12.0
bdk_file_store to 0.10.0
bdk_testenv to 0.3.0
bdk_persist to 0.2.0

#### All Submissions:
- [x] I've signed all my commits
- [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
- [x] I ran cargo fmt and cargo clippy before committing

PR #1421, opened 2024-05-02
fix: Cargo clippy lints

danielabrozzoni:
### Description

Caught when trying to release (#1420), clippy failed randomly although it worked on master, this happened because rust 1.78 had just been release and we use clippy stable. IMHO we should pin the clippy version in CI and bump it manually at each new rust release.

#### All Submissions:


notmandatory (approved):
ACK a5fb7fdf505a53b5865ed83d53745e7efa5e7818

PR #1425, opened 2024-05-05
chore: add `print_stdout`/`print_stderr` lints to workspace level

### LINKED ISSUE #1362: Update CI to prevent merging any prints to stdout
```rust
                #[cfg(feature = "std")]
                debug_assert!({
                    println!("txid={} skip={}", txid, skip);
                    true
                });
```

this piece of code in tx_graph.rs prints out stuff when run in dev profile. That's pretty annoying tbh, I don't think it's uncommon for people to use CLIs in the dev profile and this messes up with stdout for these CLIs.

I reckon this has something to do with debugging, but I think there must be better ways to get these kinds of printouts without bothering users?

oleonardolima:
potentially fixes #1362

### Description

It adds both `print_stdout` and `print_stderr` deny level lints on workspace level, but it does allow it on test fns through `clippy.toml` settings, and explicitly allow it on example code. 
 

### Notes to the reviewers
It currently has the setting allowing it on test fns, but open for discussion below.

### Changelog notice

- Add both `print_stdout` and `print_stderr` deny level lints on workspace level

#### All Submissions:

#### New Features:


notmandatory:
This looks like the right way to prevent prints from getting into the code. But do we even want to allow them on tests? The only place I think they should be allowed is in the examples.

oleonardolima:
> This looks like the right way to prevent prints from getting into the code. But do we even want to allow them on tests? The only place I think they should be allowed is in the examples.

I'm fine with it either way, but I don't see a problem with allowing it for tests.

evanlinjin (commented):
Concept ACK

Just needs a rebase! We got rid of the `bdk_persist` crate.

crates/chain/tests/test_local_chain.rs
```diff
@@ -32,6 +32,7 @@ enum ExpectedResult<'a> {
     Err(CannotConnectError),
 }
 
+#[allow(clippy::print_stdout)]
 impl<'a> TestLocalChain<'a> {
     fn run(mut self) {
         println!("[TestLocalChain] test: {}", self.name);
```
notmandatory: 🔧 I'd rather you remove this `println` than add the allow.

clippy.toml
```diff
@@ -1 +1,2 @@
-msrv="1.63.0"
\ No newline at end of file
+msrv="1.63.0"
+allow-print-in-tests=true
```
notmandatory: 🔧 having `println!()` in tests makes the outputs pretty messy so I don't think they should be allowed. Only exception would be for the example crates.

notmandatory (changes requested):
Thanks for cleaning this up, I'd just like to remove the `println!()` from all tests to make `cargo test` output more readable.  If someone really needs to print stuff while debugging tests they can remove or comment out the print lines.

crates/chain/tests/test_local_chain.rs
oleonardolima: Sure, I'll finish removing it in the further commits. There are a bunch of `println` used throughout the code, so would require attention and judgment of what we "might" need/want to keep.

oleonardolima: I did it in the new commits, let me know what you think.

notmandatory (approved):
ACK e063ad89bde62e60b10435260e736c66515bf447

Looks great! much easier to read through the test output with out all the println noise.

PR #1426, opened 2024-05-05
feat: add further `bitcoin::Amount` usage on public APIs

oleonardolima:
builds on top of #1411, further improves #823 

### Description

It further updates and adds the usage of `bitcoin::Amount` instead of `u64`.

### Notes to the reviewers

Open for comments and discussions.

### Changelog notice
- Updated `CreateTxError::FeeTooLow` to use `bitcoin::Amount`.
- Updated `Wallet::calculate_fee()`. to use `bitcoin::Amount`
- Updated `TxBuilder::fee_absolute()`. to use `bitcoin::Amount`.
- Updated `CalculateFeeError::NegativeFee` to use `bitcoin::SignedAmount`.
- Updated `TxGraph::calculate_fee()`. to use `bitcoin::Amount`.
- Updated `PsbUtils::fee_amount()` to use `bitcoin::Amount`.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -95,7 +95,7 @@ use crate::{
 use alloc::collections::vec_deque::VecDeque;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
-use bitcoin::{Amount, OutPoint, Script, Transaction, TxOut, Txid};
+use bitcoin::{Amount, OutPoint, Script, SignedAmount, Transaction, TxOut, Txid};
 use core::fmt::{self, Formatter};
 use core::{
     convert::Infallible,
@@ -182,7 +182,7 @@ pub enum CalculateFeeError {
     /// Missing `TxOut` for one or more of the inputs of the tx
     MissingTxOut(Vec<OutPoint>),
     /// When the transaction is invalid according to the graph it has a negative fee
-    NegativeFee(i64),
+    NegativeFee(SignedAmount),
 }
 
 impl fmt::Display for CalculateFeeError {
@@ -196,7 +196,7 @@ impl fmt::Display for CalculateFeeError {
             CalculateFeeError::NegativeFee(fee) => write!(
                 f,
                 "transaction is invalid according to the graph and has negative fee: {}",
-                fee
+                fee.display_dynamic()
             ),
         }
     }
@@ -307,7 +307,7 @@ impl<A> TxGraph<A> {
         })
     }
 
-    /// Calculates the fee of a given transaction. Returns 0 if `tx` is a coinbase transaction.
+    /// Calculates the fee of a given transaction. Returns [`Amount::ZERO`] if `tx` is a coinbase transaction.
     /// Returns `OK(_)` if we have all the [`TxOut`]s being spent by `tx` in the graph (either as
     /// the full transactions or individual txouts).
     ///
@@ -318,20 +318,20 @@ impl<A> TxGraph<A> {
     /// Note `tx` does not have to be in the graph for this to work.
     ///
     /// [`insert_txout`]: Self::insert_txout
-    pub fn calculate_fee(&self, tx: &Transaction) -> Result<u64, CalculateFeeError> {
+    pub fn calculate_fee(&self, tx: &Transaction) -> Result<Amount, CalculateFeeError> {
         if tx.is_coinbase() {
-            return Ok(0);
+            return Ok(Amount::ZERO);
         }
 
         let (inputs_sum, missing_outputs) = tx.input.iter().fold(
-            (0_i64, Vec::new()),
+            (SignedAmount::ZERO, Vec::new()),
             |(mut sum, mut missing_outpoints), txin| match self.get_txout(txin.previous_output) {
                 None => {
                     missing_outpoints.push(txin.previous_output);
                     (sum, missing_outpoints)
                 }
                 Some(txout) => {
-                    sum += txout.value.to_sat() as i64;
+                    sum += txout.value.to_signed().expect("valid `SignedAmount`");
                     (sum, missing_outpoints)
                 }
             },
@@ -343,15 +343,12 @@ impl<A> TxGraph<A> {
         let outputs_sum = tx
             .output
             .iter()
-            .map(|txout| txout.value.to_sat() as i64)
-            .sum::<i64>();
+            .map(|txout| txout.value.to_signed().expect("valid `SignedAmount`"))
+            .sum::<SignedAmount>();
 
         let fee = inputs_sum - outputs_sum;
-        if fee < 0 {
-            Err(CalculateFeeError::NegativeFee(fee))
-        } else {
-            Ok(fee as u64)
-        }
+        fee.to_unsigned()
+            .map_err(|_| CalculateFeeError::NegativeFee(fee))
     }
 
     /// The transactions spending from this output.
--- a/crates/electrum/tests/test_electrum.rs
+++ b/crates/electrum/tests/test_electrum.rs
@@ -101,7 +101,8 @@ fn scan_detects_confirmed_tx() -> anyhow::Result<()> {
             .fee
             .expect("Fee must exist")
             .abs()
-            .to_sat() as u64;
+            .to_unsigned()
+            .expect("valid `SignedAmount`");
 
         // Check that the calculated fee matches the fee from the transaction data.
         assert_eq!(fee, tx_fee);
--- a/crates/wallet/src/wallet/error.rs
+++ b/crates/wallet/src/wallet/error.rs
@@ -16,7 +16,7 @@ use crate::descriptor::DescriptorError;
 use crate::wallet::coin_selection;
 use crate::{descriptor, KeychainKind};
 use alloc::string::String;
-use bitcoin::{absolute, psbt, OutPoint, Sequence, Txid};
+use bitcoin::{absolute, psbt, Amount, OutPoint, Sequence, Txid};
 use core::fmt;
 
 /// Errors returned by miniscript when updating inconsistent PSBTs
@@ -78,8 +78,8 @@ pub enum CreateTxError {
     },
     /// When bumping a tx the absolute fee requested is lower than replaced tx absolute fee
     FeeTooLow {
-        /// Required fee absolute value (satoshi)
-        required: u64,
+        /// Required fee absolute value [`Amount`]
+        required: Amount,
     },
     /// When bumping a tx the fee rate requested is lower than required
     FeeRateTooLow {
@@ -160,7 +160,7 @@ impl fmt::Display for CreateTxError {
                 )
             }
             CreateTxError::FeeTooLow { required } => {
-                write!(f, "Fee to low: required {} sat", required)
+                write!(f, "Fee to low: required {}", required.display_dynamic())
             }
             CreateTxError::FeeRateTooLow { required } => {
                 write!(
--- a/crates/chain/tests/test_tx_graph.rs
+++ b/crates/chain/tests/test_tx_graph.rs
@@ -10,8 +10,8 @@ use bdk_chain::{
     Anchor, Append, BlockId, ChainOracle, ChainPosition, ConfirmationHeightAnchor,
 };
 use bitcoin::{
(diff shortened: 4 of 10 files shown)


oleonardolima:
@ValuedMammal Please let me know what are your thoughts on these changes.

oleonardolima:
> @ValuedMammal Please let me know what are your thoughts on these changes.

I did remove the updates where it's `pub(crate)` APIs, and kept only the ones for explicitly `pub` APIs.

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -204,16 +204,16 @@ impl<'a, Cs, Ctx> TxBuilder<'a, Cs, Ctx> {
...
+    /// the `FeePolicy` enum will be set by whichever method was called last,
+    /// as the [`FeeRate`] and `FeeAmount` are mutually exclusive.
     ///
     /// Note that this is really a minimum absolute fee -- it's possible to
     /// overshoot it slightly since adding a change output to drain the remaining
     /// excess might not be viable.
-    pub fn fee_absolute(&mut self, fee_amount: u64) -> &mut Self {
-        self.params.fee_policy = Some(FeePolicy::FeeAmount(fee_amount));
+    pub fn fee_absolute(&mut self, fee_amount: Amount) -> &mut Self {
+        self.params.fee_policy = Some(FeePolicy::FeeAmount(fee_amount.to_sat()));
```
ValuedMammal: I thought about it more and I think it would be good for `PreviousFee` and `FeePolicy` to use Amount, sorry for flip-flopping.

crates/bdk/src/wallet/tx_builder.rs
```diff
@@ -204,16 +204,16 @@ impl<'a, Cs, Ctx> TxBuilder<'a, Cs, Ctx> {
...
 
     /// Set an absolute fee
-    /// The fee_absolute method refers to the absolute transaction fee in satoshis (sats).
-    /// If anyone sets both the fee_absolute method and the fee_rate method,
-    /// the FeePolicy enum will be set by whichever method was called last,
-    /// as the FeeRate and FeeAmount are mutually exclusive.
+    /// The fee_absolute method refers to the absolute transaction fee in [`Amount`].
+    /// If anyone sets both the `fee_absolute` method and the `fee_rate` method,
+    /// the `FeePolicy` enum will be set by whichever method was called last,
+    /// as the [`FeeRate`] and `FeeAmount` are mutually exclusive.
```
ValuedMammal: Thanks for fixing this. Now that I'm looking at it, maybe referring to the "FeePolicy enum" is a bit too much detail, but we can state it more generally.

```rust
/// Set an absolute fee [`Amount`].
///
/// The `fee_absolute` method refers to the absolute transaction fee. If anyone sets both
/// this method and the [`fee_rate`] method, the fee policy for building the transaction
/// will be set by whichever method was called last, as the feerate and fee amount are
/// mutually exclusive.
///
/// Note that this is really a minimum absolute fee -- it's possible to
/// overshoot it slightly since adding a change output to drain the remaining
/// excess might not be viable.
///
/// [`fee_rate`]: Self::fee_rate 
```

crates/chain/src/tx_graph.rs
```diff
@@ -182,7 +182,7 @@ pub enum CalculateFeeError {
     /// Missing `TxOut` for one or more of the inputs of the tx
     MissingTxOut(Vec<OutPoint>),
     /// When the transaction is invalid according to the graph it has a negative fee
-    NegativeFee(i64),
+    NegativeFee(SignedAmount),
```
ValuedMammal, line 185: The `Display` impl for both `Amount` and `SignedAmount` will be denominated in BTC. Is that what we want?

https://docs.rs/bitcoin/0.31.2/bitcoin/amount/struct.SignedAmount.html#impl-Display-for-SignedAmount

crates/chain/src/tx_graph.rs
```diff
@@ -318,20 +318,20 @@ impl<A> TxGraph<A> {
...
-    pub fn calculate_fee(&self, tx: &Transaction) -> Result<u64, CalculateFeeError> {
+    pub fn calculate_fee(&self, tx: &Transaction) -> Result<Amount, CalculateFeeError> {
         if tx.is_coinbase() {
-            return Ok(0);
+            return Ok(Amount::ZERO);
         }
 
         let (inputs_sum, missing_outputs) = tx.input.iter().fold(
-            (0_i64, Vec::new()),
+            (SignedAmount::ZERO, Vec::new()),
```
ValuedMammal, line 327: This could work, but I think just create an `Amount` from the return value and leave the rest alone.

crates/bdk/src/wallet/mod.rs
```diff
@@ -1421,8 +1420,9 @@ impl Wallet {
         }
 
         // we keep it as a float while we accumulate it, and only round it at the end
-        let mut outgoing: u64 = 0;
-        let mut received: u64 = 0;
+        // (it's internally implemented as a float as Amount(u64))
+        let mut outgoing = Amount::ZERO;
+        let mut received = Amount::ZERO;
```
ValuedMammal: The original comment must be wrong, because a `u64` is definitely not a float.

ValuedMammal:
Here's how I would approach this PR with fewer lines touched ValuedMammal/bdk@0c72958e388e060a1ad6db46e7f19192938b8f1f

However I think it's worth having a discussion about where we can continue to get benefits from the Amount type so I opened an issue #1432 

--

edit: @oleonardolima Just checking, is the plan to go ahead and complete #1432 in this PR?

crates/bdk/src/wallet/tx_builder.rs
storopoli: I don't agree. I always think that these links are useful.
They are "hover" only and avoids you having to go through the codebase, both in an IDE/LSP and also in the Docs.

I would agree with you if only the thing we are referring is not public, which is not the case here

https://github.com/bitcoindevkit/bdk/blob/66abc73c3dd0e3fb1f18586f6beca2e8d457be83/crates/bdk/src/wallet/tx_builder.rs#L166-L170

oleonardolima: > Thanks for fixing this. Now that I'm looking at it, maybe referring to the "FeePolicy enum" is a bit too much detail, but we can state it more generally.

I agree with Jose's point, and that's what I had in mind and also why I updated it.

It gives the ergonomic for the developer to know exactly which type is being used, and to navigate easily on the code, at least I use it a lot with `rust-analyzer` (even navigating into the dependencies and so on).

> [...]
> I would agree with you if only the thing we are referring is not public, which is not the case here

Exactly, I updated it only for the ones that are explicitly public, the compiler does not even allow you to reference non-public APIs.

crates/chain/src/tx_graph.rs
oleonardolima: That's a good catch, thanks!

Although bitcoin::Amount is generally sats, they are using the Denomination::Bitcoin on Display, I don't think that's what we want (at least not what I had in mind).

crates/chain/src/tx_graph.rs
oleonardolima: It's "safer", I'll update it. If we decide to use bitcoin::Amount everywhere, we can update it again.

oleonardolima:
> [...]
> 
> --
> 
> edit: @oleonardolima Just checking, is the plan to go ahead and complete #1432 in this PR?

I think I can continue the work on non-public code (everywhere) on another PR, and keep this one only for the public APIs that remained from #1411. wdyt @ValuedMammal ?

Also, I was waiting if some new discussion would occur on #1432 before working on it :)

crates/bdk/src/wallet/tx_builder.rs
ValuedMammal: I based this https://github.com/bitcoindevkit/bdk/pull/1426#discussion_r1594166489 suggestion on the observation that `FeePolicy` won't be visible to someone outside the crate reading the docs for example on docs.rs, which is why I think it doesn't belong here.

oleonardolima: Good point, but still for us and other BDK developers I think it helps to navigate the codebase through IDE tools 🤔.

crates/wallet/src/wallet/error.rs
```diff
@@ -78,8 +78,8 @@ pub enum CreateTxError {
     },
     /// When bumping a tx the absolute fee requested is lower than replaced tx absolute fee
     FeeTooLow {
-        /// Required fee absolute value (satoshi)
-        required: u64,
+        /// Required fee absolute value [`Amount`]
+        required: Amount,
```
ValuedMammal, line 82: I think we should fix the displayed error message for `FeeTooLow`, maybe by using `display_dynamic` again

ValuedMammal (commented):
Thanks for working on this. There's a small typo in the changelog notice - `FeePolicy::fee_absolute` should say `TxBuilder::fee_absolute`.

oleonardolima:
> Thanks for working on this. There's a small typo in the changelog notice - `FeePolicy::fee_absolute` should say `TxBuilder::fee_absolute`.

Oh, thanks! Sorry for the oversight, fixed it.

crates/wallet/src/wallet/error.rs
oleonardolima: Yes, updated.
I did another pass on the codebase as well and currently didn't find other spots that would need this kind of fix.

crates/chain/src/tx_graph.rs
```diff
@@ -343,14 +343,13 @@ impl<A> TxGraph<A> {
...
 
         let fee = inputs_sum - outputs_sum;
-        if fee < 0 {
-            Err(CalculateFeeError::NegativeFee(fee))
-        } else {
-            Ok(fee as u64)
+        match fee.to_unsigned() {
+            Ok(fee) => Ok(fee),
+            Err(_) => Err(CalculateFeeError::NegativeFee(fee)),
         }
```
notmandatory: This can be done more concisely like this: 
```suggestion
        fee.to_unsigned()
            .map_err(|_| CalculateFeeError::NegativeFee(fee))
```

notmandatory (approved):
ACK 5f2d6cbf08106028558dda570a28de50690496d5

I made one small suggestion and I think you need to do a rebase. Otherwise looks ready to merge to me.

crates/chain/src/tx_graph.rs
oleonardolima: Thanks, that's better, updated

oleonardolima:
> ACK [5f2d6cb](https://github.com/bitcoindevkit/bdk/commit/5f2d6cbf08106028558dda570a28de50690496d5)
> 
> I made one small suggestion and I think you need to do a rebase. Otherwise looks ready to merge to me.

**EDIT:** I applied the suggestion and rebased it on top of master at [a03949a](https://github.com/bitcoindevkit/bdk/pull/1426/commits/a03949adb0c38ba4c2f44b497871070c1a2170d2), should be ready to go now.

crates/electrum/tests/test_electrum.rs
```diff
@@ -101,7 +101,8 @@ fn scan_detects_confirmed_tx() -> anyhow::Result<()> {
             .fee
             .expect("Fee must exist")
             .abs()
-            .to_sat() as u64;
+            .to_unsigned()
+            .expect("valid `SignedAmount");
```
storopoli: missing backtick, or remove the dangling one.

storopoli (changes requested):

crates/electrum/tests/test_electrum.rs
oleonardolima: done at [a03949a](https://github.com/bitcoindevkit/bdk/pull/1426/commits/a03949adb0c38ba4c2f44b497871070c1a2170d2)

storopoli (approved):
ACK a03949adb0c38ba4c2f44b497871070c1a2170d2

crates/electrum/tests/test_electrum.rs
ValuedMammal: but you called `to_unsigned` so the expect message is wrong. oh well

oleonardolima: > but you called `to_unsigned` so the expected message is wrong. oh well

Oh, sorry for the oversight, I fixed it in #1458 🤦‍♂️

PR #1428, opened 2024-05-06
Further work on #1203

evanlinjin:
### Description

This PR builds on top of #1203 to complete it. Ideally, these commits will be included in #1203 so I will not include the "Changelog notice" section.

**What this PR fixes:**

* Simplifies the `Append::append` implementation for `keychain::ChangeSet`. (Conversation: https://github.com/bitcoindevkit/bdk/pull/1203#discussion_r1510200792)
* Fixes changesets no longer being monotone (when applied to `KeychainTxOutIndex`). We rank keychain variants based on `Ord` (earliest variants have higher rank) and return the highest ranked one for our methods. (Conversation: https://github.com/bitcoindevkit/bdk/pull/1203#discussion_r1590010052)

**Further benefits:**

This properly allows multiple keychains to share the same descriptor, hence removing the need of #1390 (although we should probably test this).

#### All Submissions:

#### Bugfixes:

~* [x] This pull request breaks the existing API~

### DIFF
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -49,23 +49,26 @@ impl<K: Ord> Append for ChangeSet<K> {
     ///
     /// For each `last_revealed` in the given [`ChangeSet`]:
     /// If the keychain already exists, increase the index when the other's index > self's index.
-    fn append(&mut self, mut other: Self) {
-        for (keychain, descriptor) in &mut self.keychains_added {
-            if let Some(other_descriptor) = other.keychains_added.remove(keychain) {
-                *descriptor = other_descriptor;
-            }
-        }
-
-        for (descriptor_id, index) in &mut self.last_revealed {
-            if let Some(other_index) = other.last_revealed.remove(descriptor_id) {
-                *index = other_index.max(*index);
-            }
-        }
-
+    fn append(&mut self, other: Self) {
         // We use `extend` instead of `BTreeMap::append` due to performance issues with `append`.
         // Refer to https://github.com/rust-lang/rust/issues/34666#issuecomment-675658420
         self.keychains_added.extend(other.keychains_added);
-        self.last_revealed.extend(other.last_revealed);
+
+        // for `last_revealed`, entries of `other` will take precedence ONLY if it is greater than
+        // what was originally in `self`.
+        for (desc_id, index) in other.last_revealed {
+            use crate::collections::btree_map::Entry;
+            match self.last_revealed.entry(desc_id) {
+                Entry::Vacant(entry) => {
+                    entry.insert(index);
+                }
+                Entry::Occupied(mut entry) => {
+                    if *entry.get() < index {
+                        entry.insert(index);
+                    }
+                }
+            }
+        }
     }
 
     /// Returns whether the changeset are empty.
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -13,6 +13,71 @@ use core::{
 
 use crate::Append;
 
+
+/// Represents updates to the derivation index of a [`KeychainTxOutIndex`].
+/// It maps each keychain `K` to its last revealed index.
+///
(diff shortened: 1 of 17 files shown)


crates/chain/src/keychain/txout_index.rs
```diff
@@ -172,17 +186,17 @@ const DEFAULT_LOOKAHEAD: u32 = 25;
...
 /// [`all_unbounded_spk_iters`]: KeychainTxOutIndex::all_unbounded_spk_iters
-// Under the hood, the KeychainTxOutIndex uses a SpkTxOutIndex that keeps track of spks, indexed by
-// descriptors. Users then assign or unassign keychains to those descriptors. It's important to
-// note that descriptors, once added, never get removed from the SpkTxOutIndex; this is useful in
-// case a user unassigns a keychain from a descriptor and after some time assigns it again.
 #[derive(Clone, Debug)]
 pub struct KeychainTxOutIndex<K> {
     inner: SpkTxOutIndex<(DescriptorId, u32)>,
     // keychain -> (descriptor, descriptor id) map
     keychains_to_descriptors: BTreeMap<K, (DescriptorId, Descriptor<DescriptorPublicKey>)>,
```
LLFourn: why do we have this map from `K` to (`DescriptorId, Descriptor`). Shouldn't it just be the `DescriptorId`? With the `DescriptorId` you can lookup the `Descriptor`.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -208,11 +222,9 @@ impl<K: Clone + Ord + Debug> Indexer for KeychainTxOutIndex<K> {
             Some((descriptor_id, index)) => {
                 // We want to reveal spks for descriptors that aren't tracked by any keychain, and
                 // so we call reveal with descriptor_id
-                if let Some((_, changeset)) = self.reveal_to_target_with_id(descriptor_id, index) {
-                    changeset
-                } else {
-                    super::ChangeSet::default()
-                }
+                let (_, changeset) = self.reveal_to_target_with_id(descriptor_id, index)
+                    .expect("descriptors are added in a monotone manner, there cannot be a descriptor id with no corresponding descriptor");
```
LLFourn, line 222: 90% of the code in `reveal_to_target_with_id` is about producing the iterator which you ignore here and then panic because producing it is a fallible operation since the descriptor might not exist. Instead just call an infallible function that increases a descriptor ids index and returns a changeset (it's only 2 or 3 lines of code). The descriptor for the descriptor id doesn't have to exist to do this.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -444,45 +458,42 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-        let old_keychain_opt = self
-            .descriptor_ids_to_keychain
-            .insert(descriptor_id, (keychain.clone(), descriptor.clone()));
-
-        // If `keychain` already had a `descriptor` associated, different from the `descriptor`
-        // passed in, we remove it from the descriptor -> keychain map
-        if let Some((old_desc_id, _)) = old_descriptor_opt {
-            if old_desc_id != descriptor_id {
-                self.descriptor_ids_to_keychain.remove(&old_desc_id);
+            .insert(keychain.clone(), (desc_id, descriptor.clone()));
```
LLFourn: The claim of the commit message was that the structure is now fully monotone. But it's not, this function blindly changes the descriptor associated with a keychain. The keychian <-> descriptor relationship will depend on the order of the changessets/calls to this.

LLFourn (commented):
I reviewed this taking the premise that making the structure fully monotone is necessary (I'm not sure it is and I'm not sure this commit does that).

evanlinjin:
Closed since #1203 is merged.

crates/chain/src/keychain/txout_index.rs
evanlinjin: I think 90% is a bit of an overstatement? We still need the `next_reveal_index` and `target_index` anyway (since we need to know which spks to reveal to reveal them). The actual logic that creates the returned spk iters is two lines. However, I would simplify this method by taking in a descriptor directly and not returning an option.

PR #1430, opened 2024-05-07
ci: Pin clippy to rust 1.78.0

ValuedMammal:
This PR pins clippy check in CI to the rust 1.78 toolchain, which prevents new lints in stable releases from interrupting the usual workflow. Because rust versions are released on a predictable schedule, we can revisit this setting in the future as needed (say 3-6 months).

#### All Submissions:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -118,7 +118,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions-rs/toolchain@v1
         with:
-            toolchain: stable
+            toolchain: 1.78.0
             components: clippy
             override: true
       - name: Rust Cache


.github/workflows/cont_integration.yml
```diff
@@ -118,7 +118,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions-rs/toolchain@v1
         with:
-            toolchain: stable
+            toolchain: 1.78.0
```
storopoli, line 121:
```suggestion
            toolchain: 1.63.0
```

Why not 1.63, our MSRV?

https://github.com/bitcoindevkit/bdk/blob/dcd2d4741d46c30a930b6f673cf12016a65edc54/clippy.toml#L1C7-L1C13

cc @danielabrozzoni this might solve our clippy problems.

ValuedMammal: We want the latest lints that are compatible with MSRV, we just don't want them changing unexpectedly.

danielabrozzoni (approved):
ACK de2763a4b88e65b1c67a06148b4f9e012e561c5a

.github/workflows/cont_integration.yml
storopoli: Just saw the discord message now. Makes sense.

storopoli (approved):
ACK de2763a4b88e65b1c67a06148b4f9e012e561c5a

apoelstra:
In rust-bitcoin we have started pinning our nightly compiler (which is where we run clippy) and we have a cronjob that runs every so often (was daily, but we reduced it to twice a week, and realistically it'd be fine if it were twice a year) which opens a PR to bump the pin.

If the PR passes CI we know that we can accept it and increase our pin. If it fails CI then we can see that there's an upcoming lint failure and we can address it at our own pace.

notmandatory (approved):
ACK de2763a4b88e65b1c67a06148b4f9e012e561c5a

oleonardolima (approved):
ACK de2763a4b88e65b1c67a06148b4f9e012e561c5a

danielabrozzoni:
> In rust-bitcoin we have started pinning our nightly compiler (which is where we run clippy) and we have a cronjob that runs every so often (was daily, but we reduced it to twice a week, and realistically it'd be fine if it were twice a year) which opens a PR to bump the pin.

@apoelstra thanks for sharing! Definitely something we should do in the future, we did forget to bump the pin on CI a few times. Opened a new good first issue: https://github.com/bitcoindevkit/bdk/issues/1433

Issue #1431, opened 2024-05-07
Create nightly job to bump clippy rust version


notmandatory:
We should do what the `rust-bitcoin` team does to have CI check weekly if the latest nightly rust version can run `clippy` without errors. If `clippy` passes then CI should open a PR to bump the rust version for `clippy`. If not we can create an issue to fix later.

"In rust-bitcoin we have started pinning our nightly compiler (which is where we run clippy) and we have a cronjob that runs every so often (was daily, but we reduced it to twice a week, and realistically it'd be fine if it were twice a year) which opens a PR to bump the pin.

If the PR passes CI we know that we can accept it and increase our pin. If it fails CI then we can see that there's an upcoming lint failure and we can address it at our own pace."

_Originally posted by @apoelstra in https://github.com/bitcoindevkit/bdk/issues/1430#issuecomment-2098609403_

PR #1438, opened 2024-05-10
Various improvements not included in #1203

evanlinjin:
### Description

Adds various improvements to the work of #1203. These were missed out while cherry-picking, or review comments left in #1428 that were forgotten:

* Change `keychains_to_descriptors` to `keychains_to_descriptor_ids` which simplifies the field. This was mentioned [here](https://github.com/bitcoindevkit/bdk/pull/1428#discussion_r1591795543) and included in #1428, but an older commit was cherry-picked.
* Rename `KeychainTxOutIndex` field `descriptor_ids_to_keychain_set` to `descriptor_ids_to_keychains` was missed out as an older commit was cherry-picked. This change to naming shows the direct relationship between `keychains_to_desriptor_ids` and `descriptor_ids_to_keychains` (one is directly a reverse lookup of the other).
* Change `reveal_to_target_with_id` to `reveal_to_target_with_descriptor`, reasoning mentioned [here](https://github.com/bitcoindevkit/bdk/pull/1428#discussion_r1596527461).

In addition to this, I changed the output signature of `reveal_to_target`, `reveal_next_spk` and `next_unused_spk` methods to return `(Option<spk(s)>, changeset)`, whereas previously it was `Option<(spk(s), changeset)>`. This makes the API more consistent as the `ChangeSet` is always returned, and `reveal_to_target` and `unbounded_spk_iter`-esc methods all return `Option<SpkIterator>` (which we can `.flatten()`).

### Notes to the reviewers

Not all changes in this PR are Changelog-worthy. I.e. renaming of internal variables to increase readability, changing code comments, refactoring private methods - are all excluded from the changelog.

### Changelog notice

* Change `KeychainTxOutIndex` methods to always return a changeset. This makes the API more consistent.

#### All Submissions:

### DIFF
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -207,14 +207,16 @@ const DEFAULT_LOOKAHEAD: u32 = 25;
 #[derive(Clone, Debug)]
 pub struct KeychainTxOutIndex<K> {
     inner: SpkTxOutIndex<(DescriptorId, u32)>,
-    // keychain -> (descriptor, descriptor id) map
-    keychains_to_descriptors: BTreeMap<K, (DescriptorId, Descriptor<DescriptorPublicKey>)>,
-    // descriptor id -> keychain set
-    // Because different keychains can have the same descriptor, we rank keychains by `Ord` so that
-    // that the first keychain variant (according to `Ord`) has the highest rank. When associated
-    // data (such as spks, outpoints) are returned with a keychain, we return the highest-ranked
-    // keychain with it.
-    descriptor_ids_to_keychain_set: HashMap<DescriptorId, BTreeSet<K>>,
+    // keychain -> descriptor_id map
+    keychains_to_descriptor_ids: BTreeMap<K, DescriptorId>,
+    // descriptor_id -> keychain set
+    // This is a reverse map of `keychains_to_descriptors`. Although there is only one descriptor
+    // per keychain, different keychains can refer to the same descriptor, therefore we have a set
+    // of keychains per descriptor. When associated data (such as spks, outpoints) are returned with
+    // a keychain, we return it with the highest-ranked keychain with it. We rank keychains by
+    // `Ord`, therefore the keychain set is a `BTreeSet`. The earliest keychain variant (according
+    // to `Ord`) has precedence.
+    descriptor_ids_to_keychains: HashMap<DescriptorId, BTreeSet<K>>,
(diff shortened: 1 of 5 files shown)


ValuedMammal:
In terms of super simple names you could have `descriptor_ids_to_descriptors` be named `descriptors` and `descriptor_ids_to_keychains` be `keychains`. With a method like `get`, it's still clear what the key-value relationship is.

ValuedMammal:
I got tripped up at this comment. I think what's happening is we're removing this keychain from the old descriptor's set

https://github.com/bitcoindevkit/bdk/blob/78e32645957ad2c65e4b0d2da77cc996d4c0318c/crates/chain/src/keychain/txout_index.rs#L509-L516

ValuedMammal:
Maybe more of a philosophical point, but I think an argument can be made that methods which previously returned `K`, such as `index_of_spk` et al, can return the descriptor id and let the caller match the id with a pre-determined label. If a user is interested in getting the set of keychains marking a descriptor, there can be a method for that called `keychains_of_descriptor{_id}`

notmandatory:
Can you also add changes from #1341?

evanlinjin:
> Can you also add changes from #1341?
> 

Good idea

ValuedMammal:
Is it worth revisiting a discussion of returning `Option` vs empty iterators? I don't see an issue with returning bogus values (given the wrong inputs) if it means reducing the syntactic overhead in function signatures. We could expand this concept by returning `ScriptBuf::new` when we don't have an spk to reveal, and returning an empty `SpkIterator` (which we'd have to implement) for `reveal_to_target`.

ValuedMammal:
Reviewed up through 78e32645. For a minute I thought all of #1341 was done but it looks like there's a few more places like `spk_at_index`, `revealed_spks`, `unused_spks`.

documentation nit: There's a missing backtick in the docs for keychain `ChangeSet`, which is strange because I think this was fixed in an older version.

evanlinjin:
@LagginTimes can you complete this PR for me?

* Rename field `descriptor_ids_to_descriptors` to `descriptors`.
* Rename field `descriptor_ids_to_keychains` to `keychains`.
* Fix this comment: https://github.com/bitcoindevkit/bdk/pull/1438#issuecomment-2113478065 to that of the version in https://github.com/bitcoindevkit/bdk/pull/1428
* Also do #1431 in this PR too!

evanlinjin:
Replaced by #1451

PR #1441, opened 2024-05-13
Remove duplicated InsufficientFunds error member

### LINKED ISSUE #1440: InsufficientFunds is duplicated within coin_selection::Error and CreateTxError enums
**Describe the bug**  
Both `coin_selection::Error` and `CreateTxError` enums have an InsufficientFunds member. It is quite confusion prone

**Expected behavior**  
We shouldn't have duplicated enums

**Proposed solution**  
keep only `coin_selection::Error`'s member

e1a0a0ea:
closes #1440 

### Description

- Replace `CreateTxError::InsufficientFunds` use by `coin_selection::Error::InsufficientFunds`
- Remove `InsufficientFunds` member from `CreateTxError` enum
- Rename `coin_selection::Error` to `coin_selection::CoinSelectionError`

### Notes to the reviewers

- We could also keep both members but rename one of them to avoid confusion

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/error.rs
+++ b/crates/wallet/src/wallet/error.rs
@@ -94,13 +94,6 @@ pub enum CreateTxError {
     ChangePolicyDescriptor,
     /// There was an error with coin selection
     CoinSelection(coin_selection::Error),
-    /// Wallet's UTXO set is not enough to cover recipient's requested plus fee
-    InsufficientFunds {
-        /// Sats needed for some transaction
-        needed: u64,
-        /// Sats available for spending
-        available: u64,
-    },
     /// Cannot build a tx without recipients
     NoRecipients,
     /// Partially signed bitcoin transaction error
@@ -184,13 +177,6 @@ impl fmt::Display for CreateTxError {
                 )
             }
             CreateTxError::CoinSelection(e) => e.fmt(f),
-            CreateTxError::InsufficientFunds { needed, available } => {
-                write!(
-                    f,
-                    "Insufficient funds: {} sat available of {} sat needed",
-                    available, needed
-                )
-            }
             CreateTxError::NoRecipients => {
                 write!(f, "Cannot build tx without recipients")
             }
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -73,6 +73,8 @@ use crate::types::*;
 use crate::wallet::coin_selection::Excess::{Change, NoChange};
 use crate::wallet::error::{BuildFeeBumpError, CreateTxError, MiniscriptPsbtError};
 
+use self::coin_selection::Error;
+
 const COINBASE_MATURITY: u32 = 100;
 
 /// A Bitcoin wallet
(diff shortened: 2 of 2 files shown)


notmandatory:
Please also setup commit signing, see: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

matthiasdebernardini:
Maybe this belongs here better, basically, can this be disabled in anything but Network::Bitcoin?

https://github.com/bitcoindevkit/bdk/issues/1440#issuecomment-2108275322

oleonardolima (commented):
Concept ACK, I do agree with Value's and Steve's comments above.

oleonardolima:
> Maybe this belongs here better, basically, can this be disabled in anything but Network::Bitcoin?
> 
> [#1440 (comment)](https://github.com/bitcoindevkit/bdk/issues/1440#issuecomment-2108275322)

I personally don't think it's a good idea to "error-gate" based on the bitcoin::Network type, I would expect all behaviors that happen on mainnet (production per se) to happen on others, and even some nasty behaviors we may catch while testing on testnet/signet being prevented to land on mainnet code 😅.

matthiasdebernardini:
That's a good point, @oleonardolima maybe the juice is not worth the squeeze on making this change.

oleonardolima (commented):
Looking good! I left some comments, and I also think that it could be squashed into a single commit.

crates/bdk/tests/wallet.rs
```diff
@@ -4,7 +4,7 @@ use assert_matches::assert_matches;
 use bdk::descriptor::{calc_checksum, IntoWalletDescriptor};
 use bdk::psbt::PsbtUtils;
 use bdk::signer::{SignOptions, SignerError};
-use bdk::wallet::coin_selection::{self, LargestFirstCoinSelection};
+use bdk::wallet::coin_selection::{self, CoinSelectionError, LargestFirstCoinSelection};
```
oleonardolima: @e1a0a0ea Does this change still applies?

crates/bdk/src/wallet/mod.rs
```diff
@@ -1662,10 +1664,12 @@ impl Wallet {
                     change_fee,
                 } = excess
                 {
-                    return Err(CreateTxError::InsufficientFunds {
-                        needed: *dust_threshold,
-                        available: remaining_amount.saturating_sub(*change_fee),
-                    });
+                    return Err(CreateTxError::CoinSelection(
+                        Error::InsufficientFunds {
```
oleonardolima: @e1a0a0ea You could probably use a qualified version: `coin_selection::Error::InsufficientFunds` instead 🤔

oleonardolima (commented):
Looking good! I left some comments, and I also think that it could be squashed into a single commit.

crates/bdk/src/wallet/mod.rs
ValuedMammal: Or just import the `InsufficientFunds` variant itself.

notmandatory:
Please rebase again and fixup commits into one commit then this should be ready to merge.

crates/bdk/src/wallet/mod.rs
evanlinjin: I'm not a fan of importing the enum variant (can become confusing imo). I prefer the more verbose, but clearer option.

evanlinjin (approved):
ACK 29c8a00b435bdf2fa1d7c707ad09c1a47edc07b3

notmandatory (approved):
ACK 29c8a00b435bdf2fa1d7c707ad09c1a47edc07b3

PR #1400, opened 2024-04-10
Test the bitcoin `v0.32.0-rc1` release

tcharding:
Please excuse the noise if this is not of interest to you guys right now. No action required from the `bdk` team - this is just informational and so the `rust-bitcoin` team can see that bitcoin rc1 release is ok. 

Test the latest bitcoin release candidate. Includes an unreleased patched version of `rust-miniscript` (`master` branch + depend on bitcoin rc release).

Notes:

- This wouldn't have been too hard if I knew my way around the codebase and didn't try to rush. The new `miniscript::Threshold` usage took me down a dark path for a while :) 
- Leaves a few calls to `expect` that require more invasive error handling improvements
- Uses the ugly `weight.to_wu() as usize` instead of using the new `Weight` type more thoroughly
- Uses `network.into()` liberally instead of using the new `NetworkKind` more thoroughly
- The `Sequence` / `RelLockTime` stuff is a bit rough too

But I believe this proves the rc1 release is pretty good as it is.

`cargo test  --color=always --workspace --all-features --all-targets` passes all tests on my machine.

PR #1424, opened 2024-05-03
refactor(wallet): Use Psbt::sighash_ecdsa for computing sighashes

ValuedMammal:
This PR does some cleanup of the `bdk_wallet` signer module most notably by removing the internal trait `ComputeSighash` and replacing old code for computing the sighash (for legacy and segwit context) with a single method [`Psbt::sighash_ecdsa`](https://docs.rs/bitcoin/0.31.2/bitcoin/psbt/struct.Psbt.html#method.sighash_ecdsa). The logic for computing the taproot sighash is unchanged and extracted to a new helper function `compute_tap_sighash`.

- [x] Unimplement `ComputeSighash`
- [x] Try de-duplicating code by using `Psbt::sighash_ecdsa`. see https://github.com/bitcoindevkit/bdk/pull/1023#discussion_r1263140218
- Not done in this PR: Consider removing unused `SignerError` variants

fixes bitcoindevkit/bdk#1038

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #1448, opened 2024-05-22
bump(deps): upgrade rust `bitcoin` to `0.32.0`, `miniscript` to `0.12.0` and others

### LINKED ISSUE #1422: Upgrade rust-bitcoin to 0.32
**Describe the enhancement**  

The LDK team would like to use `rust-bitcoin` version 0.32 so BDK should update too. To do so we also need to update the following crates to the versions that depend on `rust-bitcoin` 0.32:

- [x] [rust-miniscript](https://github.com/rust-bitcoin/rust-miniscript)
- [x] [rust-bitcoincore-rpc](https://github.com/rust-bitcoin/rust-bitcoincore-rpc)
- [x] [bitcoind](https://github.com/rust-bitcoin/bitcoind)
- [x] [electrsd](https://github.com/rcasatta/electrsd)
- [x] [rust-electrum-client](https://github.com/bitcoindevkit/rust-electrum-client)
- [x] [rust-esplora-client](https://github.com/bitcoindevkit/rust-esplora-client)
- [x] [rust-hwi](https://github.com/bitcoindevkit/rust-hwi)

**Use case**  

Updating to `rust-bitcoin` 0.32 removes the need for a specific `no-std` feature, it will be implied by not enabling the `std` feature. It also includes a new io crate.

**Additional context**

https://discord.com/channels/753336465005608961/753367451319926827/1228314439133171795

oleonardolima:
fixes bitcoindevkit/bdk#1422

### Description

This PR focuses on upgrading the `rust-bitcoin` and `miniscript` versions, to `0.32.0` and `0.12.0`, respectively. It also bumps the versions of other BDK ecosystem crates that also rely on both `rust-bitcoin` and `miniscript`, being: 

- electrum-client https://github.com/bitcoindevkit/rust-electrum-client/pull/133
- esplora-client https://github.com/bitcoindevkit/rust-esplora-client/pull/85
- hwi https://github.com/bitcoindevkit/rust-hwi/pull/99

<ins>I structured the PR in multiple commits, with closed scope, being one for each BDK crate being upgraded, and one for each kind of fix and upgrade required, it seems like a lot of commits (**that should be squashed before merging**), but I think it'll make it easier during review phase.</ins>

In summary I can already mention some of the main changes:
- using `compute_txid()` instead of deprecated `txid()`
- using `minimal_non_dust()` instead of `dust_value()`
- using the renamed `signature` and `sighash_type` fields 
- using proper `sighash::P2wpkhError`,  `sighash::TaprootError` instead of older `sighash::Error`
- conversion from `Network` to new expected `NetworkKind` bitcoindevkit/bdk_wallet#94 
- conversion from the new `Weight` type to current expected `usize` bitcoindevkit/bdk#1466 
- using `.into()` to convert from AbsLockTime and `RelLockTime` to `absolute::LockTime` and `relative::LockTime`
- using Message::from_digest() instead of relying on deprecated `ThirtyTwoByteHash` trait.
- updating the miniscript policy and dsl to proper expect and handle new `Threshold` type, instead of the previous two parameters.

### Notes to the reviewers
<ins>Again, I structured the PR in multiple commits, with closed scope, being one for each BDK crate being upgraded, and one for each kind of fix and upgrade required, it seems like a lot of commits (**that should be squashed before merging**), but I think it'll make it easier during review phase.</ins>

It should definitely be carefully reviewed, especially the last commits for the wallet crate scope, the ones with the semantic `fix(wallet)`.

I would also appreciate if @tcharding reviewed it, as he gave a try in the past (#1400 ), and I relied on some of it for the  policy part of it, other rust-bitcoin maintainers reviews are a definitely welcome 😁 

### Changelog notice
- Use `compute_txid()` instead of deprecated `txid()`
- Use `minimal_non_dust()` instead of `dust_value()`
- Use `signature` and `sighash_type` fields, instead of previous `sig` and `hash_type` 
- Use `sighash::P2wpkhError`,  and `sighash::TaprootError` instead of older `sighash::Error`
- Converts from `Network` to `NetworkKind`, where expected
- Converts from `Weight` type to current used `usize`
- Use `.into()` to convert from `AbsLockTime` and `RelLockTime` to `absolute::LockTime` and `relative::LockTime`
- Remove use of  deprecated `ThirtyTwoByteHash` trait, use `Message::from_digest()` 
- Update the miniscript policy and dsl macros to proper expect and handle new `Threshold` type, instead of the previous two parameters.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/descriptor_ext.rs
+++ b/crates/chain/src/descriptor_ext.rs
@@ -31,7 +31,7 @@ impl DescriptorExt for Descriptor<DescriptorPublicKey> {
         self.at_derivation_index(0)
             .expect("descriptor can't have hardened derivation")
             .script_pubkey()
-            .dust_value()
+            .minimal_non_dust()
             .to_sat()
     }
 
--- a/crates/esplora/Cargo.toml
+++ b/crates/esplora/Cargo.toml
@@ -13,13 +13,12 @@ readme = "README.md"
 
 [dependencies]
 bdk_chain = { path = "../chain", version = "0.15.0", default-features = false }
-esplora-client = { version = "0.7.0", default-features = false }
+esplora-client = { version = "0.8.0", default-features = false }
 async-trait = { version = "0.1.66", optional = true }
 futures = { version = "0.3.26", optional = true }
 
-# use these dependencies if you need to enable their /no-std features
-bitcoin = { version = "0.31.0", optional = true, default-features = false }
-miniscript = { version = "11.0.0", optional = true, default-features = false }
+bitcoin = { version = "0.32.0", optional = true, default-features = false }
+miniscript = { version = "12.0.0", optional = true, default-features = false }
 
 [dev-dependencies]
 bdk_testenv = { path = "../testenv", default-features = false }
@@ -27,7 +26,7 @@ tokio = { version = "1", features = ["rt", "rt-multi-thread", "macros"] }
 
 [features]
 default = ["std", "async-https", "blocking-https-rustls"]
-std = ["bdk_chain/std"]
+std = ["bdk_chain/std", "miniscript?/std"]
 async = ["async-trait", "futures", "esplora-client/async"]
 async-https = ["async", "esplora-client/async-https"]
 async-https-rustls = ["async", "esplora-client/async-https-rustls"]
--- a/crates/wallet/src/descriptor/error.rs
+++ b/crates/wallet/src/descriptor/error.rs
@@ -37,7 +37,7 @@ pub enum Error {
     /// Error during base58 decoding
     Base58(bitcoin::base58::Error),
     /// Key-related error
-    Pk(bitcoin::key::Error),
+    Pk(bitcoin::key::ParsePublicKeyError),
     /// Miniscript error
     Miniscript(miniscript::Error),
     /// Hex decoding error
@@ -103,8 +103,8 @@ impl From<bitcoin::base58::Error> for Error {
     }
 }
 
-impl From<bitcoin::key::Error> for Error {
-    fn from(err: bitcoin::key::Error) -> Self {
+impl From<bitcoin::key::ParsePublicKeyError> for Error {
+    fn from(err: bitcoin::key::ParsePublicKeyError) -> Self {
         Error::Pk(err)
     }
 }
--- a/crates/wallet/src/descriptor/policy.rs
+++ b/crates/wallet/src/descriptor/policy.rs
@@ -40,6 +40,7 @@ use crate::collections::{BTreeMap, HashSet, VecDeque};
 use alloc::string::String;
 use alloc::vec::Vec;
 use core::cmp::max;
+use miniscript::miniscript::limits::{MAX_PUBKEYS_IN_CHECKSIGADD, MAX_PUBKEYS_PER_MULTISIG};
 
 use core::fmt;
 
@@ -48,12 +49,12 @@ use serde::{Serialize, Serializer};
 
 use bitcoin::bip32::Fingerprint;
 use bitcoin::hashes::{hash160, ripemd160, sha256};
-use bitcoin::{absolute, key::XOnlyPublicKey, PublicKey, Sequence};
+use bitcoin::{absolute, key::XOnlyPublicKey, relative, PublicKey, Sequence};
 
 use miniscript::descriptor::{
     DescriptorPublicKey, ShInner, SinglePub, SinglePubKey, SortedMultiVec, WshInner,
 };
-use miniscript::hash256;
+use miniscript::{hash256, Threshold};
 use miniscript::{
     Descriptor, Miniscript, Satisfier, ScriptContext, SigType, Terminal, ToPublicKey,
 };
@@ -137,7 +138,7 @@ pub enum SatisfiableItem {
     /// Relative timelock locktime
     RelativeTimelock {
         /// The timelock value
-        value: Sequence,
+        value: relative::LockTime,
     },
     /// Multi-signature public keys with threshold count
     Multisig {
@@ -586,30 +587,25 @@ impl Policy {
         Ok(Some(policy))
     }
 
-    fn make_multisig<Ctx: ScriptContext + 'static>(
-        keys: &[DescriptorPublicKey],
+    fn make_multi<Ctx: ScriptContext + 'static>(
+        threshold: &Threshold<DescriptorPublicKey, MAX_PUBKEYS_PER_MULTISIG>,
         signers: &SignersContainer,
         build_sat: BuildSatisfaction,
-        threshold: usize,
         sorted: bool,
         secp: &SecpCtx,
     ) -> Result<Option<Policy>, PolicyError> {
-        if threshold == 0 {
-            return Ok(None);
-        }
-
-        let parsed_keys = keys.iter().map(|k| PkOrF::from_key(k, secp)).collect();
+        let parsed_keys = threshold.iter().map(|k| PkOrF::from_key(k, secp)).collect();
 
         let mut contribution = Satisfaction::Partial {
-            n: keys.len(),
-            m: threshold,
+            n: threshold.n(),
+            m: threshold.k(),
             items: vec![],
             conditions: Default::default(),
             sorted: Some(sorted),
         };
         let mut satisfaction = contribution.clone();
 
-        for (index, key) in keys.iter().enumerate() {
+        for (index, key) in threshold.iter().enumerate() {
             if signers.find(signer_id(key, secp)).is_some() {
                 contribution.add(
                     &Satisfaction::Complete {
@@ -635,12 +631,63 @@ impl Policy {
 
         let mut policy: Policy = SatisfiableItem::Multisig {
             keys: parsed_keys,
-            threshold,
+            threshold: threshold.k(),
         }
         .into();
         policy.contribution = contribution;
         policy.satisfaction = satisfaction;
 
+        Ok(Some(policy))
+    }
+
+    fn make_multi_a<Ctx: ScriptContext + 'static>(
+        threshold: &Threshold<DescriptorPublicKey, MAX_PUBKEYS_IN_CHECKSIGADD>,
+        signers: &SignersContainer,
+        build_sat: BuildSatisfaction,
+        sorted: bool,
+        secp: &SecpCtx,
+    ) -> Result<Option<Policy>, PolicyError> {
+        let parsed_keys = threshold.iter().map(|k| PkOrF::from_key(k, secp)).collect();
+
+        let mut contribution = Satisfaction::Partial {
+            n: threshold.n(),
+            m: threshold.k(),
+            items: vec![],
+            conditions: Default::default(),
+            sorted: Some(sorted),
+        };
+        let mut satisfaction = contribution.clone();
+
+        for (index, key) in threshold.iter().enumerate() {
+            if signers.find(signer_id(key, secp)).is_some() {
+                contribution.add(
+                    &Satisfaction::Complete {
+                        condition: Default::default(),
+                    },
+                    index,
+                )?;
+            }
+            if let Some(psbt) = build_sat.psbt() {
+                if Ctx::find_signature(psbt, key, secp) {
+                    satisfaction.add(
+                        &Satisfaction::Complete {
+                            condition: Default::default(),
+                        },
+                        index,
+                    )?;
+                }
+            }
+        }
+        satisfaction.finalize();
+        contribution.finalize();
+
+        let mut policy: Policy = SatisfiableItem::Multisig {
+            keys: parsed_keys,
+            threshold: threshold.k(),
+        }
+        .into();
+        policy.contribution = contribution;
+        policy.satisfaction = satisfaction;
         Ok(Some(policy))
     }
 
@@ -725,7 +772,7 @@ impl Policy {
                 timelock: Some(*value),
             }),
             SatisfiableItem::RelativeTimelock { value } => Ok(Condition {
-                csv: Some(*value),
+                csv: Some((*value).into()),
                 timelock: None,
             }),
             _ => Ok(Condition::default()),
@@ -952,11 +999,14 @@ impl<Ctx: ScriptContext + 'static> ExtractPolicy for Miniscript<DescriptorPublic
                 Some(policy)
             }
             Terminal::Older(value) => {
-                let mut policy: Policy = SatisfiableItem::RelativeTimelock { value: *value }.into();
+                let mut policy: Policy = SatisfiableItem::RelativeTimelock {
+                    value: (*value).into(),
+                }
+                .into();
                 policy.contribution = Satisfaction::Complete {
                     condition: Condition {
                         timelock: None,
-                        csv: Some(*value),
+                        csv: Some((*value).into()),
                     },
                 };
                 if let BuildSatisfaction::PsbtTimelocks {
@@ -966,9 +1016,11 @@ impl<Ctx: ScriptContext + 'static> ExtractPolicy for Miniscript<DescriptorPublic
                 } = build_sat
                 {
                     let older = Older::new(Some(current_height), Some(input_max_height), false);
-                    let older_sat = Satisfier::<bitcoin::PublicKey>::check_older(&older, *value);
-                    let inputs_sat = psbt_inputs_sat(psbt)
-                        .all(|sat| Satisfier::<bitcoin::PublicKey>::check_older(&sat, *value));
+                    let older_sat =
+                        Satisfier::<bitcoin::PublicKey>::check_older(&older, (*value).into());
+                    let inputs_sat = psbt_inputs_sat(psbt).all(|sat| {
+                        Satisfier::<bitcoin::PublicKey>::check_older(&sat, (*value).into())
+                    });
                     if older_sat && inputs_sat {
                         policy.satisfaction = policy.contribution.clone();
                     }
@@ -986,8 +1038,11 @@ impl<Ctx: ScriptContext + 'static> ExtractPolicy for Miniscript<DescriptorPublic
             Terminal::Hash160(hash) => {
                 Some(SatisfiableItem::Hash160Preimage { hash: *hash }.into())
             }
-            Terminal::Multi(k, pks) | Terminal::MultiA(k, pks) => {
-                Policy::make_multisig::<Ctx>(pks, signers, build_sat, *k, false, secp)?
+            Terminal::Multi(threshold) => {
+                Policy::make_multi::<Ctx>(threshold, signers, build_sat, false, secp)?
+            }
+            Terminal::MultiA(threshold) => {
+                Policy::make_multi_a::<Ctx>(threshold, signers, build_sat, false, secp)?
             }
             // Identities
             Terminal::Alt(inner)
@@ -1016,8 +1071,9 @@ impl<Ctx: ScriptContext + 'static> ExtractPolicy for Miniscript<DescriptorPublic
                 a.extract_policy(signers, build_sat, secp)?,
                 b.extract_policy(signers, build_sat, secp)?,
             )?,
-            Terminal::Thresh(k, nodes) => {
-                let mut threshold = *k;
+            Terminal::Thresh(threshold) => {
+                let mut k = threshold.k();
+                let nodes = threshold.data();
                 let mapped: Vec<_> = nodes
                     .iter()
                     .map(|n| n.extract_policy(signers, build_sat, secp))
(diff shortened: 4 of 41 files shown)


oleonardolima:
NOTE: related to the mentioned todo's on the description, I'll open a new PR to add more scope/context to this one with all the changes required due to the new `NetworkKind` and new `Weight`.

crates/wallet/src/descriptor/error.rs
```diff
@@ -37,7 +37,7 @@ pub enum Error {
     /// Error during base58 decoding
     Base58(bitcoin::base58::Error),
     /// Key-related error
-    Pk(bitcoin::key::Error),
+    Pk(bitcoin::key::ParsePublicKeyError),
```
ValuedMammal, line 40: I'm undecided if we need this `Pk` variant

crates/wallet/src/descriptor/policy.rs
```diff
@@ -952,11 +999,14 @@ impl<Ctx: ScriptContext + 'static> ExtractPolicy for Miniscript<DescriptorPublic
                 Some(policy)
             }
             Terminal::Older(value) => {
-                let mut policy: Policy = SatisfiableItem::RelativeTimelock { value: *value }.into();
+                let mut policy: Policy = SatisfiableItem::RelativeTimelock {
+                    value: (*value).into(),
+                }
+                .into();
```
ValuedMammal, line 952: What do you think about having `SatisfiableItem::RelativeTimelock` hold a value of type `relative::LockTime` instead of `Sequence` ?

crates/wallet/src/wallet/mod.rs
```diff
@@ -1502,10 +1502,12 @@ impl Wallet {
 
         let recipients = params.recipients.iter().map(|(r, v)| (r, *v));
 
+        #[allow(deprecated)]
         for (index, (script_pubkey, value)) in recipients.enumerate() {
             if !params.allow_dust
                 && value.is_dust(script_pubkey)
                 && !script_pubkey.is_provably_unspendable()
+            // FIXME: (@leonardo) I think it needs further discussion, should we rely on the suggested `is_op_return` ?
```
ValuedMammal: Yes I'm in favor of using `is_op_return`

crates/wallet/src/wallet/mod.rs
```diff
@@ -1704,7 +1706,9 @@ impl Wallet {
             .iter()
             .any(|txin| txin.sequence.to_consensus_u32() <= 0xFFFFFFFD)
         {
-            return Err(BuildFeeBumpError::IrreplaceableTransaction(tx.txid()));
+            return Err(BuildFeeBumpError::IrreplaceableTransaction(
+                tx.compute_txid(),
```
ValuedMammal, line 1674: Also, L1671 there's an old `.txid()` in a doc comment

crates/wallet/src/wallet/signer.rs
```diff
@@ -548,21 +566,24 @@ fn sign_psbt_ecdsa(
...
-    hash: impl bitcoin::hashes::Hash + bitcoin::secp256k1::ThirtyTwoByteHash,
-    hash_ty: EcdsaSighashType,
+    hash: impl bitcoin::hashes::Hash<Bytes = [u8; 32]>,
+    sighash_type: EcdsaSighashType,
     secp: &SecpCtx,
     allow_grinding: bool,
 ) {
-    let msg = &Message::from(hash);
-    let sig = if allow_grinding {
+    let msg = &Message::from_digest(hash.to_byte_array());
```
ValuedMammal, line 575: We could also just pass a `&Message` to the function `sign_psbt_ecdsa`

crates/wallet/src/wallet/signer.rs
```diff
@@ -158,8 +158,12 @@ pub enum SignerError {
...
     /// Invalid SIGHASH for the signing context in use
     InvalidSighash,
-    /// Error while computing the hash to sign
-    SighashError(sighash::Error),
+    /// Error while computing the hash to sign a P2WPKH input.
+    SighashP2wpkh(sighash::P2wpkhError),
+    /// Error while computing the hash to sign a Taproot input.
+    SighashTaproot(sighash::TaprootError),
+    /// Error while computing the hash, out of bounds access on the transaction inputs.
+    InputsIndexError(transaction::InputsIndexError),
```
ValuedMammal: 📌  Considering we already have a `InputIndexOutOfRange` error, I recommend consolidating these into one variant if possible.

ValuedMammal:
@tcharding Do the changes related to the new `Threshold` still look ok to you?

crates/chain/src/descriptor_ext.rs
```diff
@@ -31,7 +31,7 @@ impl DescriptorExt for Descriptor<DescriptorPublicKey> {
         self.at_derivation_index(0)
             .expect("descriptor can't have hardened derivation")
             .script_pubkey()
-            .dust_value()
+            .minimal_non_dust()
             .to_sat()
```
notmandatory, line 35: 📌 since this is a public API should we change the `DescriptorExt` trait to make this an `Amount` instead of u64?

crates/esplora/Cargo.toml
```diff
@@ -13,21 +13,21 @@ readme = "README.md"
...
 bdk_chain = { path = "../chain", version = "0.15.0", default-features = false }
-esplora-client = { version = "0.7.0", default-features = false }
+esplora-client = { version = "0.8.0", default-features = false }
 async-trait = { version = "0.1.66", optional = true }
 futures = { version = "0.3.26", optional = true }
 
-# use these dependencies if you need to enable their /no-std features
-bitcoin = { version = "0.31.0", optional = true, default-features = false }
-miniscript = { version = "11.0.0", optional = true, default-features = false }
+# For no-std, remember it's already implied when not enabling the `std` feature.
```
notmandatory: ⛏️ nit, I think this description would be more clearly stated (here and other files) as:
```suggestion
# The `no-std` feature it's implied when the `std` feature is disabled.
```

crates/wallet/src/descriptor/policy.rs
notmandatory: Keep in mind that this module will likely be deprecated in the next major release (2.0) in favor of the new miniscript [plan module](https://docs.rs/miniscript/latest/miniscript/plan/index.html). Unless we're fixing a bug it may not be worth the refactoring to improve the API.

crates/wallet/src/types.rs
```diff
@@ -72,7 +72,7 @@ pub struct WeightedUtxo {
     /// properly maintain the feerate when adding this input to a transaction during coin selection.
     ///
     /// [weight units]: https://en.bitcoin.it/wiki/Weight_units
-    pub satisfaction_weight: usize,
+    pub satisfaction_weight: usize, // FIXME: (@leonardo) I think it should expect the new bitcoin-units `Weight` type instead of usize.
```
notmandatory: 👍 agreed, now that there's a `Weight` type we should use it. If you think it's too big a change to add to this PR feel free to make an issue to fix it in a future PR.

crates/wallet/src/wallet/mod.rs
notmandatory: Agreed, we should use `is_op_return()`, especially since `is_provably_unspendable()` is now deprecated.

crates/wallet/src/wallet/mod.rs
notmandatory: A few other spots in doc code that need to have tx.txid() fixed:
chain, tx_data_traits.rs: L46, L61, L79
wallet, wallet/mod.rs: L1671

notmandatory (commented):
I have a couple small suggestions but otherwise this looks almost ready. Great work!

I also created #1469 to remove `tmp_plan`, should it be done with this PR?

crates/wallet/src/descriptor/error.rs
oleonardolima: What are your thoughts about it?

crates/wallet/src/descriptor/policy.rs
oleonardolima: @ValuedMammal I think that's a great suggestion, but I'm now unsure if it's worth it considering what @notmandatory, I'll take a look if it's only a few LOC/changes that would be needed for the improvement.

crates/wallet/src/wallet/mod.rs
oleonardolima: Great, I'll fix it to use it then.

crates/wallet/src/wallet/mod.rs
oleonardolima: Thanks, sorry for the oversight, it's probably because `rust-analyzer` does not handle these due to being under feature/docs codes, and it didn't show up as failure. I'll handle these.

crates/wallet/src/wallet/signer.rs
oleonardolima: @ValuedMammal Wouldn't it add more complexity/verbosity ? 🤔

crates/wallet/src/wallet/signer.rs
oleonardolima: @ValuedMammal Oh, thanks! Yes, that's much better, I'll fix it.

crates/chain/src/descriptor_ext.rs
oleonardolima: @notmandatory Yes, I do agree. I'll update it as well.

crates/esplora/Cargo.toml
oleonardolima: @notmandatory Great, thanks for the suggestion.

crates/wallet/src/types.rs
oleonardolima: @notmandatory I wasn't so sure about, but I think it adds unrelated scope/context, I'm addressing this one in  https://github.com/bitcoindevkit/bdk/pull/1468

crates/wallet/src/descriptor/error.rs
ValuedMammal: afaik it's not being used anywhere

crates/wallet/src/wallet/signer.rs
ValuedMammal: I'm thinking this module might eventually look more like in #1424, but if this works for now then that's great too

crates/chain/src/descriptor_ext.rs
oleonardolima: @notmandatory I'll address it in another PR, it ends up breaking and requiring changes on the `CoinSelectorOpt`, and probably will also raise some discussions, which is outside the scope of this PR.

crates/wallet/src/wallet/signer.rs
oleonardolima: @ValuedMammal I did a deep-dive in rust-bitcoin, and now I'm wondering if we should have two variants, one for Psbt and another one for Transaction, as they have two in rust-bitcoin crate:
- bitcoin::psbt::IndexOutOfBoundsError
- bitcoin::transaction::InputsIndexError & bitcoin::transaction::OutputsIndexError

crates/esplora/Cargo.toml
```diff
@@ -13,21 +13,21 @@ readme = "README.md"
...
+miniscript = { version = "12.0.0", optional = true, default-features = false }
 
 [dev-dependencies]
 bdk_testenv = { path = "../testenv", default-features = false }
 tokio = { version = "1", features = ["rt", "rt-multi-thread", "macros"] }
 
 [features]
 default = ["std", "async-https", "blocking-https-rustls"]
-std = ["bdk_chain/std"]
+std = ["bdk_chain/std", "miniscript?/std"]
```
tcharding, line 29: What does the `?` do in this line?

crates/esplora/Cargo.toml
```diff
@@ -13,21 +13,21 @@ readme = "README.md"
...
 bdk_chain = { path = "../chain", version = "0.15.0", default-features = false }
-esplora-client = { version = "0.7.0", default-features = false }
+esplora-client = { version = "0.8.0", default-features = false }
 async-trait = { version = "0.1.66", optional = true }
 futures = { version = "0.3.26", optional = true }
 
-# use these dependencies if you need to enable their /no-std features
-bitcoin = { version = "0.31.0", optional = true, default-features = false }
-miniscript = { version = "11.0.0", optional = true, default-features = false }
+# The `no-std` feature it's implied when the `std` feature is disabled.
```
tcharding: This comment is not really correct, there is no `no-std` feature anymore in `rust-bitcoin`. Its in another place also.

tcharding:
> @tcharding Do the changes related to the new `Threshold` still look ok to you?

Handwavey response, if it builds ship it.

Take everything I saw with a pinch of salt, I'm not that good with `miniscript`.

I'm not sure, I don't exactly know what you are aiming for:

1. Just make it build - then CI can answer that.
2. Make it build and not be obviously wrong - seems like the usage is ok to me.
3. Fully utilise `miniscript`'s new `Threshold` type - definitely currently not doing this.

I didn't grok the `fragment` macro totally but was surprised to see `expect` called if inputs to the macro are invalid thresholds.  Passing the closure is a bit hacky but I totally feel your pain, patching macros is an epic PITA.

Sorry I can't be much more help than that.

crates/esplora/Cargo.toml
ValuedMammal: Basically, only enable the std feature if something else already enabled the miniscript dep.

https://doc.rust-lang.org/cargo/reference/features.html#dependency-features

tcharding: Oh sweet, TIL. (Doesn't help us in `rust-bitcoin` yet though because it needs Rust 1.60)

crates/esplora/Cargo.toml
oleonardolima: Good point! I think it's best to just remove the comments altogether then.
fyi @notmandatory

crates/wallet/src/wallet/signer.rs
oleonardolima: I updated it to be `TxInputsIndexError` instead, I also think for the current `InputIndexOutOfRange` it should be `PsbtInputIndexOutOfRange(psbt::IndexOutOfBoundsError)` instead, let me know what you think.

crates/wallet/src/descriptor/policy.rs
oleonardolima: @ValuedMammal I updated the variant, it was trivial. I think the most of work would be updating `Condition` (below), but I don't its work as @notmandatory  mentioned 🤔  https://github.com/bitcoindevkit/bdk/blob/473ef9714f642ee2ad393541763881cf69132262/crates/wallet/src/descriptor/policy.rs#L446-L456

ValuedMammal:
The summary of changes in the description certainly helps, thanks for that. In terms of review, the individual commits are less useful since realistically this should be a single commit - the one that finally builds with tests passing.

ValuedMammal:
> I'm not sure, I don't exactly know what you are aiming for:
> 
>     1. Just make it build - then CI can answer that.
> 
>     2. Make it build and not be obviously wrong - seems like the usage is ok to me.
> 
>     3. Fully utilise `miniscript`'s new `Threshold` type - definitely currently not doing this.

Mostly number 2 for now, ha. All good, appreciate you @tcharding !

crates/wallet/src/descriptor/policy.rs
ValuedMammal: `Condition` should be fine as it is now

crates/wallet/src/wallet/signer.rs
```diff
@@ -159,8 +159,14 @@ pub enum SignerError {
...
-    /// Error while computing the hash to sign
-    SighashError(sighash::Error),
+    /// Error while computing the hash to sign a P2WPKH input.
+    SighashP2wpkh(sighash::P2wpkhError),
+    /// Error while computing the hash to sign a Taproot input.
+    SighashTaproot(sighash::TaprootError),
+    /// Error while computing the hash, out of bounds access on the transaction inputs.
+    TxInputsIndexError(transaction::InputsIndexError),
+    /// Error while computing the hash, out of bounds access on the transaction outputs.
+    TxOutputsIndexError(transaction::OutputsIndexError),
```
ValuedMammal: I'm not sure the outputs are relevant to computing the sighash

```suggestion
```

ValuedMammal (changes requested):

oleonardolima:
> The summary of changes in the description certainly helps, thanks for that. In terms of review, the individual commits are less useful since realistically this should be a single commit - the one that finally builds with tests passing.

Nice, thanks for the feedback! I'll squash the commits into a single one.

crates/wallet/src/wallet/signer.rs
ValuedMammal: A simple solution would be to make `SignerError::InputIndexOutOfRange` into a struct with fields for `index` and `length` (essentially reimplementing [`IndexOutOfBoundsError`](https://docs.rs/bitcoin/latest/bitcoin/blockdata/transaction/struct.IndexOutOfBoundsError.html)) and then we have

```rust
impl From<transaction::InputsIndexError> for SignerError {}
```

similar to what you have here, but it shouldn't be a blocker to this PR. Accessing the outputs of a tx or psbt doesn't appear to be a concern in this module.

ValuedMammal:
ACK 2a4564097bd70905ed9fa21b92cdf074a72f878e

ValuedMammal (approved):

crates/esplora/Cargo.toml
evanlinjin: TIL +1

crates/wallet/src/descriptor/policy.rs
```diff
@@ -644,6 +640,57 @@ impl Policy {
         Ok(Some(policy))
     }
 
+    fn make_multi_a<Ctx: ScriptContext + 'static>(
```
evanlinjin: ~~Thanks for the attempt to rewrite what is here! I would not bother though, since we want to replace this `policy` stuff with `miniscript`'s planning module.~~

Edit: okay I'm trying to figure out why the policy will be different here?

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -295,7 +295,9 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
 
             for utxo in utxos {
                 let descriptor = wallet.get_descriptor_for_keychain(utxo.keychain);
-                let satisfaction_weight = descriptor.max_weight_to_satisfy().unwrap();
+
+                let satisfaction_weight =
+                    descriptor.max_weight_to_satisfy().unwrap().to_wu() as usize;
```
evanlinjin, line 300: Okay we need to create a ticket for what we should do when a caller creates a wallet with descriptors that are insatiable. Panicing with no clear reason when they try construct a tx is not the right behavior imo. Out of scope for this PR.

oleonardolima: I agree, I tried to solve that problem in #1468 by propagating the error, but it needs further discussion.

crates/wallet/src/descriptor/policy.rs
evanlinjin: Now I understand. The only difference is when constructing the `Threshold`, we enforce different max `n` (of the `k-of-on` theshold) for different script constructions.

Can we use a const generic instead of duplicating code?

```rust
    fn make_multi<Ctx: ScriptContext + 'static, const MAX: usize>(
        threshold: &Threshold<DescriptorPublicKey, MAX>,
        signers: &SignersContainer,
        build_sat: BuildSatisfaction,
        sorted: bool,
        secp: &SecpCtx,
    ) -> Result<Option<Policy>, PolicyError> { todo!() }
```

evanlinjin: Due to time constraints I'm going to do this and push a commit on top.

evanlinjin: Done in 11200810d048abb8deb3c7961ca93c68011d41fd

notmandatory (approved):
ACK 11200810d048abb8deb3c7961ca93c68011d41fd

crates/wallet/src/descriptor/policy.rs
oleonardolima: Great! That's much better, thanks for the suggestion and for addressing this @evanlinjin!

crates/wallet/src/wallet/tx_builder.rs
oleonardolima: I moved #1468 to ready to review now, we can move this discussion there then.

apoelstra: Maybe upstream could add a field to the `Params` structure which would refuse to parse un-satiable descriptors. Could you file an issue in rust-miniscript asking for strategies to handle the error return of max_weight_to_satisfy?

Alternately you can just use `unwrap_or(usize::MAX)` or something silly.

oleonardolima: Yes, I did it https://github.com/rust-bitcoin/rust-miniscript/issues/695

Issue #1317, opened 2024-02-02
[chain] keychain module doesn't need to exist


LLFourn:
The module tries to explain itself as follows:

> A keychain here is a set of application-defined indexes for a miniscript descriptor where we can derive script pubkeys at a particular derivation index.

Which is very difficult to read. More importantly it only has one thing in it `KeyChainTxOutIndex`. We can just have `keychain_txout_index.rs`. We could move indexers into their own module so `indexers/keychain_txout.rs`, `indexers/spk_txout.rs`.

evanlinjin:
Let's do this after #1203. I'll take responsibility for the convoluted crate docs.

ValuedMammal:
@evanlinjin If you want I can open a PR and have you make suggestions for the crate docs.. Actually, removing this module means we won't need the old docs. This also gives us a chance to finally relocate the Balance struct

PR #1451, opened 2024-05-28
Various improvements not included in #1203 (new)

### LINKED ISSUE #1439: txout_index, `Descriptor` not needed in `keychains_to_descriptors`
The `Descriptor` does not need to be here. You can just look it up in `descriptor_ids_to_descriptors`.

_Originally posted by @LLFourn in https://github.com/bitcoindevkit/bdk/pull/1203#discussion_r1596118215_

LagginTimes:
Closes #1439

### Description

These were tasks that were intended to be included in #1438.
See https://github.com/bitcoindevkit/bdk/pull/1438#issuecomment-2134400460.

* [x] Rename field `descriptor_ids_to_descriptors` to `descriptors`.
* [x] Rename field `descriptor_ids_to_keychains` to `keychains`.
* [x] Fix this comment:
https://github.com/bitcoindevkit/bdk/pull/1438#issuecomment-2113478065 to that of the version in
https://github.com/bitcoindevkit/bdk/pull/1428
* [x] https://github.com/bitcoindevkit/bdk/pull/1341
* [x] Make returned keychains non-reference.

#### Original Ticket Description

Adds various improvements to the work of #1203. These were missed out while cherry-picking, or review comments left in #1428 that were forgotten:

* Change `keychains_to_descriptors` to `keychains_to_descriptor_ids` which simplifies the field. This was mentioned [here](https://github.com/bitcoindevkit/bdk/pull/1428#discussion_r1591795543) and included in #1428, but an older commit was cherry-picked.
* Rename `KeychainTxOutIndex` field `descriptor_ids_to_keychain_set` to `descriptor_ids_to_keychains` was missed out as an older commit was cherry-picked. This change to naming shows the direct relationship between `keychains_to_desriptor_ids` and `descriptor_ids_to_keychains` (one is directly a reverse lookup of the other).
* Change `reveal_to_target_with_id` to `reveal_to_target_with_descriptor`, reasoning mentioned [here](https://github.com/bitcoindevkit/bdk/pull/1428#discussion_r1596527461).

In addition to this, I changed the output signature of `reveal_to_target`, `reveal_next_spk` and `next_unused_spk` methods to return `(Option<spk(s)>, changeset)`, whereas previously it was `Option<(spk(s), changeset)>`. This makes the API more consistent as the `ChangeSet` is always returned, and `reveal_to_target` and `unbounded_spk_iter`-esc methods all return `Option<SpkIterator>` (which we can `.flatten()`).

### Notes to the reviewers

Not all changes in this PR are Changelog-worthy. I.e. renaming of internal variables to increase readability, changing code comments, refactoring private methods - are all excluded from the changelog.

### Changelog notice

* Change `KeychainTxOutIndex` methods to always return a changeset. This makes the API more consistent.
* Change `KeychainTxOutIndex` methods to return spks as `ScriptBuf` instead of references (`&Script`) to reduce lifetimes of mutable borrows.

#### All Submissions:

### DIFF
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -506,9 +506,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
             if old_desc_id == desc_id {
                 return changeset;
             }
-            // we should remove old descriptor that is associated with this keychain as the index
-            // is designed to track one descriptor per keychain (however different keychains can
-            // share the same descriptor)
+            // remove keychain from reverse index
             let _is_keychain_removed = self
                 .keychains
                 .get_mut(&old_desc_id)
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -207,14 +207,16 @@ const DEFAULT_LOOKAHEAD: u32 = 25;
 #[derive(Clone, Debug)]
 pub struct KeychainTxOutIndex<K> {
     inner: SpkTxOutIndex<(DescriptorId, u32)>,
-    // keychain -> (descriptor, descriptor id) map
-    keychains_to_descriptors: BTreeMap<K, (DescriptorId, Descriptor<DescriptorPublicKey>)>,
-    // descriptor id -> keychain set
-    // Because different keychains can have the same descriptor, we rank keychains by `Ord` so that
-    // that the first keychain variant (according to `Ord`) has the highest rank. When associated
-    // data (such as spks, outpoints) are returned with a keychain, we return the highest-ranked
-    // keychain with it.
-    descriptor_ids_to_keychain_set: HashMap<DescriptorId, BTreeSet<K>>,
+    // keychain -> descriptor_id map
+    keychains_to_descriptor_ids: BTreeMap<K, DescriptorId>,
+    // descriptor_id -> keychain set
+    // This is a reverse map of `keychains_to_descriptors`. Although there is only one descriptor
+    // per keychain, different keychains can refer to the same descriptor, therefore we have a set
+    // of keychains per descriptor. When associated data (such as spks, outpoints) are returned with
+    // a keychain, we return it with the highest-ranked keychain with it. We rank keychains by
+    // `Ord`, therefore the keychain set is a `BTreeSet`. The earliest keychain variant (according
+    // to `Ord`) has precedence.
+    descriptor_ids_to_keychains: HashMap<DescriptorId, BTreeSet<K>>,
     // descriptor_id -> descriptor map
     // This is a "monotone" map, meaning that its size keeps growing, i.e., we never delete
     // descriptors from it. This is useful for revealing spks for descriptors that don't have
@@ -289,8 +291,8 @@ impl<K> KeychainTxOutIndex<K> {
     pub fn new(lookahead: u32) -> Self {
         Self {
             inner: SpkTxOutIndex::default(),
-            keychains_to_descriptors: BTreeMap::new(),
-            descriptor_ids_to_keychain_set: HashMap::new(),
+            keychains_to_descriptor_ids: BTreeMap::new(),
+            descriptor_ids_to_keychains: HashMap::new(),
             descriptor_ids_to_descriptors: BTreeMap::new(),
             last_revealed: BTreeMap::new(),
             lookahead,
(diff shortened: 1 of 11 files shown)


evanlinjin:
@LagginTimes could you also rebase this please? Thanks!

crates/chain/src/keychain/txout_index.rs
```diff
@@ -207,19 +210,21 @@ const DEFAULT_LOOKAHEAD: u32 = 25;
...
+    // keychain -> descriptor_id map
+    keychains_to_descriptor_ids: BTreeMap<K, DescriptorId>,
+    // descriptor_id -> keychain set
+    // This is a reverse map of `keychains_to_descriptors`. Although there is only one descriptor
+    // per keychain, different keychains can refer to the same descriptor, therefore we have a set
+    // of keychains per descriptor. When associated data (such as spks, outpoints) are returned with
+    // a keychain, we return it with the highest-ranked keychain with it. We rank keychains by
+    // `Ord`, therefore the keychain set is a `BTreeSet`. The earliest keychain variant (according
+    // to `Ord`) has precedence.
+    keychains: HashMap<DescriptorId, BTreeSet<K>>,
```
evanlinjin, line 222: I have a second thought about this renaming. What do you think about: `ids_by_keychain` and `keychains_by_id`?

LagginTimes: My initial thought is, if we're simplifying `descriptor_ids_to_keychains` to something like `keychains_by_id`, would it make sense to do something similar to `keychains_to_descriptor_ids`?

crates/chain/src/keychain/txout_index.rs
```diff
@@ -580,28 +601,39 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
             .collect()
     }
 
     /// Iterate over revealed spks of keychains in `range`
     pub fn revealed_spks(
         &self,
         range: impl RangeBounds<K>,
-    ) -> impl DoubleEndedIterator<Item = (&K, u32, &Script)> + Clone {
-        self.keychains_to_descriptors
+    ) -> impl DoubleEndedIterator<Item = (&K, u32, ScriptBuf)> + Clone {
```
evanlinjin: What are our thoughts about making returned keychains non-reference as well?

crates/chain/src/keychain/txout_index.rs
evanlinjin: ~~I meant `keychains_by_ids` to replace `descriptor_ids_to_keychains`. I was suggesting it because the wording makes more sense to me. However, `ids_to_keychains` can work too.~~

Edit: Sorry @LagginTimes I completely misread the comment. Yes, since one field is a reverse-index of the other so it makes sense that the names correlate.

evanlinjin:
This is looking good. Curious what others think about my comments.

crates/chain/src/keychain/txout_index.rs
notmandatory: Since K is already `Clone` this sounds like a good idea, it should simplify lifetime issues for users.

crates/chain/src/keychain/txout_index.rs
notmandatory: 🤔 my 2 sats is names should match plurality of the related data. So something like:
- `keychain_to_descriptor_id`, maps one keychain to one descriptor_id
- `descriptor_id_to_keychains`, maps one descriptor_id to multiple keychains

notmandatory (approved):
Looks good to me. I just added a couple comments, but not show stoppers. Also since this is a cleanup PR would you like to add a commit to fix #1317 also? 

ACK a5142decfcd0581bf871755b8130782666329017

evanlinjin:
I was just on a call with @LLFourn this morning. It turns out that having the internal `SpkTxOutIndex` use `DescriptorId`s is causing issues when using range-based methods.

The problem is evident when we query outputs in a given keychain (`K`) range, where the range is across multiple keychains. The keychain (`K`) range query does not translate well since the `Ord` of `DescriptorId`s is not guaranteed to be one-to-one with the `Ord` of keychains (`K`).

The first solution (the one @LLFourn is working on) is to use keychain (`K`) with the internal `SpkTxOutIndex`. To maintain the "consistency", we disallow reassigning keychains and assigning the same descriptor to multiple keychains. The `ChangeSet::append` method will ignore keychain-reassignments and assigning the same descriptor to another keychain. This is the solution with the simplest codebase. Downsides is it disallows certain usecases (which we aren't even sure if users will use) - i.e. assigning the same descriptor to multiple keychains, and replacing the descriptor for a keychain.

The alternative is to continue to have the internal `SpkTxOutIndex` use `DescriptorId`s and when we range keychains, we first get corresponding `DescriptorId`s from `KeychainTxOutIndex::keychains_to_descriptor_ids` and range for individual `DescriptorId`s when querying the internal `SpkTxOutIndex`. This does introduce even more code, however it allows us complete flexibility and how changesets are appended will make more sense (in my opinion).

---

Edit: I created a ticket (#1459).

evanlinjin:
Replaced by #1463

PR #1453, opened 2024-05-31
refactor(electrum) put the tx cache in electrum

LLFourn:
Previously there was a `TxCache` that you passed in as part of the sync request. There are lots of downsides to this:

1. If the user forgets to do this you cache nothing
2. where are you meant to keep this cache? The example shows it being recreated every time which seems very suboptimal.
3. More API and documentation surface area.

Instead just do a plain old simple cache inside the electrum client. This way at least you only download transactions once. You can pre-populate the cache with a method also and I did this in the examples.

### DIFF
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -65,6 +65,13 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
         Ok(tx)
     }
 
+    /// Broadcasts a transaction to the network.
+    ///
+    /// This is a re-export of [`ElectrumApi::transaction_broadcast`].
+    pub fn transaction_broadcast(&self, tx: &Transaction) -> Result<Txid, Error> {
+        self.inner.transaction_broadcast(tx)
+    }
+
     /// Full scan the keychain scripts specified with the blockchain (via an Electrum client) and
     /// returns updates for [`bdk_chain`] data structures.
     ///
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -2,19 +2,69 @@ use bdk_chain::{
     bitcoin::{OutPoint, ScriptBuf, Transaction, Txid},
     collections::{BTreeMap, HashMap, HashSet},
     local_chain::CheckPoint,
-    spk_client::{FullScanRequest, FullScanResult, SyncRequest, SyncResult, TxCache},
+    spk_client::{FullScanRequest, FullScanResult, SyncRequest, SyncResult},
     tx_graph::TxGraph,
     BlockId, ConfirmationHeightAnchor, ConfirmationTimeHeightAnchor,
 };
 use core::str::FromStr;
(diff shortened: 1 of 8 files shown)


crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -2,19 +2,69 @@ use bdk_chain::{
...
 
 /// We include a chain suffix of a certain length for the purpose of robustness.
 const CHAIN_SUFFIX_LENGTH: u32 = 8;
 
-/// Trait to extend [`electrum_client::Client`] functionality.
-pub trait ElectrumExt {
+/// Wrapper around an [`electrum_client::ElectrumApi`] which includes an internal in-memory
+/// transaction cache to avoid re-fetching already downloaded transactions.
+#[derive(Debug)]
+pub struct BdkElectrumClient<E> {
```
notmandatory, line 19: I like this new name. 

📌 After this goes in we should rename `EsploraExt` to `BdkEsploraClient.

example-crates/wallet_electrum/src/main.rs
```diff
@@ -89,7 +93,7 @@ fn main() -> Result<(), anyhow::Error> {
     assert!(finalized);
 
     let tx = psbt.extract_tx()?;
-    client.transaction_broadcast(&tx)?;
+    client.inner.transaction_broadcast(&tx)?;
```
notmandatory: ⛏️ Can we also expose `transaction_broadcast()` as`client.transaction_broadcast(..)`? Most users will only be doing `sync()`, `full_scan()` and `transaction_broadcast()`.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -242,11 +428,11 @@ impl ElectrumSyncResult {
     /// This requires additional calls to the Electrum server.
     pub fn with_confirmation_time_height_anchor(
         self,
-        client: &impl ElectrumApi,
+        client: &BdkElectrumClient<impl ElectrumApi>,
```
notmandatory, line 431: ❓ Not introduced by this PR, but why do we need the `client` param? Wouldn't we always use &self.inner? Same above for `full_scan()`.

notmandatory (commented):
Concept ACK

crates/electrum/src/bdk_electrum_client.rs
evanlinjin: But `BdkElectrumClient` is a struct and `EsploraExt` is an extension trait.

notmandatory: Ah ok, that makes sense, disregard my above suggestion.

evanlinjin (approved):
self-ACK 2d2656acfa83ab4c4846c0aab14072efb64c5cc3

@LLFourn I re-exported a method `transaction_broadcast` (as per @notmandatory's suggestion) and renamed the first commit's message to match convention.

notmandatory (approved):
ACK 2d2656acfa83ab4c4846c0aab14072efb64c5cc3

PR #1454, opened 2024-05-31
Refactor wallet and persist mod, remove bdk_persist crate

notmandatory:
### Description

Sorry to submit another refactor PR for the persist related stuff, but I think it's worth revisiting. My primary motivations are: 

1. remove `db` from `Wallet` so users have the ability to use `async` storage crates, for example using `sqlx`. I updated docs and examples to let users know they are responsible for persisting changes.
2. remove the `anyhow` dependency everywhere (except as a dev test dependency). It really doesn't belong in a lib and by removing persistence from `Wallet` it isn't needed.
3. remove the `bdk_persist` crate and revert back to the original design with generic error types. I kept the `Debug` and `Display` constrains on persist errors so they could still be used with the `anyhow!` macro.

### Notes to the reviewers

I also replaced/renamed old `Persist` with `StagedPersist` struct inspired by bitcoindevkit/bdk#1453, it is only used in examples. The `Wallet` handles it's own staging.  

### Changelog notice

Changed

- Removed `db` from `Wallet`, users are now responsible for persisting changes, see docs and examples.
- Removed the `bdk_persist` crate and moved logic back to `bdk_chain::persist` module.
- Added `PersistBackendAsync`, `StagedExt`, and `StageExtAsync` traits.
- Added `Wallet` functions `commit_to` and `commit_to_async`.

#### All Submissions:


crates/chain/src/persist.rs
```diff
@@ -0,0 +1,317 @@
...
+    }
+}
+
+/// A persistence backend for writing and loading changesets.
+///
+/// `C` represents the changeset; a datatype that records changes made to in-memory data structures
+/// that are to be persisted, or retrieved from persistence.
+pub trait Persist<C> {
+    /// The error the backend returns when it fails to write.
+    type WriteError: Debug + Display;
```
rustaceanrob, line 112: I wondering if there is any advantage to having `Persist` be generic over an additional error type `E: Debug + Display` and having the function signatures be `Result<_, E>`. There is no difference in functionality from what I can tell but the user does not have to declare the `type WriteError = WhateverErrorTheyMade` internally within `Persist`.

notmandatory: That does seem simpler, but I gave it a shot just now and the problem is that I need to support different error types for `WriteError` and `LoadError`. This means the generic for Persist would have to be something like `pub trait Persist<C, WE: Debug + Display, LE: Debug + Default>` and then I'd also have to make the StagedPersist type generics `pub struct StagedPersist<C, WE: Debug + Display, LE: Debug + Display, P: Persist<C, WE, LE>>`, which looks more confusing to me than just keeping the error type definitions.

notmandatory:
@matthiasdebernardini I added an async version of the persist trait called `PersistAsync` for you to test out with your `sqlx` crate, see 4ab0648e8435d13539087792fc7a935ef9ce6cd4.

evanlinjin:
This is looking good. Removing the persistence backend from wallet is a welcome change. The main concern with doing so is newly revealed addresses not being persisted, but this is addressed with good documentation. ~Will have a proper review tomorrow.~ (give me a couple more days)

crates/wallet/src/wallet/mod.rs
```diff
@@ -744,25 +692,53 @@ impl Wallet {
     /// index defined in [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki),
     /// then returns the last revealed address.
     ///
-    /// # Errors
+    /// **WARNING:** You must persist the staged wallet changes after making this call. For example:
```
ValuedMammal: This comment should say what will happen if we forget to persist changes, but my hunch is that the method will continue to work as expected as long as all staged changes are eventually written before dropping the wallet. any thoughts on this?

```suggestion
    /// **WARNING:** To avoid address reuse you must persist the changes resulting from one or more calls to this method before closing the wallet.
```

I guess this is a similar to what you conveyed in other spots by saying "you must persist it later". In practice I don't know if it's better to commit the change immediately or along with a `wallet::Update`.

We have to demonstrate the example somehow, so I'm sure it's fine the way you have it. We could also factor out most of the documented warnings to the module-level or the docs for the `Wallet` type itself.

crates/wallet/src/wallet/mod.rs
```diff
@@ -533,168 +471,178 @@ impl Wallet {
...
+    /// let secp = Secp256k1::new();
+    ///
+    /// let external_descriptor = Descriptor::parse_descriptor(&secp, "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/0/*)").unwrap();
+    /// let internal_descriptor = Descriptor::parse_descriptor(&secp, "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/1/*)").unwrap();
+    ///
+    /// let changeset = db.load_changes()?;
+    /// let mut wallet = Wallet::new_or_load(external_descriptor, Some(internal_descriptor), changeset, Testnet)?;
+    /// # Ok(())
+    /// # }
+    /// ```
```
ValuedMammal, line 498:
```rust
/// ```rust,no_run
/// # use bdk_chain::persist::Persist;
/// # use bdk_wallet::Wallet;
/// # use bdk_sqlite::{Store, rusqlite::Connection};
/// # use bitcoin::Network::Testnet;
/// # let conn = Connection::open_in_memory().expect("must open connection");
/// let mut db = Store::new(conn).expect("must create db");
/// let changeset = db.load_changes()?;
///
/// let external_descriptor = "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/0/*)";
/// let internal_descriptor = "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/1/*)";
///
/// let mut wallet = Wallet::new_or_load(external_descriptor, Some(internal_descriptor), changeset, Testnet)?;
/// # Ok::<(), anyhow::Error>(())
/// ```    
```

crates/chain/src/persist.rs
```diff
@@ -0,0 +1,454 @@
...
+}
+
+/// `StagedPersist` adds a convenient staging area for changesets before they are persisted.
+///
+/// Not all changes to the in-memory representation needs to be written to disk right away, so
+/// [`crate::persist::StagedPersist::stage`] can be used to *stage* changes first and then
+/// [`crate::persist::StagedPersist::commit`] can be used to write changes to disk.
+pub struct StagedPersist<C, P: Persist<C>> {
+    inner: P,
+    staged: C,
```
ValuedMammal: nit: I like the old name `stage`.

crates/sqlite/src/store.rs
```diff
@@ -57,21 +57,23 @@ where
...
+    ) -> Result<(), Self::WriteError> {
+        self.write(changeset)
     }
 
-    fn load_from_persistence(&mut self) -> anyhow::Result<Option<C>> {
-        self.read()
-            .map(|c| c.map(Into::into))
-            .map_err(|e| anyhow::anyhow!(e).context("unable to read changes from sqlite database"))
+    fn load_changes(&mut self) -> Result<Option<CombinedChangeSet<K, A>>, Self::LoadError> {
+        self.read().map(|c| c.map(Into::into))
```
ValuedMammal:
```suggestion
        self.read()
```

crates/wallet/src/wallet/mod.rs
ValuedMammal: side note that the doc for `reveal_next_address` says

> This will increment the internal derivation index

"internal" is somewhat ambiguous here. Maybe just say it will "increment the keychain's derivation index."

ValuedMammal:
Maybe worth keeping the `Wallet::staged` method in order to preview what's currently staged? or else consider returning a cloned changeset from `apply_update`.

crates/wallet/src/wallet/mod.rs
```diff
@@ -744,25 +692,53 @@ impl Wallet {
...
+    /// # let internal_descriptor = Descriptor::parse_descriptor(&secp, "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/1/*)").unwrap();
+    /// # let changeset = db.load_changes()?.expect("Changeset must exist");
+    /// # let mut wallet = Wallet::load_from_changeset(changeset)?;
+    /// let next_address = wallet.reveal_next_address(KeychainKind::External);
+    /// db.write_changes(&wallet.take_staged())?;
+    /// // Now it's safe to show the user their next address!
+    /// println!("Next address: {}", next_address.address);
+    /// # Ok(())
+    /// # }
+    /// ```
```
ValuedMammal, line 683: somewhat minimal but I think it compiles

```rust
/// ```rust,no_run
/// # use bdk_chain::persist::Persist;
/// # use bdk_wallet::wallet::{Wallet, ChangeSet};
/// # use bdk_wallet::KeychainKind;
/// # use bdk_sqlite::{Store, rusqlite::Connection};
/// # let conn = Connection::open_in_memory().expect("must open connection");
/// # let mut db = Store::new(conn).expect("must create store");
/// # let changeset = ChangeSet::default();
/// # let mut wallet = Wallet::load_from_changeset(changeset).expect("load wallet");
/// let next_address = wallet.reveal_next_address(KeychainKind::External);
/// db.write_changes(&wallet.take_staged())?;
///
/// // Now it's safe to show the user their next address!
/// println!("Next address: {}", next_address.address);
/// # Ok::<(), anyhow::Error>(())
/// ```
```

crates/wallet/src/wallet/mod.rs
notmandatory: I like the wording you suggested since it's more specific about what will happen if you don't persist the changes. I also did some rewording for `reveal_next_address` along the lines you suggested. 

As for moving the warning about persisting updates at the module level, I ike putting it on each affected method so it's super clear which calls need the user to persist changes for, and what will happen if they don't. To that end I added more warnings, hopefully it doesn't look too annoying.

crates/chain/src/persist.rs
notmandatory: I was on the fence with the name here too, I renamed it back to `stage`.

notmandatory:
> Maybe worth keeping the `Wallet::staged` method in order to preview what's currently staged? or else consider returning a cloned changeset from `apply_update`.

@ValuedMammal Did you have a particular use in mind for previewing the staged? I can't think of a reason for it and don't want to add new functionality that may never be used.

crates/sqlite/src/store.rs
notmandatory: Good call, fixed.

crates/wallet/src/wallet/mod.rs
notmandatory: Thanks for making this example and the one below more succinct. I've updated as you suggest.

ValuedMammal:
> Did you have a particular use in mind for previewing the staged?

Nothing in particular, mostly thinking about issue bitcoindevkit/bdk_wallet#6 but I guess a raw changeset is not exactly a convenient list of events.

notmandatory:
Ah ya, for bitcoindevkit/bdk_wallet#6 the changeset should have the required data but probably not in a very easy to use form. Let's tackle that issue in a new PR.

notmandatory:
FYI CI is currently broken due to a dependency changing their MSRV on a patch release 😞  servo/rust-url#937. Before pinning it I want to see if the dev is going to fix it.

notmandatory:
I added a commit to pin the `url` dependency, hopefully upstream will fix the issue so I can remove the pin before this PR is ready to merge. 94e26270988c15aa741d3ff382d4cd56723acf7e

ValuedMammal:
Your base seems off, was this duplicated? bb23739c

notmandatory:
> Your base seems off, was this duplicated? [bb23739](https://github.com/bitcoindevkit/bdk/commit/bb23739c2079aab4c777945f64f610cdd170858e)

Oh shoot. Ya I accidentally included one too many commits in an interactive rebase. It's fixed now and I also re-rebased on master to pick up the `url` dependency pinning.

crates/chain/src/persist.rs
```diff
@@ -0,0 +1,454 @@
...
+            ..Default::default()
+        }
+    }
+}
+
+/// A persistence backend for writing and loading changesets.
+///
+/// `C` represents the changeset; a datatype that records changes made to in-memory data structures
+/// that are to be persisted, or retrieved from persistence.
+pub trait Persist<C> {
```
evanlinjin: I prefer the old name better `PersistBackend`. `Persist` is a little more ambiguous since we have many types here that start with `Persist`.

notmandatory: I don't have a strong feeling either way so I renamed to `Persist[Async]` -> `PersistBackend[Async]`.  Also `StagedPersist[Async]` -> `StagedPersistBackend[Async]`

crates/chain/Cargo.toml
```diff
@@ -20,12 +20,15 @@ serde_crate = { package = "serde", version = "1", optional = true, features = ["
...
 [dev-dependencies]
 rand = "0.8"
 proptest = "1.2.0"
 
 [features]
-default = ["std", "miniscript"]
+default = ["std", "miniscript", "persist"]
 std = ["bitcoin/std", "miniscript?/std"]
 serde = ["serde_crate", "bitcoin/serde", "miniscript?/serde"]
+persist = ["miniscript"]
```
evanlinjin: I don't think we should have a `persist` feature. Instead, we disable `CombinedChangeSet` when we don't have `miniscript`.

crates/chain/src/persist.rs
```diff
@@ -0,0 +1,456 @@
...
+    type LoadError = P::LoadError;
+
+    fn write_changes(&mut self, changeset: &C) -> Result<(), Self::WriteError> {
+        self.inner.write_changes(changeset)
+    }
+
+    fn load_changes(&mut self) -> Result<Option<C>, Self::LoadError> {
+        self.inner.load_changes()
+    }
+}
```
evanlinjin: I don't think the staged area should impl `PersistBackend`. The `load_changes` method would not load everything pushed in `write_changes`. I guess we can change `load_changes` to merge on whatever is in `stage`, but I don't think that is helpful?

evanlinjin: It no longer makes sense to have `StagedPersistBackend` since we don't use it! The examples can do what we do in wallet (as discussed in DMs).

evanlinjin: Change of plan (as discussed in DMs). The main selling point of `StagedPersist..` is in the `commit`-esc methods. Since the `commit` method checks whether it's an empty changeset (does not persist empty changeset) and not taking from the staged error if writing has failed. We should still keep this, but have `StagedPersist::inner` as in external input (so we can use it in wallet).

evanlinjin (approved):
ACK c9d1b418ba06ae47688bc09523966639c4b0fe48

notmandatory (commented):
ACK 6323d55d9d6142bc2ef16a0c6173e0d6d1e4e408

I rebased to add one more small fix to the bdk_wallet README.md.

notmandatory:
Had to do one final force push after rebasing on master.

ValuedMammal:
ACK ec36c7eccacc8347568b77ed1dfc0c50a2524906

ValuedMammal (approved):

PR #1463, opened 2024-06-06
No descriptor ids in spk txout index

### LINKED ISSUE #1459: `KeychainTxOutIndex` range-based methods are broken
**Describe the bug**

It turns out that having the internal `SpkTxOutIndex` use `DescriptorId`s is causing issues when using range-based methods.

The problem is evident when we query outputs in a given keychain (`K`) range, where the range is across multiple keychains. The keychain (`K`) range query does not translate well since the `Ord` of `DescriptorId`s is not guaranteed to be one-to-one with the `Ord` of keychains (`K`).

**Potential fixes**  

The first solution (the one @LLFourn is working on) is to use keychain (`K`) with the internal `SpkTxOutIndex`. To maintain the "consistency", we disallow reassigning keychains and assigning the same descriptor to multiple keychains. The `ChangeSet::append` method will ignore keychain-reassignments and assigning the same descriptor to another keychain. This is the solution with the simplest codebase. Downsides is it disallows certain usecases (which we aren't even sure if users will use) - i.e. assigning the same descriptor to multiple keychains, and replacing the descriptor for a keychain.

The alternative is to continue to have the internal `SpkTxOutIndex` use `DescriptorId`s and when we range keychains, we first get corresponding `DescriptorId`s from `KeychainTxOutIndex::keychains_to_descriptor_ids` and range for individual `DescriptorId`s when querying the internal `SpkTxOutIndex`. This does introduce even more code, however it allows us complete flexibility and how changesets are appended will make more sense (in my opinion).

LLFourn:
Fixes #1459

This reverts part of the changes in #1203. There the `SpkTxOutIndex<(K,u32)>` was changed to `SpkTxOutIndex<(DescriptorId, u32>)`. This led to a complicated translation logic in  `KeychainTxOutIndex` (where the API is based on `K`) to transform calls to it to calls to the underlying `SpkTxOutIndex` (which now indexes by `DescriptorId`). The translation layer was broken when it came to translating range queries from the `KeychainTxOutIndex`. My solution was just to revert this part of the change and remove the need for a translation layer (almost) altogether. A thin translation layer remains to ensure that un-revealed spks are filtered out before being returned from the `KeychainTxOutIndex` methods.

I feel like this PR could be extended to include a bunch of ergonomics improvements that are easier to implement now. But I think that's the point of https://github.com/bitcoindevkit/bdk/pull/1451 so I held off and should probably go and scope creep that one instead.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,2 +1,1 @@
 msrv="1.63.0"
-type-complexity-threshold = 275
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,1 +1,2 @@
 msrv="1.63.0"
+type-complexity-threshold = 275
--- a/crates/chain/src/keychain.rs
+++ b/crates/chain/src/keychain.rs
@@ -12,7 +12,7 @@
 
 #[cfg(feature = "miniscript")]
 mod txout_index;
-use bitcoin::Amount;
+use bitcoin::{Amount, ScriptBuf};
 #[cfg(feature = "miniscript")]
 pub use txout_index::*;
 
@@ -48,6 +48,11 @@ impl Balance {
         self.confirmed + self.trusted_pending + self.untrusted_pending + self.immature
     }
 }
+
+/// A tuple of keychain index and corresponding [`ScriptBuf`].
+pub type Indexed<T> = (u32, T);
+/// A tuple of keychain, index and the [`ScriptBuf`] derived at that location.
+pub type KeychainIndexed<K, T> = ((K, u32), T);
 
 impl core::fmt::Display for Balance {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
--- a/crates/chain/src/keychain.rs
+++ b/crates/chain/src/keychain.rs
@@ -49,9 +49,9 @@ impl Balance {
     }
 }
 
-/// A tuple of keychain index and corresponding [`ScriptBuf`].
+/// A tuple of keychain index and `T` representing the indexed value.
 pub type Indexed<T> = (u32, T);
-/// A tuple of keychain, index and the [`ScriptBuf`] derived at that location.
+/// A tuple of keychain `K`, derivation index (`u32`) and a `T` associated with them.
 pub type KeychainIndexed<K, T> = ((K, u32), T);
 
 impl core::fmt::Display for Balance {
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -15,87 +15,6 @@ use core::{
 use super::*;
 use crate::Append;
 
-/// Represents updates to the derivation index of a [`KeychainTxOutIndex`].
-/// It maps each keychain `K` to a descriptor and its last revealed index.
-///
-/// It can be applied to [`KeychainTxOutIndex`] with [`apply_changeset`]. [`ChangeSet] are
-/// monotone in that they will never decrease the revealed derivation index.
-///
-/// [`KeychainTxOutIndex`]: crate::keychain::KeychainTxOutIndex
-/// [`apply_changeset`]: crate::keychain::KeychainTxOutIndex::apply_changeset
-#[derive(Clone, Debug, PartialEq)]
-#[cfg_attr(
-    feature = "serde",
-    derive(serde::Deserialize, serde::Serialize),
-    serde(
-        crate = "serde_crate",
-        bound(
-            deserialize = "K: Ord + serde::Deserialize<'de>",
-            serialize = "K: Ord + serde::Serialize"
-        )
-    )
-)]
-#[must_use]
-pub struct ChangeSet<K> {
-    /// Contains the keychains that have been added and their respective descriptor
-    pub keychains_added: BTreeMap<K, Descriptor<DescriptorPublicKey>>,
-    /// Contains for each descriptor_id the last revealed index of derivation
-    pub last_revealed: BTreeMap<DescriptorId, u32>,
-}
-
-impl<K: Ord> Append for ChangeSet<K> {
-    /// Append another [`ChangeSet`] into self.
-    ///
-    /// For each keychain in `keychains_added` in the given [`ChangeSet`]:
-    /// If the keychain already exist with a different descriptor, we overwrite the old descriptor.
-    ///
-    /// For each `last_revealed` in the given [`ChangeSet`]:
-    /// If the keychain already exists, increase the index when the other's index > self's index.
-    fn append(&mut self, other: Self) {
-        for (new_keychain, new_descriptor) in other.keychains_added {
-            if !self.keychains_added.contains_key(&new_keychain)
-                // FIXME: very inefficient
-                && self
-                    .keychains_added
-                    .values()
-                    .all(|descriptor| descriptor != &new_descriptor)
-            {
-                self.keychains_added.insert(new_keychain, new_descriptor);
-            }
-        }
-
-        // for `last_revealed`, entries of `other` will take precedence ONLY if it is greater than
-        // what was originally in `self`.
-        for (desc_id, index) in other.last_revealed {
-            use crate::collections::btree_map::Entry;
-            match self.last_revealed.entry(desc_id) {
-                Entry::Vacant(entry) => {
-                    entry.insert(index);
-                }
-                Entry::Occupied(mut entry) => {
-                    if *entry.get() < index {
-                        entry.insert(index);
-                    }
-                }
-            }
-        }
-    }
-
-    /// Returns whether the changeset are empty.
-    fn is_empty(&self) -> bool {
-        self.last_revealed.is_empty() && self.keychains_added.is_empty()
-    }
-}
-
-impl<K> Default for ChangeSet<K> {
-    fn default() -> Self {
-        Self {
-            last_revealed: BTreeMap::default(),
-            keychains_added: BTreeMap::default(),
-        }
-    }
-}
-
 /// The default lookahead for a [`KeychainTxOutIndex`]
 pub const DEFAULT_LOOKAHEAD: u32 = 25;
 
@@ -105,6 +24,10 @@ pub const DEFAULT_LOOKAHEAD: u32 = 25;
 /// A single keychain is a chain of script pubkeys derived from a single [`Descriptor`]. Keychains
 /// are identified using the `K` generic. Script pubkeys are identified by the keychain that they
 /// are derived from `K`, as well as the derivation index `u32`.
+///
+/// There is a strict 1-to-1 relationship between descriptors and keychains. Each keychain has one
+/// and only one descriptor and each descriptor has one and only one keychain. The
+/// [`insert_descriptor`] method will return an error if you try and violate this invariant.
 ///
 /// # Revealed script pubkeys
 ///
@@ -177,18 +100,19 @@ pub const DEFAULT_LOOKAHEAD: u32 = 25;
 /// [`Ord`]: core::cmp::Ord
 /// [`SpkTxOutIndex`]: crate::spk_txout_index::SpkTxOutIndex
 /// [`Descriptor`]: crate::miniscript::Descriptor
-/// [`reveal_to_target`]: KeychainTxOutIndex::reveal_to_target
-/// [`reveal_next_spk`]: KeychainTxOutIndex::reveal_next_spk
-/// [`revealed_keychain_spks`]: KeychainTxOutIndex::revealed_keychain_spks
-/// [`revealed_spks`]: KeychainTxOutIndex::revealed_spks
-/// [`index_tx`]: KeychainTxOutIndex::index_tx
-/// [`index_txout`]: KeychainTxOutIndex::index_txout
-/// [`new`]: KeychainTxOutIndex::new
-/// [`unbounded_spk_iter`]: KeychainTxOutIndex::unbounded_spk_iter
-/// [`all_unbounded_spk_iters`]: KeychainTxOutIndex::all_unbounded_spk_iters
-/// [`outpoints`]: KeychainTxOutIndex::outpoints
-/// [`txouts`]: KeychainTxOutIndex::txouts
-/// [`unused_spks`]: KeychainTxOutIndex::unused_spks
+/// [`reveal_to_target`]: Self::reveal_to_target
+/// [`reveal_next_spk`]: Self::reveal_next_spk
+/// [`revealed_keychain_spks`]: Self::revealed_keychain_spks
+/// [`revealed_spks`]: Self::revealed_spks
+/// [`index_tx`]: Self::index_tx
+/// [`index_txout`]: Self::index_txout
+/// [`new`]: Self::new
+/// [`unbounded_spk_iter`]: Self::unbounded_spk_iter
+/// [`all_unbounded_spk_iters`]: Self::all_unbounded_spk_iters
+/// [`outpoints`]: Self::outpoints
+/// [`txouts`]: Self::txouts
+/// [`unused_spks`]: Self::unused_spks
+/// [`insert_descriptor`]: Self::insert_descriptor
 #[derive(Clone, Debug)]
 pub struct KeychainTxOutIndex<K> {
     inner: SpkTxOutIndex<(K, u32)>,
@@ -879,19 +803,18 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
             .collect()
     }
 
-    /// Applies the derivation changeset to the [`KeychainTxOutIndex`], as specified in the
-    /// [`ChangeSet::append`] documentation:
-    /// - Extends the number of derived scripts per keychain
-    /// - Adds new descriptors introduced
-    /// - If a descriptor is introduced for a keychain that already had a descriptor, overwrites
-    /// the old descriptor
+    /// Applies the `ChangeSet<K>` to the [`KeychainTxOutIndex<K>`]
+    ///
+    /// Keychains added by the `keychains_added` field of `ChangeSet<K>` respect the one-to-one
+    /// keychain <-> descriptor invariant by silently ignoring attempts to violate it (but will
+    /// panic if `debug_assertions` are enabled).
     pub fn apply_changeset(&mut self, changeset: ChangeSet<K>) {
         let ChangeSet {
             keychains_added,
             last_revealed,
         } = changeset;
         for (keychain, descriptor) in keychains_added {
-            let _ = self.insert_descriptor(keychain, descriptor);
+            let _ignore_invariant_violation = self.insert_descriptor(keychain, descriptor);
         }
 
         for (&desc_id, &index) in &last_revealed {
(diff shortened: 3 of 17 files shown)


crates/chain/src/keychain/txout_index.rs
```diff
@@ -50,9 +51,17 @@ impl<K: Ord> Append for ChangeSet<K> {
...
+            if !self.keychains_added.contains_key(&new_keychain)
+                // FIXME: very inefficient
+                && self
+                    .keychains_added
+                    .values()
+                    .all(|descriptor| descriptor != &new_descriptor)
+            {
+                self.keychains_added.insert(new_keychain, new_descriptor);
+            }
+        }
```
evanlinjin: Can we have a simple comment explaining why we are doing this?
I.e. We disallow re-assigning a keychain's descriptor or assigning the same descriptor to multiple keychains.

clippy.toml
```diff
@@ -1 +1,2 @@
 msrv="1.63.0"
+type-complexity-threshold = 275
```
notmandatory: 🔧 Since our types really are too complex can we just add a new `IndexSpk` type alias to fix the problem instead? It would also improve the docs for other APIs in the modules that also use `(u32,ScriptBuf)`. See my proposed patch here (feel free to cherry-pick or copy paste): https://github.com/notmandatory/bdk/commit/e15565571dd2d55d54c5d31de1042bbcc381daf4

We don't need to use the type outside of the `bdk_chain` crate but it shouldn't hurt since the other crates will always depend on chain.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -50,9 +51,17 @@ impl<K: Ord> Append for ChangeSet<K> {
...
-        // Refer to https://github.com/rust-lang/rust/issues/34666#issuecomment-675658420
-        self.keychains_added.extend(other.keychains_added);
+        for (new_keychain, new_descriptor) in other.keychains_added {
+            if !self.keychains_added.contains_key(&new_keychain)
+                // FIXME: very inefficient
+                && self
+                    .keychains_added
+                    .values()
+                    .all(|descriptor| descriptor != &new_descriptor)
+            {
```
notmandatory: 🔧 Would this work as a more efficient single pass, short-circuiting way to do this?
```suggestion
            if self.keychains_added.iter().all(|(keychain, descriptor)| {
                keychain != &new_keychain && descriptor != &new_descriptor
            }) {
```

crates/chain/src/keychain/txout_index.rs
```diff
@@ -112,22 +122,24 @@ const DEFAULT_LOOKAHEAD: u32 = 25;
...
+/// require holding a reference to the index.
 ///
 /// Call [`unbounded_spk_iter`] to get an unbounded script pubkey iterator for a given keychain.
 /// Call [`all_unbounded_spk_iters`] to get unbounded script pubkey iterators for all keychains.
 ///
 /// # Change sets
 ///
-/// Methods that can update the last revealed index or add keychains will return [`super::ChangeSet`] to report
+/// Methods that can update the last revealed index or add keychains will return [`ChangeSet`] to report
 /// these changes. This can be persisted for future recovery.
```
notmandatory: ⛏️  Would be good to make this recommendation stronger here.
```suggestion
/// these changes. This [`ChangeSet`] should be persisted for future recovery.
```

crates/chain/src/keychain/txout_index.rs
```diff
@@ -206,22 +190,18 @@ const DEFAULT_LOOKAHEAD: u32 = 25;
 /// [`unused_spks`]: KeychainTxOutIndex::unused_spks
 #[derive(Clone, Debug)]
 pub struct KeychainTxOutIndex<K> {
-    inner: SpkTxOutIndex<(DescriptorId, u32)>,
+    inner: SpkTxOutIndex<(K, u32)>,
     // keychain -> (descriptor, descriptor id) map
```
notmandatory: 🔧  needs to match the code change.
```suggestion
    // keychain -> descriptor id map
```

crates/chain/src/keychain/txout_index.rs
```diff
@@ -462,66 +408,82 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
     ///
     /// Adding a descriptor means you will be able to derive new script pubkeys under it
     /// and the txout index will discover transaction outputs with those script pubkeys.
     ///
-    /// When trying to add a keychain that already existed under a different descriptor, or a descriptor
-    /// that already existed with a different keychain, the old keychain (or descriptor) will be
-    /// overwritten.
+    /// keychain <-> descriptor is a one-to-one mapping -- in `--release` this method will ignore calls that try to
+    /// associate a keychain with the descriptor of another keychain or to re-assign the keychain to
+    /// new descriptor. If `debug_assertions` are enabled then it will panic.
```
notmandatory: 🔧 This code now throws an error and should not panic, docs need to be updated to reflect new behavior.
```suggestion
    /// The keychain <-> descriptor relationship is a one-to-one mapping that can not be changed. This method 
    /// will throw an error if you try to associate a new keychain with an already assigned descriptor or to re-assign 
    /// an existing keychain to a new descriptor.
```

notmandatory (changes requested):
LGTM but I have a few suggestions before ACKing.

crates/chain/src/keychain/txout_index.rs
LLFourn: No this code looks wrong. It's meant to check that no descriptor already in the changeset is the same as the one we're adding. The correct way to do this is a bijective map struct i.e. a BTreeMap one way and HashMap the other way. This might be handy to have in general but I guess it belongs in another PR.

clippy.toml
LLFourn: I made another commit to take this idea further: 4eb1e288a9362803b034590e0c56c8cf9cf8b0c2

There's still a bit to be done to have good ergonomics but I think that's better in followup.

I made the spks owned like in #1341

crates/chain/src/keychain/txout_index.rs
```diff
@@ -233,36 +214,39 @@ impl<K> Default for KeychainTxOutIndex<K> {
...
 
     fn initial_changeset(&self) -> Self::ChangeSet {
-        super::ChangeSet {
+        ChangeSet {
             keychains_added: self
                 .keychains()
                 .map(|(k, v)| (k.clone(), v.clone()))
                 .collect(),
-            last_revealed: self.last_revealed.clone(),
+            last_revealed: self.last_revealed.clone().into_iter().collect(),
```
evanlinjin, line 170: Nit:

```suggestion
            last_revealed: self.last_revealed.iter().map(|(did, i)| (*did, *i)).collect(),
```

crates/chain/src/keychain/txout_index.rs
```diff
@@ -462,66 +418,82 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+                .insert(desc_id, descriptor.clone());
+            self.keychains_to_descriptor_ids
+                .insert(keychain.clone(), desc_id);
+            self.descriptor_ids_to_keychains
+                .insert(desc_id, keychain.clone());
+            self.replenish_lookahead(&keychain, self.lookahead);
+            changeset
+                .keychains_added
+                .insert(keychain.clone(), descriptor);
+        } else {
```
evanlinjin, line 372: Nit: else branch not needed.

crates/chain/src/keychain/txout_index.rs
LLFourn: I added more docs and comment on this: 558a33a205d80c0e4bb3d3a24bdc7b826a90c8e0

crates/chain/src/keychain/txout_index.rs
```diff
@@ -548,114 +444,146 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-    fn replenish_lookahead(&mut self, keychain: &K, lookahead: u32) {
-        let descriptor_opt = self.keychains_to_descriptors.get(keychain).cloned();
-        if let Some((descriptor_id, descriptor)) = descriptor_opt {
-            let next_store_index = self.next_store_index(descriptor_id);
-            let next_reveal_index = self.last_revealed.get(&descriptor_id).map_or(0, |v| *v + 1);
+    fn replenish_lookahead_did(&mut self, did: DescriptorId) {
+        if let Some(keychain) = self.descriptor_ids_to_keychains.get(&did).cloned() {
+            self.replenish_lookahead(&keychain, self.lookahead);
+        }
+    }
```
evanlinjin: I don't think this method is necessary. We are only using it in 2 places; `Indexer::index_txout` and `apply_changeset`.

For `index_txout`, the keychain is already available for us (however we may need to clone the `inner.scan_txout` result because of the referencing rules of rust).

For `apply_changeset`, we can just query `Self::descriptor_ids_to_keychains` within the loop.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -548,114 +444,146 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+            // we skip ahead.
+            while current_keychain?.0 < keychain {
+                current_keychain = iter_last_revealed.next();
+            }
+            let (current_keychain, last_revealed) = current_keychain?;
+
+            if current_keychain == keychain && Some(*index) <= last_revealed {
+                break Some(((keychain.clone(), *index), spk.as_script()));
+            }
+        })
```
evanlinjin, line 519: I'm trying to understand the performance implications of this (since this complex-ish logic seems to be here for performance reasons, otherwise a `.filter` can work too).

I.e. let's say the caller is using a large lookahead (1000+) with a large amount of keychains. Would it be more performant to loop through all lookahead spks (except for the last keychain) - as we are doing here. Or, would it be more performant to do multiple `.range` calls on `inner.all_spks()` per keychain?

My argument is that performance is more of an issue for usecases with larger keychains-counts/spk-counts/lookahead-values.

evanlinjin (commented):
Thanks for working on this. The range-based method bugs definitely need to be fixed. However, I'm not sure if a total-revamp of the behavior is the correct way forward?

I'm okay with disallowing reassignment of a keychain's descriptor. However, I still think assigning the same descriptor to multiple keychains is important (for consistency). Let me explain.

Currently, our API allows adding *overlapping* descriptors. I.e. a non-wildcard descriptor which derives a spk that exists in a wildcard descriptor. There is just no good way to detect for this. With the changes presented in this PR, doing the above (which is impossible to detect) will cause undefined behavior, whereas the previous implementation handles this perfectly.

Therefore, for consistency, I think we should just enable both and scenarios and handle things consistently (as we are doing before).

evanlinjin:
However, I really like the API changes done in this PR. I.e.
* Returning a `Vec` of revealed spks (instead of iterator)
* Introducing `KeychainIndexed` and `Indexed` type parameters (however, can we think of a better name?)

ValuedMammal:
> There is just no good way to detect for this.

The way to detect it is to check the value returned by `self.inner.insert_spk()` instead of ignoring it.

crates/chain/src/keychain/txout_index.rs
notmandatory: My suggested code does check that neither the new `keychain` or `descriptor` being added exists in the set of `keychains_added` just as the original code, but should be O(n) instead of O(2n). 

But I'm ok leaving this as is for now and worrying about the efficiency later if it ever becomes a problem.

crates/chain/src/keychain/txout_index.rs
```diff
@@ -803,50 +665,67 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+            match self.reveal_next_spk(keychain) {
+                Some(((i, spk), change)) => {
+                    spks.push((i, spk));
+                    changeset.append(change);
+                }
+                None => break,
+            }
+        }
+
+        Some((spks, changeset))
```
ValuedMammal, line 667: The way it's written this will never return `None`, I think?

crates/chain/src/keychain/txout_index.rs
```diff
@@ -803,50 +665,67 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+            let did = self.keychains_to_descriptor_ids.get(keychain)?;
+            let descriptor = self.descriptor_ids_to_descriptors.get(did)?;
+            let spk = descriptor
+                .at_derivation_index(next_index)
+                .expect("already checked index is not too high")
+                .script_pubkey();
+            let _ = self.inner.insert_spk((keychain.clone(), next_index), spk);
+            self.last_revealed.insert(*did, next_index);
+            changeset.last_revealed.insert(*did, next_index);
+            self.replenish_lookahead(keychain, self.lookahead);
```
ValuedMammal: I'm thinking as long as we update `self.last_revealed` and then call `replenish_lookahead`, we shouldn't have to insert the spk here.

evanlinjin:
> > There is just no good way to detect for this.
> 
> The way to detect it is to check the value returned by `self.inner.insert_spk()` instead of ignoring it.

Yes, but that means we need to error when that happens (since it's non-viable to iterate through all of the wildcard-descriptor's spks). This means all methods that may need to derive more spks have to return an error. I don't think this is a reasonable API.

crates/chain/src/keychain/txout_index.rs
evanlinjin: What if the keychain does not exist?

crates/chain/src/keychain.rs
```diff
@@ -49,6 +49,11 @@ impl Balance {
     }
 }
 
+/// A tuple of keychain index and corresponding [`ScriptBuf`].
+pub type Indexed<T> = (u32, T);
```
ValuedMammal: maybe something like

```suggestion
/// A tuple of keychain index and an arbitrary `T` representing the indexed value.
pub type Indexed<T> = (u32, T);
```

crates/chain/src/keychain/txout_index.rs
LLFourn: Sorry I understand what you are saying now. `contains_key` is O(log(n)) so keeping the fast one fast and separate seemed sensible.

crates/chain/src/keychain/txout_index.rs
LLFourn: why is it not needed?

crates/chain/src/keychain/txout_index.rs
LLFourn: I think the method is useful to explicitly make the choice to silently ignore the corresponding `did` or keychain not existing. Doing the lookups in place means me or someone else is likely to just do `.expect("invariant")` in that case without understanding context and reviewers are unlikely to notice it.

I take your point that in one place the keychain already exists. I refactored the methods to take advantage of that.

crates/chain/src/keychain/txout_index.rs
LLFourn: Indeed this is written so it hits the optimal time complexity in the number of keychains.

Let $k$ is the number of keychains total, $s$ is the average number of spks per keychain and $h$ is the lookahead value. 

Let's consider the worst case performance when the query matches all keychains and $h = 0$. For the code above, there are two map queries which take negligible time and then we iterate through all the spks of every matched keychain so the time is dominated by the $sk$ spks you iterate through.

For the previous implementation where you do two lookups you have to do $sk$ iterations but you have another term that could take more time: $k\log(sk)$ time to do the btree map lookups which domainates when $\log(sk) > s$. The bottom line is that keeping everything other than $k$ constant the complexity is of PR impl is $O(k)$ while the previous impl was $O(k\log(k))$ so it's a significant improvement in theory.

Setting $h > 0$ it becomes $k(s+h)$ (this PR) vs $k\log(k(s+h))$ (prev). In the case that $h$ dominates $s$ which is maximally advantages the former implementation you get $kh$ vs $k(\log(k) + \log(h))$. Therefore the prev impl would always be faster where the lookahead dominates the actual number of spks. To improve performance for applications where $h >> s$ I'd say we should simply add a revleaed spks cache BTreeMap.

LLFourn:
> With the changes presented in this PR, doing the above (which is impossible to detect) will cause undefined behavior

The two claims made here are wrong in my view:

1. **impossible to detect**: Figuring out if two descriptors can produce the same spk is not a hard problem of computer science. I bet you can come up with a pretty good algorithm and implementation for this in less than a day. You can do a decent job of it by just by replacing the definite derivation indices with wildcards of any descriptor your adding and using that as a key in a lookup. I really don't think you should do this -- it's not worth your time because checking whether they're equal and documenting that the thing should not be used that way is good enough. This is just the place to solve the problem. Making the internals here complicated is certainly not.
2. **undefined behavior**: The first first keychain to derive that descriptor will be the owner of it. To me this seems like a simple, good and well defined behavior. Perhaps you mean it hasn't been defined in terms of writing a test. Feel free to do so.

update: I improved docs around this in d2629cbc320192e846d4c6dd9b56c7cc18764d3c also mentioning that it's up to the user not to insert overlapping descriptors.

crates/chain/src/keychain/txout_index.rs
LLFourn: I think I might try and make an Error for the keychain not exisitng.

LLFourn: (nope it is fine as it is -- and evan is right it will return `None` when there is not keychain).

crates/chain/src/keychain/txout_index.rs
LLFourn: Ah nice catch. Hmm I would then change the name of the `lookahead` functions so that they are not just for lookaheads.

see 1b6fe8cb0ae0104dac81730e34972804c04114b5

evanlinjin:
I've been thinking about `KeychainTxOutIndex` overnight, and the direction that this PR takes us is growing on me.

The keychain `K` is useful for the lifetime of `KeychainTxOutIndex` so that we can do range queries using `K` or `(K, u32)` for spks & indexed outputs.

For persisting, it's a more accurate representation to persist the `last_revealed_index` with the corresponding descriptor (instead of the keychain `K`). It's also handy if the user wishes to reconstruct the `KeychainTxOutIndex` later with a different set of keychains.

In addition, I was in discussion with @LLFourn yesterday. He suggested that we should consider NOT persisting descriptors via the `keychain::ChangeSet` since some users do not care about persisting the descriptor here.

evanlinjin:
Steps after this PR (not in any specific order):
* Rename generic `K` (keychain) to `L` (label). Label represents how we want to "label" the keychains for the lifetime of `KeychainTxOutIndex`.
* Introduce `trait Keychain` that produces `KeychainId` (which replaces `DescriptorId`) and a method that produces spks. A keychain always produces the same spks.
* `KeychainId` is created from a hash of the concatenation of 0th derived spk and a flag for wildcard/non-wildcard. This makes the `KeychainId` better for uniquely-identifying a keychain (than the hash of the miniscript descriptor).

crates/chain/src/keychain/txout_index.rs
evanlinjin: Oh it is needed sorry.

evanlinjin:
I'm tempted to remove all instances where we prematurely collect, but can leave it for another PR.

evanlinjin (approved):
self-ACK 3070b2dfaf7b48e4280dc2c5805fa76b5058245c

crates/chain/src/keychain/txout_index.rs
ValuedMammal: You could add a line to check `self.keychain_to_descriptor_id`

```diff
     pub fn reveal_to_target(
         &mut self,
         keychain: &K,
         target_index: u32,
     ) -> Option<(Vec<Indexed<ScriptBuf>>, ChangeSet<K>)> {
+        let _ = self.keychain_to_descriptor_id.get(keychain)?;
         let mut changeset = ChangeSet::default();
+
         let mut spks: Vec<Indexed<ScriptBuf>> = vec![];
         while let Some((i, new)) = self.next_index(keychain) {
             if !new || i > target_index {
                 break;
             }
             match self.reveal_next_spk(keychain) {
                 Some(((i, spk), change)) => {
                     spks.push((i, spk));
                     changeset.append(change);
                 }
                 None => break,
             }
         }
 
         Some((spks, changeset))
     }
```

ValuedMammal (commented):
ACK 3070b2dfaf7b48e4280dc2c5805fa76b5058245c

plus an additional comment about `reveal_to_target`

notmandatory (approved):
tACK 3070b2dfaf7b48e4280dc2c5805fa76b5058245c

I also manually tested the examples with signet descriptors and got back no errors and expected balance after scan and sync.

evanlinjin (approved):
ACK 8dd174479f9719309663ed979a5b4b86aca0a6e9

PR #1468, opened 2024-06-07
refactor(wallet)!: Use `Weight` type instead of `usize`

### LINKED ISSUE #1466: feat: use new `Weight` type throughout the BDK codebase, instead of `usize` for weight units
**Describe the enhancement**  

In the new release of rust-bitcoin 0.32.0 the new Weight type has been introduced through `bitcoin-units` `0.1.1`, it's not being explicitly used in the PR https://github.com/bitcoindevkit/bdk/pull/1448, but it should be instead of using `.to_wu() as usize` conversions.

**Use case**  
 
Use the proper `Weight` type/API, instead of the current `usize`, for referring to weight units.

**Additional context**

oleonardolima:
fixes bitcoindevkit/bdk#1466 
depends on bitcoindevkit/bdk#1448 

### Description

This PR is a follow-up on top of bitcoindevkit/bdk#1448, and should be rebased and merged after it, it uses the rust-bitcoin `Weight` type instead of the current `usize` usage.

NOTE: ~~It also adds a new `MiniscriptError` variant, and remove the `.unwrap()` usage.~~ As suggested I'll address this on another issue bitcoindevkit/bdk_wallet#92, trying to reproduce the error first.

### Notes to the reviewers
It should be ready to review after bitcoindevkit/bdk#1448 gets merged.

### Changelog notice
- Change `WeightedUtxo` `satisfaction_weight` has been changed to use `Weight` type. 
- Change `TxBuilder` methods `add_foreign_utxo` and `add_foreign_utxo_with_sequence` to expect `Weight` as `satisfaction_weight` type.

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/coin_selection.rs
+++ b/crates/wallet/src/wallet/coin_selection.rs
@@ -52,11 +52,10 @@
 //!                 (&mut selected_amount, &mut additional_weight),
 //!                 |(selected_amount, additional_weight), weighted_utxo| {
 //!                     **selected_amount += weighted_utxo.utxo.txout().value.to_sat();
-//!                     **additional_weight += Weight::from_wu(
-//!                         (TxIn::default().segwit_weight().to_wu()
-//!                             + weighted_utxo.satisfaction_weight as u64)
-//!                             as u64,
-//!                     );
+//!                     **additional_weight += TxIn::default()
+//!                         .segwit_weight()
+//!                         .checked_add(weighted_utxo.satisfaction_weight)
+//!                         .expect("`Weight` addition should not cause an integer overflow");
 //!                     Some(weighted_utxo.utxo)
 //!                 },
 //!             )
@@ -344,10 +343,10 @@ fn select_sorted_utxos(
             |(selected_amount, fee_amount), (must_use, weighted_utxo)| {
                 if must_use || **selected_amount < target_amount + **fee_amount {
                     **fee_amount += (fee_rate
-                        * Weight::from_wu(
-                            TxIn::default().segwit_weight().to_wu()
-                                + weighted_utxo.satisfaction_weight as u64,
-                        ))
+                        * (TxIn::default()
+                            .segwit_weight()
+                            .checked_add(weighted_utxo.satisfaction_weight)
+                            .expect("`Weight` addition should not cause an integer overflow")))
                     .to_sat();
                     **selected_amount += weighted_utxo.utxo.txout().value.to_sat();
                     Some(weighted_utxo.utxo)
@@ -390,9 +389,10 @@ struct OutputGroup {
 impl OutputGroup {
     fn new(weighted_utxo: WeightedUtxo, fee_rate: FeeRate) -> Self {
         let fee = (fee_rate
-            * Weight::from_wu(
-                TxIn::default().segwit_weight().to_wu() + weighted_utxo.satisfaction_weight as u64,
-            ))
+            * (TxIn::default()
+                .segwit_weight()
+                .checked_add(weighted_utxo.satisfaction_weight)
+                .expect("`Weight` addition should not cause an integer overflow")))
         .to_sat();
         let effective_value = weighted_utxo.utxo.txout().value.to_sat() as i64 - fee as i64;
         OutputGroup {
@@ -774,7 +774,7 @@ mod test {
         ))
         .unwrap();
         WeightedUtxo {
-            satisfaction_weight: P2WPKH_SATISFACTION_SIZE,
+            satisfaction_weight: Weight::from_wu_usize(P2WPKH_SATISFACTION_SIZE),
             utxo: Utxo::Local(LocalOutput {
                 outpoint,
                 txout: TxOut {
@@ -834,7 +834,7 @@ mod test {
         let mut res = Vec::new();
         for i in 0..utxos_number {
             res.push(WeightedUtxo {
-                satisfaction_weight: P2WPKH_SATISFACTION_SIZE,
+                satisfaction_weight: Weight::from_wu_usize(P2WPKH_SATISFACTION_SIZE),
                 utxo: Utxo::Local(LocalOutput {
                     outpoint: OutPoint::from_str(&format!(
                         "ebd9813ecebc57ff8f30797de7c205e3c7498ca950ea4341ee51a685ff2fa30a:{}",
@@ -865,7 +865,7 @@ mod test {
     fn generate_same_value_utxos(utxos_value: u64, utxos_number: usize) -> Vec<WeightedUtxo> {
         (0..utxos_number)
             .map(|i| WeightedUtxo {
-                satisfaction_weight: P2WPKH_SATISFACTION_SIZE,
+                satisfaction_weight: Weight::from_wu_usize(P2WPKH_SATISFACTION_SIZE),
                 utxo: Utxo::Local(LocalOutput {
                     outpoint: OutPoint::from_str(&format!(
                         "ebd9813ecebc57ff8f30797de7c205e3c7498ca950ea4341ee51a685ff2fa30a:{}",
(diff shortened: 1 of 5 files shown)


oleonardolima:
> Okay we need to create a ticket for what we should do when a caller creates a wallet with descriptors that are insatiable. Panicing with no clear reason when they try construct a tx is not the right behavior imo. Out of scope for this PR.

@evanlinjin mentioned it the https://github.com/bitcoindevkit/bdk/pull/1448#discussion_r1636641512. I'm moving the discussion to here as I've tried to address it here at [51e6f6d](https://github.com/bitcoindevkit/bdk/pull/1468/commits/51e6f6dbe91b80c7239c5367e6c654a749678962).

edit: also refer to upstream issue https://github.com/rust-bitcoin/rust-miniscript/issues/695

apoelstra:
Ok, moving my comment here:

Regarding the error path of `max_weight_to_satisfy` when you try to give it a (literally) unspendable descriptor like `wsh(0)` or something, I see a couple ways you can handle this:

* Try to propagate the error from the `max_weight_to_satisfy` call, treating it similarly to a transient "cannot spend" condition.
* Use `.unwrap_or(Weight::MAX)` which will have roughly the correct semantics without causing errors (though it could cause panics down the line if you add with it)
* Petition upstream to add a flag to [ExtParams](https://docs.rs/miniscript/latest/miniscript/miniscript/analyzable/struct.ExtParams.html) that would refuse to parse such scripts, then you can `unwrap`. Then users would be unable to import such scripts. Maybe you want them to be able to, though, for watch-only wallets?

oleonardolima:
Thanks for the suggestions! Although I've tried to do something similar to option (1), I'm now wondering about option (2), still need to think about the possible panics down the line 🤔

ValuedMammal:
I'm wondering whether it's even possible to create a bdk Wallet with an unsatisfiable descriptor

storopoli:
I think that what idea (2) meant was to use the `Weight::MAX` which is actually `u64::MAX` (Ref: https://docs.rs/bitcoin-units/0.1.1/src/bitcoin_units/weight.rs.html#35).
That might hit the `DEFAULT_MAX_FEE_RATE` which is `25_000` sats/vB (Ref: https://docs.rs/bitcoin/latest/src/bitcoin/psbt/mod.rs.html#133.

If we hit that, then we get a [`ExtractTxError`](https://docs.rs/bitcoin/latest/bitcoin/psbt/struct.Psbt.html#method.extract_tx_fee_rate_limit).

I think that's what it is meant as could cause panics.

What am I missing? I probably could be wrong here.
I am assuming that `Psbt::extract_tx_fee_rate_limit()` is being called somewhere down the line, right?

crates/wallet/src/wallet/error.rs
```diff
@@ -228,6 +228,14 @@ pub enum BuildFeeBumpError {
     IrreplaceableTransaction(Txid),
     /// Node doesn't have data to estimate a fee rate
     FeeRateUnavailable,
+    /// Happens when the descriptor is impossible to satisfy
+    MiniscriptError(miniscript::Error),
```
ValuedMammal: Can we write a test to show that it's possible to hit this error?

ValuedMammal:
I'm also ok with just fixing the `usize` to `Weight` stuff in this PR and creating a separate issue to decide how to handle an unsatisfiable descriptor

crates/wallet/src/wallet/error.rs
oleonardolima: Sure, I'll try working on that.

oleonardolima: I'll be working on that in another issue as you suggested, #1485

oleonardolima:
> I'm also ok with just fixing the `usize` to `Weight` stuff in this PR and creating a separate issue to decide how to handle an unsatisfiable descriptor

Yes, I removed the other commit. I only fixes the usage of `usize` now.

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -52,11 +52,10 @@
...
 //!                     **selected_amount += weighted_utxo.utxo.txout().value.to_sat();
-//!                     **additional_weight += Weight::from_wu(
-//!                         (TxIn::default().segwit_weight().to_wu()
-//!                             + weighted_utxo.satisfaction_weight as u64)
-//!                             as u64,
-//!                     );
+//!                     **additional_weight += TxIn::default()
+//!                         .segwit_weight()
+//!                         .checked_add(weighted_utxo.satisfaction_weight)
+//!                         .expect("valid `Weight`'s");
```
ValuedMammal: Note the expectation here is not just that the individual weights are valid but that summing them won't cause an overflow, maybe the expect message should clarify that.

But I guess the choice to use `checked_add` instead of `+` depends on whether we decide to use `Weight::MAX` in the case `max_weight_to_satisfy` returns an error. thoughts?

oleonardolima: > Note the expectation here is not just that the individual weights are valid but that summing them won't cause an overflow, maybe the expect message should clarify that.

Sure, I can improve the expectation message.

> But I guess the choice to use checked_add instead of + depends on whether we decide to use Weight::MAX in the case max_weight_to_satisfy returns an error. thoughts?

I don't think it strictly depends on that, but it's already as a safe/prevention for any sum that might cause overflow, in that scenario (use Weight::MAX) ends up it'll already be covered 🤔

ValuedMammal: Ok let's stick with `checked_add` and make the expect say something like "integer overflow occurred"

oleonardolima: I updated it to: "`Weight` addition should not cause an integer overflow", trying to follow the [common styles/conventions](https://doc.rust-lang.org/std/error/index.html#common-message-styles).

Let me know what you think.

ValuedMammal (approved):
ACK 729fec62d68b4824a6e36e64556798fd69caea4f

I know this is picky but since this change is API breaking, the commit should tell us that in conventional commits style, so something like

> refactor(wallet)!: change `WeightedUtxo` to use `Weight` type

In general the changelog should communicate any API changes so in this case the fact that `WeightedUtxo` is changed as well as the function signatures for `add_foreign_utxo{_with_sequence}`

Note to reviewers we didn't fix the `unwrap` for `max_weight_to_satisfy` but an issue has been created to discuss further.

oleonardolima:
> ACK [729fec6](https://github.com/bitcoindevkit/bdk/commit/729fec62d68b4824a6e36e64556798fd69caea4f)
> 
> I know this is picky but since this change is API breaking, the commit should tell us that in conventional commits style, so something like
> 
> > refactor(wallet)!: change `WeightedUtxo` to use `Weight` type
> 
> In general the changelog should communicate any API changes so in this case the fact that `WeightedUtxo` is changed as well as the function signatures for `add_foreign_utxo{_with_sequence}`
> 
> Note to reviewers we didn't fix the `unwrap` for `max_weight_to_satisfy` but an issue has been created to discuss further.

No worries, thanks for the suggestions. 🚀

storopoli (approved):
> I'm also ok with just fixing the `usize` to `Weight` stuff in this PR and creating a separate issue to decide how to handle an unsatisfiable descriptor

I second this. Do we have the issue open? I did not see linked to this PR.

Anyways, ACK 438cd4682d14b5f4c6c9c653c9f05ccaaecb815d

oleonardolima:
> I second this. Do we have the issue open? I did not see linked to this PR.

Yes, it's on the PR description, issue: https://github.com/bitcoindevkit/bdk_wallet/issues/92

notmandatory (approved):
ACK 438cd4682d14b5f4c6c9c653c9f05ccaaecb815d

Issue #1103, opened 2023-08-29
Backwards compatibility of `bdk::ChangeSet` data structure


evanlinjin:
**Describe the enhancement**  

The `bdk_wallet::ChangeSet` is what the persistence interfaces with. As suggested by @tnull, this structure should be backwards compatible (in case we make changes to this structure).

An easy solution is to make `bdk_wallet::ChangeSet` a enum:

```rust
#[non_exhaustive]
pub enum ChangeSet {
    V0(ChangeSet0),
    V1(ChangeSet1), // in the future this may exist
}

// We always convert to the latest version
impl From<ChangeSet0> for ChangeSet1 { /* TODO */ }
```

tnull:
Note that it would also be nice to consider whether even *forwards* compatibility could be guaranteed. It may make things a bit more complicated, but it's worth exploring, especially since these properties are transitive for any downstream users (e.g., LDK Node can only guarantee what BDK guarantees...).

> An easy solution is to make `bdk::ChangeSet` a enum:

Mh, but then you'd need to define explicit migration functions to allow user migrating from `V0` to `V1`? Another way would be to a) don't break compat for any existing fields and b) introduce any new fields as `Option`s so that users upgrading could deal with 'old' objects in which the fields are still `None`.

evanlinjin:
@tnull so you want an older version of BDK to understand changesets that a newer version of BDK outputs. Can you elaborate why this would be useful? Thank you

tnull:
> @tnull so you want an older version of BDK to understand changesets that a newer version of BDK outputs. Can you elaborate why this would be useful? Thank you

Forwards compatibility is really useful to allow downgrades for various reasons: say a user upgrades to a newer BDK version and then a bug is discovered. If the upgrade breaks serialization forwards compatibility they couldn't downgrade to get going again and would be stuck until BDK ships a fix. Similarly, it would be necessary to be a bit more forgiving when restoring from backups or when having multiple apps that operate on the same data set.

In LDK Node's case we for example plan to allow multi-device support synced via a VSS backend. Not having forwards compatibility of the serialized data would require our users to upgrade all apps and devices in lockstep as otherwise an outdated device would refuse to read any data written by the already upgraded devices. 

As said above, I'm aware that it may complicate things, but it's at least worth considering, i.e., it should be a dedicated design choice not to provide these guarantees.

LLFourn:
[...]

PR #1473, opened 2024-06-14
Remove `persist` submodule

evanlinjin:
### Description

@LLFourn suggested these changes which greatly simplifies the amount of code we have to maintain and removes the `async-trait` dependency. We remove `PersistBackend`, `PersistBackendAsync`, `StageExt` and `StageExtAsync` completely. Instead, we introduce `Wallet::take_staged(&mut self) -> Option<ChangeSet>`.

The original intention to keep a staging area (`StageExt`, `StageExtAsync`) is to enforce:
1. The caller will not persist an empty changeset.
2. The caller does not take the staged changeset if the database (`PersistBackend`) fails.

We achieve `1.` by returning `None` if the staged changeset is empty.

`2.` is not too important. The caller can try figure out what to do with the changeset if persisting to db fails. 

### Notes to the reviewers

I added a `take` convenience method to the `Append` trait. I thought it would be handy for the caller to do a staging area with this.

### Changelog notice

* Remove `persist` submodule from `bdk_chain`.
* Change `Wallet` to outsource it's persistence logic by introducing `Wallet::take_staged`.
* Add `take` convenience method to `Append` trait.

#### All Submissions:

#### New Features:


crates/chain/src/changeset.rs
```diff
@@ -0,0 +1,95 @@
+//! This module is home to the [`PersistBackend`] trait which defines the behavior of a data store
+//! required to persist changes made to BDK data structures.
+//!
```
ValuedMammal:
```suggestion
```

crates/wallet/src/wallet/mod.rs
```diff
@@ -2283,44 +2280,18 @@ impl Wallet {
...
-        persist_backend: &mut B,
-    ) -> Result<bool, B::WriteError>
-    where
-        B: bdk_chain::persist::PersistBackendAsync<ChangeSet> + Send + Sync,
-    {
-        let committed =
-            bdk_chain::persist::StageExtAsync::commit_to(&mut self.stage, persist_backend).await?;
-        Ok(committed.is_some())
+    /// Get a reference of the staged [`ChangeSet`] that are yet to be committed (if any).
+    pub fn staged(&self) -> Option<&ChangeSet> {
```
ValuedMammal, line 2284: I thought the previous implementation of this was fine too.

ValuedMammal (commented):
Approach ACK

crates/chain/src/tx_data_traits.rs
```diff
@@ -114,12 +114,21 @@ pub trait AnchorFromBlockPosition: Anchor {
 }
 
 /// Trait that makes an object appendable.
-pub trait Append {
+pub trait Append: Default {
```
notmandatory, line 117: 👍 This has the added benefit that it should solve #1103 by requiring all change set fields implement `Default`. So if we load a wallet data file from an older bdk version we should gracefully handle fields added in newer bdk versions since they must have a reasonable default value.

crates/chain/src/changeset.rs
notmandatory: Fixed.

notmandatory (approved):
ACK a0bf45bef1b53f8aec273f7fe89915f3d41974c0

notmandatory:
I pushed a commit to fix some docs that still mentioned `PersistBackend` and updated the `bdk_wallet` README example.

crates/chain/src/tx_data_traits.rs
evanlinjin: This doesn't really solve #1033. Backwards compatibility is only important for wallet, and we need the wallet's changeset to impl default already. Solving #1033 requires the serialized changeset to be deserializable with newer versions

crates/wallet/src/wallet/mod.rs
evanlinjin: You mean without the option?

evanlinjin: I think I was worried that people will try persist using this and then call `take_staged` only after saving to db succeeded.

evanlinjin (commented):
self-ACK a0bf45bef1b53f8aec273f7fe89915f3d41974c0

PR #1476, opened 2024-06-17
fix(wallet)!: Simplify `SignOptions` and improve finalization logic

### LINKED ISSUE #1461: wallet: `finalize_psbt` should clear BIP32 derivations
According to [BIP174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#input-finalizer) after building the final scriptWitness and scriptSig "all other data except the UTXO and unknown fields in the input key-value map should be cleared from the PSBT". This would include the BIP32 derivation paths for known pubkeys.

**To Reproduce**  
```rust
let (mut wallet, _) = get_funded_wallet(get_test_wpkh());
let addr = wallet.next_unused_address(KeychainKind::External).unwrap();
builder.drain_to(addr.script_pubkey()).drain_wallet();
assert!(wallet.sign(&mut psbt, SignOptions::default(),).unwrap());

assert!(psbt.inputs[0].bip32_derivation.is_empty());
assert!(psbt.outputs[0].bip32_derivation.is_empty());
```

**Additional context** 

I seem to recall @notmandatory mentioned there is a reason for having more granularity in choosing which fields should remain as opposed to just clearing the necessary fields by default. Is that still the case?

ValuedMammal:
Rather than comingle various `SignOptions` with the finalization step, we simply clear all fields when finalizing as per the PSBT spec in BIPs 174 and 371 which is more in line with user expectations.

### Notes to the reviewers

I chose to re-implement some parts of [`finalize_input`](https://docs.rs/miniscript/latest/src/miniscript/psbt/finalizer.rs.html#434) since it's fairly straightforward, see also https://github.com/bitcoindevkit/bdk/issues/1461#issuecomment-2171983426. I had to refactor some wallet tests but didn't go out of my way to add additional tests.

closes #1461 

### Changelog notice

- Removed field `remove_partial_sigs` from `SignOptions`
- Removed field `remove_taproot_extras` from `SignOptions`

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -40,6 +40,7 @@ use bitcoin::{
 use bitcoin::{consensus::encode::serialize, transaction, BlockHash, Psbt};
 use bitcoin::{constants::genesis_block, Amount};
 use core::fmt;
+use core::mem;
 use core::ops::Deref;
 use descriptor::error::Error as DescriptorError;
 use miniscript::psbt::{PsbtExt, PsbtInputExt, PsbtInputSatisfier};
@@ -1821,7 +1822,8 @@ impl Wallet {
 
     /// Finalize a PSBT, i.e., for each input determine if sufficient data is available to pass
     /// validation and construct the respective `scriptSig` or `scriptWitness`. Please refer to
-    /// [BIP174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#Input_Finalizer)
+    /// [BIP174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#Input_Finalizer),
+    /// and [BIP371](https://github.com/bitcoin/bips/blob/master/bip-0371.mediawiki)
     /// for further information.
     ///
     /// Returns `true` if the PSBT could be finalized, and `false` otherwise.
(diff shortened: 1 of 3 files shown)


storopoli (approved):
Concept ACK.

Nice trick with `mem::take`.

crates/wallet/src/wallet/mod.rs
```diff
@@ -1884,21 +1886,22 @@ impl Wallet {
...
+                            psbt_input.final_script_sig = if tmp_input.script_sig.is_empty() {
+                                None
+                            } else {
+                                Some(tmp_input.script_sig)
+                            };
+                            psbt_input.final_script_witness = if tmp_input.witness.is_empty() {
+                                None
+                            } else {
+                                Some(tmp_input.witness)
+                            };
```
evanlinjin: Nit:

```suggestion
                            if !tmp_input.script_sig.is_empty() {
                                psbt_input.final_script_sig = Some(tmp_input.script_sig);
                            }
                            if !tmp_input.witness.is_empty() {
                                psbt_input.final_script_witness = Some(tmp_input.witness);
                            }

```

evanlinjin (approved):
ACK ee8d410eb0f1feff30027323d8587dcb4148408e

Just a small nit lol

crates/wallet/src/wallet/mod.rs
ValuedMammal: thank you

evanlinjin (approved):
re-ACK 996605f2bf9440dd42647123a127c038253f0247

PR #1477, opened 2024-06-18
feat(wallet): add transactions_sort_by function

### LINKED ISSUE #794: Transaction Sorting and Querying
**Describe the enhancement**  
We'd like to be able to sort transactions at the DB layer and be able to paginate (and maybe in the future search) queries to the transactions.

**Use case**  
We want to display transactions to our customers in a sorted manner. We can do this by first querying all transactions and then sorting them after the fact, but this won't perform well once customers have a large number of transactions.

Currently we are thinking of sorting based on confirmation time for confirmed transactions and no sorting for unconfirmed transactions, but we may explore associating metadata with transactions (i.e. "initiation date" for when the transaction was initiated) and would want to be able to sort transactions based on that (this would help us sort unconfirmed transactions).

In the future, we may also allow customers to filter / search their transactions – ideally we could associate some sort of metadata here that we could filter / search by, but in a simpler use case could also imagine looking up date ranges (i.e. searching / grouping by confirmation time).

notmandatory:
### Description

Added new type alias `WalletTx` which represents a `CanonicalTx<'a, Arc<Transaction>, ConfirmationTimeHeightAnchor>` and new `Wallet::transactions_sort_by` that returns a `Vec<WalletTx>` sorted by the given compare function. 

### Notes to the reviewers

fixes bitcoindevkit/bdk#794 

### Changelog notice

* Add new type alias `WalletTx` which represents a `CanonicalTx<'a, Arc<Transaction>, ConfirmationTimeHeightAnchor>`.
* Add `Wallet::transactions_sort_by()` that returns a `Vec<WalletTx>` sorted by a given compare function. 

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -45,6 +45,7 @@ use bitcoin::{
 use bitcoin::{consensus::encode::serialize, transaction, BlockHash, Psbt};
 use bitcoin::{constants::genesis_block, Amount};
 use bitcoin::{secp256k1::Secp256k1, Weight};
+use core::cmp::Ordering;
 use core::fmt;
 use core::mem;
 use core::ops::Deref;
@@ -290,6 +291,9 @@ impl fmt::Display for ApplyBlockError {
 
 #[cfg(feature = "std")]
 impl std::error::Error for ApplyBlockError {}
+
+/// A `CanonicalTx` managed by a `Wallet`.
+pub type WalletTx<'a> = CanonicalTx<'a, Arc<Transaction>, ConfirmationBlockTime>;
 
 impl Wallet {
     /// Build a new single descriptor [`Wallet`].
@@ -1002,9 +1006,9 @@ impl Wallet {
         self.indexed_graph.index.sent_and_received(tx, ..)
     }
 
-    /// Get a single transaction from the wallet as a [`CanonicalTx`] (if the transaction exists).
+    /// Get a single transaction from the wallet as a [`WalletTx`] (if the transaction exists).
     ///
-    /// `CanonicalTx` contains the full transaction alongside meta-data such as:
+    /// `WalletTx` contains the full transaction alongside meta-data such as:
     /// * Blocks that the transaction is [`Anchor`]ed in. These may or may not be blocks that exist
     ///   in the best chain.
     /// * The [`ChainPosition`] of the transaction in the best chain - whether the transaction is
@@ -1018,13 +1022,13 @@ impl Wallet {
     /// # let wallet: Wallet = todo!();
     /// # let my_txid: bitcoin::Txid = todo!();
     ///
-    /// let canonical_tx = wallet.get_tx(my_txid).expect("panic if tx does not exist");
+    /// let wallet_tx = wallet.get_tx(my_txid).expect("panic if tx does not exist");
     ///
     /// // get reference to full transaction
-    /// println!("my tx: {:#?}", canonical_tx.tx_node.tx);
+    /// println!("my tx: {:#?}", wallet_tx.tx_node.tx);
     ///
     /// // list all transaction anchors
-    /// for anchor in canonical_tx.tx_node.anchors {
+    /// for anchor in wallet_tx.tx_node.anchors {
     ///     println!(
     ///         "tx is anchored by block of hash {}",
     ///         anchor.anchor_block().hash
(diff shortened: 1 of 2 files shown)


crates/wallet/src/wallet/mod.rs
```diff
@@ -1136,15 +1137,33 @@ impl Wallet {
...
+    /// # Example
+    ///
+    /// ```rust,no_run
+    /// # use bdk_wallet::Wallet;
+    /// # use bdk_wallet::wallet::WalletTx;
+    /// # let changeset = Default::default();
+    /// # let mut wallet = Wallet::load_from_changeset(changeset)?;
+    /// // Transactions by chain position: first unconfirmed then descending by confirmed height.
+    /// let sorted_txs:Vec<WalletTx> = wallet.transactions_sort_by(|tx1, tx2| tx2.chain_position.cmp(&tx1.chain_position));
+    /// # Ok::<(), anyhow::Error>(())
```
storopoli, line 1122: We are missing a closing ```
```suggestion
    /// # Ok::<(), anyhow::Error>(())
    /// ```
```

storopoli (commented):
A minor comment.

crates/wallet/src/wallet/mod.rs
```diff
@@ -1136,15 +1137,33 @@ impl Wallet {
...
+    pub fn transactions_sort_by<F>(&self, compare: F) -> Vec<WalletTx>
+    where
+        F: FnMut(&WalletTx, &WalletTx) -> Ordering,
+    {
+        let mut txs: Vec<WalletTx> = self.transactions().collect();
+        txs.sort_by(compare);
```
ValuedMammal: perf: Consider using `sort_unstable_by` ?

crates/wallet/src/wallet/mod.rs
notmandatory: oops! fixed.

crates/wallet/src/wallet/mod.rs
notmandatory: I didn't notice this, nice optimization.  Fixed.

ValuedMammal (approved):
ACK fb1599d50a38b37763a1246ba2c0de64587a150d

This looks good. I'm curious if users will also be interested in a list that includes transactions that may not be in the current active chain as was mentioned in https://github.com/bitcoindevkit/bdk/issues/794#issuecomment-2002656625

notmandatory:
> I'm curious if users will also be interested in a list that includes transactions that may not be in the current active chain as was mentioned in [#794 (comment)](https://github.com/bitcoindevkit/bdk/issues/794#issuecomment-2002656625)

I like the idea of being able to show TX that are not in the valid chain due to RBF or reorg for a future PR. The idea with this one is to provide only the basic sort functionality for lang bindings to expose.

evanlinjin (approved):
ACK fb1599d50a38b37763a1246ba2c0de64587a150d

notmandatory:
This only adds a new function and doesn't break any existing APIs so I'm moving this to the beta milestone so we can focus on breaking changes.

crates/wallet/src/wallet/mod.rs
```diff
@@ -1037,13 +1041,10 @@ impl Wallet {
     /// ```
     ///
     /// [`Anchor`]: bdk_chain::Anchor
-    pub fn get_tx(
-        &self,
-        txid: Txid,
-    ) -> Option<CanonicalTx<'_, Arc<Transaction>, ConfirmationTimeHeightAnchor>> {
+    pub fn get_tx(&self, txid: Txid) -> Option<WalletTx> {
```
oleonardolima, line 1052: @notmandatory I think the documentation needs an update now, referring to `WalletTx` instead, wdyt ?

oleonardolima (commented):
Looks good! I just left a single comment about the documentation.

oleonardolima:
Also, we'll still need to work on https://github.com/bitcoindevkit/bdk_wallet/issues/106 in the future, right ?

storopoli:
> Also, we'll still need to work on bitcoindevkit/bdk_wallet#106 in the future, right ?

Yes because this is a simple `collect()` into a `Vec` and then sorting the vec in the Wallet API.
bitcoindevkit/bdk_wallet#106 is in the Chain API.

evanlinjin:
@notmandatory I'm happy to merge this now tbh. It doesn't conflict or break anything.

reez:
Just adding a +1 for merging this PR 🚀

crates/wallet/src/wallet/mod.rs
notmandatory: Good catch, fixed.

notmandatory:
Rebased and fixed docs.

notmandatory:
Back by popular demand, merging this now instead of later.

reez:
woohoo!

PR #1478, opened 2024-06-18
Allow opting out of getting `LocalChain` updates with `FullScanRequest`/`SyncRequest` structures

evanlinjin:
Closes #1528

### Description

Some users use `bdk_esplora` to update `bdk_chain` structures *without* a `LocalChain`. ~~This PR introduces "low-level" methods to `EsploraExt` and `EsploraAsyncExt` which populates a `TxGraph` update with associated data.~~

We change `FullScanRequest`/`SyncRequest` to take in the `chain_tip` parameter as an option. Spk-based chain sources (`bdk_electrum` and `bdk_esplora`) will not fetch a chain-update if `chain_tip` is `None`, allowing callers to opt-out of receiving updates for `LocalChain`.

We change `FullScanRequest`/`SyncRequest` to have better ergonomics when inspecting the progress of syncing (refer to #1528).

We change `FullScanRequest`/`SyncRequest` to be constructed with a builder pattern. This is a better API since we separate request-construction and request-consumption.

Additionally, much of the `bdk_esplora` logic has been made more efficient (less calls to Esplora) by utilizing the `/tx/:txid` endpoint (`get_tx_info` method). This method returns the tx and tx_status in one call. The logic for fetching updates for outpoints has been reworked to support parallelism.

Documentation has also been updated.

### Notes to reviewers

This PR has evolved somewhat. Initially, we were adding more methods on `EsploraExt`/`EsploraAsyncExt` to make syncing/scanning more modular. However, it turns out we can just make the `chain_tip` parameter of the request structures optional. Since we are changing the request structures, we might as well go further and improve the ergonomics of the whole structures (refer to #1528). This is where we are at with this PR.

Unfortunately, the changes are now breaking. I think this is an okay tradeoff for the API improvements (before we get to stable).

### Changelog notice

* Change request structures in `bdk_chain::spk_client` to be constructed via a builder pattern, make providing a `chain_tip` optional, and have better ergonomics for inspecting progress while syncing.
* Change `bdk_esplora` to be more efficient by reducing the number of calls via the `/tx/:txid` endpoint. The logic for fetching outpoint updates has been reworked to support parallelism.
* Change `bdk_esplora` to always add prev-txouts to the `TxGraph` update.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/spk_client.rs
+++ b/crates/chain/src/spk_client.rs
@@ -110,14 +110,7 @@ impl<K: Clone + Ord + core::fmt::Debug + Send + Sync> SyncRequestBuilder<(K, u32
         indexer: &crate::indexer::keychain_txout::KeychainTxOutIndex<K>,
         spk_range: impl core::ops::RangeBounds<K>,
     ) -> Self {
-        use crate::alloc::borrow::ToOwned;
-        use crate::alloc::vec::Vec;
-        self.spks_with_labels(
-            indexer
-                .revealed_spks(spk_range)
-                .map(|(i, spk)| (i, spk.to_owned()))
-                .collect::<Vec<_>>(),
-        )
+        self.spks_with_labels(indexer.revealed_spks(spk_range))
     }
 
     /// Add [`Script`]s that are revealed by the `indexer` but currently unused.
@@ -125,14 +118,7 @@ impl<K: Clone + Ord + core::fmt::Debug + Send + Sync> SyncRequestBuilder<(K, u32
         self,
         indexer: &crate::indexer::keychain_txout::KeychainTxOutIndex<K>,
     ) -> Self {
-        use crate::alloc::borrow::ToOwned;
-        use crate::alloc::vec::Vec;
-        self.spks_with_labels(
-            indexer
-                .unused_spks()
-                .map(|(i, spk)| (i, spk.to_owned()))
-                .collect::<Vec<_>>(),
-        )
+        self.spks_with_labels(indexer.unused_spks())
     }
 }
 
--- a/crates/chain/src/spk_client.rs
+++ b/crates/chain/src/spk_client.rs
@@ -89,11 +89,11 @@ impl SyncProgress {
 
 /// Builds a [`SyncRequest`].
 #[must_use]
-pub struct SyncRequestBuilder<SpkLabel = ()> {
-    inner: SyncRequest<SpkLabel>,
+pub struct SyncRequestBuilder<I = ()> {
+    inner: SyncRequest<I>,
 }
 
-impl<SpkLabel> Default for SyncRequestBuilder<SpkLabel> {
+impl<I> Default for SyncRequestBuilder<I> {
     fn default() -> Self {
         Self {
             inner: Default::default(),
@@ -110,7 +110,7 @@ impl<K: Clone + Ord + core::fmt::Debug + Send + Sync> SyncRequestBuilder<(K, u32
         indexer: &crate::indexer::keychain_txout::KeychainTxOutIndex<K>,
         spk_range: impl core::ops::RangeBounds<K>,
     ) -> Self {
-        self.spks_with_labels(indexer.revealed_spks(spk_range))
+        self.spks_with_indexes(indexer.revealed_spks(spk_range))
     }
 
     /// Add [`Script`]s that are revealed by the `indexer` but currently unused.
@@ -118,18 +118,18 @@ impl<K: Clone + Ord + core::fmt::Debug + Send + Sync> SyncRequestBuilder<(K, u32
         self,
         indexer: &crate::indexer::keychain_txout::KeychainTxOutIndex<K>,
     ) -> Self {
-        self.spks_with_labels(indexer.unused_spks())
+        self.spks_with_indexes(indexer.unused_spks())
     }
 }
 
 impl SyncRequestBuilder<()> {
     /// Add [`Script`]s that will be synced against.
     pub fn spks(self, spks: impl IntoIterator<Item = ScriptBuf>) -> Self {
-        self.spks_with_labels(spks.into_iter().map(|spk| ((), spk)))
+        self.spks_with_indexes(spks.into_iter().map(|spk| ((), spk)))
     }
 }
 
-impl<SpkLabel> SyncRequestBuilder<SpkLabel> {
+impl<I> SyncRequestBuilder<I> {
     /// Set the initial chain tip for the sync request.
     ///
     /// This is used to update [`LocalChain`](crate::local_chain::LocalChain).
@@ -138,7 +138,7 @@ impl<SpkLabel> SyncRequestBuilder<SpkLabel> {
         self
     }
 
-    /// Add [`Script`]s coupled with an associated label that will be synced against.
+    /// Add [`Script`]s coupled with associated indexes that will be synced against.
     ///
     /// # Example
     ///
@@ -158,28 +158,25 @@ impl<SpkLabel> SyncRequestBuilder<SpkLabel> {
     ///
     /// /* Assume that the caller does more mutations to the `indexer` here... */
     ///
-    /// // Reveal spks for "descriptor_a", then build a sync request. Each spk will be labelled with
+    /// // Reveal spks for "descriptor_a", then build a sync request. Each spk will be indexed with
     /// // `u32`, which represents the derivation index of the associated spk from "descriptor_a".
     /// let (newly_revealed_spks, _changeset) = indexer
     ///     .reveal_to_target("descriptor_a", 21)
     ///     .expect("keychain must exist");
     /// let _request = SyncRequest::builder()
-    ///     .spks_with_labels(newly_revealed_spks)
+    ///     .spks_with_indexes(newly_revealed_spks)
     ///     .build();
     ///
     /// // Sync all revealed spks in the indexer. This time, spks may be derived from different
-    /// // keychains. Each spk will be labelled with `(&'static str, u32)` where `&'static str` is
+    /// // keychains. Each spk will be indexed with `(&'static str, u32)` where `&'static str` is
     /// // the keychain identifier and `u32` is the derivation index.
     /// let all_revealed_spks = indexer.revealed_spks(..);
     /// let _request = SyncRequest::builder()
-    ///     .spks_with_labels(all_revealed_spks)
+    ///     .spks_with_indexes(all_revealed_spks)
     ///     .build();
     /// # Ok::<_, bdk_chain::keychain_txout::InsertDescriptorError<_>>(())
     /// ```
-    pub fn spks_with_labels(
-        mut self,
-        spks: impl IntoIterator<Item = (SpkLabel, ScriptBuf)>,
-    ) -> Self {
+    pub fn spks_with_indexes(mut self, spks: impl IntoIterator<Item = (I, ScriptBuf)>) -> Self {
         self.inner.spks.extend(spks);
         self
     }
@@ -199,14 +196,14 @@ impl<SpkLabel> SyncRequestBuilder<SpkLabel> {
     /// Set the closure that will inspect every sync item visited.
     pub fn inspect<F>(mut self, inspect: F) -> Self
     where
-        F: FnMut(SyncItem<SpkLabel>, SyncProgress) + Send + 'static,
+        F: FnMut(SyncItem<I>, SyncProgress) + Send + 'static,
     {
         self.inner.inspect = Box::new(inspect);
         self
     }
 
     /// Build the [`SyncRequest`].
-    pub fn build(self) -> SyncRequest<SpkLabel> {
+    pub fn build(self) -> SyncRequest<I> {
         self.inner
     }
 }
--- a/crates/chain/src/spk_client.rs
+++ b/crates/chain/src/spk_client.rs
@@ -557,8 +557,8 @@ impl<'r, I> Iterator for SyncIter<'r, I, ScriptBuf> {
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let consumed = self.request.spks_consumed;
-        (consumed, Some(consumed))
+        let remaining = self.request.spks.len();
+        (remaining, Some(remaining))
     }
 }
 
@@ -570,8 +570,8 @@ impl<'r, I> Iterator for SyncIter<'r, I, Txid> {
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let consumed = self.request.txids_consumed;
-        (consumed, Some(consumed))
+        let remaining = self.request.txids.len();
+        (remaining, Some(remaining))
     }
 }
 
@@ -583,7 +583,7 @@ impl<'r, I> Iterator for SyncIter<'r, I, OutPoint> {
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let consumed = self.request.outpoints_consumed;
-        (consumed, Some(consumed))
+        let remaining = self.request.outpoints.len();
+        (remaining, Some(remaining))
     }
 }
--- a/crates/chain/src/spk_client.rs
+++ b/crates/chain/src/spk_client.rs
@@ -139,6 +139,43 @@ impl<SpkLabel> SyncRequestBuilder<SpkLabel> {
     }
 
     /// Add [`Script`]s coupled with an associated label that will be synced against.
+    ///
+    /// # Example
+    ///
+    /// Sync revealed script pubkeys obtained from a
+    /// [`KeychainTxOutIndex`](crate::keychain_txout::KeychainTxOutIndex).
+    ///
+    /// ```rust
+    /// # use bdk_chain::spk_client::SyncRequest;
+    /// # use bdk_chain::indexer::keychain_txout::KeychainTxOutIndex;
+    /// # use bdk_chain::miniscript::{Descriptor, DescriptorPublicKey};
+    /// # let secp = bdk_chain::bitcoin::secp256k1::Secp256k1::signing_only();
+    /// # let (descriptor_a,_) = Descriptor::<DescriptorPublicKey>::parse_descriptor(&secp, "tr([73c5da0a/86'/0'/0']xprv9xgqHN7yz9MwCkxsBPN5qetuNdQSUttZNKw1dcYTV4mkaAFiBVGQziHs3NRSWMkCzvgjEe3n9xV8oYywvM8at9yRqyaZVz6TYYhX98VjsUk/0/*)").unwrap();
+    /// # let (descriptor_b,_) = Descriptor::<DescriptorPublicKey>::parse_descriptor(&secp, "tr([73c5da0a/86'/0'/0']xprv9xgqHN7yz9MwCkxsBPN5qetuNdQSUttZNKw1dcYTV4mkaAFiBVGQziHs3NRSWMkCzvgjEe3n9xV8oYywvM8at9yRqyaZVz6TYYhX98VjsUk/1/*)").unwrap();
+    /// let mut indexer = KeychainTxOutIndex::<&'static str>::default();
+    /// indexer.insert_descriptor("descriptor_a", descriptor_a)?;
+    /// indexer.insert_descriptor("descriptor_b", descriptor_b)?;
+    ///
+    /// /* Assume that the caller does more mutations to the `indexer` here... */
+    ///
+    /// // Reveal spks for "descriptor_a", then build a sync request. Each spk will be labelled with
+    /// // `u32`, which represents the derivation index of the associated spk from "descriptor_a".
+    /// let (newly_revealed_spks, _changeset) = indexer
+    ///     .reveal_to_target("descriptor_a", 21)
+    ///     .expect("keychain must exist");
+    /// let _request = SyncRequest::builder()
+    ///     .spks_with_labels(newly_revealed_spks)
+    ///     .build();
+    ///
+    /// // Sync all revealed spks in the indexer. This time, spks may be derived from different
+    /// // keychains. Each spk will be labelled with `(&'static str, u32)` where `&'static str` is
+    /// // the keychain identifier and `u32` is the derivation index.
+    /// let all_revealed_spks = indexer.revealed_spks(..);
+    /// let _request = SyncRequest::builder()
+    ///     .spks_with_labels(all_revealed_spks)
+    ///     .build();
+    /// # Ok::<_, bdk_chain::keychain_txout::InsertDescriptorError<_>>(())
+    /// ```
     pub fn spks_with_labels(
         mut self,
         spks: impl IntoIterator<Item = (SpkLabel, ScriptBuf)>,
--- a/crates/chain/src/spk_client.rs
+++ b/crates/chain/src/spk_client.rs
@@ -1,186 +1,407 @@
 //! Helper types for spk-based blockchain clients.
-
 use crate::{
-    collections::BTreeMap, local_chain::CheckPoint, ConfirmationBlockTime, Indexed, TxGraph,
+    alloc::{boxed::Box, collections::VecDeque, vec::Vec},
+    collections::BTreeMap,
+    local_chain::CheckPoint,
+    ConfirmationBlockTime, Indexed, TxGraph,
 };
-use alloc::boxed::Box;
 use bitcoin::{OutPoint, Script, ScriptBuf, Txid};
-use core::marker::PhantomData;
+
+type InspectSync<I> = dyn FnMut(SyncItem<I>, SyncProgress) + Send + 'static;
+
+type InspectFullScan<K> = dyn FnMut(K, u32, &Script) + Send + 'static;
+
+/// An item reported to the [`inspect`](SyncRequestBuilder::inspect) closure of [`SyncRequest`].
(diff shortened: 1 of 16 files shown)


crates/esplora/src/lib.rs
```diff
@@ -1,19 +1,59 @@
...
+//! For a sync or full scan, the user receives relevant blockchain data and output updates for
+//! [`bdk_chain`] .
+//!
+//! # Low-Level Methods
+//!
+//! [`EsploraExt::sync`] and [`EsploraExt::full_scan`] returns updates which are *complete* and can
+//! be used directly to determine confirmation statuses of each transaction. This is because a
+//! [`LocalChain`] update is contained in the returned update structures. However, sometimes the
+//! caller wishes to use a custom [`ChainOracle`] implementation (something other than
+//! [`LocalChain`]). The following methods ONLY populates an update [`TxGraph`].
```
storopoli:
```suggestion
//! [`LocalChain`]). The following methods ONLY populates an update [`TxGraph`]:
```

crates/esplora/src/lib.rs
```diff
@@ -1,19 +1,59 @@
...
 //!
-//! Refer to [`example_esplora`] for a complete example.
+//! [`EsploraExt::sync`] is used to sync against a subset of wallet data. For example, transaction
+//! histories of revealed and unused script from the external (public) keychain and/or statuses of
+//! wallet-owned UTXOs and spending transactions from them. The policy of what to sync against can
+//! be customized.
+//!
+//! [`EsploraExt::full_scan`] is designed to be used when importing or restoring a keychain where
+//! the range of possibly used scripts is not known. In this case it is necessary to scan all
+//! keychain scripts until a number (the `stop gap`) of unused scripts is discovered.
```
storopoli:
```suggestion
//! keychain scripts until a number (the `stop_gap`) of unused scripts is discovered.
```

storopoli (commented):
Some high-level crate doc comments.
I might go over the code until the end of the weekend.

notmandatory:
Since this is a refactor and won't break the `bdk_wallet` API I'm not going to assign it a milestone until we get together for a team backlog prioritization call.

crates/esplora/src/blocking_ext.rs
```diff
@@ -112,6 +164,235 @@ impl EsploraExt for esplora_client::BlockingClient {
...
+            for handle in handles {
+                let (txid, resp) = handle.join().expect("thread must not panic")?;
+                match resp {
+                    EsploraResp::TxStatus(status) => {
+                        if let Some(anchor) = anchor_from_status(&status) {
+                            let _ = tx_graph.insert_anchor(txid, anchor);
+                        }
+                    }
+                    EsploraResp::Tx(Some(tx_info)) => {
+                        let _ = tx_graph.insert_tx(tx_info.to_tx());
```
LagginTimes: There was a discussion when the fee calculation for Electrum was implemented, that when we populate by Txids and Outpoints, we still want to add the prev TxOuts for fee calculation because there might be new relevant data. I just noticed that both the previous and new Esplora implementations appear to not actually do this, and was wondering if this needs to be added in?

crates/esplora/src/blocking_ext.rs
```diff
@@ -112,6 +164,235 @@ impl EsploraExt for esplora_client::BlockingClient {
...
+        self.populate_using_keychain_spks(
+            tx_graph,
+            spks.into_iter().enumerate().map(|(i, spk)| (i as u32, spk)),
+            usize::MAX,
+            parallel_requests,
+        )
+        .map(|_| ())
+    }
+
+    fn populate_using_txids<I: IntoIterator<Item = Txid>>(
```
LagginTimes: This is probably just a personal nit, but I'm of the opinion that the codebase would look "neater" if the naming scheme of conceptually similar methods were the same across both Electrum and Esplora, i.e. match `bdk_electrum`'s `populate_with_txids`.

evanlinjin: Thanks for the suggestion. The reason why I have a different name is because we aren't exactly populating the update graph with txids. Instead, we are using txids to fetch txs & anchors for the tx graph. However, I do feel the name can be improved. Let me know if you have any suggestions.

LagginTimes: I think I lean towards something similar to `chain_update`, maybe `graph_update_with_txids` or something similar? What do you think?

evanlinjin:
> Since this is a refactor and won't break the `bdk_wallet` API I'm not going to assign it a milestone until we get together for a team backlog prioritization call. 

That is fine for me. Because some users are waiting on this, I would appreciate it if we have it in soon. It doesn't break any API since it only adds new exposed methods.

crates/esplora/src/blocking_ext.rs
LLFourn: it's not populating, it's fetching transactions from a server. The fact that it's doing IO and fertching txs should be emphasized.

LLFourn:
Concept ACK

crates/esplora/src/blocking_ext.rs
```diff
@@ -10,59 +10,108 @@ use bdk_chain::{
     BlockId, ConfirmationTimeHeightAnchor, TxGraph,
 };
 use bdk_chain::{Anchor, Indexed};
-use esplora_client::TxStatus;
+use esplora_client::{OutputStatus, Tx, TxStatus};
 
 use crate::anchor_from_status;
 
-/// [`esplora_client::Error`]
+/// [`esplora_client::Error`] wrapped in a [`Box`].
```
LLFourn: you don't need to say it's wrapped in a `Box`. No one cares about this and it's right there in the type definition.

crates/esplora/src/blocking_ext.rs
evanlinjin: I mean technically the method is doing both? It's fetching txs and tx metadata from Esplora and then populating a `TxGraph` with it?

However, I'm thinking it's better to change the method signatures to **output** a `TxGraph` instead - this creates a nicer API, and the performance sacrifice is not too much. I had `&mut TxGraph` as an input because that was how the internal logic worked and it "felt more efficient" (which it is, but not by much).

Then we can change the methods to `fetch_txs_of_txids`, `fetch_txs_of_outpoints`, `fetch_txs_of_spks`, `fetch_txs_of_keychain_spks`, etc.

evanlinjin: Changed the methods to be as mentioned above. Except I've decided to use `with` as a preposition (instead of `of`).

crates/esplora/src/blocking_ext.rs
evanlinjin: Added.

crates/chain/src/changeset.rs
```diff
@@ -1,5 +1,4 @@
 /// A changeset containing [`crate`] structures typically persisted together.
-#[cfg(feature = "miniscript")]
```
LLFourn: If it's not necessary part of the PR revert it.

The correct way to clean this up is to remove it. If we're not going to remove it yet at least call the file `wallet_changeset.rs` and the type `WalletChangeSet` with a big notice at the top saying "This is going to be moved soon". It's confusing to all it `changeset.rs` because people will want to know how to create their own changeset but the trait for that is actually called `Append`.

crates/esplora/src/lib.rs
```diff
@@ -1,22 +1,63 @@
...
+//! [`EsploraExt::sync`] is used to sync against a subset of wallet data. For example, transaction
+//! histories of revealed and unused script from the external (public) keychain and/or statuses of
+//! wallet-owned UTXOs and spending transactions from them. The policy of what to sync against can
+//! be customized.
+//!
+//! [`EsploraExt::full_scan`] is designed to be used when importing or restoring a keychain where
+//! the range of possibly used scripts is not known. In this case it is necessary to scan all
+//! keychain scripts until a number (the `stop_gap`) of unused scripts is discovered.
+//!
+//! For a sync or full scan, the user receives relevant blockchain data and output updates for
```
LLFourn: This sentence needs to be reworked. The user receives it? (I think you mean caller) and how exactly do they output updates. Actually I thought the blockchain data was the update. Probably just delete all this with the other suggestion.

crates/esplora/src/lib.rs
```diff
@@ -1,22 +1,63 @@
...
-//! cares about, for example the scripts for all the receive addresses of a Wallet's keychain that it
-//! has shown a user. [`EsploraExt::full_scan`] is meant to be used when importing or restoring a
-//! keychain where the range of possibly used scripts is not known. In this case it is necessary to
-//! scan all keychain scripts until a number (the "stop gap") of unused scripts is discovered. For a
-//! sync or full scan the user receives relevant blockchain data and output updates for [`bdk_chain`]
-//! via a new [`TxGraph`] to be appended to any existing [`TxGraph`] data.
+//! The two primary methods are [`EsploraExt::sync`] and [`EsploraExt::full_scan`].
 //!
-//! Refer to [`example_esplora`] for a complete example.
+//! [`EsploraExt::sync`] is used to sync against a subset of wallet data. For example, transaction
```
LLFourn: All this documentation seems redundant since we have `SynRequest` and `FullScanRequest`.

> This library is primarily intended to satisfy `SyncRequest`s with [`sync`] and [`FullScanRequest`]s with [`full_scan`]. 

To know what that means they can read the documentation for those types.

LLFourn (dismissed):
Small doc improvements requested.

LLFourn:
Hmm I wonder what would happen if we just made updating the chain optional as part of `SyncRequest/SyncResult`?

crates/chain/src/changeset.rs
evanlinjin: Removed the commit for now!

crates/esplora/src/lib.rs
evanlinjin: Removed.

crates/esplora/src/lib.rs
evanlinjin: Removed.

oleonardolima (commented):
Concept ACK

crates/esplora/src/blocking_ext.rs
ValuedMammal: we may not need to box the error anymore after rust-esplora-client moved to using minreq

crates/esplora/src/lib.rs
```diff
@@ -1,22 +1,46 @@
...
+//! [BTCPay Server](https://docs.btcpayserver.org/FAQ/Wallet/#the-gap-limit-problem),
+//! and [Sparrow](https://www.sparrowwallet.com/docs/faq.html#ive-restored-my-wallet-but-some-of-my-funds-are-missing).
 //!
-//! Refer to [`example_esplora`] for a complete example.
+//! A `stop_gap` of 0 will be treated as a `stop_gap` of 1.
+//!
+//! # Async
+//!
+//! Just like how [`EsploraExt`] extends the functionality of an
+//! [`esplora_client::BlockingClient`], [`EsploraExt`] is the async version which extends
```
ValuedMammal:
```suggestion
//! [`esplora_client::BlockingClient`], [`EsploraAsyncExt`] is the async version which extends
```

crates/esplora/src/async_ext.rs
```diff
@@ -19,56 +20,102 @@ type Error = Box<esplora_client::Error>;
...
+    /// derived from a keychain. The scanning logic stops after a `stop_gap` number of consecutive
+    /// scripts with no transaction history is reached. `parallel_requests` specifies the maximum
+    /// number of HTTP requests to make in parallel.
+    ///
+    /// A [`TxGraph`] (containing the fetched transactions and anchors) and the last active
+    /// keychain index (if any) is returned. The last active keychain index is the keychain's last
+    /// script pubkey that contains a non-empty transaction history.
+    ///
+    /// Refer to [crate-level docs](crate) for more.
+    async fn fetch_txs_with_keychain_spks<I: Iterator<Item = Indexed<ScriptBuf>> + Send>(
```
LLFourn: do we actually need this as part of the API now?

crates/chain/src/spk_client.rs
```diff
@@ -1,388 +1,552 @@
...
         self
     }
 
-    /// Chain on additional [`Script`]s that will be synced against.
-    ///
-    /// This consumes the [`SyncRequest`] and returns the updated one.
-    #[must_use]
-    pub fn chain_spks(
+    /// Add [`Script`]s coupled with an associated label that will be synced against.
+    pub fn spks_with_labels(
```
ValuedMammal: I think an example could help for knowing how to include the `SpkLabel`

```rust
/// # Example
/// 
/// Sync revealed script pubkeys obtained from a [`KeychainTxOutIndex`].
///
/// ```rust
/// # use bdk_chain::spk_client::SyncRequest;
/// # use bdk_chain::indexer::keychain_txout::KeychainTxOutIndex;
/// # let indexer = KeychainTxOutIndex::<()>::default();
/// let indexed_spks = indexer.revealed_spks(..);
/// let request = SyncRequest::builder()
///     .spks_with_labels(indexed_spks)
///     .build();
/// ```
```

crates/chain/src/spk_client.rs
```diff
@@ -1,388 +1,552 @@
...
+    type Item = ScriptBuf;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.request.next_spk()
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let consumed = self.request.spks_consumed;
+        (consumed, Some(consumed))
+    }
```
ValuedMammal: Is `size_hint` meant to return the number of items consumed or remaining?

ValuedMammal (commented):
I haven't reviewed the whole diff but I can say that tests pass locally for me and the electrum/esplora example crates all work nicely.

notmandatory:
Concept ACK since it looks like the only changes to the `bdk_wallet` beta API is renaming `start_sync_with_revealed_spks()` to `build_sync_with_revealed_spks()` and `start_full_scan()` to `build_full_scan()`. Though in general I'd like to keep API changes in this milestone as close as we can to none.

The return type also changes but it isn't something a user would be interacting with unless they're developing their own SPK based blockchain client (which I haven't heard of anyone doing).

evanlinjin:
> Concept ACK since it looks like the only changes to the `bdk_wallet` beta API is renaming `start_sync_with_revealed_spks()` to `build_sync_with_revealed_spks()` and `start_full_scan()` to `build_full_scan()`. Though in general I'd like to keep API changes in this milestone as close as we can to none.
> 
> The return type also changes but it isn't something a user would be interacting with unless they're developing their own SPK based blockchain client (which I haven't heard of anyone doing).

Looking at your comment now, I don't think my renaming adds much value, I will change it back.

crates/chain/src/spk_client.rs
evanlinjin: Good point, it is meant to be remaining.

LagginTimes:
LGTM so far, will continue reviewing this tomorrow. 👍

crates/wallet/src/wallet/mod.rs
```diff
@@ -2319,9 +2322,10 @@ impl Wallet {
     /// This is the first step when performing a spk-based wallet partial sync, the returned
     /// [`SyncRequest`] collects all revealed script pubkeys from the wallet keychain needed to
     /// start a blockchain sync with a spk based blockchain client.
-    pub fn start_sync_with_revealed_spks(&self) -> SyncRequest {
-        SyncRequest::from_chain_tip(self.chain.tip())
-            .populate_with_revealed_spks(&self.indexed_graph.index, ..)
+    pub fn build_sync_with_revealed_spks(&self) -> SyncRequestBuilder<(KeychainKind, u32)> {
+        SyncRequest::builder()
+            .chain_tip(self.chain.tip())
+            .revealed_spks_from_indexer(&self.indexed_graph.index, ..)
```
notmandatory, line 2446: ❓ I like the new builder pattern for sync and full scan requests, but do we really need to return the builder instead of the final built request (also for full scan) here in Wallet? My understanding is that we'd add new Wallet methods for other useful request options (ie. build_sync_with_revealed_unused_spks(), or something like that).
```suggestion
            .revealed_spks_from_indexer(&self.indexed_graph.index, ..)
            .build()
```

evanlinjin: I think it's better to return the builder, since the caller can further customize it. I've also changed chain sources to take in `impl Into<...Request>` so they can directly take in a builder struct.

crates/esplora/README.md
```diff
@@ -26,11 +27,17 @@ To use the extension traits:
 ```rust
 // for blocking
 use bdk_esplora::EsploraExt;
-// for async
+// // for async
 // use bdk_esplora::EsploraAsyncExt;
```
LagginTimes: Can probably do without the additional commenting.

```suggestion
// for async
use bdk_esplora::EsploraAsyncExt;
```

crates/esplora/src/async_ext.rs
evanlinjin: This is just a private helper-function now.

crates/chain/src/spk_client.rs
```diff
@@ -1,388 +1,589 @@
...
+
+    /// Total remaining items of the request.
+    pub fn remaining(&self) -> usize {
+        self.spks_remaining + self.txids_remaining + self.outpoints_remaining
+    }
+}
+
+/// Builds a [`SyncRequest`].
+#[must_use]
+pub struct SyncRequestBuilder<SpkLabel = ()> {
```
notmandatory: Is `SpkLabel` essentially the same thing as `I` (index?) in `SyncRequest<I = ()>` ? It's confusing to use different identifiers for generic types that represent the same thing. I'm also worried that the "Label" concept might get confused with BIP-329 wallet labels which are unrelated to your change.  One fix is to rename `SpkLabel` to  `SpkIndex` and the related method to `spks_with_indexes`. Then also update `SyncRequest<SpkIndex = ()>`.  Same for `FullScan`.

ValuedMammal (commented):
After looking over the changes to esplora, I did confirm it's possible to calculate fee for all wallet transactions after doing a full scan

evanlinjin:
@ValuedMammal it should also be possible to calculate fee for all wallet transactions after doing sync

notmandatory (changes requested):
I tested all the examples (rebased on #1549) and confirmed scan, sync. I have one question/request but otherwise this PR looks good to merge. 

I also have one complaint, this PR should have been split up into at least two PRs, 1. to refactor the electrum and esplora scan and sync and make chain optional, and 2. to refactor sync and scan requests with the builder and more progress details. Big PRs like this are hard to properly review and prioritize.

But since I'd like to get this in for a beta.2 this week we don't have time to split and re-review separate PRs for this one. For all PR authors going forward we need to work on making smaller more focused single purpose PRs.

crates/esplora/src/async_ext.rs
```diff
@@ -230,176 +237,235 @@ async fn chain_update<A: Anchor>(
...
+{
+    let outpoints = outpoints.into_iter().collect::<Vec<_>>();
+
+    // make sure txs exists in graph and tx statuses are updated
+    // TODO: We should maintain a tx cache (like we do with Electrum).
+    let mut tx_graph = fetch_txs_with_txids(
+        client,
+        // NOTE: @evanlinjin We need to collect here otherwise we run into lifetime issues, but I
+        // do not know why. If someone can figure this out, that will be amazing.
+        outpoints.iter().map(|op| op.txid).collect::<Vec<_>>(),
```
ValuedMammal: not sure if this is what you want, but it seems to compile

```suggestion
        outpoints.iter().copied().map(|op| op.txid),
```

crates/chain/src/spk_client.rs
evanlinjin: For `FullScan`, what should we name the `K` generic parameter? Would `Keychain` worK?

evanlinjin: I think just having `I` for spk index and `K` for keychain should be okay? I renamed the generic type parameters like so.

crates/esplora/src/async_ext.rs
evanlinjin: Thank you @ValuedMammal, that seems to work. I'm still quite confused why your solution works and why `outpoints.iter().map(|op| op.txid)` does not. If someone can explain it to me, that will be wonderful.

crates/chain/src/spk_client.rs
notmandatory: Yes `I` for spk index and `K` for keychain makes sense. My main concern is that what ever name we use it's used consistently.  Will you also rename `SkpLabel` on this line and below to `I` ?

crates/chain/src/spk_client.rs
```diff
@@ -1,388 +1,586 @@
...
-    ) -> Self {
-        self.outpoints = Box::new(outpoints.into_iter());
+    /// This is used to update [`LocalChain`](crate::local_chain::LocalChain).
+    pub fn chain_tip(mut self, cp: CheckPoint) -> Self {
+        self.inner.chain_tip = Some(cp);
         self
     }
 
-    /// Chain on additional [`Script`]s that will be synced against.
+    /// Add [`Script`]s coupled with an associated label that will be synced against.
```
notmandatory:
```suggestion
    /// Add [`Script`]s coupled with an associated index that will be synced against.
```

crates/chain/src/spk_client.rs
```diff
@@ -1,388 +1,586 @@
...
+    }
+}
+
+/// Builds a [`SyncRequest`].
+#[must_use]
+pub struct SyncRequestBuilder<SpkLabel = ()> {
+    inner: SyncRequest<SpkLabel>,
+}
+
+impl<SpkLabel> Default for SyncRequestBuilder<SpkLabel> {
```
notmandatory:
```suggestion
pub struct SyncRequestBuilder<I = ()> {
    inner: SyncRequest<I>,
}

impl<I> Default for SyncRequestBuilder<I> {
```

notmandatory (changes requested):
Looks like a few more spots where `SpkLabel` needs to be renamed to `I`.

crates/chain/src/spk_client.rs
evanlinjin: Oh no! Fixing now

crates/chain/src/spk_client.rs
```diff
@@ -1,388 +1,586 @@
...
+    /// # use bdk_chain::indexer::keychain_txout::KeychainTxOutIndex;
+    /// # use bdk_chain::miniscript::{Descriptor, DescriptorPublicKey};
+    /// # let secp = bdk_chain::bitcoin::secp256k1::Secp256k1::signing_only();
+    /// # let (descriptor_a,_) = Descriptor::<DescriptorPublicKey>::parse_descriptor(&secp, "tr([73c5da0a/86'/0'/0']xprv9xgqHN7yz9MwCkxsBPN5qetuNdQSUttZNKw1dcYTV4mkaAFiBVGQziHs3NRSWMkCzvgjEe3n9xV8oYywvM8at9yRqyaZVz6TYYhX98VjsUk/0/*)").unwrap();
+    /// # let (descriptor_b,_) = Descriptor::<DescriptorPublicKey>::parse_descriptor(&secp, "tr([73c5da0a/86'/0'/0']xprv9xgqHN7yz9MwCkxsBPN5qetuNdQSUttZNKw1dcYTV4mkaAFiBVGQziHs3NRSWMkCzvgjEe3n9xV8oYywvM8at9yRqyaZVz6TYYhX98VjsUk/1/*)").unwrap();
+    /// let mut indexer = KeychainTxOutIndex::<&'static str>::default();
+    /// indexer.insert_descriptor("descriptor_a", descriptor_a)?;
+    /// indexer.insert_descriptor("descriptor_b", descriptor_b)?;
+    ///
+    /// /* Assume that the caller does more mutations to the `indexer` here... */
```
ValuedMammal, line 159: What extra assumptions are needed for this example to make sense? (I would think none)

ValuedMammal (approved):
ACK 6d77e2e2c3e7c8d49c50bd6b41e7dbc3ddd1d275

crates/esplora/src/async_ext.rs
ValuedMammal: I found this?

https://users.rust-lang.org/t/implementation-of-send-is-not-general-enough-but-cannot-make-it-more-general/115087

rustaceanrob:
It is non-intuitive to me to have `SyncProgress` be a collection of `usize` fields. It seems as though you can "consume" the SPK and still have the request fail? I might be interpreting this diff wrong but I feel as though the `SyncProgress` should be derived from a collection of `SyncItem` rather than being a distinct object

i.e. `impl SyncProgress for Vec<SyncItem>`

notmandatory (approved):
ACK 6d77e2e2c3e7c8d49c50bd6b41e7dbc3ddd1d275

notmandatory:
> It is non-intuitive to me to have `SyncProgress` be a collection of `usize` fields. It seems as though you can "consume" the SPK and still have the request fail? I might be interpreting this diff wrong but I feel as though the `SyncProgress` should be derived from a collection of `SyncItem` rather than being a distinct object
> 
> i.e. `impl SyncProgress for Vec<SyncItem>`

I see your point but in practice the `SyncProgress` will be used to just show the percent completed for these items and especially for mobile apps using language bindings transferring smaller and simpler structs is better.

PR #1486, opened 2024-06-24
refactor(chain): calculate DescriptorId as the sha256 hash of spk at index 0

### LINKED ISSUE #1483: `DescriptorId` is not unique enough
**Describe the bug**  

As mentioned on [`Discord#development`](https://discord.com/channels/753336465005608961/753367451319926827/1253198896897724446), there is a fundamental problem with how we define a `DescriptorId`. Currently `DescriptorId` does not uniquely identify a descriptor. I.e. we can change how the xpub is portrayed and it would completely change the `DescriptorId`. This is a problem because the whole point of `DescriptorId` is so that the last derivation index is stored with something that uniquely represents a descriptor.

**Proposed Solutions**

An idea (first mentioned to me by @LLFourn) is to use the hash of the first derived spk for the `DescriptorId`. We can consider disallowing non-wildcard descriptors in `KeychainTxOutIndex` (as they aren't keychains) and it allows us full uniqueness for `DescriptorId`s.

notmandatory:
### Description

Rename `DescriptorId` to `KeychainId` and `descriptor_id()` to `keychain_id()`.  

Calculate keychain ids as the hash of the spk derived from its descriptor as index 0. 

Added docs to `Wallet` and `KeychainTxOutIndex::insert_descriptor()` explaining that it's the users responsibility not to use wildcard and non-wildcard descriptors that can derive the same script pubkey. I also recommended for `Wallet` that legacy non-wildcard descriptors be added to a temporary `Wallet` and swept into a modern wildcard descriptor.

### Notes to the reviewers

fixes #1483

### Changelog notice

changed

- Calculate DescriptorId as the sha256 hash of the spk at index 0.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/descriptor_ext.rs
+++ b/crates/chain/src/descriptor_ext.rs
@@ -1,12 +1,8 @@
-use crate::{
-    alloc::{string::ToString, vec::Vec},
-    miniscript::{Descriptor, DescriptorPublicKey},
-};
+use crate::miniscript::{Descriptor, DescriptorPublicKey};
 use bitcoin::hashes::{hash_newtype, sha256, Hash};
 
 hash_newtype! {
-    /// Represents the ID of a descriptor, defined as the sha256 hash of
-    /// the descriptor string, checksum excluded.
+    /// Represents the unique ID of a descriptor.
     ///
     /// This is useful for having a fixed-length unique representation of a descriptor,
     /// in particular, we use it to persist application state changes related to the
@@ -21,8 +17,8 @@ pub trait DescriptorExt {
     /// Panics if the descriptor wildcard is hardened.
     fn dust_value(&self) -> u64;
 
-    /// Returns the descriptor id, calculated as the sha256 of the descriptor, checksum not
-    /// included.
+    /// Returns the descriptor ID, calculated as the sha256 hash of the spk derived from the
+    /// descriptor at index 0.
     fn descriptor_id(&self) -> DescriptorId;
 }
 
@@ -36,9 +32,7 @@ impl DescriptorExt for Descriptor<DescriptorPublicKey> {
     }
 
     fn descriptor_id(&self) -> DescriptorId {
-        let desc = self.to_string();
-        let desc_without_checksum = desc.split('#').next().expect("Must be here");
-        let descriptor_bytes = <Vec<u8>>::from(desc_without_checksum.as_bytes());
-        DescriptorId(sha256::Hash::hash(&descriptor_bytes))
+        let spk = self.at_derivation_index(0).unwrap().script_pubkey();
+        DescriptorId(sha256::Hash::hash(spk.as_bytes()))
     }
 }
--- a/crates/wallet/tests/wallet.rs
+++ b/crates/wallet/tests/wallet.rs
@@ -31,7 +31,16 @@ mod common;
 use common::*;
 
 fn receive_output(wallet: &mut Wallet, value: u64, height: ConfirmationTime) -> OutPoint {
-    let addr = wallet.next_unused_address(KeychainKind::External);
+    let addr = wallet.next_unused_address(KeychainKind::External).address;
+    receive_output_to_address(wallet, addr, value, height)
+}
+
+fn receive_output_to_address(
+    wallet: &mut Wallet,
+    addr: Address,
+    value: u64,
+    height: ConfirmationTime,
+) -> OutPoint {
     let tx = Transaction {
         version: transaction::Version::ONE,
         lock_time: absolute::LockTime::ZERO,
@@ -3995,6 +4004,44 @@ fn test_taproot_load_descriptor_duplicated_keys() {
         addr.to_string(),
         "bcrt1pvysh4nmh85ysrkpwtrr8q8gdadhgdejpy6f9v424a8v9htjxjhyqw9c5s5"
     );
+}
+
+/// In dev mode this test panics, but in release mode, or if the `debug_panic` in `TxOutIndex::replenish_inner_index`
+/// is commented out, there is no panic and the balance is calculated correctly. See issue [#1483]
+/// and PR [#1486] for discussion on mixing non-wildcard and wildcard descriptors.
+///
+/// [#1483]: https://github.com/bitcoindevkit/bdk/issues/1483
+/// [#1486]: https://github.com/bitcoindevkit/bdk/pull/1486
+#[test]
+#[should_panic(
+    expected = "replenish lookahead: must not have existing spk: keychain=Internal, lookahead=25, next_store_index=0, next_reveal_index=0"
+)]
+fn test_keychains_with_overlapping_spks() {
+    // this can happen if a non-wildcard descriptor keychain derives an spk that a
+    // wildcard descriptor keychain in the same wallet also derives.
+
+    // index 1 spk overlaps with non-wildcard change descriptor
+    let wildcard_keychain = "wpkh(tprv8ZgxMBicQKsPdDArR4xSAECuVxeX1jwwSXR4ApKbkYgZiziDc4LdBy2WvJeGDfUSE4UT4hHhbgEwbdq8ajjUHiKDegkwrNU6V55CxcxonVN/*)";
+    let non_wildcard_keychain = "wpkh(tprv8ZgxMBicQKsPdDArR4xSAECuVxeX1jwwSXR4ApKbkYgZiziDc4LdBy2WvJeGDfUSE4UT4hHhbgEwbdq8ajjUHiKDegkwrNU6V55CxcxonVN/1)";
+
+    let (mut wallet, _) = get_funded_wallet_with_change(wildcard_keychain, non_wildcard_keychain);
+    assert_eq!(wallet.balance().confirmed, Amount::from_sat(50000));
+
+    let addr = wallet
+        .reveal_addresses_to(KeychainKind::External, 1)
+        .last()
+        .unwrap()
+        .address;
+    let _outpoint = receive_output_to_address(
+        &mut wallet,
+        addr,
+        8000,
+        ConfirmationTime::Confirmed {
+            height: 2000,
+            time: 0,
+        },
+    );
+    assert_eq!(wallet.balance().confirmed, Amount::from_sat(58000));
 }
 
 #[test]
--- a/crates/chain/src/keychain/txout_index.rs
+++ b/crates/chain/src/keychain/txout_index.rs
@@ -352,6 +352,13 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     ///
     /// keychain <-> descriptor is a one-to-one mapping that cannot be changed. Attempting to do so
     /// will return a [`InsertDescriptorError<K>`].
+    ///
+    /// `[KeychainTxOutIndex]` will prevent you from inserting two descriptors which derive the same
+    /// script pubkey at index 0, but it's up to you to ensure that descriptors don't collide at
+    /// other indices. If they do nothing catastrophic happens at the `KeychainTxOutIndex` level
+    /// (one keychain just becomes the defacto owner of that spk arbitrarily) but this may have
+    /// subtle implications up the application stack like one UTXO being missing from one keychain
+    /// because it has been assigned to another which produces the same script pubkey.
     pub fn insert_descriptor(
         &mut self,
         keychain: K,
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -91,6 +91,9 @@ const COINBASE_MATURITY: u32 = 100;
 /// [`ChangeSet`]s (see [`take_staged`]). Also see individual functions and example for instructions
 /// on when [`Wallet`] state needs to be persisted.
 ///
+/// The `Wallet` descriptor (external) and change descriptor (internal) must not derive the same
+/// script pubkeys. See [`KeychainTxOutIndex::insert_descriptor()`] for more details.
+///
 /// [`signer`]: crate::signer
 /// [`take_staged`]: Wallet::take_staged
 #[derive(Debug)]


LLFourn:
> ### Description
> 
> **This is a proof of concept and likely incomplete!**
> 
> Refactored `DescriptorExt::descriptor_id()` to calculate descriptor ids as the hash of the SPK as index 0. For any xkeys in the descriptor that don't end in a wildcard, replaces the last derivation index with a wildcard for the purpose of calculating the descriptor id.

I don't feel doing a blind wildcard replacement is worth the trouble here and actually interferes with the design @evanlinjin is implementing I think. It should just be the SPK at index 0 while leaving the descriptor as is.

evanlinjin:
I'm not a fan of doing the wildcard-replacement as it creates a behavior that I find hard to reason with: We cannot have 2 non-wildcard descriptors that are part of the same keychain (even though their derived spks do not conflict).

I think it's better to disallow non-wildcard descriptors in `KeychainTxOutIndex`. This way, `DesciptorId` is always guaranteed to be unique. If the caller wishes to use single addresses, they can use `SpkTxOutIndex`. This means `bdk_wallet` cannot have non-wildcard descriptors. I think it's okay because `bdk_wallet` is meant to be opinionated and simple.

I'm going to provide a few suggestions that I think is enough for us to move into `bdk_wallet` beta territory.

1. Rename `DescriptorId` into `KeychainId`. We only want unique id's for keychains (chains of keys). Single-spk descriptors are not keychains (by this definition).
2. Not all descriptors have a `KeychainId`. Refactor `DescriptorExt::descriptor_id()` to `DescriptorExt::keychain_id() -> Option<KeychainId>`.
3. `KeychainTxOutIndex` should error out when we try introduce a non-wildcard descriptor (therefore, so does `bdk_wallet`). If it is introduced in the changeset, we `debug_assert!` or ignore it (for non-debug).

I'm already working on renaming the `K` generic of `KeychainTxOutIndex` to `L` for "label" (which is used to label keychains). We can cherry-pick commits to combine these into one PR.

notmandatory:
Thanks for the feedback, I'll update this PR as suggested. 

@evanlinjin I'd still like to have some way to sweep single SPK descriptors into a `bdk_wallet`, can that be done with `SpkTxOutIndex` somehow?

notmandatory:
There are still a few tests failing that I could use some help fixing.

crates/chain/src/descriptor_ext.rs
```diff
@@ -21,9 +21,9 @@ pub trait DescriptorExt {
     /// Panics if the descriptor wildcard is hardened.
     fn dust_value(&self) -> u64;
 
-    /// Returns the descriptor id, calculated as the sha256 of the descriptor, checksum not
+    /// Returns the keychain id, calculated as the sha256 of the descriptor, checksum not
     /// included.
-    fn descriptor_id(&self) -> DescriptorId;
+    fn keychain_id(&self) -> KeychainId;
```
evanlinjin: Let's have this return an `Option`, non-wildcard descriptors are not keychains.

notmandatory:
I had to ignore 3 PSBT tests and 1 Wallet test, added TODOs to fix them during 1.0.0-beta phase.

storopoli:
> I had to ignore 3 PSBT tests and 1 Wallet test, added TODOs to fix them during 1.0.0-beta phase.

We should also open an issue no (if this is merged)? TODO comments get lost and forgotten quite quickly

crates/chain/src/descriptor_ext.rs
notmandatory: In my latest push I removed the restriction against using non-wildcard descriptors so this function will never fail and doesn't need an `Option`. See updated docs.

notmandatory:
I simplified this PR back to allowing wildcard and non-wildcard descriptors in a `Wallet` or `KeychainTxOutIndex`. We determine the `KeychainId` based on the hash of the derived index 0 script pubkey. No new errors or other API changes are needed and all existing tests pass.

I update docs to explain that it's the developer's responsibility to ensure that wildcard and non-wildcard descriptors don't derive the same spk if used in the same `Wallet` or the same `KeychainTxOutIndex`.

notmandatory:
@evanlinjin if you're OK with this solution I'd like to move it to the beta milestone since it won't change the wallet API.

evanlinjin:
@notmandatory this is a breaking change because `KeychainId` is part of the changeset.

notmandatory:
@evanlinjin ah ok I wasn't thinking about the changeset, I'll keep this is the alpha milestone.

evanlinjin:
@notmandatory I've been thinking about this. The problem with your solution is that any descriptor/key can be turned into a `DefiniteDescriptor`/`DefiniteDescriptorKey`. If done so, we cannot detect conflicts of spks.

If we want to support single spk descriptors, I think we should do it properly. The way to do this is to have two versions of `Wallet` or do something like `Wallet<Indexer>`. Also, I imagine the behaviour of single-spk wallets will be quite different to keychain wallets.

I would first focus on a keychain wallet first and get it correct. Then we can build a single-spk wallet later (it shouldn't be too hard?).

notmandatory:
@evanlinjin I'd rather not do any more redesign on `Wallet` right now for this when BDK pre-1.0 has also likely had this issue and it's never been an problem. The single key functionality can still be safely used if the few users who might need it mind the caveats I put in the docs.  I'd be happy to make fixing this properly one of the 2.0 milestone goals.

evanlinjin:
The problem is the `KeychainId` definition cannot easily be changed later.

Edit: Okay I guess for actual keychains it'll be the same. We won't have `KeychainId`s for non-keychains in the future.

crates/chain/src/descriptor_ext.rs
```diff
@@ -21,9 +21,9 @@ pub trait DescriptorExt {
     /// Panics if the descriptor wildcard is hardened.
     fn dust_value(&self) -> u64;
 
-    /// Returns the descriptor id, calculated as the sha256 of the descriptor, checksum not
+    /// Returns the keychain id, calculated as the sha256 of the descriptor, checksum not
     /// included.
```
evanlinjin: Docs need updating. It's now the sha256 of the spk at derivation index 0

crates/chain/src/descriptor_ext.rs
```diff
@@ -35,10 +35,10 @@ impl DescriptorExt for Descriptor<DescriptorPublicKey> {
             .to_sat()
     }
 
-    fn descriptor_id(&self) -> DescriptorId {
-        let desc = self.to_string();
-        let desc_without_checksum = desc.split('#').next().expect("Must be here");
-        let descriptor_bytes = <Vec<u8>>::from(desc_without_checksum.as_bytes());
-        DescriptorId(sha256::Hash::hash(&descriptor_bytes))
+    /// Generate an id for the keychain using the spk derived at keychain index 0.
```
evanlinjin: Remove?

evanlinjin (commented):
Just some docs to update, but looks good to me.

crates/chain/src/descriptor_ext.rs
notmandatory: fixed.

crates/chain/src/descriptor_ext.rs
notmandatory: removed.

notmandatory:
> The problem is the `KeychainId` definition cannot easily be changed later.
> 
> Edit: Okay I guess for actual keychains it'll be the same. We won't have `KeychainId`s for non-keychains in the future.

Maybe this is just semantics, but technically even a non-wildcard descriptor could be called a keychain since as currently coded it's pks still derive a key at each index, they're just all the same. 🙃 

https://docs.rs/miniscript/latest/miniscript/descriptor/enum.DescriptorPublicKey.html#method.at_derivation_index

evanlinjin:
> Maybe this is just semantics, but technically even a non-wildcard descriptor could be called a keychain since as currently coded it's pks still derive a key at each index, they're just all the same. 🙃
> 
> https://docs.rs/miniscript/latest/miniscript/descriptor/enum.DescriptorPublicKey.html#method.at_derivation_index

Sure, but as long as there is a pathway to achieve correctness (perfect uniqueness) of `KeychainId`, I am happy to have this as a compromise for now. Also, I would argue that having keychain-wallet and spk-wallet separate would be a better API. I.e. you don't need to "reveal" more spks if there is no keychain.

evanlinjin:
I would be happy to ACK and merge this soon so I can base my changes on this. I am moving `keychains_added` out of `keychain::ChangeSet` and into `CombinedChangeSet` (since we can't do this later on as it's a breaking change).

notmandatory:
I also added a little test to clarify what happens if overlapping keychains are used in a `Wallet`, at least in dev mode this situation will panic, and in non-dev the correct balance is still given. We can build on this test when developing the future redesign.

crates/wallet/tests/wallet.rs
```diff
@@ -3997,6 +4006,41 @@ fn test_taproot_load_descriptor_duplicated_keys() {
...
 }
 
+// In dev mode this test panics, but in release mode, or if the debug_panic is commented out,
+// there is no panic and the balance is calculated correctly. See #1483 and #1486 for discussion
+// on mixing non-wildcard and wildcard descriptors.
+#[test]
+#[should_panic(
+    expected = "replenish lookahead: must not have existing spk: keychain=Internal, lookahead=25, next_store_index=0, next_reveal_index=0"
+)]
+fn test_keychains_with_overlapping_spks() {
```
storopoli:
```suggestion
#[test]
#[should_panic(
    expected = "replenish lookahead: must not have existing spk: keychain=Internal, lookahead=25, next_store_index=0, next_reveal_index=0"
)]
/// In `dev` mode this test panics, but in `release` mode, or if the `debug_panic` is commented out,
/// there is no panic and the balance is calculated correctly. See issues #1483 and #1486 for discussion
/// on mixing non-wildcard and wildcard descriptors.
fn test_keychains_with_overlapping_spks() {
```

The function below this one has docstrings.
Maybe these can be docstrings as well?
I also added the `issue` identifier to 1483 and 1486.

notmandatory: I fixed up the docs as suggested. Even though these docs won't show up in the official rust docs it's still nice to see them highlighted and linked, at least in my IDE.

crates/chain/src/descriptor_ext.rs
```diff
@@ -35,10 +35,9 @@ impl DescriptorExt for Descriptor<DescriptorPublicKey> {
...
 
-    fn descriptor_id(&self) -> DescriptorId {
-        let desc = self.to_string();
-        let desc_without_checksum = desc.split('#').next().expect("Must be here");
-        let descriptor_bytes = <Vec<u8>>::from(desc_without_checksum.as_bytes());
-        DescriptorId(sha256::Hash::hash(&descriptor_bytes))
+    fn keychain_id(&self) -> KeychainId {
+        let spk = self.at_derivation_index(0).unwrap().script_pubkey();
+        let spk_bytes = <Vec<u8>>::from(spk.as_bytes());
+        KeychainId(Hash::hash(&spk_bytes))
```
LLFourn: :pick:  can you put sha256::Hash::hash or something. It's not clear which hash algorithm you're using. 

:pick: you shouldn't need to allocate a `Vec` here. Why can't you just put the `as_bytes` into the hash.

LLFourn (commented):
I think this PR was better when it didn't rename `DescriptorId` to `KeychainId`. There is not concept of a `Keychain` in this PR and only descriptors have a `KeychainId` apparently. In my mind this should have just been about calculating `DescriptorId` differently. Anyway @evanlinjin explicitly requested it....

> I think it's better to disallow non-wildcard descriptors in KeychainTxOutIndex. This way, DesciptorId is always guaranteed to be unique.

Unfortunately this is not the case.  `wsh(multi(2,xpubA/42, xpubB/*)` and `wsh(multi(2, xpubA/*, xpubB/42)` have the same spk at index `42` but different at every other index. It is actually impossible to ensure that a malicious party cannot generate two completely different descriptors that happen to have the same spks at two different derivation indices. This is because `wpkh`, `pkh` and `sh` descriptors all use 20 byte hashes which are *not* sufficiently collision resistant.

Once again in order to have this  "zero collisions" property you must enforce it at the application level. If this needs to be enforced against a malicious adversary who can insert whatever descriptor they want then you must disallow all the 20-byte hash spk descriptor types and make sure that all keys are xpubs with wildcards...I think! We can document this but there's nothing else we can do. The design here is still right. If we can make what happens when there are collisions less surprising that's good but otherwise we should document and ignore this problem.

crates/chain/src/descriptor_ext.rs
notmandatory: Thanks, fixed.

notmandatory:
> I think this PR was better when it didn't rename `DescriptorId` to `KeychainId`. There is not concept of a `Keychain` in this PR and only descriptors have a `KeychainId` apparently. In my mind this should have just been about calculating `DescriptorId` differently. 

I rolled back the `DescriptorId` name change. If @evanlinjin or anyone else has really strong opinions about the name we can discuss but for now I agree it doesn't add anything to the main purpose of this PR.

notmandatory:
> Once again in order to have this "zero collisions" property you must enforce it at the application level. If this needs to be enforced against a malicious adversary who can insert whatever descriptor they want then you must disallow all the 20-byte hash spk descriptor types and make sure that all keys are xpubs with wildcards...I think! We can document this but there's nothing else we can do. The design here is still right. If we can make what happens when there are collisions less surprising that's good but otherwise we should document and ignore this problem.

Is my current warning for `Wallet` and `KeychainTxOutIndex::insert_descriptor()` adequate for now? For a typical wallet use case I expect the most likely collisions would be someone trying to use/misuse non-wildcard descriptors.

LLFourn:
Thanks. The PR is now much easier to review without the name change.

Here's my attempt at documenting this:

> `KeychainTxOutIndex` will prevent you from inserting two descriptors which derive the same script pubkey at index 0 but it's up to you to ensure that descriptors don't collide at other indices. If they do nothing catastrophic happens at the `KeychainTxOutIndex` level (one keychain just becomes the defacto owner of that spk arbitrarily) but this may have subtle implications up the application stack like one UTXO being missing from one keychain because it has been assigned to another which produces the same script pubkey.

notmandatory:
@LLFourn I updated the docs with your much more nuanced explanation.

oleonardolima (commented):
Concept ACK 8f5b172e59b4d5e199b98777855b0d9b46742f66

LLFourn:
ACK 8f5b172e59b4d5e199b98777855b0d9b46742f66

LLFourn (approved):

PR #1487, opened 2024-06-24
refactor(wallet)!: Add support for custom sorting and deprecate BIP69

### LINKED ISSUE #534: Remove BIP 69 sorting

nymius:
### Description

Resolves https://github.com/bitcoindevkit/bdk/issues/534.

Resumes from the work in https://github.com/bitcoindevkit/bdk/pull/556.

Add custom sorting function for inputs and outputs through `TxOrdering::Custom` and deprecates `TxOrdering::Bip69Lexicographic`.

### Notes to the reviewers

I tried consider all discussions in https://github.com/bitcoindevkit/bdk/issues/534 while implementing some changes to the original PR. I created a summary of the considerations I had while implementing this:

##### Why use smart pointers?
The size of enums and structs should be known at compilation time. A struct whose fields implements some kind of trait cannot be specified without using a smart pointer because the size of the implementations of the trait cannot be known beforehand.

##### Why `Arc` or `Rc` instead of `Box`?
The majority of the useful smart pointers that I know (`Arc`, `Box`, `Rc`) for this case implement `Drop` which rules out the implementation of `Copy`, making harder to manipulate a simple enum like `TxOrdering`. `Clone` can be used instead, implemented by `Arc` and `Rc`, but not implemented by `Box`.

#####  Why `Arc` instead of `Rc`?
Multi threading I guess.

##### Why using a type alias like `TxVecSort`?
cargo-clippy was accusing a too complex type if using the whole type inlined in the struct inside the enum.

##### Why `Fn` and not `FnMut`?
`FnMut` is not allowed inside `Arc`. I think this is due to the `&mut self` ocupies the first parameter of the `call` method when desugared (https://rustyyato.github.io/rust/syntactic/sugar/2019/01/17/Closures-Magic-Functions.html), which doesn't respects `Arc` limitation of not having mutable references to data stored inside `Arc`:
Quoting the [docs](https://doc.rust-lang.org/std/sync/struct.Arc.html):
> you cannot generally obtain a mutable reference to something inside an `Arc`.

`FnOnce` > `FnMut` > `Fn`, where `>` stands for "is supertrait of", so, `Fn` can be used everywhere `FnMut` is expected.

##### Why not `&'a dyn FnMut`?
It needs to include a lifetime parameter in `TxOrdering`, which will force the addition of a lifetime parameter in `TxParams`, which will require the addition of a lifetime parameter in a lot of places more. **Which one is preferable?**

### Changelog notice

- Adds new `TxOrdering` variant: `TxOrdering::Custom`. A structure that stores the ordering functions to sort the inputs and outputs of a transaction.
- Deprecates `TxOrdering::Bip69Lexicographic`.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/wallet/src/wallet/tx_builder.rs
+++ b/crates/wallet/src/wallet/tx_builder.rs
@@ -42,10 +42,13 @@ use alloc::{boxed::Box, rc::Rc, string::String, vec::Vec};
 use core::cell::RefCell;
 use core::fmt;
 
+use alloc::sync::Arc;
+
 use bitcoin::psbt::{self, Psbt};
 use bitcoin::script::PushBytes;
 use bitcoin::{
-    absolute, Amount, FeeRate, OutPoint, ScriptBuf, Sequence, Transaction, Txid, Weight,
+    absolute, Amount, FeeRate, OutPoint, ScriptBuf, Sequence, Transaction, TxIn, TxOut, Txid,
+    Weight,
 };
 use rand_core::RngCore;
 
@@ -763,8 +766,10 @@ impl fmt::Display for AddForeignUtxoError {
 #[cfg(feature = "std")]
 impl std::error::Error for AddForeignUtxoError {}
 
+type TxSort<T> = dyn Fn(&T, &T) -> core::cmp::Ordering;
+
 /// Ordering of the transaction's inputs and outputs
-#[derive(Default, Debug, Ord, PartialOrd, Eq, PartialEq, Hash, Clone, Copy)]
+#[derive(Clone, Default)]
 pub enum TxOrdering {
     /// Randomized (default)
     #[default]
@@ -773,6 +778,24 @@ pub enum TxOrdering {
     Untouched,
     /// BIP69 / Lexicographic
     Bip69Lexicographic,
+    /// Provide custom comparison functions for sorting
+    Custom {
+        /// Transaction inputs sort function
+        input_sort: Arc<TxSort<TxIn>>,
+        /// Transaction outputs sort function
+        output_sort: Arc<TxSort<TxOut>>,
+    },
+}
+
+impl core::fmt::Debug for TxOrdering {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        match self {
+            TxOrdering::Shuffle => write!(f, "Shuffle"),
+            TxOrdering::Untouched => write!(f, "Untouched"),
+            TxOrdering::Bip69Lexicographic => write!(f, "Bip69Lexicographic"),
+            TxOrdering::Custom { .. } => write!(f, "Custom"),
+        }
+    }
 }
 
 impl TxOrdering {
@@ -780,14 +803,14 @@ impl TxOrdering {
     ///
     /// Uses the thread-local random number generator (rng).
     #[cfg(feature = "std")]
-    pub fn sort_tx(self, tx: &mut Transaction) {
+    pub fn sort_tx(&self, tx: &mut Transaction) {
         self.sort_tx_with_aux_rand(tx, &mut bitcoin::key::rand::thread_rng())
     }
 
     /// Sort transaction inputs and outputs by [`TxOrdering`] variant.
     ///
     /// Uses a provided random number generator (rng).
-    pub fn sort_tx_with_aux_rand(self, tx: &mut Transaction, rng: &mut impl RngCore) {
+    pub fn sort_tx_with_aux_rand(&self, tx: &mut Transaction, rng: &mut impl RngCore) {
         match self {
             TxOrdering::Untouched => {}
             TxOrdering::Shuffle => {
@@ -800,6 +823,13 @@ impl TxOrdering {
                 });
                 tx.output
                     .sort_unstable_by_key(|txout| (txout.value, txout.script_pubkey.clone()));
+            }
+            TxOrdering::Custom {
+                input_sort,
+                output_sort,
+            } => {
+                tx.input.sort_unstable_by(|a, b| input_sort(a, b));
+                tx.output.sort_unstable_by(|a, b| output_sort(a, b));
             }
         }
     }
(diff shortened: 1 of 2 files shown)


crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -763,16 +766,38 @@ impl fmt::Display for AddForeignUtxoError {
...
+    Custom {
+        /// Transaction inputs sort function
+        input_sort: Arc<TxVecSort<TxIn>>,
+        /// Transaction outputs sort function
+        output_sort: Arc<TxVecSort<TxOut>>,
+    },
+}
+
+impl std::fmt::Debug for TxOrdering {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
```
rustaceanrob: `core::fmt`

crates/wallet/src/wallet/mod.rs
```diff
@@ -1555,7 +1555,7 @@ impl Wallet {
         };
 
         // sort input/outputs according to the chosen algorithm
-        params.ordering.sort_tx_with_aux_rand(&mut tx, rng);
+        params.ordering.clone().sort_tx_with_aux_rand(&mut tx, rng);
```
rustaceanrob: Forgive my ignorance but why do we need clone here

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -763,16 +766,38 @@ impl fmt::Display for AddForeignUtxoError {
 #[cfg(feature = "std")]
 impl std::error::Error for AddForeignUtxoError {}
 
+type TxVecSort<T> = dyn Fn(&T, &T) -> std::cmp::Ordering;
+
```
rustaceanrob, line 770: Not sure the caller needs to know about the internal representation. I think `TxSort` is fine

rustaceanrob (commented):
Thanks for that detailed explanation. I think `Arc` is in line with what we are going for. I wonder how maintainers would feel to just remove BIP69 entirely since this allows for users to add it back if they must.

crates/wallet/src/wallet/mod.rs
nymius: For  the reasons above we need to use some kind of smart pointer inside `TxOrdering`. Smart pointers manage data outside of their `size`. On the other hand, and I'm quoting a Stack Overflow answer here:
> The Copy trait represents values that can be safely duplicated via memcpy

The idea here is, if you have to manage data outside of your `size` you cannot copy it with memcpy, so you cannot have smart pointers implementing `Copy`.

Before my changes to `TxOrdering`, because of the `sort_tx_with_aux_rand` signature (`sort_tx_with_aux_rand(self...`, notice self is owned by the function), and as `params` is later used in `create_tx`, Rust[1] performed a copy to avoid the partial move.

That's not possible anymore as `TxOrdering` cannot implement `Copy` with a smart pointer in its fields, so the new way to avoid the partial move is with `clone()`.

[1] Here I'm guessing based on the suggestion from clippy.

crates/wallet/src/wallet/tx_builder.rs
nymius: I wanted to avoid ambiguity. I don't know if there is any valid application for some `TxSort` actually sorting whole transactions.

Confirm me here and I will change it.

crates/wallet/src/wallet/mod.rs
rustaceanrob: Great thanks for clarifying

crates/wallet/src/wallet/tx_builder.rs
rustaceanrob: > actually sorting whole transactions

Can you elaborate on this?

notmandatory:
>  I wonder how maintainers would feel to just remove BIP69 entirely since this allows for users to add it back if they must.

We discussed on the dev call today and the consensus is to remove bip69 support completely instead of just deprecating it.

crates/wallet/src/wallet/tx_builder.rs
notmandatory: I think what @rustaceanrob is saying is that you don't need to mention `Vec` in the name of this type, it could just be `TxSort<T>`. It's a small thing but I agree there's no need to mention Vec in the name.

nymius: > > actually sorting whole transactions
> 
> Can you elaborate on this?

It's more of a naming approach, like _if you are specific enough, you don't risk to overlap with future changes_ that may require similar names. As both of you don't see it plausible, and I'm not strongly convinced about those future overlapping names, I agree to change it.

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -42,6 +42,9 @@ use alloc::{boxed::Box, rc::Rc, string::String, vec::Vec};
 use core::cell::RefCell;
 use core::fmt;
 
+use std::sync::Arc;
```
notmandatory: 🔧 Need to change the `Arc` type to be compatible with no-std (and WASM).

```suggestion
use alloc::sync::Arc;
```

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -948,6 +983,63 @@ mod test {
...
+                        &txin.script_pubkey.clone().into_bytes()[..],
+                        shared_secret.as_bytes(),
+                    ]
+                    .concat(),
+                )
+            };
+            secret_digest_from_txout(tx_a).cmp(&secret_digest_from_txout(tx_b))
+        });
+
+        let custom_bip69_ordering_1 = TxOrdering::Custom {
```
notmandatory: ⛏️ A nit,  I wouldn't call these variables `custom_bip69_ordering_?` since bip69 is only for a specific deterministic lexicographical sort, and not this one with the hashing. Calling these `custom_ordering_?` would be better.

notmandatory (commented):
I have just a couple small comments, this one looks almost ready.

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -910,13 +930,28 @@ mod test {
     }
 
     #[test]
-    fn test_output_ordering_bip69() {
+    fn test_output_ordering_custom_but_bip69() {
```
notmandatory, line 933: 👍 Glad you were able to keep this test by reproducing the bip69 ordering with custom functions.

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -763,16 +766,33 @@ impl fmt::Display for AddForeignUtxoError {
 #[cfg(feature = "std")]
 impl std::error::Error for AddForeignUtxoError {}
 
+type TxSort<T> = dyn Fn(&T, &T) -> std::cmp::Ordering;
```
notmandatory: 🔧 Need to change the `Ordering` type to be compatible with no-std (and WASM).

```suggestion
type TxSort<T> = dyn Fn(&T, &T) -> core::cmp::Ordering;
```

crates/wallet/src/wallet/tx_builder.rs
nymius: Totally, probably leftovers of the `Bip69Lexicographic` test I use as template.

nymius:
As you are the best to judge if your requests have been fulfilled, @notmandatory @rustaceanrob, please, consider resolving conversations as we move this forward.

I plan to squash commits once we agree the work here is final.

crates/wallet/tests/wallet.rs
```diff
@@ -956,27 +956,6 @@ fn test_create_tx_drain_to_dust_amount() {
     builder.finish().unwrap();
 }
 
-#[test]
-fn test_create_tx_ordering_respected() {
```
notmandatory, line 968: 🔧 rather than removing this test can you just copy/paste the code for the `custom_bip69_ordering` and use that here? I don't want to lose the coverage this tests give us that confirms the `TxBuilder` uses the sorting you give it.

crates/wallet/src/wallet/mod.rs
notmandatory: 🔧 You should be able to remove this `.clone()` by changing the `TxOrdering` functions to borrow `&self` instead of taking ownership of `self`. 
```rust
impl TxOrdering {

    pub fn sort_tx(&self, tx: &mut Transaction) ...
     
    pub fn sort_tx_with_aux_rand(&self, tx: &mut Transaction, rng: &mut impl RngCore) ...
}
```

notmandatory:
When you're ready for a final review be sure to rebase and squash the commits into one or two.

nymius:
Rebased and updated. There is something extra you would like to see in the docs?

crates/wallet/src/wallet/mod.rs
notmandatory: @nymius what do you think about removing this `.clone()` ?

nymius: Great! I thought there were stronger reasons to take ownership, but If that's not the case, I prefer the borrow.

rustaceanrob:
LGTM

notmandatory (approved):
ACK 3e1fc40adf643043c3fa3ccee2b755340e0e4729

notmandatory:
@nymius I you need to sign your commits before I can merge this PR.

notmandatory:
I signed the commits for @nymius .

PR #1488, opened 2024-06-25
Add generic to `LocalChain`-based types

evanlinjin:
### Description

This is so that we can do future upgrades to `LocalChain` (and associated types) in a backwards-compatible way.

@LLFourn mentioned that `bdk_chain` changesets can be non-changing. A `LocalChain` with monotone changesets can be a different type completely (for example).

However, adding "metadata" to localchain is important for merkle proofs (for electrum, cbf, etc) as we need to store the header somewhere - and it makes sense for it to be in `LocalChain`.

### Notes to the reviewers

I will add more info later

### Changelog notice

* Added a generic `B` to `LocalChain` types.

#### All Submissions:


evanlinjin:
I just realized that since `local_chain::ChangeSet<BlockId>` is the same as `local_chain::ChangeSet`, we actually do not need this PR. We can do this change later and still be backwards compatible.

PR #1489, opened 2024-06-25
feat(electrum)!: Update `bdk_electrum` to use merkle proofs

### LINKED ISSUE #980: Update `bdk_electrum` to use electrum's merkle proof API
**Describe the enhancement**

As mentioned by @LLFourn in https://github.com/bitcoindevkit/bdk/pull/976#discussion_r1193920143:

> ... I think electrum really intends for you to use the merkle proof API. Best solution is to do that and attach confirmation time of each block to `LocalChain`.

**Additional context**

* This should be attempted after #976 is merged.
~* This may require changing `LocalChain`.~

LagginTimes:
Fixes #980.

### Description

This PR is the first step in reworking `bdk_electrum` to use merkle proofs. When we fetch a transaction, we now also obtain the merkle proof and block header for verification. We then insert an anchor only after validation that the transaction exists in a confirmed block. The loop logic that previously existed in `full_scan` to account for re-orgs has also been removed as part of this rework.

This is a breaking change because `graph_update`s now provide the full `ConfirmationTimeHeightAnchor` type. This removes the need for the `ElectrumFullScanResult` and `ElectrumSyncResult` structs that existed only to provide the option for converting the anchor type from `ConfirmationHeightAnchor` into `ConfirmationTimeHeightAnchor`.

### Changelog notice

* `ConfirmationTimeHeightAnchor` and `ConfirmationHeightAnchor` have been removed.
* `ConfirmationBlockTime` has been introduced as a new anchor type.
* `bdk_electrum`'s `full_scan` and `sync` now return `graph_update`s with `ConfirmationBlockTime`.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/chain_data.rs
+++ b/crates/chain/src/chain_data.rs
@@ -74,11 +74,11 @@ impl ConfirmationTime {
     }
 }
 
-impl From<ChainPosition<ConfirmationTimeHeightAnchor>> for ConfirmationTime {
-    fn from(observed_as: ChainPosition<ConfirmationTimeHeightAnchor>) -> Self {
+impl From<ChainPosition<ConfirmationBlockTime>> for ConfirmationTime {
+    fn from(observed_as: ChainPosition<ConfirmationBlockTime>) -> Self {
         match observed_as {
             ChainPosition::Confirmed(a) => Self::Confirmed {
-                height: a.confirmation_height,
+                height: a.block_id.height,
                 time: a.confirmation_time,
             },
             ChainPosition::Unconfirmed(last_seen) => Self::Unconfirmed { last_seen },
@@ -145,9 +145,7 @@ impl From<(&u32, &BlockHash)> for BlockId {
     }
 }
 
-/// An [`Anchor`] implementation that also records the exact confirmation height of the transaction.
-///
-/// Note that the confirmation block and the anchor block can be different here.
+/// An [`Anchor`] implementation that also records the exact confirmation time of the transaction.
 ///
 /// Refer to [`Anchor`] for more details.
 #[derive(Debug, Default, Clone, PartialEq, Eq, Copy, PartialOrd, Ord, core::hash::Hash)]
@@ -156,70 +154,27 @@ impl From<(&u32, &BlockHash)> for BlockId {
     derive(serde::Deserialize, serde::Serialize),
     serde(crate = "serde_crate")
 )]
-pub struct ConfirmationHeightAnchor {
-    /// The exact confirmation height of the transaction.
-    ///
-    /// It is assumed that this value is never larger than the height of the anchor block.
-    pub confirmation_height: u32,
+pub struct ConfirmationBlockTime {
     /// The anchor block.
-    pub anchor_block: BlockId,
-}
-
-impl Anchor for ConfirmationHeightAnchor {
-    fn anchor_block(&self) -> BlockId {
-        self.anchor_block
-    }
-
-    fn confirmation_height_upper_bound(&self) -> u32 {
-        self.confirmation_height
-    }
-}
-
-impl AnchorFromBlockPosition for ConfirmationHeightAnchor {
-    fn from_block_position(_block: &bitcoin::Block, block_id: BlockId, _tx_pos: usize) -> Self {
-        Self {
-            anchor_block: block_id,
-            confirmation_height: block_id.height,
-        }
-    }
-}
-
-/// An [`Anchor`] implementation that also records the exact confirmation time and height of the
-/// transaction.
-///
-/// Note that the confirmation block and the anchor block can be different here.
-///
-/// Refer to [`Anchor`] for more details.
-#[derive(Debug, Default, Clone, PartialEq, Eq, Copy, PartialOrd, Ord, core::hash::Hash)]
-#[cfg_attr(
-    feature = "serde",
-    derive(serde::Deserialize, serde::Serialize),
-    serde(crate = "serde_crate")
-)]
-pub struct ConfirmationTimeHeightAnchor {
-    /// The confirmation height of the transaction being anchored.
-    pub confirmation_height: u32,
+    pub block_id: BlockId,
     /// The confirmation time of the transaction being anchored.
     pub confirmation_time: u64,
-    /// The anchor block.
-    pub anchor_block: BlockId,
 }
 
-impl Anchor for ConfirmationTimeHeightAnchor {
+impl Anchor for ConfirmationBlockTime {
     fn anchor_block(&self) -> BlockId {
-        self.anchor_block
+        self.block_id
     }
 
     fn confirmation_height_upper_bound(&self) -> u32 {
-        self.confirmation_height
+        self.block_id.height
     }
 }
 
-impl AnchorFromBlockPosition for ConfirmationTimeHeightAnchor {
+impl AnchorFromBlockPosition for ConfirmationBlockTime {
     fn from_block_position(block: &bitcoin::Block, block_id: BlockId, _tx_pos: usize) -> Self {
         Self {
-            anchor_block: block_id,
-            confirmation_height: block_id.height,
+            block_id,
             confirmation_time: block.header.time as _,
         }
     }
@@ -305,19 +260,19 @@ mod test {
 
     #[test]
     fn chain_position_ord() {
-        let unconf1 = ChainPosition::<ConfirmationHeightAnchor>::Unconfirmed(10);
-        let unconf2 = ChainPosition::<ConfirmationHeightAnchor>::Unconfirmed(20);
-        let conf1 = ChainPosition::Confirmed(ConfirmationHeightAnchor {
-            confirmation_height: 9,
-            anchor_block: BlockId {
-                height: 20,
+        let unconf1 = ChainPosition::<ConfirmationBlockTime>::Unconfirmed(10);
+        let unconf2 = ChainPosition::<ConfirmationBlockTime>::Unconfirmed(20);
+        let conf1 = ChainPosition::Confirmed(ConfirmationBlockTime {
+            confirmation_time: 20,
+            block_id: BlockId {
+                height: 9,
                 ..Default::default()
             },
         });
-        let conf2 = ChainPosition::Confirmed(ConfirmationHeightAnchor {
-            confirmation_height: 12,
-            anchor_block: BlockId {
-                height: 15,
+        let conf2 = ChainPosition::Confirmed(ConfirmationBlockTime {
+            confirmation_time: 15,
+            block_id: BlockId {
+                height: 12,
                 ..Default::default()
             },
         });
--- a/crates/chain/tests/test_tx_graph.rs
+++ b/crates/chain/tests/test_tx_graph.rs
@@ -7,7 +7,7 @@ use bdk_chain::{
     collections::*,
     local_chain::LocalChain,
     tx_graph::{ChangeSet, TxGraph},
-    Anchor, BlockId, ChainOracle, ChainPosition, ConfirmationHeightAnchor, Merge,
+    Anchor, BlockId, ChainOracle, ChainPosition, ConfirmationBlockTime, Merge,
 };
 use bitcoin::{
     absolute, hashes::Hash, transaction, Amount, BlockHash, OutPoint, ScriptBuf, SignedAmount,
@@ -935,7 +935,7 @@ fn test_chain_spends() {
         ..common::new_tx(0)
     };
 
-    let mut graph = TxGraph::<ConfirmationHeightAnchor>::default();
+    let mut graph = TxGraph::<ConfirmationBlockTime>::default();
 
     let _ = graph.insert_tx(tx_0.clone());
     let _ = graph.insert_tx(tx_1.clone());
@@ -944,9 +944,9 @@ fn test_chain_spends() {
     for (ht, tx) in [(95, &tx_0), (98, &tx_1)] {
         let _ = graph.insert_anchor(
             tx.compute_txid(),
-            ConfirmationHeightAnchor {
-                anchor_block: tip.block_id(),
-                confirmation_height: ht,
+            ConfirmationBlockTime {
+                block_id: tip.get(ht).unwrap().block_id(),
+                confirmation_time: 100,
             },
         );
     }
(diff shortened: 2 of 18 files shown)


notmandatory:
Since this doesn't touch the wallet APIs I moved it to the beta milestone.

LagginTimes:
> Since this doesn't touch the wallet APIs I moved it to the beta milestone.

I'm okay with this!

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -371,120 +287,90 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+        &self,
+        graph_update: &mut TxGraph<ConfirmationTimeHeightAnchor>,
+        txid: Txid,
+        confirmation_height: i32,
+    ) -> Result<(), Error> {
+        if let Ok(merkle_res) = self
+            .inner
+            .transaction_get_merkle(&txid, confirmation_height as usize)
+        {
+            let header = self.inner.block_header(merkle_res.block_height)?;
```
LLFourn: I think we should have a `block_header_cache` here too. Then you can get the last block header we had at that height. If the merkle proof validation fails we can fetch it again, and validate against the new one. This will reduce the number of requests we make to electrum by a bit.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -371,120 +287,90 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+            let is_confirmed_tx = electrum_client::utils::validate_merkle_proof(
+                &txid,
+                &header.merkle_root,
+                &merkle_res,
+            );
+
+            if is_confirmed_tx {
+                let _ = graph_update.insert_anchor(
+                    txid,
+                    ConfirmationTimeHeightAnchor {
```
LLFourn: This seems like this is a suboptimal anchor to use since you're entering the same `block_height` twice. It would actually be good if you could delete `ConfirmationTimeHeightAnchor` and `ConfirmationHeightAnchor` from the codebase now since both electrum and esplora will report the exact block that the tx is in with this PR. It only needs to be recorded once.

For now this would be good to use for both:

```rust
struct ConfirmationBlockTime {
    block_id: BlockId,
    confirmation_time: u64
}
```

Later we can add the block position to this and maybe even the proof. 

Deleting `ConfirmationTimeHeight` anchor would need ACK from @evanlinjin  and @notmandatory since it's used by `Wallet`. Just for background the only reason `ConfirmationTimeHeightAnchor` exists is because of how wacky our electrum client was before this PR. I think simplifying the anchors to make use of this should be done ASAP.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -371,120 +287,90 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+        graph_update: &mut TxGraph<ConfirmationTimeHeightAnchor>,
+        txid: Txid,
+        confirmation_height: i32,
+    ) -> Result<(), Error> {
+        if let Ok(merkle_res) = self
+            .inner
+            .transaction_get_merkle(&txid, confirmation_height as usize)
+        {
+            let header = self.inner.block_header(merkle_res.block_height)?;
+            let is_confirmed_tx = electrum_client::utils::validate_merkle_proof(
```
LLFourn: Is it possible to use rust bitcoin's merkle proof validator for this? Why does electrum have its own validation logic?

LLFourn (commented):
Thanks this simplifies a lot. Left a few comments around the anchors/merkle proof bit.

crates/electrum/src/bdk_electrum_client.rs
LLFourn: If we don't want to delete the old anchors we should still return the correct thing from here. It can be `Into` converted into the old one later.

evanlinjin: ACK

notmandatory:
@LagginTimes oops, I didn't read all the comments before, looks like the wallet API will need to be updated so I've moved this back to the alpha milestone.

crates/electrum/src/bdk_electrum_client.rs
notmandatory: I support simplifying the anchors now. I'm happy to do whatever changes are required for to the sqlite storage crate. 

cc: @matthiasdebernardini this anchor change will also impact the postgres store you're working on.

crates/electrum/src/bdk_electrum_client.rs
LagginTimes: The validation logic being used right now was implemented as a utility in `rust-electrum-client`, which was needed for LDK's electrum integration:
https://github.com/bitcoindevkit/rust-electrum-client/pull/122

Would using rust bitcoin be more advantageous in this scenario?

LLFourn: I have no idea. I am just curious why @tnull added that when there is presumably this algorithm is implemented in `rust-bitcoin`.

tnull: IIRC, Electrum's `get_merkle`/`GetMerkleRes` doesn't provide sufficient/slightly different data, so that we couldn't construct `bitcoin::MerkleBlock`/ `PartialMerkleTree` to use upstream's validation logic (which of course would have been our preferred way).

crates/chain/src/chain_data.rs
```diff
@@ -156,70 +156,27 @@ impl From<(&u32, &BlockHash)> for BlockId {
     derive(serde::Deserialize, serde::Serialize),
     serde(crate = "serde_crate")
 )]
-pub struct ConfirmationHeightAnchor {
-    /// The exact confirmation height of the transaction.
-    ///
-    /// It is assumed that this value is never larger than the height of the anchor block.
-    pub confirmation_height: u32,
+pub struct ConfirmationBlockTime {
```
ValuedMammal, line 157: docs: this line no longer applies

> Note that the confirmation block and the anchor block can be different here.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -944,9 +944,12 @@ fn test_chain_spends() {
...
         let _ = graph.insert_anchor(
             tx.compute_txid(),
-            ConfirmationHeightAnchor {
-                anchor_block: tip.block_id(),
-                confirmation_height: ht,
+            ConfirmationBlockTime {
+                block_id: BlockId {
+                    hash: tip.get(ht).unwrap().hash(),
+                    height: ht,
+                },
```
ValuedMammal:
```suggestion
                block_id: tip.get(ht).unwrap().block_id(),
```

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -88,87 +120,32 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
-            let server_blockhash = self.inner.block_header(tip.height() as usize)?.block_hash();
-            if tip.hash() != server_blockhash {
-                continue; // reorg
+    ) -> Result<FullScanResult<K>, Error> {
+        let (tip, latest_blocks) =
+            fetch_tip_and_latest_blocks(&self.inner, request.chain_tip.clone())?;
+        let mut graph_update = TxGraph::<ConfirmationBlockTime>::default();
+        let mut last_active_indices = BTreeMap::<K, u32>::new();
+
+        for (keychain, keychain_spks) in request.spks_by_keychain {
```
ValuedMammal:
```suggestion
        for (keychain, spks) in request.spks_by_keychain {
```

ValuedMammal: 📌  The docs for `full_scan` are a bit lacking with respect to `fetch_prev_txouts`, see https://github.com/bitcoindevkit/bdk/pull/1494#issuecomment-2204199786

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -227,70 +203,46 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
     /// Checkpoints (in `cps`) are used to create anchors. The `tx_cache` is self-explanatory.
     fn populate_with_spks<I: Ord + Clone>(
         &self,
-        cps: &BTreeMap<u32, CheckPoint>,
-        graph_update: &mut TxGraph<ConfirmationHeightAnchor>,
-        spks: &mut impl Iterator<Item = (I, ScriptBuf)>,
+        graph_update: &mut TxGraph<ConfirmationBlockTime>,
+        mut spks: impl Iterator<Item = (I, ScriptBuf)>,
```
ValuedMammal: docs: "cps" and "tx_cache" are no longer relevant parameters, same for `populate_with_outpoints` below

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -371,120 +317,100 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+                        confirmation_time: header.time as u64,
+                        block_id: BlockId {
+                            height: merkle_res.block_height as u32,
+                            hash: header.block_hash(),
+                        },
+                    },
+                );
+            }
+        }
+        Ok(())
```
ValuedMammal, line 373: question: Is there a scenario where the `validate_merkle_for_anchor` function could fail and we'd want to know about it? in other words should we be worried if validating the proof fails even after updating the header?

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -527,6 +453,7 @@ fn construct_update_tip(
     let agreement_height = agreement_cp.as_ref().map(CheckPoint::height);
 
     let new_tip = new_blocks
+        .clone()
         .into_iter()
```
ValuedMammal: if we just use `.iter()` we can remove the call to `clone`

crates/electrum/tests/test_electrum.rs
```diff
@@ -155,13 +152,11 @@ fn tx_can_become_unconfirmed_after_reorg() -> anyhow::Result<()> {
...
-            SyncRequest::from_chain_tip(recv_chain.tip()).chain_spks([spk_to_track.clone()]),
-            5,
-            false,
-        )?
-        .with_confirmation_time_height_anchor(&client)?;
+    let update = client.sync(
+        SyncRequest::from_chain_tip(recv_chain.tip()).chain_spks([spk_to_track.clone()]),
+        5,
+        false,
+    )?;
```
ValuedMammal, line 378: We should be able to make assertions on the expected graph anchors, something like:

```diff
     // Create transactions that are tracked by our receiver.
+    let mut txids = vec![];
+    let mut blockhashes = vec![];
     for _ in 0..REORG_COUNT {
-        env.send(&addr_to_track, SEND_AMOUNT)?;
-        env.mine_blocks(1, None)?;
+        txids.push(env.send(&addr_to_track, SEND_AMOUNT)?);
+        blockhashes.extend(env.mine_blocks(1, None)?);
     }
 
     // Sync up to tip.
     env.wait_until_electrum_sees_block()?;
     let update = client.sync(
@@ -163,10 +161,17 @@ fn tx_can_become_unconfirmed_after_reorg() -> anyhow::Result<()> {
         .map_err(|err| anyhow::anyhow!("LocalChain update error: {:?}", err))?;
     let _ = recv_graph.apply_update(update.graph_update.clone());
 
     // Retain a snapshot of all anchors before reorg process.
     let initial_anchors = update.graph_update.all_anchors();
+    let anchors: Vec<_> = initial_anchors.iter().cloned().collect();
+    assert_eq!(anchors.len(), REORG_COUNT);
+    for i in 0..REORG_COUNT {
+        let (anchor, txid) = anchors[i];
+        assert_eq!(anchor.anchor_block().hash, blockhashes[i]);
+        assert_eq!(txid, txids[i]);
+    }
```

ValuedMammal (commented):
I did a first pass, mostly documentation nits. Sorry I was reviewing while you were pushing, so comments might be outdated

notmandatory:
@LagginTimes hey sorry I just merged some smaller PRs and now this one needs a rebase.

evanlinjin:
I think we should not wait for complete tests before merging this PR. As long as the examples work fine it is good. This is for "beta" which does not need to be bug free. Also having this PR merged means we can work on the other anchor changes (which I need to create a ticket for).

evanlinjin:
Also, the anchor changes that I am suggesting is going to conflict with `ConfirmationBlockTime`.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -190,117 +167,87 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
 
-            for ((spk_index, spk), spk_history) in spks.into_iter().zip(spk_histories) {
+            for ((spk_index, _spk), spk_history) in spks.into_iter().zip(spk_histories) {
                 if spk_history.is_empty() {
-                    scanned_spks.insert(spk_index, (spk, false));
-                    unused_spk_count += 1;
-                    if unused_spk_count > stop_gap {
-                        return Ok(scanned_spks);
+                    unused_spk_count = unused_spk_count.saturating_add(1);
+                    if unused_spk_count >= stop_gap {
```
ValuedMammal, line 227: 💯

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -527,10 +449,13 @@ fn construct_update_tip(
     let agreement_height = agreement_cp.as_ref().map(CheckPoint::height);
 
     let new_tip = new_blocks
-        .into_iter()
+        .iter()
         // Prune `new_blocks` to only include blocks that are actually new.
-        .filter(|(height, _)| Some(*height) > agreement_height)
-        .map(|(height, hash)| BlockId { height, hash })
+        .filter(|(height, _)| Some(*<&u32>::clone(height)) > agreement_height)
```
ValuedMammal, line 454: nit, feel free to ignore

```suggestion
        .filter(|(&height, _)| Some(height) > agreement_height)
```

crates/electrum/tests/test_electrum.rs
```diff
@@ -19,6 +21,222 @@ fn get_balance(
     Ok(balance)
 }
 
+#[test]
+pub fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
```
ValuedMammal, line 25: 📌 If I recall, the name of this test is a relic from before the `sync` and `full_scan` concepts were developed. I might now call it `sync_should_update_chain_and_tx_graph` or similar, but not terribly important at the moment.

ValuedMammal (approved):
ACK 1a62488abfd6213c2780c1465aebe01dfbb6b20a

crates/electrum/src/bdk_electrum_client.rs
notmandatory: 📌  This is a good question and it makes sense to have a runtime option or feature to throw an error if merkle proof validation fails. But I don't think this has to be done in this PR or for the beta release since it will only change the API for the `bdk_electrum` crate. @ValuedMammal would you mind just making an issue for this?

notmandatory (approved):
ACK 1a62488abfd6213c2780c1465aebe01dfbb6b20a

Thanks for also adding the extra esplora tests for electrum!

PR #1491, opened 2024-06-26
Enable selecting use-rustls-ring feature on electrum client

thunderbiscuit:
This PR is a companion to bitcoindevkit/rust-electrum-client#135. It enables choosing the `ring` dependency on rustls instead of the new default (as of 0.23.0) `aws-lc-rs`. The AWS dependency breaks the Android and Swift builds. I wrote a more detailed explanation on [#135](https://github.com/bitcoindevkit/rust-electrum-client/pull/135).

### Notes to the reviewers

Do not merge before:
- [x] [#135](https://github.com/bitcoindevkit/rust-electrum-client/pull/135) is merged
- [x] A new version of rust-electrum-client is released (will be 0.21.0) 
- [x] The dependency points to the new version of the client rather than my fork of it.

### Changelog notice

```md
Added
    - bdk_electrum now enables choosing either the `use-rustls` or `use-rustls-ring` feature
```

#### All Submissions:

### DIFF
--- a/crates/electrum/Cargo.toml
+++ b/crates/electrum/Cargo.toml
@@ -9,12 +9,14 @@ description = "Fetch data from electrum in the form BDK accepts"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [dependencies]
 bdk_chain = { path = "../chain", version = "0.17.0" }
-electrum-client = { version = "0.20" }
-#rustls = { version = "=0.21.1", optional = true, features = ["dangerous_configuration"] }
+electrum-client = { version = "0.21", features = ["proxy"], default-features = false }
 
 [dev-dependencies]
 bdk_testenv = { path = "../testenv", default-features = false }
+
+[features]
+default = ["use-rustls"]
+use-rustls = ["electrum-client/use-rustls"]
+use-rustls-ring = ["electrum-client/use-rustls-ring"]


crates/electrum/Cargo.toml
```diff
@@ -12,9 +12,17 @@ readme = "README.md"
...
+bdk_chain = { version = "0.16.0" }
+# Switch the lines below once https://github.com/bitcoindevkit/rust-electrum-client/pull/135 is merged and a new version is released
+# electrum-client = { version = "0.21", default-features = false }
+electrum-client = { git = "https://github.com/thunderbiscuit/rust-electrum-client/", branch = "feature/rustls-ring", default-features = false }
 
 [dev-dependencies]
 bdk_testenv = { path = "../testenv", default-features = false }
+
+[features]
+use-rustls = ["electrum-client/use-rustls"]
```
oleonardolima, line 21: @thunderbiscuit I'm wondering if we do need to expose the feature which uses the aws-lc-rs, or have it opinionated and only use ring one 🤔

thunderbiscuit: Hey thanks for taking a look! Yeah I'm not sure what's best here, but since the rustls has made the AWS one their default I figured that might be what people want to use if they can. Happy to change the PR to simply always use ring (which would simplify the flags).

thunderbiscuit:
Quick update @oleonardolima: I added a default feature that in turn uses the rustls default features on the electrum client. This way this PR doesn't break anyone's code while offering the option for ring.

crates/electrum/Cargo.toml
```diff
@@ -9,12 +9,16 @@ description = "Fetch data from electrum in the form BDK accepts"
...
+electrum-client = { git = "https://github.com/thunderbiscuit/rust-electrum-client/", branch = "feature/rustls-ring", default-features = false }
+# electrum-client = { version = "0.21", default-features = false }
 
 [dev-dependencies]
 bdk_testenv = { path = "../testenv", default-features = false }
+
+[features]
+default = ["use-rustls"]
+use-rustls = ["electrum-client/proxy", "electrum-client/use-rustls"]
+use-rustls-ring = ["electrum-client/use-rustls-ring"]
```
oleonardolima, line 22: @thunderbiscuit maybe to maintain a standard the electrum/use-rustls-ring should not have the proxy feature by default an be explicitly expected/passed ?

crates/electrum/Cargo.toml
oleonardolima: Sure, that's a good approach

praveenperera:
Is it possible to set up the features so that you don't have to set `default-features = false`, when you want `use-rustls-ring`?

Not a big deal but a little nicer for the user

oleonardolima:
> Is it possible to set up the features so that you don't have to set `default-features = false`, when you want `use-rustls-ring`?
> 
> 
> 
> Not a big deal but a little nicer for the user

IIRC its doing this way, in order to match default features being the usage of dependencies default features too. (Although I agree and think it's better ergonomics for the BDKs user to not have to default-features = false

praveenperera:
Hey @thunderbiscuit do you know if it be possible to get this merged in soon? I was using bdk_electrum set to this branch, but i'm getting version mismatch errors on bdk_chain package.

crates/electrum/Cargo.toml
thunderbiscuit: Yes good catch! Fixed now.

thunderbiscuit:
Sorry folks I'm coming back to this. Would love to get it merged for the summit.

@praveenperera 
> Is it possible to set up the features so that you don't have to set default-features = false, when you want use-rustls-ring?

The way to do this would be to disable rustls as a default feature completely and force users to choose either the `use-rustls` or `use-rustls-ring` feature. The reason why I chose to not do it this way is that this would break all existing users' code (the default rustls feature gets removed under their feet).

Sorry to tag but here are some interested parties in this PR:
@oleonardolima 
@ValuedMammal 
@LLFourn 
@notmandatory 
@praveenperera 

See the sister discussion on https://github.com/bitcoindevkit/rust-electrum-client/pull/135. This will fix a number of users' builds (BDK bindings, Anchorwatch, Cove). I want to make sure we get it in for the next release.

oleonardolima (approved):
utACK 9a8fc8fcdf706647a016578b3aa73695feb2bacc

praveenperera:
@thunderbiscuit got error. 

```
44 | extern crate webpki_roots;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate
```

`use-rustls-ring` does not enable the webpki_roots feature

```
use-rustls-ring = ["rustls/ring", "rustls/logging", "rustls/std", "rustls/tls12"]
```

But it is imported here

```
#[cfg(any(
    feature = "default",
    feature = "use-rustls",
    feature = "use-rustls-ring"
))]
extern crate webpki_roots;
```

crates/electrum/Cargo.toml
oleonardolima: You should now be able to use the `0.21` release, use the `electrum-client/proxy` and this PR should be ready to go 🚀

crates/electrum/Cargo.toml
```diff
@@ -9,12 +9,16 @@ description = "Fetch data from electrum in the form BDK accepts"
...
 readme = "README.md"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
 [dependencies]
 bdk_chain = { path = "../chain", version = "0.17.0" }
-electrum-client = { version = "0.20" }
-#rustls = { version = "=0.21.1", optional = true, features = ["dangerous_configuration"] }
+# Switch the lines below once a new version is released
+# electrum-client = { version = "0.21", default-features = false }
```
notmandatory: I'll close #1538 so you can make the change in this PR ok?

thunderbiscuit:
Rebased and ready for final review.

crates/electrum/Cargo.toml
```diff
@@ -9,12 +9,14 @@ description = "Fetch data from electrum in the form BDK accepts"
...
-electrum-client = { version = "0.20" }
-#rustls = { version = "=0.21.1", optional = true, features = ["dangerous_configuration"] }
+electrum-client = { version = "0.21", default-features = false }
 
 [dev-dependencies]
 bdk_testenv = { path = "../testenv", default-features = false }
+
+[features]
+default = ["use-rustls"]
+use-rustls = ["electrum-client/proxy", "electrum-client/use-rustls"]
```
tnull: The addition of `"electrum-client/proxy"` here seems unrelated? IMO, we should just enable it by default in the `electrum-client` above as unfortunately it is a mandatory feature really (see https://github.com/bitcoindevkit/rust-electrum-client/issues/91).

oleonardolima: I agree with tnull's comment. I didn't notice it before, but I think we should also have it as the default as it's currently "mandatory/required" to properly work.

I don't think the way we have it on `rust-electrum-client` is the way to go, we should fix it there.

crates/electrum/Cargo.toml
```diff
@@ -9,12 +9,14 @@ description = "Fetch data from electrum in the form BDK accepts"
...
+use-rustls-ring = ["electrum-client/proxy", "electrum-client/use-rustls-ring"]
```
oleonardolima: Leaving a suggestion below, to make it easy for a quick change:

```suggestion
default = ["electrum-client/proxy", "use-rustls"]
use-rustls = ["electrum-client/use-rustls"]
use-rustls-ring = ["electrum-client/use-rustls-ring"]
```

oleonardolima: Oh, scratch that ^

You can basically add it like `electrum-client = { version = "0.21", features = ["proxy"], default-features = false }` instead, right ?

crates/electrum/Cargo.toml
thunderbiscuit: Take a look at bitcoindevkit/rust-electrum-client#143. If that's the best way forward, we can update the bdk_electrum crate once that's merged (but maybe not in this PR? The bindings are stopped because of this so I'd like to get a fix in for the next beta release).

crates/electrum/Cargo.toml
thunderbiscuit: Yep thanks for the suggestion. Committed.

thunderbiscuit:
Thanks for the review @oleonardolima @tnull! I just got to your comments today. Rebased and fixed now.

notmandatory (approved):
utACK f965f95721c719ac755da07aa997d9b858583856

LLFourn (approved):

LLFourn:
ACK f965f95721c719ac755da07aa997d9b858583856

crates/electrum/Cargo.toml
oleonardolima: Sure, I'll take another look there. I agree, we can bump the `bdk_electrum` `electrum-client` dependency in another PR, and move forward with this one and bindings.

oleonardolima (approved):
ACK f965f95721c719ac755da07aa997d9b858583856

PR #1493, opened 2024-06-27
refactor(chain)!: Create module `indexer`

### LINKED ISSUE #1317: [chain] keychain module doesn't need to exist

ValuedMammal:
Introduce module `indexer` and replace old keychain module with `balance.rs`

fixes #1317 
replaces #1175 

### Changelog notice

Removed `keychain` module from `bdk_chain`. The `Balance` type is now re-exported from the top level

#### All Submissions:

### DIFF
--- a/crates/chain/src/indexer.rs
+++ b/crates/chain/src/indexer.rs
@@ -1,0 +1,33 @@
+//! [`Indexer`] provides utilities for indexing transaction data.
+
+use bitcoin::{OutPoint, Transaction, TxOut};
+
+#[cfg(feature = "miniscript")]
+pub mod keychain_txout;
+pub mod spk_txout;
+
+/// Utilities for indexing transaction data.
+///
+/// Types which implement this trait can be used to construct an [`IndexedTxGraph`].
+/// This trait's methods should rarely be called directly.
+///
+/// [`IndexedTxGraph`]: crate::IndexedTxGraph
+pub trait Indexer {
+    /// The resultant "changeset" when new transaction data is indexed.
+    type ChangeSet;
+
+    /// Scan and index the given `outpoint` and `txout`.
+    fn index_txout(&mut self, outpoint: OutPoint, txout: &TxOut) -> Self::ChangeSet;
+
+    /// Scans a transaction for relevant outpoints, which are stored and indexed internally.
+    fn index_tx(&mut self, tx: &Transaction) -> Self::ChangeSet;
+
+    /// Apply changeset to itself.
+    fn apply_changeset(&mut self, changeset: Self::ChangeSet);
+
+    /// Determines the [`ChangeSet`](Indexer::ChangeSet) between `self` and an empty [`Indexer`].
+    fn initial_changeset(&self) -> Self::ChangeSet;
+
+    /// Determines whether the transaction should be included in the index.
+    fn is_tx_relevant(&self, tx: &Transaction) -> bool;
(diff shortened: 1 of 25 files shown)


LLFourn:
I wonder what we think about:

```
indexer.rs // the trait in here
indexer/
  ├── keychain_txout.rs
  └── spk_txout.rs
```

ValuedMammal:
@LLFourn Should indexer be a `pub mod` or do we just re export everything under it?

storopoli:
I think re-export, despite pedantic, is more future-proof.

crates/chain/src/indexer.rs
```diff
@@ -0,0 +1,31 @@
+use bitcoin::{OutPoint, Transaction, TxOut};
```
notmandatory, line 3: ❓ Why not make this file `src/indexer/mod.rs` ?

notmandatory (commented):
Overall looks good, just one question.

crates/chain/src/indexer.rs
ValuedMammal: Done in 884ac9009350d2e5fb0b7459935905216d5fcf2e

ValuedMammal:
Rebased onto a112b4d9

notmandatory (approved):
ACK 884ac9009350d2e5fb0b7459935905216d5fcf2e

crates/chain/src/indexer/mod.rs
```diff
@@ -0,0 +1,31 @@
+use bitcoin::{OutPoint, Transaction, TxOut};
```
LLFourn: Can this file be `src/indexer.rs` instead. It's easier to find it that way.

LLFourn:
> @LLFourn Should indexer be a `pub mod` or do we just re export everything under it?

I think it must be `pub mod` otherwise you will be re-exporting the `ChangeSet` in the `keychain_txout.rs` file at the root! I would just make people import by `indexer::keychain_txout::KeychainTxOutIndex`.

I wouldn't mind dropping the `TxOut` from these which seems pretty redundant.

crates/chain/src/indexer/mod.rs
notmandatory: Sorry @ValuedMammal I thought `indexer/mod.rs` was the more common rust style. It was a long time ago but apparently now `src/indexer.rs` is the recommended way so I'm good with reverting back to `src/indexer.rs`. 
https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames

ValuedMammal:
- Fixed file indexer.rs
- Made `indexer` a public module

notmandatory (approved):
reACK c3fc1dd12337bf9bdb99323a51110b7ad6a00221

notmandatory:
I'd prefer to merge this as is rather than doing any more renaming at this point.

LLFourn (approved):
ACK c3fc1dd12337bf9bdb99323a51110b7ad6a00221

PR #1494, opened 2024-06-28
Add documentation for Electrum's full_scan/sync

thunderbiscuit:
This PR also serves as an issue; it can't really be merged as is. I just didn't want to open an issue and just ask for better docs and instead decided to open a PR with some unfinished new API docs. 

I am working on a page for the Book of BDK that focuses on bdk_wallet + bdk_electrum syncing. A few things have been unclear to me, and I think slight additions to the API docs would fix that for others.

~~1. I was wondering what exactly this `confirmation_time` field on the `bdk_chain::ConfirmationTimeHeightAnchor` type represents. After digging into it (at least for the electrum client), it represents the timestamp specified by the block header where the tx was confirmed. _Question: I'd like to confirm that this always the case, e.g. that this is the timestamp used in cases where your client is an Esplora or bitcoind RPC node?_ If so, my addition to the sentence will make sense and is ready to go.~~ Edit: this is no longer a type after the rebase.
2. I think it'd be great to add context to the `fetch_prev_txouts` argument on the full_scan and sync methods on the `BdkElectrumClient`. It says that this is necessary for "fee calculation". What does that mean? I assume it means "for calculating the fee rate on the given transactions"? (let me know if that's wrong). Even so, I'm left wondering what happens if I don't fetch them. Will my fee calculations be just plain wrong? Or will they be unavailable? A bit more context for the caller of the method would be great here. If someone knows the definite answer to this let me know and feel free to propose a doc line and I'll amend the commit!

### Checklists

### DIFF
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -108,12 +108,22 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
     /// returns updates for [`bdk_chain`] data structures.
     ///
     /// - `request`: struct with data required to perform a spk-based blockchain client full scan,
-    ///              see [`FullScanRequest`]
+    ///              see [`FullScanRequest`].
     /// - `stop_gap`: the full scan for each keychain stops after a gap of script pubkeys with no
-    ///              associated transactions
+    ///               associated transactions.
     /// - `batch_size`: specifies the max number of script pubkeys to request for in a single batch
-    ///              request
-    /// - `fetch_prev_txouts`: specifies whether or not we want previous `TxOut`s for fee
+    ///                 request.
+    /// - `fetch_prev_txouts`: specifies whether we want previous `TxOut`s for fee calculation.
+    ///                        Note that this requires additional calls to the Electrum server, but
+    ///                        is necessary for calculating the fee on a transaction if your wallet
+    ///                        does not own the inputs. Methods like [`Wallet.calculate_fee`] and
+    ///                        [`Wallet.calculate_fee_rate`] will return a
+    ///                        [`CalculateFeeError::MissingTxOut`] error if those `TxOut`s are not
+    ///                        present in the transaction graph.
+    ///
+    /// [`CalculateFeeError::MissingTxOut`]: bdk_chain::tx_graph::CalculateFeeError::MissingTxOut
+    /// [`Wallet.calculate_fee`]: https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.Wallet.html#method.calculate_fee
+    /// [`Wallet.calculate_fee_rate`]: https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.Wallet.html#method.calculate_fee_rate
     pub fn full_scan<K: Ord + Clone>(
         &self,
         request: FullScanRequest<K>,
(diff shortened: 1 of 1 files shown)


evanlinjin:
1. For all the chain sources right now, the confirmation time is the block header's timestamp. I think this should always be the defacto-definition (so, yes).
2. Fee here is for absolute fee and fee rate (you can't calculate fee rate without absolute fee). It's necessary for "fee calculation" because tx inputs do not contain amounts of prev outputs. So we need to fetch those prev outputs to do (output amounts - input amounts). If you don't include prev outputs, we cannot determine the fee (and associated methods will return `None`).

evanlinjin:
Thanks for the initiative on this, better docs are always better.

LLFourn:
> 2\. Fee here is for absolute fee and fee rate (you can't calculate fee rate without absolute fee). It's necessary for "fee calculation" because tx inputs do not contain amounts of prev outputs. So we need to fetch those prev outputs to do (output amounts - input amounts). If you don't include prev outputs, we cannot determine the fee (and associated methods will return `None`).

To clarify, this means that txs spending to your wallet from a foreign wallet will not have the input info and so you can't calculate the fee without `fetch_prev_txouts`. Txs spending from your wallet will have the input info and so you will be able to get the fee regardless of `fetch_prev_txouts`.

thunderbiscuit:
Thanks guys! I'll update the PR shortly.

thunderbiscuit:
This is what I ended up adding:
```md
    /// - `fetch_prev_txouts`: specifies whether we want previous `TxOut`s for fee calculation.
    ///                        Note that this requires additional calls to the Electrum server, but
    ///                        is necessary for calculating the fee on a transaction if your wallet
    ///                        does not own the inputs. Methods like [`Wallet.calculate_fee`] and
    ///                        [`Wallet.calculate_fee_rate`] will return a
    ///                        [`CalculateFeeError::MissingTxOut`] error if those `TxOut`s are not
    ///                        present in the transaction graph.
```

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -177,9 +183,14 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
-    /// - `fetch_prev_txouts`: specifies whether or not we want previous `TxOut`s for fee
-    ///              calculation
+    ///                 request
+    /// - `fetch_prev_txouts`: specifies whether we want previous `TxOut`s for fee calculation. Note
+    //                         that this requires additional calls to the Electrum server, but is
+    //                         necessary for calculating the fee on a transaction if your wallet
+    //                         does not own the inputs. Methods like Wallet.calculate_fee and
+    //                         Wallet.calculate_fee_rate will return a
+    //                         CalculateFeeError::MissingTxOut error if those TxOuts are not present
+    //                         in the transaction graph.
```
ValuedMammal: This reads well but maybe a little verbose. also lines seem to be missing an extra forward slash `/`

```rust
/// - `fetch_prev_txouts`: specifies whether we want previous `TxOut`s for fee calculation. Note
///                        that this requires additional calls to the Electrum server, but is
///                        necessary for calculating the fee on a transaction if your wallet
///                        does not own the inputs. Methods like [`TxGraph::calculate_fee`]
///                        will return an error if those `TxOut`s are missing from the
///                        transaction graph.
```

thunderbiscuit: Verbose is my middle name 😆; but people know what I mean when I'm done.

thunderbiscuit:
Rebased and ready to go.

crates/electrum/src/bdk_electrum_client.rs
ValuedMammal: Still it would help to have backticks and/or rustdoc links for any types that can be resolved such as `CalculateFeeError::MissingTxOut`

thunderbiscuit: Sounds good. I added the links for the error as well as the Wallet::calculate_fee and Wallet::calculate_fee_rate methods.

ValuedMammal (approved):
ACK 056f26ca611399512964bd69240e98c6f190845b

evanlinjin (approved):
ACK 056f26ca611399512964bd69240e98c6f190845b

LagginTimes:
ACK 056f26ca611399512964bd69240e98c6f190845b

thunderbiscuit:
Rebased.

evanlinjin (approved):
re-ACK 7b5657e23e17c6934773dfdff0ed214ff3b8f110

PR #1506, opened 2024-07-08
refactor(chain)!: Update KeychainTxOutIndex methods to use owned K and ScriptBuf

### LINKED ISSUE #1482: KeychainTxOutIndex API inconsistent with respect to references (should be owned)
Methods often take a reference to a `K` or an owned `K`. They often return a reference or sometimes an owned `K`. I think we should always just used owned. Also I think we should owned `ScriptBuf`'s rather than `Script`. The reason is they are both usually small and fine to copy:

- `K` is usually small and needs to be anyway if you derive lots of spks from a keychain (it is copied each time you do that into the underlying `SpkTxOutIndex`.
- script pubkeys are usually small unless you are doing stuff like DoSing the mempool/utxo set which we don't need to support.

It should be owned everywhere and this should be brought up to the `Wallet` API.

rustaceanrob:
### Description

Make all method signatures of `KeychainTxOutIndex` take owned `K` and use `ScriptBuf` instead of its borrowed counterpart `&Script`. Fixes #1482 

### Notes to the reviewers

Steve also added a CI fix as well

### Changelog notice

- Make all method signatures of `KeychainTxOutIndex` take owned `K`
- Update `KeychainTxOutIndex` methods to use `ScriptBuf`

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/indexer/keychain_txout.rs
+++ b/crates/chain/src/indexer/keychain_txout.rs
@@ -9,7 +9,7 @@ use crate::{
     DescriptorExt, DescriptorId, Indexed, Indexer, KeychainIndexed, SpkIterator,
 };
 use alloc::{borrow::ToOwned, vec::Vec};
-use bitcoin::{Amount, OutPoint, Script, ScriptBuf, SignedAmount, Transaction, TxOut, Txid};
+use bitcoin::{Amount, OutPoint, ScriptBuf, SignedAmount, Transaction, TxOut, Txid};
 use core::{
     fmt::Debug,
     ops::{Bound, RangeBounds},
@@ -251,14 +251,14 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// Return the script that exists under the given `keychain`'s `index`.
     ///
     /// This calls [`SpkTxOutIndex::spk_at_index`] internally.
-    pub fn spk_at_index(&self, keychain: K, index: u32) -> Option<&Script> {
+    pub fn spk_at_index(&self, keychain: K, index: u32) -> Option<ScriptBuf> {
         self.inner.spk_at_index(&(keychain.clone(), index))
     }
 
     /// Returns the keychain and keychain index associated with the spk.
     ///
     /// This calls [`SpkTxOutIndex::index_of_spk`] internally.
-    pub fn index_of_spk(&self, script: &Script) -> Option<&(K, u32)> {
+    pub fn index_of_spk(&self, script: ScriptBuf) -> Option<&(K, u32)> {
         self.inner.index_of_spk(script)
     }
 
@@ -489,7 +489,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     pub fn revealed_spks(
         &self,
         range: impl RangeBounds<K>,
-    ) -> impl Iterator<Item = KeychainIndexed<K, &Script>> {
+    ) -> impl Iterator<Item = KeychainIndexed<K, ScriptBuf>> + '_ {
         let start = range.start_bound();
         let end = range.end_bound();
         let mut iter_last_revealed = self
@@ -516,7 +516,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
             let (current_keychain, last_revealed) = current_keychain?;
 
             if current_keychain == keychain && Some(*index) <= last_revealed {
-                break Some(((keychain.clone(), *index), spk.as_script()));
+                break Some(((keychain.clone(), *index), spk.clone()));
             }
         })
     }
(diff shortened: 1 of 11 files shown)


notmandatory:
I pushed commit to also update `KeychainTxOutIndex` to use owned `ScriptBuf`s instead of `&Script`s.

notmandatory:
⛏️ @rustaceanrob one small thing, can you change your commit description to be a little more descriptive, something like:

```
refactor(chain)!: update KeychainTxOutIndex methods to use owned K
```

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -147,7 +147,7 @@ impl<K: Clone + Ord + Debug> Indexer for KeychainTxOutIndex<K> {
             if self.last_revealed.get(did) < Some(&index) {
                 self.last_revealed.insert(*did, index);
                 changeset.last_revealed.insert(*did, index);
-                self.replenish_inner_index(*did, &keychain, self.lookahead);
+                self.replenish_inner_index(*did, keychain, self.lookahead);
```
storopoli: Is this just removing an additional `&` since the `.scan_txout(I) -> &I` (returns a reference) and `keychain` here is already a reference?

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -254,14 +254,14 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// Return the script that exists under the given `keychain`'s `index`.
     ///
     /// This calls [`SpkTxOutIndex::spk_at_index`] internally.
-    pub fn spk_at_index(&self, keychain: K, index: u32) -> Option<&Script> {
+    pub fn spk_at_index(&self, keychain: K, index: u32) -> Option<ScriptBuf> {
         self.inner.spk_at_index(&(keychain.clone(), index))
```
storopoli, line 255: What happens if we enforce that `K` is also `Copy`.

Ref https://github.com/bitcoindevkit/bdk/blob/37cef1da7834d6d29f91c382ae841baf66764c51/crates/chain/src/indexer/keychain_txout.rs#L137

I see from the example in the docstrings that probably `K` can be trait-bounded to `Copy` and then we don't need all those `.clone()`

notmandatory: Limiting K to have to implement `Copy` I think would be too restrictive. For instance we wouldn't be able to use types like `String` as K or as fields of a K.

storopoli: Yes, then `clone()` is fine.

crates/chain/src/indexer/keychain_txout.rs
rustaceanrob: You are right I will fix that

notmandatory: A little higher up in the code (line 142) we've `cloned()` the result of `scan_txout(tx)` so by line 150 it's already an owned K. 
 
But bigger picture, since `replenish_inner_index()` is an internal API an owned K maybe isn't strictly needed here, but since we're making the assumption K is always small I think it's OK to do for consistency.

notmandatory: Oops, I see @rustaceanrob took the other approach of changing `replenish_inner_index()` I'm fine with going that way too.

rustaceanrob: Oh shoot I already changed the inner functions to use a `&K` if there was unneeded clones in my rebase. I agree that the consistency is a nice to have, but since the `replenish_inner_index` logic probably won't change I feel like we might as well take a reference?

notmandatory:
I added a commit to limit tests to 2 threads, this is what I've always had to do locally to fix an issue with `bitcoind` based tests making calls to quickly.

crates/chain/src/indexer/keychain_txout.rs
notmandatory: Yup it's fine that you changed `replenish_inner_index` to take a ref.

storopoli:
> I added a commit to limit tests to 2 threads, this is what I've always had to do locally to fix an issue with `bitcoind` based tests making calls to quickly.

Oh the nasty `-- --test-threads=2` strikes back.
Yeah I had tons of issues with that in the past while prepping the Nix CI PR.

notmandatory (approved):
ACK ade5999bf551bc47fded49ec5c2376073d8d4d5c

We should hold off on merging this until after PR for #1103 is merged so we don't disrupt that work. But may also require a re-base for this PR.

crates/chain/src/indexer/keychain_txout.rs
ValuedMammal: I think it's fine to pass in an owned value of `K` since it's cloned in the function body anyway

ValuedMammal (approved):
ACK ade5999bf551bc47fded49ec5c2376073d8d4d5c

notmandatory:
This needs to be rebased after #1514 is merged.

notmandatory:
I pushed a test rebase, will do final rebase after #1514 is merged.

PR #1512, opened 2024-07-11
refactor(wallet)!: Replace `new_or_load()` with `load_with_descriptors()`

notmandatory:
### Description

Replace `Wallet::new_or_load()` with `Wallet::load_with_descriptors()`. Updated `Wallet::load_with_descriptors()` now requires a `ChangeSet` and descriptors, and validates they match, if the given descriptors have private keys it adds signers.

Also rename `Wallet::load_from_changeset()` to `Wallet::load()` and remove no longer needed `NewOrLoadError`.

### Notes to the reviewers

Based on discussion in #1500.

### Changelog notice

Changed

* Replace Wallet::new_or_load() with Wallet::load_with_descriptors().
* Rename Wallet::load_from_changeset() to load() and remove no longer needed NewOrLoadError.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [x] I've added tests for the new feature
* [x] I've added docs for the new feature

PR #1513, opened 2024-07-13
refactor(wallet)!: make internal descriptor optional in constructors

### LINKED ISSUE #1511: Please unbreak single-key wallets (and potentially remove internal/external distinction entirely)
In alpha.9 or something, I could pass `None` for the internal descriptor. Now it is no longer an `Option`, so I passed in the same descriptor, but we get an error that they can't be the same.

We have a wallet that is a single key, simple `tr(<pubkey>)` descriptor where internal and external addresses are identical. This used to work perfectly before.

We are hitting this message: `write!(f, "External and internal descriptors are the same")`

notmandatory:
### Description

WIP! needs more testing to ensure everything works without a change descriptor.

If OK then can add new constructors for single descriptor wallets in a post 1.0 release.

Built on #1512 and should fix #1511.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/hwi/src/lib.rs
+++ b/crates/hwi/src/lib.rs
@@ -20,7 +20,7 @@
 //!
 //! # let mut wallet = Wallet::new(
 //! #     "",
-//! #     "",
+//! #     Some(""),
 //! #     Network::Testnet,
 //! # )?;
 //! #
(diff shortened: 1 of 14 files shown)


storopoli:
Nice! I think it is a good compromise for now.

By the way, this still checks if the descriptors are unique right?
What we're actually doing is making the internal optional, but the checks that we've did in #1390 still apply, right?

notmandatory:
> By the way, this still checks if the descriptors are unique right? What we're actually doing is making the internal optional, but the checks that we've did in #1390 still apply, right?

Yes I kept the check that prevents duplicate descriptors, I just don't add a second descriptor and use the external one for change addresses. ~~So for example if the user asked for addresses or utxos from the internal keychain they'll always get `None`.~~ Correction this would break the API since we don't currently return an `Option` when requesting addresses 🙁. This could be a bigger problem. 

To keep the current API the same (not add `Option` for returned `AddressInfo`) we'd need to either 1. allow it to panic if requesting address for non-existent internal keychain, 2. if internal doesn't exist pull address from external keychain instead. In ether case would have to explain careful what could happen.

notmandatory:
closing in favor of #1533

PR #1523, opened 2024-07-25
bdk_wallet: Don't reimplement descriptor checksum, use the one from rust-miniscript

darosior:
Skimming through the BDK code today i noticed the descriptor checksum algorithm is re-implemented instead of using the implementation from rust-miniscript. Re-implementing it here is more code to review and maintain, more room for mistakes and overall less eyes over the code than centralizing a single implementation over at `rust-miniscript`.

While doing this i noticed that one of the variants was completely unnecessary (`calc_checksum_bytes`), so i removed it. I realise it's breaking the API, so if we want to avoid that i can remove this part. I also added a commit to remove redundant calls to `calc_checsum`.

Overall this is just a quick PoC as i noticed it i figured i'd send a PR, my bad if i missed anything!

### DIFF
--- a/crates/wallet/src/descriptor/checksum.rs
+++ b/crates/wallet/src/descriptor/checksum.rs
@@ -43,7 +43,7 @@ fn poly_mod(mut c: u64, val: u64) -> u64 {
 }
 
 /// Compute the checksum bytes of a descriptor, excludes any existing checksum in the descriptor string from the calculation
-pub fn calc_checksum_bytes(mut desc: &str) -> Result<[u8; 8], DescriptorError> {
+fn calc_checksum_bytes(mut desc: &str) -> Result<[u8; 8], DescriptorError> {
     let mut c = 1;
     let mut cls = 0;
     let mut clscount = 0;
--- a/crates/wallet/src/descriptor/checksum.rs
+++ b/crates/wallet/src/descriptor/checksum.rs
@@ -17,82 +17,20 @@
 use crate::descriptor::DescriptorError;
 use alloc::string::String;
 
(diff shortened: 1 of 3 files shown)


darosior:
CI failure looks unrelated to this PR:
> error: package `tokio v1.39.1` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.63.0

EDIT: fixed in https://github.com/bitcoindevkit/bdk/pull/1524.

notmandatory:
Thanks for this! we'll get #1524 merged soon so you can rebase.

darosior:
Rebased on master after the merge of #1524 to fix CI. Looks like [CI is failing on master](https://github.com/bitcoindevkit/bdk/actions/runs/10123843157/job/27997540367) though. Maybe a flaky unit test?

storopoli (approved):
utACK 70c7d6b1c064103b359ad2e774a2648fbd85a414

crates/wallet/src/descriptor/checksum.rs
```diff
@@ -17,82 +17,20 @@
...
-
-    let mut checksum = [0_u8; 8];
-    for j in 0..8 {
-        checksum[j] = CHECKSUM_CHARSET[((c >> (5 * (7 - j))) & 31) as usize];
-    }
-
-    // if input data already had a checksum, check calculated checksum against original checksum
-    if let Some(original_checksum) = original_checksum {
-        if original_checksum.as_bytes() != checksum {
+        let og_checksum = split.1;
```
evanlinjin, line 25: Who is the original gansta we are referring to here? :laughing:

evanlinjin (approved):
ACK 70c7d6b1c064103b359ad2e774a2648fbd85a414

crates/wallet/src/descriptor/checksum.rs
darosior: Haha, i meant "original" of course. :)

Issue #1499, opened 2024-07-02
Introduce TLV for bdk_file_store


notmandatory:
Sub-task for #1103

evanlinjin:
Okay technically we don't need the `T` since we always append the same type. I propose the following:

## Encoding

Encode each element as:
```
[Version: VarInt] [Length: VarInt] [Data: Bytes]
```
Where `VarInt` is defined here: https://en.bitcoin.it/wiki/Protocol_documentation#Variable_length_integer.

## `VersionedElement` trait

```rust
pub trait VersionedElement: Sized {
    /// Decode the `data` of given `version`.
    fn decode_with_version(version: u64, data: &[u8]) -> Result<Self, VersionedDecodeError>;

    /// The version to encode the data with.
    fn encode_version() -> u64;
}

pub enum VersionedDecodeError {
    UnknownVersion(u64),
    Bincode(bincode::Error),
}
```

LLFourn:
I did a lot of thinking about this. 

## Backwards compat

```rust
/// Persists an append-only list of changesets (`C`) to a single file.
#[derive(Debug)]
pub struct Store<C, V = C> {
    magic_len: usize,
    db_file: File,
    marker: PhantomData<(V, C)>,
}

impl<V, C> Store<V, C>
where
    C: Into<V> + Append,
    V: Into<C> + serde::Serialize + serde::de::DeserializeOwned,
```

As you can see there is a new type parameter `V` -- this is the type that is serialized to and from the actual disk.

So in practice you'd have:

```rust
pub struct MyChangeSetV0_17_0 {... }
pub struct MyChangeSet { ... }

// serde serde etc
pub enum VersionedMyChangeSet {
    V0_17_0(MyChangeSetV0_17_0)
    Current(MyChangeSet)
}

impl From<MyChangeSet> for VersionedMyChangeSet {
     fn from(value: VersionedMyChangeSet) -> MyChangeSet {
         // do the obvious
     }
}
```

Then you'd set `V` as `VersionedMyChangeSet` and `C` as `MyChangeSet`. 

If we do have forwards compatibility we need a length prefix. This can be done by just encoding each entry into a `Vec<u8>` and bincode encoding the result to the file.

This makes backwards compat optional and straight forward to achieve. You just have to nudge users to create this enum. For wallet we can create `VersionedChangeSet` and do all this without depending on `bdk_file_persist` I'm sure it's likely to be useful in the future anyway.

[...]

PR #1495, opened 2024-07-01
example: Update `example_cli` and retire old nursery crates

ValuedMammal:
- Adds two commands `init` and `generate`. Loading database doesn't require descriptors
- Replaces `send` command with `psbt` (new, sign, and extract). Supports tap key spend
- Uses `bdk-coin-select` and miniscript `plan` module

fixes #1469 
partially addresses #1462

### Notes to the reviewers

Note the `example_cli` lib now defines the `ChangeSet` and `Anchor` type whereas before these were generic

### Changelog notice

* Improvements to `example_cli` that include generating descriptors and creating PSBTs

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #1527, opened 2024-07-28
Introduce backwards/forwards compatibility for `bdk_file_store`.

nymius:
### Description

Addresses issue #1499.

### Notes to the reviewers

~The changes have been rebased over https://github.com/bitcoindevkit/bdk/pull/1495, because it will be merged sooner and there is a great number of conflicts that will arise later with master if not done.~

~I created this draft to gather feedback.
I'm not understanding properly the issue and cannot visualize how the end user will use this feature.~

I noticed that the `expected variant index` error will be risen only when the decoded version is newer than the one implemented in the code. There is no need to check the indices in that case, as the error is already signaling the encoded variant is higher than the existing in the enum.
If the decoded variant is in the range of the enum variants, but still decoding of the inner struct fails, it means the error is somewhere else.
Also, the decoded variant shouldn't be below the range, as that's not possible, at least using the default variant representation.

Another point to have in mind is I was expecting the above situation raised `InvalidTagEncoding`, but it didn't happen. I'm tracking the trigger of this error, which seems more fitted to describe the situation we want to handle, but couldn't find anything yet.

All this means I could refactor the decoding `entry_iter` code and make it shorter.

### FAQ:

<details>
<summary>What's the idea behind versioned changesets?</summary>

To enable backward and forward compatibility @LLFourn crafted _versioned changeset_ with the following properties:
1. Each new version contains the variants of the previous version and at least one extra variant. I.e.: $\forall\ i,\ j\ \in\ \{1, 2, \cdots, n\}\ V_i > V_j$ if $i > j$ and $V_i \supset V_j$

E.g.:
```rust
enum Versioned1 {
	V1(StructA),
	V2(StructB),
	V3(StructC)
}

// Next code version
enum Versioned2 {
	V1(StructA),
	V2(StructB),
	V3(StructC),
	V4(StructD),
	V5(StructE)
}
```

2. Each variant contains the fields of the previous variant in the same order and at least one extra field. I.e.: $\forall\ r \in\ \{1, 2, \cdots, k\}\ |\ c_r,\ c_{r+1}\ \in\ V_i,\ c_{r+1}\ \supset\ c_r$
E.g.:
```rust
enum Versioned1 {
	V1(StructA),
	V2(StructB),
	V3(StructB)
}

// Next code version
enum Versioned2 {
	V1(StructA),
	V2(StructB),
	V3(StructC),
	V4(StructD),
	V5(StructE)
}

struct StructA {
	field_a: u64
}

struct StructB {
	field_a: u64,
	field_b: u64
}

struct StructC {
	field_a: u64,
	field_b: u64,
	field_c: u64
}

struct StructD {
	field_a: u64,
	field_b: u64,
	field_c: u64,
	field_d: u64,
	field_e: u64,
	field_f: u64
}

struct StructE {
	field_a: u64,
	field_b: u64,
	field_c: u64,
	field_d: u64,
	field_e: u64,
	field_f: u64,
	field_g: u64
}
```
</details>

<details>
<summary>How versioned changesets solve backward compatibility?</summary>
Considering the previous definition of versioned changesets, backward compatibility means:

> Decode $V_j$ using $V_i$ when $i\ >\ j$

This is trivial, as $V_i$ already includes all the variants in $V_j$ so it is a matter of just letting bincode select the variant of $V_i$ based on the encoded variant index.
</details>

<details>
<summary>How versioned changesets solve forward compatibility?</summary>
Defining forward compatibility in terms of the versioned changeset definition:

> Decode changeset $c_w$ from $V_i$ using code only knowing about $V_j$ when $i\ >\ j$ and $c_w$ is not included in $V_j$ . 

Because of property 1 and 2 we know the max variant in $V_j$ , let's call it $c_r$ , fulfills property 2: $c_r \subset c_w$ . So we can try to decode $c_r$ from the encoded data ignoring the remaining fields from $c_w$.
</details>

<details>
<summary>How does the encoding format look like?</summary>
Using the following format:

```text
[L][V]
```

where `[L]` is the length of the encoded data, and `[V]` is the encoding of a variant of the enum representing the versioned changeset, so it is encoded always prefixed with the variant prefix, let's call it `i` followed by the actual data, let's call it `C`. Being the final encoding as the following:

```text
[L][(i)(C)]
```
</details>

<details>
<summary>Why we need to prepend a length to every encoded versioned changeset?</summary>
Typically the encoding length of a versioned changeset $V_i$ would be determined by its type definition, however, a newer version $V_j$ with $j > i$ will have new variants, and the encoding length of those variants cannot be determined if the code only knows about $V_i$.
</details>

<details>
<summary>How we realize we are decoding a higher encoded version than we actually have?</summary>
As bincode encodes each enum variant prefixed with its variant index, if the versioned changeset $V_i$ we are currently working with doesn't contain the variant index we can be sure the data is of a newer type. On the flip side, if the data is of a variant index that exists in $V_i$, we can be sure the data is corrupted.
</details>

<details>
<summary>How encoding works?</summary>

1. Transform the changeset into a versioned-changeset first.
2. Get serialized size of versioned changeset: `[L]`.
3. Encode `[L]` as `u64` and write to file.
4. Encode `[V]` and write to file. Remember bincode encodes it prefixed with the variant index: `[(i)(C)]`
</details>

<details>
<summary>How decoding works?</summary>

1. Decode `[L]`as `u64` from file.
2. Attempt to decode first `[L]` bytes as `[V]`.
	a. On success: Extract `C` from `[V]` (`[(i)(C)]`) and return it.
	b. On failure: Decode `[V]` first bytes as `u64` to extract the enum variant index. If it is too high, trim the variant index from the data, and attempt to decode `C` , the written changeset $c_r$ into the version in use $V_i$ max variant $c_r$, the reading changeset.
</details>

<details>
<summary>How do we determine if a variant is too high?</summary>

Code available to extract the number of variants in an enum is unsafe or it is in the nightly API, so we force versioned changeset to implement the `Default` trait expecting it to always return the max variant (latest changeset version) of the versioned changeset. Then we can encode it using bincode and decode the prefixed variant index as `u64` ignoring the remaining encoded bytes.

```rust
let serialized_max_variant = bincode_options().serialize(&T1::default())?;
// allow trailing bytes because we are serializing the variant, not an u64, and
// we want to extract the varint index prefixed
let max_variant_index = bincode_options()
    .allow_trailing_bytes()
    .deserialize::<u64>(&serialized_max_variant)?;
```
</details>

<details>
<summary>Is it necessary to check if the encoded variant index is higher than the max variant index of the versioned changeset in use?
</summary>

I think it's not. While testing the implementation the code panicked with a particular `bincode::ErrorKind::Custom` error announcing the decoded variant index wasn't expected as it exceeded the limits of the variant indexes available. So it seems just handling this error is enough to detect higher encoded versions.

I **haven't removed the check** because I want someone else to confirm this.

</details>

<details>
<summary>How to decode the higher version encoded changeset $c_w$ with the lower version decoding changeset $c_r$?</summary>

 As I said, the code decodes `[V]` first bytes as `u64` to extract the enum variant index and if it is too high, trims the variant index from the data, and attempts to decode `C` , the written changeset $c_r$ into the version in use $V_i$ max variant $c_r$, the reading changeset.
 As this step is decoding less data than the available in the buffer, we risk to left this data in the buffer losing "data alignment" and ending with a bincode `SizeLimit` error produced after falling short while decoding the last portion of the encoded data.
 To avoid this, we need to keep track of the encoded versioned changeset size `[L]` and after decoding $c_r$, substract the serialized size of $c_r$ (which includes the variant index) to the decoded `[L]`, and use this as an offset to move the reader position after that amount of bytes.

To summarize:
1. Keep track of the encoded size of the changeset
2. Get the size of the decoded $c_r$
3. Get the remaining data to read from the subtraction of 1 minus 2
4. Skip the obtained number in 3 of bytes (the extra fields from $c_w$)
5. Keep decoding the remaining data, coming back to 1 if needed

To illustrate this look at the following example:
$c_w$ (greater encoded changeset version)
$c_r$ (decoding changeset version with less fields)
$c_w$ = `[0][0][0][0]`
$c_r$ = `[x][x]`
`[S]` means the byte has been skipped

Ignore for the sake of the example the length prefix, and the variant index, and give me the license to use a delimiter that bincode doesn't have, the pipe `|`.

initial encoded buffer = `[0][0][0]|[0] | [0][0][0][0] | [0][0][0]|[0] | [0][0][0][0]`
first decode = `[X][X][S]|[S] | [0][0][0][0] | [0][0][0]|[0] | [0][0][0][0]`
second decode = `[X][X][S]|[S] | [X][X][S][S] | [0][0][0]|[0] | [0][0][0][0]`
third decode = `[X][X][S]|[S] | [X][X][S][S] | [X][X][S]|[S] | [0][0][0][0]`
fourth decode = `[X][X][S]|[S] | [X][X][S][S] | [X][X][S]|[S] | [X][X][S][S]`

</details>

<details>
<summary>Which problems this realignment of data may have?</summary>

If not used carefully a version $V_i$ and a higher $V_j$ may have the same variant indexes with different types, and the change above will decode it without returning any error. This is obviously something against the definition of versioned changeset, but without the realignment code this error was easily spotted.
</details>

<details>
<summary>All users should implement versioned changesets to work with the `file_store` crate?</summary>

No, the current code is completely compatible with a simple changeset without using any enum versioned changeset or variant index prefix.
</details>

<details>
<summary>Why there is an pattern matching arm handling `bincode::ErrorKind::InvalidTagEncoding`?</summary>

While implementing the changes and investigating the bincode I tumbled over this error that at first seemed the most suitable to handle in relation to a variant index off limits, but I never got it to reproduce.

**I think it is removable**, but I would like another eye there before doing it.
</details>

<details>
<summary>What tests have been added?</summary>

As the backward compatibility case is trivial, it just depends of bincode logic, the only added tests involves the forward compatibility hard case: when changeset $c_w$ has been encoded with a versioned changeset $V_j$ and its being decoded with a lesser versioned changeset $V_i$ with $i < j$ , and the maximum available variant is $c_r$.
For this test multiple changesets have been appended to replicate the alignment explained above (which may not rise if the changeset encoded was just one).
Also, `Option` type has been avoided because there was a misleading notion of how bincode encoded and decoded these fields that might introduce users to think any additional variants added to higher versions should be of this type.
</details>

<details>
<summary>What I would like to have feedback about?</summary>

To figure out parts of the code where I would like to have feedback on, check out:

- The only implemented test
- Question `Is it necessary to check if the encoded variant index is higher than the max variant index of the versioned changeset in use?` of this **FAQ**.
- Question  `Why there is an arm handling bincode::ErrorKind::InvalidTagEncoding?` of this **FAQ**.
- Question `Which problems this realignment of data may have?` of this **FAQ**.
</details>

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/file_store/src/store.rs
+++ b/crates/file_store/src/store.rs
@@ -11,22 +11,16 @@ use std::{
 
 /// Persists an append-only list of changesets (`C`) to a single file.
 #[derive(Debug)]
-pub struct Store<C>
-where
-    C: Sync + Send,
-{
+pub struct Store<C, V = C> {
     magic_len: usize,
     db_file: File,
-    marker: PhantomData<C>,
+    marker: PhantomData<(C, V)>,
 }
 
-impl<C> Store<C>
+impl<C, V> Store<C, V>
 where
-    C: Merge
-        + serde::Serialize
-        + serde::de::DeserializeOwned
-        + core::marker::Send
-        + core::marker::Sync,
+    C: Into<V> + Merge + Clone + Default + serde::de::DeserializeOwned,
+    V: Into<C> + Default + serde::Serialize + serde::de::DeserializeOwned,
 {
     /// Create a new [`Store`] file in write-only mode; error if the file exists.
     ///
@@ -109,6 +103,19 @@ where
         }
     }
 
+    /// Iterates over the stored **versioned** changesets from first to last, changing the seek
+    /// position at each iteration.
+    ///
+    /// The iterator may fail to read an entry and therefore return an error. However, the first time
+    /// it returns an error will be the last. After doing so, the iterator will always yield `None`.
+    ///
+    /// **WARNING**: This method changes the write position in the underlying file. You should
+    /// always iterate over all entries until `None` is returned if you want your next write to go
+    /// at the end; otherwise, you will write over existing entries.
+    pub fn iter_versioned_changesets(&mut self) -> EntryIter<V, C> {
+        EntryIter::new(self.magic_len as u64, &mut self.db_file)
+    }
+
     /// Iterates over the stored changeset from first to last, changing the seek position at each
     /// iteration.
     ///
@@ -118,8 +125,8 @@ where
     /// **WARNING**: This method changes the write position in the underlying file. You should
     /// always iterate over all entries until `None` is returned if you want your next write to go
     /// at the end; otherwise, you will write over existing entries.
-    pub fn iter_changesets(&mut self) -> EntryIter<C> {
-        EntryIter::new(self.magic_len as u64, &mut self.db_file)
+    pub fn iter_changesets(&mut self) -> impl Iterator<Item = Result<C, IterError>> + '_ {
+        self.iter_versioned_changesets().map(|r| r.map(Into::into))
     }
 
     /// Loads all the changesets that have been stored as one giant changeset.
@@ -135,23 +142,22 @@ where
     /// **WARNING**: This method changes the write position of the underlying file. The next
     /// changeset will be written over the erroring entry (or the end of the file if none existed).
     pub fn aggregate_changesets(&mut self) -> Result<Option<C>, AggregateChangesetsError<C>> {
-        let mut changeset = Option::<C>::None;
-        for next_changeset in self.iter_changesets() {
-            let next_changeset = match next_changeset {
-                Ok(next_changeset) => next_changeset,
-                Err(iter_error) => {
-                    return Err(AggregateChangesetsError {
-                        changeset,
-                        iter_error,
-                    })
+        self.iter_changesets().try_fold(
+            Option::<C>::None,
+            |mut aggregated_changeset: Option<C>, next_changeset| match next_changeset {
+                Ok(next_changeset) => {
+                    match &mut aggregated_changeset {
+                        Some(aggregated_changeset) => aggregated_changeset.merge(next_changeset),
+                        aggregated_changeset => *aggregated_changeset = Some(next_changeset),
+                    }
+                    Ok(aggregated_changeset)
                 }
-            };
-            match &mut changeset {
-                Some(changeset) => changeset.merge(next_changeset),
-                changeset => *changeset = Some(next_changeset),
-            }
-        }
-        Ok(changeset)
+                Err(iter_error) => Err(AggregateChangesetsError {
+                    changeset: aggregated_changeset,
+                    iter_error,
+                }),
+            },
+        )
     }
 
     /// Append a new changeset to the file and truncate the file to the end of the appended
@@ -165,8 +171,11 @@ where
             return Ok(());
         }
 
+        let versioned_changeset: V = changeset.clone().into();
         bincode_options()
-            .serialize_into(&mut self.db_file, changeset)
+            .serialized_size(&versioned_changeset)
+            .and_then(|size| bincode_options().serialize_into::<_, u64>(&mut self.db_file, &size))
+            .and_then(|_| bincode_options().serialize_into(&mut self.db_file, &versioned_changeset))
             .map_err(|e| match *e {
                 bincode::ErrorKind::Io(error) => error,
                 unexpected_err => panic!("unexpected bincode error: {}", unexpected_err),
--- a/crates/file_store/src/store.rs
+++ b/crates/file_store/src/store.rs
@@ -465,7 +465,7 @@ mod test {
     }
 
     #[test]
-    fn deserialize_newer_changeset_version_with_old_code() {
+    fn decode_multiple_vw_enum_variant_index_not_in_vr_with_cr_when_cw_adds_any_kind_of_field() {
         #[derive(PartialEq, Debug, Default, Clone, serde::Deserialize, serde::Serialize)]
         struct TestChangeSetV1 {
             field_a: BTreeSet<String>,
@@ -483,31 +483,22 @@ mod test {
         #[derive(Debug, Default, Clone, serde::Deserialize, serde::Serialize)]
         struct TestChangeSetV2 {
             field_a: BTreeSet<String>,
-            field_b: Option<BTreeSet<String>>,
+            field_b: BTreeSet<String>,
         }
 
         impl Merge for TestChangeSetV2 {
             fn merge(&mut self, other: Self) {
                 self.field_a.extend(other.field_a);
-                if let Some(ref mut field_b) = self.field_b {
-                    if let Some(other_field_b) = other.field_b {
-                        field_b.extend(other_field_b)
-                    }
-                }
+                self.field_b.extend(other.field_b);
             }
 
             fn is_empty(&self) -> bool {
-                if self.field_b.is_none() {
-                    false
-                } else {
-                    self.field_a.is_empty()
-                }
+                self.field_a.is_empty() && self.field_b.is_empty()
             }
         }
 
         let temp_dir = tempfile::tempdir().unwrap();
         let file_path = temp_dir.path().join("20.dat");
-        println!("Test file: {:?}", file_path);
 
         let new_code_aggregation = {
             #[derive(serde::Serialize, serde::Deserialize)]
@@ -549,10 +540,10 @@ mod test {
                 }
             }
 
-            let changesets = (0..2)
+            let changesets = (0..4)
                 .map(|n| TestChangeSetV2 {
                     field_a: BTreeSet::from([format!("{}", n)]),
-                    field_b: None,
+                    field_b: BTreeSet::from([format!("{}", n)]),
                 })
                 .collect::<Vec<_>>();
 
@@ -574,6 +565,7 @@ mod test {
                     acc
                 })
         };
+
         let old_code_aggregation = {
             #[derive(serde::Serialize, serde::Deserialize)]
             enum VersionedTestChangeSet {
@@ -600,21 +592,9 @@ mod test {
                 }
             }
 
-            impl From<VersionedTestChangeSet> for TestChangeSetV2 {
-                fn from(_: VersionedTestChangeSet) -> Self {
-                    Self::default()
-                }
-            }
-
-            impl From<TestChangeSetV2> for VersionedTestChangeSet {
-                fn from(_: TestChangeSetV2) -> Self {
-                    VersionedTestChangeSet::default()
-                }
-            }
-
             // We re-open the file and read all the versioned changesets using the "old"
             // VersionedTestChangeSet
-            let mut db = Store::<TestChangeSetV2, VersionedTestChangeSet>::open(
+            let mut db = Store::<TestChangeSetV1, VersionedTestChangeSet>::open(
                 &TEST_MAGIC_BYTES,
                 &file_path,
             )
@@ -622,7 +602,7 @@ mod test {
             // Merge all versioned changesets in the aggregated correct changeset
             db.iter_changesets()
                 .map(|r| r.expect("must read valid changeset"))
-                .fold(TestChangeSetV2::default(), |mut acc, v| {
+                .fold(TestChangeSetV1::default(), |mut acc, v| {
                     Merge::merge(&mut acc, v);
                     acc
                 })
--- a/crates/file_store/src/store.rs
+++ b/crates/file_store/src/store.rs
@@ -463,4 +463,171 @@ mod test {
             assert_eq!(aggregation, exp_aggregation);
         }
     }
+
+    #[test]
+    fn deserialize_newer_changeset_version_with_old_code() {
+        #[derive(PartialEq, Debug, Default, Clone, serde::Deserialize, serde::Serialize)]
+        struct TestChangeSetV1 {
+            field_a: BTreeSet<String>,
+        }
+        impl Merge for TestChangeSetV1 {
+            fn merge(&mut self, other: Self) {
+                self.field_a.extend(other.field_a);
+            }
+
+            fn is_empty(&self) -> bool {
+                self.field_a.is_empty()
+            }
+        }
+
+        #[derive(Debug, Default, Clone, serde::Deserialize, serde::Serialize)]
+        struct TestChangeSetV2 {
+            field_a: BTreeSet<String>,
+            field_b: Option<BTreeSet<String>>,
+        }
+
+        impl Merge for TestChangeSetV2 {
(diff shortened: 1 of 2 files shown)


crates/file_store/src/store.rs
```diff
@@ -118,7 +126,7 @@ where
     /// **WARNING**: This method changes the write position in the underlying file. You should
     /// always iterate over all entries until `None` is returned if you want your next write to go
     /// at the end; otherwise, you will write over existing entries.
-    pub fn iter_changesets(&mut self) -> EntryIter<C> {
+    pub fn iter_changesets(&mut self) -> EntryIter<V> {
```
evanlinjin: I think it will be helpful to have two versions of this method. One that iterates over versioned changesets; `iter_versioned_changesets(&mut self) -> EntryIter<V>` and one that iterates over changesets; `iter_changesets(&mut self) -> EntryIter<C>`.

nymius: Will do, thanks!

evanlinjin:
There are too many changes in the examples. I think there is a bit of a confusion with the goal of the ticket. We don't need to make our new code understand data persisted with our old code. We want data persisted with the new code to be backwards compatible in the future.

Edit: Okay I see why the examples are changed. It's because we can't impl `Into` and `From` on tuples that don't contain a locally defined type...

At the same time, changing the example changesets to store descriptors is out of scope for this PR.

evanlinjin:
For the `bdk_file_store::Store` methods that require transforming `C` to `V` (or vise versa), I think we should have 2 versions of each. One for when the `C: Into<V>` and `V: Into<C>` bounds are satisfied, one where it is not. For the *not satisfied* case, we take in closures that can handle such transformations.

This way, we can keep the examples mostly the same, and just call the methods with the transform closures as input.

evanlinjin:
> I'm not understanding properly the issue and cannot visualize how the end user will use this feature.

We want to be able to persist changesets in `bdk_file_store` that are compatible with newer versions of the software. The newer version of the software may have new fields or modified fields in changeset.

Instead of storing changesets directly, we wrap the changesets in an enum before storing. This acts as some sort of versioning.

crates/file_store/src/store.rs
```diff
@@ -109,6 +117,19 @@ where
...
+    /// The iterator may fail to read an entry and therefore return an error. However, the first time
+    /// it returns an error will be the last. After doing so, the iterator will always yield `None`.
+    ///
+    /// **WARNING**: This method changes the write position in the underlying file. You should
+    /// always iterate over all entries until `None` is returned if you want your next write to go
+    /// at the end; otherwise, you will write over existing entries.
+    pub fn iter_versioned_changesets(&mut self) -> EntryIter<V> {
+        EntryIter::new(self.magic_len as u64, &mut self.db_file)
+    }
+
```
evanlinjin: `iter_changesets` should be built from `iter_versioned_changesets`. We only want to read versioned changesets directly from the file.

evanlinjin:
```rust
    pub fn iter_changesets(&mut self) -> impl Iterator<Item = Result<C, IterError>> + '_ {
        self.iter_versioned_changesets().map(|r| r.map(Into::into))
    }
```

crates/file_store/src/store.rs
```diff
@@ -11,18 +11,26 @@ use std::{
...
+        + Clone
+        + Default
+        + serde::de::DeserializeOwned
+        + core::marker::Send
+        + core::marker::Sync,
+    V: Into<C>
+        + Default
         + serde::Serialize
         + serde::de::DeserializeOwned
         + core::marker::Send
```
```suggestion
    C: Into<V> + Merge + Clone + Default,
    V: Into<C>
        + Default
        + serde::Serialize
        + serde::de::DeserializeOwned,
```

There are a lot of unnecessary bounds here.

crates/file_store/src/store.rs
```diff
@@ -11,18 +11,26 @@ use std::{
 
 /// Persists an append-only list of changesets (`C`) to a single file.
 #[derive(Debug)]
-pub struct Store<C>
+pub struct Store<C, V = C>
 where
     C: Sync + Send,
```
```suggestion
```

No longer needed.

crates/file_store/src/store.rs
```diff
@@ -135,23 +156,32 @@ where
...
+            .try_fold(Option::<C>::None, changeset_aggregator);
+
+        // But try first versioned V changeset deserializer and if it fails, proceed with C
+        // unversioned deserializer
+        self.iter_versioned_changesets()
+            .try_fold(Option::<C>::None, |acc, x| {
+                changeset_aggregator(acc, x.map(|v| v.into()))
+            })
+            .or(unversioned_parsed_changeset)
     }
```
evanlinjin: I think there is a bit of a misunderstanding of what @LLFourn is suggesting [here](https://github.com/bitcoindevkit/bdk/issues/1499#issuecomment-2205363144) and [here](https://github.com/bitcoindevkit/bdk/issues/1499#issuecomment-2207582765).

Forward compatibility here means that the encoded `V` is of a newer version than what our software has, but our software needs to support deserializing it.

### To have forward compatibility, we need to prepend a length to every encoded `V`.

Why? Typically, the encoding length can be determined by the type definition of `V`. However, a newer version of `V` will have new variants, and the encoding lengths of those variants cannot be determined (because we don't have it).

### When we fail to deserialize into a variable of type `V`, we need to check the "variant index" encoding first.

Why? If the variant index does not exist in our type `V` (which is an enum), then we are sure the data if of a newer type. Since the data is of a newer type, we can apply the forward-compatibility logic. On the flip side, if the data is of a variant index that that exists in `V`, then we are sure that the data is corrupted.

What is "variant index"? `bincode` encodes enums with the variant index first. The variant index is encoded with `VarintEncoding` by default. Refer to how `bincode` encodes integers [here](https://github.com/bincode-org/bincode/blob/trunk/docs/spec.md#intencoding), and enums [here](https://github.com/bincode-org/bincode/blob/trunk/docs/spec.md#enums).

## The Encoding Format

Each changeset should be encoded as follows:

```
[L][V]
```

* `[L]` is a `bincode` encoded `u64`. Use [`serialized_size`](https://docs.rs/bincode/latest/bincode/config/trait.Options.html#method.serialized_size) on `V` to figure out the value of `L`.
* `[V]` is a `bincode` encoded enum type that represents a versioned changeset. Because it is an enum, it is guaranteed that the encoded data begins with a `VariantEncoding`.

### When encoding a changeset

1. Transform the changeset into a versioned-changeset first.
2. Determine the `serialized_size` of the versioned-changeset. The `serialized_size` of `V` is `L`.
3. Encode `L: u64` and write to file.
4. Enncode `V` and write to file.

### When decoding a changeset

1. Decode `L: u64` from file.
2. Read `L` bytes from file and attempt to decode as `V`.
    * On success: Great! Extract `C` from `V` and return it.
    * On failure: Decode `V` as `u64` (to extract the enum-variant-varint-index). If variant index is too high, trim the variant-index from the data, and attempt to decode into `C` (changeset) directly.

### How do we determine if a variant-index is too high?

Ideally, we can use https://doc.rust-lang.org/std/mem/fn.variant_count.html on our definition of `V`. However, this is a nightly-only API.

Instead, let's use the hack that @LLFourn has suggested: Encode `V::default()` (which should always be the latest-version variant), and then decode the result as `u64` to extract the variant-index.

evanlinjin: I also propose to update this PR description to: Introduce backwards/forwards compatibility for `bdk_file_store`.

crates/file_store/src/store.rs
nymius: Why they were here in the first place? Change sets cannot be modified by multiple threads?

crates/file_store/src/store.rs
evanlinjin: Btw, to have forward compatibility, we are assuming that the new changeset `C` only adds new fields at the end ~~as `Option`s (when diffed with the old changeset `C`).~~

nymius: Thanks for the thorough review and write up!

nymius: When decoding a changeset, the `On failure` case for step 2, we hope to succeed if C is of a lower version than the encoded one but without the extra probably `Option` fields or code should fail? Or we expect C to be of the newer version but V doesn't account for it yet? (like in the test I added)

cc @evanlinjin

crates/file_store/src/store.rs
evanlinjin: In the past, we used a trait object that is kept internally in `Wallet` for the `PersistenceBackend`... However, I'm not sure, maybe it was just a mistake. Those bounds aren't needed for the methods we implement below.

crates/file_store/src/store.rs
evanlinjin: @nymius

Just to be clear with what we are talking about, `V` and `C` can either be type definitions defined by the software version we use to read the data, or by the software version which encoded and wrote the data to the file we are reading.

* For type definitions defined by the software that reads, let's use **V<sub>r</sub>** and **C<sub>r</sub>** (Where **C<sub>r</sub>** is the body of **V<sub>r</sub>**'s latest enum variant).
* For type definitions defined by the software the writes, let's use **V<sub>w</sub>** and **C<sub>w</sub>** (where **C<sub>w</sub>** is the body of **V<sub>w</sub>**'s latest enum variant).

When decoding the data as **V<sub>r</sub>** fails, it either means:

1. **V<sub>w</sub>**'s enum variant index does not exist in **V<sub>r</sub>**.
2. The data to decode is invalid. This is either data corruption, or **V<sub>r</sub>** has a variant with a different `C` definition to that of **V<sub>w</sub>**.

For `1.`, we still try decode the data (excluding the enum variant index) to **C<sub>r</sub>**. This will work if **C<sub>w</sub>** only adds new fields on **C<sub>r</sub>** (which still exists in **V<sub>w</sub>** as a lower enum variant).

For `2.`, we can just error out. It's either a programming error or data corruption.

LLFourn:
Given discussion on https://github.com/bitcoindevkit/bdk/issues/1499 I thank @nymius for his work exploring this and vote to close this in favor of status quo + updated documentation explaining how this is a development database.

nymius:
Thanks @LLFourn for the follow up and thanks @evanlinjin for all the thorough feedback! Should I close this now @notmandatory?

nymius:
Closed in favor of keeping actual store design and update documentation explaining how this is oriented to be used as a development database. Read above and #1499 for further information.

PR #1533, opened 2024-08-01
[wallet] Enable support for single descriptor wallets

### LINKED ISSUE #1511: Please unbreak single-key wallets (and potentially remove internal/external distinction entirely)

ValuedMammal:
The change descriptor is made optional, making this an effective reversion of #1390 and enables creating wallets with a single descriptor.

fixes #1511 

### Notes to the reviewers

PR 1390 also removed an error case [`ChangePolicyDescriptor`](https://github.com/bitcoindevkit/bdk/blob/8eef350bd08057acc39b6fc50b1217db5e29b968/crates/wallet/src/wallet/mod.rs#L1529-L1533) and this can possibly be added back. In the case the wallet only has a single descriptor we allow any utxos to fund a tx that aren't specifically marked unspendable regardless of the change spend policy.

### Changelog notice

* Added method `Wallet::create_single` that expects a single `D: IntoWalletDescriptor` as input and enables building a wallet with no internal keychain.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/wallet/examples/compiler.rs
+++ b/crates/wallet/examples/compiler.rs
@@ -56,28 +56,8 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     println!("Compiled into Descriptor: \n{}", descriptor);
 
-    // Do the same for another (internal) keychain
-    let policy_str = "or(
-        10@thresh(2,
-            pk(029ffbe722b147f3035c87cb1c60b9a5947dd49c774cc31e94773478711a929ac0),pk(025f05815e3a1a8a83bfbb03ce016c9a2ee31066b98f567f6227df1d76ec4bd143),pk(025625f41e4a065efc06d5019cbbd56fe8c07595af1231e7cbc03fafb87ebb71ec)
-        ),1@and(
-            pk(03deae92101c790b12653231439f27b8897264125ecb2f46f48278603102573165),
-            older(12960)
-        )
-    )"
-    .replace(&[' ', '\n', '\t'][..], "");
-
-    println!("Compiling internal policy: \n{}", policy_str);
-
-    let policy = Concrete::<String>::from_str(&policy_str)?;
-    let internal_descriptor = Descriptor::new_wsh(policy.compile()?)?.to_string();
-    println!(
-        "Compiled into internal Descriptor: \n{}",
-        internal_descriptor
-    );
-
     // Create a new wallet from descriptors
-    let mut wallet = Wallet::create(descriptor, internal_descriptor)
+    let mut wallet = Wallet::create_single(descriptor)
         .network(Network::Regtest)
         .create_wallet_no_persist()?;
 
--- a/crates/wallet/src/wallet/changeset.rs
+++ b/crates/wallet/src/wallet/changeset.rs
@@ -103,16 +103,18 @@ impl ChangeSet {
         let row = wallet_statement
             .query_row([], |row| {
                 Ok((
-                    row.get::<_, Impl<Descriptor<DescriptorPublicKey>>>("descriptor")?,
-                    row.get::<_, Impl<Descriptor<DescriptorPublicKey>>>("change_descriptor")?,
-                    row.get::<_, Impl<bitcoin::Network>>("network")?,
+                    row.get::<_, Option<Impl<Descriptor<DescriptorPublicKey>>>>("descriptor")?,
+                    row.get::<_, Option<Impl<Descriptor<DescriptorPublicKey>>>>(
+                        "change_descriptor",
+                    )?,
+                    row.get::<_, Option<Impl<bitcoin::Network>>>("network")?,
                 ))
             })
             .optional()?;
-        if let Some((Impl(desc), Impl(change_desc), Impl(network))) = row {
-            changeset.descriptor = Some(desc);
-            changeset.change_descriptor = Some(change_desc);
-            changeset.network = Some(network);
+        if let Some((desc, change_desc, network)) = row {
+            changeset.descriptor = desc.map(Impl::into_inner);
+            changeset.change_descriptor = change_desc.map(Impl::into_inner);
+            changeset.network = network.map(Impl::into_inner);
         }
 
         changeset.local_chain = local_chain::ChangeSet::from_sqlite(db_tx)?;
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -246,9 +246,9 @@ pub enum LoadMismatch {
         /// Keychain identifying the descriptor.
         keychain: KeychainKind,
         /// The loaded descriptor.
-        loaded: ExtendedDescriptor,
+        loaded: Option<ExtendedDescriptor>,
         /// The expected descriptor.
-        expected: ExtendedDescriptor,
+        expected: Option<ExtendedDescriptor>,
     },
 }
 
@@ -401,11 +401,15 @@ impl Wallet {
         ));
 
         let (change_descriptor, change_signers) = match params.change_descriptor {
-            Some(desc_fn) => {
-                let (descriptor, mut keymap) = desc_fn(&secp, network)?;
-                keymap.extend(params.change_descriptor_keymap);
-                let change_signers = Arc::new(SignersContainer::build(keymap, &descriptor, &secp));
-                (Some(descriptor), change_signers)
+            Some(make_desc) => {
+                let (change_descriptor, mut internal_keymap) = make_desc(&secp, network)?;
+                internal_keymap.extend(params.change_descriptor_keymap);
+                let change_signers = Arc::new(SignersContainer::build(
+                    internal_keymap,
+                    &change_descriptor,
+                    &secp,
+                ));
+                (Some(change_descriptor), change_signers)
             }
             None => (None, Arc::new(SignersContainer::new())),
         };
@@ -442,7 +446,7 @@ impl Wallet {
     /// Note that the descriptor secret keys are not persisted to the db. You can either add
     /// signers after-the-fact with [`Wallet::add_signer`] or [`Wallet::set_keymap`]. Or you can
     /// add keys when building the wallet using [`LoadParams::keymap`] and/or
-    /// [`LoadParams::descriptors`].
+    /// [`LoadParams::descriptor`].
     ///
     /// # Synopsis
     ///
@@ -467,7 +471,9 @@ impl Wallet {
     /// let mut conn = bdk_wallet::rusqlite::Connection::open(file_path)?;
     /// let mut wallet = Wallet::load()
     ///     // check loaded descriptors matches these values and extract private keys
-    ///     .descriptors(EXTERNAL_DESC, INTERNAL_DESC)
+    ///     .descriptor(KeychainKind::External, Some(EXTERNAL_DESC))
+    ///     .descriptor(KeychainKind::Internal, Some(INTERNAL_DESC))
+    ///     .extract_keys()
     ///     // you can also manually add private keys
     ///     .keymap(KeychainKind::External, external_keymap)
     ///     .keymap(KeychainKind::Internal, internal_keymap)
@@ -499,42 +505,6 @@ impl Wallet {
         let chain = LocalChain::from_changeset(changeset.local_chain)
             .map_err(|_| LoadError::MissingGenesis)?;
 
-        let mut descriptor_keymap = params.descriptor_keymap;
-        let descriptor = changeset
-            .descriptor
-            .ok_or(LoadError::MissingDescriptor(KeychainKind::External))?;
-        check_wallet_descriptor(&descriptor).map_err(LoadError::Descriptor)?;
-
-        let (change_descriptor, change_signers) = match changeset.change_descriptor {
-            Some(change_descriptor) => {
-                check_wallet_descriptor(&change_descriptor).map_err(LoadError::Descriptor)?;
-                let mut change_descriptor_keymap = params.change_descriptor_keymap;
-
-                // check params match loaded
-                if let Some(exp_change_descriptor) = params.check_change_descriptor {
-                    let (exp_change_descriptor, keymap) =
-                        (exp_change_descriptor)(&secp, network).map_err(LoadError::Descriptor)?;
-                    change_descriptor_keymap.extend(keymap);
-
-                    if change_descriptor.descriptor_id() != exp_change_descriptor.descriptor_id() {
-                        return Err(LoadError::Mismatch(LoadMismatch::Descriptor {
-                            keychain: KeychainKind::Internal,
-                            loaded: change_descriptor,
-                            expected: exp_change_descriptor,
-                        }));
-                    }
-                }
-                let change_signers = Arc::new(SignersContainer::build(
-                    change_descriptor_keymap,
-                    &change_descriptor,
-                    &secp,
-                ));
-                (Some(change_descriptor), change_signers)
-            }
-            None => (None, Arc::new(SignersContainer::new())),
-        };
-
-        // checks
         if let Some(exp_network) = params.check_network {
             if network != exp_network {
                 return Err(LoadError::Mismatch(LoadMismatch::Network {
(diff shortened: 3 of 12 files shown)


crates/wallet/src/wallet/params.rs
```diff
@@ -32,22 +32,38 @@ where
...
+        Self {
+            descriptor: make_descriptor_to_extract(descriptor),
+            descriptor_keymap: KeyMap::default(),
+            change_descriptor: None,
+            change_descriptor_keymap: KeyMap::default(),
+            network: Network::Bitcoin,
+            genesis_hash: None,
+            lookahead: DEFAULT_LOOKAHEAD,
+        }
+    }
```
evanlinjin: Instead of having a new constructor, I would prefer changing `new` to only take in a single `descriptor` and creating:
```rust
pub fn change_descriptor(mut self, change_descriptor: D) -> Self { todo!() }
```

Ergonomically, maybe we should have two methods for `keymap`, one for each keychain?

ValuedMammal: That sounds reasonable.

cc @notmandatory

storopoli: If we change `CreateParams::new` to take a single descriptor (right now takes 2) it will  be a breaking change.
I think that it is acceptable and the time to do it is now!

thunderbiscuit: If you are considering a breaking change, then IMO the cleanest way to do it is similar to the [Wallet::load()](https://docs.rs/bdk_wallet/1.0.0-beta.1/bdk_wallet/struct.Wallet.html#method.load) API, with the `Wallet::new()` not taking any descriptors at all, and a single method `descriptors()` that takes an optional change_descriptor.

```rust
pub fn descriptors(mut self, descriptor: D, change_descriptor: Option<D>) -> Self { todo!() }
```

stevenroose:
Much love for this one :wink:

crates/wallet/src/wallet/params.rs
evanlinjin: @thunderbiscuit but that means we need an error case when `descriptor` is not provided.

I think it makes sense for the builder struct to take in all that is necessary in `new`, and all that is optional with additional methods. Hence, my original suggestion.

@storopoli I think a bit of breaking API changes is okay in beta. The worst case scenario is users need to change a few lines of code.

ValuedMammal:
What should be the function signature and behavior of [`Wallet::public_descriptor`](https://github.com/bitcoindevkit/bdk/blob/785371e0a113f1be40a9350f2393846a28971f95/crates/wallet/src/wallet/mod.rs#L1735) ?

crates/wallet/src/wallet/params.rs
thunderbiscuit: Sounds good thanks for expanding on the constructor design @evanlinjin I had not seen that. Cheers! I have more questions which I'll ask in the thread below.

thunderbiscuit:
Thanks for the input Evan about the new/load taking only and all that's required to build, and further methods that provide setter/validators to be optional. 

While we're on the topic and refactoring of the builder I figured I would just ask here instead of opening a new issue:
### Q1
Most of the fields on the [`LoadParams` type](https://docs.rs/bdk_wallet/1.0.0-beta.1/src/bdk_wallet/wallet/params.rs.html#116-124) are "checkers/validators" (for example `check_network`, `check_genesis_hash`, `check_descriptor`), but the methods on the type are simply `network`, `descriptor`, and `genesis_hash`. This makes it unintuitive when reading the docs and the examples, because you're left wondering whether the network is persisted at all or if you need to set it correctly while using the load builder, or whether you might in fact be able to change it between loads. My suggestion would be to use this PR to fix the name on the validator methods to be either in the style of `validate_network` or `check_network` for all 4 methods. Let me know everyone if I've missed something or if there is a better way or reason to keep them unclear.
### Q2
I see that the `lookahead` is a setter, but even after reading the docs I'm not sure what it's setting exactly because it just sounds like a sort of stop gap but not a stop gap. See the [docs here on lookahead](https://docs.rs/bdk_chain/latest/bdk_chain/indexer/keychain_txout/struct.KeychainTxOutIndex.html#method.new) and the more explicit [lookahead in KeychainTxOutIndex](https://docs.rs/bdk_chain/latest/bdk_chain/indexer/keychain_txout/struct.KeychainTxOutIndex.html#lookahead-script-pubkeys). 
> When an user first recovers a wallet (i.e. from a recovery phrase and/or descriptor), we will NOT have knowledge of which script pubkeys are revealed. So when we index a transaction or txout (using [index_tx](https://docs.rs/bdk_chain/latest/bdk_chain/indexer/keychain_txout/struct.KeychainTxOutIndex.html#method.index_tx)/[index_txout](https://docs.rs/bdk_chain/latest/bdk_chain/indexer/keychain_txout/struct.KeychainTxOutIndex.html#method.index_txout)) we scan the txouts against script pubkeys derived above the last revealed index. These additionally-derived script pubkeys are called the lookahead. 

Don't we scan up to the stop gap? I know this from experience when looking at the Esplora logs on full scan and syncs (for example a stop gap of 3 with a lookahead of 100 will query the Esplora instance for only 4 addresses), so not sure if maybe it refers to something else?

> The [KeychainTxOutIndex](https://docs.rs/bdk_chain/latest/bdk_chain/indexer/keychain_txout/struct.KeychainTxOutIndex.html) is constructed with the lookahead and cannot be altered. 

Wait so is it persisted then? Or by that do we mean it can't be altered per session but altered on every `load` of the wallet?

### Q3
Related to (Q2), I know (think?) the `lookahead` is not persisted, but is there a danger in having two different lookahead between sessions? What if I set a lookahead of 3 when I create a wallet, and 300 the next time I load the wallet? Or the opposite? 

Sorry I know it's a lot but I did not have time to review #1514 and I'm just now after the week in Nashville confident enough to ask my questions and comment on the shape of the API. Cheers!

ValuedMammal:
> Don't we scan up to the stop gap?

The docs describing the "lookahead" pertain specifically to the correct functioning of an `Indexer` when indexing transactions. Some number of lookahead SPKs are necessary for the indexer to recognize when a tx is relevant to its tracked SPKs. ~~The stop-gap used during a full scan depends on there being a lookahead, in other words having a large stop-gap won't help if you haven't stored any lookaheads~~.

edit: see below

> is there a danger in having two different lookahead between sessions?

I think it's a valid question..

crates/wallet/src/wallet/params.rs
notmandatory: > Instead of having a new constructor, I would prefer changing `new` to only take in a single `descriptor` and creating ...

Since our recommended practice is to provide two descriptors I'd rather leave `new(...)` as is and only add `new_single_descriptor(...)` as @ValuedMammal has it for the few projects that need it.

> Ergonomically, maybe we should have two methods for `keymap`, one for each keychain?

I'd rather not have separate methods since it's an API change and doesn't add very much.

notmandatory:
> What should be the function signature and behavior of [`Wallet::public_descriptor`](https://github.com/bitcoindevkit/bdk/blob/785371e0a113f1be40a9350f2393846a28971f95/crates/wallet/src/wallet/mod.rs#L1735) ?

Since single descriptor wallets are the advanced user/unusual case I think any `Wallet` methods that require a `keychain_kind` like `Wallet::public_descriptor` should panic if called with `KeychainKind::Internal`. This behavior should be documented on the `Wallet::new_single_descriptor` so the user knows it is their responsibility to only use `KeychainKind::External`.  If this seems too harsh maybe we should hide the `Wallet::new_single_descriptor` option behind a non-default feature flag.

thunderbiscuit:
> Some number of lookahead SPKs are necessary for the indexer to recognize when a tx is relevant to its tracked SPKs. The stop-gap used during a full scan depends on there being a lookahead, in other words having a large stop-gap won't help if you haven't stored any lookaheads.

That sort of makes it sound like it could be a pretty good footgun then no? Big stop gap, tiny lookahead, your wallet not finding your txs? I have not, however, been able to reproduce this locally, so I'm still unclear about what this lookahead is doing, and because it's part of the public API of the `Wallet` type I want to make sure I really get it. Here's what I have tried on regtest, and looking at the logs for my esplora instance:
- Stop gap of 3, lookahead of 100. The wallet syncs 6 addresses total (3 per keychain). The lookahead does not impact the spks the wallet attempts to sync.
- Stop gap of 25, lookahead of 7. The wallet syncs 50 addresses, (lookahead didn't seem to be a problem, again not impacting the syncing process)
- Send coins to address 19, set the lookahead to 7 and the stop gap to 25. The coins are seen and wallet shows correct balance. Lookahead below the stop gap did not cause any problems.

More notes on this: [the docs here](https://docs.rs/bdk_chain/latest/bdk_chain/indexer/keychain_txout/struct.KeychainTxOutIndex.html#method.new) say:
> Without a lookahead the index will miss outputs you own when processing transactions whose output script pubkeys lie beyond the last revealed index. In certain situations, such as when performing an initial scan of the blockchain during wallet import, it may be uncertain or unknown what the index of the last revealed script pubkey actually is.

Again this makes it sound close to the concept of the stop gap, but I assume it's different somehow or a more generalized version of it that can be used in the deeper logic of the KeychainIndex? That's as far as the API docs go. I'll try to look at where it's actually used in the chain crate tomorrow and report back.

ValuedMammal:
@thunderbiscuit Ah, you're right, the lookahead doesn't appear to have a role during full scan for esplora because we get the SPKs to sync by calling `all_unbounded_spk_iters`

For sync it's also not relevant because we only sync SPKs that are already revealed.

In short, the lookahead defines a number of scripts to derive ahead of time without revealing them. If you do a full scan with the default lookahead of 25 and persist the wallet, then on subsequent loads ~~those SPKs are still stored in the index and there shouldn't be any harm in setting a new lookahead.~~

edit: Nope that's also wrong. we don't persist derived SPKs, only the value of `last_revealed`. So yeah we probably need more testing.

One way to illustrate the effect of the lookahead is to create a wallet with the default value of 25. Then peek at address index 25. The method `Wallet::is_mine` will return false because internally the address is not yet derived (remembering that SPKs are 0-indexed).

For context one of the reasons we chose to make the lookahead configurable is because for a short time we had it set to 1000 and users reported unbearably long load times. Beyond that, I don't think it's something users should generally care about other than to know it exists

evanlinjin:
> Since single descriptor wallets are the advanced user/unusual case I think any `Wallet` methods that require a `keychain_kind` like `Wallet::public_descriptor` should panic if called with `KeychainKind::Internal`. This behavior should be documented on the `Wallet::new_single_descriptor` so the user knows it is their responsibility to only use `KeychainKind::External`. If this seems too harsh maybe we should hide the `Wallet::new_single_descriptor` option behind a non-default feature flag.

I think the better solution is to have `KeychainKind` as a type parameter in `Wallet<K = KeychainKind>`. And have a second type `pub struct KeychainSingle` that is for single-descriptor wallets specifically. This way, we have methods specific to each keychain arrangement, thus no need for panicing.

We can do it so that it is not a breaking change by setting a default type for `K = KeychainKind`. Therefore `Wallet` is equivalent to `Wallet<KeychainKind>`.

Additionally, there are a bunch of other methods that take in `KeychainKind` as an input. It wouldn't make sense to have methods that take in/ return `KeychainKind` for a single-desriptor wallet at all. With this solution, we can have single-descriptor wallet methods not take in a `keychain` parameter.

~~There is also one more thing that I would like to figure out. Is having two fields for `signers` even useful (`::signers` and `::change_signers`)? If not, we can combine the two into one field.~~ Too much work I think.

Edit: The downside of this solution is that it is more work, but in my opinion, it results in a superior API.

ValuedMammal:
Would it be easier to just bring back the `map_keychain` functionality as a way to avoid panics?

crates/wallet/src/wallet/params.rs
ValuedMammal: > I think it makes sense for the builder struct to take in all that is necessary in `new`, and all that is optional with additional methods. Hence, my original suggestion.

I generally agree with this and do think it would be cleaner than having two `new` constructors but I don't have a strong preference in terms of API

evanlinjin:
@ValuedMammal I think I agree with you here. What I am suggesting requires too much work. What you suggest works good enough to make the users happy.

notmandatory:
> Would it be easier to just bring back the `map_keychain` functionality as a way to avoid panics?

After sleeping on it I also support avoiding the panics by just pulling from the `External` keychain when ever the `Internal` one is requested, which is what the TxBuilder is going to do. We just need to document it (and the TxBuilder "change" related functions won't work anymore).

I can also see that @evanlinjin's suggested builder change of making `new()` take the `External` and a different function to set the `Internal` isn't too big a breaking change. This way also sets us up for supporting multipath descriptors (BIP-389).

We should finalize these decisions first thing in tomorrow's dev call so @ValuedMammal can finish up this PR.

ValuedMammal:
I didn't touch the method names for `LoadParams` as noted in https://github.com/bitcoindevkit/bdk/pull/1533#issuecomment-2267641603. I think the docs are clear enough but I'm open to more discussion. Appreciate everyone's input

crates/wallet/src/wallet/params.rs
ValuedMammal: After more discussion I think it's better to be explicit and provide another constructor along the lines of `new_single_descriptor`. This also makes for a convenient place to add docs that are specific to single keychain wallets.

ValuedMammal:
Attempting to load a persisted wallet with no change desc caused an error when reading from sqlite - currently investigating

Pushed a fix in d4fbe6677286956fcec1d1db4890c8c007e097b2

thunderbiscuit:
As requested over the dev call my comment above about the method names on the load builder does not need to be addressed here, and instead can be reviewed in a separate PR: #1537.

evanlinjin:
I'm wondering whether it will be worth it to panic when the caller provides an internal keymap for a single descriptor wallet (and document the panic). Just in case the caller adds a keymap and accidentally uses the wrong keychain variant.

crates/wallet/src/wallet/changeset.rs
```diff
@@ -104,14 +104,16 @@ impl ChangeSet {
             .query_row([], |row| {
                 Ok((
                     row.get::<_, Impl<Descriptor<DescriptorPublicKey>>>("descriptor")?,
```
evanlinjin:
```suggestion
                    row.get::<_, Option<Impl<Descriptor<DescriptorPublicKey>>>>("descriptor")?,
```

I'm thinking it also makes sense to NOT error when we don't find an external descriptor. If we don't have an external descriptor, the problem isn't with our persistence implementation (the problem is with the caller not persisting an external descriptor).

We already catch the error if we have a missing descriptor in `Wallet::load_with_params`. This is the appropriate place to catch this error.

crates/wallet/tests/wallet.rs
```diff
@@ -4116,3 +4116,40 @@ fn test_insert_tx_balance_and_utxos() {
     assert!(wallet.list_unspent().next().is_none());
     assert_eq!(wallet.balance().total().to_sat(), 0);
 }
+
+#[test]
```
evanlinjin, line 4186: Just in case, maybe we should create a test that also recovers a single descriptor wallet from persistence.

ValuedMammal:
@evanlinjin Right, the current load API makes it a little awkward to accommodate the single keychain case. Another example is the `descriptors` method on `LoadParams` that expects two descriptors. (although with this PR we just ignore the params if there's no change descriptor in the loaded changeset). Maybe better to change `LoadParams::descriptors` to `descriptor` that takes in a keychain kind and descriptor?

> I'm wondering whether it will be worth it to panic when the caller provides an internal keymap for a single descriptor wallet (and document the panic). Just in case the caller adds a keymap and accidentally uses the wrong keychain variant.

You're referring to setting the keymap on `LoadParams` or the other method `Wallet::set_keymap`? or both?

crates/wallet/src/wallet/changeset.rs
ValuedMammal: Do the same for `network` as well?

evanlinjin: > Do the same for `network` as well?

I think so, yes!

crates/wallet/tests/wallet.rs
ValuedMammal: Done in 27da178c4a5581fca9ca1e7706ec47629ba65b4c

notmandatory:
> @evanlinjin Right, the current load API makes it a little awkward to accommodate the single keychain case. Another example is the `descriptors` method on `LoadParams` that expects two descriptors. (although with this PR we just ignore the params if there's no change descriptor in the loaded changeset). Maybe better to change `LoadParams::descriptors` to `descriptor` that takes in a keychain kind and descriptor?

I like this idea to replace `LoadParams::descriptors(descriptor,change_descriptor)` with `LoadParams::descriptor(keychain, descriptor)`.

evanlinjin:
~~Another idea is to have separate builders for single/dual descriptors. This may make things easier to reason about?~~

Edit: I changed my mind. It makes sense to be able to load both single/dual descriptor wallets from the same load constructor.

I cast a vote for `LoadParams::descriptor(keychain, descriptor)`.

> You're referring to setting the keymap on LoadParams or the other method Wallet::set_keymap? or both?

I'm referring to when we create a fresh wallet with `CreateParams`. I think `Wallet::set_keymap` would already panic?

ValuedMammal:
Followup question: when loading a single descriptor wallet `changeset.change_descriptor` will be `None`. In this case if the user calls `descriptor` on `LoadParams` passing `KeychainKind::Internal`, should we

- add a change descriptor to the wallet
- error because it doesn't match what was loaded
- or silently ignore the param?

evanlinjin:
> Followup question: when loading a single descriptor wallet `changeset.change_descriptor` will be `None`. In this case if the user calls `descriptor` on `LoadParams` passing `KeychainKind::Internal`, should we
> 
>     * add a change descriptor to the wallet
> 
>     * error because it doesn't match what was loaded
> 
>     * or silently ignore the param?

I think we should error since we are meant to check the loaded changeset's descriptor against the provided descriptor.

Also refer to my comment https://github.com/bitcoindevkit/bdk/pull/1537#discussion_r1712948216 which would (in my opinion) make our API communicate the "checking" part more clearer.

notmandatory:
I also think we should error if LoadParams doesn't exactly match what's in the loaded file, including if the change_descriptor value in file not matching LoadParams as Some or None.

evanlinjin:
> I also think we should error if LoadParams doesn't exactly match what's in the loaded file, including if the change_descriptor value in file not matching LoadParams as Some or None.

I think we need to rethink the `LoadParams` constructor API if we want to do this. What if the caller does not want to check anything? What if the caller only wants to check the external descriptor (because they don't know what the internal descriptor is)?

```rust
pub fn check_descriptor<D>(mut self, keychain: KeychainKind, exp: Option<D>, extract_keys: bool) -> Self {
    todo!()
}
```

I think an API like this makes sense for what you are proposing. `exp` is what we expect the loaded changeset should have.

ValuedMammal:
@evanlinjin I added your suggestion, happy to make further tweaks if anything is missing.

Edit: still needs work

ValuedMammal:
I changed the `extract_keys` bool to a separate method on `LoadParams` that ensures we either extract keys for both keychains or for none of them

crates/wallet/src/wallet/mod.rs
```diff
@@ -393,7 +446,7 @@ impl Wallet {
     /// Note that the descriptor secret keys are not persisted to the db. You can either add
     /// signers after-the-fact with [`Wallet::add_signer`] or [`Wallet::set_keymap`]. Or you can
     /// add keys when building the wallet using [`LoadParams::keymap`] and/or
-    /// [`LoadParams::descriptors`].
+    /// [`LoadParams::descriptor`].
```
notmandatory: ⛏️ Should mention `extract_keys` here:
```suggestion
    /// [`LoadParams::descriptor`] with a private key descriptor and [`LoadParams::extract_keys`].
```

notmandatory (approved):
ACK 3951110bb596d56a30ddc4cf09903cee4eebedea

This turned out bigger than I expected but should be a good solution for the projects that need it.

crates/wallet/src/wallet/mod.rs
ValuedMammal: @notmandatory fixed

evanlinjin (approved):
ACK 13e7008f00b77802481733eb74daf38f48a2faef

crates/wallet/examples/compiler.rs
```diff
@@ -56,28 +56,8 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-    let policy = Concrete::<String>::from_str(&policy_str)?;
-    let internal_descriptor = Descriptor::new_wsh(policy.compile()?)?.to_string();
-    println!(
-        "Compiled into internal Descriptor: \n{}",
-        internal_descriptor
-    );
-
     // Create a new wallet from descriptors
-    let mut wallet = Wallet::create(descriptor, internal_descriptor)
+    let mut wallet = Wallet::create_single(descriptor)
```
cryptoquick, line 60: Nice! This looks great.

toskaen:
is there a reason why bdk_wallet 1.1.0 did not adopt it? or am I missing something.

oh, its coming into 1.2.0 - will watch

PR #1535, opened 2024-08-03
test(electrum): Test sync in reorg and no-reorg situations

### LINKED ISSUE #1125: Calling `example_electrum sync` on an unconfirmed tx may result in the tx being stuck as unconfirmed eternally (according to the cli))

LagginTimes:
Closes #1125.

### Description

Add new test for `bdk_electrum` to make sure previously unconfirmed transactions get confirmed again in both reorg and no-reorg situations.

### Changelog notice

* Added `wait_until_electrum_sees_txid` method to `TestEnv`.
* `wait_until_electrum_sees_block` now has a `Duration` input for timeout.
* Removed exponential polling delay in `wait_until_electrum_sees_block`.
* Added `test_sync` to `bdk_electrum` to make sure previously unconfirmed transactions get confirmed in both reorg and no-org situations.

#### All Submissions:

### DIFF
--- a/crates/electrum/tests/test_electrum.rs
+++ b/crates/electrum/tests/test_electrum.rs
@@ -275,7 +275,9 @@ pub fn test_update_tx_graph_stop_gap() -> anyhow::Result<()> {
     Ok(())
 }
 
-/// Ensure that [`BdkElectrumClient::sync`] can confirm previously unconfirmed transactions in both reorg and no-reorg situations.
+/// Ensure that [`BdkElectrumClient::sync`] can confirm previously unconfirmed transactions in both
+/// reorg and no-reorg situations. After the transaction is confirmed after reorg, check if floating
+/// txouts for previous outputs were inserted for transaction fee calculation.
 #[test]
 fn test_sync() -> anyhow::Result<()> {
     const SEND_AMOUNT: Amount = Amount::from_sat(10_000);
@@ -316,7 +318,7 @@ fn test_sync() -> anyhow::Result<()> {
         &mut recv_graph,
     )?;
 
-    // Check if balance is zero when transaction exists only in mempool.
+    // Check for unconfirmed balance when transaction exists only in mempool.
     assert_eq!(
         get_balance(&recv_chain, &recv_graph)?,
         Balance {
--- a/crates/electrum/tests/test_electrum.rs
+++ b/crates/electrum/tests/test_electrum.rs
@@ -1,14 +1,18 @@
 use bdk_chain::{
     bitcoin::{hashes::Hash, Address, Amount, ScriptBuf, Txid, WScriptHash},
     local_chain::LocalChain,
-    spk_client::{FullScanRequest, SyncRequest},
+    spk_client::{FullScanRequest, SyncRequest, SyncResult},
     spk_txout::SpkTxOutIndex,
-    Balance, ConfirmationBlockTime, IndexedTxGraph,
+    Balance, ConfirmationBlockTime, IndexedTxGraph, Indexer, Merge,
 };
 use bdk_electrum::BdkElectrumClient;
 use bdk_testenv::{anyhow, bitcoincore_rpc::RpcApi, TestEnv};
+use core::time::Duration;
 use std::collections::{BTreeSet, HashSet};
 use std::str::FromStr;
+
+// Batch size for `sync_with_electrum`.
+const BATCH_SIZE: usize = 5;
 
 fn get_balance(
     recv_chain: &LocalChain,
@@ -20,6 +24,39 @@ fn get_balance(
         .graph()
         .balance(recv_chain, chain_tip, outpoints, |_, _| true);
     Ok(balance)
+}
+
+fn sync_with_electrum<I, Spks>(
+    client: &BdkElectrumClient<electrum_client::Client>,
+    spks: Spks,
+    chain: &mut LocalChain,
+    graph: &mut IndexedTxGraph<ConfirmationBlockTime, I>,
+) -> anyhow::Result<SyncResult>
+where
+    I: Indexer,
+    I::ChangeSet: Default + Merge,
+    Spks: IntoIterator<Item = ScriptBuf>,
+    Spks::IntoIter: ExactSizeIterator + Send + 'static,
+{
+    let mut update = client.sync(
+        SyncRequest::from_chain_tip(chain.tip()).chain_spks(spks),
+        BATCH_SIZE,
+        true,
+    )?;
+
+    // Update `last_seen` to be able to calculate balance for unconfirmed transactions.
+    let now = std::time::UNIX_EPOCH
+        .elapsed()
+        .expect("must get time")
+        .as_secs();
+    let _ = update.graph_update.update_last_seen_unconfirmed(now);
+
+    let _ = chain
+        .apply_update(update.chain_update.clone())
+        .map_err(|err| anyhow::anyhow!("LocalChain update error: {:?}", err))?;
+    let _ = graph.apply_update(update.graph_update.clone());
+
+    Ok(update)
 }
 
 #[test]
@@ -60,7 +97,7 @@ pub fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
         None,
     )?;
     env.mine_blocks(1, None)?;
-    env.wait_until_electrum_sees_block()?;
+    env.wait_until_electrum_sees_block(Duration::from_secs(6))?;
 
     // use a full checkpoint linked list (since this is not what we are testing)
     let cp_tip = env.make_checkpoint_tip();
@@ -162,7 +199,7 @@ pub fn test_update_tx_graph_stop_gap() -> anyhow::Result<()> {
         None,
     )?;
     env.mine_blocks(1, None)?;
-    env.wait_until_electrum_sees_block()?;
+    env.wait_until_electrum_sees_block(Duration::from_secs(6))?;
 
     // use a full checkpoint linked list (since this is not what we are testing)
     let cp_tip = env.make_checkpoint_tip();
@@ -204,7 +241,7 @@ pub fn test_update_tx_graph_stop_gap() -> anyhow::Result<()> {
         None,
     )?;
     env.mine_blocks(1, None)?;
-    env.wait_until_electrum_sees_block()?;
+    env.wait_until_electrum_sees_block(Duration::from_secs(6))?;
 
     // A scan with gap limit 5 won't find the second transaction, but a scan with gap limit 6 will.
     // The last active indice won't be updated in the first case but will in the second one.
@@ -238,14 +275,9 @@ pub fn test_update_tx_graph_stop_gap() -> anyhow::Result<()> {
     Ok(())
 }
 
-/// Ensure that [`ElectrumExt`] can sync properly.
-///
-/// 1. Mine 101 blocks.
-/// 2. Send a tx.
-/// 3. Mine extra block to confirm sent tx.
-/// 4. Check [`Balance`] to ensure tx is confirmed.
+/// Ensure that [`BdkElectrumClient::sync`] can confirm previously unconfirmed transactions in both reorg and no-reorg situations.
 #[test]
-fn scan_detects_confirmed_tx() -> anyhow::Result<()> {
+fn test_sync() -> anyhow::Result<()> {
     const SEND_AMOUNT: Amount = Amount::from_sat(10_000);
 
     let env = TestEnv::new()?;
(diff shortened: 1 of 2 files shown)


crates/electrum/tests/test_electrum.rs
```diff
@@ -22,10 +23,34 @@ fn get_balance(
     Ok(balance)
 }
 
+fn electrum_sync<I>(
```
evanlinjin: **Nit:** according to [standard naming conventions in Rust](https://rust-lang.github.io/api-guidelines/naming.html#names-use-a-consistent-word-order-c-word-order), let's start with the verb for this function (since this is how we name things in BDK, let me know if you find an exception though). I.e. `sync_with_electrum`.

crates/electrum/tests/test_electrum.rs
```diff
@@ -22,10 +23,34 @@ fn get_balance(
...
+    let _ = graph.apply_update(update.graph_update.clone());
+
+    Ok(update)
+}
+
 #[test]
 pub fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
     let env = TestEnv::new()?;
-    let electrum_client = electrum_client::Client::new(env.electrsd.electrum_url.as_str())?;
+    let electrum_client = Client::new(env.electrsd.electrum_url.as_str())?;
```
evanlinjin: **Nit:** Ideally, we don't change lines of code that we don't need to touch. Same as line `151`. In this file, could we use the full module path to refer to `Client`?

However, if you see a problem with using the full path, I think it makes sense to include this change in a separate commit.

crates/electrum/tests/test_electrum.rs
```diff
@@ -238,18 +263,13 @@ pub fn test_update_tx_graph_stop_gap() -> anyhow::Result<()> {
     Ok(())
 }
 
-/// Ensure that [`ElectrumExt`] can sync properly.
-///
-/// 1. Mine 101 blocks.
-/// 2. Send a tx.
-/// 3. Mine extra block to confirm sent tx.
-/// 4. Check [`Balance`] to ensure tx is confirmed.
+/// Ensure that [`BdkElectrumClient`] can sync properly in both reorg and no-reorg situations.
```
evanlinjin: This doc comment is a little ambiguous. We may need to expand on "sync properly" here and describe the scenario in more detail. I think you did a better job describing the test in the PR description than in this doc comment.

crates/electrum/tests/test_electrum.rs
```diff
@@ -22,10 +23,34 @@ fn get_balance(
     Ok(balance)
 }
 
+fn electrum_sync<I>(
+    client: &BdkElectrumClient<Client>,
+    spks: &[ScriptBuf],
```
evanlinjin: **Nit:** there is no point having a reference here, since we are cloning this later on (with `.to_owned()`).

crates/electrum/tests/test_electrum.rs
```diff
@@ -22,10 +23,34 @@ fn get_balance(
...
+    batch_size: usize,
+    chain: &mut LocalChain,
+    graph: &mut IndexedTxGraph<ConfirmationBlockTime, I>,
+) -> anyhow::Result<SyncResult>
+where
+    I: Indexer,
+    I::ChangeSet: Default + Merge,
```
evanlinjin: **Nit:** This removes the unnecessary reference and allows more flexibility for what we can input in `spks`.

```suggestion
fn electrum_sync<I, Spks>(
    client: &BdkElectrumClient<Client>,
    spks: Spks,
    batch_size: usize,
    chain: &mut LocalChain,
    graph: &mut IndexedTxGraph<ConfirmationBlockTime, I>,
) -> anyhow::Result<SyncResult>
where
    I: Indexer,
    I::ChangeSet: Default + Merge,
    Spks: IntoIterator<Item = ScriptBuf>,
    Spks::IntoIter: ExactSizeIterator + Send + 'static,

```

crates/electrum/tests/test_electrum.rs
```diff
@@ -22,10 +23,34 @@ fn get_balance(
     Ok(balance)
 }
 
+fn electrum_sync<I>(
+    client: &BdkElectrumClient<Client>,
+    spks: &[ScriptBuf],
+    batch_size: usize,
```
evanlinjin: I think it's okay to hard-code the batch size and have it as a constant at the top of the file (instead of having it as an input to this function). The value is not relevant for what we are testing and not having it slightly increases readability for the tests.

crates/electrum/tests/test_electrum.rs
```diff
@@ -272,32 +292,91 @@ fn scan_detects_confirmed_tx() -> anyhow::Result<()> {
     // Mine some blocks.
     env.mine_blocks(101, Some(addr_to_mine))?;
 
-    // Create transaction that is tracked by our receiver.
+    // Broadcast transaction to mempool.
     env.send(&addr_to_track, SEND_AMOUNT)?;
+    env.wait_until_electrum_sees_block()?;
```
evanlinjin: We are sending tx to mempool, but we are waiting to see a block? Is there a method that waits for electrum to see a certain tx in mempool? If not, we should add it.

crates/electrum/tests/test_electrum.rs
```diff
@@ -272,32 +292,91 @@ fn scan_detects_confirmed_tx() -> anyhow::Result<()> {
...
+
+    // Check if balance is zero when transaction exists only in mempool.
+    assert_eq!(
+        get_balance(&recv_chain, &recv_graph)?,
+        Balance {
+            confirmed: Amount::from_sat(0),
+            ..Balance::default()
+        },
+        "initial balance must be correct",
+    );
```
evanlinjin: Since we have a tx in mempool, shouldn't we have an unconfirmed balance?

If the intention is to sync BEFORE seeing the tx in mempool, it should be made more obvious in code.

crates/electrum/tests/test_electrum.rs
```diff
@@ -272,32 +292,91 @@ fn scan_detects_confirmed_tx() -> anyhow::Result<()> {
...
-        .map_err(|err| anyhow::anyhow!("LocalChain update error: {:?}", err))?;
-    let _ = recv_graph.apply_update(update.graph_update);
+    // Check if balance is correct when transaction returns to mempool.
+    assert_eq!(
+        get_balance(&recv_chain, &recv_graph)?,
+        Balance {
+            confirmed: Amount::from_sat(0),
+            ..Balance::default()
+        },
+    );
```
evanlinjin: This result is assuming that the receiving structures have not seen the tx in mempool yet.

If the receiving structures are aware of the tx being in mempool, there should be an unconfirmed balance here.

crates/electrum/tests/test_electrum.rs
```diff
@@ -272,32 +292,91 @@ fn scan_detects_confirmed_tx() -> anyhow::Result<()> {
...
     assert_eq!(
         get_balance(&recv_chain, &recv_graph)?,
         Balance {
             confirmed: SEND_AMOUNT,
             ..Balance::default()
         },
+        "initial balance must be correct",
     );
 
     for tx in recv_graph.graph().full_txs() {
```
evanlinjin, line 390: We need a comment before the for loop to explain what is being tested in the for loop.

evanlinjin (changes requested):
Good work with this. Moving forward, I recommend improving the docs and adding more clarity to the testing logic. I also included a bunch of nits for you to consider including.

crates/electrum/tests/test_electrum.rs
evanlinjin: @LagginTimes I think you attempted to change this, but you ended up changing the wrong thing 😅

LagginTimes: Oops.

evanlinjin:
The commit also adds a method to the `TestEnv` (which is not documented). Please either have this change as a separate commit, or mention the change in the commit.

crates/testenv/src/lib.rs
```diff
@@ -186,6 +186,22 @@ impl TestEnv {
         }
     }
 
+    /// This method waits for Electrum to see a transaction with given `txid`.
```
evanlinjin:
```suggestion
    /// This method waits for Electrsd to see a transaction with given `txid`.
```

crates/testenv/src/lib.rs
```diff
@@ -186,6 +186,22 @@ impl TestEnv {
...
+    pub fn wait_until_electrum_sees_txid(&self, txid: Txid) -> anyhow::Result<()> {
+        let mut delay = Duration::from_millis(64);
+
+        loop {
+            if self.electrsd.client.transaction_get(&txid).is_ok() {
+                return Ok(());
+            }
```
evanlinjin: This will create a situation where we may loop forever.

Are we able to figure out if the error is a not-exist error? Ideally, the `rust_esplora` crate has something to figure this out easily. If not, we should upstream some changes later.

Also, would be good to add a timeout as an input. Just in case the test fails (the txid never shows up). We at least want to exit the test!

evanlinjin (changes requested):

crates/testenv/src/lib.rs
evanlinjin: The timeout thing can be added in a separate PR. Can also add timeouts to the other wait methods.

LagginTimes: The error returned is: `Err(Protocol(String("missing transaction")))`

crates/testenv/src/lib.rs
```diff
@@ -168,22 +168,36 @@ impl TestEnv {
     }
 
     /// This method waits for the Electrum notification indicating that a new block has been mined.
-    pub fn wait_until_electrum_sees_block(&self) -> anyhow::Result<()> {
+    /// `timeout` is the maximum amount of seconds we want to wait for a response from Electrsd.
+    pub fn wait_until_electrum_sees_block(&self, timeout: u32) -> anyhow::Result<()> {
```
evanlinjin: Can we please make `timeout` a `Duration`?

You can instantiate an `Instant` before the loop and check `Instant::elapsed` against `timeout` within the loop to determine if we need to break.

https://doc.rust-lang.org/std/time/struct.Instant.html#method.elapsed

crates/testenv/src/lib.rs
```diff
@@ -168,22 +168,42 @@ impl TestEnv {
...
             }
+            std::thread::sleep(delay);
+        }
+        Err(anyhow::Error::msg(
+            "Timed out waiting for Electrsd to get block header",
+        ))
+    }
 
-            if delay.as_millis() < 512 {
-                delay = delay.mul_f32(2.0);
```
evanlinjin, line 183: Is there a reason why we got rid of the exponential backoff strategy?

evanlinjin: As mentioned in DMs, I'm okay with removing this, but can we please document this removal in the commit description? Thanks!

crates/electrum/tests/test_electrum.rs
```diff
@@ -271,35 +303,88 @@ fn scan_detects_confirmed_tx() -> anyhow::Result<()> {
...
 
-    // Mine a block to confirm sent tx.
+    sync_with_electrum(
+        &client,
+        [spk_to_track.clone()],
+        &mut recv_chain,
+        &mut recv_graph,
+    )?;
+
+    // Check if balance is zero when transaction exists only in mempool.
```
evanlinjin: Might need to update this comment now, since we have a pending balance.

crates/electrum/tests/test_electrum.rs
```diff
@@ -271,35 +303,88 @@ fn scan_detects_confirmed_tx() -> anyhow::Result<()> {
...
         get_balance(&recv_chain, &recv_graph)?,
         Balance {
             confirmed: SEND_AMOUNT,
             ..Balance::default()
         },
+        "balance must be correct",
     );
 
+    // Check to see if we have the floating txouts available from our transactions' previous outputs
+    // in order to calculate transaction fees.
```
evanlinjin, line 389: Since we are testing this, might make sense to mention this in the test's doc comments.

evanlinjin (changes requested):
LGTM. Once these are addressed and changelog is added to PR description, I am happy to merge.

evanlinjin (approved):
ACK 2c0bc45ecf0241abc8bf9d72c8bbeded53d32a45

crates/electrum/tests/test_electrum.rs
```diff
@@ -407,24 +490,21 @@ fn tx_can_become_unconfirmed_after_reorg() -> anyhow::Result<()> {
...
-            .map_err(|err| anyhow::anyhow!("LocalChain update error: {:?}", err))?;
-
         // Check that no new anchors are added during current reorg.
         assert!(initial_anchors.is_superset(update.graph_update.all_anchors()));
-        let _ = recv_graph.apply_update(update.graph_update);
 
         assert_eq!(
             get_balance(&recv_chain, &recv_graph)?,
             Balance {
+                trusted_pending: SEND_AMOUNT * depth as u64,
```
ValuedMammal, line 507: I agree this line should be here. It made me wonder why after #1416 the assertion failed, but I think the difference here is that we're updating last-seen during `sync_with_electrum` as we should.

see related #1550

PR #1541, opened 2024-08-08
Force `Persisted<T>` to mutate and persist atomically

evanlinjin:
### Description

@tnull asked why we have to do a separate `.persist` call when using `Persisted<Wallet>`. This is my solution to not having to do that.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/persist.rs
+++ b/crates/chain/src/persist.rs
@@ -1,8 +1,4 @@
-use core::{
-    future::Future,
-    ops::{Deref, DerefMut},
-    pin::Pin,
-};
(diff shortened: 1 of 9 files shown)


tnull:
Thank you for looking into this! It seems like a viable option even though it somewhat complicates the wallet API.

Although, as noted on Discord, I personally might even prefer the current simpler API as long as it clearly communicates when persistence should/may happen (e.g., would return a result along the lines of enum PendingPersistence { Immediately, Eventually } for all codepaths that need persistence), and the docs would provide clear guidance how to handle persistence failures.

evanlinjin:
Closing this because I don't think it adds much.

PR #1544, opened 2024-08-09
Remove crate-renaming

evanlinjin:
### Description

[Rust 1.60 introduces *"namespaced features"*](https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#new-syntax-for-cargo-features) so we can add a dependency without implicitly also defining a feature of the same name.

We can now get rid of crate-renaming and use "namespaced features" instead.

Thanks to @LLFourn for pointing this out.

### Changelog notice

* Remove crate-renaming in `bdk_chain` and use "namespaced features" instead.

#### All Submissions:

### DIFF
--- a/crates/chain/Cargo.toml
+++ b/crates/chain/Cargo.toml
@@ -14,14 +14,14 @@ readme = "README.md"
 
 [dependencies]
 bitcoin = { version = "0.32.0", default-features = false }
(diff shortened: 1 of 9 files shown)


LLFourn:
ACK 318ce095f2adec9d03077ddf65470ff475547f39

LLFourn (approved):

PR #1545, opened 2024-08-09
feat(testenv): Add method `new_with_config`

ValuedMammal:
I think this will be useful for passing a custom bitcoin config including e.g. `-blockfilterindex` that would not otherwise be included in the default config.

### Notes to the reviewers

Any other suggestions let me know @LagginTimes 

### Changelog notice

* Added method `TestEnv::new_with_config` for passing custom `bitcoind::Conf`, `electrsd::Conf` config objects

#### All Submissions:

### DIFF
--- a/crates/testenv/src/lib.rs
+++ b/crates/testenv/src/lib.rs
@@ -11,6 +11,8 @@ use bitcoincore_rpc::{
     bitcoincore_rpc_json::{GetBlockTemplateModes, GetBlockTemplateRules},
     RpcApi,
 };
+use electrsd::bitcoind::anyhow::Context;
+
 pub use electrsd;
 pub use electrsd::bitcoind;
 pub use electrsd::bitcoind::anyhow;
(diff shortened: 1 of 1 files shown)


crates/testenv/src/lib.rs
```diff
@@ -29,32 +31,48 @@ pub struct TestEnv {
...
+                conf.http_enabled = true;
+                conf
             }
-        }?;
+        };
+        TestEnv::_new(&bitcoind_conf, &electrsd_conf)
+    }
+
+    /// Construct a new [`TestEnv`] instance with the given configs.
+    fn _new(
```
evanlinjin: The intention of this is so that we have our own defaults for the config structures? (Enable http)

If this is the case, I would prefer we have our own config struct and impl default there. I think it would be easier to reason about. 

Also, I'm curious about the intention to enable http on default. Is it because electrsd requires it?

ValuedMammal: I pushed a commit implementing your suggestion. The other option is to upstream the change to `electrsd` crate making `http_enabled` the default unless there's a particular reason they chose to set it `false`.

ValuedMammal:
> I'm curious about the intention to enable http on default. Is it because electrsd requires it?

I tried turning it off, and running

`cargo t -p bdk_esplora --all-features -- --test-threads=1`

gave this error

```
running 3 tests
test async_ext::test::test_finalize_chain_update ... FAILED
test blocking_ext::test::test_finalize_chain_update ... FAILED
test blocking_ext::test::update_local_chain ... ok

failures:

---- async_ext::test::test_finalize_chain_update stdout ----
[0] running test case: chain_extends
thread 'async_ext::test::test_finalize_chain_update' panicked at crates/esplora/src/async_ext.rs:469:83:
called `Option::unwrap()` on a `None` value

---- blocking_ext::test::test_finalize_chain_update stdout ----
[0] running test case: chain_extends
thread 'blocking_ext::test::test_finalize_chain_update' panicked at crates/esplora/src/blocking_ext.rs:466:83:
called `Option::unwrap()` on a `None` value
```

crates/testenv/src/lib.rs
evanlinjin: I meant to have a config struct that incorporated both `bitcoind::Conf` and `electrsd::Conf`. This way, we can get rid of the `_new` method.

ValuedMammal: Ah, ok. will update soon

LagginTimes:
ACK b3b65514b7243664d24e4a9623d93ef619650d28

evanlinjin (approved):
ACK b3b65514b7243664d24e4a9623d93ef619650d28

PR #1547, opened 2024-08-11
Simplify wallet persistence

evanlinjin:
### Description

Removed the persistence module in `bdk_chain` (which contained the `PersistWith` and `PersistAsyncWith` traits and a `Persisted<T>` wrapper). Replaced it with simplified versions that are `Wallet`-specific.

The new traits (`WalletPersister` and `AsyncWalletPersister`) are simpler since they have less type-parameters (being wallet-specific). The old versions were more complex as they were intended to be used with any type. However, we need more time to finalize a works-for-all-bdk-types persistence API.

Additionally, `WalletPersister` and `AsyncWalletPersister` also introduce the `initialize` method. It is handy to contain db-initialization (i.e. create tables for SQL) here. We can call `initialize` in `PersistedWallet` constructors. So the `PersistedWallet::persist` method does not need to ensure the database is initialized. To accommodate this, I made the `.init_sqlite_table` methods on changeset-types public, and removed the call of this in `from_sqlite` and `.persist_to_sqlite`. The `initialize` method now loads from the persister, so the `load` associated function is removed.

There was a bug in the old `PersistAsyncWith` trait where the lifetime bounds were not strict enough (refer to the conversation in #1552). This is now fixed in the new `AsyncWalletPersister` trait.

Docs for the new types are clearer (hopefully). I mentioned implementation details about the new traits and what safety checks were guaranteed in `PersistedWallet`.

I think it makes sense just to have a wallet-specific persistence API which we will maintain alongside wallet for v1.0. This is less baggage for users and maintainers alike.

### Notes to the reviewers

**How breaking are these changes?**

Unless if you were implementing your own persistence, then not breaking at all! If you were implementing your own persistence for BDK, the breakages are minimal. The main change is introducing the `initialize` method to persistence traits (which replaces `load`).

**Acknowledgements**

I came up with the idea of this while responding to @tnull's suggestions on Discord. Unfortunately we cannot fulfill all of his requests. However, I do believe that this will make things easier for our users.

### Changelog notice

* Removed the `persist` module in `bdk_chain`.
* Added `WalletPersister`/`AsyncWalletPersister` traits and `PersistedWallet` struct to `bdk_wallet`. These are simplified and safer versions of old structs provided by the `persist` module in `bdk_chain`.
* Change `.init_sqlite_tables` method to be public on changeset types. `from_sqlite` and `persist_into_sqlite` methods no longer call `.init_sqlite_tables` internally.

#### All Submissions:

#### What is left to do:

- [x] Better docs.


notmandatory:
Concept ACK

Even though we need to stop making breaking changes, this looks like like a reasonable isolated change that will simplify the persistence API for Wallet users.

notmandatory:
LGTM, needs `fmt`, rebase on master, and commits cleanup. I tested with wallet and cli examples and worked fine, including loading existing db created with prior version.

evanlinjin:
An idea would be to introduce some sort of test framework for `WalletPersister` and `AsyncWalletPersister` implementations.

Checks include:

* Calling `initialize` multiple times should not error, or make database inconsistent.
* Calling `initialize` after `persist` should not error, or make database inconsistent.
* What previously went in should come out.

Maybe a more elaborate version of https://github.com/bitcoindevkit/bdk/blob/e0822d7a7654a9e0669a43bdd50894872aa49f1a/crates/wallet/tests/wallet.rs#L105?

crates/wallet/src/wallet/params.rs
```diff
@@ -109,25 +110,25 @@ impl CreateParams {
     }
 
     /// Create [`PersistedWallet`] with the given `Db`.
```
notmandatory:
```suggestion
    /// Create [`PersistedWallet`] with the given [`WalletPersister`].
```

crates/wallet/src/wallet/params.rs
```diff
@@ -219,26 +220,26 @@ impl LoadParams {
         self
     }
 
-    /// Load [`PersistedWallet`] with the given `Db`.
-    pub fn load_wallet<Db>(
+    /// Load [`PersistedWallet`] with the given `persister`.
```
notmandatory:
```suggestion
    /// Load [`PersistedWallet`] with the given [`WalletPersister`].
```

crates/wallet/src/wallet/params.rs
```diff
@@ -219,26 +220,26 @@ impl LoadParams {
...
-        Wallet: PersistWith<Db, LoadParams = Self>,
+        P: WalletPersister,
     {
-        PersistedWallet::load(db, self)
+        PersistedWallet::load(persister, self)
     }
 
-    /// Load [`PersistedWallet`] with the given async `Db`.
-    pub async fn load_wallet_async<Db>(
+    /// Load [`PersistedWallet`] with the given async `persister`.
```
notmandatory:
```suggestion
    /// Load [`PersistedWallet`] with the given [`AsyncWalletPersister`].
```

crates/wallet/src/wallet/params.rs
```diff
@@ -109,25 +110,25 @@ impl CreateParams {
...
+    ) -> Result<PersistedWallet<P>, CreateWithPersistError<P::Error>>
     where
-        Wallet: PersistWith<Db, CreateParams = Self>,
+        P: WalletPersister,
     {
-        PersistedWallet::create(db, self)
+        PersistedWallet::create(persister, self)
     }
 
     /// Create [`PersistedWallet`] with the given async `Db`.
```
notmandatory:
```suggestion
    /// Create [`PersistedWallet`] with the given [`AsyncWalletPersister`].
```

crates/wallet/src/wallet/persisted.rs
```diff
@@ -1,130 +1,320 @@
...
+    ///
+    /// # Implementation Details
+    ///
+    /// The database schema of the `persister` (if any), should be initialized and migrated here.
+    ///
+    /// The implementation must return all data currently stored in the `persister`. If there is no
+    /// data, return an empty changeset (using [`ChangeSet::default()`]).
+    ///
+    /// Error should only occur on database failure. Multiple calls to `initialize` should not
+    /// error. Calling [`persist`] before calling `initialize` should not error either.
```
notmandatory: In the docs below for `persist` you said that "The method can fail if the `persister` is not [`initialize`]d" so this comment should be updated to agree right?

```suggestion
    /// error. Calling [`persist`] before calling `initialize` can error.
```

crates/wallet/src/wallet/persisted.rs
```diff
@@ -1,130 +1,320 @@
...
```
notmandatory: Same comment as above.
```suggestion
    /// error. Calling [`persist`] before calling `initialize` can error.
```

crates/wallet/src/wallet/persisted.rs
```diff
@@ -1,130 +1,320 @@
...
+                Ok(true)
+            }
+            None => Ok(false),
+        }
+    }
+}
+
+/// Methods when `P` is an [`AsyncWalletPersister`].
+impl<P: AsyncWalletPersister> PersistedWallet<P> {
+    /// Create a new [`PersistedWallet`] witht the given async `persister` and `params`.
```
notmandatory:
```suggestion
    /// Create a new [`PersistedWallet`] with the given async `persister` and `params`.
```

notmandatory (changes requested):
I did a quick review and found a few doc fixes but otherwise looks good. I'll do a deeper look this afternoon but so far looks ready to go.

crates/wallet/src/wallet/persisted.rs
ValuedMammal: Or maybe docs for `persist` should say "This method should ensure the persister is initialized"

crates/wallet/src/wallet/mod.rs
```diff
@@ -2312,6 +2303,15 @@ impl Wallet {
         }
     }
 
+    /// Get a mutable reference of the staged [`ChangeSet`] that is yet to be commited (if any).
+    pub fn staged_mut(&mut self) -> Option<&mut ChangeSet> {
```
ValuedMammal, line 2307:
```suggestion
    pub(crate) fn staged_mut(&mut self) -> Option<&mut ChangeSet> {
```

crates/wallet/src/wallet/persisted.rs
```diff
@@ -1,130 +1,320 @@
...
+    ///
+    /// This method can fail if the `persister` is not [`initialize`]d.
+    ///
+    /// [`initialize`]: AsyncWalletPersister::initialize
+    fn persist<'a>(
+        persister: &'a mut Self,
+        changeset: &'a ChangeSet,
+    ) -> FutureResult<'a, (), Self::Error>
+    where
+        Self: 'a;
```
ValuedMammal, line 102: wondering if this would achieve the same thing

```suggestion
    fn persist<'a>(&'a mut self, changeset: &'a ChangeSet) -> FutureResult<'a, (), Self::Error>;
```

crates/wallet/src/wallet/persisted.rs
```diff
@@ -1,130 +1,320 @@
...
+    ) -> Result<Self, CreateWithPersistError<P::Error>> {
+        let existing = P::initialize(persister).map_err(CreateWithPersistError::Persist)?;
+        if !existing.is_empty() {
+            return Err(CreateWithPersistError::DataAlreadyExists(existing));
+        }
+        let mut inner =
+            Wallet::create_with_params(params).map_err(CreateWithPersistError::Descriptor)?;
+        if let Some(changeset) = inner.take_staged() {
+            P::persist(persister, &changeset).map_err(CreateWithPersistError::Persist)?;
+        }
```
ValuedMammal, line 155: 🔧  Here you're taking the stage before the call to `persist` has returned

crates/wallet/src/wallet/persisted.rs
```diff
@@ -1,130 +1,320 @@
...
+                    inner,
+                    marker: PhantomData,
+                })
+            })
+            .map_err(LoadWithPersistError::InvalidChangeSet)
+    }
+
+    /// Persist staged changes of wallet into `persister`.
+    ///
+    /// If the `persister` errors, the staged changes will not be cleared.
```
ValuedMammal, line 182:
```suggestion
    /// If the `persister` errors, the staged changes will not be cleared. Returns whether any new changes were persisted.
```

crates/wallet/src/wallet/persisted.rs
```diff
@@ -1,130 +1,320 @@
...
-        if let Some(changeset) = wallet.take_staged() {
-            db.append_changeset(&changeset)
-                .map_err(CreateWithPersistError::Persist)?;
+#[derive(Debug)]
+pub enum FileStoreError {
+    /// Error when loading from the store.
+    Load(bdk_file_store::AggregateChangesetsError<ChangeSet>),
+    /// Error when writing to the store.
+    Write(std::io::Error),
+}
```
ValuedMammal, line 297: Maybe this error should be defined in `bdk_file_store` and we can just import it here?

ValuedMammal (commented):
I did a first pass. Also spent some time implementing `AsyncWalletPersister` for a locally defined type and so far results are positive.

crates/wallet/src/wallet/persisted.rs
ValuedMammal: oh, I think we still need `Self: 'a`.

crates/wallet/src/wallet/persisted.rs
evanlinjin: @ValuedMammal it's a "may" though, since some databases don't require initialization.

```rust
    /// Error should only occur on database failure. Multiple calls to `initialize` should not
    /// error. Calling `initialize` inbetween calls to `persist` should not error.
    ///
    /// Calling [`persist`] before the `persister` is `initialize`d may error. However, some
    /// persister implementations may NOT require initialization at all (and not error).

```

@notmandatory @ValuedMammal what do you guys think about this?

crates/wallet/src/wallet/mod.rs
evanlinjin: I was thinking that this method may be useful for some users. I.e. they don't need to query the staged area twice if they wanted to do something like we do in `PersistedWallet::persist`.

crates/wallet/src/wallet/persisted.rs
evanlinjin: Also, I made purposely made `persist` a non-member so that it's harder to call haha

crates/wallet/src/wallet/persisted.rs
evanlinjin: I think it's okay here since it's within `create`. If we error here, we will loose `Wallet` regardless.

The other option would be to contain the non-persisted `Wallet` in the error type. However, I'm not sure how useful it would be.

crates/wallet/src/wallet/persisted.rs
evanlinjin: Hmmm, it'll be weird in the sense that no methods/functions will return `FileStoreError` in `bdk_file_store`. What would be the benefit of this?

evanlinjin:
@ValuedMammal @notmandatory I updated the docs! Let me know if everything makes sense.

I'll be disappearing this weekend and be back Tuesday night. Feel free to push on commits for more doc fixes and merge (if appropriate).

crates/wallet/src/wallet/mod.rs
ValuedMammal: For the sake of discussion I think you can do it without ever getting the stage mutably such as having a method on `Wallet` that resets the stage but doesn't return it, and this would prevent users from freely modifying their changesets.

```rust
  /// Clears the stage.
  pub fn clear_stage(&mut self) { self.stage = ChangeSet::default(); }
```

Then `persist` would look like

```rust
  /// Persist staged changes.
  pub fn persist(&mut self, persister: &mut P) -> Result<bool, P::Error> {
      if let Some(change) = self.inner.staged() {
          P::persist(persister, change)?;
          self.inner.clear_stage();
          return Ok(true);
      }
      Ok(false)
  }
```

crates/wallet/src/wallet/persisted.rs
```diff
@@ -1,130 +1,305 @@
...
+
+type FutureResult<'a, T, E> = Pin<Box<dyn Future<Output = Result<T, E>> + Send + 'a>>;
+
+/// Async trait that persists [`Wallet`].
+///
+/// For a blocking version, use [`WalletPersister`].
+///
+/// Associated functions of this trait should not be called directly, and the trait is designed so
+/// that associated functions are hard to find (since they are not methods!). [`WalletPersister`] is
```
oleonardolima:
```suggestion
/// that associated functions are hard to find (since they are not methods!). [`AsyncWalletPersister`] is
```

crates/wallet/src/wallet/persisted.rs
```diff
@@ -1,130 +1,305 @@
...
+/// used by [`PersistedWallet`] (a light wrapper around [`Wallet`]) which enforces some level of
+/// safety. Refer to [`PersistedWallet`] for more about the safety checks.
+pub trait AsyncWalletPersister {
+    /// Error type of the persister.
+    type Error;
+
+    /// Initialize the `persister` and load all data.
+    ///
+    /// This is called by [`PersistedWallet::create_async`] and [`PersistedWallet::load_async`] to
+    /// ensure the [`WalletPersister`] is initialized and returns all data in the `persister`.
```
oleonardolima:
```suggestion
    /// ensure the [`AsyncWalletPersister`] is initialized and returns all data in the `persister`.
```

oleonardolima:
Concept ACK
It's looking good to me. I guess I found only two documentation fixes.

crates/wallet/src/wallet/mod.rs
oleonardolima: Agree, I think it's a better approach to have/use a `clear_stage` method, having explicit and clear meaning.

crates/wallet/src/wallet/persisted.rs
```diff
@@ -1,130 +1,330 @@
...
+    /// data, return an empty changeset (using [`ChangeSet::default()`]).
+    ///
+    /// Error should only occur on database failure. Multiple calls to `initialize` should not
+    /// error. Calling `initialize` inbetween calls to `persist` should not error.
+    ///
+    /// Calling [`persist`] before the `persister` is `initialize`d may error. However, some
+    /// persister implementations may NOT require initialization at all (and not error).
+    ///
+    /// [`persist`]: WalletPersister::persist
+    fn initialize(persister: &mut Self) -> Result<ChangeSet, Self::Error>;
```
ValuedMammal, line 45: small nit: 

It might be more convenient for the implementer if the return type is `Result<Option<ChangeSet>, Self::Error>`.

ValuedMammal (approved):
ACK f434c93a788c6334319c40de7fe961113b9e415f (pending comments by @oleonardolima )

crates/wallet/src/wallet/persisted.rs
notmandatory: I amended the last push with this fix.

crates/wallet/src/wallet/persisted.rs
notmandatory: I amended the last push with this fix.

crates/wallet/src/wallet/persisted.rs
notmandatory: I amended the last push with this fix.

notmandatory (approved):
ACK 340808e820cc144c8b2756dd9c235fad1400ad76

PR #1549, opened 2024-08-12
fix(example_cli): add bitcoin and rand dependencies

notmandatory:
### Description

Fix building `example_cli` by adding the bitcoin and rand dependencies so it, and the crates that depend on it, can be built independently and not only from the workspace.

### Notes to the reviewers

I also added the  build-examples CI job to verify we can build examples individually.

### Changelog notice

None.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/example-crates/example_cli/src/lib.rs
+++ b/example-crates/example_cli/src/lib.rs
@@ -10,14 +10,9 @@ use std::sync::Mutex;
 use anyhow::bail;
 use anyhow::Context;
 use bdk_chain::bitcoin::{
-    absolute,
-    address::NetworkUnchecked,
-    bip32, consensus, constants,
-    hex::DisplayHex,
-    relative,
-    secp256k1::{rand::prelude::*, Secp256k1},
-    transaction, Address, Amount, Network, NetworkKind, PrivateKey, Psbt, PublicKey, Sequence,
-    Transaction, TxIn, TxOut,
+    absolute, address::NetworkUnchecked, bip32, consensus, constants, hex::DisplayHex, relative,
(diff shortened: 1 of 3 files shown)


example-crates/example_cli/src/lib.rs
```diff
@@ -667,15 +667,15 @@ pub fn handle_commands<CS: clap::Subcommand, S: clap::Args>(
                     // print base64 encoded psbt
                     let fee = psbt.fee()?.to_sat();
                     let mut obj = serde_json::Map::new();
-                    obj.insert("psbt".to_string(), json!(psbt.to_string()));
+                    obj.insert("psbt".to_string(), json!(psbt));
```
ValuedMammal: I think `to_string` is needed to get the PSBT in base64. You can debug print the struct using `-d`

notmandatory: Good catch! fixed.

I also verified I could create and broadcast a PSBT with the base64 encoded output of this command.

ValuedMammal:
I'm not sure if it matters but we could possibly just add a bitcoin dependency to `example_cli` with the necessary features and not have to add new features to `bdk_chain`

notmandatory:
> I'm not sure if it matters but we could possibly just add a bitcoin dependency to `example_cli` with the necessary features and not have to add new features to `bdk_chain`

Good idea, the `bdk_wallet` crate had it's own dependency on `bitcoin` and `rand` (for tests). I updated this PR to not add features to `bdk_chain` and instead add the missing local dependencies to `example_cli`.

ValuedMammal (approved):
ACK 3675a9e4921f2002f23a32d2f76d65549899edd4

storopoli (approved):
ACK 3675a9e4921f2002f23a32d2f76d65549899edd4

Issue #1550, opened 2024-08-13
Should `bdk_electrum` `sync`/`full_scan` set `last_seen`?


LagginTimes:
Getting a balance from `TxGraph` will not return a balance for unconfirmed transactions immediately after `sync`/`full_scan` with `bdk_electrum`. This is because the `last_seen` for these transactions is missing, and has to be set separately from `sync`/`full_scan` with `TxGraph::update_last_seen_unconfirmed`. Would it make sense to set the `last_seen` for unconfirmed transactions during the `sync`/`full_scan` process?

ValuedMammal:
Yeah we could have something like `time_of_sync` be a member of `SyncRequest` for example.

evanlinjin:
> Yeah we could have something like `time_of_sync` be a member of `SyncRequest` for example.

I think that's a great idea.

@LagginTimes thoughts?

notmandatory:
This looks like the kind of issue we should try to fix in the beta to prevent users from running into it. The fix proposed by @ValuedMammal sounds relatively painless.

notmandatory:
Also @ValuedMammal mentioned in the call today that we need to be careful not to call functions like `std::time` internally and break the `no-std` (and WASM) compatibility.  Passing `time_of_sync` in with the sync/full_scan request should also work with `no-std` apps.

Issue #1543, opened 2024-08-09
Make chain-sources non-dependent on a specific `bdk_chain` version.


evanlinjin:
## Describe the problem

We may want to be able to iterate fast on `bdk_chain` without breaking `Wallet`. This means `Wallet` may be of an older version of `bdk_chain`. Therefore, `Wallet` (currently) will have to use older versions of chain sources (since our chain sources depend on `bdk_chain`).

It will be a nice property if projects that use BDK can stay put on a specific `bdk_chain` version, but still use the latest chain-source crate version.

## Additional context

**Why would we need to break `bdk_chain`?**

There are a couple a breaking changes that we simply do not have enough time to include in the `v1.0-beta` release.

1. #1510
    Get rid of `IndexedTxGraph` and just put the `Indexer` in `TxGraph`. This removes a bunch of overhead and is in general much cleaner.
2. #1515
    Removing the anchor trait, and representing the `Anchor` as a concrete type is much cleaner and a less error-prone API.

Both of these changes will break `ChangeSet`.

## Proposed solution

Introduce `bdk_core` crate that moves structures and types from `bdk_chain` that will mostly stay constant.

* Types: `BlockId`, `ConfirmationBlockTime`.
* Updates: `CheckPoint`, introduce `tx_graph::Update` (instead of using `TxGraph` to update `TxGraph`).
* Spk-client types: `FullScanRequest`, `FullScanResult`, `SyncRequest`, `SyncResult`.

Chain sources will only depend on `bdk_core`. Note that we need to introduce `tx_graph::Update` though.

## Potential downsides

`bdk_core` will still need to depend on `bitcoin`. Any major version updates to `bitcoin` will be a breaking change and we will need to introduce a new major version for `bdk_core`. To make these suggested changes useful, we need to lock down on specific major releases of `bitcoin`.

## Acknowledgement

Thank you to @LLFourn for bringing up this idea.

oleonardolima:
I'm not well versed on the `bdk_chain`, but it looks like a good approach.

Concept ACK, as long as we settle on the same rust-bitcoin version of our other crates and the ecosystem (0.32.x ?), in that way I think we "minimize" any further downsides.

I'll put more thought into it overnight, but we might've other structures/types that could be moved to `bdk_core` too 🤔.

LagginTimes:
I'll work towards making a concept PR for this.

notmandatory:
[...]

PR #1553, opened 2024-08-13
feat(chain): Introduce `tx_graph::Update` WIP

LagginTimes:
### Description

Introduces the `tx_graph::Update` struct, which will be used to update a `TxGraph`. This will replace the old method of updating a `TxGraph` with a `TxGraph`.

### Notes to the reviewers

With the current implementation, there is some goofiness in the `esplora`/`electrum` wallet examples and `test_electrum::sync_with_electrum()`, where an `Update` is converted into a `TxGraph` to execute `update_last_seen_unconfirmed()`, before reverting the `TxGraph` back into an `Update` again. Implementing #1550 would prevent the need for this.

### Changelog notice

WIP

#### All Submissions:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -89,7 +89,8 @@
 //! [`insert_txout`]: TxGraph::insert_txout
 
 use crate::{
-    collections::*, Anchor, Balance, BlockId, ChainOracle, ChainPosition, FullTxOut, Merge,
+    collections::*, Anchor, Balance, BlockId, ChainOracle, ChainPosition, ConfirmationBlockTime,
+    FullTxOut, Merge,
 };
 use alloc::collections::vec_deque::VecDeque;
 use alloc::sync::Arc;
@@ -640,15 +641,15 @@ impl<A: Clone + Ord> TxGraph<A> {
     ///
     /// The returned [`ChangeSet`] is the set difference between `update` and `self` (transactions that
     /// exist in `update` but not in `self`).
-    pub fn apply_update(&mut self, update: TxGraph<A>) -> ChangeSet<A> {
-        let changeset = self.determine_changeset(update);
+    pub fn apply_update(&mut self, update: impl Into<Update<A>>) -> ChangeSet<A> {
+        let changeset = self.determine_changeset(update.into());
         self.apply_changeset(changeset.clone());
         changeset
     }
 
     /// Determines the [`ChangeSet`] between `self` and an empty [`TxGraph`].
     pub fn initial_changeset(&self) -> ChangeSet<A> {
-        Self::default().determine_changeset(self.clone())
+        Self::default().determine_changeset(Update::from(self.clone()))
     }
 
     /// Applies [`ChangeSet`] to [`TxGraph`].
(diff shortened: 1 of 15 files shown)


evanlinjin:
@LagginTimes forgot to mention that this is step 1 for #1543

crates/chain/src/tx_graph.rs
```diff
@@ -640,15 +633,15 @@ impl<A: Clone + Ord> TxGraph<A> {
     ///
     /// The returned [`ChangeSet`] is the set difference between `update` and `self` (transactions that
     /// exist in `update` but not in `self`).
-    pub fn apply_update(&mut self, update: TxGraph<A>) -> ChangeSet<A> {
+    pub fn apply_update(&mut self, update: Update<A>) -> ChangeSet<A> {
```
evanlinjin: Maybe `update: impl Into<Update>` will make things flexible and less breaking.

crates/chain/src/tx_graph.rs
```diff
@@ -715,33 +708,38 @@ impl<A: Clone + Ord> TxGraph<A> {
...
-    pub(crate) fn determine_changeset(&self, update: TxGraph<A>) -> ChangeSet<A> {
+    pub(crate) fn determine_changeset(&self, mut update: Update<A>) -> ChangeSet<A> {
         let mut changeset = ChangeSet::<A>::default();
 
-        for (&txid, (update_tx_node, _)) in &update.txs {
-            match (self.txs.get(&txid), update_tx_node) {
-                (None, TxNodeInternal::Whole(update_tx)) => {
-                    changeset.txs.insert(update_tx.clone());
+        for (txid, tx) in update.whole_txs {
+            match self.txs.get(&txid) {
```
evanlinjin, line 723: ~~Using the `.entry` API will be more efficient and cleaner.~~ Sorry I misread, we are modifying changeset and reading from self/update.

notmandatory:
I would like to avoid unnecessary refactoring like this for the 1.0 beta milestone, [see my comment in #1543](https://github.com/bitcoindevkit/bdk/issues/1543#issuecomment-2287139890).

evanlinjin:
@notmandatory I suggested this change because I thought it was necessary for 1.0. Please refer to @LLFourn's comment here: https://github.com/bitcoindevkit/bdk/issues/1543#issuecomment-2287912658

crates/chain/src/tx_graph.rs
```diff
@@ -715,35 +715,29 @@ impl<A: Clone + Ord> TxGraph<A> {
...
+                None => {
+                    changeset.txs.insert(tx);
                 }
-                (None, TxNodeInternal::Partial(update_txos)) => {
-                    changeset.txouts.extend(
-                        update_txos
-                            .iter()
-                            .map(|(&vout, txo)| (OutPoint::new(txid, vout), txo.clone())),
-                    );
+                Some((TxNodeInternal::Whole(_), _)) => {}
```
evanlinjin: Might be worth to do a `debug_assert_eq!` if for some strange reason, the `tx` does not match up (between the update's tx and the graph's tx).

crates/chain/src/tx_graph.rs
```diff
@@ -761,6 +749,42 @@ impl<A: Clone + Ord> TxGraph<A> {
     }
 }
 
+/// An update to [`TxGraph`].
+pub struct Update<A> {
+    whole_txs: HashMap<Txid, Arc<Transaction>>,
+    partial_txs: HashMap<OutPoint, TxOut>,
+    last_seen: HashMap<Txid, u64>,
+    anchors: BTreeSet<(A, Txid)>,
+}
```
evanlinjin: Remember that we will be moving `Update` into a separate crate `bdk_core` so we won't have access to private members.

Additionally, we don't want to provide floating outputs to our update logic when we already have the full tx in the same update struct.

**With these two requirements in mind, let's see various solutions.**

* Publicize all members. This is a bad idea, since we can't enforce our invariances.
* Enforce invariance when caller inserts data. When inserting full tx, rm all associated floating txouts of the tx's txid. Don't insert floating txout when there is a full tx of same txid.
* Enforce invariance when extracting data. When extracting full txs, also rm floating txouts with the same txid. When extracting floating txouts, ignore elements which have an associated full tx.

I think enforcing invariance when caller inserts data is easier to reason with.

notmandatory:
This is replaced by #1568 right?

evanlinjin:
Replaced by #1568

PR #1559, opened 2024-08-15
docs(wallet): add example usage of descriptor and plan

futurechimp:
### Description

Example usage of a vault policy, descriptor, and spend plan with a BDK wallet. 

### Changelog notice

docs(wallet): add example usage of descriptor and plan

#### All Submissions:

#### New Features:

* [x] I've added docs for the new feature (they are docs)

#### Bugfixes:

### DIFF
--- a/crates/wallet/examples/descriptor_with_plan.rs
+++ b/crates/wallet/examples/descriptor_with_plan.rs
@@ -1,0 +1,343 @@
+#![allow(unused)]
+use bdk_chain::{BlockId, CheckPoint};
+use bdk_wallet::bitcoin::bip32::Xpriv;
+use bdk_wallet::bitcoin::hashes::Hash;
+use bdk_wallet::bitcoin::key::Secp256k1;
+use std::str::FromStr;
+
+use bdk_wallet::bitcoin::{
+    self, psbt, Address, Network, OutPoint, Psbt, Script, Sequence, Transaction, TxIn, TxOut, Txid,
+};
+
+use bdk_wallet::keys::DescriptorPublicKey;
+use bdk_wallet::miniscript::plan::Assets;
+use bdk_wallet::miniscript::policy::Concrete;
+use bdk_wallet::miniscript::psbt::PsbtExt;
+use bdk_wallet::miniscript::{DefiniteDescriptorKey, Descriptor};
+use bdk_wallet::{KeychainKind, Wallet};
+use bitcoin::{absolute, transaction, Amount};
+
+// Using a descriptor and spending plans with BDK wallets.
+//
+// Consider the basic flow of using a descriptor. The steps are:
+// 1. Set up the Descriptor
+// 2. Deposit sats into the descriptor
+// 3. Get the previous output and witness from the deposit transaction
+// 4. Set up a psbt to spend the deposited funds
+// 5. If there are multiple spend paths, use the `plan` module to format the psbt properly
+// 6. Sign the spending psbt
+// 7. Finalize the psbt. At this point, miniscript will check whether the transaction
+//   satisfies the descriptor, and will notify you if it doesn't.
+// 8. If desired, extract the transaction from the psbt and broadcast it.
+#[allow(clippy::print_stdout)]
+fn main() {
+    // In order to try out descriptors, let's define a Bitcoin vault with two spend paths.
+    //
+    // The vault works like this:
+    //
+    // A. If you have the `unvault_key`, you can spend the funds, but only *after* a specified block height
+    // B. If you have the `emergency_key`, presumably kept in deep cold storage, you can spend at any time.
+
+    // Let's set up some wallets so we have keys to work with.
+
+    // Regular unvault spend path keys + blockheight. You can use wallet descriptors like this,
+    // or you could potentially use a mnemonic and derive from that. See the `mnemonic_to_descriptors.rs`
+    // example if you want to do that.
+    let unvault_tprv = "tprv8ZgxMBicQKsPdKyH699thnjrFcmJMrUUoaNZvHYxxqvhySPhAYZpmxtR39u5QAYnhtYSfMBuBBH6pGuSgmoK3NpfNDU3RAbrVpcbpLmz5ot";
+    let unvault_pk = "02e7c62fd3a65abdc7ff233fba5637f89c9eaba7fe6baaf15ca99d81e0f5145bf8";
+    let after = 1311208;
+
+    // Emergency path keys
+    let emergency_tprv = "tprv8ZgxMBicQKsPekKEvzvCnK7qe5r6ausugHDyrPeX9TLQ4oADSYLWtA4m3XsEMmUZEbVaeJtuZimakomLkecLTMwerVJKpAZFtXoo7DYb84B";
+    let emergency_pk = "033b4ac89f5d83de29af72d8b99963c4dbd416fa7c8a8aee6b4761f8f85e588f80";
+
+    // Make a wallet for the unvault user
+    let unvault_desc = format!("wpkh({unvault_tprv}/84'/1'/0'/0/*)");
+    let unvault_change_desc = format!("wpkh({unvault_tprv}/84'/1'/0'/1/*)");
+    let mut unvault_wallet = Wallet::create(unvault_desc, unvault_change_desc)
+        .network(Network::Testnet)
+        .create_wallet_no_persist()
+        .expect("couldn't create unvault_wallet");
+
+    // Make a wallet for the emergency user
+    let emergency_desc = format!("wpkh({emergency_tprv}/84'/1'/0'/0/*)");
+    let emergency_change_desc = format!("wpkh({emergency_tprv}/84'/1'/0'/1/*)");
+    let mut emergency_wallet = Wallet::create(emergency_desc, emergency_change_desc)
+        .network(Network::Testnet)
+        .create_wallet_no_persist()
+        .expect("couldn't create emergency_wallet");
+
+    // 1. Set up the Descriptor
+
+    // The following string defines a miniscript vault policy with two possible spend paths (`or`):
+    // * spend at any time with the `emergency_pk`
+    // * spend `after` the timelock with the `unvault_pk`
+    let policy_str = format!("or(pk({emergency_pk}),and(pk({unvault_pk}),after({after})))");
+
+    // Refer to `examples/compiler.rs` for compiling a miniscript descriptor from a policy string.
+    let desc_str = "wsh(or_d(pk(033b4ac89f5d83de29af72d8b99963c4dbd416fa7c8a8aee6b4761f8f85e588f80),and_v(v:pk(02e7c62fd3a65abdc7ff233fba5637f89c9eaba7fe6baaf15ca99d81e0f5145bf8),after(1311208))))#9xvht4sc";
+    println!("The vault descriptor is: {}\n", desc_str);
+
+    // Alternately, we can make a wallet for our vault and get its address:
+    let mut vault = Wallet::create_single(desc_str)
+        .network(Network::Testnet)
+        .create_wallet_no_persist()
+        .unwrap();
+    let vault_address = vault.peek_address(KeychainKind::External, 0).address;
+    println!("The vault address is {:?}", vault_address);
+    let vault_descriptor = vault.public_descriptor(KeychainKind::External).clone();
+    let definite_descriptor = vault_descriptor.at_derivation_index(0).unwrap();
+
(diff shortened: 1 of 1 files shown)


oleonardolima (commented):
Thank you for working on this! It looks good and it's a great addition.

I guess you could rename the commits to the standard [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary), to something like: `docs(wallet): add example usage of descriptor and plan` and maybe squash the clippy fixes into a single commit.

futurechimp:
All done, thanks for the suggestions, and if there's anything else needed just let me know.

crates/wallet/examples/descriptor_with_plan.rs
```diff
@@ -0,0 +1,479 @@
...
+    // A. If you have the `unvault_key`, you can spend the funds, but only *after* a specified block height
+    // B. If you have the `emergency_key`, presumably kept in deep cold storage, you can spend at any time.
+
+    // Let's set up some wallets so we have keys to work with.
+
+    // Regular unvault spend path keys + blockheight. You can use wallet descriptors like this,
+    // or you could potentially use a mnemonic and derive from that. See the `mnemonic_to_descriptor.rs`
+    // example if you want to do that.
+    let unvault_tprv = "tprv8ZgxMBicQKsPdKyH699thnjrFcmJMrUUoaNZvHYxxqvhySPhAYZpmxtR39u5QAYnhtYSfMBuBBH6pGuSgmoK3NpfNDU3RAbrVpcbpLmz5ot";
+    let unvault_pk = "02e7c62fd3a65abdc7ff233fba5637f89c9eaba7fe6baaf15ca99d81e0f5145bf8";
```
ValuedMammal, line 47: I'm curious where does this `unvault_pk` come from?

ValuedMammal: Ok in terms of rust it's just

```rust
Xpriv::from_str(unvault_tprv)
    .unwrap()
    .private_key
    .public_key(&secp)
```

futurechimp: Yep, that's right.

crates/wallet/examples/descriptor_with_plan.rs
```diff
@@ -0,0 +1,479 @@
...
+
+    // 8. If desired, extract the transaction from the psbt and broadcast it. We won't do this
+    // here as it saves messing around with faucets, wallets, etc.
+    let _my_emergency_spend_tx = psbt.extract_tx().expect("failed to extract emergency tx");
+
+    // Let's now try the same thing with the unvault transaction. We just need to make a new
+    // plan, sign a new spending psbt, and finalize it.
+
+    // Build a spend transaction the unvault key path
+    let timelock = LockTime::from_height(after).expect("couldn't format locktime");
```
ValuedMammal: nit: For clarity you can qualify it as `absolute::LockTime` but I see you imported the right type so this is fine too.

crates/wallet/examples/descriptor_with_plan.rs
```diff
@@ -0,0 +1,479 @@
...
+fn blank_transaction_with(lock_time: LockTime) -> bitcoin::Transaction {
+    bitcoin::Transaction {
+        version: transaction::Version::TWO,
+        lock_time,
+        input: vec![],
+        output: vec![],
+    }
+}
+
+fn psbt_with_unsigned_tx(tx: bitcoin::Transaction) -> Psbt {
```
ValuedMammal: There's a method for this [`Psbt::from_unsigned_tx`](https://docs.rs/bitcoin/latest/bitcoin/psbt/struct.Psbt.html#method.from_unsigned_tx) - with a slight difference that instead of pushing onto `psbt.inputs` you would just assign to `psbt.inputs[0]`.

crates/wallet/examples/descriptor_with_plan.rs
```diff
@@ -0,0 +1,479 @@
...
+
+    // Build a spend transaction the unvault key path
+    let timelock = LockTime::from_height(after).expect("couldn't format locktime");
+    let unvault_spend_transaction = blank_transaction_with(timelock);
+    let mut psbt = psbt_with_unsigned_tx(unvault_spend_transaction);
+
+    // Format an input containing the previous output (we got that already using `get_vout()`)
+    let txin = TxIn {
+        previous_output,
+        sequence: Sequence::ZERO, // you MUST set this if you have a timelock
```
ValuedMammal: This should also work?

```suggestion
        sequence: Sequence::ENABLE_RBF_NO_LOCKTIME, // disables relative timelock 
```

ValuedMammal (commented):
Nice work on this. My initial thoughts are that we should try to narrow the scope to a few concepts including the descriptor, plan, and psbt.

It would be nice if the vault was a bdk `Wallet` which we can simulate funding by just inserting a transaction into it (no need for `get_funded_wallet` or `TxBuilder`, and you can just hardcode a receive address for spending out of the vault).

We can use the [`Psbt::sign`](https://docs.rs/bitcoin/latest/bitcoin/psbt/struct.Psbt.html#method.sign) API and just pass in the xpriv instead of using `sign_ecdsa`.

Maybe add more `println`s and assertions along the way if you think that would help. Another future enhancement would be to use the wallet to create a psbt, but I think more needs to be done in the library to fully utilize the `Plan` construct. Having this example helps demonstrate how that could be implemented.

futurechimp:
Ok great, happy to get rid of the ecdsa signing (I did it this way only because the only code example I could see that seemed as if it'd work came from the rust-miniscript repository). I'll try using `Psbt::sign()` although I haven't run across any examples of how to set that up. 

And I will switch to using a `Wallet` for the descriptor and try to simplify the usage of it to reduce code. And add some `println!`s. 

Do the transaction inputs, outputs, and witness appear to be set up correctly? 

I'll keep working on it in a series of commits in my fork, and we can squash it down at the end once the example looks good. This is my first attempt with bdk so I'm very happy to receive feedback about how to structure things.

crates/wallet/examples/descriptor_with_plan.rs
futurechimp: Will do.

crates/wallet/examples/descriptor_with_plan.rs
futurechimp: I've added this locally

crates/wallet/examples/descriptor_with_plan.rs
futurechimp: I have switched to `Psbt::from_unsigned_tx`, thanks. Just to note,`psbt.inputs.push(foo)` still appears to be there.

futurechimp:
Ok, updated with the following: 

1. turned the vault descriptor in a `Wallet`
2. used `Psbt::sign` instead of Ecdsa signing
3. added printlns to show the general flow. I didn't print out the internals of many things as the output quickly gets pretty big - happy to show more things if helpful
4. removed the `get_funded_wallet` function and just inserted the deposit transaction into the `vault` wallet directly
5. prefixed with type path `absolute::LockTime` so that types are more clear
6. used `Sequence::ENABLE_RBF_NO_LOCKTIME` instead of `Sequence::ZERO`.
7. used `Psbt::from_unsigned_tx` instead of rolling my own

ValuedMammal:
I think it's looking good. I put some ideas for extra improvements in this commit ValuedMammal/bdk@60601f1ee6ceb1ba981f0e8cf589d574bcba49cd that include

- refactor `deposit_transaction` and print the vault balance    
- refer to compiler example (instead of duplicating it)
- get the vault descriptor from the `Wallet` structure

futurechimp:
Looking quite a bit nicer now. Is it possible for you to squash merge it into the main bdk repo when ready? I've tied myself in a few git knots in my branch/repo, missed squashing your PR to my PR branch.

futurechimp:
Ok, all squashed in place now, sorry for the noise.

crates/wallet/examples/descriptor_with_plan.rs
```diff
@@ -0,0 +1,320 @@
...
+        ..Default::default()
+    };
+
+    // Format an output which spends some of the funds in the vault
+    let txout = TxOut {
+        script_pubkey: emergency_wallet
+            .next_unused_address(KeychainKind::External)
+            .script_pubkey(),
+        value: Amount::from_sat(750),
+    };
```
ValuedMammal: I realized we're not sending any change back to the vault, so in this case the majority of funds would be lost to fees. If we want to keep it to 1-input 1-output for simplicity we could just send all the original funds (`76_000`) minus some reasonable amount for fees.

futurechimp: Good point. I've added a 500 sat fee subtraction and sent the rest of the change back to the `emergency_wallet`'s change address.

futurechimp:
It looks like the API for `wallet.apply_update()` has been changed in the past week, so this PR fails CI: https://github.com/bitcoindevkit/bdk/actions/runs/10577332007/job/29489915320?pr=1559

The old API took a `graph` in an Update struct: 

```rust
pub struct Update {
    pub last_active_indices: BTreeMap<KeychainKind, u32>,
    pub graph: TxGraph<ConfirmationBlockTime>,
    pub chain: Option<CheckPoint>,
}
```

The new API is basically the same but has  a `tx_update` in place of the `graph`:

```rust
pub struct Update {
    pub last_active_indices: BTreeMap<KeychainKind, u32>,
    pub tx_update: TxUpdate<ConfirmationBlockTime>,
    pub chain: Option<CheckPoint>,
}
```

I can't see how to get access to `TxUpdate` though, because that's now in the `core` crate, not the `wallet` crate, and I not seeing an import path which allows me to import the type. 

Am I missing an easy way to get access to bdk_core from the wallet? 

Or is there a way to easily change this for a fake wallet funding transaction so that it doesn't depend on `graph`/`tx_update`? 

```rust
fn deposit_transaction(wallet: &mut Wallet, tx: Transaction) -> OutPoint {
    use bdk_chain::{ConfirmationBlockTime, TxGraph};
    use bdk_wallet::Update;

    let txid = tx.compute_txid();
    let vout = 0;
    let mut graph = TxGraph::<ConfirmationBlockTime>::new([tx]);
    let _ = graph.insert_seen_at(txid, 42);
    wallet
        .apply_update(Update {
            graph,
            ..Default::default()
        })
        .unwrap();

    OutPoint { txid, vout }
}

```

ValuedMammal:
You could try rebasing on current master to pick up the `bdk_core` changes and then import `TxUpdate` via `bdk_chain` crate, so `use bdk_chain::TxUpdate` then call `into` on a `TxGraph`. 

Note after #1425 you might also need to add `#![allow(clippy::print_stdout)]` to the file `descriptor_with_plan.rs`

futurechimp:
I've synced my branch with the bdk master and fixed compilation on top of that. Should be ready for review now.

tvpeter:
Awesome job! Super helpful example for using descriptors and vault policies. 

NIT: As was earlier noted https://github.com/bitcoindevkit/bdk/pull/1559#discussion_r1731401026, the unvault spend path sent majority of funds to miner.

Test output:
--------------
Screenshot 2024-12-19 at 10 28 06 PM

luisschwab:
Hey @futurechimp, this PR needs to be closed and moved to the new [bdk_wallet](https://github.com/bitcoindevkit/bdk_wallet) repo. Here's how:

1. From your Github account, fork https://github.com/bitcoindevkit/bdk_wallet
2. From your local system `bdk` repo folder, push your PR changes to your `bdk_wallet` fork on Github:
   ```
   git checkout <PR branch path>
   git push git@github.com:<username>/bdk_wallet.git
   ```
3. Clone to your forked `bdk_wallet` repo to your local system and switch to your PR branch:
   ```
   git clone git@github.com:<username>/bdk_wallet.git
   cd bdk_wallet
   git fetch origin <PR branch path>
   git checkout -b <PR branch path> FETCH_HEAD
   git rebase master
   ```
4. Resolve rebase conflicts (if any)
5. Force push your PR branch: `git push origin -f`
6. Create new PR in [bitcoindevkit/bdk_wallet](https://github.com/bitcoindevkit/bdk_wallet/pulls) repo from your fork branch
7. Copy/paste title and description from your original `bdk` PR and add a link back to it in your new PR.
8. Close original PR with comment that includes link to new PR.

ValuedMammal:
Closing as it is now stale. Also it's not clear whether a descriptor-and-plan example is truly needed since the impact of the planning module is mostly internal (https://github.com/bitcoindevkit/bdk_wallet/issues/64#issuecomment-2784704405).

PR #1564, opened 2024-08-21
ci: add cron-update-rust.yml

### LINKED ISSUE #1431: Create nightly job to bump clippy rust version

ValuedMammal:
This adds a cronjob to watch for updates to rust stable once a month. If the latest version changed, then a pull request is opened to see if clippy passes CI using the new toolchain.

### Notes to the reviewers

Let me know what you think @notmandatory 

fixes #1431

#### TODO

* [ ] Setup access token for [create-pull-request](https://github.com/marketplace/actions/create-pull-request)

#### All Submissions:


.github/workflows/cron-update-rust.yml
```diff
@@ -0,0 +1,41 @@
...
+    - cron: "0 0 15 * *" # At 00:00 on day-of-month 15.
+  workflow_dispatch: # allows manual triggering
+jobs:
+  format:
+    name: Update rustc
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Update rust.yml to use latest stable
```
notmandatory: Should this be `rust-version` instead of `rust.yml`, on line 39 also? 
```suggestion
      - name: Update rust-version to use latest stable
```

notmandatory (commented):
Looks good to me. I just have one small comment.

storopoli (approved):
Looks good exactly what we do in `rust-bitcoin`.
You got the idea! Really nice to see what we've discussed briefly in the summit being implemented.

Just need to set the token stuff in the repo/org and then we are good to go to merge it.

.github/workflows/cron-update-rust.yml
ValuedMammal: Yes, thanks. On line 39 I changed it to refer to the actual workflow file `cont_integration.yml`

notmandatory (approved):
ACK e5a97515602832312d2925d7ea1a3c0ff351d05b

@ValuedMammal if you need more permission to create `secrets.MY_TOKEN` let me know.

.github/workflows/cron-update-rust.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: peter-evans/create-pull-request@v6
+        with:
+          # TODO: Allow `MY_TOKEN` to open a pull request
+          # token: ${{ secrets.MY_TOKEN }}
+          author: Update Rustc Bot <bot@example.com>
+          committer: Update Rustc Bot <bot@example.com>
+          title: Automated update to rustc (to ${{ env.rust_version }})
+          body: |
+           Automated update to Github CI workflow `cont_integration.yml` by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action
+          commit-message: Automated update to Github CI to rustc ${{ env.rust_version }}
```
oleonardolima: nit: should we use the convetional commits here, and maybe on the title (e.g as dependabot does) ?

oleonardolima (commented):
ACK e5a97515602832312d2925d7ea1a3c0ff351d05b

.github/workflows/cron-update-rust.yml
notmandatory: Oh good point! how about something like:
```
ci: automated update to rustc ${{ env.rust_version }}
```

ValuedMammal:
> ACK [e5a9751](https://github.com/bitcoindevkit/bdk/commit/e5a97515602832312d2925d7ea1a3c0ff351d05b)
> 
> @ValuedMammal if you need more permission to create `secrets.MY_TOKEN` let me know.

As far as I can tell I'm not able to add secrets to the repo

.github/workflows/cron-update-rust.yml
```diff
@@ -0,0 +1,43 @@
...
+          else
+              echo "Attempted to update rustc but the latest stable date did not change. Not opening any PR."
+              echo "changes_made=false" >> $GITHUB_ENV
+          fi
+      - name: Create Pull Request
+        if: env.changes_made == 'true'
+        uses: peter-evans/create-pull-request@v6
+        with:
+          # TODO: Allow `MY_TOKEN` to open a pull request
+          # token: ${{ secrets.MY_TOKEN }}
```
notmandatory: I think we don't need a custom token and can use [GITHUB_TOKEN](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) instead.

```suggestion
          # token: ${{ secrets.GITHUB_TOKEN }}
```

ValuedMammal: Ok changed to `GITHUB_TOKEN`. As far as permissions I guess it needs to trigger other workflows in addition to creating a PR. We should be able to manually run the cron update to check that it's working

oleonardolima (approved):
tACK b92f8c9ac1e000ccced89dc597a8ff278842d5ed

I don't think you'll need any further `secrets.GITHUB_TOKEN` changes, I tested it in my forked repository and it worked just fine.

I [manually triggered the workflow](https://github.com/oleonardolima/bdk/actions/runs/10525851656) and it [automatically created a new PR](https://github.com/oleonardolima/bdk/pull/1) as expected.

PR #1566, opened 2024-08-21
feat(chain)!: Add `time_of_sync` to `SyncRequest` and `FullScanRequest` WIP

LagginTimes:
Implements #1550.

### Description

Adds `time_of_sync` as a member of `SyncRequest` and `FullSyncRequest` so we can set the `last_seen` for unconfirmed transactions during the sync/scan process in `bdk_electrum`.

### Notes to the reviewers

This PR is a WIP because setting the `last_seen` for unconfirmed transactions during the sync/scan process is yet to be implemented for `bdk_esplora`. I pushed my WIP towards setting the `last_seen` for `bdk_esplora` but I'm unsure if it is correct, as I did not have enough time to confirm and write a test for it before my travel. Please feel free to drop that commit if the concept is incorrect.

### Changelog notice

* Removed `update_last_seen_unconfirmed()` from `TxGraph`.
* Added `time_of_sync` as a member of `SyncRequest`.
* Added `time_of_sync` as a member of `FullScanRequest`.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/spk_client.rs
+++ b/crates/chain/src/spk_client.rs
@@ -193,6 +193,12 @@ impl<I> SyncRequestBuilder<I> {
         self
     }
 
+    /// Set the `time_of_sync` for unconfirmed transactions.
+    pub fn time_of_sync(mut self, time_of_sync: u64) -> Self {
+        self.inner.time_of_sync = Some(time_of_sync);
+        self
+    }
+
     /// Set the closure that will inspect every sync item visited.
     pub fn inspect<F>(mut self, inspect: F) -> Self
     where
@@ -239,6 +245,7 @@ pub struct SyncRequest<I = ()> {
     txids_consumed: usize,
     outpoints: VecDeque<OutPoint>,
     outpoints_consumed: usize,
+    time_of_sync: Option<u64>,
     inspect: Box<InspectSync<I>>,
 }
 
@@ -252,6 +259,7 @@ impl<I> Default for SyncRequest<I> {
             txids_consumed: 0,
             outpoints: VecDeque::new(),
             outpoints_consumed: 0,
+            time_of_sync: None,
             inspect: Box::new(|_, _| {}),
         }
     }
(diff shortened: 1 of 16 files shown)


crates/chain/src/spk_client.rs
```diff
@@ -193,6 +193,12 @@ impl<I> SyncRequestBuilder<I> {
         self
     }
 
+    /// Set the `time_of_sync` for unconfirmed transactions.
+    pub fn time_of_sync(mut self, time_of_sync: u64) -> Self {
```
notmandatory, line 197: Is this value required ? If so should it be a parameter of `SyncRequest::builder()`?  Otherwise if someone forgets to set it it will always be 0. (And same for FullScanRequest::builder()).  

Or if it's not required as you have it is fine.

LagginTimes: At first I thought to make it user-configurable, but I think you're right, it makes significantly more sense to just have the `time_of_sync` set by default.

But we would run into the `no_std` issue we we were to set it from `SyncRequest`/`FullScanRequest`.

crates/chain/src/spk_client.rs
```diff
@@ -239,6 +239,7 @@ pub struct SyncRequest<I = ()> {
     txids_consumed: usize,
     outpoints: VecDeque<OutPoint>,
     outpoints_consumed: usize,
+    time_of_sync: u64,
```
evanlinjin:
```suggestion
    time_of_sync: Option<u64>,
```

I think it's important that the caller can opt out of setting a seen_at for unconfirmed transactions.

Additionally, I don't think having a default of 0 is a good idea... However, I need to think more on this.

crates/chain/src/tx_graph.rs
```diff
@@ -565,76 +565,13 @@ impl<A: Clone + Ord> TxGraph<A> {
...
-                },
-            )
-            .collect();
-
-        for txid in unanchored_txs {
-            changeset.merge(self.insert_seen_at(txid, seen_at));
-        }
-        changeset
-    }
-
```
evanlinjin, line 637: I think removing this method should really be a separate commit. However, I'm wondering if there is an argument to keep it.

crates/chain/src/spk_client.rs
notmandatory: Now that you have internal `time_of_sync` field as an `Option<u64>` it makes sense to leave this function as-is.

crates/chain/src/tx_graph.rs
notmandatory: My 2 sats is if both wallet and non-wallet based app will use `SyncRequest` and `FullScanRequest` then this shouldn't be needed.

PR #1568, opened 2024-08-23
Introduce `tx_graph::Update` and simplify `TxGraph` update logic

### LINKED ISSUE #1543: Make chain-sources non-dependent on a specific `bdk_chain` version.

### LINKED ISSUE #1550: Should `bdk_electrum` `sync`/`full_scan` set `last_seen`?

evanlinjin:
Part of #1543
Closes #1550

### Description

Instead of updating a `TxGraph` with another `TxGraph` in `.apply_update()`, we introduce `tx_graph::Update`. `tx_graph::Update` is a simple data object. This is the first step of #1543. This also makes it slightly less expensive to create an update.

Additionally, we simplify the update logic of `TxGraph` by containing most of the update logic in `.insert_{}`-esc methods. `.apply_update` and `.apply_changeset` will call `.insert_{}` methods internally and is greatly simplified. Thus, we also get rid of `.determine_changeset`.

We change `.apply_update` methods of `TxGraph`,`IndexedTxGraph` and `Wallet` to implicitly set `seen_at` to the current timestamp for unconfirmed update transactions. This makes these methods only available when the `"std"` cargo feature is enabled. For a non-std environment, we introduce `.apply_update_at` (as shown below).

```rust
pub fn apply_update_at(&mut self, update: Update<A>, seen_at: Option<u64>) -> ChangeSet<A> {
    todo!()
}
```

During this process, I've fixed some tests that didn't make much sense.

### Notes to the reviewers

There is a slight scope-creep, but they all address the update logic of `TxGraph` and structures that build on top.

### Changelog notice

* Changed update API (`.apply_update`) of `TxGraph` to take in a simple data object (`tx_graph::Update`).
* Changed `.apply_update` methods of `TxGraph`, `IndexedTxGraph` and `Wallet` to implicitly set the `seen_at` to the current timestamp for unconfirmed update transactions (this making them depend on the "std" cargo feature).
* Add `.apply_update_at` which is the no-std version of `.apply_update`.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -146,29 +146,12 @@ impl<A> From<TxGraph<A>> for Update<A> {
 impl<A: Ord + Clone> From<Update<A>> for TxGraph<A> {
     fn from(update: Update<A>) -> Self {
         let mut graph = TxGraph::<A>::default();
-        let _ = graph.apply_update(update);
+        let _ = graph.apply_update_at(update, None);
         graph
     }
 }
 
 impl<A: Ord> Update<A> {
-    /// Update the [`seen_ats`](Self::seen_ats) for all unanchored transactions.
-    pub fn update_last_seen_unconfirmed(&mut self, seen_at: u64) {
-        let seen_ats = &mut self.seen_ats;
-        let anchors = &self.anchors;
-        let unanchored_txids = self.txs.iter().map(|tx| tx.compute_txid()).filter(|txid| {
-            for (_, anchor_txid) in anchors {
-                if txid == anchor_txid {
-                    return false;
-                }
-            }
-            true
-        });
-        for txid in unanchored_txids {
-            seen_ats.insert(txid, seen_at);
-        }
-    }
-
     /// Extend this update with `other`.
     pub fn extend(&mut self, other: Update<A>) {
         self.txs.extend(other.txs);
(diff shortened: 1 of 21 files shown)


crates/chain/src/tx_graph.rs
```diff
@@ -640,125 +691,70 @@ impl<A: Clone + Ord> TxGraph<A> {
     ///
     /// The returned [`ChangeSet`] is the set difference between `update` and `self` (transactions that
     /// exist in `update` but not in `self`).
-    pub fn apply_update(&mut self, update: TxGraph<A>) -> ChangeSet<A> {
-        let changeset = self.determine_changeset(update);
-        self.apply_changeset(changeset.clone());
+    #[cfg(feature = "std")]
```
LLFourn, line 689:
```suggestion
    #[cfg(feature = "std")]
    #[cfg_attr(docsrs, doc(cfg(feature = "std")))]
```

Explain what's going on here internally.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -385,7 +310,7 @@ fn insert_txout_does_not_displace_tx() {
     let _ = tx_graph.insert_txout(
         OutPoint {
             txid: tx.compute_txid(),
-            vout: 0,
+            vout: 1,
```
LLFourn, line 313: looks like a mistaken change.

crates/esplora/src/blocking_ext.rs
```diff
@@ -339,59 +354,48 @@ fn fetch_txs_with_spks<I: IntoIterator<Item = ScriptBuf>>(
...
         let handles = txids
             .by_ref()
             .take(parallel_requests)
             .map(|txid| {
                 let client = client.clone();
-                let tx_already_exists = tx_graph.get_tx(txid).is_some();
                 std::thread::spawn(move || {
-                    if tx_already_exists {
-                        client
-                            .get_tx_status(&txid)
```
LLFourn, line 362: It looks like we were calling `get_tx_status` if we already had it but this seems like a bug so you removed it?

crates/chain/tests/test_tx_graph.rs
```diff
@@ -88,57 +88,21 @@ fn insert_txouts() {
...
+        let mut update = tx_graph::Update::default();
         for (outpoint, txout) in &update_ops {
-            // Insert partials transactions
-            assert_eq!(
-                graph.insert_txout(*outpoint, txout.clone()),
-                ChangeSet {
-                    txouts: [(*outpoint, txout.clone())].into(),
-                    ..Default::default()
-                }
-            );
```
LLFourn, line 100: note sure if it's worth it but you could keep all the code here the same and then just turn the `TxGraph` into an `Update` at the end. Might as well to test the `From` impl but up to you.

crates/esplora/src/blocking_ext.rs
evanlinjin: Yes. I wouldn't call it a bug. It's just that esplora always returns a tx with a status. If we already have a tx, that means we already have the status. No need to getting the status here again

crates/chain/tests/test_tx_graph.rs
LLFourn: oh no it's right.

crates/chain/tests/test_tx_graph.rs
evanlinjin: I think it's better to add separate tests for those!

evanlinjin: Tests added.

notmandatory (approved):
ACK ccb8c796b259ae348cb8b8ef35b8c60e5621ccb5

Looks good and I can see how this will help decouple the blockchain clients from the wallet once `tx_graph:Update` is moved to a new crate. 

I ran all the examples also as a little smoke test and they worked fine.

LLFourn (approved):
ACK ccb8c796b259ae348cb8b8ef35b8c60e5621ccb5

PR #1569, opened 2024-08-23
Introduce `bdk_core`

### LINKED ISSUE #1543: Make chain-sources non-dependent on a specific `bdk_chain` version.

evanlinjin:
Based on #1568
Closes #1543

### Description

Introduce `bdk_core` crate. Move types over from `bdk_chain`. Chain sources (`bdk_electrum`, `bdk_esplora` and `bdk_bitcoind_rpc`) now only depend on `bdk_core`.

### Notes to the reviewers

Please review commit-by-commit. I've moved things over, but slight API changes were necessary (mentioned in the commit messages).

### Changelog notice

* Add `bdk_core` crate which contains core types that were previously in `bdk_chain`. Including: `BlockId`, `ConfirmationBlockTime`, `CheckPoint`, `CheckPointIter`, `tx_graph::Update` and `spk_client`-types.
* Change chain sources (`bdk_esplora`, `bdk_electrum` and `bdk_bitcoind_rpc`) to only depend on `bdk_core`.

#### All Submissions:


notmandatory:
Please keep track if any types or functions were modified other than just because things were moved around.  Even though there are a lot of changes here if no types or logic were changed it shouldn't be too hard to review.

crates/core/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+
+mod checkpoint;
+pub use checkpoint::*;
+
+pub mod spk_client;
+
+/// Core structures for [`TxGraph`].
+///
+/// [`TxGraph`]: https://docs.rs/bdk_chain/latest/bdk_chain/tx_graph/struct.TxGraph.html
+pub mod tx_graph {
```
notmandatory: Is there a reason to put `tx_graph` here instead of it's own file?

evanlinjin: No specific reason. It's just for convenience and there is not much to the `tx_graph::Update` declaration anyway. Is there any downsides to this arrangement?

notmandatory: No particular downside I can see, just didn't know if there was some advantage I wasn't aware of.

notmandatory (approved):
ACK dafb9aaef7d8d4bc83c8ba0845323b92b82fbb3d

Great to see this PR only moves code around and decouple the chain clients from the chain module; plus some small docs and helper function cleanup. Your commits were concise and well documented which helped a lot with review. 

I reran examples as a simple smoke test with no problems.

crates/core/src/lib.rs
LLFourn: We shouldn't mention `TxGraph` here too much. Now this thing lives here the `Update` can be used by any other software to process transaction updates. I'll push a commit.

crates/core/src/spk_client.rs
```diff
@@ -345,9 +323,11 @@ impl<I> SyncRequest<I> {
...
 #[derive(Debug)]
 pub struct SyncResult<A = ConfirmationBlockTime> {
-    /// The update to apply to the receiving [`TxGraph`].
-    pub graph_update: TxGraph<A>,
-    /// The update to apply to the receiving [`LocalChain`](crate::local_chain::LocalChain).
+    /// The update to apply to the receiving
+    /// [`TxGraph`](../../bdk_chain/tx_graph/struct.TxGraph.html).
+    pub graph_update: crate::tx_graph::Update<A>,
+    /// The update to apply to the receiving
+    /// [`LocalChain`](../../bdk_chain/local_chain/struct.LocalChain.html).
```
LLFourn: Heh I guess this link is one way to do it but I think it's a mistake to mention `LocalChain` and `TxGraph` here. We are decoupling these things. What if we make a new chain type wouldn't a CheckPoint also be a valid update for it? (I'm pushing a commit).

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -189,9 +185,10 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
     /// may include scripts that have been used, use [`full_scan`] with the keychain.
     ///
     /// [`full_scan`]: Self::full_scan
-    /// [`CalculateFeeError::MissingTxOut`]: bdk_chain::tx_graph::CalculateFeeError::MissingTxOut
-    /// [`Wallet.calculate_fee`]: https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.Wallet.html#method.calculate_fee
-    /// [`Wallet.calculate_fee_rate`]: https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.Wallet.html#method.calculate_fee_rate
+    /// [`bdk_chain`]: ../bdk_chain/index.html
+    /// [`CalculateFeeError::MissingTxOut`]: ../bdk_chain/tx_graph/enum.CalculateFeeError.html#variant.MissingTxOut
+    /// [`Wallet.calculate_fee`]: ../bdk_wallet/struct.Wallet.html#method.calculate_fee
+    /// [`Wallet.calculate_fee_rate`]: ../bdk_wallet/struct.Wallet.html#method.calculate_fee_rate
```
LLFourn, line 191: :pick: I'm against these kinds of doc links because they go stale without us noticing. Also they potentially link to incompatible versions. What if we change that method name etc?

It's enough to explain the concept.

Here are the remaining ones after a5d076f215cd91173f55bda0d1cc59b9dde75511:

```
crates/core/src/spk_client.rs:113:    /// This is used to update [`LocalChain`](../../bdk_chain/local_chain/struct.LocalChain.html).
crates/core/src/spk_client.rs:124:    /// [`KeychainTxOutIndex`](../../bdk_chain/indexer/keychain_txout/struct.KeychainTxOutIndex.html).
crates/core/src/spk_client.rs:358:    /// This is used to update [`LocalChain`](../../bdk_chain/local_chain/struct.LocalChain.html).
crates/electrum/src/bdk_electrum_client.rs:119:    /// [`bdk_chain`]: ../bdk_chain/index.html
crates/electrum/src/bdk_electrum_client.rs:120:    /// [`CalculateFeeError::MissingTxOut`]: ../bdk_chain/tx_graph/enum.CalculateFeeError.html#variant.MissingTxOut
crates/electrum/src/bdk_electrum_client.rs:121:    /// [`Wallet.calculate_fee`]: ../bdk_wallet/struct.Wallet.html#method.calculate_fee
crates/electrum/src/bdk_electrum_client.rs:122:    /// [`Wallet.calculate_fee_rate`]: ../bdk_wallet/struct.Wallet.html#method.calculate_fee_rate
crates/electrum/src/bdk_electrum_client.rs:188:    /// [`bdk_chain`]: ../bdk_chain/index.html
crates/electrum/src/bdk_electrum_client.rs:189:    /// [`CalculateFeeError::MissingTxOut`]: ../bdk_chain/tx_graph/enum.CalculateFeeError.html#variant.MissingTxOut
crates/electrum/src/bdk_electrum_client.rs:190:    /// [`Wallet.calculate_fee`]: ../bdk_wallet/struct.Wallet.html#method.calculate_fee
crates/electrum/src/bdk_electrum_client.rs:191:    /// [`Wallet.calculate_fee_rate`]: ../bdk_wallet/struct.Wallet.html#method.calculate_fee_rate
```

LLFourn (approved):
I added a a5d076f215cd91173f55bda0d1cc59b9dde75511: 

> We shouldn't refer to `tx_graph` in `bdk_core`. TxGraph happens to consume `Update` but it doesn't conceptually own the definition of an update to transactions.
> I tried to also remove a lot of references to "graph" where they shouldn't be. "graph" is a very general kind of data structure so we should avoid referring to it where it's not relevant. `tx_update` is much better than `graph_update`.

Self-ACK: a5d076f215cd91173f55bda0d1cc59b9dde75511

evanlinjin (commented):
ACK a5d076f215cd91173f55bda0d1cc59b9dde75511

`TxUpdate` makes sense. The doc changes make things clearer and easier to maintain.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -39,14 +37,11 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
 
     /// Inserts transactions into the transaction cache so that the client will not fetch these
     /// transactions.
-    pub fn populate_tx_cache<A>(&self, tx_graph: impl AsRef<TxGraph<A>>) {
-        let txs = tx_graph
-            .as_ref()
-            .full_txs()
-            .map(|tx_node| (tx_node.txid, tx_node.tx));
-
+    pub fn populate_tx_cache(&self, txs: impl IntoIterator<Item = impl Into<Arc<Transaction>>>) {
```
ValuedMammal, line 40: Ideally this function could accept an iterator of `(Txid, Arc<Transaction>)` tuples

LLFourn: This function is called at startup time so computing txid here is not really worth optimizing out.

PR #1579, opened 2024-08-27
fix(wallet): only mark change address used if `create_tx` succeeds

### LINKED ISSUE #1578: PSBT Creation: Change Address Index Skipped on InsufficientFunds
**Describe the bug**  

When I create a PSBT using TxBuilder, we typically add the recipients, amounts, and other necessary data, and everything works as expected. However, sometimes when we receive an InsufficientFunds exception, the change address index increases, causing it to be skipped in next call to create_psbt, which then selects a new unused one.

In the create_psbt function and noticed that the latest unused index in the KeychainKind::Internal is marked as used first, but when the InsufficientFunds exception occurs, it doesn’t reset back to unused.

please take a look thx

ValuedMammal:
If no drain script is specified in tx params then we get it from the change keychain by looking at the next unused address. Before this PR we marked the index used regardless of whether a change output is finally added. Then if creating a psbt failed, we never restored the unused status of the change address, so creating the next tx would have revealed an extra one.

We want to mark the change address used so that other callers won't attempt to use the same address between the time we create the tx and when it appears on chain. With this PR we only mark the change address used if we successfully create a psbt and the drain script is used in the change output.

fixes bitcoindevkit/bdk#1578 

### Notes to the reviewers

An early idea was to unmark the change address used if we fail to create a tx due to `InsufficientFunds`, but after looking into it I figure it doesn't totally make sense to mark the address used before we've determined that a change output is necessary. Further, `create_tx` can fail in other ways besides running coin selection, so I moved the `mark_used` logic to the end of the function.

### Changelog notice

Fixed an issue that caused an unused internal address to be skipped when creating transactions (#1578)

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -88,7 +88,7 @@ use crate::descriptor::{
 use crate::psbt::PsbtUtils;
 use crate::signer::SignerError;
 use crate::types::*;
-use crate::wallet::coin_selection::Excess::{Change, NoChange};
+use crate::wallet::coin_selection::Excess::{self, Change, NoChange};
 use crate::wallet::error::{BuildFeeBumpError, CreateTxError, MiniscriptPsbtError};
 
 use self::coin_selection::Error;
@@ -1468,17 +1468,28 @@ impl Wallet {
             self.preselect_utxos(&params, Some(current_height.to_consensus_u32()));
 
         // get drain script
+        let mut drain_index = Option::<(KeychainKind, u32)>::None;
         let drain_script = match params.drain_to {
             Some(ref drain_recipient) => drain_recipient.clone(),
             None => {
                 let change_keychain = self.map_keychain(KeychainKind::Internal);
-                let ((index, spk), index_changeset) = self
+                let (index, spk) = self
                     .indexed_graph
                     .index
-                    .next_unused_spk(change_keychain)
-                    .expect("keychain must exist");
-                self.indexed_graph.index.mark_used(change_keychain, index);
-                self.stage.merge(index_changeset.into());
+                    .unused_keychain_spks(change_keychain)
+                    .next()
+                    .unwrap_or_else(|| {
+                        let (next_index, _) = self
+                            .indexed_graph
+                            .index
+                            .next_index(change_keychain)
+                            .expect("keychain must exist");
+                        let spk = self
+                            .peek_address(change_keychain, next_index)
+                            .script_pubkey();
+                        (next_index, spk)
+                    });
+                drain_index = Some((change_keychain, index));
                 spk
             }
         };
@@ -1577,6 +1588,18 @@ impl Wallet {
         params.ordering.sort_tx_with_aux_rand(&mut tx, rng);
 
         let psbt = self.complete_transaction(tx, coin_selection.selected, params)?;
+
+        // recording changes to the change keychain
+        if let (Excess::Change { .. }, Some((keychain, index))) = (excess, drain_index) {
+            let (_, index_changeset) = self
+                .indexed_graph
+                .index
+                .reveal_to_target(keychain, index)
+                .expect("must not be None");
+            self.stage.merge(index_changeset.into());
+            self.mark_used(keychain, index);
+        }
+
         Ok(psbt)
     }
 
--- a/crates/wallet/tests/wallet.rs
+++ b/crates/wallet/tests/wallet.rs
@@ -1393,6 +1393,127 @@ fn test_create_tx_global_xpubs_with_origin() {
 }
 
 #[test]
+fn test_create_tx_increment_change_index() {
+    // Test derivation index and unused index of change keychain when creating a transaction
+    // Cases include wildcard and non-wildcard descriptors with and without an internal keychain
+    // note the test assumes that the first external address is revealed since we're using
+    // `receive_output`
+    struct TestCase {
+        name: &'static str,
+        descriptor: &'static str,
+        change_descriptor: Option<&'static str>,
+        // amount to send
+        to_send: u64,
+        // (derivation index, next unused index) of *change keychain*
+        expect: (Option<u32>, u32),
+    }
+    // total wallet funds
+    let amount = 10_000;
+    let recipient = Address::from_str("bcrt1q3qtze4ys45tgdvguj66zrk4fu6hq3a3v9pfly5")
+        .unwrap()
+        .assume_checked()
+        .script_pubkey();
+    let (desc, change_desc) = get_test_tr_single_sig_xprv_with_change_desc();
+    [
+        TestCase {
+            name: "two wildcard, builder error",
+            descriptor: desc,
+            change_descriptor: Some(change_desc),
+            to_send: amount + 1,
+            // should not use or derive change index
+            expect: (None, 0),
+        },
+        TestCase {
+            name: "two wildcard, create change",
+            descriptor: desc,
+            change_descriptor: Some(change_desc),
+            to_send: 5_000,
+            // should use change index
+            expect: (Some(0), 1),
+        },
+        TestCase {
+            name: "two wildcard, no change",
+            descriptor: desc,
+            change_descriptor: Some(change_desc),
+            to_send: 9_850,
+            // should not use change index
+            expect: (None, 0),
+        },
+        TestCase {
+            name: "one wildcard, create change",
+            descriptor: desc,
+            change_descriptor: None,
+            to_send: 5_000,
+            // should use change index of external keychain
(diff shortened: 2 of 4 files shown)


ValuedMammal:
This might also be related to bitcoindevkit/bdk_wallet#74

crates/wallet/src/wallet/mod.rs
```diff
@@ -1452,12 +1452,11 @@ impl Wallet {
             Some(ref drain_recipient) => drain_recipient.clone(),
             None => {
                 let change_keychain = self.map_keychain(KeychainKind::Internal);
-                let ((index, spk), index_changeset) = self
+                let ((_index, spk), index_changeset) = self
                     .indexed_graph
                     .index
                     .next_unused_spk(change_keychain)
```
evanlinjin: I'm wondering whether it will make sense to introduce a `KeychainTxOutIndex::next_unused_spk_without_revealing`.

```rust
    pub fn next_unused_spk_without_revealing(&self, keychain: K) -> Option<Indexed<ScriptBuf>> {
        let next_unused = self
            .unused_keychain_spks(keychain.clone())
            .next()
            .map(|(i, spk)| (i, spk.to_owned()));
        let (next_index_to_reveal, _) = self.next_index(keychain.clone())?;
        next_unused.or_else(|| {
            self.unbounded_spk_iter(keychain)
                .expect("descriptor must exist")
                .nth(next_index_to_reveal as _)
        })
    }

```

This way, we can avoid potentially-revealing an spk here (which might not be used), and avoid recording the new revealed index in `self.stage`.

We can use a new variable `let mut drain_keychain_index = Option::<(KeychainKind, u32)>::None` to actually do a mutation later on (when we determine that we will have a change output in our tx).

crates/wallet/src/wallet/mod.rs
```diff
@@ -1557,6 +1556,19 @@ impl Wallet {
         params.ordering.sort_tx_with_aux_rand(&mut tx, rng);
 
         let psbt = self.complete_transaction(tx, coin_selection.selected, params)?;
+
+        if psbt
+            .unsigned_tx
+            .output
+            .iter()
+            .any(|txout| txout.script_pubkey == drain_script)
+        {
```
evanlinjin: It will be more *correct* to check the `excess` variable. The caller may set the recipient address as the drain address that we obtained from the index. In this case, it might be unexpected for the caller to have this spk marked as used.

```rust
        if matches!(excess, &coin_selection::Excess::Change { .. }) { }
```

evanlinjin (commented):
Thanks for taking this forward. Let me know what you think about the `next_unused_spk_without_revealing` idea.

crates/wallet/src/wallet/mod.rs
ValuedMammal: Yes that will be better, I think it's resolved by 6afde0b7761113775bdca11ee720777cd178acf7

crates/wallet/src/wallet/mod.rs
ValuedMammal: Thinking through this what I think you're advocating is something like "next unused or else peek next index". Would you then reveal the next index in the event that the change address is used in the tx?

Also if we don't reveal it here are we relying on `index_of_spk` returning a value in order to mark it used? (which should be the case as long as the lookahead > 0)

ValuedMammal: > are we relying on index_of_spk returning a value in order to mark it used?

I think no because we'll just use `drain_keychain_index` instead of `index_of_spk`

ValuedMammal: I pushed 9bd6c4703fecedf4237dff3091e515d3d000951a to show what that would look like, but I'm not certain this is truly better than revealing one if needed. If we reveal an address but don't use it, then future calls to `create_tx` shouldn't cause a runaway of revealed addresses

ValuedMammal: alternatively I'm considering the approach in 9341041cfdadc329f7850cf028059db84d4eb658

crates/wallet/src/wallet/mod.rs
```diff
@@ -1457,8 +1458,8 @@ impl Wallet {
                     .index
                     .next_unused_spk(change_keychain)
```
evanlinjin: It's a bug to reveal something when we error out. Calling `.next_unused_spk` may reveal an spk here. I think it makes sense to fix this in the same PR.

crates/wallet/src/wallet/mod.rs
```diff
@@ -1557,6 +1568,20 @@ impl Wallet {
...
+                let (_, index_changeset) = self
+                    .indexed_graph
+                    .index
+                    .reveal_to_target(keychain, index)
+                    .expect("must not be None");
+                self.stage.merge(index_changeset.into());
+                self.mark_used(keychain, index);
+            }
+        }
+
```
evanlinjin: Nit

```suggestion
        if let (Excess::Change { .. }, Some((keychain, index))) = (excess, drain_index) {
            let (_, index_changeset) = self
                .indexed_graph
                .index
                .reveal_to_target(keychain, index)
                .expect("must not be None");
            self.stage.merge(index_changeset.into());
            self.mark_used(keychain, index);
        }

```

crates/wallet/tests/wallet.rs
```diff
@@ -1392,6 +1392,55 @@ fn test_create_tx_global_xpubs_with_origin() {
...
+    let err = builder.finish().unwrap_err();
+    assert!(matches!(
+        err,
+        CreateTxError::CoinSelection(Error::InsufficientFunds { .. }),
+    ));
+    assert_eq!(
+        wallet.next_unused_address(KeychainKind::Internal),
+        internal_addr0,
+        "create tx fail should not mark change address used",
+    );
```
evanlinjin: Nit: Can also check that the `last_derivation_index` is not incremented. An easy way to do this is to check whether `Wallet::stage` is empty.

evanlinjin (approved):
ACK 18ecd0641438c455e8ca9a666d0306b31c2a0d8f

crates/wallet/tests/wallet.rs
ValuedMammal: Added checks for derivation index in 004b0caf30c08b4177a92076fdc448191ce9cf89

crates/wallet/tests/wallet.rs
```diff
@@ -1392,6 +1392,60 @@ fn test_create_tx_global_xpubs_with_origin() {
...
+    assert!(
+        !wallet.mark_used(KeychainKind::Internal, 0),
+        "creating tx should mark change address used",
+    );
+    assert!(wallet
+        .list_unused_addresses(KeychainKind::Internal)
+        .next()
+        .is_none());
+    assert_eq!(wallet.reveal_next_address(KeychainKind::Internal).index, 1);
+}
```
oleonardolima: nit: you could also check the `derivation_index` as above.

crates/wallet/tests/wallet.rs
```diff
@@ -1392,6 +1392,60 @@ fn test_create_tx_global_xpubs_with_origin() {
     assert_eq!(psbt.xpub.get(&key), Some(&(fingerprint, path)));
 }
 
+#[test]
+fn test_create_tx_increment_change_index() {
+    use coin_selection::Error;
+    let (desc, change_desc) = get_test_tr_single_sig_xprv_with_change_desc();
```
oleonardolima, line 1416: I guess you could have a test that tests for a single descriptor wallet, without the change descriptor. WDYT ?

oleonardolima (commented):
utACK 004b0caf30c08b4177a92076fdc448191ce9cf89

It looks good! Thanks for addressing this issue.
I left some suggestions and nit below.

crates/wallet/tests/wallet.rs
ValuedMammal: Let me know if you like the test better in 849e13691031cb811567775c54325ae51e67a901

oleonardolima (approved):
ACK 849e13691031cb811567775c54325ae51e67a901

crates/wallet/tests/wallet.rs
oleonardolima: It looks good, thanks!

ValuedMammal:
Made last minute [changes](https://github.com/bitcoindevkit/bdk/compare/849e13691031cb811567775c54325ae51e67a901..606fa0874db0f10cd1c64de0f1f097b12db3a16d) @notmandatory

notmandatory (approved):
ACK 606fa0874db0f10cd1c64de0f1f097b12db3a16d

PR #1580, opened 2024-08-28
ci: add token for cron-update-rust.yml

notmandatory:
### Description

Add organization app token and GPG signing key for cron-update-rust.yml.

### Notes to the reviewers

I went with the organization github app token option mentioned here:
https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs

I added gpg commit signing with below instructions. The instructions say to use PAT for signing but the plugin doesn't mention it's needed so I want to try it with only the github app token.
https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#gpg-commit-signature-verification

### DIFF
--- a/.github/workflows/cron-update-rust.yml
+++ b/.github/workflows/cron-update-rust.yml
@@ -10,6 +10,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
+      - uses: tibdex/github-app-token@v1
+        id: generate-token
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.APP_PRIVATE_KEY }}
       - name: Update rust-version to use latest stable
         run: |
           set -x
@@ -30,7 +35,7 @@ jobs:
         if: env.changes_made == 'true'
         uses: peter-evans/create-pull-request@v6
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ steps.generate-token.outputs.token }}
           author: Update Rustc Bot <bot@example.com>
           committer: Update Rustc Bot <bot@example.com>
           branch: create-pull-request/update-rust-version
--- a/.github/workflows/cron-update-rust.yml
+++ b/.github/workflows/cron-update-rust.yml
@@ -15,6 +15,11 @@ jobs:
         with:
           app_id: ${{ secrets.APP_ID }}
           private_key: ${{ secrets.APP_PRIVATE_KEY }}
+      - uses: crazy-max/ghaction-import-gpg@v5
+        with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
       - name: Update rust-version to use latest stable
         run: |
           set -x
@@ -36,8 +41,8 @@ jobs:
         uses: peter-evans/create-pull-request@v6
         with:
           token: ${{ steps.generate-token.outputs.token }}
-          author: Update Rustc Bot <bot@example.com>
-          committer: Update Rustc Bot <bot@example.com>
+          author: Github Action <github@bitcoindevkit.org>
+          committer: Github Action <github@bitcoindevkit.org>
           branch: create-pull-request/update-rust-version
           title: |
             ci: automated update to rustc ${{ env.rust_version }}


.github/workflows/cron-update-rust.yml
```diff
@@ -10,6 +10,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
+      - uses: tibdex/github-app-token@v1
+        id: generate-token
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.APP_PRIVATE_KEY }}
```
oleonardolima, line 17: Do we need this new GH Action ?

Although it's suggested in the `peter-evans/create-pull-request`, can't we just follow the same approach as rust-bitcoin ([here](https://github.com/rust-bitcoin/rust-bitcoin/blob/de8c70ccfd59d0ed0758936a5fc7a40afa266ac6/.github/workflows/cron-semi-weekly-update-nightly.yml#L30-L41)) and use a new token ?

.github/workflows/cron-update-rust.yml
```diff
@@ -15,6 +15,11 @@ jobs:
         with:
           app_id: ${{ secrets.APP_ID }}
           private_key: ${{ secrets.APP_PRIVATE_KEY }}
+      - uses: crazy-max/ghaction-import-gpg@v5
+        with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
```
oleonardolima, line 22: I'm unsure how this action impact security-wise, but it doesn't seem very used/maintained :think:. We can always sign off the commits if we opt to not use this action.

notmandatory: It looks like the rust-bitcoin repo doesn't enforce that the commits be signed. With our current repo rules we can't merge unsigned commits. https://github.com/rust-bitcoin/rust-bitcoin/pull/3267/commits

So we'd need to take off this protection, or figure out a way to get the github action to create a properly signed commit.

ValuedMammal: This looks like it will work, but I want to make sure we've considered the alternatives and that these new actions won't be a hassle to maintain for someone like  @notmandatory 

Options:

- for security, fork or pin third-party actions to a commit SHA (requires code review)
- drop requirement to sign commits (not ideal)
- allow actions to create PR but don't merge it (not ideal)
- just pin clippy to stable and forgo the extra automation

notmandatory: I'm OK with these automations since they will prevent us from getting unexpected breaks when `stable` updates behind the scenes.  My main concern is I don't want to use my (or anyone else's) github or gpg credentials to create and sign the PR commit. 

I think the github app credentials for PR creation as I have it now are a fine way to go. If the GPG plugin I added doesn't work then we could just let the action create the PR and who ever merges it will need to first rebase-resign the commit manually. It shouldn't happen too often so I'm not worried about the extra step.

notmandatory: BTW I think I can test this in my personal repo, will post here to confirm if it works as expected. 

Also I'm not so worried about the plugin versions being pinned since this can only create PRs it can't commit them.

notmandatory: I tested on my local repo and it works, see: https://github.com/notmandatory/bdk/pull/9/commits

I had to create a new GPG key for the action, add the corresponding GPG pubkey to my account, and finally verify the email for this GPG key with my account. But after all that github created the PR, automatically runs the CI, and is able to verify the commit so the PR can be merged. 🎉

oleonardolima (approved):
ACK b140b32648241884e0e5e222fd7a0fc4e05908f1

ValuedMammal (approved):
ACK b140b32648241884e0e5e222fd7a0fc4e05908f1

.github/workflows/cron-update-rust.yml
ValuedMammal: > I tested on my local repo and it works, see: https://github.com/notmandatory/bdk/pull/9/commits

Do you happen to know why the ci ran 41 checks on that PR?

oleonardolima: I guess it's because it ran on pull_request and push dispatches 🤔

PR #1581, opened 2024-08-28
Allow custom fallback algorithm for bnb

evanlinjin:
### Description

This allows the caller to set a custom fallback algorithm when using `BranchAndBoundCoinSelection`. Previously, you were forced into using `SingleRandomDraw`.

Signature of `CoinSelectionAlgorithm::coin_select` has been changed to take in a `&mut RangCore`. This allows us to pass the random number generator directly to the cs algorithm.

Single random draw is now it's own type `SingleRandomDraw` and impls `CoinSelectionAlgorithm`.

`BranchAndBoundCoinSelection` now handles it's own fallback algorithm internally, and a generic type parameter is added to specify the fallback algorithm.

`coin_selection::Error` is renamed to `InsufficientFunds` and the BnB error variants are removed. The BnB error variants are no longer needed since those cases are handled internally by `BranchAndBoundCoinSelection` (via calling the fallback algorithm).

### Notes to the reviewers

This is breaking change. Not sure how useful this is for our users. If it's deemed useful, consider including in beta.

### Changelog notice

* Changed `CoinSelectionAlgorithm::coin_select` to take in an additional `&mut RangCore` variable. This allows us to pass a random number generator directly to the cs algorithm.
* Added `SingeRandomDraw` type which impls `CoinSelectAlgorithm`.
* Changed `BranchAndBoundCoinSelection` to call the fallback internally. An additional generic parameter is added set this.
* Changed `coin_selection::Error` to `coin_selection::InsufficientFunds` (which is now a struct) and therefore removing bnb error variants.

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/coin_selection.rs
+++ b/crates/wallet/src/wallet/coin_selection.rs
@@ -31,18 +31,20 @@
 //! # use bdk_wallet::*;
 //! # use bdk_wallet::coin_selection::decide_change;
 //! # use anyhow::Error;
+//! # use rand_core::RngCore;
 //! #[derive(Debug)]
 //! struct AlwaysSpendEverything;
 //!
 //! impl CoinSelectionAlgorithm for AlwaysSpendEverything {
-//!     fn coin_select(
+//!     fn coin_select<R: RngCore>(
 //!         &self,
 //!         required_utxos: Vec<WeightedUtxo>,
 //!         optional_utxos: Vec<WeightedUtxo>,
 //!         fee_rate: FeeRate,
 //!         target_amount: u64,
 //!         drain_script: &Script,
-//!     ) -> Result<CoinSelectionResult, coin_selection::Error> {
+//!         rand: &mut R,
+//!     ) -> Result<CoinSelectionResult, coin_selection::InsufficientFunds> {
 //!         let mut selected_amount = 0;
 //!         let mut additional_weight = Weight::ZERO;
 //!         let all_utxos_selected = required_utxos
@@ -63,7 +65,7 @@
 //!         let additional_fees = (fee_rate * additional_weight).to_sat();
 //!         let amount_needed_with_fees = additional_fees + target_amount;
 //!         if selected_amount < amount_needed_with_fees {
-//!             return Err(coin_selection::Error::InsufficientFunds {
+//!             return Err(coin_selection::InsufficientFunds {
 //!                 needed: amount_needed_with_fees,
 //!                 available: selected_amount,
 //!             });
@@ -118,44 +120,31 @@ use rand_core::RngCore;
 use super::utils::shuffle_slice;
 /// Default coin selection algorithm used by [`TxBuilder`](super::tx_builder::TxBuilder) if not
 /// overridden
-pub type DefaultCoinSelectionAlgorithm = BranchAndBoundCoinSelection;
+pub type DefaultCoinSelectionAlgorithm = BranchAndBoundCoinSelection<SingleRandomDraw>;
 
-/// Errors that can be thrown by the [`coin_selection`](crate::wallet::coin_selection) module
-#[derive(Debug)]
-pub enum Error {
-    /// Wallet's UTXO set is not enough to cover recipient's requested plus fee
-    InsufficientFunds {
-        /// Sats needed for some transaction
-        needed: u64,
-        /// Sats available for spending
-        available: u64,
-    },
-    /// Branch and bound coin selection tries to avoid needing a change by finding the right inputs for
-    /// the desired outputs plus fee, if there is not such combination this error is thrown
-    BnBNoExactMatch,
-    /// Branch and bound coin selection possible attempts with sufficiently big UTXO set could grow
-    /// exponentially, thus a limit is set, and when hit, this error is thrown
-    BnBTotalTriesExceeded,
+/// Wallet's UTXO set is not enough to cover recipient's requested plus fee.
+///
+/// This is thrown by [`CoinSelectionAlgorithm`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct InsufficientFunds {
+    /// Sats needed for some transaction
+    pub needed: u64,
+    /// Sats available for spending
+    pub available: u64,
 }
 
-impl fmt::Display for Error {
+impl fmt::Display for InsufficientFunds {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        match self {
-            Self::InsufficientFunds { needed, available } => write!(
-                f,
-                "Insufficient funds: {} sat available of {} sat needed",
-                available, needed
-            ),
-            Self::BnBTotalTriesExceeded => {
-                write!(f, "Branch and bound coin selection: total tries exceeded")
-            }
-            Self::BnBNoExactMatch => write!(f, "Branch and bound coin selection: not exact match"),
-        }
+        write!(
+            f,
+            "Insufficient funds: {} sat available of {} sat needed",
+            self.available, self.needed
+        )
     }
 }
 
 #[cfg(feature = "std")]
-impl std::error::Error for Error {}
+impl std::error::Error for InsufficientFunds {}
 
 #[derive(Debug)]
 /// Remaining amount after performing coin selection
@@ -216,8 +205,6 @@ impl CoinSelectionResult {
 pub trait CoinSelectionAlgorithm: core::fmt::Debug {
     /// Perform the coin selection
     ///
-    /// - `database`: a reference to the wallet's database that can be used to lookup additional
-    ///               details for a specific UTXO
     /// - `required_utxos`: the utxos that must be spent regardless of `target_amount` with their
     ///                     weight cost
     /// - `optional_utxos`: the remaining available utxos to satisfy `target_amount` with their
@@ -226,15 +213,17 @@ pub trait CoinSelectionAlgorithm: core::fmt::Debug {
     /// - `target_amount`: the outgoing amount in satoshis and the fees already
     ///                    accumulated from added outputs and transaction’s header.
     /// - `drain_script`: the script to use in case of change
+    /// - `rand`: random number generated used by some coin selection algorithms such as [`SingleRandomDraw`]
     #[allow(clippy::too_many_arguments)]
-    fn coin_select(
+    fn coin_select<R: RngCore>(
         &self,
         required_utxos: Vec<WeightedUtxo>,
         optional_utxos: Vec<WeightedUtxo>,
         fee_rate: FeeRate,
         target_amount: u64,
         drain_script: &Script,
-    ) -> Result<CoinSelectionResult, Error>;
+        rand: &mut R,
+    ) -> Result<CoinSelectionResult, InsufficientFunds>;
 }
 
 /// Simple and dumb coin selection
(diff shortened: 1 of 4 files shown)


ValuedMammal:
Concept ACK. I want to take a closer look at it. It would also be nice to try and fix the ignored tests in `coin_selection` module

ValuedMammal:
Simple test suggestion 
```rs
    #[test]
    fn test_bnb_fallback_algorithm() {
        // utxo value
        // 120k + 80k + 300k
        let optional_utxos = get_oldest_first_test_utxos();
        let feerate = FeeRate::BROADCAST_MIN;
        let target_amount = 190_000;
        let drain_script = ScriptBuf::new();
        // bnb won't find exact match and should select oldest first
        let res = BranchAndBoundCoinSelection::<OldestFirstCoinSelection>::default()
            .coin_select(
                vec![],
                optional_utxos,
                feerate,
                target_amount,
                &drain_script,
                &mut thread_rng(),
            )
            .unwrap();
        assert_eq!(res.selected_amount(), 200_000);
    }
```

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -227,14 +216,15 @@ pub trait CoinSelectionAlgorithm: core::fmt::Debug {
     ///                    accumulated from added outputs and transaction’s header.
     /// - `drain_script`: the script to use in case of change
     #[allow(clippy::too_many_arguments)]
-    fn coin_select(
+    fn coin_select<R: RngCore>(
```
ValuedMammal, line 217: Docs here can be updated

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -407,56 +399,73 @@ impl OutputGroup {
...
         }
     }
 }
 
-impl BranchAndBoundCoinSelection {
+impl<FA> BranchAndBoundCoinSelection<FA> {
     /// Create new instance with target size for change output
-    pub fn new(size_of_change: u64) -> Self {
-        Self { size_of_change }
+    pub fn new(size_of_change: u64, fallback_algorithm: FA) -> Self {
```
ValuedMammal:
```suggestion
    /// Create new instance with a target `size_of_change` and `fallback_algorithm`.
    pub fn new(size_of_change: u64, fallback_algorithm: FA) -> Self {
```

evanlinjin:
Thanks @ValuedMammal feel free to push a commit to introduce the test (if you wish).

oleonardolima (commented):
Concept ACK 36f7d39362ed05dc99071195e7eb739a904f2ed5
Overall looks good! It's getting late, I'll do another review and test it tomorrow :)

ValuedMammal:
My main gripe is making the RNG required in order to implement `CoinSelectionAlgorithm` when previously it wasn't and technically none of our other algos require randomness. But there may be a way to keep the new struct `SingleRandomDraw` and implement the trait feature-gated behind "std" so that we can use `thread_rng`. Then BnB can fallback to single random draw if "std" feature is enabled, otherwise a user-provided algorithm. I haven't fully fleshed it out in code yet.

oleonardolima:
> [...] It would also be nice to try and fix the ignored tests in `coin_selection` module

AFAICT it only needs to removes the `#[ignore = ...]` from lines 1148 and 1193, as SRD is the fallback algorithm now. At least those two worked just fine testing locally.

Maybe renaming the test fn to mention it relies on fallback is good too.

edit: I still haven't figured out the problem with the other one from line 1236, as it's been ignored for a while.

notmandatory:
> My main gripe is making the RNG required in order to implement `CoinSelectionAlgorithm` when previously it wasn't and technically none of our other algos require randomness. But there may be a way to keep the new struct `SingleRandomDraw` and implement the trait feature-gated behind "std" so that we can use `thread_rng`. Then BnB can fallback to single random draw if "std" feature is enabled, otherwise a user-provided algorithm. I haven't fully fleshed it out in code yet.

I also took a stab at encapsulating the RNG for `SingleRandomDraw` and it's cleaner in some ways but breaks the ability to use `DefaultCoinSelectionAlgorithm` in the `no_std` mode.  So for now I think the way it is here is best.

notmandatory:
@evanlinjin feel free to squash 89e619edbbda9c740765a3a13715bd78ad338da8 which has @ValuedMammal's suggestions in it and I think this one also needs a rebase.

notmandatory (approved):
ACK 89e619edbbda9c740765a3a13715bd78ad338da8

evanlinjin (commented):
I think this PR still requires a few changes. If someone can take over this PR (and review and come to consensus on the following suggestions), that will be absolutely wonderful.

1. We currently use `.sort_unstable_by_key` to sort UTXOs for our coin selection algorithm impls. This is fine, but to make things deterministic (i.e. equal-key elements are ordered the same each time), we should add something like `OutPoint` at the end of the key used to sort.
2. In bnb, we use `.reverse`. However, to make things consistent with the way we use `OutPoint` as part of the sort-key, consider using [`Reverse(effective_value)`](https://doc.rust-lang.org/std/cmp/struct.Reverse.html) as the first tuple element of the sort-key.
3. This PR needs a rebase.

crates/wallet/src/wallet/coin_selection.rs
notmandatory: fixed.

crates/wallet/src/wallet/coin_selection.rs
notmandatory: fixed.

notmandatory:
I've rebased this PR and incorporated @ValuedMammal 's docs and new bnb fallback test.

notmandatory:
> I think this PR still requires a few changes. If someone can take over this PR (and review and come to consensus on the following suggestions), that will be absolutely wonderful.
> 
>     1. We currently use `.sort_unstable_by_key` to sort UTXOs for our coin selection algorithm impls. This is fine, but to make things deterministic (i.e. equal-key elements are ordered the same each time), we should add something like `OutPoint` at the end of the key used to sort.

I added a test to check if deterministic coin selection algos were always returning the same utxos and even without changing `.sort_unstable_by_key` to `.sort_by_key` I'm getting deterministic results.  I still changed to using `.sort_by_key` just to be safe.

>     2. In bnb, we use `.reverse`. However, to make things consistent with the way we use `OutPoint` as part of the sort-key, consider using [`Reverse(effective_value)`](https://doc.rust-lang.org/std/cmp/struct.Reverse.html) as the first tuple element of the sort-key.

 I didn't have to change `.reverse` in bnb to get deterministic results. 

I also did some test cleanup and added back ignored tests that work as long as we calculate the correct target amount that bnb needs to select the expected utxos.

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -894,6 +929,36 @@ mod test {
             .sum()
     }
 
+    /// Use this when you want to ensure that the fallback coin selection is NOT used.
+    #[derive(Debug)]
+    struct PanicFallBackCoinSelection {}
```
ValuedMammal: I would like to get rid of this

notmandatory: Are you thinking about making fallback optional instead? or just some other way to ensure we don't call the fallback during testing?

ValuedMammal: We're already testing bnb with no fallback by just calling the `bnb` function

ValuedMammal:
Regarding utxo sorting I think we're also saved by the fact that `list_unspent` calls `index.outpoints()` that returns an ordered set of `((K, u32), OutPoint)` - it seems like we can rely on this sorting. Also isn't it true that no two `WeightedUtxos` should be the same because it already contains the outpoint?

evanlinjin:
I think I pushed everyone onto the wrong rabbit-hole with regards to `.sort_unstable_by_key`. The "unstable" part only means that equal elements may be reordered. There is no randomness here however, so with the same UTXO set, the sorted order will still be the same (as you probably found in your tests @notmandatory).

I think we should revert back to using `.sort_unstable_by_key` as it is more performant and there is no benefit for keeping equal elements ordered the same.

crates/wallet/src/wallet/coin_selection.rs
ValuedMammal: I added suggestions in evanlinjin/bdk#13

notmandatory: Thanks @ValuedMammal, I'll pulled in your suggested changes and remove the `PanicFallBackCoinSelection` stuff.

notmandatory:
> I think we should revert back to using `.sort_unstable_by_key` as it is more performant and there is no benefit for keeping equal elements ordered the same.

Makes sense, reverted.

oleonardolima (commented):
utACK 65be4ead03126430ab47bf192d0de0db1b1b883a

Cool, the new deterministic UTXO selection tests look like a great addition!

evanlinjin (commented):
self-ACK 65be4ead03126430ab47bf192d0de0db1b1b883a

crates/wallet/src/wallet/coin_selection.rs
ValuedMammal, line 1:
```diff
@@ -402,7 +402,7 @@ pub struct BranchAndBoundCoinSelection<Cs = SingleRandomDraw> {
     fallback_algorithm: Cs,
 }
 
-/// Error returned by branch and bond coin selection.
+/// Error returned by branch and bound coin selection.
 #[derive(Debug)]
 enum BnbError {
     /// Branch and bound coin selection tries to avoid needing a change by finding the right inputs for
@@ -521,8 +521,8 @@ impl<Cs: CoinSelectionAlgorithm> CoinSelectionAlgorithm for BranchAndBoundCoinSe
         }
 
         match self.bnb(
-            required_ogs.clone(),
-            optional_ogs.clone(),
+            required_ogs,
+            optional_ogs,
             curr_value,
             curr_available_value,
             signed_target_amount,
@@ -1383,7 +1383,7 @@ mod test {
 
         let drain_script = ScriptBuf::default();
         let target_amount = 20_000 + FEE_AMOUNT;
-        BranchAndBoundCoinSelection::<SingleRandomDraw>::default()
+        BranchAndBoundCoinSelection::new(size_of_change, SingleRandomDraw)
             .bnb(
                 vec![],
                 utxos,
@@ -1414,7 +1414,7 @@ mod test {
 
         let drain_script = ScriptBuf::default();
 
-        BranchAndBoundCoinSelection::<SingleRandomDraw>::default()
+        BranchAndBoundCoinSelection::new(size_of_change, SingleRandomDraw)
             .bnb(
                 vec![],
                 utxos,
@@ -1450,7 +1450,7 @@ mod test {
 
         let drain_script = ScriptBuf::default();
 
-        let result = BranchAndBoundCoinSelection::<SingleRandomDraw>::default()
+        let result = BranchAndBoundCoinSelection::new(size_of_change, SingleRandomDraw)
             .bnb(
                 vec![],
                 utxos,
```

notmandatory: I'm working on a follow-up PR. I also think the tests need to be improved and there's a bigger issue with `SingleRandomDraw` in that if used on its own it doesn't fail when there's not enough sats to meet the required amount.

notmandatory: Fixed in #1605

PR #1582, opened 2024-08-29
refactor!: Implement generics for `CheckPoint`, `LocalChain`, and `spk_client` types

### LINKED ISSUE #1937: Implement generics for `CheckPoint`, `LocalChain`, and `spk_client` types
Currently, the `CheckPoint`, `LocalChain`, and `spk_client` components are not designed to handle generic types, which limits their flexibility. To support caching of Merkle proofs (#1699) and median time passed calculations, it would be beneficial for these components to accept a generic type, specifically to accommodate `Header`s.

LagginTimes:
Implements bitcoindevkit/bdk#1937.

### Description

This PR is a step towards header checkpointing for `bdk_electrum`. The goal is to be able to store whole headers in `CheckPoint` so they do not have to be re-downloaded. Storing headers this way would be a prerequisite for caching of merkle proofs and for median time passed.

### Changelog notice

* `CheckPoint` takes in a generic.
* `LocalChain` and `ChangeSet` take in generics.
* `spk_client` types can take in generics.

#### All Submissions:

### DIFF
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -1,38 +1,43 @@
 //! The [`LocalChain`] is a local implementation of [`ChainOracle`].
 
 use core::convert::Infallible;
+use core::fmt;
 use core::ops::RangeBounds;
 
 use crate::collections::BTreeMap;
 use crate::{BlockId, ChainOracle, Merge};
+use bdk_core::ToBlockHash;
 pub use bdk_core::{CheckPoint, CheckPointIter};
 use bitcoin::block::Header;
 use bitcoin::BlockHash;
 
 /// Apply `changeset` to the checkpoint.
-fn apply_changeset_to_checkpoint(
-    mut init_cp: CheckPoint,
-    changeset: &ChangeSet,
-) -> Result<CheckPoint, MissingGenesisError> {
+fn apply_changeset_to_checkpoint<D>(
+    mut init_cp: CheckPoint<D>,
+    changeset: &ChangeSet<D>,
+) -> Result<CheckPoint<D>, MissingGenesisError>
+where
+    D: ToBlockHash + fmt::Debug + Copy,
+{
     if let Some(start_height) = changeset.blocks.keys().next().cloned() {
         // changes after point of agreement
         let mut extension = BTreeMap::default();
         // point of agreement
-        let mut base: Option<CheckPoint> = None;
+        let mut base: Option<CheckPoint<D>> = None;
 
         for cp in init_cp.iter() {
             if cp.height() >= start_height {
-                extension.insert(cp.height(), cp.hash());
+                extension.insert(cp.height(), cp.data());
             } else {
                 base = Some(cp);
                 break;
             }
         }
 
-        for (&height, &hash) in &changeset.blocks {
-            match hash {
-                Some(hash) => {
-                    extension.insert(height, hash);
+        for (&height, &data) in &changeset.blocks {
+            match data {
+                Some(data) => {
+                    extension.insert(height, data);
                 }
                 None => {
                     extension.remove(&height);
@@ -42,7 +47,7 @@ fn apply_changeset_to_checkpoint(
 
         let new_tip = match base {
             Some(base) => base
-                .extend(extension.into_iter().map(BlockId::from))
+                .extend(extension)
                 .expect("extension is strictly greater than base"),
             None => LocalChain::from_blocks(extension)?.tip(),
         };
@@ -53,12 +58,18 @@ fn apply_changeset_to_checkpoint(
 }
 
 /// This is a local implementation of [`ChainOracle`].
-#[derive(Debug, Clone, PartialEq)]
-pub struct LocalChain {
-    tip: CheckPoint,
+#[derive(Debug, Clone)]
+pub struct LocalChain<D = BlockHash> {
+    tip: CheckPoint<D>,
 }
 
-impl ChainOracle for LocalChain {
+impl<D> PartialEq for LocalChain<D> {
+    fn eq(&self, other: &Self) -> bool {
+        self.tip == other.tip
+    }
+}
+
+impl<D> ChainOracle for LocalChain<D> {
     type Error = Infallible;
 
     fn is_block_in_chain(
@@ -83,101 +94,8 @@ impl ChainOracle for LocalChain {
     }
 }
 
-impl LocalChain {
-    /// Get the genesis hash.
-    pub fn genesis_hash(&self) -> BlockHash {
-        self.tip.get(0).expect("genesis must exist").hash()
-    }
-
-    /// Construct [`LocalChain`] from genesis `hash`.
-    #[must_use]
-    pub fn from_genesis_hash(hash: BlockHash) -> (Self, ChangeSet) {
-        let height = 0;
-        let chain = Self {
-            tip: CheckPoint::new(BlockId { height, hash }),
-        };
-        let changeset = chain.initial_changeset();
-        (chain, changeset)
-    }
-
-    /// Construct a [`LocalChain`] from an initial `changeset`.
-    pub fn from_changeset(changeset: ChangeSet) -> Result<Self, MissingGenesisError> {
-        let genesis_entry = changeset.blocks.get(&0).copied().flatten();
-        let genesis_hash = match genesis_entry {
-            Some(hash) => hash,
-            None => return Err(MissingGenesisError),
-        };
-
-        let (mut chain, _) = Self::from_genesis_hash(genesis_hash);
-        chain.apply_changeset(&changeset)?;
-
-        debug_assert!(chain._check_changeset_is_applied(&changeset));
-
-        Ok(chain)
-    }
-
-    /// Construct a [`LocalChain`] from a given `checkpoint` tip.
-    pub fn from_tip(tip: CheckPoint) -> Result<Self, MissingGenesisError> {
-        let genesis_cp = tip.iter().last().expect("must have at least one element");
-        if genesis_cp.height() != 0 {
-            return Err(MissingGenesisError);
-        }
-        Ok(Self { tip })
-    }
-
-    /// Constructs a [`LocalChain`] from a [`BTreeMap`] of height to [`BlockHash`].
-    ///
-    /// The [`BTreeMap`] enforces the height order. However, the caller must ensure the blocks are
-    /// all of the same chain.
-    pub fn from_blocks(blocks: BTreeMap<u32, BlockHash>) -> Result<Self, MissingGenesisError> {
-        if !blocks.contains_key(&0) {
-            return Err(MissingGenesisError);
-        }
-
-        let mut tip: Option<CheckPoint> = None;
-        for block in &blocks {
-            match tip {
-                Some(curr) => {
-                    tip = Some(
-                        curr.push(BlockId::from(block))
-                            .expect("BTreeMap is ordered"),
-                    )
-                }
-                None => tip = Some(CheckPoint::new(BlockId::from(block))),
-            }
-        }
-
-        Ok(Self {
-            tip: tip.expect("already checked to have genesis"),
-        })
-    }
-
-    /// Get the highest checkpoint.
-    pub fn tip(&self) -> CheckPoint {
-        self.tip.clone()
-    }
-
-    /// Applies the given `update` to the chain.
-    ///
-    /// The method returns [`ChangeSet`] on success. This represents the changes applied to `self`.
-    ///
-    /// There must be no ambiguity about which of the existing chain's blocks are still valid and
-    /// which are now invalid. That is, the new chain must implicitly connect to a definite block in
-    /// the existing chain and invalidate the block after it (if it exists) by including a block at
-    /// the same height but with a different hash to explicitly exclude it as a connection point.
-    ///
-    /// # Errors
-    ///
-    /// An error will occur if the update does not correctly connect with `self`.
-    ///
-    /// [module-level documentation]: crate::local_chain
-    pub fn apply_update(&mut self, update: CheckPoint) -> Result<ChangeSet, CannotConnectError> {
-        let (new_tip, changeset) = merge_chains(self.tip.clone(), update)?;
-        self.tip = new_tip;
-        debug_assert!(self._check_changeset_is_applied(&changeset));
-        Ok(changeset)
-    }
-
+// Methods for `LocalChain<BlockHash>`
+impl LocalChain<BlockHash> {
     /// Update the chain with a given [`Header`] at `height` which you claim is connected to a
     /// existing block in the chain.
     ///
@@ -227,8 +145,13 @@ impl LocalChain {
             conn => Some(conn),
         };
 
-        let update = CheckPoint::from_block_ids([conn, prev, Some(this)].into_iter().flatten())
-            .expect("block ids must be in order");
+        let update = CheckPoint::from_blocks(
+            [conn, prev, Some(this)]
+                .into_iter()
+                .flatten()
+                .map(|block_id| (block_id.height, block_id.hash)),
+        )
+        .expect("block ids must be in order");
 
         self.apply_update(update)
             .map_err(ApplyHeaderError::CannotConnect)
@@ -264,9 +187,130 @@ impl LocalChain {
                 ApplyHeaderError::CannotConnect(err) => err,
             })
     }
+}
+
+// Methods for any `D`
+impl<D> LocalChain<D> {
+    /// Get the highest checkpoint.
+    pub fn tip(&self) -> CheckPoint<D> {
+        self.tip.clone()
+    }
+
+    /// Get the genesis hash.
+    pub fn genesis_hash(&self) -> BlockHash {
+        self.tip.get(0).expect("genesis must exist").block_id().hash
+    }
+
+    /// Iterate over checkpoints in descending height order.
+    pub fn iter_checkpoints(&self) -> CheckPointIter<D> {
+        self.tip.iter()
+    }
+
+    /// Get checkpoint at given `height` (if it exists).
+    ///
+    /// This is a shorthand for calling [`CheckPoint::get`] on the [`tip`].
+    ///
+    /// [`tip`]: LocalChain::tip
+    pub fn get(&self, height: u32) -> Option<CheckPoint<D>> {
+        self.tip.get(height)
+    }
+
+    /// Iterate checkpoints over a height range.
+    ///
+    /// Note that we always iterate checkpoints in reverse height order (iteration starts at tip
+    /// height).
+    ///
+    /// This is a shorthand for calling [`CheckPoint::range`] on the [`tip`].
+    ///
+    /// [`tip`]: LocalChain::tip
+    pub fn range<R>(&self, range: R) -> impl Iterator<Item = CheckPoint<D>>
+    where
+        R: RangeBounds<u32>,
+    {
+        self.tip.range(range)
+    }
+}
+
+// Methods where `D: ToBlockHash`
+impl<D> LocalChain<D>
+where
+    D: ToBlockHash + fmt::Debug + Copy,
+{
+    /// Constructs a [`LocalChain`] from genesis data.
+    pub fn from_genesis(data: D) -> (Self, ChangeSet<D>) {
+        let height = 0;
+        let chain = Self {
+            tip: CheckPoint::new(height, data),
+        };
+        let changeset = chain.initial_changeset();
+
+        (chain, changeset)
+    }
+
+    /// Constructs a [`LocalChain`] from a [`BTreeMap`] of height and data `D`.
+    ///
+    /// The [`BTreeMap`] enforces the height order. However, the caller must ensure the blocks are
+    /// all of the same chain.
+    pub fn from_blocks(blocks: BTreeMap<u32, D>) -> Result<Self, MissingGenesisError> {
+        if !blocks.contains_key(&0) {
+            return Err(MissingGenesisError);
+        }
+
+        Ok(Self {
+            tip: CheckPoint::from_blocks(blocks).expect("blocks must be in order"),
+        })
+    }
+
+    /// Construct a [`LocalChain`] from an initial `changeset`.
+    pub fn from_changeset(changeset: ChangeSet<D>) -> Result<Self, MissingGenesisError> {
+        let genesis_entry = changeset.blocks.get(&0).copied().flatten();
+        let genesis_data = match genesis_entry {
+            Some(data) => data,
+            None => return Err(MissingGenesisError),
+        };
+
+        let (mut chain, _) = Self::from_genesis(genesis_data);
+        chain.apply_changeset(&changeset)?;
+        debug_assert!(chain._check_changeset_is_applied(&changeset));
+        Ok(chain)
+    }
+
+    /// Construct a [`LocalChain`] from a given `checkpoint` tip.
+    pub fn from_tip(tip: CheckPoint<D>) -> Result<Self, MissingGenesisError> {
+        let genesis_cp = tip.iter().last().expect("must have at least one element");
+        if genesis_cp.height() != 0 {
+            return Err(MissingGenesisError);
+        }
+
+        Ok(Self { tip })
+    }
+
+    /// Applies the given `update` to the chain.
+    ///
+    /// The method returns [`ChangeSet`] on success. This represents the changes applied to `self`.
+    ///
+    /// There must be no ambiguity about which of the existing chain's blocks are still valid and
+    /// which are now invalid. That is, the new chain must implicitly connect to a definite block in
+    /// the existing chain and invalidate the block after it (if it exists) by including a block at
+    /// the same height but with a different hash to explicitly exclude it as a connection point.
+    ///
+    /// # Errors
+    ///
+    /// An error will occur if the update does not correctly connect with `self`.
+    ///
+    /// [module-level documentation]: crate::local_chain
+    pub fn apply_update(
+        &mut self,
+        update: CheckPoint<D>,
+    ) -> Result<ChangeSet<D>, CannotConnectError> {
+        let (new_tip, changeset) = merge_chains(self.tip.clone(), update)?;
+        self.tip = new_tip;
+        debug_assert!(self._check_changeset_is_applied(&changeset));
+        Ok(changeset)
+    }
 
     /// Apply the given `changeset`.
-    pub fn apply_changeset(&mut self, changeset: &ChangeSet) -> Result<(), MissingGenesisError> {
+    pub fn apply_changeset(&mut self, changeset: &ChangeSet<D>) -> Result<(), MissingGenesisError> {
         let old_tip = self.tip.clone();
         let new_tip = apply_changeset_to_checkpoint(old_tip, changeset)?;
         self.tip = new_tip;
@@ -274,33 +318,52 @@ impl LocalChain {
         Ok(())
     }
 
-    /// Insert a [`BlockId`].
+    /// Derives an initial [`ChangeSet`], meaning that it can be applied to an empty chain to
+    /// recover the current chain.
+    pub fn initial_changeset(&self) -> ChangeSet<D> {
+        ChangeSet {
+            blocks: self
+                .tip
+                .iter()
+                .map(|cp| (cp.height(), Some(cp.data())))
+                .collect(),
+        }
+    }
+
+    /// Insert block into a [`LocalChain`].
     ///
     /// # Errors
     ///
     /// Replacing the block hash of an existing checkpoint will result in an error.
-    pub fn insert_block(&mut self, block_id: BlockId) -> Result<ChangeSet, AlterCheckPointError> {
-        if let Some(original_cp) = self.tip.get(block_id.height) {
+    pub fn insert_block(
+        &mut self,
+        height: u32,
+        data: D,
+    ) -> Result<ChangeSet<D>, AlterCheckPointError> {
+        if let Some(original_cp) = self.tip.get(height) {
             let original_hash = original_cp.hash();
-            if original_hash != block_id.hash {
+            if original_hash != data.to_blockhash() {
                 return Err(AlterCheckPointError {
-                    height: block_id.height,
+                    height,
                     original_hash,
-                    update_hash: Some(block_id.hash),
+                    update_hash: Some(data.to_blockhash()),
                 });
             }
             return Ok(ChangeSet::default());
         }
 
-        let mut changeset = ChangeSet::default();
-        changeset
-            .blocks
-            .insert(block_id.height, Some(block_id.hash));
+        let mut changeset = ChangeSet::<D>::default();
+        changeset.blocks.insert(height, Some(data));
         self.apply_changeset(&changeset)
             .map_err(|_| AlterCheckPointError {
                 height: 0,
                 original_hash: self.genesis_hash(),
-                update_hash: changeset.blocks.get(&0).cloned().flatten(),
+                update_hash: changeset
+                    .blocks
+                    .get(&0)
+                    .cloned()
+                    .flatten()
+                    .map(|d| d.to_blockhash()),
             })?;
         Ok(changeset)
     }
@@ -314,8 +377,11 @@ impl LocalChain {
     ///
     /// This will fail with [`MissingGenesisError`] if the caller attempts to disconnect from the
     /// genesis block.
-    pub fn disconnect_from(&mut self, block_id: BlockId) -> Result<ChangeSet, MissingGenesisError> {
-        let mut remove_from = Option::<CheckPoint>::None;
+    pub fn disconnect_from(
+        &mut self,
+        block_id: BlockId,
+    ) -> Result<ChangeSet<D>, MissingGenesisError> {
+        let mut remove_from = Option::<CheckPoint<D>>::None;
         let mut changeset = ChangeSet::default();
         for cp in self.tip().iter() {
             let cp_id = cp.block_id();
@@ -340,38 +406,20 @@ impl LocalChain {
         Ok(changeset)
     }
 
-    /// Derives an initial [`ChangeSet`], meaning that it can be applied to an empty chain to
-    /// recover the current chain.
-    pub fn initial_changeset(&self) -> ChangeSet {
-        ChangeSet {
-            blocks: self
-                .tip
-                .iter()
-                .map(|cp| {
-                    let block_id = cp.block_id();
-                    (block_id.height, Some(block_id.hash))
-                })
-                .collect(),
-        }
-    }
-
-    /// Iterate over checkpoints in descending height order.
-    pub fn iter_checkpoints(&self) -> CheckPointIter {
-        self.tip.iter()
-    }
-
-    fn _check_changeset_is_applied(&self, changeset: &ChangeSet) -> bool {
-        let mut curr_cp = self.tip.clone();
-        for (height, exp_hash) in changeset.blocks.iter().rev() {
-            match curr_cp.get(*height) {
-                Some(query_cp) => {
-                    if query_cp.height() != *height || Some(query_cp.hash()) != *exp_hash {
+    fn _check_changeset_is_applied(&self, changeset: &ChangeSet<D>) -> bool {
+        let mut cur = self.tip.clone();
+        for (&exp_height, exp_data) in changeset.blocks.iter().rev() {
+            match cur.get(exp_height) {
+                Some(cp) => {
+                    if cp.height() != exp_height
+                        || Some(cp.hash()) != exp_data.map(|d| d.to_blockhash())
+                    {
                         return false;
                     }
-                    curr_cp = query_cp;
+                    cur = cp;
                 }
                 None => {
-                    if exp_hash.is_some() {
+                    if exp_data.is_some() {
                         return false;
                     }
                 }
@@ -379,44 +427,28 @@ impl LocalChain {
         }
         true
     }
-
-    /// Get checkpoint at given `height` (if it exists).
-    ///
-    /// This is a shorthand for calling [`CheckPoint::get`] on the [`tip`].
-    ///
-    /// [`tip`]: LocalChain::tip
-    pub fn get(&self, height: u32) -> Option<CheckPoint> {
-        self.tip.get(height)
-    }
-
-    /// Iterate checkpoints over a height range.
-    ///
-    /// Note that we always iterate checkpoints in reverse height order (iteration starts at tip
-    /// height).
-    ///
-    /// This is a shorthand for calling [`CheckPoint::range`] on the [`tip`].
-    ///
-    /// [`tip`]: LocalChain::tip
-    pub fn range<R>(&self, range: R) -> impl Iterator<Item = CheckPoint>
-    where
-        R: RangeBounds<u32>,
-    {
-        self.tip.range(range)
-    }
 }
 
 /// The [`ChangeSet`] represents changes to [`LocalChain`].
-#[derive(Debug, Default, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
-pub struct ChangeSet {
+pub struct ChangeSet<D = BlockHash> {
     /// Changes to the [`LocalChain`] blocks.
     ///
     /// The key represents the block height, and the value either represents added a new
     /// [`CheckPoint`] (if [`Some`]), or removing a [`CheckPoint`] (if [`None`]).
-    pub blocks: BTreeMap<u32, Option<BlockHash>>,
+    pub blocks: BTreeMap<u32, Option<D>>,
 }
 
-impl Merge for ChangeSet {
+impl<D> Default for ChangeSet<D> {
+    fn default() -> Self {
+        ChangeSet {
+            blocks: BTreeMap::default(),
+        }
+    }
+}
+
+impl<D> Merge for ChangeSet<D> {
     fn merge(&mut self, other: Self) {
         Merge::merge(&mut self.blocks, other.blocks)
     }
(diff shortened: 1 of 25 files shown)


crates/core/src/checkpoint.rs
```diff
@@ -89,48 +181,46 @@ impl CheckPoint {
...
+        Self(Arc::new(CPInner {
+            height,
+            data,
+            prev: None,
+        }))
     }
 
-    /// Get the [`BlockId`] of the checkpoint.
-    pub fn block_id(&self) -> BlockId {
-        self.0.block
```
evanlinjin: This can be a definition on `CheckPoint<B: ToBlockHash>`.

crates/core/src/checkpoint.rs
```diff
@@ -50,36 +104,74 @@ impl CheckPoint {
...
+        for block in blockdata {
+            curr = curr.push(block).map_err(|_| self.clone())?;
+        }
+        Ok(curr)
+    }
+
+    /// Get the block hash of the checkpoint.
+    pub fn hash(&self) -> BlockHash {
+        self.0.data
+    }
```
evanlinjin: Have this for `B: ToBlockHash`.

crates/core/src/checkpoint.rs
```diff
@@ -10,29 +10,83 @@ use crate::BlockId;
...
+pub trait ToBlockHash {
+    /// TODO: to_blockhash doc
+    fn to_blockhash(&self) -> BlockHash;
+}
+
+impl<B: ToBlockHash> ToBlockHash for CheckPoint<B> {
+    fn to_blockhash(&self) -> BlockHash {
+        self.0.data.to_blockhash()
+    }
+}
```
evanlinjin: I don't think this is a good idea? We don't want nested `CheckPoint`s.

crates/core/src/checkpoint.rs
```diff
@@ -10,29 +10,83 @@ use crate::BlockId;
...
-    prev: Option<Arc<CPInner>>,
+    prev: Option<Arc<CPInner<B>>>,
 }
 
-impl PartialEq for CheckPoint {
+/// TODO: ToBlockHash doc
+pub trait ToBlockHash {
+    /// TODO: to_blockhash doc
+    fn to_blockhash(&self) -> BlockHash;
+}
```
evanlinjin: impl this for `BlockHash`.

crates/core/src/checkpoint.rs
```diff
@@ -164,6 +245,40 @@ impl CheckPoint {
             })
     }
 
+    /// This method tests for `self` and `other` to have equal internal pointers.
+    pub fn eq_ptr(&self, other: &Self) -> bool {
+        Arc::as_ptr(&self.0) == Arc::as_ptr(&other.0)
+    }
```
evanlinjin, line 190: This can be defined for `CheckPoint<B: ToBlockHash>`.

crates/core/src/checkpoint.rs
```diff
@@ -164,6 +245,40 @@ impl CheckPoint {
...
+}
+
+impl<B> CheckPoint<B>
+where
+    B: Copy + core::fmt::Debug + ToBlockHash + From<BlockHash>,
```
evanlinjin:
```suggestion
    B: Copy + core::fmt::Debug + ToBlockHash,
```

Why do we need `From<BlockHash>`?

crates/core/src/checkpoint.rs
```diff
@@ -10,29 +10,83 @@ use crate::BlockId;
...
+    fn to_blockhash(&self) -> BlockHash {
+        *self
+    }
+}
+
+impl ToBlockHash for Header {
+    fn to_blockhash(&self) -> BlockHash {
+        let mut bytes = vec![];
+        self.consensus_encode(&mut bytes).unwrap_or_default();
+        BlockHash::hash(&bytes)
```
evanlinjin: https://docs.rs/bitcoin/latest/bitcoin/block/struct.Header.html#method.block_hash

crates/core/src/checkpoint.rs
```diff
@@ -10,29 +10,83 @@ use crate::BlockId;
...
+
+impl<B> PartialEq for CheckPoint<B>
+where
+    B: Copy + core::cmp::PartialEq,
+{
     fn eq(&self, other: &Self) -> bool {
-        let self_cps = self.iter().map(|cp| cp.block_id());
-        let other_cps = other.iter().map(|cp| cp.block_id());
+        let self_cps = self.iter().map(|cp| *cp.inner());
+        let other_cps = other.iter().map(|cp| *cp.inner());
```
evanlinjin: Technically, we can't have two blockhashes that are the same. However, let's not change the behavior. I.e. also compare the heights along the hashes.

crates/core/src/checkpoint.rs
```diff
@@ -90,47 +169,36 @@ impl CheckPoint {
...
-
-    /// Get the [`BlockId`] of the checkpoint.
-    pub fn block_id(&self) -> BlockId {
-        self.0.block
+impl<B> CheckPoint<B>
+where
+    B: Copy,
+{
+    /// Get reference to the inner type.
+    pub fn inner(&self) -> &B {
```
evanlinjin: Let's rename this to `data`.

crates/core/src/checkpoint.rs
```diff
@@ -10,29 +10,83 @@ use crate::BlockId;
...
-pub struct CheckPoint(Arc<CPInner>);
+pub struct CheckPoint<B = BlockHash>(Arc<CPInner<B>>);
 
 /// The internal contents of [`CheckPoint`].
 #[derive(Debug, Clone)]
-struct CPInner {
-    /// Block id (hash and height).
+struct CPInner<B> {
+    /// Block data.
     block: BlockId,
```
evanlinjin: I think it makes sense to rename this to either `block_id` or `id`.

crates/core/src/checkpoint.rs
evanlinjin: So I think for now, just compare the block ids (like the old behavior). However, I'm wondering whether in the future, we want to be able to insert changeable data (in which case, we need to compare the `data` field).

crates/core/src/checkpoint.rs
```diff
@@ -10,29 +10,83 @@ use crate::BlockId;
...
+        let self_cps = self.iter().map(|cp| *cp.inner());
+        let other_cps = other.iter().map(|cp| *cp.inner());
         self_cps.eq(other_cps)
     }
 }
 
-impl CheckPoint {
-    /// Construct a new base block at the front of a linked list.
+impl CheckPoint<BlockHash> {
+    /// Construct a new [`CheckPoint`] at the front of a linked list.
```
evanlinjin: The keyword in the docs is "base". Let's keep it!

crates/core/src/checkpoint.rs
```diff
@@ -10,29 +10,83 @@ use crate::BlockId;
...
+        }))
+    }
+
+    /// Construct a checkpoint from the given `header` and block `height`.
+    ///
+    /// If `header` is of the genesis block, the checkpoint won't have a [`prev`] node. Otherwise,
+    /// we return a checkpoint linked with the previous block.
+    ///
+    /// [`prev`]: CheckPoint::prev
+    pub fn from_header(header: &bitcoin::block::Header, height: u32) -> Self {
```
evanlinjin: This method name is going to be confusing, since we can also create a `CheckPoint<Header>` now. I think we should deprecate this method with additional docs stating that if you want to create `CheckPoint<Header>`, use `from_data`. We should introduce a new method with a better name (and same behavior). `blockhash_checkpoint_from_header` may be clear?

crates/core/src/checkpoint.rs
```diff
@@ -50,36 +104,61 @@ impl CheckPoint {
...
-    /// we return a checkpoint linked with the previous block.
+    /// Returns an `Err(self)` if there is block which does not have a greater height than the
+    /// previous one.
+    pub fn extend(self, blockdata: impl IntoIterator<Item = BlockId>) -> Result<Self, Self> {
+        let mut curr = self.clone();
+        for block in blockdata {
+            curr = curr.push(block).map_err(|_| self.clone())?;
+        }
+        Ok(curr)
+    }
```
evanlinjin: Reuse logic from `extend_data`.

crates/core/src/checkpoint.rs
```diff
@@ -50,36 +104,61 @@ impl CheckPoint {
...
+            cp = cp.prev().expect("will break before genesis block");
         };
 
-        CheckPoint::new(prev_block_id)
-            .push(this_block_id)
-            .expect("must construct checkpoint")
+        let new_cp = core::iter::once(block_id).chain(tail.into_iter().rev());
+
+        base.extend(new_cp).expect("tail is in order")
     }
```
evanlinjin: Reuse logic from `insert_data`.

crates/core/src/checkpoint.rs
```diff
@@ -10,29 +10,83 @@ use crate::BlockId;
...
-    /// Construct a new base block at the front of a linked list.
+impl CheckPoint<BlockHash> {
+    /// Construct a new [`CheckPoint`] at the front of a linked list.
     pub fn new(block: BlockId) -> Self {
-        Self(Arc::new(CPInner { block, prev: None }))
+        Self(Arc::new(CPInner {
+            block,
+            data: block.hash,
+            prev: None,
+        }))
```
evanlinjin: Reuse logic from `from_data`.

crates/chain/tests/test_local_chain.rs
```diff
@@ -380,6 +381,88 @@ fn local_chain_insert_block() {
     }
 }
 
+#[test]
+fn local_chain_insert_header() {
```
evanlinjin, line 383: What do you think about constructing an array of consecutive headers first? 

Then test cases can be constructed from a subset from this array.

The benefits of this is more correctness. The block hashes line up. The header-to-insert can be constructed from the previous block hash.

crates/core/src/checkpoint.rs
```diff
@@ -49,40 +57,51 @@ impl Drop for CPInner {
...
-        let mut blocks = block_ids.into_iter();
-        let mut acc = CheckPoint::new(blocks.next().ok_or(None)?);
-        for id in blocks {
-            acc = acc.push(id).map_err(Some)?;
-        }
-        Ok(acc)
+    /// If `header` is of the genesis block, the checkpoint won't have a `prev` node. Otherwise,
+    /// we return a checkpoint linked with the previous block.
+    #[deprecated(
+        since = "0.1.1",
```
oleonardolima: ❓ Is this version correct ?

crates/core/src/checkpoint.rs
```diff
@@ -110,55 +129,90 @@ impl CheckPoint {
...
+                .map(|block| (block.height, block.hash)),
+        )
+    }
+
+    /// Inserts `block_id` at its height within the chain.
+    ///
+    /// The effect of `insert` depends on whether a height already exists. If it doesn't the
+    /// `block_id` we inserted and all pre-existing blocks higher than it will be re-inserted after
+    /// it. If the height already existed and has a conflicting block hash then it will be purged
+    /// along with all block followin it. The returned chain will have a tip of the `block_id`
```
oleonardolima:
``` suggestion
/// along with all blocks following it. The returned chain will have a tip of the `block_id`
```

crates/core/src/checkpoint.rs
LagginTimes: Thanks for pointing this out! The version number for `bdk_core` has been updated since this PR was created. This will need to be updated as well.

crates/core/src/checkpoint.rs
```diff
@@ -1,32 +1,40 @@
...
-#[derive(Debug, Clone)]
-struct CPInner {
-    /// Block id (hash and height).
-    block: BlockId,
+#[derive(Debug)]
+struct CPInner<B> {
+    /// Block data.
+    block_id: BlockId,
+    /// Data.
+    data: B,
```
ValuedMammal: When `B` is a blockhash, won't this be storing the same block hash twice? 🤔

nymius: +1 on this. Maybe `data: Option<B>` or just keep the height as part of `CPInner`?
```rust
struct CPInner<B> {
  height: u32,
  data: B,
  prev: Option<Arc<CPinner<B>>>,
}
```

crates/core/src/checkpoint.rs
```diff
@@ -49,40 +57,51 @@ impl Drop for CPInner {
...
+}
+
+impl<B> PartialEq for CheckPoint<B>
+where
+    B: core::cmp::PartialEq,
+{
     fn eq(&self, other: &Self) -> bool {
-        let self_cps = self.iter().map(|cp| cp.block_id());
-        let other_cps = other.iter().map(|cp| cp.block_id());
+        let self_cps = self.iter().map(|cp| cp.0.block_id);
```
nymius: Out of curiosity, Is there any reason to prefer the field access over the method call?

crates/core/src/checkpoint.rs
evanlinjin: Let's look at this from the API/performance perspective.

I think the "killer feature" of this PR is to be able to put Headers in Checkpoints.

Ideally, we want something like this: `CheckPoint<Header>`. However, with the proposed changes, it will mean we will be hashing the `Header` every time we want the block hash from the checkpoint (which will be extremely frequent).

To fix this, we can introduce another data type `HeaderWithHash` - which, as the name suggests, contains the `Header` and the `BlockHash` together. I'm honestly not a fan of this as it becomes clunky.

Since memory is cheap, I think it is okay to duplicate the `BlockHash`.

evanlinjin: If we really don't want to duplicate the `BlockHash`, another solution would be to use an enum or option internally. It's honestly not pretty and we need to worry about unwraps and how to store the data. Not a fan.

crates/core/src/checkpoint.rs
evanlinjin: No.

notmandatory:
Since this one is still having design discussions can I move it out of the 1.1 milestone?

crates/core/src/checkpoint.rs
ValuedMammal: `HeaderWithHash` sounds reasonable. (or `HashedHeader`)

notmandatory:

rustaceanrob:
Dropping in to note that this would be incredible useful for `bdk_kyoto`. It is likely that a `LocalChain` generic over `Header` would allow the CBF system to simply use the `LocalChain` the chain backend, and the persistence may be shared between a `Wallet` and the node.

luisschwab:
> Dropping in to note that this would be incredible useful for bdk_kyoto. It is likely that a LocalChain generic over Header would allow the CBF system to simply use the LocalChain the chain backend, and the persistence may be shared between a Wallet and the node.

+1 for this feature, `bdk_floresta` could share persistence with `bdk_wallet` as well.

ValuedMammal:
At a high level I think the API would be cleaner if the generic methods had the simpler names (new, push, extend, insert) and the old methods were renamed `push_block_id`, `extend_block_ids`, etc.

`CheckPoint::data` method can be two methods:

```rust
fn data_ref(&self) -> &B {}

fn data(&self) -> B where B: Clone {} // or Copy
```

`LocalChain` should have more of a unified interface. Most of the methods should be generic and the only methods under `impl LocalChain<BlockHash> {}` should be the ones that only make sense in the old design (from_genesis_hash, insert_block, etc). Having two versions of `apply_changeset` and `_check_changeset_is_applied` isn't very helpful.

Remaining questions:
1) Will it be possible to migrate an existing sqlite database to one that persists a local chain of headers, or will it require adding a new SQL table?
2) Are there plans to enable syncing a local chain of headers using bdk_electrum / esplora?

Edit: I added my review with some API suggestions in LagginTimes/bdk#5

crates/core/src/checkpoint.rs
evanlinjin: `CheckPoint<HeaderWithHash>` is quite unpleasing. I guess it will be a win if there was no duplicate `BlockHash` internally. I've changed my mind and I think an internal option solution is best.

```rust
pub trait ToBlockHash {
    fn to_blockhash(&self) -> BlockHash;
    /// Whether this type needs the `BlockHash` to be cached.
    /// For example, `Header` will have this as `true`.
    fn needs_cache() -> bool;
}

/// Internal struct
struct CPInner<B> {
    height: u32,
    data: B,
    cached_block_hash: Option<BlockHash>,
    /* missing fields */
}

impl<B: ToBlockHash> CPInner {
    /// Example construction only. This is missing `prev`.
    fn new(height: u32, data: B) -> Self {
        let cached_block_hash = if B::needs_cache() {
            Some(data.to_block_hash())
        } else {
            None
        };
        return Self { height, data, cached_block_hash }
    }

    fn block_hash(&self) -> BlockHash {
        self.cached_block_hash.unwrap_or_else(|| self.data.to_block_hash())
    }
}
```

evanlinjin: Rethinking about this, I don't think there is a point in doing what I suggested in the previous comment.

I misunderstood how enums work in Rust. Enums are as large as their largest variant in memory. So even if `cached_block_hash` is `None`, it will still be taking up 32 bytes (+ some more). To fix this we can throw the cached hash onto the heap with `Box`, but I don't think overloading the heap to save `32` bytes is worth it.

Now the question is whether we do `CheckPoint<Header>` or `CheckPoint<HeaderWithHash>`. I no longer have a strong opinion on this.

ValuedMammal: I agree with `CheckPoint<Header>` and the inner block id field can stay.

evanlinjin:
@LagginTimes is this ready to review again? Remember to mark it as so and request reviewers, otherwise no one will know about it.

evanlinjin:
What was the discussion around breaking changes? Were we okay to break the API?

LagginTimes:
> What was the discussion around breaking changes? Were we okay to break the API?

Iirc the consensus was for this to be a breaking change to be a cleaner API.

crates/chain/src/local_chain.rs
```diff
@@ -53,12 +56,24 @@ fn apply_changeset_to_checkpoint(
 }
 
 /// This is a local implementation of [`ChainOracle`].
-#[derive(Debug, Clone, PartialEq)]
-pub struct LocalChain {
-    tip: CheckPoint,
+#[derive(Debug, Clone)]
+pub struct LocalChain<H = BlockHash> {
```
evanlinjin: I would use `B` (as in block) or `D` (as in data) for the type generic instead of `H` (which I'm assuming stands for header, but we aren't necessarily using headers here).

crates/chain/src/local_chain.rs
```diff
@@ -83,18 +98,13 @@ impl ChainOracle for LocalChain {
...
-impl LocalChain {
-    /// Get the genesis hash.
-    pub fn genesis_hash(&self) -> BlockHash {
-        self.tip.get(0).expect("genesis must exist").hash()
-    }
-
+impl LocalChain<BlockHash> {
     /// Construct [`LocalChain`] from genesis `hash`.
     #[must_use]
     pub fn from_genesis_hash(hash: BlockHash) -> (Self, ChangeSet) {
```
evanlinjin: I would remove this and have `from_genesis(data: B)` instead which would work with all types.

crates/chain/src/local_chain.rs
```diff
@@ -83,18 +98,13 @@ impl ChainOracle for LocalChain {
     }
 }
 
-impl LocalChain {
-    /// Get the genesis hash.
-    pub fn genesis_hash(&self) -> BlockHash {
-        self.tip.get(0).expect("genesis must exist").hash()
-    }
-
+impl LocalChain<BlockHash> {
```
evanlinjin, line 98: Let's make these following methods generic for all `H`s.

```rust
impl LocalChain<D> {
    pub fn from_genesis(data: D) -> (Self, ChangeSet<D>);
    pub fn from_changeset(changeset: ChangeSet<D>) -> Result<Self>;
    pub fn from_blocks(blocks: BTreeMap<u32, D>) -> Result<Self>;
    pub fn apply_update(&mut self, update: CheckPoint<D>) -> Result<ChangeSet>;
```

crates/chain/src/local_chain.rs
```diff
@@ -265,12 +245,11 @@ impl LocalChain {
...
-        let old_tip = self.tip.clone();
-        let new_tip = apply_changeset_to_checkpoint(old_tip, changeset)?;
-        self.tip = new_tip;
-        debug_assert!(self._check_changeset_is_applied(changeset));
-        Ok(())
+    pub fn apply_blockhash_changeset(
+        &mut self,
+        changeset: &ChangeSet,
+    ) -> Result<(), MissingGenesisError> {
+        self.apply_changeset(changeset)
```
evanlinjin: This doesn't look helpful as `.apply_changeset` does the exact same thing.

crates/chain/src/local_chain.rs
```diff
@@ -279,29 +258,7 @@ impl LocalChain {
     ///
     /// Replacing the block hash of an existing checkpoint will result in an error.
     pub fn insert_block(&mut self, block_id: BlockId) -> Result<ChangeSet, AlterCheckPointError> {
```
evanlinjin: This method is no longer useful as `.insert_data` does the same thing. Just rename `.insert_data` to `.insert_block`.

crates/chain/src/local_chain.rs
```diff
@@ -314,7 +271,7 @@ impl LocalChain {
     /// This will fail with [`MissingGenesisError`] if the caller attempts to disconnect from the
     /// genesis block.
     pub fn disconnect_from(&mut self, block_id: BlockId) -> Result<ChangeSet, MissingGenesisError> {
```
evanlinjin: This method can be made to be generic over `H`.

crates/chain/src/local_chain.rs
```diff
@@ -355,36 +312,41 @@ impl LocalChain {
     }
 
     /// Iterate over checkpoints in descending height order.
-    pub fn iter_checkpoints(&self) -> CheckPointIter {
+    pub fn iter_checkpoints(&self) -> CheckPointIter<BlockHash> {
```
evanlinjin: Can be made to be generic over `H`.

crates/chain/src/local_chain.rs
```diff
@@ -355,36 +312,41 @@ impl LocalChain {
...
-                    curr_cp = query_cp;
-                }
-                None => {
-                    if exp_hash.is_some() {
-                        return false;
-                    }
-                }
-            }
+        self._check_data_changeset_is_applied(changeset)
+    }
```
evanlinjin: Not needed anymore.

crates/chain/src/local_chain.rs
```diff
@@ -396,26 +358,135 @@ impl LocalChain {
...
+        }
+        true
+    }
+
+    /// Insert data into a [`LocalChain`].
+    ///
+    /// # Errors
+    ///
+    /// Replacing the block hash of an existing checkpoint will result in an error.
+    pub fn insert_data(
```
evanlinjin: To be consistent, I think we can rename this to `insert_block`.

crates/chain/src/local_chain.rs
```diff
@@ -396,26 +358,135 @@ impl LocalChain {
...
+                ),
+            })?;
+        Ok(changeset)
+    }
+
+    /// Constructs a [`LocalChain`] from a [`BTreeMap`] of height and data.
+    ///
+    /// The [`BTreeMap`] enforces the height order. However, the caller must ensure the blocks are
+    /// all of the same chain.
+    pub fn from_data(blocks: BTreeMap<u32, H>) -> Result<Self, MissingGenesisError> {
```
evanlinjin: I think it makes sense to rename this to `from_blocks`.

crates/core/src/checkpoint.rs
```diff
@@ -49,40 +57,51 @@ impl Drop for CPInner {
...
-            acc = acc.push(id).map_err(Some)?;
-        }
-        Ok(acc)
+    /// If `header` is of the genesis block, the checkpoint won't have a `prev` node. Otherwise,
+    /// we return a checkpoint linked with the previous block.
+    #[deprecated(
+        since = "0.5.0",
+        note = "Please use [`CheckPoint::blockhash_checkpoint_from_header`] instead. To create a CheckPoint<Header>, please use [`CheckPoint::from_data`]."
+    )]
+    pub fn from_header(header: &bitcoin::block::Header, height: u32) -> Self {
```
evanlinjin: We can remove this now.

crates/core/src/checkpoint.rs
```diff
@@ -1,32 +1,40 @@
...
 use crate::BlockId;
 
 /// A checkpoint is a node of a reference-counted linked list of [`BlockId`]s.
 ///
 /// Checkpoints are cheaply cloneable and are useful to find the agreement point between two sparse
 /// block chains.
-#[derive(Debug, Clone)]
-pub struct CheckPoint(Arc<CPInner>);
+#[derive(Debug)]
+pub struct CheckPoint<H = Header>(Arc<CPInner<H>>);
```
evanlinjin: For now, I think the default for `H` should still be `BlockHash`. Also, I would prefer `D` over `H` since it's not necessarily headers that go here.

crates/core/src/checkpoint.rs
```diff
@@ -49,40 +57,51 @@ impl Drop for CPInner {
...
+    fn to_blockhash(&self) -> BlockHash {
+        *self
+    }
+}
+
+impl ToBlockHash for Header {
+    fn to_blockhash(&self) -> BlockHash {
+        self.block_hash()
+    }
+}
```
evanlinjin: Let's have one more impl for `Block`.

crates/core/src/checkpoint.rs
```diff
@@ -49,40 +57,51 @@ impl Drop for CPInner {
     }
 }
 
-impl PartialEq for CheckPoint {
+/// Trait that converts [`CheckPoint`] `data` to [`BlockHash`].
+pub trait ToBlockHash {
```
evanlinjin: We should add to the docs:

```md
Implementations of `ToBlockHash` must consist exclusively of the data to be hashed — do not include any extraneous fields.
```

Otherwise, additional change-detection logic will be required to distinguish `D` instances that share the same height and hash. I'm pretty sure no one wants this feature?

crates/core/src/checkpoint.rs
```diff
@@ -49,40 +57,51 @@ impl Drop for CPInner {
...
         let other_cps = other.iter().map(|cp| cp.block_id());
         self_cps.eq(other_cps)
     }
 }
 
-impl CheckPoint {
-    /// Construct a new base block at the front of a linked list.
+impl CheckPoint<BlockHash> {
+    /// Construct a new base [`CheckPoint`] at the front of a linked list.
     pub fn new(block: BlockId) -> Self {
```
evanlinjin: This method can be changed to be generic: `new(height: u32, data: B)`. `from_data` can be removed.

crates/core/src/checkpoint.rs
```diff
@@ -106,59 +125,105 @@ impl CheckPoint {
...
+    /// This method will error if any of the follow occurs:
+    ///
+    /// - The `blocks` iterator is empty, in which case, the error will be `None`.
+    /// - The `blocks` iterator is not in ascending height order.
+    /// - The `blocks` iterator contains multiple [`BlockId`]s of the same height.
+    ///
+    /// The error type is the last successful checkpoint constructed (if any).
+    pub fn from_block_ids(
+        block_ids: impl IntoIterator<Item = BlockId>,
+    ) -> Result<Self, Option<Self>> {
```
evanlinjin: Let's rework this method to be generic over `D`:

```suggestion
    pub fn from_blocks(
        blocks: impl IntoIterator<Item = (u32, D)>,
    ) -> Result<Self, Option<Self>> {
```

crates/core/src/checkpoint.rs
```diff
@@ -106,59 +125,105 @@ impl CheckPoint {
...
+        self,
+        blockdata: impl IntoIterator<Item = BlockId>,
+    ) -> Result<Self, Self> {
+        self.extend(
+            blockdata
+                .into_iter()
+                .map(|block| (block.height, block.hash)),
+        )
+    }
+
```
evanlinjin: Let's remove this as `extend` does the same thing.

crates/core/src/checkpoint.rs
```diff
@@ -106,59 +125,105 @@ impl CheckPoint {
...
+    ///
+    /// The effect of `insert` depends on whether a height already exists. If it doesn't the
+    /// `block_id` we inserted and all pre-existing blocks higher than it will be re-inserted after
+    /// it. If the height already existed and has a conflicting block hash then it will be purged
+    /// along with all blocks following it. The returned chain will have a tip of the `block_id`
+    /// passed in. Of course, if the `block_id` was already present then this just returns `self`.
+    #[must_use]
+    pub fn insert_block_id(self, block_id: BlockId) -> Self {
+        self.insert(block_id.height, block_id.hash)
+    }
```
evanlinjin: Remove.

crates/core/src/checkpoint.rs
```diff
@@ -106,59 +125,105 @@ impl CheckPoint {
...
+
+    /// Puts another checkpoint onto the linked list representing the blockchain.
+    ///
+    /// Returns an `Err(self)` if the block you are pushing on is not at a greater height that the one you
+    /// are pushing on to.
+    pub fn push_block_id(self, block: BlockId) -> Result<Self, Self> {
+        self.push(block.height, block.hash)
+    }
```
evanlinjin: Remove.

crates/core/src/spk_client.rs
```diff
@@ -387,14 +388,14 @@ impl<I> SyncRequest<I> {
 /// See also [`SyncRequest`].
 #[must_use]
 #[derive(Debug)]
-pub struct SyncResponse<A = ConfirmationBlockTime> {
+pub struct SyncResponse<B = BlockHash, A = ConfirmationBlockTime> {
```
evanlinjin: I know this is a breaking change, but introducing a generic parameter before one that already exists can be really annoying. Introduce it at the end please.

evanlinjin (changes requested):
Thanks for pushing this forward.

Please double check (brownie points for triple check) over all the methods of a struct to see whether they are coherent, whether any more methods can be made generic and whether the method names and generic parameter names are coherent and make sense.

Let's make all generic parameter names for block data be consistent across the board. I'm in favor of `B`.

crates/chain/src/local_chain.rs
```diff
@@ -264,43 +187,183 @@ impl LocalChain {
...
-                update_hash: changeset.blocks.get(&0).cloned().flatten(),
+                update_hash: Some(
+                    changeset
+                        .blocks
+                        .get(&0)
+                        .cloned()
+                        .flatten()
+                        .unwrap()
+                        .to_blockhash(),
+                ),
```
ValuedMammal: I spotted an `unwrap`, not sure if it was intentional.
```suggestion
                update_hash: changeset
                    .blocks
                    .get(&0)
                    .cloned()
                    .flatten()
                    .map(|d| d.to_blockhash()),
```

ValuedMammal:
Tests are passing for me.

But I think we should still have the function `_check_changeset_is_applied`. Edit: unless I misread an earlier comment saying it is no longer needed. I tested with and without it, plus the unit tests should be sufficient, so feel free to ignore.

<details>
  <summary>diff</summary>

```diff
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -270,8 +270,9 @@ where
         };
 
         let (mut chain, _) = Self::from_genesis(genesis_data);
         chain.apply_changeset(&changeset)?;
+        debug_assert!(chain._check_changeset_is_applied(&changeset));
         Ok(chain)
     }
 
     /// Construct a [`LocalChain`] from a given `checkpoint` tip.
@@ -303,16 +304,18 @@ where
         update: CheckPoint<D>,
     ) -> Result<ChangeSet<D>, CannotConnectError> {
         let (new_tip, changeset) = merge_chains(self.tip.clone(), update)?;
         self.tip = new_tip;
+        debug_assert!(self._check_changeset_is_applied(&changeset));
         Ok(changeset)
     }
 
     /// Apply the given `changeset`.
     pub fn apply_changeset(&mut self, changeset: &ChangeSet<D>) -> Result<(), MissingGenesisError> {
         let old_tip = self.tip.clone();
         let new_tip = apply_changeset_to_checkpoint(old_tip, changeset)?;
         self.tip = new_tip;
+        debug_assert!(self._check_changeset_is_applied(changeset));
         Ok(())
     }
 
     /// Derives an initial [`ChangeSet`], meaning that it can be applied to an empty chain to
@@ -401,8 +404,30 @@ where
             None => return Ok(ChangeSet::default()),
         };
         Ok(changeset)
     }
+
+    fn _check_changeset_is_applied(&self, changeset: &ChangeSet<D>) -> bool {
+        let mut cur = self.tip.clone();
+        for (&exp_height, exp_data) in changeset.blocks.iter().rev() {
+            match cur.get(exp_height) {
+                Some(cp) => {
+                    if cp.height() != exp_height
+                        || Some(cp.hash()) != exp_data.map(|d| d.to_blockhash())
+                    {
+                        return false;
+                    }
+                    cur = cp;
+                }
+                None => {
+                    if exp_data.is_some() {
+                        return false;
+                    }
+                }
+            }
+        }
+        true
+    }
 }

```
</details>

ValuedMammal (approved):
ACK e1e19599e04225b916afd223d283f4437cb88e26

crates/core/src/checkpoint.rs
```diff
@@ -192,7 +163,56 @@ impl CheckPoint {
...
+            },
+            data,
+            prev: None,
+        }))
+    }
+
+    /// Construct from an iterator of block data.
+    ///
+    /// Returns `Err(None)` if `blocks` doesn't yield any data. If the blocks are not in ascending
+    /// height order, then returns the last [`CheckPoint`] that would have been extended.
```
ValuedMammal:
```suggestion
    /// height order, then returns an `Err(..)` containing the last checkpoint that would have been extended.
```

ValuedMammal:
ACK dabb869f77b6cb548e4a82c15b250df460695b24

crates/core/src/checkpoint.rs
```diff
@@ -49,116 +58,78 @@ impl Drop for CPInner {
...
-impl PartialEq for CheckPoint {
-    fn eq(&self, other: &Self) -> bool {
-        let self_cps = self.iter().map(|cp| cp.block_id());
-        let other_cps = other.iter().map(|cp| cp.block_id());
-        self_cps.eq(other_cps)
-    }
+/// Trait that converts [`CheckPoint`] `data` to [`BlockHash`].
+///
+/// Implementations of [`ToBlockHash`] must consist exclusively of the data to be hashed — do not
+/// include any extraneous fields.
```
ValuedMammal: ❓ I don't fully understand what this is saying. Is there an example of a type which doesn't have a valid implementation of `ToBlockHash`?

crates/core/src/checkpoint.rs
```diff
@@ -49,116 +58,78 @@ impl Drop for CPInner {
...
-
-        let prev_block_id = BlockId {
-            height: prev_height,
-            hash: header.prev_blockhash,
-        };
+impl ToBlockHash for Block {
+    fn to_blockhash(&self) -> BlockHash {
+        self.block_hash()
+    }
+}
```
ValuedMammal: Because the `CheckPoint` API constrains `D: Copy`, I don't see how one could ever have `CheckPoint<Block>` at least in the current implementation. But if we go the way of #1997 I'm guessing we could relax the `Copy` constraint to `Clone` instead?

crates/core/src/spk_client.rs
```diff
@@ -158,15 +159,15 @@ impl<I> SyncRequestBuilder<I> {
...
+    /// let _request: SyncRequest<u32, BlockHash> = SyncRequest::builder()
     ///     .spks_with_indexes(newly_revealed_spks)
     ///     .build();
     ///
     /// // Sync all revealed spks in the indexer. This time, spks may be derived from different
     /// // keychains. Each spk will be indexed with `(&'static str, u32)` where `&'static str` is
     /// // the keychain identifier and `u32` is the derivation index.
     /// let all_revealed_spks = indexer.revealed_spks(..);
-    /// let _request = SyncRequest::builder()
+    /// let _request: SyncRequest<(&str, u32), BlockHash> = SyncRequest::builder()
```
ValuedMammal, line 170: nit: The comment above refers to `(&'static str, u32)` (not `&str`). Feel free to ignore

ValuedMammal (commented):
ACK 0a39fa76e777204e9c774e5fd40c446605d2783e pending a CI fix.

I left comments mostly for discussion. Also there was a comment https://github.com/bitcoindevkit/bdk/pull/1582#pullrequestreview-2846419311 suggesting to use `B` instead of `D`. I don't have much preference though.

notmandatory:
I haven't done a deep review but I support merging this now that we're done with the `bdk_wallet` 2.1 release.

crates/chain/src/local_chain.rs
```diff
@@ -264,43 +187,183 @@ impl LocalChain {
...
+        let mut changeset = ChangeSet::<D>::default();
+        changeset.blocks.insert(height, Some(data));
         self.apply_changeset(&changeset)
             .map_err(|_| AlterCheckPointError {
                 height: 0,
                 original_hash: self.genesis_hash(),
-                update_hash: changeset.blocks.get(&0).cloned().flatten(),
+                update_hash: changeset
+                    .blocks
+                    .get(&0)
```
oleonardolima, line 363: ~~question: I'm wondering about this error, is it correct to always get the height 0 ?~~

scratch that, I didn't notice the upstream error: `MissingGenesisError` due to a lack of valid PoA and genesis on the extension changeset.

crates/core/src/checkpoint.rs
```diff
@@ -192,7 +163,57 @@ impl CheckPoint {
...
+        let mut cp = self.clone();
+        for (height, data) in blockdata {
+            cp = cp.push(height, data)?;
+        }
+        Ok(cp)
+    }
+
+    /// Inserts `data` at its `height` within the chain.
     ///
     /// The effect of `insert` depends on whether a height already exists. If it doesn't the
```
oleonardolima: nit: it's not expecting `BlockId` as a parameter anymore, although it's constructed internally for the checkpoint, I think we could update its references in the doc.

oleonardolima (commented):
utACK d4bd7e5ccb598e010808823ffdc3d1e66f468bce

The code looks good to me, I only left a minor ~~question~~ and nit. I didn't test it on a complicated example yet.

ValuedMammal:
ACK 4dce5add0801bdbb52e1997386aed8f3455ff88e

notmandatory (approved):
utACK 4dce5add0801bdbb52e1997386aed8f3455ff88e

I also re-read through the whole history and this one has certainly been a journey so thanks to @LagginTimes and all the reviewers for sticking with it. 

I look forward to the followup PRs to integrate chain client specific data in the checkpoints. I expect custom data will break existing persistence so being able to persist the data in a generic way and having a good persistence test suite such as #2012 will be handy.

oleonardolima (commented):
ACK 4dce5add0801bdbb52e1997386aed8f3455ff88e

evanlinjin:
Looking back through the codebase, I noticed that we've introduced a condition that wasn't there before, and I’m wondering about the best way to handle it.

Since a `Header` contains the previous block hash, two `CheckPoint<Header>` chains can be merged even if their heights differ. For example, if chain A ends at height 10 and chain B only has a single block at height 11, we can still connect them if B’s previous block points to A’s height 10.

---

One idea is to modify `ToBlockHash` to have knowledge of whether the type has the previous blockhash.

```rust
pub trait ToBlockHash {
    /* OTHER METHODS */

    /// Returns `None` if the type has no knowledge of the previous blockhash.
}
```

Then we need to modify `merge_chains` to take it into account.

~~Should we modify `CheckPoint` to allow empty values (knowing the block hash & height, but not the data yet)?~~
  
* **Edit:** On second thought, no. Allowing `CheckPoint` to hold empty data would increase the risk of inconsistent representations—for example, having a `Header` at height 10 while forgetting to include the “ghost checkpoint” at height 9.
* **A better idea?:** I think we should modify the `CheckPointIter` instead to also return "ghost checkpoint"s. The `Iterator::Item` type needs to be changed to `MaybeCheckPoint<D>` or something.
* **Caveat:** Methods that construct/extend/insert-into `CheckPoint`-chains should really error/fail if an inserted block's prevhash does not match the previous value (if it exists).

ValuedMammal:
> Since a `Header` contains the previous block hash, two `CheckPoint<Header>` chains can be merged even if their heights differ. For example, if chain A ends at height 10 and chain B only has a single block at height 11, we can still connect them if B’s previous block points to A’s height 10.
> 
> One idea is to modify `ToBlockHash` to have knowledge of whether the type has the previous blockhash.
> 
> ```rust
> pub trait ToBlockHash {
>     /* OTHER METHODS */
> 
>     /// Returns `None` if the type has no knowledge of the previous blockhash.
>     fn prev_blockhash(&self) -> Option<BlockHash>;
> }
> ```
> 
> Then we need to modify `merge_chains` to take it into account.

Yes in principle I agree with this if the update is strictly extending the parent chain by the next consecutive block. If the type doesn't know the previous hash, like in the case of `CheckPoint<BlockHash>`, then I think we'd still require the existence of a point of agreement.

evanlinjin:
I've started experimenting with my suggestion in https://github.com/bitcoindevkit/bdk/pull/1582#issuecomment-3214838670.

This is the WIP commit: https://github.com/evanlinjin/bdk/commit/8f374ccb639e98836b482d3cb0e704ec64f50aa1. I'm quite happy with the implementation of it (unless proven otherwise).

What is missing:
* Figure out best practice for handling `CheckPointEntry::Backref`s (ghost-checkpoints) that connect.
* Methods that build on `CheckPoint` (i.e. `push`, `from_blocks`, `extend`, `insert`) need to check `Backref`s match.
* Tests.
* More docs.

Feel free to continue this work as you see fit.

---

I'm wondering if we should merge this PR as is, and introduce "ghost-checkpoint"s in a separate PR?

LagginTimes:
> I've started experimenting with my suggestion in [#1582 (comment)](https://github.com/bitcoindevkit/bdk/pull/1582#issuecomment-3214838670).
> 
> This is the WIP commit: [evanlinjin@8f374cc](https://github.com/evanlinjin/bdk/commit/8f374ccb639e98836b482d3cb0e704ec64f50aa1). I'm quite happy with the implementation of it (unless proven otherwise).
> 
> What is missing:
> 
>     * Figure out best practice for handling `CheckPointEntry::Backref`s (ghost-checkpoints) that connect.
> 
>     * Methods that build on `CheckPoint` (i.e. `push`, `from_blocks`, `extend`, `insert`) need to check `Backref`s match.
> 
>     * Tests.
> 
>     * More docs.
> 
> 
> Feel free to continue this work as you see fit.
> 
> I'm wondering if we should merge this PR as is, and introduce "ghost-checkpoint"s in a separate PR?

I think it makes sense to merge #1582 as it is now. The additional work around ghost-checkpoints, backref checks, and more docs/tests can be done in a follow-up PR to keep the scope focused and avoid resetting review. I would be happy to continue the work in a follow-up PR.

evanlinjin:
@LagginTimes I'll be happy with that. Do we have a plan for back-porting the non-merged chain-source fixes (#2005, #2011, #2000)? Note that the `CheckPointEntry` PR may break the API again.

ValuedMammal:
> if chain A ends at height 10 and chain B only has a single block at height 11, we can still connect them if B’s previous block points to A’s height 10.

For this I would look at things we already have such as [`insert_block`](https://github.com/bitcoindevkit/bdk/blob/4dce5add0801bdbb52e1997386aed8f3455ff88e/crates/chain/src/local_chain.rs#L338).

evanlinjin (approved):
ACK 8bc239168a2d81430244ee7633723f7eefe0d0f4

PR #1586, opened 2024-09-02
Change methods of `IndexedTxGraph`/`TxGraph`/`Wallet` that insert txs to be more generic

evanlinjin:
### Description

We want to reuse the `Arc` pointer whenever possible. However, some methods on `TxGraph` and `IndexedTxGraph` that insert transactions take in `&Transaction` or `Transaction` (thus forcing us to create a new `Arc<Transaction>` internally by cloning, even if the input tx is under a `Arc`).

This PR changes these methods to take in a generic parameter, `T: Into<Arc<Transaction>>`, allowing us to reuse the `Arc` pointer whenever possible.

### Notes to the reviewers

Methods that previously took in `Transaction` can be changed to take in `T: Into<Arc<Transaction>>` and be non-breaking (since `Arc<T>` impls `From<T>` automatically). These changes are contained in the first two commits.

Methods that previously took in `&Transaction` will break. However, I think these api changes are small and the improvements are substantial enough to be worth it. These changes are contained in the last commit.

### Changelog notice

* Changed `IndexedTxGraph` methods `insert_tx`, `batch_insert_relevant`, `batch_insert_relevant_unconfirmed`, `batch_insert_unconfirmed` to take in `T: Into<Arc<Transaction>>` instead of `Transaction` or `&Transaction`.
* Changed `TxGraph` method `batch_insert_unconfirmed` to take in `T: Into<Arc<Transaction>>` instead of `Transaction`.
* Changed `Wallet` methods `insert_tx`, `apply_unconfirmed_txs` to take in `T: Into<Arc<Transaction>>` instead of `Transaction` or `&Transaction`.

#### All Submissions:

### DIFF
--- a/crates/chain/src/indexed_tx_graph.rs
+++ b/crates/chain/src/indexed_tx_graph.rs
@@ -2,7 +2,7 @@
 //! [`IndexedTxGraph`] documentation for more.
 use core::fmt::Debug;
 
-use alloc::vec::Vec;
+use alloc::{sync::Arc, vec::Vec};
 use bitcoin::{Block, OutPoint, Transaction, TxOut, Txid};
 
 use crate::{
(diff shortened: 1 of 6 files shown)


oleonardolima (commented):
utACK 87e61212f556bccfaa2a64b74f727cedc0f70e9c

Looks good! I do agree that it definitely helps being able to reuse the `Arc` pointer. It looks fine to me, but I wonder what others think about the breaking changes.

notmandatory (approved):
ACK 87e61212f556bccfaa2a64b74f727cedc0f70e9c

This minor API change also works better with the UniFFI language bindings where all return types are wrapped in an `Arc`.

PR #1594, opened 2024-09-08
Replace trait `AnchorFromBlockPosition` with new struct

### LINKED ISSUE #1266: An alternative approach to `AnchorFromBlockPosition`.
> This was done previously in #1041 but I missed it. Why can't we just have `BlockPosition: Into<A>` bound instead of adding our own conversion trait?

_Originally posted by @LLFourn in https://github.com/bitcoindevkit/bdk/pull/1172#discussion_r1444094122_

@evanlinjin's response:

> It's because there are 3 inputs and doing `A: From<(&Block, BlockId, usize)>` is confusing and we won't know what the `usize` is for.
> 
> An alternative would be to have:
> 
> ```rust
> /// Parameters for constructing an [`Anchor`].
> pub struct AnchorParams<'b> {
>     pub block: &'b Block,
>     pub block_id: BlockId,
>     pub tx_pos: usize,
> }
> 
> fn some_method<A: Anchor + From<AnchorParams>>() { todo!() }
> ```
> 
> What do you think? (`AnchorParams` will be nicely documented of course!)

jirijakes:
### Description

This change replaces a way of creating new generic anchor from block and its height. Previous way was using conversion trait, newly it is a struct and `From`.

Closes #1266.

### Notes to the reviewers

- Removed `tx_pos`: I did not see its relevance (`Anchor` does not give access to it). Was it a relic from the past or something I overlooked?
- I put `BlockPosition` into `tx_data_traits.rs` because I believe it should be next to `Anchor`. But then it's not a module just for traits. What would be better place?

#### All Submissions:

### DIFF
--- a/crates/chain/src/indexed_tx_graph.rs
+++ b/crates/chain/src/indexed_tx_graph.rs
@@ -7,7 +7,7 @@ use bitcoin::{Block, OutPoint, Transaction, TxOut, Txid};
 
 use crate::{
     tx_graph::{self, TxGraph},
-    Anchor, AnchorFromBlockPosition, BlockId, Indexer, Merge,
+    Anchor, BlockId, Indexer, Merge, TxPosInBlock,
 };
 
 /// The [`IndexedTxGraph`] combines a [`TxGraph`] and an [`Indexer`] implementation.
@@ -252,17 +252,17 @@ where
     }
 }
 
-/// Methods are available if the anchor (`A`) implements [`AnchorFromBlockPosition`].
-impl<A: Anchor, I: Indexer> IndexedTxGraph<A, I>
+/// Methods are available if the anchor (`A`) can be created from [`TxPosInBlock`].
+impl<A, I> IndexedTxGraph<A, I>
 where
     I::ChangeSet: Default + Merge,
-    A: AnchorFromBlockPosition,
+    for<'b> A: Anchor + From<TxPosInBlock<'b>>,
+    I: Indexer,
 {
     /// Batch insert all transactions of the given `block` of `height`, filtering out those that are
     /// irrelevant.
     ///
-    /// Each inserted transaction's anchor will be constructed from
-    /// [`AnchorFromBlockPosition::from_block_position`].
+    /// Each inserted transaction's anchor will be constructed using [`TxPosInBlock`].
     ///
     /// Relevancy is determined by the internal [`Indexer::is_tx_relevant`] implementation of `I`.
     /// Irrelevant transactions in `txs` will be ignored.
@@ -280,7 +280,12 @@ where
             changeset.indexer.merge(self.index.index_tx(tx));
             if self.index.is_tx_relevant(tx) {
                 let txid = tx.compute_txid();
-                let anchor = A::from_block_position(block, block_id, tx_pos);
+                let anchor = TxPosInBlock {
+                    block,
+                    block_id,
+                    tx_pos,
+                }
+                .into();
                 changeset.tx_graph.merge(self.graph.insert_tx(tx.clone()));
                 changeset
                     .tx_graph
@@ -292,8 +297,7 @@ where
 
     /// Batch insert all transactions of the given `block` of `height`.
     ///
-    /// Each inserted transaction's anchor will be constructed from
-    /// [`AnchorFromBlockPosition::from_block_position`].
+    /// Each inserted transaction's anchor will be constructed using [`TxPosInBlock`].
     ///
     /// To only insert relevant transactions, use [`apply_block_relevant`] instead.
     ///
@@ -305,7 +309,12 @@ where
         };
         let mut graph = tx_graph::ChangeSet::default();
         for (tx_pos, tx) in block.txdata.iter().enumerate() {
-            let anchor = A::from_block_position(&block, block_id, tx_pos);
+            let anchor = TxPosInBlock {
+                block: &block,
+                block_id,
+                tx_pos,
+            }
+            .into();
             graph.merge(self.graph.insert_anchor(tx.compute_txid(), anchor));
             graph.merge(self.graph.insert_tx(tx.clone()));
         }
--- a/crates/chain/src/tx_data_traits.rs
+++ b/crates/chain/src/tx_data_traits.rs
@@ -100,24 +100,31 @@ impl Anchor for ConfirmationBlockTime {
     }
 }
 
-/// An [`Anchor`] that can be constructed from a given block, block height and transaction position
-/// within the block.
-pub trait AnchorFromBlockPosition: Anchor {
-    /// Construct the anchor from a given `block`, block height and `tx_pos` within the block.
-    fn from_block_position(block: &bitcoin::Block, block_id: BlockId, tx_pos: usize) -> Self;
+/// Set of parameters sufficient to construct an [`Anchor`].
+///
+/// Typically used as an additional constraint on anchor:
+/// `for<'b> A: Anchor + From<TxPosInBlock<'b>>`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct TxPosInBlock<'b> {
+    /// Block in which the transaction appeared.
+    pub block: &'b bitcoin::Block,
+    /// Block's [`BlockId`].
+    pub block_id: BlockId,
+    /// Position in the block on which the transaction appeared.
+    pub tx_pos: usize,
 }
 
-impl AnchorFromBlockPosition for BlockId {
-    fn from_block_position(_block: &bitcoin::Block, block_id: BlockId, _tx_pos: usize) -> Self {
-        block_id
+impl<'b> From<TxPosInBlock<'b>> for BlockId {
+    fn from(pos: TxPosInBlock) -> Self {
+        pos.block_id
     }
 }
 
-impl AnchorFromBlockPosition for ConfirmationBlockTime {
-    fn from_block_position(block: &bitcoin::Block, block_id: BlockId, _tx_pos: usize) -> Self {
+impl<'b> From<TxPosInBlock<'b>> for ConfirmationBlockTime {
+    fn from(pos: TxPosInBlock) -> Self {
         Self {
-            block_id,
-            confirmation_time: block.header.time as _,
+            block_id: pos.block_id,
+            confirmation_time: pos.block.header.time as _,
         }
     }
 }


jirijakes:
@evanlinjin @LLFourn, is this what you meant in #1266?

jirijakes:
I am not sure about the failing test. Is there some flakiness? It ran [alright](https://github.com/jirijakes/bdk/actions/runs/10758395831) in my fork and also my machine. 

```
     Running tests/test_electrum.rs (target/debug/deps/test_electrum-5e966bdb43ca0ef0)

running 4 tests
test test_update_tx_graph_stop_gap ... FAILED
test test_update_tx_graph_without_keychain ... ok
test tx_can_become_unconfirmed_after_reorg ... ok
test test_sync ... ok

failures:

---- test_update_tx_graph_stop_gap stdout ----
Error: expected value at line 1 column 1


failures:
    test_update_tx_graph_stop_gap

error: test failed, to rerun pass `-p bdk_electrum --test test_electrum`
test result: FAILED. 3 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 7.73s

```

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -252,61 +252,62 @@ where
...
 }
 
-/// Methods are available if the anchor (`A`) implements [`AnchorFromBlockPosition`].
-impl<A: Anchor, I: Indexer> IndexedTxGraph<A, I>
+/// Methods are available if the anchor (`A`) can be created from [`BlockPosition`].
+impl<'b, A, I> IndexedTxGraph<A, I>
 where
     I::ChangeSet: Default + Merge,
-    A: AnchorFromBlockPosition,
+    A: Anchor + From<BlockPosition<'b>>,
```
ValuedMammal: Maybe instead of adding a new type this can be `A: Anchor + From<ConfirmationBlockTime>` and below we just construct a `ConfirmationBlockTime` from a block id and header time.

My other thought is that `BlockPosition` might not be the right name if originally it referred to the position of the txid in the `txdata` vector, but that could be up to interpretation.

crates/chain/src/tx_data_traits.rs
```diff
@@ -100,24 +100,42 @@ impl Anchor for ConfirmationBlockTime {
...
+    block: &'b bitcoin::Block,
+    block_id: BlockId,
 }
 
-impl AnchorFromBlockPosition for BlockId {
-    fn from_block_position(_block: &bitcoin::Block, block_id: BlockId, _tx_pos: usize) -> Self {
-        block_id
+impl<'b> BlockPosition<'b> {
+    /// Creates new [`BlockPosition`].
+    pub fn new(block: &'b bitcoin::Block, block_id: BlockId) -> Self {
```
LLFourn: Shouldn't the block hash just come from the block here. Pass the height in only.

crates/chain/src/tx_data_traits.rs
```diff
@@ -100,24 +100,42 @@ impl Anchor for ConfirmationBlockTime {
...
 
-/// An [`Anchor`] that can be constructed from a given block, block height and transaction position
-/// within the block.
-pub trait AnchorFromBlockPosition: Anchor {
-    /// Construct the anchor from a given `block`, block height and `tx_pos` within the block.
-    fn from_block_position(block: &bitcoin::Block, block_id: BlockId, tx_pos: usize) -> Self;
+/// Minimal set of parameters that are needed to construct a generic [`Anchor`].
+/// Typically used as an additional constraint on anchor:
+/// `A: Anchor + From<BlockPosition<'b>>`.
+pub struct BlockPosition<'b> {
```
LLFourn: I think this should be called `BlockAndHeight` and should be documented simply as that. I'm not sure it should be described as the "minimal" set of parameters. It's more like like the maximal amount of detail about the block that could be needed to make an anchor i.e. any anchor should be able to be made from a `BlockPosition`. We could even type constraint that `Anchor: for<'a> From<BlockPosition<'a>>`.

crates/chain/src/indexed_tx_graph.rs
LLFourn: @ValuedMammal The point of the type constraint is that you can get the anchor type from a block and height. We don't want to constrain it so that it must come from a `ConfirmationBlockTime` since not all anchors can do that -- they can all come from a full block + height though.

I agree about the name.

I think the constraint should be `A: for<'b> From<BlockPosition<'b>>` since it should be from any lifetime.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -252,61 +252,62 @@ where
...
     ///
-    /// Each inserted transaction's anchor will be constructed from
-    /// [`AnchorFromBlockPosition::from_block_position`].
+    /// Each inserted transaction's anchor will be constructed using [`BlockPosition`].
     ///
     /// To only insert relevant transactions, use [`apply_block_relevant`] instead.
     ///
     /// [`apply_block_relevant`]: IndexedTxGraph::apply_block_relevant
-    pub fn apply_block(&mut self, block: Block, height: u32) -> ChangeSet<A, I::ChangeSet> {
+    pub fn apply_block(&'b mut self, block: &'b Block, height: u32) -> ChangeSet<A, I::ChangeSet> {
```
LLFourn: This can take a `BlockPosition` (what I would call `BlockAndHeight`).

LLFourn (commented):
Concept ACK.

crates/chain/src/tx_data_traits.rs
evanlinjin: If we were to have a merkle-proof anchor (which I think we should have), don't we also need a tx position in block? I'm not too great with naming, but I'm thinking `TxPosInBlock`?

```rust
pub struct TxPosInBlock<'b> {
    block: &'b Block,
    height: u32,
    tx_pos: usize, 
}

Comment (evanlinjin) [reply crates/chain/src/tx_data_traits.rs]: Also, we are missing some derives. `#[derive(Debug, Clone, Copy, PartialEq, Eq)]` I think makes sense for this type.

File: crates/chain/src/indexed_tx_graph.rs (hunk 4 above)
Comment (evanlinjin) [reply crates/chain/src/indexed_tx_graph.rs]: @LLFourn We shouldn't do this if `BlockPosition` (which I'm thinking should be called `TxPosInBlock`) needs `tx_pos` field.

File: crates/chain/src/indexed_tx_graph.rs (hunk 1 above)
Comment (jirijakes) [reply crates/chain/src/indexed_tx_graph.rs]: > I think the constraint should be A: for<'b> From<BlockPosition<'b>> since it should be from any lifetime.

Good idea!

File: crates/chain/src/tx_data_traits.rs (hunk 2 above)
Comment (jirijakes) [reply crates/chain/src/tx_data_traits.rs]: It's better, yes.

File: crates/chain/src/tx_data_traits.rs (hunk 3 above)
Comment (jirijakes) [reply crates/chain/src/tx_data_traits.rs]: > Also, we are missing some derives. #[derive(Debug, Clone, Copy, PartialEq, Eq)] I think makes sense for this type.

Added, thanks.

**Comment by jirijakes** [comment]
I reflected some of the comments but have not figured out the name yet. I will sleep on it and will see tomorrow if I come with a better one. Until then, I will keep it as draft.

File: crates/chain/src/tx_data_traits.rs (hunk 3 above)
Comment (evanlinjin) [reply crates/chain/src/tx_data_traits.rs]: ~~All fields should be public as well, otherwise external modules/crates can't impl `From<BlockPosition>`.~~

Ohh I see u have a constructor and methods on it. I guess you did it this way instead of... 

```rust
pub struct BlockPosition<'a> {
    pub block: &'a Block,
    pub height: u32,
    pub tx_pos: usize,
}
```

So that the blockhash can be precalculated?

jirijakes: Hmm, it is more of my habit to start with hidden fields (in order to hide internals) and use getters. The blockhash was a bonus.

However, as BDK typically exposes publicly all fields, I guess it would be appropriate to follow also in this case.

jirijakes: > We could even type constraint that `Anchor: for<'a> From<BlockPosition<'a>>`.

`From<BlockPosition>` is currently used only by two methods. Wouldn't that be too strict to require that of every future implementation of Anchor?

> I'm not sure it should be described as the "minimal" set of parameters. It's more like like the maximal amount of detail about the block that could be needed to make an anchor 

Mmm, yeah, I see how it can be unclear. My idea was “if you have _at least_ these parameters at disposal, you can create any anchor.“ I will reword.

jirijakes:
- Renamed to `TxPosInBlock`. Although this name still does not feel perfect, it is better than `BlockPosition` and I cannot think of other.
- Reworded docomment.
- I still kept `TxPosInBlock`'s fields private, accessible via getters. It feels a bit better suitable for this purpose: with public fields instead of a constructor, one of the fields would  be height and therefore block hash would have to be calculated on every access. Please let me know if public fields would be better regardless.

Thanks everybody for the comments!

oleonardolima (commented):
Concept ACK

I agree, the naming is hard. I wasn't able to come up with any other idea than `TxBlockId` (which I don't think really suits 😅)

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -271,42 +271,35 @@ where
...
-        };
+    pub fn apply_block(&mut self, block: &Block, height: u32) -> ChangeSet<A, I::ChangeSet> {
         let mut graph = tx_graph::ChangeSet::default();
         for (tx_pos, tx) in block.txdata.iter().enumerate() {
-            let anchor = A::from_block_position(&block, block_id, tx_pos);
-            graph.merge(self.graph.insert_anchor(tx.compute_txid(), anchor));
+            graph.merge(self.graph.insert_anchor(
+                tx.compute_txid(),
+                TxPosInBlock::new(block, height, tx_pos).into(),
+            ));
```
evanlinjin: This means we are constructing `TxPosInBlock` on every tx, thus re-computing the block hash for every tx... I have a couple of ideas to fix this.

1. Make all fields of `TxPositionInBlock` public and get rid of construction and getter methods.
2. Add a method to change the `tx_pos`. I.e. `fn with_tx_pos(self, tx_pos: usize) -> Self`.

I'm in favor of 1. because it is simpler and I don't see much value in having an API that enforces the consistency of the type. I don't see the caller constructing this type. The main usecase of this type (as I understand) is for implementing `From<TxPositionInBlock>` for anchors types that support it.

crates/chain/src/tx_data_traits.rs
LLFourn: Yes @evanlinjin's design looks right to me.

> From<BlockPosition> is currently used only by two methods. Wouldn't that be too strict to require that of every future implementation of Anchor?

I think if you can't get your anchor from the full block and position you may be using anchors in a way you probably shouldn't.

crates/chain/src/indexed_tx_graph.rs
jirijakes: > Make all fields of TxPositionInBlock public and get rid of construction and getter methods.

Made fields public as suggested.

Although that puts us back [to the beginning](https://github.com/bitcoindevkit/bdk/pull/1594#discussion_r1751085427) (suggestion to pass only height, not BlockId), the repeated calculation of block hash should be avoided. Most likely why the creation of `BlockId` was in the original code in a first place.

jirijakes:
Another round:

- fields in `TxPosInBlock` public, one of them is `block_id` instead of `height` to avoid repetead recalculations of block hash
- if we want to have all anchors implement `From<TxPosInBlock>`, as [hinted](https://github.com/bitcoindevkit/bdk/pull/1594#discussion_r1751086536) by @LLFourn, I believe it could be done in a followup rather than here.

crates/chain/src/tx_data_traits.rs
ValuedMammal: > If we were to have a merkle-proof anchor (which I think we should have), don't we also need a tx position in block?

If we decide this new struct should include the position of the tx in block, then the original name `BlockPosition` would again makes sense to me but I won't belabor the point.

evanlinjin: `BlockPosition` may be interpreted as position of block in the Blockchain though

evanlinjin:
I think it makes sense to include this change in v1.0 since this is a minor breaking change and the work is done.

The worst case scenario is that users that created custom anchors that used methods which required `AnchorFromBlockPosition` will need to change to impl `From<TxPosInBlock>`.

If we deem this breaking change as unacceptable, there is a way to make this non-breaking. We can keep `AnchorFromBlockPosition` and anything that impls it will impl `From<TxPosInBlock>`. We also mark `AnchorFromBlockPosition` as deprecated. 

I'm in favor of just making this change breaking since it's so minor.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -292,20 +297,24 @@ where
...
     ///
-    /// Each inserted transaction's anchor will be constructed from
-    /// [`AnchorFromBlockPosition::from_block_position`].
+    /// Each inserted transaction's anchor will be constructed using [`TxPosInBlock`].
     ///
     /// To only insert relevant transactions, use [`apply_block_relevant`] instead.
     ///
     /// [`apply_block_relevant`]: IndexedTxGraph::apply_block_relevant
-    pub fn apply_block(&mut self, block: Block, height: u32) -> ChangeSet<A, I::ChangeSet> {
+    pub fn apply_block(&mut self, block: &Block, height: u32) -> ChangeSet<A, I::ChangeSet> {
```
evanlinjin: Is there a specific reason for this change?

jirijakes: Ah, not anymore. If I remember, I had to add it in the first iteration when the lifetime for the struct was placed on the impl. No need for it anymore.

I reverted to the original.

ValuedMammal:
How about calling it `TxBlockPosition` @jirijakes @evanlinjin

evanlinjin:
> How about calling it `TxBlockPosition` @jirijakes @evanlinjin

How about `PositionInBlock`? :sweat_smile:

LagginTimes:
> > How about calling it `TxBlockPosition` @jirijakes @evanlinjin
> 
> How about `PositionInBlock`? 😅

Maybe `BlockTxPosition` might be preferable? Or just `TxPosition`? 😅

crates/chain/src/tx_data_traits.rs
```diff
@@ -100,24 +100,30 @@ impl Anchor for ConfirmationBlockTime {
     }
 }
 
-/// An [`Anchor`] that can be constructed from a given block, block height and transaction position
-/// within the block.
-pub trait AnchorFromBlockPosition: Anchor {
-    /// Construct the anchor from a given `block`, block height and `tx_pos` within the block.
-    fn from_block_position(block: &bitcoin::Block, block_id: BlockId, tx_pos: usize) -> Self;
+/// Set of parameters sufficient to construct any [`Anchor`].
```
evanlinjin:
```suggestion
/// Set of parameters sufficient to construct an [`Anchor`].
```

We can't guarantee that this can construct any [`Anchor`], since anyone can implement [`Anchor`] on a new type and the type may not impl `From<TxPosInBlock>`.

crates/chain/src/tx_data_traits.rs
```diff
@@ -100,24 +100,30 @@ impl Anchor for ConfirmationBlockTime {
...
+/// Typically used as an additional constraint on anchor:
+/// `for<'b> A: Anchor + From<TxPosInBlock<'b>>`.
```
evanlinjin, line 106:
```suggestion
///
/// Typically used as an additional constraint on anchor:
/// `for<'b> A: Anchor + From<TxPosInBlock<'b>>`.
```

https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

Let's have this as a separate section.

evanlinjin (commented):
Looks good to me, just some doc changes and I'll happy to ACK.

There is a bit of bike-shedding going on with the naming.

I'm happy with how it is currently, or `BlockTxPosition`, or `PositionInBlock`, or `TxBlockPosition`. Whatever you choose, I will ACK.

oleonardolima:
Alongside Evan's comments, please do mind the commit message to follow https://www.conventionalcommits.org/en/v1.0.0/ too.

I'm fine with the name suggestions above, although I tend more toward either: `TxPosInBlock` (current) or `TxBlockPosition` (these seem clearer to me).

crates/chain/src/tx_data_traits.rs
jirijakes: Done.

crates/chain/src/tx_data_traits.rs
jirijakes: Done.

jirijakes:
Addressed @evanlinjin's comments and changed commit message (hope it's right).

Apologies for the delay.

LagginTimes:
Nit: I would probably reword the commit description to be something like:

```
This change revises how a generic anchor is created from a block, its
`BlockId`, and the transaction's position. The previous approach used a
conversion trait, while the new approach uses a struct and the `From`
trait.
```

LLFourn (approved):
ACK ab8068b90ee476f8edf7d3d4a21d621818e30969

a trait is deleted I ack. Would like @evanlinjin and @LagginTimes  to ACK too.

LagginTimes:
ACK ab8068b90ee476f8edf7d3d4a21d621818e30969

Issue #1432, opened 2024-05-08
Use `bitcoin::Amount` everywhere


ValuedMammal:
It was decided initially to only include `bitcoin::Amount` at the API boundary. No doubt this makes for a better UX. Would it be worth replacing all satoshi amounts represented internally as `u64` with the `Amount` type?

One concern would be: why introduce an abstraction over the denomination when sats are already the standard used throughout the library, but it's possible this fear is overblown.

Alternatives:

- Use a type alias such as `pub type Satoshis = u64;` However this serves no real purpose at the type level other than to enhance readability.
- Use a custom new type and provide conversions to/from `Amount`. The problem with this is re-inventing the wheel when the `Amount` type already exists.

So with respect to the internals we should either use `Amount` everywhere or (almost) nowhere, which seems to be similarly expressed here https://github.com/bitcoindevkit/bdk/issues/823#issuecomment-1354092490

tnull:
I still like the idea of using `Amount` everywhere as it clearly communicates the type. Furthermore, we'll eventually look to introduce similar type in LDK, as the convertion between msat and sat is an error source that keeps on giving.

oleonardolima:
> It was decided initially to only include bitcoin::Amount at the API boundary. No doubt this makes for a better UX. Would it be worth replacing all satoshi amounts represented internally as u64 with the Amount type?

I might be biased, but I think that using `bitcoin::Amount` everywhere we use previously used `u64` as sats makes it clear and easier for the user to understand what type is being expected/used.

It internally adopts `sats` as the standard way to represent Bitcoin amounts (because it just wraps: `Amount(u64)` and `SignedAmount(i64)`, while still having all the support for other operations and conversions between [denominations](https://docs.rs/bitcoin/0.32.0/bitcoin/enum.Denomination.html).


> Alternatives:
> 
> * Use a type alias such as `pub type Satoshis = u64;` However this serves no real purpose at the type level other than to enhance readability.
> * Use a custom new type and provide conversions to/from `Amount`. The problem with this is re-inventing the wheel when the `Amount` type already exists.
> 

I understand that on both scenarios we are reinventing the wheel, and even at a loss of all the conversions between denominations and error treatment already implemented on `bitcoin::Amount`

ValuedMammal:
Using `Amount` everywhere means we'll have to remember to change any error messages returned by bdk to no longer include the word "sats" https://github.com/bitcoindevkit/bdk/pull/1426#discussion_r1594222444

[...]

PR #1595, opened 2024-09-09
refactor(wallet): use `Amount` everywhere

### LINKED ISSUE #1434: example: Fix wallet examples
These are some fixes that can be implemented for the `wallet_*` examples in the example-crates dir.

- [x] Print balances either in BTC or sats (but not both)
- [x] `wallet_esplora_async` appears to be printing the KeychainKind twice, and forgets to print spk0
- [x] Use [tempfile](https://docs.rs/tempfile/latest/tempfile/) instead of `std::env::tempdir`, as this should automatically clean up temporary files
- [x] Consider changing wallet examples to use signet and possibly a different server URL to fix rate limiting issues
- [x] Try to recycle the sats so that running an example doesn't result in a net loss of "funds"

ValuedMammal:
This is a followup to #1426 that refactors wallet internals to use `bitcoin::Amount` (nearly) everywhere. I chose not to change any public types in `coin_selection.rs` that still use `u64` as that might require more discussion.

partially addresses #1432 
fixes #1434 

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/utils.rs
+++ b/crates/wallet/src/wallet/utils.rs
@@ -10,7 +10,7 @@
 // licenses.
 
 use bitcoin::secp256k1::{All, Secp256k1};
-use bitcoin::{absolute, relative, Script, Sequence};
+use bitcoin::{absolute, relative, Amount, Script, Sequence};
 
 use miniscript::{MiniscriptKey, Satisfier, ToPublicKey};
 
@@ -26,9 +26,15 @@ pub trait IsDust {
     fn is_dust(&self, script: &Script) -> bool;
 }
 
+impl IsDust for Amount {
+    fn is_dust(&self, script: &Script) -> bool {
+        *self < script.minimal_non_dust()
+    }
+}
+
 impl IsDust for u64 {
     fn is_dust(&self, script: &Script) -> bool {
-        *self < script.minimal_non_dust().to_sat()
+        Amount::from_sat(*self).is_dust(script)
     }
 }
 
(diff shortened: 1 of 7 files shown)


crates/wallet/src/wallet/utils.rs
```diff
@@ -26,9 +26,15 @@ pub trait IsDust {
...
+        *self < script.minimal_non_dust()
+    }
+}
+
 impl IsDust for u64 {
     fn is_dust(&self, script: &Script) -> bool {
-        *self < script.minimal_non_dust().to_sat()
+        Amount::from_sat(*self).is_dust(script)
     }
 }
```
oleonardolima, line 39: nit: I guess this could be moved to `coin_selection`, it's the only place still using `is_dust`   with a `u64`, right ?

oleonardolima (commented):
ACK 292ec3cb3a41d7b9f42cfea536c39b27dba16d4c

example-crates/wallet_electrum/src/main.rs
```diff
@@ -72,22 +71,19 @@ fn main() -> Result<(), anyhow::Error> {
...
         std::process::exit(0);
     }
 
-    let faucet_address = Address::from_str("mkHS9ne12qx9pS9VojpwU5xtRd4T7X7ZUt")?
-        .require_network(Network::Testnet)?;
-
     let mut tx_builder = wallet.build_tx();
     tx_builder
-        .add_recipient(faucet_address.script_pubkey(), SEND_AMOUNT)
+        .add_recipient(address.script_pubkey(), SEND_AMOUNT)
```
notmandatory, line 86: 👍 thanks for cleaning this up!

notmandatory (approved):
ACK 292ec3cb3a41d7b9f42cfea536c39b27dba16d4c

PR #1597, opened 2024-09-09
fix(wallet): do `check_wallet_descriptor` when creating and loading

### LINKED ISSUE #1557: Inconsistent use of `check_wallet_descriptor`
After #1533 we fail to call [`check_wallet_descriptor`](https://github.com/bitcoindevkit/bdk/blob/76aec62cea852344b1113228645a00413c7fcba9/crates/wallet/src/descriptor/mod.rs#L294) in all arms of a match block during load, but this was correct in v1.0.0-beta.1. Also, we apparently never apply the check during new wallet creation.

https://github.com/bitcoindevkit/bdk/blob/76aec62cea852344b1113228645a00413c7fcba9/crates/wallet/src/wallet/mod.rs#L585-L595

ValuedMammal:
Do `check_wallet_descriptor` for both external and internal descriptors when creating a wallet, and as a defensive measure also perform the check on the descriptors loaded from a changeset.

fixes #1557

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -390,6 +390,7 @@ impl Wallet {
         let (chain, chain_changeset) = LocalChain::from_genesis_hash(genesis_hash);
 
         let (descriptor, mut descriptor_keymap) = (params.descriptor)(&secp, network)?;
+        check_wallet_descriptor(&descriptor)?;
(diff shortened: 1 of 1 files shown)


notmandatory (approved):
ACK 5aecf4d6c48b75e610bbb9e31ad552878d797b5c

PR #1601, opened 2024-09-12
fix(core): calling `CheckPoint::insert` with existing block must succeed

evanlinjin:
### Description

Previously, we were panicking when the caller tried to insert a block at height 0. However, this should succeed if the block hash does not change.

### Notes to the reviewers

For context:

* https://github.com/lightningdevkit/ldk-node/pull/358
* https://discord.com/channels/753336465005608961/978744259693916230/1283050849429356629

### Changelog notice

* Fix `CheckPoint::insert` to not panic when inserting existing genesis block (same block height and hash).

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/core/tests/test_checkpoint.rs
+++ b/crates/core/tests/test_checkpoint.rs
@@ -1,0 +1,36 @@
+#[macro_use]
+mod common;
+
+use bdk_core::CheckPoint;
+
+/// Inserting a block that already exists in the checkpoint chain must always succeed.
+#[test]
+fn checkpoint_insert_existing() {
+    let blocks = &[
+        block_id!(0, "genesis"),
+        block_id!(1, "A"),
+        block_id!(2, "B"),
+        block_id!(3, "C"),
+    ];
+
+    // Index `i` allows us to test with chains of different length.
+    // Index `j` allows us to test inserting different block heights.
+    for i in 0..blocks.len() {
+        let cp_chain = CheckPoint::from_block_ids(blocks[..=i].iter().copied())
+            .expect("must construct valid chain");
+
+        for j in 0..i {
+            let block_to_insert = cp_chain
+                .get(j as u32)
+                .expect("cp of height must exist")
+                .block_id();
+            let new_cp_chain = cp_chain.clone().insert(block_to_insert);
+
+            assert_eq!(
+                new_cp_chain, cp_chain,
+                "must not divert from original chain"
+            );
+            assert!(new_cp_chain.eq_ptr(&cp_chain), "pointers must still match");
+        }
+    }
+}
--- a/crates/core/tests/test_checkpoint.rs
+++ b/crates/core/tests/test_checkpoint.rs
@@ -19,7 +19,7 @@ fn checkpoint_insert_existing() {
         let cp_chain = CheckPoint::from_block_ids(blocks[..=i].iter().copied())
             .expect("must construct valid chain");
 
-        for j in 0..i {
+        for j in 0..=i {
             let block_to_insert = cp_chain
                 .get(j as u32)
                 .expect("cp of height must exist")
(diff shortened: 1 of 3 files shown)


crates/core/tests/test_checkpoint.rs
```diff
@@ -0,0 +1,50 @@
...
+
+#[allow(unused_macros)]
+macro_rules! block_id {
+    ($height:expr, $hash:literal) => {{
+        bdk_chain::BlockId {
+            height: $height,
+            hash: bitcoin::hashes::Hash::hash($hash.as_bytes()),
+        }
+    }};
+}
```
LagginTimes: Would it make sense to have these in a separate `mod.rs` file? If more tests are added to `bdk_core` in the future, these may be useful to have.

crates/core/tests/test_checkpoint.rs
```diff
@@ -0,0 +1,50 @@
+use bdk_core::CheckPoint;
+
+#[allow(unused_macros)]
+macro_rules! h {
+    ($index:literal) => {{
+        bitcoin::hashes::Hash::hash($index.as_bytes())
+    }};
+}
```
oleonardolima: Is this one being used? 🤔

crates/core/tests/test_checkpoint.rs
oleonardolima: Yes, I do agree, +1 on the suggestion.

oleonardolima (commented):
Concept ACK

crates/core/tests/test_checkpoint.rs
evanlinjin: Woops, no it is not.

crates/core/tests/test_checkpoint.rs
evanlinjin: Agree. However, I think a long-term solution should be something like: https://github.com/bitcoindevkit/bdk/issues/1602

evanlinjin: * `common.rs` file.

evanlinjin: Moved `block_id` macro to `tests/common.rs`.

crates/core/tests/test_checkpoint.rs
evanlinjin: Removed.

oleonardolima (approved):
ACK 2970b83f30ca7071f0502de395327eb3671a512b

LagginTimes:
ACK 2970b83f30ca7071f0502de395327eb3671a512b

crates/core/tests/test_checkpoint.rs
```diff
@@ -0,0 +1,36 @@
...
+        block_id!(3, "C"),
+    ];
+
+    // Index `i` allows us to test with chains of different length.
+    // Index `j` allows us to test inserting different block heights.
+    for i in 0..blocks.len() {
+        let cp_chain = CheckPoint::from_block_ids(blocks[..=i].iter().copied())
+            .expect("must construct valid chain");
+
+        for j in 0..i {
```
ValuedMammal:
```suggestion
        for j in 0..=i {
```

ValuedMammal (approved):
I think it's a fine solution but ideally we should ensure bdk never calls `CheckPoint::insert` with a block height of 0 for example in electrum and esplora (still looking for a way to reproduce it).

We can add a Panics section to the docs, something like

```rust
/// # Panics
///
/// This panics if called with a genesis block that differs from that of `self`.
```

ValuedMammal (approved):
Self ACK 3ae9ecba8c893750fa2ed5dfdbb1f4ee84a0b228

notmandatory (approved):
ACK 3ae9ecba8c893750fa2ed5dfdbb1f4ee84a0b228

PR #1603, opened 2024-09-12
chore: fix RUSTSEC-2024-0370 proc-macro-error is unmaintained

oleonardolima:
A candidate for fixing bitcoindevkit/bdk#1593. It builds on top of bitcoindevkit/bdk#1615 

### Description

In https://github.com/bitcoindevkit/bdk/issues/1593 it's mentioned that `proc-macro-error` is unmaintained for the
past few years, with no fix other than using proc-macro-error2 instead.

As on our scenario it's merely a transitive dependency of `clap`,
through `clap_derive` feature, which in latest releases doesn't depend on
`proc-macro-error` we can just bump it to latest.

It's valid to note that by bumping it, both examples that relies on clap
are no longer MSRV (1.63) compliant.

That said, this PR does:

- Standardize the example packages to have `example_` prefix.
- Exclude examples from running in main `Build & Test` CI job.
- Add new testing step to `Build & Test Example` CI job.
- Bumps the `clap` to `4.5.17`.

### Changelog notice

- Standardize the example packages to have `example_` prefix.
- Exclude examples from running in main `Build & Test` CI job.
- Add new testing step to `Build & Test Example` CI job.
- Bumps the `clap` to `4.5.17`.

#### All Submissions:

#### Bugfixes:


oleonardolima:
I also wonder if this could become even simpler by adopting [Examples Cargo Target](https://doc.rust-lang.org/cargo/reference/cargo-targets.html?highlight=examples#examples).

.github/workflows/cont_integration.yml
```diff
@@ -50,9 +50,9 @@ jobs:
           cargo update -p tokio --precise "1.38.1"
           cargo update -p tokio-util --precise "0.7.11"
       - name: Build
-        run: cargo build ${{ matrix.features }}
+        run: cargo build --workspace --exclude 'example_*' ${{ matrix.features }}
```
ValuedMammal, line 53: We could try just doing `--all-features` 🤔

notmandatory: I think in the past we ran into errors that only happened when we built with `--no-default-features` so I'd like to keep it.

ValuedMammal: keep which one?

notmandatory: I mean keep testing with `--no-default-features` and `--all-features`. I don't mind only building the examples with the current `stable` rust though if that's the issue.

oleonardolima: I think it's better to keep testing with both.

oleonardolima:
I think I also found a bug in how the `bdk_esplora` features were structured, it's somewhat depending on `blocking` feature being set up through the `wallet_esplora` example, and that's why it's failing in the CI when excluding the examples from the workspace.

oleonardolima:
> I think I also found a bug in how the `bdk_esplora` features were structured, it's somewhat depending on `blocking` feature being set up through the `wallet_esplora` example, and that's why it's failing in the CI when excluding the examples from the workspace.

Also, it's something that we'd find out through bitcoindevkit/bdk#1944 if miniscript std/no-std weren't a blocker.

.github/workflows/cont_integration.yml
```diff
@@ -172,3 +172,6 @@ jobs:
       - name: Build
         working-directory: example-crates/${{ matrix.example-dir }}
         run: cargo build
+      - name: Test
+        working-directory: example-crates/${{ matrix.example-dir }}
+        run: cargo test
```
ValuedMammal: Just noting I don't think there are any tests here to run.

oleonardolima: Oh, yes. Thank you!

oleonardolima:
> I think I also found a bug in how the `bdk_esplora` features were structured, it's somewhat depending on `blocking` feature being set up through the `wallet_esplora` example, and that's why it's failing in the CI when excluding the examples from the workspace.

I was testing the fixes and got it in 8752555, but I'll move it to a separate PR to make it easier to review (#1615).

ValuedMammal:
Rather than rename the example-crates can't we use the "exclude" key in the workspace Cargo.toml to exclude them from invocations of `cargo test --workspace`?

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields

oleonardolima:
> https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields

I thought about it, but AFAICT it'll exclude from the workspace for any `cargo` command, and not only for the `cargo test` one.

ValuedMammal (approved):
After considering the options I don't see any downside to using this solution.

oleonardolima:
> After considering the options I don't see any downside to using this solution.

I rebased this one, as I updated the approach in bitcoindevkit/bdk#1615.

ValuedMammal (approved):
ACK d802d00a06035bec582eb5aadf0281f26bff6acd

PR #1605, opened 2024-09-12
fix(wallet): fix SingleRandomDraw to error if insufficient funds

notmandatory:
### Description

* fix SingleRandomDraw to error if insufficient funds
* fixed spelling and clippy errors (see: https://github.com/bitcoindevkit/bdk/pull/1581#discussion_r1756936178)
* updated tests to check for error variant instead of a panic

### Notes to the reviewers

Since the single random draw algo can be used on its own it needs to be able to return an insufficient funds error. I think the reason we didn't catch this before is that single random draw already check for sufficient required + optional utxo amounts and returns the insufficient funds error.

### Changelog notice

* fix SingleRandomDraw coin selection to error if there are insufficient funds for a requested payment amount.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/coin_selection.rs
+++ b/crates/wallet/src/wallet/coin_selection.rs
@@ -402,7 +402,7 @@ pub struct BranchAndBoundCoinSelection<Cs = SingleRandomDraw> {
     fallback_algorithm: Cs,
 }
 
-/// Error returned by branch and bond coin selection.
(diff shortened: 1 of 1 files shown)


ValuedMammal (approved):
ACK 22a2f83db5df5f48974a06deef6b45d41a05ef3b

PR #1612, opened 2024-09-14
Refactor: Move transaction testing utilities from `crates/chain/tests/common` to `testenv` crate

### LINKED ISSUE #1602: Move stuff from `crates/chain/tests/common` to `bdk_testenv` and make them publicly-accessible
**Describe the enhancement**  

These macros, fake-tx-construction-tools are helpful in other crates too. Let's move them to `bdk_testenv` and make them publicly available. 

**Use case**  

Better test tooling.

tvpeter:
### Description

This change moves all transaction testing utilities (macros and functions) from `crates/chain/tests/common` to `crates/testenv/tx_utils` to minimize duplication.

Closes #1602 

#### All Submissions:

#### New Features:

#### Bugfixes:


evanlinjin:
Hey @tvpeter thank you for taking this on! Please follow the [https://www.conventionalcommits.org/en/v1.0.0/](Conventional Commit Specification) for the commit messages. I'll review this properly soon!

tvpeter:
@evanlinjin thank you.
I have reworded the commit messages to follow the specification.

crates/testenv/Cargo.toml
```diff
@@ -18,6 +18,7 @@ workspace = true
 [dependencies]
 bdk_chain = { path = "../chain", version = "0.19.0", default-features = false }
 electrsd = { version = "0.28.0", features = [ "bitcoind_25_0", "esplora_a33e97e1", "legacy" ] }
+bitcoin = { version = "0.32.0", default-features = false }
```
LagginTimes: Hello! Since `bdk_chain` already re-exports `bitcoin`, we don't need to add `bitcoin` as a dependency here.

tvpeter: Fixed. Thank you

oleonardolima (commented):
Concept ACK

I only wonder about naming/structure, maybe reword it to only `utils.rs` or have both `chain_utils.rs` and `tx_utils.rs`, but I'm not sure about it.

tvpeter:
Alright @oleonardolima. I initially thought that it will contain transaction related utilities only but on a second thought most of the utilities are not just for transactions. So I will rename it to just `utils.rs` and move the `descriptors` sample array there too.  Thank you.

evanlinjin:
Please use the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) format for commit messages.

Looking at this now, maybe we should create a `bdk_testutil` crate for this instead? These are light macros, functions. Whereas `bdk_testenv` requires downloading binaries during the build process.

cc. @ValuedMammal @notmandatory

ValuedMammal:
> Looking at this now, maybe we should create a `bdk_testutil` crate for this instead? These are light macros, functions. Whereas `bdk_testenv` requires downloading binaries during the build process.

Another option that comes to mind is to have conditional compilation of testenv with something like a "full" feature flag, but I don't have a strong opinion on it at the moment.

LagginTimes:
> > Looking at this now, maybe we should create a `bdk_testutil` crate for this instead? These are light macros, functions. Whereas `bdk_testenv` requires downloading binaries during the build process.
> 
> Another option that comes to mind is to have conditional compilation of testenv with something like a "full" feature flag, but I don't have a strong opinion on it at the moment.

I think conditional compilation sounds like the better option here vs. having an additional crate to manage, since both crates would be primarily focused on testing.

ValuedMammal:
> > > Looking at this now, maybe we should create a `bdk_testutil` crate for this instead? These are light macros, functions. Whereas `bdk_testenv` requires downloading binaries during the build process.
> > 
> > 
> > Another option that comes to mind is to have conditional compilation of testenv with something like a "full" feature flag, but I don't have a strong opinion on it at the moment.
> 
> I think conditional compilation sounds like the better option here vs. having an additional crate to manage, since both crates would be primarily focused on testing.

~~Considering the test utils are common to other crates and change infrequently, it might be just as easy to move them to a `test_util` module in `bdk_core`~~

Edit: As discussed on dev call the current direction on the PR makes sense

LagginTimes:
Nit: I think squashing the commits down to one would make the history a bit cleaner.

tvpeter:
> Nit: I think squashing the commits down to one would make the history a bit cleaner.

Done

crates/chain/tests/test_local_chain.rs
```diff
@@ -9,6 +9,7 @@ use bdk_chain::{
     },
     BlockId,
 };
+use bdk_testenv::{chain_update, h, local_chain};
```
oleonardolima: As we are importing this now from bdk_testenv, and deleted the macros from `commond/mod.rs` I guess we can remove the code at L16-17.

``` rust
#[macro_use]
mod common;
```

crates/testenv/src/utils.rs
```diff
@@ -0,0 +1,90 @@
...
+    }};
+}
+
+#[allow(unused_macros)]
+#[macro_export]
+macro_rules! h {
+    ($index:literal) => {{
+        bitcoin::hashes::Hash::hash($index.as_bytes())
+    }};
+}
```
oleonardolima: IIRC we discussed during the call that this could have a meaningful name, I  would use `hash` for instance.

oleonardolima (commented):
I did another review and left some comments. Overall, it's looking good!

I think you still need to change the commit message to lowercase.

oleonardolima:
You also probably need to rebase on top of the master, so the CI issues are fixed (in case you are curious, check: #1636).

tvpeter:
> I did another review and left some comments. Overall, it's looking good!
> 
> I think you still need to change the commit message to lowercase.

Done. Thank you

crates/testenv/src/utils.rs
```diff
@@ -0,0 +1,90 @@
...
+
+#[allow(unused_macros)]
+#[macro_export]
+macro_rules! local_chain {
+    [ $(($height:expr, $block_hash:expr)), * ] => {{
+        #[allow(unused_mut)]
+        bdk_chain::local_chain::LocalChain::from_blocks([$(($height, $block_hash).into()),*].into_iter().collect())
+            .expect("chain must have genesis block")
+    }};
+}
```
LagginTimes: Nit: I suggest renaming this expression to `hash` for consistency with the other macros.

tvpeter: The previous `h!` macro has been renamed to `hash!`, so it will conflict. Since it builds a `LocalChain`, is it better to rename it `new_local_chain`?

LagginTimes: Oops! Just to clarify: I meant to change `block_hash` to `hash`, not rename the `local_chain` macro. Apologies for the confusion!

tvpeter: done

ValuedMammal (approved):
ACK 88a8403707038264251b3e7fa18b8f21af8646ff

notmandatory (approved):
ACK 88a8403707038264251b3e7fa18b8f21af8646ff

Thanks for taking this one on.

PR #1614, opened 2024-09-15
feat(rpc): introduce FilterIter

ValuedMammal:
The PR adds a `bip158` module to the `bdk_bitcoind_rpc` crate along with a new type `FilterIter` that can be used for retrieving blocks from a full node which contain transactions relevant to a list of script pubkeys.

### Changelog notice

`bdk_bitcoind_rpc`: Added `bip158` module as a means of updating `bdk_chain` structures

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/bitcoind_rpc/src/bip158.rs
+++ b/crates/bitcoind_rpc/src/bip158.rs
@@ -1,0 +1,267 @@
+//! Compact block filters sync over RPC. For more details refer to [BIP157][0].
+//!
+//! This module is home to [`FilterIter`], a structure that returns bitcoin blocks by matching
+//! a list of script pubkeys against a [BIP158][1] [`BlockFilter`].
+//!
+//! [0]: https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki
+//! [1]: https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki
+
+use bdk_core::collections::BTreeMap;
+use core::fmt;
+
+use bdk_core::bitcoin;
+use bdk_core::{BlockId, CheckPoint};
+use bitcoin::{
+    bip158::{self, BlockFilter},
+    Block, BlockHash, ScriptBuf,
+};
+use bitcoincore_rpc;
+use bitcoincore_rpc::RpcApi;
+
+/// Block height
+type Height = u32;
+
+/// Type that generates block [`Event`]s by matching a list of script pubkeys against a
+/// [`BlockFilter`].
+#[derive(Debug)]
+pub struct FilterIter<'c, C> {
+    // RPC client
+    client: &'c C,
+    // SPK inventory
+    spks: Vec<ScriptBuf>,
+    // local cp
+    cp: Option<CheckPoint>,
+    // blocks map
+    blocks: BTreeMap<Height, BlockHash>,
+    // best height counter
+    height: Height,
+    // stop height
+    stop: Height,
+}
+
+impl<'c, C: RpcApi> FilterIter<'c, C> {
+    /// Construct [`FilterIter`] from a given `client` and start `height`.
+    pub fn new_with_height(client: &'c C, height: u32) -> Self {
+        Self {
+            client,
+            spks: vec![],
+            cp: None,
+            blocks: BTreeMap::new(),
+            height,
+            stop: 0,
+        }
+    }
+
+    /// Construct [`FilterIter`] from a given `client` and [`CheckPoint`].
+    pub fn new_with_checkpoint(client: &'c C, cp: CheckPoint) -> Self {
+        let mut filter_iter = Self::new_with_height(client, cp.height());
+        filter_iter.cp = Some(cp);
+        filter_iter
+    }
+
+    /// Extends `self` with an iterator of spks.
+    pub fn add_spks(&mut self, spks: impl IntoIterator<Item = ScriptBuf>) {
+        self.spks.extend(spks)
+    }
+
+    /// Add spk to the list of spks to scan with.
+    pub fn add_spk(&mut self, spk: ScriptBuf) {
+        self.spks.push(spk);
+    }
+
+    /// Get the next filter and increment the current best height.
+    ///
+    /// Returns `Ok(None)` when the stop height is exceeded.
+    fn next_filter(&mut self) -> Result<Option<NextFilter>, Error> {
+        if self.height > self.stop {
+            return Ok(None);
+        }
+        let height = self.height;
+        let hash = match self.blocks.get(&height) {
+            Some(h) => *h,
+            None => self.client.get_block_hash(height as u64)?,
+        };
+        let filter_bytes = self.client.get_block_filter(&hash)?.filter;
+        let filter = BlockFilter::new(&filter_bytes);
+        self.height += 1;
+        Ok(Some((BlockId { height, hash }, filter)))
+    }
+
+    /// Get the remote tip.
+    ///
+    /// Returns `None` if the remote height is not strictly greater than the height of this
+    /// [`FilterIter`].
+    pub fn get_tip(&mut self) -> Result<Option<BlockId>, Error> {
+        let tip_hash = self.client.get_best_block_hash()?;
+        let mut header = self.client.get_block_header_info(&tip_hash)?;
+        let tip_height = header.height as u32;
+        if self.height >= tip_height {
+            // nothing to do
+            return Ok(None);
+        }
+        self.blocks.insert(tip_height, tip_hash);
+
+        // if we have a checkpoint we use a lookback of ten blocks
+        // to ensure consistency of the local chain
+        if let Some(cp) = self.cp.as_ref() {
+            // adjust start height to point of agreement + 1
+            let base = self.find_base_with(cp.clone())?;
+            self.height = base.height + 1;
+
+            for _ in 0..9 {
+                let hash = match header.previous_block_hash {
+                    Some(hash) => hash,
+                    None => break,
+                };
+                header = self.client.get_block_header_info(&hash)?;
+                let height = header.height as u32;
+                if height < self.height {
+                    break;
+                }
+                self.blocks.insert(height, hash);
+            }
+        }
+
+        self.stop = tip_height;
+
+        Ok(Some(BlockId {
+            height: tip_height,
+            hash: tip_hash,
+        }))
+    }
+}
+
+/// Alias for a compact filter and associated block id.
+type NextFilter = (BlockId, BlockFilter);
+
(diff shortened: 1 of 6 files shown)


evanlinjin:
This is what we've all been waiting for.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -0,0 +1,261 @@
...
+                    None => break,
+                    Some(hash) => {
+                        header_info = self.client.get_block_header_info(&hash)?;
+                        let height = header_info.height as u32;
+                        self.blocks.insert(height, hash);
+                    }
+                }
+            }
+            let start_height = self.blocks.keys().next().expect("blocks not empty");
+            self.height = cp.height().min(*start_height);
```
ValuedMammal: This looks to be inefficient in that it could cause us to re-emit full blocks redundantly. I think the start height should always begin one block after the point of agreement with the local chain

LLFourn:
ConceptACK. Don't have time for a full review now. @evanlinjin?

crates/bitcoind_rpc/examples/bip158.rs
```diff
@@ -0,0 +1,75 @@
...
+        let mut index = KeychainTxOutIndex::default();
+        index.insert_descriptor(0, descriptor.clone())?;
+        index.insert_descriptor(1, change_descriptor.clone())?;
+        index
+    });
+
+    // Assume a minimum birthday height
+    let block = BlockId {
+        height: 205_000,
+        hash: "0000002bd0f82f8c0c0f1e19128f84c938763641dba85c44bdb6aed1678d16cb".parse()?,
```
oleonardolima: nit: I guess these fields could be moved as constants above, wdyt?

crates/bitcoind_rpc/examples/bip158.rs
```diff
@@ -0,0 +1,75 @@
...
+    };
+    let _ = chain.insert_block(block)?;
+
+    // Configure RPC client
+    let rpc_client = bitcoincore_rpc::Client::new(
+        "127.0.0.1:38332",
```
oleonardolima: nit: It could be used either as a constant above, or an environment variable (preferrable in my opinion).

crates/bitcoind_rpc/tests/bip158.rs
```diff
@@ -0,0 +1,114 @@
...
+use bitcoincore_rpc::RpcApi;
+
+macro_rules! block_id {
+    ($height:expr, $hash:literal) => {{
+        BlockId {
+            height: $height,
+            hash: bdk_core::bitcoin::hashes::Hash::hash($hash.as_bytes()),
+        }
+    }};
+}
```
oleonardolima: nit: you can now use this `block_id` from `bdk_testenv`.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -177,16 +177,18 @@ impl<'c, C: RpcApi> Iterator for FilterIter<'c, C> {
     type Item = Result<Event, Error>;
 
     fn next(&mut self) -> Option<Self::Item> {
+        if self.spks.is_empty() {
+            return None;
+        }
```
oleonardolima: I assume that the user would properly use the `extend` or `add` methods, however, I wonder how could we make it clear here that it's missing the `spks`, otherwise the user would only get Nones without much information 🤔

ValuedMammal: In this case would it be better to return `Event::NoMatch`? If I recall, the reason for adding this check was that `FilterIter` seemed to be returning empty blocks as matches even without spks (on regtest) which is also confusing. I'll do some more testing though to confirm it.

ValuedMammal:
Thanks for the review @oleonardolima

crates/bitcoind_rpc/src/bip158.rs
oleonardolima: I would say it could be a specific error variant indicating the spks are empty.

ValuedMammal:
Rebased and addressed comments from @oleonardolima

crates/bitcoind_rpc/tests/test_filter_iter.rs
```diff
@@ -0,0 +1,142 @@
...
+fn filter_iter_error_no_scripts() -> anyhow::Result<()> {
+    use bdk_bitcoind_rpc::bip158::Error;
+    let env = testenv()?;
+
+    let addr = env
+        .rpc_client()
+        .get_new_address(None, None)?
+        .assume_checked();
+    let spk = addr.script_pubkey();
+    let _ = env.mine_blocks(9, Some(addr))?;
```
nymius: If the goal of this test is to check the behavior when no spk has been added to the filter, I would avoid reiterative checks over the same thing. In this case, it would mean to just mine 1 or 2 blocks.

```suggestion
    let _ = env.mine_blocks(1, Some(addr))?;

    let mut iter = FilterIter::new_with_height(env.rpc_client(), 1);
    assert_eq!(iter.get_tip().unwrap().map(|b| b.height), Some(2));

    // iterator should return two errors
    for res in iter.by_ref().take(2) {
        assert!(matches!(res, Err(Error::NoScripts)));
```

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -0,0 +1,277 @@
...
+            {
+                let block = self.client.get_block(&hash)?;
+                self.blocks.insert(height, hash);
+                let inner = EventInner { height, block };
+                Ok(Some(Event::Block(inner)))
+            } else {
+                Ok(Some(Event::NoMatch(height)))
+            };
+
+            self.next_filter = self.next_filter()?;
```
nymius: I felt `next_iter` field was redundant, as it is a result of the state of other fields already kept by `FilterIter`, and not something accessed by the other struct methods besides `next` from `Iterator`.  If you agree, I came up with the following:

```diff
diff --git a/crates/bitcoind_rpc/src/bip158.rs b/crates/bitcoind_rpc/src/bip158.rs
index 2c5bc114..5419716b 100644
--- a/crates/bitcoind_rpc/src/bip158.rs
+++ b/crates/bitcoind_rpc/src/bip158.rs
@@ -33,8 +33,6 @@ pub struct FilterIter<'c, C> {
     cp: Option<CheckPoint>,
     // blocks map
     blocks: BTreeMap<Height, BlockHash>,
-    // next filter
-    next_filter: Option<NextFilter>,
     // best height counter
     height: Height,
     // stop height
@@ -49,7 +47,6 @@ impl<'c, C: RpcApi> FilterIter<'c, C> {
             spks: vec![],
             cp: None,
             blocks: BTreeMap::new(),
-            next_filter: None,
             height,
             stop: 0,
         }
@@ -127,9 +124,6 @@ impl<'c, C: RpcApi> FilterIter<'c, C> {
 
         self.stop = tip_height;
 
-        // get the first filter
-        self.next_filter = self.next_filter()?;
-
         Ok(Some(BlockId {
             height: tip_height,
             hash: tip_hash,
@@ -177,31 +171,27 @@ impl<C: RpcApi> Iterator for FilterIter<'_, C> {
     type Item = Result<Event, Error>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let (block, filter) = self.next_filter.clone()?;
-
         (|| -> Result<_, Error> {
             // if the next filter matches any of our watched spks, get the block
             // and return it, inserting relevant block ids along the way
-            let height = block.height;
-            let hash = block.hash;
-
-            let result = if self.spks.is_empty() {
-                Err(Error::NoScripts)
-            } else if filter
-                .match_any(&hash, self.spks.iter().map(|script| script.as_bytes()))
-                .map_err(Error::Bip158)?
-            {
-                let block = self.client.get_block(&hash)?;
-                self.blocks.insert(height, hash);
-                let inner = EventInner { height, block };
-                Ok(Some(Event::Block(inner)))
-            } else {
-                Ok(Some(Event::NoMatch(height)))
-            };
-
-            self.next_filter = self.next_filter()?;
-
-            result
+            self.next_filter()?.map_or(Ok(None), |(block, filter)| {
+                let height = block.height;
+                let hash = block.hash;
+
+                if self.spks.is_empty() {
+                    Err(Error::NoScripts)
+                } else if filter
+                    .match_any(&hash, self.spks.iter().map(|script| script.as_bytes()))
+                    .map_err(Error::Bip158)?
+                {
+                    let block = self.client.get_block(&hash)?;
+                    self.blocks.insert(height, hash);
+                    let inner = EventInner { height, block };
+                    Ok(Some(Event::Block(inner)))
+                } else {
+                    Ok(Some(Event::NoMatch(height)))
+                }
+            })
         })()
         .transpose()
     }
```

ValuedMammal: Brilliant. will fix

ValuedMammal:
In the last push

- Removed unneeded `next_filter` field
- Fixed up test checking for `Error::NoScripts`
- Added test `filter_iter_returns_matched_blocks`

CI failure seems unrelated

oleonardolima (commented):
cACK db65d96a98073b1d96f62a0b5643fa997fadb4ca

ValuedMammal:
Rebased

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -0,0 +1,267 @@
...
+    pub fn add_spks(&mut self, spks: impl IntoIterator<Item = ScriptBuf>) {
+        self.spks.extend(spks)
+    }
+
+    /// Add spk to the list of spks to scan with.
+    pub fn add_spk(&mut self, spk: ScriptBuf) {
+        self.spks.push(spk);
+    }
+
+    /// Get the next filter and increment the current best height.
```
nymius, line 72: Shouldn't this be _line comments_ (`// Get the next filt...`) instead of _line docs_ (`/// Get the next filt...`)? If not, documentation for this private method will only be shown if compiled with `cargo doc --document-private-items`.

nymius: Addressed by @ValuedMammal in weekly call. Is used to have documentation snippets of private methods in text editor, and I think is reasonable to have it there, if at the end is not going to collude with the public documentation.

nymius (approved):
tACK 7ad95d35

ValuedMammal:
Rebased

notmandatory (approved):
tACK a6364e20500fba7c16f31fa97d106893110751c6

This is cool stuff, I added some docs to make it easier for others to run.

ValuedMammal:
Thanks @notmandatory

evanlinjin:

4. Call `.next` again.

PR #1615, opened 2024-09-16
fix(bdk_esplora, bdk_electrum): build and test with `--no-default-features`

oleonardolima:
### Description

It was noticed in #1603 that both `bdk_esplora` and `bdk_electrum` were not able to build and test successfully when `--no-default-features` was passed.

It wasn't noticed in CI, because the `example-crates` are in the same workspace as the main crates, therefore even though when passing `--no-default-features`, the examples were making the `blocking` feature to `bdk_esplora` and `use-rustls` to `bdk_electrum` available, therefore not failing in the tests.

This PR attempts to fix that, and is now used as base for #1603.

### Changelog notice

- Add `blocking-https` as default feature to `bdk_esplora`, and make `{async|blocking}-https-native available too.
- Add tests behind features for both `bdk_esplora` and `bdk_electrum`.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/electrum/Cargo.toml
+++ b/crates/electrum/Cargo.toml
@@ -24,3 +24,8 @@ bdk_chain = { path = "../chain" }
 default = ["use-rustls"]
 use-rustls = ["electrum-client/use-rustls"]
 use-rustls-ring = ["electrum-client/use-rustls-ring"]
+
+[[test]]
+name = "use-rustls"
+path = "tests/test_electrum.rs"
+required-features = ["use-rustls"]
--- a/crates/electrum/Cargo.toml
+++ b/crates/electrum/Cargo.toml
@@ -26,6 +26,5 @@ use-rustls = ["electrum-client/use-rustls"]
 use-rustls-ring = ["electrum-client/use-rustls-ring"]
 
 [[test]]
-name = "use-rustls"
-path = "tests/test_electrum.rs"
+name = "test_electrum"
 required-features = ["use-rustls"]
--- a/crates/esplora/Cargo.toml
+++ b/crates/esplora/Cargo.toml
@@ -39,11 +39,9 @@ blocking-https-rustls = ["blocking", "esplora-client/blocking-https-rustls"]
 blocking-https-native = ["blocking", "esplora-client/blocking-https-native"]
 
 [[test]]
-name = "blocking"
-path = "tests/blocking_ext.rs"
+name = "blocking_ext"
 required-features = ["blocking"]
 
 [[test]]
-name = "async"
-path = "tests/async_ext.rs"
+name = "async_ext"
 required-features = ["async"]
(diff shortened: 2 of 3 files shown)


crates/esplora/tests/async_ext.rs
```diff
@@ -1,5 +1,6 @@
 use bdk_chain::spk_client::{FullScanRequest, SyncRequest};
 use bdk_chain::{ConfirmationBlockTime, TxGraph};
+#[cfg(feature = "async")]
```
ValuedMammal: Would it accomplish the same thing if we enable these features as dev-dependencies?

```rs
[dev-dependencies]
bdk_esplora = { path = ".", features = ["async", "blocking"] }
```

oleonardolima: I was wondering about it, if either (1) using `#[cfg(feature = "..."]`, (2) your suggestion, or using [`required-features`](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field) would be the best approach.

ValuedMammal: I guess a better question is should these tests run when doing `cargo test --workspace --no-default-features` or is it fine to skip them as long as they are picked up when doing `--all-features`

oleonardolima: I think it's fine to skip them, with `required-features` approach, as they're strictly linked to both features (e.g: blocking, async).

ValuedMammal (approved):
ACK 086eb6ed4794e20dbb7328b3a1c192c5a100049f

crates/esplora/tests/async_ext.rs
oleonardolima: @ValuedMammal Using the `required-features` approach would look like this in `Cargo.toml`, for example: 

``` toml
[[test]]
name = "blocking"
path = "tests/blocking_ext.rs"
required-features = ["blocking"]

[[test]]
name = "async"
path = "tests/async_ext.rs"
required-features = ["async"]
```

P.S.: Then we wouldn't require the `cfg![feature = "async|blocking"] in the test files.

ValuedMammal: Yes I think that sounds reasonable

oleonardolima: Sweet! I updated both commits, and it's ready for re-review.

crates/electrum/Cargo.toml
```diff
@@ -24,3 +24,8 @@ bdk_chain = { path = "../chain" }
 default = ["use-rustls"]
 use-rustls = ["electrum-client/use-rustls"]
 use-rustls-ring = ["electrum-client/use-rustls-ring"]
+
+[[test]]
+name = "use-rustls"
+path = "tests/test_electrum.rs"
+required-features = ["use-rustls"]
```
ValuedMammal: Can we do this? - use the same name as the path
```suggestion
[[test]]
name = "test_electrum"
required-features = ["use-rustls"]
```

oleonardolima: Yes, I pushed an update with it in [9b7b195](https://github.com/bitcoindevkit/bdk/pull/1615/commits/9b7b19586e1f74613f71b6c2403a7df419e13b65)

ValuedMammal (approved):
ACK 9b7b19586e1f74613f71b6c2403a7df419e13b65

PR #1616, opened 2024-09-17
feat(wallet)!: enable RBF by default on TxBuilder

### LINKED ISSUE #791: Substitute `enable_rbf` in `TxBuilder` with `disable_rbf`
i.e., maybe all our txs should be RBF by default :)

luisschwab:
### Description

Closes #791 

This PR enables RBF by default on `TxBuilder`

### Changelog notice
- On TxParams, `rbf` becomes `sequence`
- Added `set_exact_sequence()`, so the user can define an arbitrary sequence value
- `n_sequence` now defaults to `0xFFFFFFFD` 
- Adjusted tests accordingly

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -1366,36 +1366,20 @@ impl Wallet {
             }
         };
 
-        // The nSequence to be by default for inputs unless an explicit sequence is specified.
-        let n_sequence = match (params.rbf, requirements.csv) {
-            // No RBF or CSV but there's an nLockTime, so the nSequence cannot be final
-            (None, None) if lock_time != absolute::LockTime::ZERO => {
-                Sequence::ENABLE_LOCKTIME_NO_RBF
-            }
-            // No RBF, CSV or nLockTime, make the transaction final
-            (None, None) => Sequence::MAX,
-
-            // No RBF requested, use the value from CSV. Note that this value is by definition
-            // non-final, so even if a timelock is enabled this nSequence is fine, hence why we
-            // don't bother checking for it here. The same is true for all the other branches below
+        // nSequence value for inputs
+        // When not explicitly specified, defaults to 0xFFFFFFFD,
+        // meaning RBF signaling is enabled
+        let n_sequence = match (params.sequence, requirements.csv) {
+            // Enable RBF by default
+            (None, None) => Sequence::ENABLE_RBF_NO_LOCKTIME,
+            // None requested, use required
             (None, Some(csv)) => csv,
-
-            // RBF with a specific value but that value is too high
-            (Some(tx_builder::RbfValue::Value(rbf)), _) if !rbf.is_rbf() => {
-                return Err(CreateTxError::RbfSequence)
+            // Requested sequence is incompatible with requirements
+            (Some(sequence), Some(csv)) if !check_nsequence_rbf(sequence, csv) => {
+                return Err(CreateTxError::RbfSequenceCsv { sequence, csv })
             }
-            // RBF with a specific value requested, but the value is incompatible with CSV
-            (Some(tx_builder::RbfValue::Value(rbf)), Some(csv))
-                if !check_nsequence_rbf(rbf, csv) =>
-            {
-                return Err(CreateTxError::RbfSequenceCsv { rbf, csv })
-            }
-
-            // RBF enabled with the default value with CSV also enabled. CSV takes precedence
-            (Some(tx_builder::RbfValue::Default), Some(csv)) => csv,
-            // Valid RBF, either default or with a specific value. We ignore the `CSV` value
-            // because we've already checked it before
-            (Some(rbf), _) => rbf.get_value(),
+            // Use requested nSequence value
+            (Some(sequence), _) => sequence,
         };
 
         let (fee_rate, mut fee_amount) = match params.fee_policy.unwrap_or_default() {
@@ -1609,8 +1593,7 @@ impl Wallet {
     /// let mut psbt = {
     ///     let mut builder = wallet.build_tx();
     ///     builder
-    ///         .add_recipient(to_address.script_pubkey(), Amount::from_sat(50_000))
-    ///         .enable_rbf();
+    ///         .add_recipient(to_address.script_pubkey(), Amount::from_sat(50_000));
     ///     builder.finish()?
     /// };
     /// let _ = wallet.sign(&mut psbt, SignOptions::default())?;
--- a/crates/wallet/src/wallet/tx_builder.rs
+++ b/crates/wallet/src/wallet/tx_builder.rs
@@ -31,9 +31,7 @@
 //!     // With a custom fee rate of 5.0 satoshi/vbyte
 //!     .fee_rate(FeeRate::from_sat_per_vb(5).expect("valid feerate"))
 //!     // Only spend non-change outputs
-//!     .do_not_spend_change()
-//!     // Turn on RBF signaling
-//!     .enable_rbf();
+//!     .do_not_spend_change();
 //! let psbt = tx_builder.finish()?;
 //! # Ok::<(), anyhow::Error>(())
 //! ```
@@ -134,7 +132,7 @@ pub(crate) struct TxParams {
     pub(crate) sighash: Option<psbt::PsbtSighashType>,
     pub(crate) ordering: TxOrdering,
     pub(crate) locktime: Option<absolute::LockTime>,
-    pub(crate) rbf: Option<RbfValue>,
+    pub(crate) sequence: Option<Sequence>,
     pub(crate) version: Option<Version>,
     pub(crate) change_policy: ChangeSpendPolicy,
     pub(crate) only_witness_utxo: bool,
@@ -554,23 +552,12 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
         }
     }
 
-    /// Enable signaling RBF
+    /// Set an exact nSequence value
     ///
-    /// This will use the default nSequence value of `0xFFFFFFFD`.
-    pub fn enable_rbf(&mut self) -> &mut Self {
-        self.params.rbf = Some(RbfValue::Default);
-        self
-    }
-
-    /// Enable signaling RBF with a specific nSequence value
-    ///
-    /// This can cause conflicts if the wallet's descriptors contain an "older" (OP_CSV) operator
-    /// and the given `nsequence` is lower than the CSV value.
-    ///
-    /// If the `nsequence` is higher than `0xFFFFFFFD` an error will be thrown, since it would not
-    /// be a valid nSequence to signal RBF.
-    pub fn enable_rbf_with_sequence(&mut self, nsequence: Sequence) -> &mut Self {
-        self.params.rbf = Some(RbfValue::Value(nsequence));
+    /// This can cause conflicts if the wallet's descriptors contain an
+    /// "older" (OP_CSV) operator and the given `nsequence` is lower than the CSV value.
+    pub fn set_exact_sequence(&mut self, n_sequence: Sequence) -> &mut Self {
+        self.params.sequence = Some(n_sequence);
         self
     }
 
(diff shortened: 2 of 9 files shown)


ValuedMammal:
Thanks for working on this @luisschwab

rustaceanrob:
I am inclined to think everyone with production applications will have `enable_rbf` selected. I wonder how we feel about adding a `deprecated` macro and having the method essentially do nothing instead of removing it. That way an LSP will show the deprecation warning but this would technically not be a breaking change.

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -145,6 +143,34 @@ pub(crate) struct TxParams {
...
+            fee_policy: None,
+            internal_policy_path: None,
+            external_policy_path: None,
+            utxos: Vec::new(),
+            unspendable: HashSet::new(),
+            manually_selected_only: false,
+            sighash: None,
+            ordering: TxOrdering::default(),
+            locktime: None,
+            rbf: Some(RbfValue::Default),
```
ValuedMammal: I think this would be cleaner if we had `RbfValue` implement Default. Or we can let it default to None as it does now and handle the option when creating the tx, but that's not my first choice.

ValuedMammal (commented):
I have a slightly different take but I think we agree on the end goal of making RBF the default for new transactions. I think the place to address it is by fixing the logic in [`create_tx`](https://github.com/bitcoindevkit/bdk/blob/2cf46a2a9e513d3f2612b7c363446fb1acf958ae/crates/wallet/src/wallet/mod.rs#L1370) to first check the value of `requirements.csv`, then allow any Sequence value passed in by the user, or else fall back to the new default `ENABLE_RBF_NO_LOCKTIME` (0xFFFFFFFD). To make it simple `RbfValue` should implement Default but maybe not strictly necessary.

Instead of `disable_rbf` I think we should have a method `nsequence` that accepts a `Sequence` and internally sets the `rbf` value. If we go this way, it's likely we'd then have to remove the error `CreateTxError::RbfSequence` that tries to stop you from setting an rbf value too high. Let me know what you think.

notmandatory:
I added this to the 1.0.0-beta milestone since it's an breaking API change that should be done for 1.0.

crates/wallet/src/wallet/tx_builder.rs
evanlinjin: If `RbfValue` does impl `Default`, this field will still [default to `None`](https://doc.rust-lang.org/std/option/enum.Option.html#impl-Default-for-Option%3CT%3E).

I'm thinking to do the following:

```rust
// Default is none.
sequence_policy: Option<SequencePolicy>

enum SequencePolicy {
    SignalNoRbf,
    Exact(Sequence),
}
```

`SequencePolicy::SignalNoRbf` Behavior:

* Set `nSequence` to `0xFFFFFFFE` if we have `nLockTime` set. (https://github.com/Bitcoin/bips/blob/master/bip-0065.mediawiki#summary)
* Set `nSequence` to `0xFFFFFFFF` if we do not have `nLockTime` set.
* If this conflicts with `TxParams::locktime` or spk satisfaction requirements, error out.

`SequencePolicy::Exact(_)`:

* `nSequence` will be this value.
* If spk has `nSequence` requirements, and this value does not satisfy it, error out.

evanlinjin: I think it makes sense to have corresponding methods `.signal_no_rbf` and `.set_exact_sequence`.

evanlinjin: If no `SequencePolicy` is defined, and spk has no `OP_CHECKSEQUENCEVERIFY`, default to `0xFFFFFFFD` imo.

ValuedMammal: Yes, we want to be able to derive Default for TxParams, so why not make `RbfValue` (or whatever we decide to call it) into a tuple struct that holds a `Sequence` and has a default value of 0xFFFFFFFD. Then the field of TxParams can just be `rbf: RbfValue` instead of an Option.

luisschwab: I agree, my implementation of Default is too verbose for what it accomplishes. 

evanjlin: you mean removing `enable_rbf`, `enable_rbf_with_sequence` and `disable_rbf`, and then adding `signal_no_rbf` and `set_exact_sequence`?

ValuedMammal: I think this is a good approach. I'm inclined to go with it, unless we plan to change anything else that touches this part of the code.

evanlinjin: > Yes, we want to be able to derive Default for TxParams, so why not make `RbfValue` (or whatever we decide to call it) into a tuple struct that holds a `Sequence` and has a default value of 0xFFFFFFFD. Then the field of TxParams can just be `rbf: RbfValue` instead of an Option.

So by default we want to set `nSequence` to `0xFFFFFFFD` (according to the `RbfValue` default). What if the SPK we are spending contains a relative timelock? Who do we listen to? Our default `RbfValue` setting? Or the SPK?

evanlinjin: @ValuedMammal I think it makes sense to have `RbfValue` (or whatever we name it) be wrapped in `Option`, where `None` is the default (by default).

`None` would just mean set to the default enable RBF flag (`0xFFFFFFFD`), unless if a RTL is needed as according to the SPK we are spending.

ValuedMammal: I struggle to think of a reason why a user would actually need a method like `disable_rbf` when we have `set_exact_sequence` (or what I'm calling `nsequence` in keeping with tradition). With this PR we will automatically be compliant with nodes that enforce explicit signaling under bip125. To the rest of the network, this signaling mechanism via the nSequence field is not relevant for evaluating replacements (AFAIK). Hence my claim that neither `enable_rbf` nor `disable_rbf` are necessary parts of the API. I agree that if the descriptor requires an absolute locktime, then we should disallow setting `Sequence::MAX`.

ValuedMammal:
How's it going on this @luisschwab ? I hope we didn't derail progress too much with the previous discussion. I like that you made an effort not to change the existing API - for what it's worth, the bdk team had a talk and determined that if we're going to break the API as it relates to enabling RBF then it's better to do it now while still in beta. While the implementation here looks correct I'd prefer that we get the API sorted out rather than postpone it.

After thinking more about it I want to update my suggestion to instead change the TxParams `rbf` field to `sequence: Option<Sequence>` and simply remove the `RbfValue` type from the codebase, the rationale being that if we enable RBF by default then the TxBuilder API only needs to accommodate requesting a specific nSequence. In my opinion this is simpler, but let me know if that sounds radically off-course.

luisschwab:
Sorry, I've been busy these last few days, but will resume work on this later today. Yes, since we're still in beta it's best to get it right rather than breaking the API later.

evanlinjin:
@ValuedMammal I agree with your thinking. `Option<Sequence>` makes sense to me. People who want to signal no RBF can use `0xFFFFFFFE`.

luisschwab:
Sounds good, I'll update the PR taking this into consideration.

evanlinjin:
@luisschwab how is progress on this so far? Do you need a hand with this?

crates/wallet/src/wallet/mod.rs
```diff
@@ -1367,37 +1367,20 @@ impl Wallet {
...
-            // RBF enabled with the default value with CSV also enabled. CSV takes precedence
-            (Some(tx_builder::RbfValue::Default), Some(csv)) => csv,
-            // Valid RBF, either default or with a specific value. We ignore the `CSV` value
-            // because we've already checked it before
-            (Some(rbf), _) => rbf.get_value(),
+            // Custom nSequence value
+            (Some(sequence), None) => sequence,
+            // CSV takes precedence over nSequence
+            (_, Some(csv)) => csv,               
         };
```
ValuedMammal, line 1383: It's an interesting choice getting rid of `CreateTxError::RbfSequenceCsv` and just forcing the new tx input to obey the descriptor. One potential issue I see is that it seems to not allow setting a sequence value _higher_ than what's required. Technically this is possible on current master, for example one could imagine a test like this

```rust
#[test]
fn test_create_tx_custom_csv() {
    // desc: wsh(and_v(v:pk(cVpPVruEDdmutPzisEsYvtST1usBR3ntr8pXSyt6D2YYqXRyPcFW),older(6)))
    let (mut wallet, _) = get_funded_wallet(get_test_single_sig_csv());
    let addr = wallet.next_unused_address(KeychainKind::External);
    builder
        .set_exact_sequence(Sequence(42))
        .add_recipient(addr.script_pubkey(), Amount::from_sat(25_000));
    let psbt = builder.finish().unwrap();
    // we allow setting a sequence higher than required
    assert_eq!(psbt.unsigned_tx.input[0].sequence, Sequence(42));
}
```

ValuedMammal: I think one way to achieve it while still restricting the requested sequence to at least the value dictated by the descriptor is to do something like the following

```rust
let n_sequence = match (params.sequence, requirements.csv) {
    // enable rbf by default
    (None, None) => Sequence::ENABLE_RBF_NO_LOCKTIME,
    // none requested, use required
    (None, Some(csv)) => csv,
    // requested sequence incompatible with requirements
    (Some(seq), Some(csv)) if !check_nsequence_rbf(seq, csv) => {
        return Err(CreateTxError::RbfSequenceCsv { rbf: seq, csv })
    }
    // use requested
    (Some(seq), _) => seq,
};
```

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -554,23 +552,9 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
-    /// This can cause conflicts if the wallet's descriptors contain an "older" (OP_CSV) operator
-    /// and the given `nsequence` is lower than the CSV value.
-    ///
-    /// If the `nsequence` is higher than `0xFFFFFFFD` an error will be thrown, since it would not
-    /// be a valid nSequence to signal RBF.
-    pub fn enable_rbf_with_sequence(&mut self, nsequence: Sequence) -> &mut Self {
-        self.params.rbf = Some(RbfValue::Value(nsequence));
+    /// Set an exact nSequence value
+    pub fn set_exact_sequence(&mut self, n_sequence: Sequence) -> &mut Self {
+        self.params.sequence = Some(n_sequence);
```
ValuedMammal: It would still be valuable to have in the docs that the given sequence value may conflict with the descriptor if it contains an "older" fragment

ValuedMammal (commented):
Try to fix the formatting checks caught by CI, and also I left a few comments that I wonder if you or anyone has an opinion on.

crates/wallet/src/wallet/mod.rs
luisschwab: True, I overlooked this.

ValuedMammal (approved):

luisschwab:
Forgot to rename some things.

luisschwab:
Rebased due to 519728cd596c3bda60c6e70cbc12b611017e2d7f

crates/wallet/tests/wallet.rs
```diff
@@ -654,58 +668,33 @@ fn test_create_tx_with_default_rbf_csv() {
...
-fn test_create_tx_with_custom_rbf_csv() {
-    let (mut wallet, _) = get_funded_wallet(get_test_single_sig_csv());
-    let addr = wallet.next_unused_address(KeychainKind::External);
-    let mut builder = wallet.build_tx();
-    builder
-        .add_recipient(addr.script_pubkey(), Amount::from_sat(25_000))
-        .enable_rbf_with_sequence(Sequence(3));
-    assert!(matches!(builder.finish(),
-        Err(CreateTxError::RbfSequenceCsv { rbf, csv })
-        if rbf.to_consensus_u32() == 3 && csv.to_consensus_u32() == 6));
```
ValuedMammal, line 676: 📌 Just noting this test should still be valid using the new API, but to avoid confusion it could be renamed to reflect the existing logic, something like

- `test_create_tx_incompatible_csv`
- `test_create_tx_check_nsequence_rbf`

ValuedMammal (approved):
ACK f6fa15e036814c60f459f6ec8a45f300917b6904

In the interest of building the whole workspace I suppose f6fa15e036814c60f459f6ec8a45f300917b6904 could be squashed in.

luisschwab:
Re-added test and squashed commits.

notmandatory (approved):
ACK f15551b56fb4eb91bc0be5f940950c40ac465084

Looks good!

notmandatory:
This change was also recommended by external code audit: "RBF signalling should probably be turned on by default. First because it could be misinterpreted by users as "this way my transaction has significantly less chances of getting replaced" while most of the network hashrate runs with full-RBF and from 28.0 it's also the Bitcoin Core default (nodes will relay replacements for transaction not explicitly signalling RBF). It is also a fingerprint as around 70% of transactions signal RBF right now, and this number is growing."

Issue #1598, opened 2024-09-09
electrum_client: disrepancies w/ certificate verification between `use-rustls` & `use-openssl`


pythcoiner:
**Describe the bug**  
trying to connect to an electrum server w/  a self-signed certificate using `use-rustls` fails but succeed w/ `use-openssl`

**To Reproduce**  
code snippet:
```rust
use electrum_client::{ConfigBuilder, ElectrumApi};

fn main() {
    let config = ConfigBuilder::new()
        .retry(1)
        .timeout(Some(1))
        .validate_domain(false)
        .socks5(None)
        .build();
    let client =
        electrum_client::Client::from_config("ssl://testnet.aranguren.org:51002", config).unwrap();

    match client.ping() {
        Ok(_) => println!("Pong"),
        Err(e) => {
            println!("{:?}", e);
        }
    }
}

```

**Expected behavior**  
should succeed w/ both features, `use-rusttls` fails w/:
```
AllAttemptsErrored([IOError(Kind(UnexpectedEof))])
```

trying w/ a local electrs + nginx + self-sign certificate i got this in nginx error log w/ `use-rustls`:
```
2024/09/09 10:31:50 [info] 493372#493372: *16 client 192.168.1.6:56774 connected to 0.0.0.0:60002
2024/09/09 10:31:50 [debug] 493372#493372: *16 posix_memalign: 00005583228C41D0:256 @16
2024/09/09 10:31:50 [debug] 493372#493372: *16 posix_memalign: 00005583227FA5D0:256 @16
2024/09/09 10:31:50 [debug] 493372#493372: *16 generic phase: 0
2024/09/09 10:31:50 [debug] 493372#493372: *16 generic phase: 1
2024/09/09 10:31:50 [debug] 493372#493372: *16 generic phase: 2
2024/09/09 10:31:50 [debug] 493372#493372: *16 generic phase: 3
2024/09/09 10:31:50 [debug] 493372#493372: *16 generic phase: 4
2024/09/09 10:31:50 [debug] 493372#493372: *16 tcp_nodelay
2024/09/09 10:31:50 [debug] 493372#493372: *16 posix_memalign: 00005583227FAC90:256 @16
2024/09/09 10:31:50 [debug] 493372#493372: *16 SSL_do_handshake: -1
2024/09/09 10:31:50 [debug] 493372#493372: *16 SSL_get_error: 1
[...]

PR #1620, opened 2024-09-20
feat(bdk_electrum): add `use-openssl` as a feature

oleonardolima:
partially addresses #1598

### Description

It's a simple PR to expose the `use-openssl` from `electrum-client` to `bdk_electrum`. It partially addresses #1598, allowing the user to use `openssl` as an alternative to `rustls`, as there are some problems with it when handling some types of TLS certificates.

Do we need to add some sort of `bdk_electrum` tests using the new exposed `use-openssl` feature ?

### Changelog notice

- Adds `use-openssl` as feature to `bdk_electrum`, exposing `electrum-client` `use-openssl` feature.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/electrum/Cargo.toml
+++ b/crates/electrum/Cargo.toml
@@ -24,6 +24,7 @@ bdk_chain = { path = "../chain" }
 default = ["use-rustls"]
 use-rustls = ["electrum-client/use-rustls"]
 use-rustls-ring = ["electrum-client/use-rustls-ring"]
+use-openssl = ["electrum-client/use-openssl"]
 
 [[test]]
 name = "test_electrum"


notmandatory (approved):
tACK 19849bdf09b9c2cbe5c267c7e6dde4561e830149

I tested with `example_electrum` and confirmed that if I use the default `bdk_electrum` features ("use-rustls") then I'm unable to sync to "ssl://electrum.blockstream.info:60002". But if I disable default features and enable the new "use-openssl" feature then I am able to sync to "ssl://electrum.blockstream.info:60002".

notmandatory:
I just have one question, should we make `use-openssl` the default instead of `use-rustls` since it seems to be more compatible with self-signed certs?

oleonardolima:
> I just have one question, should we make `use-openssl` the default instead of `use-rustls` since it seems to be more compatible with self-signed certs?

I'm fine with it, but I'm not sure about other impacts. Do bindings and other users rely on the default feature being `rustls`? Does `OpenSSL` work fine on bindings/mobile stuff? cc @thunderbiscuit @reez

tnull:
> Does `OpenSSL` work fine on bindings/mobile stuff?

I can only speak to this part, and the answer is: no, unfortunately it doesn't work on mobile as nobody has figured out a way to have it link in the right libraries when cross-compiling for Android. It would be great if we could (which would allow dropping a bunch of dependencies and would be more secure), but so far we seem to be stuck on `rustls`.

thunderbiscuit:
Yes I second what tnull said: open-ssl is not currently a solution for the Android libraries as we built them. It's not impossible to get it to work and would be great to figure it out one day, but as is `rustls` is our best option.

@oleonardolima to answer your question "Do bindings and other users rely on the default feature being rustls?" I think the answer is no (at least for bdk-ffi). We currently disable the default features and simply enable `use-rustls-ring`, so changing the default for the bdk client would not impact us. If you think it's a better default to use `open-ssl`, I'm open to it.

notmandatory:
I had to rebase this one due to a small merge conflict.

oleonardolima:
I had to rebase it again to solve the CI issue (although I think a re-run would've solved it 😅 ).

oleonardolima:
Thanks for the comment @tnull and @thunderbiscuit, I wasn't aware of it.

PR #1621, opened 2024-09-20
DEMO: Depend on a new psbt-v0 crate

tcharding:
EDIT: Please see simpler idea posted below before bothering to read this.

**TL;DR What if we created a`psbt` crate in the `github.com/rust-bitcoin` org?**

This is just an idea, raising a PR to see if you guys would likely get behind it and/or involved.

This PR shows the diff required to make it work. The diff is on purpose verbose, it could be made much smaller. I did it like this as a demo to make sure I got everything.

I've hacked up a new `psbt-v0` crate with the goals of:

- Create crate directly from `rust-bitcoin v0.32.2` and `rust-miniscript 12.0.0` with minimal changes so it can be diff'ed an verified.
- Be a drop in replacement for the `psbt` modules in `bitcoin` and `miniscript`.

This hopes to enable the following:

- Make development on PSBT separate from development on `rust-bitcoin`
- Get a different set of maintainers involved
- Make development of PSBT crate more rapid, more fun, and more productive

My claim is that PSBT has different stakeholders to other code in `bitcoin` and `miniscript` and the fact that it lives in these two crates (specifically `bitcoin`) makes it harder to attract devs and harder to patch than necessary.

github: https://github.com/tcharding/rust-psbt-v0

Potential Roadmap:

`v0.1.0` would be the drop in replacement with minimal changes. Next we could potentially do:

- Re-write signing API
- Implement PSBTv2
- Re-write the whole thing with what we've learned over the last few years

### notes on testing

If you want to prove this works you'll need something like
- https://github.com/tcharding/rust-psbt-v0/pull/6
- https://github.com/rust-bitcoin/rust-miniscript/pull/750

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,3 +25,6 @@ authors = ["Bitcoin Dev Kit Developers"]
 [workspace.lints.clippy]
 print_stdout = "deny"
 print_stderr = "deny"
+
+[patch.crates-io.miniscript]
+path = "/home/tobin/build/github.com/tcharding/rust-miniscript/master"
--- a/crates/wallet/Cargo.toml
+++ b/crates/wallet/Cargo.toml
@@ -23,13 +23,14 @@ serde = { version = "^1.0", features = ["derive"] }
 serde_json = { version = "^1.0" }
 bdk_chain = { path = "../chain", version = "0.19.0", features = [ "miniscript", "serde" ], default-features = false }
 bdk_file_store = { path = "../file_store", version = "0.16.0", optional = true }
+psbt-v0 = { path = "/home/tobin/build/github.com/tcharding/rust-psbt-v0/master", version = "0.1.0", features = ["miniscript"], default-features = false }
 
 # Optional dependencies
 bip39 = { version = "2.0", optional = true }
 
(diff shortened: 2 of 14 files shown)


tcharding:
Had an even simpler idea:
1. Re-expose `bitcoin::psbt::*` from `miniscript`, and release in minor version
2. Patch `bdk` to only use PSBT stuff from `miniscript` (internal change only)
3. Create `psbt-v0` which contains stuff imported from `bitcoin::psbt`
4. Patch `miniscript` to use `psbt-v0`, release in minor version [0]
5. Deprecate `bitcoin::psbt` and release minor version, having no effect on `bdk` - WIN

[0] - Open question, is step 4 semver compliant, and if not can we fudge it like this since we know the code is the same?

apoelstra:
Step 4 unfortunately will *not* be semver-compliant, because it will change a `psbt` re-export which is compatible with rust-bitcoin to one that isn't.

tcharding:
hmm, yeah and this would break builds for other people. Drats.

tcharding:
Demo not needed now, crate is live.

PR #1623, opened 2024-09-26
fix!(wallet): `ChangeSet` should not be non-exhaustive

### LINKED ISSUE #1591: Should `bdk_wallet::ChangeSet` really be `non_exhaustive`?
**Describe the enhancement**  
The `non_exhaustive` attribute disallows the construction of a type via struct expressions to make sure existing code doesn't break at compile time when new variants are added. 

However, in case of `ChangeSet` I'd want to raise the question if we really want this behavior, or if you'd *explicitly want* existing code to break at compile time so that, if you implement custom serialization logic, you're forced to look at how to deal with the newly added field during version upgrades.
 
Otherwise, newly added field might get unnoticed by custom implementations as backwards compatibility should ensure prior code continues to be working, although then possibly not with the full feature set of the upgraded version.

FWIW, it's currently also inconsistent across `ChangeSet` types as all others are *not* `non_exhaustive`, meaning that any custom serialization implementation would break anyways as soon fields are added to one of them, just in case of `bdk_wallet::ChangeSet` the API is different.

**Use case**  
Implementing custom serialization logic. Currently, you have to call `ChangeSet::default()` and override any deserialized fields manually.

evanlinjin:
Closes #1591 

### Description

As stated in #1591, it is handy to know when new values are added to `ChangeSet`.

### Notes to the reviewers

~

### Changelog notice

* Remove `#[non_exhaustive]` attribute from `wallet::ChangeSet`.

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/changeset.rs
+++ b/crates/wallet/src/wallet/changeset.rs
@@ -8,7 +8,6 @@ type IndexedTxGraphChangeSet =
 
 /// A changeset for [`Wallet`](crate::Wallet).
 #[derive(Default, Debug, Clone, PartialEq, serde::Deserialize, serde::Serialize)]
-#[non_exhaustive]
 pub struct ChangeSet {
     /// Descriptor for recipient addresses.
     pub descriptor: Option<Descriptor<DescriptorPublicKey>>,


oleonardolima (approved):
utACK d7dfe38ff6c0f1b984e15defc83c79e52a44afa9

ValuedMammal (approved):
ACK d7dfe38ff6c0f1b984e15defc83c79e52a44afa9

PR #1632, opened 2024-09-29
Recover file store previous state at open

### LINKED ISSUE #1517: Appending changesets to `Store` after opening causes overwriting
**Describe the bug**  
* let's say a `Store` instance initially contains `changeset1` stored in its file.
https://github.com/bitcoindevkit/bdk/blob/d99b3ef4b449a5c7b8facf04db3cb9be829a0ac1/crates/file_store/src/store.rs#L14

* Then after opening the store instance using `Store::open` and appending a new changeset (`changeset2`) using `append_changeset`, the store will only contain `changeset2`, not `changeset1 + changeset2`.

https://github.com/bitcoindevkit/bdk/blob/d99b3ef4b449a5c7b8facf04db3cb9be829a0ac1/crates/file_store/src/store.rs#L73
https://github.com/bitcoindevkit/bdk/blob/d99b3ef4b449a5c7b8facf04db3cb9be829a0ac1/crates/file_store/src/store.rs#L162

**To Reproduce**  
 ``` rust
   #[test]
    fn test() {
        let temp_dir = tempfile::tempdir().unwrap();
        let file_path = temp_dir.path().join("db_file");
        let changeset1 = BTreeSet::from(["hello".to_string(), "world".to_string()]);
        let changeset2 = BTreeSet::from(["got the error".to_string()]);

        {
            // create new  db
            let mut db = Store::<TestChangeSet>::create_new(&TEST_MAGIC_BYTES, &file_path)
                .expect("must create");

            // append first changeset to db
            db.append_changeset(&changeset1).expect("must succeed");
        }

        {
            // open db
            let mut db = Store::<TestChangeSet>::open_or_create_new(&TEST_MAGIC_BYTES, &file_path)
                .expect("failed to load db");

            // now append the second changeset
            db.append_changeset(&changeset2).expect("must succeed");

            // Retrieve stored changesets from the database
            let stored_changesets = db
                .aggregate_changesets()
                .expect("must succeed")
                .expect("must succeed");

            // expected changeset must  be changeset2 + changeset1
            let mut expected_changeset = changeset2.clone();
            expected_changeset.extend(changeset1);

            // Assert that stored_changesets matches changeset2 but not expected_changeset
            assert_eq!(stored_changesets, changeset2);
            assert_ne!(stored_changesets, expected_changeset);
        }

        // Reason:
        //
        // On loading the pre-existing store instance,
        // it contains all the pre-appended changesets but the file pointer is positioned back to 12
        // i.e till the size of TEST_MAGIC_BYTES array.
        // thus on appending any new changeset to db -> it overwrites those past changesets
        // thus only contain the latest changeset.

        // Open the store again to verify file pointer position at the end of the file
        let mut db =
            Store::<TestChangeSet>::open(&TEST_MAGIC_BYTES, &file_path).expect("failed to load db");

        // get the current position of file pointer just after loading store
        let current_pointer = db.db_file.stream_position().expect("must suceed");

        assert_eq!(current_pointer, 12);

        // end pointer for the loaded db
        let expected_pointer = db.db_file.seek(SeekFrom::End(0)).expect("must succeed");

        // both doesn't match
        assert_ne!(current_pointer, expected_pointer);
    }
  ```

**Expected behavior**  
   * On opening the `store` -> its file pointer must be set to end.

**Build environment**  
 - BDK tag/commit: d99b3ef4b449a5c7b8facf04db3cb9be829a0ac1
 - OS+version: `Fedora 37`
 - Rust/Cargo version: `rustc 1.81.0-nightly (f21554f7f 2024-06-08)`
 - Rust/Cargo target: `nightly-2024-06-09-x86_64-unknown-linux-gnu `

nymius:
### Description

The `Store::open` method doesn't recovers the previous `Store` state saved in the file and emplaces the file pointer just after the magic bytes prefix, this later is agravated by `Store::append_changeset` which sets the file pointer after the last written changeset. The combination of both causes the lost of any previous changeset there may have been in the file.

Is natural to think this shouldn't be the expected behavior, as @KnowWhoami pointed out in #1517, and the `Store` should recover the previous changesets stored in the file store.

To avoid producing breaking changes the expected logic has been implemented in a new `Store::reopen` method, which opens the file, iterates the changesets and returns a new `Store` with the file pointer correctly placed at the end of the already existing changesets. As this method checks the integrity of the stored changesets while iterating them, if any error is found, the method will return a `StoreError::EntryIter` error with the index of the offending changeset, the error caused by it and the number of bytes read when the error happened.

The `StoreError` is a new error enum including the following variants:
- `EntryIter`: any error related to the decoding of the changesets of the file.
- `Io`: errors related with file management internals.
- `InvalidMagicBytes`: error caused by mismatching with the expected magic bytes.
  
`StoreError::Io` and `StoreError::InvalidMagicBytes` are the same than `FileError::Io` and `FileError::InvalidMagicBytes`.  My idea is to propose `StoreError` as a replacement of `FileError` in the future. It has been implemented in this way to avoid the addition of extra nested variants at the moment of matching the error. 

Fixes #1517.

### Notes to the reviewers

I have committed the changes as `feat` and not `fix` because, although this PR fixes #1517, the changes do not imply a change in the current `Store` methods, but the addition of new one and a new `Error` enum.

### Changelog notice

- New `Store::reopen` method in `file_store`.
- New `StoreError` enum in `file_store`.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/file_store/src/store.rs
+++ b/crates/file_store/src/store.rs
@@ -1,4 +1,4 @@
-use crate::{bincode_options, EntryIter, FileError, IterError};
+use crate::{bincode_options, EntryIter, FileError, IterError, StoreError};
 use bdk_chain::Merge;
 use bincode::Options;
 use std::{
@@ -90,6 +90,62 @@ where
             db_file: f,
             marker: Default::default(),
         })
+    }
+
+    /// Open an existing [`Store`], read its content, and return it ready to start receiving new
+    /// changesets.
+    ///
+    /// Use [`create_new`] to create a new `Store`.
+    ///
+    /// # Errors
+    ///
+    /// If the prefixed bytes of the opened file does not match the provided `magic`, the
+    /// [`StoreError::InvalidMagicBytes`] error variant will be returned.
+    ///
+    /// If there is an error while decoding the changesets stored, the [`StoreError::EntryIter`]
+    /// error variant will be returned, with the index of the failing changeset and the error it
+    /// caused.
+    ///
+    /// [`create_new`]: Store::create_new
+    pub fn reopen<P>(magic: &[u8], file_path: P) -> Result<Self, StoreError>
+    where
+        P: AsRef<Path>,
+    {
+        let mut f = OpenOptions::new().read(true).write(true).open(file_path)?;
+
+        let mut magic_buf = vec![0_u8; magic.len()];
+        f.read_exact(&mut magic_buf)?;
+        if magic_buf != magic {
+            return Err(StoreError::InvalidMagicBytes {
+                got: magic_buf,
+                expected: magic.to_vec(),
+            });
+        }
+
+        let mut store = Self {
+            magic_len: magic.len(),
+            db_file: f,
+            marker: Default::default(),
+        };
+
+        let mut index: usize = 0;
+        let mut error: Option<IterError> = None;
+        for (idx, next_changeset) in store.iter_changesets().enumerate() {
+            if let Err(iter_error) = next_changeset {
+                index = idx;
+                error = Some(iter_error);
+            };
+        }
+
+        if let Some(iter) = error {
+            return Err(StoreError::EntryIter {
+                index,
+                iter,
+                bytes_read: store.db_file.stream_position()?,
+            });
+        }
+
+        Ok(store)
     }
 
     /// Attempt to open existing [`Store`] file; create it if the file is non-existent.
--- a/crates/file_store/src/store.rs
+++ b/crates/file_store/src/store.rs
@@ -494,4 +494,171 @@ mod test {
             assert_eq!(aggregation, exp_aggregation);
         }
     }
+
+    #[test]
+    fn reopen_recovers_state_after_last_write() {
+        let temp_dir = tempfile::tempdir().unwrap();
+        let file_path = temp_dir.path().join("db_file");
+
+        let changeset1 = TestChangeSet::from(["1".into(), "2".into(), "3".into()]);
+        let changeset2 = TestChangeSet::from(["4".into(), "5".into(), "6".into()]);
+
+        {
+            // create new  db
+            let mut db = Store::<TestChangeSet>::create_new(&TEST_MAGIC_BYTES, &file_path)
+                .expect("must create");
+
+            // append first changeset to db
+            db.append_changeset(&changeset1).expect("must succeed");
+        }
+
+        {
+            // open db
+            let mut db = Store::<TestChangeSet>::reopen(&TEST_MAGIC_BYTES, &file_path)
+                .expect("failed to load db");
+
+            // now append the second changeset
+            db.append_changeset(&changeset2).expect("must succeed");
+
+            // Retrieve stored changesets from the database
+            let stored_changesets = db
+                .aggregate_changesets()
+                .expect("must succeed")
+                .expect("must succeed");
+
+            // expected changeset must  be changeset2 + changeset1
+            let mut expected_changeset = changeset2.clone();
+            expected_changeset.extend(changeset1);
+
+            // Assert that stored_changesets matches expected_changeset but not changeset2
+            assert_eq!(stored_changesets, expected_changeset);
+            assert_ne!(stored_changesets, changeset2);
+        }
+
+        // Open the store again to verify file pointer position at the EOF
+        let mut db = Store::<TestChangeSet>::reopen(&TEST_MAGIC_BYTES, &file_path)
+            .expect("failed to load db");
+
+        // get the current position of file pointer just after loading store
+        let current_pointer = db.db_file.stream_position().expect("must suceed");
+
+        // get pointer to last position for the loaded db
+        let expected_pointer = db.db_file.seek(io::SeekFrom::End(0)).expect("must succeed");
+
+        // both should match
+        assert_eq!(current_pointer, expected_pointer);
+    }
+
+    #[test]
+    fn fail_to_reopen_if_write_is_short() {
+        let temp_dir = tempfile::tempdir().unwrap();
+
+        let changesets = [
+            TestChangeSet::from(["1".into()]),
+            TestChangeSet::from(["2".into(), "3".into()]),
+            TestChangeSet::from(["4".into(), "5".into(), "6".into()]),
+        ];
+        let last_changeset = TestChangeSet::from(["7".into(), "8".into(), "9".into()]);
+        let last_changeset_bytes = bincode_options().serialize(&last_changeset).unwrap();
+
+        for short_write_len in 1..last_changeset_bytes.len() - 1 {
+            let file_path = temp_dir.path().join(format!("{}.dat", short_write_len));
+
+            // simulate creating a file, writing data where the last write is incomplete
+            {
+                let mut db =
+                    Store::<TestChangeSet>::create_new(&TEST_MAGIC_BYTES, &file_path).unwrap();
+                for changeset in &changesets {
+                    db.append_changeset(changeset).unwrap();
+                }
+                // this is the incomplete write
+                db.db_file
+                    .write_all(&last_changeset_bytes[..short_write_len])
+                    .unwrap();
+            }
+
+            // Reopen file and fail, but recover once file is truncated to valid bytes
+            {
+                match Store::<TestChangeSet>::reopen(&TEST_MAGIC_BYTES, &file_path) {
+                    Err(StoreError::EntryIter {
+                        index, bytes_read, ..
+                    }) => {
+                        // Open file again and truncate file to valid content
+                        let mut f = OpenOptions::new()
+                            .read(true)
+                            .write(true)
+                            .open(&file_path)
+                            .expect("should open");
+                        f.set_len(bytes_read)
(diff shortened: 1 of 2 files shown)


crates/file_store/src/store.rs
```diff
@@ -92,6 +92,62 @@ where
...
+    /// # Errors
+    ///
+    /// If the prefixed bytes of the opened file does not match the provided `magic`, the
+    /// [`StoreError::InvalidMagicBytes`] error variant will be returned.
+    ///
+    /// If there is an error while decoding the changesets stored, the [`StoreError::EntryIter`]
+    /// error variant will be returned, with the index of the failing changeset and the error it
+    /// caused.
+    ///
+    /// [`create_new`]: Store::create_new
```
LLFourn, line 109: I'm probably missing some background but what's the point of doing this unless you return the data you read?

I think a PR here makes sense because we got rid of the "load" system we had before. Maybe this needs a big re-think including deleting lots of the API here. We probably just want `append_changeset` and `create` and `load` as the API. Does anyone need to actually iter changesets?

KnowWhoami: I agree that the FileStore has APIs that users likely don't need.

IMO, implementing these api's in the following fashion could make `FileStore`  much simpler.

## Key APIs:
- **append_changeset**:
We should ensure that duplicate changesets are not stored. Currently, users must call `aggregate_changesets` to remove duplicates, but the store isn't updated with the final unique changeset. Instead, we can make `append_changeset` automatically append the given changeset, aggregate them to remove duplicates, and then store the final aggregated changeset back into the `FileStore`.

- **load**:
As suggested by @ValuedMammal, it makes sense to open the `FileStore` in `append` mode rather than `write` mode. This will place the file pointer at the end of the file, avoiding overwriting and making it easier to append new changesets.

## Other Considerations:
- **aggregate_changesets**:
While I'm not certain if downstream users will need this API, it seems valuable to keep it exposed for now.

- **Iteration Over Changesets**:
From a user's perspective, iterating changesets is unnecessary. Currently, it's only used to collect changesets for aggregation. We could either make it private or move the iteration logic directly into `aggregate_changesets`.

By handling changeset duplication within append_changeset, we can eliminate the need for users to manually aggregate changesets or iterate over them..

nymius: I agree with your vision.

In this PR I proposed a solution of compromise guided by the principle of stabilizing the API rather than breaking it. I don't know if that's currently something BDK wants to do.

---
>  As suggested by @ValuedMammal, it makes sense to open the `FileStore` in `append` mode rather than `write` mode. This will place the file pointer at the end of the file, avoiding overwriting and making it easier to append new changesets.

For example, I discarded the `append` solution because there seemed to be an expected behavior hidden behind the interaction of two functions:
1. `open` not moving the file pointer to the EOF when accessing the file and,
2. `append_changeset` setting the file pointer (and length) to the end of the last changeset written.

The interaction of both is what made the `append_changeset_truncates_invalid_bytes` test to pass in the original code, because **1** leaves the file pointer at the end of the `MAGIC_BYTES` and **2** removes the invalid bytes completely by shorting the file to the end of the last written changeset.

---
> I'm probably missing some background but what's the point of doing this unless you return the data you read?

That's also why I don't provide the user with the well encoded changesets on error, as that would have implied a more complex structure, using generics (to consider the changeset).
I traded it with the minimum data to locate and isolate the failure in the file.

You may find a clearer idea of this in the `fail_to_reopen_if_write_is_short` test, at [this](https://github.com/bitcoindevkit/bdk/pull/1632/files#diff-a2a5264e9fd43373450a32f3b6981e1e5307bc8fff7bacdb008c7c8a379ef379R581) line.

---
If there is a common agreement to make the refactoring above instead of trying to fix the current API I can close this PR and try to implement the changes in another one.

LLFourn: I don't understand why we want to ensure duplicate changesets are not stored. They do no harm and only happen in the case of some kind of programmer error.

There's no issue with stabalizing the API of this crate I think. We can do what we want with it. To me at least, this `reopen` makes it make even less sense because it reads in every entry but doesn't return it which is a waste.

I agree with @nymius that opening the file with append is a nice idea in theory but it's actually counter productive here. There's no magic to opening files with append it just makes the operating system handle what we can do better.

---

I vote for a PR cutting the API down to:

- `create` -- rename `create_new`.
- `load` -- does what `open` does and aggregates the changeset and leaves the file pointer at the right spot. Errors if the file didn't exist or wrong magic bytes.
- `append` -- does what `append_changeset` does
- `create_or_load` -- does what `open_or_create_new` does

Then updating the README to explain that this is the simplest implementation of a bdk compatible database. It's useful for development but should probably not be used beyond that. No effort has been made to make it backwards compatible so you have to do that yourself. Link them to docs on sqlite feature for what they should use in production.

evanlinjin: I think `load` should essentially return `Result<(C, Self), Error>` so that we can get rid of `.aggregate_changeset`. This forces the caller to only read once.

KnowWhoami: > The interaction of both is what made the append_changeset_truncates_invalid_bytes test to pass in the original code, because 1 leaves the file pointer at the end of the MAGIC_BYTES and 2 removes the invalid bytes completely by shorting the file to the end of the last written changeset.

Got the point.

> I don't understand why we want to ensure duplicate changesets are not stored.

On ensuring this -> we can get rid of iterating over the changeset  each time user want to get the total changeset.

> Instead, we can make append_changeset automatically append the given changeset, aggregate them to remove duplicates, and then store the final aggregated changeset back into the FileStore

Though in my proposed impl ->  we have to iterate at the end -> so doesn't make much diference. 

>  They do no harm and only happen in the case of some kind of programmer error.

Though I have no idea regarding this but if that's the case -> then it does not make sense to  try to keep unique set of changesets.

> I think load should essentially return Result<(C, Self), Error> so that we can get rid of .aggregate_changeset. This forces the caller to only read once.

Combining both the `load/open` and `aggregate_changeset` logic does make sense because IMO, both are correlated to each other as in order to load a wallet , a user has to get the aggregated changeset for which they have to load the `store` instance.
By this, we don't require to create another public api and users don't have to manually call a function to get the required changeset for loading wallet.

evanlinjin: > On ensuring this -> we can get rid of iterating over the changeset each time user want to get the total changeset.

If loading gets too slow, we can just have a method to merge changesets. I.e. overwrite the file with just one entry, the new aggregate changeset. Or aggregate x number of changesets at the end.

It does NOT make sense to have a check that stops duplicate changesets being stored. How are we going to implement this? Have an aggregate changeset of all persisted changes in memory that ensures the new changeset is not a subset of it? That typically doesn't happen, and if it does, it means that there is a problem with the in-memory data structures. The in-memory data structures should only output a changeset because there are changes to it.

nymius: > I think `load` should essentially return `Result<(C, Self), Error>` so that we can get rid of `.aggregate_changeset`. This forces the caller to only read once.

I have found an issue with this approach. The `(C, Self)` return type is hard to adapt to the `WalletPersister::initialize` trait.

By the documentation I don't have available any file path to load the store from and I must return all data stored in the `persister` (i.e. return the `AggregatedChangeset`):
> - The database schema of the `persister` (if any), **should be initialized**...
> - The **implementation must return all data currently stored in the `persister`**. If there is no data, return an empty changeset (using [`ChangeSet::default()`]).
> - **some persister implementations may NOT require initialization at all (and not error)**.

```
    /// [`persist`]: WalletPersister::persist
    fn initialize(persister: &mut Self) -> Result<ChangeSet, Self::Error>;
```

`file_store`'s trait implementation currently in master assumes `persister` (i.e. `file_store::Store`) has been opened or loaded before reaching the trait method (look how `aggregate_changesets()` is called without `create_new/open` ceremony).

```
#[cfg(feature = "file_store")]
impl WalletPersister for bdk_file_store::Store<ChangeSet> {
    type Error = FileStoreError;

    fn initialize(persister: &mut Self) -> Result<ChangeSet, Self::Error> {
        persister
            .aggregate_changesets()
            .map(Option::unwrap_or_default)
            .map_err(FileStoreError::Load)
    }
```

Summing up, the constraints for `file_store::Store` are:
- Must return aggregated changeset.
- Cannot load from file at `WalletPersister::initialize` because `file_path` is not at scope at that moment
- including a `file_path` parameter in `WalletPersister::initialize` will break API and force changes on `rusqlite` implementation.

Following LLForun's idea, I propose the following new API for `file_store::Store`:
- `create`: rename of `create_new`
- `load`: open file, place file pointer at EOF and return `file_store::Store`.
- **`dump`: aggregate changesets and return the aggregated changeset.**
- `append`: rename of `append_changeset`
- `load_or_create`: `Store::create` or `Store::load`.

nymius: Working on this I realized we can have both, `(C, Store)` as return type of `Store::load` and a `Store::dump` method to call on `<bdk_file_store::Store<ChangeSet> as WalletPersister>::initialize`, so I will keep both.

nymius:
Overrode by #1684

PR #1625, opened 2024-09-26
feat(chain,core)!: move `Merge` to `bdk_core`

evanlinjin:
Fixes #1624

### Description

Moving `Merge` into `bdk_core` (from `bdk_chain`) allows persist crates to only depend on `bdk_core`.

### Changelog notice

* Move `Merge` into `bdk_core`.
* Remove `bdk_chain` dependency from `bdk_file_store`.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #1575, opened 2024-08-26
ci: add `workspace` or single `package` to build-test

oleonardolima:
fixes bitcoindevkit/bdk#1944

### Description

I was going through the Project board and gave it a try as I'm already working with CI on `rust-esplora-client`, and it looked simple.

I was wondering if it should really be another job, I felt that adding to the matrix strategy it's simpler and clearer, let me know what you think.

### Notes to the reviewers

I'm happy with either a new job or this approach (if it's sound), please let me know of any flaws with this one.

### Changelog notice

- Adds `--workspace` and `-p <bdk-package>` to CI `build-test` job matrix strategy, to build and test the workspace and every single BDK package. 

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

Issue #1018, opened 2023-06-26
Add CI tests for building individual packages


notmandatory:
**Describe the enhancement**  

As described in #1017 it's possible in the `build-test` job for `--no-default-features` and `--all-features` tests to pass, but building individual packages can still fail. 

To make sure this doesn't happen again we should test building each package individually with `default`, `--all-features`, and `--no-default-features`. 

**Use case**  

Catch broken feature configurations sooner.

oleonardolima:
I gave it a try, and it doesn't seem each package can be built with `--no-default-features`, at least is not the case for `bdk_wallet` because of miniscript `std/no-std` requirement, see https://github.com/bitcoindevkit/bdk/actions/runs/10565729093/job/29270897594?pr=1575

oleonardolima:
Oh, I guess that's the reason we have the `check-no-std` job :think:

notmandatory:
I think we won't be able to fix this issue until rust-bitcoin/rust-miniscript#681 is done.

oleonardolima:
I'll check if the changes from miniscript#681 would unblock this.

PR #1636, opened 2024-10-01
ci: fix build-test job with --no-default-features, add miniscript/no-std

notmandatory:
### Description

Fixes the CI `build-test` job with `--no-default-features` by also adding `--features miniscript/no-std`.

Until `rust-miniscript` removes the `no-std` feature we need to enable it when `--no-default-features` is used to build `bdk_wallet` or the whole workspace. See also the `check-no-std` job which does the same plus enables the `bdk_chain/hashbrown` feature which is also needed to build `bdk_wallet` with `--no-default-features` but is already enabled when building the whole workspace.

### Notes to the reviewers

I think we didn't catch this on bitcoindevkit/bdk#1625 because the CI job names changed and I didn't update the branch merge requirements. Another possibility is it was passing because of cached build artifacts which I removed last night when I was trying to troubleshoot something else. I've updated the required CI jobs that need to pass before allowing a PR to be merged to `master` to include the ones with `--no-default-features --features bdk_chain/hashbrown` in the name.

#### All Submissions:

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -25,7 +25,7 @@ jobs:
             clippy: true
           - version: 1.63.0 # MSRV
         features:
-          - --no-default-features
+          - --no-default-features --features miniscript/no-std
           - --all-features
     steps:
       - name: checkout


ValuedMammal (approved):
ACK 0e8082437401eccdef30737e739381d295530547

oleonardolima (commented):
post-merge ACK 0e8082437401eccdef30737e739381d295530547

I think this approach also partially unblocks #1018 and #1575, should we move forward with those?

PR #1640, opened 2024-10-04
Refactor `TestEnv::wait_until_electrum_sees_block` to be more concrete.

evanlinjin:
### Description

Before, the signature of `TestEnv::wait_until_electrum_sees_block` only took in a `delay: Duration` input. It was not clear which block we were waiting for exactly. This PR adds 2 inputs to this method: a `block_height` target, and an optional `block_hash` method. Just having the `block_height` method waits until Electrs sees a block at that height, and all spk histories are up to date to that block. If the caller also includes a `block_hash`, we wait until Electrs sees a block at the given height, also has that block hash. 

We also introduce a new method: `TestEnv::wait_until_electrum_tip_syncs_with_bitcoind`. The method name is self-explanatory.

### Notes to the reviewers

It's a breaking change for `bdk_testenv`. The API now makes more sense.

### Changelog notice

- Change `TestEnv::wait_until_electrum_sees_block` to have a height target, and an optional block hash target.
- Add `TestEnv::wait_until_electrum_tip_syncs_with_bitcoind`.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/electrum/tests/test_electrum.rs
+++ b/crates/electrum/tests/test_electrum.rs
@@ -297,7 +297,7 @@ pub fn test_update_tx_graph_without_keychain() -> anyhow::Result<()> {
         None,
     )?;
     env.mine_blocks(1, None)?;
-    env.wait_until_electrum_sees_block(Duration::from_secs(6))?;
+    env.wait_until_electrum_tip_syncs_with_bitcoind(Duration::from_secs(6))?;
 
     // use a full checkpoint linked list (since this is not what we are testing)
     let cp_tip = env.make_checkpoint_tip();
@@ -409,7 +409,7 @@ pub fn test_update_tx_graph_stop_gap() -> anyhow::Result<()> {
         None,
     )?;
     env.mine_blocks(1, None)?;
-    env.wait_until_electrum_sees_block(Duration::from_secs(6))?;
+    env.wait_until_electrum_tip_syncs_with_bitcoind(Duration::from_secs(6))?;
 
     // use a full checkpoint linked list (since this is not what we are testing)
     let cp_tip = env.make_checkpoint_tip();
@@ -453,7 +453,7 @@ pub fn test_update_tx_graph_stop_gap() -> anyhow::Result<()> {
         None,
     )?;
     env.mine_blocks(1, None)?;
-    env.wait_until_electrum_sees_block(Duration::from_secs(6))?;
+    env.wait_until_electrum_tip_syncs_with_bitcoind(Duration::from_secs(6))?;
 
     // A scan with gap limit 5 won't find the second transaction, but a scan with gap limit 6 will.
     // The last active indice won't be updated in the first case but will in the second one.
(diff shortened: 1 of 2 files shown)


crates/testenv/src/lib.rs
```diff
@@ -186,18 +186,45 @@ impl TestEnv {
...
+        timeout: Duration,
+    ) -> anyhow::Result<()> {
+        self.electrsd.trigger()?;
+        // NOTE: There is a reason why we use the subscribe endpoint specifically. You may think
+        // just polling Electrs via Electrum for a block header at height `block_height` and
+        // checking whether `block_hash` matches is enough. However, having the header polling call
+        // up to date does not necessarily mean the spk histories are up to date. On the other hand,
+        // getting a notification for a new block tip does mean that the confirmed spk histories
+        // are up to date up to and including the new notified tip. This is all due to the internal
+        // workings of Electrs.
```
evanlinjin: I may need to word this better.

LagginTimes: How about:

```
// NOTE: We use the subscribe endpoint because polling Electrs for a block header at 
// `block_height` and verifying the `block_hash` does not ensure that the spk histories 
// are current. In contrast, getting a notification for a new block tip ensures that the
// confirmed spk histories are current, including the new notified tip. This is a result of
// the internal workings of Electrs.
```

evanlinjin: @LagginTimes thanks for the suggestion, I've used what you wrote.

crates/electrum/tests/test_electrum.rs
```diff
@@ -716,6 +717,7 @@ fn tx_can_become_unconfirmed_after_reorg() -> anyhow::Result<()> {
         // Check that no new anchors are added during current reorg.
         assert!(initial_anchors.is_superset(&update.tx_update.anchors));
 
+        // TODO: Fails here.
```
LagginTimes, line 720: Nit: This is now outdated since the test passes.

crates/testenv/src/lib.rs
```diff
@@ -188,18 +188,43 @@ impl TestEnv {
         Ok((bt.height as usize, block.block_hash()))
     }
 
-    /// This method waits for the Electrum notification indicating that a new block has been mined.
-    /// `timeout` is the maximum [`Duration`] we want to wait for a response from Electrsd.
-    pub fn wait_until_electrum_sees_block(&self, timeout: Duration) -> anyhow::Result<()> {
-        self.electrsd.client.block_headers_subscribe()?;
+    /// Wait until Electrum is aware of a block of a given `block_height` (and optionally, matches
+    /// the given `block_hash`).
+    pub fn wait_until_electrum_sees_block(
```
LagginTimes, line 193: Since `wait_until_electrum_sees_block` now accepts any height and optional hash, would it be helpful to add unit tests for a non-tip height or a mismatched hash?

LagginTimes: Is `wait_until_electrum_sees_block(non_tip_height, Some(hash), …)` still expected to succeed when invoked *after* `wait_until_electrum_tip_syncs_with_bitcoind()`, given that the chain tip has already advanced past `non_tip_height`?

See: https://github.com/LagginTimes/bdk/commit/1a055c7694e3f68d50bd3931686c4cd2ff96b52b for the tests that prompted this question.

oleonardolima (commented):
cACK 0025f14766dc867d783748b83dffdbb7edc1c7c2

evanlinjin:
Replaced by #2231

PR #1641, opened 2024-10-04
PoWChain WIP

evanlinjin:
#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/pow_chain.rs
+++ b/crates/chain/src/pow_chain.rs
@@ -1,0 +1,574 @@
+//! The [`PoWChain`] is an implementation of [`ChainOracle`] that verifies proof of work.
+
+use crate::{BlockId, ChainOracle, CheckPoint, Merge};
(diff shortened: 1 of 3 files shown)


crates/chain/src/pow_chain.rs
```diff
@@ -0,0 +1,574 @@
...
+
+    /// Apply the given `changeset`.
+    pub fn apply_changeset(&mut self, changeset: &ChangeSet) -> Result<(), MissingGenesisError> {
+        let old_tip = self.tip.clone();
+        let new_tip =
+            apply_changeset_to_checkpoint(old_tip, changeset, self.difficulty_map.clone())?;
+        self.tip = new_tip;
+        debug_assert!(self._check_changeset_is_applied(changeset));
+        Ok(())
+    }
```
evanlinjin, line 203: I'm finding this quite hard to reason about and it seems rather dangerous. How do you envision it being used here?

PR #1643, opened 2024-10-08
feat(chain,wallet)!: rm `ConfirmationTime`

evanlinjin:
### Description

This PR removes `ConfirmationTime`, and favors `ChainPosition<ConfirmationBlockTime>` instead. The only difference between these two structures is that `ChainPosition<ConfirmationBlockTime>` contains an additional `BlockHash`. Additionally, `ConfirmationTime` was not used in many places. It was mainly for displaying information in `bdk_wallet::Wallet`.

We also impl `serde::Deserialize` and `serde::Serialize` for `ChainPosition`.

### Changelog notice

* Remove `bdk_chain::ConfirmationTime`. Use `ChainPosition<ConfirmationBlockTime>` in place.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #1644, opened 2024-10-09
Inserting transactions in the `Wallet` should always include a `last_seen`.

### LINKED ISSUE #1642: Transactions inserted with `Wallet::insert_tx` will be replaced by the wallet when creating a new tx.
**Describe the bug**  

When a caller manually inserts a transaction into wallet via `Wallet::insert_tx`, it is an intuitive assumption that the transaction inserted will not be replaced by a new transaction created with `Wallet::build_tx`.

However, the current default behavior of `Wallet::build_tx` replaces the transaction inserted with `Wallet::insert_tx` since purely inserting a transaction without a `last_seen` or `Anchor` to the best chain means that the tx will not be part of the canonical chain. 

Thank you @stevenroose for bringing this to my attention. Am I missing anything here?

**To Reproduce**  

Some simple steps for reproducing.

* Create a wallet with a single UTXO.
* Create a tx (`T1`) from that wallet and send to foreign address A.
* Insert `T1` into the wallet via `Wallet::insert_tx`.
* Create another tx (`T2`) that sends to foreign address B.
* Compare inputs of A & B. They should not be the same - but they are.

**Proposed solutions**

1. Add another parameter to `insert_tx`: `last_seen: Option<u64>`, and document that we need the last_seen value in order for the tx to be seen in the canonical chain.
2. Have wallet keep track of which transactions it has created. We shouldn't be replacing transactions that we create.

I think `1.` should be the solution for now. `2.` will need more thinking.

evanlinjin:
Partially fixes #1642

### Description

As mentioned in #1642, `Wallet::insert_tx` has some confusing behavior. Callers typically expect that anything inserted into the wallet will not be replaced, and transaction inserted later will have precedence (in the case when we have conflicting txs).

This PR modifies the behavior of `Wallet::insert_tx` to always add the current unix timestamp as the `last_seen`-in-mempool value alongside the transaction being inserted. Since this is now a `std`-only method, we also introduce a `Wallet::insert_tx_at` where the `last_seen` timestamp is provided by the caller. We expect callers to only insert txs into the wallet after a successful broadcast. The documentation is updated to reflect this. `Wallet::unbroadcast_transactions` is removed since callers are not expected to store unbroadcasted transactions in wallet anymore.

In addition, the `seen_at` input is made mandatory for `Wallet::apply_update_at`. Without this, callers may end up replacing unconfirmed txs when building txs.

### Notes to the reviewers

I don't think this is the perfect solution because a failed broadcast may be due to a network issue. In this case, the caller may need to persist the transaction external to wallet.

### Changelog notice

* Change `Wallet::insert_tx` to always insert the tx alongside a `last_seen`-in-mempool timestamp (using the current timestamp).
* Add `Wallet::insert_tx_at`, which is a version og `insert_tx` where the user can explicitly set the `last_seen` timestamp.
* Change `Wallet::apply_update_at` to always expect a `last_seen` value (not have it wrapped in `Option`).
* Remove `Wallet::unbroadcast_transactions`.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/chain_data.rs
+++ b/crates/chain/src/chain_data.rs
@@ -1,4 +1,3 @@
-use crate::ConfirmationBlockTime;
 use bitcoin::{OutPoint, TxOut, Txid};
 
 use crate::{Anchor, COINBASE_MATURITY};
@@ -7,6 +6,14 @@ use crate::{Anchor, COINBASE_MATURITY};
 ///
 /// The generic `A` should be a [`Anchor`] implementation.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, core::hash::Hash)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Deserialize, serde::Serialize),
+    serde(bound(
+        deserialize = "A: Ord + serde::Deserialize<'de>",
+        serialize = "A: Ord + serde::Serialize",
+    ))
+)]
 pub enum ChainPosition<A> {
     /// The chain data is seen as confirmed, and in anchored by `A`.
     Confirmed(A),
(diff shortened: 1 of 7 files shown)


oleonardolima (commented):
Concept ACK 4f73a770a789ce7b49c9f816291f7f7db9c8caf2

However, a3d4eef77d427b290e1d0041d3c760b8fe29f6a2 should be dropped as it has another context and is covered in #1643.

I'd try to repurpose the `test_insert_tx_balance_and_utxos` instead of removing it.

notmandatory:
Can you rebase this now that #1658 is merged with just what you need for a3d4eef77d427b290e1d0041d3c760b8fe29f6a2 ?

oleonardolima:
> Can you rebase this now that #1658 is merged with just what you need for [a3d4eef](https://github.com/bitcoindevkit/bdk/commit/a3d4eef77d427b290e1d0041d3c760b8fe29f6a2) ?

After #1658, is https://github.com/bitcoindevkit/bdk/commit/a3d4eef77d427b290e1d0041d3c760b8fe29f6a2 the only commit that interests here? 🤔 (I guess so)

I guess we could close this PR then, as a3d4eef77d427b290e1d0041d3c760b8fe29f6a2 has already been merged in #1643.

notmandatory:
@oleonardolima good catch, I didn't notice a3d4eef77d427b290e1d0041d3c760b8fe29f6a2 was already merged. Closing for now and @evanlinjin please re-open if you don't agree all the commits here were already merged.

PR #1648, opened 2024-10-15
docs: show example migration from bdk to bdk_wallet

### LINKED ISSUE #1606: Create wallet data migration process
**Describe the enhancement**  

Create code and/or documentation to help pre-1.0 `bdk` wallet users migrate wallet data to a new 1.0 `bdk_wallet`.  The essential information that needs to be migrated are the descriptors and their corresponding revealed indexes. Transaction and other wallet data does not need to be migrated, it can be re-loaded from a blockchain client.  

The migration process should be database agnostic. Example docs should at least show how to do it with a SQLite database.

**Use case**  

This is needed to help existing pre-1.0 `bdk` users upgrade to `bdk_wallet` 1.0+. 

**Additional context**

This was requested by @tnull but other users will also need this feature to be able to easily upgrade their projects to the `bdk_wallet` 1.0+ releases. 

One option is to enhance the existing wallet `wallet::export` code to export and import descriptors with their revealed index info.

ValuedMammal:
I created a migration workflow in the form of a runnable example that does everything in one go. In the example we

1. Open an existing sqlite db
2. Get the last revealed addresses
3. Create a new `bdk_wallet::Wallet` with new database
4. Restore previously revealed addresses
5. Sync with electrum

fixes #1606

#### All Submissions:

### DIFF
--- a/example-crates/example_migrate_wallet/src/main.rs
+++ b/example-crates/example_migrate_wallet/src/main.rs
@@ -1,0 +1,131 @@
+use std::collections::{BTreeSet, HashSet};
+use std::io::Write;
+
+use bdk::bitcoin::Network;
+use bdk::database::SqliteDatabase;
+use bdk::wallet::{AddressIndex, Wallet};
+
+use bdk_electrum::{electrum_client, BdkElectrumClient};
+use bdk_wallet::rusqlite;
+
+const EXTERNAL: &str = "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/0/*)";
+const INTERNAL: &str = "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/1/*)";
+const ELECTRUM_URL: &str = "ssl://electrum.blockstream.info:60002";
+const NETWORK: Network = Network::Testnet;
+
+const BDK_DB_PATH: &str = ".bdk-example.sqlite";
+const BDK_WALLET_DB_PATH: &str = ".bdk-wallet-example.sqlite";
+
+// Steps for migrating wallet parameters from the old `bdk` 0.30 to the new `bdk_wallet` 1.0.
+// These steps can be applied to wallets backed by a SQLite database. For example to read an
+// existing database, change `BDK_DB_PATH` above to point to the location of the old database
+// file. You may also want to hard code the remaining parameters (descriptors, network, etc)
+// to fit your setup. Note: because we're migrating to a new database there must not already
+// exist a persisted wallet at the new path `BDK_WALLET_DB_PATH`.
+
+// Usage: `cargo run --bin example_migrate_wallet`
+
+fn main() -> anyhow::Result<()> {
+    // Open old wallet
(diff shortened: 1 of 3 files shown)


notmandatory:
Looks good, simpler and safer than dealing directly with database or intermediate text files as I'd originally proposed. A few suggestions:
1. would read better with more descriptive variable names, ie. `old_wallet`, `new_wallet`
2. should add sanity check asserts after sync for things like balances, indexes; maybe also utxos and tx ids.
3. need to update PR description with sqlite instead of sled
4. make it clearer in code comments that you're opening an "old" (will be) 0.30 wallet and a "new" 1.0 wallet

ValuedMammal:
I tested this by first syncing a `bdk` wallet using [`electrum_backend.rs`](https://github.com/bitcoindevkit/bdk/blob/release/0.29/examples/electrum_backend.rs), adapted to use this tprv

```
tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L
```

and persisted with sqlite. Then I ran the example in this PR, initializing `BDK_DB_PATH` to the old db path.

> should add sanity check asserts after sync for things like balances, indexes; maybe also utxos and tx ids

I added assertions for utxos, balance, and revealed addresses. Just to note: this assumes the old database is up to date and no new information is presented as a result of syncing the new wallet.

I'm wondering whether `example-crates/` is the best place for this compared to maybe a page in Book of BDK - I guess either is fine

notmandatory:
Yes I agree it might make a better book section with companion code to take snippets from.

notmandatory:
Maintenance version v0.30.0 of `bdk` crate with upgrade to rusqlite 0.31 is now published: 

https://crates.io/crates/bdk/0.30.0

example-crates/example_migrate_wallet/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+
+// Usage: `cargo run --bin example_migrate_wallet`
+
+fn main() -> anyhow::Result<()> {
+    // Open old wallet
+    let db = SqliteDatabase::new(BDK_DB_PATH);
+    let old_wallet = Wallet::new(EXTERNAL, Some(INTERNAL), NETWORK, db)?;
+
+    // Get last revealed addresses for each keychain
+    let addr = old_wallet.get_address(AddressIndex::LastUnused)?;
```
notmandatory, line 34: Should sync `old_wallet` to make sure we have an accurate `AddressIndex::LastUnused` or could use `AddressIndex::New` here which makes sure we are working with the last exposed (possibly not yet used) index from `old_wallet`.

example-crates/example_migrate_wallet/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+    assert_eq!(addr.to_string(), last_revealed_internal);
+
+    // Now that we migrated the wallet details, you likely want to rescan the blockchain
+    // to restore the wallet's transaction data. Here we're using the bdk_electrum client
+    // with a gap limit of 10. We also pass the `true` parameter to `full_scan` to indicate
+    // we want to collect additional transaction data (previous txouts) that are important
+    // for a fully functioning wallet.
+
+    let client = BdkElectrumClient::new(electrum_client::Client::new(ELECTRUM_URL)?);
+    let request = new_wallet.start_full_scan().inspect({
```
notmandatory, line 96: Do we really need a `full_scan()` here?  Since we already know the last exposed SPK we should be able to do a `sync()` instead which will be quicker and can still retrieve the previous txouts.

notmandatory:
Overall looks great, short and too the point.  I have a couple suggestions but otherwise this looks ready to go.

notmandatory:
As discussed on call today, this can be closed in favor of  bitcoindevkit/book-of-bdk#81. That PR is already using `sync` instead of `full_scan` and I understand and agree it's best to use `AddressIndex::LastUnused`.

PR #1659, opened 2024-10-24
Draft: Better canonial history construction.

evanlinjin:
### Description

The way we are determining the canonical history of transactions/outputs in `bdk_chain` is inefficient and inflexible:

* The algorithm used in [`TxGraph::try_get_chain_position`](https://github.com/bitcoindevkit/bdk/blob/647d2855941d0c46412efa797db4ae157fb544d5/crates/chain/src/tx_graph.rs#L760) is costly when we have many conflicting unconfirmed txs `O(n^2)`.
* There is only one way (one algorithm) to struct a "canonical history" and transactions with no anchor or last_seen will not be considered to be part of it. Because we use the canonical history for the UTXO set, non-broadcasted transactions added to the `TxGraph` may be replaced. 
  * #1642
  * #1644

To fix this, I have introduced the following in the PR:

* `UnconfirmedOracle` is used to determine the *maybe* canonical transaction between a set of conflicting unconfirmed transactions. It can return `None` to signal that none of the presented transactions are canonical. This can be used to create different policies as to which transactions to include in the canonical history.
* `CanonicalView` represents a canonical view of transactions. The algorithm to create it iterates from the "roots" of the `TxGraph` (roots are either coinbase transactions, or outpoints which we have no residing transaction). Each set of conflicting spends only needs to be iterated over once.

### Notes to the reviewers

This is a proof of concept. Not sure which parts should be included in `v1.0`. Tests are passing (except of doc tests).

### Changelog notice

TODO

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bitcoind_rpc/tests/test_emitter.rs
+++ b/crates/bitcoind_rpc/tests/test_emitter.rs
@@ -5,7 +5,7 @@ use bdk_chain::{
     bitcoin::{Address, Amount, Txid},
     local_chain::{CheckPoint, LocalChain},
     spk_txout::SpkTxOutIndex,
-    Balance, BlockId, IndexedTxGraph, Merge,
+    Balance, BlockId, IndexedTxGraph, LastSeenPrioritizer, Merge,
 };
 use bdk_testenv::{anyhow, TestEnv};
 use bitcoin::{hashes::Hash, Block, OutPoint, ScriptBuf, WScriptHash};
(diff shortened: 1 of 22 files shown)


evanlinjin:
Replaced by #1670. Although it would be nice to reintroduce this `CanonicalView` structure later on. This struct can be reused for generating the UTXO/TXO set and canonical tx list.

PR #1662, opened 2024-10-28
fix(wallet): fix building change signers in `load_with_params`

ValuedMammal:
This fixes an issue that prevented change signers from being built when using the `keymap` method on `LoadParams`. Now we always build a `SignersContainer` with the internal keymap whenever a change descriptor is loaded. Note, the signer may still be "empty" if neither `keymap` or `extract_keys` is used. 

### Notes to the reviewers

Building a `SignersContainer` now happens outside of the [match statement](https://github.com/bitcoindevkit/bdk/blob/b6b767f3fc12062d6787eec8b3e1021458283985/crates/wallet/src/wallet/mod.rs#L557) in `load_with_params` which simplifies the logic such that we can interpret this area of code as "initialize a value for `change_descriptor` and optionally extend the internal keymap, returning a `LoadMismatch` error if the changeset doesn't match the expected params."

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -551,7 +551,9 @@ impl Wallet {
         }
         let signers = Arc::new(SignersContainer::build(external_keymap, &descriptor, &secp));
 
-        let (mut change_descriptor, mut change_signers) = (None, Arc::new(SignersContainer::new()));
+        let mut change_descriptor = None;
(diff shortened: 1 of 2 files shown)


evanlinjin (approved):
ACK b6b767f3fc12062d6787eec8b3e1021458283985

This was my oversight. Thanks for fixing this.

PR #1663, opened 2024-10-28
test(esplora): speed up `test_finalize_chain_update`

ValuedMammal:
The test `test_finalize_chain_update` was taking up a lot of time during testing. This PR reduces the time to run the tests to around 6 seconds. I stripped down the test logic to the bare essentials. I think it is adequate but let me know if you have other ideas.

#### All Submissions:

### DIFF
--- a/crates/esplora/src/async_ext.rs
+++ b/crates/esplora/src/async_ext.rs
@@ -483,186 +483,74 @@ where
 
 #[cfg(test)]
 mod test {
-    use std::{collections::BTreeSet, time::Duration};
+    use std::collections::{BTreeMap, BTreeSet};
+    use std::time::Duration;
 
-    use bdk_chain::{
-        bitcoin::{hashes::Hash, Txid},
-        local_chain::LocalChain,
-        BlockId,
-    };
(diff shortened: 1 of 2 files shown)


evanlinjin (changes requested):
Concept ACK, Approach NACK

IMO the old tests were a bit more thorough with the testing, and also easier to extend with more test cases. For performance, I think the main culprit is the starting and stopping of the test env.

I think it would make sense to make the test cases use relative heights.

```rust
struct TestCase {
    /// Initial checkpoint heights, relative to the starting tip height of the env.
    initial_cps: &'a [i32],
    /// Number of blocks mined since starting env tip. The new tip should be introduced by the update.
    chain_extends_by: i32,
    /// Anchors passed into `chain_update`: `(height_relative_to_starting_tip, txid)`.
    anchors: &[(i32, Txid)],
}
```

Another option is to keep how `TestCase` is currently, reuse the `TestEnv` between tests, but make sure the a `TestCase`'s `initial_env_height` is always equal or greater than the last `TestCase`'s `final_env_height`. I think I like this solution better.

ValuedMammal:
Closing for now but I might make another attempt at it in the future.

PR #1592, opened 2024-09-05
Architectural Decision Records

ValuedMammal:
Opening this up as a record of past decisions. The template is inspired by the one used by uniffi. Whether or not we decide to add these to the repo, it's an opportunity to learn, discuss, and collab when it comes to design and architecture.

The first 2 ADRs pertain to bdk's approach to data persistence. Another recent suggestion was to document how we're thinking about single vs multiple descriptor wallets. Feedback welcome.

fixes #1309

Issue #1398, opened 2024-04-05
Stop pinning everything in CI


LLFourn:
Why do we need all these pins?

We don't have to test on MSRV we just have to check that it builds on MSRV. dev-dependencies shouldn't have to be pinned. Or at least that's the policy I'd like to adopt.

See sophisticated technology on how to do this: https://github.com/bitcoindevkit/coin-select/blob/990226a982bd6036bed5a7674fe57d6f9d48e49e/.github/workflows/test.yml#L40

cc @notmandatory

_Originally posted by @LLFourn in https://github.com/bitcoindevkit/bdk/pull/1397#discussion_r1553437195_

LLFourn:
> > We don't have to test on MSRV we just have to check that it builds on MSRV. dev-dependencies shouldn't have to be pinned.
> 
> I disagree: MSRV means "This project supports/works with this rust version", and, everything untested may be considered broken. I regularly run into MSRV breakages for depedencies that 'check' but don't test their MSRV on all supported platforms.

This "regularly" is quite concerning tbh :sweat_smile:. It's a good point that if our tests don't have the same MSRV as build then there's no way to verify that the tests actually pass on the MSRV. I want to add that this situation would imply a bug in the rust compiler, either at the version in CI or at the MSRV. Either that or some weird rust version based conditional compilation in a dependency.

For me the monotonically growing list of dependency pins and using old versions of dev dependencies is a worse problem than this. I definitely don't take the attitude that "everything untested may be considered broken" when developing in rust. Logic that is not tested can be considered broken but not *everything* :)

Not a hill I'll die on though and if the maintainers who are really affected by this want to take the stricter interpretation of MSRV that's fine with me!

notmandatory:
Since updating MSRV to 1.63 we were able to remove many of the originally pinned versions in CI. As 1.63 gets older this list will likely grow again and we'll need to re-verify they're still needed but otherwise I don't see any other way to reduce the pins except bumping our MSRV at least when ever a new Debian stable release comes out with an updated Rust version.

evanlinjin:
I don't think we need to make sure tests run under MSRV, it just makes testing unnecessarily difficult. We can't use the latest tooling. There is no workaround to this because dev-dependencies cannot be optional.

notmandatory:
[...]

PR #1670, opened 2024-11-03
Introduce `O(n)` canonicalization algorithm

### LINKED ISSUE #1665: Performance issue with `get_chain_position` and associated methods for unconfirmed transactions
**Describe the bug**  

`get_chain_position` for unconfirmed transactions is very inefficient if we have "nested conflicts". Since unconfirmed transactions that spend from evicted transactions are also evicted, we need to iterate over and check all unconfirmed descendants.

This problem is exacerbated in `list_canonical_txs`, `filter_chain_txouts`, etc. as these methods call `get_chain_position` for each transaction.

Worst case is `O(n^2)` for building the canonical tx history.

**Proposed Solution**  

Let's pass an `evicted_txs: &mut HashSet<Txid>` input to `get_chain_position` which takes record of any tx that is deemed to be evicted from the canonical history. Note that descendants of evicted transactions are also evicted and should be included in `evicted_txs`. The logic to fill `evicted_txs` can stop when we insert an already-existing `txid` (since descendants of that `txid` will already be included in `evicted_txs`).

`list_canonical_txs`, `filter_chain_txouts`, etc. methods will construct an `evicted_txs` hashset which is passed to each internal call to `try_get_chain_position`.

evanlinjin:
Fixes bitcoindevkit/bdk#1665
Replaces bitcoindevkit/bdk#1659

### Description

Previously, getting the canonical history of transactions/UTXOs required calling `TxGraph::get_chain_position` on each transaction. This was highly inefficient and resulted in an `O(n^2)` algorithm. The situation is especially problematic when we have many unconfirmed conflicts.

This PR introduces an `O(n)` algorithm to determine the canonical set of transactions in `TxGraph`. The algorithm's premise is as follows:

1. If transaction `A` is determined to be canonical, all of `A`'s ancestors must also be canonical.
2. If transaction `B` is determined to be NOT canonical, all of `B`'s descendants must also be NOT canonical.
3. If a transaction is anchored in the best chain, it is canonical.
4. If a transaction conflicts with a canonical transaction, it is NOT canonical.
5. A transaction with a higher last-seen has precedence.
6. Last-seen values are transitive. A transaction's collective last-seen value is the max of it's last-seen value and all of it's descendants.

We maintain two mutually-exclusive `txid` sets: `canoncial` and `not_canonical`.

Imagine a method `mark_canonical(A)` that is based on premise 1 and 2. This method will mark transaction `A` and all of it's ancestors as canonical. For each transaction that is marked canonical, we can iterate all of it's conflicts and mark those as `non_canonical`. If a transaction already exists in `canoncial` or `not_canonical`, we can break early, avoiding duplicate work.

This algorithm iterates transactions in 3 runs.

1. Iterate over all transactions with anchors in descending anchor-height order. For any transaction that has an anchor pointing to the best chain, we call `mark_canonical` on it. We iterate in descending-height order to reduce the number of anchors we need to check against the `ChainOracle` (premise 1). The purpose of this run is to populate `non_canonical` with all transactions that directly conflict with anchored transactions and populate `canonical` with all anchored transactions and ancestors of anchors transactions (transitive anchors).
2. Iterate over all transactions with last-seen values, in descending last-seen order. We can call `mark_canonical` on all of these that do not already exist in `canonical` or `not_canonical`.
3. Iterate over remaining transactions that contains anchors (but not in the best chain) and have no last-seen value. We treat these transactions in the same way as we do in run 2.

#### Benchmarks

Thank you to @ValuedMammal for working on this.

```sh
$ cargo bench -p bdk_chain --bench canonicalization
```

Benchmark results (this PR):

```
many_conflicting_unconfirmed::list_canonical_txs                                                                            
                        time:   [709.46 us 710.36 us 711.35 us]
many_conflicting_unconfirmed::filter_chain_txouts                                                                            
                        time:   [712.59 us 713.23 us 713.90 us]
many_conflicting_unconfirmed::filter_chain_unspents                                                                            
                        time:   [709.95 us 711.16 us 712.45 us]
many_chained_unconfirmed::list_canonical_txs                                                                             
                        time:   [2.2604 ms 2.2641 ms 2.2680 ms]
many_chained_unconfirmed::filter_chain_txouts                                                                             
                        time:   [3.5763 ms 3.5869 ms 3.5979 ms]
many_chained_unconfirmed::filter_chain_unspents                                                                             
                        time:   [3.5540 ms 3.5596 ms 3.5652 ms]
nested_conflicts_unconfirmed::list_canonical_txs                                                                            
                        time:   [660.06 us 661.75 us 663.60 us]
nested_conflicts_unconfirmed::filter_chain_txouts                                                                            
                        time:   [650.15 us 651.36 us 652.71 us]
nested_conflicts_unconfirmed::filter_chain_unspents                                                                            
                        time:   [658.37 us 661.54 us 664.81 us]
```

Benchmark results (master): https://github.com/evanlinjin/bdk/tree/fix/1665-master-bench

```
many_conflicting_unconfirmed::list_canonical_txs                                                                             
                        time:   [94.618 ms 94.966 ms 95.338 ms]
many_conflicting_unconfirmed::filter_chain_txouts                                                                             
                        time:   [159.31 ms 159.76 ms 160.22 ms]
many_conflicting_unconfirmed::filter_chain_unspents                                                                             
                        time:   [163.29 ms 163.61 ms 163.96 ms]

# I gave up running the rest of the benchmarks since they were taking too long.
```

### Notes to the reviewers

* ***PLEASE MERGE bitcoindevkit/bdk#1733 BEFORE THIS PR!*** We had to change the signature of `ChainPosition` to account for transitive anchors and unconfirmed transactions with no `last-seen` value.

* The canonicalization algorithm is contained in `/crates/chain/src/canonical_iter.rs`.

* Since the algorithm requires traversing transactions ordered by anchor height, and then last-seen values, we introduce two index fields in `TxGraph`; `txs_by_anchor` and `txs_by_last_seen`. Methods `insert_anchor` and `insert_seen_at` are changed to populate these index fields.

* An ADR is added: `docs/adr/0003_canonicalization_algorithm.md`. This is based on the work in bitcoindevkit/bdk#1592.

### Changelog notice

* Added: Introduce an `O(n)` canonicalization algorithm. This logic is contained in `/crates/chain/src/canonical_iter.rs`.
* Added: Indexing fields in `TxGraph`; `txs_by_anchor_height` and `txs_by_last_seen`. Pre-indexing allows us to construct the canonical history more efficiently.
* Removed: `TxGraph` methods: `try_get_chain_position` and `get_chain_position`. This is superseded by the new canonicalization algorithm.

#### All Submissions:

#### New Features:

#### Bugfixes:


crates/chain/src/tx_graph.rs
```diff
@@ -135,15 +139,15 @@ impl<A: Ord + Clone> From<TxUpdate<A>> for TxGraph<A> {
...
 pub struct TxGraph<A = ()> {
-    // all transactions that the graph is aware of in format: `(tx_node, tx_anchors)`
-    txs: HashMap<Txid, (TxNodeInternal, BTreeSet<A>)>,
+    txs: HashMap<Txid, TxNodeInternal>,
     spends: BTreeMap<OutPoint, HashSet<Txid>>,
-    anchors: BTreeSet<(A, Txid)>,
+    anchors: HashMap<Txid, BTreeSet<A>>,
     last_seen: HashMap<Txid, u64>,
 
     // This atrocity exists so that `TxGraph::outspends()` can return a reference.
```
notmandatory: No need to call this an "atrocity" since it's the best way to do it.
```suggestion
    // This exists so that `TxGraph::outspends()` can return a reference to an empty set of `Txid`s.
```

crates/chain/src/tx_graph.rs
```diff
@@ -135,15 +139,15 @@ impl<A: Ord + Clone> From<TxUpdate<A>> for TxGraph<A> {
...
     // FIXME: This can be removed once `HashSet::new` is a const fn.
     empty_outspends: HashSet<Txid>,
+    empty_anchors: BTreeSet<A>,
```
notmandatory, line 155:
```suggestion
    // This exists so that `TxGraph` functions can return a reference to an empty set of `TxNode.anchors`.
    empty_anchors: BTreeSet<A>,
```

crates/wallet/src/wallet/mod.rs
```diff
@@ -1050,10 +1050,17 @@ impl Wallet {
...
+    ///     ChainPosition::ConfirmedByTransitivity(txid, anchor) => println!(
+    ///         "tx is confirmed because a descendant '{}' is in the best chain at {}:{}",
+    ///         txid, anchor.block_id.height, anchor.block_id.hash,
+    ///     ),
     ///     ChainPosition::Unconfirmed(last_seen) => println!(
     ///         "tx is last seen at {}, it is unconfirmed as it is not anchored in the best chain",
     ///         last_seen,
     ///     ),
+    ///     ChainPosition::UnconfirmedAndNotSeen => println!(
+    ///         "tx does not conflict with anything canonical, but we never seen it in mempool",
```
notmandatory: ⛏️ grammar nit
```suggestion
    ///         "tx does not conflict with anything canonical, but we have never seen it in the mempool",
```

notmandatory:
Per call today @evanlinjin will open a new PR for the 1.0.0-beta milestone that only makes expected breaking changes for `Wallet`,  likely only the `ChainPosition` enum. Then the algorithm improvements in this PR can be released in a future non-breaking patch version.

notmandatory:
On further discussion today at release planning call this PR was moved back into the 1.0 milestone if it can be completed and reviewed in time. If not it will have to wait for a 2.0 milestone because it required breaking changes to chain crate APIs that are exposed in the `Wallet` API.  

Alternatively we could remove the following `Wallet` functions and move any error types from chain into the core module which should insulate the wallet crate from this chain crate API changes. 

```rust
    /// Get a reference to the inner [`TxGraph`].
    pub fn tx_graph(&self) -> &TxGraph<ConfirmationBlockTime> {
        self.indexed_graph.graph()
    }

    /// Get a reference to the inner [`KeychainTxOutIndex`].
    pub fn spk_index(&self) -> &KeychainTxOutIndex<KeychainKind> {
        &self.indexed_graph.index
    }

    /// Get a reference to the inner [`LocalChain`].
    pub fn local_chain(&self) -> &LocalChain {
        &self.chain
    }
```

evanlinjin:
@notmandatory how does this solution compare with just giving wallet a major version bump?

notmandatory:
Removing the above functions and moving required chain error type to core should allow us to do breaking chain crate api changes without having to do a major wallet crate release. I also don't see why Wallet users need to access the inner chain types directly instead of using higher level functions like `transactions()`, or we can add new helper functions without breaking any APIs. 

But all that said if it's less risky to just keep everything as is and do a 2.0 release in 6 mo or so I'd be fine with that too.

crates/chain/src/tx_graph.rs
evanlinjin: ~This is out of scope for this PR. I'll turn this into a ticket.~

I realized that this is not out of scope, since we need to explain the purpose of `empty_anchors` below and might as well do this together.

crates/wallet/src/test_utils.rs
```diff
@@ -224,37 +224,63 @@ pub fn feerate_unchecked(sat_vb: f64) -> FeeRate {
...
+    // } else {
+    //     ChainPosition::Confirmed {
+    //         anchor: ConfirmationBlockTime {
+    //             block_id: latest_cp.block_id(),
+    //             confirmation_time: 0,
+    //         },
+    //         transitively: None,
+    //     }
+    // };
+    // receive_output(wallet, value, anchor)
```
evanlinjin: woops.

evanlinjin:
All that remains is docs and updating the PR description.

README.md
```diff
@@ -78,6 +78,10 @@ cargo update -p tokio --precise "1.38.1"
 cargo update -p tokio-util --precise "0.7.11"
 cargo update -p indexmap --precise "2.5.0"
 cargo update -p security-framework-sys --precise "2.11.1"
+
+# Additional pins for running tests
+cargo update -p csv --precise "1.3.0"
+cargo update -p unicode-width --precise "0.1.13"
```
evanlinjin: This is unfortunate. I think we should do #1398.

ValuedMammal:
Concept ACK, I'm still reviewing

Anecdotally I can attest that computation scales linearly with `TxGraph` size.

README.md
```diff
@@ -78,6 +78,10 @@ cargo update -p tokio --precise "1.38.1"
 cargo update -p tokio-util --precise "0.7.11"
 cargo update -p indexmap --precise "2.5.0"
 cargo update -p security-framework-sys --precise "2.11.1"
+
+# Additional pins for running tests
```
LLFourn: aren't all these pins for running tests...? I mean if you are just building a library then you likely won't have many of these dependencies.

crates/chain/src/tx_graph.rs
```diff
@@ -134,16 +140,19 @@ impl<A: Ord + Clone> From<TxUpdate<A>> for TxGraph<A> {
...
+pub struct TxGraph<A = ConfirmationBlockTime> {
+    txs: HashMap<Txid, TxNodeInternal>,
     spends: BTreeMap<OutPoint, HashSet<Txid>>,
-    anchors: BTreeSet<(A, Txid)>,
+    anchors: HashMap<Txid, BTreeSet<A>>,
     last_seen: HashMap<Txid, u64>,
 
-    // This atrocity exists so that `TxGraph::outspends()` can return a reference.
-    // FIXME: This can be removed once `HashSet::new` is a const fn.
+    txs_by_anchor: BTreeSet<(u32, Txid)>,
```
LLFourn: :question: 
```suggestion
    highest_confirmation_heights: BTreeSet<(u32, Txid)>,
```

crates/chain/src/canonical_iter.rs
```diff
@@ -0,0 +1,319 @@
...
+            pending_remaining: VecDeque::new(),
+            canonical: HashMap::new(),
+            not_canonical: HashSet::new(),
+            queue: VecDeque::new(),
+        }
+    }
+
+    /// Whether this transaction is already visited.
+    fn is_visited(&self, txid: Txid) -> bool {
+        self.canonical.contains_key(&txid) || self.not_canonical.contains(&txid)
```
LLFourn, line 56: `is_visited`  is a bit of a lie. `scan_anchors` is visiting the tx but it doesn't always make it `is_visited`. This should be `is_canonicalized`.

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -591,10 +591,84 @@ fn test_tx_conflict_handling() {
...
+                },
+            ],
+            exp_chain_txs: HashSet::from(["root", "transitively_anchored_conflict", "anchored"]),
+            exp_chain_txouts: HashSet::from([("root", 0), ("transitively_anchored_conflict", 0), ("anchored", 0)]),
+            exp_unspents: HashSet::from([("anchored", 0)]),
+            exp_balance: Balance {
+                confirmed: Amount::from_sat(8000),
+                ..Default::default()
+            }
+        }
```
LLFourn, line 689: There seems to be a test missing for an anchored tx but not confirmed tx being canonical (or not). The orphaned block tests we have all are about conflicts.

P.S. this file is test_tx_graph_conflicts.rs but maybe it should just be `test_tx_graph_canonicalization.rs` .

LLFourn (approved):
ACK d107e4be516b1282cd5301322abbb15257934a40

Really excellent work. Thanks so much for doing an in-depth ADR. Made minor comments and a request for a test.

jirijakes:
I realize it's a lot of code but would it be possible to turn commits https://github.com/bitcoindevkit/bdk/pull/1670/commits/759be4c8f55adb266559fbdec90e1dc5ff96f9e6 and https://github.com/bitcoindevkit/bdk/pull/1670/commits/3577c5f25f3725f706627cf1786a24860cce63aa into one? The former commit introduces some changes (e. g. creates `canonical_iter.rs`) but the latter one changes them, which makes it more difficult to review commit by commit.

thunderbiscuit:
Thanks for the fantastic ADR @evanlinjin!

Perfect for me to stay up to speed on the changes without requiring a full code review, which I would not have had time to do! 🔥🔥

crates/chain/benches/canonicalization.rs
nymius, line 1: Shouldn't this be separated in another PR to review (and merge) first in master? Then benchmark test could be run against master before this PR get merge and have a reproducible way to check the performance increment after it does.

crates/chain/src/lib.rs
```diff
@@ -43,6 +43,8 @@ pub mod tx_graph;
 pub use tx_graph::TxGraph;
 mod chain_oracle;
 pub use chain_oracle::*;
+mod canonical_iter;
+pub use canonical_iter::*;
```
ValuedMammal, line 47: I would have opted for less public visibility for most of the items in here.

crates/chain/src/canonical_iter.rs
```diff
@@ -0,0 +1,319 @@
...
+
+            return None;
+        }
+    }
+}
+
+/// Represents when and where a given transaction is last seen.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, core::hash::Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+pub enum LastSeenIn {
```
ValuedMammal: Name suggestion to avoid overloading the term last-seen. Also why does it implement serde traits?

```suggestion
pub enum ObservedIn {
```

crates/chain/src/canonical_iter.rs
```diff
@@ -0,0 +1,319 @@
...
+            None => ChainPosition::Confirmed {
+                anchor,
+                transitively: None,
+            },
+        },
+        CanonicalReason::LastSeen { last_seen, .. } => match last_seen {
+            LastSeenIn::Mempool(last_seen) => ChainPosition::Unconfirmed {
+                last_seen: Some(last_seen),
+            },
+            LastSeenIn::Block(_) => ChainPosition::Unconfirmed { last_seen: None },
```
ValuedMammal: I would have expected this to be not canon. based on my reading of the definition of `CanonicalTx`.

docs/adr/0003_canonicalization_algorithm.md
```diff
@@ -0,0 +1,83 @@
...
+
+* Status: Proposed
+* Authors: @LLFourn, @evanlinjin
+* Date: 2024-12-01
+* Targeted modules: `bdk_chain`
+* Associated Tickets/PRs: Issue #1665, ~PR #1659~, PR #1670
+
+## Context and Problem Statement
+
+The Wizardsardine BDK audit uncovered the severity of the performance issues in the original canonicalization logic. The problem is especially severe for wallet histories with many unconfirmed and conflicting transactions. This can be a dDos vector if BDK is used in server-side applications. The time complexity of the original canonicalization logic is `O(n^2)`.
```
ValuedMammal: Consider using math notation, e.g. $O(n^2)$

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/writing-mathematical-expressions

ValuedMammal (approved):
ACK d107e4be516b1282cd5301322abbb15257934a40

📌 Comments are mostly for discussion and no action is required unless you plan to retouch.

docs/adr/0003_canonicalization_algorithm.md
notmandatory: ⛏️ Should include a link to the audit report.
```suggestion
The [2024 Wizardsardine BDK code audit](https://gist.github.com/darosior/4aeb9512d7f1ac7666abc317d6f9453b) uncovered the severity of the performance issues in the original canonicalization logic. The problem is especially severe for wallet histories with many unconfirmed and conflicting transactions. This can be a dDos vector if BDK is used in server-side applications. The time complexity of the original canonicalization logic is `O(n^2)`.
```

crates/chain/benches/canonicalization.rs
```diff
@@ -0,0 +1,263 @@
...
+        let (tx_graph, chain) = (tx_graph.clone(), chain.clone());
+        move |b| b.iter(|| run_filter_chain_txouts(&tx_graph, &chain, 1))
+    });
+    c.bench_function("many_chained_unconfirmed::filter_chain_unspents", {
+        let (tx_graph, chain) = (tx_graph.clone(), chain.clone());
+        move |b| b.iter(|| run_filter_chain_unspents(&tx_graph, &chain, 0))
+    });
+}
+
+pub fn nested_conflicts(c: &mut Criterion) {
```
nymius, line 195: I'm having problems grokking the idea of "nested" conflicts. The code below creates the following transaction dependency tree, right?
```text
	    A                  B
	   /|\                /|\ 
	  / | \              / | \
	 /  |  \            /  |  \
	/   |   \          /   |   \
       /    |    \        /    |    \
      /     |     \      /     |     \
     /      |      \    /      |      \
    tx1...tx2099   tx1050     tx2...tx2100
```
Then the "answers" to the assert would be:
`nested_conflicts_unconfirmed::list_canonical_txs` -->  `A,B,tx1,tx2`
`nested_conflicts_unconfirmed::filter_chain_txouts` --> `TxOut(A), TxOut(B)`
`nested_conflicts_unconfirmed::filter_chain_unspents` --> `0 as TxOut(A) is spent by (unconfirmed) tx1 and TxOut(B) is spent by (unconfirmed) tx2` 

So, why "nested"?

docs/adr/0003_canonicalization_algorithm.md
```diff
@@ -0,0 +1,83 @@
...
+
+1. Iterate over all transactions with anchors in descending anchor-height order. For any transaction that has an anchor pointing to the best chain, we call `mark_canonical` on it. We iterate in descending-height order to reduce the number of anchors we need to check against the `ChainOracle` (premise 1). The purpose of this run is to populate `non_canonical` with all transactions that directly conflict with anchored transactions and populate `canonical` with all anchored transactions and ancestors of anchors transactions (transitive anchors).
+2. Iterate over all transactions with last-seen values, in descending last-seen order. We can call `mark_canonical` on all of these that do not already exist in `canonical` or `not_canonical`.
+3. Iterate over remaining transactions that contains anchors (but not in the best chain) and have no last-seen value. We treat these transactions in the same way as we do in run 2.
+
+**Pros:**
+* Transitive anchors are handled correctly.
+* We can efficiently mark large subsets as canonical/non-canonical.
+
+**Cons:** ?
```
notmandatory, line 79:
```suggestion
**Cons:** More API and code changes.
```

crates/chain/src/lib.rs
evanlinjin: We have an iterator `CanonicalIter` which should be public since it's returned by `TxGraph::canonical_iter`. `<CanonicalIter as Iterator>::Item` contains `CanonicalReason` so that needs to be public. Subsequently, `LastSeenIn` also needs to be public.

Otherwise, I've tried to reduce the publicly exposed items in the latest push.

crates/chain/src/canonical_iter.rs
evanlinjin: The docs of `CanonicalTx` needs to be changed. If a tx is not seen in the mempool, but we've seen it in a conflicting block, we know that the tx is seen by the network so we should still consider it canonical.

crates/chain/src/canonical_iter.rs
evanlinjin: Changed in the latest push.

docs/adr/0003_canonicalization_algorithm.md
evanlinjin: I didn't know this was a thing. Thanks!

crates/chain/src/tx_graph.rs
evanlinjin: Renamed to `txs_by_highest_conf_height`.

README.md
evanlinjin: Good point.

crates/chain/src/canonical_iter.rs
evanlinjin: Updated in the latest push.

crates/chain/src/canonical_iter.rs
evanlinjin: Changed.

docs/adr/0003_canonicalization_algorithm.md
evanlinjin: I don't think the API surface changes are substantial - it's still mostly the same. Definitely more code changes though. However, I don't think this is a con in terms of architecture? Just the implementer and reviewers need to do more work.

nymius (commented):
cACK

These are my results for running the benchmark on this PR:
```text
many_conflicting_unconfirmed::list_canonical_txs [529.28 us 533.25 us 537.34 us]
many_conflicting_unconfirmed::filter_chain_txouts [560.02 us 566.69 us 574.36 us]
many_conflicting_unconfirmed::filter_chain_unspents [570.67 us 576.12 us 581.60 us]
many_chained_unconfirmed::list_canonical_txs [1.8421 ms 1.8820 ms 1.9277 ms]
many_chained_unconfirmed::filter_chain_txouts [2.5919 ms 2.6256 ms 2.6601 ms]
many_chained_unconfirmed::filter_chain_unspents [2.7202 ms 2.7679 ms 2.8201 ms]
nested_conflicts_unconfirmed::list_canonical_txs [505.92 us 509.53 us 513.44 us]
nested_conflicts_unconfirmed::filter_chain_txouts [502.93 us 511.17 us 519.27 us]
nested_conflicts_unconfirmed::filter_chain_unspents [485.62 us 488.70 us 492.61 us]
```

### Nits:
1. `crates/chain/tests/test_indexed_tx_graph.rs` lines 529, 536 and 566 still exhibit the old `get_chain_position` name.
2. The `CHANGELOG NOTICE` mentions it removes the `try_get_chain_position` twice. Should it be `try_get_chain_position` first and then `try_get_chain_spend` second?
3. I stumbled with the following _warning_ while running the bench code:
```bash
warning: the following packages contain code that will be rejected by a future version of Rust: criterion v0.2.11
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
```

crates/chain/benches/canonicalization.rs
evanlinjin: I've cherry-picked the benchmark commits into a separate branch that is based on master here: https://github.com/evanlinjin/bdk/tree/fix/1665-master-bench.

The problem is that the benchmarks on master takes so long that I don't think there is a point to check the performance increment against master.

crates/chain/benches/canonicalization.rs
evanlinjin: Good catch, this benchmark is implemented wrong.

evanlinjin: Fixed the test in the latest push.

README.md
evanlinjin: Removed this comment in the latest push.

crates/chain/tests/test_tx_graph_conflicts.rs
evanlinjin: In terms of naming, I think conflicts and canonicalization are mostly synonymous. We don't need to canonicalize if we don't have conflicts.

evanlinjin: Test added.

evanlinjin:
```
warning: the following packages contain code that will be rejected by a future version of Rust: criterion v0.2.11
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
```

@nymius Yes, this warning is annoying. Right now our MSRV policy is enforced in the tests (which I don't think is a good idea) so we are forced to use a very old version of `criterion`. I think we should do bitcoindevkit/bdk_wallet#101 and update our dev dependencies.

crates/wallet/src/wallet/mod.rs
```diff
@@ -2014,20 +2026,22 @@ impl Wallet {
...
-                    &self.chain,
-                    chain_tip,
-                    txid,
-                ) {
-                    Some(chain_position) => chain_position.cloned(),
+                let (chain_position, tx) = match canon_txs.get(&txid) {
+                    Some(CanonicalTx {
+                        chain_position,
+                        tx_node,
+                    }) => (chain_position, tx_node.tx.clone()),
```
jirijakes: In 5456f8b4 feat(chain)!: rm `get_chain_position` and associated methods

```suggestion
                    }) => (chain_position, &tx_node.tx),
```

crates/wallet/src/wallet/mod.rs
```diff
@@ -2014,20 +2026,22 @@ impl Wallet {
             return (must_spend, vec![]);
         }
 
+        let canon_txs = self
+            .indexed_graph
+            .graph()
+            .list_canonical_txs(&self.chain, chain_tip)
+            .map(|canon_tx| (canon_tx.tx_node.txid, canon_tx))
+            .collect::<HashMap<Txid, _>>();
+
```
jirijakes: In https://github.com/bitcoindevkit/bdk/commit/5456f8b4cd3fabaf1215d2bdec06bd0eae67a86f feat(chain)!: rm get_chain_position and associated methods:

Is it necessary to build list of canonical transactions here?

If it's just to have `chain_position` available, then that should not be necessary, chain position is already contained in `u.0` (`LocalOutput`) and can be extracted from there (see #1758).

If I remove `canon_txs` and apply the same change as in #1758, tests still pass.

```rs
        //    must_spend <- manually selected utxos
        //    may_spend  <- all other available utxos
        let mut may_spend = self.get_available_utxos();

        may_spend.retain(|may_spend| {
            !manually_selected
                .iter()
                .any(|manually_selected| manually_selected.utxo.outpoint() == may_spend.0.outpoint)
        });
        let mut must_spend = manually_selected;

        // NOTE: we are intentionally ignoring `unspendable` here. i.e manual
        // selection overrides unspendable.
        if *manually_selected_only {
            return (must_spend, vec![]);
        }

        let satisfies_confirmed = may_spend
            .iter()
            .map(|u| -> bool {
                let txid = u.0.outpoint.txid;
                let tx = match self.indexed_graph.graph().get_tx(txid) {
                    Some(tx) => tx,
                    None => return false,
                };

                // Whether the UTXO is mature and, if needed, confirmed
                let mut spendable = true;
                let chain_position = u.0.chain_position;

```

crates/wallet/src/wallet/mod.rs
```diff
@@ -1833,9 +1832,28 @@ impl Wallet {
...
+            .input
+            .iter()
+            .map(|txin| txin.previous_output.txid)
+            .collect::<HashSet<Txid>>();
+        let confirmation_heights = self
+            .indexed_graph
+            .graph()
+            .list_canonical_txs(&self.chain, chain_tip)
+            .filter(|canon_tx| prev_txids.contains(&canon_tx.tx_node.txid))
+            .take(prev_txids.len())
```
jirijakes, line 1850: In 5456f8b4 feat(chain)!: rm `get_chain_position` and associated methods:

`.take(prev_txids.len())` should not be needed. After filtering there can't be more than `prev_txids.len()` transactions. And then `filter_map` could be used.

If I make both `prev_txids` and `confirmation_heights` empty, no test in `wallet` fails. It seems like some execution paths are not covered by tests, so it is hard to tell whether logic of the implementation changed or not.

crates/chain/tests/test_indexed_tx_graph.rs
jirijakes, line 1: In this file, there are still some references to `get_chain_position` which does not exist anymore.

jirijakes (commented):
Thanks for rerebasing and resquashing :pray:

A few comments from first round of reviewing. I have not yet reviewed the algorithm itself and ADR, I intend to do so during Saturday.

crates/chain/src/canonical_iter.rs
```diff
@@ -0,0 +1,251 @@
...
+use bdk_core::BlockId;
+use bitcoin::{Transaction, Txid};
+
+/// Iterates over canonical txs.
+pub struct CanonicalIter<'g, A, C> {
+    tx_graph: &'g TxGraph<A>,
+    chain: &'g C,
+    chain_tip: BlockId,
+
+    pending_anchored: Box<dyn Iterator<Item = (Txid, Arc<Transaction>, &'g BTreeSet<A>)> + 'g>,
```
nymius: `pending` in bitcoin context may be ambiguous. From what I understand, `pending` here refers to those transactions that haven't been canonicalized yet. Could the struct be documented considering this?

nymius (commented):
Great work!

Final NIT. From PR branch:
```bash
rg -l "canoncical|canoncial|canonial" | xargs sed -i "s/\(canoncical\|canoncial\|canonial\)/canonical/g"
```

crates/chain/src/canonical_iter.rs
```diff
@@ -0,0 +1,251 @@
...
+                canonical_entry.insert((tx, this_reason));
+                self.queue.push_back(this_txid);
+                Some(())
+            },
+        )
+        .for_each(|_| {})
+    }
+}
+
+impl<'g, A: Anchor, C: ChainOracle> Iterator for CanonicalIter<'g, A, C> {
```
jirijakes:
```suggestion
impl<A: Anchor, C: ChainOracle> Iterator for CanonicalIter<'_, A, C> {
```

crates/chain/src/canonical_iter.rs
```diff
@@ -0,0 +1,251 @@
...
+        txid: Txid,
+        tx: Arc<Transaction>,
+        anchors: &BTreeSet<A>,
+    ) -> Result<(), C::Error> {
+        for anchor in anchors {
+            let in_chain_opt = self
+                .chain
+                .is_block_in_chain(anchor.anchor_block(), self.chain_tip)?;
+            if in_chain_opt == Some(true) {
+                self.mark_canonical(tx, CanonicalReason::from_anchor(anchor.clone()));
```
jirijakes: :wrench: I believe cloning should not be necessary here. However, without further refactoring it might be needed on a higher level. It could be tried in a follow-up.

crates/chain/src/canonical_iter.rs
```diff
@@ -0,0 +1,251 @@
...
+                .unwrap()
+                .confirmation_height_upper_bound(),
+        ));
+        Ok(())
+    }
+
+    /// Marks a transaction and it's ancestors as canoncial. Mark all conflicts of these as
+    /// `not_canonical`.
+    fn mark_canonical(&mut self, tx: Arc<Transaction>, reason: CanonicalReason<A>) {
+        let starting_txid = tx.compute_txid();
```
jirijakes: Txid is already available everytime `mark_canonical` is called. We could avoid recomputing it again.

crates/chain/src/canonical_iter.rs
```diff
@@ -0,0 +1,251 @@
...
+                        },
+                    )
+                    .run_until_finished()
+                }
+                canonical_entry.insert((tx, this_reason));
+                self.queue.push_back(this_txid);
+                Some(())
+            },
+        )
+        .for_each(|_| {})
```
jirijakes:
```suggestion
        .run_until_finished()
```

crates/chain/src/canonical_iter.rs
jirijakes: :+1: 

Or simply `anchored_iter` and `last_seen_iter`.

crates/chain/src/canonical_iter.rs
```diff
@@ -0,0 +1,251 @@
...
+                .last()
+                .unwrap()
+                .confirmation_height_upper_bound(),
+        ));
+        Ok(())
+    }
+
+    /// Marks a transaction and it's ancestors as canoncial. Mark all conflicts of these as
+    /// `not_canonical`.
+    fn mark_canonical(&mut self, tx: Arc<Transaction>, reason: CanonicalReason<A>) {
```
jirijakes: If this method is called `canonicalize`, would it make sense? Is the meaning of `is_canonicalized` the same as “gone through `mark_canonical`”?

I am not necessarily suggesting to rename it, I am trying to understand.

crates/chain/src/canonical_iter.rs
```diff
@@ -0,0 +1,251 @@
...
+            }
+        }
+        // cannot determine
+        self.pending_remaining.push_back((
+            txid,
+            tx,
+            anchors
+                .iter()
+                .last()
+                .unwrap()
```
jirijakes: Maybe add `debug_assert` to check that `anchors` is not empty?

crates/chain/src/tx_graph.rs
```diff
@@ -624,14 +634,36 @@ impl<A: Clone + Ord> TxGraph<A> {
...
+            hash_map::Entry::Occupied(mut e) => {
+                fn top_anchor_height<A: Anchor>(anchors: &BTreeSet<A>) -> Option<u32> {
+                    anchors
+                        .iter()
+                        .last()
+                        .map(Anchor::confirmation_height_upper_bound)
+                }
+                old_top_h = top_anchor_height(e.get());
+                let is_changed = e.get_mut().insert(anchor.clone());
+                new_top_h = top_anchor_height(e.get()).expect("must have anchor");
```
jirijakes: 🔧 It is a bit hard to understand that what this does is to determine whether a newly inserted anchor has higher height than any of previously existing anchors. It is especially confusing that later the two values are compared as `old_top_h != Some(new_top_h)`.

A comment may help. Or perhaps to find `new_top_h`, compare `old_top_h` with `anchor` directly instead of finding BTreeSet's max value again: either just-inserted `anchor` will be higher or `old_top_h`.

jirijakes (commented):
Reviewed the new algorithm and ADR. Left a few comments but nothing extremely important.

Great work!

crates/chain/src/canonical_iter.rs
evanlinjin: How about `unprocessed_anchored_txs`?

crates/chain/src/canonical_iter.rs
evanlinjin: This whole algorithm is a canonicalization algorithm. So I'm that sense, we are canonicalizing everything and the name `canonicalize` will be ambiguous. 

We are explicitly meeting a tx as canonical here. We make use of the transitive nature of a canonical tx to mark more things. 

`is_canonicalized` returns whether we have an answer for a given tx. If so, we can skip it.

notmandatory (approved):
ACK 664bb9ae97a5b5e347c5b2bb9c989f0bfb5fa8d3

Great job with the ADR, it's a big help for understand the logic. The benchmarks are also a first for the project and do a fantastic job quantifying the value of this change.

crates/chain/src/tx_graph.rs
```diff
@@ -138,12 +142,15 @@ impl<A: Ord + Clone> From<TxUpdate<A>> for TxGraph<A> {
...
 #[derive(Clone, Debug, PartialEq)]
-pub struct TxGraph<A = ()> {
+pub struct TxGraph<A = ConfirmationBlockTime> {
     txs: HashMap<Txid, TxNodeInternal>,
     spends: BTreeMap<OutPoint, HashSet<Txid>>,
     anchors: HashMap<Txid, BTreeSet<A>>,
     last_seen: HashMap<Txid, u64>,
 
+    txs_by_highest_conf_heights: BTreeSet<(u32, Txid)>,
+    txs_by_last_seen: BTreeSet<(u64, Txid)>,
```
oleonardolima, line 150: nit: we could probably use `txids_by_...` instead, wdyt ?

crates/chain/src/canonical_iter.rs
nymius: `unprocessed` seems right

crates/chain/tests/test_indexed_tx_graph.rs
oleonardolima: Yes, some are still at L529 (the test name), L536, L566.

notmandatory:
Might need a rebase.

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -595,6 +597,7 @@ fn test_tx_conflict_handling() {
 
     for scenario in scenarios {
         let (tx_graph, spk_index, exp_tx_ids) = init_graph(scenario.tx_templates.iter());
+        let exp_tx_ids = dbg!(exp_tx_ids);
```
oleonardolima: nit: Did you mean to keep this `dbg!()` ?

crates/chain/src/canonical_iter.rs
```diff
@@ -214,36 +213,39 @@ impl<A: Clone> CanonicalReason<A> {
     }
 
     /// Constructs a [`CanonicalReason`] from a `last_seen` value.
```
oleonardolima: nit: I guess the docs need an update too.

``` suggestion
    /// Constructs a [`CanonicalReason`] from a `observed_in` value.
```

crates/chain/src/canonical_iter.rs
```diff
@@ -214,36 +213,39 @@ impl<A: Clone> CanonicalReason<A> {
...
                 anchor: anchor.clone(),
                 descendant: Some(descendant),
             },
-            CanonicalReason::LastSeen { last_seen, .. } => Self::LastSeen {
-                last_seen: *last_seen,
+            CanonicalReason::ObservedIn {
+                observed_in: last_seen,
+                ..
+            } => Self::ObservedIn {
+                observed_in: *last_seen,
```
oleonardolima: Is it intentional to keep using `last_seen` here instead of all `observed_in`?

oleonardolima (commented):
Concept ACK

Great work! I don't have much experience with the `bdk_chain`, so I'll need another round to wrap my mind around the changes, but the docs and ADR already helped a lot 😁.

I left a few comments with some possibly small applicable nits.

crates/chain/src/canonical_iter.rs
oleonardolima: Yes, +1 on this.

crates/chain/src/canonical_iter.rs
evanlinjin: Are you thinking about doing this with references or by making `Anchor` be bound by `Copy`?

crates/chain/src/canonical_iter.rs
evanlinjin: What are we trying to achieve with this? Would changing `.unwrap()` to `.expect()` suffice?

evanlinjin: Oh, at the top of the method?

evanlinjin: I guess the problem will still get picked up by the `.expect()`?

crates/chain/src/tx_graph.rs
evanlinjin: I don't think this will make a significant difference to the readability of the codebase.

crates/chain/src/tx_graph.rs
evanlinjin: Good catch. I tried reorganizing this logic in 87528963805da281c2f674eba63e22a075613117. Let me know what you think.

crates/wallet/src/wallet/mod.rs
evanlinjin: It's for a small performance gain. We may not return anything since `filter` filters it out. However, the `CanonicalIter` will continue to try canonicalize. Will add a comment.

> If I make both prev_txids and confirmation_heights empty, no test in wallet fails. It seems like some execution paths are not covered by tests, so it is hard to tell whether logic of the implementation changed or not.

Lets create a ticket since this is out of scope for this PR. However, currently BDK does not handle median-time-past properly.

evanlinjin: I decided to include a comment on #642 instead.

crates/wallet/src/wallet/mod.rs
evanlinjin: Good catch, included your commit.

crates/wallet/src/wallet/mod.rs
evanlinjin: This line no longer exists.

crates/chain/src/canonical_iter.rs
evanlinjin: Fixed

crates/chain/tests/test_indexed_tx_graph.rs
evanlinjin: Should be somewhat resolved by latest push.

ValuedMammal (approved):
ACK 956d0a93bab36fd919e81718832fbc1033b677e9

nymius:
ACK 956d0a93

nymius (approved):
ACK 956d0a93

oleonardolima (commented):
utACK 956d0a93bab36fd919e81718832fbc1033b677e9

crates/chain/src/canonical_iter.rs
jirijakes: I meant reference. But of course, unnecessary in this PR. I may give it a try after it's merged.

crates/chain/src/canonical_iter.rs
jirijakes: Yes, at the top. True, in this specific case, the expect would work similarly and adding `debug_assert` now is not worth it.

(I came to believe that in general, putting `debug_assert` around the code is a [good practice](https://tratt.net/laurie/blog/2023/rusts_two_kinds_of_assert_make_for_better_code.html) for documentation purposes and for increasing a chance of discovering bugs.)

jirijakes (approved):
ACK 956d0a93bab36fd919e81718832fbc1033b677e9

:fire:

crates/chain/src/canonical_iter.rs
evanlinjin: That's a great read and thanks for sharing.

notmandatory:
Thanks @evanlinjin and @ValuedMammal for all the work on getting this coded, documented, and benchmark tested, and to everyone who helped with review!

PR #1675, opened 2024-11-06
fix(electrum): `bdk_electrum` coinbase merkle and `populate_with_txids` exploit fixes

### LINKED ISSUE #1698: bdk_electrum: don't query prevouts of coinbase tx
"fetch_prev_txout should not try to query the prevouts of coinbase transactions, this will query a 000...000 transaction to the Electrum server which will return an error and will make the overall sync() fail."

LagginTimes:
Fixes bitcoindevkit/bdk#1698, bitcoindevkit/bdk#2249.

### Description

The aim of this PR is to fix a few exploits that were found within `bdk_electrum`. Most notably:

- Transaction Merkle proof verification does not check the Merkle proof for the coinbase. This enables an attack by which an attacker can fake a deeply-confirmed payment to a BDK wallet. (https://delvingbitcoin.org/t/great-consensus-cleanup-revival/710#merkle-tree-attacks-using-64-bytes-transactions-8)
- An Electrum server can trigger a panic in `populate_with_txids` by returning a transaction with no output.
- `fetch_prev_txout` should not try to query the prevouts of coinbase transactions. This will query a transaction to the Electrum server which will return an error and will make the overall `sync` fail.

### Notes to the reviewers

- The "validate coinbase merkle test" commit needs significant review. My initial thoughts were to put `MockElectrumClient` into `bdk_testenv`, but because `bdk_testenv`'s `electrum-client` is re-exported from `electrsd`, a separate dependency would have to be added just for `electrum-client` 0.22 support. However, since we are doing in-line tests to access private functions, there is a strong argument for leaving the test structures where they are currently at. Or perhaps there is a better way to specifically test `validate_merkle_for_anchor` that does not involve creating a `MockElectrumClient`, similar to what @evanlinjin is doing here: https://github.com/evanlinjin/experiments/blob/main/electrum_c/src/state.rs.

### Changelog notice

* `fetch_prev_txout` no longer queries coinbase transactions.
* Coinbase tx merkle path is also validated during tx merkle path verification.
* `populate_with_txids` will no longer panic on transactions with no outputs.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -346,24 +346,22 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
                 Err(other_err) => return Err(other_err),
             };
 
-            let spk = tx
-                .output
-                .first()
-                .map(|txo| &txo.script_pubkey)
-                .expect("tx must have an output");
+            if let Some(txo) = tx.output.first() {
+                let spk = &txo.script_pubkey;
 
-            // because of restrictions of the Electrum API, we have to use the `script_get_history`
-            // call to get confirmation status of our transaction
-            if let Some(r) = self
-                .inner
-                .script_get_history(spk)?
-                .into_iter()
-                .find(|r| r.tx_hash == txid)
-            {
-                self.validate_merkle_for_anchor(tx_update, txid, r.height)?;
+                // because of restrictions of the Electrum API, we have to use the `script_get_history`
+                // call to get confirmation status of our transaction
+                if let Some(r) = self
+                    .inner
+                    .script_get_history(spk)?
+                    .into_iter()
+                    .find(|r| r.tx_hash == txid)
+                {
+                    self.validate_merkle_for_anchor(tx_update, txid, r.height)?;
+                }
+
+                tx_update.txs.push(tx);
             }
-
-            tx_update.txs.push(tx);
         }
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


oleonardolima:
Isn't this one part of 1.0 milestone? I guess it's missing it and not appearing on the project board.

notmandatory:
For the audit issue bitcoindevkit/bdk#1698 you fixed it with bitcoindevkit/bdk#1756.  Can you update this to just be checking Merkle proofs? And if so then I'd say this is a new feature that can be put off to a post 1.0.0 release.

oleonardolima (commented):
Concept ACK

I'm planning giving this one a test throughout the weekend, and think more about the approach with `MockElectrumClient`.

> For the audit issue https://github.com/bitcoindevkit/bdk/issues/1698 you fixed it with https://github.com/bitcoindevkit/bdk/pull/1756. Can you update this to just be checking Merkle proofs? And if so then I'd say this is a new feature that can be put off to a post 1.0.0 release.

Agree, the commit with the fix on coinbase prevout should be dropped. The remaining two fixes (I guess) can either be together in this PR or separate, it'd depend on the priority of each one on `1.0.0-beta` milestone.

evanlinjin:
Thank you for working on this. However, I'm not sure how useful this is here, since we aren't even verifying the PoW of the chain in `bdk_electrum`. `bdk_electrum` assumes that we can trust the server that they are connected to.

In the future, we should have a version of `bdk_electrum` which checks PoW and addresses this exploit.

evanlinjin:
Closing as it's stale.

PR #1682, opened 2024-11-13
test(wallet): create tx locktime cltv for a specific time

f3r10:
### Description
This PR adds a unit test that checks the satisfaction of timestamp-based timelocks. The goal is to test the absolute time therefore the variable passed to the miniscript fragment `after` has to be equal to or greater that 500_000_000 otherwise it would be checking the block height.

### Notes to the reviewers

This unit test tries to check if bitcoindevkit/bdk_wallet#183 is still an issue.

#### All Submissions:

### DIFF
--- a/crates/wallet/src/test_utils.rs
+++ b/crates/wallet/src/test_utils.rs
@@ -160,6 +160,13 @@ pub fn get_test_single_sig_cltv() -> &'static str {
     "wsh(and_v(v:pk(cVpPVruEDdmutPzisEsYvtST1usBR3ntr8pXSyt6D2YYqXRyPcFW),after(100000)))"
 }
 
+/// `wsh` descriptor with policy `and(pk(A),after(1_734_230_218))`
+// the parameter passed to miniscript fragment `after` has to equal or greater than 500_000_000
+// in order to use a lock based on unix time
+pub fn get_test_single_sig_cltv_timestamp() -> &'static str {
+    "wsh(and_v(v:pk(cVpPVruEDdmutPzisEsYvtST1usBR3ntr8pXSyt6D2YYqXRyPcFW),after(1734230218)))"
+}
+
 /// taproot single key descriptor
 pub fn get_test_tr_single_sig() -> &'static str {
     "tr(cNJmN3fH9DDbDt131fQNkVakkpzawJBSeybCUNmP1BovpmGQ45xG)"
--- a/crates/wallet/tests/wallet.rs
+++ b/crates/wallet/tests/wallet.rs
@@ -610,6 +610,21 @@ fn test_create_tx_default_locktime_cltv() {
 }
 
 #[test]
+fn test_create_tx_locktime_cltv_timestamp() {
+    let (mut wallet, _) = get_funded_wallet_single(get_test_single_sig_cltv_timestamp());
+    let addr = wallet.next_unused_address(KeychainKind::External);
+    let mut builder = wallet.build_tx();
+    builder.add_recipient(addr.script_pubkey(), Amount::from_sat(25_000));
+    let mut psbt = builder.finish().unwrap();
+
+    assert_eq!(psbt.unsigned_tx.lock_time.to_consensus_u32(), 1_734_230_218);
+
+    let finalized = wallet.sign(&mut psbt, SignOptions::default()).unwrap();
+
+    assert!(finalized);
+}
+
+#[test]
 fn test_create_tx_custom_locktime() {
     let (mut wallet, _) = get_funded_wallet_wpkh();
     let addr = wallet.next_unused_address(KeychainKind::External);


crates/wallet/tests/wallet.rs
```diff
@@ -553,6 +553,17 @@ fn test_create_tx_default_locktime_cltv() {
     assert_eq!(psbt.unsigned_tx.lock_time.to_consensus_u32(), 100_000);
 }
 
+#[test]
+fn test_create_tx_locktime_cltv_for_specific_time() {
+    let (mut wallet, _) = get_funded_wallet_single(get_test_single_sig_cltv_for_specific_time());
+    let addr = wallet.next_unused_address(KeychainKind::External);
+    let mut builder = wallet.build_tx();
+    builder.add_recipient(addr.script_pubkey(), Amount::from_sat(25_000));
+    let psbt = builder.finish().unwrap();
```
ValuedMammal: To check that finalization works, you'll want to sign the PSBT also.

```suggestion
    let mut psbt = builder.finish().unwrap();
    let finalized = wallet.sign(&mut psbt, SignOptions::default()).unwrap();
    assert!(finalized);
```

crates/wallet/src/test_utils.rs
```diff
@@ -160,6 +160,13 @@ pub fn get_test_single_sig_cltv() -> &'static str {
     "wsh(and_v(v:pk(cVpPVruEDdmutPzisEsYvtST1usBR3ntr8pXSyt6D2YYqXRyPcFW),after(100000)))"
 }
 
+/// `wsh` descriptor with policy `and(pk(A),after(500_000_000))`
+/// the parameter passed to miniscript fragment `after` has to equal or grather than 500_000_000
+/// in order to use a lock based on unix time
+pub fn get_test_single_sig_cltv_for_specific_time() -> &'static str {
+    "wsh(and_v(v:pk(cVpPVruEDdmutPzisEsYvtST1usBR3ntr8pXSyt6D2YYqXRyPcFW),after(500000000)))"
```
notmandatory: ⛏️  just a nit but we should use a timestamp that's sometime after the genesis block. Sometimes people use these tests as examples so will make more sense to use a more recent time. Also the number is easier to read in the code with the underscores.
```suggestion
    "wsh(and_v(v:pk(cVpPVruEDdmutPzisEsYvtST1usBR3ntr8pXSyt6D2YYqXRyPcFW),after(1_734_230_218)))"
```

crates/wallet/src/test_utils.rs
```diff
@@ -160,6 +160,13 @@ pub fn get_test_single_sig_cltv() -> &'static str {
     "wsh(and_v(v:pk(cVpPVruEDdmutPzisEsYvtST1usBR3ntr8pXSyt6D2YYqXRyPcFW),after(100000)))"
 }
 
+/// `wsh` descriptor with policy `and(pk(A),after(500_000_000))`
+/// the parameter passed to miniscript fragment `after` has to equal or grather than 500_000_000
```
notmandatory:
```suggestion
/// the parameter passed to miniscript fragment `after` has to equal or greater than 500_000_000
```

crates/wallet/src/test_utils.rs
f3r10: I tried to update the descriptor as you mentioned, but there is an error with the format:
```descriptors must be valid: Miniscript(Unexpected("unexpected «1_734_230_218»"))```
It looks like miniscript does not have support for numbers with underscores.

File: crates/wallet/src/test_utils.rs (hunk 3 above)
Comment (ValuedMammal) [reply crates/wallet/src/test_utils.rs]: @f3r10 I don't think the extra commentary is needed, but we can make it a code comment as a note to developers and leave the doc string as a single line:

```rs
/// `wsh` descriptor with policy `and(pk(A),after(1734230218))`
```

Can you give this helper a better name, like `get_test_single_sig_cltv_timestamp`. I agree with the suggestion to use a realistic timestamp (without underscores).

f3r10: @ValuedMammal fixed with your comments.

However, based on what we talked about at the previous meeting, I was about to close this PR. 
Do you think that this test is testing what is supposed to?

Maybe when the implementation of the planning module that you are working on is finished, this test will be more relevant. :thinking:

ValuedMammal (approved):
ACK af30026d6b4159f7c3001411996fa038319c8bff

crates/wallet/src/test_utils.rs
ValuedMammal: yes I think we should have the test, I'm not sure of another way to test that time based timelocks are satisfied.

crates/wallet/src/test_utils.rs
```diff
@@ -160,6 +160,13 @@ pub fn get_test_single_sig_cltv() -> &'static str {
     "wsh(and_v(v:pk(cVpPVruEDdmutPzisEsYvtST1usBR3ntr8pXSyt6D2YYqXRyPcFW),after(100000)))"
 }
 
+/// `wsh` descriptor with policy `and(pk(A),after(1_734_230_218))`
+// the parameter passed to miniscript fragment `after` has to equal or grather than 500_000_000
```
oleonardolima:
```suggestion
// the parameter passed to miniscript fragment `after` has to equal or greater than 500_000_000
```

oleonardolima (commented):
tACK af30026d6b4159f7c3001411996fa038319c8bff

I just left a suggestion to fix a typo on the code comment.

ValuedMammal:
In 5471c3b2939ad0f4555cc273355bc1ede5ce6f44: the commit doesn't appear to be signed

ValuedMammal (approved):
ACK 2c11d46d 

Thanks @f3r10

PR #1684, opened 2024-11-13
Refactor file store

### LINKED ISSUE #1517: Appending changesets to `Store` after opening causes overwriting

nymius:
### Description

The `Store::open` method doesn't recovers the previous `Store` state saved in the file and emplaces the file pointer just after the magic bytes prefix, this later is agravated by `Store::append_changeset` which sets the file pointer after the last written changeset. The combination of both causes the lost of any previous changeset there may have been in the file.

Is natural to think this shouldn't be the expected behavior, as @KnowWhoami pointed out in #1517, and the `Store` should recover the previous changesets stored in the file store.

The approach proposed in #1632 triggered a discussion about more changes in `Store`, which motivated the current refactor.

Besides the fix for #1517, the following methods have been changed/renamed/repurposed in Store:
- `create`: create file and retrieve store, if exists fail.
- `load`: load changesets from file, aggregate them and return aggregated changeset and `Store`. If there are problems with decoding, fail.
- `dump`: aggregate all changesets and return them.
- `load_or_create`: try load or fallback to create.

Fixes #1517.
Overrides #1632.

### Notes to the reviewers

**Load** return type is `Result<(Option<C>, Store), StoreErrorWithDump>` to allow methods which doesn't use `WalletPersister` to get the aggregated changeset right away.

**Dump** is kept to allow `WalletPersister::initialize` method to retrieve the aggregated changeset without forcing the inclusion of the additional parameters needed by `store::load` to the trait, which would also affect the `rusqlite` implementation.

#### Added:
- `StoreError` enum, which includes `Io`, `Bincode` and `InvalidMagicBytes`.
- "not intended for production" notice in general `README` for `file store`.

#### Changed:
- `Store::create_new` to `Store::create`, with new return type: `Result<Self, StoreError>`
- `Store::open` to `Store::load`, with new return type: `Result<(Self, Option<C>), StoreErrorWithDump<C>>`
- `Store::open_or_create` to `Store::load_or_create`, with new return type: `Result<(Option<C>, Self), StoreErrorWithDump<C>>
`
- `Store::aggregate_changesets` to `Store::dump`, with new return type: `Result<Option<C>, StoreErrorWithDump<C>>`
- `FileError` to `StoreError`
- `AggregateChangesetsError` to `StoreErrorWithDump`, which now can include all the variants of `StoreError` in the error field.

#### Deleted:
- `IterError` deleted.

#### All Submissions:

* ~~I've signed all my commits~~
* ~~I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)~~
* ~~I ran `cargo fmt` and `cargo clippy` before committing~~

#### New Features:

* ~~I've added tests for the new feature~~
* ~~I've added docs for the new feature~~

#### Bugfixes:

* ~~This pull request breaks the existing API~~
* ~~I've added tests to reproduce the issue which are now passing~~
* ~~I'm linking the issue being fixed by this PR~~

### DIFF
--- a/crates/file_store/src/lib.rs
+++ b/crates/file_store/src/lib.rs
@@ -13,7 +13,7 @@ pub(crate) fn bincode_options() -> impl bincode::Options {
 
 /// Error that occurs due to problems encountered with the file.
 #[derive(Debug)]
-pub enum FileError {
+pub enum StoreError {
     /// IO error, this may mean that the file is too short.
     Io(io::Error),
     /// Magic bytes do not match what is expected.
@@ -22,7 +22,7 @@ pub enum FileError {
     Bincode(bincode::ErrorKind),
 }
 
-impl core::fmt::Display for FileError {
+impl core::fmt::Display for StoreError {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         match self {
             Self::Io(e) => write!(f, "io error trying to read file: {}", e),
@@ -36,10 +36,10 @@ impl core::fmt::Display for FileError {
     }
 }
 
-impl From<io::Error> for FileError {
+impl From<io::Error> for StoreError {
     fn from(value: io::Error) -> Self {
         Self::Io(value)
     }
 }
 
-impl std::error::Error for FileError {}
+impl std::error::Error for StoreError {}
--- a/crates/file_store/src/lib.rs
+++ b/crates/file_store/src/lib.rs
@@ -18,6 +18,8 @@ pub enum FileError {
     Io(io::Error),
     /// Magic bytes do not match what is expected.
     InvalidMagicBytes { got: Vec<u8>, expected: Vec<u8> },
+    /// Failure to decode data from the file.
+    Bincode(bincode::ErrorKind),
 }
 
 impl core::fmt::Display for FileError {
@@ -29,6 +31,7 @@ impl core::fmt::Display for FileError {
                 "file has invalid magic bytes: expected={:?} got={:?}",
                 expected, got,
             ),
+            Self::Bincode(e) => write!(f, "bincode error while reading entry {}", e),
         }
     }
 }
--- a/crates/file_store/src/store.rs
+++ b/crates/file_store/src/store.rs
@@ -41,16 +41,8 @@ where
     where
         P: AsRef<Path>,
     {
-        if file_path.as_ref().exists() {
-            // `io::Error` is used instead of a variant on `FileError` because there is already a
-            // nightly-only `File::create_new` method
-            return Err(FileError::Io(io::Error::new(
-                io::ErrorKind::Other,
-                "file already exists",
-            )));
-        }
         let mut f = OpenOptions::new()
-            .create(true)
+            .create_new(true)
             .read(true)
             .write(true)
             .truncate(true)
--- a/crates/file_store/src/store.rs
+++ b/crates/file_store/src/store.rs
@@ -109,18 +109,24 @@ where
     /// # let new_len = file.seek(SeekFrom::End(-2))?;
     /// # file.set_len(new_len)?;
     ///
-    /// let mut new_store = match Store::<TestChangeSet>::load(&MAGIC_BYTES, &file_path) {
+    /// let (mut new_store, _aggregate_changeset) =
+    ///     match Store::<TestChangeSet>::load(&MAGIC_BYTES, &file_path) {
     /// #   Ok(_) => panic!("should have errored"),
-    ///     Ok((store, _aggregated_changeset)) => store,
-    ///     Err(StoreErrorWithDump { changeset, .. }) => {
-    ///         let new_file_path = file_path.with_extension("bkp");
-    ///         let mut new_store = Store::create(&MAGIC_BYTES, &new_file_path).unwrap();
-    ///         if let Some(aggregated_changeset) = changeset {
-    ///             new_store.append(&aggregated_changeset)?;
+    ///         Ok((store, changeset)) => (store, changeset),
+    ///         Err(StoreErrorWithDump { changeset, .. }) => {
+    ///             let new_file_path = file_path.with_extension("backup");
+    ///             let mut new_store =
+    ///                 Store::create(&MAGIC_BYTES, &new_file_path).expect("must create new file");
+    ///             if let Some(aggregated_changeset) = changeset {
+    ///                 new_store.append(&aggregated_changeset)?;
+    ///             }
+    ///             // The following will overwrite the original file. You will loose the corrupted
+    ///             // portion of the original file forever.
+    ///             drop(new_store);
+    ///             std::fs::rename(&new_file_path, &file_path)?;
+    ///             Store::load(&MAGIC_BYTES, &file_path).expect("must load new file")
     ///         }
-    ///         new_store
-    ///     }
-    /// };
+    ///     };
     /// #
     /// # assert_eq!(
     /// #     new_store.dump().expect("should dump changeset: {1, 2, 3} "),
--- a/crates/file_store/src/store.rs
+++ b/crates/file_store/src/store.rs
@@ -1,10 +1,10 @@
-use crate::{bincode_options, EntryIter, FileError, IterError};
+use crate::{bincode_options, EntryIter, StoreError};
 use bdk_core::Merge;
 use bincode::Options;
 use std::{
     fmt::{self, Debug},
     fs::{File, OpenOptions},
-    io::{self, Read, Seek, Write},
+    io::{self, Read, Write},
     marker::PhantomData,
     path::Path,
 };
@@ -14,10 +14,7 @@ use std::{
 /// > ⚠ This is a development/testing database. It does not natively support backwards compatible
 /// > BDK version upgrades so should not be used in production.
 #[derive(Debug)]
-pub struct Store<C>
-where
-    C: Sync + Send,
-{
+pub struct Store<C> {
     magic_len: usize,
     db_file: File,
     marker: PhantomData<C>,
@@ -25,19 +22,15 @@ where
 
 impl<C> Store<C>
 where
-    C: Merge
-        + serde::Serialize
-        + serde::de::DeserializeOwned
-        + core::marker::Send
-        + core::marker::Sync,
+    C: Merge + serde::Serialize + serde::de::DeserializeOwned,
 {
     /// Create a new [`Store`] file in write-only mode; error if the file exists.
     ///
-    /// `magic` is the prefixed bytes to write to the new file. This will be checked when opening
-    /// the `Store` in the future with [`open`].
+    /// `magic` is the prefixed bytes to write to the new file. This will be checked when loading
+    /// the [`Store`] in the future with [`load`].
     ///
-    /// [`open`]: Store::open
-    pub fn create_new<P>(magic: &[u8], file_path: P) -> Result<Self, FileError>
+    /// [`load`]: Store::load
+    pub fn create<P>(magic: &[u8], file_path: P) -> Result<Self, StoreError>
     where
         P: AsRef<Path>,
     {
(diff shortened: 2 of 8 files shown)


crates/wallet/tests/wallet.rs
```diff
@@ -116,8 +116,11 @@ fn wallet_is_persisted() -> anyhow::Result<()> {
 
     run(
         "store.db",
-        |path| Ok(bdk_file_store::Store::create_new(DB_MAGIC, path)?),
-        |path| Ok(bdk_file_store::Store::open(DB_MAGIC, path)?),
+        |path| Ok(bdk_file_store::Store::create(DB_MAGIC, path)?),
+        |path| {
+            let (_, store) = bdk_file_store::Store::load(DB_MAGIC, path)?;
+            Ok(store)
+        },
```
ValuedMammal: nit: This could be condensed slightly. similar suggestion below in `wallet_load_checks`.

```suggestion
        |path| Ok(bdk_file_store::Store::load(DB_MAGIC, path)?.1),
```

crates/file_store/src/store.rs
```diff
@@ -82,87 +81,92 @@ where
...
-
-    /// Append a new changeset to the file and truncate the file to the end of the appended
-    /// changeset.
-    ///
-    /// The truncation is to avoid the possibility of having a valid but inconsistent changeset
-    /// directly after the appended changeset.
-    pub fn append_changeset(&mut self, changeset: &C) -> Result<(), io::Error> {
+    /// If garbage data is written and then valid changesets, the next load will still only
+    /// retrieve the first chunk of valid changesets. The recovery of those valied changesets after
+    /// the garbage data is responsability of the user
```
ValuedMammal:
```suggestion
    /// retrieve the first chunk of valid changesets. The recovery of those valid changesets after
    /// the garbage data is responsibility of the user.
```

crates/file_store/src/store.rs
```diff
@@ -372,17 +436,26 @@ mod test {
...
                 );
-                db.db_file.write_all(&last_changeset_bytes).unwrap();
+                // Remove file and start again
+                fs::remove_file(&file_path).expect("should remove file");
+                let mut store =
+                    Store::<TestChangeSet>::create(&TEST_MAGIC_BYTES, &file_path).unwrap();
+                for changeset in &changesets {
+                    store.append(changeset).unwrap();
+                }
+                // this is the incomplete write
```
ValuedMammal: This is the complete write, no?

It's a little unfortunate that we can only recover from a corrupt file by removing it and starting again. To better simulate a recovery you could use the agg. changeset given by the load error to write back to the new file. But I think another solution is to have `load` return a `Store` instance with the partially aggregated changeset even in case of a bincode error. Then we can just append the `last_changeset` to the existing store. What do you think?

crates/file_store/src/store.rs
```diff
@@ -63,17 +55,24 @@ where
...
-    /// [`create_new`]: Store::create_new
-    pub fn open<P>(magic: &[u8], file_path: P) -> Result<Self, FileError>
+    /// If there exist changesets in the file, [`load`] will try to aggregate them in
+    /// a single changeset to verify their integrity. If aggregation fails
+    /// [`StoreErrorWithDump`] will be returned with the [`StoreError::Bincode`] error variant in
+    /// its error field and the aggregated changeset so far in the changeset field.
+    ///
+    /// [`create`]: Store::create
+    /// [`load`]: Store::load
+    pub fn load<P>(magic: &[u8], file_path: P) -> Result<(Option<C>, Self), StoreErrorWithDump<C>>
```
ValuedMammal: I have a slight preference for returning `(Self, Option<C>)`, i.e. make `Self` the first item in the tuple, but I don't have a good reason other than it feels more ergonomic.

ValuedMammal (commented):
I did a first pass. Overall it looks good. I also made suggestions in this commit ValuedMammal/bdk@6312ca39d0ba6514450b4b64bcd9c6e9e9a139b0

If you plan to rebase make sure commits compile on their own. For example changing a method name and updating the method call sites should be in the same commit in my opinion.

crates/file_store/src/store.rs
nymius: > This is the complete write, no?

I don't think so, the following line just writes the changeset bytes up to the `short_write_len` limit, which will produce a partial write. There's something I'm not seeing?

> It's a little unfortunate that we can only recover from a corrupt file by removing it and starting again. To better simulate a recovery you could use the agg. changeset given by the load error to write back to the new file.

I prefer the above...

> But I think another solution is to have `load` return a `Store` instance with the partially aggregated changeset even in case of a bincode error. Then we can just append the `last_changeset` to the existing store. What do you think?

rather than this, because if user has stored 5 changesets: `{1, 2, 3, 4, 5}` and then 3 and 4 get corrupted: `{1, 2, x, x, 5}`, this solution will just return the aggregated `{1, 2}` changeset and add the 6th changeset: `{{1, 2}, 6}` without giving the user any notion something went wrong.

That's why I included changeset in `StoreErrorWithDump`: the user receives an error, but still has the partial aggregated changeset to create a new store if he wants. With our code signaling there are no guarantee the aggregated changeset contains all the changesets the user originally stored.

crates/wallet/tests/wallet.rs
nymius: I will change it. Consider I have a bias towards explicit semantics.

nymius:
@ValuedMammal thanks for the review, I'm already working in the changes.

crates/file_store/src/store.rs
nymius: Fixed by 68f52fd1d60cd94cd3e012d906654443a0202393

crates/wallet/tests/wallet.rs
nymius: Fixed by b93cec96d60f71025795678dda0c29586f7a27b6

crates/file_store/src/store.rs
nymius: Fixed by 5a7fe4259c7a7c62b24973acdcd1961c46d69bdd

nymius:
Rebased on 03a08bb7fe85774079447e985501f5718be91922
Haven't applied changes solicited in https://github.com/bitcoindevkit/bdk/pull/1684#discussion_r1888778129 because now I find more valuable to report the error and allow the user to recover from it instead of returning whatever aggregated changeset we could parse in a best effort approach.

ValuedMammal:
In 0f104cca:

```sh
error[E0432]: unresolved import `crate::FileError`
 --> crates/file_store/src/store.rs:1:41
  |
1 | use crate::{bincode_options, EntryIter, FileError, IterError};
  |                                         ^^^^^^^^^ no `FileError` in the root

For more information about this error, try `rustc --explain E0432`.
error: could not compile `bdk_file_store` (lib) due to 1 previous error
```

Maybe just squash in the next commit 8897a201 but keep the commit message from 8897a201 ?

nymius:
> In [0f104cc](https://github.com/bitcoindevkit/bdk/commit/0f104cca04bfd9f2f8650b4d55efa9d72fcb9329):
> 
> ```shell
> error[E0432]: unresolved import `crate::FileError`
>  --> crates/file_store/src/store.rs:1:41
>   |
> 1 | use crate::{bincode_options, EntryIter, FileError, IterError};
>   |                                         ^^^^^^^^^ no `FileError` in the root
> 
> For more information about this error, try `rustc --explain E0432`.
> error: could not compile `bdk_file_store` (lib) due to 1 previous error
> ```
> 
> Maybe just squash in the next commit [8897a20](https://github.com/bitcoindevkit/bdk/commit/8897a201c3fb18a2043df5a83d4572f3307d39a1) but keep the commit message from [8897a20](https://github.com/bitcoindevkit/bdk/commit/8897a201c3fb18a2043df5a83d4572f3307d39a1) ?

Thanks! should be fixed now. I checked each commit with:
```rust
 git rebase --exec 'cargo build; cargo test; cargo clippy; cargo fmt' 0f104c^
```

crates/file_store/src/store.rs
ValuedMammal: Gotcha. It seems like either way changesets 3, 4, and 5 will be unrecoverable unless they're somehow predetermined ahead of time?

> I don't think so, the following line just writes the changeset bytes up to the short_write_len limit, which will produce a partial write. There's something I'm not seeing?

Hm, I was thinking that the short write was above on L412

nymius: > Gotcha. It seems like either way changesets 3, 4, and 5 will be unrecoverable unless they're somehow predetermined ahead of time?

Yes, they will be unrecoverable, but I don't get what you mean by "predetermined ahead of time".

> > I don't think so, the following line just writes the changeset bytes up to the short_write_len limit, which will produce a partial write. There's something I'm not seeing?
> 
> Hm, I was thinking that the short write was above on L412

Indeed, there was something I was not seeing. You're right, it is the complete write. I was matching the comment with the wrong line.

ValuedMammal (approved):
ACK 18aa75eb I tested on `example_esplora`

crates/file_store/src/store.rs
ValuedMammal: > but I don't get what you mean by "predetermined ahead of time".

Just referring to the fact that the changesets are hardcoded in the test

nymius:
@notmandatory we have been trying to figure out with @ValuedMammal what would be the correct handling for this change.

[This semver FAQ entry](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api) implies it isn't a breaking change for wallet as long as we don't change the API, but [this single line](https://github.com/bitcoindevkit/bdk/pull/1684/files#diff-8bbf758809da8b212b6fe8f7f512ad335e7a620672c66588a8f6ed8351ecb8c8R294) is a breaking change, so I think if we don't rollback the error name to the previous one, we should wait until `wallet-2.0` .
If we rollback the name change to the old one, I think we can introduce all the changes in a minor version for wallet, and then move the name change to a future update closer to the wallet 2.0 release, or drop it altogether.

What do you think we should do?

nymius:
I tried to just change the name of `StoreErrorWithDump` and roll it back to `AggreagateChangesetsError` but the field `error` of the former cannot be changed back to `iter_error` without breaking the logic of the new changes.

So I decided to go with the option to use `bdk_file_store` as a registry dependency instead of a path dependency to avoid the force version bump of `bdk_wallet` when merging these changes.
It showed up to be more complex than I thought initially and there are some weird things like `bdk_core` depending solely on `bdk_testenv` and `bdk_chain` and not in the actual crate. I need to think more about this solution, but is the only working so far.

cc: @notmandatory @ValuedMammal

nymius:
#1860 allow this to be a breaking change without bumping `bdk_wallet` major version. I have squashed all commits addressing comments requesting changes and rebased onto latest b26ff891

crates/file_store/src/store.rs
```diff
@@ -82,87 +74,92 @@ where
...
+        };
+
+        // Get aggregated changeset
+        let aggregated_changeset = store.dump()?;
+
+        Ok((store, aggregated_changeset))
     }
 
-    /// Attempt to open existing [`Store`] file; create it if the file is non-existent.
+    /// Aggregate [`Store`] changesets and return them as a single changeset.
```
```suggestion
    /// Dump the aggregate of all changesets in [`Store`].
```

crates/file_store/src/store.rs
```diff
@@ -63,17 +48,24 @@ where
...
-    /// [`create_new`]: Store::create_new
-    pub fn open<P>(magic: &[u8], file_path: P) -> Result<Self, FileError>
+    /// If there exist changesets in the file, [`load`] will try to aggregate them in
+    /// a single changeset to verify their integrity. If aggregation fails
+    /// [`StoreErrorWithDump`] will be returned with the [`StoreError::Bincode`] error variant in
+    /// its error field and the aggregated changeset so far in the changeset field.
+    ///
+    /// [`create`]: Store::create
+    /// [`load`]: Store::load
+    pub fn load<P>(magic: &[u8], file_path: P) -> Result<(Self, Option<C>), StoreErrorWithDump<C>>
```
evanlinjin, line 145:
```suggestion
    pub fn load<P>(magic: &[u8], file_path: P) -> (Option<Self>, Result<Option<C>, StoreErrorWithDump<C>>)
```

I think we would still like to have access to `Store` even if the last changeset is corrupted in the file.

The only case where we would not like access to `Store` is when the magic bytes do not match (hence the `Option<Self>`).

I'm open for ideas for making this return type more intuitive.

crates/file_store/src/store.rs
```diff
@@ -82,87 +74,92 @@ where
...
-    /// changeset.
+    /// If multiple garbage writes are produced on the file, the next load will only retrieve the
+    /// first chunk of valid changesets.
     ///
-    /// The truncation is to avoid the possibility of having a valid but inconsistent changeset
-    /// directly after the appended changeset.
-    pub fn append_changeset(&mut self, changeset: &C) -> Result<(), io::Error> {
+    /// If garbage data is written and then valid changesets, the next load will still only
+    /// retrieve the first chunk of valid changesets. The recovery of those valid changesets after
+    /// the garbage data is responsibility of the user.
```
evanlinjin, line 238: Why is this important? How do we write garbage data with this API?

crates/file_store/src/lib.rs
```diff
@@ -13,14 +13,16 @@ pub(crate) fn bincode_options() -> impl bincode::Options {
...
 /// Error that occurs due to problems encountered with the file.
 #[derive(Debug)]
-pub enum FileError {
+pub enum StoreError {
     /// IO error, this may mean that the file is too short.
     Io(io::Error),
     /// Magic bytes do not match what is expected.
     InvalidMagicBytes { got: Vec<u8>, expected: Vec<u8> },
+    /// Failure to decode data from the file.
+    Bincode(bincode::ErrorKind),
```
evanlinjin, line 22: @nymius mentioned that it might be useful to separate out the "corrupted" portion of the file. In this case, adding a field for the file position of the start of the corrupted portion will be helpful.

evanlinjin (changes requested):
I like where this is going.

crates/file_store/src/store.rs
nymius: I consider there is no case when this API writes garbage data. I think the only way to corrupt the database file is to do it externally. Maybe "write" is not the best way to express that.

crates/file_store/src/lib.rs
nymius: [I considered that solution before](https://github.com/bitcoindevkit/bdk/pull/1632/files#diff-d54e16114a77406783f25f85482ebc29cafd36ba0a1e3907e35855e9d5a8ccb5R50), but we agreed with @ValuedMammal it was not needed considering the goals of the crate.

crates/file_store/src/store.rs
nymius: I'm dismissing this change in favor of explaining [how to recover the `Store` after a failed read](https://github.com/bitcoindevkit/bdk/pull/1684/files#diff-a2a5264e9fd43373450a32f3b6981e1e5307bc8fff7bacdb008c7c8a379ef379R61-R135)

nymius:
Rebased onto 362c3dc8

evanlinjin (approved):
ACK 54251a7c9fdf08bced132be3851207917cbc36ab

PR #1721, opened 2024-11-15
chore(deps): bump hashbrown to v0.14.5

tvolk131:
Bump [`hashbrown`](https://crates.io/crates/hashbrown) to v0.15

I ran [`cargo build-all-features` and `cargo test-all-features`](https://github.com/frewsxcv/cargo-all-features) locally for the `core` crate with no issues. Here's the output for the former:

```
cargo build-all-features

    Building crate=bdk_core features=[]
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
    Building crate=bdk_core features=[hashbrown]
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
    Building crate=bdk_core features=[serde]
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.05s
    Building crate=bdk_core features=[std]
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
    Building crate=bdk_core features=[hashbrown,serde]
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
    Building crate=bdk_core features=[hashbrown,std]
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
    Building crate=bdk_core features=[serde,std]
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
    Building crate=bdk_core features=[hashbrown,serde,std]
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
```

The relevant changelog can be found between [here](https://github.com/rust-lang/hashbrown/blob/master/CHANGELOG.md#v0100---2021-01-16) and [here](https://github.com/rust-lang/hashbrown/blob/master/CHANGELOG.md#v0151---2024-11-03)

The only notable breaking change I found was that the MSRV was bumped to v1.63. However, this is the same MSRV as we already use, so it's not a breaking change for us.

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 [dependencies]
 bitcoin = { version = "0.32", default-features = false }
 serde = { version = "1", optional = true, features = ["derive", "rc"] }
-hashbrown = { version = "0.9.1", optional = true }
+hashbrown = { version = "0.14.5", optional = true, default-features = false, features = ["ahash", "inline-more"] }
 
 [features]
 default = ["std"]


crates/core/Cargo.toml
```diff
@@ -13,7 +13,7 @@ readme = "README.md"
 [dependencies]
 bitcoin = { version = "0.32", default-features = false }
 serde = { version = "1", optional = true, features = ["derive", "rc"] }
-hashbrown = { version = "0.9.1", optional = true }
+hashbrown = { version = "0.15", optional = true }
```
notmandatory: Looks like hashbrown 0.15.1 bumped the MSRV to 1.65 😞 . Are you OK with making this a pinned version?
```suggestion
hashbrown = { version = "=0.15.0", optional = true }
```

tnull: I don't think BDK should hard-pin this in `Cargo.toml`. But since they unyanked 0.15.0 (see https://github.com/rust-lang/hashbrown/issues/585 for the discussion around that), it should be fine to pin it back in CI.

tnull: While we're here, it might be a good opportunity to set `default-features = false` and only enable features BDK actually uses/requires (maybe even none are required?).

notmandatory: Agreed, if we don't need the default features let's turn them off.

notmandatory:
Sorry about the force pushes, I committed my suggestion instead of closing it. ~~But at least this give the branch a chance to run with the fixed hashbrown dependency.~~

Looks like hashbrown team doesn't want to fix this and Tokio pinned it in their CI and readme. We can do the same for this PR. Let me know if you want me to push a commit with those changes.

tvolk131:
> Let me know if you want me to push a commit with those changes.

Pushed! Looks like the only feature needed is `default-hasher`. I re-ran `cargo build-all-features` and everything's passing locally.

crates/core/Cargo.toml
tvolk131: Done. Only `default-hasher` is needed.

oleonardolima (commented):
utACK eba1dd08d1aa3324bc1a08285ef8636d2367a2af

It's failing on CI at the MSRV step, you'll need to pin it to 0.15.0 [here](https://github.com/oleonardolima/bdk/blob/3bc45b5431014af3a8255a201e49227584e953b5/.github/workflows/cont_integration.yml#L41-L53) and update the `README.md` as well.

notmandatory:
For some reason getting errors with hashbrown pinned to 0.15.0 for MSRV 1.63.0, also reproducible locally:
```
---- test_into_tx_graph stdout ----
thread 'test_into_tx_graph' panicked at 'a failure ordering can't be stronger than a success ordering', /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/sync/atomic.rs:2700:18

---- tx_can_become_unconfirmed_after_reorg stdout ----
thread 'tx_can_become_unconfirmed_after_reorg' panicked at 'a failure ordering can't be stronger than a success ordering', /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/sync/atomic.rs:2700:18

```

notmandatory:
I force pushed a commit to only bump hashbrown to `0.14.5` ~~and disable default features~~, building and testing for MSRV looks like it's working now. EDIT: with this version I have to leave default features on.

tnull:
> EDIT: with this version I have to leave default features on.

No, I believe you simply need to enable `ahash` as a default hasher, although `inline-more` probably also makes sense to leave enabled.

So 
```
hashbrown = { version = "0.14.5", optional = true, default-features = false, features = ["ahash", "inline-more"] }
```
works for me.

notmandatory:
> > EDIT: with this version I have to leave default features on.
> 
> No, I believe you simply need to enable `ahash` as a default hasher, although `inline-more` probably also makes sense to leave enabled.

@tnull thanks for testing it out, I pushed your suggestion and CI passing so all good. Just FYI the only default feature not included is: 

allocator-api2: Enables support for allocators that support allocator-api2.

tvolk131:
Thanks for updating and testing this guys! Is there anything else needed here, or is this good to merge?

notmandatory (approved):
ACK 52fa5404bdecc96939f47f7567e2605f449b4b2a

notmandatory:
@tvolk131 I was just giving anyone else interested a chance to comment, but looks good to me so I'll merge it. Thanks for the suggestion to do this version bump.

PR #1725, opened 2024-11-16
Check recent blocks are not empty

### LINKED ISSUE #1696: bdk_esplora: handle empty /blocks response
"a rogue Esplora server can make an application crash by returning an empty list of blocks to a /blocks request, as latest_blocks is assumed to never be empty in fetch_block()"

rustaceanrob:
Closes #1696 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/esplora/src/async_ext.rs
+++ b/crates/esplora/src/async_ext.rs
@@ -211,6 +211,9 @@ async fn chain_update(
     local_tip: &CheckPoint,
     anchors: &BTreeSet<(ConfirmationBlockTime, Txid)>,
 ) -> Result<CheckPoint, Error> {
+    if latest_blocks.is_empty() {
+        return Ok(local_tip.clone());
+    }
     let mut point_of_agreement = None;
     let mut conflicts = vec![];
     for local_cp in local_tip.iter() {
--- a/crates/esplora/src/blocking_ext.rs
+++ b/crates/esplora/src/blocking_ext.rs
@@ -201,6 +201,9 @@ fn chain_update(
     local_tip: &CheckPoint,
     anchors: &BTreeSet<(ConfirmationBlockTime, Txid)>,
 ) -> Result<CheckPoint, Error> {
+    if latest_blocks.is_empty() {
+        return Ok(local_tip.clone());
+    }
     let mut point_of_agreement = None;
     let mut conflicts = vec![];
     for local_cp in local_tip.iter() {


oleonardolima (commented):
It looks good, but it seems to partially address the problem.

I'm trying to wrap my mind on the behavior when the `latest_blocks` is empty. I wonder if it shouldn't be handled on the upstream fns too?

Otherwise, the same rogue Esplora server would still be able to crash the application as we expect to have a valid point of agreement here:
https://github.com/bitcoindevkit/bdk/blob/3bc45b5431014af3a8255a201e49227584e953b5/crates/esplora/src/blocking_ext.rs#L225

rustaceanrob:
That's a good point. I think these `is_empty` checks should simply be done in the first line of `chain_update` and this would be sufficient.

oleonardolima (commented):
Concept ACK f3efbdaea95ceee762c31219f14a4181bcd92eeb

notmandatory (approved):
utACK f3efbdaea95ceee762c31219f14a4181bcd92eeb

notmandatory:
Can you add a simple test to exercise this condition in case we change the fix implementation later we make sure this remains fixed.

I also had some discussion with @oleonardolima and @ValuedMammal and it would be good to have the test first and make sure without any changes this does in fact gets to and fails an at expect.

rustaceanrob:
I can try to work on that tomorrow. Heads up I will not be able to work on much next week

ValuedMammal:
Since the checkpoint is an optional field of the sync/scan response it may also make sense to have `chain_update` return a `Result<Option<CheckPoint>, Error>`.

rustaceanrob:
I was thinking that this should be an error variant in the Esplora client, as it should not ever respond with empty blocks when requested. I am unable to do much this week but I can review any new PRs that replace this one

crates/esplora/src/async_ext.rs
```diff
@@ -211,6 +211,9 @@ async fn chain_update(
     local_tip: &CheckPoint,
     anchors: &BTreeSet<(ConfirmationBlockTime, Txid)>,
 ) -> Result<CheckPoint, Error> {
+    if latest_blocks.is_empty() {
+        return Ok(local_tip.clone());
+    }
```
evanlinjin, line 216: Please include docs explaining why we are doing this.

evanlinjin (changes requested):

evanlinjin:
I'm thinking, if the server does something unexpected, maybe it is best to error out? Otherwise there is no way the caller will figure it out.

rustaceanrob:
> I'm thinking, if the server does something unexpected, maybe it is best to error out? Otherwise there is no way the caller will figure it out.

Right, I agree, but that would push the error variant into `rust_esplora_client` and would no longer be a `bdk_esplora` patch correct?

rustaceanrob:
Would be resolved by bitcoindevkit/rust-esplora-client#113

rustaceanrob:
Closed by bitcoindevkit/rust-esplora-client#113

PR #1726, opened 2024-11-16
Enforce OP_RETURN standardness

### LINKED ISSUE #49: Enforce standardness rules for TxBuilder::add_data
~~"TxBuilder::add_data could enforce common standardness rules (max 80 bytes and at most one such output per transaction). It would also prevent a potential panic to be triggered from within rust-bitcoin (in push_slice_no_opt) if the data is too large, although i don't think it's realistic as a request for creating a 4GiB OP_RETURN would probably fail before then."~~

The policy for OP_RETURN has changed since [Bitcoin Core v30.0](https://bitcoincore.org/en/releases/30.0/)+, per release notes:

"-datacarriersize is increased to 100,000 by default, which effectively uncaps the limit (as the maximum transaction size limit will be hit first). It can be overridden with -datacarriersize=83 to revert to the limit enforced in previous versions. ([#32406](https://github.com/bitcoin/bitcoin/pull/32406))

Multiple data carrier (OP_RETURN) outputs in a transaction are now permitted for relay and mining. The -datacarriersize limit applies to the aggregate size of the scriptPubKeys across all such outputs in a transaction, not including the scriptPubKey size itself. ([#32406](https://github.com/bitcoin/bitcoin/pull/32406))"

### LINKED ISSUE #1718: Enforce standardness rules for TxBuilder::add_data
"TxBuilder::add_data could enforce common standardness rules (max 80 bytes and at most one such output per transaction). It would also prevent a potential panic to be triggered from within rust-bitcoin (in push_slice_no_opt) if the data is too large, although i don't think it's realistic as a request for creating a 4GiB OP_RETURN would probably fail before then."

rustaceanrob:
Closes  bitcoindevkit/bdk-tx#49

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/error.rs
+++ b/crates/wallet/src/wallet/error.rs
@@ -104,6 +104,10 @@ pub enum CreateTxError {
     MissingNonWitnessUtxo(OutPoint),
     /// Miniscript PSBT error
     MiniscriptPsbt(MiniscriptPsbtError),
+    /// Multiple recipients are OP_RETURN outputs
+    MultipleOpReturn,
+    /// The OP_RETURN data contains too many bytes
+    MaxDataCarrierSize,
 }
 
 impl fmt::Display for CreateTxError {
@@ -170,6 +174,12 @@ impl fmt::Display for CreateTxError {
             }
             CreateTxError::MiniscriptPsbt(err) => {
                 write!(f, "Miniscript PSBT error: {}", err)
+            }
+            CreateTxError::MultipleOpReturn => {
+                write!(f, "Multiple recipients are OP_RETURN outputs")
+            }
+            CreateTxError::MaxDataCarrierSize => {
+                write!(f, "The OP_RETURN data contains too many bytes")
             }
         }
     }
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -87,6 +87,7 @@ pub use persisted::*;
 pub use utils::IsDust;
 
 const COINBASE_MATURITY: u32 = 100;
+const MAX_OP_RETURN_BYTES: usize = 83;
 
 /// A Bitcoin wallet
 ///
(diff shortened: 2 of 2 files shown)


ValuedMammal:
Should we use a default max size of 83 as shown here?

https://github.com/bitcoin/bitcoin/blob/110183746150428e6385880c79f8c5733b1361ba/src/policy/policy.h#L68-L72

crates/wallet/src/wallet/mod.rs
```diff
@@ -1388,12 +1389,23 @@ impl Wallet {
...
 
+            if script_pubkey.is_op_return() {
+                if contains_op_return {
+                    return Err(CreateTxError::MultipleOpReturn);
+                }
+                if script_pubkey.len() > MAX_OP_RETURN_BYTES {
+                    return Err(CreateTxError::MaxDataCarrierSize);
+                }
+                contains_op_return = true;
+            }
```
oleonardolima, line 1407: Any reason to not use `is_standard_op_return` instead ?

rustaceanrob: I don't think that is a member of `bitcoin 0.32.4`

oleonardolima: Oh, scratch that. Unfortunately, it's not available int that version.

oleonardolima (approved):
utACK f7d6dcf2a453f31170107870cb56775207fc2faf

crates/wallet/src/wallet/error.rs
```diff
@@ -104,6 +104,10 @@ pub enum CreateTxError {
     MissingNonWitnessUtxo(OutPoint),
     /// Miniscript PSBT error
     MiniscriptPsbt(MiniscriptPsbtError),
+    /// Multiple recipients are OP_RETURN outputs
+    MultipleOpReturn,
```
evanlinjin, line 108: Why are we disallowing multiple `OP_RETURN` outputs?

crates/wallet/src/wallet/mod.rs
evanlinjin, line 1: Although `OP_RETURN` outputs larger than 83 are non-standard, they are still considered to be valid transactions. I think we should still allow building non-standard txs in `build_tx`.

I think the better approach would be to have `TxBuilder::add_data` return an error. This provides instant-feedback for the caller and we can document the restriction on the method.

crates/wallet/src/wallet/error.rs
rustaceanrob: For the same reason as the byte limitation, it would be non-standard

rustaceanrob:
I think this is best put off for 2.0.0. The functionality to build OP_RETURN transactions is present and this is really a matter of standardness, which is up for debate

crates/wallet/src/wallet/error.rs
evanlinjin: Fair, I didn't know this was a policy. I think it's best to document this and include a 🔗 to the code enforcing the policy.

https://github.com/bitcoin/bitcoin/blob/3867d2421ae45c9093ceb355f5373c09e6f93c5d/src/policy/policy.cpp#L162

rustaceanrob:
This will likely be resolved with a new `tx_builder` crate by `2.0.0` and wouldn't be worth the rebase. Closing

PR #1727, opened 2024-11-16
Use `bitcoin::constants::COINBASE_MATURITY`

### LINKED ISSUE #1692: Use rust-bitcoin's COINBASE_MATURITY
"No need to re-define rust-bitcoin's COINBASE_MATURITY."

rustaceanrob:
Closes #1692 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/lib.rs
+++ b/crates/chain/src/lib.rs
@@ -77,9 +77,6 @@ pub extern crate serde;
 #[macro_use]
 extern crate std;
 
-/// How many confirmations are needed f or a coinbase output to be spent.
-pub const COINBASE_MATURITY: u32 = 100;
-
 /// A wrapper that we use to impl remote traits for types in our crate or dependency crates.
 pub struct Impl<T>(pub T);
 
--- a/crates/chain/src/chain_data.rs
+++ b/crates/chain/src/chain_data.rs
@@ -1,6 +1,6 @@
-use bitcoin::{OutPoint, TxOut, Txid};
+use bitcoin::{constants::COINBASE_MATURITY, OutPoint, TxOut, Txid};
 
-use crate::{Anchor, COINBASE_MATURITY};
+use crate::Anchor;
 
 /// Represents the observed position of some chain data.
 ///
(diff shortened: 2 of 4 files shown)


crates/chain/src/lib.rs
```diff
@@ -77,9 +77,6 @@ pub extern crate serde;
 #[macro_use]
 extern crate std;
 
-/// How many confirmations are needed f or a coinbase output to be spent.
-pub const COINBASE_MATURITY: u32 = 100;
-
```
ErikDeSmedt, line 82: This was a public method and removing it breaks the public API.

Could you add `pub use bitcoin::constants::COINBASE_MATURITY` in the top of this file?

rustaceanrob: I am unsure if stylistic or for another reason, but BDK does not export types via `pub use` anywhere else in the crate. What do you think @notmandatory @ValuedMammal

oleonardolima (commented):
utACK 3869705bc90e85e958f26ba11e2675383a569f8e

crates/chain/src/lib.rs
oleonardolima: Although it's a breaking change on the API, it's fixing the API and you can use it from rust-bitcoin (e.g. `bdk_chain::bitcoin::constants::COINBASE_MATURITY`), which is re-exported.

oleonardolima (commented):
After answering the comment above I noticed it, and I guess you'd need to update the commit message to `fix(wallet, chain)!: ...` instead.

crates/chain/src/lib.rs
notmandatory: I agree this is a breaking change, but it's OK to require downstream projects to switch to explicitly use `bdk_chain::bitcoin::constants::COINBASE_MATURITY` since we'll be rolling this out in the 1.0.0 major release.

notmandatory (approved):
ACK 309a6077d0fa263c03e5629568ae94844be0c90f

oleonardolima (approved):
ACK 309a6077d0fa263c03e5629568ae94844be0c90f

PR #1728, opened 2024-11-18
Feature(testenv)/run regtest tests in series

riverKanies:
### Description

This PR is a proposed solution to https://github.com/bitcoindevkit/bdk_wallet/issues/43
It enforces running tests that use bitcoind or electrsd in series (rather than parallel)

### Notes to the reviewers

Originally I had suggested just including a test troubleshooting section in docs to show ways to run in series

#### Bugfixes:

### DIFF
--- a/crates/bitcoind_rpc/Cargo.toml
+++ b/crates/bitcoind_rpc/Cargo.toml
@@ -23,6 +23,7 @@ bdk_core = { path = "../core", version = "0.3.0", default-features = false }
 [dev-dependencies]
 bdk_testenv = { path = "../testenv", default-features = false }
 bdk_chain = { path = "../chain" }
+serial_test = { version = "=0.8.0" }
 
 [features]
 default = ["std"]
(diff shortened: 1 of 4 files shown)


oleonardolima (commented):
Even though it's under `dev-dependencies` and being used exclusively for tests, I'm not entirely sure if we do need this one. Are there any major improvements that it adds to the tests?

riverKanies:
@oleonardolima this is so tests work with `cargo test` (on mac or wherever). I guess it could make tests on linux a little slower if it doesn't have the issue running parallel. I softof feel that it makes sense to run blockchain client tests in series regardless since tests would be sharing state otherwise. maybe linux just does this by default, which would mean there's no performance cost

riverKanies:
should we update rustc version? https://github.com/bitcoindevkit/bdk/actions/runs/11896104128/job/33175482146?pr=1728

I can just downgrade serial_test for now

update: setting serial_test = { version = "=3.0.0" } to remove scc dep which requires rustc 1.65

riverKanies:
hmm, looks like 3.0.0 also has a dep on 1.65
ended up having to force my local to build on 1.63 and set msrv and try every version all the way down to 0.8.0 😅 

(build failed again https://github.com/bitcoindevkit/bdk/actions/runs/11924661651/job/33299306353?pr=1728)

notmandatory:
As [mentioned](https://github.com/bitcoindevkit/bdk_wallet/issues/43) on the issue I'm not on board with serializing all tests. This is probably something better to just document as a caveat for MacOS users so we can still by default run tests in parallel for linux users (and CI).

riverKanies:
@notmandatory ok, I was wondering if it would effect performance. I was thinking maybe linux ran in series by default or something and that's why we're only seeing this on mac. I'll close this and we can just document the issue in the other pr

PR #1732, opened 2024-11-19
chore(core)!: rename `SyncResult` to `SyncResponse`

### LINKED ISSUE #1647: SyncResult should not have "Result" in the name
A "Result" type in rust is conventionally always meant to be a enum with an error variant. 

This should be called `SyncResponse` since it is what you get after sending a `SyncRequest`.

The present name should be deprecated and type alias to `SyncRequest`.

oleonardolima:
fixes #1647 

### Description

As mentioned in #1647 issue, the usage of `Result` in Rust is conventionally meant to be an enum type with error variant, and it's appropriate here.

It should be `SyncResponse` instead, as it's the most appropriate alongside the other types `SyncRequest` and `SyncProgress` already being used.

### Changelog notice

- Change `bdk_core::spk_client`'s `SyncResult` to `SyncResponse`.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/core/src/spk_client.rs
+++ b/crates/core/src/spk_client.rs
@@ -322,14 +322,14 @@ impl<I> SyncRequest<I> {
 /// See also [`SyncRequest`].
 #[must_use]
 #[derive(Debug)]
-pub struct SyncResult<A = ConfirmationBlockTime> {
+pub struct SyncResponse<A = ConfirmationBlockTime> {
     /// Relevant transaction data discovered during the scan.
     pub tx_update: crate::TxUpdate<A>,
     /// Changes to the chain discovered during the scan.
     pub chain_update: Option<CheckPoint>,
 }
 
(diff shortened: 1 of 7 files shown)


notmandatory (approved):
ACK 1411cb8d58f669054c10c0dcfe67e6b271368544

Looks good, all `SyncResult`s renamed.

ValuedMammal:
Would it make sense to apply the same reasoning to `FullScanResult` by renaming it to `FullScanResponse`?

oleonardolima:
> Would it make sense to apply the same reasoning to `FullScanResult` by renaming it to `FullScanResponse`?

Good point! I think it does make sense, added a new commit with the changes so I can either squash it or drop it depending on other opinions.

notmandatory (approved):
reACK 3b03c7bed2d132fbcdb027a8445f47f0b85d99fa

PR #1733, opened 2024-11-19
feat(chain,wallet)!: Transitive `ChainPosition`

evanlinjin:
### Description

Change `ChainPosition` to be able to represent transitive anchors and unconfirm-without-last-seen values.

As mentioned in https://github.com/bitcoindevkit/bdk/pull/1670#issuecomment-2484535021, we want this merged first so that we have minimal changes to the API after 1670 is merged.

### Changelog notice

* Change `ChainPosition` so that it is able to represent transitive anchors and unconfirmed-without-last-seen values.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/chain_data.rs
+++ b/crates/chain/src/chain_data.rs
@@ -15,16 +15,30 @@ use crate::Anchor;
     ))
 )]
 pub enum ChainPosition<A> {
-    /// The chain data is seen as confirmed, and in anchored by `A`.
-    Confirmed(A),
-    /// The chain data is not confirmed and last seen in the mempool at this timestamp.
-    Unconfirmed(u64),
+    /// The chain data is confirmed as it is anchored in the best chain by `A`.
+    Confirmed {
+        /// The [`Anchor`].
+        anchor: A,
+        /// Whether the chain data is anchored transitively by a child transaction.
+        ///
+        /// If the value is `Some`, it means we have incomplete data. We can only deduce that the
+        /// chain data is confirmed at a block equal to or lower than the block referenced by `A`.
+        transitively: Option<Txid>,
+    },
+    /// The chain data is not confirmed.
+    Unconfirmed {
+        /// When the chain data is last seen in the mempool.
+        ///
+        /// This value will be `None` if the chain data was never seen in the mempool and only seen
+        /// in a conflicting chain.
+        last_seen: Option<u64>,
+    },
 }
 
 impl<A> ChainPosition<A> {
     /// Returns whether [`ChainPosition`] is confirmed or not.
     pub fn is_confirmed(&self) -> bool {
-        matches!(self, Self::Confirmed(_))
+        matches!(self, Self::Confirmed { .. })
     }
 }
 
@@ -32,8 +46,14 @@ impl<A: Clone> ChainPosition<&A> {
     /// Maps a [`ChainPosition<&A>`] into a [`ChainPosition<A>`] by cloning the contents.
     pub fn cloned(self) -> ChainPosition<A> {
         match self {
-            ChainPosition::Confirmed(a) => ChainPosition::Confirmed(a.clone()),
-            ChainPosition::Unconfirmed(last_seen) => ChainPosition::Unconfirmed(last_seen),
+            ChainPosition::Confirmed {
+                anchor,
+                transitively,
+            } => ChainPosition::Confirmed {
+                anchor: anchor.clone(),
+                transitively,
+            },
+            ChainPosition::Unconfirmed { last_seen } => ChainPosition::Unconfirmed { last_seen },
         }
     }
 }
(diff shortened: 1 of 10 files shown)


crates/chain/src/chain_data.rs
```diff
@@ -15,25 +15,45 @@ use crate::{Anchor, COINBASE_MATURITY};
...
+        /// If the value is `Some`, it means we have incomplete data. We can only deduce that the
+        /// chain data is confirmed at a block equal to or lower than the block referenced by `A`.
+        transitively: Option<Txid>,
+    },
+    /// The chain data is not confirmed.
+    Unconfirmed {
+        /// When the chain data is last seen in the mempool (if seen at all).
+        ///
+        /// It may be possible that the chain data was last seen as confirmed in a conflicting
+        /// chain.
```
notmandatory: Would this more clearly describe when this value is Some or None? 
```suggestion
        /// If the value is `Some` then this is when the chain data was last seen in a mempool. 
        /// If `None` then it was only seen as confirmed in a conflicting chain.
```

evanlinjin:
```
        /// When the chain data is last seen in the mempool.
        ///
        /// This value will be `None` if the chain data was never seen in the mempool and only seen
        /// in a conflicting chain.
```

How about this?

notmandatory: Yes this is clear and more concise.

crates/chain/src/chain_data.rs
```diff
@@ -15,25 +15,45 @@ use crate::Anchor;
...
-    Unconfirmed(u64),
+    /// The chain data is confirmed as it is anchored in the best chain by `A`.
+    Confirmed {
+        /// The [`Anchor`].
+        anchor: A,
+        /// Whether the chain data is anchored transitively by a child transaction.
+        ///
+        /// If the value is `Some`, it means we have incomplete data. We can only deduce that the
+        /// chain data is confirmed at a block equal to or lower than the block referenced by `A`.
+        transitively: Option<Txid>,
```
ValuedMammal, line 26: ⛏️ I think calling the field either "child" or "descendant" would be best.

ValuedMammal (commented):
Approach ACK

crates/chain/src/chain_data.rs
evanlinjin: I think `transitively` is still more descriptive. If it is `Some`, we know that the tx is "confirmed transitively".

If we name it "child" or "descendant", my first thought when looking at it is; child of what? How does this relate to the anchor?

ValuedMammal: I'm on the fence, @LagginTimes what do you think?

LagginTimes: I think in this naming scheme, `anchor` / `transitively` makes more sense to me. 🤔

crates/chain/src/chain_data.rs
```diff
@@ -73,9 +95,9 @@ impl<A: Anchor> FullTxOut<A> {
     /// [`confirmation_height_upper_bound`]: Anchor::confirmation_height_upper_bound
     pub fn is_mature(&self, tip: u32) -> bool {
         if self.is_on_coinbase {
-            let tx_height = match &self.chain_position {
-                ChainPosition::Confirmed(anchor) => anchor.confirmation_height_upper_bound(),
-                ChainPosition::Unconfirmed(_) => {
+            let tx_height = match self.chain_position.confirmation_height_upper_bound() {
+                Some(conf_height_upper_bound) => conf_height_upper_bound,
```
LagginTimes:
```suggestion
                Some(height) => height,
```

crates/chain/src/chain_data.rs
```diff
@@ -103,17 +125,21 @@ impl<A: Anchor> FullTxOut<A> {
             return false;
         }
 
-        let confirmation_height = match &self.chain_position {
-            ChainPosition::Confirmed(anchor) => anchor.confirmation_height_upper_bound(),
-            ChainPosition::Unconfirmed(_) => return false,
+        let confirmation_height = match self.chain_position.confirmation_height_upper_bound() {
+            Some(h) => h,
```
LagginTimes:
```suggestion
            Some(height) => height,
```

LagginTimes (commented):
Nit: It might be a good idea to adopt the same naming scheme for the sake of consistency.

crates/wallet/src/wallet/mod.rs
```diff
@@ -1037,12 +1037,22 @@ impl Wallet {
...
     ///         "tx is confirmed at height {}, we know this since {}:{} is in the best chain",
     ///         anchor.block_id.height, anchor.block_id.height, anchor.block_id.hash,
     ///     ),
-    ///     ChainPosition::Unconfirmed(last_seen) => println!(
-    ///         "tx is last seen at {}, it is unconfirmed as it is not anchored in the best chain",
+    ///     ChainPosition::Confirmed {
+    ///         anchor,
+    ///         transitively: Some(_),
+    ///     } => println!(
+    ///         "tx is an ancestor of a tx anchored in {}:{}",
```
LagginTimes, line 1051:
```suggestion
    ///         "tx is assumed to be confirmed because it is an ancestor of a tx anchored in {}:{}",
```

LagginTimes (commented):
LGTM, just a documentation suggestion to improve clarity and consistency.

ValuedMammal (approved):
ACK 29b374e383bc9317b55a2a58019e1948e170fabd

Issue #1687, opened 2024-11-13
Benchmark the transaction graph reconciliation


notmandatory:
"Benchmark the transaction graph reconciliation with the canonical chain in the presence of large unconfirmed transaction chains and/or high number of conflicts."

notmandatory:
This should be done before and after #1670 is merged.

ValuedMammal:
This is some data collected around the time of 1.0.0-beta.5 https://hackmd.io/@valuedmammal/rywvTTgQkl

notmandatory:
Fixed in #1670

PR #1735, opened 2024-11-19
WIP: Benchmark `TxGraph` queries

ValuedMammal:
Add `bench` directory to the workspace and begin benchmarking performance of tx-graph queries. I'm currently working on creating benches for these scenarios

- [ ] Finding the canonical set from a large number of tx conflicts
- [ ] Same as above, but further complicated by "nesting" conflicts
- [ ] Conflicts with potentially many descendants

cc #1687 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/.github/workflows/nightly_docs.yml
+++ b/.github/workflows/nightly_docs.yml
@@ -10,7 +10,7 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@v4
       - name: Set default toolchain
-        run: rustup default nightly-2024-05-12
+        run: rustup default nightly-2024-11-17
       - name: Set profile
         run: rustup set profile minimal
       - name: Update toolchain
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-/target
+**/target
 Cargo.lock
 /.vscode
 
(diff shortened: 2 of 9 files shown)


evanlinjin:
Why do we need a separate crate for benchmarks?

ValuedMammal:
Quick status update: I used the bench tests as of a65040847e6ec9816e97c4fd876975ddcaaea3d5 to produce some initial performance numbers, see https://github.com/bitcoindevkit/bdk/issues/1687#issuecomment-2496118577. I ran the same tests on #1670 and the result was such a vast improvement that the current benches in their basic form may not even be very useful post 1.0 but it gives us a starting point. @evanlinjin has also offered to continue work on this PR. Overall it's good to know that ongoing development is backed up by the data.

I also had a question about whether it would be worth having a CI job for benchmarks and the process by which to monitor changes in performance, thoughts?

> Why do we need a separate crate for benchmarks?

In theory you could have several bench tests from different crates run together as a single `criterion_group`, but maybe not necessary for what we need at present. Also it seems that making it a dev-dependency could add to compile times even for normal non-benching development?

evanlinjin:
Replaced by #1670

PR #1737, opened 2024-11-21
fix(tx_builder)!: make TxBuilder Send safe, remove Clone trait

notmandatory:
### Description

Inspired by discord chat with @stevenroose as a way to make the `TxBuilder` Send safe. 

See his original patch on 1.0.0-beta.5: https://gist.github.com/stevenroose/f7736dfedfaa64bbdbb0da5875df28fc

### Notes to the reviewers

I had to remove the `Clone` trait on `TxBuilder` but it was only being used in tests.

### Changelog notice

- TxBuilder is now Send safe and does not implement the Clone trait

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -1195,7 +1195,7 @@ impl Wallet {
     /// [`TxBuilder`]: crate::TxBuilder
     pub fn build_tx(&mut self) -> TxBuilder<'_, DefaultCoinSelectionAlgorithm> {
         TxBuilder {
-            wallet: alloc::rc::Rc::new(core::cell::RefCell::new(self)),
+            wallet: self,
             params: TxParams::default(),
             coin_selection: DefaultCoinSelectionAlgorithm::default(),
         }
@@ -1701,7 +1701,7 @@ impl Wallet {
         };
 
         Ok(TxBuilder {
-            wallet: alloc::rc::Rc::new(core::cell::RefCell::new(self)),
+            wallet: self,
             params,
             coin_selection: DefaultCoinSelectionAlgorithm::default(),
         })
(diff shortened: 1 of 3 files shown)


notmandatory:
@ValuedMammal you were right, I didn't need the `Arc::new(Box::new())` for the test with `custom_bip69_ordering`. I put it back as it was originally and all still working. This is ready to review.

stevenroose:
Do we know if there are users using the `Clone` implementation of `TxBuilder`? I can see how it might be useful to have for some people..

Otherwise ack 663fb133a4aeac7e50f5d32bdba2b316b17a7e2b, changes look good.

Upon further thinking, I think any use of the `Clone` function can be replaced by a small function re-creating the same builder, I guess.

notmandatory:
> Upon further thinking, I think any use of the Clone function can be replaced by a small function re-creating the same builder, I guess.

We should be fine removing `Clone` since I don't think anyone is using it and if anyone is there is the work-around to re-creating the builder with the same params. Plus I have gotten feedback from other projects who want `TxBuilder` to be thread safe so that seems the more valuable thing to do right now.

Long term the goal is to decouple `TxBuilder` from  `Wallet` and at that point it will be much easier to make `TxBuilder` cloneable.

notmandatory:
I moved this back to in-progress, I need to verify bitcoindevkit/bdk-ffi#611 doesn't need `TxBuilder` to impl `Clone`.

evanlinjin (approved):
ACK 663fb133a4aeac7e50f5d32bdba2b316b17a7e2b

This makes a lot of sense. I don't see how this will be a problem for ffi, since the `TxBuilder` struct is only used within the lifetime of the ffi `TxBuilder::finish` method: https://github.com/bitcoindevkit/bdk-ffi/blob/cf5f714bed70cd7ee09f203f55548fb2a2c478f7/bdk-ffi/src/tx_builder.rs#L371-L399

notmandatory:
@evanlinjin thanks for checking this for me. You're right that there's no need for `.clone()` on the `bdk_wallet` version of `TxBuilder` in `bdk-ffi`.

PR #1758, opened 2024-12-05
refactor(wallet): Reuse chain position instead of obtaining new one

### LINKED ISSUE #1677: wallet: Don't call `get_chain_position` redundantly
`chain_position` is a field of `LocalOutput`, so shouldn't this just get the chain position of the utxo rather than call `get_chain_position` again?

https://github.com/bitcoindevkit/bdk/blob/dc0511fcaba0a94e127ca52560e5f10bf18edff9/crates/wallet/src/wallet/mod.rs#L2061-L2068

jirijakes:
In `Wallet::preselect_utxos()`, removes redundant call to `get_chain_position()` and reuses chain position that is already available.

Closes #1677.

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -1996,7 +1996,6 @@ impl Wallet {
         let must_only_use_confirmed_tx = bumping_fee.is_some();
         let must_use_all_available = *drain_wallet;
 
-        let chain_tip = self.chain.tip().block_id();
         //    must_spend <- manually selected utxos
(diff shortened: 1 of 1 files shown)


ValuedMammal:
ACK however if you look at how it was handled in #1670 I'm not entirely sure this would be needed now, what do you think @jirijakes?

jirijakes:
> […] it was handled in #1670 I'm not entirely sure this would be needed now

Ah, of course. I did not notice it was going to be changed there. Thanks!


Closing.

PR #1763, opened 2024-12-08
Change coin_selection and DescriptorExt::dust_value to use Amount type

notmandatory:
### Description

refactor(coin_selection)!: use Amount and SignedAmount for API and internally
refactor(chain)!: use Amount for DescriptorExt::dust_value()

Using named types make the API and internal code easier to read and reason about since it makes it clear that the values are bitcoin amounts. Also to create these types the units (ie .from_sat()) must be specified.

### Notes to the reviewers

For coin_selection using Amount and SignedAmount makes internal code safer against overflow errors. In particular because these types will panic if an amount overflow occurs. Using u64/i64 on the other hand can silently rollover. See: https://doc.rust-lang.org/book/ch03-02-data-types.html#integer-overflow

This is a continuation of the work done in #1595.  Since this is an API breaking change I would like to include it in the 1.0.0-beta milestone if possible.

### Changelog notice

- Change coin_selection to use Amount instead of u64 for all bitcoin amounts.
- Change DescriptorExt::dust_value() to return an Amount instead of u64.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/coin_selection.rs
+++ b/crates/wallet/src/wallet/coin_selection.rs
@@ -41,11 +41,11 @@
 //!         required_utxos: Vec<WeightedUtxo>,
 //!         optional_utxos: Vec<WeightedUtxo>,
 //!         fee_rate: FeeRate,
-//!         target_amount: u64,
+//!         target_amount: Amount,
 //!         drain_script: &Script,
 //!         rand: &mut R,
 //!     ) -> Result<CoinSelectionResult, coin_selection::InsufficientFunds> {
-//!         let mut selected_amount = 0;
+//!         let mut selected_amount = Amount::ZERO;
 //!         let mut additional_weight = Weight::ZERO;
 //!         let all_utxos_selected = required_utxos
 //!             .into_iter()
@@ -53,7 +53,7 @@
 //!             .scan(
 //!                 (&mut selected_amount, &mut additional_weight),
 //!                 |(selected_amount, additional_weight), weighted_utxo| {
-//!                     **selected_amount += weighted_utxo.utxo.txout().value.to_sat();
+//!                     **selected_amount += weighted_utxo.utxo.txout().value;
 //!                     **additional_weight += TxIn::default()
 //!                         .segwit_weight()
 //!                         .checked_add(weighted_utxo.satisfaction_weight)
@@ -62,7 +62,7 @@
 //!                 },
 //!             )
 //!             .collect::<Vec<_>>();
-//!         let additional_fees = (fee_rate * additional_weight).to_sat();
+//!         let additional_fees = fee_rate * additional_weight;
 //!         let amount_needed_with_fees = additional_fees + target_amount;
 //!         if selected_amount < amount_needed_with_fees {
 //!             return Err(coin_selection::InsufficientFunds {
@@ -105,7 +105,7 @@ use crate::chain::collections::HashSet;
 use crate::wallet::utils::IsDust;
 use crate::Utxo;
 use crate::WeightedUtxo;
-use bitcoin::FeeRate;
+use bitcoin::{Amount, FeeRate, SignedAmount};
 
 use alloc::vec::Vec;
 use bitcoin::consensus::encode::serialize;
@@ -127,17 +127,17 @@ pub type DefaultCoinSelectionAlgorithm = BranchAndBoundCoinSelection<SingleRando
 /// This is thrown by [`CoinSelectionAlgorithm`].
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct InsufficientFunds {
-    /// Sats needed for some transaction
-    pub needed: u64,
-    /// Sats available for spending
-    pub available: u64,
+    /// Amount needed for the transaction
+    pub needed: Amount,
+    /// Amount available for spending
+    pub available: Amount,
 }
 
 impl fmt::Display for InsufficientFunds {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         write!(
             f,
-            "Insufficient funds: {} sat available of {} sat needed",
+            "Insufficient funds: {} available of {} needed",
             self.available, self.needed
         )
     }
@@ -152,18 +152,18 @@ pub enum Excess {
     /// It's not possible to create spendable output from excess using the current drain output
     NoChange {
         /// Threshold to consider amount as dust for this particular change script_pubkey
-        dust_threshold: u64,
+        dust_threshold: Amount,
         /// Exceeding amount of current selection over outgoing value and fee costs
-        remaining_amount: u64,
+        remaining_amount: Amount,
         /// The calculated fee for the drain TxOut with the selected script_pubkey
-        change_fee: u64,
+        change_fee: Amount,
     },
     /// It's possible to create spendable output from excess using the current drain output
     Change {
         /// Effective amount available to create change after deducting the change output fee
-        amount: u64,
+        amount: Amount,
         /// The deducted change output fee
-        fee: u64,
+        fee: Amount,
     },
 }
 
@@ -172,24 +172,24 @@ pub enum Excess {
 pub struct CoinSelectionResult {
     /// List of outputs selected for use as inputs
     pub selected: Vec<Utxo>,
-    /// Total fee amount for the selected utxos in satoshis
-    pub fee_amount: u64,
+    /// Total fee amount for the selected utxos
+    pub fee_amount: Amount,
     /// Remaining amount after deducing fees and outgoing outputs
     pub excess: Excess,
 }
 
 impl CoinSelectionResult {
     /// The total value of the inputs selected.
-    pub fn selected_amount(&self) -> u64 {
-        self.selected.iter().map(|u| u.txout().value.to_sat()).sum()
+    pub fn selected_amount(&self) -> Amount {
+        self.selected.iter().map(|u| u.txout().value).sum()
     }
 
     /// The total value of the inputs selected from the local wallet.
-    pub fn local_selected_amount(&self) -> u64 {
+    pub fn local_selected_amount(&self) -> Amount {
         self.selected
             .iter()
             .filter_map(|u| match u {
-                Utxo::Local(_) => Some(u.txout().value.to_sat()),
+                Utxo::Local(_) => Some(u.txout().value),
                 _ => None,
             })
             .sum()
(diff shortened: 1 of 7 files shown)


notmandatory:
Thanks to @thesimplekid for bringing to my attention that native rust integer types can silently roll-over. Even though this is very unlikely to happen, using `Amount`/`SignedAmount` in coin_selection is a good way to make sure it never happens. 

The rest of the code base is already updated to use `Amount` in APIs and internally.

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -297,15 +297,15 @@ impl CoinSelectionAlgorithm for OldestFirstCoinSelection {
...
 /// - `drain_script`: script to consider change creation
-pub fn decide_change(remaining_amount: u64, fee_rate: FeeRate, drain_script: &Script) -> Excess {
+pub fn decide_change(remaining_amount: Amount, fee_rate: FeeRate, drain_script: &Script) -> Excess {
     // drain_output_len = size(len(script_pubkey)) + len(script_pubkey) + size(output_value)
     let drain_output_len = serialize(drain_script).len() + 8usize;
     let change_fee =
-        (fee_rate * Weight::from_vb(drain_output_len as u64).expect("overflow occurred")).to_sat();
-    let drain_val = remaining_amount.saturating_sub(change_fee);
+        fee_rate * Weight::from_vb(drain_output_len as u64).expect("overflow occurred");
+    let drain_val = remaining_amount.checked_sub(change_fee).unwrap_or_default();
```
rustaceanrob, line 305: Is `Default` here `Amount::ZERO`?

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -62,7 +62,7 @@
 //!                 },
 //!             )
 //!             .collect::<Vec<_>>();
-//!         let additional_fees = (fee_rate * additional_weight).to_sat();
+//!         let additional_fees = (fee_rate * additional_weight);
```
ValuedMammal: nit: extra set of round brackets
```suggestion
//!         let additional_fees = fee_rate * additional_weight;
```

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -128,9 +128,9 @@ pub type DefaultCoinSelectionAlgorithm = BranchAndBoundCoinSelection<SingleRando
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct InsufficientFunds {
     /// Sats needed for some transaction
-    pub needed: u64,
+    pub needed: Amount,
     /// Sats available for spending
-    pub available: u64,
+    pub available: Amount,
```
ValuedMammal, line 133: I guess it would be more clear to change the doc to say "Amount needed..." and "Amount available...". We should also update the `Display` impl for `InsufficientFunds` which still assumes the amount is satoshis.

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -219,7 +219,7 @@ pub trait CoinSelectionAlgorithm: core::fmt::Debug {
         required_utxos: Vec<WeightedUtxo>,
         optional_utxos: Vec<WeightedUtxo>,
         fee_rate: FeeRate,
-        target_amount: u64,
+        target_amount: Amount,
```
ValuedMammal, line 222: We should update the doc describing the `target_amount` parameter.

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -322,23 +322,22 @@ pub fn decide_change(remaining_amount: u64, fee_rate: FeeRate, drain_script: &Sc
...
                 if must_use || **selected_amount < target_amount + **fee_amount {
-                    **fee_amount += (fee_rate
+                    **fee_amount += fee_rate
                         * (TxIn::default()
                             .segwit_weight()
                             .checked_add(weighted_utxo.satisfaction_weight)
-                            .expect("`Weight` addition should not cause an integer overflow")))
-                    .to_sat();
-                    **selected_amount += weighted_utxo.utxo.txout().value.to_sat();
+                            .expect("`Weight` addition should not cause an integer overflow"));
```
ValuedMammal: nit: we can remove the extra round brackets starting at `TxIn::default()`. same suggestion on L381.

crates/wallet/src/wallet/coin_selection.rs
notmandatory: Yup default for `Amount` is `Amount::ZERO`.

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -513,7 +518,9 @@ impl<Cs: CoinSelectionAlgorithm> CoinSelectionAlgorithm for BranchAndBoundCoinSe
             // remaining_amount can't be negative as that would mean the
             // selection wasn't successful
             // target_amount = amount_needed + (fee_amount - vin_fees)
-            let remaining_amount = (curr_value - signed_target_amount) as u64;
+            let remaining_amount = (curr_value - signed_target_amount)
+                .to_unsigned()
+                .expect("unsigned amount");
```
ValuedMammal: This sounds like a bit stronger of a reason. same suggestion on L667
```suggestion
                .expect("remaining amount can't be negative");
```

crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -1500,15 +1530,16 @@ mod test {
...
             &drain_script,
             &mut thread_rng(),
         );
 
+        const EXPECTED_AMOUNT: Amount = Amount::from_sat(300_000);
         assert_matches!(
             selection,
             Err(InsufficientFunds {
-                available: 300_000,
+                available: EXPECTED_AMOUNT,
```
ValuedMammal: Can we also use a match guard here?
```suggestion
                available,
                ..
            })
            if available.to_sat() == 300_000
```

crates/wallet/src/wallet/mod.rs
```diff
@@ -1456,12 +1456,12 @@ impl Wallet {
                 required_utxos.clone(),
                 optional_utxos.clone(),
                 fee_rate,
-                outgoing.to_sat() + fee_amount.to_sat(),
+                outgoing + fee_amount,
```
ValuedMammal, line 1449: We can remove the call to `clone` above when passing the required/optional utxos to `coin_select`.

crates/wallet/src/wallet/mod.rs
```diff
@@ -1456,12 +1456,12 @@ impl Wallet {
...
                 &drain_script,
                 rng,
             )
             .map_err(CreateTxError::CoinSelection)?;
-        fee_amount += Amount::from_sat(coin_selection.fee_amount);
+        fee_amount += coin_selection.fee_amount;
```
ValuedMammal, line 1454: `fee_amount` isn't used after this point (it is only assigned to), so we should remove it.

notmandatory:
Just as an FYI, I did a little test to confirm that rust integer underflow does wrap and isn't caught as a static check if the values aren't known at compile time. This only applies when built in release mode, in debug mode it panics.

```rust
#[test]
#[cfg(not(debug_assertions))]
fn test_integer_underflow() {
    use rand::Rng;
    let num = rand::thread_rng().gen_range(11..21);
    let ten: u64 = 10;
    let underflow = ten - num;
    // println!("10 - {} => underflow {}", num, underflow);
    assert!(underflow > 0)
}
```

crates/wallet/src/wallet/mod.rs
notmandatory: I see `fee_amount` is assigned to after L1464 on L1508, L1513, but as you say it isn't used. The code is pretty convoluted but it seems none of the lines that changes `fee_amount` after L1459 are needed. That also seems to be true for the `received` value on L1511. 

I suspect some of this code was added during development for debugging and never removed. I'd rather not make unrelated changes in this PR, do you mind if I create a new issue to research and clean up the `Wallet::create_tx()` logic?

I did a quick test after removing unused code mentioned above and all tests still pass. But want to understand better why it was there in the first place.

crates/wallet/src/wallet/coin_selection.rs
notmandatory: Fixed.

crates/wallet/src/wallet/coin_selection.rs
notmandatory: Fixed.

crates/wallet/src/wallet/coin_selection.rs
notmandatory: Fixed.

crates/wallet/src/wallet/coin_selection.rs
notmandatory: Fixed.

crates/wallet/src/wallet/coin_selection.rs
notmandatory: Fixed.

crates/wallet/src/wallet/coin_selection.rs
notmandatory: Great suggestion, I should have known there was a better way to do this.  Fixed.

crates/wallet/src/wallet/mod.rs
notmandatory: Fixed.

crates/wallet/src/wallet/mod.rs
ValuedMammal: I assumed it was left over from when `create_tx` returned a tuple of `(Psbt, TransactionDetails)`. https://docs.rs/bdk/latest/src/bdk/wallet/mod.rs.html#603

notmandatory: Ah ha, I think you're correct and I'm responsible for leaving it there when I did #1048. I'll create a new PR to clean it up, it doesn't affect any APIs so can wait until post 1.0.0.

notmandatory: See #1768 .

crates/wallet/src/wallet/coin_selection.rs
ValuedMammal: I think the `Display` impl should be changed, because `Amount` will display itself in units of BTC.

```diff
 impl fmt::Display for InsufficientFunds {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         write!(
             f,
-            "Insufficient funds: {} sat available of {} sat needed",
+            "Insufficient funds: {} available of {} needed",
             self.available, self.needed
         )
     }
 }
```

oleonardolima (commented):
utACK d2f49a7e995d5d75b62eeac61722d1b1429b41eb

crates/wallet/src/wallet/coin_selection.rs
notmandatory: Oops sorry I mixed that part, fixed.

oleonardolima (commented):
utACK 7f1e4d2ecddf79ec57e07fb50bdf319794188cd1

ValuedMammal (approved):
ACK 7f1e4d2ecddf79ec57e07fb50bdf319794188cd1

ValuedMammal (approved):
ACK 2a9eeed302a3c7faaa7e2d26ff3d806f2317a850

Issue #1748, opened 2024-11-27
Don't double spend created transaction just because they haven't been broadcasted


LLFourn:
This was a comment from a discussion elsewhere. I don't want to restrict the design space too much but I personally think the right way to deal with this is:

1. Insert the transactions into the tx graph right after they are created (without signatures)
2. Mark the txid as "unbroadcasted" at the tx graph level or application level
3. Assume it's canonical whenever you do canonicalization
4. Once it's signed, add the signatures to the tx graph. This can be done by re-inserting the transaction and adding it to the changeset with the new signatures.
5. When it gets a `last_seen` remove it from the unbraodcasted state.

This makes sense to on the `TxGraph`

---

The user has to decide the fate of their transactions. And I claim each unconfirmed tx is either in a state of:

1. The user has not broadcasted it (maybe it's still being signed) and wants to confirm it eventually
7. It's broadcasted and the user still wants it to be confirmed
8. It's broadcasted but the user wants to replace it
9. It's not broadcasted and the user no longer wants to broadcast it.

We have to consider the state when creating a new tx: in (1,2) we want to treat the tx as canonical regardless of whether it has been seen in the mempool recently or not and therefore it's UTXOs can be candidate inputs. In (2), if it hasn't been seen for a while this will mean we want the user to broadcast it again along with the new child tx.
In state (3) the application has to treat it as non-canonical and double spend it with the replacing tx.  In state (4) it's just non-canonical and remains that way.

I don't think there is a fifth state we have to model where the user is thinking "I broadcasted this tx but you know I don't really care if it gets confirmed. If it hasn't been seen for a while I kinda want to prune it so I might inadvertently double spend it with my next tx...or I might not..whatever". 

There are several ways in which these three tx states are not modeled correctly in BDK so there definitely is a design bug but it seems we are not in agreement to the nature of it. I think states (2) and (4) are easy to express, YMWV with (3) and (1) doesn't exist.

_Originally posted by @LLFourn in https://github.com/bitcoindevkit/bdk/issues/1666#issuecomment-2502098294_

PR #1764, opened 2024-12-08
proof of concept: implement `cancel_tx`

ValuedMammal:
### Description

This is not a merge candidate. The goal of this PR is implement `cancel_tx`, as doing so has the potential to fix a long-standing TODO in the library saying "make this free up reserved utxos when that's implemented". This was initiated in response to user feedback claiming that the method has no effect.

issue bitcoindevkit/bdk_wallet#41 

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/tx_details.rs
+++ b/crates/wallet/src/wallet/tx_details.rs
@@ -1,0 +1,83 @@
+//! `TxDetails`
+
+use alloc::vec::Vec;
+use serde::{Deserialize, Serialize};
+
+use bitcoin::Txid;
+
+use crate::collections::HashMap;
+
+/// Kind of change that can be applied to [`TxDetails`]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum Record {
+    /// tx canceled
+    Canceled,
+    /// new details added. note, when applied this will overwrite existing
+    Details(Details),
+}
+
+/// Communicates changes to [`TxDetails`]
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ChangeSet {
+    /// records
+    pub records: Vec<(Txid, Record)>,
+}
+
+impl bdk_chain::Merge for ChangeSet {
+    fn merge(&mut self, other: Self) {
+        self.records.extend(other.records);
+    }
+
+    fn is_empty(&self) -> bool {
+        self.records.is_empty()
+    }
+}
+
+/// The details of a wallet transaction
+///
+/// In the future this may be extended to include more metadata.
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Details {
+    /// Cancellation status, `true` if the tx was canceled
+    pub is_canceled: bool,
+}
+
+/// Object that stores transaction metadata
+#[derive(Debug, Clone, Default)]
+pub(crate) struct TxDetails {
+    /// maps txid to details
+    pub map: HashMap<Txid, Details>,
+}
+
+impl TxDetails {
+    /// Marks a tx canceled
+    pub fn cancel_tx(&mut self, txid: Txid) -> ChangeSet {
+        let mut change = ChangeSet::default();
+
+        let val = self.map.entry(txid).or_default();
+
+        if val.is_canceled {
+            return change;
+        }
+
+        val.is_canceled = true;
+
+        change.records = vec![(txid, Record::Canceled)];
+
+        change
+    }
+
+    /// Applies the given `changeset` to `self`
+    pub fn apply_changeset(&mut self, changeset: ChangeSet) {
+        for (txid, record) in changeset.records {
+            match record {
+                Record::Details(det) => {
+                    let _ = self.map.insert(txid, det);
+                }
+                Record::Canceled => {
+                    let _ = self.cancel_tx(txid);
+                }
+            }
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


crates/wallet/src/wallet/tx_details.rs
```diff
@@ -0,0 +1,83 @@
...
+use crate::collections::HashMap;
+
+/// Kind of change that can be applied to [`TxDetails`]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum Record {
+    /// tx canceled
+    Canceled,
+    /// new details added. note, when applied this will overwrite existing
+    Details(Details),
+}
```
ErikDeSmedt, line 17: I don't really understand the design of this enum.

What is the difference between `Record::Canceled` and `Record::Details { canceled: true}`.
Why are both structs required?

crates/wallet/tests/wallet.rs
```diff
@@ -44,6 +44,95 @@ const P2WPKH_FAKE_WITNESS_SIZE: usize = 106;
 
 const DB_MAGIC: &[u8] = &[0x21, 0x24, 0x48];
 
+#[test]
+fn cancel_tx_frees_spent_inputs() -> anyhow::Result<()> {
```
ErikDeSmedt, line 48: Nice, this test behaves exactly as I would expect

ErikDeSmedt (commented):
Thanks for looking into this.

I'm not very familiar with the BDK-codebase but I am happy to see where this is going

evanlinjin:
@ValuedMammal could you further elaborate on the goals of this PR? Thanks!

buffrr:
Even after PR bitcoindevkit/bdk#1765, I think `cancel_tx` would still be needed for different reasons. If a transaction gets dropped from the mempool but persisted in the wallet as unconfirmed, we need a way to handle it - otherwise it would stay as unconfirmed in the wallet db forever, right? Users would have no choice but to rescan their wallet (I guess they could try to rebroadcast or bump the fee)

evanlinjin:
@buffrr The only way we can confidently say "a transaction is dropped" is if a conflicting transaction gets confirmed. Therefore, having `cancel_tx` is not the correct fix here. Instead, we should somehow introduce the conflicting tx into our `TxGraph`.

buffrr:
It's possible for a transaction to be dropped from the mempool without a conflicting tx - such as when it has a low fee and stays in the mempool too long. While we can have some level of confidence after seeing a conflicting tx, we also can't be certain that it will be mined over the original. Waiting for a confirmation of a foreign/irrelevant tx might not be desirable as it locks the users funds during that period which could be hours or days.

It would be helpful to give developers the option to explicitly mark a tx as 'ok to double spend' and I guess that's what would `cancel_tx` enable. I'm grateful for https://github.com/bitcoindevkit/bdk/pull/1765 so this is more of a nice to have!

evanlinjin:
Imagine this from a UX perspective. The user creates a tx because they have intent to spend their Bitcoin. Then because feerate too low, it gets dropped from mempool and the tx just disappears from the wallet. I would say this is unexpected behavior. I expect the user would want to RBF the tx?

What if the feerate is acceptable after some time, and someone, all this time, has a mempool that didn't drop txs and this tx gets rebroadcasted? So now do we need a method to uncancel the tx?

There are other ways to explicitly mark a tx as "okay to double spend".

buffrr:
> Then because feerate too low, it gets dropped from mempool and the tx just disappears from the wallet. I would say this is unexpected behavior. I expect the user would want to RBF the tx?

While I agree in that specific case for typical transactions, it really depends on the use case and how developers would want to tailor the UX for end users. 

> What if the feerate is acceptable after some time, and someone, all this time, has a mempool that didn't drop txs and this tx gets rebroadcasted? So now do we need a method to uncancel the tx?

There's no need for such a method as that wouldn't result in locking up funds. The worst that could happen is they end up replacing it, or it gets confirmed, which, depending on the use case, might be ok. 

> There are other ways to explicitly mark a tx as "okay to double spend".

I’m not sure that’s easily achievable. Perhaps only by attempting to manually find and double-spend the locked-up funds?

crates/wallet/src/wallet/tx_details.rs
ValuedMammal: This made it easy to match the kind of change to the specific SQL statement, with the assumption that the real `TxDetails` would have more than a single field and hence be subject to different kinds of changes. Granted it's a bit contrived at the moment. Anyway, appreciate your comments!

ValuedMammal:
Sorry for the late reply, I do have several thoughts

I'm still at a light concept ACK for this feature, because it allows you to abandon an untrusted, unconfirmed tx that has no viability on chain but would nonetheless be stuck pending in the wallet and risk poisoning the local utxo pool with bad outputs. That said, I agree that simply introducing the spending tx should be enough to resolve the conflict internally.

#### To reiterate the rules of `cancel_tx`:

1. Txouts of a canceled tx are ignored, provided the canceled tx is unconfirmed 
2. Unspents are kept if spent by a canceled tx, provided the canceled tx is unconfirmed

#### What if someone rebroadcasts?

- In the event the tx confirms there is no issue because confirmation takes precedence over cancellation, no need to un-cancel
- If we cancel tx1 and then reuse the coins in tx2 and tx2 confirms, then a rebroadcast of tx1 will be rejected by the network.

The Bitcoin Core wallet has an `abandontransaction` RPC and apparently does this on its own for txs that have been stuck for two weeks (`-mempoolexpiry`), so in terms of wallet behavior this isn't particularly groundbreaking.

#### There are a few caveats though:

* With this approach we can't actually mark the change address unused if the tx is indexed internally, so in that sense we could be trading off one weakness for another.
* In terms of implementation I think a more permanent solution could be done in `bdk_chain` where an abandoned tx would be represented by transaction state 4 (#1748). But I'm not opposed to a solution similar to this PR if there's enough interest in further developing `TxDetails`.

I agree that in the absence of this feature, we will likely need better solutions for RBF and CPFP.

evanlinjin:
@ValuedMammal Thank you for clarifying. This now makes sense for me.

PR #1765, opened 2024-12-09
fix(chain): detect incoming transaction being replaced/canceled

### LINKED ISSUE #1740: Undetected double-spent
**Describe the bug**  

1. Alice has an empty wallet

**To Reproduce**  

**Expected behavior**  

**Build environment**  
 - OS+version: debian

**Additional context**  

LagginTimes:
Fixes #1740.

### Description

This PR addresses a potential double-spending issue by implementing the following changes:

#### Redefine Transaction Relevancy in `IndexedTxGraph`
Transactions are now considered relevant if they:
* Reference the wallet's spks in their inputs/outputs.
* Share inputs with any transaction already in the `TxGraph`.

#### Include Unconfirmed Spends in Chain Sync Requests
* Unconfirmed outputs can now be provided via `SyncRequestExt::unconfirmed_spends`, which enables the chain source to detect replaced or canceled transactions.
* Deprecated `Wallet::start_sync_with_revealed_spks` in favor of `start_sync`, which supports unconfirmed spends.

### Notes to the reviewers

TODO:
* [ ] Add esplora test.

### Changelog notice

* Redefine what a relevant transaction is in `IndexedTxGraph`.
* Providing unconfirmed spends in chain sync requests, enabling detection of replaced or canceled transactions.
* Deprecated `Wallet::start_sync_with_revealed_spks` in favor of the more comprehensive `Wallet::start_sync`.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/indexed_tx_graph.rs
+++ b/crates/chain/src/indexed_tx_graph.rs
@@ -1,5 +1,6 @@
 //! Contains the [`IndexedTxGraph`] and associated types. Refer to the
 //! [`IndexedTxGraph`] documentation for more.
+use crate::collections::HashSet;
 use core::fmt::Debug;
 
 use alloc::{sync::Arc, vec::Vec};
@@ -178,7 +179,7 @@ where
 
         let mut tx_graph = tx_graph::ChangeSet::default();
         for (tx, anchors) in txs {
-            if self.index.is_tx_relevant(&tx) {
+            if self.index.is_tx_relevant(&tx) || self.graph.direct_conflicts(&tx).next().is_some() {
                 let txid = tx.compute_txid();
                 tx_graph.merge(self.graph.insert_tx(tx.clone()));
                 for anchor in anchors {
@@ -218,10 +219,37 @@ where
             indexer.merge(self.index.index_tx(tx));
         }
 
+        // It is possible that a spk-relevant transaction can be replaced with a non-spk-relevant
+        // transaction. Furthermore, the existence of the replacement transaction may mean that the
+        // original spk-relevant transaction is evicted. Therefore, we need to include the new
+        // non-spk-relevant transaction in the graph.
+        //
+        // `relevant_spends` keeps tracks of outpoints that are associated with spk-relevant
+        // transactions in `unconfirmed_txs`. This allows us to detect non-spk-relevant replacements
+        // when both the original and the replacement are introduced together.
+        let mut relevant_spends = HashSet::<OutPoint>::new();
+
+        let mut indexer_changeset = I::ChangeSet::default();
+        for (tx, _) in &txs {
+            indexer_changeset.merge(self.index.index_tx(tx));
+
+            if self.index.is_tx_relevant(tx) && !tx.is_coinbase() {
+                relevant_spends.extend(tx.input.iter().map(|txin| txin.previous_output));
+            }
+        }
+
         let graph = self.graph.batch_insert_unconfirmed(
             txs.into_iter()
-                .filter(|(tx, _)| self.index.is_tx_relevant(tx))
-                .map(|(tx, seen_at)| (tx.clone(), seen_at)),
+                .filter(|(tx, _)| {
+                    self.index.is_tx_relevant(tx)
+                        || tx
+                            .input
+                            .iter()
+                            .any(|txin| relevant_spends.contains(&txin.previous_output))
+                        || self.graph.direct_conflicts(tx).next().is_some()
+                })
+                .map(|(tx, seen_at)| (tx.clone(), seen_at))
+                .collect::<Vec<_>>(),
         );
 
         ChangeSet {
@@ -278,7 +306,8 @@ where
         let mut changeset = ChangeSet::<A, I::ChangeSet>::default();
         for (tx_pos, tx) in block.txdata.iter().enumerate() {
             changeset.indexer.merge(self.index.index_tx(tx));
-            if self.index.is_tx_relevant(tx) {
+
+            if self.index.is_tx_relevant(tx) || self.graph.direct_conflicts(tx).next().is_some() {
                 let txid = tx.compute_txid();
                 let anchor = TxPosInBlock {
                     block,
--- a/crates/chain/tests/test_indexed_tx_graph.rs
+++ b/crates/chain/tests/test_indexed_tx_graph.rs
@@ -3,12 +3,16 @@
 #[macro_use]
 mod common;
 
-use std::{collections::BTreeSet, sync::Arc};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    sync::Arc,
+};
 
 use bdk_chain::{
     indexed_tx_graph::{self, IndexedTxGraph},
     indexer::keychain_txout::KeychainTxOutIndex,
     local_chain::LocalChain,
+    spk_txout::SpkTxOutIndex,
     tx_graph, Balance, ChainPosition, ConfirmationBlockTime, DescriptorExt,
 };
 use bdk_testenv::{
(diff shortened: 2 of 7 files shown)


buffrr (commented):
This behaves as expected! However, I think including the conflicting transaction in the wallet's Tx graph may create unnecessary noise/confusion. Since these conflicts don't represent actual wallet activity. 

Perhaps irrelevant transactions could be filtered out either directly in `list_canonical_txs` or in the wallet crate. Alternatively, marking the existing wallet tx as conflicting without inserting the external conflicting transactions into the graph.

evanlinjin:
> Perhaps irrelevant transactions could be filtered out either directly in list_canonical_txs or in the wallet crate. Alternatively, marking the existing wallet tx as conflicting without inserting the external conflicting transactions into the graph.

That's a good point. It's a one line change to have `Wallet::transactions` filter out *irrelevant* txs. For `bdk_chain`, I think having docs should be sufficient?

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -99,6 +99,97 @@ fn insert_relevant_txs() {
     assert_eq!(graph.initial_changeset(), initial_changeset);
 }
 
+/// Ensure that [`IndexedTxGraph::batch_insert_relevant`] adds transactions that are direct
+/// conflicts with transactions in our graph but are not directly relevant to it.
```
evanlinjin: Nit

```suggestion
/// Ensure that [`IndexedTxGraph::batch_insert_relevant`] adds transactions that are direct
/// conflicts with transactions in our graph but are not directly tracked by our indexer.
```

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -99,6 +99,97 @@ fn insert_relevant_txs() {
...
+///
+/// The graph contains three transactions (A, B, and conflicting_tx), with A and B being relevant
+/// because B spends from A. This test verifies that `conflicting_tx` is inserted into the graph
+/// solely because it is directly conflicting with B.
```
evanlinjin: > with A and B being relevant because B spends from A

do you mean: A is tracked by the indexer and B conflicts with A so A is relevant?

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -99,6 +99,97 @@ fn insert_relevant_txs() {
...
+    let changeset = graph.batch_insert_relevant(txs.iter().cloned().map(|tx| (tx, None)));
+
+    // Confirm that `conflicting_tx` is added to `ChangeSet` only due to it being a conflicting
+    // transaction, and not because it is directly relevant to our graph.
+    assert!(!graph.index.is_tx_relevant(&conflicting_tx));
+    assert!(graph
+        .graph()
+        .direct_conflicts(&conflicting_tx)
+        .next()
+        .is_some());
```
evanlinjin: I don't think this is needed, we can see it by just looking at the txs.

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -99,6 +99,97 @@ fn insert_relevant_txs() {
...
 }
 
+/// Ensure that [`IndexedTxGraph::batch_insert_relevant`] adds transactions that are direct
+/// conflicts with transactions in our graph but are not directly relevant to it.
+///
+/// The graph contains three transactions (A, B, and conflicting_tx), with A and B being relevant
+/// because B spends from A. This test verifies that `conflicting_tx` is inserted into the graph
+/// solely because it is directly conflicting with B.
+#[test]
+fn insert_relevant_conflicting_txs() {
```
evanlinjin, line 109: I think we have better clarity with these tests. I propose having multiple test cases.

### For all these test cases, reuse the following 3 transactions

```
sender_prev_tx ---> send_tx
               \__> undo_send_tx
```

* `sender_prev_tx` has 1 input and 1 output. It is not relevant to the receiver's wallet.
* `send_tx` is the initial tx that was meant to pay the receiver. It has an input spending `sender_prev_tx` and an output which is tracked by the receiver's indexer.
* `undo_send_tx` replaces `send_tx` and sends the tx back to the sender.

### For all these test cases, have more meaningful variable names and data types

* 1 spk for the sender and 1 spk for the receiver. Name them `sender_spk` and `receiver_spk`.
* Use `SpkTxOutIndex` instead of `KeychainTxOutIndex`. Derivation is not important for the tests.

### Test case 1

* Given: An empty receiver `IndexedTxGraph`.
* When: `batch_insert_relevant_unconfirmed` is called with `[send_tx, undo_send_tx]` or `[undo_send_tx, send_tx]`.
* Assert: Both txs are included in the changeset no matter which order we input the txs.

### Test case 2

* Given: A receiver `IndexedTxGraph` which already contains `send_tx`.
* When: `batch_insert_relevant_unconfirmed` is called with `undo_send_tx`.
* Assert: `undo_send_tx` is included in the returned changeset.
* Variations: Try with `batch_insert_relevant`.

evanlinjin (changes requested):
ConceptACK

I'm requesting the following changes:
* Better clarity with the tests.
* Update docs of the methods that are changed (since we changed the definition of what is relevant).

I expect a follow-up PR which tests against a block-by-block chain source (`bdk_bitcoind_rpc`). I think this test can exist in `bdk_bitcoind_rpc`.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -218,11 +218,13 @@ where
...
-                .filter(|(tx, _)| self.index.is_tx_relevant(tx))
-                .map(|(tx, seen_at)| (tx.clone(), seen_at)),
-        );
+        let mut relevant_txs = Vec::new();
+        for (tx, seen_at) in txs.into_iter() {
+            if self.index.is_tx_relevant(&tx) || self.graph.direct_conflicts(&tx).next().is_some() {
+                relevant_txs.push((tx.clone(), seen_at));
+            }
+        }
+        let graph = self.graph.batch_insert_unconfirmed(relevant_txs);
```
evanlinjin: What was wrong with the iterator?

evanlinjin:
@LagginTimes I rebased and forced push. The last commit I added provides unconfirmed spends to spk-based chain sources. This should be enough to fully fix #1740 but we need to test.

I expect a `detect_receive_tx_cancel` test for all chain sources (defined in their respective chain source crates). Also the PR description/title needs to be updated.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -218,9 +218,16 @@ where
...
+
         let graph = self.graph.batch_insert_unconfirmed(
             txs.into_iter()
-                .filter(|(tx, _)| self.index.is_tx_relevant(tx))
+                .filter(|(tx, _)| {
+                    (self.index.is_tx_relevant(tx))
+                        || temp_graph.direct_conflicts(tx).next().is_some()
+                })
                 .map(|(tx, seen_at)| (tx.clone(), seen_at)),
         );
```
evanlinjin:
```rust
        let graph = self.graph.batch_insert_unconfirmed(
            txs.into_iter()
                .filter(|(tx, _)| {
                    (self.index.is_tx_relevant(tx))
                        || self.graph.direct_conflicts(tx).next().is_some()
                })
                .map(|(tx, seen_at)| (tx.clone(), seen_at))
                .collect::<Vec<_>>(),
        );
```

evanlinjin: If we collect before calling `batch_insert_unconfirmed`, we don't have to hold the immutable reference across it.

evanlinjin:
The title needs to highlight what is being fixed, not what is changed to do the fix.

Suggestion: Detect incoming transaction being replaced/canceled.

crates/chain/src/indexed_tx_graph.rs
LagginTimes: The `temp_graph` is used here to accurately detect conflicts. Without it, the filtering in this statement happens before any transactions are added to `self.graph`, causing `self.graph.direct_conflicts()` to always return false.

evanlinjin: @LagginTimes You'll need to include this in the comment since it's not immediately obvious.

> causing self.graph.direct_conflicts() to always return false.

I don't think this is true in every case. If the indexed tx (the tx which has an output which we own) already exists in the graph, then this will work properly.

I think what you want to say is: This allows us to detect replacement txs which reverts sending to a tracked spk in the situation where the original tx and the replacement tx is added in the same call to `batch_insert_relevant_unconfirmed`.

However, the logic is not fully sound here. Imagine if `batch_insert_relevant_unconfirmed` is called with two conflicting txs which both don't spend or spend to a tracked spk. We expect that this method should filter both these txs out.

evanlinjin:
```rust
        // It is possible that a spk-relevant transaction can be replaced with a non-spk-relevant
        // transaction. Furthermore, the exsistance of the replacement transaction may mean that the
        // original spk-relevant transaction is evicted. Therefore, we need to include the new
        // non-spk-relevant transaction in the graph.
        //
        // `relevant_spends` keeps tracks of outpoints that spk-relevant transactions in
        // `unconfirmed_txs`. This allows us to detect non-spk-relevant replacements when both the
        // original and the replacement are introduced together.
        let mut relevant_spends = HashSet::<OutPoint>::new();

        let mut indexer_changeset = I::ChangeSet::default();
        for (tx, _) in &txs {
            indexer_changeset.merge(self.index.index_tx(tx));

            if self.index.is_tx_relevant(tx) && !tx.is_coinbase() {
                relevant_spends.extend(tx.input.iter().map(|txin| txin.previous_output));
            }
        }

        let tx_graph_changeset = self.graph.batch_insert_unconfirmed(
            txs.into_iter()
                .filter(|(tx, _)| {
                    self.index.is_tx_relevant(tx)
                        || self.graph.direct_conflicts(tx).next().is_some()
                        || tx
                            .input
                            .iter()
                            .any(|txin| relevant_spends.contains(&txin.previous_output))
                })
                .map(|(tx, seen_at)| (tx.clone(), seen_at))
                .collect::<Vec<_>>(),
        );
```

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -1,6 +1,7 @@
 //! Contains the [`IndexedTxGraph`] and associated types. Refer to the
 //! [`IndexedTxGraph`] documentation for more.
 use core::fmt::Debug;
+use std::collections::HashSet;
```
evanlinjin:
```suggestion
use crate::collections::HashSet;
```

ValuedMammal:
Since we can reliably determine the "unconfirmed outpoints" of relevant txs, would it not suffice to have a method `apply_block_relevant_spends` that applies both relevant txs along with any spends given a list of semi-relevant outpoints. This wouldn't change the definition of relevant and the other batch-insert methods could continue to do what they say.

evanlinjin:
After reviewing comments from @buffrr and @ValuedMammal, and learning about perspectives from @LLFourn and Yuval Kogman, I realized the approach in this PR is wrong.

We failed to take into consideration that the malicious replacement can be replacing an ancestor/parent of the incoming transaction. To fix the problem with our current approach of including the conflicting transaction(s), we will need to traverse all unconfirmed ancestors (and find their conflicts). This is obviously a DoS vector and very bandwidth-intensive in general.

The better solution would be to introduce a `MissingMarker` which informs about the latest timestamp that a transaction was deemed "missing" from the mempool.

When requesting a sync from spk-based chain sources, we also include a list of "expected txids" per spk. Anything missing from the list will result in a `MissingMarker` being added to the `SyncResponse`.

For block-based chain sources, can send the txid list of unconfirmed to it and it can return those that are missing from mempool.

LagginTimes:
Closed, replaced by #1811.

PR #1777, opened 2024-12-16
feat(wallet): add check for network and genesis_hash consistency

### LINKED ISSUE #46: Verify network consistency between chain genesis and descriptors
"Consistency between descriptor network and chain genesis hash isn't checked in load_with_params nor in create_with_params."

### LINKED ISSUE #1716: Verify network consistency between chain genesis and descriptors
"Consistency between descriptor network and chain genesis hash isn't checked in load_with_params nor in create_with_params."

oleonardolima:
fixes bitcoindevkit/bdk_wallet#46

### Description

Adds a new check on both `create_with_params` and `load_with_params` to validate that the given `genesis_hash` parameter matches the `genesis_hash` for expected `Network`, returns a new `DescriptorError::Mismatch` or `LoadError::LoadMismatch` error variants, respectively.

### Notes to the reviewers

I'm not sure if adding the check on `load_with_params` is needed, and if that branch will ever be reached, at least it seems not trivial to test it using the standard of tests we currently have, that its: 1) creating a wallet; 2) loading and asserting the expected checks.

### Changelog notice
- Adds a new check for consistency of `genesis_hash` and `network` parameters.
- Adds a new error `MismatchError` enum and `Descriptor::Mismatch` variant.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -59,7 +59,6 @@ pub mod signer;
 pub mod tx_builder;
 pub(crate) mod utils;
 
-use crate::collections::{BTreeMap, HashMap, HashSet};
 use crate::descriptor::{
     check_wallet_descriptor, error::Error as DescriptorError, policy::BuildSatisfaction,
     DerivedDescriptor, DescriptorMeta, ExtendedDescriptor, ExtractPolicy, IntoWalletDescriptor,
@@ -73,6 +72,10 @@ use crate::wallet::{
     signer::{SignOptions, SignerError, SignerOrdering, SignersContainer, TransactionSigner},
     tx_builder::{FeePolicy, TxBuilder, TxParams},
     utils::{check_nsequence_rbf, After, Older, SecpCtx},
+};
+use crate::{
+    collections::{BTreeMap, HashMap, HashSet},
+    descriptor::error::MismatchError,
 };
 
 // re-exports
@@ -372,6 +375,14 @@ impl Wallet {
         let genesis_hash = params
             .genesis_hash
             .unwrap_or(genesis_block(network).block_hash());
+
+        if genesis_hash.ne(&genesis_block(network).block_hash()) {
+            return Err(DescriptorError::Mismatch(MismatchError::Genesis {
+                network: genesis_block(network).block_hash(),
+                parameter: genesis_hash,
+            }));
+        }
+
         let (chain, chain_changeset) = LocalChain::from_genesis_hash(genesis_hash);
 
         let (descriptor, mut descriptor_keymap) = (params.descriptor)(&secp, network)?;
@@ -503,6 +514,14 @@ impl Wallet {
             if chain.genesis_hash() != exp_genesis_hash {
                 return Err(LoadError::Mismatch(LoadMismatch::Genesis {
                     loaded: chain.genesis_hash(),
+                    expected: exp_genesis_hash,
+                }));
+            }
+
+            let network_genesis_hash = genesis_block(network).block_hash();
+            if network_genesis_hash != exp_genesis_hash {
+                return Err(LoadError::Mismatch(LoadMismatch::Genesis {
+                    loaded: network_genesis_hash,
                     expected: exp_genesis_hash,
                 }));
             }
--- a/crates/wallet/src/descriptor/error.rs
+++ b/crates/wallet/src/descriptor/error.rs
@@ -43,6 +43,8 @@ pub enum Error {
     Hex(bitcoin::hex::HexToBytesError),
     /// The provided wallet descriptors are identical
     ExternalAndInternalAreTheSame,
+    /// The provided parameters mismatch
+    Mismatch(MismatchError),
 }
 
 impl From<crate::keys::KeyError> for Error {
@@ -83,6 +85,7 @@ impl fmt::Display for Error {
             Self::ExternalAndInternalAreTheSame => {
                 write!(f, "External and internal descriptors are the same")
             }
+            Self::Mismatch(mismatch) => write!(f, "The provided parameters mismatch: {mismatch:?}"),
         }
     }
 }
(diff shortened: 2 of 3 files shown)


crates/wallet/src/wallet/mod.rs
```diff
@@ -506,6 +517,14 @@ impl Wallet {
...
                 }));
             }
+
+            let network_genesis_hash = genesis_block(network).block_hash();
+            if network_genesis_hash != exp_genesis_hash {
+                return Err(LoadError::Mismatch(LoadMismatch::Genesis {
+                    loaded: network_genesis_hash,
+                    expected: exp_genesis_hash,
+                }));
+            }
```
oleonardolima, line 527: I'm still unsure of how to properly test this one (or if it's even needed), because I can't initially create a wallet with `network` and `genesis_hash` that are valid (consistent) on `create_with_params`, but it will fail with an inconsistent `check_genesis_hash` on `load_with_params`.

It'll always fail on the previous check.

ValuedMammal:
I had previously understood that bdk allowed you to use a custom genesis hash, meaning one that doesn't match the chain hash of the specified network. I interpreted the issue as saying the network isn't checked to be among the valid networks of a descriptor, but as you said that logic exists in `into_wallet_descriptor`.

crates/wallet/src/wallet/mod.rs
luisschwab: Would it be possible to `create_with_params()` with one set of params and then `load_with_params()` with another? This could trigger the second test, no? Or is this scenario captured by another test?

crates/wallet/src/wallet/mod.rs
```diff
@@ -372,6 +375,14 @@ impl Wallet {
...
             .genesis_hash
             .unwrap_or(genesis_block(network).block_hash());
+
+        if genesis_hash.ne(&genesis_block(network).block_hash()) {
+            return Err(DescriptorError::Mismatch(MismatchError::Genesis {
+                network: genesis_block(network).block_hash(),
+                parameter: genesis_hash,
+            }));
+        }
+
```
evanlinjin, line 385: As mentioned by @ValuedMammal, the intention here is so the caller can provide a custom genesis hash.

However, if the intention is to enforce a 1-possible-genesis-hash-per-network-variant invariance, changing the data structure to disallow the invariance would make a lot more sense. I.e. remove `params.genesis_hash`.

evanlinjin (commented):
Approach-NACK, Concept-NACK

It seems the intention of the PR is to enforce a 1-possible-genesis-hash-per-network-variant invariance. The correct way to do this is to change the datastructure, not enforce the invariance ad-hoc with errors.

So now the question is whether we want to be able have the caller provide a custom genesis hash. I think we still should, but maybe our approach needs to be a bit different.

It is true that the `Network` enum is intended to represent a single network per variant so the current code is hard to reason with. To fix this, maybe we can introduce a new enum:

```rust
pub enum BdkNetwork {
    Network(bitcoin::Network),
    Custom(BlockHash),
}

impl BdkNetwork {
    /// Handy to work with keys.
    ///
    /// The `Custom` variant will always be `Text`.
    pub fn kind(&self) -> bitcoin::NetworkKind {}
}
```

oleonardolima:
> Approach-NACK, Concept-NACK
> 
> It seems the intention of the PR is to enforce a 1-possible-genesis-hash-per-network-variant invariance. The correct way to do this is to change the datastructure, not enforce the invariance ad-hoc with errors.

Agree, if we opt-out from allowing custom genesis_hash/networks, we can just remove the parameter altogether.

> So now the question is whether we want to be able have the caller provide a custom genesis hash. I think we still should, but maybe our approach needs to be a bit different.

Yes, I agree with also allowing a custom genesis_hash. I didn't have that in mind when working on the changes. However, by the initial issue description, I reasoned that the current implementation is too loose and would allow a user to use a custom Genesis hash on Mainnet and Testnet.

> 
> It is true that the `Network` enum is intended to represent a single network per variant so the current code is hard to reason with. To fix this, maybe we can introduce a new enum:
> 
> ```rust
> pub enum BdkNetwork {
>     Network(bitcoin::Network),
>     Custom(BlockHash),
> }
> 
> impl BdkNetwork {
>     /// Handy to work with keys.
>     ///
>     /// The `Custom` variant will always be `Text`.
>     pub fn kind(&self) -> bitcoin::NetworkKind {}
> }
> ```

Thanks, this seems like a better approach.

oleonardolima:
As discussed with VM, another possible approach would be to:

1. only check for consistency only in `create_with_params`.
2. check for consistency only for mainnet network, that said only networks from the variant `bitcoin::NetworkKind::Main`, returning a proper error otherwise (we currently don't have any suitable variants in current `DescriptorError`).

It would still allow the usage of a custom `genesis_hash` by the user in any other scenario (e.g. Signet/Testnet/Regtest networks) but without the need to introduce a new `BdkNetwork` type, relying on rust-bitcoin's one. (@ValuedMammal please correct me if I missed something)

However, any of both approaches: stated above or introducing a new type `BdkNetwork`, are breaking changes API-wise, that said it's not applicable for any near-term release, and would only be suitable to the 2.0 one.

Although the current API allows some inadequate usage/inconsistency as the code snippet below, it seems a better approach to put this PR aside now and return to it later.

Any other thoughts?


``` rust
#[test]
fn test_create_genesis_hash_and_network_inconsistency() {
    let (external, internal) = get_test_tr_single_sig_xprv_and_change_desc();
    let mainnet_genesis_hash = BlockHash::from_byte_array(ChainHash::BITCOIN.to_bytes());

    let _ = Wallet::create(external, internal)
        .genesis_hash(mainnet_genesis_hash)
        .network(Network::Regtest)
        .create_wallet_no_persist()
        .unwrap();
}
```

ValuedMammal:
I would be in favor of something like this, that is bdk should only stop you from trying to customize the mainnet genesis hash.

> `create_with_params`
```rs
    let genesis_hash = match params.genesis_hash {
        Some(hash) => {
            // only permit custom genesis hash for NetworkKind::Test
            if NetworkKind::from(network).is_mainnet()
                && ChainHash::from_genesis_block_hash(hash) != ChainHash::BITCOIN
            {
                // TODO: Error: "genesis hash does not match network kind"
            }
            hash
        }
        None => genesis_block(network).block_hash(),
    };
```

As you said we need to figure out the exact error to return. I guess it would be prudent to have a similar validation in `load_with_params` too.

evanlinjin:
@oleonardolima This needs to be moved to `bdk_wallet`.

PR #1779, opened 2024-12-17
`transactions` method should only return relevant transactions

### LINKED ISSUE #1239: The fact that the `Wallet.transactions()` API can return more than wallet-related txs is not explicit
One might expect that the `Wallet.transactions()` method replaced the `Wallet.list_transactions()` and will return similar transactions (the transactions related to the descriptors the wallet tracks). This is not the case, as the method returns an iterator that traverses the entire TxGraph. If I understand this TxGraph correctly, any number of transactions can be added to it, even conflicting and never mined txs. The API docs are simply: `Iterate over the transactions in the wallet.` 

For typical use, a wallet will likely want to see transactions related to itself, and I assume the user is expected to then filter the transactions in the iterator using the `is_mine()` method. Let me know if I misunderstand the intended use of the API!

But if the above is correct, I find the API more cumbersome than it needs to be for the most typical uses and what one might expect of something like a `Wallet.transactions()` method. Here are a few thoughts I have to maybe make the API simpler/more explicit. Would love to hear what others think:

1. The `transactions()` method could take an argument `mine_only: bool` that would ensure the method returns only wallet-related txs, or maybe the opposite, an argument `full_tx_graph: bool` that would return what it currently returns.
2. The `transactions()` method could be renamed entirely to imply its more complex nature, and a new, simpler `transactions()` method could take its place and return a more filtered list of transactions that relate to the wallet.

Let me know if there is more to this that I haven't understood well!

evanlinjin:
Fixes #1239

### Description

Currently the behavior of `Wallet::transactions` is not well defined and unintuitive.

The approach taken in this PR is to make `Wallet::transactions` return what most wallets would like the caller to see (i.e. transactions that are part of the canonical history and spend from/to a tracked spk). A.k.a make the method more restrictive.

Documentation is updated to refer the caller to the underlying `bdk_chain` structures for any over usecase.

After #1765 gets merged, the behavior of `Wallet::transactions` will become even more unintuitive. Refer to https://github.com/bitcoindevkit/bdk/pull/1765#pullrequestreview-2503968540. 

### Notes to the reviewers

**Why not have multiple methods in `Wallet` that return different sets of transactions?**

I think it's better to only provide common usecase histories from `Wallet` and I can only think of one.

### Changelog notice

* Change `Wallet::transactions` to only include "relevant" transactions.

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -19,6 +19,7 @@ use alloc::{
     sync::Arc,
     vec::Vec,
 };
+use chain::Indexer;
 use core::{cmp::Ordering, fmt, mem, ops::Deref};
 
 use bdk_chain::{
@@ -1062,14 +1063,30 @@ impl Wallet {
             .find(|tx| tx.tx_node.txid == txid)
     }
 
-    /// Iterate over the transactions in the wallet.
+    /// Iterate over relevant and canonical transactions in the wallet.
+    ///
+    /// A transaction is relevant when it spends from or spends to at least one tracked output. A
+    /// transaction is canonical when it is confirmed in the best chain, or does not conflict
+    /// with any transaction confirmed in the best chain.
+    ///
+    /// To iterate over all transactions, including those that are irrelevant and not canonical, use
+    /// [`TxGraph::full_txs`].
+    ///
+    /// To iterate over all canonical transactions, including those that are irrelevant, use
+    /// [`TxGraph::list_canonical_txs`].
     pub fn transactions(&self) -> impl Iterator<Item = WalletTx> + '_ {
-        self.indexed_graph
-            .graph()
+        let tx_graph = self.indexed_graph.graph();
+        let tx_index = &self.indexed_graph.index;
+        tx_graph
             .list_canonical_txs(&self.chain, self.chain.tip().block_id())
+            .filter(|c_tx| tx_index.is_tx_relevant(&c_tx.tx_node.tx))
     }
 
-    /// Array of transactions in the wallet sorted with a comparator function.
+    /// Array of relevant and canonical transactions in the wallet sorted with a comparator
+    /// function.
+    ///
+    /// This is a helper method equivalent to collecting the result of [`Wallet::transactions`]
+    /// into a [`Vec`] and then sorting it.
     ///
     /// # Example
     ///
--- a/crates/wallet/tests/wallet.rs
+++ b/crates/wallet/tests/wallet.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 
 use anyhow::Context;
 use assert_matches::assert_matches;
-use bdk_chain::{BlockId, ChainPosition, ConfirmationBlockTime};
+use bdk_chain::{BlockId, ChainPosition, ConfirmationBlockTime, TxUpdate};
 use bdk_wallet::coin_selection::{self, LargestFirstCoinSelection};
 use bdk_wallet::descriptor::{calc_checksum, DescriptorError, IntoWalletDescriptor};
 use bdk_wallet::error::CreateTxError;
@@ -12,7 +12,7 @@ use bdk_wallet::psbt::PsbtUtils;
 use bdk_wallet::signer::{SignOptions, SignerError};
 use bdk_wallet::test_utils::*;
 use bdk_wallet::tx_builder::AddForeignUtxoError;
-use bdk_wallet::{AddressInfo, Balance, ChangeSet, Wallet, WalletPersister, WalletTx};
+use bdk_wallet::{AddressInfo, Balance, ChangeSet, Update, Wallet, WalletPersister, WalletTx};
 use bdk_wallet::{KeychainKind, LoadError, LoadMismatch, LoadWithPersistError};
 use bitcoin::constants::{ChainHash, COINBASE_MATURITY};
 use bitcoin::hashes::Hash;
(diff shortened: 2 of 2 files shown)


crates/wallet/src/wallet/mod.rs
```diff
@@ -1062,14 +1063,30 @@ impl Wallet {
             .find(|tx| tx.tx_node.txid == txid)
     }
 
-    /// Iterate over the transactions in the wallet.
+    /// Iterate over relevant and canonical transactions in the wallet.
+    ///
+    /// Transactions are relevant when it spends from or spends to at least one tracked ouput.
```
notmandatory: ⛏️ grammar nit:
```suggestion
    /// Transactions are relevant when they spend from or spend to at least one tracked output.
```

evanlinjin: I did:
```rust
    /// Iterate over relevant and canonical transactions in the wallet.
    ///
    /// A transaction is relevant when it spends from or spends to at least one tracked ouput. A
    /// transactions is canonical when it is confirmed in the best chain, or does not conflict
    /// with any transaction confirmed in the best chain.
    ///
    /// To iterate over all transactions, including those that are irrelevant and not canonical, use
    /// [`TxGraph::full_txs`].
    ///
    /// To iterate over all canonical transactions, including those that are irrelevant, use
    /// [`TxGraph::list_canonical_txs`].
```

oleonardolima (commented):
utACK 3fe2bb0e53b1ac720c08d30617fc69aeec8ee6a3

thunderbiscuit:
Concept ACK! Reminds me of #1239, and should probably close it once it gets in.

_Edit: just noticed Evan you'd already linked the issue up. Cheers!_

notmandatory:
I added 75fae3ef56be838787306fc81862eb5ae500f635 to verify only relevant transactions are returned. If there's a better way to test this feel free to replace or change this commit. 

I also re-based on master to pickup the CI fixes in #1776 and fixed typos that @oleonardolima found.

notmandatory (approved):
tACK 75fae3ef56be838787306fc81862eb5ae500f635

crates/wallet/src/wallet/mod.rs
```diff
@@ -1062,14 +1063,30 @@ impl Wallet {
             .find(|tx| tx.tx_node.txid == txid)
     }
 
-    /// Iterate over the transactions in the wallet.
+    /// Iterate over relevant and canonical transactions in the wallet.
+    ///
+    /// A transaction is relevant when it spends from or spends to at least one tracked ouput. A
```
oleonardolima:
``` suggestion
    /// A transaction is relevant when it spends from or spends to at least one tracked output. A
```

notmandatory: Oh good catch, I should have caught this, fixed.

oleonardolima:
> [...] If there's a better way to test this feel free to replace or change this commit.

I wonder if another thorough testing approach would be explicitly creating a new tx sending from A -> B, RBF it, and see that it's not being returned as relevant now, or that would still be relevant ? Does that make sense, or am I missing a point here ? (I guess same is being achieved in a simpler approach)

oleonardolima (commented):
tACK 75fae3ef56be838787306fc81862eb5ae500f635

luisschwab:
>  oleonardolima: wonder if another thorough testing approach would be explicitly creating a new tx sending from A -> B, RBF it, and see that it's not being returned as relevant now, or that would still be relevant ?

IMO, a transaction `T` and it's replacement `T'` are both relevant, until `T'` get's confirmed to the chain, then only `T'` is relevant. That's what I expect after reading this doc:

>  /// Iterate over relevant and canonical transactions in the wallet.
    ///
   /// A transaction is relevant when it spends from or spends to at least one tracked output. A
   /// transaction is canonical when it is confirmed in the best chain, or does not conflict
   /// with any transaction confirmed in the best chain.

notmandatory:
> I wonder if another thorough testing approach would be explicitly creating a new tx sending from A -> B, RBF it, and see that it's not being returned as relevant now, or that would still be relevant ? Does that make sense, or am I missing a point here ? (I guess same is being achieved in a simpler approach)

The scenario I had in mind is when we have a new version of electrum or esplora that also retrieves transactions that have double spent a utxo that we received in an unconfirmed tx. Those "foreign" tx don't have any scripts that are relevant to our wallet but still need to be in our wallet to invalidate our unconfirmed (double spent) tx. 

My test doesn't repro all the details of this scenario but at least shows that a not relevant tx can be added to the tx_graph and will be excluded from `Wallet::transactions`.

notmandatory:
> IMO, a transaction `T` and it's replacement `T'` are both relevant, until `T'` get's confirmed to the chain, then only `T'` is relevant. That's what I expect after reading this doc:

In this RBF scenario I think `T` and `T'` will both still be relevant since they both spend to or from a tracked SPK. But after `T'` is confirmed `T` is no longer canonical since it is not confirmed in the best chain.

luisschwab:
Is TxGraph monotonic? If so I agree that `T` remains relevant and `T'` becomes cannonical.

luisschwab:
tACK 75fae3ef56be838787306fc81862eb5ae500f635

Would be good to add a test fot the RBF scenario as well.

crates/wallet/src/wallet/mod.rs
```diff
@@ -1062,14 +1063,30 @@ impl Wallet {
...
+    ///
+    /// A transaction is relevant when it spends from or spends to at least one tracked output. A
+    /// transaction is canonical when it is confirmed in the best chain, or does not conflict
+    /// with any transaction confirmed in the best chain.
+    ///
+    /// To iterate over all transactions, including those that are irrelevant and not canonical, use
+    /// [`TxGraph::full_txs`].
+    ///
+    /// To iterate over all canonical transactions, including those that are irrelevant, use
+    /// [`TxGraph::list_canonical_txs`].
```
ValuedMammal, line 1076: 💯 This does a good job of outlining the distinction between relevant, canonical, and everything else.

crates/wallet/tests/wallet.rs
```diff
@@ -4239,3 +4239,52 @@ fn test_tx_builder_is_send_safe() {
...
+        txs: vec![other_tx_node.tx],
+        txouts: Default::default(),
+        anchors: [(*other_tx_confirmationblocktime, other_txid)].into(),
+        seen_ats: [(other_txid, other_tx_confirmationblocktime.confirmation_time)].into(),
+    };
+    let test_wallet_update = Update {
+        last_active_indices: Default::default(),
+        tx_update: other_tx_update,
+        chain: None,
+    };
```
ValuedMammal, line 4269: This can be simplified by getting a `TxUpdate` directly from the other tx graph.

```diff
-    let other_tx_node = other_wallet.get_tx(other_txid).unwrap().tx_node;
-    let other_tx_confirmationblocktime = other_tx_node.anchors.iter().last().unwrap();
-    let other_tx_update = TxUpdate {
-        txs: vec![other_tx_node.tx],
-        txouts: Default::default(),
-        anchors: [(*other_tx_confirmationblocktime, other_txid)].into(),
-        seen_ats: [(other_txid, other_tx_confirmationblocktime.confirmation_time)].into(),
-    };
     let test_wallet_update = Update {
-        last_active_indices: Default::default(),
-        tx_update: other_tx_update,
-        chain: None,
+        tx_update: other_wallet.tx_graph().clone().into(),
+        ..Default::default()
     };
```

crates/wallet/tests/wallet.rs
```diff
@@ -4239,3 +4239,52 @@ fn test_tx_builder_is_send_safe() {
...
+    test_wallet.apply_update(test_wallet_update).unwrap();
+
+    // verify transaction from other wallet was added but is not it relevant transactions list.
```
ValuedMammal, line 4272:
```suggestion
    // verify transaction from other wallet was added but is not in relevant transactions list.
```

crates/wallet/src/wallet/mod.rs
```diff
@@ -19,6 +19,7 @@ use alloc::{
     sync::Arc,
     vec::Vec,
 };
+use chain::Indexer;
```
ValuedMammal, line 22: ⛏️ This should go under `use bdk_chain::`

ValuedMammal (approved):
ACK 75fae3ef

oleonardolima:
> [...] Those "foreign" tx don't have any scripts that are relevant to our wallet but still need to be in our wallet to invalidate our unconfirmed (double spent) tx.
> 
> My test doesn't repro all the details of this scenario but at least shows that a not relevant tx can be added to the tx_graph and will be excluded from `Wallet::transactions`.

I had this scenario in mind when writing RBF (e.g RBF'ed by/to an external wallet/spk). Thanks, it's clearer now.

crates/wallet/src/wallet/mod.rs
notmandatory: Good suggestion, fixed in #1785.

crates/wallet/tests/wallet.rs
notmandatory: Good suggestion, fixed in #1785.

crates/wallet/tests/wallet.rs
notmandatory: Good suggestion, fixed in #1785.

PR #1789, opened 2025-01-02
Change default tx to version 2

### LINKED ISSUE #1676: Wallet should use transaction version 2 when creating transactions
"You create transactions with nVersion set to 1 if no CSV. Nowadays >85% of transactions on the network are version 2, i think it would make you stick out. It's likely the <15% of version 1 transactions are created by old wallet. It's likely that BDK would be used with newer output types. If these two assumptions hold, it would make it possible to identify a BDK transaction with near certainty. Version 2 is also necessary if you want to eventually implement BIP 326 nSequence-based anti-fee-sniping."

benalleng:
### Description

Changes the default version number to version 2 to enchance privacy for the type of wallet used 

Fixes #1676

### Changelog notice

- Changed the default transaction version number in the transaction builder.
- Updated the test to test for the default transaction number.

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -1284,13 +1284,12 @@ impl Wallet {
             external_requirements.merge(&internal_requirements.unwrap_or_default())?;
 
         let version = match params.version {
-            Some(tx_builder::Version(0)) => return Err(CreateTxError::Version0),
-            Some(tx_builder::Version(1)) if requirements.csv.is_some() => {
+            Some(transaction::Version(0)) => return Err(CreateTxError::Version0),
+            Some(transaction::Version::ONE) if requirements.csv.is_some() => {
                 return Err(CreateTxError::Version1Csv)
             }
-            Some(tx_builder::Version(x)) => x,
-            None if requirements.csv.is_some() => 2,
-            None => 1,
+            Some(v) => v,
+            None => transaction::Version::TWO,
         };
 
         // We use a match here instead of a unwrap_or_else as it's way more readable :)
@@ -1388,7 +1387,7 @@ impl Wallet {
         };
 
         let mut tx = Transaction {
-            version: transaction::Version::non_standard(version),
+            version,
             lock_time,
             input: vec![],
             output: vec![],
@@ -1693,7 +1692,7 @@ impl Wallet {
 
         let params = TxParams {
             // TODO: figure out what rbf option should be?
-            version: Some(tx_builder::Version(tx.version.0)),
+            version: Some(tx.version),
             recipients: tx
                 .output
                 .into_iter()
@@ -2581,7 +2580,7 @@ macro_rules! doctest_wallet {
             .unwrap();
         let address = wallet.peek_address(KeychainKind::External, 0).address;
         let tx = Transaction {
-            version: transaction::Version::ONE,
+            version: transaction::Version::TWO,
             lock_time: absolute::LockTime::ZERO,
             input: vec![],
             output: vec![TxOut {
--- a/crates/wallet/src/wallet/tx_builder.rs
+++ b/crates/wallet/src/wallet/tx_builder.rs
@@ -44,8 +44,8 @@ use alloc::sync::Arc;
 use bitcoin::psbt::{self, Psbt};
 use bitcoin::script::PushBytes;
 use bitcoin::{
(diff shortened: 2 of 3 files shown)


benalleng:
Should any of the tests be updated to use a version 2 tx after setting it as a new default?

ValuedMammal:
> Should any of the tests be updated to use a version 2 tx after setting it as a new default?

You could possibly add an assertion to one of the existing tests in `tests/wallet.rs` to check that transactions we create are using the new default.

crates/wallet/src/wallet/mod.rs
```diff
@@ -2581,7 +2580,7 @@ macro_rules! doctest_wallet {
             .unwrap();
         let address = wallet.peek_address(KeychainKind::External, 0).address;
         let tx = Transaction {
-            version: transaction::Version::ONE,
+            version: transaction::Version::TWO,
```
benalleng, line 2583: This does not create any doctest failures, though this is expected as there aren't any doctests that look for versioning, figure it is a nice thing to have to keep these tests consistent with the actual `tx_builder`

ValuedMammal (approved):
Thanks @benalleng overall this looks good to me. My only comment is I think some additional cleanups can be done like using `bitcoin::transaction::Version` everywhere internally instead of this roundabout way that we're doing now. But I'm ok with pushing that to a followup since this PR effectively resolves the current issue.

https://github.com/bitcoindevkit/bdk/blob/4a3675fc8729dab76619d882f46264033492eed3/crates/wallet/src/wallet/mod.rs#L1390-L1395

benalleng:
Just to confirm you had meant removing the version default in the tx_builder completely in favor of the bitcoin::transaction::Version?

crates/wallet/src/wallet/mod.rs
```diff
@@ -1693,7 +1692,7 @@ impl Wallet {
 
         let params = TxParams {
             // TODO: figure out what rbf option should be?
-            version: Some(tx_builder::Version(tx.version.0)),
+            version: Some(transaction::Version::non_standard(tx.version.0)),
```
ValuedMammal: nit:
```suggestion
            version: Some(tx.version),
```

crates/wallet/src/wallet/mod.rs
```diff
@@ -1284,13 +1284,12 @@ impl Wallet {
...
-            Some(tx_builder::Version(1)) if requirements.csv.is_some() => {
+            Some(transaction::Version(0)) => return Err(CreateTxError::Version0),
+            Some(transaction::Version(1)) if requirements.csv.is_some() => {
                 return Err(CreateTxError::Version1Csv)
             }
-            Some(tx_builder::Version(x)) => x,
-            None if requirements.csv.is_some() => 2,
-            None => 1,
+            Some(transaction::Version(x)) => x,
+            None => 2,
```
ValuedMammal: nit: Can we initialize a `Version` here (not just an integer) to pass directly to the tx?

```diff
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -1288,8 +1288,8 @@ impl Wallet {
             Some(transaction::Version(1)) if requirements.csv.is_some() => {
                 return Err(CreateTxError::Version1Csv)
             }
-            Some(transaction::Version(x)) => x,
-            None => 2,
+            Some(v) => v,
+            None => transaction::Version::TWO,
         };
 
         // We use a match here instead of a unwrap_or_else as it's way more readable :)
@@ -1387,7 +1387,7 @@ impl Wallet {
         };
 
         let mut tx = Transaction {
-            version: transaction::Version::non_standard(version),
+            version,
             lock_time,
             input: vec![],
             output: vec![],
```

notmandatory:
Overall this looks good, but should be squashed into one or at most two commits with ["conventional commit"](https://www.conventionalcommits.org/en/v1.0.0/) style messages. Something like:

```
fix(tx_builder): change default transaction version to 2

This change improves privacy since >85% of transactions on the network are version 2. Version 2 is also necessary to eventually implement BIP 326 nSequence-based anti-fee-sniping.
```

notmandatory (approved):
ACK 2219b7cd55bfb8023e9b426de0ef118b32bfa372

benalleng:
Thanks for holding my hand through this one!

PR #1790, opened 2025-01-05
fix(wallet): improve safety on finaize psbt

### LINKED ISSUE #1711: Improve safety of finalize_psbt
"In finalize_psbt, the code at lines 1910-1911 assumes a PSBT input corresponds to the inner transaction input. This only holds because it was checked 50 lines before (L1862). It'd be more careful to use get(n).ok_or(...)? to access the input around line 1910 too, in case the check 50 lines earlier was removed or updated. This is the kind of bug that could slip through review as it'd affect code outside the diff, and could realistically cause an issue as a PSBT is often externally provided: it can be reasonably anticipated to be turned into a remote crasher for some applications."

f3r10:
fix #1711 
### Description

This PR replaces how currently potentially bug-prone PSBT's input access is done using a more carefully approach. 

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -1879,8 +1879,11 @@ impl Wallet {
                         Ok(_) => {
                             // Set the UTXO fields, final script_sig and witness
                             // and clear everything else.
-                            let original = mem::take(&mut psbt.inputs[n]);
-                            let psbt_input = &mut psbt.inputs[n];
+                            let psbt_input = psbt
+                                .inputs
+                                .get_mut(n)
+                                .ok_or(SignerError::InputIndexOutOfRange)?;
+                            let original = mem::take(psbt_input);
                             psbt_input.non_witness_utxo = original.non_witness_utxo;
                             psbt_input.witness_utxo = original.witness_utxo;
                             if !tmp_input.script_sig.is_empty() {


crates/wallet/src/wallet/mod.rs
```diff
@@ -1879,8 +1879,11 @@ impl Wallet {
                         Ok(_) => {
                             // Set the UTXO fields, final script_sig and witness
                             // and clear everything else.
-                            let original = mem::take(&mut psbt.inputs[n]);
-                            let psbt_input = &mut psbt.inputs[n];
```
rustaceanrob, line 1883: Interesting here that `psbt_input` is assigned to the `n`th value after `take` has already been called. The result should be `psbt_input` is assigned to `Input::default`, which seems like a bug to me in the old code. Curious what you think @ValuedMammal ?

rustaceanrob:
ACK e8a96389bac78961220cbc095c820a551ce4dacf

crates/wallet/src/wallet/mod.rs
ValuedMammal: I think it's right before and after.

evanlinjin: @rustaceanrob it's not a bug. `mem::take` replaces the original with a default value (in this case, it'll be an empty `psbt::Input`). In essence, this block of code will clear the `psbt::Input` and fill it in with only things that we need.

evanlinjin (approved):
ACK e8a96389bac78961220cbc095c820a551ce4dacf

PR #1792, opened 2025-01-07
fix(chain): sanitize derivation index before apply changeset

### LINKED ISSUE #60: 'apply_changeset' the derivation index aren't sanitized
"In apply_changeset the derivation index aren't sanitized. It could include a hardened index, which would break an invariant and trigger an internal crash."

f3r10:
fix: bitcoindevkit/bdk_wallet#60 

### Description

Right now, ff a descriptor has hardened child steps when inserting a descriptor throws a panic. This PR adds a previous descriptor sanitization controlling such cases.

An additional problem pointed out by @ValuedMammal is that when the apply_changest is applied a malicious crafter index could be greater that `BIP32_MAX_INDEX`. This PR adds a verification to handle such cases.
#### All Submissions:

### DIFF
--- a/crates/chain/src/indexer/keychain_txout.rs
+++ b/crates/chain/src/indexer/keychain_txout.rs
@@ -793,7 +793,12 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     pub fn apply_changeset(&mut self, changeset: ChangeSet) {
         for (&desc_id, &index) in &changeset.last_revealed {
             let v = self.last_revealed.entry(desc_id).or_default();
-            *v = index.max(*v);
+            let sanitized_index = if index > BIP32_MAX_INDEX {
+                BIP32_MAX_INDEX
+            } else {
+                index
+            };
+            *v = sanitized_index.max(*v);
             self.replenish_inner_index_did(desc_id, self.lookahead);
         }
     }
--- a/crates/chain/src/indexer/keychain_txout.rs
+++ b/crates/chain/src/indexer/keychain_txout.rs
@@ -10,10 +10,16 @@ use crate::{
     DescriptorExt, DescriptorId, Indexed, Indexer, KeychainIndexed, SpkIterator,
 };
 use alloc::{borrow::ToOwned, vec::Vec};
-use bitcoin::{Amount, OutPoint, ScriptBuf, SignedAmount, Transaction, TxOut, Txid};
+use bitcoin::{
+    bip32::ChildNumber, Amount, OutPoint, ScriptBuf, SignedAmount, Transaction, TxOut, Txid,
+};
 use core::{
     fmt::Debug,
     ops::{Bound, RangeBounds},
+};
+use miniscript::{
+    descriptor::{DescriptorXKey, Wildcard},
+    ForEachKey,
 };
 
 use crate::Merge;
@@ -355,6 +361,23 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         keychain: K,
         descriptor: Descriptor<DescriptorPublicKey>,
     ) -> Result<bool, InsertDescriptorError<K>> {
+        // Ensure the keys don't contain any hardened derivation steps or hardened wildcards
+        let descriptor_contains_hardened_steps = descriptor.for_any_key(|k| {
+            if let DescriptorPublicKey::XPub(DescriptorXKey {
+                derivation_path,
+                wildcard,
+                ..
+            }) = k
+            {
+                return *wildcard == Wildcard::Hardened
+                    || derivation_path.into_iter().any(ChildNumber::is_hardened);
+            }
+            false
+        });
+        if descriptor_contains_hardened_steps {
+            return Err(InsertDescriptorError::HardenedDerivationXpub);
+        }
+        descriptor.sanity_check()?;
         let did = descriptor.descriptor_id();
         if !self.keychain_to_descriptor_id.contains_key(&keychain)
             && !self.descriptor_id_to_keychain.contains_key(&did)
@@ -776,7 +799,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     }
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Debug, PartialEq)]
 /// Error returned from [`KeychainTxOutIndex::insert_descriptor`]
 pub enum InsertDescriptorError<K> {
     /// The descriptor has already been assigned to a keychain so you can't assign it to another
@@ -793,6 +816,16 @@ pub enum InsertDescriptorError<K> {
         /// The descriptor that the keychain is already assigned to
         existing_assignment: Descriptor<DescriptorPublicKey>,
     },
+    /// Miniscript error
+    Miniscript(miniscript::Error),
+    /// The descriptor contains hardened derivation steps on public extended keys
+    HardenedDerivationXpub,
+}
+
+impl<K> From<miniscript::Error> for InsertDescriptorError<K> {
+    fn from(err: miniscript::Error) -> Self {
+        InsertDescriptorError::Miniscript(err)
+    }
 }
 
 impl<K: core::fmt::Debug> core::fmt::Display for InsertDescriptorError<K> {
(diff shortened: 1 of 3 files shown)


f3r10:
@notmandatory I am also working on this PR but is not been assigned to me yet.

ValuedMammal:
If this helps, it appears that `insert_descriptor` can cause a panic in miniscript if passed a descriptor with hardened steps.

```rust
#[test]
fn insert_descriptor_should_reject_hardened_steps() {
    use bdk_chain::keychain_txout::KeychainTxOutIndex;

    // This descriptor has hardened child steps
    let s = "wpkh([e273fe42/84h/1h/0h/0h]tpubDEBY7DLZ5kK6pMC2qdtVvKpe6CiAmVDx1SiLtLS3V4GAGyRvsuVbCYReJ9oQz1rfMapghJyUAYLqkqJ3Xadp3GSTCtdAFcKPgzAXC1hzz8a/*h)";
    let (desc, _) =
        <Descriptor<DescriptorPublicKey>>::parse_descriptor(&Secp256k1::new(), s).unwrap();

    // FIXME: This will panic by attempting to derive a script pubkey from a hardened step
    let mut indexer = KeychainTxOutIndex::<&str>::new(10);
    let _ = indexer.insert_descriptor("keychain", desc);
}
```

f3r10:
> it appears that `insert_descriptor` can cause a panic in miniscript if passed a descriptor with hardened steps.

yes thanks @ValuedMammal  I did a similar test :sweat_smile: It panics because of  ```desc.descriptor_id()``` :thinking: 

So I am going to add the sanitize part in the `insert_descriptor` function trying to handle that error. WDYT?

ValuedMammal:
I think it makes sense to handle the error in `insert_descriptor`, but also I thought about it more and I guess we could have a test showing that `KeychainTxOutIndex::apply_changeset` can handle a maliciously crafted changeset e.g. where the index is >`BIP32_MAX_INDEX`, although I'm fairly certain that `SpkIterator` always caps the range at this value.

f3r10:
> I think it makes sense to handle the error in `insert_descriptor`, but also I thought about it more and I guess we could have a test showing that `KeychainTxOutIndex::apply_changeset` can handle a maliciously crafted changeset e.g. where the index is >`BIP32_MAX_INDEX`, although I'm fairly certain that `SpkIterator` always caps the range at this value.

I tried to create such a test but it takes a lot of time to complete ```.. has been running for over 60 seconds ``` :thinking: 

```rust
#[test]
fn insert_descriptor_maliciously() {
    use bdk_chain::keychain_txout::KeychainTxOutIndex;

    let s = "wpkh([e273fe42/84h/1h/0h/0h]tpubDEBY7DLZ5kK6pMC2qdtVvKpe6CiAmVDx1SiLtLS3V4GAGyRvsuVbCYReJ9oQz1rfMapghJyUAYLqkqJ3Xadp3GSTCtdAFcKPgzAXC1hzz8a/*)";
    let (desc, _) =
        <Descriptor<DescriptorPublicKey>>::parse_descriptor(&Secp256k1::new(), s).unwrap();

     let changeset: ChangeSet = ChangeSet {
        last_revealed: [(desc.descriptor_id(), bdk_chain::BIP32_MAX_INDEX + 1)].into(),
    };

    let mut indexer_a = KeychainTxOutIndex::<TestKeychain>::new(10);
    indexer_a
        .insert_descriptor(TestKeychain::External, desc.clone())
        .expect("must insert keychain");
    indexer_a
        .apply_changeset(changeset.clone())
        .expect("must apply changeset");

}
```

f3r10:
@ValuedMammal just to be sure, the sanitize check has to be only made in `insert_descriptor` and not in `KeychainTxOutIndex::apply_changeset`, right?

evanlinjin (changes requested):
Approach NACK

The thing with changesets is that they are designed to be loaded from persistence and ideally be fully monotone in nature. Erroring when applying the changeset is against this design principle.

I think the correct solution would be to modify the `ChangeSet::last_revealed` values in `KeychainTxOutIndex::apply_changeset` to be maxed at 2^31-1.

Whether the descriptor has any hardened steps in the derivation path should be checked in `insert_descriptor` (which already returns an error).

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -767,18 +790,22 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     }
 
     /// Applies the `ChangeSet<K>` to the [`KeychainTxOutIndex<K>`]
-    pub fn apply_changeset(&mut self, changeset: ChangeSet) {
+    pub fn apply_changeset(&mut self, changeset: ChangeSet) -> Result<(), InsertDescriptorError> {
         for (&desc_id, &index) in &changeset.last_revealed {
             let v = self.last_revealed.entry(desc_id).or_default();
+            if index > BIP32_MAX_INDEX {
+                return Err(InsertDescriptorError::OutOfBounds);
+            }
```
ValuedMammal: 🔧  This comment https://github.com/bitcoindevkit/bdk/pull/1792#pullrequestreview-2541613102 is saying, and I agree, that in principle `apply_changeset` should not have to return a `Result`.

notmandatory:
I moved the state of this PR back to "In Progress" and took it out of the 1.1 release since https://github.com/bitcoindevkit/bdk/pull/1792#pullrequestreview-2541613102 still needs to be addressed.

crates/chain/src/indexer/keychain_txout.rs
f3r10: ok I did not understand the comment at first :sweat: 
I just added a simple conditional that checks if any of the last_revealed values is greater than `BIP32_MAX_INDEX` if it is then that value would be limit to `BIP32_MAX_INDEX` WDYT @ValuedMammal ?

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -355,6 +361,23 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+            }) = k
+            {
+                return *wildcard == Wildcard::Hardened
+                    || derivation_path.into_iter().any(ChildNumber::is_hardened);
+            }
+            false
+        });
+        if descriptor_contains_hardened_steps {
+            return Err(InsertDescriptorError::HardenedDerivationXpub);
+        }
```
ValuedMammal, line 379: What do you think about reusing the miniscript `ConversionError` here?

```rust
  // Ensure the keys don't contain any hardened derivation steps
  let _ = descriptor
      .derived_descriptor(&Secp256k1::verification_only(), 0)
      .map_err(InsertDescriptorError::Conversion)?;
```

crates/chain/tests/test_keychain_txout_index.rs
```diff
@@ -600,6 +600,19 @@ fn lookahead_to_target() {
...
+fn insert_descriptor_should_reject_hardened_steps() {
+    use bdk_chain::keychain_txout::KeychainTxOutIndex;
+
+    // This descriptor has hardened child steps
+    let s = "wpkh([e273fe42/84h/1h/0h/0h]tpubDEBY7DLZ5kK6pMC2qdtVvKpe6CiAmVDx1SiLtLS3V4GAGyRvsuVbCYReJ9oQz1rfMapghJyUAYLqkqJ3Xadp3GSTCtdAFcKPgzAXC1hzz8a/*h)";
+    let (desc, _) =
+        <Descriptor<DescriptorPublicKey>>::parse_descriptor(&Secp256k1::new(), s).unwrap();
+
+    let mut indexer = KeychainTxOutIndex::<&str>::new(10);
+    assert!(indexer.insert_descriptor("keychain", desc).is_err())
```
ValuedMammal, line 613: Ideally we should assert that the returned error matches a specific variant of `InsertDescriptorError`.

crates/chain/src/indexer/keychain_txout.rs
f3r10: oh!! you are right I am going to do that :+1:

notmandatory:
Since we've moved `bdk_wallet` into it's own repo (#1928) you'll need to remove the `bdk_wallet` change from this PR and put it in it's own PR in the new `bdk_wallet` repo. I recommend doing that after this the `bdk_chain` parts of this PR are approved. Thanks!

evanlinjin:
@f3r10 Let us know if you would like someone to take over.

evanlinjin:
Stale.

PR #1799, opened 2025-01-13
feat(wallet): add `TxBuilder::replace_tx`

ValuedMammal:
This PR adds a new method `TxBuilder::replace_tx` that offers a simple way of creating replacements without relying on the mechanics of `build_fee_bump`. See #1374 for context and motivating discussion.

In summary, there are a few limitations with the current `build_fee_bump`:

1. It doesn't prevent the wallet from accidentally selecting the outputs of the replaced tx as inputs to the replacement
2. The UX is somewhat rigid because it assumes all of the same inputs and outputs should appear in the replacement. (however despite this second point we somehow neglect to reuse the original drain script?)

In contrast `replace_tx` is useful when we only need to create a conflict and flexible enough to further tweak the parameters as desired.

### Notes to the reviewers

- `TxBuilder::add_utxos` is modified to look for potentially spent outputs, provided that the given outpoint exists in the replaceable set.

### Changelog notice

- Added method `TxBuilder::replace_tx`
- Added method `TxBuilder::previous_fee`

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/wallet/src/wallet/tx_builder.rs
+++ b/crates/wallet/src/wallet/tx_builder.rs
@@ -138,6 +138,7 @@ pub(crate) struct TxParams {
     pub(crate) add_global_xpubs: bool,
     pub(crate) include_output_redeem_witness_script: bool,
     pub(crate) bumping_fee: Option<PreviousFee>,
+    pub(crate) replacing: HashSet<OutPoint>,
     pub(crate) current_height: Option<absolute::LockTime>,
     pub(crate) allow_dust: bool,
 }
@@ -281,17 +282,17 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
             .collect::<HashMap<_, _>>();
         let utxo_batch = outpoints
             .iter()
-            .map(|outpoint| {
+            .map(|&outpoint| -> Result<_, AddUtxoError> {
                 let output = outputs
-                    .get(outpoint)
+                    .get(&outpoint)
                     .cloned()
-                    .ok_or(AddUtxoError::UnknownUtxo(*outpoint))?;
-                // the output should be unspent unless we're doing a RBF
-                if self.params.bumping_fee.is_none() && output.is_spent {
-                    return Err(AddUtxoError::UnknownUtxo(*outpoint));
+                    .ok_or(AddUtxoError::UnknownUtxo(outpoint))?;
+                // The output must be unspent unless the outpoint opts-in to replacement
+                if output.is_spent && !self.params.replacing.contains(&outpoint) {
+                    return Err(AddUtxoError::UnknownUtxo(outpoint));
                 }
                 Ok((
-                    *outpoint,
+                    outpoint,
                     WeightedUtxo {
                         satisfaction_weight: self
                             .wallet
@@ -393,14 +394,22 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
             .map(|(op, _)| op)
             .ok_or(ReplaceTxError::NonReplaceable)?;
 
-        // add previous fee
-        let absolute = self.wallet.calculate_fee(&tx).unwrap_or_default();
-        let rate = absolute / tx.weight();
-        self.params.bumping_fee = Some(PreviousFee { absolute, rate });
+        // Inform the builder that all of the previous outputs *may* be replaced
+        // before adding our chosen outpoint.
+        self.params
+            .replacing
+            .extend(tx.input.iter().map(|txin| txin.previous_output));
 
-        self.add_utxo(outpoint).expect("we must have the utxo");
+        self.add_utxo(outpoint)
+            .expect("we have the utxo and opted to replace it");
 
-        // do not allow spending outputs descending from the replaced tx
+        // Add the previous fee
+        if let Ok(absolute) = self.wallet.calculate_fee(&tx) {
+            let rate = absolute / tx.weight();
+            self.params.bumping_fee = Some(PreviousFee { absolute, rate });
+        }
+
+        // Do not allow spending outputs descending from the replaced tx
         core::iter::once((txid, tx))
             .chain(
                 self.wallet
@@ -1520,16 +1529,17 @@ mod test {
         // now replace tx1 with a new transaction
         let mut builder = wallet.build_tx();
         builder.replace_tx(txid1).expect("should replace input");
+
+        // `replace_tx` should allow replacing any of the remaining inputs
+        // of the original tx
+        assert_eq!(builder.params.replacing, [outpoint_1, outpoint_2].into());
+        builder.add_utxo(outpoint_2).expect("should add output");
+
         let prev_feerate = builder.previous_fee().unwrap().1;
         builder.add_recipient(recip, Amount::from_sat(98_500));
         builder.fee_rate(FeeRate::from_sat_per_kwu(
             prev_feerate.to_sat_per_kwu() + 250,
         ));
-
-        // Because outpoint 2 was spent in tx1, by default it won't be available for selection,
-        // but we can add it manually, with the caveat that the builder is in a bump-fee
-        // context.
-        builder.add_utxo(outpoint_2).expect("should add output");
         let psbt = builder.finish().unwrap();
 
         assert!(psbt
--- a/crates/wallet/src/wallet/tx_builder.rs
+++ b/crates/wallet/src/wallet/tx_builder.rs
@@ -274,27 +274,36 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
     /// These have priority over the "unspendable" utxos, meaning that if a utxo is present both in
     /// the "utxos" and the "unspendable" list, it will be spent.
     pub fn add_utxos(&mut self, outpoints: &[OutPoint]) -> Result<&mut Self, AddUtxoError> {
+        let outputs = self
+            .wallet
+            .list_output()
+            .map(|out| (out.outpoint, out))
+            .collect::<HashMap<_, _>>();
         let utxo_batch = outpoints
             .iter()
             .map(|outpoint| {
-                self.wallet
-                    .get_utxo(*outpoint)
-                    .ok_or(AddUtxoError::UnknownUtxo(*outpoint))
-                    .map(|output| {
-                        (
-                            *outpoint,
-                            WeightedUtxo {
-                                satisfaction_weight: self
-                                    .wallet
-                                    .public_descriptor(output.keychain)
-                                    .max_weight_to_satisfy()
-                                    .unwrap(),
-                                utxo: Utxo::Local(output),
-                            },
-                        )
-                    })
+                let output = outputs
+                    .get(outpoint)
+                    .cloned()
+                    .ok_or(AddUtxoError::UnknownUtxo(*outpoint))?;
+                // the output should be unspent unless we're doing a RBF
+                if self.params.bumping_fee.is_none() && output.is_spent {
+                    return Err(AddUtxoError::UnknownUtxo(*outpoint));
+                }
+                Ok((
+                    *outpoint,
+                    WeightedUtxo {
+                        satisfaction_weight: self
+                            .wallet
+                            .public_descriptor(output.keychain)
+                            .max_weight_to_satisfy()
+                            .unwrap(),
+                        utxo: Utxo::Local(output),
+                    },
+                ))
             })
             .collect::<Result<HashMap<OutPoint, WeightedUtxo>, AddUtxoError>>()?;
+
         self.params.utxos.extend(utxo_batch);
 
         Ok(self)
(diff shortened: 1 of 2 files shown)


crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -306,6 +310,106 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
+            .max_by_key(|(_, value)| *value)
+            .map(|(op, _)| op)
+            .ok_or(ReplaceTxError::NonReplaceable)?;
+
+        // add previous fee
+        if let Ok(absolute) = self.wallet.calculate_fee(&tx) {
+            let rate = absolute / tx.weight();
+            let previous_fee = PreviousFee { absolute, rate };
+            self.params.bumping_fee = Some(previous_fee);
+        }
```
evanlinjin: Should we still allow replacement if we can't determine the previous tx's fee/feerate?

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -306,6 +310,106 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
+
+        self.add_utxo(outpoint).map_err(|e| match e {
+            AddUtxoError::UnknownUtxo(op) => ReplaceTxError::MissingOutput(op),
+        })?;
+
+        // do not try to spend the outputs of the tx being replaced
+        self.params
+            .unspendable
+            .extend((0..tx.output.len()).map(|vout| OutPoint::new(txid, vout as u32)));
```
evanlinjin: I think we also need to add the children outputs of the tx being replaced.

crates/wallet/tests/wallet.rs
evanlinjin, line 1: Maybe one more test to make sure we don't use descendant outputs of the tx we are trying to replace.

evanlinjin (changes requested):
ConceptACK, ApproachACK.

Just need to make sure we don't use descendant outputs from tx we are trying to replace.

It's out of the scope of this PR, the inability to set both feerate/fee amount constraints together really bothers me. We should probably switch to the `bdk_coin_select` crate soon.

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -306,6 +310,106 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
+    /// ```
+    ///
+    /// # Errors
+    ///
+    /// - If the original transaction is not found in the tx graph
+    /// - If the orginal transaction is confirmed
+    /// - If none of the inputs are owned by this wallet
+    ///
+    /// [`finish`]: TxBuilder::finish
+    pub fn replace_tx(&mut self, txid: Txid) -> Result<&mut Self, ReplaceTxError> {
```
evanlinjin, line 361: Should probably warn that we don't help keep the original tx's recipient.

crates/wallet/src/wallet/tx_builder.rs
ValuedMammal: Thanks, will fix.

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -306,6 +310,106 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
 
+    /// Replace an unconfirmed transaction.
+    ///
+    /// This method attempts to create a replacement for the transaction with `txid` by
+    /// looking for the largest input that is owned by this wallet and adding it to the
+    /// list of UTXOs to spend.
+    ///
+    /// # Note
+    ///
+    /// Aside from reusing one of the inputs, the method makes no assumptions about the
```
rustaceanrob, line 329: I would be wary of leaving this caveat up to documentation. The way I see it, one could either declare the recipient when calling `replace_tx` as a parameter or a new variant on the error path could be added for `NoRecipient` such that `set_recipient` must be called before `replace_tx`.

ValuedMammal: I see what you mean, but isn't this basically true of all of the builder methods - that we don't know whether a recipient has been added until the user calls `finish`?

ValuedMammal:
I made a few small [changes](https://github.com/bitcoindevkit/bdk/compare/3acc39b3c8875c25d13da1af299d087107f4e4a1..f753f5ec59c03ef1b1cc446525b690d55ff95d4c) including a test for unspendable descendants. I'm still open to discussing the right UX for `replace_tx`.

crates/wallet/src/wallet/tx_builder.rs
rustaceanrob: Fair enough. Forgive my ignorance but what happens if the recipient list is empty?

ValuedMammal: Normally it will be a `NoRecipients` error, unless you're doing a sweep using `drain_wallet`.

notmandatory:
Since still some open ux design questions I removed this from the 1.1 milestone.

crates/wallet/src/wallet/tx_builder.rs
rustaceanrob: Got it. In that case I would expect the user to set the recipients to the original recipients if that is their intended goal.

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -306,6 +311,120 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
+
+        self.add_utxo(outpoint).expect("we must have the utxo");
+
+        // do not try to spend the outputs of the replaced tx including descendants
+        core::iter::once((txid, tx))
+            .chain(
+                self.wallet
+                    .tx_graph()
+                    .walk_descendants(txid, |_, descendant_txid| {
+                        let tx = self.wallet.tx_graph().get_tx(txid)?;
```
ValuedMammal: 🔧 Looks like a bug, we should be getting the tx by descendant txid

```suggestion
                        let tx = self.wallet.tx_graph().get_tx(descendant_txid)?;
```

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -306,6 +316,122 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
+    ///
+    ///     let psbt = builder.finish()?;
+    /// }
+    /// # Ok::<_, anyhow::Error>(())
+    /// ```
+    ///
+    /// # Errors
+    ///
+    /// - If the original transaction is not found in the tx graph
+    /// - If the orginal transaction is confirmed
```
oleonardolima: nit:

``` suggestion
    /// - If the original transaction is confirmed
```

ValuedMammal:
Since `replace_tx` doesn't currently let you to choose which input to replace, another idea could be to have `replace_tx` internally call a new method `replace_tx_input` with an extra argument `input_index: Option<usize>`.

crates/wallet/src/wallet/tx_builder.rs
ValuedMammal: Fixed by using `unwrap_or_default` although it is kinda hacky. But if we reach this part of the code then we must have the previous tx and presumably have no issue calculating the fee. This ensures that `params.bumping_fee` has a value when `add_utxo` is called.

An alternative is to reintroduce a context generic to TxBuilder and have methods only available in that context.

Another alternative is to add a method similar to `add_utxo` but doesn't throw an error if the output is spent.

crates/wallet/src/wallet/tx_builder.rs
ValuedMammal: "warning" sounds a bit too dramatic.

notmandatory (approved):
utACK dae5d73fd4829bcd4835a3b1cc70922e28adc295

My 2 sats on adding `replace_tx_input` is it should wait until if/when someone has a need for it.

tvpeter:
tACK dae5d73fd4829bcd4835a3b1cc70922e28adc295

nit: When the discussion is finalized, it would be great to consider adding a test case for `replace_tx` in the `tests/wallet`, as it is an important wallet feature.

oleonardolima (commented):
utACK dae5d73fd4829bcd4835a3b1cc70922e28adc295

notmandatory:
@evanlinjin your above comments were addressed ok to resolve those comments so we can merge this one in 1.2?

evanlinjin:
Looks pretty good to me on a first glance. I'll double check over it tomorrow to make sure and I'll ACK/merge.

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -274,27 +274,36 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
-                        )
-                    })
+                let output = outputs
+                    .get(outpoint)
+                    .cloned()
+                    .ok_or(AddUtxoError::UnknownUtxo(*outpoint))?;
+                // the output should be unspent unless we're doing a RBF
+                if self.params.bumping_fee.is_none() && output.is_spent {
+                    return Err(AddUtxoError::UnknownUtxo(*outpoint));
+                }
```
evanlinjin: I guess we changed this so that we can call it in `replace_tx` without having it return an error?

However, when we do RBF, it doesn't mean we want to be able to replace every spend.

To avoid this potential footgun, the ideal approach would be to use something like `CanonicalizationParams::assume_not_canonical` in #1809 (which is not helpful here).

Another idea would be to NOT modify `add_utxos` and use something custom for `replace_tx`.

However, `add_utxos` is problematic anyway because "these have priority over the unspendable utxos". So we can still create txs that selects "outputs of the replaced tx as inputs to the replacement".

Would it make more sense to change our logic to make "unspendables" have precedence? It does seem to be safer in this regard.

Ideally, we want to change `add_utxos` to disregard the `is_spent` of outputs that are inputs of the tx-to-replace. ~~We can do this by observing the "unspendables" list and if the output is spent by something of the unspendables, we consider it unspent.~~ We can do this by keeping track of the spends of the tx-to-replace.

ValuedMammal: > We can do this by keeping track of the spends of the tx-to-replace.

I went with this idea and pushed an update f1f57d42.

evanlinjin (commented):
There are still a bunch of issues with this implementation that may result in footguns for the caller. However, I'm not sure if it is worth tackling here because the users really want this now and the only way to solve everything correctly is to use `CanonicalizationParams` which will not be available until `bdk_wallet 2.0`. Let me list them out below:

---

Given 2 unconfirmed txs `(A, B)`, where `B` spends `A`. If the caller does `replace_tx(A)`, then calls `add_utxo(some-output-of-B)`, we will create a tx that simultaneously replaces `A` and spends from a descendant of `A`.

---

Given 2 unconfirmed txs `(A, B)`, where `B` spends `A`. If the caller does `replace_tx` on both `A` and `B`, we will be creating a replacement of `A` & `B` that tries to spend from `A`.

---

Doing this logic in `TxBuilder` means introducing more footguns.

I would prefer if we just create a no-bdk-wallet dependency function that takes in a bunch of params to create an RBF tx. `create_rbf(params: RBFParams) -> Result<Psbt, Error>`. This way this feature is release-agnostic, does not introduce footguns and can be used without a `bdk_wallet` dependency.

I created a some [proposed specifications for this `create_rbf` function](https://drive.proton.me/urls/1BSASWXXYG#ZB1tDXe0C7hU).

thunderbiscuit:
I’m glad to see this getting solid attention in review and and hope we find a good API for users (good meaning usable with minimal footguns, but also existing!).

I am, however, hesitant about shelving an approach that has working code and could be immediately useful, especially in favour of an alternative that, while promising on paper, currently has:
- No working implementation (still at the spec stage, so edge cases may yet surface)
- No tests
- An unclear timeline
- And potentially a dependency on 2.0?

I’m not trying to block progress or be antagonistic—I just want to advocate a bit for users of bdk_wallet. They form our biggest user base, and shipping something that’s imperfect but works today often ends up being more valuable for them than waiting on something ideal that might still be quite a ways off.

evanlinjin:
@thunderbiscuit there is a working implementation.
 https://github.com/bitcoindevkit/bdk-tx/pull/1

I did this in a single day. Check `/tests/synopsis.rs`. Give me a few more days (a week at most) and can make it feature complete.

We won't need to wait until v2.0 to have this since it lives outside of `bdk_wallet`.

In terms of tests, a bunch of the heavy lifting is in `bdk_coin_select` and that is well tested. Also some of the tests in the old wallet-tx-builder are for making sure the policy module works. We're using the planning module in the new implementation. 

The old codebase is a mess and is hard to reason about. I spent half a day reviewing this PR. I prefer putting my time into fixing it up for good.

thunderbiscuit:
I want to give an update for anyone following this PR/thread.

I had a really good offline conversation with @evanlinjin, and it gave me an idea of what the next steps are for me. 

The gist of my hesitation above revolves around the giving up of a useful feature that's ready to go (this PR) in exchange for an alternative for which I don't have a clear understanding of "readiness". Are we at the design phase? The hammering the details phase? How probable is it that users could leverage it in the 1.3/1.4 timeline? (June/August respectively).

This new crate ([bdk-tx](https://github.com/bitcoindevkit/bdk-tx)) has 2 PRs. It's fairly new. But also super promising. Doing away with code that is hard to reason about and leveraging the lessons we've learned over the past 4 years of using the TxBuilder is very appealing. I want to be on board, at least to understand better where this is going.

Here is what I think is missing at this point for bdk-tx to make a strong case for itself:
- There are no core/root issue on the bdk or bdk_wallet repositories making the sales pitch for it. What is it for? Is it intended as the complete replacement of TxBuilder? Can it be done within the 1.0 versions, or will it break anything? Could bdk_wallet maybe start using it one bit at a time?
- I hear here and elsewhere that the TxBuilder "is broken" and "is hard to reason about", and that we should not put more time into building new features into it, but rather focus on a new, better approach. But this TxBuilder has been with BDK since the beginning, and a migration away from it is no small affair. If this is where we're going, a clear outline of what is wrong with it, the issues we hope to fix in a replacement for it, and lessons learned should be written somewhere. May I suggest an ADR for it?

Here is what I commit to:
1. I want to understand better this alternative bdk-tx, and will be an ally for it moving forward in a few ways, namely:
2. Exploring how to use it in conjunction with bdk_wallet. You can find PRs [here (bdk-tx)](https://github.com/bitcoindevkit/bdk-tx/pull/2) and [here (bdk_wallet)](https://github.com/bitcoindevkit/bdk_wallet/pull/15) if you'd like to follow along my poking around these APIs.
3. I'll open some of the issues I mention above that are missing to drive the discussion forward.
4. I will bring this up directly on dev calls and reach out to people to discuss it more and understand better the solution space for this.

PR #1803, opened 2025-01-16
ci: add secondary MSRV job of 1.75.0

### LINKED ISSUE #1750: Consider bumping MSRV
Unfortunately `rustls` recently decided to bump their MSRV to 1.71 (see https://github.com/rustls/rustls/issues/2239). While they now at least intermittently reverted the MSRV and pushed out version 0.23.19 that fixes [RUSTSEC-2024-0399](https://rustsec.org/advisories/RUSTSEC-2024-0399.html) *without* at bumping MSRV at the same time, they are commited to re-bump their MSRV to 1.71 with the follow-up (i.e., 0.23.20) release.

Given how security-critical TLS libraries are, I don't think an MSRV of 1.63 for the dependent crates can be maintained through pinning, as users might miss out on critical security patches. That puts any dependent crates in a tough spot, and essentially forces them to bump their MSRV to at least 1.71, too.

In LDK we're probably going to introduce a separate MSRV just for the `lightning-transaction-sync` crate (currently depending on `rustls`), which will have further consequences, such as bumping MSRV on LDK Node and other dependent projects down the dependency chain. We might also re-evaluate whether `native-tls` can be made to work on Android/Kotlin bindings.

I'm opening this discussion topic as BDK probably also needs to act on this in one form or another, i.e., either bumping MSRV in general, or introducing a separate MSRV for anything depending on `rustls`. 

It probably would also make sense to figure out a common MSRV target for the affected crates to at least somewhat maintain a standard that users can lean on here. It might just be 1.71, but then again `rustls`'s newly defined [MSRV policy](https://github.com/rustls/rustls/blob/main/SECURITY.md#minimum-supported-rust-version) would 'allow' (not that they seem to care too much in general) them to bump up to 1.73 currently.

oleonardolima:
partially fixes #1750

### Description

It adds a secondary MSRV version of `1.75.0` for the `bdk_electrum` crate, excluding it from the `1.63.0` MSRV dependency pinning and CI step.

As it's the new CI will run without `bdk_electrum` for `1.63.0` MSRV step, will run another step for `1.75.0` MSRV and the latest stable.

Currently,, the only affected crate is `electrum-client`, therefore `bdk-electrum`, as it relies directly on `rustls` which migrated to `1.71.0` MSRV. The `esplora-client` relies on `rustls` as a dependency of `minreq` or `reqwest`, but those crates didn't upgraded it's `rustls` versions yet.

In a further PR it should also bump the `electrum-client` crate to it's latest version, which relies on `1.75.0` MSRV, see https://github.com/bitcoindevkit/rust-electrum-client/pull/159.

### Notes to the reviewers

It's open for discussion if this approach is the expected one, or another one would be better.

### Changelog notice

- Add a secondary MSRV of `1.75.0` for `bdk_electrum`.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Fully working examples of how to use these components are in `/example-crates`:
 [`bdk_chain`]: https://docs.rs/bdk-chain/
 
 ## Minimum Supported Rust Version (MSRV)
-The BDK library maintains a MSRV of 1.63.0. This includes the following crates 
+The BDK library maintains a MSRV of 1.63.0. This includes the following crates:
 
 - `bdk_core`
 - `bdk_chain`
@@ -73,7 +73,7 @@ The BDK library maintains a MSRV of 1.63.0. This includes the following crates
 
 The MSRV of `bdk_electrum` is 1.75.0.
 
-To build with the MSRV of 1.63.0 you will need to pin dependencies by running the [`pin-msrv.sh`](./ci/pin-msrv.sh) script
+To build with the MSRV of 1.63.0 you will need to pin dependencies by running the [`pin-msrv.sh`](./ci/pin-msrv.sh) script.
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -73,22 +73,7 @@ The BDK library maintains a MSRV of 1.63.0. This includes the following crates
 
 The MSRV of `bdk_electrum` is 1.75.0.
 
-To build with the MSRV you will need to pin dependencies as follows:
-
-```shell
-cargo update -p zstd-sys --precise "2.0.8+zstd.1.5.5"
-cargo update -p time --precise "0.3.20"
-cargo update -p home --precise "0.5.5"
-cargo update -p proptest --precise "1.2.0"
-cargo update -p url --precise "2.5.0"
-cargo update -p cc --precise "1.0.105"
-cargo update -p tokio --precise "1.38.1"
-cargo update -p tokio-util --precise "0.7.11"
-cargo update -p indexmap --precise "2.5.0"
-cargo update -p security-framework-sys --precise "2.11.1"
-cargo update -p csv --precise "1.3.0"
-cargo update -p unicode-width --precise "0.1.13"
-```
+To build with the MSRV of 1.63.0 you will need to pin dependencies by running the [`pin-msrv.sh`](./ci/pin-msrv.sh) script
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -63,7 +63,15 @@ Fully working examples of how to use these components are in `/example-crates`:
 [`bdk_chain`]: https://docs.rs/bdk-chain/
 
 ## Minimum Supported Rust Version (MSRV)
-This library should compile with any combination of features with Rust 1.63.0.
+The BDK library maintains a MSRV of 1.63.0. This includes the following crates 
+
+- `bdk_core`
+- `bdk_chain`
+- `bdk_bitcoind_rpc`.
+- `bdk_esplora`.
+- `bdk_wallet`.
+
+The MSRV of `bdk_electrum` is 1.75.0.
 
 To build with the MSRV you will need to pin dependencies as follows:
 
@@ -80,7 +88,6 @@ cargo update -p indexmap --precise "2.5.0"
 cargo update -p security-framework-sys --precise "2.11.1"
 cargo update -p csv --precise "1.3.0"
 cargo update -p unicode-width --precise "0.1.13"
-cargo update -p rustls@0.23.21 --precise "0.23.19"
 ```
 
 ## License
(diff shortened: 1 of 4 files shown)


ValuedMammal:
I sort of like the idea of separating the "check" and "test" jobs. For example this CI run https://github.com/ValuedMammal/bdk/actions/runs/12872751304

oleonardolima:
> I sort of like the idea of separating the "check" and "test" jobs. For example this CI run https://github.com/ValuedMammal/bdk/actions/runs/12872751304

That's a good idea, I'll cherry-pick the commit.

oleonardolima:
> > I sort of like the idea of separating the "check" and "test" jobs. For example this CI run https://github.com/ValuedMammal/bdk/actions/runs/12872751304
> 
> That's a good idea, I'll cherry-pick the commit.

However, @ValuedMammal any reason to not have tests run with `--no-default-features` too?

ValuedMammal:
> > > I sort of like the idea of separating the "check" and "test" jobs. For example this CI run https://github.com/ValuedMammal/bdk/actions/runs/12872751304
> > 
> > 
> > That's a good idea, I'll cherry-pick the commit.
> 
> However, @ValuedMammal any reason to not have tests run with `--no-default-features` too?

I thought the same thing and I don't know. It didn't seem like that would pick up any tests that weren't already included in `--all-features`, but in any case I'm in favor of running all possible tests.

oleonardolima:
> > > > I sort of like the idea of separating the "check" and "test" jobs. For example this CI run https://github.com/ValuedMammal/bdk/actions/runs/12872751304
> > > 
> > > 
> > > That's a good idea, I'll cherry-pick the commit.
> > 
> > 
> > However, @ValuedMammal any reason to not have tests run with `--no-default-features` too?
> 
> I thought the same thing and I don't know. It didn't seem like that would pick up any tests that weren't already included in `--all-features`, but in any case I'm in favor of running all possible tests.

I cherry-picked and add the features to matrix, but it now fails because not all crates has `miniscript/no-std`, I'll need to sort that out.

oleonardolima:
Alright, I kept it all under a single build-test job, using the `./ci/pin-msrv.sh` script and referring to it on README.md.

It should be ready for review.

README.md
```diff
@@ -63,25 +63,17 @@ Fully working examples of how to use these components are in `/example-crates`:
...
-cargo update -p cc --precise "1.0.105"
-cargo update -p tokio --precise "1.38.1"
-cargo update -p tokio-util --precise "0.7.11"
-cargo update -p indexmap --precise "2.5.0"
-cargo update -p security-framework-sys --precise "2.11.1"
-cargo update -p csv --precise "1.3.0"
-cargo update -p unicode-width --precise "0.1.13"
-cargo update -p rustls@0.23.21 --precise "0.23.19"
-```
+The BDK library maintains a MSRV of 1.63.0. This includes the following crates
```
notmandatory:
```suggestion
The BDK library maintains a MSRV of 1.63.0. This includes the following crates:
```

README.md
```diff
@@ -63,25 +63,17 @@ Fully working examples of how to use these components are in `/example-crates`:
...
+
+- `bdk_core`
+- `bdk_chain`
+- `bdk_bitcoind_rpc`.
+- `bdk_esplora`.
+- `bdk_wallet`.
+
+The MSRV of `bdk_electrum` is 1.75.0.
+
+To build with the MSRV of 1.63.0 you will need to pin dependencies by running the [`pin-msrv.sh`](./ci/pin-msrv.sh) script
```
notmandatory:
```suggestion
To build with the MSRV of 1.63.0 you will need to pin dependencies by running the [`pin-msrv.sh`](./ci/pin-msrv.sh) script.
```

ci/pin-msrv.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -x
+set -euo pipefail
+
+# Pin dependencies for MSRV
+
+# To pin deps, switch toolchain to MSRV and execute the below updates
+
+# cargo clean
+# rustup default 1.63.0
```
notmandatory: I had to use this command:
```suggestion
# rustup override set 1.63.0
```

notmandatory (commented):
Looks good, I found a couple nits to fix and it needs to be rebased now that #1801 is merged.

ci/pin-msrv.sh
oleonardolima: It depends on whether the user is using `overrides` for the directory or not 🤔. I guess it's best to recommend using a command that has only side-effects within the directory, will update.

notmandatory (approved):
ACK c8b55da39913ddaf11d26f09d92034aef4f550e0

notmandatory:
I updated Github required checks to match changes in this PR.

Issue #1576, opened 2024-08-27
[chain] Make `LocalChain::apply_changeset` infallible


ValuedMammal:
It seems like the only way this can fail is if we don't find a base AND `extension` doesn't include block 0, but I think one of these will have to be true, i.e. `extension` must have a genesis block or else we have a base.

https://github.com/bitcoindevkit/bdk/blob/775e4aee3a9ec2d99adb129f7a9367a3a15c7032/crates/chain/src/local_chain.rs#L23-L30

https://github.com/bitcoindevkit/bdk/blob/775e4aee3a9ec2d99adb129f7a9367a3a15c7032/crates/chain/src/local_chain.rs#L43-L48

ValuedMammal:
Perhaps more importantly, this function actually allows you to change the genesis block if the changeset happens to include it (although why would it?) which seems incongruent with most other methods

notmandatory:
Is this something you think should be done for the 1.0.0-beta milestone? would it help making any user wallet facing functions infallible ?

ValuedMammal:
I don't think it's terribly urgent and seems like it mostly affects developers who are working with `LocalChain`. The issue was originally motivated by a discussion with @rustaceanrob who pointed out that `MissingGenesisError` is surprisingly common for something that should be an inherent property of a `LocalChain` object.

rustaceanrob:
What I mean by inherent or an invariant of `LocalChain` is if all constructors of `LocalChain` require a genesis block then calls of merging chains may only be fallible in that they do not connect

PR #1806, opened 2025-01-21
fix(docs): `merge_chains` outdated documentation

oleonardolima:
### Description

It's a minor documentation fix, as reported during the audit and referenced in bitcoindevkit/bdk_wallet#59.

### Notes to the reviewers

I'm unsure if anything else / any other explanation should be included in the documentation. Let me know if you think more context should be added to it.

### Changelog notice

- Update `bdk_chain::local_chain::merge_chains` documentation.

#### All Submissions:

### DIFF
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -545,31 +545,46 @@ impl std::error::Error for ApplyHeaderError {}
 
 /// Applies `update_tip` onto `original_tip`.
 ///
-/// On success, a tuple is returned `(changeset, can_replace)`. If `can_replace` is true, then the
-/// `update_tip` can replace the `original_tip`.
+/// On success, a tuple is returned ([`CheckPoint`], [`ChangeSet`]).
+///
+/// # Errors
+///
+/// [`CannotConnectError`] occurs when the `original_tip` and `update_tip` chains are disjoint:
+///
+/// - If no point of agreement is found between the update and original chains.
+/// - A point of agreement is found but the update is ambiguous above the point of agreement (a.k.a.
+///   the update and original chain both have a block above the point of agreement, but their
+///   heights do not overlap).
+/// - The update attempts to replace the genesis block of the original chain.
 fn merge_chains(
     original_tip: CheckPoint,
     update_tip: CheckPoint,
 ) -> Result<(CheckPoint, ChangeSet), CannotConnectError> {
     let mut changeset = ChangeSet::default();
+
     let mut orig = original_tip.iter();
     let mut update = update_tip.iter();
+
     let mut curr_orig = None;
     let mut curr_update = None;
+
     let mut prev_orig: Option<CheckPoint> = None;
     let mut prev_update: Option<CheckPoint> = None;
+
     let mut point_of_agreement_found = false;
+
     let mut prev_orig_was_invalidated = false;
+
     let mut potentially_invalidated_heights = vec![];
 
     // If we can, we want to return the update tip as the new tip because this allows checkpoints
     // in multiple locations to keep the same `Arc` pointers when they are being updated from each
-    // other using this function. We can do this as long as long as the update contains every
+    // other using this function. We can do this as long as the update contains every
     // block's height of the original chain.
     let mut is_update_height_superset_of_original = true;
 
     // To find the difference between the new chain and the original we iterate over both of them
-    // from the tip backwards in tandem. We always dealing with the highest one from either chain
+    // from the tip backwards in tandem. We are always dealing with the highest one from either chain
     // first and move to the next highest. The crucial logic is applied when they have blocks at the
     // same height.
     loop {


crates/chain/src/local_chain.rs
```diff
@@ -545,8 +545,10 @@ impl std::error::Error for ApplyHeaderError {}
 
 /// Applies `update_tip` onto `original_tip`.
 ///
-/// On success, a tuple is returned `(changeset, can_replace)`. If `can_replace` is true, then the
-/// `update_tip` can replace the `original_tip`.
+/// On success, a tuple is returned `([`CheckPoint`], [`ChangeSet`])`.
+///
+/// [`CheckPoint`]: bdk_core::CheckPoint
+/// [`ChangeSet`]: bdk_chain::local_chain::CheckPoint
```
ValuedMammal: We can add a note about how this could fail when the chains don't connect. The test [`update_local_chain`](https://github.com/bitcoindevkit/bdk/blob/43f0f8dea54aec92ddf7e088245c4ac9ba7afcd6/crates/chain/tests/test_local_chain.rs#L75) has some examples of this where the error is typically a variation of "two disjoint chains cannot merge".

I see we also propagate the error from `apply_changeset_to_checkpoint`, but as far as I know we don't have coverage for it (#1576).

oleonardolima: Yes, I'll take a look at those.

oleonardolima: I'm not sure if I covered all scenarios, but hopefully the new errors section is a good start for discussion. Let me know what you think.

ValuedMammal: I think it looks good. I might enumerate the error cases like this, which is roughly the same as what you came up with

```rust
/// - If no point of agreement is found between the update and original chains.
/// - A point of agreement is found but we were unable to invalidate heights of the original chain.
/// - A point of agreement is found but we failed to apply the changeset to `CheckPoint`.
```

ValuedMammal: All of these types are in scope and shouldn't need markdown links

- `CheckPoint`
- `ChangeSet`
- `CannotConnectError`

crates/chain/src/local_chain.rs
```diff
@@ -545,31 +545,50 @@ impl std::error::Error for ApplyHeaderError {}
 
 /// Applies `update_tip` onto `original_tip`.
 ///
-/// On success, a tuple is returned `(changeset, can_replace)`. If `can_replace` is true, then the
-/// `update_tip` can replace the `original_tip`.
+/// On success, a tuple is returned `([`CheckPoint`], [`ChangeSet`])`.
```
ValuedMammal: Maybe remove the outer set of backticks
```suggestion
/// On success, a tuple is returned ([`CheckPoint`], [`ChangeSet`]).
```

crates/chain/src/local_chain.rs
```diff
@@ -545,31 +545,50 @@ impl std::error::Error for ApplyHeaderError {}
...
+///
+/// # Errors
+///
+/// [`CannotConnectError`] occurs when the `original_tip`` and `update_tip` chains are disjoint:
```
ValuedMammal:
```suggestion
/// [`CannotConnectError`] occurs when the `original_tip` and `update_tip` chains are disjoint:
```

oleonardolima:
@ValuedMammal Great, thanks for the suggestions! I've rebased the branch and applied the suggestions.

ValuedMammal (approved):
ACK 166a8f92ec60cd4e4091e46db5ba2541da34a59e

crates/chain/src/local_chain.rs
```diff
@@ -545,31 +545,44 @@ impl std::error::Error for ApplyHeaderError {}
 
 /// Applies `update_tip` onto `original_tip`.
 ///
-/// On success, a tuple is returned `(changeset, can_replace)`. If `can_replace` is true, then the
-/// `update_tip` can replace the `original_tip`.
+/// On success, a tuple is returned ([`CheckPoint`], [`ChangeSet`]).
```
notmandatory, line 548:
```suggestion
/// On success, a tuple is returned ([`CheckPoint`], [`ChangeSet`]). A [`CannotConnectError`] occurs when the `original_tip` and `update_tip` chains are disjoint.
```

I'd move the rest of the explanation for `CannotConnectError` to that type's doc. 

But I don't want to bike-shed it too much so I'll also ACK this as is and you decide.

notmandatory (approved):
ACK 166a8f92ec60cd4e4091e46db5ba2541da34a59e

crates/chain/src/local_chain.rs
```diff
@@ -545,31 +545,44 @@ impl std::error::Error for ApplyHeaderError {}
...
+///
+/// # Errors
+///
+/// [`CannotConnectError`] occurs when the `original_tip` and `update_tip` chains are disjoint:
+///
+/// - If no point of agreement is found between the update and original chains.
+/// - A point of agreement is found but we were unable to invalidate heights of the original chain.
```
evanlinjin:
```suggestion
/// - A point of agreement is found but the update is ambiguous above the point of agreement (a.k.a. the update and original chain both have a block above the point of agreement, but their heights do not overlap).
```

crates/chain/src/local_chain.rs
```diff
@@ -545,31 +545,44 @@ impl std::error::Error for ApplyHeaderError {}
...
+/// - A point of agreement is found but we failed to apply the changeset to `CheckPoint`.
```
evanlinjin:
```suggestion
/// - The update attempts to replace the genesis block of the original chain.
```

evanlinjin (approved):
self-ACK 5647241f45b29fe5f7ad5aa5a94c0c4f72f9adea

Just fixed up some documentation errors.

CosmicJesterX (commented):
hi

PR #1808, opened 2025-01-23
Introduce canonicalization parameters

### LINKED ISSUE #40: Don't double spend created transaction just because they haven't been broadcasted
This was a comment from a discussion elsewhere. I don't want to restrict the design space too much but I personally think the right way to deal with this is:

1. Insert the transactions into the tx graph right after they are created (without signatures)
2. Mark the txid as "unbroadcasted" at the tx graph level or application level
3. Assume it's canonical whenever you do canonicalization
4. Once it's signed, add the signatures to the tx graph. This can be done by re-inserting the transaction and adding it to the changeset with the new signatures.
5. When it gets a `last_seen` remove it from the unbraodcasted state.

I think in (2) it probably should be done inside the tx graph. bitcoindevkit/bdk_wallet#130 should be done first.

---

The user has to decide the fate of their transactions. And I claim each unconfirmed tx is either in a state of:

1. The user has not broadcasted it (maybe it's still being signed) and wants to confirm it eventually
7. It's broadcasted and the user still wants it to be confirmed
8. It's broadcasted but the user wants to replace it
9. It's not broadcasted and the user no longer wants to broadcast it.

We have to consider the state when creating a new tx: in (1,2) we want to treat the tx as canonical regardless of whether it has been seen in the mempool recently or not and therefore it's UTXOs can be candidate inputs. In (2), if it hasn't been seen for a while this will mean we want the user to broadcast it again along with the new child tx.
In state (3) the application has to treat it as non-canonical and double spend it with the replacing tx.  In state (4) it's just non-canonical and remains that way.

I don't think there is a fifth state we have to model where the user is thinking "I broadcasted this tx but you know I don't really care if it gets confirmed. If it hasn't been seen for a while I kinda want to prune it so I might inadvertently double spend it with my next tx...or I might not..whatever". 

There are several ways in which these four tx states are not modeled correctly in BDK so there definitely is a design bug but it seems we are not in agreement to the nature of it. I think states (2) and (4) are easy to express, YMWV with (3) and (1) doesn't exist.

_Originally posted by @LLFourn in https://github.com/bitcoindevkit/bdk/issues/1666#issuecomment-2502098294_

### LINKED ISSUE #1748: Don't double spend created transaction just because they haven't been broadcasted

evanlinjin:
Partially Fixes bitcoindevkit/bdk_wallet#40

### Description

Add the ability to modify the canonicalization algorithm. Right now, the only modifier is `assume_canonical` which takes in a `Vec` (ordered list) of txids and superimposes it on the canonicalization algorithm. Txs later in the list (higher index) have a higher priority (in case of conflicts).

### Notes to the reviewers

None at the moment.

### Changelog notice

* Added `CanonicalizationParams` to allow the caller to modify the canonicalization algorithm. This in a new parameter on `CanonicalIter::new`.
* Changed `TxGraph::insert_tx` to allow for updating a transaction's witness field. This is useful for initially introducing an unsigned tx and adding witnesses later on.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/canonical_iter.rs
+++ b/crates/chain/src/canonical_iter.rs
@@ -17,10 +17,10 @@ pub struct CanonicalIter<'g, A, C> {
     chain: &'g C,
     chain_tip: BlockId,
 
-    unprocessed_txs_with_anchors:
+    unprocessed_anchored_txs:
         Box<dyn Iterator<Item = (Txid, Arc<Transaction>, &'g BTreeSet<A>)> + 'g>,
-    unprocessed_txs_with_last_seens: Box<dyn Iterator<Item = (Txid, Arc<Transaction>, u64)> + 'g>,
-    unprocessed_txs_left_over: VecDeque<(Txid, Arc<Transaction>, u32)>,
+    unprocessed_seen_txs: Box<dyn Iterator<Item = (Txid, Arc<Transaction>, u64)> + 'g>,
+    unprocessed_leftover_txs: VecDeque<(Txid, Arc<Transaction>, u32)>,
 
     canonical: CanonicalMap<A>,
     not_canonical: NotCanonicalSet,
@@ -32,12 +32,12 @@ impl<'g, A: Anchor, C: ChainOracle> CanonicalIter<'g, A, C> {
     /// Constructs [`CanonicalIter`].
     pub fn new(tx_graph: &'g TxGraph<A>, chain: &'g C, chain_tip: BlockId) -> Self {
         let anchors = tx_graph.all_anchors();
-        let pending_anchored = Box::new(
+        let unprocessed_anchored_txs = Box::new(
             tx_graph
                 .txids_by_descending_anchor_height()
                 .filter_map(|(_, txid)| Some((txid, tx_graph.get_tx(txid)?, anchors.get(&txid)?))),
         );
-        let pending_last_seen = Box::new(
+        let unprocessed_seen_txs = Box::new(
             tx_graph
                 .txids_by_descending_last_seen()
                 .filter_map(|(last_seen, txid)| Some((txid, tx_graph.get_tx(txid)?, last_seen))),
@@ -46,9 +46,9 @@ impl<'g, A: Anchor, C: ChainOracle> CanonicalIter<'g, A, C> {
             tx_graph,
             chain,
             chain_tip,
-            unprocessed_txs_with_anchors: pending_anchored,
-            unprocessed_txs_with_last_seens: pending_last_seen,
-            unprocessed_txs_left_over: VecDeque::new(),
+            unprocessed_anchored_txs,
+            unprocessed_seen_txs,
+            unprocessed_leftover_txs: VecDeque::new(),
             canonical: HashMap::new(),
             not_canonical: HashSet::new(),
             queue: VecDeque::new(),
@@ -77,7 +77,7 @@ impl<'g, A: Anchor, C: ChainOracle> CanonicalIter<'g, A, C> {
             }
         }
         // cannot determine
-        self.unprocessed_txs_left_over.push_back((
+        self.unprocessed_leftover_txs.push_back((
             txid,
             tx,
             anchors
@@ -190,7 +190,7 @@ impl<A: Anchor, C: ChainOracle> Iterator for CanonicalIter<'_, A, C> {
                 return Some(Ok((txid, tx, reason)));
             }
 
-            if let Some((txid, tx, anchors)) = self.unprocessed_txs_with_anchors.next() {
+            if let Some((txid, tx, anchors)) = self.unprocessed_anchored_txs.next() {
                 if !self.is_canonicalized(txid) {
                     if let Err(err) = self.scan_anchors(txid, tx, anchors) {
                         return Some(Err(err));
@@ -199,7 +199,7 @@ impl<A: Anchor, C: ChainOracle> Iterator for CanonicalIter<'_, A, C> {
                 continue;
             }
 
-            if let Some((txid, tx, last_seen)) = self.unprocessed_txs_with_last_seens.next() {
+            if let Some((txid, tx, last_seen)) = self.unprocessed_seen_txs.next() {
                 if !self.is_canonicalized(txid) {
                     let observed_in = ObservedIn::Mempool(last_seen);
                     self.mark_canonical(txid, tx, CanonicalReason::from_observed_in(observed_in));
@@ -207,7 +207,7 @@ impl<A: Anchor, C: ChainOracle> Iterator for CanonicalIter<'_, A, C> {
                 continue;
             }
 
-            if let Some((txid, tx, height)) = self.unprocessed_txs_left_over.pop_front() {
+            if let Some((txid, tx, height)) = self.unprocessed_leftover_txs.pop_front() {
                 if !self.is_canonicalized(txid) {
                     let observed_in = ObservedIn::Block(height);
                     self.mark_canonical(txid, tx, CanonicalReason::from_observed_in(observed_in));
--- a/crates/chain/src/canonical_iter.rs
+++ b/crates/chain/src/canonical_iter.rs
@@ -11,12 +11,23 @@ use bitcoin::{Transaction, Txid};
 type CanonicalMap<A> = HashMap<Txid, (Arc<Transaction>, CanonicalReason<A>)>;
 type NotCanonicalSet = HashSet<Txid>;
 
+/// Modifies the canonicalization algorithm.
+#[derive(Debug, Default, Clone)]
+pub struct CanonicalizationParams {
+    /// Transactions that will supercede all other transactions.
+    ///
+    /// In case of conflicting transactions within `assume_canonical`, transactions that appear
+    /// later in the list (have higher index) have precedence.
+    pub assume_canonical: Vec<Txid>,
+}
+
 /// Iterates over canonical txs.
 pub struct CanonicalIter<'g, A, C> {
     tx_graph: &'g TxGraph<A>,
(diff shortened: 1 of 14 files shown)


notmandatory:
Would it make sense to add `CanonicalizationMods` as a field of `TxGraph` ? 

This way the existing `TxGraph` method signatures don't change, you'd just need a new function to add `TxId`s to the `CanonicalizationMods` in the `TxGraph`.   

And if you could only add new `TxIds` to your `CanonicalizationMods` this should still be monotone right? 

Then you could also add the `CanonicalizationMods` data to the `tx_graph::Changeset` so it's persisted. Of course you'd need to update the persistence crates but wouldn't users need/want to persist these mods anyway so a reloaded wallet shows the same balance and utxos? 

For `Wallet` I'm imaging some higher level functions for the things a wallet user needs. For instance (based on suggestions in bitcoindevkit/bdk_wallet#40): 

1. change TxBuilder::finish to insert new unbroadcast tx and add a canonical mod so their inputs won't be double spent
2. add a canonical mod for a new RBF tx so it has precedence over the unconfirmed tx it's replacing
3. list unbroadcast tx (found in CanonicalizationMods but not seen)
4. add an option to TxBuilder to include inputs from unbroadcast tx(s), the new tx has higher canonicalization
5. remove tx from CanonicalizationMods after seen in mempool (or would this make TxGraph not monotone?)

crates/chain/src/canonical_iter.rs
```diff
@@ -4,19 +4,38 @@ use crate::{Anchor, ChainOracle, TxGraph};
 use alloc::boxed::Box;
 use alloc::collections::BTreeSet;
 use alloc::sync::Arc;
+use alloc::vec::Vec;
 use bdk_core::BlockId;
 use bitcoin::{Transaction, Txid};
 
+/// Modifies the canonicalization algorithm.
+#[derive(Debug, Default, Clone)]
+pub struct CanonicalizationMods {
```
LLFourn: "Params" feels more idiomatic.

crates/chain/src/canonical_iter.rs
```diff
@@ -4,19 +4,38 @@ use crate::{Anchor, ChainOracle, TxGraph};
...
+    /// Transactions that will supercede all other transactions.
+    ///
+    /// In case of conflicting transactions within `assume_canonical`, transactions that appear
+    /// later in the list (have higher index) have precedence.
+    pub assume_canonical: Vec<Txid>,
+}
+
+impl CanonicalizationMods {
+    /// No mods.
+    pub const NONE: Self = Self {
```
LLFourn: Dunno if we need this. `Default` is enough.

crates/chain/src/tx_graph.rs
```diff
@@ -590,11 +591,20 @@ impl<A: Anchor> TxGraph<A> {
...
             TxNodeInternal::Whole(existing_tx) => {
-                debug_assert_eq!(
-                    existing_tx.as_ref(),
-                    tx.as_ref(),
-                    "tx of same txid should never change"
-                );
+                // We want to be able to replace an unsigned tx with a signed tx.
+                // The tx with more weight has precedence (and tiebreak with the actual tx data).
+                // We can also check whether the witness is valid and also prioritize signatures
+                // with less weight, but that is more work and this solution is good enough.
```
LLFourn: Prediction: It's less lines of code than this comment to just do it correctly.

LLFourn (dismissed):

LLFourn:
ConceptACK. Would be nice to have draft PR built on this that resolves bitcoindevkit/bdk_wallet#40 so we know that this is all that's needed at the chain level before merging this.

crates/chain/src/canonical_iter.rs
oleonardolima: Agree, it also aligns better with other usages in the project, i.e: TxParams, LoadParams.

oleonardolima (commented):
cACK 0f00cdb4750d06dd1fa52bf9878933e5f6f3dd4d

notmandatory:

evanlinjin:
> Would it make sense to add CanonicalizationMods as a field of TxGraph ?

@notmandatory I do not think so. Different calls to `canonicalize` will be for different purposes. Especially when we add more fields to `CanonicalizationMods` (i.e. in bitcoindevkit/bdk#1809).

evanlinjin:
> And if you could only add new TxIds to your CanonicalizationMods this should still be monotone right?

We want ordering for these txids. In case of conflict between them, we need a way to prioritize ones over others. This order should be controlled by the caller.

crates/chain/src/tx_graph.rs
evanlinjin: Changed it so we zip across both tx inputs and compare witnesses. We prioritize non-empty witnesses. If witnesses both have data, we prioritize the witness that is smaller. If both witnesses are the same size, we prioritize based on lexicographical order.

evanlinjin:
@LLFourn I have a draft PR (bitcoindevkit/bdk_wallet#220) which showcases that this API is enough for `bdk_wallet` to keep track of unbroadcasted transactions (and not accidentally replace them).

crates/chain/tests/test_tx_graph.rs
```diff
@@ -284,6 +286,74 @@ fn insert_tx_displaces_txouts() {
     assert_eq!(tx_graph.get_txout(outpoint), Some(txout));
 }
 
+#[test]
+fn insert_signed_tx_displaces_unsigned() {
```
LagginTimes: Could we add tests for witness size preference and lexicographic tiebreaking? Maybe something like this:
https://github.com/LagginTimes/bdk/commit/354ea8ecabe1751a5eaca7e71d01108c35e0fad7

crates/chain/src/canonical_iter.rs
```diff
@@ -30,8 +41,19 @@ pub struct CanonicalIter<'g, A, C> {
 
 impl<'g, A: Anchor, C: ChainOracle> CanonicalIter<'g, A, C> {
     /// Constructs [`CanonicalIter`].
-    pub fn new(tx_graph: &'g TxGraph<A>, chain: &'g C, chain_tip: BlockId) -> Self {
+    pub fn new(
+        tx_graph: &'g TxGraph<A>,
+        chain: &'g C,
+        chain_tip: BlockId,
+        mods: CanonicalizationParams,
```
LagginTimes: Nit: Might be better to rename the parameter to `params` to match the renaming from `CanonicalizationMods` to `CanonicalizationParams`. This will have to be done in `tx_graph.rs` as well.

crates/chain/tests/common/tx_template.rs
```diff
@@ -51,25 +52,34 @@ impl TxOutTemplate {
     }
 }
 
+#[allow(dead_code)]
+pub struct TxTemplateEnv<'a, A> {
+    pub tx_graph: TxGraph<A>,
+    pub indexer: SpkTxOutIndex<u32>,
+    pub txid_to_name: HashMap<&'a str, Txid>,
+    pub canonicalization_mods: CanonicalizationParams,
```
LagginTimes: Nit: This will probably need updating as well.
```suggestion
    pub canonicalization_params: CanonicalizationParams,
```

LagginTimes (commented):
Just a few consistency nits:

* Commit message 53d6fa1c2f86cd589d5b827ba5aab45a91a6d66c needs to be updated to rename `CanonicalizationMods` → `CanonicalizationParams`.
* Same change needs to be done in the PR description changelog.

Besides these, this LGTM!

LagginTimes (approved):
ACK 53afb3511cb472c0fd91ccf2815fa0c44ffc56ba

ValuedMammal (approved):
ACK 53afb3511cb472c0fd91ccf2815fa0c44ffc56ba

ValuedMammal:
Looks good to me.

@LLFourn Please re-review if you find any issues

PR #1809, opened 2025-01-23
Add canonicalization modifier: assume not canonical.

evanlinjin:
### Description

This is useful for creating RBF transactions where you do not want to spend from txs you are replacing or descendants of txs you are replacing.

I noticed that some users think it is useful to implicitly cancel a tx. This can also be used for that. cc. #1799, #1764.

### Notes to the reviewers

This PR is based on #1808. Please review and merge that first.

### Changelog notice

WIP

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/bitcoind_rpc/tests/test_emitter.rs
+++ b/crates/bitcoind_rpc/tests/test_emitter.rs
@@ -5,7 +5,7 @@ use bdk_chain::{
     bitcoin::{Address, Amount, Txid},
     local_chain::{CheckPoint, LocalChain},
     spk_txout::SpkTxOutIndex,
-    Balance, BlockId, IndexedTxGraph, Merge,
+    Balance, BlockId, CanonicalizationMods, IndexedTxGraph, Merge,
 };
 use bdk_testenv::{anyhow, TestEnv};
 use bitcoin::{hashes::Hash, Block, OutPoint, ScriptBuf, WScriptHash};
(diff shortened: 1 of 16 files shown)


LLFourn:
As in #1808 it would be nice to have a draft PR that uses this to clean up RBF in wallet to see if we've got everything we need here.

notmandatory:

evanlinjin:
I've realized there's a flaw with the `assume_not_canonical` approach.

Consider a case where transaction `A` has one output `A:0`. Both `B` and `C` spend `A:0`, making them conflicting transactions. If we simply assume that `B` is not canonical, `C` will still end up spending `A:0` in the canonical result — assuming nothing else conflicts with `A`, `C`, or their ancestors.

Since this feature is intended to support RBF behavior, this logic doesn't actually help us. The correct `CanonicalizationParams` field would be something like:

```rust
assume_not_spent: HashSet<OutPoint>,
```

evanlinjin:
I've realized there's a flaw with the `assume_not_canonical` approach.

Consider a case where transaction `A` has one output `A:0`. Both `B` and `C` spend `A:0`, making them conflicting transactions. If we simply assume that `B` is not canonical, `C` will still end up spending `A:0` in the canonical result — assuming nothing else conflicts with `A`, `C`, or their ancestors.

Since this feature is intended to support RBF behavior, this logic doesn't actually help us. The correct `CanonicalizationParams` field would be something like:

```rust
assume_not_spent: HashSet<OutPoint>,
```

PR #1811, opened 2025-01-24
Introduce `evicted-at`/`last-evicted` timestamps

### LINKED ISSUE #1740: Undetected double-spent

evanlinjin:
Fixes #1740.

### Description

This PR replaces #1765. For context and the original discussion that led to this change, please refer to [this comment](https://github.com/bitcoindevkit/bdk/pull/1765#issuecomment-2611491174).

This PR addresses a potential malicious double-spending issue by introducing improvements to unconfirmed transaction tracking. Key changes include the addition of `TxUpdate::missing` that tracks transactions that have been replaced and are no longer in the mempool, and the inclusion of `last_evicted` and `evicted_at` timestamps in `TxGraph` to track when a transaction was last deemed missing. 

`SpkWithExpectedTxids` is introduced in `SpkClient` to track expected `Txid`s for each `spk`. During a sync, if any `Txid`s from `SpkWithExpectedTxids` are not in the current history of an `spk` obtained from the chain source, those `Txid`s are considered missing. Support for `SpkWithExpectedTxids` has been added to both `bdk_electrum` and `bdk_esplora` chain source crates.

The canonicalization algorithm is updated to disregard transactions with a `last_evicted` timestamp greater than or equal to their `last_seen` timestamp, except in cases where transitivity rules apply.

### Changelog notice

* `TxUpdate::missing` tracks transactions that have been replaced and are no longer present in mempool.
* `last_evicted` and `evicted_at` timestamps in `TxGraph` track when a transaction was last marked as missing from the mempool.
* The canonicalization algorithm now disregards transactions with a `last_evicted` timestamp greater than or equal to it's `last_seen` timestamp, except when a canonical descendant exists due to rules of transitivity.
* `SpkWithExpectedTxids` in `SpkClient` keeps track of expected `Txid`s for each `spk`.
* `SpkWithExpectedTxids` support added for `bdk_electrum` and `bdk_esplora`.
* `SyncRequestBuilder::expected_txids_of_spk` adds an association between `Txid` and `spk`.
* `SyncRequestExt::check_unconfirmed_statuses` adds unconfirmed transactions alongside their `spk`s during sync.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bitcoind_rpc/src/lib.rs
+++ b/crates/bitcoind_rpc/src/lib.rs
@@ -10,8 +10,9 @@
 #![warn(missing_docs)]
 
 use bdk_core::{BlockId, CheckPoint};
-use bitcoin::{block::Header, Block, BlockHash, Transaction};
+use bitcoin::{block::Header, Block, BlockHash, Transaction, Txid};
 use bitcoincore_rpc::bitcoincore_rpc_json;
+use std::collections::HashSet;
 
 pub mod bip158;
 
@@ -64,17 +65,19 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
         }
     }
 
-    /// Emit mempool transactions, alongside their first-seen unix timestamps.
+    /// Emit mempool transactions and capture the initial snapshot of all mempool [`Txid`]s.
     ///
-    /// This method emits each transaction only once, unless we cannot guarantee the transaction's
-    /// ancestors are already emitted.
+    /// This method returns a [`MempoolEvent`] containing the full transactions (with their
+    /// first-seen unix timestamps) that were emitted, and the set of all [`Txid`]s present from the
+    /// initial mempool query. Each transaction is emitted only once, unless we cannot guarantee the
+    /// transaction's ancestors are already emitted.
     ///
     /// To understand why, consider a receiver which filters transactions based on whether it
     /// alters the UTXO set of tracked script pubkeys. If an emitted mempool transaction spends a
     /// tracked UTXO which is confirmed at height `h`, but the receiver has only seen up to block
     /// of height `h-1`, we want to re-emit this transaction until the receiver has seen the block
     /// at height `h`.
-    pub fn mempool(&mut self) -> Result<Vec<(Transaction, u64)>, bitcoincore_rpc::Error> {
+    pub fn mempool(&mut self) -> Result<MempoolEvent, bitcoincore_rpc::Error> {
         let client = self.client;
 
         // This is the emitted tip height during the last mempool emission.
@@ -91,8 +94,11 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
         let prev_mempool_time = self.last_mempool_time;
         let mut latest_time = prev_mempool_time;
 
-        let txs_to_emit = client
-            .get_raw_mempool_verbose()?
+        // Get the raw mempool result from the RPC client.
+        let raw_mempool = client.get_raw_mempool_verbose()?;
+        let raw_mempool_txids = raw_mempool.keys().copied().collect::<HashSet<_>>();
+
+        let emitted_txs = raw_mempool
             .into_iter()
             .filter_map({
                 let latest_time = &mut latest_time;
@@ -128,7 +134,11 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
         self.last_mempool_time = latest_time;
         self.last_mempool_tip = Some(self.last_cp.height());
 
-        Ok(txs_to_emit)
+        Ok(MempoolEvent {
+            emitted_txs,
+            raw_mempool_txids,
+            last_seen: latest_time as u64,
+        })
     }
 
     /// Emit the next block height and header (if any).
@@ -141,6 +151,37 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
     pub fn next_block(&mut self) -> Result<Option<BlockEvent<Block>>, bitcoincore_rpc::Error> {
         Ok(poll(self, |hash| self.client.get_block(hash))?
             .map(|(checkpoint, block)| BlockEvent { block, checkpoint }))
+    }
+}
+
+/// A new emission from mempool.
+#[derive(Debug)]
+pub struct MempoolEvent {
+    /// Emitted mempool transactions with their first‐seen unix timestamps.
+    pub emitted_txs: Vec<(Transaction, u64)>,
+
+    /// Set of all [`Txid`]s from the raw mempool result, including transactions that may have been
+    /// confirmed or evicted during processing. This is used to determine which expected
+    /// transactions are missing.
+    pub raw_mempool_txids: HashSet<Txid>,
+
+    /// The latest first-seen epoch of emitted mempool transactions.
+    pub last_seen: u64,
+}
+
+impl MempoolEvent {
+    /// Given an iterator of expected [`Txid`]s, return those that are missing from the mempool.
+    pub fn evicted_txids(
+        &self,
+        expected_unconfirmed_txids: impl IntoIterator<Item = Txid>,
+    ) -> HashSet<Txid> {
+        let expected_set = expected_unconfirmed_txids
+            .into_iter()
+            .collect::<HashSet<_>>();
+        expected_set
+            .difference(&self.raw_mempool_txids)
+            .copied()
+            .collect()
     }
 }
 
--- a/crates/bitcoind_rpc/tests/test_emitter.rs
+++ b/crates/bitcoind_rpc/tests/test_emitter.rs
@@ -189,7 +189,7 @@ fn test_into_tx_graph() -> anyhow::Result<()> {
         assert!(emitter.next_block()?.is_none());
 
         let mempool_txs = emitter.mempool()?;
-        let indexed_additions = indexed_tx_graph.batch_insert_unconfirmed(mempool_txs);
+        let indexed_additions = indexed_tx_graph.batch_insert_unconfirmed(mempool_txs.emitted_txs);
         assert_eq!(
             indexed_additions
                 .tx_graph
@@ -437,6 +437,7 @@ fn mempool_avoids_re_emission() -> anyhow::Result<()> {
     // the first emission should include all transactions
     let emitted_txids = emitter
         .mempool()?
+        .emitted_txs
         .into_iter()
         .map(|(tx, _)| tx.compute_txid())
         .collect::<BTreeSet<Txid>>();
@@ -447,7 +448,7 @@ fn mempool_avoids_re_emission() -> anyhow::Result<()> {
 
     // second emission should be empty
     assert!(
-        emitter.mempool()?.is_empty(),
+        emitter.mempool()?.emitted_txs.is_empty(),
         "second emission should be empty"
     );
 
@@ -457,7 +458,7 @@ fn mempool_avoids_re_emission() -> anyhow::Result<()> {
     }
     while emitter.next_header()?.is_some() {}
     assert!(
-        emitter.mempool()?.is_empty(),
+        emitter.mempool()?.emitted_txs.is_empty(),
         "third emission, after chain tip is extended, should also be empty"
     );
 
@@ -506,6 +507,7 @@ fn mempool_re_emits_if_tx_introduction_height_not_reached() -> anyhow::Result<()
     assert_eq!(
         emitter
             .mempool()?
+            .emitted_txs
             .into_iter()
             .map(|(tx, _)| tx.compute_txid())
             .collect::<BTreeSet<_>>(),
@@ -515,6 +517,7 @@ fn mempool_re_emits_if_tx_introduction_height_not_reached() -> anyhow::Result<()
     assert_eq!(
         emitter
             .mempool()?
+            .emitted_txs
             .into_iter()
             .map(|(tx, _)| tx.compute_txid())
             .collect::<BTreeSet<_>>(),
@@ -535,6 +538,7 @@ fn mempool_re_emits_if_tx_introduction_height_not_reached() -> anyhow::Result<()
                 .collect::<BTreeSet<_>>();
             let emitted_txids = emitter
                 .mempool()?
+                .emitted_txs
                 .into_iter()
                 .map(|(tx, _)| tx.compute_txid())
                 .collect::<BTreeSet<_>>();
@@ -593,6 +597,7 @@ fn mempool_during_reorg() -> anyhow::Result<()> {
     assert_eq!(
         emitter
             .mempool()?
+            .emitted_txs
             .into_iter()
             .map(|(tx, _)| tx.compute_txid())
             .collect::<BTreeSet<_>>(),
@@ -628,6 +633,7 @@ fn mempool_during_reorg() -> anyhow::Result<()> {
             // include mempool txs introduced at reorg height or greater
             let mempool = emitter
                 .mempool()?
+                .emitted_txs
                 .into_iter()
                 .map(|(tx, _)| tx.compute_txid())
                 .collect::<BTreeSet<_>>();
@@ -643,6 +649,7 @@ fn mempool_during_reorg() -> anyhow::Result<()> {
 
             let mempool = emitter
                 .mempool()?
+                .emitted_txs
                 .into_iter()
                 .map(|(tx, _)| tx.compute_txid())
                 .collect::<BTreeSet<_>>();
@@ -766,7 +773,7 @@ fn test_expect_tx_evicted() -> anyhow::Result<()> {
     )?;
 
     let mut emitter = Emitter::new(env.rpc_client(), chain.tip(), 1);
-    let changeset = graph.batch_insert_unconfirmed(emitter.mempool()?);
+    let changeset = graph.batch_insert_unconfirmed(emitter.mempool()?.emitted_txs);
     assert!(changeset
         .tx_graph
         .txs
@@ -806,14 +813,15 @@ fn test_expect_tx_evicted() -> anyhow::Result<()> {
     // Send the tx.
     let txid_2 = core.send_raw_transaction(&tx1b)?;
 
-    // We evict the expected txs that are missing from mempool.
-    let exp_txids = graph
-        .expected_unconfirmed_spk_txids(&chain, chain_tip, ..)?
-        .collect::<Vec<_>>();
-    assert_eq!(exp_txids, vec![(txid_1, spk)]);
-    let mempool = emitter
-        .mempool()?
-        .into_iter()
+    // Retrieve the expected unconfirmed txids and spks from the graph.
+    let exp_spk_txids = graph.expected_unconfirmed_spk_txids(&chain, chain_tip, ..)?;
+    assert_eq!(exp_spk_txids, vec![(txid_1, spk)]);
+
+    // Check that mempool emission contains evicted txid.
+    let mempool_event = emitter.mempool()?;
+    let unseen_txids: Vec<Txid> = mempool_event
+        .emitted_txs
+        .iter()
         .map(|(tx, _)| tx.compute_txid())
         .collect();
     assert!(unseen_txids.contains(&txid_2));
@@ -825,6 +833,7 @@ fn test_expect_tx_evicted() -> anyhow::Result<()> {
         let _ = graph.insert_evicted_at(txid, seen_at);
     }
 
+    // tx1 should no longer be canonical.
     assert!(graph
         .graph()
         .list_canonical_txs(&chain, chain_tip)
--- a/crates/bitcoind_rpc/tests/test_emitter.rs
+++ b/crates/bitcoind_rpc/tests/test_emitter.rs
@@ -731,3 +731,106 @@ fn no_agreement_point() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn test_expect_tx_evicted() -> anyhow::Result<()> {
+    use bdk_bitcoind_rpc::bitcoincore_rpc::bitcoin;
+    use bdk_bitcoind_rpc::bitcoincore_rpc::bitcoincore_rpc_json::CreateRawTransactionInput;
+    use bdk_chain::miniscript;
+    use bdk_chain::spk_txout::SpkTxOutIndex;
+    use bdk_chain::ConfirmationBlockTime;
+    use bitcoin::constants::genesis_block;
+    use bitcoin::secp256k1::Secp256k1;
+    use bitcoin::Network;
+    use std::collections::HashMap;
+    let env = TestEnv::new()?;
+
+    let s = bdk_testenv::utils::DESCRIPTORS[0];
+    let desc = miniscript::Descriptor::parse_descriptor(&Secp256k1::new(), s)
+        .unwrap()
+        .0;
+    let spk = desc.at_derivation_index(0)?.script_pubkey();
+
+    let chain = LocalChain::from_genesis_hash(genesis_block(Network::Regtest).block_hash()).0;
+    let chain_tip = chain.tip().block_id();
+
+    let mut index = SpkTxOutIndex::default();
+    index.insert_spk(("external", 0u32), spk.clone());
+    let mut graph = IndexedTxGraph::<ConfirmationBlockTime, _>::new(index);
+
+    // Receive tx1.
+    let _ = env.mine_blocks(100, None)?;
+    let txid_1 = env.send(
+        &Address::from_script(&spk, Network::Regtest)?,
+        Amount::ONE_BTC,
+    )?;
+
+    let mut emitter = Emitter::new(env.rpc_client(), chain.tip(), 1);
+    let changeset = graph.batch_insert_unconfirmed(emitter.mempool()?);
+    assert!(changeset
+        .tx_graph
+        .txs
+        .iter()
+        .any(|tx| tx.compute_txid() == txid_1));
+    let seen_at = graph
+        .graph()
+        .get_tx_node(txid_1)
+        .unwrap()
+        .last_seen_unconfirmed
+        .unwrap();
+
+    // Double spend tx1.
+
+    // Get `prevout` from core.
+    let core = env.rpc_client();
+    let tx1 = &core.get_raw_transaction(&txid_1, None)?;
+    let txin = &tx1.input[0];
+    let op = txin.previous_output;
+
+    // Create `tx1b` using the previous output from tx1.
+    let utxo = CreateRawTransactionInput {
+        txid: op.txid,
+        vout: op.vout,
+        sequence: None,
+    };
+    let addr = core.get_new_address(None, None)?.assume_checked();
+    let tx = core.create_raw_transaction(
+        &[utxo],
+        &HashMap::from([(addr.to_string(), Amount::from_btc(49.99)?)]),
+        None,
+        None,
+    )?;
+    let res = core.sign_raw_transaction_with_wallet(&tx, None, None)?;
+    let tx1b = res.transaction()?;
+
+    // Send the tx.
+    let txid_2 = core.send_raw_transaction(&tx1b)?;
+
+    // Retrieve the expected unconfirmed txids and spks from the graph.
+    let exp_spk_txids = graph.expected_unconfirmed_spk_txids(&chain, chain_tip, ..)?;
+    assert_eq!(exp_spk_txids, vec![(txid_1, spk)]);
+
+    // Check that mempool emission contains evicted txid.
+    let mempool_event = emitter.mempool()?;
+    let unseen_txids: Vec<Txid> = mempool_event
+        .emitted_txs
+        .iter()
+        .map(|(tx, _)| tx.compute_txid())
+        .collect();
+    assert!(unseen_txids.contains(&txid_2));
+
+    // Update graph with evicted tx.
+    let exp_txids = exp_spk_txids.into_iter().map(|(txid, _)| txid);
+    let evicted_txids = mempool_event.evicted_txids(exp_txids);
+    for txid in evicted_txids {
+        let _ = graph.insert_evicted_at(txid, seen_at);
+    }
+
+    assert!(graph
+        .graph()
+        .list_canonical_txs(&chain, chain_tip)
+        .next()
+        .is_none());
+
+    Ok(())
+}
--- a/crates/bitcoind_rpc/tests/test_emitter.rs
+++ b/crates/bitcoind_rpc/tests/test_emitter.rs
@@ -806,15 +806,14 @@ fn test_expect_tx_evicted() -> anyhow::Result<()> {
     // Send the tx.
     let txid_2 = core.send_raw_transaction(&tx1b)?;
 
-    // Retrieve the expected unconfirmed txids and spks from the graph.
-    let exp_spk_txids = graph.expected_unconfirmed_spk_txids(&chain, chain_tip, ..)?;
-    assert_eq!(exp_spk_txids, vec![(txid_1, spk)]);
-
-    // Check that mempool emission contains evicted txid.
-    let mempool_event = emitter.mempool()?;
-    let unseen_txids: Vec<Txid> = mempool_event
-        .emitted_txs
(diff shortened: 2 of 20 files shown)


notmandatory:
Looks like only breaking changes are on `core` and `chain`, so we'll be able to roll this out in a non-breaking 1.x `wallet` release?

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -879,6 +881,20 @@ pub trait SyncRequestBuilderExt<K> {
 
     /// Add [`Script`](bitcoin::Script)s that are revealed by the `indexer` but currently unused.
     fn unused_spks_from_indexer(self, indexer: &KeychainTxOutIndex<K>) -> Self;
+
+    /// Add unconfirmed txids and their associated spks.
+    ///
+    /// We expect that the chain source should include these txids in their spk histories. If not,
+    /// the transaction has been evicted for some reason and we will inform the receiving
+    /// structures in the response.
```
evanlinjin: Update docs here.

crates/core/src/spk_client.rs
```diff
@@ -86,6 +86,14 @@ impl SyncProgress {
     }
 }
 
+/// [`Script`] with corresponding [`Txid`] histories.
+pub struct SpkWithHistory {
+    /// Script pubkey.
+    pub spk: ScriptBuf,
+    /// Txid history.
+    pub txids: HashSet<Txid>,
+}
```
evanlinjin: Maybe `SpkWithExpectedTxids` to be more explicit?

crates/core/src/spk_client.rs
```diff
@@ -301,6 +369,11 @@ impl<I> SyncRequest<I> {
         SyncIter::<I, ScriptBuf>::new(self)
     }
 
+    /// Iterate over [`ScriptBuf`]s with corresponding [`Txid`] histories contained in this request.
+    pub fn iter_spks_with_history(&mut self) -> impl ExactSizeIterator<Item = SpkWithHistory> + '_ {
```
evanlinjin: Rename to `iter_spks_with_expected_txids`

LLFourn:
@notmandatory core breaking change is also a wallet crate breaking change unfortunately. It is a minor breaking change (some struct from core is changing that no one inspects directly in their code probably) but it technically is a breaking change so it would have to be in bdk_wallet v2. This is a pretty important improvement so I'd say it justifies a v2 release by itself when it's merged.

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -879,6 +881,23 @@ pub trait SyncRequestBuilderExt<K> {
 
     /// Add [`Script`](bitcoin::Script)s that are revealed by the `indexer` but currently unused.
     fn unused_spks_from_indexer(self, indexer: &KeychainTxOutIndex<K>) -> Self;
+
```
LLFourn: :pick: I don't really see the attraction of a extension trait for this stuff. I think this should all be just member methods.

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -892,6 +911,29 @@ impl<K: Clone + Ord + core::fmt::Debug> SyncRequestBuilderExt<K> for SyncRequest
     fn unused_spks_from_indexer(self, indexer: &KeychainTxOutIndex<K>) -> Self {
         self.spks_with_indexes(indexer.unused_spks())
     }
+
+    fn check_for_missing_txs<A, C>(
+        self,
+        indexer: &KeychainTxOutIndex<K>,
+        canonical_iter: CanonicalIter<A, C>,
```
LLFourn: This seems extremely awkward. What about ChainOracles that are not infallible? Why does it need be an `KeychainTxOutIndex` -- it just calls `inner` in the implementation.

How about you just put this method on the tx graph. Like:

```rust
impl<I, D: AsRef<SpkTxOutIndex<I>>> IndexedTxGraph<D> {
    fn sync_request_add_expected_unconfirmed_txs<O: ChainOracle>(&self, sync_request: &mut SyncRequestBuilder, chain_oracle: &O, range: impl RangeBounds<I>) -> Result<(), O::Error> { ... }
}

impl<K> IndexedTxGraph<KeychainTxOutIndex<K>> {
    fn keychain_sync_request_add_expected_unconfirmed_txs<O: ChainOracle>(&self, sync_request: &mut SyncRequestBuilder, chain_oracle: &O, range: impl RangeBounds<K>) -> Result<(), O::Error> { ... }
}
```

crates/chain/src/indexer/spk_txout.rs
```diff
@@ -334,4 +334,21 @@ impl<I: Clone + Ord + core::fmt::Debug> SpkTxOutIndex<I> {
             .any(|output| self.spk_indices.contains_key(&output.script_pubkey));
         input_matches || output_matches
     }
+
+    /// Find relevant script pubkeys associated with a transaction.
```
LLFourn: :pick: doesn't define relevant here.

crates/chain/src/indexer/spk_txout.rs
```diff
@@ -334,4 +334,21 @@ impl<I: Clone + Ord + core::fmt::Debug> SpkTxOutIndex<I> {
...
+    ///
+    /// The script pubkeys of the transaction's outputs and previous outputs as scanned.
```
LLFourn: unhelpful sentence. Why. What do I use this method for.

Also needs to explain that inputs would not be detected as relevant unless the parents have been scanned.

crates/chain/src/tx_graph.rs
```diff
@@ -715,6 +717,30 @@ impl<A: Anchor> TxGraph<A> {
         changeset
     }
 
+    /// Inserts the given `missing_at` for `txid`
```
LLFourn: :wrench: there is no explanation of `missing_at`.

LLFourn (changes requested):
ConceptACK. Nice ideas here. I think chain API needs a clean up.

crates/chain/src/tx_graph.rs
```diff
@@ -145,6 +145,7 @@ pub struct TxGraph<A = ConfirmationBlockTime> {
     spends: BTreeMap<OutPoint, HashSet<Txid>>,
     anchors: HashMap<Txid, BTreeSet<A>>,
     last_seen: HashMap<Txid, u64>,
+    last_missing: HashMap<Txid, u64>,
```
ValuedMammal: I think it would be nice if we had a data structure like this and then `last_seen` would become `HashMap<Txid, LastSeen>`

```rust
#[derive(Debug, Clone, Copy, Default, PartialEq)]
struct LastSeen {
    /// seen at
    seen_at: u64,
    /// evicted at
    evicted_at: Option<u64>,
}
```

crates/chain/src/tx_graph.rs
```diff
@@ -110,19 +110,19 @@ use core::{
...
+            .floating_txouts()
+            .map(|(op, txo)| (op, txo.clone()))
+            .collect();
+        tx_update.anchors = graph
+            .anchors
+            .into_iter()
+            .flat_map(|(txid, anchors)| anchors.into_iter().map(move |a| (a, txid)))
+            .collect();
+        tx_update.seen_ats = graph.last_seen.into_iter().collect();
+        tx_update
```
ValuedMammal, line 155: It seems like `TxUpdate` would need another field `evicted_at` in order to maintain roundtrip convertibility between `TxUpdate` and `TxGraph`.

crates/chain/src/tx_graph.rs
```diff
@@ -715,6 +717,30 @@ impl<A: Anchor> TxGraph<A> {
         changeset
     }
 
+    /// Inserts the given `missing_at` for `txid`
+    pub fn insert_missing_at(&mut self, txid: Txid, missing_at: u64) -> ChangeSet<A> {
```
ValuedMammal: `insert_evicted_at`?

```suggestion
    pub fn insert_evicted_at(&mut self, txid: Txid, evicted_at: u64) -> ChangeSet<A> {
```

crates/chain/src/tx_graph.rs
```diff
@@ -1139,6 +1182,8 @@ pub struct ChangeSet<A = ()> {
     pub anchors: BTreeSet<(A, Txid)>,
     /// Added last-seen unix timestamps of transactions.
     pub last_seen: BTreeMap<Txid, u64>,
+    /// Added timestamps of when a transaction is last missing from the mempool.
+    pub last_missing: BTreeMap<Txid, u64>,
```
ValuedMammal: `last_evicted`?
```suggestion
    pub last_evicted: BTreeMap<Txid, u64>,
```

crates/chain/src/indexer/keychain_txout.rs
ValuedMammal: Building on this comment I put some ideas for a `IndexedTxGraph` method in evanlinjin/bdk#14

crates/chain/src/tx_graph.rs
evanlinjin: I quite like this idea since there has been talks about adding more timestamps.

I.e. `first-seen` timestamp, which will help with tx ordering when listing transactions.

```rust
struct MempoolTimestamps {
    pub first_seen: u64,
    pub last_seen: u64,
    pub last_evicted: Option<u64>,
}
```

I also like the term `evicted` as it's more descriptive than `missing`.

crates/chain/src/tx_graph.rs
evanlinjin: Do you think the convertibility between `TxGraph` and `TxUpdate` is important? I'm thinking maybe we should just get rid of it. One can just construct an empty `TxGraph` and apply an `TxUpdate`.

crates/chain/src/indexer/keychain_txout.rs
evanlinjin: Thanks @LLFourn, the infallible part was pure laziness.

I wanted to have the extension traits since `bdk_core::spk_client` types have a builder API. However, @ValuedMammal's has already provided a superior solution - I'm happy to go with that!

crates/chain/src/tx_graph.rs
ValuedMammal: It's probably ok to forego complete fungibility between graphs and updates, but at the same time I wonder how useful are the seen-ats without the corresponding evictions?

crates/chain/src/tx_graph.rs
```diff
@@ -969,9 +1007,14 @@ impl<A: Anchor> TxGraph<A> {
 
     /// List txids by descending last-seen order.
     ///
-    /// Transactions without last-seens are excluded.
-    pub fn txids_by_descending_last_seen(&self) -> impl ExactSizeIterator<Item = (u64, Txid)> + '_ {
-        self.txs_by_last_seen.iter().copied().rev()
+    /// Transactions without last-seens are excluded. Transactions with a last-missing timestamp
```
oleonardolima:
``` suggestion
    /// Transactions without last-seens are excluded. Transactions with a last-evicted timestamp
```

crates/chain/src/tx_graph.rs
```diff
@@ -1139,6 +1182,8 @@ pub struct ChangeSet<A = ()> {
     pub anchors: BTreeSet<(A, Txid)>,
     /// Added last-seen unix timestamps of transactions.
     pub last_seen: BTreeMap<Txid, u64>,
+    /// Added timestamps of when a transaction is last missing from the mempool.
```
oleonardolima:
``` suggestion
    /// Added timestamps of when a transaction is last evicted from the mempool.
```

crates/chain/src/tx_graph.rs
oleonardolima: I agree, and +1 on this idea, it makes it clear, and easy to "derive" its lifecycle and it's and can be useful info to expose to the users.

oleonardolima (commented):
A general ConceptACK, some small nits.

crates/chain/src/tx_graph.rs
```diff
@@ -765,6 +791,14 @@ impl<A: Anchor> TxGraph<A> {
                 changeset.merge(self.insert_seen_at(txid, seen_at));
             }
         }
+        for txid in update.missing {
```
oleonardolima: I think it'd help to update `apply_update_at` documentation in order to mention the `evicted_at` too, in [lines](https://github.com/bitcoindevkit/bdk/pull/1811/files#diff-edb82669f839091d04fe6d361bb8cf7ad7f738ba717663cb2735a5bc64f525cfR758-R770).

crates/chain/src/tx_graph.rs
LagginTimes: After some discussion with @evanlinjin, it was decided that this was best done in a separate PR as it may be too breaking of a change.

evanlinjin: Yes. On second thought, this introduces an invariant that `last_evicted` must exist only if we have had a `seen_at` value. Because of the existence of this invariant, this invariant must be reflected in `tx_graph::ChangeSet` by changing `ChangeSet::seen_at` to have values of `MempoolTimestamps`. This change breaks our policy of changes to changeset which was to only add new fields.

Edit: Maybe we can make it a user-facing API.

LagginTimes:
Closed, replaced by #1811.

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -731,3 +731,103 @@ fn no_agreement_point() -> anyhow::Result<()> {
...
+    // send it
+    let txid_2 = core.send_raw_transaction(&tx1b)?;
+
+    // We evict the expected txs that are missing from mempool
+    let exp_txids = graph
+        .expected_unconfirmed_spk_txids(&chain, ..)
+        .collect::<Vec<_>>();
+    assert_eq!(exp_txids, vec![(txid_1, spk)]);
+    let mempool = emitter
+        .mempool()?
```
evanlinjin: I'm wondering whether it will make sense to change the signature of `Emitter::mempool` to take in `expected_unconfirmed_txids` and output `TxUpdate`. This way the caller won't accidentally forget to check `expected_unconfirmed_txids` of the graph.

We can also take the opportunity to keep track of full txs that are already emitted and not re-emit them.

crates/chain/src/indexer/spk_txout.rs
```diff
@@ -334,4 +340,23 @@ impl<I: Clone + Ord + core::fmt::Debug> SpkTxOutIndex<I> {
             .any(|output| self.spk_indices.contains_key(&output.script_pubkey));
         input_matches || output_matches
     }
+
+    /// Find relevant script pubkeys associated with a transaction for tracking and validation.
+    ///
+    /// Returns a set of script pubkeys from [`SpkTxOutIndex`] that are pertinent to the outputs and
```
evanlinjin: `pertinent` is synonymous with `relevant` right? Can we just use `relevant` to be consistent and it's a much more common term.

crates/chain/src/tx_graph.rs
evanlinjin: > but at the same time I wonder how useful are the seen-ats without the corresponding evictions?

@ValuedMammal what do you mean by this?

I think it makes sense to completely remove convertibility of `TxGraph -> TxUpdate` and `TxUpdate -> TxGraph`.

ValuedMammal: Just pointing out that we lose information going from `TxGraph` to `TxUpdate`. Since both last-seen and last-evicted can influence canonicalization, it might make sense to include both (or neither).

Maybe it depends on the use case for `TxUpdate`. If updating a tx graph from a `spk_client` based chain source, the `seen_ats` don't seem to be as important compared to treating graphs and updates as interchangeable. If we remove convertibility, would it make sense to change this `From` impl to a method on `TxGraph` that returns a `TxUpdate`? We could potentially develop different kinds of updates, like one that is only relevant to a subset of keychains.

crates/bitcoind_rpc/tests/test_emitter.rs
ValuedMammal: I assumed it would be more of an opt-in feature, since it might feel like a significant behavior change to "eagerly evict" a tx the first time it appears missing.

crates/chain/src/tx_graph.rs
```diff
@@ -1110,6 +1192,51 @@ impl<A: Anchor> TxGraph<A> {
...
+                    )
+                    .flat_map(|op| match indexer.txout(op) {
+                        Some((i, txo)) if range.contains(i) => Some(txo.script_pubkey.clone()),
+                        _ => None,
+                    })
+                    .collect::<Vec<_>>();
+
+                core::iter::repeat(txid).zip(spks)
+            })
+    }
```
ValuedMammal, line 1235: To make it fallible I allocated a vec for the return value, not sure which way you prefer.

<details>

<summary>expand</summary>

```rust
pub fn expected_unconfirmed_spk_txids<'a, C, I>(
    &'a self,
    chain: &'a C,
    chain_tip: BlockId,
    indexer: &'a impl AsRef<SpkTxOutIndex<I>>,
    range: impl RangeBounds<I> + 'a,
) -> Result<impl Iterator<Item = (Txid, ScriptBuf)> + 'a, C::Error>
where
    C: ChainOracle,
    I: fmt::Debug + Clone + Ord + 'a,
{
    let mut spk_txs = vec![];
    for res in self.try_list_canonical_txs(chain, chain_tip) {
        let canon_tx = res?;
        if canon_tx.chain_position.is_confirmed() {
            continue;
        }
        let txid = canon_tx.tx_node.txid;
        let tx = canon_tx.tx_node.tx;
        let outpoints = tx.input.iter().map(|txin| txin.previous_output).chain(
            tx.output
                .iter()
                .enumerate()
                .map(|(vout, _)| OutPoint::new(txid, vout as u32)),
        );
        for op in outpoints {
            if let Some((index, txo)) = indexer.as_ref().txout(op) {
                if range.contains(index) {
                    spk_txs.push((txid, txo.script_pubkey.clone()));
                }
            }
        }
    }
    Ok(spk_txs.into_iter())
}
```

</details>

crates/bitcoind_rpc/tests/test_emitter.rs
evanlinjin: It would still be an opt-in feature though? If you provide an empty `expected_unconfirmed_txids` input, nothing would be marked as evicted.

ValuedMammal: Oh gotcha. I think that could work.

crates/esplora/tests/async_ext.rs
```diff
@@ -1,14 +1,146 @@
...
+    let _ = graph.index.insert_descriptor((), receiver_desc.clone())?;
+    let (chain, _) = LocalChain::from_genesis_hash(env.bitcoind.client.get_block_hash(0)?);
+
+    // Derive the receiving address from the descriptor.
+    let ((_, receiver_spk), _) = graph.index.reveal_next_spk(()).unwrap();
+    let receiver_addr = Address::from_script(&receiver_spk, bdk_chain::bitcoin::Network::Regtest)?;
+
+    env.mine_blocks(101, None)?;
+
+    // Select a UTXO to use as an input for constructing our test transactions.
```
evanlinjin, line 49: Which test transaction?

example-crates/example_bitcoind_rpc_polling/src/main.rs
```diff
@@ -256,7 +253,22 @@ fn main() -> anyhow::Result<()> {
...
+                                        .unwrap_or_else(|arc| (*arc).clone());
+                                    let seen_at = update
+                                        .seen_ats
+                                        .get(&tx.compute_txid())
+                                        .copied()
+                                        .unwrap_or(0);
+                                    (tx, seen_at)
+                                })
+                                .collect();
+                            tx.send(Emission::Mempool(mempool_txs))?;
```
evanlinjin: ~It will make more sense to change `Emission::Mempool` to contain a `TxUpdate`.~

Edit: `Emission::Mempool` should contain `MempoolEvent` instead. Refer to https://github.com/bitcoindevkit/bdk/pull/1811#discussion_r1956619412

example-crates/example_wallet_rpc/src/main.rs
```diff
@@ -130,7 +131,21 @@ fn main() -> anyhow::Result<()> {
...
+                let tx = std::sync::Arc::try_unwrap(arc_tx).unwrap_or_else(|arc| (*arc).clone());
+                let seen_at = update
+                    .seen_ats
+                    .get(&tx.compute_txid())
+                    .copied()
+                    .unwrap_or(0);
+                (tx, seen_at)
+            })
+            .collect();
+        sender.send(Emission::Mempool(mempool_txs))?;
```
evanlinjin: ~Same here. Change `Emission::Mempool` instead.~

Edit: `Emission::Mempool` should contain `MempoolEvent` instead. Refer to https://github.com/bitcoindevkit/bdk/pull/1811#discussion_r1956619412

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -91,44 +95,51 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
         self.last_mempool_tip = Some(self.last_cp.height());
 
-        Ok(txs_to_emit)
+        // After processing the mempool, any expected unconfirmed transactions that were not seen
+        // should be marked as evicted.
+        for expected_txid in expected_unconfirmed_txids {
+            if !tx_update.seen_ats.contains_key(&expected_txid) {
+                tx_update.evicted.insert(expected_txid);
+            }
+        }
```
evanlinjin: This is wrong as we skip including some mempool txs into `TxUpdate::txs` (check line 112).

Maybe create a `HashSet<Txid>` from the `.get_raw_mempool_verbose()` result directly to compare the `expected_unconfirmed_txids` against.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -74,8 +74,12 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
     /// tracked UTXO which is confirmed at height `h`, but the receiver has only seen up to block
     /// of height `h-1`, we want to re-emit this transaction until the receiver has seen the block
     /// at height `h`.
-    pub fn mempool(&mut self) -> Result<Vec<(Transaction, u64)>, bitcoincore_rpc::Error> {
+    pub fn mempool<A>(
+        &mut self,
+        expected_unconfirmed_txids: impl IntoIterator<Item = Txid>,
+    ) -> Result<TxUpdate<A>, bitcoincore_rpc::Error> {
```
evanlinjin: Okay I just realized that returning `TxUpdate` here is a mistake because when we apply `TxUpdate` we don't filter out irrelevant txs.

I think we should do the following:

`mempool()` should return a `MempoolEvent` struct.

```rust
pub struct MempoolEvent {
    pub unseen_txs: Vec<(Transaction, u64)>,
    pub all_txids: HashSet<Txid>,
}
```

We can remove the `expected_unconfirmed_txids` from the `mempool` method as we can use `MempoolEvent::all_txids` to find evicted txs (let's create a helper method on `MempoolEvent` to do this!

evanlinjin: Also the `A` generic is annoying here. Mempool txs wouldn't be anchored.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -91,44 +95,49 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
+            if tx_time > latest_time {
+                latest_time = tx_time;
+            }
 
-                    let tx = match client.get_raw_transaction(&txid, None) {
-                        Ok(tx) => tx,
-                        // the tx is confirmed or evicted since `get_raw_mempool_verbose`
-                        Err(err) if err.is_not_found_error() => return None,
-                        Err(err) => return Some(Err(err)),
-                    };
```
evanlinjin: No need to modify this now. Can you reduce the diff by not changing to a for loop?

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -806,23 +813,29 @@ fn test_expect_tx_evicted() -> anyhow::Result<()> {
...
-            let _ = graph.insert_evicted_at(txid, seen_at + 1);
-        }
+        .collect();
+    assert!(unseen_txids.contains(&txid_2));
+
+    // Update graph with evicted tx.
+    let exp_txids = exp_spk_txids.into_iter().map(|(txid, _)| txid);
+    let evicted_txids = mempool_event.evicted_txids(exp_txids);
+    for txid in evicted_txids {
+        let _ = graph.insert_evicted_at(txid, seen_at + 1);
```
evanlinjin: Why do we need the `+1`?

example-crates/example_bitcoind_rpc_polling/src/main.rs
```diff
@@ -287,7 +284,8 @@ fn main() -> anyhow::Result<()> {
                         (chain_changeset, graph_changeset)
                     }
                     Emission::Mempool(mempool_txs) => {
-                        let graph_changeset = graph.batch_insert_relevant_unconfirmed(mempool_txs);
+                        let graph_changeset =
+                            graph.batch_insert_relevant_unconfirmed(mempool_txs.emitted_txs);
```
evanlinjin: Could you modify this to check for evicted txs?

crates/chain/src/tx_graph.rs
```diff
@@ -1161,6 +1163,50 @@ impl<A: Anchor> TxGraph<A> {
...
+        range: impl RangeBounds<I> + 'a,
+    ) -> impl Iterator<Item = (Txid, ScriptBuf)> + 'a
+    where
+        C: ChainOracle<Error = core::convert::Infallible>,
+        I: fmt::Debug + Clone + Ord,
+    {
+        let chain_tip = chain.get_chain_tip().unwrap();
+
+        self.list_canonical_txs(chain, chain_tip)
+            .filter(|c| !c.chain_position.is_confirmed() && indexer.is_tx_relevant(&c.tx_node))
```
evanlinjin: Any confirmed tx that is missing from the Electrum's spk history should also be evicted. Since this means that there is a reorg and the tx is not re-introduced to the mempool (evicted).

crates/chain/src/tx_graph.rs
```diff
@@ -1164,48 +1163,46 @@ impl<A: Anchor> TxGraph<A> {
...
+    /// The spk range can be contrained with `range`.
+    pub fn expected_unconfirmed_spk_txids<'a, C, I>(
         &'a self,
         chain: &'a C,
-        indexer: &'a SpkTxOutIndex<I>,
+        chain_tip: BlockId,
+        indexer: &'a impl AsRef<SpkTxOutIndex<I>>,
         range: impl RangeBounds<I> + 'a,
-    ) -> impl Iterator<Item = (Txid, ScriptBuf)> + 'a
+    ) -> Result<impl Iterator<Item = (Txid, ScriptBuf)> + 'a, C::Error>
```
evanlinjin: There is no point returning an `impl Iterator` if it's just a `Vec`.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -144,6 +154,37 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
     }
 }
 
+/// A new emission from mempool.
+#[derive(Debug)]
+pub struct MempoolEvent {
+    /// Emitted mempool transactions with their first‐seen unix timestamps.
+    pub emitted_txs: Vec<(Transaction, u64)>,
```
evanlinjin, line 161:
```suggestion
    /// Unemitted transactions or transactions with ancestors that are unseen by the receiver.
    ///
    /// To understand the second condition, consider a receiver which filters transactions based on
    /// whether it alters the UTXO set of tracked script pubkeys. If an emitted mempool transaction
    /// spends a tracked UTXO which is confirmed at height `h`, but the receiver has only seen up to
    /// block of height `h-1`, we want to re-emit this transaction until the receiver has seen the
    /// block at height `h`.
    pub new_txs: Vec<(bitcoin::Transaction, u64)>,

```

"Emitted" is not descriptive as the whole `MempoolEvent` struct is emitted. Also I think we need to describe clearer what exactly is contained in this field.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -144,6 +154,37 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
+
+    /// Set of all [`Txid`]s from the raw mempool result, including transactions that may have been
+    /// confirmed or evicted during processing. This is used to determine which expected
+    /// transactions are missing.
+    pub raw_mempool_txids: HashSet<Txid>,
```
evanlinjin, line 166:
```suggestion
    /// Txids of all transactions currently present in the mempool.
    ///
    /// This is useful to determine whether transactions are evicted from the mempool.
    pub all_txids: HashSet<Txid>,
```

I think this is more concise and clearer.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -144,6 +154,37 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
+
+    /// The latest first-seen epoch of emitted mempool transactions.
+    pub last_seen: u64,
```
evanlinjin, line 169:
```suggestion
    /// The latest timestamp of when a transaction entered the mempool.
    ///
    /// This is useful for setting the timestamp for evicted transactions.
    pub latest_update_time: u64,
```

What do you think about this?

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -144,6 +154,37 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
+}
+
+impl MempoolEvent {
+    /// Given an iterator of expected [`Txid`]s, return those that are missing from the mempool.
+    pub fn evicted_txids(
+        &self,
+        expected_unconfirmed_txids: impl IntoIterator<Item = Txid>,
```
evanlinjin, line 176: For spk-based syncing we want to use all canonical txids (including confirmed) to determine evicted txs. The reason is because spk-based APIs returns both confirmed and unconfirmed under a given spk and that a missing previously-confirmed tx is also evicted from the mempool.

For rpc-based, we obviously want to only input unconfirmed txs as block (confirmed events) are handled separately...

We can add an `include_confirmed: bool` param on `TxGraph`/`IndexedTxGraph` methods that lists expected txs... but that does feel seen like it's easy to screw up.

Another idea is to compare the `MempoolEvent` txids to the previous `MempoolEvent` txids and the difference that does not end up in a block is evicted... However, when we initiate the `Emitter`, we still need the wallet state of unconfirmed txs as we no longer have all the data from the last `MempoolEvent` emission.

evanlinjin: Maybe it's best to combine those two ideas.

This way we don't have to get something from our receiving structures (`expected_unconfirmed_txids`) before applying the update.

evanlinjin:
Replaced by #1839

crates/bitcoind_rpc/src/lib.rs
ValuedMammal: I'm not sure that the entire txid history is strictly needed in the expected set.

> a missing previously-confirmed tx is also evicted from the mempool.

Perhaps, but maybe we're only concerned with reorgs below some number of confirmations? For a busy wallet the set of expected txids could be quite large which could hinder sync performance over time, considering that the normal fate of a transaction is to gain confirmations and never disappear from the blockchain.

ValuedMammal: > We can add an `include_confirmed: bool` param on `TxGraph`/`IndexedTxGraph` methods that lists expected txs... but that does feel seen like it's easy to screw up.

Or have a method that takes an iterator of canonical txs, that way the caller can filter them on a number of confirmations for example.

PR #1812, opened 2025-01-24
fix(wallet): use `map_keychain` in `Wallet::build_fee_bump`

### LINKED ISSUE #1807: [wallet] `build_fee_bump` fails to remove the change output for single descriptor wallet

ValuedMammal:
Previously we failed to remove the change output if the wallet has no internal keychain which caused tx building to fail at the new higher feerate. Fix this by mapping the internal keychain to the de-facto change keychain so that the drain output can be recalculated.

fixes #1807

### Changelog notice

Fixed

- wallet: Fixed an issue preventing `build_fee_bump` from re-targeting the drain value for some wallets

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -1676,7 +1676,7 @@ impl Wallet {
         if tx.output.len() > 1 {
             let mut change_index = None;
             for (index, txout) in tx.output.iter().enumerate() {
-                let change_keychain = KeychainKind::Internal;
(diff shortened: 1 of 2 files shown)


notmandatory (approved):
Thanks for catching this! 

ACK 515309e99fb92335a35ef994839a37a26f23ea69

evanlinjin (approved):
ACK 515309e99fb92335a35ef994839a37a26f23ea69

PR #1820, opened 2025-02-04
Have `bdk_electrum` take Electrum client by reference

tnull:
### Description

Previously, `BdkElectrumClient::new` would unnecessarily take an owned `electrum_client::Client` value, which rendered reusing the same client (i.e., connection) for other tasks impossible.

Here, we switch to have `BdkElectrumClient` take a `Deref` with target `ElectrumApi`, which allows to give the `electrum_client::Client` *by reference* to `BdkElectrumClient`.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -7,6 +7,7 @@ use bdk_core::{
 use electrum_client::{ElectrumApi, Error, HeaderNotification};
 use std::{
     collections::HashSet,
+    ops::Deref,
     sync::{Arc, Mutex},
 };
 
@@ -16,7 +17,10 @@ const CHAIN_SUFFIX_LENGTH: u32 = 8;
 /// Wrapper around an [`electrum_client::ElectrumApi`] which includes an internal in-memory
 /// transaction cache to avoid re-fetching already downloaded transactions.
 #[derive(Debug)]
-pub struct BdkElectrumClient<E> {
+pub struct BdkElectrumClient<E: Deref>
+where
+    E::Target: ElectrumApi,
+{
     /// The internal [`electrum_client::ElectrumApi`]
     pub inner: E,
     /// The transaction cache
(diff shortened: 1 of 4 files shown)


evanlinjin:
> which rendered reusing the same client (i.e., connection) for other tasks impossible.

This is not true. The `BdkElectrumClient::inner` field is public.

tnull:
> This is not true. The `BdkElectrumClient::inner` field is public.

Well, that's still very limiting, no? It has a simple `E: ElectrumApi` bound, so it's not `Clone`, for example. Yes, you could use that to make a request using the same client, but you can't for example, reuse the same client that would require moving the reference to another context?

evanlinjin:
I created #1825 as an alternative, less-breaking solution.

tnull:
> I created #1825 as an alternative, less-breaking solution.

I'm not sure I understand why it's considered less breaking (note the majority of the diff here is simply moving code)?

evanlinjin:
> I'm not sure I understand why it's considered less breaking (note the majority of the diff here is simply moving code)?

It's breaking because `Deref` does not have a blanket implementation for `impl<T> Deref for T { Target: T }`. In other words, this PR forces `BdkElectrumClient::inner` to be of a type that impls `Deref<Target: ElectrumApi>` which does not include all types that impls `ElectrumApi` (which was the old constraint).

Anyhow, I closed #1825 as I think https://github.com/bitcoindevkit/rust-electrum-client/pull/163 is the superior solution.

tnull:
> It's breaking because `Deref` does not have a blanket implementation for `impl<T> Deref for T { Target: T }`. In other words, this PR forces `BdkElectrumClient::inner` to be of a type that impls `Deref<Target: ElectrumApi>` which does not include all types that impls `ElectrumApi` (which was the old constraint).
> 
> Anyhow, I closed #1825 as I think [bitcoindevkit/rust-electrum-client#163](https://github.com/bitcoindevkit/rust-electrum-client/pull/163) is the superior solution.

Okay. I'm not sure adding that boilerplate is much preferable, but your choice. Going ahead and closing this.

PR #1823, opened 2025-02-05
Check foreign utxos are not local before inclusion

### LINKED ISSUE #29: [wallet] Document about extra validation when signing transactions including foreign utxos
The wallet should check that a foreign utxo is indeed foreign before proceeding to build a transaction with it.

See a related discord comment

https://discord.com/channels/753336465005608961/978744259693916230/1326206434144288821

### LINKED ISSUE #1819: [wallet] Validate inputs added with `add_foreign_utxo`
The wallet should check that a foreign utxo is indeed foreign before proceeding to build a transaction with it.

See a related discord comment

https://discord.com/channels/753336465005608961/978744259693916230/1326206434144288821

nymius:
### Description
Considering all foreign utxos must not be owned by wallet
The new check is added to ensure this invariant holds.

Before this change, foreign utxos could belong to the wallet, breaking the expected invariant, as exposed by [this test](https://github.com/bitcoindevkit/bdk/blob/cc15e5d316b0827c3b81e74d586b96c26c174a2e/crates/wallet/tests/wallet.rs#L1647-L1659) at cc15e5d316b0827c3b81e74d586b96c26c174a2e:
```rust
fn test_add_foreign_utxo_invalid_psbt_input() {
    let (mut wallet, _) = get_funded_wallet_wpkh();
    let outpoint = wallet.list_unspent().next().expect("must exist").outpoint; // The outpoint is extracted from the wallet itself!
    let foreign_utxo_satisfaction = wallet
        .public_descriptor(KeychainKind::External)
        .max_weight_to_satisfy()
        .unwrap();


    let result =
        builder.add_foreign_utxo(outpoint, psbt::Input::default(), foreign_utxo_satisfaction);
    assert!(matches!(result, Err(AddForeignUtxoError::MissingUtxo)));
}
```

Fixes bitcoindevkit/bdk_wallet#29

### Changelog notice

No API changes.

### Checklists

* [x] This pull request doesn't break the existing API

### DIFF
--- a/crates/wallet/src/wallet/tx_builder.rs
+++ b/crates/wallet/src/wallet/tx_builder.rs
@@ -351,8 +351,9 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
     ///
     /// This method returns errors in the following circumstances:
     ///
-    /// 1. The `psbt_input` does not contain a `witness_utxo` or `non_witness_utxo`.
-    /// 2. The data in `non_witness_utxo` does not match what is in `outpoint`.
+    /// 1. The provided outpoint is associated to a [`LocalOutput`].
+    /// 2. The `psbt_input` does not contain a `witness_utxo` or `non_witness_utxo`.
+    /// 3. The data in `non_witness_utxo` does not match what is in `outpoint`.
     ///
     /// Note unless you set [`only_witness_utxo`] any non-taproot `psbt_input` you pass to this
     /// method must have `non_witness_utxo` set otherwise you will get an error when [`finish`]
@@ -383,24 +384,27 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
         satisfaction_weight: Weight,
         sequence: Sequence,
     ) -> Result<&mut Self, AddForeignUtxoError> {
-        if psbt_input.witness_utxo.is_none() {
-            match psbt_input.non_witness_utxo.as_ref() {
-                Some(tx) => {
-                    if tx.compute_txid() != outpoint.txid {
-                        return Err(AddForeignUtxoError::InvalidTxid {
-                            input_txid: tx.compute_txid(),
-                            foreign_utxo: outpoint,
-                        });
-                    }
-                    if tx.output.len() <= outpoint.vout as usize {
-                        return Err(AddForeignUtxoError::InvalidOutpoint(outpoint));
-                    }
-                }
-                None => {
-                    return Err(AddForeignUtxoError::MissingUtxo);
+        let txout = match (&psbt_input.witness_utxo, &psbt_input.non_witness_utxo) {
+            (Some(ref txout), _) => Ok(txout.clone()),
+            (_, Some(tx)) => {
+                if tx.compute_txid() != outpoint.txid {
+                    Err(AddForeignUtxoError::InvalidTxid {
+                        input_txid: tx.compute_txid(),
+                        foreign_utxo: outpoint,
+                    })
+                } else {
+                    tx.tx_out(outpoint.vout as usize)
+                        .map_err(|_| AddForeignUtxoError::InvalidOutpoint(outpoint))
+                        .cloned()
                 }
             }
-        }
+            (_, _) => Err(AddForeignUtxoError::MissingUtxo),
+        }?;
+
+        // Avoid the inclusion of local utxos as foreign utxos
+        if self.wallet.is_mine(txout.script_pubkey) {
+            return Err(AddForeignUtxoError::NotForeignUtxo);
+        };
 
         if let Some(WeightedUtxo {
             utxo: Utxo::Local { .. },
@@ -731,6 +735,8 @@ pub enum AddForeignUtxoError {
     InvalidOutpoint(OutPoint),
     /// Foreign utxo missing witness_utxo or non_witness_utxo
     MissingUtxo,
+    /// UTxO is owned by wallet
+    NotForeignUtxo,
 }
 
 impl fmt::Display for AddForeignUtxoError {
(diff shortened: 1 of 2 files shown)


nymius:
As part of BDK review club for bitcoindevkit/bdk#1798 we talked about bitcoindevkit/bdk_wallet#29  and its fix candidate bitcoindevkit/bdk#1823 :
	
1. As bitcoindevkit/bdk#1823 introduces breaking changes, a compromise solution would be to add a silent check which works as a no-op in case transaction is found to be owned by the wallet. Also, one possible approach is to delay the introduction of the new `NotForeignUtxo` variant error (breaking change) until the release of the new major version, `2.0.0` with a `TODO` comment.
2. It wasn't clear the ownership of the transactions stored in `TxGraph`. The terms `canonical` and `relevant` were mentioned as a good source to get an idea of the topic from the documentation.
`Wallet::transactions` docs have a good explanation of this.
`TxGraph`'s transactions don't have to belong to the wallet necessarily, and `get_txout` shouldn't be considered a good source of truth to establish if an `Outpoint` can be unlocked by `Wallet` or not. `Wallet::is_mine` is the correct method for this.

nymius:
**Next steps based on review club**:
- [x] Use `Wallet::is_mine` method to check if UTxO is local or not.
- [x] Comment API breaking changes to avoid delaying the changes until next major version, but keep them around to be ready for it.
- [x] Make failure to add local UTxO as foreign `do nothing and return` logic.

nymius:
Applied suggestions in [above comment checklist](https://github.com/bitcoindevkit/bdk/pull/1823#issuecomment-2648181368) in commit 995df71bbfcda861b4e15257bd4d58a0edd35e77.

995df71bbfcda861b4e15257bd4d58a0edd35e77 will be squashed to 81a1cef1 once reviewed.

**TODO**:
- [x] Squash 995df71bbfcda861b4e15257bd4d58a0edd35e77 to 81a1cef1

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -374,24 +375,34 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
+                return Err(AddForeignUtxoError::InvalidTxid {
+                    input_txid: tx.compute_txid(),
+                    foreign_utxo: outpoint,
+                });
             }
-        }
+            if tx.output.len() <= outpoint.vout as usize {
+                return Err(AddForeignUtxoError::InvalidOutpoint(outpoint));
+            }
+
```
evanlinjin:
```suggestion
```

Already checked 2 lines below.

evanlinjin (commented):
I have no strong opinion about whether we should make this breaking/not-breaking.

Generally, I think it's easier for maintainers if we are allowed to break everything and tell everyone to upgrade. For BDK, I think as long as we make the persistence crates backwards-compatible we are good.

crates/wallet/src/wallet/tx_builder.rs
nymius: Thanks! I didn't see it

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -374,24 +375,31 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
-                        });
-                    }
-                    if tx.output.len() <= outpoint.vout as usize {
-                        return Err(AddForeignUtxoError::InvalidOutpoint(outpoint));
-                    }
-                }
-                None => {
-                    return Err(AddForeignUtxoError::MissingUtxo);
-                }
+        let script_pubkey = if let Some(ref txout) = psbt_input.witness_utxo {
```
ValuedMammal: Style nit, but I'm inclined to write this as a match statement with the primary goal of locating the txout.

```rust
let txout = match (&psbt_input.witness_utxo, &psbt_input.non_witness_utxo) { ... };
```

crates/wallet/src/wallet/tx_builder.rs
```diff
@@ -1098,4 +1110,21 @@ mod test {
...
+        let foreign_utxo_satisfaction = wallet
+            .public_descriptor(KeychainKind::External)
+            .max_weight_to_satisfy()
+            .unwrap();
+
+        let mut builder = wallet.build_tx();
+        let _ =
+            builder.add_foreign_utxo(outpoint, psbt::Input::default(), foreign_utxo_satisfaction);
+        // TODO(2.0.0): assert!(matches!(result, Err(AddForeignUtxoError::NotForeignUtxo)));
+        assert!(builder.params.utxos.is_empty());
```
ValuedMammal: I think you still need to provide the witness utxo if you want this to reach the `is_mine` check.

ValuedMammal:
> I have no strong opinion about whether we should make this breaking/not-breaking.
> 
> Generally, I think it's easier for maintainers if we are allowed to break everything and tell everyone to upgrade. For BDK, I think as long as we make the persistence crates backwards-compatible we are good.

I agree it's better to implement the appropriate fix than one that is potentially sub-optimal, so in that case I'd be in favor of adding the `NotForeignUtxo` error in this PR.

nymius:
I have rebased, added some test to check all execution paths and addressed the comments.

nymius:
Rebased again. CI errors are related to dependencies and not this PR code.

nymius:
Rebased onto b26ff891

nymius:
Rebased onto 362c3dc8

nymius:
Rebased onto 739b54ff

nymius:
Already commented on bitcoindevkit/bdk_wallet#29, but adding it here too for an easier context.

As discussed in today lib dev team call, the behavior explained above is something we should not constrain, and the issue mentioned in the [discord chat](https://discord.com/channels/753336465005608961/978744259693916230/1326206434144288821) should be controlled by the user, which makes this PR irrelevant for the issue.

PR #1825, opened 2025-02-06
Allow `BdkElectrumClient` to take client by reference

evanlinjin:
### Description

Previously, `BdkElectrumClient` has to take ownership of the internal electrum client.

This is an alternative of #1820. #1820 removes the ability to have `BdkElectrumClient` take ownership.

### Notes to the reviewers

Is the `C` generic necessary? It's there just in case a caller would like to use something other than `electrum_client::Client`. Is anyone using the `RawClient`s directly?

If all we want to for `BdkElectrumClient` to take in `&electrum_client::Client` we can just do the following in the `electrum_client` crate.

```rust
impl<E: ElectrumApi> ElectrumApi for &E { /* todo */ }
```

### Changelog notice

* Allow `BdkElectrumClient` to take in the internal client as reference or smart pointer.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -6,7 +6,9 @@ use bdk_core::{
 };
 use electrum_client::{ElectrumApi, Error, HeaderNotification};
 use std::{
+    borrow::Borrow,
     collections::HashSet,
+    marker::PhantomData,
     sync::{Arc, Mutex},
 };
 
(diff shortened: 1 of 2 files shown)


crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -6,7 +6,9 @@ use bdk_core::{
 };
 use electrum_client::{ElectrumApi, Error, HeaderNotification};
 use std::{
+    borrow::Borrow,
```
tnull, line 9: Could you clarify why `Borrow` would be preferable to `Deref` in this case, especially since it requires a bunch of additional boilerplate, etc?

evanlinjin: Because `Borrow` has a [blanket implementation](https://doc.rust-lang.org/std/convert/trait.AsRef.html#relation-to-borrow) so that any `T` impls `Borrow<T>`.

We still want to keep the ability to take a client by ownership imo.

evanlinjin:
Closing, as https://github.com/bitcoindevkit/rust-electrum-client/pull/163 is a superior solution.

PR #1826, opened 2025-02-06
deps(bdk_testenv)!: bump `electrsd` to `0.36.1`

oleonardolima:
partially addresses #1949

### Description

Upgrades the `electrsd` version from `0.28.0` to `0.36.1`, migrating from old `bitcoind` to `corepc-node` and using the `corepc-node_28_2` feature.
 
Updates the `bitcoind` client type for `TestEnv`, and its internal methods. Also, updates: bdk_bitcoind_rpc, bdk_esplora, bdk_electrum usage of `bdk_testenv` and `TestEnv`.

### Notes to the reviewers

As we are upgrading to the latest `electrsd`, we can drop for now the pinning for `home` in the MSRV step.

### Changelog notice

```
### Added
- deps(bdk_testenv): introduces `bitcoin and it's `bitcoin/rand-std` feature under `bdk_testenv/std`.
- test(bdk_bitcoind_rpc): introcuce `ClientExt` trait to use custom testing RPC client with current `bitcoincore-rpc` one.

### Changed
- deps(bdk_testenv)!: bump `electrsd` to `0.36.1`.
- deps(bdk_testenv): updates `electrsd` feature to `corepc-node_28_2`.
- test(bdk_bitcoind_rpc): update all tests to use new `ClientExt` trait when building the `Emitter`.
- test(bdk_chain): update all tests to use new `bdk_testenv`, `electrsd` and `corepc-node` APIs.
- test(bdk_esplora, bdk_electrum): update all tests to use new `bdk_testenv`, `electrsd` and `corepc-node` APIs.
- build(testenv): feature gate `mine_empty_blocks`, `reorg_empty_blocks` behind `std` feature.

```

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/bitcoind_rpc/src/lib.rs
+++ b/crates/bitcoind_rpc/src/lib.rs
@@ -399,7 +399,7 @@ impl BitcoindRpcErrorExt for bitcoincore_rpc::Error {
 #[cfg(test)]
 #[cfg_attr(coverage_nightly, coverage(off))]
 mod test {
-    use crate::{bitcoincore_rpc::RpcApi, Emitter, NO_EXPECTED_MEMPOOL_TXS};
+    use crate::{Emitter, NO_EXPECTED_MEMPOOL_TXS};
     use bdk_chain::local_chain::LocalChain;
     use bdk_testenv::{anyhow, TestEnv};
     use bitcoin::{hashes::Hash, Address, Amount, ScriptBuf, Txid, WScriptHash};
@@ -408,14 +408,15 @@ mod test {
     #[test]
     fn test_expected_mempool_txids_accumulate_and_remove() -> anyhow::Result<()> {
         let env = TestEnv::new()?;
-        let chain = LocalChain::from_genesis(env.rpc_client().get_block_hash(0)?).0;
+        let (chain, _) = LocalChain::from_genesis(env.genesis_hash()?);
         let chain_tip = chain.tip();
-        let mut emitter = Emitter::new(
-            env.rpc_client(),
-            chain_tip.clone(),
-            1,
-            NO_EXPECTED_MEMPOOL_TXS,
-        );
+
+        let rpc_client = bitcoincore_rpc::Client::new(
+            &env.bitcoind.rpc_url(),
+            bitcoincore_rpc::Auth::CookieFile(env.bitcoind.params.cookie_file.clone()),
+        )?;
+
+        let mut emitter = Emitter::new(&rpc_client, chain_tip.clone(), 1, NO_EXPECTED_MEMPOOL_TXS);
 
         env.mine_blocks(100, None)?;
         while emitter.next_block()?.is_some() {}
--- a/crates/bitcoind_rpc/tests/test_emitter.rs
+++ b/crates/bitcoind_rpc/tests/test_emitter.rs
@@ -7,9 +7,16 @@ use bdk_chain::{
     spk_txout::SpkTxOutIndex,
     Balance, BlockId, CanonicalizationParams, IndexedTxGraph, Merge,
 };
-use bdk_testenv::{anyhow, TestEnv};
-use bitcoin::{hashes::Hash, Block, Network, OutPoint, ScriptBuf, WScriptHash};
-use bitcoincore_rpc::RpcApi;
+use bdk_testenv::{
+    anyhow,
+    corepc_node::{Input, Output},
+    TestEnv,
+};
+use bitcoin::{hashes::Hash, Block, Network, ScriptBuf, WScriptHash};
+
+use crate::common::ClientExt;
+
+mod common;
 
 /// Ensure that blocks are emitted in order even after reorg.
 ///
@@ -20,21 +27,18 @@ use bitcoincore_rpc::RpcApi;
 #[test]
 pub fn test_sync_local_chain() -> anyhow::Result<()> {
     let env = TestEnv::new()?;
-    let network_tip = env.rpc_client().get_block_count()?;
-    let (mut local_chain, _) = LocalChain::from_genesis(env.rpc_client().get_block_hash(0)?);
-    let mut emitter = Emitter::new(
-        env.rpc_client(),
-        local_chain.tip(),
-        0,
-        NO_EXPECTED_MEMPOOL_TXS,
-    );
+    let network_tip = env.rpc_client().get_block_count()?.into_model().0;
+    let (mut local_chain, _) = LocalChain::from_genesis(env.genesis_hash()?);
+
+    let client = ClientExt::get_rpc_client(&env)?;
+    let mut emitter = Emitter::new(&client, local_chain.tip(), 0, NO_EXPECTED_MEMPOOL_TXS);
 
     // Mine some blocks and return the actual block hashes.
     // Because initializing `ElectrsD` already mines some blocks, we must include those too when
     // returning block hashes.
     let exp_hashes = {
         let mut hashes = (0..=network_tip)
-            .map(|height| env.rpc_client().get_block_hash(height))
+            .map(|height| env.get_block_hash(height))
             .collect::<Result<Vec<_>, _>>()?;
         hashes.extend(env.mine_blocks(101 - network_tip as usize, None)?);
         hashes
@@ -140,19 +144,24 @@ fn test_into_tx_graph() -> anyhow::Result<()> {
     let addr_0 = env
         .rpc_client()
         .get_new_address(None, None)?
+        .address()?
         .assume_checked();
+
     let addr_1 = env
         .rpc_client()
         .get_new_address(None, None)?
+        .address()?
         .assume_checked();
+
     let addr_2 = env
         .rpc_client()
         .get_new_address(None, None)?
+        .address()?
         .assume_checked();
 
     env.mine_blocks(101, None)?;
 
-    let (mut chain, _) = LocalChain::from_genesis(env.rpc_client().get_block_hash(0)?);
+    let (mut chain, _) = LocalChain::from_genesis(env.genesis_hash()?);
     let mut indexed_tx_graph = IndexedTxGraph::<BlockId, _>::new({
         let mut index = SpkTxOutIndex::<usize>::default();
         index.insert_spk(0, addr_0.script_pubkey());
@@ -161,7 +170,8 @@ fn test_into_tx_graph() -> anyhow::Result<()> {
         index
     });
 
-    let emitter = &mut Emitter::new(env.rpc_client(), chain.tip(), 0, NO_EXPECTED_MEMPOOL_TXS);
+    let client = ClientExt::get_rpc_client(&env)?;
+    let emitter = &mut Emitter::new(&client, chain.tip(), 0, NO_EXPECTED_MEMPOOL_TXS);
 
     while let Some(emission) = emitter.next_block()? {
         let height = emission.block_height();
@@ -174,16 +184,11 @@ fn test_into_tx_graph() -> anyhow::Result<()> {
     let exp_txids = {
         let mut txids = BTreeSet::new();
         for _ in 0..3 {
-            txids.insert(env.rpc_client().send_to_address(
-                &addr_0,
-                Amount::from_sat(10_000),
-                None,
-                None,
-                None,
-                None,
-                None,
-                None,
-            )?);
+            txids.insert(
+                env.rpc_client()
+                    .send_to_address(&addr_0, Amount::from_sat(10_000))?
+                    .txid()?,
+            );
         }
         txids
     };
@@ -210,7 +215,10 @@ fn test_into_tx_graph() -> anyhow::Result<()> {
 
     // mine a block that confirms the 3 txs
     let exp_block_hash = env.mine_blocks(1, None)?[0];
-    let exp_block_height = env.rpc_client().get_block_info(&exp_block_hash)?.height as u32;
+    let exp_block_height = env
+        .rpc_client()
+        .get_block_verbose_one(exp_block_hash)?
+        .height as u32;
     let exp_anchors = exp_txids
         .iter()
         .map({
@@ -250,9 +258,11 @@ fn ensure_block_emitted_after_reorg_is_at_reorg_height() -> anyhow::Result<()> {
     const CHAIN_TIP_HEIGHT: usize = 110;
 
     let env = TestEnv::new()?;
+
+    let client = ClientExt::get_rpc_client(&env)?;
     let mut emitter = Emitter::new(
-        env.rpc_client(),
-        CheckPoint::new(0, env.rpc_client().get_block_hash(0)?),
+        &client,
+        CheckPoint::new(0, env.genesis_hash()?),
         EMITTER_START_HEIGHT as _,
         NO_EXPECTED_MEMPOOL_TXS,
     );
@@ -325,9 +335,11 @@ fn tx_can_become_unconfirmed_after_reorg() -> anyhow::Result<()> {
     const SEND_AMOUNT: Amount = Amount::from_sat(10_000);
 
     let env = TestEnv::new()?;
+
+    let client = ClientExt::get_rpc_client(&env)?;
     let mut emitter = Emitter::new(
-        env.rpc_client(),
-        CheckPoint::new(0, env.rpc_client().get_block_hash(0)?),
+        &client,
+        CheckPoint::new(0, env.genesis_hash()?),
         0,
         NO_EXPECTED_MEMPOOL_TXS,
     );
@@ -336,12 +348,13 @@ fn tx_can_become_unconfirmed_after_reorg() -> anyhow::Result<()> {
     let addr_to_mine = env
         .rpc_client()
         .get_new_address(None, None)?
+        .address()?
         .assume_checked();
     let spk_to_track = ScriptBuf::new_p2wsh(&WScriptHash::all_zeros());
     let addr_to_track = Address::from_script(&spk_to_track, Network::Regtest)?;
 
     // setup receiver
-    let (mut recv_chain, _) = LocalChain::from_genesis(env.rpc_client().get_block_hash(0)?);
+    let (mut recv_chain, _) = LocalChain::from_genesis(env.genesis_hash()?);
     let mut recv_graph = IndexedTxGraph::<BlockId, _>::new({
         let mut recv_index = SpkTxOutIndex::default();
         recv_index.insert_spk((), spk_to_track.clone());
@@ -358,14 +371,16 @@ fn tx_can_become_unconfirmed_after_reorg() -> anyhow::Result<()> {
         // lock outputs that send to `addr_to_track`
         let outpoints_to_lock = env
             .rpc_client()
-            .get_transaction(&txid, None)?
-            .transaction()?
+            .get_transaction(txid)?
+            .into_model()?
+            .tx
             .output
             .into_iter()
             .enumerate()
             .filter(|(_, txo)| txo.script_pubkey == spk_to_track)
-            .map(|(vout, _)| OutPoint::new(txid, vout as _))
+            .map(|(vout, _)| (txid, vout as u32))
             .collect::<Vec<_>>();
+
         env.rpc_client().lock_unspent(&outpoints_to_lock)?;
 
         let _ = env.mine_blocks(1, None)?;
@@ -413,9 +428,11 @@ fn mempool_avoids_re_emission() -> anyhow::Result<()> {
     const MEMPOOL_TX_COUNT: usize = 2;
 
     let env = TestEnv::new()?;
+
+    let client = ClientExt::get_rpc_client(&env)?;
     let mut emitter = Emitter::new(
-        env.rpc_client(),
-        CheckPoint::new(0, env.rpc_client().get_block_hash(0)?),
+        &client,
+        CheckPoint::new(0, env.genesis_hash()?),
         0,
         NO_EXPECTED_MEMPOOL_TXS,
     );
@@ -424,6 +441,7 @@ fn mempool_avoids_re_emission() -> anyhow::Result<()> {
     let addr = env
         .rpc_client()
         .get_new_address(None, None)?
+        .address()?
         .assume_checked();
     env.mine_blocks(BLOCKS_TO_MINE, Some(addr.clone()))?;
     while emitter.next_block()?.is_some() {}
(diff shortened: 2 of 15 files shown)


oleonardolima:
It looks like I'll need to open a PR on corepc-client to add `get-block-template` RPC, then I'll get back to this one :(

I'll do the same in `rust-electrum-client` for the tests, but I'll do it to 0.30.x instead to prevent this issue from blocking the PR there.

ValuedMammal:
@oleonardolima I noticed that the [`getblocktemplate`](https://docs.rs/corepc-node/0.8.0/corepc_node/struct.Client.html#method.get_block_template) RPC is implemented in `corepc-client`.

oleonardolima:
> @oleonardolima I noticed that the [`getblocktemplate`](https://docs.rs/corepc-node/0.8.0/corepc_node/struct.Client.html#method.get_block_template) RPC is implemented in `corepc-client`.

Yes, it was added a few releases after I started this PR, but IIRC some other methods were missing too. I'll check that again.

ValuedMammal:
> Yes, it was added a few releases after I started this PR, but IIRC some other methods were missing too. I'll check that again.

Seems to be lacking good support for `get_raw_mempool_verbose`, at least for versions of Bitcoin Core later than 0.17.

oleonardolima:
Once #1988 lands, we won't be relying on `get_raw_mempool_verbose`; instead, we will use `get_raw_mempool`. It should unblock this one.

oleonardolima:
> Once #1988 lands, we won't be relying on `get_raw_mempool_verbose`; instead, we will use `get_raw_mempool`. It should unblock this one.

It appears that other methods are missing in the corepc-client at electrsd `0.34`. They're released on corepc-node 0.8, which will be included in the next release of electrsd. Will check if there are alternative methods for those.

oleonardolima:
> It appears that other methods are missing in the corepc-client at electrsd `0.34`. They're released on corepc-node 0.8, which will be included in the next release of electrsd. Will check if there are alternative methods for those.

As of the latest v0.35 release, it appears that all the necessary methods are implemented in `corepc-{node|client}`. I went ahead and updated the bdk-testenv and bdk-electrum/esplora crates.

It requires a more significant refactor and changes for the `bdk_bitcoind_rpc`, as the Emitter relies on the `bitcoincore-rpc` ' RpcApi' trait. I can cover it all in a single PR, but would like to know what you guys think as well.

cc @ValuedMammal @LagginTimes

ValuedMammal:
Replaying a [comment](https://discord.com/channels/753336465005608961/753336465005608965/1404834140167213148) from discord

> I also noticed that updating `electrsd` means potentially changing the Emitter to use `corepc` Client because of how it relates to testenv, etc. We have to make sure switching clients doesn't negatively impact behavior, performance and stuff

ValuedMammal:
We may be able to unblock this by having a way to create a new RPC client using the node credentials of a `TestEnv` instance but independent of the actual test client that is used by `electrsd`. oleonardolima/bdk#3

oleonardolima:
> We may be able to unblock this by having a way to create a new RPC client using the node credentials of a `TestEnv` instance but independent of the actual test client that is used by `electrsd`. [oleonardolima#3](https://github.com/oleonardolima/bdk/pull/3)

Cool, I'll take a look at your PR this morning.

oleonardolima:
> We may be able to unblock this by having a way to create a new RPC client using the node credentials of a `TestEnv` instance but independent of the actual test client that is used by `electrsd`. [oleonardolima#3](https://github.com/oleonardolima/bdk/pull/3)

Alright, but if I got it right, the idea would to still rely on `bitcoincore_rpc` API for `bdk_bitcoind_rpc` tests (which I think makes sense), right ? And when we land bdk_bitcoind_client that wouldn't an issue anymore, right

oleonardolima:
> re-posting from https://github.com/oleonardolima/bdk/pull/3#issuecomment-3572074616

Thanks for the suggestion, I'll cherry-pick this one and update the remaining tests.

oleonardolima:
Also, it looks like the latest one is v0.36.1, I'll give it a try on that one too.

ValuedMammal:
> Alright, but if I got it right, the idea would to still rely on `bitcoincore_rpc` API for `bdk_bitcoind_rpc` tests (which I think makes sense), right ? And when we land bdk_bitcoind_client that wouldn't an issue anymore, right

The approach I had in mind is to bump `electrsd` to the latest version in `bdk_testenv` which might entail updating some methods in order to benefit from all new `corepc` types. In `bdk_bitcoind_rpc` we can start using the `TestEnv` client for tests, and this `ClientExt` trait or whatever we're calling it can be used to create a client just for the `Emitter` with no other code changes needed. When `bdk_bitcoind_client` is available we'll update the `Emitter` to use that and remove `bitcoincore_rpc`. Don't know if I've fully answered the question.

oleonardolima:
AFAICT the only blocker now is: https://github.com/rust-bitcoin/corepc/issues/425

oleonardolima:
> AFAICT the only blocker now is: [rust-bitcoin/corepc#425](https://github.com/rust-bitcoin/corepc/issues/425)

It's not a block anymore, by using the latest `electrsd`, the corepc-types are already fixed as it was suggested in the issue.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -408,10 +408,17 @@ mod test {
     #[test]
     fn test_expected_mempool_txids_accumulate_and_remove() -> anyhow::Result<()> {
         let env = TestEnv::new()?;
-        let chain = LocalChain::from_genesis(env.rpc_client().get_block_hash(0)?).0;
+        let (chain, _) = LocalChain::from_genesis(env.genesis_hash()?);
         let chain_tip = chain.tip();
+
+        // TODO: (@oleonardolima) We should use the `ClientExt` trait instead.
```
luisschwab: Is this currently possible or do we need the new `bdk_bitcoind_rpc` client ready?

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -78,31 +77,39 @@ fn relevant_conflicts() -> anyhow::Result<()> {
             env.mine_blocks(1, Some(sender_addr.clone()))?;
             env.mine_blocks(101, None)?;
 
+            // TODO: (@oleonardolima) remove unwraps
```
luisschwab: Lingering TODO here

crates/chain/tests/test_indexed_tx_graph.rs
luisschwab, line 64: Thought: why not drop `TestEnv::rpc_client` altogether and rely only on `ClientExt::get_rpc_client(&env)`?

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -118,31 +125,53 @@ fn relevant_conflicts() -> anyhow::Result<()> {
...
+            // })?     .map(
+            //         |txid| -> Result<
+            //             GetRawTransaction,
+            //             bdk_testenv::corepc_client::client_sync::Error,
+            //         > { client.get_raw_transaction(*txid) },
+            //     )
+            //     .collect::<Vec<Result<_, _>>>()
+            //     .map(|get_raw_tx| get_raw_tx.into_model())
+            //     .map(|tx| (tx.0, 0))
+            //     .collect::<Vec<(Transaction, u64)>>();
```
luisschwab: Needs resolving

crates/testenv/Cargo.toml
luisschwab, line 3: Need to bump `bdk_testenv`'s version as well

luisschwab:
Tests are failing locally for me:
```
cargo test -p bdk_chain --all-features
   Compiling bdk_testenv v0.13.1 (/Users/luisschwab/git/bdk/crates/testenv)
error[E0432]: unresolved import `bdk_chain::bitcoin::secp256k1::rand`
   --> crates/testenv/src/lib.rs:8:20
    |
8   | ...   secp256k1::rand::random, transaction, Address, Amount, Block, BlockHas...
    |                  ^^^^ could not find `rand` in `secp256k1`
    |
note: found an item that was configured out
   --> /Users/luisschwab/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/secp256k1-0.29.1/src/lib.rs:186:9
    |
186 | pub use rand;
    |         ^^^^
note: the item is gated behind the `rand` feature
   --> /Users/luisschwab/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/secp256k1-0.29.1/src/lib.rs:185:7
    |
185 | #[cfg(feature = "rand")]
    |       ^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0432`.
error: could not compile `bdk_testenv` (lib) due to 1 previous error
error: Recipe `_test-chain` failed on line 41 with exit code 101
error: Recipe `test` failed on line 30 with exit code 101
```

crates/testenv/Cargo.toml
oleonardolima: It's only bumped in the release PRs.

crates/bitcoind_rpc/src/lib.rs
oleonardolima: It's already possible, lingering TODO.

crates/chain/tests/test_indexed_tx_graph.rs
oleonardolima: Already solved those, forgot to remove the comment.

crates/chain/tests/test_indexed_tx_graph.rs
oleonardolima: It's solved, forgot to remove the comment.

crates/chain/tests/test_indexed_tx_graph.rs
oleonardolima: It's the standard-way to use testenv's rpc client, the `ClientExt` is just a workaround to upgrade the `bdk_testenv` in `bdk_bitcoind_rpc` while we don't have the new client that relies on the same corepc types.

oleonardolima:
> Tests are failing locally for me:
> 
> ```
> cargo test -p bdk_chain --all-features
>    Compiling bdk_testenv v0.13.1 (/Users/luisschwab/git/bdk/crates/testenv)
> error[E0432]: unresolved import `bdk_chain::bitcoin::secp256k1::rand`
>    --> crates/testenv/src/lib.rs:8:20
>     |
> 8   | ...   secp256k1::rand::random, transaction, Address, Amount, Block, BlockHas...
>     |                  ^^^^ could not find `rand` in `secp256k1`
>     |
> note: found an item that was configured out
>    --> /Users/luisschwab/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/secp256k1-0.29.1/src/lib.rs:186:9
>     |
> 186 | pub use rand;
>     |         ^^^^
> note: the item is gated behind the `rand` feature
>    --> /Users/luisschwab/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/secp256k1-0.29.1/src/lib.rs:185:7
>     |
> 185 | #[cfg(feature = "rand")]
>     |       ^^^^^^^^^^^^^^^^
> 
> For more information about this error, try `rustc --explain E0432`.
> error: could not compile `bdk_testenv` (lib) due to 1 previous error
> error: Recipe `_test-chain` failed on line 41 with exit code 101
> error: Recipe `test` failed on line 30 with exit code 101
> ```

Indeed, needs debugging. Still unsure why it's now failing.

luisschwab:
I think you need to import `bitcoin` with the `rand` feature so it re-exports `secp256k1` with it: https://docs.rs/crate/bitcoin/latest/features#rand

crates/testenv/src/lib.rs
```diff
@@ -238,18 +239,16 @@ impl TestEnv {
...
-                None => break,
-            }
+            let prev_hash = self.bitcoind.client.get_block(hash)?.header.prev_blockhash;
+            self.bitcoind.client.invalidate_block(hash)?;
+            hash = prev_hash
+            // TODO: (@leonardo) It requires a double check if there is any side-effect with this
+            // break removal. match prev_hash {
+            //     Some(prev_hash) => hash = prev_hash,
+            //     None => break,
+            // }
```
ValuedMammal: It seems that the `break` is necessary in case `count` is greater than the current block height, we'd end up trying to `get_block` with an all-zeros `hash` which doesn't exist. Another idea is to move `get_best_block_hash` to inside the loop and call `invalidate_block` on the resulting hash.

crates/testenv/src/lib.rs
```diff
@@ -301,24 +301,57 @@ impl TestEnv {
                 .client
                 .get_block_hash(height as u64)
                 .ok()
-                .map(|hash| (height, hash))
+                .map(|get_block_hash| {
+                    let hash = get_block_hash
+                        .block_hash()
+                        .expect("should `successfully convert to `BlockHash` from `GetBlockHash`");
+                    (height, hash)
+                })
```
ValuedMammal: As a convenience you could implement `get_block_hash` on `TestEnv` and use it in the implementation of `make_checkpoint_tip`.

```rust
/// Get block hash by `height` from the bitcoind client.
pub fn get_block_hash(&self, height: u32) -> anyhow::Result<BlockHash> { todo!() }
```

oleonardolima:
> I think you need to import `bitcoin` with the `rand` feature so it re-exports `secp256k1` with it: https://docs.rs/crate/bitcoin/latest/features#rand

It's weird, because it's not a new import, but it does not fail on master though.

crates/bitcoind_rpc/src/lib.rs
oleonardolima: Actually, the way I did we shouldn't use the `ClientExt` here, as I've defined as common module under tests. It was just a lingering TODO, removed it.

luisschwab:
Almost there, we just need #2099 merged to fix the `no_std` issue on CI.

oleonardolima:
> Almost there, we just need #2099 merged to fix the `no_std` issue on CI.

Yep, I'm merging it and will add VM's suggestions.

crates/bitcoind_rpc/tests/test_emitter.rs
ValuedMammal, line 1: In `test_emitter.rs`:

- ⛏️  I think it's fine to continue using the corepc client for various RPC calls in the tests, and only use `ClientExt` when creating the `Emitter`. This will also help to minimize the dependency on `bitcoincore_rpc`.

- ⛏️  We're passing `&mut bitcoincore_rpc_client` to `Emitter::new` in a few places, where an immutable `&` reference should be fine.

- ❓ In `tx_can_become_unconfirmed_after_reorg` I don't see the point of having `lock_unspent` here (Bitcoin Core can't spend from `addr_to_track` anyway), but maybe I missed something.

crates/testenv/src/lib.rs
oleonardolima: Good suggestion, thanks!

crates/testenv/src/lib.rs
oleonardolima: I updated it to have the same behavior as before.

crates/electrum/tests/test_electrum.rs
```diff
@@ -206,43 +216,47 @@ pub fn chained_mempool_tx_sync() -> anyhow::Result<()> {
...
-        .get_new_address(None, None)?
-        .require_network(Network::Regtest)?;
+    let tracked_addr = rpc_client.new_address()?;
 
     env.mine_blocks(100, None)?;
 
     // First unconfirmed tx.
     let txid1 = env.send(&tracked_addr, Amount::from_btc(1.0)?)?;
 
+    let list_unspent = env.list_unspent()?;
```
ValuedMammal: You can use `get_raw_transaction` to get `txid1` from the mempool and get the utxo the same way by iterating over the outputs to find the one whose SPK matches the `tracked_addr`.

```suggestion
    let tx = rpc_client.get_raw_transaction(txid1)?.transaction()?;
```

crates/electrum/tests/test_electrum.rs
```diff
@@ -792,42 +780,43 @@ fn test_check_fee_calculation() -> anyhow::Result<()> {
...
     // Look at the tx we just sent, it should have 1 input and 1 output
-    let tx = env
-        .rpc_client()
-        .get_raw_transaction_info(&txid, block_hash.as_ref())?;
-    assert_eq!(tx.vin.len(), 1);
-    assert_eq!(tx.vout.len(), 1);
-    let vin = &tx.vin[0];
-    let prev_txid = vin.txid.unwrap();
-    let vout = vin.vout.unwrap();
+    let tx = env.get_raw_transaction_verbose(&txid, &block_hash)?;
```
ValuedMammal: You can use `get_transaction` here without the additional `block_hash` parameter, since this is a transaction in Core's wallet.

```suggestion
    let tx = env.rpc_client().get_transaction(txid)?.into_model()?.tx;
```

ValuedMammal: Or use `rpc_client().get_raw_transaction()` and add `-txindex=1` to the default testenv [`Config`](https://github.com/bitcoindevkit/bdk/blob/d0eaadff02535b24b6a35e9d6d2c6df5505b4a0c/crates/testenv/src/lib.rs#L46).

crates/testenv/src/lib.rs
ValuedMammal, line 1: My only comment is to avoid implementing new RPC calls in `bdk_testenv` (e.g. `list_unspent` and `get_raw_transaction_verbose`) if an alternative can be found. Only reason is to keep the API lean and only proxy calls to the inner `bitcoind.client`. The RPCs with parameters can always be upstreamed to `corepc`. Not saying it's wrong, just more to maintain.

oleonardolima: I didn't want to change the RPC's being used in the tests at first, and added the new methods just to not have to wait for a new release from corepc + electrsd.

I'm fine with using the other RPCs and adding -txindex to the node config, if you guys are onboard with it.

ValuedMammal:
> > I think you need to import `bitcoin` with the `rand` feature so it re-exports `secp256k1` with it: https://docs.rs/crate/bitcoin/latest/features#rand
> 
> It's weird, because it's not a new import, but it does not fail on master though.

The old `electrsd` 0.28 may have been pulling in `rand` indirectly via `bitcoincore_rpc`. I was going to suggest making the `std` feature enable `bitcoin/rand-std` and feature-gating the methods that use it.

> `testenv/Cargo.toml`

```toml
[dependencies]
bitcoin = { version = "0.32.8", default-features = false }

[features]
std = ["bdk_chain/std", "bitcoin/rand-std"]
```

oleonardolima:
> > > I think you need to import `bitcoin` with the `rand` feature so it re-exports `secp256k1` with it: https://docs.rs/crate/bitcoin/latest/features#rand
> > 
> > 
> > It's weird, because it's not a new import, but it does not fail on master though.
> 
> The old `electrsd` 0.28 may have been pulling in `rand` indirectly via `bitcoincore_rpc`. I was going to suggest making the `std` feature enable `bitcoin/rand-std` and feature-gating the methods that use it.
> 
> > `testenv/Cargo.toml`
> 
> ```toml
> [dependencies]
> bitcoin = { version = "0.32.8", default-features = false }
> 
> [features]
> std = ["bdk_chain/std", "bitcoin/rand-std"]
> ```

Yes, the old `bitcoind` was using it from the `bitcoincore-rpc`.

crates/electrum/tests/test_electrum.rs
oleonardolima: I updated it to use the `get_raw_transaction` instead.

crates/electrum/tests/test_electrum.rs
oleonardolima: I updated it to use the `get_transaction` instead.

crates/testenv/src/lib.rs
oleonardolima: Alright, I've updated it in the latest commits.

crates/bitcoind_rpc/tests/test_emitter.rs
oleonardolima: I'll address the first two points, the last one can be handled in follow-up.

oleonardolima: It's now updated.

crates/esplora/src/async_ext.rs
```diff
@@ -577,7 +577,7 @@ mod test {
         let env = TestEnv::new()?;
         let base_url = format!("http://{}", &env.electrsd.esplora_url.clone().unwrap());
         let client = Builder::new(base_url.as_str()).build_async()?;
-        let initial_height = env.rpc_client().get_block_count()? as u32;
+        let initial_height = env.rpc_client().get_block_count()?.into_model().0 as u32;
```
ValuedMammal, line 580: 📌 We could have a `TestEnv::get_block_count` that returns `Result<u32>`. Not a blocker for this PR though.

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -118,31 +124,40 @@ fn relevant_conflicts() -> anyhow::Result<()> {
...
+                .iter()
+                .map(|txid| {
+                    client
+                        .get_raw_transaction(*txid)
+                        .unwrap()
+                        .into_model()
+                        .unwrap()
+                })
+                .map(|get_raw_transaction| (get_raw_transaction.0, 0))
+                .collect::<Vec<_>>();
```
ValuedMammal: Maybe a slightly nicer way to express the same idea

```suggestion
            let unconfirmed_txs: Vec<(Transaction, u64)> = mempool_txids
                .iter()
                .map(|txid| -> anyhow::Result<_> {
                    let tx = client.get_raw_transaction(*txid)?.transaction()?;
                    Ok((tx, 0))
                })
                .collect::<Result<Vec<_>, _>>()?;
```

ValuedMammal (commented):
I left a few non-blocking comments and also added 2 commits in oleonardolima/bdk#4. Let me know what you think.

crates/esplora/src/async_ext.rs
oleonardolima: I'll leave this for a follow-up.

crates/chain/tests/test_indexed_tx_graph.rs
oleonardolima: I've applied it into b78a9dedf8be14aef09d15ba2ad362d268b742c7

oleonardolima:
> I left a few non-blocking comments and also added 2 commits in [oleonardolima#4](https://github.com/oleonardolima/bdk/pull/4). Let me know what you think.

Thanks, those looked good, I cherry-picked the commits.

ValuedMammal (approved):
ACK 303cc7010006dcb546d92d8a4c8c52e43ad14ead

luisschwab (approved):
ACK 303cc7010006dcb546d92d8a4c8c52e43ad14ead

crates/esplora/src/async_ext.rs
oleonardolima: https://github.com/oleonardolima/bdk/commit/2ebe82dee94ff34eb0104c7c56e6cc0a20f5d234

PR #1828, opened 2025-02-09
Fix SQLite panic when syncing many UTXOs

### LINKED ISSUE #1827: [bdk 0.29] Syncing a Wallet that received many small UTXOs panics with SQLite error
**Describe the bug**  

Syncing a Wallet (`bdk` crate version 0.29) that received many small UTXOs (50 TX with 700 UTXOs each) fails with the error "SQLite error: database is locked".  This was originally found on a mobile `bdk` app using `bdk-ffi` but can also be reproduced with only the rust version of `bdk`. 

**To Reproduce**  

To reproduce the issue send 50 transactions that have 700 UTXOs each to the target wallet (total of 35_000 UTXOs), the UTXO values can be small (eg. 2715 sats each). See the below[^1] unit test which does the following:

1. fund bdk wallet 1 from regtest node coinbase txs (5 BTC)
2. bdk wallet 1 creates 50 tx with 700 utxos each and sends them back to itself
3. bdk wallet 2 receives 50 tx with 700 utxos each from wallet 1
4. sync wallet 2 <-- PANICS with "SQLite error: database is locked" error

Note: bdk wallet 1 uses a memory DB which does not exhibit this issue. The reason the original UTXOs are created in bdk wallet 1 is because the `rust-bitcoincore-rpc` library does not support the `sendmany` RPC call. 

**Expected behavior**  

Syncing a large number of utxos should not cause a panic. It should also sync much faster.

**Build environment**  
 - BDK tag/commit: **v0.29.0**
 - OS+version: macOS (but also seen on Android, iOS)
 - Rust/Cargo version: 1.84.1
 - Rust/Cargo target: macOS

**Additional context**  

This was found by a user with a production mobile app using bdk-ffi bindings based on bdk 0.29. It should be considered a high priority to at least find a workaround so the user can sync their wallet. 

I was able to get the repro test[^1] to complete without a panic with a small change[^2] to `impl BatchDatabase for SqliteDatabase`. This work around may not be safe since it removes the `begin_batch`/`commit_batch` semantics currently used during syncing. On the other hand I don't think any of the individual steps in syncing would need to be rolled back if one step fails. This approach needs more analysis. Even with this change, syncing the repro test[^1] takes ~1 hr on a desktop system.

Once fixed this scenario should also be tested with `bdk-ffi` and on iOS and Android example apps.

The repro test[^1] takes ~1 hr on a desktop system so it shouldn't be added as a normal unit tests. 

This scenario should also be tested with bdk 1.0.

[^1]: This test added to the `blockchain::electrum::test` module reproduces the panic. 
    
    ```rust
    #[test]
        fn test_electrum_large_num_utxos() {
            env_logger::init();
            let mut test_client = TestClient::default();
            let electrum_blockchain =
                ElectrumBlockchain::from(Client::new(&test_client.electrsd.electrum_url).unwrap());
    
            // fund bdk wallet 1 with regtest node coinbase txs
            let mem_db = MemoryDatabase::new();
            let wallet1_descriptor = "wpkh(tprv8i8F4EhYDMquzqiecEX8SKYMXqfmmb1Sm7deoA1Hokxzn281XgTkwsd6gL8aJevLE4aJugfVf9MKMvrcRvPawGMenqMBA3bRRfp4s1V7Eg3/0/*)";
            let wallet1 =
                Wallet::new(wallet1_descriptor, None, bitcoin::Network::Regtest, mem_db).unwrap();
            let wallet1_address = wallet1.get_address(AddressIndex::New).unwrap().address;
            test_client
                .send_to_address(
                    &wallet1_address,
                    Amount::from_btc(5.0).unwrap(),
                    None,
                    None,
                    None,
                    None,
                    None,
                    None,
                )
                .unwrap();
            test_client.generate(1, None);
            wallet1
                .sync(&electrum_blockchain, Default::default())
                .unwrap();
            assert_eq!(wallet1.get_balance().unwrap().confirmed, 5_0000_0000);
            // bdk wallet 1 creates 50 tx * 700 utxos and sends them back to itself
            for _ in 0..50 {
                let amount = 2715;
                let address_amounts = (0..700)
                    .map(|_| {
                        (
                            wallet1
                                .get_address(AddressIndex::New)
                                .unwrap()
                                .address
                                .script_pubkey(),
                            amount,
                        )
                    })
                    .collect::<Vec<_>>();
                let mut tx_builder = wallet1.build_tx();
                // only allow spending utxos greater than 2715 sats
                let unspendable = wallet1
                    .list_unspent()
                    .unwrap()
                    .iter()
                    .filter(|utxo| utxo.txout.value <= amount)
                    .map(|utxo| utxo.outpoint)
                    .collect::<Vec<_>>();
                tx_builder
                    .set_recipients(address_amounts)
                    .unspendable(unspendable);
                let (mut psbt, details) = tx_builder.finish().unwrap();
                assert!(wallet1.sign(&mut psbt, SignOptions::default()).unwrap());
                let tx = psbt.extract_tx();
                electrum_blockchain.broadcast(&tx).unwrap();
                // include test txs in a block
                test_client.generate(1, None);
                wallet1
                    .sync(&electrum_blockchain, Default::default())
                    .unwrap()
            }
            assert_eq!(
                50 * 700,
                wallet1
                    .list_unspent()
                    .unwrap()
                    .iter()
                    .filter(|utxo| utxo.txout.value == 2715)
                    .count()
            );
    
            // bdk wallet 2 to receives 50 tx with 700 utxos from wallet 1
            let time = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
            let mut dir = std::env::temp_dir();
            dir.push(format!("bdk_{}", time.as_nanos()));
            let sqlite_db = SqliteDatabase::new(String::from(dir.to_str().unwrap()));
            let wallet2_descriptor = "wpkh(tprv8i8F4EhYDMquzqiecEX8SKYMXqfmmb1Sm7deoA1Hokxzn281XgTkwsd6gL8aJevLE4aJugfVf9MKMvrcRvPawGMenqMBA3bRRfp4s1V7Eg3/1/*)";
            let wallet2 = Wallet::new(
                wallet2_descriptor,
                None,
                bitcoin::Network::Regtest,
                sqlite_db,
            )
            .unwrap();
            wallet2
                .sync(&electrum_blockchain, Default::default())
                .unwrap();
            assert_eq!(0, wallet2.get_balance().unwrap().confirmed);
    
            // send 50 tx with 700 utxos each from wallet1 to wallet2
            for _ in 0..50 {
                let amount = 2715;
                let address_amounts = (0..700)
                    .map(|_| {
                        (
                            wallet2
                                .get_address(AddressIndex::New)
                                .unwrap()
                                .address
                                .script_pubkey(),
                            amount,
                        )
                    })
                    .collect::<Vec<_>>();
                let fee_utxo = wallet1
                    .list_unspent()
                    .unwrap()
                    .iter()
                    .filter(|utxo| utxo.txout.value > amount)
                    .map(|utxo| utxo.outpoint)
                    .last().unwrap().clone();
                let spend_utxos = wallet1
                    .list_unspent()
                    .unwrap()
                    .iter()
                    .filter(|utxo| utxo.txout.value == amount)
                    .map(|utxo| utxo.outpoint)
                    .take(700).collect::<Vec<_>>();
                let mut tx_builder = wallet1.build_tx();
                tx_builder
                    .manually_selected_only()
                    .set_recipients(address_amounts)
                    .add_utxos(&spend_utxos).unwrap()
                    .add_utxo(fee_utxo).unwrap();
                let (mut psbt, details) = tx_builder.finish().unwrap();
                assert!(wallet1.sign(&mut psbt, SignOptions::default()).unwrap());
                let tx = psbt.extract_tx();
                electrum_blockchain.broadcast(&tx).unwrap();
                // include test txs in a block
                test_client.generate(1, None);
                wallet1
                    .sync(&electrum_blockchain, Default::default())
                    .unwrap()
            }
            wallet2
                .sync(&electrum_blockchain, Default::default())
                .unwrap();
            assert_eq!(
                50 * 700,
                wallet2
                    .list_unspent()
                    .unwrap()
                    .iter()
                    .filter(|utxo| utxo.txout.value == 2715)
                    .count()
            );
            assert_eq!(wallet2.get_balance().unwrap().confirmed, 2715 * 700 * 50);
        }
    ```

[^2]: Quick and dirty work around for SQLite panic:
    ```rust
    impl BatchDatabase for SqliteDatabase {
        type Batch = SqliteDatabase;
    
        fn begin_batch(&self) -> Self::Batch {
            let db = SqliteDatabase::new(self.path.clone());
            //db.connection.execute("BEGIN TRANSACTION", []).unwrap();
            db
        }
    
        fn commit_batch(&mut self, batch: Self::Batch) -> Result<(), Error> {
            //batch.connection.execute("COMMIT TRANSACTION", [])?;
            Ok(())
        }
    }
    ```

notmandatory:
### Description

fixes #1827 

I changed the `SqliteDatabase` struct to contain a `r2d2` with [`r2d2_sqlite`](https://docs.rs/r2d2_sqlite/latest/r2d2_sqlite/) connection pool instead of a sqlite `Connection` so the `SqliteDatabase` can be safely cloned and reused when a batch is started. This prevents a large sync from trying to start and commit a batch (db transaction) while the initial non-batch connection is still busy writing it's data.

### Notes to the reviewers

I had to comment out the `test_batch_script_pubkey()` test for `sqlite` since the fix in this PR prevents a new sqlite DB connections from being created when a batch is started. In practice our sync'ing code does all its initial work without starting a batch (db transaction), then before finishing it starts a batch to do some cleanup work and commits it.

### Changelog notice

- Fix SQLite panic when syncing many UTXOs #1827

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/blockchain/electrum.rs
+++ b/src/blockchain/electrum.rs
@@ -346,10 +346,9 @@ impl ConfigurableBlockchain for ElectrumBlockchain {
 #[cfg(feature = "test-electrum")]
 mod test {
     use super::*;
-    use crate::database::{MemoryDatabase, SqliteDatabase};
+    use crate::database::MemoryDatabase;
     use crate::testutils::blockchain_tests::TestClient;
     use crate::testutils::configurable_blockchain_tests::ConfigurableBlockchainTester;
-    use crate::wallet::coin_selection::OldestFirstCoinSelection;
     use crate::wallet::{AddressIndex, Wallet};
 
     crate::bdk_blockchain_tests! {
@@ -438,6 +437,13 @@ mod test {
     #[test]
     #[ignore] // takes ~1 hr to complete, here as reference for future testing
     fn test_electrum_large_num_utxos() {
+        use crate::database::SqliteDatabase;
+        use crate::wallet::coin_selection::OldestFirstCoinSelection;
+        use crate::SignOptions;
+        use bitcoin::Amount;
+        use bitcoincore_rpc::RpcApi;
+        use std::time::{SystemTime, UNIX_EPOCH};
+
         const NUM_TX: u32 = 50;
         const NUM_UTXO: u32 = 700;
 
--- a/src/blockchain/electrum.rs
+++ b/src/blockchain/electrum.rs
@@ -278,15 +278,20 @@ impl<'a, 'b, D: Database> TxCache<'a, 'b, D> {
             }
         }
 
-        if !need_fetch.is_empty() {
+        // For some wallets there exists a pathological case where we may try to fetch many thousands
+        // of transactions at once, which creates enormous memory pressure. By chunking the batch
+        // into more reasonably sized sub-queries, we allow time for memory to be freed.
+        for chunk in need_fetch.chunks(1000) {
             let txs = self
                 .client
-                .batch_transaction_get(need_fetch.clone())
+                .batch_transaction_get(chunk)
                 .map_err(Error::Electrum)?;
+
             let mut txs: HashMap<_, _> = txs.into_iter().map(|tx| (tx.txid(), tx)).collect();
-            for txid in need_fetch {
-                if let Some(tx) = txs.remove(txid) {
-                    self.cache.insert(*txid, tx);
+
+            for txid in chunk {
+                if let Some(tx) = txs.remove(*txid) {
+                    self.cache.insert(**txid, tx);
                 }
             }
         }
--- a/src/blockchain/electrum.rs
+++ b/src/blockchain/electrum.rs
@@ -131,8 +131,8 @@ impl WalletSync for ElectrumBlockchain {
         let chunk_size = self.stop_gap + 1;
 
         // The electrum server has been inconsistent somehow in its responses during sync. For
-        // example, we do a batch request of transactions and the response contains less
-        // tranascations than in the request. This should never happen but we don't want to panic.
+        // example, we do a batch request of transactions and the response contains fewer
+        // transactions than in the request. This should never happen, but we don't want to panic.
         let electrum_goof = || Error::Generic("electrum server misbehaving".to_string());
 
         let batch_update = loop {
@@ -345,12 +345,11 @@ impl ConfigurableBlockchain for ElectrumBlockchain {
 #[cfg(test)]
 #[cfg(feature = "test-electrum")]
 mod test {
-    use std::sync::Arc;
-
     use super::*;
-    use crate::database::MemoryDatabase;
+    use crate::database::{MemoryDatabase, SqliteDatabase};
     use crate::testutils::blockchain_tests::TestClient;
     use crate::testutils::configurable_blockchain_tests::ConfigurableBlockchainTester;
+    use crate::wallet::coin_selection::OldestFirstCoinSelection;
     use crate::wallet::{AddressIndex, Wallet};
 
     crate::bdk_blockchain_tests! {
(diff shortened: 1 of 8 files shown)


notmandatory:
Test completes without panic in ~58 mins.
```
test blockchain::electrum::test::test_electrum_large_num_utxos ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 327 filtered out; finished in 3501.34s
```

src/database/sqlite.rs
```diff
@@ -76,10 +76,8 @@ static MIGRATIONS: &[&str] = &[
 /// [`crate::database`]
 #[derive(Debug)]
 pub struct SqliteDatabase {
-    /// Path on the local filesystem to store the sqlite file
-    pub path: PathBuf,
```
notmandatory, line 80: I removed this since it's no longer needed.

notmandatory:
Rebased on release/0.30 branch.

notmandatory:
The sync is still very slow, but that's due to how  electrum syncing works.  It requests complete histories for all  addresses for every sync. To support these sort of very large UTXO wallets on a mobile device I think a protocol like CBF (kyoto) is the only way to do it. CBF syncs block by block, so once blocks with new TX are processed they don't get processed again. Down side of CBF is it doesn't check the mempool, but a way to combine CBF for committed TX and electrum or esplora to get unconfirmed mempool TX should work.

src/database/sqlite.rs
```diff
@@ -76,10 +74,8 @@ static MIGRATIONS: &[&str] = &[
...
     /// A rusqlite connection object to the sqlite database
-    pub connection: Connection,
+    pub connection: Rc<Connection>,
```
ankitgoswami: Could Arc be used instead to maintain thread safety?

nymius:
I'm trying to run `test_electrum_large_num_utxos`, reducing `NUM_TX` to 10, as you mentioned in [this comment](https://github.com/bitcoindevkit/bdk/issues/1827#issuecomment-2646023260), but I'm getting dependency errors. Could you provide more insight about how to test it? I think I missing an important part of the setup.

reez:
Just adding some info when pointing to this branch from bdk-ffi and building.

I'm using a branch based on the `bdk-ffi` 0.31.2 tag.

That branch was pointing to `bdk` version 0.29.0, so I pointed to this PR branch instead `"https://github.com/notmandatory/bdk", branch = "fix/many_utxos"`

I won't list every error because from what I can tell the errors all seem to stem from the same root cause of Rc<Connection> not being Send + Sync, but definitely let me know if you'd like to me to follow up with them.

```
bdk-ffi % cargo build

error[E0277]: `Rc<Connection>` cannot be sent between threads safely
     |
1461 | uniffi::deps::static_assertions::assert_impl_all!(r#Wallet: Sync, Send);
     |                                                   ^^^^^^^^ `Rc<Connection>` cannot be sent between threads safely
     |
     = help: within `AnyDatabase`, the trait `std::marker::Send` is not implemented for `Rc<Connection>`, which is required by `wallet::Wallet: Sync`
```

notmandatory:
I'm replacing the `Rc` with a `r2d2::Pool` which is Send + Sync and builds with `bdk-ffi`.  But I'm also having trouble re-running the test so trying to figure that out now.

notmandatory:
I retested with the `r2d2` connection pool and test passes with no errors: 
```
test blockchain::electrum::test::test_electrum_large_num_utxos ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 327 filtered out; finished in 3427.42s
```

I also verified `bdk-ffi` builds with no errors. We just need to add one new `bdk-ffi::BdkError` enum variant in the UDL:
```
[Error]
enum BdkError {
  ...
  "R2d2",
};
```

src/database/sqlite.rs
notmandatory: I did a test with `Arc<RwLock<Connection>>` but this causes deadlocks. The `rusqlite::Connection` type is a little strange in that it is `Send` but that guarantee is provided by the underlying C library not by Rust. I ended up using the `r2d2_sqlite` connection pool which supports thread safety and cloning, which is what we needed.

notmandatory:
> I'm trying to run `test_electrum_large_num_utxos`, reducing `NUM_TX` to 10, as you mentioned in [this comment](https://github.com/bitcoindevkit/bdk/issues/1827#issuecomment-2646023260), but I'm getting dependency errors. Could you provide more insight about how to test it? I think I missing an important part of the setup.

You should be able to test it now. I think I forgot to check in some dependency fixes that were preventing the electrum tests from building.

nymius:
> You should be able to test it now. I think I forgot to check in some dependency fixes that were preventing the electrum tests from building.

Thanks! It works now

src/blockchain/electrum.rs
```diff
@@ -429,4 +427,179 @@ mod test {
 
         ElectrumTester.run();
     }
+
+    #[cfg(feature = "sqlite")]
+    #[test]
+    #[ignore] // takes ~1 hr to complete, here as reference for future testing
```
nymius, line 438: Should we remove this comment or update it? With `NUM_TX = 2` (as it is now) it took ~106 seconds on average to run for my computer (I tested it multiple times).

src/blockchain/electrum.rs
```diff
@@ -429,4 +427,179 @@ mod test {
...
+                None,
+            )
+            .unwrap();
+        test_client.generate(1, None);
+        wallet1
+            .sync(&electrum_blockchain, Default::default())
+            .unwrap();
+        assert_eq!(wallet1.get_balance().unwrap().confirmed, 5_0000_0000);
+        // bdk wallet 1 creates NUM_TX tx * NUM_UTXO utxos and sends them back to itself
+        for _ in 0..NUM_TX {
```
nymius, line 479: If not so time consuming I would implement these as inner functions `fund_wallet_with_many_utxos` and `send_many_utxos`. Will improve discoverability of the code for future occasions if we have to reproduce a similar scenario.

nymius (commented):
tACK

Didn't manage to produce the failure at commit a75f749 using `test_electrum_large_num_utxos` with `NUM_TX = 5`

reez:
> I also verified `bdk-ffi` builds with no errors. We just need to add one new `bdk-ffi::BdkError` enum variant in the UDL:

I verified `bdk-ffi` builds w no errors as well, and also built the `bdk-swift` bindings locally & used those local bindings in BDK Swift Example Wallet.

src/blockchain/electrum.rs
notmandatory: oh ya I forgot to change it back from 2, it might take 10 to show the error or more on a faster computer. I'm going to see if I can repro the error with the test but not the fix from this PR but with the fix in #1831 instead.

notmandatory:
I rebased this PR on #1831 which fixes another issue related to syncing large transactions. After #1831 is merged this one can go in.

src/blockchain/electrum.rs
notmandatory: I'll work on a cleaner version for the 1.0 master branch since this 0.30.x branch of code will be ending soon.

src/blockchain/electrum.rs
notmandatory: I also had a test run that didn't fail with 5 but did with 10 tx. I put the test back to 50 anyway.

src/database/sqlite.rs
```diff
@@ -919,23 +925,20 @@ impl BatchDatabase for SqliteDatabase {
...
+        connection.execute("BEGIN TRANSACTION", []).unwrap();
+        SqliteDatabase {
+            pool: self.pool.clone(),
+        }
     }
 
     fn commit_batch(&mut self, batch: Self::Batch) -> Result<(), Error> {
-        batch.connection.execute("COMMIT TRANSACTION", [])?;
+        let connection = batch.pool.get()?;
+        connection.execute("COMMIT TRANSACTION", [])?;
```
evanlinjin, line 937: From my understanding, an [sqlite transaction is isolated to a single connection](https://sqlite.org/isolation.html), unless we are in "shared cache mode".

~Even in "shared cache mode", is it even possible to start a transaction in one connection and commit it in another?~ Edit: It is definitely not possible. [Almost all statements in sqlite are automatically a transaction](https://sqlite.org/lang_transaction.html) so it would make no sense that transactions are begun and committed from different connections.

thunderbiscuit:
We'll be ready to build a patch release of the bdk-ffi libraries once both those fixes are in.

src/database/sqlite.rs
notmandatory: Yes by default all the individual statements in the syncing happen in their on transaction. But by starting and ending a db transaction it ensures that a group of statements are committed or rolled back atomically. For this final phase of syncing `into_db_update` and also in `cache_addresses` there are multi statements that need to commit or roll back together. 

The main thing I'm fixing by introducing this `r2d2` pool is that we don't have different threads making their own `Connection` to the same sqlite DB file. Even when the `SqliteDatabase.pool` is cloned it's still using the same `Connection`. There's probably a way to do this without using a connection pool like `r2d2` but the lifetimes and ownership gets hairy and I didn't want to roll my own when a common solution is to use a pool.

notmandatory: ~~For reference, the `r2d2_sqlite` connection manager uses a `Mutex<Option<Connection>>` so that a single connection can be shared. I could remove `r2d2` and do the same but then I'd have to replicate the connection logic they already have.~~ Sorry this is wrong, the Mutex is only used for sqlite in memory connections.

https://github.com/ivanceras/r2d2-sqlite/blob/fbba255ff23aeaa005e810ce16fae19f1e44b0c2/src/lib.rs#L60-L65

evanlinjin: @notmandatory Okay that makes sense, thanks for clarifying! Maybe a comment somewhere stating that `ConnectionPool` only manages a single connection will be helpful.

evanlinjin (approved):
ACK 29f6935b440762af6768eaddd127435b4b088587

Only request: Comment somewhere saying that the Sqlite Connection Pool will always only have a single connection.

src/database/sqlite.rs
evanlinjin: @notmandatory Okay I see you edited your previous comment...

evanlinjin: I guess we can create an `r2d2` Pool with the `max_size` set to 1.

evanlinjin (changes requested):
Need to ensure we handle db transactions properly (as discussed above).

src/database/sqlite.rs
```diff
@@ -919,23 +925,20 @@ impl BatchDatabase for SqliteDatabase {
     type Batch = SqliteDatabase;
 
     fn begin_batch(&self) -> Self::Batch {
-        let db = SqliteDatabase::new(self.path.clone());
```
evanlinjin, line 922: I think a more straight-forward fix would be to just somehow clone `self` and return it. Trying it out now.

src/database/sqlite.rs
notmandatory: I'm trying to find out exactly what `r2d2` does but it defaults to 10 connections and everything still works. I think the reason is `sqlite` allows multiple connections on the same file but only a single "writer" as described by https://github.com/rusqlite/rusqlite/issues/920. I think `r2d2` is doing something else to avoid the "DatabaseBusy" error. The `r2d2_sqlite` crate has a `is_valid()` function that starts an empty "batch" which I believe fails if the connection is not writable. https://github.com/ivanceras/r2d2-sqlite/blob/fbba255ff23aeaa005e810ce16fae19f1e44b0c2/src/lib.rs#L163C5-L165C6

notmandatory: Now I believe we actually don't want a pool of `max_size` 1 since it will block safe read-only queries we make during write transactions.

src/database/sqlite.rs
notmandatory: Unfortunately `rusqlite::Connection` doesn't support `Clone`. I tried `Arc<Mutex>` and separating read and write connections, including setting the sqlite connection flag on the read connection to be READ_ONLY, but continue to get the original err.

I'd like to stick with this pool solution since I don't think there's any reason to doubt `r2d2` handles multiple concurrent transactions, even if I can't pinpoint exactly how. Moreover by using this connection pool 1) our tests pass, 2) the `bdk_ffi` project builds and bitkey's app doesn't crash on their example big wallet 3) the `r2d2_sqlite` example demos multiple concurrent threads writing to the DB in their own automatic transactions. Bottom line is I don't want to re-invent a multi-threaded sqlite connection pool when we already have one that works.

As a side note, I did test with the `r2d2` pool configured to use only one thread and my test still passes, but I'd rather leave it at the default max 10 connections so we don't unnecessarily limit performance.

notmandatory:
Replaced by #1836

PR #1829, opened 2025-02-10
ci: remove `cc` pin to fix CI error

LagginTimes:
### Description
`cc` dependency was unpinned to fix a clash with `cmake` that was causing CI errors.

### Changelog notice

* Unpinned cc dependency.

#### All Submissions:


ValuedMammal:
I wonder if we could instead just not pin `cc`?

LagginTimes:
`cc = ^1.0.106` would require a MSRV above 1.63 ([reference](https://github.com/rust-lang/cc-rs/releases/tag/cc-v1.0.106)).

ValuedMammal:
I'm confused because it looks like `cc` has since reversed the decision to raise their MSRV, plus this CI failure https://github.com/bitcoindevkit/bdk/actions/runs/13237150987/job/36944325715 points to bdk_electrum as the cause which is supposed to be exempt from our MSRV of 1.63

LagginTimes:
Ah yes good catch, looks like [`cc 1.1.3`](https://github.com/rust-lang/cc-rs/releases/tag/cc-v1.1.3) reduced the MSRV back to 1.63. I'll make the changes to this PR accordingly.

Edit:
> points to bdk_electrum as the cause which is supposed to be exempt from our MSRV of 1.63

I believe this may be because the `cargo update`s from `pin-msrv.sh` run before the `build` and `test` commands that explicitly exclude `bdk_electrum`.

ValuedMammal (approved):
ACK 979f9d1e6297fbd43f3f8e1061405ae3a901efc9

PR #1830, opened 2025-02-10
Fix off-by-one error checking coinbase maturity in optional UTxOs

### LINKED ISSUE #1810: Off-by-one error is making optional UTxOs selection too restrictive for coinbase outputs
**Describe the bug**  
The `preselect_utxos` has an **off-by-one** error that is making the selection of optional UTxOs too restrictive, by requiring the coinbase outputs to surpass or equal coinbase maturity time **at the current height of the selection**, and not in the block in which the transaction may be included in the blockchain (be spent).

**To Reproduce**
Apply the following changes to`test_spend_coinbase`:
```diff
diff --git a/crates/wallet/tests/wallet.rs b/crates/wallet/tests/wallet.rs
index 174a628d..ac4fa312 100644
--- a/crates/wallet/tests/wallet.rs
+++ b/crates/wallet/tests/wallet.rs
@@ -3875,8 +3875,8 @@ fn test_spend_coinbase() {
     };
     insert_anchor(&mut wallet, txid, anchor);
 
-    let not_yet_mature_time = confirmation_height + COINBASE_MATURITY - 1;
-    let maturity_time = confirmation_height + COINBASE_MATURITY;
+    let not_yet_mature_time = confirmation_height + COINBASE_MATURITY - 2;
+    let maturity_time = confirmation_height + COINBASE_MATURITY - 1;
 
     let balance = wallet.balance();
     assert_eq!(
```

**Expected behavior**  
`cargo test -- test_spend_coinbase` should pass with the applied changes, but it fails with `CoinSelection::InsufficientFunds` because the only spendable UTxO is marked as immature.

**Build environment**  
 - BDK tag/commit: 43f0f8de
 - OS+version: `Ubuntu 24.04.1 LTS`
 - Rust/Cargo version:  `cargo 1.83.0 (5ffbef321 2024-10-29)`
 - Rust/Cargo target: `x86_64-unknown-linux-gnu`

**Additional context**  
Why I think the correct behavior is **to signal the output as mature in $h+99$[^1]instead of $h+100$[^1]**?

Because as this [stack overflow answer](https://bitcoin.stackexchange.com/a/119017) states:
> A transaction that spends an output of the coinbase transaction at height h, is eligible to be included in block h+100[^1] or higher.
> ...
> This means that miners may include such a transaction in their block template when their chain tip has a height of h+99[^1].

The bug is in [these lines in wallet crate](https://github.com/bitcoindevkit/bdk/blob/1afa439d1756d1294cce9d5e8917113062de442b/crates/wallet/src/wallet/mod.rs#L2053-L2057) where the difference between the current height and the UTxO mining height is forced to already be of 100 blocks, effectively marking it for inclusion in chain starting from block $h+101$[^1], one more than it should be.

The lines mentioned are also referencing [bitcoin core code](https://github.com/bitcoin/bitcoin/blob/c5e67be03bb06a5d7885c55db1f016fbf2333fe3/src/validation.cpp#L372-L375). The code is checking transaction validity after reorg. The linked lines are particularly discarding transactions if some input is spending a coinbase UTxO and the difference between the spend height (**current height + 1**) and the **mine height of the UTxO** are less than coinbase maturity. So, to consider its inclusion in a set of candidates for coin selection we should check the same condition. And this [chain crate code](https://github.com/bitcoindevkit/bdk/blob/1afa439d1756d1294cce9d5e8917113062de442b/crates/chain/src/chain_data.rs#L106) in BDK is exactly doing that.

There is a test asserting coinbase maturity but it is faulty because of the same mistake: checking transaction can't be spend at $h+100$ and can be spend at $h+101$.
The correct test should be: **can't be spend at $h+98$** (because in the block $h+99$ which would hypothetically include the transaction, the input wouldn't have 100 blocks since mined). And **can be spend at h+99** because then transaction will be a candidate for inclusion starting from block $h+100$.

[^1]: where h is the height at which the UTxO transaction was mined

nymius:
### Description

As I was developing the changes in #1798 I discover issue #1810. So I introduced the fixes in that PR but later I split them in two to ease the review by suggestion of @oleonardolima .

The `preselect_utxos` method has an off-by-one error that is making the selection of optional UTxOs too restrictive, by
requiring the coinbase outputs to surpass or equal coinbase maturity time at the current height of the selection, and
not in the block in which the transaction may be included in the blockchain.

The changes in this commit fix it by considering the maturity of the coinbase output at the spending height and not
the transaction creation height, this means, a +1 at the considered height at the moment of building the
transaction.

Fixes #1810.

### Notes to the reviewers

Tests for issue #1810 have not been explicitly added, as there already was a `text_spend_coinbase` test which was corrected to ensure coinbase maturation is considered in alignment with the new logic.

Changes are not breaking but I'm modifying slightly the documentation for the public method `TxBuilder::current_height` to adjust to the fixed code. Does this merit an entry in the CHANGELOG?

### Changelog notice

`Wallet` now considers a utxo originated from a coinbase transaction (`coinbase utxo`) as available for selection if it will mature in the next block after the height provided to the selection, the current height by default. The previous behavior allowed selecting a `coinbase utxo` only when the height at the moment of selection was equal to maturity height or greater.

### Checklists

* [x] I've updated existing tests to match the fix
* [x] I've updated docs to match the fix logic
* [x] This pull request DOES NOT break the existing API

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -2052,9 +2052,10 @@ impl Wallet {
                         match chain_position {
                             ChainPosition::Confirmed { anchor, .. } => {
                                 // https://github.com/bitcoin/bitcoin/blob/c5e67be03bb06a5d7885c55db1f016fbf2333fe3/src/validation.cpp#L373-L375
-                                spendable &= (current_height
-                                    .saturating_sub(anchor.block_id.height))
-                                    >= COINBASE_MATURITY;
+                                let spend_height = current_height + 1;
+                                let coin_age_at_spend_height =
+                                    spend_height.saturating_sub(anchor.block_id.height);
+                                spendable &= coin_age_at_spend_height >= COINBASE_MATURITY;
                             }
                             ChainPosition::Unconfirmed { .. } => spendable = false,
                         }
--- a/crates/wallet/src/wallet/tx_builder.rs
+++ b/crates/wallet/src/wallet/tx_builder.rs
@@ -556,9 +556,9 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
     /// 1. Set the nLockTime for preventing fee sniping.
     ///    **Note**: This will be ignored if you manually specify a nlocktime using [`TxBuilder::nlocktime`].
     /// 2. Decide whether coinbase outputs are mature or not. If the coinbase outputs are not
-    ///    mature at `current_height`, we ignore them in the coin selection.
-    ///    If you want to create a transaction that spends immature coinbase inputs, manually
-    ///    add them using [`TxBuilder::add_utxos`].
+    ///    mature at spending height, which is `current_height` + 1, we ignore them in the coin
+    ///    selection. If you want to create a transaction that spends immature coinbase inputs,
+    ///    manually add them using [`TxBuilder::add_utxos`].
     ///
     /// In both cases, if you don't provide a current height, we use the last sync height.
     pub fn current_height(&mut self, height: u32) -> &mut Self {
--- a/crates/wallet/tests/wallet.rs
+++ b/crates/wallet/tests/wallet.rs
@@ -3875,8 +3875,16 @@ fn test_spend_coinbase() {
     };
     insert_anchor(&mut wallet, txid, anchor);
 
-    let not_yet_mature_time = confirmation_height + COINBASE_MATURITY - 1;
-    let maturity_time = confirmation_height + COINBASE_MATURITY;
+    // NOTE: A transaction spending an output coming from the coinbase tx at height h, is eligible
+    // to be included in block h + [100 = COINBASE_MATURITY] or higher.
+    // Tx elibible to be included in the next block will be accepted in the mempool, used in block
+    // templates and relayed on the network.
+    // Miners may include such tx in a block when their chaintip is at h + [99 = COINBASE_MATURITY - 1].
+    // This means these coins are available for selection at height h + 99.
+    //
+    // By https://bitcoin.stackexchange.com/a/119017
+    let not_yet_mature_time = confirmation_height + COINBASE_MATURITY - 2;
+    let maturity_time = confirmation_height + COINBASE_MATURITY - 1;
 
     let balance = wallet.balance();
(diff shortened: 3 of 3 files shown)


oleonardolima (commented):
cACK efe9b4a93beb1a7a1683c2e12677b47c919ef25d

LagginTimes (approved):
ACK 03b7ecaea1176f55c6792d8798ead72b23082af7

ValuedMammal:
Looks good to me. I would add to Changelog something like this

`Wallet` now considers a coinbase output available for selection if it will mature in the next block after the current height; the previous behavior allowed selecting a coinbase output only when the current height reached the maturity height or greater.

nymius:
> Looks good to me. I would add to Changelog something like this
> 
> `Wallet` now considers a coinbase output available for selection if it will mature in the next block after the current height; the previous behavior allowed selecting a coinbase output only when the current height reached the maturity height or greater.

I got confused and added an entry to `crates/wallet/CHANGELOG.md`. Fortunately, I checked the CHANGELOG policy again, remove it and added the comment to the CHANGELOG section of the PR header.

crates/wallet/src/wallet/mod.rs
```diff
@@ -2052,9 +2052,10 @@ impl Wallet {
                         match chain_position {
                             ChainPosition::Confirmed { anchor, .. } => {
                                 // https://github.com/bitcoin/bitcoin/blob/c5e67be03bb06a5d7885c55db1f016fbf2333fe3/src/validation.cpp#L373-L375
-                                spendable &= (current_height
-                                    .saturating_sub(anchor.block_id.height))
-                                    >= COINBASE_MATURITY;
+                                let spend_height = current_height + 1;
```
evanlinjin, line 2055: Nit: I think a better name would be `inclusion_height`.

evanlinjin (approved):
ACK 03b7ecaea1176f55c6792d8798ead72b23082af7

crates/wallet/src/wallet/mod.rs
nymius: Won't be needed if #1798 is merged.

PR #1833, opened 2025-02-14
wallet: unnecessary coinbase chain position assert

torkelrogstad:
### Description

Removes an unneeded assertion from `preselect_utxos`. I'm not precisely sure how I triggered this, but I encountered this while syncing a rather large signet wallet containing lots of freshly generated UTXOs. 

This assertion is not needed, as unconfirmed coinbase TXs get filtered away in the branch below anyways.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -2073,10 +2073,6 @@ impl Wallet {
                     return false;
                 }
                 if tx.is_coinbase() {
-                    debug_assert!(
-                        chain_position.is_confirmed(),
-                        "coinbase must always be confirmed"
-                    );
                     if let Some(current_height) = current_height {
                         match chain_position {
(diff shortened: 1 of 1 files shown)


evanlinjin:
> I'm not precisely sure how I triggered this, but I encountered this while syncing a rather large signet wallet containing lots of freshly generated UTXOs.

The fact that there is an unconfirmed coinbase tx signals that there is a bug.

Would you mind describing the conditions that triggered this?

I.e.
* What chain source you used. `bdk_electrum`, `bdk_esplora`, something custom, etc.
* whether you did a full-scan request or a sync request (if electrum/esplora) before trying to create tx.
* Wallet state before the sync.

torkelrogstad:
This might actually be fixed by this recent PR: https://github.com/bitcoindevkit/bdk/pull/1830

PR #1837, opened 2025-02-16
`bdk_electrum`: Handle negative heights properly

evanlinjin:
### Description

This fixes a problem with `bdk_electrum` making bogus `transaction_get_merkle` calls to the Electrum server.

In electrum, heights returned alongside txs may be 0 or -1. 0 means the tx is unconfirmed. -1 means the tx is unconfirmed and spends an unconfirmed tx.

Previously, the codebase assumed that heights cannot be negative and used a `i32 as usize` cast (which would lead to the casted height being 18446744073709551615). Then the codebase will try to call `transaction_get_merkle` with the overflowed height.

### Notes to the reviewers

The test introduced in this PR does not fail with the old codebase. What ends up happening is that `transaction_get_merke` will be called with the overflow height and the Electrum server will return with "merkle not found".

To see the failure, you can change the `height as usize` casts to use `.try_into().expect()` then you will see a panic.

These changes should be applied as `1.0.1` and `1.1.1` fixes.

### Changelog notice

* Fix `bdk_electrum` handling of negative spk-history height.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -276,7 +276,9 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
 
                 for tx_res in spk_history {
                     tx_update.txs.push(self.fetch_tx(tx_res.tx_hash)?);
-                    self.validate_merkle_for_anchor(tx_update, tx_res.tx_hash, tx_res.height)?;
+                    if let Ok(height) = tx_res.height.try_into() {
+                        self.validate_merkle_for_anchor(tx_update, tx_res.tx_hash, height)?;
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -276,7 +276,13 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
 
                 for tx_res in spk_history {
                     tx_update.txs.push(self.fetch_tx(tx_res.tx_hash)?);
-                    self.validate_merkle_for_anchor(tx_update, tx_res.tx_hash, tx_res.height)?;
+                    if tx_res.height > 0 {
+                        self.validate_merkle_for_anchor(
+                            tx_update,
+                            tx_res.tx_hash,
+                            tx_res.height.try_into().expect("must not overflow"),
+                        )?;
```
LLFourn: :pick: We can just match on the try into result and branch based on that to avoid the unwrap/expect.

LLFourn (approved):
Nice work. Especially with the test.

LagginTimes (approved):
ACK 2afdae587fa19910cd2f55ade8571cd7ba3075c9

LagginTimes (approved):
ACK 161d715bedd6af5996c4cf24b6bc1f733c1228ea

PR #1838, opened 2025-02-16
Make full-scan/sync flow easier to reason about.

evanlinjin:
### Description

#1811 introduces the `evicted-at` concept. While adding this to `TxUpdate`, I realized there were some shortcomings in our full-scan & sync flow:

* Chain sources that use `TxUpdate` to convey updates cannot choose to add transactions without a `seen_at` timestamp as the `TxGraph::apply_update_at` logic adds this timestamp for all unanchored txs if the `seen_at` parameter is specified. Purposefully adding uncanonical txs is useful for wallets that want to know about replaced txs.
* The `TxGraph::apply_update_at` logic is hard to reason about. `TxUpdate::seen_ats` already has the `seen_at` timestamp per tx, but `apply_update_at()` also takes in a `seen_at` timestamp`.
* `TxUpdate::seen_ats` currently forces us to only have one `seen-at` per tx as it's a map of `txid -> seen_at`. However, in the future we want to add a `first-seen` timestamp to `TxGraph` which is basically the earliest `seen-at` timestamp introduced so we may want to have multiple `seen_at`s per tx. The other issue is if we merge `TxUpdate`s, we will loose data. I.e. we can only keep the `last_seen` or the `first_seen`.

These problems were exacerbated when introducing `evicted-at`. In the old design, the chain-source has no concept of sync time (as sync time was introduced after-the-fact when applying the `TxUpdate`). Therefore the introduced `TxUpdate::evicted` field was a `HashSet<Txid>` (no timestamps) and relied on the `TxGraph::apply_upate_at` logic to introduce the `evicted-at` timestamp. All this makes the sync logic harder to understand. What happens if the `seen_at` input is `None`? What happens if updates were applied out of order? What happens when we merge `TxUpdates` before applying?

The following changes are made in this PR to simplify the sync/full-scan logic to be easier to understand and robust:

* The `sync_time` is added to the `SyncRequest` and `FullScanRequest`. `sync_time` is mandatory and is added as an input of `builder()`. If the `std` feature is enabled, the `builder_now()` is available which uses the current timestamp. The chain source is now responsible to add this `sync_time` timestamp as `seen_at` for mempool txs. Non-canonical and non-anchored txs can be added without the `seen_at` timestamp.
* `TxUpdate::seen_ats` is now a `HashSet` of `(Txid, u64)`. This allows multiple `seen_at`s per tx. This is also a much easier to use API as the chain source can just insert into this `HashSet` without checking previous data.
* `TxGraph::apply_update_at` is removed as we no longer needs to introduce a fallback `seen_at` timestamp after-the-fact. `TxGraph::apply_update` is no longer `std`-only and the logic of applying updates is greatly simplified.

Additionally, `TxUpdate` is updated to be `#[non_exhaustive]` for backwards-compatibility protection.

### Notes to the reviewers

This is based on #1837. Merge that first.

These are breaking changes to `bdk_core`. It needs to be breaking to properly fix all the issues.

As mentioned by @notmandatory, `bdk_wallet` changes will be added to a new PR once the new `bdk_wallet` repo is ready. But the PR won't be merged until we're ready for a `bdk_wallet` 2.0.

### Changelog notice

* Change `FullScanRequest::builder` and `SyncRequest::builder` methods to depend on `feature = "std"`. This is because requests now have a `start_time`, instead of specifying a `seen_at` when applying the update.
* Add `FullScanRequest::builder_at` and `SyncRequest::builder_at` methods which are the non-std version of the `..Request::builder` methods.
* Change `TxUpdate::seen_ats` field to be a `HashSet` of `(Txid, u64)`. This allows a single update to have multiple `seen_at`s per tx.
* Change `TxUpdate` to be `non-exhaustive`.
* Remove `apply_update_at` as we no longer need to apply with a timestamp after-the-fact.

#### All Submissions:

#### New Features:

~* [ ] I've added tests for the new feature~ No tests needed as it's more of a refactor.

### DIFF
--- a/crates/core/src/spk_client.rs
+++ b/crates/core/src/spk_client.rs
@@ -87,17 +87,11 @@ impl SyncProgress {
 }
 
 /// Builds a [`SyncRequest`].
+///
+/// Construct with [`SyncRequest::builder`].
 #[must_use]
 pub struct SyncRequestBuilder<I = ()> {
     inner: SyncRequest<I>,
-}
-
-impl<I> Default for SyncRequestBuilder<I> {
-    fn default() -> Self {
-        Self {
-            inner: Default::default(),
-        }
-    }
 }
 
 impl SyncRequestBuilder<()> {
@@ -210,6 +204,7 @@ impl<I> SyncRequestBuilder<I> {
 /// ```
 #[must_use]
 pub struct SyncRequest<I = ()> {
+    start_time: u64,
     chain_tip: Option<CheckPoint>,
     spks: VecDeque<(I, ScriptBuf)>,
     spks_consumed: usize,
@@ -220,21 +215,6 @@ pub struct SyncRequest<I = ()> {
     inspect: Box<InspectSync<I>>,
 }
 
-impl<I> Default for SyncRequest<I> {
-    fn default() -> Self {
-        Self {
-            chain_tip: None,
-            spks: VecDeque::new(),
-            spks_consumed: 0,
-            txids: VecDeque::new(),
-            txids_consumed: 0,
-            outpoints: VecDeque::new(),
-            outpoints_consumed: 0,
-            inspect: Box::new(|_, _| {}),
-        }
-    }
-}
-
 impl<I> From<SyncRequestBuilder<I>> for SyncRequest<I> {
     fn from(builder: SyncRequestBuilder<I>) -> Self {
         builder.inner
@@ -242,11 +222,47 @@ impl<I> From<SyncRequestBuilder<I>> for SyncRequest<I> {
 }
 
 impl<I> SyncRequest<I> {
-    /// Start building a [`SyncRequest`].
-    pub fn builder() -> SyncRequestBuilder<I> {
+    /// Start building [`SyncRequest`] with a given `start_time`.
+    ///
+    /// `start_time` specifies the start time of sync. Chain sources can use this value to set
+    /// [`TxUpdate::seen_ats`](crate::TxUpdate::seen_ats) for mempool transactions. A transaction
+    /// without any `seen_ats` is assumed to be unseen in the mempool.
+    ///
+    /// Use [`SyncRequest::builder`] to use the current timestamp as `start_time` (this requires
+    /// `feature = "std"`).
+    pub fn builder_at(start_time: u64) -> SyncRequestBuilder<I> {
         SyncRequestBuilder {
-            inner: Default::default(),
+            inner: Self {
+                start_time,
+                chain_tip: None,
+                spks: VecDeque::new(),
+                spks_consumed: 0,
+                txids: VecDeque::new(),
+                txids_consumed: 0,
+                outpoints: VecDeque::new(),
+                outpoints_consumed: 0,
+                inspect: Box::new(|_, _| ()),
+            },
         }
+    }
+
+    /// Start building [`SyncRequest`] with the current timestamp as the `start_time`.
+    ///
+    /// Use [`SyncRequest::builder_at`] to manually set the `start_time`, or if `feature = "std"`
+    /// is not available.
+    #[cfg(feature = "std")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "std")))]
+    pub fn builder() -> SyncRequestBuilder<I> {
+        let start_time = std::time::UNIX_EPOCH
+            .elapsed()
+            .expect("failed to get current timestamp")
+            .as_secs();
+        Self::builder_at(start_time)
+    }
+
+    /// When the sync-request was initiated.
+    pub fn start_time(&self) -> u64 {
+        self.start_time
     }
 
     /// Get the [`SyncProgress`] of this request.
@@ -339,17 +355,11 @@ impl<A> Default for SyncResponse<A> {
 }
 
 /// Builds a [`FullScanRequest`].
+///
+/// Construct with [`FullScanRequest::builder`].
 #[must_use]
 pub struct FullScanRequestBuilder<K> {
     inner: FullScanRequest<K>,
-}
-
-impl<K> Default for FullScanRequestBuilder<K> {
-    fn default() -> Self {
-        Self {
-            inner: Default::default(),
-        }
-    }
 }
 
 impl<K: Ord> FullScanRequestBuilder<K> {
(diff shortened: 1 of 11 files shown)


crates/core/src/spk_client.rs
```diff
@@ -141,15 +136,15 @@ impl<I> SyncRequestBuilder<I> {
     /// let (newly_revealed_spks, _changeset) = indexer
     ///     .reveal_to_target("descriptor_a", 21)
     ///     .expect("keychain must exist");
-    /// let _request = SyncRequest::builder()
+    /// let _request = SyncRequest::builder(current_epoch)
```
LLFourn: use `builder_now` instead since that's what most people will want to use.

crates/core/src/spk_client.rs
```diff
@@ -220,35 +217,45 @@ pub struct SyncRequest<I = ()> {
...
     fn from(builder: SyncRequestBuilder<I>) -> Self {
         builder.inner
     }
 }
 
 impl<I> SyncRequest<I> {
-    /// Start building a [`SyncRequest`].
-    pub fn builder() -> SyncRequestBuilder<I> {
+    /// Start building [`SyncRequest`] with a given `start_time`.
+    pub fn builder(start_time: u64) -> SyncRequestBuilder<I> {
```
LLFourn: We need to explain what's the purpose of `start_time` here.

LLFourn (approved):
I left a couple of documentation comments but this LGTM.

crates/core/src/spk_client.rs
evanlinjin: Added.

crates/core/src/spk_client.rs
```diff
@@ -220,35 +215,53 @@ pub struct SyncRequest<I = ()> {
...
 }
 
 impl<I> SyncRequest<I> {
-    /// Start building a [`SyncRequest`].
-    pub fn builder() -> SyncRequestBuilder<I> {
+    /// Start building [`SyncRequest`] with a given `start_time`.
+    ///
+    /// `start_time` specifies the start time of sync. This can be used by chain sources as the
+    /// [`TxUpdate::seen_ats`](crate::TxUpdate::seen_ats) timestamp for mempool transactions.
+    /// Transactions it, transactions are assumed to be unseen in the mempool.
```
ValuedMammal: > Transactions it, transactions are assumed to be unseen in the mempool.

This sentence seems awkward.

crates/electrum/tests/test_electrum.rs
```diff
@@ -54,6 +60,63 @@ where
...
+
+    let spk_history = electrum_client.script_get_history(&tracked_addr.script_pubkey())?;
+    assert!(
+        spk_history.into_iter().any(|tx_res| tx_res.height < 0),
+        "must find tx with negative height"
+    );
+
+    let client = BdkElectrumClient::new(electrum_client);
+    let request = SyncRequest::builder_now().spks(core::iter::once(tracked_addr.script_pubkey()));
+    let _response = client.sync(request, 1, false)?;
```
ValuedMammal: 📌 We can check that the TxUpdate now contains the seen_at values.

ValuedMammal (commented):
Approach ACK

LLFourn (approved):
ACK 86824872020cd4f0e9e2c253fe065c8ad6ec20fa

nymius:
ACK 8682487

nymius (approved):
ACK 8682487

LagginTimes (commented):
Nit: typo in commit message.

> This may, if we introduce new fields to `TxUdpate`, they can be minor
> non-breaking updates.

Otherwise LGTM.

crates/wallet/src/wallet/mod.rs
```diff
@@ -2463,9 +2455,26 @@ impl Wallet {
...
+    #[cfg_attr(docsrs, doc(cfg(feature = "std")))]
     pub fn start_full_scan(&self) -> FullScanRequestBuilder<KeychainKind> {
+        let start_time = std::time::UNIX_EPOCH
+            .elapsed()
+            .expect("time must no go backwards")
+            .as_secs();
+        self.start_full_scan_at(start_time)
+    }
+
+    /// Create a [`FullScanRequest] for this wallet.
```
LagginTimes: Nit: Looks like this was pre-existing. Same typo exists in `start_full_scan()` docs as well.

```suggestion
    /// Create a [`FullScanRequest`] for this wallet.
```

crates/wallet/src/wallet/mod.rs
```diff
@@ -2265,31 +2265,7 @@ impl Wallet {
...
-    ///
-    /// `seen_at` represents when the update is seen (in unix seconds). It is used to determine the
-    /// `last_seen`s for all transactions in the update which have no corresponding anchor(s). The
-    /// `last_seen` value is used internally to determine precedence of conflicting unconfirmed
-    /// transactions (where the transaction with the lower `last_seen` value is omitted from the
-    /// canonical history).
-    ///
-    /// Use [`apply_update`](Wallet::apply_update) to have the `seen_at` value automatically set to
-    /// the current time.
-    pub fn apply_update_at(
```
notmandatory: Hey removing this would be breaking the `bdk_wallet 1.x` API so we need to discuss again how to release this change.

evanlinjin: I guess we can deprecate it and state that the timestamp here does nothing. 

However changing a method to require a feature flag is also breaking.

evanlinjin:
Rebased on #1860 and removed the changes to `bdk_wallet`.

crates/wallet/src/wallet/mod.rs
evanlinjin: After the [lib call](https://discord.com/channels/753336465005608961/753336465005608965/1344807550217883688), the consensus was to put `bdk_wallet` on a different release schedule to the other crates. This PR is modified to not affect `bdk_wallet`.

notmandatory (approved):
utACK 800f3580f8518e15ccaf9622fdd2e5141a50d5e5

Note to other reviewers: bdk_wallet changes will be added to a new PR once the new `bdk_wallet` repo is ready. But the PR won't be merged until we're ready for a `bdk_wallet` 2.0.

PR #1841, opened 2025-02-16
Generalize `add_recipient` to accept `Address`

rustaceanrob:
### Description

I would imagine many users would be handling a `Address<NetworkChecked>` when building a transaction. They may pass this structure directly with this patch, or continue to use `ScriptBuf`.

### Notes to the reviewers

To my knowledge this is non-breaking, but it is a change in the function signature so I am not sure.

### Changelog notice

Accept any type that is convertible to a `ScriptBuf` in `TxBuilder::add_recipient`

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/tx_builder.rs
+++ b/crates/wallet/src/wallet/tx_builder.rs
@@ -602,8 +602,12 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
     }
 
     /// Add a recipient to the internal list
-    pub fn add_recipient(&mut self, script_pubkey: ScriptBuf, amount: Amount) -> &mut Self {
-        self.params.recipients.push((script_pubkey, amount));
+    pub fn add_recipient(
+        &mut self,
+        script_pubkey: impl Into<ScriptBuf>,
+        amount: Amount,
+    ) -> &mut Self {
+        self.params.recipients.push((script_pubkey.into(), amount));
         self
     }
 


ValuedMammal:
Concept ACK, I guess the same could be done for `set_recipients`?

rustaceanrob:
That change would require a `Box` for the `dyn Into` trait. I am not necessarily opposed to that, but that does feel a little bit strange. I don't think it's the end of the world if this doesn't get in.

For 2.0.0, I would consider adding a `Recipient` type with the script and amount, and having useful constructors on it that use the `impl Into` syntax

ValuedMammal (approved):
ACK 3bb94e43ecf817be4d27e26c9d430fda156e6b25

evanlinjin:
> That change would require a `Box` for the `dyn Into` trait. I am not necessarily opposed to that, but that does feel a little bit strange. I don't think it's the end of the world if this doesn't get in.

Why would that change require a `Box`?

```rust
    pub fn set_recipients(
        &mut self,
        recipients: impl IntoIterator<Item = (impl Into<ScriptBuf>, Amount)>,
    ) -> &mut Self {
        self.params.recipients = recipients
            .into_iter()
            .map(|(spk, amount)| (spk.into(), amount))
            .collect();
        self
    }
```

The only problem with this is that it will be a breaking change because the compiler needs to know the explicit type of `recipients`. I.e. `tx_builder.set_recipients(Vec::new())` will no longer work because of missing annotations.

evanlinjin (approved):
ACK 3bb94e43ecf817be4d27e26c9d430fda156e6b25

rustaceanrob:
> Why would that change require a Box?

I was thinking dynamic dispatch would be required, but it is cool to see that is not the case. Good catch on the potentially missing annotations in `set_recipients` if the change were to be applied. Cheers

oleonardolima (commented):
utACK 3bb94e43ecf817be4d27e26c9d430fda156e6b25

ValuedMammal (approved):
reACK de2c44fc183b3b43816cfe6c868c8cea7d749d7b

Issue #1815, opened 2025-01-30
Add spent trait to improve sent_and_received()


matthiasdebernardini:
**Describe the enhancement**
Add a `TxAmountSpent` trait to improve how transaction spending amounts are calculated. The trait will add a `.spent()` method to tuples returned by `sent_and_received()`, calculating total amount spent including fees.

**Use case**
```rust
// Current: Manual calculation needed
let (sent, received) = wallet.sent_and_received(&tx);
let fee = wallet.calculate_fee(&tx)?;
let total = sent - received + fee; // Error-prone 

// Proposed: Clean trait implementation
let (sent, received) = wallet.sent_and_received(&tx);
let total = (sent, received, fee).spent(); // Clear and consistent
```

**Additional context**
- Non-breaking change that extends existing functionality
- Makes transaction cost calculations more intuitive
- Helps prevent calculation errors
- Small, focused improvement suitable for a first-time contributor

thunderbiscuit:
I had to review this for myself this morning because I've had issues in the past with this type of calculation (an amount was sent but it includes the fee, so it's not what the recipient(s) get).

Reading your line:
```rust
let total = sent - received + fee; // Error-prone
```

I'm not sure this is correct (unless I misunderstand what you intend to calculate). What do you mean by "total" here? The `Wallet::send_and_received()` method does include the fee. So the trickiness comes if you want to know exactly how much was sent to your recipients.

For example, if you have one recipient and you send 4200 sats to them and set a fee of 281 sats, while using a single 8000 sat UTXO, the `sent_and_received()` method should return the following, where the received is `(utxo - recipients amounts - fee)`.

```rust
sent: 8,000
received: 3,519
```

It doesn't give you how much was actually sent to the recipient (4200) but I'm not sure if that's what you're after, or your "total" above is meant be something else.

matthiasdebernardini:
I think thats a great point, right now I am trying to calculate net change in the wallet (which includes the fee). 

So it should include both, how many sats did the user put in the amount field when requesting the fee and by how many sats did the wallet balance change as a result of this transaction.

ValuedMammal:
By "total spent" aren't you referring to the `net_value`?

https://docs.rs/bdk_chain/latest/bdk_chain/indexer/keychain_txout/struct.KeychainTxOutIndex.html#method.net_value

YuriQueirozAndrade:
i can take this issue ?

khadar1020:
can I work on this

[...]

PR #1842, opened 2025-02-17
feat(wallet): Introduce `TxAmountSpend` trait for improved spend calculation

YuriQueirozAndrade:
### Description

This PR adds the new trait `TxAmountSpend` and a method `spend()` to improve the spend calculation. See [#1815](https://github.com/bitcoindevkit/bdk_wallet/issues/31) for more context.

### Notes to Reviewers
This PR adds `TxAmountSpend` in `utils.rs` and calls it in `mod.rs`. There are no breaking changes.

### Changelog Notice

- Added trait `TxAmountSpend`
- Added method `TxAmountSpend::spend()`

#### All Submissions:

* [x] I've signed all my commits.
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md).
* [x] I ran `cargo fmt` and `cargo clippy` before committing.

#### New Features:

* [x] I've added tests for the new feature.
* [x] I've added docs for the new feature.

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -80,7 +80,7 @@ pub use bdk_chain::Balance;
 pub use changeset::ChangeSet;
 pub use params::*;
 pub use persisted::*;
-pub use utils::IsDust;
+pub use utils::{IsDust, TxAmountSpend};
 
 /// A Bitcoin wallet
 ///
--- a/crates/wallet/src/wallet/utils.rs
+++ b/crates/wallet/src/wallet/utils.rs
@@ -152,8 +152,8 @@ mod test {
     // otherwise it's time-based
     pub(crate) const SEQUENCE_LOCKTIME_TYPE_FLAG: u32 = 1 << 22;
 
-    use super::{check_nsequence_rbf, shuffle_slice, IsDust};
-    use crate::bitcoin::{Address, Network, Sequence};
+    use super::{check_nsequence_rbf, shuffle_slice, IsDust, TxAmountSpend};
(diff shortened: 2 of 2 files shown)


ValuedMammal:
NACK. If you sent `sent`, then how could you have sent `sent + fee` ?

YuriQueirozAndrade:
> NACK. If you sent `sent`, then how could you have sent `sent + fee` ?

I might be making a mistake, but from what I understand about the issue, `sent_and_received` returns a tuple with two values, `sent` and `received` (which does not include fees). What we want is to calculate the total amount we will spend including fees, hence: `sent - received + fee.` Is that correct? Could you explain better what you are thinking exactly? Maybe I didn't understand it right.

evanlinjin:
NACK

I don't think having a trait here makes sense. Helper functions or creating another `Wallet` method will do (maybe need more discussion on the ticket?).

Also we need to convert to `SignedAmount` because the total may be negative.

Continue discussion on bitcoindevkit/bdk_wallet#31.

YuriQueirozAndrade:
I'm going to close this post, apparently there are more discussions ahead, sorry for the lost of time to those involved 😅, i'm going to gain more context about the project

ValuedMammal:
> I might be making a mistake, but from what I understand about the issue, `sent_and_received` returns a tuple with two values, `sent` and `received` (which does not include fees). What we want is to calculate the total amount we will spend including fees, hence: `sent - received + fee.` Is that correct? Could you explain better what you are thinking exactly? Maybe I didn't understand it right.

Good question. The fee is implicit in the fact that you received what you did, in other words you can think of the fee as a deduction on the amount you might have otherwise received (had there been no fee).

PR #1844, opened 2025-02-19
docs(wallet): Improve docs on `ChangeSet` persistence requirements

### LINKED ISSUE #9: Improve docs on `ChangeSet` persistence requirements
**Describe the enhancement**  
Currently, the docs for `bdk_wallet::ChangeSet` are not super clear on what requirements implementations of `WalletPersister` need to fulfill.

In Discord, we established that:
- Sub-`ChangeSet`s (i.e., local_chain, tx_graph, indexer) may get persisted independently/non-atomically. I.e., in case of a persistence-failure-and-restart scenario subsequent `persist` calls would deliver sub-`ChangeSet`s that are still mergable. This means that the persisted state wouldn't be corrupted by a persistence failure of one of the sub-`ChangeSet`s.
- However, the ordering of persistence calls might matter, as "(..) the localchain tip is used by block-by-block chain sources as a reference as to where to sync from, so I would persist that last", i.e., "I would write in this order: indexer, tx_graph, local_chain" (@evanlinjin)

It would be good to add all of these specifics and details to the `ChangeSet`/`WalletPersister` docs.
**Use case**  
Being able to implement `WalletPersister` without screwing up.

ValuedMammal:
This work in progress PR is meant to expand on and clarify the documentation for `bdk_wallet::ChangeSet`.

fixes bitcoindevkit/bdk_wallet#9

#### All Submissions:

### DIFF
--- a/crates/wallet/src/wallet/changeset.rs
+++ b/crates/wallet/src/wallet/changeset.rs
@@ -76,10 +76,10 @@ type IndexedTxGraphChangeSet =
 /// are likely to be added as new features and core components are implemented. Existing fields may
 /// be removed in future versions of the library.
 ///
-/// BDK reserves the right to introduce breaking changes to the [`ChangeSet`] structure in a major
-/// version release. API changes that affect the kinds of data persisted will be displayed
-/// prominently in the release notes of new versions. Users are advised to look for any such
-/// changes and update their application accordingly.
+/// The authors reserve the right to make breaking changes to the [`ChangeSet`] structure in
+/// a major version release. API changes affecting the types of data persisted will display
+/// prominently in the release notes. Users are advised to look for such changes and update their
+/// application accordingly.
 ///
 /// The resulting interface is designed to give the user more control of what to persist and when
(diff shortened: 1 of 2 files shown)


ValuedMammal:
@tnull Is there any additional info or insight that you think should be included in the docs regarding the wallet `ChangeSet`, `WalletPersister`, etc?

tnull:
> @tnull Is there any additional info or insight that you think should be included in the docs regarding the wallet `ChangeSet`, `WalletPersister`, etc?

Ah, thank you for this PR! I think it would be great if the compatibility guarantees of `ChangeSet` could be spelled out as explicitly as possible (probably in its own sub-section of the `ChangeSet` docs?). Is the type ever going to change? If so, what types of changes to the data type do users implementing their own persistence need to prepare for? How do they need to detect and handle them? Can users persist the sub-changesets individually? Will forward compatibility (i.e., downgrades) also be supported? etc.

ValuedMammal:
Reopened bitcoindevkit/bdk_wallet#198

PR #1851, opened 2025-02-24
Use destination groups instead of coins in coin_select to improve privacy

martinsaposnic:
In https://github.com/bitcoindevkit/bdk_wallet/issues/28 this PR from Bitcoin Core was mentioned https://github.com/bitcoin/bitcoin/pull/12257

This PR adds a param when building the tx called avoid_partial_spends. When true, it changes coin select to use output groups rather than outputs, where each output group corresponds to all outputs with the same destination.

It is a privacy improvement, as each time you spend some output, any other output that is publicly associated with the destination (address) will also be spent at the same time, at the cost of fee increase for cases where coin select without group restriction would find a more optimal set of coins (see example below).

For regular use without address reuse, this PR should have no effect on the user experience whatsoever; it only affects users who, for some reason, have multiple outputs with the same destination (i.e. address reuse).

Example: a node has four outputs linked to two addresses A and B:

- 1.0 btc to A
- 0.5 btc to A
- 1.0 btc to B
- 0.5 btc to B

The node sends 0.2 btc to C. Without -avoidpartialspends, the following coin selection will occur:

- 0.5 btc to A or B is picked
- 0.2 btc is output to C
- 0.3 - fee is output to (unique change address)

With -avoidpartialspends, the following will instead happen:

- Both of (0.5, 1.0) btc to A or B is picked (one or the other pair)
- 0.2 btc is output to C
- 1.3 - fee is output to (unique change address)

Assuming nobody sends to the address after you spend from it, you will only ever use one address once. The con is that the transaction becomes slightly larger in this case, because it is overpicking outputs to adhere to the no partial spending rule.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/coin_selection.rs
+++ b/crates/wallet/src/wallet/coin_selection.rs
@@ -117,6 +117,7 @@ use bitcoin::OutPoint;
 use bitcoin::TxIn;
 use bitcoin::{Script, Weight};
 
+use chain::bdk_core::collections::HashMap;
 use core::convert::TryInto;
 use core::fmt::{self, Formatter};
 use rand_core::RngCore;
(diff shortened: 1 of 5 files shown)


crates/wallet/src/wallet/coin_selection.rs
```diff
@@ -232,6 +232,10 @@ pub trait CoinSelectionAlgorithm: core::fmt::Debug {
         params: CoinSelectionParams<'_, R>,
     ) -> Result<CoinSelectionResult, InsufficientFunds>;
 }
+fn group_utxos_if_applies(utxos: Vec<WeightedUtxo>, _: bool) -> Vec<Vec<WeightedUtxo>> {
+    // No grouping: every UTXO is its own group.
+    return utxos.into_iter().map(|u| vec![u]).collect();
```
YuriQueirozAndrade: why `return utxos.into_iter().map(|u| vec![u]).collect();` and not `utxos.into_iter().map(|u| vec![u]).collect()` ?

notmandatory:
Thanks for working on this, might take some time to get it reviewed but I'll bring it up on our next dev call (on our discord server, open to everyone if you'd like to join).

martinsaposnic:
FYI, I'm working on a feature based on this PR. Nodes with avoid_partial_spends off will still try to avoid partial spending if the new transaction's fee is not higher than the original. PR coming soon

nymius:
Looking at past PRs, `OutputGroups` was also tried before, in bitcoindevkit/bdk#727.  [There was a structure to wrap for `WeightedUtxos` , named `OutputGroup` too.](https://github.com/bitcoindevkit/bdk/pull/727/files#diff-fd2a47357bd088bed85419710e86d949d1a172691db87c978be1336a725a42afR466)
However, maybe this is not the most valuable thing to work right now as it would be extra work to match with in the future on [`bdk-tx`](https://github.com/bitcoindevkit/bdk-tx), as we want to upstream it for `bdk_wallet`.

martinsaposnic:
> However, maybe this is not the most valuable thing to work right now as it would be extra work to match with in the future on [bdk-tx](https://github.com/bitcoindevkit/bdk-tx), as we want to upstream it for bdk_wallet.

closing and will work in the future

PR #1839, opened 2025-02-16
Introduce `evicted-at`/`last-evicted` timestamps

evanlinjin:
Partially Fixes #1740.
Replaces #1765.
Replaces #1811.

### Description

This PR allows the receiving structures (`bdk_chain`, `bdk_wallet`) to detect and evict incoming transactions that are double spent (cancelled).

We add a new field to `TxUpdate` (`TxUpdate::evicted_ats`), which in turn, updates the `last_evicted` timestamps that are tracked/persisted by `TxGraph`. This is similar to how `TxUpdate::seen_ats` update the `last_seen` timestamp in `TxGraph`. Transactions with a `last_evicted` timestamp higher than their `last_seen` timestamp are considered evicted.

`SpkWithExpectedTxids` is introduced in `SpkClient` to track expected `Txid`s for each `spk`. During a sync, if any `Txid`s from `SpkWithExpectedTxids` are not in the current history of an `spk` obtained from the chain source, those `Txid`s are considered missing. Support for `SpkWithExpectedTxids` has been added to both `bdk_electrum` and `bdk_esplora` chain source crates.

The canonicalization algorithm is updated to disregard transactions with a `last_evicted` timestamp greater than or equal to their `last_seen` timestamp, except in cases where transitivity rules apply.

### Notes to the reviewers

This PR does not fix #1740 for block-by-block chain source (such as `bdk_bitcoind_rpc`). This work is done in a separate PR (#1857).

### Changelog notice

* Add `TxUpdate::evicted_ats` which tracks transactions that have been replaced and are no longer present in mempool.
* Change `TxGraph` to track `last_evicted` timestamps. This is when a transaction is last marked as missing from the mempool.
* The canonicalization algorithm now disregards transactions with a `last_evicted` timestamp greater than or equal to it's `last_seen` timestamp, except when a canonical descendant exists due to rules of transitivity.
* Add `SpkWithExpectedTxids` in `spk_client` which keeps track of expected `Txid`s for each `spk`.
* Change `bdk_electrum` and `bdk_esplora` to understand `SpkWithExpectedTxids`.
* Add `SyncRequestBuilder::expected_txids_of_spk` method which adds an association between `txid`s and `spk`s.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

[...]

PR #1857, opened 2025-02-26
feat(rpc)!: Update `Emitter::mempool` to support `evicted_at`

### LINKED ISSUE #1740: Undetected double-spent

LagginTimes:
Fixes #1740

### Description

Work for this began as part of #1811, based on this [comment](https://github.com/bitcoindevkit/bdk/pull/1811#discussion_r1945941691).

`Emitter::mempool` now returns a `MempoolEvent` which provides the data for tracking `evicted_at`:

```
pub struct MempoolEvent {
    /// Unemitted transactions or transactions with ancestors that are unseen by the receiver.
    ///
    /// To understand the second condition, consider a receiver which filters transactions based on
    /// whether it alters the UTXO set of tracked script pubkeys. If an emitted mempool transaction
    /// spends a tracked UTXO which is confirmed at height `h`, but the receiver has only seen up to
    /// block of height `h-1`, we want to re-emit this transaction until the receiver has seen the
    /// block at height `h`.
    pub new_txs: Vec<(Transaction, u64)>,

    /// [`Txid`]s of all transactions that have been evicted from mempool.
    pub evicted_txids: HashSet<Txid>,

    /// The latest timestamp of when a transaction entered the mempool.
    ///
    /// This is useful for setting the timestamp for evicted transactions.
    pub latest_update_time: u64,
}
```

### Changelog notice

* Change `Emitter::mempool` to return `MempoolEvent`s which contain mempool-eviction data.
* Change `Emitter::client` to have more relaxed generic bounds. `C: Deref, C::Target: RpcApi` are the new bounds.
* Add conversion impls for `CanonicalTx` to `Txid`/`Arc<Transaction>`.
* Add `ChainPosition::is_unconfirmed` method.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bitcoind_rpc/src/lib.rs
+++ b/crates/bitcoind_rpc/src/lib.rs
@@ -10,8 +10,9 @@
 #![warn(missing_docs)]
 
 use bdk_core::{BlockId, CheckPoint};
-use bitcoin::{block::Header, Block, BlockHash, Transaction};
+use bitcoin::{Block, BlockHash, Transaction, Txid};
 use bitcoincore_rpc::bitcoincore_rpc_json;
+use std::collections::HashSet;
 
 pub mod bip158;
 
@@ -43,6 +44,16 @@ pub struct Emitter<'c, C> {
     /// The last emitted block during our last mempool emission. This is used to determine whether
     /// there has been a reorg since our last mempool emission.
     last_mempool_tip: Option<u32>,
+
+    /// A set of txids currently assumed to still be in the mempool.
+    ///
+    /// This is used to detect mempool evictions by comparing the set against the latest mempool
+    /// snapshot from bitcoind. Any txid in this set that is missing from the snapshot is considered
+    /// evicted.
+    ///
+    /// When the emitter emits a block, confirmed txids are removed from this set. This prevents
+    /// confirmed transactions from being mistakenly marked with an `evicted_at` timestamp.
+    expected_mempool_txids: HashSet<Txid>,
 }
 
 impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
@@ -53,7 +64,15 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
     ///
     /// `start_height` starts emission from a given height (if there are no conflicts with the
     /// original chain).
-    pub fn new(client: &'c C, last_cp: CheckPoint, start_height: u32) -> Self {
+    ///
+    /// `expected_mempool_txids` is the initial set of unconfirmed txids provided by the wallet.
+    /// This allows the [`Emitter`] to inform the wallet about relevant mempool evictions.
+    pub fn new(
+        client: &'c C,
+        last_cp: CheckPoint,
+        start_height: u32,
+        expected_mempool_txids: HashSet<Txid>,
+    ) -> Self {
         Self {
             client,
             start_height,
@@ -61,10 +80,18 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
             last_block: None,
             last_mempool_time: 0,
             last_mempool_tip: None,
+            expected_mempool_txids,
         }
     }
 
-    /// Emit mempool transactions, alongside their first-seen unix timestamps.
+    /// Emit mempool transactions and any evicted [`Txid`]s.
+    ///
+    /// This method returns a [`MempoolEvent`] containing the full transactions (with their
+    /// first-seen unix timestamps) that were emitted, and [`MempoolEvent::evicted_txids`] which are
+    /// any [`Txid`]s which were previously seen in the mempool and are now missing. Evicted txids
+    /// are only reported once the emitter’s checkpoint matches the RPC’s best block in both height
+    /// and hash. Until `next_block()` advances the checkpoint to tip, `mempool()` will always
+    /// return an empty `evicted_txids` set.
     ///
     /// This method emits each transaction only once, unless we cannot guarantee the transaction's
     /// ancestors are already emitted.
@@ -74,7 +101,7 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
     /// tracked UTXO which is confirmed at height `h`, but the receiver has only seen up to block
     /// of height `h-1`, we want to re-emit this transaction until the receiver has seen the block
     /// at height `h`.
-    pub fn mempool(&mut self) -> Result<Vec<(Transaction, u64)>, bitcoincore_rpc::Error> {
+    pub fn mempool(&mut self) -> Result<MempoolEvent, bitcoincore_rpc::Error> {
         let client = self.client;
 
         // This is the emitted tip height during the last mempool emission.
@@ -84,6 +111,38 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
             // `start_height` has been emitted.
             .unwrap_or(self.start_height.saturating_sub(1));
 
+        // Loop to make sure that the fetched mempool content and the fetched tip are consistent
+        // with one another.
+        let (raw_mempool, raw_mempool_txids, rpc_height, rpc_block_hash) = loop {
+            // Determine if height and hash matches the best block from the RPC. Evictions are deferred
+            // if we are not at the best block.
+            let height = client.get_block_count()?;
+            let hash = client.get_block_hash(height)?;
+
+            // Get the raw mempool result from the RPC client which will be used to determine if any
+            // transactions have been evicted.
+            let mp = client.get_raw_mempool_verbose()?;
+            let mp_txids: HashSet<Txid> = mp.keys().copied().collect();
+
+            if height == client.get_block_count()? && hash == client.get_block_hash(height)? {
+                break (mp, mp_txids, height, hash);
+            }
+        };
+
+        let at_tip =
+            rpc_height == self.last_cp.height() as u64 && rpc_block_hash == self.last_cp.hash();
+
+        // If at tip, any expected txid missing from raw mempool is considered evicted;
+        // if not at tip, we don't evict anything.
+        let evicted_txids: HashSet<Txid> = if at_tip {
+            self.expected_mempool_txids
+                .difference(&raw_mempool_txids)
+                .copied()
+                .collect()
+        } else {
+            HashSet::new()
+        };
+
         // Mempool txs come with a timestamp of when the tx is introduced to the mempool. We keep
         // track of the latest mempool tx's timestamp to determine whether we have seen a tx
         // before. `prev_mempool_time` is the previous timestamp and `last_time` records what will
@@ -91,8 +150,7 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
         let prev_mempool_time = self.last_mempool_time;
         let mut latest_time = prev_mempool_time;
 
-        let txs_to_emit = client
-            .get_raw_mempool_verbose()?
+        let new_txs = raw_mempool
             .into_iter()
             .filter_map({
                 let latest_time = &mut latest_time;
@@ -101,25 +159,25 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
                     if tx_time > *latest_time {
                         *latest_time = tx_time;
                     }
-
-                    // Avoid emitting transactions that are already emitted if we can guarantee
-                    // blocks containing ancestors are already emitted. The bitcoind rpc interface
-                    // provides us with the block height that the tx is introduced to the mempool.
-                    // If we have already emitted the block of height, we can assume that all
-                    // ancestor txs have been processed by the receiver.
+                    // Best-effort check to avoid re-emitting transactions we've already emitted.
+                    //
+                    // Complete suppression isn't possible, since a transaction may spend outputs
+                    // owned by the wallet. To determine if such a transaction is relevant, we must
+                    // have already seen its ancestor(s) that contain the spent prevouts.
+                    //
+                    // Fortunately, bitcoind provides the block height at which the transaction
+                    // entered the mempool. If we've already emitted that block height, we can
+                    // reasonably assume the receiver has seen all ancestor transactions.
                     let is_already_emitted = tx_time <= prev_mempool_time;
                     let is_within_height = tx_entry.height <= prev_mempool_tip as _;
                     if is_already_emitted && is_within_height {
                         return None;
                     }
-
                     let tx = match client.get_raw_transaction(&txid, None) {
                         Ok(tx) => tx,
-                        // the tx is confirmed or evicted since `get_raw_mempool_verbose`
                         Err(err) if err.is_not_found_error() => return None,
                         Err(err) => return Some(Err(err)),
                     };
-
                     Some(Ok((tx, tx_time as u64)))
                 }
             })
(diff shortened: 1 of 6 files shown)


evanlinjin:
Instead of rebasing, it's probably best to `git switch -c <new-branch>` from #1839 and cherry-pick on the new commits.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -84,24 +96,54 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
             // `start_height` has been emitted.
             .unwrap_or(self.start_height.saturating_sub(1));
 
+        // Get the raw mempool result from the RPC client.
+        let raw_mempool = client.get_raw_mempool_verbose()?;
+        let raw_mempool_txids: HashSet<Txid> = raw_mempool.keys().copied().collect();
+
+        // Determine the latest block height from the raw mempool data.
```
evanlinjin: Why would you determine the latest block height from the mempool data?

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -84,24 +96,41 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
             // `start_height` has been emitted.
             .unwrap_or(self.start_height.saturating_sub(1));
 
+        // Get the raw mempool result from the RPC client.
+        let raw_mempool = client.get_raw_mempool_verbose()?;
+        let raw_mempool_txids: HashSet<Txid> = raw_mempool.keys().copied().collect();
+
+        // Determine if height matches last emitted block.
+        let at_tip = client.get_block_count()? == self.last_cp.height() as u64;
```
evanlinjin: What if it's the same height but different block due to reorg?

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -112,23 +141,40 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
         self.last_mempool_time = latest_time;
         self.last_mempool_tip = Some(self.last_cp.height());
+        if at_tip {
+            self.expected_mempool_txids = new_txs.iter().map(|(tx, _)| tx.compute_txid()).collect();
+        } else {
+            self.expected_mempool_txids.extend(
+                new_txs
+                    .iter()
+                    .map(|(tx, _)| tx.compute_txid())
+                    .collect::<Vec<_>>(),
```
evanlinjin: I don't think a collect is necessary here.

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -731,3 +744,107 @@ fn no_agreement_point() -> anyhow::Result<()> {
...
+    assert_eq!(exp_spk_txids, vec![(spk, txid_1)]);
+
+    // Check that mempool emission contains evicted txid.
+    let mempool_event = emitter.mempool()?;
+    assert!(mempool_event.evicted_txids.contains(&txid_1));
+
+    // Update graph with evicted tx.
+    for txid in mempool_event.evicted_txids {
+        let _ = graph.insert_evicted_at(txid, seen_at);
+    }
```
evanlinjin: We need to filter out irrelevant `evicted_at`s. I.e. if the txid is not part of our wallet.

evanlinjin (changes requested):
Documentation is outdated, but I understand this is WIP.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -43,6 +44,9 @@ pub struct Emitter<'c, C> {
     /// The last emitted block during our last mempool emission. This is used to determine whether
     /// there has been a reorg since our last mempool emission.
     last_mempool_tip: Option<u32>,
+
+    /// Expected mempool txs. TODO: Docs.
```
evanlinjin: Are you planning to address this TODO?

LagginTimes: Oops, looks like I goofed the rebase and undid all my changes to `lib.rs`... should be fixed now.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -53,18 +58,34 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
     ///
     /// `start_height` starts emission from a given height (if there are no conflicts with the
     /// original chain).
-    pub fn new(client: &'c C, last_cp: CheckPoint, start_height: u32) -> Self {
+    ///
+    /// `expected_mempool_txids` is the initial set of unconfirmed txids. Once at tip, any that are
+    /// no longer in mempool are marked evicted.
```
evanlinjin:
```suggestion
    /// `expected_mempool_txids` is the initial set of unconfirmed txids provided by the wallet. This
    /// allows the [`Emitter`] to inform the wallet about relevant mempool evictions.
```

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -43,6 +44,10 @@ pub struct Emitter<'c, C> {
     /// The last emitted block during our last mempool emission. This is used to determine whether
     /// there has been a reorg since our last mempool emission.
     last_mempool_tip: Option<u32>,
+
+    /// Unconfirmed txids that are expected to appear in mempool. This is used to determine if any
+    /// known txids have been evicted.
```
evanlinjin:
```suggestion
/// A set of txids that we currently believe are still in the mempool.
///
/// This is used to detect mempool evictions by comparing the set against the latest
/// mempool snapshot from bitcoind. Any txid in this set that is missing from the
/// snapshot is considered evicted.
///
/// When the emitter emits a block, confirmed txids are removed from this set. This prevents confirmed transactions from being
/// mistakenly marked with an `evicted_at` timestamp.
```

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -53,18 +58,34 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
             last_block: None,
             last_mempool_time: 0,
             last_mempool_tip: None,
+            expected_mempool_txids,
         }
     }
 
-    /// Emit mempool transactions, alongside their first-seen unix timestamps.
+    /// Emit mempool transactions and any evicted [`Txid`]s. Returns a `latest_update_time` which is
+    /// used for setting the timestamp for evicted transactions.
```
evanlinjin:
```suggestion
    /// Emit mempool transactions and any evicted [`Txid`]s.
```

I don't think this belongs here.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -112,14 +158,17 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
                     let tx = match client.get_raw_transaction(&txid, None) {
                         Ok(tx) => tx,
-                        // the tx is confirmed or evicted since `get_raw_mempool_verbose`
-                        Err(err) if err.is_not_found_error() => return None,
+                        Err(err) if err.is_not_found_error() => {
+                            // If at tip and the transaction isn't found, mark it as evicted.
+                            if at_tip {
+                                evicted_txids.insert(txid);
+                            }
+                            return None;
```
evanlinjin: It's also possible that a new block is found while we were looping through the mempool txs.

If this tx is confirmed in the new block (and therefore no longer exists in the mempool), the correct behavior is just to ignore it otherwise we may emit a false `evicted_at` for that tx here.

If the tx is actually evicted from the mempool after the `client.get_raw_mempool_verbose()` call, we will still detect the eviction will another `.mempool` call. 

I think we should retain the old logic.

evanlinjin (changes requested):
I haven't finished reviewing yet. Some thoughts:

1. We probably need something like `insert_relevant_evicted_ats(&mut self, evicted_txs: impl IntoIterator<Item = (Txid, u64)>)` on `TxGraph` for convenience.

2. The old best-effort-avoid-reemitting-mempool-txs logic might actually be replaceable by caching `Arc<Transaction>` in `expected_mempool_txids`.

    ```rust
    prior_mempool_txs: HashMap<Txid, Arc<Transaction>>,
    ```

    The idea is we always emit the whole mempool but it is still memory-efficient.

For `2.`, I guess we can tackle that in a separate PR.

crates/bitcoind_rpc/tests/test_emitter.rs
```diff
@@ -731,3 +744,109 @@ fn no_agreement_point() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn test_expect_tx_evicted() -> anyhow::Result<()> {
```
evanlinjin, line 777: Would you be able to add a brief summary at the top of the test (e.g. as a doc comment) explaining what the test is validating?

Without a summary, it’s a bit hard to grasp the intent of the test without reading through every line and trying to hold the context in mind. A short explanation would really help with readability, especially for larger tests like this one.

evanlinjin (commented):
The logic looks solid.

There’s room for improvement in documentation clarity across tests, inline code comments, and public doc comments:

- **Code comments**: Focus on the _why_ behind the logic, rather than describing _what_ the code is doing—since that’s usually self-evident from the code itself.
- **Tests**: For larger tests, consider adding a summary doc comment at the top. This helps readers quickly understand what the test is verifying without having to trace through each line and reconstruct the intent.
- **Public doc comments**: Prioritize what’s important for the _caller_ to know—avoid going into implementation details unless they directly impact usage.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -84,24 +105,49 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
             .filter_map({
                 let latest_time = &mut latest_time;
+                let evicted_txids = &mut evicted_txids;
                 move |(txid, tx_entry)| -> Option<Result<_, bitcoincore_rpc::Error>> {
                     let tx_time = tx_entry.time as usize;
                     if tx_time > *latest_time {
                         *latest_time = tx_time;
                     }
-
                     // Avoid emitting transactions that are already emitted if we can guarantee
```
evanlinjin: Could you help me replace this code comment with this?

```rust
                    // Best-effort check to avoid re-emitting transactions we've already emitted.
                    //
                    // Complete suppression isn't possible, since a transaction may spend outputs
                    // owned by the wallet. To determine if such a transaction is relevant, we must
                    // have already seen its ancestor(s) that contain the spent prevouts.
                    //
                    // Fortunately, bitcoind provides the block height at which the transaction
                    // entered the mempool. If we've already emitted that block height, we can
                    // reasonably assume the receiver has seen all ancestor transactions.
```

crates/chain/src/tx_graph.rs
```diff
@@ -777,6 +777,27 @@ impl<A: Anchor> TxGraph<A> {
...
+    /// `evicted_at` value, it is updated only if the new value is greater.
+    pub fn insert_relevant_evicted_ats(
+        &mut self,
+        txids: impl IntoIterator<Item = Txid>,
+        evicted_at: u64,
+    ) -> ChangeSet<A> {
+        let mut changeset = ChangeSet::default();
+        for txid in txids {
+            // Only record evictions for transactions the graph is tracking.
+            if self.get_tx_node(txid).is_some() {
```
```suggestion
            if self.txs.contains(txid) {
```

examples/example_bitcoind_rpc_polling/src/main.rs
```diff
@@ -287,7 +286,13 @@ fn main() -> anyhow::Result<()> {
...
                     }
                     Emission::Mempool(mempool_txs) => {
-                        let graph_changeset = graph.batch_insert_relevant_unconfirmed(mempool_txs);
+                        let mut graph_changeset =
+                            graph.batch_insert_relevant_unconfirmed(mempool_txs.new_txs.clone());
+                        for txid in mempool_txs.evicted_txids {
+                            graph_changeset.merge(
+                                graph.insert_evicted_at(txid, mempool_txs.latest_update_time),
+                            );
+                        }
```
evanlinjin: Needs update to use `insert_relevant_evicted_ats`

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -145,6 +145,23 @@ where
         }
     }
 
+    /// Inserts the given `evicted_at` for a set of `txid`s that the graph is tracking.
+    ///
+    /// The `evicted_at` timestamp represents the last known time when the transaction was observed
+    /// to be missing from the mempool. If `txid` was previously recorded with an earlier
+    /// `evicted_at` value, it is updated only if the new value is greater.
+    pub fn insert_relevant_evicted_ats(
```
evanlinjin: For consistency with other methods in `IndexedTxGraph`, I propose the following:

1. Rename the method to `batch_insert_relevant_evicted_at`.
2. Update the method signature to:  
   `batch_insert_relevant_evicted_at(&mut self, evicted_ats: impl IntoIterator<Item = (Txid, u64)>)`

This API more directly reflects the internal logic — applying individual `evicted_at` timestamps to each `txid`.

To make this integrate better with `MempoolEvent`, we could either:

1. Change the `MempoolEvent::evicted_txids` field to a `HashMap<Txid, u64>`, and remove the `latest_update_time` field.
2. Add a helper method to `MempoolEvent` that returns an `impl ExactSizeIterator<Item = (Txid, u64)>`.

Option 2 might be more direct and would require fewer changes.

Let me know what you think.

evanlinjin (commented):
Looks good, just a final request.

crates/chain/src/indexed_tx_graph.rs
LagginTimes: I think option 2 looks good to me. Let me know if the implementation looks good to you.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -84,15 +110,40 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
+
+        // Determine if height and hash matches the best block from the RPC. Evictions are deferred
+        // if we are not at the best block.
+        let height = client.get_block_count()?;
+        let at_tip = if height != self.last_cp.height() as u64 {
+            false
+        } else {
+            // Verify if block hash matches in case of re-org.
+            client.get_block_hash(height)? == self.last_cp.hash()
+        };
```
evanlinjin: We want to make sure that the fetched mempool contents and the fetched tip are always guaranteed to be consistent with one another. I.e. in other words, txs returned by `get_raw_mempool_verbose` should not exist in the block of `get_block_hash`.

The current logic does not enforce this. There is a slim chance the node might find a new block between the `get_raw_mempool_verbose` call and the `get_block_hash` call. If this happens, the current logic will advertise confirmed transactions as evicted.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -53,18 +64,33 @@ impl<'c, C: bitcoincore_rpc::RpcApi> Emitter<'c, C> {
...
     }
 
-    /// Emit mempool transactions, alongside their first-seen unix timestamps.
+    /// Emit mempool transactions and any evicted [`Txid`]s.
+    ///
+    /// This method returns a [`MempoolEvent`] containing the full transactions (with their
+    /// first-seen unix timestamps) that were emitted, and [`MempoolEvent::evicted_txids`] which are
+    /// any [`Txid`]s which were previously expected and are now missing from the mempool. Note that
+    /// [`Txid`]s are only evicted if the emitter is at the chain tip with the same height and hash
+    /// as the best block from the RPC.
```
evanlinjin: "Previously expected" can be a bit unclear. Suggestion:  
**"Were previously seen in the mempool and are now missing."**

Also, the content in the `Note that...` sentence is important for the caller to understand, but it's not stated clearly enough.  
Specifically, the caller should be made aware that if they don't call `next_block`, they will never be informed about transaction evictions.

evanlinjin (commented):
Apologies for the incremental reviews — it's been eye-opening for me as well how tricky the `Emitter` is to reason about. Let's aim to simplify it in follow-up PRs.

One thing I’ve come to realize: the `next_header` method is inconsistent with the eviction-detection logic we're introducing via `next_block` and `mempool`. I suggest removing `next_header` entirely — everything it does can already be achieved through `next_block`.

examples/example_bitcoind_rpc_polling/src/main.rs
```diff
@@ -141,7 +139,7 @@ fn main() -> anyhow::Result<()> {
 
             let chain_tip = chain.lock().unwrap().tip();
             let rpc_client = rpc_args.new_client()?;
-            let mut emitter = Emitter::new(&rpc_client, chain_tip, fallback_height);
+            let mut emitter = Emitter::new(&rpc_client, chain_tip, fallback_height, HashSet::new());
```
evanlinjin: We should probably provide the initial unconfirmed contents here.

evanlinjin: I fixed this. Also made some API changes to reduce the friction of populating `expected_mempool_txids` from `bdk_chain`.

evanlinjin (approved):
self-ACK e1c53d737b396474cc1015183f3bb1a43afd12d0

ValuedMammal:
@LagginTimes Can you rebase on master and also squash 6e1178ca into the previous commit?

ValuedMammal (approved):
ACK 8513d83a2ef2fe748e9007984187a7f08d256ee6

ValuedMammal:
Thanks @LagginTimes 

I didn't mean to become author on 28ef7c9a 😅

PR #1856, opened 2025-02-25
docs: fix dead links in `README.md`

rebustron:
### Description

Hi! I fixed outdated and incorrect paths in `README.md`, ensuring all internal links point to the correct locations.

### Changes:
- **Removed:** `persist` crate reference #1454 
- **Fixed:** Incorrect paths for example projects:
  - `wallet_esplora_blocking` → `example_wallet_esplora_blocking`
  - `wallet_esplora_async` → `example_wallet_esplora_async`
  - `wallet_electrum` → `example_wallet_electrum`

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [ ] I ran `cargo fmt` and `cargo clippy` before committing

#### New Features:

* [ ] I've added tests for the new feature
* [ ] I've added docs for the new feature

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [ ] I'm linking the issue being fixed by this PR

PR #1861, opened 2025-02-28
Update README.md

nnsW3:
Hello
I fixed 1 outdated link.
Br, Elias.


oleonardolima:
Thanks, but it's actually a dead link and should be removed, which is covered by https://github.com/bitcoindevkit/bdk/pull/1856

PR #1863, opened 2025-03-01
Delegate rustqlite connection persister to transaction persister

casey:
This is a small refactor that hopefully improves readability slightly.

Currently, the rusqlite connection persister and transaction persister are entirely different implementations, even though the connection persister just creates a new transaction, does what the transaction persister does, and then commits.

This makes the relationship a little clearly by explicitly delegating the connection persister implementation to the transaction persister implementation, so the reader doesn't need to manually compare the two implementations to see how they're related.

### DIFF
--- a/crates/wallet/src/wallet/persisted.rs
+++ b/crates/wallet/src/wallet/persisted.rs
@@ -272,16 +272,15 @@ impl WalletPersister for bdk_chain::rusqlite::Connection {
     type Error = bdk_chain::rusqlite::Error;
 
     fn initialize(persister: &mut Self) -> Result<ChangeSet, Self::Error> {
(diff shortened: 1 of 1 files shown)


evanlinjin:
Thanks for the PR! I would argue that the original was clearer because it more directly expressed what was happening, instead of having the reader go through one more layer of methods.

casey:
I can see that. It's a trade off: Either you see what it's doing, or you see that it's the same as the other persister. In my opinion, knowing that it's the same as the other persister is more valuable.

ValuedMammal:
Closing as it is now stale.

I opened bitcoindevkit/bdk_wallet#279.

PR #1874, opened 2025-03-06
fix(wallet): allow PersistedWallet to be Send + Sync

### LINKED ISSUE #1873: Make Sqlite PersistedWallet<Connection> Send+Sync
**Describe the bug**  

Using `PersistedWallet<rusqlite::Connection>` is not thread-safe because the `PersistedWallet.marker` because the `rusqlite::Connection` type is `!Sync`. 

**To Reproduce**  

Try to use a `PersistedWallet<rusqlite::Connection>` where a Send + Sync type is required.

**Expected behavior**  

Since `PersistedWallet.marker` is only `PhantomData` used by the type system it should not prevent `PersistedWallet` from being Send + Sync. 

**Build environment**  
 - BDK tag/commit: 1.2.0
 - OS+version: any
 - Rust/Cargo version: any 
 - Rust/Cargo target: any

**Additional context**

This issue was identified by @praveenperera who also suggested a solution. See discord chat:

https://discord.com/channels/753336465005608961/978744259693916230/1346317604335386705

notmandatory:
### Description

fixes #1873 based on solution proposed by @praveenperera . 

### Notes to the reviewers

This is not a breaking change since it's only changing the internal `_marker` field's type.

### Changelog notice

- Fix PersistedWallet to be Send + Sync, even when used with a !Sync persister type such as rusqlite::Connection.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/wallet/src/wallet/persisted.rs
+++ b/crates/wallet/src/wallet/persisted.rs
@@ -120,7 +120,7 @@ pub trait AsyncWalletPersister {
 #[derive(Debug)]
 pub struct PersistedWallet<P> {
     inner: Wallet,
-    marker: PhantomData<P>,
+    _marker: PhantomData<fn(&mut P)>,
 }
 
 impl<P> Deref for PersistedWallet<P> {
@@ -155,7 +155,7 @@ impl<P: WalletPersister> PersistedWallet<P> {
         }
         Ok(Self {
             inner,
-            marker: PhantomData,
+            _marker: PhantomData,
         })
     }
 
@@ -169,7 +169,7 @@ impl<P: WalletPersister> PersistedWallet<P> {
             .map(|opt| {
                 opt.map(|inner| PersistedWallet {
                     inner,
-                    marker: PhantomData,
+                    _marker: PhantomData,
                 })
             })
             .map_err(LoadWithPersistError::InvalidChangeSet)
@@ -214,7 +214,7 @@ impl<P: AsyncWalletPersister> PersistedWallet<P> {
         }
         Ok(Self {
             inner,
-            marker: PhantomData,
+            _marker: PhantomData,
         })
     }
 
@@ -230,7 +230,7 @@ impl<P: AsyncWalletPersister> PersistedWallet<P> {
             .map(|opt| {
                 opt.map(|inner| PersistedWallet {
                     inner,
-                    marker: PhantomData,
+                    _marker: PhantomData,
                 })
             })
             .map_err(LoadWithPersistError::InvalidChangeSet)
--- a/crates/wallet/tests/wallet.rs
+++ b/crates/wallet/tests/wallet.rs
@@ -12,7 +12,9 @@ use bdk_wallet::psbt::PsbtUtils;
 use bdk_wallet::signer::{SignOptions, SignerError};
 use bdk_wallet::test_utils::*;
 use bdk_wallet::tx_builder::AddForeignUtxoError;
-use bdk_wallet::{AddressInfo, Balance, ChangeSet, Update, Wallet, WalletPersister, WalletTx};
+use bdk_wallet::{
+    AddressInfo, Balance, ChangeSet, PersistedWallet, Update, Wallet, WalletPersister, WalletTx,
+};
 use bdk_wallet::{KeychainKind, LoadError, LoadMismatch, LoadWithPersistError};
 use bitcoin::constants::{ChainHash, COINBASE_MATURITY};
 use bitcoin::hashes::Hash;
@@ -4209,6 +4211,7 @@ fn test_tx_cancellation() {
 fn test_thread_safety() {
     fn thread_safe<T: Send + Sync>() {}
     thread_safe::<Wallet>(); // compiles only if true
+    thread_safe::<PersistedWallet<bdk_chain::rusqlite::Connection>>();
 }
 
 #[test]


YuriQueirozAndrade:
ACK, i run build, test, clippy and fmt. All looks goods

oleonardolima (commented):
I'm quite divided on this one, still taking a look in SQLite docs, but it doesn't feel right to override the `Send + Sync` implementation for any given persister.

I think there's no need for this change and it should inherit the `Send + Sync` implementation as it's [patterns](https://doc.rust-lang.org/nomicon/phantom-data.html#table-of-phantomdata-patterns) and use a Mutex instead, there's a few issues/discussions on rusqlite about this and it seems their decision is related on how sqlite flags are used, see https://github.com/rusqlite/rusqlite/issues/188.

praveenperera:
@oleonardolima when you call `persist` you have to pass in the `Connection` if this `Connection` is stored in the same struct as the Wallet you still need to wrap the `Connection` in a mutex.

Without this change both the `Connection` and the `Wallet` would both need to be wrapped in a mutex for no reason.

Example without this change 

```rust
#[derive(Debug, uniffi::Object)]
pub struct Wallet {
    pub id: WalletId,
    pub network: Network,
    pub bdk: Mutex<bdk_wallet::PersistedWallet<Connection>>,
    pub metadata: WalletMetadata,
    db: Mutex<Connection>,
}
```

Example with this change:

```rust
#[derive(Debug, uniffi::Object)]
pub struct Wallet {
    pub id: WalletId,
    pub network: Network,
    pub bdk: bdk_wallet::PersistedWallet<Connection>,
    pub metadata: WalletMetadata,
    db: Mutex<Connection>,
}
```

There isn't a safety issue here because the wallet doesn't actually store the `Connection` which is what is not thread safe. The only reason the `Connection` is included in the type at all is to separate it from other persister types.

crates/wallet/src/wallet/persisted.rs
```diff
@@ -120,7 +120,7 @@ pub trait AsyncWalletPersister {
 #[derive(Debug)]
 pub struct PersistedWallet<P> {
     inner: Wallet,
-    marker: PhantomData<P>,
+    _marker: PhantomData<fn(P) -> P>,
```
evanlinjin:
```suggestion
    _marker: PhantomData<*const P>,
```

According to @LLFourn

notmandatory: I tried this approach and get this error running `test_thread_safety`:
```
error[E0277]: `*const Connection` cannot be sent between threads safely
    --> crates/wallet/tests/wallet.rs:4215:19
     |
4215 |     thread_safe::<PersistedWallet<bdk_chain::rusqlite::Connection>>();
     |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `*const Connection` cannot be sent between threads safely
     |
     = help: within `PersistedWallet<Connection>`, the trait `Send` is not implemented for `*const Connection`
     = help: the trait `Send` is implemented for `Connection`
note: required because it appears within the type `PhantomData<*const Connection>`
    --> /Users/steve/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/marker.rs:757:12
     |
757  | pub struct PhantomData<T: ?Sized>;
     |            ^^^^^^^^^^^
note: required because it appears within the type `PersistedWallet<Connection>`
    --> /Users/steve/git/notmandatory/bdk/crates/wallet/src/wallet/persisted.rs:121:12
     |
121  | pub struct PersistedWallet<P> {
     |            ^^^^^^^^^^^^^^^
note: required by a bound in `thread_safe`
    --> crates/wallet/tests/wallet.rs:4212:23
     |
4212 |     fn thread_safe<T: Send + Sync>() {}
     |                       ^^^^ required by this bound in `thread_safe`
```

praveenperera: `PhantomData<*const P>` makes it `!Sync`: https://doc.rust-lang.org/nomicon/phantom-data.html#table-of-phantomdata-patterns

notmandatory:
According to `Llama 3.3 70B` I should use `PhantomData<fn(&mut T)>`:

"In this case, you want to ensure that your struct remains Send and Sync, even if the functions implemented on it have a &mut T parameter and T is not Sync.

To achieve this, you can use PhantomData<fn(&mut T)> instead of PhantomData<fn(T)>. This tells Rust that your struct is a "consumer" of &mut T, but it does not own T and therefore should not be tied to its lifetime."

notmandatory:
I updated the code per comment above and also included some of the explanation in the commit message.

praveenperera:
@notmandatory Perplexity seems to agree too

Screenshot 2025-03-06 at 7 15 55 PM

notmandatory:
I pushed another commit to fix the CI issue. Ring just (4hrs ago) bumped their MSRV from 1.63 to 1.66 with a patch release 😞 (see briansmith/ring#2449).

praveenperera:
@notmandatory I think in our case since we aren't dealing with lifetimes of any unsafe things, I think all three are practically equal, all of this compiles. I think the AIs are wrong.

```rust
PhantomData<fn(P) -> P>
PhantomData<fn(&mut P)>
PhantomData<fn(P)>
```

```rust
use std::marker::PhantomData;
use rusqlite::Connection;

pub struct PersistedWallet<P> {
    _marker: PhantomData<fn(&mut P)>,
}

impl<P> PersistedWallet<P> {
    pub fn new() -> Self {
        Self {
            _marker: PhantomData,
        }
    }
    pub fn persist(&self, _wallet: &mut P) {}
    pub fn persist_2(&self, _wallet: P) {}
}

pub struct PersistedWallet2<P> {
    _marker: PhantomData<fn(P)>,
}

impl<P> PersistedWallet2<P> {
    pub fn new() -> Self {
        Self {
            _marker: PhantomData,
        }
    }
    pub fn persist(&self, _wallet: &mut P) {}
    pub fn persist_2(&self, _wallet: P) {}
}

pub struct PersistedWallet3<P> {
    _marker: PhantomData<fn(P) -> P>,
}

impl<P> PersistedWallet3<P> {
    pub fn new() -> Self {
        Self {
            _marker: PhantomData,
        }
    }
    pub fn persist(&self, _wallet: &mut P) {}
    pub fn persist_2(&self, _wallet: P) {}
}

pub type Wallet = PersistedWallet<Connection>;
pub type Wallet2 = PersistedWallet2<Connection>;
pub type Wallet3 = PersistedWallet3<Connection>;

fn thead_safe<T>(_wallet: T)
where
    T: Sync + Sync,
{
}

fn main() {
    thead_safe(Wallet::new());
    thead_safe(Wallet2::new());
    thead_safe(Wallet3::new());
}
```

notmandatory:
If all are safe and enable the same functionality then I think the one suggested by the AIs is still the best since it communicates that PersistWallet "consumes" the &mut P type in it's functions without owning P. But happy to change it to one of the other options if there's some compelling reason to.

praveenperera:
That makes sense to me

crates/wallet/src/wallet/persisted.rs
evanlinjin: @notmandatory Sorry yeah my suggestion does the complete opposite of what we need.

evanlinjin (approved):
ACK cdba39d2b831e807f54e18d8f8bc14022424911d

TIL something about `PhantomData`.

crates/wallet/tests/wallet.rs
```diff
@@ -4209,6 +4211,8 @@ fn test_tx_cancellation() {
 fn test_thread_safety() {
     fn thread_safe<T: Send + Sync>() {}
     thread_safe::<Wallet>(); // compiles only if true
+    #[cfg(feature = "rusqlite")]
```
ValuedMammal:
```suggestion
```

ValuedMammal (commented):
ACK, needs a rebase now that #1876 is merged

crates/wallet/tests/wallet.rs
notmandatory: ~~I think I need this `cfg` or else testing `bdk_wallet` with `--no-default-features` won't compile.~~ Nevermind, you're right we don't need this since `rusqlite` is a dev dependency.

notmandatory:
@ValuedMammal I rebased this PR, OK to merge now?

ValuedMammal (approved):
ACK 14251a43be73796f0db98cee8a0b5e88e4b4c51a

PR #1881, opened 2025-03-09
fix(wallet): no debug on Display implementations

### LINKED ISSUE #1868: Display implementations should not expose `Debug` output
Currently, the `Display` implementation for `LoadError` displays`Debug` output for `LoadError::MissingDescriptor` and `LoadError::Mismatch`:

```rust
impl fmt::Display for LoadError {
        match self {
            LoadError::Descriptor(e) => e.fmt(f),
            LoadError::MissingNetwork => write!(f, "loaded data is missing network type"),
            LoadError::MissingGenesis => write!(f, "loaded data is missing genesis hash"),
            LoadError::MissingDescriptor(k) => {
                write!(f, "loaded data is missing descriptor for keychain {k:?}")
            }
            LoadError::Mismatch(mismatch) => write!(f, "data mismatch: {mismatch:?}"),
        }
    }
}
```

I think this is generally not a good practice. `Debug` output is automatically generated, and is not as nice to read as a hand-written `Display` implementation's output.

I noticed this in a test we were loading a wallet with the incorrect network, and instead of getting a nice error message, we got `data mismatch: Network { loaded: Network::Testnet, expected: Network::Bitcoin }`. Our project never exposes `Debug` output to the user, and we'd like to avoid doing this if we use BDK, but still be able to expose BDK errors to the user via `Display`.

### LINKED ISSUE #1933: Display implementations should not expose `Debug` output
Currently, the `Display` implementation for `LoadError` displays`Debug` output for `LoadError::MissingDescriptor` and `LoadError::Mismatch`:

```rust
impl fmt::Display for LoadError {
        match self {
            LoadError::Descriptor(e) => e.fmt(f),
            LoadError::MissingNetwork => write!(f, "loaded data is missing network type"),
            LoadError::MissingGenesis => write!(f, "loaded data is missing genesis hash"),
            LoadError::MissingDescriptor(k) => {
                write!(f, "loaded data is missing descriptor for keychain {k:?}")
            }
            LoadError::Mismatch(mismatch) => write!(f, "data mismatch: {mismatch:?}"),
        }
    }
}
```

I think this is generally not a good practice. `Debug` output is automatically generated, and is not as nice to read as a hand-written `Display` implementation's output.

I noticed this in a test we were loading a wallet with the incorrect network, and instead of getting a nice error message, we got `data mismatch: Network { loaded: Network::Testnet, expected: Network::Bitcoin }`. Our project never exposes `Debug` output to the user, and we'd like to avoid doing this if we use BDK, but still be able to expose BDK errors to the user via `Display`.

luisschwab:
### Description
Closes bitcoindevkit/bdk#1933.

#### All Submissions:

### DIFF
--- a/crates/wallet/src/types.rs
+++ b/crates/wallet/src/types.rs
@@ -12,6 +12,7 @@
 use alloc::boxed::Box;
 use chain::{ChainPosition, ConfirmationBlockTime};
 use core::convert::AsRef;
+use core::fmt;
 
 use bitcoin::transaction::{OutPoint, Sequence, TxOut};
 use bitcoin::{psbt, Weight};
@@ -33,6 +34,15 @@ impl KeychainKind {
         match self {
             KeychainKind::External => b'e',
             KeychainKind::Internal => b'i',
+        }
+    }
+}
+
+impl fmt::Display for KeychainKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            KeychainKind::External => write!(f, "External"),
+            KeychainKind::Internal => write!(f, "Internal"),
         }
     }
 }
--- a/crates/wallet/src/wallet/error.rs
+++ b/crates/wallet/src/wallet/error.rs
@@ -112,7 +112,7 @@ impl fmt::Display for CreateTxError {
             Self::Descriptor(e) => e.fmt(f),
             Self::Policy(e) => e.fmt(f),
             CreateTxError::SpendingPolicyRequired(keychain_kind) => {
-                write!(f, "Spending policy required: {:?}", keychain_kind)
+                write!(f, "Spending policy required: {}", keychain_kind)
             }
             CreateTxError::Version0 => {
                 write!(f, "Invalid version `0`")
@@ -127,12 +127,12 @@ impl fmt::Display for CreateTxError {
                 requested,
                 required,
             } => {
-                write!(f, "TxBuilder requested timelock of `{:?}`, but at least `{:?}` is required to spend from this script", required, requested)
+                write!(f, "TxBuilder requested timelock of `{}`, but at least `{}` is required to spend from this script", required, requested)
             }
             CreateTxError::RbfSequenceCsv { sequence, csv } => {
                 write!(
                     f,
-                    "Cannot enable RBF with nSequence `{:?}` given a required OP_CSV of `{:?}`",
+                    "Cannot enable RBF with nSequence `{}` given a required OP_CSV of `{}`",
                     sequence, csv
                 )
             }
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -194,9 +194,9 @@ impl fmt::Display for LoadError {
             LoadError::MissingNetwork => write!(f, "loaded data is missing network type"),
             LoadError::MissingGenesis => write!(f, "loaded data is missing genesis hash"),
             LoadError::MissingDescriptor(k) => {
-                write!(f, "loaded data is missing descriptor for keychain {k:?}")
+                write!(f, "loaded data is missing descriptor for {k} keychain")
             }
-            LoadError::Mismatch(mismatch) => write!(f, "data mismatch: {mismatch:?}"),
+            LoadError::Mismatch(e) => write!(f, "{e}"),
         }
     }
 }
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -232,6 +232,44 @@ pub enum LoadMismatch {
     },
 }
 
+impl fmt::Display for LoadMismatch {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            LoadMismatch::Network { loaded, expected } => {
+                write!(
+                    f,
+                    "Network mismatch: loaded {}, expected {}",
+                    loaded, expected
+                )
+            }
+            LoadMismatch::Genesis { loaded, expected } => {
+                write!(
+                    f,
+                    "Genesis hash mismatch: loaded {}, expected {}",
+                    loaded, expected
+                )
+            }
+            LoadMismatch::Descriptor {
+                keychain,
+                loaded,
+                expected,
+            } => {
+                write!(
+                    f,
+                    "Descriptor mismatch for {} keychain: loaded {}, expected {}",
+                    keychain,
+                    loaded
+                        .as_ref()
+                        .map_or("None".to_string(), |d| d.to_string()),
+                    expected
+                        .as_ref()
+                        .map_or("None".to_string(), |d| d.to_string())
+                )
+            }
+        }
+    }
+}
+
 impl From<LoadMismatch> for LoadError {
     fn from(mismatch: LoadMismatch) -> Self {
         Self::Mismatch(mismatch)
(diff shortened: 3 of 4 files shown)


crates/wallet/src/wallet/error.rs
110CodingP, line 1: I noticed that the Debug formats on lines 130 and 135 have been left as it is. Was this intentional? I'm still new to Rust and BDK, so might be missing something. Thanks!

110CodingP:
Although this PR is scoped to the wallet, I was wondering if we should also remove the Debug outputs from Display implementations in keychain_txout.rs (line 798), tx_graph.rs (line 228), spk_client.rs (line 24), and file_store/src/lib.rs (line 25)?

crates/wallet/src/wallet/error.rs
luisschwab: Yes, check out the `Debug` and `Display` implementations for `LockTime`:

```rs
impl fmt::Debug for LockTime {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        use LockTime::*;

        match *self {
            Blocks(ref h) => write!(f, "{} blocks", h),
            Seconds(ref t) => write!(f, "{} seconds", t),
        }
    }
}

impl fmt::Display for LockTime {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        use LockTime::*;

        if f.alternate() {
            match *self {
                Blocks(ref h) => write!(f, "block-height {}", h),
                Seconds(ref t) => write!(f, "block-time {} (seconds since epoch)", t),
            }
        } else {
            match *self {
                Blocks(ref h) => fmt::Display::fmt(h, f),
                Seconds(ref t) => fmt::Display::fmt(t, f),
            }
        }
    }
}
```
`Display` doesn't fit our need. Since this is a type from `rust-bitcoin`, I don't think it's worth the trouble to change it there.

luisschwab:
> Although this PR is scoped to the wallet, I was wondering if we should also remove the Debug outputs from Display implementations in keychain_txout.rs (line 798), tx_graph.rs (line 228), spk_client.rs (line 24), and file_store/src/lib.rs (line 25)?

Yes, but those should go on different PRs.

crates/wallet/src/wallet/error.rs
ValuedMammal: I agree we should be using the Display impl for `LockTime` and `Sequence`, etc.

Also found another instance of debug use on Display for `CreateWithPersistError`, although I never considered what the Display for `ChangeSet` would look like..

luisschwab: > Also found another instance of debug use on Display for CreateWithPersistError, although I never considered what the Display for ChangeSet would look like..

I'm working on it

ValuedMammal: I don't think we actually need to implement Display for `ChangeSet`, but we should think about what is the logical thing to return in the error message.

luisschwab: what do you think @ValuedMammal?
```rs
impl<E: fmt::Display> fmt::Display for CreateWithPersistError<E> {
        match self {
            Self::Persist(err) => write!(
                f,
                "Persistence error: {}",
                err
            ),
            Self::DataAlreadyExists(changeset) => write!(
                f,
                "Cannot create wallet in persister which already contains wallet data: {:?}",
                changeset
            ),
            Self::Descriptor(err) => write!(
                f,
                "Cannot construct wallet from loaded changeset: {}",
                err
            ),
        }
    }
}
```

crates/wallet/src/types.rs
```diff
@@ -37,6 +38,15 @@ impl KeychainKind {
     }
 }
 
+impl fmt::Display for KeychainKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            KeychainKind::External => write!(f, "external"),
+            KeychainKind::Internal => write!(f, "internal"),
```
ValuedMammal: Maybe more clear to use uppercase?
```suggestion
            KeychainKind::External => write!(f, "External"),
            KeychainKind::Internal => write!(f, "Internal"),
```

crates/wallet/src/wallet/error.rs
```diff
@@ -127,12 +127,16 @@ impl fmt::Display for CreateTxError {
                 requested,
                 required,
             } => {
-                write!(f, "TxBuilder requested timelock of `{:?}`, but at least `{:?}` is required to spend from this script", required, requested)
+                if requested.is_block_height() && required.is_block_height() {
+                    write!(f, "TxBuilder requested timelock of `{}` blocks, but at least `{}` blocks are required to spend from this script", required, requested)
+                } else {
+                    write!(f, "TxBuilder requested timelock of `{}` seconds, but at least `{}` seconds are required to spend from this script", required, requested)
+                }
```
ValuedMammal: I think it's fine to display LockTime unit-less, but if we want the units you could use the alternate format syntax `{:#}`.
```suggestion
                write!(f, "TxBuilder requested timelock of `{}`, but at least `{}` is required to spend from this script", required, requested)

```

crates/wallet/src/wallet/mod.rs
```diff
@@ -194,9 +194,9 @@ impl fmt::Display for LoadError {
             LoadError::MissingNetwork => write!(f, "loaded data is missing network type"),
             LoadError::MissingGenesis => write!(f, "loaded data is missing genesis hash"),
             LoadError::MissingDescriptor(k) => {
-                write!(f, "loaded data is missing descriptor for keychain {k:?}")
+                write!(f, "loaded data is missing descriptor for {k} keychain")
             }
-            LoadError::Mismatch(mismatch) => write!(f, "data mismatch: {mismatch:?}"),
+            LoadError::Mismatch(mismatch) => write!(f, "data mismatch: {mismatch}"),
```
ValuedMammal: nit: Now that `LoadMismatch` properly describes the error kinds, I would remove the extra verbiage.
```suggestion
            LoadError::Mismatch(e) => write!(f, "{e}"),
```

crates/wallet/src/wallet/persisted.rs
```diff
@@ -363,13 +363,15 @@ pub enum CreateWithPersistError<E> {
...
+            Self::Persist(err) => write!(f, "Persistence error: {}", err),
             Self::DataAlreadyExists(changeset) => write!(
                 f,
                 "Cannot create wallet in persister which already contains wallet data: {:?}",
                 changeset
             ),
-            Self::Descriptor(err) => fmt::Display::fmt(&err, f),
+            Self::Descriptor(err) => {
+                write!(f, "Cannot construct wallet from loaded changeset: {}", err)
+            }
```
ValuedMammal: It should be enough to just forward to the Display implementation of the inner error, so I don't think this adds much compared to what was there before.

crates/wallet/src/wallet/error.rs
ValuedMammal: It will be nice if the `DataAlreadyExists` error gave us a hint about the contents of the existing database without revealing every detail. We could have a helper function that takes a `&ChangeSet` and returns a formatted string with some basic metadata such as the network and descriptor checksum. As an example the end result would look like this

```
Cannot create wallet in persister which already contains wallet data: network testnet, descriptor checksum: celxt6vnfd68k0ut
```

luisschwab:
@ValuedMammal I think this is RTM

luisschwab:
Closing this and moving to https://github.com/bitcoindevkit/bdk_wallet/pull/12

PR #1840, opened 2025-02-16
example_cli: add feerate option to psbt new cmd

ValuedMammal:
The PR aims to add a `--feerate` option to the psbt `New` command.

Also in this PR:

- bump `bdk_coin_select` to 0.4
- fix collecting `Assets` and enable support for more descriptor types

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #1894, opened 2025-03-16
example_cli: use `bdk_tx` to create PSBTs

ValuedMammal:
This PR is intended to be used for testing transaction creation using "bdk-tx" which is still a new and mostly untested project. I may also use this as a place to collect any issues or findings that come up during testing.

builds on #1840 

#### All Submissions:

### DIFF
--- a/examples/example_cli/Cargo.toml
+++ b/examples/example_cli/Cargo.toml
@@ -9,6 +9,7 @@ edition = "2021"
 bdk_chain = { path = "../../crates/chain", features = ["serde", "miniscript"]}
 bdk_coin_select = "0.4"
 bdk_file_store = { path = "../../crates/file_store" }
+bdk_tx = { version = "0.1.0" }
 bitcoin = { version = "0.32.0", features = ["base64"], default-features = false }
 
 anyhow = "1"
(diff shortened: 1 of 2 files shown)


example-crates/example_esplora/src/main.rs
```diff
@@ -87,7 +87,7 @@ impl EsploraArgs {
 #[derive(Parser, Debug, Clone, PartialEq)]
 pub struct ScanOptions {
     /// Max number of concurrent esplora server requests.
-    #[clap(long, default_value = "5")]
+    #[clap(long, default_value = "2")]
```
aagbotemi: I think the default value of concurrent request is good at 5. While lower value might reduce the risk of overloading the system, using 5 allows more operation to run in parallel, and make a process to complete faster, which will help to improve overall performace.

aagbotemi (commented):
Hi @ValuedMammal, thanks for your hard work on this PR

example-crates/example_esplora/src/main.rs
ValuedMammal: Thanks @aagbotemi Since this was part of #1840 I left it as is. The option is configurable from the command line and if the default turns out to be too slow, we can change it back.

ValuedMammal:
Closed in favor of #2006.

PR #1902, opened 2025-03-20
fix(wallet): fix `into_descriptor_key` for DescriptorPublicKey

ValuedMammal:
Fix `into_descriptor_key` for `DescriptorPublicKey` by including the missing match arm for `DescriptorPublicKey::MultiXPub` when initializing `networks`.

Test that `into_wallet_descriptor` correctly parses a multipath descriptor and fails if passed a network that is invalid for the descriptor.

Note that rust-miniscript doesn't support parsing a multipath descriptor containing extended private keys.

fixes bitcoindevkit/bdk_wallet#10

### Changelog notice

Fixed an issue preventing `into_wallet_descriptor` from correctly mapping a DescriptorPublicKey to the set of available networks.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

#### Bugfixes:

* [x] I've added tests to reproduce the issue which are now passing
* [x] I'm linking the issue being fixed by this PR

PR #1906, opened 2025-03-21
feat(wallet): `create_single` accepts a multipath descriptor

### LINKED ISSUE #11: Support multipath descriptors
**Describe the enhancement**  

The release of [`rust-miniscript` 10.0.0](https://github.com/rust-bitcoin/rust-miniscript/blob/master/CHANGELOG.md#1000---may-24-2023) has [multipath descriptor](https://github.com/rust-bitcoin/rust-miniscript/pull/470) support. We should support this also in bdk 1.0 and backported to a bdk 0.2X (maintenance) release.

**Use case**  

See bitcoin/bitcoin#17190

**Additional context**

This was requested by @andreasgriffin for his project [bitcoin-safe](https://github.com/andreasgriffin/bitcoin-safe).

[BIP-389](https://github.com/bitcoin/bips/blob/master/bip-0389.mediawiki)

### LINKED ISSUE #1021: Support multipath descriptors
**Describe the enhancement**  

The release of [`rust-miniscript` 10.0.0](https://github.com/rust-bitcoin/rust-miniscript/blob/master/CHANGELOG.md#1000---may-24-2023) has [multipath descriptor](https://github.com/rust-bitcoin/rust-miniscript/pull/470) support. We should support this also in bdk 1.0 and backported to a bdk 0.2X (maintenance) release.

**Use case**  

See bitcoin/bitcoin#17190

**Additional context**

This was requested by @andreasgriffin for his project [bitcoin-safe](https://github.com/andreasgriffin/bitcoin-safe).

ValuedMammal:
### Notes to the reviewers

We reuse `Wallet::create_single` to accommodate a wallet descriptor that parses as a MultiXPub. The documentation is updated to clarify the distinction between the single vs. dual-keychain use cases.

The logical change happens in `create_with_params` where we now look for whether the external descriptor `is_multipath` and if so, use the parsed single descriptors to update the `descriptor` and `change_descriptor` and everything proceeds as normal. This means that the wallet change set still consists of the individually split descriptors.

An alternative that was discussed was to make a new dedicated function such as `create_from_multipath` that enforces the provided input to be a multipath descriptor. However we arrived at the approach in this PR in order to promote `create_single` as a more central API long term.

This is based on bitcoindevkit/bdk#1902 as that seemed to be relatively non controversial.

fixes bitcoindevkit/bdk_wallet#11 

Follow ups: Consider whether it might be needed to parse a multipath descriptor at load time https://github.com/bitcoindevkit/bdk_wallet/issues/11

### Changelog notice

Changed:

`Wallet::create_single` can be used to create a Wallet from a BIP389 multipath descriptor

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/wallet/src/wallet/mod.rs
+++ b/crates/wallet/src/wallet/mod.rs
@@ -44,7 +44,7 @@ use bitcoin::{
     Sequence, Transaction, TxOut, Txid, Weight, Witness,
 };
 use miniscript::{
-    descriptor::KeyMap,
+    descriptor::{Descriptor, KeyMap},
     psbt::{PsbtExt, PsbtInputExt, PsbtInputSatisfier},
 };
 use rand_core::RngCore;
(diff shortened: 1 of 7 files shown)


crates/wallet/src/wallet/mod.rs
```diff
@@ -356,25 +395,64 @@ impl Wallet {
...
+        let (mut descriptor, mut descriptor_keymap) = (params.descriptor)(&secp, network)?;
+
+        // In case of multipath we parse the descriptor into single descriptors and
+        // assign each one back to the `descriptor` and `change_descriptor`.
+        if descriptor.is_multipath() {
+            let mut desc_iter = descriptor.into_single_descriptors()?.into_iter();
+            descriptor = desc_iter.next().expect("must have a descriptor");
+            descriptor_keymap = Descriptor::parse_descriptor(&secp, &descriptor.to_string())?.1;
+            params.change_descriptor = desc_iter.next().map(params::make_descriptor_to_extract);
+        }
```
luisschwab, line 454: I think it should throw an error if there are more than 2 descriptors.

crates/wallet/src/wallet/mod.rs
```diff
@@ -356,25 +395,64 @@ impl Wallet {
...
         let (chain, chain_changeset) = LocalChain::from_genesis_hash(genesis_hash);
 
-        let (descriptor, mut descriptor_keymap) = (params.descriptor)(&secp, network)?;
+        let (mut descriptor, mut descriptor_keymap) = (params.descriptor)(&secp, network)?;
+
+        // In case of multipath we parse the descriptor into single descriptors and
+        // assign each one back to the `descriptor` and `change_descriptor`.
+        if descriptor.is_multipath() {
+            let mut desc_iter = descriptor.into_single_descriptors()?.into_iter();
+            descriptor = desc_iter.next().expect("must have a descriptor");
```
aagbotemi, line 451: Hi @ValuedMammal, your implementation is solid.
You can return this error gracefully, rather than panic here.

ValuedMammal: What error to return though?

aagbotemi: Probably another variant like `MissingDescriptor` can be added to the error enum

ValuedMammal: I would say the presence of a `descriptor` is invariant in the structure of CreateParams, meaning we've ruled out the possibility of a missing descriptor.

aagbotemi: Okay. I understand.

aagbotemi: Okay. I understand. That's a good development. I was initially concerned about the panic.

crates/wallet/src/wallet/mod.rs
```diff
@@ -332,6 +367,10 @@ impl Wallet {
     ///
     /// If you have previously created a wallet, use [`load`](Self::load) instead.
     ///
+    /// This should be used when the parsed `descriptor` and `change_desctiptor` are either
```
oleonardolima:
``` suggestion
    /// This should be used when the parsed `descriptor` and `change_descriptor` are either
```

crates/wallet/src/wallet/mod.rs
oleonardolima: Although you've explained that only the first 2 are used in the documentation, I also wonder if returning an error would be a better idea in order to prevent the user's unexpected behavior.

crates/wallet/src/wallet/mod.rs
```diff
@@ -356,25 +395,64 @@ impl Wallet {
...
-        let (descriptor, mut descriptor_keymap) = (params.descriptor)(&secp, network)?;
+        let (mut descriptor, mut descriptor_keymap) = (params.descriptor)(&secp, network)?;
+
+        // In case of multipath we parse the descriptor into single descriptors and
+        // assign each one back to the `descriptor` and `change_descriptor`.
+        if descriptor.is_multipath() {
+            let mut desc_iter = descriptor.into_single_descriptors()?.into_iter();
+            descriptor = desc_iter.next().expect("must have a descriptor");
+            descriptor_keymap = Descriptor::parse_descriptor(&secp, &descriptor.to_string())?.1;
+            params.change_descriptor = desc_iter.next().map(params::make_descriptor_to_extract);
```
oleonardolima, line 453: question: Is there any reason not to override the `params.descriptor` with the new single descriptor, too?

oleonardolima (commented):
cACK 4cf5537b3d87bd8ccb1b20d2151053c50a1191cb

Great work! I left a suggestion for a fix documentation typo, and another minor question. I agree with Schwab's, and wonder if we shouldn't return an error instead when multipath's with 3+ descriptors are used.

crates/wallet/src/wallet/mod.rs
ValuedMammal: Because we already parsed it.

crates/wallet/src/wallet/mod.rs
ValuedMammal: Can you make a code suggestion (preferably non breaking)?

ValuedMammal: I can't say I fully agree with throwing an error, and instead lean toward managing expectations through effective documentation. But I guess it can go either way.

thunderbiscuit:
Pardon me jumping here a bit late (and maybe with some missing information). 

One comment I have here is that once this feature is in, devs cannot expect that a wallet created with `Wallet::create_single()` will not have an internal keychain. Note also that as mentioned in the docs:

_for a wallet with no internal (change) keychain, any methods requiring a [`KeychainKind`] as input, e.g. [`reveal_next_address`] should only be called with the [`External`] variant. In some cases passing [`Internal`] is treated the same as [`External`] but this behavior must not be relied on._

I'm wondering where that certainty (around the number of keychains and hence reliance on certain APIs) should come from (and maybe relying on `Wallet::create_single` was not the right place to do that anyway?). A wallet could parse the descriptor before using it in its construction phase (Descriptor::is_multipath or similar), or call the `Wallet::keychains()` method and then pull the length of that iterator (seems hacky, or at least not what the method was intended for) to know if it has an internal keychain.

I initially thought of those as constructors that offer certainties about what types of wallets (i.e. single vs double keychains wallets) will come out of them (`Wallet::create` needs 2 descriptors and if you don't it errors out, `Wallet::create_single` takes only one descriptor and errors out if you give it more, and `Wallet::create_multi` takes one descriptors but must expand into 2 or more). Was this maybe the wrong mental model for it? Should this logic/checks live somewhere else maybe and I was sort of relying on the API in the wrong way?

Happy to hear everyone's thoughts. Again I'm coming in last minute a little bit on this, and I feel like I don't have complete knowledge on it so feel free to tell me I was wrong all along haha.

Issue #1891, opened 2025-03-15
Electrum client Performance issues


andreasgriffin:
The electrum client seems to have significantly suffered in performance in the upgrade to bdk 1.0

The full sync takes
- **4s** for bdk-ffi 0.32  
- **26s**  for bdk-ffi 1.1.0  (batch_size=100,  changing batch_size just increases the sync time) 

My setup:  Using a fulcrum server (very performant) and a wallet  with  a few thousand addresses.

evanlinjin:
Thank you for this. Could you please be a little more specific so it's easier to identify the performance issues? I.e. are you doing full scan or sync? What parameters are you using? How many txs per address and txs in total?

However, just off the top of my head, I think it is very likely it's because we are fetching and checking the merkle proofs per tx in the new `bdk_electrum` (I'm assuming the fetching part is more of an issue).

Not fetching the merkle proofs is a no go, as the electrum API only returns the height that a tx was included in (not the block hash). The old BDK was doing things incorrectly.

There are definitely things we can do to help the performance here. The most obvious thing would be to use the Electrum API properly. The Electrum API was intended as a pub-sub API. Work on this is done in an experimental client here: https://github.com/evanlinjin/experiments

We can also fetch merkle proofs asynchronously. This would be much easier if could use `async` rust code for the networking side (as I am doing in the aforementioned repo).

We can also cache/store/persist the merkle proofs if the tx can already be linked to the current best chain. This requires us to persist/cache headers.

---

For the short term, based on the way I understand the problem, I think the quickest way to speed things up is to fetch merkle proofs and headers in a batch and then do processing later on. Instead of fetch -> process -> fetch -> process as we are doing now. This will reduce round trips.

andreasgriffin:
thanks for the fast answer.

the parameters are:
- full_scan  stop_gap=500, batch_size=100, fetch_prev_txouts=True

evanlinjin:
Having a stop gap of 500 means you'll be scanning 500 extra addresses. Does reducing this improve the situation?

andreasgriffin:
The stop gap is user defined in Bitcoin Safe, and therefore is not up to me to improve performance.  
For this particular case the stop gap cannot be lowered without missing txs.

jp1ac4:
[...]

PR #1908, opened 2025-03-22
Electrum: optimize merkle proof validation with batching

Keerthi421:
This PR optimizes the Electrum client's performance by improving Merkle proof validation, addressing the significant performance regression in BDK 1.1.0 where full sync time increased from 4s to 26s.

Key improvements:
- Implemented batch processing for Merkle proof validations
- Added Merkle proof caching to prevent redundant network calls
- Optimized header handling with pre-fetching and reuse
- Modified core functions to use batch operations instead of individual calls

### Notes to the reviewers

The optimization approach focuses on three main areas:
1. Reducing network round trips through batched Merkle proof requests
2. Minimizing redundant operations with a new Merkle proof cache
3. Improving header handling efficiency with pre-fetching

The batch size is set to 100 as a balance between performance and memory usage. This value can be adjusted based on testing results.

### Added
- New Merkle proof cache to prevent redundant network calls
- Batch processing for Merkle proof validations
- Performance tests to verify sync time improvements
Solves issue #1891

### DIFF
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -6,7 +6,7 @@ use bdk_core::{
     },
     BlockId, CheckPoint, ConfirmationBlockTime, TxUpdate,
 };
-use electrum_client::{ElectrumApi, Error, HeaderNotification, GetMerkleRes};
+use electrum_client::{ElectrumApi, Error, GetMerkleRes, HeaderNotification};
 use std::sync::{Arc, Mutex};
 
 /// We include a chain suffix of a certain length for the purpose of robustness.
@@ -491,10 +491,11 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
         proofs: Vec<(Txid, GetMerkleRes)>,
     ) -> Result<(), Error> {
         // Pre-fetch all required headers
-        let heights: HashSet<u32> = proofs.iter()
+        let heights: HashSet<u32> = proofs
+            .iter()
             .map(|(_, proof)| proof.block_height as u32)
             .collect();
-        
+
         let mut headers = HashMap::new();
         for height in heights {
             headers.insert(height, self.fetch_header(height)?);
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -263,6 +263,10 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
         let mut txs_to_validate = Vec::new();
 
         loop {
+            if unused_spk_count >= stop_gap {
+                break;
+            }
+
             let spks = (0..batch_size)
                 .map_while(|_| spks_with_expected_txids.next())
                 .collect::<Vec<_>>();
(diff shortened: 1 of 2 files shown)


ValuedMammal:
Concept ACK

In 3ef0c83fb6a777040c100bc4b367087e439b816e:

I don't think `rustup-init.exe` should be included.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -22,6 +25,8 @@ pub struct BdkElectrumClient<E> {
     tx_cache: Mutex<HashMap<Txid, Arc<Transaction>>>,
     /// The header cache
     block_header_cache: Mutex<HashMap<u32, Header>>,
+    /// The Merkle proof cache
+    merkle_cache: Mutex<HashMap<(Txid, u32), GetMerkleRes>>,
```
evanlinjin: Let's say a tx is confirmed in block of height 100 and hash A. We then cache it's merkle proof. Then a reorg happens and the tx is now in block of hash B (but still height 100).

With the current implementation, we will never fetch the merkle proof for block B because we already have a merkle proof cached for a txid at the same height as block B.

Is there a way to cache merkle proofs against `(txid, blockhash)`?

I apologize in advance for how bad the Electrum API is designed.

Keerthi421: @evanlinjin  I will work on it

Keerthi421: @evanlinjin , @notmandatory  Please Review the changes made

Keerthi421:
@notmandatory please review the changes

crates/electrum/src/bdk_electrum_client.rs
evanlinjin: @Keerthi421 thank you for your attempt but you have not addressed the problem.

See if a transaction with `txid` get's confirmed in a block of height:hash `100:A` and then we do a sync. We will have an entry: `(txid, 100) : merkle_res_for_txid_in_A`.

If there is a reorg, and the block at height 100 is now `A'`, the entry will stay as it is and we will never fetch an anchor for "`txid` in `A'`". This will mean that this transaction will appear to never confirm on the wallet (until we restart and clear the cache).

notmandatory:
> @notmandatory please review the changes

@LagginTimes is a better person to review and he said he's taking a look at this PR.

LagginTimes:
Here are my fixes and suggestions based on your work so far https://github.com/LagginTimes/bdk/commit/412d800f03e5f3461aaa45e2395ee099d03e4271:
#### 1. Cache keyed by `(Txid, BlockHash)`
* Switched from `(Txid, height)` back to `(Txid, BlockHash)`.
* Since a transaction's block hash changes during a reorg, stale‐block proofs automatically miss the cache and are re-fetched exactly once for the new block.
#### 2. Reorg-safe eviction via `clear_stale_proofs()`
* Invoked before each batch lookup: we compare the highest cached block’s hash to the live tip.
* If they don’t match, we walk backwards, evicting proofs until we hit the fork point.

For additional efficiency, calling `clear_stale_proofs()` once at the top of `sync()` and `full_scan()` instead of before every `batch_fetch_merkle_proofs()` would ensure the check runs only once per operation. Please let me know what you think of these changes.

evanlinjin:
@Keerthi421 thank you for this work. @LagginTimes will be taking over since it'll be good to have this soon. @LagginTimes please make a new PR.

PR #1917, opened 2025-03-28
Canonicalization should handle transactions that spend from two conflicting transactions

### LINKED ISSUE #1898: panic (assertion) in try_filter_chain_txouts
**Describe the bug**  
I'm hitting this assertion in the `tx_graph.rs` `try_filter_chain_txouts` method:

```
                    assert!(
                        _res.is_none(),
                        "tried to replace {:?} with {:?}",
                        _res,
                        txid
                    );
```

Is this because one of the txs I made is being double spent?
We are not doing the most innocent bitcoin txs, but we're not doing any weird fiddling with bdk either. Just make a tx and then it gets double spent away on chain. And then the sync procedure panics with this message.

I'm not sure I understand what's going on here or in that method during sync. But it sounds like bdk will crash your program if sync is called while one of your transcations is double spent on-chain or in mempool?

**To Reproduce**  
Not sure it's easy to reproduce. I created a tx that didn't pass mempool checks, and then another tx onchain double spent that tx.

**Expected behavior**  
Sync shouldn't panic if no broken bdk functions were called?

**Build environment**  
 - BDK tag/commit: branch based on 0a20724e30316cbb68f7edf3f62028de15d26277, but assert also present in master
 - OS+version: nixos  
 - Rust/Cargo version: 1.84.0 
 - Rust/Cargo target: linux  

**Additional context**  
n/a

evanlinjin:
Fixes #1898 

### Description

When we initially created the canonicalization algorithm, we didn't expect callers to insert invalid transactions into the graph. However, user error happens and we should handle it correctly.

Before this PR, when inserting transactions that double-spend themselves (where two or more different inputs of the transactions conflict), the canonicalization result will have inconsistencies.

### Notes to the reviewers

Logic changes are all contained in `CanonicalIter::mark_canonical`. `mark_canonical` will detect whether the `tx` being passed in double spends itself. If so, we abort and undo all changes made so far.

There is a slight <2% degradation in performance with this change (except in two cases where there is a performance improvement of ~10%).

 
[bench.txt](https://github.com/user-attachments/files/19788730/bench.txt)

### Changelog notice

* Fix canonicalization mess-up when transactions that conflict with itself are inserted.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/canonical_iter.rs
+++ b/crates/chain/src/canonical_iter.rs
@@ -1,11 +1,15 @@
-use crate::collections::{hash_map, HashMap, HashSet, VecDeque};
+use crate::collections::{HashMap, HashSet, VecDeque};
 use crate::tx_graph::{TxAncestors, TxDescendants};
 use crate::{Anchor, ChainOracle, TxGraph};
 use alloc::boxed::Box;
 use alloc::collections::BTreeSet;
 use alloc::sync::Arc;
+use alloc::vec::Vec;
 use bdk_core::BlockId;
 use bitcoin::{Transaction, Txid};
+
+type CanonicalMap<A> = HashMap<Txid, (Arc<Transaction>, CanonicalReason<A>)>;
+type NotCanonicalSet = HashSet<Txid>;
 
 /// Iterates over canonical txs.
 pub struct CanonicalIter<'g, A, C> {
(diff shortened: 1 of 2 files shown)


ValuedMammal (approved):
ACK 370497c3a9c361e10c8341cbb07210f026cc3e04

LagginTimes (approved):
ACK 370497c3a9c361e10c8341cbb07210f026cc3e04

PR #1946, opened 2025-04-22
chore: create and apply rustfmt.toml

luisschwab:
This PR adds basic formatting configuration for uniform comments. Note that the nightly toolchain is necessary.

### Changelog:

- Create rustfmt.toml with this config:
```
comment_width = 100
format_code_in_doc_comments = true
wrap_comments = true
```
- Apply rustfmt.toml.
- Update the fmt CI job to use the nightly toolchain.
- Update PR template to use nightly toolchain.

Note: I had to add #[rustfmt::skip] to test_extract_satisfaction_timelock() because the base64 PSBT would get wrapped due to the slashes.

#### All Submissions:

* [X] I ran `cargo +nightly fmt` and `cargo clippy` before committing

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,3 @@
+comment_width = 100
+format_code_in_doc_comments = true
+wrap_comments = true
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -20,7 +20,7 @@ of the PR were done in a specific way -->
 
 * [ ] I've signed all my commits
 * [ ] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
-* [ ] I ran `cargo fmt` and `cargo clippy` before committing
+* [ ] I ran `cargo +nightly fmt` and `cargo clippy` before committing
 
 #### New Features:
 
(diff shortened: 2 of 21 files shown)


GideonBature:
tACK #1946 

Can you please squash all three commit messages as a single commit?

rustfmt.toml
```diff
@@ -0,0 +1,3 @@
+comment_width = 140
```
oleonardolima: Any reason for 140? I think a maximum length of 100 would be better to keep in line with `max_width`.

oleonardolima (commented):
cACK 396fff8f3fd07d78ed30dc0ed4ffeb6363288b81

rustfmt.toml
luisschwab: I think it's easier to read long comments versus long code.

evanlinjin:
Is it some sort of standard to have `comment_width = 140`?

Currently, most of the documentation is kept within 100.

luisschwab:
> Is it some sort of standard to have comment_width = 140?

Not really. Either one is fine for me.

evanlinjin:
In that case, I would prefer to stick with 100. It means you can see more if you use a split screen.

luisschwab:
I made an identical PR on bdk_wallet, once this get's merged I'll update it.

notmandatory:
Try rebasing this PR and it should build fine after that. See changes in #1955

notmandatory (approved):
ACK 0f56c3780166f444d02ebc1ed2e7fa8168318a42

luisschwab:
Rebased to trigger CI.

notmandatory (approved):
ACK 7bdbcdd2a1b6bb0fb9fafd7792615a1a3a302a5a

Issue #1947, opened 2025-04-24
Track `first_seen`


evanlinjin:
**Describe the enhancement**  

Add a `first_seen` field to `tx_graph::ChangeSet` and a way to retrieve the first-seen timestamp for a transaction.

**Use case** 

Currently, `tx_graph::ChangeSet` only tracks the `last_seen` timestamp. `last_seen` is useful for canonicalization. However, wallets would like to order unconfirmed transactions by `first_seen` so that the wallet user sees a chronological list of transactions in their transaction list.

uvuvuwu:
Can I take this on?

LagginTimes:
Resolved by https://github.com/bitcoindevkit/bdk/pull/1950 and https://github.com/bitcoindevkit/bdk/pull/1966

PR #1950, opened 2025-04-30
feat!(chain): implement `first_seen` tracking

uvuvuwu:
### Description

This PR solves issue #1947 by implementing `first_seen` tracking.

* Added `first_seen` field to `TxGraph` and `ChangeSet` so that `first_seen` timestamp can be added when inserting a new seen-at using `insert_seen_at`. 

* `first_seen` added to `TxNode` as a way to retrieve the first-seen timestamp for a transaction.

* `first_seen` added to `ChainPosition::Unconfirmed` to order unconfirmed transactions by `first_seen`.

* New tests have been added for the above described functionalities. 

### Changelog notice

* Add tracking first-seen timestamps of transactions 

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/chain_data.rs
+++ b/crates/chain/src/chain_data.rs
@@ -27,6 +27,10 @@ pub enum ChainPosition<A> {
     },
     /// The chain data is not confirmed.
     Unconfirmed {
+        /// When the chain data was first seen in the mempool.
+        ///
+        /// This value will be `None` if the chain data was never seen in the mempool.
+        first_seen: Option<u64>,
         /// When the chain data is last seen in the mempool.
         ///
         /// This value will be `None` if the chain data was never seen in the mempool and only seen
@@ -58,7 +62,13 @@ impl<A: Clone> ChainPosition<&A> {
                 anchor: anchor.clone(),
                 transitively,
             },
-            ChainPosition::Unconfirmed { last_seen } => ChainPosition::Unconfirmed { last_seen },
+            ChainPosition::Unconfirmed {
+                last_seen,
+                first_seen,
+            } => ChainPosition::Unconfirmed {
+                last_seen,
+                first_seen,
+            },
         }
     }
 }
@@ -165,9 +175,11 @@ mod test {
     fn chain_position_ord() {
         let unconf1 = ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
             last_seen: Some(10),
+            first_seen: Some(10),
         };
         let unconf2 = ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
             last_seen: Some(20),
+            first_seen: Some(20),
         };
         let conf1 = ChainPosition::Confirmed {
             anchor: ConfirmationBlockTime {
@@ -195,6 +207,52 @@ mod test {
         assert!(
             conf2 > conf1,
             "confirmation_height is higher then it should be higher ord"
+        );
+    }
+
+    #[test]
+    fn test_sort_unconfirmed_chain_position() {
+        let mut v = vec![
+            ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
+                first_seen: Some(5),
+                last_seen: Some(20),
+            },
+            ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
+                first_seen: Some(15),
+                last_seen: Some(30),
+            },
+            ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
+                first_seen: Some(1),
+                last_seen: Some(10),
+            },
+            ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
+                first_seen: Some(3),
+                last_seen: Some(6),
+            },
+        ];
+
+        v.sort();
+
+        assert_eq!(
+            v,
+            vec![
+                ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
+                    first_seen: Some(1),
+                    last_seen: Some(10)
+                },
+                ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
+                    first_seen: Some(3),
+                    last_seen: Some(6)
+                },
+                ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
+                    first_seen: Some(5),
+                    last_seen: Some(20)
+                },
+                ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
+                    first_seen: Some(15),
+                    last_seen: Some(30)
+                },
+            ]
         );
     }
 }
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -177,6 +177,7 @@ pub struct TxGraph<A = ConfirmationBlockTime> {
     txs: HashMap<Txid, TxNodeInternal>,
     spends: BTreeMap<OutPoint, HashSet<Txid>>,
     anchors: HashMap<Txid, BTreeSet<A>>,
+    first_seen: HashMap<Txid, u64>,
     last_seen: HashMap<Txid, u64>,
     last_evicted: HashMap<Txid, u64>,
 
@@ -195,6 +196,7 @@ impl<A> Default for TxGraph<A> {
             txs: Default::default(),
             spends: Default::default(),
             anchors: Default::default(),
+            first_seen: Default::default(),
             last_seen: Default::default(),
             last_evicted: Default::default(),
             txs_by_highest_conf_heights: Default::default(),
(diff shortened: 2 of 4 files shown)


crates/chain/src/chain_data.rs
```diff
@@ -32,6 +32,10 @@ pub enum ChainPosition<A> {
         /// This value will be `None` if the chain data was never seen in the mempool and only seen
         /// in a conflicting chain.
         last_seen: Option<u64>,
+        /// When the chain data was first seen in the mempool
+        ///
+        /// This value will be `None` if the chain data was never seen in the mempool
+        first_seen: Option<u64>,
```
evanlinjin: I think `first_seen` should go before `last_seen` due to how [`derive(Ord)`](https://doc.rust-lang.org/std/cmp/trait.Ord.html#derivable) works. Unconfirmed txs should be ordered by `first_seen` first.

crates/chain/src/chain_data.rs
```diff
@@ -192,4 +205,88 @@ mod test {
...
+    #[test]
+    fn test_sort_unconfirmed_chain_position() {
+        let mut v = vec![
+            ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
+                first_seen: Some(5),
+                last_seen: Some(20),
+            },
+            ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
+                first_seen: None,
+                last_seen: Some(30),
```
evanlinjin: It's impossible to have a first-seen and no last-seen. The same goes vise-versa.

crates/chain/src/chain_data.rs
```diff
@@ -192,4 +205,88 @@ mod test {
...
+        ];
+
+        // Sort by first_seen: lowest first_seen first, None at the bottom
+        v.sort_by(|a, b| match (a, b) {
+            (
+                ChainPosition::<ConfirmationBlockTime>::Unconfirmed { first_seen: a, .. },
+                ChainPosition::<ConfirmationBlockTime>::Unconfirmed { first_seen: b, .. },
+            ) => a.unwrap_or(u64::MAX).cmp(&b.unwrap_or(u64::MAX)),
+            _ => Ordering::Equal,
+        });
```
evanlinjin:
```suggestion
        v.sort();
```

This is enough. Not sure what the goal is of trying to Ord by different subsets of `ChainPosition`.

crates/chain/src/tx_graph.rs
```diff
@@ -719,8 +725,40 @@ impl<A: Anchor> TxGraph<A> {
...
+
+        let mut changeset = ChangeSet::<A>::default();
+        if is_changed {
+            changeset.first_seen.insert(txid, seen_at);
+        }
+        changeset
+    }
+
+    /// Updates `last_seen` given a new `seen_at`
+    pub fn update_last_seen(&mut self, txid: Txid, seen_at: u64) -> ChangeSet<A> {
```
evanlinjin: Please make this private. We always want to update both `last-seen` and `first-seen` whenever a timestamp is introduced.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -1325,3 +1330,89 @@ fn tx_graph_update_conversion() {
...
+        ChangeSet {
+            txs: [].into(),
+            txouts: [].into(),
+            anchors: [].into(),
+            first_seen: [(hash!("tx1"), 1000000)].into(),
+            last_seen: [(hash!("tx1"), 1000000)].into(),
+            last_evicted: [].into(),
+        }
+    );
+    graph.apply_changeset(changeset);
```
evanlinjin: I don't think this is needed. `graph.insert_seen_at` should already mutate the internal state, so this line essentially would do nothing if everything is working correctly.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -1325,3 +1330,89 @@ fn tx_graph_update_conversion() {
...
+    assert_eq!(
+        changeset,
+        ChangeSet {
+            txs: [].into(),
+            txouts: [].into(),
+            anchors: [].into(),
+            first_seen: [(hash!("tx1"), 1000000)].into(),
+            last_seen: [(hash!("tx1"), 1000000)].into(),
+            last_evicted: [].into(),
+        }
```
evanlinjin:
```suggestion
        ChangeSet {
            first_seen: [(hash!("tx1"), 1000000)].into(),
            last_seen: [(hash!("tx1"), 1000000)].into(),
            ..Default::default()
        }
```

The [`Default`](https://doc.rust-lang.org/std/default/trait.Default.html) trait is helpful to know about.

evanlinjin (changes requested):
Mostly looks good. Some quirks that requires changing in the tests. Also, `ChainPosition` field order should be changed to make proper use of `derive(Ord)`.

Other things to take note of:
* Documentation comments need full stops!
* Will be nice to include the module changed in commit messages. I.e. `feat(chain):` signals that we are changing the `bdk_chain` crate.

crates/chain/src/tx_graph.rs
```diff
@@ -1250,6 +1285,8 @@ pub struct ChangeSet<A = ()> {
     pub txouts: BTreeMap<OutPoint, TxOut>,
     /// Added anchors.
     pub anchors: BTreeSet<(A, Txid)>,
+    /// Added first-seen unix timestamps of transactions.
+    pub first_seen: BTreeMap<Txid, u64>,
```
evanlinjin: For backward compatibility, please add this field at the end of the struct and annotate it with:

```rust
#[cfg_attr(feature = "serde", serde(default))]
```

crates/chain/src/chain_data.rs
uvuvuwu: I wasn't sure if there was anything that was using `last_seen` that was relying on `Ord` working on it, so I intentionally put `first_seen` under `last_seen`. I'll change that now

crates/chain/src/tx_graph.rs
evanlinjin: The order we introduce fields matter because the order matters for some serialization formats.

`serde(default)` is important so that `serde` does not error out when this field is missing, instead puts the `Default` impl in it's place.

crates/chain/src/chain_data.rs
uvuvuwu: Do we not need to be able to order by `first_seen` and `last_seen` for different use cases?

evanlinjin: @uvuvuwu The way the tests are implemented basically only tests the `Ord` impl of stdlib types are correct.

What we want to test is that the `ChainPosition` type has `Ord` implemented properly. `ChainPosition` is a user-facing struct, so we only want to order it chronologically (therefore by `first_seen`).

There may be other situations (such as during canonicalization or other niche use cases) where we MAY want to order by `last_seen`. This is out of the concern of `ChainPosition` and types can only have one `Ord` impl anyway.

crates/chain/src/tx_graph.rs
```diff
@@ -214,6 +216,8 @@ pub struct TxNode<'a, T, A> {
     pub tx: T,
     /// The blocks that the transaction is "anchored" in.
     pub anchors: &'a BTreeSet<A>,
+    /// The first-seen unix timestamp of the transaction.
+    pub first_seen: Option<u64>,
     /// The last-seen unix timestamp of the transaction as unconfirmed.
     pub last_seen_unconfirmed: Option<u64>,
```
evanlinjin: To make things consistent, I propose the following: 

* Rename `first_seen` to `first_seen_unconfirmed`.
* Update documentation to mention that it's the first-seen timestamp of the transaction as unconfirmed.

crates/chain/src/tx_graph.rs
```diff
@@ -789,8 +795,40 @@ impl<A: Anchor> TxGraph<A> {
 
     /// Inserts the given `seen_at` for `txid` into [`TxGraph`].
     ///
-    /// Note that [`TxGraph`] only keeps track of the latest `seen_at`.
+    /// Note that [`TxGraph`] keeps track of the latest `seen_at` and the earliest `seen_at`.
```
evanlinjin: Suggestion: 

```rust
/// Updates the first- and last-seen timestamps for a given `txid` in the [`TxGraph`].
///
/// This method records the time a transaction was observed by updating both:
/// - the **first-seen** timestamp, which only changes if `seen_at` is earlier than the current value, and
/// - the **last-seen** timestamp, which only changes if `seen_at` is later.
///
/// Returns a [`ChangeSet`] representing any changes applied.
```

evanlinjin (commented):
LGTM, just documentation suggestions.

evanlinjin (approved):
ACK 8b17bcfc03c39cd35415742bc31293b7969d1024

PR #1954, opened 2025-05-09
core: allow serialization of SyncResponse

icota:
### Description

The rationale behind this is to allow a separate instance of the same wallet be updated on a device that's not necessarily connected. Like a hardware wallet.

A connected device gets the response, serializes it with serde and passes the `Update` to the air-gapped one

#### All Submissions:

### DIFF
--- a/crates/core/src/checkpoint.rs
+++ b/crates/core/src/checkpoint.rs
@@ -10,10 +10,12 @@ use crate::BlockId;
 /// Checkpoints are cheaply cloneable and are useful to find the agreement point between two sparse
 /// block chains.
 #[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 pub struct CheckPoint(Arc<CPInner>);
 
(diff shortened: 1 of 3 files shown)


evanlinjin:
@icota have you considered using `ChangeSet`s instead? Serializing `CheckPoint`s is a bit odd since it is a linked list and the resultant serialization would be super nested.

Also updates may have repeated elements, whereas `ChangeSet`s only include what is changed.

icota:
> @icota have you considered using `ChangeSet`s instead?

No, I have not. Are you suggesting I use `staged` after `apply_update` to get the `ChangeSet`? How do I apply that on the other side?

evanlinjin:
@icota Yes. That is a good point about applying it on the other side. I forgot that `bdk_wallet::Wallet` does not have an `apply_changeset` method (whereas the core structures from `bdk_chain` do).

Just made a draft PR to add `Wallet::apply_changeset`: https://github.com/bitcoindevkit/bdk_wallet/pull/231

evanlinjin:
Replaced by #2232

PR #1957, opened 2025-05-15
feat(electrum): optimize merkle proof validation with batching

### LINKED ISSUE #1891: Electrum client Performance issues

LagginTimes:
Replaces #1908, originally authored by @Keerthi421.
Fixes #1891.

### Description

This PR optimizes `sync`/`full_scan` performance by batching and caching key RPC calls to slash network round-trips and eliminate redundant work.

Key improvements:

* Gather all `blockchain.transaction.get_merkle` calls into a single `batch_call` request.
* Use `batch_script_get_history` instead of many individual `script_get_history` calls.
* Use `batch_block_header` to fetch all needed block headers in one call rather than repeatedly calling `block_header`.
* Introduce a cache of transaction anchors to skip re-validating already confirmed transactions.

#### Anchor Caching Performance Improvements

Results suggest a significant speed up with a warmed up cache. Tested on local Electrum server with:
```
$ cargo bench -p bdk_electrum --bench test_sync
```

Results before this PR (https://github.com/LagginTimes/bdk/tree/1957-master-branch):

```
sync_with_electrum      time:   [1.3702 s 1.3732 s 1.3852 s]
```

Results after this PR:

```
sync_with_electrum      time:   [851.31 ms 853.26 ms 856.23 ms]
```
#### Batch Call Performance Improvements

No persisted data was carried over between runs, so each test started with cold caches and measured only raw batching performance. Tested with`example_electrum` out of https://github.com/LagginTimes/bdk/tree/example_electrum_timing with the following parameters:

```
$ example_electrum init "tr([62f3f3af/86'/1'/0']tpubDD4Kse29e47rSP5paSuNPhWnGMcdEDAuiG42LEd5yaRDN2CFApWiLTAzxQSLS7MpvxrpxvRJBVcjhVPRk7gec4iWfwvLrEhns1LA4h7i3c2/0/*)#cn4sudyq"
$ example_electrum scan tcp://signet-electrumx.wakiyamap.dev:50001
```

Results before this PR:

```
FULL_SCAN TIME: 8.145874476s
```

Results after this PR (using this PR's [`bdk_electrum_client.rs`](https://github.com/bitcoindevkit/bdk/blob/70495e2010541acbb5d62f9b5692de20924ac53f/crates/electrum/src/bdk_electrum_client.rs)):

```
FULL_SCAN TIME: 2.594050112s
```

### Changelog notice

* Add transaction anchor cache to prevent redundant network calls.
* Batch Merkle proof, script history, and header requests.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/electrum/benches/test_sync.rs
+++ b/crates/electrum/benches/test_sync.rs
@@ -1,0 +1,111 @@
+use bdk_chain::{
+    bitcoin::{Address, Amount, ScriptBuf},
+    local_chain::LocalChain,
+    spk_client::{SyncRequest, SyncResponse},
+    spk_txout::SpkTxOutIndex,
+    ConfirmationBlockTime, IndexedTxGraph, Indexer, Merge,
+};
+use bdk_core::bitcoin::{
+    key::{Secp256k1, UntweakedPublicKey},
+    Network,
+};
+use bdk_electrum::BdkElectrumClient;
+use bdk_testenv::{anyhow, bitcoincore_rpc::RpcApi, TestEnv};
+use criterion::{criterion_group, criterion_main, Criterion};
+use std::time::Duration;
+
+// Batch size for `sync_with_electrum`.
+const BATCH_SIZE: usize = 5;
+
+pub fn get_test_spk() -> ScriptBuf {
+    const PK_BYTES: &[u8] = &[
+        12, 244, 72, 4, 163, 4, 211, 81, 159, 82, 153, 123, 125, 74, 142, 40, 55, 237, 191, 231,
+        31, 114, 89, 165, 83, 141, 8, 203, 93, 240, 53, 101,
+    ];
+    let secp = Secp256k1::new();
+    let pk = UntweakedPublicKey::from_slice(PK_BYTES).expect("Must be valid PK");
+    ScriptBuf::new_p2tr(&secp, pk, None)
+}
+
+fn sync_with_electrum<I, Spks>(
+    client: &BdkElectrumClient<electrum_client::Client>,
+    spks: Spks,
+    chain: &mut LocalChain,
+    graph: &mut IndexedTxGraph<ConfirmationBlockTime, I>,
+) -> anyhow::Result<SyncResponse>
+where
+    I: Indexer,
+    I::ChangeSet: Default + Merge,
+    Spks: IntoIterator<Item = ScriptBuf>,
+    Spks::IntoIter: ExactSizeIterator + Send + 'static,
+{
+    let update = client.sync(
+        SyncRequest::builder().chain_tip(chain.tip()).spks(spks),
+        BATCH_SIZE,
+        true,
+    )?;
+
+    assert!(
+        !update.tx_update.txs.is_empty(),
+        "expected some transactions from sync, but got none"
+    );
+
+    if let Some(chain_update) = update.chain_update.clone() {
+        let _ = chain
+            .apply_update(chain_update)
+            .map_err(|err| anyhow::anyhow!("LocalChain update error: {:?}", err))?;
+    }
+    let _ = graph.apply_update(update.tx_update.clone());
+
+    Ok(update)
+}
+
+pub fn test_sync_performance(c: &mut Criterion) {
+    let env = TestEnv::new().unwrap();
+    let electrum_client = electrum_client::Client::new(env.electrsd.electrum_url.as_str()).unwrap();
+    let client = BdkElectrumClient::new(electrum_client);
+
+    const NUM_BLOCKS: usize = 100;
+    let mut spks = Vec::with_capacity(NUM_BLOCKS);
+
+    // Mine some blocks and send transactions.
+    env.mine_blocks(101, None).unwrap();
+
+    // Scatter UTXOs across many blocks.
+    for _ in 0..NUM_BLOCKS {
+        let spk = get_test_spk();
+        let addr = Address::from_script(&spk, Network::Regtest).unwrap();
+        env.send(&addr, Amount::from_sat(10_000)).unwrap();
+        env.mine_blocks(1, None).unwrap();
+
+        spks.push(spk);
+    }
+    let _ = env.wait_until_electrum_sees_block(Duration::from_secs(6));
+
+    // Setup receiver.
+    let genesis = env.bitcoind.client.get_block_hash(0).unwrap();
+    let (chain, _) = LocalChain::from_genesis_hash(genesis);
+    let graph = IndexedTxGraph::<ConfirmationBlockTime, _>::new({
+        let mut idx = SpkTxOutIndex::default();
+        idx.insert_spk((), spks[0].clone());
+        idx
+    });
+
+    c.bench_function("sync_with_electrum", move |b| {
+        b.iter(|| {
+            let spks = spks.clone();
+            let mut recv_chain = chain.clone();
+            let mut recv_graph = graph.clone();
+
+            let _ = sync_with_electrum(&client, spks, &mut recv_chain, &mut recv_graph);
+        })
+    });
+}
+
+criterion_group! {
+    name = benches;
+    config = Criterion::default()
+        .sample_size(10);
+    targets = test_sync_performance
+}
+criterion_main!(benches);
--- a/crates/electrum/benches/test_sync.rs
+++ b/crates/electrum/benches/test_sync.rs
@@ -1,46 +1,44 @@
-use bdk_chain::{
-    bitcoin::{Address, Amount, ScriptBuf},
-    local_chain::LocalChain,
-    spk_client::{SyncRequest, SyncResponse},
-    spk_txout::SpkTxOutIndex,
-    ConfirmationBlockTime, IndexedTxGraph, Indexer, Merge,
-};
-use bdk_core::bitcoin::{
-    key::{Secp256k1, UntweakedPublicKey},
-    Network,
+use bdk_chain::bitcoin::{Address, Amount, ScriptBuf};
+use bdk_core::{
+    bitcoin::{
+        consensus::WriteExt,
+        hashes::Hash,
+        key::{Secp256k1, UntweakedPublicKey},
+        Network, TapNodeHash,
+    },
+    spk_client::SyncRequest,
+    CheckPoint,
 };
 use bdk_electrum::BdkElectrumClient;
 use bdk_testenv::{anyhow, bitcoincore_rpc::RpcApi, TestEnv};
 use criterion::{criterion_group, criterion_main, Criterion};
-use std::time::Duration;
+use electrum_client::ElectrumApi;
+use std::{collections::BTreeSet, time::Duration};
 
 // Batch size for `sync_with_electrum`.
-const BATCH_SIZE: usize = 5;
+const BATCH_SIZE: usize = 100;
 
-pub fn get_test_spk() -> ScriptBuf {
+pub fn get_test_spk(i: usize) -> ScriptBuf {
     const PK_BYTES: &[u8] = &[
         12, 244, 72, 4, 163, 4, 211, 81, 159, 82, 153, 123, 125, 74, 142, 40, 55, 237, 191, 231,
         31, 114, 89, 165, 83, 141, 8, 203, 93, 240, 53, 101,
     ];
     let secp = Secp256k1::new();
     let pk = UntweakedPublicKey::from_slice(PK_BYTES).expect("Must be valid PK");
-    ScriptBuf::new_p2tr(&secp, pk, None)
+    let mut engine = TapNodeHash::engine();
+    engine.emit_u64(i as u64).expect("must emit");
+    ScriptBuf::new_p2tr(&secp, pk, Some(TapNodeHash::from_engine(engine)))
 }
 
-fn sync_with_electrum<I, Spks>(
-    client: &BdkElectrumClient<electrum_client::Client>,
-    spks: Spks,
-    chain: &mut LocalChain,
-    graph: &mut IndexedTxGraph<ConfirmationBlockTime, I>,
-) -> anyhow::Result<SyncResponse>
-where
-    I: Indexer,
-    I::ChangeSet: Default + Merge,
-    Spks: IntoIterator<Item = ScriptBuf>,
-    Spks::IntoIter: ExactSizeIterator + Send + 'static,
-{
+fn sync_with_electrum<E: ElectrumApi>(
+    client: &BdkElectrumClient<E>,
+    spks: &[ScriptBuf],
+    chain_tip: &CheckPoint,
+) -> anyhow::Result<()> {
     let update = client.sync(
-        SyncRequest::builder().chain_tip(chain.tip()).spks(spks),
+        SyncRequest::builder()
+            .chain_tip(chain_tip.clone())
+            .spks(spks.iter().cloned()),
         BATCH_SIZE,
         true,
     )?;
@@ -50,20 +48,11 @@ where
         "expected some transactions from sync, but got none"
     );
 
-    if let Some(chain_update) = update.chain_update.clone() {
-        let _ = chain
-            .apply_update(chain_update)
-            .map_err(|err| anyhow::anyhow!("LocalChain update error: {:?}", err))?;
-    }
-    let _ = graph.apply_update(update.tx_update.clone());
-
-    Ok(update)
+    Ok(())
 }
 
 pub fn test_sync_performance(c: &mut Criterion) {
     let env = TestEnv::new().unwrap();
-    let electrum_client = electrum_client::Client::new(env.electrsd.electrum_url.as_str()).unwrap();
-    let client = BdkElectrumClient::new(electrum_client);
 
     const NUM_BLOCKS: usize = 100;
     let mut spks = Vec::with_capacity(NUM_BLOCKS);
@@ -72,8 +61,8 @@ pub fn test_sync_performance(c: &mut Criterion) {
     env.mine_blocks(101, None).unwrap();
 
     // Scatter UTXOs across many blocks.
-    for _ in 0..NUM_BLOCKS {
-        let spk = get_test_spk();
+    for i in 0..NUM_BLOCKS {
+        let spk = get_test_spk(i);
         let addr = Address::from_script(&spk, Network::Regtest).unwrap();
         env.send(&addr, Amount::from_sat(10_000)).unwrap();
         env.mine_blocks(1, None).unwrap();
@@ -81,25 +70,43 @@ pub fn test_sync_performance(c: &mut Criterion) {
         spks.push(spk);
     }
     let _ = env.wait_until_electrum_sees_block(Duration::from_secs(6));
+    assert_eq!(
+        spks.iter().cloned().collect::<BTreeSet<_>>().len(),
+        spks.len(),
+        "all spks must be unique",
+    );
 
     // Setup receiver.
-    let genesis = env.bitcoind.client.get_block_hash(0).unwrap();
-    let (chain, _) = LocalChain::from_genesis_hash(genesis);
-    let graph = IndexedTxGraph::<ConfirmationBlockTime, _>::new({
-        let mut idx = SpkTxOutIndex::default();
-        idx.insert_spk((), spks[0].clone());
-        idx
+    let genesis_cp = CheckPoint::new(bdk_core::BlockId {
+        height: 0,
+        hash: env.bitcoind.client.get_block_hash(0).unwrap(),
     });
 
-    c.bench_function("sync_with_electrum", move |b| {
-        b.iter(|| {
-            let spks = spks.clone();
-            let mut recv_chain = chain.clone();
-            let mut recv_graph = graph.clone();
+    {
+        let electrum_client =
+            electrum_client::Client::new(env.electrsd.electrum_url.as_str()).unwrap();
+        let spks = spks.clone();
+        let genesis_cp = genesis_cp.clone();
+        c.bench_function("sync_with_electrum", move |b| {
+            b.iter(|| {
+                sync_with_electrum(
+                    &BdkElectrumClient::new(&electrum_client),
+                    &spks,
+                    &genesis_cp,
+                )
+                .expect("must not error")
+            })
+        });
+    }
 
-            let _ = sync_with_electrum(&client, spks, &mut recv_chain, &mut recv_graph);
-        })
-    });
+    {
+        let client = BdkElectrumClient::new(
+            electrum_client::Client::new(env.electrsd.electrum_url.as_str()).unwrap(),
+        );
+        c.bench_function("sync_with_electrum_cached", move |b| {
+            b.iter(|| sync_with_electrum(&client, &spks, &genesis_cp).expect("must not error"))
+        });
+    }
 }
 
 criterion_group! {
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -70,33 +70,6 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
         Ok(tx)
     }
 
-    /// Fetch block header of given `height`.
-    ///
-    /// If it hits the cache it will return the cached version and avoid making the request.
-    fn fetch_header(&self, height: u32) -> Result<Header, Error> {
-        let block_header_cache = self.block_header_cache.lock().unwrap();
-
-        if let Some(header) = block_header_cache.get(&height) {
-            return Ok(*header);
-        }
-
-        drop(block_header_cache);
-
-        self.update_header(height)
-    }
-
-    /// Update a block header at given `height`. Returns the updated header.
-    fn update_header(&self, height: u32) -> Result<Header, Error> {
-        let header = self.inner.block_header(height as usize)?;
-
-        self.block_header_cache
-            .lock()
-            .unwrap()
-            .insert(height, header);
-
-        Ok(header)
-    }
-
     /// Broadcasts a transaction to the network.
     ///
     /// This is a re-export of [`ElectrumApi::transaction_broadcast`].
@@ -292,7 +265,7 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
         let mut unused_spk_count = 0;
         let mut last_active_index = None;
 
-        loop {
+        'batch_loop: loop {
             let spks = (0..batch_size)
                 .map_while(|_| spks_with_expected_txids.next())
                 .collect::<Vec<_>>();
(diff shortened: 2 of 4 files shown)


crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -22,6 +25,8 @@ pub struct BdkElectrumClient<E> {
     tx_cache: Mutex<HashMap<Txid, Arc<Transaction>>>,
     /// The header cache
     block_header_cache: Mutex<HashMap<u32, Header>>,
+    /// The Merkle proof cache
+    merkle_cache: Mutex<HashMap<(Txid, BlockHash), GetMerkleRes>>,
```
evanlinjin: It will be more efficient if we cache anchors instead of `GetMerkleRes` here.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -414,55 +444,166 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+        // Evict any stale proofs until fork point is found.
+        for (height, old_hash) in entries {
+            let current_hash = self.fetch_header(height)?.block_hash();
+            if current_hash == old_hash {
+                break;
+            }
+            cache.retain(|&(_txid, bh), _| bh != old_hash);
+        }
+        Ok(())
+    }
```
evanlinjin: Reorgs don't happen that often so we won't have much "extra data". This method looks like it's `O(n^2)`. Let's remove it.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -294,15 +301,27 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+            if unused_spk_count >= stop_gap {
+                break;
+            }
+        }
+
+        // Batch validate all collected transactions.
+        if !txs_to_validate.is_empty() {
+            let proofs = self.batch_fetch_merkle_proofs(&txs_to_validate)?;
+            self.batch_validate_merkle_proofs(tx_update, proofs)?;
         }
```
evanlinjin: Instead of having every `populate_with_{}` method call this internally, it will be more efficient and make more logical sense if we extract this so that we only call it at the end of `full_scan` and `sync`.

In other words, `populate_with_{}` should no longer fetch anchors. Instead, they should either mutate, or return a list of `(Txid, BlockId)` for which we try to fetch anchors for in a separate step.

It will be even better if full txs are fetched in a separate step too.

evanlinjin (changes requested):
Thanks for moving this forward.

This is not a full review, but I think it's enough to push this PR in a good direction.

crates/electrum/src/bdk_electrum_client.rs
LagginTimes: Partially resolved. This is the next TODO:
> It will be even better if full txs are fetched in a separate step too.

crates/electrum/tests/test_electrum.rs
```diff
@@ -881,3 +882,51 @@ fn test_check_fee_calculation() -> anyhow::Result<()> {
     }
     Ok(())
 }
+
+#[test]
+pub fn test_sync_performance() -> anyhow::Result<()> {
+    const EXPECTED_MAX_SYNC_TIME: Duration = Duration::from_secs(15);
```
evanlinjin: This isn't really the right way to do benchmarks 😅 could you please use something proper like `criterion`? We already use it to benchmark canonicalization.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -412,52 +460,99 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+
         Ok(())
     }
 
-    // Helper function which checks if a transaction is confirmed by validating the merkle proof.
-    // An anchor is inserted if the transaction is validated to be in a confirmed block.
-    fn validate_merkle_for_anchor(
+    /// Batch validate Merkle proofs, cache each confirmation anchor, and return them.
+    fn batch_validate_anchors(
```
evanlinjin: Nit: `batch_fetch_anchors`

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -412,52 +460,99 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+        let mut to_fetch = Vec::new();
+
+        // Build a map for height to block hash conversions. This is for obtaining block hash data
+        // with minimum `fetch_header` calls.
+        let mut height_to_hash: HashMap<u32, BlockHash> = HashMap::new();
+        for &(_, height) in txs_with_heights {
+            let h = height as u32;
+            if !height_to_hash.contains_key(&h) {
+                // Try to obtain hash from the header cache, or fetch the header if absent.
+                let hash = self.fetch_header(h)?.block_hash();
```
evanlinjin: We can also batch fetch headers.

crates/electrum/tests/test_electrum.rs
LagginTimes: This was a carry-over from the original PR. Test has been changed to use `criterion`.

crates/electrum/src/bdk_electrum_client.rs
LagginTimes: ~This will likely be added in a separate PR.~
Done in this PR.

crates/electrum/src/bdk_electrum_client.rs
LagginTimes: ~This will likely be included in a separate PR.~
Fetching all full txs in a batch call at the beginning of `sync` actually ended up doubling `sync` time.

evanlinjin:
@LagginTimes could you provide the benchmark results in the PR description and compare it to results before the changes in this PR?

notmandatory:
Based on above benchmark results it looks like this change is 1s faster on sync, is that due to a small test size? Do we expect it to make more of a difference with wallets with many addresses?

evanlinjin:
@LagginTimes ~~can you provide the code you used to test with a remote electrum server (instead of the testenv)?~~

Edit: how about we just test with the example-cli with a pre-populated signet wallet?

I suggested writing benchmarks with the assumption that local io (against testenv) would be slower than allocating memory (collecting requests into vec before batch requesting), however that assumption seems incorrect.

crates/electrum/src/bdk_electrum_client.rs
evanlinjin: @LagginTimes did you figure out why though?

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -292,15 +301,21 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
                         _ => {
                             tx_update.seen_ats.insert((tx_res.tx_hash, start_time));
                         }
                     }
                 }
             }
+
+            if unused_spk_count >= stop_gap {
+                break;
+            }
```
ValuedMammal: Could you ~~also~~ instead use a named lifetime to break from the outer loop on L281?

Edit: I only mention it to not have to break in two places for the same reason, but it's probably fine either way.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -377,87 +396,178 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+        for &(txid, height, _) in &to_fetch {
+            batch.raw(
+                "blockchain.transaction.get_merkle".into(),
+                vec![
+                    electrum_client::Param::String(format!("{:x}", txid)),
+                    electrum_client::Param::Usize(height),
+                ],
             );
+        }
+        let resps = self.inner.batch_call(&batch)?;
```
ValuedMammal, line 520: I guess the batch version of `transaction.get_merkle` could be implemented as an API method in `electrum-client`.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -377,87 +396,178 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+        }
+
+        let height_to_hash: HashMap<u32, BlockHash> = txs_with_heights
+            .iter()
+            .map(|&(_, h)| {
+                let h = h as u32;
+                (h, cache[&h].block_hash())
+            })
+            .collect();
+        drop(cache);
```
ValuedMammal: This part seemed a little busy, but maybe something like this would be more clear.

```rs
let mut height_to_hash = HashMap::with_capacity(needed_heights.len());

// Collect headers of missing heights, and build `height_to_hash` map.
{
    let mut cache = self.block_header_cache.lock().unwrap();

    let mut missing_heights = Vec::new();
    for &height in &needed_heights {
        if let Some(header) = cache.get(&height) {
            height_to_hash.insert(height, header.block_hash());
        } else {
            missing_heights.push(height);
        }
    }

    if !missing_heights.is_empty() {
        let headers = self.inner.batch_block_header(missing_heights.clone())?;
        for (height, header) in missing_heights.into_iter().zip(headers) {
            height_to_hash.insert(height, header.block_hash());
            cache.insert(height, header);
        }
    }
}
```

crates/electrum/src/bdk_electrum_client.rs
evanlinjin: Upstream later (as we don't know when that'll be merged).

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -377,87 +396,178 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
         start_time: u64,
         tx_update: &mut TxUpdate<ConfirmationBlockTime>,
         txids: impl IntoIterator<Item = Txid>,
+        pending_anchors: &mut Vec<(Txid, usize)>,
     ) -> Result<(), Error> {
+        let mut txs = Vec::new();
```
ValuedMammal: You could change this to `txs: Vec<(Txid, Arc<Transaction>)>` to avoid recomputing the txid later in the function.

ValuedMammal: I'm unsure if this suggestion is actually supported by the benchmark though.

crates/electrum/benches/test_sync.rs
```diff
@@ -0,0 +1,106 @@
...
+    I: Indexer,
+    I::ChangeSet: Default + Merge,
+    Spks: IntoIterator<Item = ScriptBuf>,
+    Spks::IntoIter: ExactSizeIterator + Send + 'static,
+{
+    let update = client.sync(
+        SyncRequest::builder().chain_tip(chain.tip()).spks(spks),
+        BATCH_SIZE,
+        true,
+    )?;
```
ValuedMammal: I feel like there should be a sanity check assertion to check that the update actually returns tx data - when I was testing, it looked to be empty.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -377,87 +396,175 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
-            };
+            match self.fetch_tx(txid) {
+                Ok(tx) => {
+                    let spk = tx
+                        .output
+                        .first()
+                        .map(|txo| &txo.script_pubkey)
+                        .expect("tx must have an output")
+                        .clone();
+                    txs.push((tx.compute_txid(), tx));
```
ValuedMammal:
```suggestion
                    txs.push((txid, tx));
```

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -377,87 +396,175 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
-                            height: merkle_res.block_height as u32,
-                            hash: header.block_hash(),
-                        },
+            // Build and cache the anchor if merkle proof is valid.
+            if valid {
+                let anchor = ConfirmationBlockTime {
+                    confirmation_time: header.time as u64,
+                    block_id: BlockId {
+                        height: height as u32,
+                        hash: header.block_hash(),
```
ValuedMammal: nit: Is this the same block hash from `to_fetch` ?
```suggestion
                        hash,
```

ValuedMammal (changes requested):

crates/electrum/benches/test_sync.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    // Setup receiver.
+    let genesis = env.bitcoind.client.get_block_hash(0).unwrap();
+    let (chain, _) = LocalChain::from_genesis_hash(genesis);
+    let graph = IndexedTxGraph::<ConfirmationBlockTime, _>::new({
+        let mut idx = SpkTxOutIndex::default();
+        for spk in &spks {
+            idx.insert_spk((), spk.clone());
+        }
```
ValuedMammal: If there's only one indexed spk, then I think just insert it once into the index.
```suggestion
        idx.insert_spk((), spks[0].clone());
```

crates/electrum/benches/test_sync.rs
LagginTimes: Added an explicit `wait_until_electrum_sees_block()` after `mine_blocks()` so Electrs has time to index before we sync. I also dropped the `sample_size` to keep the bench from getting killed. We can still fine-tune these settings down the road if needed.

ValuedMammal (approved):

ValuedMammal:
These are my criterion results after benching this PR. 👍 

```sh
sync_with_electrum      time:   [37.325 ms 38.220 ms 38.805 ms]                              
                        change: [-85.756% -85.556% -85.308%] (p = 0.00 < 0.05)
                        Performance has improved.
```

ValuedMammal:
In 838c24762b8b03a34a64804b9b7a5e6626f03aad:

```sh
error: this file contains an unclosed delimiter
   --> crates/electrum/src/bdk_electrum_client.rs:724:3
    |
32  | impl<E: ElectrumApi> BdkElectrumClient<E> {
    |                                           - unclosed delimiter
...
504 |             for &(txid, height, hash) in chunk {
    |                                                - this delimiter might not be properly closed...
...
547 |         }
    |         - ...as it matches this but it has different indentation
...
724 | }
```

ValuedMammal (approved):
ACK 8086e1ef752a9ef5ca3908f0a2f2cef68256473e

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -249,16 +257,17 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+        let mut last_active_index = None;
 
-        loop {
+        'batch_loop: loop {
             let spks = (0..batch_size)
                 .map_while(|_| spks_with_expected_txids.next())
                 .collect::<Vec<_>>();
             if spks.is_empty() {
-                return Ok(last_active_index);
+                break;
```
evanlinjin: Is this set of changes intended to improve readability? I think it is doing the opposite effect.

* `0_usize` is clearer than `0` as it contains the type and the reader is not required to rely on the LSP.
* `Option::<u32>::None` is clear than `None`.
* `return Ok(last_active_index)` is clearer than `break` as it is more linear to read. `break` means we need to follow where the loop ends.

I recommend reverting these changes.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -267,9 +276,9 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
 
             for ((spk_index, spk), spk_history) in spks.into_iter().zip(spk_histories) {
                 if spk_history.is_empty() {
-                    unused_spk_count = unused_spk_count.saturating_add(1);
+                    unused_spk_count += 1;
```
evanlinjin: **Nit:**

I see how `.saturating_add` can be problematic if we do reach the impossible situation of hitting `usize::MAX`.

If we want to have "logical soundness", I think this approach would be better:

```suggestion
                    match unused_spk_count.checked_add(1) {
                        Some(i) if i < stop_gap => unused_spk_count = i,
                        _ => return Ok(last_active_index),
                    };
```

crates/electrum/src/bdk_electrum_client.rs
evanlinjin: I have to disagree with this.

Directly returning is more readable than `break`ing with a label as it is linear to read. Reading a labelled `break` means searching for the labelled loop and also where the loop ends to know where it breaks to.

Also, this seems out of scope for this PR.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -377,87 +396,175 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+            };
+            let mut valid =
+                electrum_client::utils::validate_merkle_proof(&txid, &header.merkle_root, &proof);
+            if !valid {
+                let new_header = self.inner.block_header(height)?;
+                self.block_header_cache
+                    .lock()
+                    .unwrap()
+                    .insert(height as u32, new_header);
+                header = new_header;
```
```suggestion
            let mut header = {
                let cache = self.block_header_cache.lock().unwrap();
                cache
                .get(&(height as u32))
                .copied()
                .expect("header already fetched above")
            };
            let mut valid =
                electrum_client::utils::validate_merkle_proof(&txid, &header.merkle_root, &proof);
            if !valid {
                header = self.inner.block_header(height)?;
                self.block_header_cache
                    .lock()
                    .unwrap()
                    .insert(height as u32, header);
```

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -377,87 +396,175 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+    fn validate_anchor_for_update(
+        &self,
+        tx_update: &mut TxUpdate<ConfirmationBlockTime>,
+        txid: Txid,
+        confirmation_height: usize,
+    ) -> Result<(), Error> {
+        let anchors = self.batch_fetch_anchors(&[(txid, confirmation_height)])?;
+        for (txid, anchor) in anchors {
+            tx_update.anchors.insert((anchor, txid));
+        }
```
evanlinjin: Is this used anywhere?

crates/electrum/benches/test_sync.rs
```diff
@@ -0,0 +1,111 @@
...
+    let (chain, _) = LocalChain::from_genesis_hash(genesis);
+    let graph = IndexedTxGraph::<ConfirmationBlockTime, _>::new({
+        let mut idx = SpkTxOutIndex::default();
+        idx.insert_spk((), spks[0].clone());
+        idx
+    });
+
+    c.bench_function("sync_with_electrum", move |b| {
+        b.iter(|| {
+            let spks = black_box(spks.clone());
```
evanlinjin: I don't see why `black_box` will be needed here.

crates/electrum/benches/test_sync.rs
```diff
@@ -0,0 +1,111 @@
...
+        !update.tx_update.txs.is_empty(),
+        "expected some transactions from sync, but got none"
+    );
+
+    if let Some(chain_update) = update.chain_update.clone() {
+        let _ = chain
+            .apply_update(chain_update)
+            .map_err(|err| anyhow::anyhow!("LocalChain update error: {:?}", err))?;
+    }
+    let _ = graph.apply_update(update.tx_update.clone());
```
evanlinjin: I don't think applying the update should be part of this benchmark.

crates/electrum/benches/test_sync.rs
```diff
@@ -0,0 +1,111 @@
...
+    Ok(update)
+}
+
+pub fn test_sync_performance(c: &mut Criterion) {
+    let env = TestEnv::new().unwrap();
+    let electrum_client = electrum_client::Client::new(env.electrsd.electrum_url.as_str()).unwrap();
+    let client = BdkElectrumClient::new(electrum_client);
+
+    const NUM_BLOCKS: usize = 100;
+    let mut spks = Vec::with_capacity(NUM_BLOCKS);
```
evanlinjin, line 58: It seems we are constructing a list of `spks`, but each element is the same spk.

To fix this, we can change the signature of `get_test_spk` to take in an integer to tweak the spk's pk.

crates/electrum/benches/test_sync.rs
```diff
@@ -0,0 +1,111 @@
...
+    key::{Secp256k1, UntweakedPublicKey},
+    Network,
+};
+use bdk_electrum::BdkElectrumClient;
+use bdk_testenv::{anyhow, bitcoincore_rpc::RpcApi, TestEnv};
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use std::time::Duration;
+
+// Batch size for `sync_with_electrum`.
+const BATCH_SIZE: usize = 5;
```
evanlinjin: This is a very small `BATCH_SIZE`.

evanlinjin (changes requested):
Here are some suggestions to improve the benchmark.

I've implemented them in this commit: https://github.com/evanlinjin/bdk/commit/1ba324ca653a72ca410f8dff147d78823fdb6369

I also added a benchmark for testing sync without cache. Feel free to cherry-pick this commit if you agree with the changes.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -22,6 +22,8 @@ pub struct BdkElectrumClient<E> {
     tx_cache: Mutex<HashMap<Txid, Arc<Transaction>>>,
     /// The header cache
     block_header_cache: Mutex<HashMap<u32, Header>>,
+    /// Cache of transaction anchors
+    anchor_cache: Mutex<HashMap<(Txid, BlockHash), ConfirmationBlockTime>>,
```
evanlinjin, line 26: Can be done in a follow up, but we can have a `populate_anchor_cache` method. This will speed things up even more for subsequent syncs.

evanlinjin: Ticket made: https://github.com/bitcoindevkit/bdk/issues/1982

crates/electrum/src/bdk_electrum_client.rs
evanlinjin: Is this still needed?

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -292,15 +301,21 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
                     }
                 }
             }
+
+            if unused_spk_count >= stop_gap {
+                break;
+            }
         }
+
+        Ok(last_active_index)
```
evanlinjin: Also here.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -312,40 +327,43 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+        let ops_txs = outpoints
+            .into_iter()
+            .filter_map(|op| self.fetch_tx(op.txid).ok().map(|tx| (op, tx)))
+            .collect::<Vec<_>>();
 
-            // attempt to find the following transactions (alongside their chain positions), and
-            // add to our sparsechain `update`:
+        let scripts = ops_txs
+            .iter()
+            .map(|(op, tx)| tx.output[op.vout as usize].script_pubkey.as_script());
```
evanlinjin: This is dangerous if the caller provides an invalid outpoint.

Additionally, we may duplicate spk-history-fetches with this logic.

Let's kill two birds with one stone:

* Collect spk histories into a `Map<spk, history>`.
* Collect outpoints into a `Vec<(outpoint, spk-of-outpoint)>` while filtering out invalid outpoints.
* Iterate over `Vec<(outpoint, spk-of-outpoint)>` and get spk history from the `Map<spk, history>` (if available).

evanlinjin (changes requested):

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -312,62 +321,82 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
+        // Dedup `spk`s, batch-fetch all histories in one call, and store them in a map.
+        let unique_spks: Vec<_> = ops_spks_txs
+            .iter()
+            .map(|(_, spk, _)| spk.clone())
+            .collect::<HashSet<_>>()
+            .into_iter()
+            .collect();
+        let histories = self
+            .inner
+            .batch_script_get_history(unique_spks.iter().map(|spk| spk.as_script()))?;
```
oleonardolima, line 345: question to self: does this batch call guarantees order ?

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -377,88 +406,163 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
-                .find(|r| r.tx_hash == txid)
-            {
-                match r.height.try_into() {
+        for (tx, spk_history) in txs.into_iter().zip(spk_histories) {
+            if let Some(res) = spk_history.into_iter().find(|res| res.tx_hash == tx.0) {
+                match res.height.try_into() {
                     // Returned heights 0 & -1 are reserved for unconfirmed txs.
                     Ok(height) if height > 0 => {
-                        self.validate_merkle_for_anchor(tx_update, txid, height)?;
+                        pending_anchors.push((tx.0, height));
```
oleonardolima, line 443: nit: use res.tx_hash for consistency with the branch below.

oleonardolima (commented):
tACK 156cbab67f4ff91276f9f03749944f4c46210f7f

It looks good to me, I did run the example with the same descriptor, but a different server (and no TLS), here are the results: 

```
master @ `63923c63dc5dbd7850ae8fa4f4d1b832170fe957`
FULL_SCAN TIME: 26.977265542s
SYNC TIME: 23.786322458s
```

```
merkle_batching @ `156cbab67f4ff91276f9f03749944f4c46210f7f`
FULL_SCAN TIME: 3.85442175s
SYNC TIME: 4.124651292s
```

crates/electrum/src/bdk_electrum_client.rs
evanlinjin: If it doesn't, then it is a major bug in `bdk_client` since the API implies that the response order correlates with the requests.

evanlinjin (approved):
ACK 156cbab67f4ff91276f9f03749944f4c46210f7f

Styling tips for readability:

* I find that it is more readable to initialize a variable with it's type rather than relying on it being implied later on. I.e. prefer `let mut result = Vec::<(Txid, ConfirmationBlockTime)>::new()` over `let mut result = Vec::new()`. The latter requires the reader to use an LSP or scan head first to figure out the type.
* Avoid overly-nested logic. Sometimes it is better to use `continue` instead of nesting the logic in an `if` clause.
* `debug_assert!`s help us to catch bugs. Don't be afraid to add them where `panic/expect` would not be appropriate.

PR #1963, opened 2025-05-22
Persist spks derived from `KeychainTxOutIndex`

### LINKED ISSUE #1953: Generating SPK Index upon loading wallet is slow for large wallets with lots of addresses
**Describe the enhancement**  
Wallets with large `last_revealed` indexes are slow to generate the SPK cache when the wallet is loaded. I discovered this with a wallet that has an index of 7,000 addresses.

**Use case**  
I am loading my wallets using the PersistTrait and this SPK list gets generated every time, which for large numbers is slow. I would like to load wallets from persistence as quickly as possible (so I can show the balance).

**Additional context**
There are a number of ways which were discussed in a call.

- We can change KeychainTxOutIndex to persist the spks. 
    - needs to be careful to check them 

- implement a whole changeset for spk index (@ValuedMammal)

- make it optional since its not strictly needed for checking balance or listing utxos.

Some context from the call
The problem is with the KeychainTxOutIndex. @evanlinjin

### LINKED ISSUE #1964: Implement `IndexedTxGraph::reindex`
**Describe the enhancement**  

Implement `pub fn reindex(&mut self)` so that it reindexes all transactions and transaction outputs (txouts) in the `TxGraph` by passing them through the indexer.

**Use case**  

When a caller inserts a new descriptor into a `KeychainTxOutIndex`, calling `reindex()` ensures that any existing transactions or txouts in the graph matching the new descriptor are correctly indexed.

evanlinjin:
Replaces #1960
Fixes #1953
Fixes #1964 

### Description

Users with large wallet and/or complex descriptors may experience slow startup of `KeychainTxOutIndex`. This PR addresses this problem by providing the option to persist derived spks so that they no longer need to be re-derived on startup.

The `IndexedTxGraph` API has been changed for better ergonomics.

Compared to #1960, this is a more longterm solution that does not depend on multi-threading logic.

### Changelog notice

```md
Changed
  - `KeychainTxOutIndex::new` to take in an additional parameter `persist_spks` to control whether derived spks are cached and persisted across restarts. The default of `persist_spks` is false.
  - `KeychainTxOutIndex` methods (`lookahead_to_target, `next_unused_spk`, `reveal_next_spk`) now return changesets as they may derive spks to be persisted.
  - The `InsertDescriptorError` type now wraps descriptors in `Box` to reduce enum size.

Added
  - `spk_cache` field to `indexer::keychain_txout::ChangeSet` which persists derived spks.
  - `IndexedTxGraph::from_changeset` - allows constructing from `indexed_tx_graph::ChangeSet` and only indexing when ready.
  - `IndexedTxGraph::reindex` method.

Fixed
  - `KeychainTxOutIndex::reveal_to_target` so that it actually returns `None` if the `keychain` does not exist.
```

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/indexed_tx_graph.rs
+++ b/crates/chain/src/indexed_tx_graph.rs
@@ -15,12 +15,14 @@ use crate::{
     Anchor, BlockId, ChainOracle, Indexer, Merge, TxPosInBlock,
 };
 
-/// The [`IndexedTxGraph`] combines a [`TxGraph`] and an [`Indexer`] implementation.
+/// A [`TxGraph<A>`] paired with an indexer `I`, enforcing that every insertion into the graph is
+/// simultaneously fed through the indexer.
 ///
-/// It ensures that [`TxGraph`] and [`Indexer`] are updated atomically.
+/// This guarantees that `tx_graph` and `index` remain in sync: any transaction or floating txout
+/// you add to `tx_graph` has already been processed by `index`.
 #[derive(Debug, Clone)]
 pub struct IndexedTxGraph<A, I> {
-    /// Transaction index.
+    /// The indexer used for filtering transactions and floating txouts that we are interested in.
     pub index: I,
     graph: TxGraph<A>,
 }
@@ -35,14 +37,6 @@ impl<A, I: Default> Default for IndexedTxGraph<A, I> {
 }
 
 impl<A, I> IndexedTxGraph<A, I> {
-    /// Construct a new [`IndexedTxGraph`] with a given `index`.
-    pub fn new(index: I) -> Self {
-        Self {
-            index,
-            graph: TxGraph::default(),
-        }
-    }
-
     /// Get a reference of the internal transaction graph.
     pub fn graph(&self) -> &TxGraph<A> {
         &self.graph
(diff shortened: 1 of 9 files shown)


crates/chain/src/rusqlite_impl.rs
```diff
@@ -533,10 +535,28 @@ impl keychain_txout::ChangeSet {
...
     }
 
+    /// Get v1 of sqlite [keychain_txout::ChangeSet] schema
+    pub fn schema_v1() -> String {
+        format!(
+            "CREATE TABLE {} ( \
+            descriptor_id TEXT NOT NULL REFERENCES {}, \
+            spk_index INTEGER NOT NULL, \
+            spk BLOB NOT NULL, \
+            PRIMARY KEY (descriptor_id, index) \
```
ValuedMammal: Small typo in column name.

```suggestion
            PRIMARY KEY (descriptor_id, spk_index) \
```

crates/chain/src/rusqlite_impl.rs
```diff
@@ -533,10 +535,28 @@ impl keychain_txout::ChangeSet {
         )
     }
 
+    /// Get v1 of sqlite [keychain_txout::ChangeSet] schema
+    pub fn schema_v1() -> String {
+        format!(
+            "CREATE TABLE {} ( \
+            descriptor_id TEXT NOT NULL REFERENCES {}, \
```
ValuedMammal: While testing I ran into a failed foreign key constraint because bdk attempted to write to the derived spks table before any scripts were revealed hence the last_revealed table was empty.
```suggestion
            descriptor_id TEXT NOT NULL, \
```

ValuedMammal (changes requested):

ValuedMammal (approved):
ACK 62767f0702e6497d08fea702b747fd6e7fa72b8c

notmandatory:
~~Might have a problem with the sqlite store. I'm manually testing and looks like it's storing the same script for all indexes. I'll keep looking but wanted to give a headsup not to merge this yet.~~
Nevermind, it's working fine, my sqlite select just wan't showing the blog data.

notmandatory (approved):
ACK 62767f0702e6497d08fea702b747fd6e7fa72b8c

tested with `bdk-cli` with cache enabled and sqlite schema updated and new table is being populated.

crates/chain/src/indexed_tx_graph.rs
```diff
@@ -79,6 +73,87 @@ impl<A: Anchor, I: Indexer> IndexedTxGraph<A, I>
...
+    /// # use bdk_testenv::anyhow;
+    /// # use miniscript::{Descriptor, DescriptorPublicKey};
+    /// # use std::str::FromStr;
+    /// # let persisted_changeset = ChangeSet::<BlockId, _>::default();
+    /// # let persisted_desc = Some(Descriptor::<DescriptorPublicKey>::from_str("")?);
+    /// # let persisted_change_desc = Some(Descriptor::<DescriptorPublicKey>::from_str("")?);
+    ///
+    /// let (graph, reindex_cs) =
+    ///     IndexedTxGraph::from_changeset(persisted_changeset, move |idx_cs| -> anyhow::Result<_> {
+    ///         // e.g. KeychainTxOutIndex needs descriptors that weren’t in its CS
```
ValuedMammal, line 115:
```rs
// e.g. KeychainTxOutIndex needs descriptors that weren’t in its change set.
```

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -186,24 +200,44 @@ impl<K: Clone + Ord + Debug> Indexer for KeychainTxOutIndex<K> {
 }
 
 impl<K> KeychainTxOutIndex<K> {
-    /// Construct a [`KeychainTxOutIndex`] with the given `lookahead`.
+    /// Construct a [`KeychainTxOutIndex`] with the given `lookahead` and `use_spk_cache` boolean.
```
ValuedMammal, line 203:
```rs
/// Construct a [`KeychainTxOutIndex`] with the given `lookahead` and `persist_spks` boolean.
```

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -186,24 +200,44 @@ impl<K: Clone + Ord + Debug> Indexer for KeychainTxOutIndex<K> {
...
+    /// re-derived on init, affecting startup only for very large or complex wallets.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use bdk_chain::keychain_txout::KeychainTxOutIndex;
+    /// // Derive 20 future addresses per chain and persist + reload script pubkeys via ChangeSets:
+    /// let idx = KeychainTxOutIndex::<&'static str>::new(20, true);
+    ///
+    /// // Derive 10 future addresses per chain without persistence:
```
ValuedMammal, line 227: Per "keychain" ?

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -215,6 +249,64 @@ impl<K> KeychainTxOutIndex<K> {
 
 /// Methods that are *re-exposed* from the internal [`SpkTxOutIndex`].
 impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
+    /// Construct `KeychainTxOutIndex<K>` from the given `changeset`.
+    ///
+    /// Shorthand for called [`new`] and then [`apply_changeset`].
```
ValuedMammal, line 254: Shorthand for "calling"...

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -834,14 +998,22 @@ impl<K: core::fmt::Debug> core::fmt::Display for InsertDescriptorError<K> {
 #[cfg(feature = "std")]
 impl<K: core::fmt::Debug> std::error::Error for InsertDescriptorError<K> {}
 
-/// Represents updates to the derivation index of a [`KeychainTxOutIndex`].
-/// It maps each keychain `K` to a descriptor and its last revealed index.
+/// `ChangeSet` represents persistent updates to a [`KeychainTxOutIndex`].
+///
+/// It tracks:
+/// 1. `last_revealed`: the highest derivation index revealed per descriptor.
+/// 2. `spks`: the cache of derived script pubkeys to persist across runs.
```
ValuedMammal, line 1005:
```rs
/// 2. `spk_cache`: ...
```

PR #1971, opened 2025-06-06
esplora: `chain_update` errors if no point of connection

ValuedMammal:
### Description

Before, the `chain_update` function might have panicked if the local checkpoint was not on the same network as the remote server. Now if we have iterated all of the blocks of the local CP and do not find a point of agreement, then we return early with a `esplora_client::Error::HeaderHashNotFound`.

cc bitcoindevkit/bdk_wallet#30

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/esplora/src/async_ext.rs
+++ b/crates/esplora/src/async_ext.rs
@@ -205,12 +205,10 @@ async fn fetch_block<S: Sleeper>(
 
     // We avoid fetching blocks higher than previously fetched `latest_blocks` as the local chain
     // tip is used to signal for the last-synced-up-to-height.
-    let &tip_height = latest_blocks
-        .keys()
-        .last()
-        .expect("must have atleast one entry");
-    if height > tip_height {
-        return Ok(None);
+    if let Some(tip_height) = latest_blocks.keys().last().copied() {
+        if height > tip_height {
+            return Ok(None);
+        }
     }
 
     Ok(Some(client.get_block_hash(height).await?))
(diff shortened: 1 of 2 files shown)


crates/esplora/src/async_ext.rs
```diff
@@ -205,12 +205,10 @@ async fn fetch_block<S: Sleeper>(
...
-        .keys()
-        .last()
-        .expect("must have atleast one entry");
-    if height > tip_height {
-        return Ok(None);
+    if let Some(tip_height) = latest_blocks.keys().last().copied() {
+        if height > tip_height {
+            return Ok(None);
+        }
     }
```
evanlinjin: I think it's better to return `None` if there are no `latest_blocks` as the `latest_blocks` protects us from creating a chain update with inconsistent state. 

However, `latest_blocks` should really never be empty - how about we put a `debug_assert!` here?

crates/esplora/src/async_ext.rs
```diff
@@ -226,28 +224,33 @@ async fn chain_update<S: Sleeper>(
...
-    let mut point_of_agreement = None;
+    let mut cp = local_tip.clone();
     let mut conflicts = vec![];
-    for local_cp in local_tip.iter() {
-        let remote_hash = match fetch_block(client, latest_blocks, local_cp.height()).await? {
+
+    let mut tip = loop {
+        let remote_hash = match fetch_block(client, latest_blocks, cp.height()).await? {
             Some(hash) => hash,
             None => continue,
```
evanlinjin, line 240: Is the intention that we continuously try fetch the same block of height until this returns some? 

This will be problematic if the server is very behind as syncing will hang and start spamming the server.

ValuedMammal: Oops, no it wasn't intentional. I changed back to `for` loop and settled on the approach in b90f4cc.

oleonardolima (commented):
utACK 96a62074870f0cb3c9a84444b7dab69ac1645098

It's valid to note that we covered the `fetch_blocks` emptyness scenario on the client side too, at https://github.com/bitcoindevkit/rust-esplora-client/pull/113

crates/esplora/src/async_ext.rs
```diff
@@ -557,6 +570,41 @@ mod test {
...
+        let env = TestEnv::new()?;
+        let base_url = format!("http://{}", &env.electrsd.esplora_url.clone().unwrap());
+        let client = Builder::new(base_url.as_str()).build_async()?;
+        let initial_height = env.rpc_client().get_block_count()? as u32;
+
+        let mine_to = 16;
+        let _ = env.mine_blocks((mine_to - initial_height) as usize, None)?;
+        while client.get_height().await? < mine_to {
+            std::thread::sleep(Duration::from_millis(64));
+        }
```
oleonardolima, line 585: nit: FWIW, you can use `wait_until_electrum_sees_block` if #1640 ends up landing before this one.

LagginTimes (approved):
ACK 96a62074870f0cb3c9a84444b7dab69ac1645098

ValuedMammal:
I had to rebase after #1978.

ACK 568a36642c43d74132339a4e42dd41c545a764e1

PR #1979, opened 2025-06-18
feat: add `justfile`

### LINKED ISSUE #1967: Create justfile
`just` is the de facto standard for command runner utilities. I think it would be useful to have one here.

luisschwab:
### Description

Closes #1967.

This PR adds a `justfile`, updates the PR template to use it, and adds a section on the `README.md` to show available recipes.

These are the implemented recipes:
```justfile
alias b := build
alias c := check
alias f := fmt
alias t := test
alias p := pre-push

_default:
  @just --list

# Build the project
build:
   cargo build

# Check code: formatting, compilation, linting, and commit signature
check:
   cargo +nightly fmt --all -- --check
   cargo check --workspace --all-features
   cargo clippy --all-features --all-targets -- -D warnings
   @[ "$(git log --pretty='format:%G?' -1 HEAD)" = "N" ] && \
       echo "\n⚠️  Unsigned commit: BDK requires that commits be signed." || \
       true

# Format all code
fmt:
   cargo +nightly fmt

# Run all tests for all crates with all features enabled
test:
   @just _test-bitcoind_rpc
   @just _test-chain
   @just _test-core
   @just _test-electrum
   @just _test-esplora
   @just _test-file_store
   @just _test-testenv

_test-bitcoind_rpc:
    cargo test -p bdk_bitcoind_rpc --all-features

_test-chain:
    cargo test -p bdk_chain --all-features

_test-core:
    cargo test -p bdk_core --all-features

_test-electrum:
    cargo test -p bdk_electrum --all-features

_test-esplora:
    cargo test -p bdk_esplora --all-features

_test-file_store:
    cargo test -p bdk_file_store --all-features

_test-testenv:
    cargo test -p bdk_testenv --all-features

# Run pre-push suite: format, check, and test
pre-push: fmt check test
```

`check` will verify if `HEAD` was signed and echo that warning if not. It does not check all commits, but I think that checking only the last is a pretty good heuristic (who signs the last commit only?).

Before pushing, one only needs to run `just p`.

#### All Submissions:

* [X] I ran `just p` before pushing

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -18,9 +18,7 @@ of the PR were done in a specific way -->
 
 #### All Submissions:
 
-* [ ] I've signed all my commits
 * [ ] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
-* [ ] I ran `cargo +nightly fmt` and `cargo clippy` before committing
 
 #### New Features:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,7 @@ To facilitate communication with other contributors, the project is making use
 of GitHub's "assignee" field. First check that no one is assigned and then
 comment suggesting that you're working on it. If someone is already assigned,
 don't hesitate to ask if the assigned party or previous commenter are still
-working on it if it has been awhile.
+working on it if it has been a while.
 
 Deprecation policy
 ------------------
@@ -87,7 +87,7 @@ Coding Conventions
 ------------------
 
 This codebase uses spaces, not tabs.
-Use `cargo fmt` with the default settings to format code before committing.
+Run `just check` to check formatting, linting, compilation and commit signing, `just fmt` to format code before commiting, and `just test` to run tests for all crates.
 This is also enforced by the CI.
 All public items must be documented. We adhere to the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/about.html) with respect to documentation.
 
@@ -105,6 +105,10 @@ Note that BDK is currently considered "pre-production" during this time, there
 is no special handling of security issues. Please simply open an issue on
 Github.
 
+BDK requires all commits to be signed using PGP. Refer to
+[this guide](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work)
+if you don't have a PGP key set up with `git` yet.
+
 Testing
 -------
 
@@ -117,16 +121,16 @@ effort.
 First Time Contributors
 -----------------------
 
-If it is your first time contributing to the BDK family of libraries, welcome! We're glad to have you with us. If your 
-first (or few first) PRs are focused on very small fixes to documentation, however, they might not meet our threshold 
+If it is your first time contributing to the BDK family of libraries, welcome! We're glad to have you with us. If your
+first (or few first) PRs are focused on very small fixes to documentation, however, they might not meet our threshold
 for acceptance for first time contributors.
 
-Minor grammar and punctuation fixes aren't a good way to start contributing to a project, and instead we suggest you 
-start with something a little more substantial. It's better to find an issue where you can demonstrate some knowledge 
-of bitcoin or the code base, such as improving the substance of documentation, testing, or fixing some small issue 
+Minor grammar and punctuation fixes aren't a good way to start contributing to a project, and instead we suggest you
+start with something a little more substantial. It's better to find an issue where you can demonstrate some knowledge
+of bitcoin or the code base, such as improving the substance of documentation, testing, or fixing some small issue
 even if it's considered low priority.
 
-This being said we are always looking forward to working with new folks interested in contributing to our libraries. 
+This being said we are always looking forward to working with new folks interested in contributing to our libraries.
 If you are looking for issues to work on, check out the good first issue label and join our Discord server!
 
 Going further
--- a/justfile
+++ b/justfile
@@ -1,0 +1,59 @@
+alias b := build
+alias c := check
+alias f := fmt
+alias t := test
+alias p := pre-push
+
+_default:
+  @just --list
+
+# Build the project
+build:
+   cargo build
+
+# Check code: formatting, compilation, linting, and commit signature
+check:
+   cargo +nightly fmt --all -- --check
+   cargo check --workspace --all-features
+   cargo clippy --all-features --all-targets -- -D warnings
+   @[ "$(git log --pretty='format:%G?' -1 HEAD)" = "N" ] && \
+       echo "\n⚠️  Unsigned commit: BDK requires that commits be signed." || \
+       true
+
+# Format all code
+fmt:
+   cargo +nightly fmt
+
+# Run all tests for all crates with all features enabled
+test:
+   @just _test-bitcoind_rpc
+   @just _test-chain
+   @just _test-core
+   @just _test-electrum
+   @just _test-esplora
+   @just _test-file_store
+   @just _test-testenv
+
+_test-bitcoind_rpc:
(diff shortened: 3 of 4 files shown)


luisschwab:
Did some thinking on this. Should we really exclude examples from formatting?

notmandatory:
You mean exclude examples from `cargo check`?  I think the only reason we exclude them in CI was to avoid MSRV dependency issues, but for this `Justfile` it shouldn't be a problem so probably should be checking everything.

luisschwab:
Yes, `check` and also `fmt --check`.

luisschwab:
If there are no objections I'll make these changes on `bdk_wallet` as well.

ValuedMammal:
The example crates really should be excluded since they're not part of the core library. I'm still in favor of making `examples` into a separate workspace (if not outright removing them), and any just/cargo commands should be invoked from the examples/ directory. For wallet it will be easier to move the "examples" to the `wallet/examples` directory and include the necessary blockchain crates as dev-dependencies.

luisschwab:
Makes sense. I reverted the changes to exclude the `examples` directory. Any other nits? If not I think this is ready to merge.

luisschwab:
CI is broken:

```shell
running 3 tests
test blocking_ext::test::update_local_chain ... FAILED
test blocking_ext::test::test_finalize_chain_update ... ok
test async_ext::test::test_finalize_chain_update ... ok

failures:

---- blocking_ext::test::update_local_chain stdout ----
Error: Minreq(IoError(Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" }))


failures:
    blocking_ext::test::update_local_chain

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 27.60s

error: test failed, to rerun pass `-p bdk_esplora --lib`
```

luisschwab:
Friendly ping @notmandatory @ValuedMammal

ValuedMammal:
@luisschwab Thanks for the reminder. About the example crates, my comment was more about the overall structure of the repo, not to say they have to be excluded from the justfile entirely. The argument for having a separate command for the examples though is that `example_cli` already doesn't respect the MSRV, so isn't subject to the same clippy lints for example. Obviously the example code should be formatted, which is covered by `just fmt`.

justfile
```diff
@@ -0,0 +1,32 @@
...
+  @just --list
+
+# Build the project
+build:
+   cargo build
+
+# Check code: formatting, compilation, linting, and commit signature
+check:
+   cargo +nightly fmt --all -- --check
+   cargo check --workspace --exclude 'examples' --all-features
```
ValuedMammal: I was thinking this needed to be a package specifier, not a path.
```suggestion
   cargo check --workspace --exclude 'example_*' --all-features
```

justfile
```diff
@@ -0,0 +1,32 @@
...
+       echo "\n⚠️  Unsigned commit: BDK requires that commits be signed." || \
+       true
+
+# Format all code
+fmt:
+   cargo +nightly fmt
+
+# Run all tests on the workspace with all features
+test:
+   cargo test --workspace --exclude 'examples' --all-features
```
ValuedMammal: 📌 I think it would be a good idea to test each package individually. I was also going to suggest reducing `--test-threads` for tests that rely heavily on the testenv, but I just tested this on my own machine with no issues.

.github/pull_request_template.md
```diff
@@ -20,7 +20,7 @@ of the PR were done in a specific way -->
 
 * [ ] I've signed all my commits
 * [ ] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
-* [ ] I ran `cargo +nightly fmt` and `cargo clippy` before committing
+* [ ] I ran `just p` before pushing
```
ValuedMammal: nit: I think this looks tacky, and is now less readable than before. To a new contributor it's unclear what this does just by looking at it. The information about commit signing and code conventions should be included in `CONTRIBUTING.md`, and we can remove the extra check boxes from the PR template.

ValuedMammal:
If it makes things less complicated, then I also agree with @notmandatory that we don't have to be very strict about excluding the examples.

justfile
ValuedMammal: Same suggestion would apply to `just test`.

If we're following the `--exclude` pattern, then clippy would become

`cargo clippy --workspace --exclude 'example_*' --all-features --all-targets -- -D warnings`

luisschwab: My mistake, I though `--exclude` took a path and not a package.

justfile
luisschwab: We could do something like this:
```just
test:
   @just test-chain
   @just test-core
   @just test-file_store
   @just test-electrum
   @just test-esplora
   @just test-bitcoind_rpc

test-chain:
    cargo test -p bdk_chain --all-features

test-core:
    cargo test -p bdk_core --all-features

test-file_store:
    cargo test -p bdk_file_core --all-features

test-electrum:
    cargo test -p bdk_electrum --all-features

test-esplora:
    cargo test -p bdk_esplora --all-features

test-bitcoind_rpc:
    cargo test -p bdk_bitcoind_rpc --all-features
```

I could also add the underscore so those recipes don't show up when running `just`.

luisschwab: > I was also going to suggest reducing --test-threads for tests that rely heavily on the testenv, but I just tested this on my own machine with no issues.

I don't think that's necessary, running this on a low-end laptop doesn't take too long.

.github/pull_request_template.md
luisschwab: Agreed.

CONTRIBUTING.md
```diff
@@ -33,6 +33,8 @@ To contribute a patch, the workflow is as follows:
   1. Fork Repository
   2. Create topic branch
   3. Commit patches
+  4. Run checks for formatting, linting, compilation and commit signing: `just check`
+  5. Run the test suite: `just test`
```
ValuedMammal: Can we move this to somewhere under the Coding Conventions heading and fixup the stale comment on L92 that refers to using "`cargo fmt` with the default settings".

justfile
ValuedMammal, line 1:
```diff
 # Run all tests for all crates with all features enabled
 test:
    @just _test-bitcoind_rpc
    @just _test-chain
    @just _test-core
    @just _test-electrum
    @just _test-esplora
    @just _test-file_store
+   @just _test-testenv
 
 _test-bitcoind_rpc:
     cargo test -p bdk_bitcoind_rpc --all-features
 
 _test-chain:
     cargo test -p bdk_chain --all-features
 
 _test-core:
     cargo test -p bdk_core --all-features
 
 _test-electrum:
     cargo test -p bdk_electrum --all-features
 
 _test-esplora:
     cargo test -p bdk_esplora --all-features
 
 _test-file_store:
     cargo test -p bdk_file_store --all-features
 
+_test-testenv:
+    cargo test -p bdk_testenv --all-features
+
```

ValuedMammal (commented):
This looks great. Valid to note that the justfile can be used for basic sanity checking but at the moment doesn't help with switching toolchains, pinning dependencies, or checking various feature configurations, but these can be added as needed.

luisschwab:
> switching toolchains, pinning dependencies, or checking various feature configurations

What would this look like? We can add it now.

I'll also circle back to `bdk_wallet`  and add these changes there.

ValuedMammal (approved):
ACK 501766e7106f043d45b5ab6d29e9af83d6590e13

> switching toolchains, pinning dependencies, or checking various feature configurations

I think we should do more work on CI along the lines of #1944 - to help elucidate the relevant/important features, and uncover things we're not testing.

evanlinjin (approved):
ACK 501766e7106f043d45b5ab6d29e9af83d6590e13

PR #1980, opened 2025-06-18
feat!: Add #[non_exhaustive] to ChangeSet structs

notmandatory:
### Description

Add #[non_exhaustive] to ChangeSet structs in tx_graph, keychain_txout, local_chain.

Add #[non_exhaustive] attribute to the ChangeSet structs to prevent downstream crates from exhaustively matching or constructing instances using struct literal syntax, allowing future additions without breaking compatibility.

The change improves forward compatibility by allowing new fields to be added without breaking downstream code in future releases.

### Notes to the reviewers

The motivation for this is to prevent breaking changes in these structs but will still require downstream persistence crates to implement and support any new fields added. 

@ValuedMammal suggested creating a persistence test suite which I think is a great idea but should be done in a separate PR.

#### Breaking Changes
- Added `#[non_exhaustive]` attribute to `ChangeSet` structs in tx_graph, keychain_txout, local_chain modules.
  - Match expressions on these structs must now include wildcard (`..`) patterns.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/indexer/keychain_txout.rs
+++ b/crates/chain/src/indexer/keychain_txout.rs
@@ -1047,6 +1047,7 @@ impl<K: core::fmt::Display + core::fmt::Debug> core::error::Error for InsertDesc
 #[derive(Clone, Debug, Default, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[must_use]
+#[non_exhaustive]
 pub struct ChangeSet {
     /// Maps each `DescriptorId` to its last revealed derivation index.
     pub last_revealed: BTreeMap<DescriptorId, u32>,
--- a/crates/chain/tests/test_tx_graph.rs
+++ b/crates/chain/tests/test_tx_graph.rs
@@ -78,12 +78,11 @@ fn insert_txouts() {
     let mut graph = {
         let mut graph = TxGraph::<BlockId>::default();
         for (outpoint, txout) in &original_ops {
+            let mut expected_changeset = ChangeSet::default();
+            expected_changeset.txouts.insert(*outpoint, txout.clone());
             assert_eq!(
                 graph.insert_txout(*outpoint, txout.clone()),
-                ChangeSet {
-                    txouts: [(*outpoint, txout.clone())].into(),
-                    ..Default::default()
-                }
+                expected_changeset
             );
         }
         graph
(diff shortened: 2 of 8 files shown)


evanlinjin:
The persistence test suite will make this work. It's unfortunate we didn't think about that solution earlier.

notmandatory:
I fixed the tests, I don't think they look too bad with the non-exhaustive change sets.

tnull (commented):
The `wallet::ChangeSet` struct used to be `non_exhaustive`, which was reverted after I had argued in https://github.com/bitcoindevkit/bdk/issues/1591 that it shouldn't be. 

Why do we now think the sub-changesets should be `non_exhaustive`? To restate my point: having your code break is a good thing if you're handling the changeset types manually, because then you're forced to update your persistence. If you force a non-breaking API via `non_exhaustive`, users could miss persisting important fields after they had been added.

notmandatory:
Ah ok I completely forgot we discussed this last year. It seems we have conflicting goals:
* I want to be able to add new, default-able fields in persisted data as non-breaking changes so we don't need to do a new major release every time we add a feature that requires persisting new data. 
* You want existing persistence code to break when new fields are added to persisted data so all implementations are forced to stay up-to-date. 

I think a part of the problem is we're trying to support a persistence API that is tightly coupled to our ChangeSet structs. I feel like there should be a way to support both goals. Possibly with some sort of persistence store traits or persistence conformance test suite, maybe both. But I'll do some more thinking on it, and open to any other ideas.

tnull:
> * I want to be able to add new, default-able fields in persisted data as non-breaking changes so we don't need to do a new major release every time we add a feature that requires persisting new data.

`ChangeSet` in part of the public API. If you add new data fields that a user is required to persist, it's by definition a breaking change, no?

>  Possibly with some sort of persistence store traits or persistence conformance test suite

What you describe here is a BDK-internal data model/serialization logic, which was decided against some time ago. But yeah, if that is the conclusion, I'd be very happy about that outcome (especially if it considers forwards compat.). ;)

crates/chain/tests/test_tx_graph.rs
```diff
@@ -1470,39 +1451,27 @@ fn tx_graph_update_conversion() {
...
     changeset = graph.insert_seen_at(hash!("tx1"), later_seen_at);
-    assert_eq!(
-        changeset,
-        ChangeSet {
-            last_seen: [(hash!("tx1"), 1000001)].into(),
-            ..Default::default()
-        }
-    );
+    let mut expected_changeset = ChangeSet::default();
+    expected_changeset.last_seen = [(hash!("tx1"), 1000001)].into();
```
sdmg15, line 1473: Nit: Since we're formatting the number, maybe we could do that too here.

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -1047,6 +1047,7 @@ impl<K: core::fmt::Display + core::fmt::Debug> core::error::Error for InsertDesc
 #[derive(Clone, Debug, Default, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[must_use]
+#[non_exhaustive]
```
sdmg15, line 1050: Will it be interesting to either through doc comment or inside the book to add about the rational here?

sdmg15 (commented):
tAck.

`cargo test --all-targets`

Image

oleonardolima (commented):
cACK dd97143463111b49126b4e6100f5b80cb323df11

evanlinjin:
After a discussion with @oleonardolima and @ValuedMammal, we concluded that there is no way to introduce this feature without introducing dire footguns for the caller (even if we have #2012 in place).

Image the following situation:

* Latest `bdk_chain` version is `0.25.0`.
* Caller is using a persistence crate `bdk_mongodb = 0.1` which uses persistence tests from `bdk_chain = 0.25.0`.
* Caller has `bdk_chain = 0.25` in their Cargo (patch version will be bumped automatically).
* We introduce new changeset fields in `bdk_chain = 0.25.1` and release it.
* Maintainer of `bdk_mongodb` does not update persistence tests to use the version in `bdk_chain = 0.25.1`.
* Caller does `cargo update` in their project and changesets get new fields. `bdk_mongodb` ignores those new fields - potentially loosing data.

The benefits of this does not outweigh this massive footgun. We rarely introduce new changeset fields anyway.

notmandatory:
Thanks for the detailed thought on this. I was worried we'd have too many changes on this struct but in practice it seems to be going OK with us making breaking version bump when we need to add fields to it. We should still try to limit changes as much as possible.

tnull:
> * We introduce new changeset fields in `bdk_chain = 0.25.1` and release it.

Hmm, well, isn't `ChangeSet` part of the public API, so you can't/shouldn't change it on a patch release anyways?

PR #1981, opened 2025-06-25
refactor(electrum): remove `unwrap()`s and `expect()`s

### LINKED ISSUE #30: Never panic based on remote-provided inputs
**Describe the bug**  
I just hit a panic in `bdk_esplora` as I misconfigured the URL for the Esplora endpoint to a server on a different network (i.e., left it untouched when switching networks from `mainnet` to `signet`). This results in in hitting this `expect`:

https://github.com/bitcoindevkit/bdk/blob/88330f603cb415d01c88f1a579f20a21cb8c1658/crates/esplora/src/async_ext.rs#L240

While this is just a minor inconvenience, it surfaces a much larger problem: from a quick look, there are many cases of `expect` and `unwrap` in the chain-syncing crates (i.e., `bdk_esplora`,`bdk_electrum`), and at least a few of them seem to be directly or indirectly based on remote-provided inputs that aren't validated otherwise.

This should never happen, essentially crates dealing with remote-provided data should *never* `expect`  or `unwrap`. Given that all of these methods return `Result<.., Error>` there really is no reason why they couldn't just return errors in these cases.

**To Reproduce**  
For example, switch networks without switching Esplora URLs.

**Expected behavior**  
Never panic based on remote-provided inputs

LagginTimes:
Partially resolves https://github.com/bitcoindevkit/bdk_wallet/issues/30.

### Description

This PR eliminates all `unwrap()` and `expect()` calls from `bdk_electrum_client`, replacing them with proper error handling. Given that all public methods already return `Result`, we now propagate error messages instead of panicking.

### Changelog notice

* Removed all `unwrap()`s and `expect()`s from `bdk_electrum_client.rs`.

#### All Submissions:

### DIFF
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -647,25 +647,19 @@ fn fetch_tip_and_latest_blocks(
             }
         }
         agreement_cp
+            .ok_or_else(|| Error::Message("cannot find agreement block with server".to_string()))?
     };
 
-    let agreement_height = agreement_cp.as_ref().map(CheckPoint::height);
-
-    let new_tip = new_blocks
+    let extension = new_blocks
         .iter()
-        // Prune `new_blocks` to only include blocks that are actually new.
-        .filter(|(height, _)| Some(*<&u32>::clone(height)) > agreement_height)
-        .map(|(height, hash)| BlockId {
-            height: *height,
-            hash: *hash,
+        .filter({
+            let agreement_height = agreement_cp.height();
+            move |(height, _)| **height > agreement_height
         })
-        .fold(agreement_cp, |prev_cp, block| {
-            Some(match prev_cp {
-                Some(cp) => cp.push(block).ok()?,
-                None => CheckPoint::new(block),
-            })
-        })
-        .ok_or_else(|| Error::Message("failed to construct new checkpoint tip".to_string()))?;
+        .map(|(&height, &hash)| BlockId { height, hash });
+    let new_tip = agreement_cp
+        .extend(extension)
+        .expect("extension heights already checked to be greater than agreement height");
 
     Ok((new_tip, new_blocks))
 }
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -690,9 +690,11 @@ fn chain_update(
 #[cfg(test)]
 mod test {
     use crate::{bdk_electrum_client::TxUpdate, BdkElectrumClient};
-    use bdk_chain::bitcoin::{OutPoint, Transaction, TxIn};
-    use bdk_core::collections::BTreeMap;
-    use bdk_testenv::{utils::new_tx, TestEnv};
+    use bdk_chain::bitcoin::{constants, Network, OutPoint, ScriptBuf, Transaction, TxIn};
+    use bdk_chain::{BlockId, CheckPoint};
+    use bdk_core::{collections::BTreeMap, spk_client::SyncRequest};
+    use bdk_testenv::{anyhow, utils::new_tx, TestEnv};
+    use electrum_client::Error as ElectrumError;
     use std::sync::Arc;
 
     #[cfg(feature = "default")]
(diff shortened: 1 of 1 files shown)


crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -36,20 +36,32 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
-    pub fn populate_tx_cache(&self, txs: impl IntoIterator<Item = impl Into<Arc<Transaction>>>) {
-        let mut tx_cache = self.tx_cache.lock().unwrap();
+    pub fn populate_tx_cache(
+        &self,
+        txs: impl IntoIterator<Item = impl Into<Arc<Transaction>>>,
+    ) -> Result<(), Error> {
+        let mut tx_cache = self
+            .tx_cache
+            .lock()
+            .map_err(|e| Error::Message(format!("tx_cache lock poisoned: {e}")))?;
```
tnull: IMO, lock poisoning is an exception to 'never unwrap', as if we fail to acquire a `Mutex` something is so fundamentally wrong that we can't resume anyways, and might as well panic.

tnull (commented):
Thank you for looking into this! One comment.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -475,8 +473,11 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
                     let outpoint = vin.previous_output;
                     let vout = outpoint.vout;
                     let prev_tx = self.fetch_tx(outpoint.txid)?;
-                    let txout = prev_tx.output[vout as usize].clone();
-                    let _ = tx_update.txouts.insert(outpoint, txout);
+                    // Only insert txout if it exists. This avoids panicking on malformed or
+                    // incomplete transactions.
+                    if let Some(txout) = prev_tx.output.get(vout as usize) {
+                        let _ = tx_update.txouts.insert(outpoint, txout.clone());
+                    }
```
evanlinjin: Since this is a technically impossible situation it may be better to catch it.

I agree that a panic won't be appropriate here as it is more than likely a server-side issue. It seems that an error would be most appropriate?

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -579,8 +579,18 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
-                    let txout = prev_tx.output[vout as usize].clone();
-                    let _ = tx_update.txouts.insert(outpoint, txout);
+                    // Ensure server returns the expected txout.
+                    let txout = prev_tx
+                        .output
+                        .get(vout as usize)
+                        .ok_or_else(|| {
+                            electrum_client::Error::Message(format!(
+                                "server returned tx {} with no vout",
+                                prev_tx.compute_txid()
```
```suggestion
                                "prevout {outpoint} does not exist"
```

Not sure what "no vout" refers to here.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -652,11 +662,11 @@ fn fetch_tip_and_latest_blocks(
         })
         .fold(agreement_cp, |prev_cp, block| {
             Some(match prev_cp {
-                Some(cp) => cp.push(block).expect("must extend checkpoint"),
+                Some(cp) => cp.push(block).ok()?,
```
evanlinjin: `.push` would never fail because `push` only errors "if the block you are pushing on is not at a greater height than the one you are pushing to" and this would never happen because:

* We use `agreement_point` as the base.
* We filter out anything not greater than `agreement_point`.
* `new_blocks` must be in sequentially increasing order.

Instead of changing this to something that doesn't panic, I suggest having a more detailed `expect` with supplementary comments.

evanlinjin:
I created a commit on top which cleans up some of the logic so that we return a more explicit error when we cannot find a point of agreement (a.k.a. wrong network): 7c1cc8033a183f5d44b0be1cac1069ce710064bb. Feel free to cherry-pick this across if you find that it is appropriate.

I think it would make sense to include a sync-with-wrong-network test to make sure we hit this error.

evanlinjin (approved):
ACK b24ae6d71815e268417034fda44f2d57325f6166

PR #1985, opened 2025-07-01
fix(bitcoind_rpc): properly handle reorgs in `FilterIter`

### LINKED ISSUE #1848: `FilterIter` may not handle reorgs properly.

4. Call `.next` again.

_Originally posted by @evanlinjin in https://github.com/bitcoindevkit/bdk/issues/1614#issuecomment-2672023986_

LagginTimes:
Replaces #1909.
Fixes #1848.

### Description

This patch enhances `FilterIter` to robustly detect and handle chain reorgs. Reorgs are detected by checking if the block hash at the current height has changed, or if the block's `prev_blockhash` no longer matches the hash of the block at the previous height. This allows detection of reorgs that occur either at the current height or one block prior.

When a reorg is detected, the iterator rewinds up to `MAX_REORG_DEPTH` blocks to locate the most recent common ancestor. It then resets its internal state and resumes scanning from the fork point on the new chain.

This update also includes tests that simulate reorgs — including back-to-back reorgs at the same height — and verifies that `FilterIter` correctly switches to the new chain and yields accurate events.

### Changelog notice

* Improve `FilterIter` to detect and handle reorgs at the current height or its parent.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bitcoind_rpc/src/bip158.rs
+++ b/crates/bitcoind_rpc/src/bip158.rs
@@ -6,7 +6,7 @@
 //! [0]: https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki
 //! [1]: https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki
 
-use bdk_core::collections::BTreeMap;
+use bdk_core::collections::{BTreeMap, BTreeSet};
 use core::fmt;
 
 use bdk_core::bitcoin;
@@ -33,6 +33,10 @@ pub struct FilterIter<'c, C> {
     cp: Option<CheckPoint>,
     // blocks map
     blocks: BTreeMap<Height, BlockHash>,
+    // set of heights with filters that matched any watched SPK
+    matched: BTreeSet<Height>,
+    // initial height
+    start: Height,
     // best height counter
     height: Height,
     // stop height
@@ -40,6 +44,9 @@ pub struct FilterIter<'c, C> {
 }
 
 impl<'c, C: RpcApi> FilterIter<'c, C> {
+    /// Hard cap on how far to walk back when a reorg is detected.
+    const MAX_REORG_DEPTH: u32 = 100;
+
     /// Construct [`FilterIter`] from a given `client` and start `height`.
     pub fn new_with_height(client: &'c C, height: u32) -> Self {
         Self {
@@ -47,6 +54,8 @@ impl<'c, C: RpcApi> FilterIter<'c, C> {
             spks: vec![],
             cp: None,
             blocks: BTreeMap::new(),
+            matched: BTreeSet::new(),
+            start: height,
             height,
             stop: 0,
         }
@@ -69,57 +78,28 @@ impl<'c, C: RpcApi> FilterIter<'c, C> {
         self.spks.push(spk);
     }
 
-    /// Get the next filter and increment the current best height.
-    ///
-    /// Returns `Ok(None)` when the stop height is exceeded.
-    fn next_filter(&mut self) -> Result<Option<NextFilter>, Error> {
-        if self.height > self.stop {
-            return Ok(None);
-        }
-        let height = self.height;
-        let hash = match self.blocks.get(&height) {
-            Some(h) => *h,
-            None => self.client.get_block_hash(height as u64)?,
-        };
-        let filter_bytes = self.client.get_block_filter(&hash)?.filter;
-        let filter = BlockFilter::new(&filter_bytes);
-        self.height += 1;
-        Ok(Some((BlockId { height, hash }, filter)))
+    /// Get the block hash by `height` if it is found in the blocks map.
+    fn get_block_hash(&self, height: &Height) -> Option<BlockHash> {
+        self.blocks.get(height).copied()
     }
 
     /// Get the remote tip.
     ///
-    /// Returns `None` if the remote height is not strictly greater than the height of this
-    /// [`FilterIter`].
+    /// Returns `None` if the remote height is less than the height of this [`FilterIter`].
     pub fn get_tip(&mut self) -> Result<Option<BlockId>, Error> {
         let tip_hash = self.client.get_best_block_hash()?;
-        let mut header = self.client.get_block_header_info(&tip_hash)?;
+        let header = self.client.get_block_header_info(&tip_hash)?;
         let tip_height = header.height as u32;
-        if self.height >= tip_height {
+        // Allow returning tip if we're exactly at it. Return `None`` if we've already scanned past.
+        if self.height > tip_height {
             // nothing to do
             return Ok(None);
         }
-        self.blocks.insert(tip_height, tip_hash);
 
-        // if we have a checkpoint we use a lookback of ten blocks
-        // to ensure consistency of the local chain
+        // start scanning from point of agreement + 1
         if let Some(cp) = self.cp.as_ref() {
-            // adjust start height to point of agreement + 1
             let base = self.find_base_with(cp.clone())?;
-            self.height = base.height + 1;
-
-            for _ in 0..9 {
-                let hash = match header.previous_block_hash {
-                    Some(hash) => hash,
-                    None => break,
-                };
-                header = self.client.get_block_header_info(&hash)?;
-                let height = header.height as u32;
-                if height < self.height {
-                    break;
-                }
-                self.blocks.insert(height, hash);
-            }
+            self.height = base.height.saturating_add(1);
         }
 
         self.stop = tip_height;
@@ -130,9 +110,6 @@ impl<'c, C: RpcApi> FilterIter<'c, C> {
         }))
     }
 }
-
-/// Alias for a compact filter and associated block id.
-type NextFilter = (BlockId, BlockFilter);
 
 /// Event inner type
 #[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -69,22 +78,14 @@ impl<'c, C: RpcApi> FilterIter<'c, C> {
...
-        self.height += 1;
-        Ok(Some((BlockId { height, hash }, filter)))
+    /// Get the block hash by `height` if it is found in the blocks map.
+    fn get_block_hash(&self, height: &Height) -> Option<BlockHash> {
+        self.blocks.get(height).copied()
+    }
+
+    /// Insert a (non-matching) block height and hash into the blocks map.
+    fn insert_block(&mut self, height: Height, hash: BlockHash) {
+        self.blocks.insert(height, hash);
```
ValuedMammal: 🔧  We can get rid of the helpers `insert_block` and `get_block_hash` and simply call this code inline.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -93,33 +94,18 @@ impl<'c, C: RpcApi> FilterIter<'c, C> {
     /// [`FilterIter`].
     pub fn get_tip(&mut self) -> Result<Option<BlockId>, Error> {
         let tip_hash = self.client.get_best_block_hash()?;
-        let mut header = self.client.get_block_header_info(&tip_hash)?;
+        let header = self.client.get_block_header_info(&tip_hash)?;
         let tip_height = header.height as u32;
-        if self.height >= tip_height {
+        // Allow returning tip if we're exactly at it. Return `None`` if we've already scanned past.
+        if self.height > tip_height {
```
ValuedMammal, line 91: 📌  Since this is a logical change I think the documentation for `get_tip` should be reworded slightly.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -171,27 +154,84 @@ impl<C: RpcApi> Iterator for FilterIter<'_, C> {
...
-                    self.blocks.insert(height, hash);
-                    let inner = EventInner { height, block };
-                    Ok(Some(Event::Block(inner)))
-                } else {
-                    Ok(Some(Event::NoMatch(height)))
+        (|| -> Result<Option<_>, Error> {
+            if self.height > self.stop {
+                return Ok(None);
+            }
+            // Fetch next filter.
```
ValuedMammal:
```suggestion
            // Fetch next header.
```
<hr>

📌 In the future `FilterIter` should store and return the headers, that way the caller can collect a chain update into a `LocalChain<Header>`.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -171,27 +154,84 @@ impl<C: RpcApi> Iterator for FilterIter<'_, C> {
...
+            let mut height = self.height;
+            let mut hash = self.client.get_block_hash(height as _)?;
+            let mut header = self.client.get_block_header(&hash)?;
+
+            // Detect and resolve reorgs: either block at height changed, or its parent changed.
+            let stored_hash = self.blocks.get(&height).copied();
+            let prev_hash = self.blocks.get(&(height - 1)).copied();
```
ValuedMammal: 🔧 This is to avoid calling `get` with the underflow height in case `height` is 0.

```suggestion
            let prev_hash = height
                .checked_sub(1)
                .and_then(|height| self.blocks.get(&height).copied());
```

crates/bitcoind_rpc/tests/test_filter_iter.rs
```diff
@@ -163,3 +163,421 @@ fn filter_iter_error_no_scripts() -> anyhow::Result<()> {
...
+    println!("STEP: Iterating once (original block A)");
+    let event_a = iter.next().expect("Iterator should have item A")?;
+    // println!("First event: {:?}", event_a);
+    match event_a {
+        Event::Block(EventInner { height, block }) => {
+            assert_eq!(height, 104);
+            assert_eq!(block.block_hash(), hash_104);
+            assert!(block.txdata.iter().any(|tx| tx.compute_txid() == txid_a));
+        }
+        _ => panic!("Expected relevant tx at block A 102"),
```
ValuedMammal:
```suggestion
        _ => panic!("Expected relevant tx at block A 104"),
```

crates/bitcoind_rpc/tests/test_filter_iter.rs
```diff
@@ -163,3 +163,421 @@ fn filter_iter_error_no_scripts() -> anyhow::Result<()> {
...
+        }
+    }
+
+    // Check chain update tip
+    // println!("STEP: Checking chain_update");
+    let final_update = iter.chain_update();
+    assert!(
+        final_update.is_none(),
+        "We didn't instantiate FilterIter with a checkpoint"
+    );
```
ValuedMammal: 📌 It will be better to instantiate `FilterIter` with a checkpoint and check the blocks in the returned chain update as well.

ValuedMammal (commented):
Thanks for picking this up @LagginTimes

ValuedMammal (changes requested):
It's missing a test for max reorg depth.

LagginTimes:
> It's missing a test for max reorg depth.

The test for this is still WIP but I pushed out the changes that resolved the other open comments.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -171,27 +148,86 @@ impl<C: RpcApi> Iterator for FilterIter<'_, C> {
...
+                        }
+                    }
+
+                    reorg_depth += 1;
                 }
-            })
+
+                // Update self.height so we reprocess this height
+                self.height = height;
+            }
```
ValuedMammal:
```diff
             // Reorg detected, rewind to last known-good ancestor.
             if reorg_detected {
                 self.blocks.split_off(&height);
-                self.matched = self.matched.split_off(&height);
+                self.matched.split_off(&height);
 
                 let mut reorg_depth = 0;
                 loop {
                     if reorg_depth >= Self::MAX_REORG_DEPTH || height == 0 {
                         return Err(Error::ReorgDepthExceeded);
                     }
 
                     height = height.saturating_sub(1);
                     hash = self.client.get_block_hash(height as _)?;
                     header = self.client.get_block_header(&hash)?;
 
                     let prev_height = height.saturating_sub(1);
-                    let prev_hash = self.client.get_block_hash(prev_height as _)?;
-                    self.blocks.insert(prev_height, prev_hash);
-
                     if let Some(prev_hash) = self.blocks.get(&prev_height) {
                         if header.prev_blockhash == *prev_hash {
                             break;
                         }
                     }
 
                     reorg_depth += 1;
                 }
-
-                // Update self.height so we reprocess this height
-                self.height = height;
             }

```

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -171,27 +148,86 @@ impl<C: RpcApi> Iterator for FilterIter<'_, C> {
...
+            } else if let Some(expected_prev) = prev_hash {
+                header.prev_blockhash != expected_prev
+            } else {
+                false
+            };
+
+            // Reorg detected, rewind to last known-good ancestor.
+            if reorg_detected {
+                self.blocks.split_off(&height);
+                self.matched = self.matched.split_off(&height);
```
ValuedMammal: Looks like you're using `split_off` in two different ways. If I understand the logic, I think we want to drop the return value, like you have on L178.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -171,27 +148,86 @@ impl<C: RpcApi> Iterator for FilterIter<'_, C> {
...
+                        return Err(Error::ReorgDepthExceeded);
+                    }
+
+                    height = height.saturating_sub(1);
+                    hash = self.client.get_block_hash(height as _)?;
+                    header = self.client.get_block_header(&hash)?;
+
+                    let prev_height = height.saturating_sub(1);
+                    let prev_hash = self.client.get_block_hash(prev_height as _)?;
+                    self.blocks.insert(prev_height, prev_hash);
```
ValuedMammal: This looks problematic because we do the `header.prev_blockhash` check immediately following, which will most likely short-circuit the loop rather than rewind back to the last known good ancestor.

crates/bitcoind_rpc/src/bip158.rs
ValuedMammal: Maybe move the `split_off` until after we've found the rewind height - so we don't hold on to any stale entries.

ValuedMammal (dismissed):
ACK c2ae4da2ca59ee05e14a99ec7fba0297bd461695

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -69,57 +78,28 @@ impl<'c, C: RpcApi> FilterIter<'c, C> {
...
+    fn get_block_hash(&self, height: &Height) -> Option<BlockHash> {
+        self.blocks.get(height).copied()
     }
 
     /// Get the remote tip.
     ///
-    /// Returns `None` if the remote height is not strictly greater than the height of this
-    /// [`FilterIter`].
+    /// Returns `None` if the remote height is less than the height of this [`FilterIter`].
     pub fn get_tip(&mut self) -> Result<Option<BlockId>, Error> {
```
evanlinjin, line 86: Maybe a little out of scope for this PR, but the documentation here is a little lacking.

`get_tip`, as I understand, serves the following purposes:

1. Initialize the `FilerIter`. Without calling it, `FilterIter` returns nothing.
2. If it returns `None`, it means we are up to date with remote.

Is this the case?

If so, purpose `2.` is not satisfied because reorgs are not detected by `get_tip`.

@ValuedMammal are you able to provide some clarity, thanks!

ValuedMammal: `get_tip` is used to set the start and stop height of a scan and is only necessary to call once per instance of FilterIter. During review I was going to mention that I think `get_tip` should actually clear `self.blocks`, because I didn't anticipate the user interleaving calls to `get_tip` and `next`.

evanlinjin: @ValuedMammal Thank you for the update.

The internal implementation of `FilterIter` implies that it is a single-use structure (construct once, call `.next` till end). However, the API design of it implies something different.

What do you think about changing the API to properly represent a "single-use" workflow? I.e. I would change the following:

* Remove or unexpose `add_spk{s}`.
* Change constructor methods to also take in spks. Return `None` if there are no spks provided.
* Make sure `get_tip` logic is run on the first call to `next`.
* Rename `get_tip` to `init` which would return the target height.

(Separate PR of course)

Let me know what you think.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -33,20 +33,29 @@ pub struct FilterIter<'c, C> {
...
+    start: Height,
     // best height counter
     height: Height,
     // stop height
     stop: Height,
 }
 
 impl<'c, C: RpcApi> FilterIter<'c, C> {
+    /// Hard cap on how far to walk back when a reorg is detected.
+    const MAX_REORG_DEPTH: u32 = 100;
```
evanlinjin: What is the rationale for this check? Are there any downsides of not having this?

100-block reorgs are definitely possible (I.e. black swan events such as WW3). If it's the wrong network, we will emit the genesis block making it impossible to apply to the wallet anyway.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -171,27 +148,80 @@ impl<C: RpcApi> Iterator for FilterIter<'_, C> {
...
+                self.matched.split_off(&height);
+            }
+
+            let filter_bytes = self.client.get_block_filter(&hash)?.filter;
+            let filter = BlockFilter::new(&filter_bytes);
+
+            // record the scanned block
+            self.blocks.insert(height, hash);
+            // increment best height
+            self.height = height.saturating_add(1);
```
evanlinjin: Ideally, we want to deal with errors (esp. IO errors) before we mutate internal state. Otherwise, we may miss a block purely because we dropped connection for one second.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -259,6 +301,7 @@ impl fmt::Display for Error {
             Self::Bip158(e) => e.fmt(f),
             Self::NoScripts => write!(f, "no script pubkeys were provided to match with"),
             Self::Rpc(e) => e.fmt(f),
+            Self::ReorgDepthExceeded => write!(f, "maximum reorg depth exceeded"),
```
evanlinjin: Note that this change is breaking (commit message should have `!`). However, I'm proposing to remove this.

evanlinjin (changes requested):
Thanks for taking this forward.

I addressed my own review comments in commit 30e25d17f8c847039b9412338594f580947f1cba of [`evanlinjin:bdk @ filter_iter`](https://github.com/evanlinjin/bdk/tree/filter_iter). Let me know what you think. Check commit message for rationale.

crates/bitcoind_rpc/src/bip158.rs
LagginTimes: I like this idea, and can make a follow up PR to implement this.

ValuedMammal (commented):
I'm not convinced that this represents an objective improvement over Musab's PR.

evanlinjin:
> I'm not convinced that this represents an objective improvement over Musab's PR.

@ValuedMammal are you able to provide more context? I'm assuming you are referring to changes in 071fe4034cf45cbaf92eb55921634456099a7f1c?

Here is my justification why these changes are superior:

* **Separation of mutating/non-mutating code -** No state mutations made if anything errors - without this, we may miss blocks if network drops. Better readability.
* **No duplicated logic -** We only need to check whether `prev_blockhash` matches in one place (rather than two places).
* **Remove `MAX_REORG_DEPTH` -** As `MAX_REORG_DEPTH` does not provide any additional security and 100 block reorgs are possible. We are not checking PoW anyway; a malicious node can freely waste our time/return gibberish.

crates/bitcoind_rpc/src/bip158.rs
notmandatory: I agree it makes sense to make the API simpler/safer by making it a one time use struct that takes spks in the constructor. @LagginTimes please open a new issue or PR so we can continue exact implementation changes there.

LagginTimes:
~~Replaced by #2000.~~

Edit:
Re-opened for further discussion alongside #2000.

evanlinjin:
@LagginTimes @ValuedMammal There is no mention of a rationale on why this is closed and replaced by #2000 which is confusing. Let's not do that in the future.

crates/bitcoind_rpc/src/bip158.rs
ValuedMammal: On the other hand I think it would make sense to move the logic of `find_base_with` to happen in the constructor, then `get_tip` will only fetch the remote tip and hence could be called multiple times. Just an idea.

Also +1 for single-use workflow.

ValuedMammal:
When I ran the FilterIter example my canonical txs showed up, but strangely they appeared unconfirmed. I guess because nothing is being retained in `.matched`.

Further discussions surfaced more fundamental issues with the bip158 module which I started to address in #2000, namely the reliability of the internal checkpoint logic.

> `examples/filter_iter.rs`

```rs
    for canon_tx in graph.graph().list_canonical_txs(
        &chain,
        chain.tip().block_id(),
        bdk_chain::CanonicalizationParams::default(),
    ) {
        if !canon_tx.chain_position.is_confirmed() {
            eprintln!(
                "ERROR: canonical txid should be confirmed {}",
                canon_tx.tx_node.txid
            );
        }
    }
```

evanlinjin:
@ValuedMammal would you like to be responsible for finishing this work in #2000 and we can close this PR?

LagginTimes:
Replaced by #2000.

PR #1988, opened 2025-07-04
Detect new mempool txs

evanlinjin:
### Description

There is a bug in `bdk_bitcoind_rpc` where some new mempool transactions will not be emitted at all.

This problem exists because the avoid-re-emission logic depends on rounded-to-nearest-second timestamps.

The fix is to just emit all mempool transactions but wrap them in `Arc`s so that emission becomes cheap.

**Background:** I tried using `bdk_bitcoind_rpc` as the chain-source to write an example to showcase the [`IntentTracker`](https://github.com/bitcoindevkit/bdk_wallet/pull/257). However, `bdk_bitcoind_rpc` failed to emit some mempool transactions.

### Notes to the reviewers

The test added in c22c68f fails without these fixes.

Some tests are removed as they are no longer relevant.

### Changelog notice

```md
Fixed:
- Some mempool transactions not being emitted at all. The fix is to replace the avoid-re-emission-logic with one which emits all mempool transactions.
```

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/bitcoind_rpc/src/lib.rs
+++ b/crates/bitcoind_rpc/src/lib.rs
@@ -12,7 +12,11 @@
 use bdk_core::{BlockId, CheckPoint};
 use bitcoin::{Block, BlockHash, Transaction, Txid};
 use bitcoincore_rpc::{bitcoincore_rpc_json, RpcApi};
-use std::{collections::HashSet, ops::Deref};
+use std::{
+    collections::{HashMap, HashSet},
+    ops::Deref,
+    sync::Arc,
+};
 
 pub mod bip158;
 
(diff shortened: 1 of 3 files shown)


oleonardolima (commented):
cACK 
I think the new simplified approach looks better.

I'm still in the process of a deeper review. However, when I'm trying to run the `detect_new_mempool_txs`, on both commits, it takes more than 60 seconds and becomes stale, still unsure if it's a local issue.

evanlinjin:
@oleonardolima are you on Linux or OSX?

LagginTimes:
> cACK I think the new simplified approach looks better.
> 
> I'm still in the process of a deeper review. However, when I'm trying to run the `detect_new_mempool_txs`, on both commits, it takes more than 60 seconds and becomes stale, still unsure if it's a local issue.

FWIW, the `detect_new_mempool_txs` test runs successfully on my machine (Linux).

LagginTimes (approved):
Re-ACK 51ee99a39fbf9a6dc0ead964e794ca3bebd43ed0

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -115,110 +116,89 @@ where
...
+        let mut rpc_tip_height;
+        let mut rpc_tip_hash;
+        let mut rpc_mempool;
+        let mut rpc_mempool_txids;
+
+        // Ensure we get a mempool snapshot consistent with `rpc_tip_hash` as the tip.
+        loop {
+            rpc_tip_height = client.get_block_count()?;
+            rpc_tip_hash = client.get_block_hash(rpc_tip_height)?;
+            rpc_mempool = client.get_raw_mempool_verbose()?;
```
tnull: In our experience, we found polling via  `get_raw_mempool_verbose` constantly might be prohibitively expensive when the mempool is full and you 'download' a lot of data from the RPC each time. We since switched to an iterative approach that calls (non-verbose) `getrawmempool` and then only retrieving transaction data via `getmempoolentry` for the transactions that are missing from our in-memory cache (see https://github.com/lightningdevkit/ldk-node/pull/410 , although there have been further optimiziations after that).

Since you're doing the refactoring here already, I wounder if it would it make sense to do a similar change while you're at it.

oleonardolima:
> I'm still in the process of a deeper review. However, when I'm trying to run the `detect_new_mempool_txs`, on both commits, it takes more than 60 seconds and becomes stale, still unsure if it's a local issue.

> @oleonardolima are you on Linux or OSX?

It's OSX, but never mind, I ran it from a clean state with 'test-threads=2', and it worked just fine (probably had a concurrency or other problem with other tests I was running using bitcoind).

crates/bitcoind_rpc/src/lib.rs
evanlinjin: @tnull Thank you for the tip.

Changed to use `getrawmempool` without `verbose` in 7de3b71831947d1380c152286af3eb2ab2cdf019.

oleonardolima (approved):
tACK 73ab1eb06bcfaa0368a07b84a054bdd2f4f84f06

It looks good. I've tested in regtest with the bitcoind_rpc example, and it worked just fine.

nymius:
cACK 51ee99a39fbf9a6dc0ead964e794ca3bebd43ed0

PR #1993, opened 2025-07-14
keychain_txout: `apply_changeset` restores spk cache before last revealed

### LINKED ISSUE #1975: keychain_txout: `apply_changeset` should restore cache before applying last revealed
In order to reap the most benefit from the SPK cache, it seems necessary to refill the cache before applying the changes to `last_revealed`, however [we currently do things in the opposite order](https://github.com/bitcoindevkit/bdk/blob/542f9675c0716dfc5b4f3ab4e5320702e9f07a37/crates/chain/src/indexer/keychain_txout.rs#L939-L951). I noticed this while running performance tests locally.

ValuedMammal:
This patch improves readability and maintains logical consistency with the use of `spk_cache` throughout the `keychain_txout` module.

While it might offer a performance benefit, the results are mostly comparable with the current benchmarks as far as I can tell. At least there's no indication that it would negatively impact performance.

fixes #1975

#### All Submissions:


LagginTimes (approved):
ACK 6ddecc983ab3fb163303af4902bd10cf7e4a1fdf

PR #1997, opened 2025-07-20
fix(core): Memory leak bugs in `CheckPoint::drop` impl

evanlinjin:
### Description

* Fix memory leak bug in `CheckPoint::drop` by using `Arc::into_inner` if it is available (>= 1.70).
* Fix `CPInner::drop` logic so that if `CPInner::block` becomes generic and is of a type that required `drop`, it does not leak memory.
* Add tests for memory leak + stack overflow when dropping `CheckPoint`.

An alternative fix is to bump MSRV to >= 1.70 so we don't need to do conditional compilation based on version.

### Changelog notice

```md
Fix:
- Drop implementation of `CheckPoint` to avoid memory leaks for rust version >= 1.70.
```

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/core/build.rs
+++ b/crates/core/build.rs
@@ -1,0 +1,5 @@
+fn main() {
+    if version_check::is_min_version("1.70.0").unwrap_or(false) {
+        println!("cargo:rustc-cfg=has_arc_into_inner");
+    }
+}
--- a/crates/core/src/checkpoint.rs
+++ b/crates/core/src/checkpoint.rs
@@ -28,21 +28,24 @@ struct CPInner {
 /// https://github.com/bitcoindevkit/bdk/issues/1634
 impl Drop for CPInner {
     fn drop(&mut self) {
-        // Take out `prev` so its `drop` won't be called when this drop is finished
+        // Take out `prev` so its `drop` won't be called when this drop is finished.
         let mut current = self.prev.take();
+        // Collect nodes to drop later so we avoid recursive drop calls while not leaking memory.
         while let Some(arc_node) = current {
-            // Get rid of the Arc around `prev` if we're the only one holding a ref
-            // So the `drop` on it won't be called when the `Arc` is dropped.
+            // Get rid of the `Arc` around `prev` if we're the only one holding a reference so the
+            // `drop` on it won't be called when the `Arc` is dropped.
             //
-            // FIXME: When MSRV > 1.70.0 this should use Arc::into_inner which actually guarantees
-            // that no recursive drop calls can happen even with multiple threads.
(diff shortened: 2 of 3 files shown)


crates/core/src/checkpoint.rs
```diff
@@ -28,21 +28,24 @@ struct CPInner {
...
+            // `drop` on it won't be called when the `Arc` is dropped.
             //
-            // FIXME: When MSRV > 1.70.0 this should use Arc::into_inner which actually guarantees
-            // that no recursive drop calls can happen even with multiple threads.
-            match Arc::try_unwrap(arc_node).ok() {
-                Some(mut node) => {
-                    // Keep going backwards
-                    current = node.prev.take();
-                    // Don't call `drop` on `CPInner` since that risks it becoming recursive.
-                    core::mem::forget(node);
```
evanlinjin, line 44: **Rationale for removal:**

`core::mem::forget` is dangerous if `CPInner::block` is generic since the type might require a call to `drop`. This is not possible now, however it's best to remove it now to avoid forgetting about it later on.

The initial motivation for calling `forget` is to ensure no recursion happens. However, the recursive depth would always only be 1 since `prev` is `None`. Thus, there is no risk of stack overflow (proven by the `checkpoint_drop_is_not_recursive` test that is added).

LagginTimes (approved):
ACK 44018868de5da22577ad676cd3941718543a94a1

crates/core/build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    if version_check::is_min_version("1.70.0").unwrap_or(false) {
+        println!("cargo:rustc-cfg=has_arc_into_inner");
+    }
+}
```
ValuedMammal, line 5: IMO this is slightly overkill and we can wait until MSRV is at least 1.70.0 to use `Arc::into_inner`.

evanlinjin: Memory leaks aren't fun though.

tnull:
> An alternative fix is to bump MSRV to >= 1.70 so we don't need to do conditional compilation based on version.

FWIW, bumping MSRV to 1.75 would be fine by us, especially if it allows you guys to avoid these kinds of complications (also cf. https://github.com/bitcoindevkit/bdk/issues/1750)

evanlinjin:
Closing as #2055 is under way.

oleonardolima:
> Closing as #2055 is under way.

I opened a new PR #2056 with the fix based on this one, see: https://github.com/bitcoindevkit/bdk/pull/2056/commits/56c0bbb3489f6b5cd53c0ec0bc2112b4d92874d0

PR #2000, opened 2025-07-24
`FilterIter` API redesign

### LINKED ISSUE #1848: `FilterIter` may not handle reorgs properly.

ValuedMammal:
Previously `FilterIter` did not detect or handle reorgs between `next` calls, meaning that if a reorg occurred, we might process blocks from a stale fork potentially resulting in an invalid wallet state. This PR aims to fix that by adding logic to explicitly check for and respond to a reorg on every call to `next`.

### Notes to the reviewers

The old implementation required storing block IDs of scanned blocks before creating a checkpoint update, but because the interface was split across different methods, it introduced a timing risk between method calls which, when we consider the possibility of reorgs, made the implementation somewhat brittle.

To address this, we make sure that 1) Finding the start block and 2) Updating the internal checkpoint are directly tied to the logic of `next`. Since the checkpoint in practice is derived from a clone of the local chain, this ensures that the checkpoint returned by `next` can always find a connection point with the receiver. Additionally we now emit a checkpoint at every height to ensure that any "must-include" heights are not missing.

For example usage see `examples/filter_iter.rs`

fixes #1848 

### Changelog notice

    Fixed
    - `FilterIter` now handles reorgs to ensure consistency of the header chain.

    Changed
    - `Event` is now a struct instead of enum

    Added
    - `FilterIter::new` constructor that takes as input a reference to the RPC client, checkpoint, and a list of SPKs.
    - `Error::ReorgDepthExceeded` variant
    - `Error::TryFromInt` variant

    Removed
    - `FilterIter::new_with_height`
    - `FilterIter::new_with_checkpoint`
    - `EventInner` type
    - `FilterIter::get_tip`
    - `FilterIter::chain_update`
    - `Error::NoScripts` variant

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bitcoind_rpc/src/bip158.rs
+++ b/crates/bitcoind_rpc/src/bip158.rs
@@ -6,246 +6,164 @@
 //! [0]: https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki
 //! [1]: https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki
 
-use bdk_core::collections::BTreeMap;
-use core::fmt;
-
 use bdk_core::bitcoin;
 use bdk_core::{BlockId, CheckPoint};
-use bitcoin::{
-    bip158::{self, BlockFilter},
-    Block, BlockHash, ScriptBuf,
-};
+use bitcoin::{bip158::BlockFilter, Block, ScriptBuf};
 use bitcoincore_rpc;
-use bitcoincore_rpc::RpcApi;
+use bitcoincore_rpc::{json::GetBlockHeaderResult, RpcApi};
 
-/// Block height
-type Height = u32;
-
-/// Type that generates block [`Event`]s by matching a list of script pubkeys against a
-/// [`BlockFilter`].
+/// Type that returns Bitcoin blocks by matching a list of script pubkeys (SPKs) against a
+/// [`bip158::BlockFilter`].
 #[derive(Debug)]
-pub struct FilterIter<'c, C> {
-    // RPC client
-    client: &'c C,
-    // SPK inventory
+pub struct FilterIter<'a> {
+    /// RPC client
+    client: &'a bitcoincore_rpc::Client,
+    /// SPK inventory
     spks: Vec<ScriptBuf>,
-    // local cp
-    cp: Option<CheckPoint>,
-    // blocks map
-    blocks: BTreeMap<Height, BlockHash>,
-    // best height counter
-    height: Height,
-    // stop height
-    stop: Height,
+    /// checkpoint
+    cp: CheckPoint,
+    /// Header info, contains the prev and next hashes for each header.
+    header: Option<GetBlockHeaderResult>,
 }
 
-impl<'c, C: RpcApi> FilterIter<'c, C> {
-    /// Construct [`FilterIter`] from a given `client` and start `height`.
-    pub fn new_with_height(client: &'c C, height: u32) -> Self {
+impl<'a> FilterIter<'a> {
+    /// Construct [`FilterIter`] with checkpoint, RPC client and SPKs.
+    pub fn new(
+        client: &'a bitcoincore_rpc::Client,
+        cp: CheckPoint,
+        spks: impl IntoIterator<Item = ScriptBuf>,
+    ) -> Self {
         Self {
             client,
-            spks: vec![],
-            cp: None,
-            blocks: BTreeMap::new(),
-            height,
-            stop: 0,
+            spks: spks.into_iter().collect(),
+            cp,
+            header: None,
         }
     }
 
-    /// Construct [`FilterIter`] from a given `client` and [`CheckPoint`].
-    pub fn new_with_checkpoint(client: &'c C, cp: CheckPoint) -> Self {
-        let mut filter_iter = Self::new_with_height(client, cp.height());
-        filter_iter.cp = Some(cp);
-        filter_iter
-    }
-
-    /// Extends `self` with an iterator of spks.
-    pub fn add_spks(&mut self, spks: impl IntoIterator<Item = ScriptBuf>) {
-        self.spks.extend(spks)
-    }
-
-    /// Add spk to the list of spks to scan with.
-    pub fn add_spk(&mut self, spk: ScriptBuf) {
-        self.spks.push(spk);
-    }
-
-    /// Get the next filter and increment the current best height.
+    /// Return the agreement header with the remote node.
     ///
-    /// Returns `Ok(None)` when the stop height is exceeded.
-    fn next_filter(&mut self) -> Result<Option<NextFilter>, Error> {
-        if self.height > self.stop {
-            return Ok(None);
-        }
-        let height = self.height;
-        let hash = match self.blocks.get(&height) {
-            Some(h) => *h,
-            None => self.client.get_block_hash(height as u64)?,
-        };
-        let filter_bytes = self.client.get_block_filter(&hash)?.filter;
-        let filter = BlockFilter::new(&filter_bytes);
-        self.height += 1;
-        Ok(Some((BlockId { height, hash }, filter)))
-    }
-
-    /// Get the remote tip.
-    ///
-    /// Returns `None` if the remote height is not strictly greater than the height of this
-    /// [`FilterIter`].
-    pub fn get_tip(&mut self) -> Result<Option<BlockId>, Error> {
-        let tip_hash = self.client.get_best_block_hash()?;
-        let mut header = self.client.get_block_header_info(&tip_hash)?;
-        let tip_height = header.height as u32;
-        if self.height >= tip_height {
-            // nothing to do
-            return Ok(None);
-        }
-        self.blocks.insert(tip_height, tip_hash);
-
-        // if we have a checkpoint we use a lookback of ten blocks
-        // to ensure consistency of the local chain
-        if let Some(cp) = self.cp.as_ref() {
-            // adjust start height to point of agreement + 1
-            let base = self.find_base_with(cp.clone())?;
-            self.height = base.height + 1;
-
-            for _ in 0..9 {
-                let hash = match header.previous_block_hash {
-                    Some(hash) => hash,
-                    None => break,
-                };
-                header = self.client.get_block_header_info(&hash)?;
-                let height = header.height as u32;
-                if height < self.height {
-                    break;
-                }
-                self.blocks.insert(height, hash);
+    /// Error if no agreement header is found.
+    fn find_base(&self) -> Result<GetBlockHeaderResult, Error> {
+        for cp in self.cp.iter() {
+            match self.client.get_block_header_info(&cp.hash()) {
+                Err(e) if is_not_found(&e) => continue,
+                Ok(header) if header.confirmations <= 0 => continue,
+                Ok(header) => return Ok(header),
+                Err(e) => return Err(Error::Rpc(e)),
             }
         }
-
-        self.stop = tip_height;
-
-        Ok(Some(BlockId {
-            height: tip_height,
-            hash: tip_hash,
-        }))
+        Err(Error::ReorgDepthExceeded)
     }
 }
 
-/// Alias for a compact filter and associated block id.
-type NextFilter = (BlockId, BlockFilter);
-
-/// Event inner type
+/// Event returned by [`FilterIter`].
 #[derive(Debug, Clone)]
-pub struct EventInner {
-    /// Height
-    pub height: Height,
-    /// Block
-    pub block: Block,
-}
-
-/// Kind of event produced by [`FilterIter`].
-#[derive(Debug, Clone)]
-pub enum Event {
-    /// Block
-    Block(EventInner),
-    /// No match
-    NoMatch(Height),
+pub struct Event {
+    /// Checkpoint
+    pub cp: CheckPoint,
+    /// Block, will be `Some(..)` for matching blocks
+    pub block: Option<Block>,
 }
 
 impl Event {
     /// Whether this event contains a matching block.
     pub fn is_match(&self) -> bool {
-        matches!(self, Event::Block(_))
+        self.block.is_some()
     }
 
-    /// Get the height of this event.
-    pub fn height(&self) -> Height {
-        match self {
-            Self::Block(EventInner { height, .. }) => *height,
-            Self::NoMatch(h) => *h,
-        }
+    /// Return the height of the event.
+    pub fn height(&self) -> u32 {
+        self.cp.height()
     }
 }
 
-impl<C: RpcApi> Iterator for FilterIter<'_, C> {
+impl Iterator for FilterIter<'_> {
     type Item = Result<Event, Error>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        (|| -> Result<_, Error> {
-            // if the next filter matches any of our watched spks, get the block
-            // and return it, inserting relevant block ids along the way
-            self.next_filter()?.map_or(Ok(None), |(block, filter)| {
-                let height = block.height;
-                let hash = block.hash;
+        (|| -> Result<Option<_>, Error> {
+            let mut cp = self.cp.clone();
 
-                if self.spks.is_empty() {
-                    Err(Error::NoScripts)
-                } else if filter
-                    .match_any(&hash, self.spks.iter().map(|script| script.as_bytes()))
-                    .map_err(Error::Bip158)?
-                {
-                    let block = self.client.get_block(&hash)?;
-                    self.blocks.insert(height, hash);
-                    let inner = EventInner { height, block };
-                    Ok(Some(Event::Block(inner)))
-                } else {
-                    Ok(Some(Event::NoMatch(height)))
-                }
-            })
+            let header = match self.header.take() {
+                Some(header) => header,
+                // If no header is cached we need to locate a base of the local
+                // checkpoint from which the scan may proceed.
+                None => self.find_base()?,
+            };
+
+            let mut next_hash = match header.next_block_hash {
+                Some(hash) => hash,
+                None => return Ok(None),
+            };
+
+            let mut next_header = self.client.get_block_header_info(&next_hash)?;
+
+            // In case of a reorg, rewind by fetching headers of previous hashes until we find
+            // one with enough confirmations.
+            while next_header.confirmations < 0 {
+                let prev_hash = next_header
+                    .previous_block_hash
+                    .ok_or(Error::ReorgDepthExceeded)?;
+                let prev_header = self.client.get_block_header_info(&prev_hash)?;
+                next_header = prev_header;
+            }
+
+            next_hash = next_header.hash;
+            let next_height: u32 = next_header.height.try_into()?;
+
+            cp = cp.insert(BlockId {
+                height: next_height,
+                hash: next_hash,
+            });
+
+            let mut block = None;
+            let filter =
+                BlockFilter::new(self.client.get_block_filter(&next_hash)?.filter.as_slice());
+            if filter
+                .match_any(&next_hash, self.spks.iter().map(ScriptBuf::as_ref))
+                .map_err(Error::Bip158)?
+            {
+                block = Some(self.client.get_block(&next_hash)?);
+            }
+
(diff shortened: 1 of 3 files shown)


crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -31,23 +35,32 @@ pub struct FilterIter<'c, C> {
...
     height: Height,
+    // initial height
+    start: Height,
     // stop height
     stop: Height,
 }
 
 impl<'c, C: RpcApi> FilterIter<'c, C> {
+    /// Hard cap on how far to walk back when a reorg is detected.
+    const MAX_REORG_DEPTH: u32 = 100;
```
evanlinjin: I understand that the main motivation for this is to protect against attacks from a malicious/faulty node?

This is not going to work as BDK does not check PoW (not currently). The node can just cheaply construct multiple <100 block reorgs and exhaust resources that way.

The only way to protect against attacks is to check PoW.

Note that I've already mentioned this here: https://github.com/bitcoindevkit/bdk/pull/1985#issuecomment-3105650421

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -69,57 +82,21 @@ impl<'c, C: RpcApi> FilterIter<'c, C> {
...
-        self.height += 1;
-        Ok(Some((BlockId { height, hash }, filter)))
-    }
-
     /// Get the remote tip.
     ///
-    /// Returns `None` if the remote height is not strictly greater than the height of this
-    /// [`FilterIter`].
+    /// Returns `None` if the remote height is less than the height of this [`FilterIter`].
     pub fn get_tip(&mut self) -> Result<Option<BlockId>, Error> {
```
evanlinjin: If this call is required to "initiate" the iterator, we should mention it somewhere in the docs.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -202,36 +225,49 @@ impl<C: RpcApi> FilterIter<'_, C> {
...
+                // Ensure this block also exists in `self`.
+                self.headers.insert(height, (fetched_hash, header));
                 return Ok(cp.block_id());
             }
-            // remember conflicts
-            self.blocks.insert(height, fetched_hash);
-            cp = cp.prev().expect("must break before genesis");
+            // Remember conflicts.
+            self.headers.insert(height, (fetched_hash, header));
+            cp = cp.prev().ok_or(Error::ReorgDepthExceeded)?;
```
evanlinjin: Nit:

```suggestion
            // Ensure visited blocks also exist in `self`.
            self.headers.insert(height, (fetched_hash, header));

            if cp.hash() == fetched_hash {
                return Ok(cp.block_id());
            }
            cp = cp.prev().ok_or(Error::ReorgDepthExceeded)?;
```

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -202,36 +225,49 @@ impl<C: RpcApi> FilterIter<'_, C> {
...
+    pub fn chain_update(&self) -> Option<CheckPoint> {
+        if self.cp.is_none() || self.headers.is_empty() || self.height <= self.stop {
             return None;
         }
 
-        // note: to connect with the local chain we must guarantee that `self.blocks.first()`
-        // is also the point of agreement with `self.cp`.
+        // We return blocks up to and including the initial height, all of the matching blocks,
+        // and blocks in the terminal range.
+        let tail_range = self.stop.saturating_sub(9)..=self.stop;
```
evanlinjin: **Nit:** What do you think about extracting `9` out as a constant?

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -171,27 +150,71 @@ impl<C: RpcApi> Iterator for FilterIter<'_, C> {
...
-                    Ok(Some(Event::Block(inner)))
-                } else {
-                    Ok(Some(Event::NoMatch(height)))
+        (|| -> Result<Option<_>, Error> {
+            if self.height > self.stop {
+                return Ok(None);
+            }
+            // Fetch next header.
+            let mut height = self.height;
+            let mut hash = self.client.get_block_hash(height as u64)?;
```
evanlinjin: Nit: You can move this inside the loop to avoid duplicating code. Do `break (hash, header)` to access the value outside the loop.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -171,27 +150,71 @@ impl<C: RpcApi> Iterator for FilterIter<'_, C> {
...
+
+            let header = loop {
+                if reorg_depth >= Self::MAX_REORG_DEPTH {
+                    return Err(Error::ReorgDepthExceeded);
+                }
+
+                let header = self.client.get_block_header(&hash)?;
+
+                let prev_height = height.saturating_sub(1);
+                match self.headers.get(&prev_height).copied() {
```
evanlinjin: We do not want to continue looping once we reach genesis.

```suggestion
                let prev_cached_header = height
                    .checked_sub(1)
                    .and_then(|h| self.headers.get(&h))
                    .copied();
                match prev_cached_header {
```

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -171,27 +150,71 @@ impl<C: RpcApi> Iterator for FilterIter<'_, C> {
...
-            })
+            };
+
+            let filter_bytes = self.client.get_block_filter(&hash)?.filter;
+            let filter = BlockFilter::new(&filter_bytes);
+
+            // If the filter matches any of our watched SPKs, fetch the full
+            // block and prepare the next event.
+            let next_event = if self.spks.is_empty() {
+                Err(Error::NoScripts)
```
evanlinjin: Are you able to justify this error variant?

evanlinjin (changes requested):
I still need to understand the rationale of this PR (since we already had #1985 going). If you can expand on this, it would be appreciated.

I can see how the reorg-logic is simpler here (since it uses the pre-cached headers instead of having the check both the checkpoints/headers). Is this the main reasoning?

I'm assuming the main rationale for header-caching is so that it will be easier/faster to emit `CheckPoint`s with `Header`s once that becomes available. Is this correct?

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -21,42 +22,74 @@ use bitcoincore_rpc::RpcApi;
...
+    start: Height,
+    /// Stop height
     stop: Height,
 }
 
 impl<'c, C: RpcApi> FilterIter<'c, C> {
+    /// Hard cap on how far to walk back when a reorg is detected.
+    const MAX_REORG_DEPTH: u32 = 100;
+    /// Number of recent blocks from the tip to be returned in a chain update.
+    const CHAIN_SUFFIX_LEN: u32 = 10;
```
evanlinjin: Can we include a rationale for this?

From my understanding, this will make it faster to find the "point of agreement" if there is a reorg of depth < 10.

On second though, I'm not sure how impactful this is - reorgs are relatively rare and rescanning blocks should be pretty fast with filters?

evanlinjin (changes requested):
I've demonstrated that the chain-update construction logic is unsound with the following test: 21fa815a831f2908b120e4153825d643dd9f599d.

My intuition tells me that making the `FilterIter` checkpoint-centric would make it easier to ensure correctness here. To make this current approach work would mean copying everying in the `CheckPoint` into `FilterIter::headers` which would be counter-intuitive.

evanlinjin:
This is how I propose we move forward with #2000/#1985:

* Make the internals `CheckPoint`-centric. This will solve the chain-update-may-not-connect problem at it's core.
* Use `getblockheader (verbose = true)` to fetch blocks (instead of calling `getblockhash`+`getblockheader`). This reduces round trips and the `getblockheader (verbose = true)` result contains `confirmations`+`nextblockhash`+`previousblockhash` information which can be used reorg-detection (Note that `confirmations == -1` means the block is not in the best chain).
* Emit `CheckPoint` as part of the `Event` (for matched blocks). Only add on checkpoints for matched blocks and tip.
* Remove `stop`-height from internals. Stop when there is no `nextblockhash`.

### Reorg logic

Backtrack using `previousblockhash` until `confirmations >= 0`. Purge all checkpoints with a height >= current.

No need to cache multiple headers/blockhashes. Just keep track of the last emitted header/blockhash.

No need to have `Self::matched`. Matched checkpoints are emitted straight away.

evanlinjin:
To address the [claim](https://discord.com/channels/753336465005608961/1368973459551555605/1400621216754307136) that there is a bug in `LocalChain`:

I updated the test to print the update chain and the original chain (and also added block of height 50 to the original chain to show that it's not some sort of "genesis block problem"): 4f55f2fa9024f0555bf406368524c89f270f2476

* Heights of the original chain: `0, 50, 101`.
* Heights of the update chain: `101, 102`.

Note that the block hash of the block at height `101` are not the same between original and update. With this limited information, the checkpoint/localchain logic cannot connect the update to the original chain. In other words, how can we be certain that `50` and `102` exist on the same chain?

To have the update connect, the chain source should include `50`.

evanlinjin:
@ValuedMammal Sorry I pushed 4f55f2fa9024f0555bf406368524c89f270f2476 on the wrong remote! Feel free to get rid of it.

ValuedMammal:
You're right I take back the claim. I was thinking along the lines of a "transitive invalidation with no point of agreement", but since we traverse over height 50 of the original (which is still valid) the update is ambiguous.

> Emit `CheckPoint` as part of the `Event` (for matched blocks). Only add on checkpoints for matched blocks and tip.

Good idea - only thing is that the caller might eventually wish to collect the headers?

evanlinjin:
> Good idea - only thing is that the caller might eventually wish to collect the headers?

@ValuedMammal What would be the usecase of these headers? Would the caller want headers from every single block or just the relevant blocks?

ValuedMammal:
> @ValuedMammal What would be the usecase of these headers? Would the caller want headers from every single block or just the relevant blocks?

Still just the relevant ones. Maybe a better way to state the question is if `CheckPoint` becomes generic, the implementation needs to decide whether to emit checkpoints containing headers or just the classic checkpoint. Can probably be considered for future work.

evanlinjin:
@ValuedMammal yes I think it should be considered later down the line. Since the blockhash is "stable" per height, we can take in a closure for constructing any checkpoint type.

evanlinjin:
Concept ACK.

This is looking really good.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -6,262 +6,218 @@
...
             }
         }
 
-        self.stop = tip_height;
-
-        Ok(Some(BlockId {
-            height: tip_height,
-            hash: tip_hash,
-        }))
+        Err(Error::ReorgDepthExceeded)
```
evanlinjin: There is an unfortunate restriction with `LocalChain` (as it exists currently). The only safe way to ensure two `CheckPoint` chains connect is if the update chain contains all heights of the original chain from  the point of agreement. We probably need to keep track of these must-be-included-heights internally in the `FilterIter`.

We probably also need to change how we structure an `Event` since non-matching blocks which are not tips may need to return a `CheckPoint` update. I propose the following:

```rust
type SyncPoint {
    CheckPoint(CheckPoint),
    BlockId(BlockId),
} 

type Event {
    at: SyncPoint,
    match_block: Option<Block>,
}

// Suggestions for helper methods.
impl Event {
    pub fn checkpoint(&self) -> Option<CheckPoint> { /*TODO*/ }
    pub fn block_id(&self) -> BlockId { /*TODO*/ }
    pub fn height(&self) -> u32 { /*TODO*/ }
    pub fn is_match(&self) -> bool { /*TODO*/ }
}
```

Let me know what you think.

---

Additionally, what do you think about changing the `find_base` logic to call `get_block_header_info` against `cp.hash` (so no need for a separate `get_block_hash` call)? However, we may need to look at the RPC error code for if a block does not exist (instead of directly returning the error).

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -6,262 +6,218 @@
...
-        Some(
-            CheckPoint::from_block_ids(self.blocks.iter().map(BlockId::from))
-                .expect("blocks must be in order"),
-        )
+            // Purge any no longer valid checkpoints.
+            if reorg_ct.is_positive() {
+                cp = cp
+                    .range(..=next_height)
+                    .next()
+                    .ok_or(Error::ReorgDepthExceeded)?;
```
evanlinjin: We'll need to keep track of the heights we have removed to ensure the checkpoint update connects.

evanlinjin (changes requested):
Addresses some pain points in making `LocalChain` updates apply reliably.

P.S. Hopefully this highlights the value of the `BlockGraph` work you’ve been leading.

ValuedMammal:
Will it be easier to just include the checkpoint with every event?

```rs
/// Event returned by [`FilterIter`].
#[derive(Debug, Clone)]
pub struct Event {
    /// Checkpoint
    pub cp: CheckPoint,
    /// Block, will be `Some(..)` for matching blocks.
    pub block: Option<Block>,
}
```

ValuedMammal:
I added a commit ff92f306 and rebased.

evanlinjin:
> Will it be easier to just include the checkpoint with every event?
> 
> ```rs
> /// Event returned by [`FilterIter`].
> #[derive(Debug, Clone)]
> pub struct Event {
>     /// Checkpoint
>     pub cp: CheckPoint,
>     /// Block, will be `Some(..)` for matching blocks.
>     pub block: Option<Block>,
> }
> ```

@ValuedMammal yes, but the caller can't easily determine the height of non-matching events.

In which case, maybe we should only emit matched blocks + tip? This ensures that the `cp` and `block` of an event refers to the same block.

---

Edit: Sorry, I did not read the updated code. Looks like the `CheckPoint` update now includes all blocks.

ValuedMammal:
Indeed, it would be a bug if the event `cp` and `block` did not refer to the same block by hash.

evanlinjin:
@ValuedMammal This looks very good.

I'm wondering what you think about emitting all `CheckPoint`s vs only emitting relevant `CheckPoint`s. I'm assuming you are emitting all checkpoints right now for simplicity?

However, I also see the elegance with only emitting what is relevant.

---

Edit: Okay the simplicity of your implementation has really sold me. Let's assume that memory is cheap until it becomes a problem.

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -6,262 +6,182 @@
...
-/// Type that generates block [`Event`]s by matching a list of script pubkeys against a
-/// [`BlockFilter`].
+/// Type that returns Bitcoin blocks by matching a list of script pubkeys (SPKs) against a
+/// [`bip158::BlockFilter`].
 #[derive(Debug)]
-pub struct FilterIter<'c, C> {
-    // RPC client
-    client: &'c C,
-    // SPK inventory
+pub struct FilterIter<'a> {
```
evanlinjin, line 31: The `FilterIter` has an elegant behavior that isn’t currently documented:

when it reaches the tip it returns `None`, but will resume yielding `Some` as soon as the chain extends.

Let’s include this in the docs.

ValuedMammal:
> I'm assuming you are emitting all checkpoints right now for simplicity?

And it solves the issue of potentially missing but mandatory heights by placing the responsibility of forming a chain update on the caller. Refer back to https://github.com/bitcoindevkit/bdk/pull/2000#discussion_r2265233190.

evanlinjin (approved):
ACK ff92f306e0093aa0f45ff9bd66a08f14eb57dd3e

Just a small documentation nit.

crates/bitcoind_rpc/src/bip158.rs
ValuedMammal: Don't know if I described it very elegantly, but I added more struct level docs to `FilterIter`.

evanlinjin:
@ValuedMammal I added a 6-block reorg check to the test. Just to make sure. All passes.

evanlinjin (approved):
ACK 6aa81fe2ff5737add0be0664347a786573060706

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -6,262 +6,182 @@
...
-#[cfg(feature = "std")]
-impl std::error::Error for Error {}
+/// Whether the RPC error is a "not found" error (code: `-5`).
+fn is_not_found(e: &bitcoincore_rpc::Error) -> bool {
+    matches!(
+        e,
+        bitcoincore_rpc::Error::JsonRpc(bitcoincore_rpc::jsonrpc::Error::Rpc(e))
+        if e.code == -5
+    )
+}
```
oleonardolima, line 200: question: is this the same as -5 from [Bitcoin Core docs](https://github.com/bitcoin/bitcoin/blob/36e40417de3fcaa776fb7bb8ebf3a23db4266572/src/rpc/protocol.h#L42C5-L42C31) ? I wonder if we should use it as not found or use the same as `invalid_address_or_key` as used there.

oleonardolima (approved):
utACK 6aa81fe2ff5737add0be0664347a786573060706

It's looking good to me, I only left a question/nit about the `not_found` error code.

I'll give a try and test it locally.

oleonardolima (approved):
tACK 6aa81fe2ff5737add0be0664347a786573060706

I've tested it with the example in signet, and it worked just fine.

```
Matched block 205061
Matched block 208441
Matched block 219412
Matched block 228230
Matched block 235619
Matched block 237344
Matched block 250827
Matched block 251917
Matched block 262664
Matched block 266203
Matched block 268631

took: 111s
Local tip: 268631

Unspent
("external", 3) | 1.00104000 BTC | c214d5f31fd8a32e25ad16dfabf1331388329a231f156c14d5bc57f9f78cecc0:0
("external", 3) | 0.00000420 BTC | 67db9c12102be70d1ed56eefa6f98941def25b2578371e59e737d7e175c838d4:0
("external", 6) | 0.00010500 BTC | dfcb2cab93eeca6439bccb4bae0150919fa2e05d05477793fb6c8d2a3ff73366:0
("external", 7) | 0.04186496 BTC | 28d6c83e62031b25b79a31803cd5b27af56632169d9807d235d69fe63ef2d017:0
("internal", 2) | 0.01430656 BTC | f34a49113dd663325969417ee38d283144516eeea755197a4cd83a8a64b4e624:1
("internal", 3) | 0.00031314 BTC | dfcb2cab93eeca6439bccb4bae0150919fa2e05d05477793fb6c8d2a3ff73366:1
```

LagginTimes (approved):
ACK 6aa81fe2ff5737add0be0664347a786573060706

Also ran the `filter_iter` example on signet, results:
```
Matched block 205061
Matched block 208441
Matched block 219412
Matched block 228230
Matched block 235619
Matched block 237344
Matched block 250827
Matched block 251917
Matched block 262664
Matched block 266203
Matched block 268631

took: 102s
Local tip: 268715

Unspent
("external", 3) | 1.00104000 BTC | c214d5f31fd8a32e25ad16dfabf1331388329a231f156c14d5bc57f9f78cecc0:0
("external", 3) | 0.00000420 BTC | 67db9c12102be70d1ed56eefa6f98941def25b2578371e59e737d7e175c838d4:0
("external", 6) | 0.00010500 BTC | dfcb2cab93eeca6439bccb4bae0150919fa2e05d05477793fb6c8d2a3ff73366:0
("external", 7) | 0.04186496 BTC | 28d6c83e62031b25b79a31803cd5b27af56632169d9807d235d69fe63ef2d017:0
("internal", 2) | 0.01430656 BTC | f34a49113dd663325969417ee38d283144516eeea755197a4cd83a8a64b4e624:1
("internal", 3) | 0.00031314 BTC | dfcb2cab93eeca6439bccb4bae0150919fa2e05d05477793fb6c8d2a3ff73366:1
```

ValuedMammal:
ACK 6aa81fe2ff5737add0be0664347a786573060706

PR #1766, opened 2024-12-09
chore(bdk_chain): bump rusqlite dependency from 0.31.0 to 0.32

Einliterflasche:
### Description

This PR bumps the `chain` crates feature dependency on `rusqlite` to the more recent version `0.32`. 

### Notes to the reviewers

I omitted the patch version because all versions 0.32.x must be compatible per the semver spec. 

I submitted this PR because the "outdated" dependency breaks dependency resolution in my project. 

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

Issue #1949, opened 2025-04-26
Update `bdk` out of date dependencies


notmandatory:
Where possible we should update out of date dependencies, including for dev and example projects.

https://deps.rs/repo/github/bitcoindevkit/bdk

Also tracking:

- [x] ~rusqlite~
- [ ] rand
- [ ] criterion
- [ ] hashbrown
- [ ] bincode bitcoindevkit/bdk#2092
- [ ] tokio
- [x] electrsd bitcoindevkit/bdk#1826 
- [ ] ctrlc

tnull:
> * rusqlite

I'd be careful around this one, as you can't have multiple `libsqlite3-sys` versions in your dependency tree. In fact, I know that multiple projects kinda 'standardized' around the version currently used in BDK/CDK/LDK Node for this reason. So needlessly bumping the dependency would render BDK incompatible with all of these crates, requiring corresponding bumps and releases on all sides.

notmandatory:
@tnull good call. I took it off the list.

notmandatory:
Moving this to the 3.0 milestone when I hope we'll be able to bump to a higher MSRV which will make these updates easier.

luisschwab:
Since `bincode` is now unmaintained (see https://sr.ht/~stygianentity/bincode/), we need to figure out whether to keep it of find a suitable alternative.

oleonardolima:
> Since `bincode` is now unmaintained (see https://sr.ht/~stygianentity/bincode/), we need to figure out whether to keep it of find a suitable alternative.

I updated the checklist to have the tracking issue: https://github.com/bitcoindevkit/bdk/issues/2092

ANAMASGARD:
@notmandatory  @oleonardolima  @luisschwab  Hi! I’d like to work on the remaining `rand` dependency update as a focused PR.

I plan to update `rand` in `bdk_chain`, migrate the affected test code to the current API, and verify the workspace tests and MSRV compatibility. Is anyone already working on this, or is this contribution welcome?

PR #2002, opened 2025-07-29
bump(bdk_chain): rusqlite to 0.36.0

binarybaron:
### Description

This PR bumps the chain crates feature dependency on rusqlite to the more recent version 0.36.0.

### Notes to the reviewers

I submitted this PR because the "outdated" dependency breaks dependency resolution in my project.

See https://github.com/bitcoindevkit/bdk/pull/1766

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/Cargo.toml
+++ b/crates/chain/Cargo.toml
@@ -22,7 +22,7 @@ serde = { version = "1", optional = true, features = ["derive", "rc"] }
 miniscript = { version = "12.3.1", optional = true, default-features = false }
 
 # Feature dependencies
-rusqlite = { version = "0.31.0", features = ["bundled"], optional = true }
+rusqlite = { version = "0.37.0", features = ["bundled"], optional = true }
 
 [dev-dependencies]
 rand = "0.8"
--- a/crates/chain/Cargo.toml
+++ b/crates/chain/Cargo.toml
@@ -22,7 +22,7 @@ serde = { version = "1", optional = true, features = ["derive", "rc"] }
 miniscript = { version = "12.3.1", optional = true, default-features = false }
 
 # Feature dependencies
-rusqlite = { version = "0.37.0", features = ["bundled"], optional = true }
+rusqlite = { version = "0.36.0", features = ["bundled"], optional = true }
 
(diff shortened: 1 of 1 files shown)


tnull (commented):
I don't think that unilaterally bumping the `rusqlite` dependency just in BDK is a good idea, as it will result in incompatibility with other crates in the `rust-bitcoin` space. Unfortunately, 0.31 is the current 'agreed upon' standard across crates. 

If we see a need to bump it, this will take some coordination effort across projects (BDK, LDK, Cashu, Fedimint, Payjoin, at the very least would be affected).

binarybaron:
Changed to 0.36.0 because `arti-client` uses 0.36.0

notmandatory:
@tnull & @binarybaron what would you think about extracting the `bdk_sqlite` persistence store from the `bdk_chain` crate and putting it in it's own crate. If we do that we could have a `1.0.0` version with the current  0.31.0 `rusqlite`  and release a new `2.0.0` version with 0.36.0 `rusqlite`, while still back-porting bug / features for the 1.x.  Or open to other ideas along these lines.

tnull:
> @tnull & @binarybaron what would you think about extracting the `bdk_sqlite` persistence store from the `bdk_chain` crate and putting it in it's own crate. If we do that we could have a `1.0.0` version with the current 0.31.0 `rusqlite` and release a new `2.0.0` version with 0.36.0 `rusqlite`, while still back-porting bug / features for the 1.x. Or open to other ideas along these lines.

So, moving the SQLite store into an individual crate would generally be a good idea, IMO. I however don't think that it would solve the issue, at least if you don't want to go down the road of introducing features to enable individual `rusqlite` versions. As mentioned above, a lot of projects in the space by now depend on the exact `rusqlite` version, and I think it would be important to check in with most of them to coordinate such breakage. Apart from that, yes, it would of course also require a major version bump for all of the involved projects.

luisschwab:
Closing per https://github.com/bitcoindevkit/bdk/issues/1949#issuecomment-2831715363 and the comments above.

Issue #1982, opened 2025-06-26
Add `populate_anchor_cache` method to `BdkElectrumClient`


evanlinjin:
Can be done in a follow up, but we can have a `populate_anchor_cache` method. This will speed things up even more for subsequent syncs.

_Originally posted by @evanlinjin in https://github.com/bitcoindevkit/bdk/pull/1957#discussion_r2168070612_

## To Do (`bdk`)
            
- [ ] Add `BdkElectrumClient::populate_anchor_cache`
- [ ] (Maybe) Convenient method on `TxGraph`
- [ ] Update examples

## To Do (`bdk_wallet`)

- [ ] Convenient method on `Wallet`
- [ ] Update examples

Her-Code:
Hey, I'd love to work on this

evanlinjin:
@Her-Code thank you for offering. I've assigned you

Her-Code:
@evanlinjin  Awesome, thanks! I’m on it

kwsantiago:
Hi @Her-Code are you still working on this?

If not, I'd like to give this a go if possible.

Her-Code:
Heyy @kwsantiago , yes I'm still working on it.

LagginTimes:
Issue resolved with #2005.

PR #2005, opened 2025-08-01
Add populate_anchor_cache method to bdk

Her-Code:
This PR addresses #1982
It introduces functionality to populate the anchor cache, improving how BDK handles stateful or performance-critical anchor data.

### Changes
- Add `populate_anchor_cache` method to `BdkElectrumClient `in `bdk-core`
- Improves sync speed by caching anchor transactions in advance
- Enables pre-caching of anchor transactions to reduce redundant network calls

Follow-up work will extend this functionality to `bdk-wallet`.

### DIFF
--- a/examples/example_electrum/src/main.rs
+++ b/examples/example_electrum/src/main.rs
@@ -126,16 +126,13 @@ fn main() -> anyhow::Result<()> {
 
     let client = BdkElectrumClient::new(electrum_cmd.electrum_args().client(network)?);
 
-    // Tell the electrum client about the txs we've already got locally so it doesn't re-download
-    // them
-    client.populate_tx_cache(
-        graph
-            .lock()
-            .unwrap()
-            .graph()
-            .full_txs()
-            .map(|tx_node| tx_node.tx),
-    );
+    // Tell the electrum client about the txs and anchors we've already got locally so it doesn't
+    // re-download .them
+    {
+        let graph = graph.lock().unwrap();
+        client.populate_tx_cache(graph.graph().full_txs().map(|tx_node| tx_node.tx));
+        client.populate_anchor_cache(graph.graph().all_anchors().clone());
+    }
 
     let (chain_update, tx_update, keychain_update) = match electrum_cmd.clone() {
         ElectrumCommands::Scan {
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -7,6 +7,7 @@ use bdk_core::{
     BlockId, CheckPoint, ConfirmationBlockTime, TxUpdate,
 };
 use electrum_client::{ElectrumApi, Error, HeaderNotification};
+use std::convert::TryInto;
 use std::sync::{Arc, Mutex};
 
 /// We include a chain suffix of a certain length for the purpose of robustness.
@@ -37,8 +38,24 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
         }
     }
 
-    /// Inserts transactions into the transaction cache so that the client will not fetch these
-    /// transactions.
+    /// Insert anchors into the anchor cache so that the client will not re-fetch them.
+    ///
+    /// Typically used to pre-populate the cache from an existing `TxGraph`.
+    pub fn populate_anchor_cache(
+        &self,
+        tx_anchors: impl IntoIterator<Item = (Txid, impl IntoIterator<Item = ConfirmationBlockTime>)>,
+    ) {
+        let mut cache = self.anchor_cache.lock().unwrap();
+        for (txid, anchors) in tx_anchors {
+            for anchor in anchors {
+                cache.insert((txid, anchor.block_id.hash), anchor);
+            }
+        }
+    }
+
+    /// Insert transactions into the transaction cache so that the client will not re-fetch them.
+    ///
+    /// Typically used to pre-populate the cache from an existing `TxGraph`.
     pub fn populate_tx_cache(&self, txs: impl IntoIterator<Item = impl Into<Arc<Transaction>>>) {
         let mut tx_cache = self.tx_cache.lock().unwrap();
         for tx in txs {


evanlinjin:
Thank you for approaching this.

The ticket was very brief and was not specific enough - sorry about that. Let me clarify what we are looking for.

* We need a way to populate the anchor cache with pre-existing data. For example, from our receiving structures (`bdk_wallet::Wallet`, `bdk_chain::TxGraph`+`bdk_chain::LocalChain`).
* This is similar to the `BdkElectrumClient::populate_tx_cache` method that we already have (can draw inspiration from that).

In other words, `populate_anchor_cache` should NOT be handling IO (i.e. fetching data from an Electrum server) and should only use data that is provided directly to it.

LagginTimes:
Thanks for the contribution! A couple of things to take care of:

1. Rebase on the current `master` to bring it up to date with the latest changes.
2. Run `cargo clippy` and `cargo +nightly fmt` so it passes linting and formatting checks.

examples/example_electrum/src/main.rs
```diff
@@ -94,6 +94,26 @@ pub struct ScanOptions {
...
+}
+
+impl TxGraphExt for TxGraph<ConfirmationBlockTime> {
+    fn iter_anchors_ext(
+        &self,
+    ) -> Box<dyn Iterator<Item = ((Txid, BlockHash), ConfirmationBlockTime)> + '_> {
+        Box::new(self.full_txs().flat_map(move |tx_node| {
+            tx_node.anchors.iter().map(move |anchor| {
+                let block_id = anchor.block_id;
+                ((tx_node.txid, block_id.hash), *anchor)
```
evanlinjin: You are getting the block hash from the anchor, so the API doesn't need to take in the block hash separately.

examples/example_electrum/src/main.rs
```diff
@@ -94,6 +94,26 @@ pub struct ScanOptions {
     pub batch_size: usize,
 }
 
+/// Extension trait to expose anchors in a public-friendly way
+trait TxGraphExt {
```
evanlinjin: Only introduce extension traits out of necessity. Is there another method on `TxGraph` that you can use? If not, you can introduce a method on `TxGraph` directly.

evanlinjin (changes requested):

crates/chain/src/tx_graph.rs
```diff
@@ -298,6 +298,13 @@ impl std::error::Error for CalculateFeeError {}
...
     /// Iterate over all tx outputs known by [`TxGraph`].
     ///
+    /// This returns an iterator over all anchors in the graph as (txid, anchor) pairs
+    pub fn anchors(&self) -> impl Iterator<Item = (Txid, &A)> + '_ {
+        self.anchors
+            .iter()
+            .flat_map(|(txid, anchors)| anchors.iter().map(move |a| (*txid, a)))
+    }
+
     /// This includes txouts of both full transactions as well as floating transactions.
```
evanlinjin: I think you inserted `anchors` in the middle of the doc comments of `all_txouts`.

examples/example_electrum/src/main.rs
```diff
@@ -126,6 +126,16 @@ fn main() -> anyhow::Result<()> {
 
     let client = BdkElectrumClient::new(electrum_cmd.electrum_args().client(network)?);
 
+    // This avoids re-fetching known anchor confirmations from Electrum.
+    // Hold the lock for the entire anchor extraction
+    let graph_guard = graph.lock().unwrap();
```
evanlinjin: Have you tried running this? Since we lock `graph` again on line 143, wouldn't this result in a deadlock?

examples/example_electrum/src/main.rs
```diff
@@ -126,6 +126,16 @@ fn main() -> anyhow::Result<()> {
...
+    // Collect anchors to release the lock quickly
+    let mut anchors = Vec::new();
+    for (txid, anchor) in graph_guard.graph().anchors() {
+        anchors.push((txid, *anchor)); // Dereference instead of clone for Copy types
+    }
+    client.populate_anchor_cache(anchors);
+
```
evanlinjin: Some of these comments are explaining very obvious things in a way that doesn’t add much beyond reading the code itself.

crates/chain/src/tx_graph.rs
```diff
@@ -298,6 +298,13 @@ impl std::error::Error for CalculateFeeError {}
 impl<A> TxGraph<A> {
     /// Iterate over all tx outputs known by [`TxGraph`].
     ///
+    /// This returns an iterator over all anchors in the graph as (txid, anchor) pairs
+    pub fn anchors(&self) -> impl Iterator<Item = (Txid, &A)> + '_ {
+        self.anchors
+            .iter()
+            .flat_map(|(txid, anchors)| anchors.iter().map(move |a| (*txid, a)))
+    }
```
evanlinjin: Instead of introducing this `anchors` method, how about we make use of the `all_anchors` method we have already?

I propose changing `BdkElectrumClient::populate_anchor_cache` to the following:

```rust
    /// Insert anchors into the anchor cache so that they don’t have to be fetched again from
    /// Electrum. Typically used to pre-populate the cache from an existing `TxGraph`.
    pub fn populate_anchor_cache(
        &self,
        tx_anchors: impl IntoIterator<Item = (Txid, impl IntoIterator<Item = ConfirmationBlockTime>)>,
    ) {
        let mut cache = self.anchor_cache.lock().unwrap();
        for (txid, anchors) in tx_anchors {
            for anchor in anchors {
                cache.insert((txid, anchor.block_id.hash), anchor);
            }
        }
    }
```

Using the API will look like this:

```rust
        let graph = graph.lock().unwrap();
        client.populate_anchor_cache(graph.graph().all_anchors().clone());
```

examples/example_electrum/src/main.rs
evanlinjin: Also, how about we populate the anchor & tx cache under a single lock of `graph`?

evanlinjin (changes requested):
Please double check and try running the code you have written (or the code written for you).

Comments need to capture non-obvious intent or constraints, not restate what is obvious by looking at the code.

evanlinjin:
@Her-Code Thanks for taking this on! It would be really helpful to have this in soon. Do you think you’d be able to complete it by the end of the week, or would you need more time?

Her-Code:
I've implemented the anchor caching using all_anchors() as you suggested. I noticed the original implementation could potentially create nested locks:
```rust
        let graph = graph.lock().unwrap();
        client.populate_anchor_cache(graph.graph().all_anchors().clone());
```
    
so I added explicit scoping with curly braces to ensure proper lock management:

```rust
{
        // Lock graph mutex - released automatically at scope end
        // Cache all anchors while holding the lock
        let graph = graph.lock().unwrap();
        client.populate_anchor_cache(graph.graph().all_anchors().clone());
}

evanlinjin (approved):
ACK 191bdb154e99de616c2409fbfcab3ef5e2257262

Issue #1942, opened 2025-04-16
Variable block confirmation for confirmed balances


matthiasdebernardini:
Describe the enhancement

Allow wallet.balance().confirmed to be configurable based on how many blocks it should consider it to be confirmed, 1 by default then 3 or 6. Not sure what else could be useful.

Use case

Wallets designed for large balances should prefer to have this set to a high number, 6 for example.

**Additional context**

Got posted in `bdk_wallet` first, @evanlinjin noted that this can be achieved via `wallet.local_chain().tip()`
https://github.com/bitcoindevkit/bdk_wallet/issues/210

Echomod-alt:
@matthiasdebernardini If you need assistance, please don’t hesitate to contact our support team at instant.gitresponse@gmail.com. We’re here to assist.

rustaceanrob:
1 doesn't feel like a great default to me. Reorgs of depth one happen all the time. I would go with 3, which to my knowledge has never occured, and then leave 1, 3, 6 as options.

matthiasdebernardini:
I agree. If confirmed means `finalfinal` or not subject to change, then yeah it should be 3 and not 1. Though it should be optional.

uvuvuwu:
Can I take this?

LagginTimes:
Hi @uvuvuwu! Just checking in to see if this is still being actively worked on — or if there’s anything we can do to help move it forward. Let us know if you're still interested or if you’ve run into any blockers.

uvuvuwu:
Hi @LagginTimes . Sorry I am no longer able to work on this actively. Please feel free to take over.

LagginTimes:
Closed with #2007.

PR #2007, opened 2025-08-01
docs(chain): add doctest for min confirmation balance filtering

LagginTimes:
Implements #1942.

### Description

This PR adds a new doctest demonstrating how to simulate a minimum confirmation threshold for confirmed balance calculations by adjusting the `chain_tip` height passed to `TxGraph::balance`.

### Changelog notice

- Added a doctest illustrating how to filter confirmed balance results by simulating a minimum confirmation threshold via `chain_tip` height.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -1315,6 +1315,69 @@ impl<A: Anchor> TxGraph<A> {
     ///
     /// This is the infallible version of [`try_balance`].
     ///
+    /// ### Minimum confirmations
+    ///
+    /// To filter for transactions with at least `N` confirmations, pass a `chain_tip` that is
+    /// `N - 1` blocks below the actual tip. This ensures that only transactions with at least `N`
+    /// confirmations are counted as confirmed in the returned [`Balance`].
+    ///
+    /// ```
+    /// # use bdk_chain::tx_graph::TxGraph;
+    /// # use bdk_chain::{local_chain::LocalChain, CanonicalizationParams, ConfirmationBlockTime};
+    /// # use bdk_testenv::{hash, utils::new_tx};
+    /// # use bitcoin::{Amount, OutPoint, ScriptBuf, Transaction, TxIn, TxOut};
+    ///
+    /// # let spk = ScriptBuf::from_hex("0014c692ecf13534982a9a2834565cbd37add8027140").unwrap();
+    /// # let chain =
+    /// #     LocalChain::from_blocks((0..=15).map(|i| (i as u32, hash!("h"))).collect()).unwrap();
+    /// # let mut graph: TxGraph = TxGraph::default();
+    /// # let coinbase_tx = Transaction {
+    /// #     input: vec![TxIn {
+    /// #         previous_output: OutPoint::null(),
+    /// #         ..Default::default()
+    /// #     }],
+    /// #     output: vec![TxOut {
+    /// #         value: Amount::from_sat(70000),
+    /// #         script_pubkey: spk.clone(),
+    /// #     }],
+    /// #     ..new_tx(0)
+    /// # };
+    /// # let tx = Transaction {
+    /// #     input: vec![TxIn {
+    /// #         previous_output: OutPoint::new(coinbase_tx.compute_txid(), 0),
+    /// #         ..Default::default()
+    /// #     }],
+    /// #     output: vec![TxOut {
+    /// #         value: Amount::from_sat(42_000),
+    /// #         script_pubkey: spk.clone(),
+    /// #     }],
+    /// #     ..new_tx(1)
+    /// # };
+    /// # let txid = tx.compute_txid();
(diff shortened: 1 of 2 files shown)


evanlinjin (changes requested):
Thank you for working on this.

I do not think modifying the canonicalization is the right approach.

For example, if we have a min-confirmation-count of 3. We expect that a transaction confirmed at depth 2 should be prioritized over a transaction with no confirmations. With the current state of the PR, this guarantee is nonexistent.

Additionally, different min-confirmation-count values should not influence the `Balance::total` value. It should only shift values between `*pending,immature,confirmed`.

### Idea 1

The simplest solution would be to not do any code changes and provide a doctest showing how we could manipulate the `chain_tip` input of `TxGraph::[try_]balance`. I.e. with `LocalChain`/`CheckPoint`s.

```rust
let tip = chain.tip();

let target_height = tip.height().saturating_sub(confirmation_threshold);

let target_tip = tip
    .range(..=target_height)
    .next()
    .expect("checkpoint from local chain must have genesis");

// Use `target_tip` as the `chain_tip` input of `TxGraph::[try_]balance`.
```

For `bdk_wallet::Wallet`, we should probably provide a convenience method of sorts so the caller does not need to dig into the internal structures.

### Idea 2

We can change the `trust_predicate` to be a more general `predicate`.

```rust
pub fn balance(
    /* Other Inputs */
    
    // predicate inputs: (txout_meta, txout, tip_height)
    mut predicate: impl FnMut(&OI, FullTxOut<A>, u32) -> BalanceType,
) -> Balance { todo!() }

// Maybe put this in `bdk_core`?
pub enum BalanceType {
    Immature,
    TrustedPending,
    UntrustedPending,
    Confirmed,
}
```

However, this is a breaking change and it makes it more involved to implement the `predicate`.

---

Let me know what you think.

LagginTimes:
> Thank you for working on this.
> ...
> Let me know what you think.

Thanks for the suggestions. I like Idea 1, and will work on implementing this.

crates/chain/src/tx_graph.rs
```diff
@@ -118,6 +118,92 @@
...
+//!     },
+//! );
+//!
+//! // With a confirmation threshold of 6, our transaction should show in the balance.
+//! let confirmation_threshold = 6;
+//! let target_height = tip.height().saturating_sub(confirmation_threshold - 1);
+//! let target_tip = chain
+//!     .range(..=target_height)
+//!     .next()
+//!     .expect("checkpoint from local chain must have genesis");
```
evanlinjin: This is a lot of code to achieve this. Maybe we should introduce convenience methods on `CheckPoint`/`LocalChain`:

```rust
impl CheckPoint {
    /// Returns the checkpoint at `height` if one exists, otherwise the nearest checkpoint at a
    /// lower height.
    ///
    /// This is equivalent to taking the "floor" of `height` over this checkpoint chain.
    ///
    /// Returns `None` if no checkpoint exists at or below the given height.
    pub fn floor_at(&self, height: u32) -> Option<Self> {
        self.range(..=height).next()
    }

    /// Returns the checkpoint located a number of heights below this one.
    ///
    /// This is a convenience wrapper for [`CheckPoint::floor_at`], subtracting `to_subtract` from
    /// the current height.
    ///
    /// - If a checkpoint exists exactly `offset` heights below, it is returned.
    /// - Otherwise, the nearest checkpoint *below that target height* is returned.
    ///
    /// Returns `None` if `to_subtract` is greater than the current height, or if there is no
    /// checkpoint at or below the target height.
    pub fn floor_below(&self, offset: u32) -> Option<Self> {
        self.floor_at(self.height().checked_sub(offset)?)
    }
}
```

Also, I feel like this this example should belong on `.balance` (easier to find).

evanlinjin (changes requested):

crates/chain/src/tx_graph.rs
evanlinjin: The `floor_at` method will be useful for initiating a block-based chain source. I.e. if a certain spk has a birthday at height 100, we can use this method to start scanning from that height. cc. @ValuedMammal

oleonardolima: Agreed, I think this example could live in `.balance`. Also, I don't see any issue with adding the convenience methods.

oleonardolima (approved):
ACK 657a044370a102259eb1b7f5abdfee2dde998bc2

evanlinjin (approved):
ACK 657a044370a102259eb1b7f5abdfee2dde998bc2

bladed-operative:
Height is an unsigned integer so what does it mean to take the "floor" of height over this checkpoint chain.

evanlinjin:
> Height is an unsigned integer so what does it mean to take the "floor" of height over this checkpoint chain.

The checkpoint chain is sparse. If the height at x does not exist, use a height that is directly lower.

PR #2011, opened 2025-08-12
fix(electrum): fix stale anchor hash on reorg

LagginTimes:
### Description

Fixes an issue in `batch_fetch_anchors()` in `bdk_electrum` where, if a stale block header was detected and a fresh header was fetched, the confirmation anchor inserted into `anchor_cache` still used the hash from the original stale header instead of the new one.

### Changelog notice

- `batch_fetch_anchors()` no longer uses a potentially stale hash as the anchor.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -518,17 +518,16 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
                 if let Some(anchor) = anchor_cache.get(&(txid, hash)) {
                     results.push((txid, *anchor));
                 } else {
-                    to_fetch.push((txid, height, hash));
+                    to_fetch.push((txid, height));
                 }
             }
         }
 
         // Fetch merkle proofs.
-        let txids_and_heights = to_fetch.iter().map(|&(txid, height, _)| (txid, height));
-        let proofs = self.inner.batch_transaction_get_merkle(txids_and_heights)?;
+        let proofs = self.inner.batch_transaction_get_merkle(to_fetch.iter())?;
 
         // Validate each proof, retrying once for each stale header.
-        for ((txid, height, hash), proof) in to_fetch.into_iter().zip(proofs.into_iter()) {
+        for ((txid, height), proof) in to_fetch.into_iter().zip(proofs.into_iter()) {
             let mut header = {
                 let cache = self.block_header_cache.lock().unwrap();
                 cache
@@ -553,6 +552,7 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
 
             // Build and cache the anchor if merkle proof is valid.
             if valid {
+                let hash = header.block_hash();
                 let anchor = ConfirmationBlockTime {
                     confirmation_time: header.time as u64,
                     block_id: BlockId {
@@ -696,11 +696,13 @@ fn chain_update(
 #[cfg(test)]
 #[cfg_attr(coverage_nightly, coverage(off))]
 mod test {
-    use crate::{bdk_electrum_client::TxUpdate, BdkElectrumClient};
+    use crate::{bdk_electrum_client::TxUpdate, electrum_client::ElectrumApi, BdkElectrumClient};
+    use bdk_chain::bitcoin::Amount;
     use bdk_chain::bitcoin::{constants, Network, OutPoint, ScriptBuf, Transaction, TxIn};
     use bdk_chain::{BlockId, CheckPoint};
     use bdk_core::{collections::BTreeMap, spk_client::SyncRequest};
-    use bdk_testenv::{anyhow, utils::new_tx, TestEnv};
+    use bdk_testenv::{anyhow, bitcoincore_rpc::RpcApi, utils::new_tx, TestEnv};
+    use core::time::Duration;
     use electrum_client::Error as ElectrumError;
     use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


oleonardolima (commented):
utACK c8a33267a57d4b97c2959ae9a988f9d2c256a843

evanlinjin (changes requested):
In general, this looks like a solid simplification and likely fixes the problem in the PR description. Can we add a test that triggers a reorg after `fetch_tip_and_latest_blocks` to confirm?

While reviewing, I noticed the chain-update logic is a bit convoluted, which makes this (and other `bdk_electrum` changes) tricky to follow:

- `block_header_cache` is keyed by height (not block hash), so it’s unclear how we’re ensuring the right header.  
- The [`!valid` block section](https://github.com/LagginTimes/bdk/blob/c8a33267a57d4b97c2959ae9a988f9d2c256a843/crates/electrum/src/bdk_electrum_client.rs#L523-L534) updates `block_header_cache` on reorgs after `fetch_tip_and_latest_blocks`, meaning headers from different chains can co-exist. Anchors from `batch_fetch_anchors` may also not match the same chain.  
- In `chain_update`, we use blockhashes from anchors to craft the `CheckPoint` update. Because inconsistencies are possible, we also pass in `lastest_blocks`, assuming reorgs aren’t deeper than `CHAIN_SUFFIX_LENGTH`.  

**I suggest we merge this PR ASAP (since it’s an important fix).**

Then, in a follow-up PR, we could simplify the codebase:

1. Key `block_header_cache` by blockhash instead of height. Fetch by height through the `CheckPoint` chain.  
2. Remove the [`!valid` block section](https://github.com/LagginTimes/bdk/blob/c8a33267a57d4b97c2959ae9a988f9d2c256a843/crates/electrum/src/bdk_electrum_client.rs#L523-L534), since headers will always match the chain. Reorgs will naturally invalidate anchors and re-fetch them.  
3. Let `fetch_tip_and_latest_blocks()` just fetch blocks. Move chain-update logic into `chain_update()`, updated to take `heights_we_are_interested_in` instead of anchors. Call `chain_update()` after `batch_fetch_anchors`.  

This may slightly slow down `bdk_electrum` (extra header lookups via `CheckPoint`), but IO is the real bottleneck. Once the skip-list is in, `CheckPoint` lookups will be fast.

---

On second though... maybe we should only do bug fixes and focus effort in `bdk_electrum_streaming`?

ValuedMammal (approved):

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -747,4 +749,59 @@ mod test {
...
+
+    #[cfg(feature = "default")]
+    #[test]
+    fn test_batch_fetch_anchors_reorg_uses_new_hash() -> anyhow::Result<()> {
+        let env = TestEnv::new()?;
+        let client = electrum_client::Client::new(env.electrsd.electrum_url.as_str()).unwrap();
+        let electrum_client = BdkElectrumClient::new(client);
+
+        env.mine_blocks(101, None)?;
+        env.wait_until_electrum_sees_block(Duration::from_secs(6))?;
```
evanlinjin: Not needed.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -747,4 +749,59 @@ mod test {
...
+
+        let addr = env
+            .rpc_client()
+            .get_new_address(None, None)?
+            .assume_checked();
+        let txid = env.send(&addr, Amount::from_sat(50_000))?;
+        env.wait_until_electrum_sees_txid(txid, Duration::from_secs(6))?;
```
evanlinjin: Not needed.

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -747,4 +749,59 @@ mod test {
 
         Ok(())
     }
+
```
evanlinjin, line 769: A doc comment will make the test easier to understand.

```suggestion

/// This test checks that when a transaction is reorged into a different block
/// at the same height, `batch_fetch_anchors()` updates its anchor correctly:
///
/// 1. A transaction is confirmed in a block, and that block header is cached.
/// 2. A reorg happens, replacing that block with a new one at the same height.
/// 3. When we call `batch_fetch_anchors()`, it should fetch the new block header
///    and recreate the transaction’s anchor using the new block hash.
///
/// In short: reorgs should cause the anchor to point to the new block, not the
/// outdated cached one.

```

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -747,4 +749,59 @@ mod test {
...
+            .assume_checked();
+        let txid = env.send(&addr, Amount::from_sat(50_000))?;
+        env.wait_until_electrum_sees_txid(txid, Duration::from_secs(6))?;
+
+        // Mine block that confirms transaction.
+        env.mine_blocks(1, None)?;
+        env.wait_until_electrum_sees_block(Duration::from_secs(6))?;
+        let height: u32 = env.rpc_client().get_block_count()? as u32;
+
+        // Populate header cache.
```
evanlinjin: Let's be more specific.


```suggestion
        // Add the pre-reorg block (that the tx is confirmed in) to the header cache.

```

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -747,4 +749,59 @@ mod test {
...
+                .lock()
+                .unwrap()
+                .insert(height, header);
+        }
+
+        // Reorg to create a new header and hash.
+        env.reorg(1)?;
+        env.wait_until_electrum_sees_block(Duration::from_secs(6))?;
+
+        // Calling `batch_fetch_anchors` should fetch new header.
```
```suggestion
        // Calling `batch_fetch_anchors` should fetch new header, displacing the pre-reorg header.

```

evanlinjin (commented):
Looks good.

Just a few nits in terms of documenting the test and unnecessary calls.

evanlinjin (approved):
ACK fdec863ed134f2b94186a3c5d462a5d1e65956ab

PR #2012, opened 2025-08-12
Add functions to test persistence

110CodingP:
### Description

Added a helper `assert_persist_changesets` to the testenv for testing custom persistence backends. Partially solves [bdk_wallet#14](https://github.com/bitcoindevkit/bdk_wallet/issues/14) and might help with [bdk_wallet#234](https://github.com/bitcoindevkit/bdk_wallet/issues/234).

### Changelog notice

``` 
Added
   - `miniscript` feature to testenv which depends on `bdk_chain/miniscript` .
    - `assert_persist_changesets` to testenv to check `ChangeSet`s are persisted properly.
    - `tx_graph_changesets`, `keychain_txout_changesets`, `local_chain_changesets` which provide changesets
    to be used as arguments to `assert_persist_changesets`.
    - `create_test_tx`, `spk_at_index`,`parse_descriptor`, `create_txout` helpers to testenv for creating changesets for testing.
 Changed
   - default feature of testenv to also activate `miniscript` feature.
```

### Todo
* [x] Add docs
* [x] Add tests to bdk_chain based on testenv utilities

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/tests/test_rusqlite_impl.rs
+++ b/crates/chain/tests/test_rusqlite_impl.rs
@@ -1,0 +1,109 @@
+#![cfg(feature = "rusqlite")]
+use bdk_chain::{keychain_txout, local_chain, tx_graph, ConfirmationBlockTime};
+use bdk_testenv::persist_test_utils::{
+    persist_indexer_changeset, persist_local_chain_changeset, persist_txgraph_changeset,
+};
+
+fn tx_graph_changeset_init(
+    db: &mut rusqlite::Connection,
+) -> rusqlite::Result<tx_graph::ChangeSet<ConfirmationBlockTime>> {
+    let db_tx = db.transaction()?;
+    tx_graph::ChangeSet::<ConfirmationBlockTime>::init_sqlite_tables(&db_tx)?;
+    let changeset = tx_graph::ChangeSet::<ConfirmationBlockTime>::from_sqlite(&db_tx)?;
+    db_tx.commit()?;
+    Ok(changeset)
+}
+
+fn tx_graph_changeset_persist(
+    db: &mut rusqlite::Connection,
+    changeset: &tx_graph::ChangeSet<ConfirmationBlockTime>,
+) -> rusqlite::Result<()> {
+    let db_tx = db.transaction()?;
+    changeset.persist_to_sqlite(&db_tx)?;
+    db_tx.commit()
+}
+
+fn keychain_txout_changeset_init(
+    db: &mut rusqlite::Connection,
+) -> rusqlite::Result<keychain_txout::ChangeSet> {
+    let db_tx = db.transaction()?;
+    keychain_txout::ChangeSet::init_sqlite_tables(&db_tx)?;
+    let changeset = keychain_txout::ChangeSet::from_sqlite(&db_tx)?;
+    db_tx.commit()?;
+    Ok(changeset)
+}
+
+fn keychain_txout_changeset_persist(
+    db: &mut rusqlite::Connection,
+    changeset: &keychain_txout::ChangeSet,
+) -> rusqlite::Result<()> {
+    let db_tx = db.transaction()?;
+    changeset.persist_to_sqlite(&db_tx)?;
+    db_tx.commit()
+}
+
+fn local_chain_changeset_init(
+    db: &mut rusqlite::Connection,
+) -> rusqlite::Result<local_chain::ChangeSet> {
+    let db_tx = db.transaction()?;
+    local_chain::ChangeSet::init_sqlite_tables(&db_tx)?;
+    let changeset = local_chain::ChangeSet::from_sqlite(&db_tx)?;
+    db_tx.commit()?;
+    Ok(changeset)
+}
+
+fn local_chain_changeset_persist(
+    db: &mut rusqlite::Connection,
+    changeset: &local_chain::ChangeSet,
+) -> rusqlite::Result<()> {
+    let db_tx = db.transaction()?;
+    changeset.persist_to_sqlite(&db_tx)?;
+    db_tx.commit()
+}
+
+#[test]
+fn txgraph_is_persisted() -> anyhow::Result<()> {
+    let temp_dir = tempfile::tempdir().unwrap();
+    Ok(persist_txgraph_changeset::<rusqlite::Connection, _, _, _>(
+        || {
+            Ok(rusqlite::Connection::open(
+                temp_dir.path().join("wallet.sqlite"),
+            )?)
+        },
+        |db| Ok(tx_graph_changeset_init(db)?),
+        |db, changeset| Ok(tx_graph_changeset_persist(db, changeset)?),
+    )?)
+}
+
+#[test]
+fn indexer_is_persisted() -> anyhow::Result<()> {
+    let temp_dir = tempfile::tempdir().unwrap();
+    Ok(persist_indexer_changeset::<rusqlite::Connection, _, _, _>(
+        || {
+            Ok(rusqlite::Connection::open(
+                temp_dir.path().join("wallet.sqlite"),
+            )?)
+        },
+        |db| Ok(keychain_txout_changeset_init(db)?),
+        |db, changeset| Ok(keychain_txout_changeset_persist(db, changeset)?),
+    )?)
+}
+
+#[test]
+fn local_chain_is_persisted() -> anyhow::Result<()> {
+    let temp_dir = tempfile::tempdir().unwrap();
+    Ok(persist_local_chain_changeset::<
+        rusqlite::Connection,
+        _,
+        _,
+        _,
+    >(
+        || {
+            Ok(rusqlite::Connection::open(
+                temp_dir.path().join("wallet.sqlite"),
+            )?)
+        },
+        |db| Ok(local_chain_changeset_init(db)?),
+        |db, changeset| Ok(local_chain_changeset_persist(db, changeset)?),
+    )?)
+}
--- a/crates/chain/tests/test_rusqlite_impl.rs
+++ b/crates/chain/tests/test_rusqlite_impl.rs
@@ -1,109 +1,87 @@
 #![cfg(feature = "rusqlite")]
 use bdk_chain::{keychain_txout, local_chain, tx_graph, ConfirmationBlockTime};
 use bdk_testenv::persist_test_utils::{
-    persist_indexer_changeset, persist_local_chain_changeset, persist_txgraph_changeset,
+    assert_persist_changesets, keychain_txout_changesets, local_chain_changesets,
+    tx_graph_changesets,
 };
-
-fn tx_graph_changeset_init(
-    db: &mut rusqlite::Connection,
-) -> rusqlite::Result<tx_graph::ChangeSet<ConfirmationBlockTime>> {
-    let db_tx = db.transaction()?;
-    tx_graph::ChangeSet::<ConfirmationBlockTime>::init_sqlite_tables(&db_tx)?;
-    let changeset = tx_graph::ChangeSet::<ConfirmationBlockTime>::from_sqlite(&db_tx)?;
-    db_tx.commit()?;
-    Ok(changeset)
-}
-
-fn tx_graph_changeset_persist(
-    db: &mut rusqlite::Connection,
-    changeset: &tx_graph::ChangeSet<ConfirmationBlockTime>,
-) -> rusqlite::Result<()> {
-    let db_tx = db.transaction()?;
-    changeset.persist_to_sqlite(&db_tx)?;
-    db_tx.commit()
-}
-
-fn keychain_txout_changeset_init(
-    db: &mut rusqlite::Connection,
-) -> rusqlite::Result<keychain_txout::ChangeSet> {
-    let db_tx = db.transaction()?;
-    keychain_txout::ChangeSet::init_sqlite_tables(&db_tx)?;
-    let changeset = keychain_txout::ChangeSet::from_sqlite(&db_tx)?;
-    db_tx.commit()?;
-    Ok(changeset)
-}
-
-fn keychain_txout_changeset_persist(
-    db: &mut rusqlite::Connection,
-    changeset: &keychain_txout::ChangeSet,
-) -> rusqlite::Result<()> {
-    let db_tx = db.transaction()?;
-    changeset.persist_to_sqlite(&db_tx)?;
-    db_tx.commit()
-}
-
-fn local_chain_changeset_init(
-    db: &mut rusqlite::Connection,
-) -> rusqlite::Result<local_chain::ChangeSet> {
-    let db_tx = db.transaction()?;
-    local_chain::ChangeSet::init_sqlite_tables(&db_tx)?;
-    let changeset = local_chain::ChangeSet::from_sqlite(&db_tx)?;
-    db_tx.commit()?;
-    Ok(changeset)
-}
-
-fn local_chain_changeset_persist(
-    db: &mut rusqlite::Connection,
-    changeset: &local_chain::ChangeSet,
-) -> rusqlite::Result<()> {
-    let db_tx = db.transaction()?;
-    changeset.persist_to_sqlite(&db_tx)?;
-    db_tx.commit()
-}
 
 #[test]
 fn txgraph_is_persisted() -> anyhow::Result<()> {
     let temp_dir = tempfile::tempdir().unwrap();
-    Ok(persist_txgraph_changeset::<rusqlite::Connection, _, _, _>(
+    let changesets = tx_graph_changesets();
+    Ok(assert_persist_changesets(
         || {
             Ok(rusqlite::Connection::open(
                 temp_dir.path().join("wallet.sqlite"),
             )?)
         },
-        |db| Ok(tx_graph_changeset_init(db)?),
-        |db, changeset| Ok(tx_graph_changeset_persist(db, changeset)?),
+        |db| {
+            let db_tx = db.transaction()?;
+            tx_graph::ChangeSet::<ConfirmationBlockTime>::init_sqlite_tables(&db_tx)?;
+            let changeset = tx_graph::ChangeSet::<ConfirmationBlockTime>::from_sqlite(&db_tx)?;
+            db_tx.commit()?;
+            Ok(changeset)
+        },
+        |db, changeset| {
+            let db_tx = db.transaction()?;
+            changeset.persist_to_sqlite(&db_tx)?;
+            db_tx.commit()?;
+            Ok(())
+        },
+        &changesets,
     )?)
 }
 
 #[test]
 fn indexer_is_persisted() -> anyhow::Result<()> {
     let temp_dir = tempfile::tempdir().unwrap();
-    Ok(persist_indexer_changeset::<rusqlite::Connection, _, _, _>(
+    let changesets = keychain_txout_changesets();
+    Ok(assert_persist_changesets(
         || {
             Ok(rusqlite::Connection::open(
                 temp_dir.path().join("wallet.sqlite"),
             )?)
         },
-        |db| Ok(keychain_txout_changeset_init(db)?),
-        |db, changeset| Ok(keychain_txout_changeset_persist(db, changeset)?),
+        |db| {
+            let db_tx = db.transaction()?;
+            keychain_txout::ChangeSet::init_sqlite_tables(&db_tx)?;
+            let changeset = keychain_txout::ChangeSet::from_sqlite(&db_tx)?;
+            db_tx.commit()?;
+            Ok(changeset)
+        },
+        |db, changeset| {
+            let db_tx = db.transaction()?;
+            changeset.persist_to_sqlite(&db_tx)?;
+            db_tx.commit()?;
+            Ok(())
+        },
+        &changesets,
     )?)
 }
 
 #[test]
 fn local_chain_is_persisted() -> anyhow::Result<()> {
     let temp_dir = tempfile::tempdir().unwrap();
-    Ok(persist_local_chain_changeset::<
-        rusqlite::Connection,
-        _,
-        _,
-        _,
-    >(
+    let changesets = local_chain_changesets();
+    Ok(assert_persist_changesets(
         || {
             Ok(rusqlite::Connection::open(
                 temp_dir.path().join("wallet.sqlite"),
             )?)
         },
-        |db| Ok(local_chain_changeset_init(db)?),
-        |db, changeset| Ok(local_chain_changeset_persist(db, changeset)?),
+        |db| {
+            let db_tx = db.transaction()?;
+            local_chain::ChangeSet::init_sqlite_tables(&db_tx)?;
+            let changeset = local_chain::ChangeSet::from_sqlite(&db_tx)?;
+            db_tx.commit()?;
+            Ok(changeset)
+        },
+        |db, changeset| {
+            let db_tx = db.transaction()?;
+            changeset.persist_to_sqlite(&db_tx)?;
+            db_tx.commit()?;
+            Ok(())
+        },
+        &changesets,
     )?)
 }
--- a/crates/chain/tests/test_rusqlite_impl.rs
+++ b/crates/chain/tests/test_rusqlite_impl.rs
@@ -1,4 +1,5 @@
 #![cfg(feature = "rusqlite")]
+use anyhow::anyhow;
 use bdk_chain::{keychain_txout, local_chain, tx_graph, ConfirmationBlockTime};
 use bdk_testenv::persist_test_utils::{
     assert_persist_changesets, keychain_txout_changesets, local_chain_changesets,
@@ -9,17 +10,17 @@ use bdk_testenv::persist_test_utils::{
 fn txgraph_is_persisted() -> anyhow::Result<()> {
     let temp_dir = tempfile::tempdir().unwrap();
     let changesets = tx_graph_changesets();
-    Ok(assert_persist_changesets(
+    assert_persist_changesets(
         || {
-            Ok(rusqlite::Connection::open(
-                temp_dir.path().join("wallet.sqlite"),
-            )?)
+            let mut db = rusqlite::Connection::open(temp_dir.path().join("wallet.sqlite"))?;
+            let db_tx = db.transaction()?;
+            tx_graph::ChangeSet::<ConfirmationBlockTime>::init_sqlite_tables(&db_tx)?;
+            db_tx.commit()?;
+            Ok(db)
         },
         |db| {
             let db_tx = db.transaction()?;
-            tx_graph::ChangeSet::<ConfirmationBlockTime>::init_sqlite_tables(&db_tx)?;
             let changeset = tx_graph::ChangeSet::<ConfirmationBlockTime>::from_sqlite(&db_tx)?;
-            db_tx.commit()?;
             Ok(changeset)
         },
         |db, changeset| {
@@ -29,24 +30,25 @@ fn txgraph_is_persisted() -> anyhow::Result<()> {
             Ok(())
         },
         &changesets,
-    )?)
+    )
+    .map_err(|err| anyhow!(err))
 }
 
 #[test]
 fn indexer_is_persisted() -> anyhow::Result<()> {
     let temp_dir = tempfile::tempdir().unwrap();
     let changesets = keychain_txout_changesets();
-    Ok(assert_persist_changesets(
+    assert_persist_changesets(
         || {
-            Ok(rusqlite::Connection::open(
-                temp_dir.path().join("wallet.sqlite"),
-            )?)
+            let mut db = rusqlite::Connection::open(temp_dir.path().join("wallet.sqlite"))?;
+            let db_tx = db.transaction()?;
+            keychain_txout::ChangeSet::init_sqlite_tables(&db_tx)?;
+            db_tx.commit()?;
+            Ok(db)
         },
         |db| {
             let db_tx = db.transaction()?;
-            keychain_txout::ChangeSet::init_sqlite_tables(&db_tx)?;
             let changeset = keychain_txout::ChangeSet::from_sqlite(&db_tx)?;
-            db_tx.commit()?;
             Ok(changeset)
         },
         |db, changeset| {
@@ -56,24 +58,25 @@ fn indexer_is_persisted() -> anyhow::Result<()> {
             Ok(())
         },
         &changesets,
-    )?)
+    )
+    .map_err(|err| anyhow!(err))
 }
 
 #[test]
 fn local_chain_is_persisted() -> anyhow::Result<()> {
     let temp_dir = tempfile::tempdir().unwrap();
     let changesets = local_chain_changesets();
-    Ok(assert_persist_changesets(
+    assert_persist_changesets(
         || {
-            Ok(rusqlite::Connection::open(
-                temp_dir.path().join("wallet.sqlite"),
-            )?)
+            let mut db = rusqlite::Connection::open(temp_dir.path().join("wallet.sqlite"))?;
+            let db_tx = db.transaction()?;
+            local_chain::ChangeSet::init_sqlite_tables(&db_tx)?;
+            db_tx.commit()?;
+            Ok(db)
         },
         |db| {
             let db_tx = db.transaction()?;
-            local_chain::ChangeSet::init_sqlite_tables(&db_tx)?;
             let changeset = local_chain::ChangeSet::from_sqlite(&db_tx)?;
-            db_tx.commit()?;
             Ok(changeset)
         },
         |db, changeset| {
@@ -83,5 +86,6 @@ fn local_chain_is_persisted() -> anyhow::Result<()> {
             Ok(())
         },
         &changesets,
-    )?)
+    )
+    .map_err(|err| anyhow!(err))
 }
--- a/crates/chain/tests/test_rusqlite_impl.rs
+++ b/crates/chain/tests/test_rusqlite_impl.rs
@@ -1,10 +1,7 @@
 #![cfg(feature = "rusqlite")]
 use anyhow::anyhow;
 use bdk_chain::{keychain_txout, local_chain, tx_graph, ConfirmationBlockTime};
-use bdk_testenv::persist_test_utils::{
-    assert_persist_changesets, keychain_txout_changesets, local_chain_changesets,
-    tx_graph_changesets,
-};
+use bdk_testenv::persist_test_utils::*;
 
 #[test]
 fn txgraph_is_persisted() -> anyhow::Result<()> {
@@ -35,6 +32,7 @@ fn txgraph_is_persisted() -> anyhow::Result<()> {
 }
 
 #[test]
+#[cfg(feature = "miniscript")]
 fn indexer_is_persisted() -> anyhow::Result<()> {
     let temp_dir = tempfile::tempdir().unwrap();
     let changesets = keychain_txout_changesets();
--- a/crates/file_store/Cargo.toml
+++ b/crates/file_store/Cargo.toml
@@ -21,6 +21,6 @@ serde = { version = "1", features = ["derive"] }
 
 [dev-dependencies]
 tempfile = "3"
-anyhow = { version = "1.0.102", default-features = false}
+anyhow = { version = "1", default-features = false}
 bdk_testenv = {path = "../testenv"}
 bdk_chain = { path = "../chain", version = "0.23.1", default-features = false, features = ["serde"]}
--- a/crates/file_store/Cargo.toml
+++ b/crates/file_store/Cargo.toml
@@ -21,3 +21,6 @@ serde = { version = "1", features = ["derive"] }
 
 [dev-dependencies]
 tempfile = "3"
+anyhow = { version = "1.0.102", default-features = false}
+bdk_testenv = {path = "../testenv"}
+bdk_chain = { path = "../chain", version = "0.23.1", default-features = false, features = ["serde"]}
--- a/crates/file_store/Cargo.toml
+++ b/crates/file_store/Cargo.toml
@@ -22,5 +22,4 @@ serde = { version = "1", features = ["derive"] }
 [dev-dependencies]
 tempfile = "3"
 anyhow = { version = "1", default-features = false}
-bdk_testenv = {path = "../testenv"}
-bdk_chain = { path = "../chain", version = "0.23.1", default-features = false, features = ["serde"]}
+bdk_testenv = { path = "../testenv", features = ["serde"] } # `miniscript` is active by default.
--- a/crates/file_store/src/store.rs
+++ b/crates/file_store/src/store.rs
@@ -295,9 +295,9 @@ mod test {
     const TEST_MAGIC_BYTES: [u8; TEST_MAGIC_BYTES_LEN] =
         [98, 100, 107, 102, 115, 49, 49, 49, 49, 49, 49, 49];
 
-    use bdk_chain::{keychain_txout, local_chain, tx_graph, ConfirmationBlockTime};
     use bdk_testenv::persist_test_utils::{
-        persist_indexer_changeset, persist_local_chain_changeset, persist_txgraph_changeset,
+        assert_persist_changesets, keychain_txout_changesets, local_chain_changesets,
+        tx_graph_changesets,
     };
 
     type TestChangeSet = BTreeSet<String>;
@@ -608,12 +608,8 @@ mod test {
     #[test]
     fn txgraph_is_persisted() -> anyhow::Result<()> {
         let temp_dir = tempfile::tempdir().unwrap();
-        Ok(persist_txgraph_changeset::<
-            Store<tx_graph::ChangeSet<ConfirmationBlockTime>>,
-            _,
-            _,
-            _,
-        >(
+        let changesets = tx_graph_changesets();
+        Ok(assert_persist_changesets(
             || {
                 Ok(Store::create(
                     &TEST_MAGIC_BYTES,
@@ -622,18 +618,15 @@ mod test {
             },
             |db| Ok(db.dump().map(Option::unwrap_or_default)?),
             |db, changeset| Ok(db.append(changeset)?),
+            &changesets,
         )?)
     }
 
     #[test]
     fn indexer_is_persisted() -> anyhow::Result<()> {
         let temp_dir = tempfile::tempdir().unwrap();
-        Ok(persist_indexer_changeset::<
-            Store<keychain_txout::ChangeSet>,
-            _,
-            _,
-            _,
-        >(
+        let changesets = keychain_txout_changesets();
+        Ok(assert_persist_changesets(
             || {
                 Ok(Store::create(
                     &TEST_MAGIC_BYTES,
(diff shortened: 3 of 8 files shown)


ValuedMammal:
I tried re-running CI to no avail, maybe try force pushing.

evanlinjin:
Great idea. ConceptACK

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,721 @@
...
+
+    let mut tx_graph_changeset1 = ChangeSet::<ConfirmationBlockTime> {
+        txs: [tx1.clone()].into(),
+        txouts: [
+            (
+                OutPoint::new(hash!("BDK"), 0),
+                TxOut {
+                    value: Amount::from_sat(1300),
+                    script_pubkey: Address::from_str(
+                        "bcrt1q8an5jfmpq8w2hr648nn34ecf9zdtxk0qyqtrfl",
```
oleonardolima: You can extract the address into a constant as it's being used multiple times.

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,721 @@
...
+        }],
+        output: vec![TxOut {
+            value: Amount::from_sat(amount),
+            script_pubkey: Address::from_str("bcrt1q3qtze4ys45tgdvguj66zrk4fu6hq3a3v9pfly5")
+                .unwrap()
+                .assume_checked()
+                .script_pubkey(),
+        }],
+    }
+}
```
oleonardolima: I wonder if this one should live in utils.rs instead ? Also, maybe it could be generalized for n inputs, n outputs, as it'd be reusable on other scenarios.

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,721 @@
...
+
+fn spk_at_index(descriptor: &Descriptor<DescriptorPublicKey>, index: u32) -> ScriptBuf {
+    descriptor
+        .derived_descriptor(&Secp256k1::verification_only(), index)
+        .expect("must derive")
+        .script_pubkey()
+}
```
oleonardolima: Same here, I guess it could live under `utils.rs`

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,721 @@
...
+
+    assert_eq!(tx_graph_changeset1, changeset);
+}
+
+fn parse_descriptor(descriptor: &str) -> Descriptor<DescriptorPublicKey> {
+    let secp = bdk_chain::bitcoin::secp256k1::Secp256k1::signing_only();
+    Descriptor::<DescriptorPublicKey>::parse_descriptor(&secp, descriptor)
+        .unwrap()
+        .0
+}
```
oleonardolima: Same here, I think it could live in `utils.rs`.

crates/testenv/src/persist_test_utils.rs
110CodingP: Done in [b77a114](https://github.com/bitcoindevkit/bdk/pull/2012/commits/b77a1141caa8127facbea5a302b698d7d065c0e4)!

110CodingP:
In [10c42dc](https://github.com/bitcoindevkit/bdk/pull/2012/commits/10c42dcd93973bb99960426524b654687cd26a19)  I added a `miniscript` feature to `testenv` but I am not sure if this is the correct approach. Earlier I was including the `miniscript` feature of `bdk_chain` by default but that doesn't work since `miniscript` requires either `std` or `no-std` but cargo doesn't allow specifying the feature of a dependency of a dependency. Adding `miniscript` as a separate dependency also does not work since we require `bdk_chain/miniscript`. But the current solution is also not satisfactory since `miniscript` feature requires the `std` feature!

ValuedMammal:
> In [10c42dc](https://github.com/bitcoindevkit/bdk/pull/2012/commits/10c42dcd93973bb99960426524b654687cd26a19) I added a `miniscript` feature to `testenv` but I am not sure if this is the correct approach. Earlier I was including the `miniscript` feature of `bdk_chain` by default but that doesn't work since `miniscript` requires either `std` or `no-std` but cargo doesn't allow specifying the feature of a dependency of a dependency. Adding `miniscript` as a separate dependency also does not work since we require `bdk_chain/miniscript`. But the current solution is also not satisfactory since `miniscript` feature requires the `std` feature!

We could have the testenv default feature enable `bdk_chain/default`, which in turn enables `miniscript/std`. I'm not aware of anyone needing a `miniscript/no-std` feature for the testenv crate.

110CodingP:
Should the miniscript feature introduced be removed? I was thinking if some users might not want bdk_chain/miniscript feature to be enabled?

110CodingP:
`miniscript` feature still depends on `std` feature of `testenv` since ig it would be confusing otherwise like when `miniscript` is enabled but not `std` even though `miniscript` would activate `bdk_chain/std`.  Made `miniscript` a default feature too. Sorry if I am overthinking :sweat_smile: .

crates/testenv/Cargo.toml
ValuedMammal, line 1: I feel like the correct thing to do is something like this by making the dependency explicit.

```toml
[dependencies]
miniscript = { version = "12.3.5", default-features = false, optional = true }

[features]
default = ["std" , "download", "bdk_chain/default"]
miniscript = ["dep:miniscript"]
std = ["bdk_chain/std", "miniscript?/std"]
serde = ["bdk_chain/serde", "miniscript?/serde"]
```

But a simpler hack is to just have `default = ["bdk_chain/default"]` (which seems to be missing and should probably be done anyway). In that case the cfg attribute would be changed to `#[cfg(feature = "default"]`. Sorry to cause more bikeshedding on the matter. I'll be in favor of whatever is clear and easy to understand.

notmandatory:
~~Since this is currently based on the `master` branch should we re-assign this to the 3.0 milestone and then do a new PR to back port it to 2.2?~~ Nevermind, this is for the `bdk` crate so my question doesn't apply.

crates/testenv/Cargo.toml
110CodingP: Went with the second approach since the first seemed a bit complicated :sweat_smile: (like the test utils are gated by miniscript feature which might not be enabled even though all we need is bdk_chain/default).

crates/chain/tests/test_rusqlite_impl.rs
```diff
@@ -0,0 +1,227 @@
...
+fn txgraph_is_persisted() {
+    persist_txgraph_changeset::<rusqlite::Connection, _, _, _>(
+        "wallet.sqlite",
+        |path| Ok(bdk_chain::rusqlite::Connection::open(path)?),
+        |db| {
+            let db_tx = db.transaction()?;
+            tx_graph::ChangeSet::<ConfirmationBlockTime>::init_sqlite_tables(&db_tx)?;
+            let changeset = tx_graph::ChangeSet::<ConfirmationBlockTime>::from_sqlite(&db_tx)?;
+            db_tx.commit()?;
+            Ok(changeset)
```
notmandatory: Can this be extracted into a function rather than being duplicated?

crates/chain/tests/test_rusqlite_impl.rs
```diff
@@ -0,0 +1,227 @@
...
+        },
+        |db, changeset| {
+            let db_tx = db.transaction()?;
+            changeset.persist_to_sqlite(&db_tx)?;
+            Ok(db_tx.commit()?)
```
notmandatory: Same question as above.

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,877 @@
...
+            ),
+        ]
+        .into(),
+        anchors: [(conf_anchor, tx1.compute_txid())].into(),
+        last_seen: [(tx1.compute_txid(), 1755416650)].into(),
+        first_seen: [(tx1.compute_txid(), 1755416655)].into(),
+        last_evicted: [(tx1.compute_txid(), 1755416660)].into(),
+    };
+
+    // persist and load
```
notmandatory: ⛏️ Docs should follow the rust style guidelines with first line starts with a capital letter and ends with a period.

https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html

notmandatory (changes requested):
Looks good! I made a couple comments and it needs a rebase and then I think it's ready to go.

oleonardolima:
I moved this to chain-0.24.0 milestone for visibility, @110CodingP let us know when it's ready for another round of review.

crates/testenv/src/persist_test_utils.rs
110CodingP: Sorry for the delay. Done in [70a3662](https://github.com/bitcoindevkit/bdk/pull/2012/commits/70a36629d18418355fe03b01d1d6ef7c101c3927).

crates/chain/tests/test_rusqlite_impl.rs
110CodingP: Done in [6f09bde](https://github.com/bitcoindevkit/bdk/pull/2012/commits/6f09bde642b2b829b1abf9024807ea588d7c8262).

110CodingP:
Since we use v13 of `miniscript` now, I finally added a `miniscript` feature to `testenv` and made it a default feature. The test-utils are now gated by `miniscript` and `std` features of `testenv`.

notmandatory:
This PR will be important to have if we still want to do #1980 to help test persistence implementations.

crates/chain/tests/test_rusqlite_impl.rs
```diff
@@ -0,0 +1,165 @@
...
+    let db_tx = db.transaction()?;
+    changeset.persist_to_sqlite(&db_tx)?;
+    db_tx.commit()
+}
+
+#[test]
+fn txgraph_is_persisted() {
+    persist_txgraph_changeset::<rusqlite::Connection, _, _, _>(
+        "wallet.sqlite",
+        |path| Ok(bdk_chain::rusqlite::Connection::open(path)?),
```
tvpeter, line 51: nit: Is there a reason you are using different connection types? from `rusqlite::Connection` and `bdk_chain::rusqlite::Connection`

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,878 @@
...
+) where
+    CreateStore: Fn(&Path) -> anyhow::Result<Store>,
+    Initialize: Fn(&mut Store) -> anyhow::Result<tx_graph::ChangeSet<ConfirmationBlockTime>>,
+    Persist: Fn(&mut Store, &tx_graph::ChangeSet<ConfirmationBlockTime>) -> anyhow::Result<()>,
+{
+    use tx_graph::ChangeSet;
+    // Create the store.
+    let temp_dir = tempfile::tempdir().expect("must create tempdir");
+    let file_path = temp_dir.path().join(file_name);
+    let mut store = create_store(&file_path).expect("store should get created");
```
tvpeter, line 446: nit: seemed like every fn `creates the store`, could you refactor it into a separate fn?

tvpeter (commented):
Hi @110CodingP , this is incredible work on a massive PR.

I have left some `nit` comments.

In addition to the comments, I think it will also be great if you could consider including test utils for handling edge cases/conflicts, especially in the `persist_test_utils` (if that will not make the PR too big)

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,878 @@
...
+                OutPoint::new(hash!("Bitcoin_fixes_things"), 0),
+                create_txout(1400, ADDRS[1]),
+            ),
+        ]
+        .into(),
+        anchors: [(conf_anchor, tx1.compute_txid())].into(),
+        last_seen: [(tx1.compute_txid(), 1755416650)].into(),
+        first_seen: [(tx1.compute_txid(), 1755416655)].into(),
+        last_evicted: [(tx1.compute_txid(), 1755416660)].into(),
+    };
```
nymius, line 77: Here we are also testing the persistence of all the other changesets. Being such a high level test, Is it worth it to have the granularity we have, i.e., a test for each of the smaller changesets?

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,878 @@
...
+
+    // Create another changeset.
+    let tx2 = Arc::new(create_test_tx(
+        [tx1.compute_txid()],
+        [0],
+        [20_000],
+        [ADDRS[0]],
+        1,
+        0,
+    ));
```
nymius, line 93: https://github.com/rust-fuzz/arbitrary could help remove work here, at the same time is exposing the API for fuzzers and other tests.

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,878 @@
...
+    file_name: &str,
+    create_store: CreateStore,
+    initialize: Initialize,
+    persist: Persist,
+) where
+    CreateStore: Fn(&Path) -> anyhow::Result<Store>,
+    Initialize: Fn(&mut Store) -> anyhow::Result<local_chain::ChangeSet>,
+    Persist: Fn(&mut Store, &local_chain::ChangeSet) -> anyhow::Result<()>,
+{
+    use local_chain::ChangeSet;
```
nymius, line 211: Have you explored ways to abstract this? The structure of all tests is the same, and although I know each changeset is different, even an enum wrapping the changesets could improve the current situation.

crates/testenv/src/utils.rs
```diff
@@ -77,6 +83,71 @@ pub fn new_tx(lt: u32) -> bitcoin::Transaction {
...
+    assert!(version.is_standard());
+    let lock_time = absolute::LockTime::from_consensus(locktime);
+    assert_eq!(lock_time.to_consensus_u32(), locktime);
+    Transaction {
+        version,
+        lock_time,
+        input: input_vec,
+        output: output_vec,
+    }
+}
```
nymius, line 128: Again, https://github.com/rust-fuzz/arbitrary could help with this.

nymius (commented):
cACK 6f09bde6

110CodingP:
@tvpeter and @nymius thank you for the reviews! I will restructure the PR to address the concerns. Also looking at [mammal's restructuring of the wallet tests](https://github.com/bitcoindevkit/bdk_wallet/pull/343/changes/5c19c41c64f7c2babc80dd114390602fd906c45b).
Will also try adding tests for the edge cases, the ones I can think of right now are the following:
1. persisting duplicate changesets should have no effect.
2. persisting empty changeset.

crates/testenv/src/persist_test_utils.rs
110CodingP: I agree they are redundant. The reason I was doing it this way was that users will then be able to use these redundant tests as unit tests in case they have separate functions to persist each of the smaller changesets. I would be fine with removing them too if that increases maintainability.

evanlinjin (changes requested):
### Concept ACK

Thanks for working on this. Having some sort of tooling to ensure persistence implementations is not only helpful, but also crucial if we ever want to do #1980.

### Approach NACK

There is quite a lot of boilerplate happening here and lots of repeated code making this approach hard to read and hard to maintain. All the test methods essentially all do the same thing, but with different changeset types. As mentioned by @nymius [here](https://github.com/bitcoindevkit/bdk/pull/2012/changes#r3079095018), we should really abstract this.

### Proposal

We can represent all that we want to test with a single generic helper. Then just have predefined sets of changesets (`C`).

```rust
pub fn assert_persists<C, S, CREATE, LOAD, PERSIST>(
    create: CREATE,
    load: LOAD,
    persist: PERSIST,
    changesets: &[C],
) where
      C: Default + Merge + PartialEq + Clone + core::fmt::Debug,
      CREATE: Fn(&Path) -> anyhow::Result<S>,
      LOAD: Fn(&mut S) -> anyhow::Result<C>,
      PERSIST: Fn(&mut S, &C) -> anyhow::Result<()>,
  {
      // Persist one at a time, reload after each, asserting merged accumulation.
      let dir = tempfile::tempdir().expect("tempdir");
      let mut store = create(dir.path()).expect("interleaved: create store");
      assert_eq!(
          load(&mut store).expect("interleaved: load empty"),
          C::default(),
      );

      let mut acc = C::default();
      for cs in changesets {
          persist(&mut store, cs).expect("interleaved: persist");
          acc.merge(cs.clone());
          assert_eq!(
              load(&mut store).expect("interleaved: reload"),
              acc,
          );
      }

      // Persist all into a fresh store, reload once.
      let dir = tempfile::tempdir().expect("tempdir");
      let mut store = create(dir.path()).expect("batched: create store");
      for cs in changesets {
          persist(&mut store, cs).expect("batched: persist");
      }
      assert_eq!(
          load(&mut store).expect("batched: reload"),
          acc,
      );
  }
```

Predefined test changesets via builders:

```rust
pub fn tx_graph_fixture() -> [tx_graph::ChangeSet<ConfirmationBlockTime>; 2] { … }
pub fn keychain_txout_fixture() -> [keychain_txout::ChangeSet; 2] { … }
pub fn local_chain_fixture() -> [local_chain::ChangeSet; 2] { … }
```

For fuzzing, proptest, we can even create methods that generates changeset sets on the fly.

110CodingP:
Thanks a lot @evanlinjin for the review! 
I implemented them in [48912ee](https://github.com/bitcoindevkit/bdk/pull/2012/commits/48912eee0e3a1a5ec6394c2dc271e566fe8296b4). The only major difference is the error handling part which is inspired from [bdk_wallet#343](https://github.com/bitcoindevkit/bdk_wallet/pull/343). 

Also for convenience the PR still contains changeset specific persistence tests ( `persist_txgraph_changeset`). 

I will try to leverage `arbitrary` and add edge cases in later commits in this PR.

crates/chain/tests/test_rusqlite_impl.rs
evanlinjin, line 1: There is a whole bunch of functions in this file that is only called once. Let's just inline it - this will improve readability and reduce lines of code.

crates/chain/tests/test_rusqlite_impl.rs
```diff
@@ -0,0 +1,109 @@
...
+) -> rusqlite::Result<()> {
+    let db_tx = db.transaction()?;
+    changeset.persist_to_sqlite(&db_tx)?;
+    db_tx.commit()
+}
+
+#[test]
+fn txgraph_is_persisted() -> anyhow::Result<()> {
+    let temp_dir = tempfile::tempdir().unwrap();
+    Ok(persist_txgraph_changeset::<rusqlite::Connection, _, _, _>(
```
evanlinjin: Nit: The explicit turbofishing is not really needed for any of these functions.

```suggestion
    Ok(persist_txgraph_changeset(
```

crates/file_store/src/store.rs
```diff
@@ -599,4 +604,64 @@ mod test {
...
+
+    #[test]
+    fn txgraph_is_persisted() -> anyhow::Result<()> {
+        let temp_dir = tempfile::tempdir().unwrap();
+        Ok(persist_txgraph_changeset::<
+            Store<tx_graph::ChangeSet<ConfirmationBlockTime>>,
+            _,
+            _,
+            _,
+        >(
```
```suggestion
        Ok(persist_txgraph_changeset(
```

Turbofishing not necessary.

crates/file_store/Cargo.toml
```diff
@@ -21,3 +21,6 @@ serde = { version = "1", features = ["derive"] }
 
 [dev-dependencies]
 tempfile = "3"
+anyhow = { version = "1.0.102", default-features = false}
```
evanlinjin:
```suggestion
anyhow = { version = "1", default-features = false}
```

No need to explicitly define the minor and patch versions.

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+impl<C: Debug> Err for PersistErr<C> {}
+
+/// Tests if `ChangeSet` is being persisted correctly.
+///
+/// We create a dummy `ChangeSet`, persist it and check if loaded `ChangeSet` matches
+/// the persisted one. We then create another such dummy `ChangeSet`, persist it and load it to
+/// check if merged `ChangeSet` is returned.
+pub fn persist_changeset<CS, Store, CreateStore, Initialize, Persist>(
```
evanlinjin: `persist_changeset` reads like "perform persistence" but this is a persistence-assertion-helper.

```suggestion
pub fn assert_persist_changesets<CS, Store, CreateStore, Initialize, Persist>(
```

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,308 @@
...
+        first_seen: [(tx2.compute_txid(), 1755416670)].into(),
+        last_evicted: [(tx2.compute_txid(), 1755416760)].into(),
+    };
+
+    [tx_graph_changeset1, tx_graph_changeset2]
+}
+
+/// Get two [`keychain_txout::ChangeSet`](keychain_txout::ChangeSet)s.
+#[cfg(feature = "miniscript")]
+fn keychain_txout_changesets() -> [keychain_txout::ChangeSet; 2] {
```
evanlinjin: Make these methods that return fixtures of changesets pub. Then we can remove the `persist_xxx_changeset` methods. Much less code to maintain.

evanlinjin (changes requested):
Much better already.

evanlinjin:
@110CodingP How about we focus this PR on getting the API shape correct and introduce something like `arbitrary` in a separate PR?

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,308 @@
...
+    /// ChangeSet Mismatch
+    ChangeSetMismatch {
+        /// the resulting changeset
+        got: Box<C>,
+        /// the expected changeset
+        expected: Box<C>,
+    },
+    /// Errors thrown by underlying persistence backend.
+    Persister(Box<dyn Err + 'static + Send + Sync>),
+}
```
evanlinjin: This is overengineered. We don't need to programmatically inspect different failure types. These helpers are for tests - failure just needs to `panic!`/`assert!`.

`assert_persist_changesets` should not need a return type and just panic on failure.

evanlinjin (changes requested):

110CodingP:
> @110CodingP How about we focus this PR on getting the API shape correct and introduce something like `arbitrary` in a separate PR?

Yup, agreed!

crates/testenv/src/persist_test_utils.rs
110CodingP: The reason I changed to this approach is that
1.) We have more control on the error message in this case.
2.) I was initially thinking that throwing a panic! would help users identify the line number in the tests that caused the panic but then we can't expect the users to go into the library code right?
3.) Users can downcast the error and handle it accordingly?

crates/chain/tests/test_rusqlite_impl.rs
110CodingP: Done in [75fa86d](https://github.com/bitcoindevkit/bdk/pull/2012/commits/75fa86d42597681e2fab4a5d6fa177cd722ab699).

crates/chain/tests/test_rusqlite_impl.rs
110CodingP: Done in [75fa86d](https://github.com/bitcoindevkit/bdk/pull/2012/commits/75fa86d42597681e2fab4a5d6fa177cd722ab699)

crates/file_store/src/store.rs
110CodingP: Done in [75fa86d](https://github.com/bitcoindevkit/bdk/pull/2012/commits/75fa86d42597681e2fab4a5d6fa177cd722ab699).

crates/testenv/src/persist_test_utils.rs
110CodingP: Done in [75fa86d](https://github.com/bitcoindevkit/bdk/pull/2012/commits/75fa86d42597681e2fab4a5d6fa177cd722ab699).

crates/testenv/src/persist_test_utils.rs
110CodingP: Done in [75fa86d](https://github.com/bitcoindevkit/bdk/pull/2012/commits/75fa86d42597681e2fab4a5d6fa177cd722ab699).

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,223 @@
...
+impl<C: Debug> Err for PersistErr<C> {}
+
+/// Tests if `ChangeSet` is being persisted correctly.
+///
+/// We create a dummy `ChangeSet`, persist it and check if loaded `ChangeSet` matches
+/// the persisted one. We then create another such dummy `ChangeSet`, persist it and load it to
+/// check if merged `ChangeSet` is returned.
+pub fn assert_persist_changesets<CS, Store, CreateStore, Initialize, Persist>(
+    create_store: CreateStore,
+    initialize: Initialize,
```
evanlinjin:
```suggestion
    load: Load,
```

This makes it easier to reason about. As "loading from db" is the main operation we care about.

`initialize(&mut store)` reads like we are setting up the store, but then it returns a changeset which is confusing.

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,223 @@
...
+    let mut store = create_store()?;
+
+    let init_changeset = initialize(&mut store)?;
+
+    if init_changeset != CS::default() {
+        return Err(PersistErr::ChangeSetMismatch {
+            expected: Box::new(CS::default()),
+            got: Box::new(init_changeset),
+        });
+    }
```
evanlinjin: The returned error does not explicitly convey what we are testing here. This is proof that `PersistErr` is not earning it's keep.

The returned error should state what was being tested. In this case - a newly created store should load an empty changeset.

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    let mut merged_changeset = CS::default();
+
+    for changeset in changesets {
+        persist(&mut store, changeset)?;
```
evanlinjin: It will be helpful to know which changeset index we failed on alongside the db error.

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,223 @@
...
+        merged_changeset.merge(changeset.clone());
```
evanlinjin, line 77: I proposed loading from the db state after each merge - the reasoning is so that we can catch persistence problems earlier. I think it's a good idea to have this.

crates/testenv/src/persist_test_utils.rs
evanlinjin: Thanks for presenting the arguments, I have changed my stance.
 
1. Yes, however we can also just return a `String` as the error type. Currently, the error variants are missing context that are useful to have. We are reusing the `ChangeSetMismatch` for different test errors where it would be more useful to have context-specific error messages.

2. You are right about this - `Result` is better.

3. YAGNI. I don't see a usecase for this?

### My updated recommendation

Return `Result<(), String>` or `anyhow::Result<()>` where the error string gives context of the error (check my other review comments for details).

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,223 @@
...
+
+impl<C: Debug> Err for PersistErr<C> {}
+
+/// Tests if `ChangeSet` is being persisted correctly.
+///
+/// We create a dummy `ChangeSet`, persist it and check if loaded `ChangeSet` matches
+/// the persisted one. We then create another such dummy `ChangeSet`, persist it and load it to
+/// check if merged `ChangeSet` is returned.
+pub fn assert_persist_changesets<CS, Store, CreateStore, Initialize, Persist>(
+    create_store: CreateStore,
```
evanlinjin:
```suggestion
    init: Init,
```

`CreateStore` conveys "make a brand new store". However, this closure's job is actually "give me a db handle", or in other words "Open or create".

Another note is that this test does not actually test a reopen. I propose having three sections to this test:

1. Load cs from a newly created db. The cs should be empty. 
2. Interweave cs persists and loads. Ensure persists succeeds and loaded changesets are expected.
3. Reopen the db, load cs from db. Ensure cs is expected.

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+impl<C: Debug> Err for PersistErr<C> {}
+
+/// Tests if `ChangeSet` is being persisted correctly.
+///
+/// We create a dummy `ChangeSet`, persist it and check if loaded `ChangeSet` matches
+/// the persisted one. We then create another such dummy `ChangeSet`, persist it and load it to
+/// check if merged `ChangeSet` is returned.
```
evanlinjin: This does not look correct. `changesets` are provided to this method, not created by this method.

We should mention what exactly is being tested. I.e. load from newly created db, interleaved persists and loads of changesets, load after reopen should succeed.

We should also explain the responsibilities of `Init`, `Load` and `Persist`.

crates/chain/tests/test_rusqlite_impl.rs
```diff
@@ -0,0 +1,87 @@
...
+    let changesets = keychain_txout_changesets();
+    Ok(assert_persist_changesets(
+        || {
+            Ok(rusqlite::Connection::open(
+                temp_dir.path().join("wallet.sqlite"),
+            )?)
+        },
+        |db| {
+            let db_tx = db.transaction()?;
+            keychain_txout::ChangeSet::init_sqlite_tables(&db_tx)?;
```
evanlinjin: Let's call `init_sqlite_tables` AFTER opening the connection. Not here.

evanlinjin (changes requested):

crates/testenv/src/persist_test_utils.rs
110CodingP: Sorry I forgot to update the docs. Done in de27ded .

crates/chain/tests/test_rusqlite_impl.rs
```diff
@@ -0,0 +1,91 @@
+#![cfg(feature = "rusqlite")]
+use anyhow::anyhow;
+use bdk_chain::{keychain_txout, local_chain, tx_graph, ConfirmationBlockTime};
+use bdk_testenv::persist_test_utils::{
+    assert_persist_changesets, keychain_txout_changesets, local_chain_changesets,
+    tx_graph_changesets,
+};
```
evanlinjin:
```suggestion
use bdk_testenv::persist_test_utils::*;
```

`keychain_txout_changesets` requires `feature = miniscript`.

crates/chain/tests/test_rusqlite_impl.rs
```diff
@@ -0,0 +1,91 @@
...
+            changeset.persist_to_sqlite(&db_tx)?;
+            db_tx.commit()?;
+            Ok(())
+        },
+        &changesets,
+    )
+    .map_err(|err| anyhow!(err))
+}
+
+#[test]
```
evanlinjin, line 34:
```suggestion
#[test]
#[cfg(feature = "miniscript")]
```

crates/file_store/src/store.rs
```diff
@@ -291,10 +291,17 @@ mod test {
...
+
     const TEST_MAGIC_BYTES_LEN: usize = 12;
     const TEST_MAGIC_BYTES: [u8; TEST_MAGIC_BYTES_LEN] =
         [98, 100, 107, 102, 115, 49, 49, 49, 49, 49, 49, 49];
 
+    use bdk_testenv::persist_test_utils::{
+        assert_persist_changesets, keychain_txout_changesets, local_chain_changesets,
+        tx_graph_changesets,
+    };
+
```
```suggestion
    use bdk_testenv::persist_test_utils::*;

```

crates/file_store/src/store.rs
```diff
@@ -599,4 +606,43 @@ mod test {
...
+        assert_persist_changesets(
+            || Ok(Store::load_or_create(&TEST_MAGIC_BYTES, temp_dir.path().join("store.db"))?.0),
+            |db| Ok(db.dump().map(Option::unwrap_or_default)?),
+            |db, changeset| Ok(db.append(changeset)?),
+            &changesets,
+        )
+        .map_err(|err| anyhow!(err))
+    }
+
+    #[test]
```
evanlinjin, line 620:
```suggestion
    #[test]
    #[cfg(feature = "miniscript")]
```

crates/testenv/src/utils.rs
```diff
@@ -77,6 +83,71 @@ pub fn new_tx(lt: u32) -> bitcoin::Transaction {
...
+/// Utility function to create a transaction given txids, vouts of inputs and amounts (in satoshis),
+/// addresses of outputs.
+///
+/// The locktime should be in the form given to `OP_CHEKCLOCKTIMEVERIFY`.
+pub fn create_test_tx(
+    txids: impl IntoIterator<Item = Txid>,
+    vouts: impl IntoIterator<Item = u32>,
+    amounts: impl IntoIterator<Item = u64>,
+    addrs: impl IntoIterator<Item = &'static str>,
+    version: u32,
```
evanlinjin:
```suggestion
    version: transaction::Version,
```

This way, we can avoid this line:

```rust
    let version = transaction::Version::non_standard(version as i32);
```

and use the `Version::ONE`, `Version::TWO` constants as input.

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,216 @@
...
+
+/// A helper to check if a custom persistence backend persists `ChangeSet`s correctly.
+///
+/// This first tries to create a `Store` using `init`, `load`s from it and checks if
+/// the result is an empty `ChangeSet`.
+/// It then tries to `persist` each `ChangeSet` in `changesets` one by one, doing a `load`
+/// each time and checks that the aggregated `ChangeSet` matches the one loaded.
+///
+/// Finally it closes the `Store`, reopens it using `init`, `load`s from it and checks if the loaded
+/// `ChangeSet` matches the final aggregated `ChangeSet`.
```
evanlinjin: This narrates the implementation, but never explains to the caller what they need to get right when using this API. Docs should highlight the *closure contract*.

```suggestion
/// Asserts a persistence backend correctly stores and reloads a sequence of `ChangeSet`s.
///
/// Checks that `load` always returns the merge of everything `persist`ed so far - including after
/// the store is dropped and reopened. The backend is supplied as three closures:
///
/// - `init`: open-or-create a handle. This must never wipe existing data.
/// - `load`: return the full persisted changeset (`C::default()` on a fresh store).
/// - `persist`: durably write one changeset on top of what's stored; must survive the handle
///   being dropped.
///
/// Returns `Err` naming the failing step if a closure errors or a load doesn't match the expected
/// merge.
```

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,216 @@
...
+        let mut store = init().map_err(|err| {
+            format!(
+                "Encountered an error from the persister while initializing the store.\nGot:\n{}",
+                err
+            )
+        })?;
+
+        let init_changeset = load(&mut store).map_err(|err| format!("Encountered an error from the persister while loading from the new store.\nGot:\n{}",err))?;
+
+        if init_changeset != C::default() {
```
evanlinjin: Nit

```suggestion
        if !init_changeset.is_empty() {
```

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,216 @@
...
+    changesets: &[C],
+) -> Result<(), String>
+where
+    C: Debug + PartialEq + Default + Merge + Clone,
+    Init: Fn() -> Result<Store, Box<dyn Err>>,
+    Load: Fn(&mut Store) -> Result<C, Box<dyn Err>>,
+    Persist: Fn(&mut Store, &C) -> Result<(), Box<dyn Err>>,
+{
+    let mut merged_changeset = C::default();
+    {
```
evanlinjin:
```suggestion
```

Unnecessary block.

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,216 @@
...
+/// each time and checks that the aggregated `ChangeSet` matches the one loaded.
+///
+/// Finally it closes the `Store`, reopens it using `init`, `load`s from it and checks if the loaded
+/// `ChangeSet` matches the final aggregated `ChangeSet`.
+pub fn assert_persist_changesets<C, Store, Init, Load, Persist>(
+    init: Init,
+    load: Load,
+    persist: Persist,
+    changesets: &[C],
+) -> Result<(), String>
```
evanlinjin, line 42: Nit: Since every call site returns `anyhow::Result<()>` (and needs to map), I would also have this method return `anyhow::Result` for convenience!

evanlinjin (changes requested):
We are close!

* Compilation issues without `miniscript` feature.
* Incomplete docs.
* Various nits.

crates/file_store/Cargo.toml
```diff
@@ -21,3 +21,6 @@ serde = { version = "1", features = ["derive"] }
 
 [dev-dependencies]
 tempfile = "3"
+anyhow = { version = "1", default-features = false}
+bdk_testenv = {path = "../testenv"}
+bdk_chain = { path = "../chain", version = "0.23.1", default-features = false, features = ["serde"]}
```
noahjoeris:
```suggestion
bdk_testenv = { path = "../testenv", features = ["serde"] }
```

I think we don't need bdk_chain here

noahjoeris (commented):
LGTM 🚀 

Agree with Evan's last comments.

crates/testenv/src/persist_test_utils.rs
110CodingP: ~~Ok, so do you want me to use `drop` instead for dropping the store?~~ Got it, we don't need to drop the store.

crates/testenv/src/persist_test_utils.rs
110CodingP: On second thought, can we let it be as it is? Because if the application already uses `anyhow` but a different version then it would become very inconvenient for the developer?

crates/file_store/src/store.rs
110CodingP: On probing more I realised we don't have a `miniscript` feature in file-store, so this isn't needed?

crates/testenv/src/persist_test_utils.rs
110CodingP: I never thought of docs this way! Thanks!!! will keep in mind from now on.

110CodingP:
Ready for next review!

crates/testenv/src/persist_test_utils.rs
```diff
@@ -0,0 +1,228 @@
...
+        if persisted_changeset != merged_changeset {
+            Err(format!(
+                "Persisting changeset no. {} failed.\nExpected:\n\n{:?}\n\n\nLoaded:\n\n{:?};",
+                i + 1,
+                merged_changeset,
+                persisted_changeset
+            ))?;
+        }
+    }
+
```
tvpeter, line 90: I think you should consider expressly dropping the existing store before reopening on the next line
```suggestion
drop(store);
```

crates/testenv/src/utils.rs
```diff
@@ -77,6 +83,69 @@ pub fn new_tx(lt: u32) -> bitcoin::Transaction {
...
+            .unwrap()
+            .assume_checked()
+            .script_pubkey(),
+    }
+}
+
+/// Utility function to create a transaction given txids, vouts of inputs and amounts (in satoshis),
+/// addresses of outputs.
+///
+/// The locktime should be in the form given to `OP_CHEKCLOCKTIMEVERIFY`.
```
tvpeter, line 100:
```suggestion
/// The locktime should be in the form given to `OP_CHECKLOCKTIMEVERIFY`.
```

crates/testenv/src/utils.rs
```diff
@@ -77,6 +83,69 @@ pub fn new_tx(lt: u32) -> bitcoin::Transaction {
...
+        .map(|(txid, vout)| TxIn {
+            previous_output: OutPoint::new(txid, vout),
+            ..TxIn::default()
+        })
+        .collect();
+    let output_vec = core::iter::zip(amounts, addrs)
+        .map(|(amount, addr)| create_txout(amount, addr))
+        .collect();
+    let lock_time = absolute::LockTime::from_consensus(locktime);
+    assert_eq!(lock_time.to_consensus_u32(), locktime);
```
tvpeter, line 119: Won't this always return true?

tvpeter (commented):
Weldone @110CodingP,

I have left some minor comments.

crates/testenv/src/persist_test_utils.rs
evanlinjin: That's a good point. We should only have one instance of `store` at one time. Having multiple instances of `store` may cause problems for certain persistence implementations.

Another problem is alternating between loading and persisting (that I suggested - sorry). Since BDK treats persistence by reading once on startup and then append-only after, some persistence implementations may only work in that order. For now, lets remove the re-loading logic in the loop above.

evanlinjin (commented):
> **This review was generated by Claude Code** and posted from my account. I haven’t hand-verified every claim — treat it as machine-generated input, not a maintainer sign-off. Findings already raised elsewhere (@tvpeter’s `drop(store)` suggestion, the `OP_CHECKLOCKTIMEVERIFY` typo, and “Won’t this always return true?” on `utils.rs:119`) are omitted.
>
> Tested against `cb1cec09` in three configs (`-p bdk_file_store`, `-p bdk_chain --features rusqlite`, `--workspace --no-default-features --features bdk_chain/hashbrown`). All 6 new tests pass.

Opened #2284 for the test-vector gaps so they don’t block this PR.

---

### `file_store` dev-dep now downloads bitcoind + esplora binaries

`bdk_testenv = { path = "../testenv", features = ["serde"] }` keeps defaults, which pull in `download` → bitcoind/esplora binaries. `cargo test -p bdk_file_store` now compiles and downloads those; previously it only needed `tempfile`. Offline/sandboxed runs break.

```toml
bdk_testenv = { path = "../testenv", default-features = false, features = ["std", "serde", "miniscript"] }
```

(Separate from @noahjoeris’ crate-choice suggestion.)

### Wrong crate’s feature in `#[cfg]`

`#[cfg(feature = "miniscript")]` checks **bdk_chain**’s feature, but guards a symbol gated on **bdk_testenv**’s. They only align today by accident. A consumer enabling `bdk_chain/miniscript` with `bdk_testenv` defaults off gets an unresolved import instead of a skipped test.

### `zip` silently truncates in `create_test_tx`

Four independent iterators are zipped. Mismatched lengths quietly produce fewer inputs/outputs (e.g. 2 txids + 1 vout → 1-input tx) with no warning. Failures then appear in unrelated tests. A `debug_assert_eq!` on lengths would catch this at the source.

### Reference vectors only exercise inserts (tracked in #2284)

Not blocking, but third-party backends should know the gaps:

- **`local_chain_changesets`** — disjoint heights only; no reorg (same height, different hash). Plain `INSERT` still passes.
- **`tx_graph_changesets`** — changeset 2 adds a new tx, so existing txids never get updated `last_seen`/`first_seen`. Upsert path (`INSERT OR IGNORE`) is untested. (`keychain_txout` already covers this.)
- **Block removal `(height, None)`** can’t be added: `from_sqlite` deletes the row and never returns `None`, so strict `load() == merge(...)` fails for correct backends. Document as out of scope, or compare applied state.
- **`persist_test_utils.rs:144`** — `last_seen` < `first_seen` for the same txid (looks like a swap). Harmless for round-trips, but backends that validate the ordering will reject it.
- **Idempotency untested.** Same changeset is never persisted twice. Split into `assert_persists` / `assert_persists_idempotent` for the stronger claim.

PR #2017, opened 2025-08-26
feat!(chain): implement `CheckPointEntry` WIP

LagginTimes:
### Description

This PR builds on #1582 and introduces a new `CheckPointEntry` type to represent checkpoints in a more structured way. It captures both the block data and optional backreferences, enabling chains to be merged even without an explicit point of agreement.

WIP

#### All Submissions:

#### New Features:

#### Bugfixes:


evanlinjin:
I was talking to @LLFourn about this and he suggested an alternative approach.

Instead of introducing a new concept (`CheckPointEntry`), we can make the checkpoint's `data` field optional (since not introducing a new concept is simpler).

This means we can push empty `CheckPoint`s on a chain, and pushing a non-empty `CheckPoint` may push an empty `CheckPoint` before it (to take the `prev_blockhash` into account).

The downside of this is we are introducing an invariant to `CheckPoint` (it may or may not have the data). However, with the `CheckPointEntry` approach, we are already doing the same when iterating through `CheckPoint`s. Would this invariant directly on `CheckPoint` be troublesome?

LagginTimes:
Replaced with #2024.

PR #2024, opened 2025-09-10
feat(chain)!: make `CheckPoint` `data` field optional

### LINKED ISSUE #2021: Ensure `CheckPoint` chain methods validate and link via previous blockhash
**Describe the enhancement**  

**Proposed Changes:**

  ```rust
  pub trait ToBlockHash {
      /* Other method */
    
  }
  ```

**Use case**  

**Impact**

- [x] Yes
- [ ] No

**Additional context**

LagginTimes:
Resolves #2021.

### Description

This PR makes the `CheckPoint::data` field optional. Gaps inferred from `prev_blockhash` are represented as placeholder checkpoints where  `data` is `None`. Subsequent `insert()`s will provide `data` for the placeholders when a matching block arrives.

See: https://github.com/bitcoindevkit/bdk/pull/2017#issuecomment-3231844180

### Notes to the reviewers

WIP

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -648,6 +648,9 @@ where
         match (curr_orig.as_ref(), curr_update.as_ref()) {
             // Update block that doesn't exist in the original chain
             (o, Some(u)) if Some(u.height()) > o.map(|o| o.height()) => {
+                // Only append to `ChangeSet` when the update has complete data. Entries where
+                // `data` does not exist that are created via `prev_blockhash` should not alter the
+                // `ChangeSet`.
                 if let Some(data) = u.data() {
                     changeset.blocks.insert(u.height(), Some(data));
                 }
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -26,13 +26,14 @@ where
         let mut base: Option<CheckPoint<D>> = None;
 
         for cp in init_cp.iter() {
-            if cp.height() >= start_height {
-                if let Some(data) = cp.data() {
+            // Base tip should always have data.
+            if let Some(data) = cp.data() {
+                if cp.height() >= start_height {
                     extension.insert(cp.height(), data);
+                } else {
+                    base = Some(cp);
+                    break;
                 }
-            } else {
-                base = Some(cp);
-                break;
             }
         }
 
@@ -47,18 +48,12 @@ where
             };
         }
 
-        let mut new_tip = match base {
+        let new_tip = match base {
             Some(base) => base
                 .extend(extension)
                 .expect("extension is strictly greater than base"),
             None => LocalChain::from_blocks(extension)?.tip(),
         };
-
-        if new_tip.data_ref().is_none() {
-            new_tip = new_tip
-                .find_data(new_tip.height())
-                .expect("genesis checkpoint should have data");
-        }
 
         init_cp = new_tip;
     }
@@ -702,6 +697,11 @@ where
                             })?;
                             return Ok((new_tip, changeset));
                         }
+                    }
+                    // Even if the hashes are the same, the update may contain data which the
+                    // original does not have.
+                    if let (None, Some(u_data)) = (o.data_ref(), u.data()) {
+                        changeset.blocks.insert(u.height(), Some(u_data));
                     }
                 } else {
                     // We have an invalidation height so we set the height to the updated hash and
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -27,7 +27,9 @@ where
 
         for cp in init_cp.iter() {
             if cp.height() >= start_height {
-                extension.insert(cp.height(), cp.data());
+                if let Some(data) = cp.data() {
+                    extension.insert(cp.height(), data);
+                }
             } else {
                 base = Some(cp);
                 break;
@@ -45,12 +47,19 @@ where
             };
         }
 
-        let new_tip = match base {
+        let mut new_tip = match base {
             Some(base) => base
                 .extend(extension)
                 .expect("extension is strictly greater than base"),
             None => LocalChain::from_blocks(extension)?.tip(),
         };
+
+        if new_tip.data_ref().is_none() {
+            new_tip = new_tip
+                .find_data(new_tip.height())
+                .expect("genesis checkpoint should have data");
+        }
+
         init_cp = new_tip;
     }
 
@@ -322,11 +331,7 @@ where
     /// recover the current chain.
     pub fn initial_changeset(&self) -> ChangeSet<D> {
         ChangeSet {
-            blocks: self
-                .tip
-                .iter()
-                .map(|cp| (cp.height(), Some(cp.data())))
-                .collect(),
+            blocks: self.tip.iter().map(|cp| (cp.height(), cp.data())).collect(),
         }
     }
 
@@ -349,6 +354,20 @@ where
                     update_hash: Some(data.to_blockhash()),
                 });
             }
+
+            // If this `CheckPoint` is an empty placeholder, append the `data` to it.
+            if original_cp.data_ref().is_none() {
+                let mut changeset = ChangeSet::<D>::default();
+                changeset.blocks.insert(height, Some(data));
+                self.apply_changeset(&changeset)
+                    .map_err(|_| AlterCheckPointError {
+                        height: 0,
+                        original_hash: self.genesis_hash(),
+                        update_hash: None,
+                    })?;
+                return Ok(changeset);
+            }
+
             return Ok(ChangeSet::default());
         }
 
@@ -634,7 +653,9 @@ where
         match (curr_orig.as_ref(), curr_update.as_ref()) {
             // Update block that doesn't exist in the original chain
             (o, Some(u)) if Some(u.height()) > o.map(|o| o.height()) => {
-                changeset.blocks.insert(u.height(), Some(u.data()));
+                if let Some(data) = u.data() {
+                    changeset.blocks.insert(u.height(), Some(data));
+                }
                 prev_update = curr_update.take();
             }
             // Original block that isn't in the update
@@ -685,7 +706,7 @@ where
                 } else {
                     // We have an invalidation height so we set the height to the updated hash and
                     // also purge all the original chain block hashes above this block.
-                    changeset.blocks.insert(u.height(), Some(u.data()));
+                    changeset.blocks.insert(u.height(), u.data());
                     for invalidated_height in potentially_invalidated_heights.drain(..) {
                         changeset.blocks.insert(invalidated_height, None);
                     }
--- a/crates/core/src/checkpoint.rs
+++ b/crates/core/src/checkpoint.rs
@@ -24,8 +24,8 @@ impl<D> Clone for CheckPoint<D> {
 struct CPInner<D> {
     /// Block id
     block_id: BlockId,
-    /// Data.
-    data: D,
+    /// Data (if any).
+    data: Option<D>,
     /// Previous checkpoint (if any).
     prev: Option<Arc<CPInner<D>>>,
 }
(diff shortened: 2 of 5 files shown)


LagginTimes:
Solved with #2007.

crates/core/src/checkpoint.rs
```diff
@@ -134,7 +149,8 @@ impl<D> CheckPoint<D> {
     ///
     /// Returns `None` if checkpoint at `height` does not exist`.
     pub fn get(&self, height: u32) -> Option<Self> {
-        self.range(height..=height).next()
```
evanlinjin, line 133: We want this to return all checkpoints (placeholder or not).

crates/core/src/checkpoint.rs
```diff
@@ -167,7 +183,7 @@ impl<D> CheckPoint<D> {
     ///
     /// Returns `None` if no checkpoint exists at or below the given height.
     pub fn floor_at(&self, height: u32) -> Option<Self> {
-        self.range(..=height).next()
+        self.range(..=height).find(|cp| cp.data_ref().is_some())
```
evanlinjin, line 187: Instead of modifying this, how about we create a separate method `.find_next_non_empty_checkpoint`? (please come up with a better name)

crates/core/src/checkpoint.rs
```diff
@@ -203,7 +219,7 @@ where
                 height,
                 hash: data.to_blockhash(),
             },
-            data,
+            data: Some(data),
```
evanlinjin, line 245: We should put a placeholder checkpoint before this (if `data` has `prev_blockhash`). Unless if height is 0, in which case, there is nothing before genesis.

evanlinjin (changes requested):

crates/chain/src/local_chain.rs
```diff
@@ -634,7 +646,7 @@ where
         match (curr_orig.as_ref(), curr_update.as_ref()) {
             // Update block that doesn't exist in the original chain
             (o, Some(u)) if Some(u.height()) > o.map(|o| o.height()) => {
-                changeset.blocks.insert(u.height(), Some(u.data()));
+                changeset.blocks.insert(u.height(), u.data());
```
evanlinjin: This is incorrect. If we reach a placeholder checkpoint, we don't want to insert anything into the changeset. Since a value of `None` in the changeset means deletion (which we are not doing).

crates/chain/src/local_chain.rs
```diff
@@ -27,7 +27,9 @@ where
 
         for cp in init_cp.iter() {
             if cp.height() >= start_height {
-                extension.insert(cp.height(), cp.data());
+                if let Some(data) = cp.data() {
+                    extension.insert(cp.height(), data);
+                }
             } else {
                 base = Some(cp);
```
evanlinjin: Base should never be a placeholder checkpoint - remember to include a comment.

```rust
            if let Some(data) = cp.data() {
                if cp.height() >= start_height {
                    extension.insert(cp.height(), data);
                } else {
                    base = Some(cp);
                    break;
                }
            }
```

crates/chain/src/local_chain.rs
```diff
@@ -45,12 +47,19 @@ where
...
                 .extend(extension)
                 .expect("extension is strictly greater than base"),
             None => LocalChain::from_blocks(extension)?.tip(),
         };
+
+        if new_tip.data_ref().is_none() {
+            new_tip = new_tip
+                .find_data(new_tip.height())
+                .expect("genesis checkpoint should have data");
+        }
```
evanlinjin: The new tip is always coming from the update chain. The update chain should always end with a non-empty tip.

evanlinjin (commented):
.

evanlinjin:
Some recommended test scenarios:

* Update chain that adds data to a placeholder checkpoint in the original chain.
* Update chain which has a PoA with the original chain at a placeholder checkpoint (and vise-versa).

With the above tests:

* Ensure the resultant chain has no "floating" checkpoints.

oleonardolima (commented):
cACK c24dd6a4c02597fea0977927fe5f9c74578ca1f5

It looking good code-wise, though we probably need further discussion on what's the best approach, as discussed by Evan and VM in discord.

crates/core/src/checkpoint.rs
```diff
@@ -277,15 +330,35 @@ where
     ///
     /// Returns an `Err(self)` if the block you are pushing on is not at a greater height that the
     /// one you are pushing on to.
+    ///
+    /// If `height` is non-contiguous and `data.prev_blockhash()` is available, a placeholder is
+    /// created at height - 1.
     pub fn push(self, height: u32, data: D) -> Result<Self, Self> {
```
evanlinjin, line 422: We should return `Err(self)` if `data.prev_blockhash` conflicts with `self`.

```rust
    /// Extends the chain by pushing a new checkpoint.
    ///
    /// Returns `Err(self)` if the height is not greater than the current height, or if the data's
    /// `prev_blockhash` conflicts with `self`.
    ///
    /// Creates a placeholder at height - 1 if the height is non-contiguous and
    /// `data.prev_blockhash()` is available.
    pub fn push(mut self, height: u32, data: D) -> Result<Self, Self> {
        // Reject if trying to push at or below current height - chain must grow forward
        if height <= self.height() {
            return Err(self);
        }

        if let Some(prev_hash) = data.prev_blockhash() {
            if height == self.height() + 1 {
                // For contiguous height, validate that prev_blockhash matches our hash
                // to ensure chain continuity
                if self.hash() != prev_hash {
                    return Err(self);
                }
            } else {
                // For non-contiguous heights, create placeholder to maintain chain linkage
                // This allows sparse chains while preserving block relationships
                self = CheckPoint(Arc::new(CPInner {
                    block_id: BlockId {
                        height: height
                            .checked_sub(1)
                            .expect("height has previous blocks so must be greater than 0"),
                        hash: prev_hash,
                    },
                    data: None,
                    prev: Some(self.0),
                }));
            }
        }

        Ok(Self(Arc::new(CPInner {
            block_id: BlockId {
                height,
                hash: data.to_blockhash(),
            },
            data: Some(data),
            prev: Some(self.0),
        })))
    }
```

crates/core/src/checkpoint.rs
evanlinjin, line 1: `CheckPoint::insert` implementation looks incomplete.

We should be displacing original checkpoints when they conflict with what's being inserted.

Please review the tests I got Claude to write for me: https://github.com/evanlinjin/bdk/commit/0aba292fa1b8c3fbfa0ed3a34c8b2ef15b34f8cc

evanlinjin (changes requested):

crates/core/src/checkpoint.rs
```diff
@@ -284,16 +285,21 @@ where
...
+    /// When inserting data with a `prev_blockhash` that conflicts with existing checkpoints,
+    /// those checkpoints will be displaced and replaced with placeholders. When inserting data
+    /// whose block hash conflicts with the `prev_blockhash` of higher checkpoints, those higher
+    /// checkpoints will be purged.
     ///
     /// # Panics
     ///
     /// This panics if called with a genesis block that differs from that of `self`.
     #[must_use]
     pub fn insert(self, height: u32, data: D) -> Self {
```
evanlinjin, line 314: Insert now looks really complicated. Do you have any ideas to simplify it?

crates/core/tests/test_checkpoint.rs
```diff
@@ -55,3 +55,391 @@ fn checkpoint_destruction_is_sound() {
...
+    );
+}
+
+/// Test inserting between checkpoints with conflicts on both sides.
+///
+/// When inserting at height between two checkpoints where the inserted data's `prev_blockhash`
+/// conflicts with the lower checkpoint and its `blockhash` conflicts with the upper checkpoint's
+/// `prev_blockhash`, both checkpoints should be handled: lower displaced, upper purged.
+///
+/// Expected: Checkpoint at 4 displaced with placeholder, checkpoint at 6 purged.
```
oleonardolima, line 249: I'm not sure if I'm missing something, but shouldn't we also handle the block_3 (placeholder) ? We can't really make sure it connects with the new block_4 placeholder.

oleonardolima: I mean, when pushing a new block height, in case it's contiguous we check that the prev_blockhash matches, I think we'd be breaking this invariant/assumption keeping the placeholder 3 in this scenario.

evanlinjin: That's a good point. Every time we remove a checkpoint, we need to make sure the placeholder checkpoint also gets removed.

crates/core/src/checkpoint.rs
```diff
@@ -239,57 +282,153 @@ where
...
+    /// `prev_blockhash` conflicts with `self`.
+    ///
+    /// Creates a placeholder at height - 1 if the height is non-contiguous and
+    /// `data.prev_blockhash()` is available.
+    pub fn push(mut self, height: u32, data: D) -> Result<Self, Self> {
+        // Reject if trying to push at or below current height - chain must grow forward
+        if height <= self.height() {
+            return Err(self);
         }
+
```
evanlinjin, line 427: Let's prune any floating placeholder checkpoints from `self` here.

**Rationale:** It is public API to find a placeholder checkpoint and call `push` on it. If we push without pruning, we may result in a chain with 2 consecutive placeholder checkpoints.

---

**Additional commentary:** It is possible for a chain source to submit an update with a placeholder tip - and therefore not be persisted. This is the consequence of this optional-data checkpoint API. The benefits of this optional-data API is "less concepts". The alternative is the `CheckPointEntry` API (which introduces an extra concept), but avoids this oddity.

evanlinjin:
@LagginTimes please review the modifications I made to `push` and `insert` here: https://github.com/evanlinjin/bdk/commit/1939dd2e89958c4384a6360535bbe82a8a90c5fd

  This implementation builds upon and improves the original PR's approach to optional checkpoint data in several key ways:

1. Cleaner insert logic
  The original PR had complex nested conditions for handling insertions. This version:
    - Splits the chain into base and tail components upfront
    - Uses a systematic rebuild approach that leverages push for validation
    - Handles all edge cases (displacement, purging, placeholder filling) in a unified flow

2. Proper displacement handling
    - Clear distinction: When a prev_blockhash conflicts, the checkpoint is displaced (converted to placeholder) rather than purged, preserving the chain structure
    - Orphan management: Checkpoints that become orphaned after displacement are correctly purged
    - The original PR struggled with maintaining consistency when conflicts occurred at different points in the chain
  
3. Placeholder cleanup in push
    - Added logic to remove trailing placeholder checkpoints before pushing new data
    - Prevents accumulation of unnecessary placeholders at the chain tip
    - Maintains cleaner chain structure over time

evanlinjin:
@LagginTimes I've rebased on master. Would you be okay if I took over this PR?

crates/chain/src/local_chain.rs
```diff
@@ -51,6 +54,7 @@ where
                 .expect("extension is strictly greater than base"),
```
evanlinjin, line 54: This `expect` is now wrong. A block in the `extension` may have a `prev_blockhash` that doesn't line up.

crates/chain/src/local_chain.rs
evanlinjin, line 1: There seems to be a lot of additional logic in `LocalChain` to manage `ChangeSet`s. This is non-ideal as the mutating logic of `CheckPoint` is now more complex - so we really want to keep the complexity in `CheckPoint` only.

Instead of having `apply_changeset_to_checkpoint`, I'm thinking of using methods of `CheckPoint` directly and comparing the original and updated checkpoint chains to derive the checkpoint.

evanlinjin, line 611: I wonder if the `+ Copy` here will be problematic.

evanlinjin:
Replaced by #2115

PR #2029, opened 2025-09-11
`CanonicalView`

evanlinjin:
### Description

`CanonicalView` allows us to canonicalize upfront, reducing our API surface and improving performance by reducing canonicalizations we need to do.

This is also the first step to achieving many of our goals.

* Event notifications.
* Intent tracker.
* Getting rid of `CanonicalUnspents` structure in `bdk_tx`.

### Changelog notice

```md
Added
- Introduce `CanonicalView` which allows us to canonicalize once upfront.
- Added `TxGraph::canonical_view` which constructs a `CanonicalView`.

Changed
- `TxGraph` methods which require canonicalization now have `CanonicalView` equivalents.

Removed
- `TxGraph` methods which take in a fallible `ChainOracle` implementations are now removed. 

```

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/canonical_view.rs
+++ b/crates/chain/src/canonical_view.rs
@@ -41,7 +41,7 @@ use crate::{
 /// conflicted). It includes the transaction itself along with its position in the chain (confirmed
 /// or unconfirmed).
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub struct CanonicalViewTx<A> {
+pub struct CanonicalTx<A> {
     /// The position of this transaction in the chain.
     ///
     /// This indicates whether the transaction is confirmed (and at what height) or
@@ -228,11 +228,11 @@ impl<A: Anchor> CanonicalView<A> {
     ///     println!("Found tx {} at position {:?}", canonical_tx.txid, canonical_tx.pos);
     /// }
     /// ```
-    pub fn tx(&self, txid: Txid) -> Option<CanonicalViewTx<A>> {
+    pub fn tx(&self, txid: Txid) -> Option<CanonicalTx<A>> {
         self.txs
             .get(&txid)
             .cloned()
-            .map(|(tx, pos)| CanonicalViewTx { pos, txid, tx })
+            .map(|(tx, pos)| CanonicalTx { pos, txid, tx })
     }
 
     /// Get a single canonical transaction output.
@@ -302,12 +302,10 @@ impl<A: Anchor> CanonicalView<A> {
     /// // Get the total number of canonical transactions
     /// println!("Total canonical transactions: {}", view.txs().len());
     /// ```
-    pub fn txs(
-        &self,
-    ) -> impl ExactSizeIterator<Item = CanonicalViewTx<A>> + DoubleEndedIterator + '_ {
+    pub fn txs(&self) -> impl ExactSizeIterator<Item = CanonicalTx<A>> + DoubleEndedIterator + '_ {
         self.order.iter().map(|&txid| {
             let (tx, pos) = self.txs[&txid].clone();
-            CanonicalViewTx { pos, txid, tx }
+            CanonicalTx { pos, txid, tx }
         })
     }
 
--- a/crates/chain/src/canonical_view.rs
+++ b/crates/chain/src/canonical_view.rs
@@ -1,4 +1,25 @@
-//! Canonical view.
+//! Canonical view of transactions and unspent outputs.
+//!
+//! This module provides [`CanonicalView`], a utility for obtaining a canonical (ordered and
+//! conflict-resolved) view of transactions from a [`TxGraph`].
+//!
+//! ## Example
+//!
+//! ```
+//! # use bdk_chain::{CanonicalView, TxGraph, CanonicalizationParams, local_chain::LocalChain};
+//! # use bdk_core::BlockId;
+//! # use bitcoin::hashes::Hash;
+//! # let tx_graph = TxGraph::<BlockId>::default();
+//! # let chain = LocalChain::from_blocks([(0, bitcoin::BlockHash::all_zeros())].into_iter().collect()).unwrap();
+//! # let chain_tip = chain.tip().block_id();
+//! let params = CanonicalizationParams::default();
+//! let view = CanonicalView::new(&tx_graph, &chain, chain_tip, params).unwrap();
+//!
+//! // Iterate over canonical transactions
+//! for tx in view.txs() {
+//!     println!("Transaction {}: {:?}", tx.txid, tx.pos);
+//! }
+//! ```
 
 use crate::collections::HashMap;
 use alloc::sync::Arc;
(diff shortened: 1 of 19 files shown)


crates/chain/src/canonical_view.rs
```diff
@@ -0,0 +1,275 @@
...
+            spends: HashMap::new(),
+        };
+
+        for r in CanonicalIter::new(tx_graph, chain, chain_tip, params) {
+            let (txid, tx, why) = r?;
+
+            let tx_node = match tx_graph.get_tx_node(txid) {
+                Some(tx_node) => tx_node,
+                None => {
+                    // TODO: Have the `CanonicalIter` return `TxNode`s.
```
oleonardolima, line 127: @evanlinjin in the case CanonicalIter returns a `TxNode`, how should we proceed when for some reason it does not exists ? Should we just keep the same `debug_assert` and "ignore it" or throw an  error instead ?

crates/chain/src/canonical_view.rs
```diff
@@ -0,0 +1,275 @@
...
+
+        let mut view = Self {
+            tip: chain_tip,
+            order: vec![],
+            txs: HashMap::new(),
+            spends: HashMap::new(),
+        };
+
+        for r in CanonicalIter::new(tx_graph, chain, chain_tip, params) {
+            let (txid, tx, why) = r?;
```
oleonardolima, line 122: nit: I think we should stick with `reason` instead of `why`.

crates/chain/src/canonical_view.rs
```diff
@@ -209,10 +209,25 @@ impl<A: Anchor> CanonicalView<A> {
...
+    /// ### Minimum confirmations
+    ///
+    /// `min_confirmations` specifies the minimum number of confirmations required for a transaction
+    /// to be counted as confirmed in the returned [`Balance`]. Transactions with fewer than
+    /// `min_confirmations` will be treated as trusted pending (assuming the `trust_predicate`
+    /// returns `true`).
+    ///
+    /// - `min_confirmations = 0`: Include all confirmed transactions (same as `1`)
+    /// - `min_confirmations = 1`: Standard behavior - require at least 1 confirmation
+    /// - `min_confirmations = 6`: High security - require at least 6 confirmations
```
oleonardolima: I'm wondering as we discussed if an `additional_confirmations` wouldn't be a better name, and we always use 1 as the minimum confirmation.

crates/chain/src/canonical_view.rs
```diff
@@ -14,28 +35,77 @@ use crate::{
...
+    /// The full transaction.
     pub tx: Arc<Transaction>,
 }
 
-/// A view of canonical transactions.
+/// A view of canonical transactions from a [`TxGraph`].
+///
+/// `CanonicalView` provides an ordered, conflict-resolved view of transactions. It determines
+/// which transactions are canonical (non-conflicted) based on the current chain state and
+/// provides methods to query transaction data, unspent outputs, and balances.
```
oleonardolima, line 60: I think we can explain what the ordering means here, e.g topologically-spending order.

crates/chain/src/canonical_view.rs
```diff
@@ -139,15 +209,60 @@ impl<A: Anchor> CanonicalView<A> {
...
-    /// Get a single canonical txout.
+    /// Get a single canonical transaction output.
+    ///
+    /// Returns detailed information about a transaction output, including whether it has been
+    /// spent and by which transaction.
+    ///
+    /// Returns `None` if:
+    /// - The transaction doesn't exist in the canonical view
+    /// - The output index is out of bounds
+    /// - The transaction was excluded due to conflicts
```
oleonardolima, line 209: Is the `The transaction was excluded due to conflicts` redundant to it not existing in the canonical view ?

oleonardolima (commented):
cACK / utACK 809c1537cb745740f3ff3e7d6b53d5efd0b3c218

It looks good to me so far, I left some nit and discussion comments above.

LagginTimes:
LGTM so far but needs a refactor now.

crates/chain/src/canonical_view.rs
evanlinjin: If `CanonicalIter` returns `TxNode`s, then that means the `TxNode` must definitely exist.

crates/chain/src/canonical_view.rs
evanlinjin: Yes topological-spending order, with confirmation height/first-seen-in-mempool as the secondary order. However we should also mention that this ordering does not exist yet!

crates/chain/src/canonical_view.rs
```diff
@@ -0,0 +1,436 @@
...
+    /// let balance = view.balance(
+    ///     indexer.outpoints().into_iter().map(|(k, op)| (k.clone(), *op)),
+    ///     |_keychain, _script| true,  // Trust all outputs
+    ///     6,  // Require 6 confirmations
+    /// );
+    /// ```
+    pub fn balance<'v, O: Clone + 'v>(
+        &'v self,
+        outpoints: impl IntoIterator<Item = (O, OutPoint)> + 'v,
+        mut trust_predicate: impl FnMut(&O, ScriptBuf) -> bool,
```
ValuedMammal: I think `trust_predicate` is not expressive enough, for instance we're missing information about the txid. Using `FullTxOut` as the operand, we have access to more things in addition to the script pubkey.

```suggestion
        mut trust_predicate: impl FnMut(&O, &FullTxOut<A>) -> bool,
```

evanlinjin: Good idea. Addressed in cb72f4ae9d807b2c5376cc84297501961d36ae53

crates/chain/src/canonical_view.rs
evanlinjin: Maybe let's update the documentation once the transactions are actually topologically ordered.

evanlinjin: Resolved couple of lines down.

oleonardolima (approved):
utACK cb72f4ae9d807b2c5376cc84297501961d36ae53

crates/chain/src/canonical_view.rs
evanlinjin: To make it even more expressive, we can change the closure to output how many additional confirmations the UTXO needs to be considered "finalized". This will remove the `additional_confirmations` parameter on `balance`.

Wdyt?

crates/chain/src/canonical_view.rs
ValuedMammal: I think minimum confirmations would be a more intuitive API, otherwise it forces the caller to do the math of figuring out how many confirmations is enough to be considered confirmed, which seems awkward.

ValuedMammal: If `0` is a special case, maybe just consider everything to be pending unconfirmed?

crates/chain/src/canonical_view.rs
ValuedMammal: I'm not seeing the connection between trust predicate and additional confirmations, and so I think it'd be best for them to remain separate. FTR I'm sure there's other ways for a user to filter unspents by a target number of confs, so I'm not exactly married to this approach, but it also seems convenient as most devs are already familiar with the `balance` method.

oleonardolima:
I discussed with @evanlinjin and dropped the 7245631bfb0f11679ad953e0233be60f9cb84a44, which changed the `min_confirmations` to `additional_confirmations`.

~~I didn't want to rebase everything, but you probably need a rebase to get the clippy fixes.~~

evanlinjin (commented):
self-ACK 1311a2ed050534558ef738ceb754bd9b3a5552a6

ValuedMammal (approved):
ACK 1311a2ed050534558ef738ceb754bd9b3a5552a6

oleonardolima:
I'm wondering if we should already merge this to master, as we already have breaking changes on master. And with regards to the other standalone PRs that are non-breaking, if we should merge then in a specific new branch e.g. release/0.23.3.

Let me know if you guys have another releasing approach in mind.

oleonardolima:
I ran the benchmark on this PR, and so far there's no big regressions
``` bash
many_conflicting_unconfirmed::list_canonical_txs
                        time:   [746.06 us 747.07 us 748.02 us]
                        change: [-0.6272% -0.4520% -0.2739%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 13 outliers among 100 measurements (13.00%)
  7 (7.00%) low mild
  6 (6.00%) high mild

many_conflicting_unconfirmed::filter_chain_txouts
                        time:   [842.09 us 844.42 us 846.63 us]
                        change: [-0.8799% +0.1490% +0.7945%] (p = 0.80 > 0.05)
                        No change in performance detected.
Found 14 outliers among 100 measurements (14.00%)
  3 (3.00%) low mild
  11 (11.00%) high mild

many_conflicting_unconfirmed::filter_chain_unspents
                        time:   [841.98 us 844.33 us 846.80 us]
                        change: [+1.2639% +1.5485% +1.8660%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

many_chained_unconfirmed::list_canonical_txs
                        time:   [5.7585 ms 5.7663 ms 5.7751 ms]
                        change: [+6.9882% +7.3202% +7.6471%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe

many_chained_unconfirmed::filter_chain_txouts
                        time:   [5.6686 ms 5.6742 ms 5.6803 ms]
                        change: [-4.3940% -3.7067% -3.1341%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  5 (5.00%) high mild
  2 (2.00%) high severe

many_chained_unconfirmed::filter_chain_unspents
                        time:   [5.6651 ms 5.6779 ms 5.6952 ms]
                        change: [-2.2329% -1.9763% -1.7036%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  2 (2.00%) low mild
  3 (3.00%) high mild
  4 (4.00%) high severe

nested_conflicts_unconfirmed::list_canonical_txs
                        time:   [452.26 us 452.78 us 453.35 us]
                        change: [-0.3477% -0.1279% +0.0975%] (p = 0.26 > 0.05)
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe

nested_conflicts_unconfirmed::filter_chain_txouts
                        time:   [498.89 us 499.39 us 499.91 us]
                        change: [-2.0289% -1.5397% -1.1074%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe

nested_conflicts_unconfirmed::filter_chain_unspents
                        time:   [498.07 us 498.58 us 499.10 us]
                        change: [-1.5937% -1.1912% -0.8211%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) low mild
  2 (2.00%) high mild
  4 (4.00%) high severe
```

evanlinjin:
@oleonardolima I don't think we need a point release for `CanonicalView`, it adds more work with little upside.

`CanonicalView` is most useful for:
* Slight performance improvement (not too important).
* `IntentTracker` - this will probably take 1-2 more months.
* Removing `CanonicalUnspents` from `bdk-tx` - not used by `bdk_wallet v2.x` users anyway.
* Events - can be done without it also.

crates/chain/src/canonical_view.rs
```diff
@@ -0,0 +1,436 @@
...
+            spends: HashMap::new(),
+        };
+
+        for r in CanonicalIter::new(tx_graph, chain, chain_tip, params) {
+            let (txid, tx, why) = r?;
+
+            let tx_node = match tx_graph.get_tx_node(txid) {
+                Some(tx_node) => tx_node,
+                None => {
+                    // TODO: Have the `CanonicalIter` return `TxNode`s.
```
oleonardolima, line 127: I'll create an issue to address this one in a follow-up PR.

oleonardolima (approved):
ACK 1311a2ed050534558ef738ceb754bd9b3a5552a6

PR #2031, opened 2025-09-12
Add `verify-standalone` and `verify-standalone-all` recipes

### LINKED ISSUE #2030: Add CI check to verify crates build correctly without workspace `path` deps
## Background  
Currently, when we build a crate in the workspace, Cargo still prefers local `path` dependencies. This means we might accidentally publish a crate that only works inside our workspace but fails for downstream users, because it relies on unpublished changes in sibling crates.  

To catch this before release, we need a CI job (and a corresponding `just` recipe) that:  
1. Packages the crate (`cargo package`)  
2. Unpacks the tarball  
3. Builds it in isolation with `--locked`, ensuring it resolves dependencies strictly from crates.io  

This simulates exactly how downstream users will consume our crates.  

## Proposed changes  
- Add a command to our `justfile` recipe (`just check-publish <crate>`) to package, unpack, and rebuild the crate tarball.

~- Add a CI job that runs for each crate targeted for release.~  
~- Integrate this check into the release workflow so that every crate we publish has been validated against the registry, not just the workspace. We need to label-gate this check to only run for PRs with the `release` label.~

**Edit:** We don't want to run this for every release, just as a convenient manual check.

## Example `justfile` snippet
```make
# Usage: just check-publish crate-name
check-publish crate:
    cargo package -p {{crate}}
    pkg=$(ls target/package/{{crate}}-*.crate | tail -n1); \
    tmp=$(mktemp -d); \
    tar -xzf "$$pkg" -C "$$tmp"; \
    export CARGO_HOME="$$tmp/cargo-home"; \
    export RUSTUP_HOME="$$tmp/rustup-home"; \
    cargo build --locked --manifest-path "$$tmp"/{{crate}}-*/Cargo.toml
```

evanlinjin:
Fixes #2030

### Description

Adds two `just` recipes that check whether workspace crates build as they would once published:

* `verify-standalone <crate> [args]` (alias `vs`): packages the crate and builds the packaged copy outside the workspace, so sibling workspace crates resolve from crates.io instead of local paths. A failure usually means the crate uses something from a sibling crate that isn't released yet. Extra args are passed to `cargo build`.
* `verify-standalone-all [args]` (alias `vsa`): runs `verify-standalone --all-features` on every publishable workspace crate, stopping at the first failure. `--all-features` catches uses of unreleased APIs behind non-default features.

The registry cache and build artifacts are kept under the target directory, so repeat runs are fast.

### Notes to the reviewers

Neither recipe is wired into CI. On current `master`, `bdk_chain` and `bdk_bitcoind_rpc` fail because they use `bdk_core` APIs that aren't released yet, which is what the check is for.

#### All Submissions:

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/justfile
+++ b/justfile
@@ -4,6 +4,7 @@ alias f := fmt
 alias t := test
 alias p := pre-push
 alias vs := verify-standalone
+alias vsa := verify-standalone-all
 
 _default:
   @just --list
@@ -98,3 +99,36 @@ verify-standalone crate *args="":
     cargo build --locked
     
     echo "✅ {{crate}} builds successfully in isolation!"
+
+# Verify all publishable crates can be built standalone
+verify-standalone-all *args="":
+    #!/usr/bin/env bash
+    set -euo pipefail
+    
+    # Get all workspace crates (excluding examples)
+    CRATES=$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.source == null and (.name | startswith("example_") | not)) | .name' | sort)
+    
+    echo "Verifying all publishable crates can build standalone..."
+    echo "Crates to verify:"
+    echo "$CRATES" | sed 's/^/  - /'
+    echo
+    
+    for crate in $CRATES; do
+        echo "========================================="
+        echo "Verifying $crate..."
+        echo "========================================="
+        
+        if ! just verify-standalone "$crate" {{args}}; then
+            echo "❌ $crate: FAILED"
+            echo "========================================="
+            echo "❌ VERIFICATION FAILED for: $crate"
+            echo "========================================="
+            exit 1
+        fi
+        echo "✅ $crate: SUCCESS"
+        echo
+    done
+    
+    echo "========================================="
+    echo "✅ All crates build successfully in isolation!"
+    echo "========================================="
--- a/justfile
+++ b/justfile
@@ -3,6 +3,7 @@ alias c := check
 alias f := fmt
 alias t := test
 alias p := pre-push
+alias vs := verify-standalone
 
 _default:
   @just --list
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -57,3 +58,43 @@ _test-testenv:
 
 # Run pre-push suite: format, check, and test
 pre-push: fmt check test
+
+# Verify that a crate can be built standalone without workspace dependencies
+# This ensures the crate is publishable and doesn't accidentally depend on
+# breaking changes from other workspace crates
+verify-standalone crate *args="":
```
oleonardolima, line 66: I think having also another `all` option would be useful, so you can call it once and test it for all crates in workspace.

oleonardolima (commented):
tACK 63595eedc7e7b5726ee09bb319e0101c30a08999

justfile
evanlinjin: Sure! I'll get claude to do that for us.

evanlinjin: Addressed in 3fbf406c31378e747bf3d1d5ca90bc60fc6db297

oleonardolima (commented):
cACK 3fbf406c31378e747bf3d1d5ca90bc60fc6db297

ValuedMammal:
Long term I think it would be great to have a `contrib/` folder for bash scripts that can be invoked from the `justfile`.

oleonardolima:
I'm not sure why, but now bdk_testenv is failing on my machine:
```
error: failed to parse manifest at `/var/folders/1j/b7q_zgx138q99356q8t83pjr0000gn/T/tmp.kb5jdzuAMK/.cargo/registry/src/index.crates.io-6f17d22bba15001f/base64ct-1.8.0/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.83.0 (5ffbef321 2024-10-29)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
error: Recipe `verify-standalone` failed with exit code 101
```

oleonardolima (approved):
tACK 3fbf406c31378e747bf3d1d5ca90bc60fc6db297

It looks fine, the issue I was facing was exclusively on my machine.

notmandatory (approved):
tACK 3fbf406c31378e747bf3d1d5ca90bc60fc6db297


Works for rust stable, doesn't work for MSRV 1.63 or 1.75 but this is expected right?

justfile
```diff
@@ -57,3 +59,76 @@ _test-testenv:
 
 # Run pre-push suite: format, check, and test
 pre-push: fmt check test
```
oleonardolima, line 61: nit: the new `vsa` could be added to the `pre-push` recipe.

notmandatory:
Looks like this needs a rebase to pickup and rerun actions with latest `stable` and MSRV versions for required checks.

justfile
evanlinjin: I don't think so. It's only needed for publishing.

PR #2037, opened 2025-09-18
Add median-time-past (MTP) calculation to CheckPoint

### LINKED ISSUE #2036: Add median-time-past (MTP) calculation to CheckPoint
## Summary
Add median-time-past (MTP) calculation capability to `CheckPoint` for time-based validation in Bitcoin consensus rules.

## Proposed Changes

1. **Extend `ToBlockHash` trait** with optional block time:
```rust
fn block_time(&self) -> Option<u32> { None }
```

2. **Add `median_time_past()` method to `CheckPoint`**:
   - Calculate median of up to 11 previous block timestamps
   - Return `None` if:
     - Data type doesn't support block times
     - Required blocks are missing (sparse chain)
     - Insufficient block history available

## Use Cases
- Transaction time-lock verification (nLockTime, OP_CHECKLOCKTIMEVERIFY)
- BIP113 compliance

## Notes
- `CheckPoint` represents a sparse chain, so required blocks may not be available
- MTP = median of current block + up to 10 predecessors (sorted timestamps)

evanlinjin:
## Description

This PR adds the ability to calculate median-time-past (MTP) for `CheckPoint` structures, implementing the functionality described in #2036.

## Notes to the reviewers

`CheckPoint::median_time_past` calculates the MTP value by looking at the previous 11 blocks (including the current block).

### Why `ToBlockTime` is a separate trait from `ToBlockHash`

`ToBlockTime` is intentionally kept as a separate trait with a non-optional return type (`fn to_blocktime(&self) -> u32`).

For operations like MTP calculation, block times are either fully available or the calculation is meaningless — there's no useful "partial" or "best-effort" result. By using a separate trait, we get compile-time guarantees: methods like `median_time_past()` use `where D: ToBlockTime` bounds, making it explicit which checkpoint data types support time-based operations. Types without block time data (e.g., `BlockHash`, `BlockId`) simply don't implement the trait.

#### How this differs from `prev_blockhash`

In contrast, `prev_blockhash() -> Option<BlockHash>` is a method on `ToBlockHash` rather than a separate trait. The `Option` return serves a different purpose — it allows graceful degradation. A `CheckPoint<BlockId>` chain is still useful even without hash linkage validation; callers can simply skip verification when `None` is returned. Additionally, `None` at genesis is semantically meaningful (it marks the chain root), not an error condition.

For MTP, there's no equivalent "skip if unavailable" — you either have all 11 timestamps or the calculation fails. This binary requirement is better expressed as a trait bound than runtime `Option` handling.

### Added
- Introduced `ToBlockTime` trait for types that can return a block time.
- Added `median_time_past()` method to `CheckPoint` for calculating MTP according to BIP113

### All Submissions:

### New Features:

Fixes #2036

### DIFF
--- a/crates/core/src/checkpoint.rs
+++ b/crates/core/src/checkpoint.rs
@@ -2,6 +2,7 @@ use core::fmt;
 use core::ops::RangeBounds;
 
 use alloc::sync::Arc;
+use alloc::vec::Vec;
 use bitcoin::{block::Header, BlockHash};
 
 use crate::BlockId;
@@ -56,10 +57,10 @@ impl<D> Drop for CPInner<D> {
 
 /// Trait that converts [`CheckPoint`] `data` to [`BlockHash`].
 ///
-/// Implementations of [`ToBlockHash`] must always return the block’s consensus-defined hash. If
+/// Implementations of [`ToBlockHash`] must always return the block's consensus-defined hash. If
 /// your type contains extra fields (timestamps, metadata, etc.), these must be ignored. For
 /// example, [`BlockHash`] trivially returns itself, [`Header`] calls its `block_hash()`, and a
-/// wrapper type around a [`Header`] should delegate to the header’s hash rather than derive one
+/// wrapper type around a [`Header`] should delegate to the header's hash rather than derive one
 /// from other fields.
 pub trait ToBlockHash {
     /// Returns the [`BlockHash`] for the associated [`CheckPoint`] `data` type.
@@ -75,6 +76,20 @@ impl ToBlockHash for BlockHash {
 impl ToBlockHash for Header {
     fn to_blockhash(&self) -> BlockHash {
         self.block_hash()
+    }
+}
+
+/// Trait that extracts a block time from [`CheckPoint`] `data`.
+///
+/// `data` types that contain a block time should implement this.
+pub trait ToBlockTime {
+    /// Returns the block time from the [`CheckPoint`] `data`.
+    fn to_blocktime(&self) -> u32;
+}
+
+impl ToBlockTime for Header {
+    fn to_blocktime(&self) -> u32 {
+        self.time
     }
 }
 
@@ -191,6 +206,8 @@ impl<D> CheckPoint<D>
 where
     D: ToBlockHash + fmt::Debug + Copy,
 {
+    const MTP_BLOCK_COUNT: u32 = 11;
+
     /// Construct a new base [`CheckPoint`] from given `height` and `data` at the front of a linked
     /// list.
     pub fn new(height: u32, data: D) -> Self {
@@ -202,6 +219,38 @@ where
             data,
             prev: None,
         }))
+    }
+
+    /// Calculate the median time past (MTP) for this checkpoint.
+    ///
+    /// Uses 11 blocks (heights h-10 through h, where h is the current height) to compute the MTP
+    /// for the current block. This is used in Bitcoin's consensus rules for time-based validations
+    /// (BIP-0113).
+    ///
+    /// Note: This is a pseudo-median that doesn't average the two middle values.
+    ///
+    /// Returns `None` if the data type doesn't support block times or if any of the required
+    /// 11 sequential blocks are missing.
+    pub fn median_time_past(&self) -> Option<u32>
+    where
+        D: ToBlockTime,
+    {
+        let current_height = self.height();
+        let earliest_height = current_height.saturating_sub(Self::MTP_BLOCK_COUNT - 1);
+
+        let mut timestamps = (earliest_height..=current_height)
+            .map(|height| {
+                // Return `None` for missing blocks or missing block times
+                let cp = self.get(height)?;
+                let block_time = cp.data_ref().to_blocktime();
+                Some(block_time)
+            })
+            .collect::<Option<Vec<u32>>>()?;
+        timestamps.sort_unstable();
+
+        // If there are more than 1 middle values, use the higher middle value.
+        // This is mathematically incorrect, but this is the BIP-0113 specification.
+        Some(timestamps[timestamps.len() / 2])
     }
 
     /// Construct from an iterator of block data.
--- a/crates/core/tests/test_checkpoint.rs
+++ b/crates/core/tests/test_checkpoint.rs
@@ -1,5 +1,6 @@
-use bdk_core::CheckPoint;
+use bdk_core::{CheckPoint, ToBlockHash, ToBlockTime};
 use bdk_testenv::{block_id, hash};
+use bitcoin::hashes::Hash;
 use bitcoin::BlockHash;
 
 /// Inserting a block that already exists in the checkpoint chain must always succeed.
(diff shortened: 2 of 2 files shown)


oleonardolima (commented):
cACK 0f75c221dcf840a1496d82e4ee8bbbaa2e5d82b2

I'll just give it a try with some other test scenarios before final ACK.

crates/core/src/checkpoint.rs
```diff
@@ -208,6 +223,54 @@ where
...
+    ///
+    /// Note: This is a pseudo-median that doesn't average the two middle values.
+    ///
+    /// Returns `None` if the data type doesn't support block times or if any of the required
+    /// 11 sequential blocks are missing.
+    pub fn median_time_past(&self) -> Option<u32> {
+        let current_height = self.height();
+        let earliest_height = current_height.saturating_sub(Self::MTP_BLOCK_COUNT);
+        self._median_time_past(earliest_height..current_height)
+    }
```
oleonardolima: I'm curious if these new APIs could output wrong/invalid values if the blocks inserted have a wrong/invalid time field e.g malicious block source.

evanlinjin: Definitely. We are trusting the chain source here (and everywhere in BDK). Eventually, we should think about how to verify all consensus rules in BDK: difficulty, PoW, MTP, etc.

oleonardolima (approved):
cACK 74d30d80b7784fca5005ae9dcceb3a2d6028cbf1

I didn't really find other test vectors that we could use.

Another thing that should probably be handled in a future PR is updating the calculation of Balance for the immature category, we probably should update it to consider transactions that are not yet mature considering the Locktime and MTP for it.

crates/core/src/checkpoint.rs
ValuedMammal: Why is it not inclusive of the current height?

crates/core/src/checkpoint.rs
```diff
@@ -60,14 +61,22 @@ impl<D> Drop for CPInner<D> {
...
     /// Returns the [`BlockHash`] for the associated [`CheckPoint`] `data` type.
     fn to_blockhash(&self) -> BlockHash;
+
+    /// Returns the block time if available for the data type.
+    ///
+    /// Default implementation returns `None`. Data types that include timestamp information
+    /// (such as [`Header`]) should override this method.
+    fn block_time(&self) -> Option<u32> {
+        None
+    }
```
ValuedMammal: Hm but then the trait would need to be called `ToBlockHashAndHeaderTime`? Maybe just have a new impl where `D` is convertible to a `Header` (granted, the time is only a subset of the information contained in a header).

```rust
impl<D> CheckPoint<D>
where
    D: Into<Header>,
{
    /// Calculates the median time past (MTP) of this `CheckPoint`.
    pub fn median_time_past(&self) -> Option<u32> { ... }
}
```

crates/core/src/checkpoint.rs
ValuedMammal: nit: I want to avoid the temptation to resort to using these underscore methods.

evanlinjin: > Why is it not inclusive of the current height?


This method is asking "what's the MTP that was used to validate block 100".


However, if the caller is asking "what MTP should I use to validate a transaction going into the next block", `next_median_time_past()` can be used.

crates/core/src/checkpoint.rs
evanlinjin: What do you think about a separate `ToBlockTime` trait?

crates/core/src/checkpoint.rs
evanlinjin: > nit: I want to avoid the temptation to resort to using these underscore methods.

I renamed the private method to `middle_timestamp_of_range`.

ValuedMammal: > > Why is it not inclusive of the current height?
> 
> This method is asking "what's the MTP that was used to validate block 100".
> 
> However, if the caller is asking "what MTP should I use to validate a transaction going into the next block", `next_median_time_past()` can be used.

I imagine there would just be one function called `median_time_past` with a range that includes the current tip. If I wanted to know the last MTP, I would just call this function on the previous checkpoint, right?

crates/core/src/checkpoint.rs
ValuedMammal: Yes or maybe stay with 1 trait but redefine `ToBlockHash` to be a more general `Header`-like interface.

crates/core/src/checkpoint.rs
```diff
@@ -204,6 +221,64 @@ where
...
+            .into_iter()
+            .map(|height| {
+                // Return `None` for missing blocks or missing block times
+                let cp = self.get(height)?;
+                let block_time = cp.data_ref().to_blocktime();
+                Some(block_time)
+            })
+            .collect::<Option<Vec<u32>>>()?;
+        timestamps.sort_unstable();
+        Some(timestamps[timestamps.len().checked_sub(1)? / 2])
```
luisschwab: super nit: BIP-0113 specifies `len / 2`. In the case of 11 blocks, both `len / 2` and `(len - 1)/2` yield 5, which is fine, but for even lengths, we get two elements. The median of an even number of elements is the average of the middle elements. Here we must choose one of the values, so perhaps remove the `.checked_sub(1)` to keep on par with the BIP?

luisschwab (approved):
ACK 00b0ca8a721e4d432e57c90346822bbc8b7e9335

crates/core/src/checkpoint.rs
evanlinjin: I was wrong about the definition of MTP. MTP of block `h` includes the time of block `h`. However, when validating whether a tx can be included into a block at height `x`, consensus uses `MTP(x-1)`.

> I imagine there would just be one function called `median_time_past` with a range that includes the current tip.

Yes. This makes sense as `MTP(spending_block - 1)` is essentially `MTP(tip)`.

>  If I wanted to know the last MTP, I would just call this function on the previous checkpoint, right?

~~This is true. However, a prev-MTP helper will be useful for relative-timelocked txs as we may need to get the confirmation block hash as well as `MTP(conf_block - 1)`.~~

Edit: I agree with you. Less is more.

crates/core/src/checkpoint.rs
evanlinjin: Arggggg

crates/core/src/checkpoint.rs
evanlinjin: I think 2 traits is nicer otherwise the `to_blocktime` method needs to return an `Option`.

rustaceanrob:
FYI in `units = 1.0.0` there will be a [median time past](https://github.com/rust-bitcoin/rust-bitcoin/blob/f0df6c58a77a3be53de098545c1b29362b36dc14/units/src/locktime/absolute/mod.rs#L613) struct

evanlinjin:
> FYI in `units = 1.0.0` there will be a [median time past](https://github.com/rust-bitcoin/rust-bitcoin/blob/f0df6c58a77a3be53de098545c1b29362b36dc14/units/src/locktime/absolute/mod.rs#L613) struct

Thanks! Unfortunately, I don't think we can use that here 😢

evanlinjin:
CI is failing `no_std` builds. #2099 will fix this.

crates/core/src/checkpoint.rs
evanlinjin: @ValuedMammal I added a justification of having a separate trait for `to_blocktime` in the PR description.

ValuedMammal (approved):
ACK b6235360e7232ad8bba2b12049eacdd2a6f9b3ca

PR #2038, opened 2025-09-18
refactor(chain,core)!: replace `CanonicalIter` with sans-IO `CanonicalTask` + `ChainQuery` trait

### LINKED ISSUE #1816: Remove ChainOracle trait by inverting dependency
The ChainOracle trait is annoying. Initially designed to allow anyone to provide their own view of the chain. This is a noble goal but doing this via a trait means we force the ChainOracle to do it synchronously and have to handle errors from the ChainOracle.

ChainOracles are really only used for canonicalization. Instead of writing:

```rust
tx_graph.list_canonical_txs(&local_chain, chain_tip);
```

We could instead write:

```rust
let list_txs_task = tx_graph.list_canonical_txs(chain_tip);
let txs_iter = local_chain.canonicalize(list_txs_task);
```

Now if we want to replace the canonicalizing functionality with a bitcoin node or a CBF system. It can be sync or async, infallible or fallible and we don't have to care in BDK.

oleonardolima:
fixes #1816

### Description

Replaces the iterator-based `CanonicalIter` with a two-phase sans-IO canonicalization pipeline, and introduces a generic `ChainQuery` trait in `bdk_core` to decouple canonicalization from chain sources.

**Old API:**
```rust
// Direct coupling between canonicalization logic and ChainOracle
let view = tx_graph.canonical_view(&chain, chain_tip, params)?;
```

**New API:**
```rust
// Option A: Two-phase (full control)
let canonical_txs = chain.canonicalize(tx_graph.canonical_task(tip, params));
let view = chain.canonicalize(canonical_txs.view_task(&tx_graph));

// Option B: Convenience method
let view = chain.canonical_view(&tx_graph, tip, params);
```

#### Phase 1: `CanonicalTask`
Determines which transactions are canonical by processing them in stages:
1. **Assumed txs** — transactions assumed canonical via `CanonicalParams`
2. **Anchored txs** — transactions anchored in the best chain (descending height)
3. **Seen txs** — unconfirmed transactions by descending last-seen time
4. **Remaining txs** — leftover anchored transactions not in the best chain

Produces a `CanonicalTxs<A>` containing each canonical transaction with its `CanonicalReason`.

#### Phase 2: `CanonicalViewTask`
Resolves `CanonicalReason`s into concrete `ChainPosition`s (confirmed height or unconfirmed with last-seen), producing the final `CanonicalView<A>`.

Both phases implement the `ChainQuery` trait, so any chain source can drive them via the same `next_query`/`resolve_query` loop.

#### Key structural changes

- **`ChainQuery` trait** added to `bdk_core` — a generic sans-IO interface (`next_query` → `resolve_query` → `finish`) for any algorithm that needs to verify blocks against a chain source.
- **`ChainOracle` trait removed** — replaced by `ChainQuery`. `LocalChain::canonicalize()` now drives any `ChainQuery` implementor.
- **`Canonical<A, P>` generic container** — `CanonicalTxs<A>` (phase 1 output) and `CanonicalView<A>` (phase 2 output) are type aliases over `Canonical<A, P>`.
- **Module split** — `canonical_view.rs` split into `canonical.rs` (types: `Canonical`, `CanonicalTx`, `CanonicalTxOut`) and `canonical_view_task.rs` (phase 2 task). `canonical_iter.rs` replaced by `canonical_task.rs`.

### Notes to the reviewers

The changes are split into multiple commits for easier review. Also depends on #2029.

### Changelog notice

```
  ### Added
  - `bdk_core::ChainQuery` trait — generic sans-IO interface for chain verification queries
  - `bdk_core::ChainRequest` / `ChainResponse` type aliases
  - `CanonicalTask` — phase 1 sans-IO canonicalization (determines canonical txs)
  - `CanonicalViewTask` — phase 2 sans-IO canonicalization (resolves chain positions)
  - `Canonical<A, P>` generic container with `CanonicalTxs<A>` and `CanonicalView<A>` aliases
  - `LocalChain::canonicalize()` — drives any `ChainQuery` implementor
  - `LocalChain::canonical_view()` — convenience method for full two-phase canonicalization

  ### Changed
  - **Breaking:** Replace `TxGraph::canonical_iter()` / `TxGraph::canonical_view()` with `TxGraph::canonical_task()`
  - **Breaking:** Canonicalization now uses a two-phase sans-IO process via `ChainQuery`
  - **Breaking:** `ChainQuery`, `ChainRequest`, `ChainResponse` have no generics (use `BlockId` directly)
  - **Breaking:** Chain tip moved from `ChainRequest` to `ChainQuery::tip()`

  ### Removed
  - **Breaking:** `ChainOracle` trait and all implementations
  - **Breaking:** `CanonicalIter` type and `canonical_iter` module
  - **Breaking:** `TxGraph::try_canonical_view()` and `TxGraph::canonical_view()` methods
  - **Breaking:** `CanonicalView::new()` public constructor
```

#### All Submissions:

#### New Features:


crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,483 @@
...
+        /// The chain tip to check against.
+        chain_tip: BlockId,
+    },
+}
+
+/// A response to a canonicalization request.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum CanonicalizationResponse {
+    /// Response to IsBlockInChain request.
+    IsBlockInChain(Option<bool>),
```
evanlinjin: How do we associate responses to their associated requests? I would imagine that some chain sources may not respond in order.

oleonardolima: Good point! It should work fine for now because it is not optmized for batching yet, though I still need to rework this enum for Request/Response, still not happy with it.

oleonardolima: I've reworked the Request/Response API.

crates/chain/src/canonical.rs
```diff
@@ -76,115 +74,24 @@ pub struct CanonicalView<A> {
...
-                        Ok(Some(false)) | Ok(None) => None,
-                        Err(err) => Some(Err(err)),
-                    }
-                })
-                .transpose()
+    /// This internal constructor is used by [`CanonicalizationTask`] to build the view
+    /// after completing the canonicalization process. It takes the processed transaction
+    /// data including the canonical ordering, transaction map with chain positions, and
+    /// spend information.
+    pub(crate) fn new(
```
evanlinjin, line 215: Nit: I think we can remove this and make all fields `pub(crate)`.

crates/core/src/chain_query.rs
```diff
@@ -0,0 +1,68 @@
...
+    ///
+    /// This method should return `Some(request)` if more information is needed,
+    /// or `None` if no more queries are required.
+    fn next_query(&mut self) -> Option<ChainRequest<B>>;
+
+    /// Resolves a query with the given response.
+    ///
+    /// This method processes the response to a previous query request and updates
+    /// the internal state accordingly.
+    fn resolve_query(&mut self, response: ChainResponse<B>);
```
evanlinjin: We should probably mention that:

1. Queries need to be resolved in order.
2. The same `ChainRequest` will be returned from `next_query` if it's not resolved.

crates/bitcoind_rpc/examples/filter_iter.rs
```diff
@@ -69,7 +69,9 @@ fn main() -> anyhow::Result<()> {
     println!("\ntook: {}s", start.elapsed().as_secs());
     println!("Local tip: {}", chain.tip().height());
 
-    let canonical_view = graph.canonical_view(&chain, chain.tip().block_id(), Default::default());
+    let chain_tip = chain.tip().block_id();
+    let task = graph.canonicalization_task(chain_tip, Default::default());
+    let canonical_view = chain.canonicalize(task);
```
evanlinjin: What do you think about the following naming:

* `CanonicalizationTask` -> `CanonicalResolver`.
* `TxGraph::canonicalization_task` -> `TxGraph::resolver`.
* `LocalChain::canonicalize` -> `LocalChain::resolve`.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,531 @@
...
+    unprocessed_assumed_txs: Box<dyn Iterator<Item = (Txid, Arc<Transaction>)> + 'g>,
+    unprocessed_anchored_txs:
+        Box<dyn Iterator<Item = (Txid, Arc<Transaction>, &'g BTreeSet<A>)> + 'g>,
+    unprocessed_seen_txs: Box<dyn Iterator<Item = (Txid, Arc<Transaction>, u64)> + 'g>,
+    unprocessed_leftover_txs: VecDeque<(Txid, Arc<Transaction>, u32)>,
+
+    canonical: CanonicalMap<A>,
+    not_canonical: NotCanonicalSet,
+
+    pending_anchor_checks: VecDeque<(Txid, Arc<Transaction>, Vec<A>)>,
```
evanlinjin: Nit: What do you think about renaming this to `pending_anchor_queries`? This makes it clearer that we are checking against the remote.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,531 @@
...
+                            )
+                            .confirmation_height_upper_bound(),
+                    ))
+                }
+            }
+        }
+    }
+
+    fn is_finished(&mut self) -> bool {
+        self.pending_anchor_checks.is_empty() && self.unprocessed_anchored_txs.size_hint().0 == 0
```
evanlinjin: Looking throught the docs for `.size_hint`, I don't think we should use this here - it seems unreliable.

~~Additionally, `process_anchored_txs()` just moves items from `unprocessed_anchored_txs` to `pending_anchor_checks` one at a time, skipping already canonicalized txs. How about we just collect everything upfront? This way, we will know exactly how many is left by looking at that field and not relying on `.size_hint`.~~ **Edit:** Turns out `pending_anchor_checks` is also populated in `.mark_canonical` for marking a transitively-canonical tx.

crates/chain/src/local_chain.rs
```diff
@@ -96,6 +97,80 @@ impl<D> ChainOracle for LocalChain<D> {
...
+    /// # use bitcoin::hashes::Hash;
+    /// # let tx_graph: TxGraph<BlockId> = TxGraph::default();
+    /// # let chain = LocalChain::from_blocks([(0, bitcoin::BlockHash::all_zeros())].into_iter().collect()).unwrap();
+    /// let chain_tip = chain.tip().block_id();
+    /// let task = CanonicalizationTask::new(&tx_graph, chain_tip, CanonicalizationParams::default());
+    /// let view = chain.canonicalize(task);
+    /// ```
+    pub fn canonicalize<A: Anchor>(
+        &self,
+        mut task: CanonicalizationTask<'_, A>,
```
evanlinjin: Trait here!

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,531 @@
...
+
+        if detected_self_double_spend {
+            // Undo changes
+            for (txid, _, _) in staged_canonical {
+                self.canonical.remove(&txid);
+            }
+            for txid in undo_not_canonical {
+                self.not_canonical.remove(&txid);
+            }
+        } else {
```
evanlinjin: **Nit:** Have the `detected_self_double_spend` early return instead of having the `else` branch.

**Rationale:** Early return is easier to read and results in less nesting.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,531 @@
...
+
+            match best_anchor {
+                Some(best_anchor) => {
+                    self.confirmed_anchors.insert(txid, best_anchor.clone());
+                    if !self.is_canonicalized(txid) {
+                        self.mark_canonical(txid, tx, CanonicalReason::from_anchor(best_anchor));
+                    }
+                }
+                None => {
+                    self.unprocessed_leftover_txs.push_back((
```
evanlinjin: No need to hit this branch if it's a transitively-canonical tx.

evanlinjin (commented):
Great work.

This is my initial round of reviews.

Are you planning to introduce topological ordering in a separate PR?

crates/core/src/chain_query.rs
```diff
@@ -0,0 +1,68 @@
...
+use alloc::vec::Vec;
+
+/// A request to check which block identifiers are confirmed in the chain.
+///
+/// This is used to verify if specific blocks are part of the canonical chain.
+/// The generic parameter `B` represents the block identifier type, which defaults to `BlockId`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ChainRequest<B = BlockId> {
+    /// The chain tip to use as reference for the query.
+    pub chain_tip: B,
```
evanlinjin: I'm wondering whether it will be better to remove this from `ChainRequest` and have a new method on `ChainQuery` called something like `query_tip(&self) -> BlockId`.

**Rationale:** I'm assuming that we always want this tip to be consistent across the lifetime of the `ChainQuery` impl.

**Counter-argument:** We may want to have a `ChainQuery` impl which queries across conflicting chains - what would this be used for though?

WDYT

crates/chain/src/canonical_task.rs
evanlinjin: As discussed, the logic after resolving a query should be different depending on which stage we are in.

crates/core/src/chain_query.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// The block identifiers to check for confirmation in the chain.
+    pub block_ids: Vec<B>,
+}
+
+/// Response containing the best confirmed block identifier, if any.
+///
+/// Returns `Some(B)` if at least one of the requested blocks
+/// is confirmed in the chain, or `None` if none are confirmed.
+/// The generic parameter `B` represents the block identifier type, which defaults to `BlockId`.
+pub type ChainResponse<B = BlockId> = Option<B>;
```
evanlinjin: I think it's better API to have:

```rust
pub type ChainResponse<B = BlockId> = Vec<(B, bool)>;
```

**Rationale:** We only care about whether a block is in the chain or not. However, with `Option<B>`, we now need to worry about the order the data is coming in.

**Counter-argument:** This increasing internal complexity.

**Counter-counter-argument:** No, this decreases internal complexity - allowing us to represent things in the most simplistic manner internally - whether something exists or not.

```rust
    /// We don't have it yet. This is our `ChainRequest`
    pending_blocks: BTreeSet<BlockId>,
    /// We have it!
    blocks: HashMap<BlockId, bool>,
```

At the anchored-tx/transitive-tx stages, we try to advance with the `blocks` we have. If we are missing anything, just populate `pending_blocks` then try again later. Very simple.

crates/chain/src/canonical_task.rs
oleonardolima: I've addressed this a new way, as we discussed the stage processing.

crates/chain/src/canonical_task.rs
oleonardolima: As discussed, I've introduced the stages and this sort-of become the `unprocessed_transitively_anchored_txs` queue.

crates/chain/src/canonical_task.rs
oleonardolima: I should've addressed this, let me double check.

crates/bitcoind_rpc/examples/filter_iter.rs
oleonardolima: I've been thinking about this, I agree with the `CanonicalResolver`, though the `TxGraph::resolver` and `LocalChain::resolve` seems a bit off.

What do you think about (?):

* CanonicalResolver
* TxGraph::canonical_resolver
* LocalChain::canonical_resolve

oleonardolima:
> Are you planning to introduce topological ordering in a separate PR?

Yes, it would be best to add it after finalizing the API, as it can be done internally.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,644 @@
...
+                    "resolve_query called for stage {:?} which doesn't generate queries",
+                    self.current_stage
+                );
+            }
+        }
+    }
+
+    fn is_finished(&mut self) -> bool {
+        // Try to advance stages first
+        self.try_advance();
```
evanlinjin: If we call `.try_advance` at the end of `.resolve_query` and right after constructing `CanonicalTask`, can we avoid this here?

Having a method that checks the status/state of something be mutable is quite odd.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,644 @@
...
+    }
+
+    fn process_leftover_txs(&mut self) {
+        while let Some((txid, tx, height)) = self.unprocessed_leftover_txs.pop_front() {
+            if !self.is_canonicalized(txid) && !tx.is_coinbase() {
+                let observed_in = ObservedIn::Block(height);
+                self.mark_canonical(txid, tx, CanonicalReason::from_observed_in(observed_in));
+            }
+        }
+    }
```
evanlinjin: **Nit:** These methods are only needed to be called in `.try_advance`. I would inline them there to improve readability.

`process_assumed_txs` is currently not in `.try_advance`, but it really seems like it belongs there.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,644 @@
...
+    unprocessed_leftover_txs: VecDeque<(Txid, Arc<Transaction>, u32)>,
+    unprocessed_transitively_anchored_txs: VecDeque<(Txid, Arc<Transaction>, &'g BTreeSet<A>)>,
+
+    canonical: CanonicalMap<A>,
+    not_canonical: NotCanonicalSet,
+
+    // Store canonical transactions in order
+    canonical_order: Vec<Txid>,
+
+    // Track which transactions have direct anchors (not transitive)
```
evanlinjin: *direct and canonical anchors

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,644 @@
...
+                    if !self.direct_anchors.contains_key(txid) {
+                        self.unprocessed_transitively_anchored_txs.push_back((
+                            *txid,
+                            tx.clone(),
+                            anchors,
+                        ));
+                    }
+                }
+            }
+        }
```
evanlinjin: **Nit:** Since we are iterating over all canonical txs anyway, I would just do this right before transitioning to the `TransitivelyAnchoredTxs` stage.

crates/core/src/chain_query.rs
```diff
@@ -0,0 +1,65 @@
...
+    ///
+    /// This method processes the response to a previous query request and updates
+    /// the internal state accordingly.
+    fn resolve_query(&mut self, response: Self::Response);
+
+    /// Returns true if the query process is complete and ready to finish.
+    ///
+    /// The default implementation returns `true` when there are no more queries needed.
+    /// Implementors can override this for more specific behavior if needed.
+    fn is_finished(&mut self) -> bool {
```
evanlinjin: Ideally, this doesn't need to be `mut`.

evanlinjin (commented):
This is looking much better already!

crates/core/src/chain_query.rs
ValuedMammal: Yes it seems weird that the request would include the chain tip when the chain oracle should be aware of its own tip.

crates/core/src/chain_query.rs
oleonardolima: Indeed, it looks like a better API. I'll give it a try and see if I face any other tradeoffs in the code.

crates/core/src/chain_query.rs
```diff
@@ -0,0 +1,68 @@
...
+
+use crate::BlockId;
+use alloc::vec::Vec;
+
+/// A request to check which block identifiers are confirmed in the chain.
+///
+/// This is used to verify if specific blocks are part of the canonical chain.
+/// The generic parameter `B` represents the block identifier type, which defaults to `BlockId`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ChainRequest<B = BlockId> {
```
oleonardolima: @luisschwab Overall, do you think these APIs would support bdk-floresta? Supposing we choose the path to use a task similar to CanonicalizationTask for fetching the proofs (e.g FetchUtreexoProofsTask).

If you'd like to give it a try you can build on top of an example, like those in: https://github.com/bitcoindevkit/bdk/pull/2046

@evanlinjin I think that we should have a single ChainQuery, that would handle different chain implementations, but maybe that would be way too generic, and become a footgun ? WDYT ?

crates/core/src/chain_query.rs
evanlinjin: @ValuedMammal although the chain oracle knows the tip, we want the tip to stay consistent across requests. We are just deferring the responsibility to the `ChainQuery` to make the tip consistent (instead of having each `ChainRequest` be responsible for it).

crates/core/src/chain_query.rs
evanlinjin: You'll need to be a bit more specific otherwise I'm not sure exactly how you intend on using `ChainQuery` for fetching proofs, and why doing it inside a trait is the best option. Also, is sharing the same trait necessary - what does this achieve?

crates/core/src/chain_query.rs
ValuedMammal: Rather than pass `chain_tip` to the `CanonicalizationTask` constructor, where it is merely copied into every request, I think the `chain_tip` parameter could be passed to `LocalChain::canonicalize`.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,644 @@
...
+                        chain_tip: self.chain_tip,
+                        block_ids,
+                    });
+                }
+                None
+            }
+            CanonicalStage::SeenTxs | CanonicalStage::LeftOverTxs | CanonicalStage::Finished => {
+                // These stages don't need queries
+                None
+            }
```
ValuedMammal: We need to ensure we don't inadvertently return `None` from `next_query` before reaching the `Finished` state.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,644 @@
...
+    }
+
+    fn is_finished(&mut self) -> bool {
+        // Try to advance stages first
+        self.try_advance();
+        // Check if we've reached the Finished stage
+        self.current_stage == CanonicalStage::Finished
+    }
+
+    fn finish(self) -> Self::Output {
```
ValuedMammal, line 196: Maybe `finish` should error to signal to the caller if we haven't reached the `Finished` state. I tried adding this assertion and found that it doesn't always hold. 

```suggestion
    fn finish(self) -> Self::Output {
        assert!(matches!(self.current_stage, CanonicalStage::Finished));
```

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,644 @@
...
+                    false // Stop advancing
+                }
+                CanonicalStage::Finished => {
+                    false // Already finished, nothing to do
+                }
+            };
+
+            if !advanced {
+                break;
+            }
```
ValuedMammal: suggestion: Rename to `poll_state` and return the current stage after trying to advance.

<details>
  <summary>expand</summary>

```rust
fn poll_state(&mut self) -> CanonicalState {
    loop {
        match self.current_state {
            CanonicalState::AnchoredTxs => {
                // Still have work, don't advance
                if !self.unprocessed_anchored_txs.is_empty() {
                    return self.current_state;
                // Continue to next stage
                } else {
                    self.current_state = CanonicalState::SeenTxs;
                    continue;
                }
            }
            // Process all seen transactions at once
            CanonicalState::SeenTxs => {
                self.process_seen_txs();

                // Continue to next stage
                self.current_state = CanonicalState::LeftOverTxs;
                continue;
            }
            // Process all leftover transactions at once
            CanonicalState::LeftOverTxs => {
                self.process_leftover_txs();

                // Continue to next stage
                self.current_state = CanonicalState::TransitivelyAnchoredTxs;
                continue;
            }
            CanonicalState::TransitivelyAnchoredTxs | CanonicalState::Finished => {
                if self.unprocessed_transitively_anchored_txs.is_empty() {
                    self.current_state = CanonicalState::Finished;
                }
                return self.current_state;
            }
        }
    }
}
```

</details>

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,644 @@
...
+                };
+
+                // Determine chain position based on reason
+                let chain_position = match reason {
+                    CanonicalReason::Assumed { descendant } => match descendant {
+                        Some(_) => match self.direct_anchors.get(txid) {
+                            Some(confirmed_anchor) => ChainPosition::Confirmed {
+                                anchor: confirmed_anchor,
+                                transitively: None,
+                            },
```
ValuedMammal: This `ChainPosition::Confirmed` branch depends having a value of `descendant`, but it could be the case that we have a direct anchor but no descendant, in which case the reason should still map to `Confirmed`?

ValuedMammal (commented):
I feel that `ChainQuery` trait is not fully fleshed out and could be left out without hurting the overall feature. IMO there should be 2 implementors of a trait to justify its existence, otherwise how do we know that the interface will be useful in another context? For instance should `ChainResponse` also be an associated type? Is it always the case that the request and response have the same type of `B`?

I made a few comments about making the control flow a bit tighter to ensure that we've processed all of the queues before finishing the view.

`CanonicalReason::Assumed` should map to `ChainPosition::Confirmed` if a direct anchor exists (whether or not it comes from a `descendant`). I can look into raising an issue for this. #2088

crates/chain/src/canonical_task.rs
evanlinjin: Makes sense. I don't think introducing an error type here makes sense though. Maybe returning `-> Result<Self::Output, Self>` is enough.

crates/chain/src/canonical_task.rs
evanlinjin: Yes. The logic in `.next_query` should be moved into the `.try_advance` loop.

evanlinjin:
> I feel that `ChainQuery` trait is not fully fleshed out and could be left out without hurting the overall feature. IMO there should be 2 implementors of a trait to justify its existence, otherwise how do we know that the interface will be useful in another context? For instance should `ChainResponse` also be an associated type? Is it always the case that the request and response have the same type of `B`?

In the future, we would like chain source crates to impl their own `.canonicalize` without needing to depend on `bdk_chain`. The `ChainQuery` trait is essentially a dependency inversion boundary.

Your heuristic applies to *behavioural polymorphsim*, but `ChainQuery` exists for *architectural decoupling*. Whether it's useful elsewhere is an after-thought.

crates/core/src/chain_query.rs
evanlinjin: ### Some thoughts on this and MTP

We can only get MTP time from the chain - and the ability to get the MTP is a requirement for timelocked transactions.

To calculate MTP, we need a response that also includes the block time. I think it's better to have the request type to always be `u32` (height) and the response data just needs to impl `ToBlockHash` and optionally `ToBlockTime` (https://github.com/bitcoindevkit/bdk/pull/2037).

evanlinjin: The alternative is to have a separate trait for MTP calculation, but I think that won't be ideal.

evanlinjin: To do this change, we will need a more "general" `BlockQuery` though - one that just fetches blocks (instead of looking at a bunch of IDs and returning the first one that exists in the best chain).

oleonardolima:
_(in case someone is re-reviewing it)_

I rebased it and noticed that the commit structure/history is not helpful for the reviewers, and some commits could be squashed. I'm working on it and will apply some of the previous review suggestions too.

It'll be easier to go through during the review club.

crates/chain/src/canonical_task.rs
evanlinjin: @oleonardolima this is what it would look like: https://github.com/evanlinjin/bdk/commit/f5048388ad9ac083dcd76aaf01ae32d835d2c340

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,644 @@
...
+    // Track the current stage of processing
+    current_stage: CanonicalStage,
+}
+
+impl<'g, A: Anchor> ChainQuery for CanonicalizationTask<'g, A> {
+    type Output = CanonicalView<A>;
+
+    fn next_query(&mut self) -> Option<ChainRequest> {
+        // Try to advance to the next stage if needed
+        self.try_advance();
```
evanlinjin: Were we going to get rid of this?

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,644 @@
...
+                    // only check anchors we haven't already confirmed
+                    if !self.direct_anchors.contains_key(txid) {
+                        self.unprocessed_transitively_anchored_txs.push_back((
+                            *txid,
+                            tx.clone(),
+                            anchors,
+                        ));
+                    }
+                }
+            }
```
evanlinjin: @oleonardolima You mentioned to me that you aren't happy with this, and @LLFourn brought up splitting things into two steps in the past.

### Problem Description

Currently, `CanonicalizationTask` convolutes two concerns in one step:

1. **Canonicalization** - determining which transactions are canonical and why (`CanonicalReason`).
2. **Position resolution and topological ordering** - mapping reasons into ChainPosition (confirmed vs unconfirmed) and in the future, ordering transactions topologically.

This makes the logic harder to follow than it needs to be. I'm thinking of splitting into two tasks:

- `CanonicalTask` - the core algorithm. Resolves conflicts and outputs `CanonicalTxs` (transactions + `CanonicalReason`, no position data).
- `CanonicalViewTask` - thin wrapper that takes `CanonicalTxs` and resolves reasons into `ChainPosition` (and doing topological ordering in the future), producing the final `CanonicalView`.

  Both types would be public, so callers that don't need position data can stop at step 1.

### Proposed API

```rust
  let canonical_txs = chain.resolve(tx_graph.canonical_task(tip, params));
  let view = chain.resolve(canonical_txs.view_task());
```

Maybe even a helper method on `LocalChain`:

```rust
  impl LocalChain {
      pub fn canonical_view(
          &self,
          tx_graph: &TxGraph<A>,
          tip: BlockId,
          canon_params: CanonicalizationParams,
          view_params: CanonicalViewParams,
      ) -> CanonicalView<A> {
          let canonical_txs = self.resolve(tx_graph.canonical_task(tip, canon_params));
          self.resolve(canonical_txs.view_task(view_params))
      }
  }
```

Additionally, since `CanonicalTxs` and `CanonicalView` will want to share the majority of the methods, we can do something like:

```rust
struct Canonical<A, P> { ... }

type CanonicalTxs<A> = Canonical<A, CanonicalReason<A>>;
type CanonicalView<A> = Canonical<A, ChainPosition<A>>;
```

`FullTxOut` and `CanonicalTx` will also need that `P` generic as well. I would also rename `FullTxOut::chain_position` to `pos`.

crates/chain/src/canonical_task.rs
evanlinjin: I changed my stance on this.

`resolve_query` and `finish` should NOT return errors because the only way they can be called incorrectly is by violating the protocol (e.g. calling `resolve_query` without a preceding `next_query`, or calling `finish` before `next_query` returns `None`).

These are **programming errors** in the caller, not runtime conditions - the caller controls the driving loop and is responsible for upholding the protocol. Programming errors should be caught with `debug_assert!`/panics, not `Result`.

evanlinjin:
### Before ACK and merge (@evanlinjin's list)

- [x] Rename `CanonicalizationTask` to `CanonicalTask`.
- [ ] Consider renaming `ChainQuery` to `ChainTask`.
- [x] Remove and inline single-used private methods: `try_advance`, `process_assumed_txs` (I would add another variant to `CanonicalStage`) `process_last_seen`, `process_leftover_txs`. [^inline]
- [x] Remove `is_finished` (unless if you can provide justification for it's existence). My argument: "No more queries" instantly means "actually finished" - there is no state inbetween to represent.
- [x] Remove `chain_tip` from `ChainRequest`. Add `ChainQuery::tip() -> BlockId` method. [^chaintip]
- [x] Change `CanonicalTask` to return `CanonicalTxs` & introduce `CanonicalViewTask` that returns `CanonicalView`. [^viewtask]
- [x] Remove all generics on `ChainTask` (`ChainQuery`) as we will flesh out this trait in follow-up PRs.

[^inline]: https://github.com/bitcoindevkit/bdk/pull/2038#discussion_r2658462257
[^chaintip]: https://github.com/bitcoindevkit/bdk/pull/2038#discussion_r2401496855
[^viewtask]: https://github.com/bitcoindevkit/bdk/pull/2038#discussion_r2802430801

**Edit:**

I got Claude to look into the majority of these. Let me know if these commits are useful: https://github.com/evanlinjin/bdk/tree/refactor/canonical-iter-api-evanlinjin

crates/chain/src/canonical.rs
oleonardolima: All the `Canonical` fields are now `pub(crate)`, done in: 6d9a7d6f.

crates/chain/src/canonical_task.rs
oleonardolima: I agree with @evanlinjin stance on this one, and it's been solved in the latest commits by: (i) updating the docs; (ii) adding `debug_assert` on `resolve_query()` methods to assert the proper stage transition.

I'm resolving this one, but feel free to re-open it if you think it's still holds @ValuedMammal.

crates/core/src/chain_query.rs
oleonardolima: I'm resolving this one, as it's outdated/stale.

crates/bitcoind_rpc/examples/filter_iter.rs
oleonardolima: Alright, this one is a bit outdated.

As of 031de404 we have:
- `CanonicalizationTask` -> `CanonicalTask`; and also new `CanonicalViewTask`.
- `TxGraph::canonicalization_task` -> `TxGraph::canonical_task`.
- `LocalChain::canonicalize` is still the same, though we now also have `LocalChain::canonical_view`.

I'm fine with these names for now, though if there's no consensus on those we can discuss/change in a follow-up PR.

crates/chain/src/canonical_task.rs
oleonardolima: I'm resolving this as it's been solved, and it's an outdated comment.

crates/core/src/chain_query.rs
oleonardolima: In 1e27917d, the `chain_tip` has been removed from the `ChainRequest`.

crates/core/src/chain_query.rs
oleonardolima: I'm resolving this one as it's outdated, and in the latest commits it's using generics anymore. However, the need/use of generics and maybe a different API need to be discussed and addressed in a follow-up PR.

crates/chain/src/canonical_task.rs
oleonardolima: You've addressed this one, thanks for the commit 😁.

crates/chain/src/canonical_task.rs
oleonardolima: The `is_finished` method has been removed, resolving this comment.

crates/chain/src/canonical_task.rs
oleonardolima: It's been solved in the latest commits.

crates/chain/src/canonical_task.rs
oleonardolima: As it's been splitted into two different tasks, it's not applicable anymore.

crates/core/src/chain_query.rs
oleonardolima: As of the latest commits, `is_finished` has been removed.

crates/core/src/chain_query.rs
oleonardolima: Alright, it won't be applicable as I previously thought. At least for now the `ChainQuery` trait is not generic anymore, if it's required for fetching proofs any other kind of task in the future, it should be addressed in a follow-up issue/PR.

crates/chain/src/canonical_task.rs
oleonardolima: It's been solved in afa097c6, as the processed logic was reworked.

crates/chain/src/canonical_task.rs
oleonardolima: In f07b1b35, when processing the canonical txs and building the `CanonicalView`, it now handles assumed txs properly, if they have a direct anchor.

crates/chain/src/canonical_task.rs
oleonardolima: All the logic is now handled by `next_query` and `CanonicalStage::advance`, so I deem this one outdated.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,499 @@
...
+
+    fn next_query(&mut self) -> Option<ChainRequest> {
+        loop {
+            match self.current_stage {
+                CanonicalStage::AssumedTxs => {
+                    if let Some((txid, tx)) = self.unprocessed_assumed_txs.next() {
+                        if !self.is_canonicalized(txid) {
+                            self.mark_canonical(txid, tx, CanonicalReason::assumed());
+                        }
+                        continue;
```
evanlinjin: Nit: this does nothing.

```suggestion
```

evanlinjin:
Can we get rid of `ChainOracle` in the same PR?

**Edit:** Removed in 37eb136c31ca0390106b8d101e9b6dea75bd26a5

evanlinjin (approved):
ACK 37eb136c31ca0390106b8d101e9b6dea75bd26a5

evanlinjin:
I've also updated the PR description.

crates/core/src/chain_query.rs
oleonardolima: I'm resolving this one, as it's being addressed in https://github.com/bitcoindevkit/bdk/pull/2139

oleonardolima:
> Can we get rid of `ChainOracle` in the same PR?
> 
> **Edit:** Removed in [37eb136](https://github.com/bitcoindevkit/bdk/commit/37eb136c31ca0390106b8d101e9b6dea75bd26a5)

I'm fine with removing it in the same PR, thanks for the commit.

> I've also updated the PR description.

Great, thank you!

crates/chain/src/canonical_view_task.rs
```diff
@@ -0,0 +1,199 @@
...
+                    Some(tx_node) => tx_node,
+                    None => {
+                        debug_assert!(false, "tx node must exist!");
+                        continue;
+                    }
+                };
+
+                // Determine chain position based on reason
+                let chain_position = match reason {
+                    CanonicalReason::Assumed { .. } => match self.direct_anchors.get(txid) {
```
ValuedMammal: You seem to be ignoring the descendant, but it may be the case that the descendant has a direct anchor, making this tx Confirmed transitively.

In case the reason is Assumed, to determine the ChainPosition you would want to:

- First look for a direct anchor of the current tx. If a direct anchor exists, the position must be Confirmed (and transitively None)
- Otherwise look at the descendant txid. If the descendant has a direct anchor, then the position is Confirmed transitively by the descendant's anchor
- If the descendant has no direct anchor, then the position is Unconfirmed. Likewise Unconfirmed if there's no anchor and no descendant

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,499 @@
...
+        self.descendant().is_some()
+    }
+
+    /// Returns true if this reason is [`CanonicalReason::Assumed`].
+    pub fn is_assumed(&self) -> bool {
+        matches!(self, CanonicalReason::Assumed { .. })
+    }
+}
+
+#[cfg(test)]
```
ValuedMammal, line 444:
```suggestion
#[cfg(test)]
#[cfg_attr(coverage_nightly, coverage(off))]
```

crates/chain/src/local_chain.rs
```diff
@@ -69,33 +68,90 @@ impl<D> PartialEq for LocalChain<D> {
...
+    /// let view = chain.canonicalize(canonical_txs.view_task(tx_graph));
+    /// ```
+    pub fn canonical_view<A: Anchor>(
+        &self,
+        tx_graph: &TxGraph<A>,
+        tip: BlockId,
+        params: CanonicalParams,
+    ) -> CanonicalView<A> {
+        let canonical_txs = self.canonicalize(tx_graph.canonical_task(tip, params));
+        self.canonicalize(canonical_txs.view_task(tx_graph))
```
ValuedMammal, line 156: The canonical task and the view task only differ in the stages of canonicalization they handle and the finished output. You should reduce this to a single task architecture. The rationale in 6d9a7d6 is to separate concerns, but it's really the same concern, canonicalizing the TxGraph into a consistent view. That'll prevent having the algorithm spread across multiple modules.

ValuedMammal (changes requested):

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -970,9 +970,12 @@ fn test_tx_conflict_handling() {
     for scenario in scenarios {
         let env = init_graph(scenario.tx_templates.iter());
 
-        let txs = env
-            .tx_graph
-            .canonical_view(&local_chain, chain_tip, env.canonicalization_params.clone())
+        let txs = local_chain
+            .canonical_view(
```
ValuedMammal: You could reuse the canonical view here for the remaining test assertions.

crates/chain/src/local_chain.rs
nymius: I agree with this, the canonicalization task doesn't have a meaning without the view task.

crates/chain/src/canonical_view_task.rs
nymius: > You seem to be ignoring the descendant, but it may be the case that the descendant has a direct anchor, making this tx Confirmed transitively.

I think this depends of the priority we give to the different `CanonicalReason`s. As it is expressed right now, `CanonicalReason::Asumed` has higher priority than the others.

> In case the reason is Assumed, to determine the ChainPosition you would want to:
> 
> * First look for a direct anchor of the current tx. If a direct anchor exists, the position must be Confirmed (and transitively None)

In case we raise `CanonicalReason::Anchor` in the hierarchy, then is as easy as to make `AssumedTxs` also return a query for these txids.

> * Otherwise look at the descendant txid. If the descendant has a direct anchor, then the position is Confirmed transitively by the descendant's anchor

Again, in case we raise `CanonicalReason::Anchor` in the hierarchy, this case is partially covered by `mark_canonical` when walking up the `Ancestor` queue:

```rust
            |_: usize, tx: Arc<Transaction>| -> Option<Txid> {
                let this_txid = tx.compute_txid();
                let this_reason = if is_starting_tx {
                    is_starting_tx = false;
                    reason.clone()
                } else {
                    // This is an ancestor being marked transitively
                    reason.to_transitive(starting_txid)
                };

                use crate::collections::hash_map::Entry;
                let canonical_entry = match self.canonical.entry(this_txid) {
                    // Already visited tx before, exit early.
                    Entry::Occupied(_) => return None, // Here we can compare and replace `CanonicalReason::Assumed` by `this_reason` if `this_reason` is `CanonicalReason::Anchor`. 
                    Entry::Vacant(entry) => entry,
                };
```
>  * If the descendant has no direct anchor, then the position is Unconfirmed. Likewise Unconfirmed if there's no anchor and no descendant

Are you proposing a different category here? Or with Unconfirmed you refer to `CanonicalReason::Assumed`?

crates/chain/src/local_chain.rs
evanlinjin: I disagree that these are the same concern. They solve different problems and are diverging further over time.

**Phase 1 (`CanonicalTask`)** is conflict resolution — it walks the tx DAG, detects conflicts/self-double-spends, and determines *which* transactions are canonical and *why* (`CanonicalReason`). This is the complex part (~500 lines).

**Phase 2 (`CanonicalViewTask`)** is view construction — it takes the resolved canonical set and builds the final `CanonicalView` with concrete `ChainPosition`s. This is a simpler transformation (~200 lines) that is growing its own responsibilities: #2139 adds MTP computation (fetching headers, computing median-time-past per confirmed height), and we plan to add topological sorting here too. None of that belongs in the conflict resolution phase.

The intermediate `CanonicalTxs` output is also intentionally useful on its own — callers who want to inspect canonicalization reasons without paying for full view construction can stop after phase 1.

A merged task would need 6 stages (`AssumedTxs` → `AnchoredTxs` → `SeenTxs` → `LeftOverTxs` → `ResolvingPositions` → `Finished`), a larger state machine mixing conflict resolution state with position resolution state, and a more complex `finish()`. The net result would be lines of interleaved concerns instead of two focused types.

crates/chain/src/canonical_view_task.rs
evanlinjin: @nymius I think you're conflating two different things here. ValuedMammal's comment is about **phase 2 position resolution** (`ChainPosition`), not about phase 1 canonicality priority (`CanonicalReason`).

By the time we're in `CanonicalViewTask::finish()`, canonicality is already decided — we know the tx is canonical and we know its reason is `Assumed`. The question is purely: what is its **`ChainPosition`** — `Confirmed` or `Unconfirmed`?

An `Assumed` tx can still be `Confirmed` if it or its descendant has an anchor in the best chain. The current code checks `direct_anchors.get(txid)` (the tx's own anchor) but ignores the `descendant` field entirely. If the descendant has a resolved anchor in `direct_anchors`, the tx should be `Confirmed { anchor: descendant_anchor, transitively: Some(descendant) }`.

This has nothing to do with raising `CanonicalReason::Anchor` in the hierarchy or changing phase 1 processing. And "Unconfirmed" in ValuedMammal's comment refers to `ChainPosition::Unconfirmed`, not a `CanonicalReason`.

evanlinjin: @ValuedMammal thanks for pointing out that bug! I do see a logical flaw with your proposed fix though.

Given a scenario where we have txs (A, B and C), where B spends A and C spends B, C is assumed to be canonical, C is unconfirmed and B is anchored to the best chain.

With your logic when looking at A, the descendant field always points to the original assumed tx (C), not to the intermediate tx (B) that actually has the anchor. So checking `direct_anchors.get(descendant)` only works when the assumed tx itself has the anchor, not when an intermediate ancestor does.

ValuedMammal: You'd have to initiate a `walk_descendants` routine to find the first descendant having a direct anchor, if one exists, otherwise accept that the ChainPosition can't be relied on for Assumed txs.

oleonardolima: @ValuedMammal Great, thanks for finding this one! I noticed that unfortunately we also have this issue in both `master` and `release/chain-0.23.x` where the scenario above would resolve to: txA->unconfirmed ; txB->confirmed ; txC->unconfirmed.

I'm working on a test to assert for these scenarios.

> You'd have to initiate a `walk_descendants` routine to find the first descendant having a direct anchor, if one exists, otherwise accept that the ChainPosition can't be relied on for Assumed txs.

I'll give a try to this one, but it's valid to note that we can only walk  the descendants that are included in the canonical txs.

crates/chain/src/canonical_task.rs
oleonardolima: I addressed this in 470ccd12b85979d1a8f90b08cdd34cf1a65c7002.

crates/chain/tests/test_tx_graph_conflicts.rs
oleonardolima: I addressed this in 470ccd12b85979d1a8f90b08cdd34cf1a65c7002.

crates/chain/src/canonical_view_task.rs
oleonardolima: > I think this depends of the priority we give to the different `CanonicalReason`s. As it is expressed right now, `CanonicalReason::Asumed` has higher priority than the others.

I have been thinking in depth about this, and @nymius has a good comment here. At least for now we can focus on handling these scenarios in the position resolution.

However we probably need to discuss/review the primitives for the canonicalization algorithm in the future, in these scenarios where a given txA has been resolved to `CanonicalReason::Assumed` but could also resolved to `CanonicalReason::Anchored` (both transitively), should the assumed always have higher priority ?

oleonardolima: I've pushed the initial fix in 6f8ef2667e7cc707fad5fa73b33df0b4c87b3c45

ValuedMammal:
I can take another look at it.

ValuedMammal:
While reviewing I found it easier to structure the commits like this.

- [feat: Introduce CanonicalTask](https://github.com/bitcoindevkit/bdk/commit/1c1b65f543e8be92ef37c408f8b57ea87d32aed6)
> Introduces CanonicalTask, a new type that encapsulates block-anchor verification as an explicit query/response loop. The task walks the transaction graph in priority-ordered stages (assumed, anchored, seen, leftover) and emits batches of BlockIds for the caller to verify against a chain oracle, accepting responses via resolve_query. Once driven to completion, finish() returns a CanonicalTxs which can be converted to a CanonicalView via .view().

- [feat!: Update usage of CanonicalTask](https://github.com/bitcoindevkit/bdk/commit/d0248fea5e78d286c264420242de8e605901a1ee)
> Migrates the entire codebase to the new CanonicalTask-based API, removing the now-superseded canonical_iter.rs and canonical_view.rs modules. Introduces canonical.rs to house the unified Canonical<A, T, P> type (aliased as CanonicalTxs and CanonicalView) along with the CanonicalReason-to-ChainPosition resolution in .view(). All consumers — tests, benchmarks, examples, and chain-sync integrations — are updated accordingly.

- [refactor!: Remove ChainOracle](https://github.com/bitcoindevkit/bdk/commit/d1290281ec22fe6c3f1a7cf7a7b44b85e65705d5)
> Removes the ChainOracle trait, which is no longer needed now that canonicalization is driven through CanonicalTask's explicit query/response interface rather than a generic trait. LocalChain's is_block_in_chain and get_chain_tip methods are retained as plain inherent methods (returning Option instead of Result<Option, Infallible>), simplifying the API surface without losing any functionality.

crates/chain/src/local_chain.rs
ValuedMammal: ResolvingPositions is just generating more block queries as in the first task. Instead, when marking a tx canonical, anchors of transitive and assumed txs can be added to the back of the unprocessed anchor queue. In resolve_query if the tx is already canonical, then it can have its reason upgraded to include the best anchor.

In `finish` you can fetch the TxNode from the tx graph to be passed to the Canonical set. To complete the view just have `CanonicalTxs::view -> CanonicalView`.

evanlinjin: @ValuedMammal My arguments are the following:

1. The anchors aren't free. During canonicalization, you only have `CanonicalReason` — resolving actual anchors/confirmation heights requires additional `is_block_in_chain` queries (more IO). Merging the tasks doesn't eliminate that work, it just convolutes chain-position-fetching with canonicalization.

2. Simplicity through focus. Task 1 answers "is this tx canonical?" Task 2 answers "at what height is it confirmed?" These are distinct questions with distinct outputs. Keeping them separate makes each easier to understand.

4. Not all callers need confirmation heights. Some consumers only care about the canonical set, not positions. The split lets them skip the second pass entirely.

5. The duplicate-query cost is solvable orthogonally. PR #2139 shows that caching `is_block_in_chain` results addresses the efficiency concern without coupling the two phases together.

crates/chain/src/canonical_view_task.rs
```diff
@@ -152,16 +153,53 @@ impl<'g, A: Anchor> ChainQuery for CanonicalViewTask<'g, A> {
...
+                                            self.canonical_txs
+                                                .contains_key(&desc_txid)
+                                                .then(|| {
+                                                    self.direct_anchors
+                                                        .get(&desc_txid)
+                                                        .map(|anchor| (desc_txid, anchor))
+                                                })
+                                                .flatten()
+                                        },
+                                    )
```
evanlinjin, line 182: This will stop when a descendant doesn't have a direct anchor which is incorrect as a direct anchor will follow after.

### What should happen

* The closure should return everything (no filter).
* `.next` should stop once it hits a direct anchor.

### What should be figured out in a follow up

Although `TxDescendants` does a BFS (which means we will find the "shallowest" confirmed anchor) - this does not guarantee it will be the "earliest confirmed" anchor (which is the ideal value to have).

crates/chain/src/canonical_view_task.rs
```diff
@@ -0,0 +1,237 @@
...
+                            last_seen: None,
+                        },
+                    },
+                };
+
+                view_txs.insert(*txid, (tx.clone(), chain_position.cloned()));
+            }
+        }
+
+        CanonicalView::new(self.tip, view_order, view_txs, self.spends.clone())
```
evanlinjin, line 235: Nit

```suggestion
        CanonicalView::new(self.tip, view_order, view_txs, self.spends)
```

crates/core/src/chain_query.rs
```diff
@@ -0,0 +1,40 @@
...
+/// Calling `resolve_query` or `finish` out of sequence is a programming error.
+pub trait ChainQuery {
+    /// The final output type produced when the query process is complete.
+    type Output;
+
+    /// Returns the chain tip used as the reference point for all queries.
+    fn tip(&self) -> BlockId;
+
+    /// Returns the next query needed, or `None` if no more queries are required.
+    fn next_query(&mut self) -> Option<ChainRequest>;
```
evanlinjin, line 31: Suggested by Claude

```suggestion
    #[must_use]
    fn next_query(&mut self) -> Option<ChainRequest>;
```

evanlinjin (changes requested):

crates/chain/src/canonical_view_task.rs
evanlinjin: The initial use-case of `assumed_canonical` was for crafting RBF and CPFP transactions. We want to assume that the transaction we are replacing or adding a child to is available for coin selection (CPFP) or excluded from coin selection (RBF).

- If confirmed transactions have a higher priority, and a conflict is confirmed, the "assumed canonical" transaction will be non-canonical.
- If assumed transactions have a higher priority, confirmed transactions that conflict will be considered non-canonical.

I think there are practical (but niche) scenarios where we want "assumed canonical" transactions to supersede confirmed transactions.

- The caller may want to create a CPFP transaction that bumps the fee of a transaction which was evicted by a confirmed conflict (in case the confirmed tx gets unconfirmed).
- If the wallet detects a confirmed tx is at risk of being reorged (e.g. very few confirmations), the caller might want to assume a pre-signed replacement is canonical so coin selection uses the replacement's outputs preemptively.

Additionally, "assumed canonical" is simpler to understand than "assumed canonical, unless a conflict is confirmed".

evanlinjin:
@oleonardolima Thanks for taking the PR this far. To push this forward, I would like to take over.

oleonardolima:
> @oleonardolima Thanks for taking the PR this far. To push this forward, I would like to take over.

@evanlinjin Alright, I'll take a look into your pushed commits.

evanlinjin (approved):
ACK a5d91e9528c8f60c0927c6694b800397d97ea945

However, I agree with @ValuedMammal that the commits are a bit messy.

oleonardolima:
> ACK [a5d91e9](https://github.com/bitcoindevkit/bdk/commit/a5d91e9528c8f60c0927c6694b800397d97ea945)
> 
> However, I agree with @ValuedMammal that the commits are a bit messy.

Alright, so the only thing missing is fixing the commit history. I'll handle it and ping VM's for review.

crates/core/src/chain_query.rs
nymius: Marking as resolved as the change has already been done.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,500 @@
...
+    pub fn from_observed_in(observed_in: ObservedIn) -> Self {
+        Self::ObservedIn {
+            observed_in,
+            descendant: None,
+        }
+    }
+
+    /// Contruct a new [`CanonicalReason`] from the original which is transitive to `descendant`.
+    ///
+    /// This signals that either the [`ObservedIn`] or [`Anchor`] value belongs to the transaction's
```
nymius: This is confusing, what's the direction of the transitivity? `ancestor ---> descendant` or `descendant ---> ancestor`?

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,500 @@
...
+    }
+}
+
+/// Modifies the canonicalization algorithm.
+#[derive(Debug, Default, Clone)]
+pub struct CanonicalParams {
+    /// Transactions that will supersede all other transactions.
+    ///
+    /// In case of conflicting transactions within `assume_canonical`, transactions that appear
+    /// later in the list (have higher index) have precedence.
```
nymius, line 48: `have higher index` is misleading, which is the index? the `Txid` (I wouldn't call it an index)

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,500 @@
...
+                };
+
+                use crate::collections::hash_map::Entry;
+                let canonical_entry = match self.canonical.entry(this_txid) {
+                    // Already visited tx before, exit early.
+                    Entry::Occupied(_) => return None,
+                    Entry::Vacant(entry) => entry,
+                };
+
+                // Any conflicts with a canonical tx can be added to `not_canonical`. Descendants
```
nymius, line 295: Can we add a comment header here? i.e.:
```rust
// Prune conflicts
//
// Any conflicts....
```

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,500 @@
...
+
+        if detected_self_double_spend {
+            // Undo changes
+            for (txid, _, _) in staged_canonical {
+                self.canonical.remove(&txid);
+            }
+            for txid in undo_not_canonical {
+                self.not_canonical.remove(&txid);
+            }
+            return;
```
nymius, line 334: I prefer an `else` instead of an early `return` here.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,500 @@
...
+                                None
+                            }
+                        },
+                    )
+                    .run_until_finished()
+                }
+
+                if self.not_canonical.contains(&this_txid) {
+                    // Early exit if self-double-spend is detected.
+                    detected_self_double_spend = true;
```
nymius, line 315: When a self double spend is detected, shouldn't the code stop straight away instead of "running until finished"?

oleonardolima:
@evanlinjin @ValuedMammal I've pushed the new commit history, it's now only 7 commits. I think it's the best we can get without adding more complexity to some commits.

I've tried to reduce it into fewer commits, however it got even messier as some commits had too much of a change/complexity.

At least now, each commit builds/runs successfully and truly builds on top of each other.

--

_NOTE_:
- I think the commits b0e72feccf24da5595b8dd8d1675a6aa0fe405a1 and d45160d2f8f568d8d349f2edbaa9983cf74b4116 could be squashed together, but left as it's for better clarity.
- I think can be dropped 728454dd88930d8cd8c731c61c3fdf13a05ca3a4 and inserted in a follow-up PR.

crates/chain/src/canonical_view_task.rs
```diff
@@ -152,16 +153,52 @@ impl<'g, A: Anchor> ChainQuery for CanonicalViewTask<'g, A> {
...
-                    },
+                    CanonicalReason::Assumed { descendant } => {
+                        match self.direct_anchors.get(txid) {
+                            // it has a direct anchor found
+                            // regardless if it's directly or transitively assumed canonical
+                            Some(anchor) => ChainPosition::Confirmed {
+                                anchor,
+                                transitively: None,
+                            },
+                            None => match descendant {
```
evanlinjin: `descendant == None` means that this transaction may be the one marked as "assumed" and does not mean "this has no descendants". Therefore, we should remove this match statement.

evanlinjin (changes requested):
@oleonardolima I'll just go ahead and change this!

crates/chain/tests/test_canonical_view_task.rs
```diff
@@ -0,0 +1,201 @@
...
+use std::collections::HashSet;
+
+struct Scenario<'a> {
+    name: &'a str,
+    tx_templates: &'a [TxTemplate<'a, BlockId>],
+    exp_canonical_txs: HashSet<&'a str>,
+}
+
+#[test]
+fn test_assumed_canonical_scenarios() {
```
evanlinjin, line 12: This test is a bit odd.

* We have only a single scenario so the loop is dead weight.
* Also, the loop is not reusable because of the hardcoded values within it.

crates/chain/src/canonical.rs
```diff
@@ -23,37 +24,37 @@
...
-    /// The position of this transaction in the chain.
+pub struct CanonicalTx<P> {
+    /// The position of this transaction.
     ///
-    /// This indicates whether the transaction is confirmed (and at what height) or
-    /// unconfirmed (most likely pending in the mempool).
-    pub pos: ChainPosition<A>,
+    /// When `P` is [`ChainPosition`], this indicates whether the transaction is confirmed
+    /// (and at what height) or unconfirmed (most likely pending in the mempool).
+    pub pos: P,
```
nymius, line 50: nit: trivial argument here, but I would use the full name: `position`. I don't like the ambiguity of the type, but I assume it was the best tradeoff.

crates/chain/src/canonical.rs
```diff
@@ -23,37 +24,37 @@
...
-/// A single canonical transaction with its chain position.
+use crate::{spk_txout::SpkTxOutIndex, Anchor, Balance, CanonicalViewTask, ChainPosition, TxGraph};
+
+/// A single canonical transaction with its position.
 ///
 /// This struct represents a transaction that has been determined to be canonical (not
-/// conflicted). It includes the transaction itself along with its position in the chain (confirmed
-/// or unconfirmed).
+/// conflicted). It includes the transaction itself along with its position information.
+/// The position type `P` is generic — it can be [`ChainPosition`] for resolved views,
```
nymius: This character "—" is ambiguous:  U+2014 (em dash). U+002D (hyphen minus) is the one carried by most keyboards.

crates/chain/src/canonical.rs
```diff
@@ -62,151 +63,180 @@ impl<A: Ord> Ord for CanonicalTx<A> {
...
-impl<A: Ord> PartialOrd for CanonicalTx<A> {
+impl<P: Ord> PartialOrd for CanonicalTx<P> {
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+/// A canonical transaction output with position and spend information.
+///
+/// The position type `P` is generic — it can be [`ChainPosition`] for resolved views,
```
nymius: Same than above: this character "—" is ambiguous:  U+2014 (em dash). U+002D (hyphen minus) is the one carried by most keyboards.

crates/chain/src/canonical.rs
nymius, line 50: nit: trivial argument here, but I would use the full name: position. I don't like the ambiguity of the type, but I assume it was the best tradeoff.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,500 @@
...
+            observed_in,
+            descendant: None,
+        }
+    }
+
+    /// Construct a new [`CanonicalReason`] from the original which is transitive to `descendant`.
+    ///
+    /// This signals that either the [`ObservedIn`] or [`Anchor`] value belongs to the transaction's
+    /// descendant, but is transitively relevant.
+    pub fn to_transitive(&self, descendant: Txid) -> Self {
```
nymius, line 406: `transitivity` is a directed relation. Without some clues, it may be hard to understand what's the direction here. I would change `to_transitive` to `by_descendant` or something similar that shows the tx is canonical because its descendant is. So it is clear the canonical reason is transmitted from the descendant to the ancestor.

crates/chain/src/canonical_task.rs
```diff
@@ -0,0 +1,500 @@
...
+
+    /// Constructs a [`CanonicalReason`] from an `observed_in` value.
+    pub fn from_observed_in(observed_in: ObservedIn) -> Self {
+        Self::ObservedIn {
+            observed_in,
+            descendant: None,
+        }
+    }
+
+    /// Construct a new [`CanonicalReason`] from the original which is transitive to `descendant`.
```
nymius, line 402: This is confusing, what's the direction of the transitivity? `ancestor` ---> `descendant` or `descendant ---> `ancestor`?

crates/chain/src/canonical.rs
```diff
@@ -62,151 +63,180 @@ impl<A: Ord> Ord for CanonicalTx<A> {
...
 /// which transactions are canonical (non-conflicted) based on the current chain state and
 /// provides methods to query transaction data, unspent outputs, and balances.
 ///
+/// The position type `P` is generic:
+/// - [`ChainPosition<A>`] for resolved views (aka [`CanonicalView`])
+/// - [`CanonicalReason<A>`](crate::canonical_task::CanonicalReason) for unresolved results (aka
+///   [`CanonicalTxs`])
+///
 /// The view maintains:
 /// - An ordered list of canonical transactions in topological-spending order
```
nymius: This is wrong, the current order is not topological (parent before descendant) nor topological-reversed (descendant before parent).
[This test](https://github.com/nymius/bdk/commit/bafd2c935fbaa651f77b70e274fa44fa2a92efc2) shows how it can be broken with a 4-tx diamond shaped graph.
This can be simplified, [the following test](https://github.com/nymius/bdk/commit/b13a5400469ec4a4d7dedcccf7ba21c3959d3174) breaks the topological assumption again with a 3-tx chain shaped graph.

nymius (changes requested):
Most of the suggestions are nits related to first impressions from the code and code comprehension.
The only significant change I would do here, as @ValuedMammal suggested before, is the Kahn's algorithm addition.
The Canonical struct is claiming something that is not true (see linked tests) if not paired with this algorithm.
There are two chances here, to remove the claim or actually implement the sort.
I'm inclined for the second, as the implementation I've implemented with LLM help is not as complex as I thought: https://github.com/nymius/bdk/commit/dbfabad779d36dc3b25415a898b5f173e0cf7807

It also simplifies the code in canonical task, as some of the order related fields are unnecessary after implementing the sort algorithm.

I've also compared the implementation there with the implementation from https://github.com/bitcoindevkit/bdk/commit/1a77475f087ae311672ff91da5206c49bed6adec.
I've ported some of those test to the current implementation in https://github.com/nymius/bdk/commit/25049a96a59097dac047f25408df879bd28d273f.
I don't use the chain position to disambiguate order positions nor create depth based tiers in the order because I'm not aware of use cases that require this disposition.

ValuedMammal:
Thank you @nymius

crates/chain/src/local_chain.rs
ValuedMammal: I can't comment on the unfinished draft, but shouldn't the query result caching be done in this PR?

oleonardolima:
> Most of the suggestions are nits related to first impressions from the code and code comprehension. The only significant change I would do here, as @ValuedMammal suggested before, is the Kahn's algorithm addition. The Canonical struct is claiming something that is not true (see linked tests) if not paired with this algorithm. There are two chances here, to remove the claim or actually implement the sort.

@nymius thanks for the review, I'll take a look on the comments. In my opinion, this PR's already got complex enough by itself, I'd suggest adding the topological order support as a follow-up, as it was intended by #2201, I'd say the same regarding some canonicalization algorithm fixes that I've added alongside here.

crates/chain/src/canonical_task.rs
evanlinjin: This is a stylistic choice. I find the early return clearer because it makes the short-circuit explicit.

crates/chain/src/canonical.rs
evanlinjin: We currently use `pos` elsewhere in the codebase. To keep things consistent, I would just leave it as is here. If we want to change this, let's do it in one sweep in a separate PR.

evanlinjin: Not in this PR. That was the field name before. Create a ticket if need be. 

My opinion: Generally, I would agree that full names are better. However, `pos` is a common abbreviation. It's even used in stdlib: https://doc.rust-lang.org/beta/std/io/trait.Seek.html

crates/chain/src/canonical_task.rs
evanlinjin: How about rename both the field and method to be `via_descendant`?

crates/chain/tests/test_canonical_view_task.rs
oleonardolima: Agreed, I was aiming to have different scenarios but didn't add those, as thought it was redundant.

oleonardolima: Also, I vote to drop this commit on this PR and put it strictly in another PR fixing the canonicalization algorithm, it'd be easier to backport it.

nymius:
> @nymius thanks for the review, I'll take a look on the comments. In my opinion, this PR's already got complex enough by itself, I'd suggest adding the topological order support as a follow-up, as it was intended by #2201, I'd say the same regarding some canonicalization algorithm fixes that I've added alongside

I saw the topological order issue is on the same milestone than this one, so I'm ok with this, as long they go together in the same release. I would still remove any topological order claims from docs before that happens.

crates/chain/tests/test_canonical_view_task.rs
oleonardolima: addressed in 8c0e3f6780c058637ee14a32c23ff58b2d658e59

crates/chain/src/canonical.rs
nymius: Have you considered removing the `topological-spending order` attribution?

oleonardolima: Yes, however it's not a blocker as it's being addressed in a follow-up and will land in the same release.

nymius: I insist, for me this is a blocker. I'm against merging `topological-spending order` claims if the code does not fulfill them. The PR wouldn't be self contained in that case.
If we remove the comments here, we can always re-add them later once the topological order is implemented.

oleonardolima: Alright, I'll update it then. Though it's valid to note that doc's in master was already considering this feature being added in the future.

oleonardolima: done in 92f7b2a657c0819df86a35fb3d334ab1832bbe95

nymius (approved):
ACK dd69b6a81a393189e4ae73ad1750eddb373b59f5

evanlinjin:
@oleonardolima I think we accidentally added back the examples in this PR!

Edit: I'll fix it!

evanlinjin (approved):
ACK 555f774169ea934dbb2934f2fbe6763908e86670

oleonardolima:
> @oleonardolima I think we accidentally added back the examples in this PR!
> 
> Edit: I'll fix it!

Oh, thanks! I must've missed that in the last rebasing.

PR #2039, opened 2025-09-18
feat(chain)!: add `CheckPointEntry`

### LINKED ISSUE #2021: Ensure `CheckPoint` chain methods validate and link via previous blockhash

LagginTimes:
Resolves #2021.

### Description

Introduce `CheckPointEntry` to represent either a real `CheckPoint` or a `prev_blockhash` placeholder.
`merge_chains()` now handles updates containing `prev_blockhash` data while preserving previous behavior when only `CheckPoint`s are present.
`is_block_in_chain()` can infer membership using `prev_blockhash`.

### Notes to the reviewers

TODO: Add tests.

### Changelog notice

* TODO

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -6,8 +6,8 @@ use core::ops::RangeBounds;
 
 use crate::collections::BTreeMap;
 use crate::{BlockId, ChainOracle, Merge};
-use bdk_core::ToBlockHash;
 pub use bdk_core::{CheckPoint, CheckPointIter};
+use bdk_core::{CheckPointEntry, ToBlockHash};
 use bitcoin::block::Header;
 use bitcoin::BlockHash;
 
@@ -69,7 +69,10 @@ impl<D> PartialEq for LocalChain<D> {
     }
 }
 
-impl<D> ChainOracle for LocalChain<D> {
+impl<D> ChainOracle for LocalChain<D>
+where
+    D: ToBlockHash + Copy,
+{
     type Error = Infallible;
 
     fn is_block_in_chain(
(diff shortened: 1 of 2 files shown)


crates/chain/src/local_chain.rs
```diff
@@ -576,6 +588,34 @@ impl core::fmt::Display for ApplyHeaderError {
 #[cfg(feature = "std")]
 impl std::error::Error for ApplyHeaderError {}
 
+/// Convert a `CheckPoint` to a `CheckPointEntry` representation that includes `prev_blockhash`.
+///
+/// NOTE: We store a block's prev_blockhash under its parent's height `(h - 1)`,  so a PrevBlockHash
+/// at height `h` and the hash at height `h` are the same.
+fn checkpoint_to_entries<D>(checkpoint: CheckPoint<D>) -> BTreeMap<u32, CheckPointEntry<D>>
```
evanlinjin: Can we have an iterator to avoid having to visit all the checkpoints?

evanlinjin (changes requested):
The original implementation of `merge_chains` was much more efficient. You don't need to change it so drastically to take into account prevhashes. Have an enum `CheckPointEntry` where both variants have a `CheckPoint` and you can call next to get to the closest lower height.

crates/chain/src/local_chain.rs
```diff
@@ -653,59 +664,132 @@ where
...
+                        // No prev available: just add update block.
+                        changeset.blocks.insert(u.height(), Some(u.data()));
+                        prev_update = curr_update.take();
                     }
+                } else if o.height() > u.height() {
+                    // Original > Update: mark original as potentially invalidated.
+                    potentially_invalidated_heights.push(o.height());
+                    prev_orig_was_invalidated = false;
+                    prev_orig = curr_orig.take();
+                    is_update_height_superset_of_original = false;
```
oleonardolima, line 776: question: isn't this already being covered by the lines L652-656 ?

evanlinjin:
Thanks for working on this.

`CheckPointEntry` being public is what I attempted at the very start (not sure where that implementation is now). I thought you would have made it private and only used for `merge_chains`.

Even looking at the unfinished state of this, I think I would much prefer #2024 (or the initial public `CheckPointEntry`).

With this implementation, we need alternative methods to figure out if a height "exists": a height with no checkpoint where the next adjacent checkpoint has a prev-blockhash.

LagginTimes:
Closing this and keeping #2024 per feedback.

PR #2027, opened 2025-09-11
feat(chain): add new `list_ordered_canonical_txs` method

oleonardolima:
### Description

It introduces a new method for topological ordering canonical transactions, `list_ordered_canonical_txs`. It now ensures the dependency-based transaction processing, guaranteeing that parent transactions always appears before their children transaction.

The existing `list_canonical_txs` and `try_list_canonical_txs` methods have been deprecated in favor of the new ordered version.

### Notes to the reviewers

"[...] For those reviewing, and wondering why we don't have a fallible try version of this method, it's because we don't have a fallible ChainOracle implementation - we will get rid of ChainOracle trait soon anyway."

This PR is intended for a point release so that bdk_wallet 2.x users can get a topologically sorted list of transactions (we need a point release on bdk_wallet 2.x as well).

It might be useful to take a look at the new test scenarios that I've added, it shows some specific scenarios where the current implementation and output of `canonical_txs` didn't output the transactions in topological order.

Let me know if you think the TopologicalIter algorithm and/or API could be improved.

### Changelog notice
```
#### Added
- New `list_ordered_canonical_txs` method to `TxGraph` that returns canonical transactions in topological order, ensuring parent transactions always appear before their children

#### Deprecated
- `list_canonical_txs` method - use `list_ordered_canonical_txs` instead for guaranteed topological ordering
- `try_list_canonical_txs` method - use `list_ordered_canonical_txs` instead for guaranteed topological ordering
```

#### All Submissions:

* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)

#### New Features:

* [x] I've added tests for the new feature
* [x] I've added docs for the new feature

PR #2041, opened 2025-09-19
feat(chain): Add extract_subgraph() and roots() methods to CanonicalView

evanlinjin:
### Description

This PR adds two new methods to `CanonicalView` to support RBF (Replace-By-Fee) transaction workflows:
- `extract_subgraph()`: Extracts a transaction and all its descendants from the canonical view
- `roots()`: Returns transactions that have no parent transactions in the view

These methods are essential for RBF support as they allow extracting specific transaction chains that can be replaced.

**Note:** This PR depends on #2029 which introduces the `CanonicalView` struct.

### Notes to the reviewers

The `extract_subgraph()` method removes the extracted transactions from the original view and returns them in a new `CanonicalView`. This is useful for isolating transaction chains that need to be replaced in RBF scenarios.

#### Added
- `CanonicalView::extract_subgraph()` method to extract transactions with all their descendants
- `CanonicalView::roots()` method to find root transactions without parents in the view

#### All Submissions:

#### New Features:

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/crates/chain/src/canonical_view.rs
+++ b/crates/chain/src/canonical_view.rs
@@ -21,7 +21,7 @@
 //! }
 //! ```
 
-use crate::collections::HashMap;
+use crate::collections::{BTreeSet, HashMap};
 use alloc::sync::Arc;
 use core::{fmt, ops::RangeBounds};
 
@@ -81,7 +81,7 @@ impl<A: Ord> PartialOrd for CanonicalTx<A> {
 #[derive(Debug)]
 pub struct CanonicalView<A> {
     /// Ordered list of transaction IDs in in topological-spending order.
-    order: Vec<Txid>,
+    order: Vec<Txid>, // TODO: Not ordered - waiting on #2038
     /// Map of transaction IDs to their transaction data and chain position.
     txs: HashMap<Txid, (Arc<Transaction>, ChainPosition<A>)>,
     /// Map of outpoints to the transaction ID that spends them.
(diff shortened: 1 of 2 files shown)


oleonardolima (commented):
cACK f08130045e3bdd126dd040f6ac3314e3fcbb4ac6

oleonardolima:
@evanlinjin This one could use a rebase, as #2029 landed on master.

crates/chain/src/canonical_view.rs
```diff
@@ -81,7 +81,7 @@ impl<A: Ord> PartialOrd for CanonicalTx<A> {
 #[derive(Debug)]
 pub struct CanonicalView<A> {
     /// Ordered list of transaction IDs in in topological-spending order.
-    order: Vec<Txid>,
+    order: Vec<Txid>, // TODO: Not ordered - waiting on #2038
```
oleonardolima, line 84: nit: we can probably update the documentation for now.

crates/chain/src/canonical_view.rs
```diff
@@ -448,4 +448,91 @@ impl<A: Anchor> CanonicalView<A> {
...
+        for txin in txs.values().flat_map(|(tx, _)| &tx.input) {
+            self.spends.remove(&txin.previous_output);
+        }
+
+        // Remove extracted transactions from self.order
+        let extracted_txids = txs.keys().copied().collect::<BTreeSet<Txid>>();
+        self.order.retain(|txid| !extracted_txids.contains(txid));
+
+        // TODO: Use this with `TopologicalIter` once #2027 gets merged to return a view that has
+        // topologically-ordered transactions.
```
oleonardolima, line 498: nit: the #2027 only got merged into `release/chain-0.23.x`, the topological order will be added as follow-up after #2038, i don't think it's worth it add the implementation from #2027 to master now, as it'll need rework.

crates/chain/src/canonical_view.rs
```diff
@@ -448,4 +448,91 @@ impl<A: Anchor> CanonicalView<A> {
...
+
+            for op in (0_u32..tx.output.len() as u32).map(|vout| OutPoint::new(txid, vout)) {
+                let spent_by = match self.spends.remove(&op) {
+                    Some(spent_by) => spent_by,
+                    None => continue,
+                };
+                spends.insert(op, spent_by);
+                to_visit.push(spent_by);
+            }
+        }
```
oleonardolima, line 486: discussion: I've been thinking about this, it does the job. However, I wonder if it'd be a good idea to have `TxDescendants`, `TxAncestors`, or at least it's methods for the `CanonicalView` too (eg. CanonicalView::walk_descendants or CanonicalView::list_descendants).

I mean that would allow to get all the ancestors for given txids, and then remove/extract in a single batch, maybe it's only bikeshedding though 😂

crates/chain/src/canonical_view.rs
```diff
@@ -448,4 +448,91 @@ impl<A: Anchor> CanonicalView<A> {
                 .collect()
         })
     }
+
+    /// Extracts a subgraph containing the specified transactions and all their descendants.
+    ///
+    /// Takes transaction IDs and returns a new `CanonicalView` containing those transactions
+    /// plus any transactions that spend their outputs (recursively). The extracted transactions
+    /// are removed from this view.
```
oleonardolima, line 456: It might be a good idea to mention that it does not errors or asserts that the txids exists in the CanonicalView, it's up to the user to make sure of that.

oleonardolima (commented):
utACK abdf5727d040d3260db4776e116e8174c01c562a

It looks good, I left a few nits that might be helpful, and a few comments that could be addressed in a follow-up. I'm going through the tests righ now.

crates/chain/src/canonical_view.rs
```diff
@@ -448,4 +448,91 @@ impl<A: Anchor> CanonicalView<A> {
...
+        let _roots = txs
+            .iter()
+            .filter(|(_, (tx, _))| {
+                tx.is_coinbase()
+                    || tx
+                        .input
+                        .iter()
+                        .all(|txin| !spends.contains_key(&txin.previous_output))
+            })
+            .map(|(&txid, _)| txid);
```
oleonardolima, line 508: fix: As the `_roots` are not being used it can be removed, also if it's to be used we should probably think of a method that reuses the `CanonicalView::roots()` with an optional param, or something like that.

crates/chain/src/canonical_view.rs
```diff
@@ -448,4 +448,91 @@ impl<A: Anchor> CanonicalView<A> {
...
+            spends,
+            tip: self.tip,
+        }
+    }
+
+    /// Returns transactions without parent transactions in this view.
+    ///
+    /// Root transactions are either coinbase transactions or transactions whose inputs
+    /// reference outputs not present in this canonical view.
+    pub fn roots(&self) -> impl Iterator<Item = CanonicalTx<A>> + '_ {
```
oleonardolima, line 522: nit: it's a good idea to also have a test for this one, at least asserting the behavior for coinbase and txs whose inputs are not in the `TxGraph`.

oleonardolima:
@evanlinjin needs a rebase/update

Issue #1634, opened 2024-10-01
`example_bitcoind_rpc_polling` sync command ends with stack overflow


ValuedMammal:
**Describe the bug**  
I ran a sync from genesis using a local signet node. It took around 25 minutes to reach the new tip but ended abruptly before finishing.
```sh
[  1442.151s] committed to db (took 0.11852204s)
[  1443.706s] synced to 0000003dee9b4362728c83d684d8490ddf4ffeb6ad7fe5ccda3f2203bffc24f2 @ 210819 | total: 0.00020300 BTC
[ 1449.7129s] synced to 000000a55e6a22bfd21c68f46774abda1e403ea95983f7d26981ccb24984ac60 @ 211577 | total: 0.00020300 BTC
[ 1455.7239s] synced to 00000071de5c82e264071bd7f0a8a91eab1e1b5461d44fce3b0fa99092e3f314 @ 212323 | total: 0.00020300 BTC
[ 1461.7284s] synced to 000000eeb493d04aeb69f4363bd01eb30540ff125a978c25ba082ad5f405071f @ 212890 | total: 0.00020300 BTC
[ 1467.7336s] synced to 0000002fdc5b95799d03f8c150fe7ead118bb9348f532edbdd5eda1bd02c1de7 @ 213657 | total: 0.00020300 BTC
[  1473.861s] synced to 000000b71b4299131fec5052912e975a10058a0c5c7532af945da9e081c76275 @ 214637 | total: 0.00019500 BTC
[   1479.87s] synced to 0000009c12399ad6344fe09264e4292f33b8db09fd26b79ad0f5bae102e9b880 @ 215572 | total: 0.00069500 BTC

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
zsh: abort      cargo run --bin example_bitcoind_rpc_polling -- sync
```

**To Reproduce**  
```sh
DESCRIPTOR=tr([7d94197e/86'/1'/0']tpubDCyQVJj8KzjiQsFjmb3KwECVXPvMwvAxxZGCP9XmWSopmjW3bCV3wD7TgxrUhiGSueDS1MU5X1Vb1YjYcp8jitXc5fXfdC1z68hDDEyKRNr/0/*)
[...]

PR #2048, opened 2025-09-25
feat(core): add skiplist to CheckPoint for faster traversal

evanlinjin:
#### Summary

Add a skiplist to `CheckPoint` using Bitcoin Core's `CBlockIndex::pskip` pattern, adapted to operate on checkpoint indices rather than block heights. Result: O(log n) lookups for `get`, `floor_at`, and `range`, with no tuning constant. Per-node memory is unchanged — `Option<Arc<CPInner>>` is niche-optimized to 8 bytes whether or not it's populated.

The pskip approach was suggested by @ValuedMammal in https://github.com/bitcoindevkit/bdk/pull/2048#issuecomment-4328710770 and prototyped in https://github.com/ValuedMammal/block-graph/blob/master/block-graph/src/checkpoint.rs. Credit for the core idea goes there. This PR takes a different implementation path; the rationale is below.

#### Design

Every `CheckPoint` node carries one Arc skip pointer to a deterministically chosen ancestor at index `skip_index(i)`. The chosen targets give skip distances that grow exponentially as you walk back, yielding O(log n) traversal — ~17 hops at 1M blocks vs ~1M for a linear walk.

`get`, `floor_at`, and `range` all reduce to "walk back to the highest checkpoint at or below `target_height`", factored into one private `walk_to_floor` helper. Each public method is a thin wrapper.

#### Bench numbers (criterion)

Absolute pskip times, with a linear-walk baseline at 1M blocks for scale:

```
get_1000_middle             227 ns
get_10000_near_end          397 ns
floor_at_10000              565 ns
random_access_skiplist_1m   1.36 µs   (vs 3.84 ms linear walk over the same chain)
```

#### Caveats

- Insert is ~50 ns slower per push than a non-skiplist `CheckPoint` because every push wires its skip pointer via an O(log n) `ancestor_by_index` walk. For real-world per-block chain extension this is negligible (well under 1 µs/block), but bulk-rebuild paths (`insert_sparse_1000` → ~3.5 µs for 1000 sequential pushes) pay it linearly.

#### History

This PR went through three design iterations. Including the rationale here so reviewers don't have to reconstruct it from the commit graph:

**1. Fixed-stride skiplist (initial proposal).** First version added a single skip pointer every `CHECKPOINT_SKIP_INTERVAL` indices, each pointing exactly that many positions back. `INTERVAL` was originally `100`, then bumped to `1000` after benchmarking — at mainnet-scale chains (~1M blocks) the cost-optimal stride for an O(n/k + k) walk is `k ≈ √n ≈ 1000`. This delivered ~50–80× speedups over linear walks but still left lookups at O(√n) (~2,000 hops at 1M blocks) and forced a tuning constant.

**2. Bitcoin Core's `pskip` (current).** @ValuedMammal pointed out that Bitcoin Core's `CBlockIndex::pskip` pattern achieves O(log n) lookups with no tuning constant by giving every node a single skip pointer to a *deterministically chosen* ancestor (computed from the index's bit pattern). Crucially, this comes at no extra per-node memory cost: `Option<Arc<CPInner>>` was already 8 bytes per node thanks to niche-optimization, so populating the field on every node uses the same space as populating it on every 1000th. Switched to this approach; the synthetic skiplist benches went from ~80% improvement to ~94% over linear, and the tuning constant disappeared.

**3. Reuse the pskip walker.** Once `ancestor_by_index` (the O(log n) walk underlying pskip) existed, it became natural to plug it into anywhere `CheckPoint`/`CheckPointIter` was advancing through the chain by a known amount. `get`, `range`, `floor_at`, `floor_below`, and `Iterator::nth` / `last` all use it now; `count` and `size_hint` derive from the `index` counter directly. Unused traversal methods on `CheckPointEntry` were removed in the same pass since they were duplicating `CheckPoint`'s own surface and had no callers in the workspace.

#### Diffs from @ValuedMammal's prototype

The [block-graph prototype](https://github.com/ValuedMammal/block-graph/blob/master/block-graph/src/checkpoint.rs) gets the high-level idea right (every node carries one deterministic skip pointer; index-based formula so sparse chains work) but has a few choices I wanted to address differently:

**1. `n & (n - 1)` underflow.** The prototype handles it by converting `u32 → i32 → u32` with `try_into().expect(...)`, relying on signed two's-complement to give `0 & -1 == 0`. This PR uses `n & n.wrapping_sub(1)` directly on `u32`, which produces the same `0 & u32::MAX == 0` result without the round-trip conversion.

**2. Traversal heuristic.** The prototype's `get` uses simple greedy ("take skip if it doesn't undershoot, else prev"). That's correct but **not genuinely O(log n)** — for targets near genesis on a long chain, it degrades to many small geometric descents stacked end-to-end. This PR uses Bitcoin Core's `GetAncestor` heuristic verbatim: take skip unless the *predecessor's* skip would have been a strictly bigger jump that still doesn't undershoot.

**3. Single internal helper for `get`/`range`/`floor_at`.** The public surface (`get`, `range`, `floor_at`) all reduces to "walk back to the highest checkpoint at or below `target_height`". This PR factors that into one private `walk_to_floor` function; each public method becomes a thin wrapper. The prototype duplicates the skip-walk logic.

**4. `Drop` cleanup.** With every node holding a skip Arc, the manual unwind loop in `Drop for CPInner` (existing fix for #1634) is extended with `node.skip.take()` so skip-pointer drops happen on the manual loop rather than triggering ancestor drops via the implicit recursive path.

### Changelog

```md
Added:
- `O(log n)` skiplist on `CheckPoint` (Bitcoin Core's pskip pattern). Speeds up `get`, `floor_at`, `range`, and full-chain iteration.
- `Iterator::nth` / `last` / `count` / `size_hint` overrides on `CheckPointIter` and `CheckPointEntryIter`, plus `ExactSizeIterator` impls. `nth` and `last` are now `O(log n)`; `count` and `size_hint` are `O(1)`.

Removed:
- Unused traversal methods on `CheckPointEntry`: `iter`, `get`, `range`, `floor_at`, `floor_below`. They had no callers in the workspace and duplicated `CheckPoint`'s own surface.
```

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -23,3 +23,8 @@ serde = ["dep:serde", "bitcoin/serde", "hashbrown?/serde"]
 [dev-dependencies]
 bdk_chain = { path = "../chain" }
 bdk_testenv = { path = "../testenv", default-features = false }
+criterion = { version = "0.2" }
+
+[[bench]]
+name = "checkpoint_skiplist"
+harness = false
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -23,3 +23,9 @@ serde = ["dep:serde", "bitcoin/serde", "hashbrown?/serde"]
 [dev-dependencies]
 bdk_chain = { path = "../chain" }
 bdk_testenv = { path = "../testenv", default-features = false }
+criterion = { version = "0.7" }
+proptest = "1.2.0"
+
+[[bench]]
+name = "checkpoint_skiplist"
+harness = false
--- a/crates/core/benches/checkpoint_skiplist.rs
+++ b/crates/core/benches/checkpoint_skiplist.rs
@@ -1,0 +1,235 @@
+use bdk_core::CheckPoint;
+use bitcoin::hashes::Hash;
+use bitcoin::BlockHash;
+use criterion::{black_box, criterion_group, criterion_main, Bencher, Criterion};
+
+/// Create a checkpoint chain with the given length
+fn create_checkpoint_chain(length: u32) -> CheckPoint<BlockHash> {
+    let mut cp = CheckPoint::new(0, BlockHash::all_zeros());
+    for height in 1..=length {
+        let hash = BlockHash::from_byte_array([(height % 256) as u8; 32]);
+        cp = cp.push(height, hash).unwrap();
+    }
+    cp
+}
+
+/// Benchmark get() operations at various depths
+fn bench_checkpoint_get(c: &mut Criterion) {
+    // Small chain - get near start
+    c.bench_function("get_100_near_start", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(100);
+        let target = 10;
+        b.iter(|| {
+            black_box(cp.get(target));
+        });
+    });
+
+    // Medium chain - get middle
+    c.bench_function("get_1000_middle", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(1000);
+        let target = 500;
+        b.iter(|| {
+            black_box(cp.get(target));
+        });
+    });
+
+    // Large chain - get near end
+    c.bench_function("get_10000_near_end", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        let target = 9000;
+        b.iter(|| {
+            black_box(cp.get(target));
+        });
+    });
+
+    // Large chain - get near start (best case for skiplist)
+    c.bench_function("get_10000_near_start", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        let target = 100;
+        b.iter(|| {
+            black_box(cp.get(target));
+        });
+    });
+}
+
+/// Benchmark floor_at() operations
+fn bench_checkpoint_floor_at(c: &mut Criterion) {
+    c.bench_function("floor_at_1000", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(1000);
+        let target = 750; // Target that might not exist exactly
+        b.iter(|| {
+            black_box(cp.floor_at(target));
+        });
+    });
+
+    c.bench_function("floor_at_10000", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        let target = 7500;
+        b.iter(|| {
+            black_box(cp.floor_at(target));
+        });
+    });
+}
+
+/// Benchmark range() iteration
+fn bench_checkpoint_range(c: &mut Criterion) {
+    // Small range in middle (tests skip pointer efficiency)
+    c.bench_function("range_1000_middle_10pct", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(1000);
+        b.iter(|| {
+            let range: Vec<_> = cp.range(450..=550).collect();
+            black_box(range);
+        });
+    });
+
+    // Large range (tests iteration performance)
+    c.bench_function("range_10000_large_50pct", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        b.iter(|| {
+            let range: Vec<_> = cp.range(2500..=7500).collect();
+            black_box(range);
+        });
+    });
+
+    // Range from start (tests early termination)
+    c.bench_function("range_10000_from_start", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        b.iter(|| {
+            let range: Vec<_> = cp.range(..=100).collect();
+            black_box(range);
+        });
+    });
+
+    // Range near tip (minimal skip pointer usage)
+    c.bench_function("range_10000_near_tip", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        b.iter(|| {
+            let range: Vec<_> = cp.range(9900..).collect();
+            black_box(range);
+        });
+    });
+
+    // Single element range (edge case)
+    c.bench_function("range_single_element", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        b.iter(|| {
+            let range: Vec<_> = cp.range(5000..=5000).collect();
+            black_box(range);
+        });
+    });
+}
+
+/// Benchmark insert() operations
+fn bench_checkpoint_insert(c: &mut Criterion) {
+    c.bench_function("insert_sparse_1000", |b: &mut Bencher| {
+        // Create a sparse chain
+        let mut cp = CheckPoint::new(0, BlockHash::all_zeros());
+        for i in 1..=100 {
+            let height = i * 10;
+            let hash = BlockHash::from_byte_array([(height % 256) as u8; 32]);
+            cp = cp.push(height, hash).unwrap();
+        }
+
+        let insert_height = 505;
+        let insert_hash = BlockHash::from_byte_array([255; 32]);
+
+        b.iter(|| {
+            let result = cp.clone().insert(insert_height, insert_hash);
+            black_box(result);
+        });
+    });
+}
+
+/// Compare linear traversal vs skiplist-enhanced get()
+fn bench_traversal_comparison(c: &mut Criterion) {
+    // Linear traversal benchmark
+    c.bench_function("linear_traversal_10000", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        let target = 100; // Near the beginning
+
+        b.iter(|| {
+            let mut current = cp.clone();
+            while current.height() > target {
+                if let Some(prev) = current.prev() {
+                    current = prev;
+                } else {
+                    break;
+                }
+            }
+            black_box(current);
+        });
+    });
+
+    // Skiplist-enhanced get() for comparison
+    c.bench_function("skiplist_get_10000", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        let target = 100; // Same target
+
+        b.iter(|| {
+            black_box(cp.get(target));
+        });
+    });
+}
+
+/// Analyze skip pointer distribution and usage
+fn bench_skip_pointer_analysis(c: &mut Criterion) {
+    c.bench_function("count_skip_pointers_10000", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+
+        b.iter(|| {
+            let mut count = 0;
+            let mut current = cp.clone();
+            loop {
+                if current.skip().is_some() {
+                    count += 1;
+                }
+                if let Some(prev) = current.prev() {
+                    current = prev;
+                } else {
+                    break;
+                }
+            }
+            black_box(count);
+        });
+    });
+
+    // Measure actual skip pointer usage during traversal
+    c.bench_function("skip_usage_in_traversal", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        let target = 100;
+
+        b.iter(|| {
+            let mut current = cp.clone();
+            let mut skips_used = 0;
+
+            while current.height() > target {
+                if let Some(skip_cp) = current.skip() {
+                    if skip_cp.height() >= target {
+                        current = skip_cp;
+                        skips_used += 1;
+                        continue;
+                    }
+                }
+
+                if let Some(prev) = current.prev() {
+                    current = prev;
+                } else {
+                    break;
+                }
+            }
+            black_box((current, skips_used));
+        });
+    });
+}
+
+criterion_group!(
+    benches,
+    bench_checkpoint_get,
+    bench_checkpoint_floor_at,
+    bench_checkpoint_range,
+    bench_checkpoint_insert,
+    bench_traversal_comparison,
+    bench_skip_pointer_analysis
+);
+
+criterion_main!(benches);
--- a/crates/core/benches/checkpoint_skiplist.rs
+++ b/crates/core/benches/checkpoint_skiplist.rs
@@ -16,15 +16,6 @@ fn create_checkpoint_chain(length: u32) -> CheckPoint<BlockHash> {
 
 /// Benchmark get() operations at various depths
 fn bench_checkpoint_get(c: &mut Criterion) {
-    // Small chain - get near start
-    c.bench_function("get_100_near_start", |b: &mut Bencher| {
-        let cp = create_checkpoint_chain(100);
-        let target = 10;
-        b.iter(|| {
-            black_box(cp.get(target));
-        });
-    });
-
     // Medium chain - get middle
     c.bench_function("get_1000_middle", |b: &mut Bencher| {
         let cp = create_checkpoint_chain(1000);
@@ -34,10 +25,12 @@ fn bench_checkpoint_get(c: &mut Criterion) {
         });
     });
 
-    // Large chain - get near end
+    // Large chain - get near end. Target is deliberately not a power of two or a multiple of
+    // common skiplist intervals (e.g. 1000), so neither pskip nor a fixed-stride scheme gets a
+    // free 1-hop hit on a "lucky" alignment.
     c.bench_function("get_10000_near_end", |b: &mut Bencher| {
         let cp = create_checkpoint_chain(10000);
-        let target = 9000;
+        let target = 9123;
         b.iter(|| {
             black_box(cp.get(target));
         });
@@ -100,24 +93,6 @@ fn bench_checkpoint_range(c: &mut Criterion) {
             black_box(range);
         });
     });
-
-    // Range near tip (minimal skip pointer usage)
-    c.bench_function("range_10000_near_tip", |b: &mut Bencher| {
-        let cp = create_checkpoint_chain(10000);
-        b.iter(|| {
-            let range: Vec<_> = cp.range(9900..).collect();
-            black_box(range);
-        });
-    });
-
-    // Single element range (edge case)
-    c.bench_function("range_single_element", |b: &mut Bencher| {
-        let cp = create_checkpoint_chain(10000);
-        b.iter(|| {
-            let range: Vec<_> = cp.range(5000..=5000).collect();
-            black_box(range);
-        });
-    });
 }
 
 /// Benchmark insert() operations
@@ -141,84 +116,56 @@ fn bench_checkpoint_insert(c: &mut Criterion) {
     });
 }
 
-/// Compare linear traversal vs skiplist-enhanced get()
-fn bench_traversal_comparison(c: &mut Criterion) {
-    // Linear traversal benchmark
-    c.bench_function("linear_traversal_10000", |b: &mut Bencher| {
-        let cp = create_checkpoint_chain(10000);
-        let target = 100; // Near the beginning
+/// Random-access lookups over a realistic-size chain, comparing skiplist-enhanced
+/// `get()` against a plain linear walk. Targets are drawn from a deterministic
+/// xorshift sequence so the same query stream is used for both benches.
+///
+/// Chain length is sized to the order of a full Bitcoin chain (~1M blocks) so the log-scale
+/// advantage of pskip is visible.
+fn bench_random_access(c: &mut Criterion) {
+    const CHAIN_LEN: u32 = 1_000_000;
+    const QUERIES: usize = 256;
 
+    let cp = create_checkpoint_chain(CHAIN_LEN);
+
+    // Deterministic xorshift64* over the height range.
+    let mut state: u64 = 0x9E37_79B9_7F4A_7C15;
+    let targets: Vec<u32> = (0..QUERIES)
+        .map(|_| {
+            state ^= state << 13;
+            state ^= state >> 7;
+            state ^= state << 17;
+            (state % (CHAIN_LEN as u64 + 1)) as u32
+        })
+        .collect();
+
+    {
+        let cp = cp.clone();
+        let targets = targets.clone();
+        c.bench_function("random_access_skiplist_1m", move |b: &mut Bencher| {
+            let mut i = 0usize;
+            b.iter(|| {
+                let target = targets[i % QUERIES];
+                i = i.wrapping_add(1);
+                black_box(cp.get(target));
+            });
+        });
+    }
+
+    c.bench_function("random_access_linear_1m", move |b: &mut Bencher| {
+        let mut i = 0usize;
         b.iter(|| {
+            let target = targets[i % QUERIES];
+            i = i.wrapping_add(1);
+
             let mut current = cp.clone();
             while current.height() > target {
-                if let Some(prev) = current.prev() {
-                    current = prev;
-                } else {
-                    break;
+                match current.prev() {
+                    Some(prev) => current = prev,
+                    None => break,
                 }
             }
             black_box(current);
-        });
-    });
-
-    // Skiplist-enhanced get() for comparison
-    c.bench_function("skiplist_get_10000", |b: &mut Bencher| {
-        let cp = create_checkpoint_chain(10000);
-        let target = 100; // Same target
-
-        b.iter(|| {
-            black_box(cp.get(target));
-        });
-    });
-}
-
-/// Analyze skip pointer distribution and usage
-fn bench_skip_pointer_analysis(c: &mut Criterion) {
-    c.bench_function("count_skip_pointers_10000", |b: &mut Bencher| {
-        let cp = create_checkpoint_chain(10000);
-
-        b.iter(|| {
-            let mut count = 0;
-            let mut current = cp.clone();
-            loop {
-                if current.skip().is_some() {
-                    count += 1;
-                }
-                if let Some(prev) = current.prev() {
-                    current = prev;
-                } else {
-                    break;
-                }
-            }
-            black_box(count);
-        });
-    });
-
-    // Measure actual skip pointer usage during traversal
-    c.bench_function("skip_usage_in_traversal", |b: &mut Bencher| {
-        let cp = create_checkpoint_chain(10000);
-        let target = 100;
-
-        b.iter(|| {
-            let mut current = cp.clone();
-            let mut skips_used = 0;
-
-            while current.height() > target {
-                if let Some(skip_cp) = current.skip() {
-                    if skip_cp.height() >= target {
-                        current = skip_cp;
-                        skips_used += 1;
-                        continue;
-                    }
-                }
-
-                if let Some(prev) = current.prev() {
-                    current = prev;
-                } else {
-                    break;
-                }
-            }
-            black_box((current, skips_used));
         });
     });
 }
@@ -229,8 +176,7 @@ criterion_group!(
     bench_checkpoint_floor_at,
     bench_checkpoint_range,
     bench_checkpoint_insert,
-    bench_traversal_comparison,
-    bench_skip_pointer_analysis
+    bench_random_access
 );
 
 criterion_main!(benches);
--- a/crates/core/benches/checkpoint_skiplist.rs
+++ b/crates/core/benches/checkpoint_skiplist.rs
@@ -1,0 +1,236 @@
+use bdk_core::CheckPoint;
+use bitcoin::hashes::Hash;
+use bitcoin::BlockHash;
+use core::hint::black_box;
+use criterion::{criterion_group, criterion_main, Bencher, Criterion};
+
+/// Create a checkpoint chain with the given length
+fn create_checkpoint_chain(length: u32) -> CheckPoint<BlockHash> {
+    let mut cp = CheckPoint::new(0, BlockHash::all_zeros());
+    for height in 1..=length {
+        let hash = BlockHash::from_byte_array([(height % 256) as u8; 32]);
+        cp = cp.push(height, hash).unwrap();
+    }
+    cp
+}
+
+/// Benchmark get() operations at various depths
+fn bench_checkpoint_get(c: &mut Criterion) {
+    // Small chain - get near start
+    c.bench_function("get_100_near_start", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(100);
+        let target = 10;
+        b.iter(|| {
+            black_box(cp.get(target));
+        });
+    });
+
+    // Medium chain - get middle
+    c.bench_function("get_1000_middle", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(1000);
+        let target = 500;
+        b.iter(|| {
+            black_box(cp.get(target));
+        });
+    });
+
+    // Large chain - get near end
+    c.bench_function("get_10000_near_end", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        let target = 9000;
+        b.iter(|| {
+            black_box(cp.get(target));
+        });
+    });
+
+    // Large chain - get near start (best case for skiplist)
+    c.bench_function("get_10000_near_start", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        let target = 100;
+        b.iter(|| {
+            black_box(cp.get(target));
+        });
+    });
+}
+
+/// Benchmark floor_at() operations
+fn bench_checkpoint_floor_at(c: &mut Criterion) {
+    c.bench_function("floor_at_1000", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(1000);
+        let target = 750; // Target that might not exist exactly
+        b.iter(|| {
+            black_box(cp.floor_at(target));
+        });
+    });
+
+    c.bench_function("floor_at_10000", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        let target = 7500;
+        b.iter(|| {
+            black_box(cp.floor_at(target));
+        });
+    });
+}
+
+/// Benchmark range() iteration
+fn bench_checkpoint_range(c: &mut Criterion) {
+    // Small range in middle (tests skip pointer efficiency)
+    c.bench_function("range_1000_middle_10pct", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(1000);
+        b.iter(|| {
+            let range: Vec<_> = cp.range(450..=550).collect();
+            black_box(range);
+        });
+    });
+
+    // Large range (tests iteration performance)
+    c.bench_function("range_10000_large_50pct", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        b.iter(|| {
+            let range: Vec<_> = cp.range(2500..=7500).collect();
+            black_box(range);
+        });
+    });
+
+    // Range from start (tests early termination)
+    c.bench_function("range_10000_from_start", |b: &mut Bencher| {
+        let cp = create_checkpoint_chain(10000);
+        b.iter(|| {
+            let range: Vec<_> = cp.range(..=100).collect();
+            black_box(range);
+        });
+    });
+
+    // Range near tip (minimal skip pointer usage)
+    c.bench_function("range_10000_near_tip", |b: &mut Bencher| {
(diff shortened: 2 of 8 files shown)


evanlinjin:
Guys, this is purely done by Claude. I haven't reviewed it yet.

evanlinjin:
## Performance Benchmark Comparison

Benchmarks comparing the **old O(n) implementation** vs **new skiplist O(√n) implementation** for a 10,000 checkpoint chain:

### 🎯 Key Results

| Operation | Old Implementation | Skiplist Implementation | Speedup |
|-----------|-------------------|------------------------|---------|
| **get(100)** - near start | 98.270 μs | 421 ns | **233x faster** |
| **get(9000)** - near end | 9.668 μs | 44 ns | **220x faster** |
| **linear_traversal(100)** | 56.965 μs | 110.66 μs | 0.5x (expected*) |

### 📊 Detailed Benchmarks

#### Finding checkpoint at position 100 (from 10k chain):
- **Old**: `98.270 μs` - Linear search from tip
- **New**: `421 ns` - Skip pointers jump directly to target
- **Improvement**: **233x faster** 🚀

#### Finding checkpoint at position 9000 (from 10k chain):  
- **Old**: `9.668 μs` - Linear search through 1000 nodes
- **New**: `44 ns` - Skip pointers minimize traversal
- **Improvement**: **220x faster** 🚀

\* *Note: The linear_traversal benchmark shows the new implementation is slightly slower because it's doing the same linear traversal but with additional overhead from the skip/index fields. The real performance gains come from using the skiplist-aware methods like `get()`, `floor_at()`, and `range()`.*

### Summary
The skiplist implementation provides **massive performance improvements** for checkpoint lookups, especially for deep searches in long chains. The O(√n) complexity is clearly demonstrated with 200x+ speedups in real-world scenarios.

evanlinjin:
> Guys, this is purely done by Claude. I haven't reviewed it yet.

It's now fully reviewed by myself! Made many simplifications.

Let's merge #2055 and rebase this on top of that!

evanlinjin:
## Skiplist Performance Update

After the optimizations, here are the updated benchmark results:

### `get()` Performance
| Benchmark | Time | Notes |
|-----------|------|-------|
| `get_100_near_start` | 475.89 ns | Get checkpoint near start of 100-item chain |
| `get_1000_middle` | 31.07 ns | Get checkpoint in middle of 1000-item chain |
| `get_10000_near_end` | 57.12 ns | Get checkpoint near end of 10000-item chain |
| `get_10000_near_start` | 535.37 ns | Get checkpoint near start of 10000-item chain |

### `floor_at()` Performance
| Benchmark | Time | Notes |
|-----------|------|-------|
| `floor_at_1000` | 286.33 ns | Floor at height 750 in 1000-item chain |
| `floor_at_10000` | 673.27 ns | Floor at height 7500 in 10000-item chain |

### `range()` Performance
| Benchmark | Time | Notes |
|-----------|------|-------|
| `range_1000_middle_10pct` | 1.67 µs | Range 450..=550 in 1000-item chain |
| `range_10000_large_50pct` | 97.59 µs | Range 2500..=7500 in 10000-item chain |
| `range_10000_from_start` | 3.11 µs | Range ..=100 in 10000-item chain |
| `range_10000_near_tip` | 1.21 µs | Range 9900.. in 10000-item chain |
| `range_single_element` | 942.21 ns | Range 5000..=5000 in 10000-item chain |

### Traversal Comparison
| Benchmark | Time | Notes |
|-----------|------|-------|
| `linear_traversal_10000` | 140.90 µs | Linear search to height 100 in 10000-item chain |
| `skiplist_get_10000` | 539.80 ns | Skip-enhanced search to height 100 in 10000-item chain |

**Speedup: 261x faster with skip pointers!**

### Summary

The skip list implementation successfully achieves O(√n) time complexity for search operations. Key improvements from our optimizations:

1. **Cleaner two-phase traversal** in `get()` and `range()`
2. **Simplified `floor_at()`** from 33 lines to 1 line
3. **Restored elegant `insert()`** implementation (removed 60+ lines)  
4. **Refactored `push()`** with clearer skip pointer logic

All tests pass and the implementation is now both performant and maintainable.

ValuedMammal:
I think it should have more than a single level to achieve the optimal performance, but I'm not sure if that's possible without implementing a new type.

nymius (commented):
This will be an improvement over linked list for sure, but I would like to combine efforts with [block-graph/skiplist](https://github.com/ValuedMammal/block-graph/tree/master/skiplist) to avoid duplicated work.
I find some components of the skip list too tied up to the underlying data, making difficult to resonate about the functionality of the skip list by itself. I would like to detach the skiplist logic from the underlying data.
There is a lot of room for improvement for skiplist applied to our particular use case, so I'm positive there will be a lot of changes to a structure like it. I would try to make smaller PRs to check the improvements, but get results early.
In that sense, the benchmark is going to be very handy to ensure we are making progress.
A one level, fixed skip interval is a great starting point for that.

evanlinjin:
> I think it should have more than a single level to achieve the optimal performance, but I'm not sure if that's possible without implementing a new type.

@ValuedMammal  I think the current state of the PR is a good balance between performance and simplicity (for now).

* We are already achieving queries in the single-digit microsecond range for typical wallet sizes (1k-100k checkpoints). In other words, we can already easily do hundreds of thousands of block-queries in under a second.

* Doing a multi-level skiplist will increase code complexity significantly and increase memory usage by ~20% (due to Vec overhead for multiple skip pointers - admittedly, that's not a lot).

Here is a performance summary by Claude:

  | Chain Size    | Checkpoints | Current | Multi-Level | Speedup |
  |---------------|-------------|---------|-------------|---------|
  | Small wallet  | 1,000       | 23.5 ns | ~4 ns       | 6x      |
  | Medium wallet | 10,000      | 414 ns  | ~60 ns      | 7x      |
  | Large wallet  | 100,000     | 2.3 μs  | ~70 ns      | 33x     |
  | Full chain    | 870,000     | 18.2 μs | ~83 ns      | 219x    |

* The performance gain for wallets with 10,000 or less checkpoints is insignificant (probably the majority of wallets right now).

* The performance gain for wallets with 100,000 checkpoints is significant - but probably not noticeable for the user.

* **Worst-case scenario:** Let's say we have a large wallet with 10,000 transactions and we canonicalize against a full checkpoint chain: That's 182 milliseconds to query for all relevant blocks which is perfectly manageable for a local wallet.

evanlinjin:
@nymius 

> I find some components of the skip list too tied up to the underlying data, making difficult to resonate about the functionality of the skip list by itself. I would like to detach the skiplist logic from the underlying data.

The current state of the PR is self-contained and reviewable - the changes are all internal changes to `CheckPoint` + tests + benchmarks. We modified two read methods (`get`, `range`) and a single write method (`push`). We added two fields in `CheckPoint`. The review surface is extremely minimal. A total of 126 additional lines in `checkpoint.rs` (including comments).

Are you wanting to detach because you see this skiplist logic being used elsewhere? The current skiplist implementation has domain-specific simplifications which won't exist in a full skiplist implementation (i.e. append-only in our `push`, immutable - no need to modify node data). Furthermore, a more generic skiplist implementation will be 1000+ lines of code.

> I would try to make smaller PRs to check the improvements, but get results early.

I don't think splitting this PR into multiple PRs makes sense here. This is a single atomic change to `CheckPoint`. The majority of this PR is tests and benchmarks. The modification surface is tiny.

---

We may need a new internal type to make multi-level skiplist possible, and thus achieve *optimal performance*. I'm not opposed to this idea. However, I think this PR is a simple change that achieves *good enough performance*.

nymius:
You've exposed good points and I agree with all of them. Will review again ~~soon~~.

martinsaposnic:
I was about to open a PR for this. I'm hitting this problem running an LDK Node on mutinynet with bitcoind as chain source where there's a 2-second poll that calls `Wallet::transactions()`, and once the checkpoint chain is long enough the operation takes longer than 2 seconds, so the CPU never idles.

I took a different approach adding a `BTreeMap<u32, CheckPoint>` index on `LocalChain` for O(log n) lookups. Happy to share it, although I believe this change would suffice too

nymius:
@martinsaposnic would you be open to benchmark your case against this branch?

crates/core/tests/test_checkpoint_skiplist.rs
```diff
@@ -0,0 +1,245 @@
...
+
+    if let Some(skip) = current.skip() {
+        assert_eq!(skip.index(), 0);
+    } else {
+        panic!("Expected skip pointer at index 100");
+    }
+}
+
+#[test]
+fn test_skiplist_get_performance() {
```
nymius, line 38: Why are we testing performance here? Shouldn't be a task for bench tests?

crates/core/benches/checkpoint_skiplist.rs
```diff
@@ -0,0 +1,235 @@
+use bdk_core::CheckPoint;
+use bitcoin::hashes::Hash;
+use bitcoin::BlockHash;
+use criterion::{black_box, criterion_group, criterion_main, Bencher, Criterion};
+
```
nymius, line 5: A test checking random accesses to the skip list would be useful here. Also a comparison against linear traversal.

nymius (commented):
Needs rebase.

evanlinjin:
Pushed a follow-up commit bumping `CHECKPOINT_SKIP_INTERVAL` from `100` to `1000`.

### Rationale

Traversal cost with a fixed skiplist interval `k` is `O(n/k + k)`, minimized at `k ≈ √n`. The original `k=100` was tuned for `n ≈ 10,000`, which isn't the workload that motivates this PR. The reported issue ([comment](https://github.com/bitcoindevkit/bdk/pull/2048#issuecomment-4229421931)) is a dense server chain near current tip height (~946k blocks), where `√n ≈ 970`. Rounding to `1000` aligns the constant with that regime.

### Concretely

| n | k=100 hops | k=1000 hops |
|---|---|---|
| 946,000 | ~9,560 | ~1,950 (**~5× faster**) |
| 464,000 (segwit-era) | ~4,740 | ~1,460 |
| 10,000 | ~200 | ~910 (slower, but still sub-µs) |

### Memory

No meaningful difference — `Option<Arc<CPInner>>` is niche-optimized to 8 bytes regardless of `Some`/`None`, so every node carries the same `skip` field. Skip pointers also reference existing nodes in the chain, so populated pointers don't trigger new heap allocations (just refcount increments on already-allocated `ArcInner`s). `k` doesn't move the memory needle here.

### Tradeoff

Chains shorter than ~1000 nodes no longer create any skip pointers. That's fine — linear traversal at `n ≤ 1000` is already microseconds and wasn't the motivating problem. Small wallets weren't the target.

Left `k` as a fixed single-level constant rather than going multi-level / adaptive — for BDK's realistic size ceiling (~1M), the constant-factor win is sufficient and not worth the extra machinery. Happy to revisit if that changes.

_(Edited: corrected an inaccurate claim about memory overhead — `Option<Arc<_>>` is niche-optimized so `k` doesn't meaningfully change memory usage.)_

crates/core/tests/test_checkpoint_skiplist.rs
evanlinjin: Thanks for pointing this out. Removed the test completely - I don't think it added any value.

crates/core/benches/checkpoint_skiplist.rs
evanlinjin: Good call — addressed in d7410c63.

Added `bench_random_access` which draws 256 targets from a deterministic xorshift sequence and runs both a skiplist-enhanced `get()` and a plain linear walk over the same query stream on a 100k-node chain (`random_access_skiplist_100k` vs `random_access_linear_100k`). 100k is large enough to make the skiplist delta clearly visible without slowing harness setup.

Note the existing `bench_traversal_comparison` also compares linear vs skiplist, though at 10k with a single fixed target — keeping it for continuity, but the new random-access bench is the more representative measurement.

---

The above is clearly written by Claude.

evanlinjin:
@nymius For the random access benchmark introduced in d7410c631dbf047212300817ab006478bf45912d;

`random_access_linear_100k` got  1.07 ms
`random_access_skiplist_100k` got 11.86 µs

~90x speedup

ValuedMammal:
I recommend looking at how it's done in block-graph and read up on the [blog post](https://github.com/ValuedMammal/valuedmammal.github.io/#skipping-to-the-good-part-in-logarithmic-time) for a conceptual overview.

evanlinjin:
> I recommend looking at how it's done in block-graph and read up on the [blog post](https://github.com/ValuedMammal/valuedmammal.github.io/#skipping-to-the-good-part-in-logarithmic-time) for a conceptual overview.

@ValuedMammal Thank you for the comment. The `pskip` solution seems to be the perfect solution both in terms of simplicity and performance.

I've noticed that you have already completed this work ~2 months ago by copying `CheckPoint` into a separate repository (https://github.com/ValuedMammal/block-graph/tree/master) and implementing `pskip` there 😅.

To move this forward, would you instead like to make a PR directly into this repository? Or would you like me to be responsible for moving that over?

ValuedMammal:
I'm glad you see the potential in it. I hadn't planned to open a PR as a I see this one is still in progress. There's probably a way to integrate both ideas.

evanlinjin:
@ValuedMammal I've updated the PR to use pskip. Please check the PR description for details.

crates/core/Cargo.toml
```diff
@@ -23,3 +23,8 @@ serde = ["dep:serde", "bitcoin/serde", "hashbrown?/serde"]
 [dev-dependencies]
 bdk_chain = { path = "../chain" }
 bdk_testenv = { path = "../testenv", default-features = false }
+criterion = { version = "0.2" }
```
oleonardolima: question: any reason to still use the 0.2.x ? as we're now in 1.85 msrv we could use a more recent one; also i noticed we should bump the other benchmarks.

crates/core/src/checkpoint.rs
```diff
@@ -481,12 +590,15 @@ mod tests {
...
+        // genesis checkpoint:
+        // 1. The variable `genesis`
+        // 2. The chain `cp` through checkpoint 1's prev pointer
+        // 3. Checkpoint at index 100's skip pointer (points to index 0)
         assert_eq!(
             Arc::strong_count(&genesis.0),
-            2,
-            "`cp` and `genesis` should be the only strong references",
+            3,
+            "`cp`, `genesis`, and checkpoint 100's skip pointer should be the only strong references",
```
oleonardolima: question: I'm not sure if I didn't get it right, but shouldn't it be 1000 idx ?

ValuedMammal:
In 87f2bd12cf5a70582494fcaaf53f369be1cc26a8 feat(core): add skiplist to CheckPoint for faster traversal

You have criterion 0.2 in core/Cargo.toml, but have used criterion 0.7 elsewhere.

crates/core/src/checkpoint.rs
```diff
@@ -6,6 +6,87 @@ use core::ops::RangeBounds;
...
+    while curr.height() > target_height {
+        let skip = curr.skip();
+        let take_skip = match &skip {
+            Some(skip_cp) if skip_cp.height() < target_height => false,
+            Some(skip_cp) if skip_cp.height() == target_height => true,
+            // Skip lands above target. Prefer prev's skip if it's a strictly bigger jump (lands
+            // more than 2 heights lower than current's skip) that still reaches target.
+            Some(skip_cp) => match curr.prev().and_then(|p| p.skip()) {
+                Some(prev_skip_cp) => {
+                    let prev_sh = prev_skip_cp.height();
```
ValuedMammal:
```suggestion
                    let prev_skip_height = prev_skip_cp.height();
```

crates/core/tests/test_checkpoint_skiplist.rs
```diff
@@ -0,0 +1,394 @@
...
+    const N: u32 = 100_000;
+    const HOP_LIMIT: u32 = 256;
+
+    let mut cp = CheckPoint::new(0, BlockHash::all_zeros());
+    for height in 1..=N {
+        let hash = BlockHash::from_byte_array([(height % 256) as u8; 32]);
+        cp = cp.push(height, hash).unwrap();
+    }
+
+    // Mirrors `walk_step_toward` (private to bdk_core) — the same skip-vs-prev heuristic the
```
ValuedMammal: There's no `walk_step_toward` in bdk_core, probably a stale comment.

crates/core/tests/test_checkpoint_skiplist.rs
```diff
@@ -0,0 +1,394 @@
...
+    // public `get()` uses internally. Counting its hops here verifies the heuristic actually
+    // traverses logarithmically through the public skip()/prev() accessors.
+    let count_hops = |target: u32| -> u32 {
+        let mut current = cp.clone();
+        let mut hops = 0u32;
+        while current.height() > target {
+            let skip = current.skip();
+            let take_skip = match &skip {
+                Some(s) => {
```
ValuedMammal: `s` -> `skip_cp`
`sh` -> `skip_cp_height`
`ps` -> `prev_skip_cp`
`psh` -> `prev_skip_height`

crates/core/src/checkpoint.rs
```diff
@@ -166,17 +225,39 @@ impl<D> CheckPoint<D> {
...
-            })
-            .take_while(move |cp| match start_bound {
-                core::ops::Bound::Included(inc_bound) => cp.height() >= inc_bound,
-                core::ops::Bound::Excluded(exc_bound) => cp.height() > exc_bound,
-                core::ops::Bound::Unbounded => true,
-            })
+
+        let is_above_bound = |height: u32| match end_bound {
+            core::ops::Bound::Included(inc_bound) => height > inc_bound,
+            core::ops::Bound::Excluded(exc_bound) => height >= exc_bound,
```
oleonardolima: question: are these bound correct ? (it seems to already be wrong before)

ValuedMammal:
In e2ed5ad02d348f3362a618bc50e1efe6af724e4a feat(core): Optimize `CheckPointEntryIter` using pskip

Probably a typo: the commit says optimize `CheckPointEntryIter`, but it actually optimizes a different type `CheckPointIter`.

crates/core/src/checkpoint.rs
```diff
@@ -168,7 +236,15 @@ impl<D> CheckPoint<D> {
...
-    /// Get the skip pointer checkpoint if it exists.
+    /// Get this checkpoint's pskip ancestor, if one exists.
+    ///
+    /// Returns the ancestor at the [skip index](Self::index) — a deterministically chosen node that
+    /// lets traversals cover exponentially-growing distances per hop. Returns `None` for the
+    /// genesis node (index `0`); for index `1`, the skip ancestor is the same node as `prev`.
+    ///
+    /// This accessor exposes the internal pskip topology and is intended for diagnostics and
+    /// benchmarks, not for general-purpose traversal — use [`get`](Self::get),
+    /// [`range`](Self::range), or [`floor_at`](Self::floor_at) instead.
```
oleonardolima, line 254: nit: s/node/checkpoint/g

crates/core/src/checkpoint.rs
```diff
@@ -493,27 +531,10 @@ where
...
-                // predecessors.
-                current = current.prev.clone().expect("chain has enough checkpoints");
-            }
-            Some(current)
-        } else {
-            None
-        };
+        // Wire the pskip pointer to the ancestor at skip_index(new_index). This is what makes
+        // traversal O(log n): each node carries one skip Arc to a deterministically chosen
+        // ancestor, and the chosen distances grow exponentially as you walk back.
```
oleonardolima, line 543: nit: s/node/checkpoint/g

crates/core/src/checkpoint.rs
```diff
@@ -562,8 +562,46 @@ impl<D> Iterator for CheckPointIter<D> {
...
+            .map_or(0, |cp_inner| (cp_inner.index as usize).saturating_add(1))
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let n = self
+            .next
+            .as_ref()
+            .map_or(0, |cp_inner| (cp_inner.index as usize).saturating_add(1));
+        (n, Some(n))
+    }
```
oleonardolima, line 607: @evanlinjin are these strictly intended for testing purposes ? if yes, i'd suggest putting it under a feature.

oleonardolima (commented):
cACK e2ed5ad02d348f3362a618bc50e1efe6af724e4a

I did a first pass, overall it looks good! I'll just put some more time on the proptests and hand-generated scenarios.

ValuedMammal:
These are the benchmarks on my machine (Apple silicon M4).

<details>
  <summary>benches</summary>

    ```sh
    get_1000_middle         time:   [44.127 ns 44.164 ns 44.210 ns]
    Found 9 outliers among 100 measurements (9.00%)
    1 (1.00%) low mild
    6 (6.00%) high mild
    2 (2.00%) high severe

    get_10000_near_end      time:   [75.652 ns 75.668 ns 75.684 ns]
    Found 6 outliers among 100 measurements (6.00%)
    1 (1.00%) low severe
    1 (1.00%) low mild
    1 (1.00%) high mild
    3 (3.00%) high severe

    get_10000_near_start    time:   [217.38 ns 217.68 ns 217.96 ns]
    Found 2 outliers among 100 measurements (2.00%)
    1 (1.00%) low severe
    1 (1.00%) high mild

    floor_at_1000           time:   [47.786 ns 47.965 ns 48.235 ns]
    Found 17 outliers among 100 measurements (17.00%)
    4 (4.00%) low severe
    1 (1.00%) low mild
    1 (1.00%) high mild
    11 (11.00%) high severe

    floor_at_10000          time:   [109.83 ns 109.96 ns 110.12 ns]
    Found 20 outliers among 100 measurements (20.00%)
    2 (2.00%) high mild
    18 (18.00%) high severe

    range_1000_middle_10pct time:   [646.26 ns 647.67 ns 649.10 ns]

    range_10000_large_50pct time:   [27.864 µs 27.872 µs 27.880 µs]
    Found 2 outliers among 100 measurements (2.00%)
    2 (2.00%) high severe

    range_10000_from_start  time:   [772.14 ns 773.35 ns 774.55 ns]

    insert_sparse_1000      time:   [1.4229 µs 1.4312 µs 1.4396 µs]
    Found 13 outliers among 100 measurements (13.00%)
    8 (8.00%) high mild
    5 (5.00%) high severe

    random_access_skiplist_1m
                            time:   [398.92 ns 399.43 ns 399.94 ns]
    Found 5 outliers among 100 measurements (5.00%)
    1 (1.00%) low severe
    3 (3.00%) low mild
    1 (1.00%) high severe

    Benchmarking random_access_linear_1m: Warming up for 3.0000 s
    Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.2s, enable flat sampling, or reduce sample count to 60.
    random_access_linear_1m time:   [1.1687 ms 1.1769 ms 1.1862 ms]
    Found 12 outliers among 100 measurements (12.00%)
    6 (6.00%) low severe
    5 (5.00%) low mild
    1 (1.00%) high mild
    ```
</details>

ValuedMammal:
Code review ACK e2ed5ad02d348f3362a618bc50e1efe6af724e4a

ValuedMammal (approved):

nymius:
> You have criterion 0.2 in core/Cargo.toml, but have used criterion 0.7 elsewhere.

Any reason to not update criterion to the latest stable release? Last time I checked it included a new benchmark grouping features from which this PR could benefit.

Not critical here, but worth considering.

crates/core/src/checkpoint.rs
evanlinjin: No, these are for optimising `CheckPointIter`.

crates/core/Cargo.toml
evanlinjin: Other crates were using `criterion = 0.7`. Unfortunately, we can't use `criterion = 0.8` as that has a MSRV of 1.86.

However, criterion is only used for benchmarks - I don't think we should enforce MSRV for running tests and benchmarks (separate discussion).

crates/core/tests/test_checkpoint_skiplist.rs
evanlinjin: I had a relook at this test and I think it's useless so I'll remove it.

crates/core/tests/test_checkpoint_skiplist.rs
evanlinjin: Outdated as test removed.

crates/core/src/checkpoint.rs
evanlinjin: This is outdated

crates/core/src/checkpoint.rs
evanlinjin: No, remember that checkpoints are a singly linked list and we can only iterate in one direction (backwards). The "start bounds" here represent the end of the iteration (or the earliest height).

crates/core/src/checkpoint.rs
```diff
@@ -6,6 +6,88 @@ use core::ops::RangeBounds;
...
 
+/// Returns the checkpoint index that index `i` should hold a skip pointer to.
+///
+/// Mirrors Bitcoin Core's `GetSkipHeight` (operating on checkpoint indices rather than block
+/// heights, so sparse chains work). The chosen targets give skip distances that grow
+/// exponentially as you walk back, yielding `O(log n)` traversal.
+///
+/// For `i < 2` returns `0` — `prev` already covers those distances trivially.
+fn skip_index(i: u32) -> u32 {
+    fn invert_lowest_one(n: u32) -> u32 {
```
nymius, line 20: I would add a line here, mention it clears the righteous bit in the representation. That's essential to understand the range that this function unlocks.

crates/core/src/checkpoint.rs
```diff
@@ -6,6 +6,88 @@ use core::ops::RangeBounds;
...
+    let mut curr = start.clone();
+    while curr.index > target {
+        let skip_i = skip_index(curr.index);
+        let skip_i_prev = skip_index(curr.index.saturating_sub(1));
+
+        // Only use `skip` if it's available, does not overshoot and `prev_skip` isn't better.
+        let use_skip = curr.skip.is_some()
+            && (skip_i == target
+                || (skip_i > target
+                    && !(skip_i > skip_i_prev.saturating_add(2) && skip_i_prev >= target)));
```
nymius: Can we store this `!(skip_i > skip_i_prev.saturating_add(2) && skip_i_prev >= target))` inside a variable with a meaningful name? Something like "next_skip_is_not_better". I'm still trying to get around the concept:

1. `skip > target` and `skip_prev >= target` but `skip < skip_prev - 2` means -> `|genesis| .... |target| .... |skip?| .... |skip_prev?|`

2. `skip > target` and  `skip_prev >= target` but `skip > skip_prev - 2` means -> `|genesis| .... |target| .... |skip_prev| .... |skip|`

3. `skip > target` and `skip_prev < target` means -> `|genesis| .... |skip_prev| ... |target| .... |skip|`

4. `skip > target` and  `skip_prev == target` means -> `|genesis| .... |skip_prev/target| .... |skip|`

Something that is also essential here is the deterministic nature of `skip_index`, which makes `curr.skip.index` == `skip_index(curr.index)`, so I would mention it.

Is this relation what gives sense to the boolean above, because all what is doing is tracking the index of the pointed skip node, and checking if it is better to jump now or wait for the next node in the walk.

nymius: Giving it another round of thought if `curr.skip` is a pointer to `CPInner`, wouldn't using `let skip_i = curr.skip.index` be equivalent to `let skip_i = skip_index(curr.index)`? What I prefer of the first expression is it doesn't hide any assumption, like the determinism I mention above.

nymius: I also think the visual help is useful in docs, so if you think I'm not mistaken, take whatever you need from the sketch above.

crates/core/src/checkpoint.rs
```diff
@@ -6,6 +6,88 @@ use core::ops::RangeBounds;
...
+        let skip = curr.skip();
+        let take_skip = match &skip {
+            Some(skip_cp) if skip_cp.height() < target_height => false,
+            Some(skip_cp) if skip_cp.height() == target_height => true,
+            // Skip lands above target. Prefer prev's skip if it's a strictly bigger jump (lands
+            // more than 2 heights lower than current's skip) that still reaches target.
+            Some(skip_cp) => match curr.prev().and_then(|p| p.skip()) {
+                Some(prev_skip_cp) => {
+                    let prev_skip_h = prev_skip_cp.height();
+                    !(skip_cp.height().saturating_sub(2) > prev_skip_h
```
nymius: Regarding the comment above `more than 2 heights lower than current's skip`, have you considered using `skip_cp.height().saturating_sub(prev_skip_cp.height()) > 2`. There is any deterrent for using it?

crates/core/tests/test_checkpoint_skiplist.rs
```diff
@@ -0,0 +1,324 @@
...
+
+#[test]
+fn test_skiplist_indices() {
+    // Build a chain spanning multiple log levels and assert that each node carries the pskip
+    // pointer mandated by the formula.
+    let mut cp = CheckPoint::new(0, BlockHash::all_zeros());
+    assert_eq!(cp.index(), 0);
+    assert!(cp.skip().is_none(), "genesis must not have a skip pointer");
+
+    const N: u32 = 5000;
```
nymius, line 33: This could be done in terms of `N = 2^k` (`log N = k`), so the amount of levels is explicit.

P.S.: I'm not sure the underlying tree is binary.

crates/core/tests/test_checkpoint_skiplist.rs
```diff
@@ -0,0 +1,324 @@
...
+    for height in 1..=N {
+        let hash = BlockHash::from_byte_array([(height % 256) as u8; 32]);
+        cp = cp.push(height, hash).unwrap();
+        assert_eq!(cp.index(), height);
+    }
+    assert_eq!(cp.index(), N);
+
+    // Walk the chain once and verify every non-genesis node skips to expected_skip_index(i).
+    let mut current = cp.clone();
```
nymius: Would `take()` work here? Dropping the chain as nodes are traversed?

crates/core/tests/test_checkpoint_skiplist.rs
```diff
@@ -0,0 +1,324 @@
...
+    // Insert a block in the middle
+    let hash = BlockHash::from_byte_array([25; 32]);
+    cp = cp.insert(25, hash);
+
+    // Check that indices are maintained correctly
+    let check = cp.get(50).unwrap();
+    assert_eq!(check.index(), 6); // 0, 10, 20, 25, 30, 40, 50
+
+    let check = cp.get(25).unwrap();
+    assert_eq!(check.index(), 3);
```
nymius: Why including these if the loop below is going to check them again?

crates/core/tests/test_checkpoint_skiplist.rs
```diff
@@ -0,0 +1,324 @@
...
+    assert_eq!(exclusive[0].height(), 200);
+    assert_eq!(exclusive[1].height(), 100);
+
+    // Unbounded range (..)
+    let all: Vec<_> = cp.range(..).collect();
+    assert_eq!(all.len(), 6);
+    assert_eq!(all.first().unwrap().height(), 500);
+    assert_eq!(all.last().unwrap().height(), 0);
+
+    // Range beyond chain bounds
```
nymius, line 129: `Non overlapping range beyond chain bounds`, `300..700` should produce something, right?

crates/core/tests/test_checkpoint_skiplist.rs
```diff
@@ -0,0 +1,324 @@
...
+    let beyond: Vec<_> = cp.range(600..700).collect();
+    assert!(beyond.is_empty());
+
+    // Range from genesis
```
nymius, line 133: Isn't this covered by the unbounded range?

crates/core/tests/test_checkpoint_skiplist.rs
evanlinjin: It's not a strict binary tree. I'd rather keep 5000 as it's not a power of two and therefore tests both branches of the `skip_index` formula.

crates/core/tests/test_checkpoint_skiplist.rs
evanlinjin: The test says `600..700` not `300..700`. 😅

crates/core/tests/test_checkpoint_skiplist.rs
evanlinjin: They test different code paths in `CheckPoint::range`.

crates/core/src/checkpoint.rs
evanlinjin: Updated. LMK what you think

crates/core/src/checkpoint.rs
evanlinjin: Let me know if the changes make things clearer.

crates/core/src/checkpoint.rs
evanlinjin: Changed

crates/core/tests/test_checkpoint_skiplist.rs
evanlinjin: Good point. Removed.

crates/core/tests/test_checkpoint_skiplist.rs
evanlinjin: Addressed.

crates/core/tests/test_checkpoint_skiplist.rs
nymius: Yes, it wasn't a typo, I was checking my understanding. What I meant is: for this check to work, you need the range to be "non-overlapping" because if there was an overlap, you would get something.

crates/core/src/checkpoint.rs
nymius: Precise, thanks!

crates/core/src/checkpoint.rs
nymius: They do, thanks!

P.S.: What's impossible to remove is my new understanding of the thing after review, so I don't know what someone looking at it with fresh eyes would say.

nymius (changes requested):
code ACK 74e592656610eefeba93fa7be9f1a5c44784b667

I would squash any changes produced during review to the commit where the reviewed lines belong. The same for clippy changes, those shouldn't survive standalone before PR approval.

A tentative organization could be:

- [feat(core): add skiplist to CheckPoint for faster traversal](https://github.com/bitcoindevkit/bdk/pull/2048/changes/15d31e372703483ee1eeb70c63e850b17fb4c4ac):
   - squashes [feat(core): replace fixed-stride skiplist with O(log n) pskip](https://github.com/bitcoindevkit/bdk/pull/2048/changes/892a51101c07140d8e84feffbd07d4b93e293726).
    - squashes [feat(chain,electrum): Make clippy happy](https://github.com/bitcoindevkit/bdk/pull/2048/changes/4e3dbee030b86cd9e5dc0b379c35df68601c71c6), but only in the lines of code created by the main commit, it shouldn't add new lines to diff.
- [test(core): add random-access skiplist benchmark](https://github.com/bitcoindevkit/bdk/pull/2048/changes/5bbe19ae84fc6ebb8b8f090942d44aa3fe7e7f82)
[feat(core)!: Remove unused methods of CheckPointEntry](https://github.com/bitcoindevkit/bdk/pull/2048/changes/4be89b98a6cc46c52a805f30f1473c55a6b9241a)
- [feat(core): Optimize CheckPointIter using pskip](https://github.com/bitcoindevkit/bdk/pull/2048/changes/74e592656610eefeba93fa7be9f1a5c44784b667)

evanlinjin:
@nymius I cleaned up the commits although I didn't exactly follow your commit-reorg plan. Let me know if it's sound.

nymius (approved):
ACK e095266afa2d80be0c6d2ef9d140ba2c67078a1d

PR #2052, opened 2025-09-29
chain: replace `ScriptBuf` with `&Script` in SPK index methods

yukibtc:
### Description

Replace `ScriptBuf` with `&Script` in `SpkTxOutIndex::index_of_spk` and `KeychainTxOutIndex::index_of_spk` methods, to avoid the need of cloning the `ScriptBuf` for a SPK index lookup.

### Changelog notice

Breaking changes:
- Change `SpkTxOutIndex::index_of_spk` and `KeychainTxOutIndex::index_of_spk` args from `ScriptBuf` to `&Script`.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/indexer/keychain_txout.rs
+++ b/crates/chain/src/indexer/keychain_txout.rs
@@ -12,7 +12,7 @@ use crate::{
 };
 use alloc::{borrow::ToOwned, vec::Vec};
 use bitcoin::{
-    key::Secp256k1, Amount, OutPoint, ScriptBuf, SignedAmount, Transaction, TxOut, Txid,
+    key::Secp256k1, Amount, OutPoint, Script, ScriptBuf, SignedAmount, Transaction, TxOut, Txid,
 };
 use core::{
     fmt::Debug,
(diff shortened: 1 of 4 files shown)


ValuedMammal:
Thank you @yukibtc.

evanlinjin:
Concept ACK.

For ergonomics, what do you think about having `spk: impl AsRef<Script>`? This way the caller can use `ScriptBuf`, `&Script`, `Box<Script>`, etc.

For methods that take ownership of the script, maybe we should change to `spk: impl Into<ScriptBuf>` - however, this will increase the scope of the PR - should be okay since we are relaxing bounds. Let me know what you think.

yukibtc:
> For ergonomics, what do you think about having spk: impl AsRef<Script>? This way the caller can use ScriptBuf, &Script, Box<Script>, etc.

Agree, I've updated to take `AsRef<Script>` instead.

> For methods that take ownership of the script, maybe we should change to spk: impl Into<ScriptBuf> - however, this will increase the scope of the PR - should be okay since we are relaxing bounds. Let me know what you think.

Yeah, maybe is better to do this in a dedicated PR

yukibtc:
Seems that CI failed for a electrsd timeout

aagbotemi (commented):
Code tested and result look okay.

oleonardolima (commented):
cACK c095145864a97b5efb4c90ef81989185b5a06998

evanlinjin (approved):
ACK c095145864a97b5efb4c90ef81989185b5a06998

PR #2053, opened 2025-10-01
Remove panic from stop gap scan loop

reez:
### Description

This PR addresses a panic path in stop gap scan loop. Removes `expect` used to compute gap boundary and instead tracks `consecutive_unused` with `gap_limit = stop_gap.max(1)` to decide when to stop scanning.

Spurred by trying to address https://github.com/bitcoindevkit/bdk_wallet/issues/30 for Esplora

### Notes to the reviewers

Open to any and all feedback on this.

Behavior is unchanged for typical cases, it just avoids a panic in the control flow.

### Changelog notice

```
### Fix:
- Avoid a panic in the Esplora stop‑gap scan loop by tracking consecutive unused scripts to compute the gap boundary.
```

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/esplora/src/async_ext.rs
+++ b/crates/esplora/src/async_ext.rs
@@ -184,10 +184,10 @@ async fn fetch_latest_blocks<S: Sleeper>(
     client: &esplora_client::AsyncClient<S>,
 ) -> Result<BTreeMap<u32, BlockHash>, Error> {
     Ok(client
-        .get_blocks(None)
+        .get_block_infos(None)
         .await?
         .into_iter()
-        .map(|b| (b.time.height, b.id))
+        .map(|b| (b.height, b.id))
         .collect())
 }
 
@@ -314,8 +314,11 @@ where
     type TxsOfSpkIndex = (u32, Vec<esplora_client::Tx>, HashSet<Txid>);
 
     let mut update = TxUpdate::<ConfirmationBlockTime>::default();
-    let mut last_index = Option::<u32>::None;
     let mut last_active_index = Option::<u32>::None;
+    // Use consecutive_unused so unused count drives stop gap.
+    let mut consecutive_unused = 0usize;
+    // Treat stop_gap = 0 as 1 while preserving original semantics for other values.
+    let gap_limit = stop_gap.max(1);
 
     loop {
         let handles = keychain_spks
@@ -352,8 +355,10 @@ where
         }
 
         for (index, txs, evicted) in handles.try_collect::<Vec<TxsOfSpkIndex>>().await? {
-            last_index = Some(index);
-            if !txs.is_empty() {
+            if txs.is_empty() {
+                consecutive_unused = consecutive_unused.saturating_add(1);
+            } else {
+                consecutive_unused = 0;
                 last_active_index = Some(index);
             }
             for tx in txs {
@@ -368,13 +373,7 @@ where
                 .extend(evicted.into_iter().map(|txid| (txid, start_time)));
         }
 
-        let last_index = last_index.expect("Must be set since handles wasn't empty.");
-        let gap_limit_reached = if let Some(i) = last_active_index {
-            last_index >= i.saturating_add(stop_gap as u32)
-        } else {
-            last_index + 1 >= stop_gap as u32
-        };
-        if gap_limit_reached {
+        if consecutive_unused >= gap_limit {
             break;
         }
     }
@@ -596,7 +595,10 @@ mod test {
         let res = chain_update(&client, &latest_blocks, &cp, &anchors).await;
         use esplora_client::Error;
         assert!(
-            matches!(*res.unwrap_err(), Error::HeaderHashNotFound(hash) if hash == genesis_hash),
+            matches!(
+                *res.unwrap_err(),
+                Error::HeaderHashNotFound(hash) if hash == genesis_hash
+            ),
             "`chain_update` should error if it can't connect to the local CP",
         );
 
--- a/crates/esplora/src/blocking_ext.rs
+++ b/crates/esplora/src/blocking_ext.rs
@@ -170,9 +170,9 @@ fn fetch_latest_blocks(
     client: &esplora_client::BlockingClient,
 ) -> Result<BTreeMap<u32, BlockHash>, Error> {
     Ok(client
-        .get_blocks(None)?
+        .get_block_infos(None)?
         .into_iter()
-        .map(|b| (b.time.height, b.id))
+        .map(|b| (b.height, b.id))
         .collect())
 }
 
@@ -282,8 +282,11 @@ fn fetch_txs_with_keychain_spks<I: Iterator<Item = Indexed<SpkWithExpectedTxids>
     type TxsOfSpkIndex = (u32, Vec<esplora_client::Tx>, HashSet<Txid>);
 
     let mut update = TxUpdate::<ConfirmationBlockTime>::default();
-    let mut last_index = Option::<u32>::None;
     let mut last_active_index = Option::<u32>::None;
+    // Use consecutive_unused so unused count drives stop gap.
+    let mut consecutive_unused = 0usize;
+    // Treat stop_gap = 0 as 1 while preserving original semantics for other values.
+    let gap_limit = stop_gap.max(1);
 
     loop {
         let handles = keychain_spks
@@ -321,8 +324,10 @@ fn fetch_txs_with_keychain_spks<I: Iterator<Item = Indexed<SpkWithExpectedTxids>
 
         for handle in handles {
             let (index, txs, evicted) = handle.join().expect("thread must not panic")?;
-            last_index = Some(index);
-            if !txs.is_empty() {
+            if txs.is_empty() {
+                consecutive_unused = consecutive_unused.saturating_add(1);
+            } else {
+                consecutive_unused = 0;
                 last_active_index = Some(index);
             }
             for tx in txs {
@@ -337,13 +342,7 @@ fn fetch_txs_with_keychain_spks<I: Iterator<Item = Indexed<SpkWithExpectedTxids>
                 .extend(evicted.into_iter().map(|txid| (txid, start_time)));
         }
 
-        let last_index = last_index.expect("Must be set since handles wasn't empty.");
-        let gap_limit_reached = if let Some(i) = last_active_index {
-            last_index >= i.saturating_add(stop_gap as u32)
-        } else {
-            last_index + 1 >= stop_gap as u32
-        };
-        if gap_limit_reached {
+        if consecutive_unused >= gap_limit {
             break;
         }
     }


crates/esplora/src/async_ext.rs
```diff
@@ -368,7 +371,8 @@ where
                 .extend(evicted.into_iter().map(|txid| (txid, start_time)));
         }
 
-        let last_index = last_index.expect("Must be set since handles wasn't empty.");
+        let last_index =
+            last_index.ok_or_else(|| Box::new(esplora_client::Error::InvalidResponse))?;
```
ValuedMammal: One suggestion is to leave `last_index` as an Option and wrap `stop_gap` in `Some(..)` when doing the comparison.

But there's a much better way to handle this, which is to introduce an `unused_count` variable. Now when processing handles, if `txs.is_empty()` we increment the unused count, else update the last active index (and reset the unused count). Finally, `break` once the unused count reaches the gap limit. AFAICT there'd be no reason to keep track of the `last_index` scanned.

reez: Took your suggestion and worked it in; didn't want to push it to this branch yet if I had the wrong idea, but here is the commit https://github.com/reez/bdk/commit/35538e25a1c3d02bda4a190e03d7b4bfdd209692 on a branch that builds on this branch.

I felt like I had to keep two small guardrails from the old logic:
- `processed_any` (still surface the “no handles ran” error)
- `gap_limit = stop_gap.max(1)`(keeps old “stop_gap==0 still means one empty derivation” behavior)

ValuedMammal: I don't think it's an error if `handles` turns up empty - we need a way to break from the loop once all of the keychain-spks are consumed.

ValuedMammal: Maybe `stop_gap = stop_gap.max(parallel_requests)`, since we check at least 1 spk-index for every request in `parallel_requests`.

crates/esplora/src/blocking_ext.rs
```diff
@@ -316,11 +316,17 @@ fn fetch_txs_with_keychain_spks<I: Iterator<Item = Indexed<SpkWithExpectedTxids>
...
+                return Err(Box::new(esplora_client::Error::InvalidResponse));
+            }
             break;
         }
 
         for handle in handles {
-            let (index, txs, evicted) = handle.join().expect("thread must not panic")?;
+            let handle_result = handle
+                .join()
+                .map_err(|_| Box::new(esplora_client::Error::InvalidResponse))?;
```
ValuedMammal: Because you're dropping the error it won't be very informative to propagate it upward. Instead this module can define its own `Error` type to handle errors that aren't covered by `esplora_client::Error`.

crates/esplora/src/async_ext.rs
reez: Super helpful thanks mammal, I made the below changes to that my side branch `esplora+mammal` that I can eventually fold back into this `esplora` branch if I'm on the right track

1. 

> I don't think it's an error if handles turns up empty - we need a way to break from the loop once all of the keychain-spks are consumed.

Good call. I made this change https://github.com/bitcoindevkit/bdk/commit/36073d0801faa747eed63cec624f0410464117b8 based on that

2. 

> Maybe stop_gap = stop_gap.max(parallel_requests), since we check at least 1 spk-index for every request in parallel_requests.

Good catch, made another change https://github.com/bitcoindevkit/bdk/commit/590edcfbdbf824c3b8ef08c45ba873f24b6fe3fc based on this

reez: pulled these changes into this pr now

crates/esplora/src/blocking_ext.rs
reez: Tried to address this with new commit 9ea0dd064c30f4ca79dac7a0e5e736a0c1d7f2c2, trying to do the most simple thing I could think of that addresses what you mentioned

crates/esplora/src/blocking_ext.rs
```diff
@@ -7,12 +7,54 @@ use bdk_core::{
...
 use std::thread::JoinHandle;
 
 use crate::{insert_anchor_or_seen_at_from_status, insert_prevouts};
 
-/// [`esplora_client::Error`]
-pub type Error = Box<esplora_client::Error>;
+#[derive(Debug)]
+pub enum Error {
+    Client(esplora_client::Error),
+    ThreadPanic(Option<String>),
```
evanlinjin: This seems over-engineered. Thread panics are highly unlikely because:

* No explicit panics in the thread - all errors are propagated.
* We only call `esplora_client` methods - if the method panics, it's a bug in the `esplora_client` library and should be addressed in the upstream crate.

`ThreadPanic` is really only protection against bugs in the dependency chain (should be addressed upstream) or extreme memory exhaustion (which would crash the application anyway).

crates/esplora/src/async_ext.rs
```diff
@@ -571,6 +568,15 @@ mod test {
...
     }
 
+    #[test]
+    fn ensure_last_index_none_returns_error() {
+        let last_index: Option<u32> = None;
+        let err = last_index
+            .ok_or_else(|| Box::new(esplora_client::Error::InvalidResponse))
+            .unwrap_err();
+        assert!(matches!(*err, esplora_client::Error::InvalidResponse));
+    }
```
evanlinjin: I don't think this is needed. If this fails, it's a problem with rust.

crates/esplora/src/blocking_ext.rs
```diff
@@ -316,13 +358,22 @@ fn fetch_txs_with_keychain_spks<I: Iterator<Item = Indexed<SpkWithExpectedTxids>
             .collect::<Vec<JoinHandle<Result<TxsOfSpkIndex, Error>>>>();
 
         if handles.is_empty() {
+            if !processed_any {
+                return Err(esplora_client::Error::InvalidResponse.into());
+            }
```
evanlinjin: Looks like you are trying to catch a empty-input? You'll need a justification on why an empty input should result in an error. My understanding is that the caller may create a `FullScanRequest` with an empty keychain and if they do, it's not problematic at all and we should just return nothing (which looks like what the original code is doing).

crates/esplora/src/blocking_ext.rs
```diff
@@ -529,6 +580,34 @@ mod test {
...
+
+    #[test]
+    fn ensure_last_index_none_returns_error() {
+        let last_index: Option<u32> = None;
+        let err = last_index
+            .ok_or_else(|| Error::from(esplora_client::Error::InvalidResponse))
+            .unwrap_err();
+        assert!(matches!(err, Error::Client(esplora_client::Error::InvalidResponse)));
+    }
+
```
evanlinjin: No need to test rust for bugs in our repo.

evanlinjin (changes requested):
Thanks for working on this!

Some great simplifications in logic. Some parts are over-engineered.

Note that `panic`s and `expects` indicate that "we should never hit this - unless there is a bug!" and I think it's justified to have them (where reasonable).

crates/esplora/src/blocking_ext.rs
```diff
@@ -241,15 +283,13 @@ fn chain_update(
...
-                local_cp_hash,
-            )));
+            return Err(esplora_client::Error::HeaderHashNotFound(local_cp_hash).into());
         }
     };
 
     tip = tip
         .extend(conflicts.into_iter().rev().map(|b| (b.height, b.hash)))
-        .expect("evicted are in order");
+        .map_err(|_| Error::from(esplora_client::Error::InvalidResponse))?;
```
reez: Add error variant, `Checkpoint` error. 

`Checkpoint` and wraps checkpoint type.

put in lib.rs, have async and blocking use new error

crates/esplora/src/blocking_ext.rs
```diff
@@ -316,13 +358,22 @@ fn fetch_txs_with_keychain_spks<I: Iterator<Item = Indexed<SpkWithExpectedTxids>
             .collect::<Vec<JoinHandle<Result<TxsOfSpkIndex, Error>>>>();
 
         if handles.is_empty() {
+            if !processed_any {
+                return Err(esplora_client::Error::InvalidResponse.into());
```
reez: remove this

crates/esplora/src/blocking_ext.rs
```diff
@@ -316,13 +358,22 @@ fn fetch_txs_with_keychain_spks<I: Iterator<Item = Indexed<SpkWithExpectedTxids>
...
 
         for handle in handles {
-            let (index, txs, evicted) = handle.join().expect("thread must not panic")?;
-            last_index = Some(index);
-            if !txs.is_empty() {
+            let handle_result = handle
+                .join()
+                .map_err(Error::from_thread_panic)?;
+            let (index, txs, evicted) = handle_result?;
+            processed_any = true;
```
reez: dont need processed_any

crates/esplora/src/blocking_ext.rs
```diff
@@ -316,13 +358,22 @@ fn fetch_txs_with_keychain_spks<I: Iterator<Item = Indexed<SpkWithExpectedTxids>
...
             break;
         }
 
         for handle in handles {
-            let (index, txs, evicted) = handle.join().expect("thread must not panic")?;
-            last_index = Some(index);
-            if !txs.is_empty() {
+            let handle_result = handle
+                .join()
+                .map_err(Error::from_thread_panic)?;
```
reez: dont need this if not worried about panic on thread panic

crates/esplora/src/blocking_ext.rs
```diff
@@ -406,7 +451,7 @@ fn fetch_txs_with_txids<I: IntoIterator<Item = Txid>>(
                 std::thread::spawn(move || {
                     client
                         .get_tx_info(&txid)
-                        .map_err(Box::new)
+                        .map_err(Error::from)
```
reez: map this more explicitly

crates/esplora/src/blocking_ext.rs
```diff
@@ -417,7 +462,10 @@ fn fetch_txs_with_txids<I: IntoIterator<Item = Txid>>(
         }
 
         for handle in handles {
-            let (txid, tx_info) = handle.join().expect("thread must not panic")?;
+            let handle_result = handle
+                .join()
+                .map_err(Error::from_thread_panic)?;
```
reez: dont worry about panic here

crates/esplora/src/blocking_ext.rs
```diff
@@ -468,7 +516,7 @@ fn fetch_txs_with_outpoints<I: IntoIterator<Item = OutPoint>>(
                 std::thread::spawn(move || {
                     client
                         .get_output_status(&op.txid, op.vout as _)
-                        .map_err(Box::new)
+                        .map_err(Error::from)
```
reez: be more explicit here too

crates/esplora/src/blocking_ext.rs
```diff
@@ -478,7 +526,10 @@ fn fetch_txs_with_outpoints<I: IntoIterator<Item = OutPoint>>(
         }
 
         for handle in handles {
-            if let Some(op_status) = handle.join().expect("thread must not panic")? {
+            let handle_result = handle
+                .join()
+                .map_err(Error::from_thread_panic)?;
```
reez: dont worry about panic here

crates/esplora/src/blocking_ext.rs
reez: You're right. I took your advice and removed `ThreadPanic`

crates/esplora/src/async_ext.rs
reez: Agreed, dropped this test

crates/esplora/src/blocking_ext.rs
reez: good point, dropped the processed_any check

crates/esplora/src/blocking_ext.rs
reez: right again, removed these

reez:
Updated PR due to great feedback from Evan, Mammal, Leonardo.

Noteworthy:
- Introduced a crate-wide Error enum (in lib.rs) with Client and Checkpoint variants. Both async and blocking extensions now return this so real Esplora failures bubble up via Client while checkpoint ordering issues surface as Checkpoint instead of panicking or masquerading as InvalidResponse
- Removed the custom ThreadPanic and reverted as suggested so we only guard against genuine remote errors
- Dropped the processed_any guard added previously
- Eliminated tests that provided no major value

crates/esplora/src/lib.rs
```diff
@@ -37,6 +38,38 @@ mod async_ext;
 #[cfg(feature = "async")]
 pub use async_ext::*;
 
+#[derive(Debug)]
+pub enum Error {
+    Client(esplora_client::Error),
+    Checkpoint(CheckPoint<BlockHash>),
+}
```
evanlinjin: ~~We're missing doc-comments for the `Error` type and it's variants!~~

~~We should make it obvious to the reader when the `Error::CheckPoint` variant occurs.~~

**Edit:** I don't think the `CheckPoint` variant is needed, so this `Error` type is not needed either.

crates/esplora/src/blocking_ext.rs
```diff
@@ -241,15 +238,13 @@ fn chain_update(
...
-                local_cp_hash,
-            )));
+            return Err(esplora_client::Error::HeaderHashNotFound(local_cp_hash).into());
         }
     };
 
     tip = tip
         .extend(conflicts.into_iter().rev().map(|b| (b.height, b.hash)))
-        .expect("evicted are in order");
+        .map_err(Error::Checkpoint)?;
```
evanlinjin: I still think `panic`/`expect` here is more appropriate.

1. `CheckPoint::extend` only errors if a block is not higher than the previous.
2. The logic to populate `conflicts` uses heights from `local_tip`. `local_tip` is guaranteed to have ordered heights (since it is a `CheckPoint` type).

evanlinjin (changes requested):
Avoid returning errors for internal logic bugs in BDK itself.

crates/esplora/src/lib.rs
reez: sounds good, `Error` type removed

crates/esplora/src/blocking_ext.rs
reez: makes sense, I reverted this so it uses `expect` again

reez:
I believe all review comments are addressed now; to recap current state this PR adds logic around removing panic tied to remote input, while leaving panics where appropriate based on PR review feedback. Thanks again for taking time to review this as I iterated on it (happy to continue tweaking if needed)!

ValuedMammal (commented):
ACK d4244e315017dfe570220f34f75b34e60a7fa8dd

Note: I think it would make sense to use `gap_limit = stop_gap.max(parallel_requests)`. Related discussion https://github.com/bitcoindevkit/bdk/pull/2053#discussion_r2407422581 and https://github.com/bitcoindevkit/bdk_wallet/issues/63.

reez:
> ACK [d4244e3](https://github.com/bitcoindevkit/bdk/commit/d4244e315017dfe570220f34f75b34e60a7fa8dd)
> 
> Note: I think it would make sense to use `gap_limit = stop_gap.max(parallel_requests)`. Related discussion [#2053 (comment)](https://github.com/bitcoindevkit/bdk/pull/2053#discussion_r2407422581) and [bitcoindevkit/bdk_wallet#63](https://github.com/bitcoindevkit/bdk_wallet/issues/63).

@ValuedMammal I just updated PR to use `gap_limit = stop_gap.max(parallel_requests)` now 👍

ValuedMammal:
Need to debug a CI fail. Maybe try rebasing this PR onto master.

Edit: As a workaround, pinning the rust toolchain to `nightly-2025-11-26` seems to have fixed it https://github.com/ValuedMammal/bdk/actions/runs/19878874973/job/56972371610, although I haven't pinpointed the exact issue.

ACK 074aed1c2f304906eb8e1abe92f9a0df76bc1f79

thunderbiscuit:
This is a bit of a blocker for the 2.3 release of bdk-ffi, I'd like to include it in there if possible. @reez do you mind trying the pin workaround proposed by @ValuedMammal?

reez:
> This is a bit of a blocker for the 2.3 release of bdk-ffi, I'd like to include it in there if possible. @reez do you mind trying the pin workaround proposed by @ValuedMammal?

@thunderbiscuit sure let me give my context

@ValuedMammal  has ACK’d this PR a few times, so after his last comment https://github.com/bitcoindevkit/bdk/pull/2053#issuecomment-3603062288 and because I didn’t want him to have to keep review+ACK’ing if no change actually needed on this PR I pinged him to ask if he wanted me to try either/both of those in this PR or keep the PR as is. My understanding is keep the PR as is (and maybe there would be a new issue opened scoped to resolve the coverage job), so I explicitly did not do the pin workaround for those reasons.

Additionally @oleonardolima has reviewed this previously and also let me know he’d be giving this PR a final review too, I know he’s been traveling a bunch though so there’s no rush from me on that.

Then from the bdk-ffi side of things my view point is this would be nice-to-have but is not a blocker for 2.3 release.

ValuedMammal:
> and maybe there would be a new issue opened scoped to resolve the coverage job

The CI fix is now in a separate PR #2080.

reez:
> #2080

Thanks for the heads up! 

I'll try to keep an eye on that PR so when it gets merged I can rebase this PR on it if needed for this PR to be merged.

reez:
> > #2080
> 
> Thanks for the heads up!
> 
> I'll try to keep an eye on that PR so when it gets merged I can rebase this PR on it if needed for this PR to be merged.

Rebased PR on `master` now that #2080 was merged, CI passing ✅

ValuedMammal (approved):
ACK 92dc6dd33ddbbf6c15cdcae03fdcdf5e8df4321d

crates/esplora/src/async_ext.rs
```diff
@@ -594,9 +591,12 @@ mod test {
 
         let anchors = BTreeSet::new();
         let res = chain_update(&client, &latest_blocks, &cp, &anchors).await;
-        use esplora_client::Error;
+        use esplora_client::Error as ClientError;
```
oleonardolima: nit: we probably don't need to alias as `ClientError` anymore, as we're not adding a different variant as before. It's not a blocker though, can be addressed in a follow-up.

oleonardolima (commented):
utACK 92dc6dd33ddbbf6c15cdcae03fdcdf5e8df4321d

It looks good for the removal of the single `last_index` use of `.expect()`.

As a side-note in a follow-up I'd like to take a look on how the remaining ones could be reached (except the join threads one, we discussed that before).

I left a single nit that is not a blocker as well, can be addressed in a follow-up PR.

I'd just suggest updating the PR title and description before merging it.

ValuedMammal:
Totally agree on updating the PR, i.e. what panic path remained, and how does this commit fix that?

crates/esplora/src/async_ext.rs
reez: went ahead and updated this since I'm assuming I'll get another round of review/ack when I update the PR title and description and I think there's still one more ACK I need before it can get merged

reez:
@oleonardolima @ValuedMammal I removed `ClientError` alias, updated the PR title+description, and amended the commit message to hopefully reflect current state, but let me know if anything is off there. Thanks!

oleonardolima (commented):
utACK 741a30c409bfe2f148b0d11b9905c7fc5c0ef948

ValuedMammal:
ACK 741a30c409bfe2f148b0d11b9905c7fc5c0ef948

crates/esplora/src/async_ext.rs
```diff
@@ -314,8 +314,9 @@ where
     type TxsOfSpkIndex = (u32, Vec<esplora_client::Tx>, HashSet<Txid>);
 
     let mut update = TxUpdate::<ConfirmationBlockTime>::default();
-    let mut last_index = Option::<u32>::None;
     let mut last_active_index = Option::<u32>::None;
+    let mut consecutive_unused = 0usize;
+    let gap_limit = stop_gap.max(parallel_requests);
```
evanlinjin: Nit: Add comments explaining rationale. Same for the blocking version.

evanlinjin (approved):
ACK 741a30c409bfe2f148b0d11b9905c7fc5c0ef948

crates/esplora/src/async_ext.rs
reez: Updated with brief rationale comments in both async and blocking. Happy to tweak wording if anyone wants anything different.

https://github.com/bitcoindevkit/bdk/compare/741a30c409bfe2f148b0d11b9905c7fc5c0ef948..81a2d280c975d98c093ccf006ae9d5674d43147f

evanlinjin: @reez that's not really a rationale though. The reader would like to know why we are doing this, not what the code is doing.

reez: @evanlinjin is this more in line with what would be better

```
// Use consecutive_unused so stop gap is based on consecutive empty spks instead of the last scanned index.
let mut consecutive_unused = 0usize;
// Use stop_gap.max(parallel_requests) since we check at least 1 spk-index for every request in parallel_requests.
let gap_limit = stop_gap.max(parallel_requests);
```

If so I'll update the pr with these comments instead.

crates/esplora/src/async_ext.rs
```diff
@@ -368,13 +373,7 @@ where
...
         }
 
-        let last_index = last_index.expect("Must be set since handles wasn't empty.");
-        let gap_limit_reached = if let Some(i) = last_active_index {
-            last_index >= i.saturating_add(stop_gap as u32)
-        } else {
-            last_index + 1 >= stop_gap as u32
-        };
-        if gap_limit_reached {
+        if consecutive_unused >= gap_limit {
```
evanlinjin, line 376: @reez this is the only place that `gap_limit` is used. Can you explain what is functionally different when the `gap_limit` is minimally bound to `parallel_requests` vs when it's not? 

I may be missing something, but I'm failing to see a clear rationale for the bound.

evanlinjin: I took a look at this original issue: https://github.com/bitcoindevkit/bdk_wallet/issues/63

I think bounding to 1 makes more sense - since a `stop_gap` of 0 is undefined and the behaviour we want is to "treat 0 as 1".

Bounding it to parallel requests does the same thing, but the intention is obscured imo. It seems like the intention is to have `consecutive_used` be a multiple of `gap_limit` if the `gap_limit` is originally set lower than `parallel_requests`? However we are using `>=`, and this breaks down when `gap_limit > parallel_requests` anyway.

If the rationale is not obvious, the reader might assume they are missing something and spend time trying to understand it and then find out there is nothing to be understood.

evanlinjin: Consider `parallel_requests=10`, `stop_gap=5`, first batch processes indices 0-9 where indices 0-4 have transactions and 5-9 are empty:

  Original code:
  - `last_active_index=4`, `last_index=9`
  - Check: 9 >= 4 + 5? → true, break

  If we do `.max(1)`:
  - consecutive_unused=5, gap_limit=5
  - 5 >= 5? → true, break ✓ (matches original)

  If we keep this PR as is with `.max(parallel_requests)`:
  - `consecutive_unused=5`, `gap_limit=10`
  - 5 >= 10? → false, continue ✗ (different from original)

  So `.max(parallel_requests)` changes the effective `stop_gap` when the user provides a value less than `parallel_requests`, which is a behavioral change from the original code. `.max(1)` preserves the original semantics while handling the `stop_gap=0` edge case.

### Suggestion

Change to `.max(1)`.

reez: Updated to `.max(1)` thanks for the deep dive review of that. Also tweaked the rationale comments. Updated PR description to mention `.max(1)` too now. Happy to adjust further if needed.

notmandatory:
This one will need another rebase since I merged #2123, sorry.

reez:
> This one will need another rebase since I merged https://github.com/bitcoindevkit/bdk/pull/2123, sorry.

rebased

luisschwab:
@reez you need to use `get_block_info`'s instead of `get_blocks`. It was deprecated in 0.12.3. Since the `Cargo.toml` imports `esplora-client` as `"0.12.1"` and not `"=0.12.1"`, `cargo` will pull the newest patch release.


See https://github.com/bitcoindevkit/rust-esplora-client/pull/164 for the rationale.

reez:
> @reez you need to use `get_block_info`'s instead of `get_blocks`. It was deprecated in 0.12.3. Since the `Cargo.toml` imports `esplora-client` as `"0.12.1"` and not `"=0.12.1"`, `cargo` will pull the newest patch release.
> 
> See [bitcoindevkit/rust-esplora-client#164](https://github.com/bitcoindevkit/rust-esplora-client/pull/164) for the rationale.

👍updated

luisschwab (approved):
ACK 745f805068ce33ee830b497c28fd37eb4c1aafb1

evanlinjin (approved):
ACK 745f805068ce33ee830b497c28fd37eb4c1aafb1

PR #2056, opened 2025-10-02
fix(core) `Checkpoint::drop` memory leak

oleonardolima:
based on #1997, depends on #2055

### Description

Fix memory leak bug in CheckPoint::drop by using Arc::into_inner.

Add tests (from the old PR) for memory leak + stack overflow when dropping CheckPoint.

### Notes to the reviewers

It should be merged after #2055.

### Changelog notice

```
### Fix:
- Fix memory leak bug in CheckPoint::drop by using `Arc::into_inner`.
```

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/core/src/checkpoint.rs
+++ b/crates/core/src/checkpoint.rs
@@ -37,21 +37,17 @@ struct CPInner<D> {
 /// https://github.com/bitcoindevkit/bdk/issues/1634
 impl<D> Drop for CPInner<D> {
     fn drop(&mut self) {
-        // Take out `prev` so its `drop` won't be called when this drop is finished
+        // Take out `prev` so its `drop` won't be called when this drop is finished.
         let mut current = self.prev.take();
+        // Collect nodes to drop later so we avoid recursive drop calls while not leaking memory.
         while let Some(arc_node) = current {
-            // Get rid of the Arc around `prev` if we're the only one holding a ref
-            // So the `drop` on it won't be called when the `Arc` is dropped.
-            //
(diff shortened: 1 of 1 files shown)


.github/zizmor.yml
```diff
@@ -0,0 +1,7 @@
+# Zizmor config
+rules:
+  unpinned-uses:
+    config:
+      policies:
+        # Allow pin by ref/tag
+        actions-rust-lang/setup-rust-toolchain: ref-pin
```
ValuedMammal, line 7: Since we're redefining the policy for `unpinned-uses`, we may need to reintroduce zizmor's default ref-pin policy for official GitHub actions.
```suggestion
        actions-rust-lang/setup-rust-toolchain: ref-pin
        actions/*: ref-pin
```

ValuedMammal (approved):
ACK a3bbba80066935ad976809d8df7ab5f8fefd0eff

.github/zizmor.yml
oleonardolima: Alright, I'll handle this on #2055.

evanlinjin (approved):
self-ACK 55161654ec107b7889ca3296aa64b32e3bc72eff

I rebased on the current state of #2055.

oleonardolima:
I've rebased it on top of master, so it should be ready to go.

evanlinjin (approved):
ACK 0bd93dad114b90c3f08df9cc892625e250edb718

PR #2062, opened 2025-10-09
fix(chain): return full TxNode from CanonicalIter with debug logging

Rahamath-unnisa:
This pull request enhances the CanonicalIter in the chain module to return the full TxNode—which includes Txid, Transaction, and CanonicalReason—instead of just the Txid. This improvement allows better tracking of canonical transactions along with the reason why each transaction is considered canonical.

Additionally, debug logging has been added throughout the canonicalization process. This logging prints transaction IDs as they are processed, which makes it easier to trace the iteration flow and understand how conflicts and ancestor-descendant relationships are resolved.

These changes improve the testability and observability of canonical transaction processing in the BDK library, which is particularly useful when working with transaction graphs that have complex conflicts or multiple anchors.

Notes to Reviewers

The debug logs are temporary and intended to help during development and testing.

This change does not alter the underlying canonicalization logic but provides more detailed output and richer information per transaction.

No existing APIs are broken; only the iterator’s item type has been extended to provide more context.

### DIFF
--- a/crates/chain/src/canonical_iter.rs
+++ b/crates/chain/src/canonical_iter.rs
@@ -6,6 +6,7 @@ use alloc::collections::BTreeSet;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
 use bdk_core::BlockId;
+use bitcoin::hashes::Hash;
 use bitcoin::{Transaction, Txid};
 
 type CanonicalMap<A> = HashMap<Txid, (Arc<Transaction>, CanonicalReason<A>)>;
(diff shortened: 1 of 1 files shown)


crates/chain/src/canonical_iter.rs
```diff
@@ -205,19 +206,22 @@ impl<A: Anchor, C: ChainOracle> Iterator for CanonicalIter<'_, A, C> {
...
                     .expect("reason must exist");
+
                 return Some(Ok((txid, tx, reason)));
             }
 
             if let Some((txid, tx)) = self.unprocessed_assumed_txs.next() {
                 if !self.is_canonicalized(txid) {
                     self.mark_canonical(txid, tx, CanonicalReason::assumed());
                 }
+                continue;
```
oleonardolima, line 224: You shouldn't change the behavior of the canonicalization algorithm.

oleonardolima (commented):
It doesn't look like the code is really changing the types to return a TxNode, there's no change on it yet.

evanlinjin:
The quality of this PR is no different than #2063. Please spend time understanding the codebase before submitting PRs otherwise you'll end up wasting maintainers' time.

Maintainers will respond to genuine attempts to understand the codebase.

PR #2071, opened 2025-11-06
ci: Pin `home` to 0.5.11

ValuedMammal:
### Description

PR pins `home` dependency in CI to `0.5.11` in order to build on rust 1.85.0.

#### All Submissions:

### DIFF
--- a/ci/pin-msrv.sh
+++ b/ci/pin-msrv.sh
@@ -9,3 +9,5 @@ set -euo pipefail
 
 # cargo clean
 # rustup override set 1.85.0
+
+cargo update -p home --precise "0.5.11"


tnull (commented):
Instead of pinning `home` (which is just a sub-dependency of `electrsd` through `which`)  you should be able to rather bump `electrsd` to 0.36.1 that includes https://github.com/RCasatta/electrsd/pull/114, simply dropping the unnecessary dependency altogether.

tnull:
> Instead of pinning `home` (which is just a sub-dependency of `electrsd` through `which`) you should be able to rather bump `electrsd` to 0.36.1 that includes [RCasatta/electrsd#114](https://github.com/RCasatta/electrsd/pull/114), simply dropping the unnecessary dependency altogether.

Ah, nvm, I realize you haven't made the switch to `corepc-node` yet. Would be the 'real' fix here though.

evanlinjin (approved):
ACK 93155d2c50fea094c15e04de36d5b60bea75d173

PR #2074, opened 2025-11-21
Fix `ChainPosition` ordering

evanlinjin:
## Description

This PR fixes incorrect ordering of `ChainPosition` that affected how wallet transactions are displayed. Previously, unconfirmed transactions that were never seen in the mempool would incorrectly appear before transactions with mempool timestamps due to the derived `Ord` implementation treating `None` as less than `Some(_)`.

### Problem

The derived `Ord` implementation for `ChainPosition::Unconfirmed` caused incorrect transaction ordering:
- `Unconfirmed { first_seen: None, last_seen: None }` (never in mempool)
- Would appear **before** `Unconfirmed { first_seen: Some(10), last_seen: Some(10) }` (seen in mempool)

This resulted in a confusing wallet display where transactions never broadcast would appear before pending mempool transactions.

### Solution

Implemented manual `Ord` and `PartialOrd` traits for `ChainPosition` with the correct ordering:

1. **Confirmed transactions** (earliest first by block height)
2. **Unconfirmed transactions seen in mempool** (ordered by `first_seen`)
3. **Unconfirmed transactions never seen** (these come last)

Additional ordering rules:
- At the same confirmation height, transitive confirmations come before direct confirmations (as they may actually be confirmed earlier)
- Tie-breaking uses transaction IDs for deterministic ordering

### Fixed
- Fixed `ChainPosition` ordering so unconfirmed transactions never seen in mempool appear last instead of first

### Changed
- `ChainPosition`, `CanonicalTx`, and `FullTxOut` now require `A: Ord` for their `Ord` implementations
- Simplified `FullTxOut` ordering to only use essential fields (chain_position, outpoint, spent_by)

### All Submissions:

### Bugfixes:

### DIFF
--- a/crates/chain/src/canonical_view.rs
+++ b/crates/chain/src/canonical_view.rs
@@ -40,7 +40,7 @@ use crate::{
 /// This struct represents a transaction that has been determined to be canonical (not
 /// conflicted). It includes the transaction itself along with its position in the chain (confirmed
 /// or unconfirmed).
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct CanonicalTx<A> {
     /// The position of this transaction in the chain.
     ///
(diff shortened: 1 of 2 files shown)


aagbotemi (commented):
tACK d6593dad8a1ff6ed3e67cf15a15f995f7977989d

Ran the test locally and it was successful.
Screenshot from 2025-12-09 15-07-22

ValuedMammal (approved):

evanlinjin:
I rebased on master and simplified the `Ord` logic a bit. Needs a re-review, thanks!

aagbotemi (commented):
@evanlinjin thank you for simplifying the `Ord` logic. I think it would be nice to add a test checking when the `first_seen` is equal and the `last_seen` is different, to explicitly verify the tiebreaker logic.

evanlinjin:
@aagbotemi Thank you for the suggestion. There was actually a slight bug in the logic. I fixed it and simplified the `Ord` implementation even further.

aagbotemi (commented):
tACK 85bfcdc5da053d1c6794912b70d7ad84f03c7bc6

Ran the test locally and it was successful.
Screenshot from 2026-01-10 21-50-50

oleonardolima (approved):
ACK 85bfcdc5da053d1c6794912b70d7ad84f03c7bc6

evanlinjin:
@oleonardolima I rebased on master. Feel free to ACK and merge.

PR #2081, opened 2025-12-10
Add spent/created_txouts to SPK and Keychain TxOut indexes

notmandatory:
### Description

Implement spent_txouts and created_txouts methods on SpkTxOutIndex and KeychainTxOutIndex. These methods return SpentTxOut and CreatedTxOut structs allowing callers to access complete transaction output information including script pubkeys and values.

### Notes to the reviewers

This info is useful to users who use the new WalletEvent data in bitcoindevkit/bdk_wallet#319. In particular to determine the addresses and amounts for TxOuts that are sent/spent or have been received in a newly seen or confirmed transaction. See: [bdk_wallet#319 comment](https://github.com/bitcoindevkit/bdk_wallet/pull/319#issuecomment-3630027488)

If/when this PR is merged I'll make a corresponding one to `bdk_wallet::Wallet`.

### Changelog notice

Added

* New spent_txouts and created_txouts methods on SpkTxOutIndex and KeychainTxOutIndex.
* New SpentTxOut and CreatedTxOut structs returned by spent_txouts and created_txouts functions.

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/indexer/spk_txout.rs
+++ b/crates/chain/src/indexer/spk_txout.rs
@@ -5,11 +5,9 @@ use core::ops::RangeBounds;
 
 use crate::{
     collections::{hash_map::Entry, BTreeMap, BTreeSet, HashMap},
-    IndexedTxOuts, Indexer,
+    Indexer,
 };
-use bitcoin::{Amount, OutPoint, Script, ScriptBuf, SignedAmount, Transaction, TxOut, Txid};
-
-use alloc::vec::Vec;
+use bitcoin::{Amount, OutPoint, Script, ScriptBuf, SignedAmount, Transaction, TxIn, TxOut, Txid};
 
 /// An index storing [`TxOut`]s that have a script pubkey that matches those in a list.
 ///
@@ -320,17 +318,14 @@ impl<I: Clone + Ord + core::fmt::Debug> SpkTxOutIndex<I> {
         (sent, received)
     }
 
-    /// Collects the sent and received [`TxOut`]s for `tx` on the script pubkeys in `range`.
-    /// TxOuts are *sent* when a script pubkey in the `range` is on an input and *received* when
-    /// it is on an output. For `sent` to be computed correctly, the index must have already
-    /// scanned the output being spent. Calculating received just uses the [`Transaction`]
-    /// outputs directly, so it will be correct even if it has not been scanned.
+    /// Returns the relevant [`SpentTxOut`]s for a [`Transaction`]
     ///
-    /// Returns a tuple of (sent_txouts, received_txouts).
+    /// TxOuts are *spent* when an indexed script pubkey is found in one of the transaction's
+    /// inputs. For these to be computed correctly, the index must have already scanned the
+    /// output being spent.
     ///
     /// # Example
-    /// Shows the addresses of the TxOut sent from or received by a Transaction relevant to all spks
-    /// in this index.
+    /// Shows the addresses of the TxOut spent from a Transaction relevant to spks in this index.
     ///
     /// ```rust
     /// # use bdk_chain::spk_txout::SpkTxOutIndex;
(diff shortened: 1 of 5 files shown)


crates/chain/src/indexer/spk_txout.rs
```diff
@@ -318,6 +320,74 @@ impl<I: Clone + Ord + core::fmt::Debug> SpkTxOutIndex<I> {
...
+    /// # fn example() -> Result<(), Box<dyn std::error::Error>> {
+    /// let mut index = SpkTxOutIndex::<u32>::default();
+    ///
+    /// // ... scan transactions to populate the index ...
+    /// # let tx = Transaction { version: bitcoin::transaction::Version::TWO, lock_time: bitcoin::locktime::absolute::LockTime::ZERO, input: vec![], output: vec![] };
+    ///
+    /// // Get sent and received txouts for a transaction across all tracked addresses
+    /// let (sent_txouts, received_txouts) = index.sent_and_received_txouts(&tx, ..);
+    ///
+    /// // Display addresses and amounts
```
notmandatory, line 344: @tnull would this information be useful for LDKnode when handling the transactions in WalletEvents ?

tnull: Ah, yes, that would be very helpful, we just recently got requests to be able to see which address was sent to: https://github.com/lightningdevkit/ldk-node/issues/684

Do you happen to see a way to include the output's index in this API, too? (cf. https://github.com/lightningdevkit/ldk-node/issues/717). Maybe the return type could be `(Vec<(usize, TxOut)>, Vec<(usize, TxOut)>)`, or a similar struct representation?

notmandatory: Good idea, since the input & output Vecs of Transaction are in the correct order I'm able to return the index info with the TxOuts. See 6ff9f26cdbd6f4dcbe14e5b2a16ac8406e1636fb.

notmandatory: @tnull did you get a chance to look at this again? it now returns the index and TxOut as you suggested.

tnull: Cool, thank you! Looks good!

evanlinjin (changes requested):
Thanks for this, Steve. The motivation is solid. I have some API design suggestions that I think would make this more ergonomic and precise.

## Split into two methods: `spent_txouts` and `created_txouts`

"Sent" and "received" are wallet-UX terms that obscure what's actually happening at this layer. "Spent" and "created" map directly to Bitcoin's model—consuming existing UTXOs and creating new ones. This makes the API more self-documenting.

Splitting also means callers only pay for what they need. If someone only cares about detecting which UTXOs got consumed, they shouldn't have to iterate outputs too.

## Return `OutPoint`s, not just indices

For spent outputs, the caller almost certainly wants to know *which* UTXO was consumed. The current implementation gives the input index, but what you really need is `txin.previous_output`. Making the caller fish that out separately is awkward when the method already has that information.

For created outputs, returning the `OutPoint` (txid + vout) means callers can immediately track or spend that output without computing the txid themselves.

## Use structs instead of tuples

```rust
pub struct SpentTxOut<I> {
    pub txout: TxOut,
    // Have a method: .outpoint()
    pub spending_input: TxIn,
    pub spending_input_index: usize,
    pub spk_index: I,
}

pub struct CreatedTxOut<I> {
    pub outpoint: OutPoint,
    pub txout: TxOut,
    pub spk_index: I,
}
```

`spent.outpoint()` is immediately clear. `spent.0` requires checking docs every time.

## Return iterators instead of `Vec`

More idiomatic, no upfront allocation, composable with `.filter()`, `.map()`, etc. If someone wants a `Vec`, they can `.collect()`.

## Drop the range parameter

The range conflates two concerns. Filtering by SPK index is trivial with iterators:

```rust
index.spent_txouts(&tx)
    .filter(|s| my_range.contains(&s.spk_index))
```

notmandatory:
> Split into two methods: `spent_txouts` and `created_txouts`
> ...

@evanlinjin thanks for the suggestions I think they're a big improvement. I've pushed the changes in a new commit but if everyone is happy with this approach I'll squash into the last commit.

evanlinjin (approved):
ACK d17d85fd54a4783f50b08f1753b322b5060f2f4e

notmandatory:
In the above description I had a follow-up task to make a corresponding PR on `bdk_wallet` but I don't think one is needed since these new functions can already be called from `Wallet` (once `bdk_wallet` 3.0 is released with  `bdk_chain` 0.24):

```rust
let spent_txouts: Vec<SpentTxOut<_>> = wallet.spk_index().spent_txouts(&tx).collect();
dbg!(&spent_txouts);
let created_txouts: Vec<CreatedTxOut<_>> = wallet.spk_index().created_txouts(&tx).collect();
dbg!(&created_txouts);
```

Does anyone need these function functions back-ported to a `bdk_chain` 0.23.x so they can be included in a `bdk_wallet` 2.x release?

oleonardolima:
> In the above description I had a follow-up task to make a corresponding PR on `bdk_wallet` but I don't think one is needed since these new functions can already be called from `Wallet` (once `bdk_wallet` 3.0 is released with `bdk_chain` 0.24):
> 
> ```rust
> let spent_txouts: Vec<SpentTxOut<_>> = wallet.spk_index().spent_txouts(&tx).collect();
> dbg!(&spent_txouts);
> let created_txouts: Vec<CreatedTxOut<_>> = wallet.spk_index().created_txouts(&tx).collect();
> dbg!(&created_txouts);
> ```
> 
> Does anyone need these function functions back-ported to a `bdk_chain` 0.23.x so they can be included in a `bdk_wallet` 2.x release?

@notmandatory I wasn't following this discussion here, but I think it's something we need to backport now, given that bdk_wallet v3.0.0 will stick to `chain-0.23.x`. Let me know if it's required or not.

PR #2083, opened 2025-12-11
fix: no `Debug` on `Display` implementations

### LINKED ISSUE #1933: Display implementations should not expose `Debug` output

Dmenec:
### Description

Fixes https://github.com/bitcoindevkit/bdk/issues/1933, remove the usage of `Debug` on `Display` implemetations mentioned in https://github.com/bitcoindevkit/bdk/pull/1881#issuecomment-2718412512

### Changelog notice

- Replaced  `Debug` usage with `Display` messages for `InsertDescriptorError`, `CalculateFeeError` and `StoreError`.
- `InvalidMagicBytes` now displays expected and actual bytes in hexadecimal instead of using `Debug`.
- Added a new `short_descriptor` function to shorten descriptors for concise error output.
- Added detailed `SyncRequest` `Display` for Esplora and Electrum examples.

#### All Submissions:
* [x] I have signed all my commits
* [x] I ran `just check`, `just fmt` and `just test` before committing

### DIFF
--- a/crates/chain/src/indexer/keychain_txout.rs
+++ b/crates/chain/src/indexer/keychain_txout.rs
@@ -973,25 +973,27 @@ pub enum InsertDescriptorError<K> {
     },
 }
 
-impl<K: core::fmt::Debug> core::fmt::Display for InsertDescriptorError<K> {
+impl<K: core::fmt::Display> core::fmt::Display for InsertDescriptorError<K> {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         match self {
             InsertDescriptorError::DescriptorAlreadyAssigned {
-                existing_assignment: existing,
+                existing_assignment,
                 descriptor,
             } => {
                 write!(
                     f,
-                    "attempt to re-assign descriptor {descriptor:?} already assigned to {existing:?}"
+                    "descriptor '{}' is already in use by another keychain '{}'",
+                    descriptor, existing_assignment
                 )
             }
             InsertDescriptorError::KeychainAlreadyAssigned {
-                existing_assignment: existing,
+                existing_assignment,
                 keychain,
             } => {
                 write!(
                     f,
-                    "attempt to re-assign keychain {keychain:?} already assigned to {existing:?}"
+                    "keychain '{}' is already associated with another descriptor '{}'",
+                    keychain, existing_assignment
                 )
             }
         }
@@ -999,7 +1001,7 @@ impl<K: core::fmt::Debug> core::fmt::Display for InsertDescriptorError<K> {
 }
 
 #[cfg(feature = "std")]
-impl<K: core::fmt::Debug> std::error::Error for InsertDescriptorError<K> {}
+impl<K: core::fmt::Display + core::fmt::Debug> std::error::Error for InsertDescriptorError<K> {}
 
 /// `ChangeSet` represents persistent updates to a [`KeychainTxOutIndex`].
 ///
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -256,15 +256,33 @@ pub enum CalculateFeeError {
 impl fmt::Display for CalculateFeeError {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         match self {
-            CalculateFeeError::MissingTxOut(outpoints) => write!(
-                f,
-                "missing `TxOut` for one or more of the inputs of the tx: {outpoints:?}",
-            ),
-            CalculateFeeError::NegativeFee(fee) => write!(
-                f,
-                "transaction is invalid according to the graph and has negative fee: {}",
-                fee.display_dynamic()
-            ),
+            CalculateFeeError::MissingTxOut(outpoints) => {
+                let max_show = 3;
+                let shown: Vec<_> = outpoints.iter().take(max_show).collect();
+                let remaining = outpoints.len().saturating_sub(max_show);
+
+                write!(f, "cannot calculate fee, missing previous output(s): ")?;
+                if outpoints.is_empty() {
+                    write!(f, "<none>")
+                } else {
+                    write!(f, "{}", shown[0])?;
+                    for op in &shown[1..] {
+                        write!(f, ", {}", op)?;
+                    }
+                    if remaining > 0 {
+                        write!(f, " (+{} more)", remaining)?;
+                    }
+                    Ok(())
+                }
+            }
+            CalculateFeeError::NegativeFee(fee) => {
+                write!(
+                    f,
+                    "invalid transaction: negative fee {}",
+                    fee.display_dynamic()
+                )?;
+                Ok(())
+            }
         }
     }
 }
--- a/crates/file_store/src/lib.rs
+++ b/crates/file_store/src/lib.rs
@@ -25,13 +25,24 @@ pub enum StoreError {
 
 impl core::fmt::Display for StoreError {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        fn fmt_hex_bytes(f: &mut core::fmt::Formatter<'_>, bytes: &[u8]) -> core::fmt::Result {
+            for &b in bytes {
+                write!(f, "{:02x}", b)?;
+            }
+            Ok(())
+        }
+
         match self {
-            Self::Io(e) => write!(f, "io error trying to read file: {e}"),
-            Self::InvalidMagicBytes { got, expected } => write!(
-                f,
-                "file has invalid magic bytes: expected={expected:?} got={got:?}",
-            ),
-            Self::Bincode(e) => write!(f, "bincode error while reading entry {e}"),
+            Self::Io(e) => write!(f, "io error while reading store file: {}", e),
+            Self::Bincode(e) => write!(f, "bincode error while decoding entry {}", e),
+            Self::InvalidMagicBytes { got, expected } => {
+                write!(f, "invalid magic bytes: ")?;
+                write!(f, "expected 0x")?;
+                fmt_hex_bytes(f, expected)?;
+                write!(f, ", got 0x")?;
+                fmt_hex_bytes(f, got)?;
+                Ok(())
+            }
         }
     }
 }
--- a/examples/example_electrum/src/main.rs
+++ b/examples/example_electrum/src/main.rs
@@ -210,9 +210,22 @@ fn main() -> anyhow::Result<()> {
                     .chain_tip(chain_tip.clone())
                     .inspect(|item, progress| {
                         let pc = (100 * progress.consumed()) as f32 / progress.total() as f32;
-                        eprintln!("[ SCANNING {pc:03.0}% ] {item}");
+                        match item {
+                            bdk_chain::spk_client::SyncItem::Spk((keychain, index), spk) => {
+                                eprintln!(
+                                    "[ SCANNING {pc:3.0}% ] script {} {} {}",
+                                    keychain, index, spk
+                                );
+                            }
+                            bdk_chain::spk_client::SyncItem::Txid(txid) => {
+                                eprintln!("[ SCANNING {pc:3.0}% ] txid {}", txid);
+                            }
+                            bdk_chain::spk_client::SyncItem::OutPoint(op) => {
+                                eprintln!("[ SCANNING {pc:3.0}% ] outpoint {}", op);
+                            }
+                        }
+                        let _ = io::stderr().flush();
                     });
-
             let canonical_view = graph.canonical_view(
                 &*chain,
                 chain_tip.block_id(),
--- a/examples/example_esplora/src/main.rs
+++ b/examples/example_esplora/src/main.rs
@@ -215,8 +215,20 @@ fn main() -> anyhow::Result<()> {
                     .chain_tip(local_tip.clone())
                     .inspect(|item, progress| {
                         let pc = (100 * progress.consumed()) as f32 / progress.total() as f32;
-                        eprintln!("[ SCANNING {pc:03.0}% ] {item}");
-                        // Flush early to ensure we print at every iteration.
+                        match item {
+                            bdk_chain::spk_client::SyncItem::Spk((keychain, index), spk) => {
+                                eprintln!(
+                                    "[ SCANNING {pc:3.0}% ] script {} {} {}",
+                                    keychain, index, spk
+                                );
+                            }
+                            bdk_chain::spk_client::SyncItem::Txid(txid) => {
+                                eprintln!("[ SCANNING {pc:3.0}% ] txid {}", txid);
+                            }
+                            bdk_chain::spk_client::SyncItem::OutPoint(op) => {
+                                eprintln!("[ SCANNING {pc:3.0}% ] outpoint {}", op);
+                            }
+                        }
                         let _ = io::stderr().flush();
                     });
 
--- a/crates/chain/tests/test_keychain_txout_index.rs
+++ b/crates/chain/tests/test_keychain_txout_index.rs
@@ -18,6 +18,15 @@ enum TestKeychain {
     Internal,
 }
 
+impl core::fmt::Display for TestKeychain {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            TestKeychain::External => write!(f, "External"),
+            TestKeychain::Internal => write!(f, "Internal"),
+        }
+    }
+}
+
 fn parse_descriptor(descriptor: &str) -> Descriptor<DescriptorPublicKey> {
     let secp = bdk_chain::bitcoin::secp256k1::Secp256k1::signing_only();
     Descriptor::<DescriptorPublicKey>::parse_descriptor(&secp, descriptor)
--- a/crates/core/src/spk_client.rs
+++ b/crates/core/src/spk_client.rs
@@ -225,6 +225,7 @@ impl<I, D> SyncRequestBuilder<I, D> {
 /// [`chain_tip`](SyncRequestBuilder::chain_tip) (if provided).
 ///
 /// ```rust
+/// # use std::io::{self, Write};
 /// # use bdk_chain::{bitcoin::{hashes::Hash, ScriptBuf}, local_chain::LocalChain};
 /// # use bdk_chain::spk_client::SyncRequest;
 /// # let (local_chain, _) = LocalChain::from_genesis(Hash::all_zeros());
@@ -236,7 +237,22 @@ impl<I, D> SyncRequestBuilder<I, D> {
 ///     // Provide list of scripts to scan for transactions against.
 ///     .spks(scripts)
 ///     // This is called for every synced item.
-///     .inspect(|item, progress| println!("{} (remaining: {})", item, progress.remaining()))
+///     .inspect(|item, progress| {
+///         let pc = (100.0 * progress.consumed() as f32) / progress.total() as f32;
+///         match item {
+///             // In this example I = (), so the first field of Spk is unit.
+///             bdk_chain::spk_client::SyncItem::Spk((), spk) => {
+///                 eprintln!("[ SCANNING {pc:03.0}% ] script {}", spk);
+///             }
+///             bdk_chain::spk_client::SyncItem::Txid(txid) => {
+///                 eprintln!("[ SCANNING {pc:03.0}% ] txid {}", txid);
+///             }
+///             bdk_chain::spk_client::SyncItem::OutPoint(op) => {
+///                 eprintln!("[ SCANNING {pc:03.0}% ] outpoint {}", op);
+///             }
+///         }
+///         let _ = io::stderr().flush();
+///     })
 ///     // Finish constructing the sync request.
 ///     .build();
 /// ```


qatkk:
Concept ACK — this approach looks good, thanks for the changes. I ran the tests locally and they all passed.

While trying to recreate the initial issue’s scenario (https://github.com/bitcoindevkit/bdk/issues/1933), I found that `bdk_wallet` `master` still references the older API names (`list_canonical_txs`, `filter_chain_unspents`, `filter_chain_txouts`, `from_genesis_hash`), whereas this PR’s `bdk_chain` exposes the newer canonicalization API (`canonical_iter`) and `LocalChain::from_genesis`. As a result, building `bdk_wallet` against this PR failed. Is there any branch in `bdk_wallet` that already adopts the new APIs, so I can test against them?

I have a few small suggestions below. Most of these are nits, but since this PR is already addressing the area, I thought it’d be reasonable to mention them here:

- It might be nice to add a CI to prevent using `Debug` for `Display` in future changes.

- In the `short_descriptor` function, it could be useful to show the last 8 characters of the descriptor so the full checksum is visible.

- A small note on `InvalidMagicBytes`: it might be helpful to also show any difference between the length of expected and received byte.

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -973,33 +973,47 @@ pub enum InsertDescriptorError<K> {
...
             InsertDescriptorError::DescriptorAlreadyAssigned {
-                existing_assignment: existing,
+                existing_assignment,
                 descriptor,
             } => {
+                let short_desc = short_descriptor(descriptor, 4);
                 write!(
                     f,
-                    "attempt to re-assign descriptor {descriptor:?} already assigned to {existing:?}"
+                    "Descriptor '{}' is already in use by keychain '{}'. Please use a different descriptor.",
```
evanlinjin: Nit

```suggestion
                    "descriptor '{}' is already in use by another keychain '{}'",
```

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -973,33 +973,47 @@ pub enum InsertDescriptorError<K> {
...
             InsertDescriptorError::KeychainAlreadyAssigned {
-                existing_assignment: existing,
+                existing_assignment,
                 keychain,
             } => {
+                let short_desc = short_descriptor(existing_assignment, 4);
                 write!(
                     f,
-                    "attempt to re-assign keychain {keychain:?} already assigned to {existing:?}"
+                    "Keychain '{}' is already associated with descriptor '{}'. Please choose a different keychain.",
```
evanlinjin: Nit

```suggestion
                    "keychain '{}' is already associated with another descriptor '{}'",
```

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -973,33 +973,47 @@ pub enum InsertDescriptorError<K> {
...
+                    keychain, short_desc
                 )
             }
         }
     }
 }
 
 #[cfg(feature = "std")]
-impl<K: core::fmt::Debug> std::error::Error for InsertDescriptorError<K> {}
+impl<K: core::fmt::Display + core::fmt::Debug> std::error::Error for InsertDescriptorError<K> {}
```
evanlinjin, line 1004: Technically this is a breaking change, so it should be reflected in the commit message (`!`).

evanlinjin (changes requested):
Quality work.

Just a couple requested changes:

* Commit messages should contain a scope (as mentioned in the conventional commits spec). The scope should be the name of the package affected: `core`, `chain`, `example`, etc.
* Commits that break the API should be marked with `!`.

And a nit:

* Error messages should be somewhat consistent. I personally don't think we should suggest actions in the message, just state what the error is.

evanlinjin:
@Dmenec are you able to make progress on this?

Dmenec:
@evanlinjin Pushed the updates. Let me know if everything checks out :)

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -973,33 +973,47 @@ pub enum InsertDescriptorError<K> {
     },
 }
 
-impl<K: core::fmt::Debug> core::fmt::Display for InsertDescriptorError<K> {
+/// Returns a shortened string representation of a descriptor.
+/// Shows the first and last `edge_len` characters, separated by `...`.
+fn short_descriptor(desc: &Descriptor<DescriptorPublicKey>, edge_len: usize) -> String {
+    let s = ToString::to_string(desc);
+    if s.len() <= edge_len * 2 {
```
evanlinjin: Nit: `edge_len * 2 + 3`.

No point returning the shortened version if it's longer than the original (due to `...`).

Have a code comment for the rationale of `+3`.

evanlinjin (approved):
utACK 532e2ecf039f7f3ab1b17e76620137cb955c2704

I'm happy with how this is looking. Unable to test until I have access to a computer again.

Dmenec:
Having some trouble with the no_std CI check for the esplora crate... Does it actually support no_std, or should be treated as std‑only?

oleonardolima:
> Having some trouble with the no_std CI check for the esplora crate... Does it actually support no_std, or should be treated as std‑only?

@Dmenec It's now fixed on master, a rebase should fix it.

crates/chain/src/tx_graph.rs
```diff
@@ -256,10 +256,28 @@ pub enum CalculateFeeError {
...
+                    write!(f, ": ")?;
+                    for (i, op) in shown.enumerate() {
+                        if i > 0 {
+                            write!(f, ", ")?;
+                        }
+                        write!(f, "{}", op)?;
+                    }
+                    if remaining > 0 {
+                        write!(f, " (+{} more)", remaining)?;
+                    }
```
nymius, line 274: Nit:

```diff
diff --git a/crates/chain/src/tx_graph.rs b/crates/chain/src/tx_graph.rs
index 457a478a..d170eccf 100644
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -258,19 +258,16 @@ impl fmt::Display for CalculateFeeError {
         match self {
             CalculateFeeError::MissingTxOut(outpoints) => {
                 let max_show = 3;
-                let shown = outpoints.iter().take(max_show);
+                let shown: Vec<_> = outpoints.iter().take(max_show).collect();
                 let remaining = outpoints.len().saturating_sub(max_show);
 
                 write!(f, "missing `TxOut` for input(s)")?;
                 if outpoints.is_empty() {
                     write!(f, ": <none>")
                 } else {
-                    write!(f, ": ")?;
-                    for (i, op) in shown.enumerate() {
-                        if i > 0 {
-                            write!(f, ", ")?;
-                        }
-                        write!(f, "{}", op)?;
+                    write!(f, ": {}", shown[0])?;
+                    for op in &shown[1..] {
+                        write!(f, ", {}", op)?;
                     }
                     if remaining > 0 {
                         write!(f, " (+{} more)", remaining)?;
```

crates/file_store/src/lib.rs
```diff
@@ -26,11 +26,18 @@ pub enum StoreError {
...
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         match self {
-            Self::Io(e) => write!(f, "io error trying to read file: {e}"),
-            Self::InvalidMagicBytes { got, expected } => write!(
-                f,
-                "file has invalid magic bytes: expected={expected:?} got={got:?}",
-            ),
+            Self::Io(e) => write!(f, "io error trying to read file: {}", e),
+            Self::InvalidMagicBytes { got, expected } => {
+                write!(f, "file has invalid magic bytes: expected=0x")?;
```
nymius:
```bash
Error: could not open file store

Caused by:
    file has invalid magic bytes: expected=0x62646b5f6578616d706c655f656c65637472756d got=0x62646b5f6578616d706c5f656c65637472756d01
```
Given the format of the error output I would use `invalid magic bytes` without the `file has`.

This is a good opportunity to rewrite the others in the same spirit.

Also, stacking `expected` and `got` improves the feedback to the reader:
```bash
Error: could not open file store

Caused by:
    invalid magic bytes
expected: 0x62646b5f6578616d706c655f656c65637472756d
got:      0x62646b5f6578616d706c5f656c65637472756d01
```

examples/example_esplora/src/main.rs
```diff
@@ -215,8 +215,20 @@ fn main() -> anyhow::Result<()> {
                     .chain_tip(local_tip.clone())
                     .inspect(|item, progress| {
                         let pc = (100 * progress.consumed()) as f32 / progress.total() as f32;
-                        eprintln!("[ SCANNING {pc:03.0}% ] {item}");
-                        // Flush early to ensure we print at every iteration.
+                        match item {
+                            bdk_chain::spk_client::SyncItem::Spk((keychain, index), spk) => {
+                                eprintln!(
+                                    "[ SCANNING {pc:03.0}% ] script {} {} {}",
```
nymius: nit: `"[ SCANNING {pc:3.0}% ] script {} {} {}"`, instead of `"[ SCANNING {pc:03.0}% ] script {} {} {}",`

I prefer
```
[ SCANNING   6% ] outpoint 623cf110e23b15074e807a551a751856eb0a95f23a4a60f125d943d1d6e7cb6a:0
```
to
```
[ SCANNING 006% ] outpoint 623cf110e23b15074e807a551a751856eb0a95f23a4a60f125d943d1d6e7cb6a:0
```

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -973,33 +973,50 @@ pub enum InsertDescriptorError<K> {
     },
 }
 
-impl<K: core::fmt::Debug> core::fmt::Display for InsertDescriptorError<K> {
+/// Returns a shortened string representation of a descriptor.
+/// Shows the first and last `edge_len` characters, separated by `...`.
+fn short_descriptor(desc: &Descriptor<DescriptorPublicKey>, edge_len: usize) -> String {
```
nymius: I've tried this with the following `bin`:
```rust
use bdk_chain::bitcoin::key::Secp256k1;
use bdk_chain::indexer::keychain_txout::KeychainTxOutIndex;
use bdk_chain::miniscript::Descriptor;

const DESC: &str = "...";

fn main() {
    let lookahead = 10;
    let use_cache = true;
    let mut index = KeychainTxOutIndex::new(lookahead, use_cache);
    let secp = Secp256k1::new();
    let (desc, _) = Descriptor::parse_descriptor(&secp, DESC).unwrap();
    index.insert_descriptor("some", desc.clone()).unwrap();
    if let Err(e) = index.insert_descriptor("other", desc) {
        eprintln!("{}", e);
    }
}
```

I think the output is not very clear:

```bash
descriptor 'tr([...d45f9rlk' is already in use by another keychain 'some'.
```

I would keep only a fixed size prefix in the output, including the descriptor type and trying to capture the origin keypath if possible.
@qatkk proposed to increase the suffix to capture the checksum, but without any clear accessors to those components it is a "best guess" effort.
Maybe we should propose the implementation of them at the `miniscript` level, and add methods to get strings from `DescriptorType` for example.

crates/chain/src/tx_graph.rs
Dmenec: Maybe shortening the displayed txid could be an option as well? e.g. abcd1234…ef901234

crates/file_store/src/lib.rs
Dmenec: Sure! I'll update the ones that are in scope for this PR and push the proposed changes shortly.

crates/chain/src/indexer/keychain_txout.rs
Dmenec: cACK — I’ll open a follow‑up PR with the proposal and then apply it here. Thanks for the suggestion!

crates/chain/src/tx_graph.rs
evanlinjin: Nack on shortening txid 😅 personally I prefer seeing the whole thing

crates/chain/src/indexer/keychain_txout.rs
evanlinjin: @nymius that is a good point. For practicality, can we remove the `short_descriptor` concept in this PR and just print the full descriptor for now?

Let's add it back once it is more readable (including the descriptor type, key path, checksum, etc.)

nymius: I agree, let's use full descriptors for now.

crates/chain/src/tx_graph.rs
```diff
@@ -256,15 +256,29 @@ pub enum CalculateFeeError {
...
+                let remaining = outpoints.len().saturating_sub(max_show);
+
+                writeln!(f, "   missing `TxOut` for input(s)")?;
+                if outpoints.is_empty() {
+                    write!(f, "<none>")
+                } else {
+                    writeln!(f, "{}", shown[0])?;
+                    for op in &shown[1..] {
+                        writeln!(f, "{}", op)?;
+                    }
```
nymius: I haven't executed this yet, but my guess is it will print the outpoint vertically, as you are using `writeln!` instead of `write!`.

crates/chain/src/tx_graph.rs
Dmenec: Resolved in 737cff8db54031db379c583f8869b61170b268ec

examples/example_esplora/src/main.rs
Dmenec: Resolved in e46b1c9640ec8ebbb0200e16084c3e31d1824fb0

crates/file_store/src/lib.rs
```diff
@@ -25,13 +25,26 @@ pub enum StoreError {
...
-            ),
-            Self::Bincode(e) => write!(f, "bincode error while reading entry {e}"),
+            Self::Io(e) => write!(f, "io error while reading store file: {e}"),
+            Self::Bincode(e) => write!(f, "bincode error while decoding entry {e}"),
+            Self::InvalidMagicBytes { got, expected } => {
+                writeln!(f, "invalid magic bytes")?;
+                write!(f, "  expected: 0x")?;
+                fmt_hex_bytes(f, expected)?;
+                writeln!(f)?;                
+                write!(f, "  got:      0x")?;
```
nymius: Maybe the best for this is to reserve the same number of spaces for the word in the `fmt` placeholder, i.e., `writeln!("{:10}", "expected:")`. Seems a bit clunky, but is much more future proof than guessing the number of spaces. I have tried it with the following code:
```rust
fn main() {
    let hex = "0x913414332";
    let other_hex = "0x3941247913414332";
    println!("{:10}{hex}", "expected:");
    println!("{:10}{other_hex}", "got:");
}
```

examples/example_electrum/src/main.rs
```diff
@@ -210,9 +210,22 @@ fn main() -> anyhow::Result<()> {
...
+                                eprintln!(
+                                    "[ SCANNING {pc:03.0}% ] script {} {} {}",
+                                    keychain, index, spk
+                                );
+                            }
+                            bdk_chain::spk_client::SyncItem::Txid(txid) => {
+                                eprintln!("[ SCANNING {pc:03.0}% ] txid {}", txid);
+                            }
+                            bdk_chain::spk_client::SyncItem::OutPoint(op) => {
+                                eprintln!("[ SCANNING {pc:03.0}% ] outpoint {}", op);
```
nymius: I would apply the same criteria for `esplora` here.

crates/chain/src/tx_graph.rs
Dmenec: Yes, thought maybe it would be a bit more readable:

```
   missing `TxOut` for input(s)
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa:0
bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb:1
cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc:2
```

crates/file_store/src/lib.rs
Dmenec: Resolved in ae17df23280728923c6f062aa8fda3aa462766eb

examples/example_electrum/src/main.rs
Dmenec: Resolved in adb5b9954267de623c111cbd103523d7e6b07487

crates/chain/src/tx_graph.rs
evanlinjin: I'd suggest keeping `Display` single-line to follow Rust conventions. Callers typically don't expect `{}` to emit multiple lines, and it composes better with logging/error chains.

If we really want a pretty-printed multi-line version later, we can use the alternate flag (`{:#}`) for that. I don't think it's necessary for this PR (but feel free to include it).

```rust
       if f.alternate() {
           // multi-line version
           todo!()
       } else {
           // single-line version
           todo!()
       }
   }
```

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -973,33 +973,35 @@ pub enum InsertDescriptorError<K> {
...
             }
             InsertDescriptorError::KeychainAlreadyAssigned {
-                existing_assignment: existing,
+                existing_assignment,
                 keychain,
             } => {
                 write!(
                     f,
-                    "attempt to re-assign keychain {keychain:?} already assigned to {existing:?}"
+                    "keychain '{}' is already associated with another descriptor '{}'.",
```
evanlinjin: As per [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html#error-types-are-meaningful-and-well-behaved-c-good-err):

> The error message given by the Display representation of an error type should be lowercase without trailing punctuation, and typically concise.

Please remove the trailing punctuation thanks!

crates/file_store/src/lib.rs
```diff
@@ -25,13 +25,26 @@ pub enum StoreError {
...
-            Self::Io(e) => write!(f, "io error trying to read file: {e}"),
-            Self::InvalidMagicBytes { got, expected } => write!(
-                f,
-                "file has invalid magic bytes: expected={expected:?} got={got:?}",
-            ),
-            Self::Bincode(e) => write!(f, "bincode error while reading entry {e}"),
+            Self::Io(e) => write!(f, "io error while reading store file: {e}"),
+            Self::Bincode(e) => write!(f, "bincode error while decoding entry {e}"),
+            Self::InvalidMagicBytes { got, expected } => {
+                writeln!(f, "invalid magic bytes")?;
```
evanlinjin: Also here, please make this single line thanks!

evanlinjin (changes requested):
Sorry for the bike-shedding. We're very close.

crates/file_store/src/lib.rs
Dmenec: Just to clarify the final intended format:

`invalid magic bytes: expected 0x01020304, got 0x10203040`

and the one mentioned in https://github.com/bitcoindevkit/bdk/pull/2083#discussion_r2734811270

```
missing `TxOut` for input(s): aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa:0, ...
```

Does this look okay? Thank you!

crates/file_store/src/lib.rs
Dmenec: Discussing **[here](https://github.com/bitcoindevkit/bdk/pull/2083#discussion_r2734829962)**

crates/file_store/src/lib.rs
evanlinjin: > ```shell
> Error: could not open file store
> 
> Caused by:
>     file has invalid magic bytes: expected=0x62646b5f6578616d706c655f656c65637472756d got=0x62646b5f6578616d706c5f656c65637472756d01
> ```
> 
> Given the format of the error output I would use `invalid magic bytes` without the `file has`.
> 
> This is a good opportunity to rewrite the others in the same spirit.
> 
> Also, stacking `expected` and `got` improves the feedback to the reader:
> 
> ```shell
> Error: could not open file store
> 
> Caused by:
>     invalid magic bytes
> expected: 0x62646b5f6578616d706c655f656c65637472756d
> got:      0x62646b5f6578616d706c5f656c65637472756d01
> ```

Sorry for the back-and-forth, but my [comment](https://github.com/bitcoindevkit/bdk/pull/2083#discussion_r2734811270) invalidates this. Error messages should be single line as per convention! Can use the `#` for "pretty print".

crates/file_store/src/lib.rs
evanlinjin: @Dmenec looks good thanks.

**Nit:** Message for the second one can be simplified to `cannot calculate fee, missing output(s):`. 😅 Messages for `CalculateFeeError` should mention that it's a calculate fee error?

Dmenec: > **Nit:** Message for the second one can be simplified to `cannot calculate fee, missing output(s):`. 😅 Messages for `CalculateFeeError` should mention that it's a calculate fee error?

What do you think about `cannot calculate fee, missing previous output(s)`? Since we're referring to the outpoints of the inputs, I think it's clearer. Also agree on mentioning that it's a calculate fee error :)

crates/chain/src/tx_graph.rs
oleonardolima: I don't think we need the multi-line now, can add in the future if needed. Also, what's the rationale for the leading whitespace ?

oleonardolima (commented):
Overall it looks good, it just need to address Evan's comments.

crates/file_store/src/lib.rs
Dmenec: Resolved in 744a3de1bc2f9fc77203e1e89a5e95de5034948a

crates/chain/src/indexer/keychain_txout.rs
Dmenec: Resolved in 5d715c2e4c8076be058e342efe279a3fc3481312

crates/chain/src/tx_graph.rs
Dmenec: > I don't think we need the multi-line now, can add in the future if needed. Also, what's the rationale for the leading whitespace ?

Was a minor style thing, wasn't really sure about it, already removed it :)

Dmenec: Resolved in ad96bd9244fd58d34b3bb901285751e2c92a9edd

crates/chain/src/indexer/keychain_txout.rs
Dmenec: Resolved in 5d715c2e4c8076be058e342efe279a3fc3481312

crates/file_store/src/lib.rs
Dmenec: > Also here, please make this single line thanks!

Resolved in 823e4e980af03672cc6342d0e4850febcbd14c9d

> > **Nit:** Message for the second one can be simplified to `cannot calculate fee, missing output(s):`. 😅 Messages for `CalculateFeeError` should mention that it's a calculate fee error?
> 
> What do you think about `cannot calculate fee, missing previous output(s)`? Since we're referring to the outpoints of the inputs, I think it's clearer. Also agree on mentioning that it's a calculate fee error :)

Resolved in 5a85f6e25a9e14204cb66fec3f7381a59133f0be

crates/chain/src/tx_graph.rs
```diff
@@ -256,15 +256,33 @@ pub enum CalculateFeeError {
...
-                f,
-                "missing `TxOut` for one or more of the inputs of the tx: {outpoints:?}",
-            ),
-            CalculateFeeError::NegativeFee(fee) => write!(
-                f,
-                "transaction is invalid according to the graph and has negative fee: {}",
-                fee.display_dynamic()
-            ),
+            CalculateFeeError::MissingTxOut(outpoints) => {
+                let max_show = 3;
```
evanlinjin, line 260: If I were to nit-pick, I would say get rid of this and just show all prev outputs.

evanlinjin (approved):
ACK 823e4e980af03672cc6342d0e4850febcbd14c9d

I'm going to merge this as is. My nit-pick can be addressed in follow ups.

crates/chain/src/tx_graph.rs
```diff
@@ -256,15 +256,33 @@ pub enum CalculateFeeError {
...
+                let shown: Vec<_> = outpoints.iter().take(max_show).collect();
+                let remaining = outpoints.len().saturating_sub(max_show);
+
+                write!(f, "cannot calculate fee, missing previous output(s): ")?;
+                if outpoints.is_empty() {
+                    write!(f, "<none>")
```
evanlinjin, line 266: If this ever happens, it's a bug in the library - let's `debug_assert!` it?

PR #2090, opened 2026-01-07
Add to rusqlite module the `get_pre_1_wallet_keychains` migration helper

notmandatory:
### Description

Add to rusqlite module the `get_pre_1_wallet_keychains` migration to help migrate users from a pre-1.0 bdk sqlite database. This new function returns the last revealed index and checksum value for each keychain it finds. 

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/rusqlite_impl.rs
+++ b/crates/chain/src/rusqlite_impl.rs
@@ -12,10 +12,10 @@ use alloc::{
 };
 use bitcoin::consensus::{Decodable, Encodable};
 use rusqlite;
-use rusqlite::named_params;
 use rusqlite::types::{FromSql, FromSqlError, FromSqlResult, ToSql, ToSqlOutput, ValueRef};
 use rusqlite::OptionalExtension;
 use rusqlite::Transaction;
+use rusqlite::{named_params, Connection};
 
 /// Table name for schemas.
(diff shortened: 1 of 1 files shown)


thunderbiscuit (commented):
Approach ACK. I don't have an opinion as to where it should live per se, although this is mostly related to a pre-1.0 bdk `Wallet` database, so maybe in bdk_wallet would be closer to its target users? Works either way!

I tested with a local database and I get correct values. Only small comment would be that the string it pulls for the KeychainKind has escaped double quote characters and those are probably not really wanted (see example below).

```shell
Pre1WalletKeychain { keychain: "\"External\"", last_derivation_index: 3, checksum: [108, 55, 54, 100, 117, 121, 118, 55] }
```

notmandatory:
ah good catch, I didn't notice that with the double quotes, will fix.

oleonardolima (commented):
Concept ACK fde4e8745a448a9c0f01e0368915673d3b4cb6d1

Are you looking forward to keep these changes here or move it to `bdk_wallet` ? I think it'd better live in `bdk_wallet`.

notmandatory:
Thanks for the reminder, I'm closing this one, replaced by bitcoindevkit/bdk_wallet#364 and bitcoindevkit/bdk_wallet#365.

PR #2091, opened 2026-01-07
feat(chain)!: Add `prev_blockhash` to `ToBlockHash` trait

### LINKED ISSUE #2021: Ensure `CheckPoint` chain methods validate and link via previous blockhash

### LINKED ISSUE #2095: When does `LocalChain::apply_update` invalidate checkpoints?
**Describe the bug**  

### 1. Genesis block replacement

The documentation for `merge_chains` says that a `CannotConnectError` occurs when the update attempts to replace the genesis block of the original chain. However, as demonstrated by the test case "fix blockhash before agreement point" in `update_local_chain`, the implementation does allow the genesis hash to be changed if the update provides a new hash at height 0 and the rest of the update is consistent. The function doesn't prevent replacing the genesis block - it only requires that the update is unambiguous and properly invalidates any conflicting blocks above the point of agreement.

For users this mismatch between documentation and implementation can lead to confusion. If a user expects that the genesis block can never be changed as the docs suggest, they may be surprised to find that it can be "fixed" by an update. This could affect how they reason about the behavior of `LocalChain` as it relates to invalidation heights and chain reorgs.

For correctness, allowing the genesis block to be replaced may be intentional, but the documentation should clearly state this. If it is not intentional, the implementation should be fixed to match the documentation.

**To Reproduce**  

Original `[0, A]`
Update `[0', A]`
Result `[0', A]` (Ok)

**Expected behavior**  

Result `CannotConnectError { try_include_height: 0 }` (Err)

### 2. Chain reorganization

The current behavior of `merge_chains` does not allow a basic reorg by switching forks unless the update explicitly includes a point of agreement. If we continue traversing blocks below the last update height, the function cannot unambiguously determine how to connect the update to the original chain, and thus returns a `CannotConnectError`.

However if the update at the invalidation height claims to connect to a block in the original chain, or otherwise doesn't conflict with it, then presumably it ought to connect unambiguously, allowing a reorganization to occur by switching to the longest chain.

**To Reproduce**  

Original `[0, A]`
Update `[A', B']`
Result `CannotConnectError { try_include_height: 0 }` (Err)

**Expected behavior**  

Result `[0, A', B']` (Ok)

**Additional context**  

- Documentation of [`merge_chains`](https://github.com/bitcoindevkit/bdk/blob/a161ee24d38b038a51161816a9f90a87f90d157d/crates/chain/src/local_chain.rs#L592)

- The unfortunate restriction with `LocalChain` https://github.com/bitcoindevkit/bdk/pull/2000#discussion_r2265233190

**Build environment**  
- BDK tag/commit: a161ee24d38b038a51161816a9f90a87f90d157d 
- OS+version: macOS 15.7.2 (Sequoia)  
- Rust/Cargo version: 1.92.0 
- Rust/Cargo target: aarch64-apple-darwin  

- [ ] Electrum
- [ ] Esplora
- [ ] Bitcoin Core RPC

- [ ] Yes
- [x] No

ValuedMammal:
This PR adds `prev_blockhash` method to `ToBlockHash` trait.

`prev_blockhash` returns the hash of the previous block data, if it is known, and `None` otherwise. Implemented `prev_blockhash` for `Header`, which returns the header's previous block hash.

We now require checkpoints to link via their `prev_blockhash` relation.

Fixed `CheckPoint::push` to now check that if the height being pushed directly follows the current tip height, the previous hash of `data` is the same as the current tip hash, (provided that the data's `prev_blockhash` is known). This is necessary to prevent inconsistent or invalid chains from being constructed.

Fixed `insert` to displace conflicting blocks by reverting to a previous base if it's found to conflict with the inserted data. Also avoid a panic in `insert` in case `push` returns an error by returning the last successfully updated tip.

In `apply_changeset{_to checkpoint}` we now propagate the error from `extend`. Now that `apply_changeset` can fail due to inconsistent data in addition to `MissingGenesisError`, we need to propagate the error rather than expect that the chain update can always succeed.

fix #2095 
fix #2021 

#### Changed

- Changed `LocalChain::from_blocks` to return `Result<Self, CannotConnectError>`
- Changed `LocalChain::from_changeset` to return `Result<Self, CannotConnectError)`
- Changed `LocalChain::apply_changeset` to return `Result<(), CannotConnectError>`

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -13,11 +13,22 @@ pub use bdk_core::{CheckPoint, CheckPointIter};
 use bitcoin::block::Header;
 use bitcoin::BlockHash;
 
+/// Error for `apply_changeset_to_checkpoint`.
+#[derive(Debug)]
+struct ApplyChangeSetError<D> {
+    cp: CheckPoint<D>,
+}
+
 /// Apply `changeset` to the checkpoint.
+///
+/// # Errors
+///
+/// - If constructing the new chain from the provided `changeset` fails, then a
+///   [`ApplyChangeSetError`] is returned.
 fn apply_changeset_to_checkpoint<D>(
     mut init_cp: CheckPoint<D>,
     changeset: &ChangeSet<D>,
-) -> Result<CheckPoint<D>, MissingGenesisError>
+) -> Result<CheckPoint<D>, ApplyChangeSetError<D>>
 where
     D: ToBlockHash + fmt::Debug + Clone,
 {
@@ -50,8 +61,12 @@ where
         let new_tip = match base {
             Some(base) => base
                 .extend(extension)
-                .expect("extension is strictly greater than base"),
-            None => LocalChain::from_blocks(extension)?.tip(),
+                .map_err(|cp| ApplyChangeSetError { cp })?,
+            None => match CheckPoint::from_blocks(extension) {
+                Ok(cp) => cp,
+                Err(None) => init_cp,
+                Err(Some(cp)) => return Err(ApplyChangeSetError { cp }),
+            },
         };
         init_cp = new_tip;
     }
@@ -255,26 +270,34 @@ where
     ///
     /// The [`BTreeMap`] enforces the height order. However, the caller must ensure the blocks are
     /// all of the same chain.
-    pub fn from_blocks(blocks: BTreeMap<u32, D>) -> Result<Self, MissingGenesisError> {
-        let genesis_hash = blocks
-            .get(&0)
-            .map(ToBlockHash::to_blockhash)
-            .ok_or(MissingGenesisError)?;
-
-        Ok(Self {
-            genesis_hash,
-            tip: CheckPoint::from_blocks(blocks).expect("blocks must be in order"),
-        })
+    ///
+    /// If `blocks` doesn't contain a value at height `0`, a.k.a. "genesis" block, or contains
+    /// inconsistent or invalid data, then returns a [`CannotConnectError`]. See also
+    /// [`LocalChain::from_changeset`].
+    pub fn from_blocks(blocks: BTreeMap<u32, D>) -> Result<Self, CannotConnectError> {
+        let changeset = ChangeSet {
+            blocks: blocks
+                .into_iter()
+                .map(|(height, data)| (height, Some(data)))
+                .collect(),
+        };
+        Self::from_changeset(changeset)
     }
 
     /// Construct a [`LocalChain`] from an initial `changeset`.
-    pub fn from_changeset(changeset: ChangeSet<D>) -> Result<Self, MissingGenesisError> {
-        let genesis_entry = changeset.blocks.get(&0).cloned().flatten();
-        let genesis_data = match genesis_entry {
+    ///
+    /// If `blocks` doesn't contain a value at height `0`, a.k.a. "genesis" block, or contains
+    /// inconsistent or invalid data, then returns a [`CannotConnectError`].
+    /// Otherwise returns a new [`LocalChain`] with the changeset applied.
+    pub fn from_changeset(changeset: ChangeSet<D>) -> Result<Self, CannotConnectError> {
+        let genesis_data = match changeset.blocks.get(&0).cloned().flatten() {
             Some(data) => data,
-            None => return Err(MissingGenesisError),
+            None => {
+                return Err(CannotConnectError {
+                    try_include_height: 0,
+                })
+            }
         };
-
         let (mut chain, _) = Self::from_genesis(genesis_data);
         chain.apply_changeset(&changeset)?;
         debug_assert!(chain._check_changeset_is_applied(&changeset));
@@ -317,9 +340,12 @@ where
     }
 
     /// Apply the given `changeset`.
-    pub fn apply_changeset(&mut self, changeset: &ChangeSet<D>) -> Result<(), MissingGenesisError> {
+    pub fn apply_changeset(&mut self, changeset: &ChangeSet<D>) -> Result<(), CannotConnectError> {
         let old_tip = self.tip.clone();
-        let new_tip = apply_changeset_to_checkpoint(old_tip, changeset)?;
+        let new_tip =
+            apply_changeset_to_checkpoint(old_tip, changeset).map_err(|e| CannotConnectError {
+                try_include_height: e.cp.height(),
+            })?;
         self.tip = new_tip;
         debug_assert!(self._check_changeset_is_applied(changeset));
         Ok(())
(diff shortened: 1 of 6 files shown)


evanlinjin:
How does this relate to PR #2024? That PR addresses the same issue and handles validation in `insert()` and `merge_chains` as well. As written, this only validates `push()` with contiguous heights - `insert()` and `merge_chains` can still create inconsistent chains. Is this intended as a simpler alternative approach, or a foundation for the work in #2024?

I'm also wondering if this is meant to show that we can achieve simpler validation by not introducing optional data fields. However, there is a gap: without tracking "ghost checkpoints" (positions inferred from `prev_blockhash` but not yet populated), we can't properly validate in `insert()` or `merge_chains()`.

For example, if I insert a checkpoint at height 100 with `prev_blockhash = X`, and later insert at height 99 with `blockhash = Y` where `X != Y`, we need to detect that conflict. Either:

* We materialize the ghost as a placeholder node (data: Option<D>) so the conflict is caught at insertion (#2024 ), or
* We use a custom iterator that yields ghost checkpoints during traversal/merge, validating at that boundary (#2017 )

The current PR doesn't address either path, so I don't think it fully resolves #2021. Is there a different approach you have in mind for handling these cases?

ValuedMammal:
Here we don't intend to make the `CheckPoint::data` field optional, instead displacing a block will just remove it from the chain.

It's missing coverage for when `insert` properly displaces a conflicting block at the previous height.

The previous `BlockId` inferred by the current chain could in theory be implemented directly on the `CheckPoint` like this. I haven't tackled the issues with `merge_chains` yet.

```rust
/// Get the previous [`BlockId`] of this checkpoint.
///
/// I.e. the height and hash of the block immediately preceding this one by consensus,
/// if it can be inferred from the `prev_blockhash` of the inner block data.
pub fn prev_block_id(&self) -> Option<BlockId> {
    let prev_height = self.height().checked_sub(1)?;
    let prev_hash = self.0.data.prev_blockhash()?;
    Some(BlockId {
        height: prev_height,
        hash: prev_hash,
    })
}
```

crates/core/src/checkpoint.rs
```diff
@@ -233,59 +270,103 @@ where
...
+                // Same hash, no change
                 if cp.hash() == data.to_blockhash() {
                     return self;
                 }
                 assert_ne!(cp.height(), 0, "cannot replace genesis block");
-                // If we have a conflict we just return the inserted data because the tail is by
-                // implication invalid.
-                tail = vec![];
-                break cp.prev().expect("can't be called on genesis block");
+                // We're replacing a block, so the tail is no longer valid.
```
evanlinjin: We should also displace the tail if:

* `data`'s `prev_blockhash` conflicts with the block directly below. The block directly below should also be displaced.
* `data`'s `blockhash` conflicts with the the `prev_blockhash` of the block directly above.

To achieve this, I recommend using `CheckPointEntry` instead of `CheckPoint` and add a method such as `CheckPointEntry::prev()` - this requires the `CheckPointEntry::BlockId` variant to have a `CheckPoint` field.

#### Naming suggestion

```rust
   /// An entry yielded by [`CheckPointEntryIter`].
   #[derive(Debug, Clone)]
   pub enum CheckPointEntry<D> {
       /// A placeholder entry: there is no checkpoint stored st this height,
       /// but the checkpoint one height above links back here via it's `prev_blockhash`.
       Placeholder {
           /// The block ID at *this* height.
           block_id: BlockId,
           /// The checkpoint one height *above* that links back to `block_id`.
           checkpoint_above: CheckPoint<D>,
       },
       /// A real checkpoint recorded at this height.
       Occupied(CheckPoint<D>),
   }
````

File: crates/chain/src/local_chain.rs [hunk 2]
```diff
@@ -576,148 +594,138 @@ impl core::fmt::Display for ApplyHeaderError {
...
+    ///
+    /// [`CannotConnectError`] occurs when the `original_tip` and `update_tip` chains are disjoint:
+    ///
+    /// - If no point of agreement is found between the update and original chains, and no explicit
+    ///   invalidation occurred.
+    /// - A point of agreement is found but the update is ambiguous above the point of agreement
+    ///   (a.k.a. the update and original chain both have a block above the point of agreement, but
+    ///   their heights do not overlap).
+    /// - The update attempts to replace the genesis block of the original chain.
+    fn merge_chains(
```
evanlinjin, line 632: The refactored `merge_chains` does not do anything meaningfully different. It also does not make it easier to read.

The important change to make is to make sure `merge_chains` takes `prev_blockhash`es into consideration. The refactoring does not do that.

### Suggested changes

* Iterate over `CheckPointEntry` instead of `CheckPoint`. Rationale: `prev_blockhash`es are also block hashes so they can be points of agreements/invalidation.
* Only insert `Some(data)` into `ChangeSet` if there is an actual `CheckPoint` in the update chain at that height.
* Only insert `None` into `ChangeSet` if there is an actual `CheckPoint` in the original chain at that height.

### Suggested tests

These tests uses a data type that returns `Some` for `prev_blockhash`:

##### `prev_blockhash` invalidates

* Original: `1:A`, `2:B`, `4:D`
* Update: `1:A`, `3:C' (prev=2:B')`
* Result: `1:A`, `3:C' (prev=2:B')`

#### Connects due to `prev_blockhash`

* Original: `2:B`, `3:C`
* Update: `2:B`, `4:D (prev=3:C)`
* Result: `2:B`, `3:C`, `4:D`

evanlinjin (changes requested):
Since you're taking over ticket #2021, feel free to checkout my unfinished work for inspiration: https://github.com/bitcoindevkit/bdk/compare/cp_entry.

evanlinjin:
NACK

@ValuedMammal As I mentioned [here](https://github.com/bitcoindevkit/bdk/pull/2091#discussion_r2690022372), the refactoring of `merge_chains` is a big distraction from the main goal.

1. The PR's purpose is to make checkpoint construction `prev_blockhash`-aware. The `merge_chains` refactor doesn't do this at all - it's still comparing hashes at matching heights only.

2. Your commit claims it fixes ***"faulty behavior preventing the ability to switch forks in a basic reorg scenario."*** and introduces a test that is unsound:
    ```
    // "invalidate tip and extend chain"
    chain:  0:_, 1:A
    update: 1:A', 2:B'
    ```
    The update doesn't include genesis. How do we know `A'` and `B'` are even on the same blockchain as `_` and `A`? They could be from a completely different chain that happens to have blocks at heights 1 and 2. The old `merge_chains` would reject this because there's no point of agreement - which is actually correct behavior. You shouldn't be able to replace your chain's tip with blocks from an unverified source.

3. Having something like `CheckPointEntry` and `CheckPointIter` is genuinely useful. For example, when we canonicalize, if a tx is anchored to `A:3` (prev of `B:4`), but we only have `B:4` in the chain, that tx should still be considered in the best chain. If we use `CheckPointIter` to implement `merge_chains` (as I suggested), the logic will naturally become `prev_blockhash`-aware.

I would like to have #2021 done soon - so I will take over this ticket.

PR #2059, opened 2025-10-08
Add ElectrumSync builder for one-liner wallet synchronization

Her-Code:
## Summary

This PR addresses issue #1974.
It introduces a new fluent builder API for synchronizing wallets with Electrum servers, making wallet synchronization as simple as a one-liner call. The `ElectrumSync` builder provides an intuitive, chainable interface for configuring and executing wallet sync operations, significantly improving developer experience and reducing boilerplate code for common synchronization tasks.
## Features

- `Fluent Builder API `: Implement fluent API for `ElectrumSync` configuration
- `Dual Sync Modes` : Fast sync (revealed addresses only) and full scan (until stop gap)
- `Configurable Parameters `: Server URL, stop gap, batch size, and transaction output fetching
- `Type Safety` : Add `ElectrumSyncResult<K>` type alias.
- Add new `bdk_builder` module with sync functionality

PR #2094, opened 2026-01-15
feat: add one-liner wallet sync API with ElectrumSync

### LINKED ISSUE #1974: Provide a one liner to sync with electrum
**Describe the enhancement**  

Currently there is a lot of boilerplate needed to sync a wallet with electrum, it would be great to have one function that lets you pass in a mutable reference to a wallet and then have it update its internals with block data.

It used to be much easier to sync with electrum in the prior versions of bdk, it would be great to have a return to that simplicity.

**Use case**  

load a wallet, then sync it with electrum to get block data with a single line of code

**Additional context**

I personally think a builder method and some macros/convinience functions built on top of that would work well, I usually just want a full scan no matter what but every once in a while I'll need a fast sync. 

```rust
  ElectrumBuilder::new(wallet)
          .url("tcp://electrum.blockstream.info:50001")
          .network(NETWORK)
          .stop_gap(25)
          .batch_size(30)
          .fast_sync() 
          .fetch_prev_txouts()
          .request()?;
```

I'll bring this up in the Thursday meeting with @evanlinjin since I know he is working with electrum.

ShigrafS:
Closes #1974 

### Description

This PR introduces a small, focused helper abstraction to reduce the boilerplate required to synchronize a wallet index with an Electrum server, addressing the usability concerns raised in #1974.

The main addition is the `ElectrumSync` helper and `SyncOptions`, which wrap existing `SyncRequest` and `FullScanRequest` flows into a single, ergonomic call while preserving the existing `BdkElectrumClient` cache and behavior.

Concretely, this PR:

* Adds `ElectrumSync`, a lightweight helper for performing Electrum syncs using a shared `BdkElectrumClient`
* Introduces `SyncOptions` to configure fast sync vs full scan, stop gap, batch size, and prevout fetching
* Exports the helper from `bdk_electrum` for reuse
* Adds an example demonstrating a “one-liner” full scan and fast sync using public Electrum servers

The helper is intentionally minimal and builds directly on top of existing BDK primitives without introducing a new builder API or altering wallet abstractions.

---

### Notes to the reviewers

This PR is informed by the feedback and architectural concerns raised in #2059.

In particular:

* The helper preserves the `BdkElectrumClient` and its cache across sync calls, avoiding the performance regressions discussed previously.
* The API is intentionally small and avoids introducing a fluent/builder-style abstraction that would expand the public surface area.
* The helper does **not** mutate wallet state; instead, it returns the existing chain and transaction updates so that application code remains in control of how updates are applied.

The goal of this PR is to provide a **convenience layer** that demonstrates a simplified sync flow while remaining close to existing abstractions.
Feedback on whether this belongs as a public helper or should instead live purely as an example is very welcome.

---

### Changelog notice

* Added `ElectrumSync` and `SyncOptions` helpers to simplify Electrum wallet synchronization
* Added an example demonstrating one-liner Electrum sync usage

---

#### All Submissions:

* [x] I followed the [[contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)

#### New Features:

#### Bugfixes:

* [x] I'm linking the issue being fixed by this PR (#1974)


crates/electrum/Cargo.toml
```diff
@@ -16,10 +16,11 @@ workspace = true
 [dependencies]
 bdk_core = { path = "../core", version = "0.6.1" }
 electrum-client = { version = "0.24.0", features = [ "proxy" ], default-features = false }
+bdk_chain = { path = "../chain", version = "0.23.2" }
```
evanlinjin: As mentioned in the closed PR, we don't want a chain source crate to depend on `bdk_chain`.

ShigrafS: Yes, of course. 
This dependency on bdk_chain is temporary. 
I’m actively working on switching it out so the chain source crate remains independent, and I’ll push an update once that’s done.

ShigrafS:
The PR is ready for review. 
The tests are passing locally.

evanlinjin:
Where exactly is the "small, focused helper abstraction"? You changed nothing of relevance in the `bdk_electrum` crate.

There is a whole lot of unrelated changes - I'm assuming this is an attempt at making thing run/compile on windows? This does not belong in this PR - feel free to create a new PR for these changes.

ShigrafS:
Fixed it.

evanlinjin:
@ShigrafS the "one liner helper" is currently an example and not part of the library.

ShigrafS:
@evanlinjin I was thinking this fits right within the scope of the issue with respect to the last PR. 
If it makes sense to evolve it into a lightweight public API , should I proceed with that?

thunderbiscuit:
This PR doesn't add a 1-liner API to the bdk_electrum crate, nor does it make sense to add one IMO. I think the core issue is that we kept #1974 open when it should not be. Worse, we added a `good-first-issue` to it which draws in people when it's a fool's errand (sorry @ShigrafS! this is totally not your fault, I mean no offence at all). We're on the second PR on this, and honestly it just feels like it's wasting everyone's time.

My suggestion is to either:
1. Close #1974 (preferred).
2. If for whatever reason people feel strongly about this sort of thing which (in my opinion) belongs as code at the application-level not the library level, at most a single example could be added that would show how to do it. @ShigrafS your PR impacts 8 files when it should only create 1.

ShigrafS:
@evanlinjin @thunderbiscuit @luisschwab 

Thanks for the feedback so far. Based on @evanlinjin’s comment [[here](https://github.com/bitcoindevkit/bdk/pull/2059#pullrequestreview-3347616682)]) and the discussion in this thread, I’ll be **reducing the scope of this PR to an example-only implementation** demonstrating a one-liner Electrum wallet sync. This keeps it safe, minimal, and in line with current BDK library constraints.

Given that this example fully demonstrates the intended functionality from #1974, I’d suggest we **close the original issue**, as adding it to the library API is beyond the scope of the repo at present.

I’ll update the PR accordingly to reflect the example-only version.

Thanks!

evanlinjin:
NACK

Thanks for trying to contribute, but I still don't see the value in this.

ShigrafS:
@evanlinjin 
Thanks for the feedback and for taking the time to review this.

Given the consensus that this doesn’t belong in the BDK repo, I’ll close the PR.
Appreciate the guidance.

PR #2099, opened 2026-01-21
Fix no-std compilation

evanlinjin:
### Description

`serde_json` requires either `std` or `alloc` feature enabled. We just enable `alloc` as we don't need the features that `std` provides.

### Notes to the reviewers

`alloc` is just a subset of `std`. `json_serde` has `from_reader()` & `to_writer()` methods if `std` is enabled. We aren't using those methods so just having `alloc` is fine.

I'm also not sure why an explicit `alloc` cargo feature flag is needed in `serde_json`. They can just have an internal prelude based on whether the `std` feature is enabled.

#### All Submissions:


luisschwab (approved):
ACK 546b07e23b6219693976d7a4eeaacdc7c304f04a

oleonardolima (approved):
ACK 546b07e23b6219693976d7a4eeaacdc7c304f04a

PR #2100, opened 2026-01-24
Make mining more flexible in `bdk_testenv`

evanlinjin:
### Description

Add `mine_block` to `bdk_testenv::Env` with custom timestamp and coinbase address. This allows us to test timelocked transactions.

### Changelog notice

```md
Added
- `mine_block` method to `bdk_testenv::Env` with custom `MineParams`. Blocks can be mined with no transactions, contain custom timestamp and custom coinbase address.
- `min_time_for_next_block` and `get_block_template` helper methods.
```

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/testenv/src/lib.rs
+++ b/crates/testenv/src/lib.rs
@@ -3,9 +3,16 @@
 pub mod utils;
 
 use anyhow::Context;
+use bdk_chain::bitcoin::{
+    block::Header, hash_types::TxMerkleNode, hex::FromHex, script::PushBytesBuf, transaction,
+    Address, Amount, Block, BlockHash, ScriptBuf, Transaction, TxIn, TxOut, Txid,
+};
 use bdk_chain::CheckPoint;
-use bitcoin::{address::NetworkChecked, Address, Amount, BlockHash, Txid};
-use std::time::Duration;
+use bitcoin::address::NetworkChecked;
+use bitcoin::hex::HexToBytesError;
+use core::time::Duration;
+use electrsd::corepc_node::mtype::GetBlockTemplate;
+use electrsd::corepc_node::{TemplateRequest, TemplateRules};
 
 pub use electrsd;
 pub use electrsd::corepc_client;
@@ -42,6 +49,32 @@ impl Default for Config<'_> {
                 conf
             },
         }
+    }
+}
+
+/// Parameters for [`TestEnv::mine_block`].
+#[non_exhaustive]
+#[derive(Default)]
+pub struct MineParams {
+    /// If `true`, the block will be empty (no mempool transactions).
+    pub empty: bool,
+    /// Set a custom block timestamp. Defaults to `max(min_time, now)`.
+    pub time: Option<u32>,
+    /// Set a custom coinbase output script. Defaults to `OP_TRUE`.
+    pub coinbase_address: Option<ScriptBuf>,
+}
+
+impl MineParams {
+    fn address_or_anyone_can_spend(&self) -> ScriptBuf {
+        use bdk_chain::bitcoin::opcodes::OP_TRUE;
+        self.coinbase_address
+            .clone()
+            // OP_TRUE (anyone can spend)
+            .unwrap_or_else(|| {
+                bdk_chain::bitcoin::script::Builder::new()
+                    .push_opcode(OP_TRUE)
+                    .into_script()
+            })
     }
 }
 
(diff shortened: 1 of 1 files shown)


oleonardolima:
@evanlinjin I'd suggest you start working on this on top of #1826, as we're migrating to `corepc` library/API.

evanlinjin:
> @evanlinjin I'd suggest you start working on this on top of #1826, as we're migrating to `corepc` library/API.

Good point. I'll do so once that PR is merged. I would like to test this against `bdk-tx` right now and that's using the old `bdk_testenv` (for testing MTP).

crates/testenv/src/lib.rs
```diff
@@ -45,6 +54,27 @@ impl Default for Config<'_> {
...
+    /// Set a custom coinbase output script. Defaults to `OP_TRUE`.
+    pub coinbase_address: Option<ScriptBuf>,
+}
+
+impl MineParams {
+    fn address_or_anyone_can_spend(&self) -> ScriptBuf {
+        self.coinbase_address
+            .clone()
+            // OP_TRUE (anyone can spend)
+            .unwrap_or(ScriptBuf::from_bytes(vec![0x51]))
```
oleonardolima: nit: could we use the [`OP_TRUE`](https://docs.rs/bitcoin/0.32.8/bitcoin/blockdata/opcodes/static.OP_TRUE.html) instead of the hex directly ?

crates/testenv/src/lib.rs
```diff
@@ -119,70 +149,157 @@ impl TestEnv {
...
+            output: coinbase_outputs,
+        };
+
+        let txdata = if params.empty {
+            vec![coinbase_tx]
+        } else {
+            core::iter::once(coinbase_tx)
+                .chain(bt.transactions.iter().map(|tx| {
+                    let raw = Vec::<u8>::from_hex(&tx.data).expect("must be valid hex");
+                    Transaction::consensus_decode(&mut raw.as_slice()).expect("must decode tx")
```
oleonardolima: nit+question: can't you use the `.into_model()` for the `get_block_template` in order to not have to do all the decoding here and on the other usage of it above?

crates/testenv/src/lib.rs
```diff
@@ -119,70 +149,157 @@ impl TestEnv {
...
+                .collect()
+        };
 
         let mut block = Block {
             header: Header {
-                version: bt.version,
-                prev_blockhash: bt.previous_block_hash,
-                merkle_root: TxMerkleNode::all_zeros(),
-                time: Ord::max(
+                version: bdk_chain::bitcoin::blockdata::block::Version::from_consensus(bt.version),
```
oleonardolima: nit+question: same as above, i think if you use the `.into_model()` would help with all these parsing/conversions.

oleonardolima (commented):
utACK 9d7f859ee29ea45ab986c8bf482200c82abfdeee

Overall looks good. I left a few comments that might help with removing some decoding done here, didn't tried the changes yet though.

crates/testenv/src/lib.rs
```diff
@@ -119,52 +152,132 @@ impl TestEnv {
...
+                let bytes_buf = PushBytesBuf::try_from(bytes).expect("must be valid bytes");
+                builder = builder.push_slice(bytes_buf);
+            }
+            let script = builder.into_script();
+            // Ensure scriptSig is at least 2 bytes (pad with OP_0 if needed)
+            if script.len() < 2 {
+                bdk_chain::bitcoin::script::Builder::new()
+                    .push_int(bt.height as i64)
+                    .push_opcode(bdk_chain::bitcoin::opcodes::OP_0)
+                    .into_script()
```
oleonardolima: question: shouldn't this one also have the coinbase_aux values ?

oleonardolima: I think this whole coinbase_scriptsig codeblock could be reworked, maybe even be extracted to a helper fn.

evanlinjin: Woops, I over-relied on Claude here.

evanlinjin: @oleonardolima Fixed it. Let me know what you think.

crates/testenv/src/lib.rs
```diff
@@ -119,52 +152,135 @@ impl TestEnv {
...
+        self.mine_block(MineParams {
+            empty: true,
+            ..Default::default()
+        })
+    }
+
+    /// Get the minimum valid timestamp for the next block.
+    pub fn min_time_for_next_block(&self) -> anyhow::Result<u32> {
+        Ok(self.get_block_template()?.min_time)
+    }
```
oleonardolima: nit: currently it's only being used on the tests, is it really needed ? I mean, the same can be achieved directly calling the getblocktemplate.

crates/testenv/src/lib.rs
```diff
@@ -119,52 +152,135 @@ impl TestEnv {
...
+            }
+            builder.into_script()
+        }
+        let coinbase_scriptsig = {
+            let mut script = build_coinbase_scriptsig(&bt, false);
+            // Ensure scriptSig is at least 2 bytes (pad with OP_0 if needed)
+            if script.len() < 2 {
+                script = build_coinbase_scriptsig(&bt, true);
+            };
+            script
```
oleonardolima: nit: it's only being used here, so the padding could be handled directly, but it's not a blocker though.

crates/testenv/src/lib.rs
```diff
@@ -119,52 +152,135 @@ impl TestEnv {
...
+                    .as_ref()
+                    .map(|s| -> Result<_, HexToBytesError> {
+                        Ok(TxOut {
+                            value: Amount::ZERO,
+                            script_pubkey: ScriptBuf::from_hex(s)?,
+                        })
+                    })
+                    .transpose()?,
+            )
+            .collect()
```
oleonardolima, line 235: question: was this created by claude, or there's an specific reason you need this ?

evanlinjin: Which part are you referring to exactly? It's handy to have the ability to mine empty blocks.

crates/testenv/src/lib.rs
evanlinjin: What do you mean?

oleonardolima: I mean, this block could be internal to the fn above and no need for a second argument, as it's only being used here.

crates/testenv/src/lib.rs
oleonardolima: I mean, can't we always build it using the bt info ?

oleonardolima (approved):
utACK 9620a7d991bb6fe67522702acde7f6bf8da98e3a

I don't mean to keep bikeshedding on the review, all the comments can be addressed in a follow-up. I didn't test it, though as the fns are already been used in existing tests, I think it looks fine.

notmandatory:
This will need a rebase since I merged #2123.

crates/testenv/src/lib.rs
evanlinjin: Are you talking about the `default_witness_commitment`? If we don't have any transactions in the block, there will be no commitment.

oleonardolima (approved):
ACK 9b1111fd18eed69839ca10e8cacb3b513a8a7aa4

It looks good! Thanks for addressing the comments above.

_I didn't do any fancy local test though, as it's being used throughout multiple existing tests (through the `bdk_testenv`)._

PR #2102, opened 2026-01-24
docs: add minimal Esplora backend usage example

rishabhyadav03:
This PR adds a small Esplora backend example to the README to help new users quickly understand how to fetch blockchain data without running a full code.


thunderbiscuit:
NACK from me. This really doesn't belong on our root readme, let alone the fact that it showcases an example from _another crate_ not in this repository.

I'll give it a day or two for others to comment if they need to, and close then.

oleonardolima:
NACK a51efb9311bfc593501244eb78868ef268899508, agree with thunder's comments above.

PR #2103, opened 2026-01-24
Add request timeout to Esplora client in example

rishabhyadav03:
This PR adds a request timeout to the Esplora client used by the example application. Configuring a timeout helps avoid hanging network requests and makes the example more robust for real world usage.


oleonardolima (commented):
NACK 0a36b0a3b36cf7c064472ecb22d01df6d31cb5bc

As also mentioned in this other [comment](https://github.com/bitcoindevkit/bdk/pull/2102#issuecomment-3819394545) it does not belong to the `README.md`. Maybe the example could use the timeout, but that would require it to be reworked to actually be an argument/parameter, instead of just setting it up straight to 60, as it's currently being done in the PR.

Issue #2076, opened 2025-11-24
`CheckPoint` design issue


ValuedMammal:
These are some of the proposed changes to `CheckPoint`. I'm opening this as a record of ongoing discussion as it relates to the `CheckPoint` structure and functionality.

- [x] Add MTP calculation #2036
- [x] Validate CheckPoint links via previous hash #2021
- [ ] Add efficient comparison #2033
- [ ] Improve search complexity #2048

ValuedMammal:
In order to evaluate something proposing a significant change to the structure or design of the `CheckPoint`, it's important to clarify the specific requirements of the `CheckPoint`. What does this data structure need to accomplish, and what is the best way to do it while keeping the tradeoffs in mind?

- If there are enhancements or optimizations to be made, can they be done within the current design contraints?  
- If it's a design change being proposed, what is the new design? Does it uphold the same requirements?  
- Is there a chance to simplify by benefiting from types that exist either in the std library or elsewhere?  
- Is there a sufficiently motivated change that warrants the time and investment on a complete rewrite?

oleonardolima:
Awesome, thanks for compiling all the issues here! Agree, I think it's a good idea to have the big picture and the tradeoffs clear in mind. I'd add that we should probably compile all these changes in a "Checkpoint's ADR" at some point.

I'll try to come up with the answers from my perspective once I settle back home later.

evanlinjin:
#2024 is a PR attempting to solve #2021. @ValuedMammal I'll address your questions here, in the context of #2021.

---

#### What does `CheckPoint` need to accomplish?

At its core, `CheckPoint` represents a sparse, singly-linked chain of block references used to:

* Track which blocks we know about (for wallet sync and canonicalization)
* Detect reorgs by finding points of agreement between local and remote chains
* Anchor transactions to specific blocks

And in the future:

* Calculate MTP time so we can create/spend time-locked outputs properly.

I thought of the following design constraints for `CheckPoint`:

* Represent a sparse sample of the blockchain (not every block)
* Support iteration from tip towards genesis
* Detect points of agreement between two chains
* Insert checkpoints at arbitrary heights
* Merge chains (for applying sync updates)
* Multiple components can hold references to the same chain
* Chains can share common tails (diverge at some height, share everything below)
* Cloning a reference should be cheap

[...]

PR #2115, opened 2026-02-05
Add `prev_blockhash` validation to `CheckPoint`

### LINKED ISSUE #2021: Ensure `CheckPoint` chain methods validate and link via previous blockhash

evanlinjin:
Closes #2021
Related to #2076
Replaces #2024
Replaces #2091

### Description

This PR adds `prev_blockhash` awareness to `CheckPoint`, enabling proper chain validation when merging checkpoint chains that store block headers or similar data with previous block hash information.

### Notes to the reviewers

This PR replaces some prior attempts:

* #2024 - where we made the `CheckPoint::data` optional - however this resulted in internal complexity and an API with annoying edge cases. The tests from this PR were still useful.

* #2091 - This second attempt had some good ideas, but was distracted from the goal of #2021. I mostly reused the `CheckPoint::insert` implementation of that PR.

### Changelog notice

```md
Added:
- `ToBlockHash::prev_blockhash()` - optional method to expose previous block hash
- `CheckPointEntry` - new type for iterating with `prev_blockhash` awareness, yielding "placeholder" entries for heights inferred from `prev_blockhash`
- `ApplyBlockError` - this is a new error type with two variants; `MissingGenesis` and `PrevBlockhashMismatch`. The second variant is a new error case introduced by `prev_blockhash` awareness.

Changed:
- `CheckPoint::push` - now errors when `prev_blockhash` conflicts with current tip (contiguous heights)
- `CheckPoint::insert` - now evicts/displaces checkpoints on `prev_blockhash` conflict
- `merge_chains` - now validates `prev_blockhash` consistency when merging
- `LocalChain<D>` generic parameter - relaxed constraint to `D: Clone` instead of `D: Copy`.

Fixed:
- `merge_chains` no longer replaces the genesis block.
```

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -6,8 +6,8 @@ use core::ops::RangeBounds;
 
 use crate::collections::BTreeMap;
 use crate::{BlockId, ChainOracle, Merge};
-use bdk_core::{CheckPointEntry, ToBlockHash};
 pub use bdk_core::{CheckPoint, CheckPointIter};
+use bdk_core::{CheckPointEntry, ToBlockHash};
 use bitcoin::block::Header;
 use bitcoin::BlockHash;
 
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -263,7 +263,7 @@ where
         CheckPoint::from_blocks(blocks)
             .map(|tip| Self { tip })
             .map_err(|err| {
-                let last_cp = err.expect("must have atleast one block (genesis)");
+                let last_cp = err.expect("must have at least one block (genesis)");
                 ApplyBlockError::PrevBlockhashMismatch {
                     expected: last_cp.block_id(),
                 }
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -15,7 +15,7 @@ use bitcoin::BlockHash;
 fn apply_changeset_to_checkpoint<D>(
     mut init_cp: CheckPoint<D>,
     changeset: &ChangeSet<D>,
-) -> Result<CheckPoint<D>, MissingGenesisError>
+) -> Result<CheckPoint<D>, ApplyBlockError>
 where
     D: ToBlockHash + fmt::Debug + Clone,
 {
@@ -48,7 +48,11 @@ where
         let new_tip = match base {
             Some(base) => base
                 .extend(extension)
-                .expect("extension is strictly greater than base"),
+                // Since `extension` is in height order, the only failure case is `prev_blockhash`
+                // mismatch.
+                .map_err(|last_cp| ApplyBlockError::PrevBlockhashMismatch {
+                    expected: last_cp.block_id(),
+                })?,
             None => LocalChain::from_blocks(extension)?.tip(),
         };
         init_cp = new_tip;
@@ -251,22 +255,27 @@ where
     ///
     /// The [`BTreeMap`] enforces the height order. However, the caller must ensure the blocks are
     /// all of the same chain.
-    pub fn from_blocks(blocks: BTreeMap<u32, D>) -> Result<Self, MissingGenesisError> {
+    pub fn from_blocks(blocks: BTreeMap<u32, D>) -> Result<Self, ApplyBlockError> {
         if !blocks.contains_key(&0) {
-            return Err(MissingGenesisError);
+            return Err(ApplyBlockError::MissingGenesis);
         }
 
-        Ok(Self {
-            tip: CheckPoint::from_blocks(blocks).expect("blocks must be in order"),
-        })
+        CheckPoint::from_blocks(blocks)
+            .map(|tip| Self { tip })
+            .map_err(|err| {
+                let last_cp = err.expect("must have atleast one block (genesis)");
+                ApplyBlockError::PrevBlockhashMismatch {
+                    expected: last_cp.block_id(),
+                }
+            })
     }
 
     /// Construct a [`LocalChain`] from an initial `changeset`.
-    pub fn from_changeset(changeset: ChangeSet<D>) -> Result<Self, MissingGenesisError> {
+    pub fn from_changeset(changeset: ChangeSet<D>) -> Result<Self, ApplyBlockError> {
         let genesis_entry = changeset.blocks.get(&0).cloned().flatten();
         let genesis_data = match genesis_entry {
             Some(data) => data,
-            None => return Err(MissingGenesisError),
+            None => return Err(ApplyBlockError::MissingGenesis),
         };
 
         let (mut chain, _) = Self::from_genesis(genesis_data);
@@ -310,7 +319,7 @@ where
     }
 
     /// Apply the given `changeset`.
-    pub fn apply_changeset(&mut self, changeset: &ChangeSet<D>) -> Result<(), MissingGenesisError> {
+    pub fn apply_changeset(&mut self, changeset: &ChangeSet<D>) -> Result<(), ApplyBlockError> {
         let old_tip = self.tip.clone();
         let new_tip = apply_changeset_to_checkpoint(old_tip, changeset)?;
         self.tip = new_tip;
@@ -485,6 +494,35 @@ impl<D> FromIterator<(u32, D)> for ChangeSet<D> {
     }
 }
 
+/// Error when applying blocks to a local chain.
+#[derive(Clone, Debug, PartialEq)]
+pub enum ApplyBlockError {
+    /// Genesis block is missing or would be altered.
+    MissingGenesis,
+    /// Block's `prev_blockhash` doesn't match the expected block.
+    PrevBlockhashMismatch {
+        /// The block that `prev_blockhash` should reference.
+        expected: BlockId,
+    },
+}
+
+impl core::fmt::Display for ApplyBlockError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ApplyBlockError::MissingGenesis => {
+                write!(f, "genesis block is missing or would be altered")
+            }
+            ApplyBlockError::PrevBlockhashMismatch { expected } => write!(
+                f,
+                "`prev_blockhash` doesn't match block at height {} ({})",
+                expected.height, expected.hash
+            ),
+        }
+    }
+}
+
+impl core::error::Error for ApplyBlockError {}
+
 /// An error which occurs when a [`LocalChain`] is constructed without a genesis checkpoint.
 #[derive(Clone, Debug, PartialEq)]
 pub struct MissingGenesisError;
@@ -592,6 +630,30 @@ fn merge_chains<D>(
 where
     D: ToBlockHash + fmt::Debug + Clone,
 {
+    // Apply the changeset to produce the final merged chain.
+    //
+    // `PrevBlockhashMismatch` should never happen because the merge iteration detects
+    // `prev_blockhash` conflicts and resolves them by invalidating conflicting blocks (setting
+    // them to `None` in the changeset) before we reach this point.
+    fn finish<D>(
+        original_tip: CheckPoint<D>,
+        changeset: ChangeSet<D>,
+    ) -> Result<(CheckPoint<D>, ChangeSet<D>), CannotConnectError>
+    where
+        D: ToBlockHash + fmt::Debug + Clone,
+    {
+        let new_tip = apply_changeset_to_checkpoint(original_tip, &changeset).map_err(|err| {
+            debug_assert!(
+                matches!(err, ApplyBlockError::MissingGenesis),
+                "PrevBlockhashMismatch should never happen"
+            );
+            CannotConnectError {
+                try_include_height: 0,
+            }
+        })?;
+        Ok((new_tip, changeset))
+    }
+
     let mut changeset = ChangeSet::<D>::default();
 
     let mut orig = original_tip.entry_iter();
@@ -676,11 +738,13 @@ where
                         if is_update_height_superset_of_original {
                             return Ok((update_tip, changeset));
                         } else {
-                            let new_tip = apply_changeset_to_checkpoint(original_tip, &changeset)
-                                .map_err(|_| CannotConnectError {
-                                try_include_height: 0,
-                            })?;
-                            return Ok((new_tip, changeset));
+                            return finish(original_tip, changeset);
+                        }
+                    }
+                    // Update placeholder with real data (if necessary).
+                    if let Some(u_data) = u.data_ref() {
+                        if o.is_placeholder() {
+                            changeset.blocks.insert(u.height(), Some(u_data.clone()));
                         }
                     }
                 } else {
@@ -715,10 +779,5 @@ where
         }
     }
 
-    let new_tip = apply_changeset_to_checkpoint(original_tip, &changeset).map_err(|_| {
-        CannotConnectError {
-            try_include_height: 0,
-        }
-    })?;
-    Ok((new_tip, changeset))
+    finish(original_tip, changeset)
 }
--- a/crates/chain/src/local_chain.rs
+++ b/crates/chain/src/local_chain.rs
@@ -631,10 +631,6 @@ where
     D: ToBlockHash + fmt::Debug + Clone,
 {
     // Apply the changeset to produce the final merged chain.
-    //
-    // `PrevBlockhashMismatch` should never happen because the merge iteration detects
-    // `prev_blockhash` conflicts and resolves them by invalidating conflicting blocks (setting
-    // them to `None` in the changeset) before we reach this point.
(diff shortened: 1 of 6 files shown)


crates/chain/src/local_chain.rs
```diff
@@ -671,7 +676,8 @@ where
                     prev_orig_was_invalidated = false;
                     // OPTIMIZATION 2 -- if we have the same underlying pointer at this point, we
                     // can guarantee that no older blocks are introduced.
-                    if o.eq_ptr(u) {
+                    // TODO: Ensure this is correct!
```
evanlinjin: This is correct! We can remove this comment!

crates/chain/src/local_chain.rs
evanlinjin, line 1: As highlighted in f7eedf3f1e84e62d7480cb24de6f4a41204b9c76, we should do something about `apply_changeset_to_checkpoint`.

evanlinjin:
This PR is actually ready for review now!

crates/chain/src/local_chain.rs
```diff
@@ -485,6 +494,36 @@ impl<D> FromIterator<(u32, D)> for ChangeSet<D> {
     }
 }
 
+/// Error when applying blocks to a local chain.
+#[derive(Clone, Debug, PartialEq)]
+pub enum ApplyBlockError {
+    /// Genesis block is missing or would be altered.
+    MissingGenesis,
```
nymius, line 501: I would split this error in two to keep a single semantic for each one: `MissingGenesis` and `ConflictingGenesis` or similar

crates/chain/src/local_chain.rs
```diff
@@ -634,13 +697,18 @@ where
         match (curr_orig.as_ref(), curr_update.as_ref()) {
             // Update block that doesn't exist in the original chain
             (o, Some(u)) if Some(u.height()) > o.map(|o| o.height()) => {
-                changeset.blocks.insert(u.height(), Some(u.data()));
+                // Only append to `ChangeSet` when this is an actual checkpoint.
```
nymius: What's the definition of "actual checkpoint": one that isn't a Placeholder or one that contains some data?
Why would it be a checkpoint with no data but at the same time is not a Placeholder?
Wouldn't be better to rewrite this comment to not leave to the reader the definition of "actual checkpoint"?

crates/core/src/checkpoint.rs
```diff
@@ -281,74 +310,119 @@ where
...
+            } else if cp.height() < height {
+                break Some(cp);
             }
 
-            if cp.height() < height {
-                break cp;
+            // Continue traversing down (if possible).
+            match cp.prev() {
+                Some(prev) => cp = prev,
+                None => break None,
```
nymius, line 376: When we hit `None` on this iterator, it means we have exhausted `CheckPoint` so we are beyond genesis block?

crates/core/tests/test_checkpoint.rs
```diff
@@ -186,3 +186,387 @@ fn test_mtp_sparse_chain() {
...
+    // Verify height 6 is purged
+    assert!(
+        result.get(6).is_none(),
+        "checkpoint at 6 should be purged due to prev_blockhash conflict"
+    );
+
+    // Verify chain structure
+    assert_eq!(result.height(), 5, "tip should be at height 5");
+    // Should have: checkpoint 5 only
+    assert_eq!(
```
nymius: What's the difference between displaced and purged if 4 is not part of the chain?

crates/core/tests/test_checkpoint.rs
```diff
@@ -186,3 +186,387 @@ fn test_mtp_sparse_chain() {
...
+        result.is_ok(),
+        "push should succeed for non-contiguous height"
+    );
+    let new_cp = result.unwrap();
+
+    // Verify the tip is at 105
+    assert_eq!(new_cp.height(), 105);
+    assert_eq!(new_cp.hash(), hash!("block_105"));
+
+    // Verify chain structure: 100, 105
```
nymius: This is not precisely verifying chain structure. An explicit check for both `CheckPoint`s will be more accurate.

crates/core/tests/test_checkpoint.rs
```diff
@@ -186,3 +186,387 @@ fn test_mtp_sparse_chain() {
...
+
+/// Test `insert` should panic if trying to replace genesis with a different block.
+#[test]
+#[should_panic(expected = "inserted data implies different genesis")]
+fn checkpoint_insert_cannot_replace_genesis() {
+    let block_0 = TestBlock {
+        blockhash: hash!("block_0"),
+        prev_blockhash: hash!("genesis_parent"),
+    };
+    let block_1 = TestBlock {
```
nymius: Isn't this redundant for this test case?

crates/core/tests/test_checkpoint.rs
```diff
@@ -186,3 +186,387 @@ fn test_mtp_sparse_chain() {
...
+
+/// Test `insert` should panic if inserted data's prev_blockhash implies a different genesis.
+#[test]
+#[should_panic(expected = "inserted data implies different genesis")]
+fn checkpoint_insert_cannot_displace_genesis() {
+    let block_0 = TestBlock {
+        blockhash: hash!("block_0"),
+        prev_blockhash: hash!("genesis_parent"),
+    };
+    let block_1 = TestBlock {
```
nymius: Isn't this redundant for this test case?

crates/core/src/checkpoint.rs
```diff
@@ -256,21 +284,22 @@ where
     /// Construct from an iterator of block data.
     ///
     /// Returns `Err(None)` if `blocks` doesn't yield any data. If the blocks are not in ascending
-    /// height order, then returns an `Err(..)` containing the last checkpoint that would have been
-    /// extended.
+    /// height order, or there are any `prev_blockhash` mismatches, then returns an `Err(..)`
+    /// containing the last checkpoint that would have been extended.
     pub fn from_blocks(blocks: impl IntoIterator<Item = (u32, D)>) -> Result<Self, Option<Self>> {
```
nymius, line 295: Doesn't this merit an `Error` and `Returns` section?

nymius (commented):
Approach ACK 2c29ff1438f5c737fd87ece6e39e5ebd43b8eb49

I think the "ghost" checkpoint approach is good enough for managing the new "not-really-in-chain" checkpoints.

After detaining myself for a while trying to understand for the first time the inner workings of `merge_chain`, I think it is hard to read, and it's going to get hairy to maintain in the future.
I will come back with some ideas to address that.

Something similar happens with the `CheckPointEntry::prev` recursion. Some unit test will come handy in the future.

I liked the format of `TestLocalChain`, the tables were very useful.

crates/chain/src/local_chain.rs
evanlinjin: "actual checkpoint" is a checkpoint that is not a placeholder, a.k.a. a checkpoint that actually contains data.

Actual checkpoints are persisted, placeholder checkpoints are not. If this is a placeholder checkpoint, we don't want to persist it as placeholder checkpoints are already implied by actual checkpoints (and there is no data to persist anyway).

crates/core/src/checkpoint.rs
evanlinjin: No. The original chain may not have a genesis. We will not hit this if the original chain has a genesis block.

crates/core/tests/test_checkpoint.rs
evanlinjin: Purged is *"transitively displaced"*.

In this test, `5'` has a `prev_blockhash` that displaces `4`. Since, the original chain connects `4` to `6`, if `4` is displaced, then `6` is purged.

crates/chain/src/local_chain.rs
nymius: Ok, so all `Placeholder`s are "`Checkpoints` without data" and all "`Checkpoints` without data" are `Placeholders`, i.e., there is a one-to-one mapping between the wrapper iter and the actual structs.

crates/core/src/checkpoint.rs
nymius: > The original chain may not have a genesis.

There is some no-buggy reason this could happen?

evanlinjin: I haven't encountered it in person, but if there's no reason to error/panic, then we shouldn't do it.

`CheckPoint` is an update type, chain sources might merge chains together before forming an update (not sure).

evanlinjin:
> After detaining myself for a while trying to understand for the first time the inner workings of merge_chain, I think it is hard to read, and it's going to get hairy to maintain in the future.
I will come back with some ideas to address that.

@nymius thanks for looking into that! For this PR, it's good enough to just understand how `merge_chains` is meant to function, and ensure our tests covers all cases and edge cases. Just treat it like a black box for now and we can look into it later.

nymius:
> @nymius thanks for looking into that! For this PR, it's good enough to just understand how `merge_chains` is meant to function, and ensure our tests covers all cases and edge cases. Just treat it like a black box for now and we can look into it later.

Yes, not intention to turn this into a refactor. Something for another PR.

crates/chain/src/local_chain.rs
```diff
@@ -485,6 +494,36 @@ impl<D> FromIterator<(u32, D)> for ChangeSet<D> {
...
+                f,
+                "`prev_blockhash` doesn't match block at height {} ({})",
+                expected.height, expected.hash
+            ),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for ApplyBlockError {}
```
nymius: [`core::error::Error` was stabilized in 1.81](https://doc.rust-lang.org/nightly/core/error/trait.Error.html), as MSRV is 1.85 now, I think we can remove this (and all the others too)

crates/chain/src/local_chain.rs
```diff
@@ -594,18 +633,42 @@ fn merge_chains<D>(
...
+            debug_assert!(
+                matches!(err, ApplyBlockError::MissingGenesis),
+                "PrevBlockhashMismatch should never happen"
+            );
+            CannotConnectError {
+                try_include_height: 0,
+            }
+        })?;
+        Ok((new_tip, changeset))
+    }
```
nymius, line 663: This is not improving the comprehension in a meaningful way. I would remove this function and add the comment directly on top of the code, once moved to the end.

crates/chain/tests/test_local_chain.rs
```diff
@@ -953,3 +1016,296 @@ proptest! {
...
+        // chain  | _       C(prev=B)  D
+        // update | _   A              D
+        // result | _   A              D
+        // ```
+        //
+        // Both chains agree on D at height 4. Chain has C at height 2 with `prev_blockhash = B`,
+        // but no B exists. Update introduces A at height 1, which displaces C because
+        // C's `prev_blockhash` ("B") doesn't match A's hash ("A").
+        //
+        // Note: This can only happen if chains are constructed incorrectly.
```
nymius: If this can only happen if chains are constructed incorrectly, shouldn't the expected behavior be to fail then?

crates/core/tests/test_checkpoint.rs
```diff
@@ -186,3 +186,379 @@ fn test_mtp_sparse_chain() {
...
+    let result = cp.clone().push(101, conflicting_block);
+
+    assert!(
+        result.is_err(),
+        "push should fail when prev_blockhash conflicts"
+    );
+    assert!(
+        result.unwrap_err().eq_ptr(&cp),
+        "should return self on error"
+    );
```
nymius, line 458: To give a full picture of the expected behavior I would check the chain still contains block 100.

crates/core/tests/test_checkpoint.rs
```diff
@@ -186,3 +186,379 @@ fn test_mtp_sparse_chain() {
...
+
+    // Push at height 101 with matching prev_blockhash
+    let result = cp.push(101, matching_block);
+
+    assert!(
+        result.is_ok(),
+        "push should succeed when prev_blockhash matches"
+    );
+    let new_cp = result.unwrap();
+    assert_eq!(new_cp.height(), 101);
```
nymius, line 490: Check that `new_cp.len()` is 2.

nymius (commented):
The code is addressing the referenced issues, and the changes correlate with the discussion on them.
I haven't found any conceptual concerns, but would like to hear other voices @ValuedMammal.

P.S.: I wasn't expecting to do that many comments, next time I'll push them in a single review. Sorry

ValuedMammal:
I'd prefer if we moved the checkpoint entry feature to a separate PR and just focus this one on validating prev_blockhash connections.

evanlinjin:
> I'd prefer if we moved the checkpoint entry feature to a separate PR and just focus this one on validating prev_blockhash connections.

@ValuedMammal `CheckPointEntry` is actually the mechanism through which `prev_blockhash` validation works in `merge_chains` — it fills in gaps between checkpoints using `prev_blockhash`, allowing the merge logic to detect conflicts at heights where no checkpoint exists. These two features are tightly coupled, so splitting them would mean either rewriting `merge_chains` without `CheckPointEntry` or landing `CheckPointEntry` first as a dependency PR.

crates/chain/src/local_chain.rs
evanlinjin: Good suggestion. I think it's better to implement this in a separate PR. I created #2132

crates/chain/tests/test_local_chain.rs
evanlinjin: If the original chain is already in a bad state, erroring would prevent recovery. The merge should be best-effort.

crates/core/src/checkpoint.rs
evanlinjin: Added in 6191afcd6bfafd90d1eaaedc66b8c72202a5cdb2

crates/core/tests/test_checkpoint.rs
evanlinjin: Done in 6191afcd6bfafd90d1eaaedc66b8c72202a5cdb2

crates/core/tests/test_checkpoint.rs
evanlinjin: Added in 6191afcd6bfafd90d1eaaedc66b8c72202a5cdb2

crates/chain/src/local_chain.rs
evanlinjin: I'd prefer to keep `finish` — it's called from two places and contains a `debug_assert!` that documents an important invariant. Inlining would mean duplicating the assert + error mapping at both call sites. The named function also communicates intent — "we're done iterating, apply the result."

crates/chain/src/local_chain.rs
evanlinjin: This is already addressed — `ConflictingGenesis` was removed. The current enum only has `MissingGenesis` and `PrevBlockhashMismatch`, each with a single semantic.

crates/core/tests/test_checkpoint.rs
evanlinjin: Addressed — both checkpoints are now explicitly checked (height + hash) in addition to the count.

crates/core/tests/test_checkpoint.rs
evanlinjin: Addressed — the redundant blocks have been removed from these tests.

crates/core/tests/test_checkpoint.rs
evanlinjin: Addressed — same as above.

crates/chain/src/local_chain.rs
nymius: Good point, leave as it is.

crates/chain/tests/test_local_chain.rs
nymius: So merge has the chances of fixing the original chain state?

nymius (approved):
ACK 3a101d801d493da89ec0897ccf610a423d3504d5

Needs rebase.

crates/chain/tests/test_local_chain.rs
evanlinjin: Yes. The alternative is worse.

Failing loudly on a state that is, by assumption, reachable but malformed means: 
* (a) you need a separate repair API for users to dig themselves out,
* (b) the enum grows a variant nobody wants to handle,
* (c) callers have to match on it to do the thing the library could have just done for them.

evanlinjin:
self-ACK c486ba75865806bbe1d88c646385ac57df611ef5

@nymius Thanks for the review. I've rebased, got Claude to review the codebase one more time and fixed/improved docs and a few small fixes.

PR #2119, opened 2026-02-09
Refactor: update bdk-bitcoind-rpc to use bdk-bitcoind-client

tvpeter:
### Description

This PR replaces `bitcoincore-rpc` with `bdk-bitcoind-client` in `bdk-bitcoind-rpc` library and its tests and examples.

Depends on [bdk-bitcoind-client#5](https://github.com/bitcoindevkit/bdk-bitcoind-client/pull/5)

### Notes to the reviewers

- changed the Emitter to a concrete type
- In the example crate, I replaced the shared client between threads with two different clients (will come back to improve this later)

### Changelog notice

- replace `bitcoincore-rpc` with `bdk-bitcoind-client`

#### All Submissions:

### DIFF
--- a/crates/bitcoind_rpc/Cargo.toml
+++ b/crates/bitcoind_rpc/Cargo.toml
@@ -26,9 +26,10 @@ bdk_testenv = { path = "../testenv" }
 bdk_chain = { path = "../chain" }
 
 [features]
-default = ["std"]
+default = ["std", "bitcoind_28_0"]
 std = ["bitcoin/std", "bdk_core/std"]
 serde = ["bitcoin/serde", "bdk_core/serde"]
+bitcoind_28_0 =["bdk-bitcoind-client/28_0"]
 
 [[example]]
 name = "filter_iter"
--- a/crates/bitcoind_rpc/Cargo.toml
+++ b/crates/bitcoind_rpc/Cargo.toml
@@ -17,7 +17,7 @@ workspace = true
 
 [dependencies]
 bitcoin = { version = "0.32.0", default-features = false }
-bitcoincore-rpc = { version = "0.19.0" }
+bdk-bitcoind-client ={ git = "https://github.com/bitcoindevkit/bdk-bitcoind-client", rev = "06526db5c25047b61d29f81b25cfee7bd0f3abc8"}
 bdk_core = { path = "../core", version = "0.6.1", default-features = false }
 
 [dev-dependencies]
--- a/crates/bitcoind_rpc/Cargo.toml
+++ b/crates/bitcoind_rpc/Cargo.toml
@@ -17,7 +17,7 @@ workspace = true
 
 [dependencies]
 bitcoin = { version = "0.32.0", default-features = false }
-bdk-bitcoind-client ={ git = "https://github.com/bitcoindevkit/bdk-bitcoind-client", rev = "06526db5c25047b61d29f81b25cfee7bd0f3abc8"}
+bdk_bitcoind_client = { version = "0.1.0", default-features = false }
 bdk_core = { path = "../core", version = "0.6.1", default-features = false }
 
 [dev-dependencies]
@@ -29,7 +29,9 @@ bdk_chain = { path = "../chain" }
 default = ["std", "bitcoind_28_0"]
 std = ["bitcoin/std", "bdk_core/std"]
 serde = ["bitcoin/serde", "bdk_core/serde"]
-bitcoind_28_0 =["bdk-bitcoind-client/28_0"]
+bitcoind_28_0 = ["bdk_bitcoind_client/28_0"] 
+bitcoind_29_0 = ["bdk_bitcoind_client/29_0"]
+bitcoind_30_0 = ["bdk_bitcoind_client/30_0"]
 
 [[example]]
 name = "filter_iter"
--- a/crates/bitcoind_rpc/examples/filter_iter.rs
+++ b/crates/bitcoind_rpc/examples/filter_iter.rs
@@ -44,8 +44,10 @@ fn main() -> anyhow::Result<()> {
     // Configure RPC client
     let url = std::env::var("RPC_URL").context("must set RPC_URL")?;
     let cookie = std::env::var("RPC_COOKIE").context("must set RPC_COOKIE")?;
-    let rpc_client =
-        bitcoincore_rpc::Client::new(&url, bitcoincore_rpc::Auth::CookieFile(cookie.into()))?;
+    let rpc_client = bdk_bitcoind_client::Client::with_auth(
+        &url,
+        bdk_bitcoind_client::Auth::CookieFile(cookie.into()),
+    )?;
 
     // Initialize `FilterIter`
     let mut spks = vec![];
--- a/crates/bitcoind_rpc/src/bip158.rs
+++ b/crates/bitcoind_rpc/src/bip158.rs
@@ -6,18 +6,18 @@
 //! [0]: https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki
 //! [1]: https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki
 
+use bdk_bitcoind_client;
+use bdk_bitcoind_client::corepc_types::model::GetBlockHeaderVerbose;
 use bdk_core::bitcoin;
 use bdk_core::CheckPoint;
 use bitcoin::BlockHash;
 use bitcoin::{bip158::BlockFilter, Block, ScriptBuf};
-use bitcoincore_rpc;
-use bitcoincore_rpc::{json::GetBlockHeaderResult, RpcApi};
 
 /// Type that returns Bitcoin blocks by matching a list of script pubkeys (SPKs) against a
 /// [`bip158::BlockFilter`](bitcoin::bip158::BlockFilter).
 ///
 /// * `FilterIter` talks to bitcoind via JSON-RPC interface, which is handled by the
-///   [`bitcoincore_rpc::Client`].
+///   [`bdk_bitcoind_client::Client`].
 /// * Collect the script pubkeys (SPKs) you want to watch. These will usually correspond to wallet
 ///   addresses that have been handed out for receiving payments.
 /// * Construct `FilterIter` with the RPC client, SPKs, and [`CheckPoint`]. The checkpoint tip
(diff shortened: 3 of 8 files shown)


ValuedMammal:
Concept ACK

crates/bitcoind_rpc/Cargo.toml
```diff
@@ -17,7 +17,7 @@ workspace = true
 
 [dependencies]
 bitcoin = { version = "0.32.0", default-features = false }
-bitcoincore-rpc = { version = "0.19.0" }
+bdk-bitcoind-client ={ git = "https://github.com/bitcoindevkit/bdk-bitcoind-client", rev = "06526db5c25047b61d29f81b25cfee7bd0f3abc8"}
```
oleonardolima: @tvpeter you can keep it referencing to master branch instead, while we don't have a release for bdk-bitcoind-client.

crates/bitcoind_rpc/Cargo.toml
```diff
@@ -26,9 +26,10 @@ bdk_testenv = { path = "../testenv" }
 bdk_chain = { path = "../chain" }
 
 [features]
-default = ["std"]
+default = ["std", "bitcoind_28_0"]
 std = ["bitcoin/std", "bdk_core/std"]
 serde = ["bitcoin/serde", "bdk_core/serde"]
+bitcoind_28_0 =["bdk-bitcoind-client/28_0"]
```
oleonardolima: nit: we should probably establish a policy here similar to bdk-bitcoind-rcp ones, that it support the latest 3 versions (exposed as features too).

crates/bitcoind_rpc/src/lib.rs
oleonardolima: @tvpeter can't we have this in `bdk_bitcoind_client` instead ?

oleonardolima (commented):
cACK 44bfbf9892c7fea37c4e0140a7e40c14b6c7925e

crates/bitcoind_rpc/tests/common/mod.rs
```diff
@@ -1,20 +1,19 @@
 use bdk_testenv::anyhow;
 use bdk_testenv::TestEnv;
 
-/// This trait is used for testing. It allows creating a new [`bitcoincore_rpc::Client`] connected
-/// to the instance of bitcoind running in the test environment. This way the `TestEnv` and the
-/// `Emitter` aren't required to share the same client. In the future when we no longer depend on
-/// `bitcoincore-rpc`, this can be updated to return the production client that is used by BDK.
+/// This trait is used for testing. It allows creating a new [`bdk_bictoind_client::Client`]
```
ValuedMammal: typo: "bictoind"
```suggestion
/// This trait is used for testing. It allows creating a new [`bdk_bitcoind_client::Client`]
```

crates/bitcoind_rpc/src/bip158.rs
```diff
@@ -188,10 +188,13 @@ impl From<core::num::TryFromIntError> for Error {
 }
 
 /// Whether the RPC error is a "not found" error (code: `-5`).
-fn is_not_found(e: &bitcoincore_rpc::Error) -> bool {
-    matches!(
-        e,
-        bitcoincore_rpc::Error::JsonRpc(bitcoincore_rpc::jsonrpc::Error::Rpc(e))
-        if e.code == -5
-    )
+fn is_not_found(e: &bdk_bitcoind_client::Error) -> bool {
```
ValuedMammal: We can use [`Error::is_not_found_error`](https://docs.rs/bdk_bitcoind_client/latest/bdk_bitcoind_client/enum.Error.html#method.is_not_found_error) here now.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -15,34 +15,33 @@
...
-/// The [`Emitter`] is used to emit data sourced from [`bitcoincore_rpc::Client`].
+/// The [`Emitter`] is used to emit data sourced from [`bdk_bitcoind_client::Client`].
 ///
 /// Refer to [module-level documentation] for more.
 ///
 /// [module-level documentation]: crate
-pub struct Emitter<C> {
-    client: C,
+pub struct Emitter {
+    client: Client,
```
ValuedMammal: I think the `Emitter` should hold a `&Client`. It doesn't need to consume or mutate the client. Unfortunate that we lose the `RpcClient` trait, since this is now less flexible in the types it accepts.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -15,34 +15,33 @@
...
-    /// The block result returned from rpc of the last-emitted block. As this result contains the
-    /// next block's block hash (which we use to fetch the next block), we set this to `None`
-    /// whenever there are no more blocks, or the next block is no longer in the best chain. This
-    /// gives us an opportunity to re-fetch this result.
-    last_block: Option<bitcoincore_rpc_json::GetBlockResult>,
+    /// The block header result returned from rpc of the last-emitted block. As this result
+    /// contains the next block's block hash (which we use to fetch the next block), we set
+    /// this to `None` whenever there are no more blocks, or the next block is no longer in the
+    /// best chain. This gives us an opportunity to re-fetch this result.
+    last_block: Option<corepc_types::model::GetBlockHeaderVerbose>,
```
ValuedMammal, line 44: I think the equivalent of GetBlockResult is GetBlockVerboseOne, but this seems to work as well.

crates/bitcoind_rpc/src/lib.rs
ValuedMammal, line 1: Now that all heights are `u32` we can remove the unneeded `as _` casts (I counted 4).

crates/bitcoind_rpc/tests/test_filter_iter.rs
```diff
@@ -22,7 +21,8 @@ fn testenv() -> anyhow::Result<TestEnv> {
 fn filter_iter_matches_blocks() -> anyhow::Result<()> {
     let env = testenv()?;
     let addr = ClientExt::get_rpc_client(&env)?
-        .get_new_address(None, None)?
+        .call::<String>("getnewaddress", &[])?
+        .parse::<Address<_>>()?
```
ValuedMammal: I think this should be `env.bitcoind.client.new_address()`.

crates/bitcoind_rpc/Cargo.toml
```diff
@@ -26,9 +26,12 @@ bdk_testenv = { path = "../testenv" }
 bdk_chain = { path = "../chain" }
 
 [features]
-default = ["std"]
+default = ["std", "bitcoind_28_0"]
```
ValuedMammal: To be consistent with bdk-bitcoind-client's default feature I guess it can be `bitcoind_30_0`.

crates/bitcoind_rpc/Cargo.toml
```diff
@@ -17,7 +17,7 @@ workspace = true
 
 [dependencies]
 bitcoin = { version = "0.32.0", default-features = false }
-bitcoincore-rpc = { version = "0.19.0" }
+bdk_bitcoind_client ={ version = "0.1.0", default-features = false }
```
ValuedMammal: spacing nit
```suggestion
bdk_bitcoind_client = { version = "0.1.0", default-features = false }
```

crates/bitcoind_rpc/Cargo.toml
```diff
@@ -26,9 +26,12 @@ bdk_testenv = { path = "../testenv" }
 bdk_chain = { path = "../chain" }
 
 [features]
-default = ["std"]
+default = ["std", "bitcoind_28_0"]
 std = ["bitcoin/std", "bdk_core/std"]
 serde = ["bitcoin/serde", "bdk_core/serde"]
+bitcoind_28_0 =["bdk_bitcoind_client/28_0"]
```
ValuedMammal: nit
```suggestion
bitcoind_28_0 = ["bdk_bitcoind_client/28_0"] 
```

examples/example_bitcoind_rpc_polling/src/main.rs
```diff
@@ -122,7 +123,7 @@ fn main() -> anyhow::Result<()> {
                 network,
                 |rpc_args, tx| {
                     let client = rpc_args.new_client()?;
-                    client.send_raw_transaction(tx)?;
+                    let _txid: Txid = client.call("sendrawtransaction", &[json!(tx)])?;
```
ValuedMammal: This is sending the Transaction object as a JSON value, whereas `sendrawtransaction` expects a hex-encoded raw transaction string - `json!(serialize_hex(tx))`.

crates/bitcoind_rpc/Cargo.toml
ValuedMammal: Should the default feature be `bitcoind_30_0` to match the default in bdk-bitcoind-client?

crates/bitcoind_rpc/examples/filter_iter.rs
ValuedMammal, line 1: I get this error when running the `filter_iter` example, don't know if it's something about my OS. I noticed during the transition from `bitcoincore_rpc` to `bdk_bitcoind_client` using `bitreq`. FWIW I don't have the same problem with `SimpleHttpTransport` because it can create a fresh tcp socket on the fly.

```sh
Matched block 205061
Matched block 208441
Error: JSON-RPC error: transport error: bitreq: Can't assign requested address (os error 49)
```

crates/bitcoind_rpc/Cargo.toml
tvpeter: For now, this is not possible because we need to update corepc-node in testenv to 30_0, which is not yet released. Even updating to corepc-node_29_0 will have to be done in a separate PR because some types will have to be updated too

crates/bitcoind_rpc/examples/filter_iter.rs
tvpeter: I have noticed this randomly when running tests. Sometimes it passes, other times it fails.

evanlinjin:
Thanks for this PR.

Before I give it a review, can I please get a bit more context to this [Discord message](https://discord.com/channels/753336465005608961/753367451319926827/1497285328166191205)?

In this message, it claims that `bdk_bitcoind_client` is only meant to support the features we need for BDK. Where else are we talking to `bitcoind` other than via `bdk_bitcoind_rpc`?

Sounds like it should NOT be a public crate at all.

Additionally, if the goal is to "support features that we need for BDK", one feature that is missing is the ability to support async IO. The `jsonrpc` crate that is used internally for `bdk_bitcoind_rpc` does not support this at all. To elegantly support blocking and async we need some sort of sans-io architecture. I've approached this in [`electrum_streaming_client`](https://github.com/bitcoindevkit/electrum_streaming_client), however, I'm not sure if that is the best way about it - would appreciate a look and some opinions.

cc. @notmandatory

notmandatory:
As a little recap from our discussion yesterday (please correct me if I've missed something), there are two ways to go with this crate:

1. move it into a private module in the `bdk_bitcoind_rpc` crate since it's not intended for use outside of with `bdk`.
2. keep it as it's own crate for use as a dependency of `bdk_bitcoind_rpc` but remove `bdk` from the name to be more inline with other client crates in the bitcoindevkit org. So something like `bitcoind-rpc-client` or `bitcoind-client` but then need to make sure that name is available on crates.io and need to keep it clear in the README it's only going to implement RPC functions and other features needed to support `bdk`. 

The advantage I see of option 1 is it avoids any confusion about the purpose of the client and simplifies release/publishing. 

Option 2 could be useful if other related projects like LDK want to use the same bitcoind rpc client for L1 and L2 but they'd need to be OK with exactly the same features `bdk` needs.

crates/bitcoind_rpc/examples/filter_iter.rs
ValuedMammal: I realized it's less of an issue for this PR since you can opt out of bitreq using `Client::with_transport`. Still interested to know if it can or should be addressed in bitreq library 🤔

tvpeter:
Closing this PR in favour of #2244

PR #2120, opened 2026-02-14
fix(chain): forward `confirmation_height_upper_bound` in `Anchor` implementation for `&A`

evanlinjin:
### Description

The blanket `Anchor` impl for `&A` was missing the `confirmation_height_upper_bound` method, causing it to fall back to the default implementation instead of delegating to the inner type.

### Changelog notice

```md
Fixed:

- The `Anchor::confirmation_height_upper_bound` impl was missing for `&A`, causing it to fallback to the default impl.
```

#### All Submissions:

#### Bugfixes:

~~* [ ] I've added tests to reproduce the issue which are now passing~~
~~* [ ] I'm linking the issue being fixed by this PR~~

### DIFF
--- a/crates/chain/src/tx_data_traits.rs
+++ b/crates/chain/src/tx_data_traits.rs
@@ -81,6 +81,10 @@ impl<A: Anchor> Anchor for &A {
     fn anchor_block(&self) -> BlockId {
         <A as Anchor>::anchor_block(self)
     }
+
+    fn confirmation_height_upper_bound(&self) -> u32 {
+        <A as Anchor>::confirmation_height_upper_bound(self)
+    }
 }
 
 impl Anchor for BlockId {


ValuedMammal (approved):

Dmenec (changes requested):
approach ACK

replacing `&self` with `self` should fix the Clippy lint that is breaking CI

```rust
fn confirmation_height_upper_bound(&self) -> u32 {
    <A as Anchor>::confirmation_height_upper_bound(self)
}

Serious-Sam-Dev:
approach ACK, how the comment above says 

> approach ACK
> 
> replacing `&self` with `self` should fix the Clippy lint that is breaking CI
> 
> ```rust
> fn confirmation_height_upper_bound(&self) -> u32 {
>     <A as Anchor>::confirmation_height_upper_bound(self)
> }
> ```
only change need is to fix the clippy, all another test is ok

Dmenec (approved):
tACK 5d97e136ae89854c53add90d9da9131fec56c4a0

ValuedMammal:
ACK 5d97e136ae89854c53add90d9da9131fec56c4a0

PR #2130, opened 2026-02-24
fix(bdk_esplora): use `get_blocks_infos` instead of `get_blocks`

oleonardolima:
### Description

The `get_blocks` method from `esplora-client` has been deprecated on the latest release `v0.12.3`. I'm updating it to use newly added one: `get_blocks_infos` in order to fix the deprecation warning.

### Changelog notice

```
### Changed
- fix(bdk_esplora): use `get_block_infos` instead of deprecated `get_blocks`
```

#### All Submissions:

* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)

#### New Features:

* [x] I've added tests for the new feature
* [x] I've added docs for the new feature

PR #2136, opened 2026-03-05
Fix `bdk_esplora` compilation error by bumping `esplora_client` version

evanlinjin:
### Description

I didn't realize #2053 didn't actually compile when I merged it. This PR fixes this by bumping `esplora_client` to `0.12.3` so that the `.get_block_infos` method is always available.

### Changelog

```md
Fixed:
- Bump `esplora_client` to `0.12.3` so that the `.get_block_infos` method is always available.
```

#### All Submissions:

### DIFF
--- a/crates/esplora/Cargo.toml
+++ b/crates/esplora/Cargo.toml
@@ -17,7 +17,7 @@ workspace = true
 
 [dependencies]
 bdk_core = { path = "../core", version = "0.6.1", default-features = false }
-esplora-client = { version = "0.12.1", default-features = false }
+esplora-client = { version = "0.12.3", default-features = false }
 async-trait = { version = "0.1.66", optional = true }
 futures = { version = "0.3.26", optional = true }
 serde_json = { version = "1.0", features = ["alloc"] }


luisschwab:
What do you mean by didn't compile? #2130 fixed the deprecated warning by using `get_block_infos`, then #2053 rebased on top of it and was merged. This is effectively a reversal of #2130.

evanlinjin:
> What do you mean by didn't compile? #2130 fixed the deprecated warning by using `get_block_infos`, then #2053 rebased on top of it and was merged. This is effectively a reversal of #2130.

Okay I think I found the problem. I'm assuming `get_block_infos` was added in a minor release of `esplora_client`.

I never did `cargo update` so `Cargo.lock` was stuck in the past. A minimum version of `esplora_client` should be enforced though.

luisschwab:
This is weird, because CI was complaining about using `get_blocks`. Perhaps we should only use `MAJOR.MINOR` or fix it on a patch version and bump manually.

oleonardolima:
> This is weird, because CI was complaining about using `get_blocks`. Perhaps we should only use `MAJOR.MINOR` or fix it on a patch version and bump manually.

CI complained because it does not keep the `Cargo.lock` and always resolves to the latest compatible version.

oleonardolima (approved):
ACK 1804699d372ade32825e444c9939626b94731596

Do mind to update the PR title/description before merging.

luisschwab:
Perhaps we should do it like `rust-bitcoin` does it: https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/tree/master/cargo-rbmt#lock-files

luisschwab (approved):
ACK 1804699d372ade32825e444c9939626b94731596

PR #2137, opened 2026-03-06
Replace `std::error::Error` with `core::error::Error`

### LINKED ISSUE #2132: Use `core::error::Error`
As mentioned in https://github.com/bitcoindevkit/bdk/pull/2115#discussion_r2834065195:

> [`core::error::Error` was stabilized in 1.81](https://doc.rust-lang.org/nightly/core/error/trait.Error.html), as MSRV is 1.85 now, I think we can remove this (and all the others too)

luisschwab:
Closes #2132

The `Error` trait is stable as of 1.81.0, so import it from `core` instead.

### DIFF
--- a/crates/bitcoind_rpc/src/bip158.rs
+++ b/crates/bitcoind_rpc/src/bip158.rs
@@ -173,8 +173,7 @@ impl core::fmt::Display for Error {
     }
 }
 
-#[cfg(feature = "std")]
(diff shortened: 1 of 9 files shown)


evanlinjin (approved):
ACK 2a92fd365ea7f45e56b3027989b5cdf79ef4773a

PR #2139, opened 2026-03-09
Refactor `ChainQuery` to be more flexible 

evanlinjin:
### Description

This PR introduces **median-time-past (MTP)** on canonical views. MTP needs block timestamps, and the existing task (`bdk_core::ChainQuery`) interface cannot expose them.

This PR changes the task (`bdk_core::ChainQuery`) interface to request block data by height, with the response now a generic `B`. A driver such as `LocalChain` can return `Header`s (which carry timestamps), and it resolves *every* requested height rather than collapsing the request to a single answer (as MTP needs all the blocks in a window, not just one). Previously the request was `Vec<BlockId>` and the response a single `ChainResponse = Option<BlockId>`: only the first in-chain block among the candidates.

The new response shape could have been done by keeping `next_query() -> Option<…>`, but the multi-stage canonicalization is much cleaner when the driving call returns a richer `TaskProgress` enum. This PR reshapes the protocol to a poll-based **`ChainTask`** whose `TaskProgress` expresses states `next_query` couldn't:

* `TaskProgress::Blocked` signals the task has issued every request it can but still needs outstanding data before it can finish. This is the hook a future streaming (non-blocking) driver would use to know when to wait.
* `TaskProgress::Advanced` lets a task write `poll` as one unit of work per call instead of wrapping its whole implementation in an internal loop. It also lets a driver observe progress in finer-grained increments.

Canonical views can then compute two MTP values, opted into via `with_mtp()` so callers who don't need it pay nothing:

* `CanonicalTxOut::prev_mtp` — the MTP of the block *preceding* the output's confirmation height. This is the reference BIP-68 evaluates relative timelocks against, matching Bitcoin Core's `CalculateSequenceLocks`, which uses `GetAncestor(std::max(nCoinHeight - 1, 0))->GetMedianTimePast()`. Note this is *not* the MTP at the confirmation height itself.
* `Canonical::tip_mtp` — MTP at the tip height, the BIP-113 reference for absolute timelocks.

MTP is on `CanonicalTxOut` only, deliberately — a relative timelock is evaluated against the coin being spent, and a transaction has no coin. It sits alongside `is_on_coinbase` for the same reason: spendability data bundled onto the output that coin selection consumes.

Both values are `Option<u32>`, and `None` means *unknown*: MTP medians an 11-block window, so it is only `Some` when the driver resolves all eleven heights. A `LocalChain` built by syncing keeps a handful of checkpoints rather than every block, so it yields `None` until the electrum/esplora clients fetch the preceding window per confirmed height (follow-up PRs). An output whose transaction is confirmed only transitively is also `None`, since that transaction's own confirmation height isn't known. `None` must never be read as an absent or satisfied timelock.

### Notes to the reviewers

- **Naming.** In the current `master`, `LocalChain::canonicalize` is the *generic task runner* and `LocalChain::canonical_view` is the *high-level helper*. This naming is confusing. In this PR, the runner becomes `run_task` and the helper becomes `canonicalize`.
- **Relationship to `CheckPoint::median_time_past`.** This does not replace that method. `CheckPoint::median_time_past` computes MTP for a height directly from the in-memory `LocalChain` checkpoint you already hold. Canonical-view computes MTP by sourcing its blocks through `ChainTask` / `BlockQueries` so it is driver-agnostic (works for any chain source, not just `LocalChain<Header>`) and attaches the result to each `CanonicalTx` and the tip, rather than making the caller walk checkpoints per height.
- **`TaskProgress::Blocked` has no in-tree consumer yet — this is intentional.** It exists for a future *streaming* driver (e.g. a Bitcoin Core RPC driver) that issues fetches without blocking: `Query` says "launch I/O", `Blocked` says "block until some lands". The only current driver (`LocalChain::run_task`) is synchronous: it resolves every height a task announces via `Query` before polling again, so a task that announces what it needs never has anything outstanding and never blocks. `run_task` therefore treats `Blocked` as a contract violation and panics, naming it — a task that reaches that arm is waiting on heights it never announced, which would also hang a streaming driver, since nothing ever launched those fetches. A comment in `CanonicalTask` documents the intended "anchored-leaves" batching optimization for the future driver.
- The shared anchor-checking logic used by both canonicalization phases is extracted into `BlockQueries::resolve_candidates` (returning `BlockCandidateResolution`), decoupled from `Anchor` by taking the candidates as `(payload, BlockId)` pairs.
- **Anchor semantics.** This PR assumes #2298 — that an anchor names the block which *confirmed* the transaction, so at most one of a transaction's anchors can be in the best chain at a time. `resolve_candidates` therefore returns the first in-chain candidate it finds and does not tie-break: under that assumption there is no tie to break. If #2298 does not land as written, and several anchors can legitimately be in-chain at once, this needs to resolve every candidate height before deciding and pick among them deterministically — otherwise the winner depends on which heights an earlier call happened to cache.

### Changelog notice

**`bdk_core`**
- Changed: the `ChainQuery` trait is reshaped into a poll-based `ChainTask` trait with a `TaskProgress` enum (`Advanced` / `Query` / `Blocked` / `Done`), generic over block type `B` (default `BlockHash`). Drivers now fetch block *data* by height instead of resolving candidate `BlockId`s.
- Added: `BlockQueries`, a request/resolve tracker, with `resolve_candidates` and the `BlockCandidateResolution` enum.
- Added: `median_time_past(height, block_time)`, a free helper computing the BIP-113 median-time-past with the timestamp source supplied by closure. `CheckPoint::median_time_past` and the canonical-view MTP computation both delegate to it.
- Removed: the `ChainRequest` / `ChainResponse` type aliases.

**`bdk_chain`**
- Added: median-time-past on canonical views — `CanonicalTxOut::prev_mtp`, the BIP-68 MTP at `confirmation_height - 1`, plus `Canonical::tip_mtp`, the BIP-113 MTP at the tip. Opted into via `CanonicalViewTask::with_mtp()` and `LocalChain::canonicalize_with_mtp()`.
- Changed: `CanonicalTask` and `CanonicalViewTask` now implement the poll-based `ChainTask` (previously `ChainQuery`) and take a generic block-type parameter `B`.
- Changed: `LocalChain::canonicalize` (the generic task runner) is renamed to `LocalChain::run_task`, and `LocalChain::canonical_view` (the high-level helper) is renamed to `LocalChain::canonicalize`.
- Fixed: `ChainPosition::Unconfirmed::last_seen` is now reported for a transaction that was seen in the mempool, evicted, and then swept canonical as the ancestor of a stale-anchored transaction. It was forced to `None`, which gave `first_seen: Some(_), last_seen: None`.

#### All Submissions:

#### New Features:

#### Bugfixes:


oleonardolima:
@evanlinjin the PR dependency is solved, #2038 landed on master! 🚀 🚀
You can rebase this one, though I think just resetting and cherry-picking the last commits will be easier (sorry for the huge conflicts 😅).

evanlinjin:
@oleonardolima We should have done this in one go in #2038. Now reviewers need to review 2 large refactor PRs. 😢

oleonardolima:
> @oleonardolima We should have done this in one go in #2038. Now reviewers need to review 2 large refactor PRs. 😢

Oh, agree! Though #2038 was long overdue and getting too convoluted, I think having this follow-up won't be that bad.

I'll make this review my top priority, let me know when it's ready.

evanlinjin:
@oleonardolima ready for review

crates/chain/src/canonical_view_task.rs
```diff
@@ -29,195 +33,274 @@ enum ViewStage {
...
+                    })
+                }
+                CanonicalReason::Anchor { anchor, descendant } => ChainPosition::Confirmed {
+                    anchor: anchor.clone(),
+                    transitively: *descendant,
+                },
+                CanonicalReason::ObservedIn { observed_in, .. } => ChainPosition::Unconfirmed {
+                    first_seen: tx_node.first_seen,
+                    last_seen: match observed_in {
+                        ObservedIn::Block(_) => tx_node.last_seen,
```
noahjoeris, line 288: It should be `ObservedIn::Block(_) => None` right?

If it re-enters into mempool it will be `ObservedIn::Mempool`

crates/chain/src/canonical_view_task.rs
```diff
@@ -29,195 +33,274 @@ enum ViewStage {
...
-                    CanonicalReason::ObservedIn { observed_in, .. } => match observed_in {
-                        ObservedIn::Mempool(last_seen) => ChainPosition::Unconfirmed {
-                            first_seen: tx_node.first_seen,
-                            last_seen: Some(*last_seen),
-                        },
-                        ObservedIn::Block(_) => ChainPosition::Unconfirmed {
-                            first_seen: tx_node.first_seen,
-                            last_seen: None,
+            let pos = match reason {
+                // Since the canonicalization algorithm processes assumed txs first (and therfore
```
noahjoeris, line 252:
```suggestion
                // Since the canonicalization algorithm processes assumed txs first (and therefore
```

crates/chain/src/local_chain.rs
```diff
@@ -471,6 +397,99 @@ where
...
+            .map(|cp| cp.hash() == block.hash)
+    }
+
+    /// Drive a [`ChainTask<D>`] to completion against this chain.
+    ///
+    /// This method processes any type implementing [`ChainTask<D>`], handling all its requests
+    /// against this chain and returning the task's output. The chain responds with its data
+    /// type `D`, so `LocalChain<Header>` responds with headers and `LocalChain<BlockHash>`
+    /// responds with hashes.
+    pub fn run_task<Q>(&self, mut task: Q) -> Q::Output
```
noahjoeris, line 422: nit: should we now rename the generic `Q` to `T`?

noahjoeris (commented):
Looking good overall. I left a few minor comments

crates/chain/src/canonical_view_task.rs
evanlinjin: ~~That's a good point and worth discussing now that you've mentioned it. The PR's behavior is the same behavior that is currently on `master`. Let's create a ticket and discuss there.~~

Edit: Looks like I did change the behavior.

evanlinjin: Split out to #2300, as discussed.

While writing it up I checked whether this actually differed from `master`, and it did: `master` has an explicit `last_seen: None` here, and this PR had silently changed it to `tx_node.last_seen`. That was unintended in a refactor, so I have reverted the arm to `None` — the PR is now behaviour-preserving and the semantic question lives in #2300.

Worth noting for that discussion: the two are not always equal. `txids_by_descending_last_seen` filters out evicted txs, so an evicted tx with a `last_seen` never reaches the `SeenTxs` stage, but it can still be swept transitively as an ancestor of a leftover tx and land in this arm with `tx_node.last_seen == Some(_)`. So your reasoning holds for the ordinary path, with that one hole.

crates/chain/src/canonical_view_task.rs
evanlinjin: Fixed, thanks — folded into the commit that introduced the line rather than added as a follow-up.

crates/chain/src/local_chain.rs
evanlinjin: Done — `run_task<Q>` is now `run_task<T>`. `Q` was left over from when the trait was `ChainQuery`; `T` for `ChainTask` reads better. Folded into the renaming commit.

crates/chain/src/canonical_view_task.rs
evanlinjin: Following up — I dug into this properly and your instinct pointed at something real, but the conclusion lands the other way. `tx_node.last_seen` is correct and I have kept it, with a test.

The field's own doc decides it:

> This value will be `None` if the chain data was never seen in the mempool **and** only seen in a conflicting chain.

Two conditions, joined by "and". You are right that a tx which re-entered the mempool would be `ObservedIn::Mempool` — for the ordinary path `ObservedIn::Block` does mean "no mempool sighting". The hole is eviction: `TxGraph::txids_by_descending_last_seen` filters evicted txs out, so an evicted tx never reaches the mempool stage, yet it can still be swept transitively by `mark_canonical` as the ancestor of a stale-anchored tx. It then lands in this arm while genuinely having a `last_seen`.

Forcing `None` there gave `first_seen: Some(t), last_seen: None` — "first seen at t, never seen" — because `first_seen` is passed through on exactly the same terms one line above. That inconsistency is on `master` today; it is just unreachable without the evicted-ancestor path, which nothing tested.

So this is now a deliberate fix in its own commit rather than a silent change inside the refactor, with `test_evicted_ancestor_of_stale_anchored_tx_keeps_last_seen` covering it. #2300 is closed with the reasoning. Thanks for pushing on it — the bug would not have been found otherwise.

PR #2142, opened 2026-03-09
fix(chain): return TxNode from CanonicalIter

### LINKED ISSUE #2051: The `CanonicalIter` should return the `TxNode`'s.
I'll create an issue to address this one in a follow-up PR.

_Originally posted by @oleonardolima in https://github.com/bitcoindevkit/bdk/pull/2029#discussion_r2386483114_

shubham-shinde-442:
Fixes: #2051 

### Description

This patch implements a TODO comment to have CanonicalIter return TxNode directly instead of just (Txid, Arc<Transaction>). This eliminates a redundant lookup in CanonicalView::new that was fetching the same transaction node data twice.

- Modified CanonicalIter to return TxNode<'g, Arc<Transaction>, A> instead of (Txid, Arc<Transaction>) in its iterator item type
- Simplified CanonicalView::new by removing redundant tx_graph.get_tx_node(txid) lookup and TODO comment
- Added integration test test_canonical_view_construction_with_tx_nodes to verify the change

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/canonical_iter.rs
+++ b/crates/chain/src/canonical_iter.rs
@@ -1,5 +1,5 @@
 use crate::collections::{HashMap, HashSet, VecDeque};
-use crate::tx_graph::{TxAncestors, TxDescendants};
+use crate::tx_graph::{TxAncestors, TxDescendants, TxNode};
 use crate::{Anchor, ChainOracle, TxGraph};
 use alloc::boxed::Box;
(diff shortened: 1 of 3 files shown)


evanlinjin:
Thanks for the attempt, but this is no longer needed as we will remove `CanonicalIter` in #2038.

PR #2145, opened 2026-03-10
perf(chain): refactor TxGraph.spends from BTreeMap to HashMap

### LINKED ISSUE #2026: The `TxGraph` spends field does not need to be a `BTreeMap`, could be a `HashMap` instead.
**Describe the enhancement**  

I was discussing with @evanlinjin about the `CanonicalIter`, and we noticed a possible optimization on the `TxGraph` spends field.

It's currently a `BTreeMap`, but it could be updated to become a `HashMap` instead. However, it would require reworking the `TxDescendants::populate_queue()` method.

**Use case**  

All fields in `TxGraph` are only either a `HashMap`, `HashSet` or `BTreeSet`.

**Impact**
- [x] Developer experience / maintainability

- [x] Yes
- [ ] No

- [ ] Electrum
- [ ] Esplora
- [ ] Bitcoin Core RPC

**Additional context**

shubham-shinde-442:
Resolves: #2026 

### Description

Refactors `TxGraph.spends` field from `BTreeMap` to `HashMap` for improved 
performance while maintaining all existing behavioral guarantees.

**Problem:**
The `spends` field was a `BTreeMap<OutPoint, HashSet<Txid>>`, providing O(log n) 
lookups. For transaction graph traversal this is unnecessary overhead.

**Solution:**
- Switch to `HashMap<OutPoint, HashSet<Txid>>` for O(1) average-case lookups
- Add secondary index `spend_vouts_by_txid: HashMap<Txid, BTreeSet<u32>>` to:
  - Provide O(1) lookup of all vouts spent by a given txid
  - Preserve deterministic vout iteration order (required by tx_spends() and 
    descendant walking)
- Update three methods to use the new index instead of range queries

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -133,7 +133,7 @@ use bitcoin::{Amount, OutPoint, SignedAmount, Transaction, TxOut, Txid};
 use core::fmt::{self, Formatter};
 use core::{
     convert::Infallible,
-    ops::{Deref, RangeInclusive},
+    ops::Deref,
 };
 
 impl<A: Ord> From<TxGraph<A>> for TxUpdate<A> {
@@ -171,7 +171,8 @@ impl<A: Anchor> From<TxUpdate<A>> for TxGraph<A> {
 #[derive(Clone, Debug, PartialEq)]
 pub struct TxGraph<A = ConfirmationBlockTime> {
     txs: HashMap<Txid, TxNodeInternal>,
-    spends: BTreeMap<OutPoint, HashSet<Txid>>,
+    spends: HashMap<OutPoint, HashSet<Txid>>,
+    spend_vouts_by_txid: HashMap<Txid, BTreeSet<u32>>,
     anchors: HashMap<Txid, BTreeSet<A>>,
     first_seen: HashMap<Txid, u64>,
     last_seen: HashMap<Txid, u64>,
@@ -191,6 +192,7 @@ impl<A> Default for TxGraph<A> {
         Self {
             txs: Default::default(),
             spends: Default::default(),
+            spend_vouts_by_txid: Default::default(),
             anchors: Default::default(),
             first_seen: Default::default(),
             last_seen: Default::default(),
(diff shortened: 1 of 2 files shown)


evanlinjin:
Does the performance improvement justify the added complexity?

lucasdbr05:
I reviewed the implementation, and indeed the `tx_spends` function had a complexity improvement from `O(log(n) + k)` to `O(k)` (where `k` is the number of spent vouts for the txid), in addition to the direct performance improvements from switching from a `BTreeMap` to a `HashMap` (insertion and lookup in amortized O(1)).

Furthermore, in terms of developer experience, I believe the performance improvement justifies the change made, as it keeps `spends` and `spend_vouts_by_txid` synchronized.

lucasdbr05 (approved):

evanlinjin:
Thanks @lucasdbr05 for the review — a few pushbacks though.

The O(log(n) + k) → O(k) improvement for `tx_spends` is real, but the framing
obscures what was actually traded away. The original `BTreeMap` gave us range queries for free as a consequence of `OutPoint`'s natural sort order. This PR recovers that same capability by introducing a secondary index that must be kept manually synchronized. The complexity didn't disappear — it moved from the data structure into our invariant maintenance.

Also worth noting: the reviewer's conclusion that "it keeps `spends` and `spend_vouts_by_txid` synchronized" is describing the requirement, not the guarantee. There's nothing enforcing that synchronization except the correctness of the insertion code.

A few remaining concerns before this can move forward:

1. **No benchmarks.** The asymptotic argument only holds at scale. For realistic wallet graphs the constant factors of hashing `OutPoint` and `Txid` twice per lookup may dominate. Real numbers would settle this.

2. **`DoubleEndedIterator` regression.** The original `tx_spends()` returned `impl DoubleEndedIterator` via `BTreeMap::range`. The new implementation uses `flat_map`, which doesn't preserve that bound. Was this intentional? If the return type was quietly downgraded it's a breaking change.

evanlinjin:
Closing due to unresponsive author.

PR #2146, opened 2026-03-11
`chain 0.23.x`: Fix `ChainPosition` ordering

evanlinjin:
## Description

Back-ported from #2074 
Depends on #2148

This PR fixes incorrect ordering of `ChainPosition` that affected how wallet transactions are displayed. Previously, unconfirmed transactions that were never seen in the mempool would incorrectly appear before transactions with mempool timestamps due to the derived `Ord` implementation treating `None` as less than `Some(_)`.

### Problem

The derived `Ord` implementation for `ChainPosition::Unconfirmed` caused incorrect transaction ordering:
- `Unconfirmed { first_seen: None, last_seen: None }` (never in mempool)
- Would appear **before** `Unconfirmed { first_seen: Some(10), last_seen: Some(10) }` (seen in mempool)

This resulted in a confusing wallet display where transactions never broadcast would appear before pending mempool transactions.

### Solution

Implemented manual `Ord` and `PartialOrd` traits for `ChainPosition` with the correct ordering:

1. **Confirmed transactions** (earliest first by block height)
2. **Unconfirmed transactions seen in mempool** (ordered by `first_seen`)
3. **Unconfirmed transactions never seen** (these come last)

Additional ordering rules:
- At the same confirmation height, transitive confirmations come before direct confirmations (as they may actually be confirmed earlier)
- Tie-breaking uses transaction IDs for deterministic ordering

## Changelog notice

```md
### Fixed
- Fixed `ChainPosition` ordering so unconfirmed transactions never seen in mempool appear last instead of first

### Changed
- `ChainPosition`, `CanonicalTx`, and `FullTxOut` now require `A: Ord` for their `Ord` implementations
- Simplified `FullTxOut` ordering to only use essential fields (chain_position, outpoint, spent_by)
```

### All Submissions:

### Bugfixes:

### DIFF
--- a/crates/chain/src/chain_data.rs
+++ b/crates/chain/src/chain_data.rs
@@ -96,6 +96,16 @@ impl<A: Ord> Ord for ChainPosition<A> {
     fn cmp(&self, other: &Self) -> core::cmp::Ordering {
         use core::cmp::Ordering;
 
+        /// Compares options where `None` is greater than `Some` (sorts last).
+        fn cmp_none_last<T: Ord>(t1: &Option<T>, t2: &Option<T>) -> Ordering {
+            match (t1, t2) {
+                (None, None) => Ordering::Equal,
+                (None, Some(_)) => Ordering::Greater,
+                (Some(_), None) => Ordering::Less,
+                (Some(t1), Some(t2)) => t1.cmp(t2),
+            }
+        }
+
         match (self, other) {
             // Both confirmed: compare by anchor first
             (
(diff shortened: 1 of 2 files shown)


luisschwab (approved):
ACK c3e960f84b390c4bcddc71bdf73c0ec9f5bb193b

notmandatory (approved):
ACK c3e960f84b390c4bcddc71bdf73c0ec9f5bb193b

I ran it through the LLM 🤖 reviewer also and its only comments were about missing edge case tests with invalid ChainPosition data. But preventing invalid data should already be covered in other tests.

PR #2181, opened 2026-04-16
feat: unify full_scan and sync into a single scan API

### LINKED ISSUE #2057: Unintuitive and undocumented `full_scan` behavior
[Book of BDK](https://bookofbdk.com/cookbook/syncing/full-scan-vs-sync/?h=sync) has a overview of the expected sync behavior 

Image

However `full_scan` seems to completely ignore the revealed addresses and goes by the gap only.  This means that if there is a large gap it will not discover any transactions even though the addresses are already revealed.  A `sync` will discover those transactions, such that for a "full scan" one has to do  `full_scan` _and_ `sync` .

Thanks to  @design-rrr for finding this behavior.

- Reproducible in bdkpython 2.0.0

#### Why not do `sync`  most of the time?
- onchain wallets can have multiple instances running, that reveal and use scripts.  That means that the wallet always has to sync at least the gap limit ahead to discover txs that other instances used. 

#### What to do about it?
- Was this behavior intended? 
  - But then to do a full scan, one has to hit the server twice...  --> inefficient

@thunderbiscuit  @ValuedMammal

noahjoeris:
### Description

Proposal: unify `full_scan` and `sync` into a single `scan` API. `scan(...)` combines explicit sync and keychain discovery in one request/call, so callers no longer need to stitch together `sync(...)` + `full_scan(...)`. Open to naming/API changes. I also tried to increase DX a bit.

Fixes #2057. The old `spks_from_indexer(...)` used an unbounded iterator starting at index 0. A single stop-gap scan therefore covered both revealed and unrevealed scripts in one pass, which could cause it to terminate before reaching a higher revealed address. `test_scan_revealed_scripts_beyond_stop_gap` covers this now.

`scan` splits the old behavior along the right seam:
- `revealed_spks_from_indexer(...)`: explicit sync of the already revealed range `0..=last_revealed`
- `discover_from_indexer(...)`: discovery starting at `last_revealed + 1`, with `stop_gap` applied only to the unrevealed part

Other changes:
- The new builder uses `discover_keychain(...)` instead of `spks_for_keychain(...)` (renamed for clarity)
- `stop_gap` moved onto the builder. Defaults to 20 (matches BIP44/Electrum/Sparrow defaults). I assume most callers use the default unless they need something custom.
- Unified inspect/progress API via `ScanItem` / `ScanProgress`.
- `full_scan(...)` and `sync(...)` kept in place for gradual migration.

#### API changes

| New | Old / change |
|-----|--------------|
| `scan(...)` | `sync(...)` + `full_scan(...)`; new unified entrypoint for explicit sync and discovery |
| `ScanRequest` / `ScanResponse` | `SyncRequest`/`SyncResponse` + `FullScanRequest`/`FullScanResponse`; new unified request/response types |
| `ScanRequestBuilder::discover_keychain(...)` + `stop_gap(...)` | `FullScanRequestBuilder::spks_for_keychain(...)` + `full_scan(..., stop_gap, ...)`; same discovery input, but `stop_gap` moved onto the builder |
| `ScanRequestBuilderExt::revealed_spks_from_indexer(...)` + `discover_from_indexer(...)` | `FullScanRequestBuilderExt::spks_from_indexer(...)`; split into explicit sync of revealed scripts plus discovery starting at `last_revealed + 1` |

### Notes to the reviewers

Waiting for concept (N)ACK before filling in more tests, example updates, or deprecations.

**Other findings while working on this:**
- Esplora clamps `stop_gap` to 1 via `.max(1)` but electrum did not. I aligned electrum here.
- Due to `parallel_requests` in Esplora, the scan can overshoot `stop_gap`. Same behavior as the previous `full_scan`, no harm, keeping as-is.

**Limitations:**
- `expected_spk_txids` currently applies only to explicitly synced scripts (`spks` / `spks_with_indexes`), not to discovery scripts (`discover_keychain` / `discover_from_indexer`). This matches the current design where expected txids are associated with known scripts rather than the unrevealed discovery suffix. Extending to discovery is possible if we want it to be agnostic later.

**Follow-ups (post concept-ACK):**
- Update examples to use `scan`
- Mark `sync` / `full_scan` as `#[deprecated]`
- Consider migration helpers to ease adoption
- More test coverage

### Changelog notice

- Added unified `scan` API in `bdk_core::spk_client`: `ScanRequest`, `ScanResponse`, `ScanRequestBuilder`, `ScanRequestBuilderExt`, `ScanItem`, `ScanProgress`.
- Added `BdkElectrumClient::scan`, `EsploraExt::scan`, and `EsploraAsyncExt::scan`.
- Fixed stop-gap terminating before reaching higher already-revealed scripts (#2057).

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/indexer/keychain_txout.rs
+++ b/crates/chain/src/indexer/keychain_txout.rs
@@ -5,7 +5,7 @@ use crate::{
     alloc::boxed::Box,
     collections::*,
     miniscript::{Descriptor, DescriptorPublicKey},
-    spk_client::{FullScanRequestBuilder, SyncRequestBuilder},
+    spk_client::{FullScanRequestBuilder, ScanRequestBuilder, SyncRequestBuilder},
     spk_iter::BIP32_MAX_INDEX,
     spk_txout::SpkTxOutIndex,
     DescriptorExt, DescriptorId, Indexed, Indexer, KeychainIndexed, SpkIterator,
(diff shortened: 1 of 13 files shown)


evanlinjin (changes requested):
Concept NACK

First off, thanks for working on this. However, I think `full_scan` and `sync` should remain separate APIs. They have different inputs and use cases, and merging them adds complexity without adding capability.

The underlying issue (#2057) is valid, but the fix should be targeted: give `full_scan` awareness of `last_revealed` so the stop gap only applies beyond that point. See https://github.com/bitcoindevkit/bdk/issues/2057#issuecomment-4266722903.

noahjoeris:
Appreciate the feedback! 👍  Yep I can do that too. 

My idea was to address this comment by @ValuedMammal https://github.com/bitcoindevkit/bdk/issues/2057#issuecomment-3379435363 which @luisschwab also upvoted.

PR #2182, opened 2026-04-19
fix: full_scan covers revealed range before applying stop_gap

### LINKED ISSUE #2057: Unintuitive and undocumented `full_scan` behavior

noahjoeris:
Fixes #2057. `full_scan` now always covers the revealed range before applying the stop_gap

### Notes to the reviewers

Replaces #2181. That PR unified full_scan and sync into one path; per @evanlinjin's feedback we're keeping them separate and doing the minimal fix here instead

### Changelog notice

Fixed: full_scan now always scans the full revealed range before applying stop_gap past last_revealed.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/core/src/spk_client.rs
+++ b/crates/core/src/spk_client.rs
@@ -456,6 +456,18 @@ impl<K: Ord, D> FullScanRequestBuilder<K, D> {
         self
     }
 
+    /// Record the last revealed script pubkey `index` for a given `keychain`.
+    ///
+    /// `full_scan` covers `0..=index` for this keychain; `stop_gap`
+    /// applies only to indices past `index`. Keychains without a recorded last revealed
+    /// index fall back to applying `stop_gap` from index 0.
+    /// Users working with a `KeychainTxOutIndex` usually don't call this directly,
+    /// `spks_from_indexer` (from `bdk_chain`) populates it automatically.
+    pub fn last_revealed_for_keychain(mut self, keychain: K, index: u32) -> Self {
+        self.inner.last_revealed.insert(keychain, index);
+        self
+    }
+
     /// Set the closure that will inspect every sync item visited.
     pub fn inspect<F>(mut self, inspect: F) -> Self
     where
@@ -483,6 +495,7 @@ pub struct FullScanRequest<K, D = BlockHash> {
     start_time: u64,
     chain_tip: Option<CheckPoint<D>>,
     spks_by_keychain: BTreeMap<K, Box<dyn Iterator<Item = Indexed<ScriptBuf>> + Send>>,
+    last_revealed: BTreeMap<K, u32>,
     inspect: Box<InspectFullScan<K>>,
 }
 
@@ -507,6 +520,7 @@ impl<K: Ord + Clone, D> FullScanRequest<K, D> {
                 start_time,
                 chain_tip: None,
                 spks_by_keychain: BTreeMap::new(),
+                last_revealed: BTreeMap::new(),
                 inspect: Box::new(|_, _, _| ()),
             },
         }
@@ -539,6 +553,14 @@ impl<K: Ord + Clone, D> FullScanRequest<K, D> {
     /// List all keychains contained in this request.
     pub fn keychains(&self) -> Vec<K> {
         self.spks_by_keychain.keys().cloned().collect()
+    }
+
+    /// Get the last revealed script pubkey index for `keychain` (if set).
+    ///
+    /// Chain sources use this to scan `0..=last_revealed` before applying
+    /// `stop_gap` to further discovery.
+    pub fn last_revealed(&self, keychain: &K) -> Option<u32> {
+        self.last_revealed.get(keychain).copied()
     }
 
     /// Advances the full scan request and returns the next indexed [`ScriptBuf`] of the given
--- a/crates/core/src/spk_client.rs
+++ b/crates/core/src/spk_client.rs
@@ -456,6 +456,18 @@ impl<K: Ord, D> FullScanRequestBuilder<K, D> {
         self
     }
 
+    /// Record the last revealed script pubkey `index` for a given `keychain`.
+    ///
+    /// `full_scan` covers `0..=index` for this keychain; `stop_gap`
+    /// applies only to indices past `index`. Keychains without a recorded last revealed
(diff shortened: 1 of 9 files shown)


crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -267,12 +270,14 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
     /// requested spk will be added to `tx_update`. Anchors of the aforementioned transactions are
     /// also included.
+    #[allow(clippy::too_many_arguments)]
     fn populate_with_spks(
         &self,
         start_time: u64,
         tx_update: &mut TxUpdate<ConfirmationBlockTime>,
         mut spks_with_expected_txids: impl Iterator<Item = (u32, SpkWithExpectedTxids)>,
         stop_gap: usize,
+        last_revealed: Option<u32>,
```
noahjoeris, line 280: Should we combine some parameters to avoid suppressing clippy::too_many_arguments? e.g.
```
struct StopGapExt {
  stop_gap: usize
  last_revealed Option<u32>
}
```

crates/core/src/spk_client.rs
evanlinjin, line 493: We now need to mention that fetching stops after stop-gap number of scripts ABOVE the last-revealed-index. Feel free to improve my wording here.

```suggestion
/// A client full scan iterates through all the scripts for the given keychains, fetching relevant
/// data until some stop gap number of scripts is found (above the last-revealed index) that have no
/// data. This operation is generally only used when importing or restoring previously used
/// keychains in which the list of used scripts is not known. The full scan process also updates the
/// chain from the given [`chain_tip`](FullScanRequestBuilder::chain_tip) (if provided).
```

crates/electrum/tests/test_electrum.rs
```diff
@@ -490,6 +492,47 @@ pub fn test_update_tx_graph_stop_gap() -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Test that `full_scan` always scans the revealed range before applying `stop_gap`.
+#[test]
+pub fn test_stop_gap_past_last_revealed() -> anyhow::Result<()> {
```
evanlinjin, line 502: Can we expand on this test (or write a new test) to ensure that the stop gap is enforced above last-revealed?

I.e. last_revealed = 10, stop_gap = 5, tx at index = 16 (should not be discovered).

crates/electrum/tests/test_electrum.rs
```diff
@@ -490,6 +492,47 @@ pub fn test_update_tx_graph_stop_gap() -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Test that `full_scan` always scans the revealed range before applying `stop_gap`.
```
evanlinjin:
```suggestion
/// Test that `full_scan` always scans the revealed range before applying `stop_gap`.
///
///
/// Given a single keychain with a tx at index 9, we scan with `stop_gap=3` and `last_revealed=9`.
/// We expect that the tx at index 9 to be returned by the `full_scan`.
```

crates/esplora/tests/common/mod.rs
```diff
@@ -1,14 +1,32 @@
...
-use bdk_core::bitcoin::ScriptBuf;
+use bdk_core::bitcoin::{Address, ScriptBuf};
+use std::str::FromStr;
 
 const PK_BYTES: &[u8] = &[
     12, 244, 72, 4, 163, 4, 211, 81, 159, 82, 153, 123, 125, 74, 142, 40, 55, 237, 191, 231, 31,
     114, 89, 165, 83, 141, 8, 203, 93, 240, 53, 101,
 ];
 
-#[allow(dead_code)]
```
evanlinjin, line 9: Can we have this change in a new commit?

The idiomatic way would be to have `#![allow(dead_code)]` on the top of the file instead of per-item - let's do that in that new commit. This is slightly out of scope, but would be good to get it sorted here imo.

crates/electrum/tests/test_electrum.rs
evanlinjin: Can we also have this in the esplora tests?

crates/electrum/tests/test_electrum.rs
evanlinjin: Also have this in the esplora tests, thanks.

evanlinjin (changes requested):
Thanks for this work. The approach is correct and minimal. Just some small things about docs and tests before ACK and merge.

noahjoeris:
Thanks for the review! Addressed all comments. Docs updated, tests expanded and the `#![allow(dead_code)]` refactor is in its own commit

evanlinjin (approved):
ACK 187b00747ab978a3bba21df5d14835ec31ed9fa3

thunderbiscuit:
@noahjoeris @evanlinjin I'm wondering if there is a way to backport this into chain 0.23.X so users of the 3.0 wallet could benefit from this. As is, this fix will only come in the 4.0 version of `Wallet`, probably a long way out.

evanlinjin:
@thunderbiscuit definitely possible. This is a non-breaking change. Need to create a minor release for `bdk_core` and subsequent minor releases for the chain sources.

thunderbiscuit:
Awesome. Also by the way, I wasn't asking for someone else to do the work for me haha, I just wanted to check if this was an easy lift or if it needed more thinking through.

I'm happy to do the PR.

PR #2186, opened 2026-04-22
Have MTP in checkpoint data and report missing blocks

evanlinjin:
### Description

Context: https://github.com/bitcoindevkit/bdk/issues/2076#issuecomment-4261182538

Carries median-time-past (MTP) on each checkpoint at the type level, and reworks `compute_mtp` so callers can drive fetch-retry loops when ancestor blocks are missing.

- `WithMtp<D>` wrapper: `CheckPoint<WithMtp<D>>` is a compile-time guarantee that MTP is present on every checkpoint in the chain. Code that needs MTP can take `CheckPoint<WithMtp<D>>` and read it via `CheckPoint::mtp()` without ever having to handle a "MTP couldn't be computed" case — the type system rules that out. Compare to the alternative of computing MTP on demand, which forces every consumer to deal with the failure path even when the chain source already has the data.
- `compute_mtp` returns `Result<u32, MissingBlocks>` instead of `Option<u32>`. `MissingBlocks` lists every height in the MTP window (`h-10..=h`) that the chain is missing, so callers — especially light-client chain sources like Electrum that don't expose MTP directly — can batch-fetch exactly the headers they need and retry.
- Rename `median_time_past` → `compute_mtp` to distinguish the chain-walking constructor from the `mtp()` accessor on `CheckPoint<WithMtp<D>>`.

### Notes to the reviewers

Breaking change, marked `feat(core)!`. Two API breaks:
- `median_time_past` → `compute_mtp`
- `compute_mtp` returns `Result<u32, MissingBlocks>` (was `Option<u32>`)

### Changelog notice

```md
Added:
- `WithMtp<D>` wrapper: `CheckPoint<WithMtp<D>>` guarantees at the type level that MTP is present for every checkpoint.
- `CheckPoint::mtp()` on `CheckPoint<WithMtp<D>>`.
- `MissingBlocks` error type.

Changed:
- `CheckPoint::median_time_past` renamed to `CheckPoint::compute_mtp`; return type changed from `Option<u32>` to `Result<u32, MissingBlocks>`.
```

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/core/src/checkpoint.rs
+++ b/crates/core/src/checkpoint.rs
@@ -120,6 +120,25 @@ impl<D: ToBlockHash> ToBlockHash for WithMtp<D> {
     }
 }
 
+/// Error returned by [`CheckPoint::compute_mtp`] when one or more required ancestor
+/// checkpoints are missing from the chain.
+///
+/// The `heights` field lists every height in the MTP window (`h-10..=h`) that the caller
+/// needs to fetch and insert before `compute_mtp` can succeed.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MissingBlocks {
+    /// Heights whose checkpoints were expected but not found in the chain.
+    pub heights: Vec<u32>,
+}
+
+impl fmt::Display for MissingBlocks {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "missing block data for heights {:?}", self.heights)
+    }
+}
+
+impl core::error::Error for MissingBlocks {}
+
 impl<D> PartialEq for CheckPoint<D> {
     fn eq(&self, other: &Self) -> bool {
         let self_cps = self.iter().map(|cp| cp.block_id());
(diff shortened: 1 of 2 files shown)


evanlinjin:
I'm wondering whether we should get rid of the `ToBlockTime` trait and implement block-time-dependent methods directly on `CheckPoint<Header>`. This approach is more aligned with the YAGNI principle. I'm personally not expecting many custom checkpoint data types. `WithMtp<Header>` seems like what the super majority of wallets will use.

The other approach is to go the other way and introduce a `ToBlockMTP` trait to this PR.

crates/core/src/checkpoint.rs
```diff
@@ -338,6 +392,16 @@ where
     }
 }
 
+impl<D> CheckPoint<WithMtp<D>> {
+    /// Returns the precomputed median-time-past (MTP) stored on this checkpoint.
+    ///
+    /// Unlike [`CheckPoint::compute_mtp`], this does not walk the chain — the MTP was computed
+    /// when the checkpoint was constructed and is guaranteed to be present by the type system.
```
notmandatory, line 399: Is it worth mentioning here that the MTP value returned here would be in accurate in the case of a reorg that happens after the MTP was calculated?

notmandatory (commented):
This generally looks like a reasonable performance optimization. Would blockchain clients be responsible for recomputing the recent mpc values in the case of a reorg?

notmandatory:
> I'm wondering whether we should get rid of the `ToBlockTime` trait and implement block-time-dependent methods directly on `CheckPoint<Header>`. This approach is more aligned with the YAGNI principle. I'm personally not expecting many custom checkpoint data types. `WithMtp<Header>` seems like what the super majority of wallets will use.
> 
> The other approach is to go the other way and introduce a `ToBlockMTP` trait to this PR.

Wouldn't any custom `CheckPoint<D>` type need to implement `D: ToBlockTime` anyway? so probably would be better to implement these directly on `CheckPoint` rather than this wrapper approach.

evanlinjin:
@notmandatory Thanks for the review!

> This generally looks like a reasonable performance optimization. Would blockchain clients be responsible for recomputing the recent mpc values in the case of a reorg?

The main intention behind the PR is not about performance (the previous PR description was misleading and I've updated it!). The idea is to enforce MTP value per checkpoint at compile time - if you use `CheckPoint<WithMtp<Header>>`, you have access to the method `.mtp()` that returns the MTP value (not wrapped in `Option`). If you have `CheckPoint<Header>`, you `calculate_mtp()` that returns an option (as the MTP time needs to be calculated from the previous 11 headers).

So the chain source must return the MTP value per block in order to be compatible with `WithMtp<X>` types. Bitcoind's RPC methods already do this. Electrum does not (as you are expected to have most of the headers and verify PoW). For `bdk_electrum`, we can artificially allow grabbing the MTP value by querying 11 blocks.

> Wouldn't any custom `CheckPoint<D>` type need to implement `D: ToBlockTime` anyway? so probably would be better to implement these directly on `CheckPoint` rather than this wrapper approach.

So currently, yes if they want the `calculate_mtp()` method to be available. Sorry, can you expand on this point please?

notmandatory:
> Wouldn't any custom CheckPoint<D> type need to implement D: ToBlockTime anyway? so probably would be better to implement these directly on CheckPoint rather than this wrapper approach.
> > So currently, yes if they want the calculate_mtp() method to be available. Sorry, can you expand on this point please?

High level what I was thinking it would be simpler to put `pub mtp: Option<u32>` in `CPInner` rather than having to make the `WithMtp` wrapper type. But I can see how this would be messier logic that your approach avoids by counting on the type system. So you can disregard my comment.

evanlinjin:
Coming back to this, I don't think having a compile-time guarantee of MTP per checkpoint makes sense for `LocalChain` as `LocalChain` is a sparse structure. Therefore, I'll be closing this PR.

Context: https://github.com/bitcoindevkit/bdk/issues/2076#issuecomment-4376769976

PR #2188, opened 2026-04-23
fix(electrum): verify txid of server-returned transactions

tnull:
### Description

An Electrum server could return an arbitrary transaction when `fetch_tx()` requests a specific txid. The returned transaction was cached and used without verifying that its computed txid matches the requested one.

Add a verification check that `tx.compute_txid() == txid` after fetching from the server, returning an error on mismatch. Include a unit test with a mock Electrum client that exercises both the mismatch rejection and the matching-txid happy path.

### DIFF
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -78,6 +78,12 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
         drop(tx_cache);
 
         let tx = Arc::new(self.inner.transaction_get(&txid)?);
+        let returned_txid = tx.compute_txid();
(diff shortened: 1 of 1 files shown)


evanlinjin (changes requested):
ConceptACK

I think the test is overkill - let's remove it. Otherwise, LGTM.

evanlinjin:
@tnull Let me know if you would like me to take over (I know you are busy).

tnull:
> @tnull Let me know if you would like me to take over (I know you are busy).

Dropped the test.

evanlinjin (approved):
ACK d101a0973bc19ac1b64809c0d9f0cd3516e263d8

PR #2189, opened 2026-04-25
Bump `esplora-client` to v0.13.0

luisschwab:
TBD

### DIFF
--- a/ci/pin-msrv.sh
+++ b/ci/pin-msrv.sh
@@ -9,8 +9,3 @@ set -euo pipefail
 
 # cargo clean
 # rustup override set 1.85.0
-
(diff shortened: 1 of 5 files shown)


oleonardolima:
ping @luisschwab
Could you update to use the latest esplora-client master ?
I think there's something broken that VM's mentioned in the last call.

luisschwab:
It seems `get_block_infos` (plural) is missing, we only have `get_block_info` (singular).

luisschwab:
It seems that `bitreq` doesn't properly support a WASM transport. We may need to hold off on bumping it here until that is addressed.

We can make a `0.13.1` point release once `bitreq` gets this sorted out.

cc @oleonardolima

PR #2195, opened 2026-04-29
fix(electrum): do not pick unindexed outputs for history lookup

zoedberg:
### Description

`BdkElectrumClient::populate_with_txids` queries each transaction's confirmation status by calling `script_get_history` on the script of one of its outputs. It currently picks the first output unconditionally. This breaks for transactions which first output is an `OP_RETURN`, because Electrum servers don't index `OP_RETURN` scripts and will return an empty history. This is a real-world scenario: protocols like RGB place an `OP_RETURN` commitment as the first output of every transaction.

### Notes to the reviewers

The fix selects the first output which script is not `OP_RETURN` or a `OP_FALSE OP_RETURN`. If a transaction has only `OP_RETURN`/`OP_FALSE OP_RETURN` outputs, we fall back to the script of any input's previous output to query history. The only case still skipped is a coinbase with all unindexed outputs, since coinbases have no parent to fall back on.

### Changelog notice

```
Fixed:
- `BdkElectrumClient::sync` now correctly retrieves confirmation status for transactions which first output is an `OP_RETURN` or `OP_FALSE OP_RETURN`
```

#### All Submissions:

### DIFF
--- a/crates/electrum/src/bdk_electrum_client.rs
+++ b/crates/electrum/src/bdk_electrum_client.rs
@@ -1,5 +1,9 @@
 use bdk_core::{
-    bitcoin::{block::Header, BlockHash, OutPoint, Transaction, Txid},
+    bitcoin::{
+        block::Header,
+        opcodes::{all::OP_RETURN, OP_FALSE},
+        BlockHash, OutPoint, Transaction, Txid,
+    },
     collections::{BTreeMap, HashMap, HashSet},
     spk_client::{
         FullScanRequest, FullScanResponse, SpkWithExpectedTxids, SyncRequest, SyncResponse,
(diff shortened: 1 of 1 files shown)


crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -443,12 +443,15 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
-                    let spk = tx
+                    // pick the first output Electrum will return history for
+                    let spk = match tx
                         .output
-                        .first()
-                        .map(|txo| &txo.script_pubkey)
-                        .expect("tx must have an output")
-                        .clone();
+                        .iter()
+                        .find(|txo| !txo.script_pubkey.is_op_return())
```
evanlinjin: We should also filter out `OP_FALSE OP_RETURN` since that's used in ephemeral anchors and is excluded from the ElectrumX index.

https://github.com/spesmilo/electrumx/blob/24865dc3a8ac8d360e467df704777c1bbef72706/src/electrumx/lib/script.py#L75-L77

evanlinjin (changes requested):
Good catch!

Also, can you rebase on master to fix CI.

crates/electrum/src/bdk_electrum_client.rs
zoedberg: you're right, sorry I missed that, fixed!

zoedberg:
@evanlinjin Fixed and rebased

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -443,12 +443,15 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
-                        .expect("tx must have an output")
-                        .clone();
+                    // pick the first output Electrum will return history for
+                    let spk = match tx.output.iter().find_map(|txo| {
+                        let script = &txo.script_pubkey;
+                        (!script.is_op_return() && !script.as_bytes().starts_with(&[0x00, 0x6a]))
+                            .then(|| script.clone())
+                    }) {
+                        Some(spk) => spk,
+                        None => continue,
```
evanlinjin, line 481: This means some txs will not be fetched even though they are provided by `txids` and they exist.

We may need to fetch a parent tx in order to obtain a spk that we can fetch history with.

evanlinjin (changes requested):

crates/electrum/src/bdk_electrum_client.rs
zoedberg: it makes sense, thanks for the suggestion. please check if the fix is ok now

oleonardolima (commented):
cACK 88b135468799104c27e6a64860a863ba036cfd18

crates/electrum/src/bdk_electrum_client.rs
```diff
@@ -443,12 +443,36 @@ impl<E: ElectrumApi> BdkElectrumClient<E> {
...
-                    let spk = tx
-                        .output
-                        .first()
-                        .map(|txo| &txo.script_pubkey)
-                        .expect("tx must have an output")
-                        .clone();
+                    // pick the first output Electrum will return history for
+                    let mut spk = tx.output.iter().find_map(|txo| {
+                        let script = &txo.script_pubkey;
+                        (!script.is_op_return() && !script.as_bytes().starts_with(&[0x00, 0x6a]))
```
oleonardolima: nit: you can also use the op code values from `bdk_chain::bitcoin::opcodes`.

zoedberg: @oleonardolima sure, done

oleonardolima (commented):
ACK 2e3d52e80fdc504c718426b9255a30fc24b7ceda

evanlinjin (approved):
ACK 2e3d52e80fdc504c718426b9255a30fc24b7ceda

notmandatory:
fixes #2249

PR #2205, opened 2026-05-17
feat: add start_height support to FilterIter

### LINKED ISSUE #2126: Add `start_height` parameter to `bitcoind_rpc::FilterIter::new`
From evanlinjin's [comment on `bdk_wallet#368`](https://github.com/bitcoindevkit/bdk_wallet/pull/368#issuecomment-3809170329):

> bitcoind_rpc::Emitter::new has a start_height parameter. This is used in the case the closest block that connects is below the start_height. So if the tip is reorged out, and we only have tip + genesis, we will start from start_height instead. To convey the wallet birthday with the checkpoint tip, while still taking into account reorgs, we can use tip_height - FINAL_DEPTH as the start_height parameter. FINAL_DEPTH can be as conservative as you want.
>
> bitcoind_rpc::FilterIter::new does not have such a parameter. We can add it?
>
> Then we provide docs as how to convey the wallet's birthday to the chain source using the checkpoint tip.

kafeelraza:
### Description

Add `start_height` support to `bitcoind_rpc::bip158::FilterIter::new`.

This allows filter scanning to recover from deep reorgs by resuming from a configured starting height when no common ancestor is found above that height.

### Notes to the reviewers

This PR:

* adds a `start_height` parameter to `FilterIter`
* updates recovery behavior for deep reorg handling
* adds/updates tests covering recovery from invalid checkpoints
* updates docs/comments describing the new behavior

Closes #2126

### Changelog notice

Added `start_height` support to `bitcoind_rpc::bip158::FilterIter::new` to improve deep reorg recovery behavior.

#### All Submissions:

* [x] I followed the contribution guidelines

#### New Features:

### DIFF
--- a/crates/bitcoind_rpc/examples/filter_iter.rs
+++ b/crates/bitcoind_rpc/examples/filter_iter.rs
@@ -52,7 +52,11 @@ fn main() -> anyhow::Result<()> {
     for (_, desc) in graph.index.keychains() {
         spks.extend(SpkIterator::new_with_range(desc, 0..SPK_COUNT).map(|(_, s)| s));
     }
-    let iter = FilterIter::new(&rpc_client, chain.tip(), spks);
+    const FINAL_DEPTH: u32 = 100;
+
+    let start_height = chain.tip().height().saturating_sub(FINAL_DEPTH);
+
+    let iter = FilterIter::new(&rpc_client, chain.tip(), start_height, spks);
 
     let start = Instant::now();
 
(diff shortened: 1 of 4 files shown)


kafeelraza:
Hi! I'd like to work on this issue.

This PR adds `start_height` support to `bitcoind_rpc::FilterIter::new`, similar to the existing behavior in `Emitter::new`, and also updates the related documentation.

evanlinjin:
NACK.

The reason `start_height` is useful in `bdk_bitcoind_rpc::Emitter` is because it's expensive to scan through every transaction in every block. If we definitely know a wallet is created after a height, we can use that height as `start_height` so that we won't need to scan blocks below that height.

The mention of reorg-handling in my original comment is just an explanation of how it functions when there is a reorg that replaces the block at or below `start_height`.

Closing as this is the wrong direction.

PR #2209, opened 2026-05-19
ci: individual package tests

### LINKED ISSUE #1944: Add CI tests for building individual packages
**Describe the enhancement**  

As described in bitcoindevkit/bdk#1017 it's possible in the `build-test` job for `--no-default-features` and `--all-features` tests to pass, but building individual packages can still fail. 

To make sure this doesn't happen again we should test building each package individually with `default`, `--all-features`, and `--no-default-features`. 

**Use case**  

Catch broken feature configurations sooner.

ValuedMammal:
Close #1944

### Description

Replaces the monolithic `build-test` matrix job (which ran a 2-toolchain × 2-feature-set matrix across the whole workspace) with dedicated per-crate build jobs. Each crate uses named steps for every meaningful feature combination, making it easy to identify which specific crate and feature set broke in CI.

**Changes:**

- **Per-crate jobs**: `build-core`, `build-chain`, `build-bitcoind-rpc`, `build-electrum`, `build-esplora`, `build-file-store`, `build-testenv` — each with per-feature `cargo build` steps and a `cargo test --all-features` step
- **`msrv` job**: Simplified from a matrix to a single flat job targeting 1.85.0; runs `cargo build --workspace --all-targets` and `cargo test --workspace --lib --tests --all-features`
- **`check-no-std`**: Updated to check `bdk_core` and `bdk_chain` with no default features + with `hashbrown` and `miniscript`; cross-compilation against `thumbv6m-none-eabi` is stubbed out as usual to be addressed in a separate PR

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -19,63 +19,223 @@ jobs:
         id: read_toolchain
         run: echo "rust_version=$(cat rust-version)" >> $GITHUB_OUTPUT
 
-  build-test:
-    needs: prepare
-    name: Build & Test
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        rust:
-          - version: ${{ needs.prepare.outputs.rust_version }}
-          - version: 1.85.0 # MSRV
-        features:
-          - --no-default-features --features bdk_chain/hashbrown
-          - --all-features
-    steps:
-      - name: checkout
-        uses: actions/checkout@v6
-        with:
-          persist-credentials: false
-      - name: Install Rust toolchain
-        uses: actions-rust-lang/setup-rust-toolchain@v1
-        with:
-            toolchain: ${{ matrix.rust.version }}
-            override: true
-            cache: true
-      - name: Pin dependencies for MSRV
-        if: matrix.rust.version == '1.85.0'
-        run: ./ci/pin-msrv.sh
-      - name: Build + Test
-        env:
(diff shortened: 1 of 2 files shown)


ValuedMammal:
- `electrum/benches/test_sync.rs` requires use-rustls feature
- ci: Changed Build 1.85.0 (no default features) to `cargo build --workspace --all-targets --no-default-features`

ValuedMammal:
Rebased

hash-ty (approved):
@ValuedMammal This looks ready man.

.github/workflows/cont_integration.yml
```diff
@@ -97,13 +257,34 @@ jobs:
...
+          toolchain: ${{ needs.prepare.outputs.rust_version }}
+          cache: true
+          # TODO: `--target thumbv6m-none-eabi` should work but currently does not
+          # target: thumbv6m-none-eabi
+      - name: Check bdk_core (no-std, hashbrown)
+        working-directory: ./crates/core
+        run: cargo check --no-default-features --features hashbrown
+      - name: Check bdk_chain (no-std, hashbrown + miniscript)
+        working-directory: ./crates/chain
+        run: cargo check --no-default-features --features hashbrown,miniscript
```
oleonardolima, line 286: nit: As you've splitted into multiple jobs, it could live in each one of them. Unless, you're looking forward to keep this one as the one that uses an specific `target`.

PR #2006, opened 2025-08-01
refactor: replace examples with focused rustdoc examples

kwsantiago:
Closes #1973

The PR is now scoped to the first step of the work discussed in #1973:

- Remove the unmaintained example binaries in examples/
- Remove the related workspace members from Cargo.toml
- Clean up docs and CI references that pointed to the removed examples

PR #2210, opened 2026-05-23
docs(rpc): add Emitter::next_block and Emitter::mempool examples

kwsantiago:
Adds `# Example` rustdoc to `Emitter::next_block` and `Emitter::mempool`, which were flagged as missing in the #2006 review.

Originally this PR also removed `example_bitcoind_rpc_polling`, but #2006 has since merged and removed all example crates, so that part is dropped after rebase. Only the doc additions remain.

Part of the #2006 follow-up to document the components previously covered by the `examples/` crates.

### DIFF
--- a/crates/bitcoind_rpc/src/lib.rs
+++ b/crates/bitcoind_rpc/src/lib.rs
@@ -101,12 +101,42 @@ where
 
     /// Emit mempool transactions and any evicted [`Txid`]s.
     ///
-    /// This method returns a [`MempoolEvent`] containing the full transactions (with their
-    /// first-seen unix timestamps) that were emitted, and [`MempoolEvent::evicted`] which are
-    /// any [`Txid`]s which were previously seen in the mempool and are now missing. Evicted txids
+    /// This method returns a [`MempoolEvent`] containing the full transactions that were emitted,
+    /// each stamped with the call's `sync_time` (unix seconds captured at the start of the call,
+    /// not when the node first saw the transaction), and [`MempoolEvent::evicted`] which are any
+    /// [`Txid`]s which were previously seen in the mempool and are now missing. The timestamp
+    /// advances on each poll, so callers should treat it as a "last seen" value. Evicted txids
     /// are only reported once the emitter’s checkpoint matches the RPC’s best block in both height
     /// and hash. Until `next_block()` advances the checkpoint to tip, `mempool()` will always
     /// return an empty `evicted` set.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// use bdk_bitcoind_rpc::{
+    ///     bitcoincore_rpc::{Auth, Client},
+    ///     Emitter, NO_EXPECTED_MEMPOOL_TXS,
+    /// };
+    /// # use bdk_core::CheckPoint;
+    /// # use bitcoin::{constants::genesis_block, Network};
+    ///
+    /// let client = Client::new("127.0.0.1:8332", Auth::None)?;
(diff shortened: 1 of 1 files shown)


nymius (changes requested):
code ACK d7e0362b

Needs rebase.

nymius:
Could you rebase again?

nymius (approved):
ACK 78a17294a458577a10bb48fec9fe6d01397cc030

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -107,6 +107,38 @@ where
...
+    ///     Emitter, NO_EXPECTED_MEMPOOL_TXS,
+    /// };
+    /// use bdk_core::CheckPoint;
+    /// use bitcoin::{constants::genesis_block, Network};
+    ///
+    /// let client = Client::new("127.0.0.1:8332", Auth::None)?;
+    /// let genesis_hash = genesis_block(Network::Bitcoin).block_hash();
+    /// let mut emitter = Emitter::new(
+    ///     &client,
+    ///     CheckPoint::new(0, genesis_hash),
```
evanlinjin: We want to incentivise callers to use the checkpoint from the receiving structures (like `LocalChain`). Can we hide the instantiation of this and have a code comment?

evanlinjin (changes requested):

crates/bitcoind_rpc/src/lib.rs
kwsantiago: Done in 0dfeb47. Hid the `CheckPoint::new` setup behind `#` lines and added a code comment pointing callers to `LocalChain::tip()` for both `next_block` and `mempool`.

crates/bitcoind_rpc/src/lib.rs
```diff
@@ -107,6 +107,34 @@ where
...
+    /// # let last_cp = CheckPoint::new(0, genesis_block(Network::Bitcoin).block_hash());
+    /// // Use the checkpoint from your receiving structure (e.g. `LocalChain::tip()`).
+    /// let mut emitter = Emitter::new(&client, last_cp, 0, NO_EXPECTED_MEMPOOL_TXS);
+    ///
+    /// // Drain blocks first so evictions can be reported once the checkpoint reaches tip.
+    /// while emitter.next_block()?.is_some() {}
+    ///
+    /// let event = emitter.mempool()?;
+    /// for (tx, seen_at) in event.update {
+    ///     // index unconfirmed `tx` (first seen at `seen_at`)
```
hash-ty: The phrase "first-seen unix timestamps" (L105) and the "first seen at seen_at" comment in the example no longer matches the implementation. Prior to #1988, `mempool` called `get_raw_mempool_verbose` which returned per-transaction time fields representing when the node first saw each tx. After #1988 the implementation uses `get_raw_mempool` (non-verbose) and stamps every tx in the response with a single sync_time captured at call time. A tx that has been sitting in the mempool for days gets the same timestamp as one that arrived a millisecond ago, and the timestamp advances on each poll. This also aligns with how the `TxGraph` and Electrum chain source treat the timestamp field - as a "last seen" value that advances monotonically.

kwsantiago: Good catch. Fixed in 1f181c4: reworded the `mempool` doc to say the timestamp is the call's `sync_time` (last-seen, advances on each poll, not first-seen), and changed the example comment from "first seen at" to "last seen at".

nymius (approved):
ACK 1f181c455c4ed0bc88da252f9cf7f8449170415a

ValuedMammal (approved):
ACK 30fa17955b16d6a5a435bea8e2b9728e9e0b3701

PR #2219, opened 2026-06-04
refactor(chain): add `CanonicalView` topological ordering

### LINKED ISSUE #2201: Add optional topological ordering stage `CanonicalViewTask`
**Describe the enhancement**  

Add a topological ordering stage to `CanonicalViewTask`.

**Use case**  

Most wallet applications would like to present an ordered view of transactions to it's users.

oleonardolima:
~depends on #2038~
closes #2201

### Description

Add topological ordering to `CanonicalView` so that parent transactions always appear before their children when iterating canonical transactions. Uses Kahn's algorithm in `CanonicalViewTask::finish()`.

### Notes to the reviewers

* `sort_topologically` is a standalone function in `canonical_view_task.rs` that operates over the existing `spends` map, no new data structures needed.
* Only edges where both parent and child are canonical are considered.
* Order among unrelated transactions is non-deterministic (`HashMap` iteration order).

### Changelog notice

```
#### Changed

- `CanonicalView` now returns transactions in topological-spending order (parents before children).
- `Canonical` struct documentation updated to reflect the ordering guarantee.
```

#### All Submissions:

#### New Features:

### DIFF
--- a/crates/chain/src/canonical.rs
+++ b/crates/chain/src/canonical.rs
@@ -21,11 +21,15 @@
 //!     println!("Transaction {}: {:?}", tx.txid, tx.pos);
 //! }
 //! ```
+//!
+//! For an ordering where every transaction appears after the transactions it spends from, see
+//! [`CanonicalView::txs_in_topological_order`].
 
 use crate::collections::HashMap;
+use alloc::collections::{BTreeSet, BinaryHeap};
 use alloc::sync::Arc;
 use alloc::vec::Vec;
-use core::{fmt, ops::RangeBounds};
+use core::{cmp::Reverse, fmt, ops::RangeBounds};
 
 use bdk_core::BlockId;
 use bitcoin::{
@@ -167,9 +171,9 @@ impl<A: Anchor> CanonicalTxOut<ChainPosition<A>> {
     }
 }
 
-/// Canonical set of transactions from a [`TxGraph`].
+/// Canonical list of transactions from a [`TxGraph`].
 ///
-/// `Canonical` provides a conflict-resolved list of transactions. It determines
+/// `Canonical` provides an ordered, conflict-resolved list of transactions. It determines
 /// which transactions are canonical (non-conflicted) based on the current chain state and
 /// provides methods to query transaction data, unspent outputs, and balances.
 ///
@@ -179,14 +183,14 @@ impl<A: Anchor> CanonicalTxOut<ChainPosition<A>> {
 ///   [`CanonicalTxs`])
 ///
 /// The view maintains:
-/// - A list of canonical transactions
+/// - A list of canonical transactions in canonical order
 /// - A mapping of outpoints to the transactions that spend them
 /// - The chain tip used for canonicalization
 ///
 /// [`TxGraph`]: crate::TxGraph
 #[derive(Debug)]
 pub struct Canonical<A, P> {
-    /// List of canonical transaction IDs.
+    /// List of canonical transaction IDs in canonical order.
     pub(crate) order: Vec<Txid>,
     /// Map of transaction IDs to their transaction data and position.
     pub(crate) txs: HashMap<Txid, (Arc<Transaction>, P)>,
@@ -199,6 +203,10 @@ pub struct Canonical<A, P> {
 }
 
 /// Type alias for canonical transactions with resolved [`ChainPosition`]s.
+///
+/// Use [`txs`](Canonical::txs) to iterate in canonical order, or
+/// [`txs_in_topological_order`](CanonicalView::txs_in_topological_order) for an ordering where
+/// every transaction appears after the transactions it spends from.
 pub type CanonicalView<A> = Canonical<A, ChainPosition<A>>;
 
 /// Type alias for canonical transactions with unresolved
@@ -269,10 +277,14 @@ impl<A, P: Clone> Canonical<A, P> {
         })
     }
 
-    /// Get an iterator over all canonical transactions in order.
+    /// Get an iterator over all canonical transactions in canonical order.
     ///
     /// Transactions are returned in canonical order, with confirmed transactions ordered by
     /// block height and position, followed by unconfirmed transactions.
+    ///
+    /// For [`CanonicalView`], an ordering where every transaction appears after the transactions
+    /// it spends from is available via
+    /// [`txs_in_topological_order`](CanonicalView::txs_in_topological_order).
     ///
     /// # Example
     ///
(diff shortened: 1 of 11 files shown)


oleonardolima:
@nymius As a heads up I also tested with your tests: https://github.com/nymius/bdk/commit/bafd2c935fbaa651f77b70e274fa44fa2a92efc2 and https://github.com/nymius/bdk/commit/b13a5400469ec4a4d7dedcccf7ba21c3959d3174 and it worked as expected.

crates/chain/src/canonical.rs
```diff
@@ -179,14 +179,14 @@ impl<A: Anchor> CanonicalTxOut<ChainPosition<A>> {
...
-/// - A list of canonical transactions
+/// - An ordered list of canonical transactions in topological-spending order
 /// - A mapping of outpoints to the transactions that spend them
 /// - The chain tip used for canonicalization
 ///
 /// [`TxGraph`]: crate::TxGraph
 #[derive(Debug)]
 pub struct Canonical<A, P> {
-    /// List of canonical transaction IDs.
+    /// Ordered list of transaction IDs in topological-spending order.
```
evanlinjin: This is only true for `CanonicalView` right? In `CanonicalTxs`, it's just ordered by the order that the txs got canonicalized.

crates/chain/src/canonical_view_task.rs
```diff
@@ -88,6 +88,81 @@ impl<'g, A: Anchor> CanonicalViewTask<'g, A> {
...
+    // Set of canonical txids — we only consider parent→child edges where
+    // both sides are in this set.
+    let canonical_set: HashSet<Txid> = order.iter().copied().collect();
+
+    // in_degree[txid] tracks how many canonical parents this tx spends from.
+    // A tx with in_degree 0 has no canonical parents and can appear first.
+    let mut in_degree: HashMap<Txid, usize> = order.iter().map(|&txid| (txid, 0)).collect();
+
+    // Adjacency list: maps each parent txid to the children that spend its
+    // outputs. Derived from `spends` where each entry (outpoint → child_txid)
```
nymius: Use only ASCII here. Remove `→`.

crates/chain/src/canonical_view_task.rs
```diff
@@ -88,6 +88,81 @@ impl<'g, A: Anchor> CanonicalViewTask<'g, A> {
...
+///    of each of its children. Whenever a child reaches `in_degree == 0`, all its parents have been
+///    placed, so it is enqueued.
+///
+/// # Note
+///
+/// The relative order among unrelated transactions (those with no spending
+/// relationship) is not guaranteed to be deterministic across runs, since
+/// it depends on `HashMap` iteration order.
+fn sort_topologically(order: &[Txid], spends: &HashMap<OutPoint, Txid>) -> Vec<Txid> {
+    // Set of canonical txids — we only consider parent→child edges where
```
nymius: Use only ASCII here: `- != —` (use -). Remove `→`.

crates/chain/src/canonical.rs
```diff
@@ -169,7 +169,7 @@ impl<A: Anchor> CanonicalTxOut<ChainPosition<A>> {
 
 /// Canonical set of transactions from a [`TxGraph`].
 ///
-/// `Canonical` provides a conflict-resolved list of transactions. It determines
+/// `Canonical` provides an ordered, conflict-resolved set of transactions. It determines
```
nymius: Why did you choose to use `set` instead of `list`? `list` is closer to ordered sequence than set.

crates/chain/src/canonical_view_task.rs
```diff
@@ -88,6 +88,81 @@ impl<'g, A: Anchor> CanonicalViewTask<'g, A> {
...
+        if canonical_set.contains(&parent) {
+            children.entry(parent).or_default().push(child);
+            *in_degree.entry(child).or_default() += 1;
+        }
+    }
+
+    // Seed the queue with root transactions (those with no canonical parents).
+    let mut queue: VecDeque<Txid> = in_degree
+        .iter()
+        .filter(|(_, &d)| d == 0)
```
nymius: Can we do `%/s/d/degree/g`?

crates/chain/src/canonical_view_task.rs
```diff
@@ -88,6 +88,81 @@ impl<'g, A: Anchor> CanonicalViewTask<'g, A> {
...
+    let canonical_set: HashSet<Txid> = order.iter().copied().collect();
+
+    // in_degree[txid] tracks how many canonical parents this tx spends from.
+    // A tx with in_degree 0 has no canonical parents and can appear first.
+    let mut in_degree: HashMap<Txid, usize> = order.iter().map(|&txid| (txid, 0)).collect();
+
+    // Adjacency list: maps each parent txid to the children that spend its
+    // outputs. Derived from `spends` where each entry (outpoint → child_txid)
+    // gives us an edge from outpoint.txid (parent) to child_txid.
+    let mut children: HashMap<Txid, Vec<Txid>> = HashMap::new();
```
nymius: The use of `Vec<Txid>` for the children collection makes this implementation broken for the following test case:
```rust
// Demonstrates the topological-order invariant on a graph where a child transaction has two
// inputs both spending from the same parent.
//         a0              ( a0 )
//        /  \             ( |  )
//       /    \            ( |  )
//    a0:0    a0:1         ( |  )
//        \  /             ( |  )
//         \/              ( |  )
//         b0              ( b0 ) -> assumed canonical
Scenario {
    name: "b0 spends a0:0 and a0:1, and both do not have anchors or last_seen",
    tx_templates: &[
        TxTemplate {
            tx_name: "a0",
            inputs: &[TxInTemplate::Bogus],
            outputs: &[
                TxOutTemplate::new(10_000, None),
                TxOutTemplate::new(10_000, None),
            ],
            anchors: &[],
            last_seen: None,
            assume_canonical: false,
        },
        TxTemplate {
            tx_name: "b0",
            inputs: &[TxInTemplate::PrevTx("a0", 0), TxInTemplate::PrevTx("a0", 1)],
            outputs: &[TxOutTemplate::new(18_000, None)],
            anchors: &[],
            last_seen: None,
            assume_canonical: true,
        },
    ],
    exp_chain_txs: Vec::from(["a0", "b0"]),
},
```

nymius: I lean to use a `BTreeSet` for the deduplication of txids, instead of a `HashSet` because it will allow the implementation of a stable order.

crates/chain/src/canonical_view_task.rs
```diff
@@ -88,6 +88,81 @@ impl<'g, A: Anchor> CanonicalViewTask<'g, A> {
...
+        // Only consider edges where the parent is also canonical — spending
+        // from a tx outside this canonical set is not a dependency.
+        if canonical_set.contains(&parent) {
+            children.entry(parent).or_default().push(child);
+            *in_degree.entry(child).or_default() += 1;
+        }
+    }
+
+    // Seed the queue with root transactions (those with no canonical parents).
+    let mut queue: VecDeque<Txid> = in_degree
```
nymius: If this where sorted lexicographically by `Txid`, the order would be stable. Any reason to not trying it?

crates/chain/tests/test_tx_graph.rs
```diff
@@ -1525,3 +1526,512 @@ fn test_get_first_seen_of_a_tx() {
...
+        (2, hash!("C")),
+        (3, hash!("D")),
+        (4, hash!("E")),
+        (5, hash!("F")),
+        (6, hash!("G"))
+    );
+    let chain_tip = local_chain.tip().block_id();
+
+    let scenarios = [
+    // a0  b0  c0
```
nymius: If we don't have relations between nodes, what are we testing here? I would remove the test cases that don't include relations.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -1525,3 +1526,512 @@ fn test_get_first_seen_of_a_tx() {
...
+                inputs: &[],
+                outputs: &[TxOutTemplate::new(2500, Some(0))],
+                anchors: &[],
+                last_seen: Some(1000),
+                assume_canonical: false,
+            },
+        ],
+        exp_chain_txs: Vec::from(["A", "B", "C"]),
+    },
+    // a0
```
nymius, line 1635: I think test cases should be simplified to the minimum problematic subgraphs: the chain (e.g.: https://github.com/nymius/bdk/commit/b13a5400469ec4a4d7dedcccf7ba21c3959d3174) and the diamond (e.g.: https://github.com/nymius/bdk/commit/bafd2c935fbaa651f77b70e274fa44fa2a92efc2).
If would add these in their minimal form, and for the rest of the test, I would remove any test case that already includes these shapes with their problematic relations, unless there is a relation that cannot be reproduced with these subgraphs.

crates/chain/tests/test_tx_graph.rs
```diff
@@ -1525,3 +1526,512 @@ fn test_get_first_seen_of_a_tx() {
...
+            inputs: &[TxInTemplate::PrevTx("f0", 0), TxInTemplate::PrevTx("f1", 0)],
+            outputs: &[TxOutTemplate::new(1000, Some(0))],
+            anchors: &[],
+            last_seen: Some(1000),
+            assume_canonical: false,
+        }
+        ],
+        exp_chain_txs: Vec::from(["a0", "b0", "c0", "d0", "e0", "f0", "f1", "g0"]),
+    },
+    //      a0
```
nymius: I don't think "big" graphs make any difference here, It's better to create a benchmark test for the topological order with a huge number of transactions and relations than increasing the size here.

nymius (changes requested):
cACK 88b7de27ce36162c2e04b0ef2523c1e53825dd65

The creation of a new method seems the right call, no one is paying the price of topological order if they don't need it.

There is a bug needing a fix.

crates/chain/src/canonical_view_task.rs
```diff
@@ -88,6 +88,81 @@ impl<'g, A: Anchor> CanonicalViewTask<'g, A> {
...
+    // outputs. Derived from `spends` where each entry (outpoint → child_txid)
+    // gives us an edge from outpoint.txid (parent) to child_txid.
+    let mut children: HashMap<Txid, Vec<Txid>> = HashMap::new();
+
+    for (outpoint, &child) in spends {
+        let parent = outpoint.txid;
+        // Only consider edges where the parent is also canonical — spending
+        // from a tx outside this canonical set is not a dependency.
+        if canonical_set.contains(&parent) {
+            children.entry(parent).or_default().push(child);
```
evanlinjin: Can we somehow push this child into an ordered position? I.e. sort the direct children by confirmation status followed by txid. This will make the order more correct and deterministic.

nymius: Lexicographic order by `Txid` is the simplest. I would only pursue the confirmation status order if there is a gain on it outside determinism.

crates/chain/src/canonical.rs
oleonardolima: Yes, I was thinking about this.

I don't think returning the topological order in here is the best approach, as this field is shared by both canonicalization and view tasks.

As it's cheap to get the topological order, and it can be optional for the user, we can have it separated instead of in a field, the field being only the list of canonical_txs, which is common on both phases.

crates/chain/src/canonical_view_task.rs
oleonardolima: Yes, though I will change it from degree to inputs or other better terminology.

crates/chain/src/canonical_view_task.rs
oleonardolima: Thanks, I'll take a look into that.

crates/chain/src/canonical_view_task.rs
evanlinjin: One of the purposes of the ordering is for wallet display. You can have a transactions of the same topological level, but confirmed in different blocks. I think it's nicer to have those ordered by confirmation block.

crates/chain/src/canonical.rs
evanlinjin: I have also been thinking along the same lines lately.

Maybe we should completely get rid of `Canonical::order`. Currently, it only records "discovery order" which is not helpful.

We can have a separate method `Canonical::topological_txs` that does the topological ordering.

Also, for @nymius' UTXOs-with-unowned-unconfirmed-ancestors-should-be-untrusted balance categorization improvement. The best algorithm (imo) would be to traverse UTXO-containing txs in topological levels, but backwards (towards the roots) in one go, accumulating visited txids and stopping after we hit a confirmed tx, or an unconfirmed and unowned tx (in which we mark the txid set as "tainted"). For this use case, the `Canonical::order` field does not help us either as we only want to traverse a sub-portion of `Canonical` and the most efficient way is to just do Kahn's algorithm again anyway. This is just further proof that `Canonical::order` has no usecase and can be removed.

crates/chain/src/canonical_view_task.rs
luisschwab: > One of the purposes of the ordering is for wallet display. You can have a transactions of the same topological level, but confirmed in different blocks. I think it's nicer to have those ordered by confirmation block.

Yes, that's very useful.

crates/chain/src/canonical.rs
```diff
@@ -179,14 +183,18 @@ impl<A: Anchor> CanonicalTxOut<ChainPosition<A>> {
...
+/// Use [`txs`](Self::txs) to iterate in canonical order, or
+/// [`txs_in_topological_order`](Self::txs_in_topological_order) for an ordering where every
+/// transaction appears after the transactions it spends from.
+///
 /// [`TxGraph`]: crate::TxGraph
 #[derive(Debug)]
 pub struct Canonical<A, P> {
-    /// List of canonical transaction IDs.
+    /// List of canonical transaction IDs in canonical order.
     pub(crate) order: Vec<Txid>,
```
evanlinjin, line 194: Should we get rid of this?

oleonardolima:
I moved it back to draft as I still need to go through the new commits I've pushed. However, in those I asked Claude to address the review comments above.

nymius:
The commit message from 0fe25402587e9913065270fefbbf28c3b71c3fc9 is confusing:

> - fix `is_ordered_topologically` to ignore non-canonical (external)
  inputs (the actual issue behind nymius's double-spend "bug")
  
The uncovered bug is not a double spend one, but spending different outpoints from the same tx in the same spending tx, which is a valid case.

> - add the double-spend-from-same-parent test case, replacing the
  relation-less scenarios
  
  Again, this is a misnomer, this isn't a double-spend case, but an issue related to counting parents multiple times.
  
  Will review again once is out of draft.

crates/chain/src/canonical.rs
evanlinjin: I think we should do the following instead:

* Keep this field. All `ChainQuery::Output`s are topologically ordered. `CanonicalTask::finish` should canonicalize into `Canonical::order`.
* Change the signature of this field to track topological "level" as well. Maybe `Vec<(u32, Txid)>`?
* `CanonicalViewTask::finish` will tiebreak each topological "level" with anchors and first-seen values.

evanlinjin:
@oleonardolima https://github.com/bitcoindevkit/bdk/pull/2219#discussion_r3986622150

PR #2218, opened 2026-06-03
feat(chain): add walk_ancestors_from_root to TxGraph

Dmenec:
Solves https://github.com/bitcoindevkit/bdk/issues/2216

### Description

Add `walk_ancestors_from_root` method to `TxGraph`, which mirrors the existing `walk_ancestors` but includes the root transaction in the iteration. 

### Notes to the reviewers

An alternative implementation would have been to make TxAncestors::new_include_root public directly, instead of adding a wrapper method on TxGraph. Not sure which approach is preferable here.

#### Added
`TxGraph::walk_ancestors_from_root`

#### All Submissions:

* [X] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)

#### New Features:

* [ ] I've added tests for the new feature
* [X] I've added docs for the new feature

PR #2221, opened 2026-06-05
refactor(chain)!: remove `min_confirmations` parameter from `CanonicalView::balance`

Dmenec:
### Description

This builds on top of https://github.com/bitcoindevkit/bdk/pull/2038.

Removes the `min_confirmations` parameter from `CanonicalView::balance`.

This is part of the redesign of the wallet `balance` API (see bitcoindevkit/bdk_wallet#451).

#### Trust Categorization

The goal is to fix how `balance` decides if a pending output is trusted or not. Right now the wallet decides this from the keychain (internal = trusted), which has some issues (bitcoindevkit/bdk_wallet#16, bitcoindevkit/bdk_wallet#273).

The approach discussed in https://github.com/bitcoindevkit/bdk_wallet/pull/431 describes creating the trust predicate by walking the tx ancestors. An unconfirmed output is trusted only if every unconfirmed ancestor spends inputs that are ours. 

To achieve that, we need a few things together in chain:

- #2038 — the new CanonicalView, which lets us look up any tx by txid and get its canonical position.
- #2218 — makes it possible to walk the ancestors of a tx from the transaction itself.
- this PR — removing min_confirmations so a confirmed tx with few confirmations is not wrongly treated as pending while walking ancestors.

There is a small prototype test of this trust logic here: https://github.com/bitcoindevkit/bdk/commit/b0acd3ba93527e149c6afb371b4be6063a29f882#diff-d3f18e9a77fd241fa351a0cbcd75e993d90819834ccde09b7f4247c153ff26eeR586-R607. It is just how the idea may work in wallet, not the final API. 

#### Removing `min_confirmations`

The ancestor walk decides trust by checking each ancestor's canonical position. A `Confirmed` ancestor stops the walk, an `Unconfirmed` one is only trusted if all of its inputs are also ours. For this to be coherent, `Confirmed` has to mean the same thing in `balance` as it does in the walk. 

On the previous approach the parameter `min_confirmations` broke that. A `Confirmed` output below the `min_confirmations` threshold was reported as `pending`, even though the walk still saw it as `Confirmed`. 

Removing it, `balance` classifies purely from the canonical position, and `Confirmed` means the same thing everywhere. This also aligns `balance` with the usage of every other method of `CanonicalView` which already reads positions relative to the view's tip and leaves interpretation to the caller. 

With the parameter gone, the way to ask for a confirmation threshold is though the tip the `CanonicalView` is build with. A view is always a snapshot of the chain seen from a given tip, so "how many confirmations does this output have" is really the same question as "which tip am I looking from". To require N confirmations, the view is built at `tip - N +1`: 

> Building the view at tip - N + 1 makes a tx count as `Confirmed` exactly when it has **at least N confirmations relative to the real tip**.

Any tx whose block ends up above that tip is no longer in the chain from a view's perspective, so it falls back to `Unconfirmed` and is classified by the trust predicate. 

The downside is that this moves some work to the caller. The tip used to build the view must be a block that the `LocalChain` already has, since `canonical_view` checks it against the chain's checkpoints. The chain is sparse, it does not keep every height, so a checkpoint that is not there can't be used. Before, the threshold was just a number checked against the current tip and nothing else was needed. Now, if the caller wants to measure from a height the chain never kept, it has to insert or fetch that block first.

### Notes to the reviewers

In the `Confirmed` branch of balance the `if confirmations < min_confirmations` path is dropped, so a confirmed output can now only go to `confirmed` or `immature`, never to `*_pending`. The `pending` categories are now reached only from the `Unconfirmed` branch.

### Changelog notice

- **Breaking:** Removed the `min_confirmations` parameter from `CanonicalView::balance`.

#### All Submissions:


nymius:
cc: @evanlinjin , what do you think? Just last commit: dec313365602a48cb338afe0e4ffaa9140a2b39c

evanlinjin:
@nymius I'll have a look! Give me a bit of time, I just got informed for the incomplete Balance problem.

evanlinjin:
@Dmenec thank you for this PR.

Unfortunately, `min_confirmations` is different to canonicalizing with a lower tip. A transaction confirmed at the actual tip is not guaranteed to be the best canonicalization candidate at a lower tip. Canonicalizing with a lower tip may get a completely different canonical set.

I don't think `min_confirmations` is the bottleneck to move towards an UTXO-ancestor based trust categorization. I would argue that the shape of `trust_predicate` is the problem.

I agree that `min_confirmations`/`Balance::confirmed` is badly named. However, I think a similar concept is still useful as blocks in a bitcoin network aren't guaranteed to be permanent. Maybe "settled" would be a better terminology. I think the UTXO-ancestor based trust categorization should stop at "settled" (which is configurable) and not purely "confirmed" as different wallets have different trust assumptions. For some, 1 confirmation is not good enough.

I've made a prototype of the direction I think we should take in #2235. Let me know what you think.

Dmenec:
Thank you @evanlinjin for checking this!

> Unfortunately, min_confirmations is different to canonicalizing with a lower tip. A transaction confirmed at the actual tip is not guaranteed to be the best canonicalization candidate at a lower tip. Canonicalizing with a lower tip may get a completely different canonical set.

> I don't think min_confirmations is the bottleneck to move towards an UTXO-ancestor based trust categorization. I would argue that the shape of trust_predicate is the problem.

Yes, I also think it's not a bottleneck. This was just one part of the `balance` redesign. What @nymius and I discussed was whether we could use `CanonicalView` to get a canonical set of txs that acted as a limiter (i.e. a `min_confirmations` parameter meaning "I want the tip to have at least N confirmations"). 

> I agree that min_confirmations/Balance::confirmed is badly named. However, I think a similar concept is still useful as blocks in a bitcoin network aren't guaranteed to be permanent. Maybe "settled" would be a better terminology. I think the UTXO-ancestor based trust categorization should stop at "settled" (which is configurable) and not purely "confirmed" as different wallets have different trust assumptions. For some, 1 confirmation is not good enough.

Agree! `settled` is a good path forward. As far as I've seen on #2235, it's totally a better option if we want flexibility for handling more complex cases. 

> I've made a prototype of the direction I think we should take in https://github.com/bitcoindevkit/bdk/pull/2235. Let me know what you think.

Taking a look at it :) I had some similar implementation in mind for the ancestor walk, I'll write it up in a follow-up comment on #2235.

Dmenec:
Closed in favor of #2235.

nymius:
> Unfortunately, `min_confirmations` is different to canonicalizing with a lower tip. A transaction confirmed at the actual tip is not guaranteed to be the best canonicalization candidate at a lower tip. Canonicalizing with a lower tip may get a completely different canonical set.

But why it would be different than canonicalizing with a lower tip but including the txs from the actual tip as assumed canonical?

> I don't think `min_confirmations` is the bottleneck to move towards an UTXO-ancestor based trust categorization. I would argue that the shape of `trust_predicate` is the problem.

This can be done in multiple other ways, but trustiness should come after settlement. By including `min_confirmations` as a parameter, the code is blending both, making it harder to reason about IMHO.

evanlinjin:
> But why it would be different than canonicalizing with a lower tip but including the txs from the actual tip as assumed canonical?

That is a lot more work for the caller just to re-categorize. I think the better API would be to get rid of the `balance` and have something like `classify_outpoints` which returns a list of `(outpoint, category)` tuples. This way, the caller can "disagree" with our categorization without having to revisit canonicalization with different parameters in multiple runs.

> This can be done in multiple other ways, but trustiness should come after settlement. By including min_confirmations as a parameter, the code is blending both, making it harder to reason about IMHO.

If the wallet considered <3 confs as non-final, I think it's a fair policy to treat <3-conf-txs the same as unconfirmed txs for UTXO set and balance. I don't think this is hard to reason about.

---


@nymius please have a skim over what is being proposed in #2235 - I think it will make more sense.

nymius:
> I think the better API would be to get rid of the `balance` and have something like `classify_outpoints` which returns a list of `(outpoint, category)` tuples. This way, the caller can "disagree" with our categorization without having to revisit canonicalization with different parameters in multiple runs.

I agree with this.

> If the wallet considered <3 confs as non-final, I think it's a fair policy to treat <3-conf-txs the same as unconfirmed txs for UTXO set and balance. I don't think this is hard to reason about.

I also agree with this, I disagreed in the mechanism: `min_confirmations` operating on top of the `Confirmed/Unconfirmed` notion. The introduction of `Elegibility` is fixing it, so I'm happy with going down that path.


P.S.: I replied on top of your comment by mistake. I updated it later to your version.

PR #2227, opened 2026-06-13
fix(chain,core): prevent serde from implicitly leaking std into no_std environments

EliteCoder18:
### Description

Fix silent `serde/std` leak in `bdk_core` and `bdk_chain` both crates declared their `serde` optional dependency without `default features = false`, causing serde's std feature to activate unconditionally even in no_std builds.

- `bdk_core`: added `default-features = false` + `serde?/std` in std feature. 
- Dropped `rc` the only two serde-derived types (`BlockId`, `ConfirmationBlockTime`) are pure scalars needing neither `serde/rc` nor `serde/alloc`.
- `bdk_chain`: added `default-features = false` and explicitly enabled`alloc` and `rc` directly on the `serde dep` + `serde?/std` in std. Both are required because `tx_graph::ChangeSet` holds `BTreeSet<Arc<Transaction>>` and `serde` gates `Arc<T>: Serialize` behind `all(feature = "rc", any(feature = "std", feature = "alloc"))`.

This fix was identified while working on (https://github.com/bitcoindevkit/bdk_wallet/pull/494).

### Notes to the reviewers

The `alloc` and `rc` features are placed on the serde dependency directly rather than behind a separate feature flag because `bdk_chain` unconditionally requires `alloc` anyway. 

### Changelog notice
```
### Fixed

- fix(core): serde dep was missing `default-features = false`, leaking `serde/std` into no_std builds and remove unnecessary `rc` feature .
- fix(chain): same leak fixed and `serde/alloc` and `serde/rc` now explicitly enabled on the dep .
```

#### All Submissions:

### DIFF
--- a/crates/chain/Cargo.toml
+++ b/crates/chain/Cargo.toml
@@ -18,7 +18,7 @@ workspace = true
 [dependencies]
 bitcoin = { version = "0.32.0", default-features = false }
 bdk_core = { path = "../core", version = "0.6.2", default-features = false }
-serde = { version = "1", optional = true, features = ["derive", "rc"] }
+serde = { version = "1", optional = true, default-features = false, features = ["derive", "rc", "alloc"] }
 miniscript = { version = "13.0.0", optional = true, default-features = false }
 
 # Feature dependencies
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -12,12 +12,12 @@ readme = "README.md"
 
 [dependencies]
 bitcoin = { version = "0.32", default-features = false }
-serde = { version = "1", optional = true, features = ["derive", "rc"] }
+serde = { version = "1", optional = true, default-features=false, features = ["derive"] }
 hashbrown = { version = "0.14.5", optional = true, default-features = false, features = ["ahash", "inline-more"] }
 
 [features]
 default = ["std"]
-std = ["bitcoin/std"]
+std = ["bitcoin/std", "serde?/std"]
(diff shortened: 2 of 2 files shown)


notmandatory (approved):
ACK cc4cef475c3cd8e9fb930752b7dc41d2709e453c

Good catch finding that `rc` isn't needed for `bdk_core`.

crates/chain/Cargo.toml
```diff
@@ -32,7 +32,7 @@ criterion = { version = "0.7" }
 
 [features]
 default = ["std", "miniscript"]
-std = ["bitcoin/std", "miniscript?/std", "bdk_core/std"]
+std = ["bitcoin/std", "miniscript?/std", "bdk_core/std", "serde?/std"]
```
oleonardolima: nit: do we need this, as it's already being brought by `bdk_core/std` and `bdk_core/serde` below ?

oleonardolima (commented):
utACK cc4cef475c3cd8e9fb930752b7dc41d2709e453c

I merged the update of rustc, so you'll need to rebase it run with the latest CI.

crates/chain/Cargo.toml
EliteCoder18: You're right, removed. Since both crates are in the same repo the transitive path through `bdk_core/std` is guaranteed. Thanks!

oleonardolima (approved):
utACK 29d1f5b971b2aea155ef56146df3eba88c6cf4b6

evanlinjin (approved):
ACK 29d1f5b971b2aea155ef56146df3eba88c6cf4b6

PR #2231, opened 2026-06-15
refactor(testenv)!: make `wait_until_electrum_sees_block` concrete

evanlinjin:
### Description

A from-scratch reimplementation of bitcoindevkit/bdk#1640 on top of the current `master`.

`TestEnv::wait_until_electrum_sees_block` previously only took a `timeout` and returned as soon as Electrum emitted *any* new-block notification. That made it ambiguous which block was actually being waited for, and it could return before Electrum had caught up to the block the test cared about.

This PR makes the wait concrete:

- `wait_until_electrum_sees_block` now takes `block_height: usize` and `block_hash: Option<BlockHash>`, so callers state exactly which block Electrum must be aware of. If the notified tip has already advanced past `block_height`, the header at that height is fetched explicitly so the expected `block_hash` can still be verified.
- A new convenience wrapper `wait_until_electrum_tip_syncs_with_bitcoind` waits until Electrum's tip matches bitcoind's chain tip (both height and hash). All existing call sites are migrated to it, since in every case the intent was "wait until Electrum catches up to the freshly mined/reorged tip".

### Notes to the reviewers

- **Why subscribe instead of polling:** polling Electrs for the header at `block_height` and checking its hash does not guarantee the spk histories are current. Receiving a block-tip notification does, because of how Electrs indexes. This is captured in a `NOTE` in the code.
- **Tip advanced past target:** the original PR's reviewers asked what happens when the method is called after the chain tip has moved past the target height. This is handled explicitly: when the notified height is greater than `block_height`, the header at `block_height` is fetched directly to verify the hash, rather than relying on the notification header.
- **`trigger()` placement:** `electrsd.trigger()` is kept inside the polling loop (matching the previous `master` behaviour) so Electrs keeps getting nudged to re-index — this is important for reorgs, where the tip changes without a height change.
- The previous WIP artefacts from #1640 (a `// TODO: Fails here.` marker and a struck-through doc comment) are intentionally left out.

### Changelog notice

- Changed: `TestEnv::wait_until_electrum_sees_block` now takes a `block_height` and an optional `block_hash` (breaking change to `bdk_testenv`).
- Added: `TestEnv::wait_until_electrum_tip_syncs_with_bitcoind`.

#### All Submissions:

#### New Features:

#### Bugfixes:

https://claude.ai/code/session_016gJvi98Q65QnanBFxe2UUn

---
_Generated by [Claude Code](https://claude.ai/code/session_016gJvi98Q65QnanBFxe2UUn)_

### DIFF
--- a/crates/testenv/src/lib.rs
+++ b/crates/testenv/src/lib.rs
@@ -291,9 +291,21 @@ impl TestEnv {
         Err(anyhow::anyhow!("Cannot find nonce that meets the target"))
     }
 
-    /// This method waits for the Electrum notification indicating that a new block has been mined.
+    /// Wait until Electrum is aware of a block at the given `block_height` (and optionally that the
+    /// block at that height matches the given `block_hash`).
+    ///
     /// `timeout` is the maximum [`Duration`] we want to wait for a response from Electrsd.
-    pub fn wait_until_electrum_sees_block(&self, timeout: Duration) -> anyhow::Result<()> {
+    pub fn wait_until_electrum_sees_block(
+        &self,
+        block_height: usize,
+        block_hash: Option<BlockHash>,
+        timeout: Duration,
+    ) -> anyhow::Result<()> {
+        // NOTE: We use the subscribe endpoint because polling Electrs for a block header at
+        // `block_height` and verifying the `block_hash` does not ensure that the spk histories are
+        // current. In contrast, getting a notification for a new block tip ensures that the
+        // confirmed spk histories are current, including those of the newly notified tip. This is a
+        // result of the internal workings of Electrs.
         self.electrsd.client.block_headers_subscribe()?;
         let delay = Duration::from_millis(200);
         let start = std::time::Instant::now();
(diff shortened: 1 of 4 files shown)


luisschwab (approved):
ACK f6a60a34a24919383fe7ac2f474c468ae748a406

oleonardolima (approved):
ACK f6a60a34a24919383fe7ac2f474c468ae748a406

crates/testenv/src/lib.rs
```diff
@@ -291,29 +291,66 @@ impl TestEnv {
...
 
-    /// This method waits for the Electrum notification indicating that a new block has been mined.
+    /// Wait until Electrum is aware of a block at the given `block_height` (and optionally that the
+    /// block at that height matches the given `block_hash`).
+    ///
     /// `timeout` is the maximum [`Duration`] we want to wait for a response from Electrsd.
-    pub fn wait_until_electrum_sees_block(&self, timeout: Duration) -> anyhow::Result<()> {
+    pub fn wait_until_electrum_sees_block(
+        &self,
+        block_height: usize,
```
tvpeter, line 300: nit: To avoid unnecessary casting, could you specify this as either `u32` or `u64`? In the outgoing `bitcoincore-rpc`, the `block_count` returns u64 while in the new client `bdk_bitcoind_client`, it returns `u32`.

tvpeter (approved):
ACK f6a60a34a24919383fe7ac2f474c468ae748a406

Thank you for working on this. I left a `nit`.

crates/testenv/src/lib.rs
luisschwab: Yes, on the next `rust-bitcoin` release we will have a `BlockHeight` type which wraps `u32`.

PR #2235, opened 2026-06-23
feat(chain)!: taint-aware `CanonicalView::balance` via canonical ancestor walk

evanlinjin:
> [!WARNING]
> **This PR — including this description — is currently entirely AI-generated.**
> It still needs proper human review (by me, @evanlinjin) before it should be taken seriously. Treat everything below as a draft proposal, not a vetted design.

### Description

This is an alternative direction to #2221. Where #2221 makes `CanonicalView::balance` less restrictive by *removing* the `min_confirmations` parameter, this PR instead tries to make it **more useful** — handing control to the caller via closures, deriving trust from *ancestry* rather than a per-script heuristic, and exposing a per-output **spend-eligibility classifier** that's the building block for coin control.

The core new primitive is `CanonicalView::classify_outpoints`, built on a reusable, sans-`TxGraph` **canonical ancestor walk**. `balance` becomes a thin fold over it.

### What's here (commit by commit)

**1. Add `CanonicalAncestors` reverse-topological walk.** `Canonical::ancestors(seeds, map, should_walk)` walks the canonical ancestors of a set of *seed* txids backwards (the seeds are the leaf-most txs, closer to the leaves of the ancestry DAG than its roots). Each tx is visited once in reverse-topological order, folding a `Merge`able accumulator from descendants into ancestors (diamonds merge correctly). `should_walk` prunes per-tx with the `ChainPosition` available; the iterator is `ExactSizeIterator`.

**2. Add `Canonical::ancestors_inclusive`.** A variant that also yields the seeds (each with its own final accumulator), so a per-seed contribution can be folded uniformly with the ancestry.

**3. Taint-aware `CanonicalView::balance` (breaking).** Replaces `trust_predicate` + `min_confirmations` with two closures:
- `does_taint(CanonicalTx) -> bool`: a pending output is untrusted if it, or any of its *unsettled* ancestors, taints. The unsettled ancestry is walked once (deduped across outputs, stopping at settled txs) via `ancestors_inclusive`. This lets callers demote unconfirmed coins received from — or chained on top of — a third party. Trust is now derived from ancestry instead of a per-script heuristic.
- `is_settled(&ChainPosition) -> bool`: generalizes the numeric `min_confirmations` into a caller-defined "confident this won't be replaced" boundary, and is the sole authority on it — a settled output is never dropped or tainted (even an unconfirmed output a caller chooses to treat as settled is counted, not lost).
- `balance` also drops its `O` identifier generic and now takes bare `OutPoint`s.

**4. Rename `Balance::confirmed` to `Balance::settled` (breaking).** The bucket is driven by `is_settled` (confidence a tx won't be replaced), not strictly confirmation status, so the field is renamed to match. Breaking serde key change (`confirmed` → `settled`, no alias).

**5. Add `classify_outpoints` spend-eligibility classifier.** `CanonicalView::classify_outpoints` pairs each unspent output with a chain-level `Eligibility` (`Settled` / `Immature` / `TrustedPending` / `UntrustedPending`) from `is_settled` + the `does_taint` walk. This is the primitive for **coin selection / coin control** (e.g. "prefer settled coins, fall back to trusted-pending"): the caller decides how to aggregate and can layer wallet-specific categories like "locked" on top, instead of being constrained to the fixed `Balance` buckets. `balance` is re-expressed as a thin fold over it.

### Direction / open questions for human review

- The longer-term idea is that **`classify_outpoints` is the real API** and `Balance` is just one possible fold — wallets that want categories like "locked"/"reserved" aggregate themselves. This PR keeps `Balance` (as a fold) for now; a follow-up could move it down into `bdk_wallet` and drop it from `bdk_chain`.
- Is replacing `trust_predicate` with ancestry-based `does_taint` the right call, or should both coexist? (`does_taint` can't express per-output trust, but per-tx ancestry is arguably the more principled model.)
- `Balance::confirmed` → `Balance::settled`: worth the rename / serde break, or keep the conventional name?
- `bdk_wallet` will need its `balance` call updated (out of this repo's tree).

### Notes to reviewers

- Tests: ancestor-walk tests (diamond merge / reverse-topo order / pruning / exact size), a balance + `classify_outpoints` ancestry-taint test, and a regression test that a settled-everything `is_settled` never drops an unconfirmed output. Existing balance tests were migrated (the old per-keychain trust test is recomputed for the new model, since none of its txs actually spend third-party coins).
- `cargo fmt`, `clippy --all-features --all-targets -p bdk_chain`, and `RUSTDOCFLAGS="-D warnings" cargo doc` are clean; the `bdk_chain` suite + doctests pass.

#### Changed

- `CanonicalView::balance` now takes `does_taint` and `is_settled` closures instead of `trust_predicate` and `min_confirmations`, and accepts bare `OutPoint`s.
- Renamed `Balance::confirmed` to `Balance::settled`.

#### Added

- `CanonicalView::classify_outpoints` + `Eligibility`: per-output spend-eligibility classification (the building block for coin control).
- `Canonical::ancestors` / `Canonical::ancestors_inclusive`: reverse-topological canonical ancestor walk with a `Merge`able accumulator.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/chain/src/balance.rs
+++ b/crates/chain/src/balance.rs
@@ -10,22 +10,23 @@ pub struct Balance {
     pub trusted_pending: Amount,
     /// Unconfirmed UTXOs received from an external wallet
     pub untrusted_pending: Amount,
-    /// Confirmed and immediately spendable balance
-    pub confirmed: Amount,
+    /// Settled balance: outputs whose transaction we are confident will not be replaced (e.g.
+    /// confirmed deeply enough), as determined by the balance query's `is_settled` predicate.
+    pub settled: Amount,
 }
 
 impl Balance {
-    /// Get sum of trusted_pending and confirmed coins.
+    /// Get sum of trusted_pending and settled coins.
     ///
     /// This is the balance you can spend right now that shouldn't get cancelled via another party
     /// double spending it.
     pub fn trusted_spendable(&self) -> Amount {
-        self.confirmed + self.trusted_pending
+        self.settled + self.trusted_pending
     }
 
     /// Get the whole balance visible to the wallet.
     pub fn total(&self) -> Amount {
-        self.confirmed + self.trusted_pending + self.untrusted_pending + self.immature
+        self.settled + self.trusted_pending + self.untrusted_pending + self.immature
     }
 }
 
@@ -33,8 +34,8 @@ impl core::fmt::Display for Balance {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         write!(
             f,
-            "{{ immature: {}, trusted_pending: {}, untrusted_pending: {}, confirmed: {} }}",
-            self.immature, self.trusted_pending, self.untrusted_pending, self.confirmed
+            "{{ immature: {}, trusted_pending: {}, untrusted_pending: {}, settled: {} }}",
+            self.immature, self.trusted_pending, self.untrusted_pending, self.settled
         )
     }
 }
@@ -47,7 +48,7 @@ impl core::ops::Add for Balance {
             immature: self.immature + other.immature,
             trusted_pending: self.trusted_pending + other.trusted_pending,
             untrusted_pending: self.untrusted_pending + other.untrusted_pending,
-            confirmed: self.confirmed + other.confirmed,
+            settled: self.settled + other.settled,
         }
     }
 }
--- a/crates/chain/src/canonical.rs
+++ b/crates/chain/src/canonical.rs
@@ -29,9 +29,7 @@ use alloc::vec::Vec;
 use core::{fmt, ops::RangeBounds};
 
 use bdk_core::BlockId;
-use bitcoin::{
-    constants::COINBASE_MATURITY, Amount, OutPoint, ScriptBuf, Transaction, TxOut, Txid,
-};
+use bitcoin::{constants::COINBASE_MATURITY, OutPoint, ScriptBuf, Transaction, TxOut, Txid};
 
 use crate::{spk_txout::SpkTxOutIndex, Anchor, Balance, CanonicalViewTask, ChainPosition, TxGraph};
 
@@ -394,25 +392,45 @@ impl<A, P: Clone> Canonical<A, P> {
     }
 }
 
+/// The spend-eligibility classification of a canonical output, produced by
+/// [`CanonicalView::classify_outpoints`].
+///
+/// This is a *chain-level* classification: it captures only what the chain can determine
+/// (settled-ness, maturity, and taint). It deliberately knows nothing about wallet policies such as
+/// locked or reserved coins — callers layer their own categories on top.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Eligibility {
+    /// A settled coinbase output that has not yet matured. Not spendable.
+    Immature,
+    /// Settled: the output's transaction is considered unlikely to be replaced.
+    Settled,
+    /// Pending (not settled), and neither it nor any of its unsettled ancestors taints it.
+    TrustedPending,
+    /// Pending (not settled), and it or one of its unsettled ancestors taints it.
+    UntrustedPending,
+}
+
 impl<A: Anchor> CanonicalView<A> {
-    /// Calculate the total balance of the given outpoints.
+    /// Classify each of the given `outpoints` by its [spend eligibility](Eligibility).
     ///
-    /// This method computes a detailed balance breakdown for a set of outpoints, categorizing
-    /// outputs as confirmed, pending (trusted/untrusted), or immature based on their chain
-    /// position and the provided trust predicate.
+    /// This is the primitive behind [`balance`](Self::balance) and is the building block for coin
+    /// selection / coin control: it yields each unspent output paired with a chain-level
+    /// [`Eligibility`], leaving aggregation (and any wallet-specific categories like "locked") to
+    /// the caller. Spent outpoints, and outpoints not in this canonical view, are skipped.
     ///
     /// # Arguments
     ///
-    /// * `outpoints` - Iterator of outpoints to calculate balance for
+    /// * `outpoints` - Iterator of outpoints to classify.
     /// * `does_taint` - Function that returns `true` for transactions that should *taint* their
-    ///   descendants. A pending output counts toward `untrusted_pending` if its transaction, or any
-    ///   of its unsettled ancestors, taints; otherwise it counts toward `trusted_pending`. See
-    ///   [Taint](#taint) below.
+    ///   descendants. A pending output is [`UntrustedPending`](Eligibility::UntrustedPending) if
+    ///   its transaction, or any of its unsettled ancestors, taints; otherwise
+    ///   [`TrustedPending`](Eligibility::TrustedPending). See [Taint](#taint) below.
     /// * `is_settled` - Function that returns `true` for the [position](ChainPosition) of an output
     ///   whose transaction is considered settled — unlikely to be replaced (e.g. confirmed deeply
-    ///   enough). Settled outputs count toward `settled`/`immature`; the rest are pending.
-    ///   `is_settled` is the sole authority on this boundary — a settled output is never dropped or
-    ///   tainted. See [Settled](#settled) below.
+    ///   enough). Settled outputs are [`Settled`](Eligibility::Settled) or
+    ///   [`Immature`](Eligibility::Immature); the rest are pending. `is_settled` is the sole
+    ///   authority on this boundary — a settled output is never tainted. See [Settled](#settled)
+    ///   below.
     ///
     /// # Settled
     ///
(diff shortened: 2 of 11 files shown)


evanlinjin:
@oleonardolima It will be ideal if this can reuse logic from #2219.

crates/chain/src/balance.rs
```diff
@@ -10,31 +10,32 @@ pub struct Balance {
     pub trusted_pending: Amount,
     /// Unconfirmed UTXOs received from an external wallet
     pub untrusted_pending: Amount,
-    /// Confirmed and immediately spendable balance
-    pub confirmed: Amount,
+    /// Settled balance: outputs whose transaction we are confident will not be replaced (e.g.
+    /// confirmed deeply enough), as determined by the balance query's `is_settled` predicate.
+    pub settled: Amount,
```
Dmenec, line 15: Agree, `settled` better reflects that the threshold is a wallet policy decision.

crates/chain/src/canonical.rs
```diff
@@ -393,99 +392,205 @@ impl<A, P: Clone> Canonical<A, P> {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Eligibility {
+    /// A settled coinbase output that has not yet matured. Not spendable.
+    Immature,
+    /// Settled: the output's transaction is considered unlikely to be replaced.
+    Settled,
+    /// Pending (not settled), and neither it nor any of its unsettled ancestors taints it.
+    TrustedPending,
+    /// Pending (not settled), and it or one of its unsettled ancestors taints it.
+    UntrustedPending,
```
Dmenec, line 410: I like the taint concept more than saying some previous transaction is untrusted, what's actually untrusted is the pending UTXO, not the ancestor.

crates/chain/src/canonical.rs
```diff
@@ -393,99 +392,205 @@ impl<A, P: Clone> Canonical<A, P> {
...
+    ///     match eligibility {
+    ///         Eligibility::Settled | Eligibility::TrustedPending => candidates.push(txout.outpoint),
+    ///         Eligibility::Immature | Eligibility::UntrustedPending => {}
+    ///     }
+    /// }
+    /// ```
+    pub fn classify_outpoints(
+        &self,
+        outpoints: impl IntoIterator<Item = OutPoint>,
+        mut does_taint: impl FnMut(CanonicalTx<ChainPosition<A>>) -> bool,
```
Dmenec, line 492: `trust_predicate` let callers express per-output trust by keychain. `does_taint` only sees the ancestor transaction, which does not allow for a "blind trust assumption" (i.e. this output belongs to my keychain so trust it). 

Not sure if this policy now has to be layered on top of the `Eligibility` results.

crates/chain/src/canonical.rs
```diff
@@ -393,99 +392,205 @@ impl<A, P: Clone> Canonical<A, P> {
...
+            let mut tainted = HashSet::<Txid>::new();
+            // Walk each pending output together with its unsettled ancestry (deduped across seeds,
+            // stopping at settled transactions). Each transaction carries the set of descendants it
+            // reaches; when an unsettled transaction taints, all of them are tainted. The settled
+            // boundary is yielded but never taints (a settled transaction cannot be replaced).
+            for (descendants, c_tx) in self.ancestors_inclusive::<BTreeSet<Txid>, _, _>(
+                seeds.iter().copied(),
+                |c_tx| core::iter::once(c_tx.txid).collect(),
+                |c_tx| !is_settled(&c_tx.pos),
+            ) {
```
Dmenec, line 520: `does_taint` is called eagerly for every unsettled ancestor here and before the returned iterator is built. So if a caller provides an expensive closure, we'd have to evaluate it against every walked ancestor upfront, even if we only end up needing a small part of the UTXOs.

crates/chain/tests/test_canonical_ancestors.rs
```diff
@@ -0,0 +1,266 @@
...
+
+/// Builds a diamond of confirmed transactions and returns the chain, graph and the four txs.
+///
+/// ```text
+///        tx_a            (2 outputs)
+///       /    \
+///    tx_b    tx_c
+///       \    /
+///        tx_d            (root we walk ancestors from)
+/// ```
```
Dmenec, line 37: Maybe checking with different topolgies would be useful to cover. 

For instance: Disjoint seeds with no shared ancestry or a seed that is itself an ancestor of another seed.

Dmenec (commented):
Hey @evanlinjin thank you for pushing this. 

As there are many things involved, I tried to go point-by-point on some thoughts I have in here. 

Related to your questions: 

> The longer-term idea is that classify_outpoints is the real API and Balance is just one possible fold — wallets that want categories like "locked"/"reserved" aggregate themselves. This PR keeps Balance (as a fold) for now; a follow-up could move it down into bdk_wallet and drop it from bdk_chain.

Agree, `classify_outpoints` + `Eligibility` is a good primitive, `Balance` is just a convenience fold on top so it makes sense to me. I think moving `Balance` out of `bdk_chain` would be a breaking change for consumers calling it directly, so maybe the migration needs care.

> Is replacing trust_predicate with ancestry-based does_taint the right call, or should both coexist? (does_taint can't express per-output trust, but per-tx ancestry is arguably the more principled model.)

As said in one comment below, `trust_predicate` had this identifier that provided specific assumptions per-output. I don't think that's a problem though, that policy was always wallet-specific. It makes more sense for `bdk_wallet` to handle it externally. 

> `bdk_wallet` will need its balance call updated (out of this repo's tree).

This is the part I'm least sure about, so I'll just lay out what I found and leave it open.

While looking into the walk I tried a different approach: a per-UTXO memoized DFS (one `bool` cached per tx, shared across UTXOs so each ancestor is visited once) instead of the topological accumulator pass. Same `does_taint`/`is_settled` contract, and I checked it gives the same `Eligibility` as `classify_outpoints` on a diamond. Benchmarked both across a few topologies:

| topology | size | memoized | classify_outpoints | speedup |
|---|---|---|---|---|
| fan-in | 10w×3d | 9.0 µs | 21.9 µs | 2.4x |
| fan-in | 50w×5d | 77.7 µs | 192.9 µs | 2.5x |
| fan-in | 100w×10d | 300 µs | 878 µs | 2.9x |
| fan-in | 500w×20d | 3.3 ms | 14.5 ms | 4.4x |
| disjoint | 10w×3d | 9.9 µs | 24.4 µs | 2.5x |
| disjoint | 50w×5d | 83.9 µs | 194.6 µs | 2.3x |
| disjoint | 100w×10d | 311 µs | 842 µs | 2.7x |
| disjoint | 500w×20d | 3.3 ms | 14.0 ms | 4.2x |
| untrusted-fan-in | 10w×3d | 8.1 µs | 24.2 µs | 3.0x |
| untrusted-fan-in | 50w×5d | 73.2 µs | 209.7 µs | 2.9x |
| untrusted-fan-in | 100w×10d | 295 µs | 940 µs | 3.2x |
| untrusted-fan-in | 500w×20d | 3.2 ms | 16.3 ms | 5.1x |
| diamond | 5s×20u | 13.5 µs | 39.2 µs | 2.9x |
| diamond | 10s×50u | 49.2 µs | 169.1 µs | 3.4x |
| diamond | 20s×100u | 158 µs | 627 µs | 4.0x |
| diamond | 50s×500u | 1.8 ms | 8.3 ms | 4.6x |

(`w` = number of parallel chains, `d` = depth of each chain; for diamond `s` = shared ancestor txs, `u` = UTXOs spending all of them.)

Both walk the same O(U+S) txs but the `BTreeSet` accumulator grows with the number of downstream UTXOs, while the memoized walk just stores one `bool` per tx and stops early once `does_taint` fires.

What I genuinely can't decide is *where* this belongs. It doesn't use `CanonicalAncestors` at all, just the public `tx`/`txout`, so it feels a bit out of place sitting next to `classify_outpoints` in `bdk_chain`, not really taking advantage of the ancestor-walk machinery this PR adds. It could just as easily live in `bdk_wallet`. But then direct `bdk_chain` consumers of `balance()` wouldn't benefit. 

I'm not sure if it makes sense here, wallet-side, or not at all, some thoughts?

Code + bench: https://github.com/Dmenec/bdk/tree/bench/trust-classification

evanlinjin:
@Dmenec Thank you for the detailed review, analysis and experimentation on top. Would you like to take over this work?

I agree that `trust_predicate` can be implemented externally and I think `classify_outpoints` + `Eligibility` is the right shape. I don't mind keeping `balance` for now as a convenience method.

Dmenec:
> @Dmenec Thank you for the detailed review, analysis and experimentation on top. Would you like to take over this work?
> 
> I agree that `trust_predicate` can be implemented externally and I think `classify_outpoints` + `Eligibility` is the right shape. I don't mind keeping `balance` for now as a convenience method.

Yes, I'd be happy to take it over! :)

Should I open a new PR from [the same branch as the benchmarks above,](https://github.com/Dmenec/bdk/tree/bench/trust-classification) or is there another way you'd prefer to hand off?

evanlinjin:
@Dmenec no preference at all. Do whatever is convenient - as long as the work is there and well communicated.

However, for convenience, it's probably better you create a new PR.

Let me know if you would like me to close this one.

nymius (commented):
cACK 71c35d9b420971d795eb913dc3404c42c7ea1015

Thanks for the suggestions @evanlinjin 

The `Elegibility` category, is something that came up before, but we didn't consider moving out of the `Balance` enum.

I prefer the `memoized` option from _@Dmenec, as I would like the canonical sort detached from any notion of `Ancestor` or `Descendant` walk, and looking at the benchmark results, the streamed classification is more performant than the "full" classification expressed here.


> - The longer-term idea is that **`classify_outpoints` is the real API** and `Balance` is just one possible fold — wallets that want categories like "locked"/"reserved" aggregate themselves. This PR keeps `Balance` (as a fold) for now; a follow-up could move it down into `bdk_wallet` and drop it from `bdk_chain`.

I will keep the PRs related focused on the correct classification of trusted txouts, and move the coin control changes to a different PR.

> - Is replacing `trust_predicate` with ancestry-based `does_taint` the right call, or should both coexist? (`does_taint` can't express per-output trust, but per-tx ancestry is arguably the more principled model.)

I don't think they should coexist, @Dmenec [has implemented a draft of how they could work together to do the same thing `trust_predicate` was doing before](https://github.com/bitcoindevkit/bdk/commit/3ec9ccde4e7f802a6148d03536e0041104200272#diff-c5c15917b9561314ed1f8b6a87c35b198ae24bdf65ed65fe68931232fafb9649R14-R34) in its `memoized` branch.

> - `Balance::confirmed` → `Balance::settled`: worth the rename / serde break, or keep the conventional name?

Not in this PR (or related). We can discard the balance breaking change for now (`confirmed` -> `settled`) and start deprecating this fold in the next versions to later introduce it with the right names in `bdk_wallet`.

Keypoints:
- Introduce `Elegibility`, `does_taint` and `is_settled`.
- Do not introduce `CanonicalAncestors` and keep using the `memoized` version.
- Keep the benchmarks (at least for the memoized form).
- Do not change `Balance`.
- Do not consider coin control, at least for this PR. Focus on fixing: bitcoindevkit/bdk_wallet#16, bitcoindevkit/bdk_wallet#273.

evanlinjin:
Replaced by #2246

PR #2236, opened 2026-06-28
feat(core): add non-consuming SyncRequest iterators

### LINKED ISSUE #2220: Consider adding visibility into the `SyncRequest` type
Once created, a `SyncRequest` cannott be inspected to know which spks will be synced. The inspect callback provides that information but only once the sync is already in motion and there is no way to query the request at rest.

Note that for `Wallet` users there is a workaround: `wallet.spk_index().revealed_spks(..)` gives the full list of spks, but this requires going back to the wallet rather than reading the already-constructed request, and is not available to callers working directly with bdk_core.

Proposed draft addition to `SyncRequest`:

```rust
pub fn iter_spks(&self) -> impl Iterator<Item = &(I, ScriptBuf)> {
    self.spks.iter()
}
```

For symmetry, non-consuming iterators for txids and outpoints could be added at the same time, potentially with:

```rust
pub fn iter_txids(&self) -> impl Iterator<Item = &Txid> {
    self.txids.iter()
}

pub fn iter_outpoints(&self) -> impl Iterator<Item = &OutPoint> {
    self.outpoints.iter()
}
```

Note that counts for all three are already available via progress(). This issue proposes only to add access to the actual contents without consuming the request.

**Impact**
- [ ] Developer experience / maintainability

- [x] Yes
- [ ] No

- [ ] Electrum
- [ ] Esplora
- [ ] Bitcoin Core RPC

alok844937-design:
### Description
This PR introduces dedicated consuming and non-consuming iteration APIs for `SyncRequest` while preserving backwards compatibility, as proposed in #2220.

Previously, `SyncRequest` exposed `iter_txids(&mut self)` and `iter_outpoints(&mut self)` as consuming iterators(draining the request via `SyncIter`), despite their `iter_*` naming suggesting read-only iteration. At the same time, there was no API for inspecting pending `spks`, `txids`, or `outpoints` without consuming the request.

This PR addresses these issues while preserving backwards compatibility.

### API Changes

**Adds new read-only accessors:**
- `SyncRequest::spks(&self)`
- `SyncRequest::txids(&self)`
- `SyncRequest::outpoints(&self)`

**Renames the consuming API to:**
- `SyncRequest::drain_txids(&mut self)`
- `SyncRequest::drain_outpoints(&mut self)`

**Reintroduces:**
- `SyncRequest::iter_txids(&mut self)`
- `SyncRequest::iter_outpoints(&mut self)`

as deprecated compatibilty wrappers that delegate to the corresponding `drain_*` methods. This preserves existing downstream code while providing a migration path to the clearer API.

### Backend updates
Updated all in-tree call-sites to use the new consuming APIs:
- `bdk_esplora` (blocking and async)
- `bdk_electrum`

### Testing
Added tests verifying that:
- `spks()`, `txids()`, and `outpoints()` are non-consuming.
- Repeated calls to the read-only accessors return consistent results.
- `drain_txids()` continues to consume all items as before.

### Design rationale
This follows the Rust API Guidelines:

- `drain_*` clearly indicates consuming iteration over the request contents.
- `spks()`, `txids()`, and `outpoints()` provide borrowed, read-only access.
- Existing `iter_*` APIs remain available as deprecated compatibility wrappers, avoiding an immediate breaking change for downstream users.

Fixes #2220

### Notes to the reviewers
1. This revision follows the API direction suggested during review:
 - consuming methods renamed to `drain_*`
 - read-only methods exposed as `spks()`, `txids()`, and `outpoints()`
 - `iter_*` methods retained as deprecated compatibility wrappers

2. The read-only accessors return `impl ExactSizeIterator + '_`, matching the capabilities of the underlying collections.
3. I observed an intermittent failure of `bdk_electrum`'s `test_sync` during local testing (`trusted_pending` vs `confirmed` balance assertion). Re-running the same test on the same commit passed without any code changes, suggesting pre-existing test flakiness rather than a regression introduced by this PR.

**Verified locally with:**
-  `cargo fmt`
- `cargo build`
- `cargo test -p bdk_core --test test_spk_client`
- multiple runs of `cargo test -p bdk_electrum --test test_electrum`

### Added
- Added `SyncRequest::spks`, `SyncRequest::txids`, and `SyncRequest::outpoints` for non-consuming inspection of pending sync data.

### Changed
- Added `SyncRequest::drain_txids` and `SyncRequest::drain_outpoints` as the consuming iterator APIs.
- Deprecated `SyncRequest::iter_txids` and `SyncRequest::iter_outpoints` in favor of `drain_*` for consuming iteration or the new read-only accessors.

### Checklists
**All Submissions**
- [x] I followed the contribution guidelines.

**New Features**
- [x] I've added tests for the new feature.
- [x] I've added docs for the new feature.

**Bugfixes**
- [x] I'm linking the issue being fixed by this PR.

### DIFF
--- a/crates/core/tests/test_spk_client.rs
+++ b/crates/core/tests/test_spk_client.rs
@@ -1,6 +1,4 @@
-use bdk_core::bitcoin::hashes::Hash;
-use bdk_core::bitcoin::{BlockHash, OutPoint, Txid};
-use bdk_core::spk_client::{FullScanResponse, SyncRequest, SyncResponse};
+use bdk_core::spk_client::{FullScanResponse, SyncResponse};
 
 #[test]
 fn test_empty() {
@@ -11,57 +9,5 @@ fn test_empty() {
     assert!(
         SyncResponse::<()>::default().is_empty(),
         "Default `SyncResponse` must be empty"
-    );
-}
-
-#[test]
-fn test_spks_does_not_consume() {
-    let spk1 = bitcoin::ScriptBuf::new();
-    let spk2 = bitcoin::ScriptBuf::new();
-
-    let request: SyncRequest<u32, BlockHash> = SyncRequest::builder()
-        .spks_with_indexes(vec![(0u32, spk1), (1u32, spk2)])
-        .build();
-
-    assert_eq!(request.spks().count(), 2);
-    assert_eq!(request.spks().count(), 2, "must not consume");
-}
-
-#[test]
-fn test_txids_does_not_consume() {
-    let txid1 = Txid::from_byte_array([0x01; 32]);
-
-    let request: SyncRequest<(), BlockHash> = SyncRequest::builder().txids(vec![txid1]).build();
-
-    assert_eq!(request.txids().count(), 1);
-    assert_eq!(request.txids().count(), 1, "must not consume");
-}
-
-#[test]
-fn test_outpoints_does_not_consume() {
-    let outpoint1 = OutPoint::null();
-
-    let request: SyncRequest<(), BlockHash> =
-        SyncRequest::builder().outpoints(vec![outpoint1]).build();
-
-    assert_eq!(request.outpoints().count(), 1);
-    assert_eq!(request.outpoints().count(), 1, "must not consume");
-}
-
-#[test]
-fn test_drain_txids_still_consumes_all_items() {
-    let txid1 = Txid::from_byte_array([0x01; 32]);
-    let txid2 = Txid::from_byte_array([0x02; 32]);
-
-    let mut request: SyncRequest<(), BlockHash> =
-        SyncRequest::builder().txids(vec![txid1, txid2]).build();
-
-    assert_eq!(request.txids().count(), 2);
-
-    let consumed: Vec<_> = request.drain_txids().collect();
-    assert_eq!(
-        consumed.len(),
-        2,
-        "drain_txids must still consume all items"
     );
 }
--- a/crates/core/tests/test_spk_client.rs
+++ b/crates/core/tests/test_spk_client.rs
@@ -1,4 +1,6 @@
-use bdk_core::spk_client::{FullScanResponse, SyncResponse};
+use bdk_core::bitcoin::hashes::Hash;
+use bdk_core::bitcoin::{BlockHash, OutPoint, Txid};
+use bdk_core::spk_client::{FullScanResponse, SyncRequest, SyncResponse};
 
 #[test]
 fn test_empty() {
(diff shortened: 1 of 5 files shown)


alok844937-design:
@evanlinjin PTAL

evanlinjin (changes requested):
Thanks for picking this up.

This PR silently changes `iter_...` methods from consuming methods to non-consuming. This will be a problem for downstream callers.

Additionally, `into_...` is the wrong naming convention. As per [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/naming.html?highlight=c-conv#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv), the `into` prefix should be reserved for methods that consume all of `self`.

To fix these naming problems, I propose the following:

* Rename consuming methods to use the `drain_` prefix.
* Reintroduce `iter_` methods that call the `drain_` methods, but mark them as deprecated. This way these changes are backwards compatible.
* Use no prefix (`spks()`, `txids()`, etc.) for read-only methods.

alok844937-design:
Thanks for the feedback!

I've updated the PR based on your suggestions:

- Renamed the consuming methods to `drain_*`.
- Added read-only accessors as `spks()`, `txids()`, and `outpoints()`.
- Reintroduced `iter_txids` and `iter_outpoints` as deprecated compatibility wrappers to preserve backwards compatibility.
- Updated the in-tree call sites, tests, and the PR description to reflect these changes.

PTAL when you have a chance. Thanks!

crates/core/tests/test_spk_client.rs
```diff
@@ -11,3 +13,55 @@ fn test_empty() {
...
+
+    assert_eq!(request.txids().count(), 2);
+
+    let consumed: Vec<_> = request.drain_txids().collect();
+    assert_eq!(
+        consumed.len(),
+        2,
+        "drain_txids must still consume all items"
+    );
+}
```
evanlinjin: These tests don't earn their keep imo. It's apparent from the method's signature.

evanlinjin (changes requested):

crates/core/tests/test_spk_client.rs
alok844937-design: Thanks for the feedback! I removed those tests since the non-consuming behavior is already apparent from the method signatures. I've updated the PR accordingly. PTAL when you have a chance.

alok844937-design:
Hi @evanlinjin, After removing the tests per your suggestion, the Codecov patch coverage dropped because these new accessors are no longer exercised. Would you prefer adding a small focused test for the new APIs, or is the coverage drop acceptable in this case?

alok844937-design:
@evanlinjin friendly follow up on this PR

alok844937-design:
@evanlinjin PTAL

PR #2237, opened 2026-07-03
chore(chain): make TxAncestors::new_include_root public

### LINKED ISSUE #2216: Make TxAncestors::new_include_root public
**Describe the enhancement**  
TxGraph::walk_ancestors skips the root transaction, it starts directly from the parents of the tx passed in. There is currently no public way to traverse ancestors including the root itself.

**Use case**  
In bitcoindevkit/bdk_wallet@15beb0b5, `balance()` needs to inspect the inputs of every transaction in a chain, including the root, to determine whether they belong to the wallet. By including the root, the same inspection logic would not be duplicated. 

I have a few ideas to discuss:

1. **Make `TxAncestors::new_include_root` public.** It already exists as `pub(crate)`, so the change would be minimal. Not sure if there's a reason it was kept internal.

2. **Add a wrapper method on TxGraph (e.g. `walk_ancestors_include_root`)**, follows the same pattern as `walk_ancestors` and keeps `TxAncestors` as an implementation detail.

**Impact**
- Developer experience / maintainability

alok844937-design:
### Description

Makes `TxAncestors::new_include_root` public. The `TxAncestors` struct itself is already `pub`, but its `new_include_root` constructor was `pub(crate)`, preventing external users from constructing an ancestor iterator that includes the root transaction. This aligns the constructor's visibility with the struct's.

Closes #2216

### Notes to the reviewers

Only `TxAncestors::new_include_root` was changed. `TxDescendants::new_include_root` (and the corresponding `new_exclude_root` methods on both types) were left untouched since the issue scope was limited to `TxAncestors`.

### Changelog notice

- Made `TxAncestors::new_include_root` public. (#2216) 

#### All Submissions:

#### New Features:

N/A - this PR only changes the visibility of an existing constructor, it does not add new functionality.

#### Bugfixes:

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -1209,7 +1209,7 @@ where
     F: FnMut(usize, Arc<Transaction>) -> Option<O>,
 {
     /// Creates a `TxAncestors` that includes the starting `Transaction` when iterating.
-    pub(crate) fn new_include_root(
+    pub fn new_include_root(
         graph: &'g TxGraph<A>,
         tx: impl Into<Arc<Transaction>>,
         filter_map: F,


alok844937-design:
@evanlinjin PTAL

alok844937-design:
**Note:** I saw https://github.com/bitcoindevkit/bdk/pull/2218 where the original issue author explored an alternative (adding a `walk_ancestors_from_root` wrapper method) but closed it while still evaluating which approach — direct visibility change vs. a wrapper vs. a memoized walk — was best. This PR implements the direct visibility change, which is the minimal, non-breaking option.

evanlinjin:
Closing as @Dmenec is already working around this issue.

PR #2239, opened 2026-07-09
feat(fuzz): add fuzzing harness for `LocalChain` with CI job

erickcestari:
### Description

Adds an initial fuzzing harness for `bdk_chain` and a CI job that runs it.

Two fuzz targets in a new `fuzz/` crate:

- `local_chain_apply_update`: builds a `LocalChain<BlockHash>` and applies random sequences of `apply_update`, `insert_block`, `disconnect_from`, `apply_header(_connected_to)`, including updates derived from the chain's own tip to hit the `merge_chains` `eq_ptr` fast path.
- `local_chain_apply_update_header`: fuzzes `LocalChain<Header>`, exercising placeholder resolution in `apply_update` and `CheckPoint::entry_iter` via a virtual chain of linked headers with random reorgs.

After every operation, the harness asserts that a successful op's changeset reconstructs the post-state from the pre-state, a failed op leaves the chain untouched, and chain invariants hold (strictly decreasing heights, genesis present, `is_block_in_chain` consistency).

Supports AFL, honggfuzz, and libFuzzer via cargo features. The CI job runs each fuzzer on each target for 5 minutes as a smoke test.

### Notes to the reviewers

- The `fuzz` crate is its own workspace (`publish = false`) so fuzzer deps don't affect the main workspace.

### Changelog notice

- Added a fuzzing harness for `LocalChain` (AFL, honggfuzz, libFuzzer) with a CI job.

**Coverage report**

image

cc @oleonardolima

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -337,3 +337,70 @@ jobs:
           cache: true
       - name: Check docs
         run: RUSTDOCFLAGS='-D warnings' cargo doc --workspace --all-features --no-deps
+
+  fuzz:
+    needs: prepare
+    name: Fuzz (${{ matrix.fuzzer }}, ${{ matrix.target }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        fuzzer: [afl, honggfuzz, libfuzzer]
+        target: [local_chain_apply_update, local_chain_apply_update_header]
+    env:
+      FUZZ_TARGET: ${{ matrix.target }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - name: Install Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          # cargo-fuzz (libFuzzer) requires nightly; AFL and honggfuzz work on stable.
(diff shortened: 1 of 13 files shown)


oleonardolima:
It's still missing an updated CI, fixing the conflicts and running each target+harness for an `x` cycles instead of time, also it requires a proper README.md for users on how to run the fuzzer, and how to add new targets.

fuzz/fuzz_targets/local_chain_apply_update.rs
```diff
@@ -0,0 +1,202 @@
...
+    DisconnectFrom,
+    /// `apply_header` with a header that usually connects to an existing checkpoint.
+    ApplyHeader,
+    /// `apply_header_connected_to` with an arbitrarily picked connection point.
+    ApplyHeaderConnectedTo,
+    /// `apply_update` with an update derived by mutating the chain's own tip, so the
+    /// update shares `Arc` nodes with the original and exercises `merge_chains`'
+    /// `eq_ptr` fast path.
+    ApplyDerivedUpdate,
+}
```
oleonardolima, line 32: @erickcestari we probably should also add coverage for the changeset operations/APIs.

luisschwab:
@erickcestari needs rebase

luisschwab:
Put `bdk_chain` targets under a `chain` folder so we can keep it neat when adding targets from other crates.

erickcestari:
> Put `bdk_chain` targets under a `chain` folder so we can keep it neat when adding targets from other crates.

Done! Looking way better now

PR #2240, opened 2026-07-09
fix(chain): skip evicted txs in leftover canonicalization

noahjoeris:
### Description

`CanonicalStage::LeftOverTxs` doesn't filter evicted txs and can mark them canonical. So a stale-anchored tx that was evicted/replaced could still show as unconfirmed.

Example: a tx confirms, then a reorg leaves only a stale anchor. Leftover treats it as unconfirmed. If it's later RBF'd or missing from the mempool, leftover still marks it unconfirmed.

### Changelog notice
- Add `TxNode::is_evicted()` for determining mempool eviction state.
- Skip evicted transactions during leftover canonicalization.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/canonical_task.rs
+++ b/crates/chain/src/canonical_task.rs
@@ -119,7 +119,12 @@ impl<'g, A: Anchor> ChainQuery for CanonicalTask<'g, A> {
                 }
                 CanonicalStage::LeftOverTxs => {
                     if let Some((txid, tx, height)) = self.unprocessed_leftover_txs.pop_front() {
-                        if !self.is_canonicalized(txid) && !tx.is_coinbase() {
+                        let is_evicted = self
+                            .tx_graph
+                            .get_tx_node(txid)
+                            .expect("leftover transaction must exist")
+                            .is_evicted();
+                        if !self.is_canonicalized(txid) && !tx.is_coinbase() && !is_evicted {
                             let observed_in = ObservedIn::Block(height);
                             self.mark_canonical(
                                 txid,
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -225,6 +225,20 @@ pub struct TxNode<'a, T, A> {
     pub last_evicted: Option<u64>,
 }
 
+impl<T, A> TxNode<'_, T, A> {
+    /// Whether the transaction has been evicted from the mempool.
+    ///
+    /// Only mempool observation timestamps are considered; anchors have no effect. A transaction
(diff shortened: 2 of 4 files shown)


crates/chain/src/canonical_task.rs
```diff
@@ -259,6 +262,18 @@ impl<'g, A: Anchor> CanonicalTask<'g, A> {
...
+        let node = self
+            .tx_graph
+            .get_tx_node(txid)
+            .expect("leftover transaction must exist");
+        match (node.last_seen, node.last_evicted) {
+            (_, None) => false,
+            (Some(seen), Some(evicted)) => evicted >= seen,
+            (None, Some(_)) => true,
+        }
+    }
```
evanlinjin: Two things:

1. The `_leftover` suffix is wrong here - the method's logic has no understanding of what is leftover or not. 
2. The `last_evicted >= last_seen` check is also used by [`TxGraph::txids_by_descending_anchor_height`](https://github.com/bitcoindevkit/bdk/blob/1acb4d00af40c317f02de9fef678cb72edbdc47a/crates/chain/src/tx_graph.rs#L1003-L1007) but the logic there did not consider the `last_seen = None, last_evicted = Some` case (which is wrong).

Possible directions:

1. Introduce `TxNode::is_evicted -> bool` method which would be a pure mempool-observation predicate (does not take anchoring into account - docs need to be clear about this). Thus, getting rid of `is_evicted_leftover`.
2. Rename `is_evicted_leftover` to just `is_evicted` and create a ticket to introduce `TxNode::is_evicted`.

evanlinjin (changes requested):
Thanks for finding and fixing this. These changes are logically correct. I have some suggestions with two possible directions.

crates/chain/src/canonical_task.rs
noahjoeris: Thanks Evan, that sounds good. I added `TxNode::is_evicted`. Though I'm not sure what you mean here [TxGraph::txids_by_descending_anchor_height](https://github.com/bitcoindevkit/bdk/blob/1acb4d00af40c317f02de9fef678cb72edbdc47a/crates/chain/src/tx_graph.rs#L1003-L1007), could you elaborate?

evanlinjin: @noahjoeris I meant `txids_by_descending_last_seen`, but I was wrong about that! Ignore me there.

evanlinjin (approved):
ACK 2a8512b15706e3f05a4a506af0293c2d33d12581

PR #2241, opened 2026-07-09
feat(chain-sources): fetch headers at caller-requested heights (#2200)

### LINKED ISSUE #2200: Allow chain sources to fetch blocks at caller-requested heights
## Context

Related to #2199. That issue covers continuous prev-11 emission during sync. This issue covers the underlying primitive: letting a caller ask a chain source for blocks at specific heights, on demand.

## Use case

A wallet wants MTP only when it actually matters — e.g., when constructing a spend that uses relative timelocks, or surfacing confirmation timestamps in a UI. Paying for prev-11 fetches on every routine sync is wasteful when the wallet may go long stretches without needing MTP at all.

The desired flow:

1. Wallet syncs normally with no prev-11 emission. Local chain has confirmation blocks but no MTP windows.
2. When the wallet needs MTP (e.g., before a spend), it canonicalizes and discovers some confirmed txs (and/or the tip) have `None` MTPs because the surrounding heights aren't in the local chain.
3. Wallet asks the chain source to fetch a chain update covering those specific heights, passing its current local checkpoint chain as the baseline.
4. Wallet applies the update; canonicalization now yields `Some(_)` MTPs.

## Goal

Each chain source exposes a way to request a chain update for a caller-supplied set of heights. The returned update is reorg-aware (anchored to the current best chain) and applies cleanly to a `LocalChain<Header>`.

## Scope

- `bdk_electrum`
- `bdk_esplora`

## Design notes

- **Input shape.** Two things: the caller's local checkpoint chain (its current `CheckPoint`) and the set/iterator of heights it wants. The checkpoint serves a dual purpose — it tells the source which heights the caller already has (so they aren't refetched) and provides the reorg anchor for building the update.
- **Output shape.** The same chain-update type each source already produces, so application is uniform.
- **Reorg safety.** Heights must be resolved against the source's current view of the best chain; if a requested height is below a reorg point relative to the caller's checkpoint, the update should reflect that.
- **Composability with #2199.** If both this and #2199 exist, the prev-11 builder flag can be implemented in terms of this primitive — or they can coexist as two entry points.

## Acceptance criteria

- Each crate exposes a method that takes the caller's checkpoint and a set of heights, and returns a chain update.
- Returned update applies via the existing `LocalChain` apply path.
- Test covering: requesting heights `[H1, H2, H3]` yields an update whose checkpoint chain contains exactly those heights (plus whatever connecting tip is needed for application).
- Test covering: a height already represented in the caller's checkpoint chain does not trigger a fetch.
- Test covering: end-to-end MTP — sync without prev-11, canonicalize and find missing MTPs, fetch the missing heights, re-canonicalize, all MTPs present.

ANAMASGARD:
Closes #2200

### Description

Adds `fetch_headers_at_heights(local_tip, heights) -> CheckPoint<Header>` to `bdk_bitcoind_rpc`, `bdk_electrum`, and `bdk_esplora`.

Wallets can fetch missing block headers on demand (for MTP, relative timelocks, confirmation timestamps) without paying for prev-11 fetches on every sync. The returned update is reorg-aware and applies via `LocalChain::apply_update`.

**API contract:**
- Skips heights already in `local_tip`
- No-op when all requested heights are known and no reorg → returns `local_tip` unchanged (empty `ChangeSet`)
- Update may include bridge/reorg blocks beyond requested heights; no Esplora tip-suffix padding

**Shared helpers in `bdk_core`:** `bridge_heights`, `build_fetch_headers_update`

### Notes to the reviewers

- `bdk_bitcoind_rpc`: standalone `fetch_headers_at_heights` (not on `Emitter::next_block`)
- `bdk_electrum`: `BdkElectrumClient::fetch_headers_at_heights` with `batch_block_header`
- `bdk_esplora`: new `headers_at_heights` module; blocking + async trait methods
- Esplora intentionally omits the ~10-block tip suffix used in normal `sync`/`full_scan`

### Local verification : VERIFIED — all tests passed 

```
cargo test -p bdk_bitcoind_rpc fetch_headers_at_heights
cargo test -p bdk_electrum fetch_headers_at_heights
cargo test -p bdk_esplora fetch_headers_at_heights
cargo clippy -p bdk_core -p bdk_bitcoind_rpc -p bdk_electrum -p bdk_esplora --all-targets -- -D warnings
```

image

### Changelog notice

- **Added** `fetch_headers_at_heights` to `bdk_bitcoind_rpc`, `bdk_electrum`, and `bdk_esplora` for on-demand `CheckPoint<Header>` fetches at caller-requested heights
- **Added** `bridge_heights` and `build_fetch_headers_update` helpers in `bdk_core`

#### All Submissions:

#### New Features:

```
---

### DIFF
--- a/crates/bitcoind_rpc/src/fetch_headers.rs
+++ b/crates/bitcoind_rpc/src/fetch_headers.rs
@@ -1,0 +1,128 @@
+//! On-demand header fetch at caller-requested heights.
+
+use bdk_core::collections::{BTreeMap, BTreeSet};
+use bdk_core::{bridge_heights, build_fetch_headers_update, CheckPoint};
+use bitcoin::block::Header;
+use bitcoincore_rpc::RpcApi;
+use core::ops::Deref;
+
+use alloc::vec::Vec;
+
+/// Error returned by [`fetch_headers_at_heights`].
+#[derive(Debug)]
+pub enum FetchHeadersError {
+    /// An RPC error occurred.
+    Rpc(bitcoincore_rpc::Error),
+    /// No common ancestor was found between `local_tip` and the node's best chain.
+    NoAgreement,
+}
+
+impl core::fmt::Display for FetchHeadersError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Self::Rpc(err) => write!(f, "bitcoind rpc error: {err}"),
+            Self::NoAgreement => write!(f, "no agreement point with bitcoind best chain"),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for FetchHeadersError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            Self::Rpc(err) => Some(err),
+            Self::NoAgreement => None,
+        }
+    }
+}
+
+impl From<bitcoincore_rpc::Error> for FetchHeadersError {
+    fn from(err: bitcoincore_rpc::Error) -> Self {
+        Self::Rpc(err)
+    }
+}
+
+/// Fetch a reorg-aware [`CheckPoint<Header>`] update covering `heights`.
+///
+/// Heights already present in `local_tip` are not fetched from the network. When every requested
+/// height is already known and no reorg is detected, returns `local_tip` unchanged (a no-op update
+/// that applies with an empty [`bdk_chain::local_chain::ChangeSet`]).
+///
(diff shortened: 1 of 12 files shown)


evanlinjin:
Thank you for attempting this.

For electrum, is there a reason for diverging from the pre-existing block-fetching logic? What do you think about repurposing `fetch_tip_and_latest_blocks` with a provided `tip_height`?

For bitcoind, I don't think individual block fetching makes sense since both the emitter and filteriter emit all heights from the start height anyway. Let's remove the bitcoind changes.

evanlinjin:
Closing due to lack of response.

PR #2244, opened 2026-07-13
Update block Emitter

### LINKED ISSUE #2176: [bitcoind_rpc] Plans to stabilize the Emitter API?
This issue tracks the work items identified in a full review of lib.rs and test_emitter.rs on master (de7a89fd), separated into 3 broad categories: 1) Internal fixes, 2) API + Documentation, 3) Test coverage.

### 1. Quirks (needs investigation)

- [ ] **`mempool_at` consistency loop**

The inner `loop` in `mempool_at` calls 5 RPCs per iteration and retries immediately with no sleep, no ceiling on retries, and the `HashSet<Txid>` allocation is repeated on every failed iteration. It doesn't remove a race condition that exists between the time of getting the raw mempool and the time the full transactions are retrieved from the node. 

- [ ] **`let mut` declarations before the loop**

This pattern is legal but unusual Rust. The loop is really "retry until consistent snapshot"— an `fn fetch_consistent_mempool_snapshot()` that returns a named struct would be cleaner.

- [ ] **`AgreementPointNotFound` arm creates an unprotected infinite loop**

`poll` responds to `AgreementPointNotFound` by resetting to genesis and looping. If genesis is somehow inaccessible or the node has a persistent inconsistency, this spins forever. Although it's a pathological case, the recommendation would be to return a hard error after a bounded number of consecutive failures.

- [ ] **`connected_to()` returns itself when checkpoint has no parent**

The fallback `None => self.checkpoint.block_id()` (returns itself) is unreachable via the emitter in practice, since every emitted checkpoint gets pushed onto the last cp. For correctness, should use `(block_height - 1, block.header.prev_blockhash)` directly which is consistent with the intended block-by-block behavior.

---

### 2. API Design + Documentation

- [ ] **Documentation for `mempool` refers to first-seen unix timestamps** 

`MempoolEvent::update` originally documented the timestamp as a "first-seen unix timestamp". For transactions in `mempool_snapshot`, the current `sync_time` is used on every poll — not the time the tx was first observed.

- [ ] **Confusing `start_height` semantics**

`start_height` was intended as a birthday/scan-start height: "don't emit blocks before this height." PR [#2167](https://github.com/bitcoindevkit/bdk/pull/2167) exposed a bug where, after a reorg that drops the agreement point below `start_height`, the emitter skips directly to `start_height` rather than re-emitting the invalidated heights. The proposed fix modifies `start_height` internally at runtime to the agreement height, circumventing the stated purpose of the parameter and making the behavior difficult to explain.

Two options:

1. Remove `start_height` entirely. The birthday/scan-start is encoded directly in `last_cp`: callers build a checkpoint containing the birthday height. The emitter scans from there naturally.

2. Add an alternate constructor that does not take `start_height` and relies solely on `last_cp` for the start height, keeping the existing constructor for backward compatibility.

- [ ] **`mempool_at` doc should carry the canonical eviction guard explanation**

`mempool_at` contains the actual `at_tip` check and eviction logic, but its doc just says _"no-std version of `mempool`"_. The eviction-guard semantics (withheld until emitter reaches tip, rationale) belong on `mempool_at`, with `mempool()` cross-referencing it.

- [ ] **`NO_EXPECTED_MEMPOOL_TXS` constant adds friction, not clarity**

The constant is a typed alias for `core::iter::empty()` hardcoded to `T = Arc<Transaction>`. This causes type-inference friction when callers pass `Transaction` directly.

- [ ] **`BlockEvent<B>` generic is vestigial**

`BlockEvent<B>` exists solely to support the private `poll<C, V, F>` closure abstraction. `next_block` is the only public consumer and always returns `BlockEvent<Block>`.

- [ ] **`poll` / `poll_once` are undocumented**

These are the algorithmic heart of the emitter. Working branch adds state-machine docs explaining entry invariants, what each `PollResponse` variant means, and the reorg-detection flow.

- [ ] **CheckPoint should be generalized for any data type** #2105

- [ ] **Move away from outdated `bitcoincore_rpc::Client`** #2119

---

### 3. Test Coverage

- [ ] **Stale `mempool_avoids_re_emission` doc comment**

The test `mempool_avoids_re_emission` has a doc comment that says "mempool txs should not be re-emitted" but its own assertions loop twice and assert that the same txids appear on both calls. Should be renamed to something like mempool_update_is_full_snapshot and the doc rewritten to describe what it actually tests: "every call to mempool() returns all currently-known unconfirmed transactions, including ones returned on previous calls."

- [ ] **Stale `test_into_tx_graph` doc comment**

The comment references `EmittedUpdate::into_tx_graph_update` — a deleted API. The test body is fine. Comment should be rewritten to describe what the test actually covers. Additionally `addr_1` and `addr_2` are generated and tracked but never funded, adding noise without coverage.

- [ ] **Stale open design question as TODO in test doc**

`ensure_block_emitted_after_reorg_is_at_reorg_height` contains `TODO: If the reorg height is lower than the fallback height...` — this is not a test concern, it is a design question. Should be removed from the test and tracked as a separate issue.

- [ ] **Behind-tip eviction guard is untested**

`test_expect_tx_evicted` tests eviction only when already at tip. The complementary scenario — evict a tx while the emitter is still behind tip, assert `evicted` is empty in that window, advance to tip, then assert the eviction is reported — is untested.

- [x] **`expected_mempool_txs` eviction path is not end-to-end tested**

The unit test `test_expected_mempool_txids_accumulate_and_remove` verifies accumulation and removal on confirmation, but there is no test that: (1) passes a known-unconfirmed tx to `Emitter::new` via `expected_mempool_txs`, (2) evicts that tx from the node's mempool, and (3) asserts it appears in `MempoolEvent::evicted`. Edit: Covered by `test_expect_tx_evicted`.

- [ ] **Missing test coverage for `start_height`**

No test verifies the fundamental birthday guarantee: an emitter with `last_cp` at height N must not emit any block at or below N.

- [ ] **Missing test coverage for `BlockEvent::connected_to()`**

- [ ] **`mempool_at` is never called in integration tests**

`mempool_at` is never called in integration tests to verify that each mempool tx carries the custom `sync_time` passed into the API.

ValuedMammal:
### Description

**Replace `bitcoincore-rpc` with `bdk_bitcoind_client`** - `Emitter` and `FilterIter` now borrow a concrete `&bdk_bitcoind_client::Client` instead of being generic over `C: Deref<Target: RpcApi>`. Core v28 is the default target; passthrough features (`29_0`, `30_0`) let callers target newer nodes.

**Generic `CheckPoint<B>` in `Emitter` and `FilterIter`** - Both types now carry a block-data type parameter `B: From<Header>`. This lets callers thread full `Header`s through emissions rather than only `BlockHash`es. `BlockEvent<B>` and `bip158::Event<B>` follow the same bound. An `as_header` helper in `bip158.rs` reconstructs a `Header` from `GetBlockHeaderVerbose` for the emitted checkpoint.

**Tidy poll loop and `Emitter` API**:
- `EmitterError` is introduced; `next_block` and mempool methods now return it
- `EmitterError::AgreementNotFound` surfaces the wrong-network case instead of silently replacing the passed-in checkpoint with a height-0 genesis checkpoint (bounded by `MAX_AGREEMENT_FAILURES = 3`)
- `start_height` is converted from a constructor parameter to a builder method; `Emitter::new` defaults to `last_cp.height()`
- `BlockEvent::connected_to()` is fixed to derive from `(height - 1, prev_blockhash)` when no previous checkpoint exists
- `raw_mempool()` extracted as a private helper for tip-consistent snapshots
- `NO_EXPECTED_MEMPOOL_TXS` removed (was hurting type inference at call sites)

**Documentation** — README rewritten with a usage overview and a runnable example, wired in as crate-level docs via `#![doc = include_str!("../README.md")]`. Poll state-machine docs (`PollResponse`, `poll_once`, `poll`), `next_block`, and `connected_to` are tidied throughout.

**Test rewrite** — `test_emitter.rs` rewritten with 14 behavior-focused tests covering: block ordering, reorgs, mempool snapshots, eviction withholding behind tip, birthday checkpoint, `start_height` semantics, `AgreementNotFound` on wrong genesis, and out-of-bounds `start_height`.

**`electrsd` bumped to `0.41.0`** in `bdk_core` and `bdk_testenv`.

Fixes #2176

### Notes to the reviewers

The `bitcoincore-rpc` → `bdk_bitcoind_client` swap is the most structurally impactful change. The emission logic itself is unchanged — only the RPC call sites and response types are migrated. The generics on `Emitter<'a, B>` / `FilterIter<'a, B>` are purely additive: existing callers using `CheckPoint<BlockHash>` compile without changes (as `BlockHash: From<Header>` is already implemented).

The `AgreementNotFound` error is a deliberate behavior change — previously a sync against a wrong network would silently replace the caller's genesis checkpoint.

#### `bdk_bitcoind_rpc`
- **Changed**: Replace `bitcoincore-rpc` dependency with `bdk_bitcoind_client`; `Emitter` and `FilterIter` now borrow `&bdk_bitcoind_client::Client` directly
- **Changed**: `Emitter` is now `Emitter<'a, B>` and `BlockEvent` is `BlockEvent<B>` where `B: From<bitcoin::block::Header>`; same for `FilterIter<'a, B>` and `Event<B>`
- **Changed**: `start_height` moved from `Emitter::new` to a new builder method `Emitter::start_height`
- **Added**: `EmitterError` enum returned by `next_block`, `mempool`, and `mempool_at`
- **Added**: `EmitterError::AgreementNotFound` surfaces wrong-network / no-common-ancestor failures instead of silently replacing the checkpoint with genesis
- **Fixed**: `BlockEvent::connected_to()` now correctly derives `(height - 1, prev_blockhash)` when no previous checkpoint exists
- **Removed**: `NO_EXPECTED_MEMPOOL_TXS` constant
- **Docs**: README rewritten with usage overview and runnable example

#### `bdk_core` / `bdk_testenv`
- **Changed**: Bump `electrsd` to `0.41.0`

#### All Submissions:

#### New Features:

#### Bugfixes:


crates/bitcoind_rpc/src/emitter.rs
```diff
@@ -233,130 +246,164 @@ impl<B> BlockEvent<B> {
...
+                    self.last_cp = cp;
+                    self.last_block = Some(block_info);
+                    self.agreement_failure_count = 0;
+                }
+                PollResponse::AgreementNotFound => {
+                    self.agreement_failure_count += 1;
+                    if self.agreement_failure_count >= MAX_AGREEMENT_FAILURES {
+                        return Err(EmitterError::AgreementNotFound);
+                    }
+                    self.last_block = None;
```
tvpeter: I think for the loop to execute the `AgreementNotFound` arm, the `self.last_block` was already `None`, so resetting it to None does not change the outcome of the next call.

crates/bitcoind_rpc/src/emitter.rs
```diff
@@ -233,130 +246,164 @@ impl<B> BlockEvent<B> {
...
 
-fn poll_once(emitter: &Emitter<'_>) -> Result<PollResponse, bitcoind_client::Error> {
-    let client = emitter.client;
+impl<'a> Emitter<'a> {
+    /// Probe the node once and return the appropriate [`PollResponse`].
+    fn poll_once(&self) -> Result<PollResponse, bitcoind_client::Error> {
+        if let Some(last_block_info) = &self.last_block {
+            let next_hash = if last_block_info.height + 1 < self.start_height {
+                // enforce start height
+                self.client.get_block_hash(self.start_height)?
```
tvpeter, line 282: I think the removed check on whether the Emitter is connected to the Best Chain is not covered by the block confirmations check. Can you please verify this?

tvpeter (commented):
Thank you ValuedMammal for working on this. 

On my comment regarding the AgreementNotFound arm, my understanding is that the emitter immediately retry the node with identical request (last_cp hash) and may likely return the same outcome for all three calls, so it might be more efficient to immediately return `EmitterError::AgreementNotFound` immediately upon the first failure.

Also, I have checked part of this PR that replaces `bitcoincore-rpc` with `bdk_bitcoind_client` and it covers everything with the latest release, so I will be closing #2119 in favour of this PR.

Thank you.

ValuedMammal:
- Added back the best-chain check in the start height enforcement branch for correctness, although I didn't come up with a reliable single-threaded integration test for it
- Removed `MAX_AGREEMENT_FAILURES` and return `EmitterError::AgreementNotFound` if no agreement block is found after 1 attempt

tvpeter (approved):
ACK 5b256524b4c9b42031475ab10d426981ec20aafc

crates/bitcoind_rpc/src/emitter.rs
```diff
@@ -0,0 +1,481 @@
...
+    last_cp: CheckPoint<B>,
+
+    /// RPC result of the last-emitted block, kept for its `nextblockhash` so the next block can be
+    /// fetched without a height-to-hash lookup. `None` before the first emission, at tip, or after
+    /// the last block was reorged out.
+    last_block: Option<GetBlockVerboseOne>,
+
+    /// Consecutive polls that failed to find an agreement point. Bounds the retry loop so the
+    /// emitter surfaces [`EmitterError::AgreementNotFound`] instead of spinning forever.
+    agreement_failure_count: u8,
```
evanlinjin: This is a great idea as a feature. However, let's not include it in this PR - it's out of scope and it's incomplete (`agreement_failure_count` is not read anywhere).

ValuedMammal: fixed :)

tvpeter:
Also fixes #2290

PR #2246, opened 2026-07-22
feat(chain)!: taint-aware `CanonicalView::balance` 

Dmenec:
Solves #2267

### Description

Takes over #2235 (thanks @evanlinjin for the go-ahead). It reworks `CanonicalView::balance` to derive trust from an output's unconfirmed ancestry, and adds `classify_outpoints`, a per-output spend-eligibility classifier that `balance` becomes a thin fold over.

The old `balance` decided trust using a per-output `trust_predicate`, which cannot express transitive trust. As a result, owned outputs whose unconfirmed ancestry included foreign coins were counted as trusted. That is the root cause of the wallet trust-classification bugs (bitcoindevkit/bdk_wallet#16, bitcoindevkit/bdk_wallet#273).

The API now takes two separate predicates, one per concern:

- `does_taint(&tx)` - should this transaction be considered tainted? (e.g., because it spends a foreign unconfirmed output)
- `is_settled(&pos)` - do we consider this chain position settled / final? (generalizes `min_confirmations`)

For each unspent output, `classify_outpoints` reports its chain-level spend eligibility:

- `Settled` if considered settled according to `is_settled`
- `Immature` for a coinbase output that has not yet matured
- `Unsettled(Trust)` otherwise, where `Trust` is:
  - `Trusted` if the whole unconfirmed ancestry only spends owned coins
  - `Untrusted` if the output itself, or any unconfirmed ancestor, is tainted
  - `Unknown` if part of the ancestry is missing from the view, so trust can't be determined

`balance` then sums each output's value into the bucket corresponding to its `Eligibility`.

### Notes to the reviewers

- Trust is resolved through a self-contained ancestry walk. The traversal is memoized: each visited transaction is cached so that already-classified transactions (and their ancestors) do not need to be walked again.
- Kept `Balance::confirmed` and did not rename it to `settled`. That rename is out of scope here. The folding logic is slated to move into `bdk_wallet`, and the current `balance` function in chain will eventually be deprecated.
- `balance` also drops the `O` generic and now takes plain `OutPoint`s, since the taint predicate operates on transactions rather than per-outpoint associated data.
- `does_taint` is evaluated at most once per transaction.
- More sophisticated classification rules (e.g., for coin control or locked funds) can be built on top of `classify_outpoints`. Left for a follow-up.

### Changelog notice

- **Breaking**: `CanonicalView::balance` now takes `does_taint: impl FnMut(&CanonicalTx) -> bool` and `is_settled: impl Fn(&ChainPosition) -> bool` instead of a per-output trust predicate and `min_confirmations`, and plain `OutPoint`s instead of `(identifier, outpoint)` pairs (this drops the `O` generic). Trust is now derived from an output's unconfirmed ancestry.
- **Breaking**: added `Balance::unknown_pending`, where outputs whose trust can't be determined are counted instead of being lumped into `untrusted_pending`.
- Added `CanonicalView::classify_outpoints` and the `Eligibility` enum.

#### All Submissions:

* [x] I ran `just p` before pushing

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/chain/benches/trust_classification.rs
+++ b/crates/chain/benches/trust_classification.rs
@@ -1,0 +1,485 @@
+//! Benchmarks for `classify_outpoints` (the classifier `balance` folds over).
+//!
+//! Each group builds a synthetic canonical graph and times classifying every UTXO across a range
+//! of sizes. What we care about is how the cost grows with the unconfirmed ancestry, so the
+//! confirmed part is kept minimal and everything interesting lives in the mempool.
+
+use std::collections::BTreeMap;
+
+use bdk_chain::{
+    local_chain::LocalChain, CanonicalView, ChainPosition, ConfirmationBlockTime, TxGraph,
+};
+use bdk_testenv::{hash, utils::new_tx};
+use bitcoin::{Amount, BlockHash, OutPoint, Script, ScriptBuf, Transaction, TxIn, TxOut, Txid};
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+
+// Minimal 2-block chain. Confirmed txs just anchor at height 1.
+fn make_chain() -> LocalChain {
+    let blocks: BTreeMap<u32, BlockHash> = [(0, hash!("genesis")), (1, hash!("b1"))]
+        .into_iter()
+        .collect();
+    LocalChain::from_blocks(blocks).unwrap()
+}
+
+// The two predicates `classify_outpoints` takes, faking what a wallet would pass:
+// `does_taint` = the tx pulls in coins that aren't ours, `is_settled` = the tx is confirmed.
+#[allow(clippy::type_complexity)]
+fn does_taint_and_is_settled<'a>(
+    view: &'a CanonicalView<ConfirmationBlockTime>,
+    owned: &ScriptBuf,
+) -> (
+    impl FnMut(&bdk_chain::CanonicalTx<ChainPosition<ConfirmationBlockTime>>) -> bool + 'a,
+    impl Fn(&ChainPosition<ConfirmationBlockTime>) -> bool,
+) {
+    let owned = owned.clone();
+    let is_mine = move |spk: &Script| spk == owned.as_script();
+    let does_taint = move |c_tx: &bdk_chain::CanonicalTx<ChainPosition<ConfirmationBlockTime>>| {
+        c_tx.tx
+            .input
+            .iter()
+            .any(|txin| match view.tx(txin.previous_output.txid) {
+                Some(prev) => {
+                    !is_mine(&prev.tx.output[txin.previous_output.vout as usize].script_pubkey)
+                }
+                // input from a tx we don't have, treat it as foreign
+                None => true,
+            })
+    };
+    let is_settled =
+        |pos: &ChainPosition<ConfirmationBlockTime>| matches!(pos, ChainPosition::Confirmed { .. });
+    (does_taint, is_settled)
+}
+
+// Per-UTXO memoized classification (`classify_outpoints`).
+fn run_classify(
+    utxo_txids: &[Txid],
+    view: &CanonicalView<ConfirmationBlockTime>,
+    owned: &ScriptBuf,
+) {
+    let outpoints = utxo_txids.iter().map(|&txid| OutPoint::new(txid, 0));
+    let (does_taint, is_settled) = does_taint_and_is_settled(view, owned);
+    for item in view.classify_outpoints(outpoints, does_taint, is_settled) {
+        std::hint::black_box(item);
+    }
+}
+
+// One confirmed owned root with `width` outputs, each starting a `depth`-deep chain of unconfirmed
+// owned txs. Nothing taints. Walking back, every chain converges on the shared root.
+fn setup_fan_in(
+    width: usize,
+    depth: usize,
+) -> (
+    LocalChain,
+    TxGraph<ConfirmationBlockTime>,
+    Vec<Txid>,
+    ScriptBuf,
+) {
+    let chain = make_chain();
+    let mut tx_graph = TxGraph::default();
+    let owned = ScriptBuf::from(vec![1u8]);
+
+    let root_tx = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint::new(hash!("external"), 0),
+            ..Default::default()
+        }],
+        output: (0..width)
+            .map(|_| TxOut {
+                value: Amount::from_sat(100_000),
+                script_pubkey: owned.clone(),
+            })
+            .collect(),
+        ..new_tx(0)
+    };
+    let root_txid = root_tx.compute_txid();
+    let _ = tx_graph.insert_tx(root_tx);
+    let _ = tx_graph.insert_anchor(
+        root_txid,
+        ConfirmationBlockTime {
+            block_id: chain.get(1).unwrap().block_id(),
+            confirmation_time: 1,
+        },
+    );
+
+    let mut utxo_txids = vec![];
+    for chain_i in 0..width {
+        let mut prev = OutPoint::new(root_txid, chain_i as u32);
+        for layer in 0..depth {
+            let tx = Transaction {
+                input: vec![TxIn {
+                    previous_output: prev,
+                    ..Default::default()
+                }],
+                output: vec![TxOut {
+                    value: Amount::from_sat(90_000),
+                    script_pubkey: owned.clone(),
+                }],
+                ..new_tx((chain_i * depth + layer + 1) as u32)
+            };
+            let txid = tx.compute_txid();
+            let _ = tx_graph.insert_tx(tx);
+            let _ = tx_graph.insert_seen_at(txid, 100);
+            prev = OutPoint::new(txid, 0);
+            if layer == depth - 1 {
+                utxo_txids.push(txid);
+            }
+        }
+    }
+    (chain, tx_graph, utxo_txids, owned)
+}
+
+// Like fan_in but each chain gets its own confirmed root, so there is no shared ancestor.
+// Nothing taints.
+fn setup_disjoint(
+    width: usize,
+    depth: usize,
+) -> (
+    LocalChain,
+    TxGraph<ConfirmationBlockTime>,
+    Vec<Txid>,
+    ScriptBuf,
+) {
+    let chain = make_chain();
+    let mut tx_graph = TxGraph::default();
+    let owned = ScriptBuf::from(vec![1u8]);
+
+    let mut utxo_txids = vec![];
+    for chain_i in 0..width {
+        let root_tx = Transaction {
+            input: vec![TxIn {
+                previous_output: OutPoint::new(hash!("external"), chain_i as u32),
+                ..Default::default()
+            }],
+            output: vec![TxOut {
+                value: Amount::from_sat(100_000),
+                script_pubkey: owned.clone(),
+            }],
+            ..new_tx(chain_i as u32)
+        };
+        let root_txid = root_tx.compute_txid();
+        let _ = tx_graph.insert_tx(root_tx);
+        let _ = tx_graph.insert_anchor(
+            root_txid,
+            ConfirmationBlockTime {
+                block_id: chain.get(1).unwrap().block_id(),
+                confirmation_time: 1,
+            },
+        );
+
+        let mut prev = OutPoint::new(root_txid, 0);
+        for layer in 0..depth {
+            let tx = Transaction {
+                input: vec![TxIn {
+                    previous_output: prev,
+                    ..Default::default()
+                }],
+                output: vec![TxOut {
+                    value: Amount::from_sat(90_000),
+                    script_pubkey: owned.clone(),
+                }],
+                ..new_tx((width + chain_i * depth + layer) as u32)
+            };
+            let txid = tx.compute_txid();
+            let _ = tx_graph.insert_tx(tx);
+            let _ = tx_graph.insert_seen_at(txid, 100);
+            prev = OutPoint::new(txid, 0);
+            if layer == depth - 1 {
+                utxo_txids.push(txid);
+            }
+        }
+    }
+    (chain, tx_graph, utxo_txids, owned)
+}
+
+// An unconfirmed root paying foreign scripts. Each chain's first tx spends it, so it taints, and
+// the taint carries down the chain. Every UTXO ends up untrusted.
+fn setup_untrusted_fan_in(
+    width: usize,
+    depth: usize,
+) -> (
+    LocalChain,
+    TxGraph<ConfirmationBlockTime>,
+    Vec<Txid>,
+    ScriptBuf,
+) {
+    let chain = make_chain();
+    let mut tx_graph = TxGraph::default();
+    let owned = ScriptBuf::from(vec![1u8]);
+    let foreign = ScriptBuf::from(vec![2u8]); // not owned -> untrusted
+
+    // Unconfirmed root with a foreign input
+    let root_tx = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint::new(hash!("external"), 0),
+            ..Default::default()
+        }],
+        output: (0..width)
+            .map(|_| TxOut {
+                value: Amount::from_sat(100_000),
+                script_pubkey: foreign.clone(),
+            })
+            .collect(),
+        ..new_tx(0)
+    };
+    let root_txid = root_tx.compute_txid();
+    let _ = tx_graph.insert_tx(root_tx);
+    let _ = tx_graph.insert_seen_at(root_txid, 50);
+
+    let mut utxo_txids = vec![];
+    for chain_i in 0..width {
+        let mut prev = OutPoint::new(root_txid, chain_i as u32);
+        for layer in 0..depth {
+            let tx = Transaction {
+                input: vec![TxIn {
+                    previous_output: prev,
+                    ..Default::default()
+                }],
+                output: vec![TxOut {
+                    value: Amount::from_sat(90_000),
+                    script_pubkey: owned.clone(),
+                }],
+                ..new_tx((chain_i * depth + layer + 1) as u32)
+            };
+            let txid = tx.compute_txid();
+            let _ = tx_graph.insert_tx(tx);
+            let _ = tx_graph.insert_seen_at(txid, 100);
+            prev = OutPoint::new(txid, 0);
+            if layer == depth - 1 {
+                utxo_txids.push(txid);
+            }
+        }
+    }
+    (chain, tx_graph, utxo_txids, owned)
+}
+
+// `shared` confirmed base txs, and `utxos` outputs that each spend  all of the base txs (many
+// inputs converging on the same ancestors). Given importance on shared ancestry and nothing taints.
+fn setup_diamond(
+    shared: usize,
+    utxos: usize,
+) -> (
+    LocalChain,
+    TxGraph<ConfirmationBlockTime>,
+    Vec<Txid>,
+    ScriptBuf,
+) {
+    let chain = make_chain();
+    let mut tx_graph = TxGraph::default();
+    let owned = ScriptBuf::from(vec![1u8]);
+
+    let root_tx = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint::new(hash!("external"), 0),
+            ..Default::default()
+        }],
+        output: (0..shared)
+            .map(|_| TxOut {
+                value: Amount::from_sat(1_000_000),
+                script_pubkey: owned.clone(),
+            })
+            .collect(),
+        ..new_tx(0)
+    };
+    let root_txid = root_tx.compute_txid();
+    let _ = tx_graph.insert_tx(root_tx);
+    let _ = tx_graph.insert_anchor(
+        root_txid,
+        ConfirmationBlockTime {
+            block_id: chain.get(1).unwrap().block_id(),
+            confirmation_time: 1,
+        },
+    );
+
+    let mut base_txids = vec![];
+    for i in 0..shared {
+        let base_tx = Transaction {
+            input: vec![TxIn {
+                previous_output: OutPoint::new(root_txid, i as u32),
+                ..Default::default()
+            }],
+            output: (0..utxos)
+                .map(|_| TxOut {
+                    value: Amount::from_sat(90_000),
+                    script_pubkey: owned.clone(),
+                })
+                .collect(),
+            ..new_tx((i + 1) as u32)
+        };
+        let base_txid = base_tx.compute_txid();
+        let _ = tx_graph.insert_tx(base_tx);
+        let _ = tx_graph.insert_seen_at(base_txid, 50);
+        base_txids.push(base_txid);
+    }
+
+    let mut utxo_txids = vec![];
+    for j in 0..utxos {
+        let utxo_tx = Transaction {
+            input: base_txids
+                .iter()
+                .map(|&base_txid| TxIn {
+                    previous_output: OutPoint::new(base_txid, j as u32),
+                    ..Default::default()
+                })
+                .collect(),
+            output: vec![TxOut {
+                value: Amount::from_sat(10_000),
+                script_pubkey: owned.clone(),
+            }],
+            ..new_tx((shared + 1 + j) as u32)
+        };
+        let utxo_txid = utxo_tx.compute_txid();
+        let _ = tx_graph.insert_tx(utxo_tx);
+        let _ = tx_graph.insert_seen_at(utxo_txid, 100);
+        utxo_txids.push(utxo_txid);
+    }
+    (chain, tx_graph, utxo_txids, owned)
+}
+
+// Each UTXO sits below a short owned tail, a single tainting tx, and a deep unconfirmed foreign
+// ancestry above it. The taint is found right above the UTXO, so the classifier can stop there
+// instead of walking the whole foreign chain. Every UTXO ends up untrusted.
+fn setup_tainted_midchain(
+    width: usize,
+    depth: usize,
+) -> (
+    LocalChain,
+    TxGraph<ConfirmationBlockTime>,
+    Vec<Txid>,
+    ScriptBuf,
+) {
+    let chain = make_chain();
+    let mut tx_graph = TxGraph::default();
+    let owned = ScriptBuf::from(vec![1u8]);
+    let foreign = ScriptBuf::from(vec![2u8]);
+
+    let mut n: u32 = 0;
+    let mut utxo_txids = vec![];
+    for chain_i in 0..width {
+        let mut prev = OutPoint::new(hash!("external"), chain_i as u32);
+        for _ in 0..depth {
+            let tx = Transaction {
+                input: vec![TxIn {
+                    previous_output: prev,
+                    ..Default::default()
+                }],
+                output: vec![TxOut {
+                    value: Amount::from_sat(100_000),
+                    script_pubkey: foreign.clone(),
+                }],
+                ..new_tx(n)
+            };
+            n += 1;
+            let txid = tx.compute_txid();
+            let _ = tx_graph.insert_tx(tx);
+            let _ = tx_graph.insert_seen_at(txid, 100);
+            prev = OutPoint::new(txid, 0);
+        }
+        let taint_tx = Transaction {
+            input: vec![TxIn {
+                previous_output: prev,
+                ..Default::default()
+            }],
+            output: vec![TxOut {
+                value: Amount::from_sat(90_000),
+                script_pubkey: owned.clone(),
+            }],
+            ..new_tx(n)
+        };
+        n += 1;
+        let taint_txid = taint_tx.compute_txid();
+        let _ = tx_graph.insert_tx(taint_tx);
+        let _ = tx_graph.insert_seen_at(taint_txid, 100);
+        let utxo_tx = Transaction {
+            input: vec![TxIn {
+                previous_output: OutPoint::new(taint_txid, 0),
+                ..Default::default()
+            }],
+            output: vec![TxOut {
+                value: Amount::from_sat(80_000),
+                script_pubkey: owned.clone(),
+            }],
+            ..new_tx(n)
+        };
+        n += 1;
+        let utxo_txid = utxo_tx.compute_txid();
+        let _ = tx_graph.insert_tx(utxo_tx);
+        let _ = tx_graph.insert_seen_at(utxo_txid, 100);
+        utxo_txids.push(utxo_txid);
+    }
+    (chain, tx_graph, utxo_txids, owned)
+}
+
+fn bench_fan_in(c: &mut Criterion) {
+    let mut group = c.benchmark_group("fan_in");
+    for (width, depth) in [(10, 3), (50, 5), (100, 10), (500, 20)] {
+        let label = format!("{width}w×{depth}d");
+        let (chain, tx_graph, utxo_txids, owned) = setup_fan_in(width, depth);
+        let view = chain.canonical_view(&tx_graph, chain.tip().block_id(), Default::default());
+        group.bench_with_input(BenchmarkId::new("classify", &label), &(), |b, _| {
+            b.iter(|| run_classify(&utxo_txids, &view, &owned));
+        });
+    }
+    group.finish();
+}
+
+fn bench_disjoint(c: &mut Criterion) {
+    let mut group = c.benchmark_group("disjoint");
+    for (width, depth) in [(10, 3), (50, 5), (100, 10), (500, 20)] {
+        let label = format!("{width}w×{depth}d");
+        let (chain, tx_graph, utxo_txids, owned) = setup_disjoint(width, depth);
+        let view = chain.canonical_view(&tx_graph, chain.tip().block_id(), Default::default());
+        group.bench_with_input(BenchmarkId::new("classify", &label), &(), |b, _| {
+            b.iter(|| run_classify(&utxo_txids, &view, &owned));
+        });
+    }
+    group.finish();
+}
+
+fn bench_untrusted_fan_in(c: &mut Criterion) {
+    let mut group = c.benchmark_group("untrusted_fan_in");
+    for (width, depth) in [(10, 3), (50, 5), (100, 10), (500, 20)] {
+        let label = format!("{width}w×{depth}d");
+        let (chain, tx_graph, utxo_txids, owned) = setup_untrusted_fan_in(width, depth);
+        let view = chain.canonical_view(&tx_graph, chain.tip().block_id(), Default::default());
+        group.bench_with_input(BenchmarkId::new("classify", &label), &(), |b, _| {
+            b.iter(|| run_classify(&utxo_txids, &view, &owned));
+        });
+    }
+    group.finish();
+}
+
+fn bench_diamond(c: &mut Criterion) {
+    let mut group = c.benchmark_group("diamond");
+    for (shared, utxos) in [(5, 20), (10, 50), (20, 100), (50, 500)] {
+        let label = format!("{shared}s×{utxos}u");
+        let (chain, tx_graph, utxo_txids, owned) = setup_diamond(shared, utxos);
+        let view = chain.canonical_view(&tx_graph, chain.tip().block_id(), Default::default());
+        group.bench_with_input(BenchmarkId::new("classify", &label), &(), |b, _| {
+            b.iter(|| run_classify(&utxo_txids, &view, &owned));
+        });
+    }
+    group.finish();
+}
+
+fn bench_tainted_midchain(c: &mut Criterion) {
+    let mut group = c.benchmark_group("tainted_midchain");
+    for (width, depth) in [(10, 3), (50, 5), (100, 10), (500, 20)] {
+        let label = format!("{width}w×{depth}d");
+        let (chain, tx_graph, utxo_txids, owned) = setup_tainted_midchain(width, depth);
+        let view = chain.canonical_view(&tx_graph, chain.tip().block_id(), Default::default());
+        group.bench_with_input(BenchmarkId::new("classify", &label), &(), |b, _| {
+            b.iter(|| run_classify(&utxo_txids, &view, &owned));
+        });
+    }
+    group.finish();
+}
+
+criterion_group!(
+    benches,
+    bench_fan_in,
+    bench_disjoint,
+    bench_untrusted_fan_in,
+    bench_diamond,
+    bench_tainted_midchain
+);
+criterion_main!(benches);
--- a/crates/chain/src/canonical.rs
+++ b/crates/chain/src/canonical.rs
@@ -28,9 +28,7 @@ use alloc::vec::Vec;
 use core::{fmt, ops::RangeBounds};
 
 use bdk_core::BlockId;
-use bitcoin::{
-    constants::COINBASE_MATURITY, Amount, OutPoint, ScriptBuf, Transaction, TxOut, Txid,
-};
+use bitcoin::{constants::COINBASE_MATURITY, OutPoint, ScriptBuf, Transaction, TxOut, Txid};
 
 use crate::{spk_txout::SpkTxOutIndex, Anchor, Balance, CanonicalViewTask, ChainPosition, TxGraph};
 
@@ -434,7 +432,8 @@ impl<A: Anchor> CanonicalView<A> {
     /// * `does_taint` - Returns `true` for a transaction that pulls in untrusted funds (e.g. it
     ///   spends an output the wallet doesn't own). It drives the [`Trust`] of an unsettled output:
     ///   a tainting transaction in its ancestry makes it [`Untrusted`](Trust::Untrusted). Outputs
-    ///   with missing ancestry stay [`Unknown`](Trust::Unknown) regardless of this predicate.
+    ///   with missing ancestry stay [`Unknown`](Trust::Unknown) regardless of this predicate. Use
+    ///   [`taints_unowned`] to classify everything foreign as untrusted.
     /// * `is_settled` - Returns `true` for the [position](ChainPosition) of a transaction we
     ///   consider settled (unlikely to be replaced), for example one with enough confirmations.
     pub fn classify_outpoints<'a>(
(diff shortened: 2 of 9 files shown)


crates/chain/src/canonical.rs
```diff
@@ -22,18 +22,33 @@
...
-};
+use bitcoin::{constants::COINBASE_MATURITY, OutPoint, ScriptBuf, Transaction, TxOut, Txid};
 
 use crate::{spk_txout::SpkTxOutIndex, Anchor, Balance, CanonicalViewTask, ChainPosition, TxGraph};
 
+/// The spend-eligibility classification of a canonical output, produced by
+/// [`CanonicalView::classify_outpoints`].
+///
+/// This is a *chain-level* classification. It deliberately knows nothing about wallet policies such
+/// as locked or reserved coins; callers layer their own categories on top.
```
evanlinjin:
```suggestion
```

Let's remove this - I don't think it's useful.

crates/chain/src/canonical.rs
```diff
@@ -22,18 +22,33 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Eligibility {
+    /// A confirmed output unlikely to be replated.
```
evanlinjin:
```suggestion
    /// A confirmed output unlikely to be replaced.
```

crates/chain/src/canonical.rs
```diff
@@ -394,98 +409,194 @@ impl<A, P: Clone> Canonical<A, P> {
...
+        &self,
+        outpoints: impl IntoIterator<Item = OutPoint>,
+        mut does_taint: impl FnMut(&CanonicalTx<ChainPosition<A>>) -> bool,
+        is_settled: impl Fn(&ChainPosition<A>) -> bool,
+    ) -> impl Iterator<Item = (CanonicalTxOut<ChainPosition<A>>, Eligibility)> {
+        let utxos = outpoints
+            .into_iter()
+            .filter_map(|op| self.txout(op))
+            .filter(|txo| txo.spent_by.is_none())
+            .collect::<Vec<_>>();
```
evanlinjin: No need to collect here.

Dmenec:
Thinking about some cases where settledness makes a transaction fall into a different `Eligibility`. With foreign inputs (e.g. a tx that needs 6 `min_confirmations` to be settled, while the current tip is only 1 block ahead of it) the output ends up as `UntrustedPending`, which doesn't really make sense: it isn't pending, just not settled. The same happens with owned inputs that falls into `TrustedPending`.

I think it might be clearer to call them `TrustedUnsettled` / `UntrustedUnsettled` rather than `*Pending`.

crates/chain/src/canonical.rs
```diff
@@ -394,98 +409,194 @@ impl<A, P: Clone> Canonical<A, P> {
...
+
+        let tip = self.tip.height;
+        // Remembers whether a tx is tainted, so if two outpoints share an ancestor we only walk
+        // it once instead of redoing the same work for each outpoint.
+        let mut cache = HashMap::<Txid, bool>::new();
+        let results: Vec<_> = utxos
+            .into_iter()
```
evanlinjin, line 445: Can remove this once `utxos` is an iterator.
```suggestion
```

evanlinjin:
@Dmenec Good point. What do you think about this?

```rust
pub enum Eligibility {
    Settled,
    Immature,
    Unsettled(Trust),
}

pub enum Trust {
    Trusted,
    Untrusted,
}
```

I think this increases clarity and makes it a bit easier for call sites that only care about whether it's unsettled or not.

crates/chain/src/canonical.rs
```diff
@@ -394,98 +409,194 @@ impl<A, P: Clone> Canonical<A, P> {
...
+            .map(|txout| {
+                let eligibility = if is_settled(&txout.pos) {
+                    if txout.is_mature(tip) {
+                        Eligibility::Settled
+                    } else {
+                        Eligibility::Immature
+                    }
```
evanlinjin: The coinbase maturity check lives inside the `is_settled` check - this is wrong. Unsettled is not the same as unconfirmed.

**Example:** A confirmed transaction can be unsettled because the caller requires 3 confirmations to be classified as settled. With the current logic, this transaction will become "pending" instead of "immature".

crates/chain/src/canonical.rs
```diff
@@ -394,98 +409,194 @@ impl<A, P: Clone> Canonical<A, P> {
...
+                    stack.push(Frame::Exit(txid));
+                    for txin in &c_tx.tx.input {
+                        let parent_txid = txin.previous_output.txid;
+                        if !cache.contains_key(&parent_txid) && !pending.contains(&parent_txid) {
+                            stack.push(Frame::Enter(parent_txid));
+                        }
+                    }
+                }
+                Frame::Exit(txid) => {
+                    let c_tx = self.tx(txid).expect("present, checked in Enter");
```
evanlinjin: If we change the signature of `Exit` to also contain the transaction itself, we can avoid calling `self.tx()` again.

crates/chain/src/canonical.rs
```diff
@@ -394,98 +409,194 @@ impl<A, P: Clone> Canonical<A, P> {
...
+
+        while let Some(frame) = stack.pop() {
+            match frame {
+                Frame::Enter(txid) => {
+                    if cache.contains_key(&txid) || pending.contains(&txid) {
+                        continue;
+                    }
+                    let Some(c_tx) = self.tx(txid) else {
+                        // Not canonical
+                        cache.insert(txid, false);
```
evanlinjin: If a transaction does not exist in the `CanonicalView`, it just means the wallet has sparse data. Therefore, "not canonical" is incorrect.

We can guarantee that all transactions are canonical at this point as in line 440, we filtered by `self.txout()`.

### Proposal

Assuming we introduce the `Eligibility::Unsettled(Trust)` variant, we can add another variant to the `Trust` enum: `Trust::Unknown`.

The `cache` type will become `HashMap<Txid, Trust>` where `Untrusted` beats `Unknown` and `Unknown` beats `Trusted`.

evanlinjin (changes requested):
Thanks for pushing this forward - this is looking great.

I haven't looked too hard into the tests yet, follow-up reviews will come.

crates/chain/src/canonical.rs
```diff
@@ -394,98 +409,194 @@ impl<A, P: Clone> Canonical<A, P> {
...
+                    &mut cache,
+                ) {
+                    Eligibility::UntrustedPending
+                } else {
+                    Eligibility::TrustedPending
+                };
+                (txout, eligibility)
+            })
+            .collect();
+        results.into_iter()
```
evanlinjin: There is a way to write this to not `.collect` and then call `.into_iter` on it again.

crates/chain/src/canonical.rs
nymius: So any unconfirmed coinbase tx with an owned script pubkey should be treated as `Immature`, not `TrustedPending`, even if it doesn't have a single confirmation?

crates/chain/src/canonical.rs
nymius: > We can guarantee that all transactions are canonical at this point as in line 440, we filtered by `self.txout()`.

You mean we consider txs for which its txid appear as parents of a tx in the `CanonicalView` as canonical?

> ### Proposal
> 
> Assuming we introduce the `Eligibility::Unsettled(Trust)` variant, we can add another variant to the `Trust` enum: `Trust::Unknown`.
> 
> The `cache` type will become `HashMap<Txid, Trust>` where `Untrusted` beats `Unknown` and `Unknown` beats `Trusted`.

Do you expect wallets to also handle this category as "unresolved" balance? Or it's just an intermediate category to trigger request for those txids and be able to resolve the whole trustness of the txs chains?

In my mind we will end up considering this as `Untrusted` because we cannot claim that we know if these are tainted or not.

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -1028,13 +1028,9 @@ fn test_tx_conflict_handling() {
...
-            env.indexer.outpoints().iter().cloned(),
-            |_, txout| {
-                env.indexer
-                    .index_of_spk(txout.txout.script_pubkey.as_script())
-                    .is_some()
-            },
-            0,
+            env.indexer.outpoints().iter().map(|(_, op)| *op),
+            |_| false,
+            |pos| pos.is_confirmed(),
```
nymius, line 1043: The new closure isn't equivalent to the old one. You must check the indexing of the script pubkey.

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -481,9 +481,11 @@ fn test_list_owned_txouts() {
                 .collect::<Vec<_>>();
 
             let balance = canonical_view.balance(
-                graph.index.outpoints().iter().cloned(),
-                |_, txout| trusted_spks.contains(&txout.txout.script_pubkey),
-                0,
+                graph.index.outpoints().iter().map(|(_, op)| *op),
```
nymius, line 490: I think the below tests assertions shouldn't be changed to match the new behavior, but the test be reworked to match the assertions. If the test was expecting some `untrusted_pending` balance, it should be changed to still produce that untrusted pending balance.

crates/chain/tests/test_canonical_view.rs
```diff
@@ -294,3 +311,486 @@ fn test_min_confirmations_multiple_transactions() {
...
+            utxos,
+            |c_tx| {
+                c_tx.tx
+                    .input
+                    .iter()
+                    .any(|txin| !owned.contains(&txin.previous_output))
+            },
+            |pos| pos.is_confirmed(),
+        )
+        .map(|(txout, eligibility)| (txout.outpoint, eligibility))
```
nymius, line 500: Does it make sense to have `impl From<Iter<TxOut, Elegibility>>> for Balance`? Is it possible?

crates/chain/tests/test_canonical_view.rs
```diff
@@ -294,3 +311,486 @@ fn test_min_confirmations_multiple_transactions() {
...
+#[test]
+fn test_classify_immature_and_settled() {
+    let blocks: BTreeMap<u32, BlockHash> = [(0, hash!("g")), (1, hash!("b1")), (2, hash!("tip"))]
+        .into_iter()
+        .collect();
+    let chain = LocalChain::from_blocks(blocks).unwrap();
+    let mut tx_graph = TxGraph::<ConfirmationBlockTime>::default();
+    let spk = ScriptBuf::new();
+
+    // Coinbase confirmed at height 1; far below `COINBASE_MATURITY` → immature.
```
nymius: Please, remove any non-ASCII characters from output: `→` (use `->`)

crates/chain/tests/test_canonical_view.rs
```diff
@@ -294,3 +311,486 @@ fn test_min_confirmations_multiple_transactions() {
...
+    // The foreign-spending ancestor is settled, so the walk stops there and never taints `child`.
+    assert_eq!(
+        by_op[&OutPoint::new(child_txid, 0)],
+        Eligibility::TrustedPending
+    );
+}
+
+/// `classify_outpoints` distinguishes an immature coinbase from a settled output.
+#[test]
+fn test_classify_immature_and_settled() {
```
nymius, line 634: After evan's review, we should have a test that contains a mature input at current height, and `is_settled` to aim for 3 confs, and check mature is still mature.

crates/chain/tests/test_canonical_view.rs
```diff
@@ -294,3 +311,486 @@ fn test_min_confirmations_multiple_transactions() {
...
+        },
+        |pos| pos.is_confirmed(),
+    );
+    assert_eq!(balance.trusted_pending, Amount::ZERO);
+    assert_eq!(balance.untrusted_pending, Amount::from_sat(29_000 + 19_000));
+}
+
+/// `classify_outpoints` skips outpoints that are already spent or absent from the view.
+#[test]
+fn test_classify_skips_spent_and_unknown() {
```
nymius, line 798: We should consider the other kind of `unknown` evan mentioned, parent txids, that aren't indexed, and are the parent root of an unconfirmed tx chain.

crates/chain/src/canonical.rs
evanlinjin: > You mean we consider txs for which its txid appear as parents of a tx in the CanonicalView as canonical?

Yes. However note that `CanonicalView` is a sparse view as `TxGraph` itself is also sparse. Therefore, the comment and logic here is wrong since we are visiting ancestors of canonical transactions, so anything we visit must also be canonical.

> Do you expect wallets to also handle this category as "unresolved" balance? Or it's just an intermediate category to trigger request for those txids and be able to resolve the whole trustness of the txs chains?

I think it's good to be transparent to the caller. Some callers may want to find those ancestors. Other callers may prefer to just treat unknown-trust transactions as untrusted.

> In my mind we will end up considering this as Untrusted because we cannot claim that we know if these are tainted or not.

I think you're probably correct. I also assume that for the majority of cases, the caller would just want to treat those transactions as untrusted. However, a transparent API is a more understandable API.

crates/chain/src/canonical.rs
evanlinjin: Hopefully there shouldn't be such a thing as an "unconfirmed coinbase" - the canonicalization algorithm should have considered those as non-canonical! If not, let's file a bug.

nymius: Not even if you assume it canonical?

Dmenec: Tried it with a coinbase without an anchor (which, as @evanlinjin said, shouldn't be possible on a real chain, but you can still construct it in a test), and if you assume it canonical it does end up in the set as unconfirmed.

https://github.com/bitcoindevkit/bdk/blob/88975799a34a147d6c3490e2413c635d3947520f/crates/chain/src/canonical.rs#L132-L137

As you can see above, it falls back to `false` there, so such an output is treated as immature anyway.

I agree with moving the maturity check out of `is_settled`, so it stays `Immature` regardless of settledness.

crates/chain/src/canonical.rs
Dmenec: Solved in efbb66983c11ab83d0a250a6ca4052237f1f3951

crates/chain/src/canonical.rs
Dmenec: Solved in efbb66983c11ab83d0a250a6ca4052237f1f3951

crates/chain/src/canonical.rs
Dmenec: Solved in efbb66983c11ab83d0a250a6ca4052237f1f3951

crates/chain/src/canonical.rs
Dmenec: Solved in efbb66983c11ab83d0a250a6ca4052237f1f3951

crates/chain/src/canonical.rs
Dmenec: Solved in efbb66983c11ab83d0a250a6ca4052237f1f3951

crates/chain/src/canonical.rs
Dmenec: Moved the maturity check out of `is_settled`. Leaving this unresolved for now.

crates/chain/src/canonical.rs
Dmenec: Implemented `Trust { Trusted, Untrusted, Unkown }` . 
Now if some output is missing from `CanonicalView` its classification falls to `Unknown`.
Marking this as resolved in efbb66983c11ab83d0a250a6ca4052237f1f3951

crates/chain/src/canonical.rs
Dmenec: Rewritten as a single lazy iterator chain, so it no longer builds a temporary list just to iterate over it right after. `classify_outpoints` now returns `impl Iterator + 'a`.

Solved in efbb66983c11ab83d0a250a6ca4052237f1f3951

crates/chain/tests/test_canonical_view.rs
Dmenec: Solved in ebd33196afab6ed03e8b28c587ea75b336158eb4

crates/chain/tests/test_tx_graph_conflicts.rs
Dmenec: Solved in ebd33196afab6ed03e8b28c587ea75b336158eb4

crates/chain/tests/test_indexed_tx_graph.rs
Dmenec: Done without changing assertions. 

ebd33196afab6ed03e8b28c587ea75b336158eb4

crates/chain/tests/test_canonical_view.rs
Dmenec: Added `test_mature_coinbase_is_settled_not_immature` in ebd33196afab6ed03e8b28c587ea75b336158eb4.

crates/chain/tests/test_canonical_view.rs
Dmenec: Added `test_unsettled_unknown_when_parent_root_missing` in ebd33196afab6ed03e8b28c587ea75b336158eb4,

crates/chain/tests/test_canonical_view.rs
Dmenec: Yes! Implemented as `impl FromIterator<(CanonicalTxOut<ChainPosition<A>>, Eligibility)> for Balance.
Thanks for the suggestion.

Solved in a2b2be80ed6a1dffcd32a3fd83a24df6bea5bbbe

Dmenec:
Thanks a lot @evanlinjin and @nymius for reviewing this :)
I think that covers everything now, but let me know if I missed something.

crates/chain/benches/trust_classification.rs
```diff
@@ -0,0 +1,485 @@
...
+) -> (
+    LocalChain,
+    TxGraph<ConfirmationBlockTime>,
+    Vec<Txid>,
+    ScriptBuf,
+) {
+    let chain = make_chain();
+    let mut tx_graph = TxGraph::default();
+    let owned = ScriptBuf::from(vec![1u8]);
+    let foreign = ScriptBuf::from(vec![2u8]); // not owned → untrusted
```
nymius, line 208: Remove non ASCII characters like this `→` from code. Use `->` instead.

crates/chain/tests/test_canonical_view.rs
```diff
@@ -55,40 +67,41 @@ fn test_min_confirmations_parameter() {
...
     assert_eq!(balance_1_conf.confirmed, Amount::from_sat(50_000));
     assert_eq!(balance_1_conf.trusted_pending, Amount::ZERO);
 
     // Test min_confirmations = 6: Should be confirmed (has exactly 6 confirmations)
     let balance_6_conf = canonical_view.balance(
-        [((), outpoint)],
-        |_, _| true, // trust all
-        6,
+        [outpoint],
+        |_tx| false, // never taint
```
nymius, line 85: nit: I think `// never taint (trust all)` is clearer. It's missing in the balance doctests too.

crates/chain/tests/test_canonical_view.rs
```diff
@@ -2,10 +2,22 @@
 
 use std::collections::BTreeMap;
 
-use bdk_chain::{local_chain::LocalChain, BlockId, ConfirmationBlockTime, TxGraph};
+use bdk_chain::{local_chain::LocalChain, BlockId, ChainPosition, ConfirmationBlockTime, TxGraph};
 use bdk_testenv::{hash, utils::new_tx};
 use bitcoin::{Amount, BlockHash, OutPoint, ScriptBuf, Transaction, TxIn, TxOut};
 
+/// Builds an `is_settled` predicate requiring at least `min_confirmations` confirmations.
+fn settled(
```
nymius, line 10: This is an idea for a follow up change: This could be renamed to `blocks_since_conf`and exposed publicly. The docs should say: `a tx with blocks_since_conf(tip, 5) is a transaction with 6 confs in next block, i.e., spendable in that same block.`

crates/chain/tests/test_indexed_tx_graph.rs
```diff
@@ -481,9 +487,22 @@ fn test_list_owned_txouts() {
...
+                        let op = txin.previous_output;
+                        !op.is_null()
+                            && canonical_view.txout(op).is_none_or(|txo| {
+                                graph
+                                    .index
+                                    .index_of_spk(txo.txout.script_pubkey.clone())
+                                    .is_none()
+                            })
+                    })
+                },
```
nymius: This resembles the code in `ancestry_trust`. Why we decided to make that function and `is_taint` consume `txid` instead of `txout`s? Have we tried it before?

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -365,8 +365,8 @@ fn test_tx_conflict_handling() {
             exp_unspents: HashSet::from([("C", 0)]),
             exp_balance: Balance {
                 immature: Amount::ZERO,
-                trusted_pending: Amount::from_sat(30000),
-                untrusted_pending: Amount::ZERO,
+                trusted_pending: Amount::ZERO,
```
nymius, line 368: Can not `exp_balance: Balance { untrusted_pending: Amount::from_sat(3000), ..Default::default() }` be used here?

crates/chain/tests/test_tx_graph_conflicts.rs
```diff
@@ -270,8 +270,8 @@ fn test_tx_conflict_handling() {
             exp_unspents: HashSet::from([("tx_conflict_1", 0)]),
             exp_balance: Balance {
                 immature: Amount::ZERO,
-                trusted_pending: Amount::from_sat(20000),
-                untrusted_pending: Amount::ZERO,
+                trusted_pending: Amount::ZERO,
```
nymius, line 273: Can not `exp_balance: Balance { untrusted_pending: Amount::from_sat(2000), ..Default::default() }` be used here?

crates/chain/tests/test_canonical_view.rs
```diff
@@ -310,6 +367,629 @@ fn test_min_confirmations_multiple_transactions() {
...
+        .map(|(txout, eligibility)| (txout.outpoint, eligibility))
+        .collect::<std::collections::HashMap<_, _>>();
+    assert_eq!(
+        by_op[&OutPoint::new(trusted_txid, 0)],
+        Eligibility::Unsettled(Trust::Trusted)
+    );
+    assert_eq!(
+        by_op[&OutPoint::new(chained_txid, 0)],
+        Eligibility::Unsettled(Trust::Untrusted)
+    );
```
nymius, line 509: This is redundant.

crates/chain/src/canonical.rs
```diff
@@ -394,98 +416,207 @@ impl<A, P: Clone> Canonical<A, P> {
...
+                    if cache.contains_key(&txid) || pending.contains(&txid) {
+                        continue;
+                    }
+                    let Some(c_tx) = self.tx(txid) else {
+                        // Missing from the `CanonicalView`.
+                        cache.insert(txid, Trust::Unknown);
+                        continue;
+                    };
+                    if is_settled(&c_tx.pos) {
+                        cache.insert(txid, Trust::Trusted);
```
110CodingP, line 513: Maybe I am missing something, but can't a (ancestor) coinbase tx be classified as trusted even if it is immature through our current logic?

nymius (commented):
In commit ff717896e7a211013953715fcb171bfbc3c28fe0, change `BREAKING` to `BREAKING CHANGE` to use the most common marker.

crates/chain/src/canonical.rs
Dmenec: It should (but only if this ancestor coinbase has an owned address)! The issue is that the balance shouldn't be selectable as an input since it is not mature enough to spend. Therefore, its `Eligibility` is left as `Immature` before trust is ever considered.

I think it is clearer to see it in `classify_outpoints` where this UTXO is classified as `Immature`.

crates/chain/benches/trust_classification.rs
Dmenec: Solved in 982244f4bd80821c994264e1dba5804f70e29d57

crates/chain/tests/test_canonical_view.rs
Dmenec: Solved in ab2a5c15bf9de32d0ad4d7e598cd61e03859b5db

crates/chain/tests/test_canonical_view.rs
Dmenec: Agree! When I get a moment I'll open a PR for it. I think it would be useful to have it as:

```rust
pub fn blocks_since_conf(&self, tip: u32) -> Option<u32>
```

and let the caller check the threshold `(blocks_since_conf(tip) >= n)`.

crates/chain/tests/test_tx_graph_conflicts.rs
Dmenec: Solved in 9f58d3c65345edb67be4994218af599cb3f17b9c

crates/chain/tests/test_tx_graph_conflicts.rs
Dmenec: Solved in 9f58d3c65345edb67be4994218af599cb3f17b9c

crates/chain/tests/test_canonical_view.rs
Dmenec: Solved in ab2a5c15bf9de32d0ad4d7e598cd61e03859b5db

crates/chain/tests/test_canonical_view.rs
Dmenec: Linking the PR here: https://github.com/bitcoindevkit/bdk/pull/2263

Dmenec:
I'd like to add a `locked` balance category on top of this (bdk_wallet's balance needs to be aware of it), for outputs that are confirmed but not yet spendable because a descriptor timelock hasn't matured yet, as raised in bitcoindevkit/bdk_wallet#180.

I think that maybe, in a future PR, a separate `frozen`/`reserved` category could also be added for outputs the user manually locks. I'd leave it as a possible future addition, which can be folded over `classify_outpoints` without changing `Balance` in wallet.

*Edit:* still thinking about how to do it. We could do the fold directly in the wallet with a `locked` category and leave `Balance` untouched, but it might be worth modeling it in chain too. Since each `Eligibility` variant has a matching `Balance` field, adding a `Locked` variant would mean adding `Balance.locked` as well... (same on frozen/reserved)  Some thoughts @evanlinjin @nymius @110CodingP?

crates/chain/src/canonical.rs
110CodingP: I meant to say that while iterating over elements of the stack we also add the coinbase tx to the stack. Now when we encounter it in the stack the first check we do is to check its settled (even before ensuring its not immature!).

Dmenec: Yup, in the walk `is_settled` is checked before maturity, so a confirmed but immature ancestor coinbase would be marked as `Trusted` without a maturity check.

Anyways, this can only occur if such a tx is inserted directly into the `TxGraph`, since it's invalid by consensus it would never become canonical when syncing.

I thought about three solutions here:

- The first is to just do nothing, since it shouldn't happen by consensus.
- The second is to return `Unknown` in the walk, so it folds into `untrusted_pending` with no special `Balance` handling.
- The third is to add a specific `Trust` variant for this, though it would end up similar to `Unknown`.

I'd prefer the first or second one

Dmenec: After discussing it with @110CodingP, I think the best option is the first one.

As this tx can only happen in specific scenarios where the user/their source are vouching for it, classifying it as Trusted is consistent, and there's no third-party replacement risk anyway. So I'll leave it `Trusted` and not add special handling.

crates/chain/tests/test_indexed_tx_graph.rs
110CodingP, line 368: nit: I think we should update the comments to remove the trusted/untrusted keychain since we can correctly establish trust based on ancestory now.

110CodingP: nit: We can similarly rename these.

110CodingP (approved):
LGTM 🚀 , just a nit.
I especially love the test coverage, great work @Dmenec !

crates/chain/src/canonical.rs
nymius: > Yup, in the walk `is_settled` is checked before maturity, so a confirmed but immature ancestor coinbase would be marked as `Trusted` without a maturity check.

It is? I mean, when is maturity checked in the walk (if you mean `ancestry_trust`)?

I think we remove the doubt here by a simple change, loading the `Immature`, and `Settled` txids from  the `classify_outpoints` main logic into the cache.

nymius:
> _Edit:_ still thinking about how to do it. We could do the fold directly in the wallet with a `locked` category and leave `Balance` untouched, but it might be worth modeling it in chain too. Since each `Eligibility` variant has a matching `Balance` field, adding a `Locked` variant would mean adding `Balance.locked` as well... (same on frozen/reserved)

This is part of the protocol, not user decided, so it belongs to chain. I would do both changes in `Elegibility` and `Balance`.
Using the same criteria, I'm not sure frozen belongs here, but Wallet. There you can pre-filter utxos to mark them as frozen before passing them to `chain::balance`. You could even create two separated list of outpoints, one set is frozen, the other is not, and compute balance for both separatedly. You would need some extra operations to compute a common output, but is doable.

crates/chain/src/canonical.rs
```diff
@@ -394,98 +416,207 @@ impl<A, P: Clone> Canonical<A, P> {
...
+                Frame::Exit(c_tx) => {
+                    let mut trust = Trust::Trusted;
+                    for txin in &c_tx.tx.input {
+                        if txin.previous_output.is_null() {
+                            continue;
+                        }
+                        let parent_trust = cache
+                            .get(&txin.previous_output.txid)
+                            .copied()
+                            .unwrap_or(Trust::Trusted);
```
noahjoeris: Is this `unwrap_or` reachable? Maybe it's better to use .expect()

crates/chain/src/canonical.rs
```diff
@@ -394,98 +416,207 @@ impl<A, P: Clone> Canonical<A, P> {
...
+        F: FnMut(&CanonicalTx<ChainPosition<A>>) -> bool,
+        S: Fn(&ChainPosition<A>) -> bool,
+    {
+        if let Some(&trust) = cache.get(&seed_txid) {
+            return trust;
+        }
+
+        // `Enter`: if tx is unsettled and not directly tainted, queue its parents.
+        // `Exit`: by now every parent is resolved, so the tx is tainted if any parent is.
+        enum Frame<A: Anchor> {
```
noahjoeris: I think the trait bound is not necessary

crates/chain/src/canonical.rs
Dmenec: I meant in `classify_outpoints`, sorry for the misunderstanding. In any case, `ancestry_trust` is only called from `classify_outpoints`, which already does the maturity check. Anyways, I think the main point here is that this specific case shouldn't happen with real data and if it ever did I think we should follow this:

> As this tx can only happen in specific scenarios where the user/their source are vouching for it, classifying it as Trusted is consistent, and there's no third-party replacement risk anyway. So I'll leave it `Trusted` and not add special handling.

crates/chain/tests/test_indexed_tx_graph.rs
Dmenec: Agreed! I have to change it, didn't notice it.

crates/chain/src/canonical.rs
Dmenec: Yes, I don't think this fallback is reachable. By the time we hit a transaction's `Exit`, all parent transactions should already be resolved and cached.

crates/chain/src/canonical.rs
Dmenec: Solved in 875ab92495b071c8bd8a55cfd8ee9637d220784b

crates/chain/src/canonical.rs
Dmenec: Solved in 875ab92495b071c8bd8a55cfd8ee9637d220784b

crates/chain/tests/test_indexed_tx_graph.rs
Dmenec: Solved in 875ab92495b071c8bd8a55cfd8ee9637d220784b

nymius (approved):
ACK 3bbdf7822292558d7dbe9720f7b87ad39ecdef29

noahjoeris (approved):
ACK 3bbdf7822292558d7dbe9720f7b87ad39ecdef29

Great job ✨

crates/chain/tests/test_canonical_view.rs
evanlinjin, line 132: We should remove this now since it's only testing a test helper method; `settled`.

crates/chain/tests/test_canonical_view.rs
```diff
@@ -2,10 +2,25 @@
...
+) -> impl Fn(&ChainPosition<ConfirmationBlockTime>) -> bool {
+    let min_confirmations = min_confirmations.max(1); // 0 and 1 behave identically
+    move |pos| {
+        pos.confirmation_height_upper_bound()
+            .is_some_and(|h| tip_height.saturating_sub(h).saturating_add(1) >= min_confirmations)
+    }
+}
+
 #[test]
 fn test_min_confirmations_parameter() {
```
evanlinjin, line 25: There is no `min_confirmations` parameter anymore. Maybe we should rename to `test_is_settled_boundary`?

nymius (approved):
ACK 2581fb2e5f5a3a864eadbed535132883cd3d7fea

crates/chain/src/canonical.rs
```diff
@@ -22,18 +22,40 @@
...
-};
+use bitcoin::{constants::COINBASE_MATURITY, OutPoint, ScriptBuf, Transaction, TxOut, Txid};
 
 use crate::{spk_txout::SpkTxOutIndex, Anchor, Balance, CanonicalViewTask, ChainPosition, TxGraph};
 
+/// The spend-eligibility classification of a canonical output, produced by
+/// [`CanonicalView::classify_outpoints`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Eligibility {
+    /// A confirmed output unlikely to be replaced.
```
evanlinjin, line 39:
```suggestion
    /// An output the caller considers settled, per the `is_settled` predicate given to
    /// [`classify_outpoints`](CanonicalView::classify_outpoints). Typically confirmed deeply
    /// enough to be unlikely to be replaced, but the caller decides.

```

`Settled` is not necessarily synonymous with "confirmed".

crates/chain/src/canonical.rs
```diff
@@ -394,98 +416,207 @@ impl<A, P: Clone> Canonical<A, P> {
...
+    ///             .is_some_and(|h| tip_height.saturating_sub(h).saturating_add(1) >= 6)
+    ///     },
     /// );
     /// ```
-    pub fn balance<'v, O: Clone + 'v>(
-        &'v self,
-        outpoints: impl IntoIterator<Item = (O, OutPoint)> + 'v,
-        mut trust_predicate: impl FnMut(&O, &CanonicalTxOut<ChainPosition<A>>) -> bool,
-        min_confirmations: u32,
+    pub fn balance(
```
evanlinjin, line 591: It's easy for a caller to expect that setting `does_taint` as `|_| false` would put all unconfirmed balance in `Balance::trusted_pending`. However this is not true as UTXOs categorized as `Unsettled(Unknown)` will end up in `Balance::untrusted_pending`. This is a potential footgun and even the tests in the repository got confused.

Essentially, `does_taint` can never achieve "trust all". We need to make this clear in the docs. Furthermore, I also propose introducing `Balance::unknown_pending` (which means trust cannot be determined due to missing ancestry) making the behavior even more explicit.

crates/chain/tests/test_canonical_view.rs
```diff
@@ -55,40 +91,41 @@ fn test_min_confirmations_parameter() {
...
     assert_eq!(balance_1_conf.confirmed, Amount::from_sat(50_000));
     assert_eq!(balance_1_conf.trusted_pending, Amount::ZERO);
 
     // Test min_confirmations = 6: Should be confirmed (has exactly 6 confirmations)
     let balance_6_conf = canonical_view.balance(
-        [((), outpoint)],
-        |_, _| true, // trust all
-        6,
+        [outpoint],
+        |_tx| false, // never taint (trust all)
```
evanlinjin, line 109: "trust all" is an incorrect framing. Maybe "leave trust to ancestry" makes more sense?

crates/chain/src/canonical.rs
```diff
@@ -394,98 +416,207 @@ impl<A, P: Clone> Canonical<A, P> {
...
     /// # let indexer = KeychainTxOutIndex::<&str>::default();
-    /// // Calculate balance with 6 confirmations, trusting all outputs
+    /// let tip_height = view.tip().height;
+    /// // Calculate balance requiring 6 confirmations, trusting every pending output.
     /// let balance = view.balance(
-    ///     indexer.outpoints().into_iter().map(|(k, op)| (k.clone(), *op)),
-    ///     |_keychain, _script| true,  // Trust all outputs
-    ///     6,  // Require 6 confirmations
+    ///     indexer.outpoints().iter().map(|(_, op)| *op),
+    ///     |_tx| false, // Never taint (trust all)
```
evanlinjin, line 584: Let's avoid doing `|_| false` in the examples and tests as it doesn't make sense.

Callers that actually want to "trust all" should do their own custom balance - our version is correct but opinionated.

crates/chain/tests/test_canonical_view.rs
```diff
@@ -143,12 +180,13 @@ fn test_min_confirmations_with_untrusted_tx() {
...
+    let tip_height = canonical_view.tip().height;
 
-    // Test with min_confirmations = 5 and untrusted predicate
+    // Test with min_confirmations = 5 and everything tainted
     let balance = canonical_view.balance(
-        [((), outpoint)],
-        |_, _| false, // don't trust
-        5,
+        [outpoint],
+        |_tx| true, // taint everything
```
evanlinjin, line 188: Right now, even if we switch this to `|_tx| false`, the test still passes for the same reasons we mentioned above; `hash!("parent")` does not actually point to a real parent -> `Trust::Unknown` -> `Balance::untrusted_pending`.

evanlinjin (changes requested):
This is quality work.

I've found a final rough edge that should be hashed out before we merge.

I haven't finished reviewing the tests yet. However, we should probably rename some of them as the `min_confirmations` parameter no longer exists.

crates/chain/src/canonical.rs
Dmenec: Agreed! I also like the idea of adding this new field. I'll map `Unsettled(Trust::Unknown)` to it.

Dmenec: Looking at the tests I noticed something... An incoming payment doesn't usually descend from one of our own transactions, so we rarely have its full ancestry, and most of what used to be `untrusted_pending` ends up in `unknown_pending` instead.

That leaves `untrusted_pending` nearly empty for a typical wallet, so I'm not sure a new `Balance` field is worth it for such a specific case.

crates/chain/tests/test_canonical_view.rs
Dmenec: Solved in 8f2db86ccc440ecb0bbbf333b191b6d7d9aecbf2

crates/chain/tests/test_canonical_view.rs
Dmenec: Solved in 8f2db86ccc440ecb0bbbf333b191b6d7d9aecbf2

crates/chain/src/canonical.rs
Dmenec: Solved in 98a7d5b88e975d529a4c8e6446ef8c3a4314281e

crates/chain/tests/test_canonical_view.rs
Dmenec: Solved in 8f2db86ccc440ecb0bbbf333b191b6d7d9aecbf2

crates/chain/src/canonical.rs
Dmenec: Solved in 98a7d5b88e975d529a4c8e6446ef8c3a4314281e

crates/chain/tests/test_canonical_view.rs
Dmenec: Solved in 8f2db86ccc440ecb0bbbf333b191b6d7d9aecbf2

crates/chain/src/canonical.rs
evanlinjin: That's a good point... 🤔

evanlinjin: @Dmenec an incoming payment is not owned by us and the fact that it's new, it's probably unsettled - therefore making it "untrusted".

The `unknown_pending` case should basically be unreachable unless we do something like `does_taint = |_| false` in which case we will have missing history as the algorithm keeps reaching further to try prove something is not tainted.

evanlinjin: We really need to disincentivize callers from writing their own `does_taint` (especially `|_| false`). We can use positive enforcement by adding a helper function:

```rust
/// A `does_taint` predicate: a transaction taints when any input spends a coin this indexer
/// doesn't recognise as ours — including one whose parent we've never seen.
pub fn taints_unowned<P, I>(
    indexer: &impl AsRef<SpkTxOutIndex<I>>,
) -> impl Fn(&CanonicalTx<P>) -> bool + '_ {
    let indexer = indexer.as_ref();
    move |c_tx| {
        c_tx.tx.input.iter().any(|txin| {
            !txin.previous_output.is_null() && indexer.txout(txin.previous_output).is_none()
        })
    }
}
```

evanlinjin: @Dmenec I'm happy to ACK once `taints_unowned` is introduced.

* Tests can be clarified in follow-ups.
* `Balance::unknown_pending` is mostly unreachable right now - we can include it later if it's actually needed. The caller may have an indexer that also tracks unowned outputs - so outputs will either be owned, unowned and unknown. Unknown means "we haven't synced this output yet". Maybe this is useful, maybe it is not, but requires changing signature of `does_taint` to return an `Option`. Another possibility is if the caller with such an indexer decided to treat "unknown" as "does not taint" - then we will be hitting "unknown" more often. YAGNI for now - let's create a ticket.

Dmenec: You're right about everything, I've got myself confused with the test predicates... Already working on it :)

evanlinjin (approved):
ACK e2dad373b0e2023a58c67afb073406eeaf028374

Solid work.

nymius (approved):
ACK e2dad373b0e2023a58c67afb073406eeaf028374

Issue #2251, opened 2026-08-01
KeychainTxOutIndex::lookahead_to_target panics (overflow) or silently no-ops on target_index = u32::MAX


CapThunder19:
**Describe the bug**  
`KeychainTxOutIndex::lookahead_to_target` computes `target_index + 1` without overflow protection (crates/chain/src/indexer/keychain_txout.rs, line 543):

    let temp_lookahead = (target_index + 1)
        .checked_sub(next_index)
        .filter(|&index| index > 0);

When `target_index == u32::MAX`:
- In a debug build (default `cargo test`/`cargo build`, overflow-checks on): panics with "attempt to add with overflow".
- In a release build (overflow-checks off by default): `u32::MAX + 1` silently wraps to 0, so `checked_sub`/`filter` evaluate to `None` and the function silently does nothing — no error, no scripts derived, no indication to the caller that the call had no effect.

`target_index` is a raw public API parameter with no upstream validation, so it could plausibly come from untrusted input (e.g. a derivation index parsed from an external PSBT).

For comparison, the sibling method `reveal_to_target` handles the same boundary safely — it only ever compares indices (`i > target_index`) rather than doing arithmetic on the raw value. `lookahead_to_target` has no equivalent protection.

**To Reproduce**  
Add this test to crates/chain/tests/test_keychain_txout_index.rs:

    #[test]
    fn repro_lookahead_to_target_overflow() {
        let external_descriptor = parse_descriptor(DESCRIPTORS[0]);
        let internal_descriptor = parse_descriptor(DESCRIPTORS[1]);
        let mut index = init_txout_index(external_descriptor, internal_descriptor, 10, true);
        let _ = index.lookahead_to_target(TestKeychain::External, u32::MAX);
    }

Run:

    cargo test --features miniscript --test test_keychain_txout_index repro_lookahead_to_target_overflow -- --nocapture

Actual output:

    thread 'repro_lookahead_to_target_overflow' panicked at crates/chain/src/indexer/keychain_txout.rs:543:34:
    attempt to add with overflow

**Expected behavior**  
No panic in debug builds, and no silent no-op in release builds. The lookahead should either be extended as far as representable (e.g. via `target_index.saturating_add(1)` instead of `target_index + 1`), or the call should visibly report that the target was unreachable — not fail silently.

**Proposed fix**
[...]

PR #2253, opened 2026-08-04
fix(chain): prevent overflow in `KeychainTxOutIndex::lookahead_to_target`

CapThunder19:
## Description

`lookahead_to_target` computed `target_index + 1` with unchecked
addition. With `target_index == u32::MAX`, this panics in debug builds
and silently no-ops in release builds (wraps to 0, so the lookahead
request is quietly ignored).

Fixed by extracting the delta calculation into a `lookahead_delta`
helper using `saturating_add` instead of `+`, matching the pattern
already used elsewhere in this file.

## Testing

Added `lookahead_delta_does_not_overflow_at_u32_max`, a unit test on the
helper directly. (Not an integration test calling `lookahead_to_target`
end-to-end with `u32::MAX`, since that would try to derive ~2^31 real
keys on a fresh index.)

- [x] `cargo test --features miniscript --lib`
- [x] `cargo test --features miniscript --test test_keychain_txout_index`
- [x] `cargo fmt --check -p bdk_chain`
- [x] `cargo clippy -p bdk_chain --features miniscript --lib -- -D warnings`

#2251

### DIFF
--- a/crates/chain/src/indexer/keychain_txout.rs
+++ b/crates/chain/src/indexer/keychain_txout.rs
@@ -540,16 +540,23 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     pub fn lookahead_to_target(&mut self, keychain: K, target_index: u32) -> ChangeSet {
         let mut changeset = ChangeSet::default();
         if let Some((next_index, _)) = self.next_index(keychain.clone()) {
-            let temp_lookahead = (target_index + 1)
-                .checked_sub(next_index)
-                .filter(|&index| index > 0);
-
-            if let Some(temp_lookahead) = temp_lookahead {
+            if let Some(temp_lookahead) = Self::lookahead_delta(target_index, next_index) {
                 self.replenish_inner_index_keychain(keychain, temp_lookahead);
             }
         }
         self._empty_stage_into_changeset(&mut changeset);
         changeset
+    }
+
+    /// Computes how many additional lookahead scripts are needed to cover `target_index`
+    /// (inclusive), given the next index that would be derived (`next_index`).
+    ///
+    /// Returns `None` if `target_index` is already covered (i.e. `target_index < next_index`).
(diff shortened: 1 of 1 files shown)


crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -540,18 +540,25 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+    /// (inclusive), given the next index that would be derived (`next_index`).
+    ///
+    /// Returns `None` if `target_index` is already covered (i.e. `target_index < next_index`).
+    fn lookahead_delta(target_index: u32, next_index: u32) -> Option<u32> {
+        target_index
+            .saturating_add(1)
+            .checked_sub(next_index)
+            .filter(|&index| index > 0)
+    }
+
```
evanlinjin, line 561: Please inline this. A separate method here is not helpful.

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -1146,6 +1153,24 @@ mod test {
...
+        );
+        // Target already covered by next_index: no lookahead needed.
+        assert_eq!(
+            KeychainTxOutIndex::<i32>::lookahead_delta(u32::MAX, u32::MAX),
+            None
+        );
+        // Normal, non-boundary case still behaves as before.
+        assert_eq!(KeychainTxOutIndex::<i32>::lookahead_delta(10, 5), Some(6));
+    }
+
```
evanlinjin, line 1173: This unit test does not add any value. Please remove.

evanlinjin (changes requested):
Thanks for the PR. Concept ACK.

crates/chain/src/indexer/keychain_txout.rs
Amanyadav0556: looks good

CapThunder19:
pls review it

evanlinjin:
#2265 supersedes this.

PR #2257, opened 2026-08-11
fix: fix command order for importing GPG key

Psycarlo:
### Fix

Options order wrong. `--recv-keys` eat everything after as key IDs.
Put `--keyserver` first.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -14,7 +14,7 @@ The following key may be used to communicate sensitive information to BDK via em
 
 You can import the key by running the following command:
 ```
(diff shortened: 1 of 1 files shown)


thunderbiscuit (approved):
ACK 3474e0c6d02e527c1f8a38660191e243026dac0d. Tested and our current line is indeed incorrect; proposed fix works.

luisschwab (approved):
ACK 3474e0c6d02e527c1f8a38660191e243026dac0d

evanlinjin (approved):
ACK 3474e0c6d02e527c1f8a38660191e243026dac0d

PR #2258, opened 2026-08-13
Replace bincode by postcard

nymius:
`bincode` is no longer maintained.
`postcard` is the closest maintained project with >52M downloads on crates.io, regular releases, and activity on its repo.

BREAKING CHANGE:
- Magic Bytes should be changed to avoid accidentaly modifying old file store blobs.
- The internal encoding has changed as a result of using postcard. Old file stores won't be recoverable using the latest file_store version.
- As this is a development environment store, we don't provide migration utilities.
- `StoreError::Bincode` has been renamed to `StoreError::Decode`, and now contains `postcard::Error`s
- From now on, trailing bytes after decoding are rejected.
- `append` always tries to attach changesets to the latest valid end of the file.

**These changes were LLM assisted.**

### Changed

- Replaced `bincode` (unmaintained) with `postcard` for on-disk (de)serialization.
  - **BREAKING CHANGE**: the on-disk format has changed. Existing store files are not readable by this
    version; callers should bump their magic bytes so old files fail fast with
    `StoreError::InvalidMagicBytes` instead of a decode error.
  - **BREAKING CHANGE**: `StoreError::Bincode(bincode::ErrorKind)` is replaced by
    `StoreError::Decode(postcard::Error)`.
- **BREAKING CHANGE**: Entry length prefix changed from an 8-byte little-endian integer to a
  `postcard` varint (1-10 bytes). Covered by the same on-disk format change and
   magic-byte-bump guidance above.

### Fixed

- `Store::append` now seeks to the end of the file before writing, so appending through a
  stale handle no longer overwrites changesets written via another handle. A failed append
  also truncates the partial frame, leaving the file unchanged.
- Entries whose payload doesn't consume its whole declared frame length are now rejected instead of ignored.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/file_store/src/entry_iter.rs
+++ b/crates/file_store/src/entry_iter.rs
@@ -1,17 +1,17 @@
 use crate::StoreError;
-use bincode::Options;
 use std::{
     fs::File,
-    io::{self, BufReader, Seek},
+    io::{self, BufRead, BufReader, Read, Seek},
     marker::PhantomData,
 };
-
-use crate::bincode_options;
 
 /// Iterator over entries in a file store.
 ///
 /// Reads and returns an entry each time [`next`] is called. If an error occurs while reading the
 /// iterator will yield a `Result::Err(_)` instead and then `None` for the next call to `next`.
+///
+/// Each entry is stored as a `postcard`-encoded `u64` varint length prefix followed by that many
+/// bytes of `postcard`-encoded data.
 ///
 /// [`next`]: Self::next
 pub struct EntryIter<'t, T> {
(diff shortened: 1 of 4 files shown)


chukwudiikeh:
Breaking compatibility with older bincode-encoded stores without supplying a conversion tool leaves users stranded. Regardless of whether file_store is intended for dev environments, dropping support without providing a migration path or prior notice inevitably leads to silent data loss.

evanlinjin:
@chukwudiikeh There is no data loss. Everything to persist in `bdk_chain` is blockchain data, which is always available in the blockchain.

crates/file_store/src/entry_iter.rs
```diff
@@ -44,31 +44,132 @@ where
...
+        if self.db_file.fill_buf()?.is_empty() {
+            return Ok(None);
+        }
+
+        let entry = self.read_frame();
+        if entry.is_err() {
+            // Leave the file at the start of the failed entry.
+            self.db_file.seek(io::SeekFrom::Start(pos_before_read))?;
+        }
+        entry.map(Some)
```
evanlinjin: I would inline `read_frame` and `read_len_prefix` into `read_entry`.

Rationale:

1. They have only one call site (`read_frame`).
2. It is a potential footgun if we call them on their own as we want to seek to the start of the entry on error.

crates/file_store/src/entry_iter.rs
```diff
@@ -44,31 +44,132 @@ where
...
+
+    /// Reads `len` payload bytes into a fresh buffer.
+    fn read_payload(&mut self, len: u64, payload_start: u64) -> Result<Vec<u8>, StoreError> {
+        let mut payload = Vec::new();
+        // Reserve exactly `len` bytes up front. Fail fast on a corrupt, oversized length prefix.
+        // Avoids unnecessary reads and allocations.
+        let alloc_failed = usize::try_from(len)
+            .map_err(|_| ())
+            .and_then(|len| payload.try_reserve_exact(len).map_err(|_| ()))
+            .is_err();
```
evanlinjin, line 129: Is there no better way to write this? 😅

crates/file_store/src/store.rs
```diff
@@ -226,28 +225,48 @@ where
...
-            .map_err(|e| match *e {
-                bincode::ErrorKind::Io(error) => error,
-                unexpected_err => panic!("unexpected bincode error: {unexpected_err}"),
-            })?;
+        let bytes = postcard::to_allocvec(changeset).map_err(io::Error::other)?;
+        let len_bytes = postcard::to_allocvec(&(bytes.len() as u64)).map_err(io::Error::other)?;
+
+        // Always write at the current end of the file. This handle's cursor may be stale if
+        // another handle has appended since we last read, and writing at a stale offset would
+        // overwrite those changesets.
```
evanlinjin, line 258: I don't think this is the behavior we want. If we failed to read the last entry due to `DeserializeUnexpectedRead`, then further writes will also be corrupted and unreadable due to the `len_prefix`es being unaligned.

evanlinjin (changes requested):
I haven't reviewed the tests yet.

`EntryIter` changes are well written and very thorough - handling all situations elegantly and returning useful errors.

I disagree with the `Store::append` changes as a corrupted tail becomes unrecoverable.

crates/file_store/src/entry_iter.rs
evanlinjin: Could we use `postcard::from_io` instead?

evanlinjin:
> Entry length prefix changed from an 8-byte little-endian integer to a
postcard varint (1-10 bytes). Covered by the same on-disk format change and
magic-byte-bump guidance above.

The version on `master` used varint length prefixes as well.

luisschwab:
> Breaking compatibility with older bincode-encoded stores without supplying a conversion tool leaves users stranded. Regardless of whether file_store is intended for dev environments, dropping support without providing a migration path or prior notice inevitably leads to silent data loss.

This crate is explicitly for testing only, per the [README](https://github.com/bitcoindevkit/bdk/tree/master/crates/file_store#bdk-file-store).

crates/file_store/src/entry_iter.rs
nymius: A better way to write it:
```rust
        let alloc_failed = match usize::try_from(len) {
            Ok(len) => payload.try_reserve_exact(len).is_err(),
            Err(_) => true,
        };
```

`postcard::from_io` would require a buffer allocation that must account for the largest possible decodable value. That's impossible to know without looking at the first varint. We could use a large enough scratch buffer, but I prefer the other way that just creates the vector, without the size considerations.
Also, `from_io`, maps all `io::Errors` to `DeserializeUnexpectedEnd`, making the distinction between torn data of allocation failure impossible, and I would like to keep it.

nymius:
> > Entry length prefix changed from an 8-byte little-endian integer to a
> > postcard varint (1-10 bytes). Covered by the same on-disk format change and
> > magic-byte-bump guidance above.
> 
> The version on `master` used varint length prefixes as well.

Yes, this is a left over from an early iteration that I never published.

crates/file_store/src/store.rs
nymius: I didn't have that present while working on this, but what I have in mind was to not leave space for intermediate faulty states, except for concurrency (that as far as I know is not supported nor intended to do so), and bit flipping (which should be remediated by redundancy, i.e., backups).
There is a flaw in the current implementation though, because the file end setting can fail too, leaving the torn bytes behind. What if I fail loudly in that case? So the user knows the append failed and there are torn bytes at the end, and he should retry the append of fix his changesets and retry later.
In this case all intermediate faulty states (without considering concurrency and bit flipping) will be covered.

crates/file_store/src/store.rs
evanlinjin: `dump` being public is problematic - on error, `self` is still available for the caller to append more changesets.

crates/file_store/src/store.rs
```diff
@@ -599,4 +621,216 @@ mod test {
...
+            .expect("both appends must remain decodable");
+        let mut expected = initial;
+        expected.extend(first_update);
+        expected.extend(second_update);
+        assert_eq!(
+            recovered,
+            Some(expected),
+            "a stale handle overwrote an append"
+        );
+    }
```
evanlinjin, line 799: Can we just assume that there is always only one handle at one time per file? As you mentioned, we don't need to support concurrent access. So the only time this would occur is either a bug in the caller's code, or the user of the wallet tries to open the same instance of the wallet twice. I think this should be handled with a lock file (separate PR imo).

crates/file_store/src/store.rs
evanlinjin: I think we need to assume 1 handle per file and use a lockfile to enforce it. I propose reverting to the old logic in `master`; `append` should only append after valid data.

One thing that is missing in `master`, is actually rolling back the data (as you have done with `File::set_len`).

nymius: I was trying to not get into locks, but as it seems we are headed in that direction, I can roll back the stale head change, and document the expectations in this PR, and leave the lock changes for another PR.

crates/file_store/src/store.rs
nymius: Yes, as I said above, I was trying to not get into locks. I will document the expectations, and leave this out of this PR.

PR #2261, opened 2026-08-18
docs(esplora): replace misleading tx-cache TODO with explanation

### LINKED ISSUE #2260: esplora: remove misleading `TODO` about maintaining a tx cache
`fetch_txs_with_outpoints` carries this comment in both backends:

- `crates/esplora/src/async_ext.rs`
- `crates/esplora/src/blocking_ext.rs`

```rust
// TODO: We should maintain a tx cache (like we do with Electrum).
```

It should be removed. It invites a port of `BdkElectrumClient`'s cache that cannot pay off here, and it already cost a contributor a 600-line PR (#2254, #2250).

**Why the Electrum pattern doesn't transfer**

Electrum's cache exists because `blockchain.scripthash.get_history` returns *txids only*, forcing a follow-up `blockchain.transaction.get` per transaction. The cache eliminates that second call.

Esplora's `GET /scripthash/:hash/txs` returns full transactions — `esplora_client::Tx` carries `vin`/`vout` and `to_tx()` reconstructs the `Transaction` from the response body. There is no second call, so there is nothing for a tx cache to skip. A cache could only ever apply to the `iter_txids`/`iter_outpoints` paths, and those share `inserted_txs` with the spk phase that runs first — so in a `SyncRequest` carrying spks they are almost always already covered.

Substituting `get_tx_status` for `get_tx_info` is additionally unsafe: Esplora returns HTTP 200 `{"confirmed":false}` for an unknown txid, indistinguishable from a genuinely unconfirmed tx, so evicted transactions would have `last_seen` bumped on every sync and never be evicted.

**Suggested replacement**

Delete the TODO, or replace it with a note recording why a tx cache does *not* apply to Esplora, so this doesn't get rediscovered a third time.

If someone wants to optimize this function, the real target is the per-outpoint `get_output_status` round trips — those are not covered by the spk scan.

Arowolokehinde:
### Description

`fetch_txs_with_outpoints` carried a misleading `// TODO: We should maintain a tx cache (like we do with Electrum).` comment in both `async_ext.rs` and `blocking_ext.rs`. Electrum needs a cache because its history call returns txids only, forcing a follow-up fetch per tx. Esplora's scan endpoint already returns full tx bodies, so there's no equivalent fetch for a cache to skip — and `inserted_txs` already dedupes against what the spk phase fetched earlier in the same sync.

This PR replaces the TODO with a comment explaining why, and why the obvious "fix" — using `get_tx_status` instead of `get_tx_info` for cache hits — is unsafe: unknown txids return HTTP 200 `{"confirmed":false}`, indistinguishable from a real unconfirmed tx, so evicted transactions would never be evicted.

Fixes #2260

### Notes to the reviewers

The TODO already cost a contributor a 600-line PR (#2254, closing #2250) implementing exactly this cache, which was caught in review for the `get_tx_status` eviction bug described above. This comment exists to stop that from being rediscovered a third time.

### Changelog notice

Removed a misleading TODO in `fetch_txs_with_outpoints` suggesting a tx cache be added; replaced with a comment explaining why the Electrum-style cache pattern does not apply to Esplora.

#### All Submissions:

### DIFF
--- a/crates/esplora/src/async_ext.rs
+++ b/crates/esplora/src/async_ext.rs
@@ -492,7 +492,8 @@ where
     let mut update = TxUpdate::<ConfirmationBlockTime>::default();
 
     // make sure txs exists in graph and tx statuses are updated
-    // TODO: We should maintain a tx cache (like we do with Electrum).
+    // TODO: The per-outpoint get_output_status round trips can be optimized.
+    // A tx cache does not apply in this scenario.
     update.extend(
         fetch_txs_with_txids(
             client,
--- a/crates/esplora/src/blocking_ext.rs
+++ b/crates/esplora/src/blocking_ext.rs
@@ -451,7 +451,8 @@ fn fetch_txs_with_outpoints<I: IntoIterator<Item = OutPoint>>(
     let mut update = TxUpdate::<ConfirmationBlockTime>::default();
 
     // make sure txs exists in graph and tx statuses are updated
-    // TODO: We should maintain a tx cache (like we do with Electrum).
+    // TODO: The per-outpoint get_output_status round trips can be optimized.
+    // A tx cache does not apply in this scenario.
     update.extend(fetch_txs_with_txids(
         client,
         start_time,


luisschwab:
I'd just add a short comment stating that a tx cache does not apply in this scenario, and change the TODO to say that the per-outpoint `get_output_status` round trips can be optimized.

Arowolokehinde:
Changes has been made as required to the comment @luisschwab

luisschwab:
The commits should be squashed

Arowolokehinde:
i just squashed the commit
Thanks for the review

Arowolokehinde:
@luisschwab kindly review

evanlinjin:
> I'd just add a short comment stating that a tx cache does not apply in this scenario, and change the TODO to say that the per-outpoint `get_output_status` round trips can be optimized.

How do you propose that we optimize this?

I prefer if we create a ticket instead of having a TODO comment in the codebase. Tickets are easier to track.

I'm responsible for writing the initial TODO comment and I apologize for creating this precedent.

PR #2263, opened 2026-08-20
feat(chain): add ChainPosition::confirmations_lower_bound

Dmenec:
Suggested by @nymius in https://github.com/bitcoindevkit/bdk/pull/2246#discussion_r3796478212.

### Description

Adds `ChainPosition::confirmations_lower_bound(tip)`, which returns the number of blocks mined on top of a confirmed position's block, given the chain `tip`. Returns `None` if the position is unconfirmed, or if the confirmation height is above `tip`.

### Changelog notice

- Added `ChainPosition::confirmations_lower_bound`, returning the number of blocks mined on top of a confirmed position's block given the chain tip.

#### All Submissions:

* [x] I ran `just p` before committing

#### New Features:

### DIFF
--- a/crates/chain/src/chain_data.rs
+++ b/crates/chain/src/chain_data.rs
@@ -83,6 +83,21 @@ impl<A: Anchor> ChainPosition<A> {
             ChainPosition::Unconfirmed { .. } => None,
         }
     }
+
+    /// Number of confirmations of this position, given `tip`.
+    ///
+    /// Returns `0` if unconfirmed, or if the confirmation height is above `tip`.
+    pub fn confirmations_lower_bound(&self, tip: u32) -> u32 {
+        let Some(height) = self.confirmation_height_upper_bound() else {
+            return 0;
+        };
+
+        if height > tip {
+            return 0;
+        }
+
+        tip - height + 1
+    }
 }
 
 /// Ordering for `ChainPosition`:
@@ -334,5 +349,30 @@ mod test {
                 },
             ]
         );
+    }
+
+    #[test]
+    fn test_confirmations_lower_bound() {
+        let confirmed_at = |height: u32| ChainPosition::Confirmed {
+            anchor: ConfirmationBlockTime {
+                confirmation_time: 0,
+                block_id: BlockId {
+                    height,
+                    ..Default::default()
+                },
+            },
+            transitively: None,
+        };
+
+        assert_eq!(confirmed_at(100).confirmations_lower_bound(100), 1);
+        assert_eq!(confirmed_at(99).confirmations_lower_bound(100), 2);
+        assert_eq!(confirmed_at(90).confirmations_lower_bound(100), 11);
+        assert_eq!(confirmed_at(101).confirmations_lower_bound(100), 0);
+
+        let unconfirmed = ChainPosition::<ConfirmationBlockTime>::Unconfirmed {
+            first_seen: Some(1),
+            last_seen: Some(2),
+        };
+        assert_eq!(unconfirmed.confirmations_lower_bound(100), 0);
     }
 }


evanlinjin:
The concept that is more common is "confirmations". Would it be better to just have `confirmations() -> Option<u32>` and for callers that need "blocks since conf", it will be `confirmations.and_then(|n| n.checked_sub(1))`?

Edit: Note that it would be `confirmations_lower_bound`.

nymius:
> The concept that is more common is "confirmations". Would it be better to just have `confirmations() -> Option<u32>` and for callers that need "blocks since conf", it will be `confirmations.and_then(|n| n.checked_sub(1))`?
> 
> Edit: Note that it would be `confirmations_lower_bound`.

Agree, it is more familiar. Something I should have asked before suggesting this is, is it interesting from an API user perspective? Because I came up with this in the context of a test, but seems to be useful in other contexts.

evanlinjin:
@nymius I think getting confirmations from chain position is definitely useful. It's what a wallet would want to show to the user.

crates/chain/src/chain_data.rs
```diff
@@ -83,6 +83,19 @@ impl<A: Anchor> ChainPosition<A> {
...
+    ///
+    /// Returns `None` if unconfirmed, or if the confirmation height is above `tip`.
+    pub fn confirmations_lower_bound(&self, tip: u32) -> Option<u32> {
+        let height = self.confirmation_height_upper_bound()?;
+
+        if height > tip {
+            return None;
+        }
+
+        Some(tip - height + 1)
```
nymius: What happens when `tip` << `height`?

Dmenec:
```
if height > tip {
  return None;
}
```
I don't think there's any case where `tip - height + 1` can be reached when `height > tip`.

nymius: I would love to say I'm reading this through the one line screen of a beeper, but I'm not. Shortsightedness at its prime. You're right.

evanlinjin:
@nymius @Dmenec what do you think about `confirmations_lower_bound` returning `u32` where `0` means unconfirmed? Right now it is not clear what's different between `None` and `Some(0)`.

nymius:
> @nymius @Dmenec what do you think about `confirmations_lower_bound` returning `u32` where `0` means unconfirmed? Right now it is not clear what's different between `None` and `Some(0)`.

Sounds better, 0 describes clearly no confirmations.

Dmenec:
> @nymius @Dmenec what do you think about `confirmations_lower_bound` returning `u32` where `0` means unconfirmed? Right now it is not clear what's different between `None` and `Some(0)`.

Not totally sure about this. with `Option` the caller is forced to handle the unconfirmed case explicitly, if we return a `u32` something like confirmations -1 would underflow when unconfirmed. 

In case we decide to change this to`u32` maybe `confirmation_height_upper_bound` should work the same way?

Edit: Don't think `confirmation_height_upper_bound` can be `u32`, since `None` is returned for unconfirmed and `0` for confirmed at genesis block.

nymius:
> > @nymius @Dmenec what do you think about `confirmations_lower_bound` returning `u32` where `0` means unconfirmed? Right now it is not clear what's different between `None` and `Some(0)`.
> 
> Not totally sure about this. with `Option` the caller is forced to handle the unconfirmed case explicitly, if we return a `u32` something like confirmations -1 would underflow when unconfirmed.
> 
> In case we decide to change this to`u32` maybe `confirmation_height_upper_bound` should work the same way?
> 
> Edit: Don't think `confirmation_height_upper_bound` can be `u32`, since `None` is returned for unconfirmed and `0` for confirmed at genesis block.

How `-1` would be a real number of confirmations? It could be an estimate, but you would never know.

Dmenec:
> > > @nymius @Dmenec what do you think about `confirmations_lower_bound` returning `u32` where `0` means unconfirmed? Right now it is not clear what's different between `None` and `Some(0)`.
> > 
> > 
> > Not totally sure about this. with `Option` the caller is forced to handle the unconfirmed case explicitly, if we return a `u32` something like confirmations -1 would underflow when unconfirmed.
> > In case we decide to change this to`u32` maybe `confirmation_height_upper_bound` should work the same way?
> > Edit: Don't think `confirmation_height_upper_bound` can be `u32`, since `None` is returned for unconfirmed and `0` for confirmed at genesis block.
> 
> How `-1` would be a real number of confirmations? It could be an estimate, but you would never know.

Sorry, I think I was a bit ambiguous there.. I meant `confirmations - 1` as the conversion from confirmations to blocks since confirmation, not as a confirmation count itself.

What I had in mind with `-> u32` is a caller doing:

```rust
let blocks_since_conf = pos.confirmations_lower_bound(tip) - 1;
```

For an unconfirmed position that's `0 - 1`, which panics in debug and wraps to `u32::MAX` in release, so the check silently passes. With `Option` the compiler forces you to handle that case.

nymius:
> > > > @nymius @Dmenec what do you think about `confirmations_lower_bound` returning `u32` where `0` means unconfirmed? Right now it is not clear what's different between `None` and `Some(0)`.
> > > 
> > > 
> > > Not totally sure about this. with `Option` the caller is forced to handle the unconfirmed case explicitly, if we return a `u32` something like confirmations -1 would underflow when unconfirmed.
> > > In case we decide to change this to`u32` maybe `confirmation_height_upper_bound` should work the same way?
> > > Edit: Don't think `confirmation_height_upper_bound` can be `u32`, since `None` is returned for unconfirmed and `0` for confirmed at genesis block.
> > 
> > 
> > How `-1` would be a real number of confirmations? It could be an estimate, but you would never know.
> 
> Sorry, I think I was a bit ambiguous there.. I meant `confirmations - 1` as the conversion from confirmations to blocks since confirmation, not as a confirmation count itself.
> 
> What I had in mind with `-> u32` is a caller doing:
> 
> ```rust
> let blocks_since_conf = pos.confirmations_lower_bound(tip) - 1;
> ```
> 
> For an unconfirmed position that's `0 - 1`, which panics in debug and wraps to `u32::MAX` in release, so the check silently passes. With `Option` the compiler forces you to handle that case.

Can you write the complete function? I'm not getting what you mean

Dmenec:
> Can you write the complete function? I'm not getting what you mean

Sure!

```rust
// unconfirmed tx and the tip is 100, checking if tx is >= 5 blocks since confirmation

// Option: no confirmations, so there's nothing to subtract from, and it's not spendable
pos.confirmations_lower_bound(100) 
    .and_then(|c| c.checked_sub(1))
    .is_some_and(|b| b >= 5)  // false

// u32: 0 confirmations minus 1 underflows to u32::MAX, so it looks deeply confirmed
pos.confirmations_lower_bound(100) - 1  // 0 - 1 -> u32::MAX
    >= 5  // true
```

evanlinjin:
@Dmenec 

```rust
pos.confirmations_lower_bound(100).checked_sub(1).is_some_and(|n| n >= 5); // false
```

Dmenec:
Yes! I was meaning a case where the caller does not call it in the "right" way and tries to get the blocks since confirmation. But of course that's how they should be calling it in this case. I'm not sure what's preferable in those cases, I was just thinking that if we could avoid the caller doing that call it would be great.

evanlinjin:
@Dmenec Personally I think a doccomment explaining that 0 means unconfirmed is enough.

evanlinjin (approved):
ACK 7ebc9d3ca216eafb11fd1d976955f9e3500108c6

PR #2265, opened 2026-08-25
fix(chain): sanitize and bound-check derivation indices near BIP32_MAX_INDEX

### LINKED ISSUE #60: 'apply_changeset' the derivation index aren't sanitized

busayo-OD:
Fixes [bdk_wallet/issues/60](https://github.com/bitcoindevkit/bdk_wallet/issues/60)

### Description

`KeychainTxOutIndex::apply_changeset` accepted `ChangeSet::last_revealed` values with no upper bound, allowing an index greater than `BIP32_MAX_INDEX` (2^31 - 1) to be stored, violating an invariant relied on by `next_index()`.

The changes in this PR address boundary issues around derivation indices:

1. **`apply_changeset`**: clamps `last_revealed` to `BIP32_MAX_INDEX` before storing it, instead of storing the raw value.
2. **`replenish_inner_index`**: fixes an off-by-one in the `stop_index` calculation that caused a panic (`"we just inserted it"`) when a keychain's derivation index was revealed exactly up to `BIP32_MAX_INDEX`. The exclusive upper bound was clamped to `BIP32_MAX_INDEX` instead of `BIP32_MAX_INDEX + 1`, silently excluding the boundary value itself.
3. **`lookahead_to_target`**: fixes an integer overflow panic when called with `target_index == u32::MAX`, and changes out of range targets to clamp and replenish up to `BIP32_MAX_INDEX` instead of silently doing nothing.

### Notes to the reviewers

This revisits the earlier implementation attempt (#1792) and implements the clamp-in-`apply_changeset` approach from that review discussion.

The second and third fixes were found while adding the boundary tests requested in review of an earlier version of this PR.

Also added a test documenting current saturation behavior: `reveal_next_spk`/`next_unused_spk` return the last revealed script with an empty `ChangeSet`, not `None`, once the index is saturated. This matches `reveal_next_spk`'s existing doc comment.

Went through the other index-taking methods (`spk_at_index`, `is_used`, `mark_used`, `unmark_used`) as well. These are plain lookups with no boundary-sensitive logic, so no dedicated tests were added there.

### Changelog notice

- Clamped `last_revealed` to `BIP32_MAX_INDEX` in `KeychainTxOutIndex::apply_changeset` to prevent it accepting an out-of-range derivation index.
- Fixed an off-by-one in `replenish_inner_index` that could panic when revealing up to `BIP32_MAX_INDEX`.
- Fixed an integer overflow in `lookahead_to_target` when called with `target_index == u32::MAX`, and changed out of range targets to clamp to `BIP32_MAX_INDEX` instead of silently doing nothing.

### Checklists

**All Submissions**

- [x] I've signed all my commits
- [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
- [x]  I ran `just p` before committing

**Bugfixes**

- [ ] This pull request breaks the existing API
- [x] I've added tests to cover the issue which are now passing
- [x] I'm linking the issue being fixed by this PR

### DIFF
--- a/crates/chain/src/indexer/keychain_txout.rs
+++ b/crates/chain/src/indexer/keychain_txout.rs
@@ -992,6 +992,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
             }
         }
         for (did, index) in changeset.last_revealed {
+            let index = index.min(BIP32_MAX_INDEX);
             let v = self.last_revealed.entry(did).or_default();
             *v = index.max(*v);
             self.replenish_inner_index_did(did, self.lookahead);
@@ -1225,5 +1226,36 @@ mod test {
         // The cache is optional at load time
         let index = KeychainTxOutIndex::<i32>::from_changeset(lookahead, false, init_cs);
         assert!(index.spk_cache.is_empty());
+    }
+
+    #[test]
+    fn apply_changeset_clamps_out_of_range_index_and_derives_correct_spk() {
+        let s = DESCRIPTORS[0];
+        let desc = Descriptor::parse_descriptor(&Secp256k1::new(), s)
+            .unwrap()
+            .0;
+        let mut index = KeychainTxOutIndex::new(0, false);
+        let did = desc.descriptor_id();
+        let _ = index.insert_descriptor(0i32, desc.clone());
+        let seed_index = BIP32_MAX_INDEX - 1;
+        let spk = desc
+            .at_derivation_index(seed_index)
+            .unwrap()
+            .script_pubkey();
+        index.inner.insert_spk((0i32, seed_index), spk);
+        let changeset = ChangeSet {
+            last_revealed: [(did, BIP32_MAX_INDEX + 1)].into(),
+            ..Default::default()
+        };
+        index.apply_changeset(changeset);
+        assert_eq!(index.last_revealed_index(0i32), Some(BIP32_MAX_INDEX));
+        assert_eq!(
+            index.spk_at_index(0i32, BIP32_MAX_INDEX),
+            Some(
+                desc.at_derivation_index(BIP32_MAX_INDEX)
+                    .unwrap()
+                    .script_pubkey()
+            )
+        );
     }
 }
--- a/crates/chain/src/indexer/keychain_txout.rs
+++ b/crates/chain/src/indexer/keychain_txout.rs
@@ -601,7 +601,9 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         // Exclusive: index to stop at.
         let stop_index = if descriptor.has_wildcard() {
             let next_reveal_index = self.last_revealed.get(&did).map_or(0, |v| *v + 1);
-            (next_reveal_index + lookahead).min(BIP32_MAX_INDEX)
+            next_reveal_index
+                .saturating_add(lookahead)
+                .min(BIP32_MAX_INDEX + 1)
         } else {
             1
         };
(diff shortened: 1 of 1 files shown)


evanlinjin (commented):
Thanks for the PR.

Given you have already started the work, we should probably also test:

* Given the derivation index is saturated, `next_unused_spk` should return `None` and `reveal_next_spk` should return `None`.
* What happens if we use an index greater than `BIP32_MAX_INDEX` for `reveal_to_target`.
* What happens if we use `BIP32_MAX_INDEX` for `reveal_to_target`.
* Any method that takes an index value should be tested with `BIP32_MAX_INDEX` and `> BIP32_MAX_INDEX`.

busayo-OD:
> Given you have already started the work, we should probably also test:
> 
>     * Given the derivation index is saturated, `next_unused_spk` should return `None` and `reveal_next_spk` should return `None`.
> 
>     * What happens if we use an index greater than `BIP32_MAX_INDEX` for `reveal_to_target`.
> 
>     * What happens if we use `BIP32_MAX_INDEX` for `reveal_to_target`.
> 
>     * Any method that takes an index value should be tested with `BIP32_MAX_INDEX` and `> BIP32_MAX_INDEX`.

Thanks for the review. I’ve added the requested boundary tests. Two additional issues surfaced while adding them and have been fixed too.

For saturation, the current behavior is to return the last revealed SPK with an empty `ChangeSet`, rather than `None`, which matches the existing `reveal_next_spk` documentation. Is `None` expected here?

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -992,6 +992,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
             }
         }
         for (did, index) in changeset.last_revealed {
+            let index = index.min(BIP32_MAX_INDEX);
```
evanlinjin, line 996: I propose adding `debug_assert!(index < BIP32_MAX_INDEX)` above this line so that overflows can be caught during development (it shouldn't really happen).

busayo-OD: Tried the `debug_assert!` here, but it breaks `apply_changeset_clamps_out_of_range_index_and_derives_correct_spk`, which passes `BIP32_MAX_INDEX + 1` to verify that `apply_changeset` clamps the value.

I also moved the test inline and changed it to use a wildcard descriptor so it exercises the boundary arithmetic. I added an SPK correctness check at the boundary as well.

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -601,7 +602,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         // Exclusive: index to stop at.
         let stop_index = if descriptor.has_wildcard() {
             let next_reveal_index = self.last_revealed.get(&did).map_or(0, |v| *v + 1);
-            (next_reveal_index + lookahead).min(BIP32_MAX_INDEX)
+            (next_reveal_index + lookahead).min(BIP32_MAX_INDEX.saturating_add(1))
```
evanlinjin:
```suggestion
            next_reveal_index.saturating_add(lookahead).min(BIP32_MAX_INDEX + 1)
```

In `lookahead_to_target`, `temp_lookahead` may surpass `BIP32_MAX_INDEX - lookahead`, making `stop_index > BIP32_MAX_INDEX`.

`lookahead_to_target` no longer needs the "checked" calls either:

```rust
let temp_lookahead = target_index.saturating_add(1).saturating_sub(next_index);
if temp_lookahead > 0 {
    // replenish inner index.
}
```

crates/chain/src/indexer/keychain_txout.rs
```diff
@@ -1226,4 +1228,141 @@ mod test {
...
+        index.inner.insert_spk((0i32, seed_index), spk);
+        index.last_revealed.insert(did, seed_index);
+
+        let (spks, _changeset) = index.reveal_to_target(0i32, BIP32_MAX_INDEX + 5).unwrap();
+        assert_eq!(spks.len(), 1);
+        assert_eq!(spks[0].0, BIP32_MAX_INDEX);
+    }
+
+    #[test]
+    fn lookahead_to_target_with_u32_max_does_not_overflow() {
```
evanlinjin: I don't think this is the right behavior. Calling `lookahead_to_target` to a value greater than BIP32_MAX should clamp the target to BIP32_MAX (not silently do nothing).

I propose replacing this test with the following two:

```rust
    /// Build an index with `last_revealed` (and the matching spk) at `last_revealed_index`.
    fn indexer_with_last_revealed_at(last_revealed_index: u32) -> KeychainTxOutIndex<i32> {
        let s = DESCRIPTORS[0];
        let desc = Descriptor::parse_descriptor(&Secp256k1::new(), s)
            .unwrap()
            .0;
        let mut index = KeychainTxOutIndex::new(0, false);
        let did = desc.descriptor_id();
        let _ = index.insert_descriptor(0i32, desc.clone());

        let spk = desc
            .at_derivation_index(last_revealed_index)
            .unwrap()
            .script_pubkey();
        index.inner.insert_spk((0i32, last_revealed_index), spk);
        index.last_revealed.insert(did, last_revealed_index);
        index
    }

    #[test]
    fn lookahead_to_target_past_bip32_max_clamps() {
        let mut index = indexer_with_last_revealed_at(BIP32_MAX_INDEX - 1);
        // `u32::MAX` is past `BIP32_MAX_INDEX`, so it must clamp down to it and store the final
        // spk.
        let changeset = index.lookahead_to_target(0i32, u32::MAX);
        assert!(
            changeset.last_revealed.is_empty(),
            "lookahead must not reveal"
        );
        // `spk_at_index` is aware of lookahead spks.
        assert!(index.spk_at_index(0i32, BIP32_MAX_INDEX).is_some());
    }

    #[test]
    fn lookahead_to_target_at_bip32_max_does_not_overflow() {
        let mut index = indexer_with_last_revealed_at(BIP32_MAX_INDEX);
        let changeset = index.lookahead_to_target(0i32, u32::MAX - 1);
        assert!(
            changeset.last_revealed.is_empty(),
            "lookahead must not reveal"
        );
    }

```

evanlinjin (changes requested):

crates/chain/src/indexer/keychain_txout.rs
busayo-OD: Fixed, thanks

crates/chain/src/indexer/keychain_txout.rs
busayo-OD: Done. Also used the helper to clean up repeated setup in a few other tests.

evanlinjin (approved):
ACK 1ca8d29854d41269982a1f47b03603a98a221d99

notmandatory (approved):
ACK 1ca8d29854d41269982a1f47b03603a98a221d99

Thanks @busayo-OD this looks like a good focused bug fix.

PR #2276, opened 2026-09-07
fix(chain): preserve first floating txout

### LINKED ISSUE #2274: `TxGraph::insert_txout` overwrites existing floating txout but returns empty `ChangeSet`
**Describe the bug**  
`TxGraph::insert_txout` on an outpoint already present as a floating txout overwrites it in memory but returns an empty `ChangeSet`.

- Debug: panics.
- Release: live graph holds the new `TxOut`, changeset is empty, so persisted state keeps the old one. `IndexedTxGraph::insert_txout` forwards the empty changeset, so the indexer never sees the replacement.

Contradicts the doc comment ("`ChangeSet` result will be empty if the `outpoint` ... already existed") and the assert's own message ("txout of the same outpoint should never change").

**To Reproduce**  
Run with `--release` (debug panics on the assert instead):

```rust
use bdk_chain::{tx_graph::TxGraph, Merge};
use bitcoin::{hashes::Hash, Amount, OutPoint, ScriptBuf, TxOut, Txid};

let op = OutPoint::new(Txid::all_zeros(), 0);
let a = TxOut { value: Amount::from_sat(1_000), script_pubkey: ScriptBuf::new() };
let b = TxOut { value: Amount::from_sat(2_000), script_pubkey: ScriptBuf::new() };

let mut graph: TxGraph = TxGraph::default();
let _ = graph.insert_txout(op, a.clone());
let cs = graph.insert_txout(op, b);

assert!(cs.is_empty());                    // passes: reports "no change"
assert_eq!(graph.get_txout(op), Some(&a)); // FAILS: graph now holds `b`
```

**Expected behavior**  
First insert wins: an existing outpoint is never overwritten, and an empty changeset truthfully means no change. 

**Build environment**  
- BDK tag/commit: acc06e53

- [x] No

Tiancheng-Xu:
Fixes #2274

`TxGraph::insert_txout` previously used `BTreeMap::insert` before checking whether an output already existed. In release builds this replaced an existing floating output while returning an empty `ChangeSet`.

Keep the first floating output and add a regression test that verifies a conflicting second insertion is ignored.

Validation:
- `git diff --check`
- Targeted test not run: `cargo` is unavailable in the local environment.

### DIFF
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -630,12 +630,16 @@ impl<A: Anchor> TxGraph<A> {
                 // written assuming this never panics.
             }
             TxNodeInternal::Partial(partial_tx) => match partial_tx.entry(outpoint.vout) {
-                crate::collections::btree_map::Entry::Occupied(entry) => {
-                    debug_assert_eq!(
-                        entry.get(),
-                        &txout,
-                        "txout of the same outpoint should never change"
-                    );
+                crate::collections::btree_map::Entry::Occupied(mut entry) => {
+                    if entry.get() != &txout {
+                        debug_assert_eq!(
+                            entry.get(),
+                            &txout,
+                            "txout of the same outpoint should never change"
+                        );
+                        entry.insert(txout.clone());
+                        changeset.txouts.insert(outpoint, txout);
+                    }
                 }
                 crate::collections::btree_map::Entry::Vacant(entry) => {
                     entry.insert(txout.clone());
--- a/crates/chain/src/tx_graph.rs
+++ b/crates/chain/src/tx_graph.rs
@@ -629,12 +629,19 @@ impl<A: Anchor> TxGraph<A> {
                 // replaced was actually correct is a good idea but the tests have already been
                 // written assuming this never panics.
             }
-            TxNodeInternal::Partial(partial_tx) => {
-                if !partial_tx.contains_key(&outpoint.vout) {
-                    partial_tx.insert(outpoint.vout, txout.clone());
+            TxNodeInternal::Partial(partial_tx) => match partial_tx.entry(outpoint.vout) {
+                crate::collections::btree_map::Entry::Occupied(entry) => {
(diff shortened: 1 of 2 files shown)


evanlinjin:
Thank you for working on this.

I'm wondering whether this is the best behavior for handling floating txout replacements. If there is a bug in the chain source that gave us an incorrect txout for an outpoint, we will not be able to correct the behavior without "resetting" the `TxGraph`. In other words, a chain source can never correct a mistake in `TxGraph`.

I propose the following:

* If we insert a floating `TxOut` that's different to what we currently have, replace it and emit a changeset.
* For debug builds, we `debug_assert_eq!` it so that the bug can be caught during development.

cc. @noahjoeris

Tiancheng-Xu:
Followed the maintainer guidance in commit `3b8c1477`: conflicting floating `TxOut` values now replace the existing value and emit a `ChangeSet` in release builds, while debug builds retain the invariant assertion. The `BTreeMap::entry` path avoids duplicate lookups. Updated the regression to cover replacement and emitted changes. `git diff --check` passes; cargo/rustc are unavailable in this environment, so I could not run the Rust suite locally.

evanlinjin:
@Tiancheng-Xu what environment are you talking about?

evanlinjin:
Thanks for taking this on, and for iterating on the semantics after my last comment.

I'm going to close this one. The blocker is that the patch hasn't been compiled or tested by anyone before CI: `cargo` and `rustc` aren't available in the environment you're working in, so there's no way for you to iterate on review feedback in a `bdk_chain` change without a full round-trip through me. For consensus-adjacent state handling that's not a workflow I can review responsibly.

There's also more to the fix than the two files here. For replace-and-emit to actually stop live and persisted state from diverging, `ChangeSet::merge` and the persistence backends need to replace on conflict too, not ignore. That wants checking before anything lands.

If you get a working Rust toolchain set up, I'd welcome you back on this or another issue. #2274 stays open in the meantime.

PR #2283, opened 2026-09-09
fix(rpc): harden `Emitter::mempool` against a malicious RPC

### LINKED ISSUE #2282: bdk_bitcoind_rpc: Unbounded spin loop, memory growth, and missing txid validation in Emitter::mempool_at
**Describe the bug**  

In `Emitter::mempool_at`, there are three related issues regarding RPC polling behavior, memory usage, and transaction ID validation:
- Unbounded tip-consistency spin loop: The loop in `mempool_at` retries `get_block_count` and `get_block_hash` until two consecutive calls agree on the tip. If an RPC endpoint flaps or alternates responses between calls, this loop spins indefinitely without an iteration limit, timeout, or backoff causing high CPU usage and rapid RPC request flooding.
- Unbounded `mempool_snapshot` memory growth: Every txid returned by `getrawmempool` that is not already cached is fetched in full via `get_raw_transaction` and inserted into `mempool_snapshot`. There is no limit on the number of transactions fetched or tracked per poll, leading to uncontrolled memory growth if the node presents an arbitrary number of transactions.
- Missing txid validation on fetch: When inserting fetched transactions into `mempool_snapshot` via `self.mempool_snapshot.insert(txid, tx.clone())`, the map key is set to the requested txid rather than the computed ID of the returned transaction (tx.compute_txid()). If the RPC server returns a transaction body whose actual txid does not match the requested txid, the crate accepts and caches the mismatched transaction without error.

**To Reproduce**  

- Set up Emitter::mempool connected to a Bitcoin RPC mock server.
- For the spin loop: Configure the mock RPC server to alternate tip hash or height responses on consecutive calls during mempool_at. Observe that the client spins indefinitely in the retry loop.
- For unverified txid insertion: Configure the mock RPC so getrawmempool returns txid X. When getrawtransaction is called for X, have the mock RPC return the raw transaction bytes for a different transaction Y.
- Run Emitter::mempool. Observe that the crate emits transaction Y without throwing a txid mismatch error, storing the mis-keyed entry in mempool_snapshot.

**Expected behavior**  

- The tip-consistency loop should be bounded with a maximum retry cap or backoff strategy, returning an error if consistency cannot be established within N retries. 
- `mempool_snapshot` tracking should impose limits on fetched transactions per poll to guard against unbounded memory consumption.
- Transactions fetched via `get_raw_transaction` should be verified (tx.compute_txid() == txid) before being inserted into mempool_snapshot

**Build environment**  
- BDK tag/commit: 337e9d68
- Rust/Cargo version: 1.75.0 or higher

- [ ] Electrum
- [ ] Esplora
- [x] Bitcoin Core RPC

- [ ] Yes
- [x] No

**Additional context**

tvpeter:
### Description

This PR fixes `Emitter::mempool` and `mempool_at`  to protect  against DOS and data spoofing vectors originating from a malicious RPC node or a MITM actor. The updates address three vulnerabilities:
- Unbounded Tip-Consistency Loop: Prevents infinite spin loops when a node returns flapping tip headers.
- Unverified Txid Validation: Rejects fetched txs bodies whose computed hash does not match the requested txid.
- Unbounded Memory Usage: Limits tracked mempool transactions in mempool_snapshot to prevent memory exhaustion.

##### Problem and fix

- Unbounded tip-consistency spin loop: The loop in `mempool_at` retries `get_block_count` and `get_block_hash` until two consecutive calls agree on the tip. If an RPC node alternates responses between calls, this loop spins indefinitely  causing high CPU usage and rapid RPC request flooding.
_Fix_: bound the loop with MEMPOOL_TIP_CONSISTENCY_MAX_ATTEMPTS
- Unbounded mempool_snapshot memory growth: Every txid returned by `getrawmempool` that is not already cached is fetched in full via `get_raw_transaction` and inserted into `mempool_snapshot`. There is no limit on the number of transactions fetched or tracked per poll, leading to uncontrolled memory growth if the node presents an arbitrary number of transactions.
_Fix_: track at most `max_mempool_txs `transactions (default DEFAULT_MAX_MEMPOOL_TXS = 100_000, configurable via Emitter::with_max_mempool_txs). Once the cap is reached, unknown mempool transactions are skipped. The emitter’s own known transactions bypass the fetch path and are never dropped.
- Missing txid validation on fetch: When inserting fetched transactions into `mempool_snapshot` via `self.mempool_snapshot.insert(txid, tx.clone())`, the map key is set to the requested txid rather than `tx.compute_txid()`. If the RPC server returns a transaction body whose actual txid does not match the requested txid, the crate accepts and caches the mismatched tx without error.
_Fix_: recompute the txid of the fetched tx body and reject it on mismatch, before it is cached or emitted

Closes #2282 

### Notes to the reviewers

- New pub const `DEFAULT_MAX_MEMPOOL_TXS: usize = 100_000`
- New builder method `Emitter::with_max_mempool_txs(usize)`
- No changes to `Emitter::new` or any existing method signatures.
- Failure paths in the fixes reuse the existing bitcoincore_rpc::Error::UnexpectedStructure variant, so no new error type is introduced. mempool_at may now return Err or a truncated snapshot in situations where it previously hung or grew unboundedly

### Changelog notice

 - Fixed`Emitter::mempool`/`mempool_at`: bound the tip-consistency retry loop so a  flapping or malicious node cannot spin it forever, and reject transaction bodies  whose computed txid does not match the requested txid instead of caching a
mis-keyed entry.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/crates/bitcoind_rpc/src/lib.rs
+++ b/crates/bitcoind_rpc/src/lib.rs
@@ -54,6 +54,10 @@ pub struct Emitter<C> {
     /// sure the tip block is already emitted. When a block is emitted, the transactions in the
     /// block are removed from this field.
     mempool_snapshot: HashMap<Txid, Arc<Transaction>>,
+    /// Maximum number of transactions to track in `mempool_snapshot`. Once reached, unknown
+    /// mempool transactions are skipped rather than fetched and stored, bounding memory so a
+    /// malicious node cannot grow the snapshot unboundedly. See [`Emitter::with_max_mempool_txs`].
+    max_mempool_txs: usize,
 }
 
 /// Indicates that there are no initially-expected mempool transactions.
@@ -61,6 +65,18 @@ pub struct Emitter<C> {
 /// Use this as the `expected_mempool_txs` field of [`Emitter::new`] when the wallet is known
 /// to start empty (i.e. with no unconfirmed transactions).
 pub const NO_EXPECTED_MEMPOOL_TXS: core::iter::Empty<Arc<Transaction>> = core::iter::empty();
+
+/// Maximum attempts to obtain a mempool snapshot consistent with the node's tip before
+/// [`Emitter::mempool_at`] returns an error. Bounds the retry loop so a malicious
+/// node cannot spin it forever.
+const MEMPOOL_TIP_CONSISTENCY_MAX_ATTEMPTS: usize = 10;
+/// Default maximum number of transactions an [`Emitter`] tracks in its mempool snapshot.
+///
+/// Bounds the memory used by `mempool_snapshot` so a malicious or compromised node cannot grow it
+/// unboundedly by streaming distinct txids. Override with [`Emitter::with_max_mempool_txs`]. Chosen
+/// to comfortably exceed a default-configured node's mempool while still bounding worst-case
+/// memory; the wallet's own transactions are never dropped, as they bypass the fetch path.
+pub const DEFAULT_MAX_MEMPOOL_TXS: usize = 100_000;
 
 impl<C> Emitter<C>
 where
@@ -96,7 +112,23 @@ where
                     (tx.compute_txid(), tx)
                 })
                 .collect(),
+            max_mempool_txs: DEFAULT_MAX_MEMPOOL_TXS,
         }
+    }
+
+    /// Set the maximum number of transactions the [`Emitter`] tracks in its mempool snapshot.
+    ///
+    /// Once the snapshot holds this many transactions, further *unknown* mempool transactions are
+    /// skipped for the current and subsequent polls (they are not fetched or stored) until existing
+    /// entries are freed by confirmations or by catching up to the node's tip. This bounds the
+    /// memory a malicious node can force the emitter to allocate.
+    ///
+    /// Transactions already known to the emitter — those provided via `expected_mempool_txs` in
+    /// [`Emitter::new`] or seen in a previous poll — are never dropped, as they bypass the fetch
+    /// path. Defaults to [`DEFAULT_MAX_MEMPOOL_TXS`].
+    pub fn with_max_mempool_txs(mut self, max_mempool_txs: usize) -> Self {
+        self.max_mempool_txs = max_mempool_txs;
+        self
     }
 
     /// Emit mempool transactions and any evicted [`Txid`]s.
@@ -151,6 +183,20 @@ where
     /// `sync_time` is in unix seconds.
     ///
     /// This is the no-std version of [`mempool`](Self::mempool).
+    ///
+    /// # Errors
+    ///
+    /// Returns [`bitcoincore_rpc::Error::UnexpectedStructure`] if the node's tip keeps changing
+    /// across a bounded number of consistency retries, or if the node returns a transaction body
+    /// whose computed txid does not match the requested txid. In the latter case the entire poll
+    /// fails; retry on the next poll.
+    ///
+    /// # Truncation
+    ///
+    /// At most [`with_max_mempool_txs`](Self::with_max_mempool_txs) transactions are tracked. If
+    /// the node's mempool exceeds that, unknown transactions beyond the limit are silently
+    /// omitted from the returned snapshot (the emitter's own known transactions are always
+    /// included).
     pub fn mempool_at(&mut self, sync_time: u64) -> Result<MempoolEvent, bitcoincore_rpc::Error> {
         let client = &*self.client;
 
@@ -158,6 +204,7 @@ where
         let mut rpc_tip_hash;
         let mut rpc_mempool;
         let mut rpc_mempool_txids;
+        let mut attempts: usize = 0;
 
         // Ensure we get a mempool snapshot consistent with `rpc_tip_hash` as the tip.
         loop {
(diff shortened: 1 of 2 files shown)


crates/bitcoind_rpc/src/lib.rs
```diff
@@ -62,6 +66,18 @@ pub struct Emitter<C> {
...
+/// [`Emitter::mempool_at`] returns an error. Bounds the retry loop so a malicious
+/// node cannot spin it forever.
+const MEMPOOL_TIP_CONSISTENCY_MAX_ATTEMPTS: usize = 10;
+/// Default maximum number of transactions an [`Emitter`] tracks in its mempool snapshot.
+///
+/// Bounds the memory used by `mempool_snapshot` so a malicious or compromised node cannot grow it
+/// unboundedly by streaming distinct txids. Override with [`Emitter::with_max_mempool_txs`]. Chosen
+/// to comfortably exceed a default-configured node's mempool while still bounding worst-case
+/// memory; the wallet's own transactions are never dropped, as they bypass the fetch path.
+pub const DEFAULT_MAX_MEMPOOL_TXS: usize = 100_000;
```
notmandatory, line 79: How did you choose this default? it seems low to me to be a max value. The current mempool has about 80K tx in it. I'd expect even a much larger max would prevent the unbounded malicious issue without risking the number being too low and ignoring tx from a valid non-malicious node.

notmandatory:
I'm not 100% convinced this needs to be fixed. What examples can you think of where someone would be connecting to a malicious node? The general recommended practice is to only connect to core nodes you control and I don't know of any public services that provide public RPC access.

crates/bitcoind_rpc/src/lib.rs
tvpeter: I did not think thoroughly on this default count, as I assumed that users will be more inclined to set a count that serves them and that is why I added the `Emitter::with_max_mempool_txs` method to allow them to set the count that is suitable for them. But as you noted, I checked and it should ideally be on an upper ceiling of 1m based on Core’s `maxmempool` default size of 300mb.

tvpeter:
> I'm not 100% convinced this needs to be fixed. What examples can you think of where someone would be connecting to a malicious node? The general recommended practice is to only connect to core nodes you control and I don't know of any public services that provide public RPC access.

For anyone running their own setup, the risk is near zero. However, for a remote setup (node on a VPS), with no TLS, there is a (little) chance of MITM attack. However, that may not be a good argument to warrant these changes. The only sub-issue I think should be considered is the transaction validation on fetch which is already done in the [BdkElectrumClient::fetch_tx](https://github.com/bitcoindevkit/bdk/blob/38b52d99fdcc772ab46e99ba73bfd0591f8c0325/crates/electrum/src/bdk_electrum_client.rs#L75).  I'm happy to scope it to just that issue or drop the entire PR if it does not merit a change.

Thank you.

notmandatory:
> For anyone running their own setup, the risk is near zero. However, for a remote setup (node on a VPS), with no TLS, there is a (little) chance of MITM attack. However, that may not be a good argument to warrant these changes. The only sub-issue I think should be considered is the transaction validation on fetch which is already done in the [BdkElectrumClient::fetch_tx](https://github.com/bitcoindevkit/bdk/blob/38b52d99fdcc772ab46e99ba73bfd0591f8c0325/crates/electrum/src/bdk_electrum_client.rs#L75). I'm happy to scope it to just that issue or drop the entire PR if it does not merit a change.

Yes I think the code changes here are out of scope but double check if there's somewhere in the docs we need to mention that this feature assumes the user is connected to a bitcoind node they trust. And given that assumption I don't think the tx validation check is needed either. Although it seems like a basic safety thing, I have a concern about how it would impact performance. The RPC interface is ideal for users with large wallets (ie. some sort of shared custodial setup).

But I could be wrong that adding this check has any performance impact, and if it's negligible then it doesn't hurt to add and I'd support doing it. Think about how you might test to see what the impact is on a test wallet validating 1000s of tx during initial sync.  Thanks for looking into this.

notmandatory:
Also want to mention, your code and this discussion are still valuable even if nothing gets merged. It helps other devs (or LLMS 🤖) who might see the same potential issue.

tvpeter:
> Yes I think the code changes here are out of scope but double check if there's somewhere in the docs we need to mention that this feature assumes the user is connected to a bitcoind node they trust.

I think the most suitable place to consider is a fresh section at the module level documentation [here](https://github.com/bitcoindevkit/bdk/blob/38b52d99fdcc772ab46e99ba73bfd0591f8c0325/crates/bitcoind_rpc/src/lib.rs#L9)
> 
> But I could be wrong that adding this check has any performance impact, and if it's negligible then it doesn't hurt to add and I'd support doing it. Think about how you might test to see what the impact is on a test wallet validating 1000s of tx during initial sync. 

I ran a benchmark on my fork ([#1](https://github.com/tvpeter/bdk/pull/1)) and the check's impact seemed negligible. It costs a stable ~0.8 µs/tx for a typical tx (~0.6–2.7 µs across sizes), roughly 2× a bare deserialize, but both are sub-microsecond and it is <1% of the full get_raw_transaction fetch. Extrapolated, even a full ~100k-tx mempool is ~80 ms of hashing on top of a fetch phase measured in tens of seconds. It also runs only on the mempool path and once per newly-fetched tx. 

So if you consider the above performance impact level acceptable, then I will scope the PR to a documentation change and the check only otherwise I will happily close it.

Thank you

notmandatory:
I agree, based on your performance testing this can be scoped to the documentation and check only.

tvpeter:
Closing this PR in favour of #2316

PR #2297, opened 2026-09-10
fix(chain): persist corrected floating txouts

### LINKED ISSUE #2274: `TxGraph::insert_txout` overwrites existing floating txout but returns empty `ChangeSet`

Tiancheng-Xu:
### Description

Fixes #2274. Floating `TxOut` corrections now remain consistent across the in-memory graph, aggregated `ChangeSet`s, and SQLite persistence:

- make `ChangeSet::merge` explicitly keep the newer floating `TxOut` for a conflicting outpoint;
- keep the existing replacement behavior in `TxGraph::insert_txout` and clarify its documentation;
- add regressions for changeset aggregation and SQLite round-tripping.

This is a new proposal after #2276 was closed. The earlier PR's maintainer feedback about `ChangeSet::merge` and persistence is addressed here.

### Verification

- `cargo fmt --all -- --check` (stable toolchain; emits only the repository's existing nightly-option warnings)
- `cargo test -p bdk_chain --features rusqlite` — 14 unit tests, all integration suites, and 16 doc tests passed (1 ignored by repository configuration)
- `git diff --check`
- commit `66346d8c` is GPG-signed; the public key is being registered on the GitHub account

### Checklist

- [x] Added regression tests for the bug
- [x] Linked the issue being fixed
- [x] Followed the repository contribution workflow

### DIFF
--- a/crates/chain/src/rusqlite_impl.rs
+++ b/crates/chain/src/rusqlite_impl.rs
@@ -665,7 +665,7 @@ mod test {
     use super::*;
 
     use bdk_testenv::{anyhow, hash};
-    use bitcoin::{absolute, transaction, TxIn, TxOut};
+    use bitcoin::{absolute, transaction, Amount, OutPoint, ScriptBuf, TxIn, TxOut};
 
     #[test]
     fn can_persist_anchors_and_txs_independently() -> anyhow::Result<()> {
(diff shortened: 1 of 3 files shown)


evanlinjin:
Thanks for the work on this, and for getting a toolchain set up.

Please stop work here — I'm taking #2274 over. The semantics need settling before anything lands: what `insert_txout` should do on a conflicting `TxOut`, and how that interacts with `ChangeSet::merge` and persistence. That's a call for me and the other maintainers to make, not something to resolve through review rounds.

Closing the PR.

PR #2299, opened 2026-09-11
refactor(chain)!: drop the bound framing from confirmation APIs

### LINKED ISSUE #2298: Drop the "anchor may be a descendant" framing; introduce `Anchor::confirmation_height`
### Summary

The `Anchor` trait documents that an anchor block may be a *descendant* of the block that actually confirmed a transaction. Nothing in the codebase produces such an anchor, and no chain source can. The framing costs us an "upper bound" name and a trail of pessimistic caveats through the public API for a case that does not occur.

Proposal: drop the framing, and introduce `Anchor::confirmation_height` alongside a deprecated `confirmation_height_upper_bound`.

### The framing

`crates/chain/src/tx_data_traits.rs`:

> If transaction A is anchored in block B, and block B is in the best chain, we can assume that transaction A is also confirmed in the best chain. **This does not necessarily mean that transaction A is confirmed in block B. It could also mean transaction A is confirmed in a parent block of B.**

and on the method:

> Get the upper bound of the chain data's confirmation height.
> The default definition gives a pessimistic answer.

### Why it does not hold in practice

Every anchor constructed in this repo uses the confirming block itself:

- **electrum** (`crates/electrum/src/bdk_electrum_client.rs:600`) builds the anchor only after `validate_merkle_proof(&txid, &header.merkle_root, &proof)` succeeds — the block provably contains the transaction.
- **esplora** (`crates/esplora/src/lib.rs:54`, and the `*_ext.rs` sites) uses `TxStatus { block_height, block_hash }`, which is the confirming block.
- **`TxPosInBlock`** (`crates/chain/src/tx_data_traits.rs:111`) carries `tx_pos`, the transaction's index *within that block* — structurally the confirming block. Both `From<TxPosInBlock>` impls (`BlockId`, `ConfirmationBlockTime`) pass `block_id` straight through.

Both in-tree `Anchor` impls define `confirmation_height_upper_bound()` as exactly `anchor_block().height`.

### What it costs

The pessimism propagates into unrelated public docs. `CanonicalTxOut::is_mature` and `is_confirmed_and_spendable` (`crates/chain/src/canonical.rs:148`, `:175`) both carry:

> Depending on the implementation of `confirmation_height_upper_bound` in `Anchor`, this method may return false-negatives. In other words, interpreted confirmation count may be less than the actual value.

So a caller reading the maturity API is told their coin may be reported immature when it isn't — a caveat that is unreachable with any anchor type that exists.

It also leaves ambiguity in ordering decisions. `BlockQueries::resolve_candidates` picks the lowest in-chain candidate among a transaction's anchors; under the current framing "several anchors in the best chain at once" is legal, so that choice is load-bearing. If anchors always name the confirming block, at most one can be in the best chain and the tie-break is defensive only.

### Proposed change

**1. Remove the "parent block" paragraph** from the `Anchor` trait docs, and the "pessimistic answer" note from the method. State instead that the anchor block is the block that confirmed the transaction.

**2. Add `confirmation_height`, deprecate `confirmation_height_upper_bound`.**

The delegation direction matters. The obvious version is wrong:

```rust
// WRONG: silently ignores existing overrides
fn confirmation_height(&self) -> u32 { self.anchor_block().height }

#[deprecated(note = "use `confirmation_height`")]
fn confirmation_height_upper_bound(&self) -> u32 { self.confirmation_height() }
```

A downstream impl that overrides `confirmation_height_upper_bound` today keeps compiling, but every internal call site moves to `confirmation_height` and stops seeing that override — a silent behaviour change, the worst outcome for a deprecation.

Default the *new* method to the old one instead, so existing overrides continue to be honoured:

```rust
fn confirmation_height(&self) -> u32 {
    #[allow(deprecated)]
    self.confirmation_height_upper_bound()
}

#[deprecated(since = "TBD", note = "use `confirmation_height`")]
fn confirmation_height_upper_bound(&self) -> u32 {
    self.anchor_block().height
}
```

Then drop `confirmation_height_upper_bound` in a later release, at which point `confirmation_height` takes the `anchor_block().height` default.

**3. Rename `ChainPosition::confirmation_height_upper_bound`** (`crates/chain/src/chain_data.rs:78`) to match. This is an inherent method on a public type, so it needs its own `#[deprecated]` shim.

**4. Update the caveats** in `canonical.rs:148` and `:175` that only exist to describe the loose bound.

### Scope

Roughly 12 call sites across `canonical.rs`, `canonical_task.rs`, `canonical_view_task.rs`, `chain_data.rs`, `tx_graph.rs` (`insert_anchor`'s `txs_by_highest_conf_heights` bookkeeping), and `tx_data_traits.rs`. No behaviour change — both in-tree impls already return `anchor_block().height`.

### Open question

Is the "parent block of B" case something a chain source was ever expected to use — a compact-block or filter-based source that learns "confirmed at or before height N" without learning the exact block? If a design has that in mind, this should stay and the docs should name that source instead of leaving it hypothetical.

evanlinjin:
### Description

Fixes #2298.

The `Anchor` trait previously said an anchor block *might* be a descendant of the block that actually confirmed a transaction. Nothing in this repo ever produces that kind of anchor:

- Electrum validates a merkle proof before building one
- Esplora uses the confirming block from `TxStatus`
- `TxPosInBlock` carries the transaction’s index inside that block

So the height is exact, not an upper bound. This PR removes that framing and renames:

- `confirmation_height_upper_bound` → `confirmation_height` (on both `Anchor` and `ChainPosition`)
- `confirmations_lower_bound` → `confirmations`

The second rename follows for the same reason: a higher height produces fewer confirmations, so an upper-bounded height used to yield a lower-bounded count. With an exact height the count is exact.

### Notes to reviewers

**No behaviour change.** Every implementation already returned `anchor_block().height`. The override on `ConfirmationBlockTime` is removed because it was identical to the trait default.

**Hard renames, no deprecated shims.** The next `bdk_chain` release is already breaking, so a downstream impl that still overrides the old name should fail to compile rather than be silently ignored.

### Changelog notice

**`bdk_chain`**

- Renamed `Anchor::confirmation_height_upper_bound` and `ChainPosition::confirmation_height_upper_bound` to `confirmation_height`. An anchor names the confirming block, so the height is exact. No deprecated alias.
- Renamed `ChainPosition::confirmations_lower_bound` to `ChainPosition::confirmations` (the derived count is now exact).
- Updated `Anchor` trait docs: they no longer claim an anchor block may be a descendant of the confirming block.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/crates/chain/src/canonical.rs
+++ b/crates/chain/src/canonical.rs
@@ -107,14 +107,12 @@ impl<P: Ord> PartialOrd for CanonicalTxOut<P> {
 impl<A: Anchor> CanonicalTxOut<ChainPosition<A>> {
     /// Whether the `txout` is considered mature.
     ///
-    /// Depending on the implementation of [`confirmation_height_upper_bound`] in [`Anchor`], this
-    /// method may return false-negatives. In other words, interpreted confirmation count may be
-    /// less than the actual value.
-    ///
-    /// [`confirmation_height_upper_bound`]: Anchor::confirmation_height_upper_bound
+    /// A coinbase output is mature once [`COINBASE_MATURITY`] blocks (including the block that
+    /// confirmed it) have been mined up to and including `tip`. Non-coinbase outputs are always
+    /// mature.
     pub fn is_mature(&self, tip: u32) -> bool {
         if self.is_on_coinbase {
-            let conf_height = match self.pos.confirmation_height_upper_bound() {
+            let conf_height = match self.pos.confirmation_height() {
                 Some(height) => height,
                 None => {
                     debug_assert!(false, "coinbase tx can never be unconfirmed");
@@ -133,18 +131,12 @@ impl<A: Anchor> CanonicalTxOut<ChainPosition<A>> {
     /// Whether the utxo is/was/will be spendable with chain `tip`.
     ///
     /// This method does not take into account the lock time.
-    ///
-    /// Depending on the implementation of [`confirmation_height_upper_bound`] in [`Anchor`], this
-    /// method may return false-negatives. In other words, interpreted confirmation count may be
-    /// less than the actual value.
-    ///
-    /// [`confirmation_height_upper_bound`]: Anchor::confirmation_height_upper_bound
     pub fn is_confirmed_and_spendable(&self, tip: u32) -> bool {
         if !self.is_mature(tip) {
             return false;
         }
 
-        let conf_height = match self.pos.confirmation_height_upper_bound() {
+        let conf_height = match self.pos.confirmation_height() {
             Some(height) => height,
             None => return false,
         };
@@ -156,7 +148,7 @@ impl<A: Anchor> CanonicalTxOut<ChainPosition<A>> {
         if let Some(spend_height) = self
             .spent_by
             .as_ref()
-            .and_then(|(pos, _)| pos.confirmation_height_upper_bound())
+            .and_then(|(pos, _)| pos.confirmation_height())
         {
             if spend_height <= tip {
                 return false;
@@ -449,7 +441,7 @@ impl<A: Anchor> CanonicalView<A> {
         for (spk_i, txout) in self.filter_unspent_outpoints(outpoints) {
             match &txout.pos {
                 ChainPosition::Confirmed { anchor, .. } => {
-                    let confirmation_height = anchor.confirmation_height_upper_bound();
+                    let confirmation_height = anchor.confirmation_height();
                     let confirmations = self
                         .tip
                         .height
--- a/crates/chain/src/chain_data.rs
+++ b/crates/chain/src/chain_data.rs
@@ -74,12 +74,10 @@ impl<A: Clone> ChainPosition<&A> {
 }
 
 impl<A: Anchor> ChainPosition<A> {
-    /// Determines the upper bound of the confirmation height.
-    pub fn confirmation_height_upper_bound(&self) -> Option<u32> {
+    /// Height of the block that confirmed this position, if it is confirmed.
+    pub fn confirmation_height(&self) -> Option<u32> {
         match self {
-            ChainPosition::Confirmed { anchor, .. } => {
-                Some(anchor.confirmation_height_upper_bound())
-            }
+            ChainPosition::Confirmed { anchor, .. } => Some(anchor.confirmation_height()),
             ChainPosition::Unconfirmed { .. } => None,
         }
     }
@@ -87,8 +85,8 @@ impl<A: Anchor> ChainPosition<A> {
     /// Number of confirmations of this position, given `tip`.
     ///
     /// Returns `0` if unconfirmed, or if the confirmation height is above `tip`.
-    pub fn confirmations_lower_bound(&self, tip: u32) -> u32 {
-        let Some(height) = self.confirmation_height_upper_bound() else {
+    pub fn confirmations(&self, tip: u32) -> u32 {
+        let Some(height) = self.confirmation_height() else {
             return 0;
         };
 
@@ -352,7 +350,7 @@ mod test {
     }
 
     #[test]
-    fn test_confirmations_lower_bound() {
+    fn test_confirmations() {
         let confirmed_at = |height: u32| ChainPosition::Confirmed {
             anchor: ConfirmationBlockTime {
                 confirmation_time: 0,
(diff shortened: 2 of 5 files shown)


crates/chain/src/chain_data.rs
```diff
@@ -88,7 +86,7 @@ impl<A: Anchor> ChainPosition<A> {
     ///
     /// Returns `0` if unconfirmed, or if the confirmation height is above `tip`.
     pub fn confirmations_lower_bound(&self, tip: u32) -> u32 {
```
evanlinjin:
```suggestion
    pub fn confirmations(&self, tip: u32) -> u32 {
```

crates/chain/src/canonical.rs
```diff
@@ -107,14 +107,12 @@ impl<P: Ord> PartialOrd for CanonicalTxOut<P> {
 impl<A: Anchor> CanonicalTxOut<ChainPosition<A>> {
     /// Whether the `txout` is considered mature.
     ///
-    /// Depending on the implementation of [`confirmation_height_upper_bound`] in [`Anchor`], this
-    /// method may return false-negatives. In other words, interpreted confirmation count may be
-    /// less than the actual value.
-    ///
-    /// [`confirmation_height_upper_bound`]: Anchor::confirmation_height_upper_bound
+    /// A coinbase output is mature once [`COINBASE_MATURITY`] blocks (including the block that
+    /// confirmed it) have been mined up to and including `tip`. Non-coinbase outputs are always
```
nymius, line 111: I don't know the best way to word this, but a coin is mature if it's eligible to be included in the next block. The minimum such height is `COINBASE_MATURITY - 1` , so this should say something like: 
```rust
/// A coinbase output is mature once it can be included in the next block after
/// [`COINBASE_MATURITY`] blocks have been mined after its confirmation
```

nymius (commented):
cACK 40cb3d1bbb8d53966ff36a72b1e3ba537c5bd743

For now, if we have no use for the probabilistic case, I agree with removing it. If we need it in the future (e.g., Utreexo?), I would reintroduce it under a different name to avoid colliding with previous versions.

evanlinjin:
@nymius Thank you for the review. How do you envision UTreexo potentially using this feature?

noahjoeris (commented):
Does this also hold for transitive confirmations through a child? If the parent has no direct anchor wouldn’t `ChainPosition::confirmation_height()` still return only an upper bound, and `confirmations()` a lower bound?

noahjoeris:
for example like this

```rust
#[test]
fn transitively_confirmed_tx_reports_descendant_anchor_height() {
    let chain = local_chain![
        (0, hash!("genesis")),
        (1, hash!("parent_block")),
        (5, hash!("child_block"))
    ];
    let tip = chain.tip().block_id();
    let parent_anchor = block_id!(1, "parent_block");
    let child_anchor = block_id!(5, "child_block");

    let templates = [
        TxTemplate {
            tx_name: "parent",
            inputs: &[TxInTemplate::Bogus],
            outputs: &[TxOutTemplate::new(100_000, Some(0))],
            // Seen in the mempool, but its confirmation has not been fetched yet.
            last_seen: Some(1),
            ..Default::default()
        },
        TxTemplate {
            tx_name: "child",
            inputs: &[TxInTemplate::PrevTx("parent", 0)],
            outputs: &[TxOutTemplate::new(90_000, Some(0))],
            anchors: &[child_anchor],
            ..Default::default()
        },
    ];
    let mut env = init_graph(&templates);
    let parent_txid = env.txid_to_name["parent"];
    let child_txid = env.txid_to_name["child"];

    let parent_position = |tx_graph: &TxGraph<BlockId>| {
        chain
            .canonical_view(tx_graph, tip, Default::default())
            .txs()
            .find(|tx| tx.txid == parent_txid)
            .expect("parent must be canonical")
            .pos
    };

    // Without its own anchor, the parent borrows the child's anchor. It was really confirmed in
    // block 1 (5 confirmations at tip), but the position can only report the child's block 5.
    let pos = parent_position(&env.tx_graph);
    assert_eq!(
        pos,
        ChainPosition::Confirmed {
            anchor: child_anchor,
            transitively: Some(child_txid),
        }
    );
    assert_eq!(pos.confirmation_height(), Some(5));
    assert_eq!(pos.confirmations(tip.height), 1);

    // Once the parent's own anchor is known, the position becomes exact.
    let _ = env.tx_graph.insert_anchor(parent_txid, parent_anchor);
    let pos = parent_position(&env.tx_graph);
    assert_eq!(
        pos,
        ChainPosition::Confirmed {
            anchor: parent_anchor,
            transitively: None,
        }
    );
    assert_eq!(pos.confirmation_height(), Some(1));
    assert_eq!(pos.confirmations(tip.height), 5);
}
```

evanlinjin:
@noahjoeris That's a good point - I completely overlooked this. Yes, we should still support transitively anchored transactions found during canonicalization. The renames in `ChainPosition` are therefore unsound.

nymius:
> @nymius Thank you for the review. How do you envision UTreexo potentially using this feature?

I'm not envisioning anything yet, I was thinking along my comment to the referenced issue https://github.com/bitcoindevkit/bdk/issues/2298#issuecomment-5637328751, trying to come up with an answer to the open question.

nymius:
> @nymius Thank you for the review. How do you envision UTreexo potentially using this feature?

This has been discarded by @luisschwab in https://github.com/bitcoindevkit/bdk/issues/2298#issuecomment-5719822509

Issue #2309, opened 2026-09-13
`LocalChain::apply_changeset` silently replaces the genesis block


noahjoeris:
**Describe the bug**

`LocalChain::apply_changeset` accepts a `ChangeSet` containing `(0, Some(hash))` with a hash different from the chain's current genesis and silently rewrites the genesis block. `apply_changeset_to_checkpoint` (`crates/chain/src/local_chain.rs`) collects the existing checkpoints from the changeset's lowest height upwards, overwrites height 0 with the changeset entry, and rebuilds the chain via `LocalChain::from_blocks`, which never compares against the previous genesis.

The other mutation entrypoints enforce this invariant: `apply_update` (via `merge_chains`) returns `CannotConnectError` when the update disagrees on genesis, and `insert_block` / `disconnect_from` also refuse to alter height 0. `apply_changeset` is the only one that does not, so a persisted or externally constructed changeset with a wrong height-0 entry can move a chain onto a different genesis (e.g. another network) without any error.

This issue was found by AI.

**To Reproduce**

Add `crates/chain/tests/test_genesis_changeset.rs` and run `cargo test -p bdk_chain --test test_genesis_changeset`:

```rust
use bdk_chain::local_chain::{ChangeSet, LocalChain};
use bdk_testenv::{hash, local_chain};
use bitcoin::BlockHash;

#[test]
fn apply_changeset_does_not_replace_genesis() {
    let mut chain: LocalChain = local_chain![(0, hash!("G")), (1, hash!("A"))];
    let changeset: ChangeSet<BlockHash> = [(0, Some(hash!("not_G")))].into_iter().collect();

    let result = chain.apply_changeset(&changeset);

    assert_eq!(chain.genesis_hash(), hash!("G"), "apply_changeset returned {result:?}");
}
```

Current output:

```
assertion `left == right` failed: apply_changeset returned Ok(())
  left: 7fa96a46d03a598e64ddfaadb1525023753ae80470cacac37ebb5fbc837ebeb9
 right: dc4e3f5d6f1c6736abbdfa7a86e999ebdc7b132b82411571fc3909ac0ebc41fa
```

**Expected behavior**

`apply_changeset` should not silently replace an existing genesis block; a changeset that disagrees on height 0 should be handled consistently with the other `LocalChain` mutation methods.

PR #2311, opened 2026-09-13
fix(chain): reject a changeset that replaces the genesis block

wangzhengzhuo05:
Solves #2309

### Description

`LocalChain::apply_changeset` was the only mutation entrypoint that did not enforce the
height-0 invariant. A `ChangeSet` carrying `(0, Some(hash))` whose hash differs from the
chain's genesis was accepted: `apply_changeset_to_checkpoint` overwrote height 0 and rebuilt
the chain through `LocalChain::from_blocks`, so the chain silently moved onto a different
chain (e.g. another network). `apply_update` (via `merge_chains`), `insert_block` and
`disconnect_from` all refuse to alter height 0.

This adds `ApplyBlockError::CannotReplaceGenesis` and rejects the changeset before anything
is applied:

- `apply_changeset_to_checkpoint` checks a `Some(_)` height-0 entry against the existing
  genesis. If that genesis is not a placeholder and the hashes disagree, it returns
  `Err(ApplyBlockError::CannotReplaceGenesis { expected })` and leaves the chain untouched.
- A height-0 entry that agrees with the current genesis, and one that upgrades a
  *placeholder* genesis to real data, still apply — the same case `merge_chains` already
  allows via `!o.is_placeholder()`.
- `merge_chains` maps the new variant to `CannotConnectError { try_include_height: 0 }`,
  consistent with its own genesis guard.
- `apply_changeset`'s public signature is unchanged. `ApplyBlockError` gains one variant,
  which is a breaking change for exhaustive downstream matches.

### Tests

`crates/chain/tests/test_genesis_changeset.rs` (new file, the path suggested in the issue):

- `apply_changeset_does_not_replace_genesis` — the issue's reproduction: the changeset is
  rejected and both the genesis hash and the tip are unchanged.
- `apply_changeset_with_matching_genesis_extends_chain` — a `(0, Some(genesis))` entry plus a
  new higher block still applies.
- `apply_changeset_above_genesis_still_applies` — the guard is not over-broad.

```
cargo test -p bdk_chain --all-features                                  # ok, 100 tests, 0 failed
cargo clippy -p bdk_chain --all-features --all-targets -- -D warnings   # clean
cargo fmt --all -- --check                                              # clean
```

Mutation check: with the height-0 comparison neutered (`changeset.blocks.get(&1)`),
`apply_changeset_does_not_replace_genesis` fails with `assertion failed: result.is_err()`;
restoring the fix turns the file green again. So the new test really guards the fix.

### Notes

This change was implemented with AI assistance (OpenCode + muse-spark 1.3) and verified
locally as described above. The commit is not GPG-signed — no signing key is configured on
this machine; I am happy to re-push a signed commit if the project requires it.


wangzhengzhuo05:
Closing this out.

It has been waiting 5 days without any maintainer review, and we are winding down this contribution effort — a stale PR sitting in the queue costs maintainers attention without giving them a decision to make, so it is better closed than left open indefinitely.

If the direction is useful, the branch is intact on the fork (`fix/apply-changeset-genesis`, +65/−0 across 2 file(s)) and can be picked up as-is.

What it contains: *fix(chain): reject a changeset that replaces the genesis block* — Solves #2309

It was verified locally before opening (target test suite green, plus a mutation check that reverts only the fix and confirms the new regression test fails). Thanks for the time.

PR #2313, opened 2026-09-14
fix(file_store): reject an entry length prefix larger than the file

### LINKED ISSUE #2306: file_store: oversized entry length prefix panics `Store::load` instead of returning an error
**Describe the bug**

`EntryIter::next` decodes each entry with `bincode_options().deserialize_from(..)`, and `bincode_options()` (`crates/file_store/src/lib.rs`) sets no size limit. A length prefix read from the file is therefore passed straight to the buffer allocation. An entry declaring a `String` length of `u64::MAX` panics with `capacity overflow` inside `Store::load`; smaller but still oversized lengths attempt a huge allocation instead of returning `StoreError::Bincode`.

This affects loading a corrupted or hand-edited store file. #2258 appears to address this by bounding the frame allocation.

This issue was found by AI.

**To Reproduce**

Add `crates/file_store/tests/test_oversized_len.rs` and run `cargo test -p bdk_file_store --test test_oversized_len`:

```rust
use bdk_file_store::Store;
use std::collections::BTreeSet;

const MAGIC: &[u8] = b"bdk_test_magic";

#[test]
fn load_returns_error_on_oversized_length_prefix() {
    let dir = tempfile::tempdir().unwrap();
    let path = dir.path().join("db");

    let mut bytes = MAGIC.to_vec();
    bytes.push(1); // BTreeSet length: 1 element
    bytes.push(253); // bincode varint tag: u64 follows
    bytes.extend_from_slice(&u64::MAX.to_le_bytes()); // String length: u64::MAX
    std::fs::write(&path, &bytes).unwrap();

    let result = Store::<BTreeSet<String>>::load(MAGIC, &path);
    assert!(result.is_err(), "load should fail with an error, not panic");
}
```

Current output:

```
thread 'load_returns_error_on_oversized_length_prefix' panicked at library/alloc/src/raw_vec/mod.rs:28:5:
capacity overflow
```

**Expected behavior**

Loading a store file whose entry declares a length that cannot be satisfied should return an error rather than panic or attempt an unbounded allocation.

wangzhengzhuo05:
Fixes #2306

## What

`EntryIter::next` decoded each entry with `bincode_options().deserialize_from(..)` and no size
limit, so a length prefix read from the file was passed straight to the buffer allocation. A
store file whose entry declares a `String` length of `u64::MAX` panicked inside `Store::load`:

```
thread 'load_returns_error_on_oversized_length_prefix' panicked at library/alloc/src/raw_vec/mod.rs:28:5:
capacity overflow
```

## How

Each entry is now decoded with the deserializer bounded by the bytes remaining in the file,
multiplied by nine:

- bincode charges `size_of::<u64>()` (8 bytes) for every length literal it decodes, even when the
  varint encoding occupies a single byte on disk, and then charges the payload length again. A
  valid entry therefore charges at most `8 * bytes + bytes`.
- Every length literal occupies at least one byte on disk, so the total charge can never exceed
  nine times the bytes remaining — the bound cannot reject a store that fits in the file.
- Any length prefix that cannot be satisfied now fails with `bincode::ErrorKind::SizeLimit`
  *before* an allocation is attempted, and the existing error arm maps that to
  `StoreError::Bincode`.
- `remaining == 0` (end of file) short-circuits to the existing clean-EOF return instead of
  handing the deserializer a zero limit.

Only `crates/file_store/src/entry_iter.rs` changes; `lib.rs` and `store.rs` are untouched.

## Tests

Added `crates/file_store/tests/test_oversized_entry_len.rs`:

- `load_returns_error_on_oversized_length_prefix` — the reproduction from the issue: `load`
  returns an error instead of panicking.
- `valid_store_still_round_trips` — create/append/load two changesets, so the bound is not
  tighter than a real store needs.

```
cargo test -p bdk_file_store                                 # 10 + 2 + 1 passed
cargo clippy -p bdk_file_store --all-targets -- -D warnings   # clean
cargo fmt --all -- --check                                    # clean
```

Mutation check: with this change reverted,
`load_returns_error_on_oversized_length_prefix` panics with `capacity overflow`; with the change
applied, it passes.

## Note on #2258

#2258 (replace bincode with postcard) also bounds the frame allocation and rewrites this same
file, so it supersedes this fix if it lands first. This change is scoped to the current framing
and is independently correct and testable today; I am happy to rebase or drop it in favour of
#2258 if that is the preferred path.

Implemented with AI assistance (OpenCode + muse-spark 1.3); the reproduction, the full crate
test suite, clippy, rustfmt and the mutation check above were run locally and reviewed before
submission.


nymius:
Hi @wangzhengzhuo05, `bincode` must be replaced because is unmaintained. `postcard` is the best candidate we have so far.
As this is the priority, this PR is not going to get merged before #2258, which would make this PR obsolete.
I recommend you to close this PR and focus on review of #2258 or look for other issues to fix in BDK, if you are up to it.
Thanks!

wangzhengzhuo05:
Thanks @nymius — understood, closing as you recommend.

With `bincode` being replaced by `postcard` in #2258, the bound this PR adds would simply be re-implemented against the new format, so keeping it open only adds review noise to your queue.

For the record, the local evidence from this branch (in case it is useful when the migration lands):

- The bug: `EntryIter::next` deserialized each entry's length prefix with no upper bound, so a corrupt `u64::MAX` prefix panicked with `capacity overflow` inside `Store::load` instead of failing cleanly.
- The fix bounded the read by `remaining * 9` (provably >= the charge of any valid entry's length literal, since `bincode` charges `size_of::<u64>()` even for a 1-byte varint — a plain `remaining` bound broke 6 existing tests) plus an explicit clean-EOF short-circuit for `remaining == 0`.
- Verified: `cargo test -p bdk_file_store` green, `clippy -D warnings` + `fmt` clean, and a mutation check (`git checkout HEAD~1 -- entry_iter.rs`) reproduced the original `capacity overflow` panic and failed the new regression test.

Happy to re-target this against the `postcard` migration once it is in, if that would be useful.

